PR #2064, opened 2021-07-22
Fix lucene filtering

### LINKED ISSUE #2038: Data Provider' Filter for Boolean type fields stopped working
Hi All,

v. 0.9.78

Data Provider' Filtering by Date/Time stopped working across all my apps. Returns always 0 records. 
No changes in formulas/fields/records were made. Even tried with the only Date/Time filed.

This is example where formula is using:
Filter_Date

P.S. It doesn't matter if an app is existing or re-imported.

Please solve the issue.

aptkingston:
## Description
This PR fixes a couple of issues with server side searching, particularly of internal tables using lucene. Lucene is very picky about what values should and should not be wrapped in quotes, and this PR wraps all required values and leaves other unwrapped. This also helps with server side searching of external SQL DB's as filter values are parsed to the correct type depending on the table schema before being sent to the server.

Fixes https://github.com/Budibase/budibase/issues/2038.

### DIFF
--- a/packages/server/src/api/controllers/row/internalSearch.js
+++ b/packages/server/src/api/controllers/row/internalSearch.js
@@ -335,6 +335,7 @@ exports.paginatedSearch = async (appId, query, params) => {
   }
   limit = Math.min(limit, 200)
   const search = new QueryBuilder(appId, query)
+    .setVersion(params.version)
     .setTable(params.tableId)
     .setSort(params.sort)
     .setSortOrder(params.sortOrder)
--- a/packages/server/src/api/controllers/row/internalSearch.js
+++ b/packages/server/src/api/controllers/row/internalSearch.js
@@ -9,25 +9,22 @@ const fetch = require("node-fetch")
  * @param options The preprocess options
  * @returns {string|*}
  */
-const preprocess = (
-  value,
-  options = { escape: true, lowercase: true, wrap: true }
-) => {
+const preprocess = (value, { escape, lowercase, wrap } = {}) => {
   // Determine if type needs wrapped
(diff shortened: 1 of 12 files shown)


packages/server/src/api/controllers/row/internalSearch.js
```diff
@@ -3,13 +3,35 @@ const env = require("../../../environment")
 const fetch = require("node-fetch")
 
 /**
- * Escapes any characters in a string which lucene searches require to be
- * escaped.
- * @param value The value to escape
- * @returns {string}
+ * Preprocesses a value before going into a lucene search.
+ * Transforms strings to lowercase and wraps strings and bools in quotes.
```
Rory-Powell, line 7: Nice description thank you 👍

Rory-Powell (approved):

packages/server/src/api/controllers/row/internalSearch.js
```diff
@@ -124,7 +149,14 @@ class QueryBuilder {
     // Construct the actual lucene search query string from JSON structure
     if (this.query.string) {
       build(this.query.string, (key, value) => {
-        return value ? `${key}:${luceneEscape(value.toLowerCase())}*` : null
+        if (!value) {
+          return null
+        }
+        value = preprocess(value, {
+          escape: true,
```
mike12345567, line 156: since all the pre-process stuff is true by default, do we need to specify it again here? By the look of things we don't actually need the options at all cause nothing actually sets anything to false, everything is always doing everything.

Btw love the pre-process thing, was gonna do this myself, nice little improvement!

PR #2066, opened 2021-07-22
Add lerna prereleases

Rory-Powell:
## Description
Adding in a prerelease publish on the develop branch to publish packages to NPM. 

This uses the lerna prerelease functionality:
`lerna publish prerelease --yes --force-publish --dist-tag develop`

This will create a new version based on the current lerna version, adding a suffix `-alpha.0`, then `-alpha.1` and so on until the version is released. 

The tag in NPM has been updated to use `develop` in this scenario instead of the default `latest`. 
https://github.com/lerna/lerna/tree/main/commands/publish#--pre-dist-tag-tag

Example repo / package
- https://www.npmjs.com/package/@rpowell/lerna-npm-1
- https://github.com/Rory-Powell/lerna-npm

Example commit history:
Screenshot 2021-07-22 at 17 18 31

### DIFF
--- a/.github/workflows/release-develop.yml
+++ b/.github/workflows/release-develop.yml
@@ -1,0 +1,52 @@
+name: Budibase Release
+
+on: 
+ push:
+    branches:
+      - develop
+
+env:
+  POSTHOG_TOKEN: ${{ secrets.POSTHOG_TOKEN }}
+  POSTHOG_URL: ${{ secrets.POSTHOG_URL }}
+  SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+      
+jobs:
+  release:
+    runs-on: ubuntu-latest 
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - run: yarn 
+      - run: yarn bootstrap 
+      - run: yarn lint 
+      - run: yarn build 
+      - run: yarn test
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-develop.yml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - run: yarn 
+      - run: yarn bootstrap 
+      - run: yarn lint 
+      - run: yarn build 
+      - run: yarn test
+
+      - name: Configure AWS Credentials
```
Rory-Powell, line 28: Not sure if this is needed?

.github/workflows/release-develop.yml
```diff
@@ -0,0 +1,52 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: eu-west-1
+
+      - name: Publish budibase packages to NPM
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: | 
+          # setup the username and email. I tend to use 'GitHub Actions Bot' with no email by default
+          git config user.name "Budibase Release Bot"
```
Rory-Powell, line 40: `Budibase Develop Release Bot` or keep the same?

mike12345567: Probably use the develop release bot, just for the differentiation in commits.

.github/workflows/release-develop.yml
mike12345567: Not sure if its needed either, we release to Docker hub rather than ECR, so I think the AWS creds are kind of pointless now.

Rory-Powell: That's what I was thinking - I'll make a note to revisit it for both the master and develop release and leave it in for now so they are at least consistent

PR #2072, opened 2021-07-23
Mark google activated by default for old configs

Rory-Powell:
## Description
The google button was't showing in staging because the existing config did not have the enabled flag. 
Enable config by default when it does not exists as this is an old one.


mike12345567 (approved):
LGTM!

PR #2074, opened 2021-07-23
Fix/oidc fixes

Rory-Powell:
## Description
Two fixes:
1. Require https on oidc callback in production
   - Previously we reused the scheme of the request to work out the protocol
   - Envoy relays traffic with http so the original https value was being dropped
   - oidc providers only allow http for localhost, so use a environment flag to set the protocol instead
2. Set the google auth form `activated` field to true when it doesn't exist
   - Existing configs were showing as activated false


PClmnt (approved):
LGTM

PR #2076, opened 2021-07-23
Multi-tenancy/organisations

mike12345567:
## Description
Created as a draft to run Cypress etc and start to get review comments, this should not be merged until after develop has been released.

For the sandbox we need the ability to separate users from each other, the goal with this PR is to completely separate user databases, meaning that no information of any user within the sandbox is stored in the same DB (unless they are invited/join the same organisation).

This work will lead onto the accounts portal that is currently being designed/discussed as part of the sandbox onboarding flow.

There are a few key changes that this introduces:
1. The `MULTI_TENANCY` environment variable supported by the worker and server services, if these flags are not set then apps will not be created in multi-tenancy mode, the login pages won't be updated, etc.
2. If no tenant is specified it is considered the `default` tenant, this mechanism means that this feature is backwards compatible.
3. The login, reset password and admin creation all have a page in front of them in the multi-tenancy update, the ability to set the organisation you are logging into. This may never be seen in the sandbox as the accounts portal will do the login, it is also possible to skip the organisation setting phase by specifying it in the query string, like `login?tenantId=orgName`.
4. App IDs now contain the `tenantId`, as well as the app metadata doc, this means that during app lookup we only need to lookup the list of databases to filter it down to the ones within the tenant (no need to look up all the app metadata docs, which is unlikely to be performant at scale). We can also consider building a redis cache of all the app databases for performance purposes.
5. Automations now have to look up the app they are within, so that they can find the `tenantId`, this allows them to look up global config (for say sending an SMTP email).
6. Everything that is not logged in, or not within an app is not within the scope of a tenant, so all of the auth endpoints had to be updated to include the `tenantId` as part of the URL params.
7. OIDC/Google pre-auth and post-auth had to be updated, specifically around the `callbackURL` as now they needed to include what tenant they were logging into. There are fallbacks in place which assume the default/no tenant for systems that are already out in the field.
8. User invite and auth sessions now include the `tenantId` when stored in Redis, as this is required when creating the user (to lookup the correct global DB).
9. Everywhere in the system it is assumed, whether in multi-tenancy mode or not that a tenantId may be null, in which case it assumes the default DB. The default will always be our old/current DB `global-db`.
10. A platform DB has been added, which is a new totally global DB, this doesn't replace the current global DB for backwards compat reasons. The platform DB has a list of tenants and then stores user IDs/emails as docs containing the tenant they are within (in case this needs to be looked up).
11. A tenancy API has been put together which can be used down the line to list some information about the tenants within the system. This hasn't been implemented just yet as I thought I'd rather get this feature in so people can play with it before putting some info around it.

## Screenshots
Org set page
image

Added button to login page to go back if desired
image

Added back button to reset (to login)
image


packages/worker/src/api/routes/global/auth.js
```diff
@@ -0,0 +1,61 @@
...
+  .get(
+    "/api/global/auth/:tenantId/oidc/configs/:configId",
+    authController.oidcPreAuth
+  )
+  .get("/api/global/auth/:tenantId/oidc/callback", authController.oidcAuth)
+  // deprecated - used by the default system before tenancy
+  .get("/api/global/auth/google", authController.googlePreAuth)
+  .get("/api/global/auth/google/callback", authController.googleAuth)
+  .get("/api/global/auth/oidc/configs/:configId", authController.oidcPreAuth)
+  .get("/api/global/auth/oidc/callback", authController.oidcAuth)
```
Rory-Powell, line 59: I think we could get away with the two callback urls only here. 
Should we keep them as `/api/admin/` as before for backwards compatibility?

mike12345567: ah yes, I forgot about the admin backwards compat too, will add it in!

joebudi:
@all-contributors please add @Rory-Powell  for infrastructure, tests and code

packages/auth/src/tenancy/FunctionContext.js
```diff
@@ -0,0 +1,73 @@
+const cls = require("cls-hooked")
+const { newid } = require("../hashing")
+
+const REQUEST_ID_KEY = "requestId"
+
+class FunctionContext {
```
shogunpurple, line 6: I like this abstraction for using `cls`. We will likely start to lean on this more heavily in future for any kind of system wide state like correlation IDs of requests and other production information.

shogunpurple (approved):
Great work - this is so fundamental and it's been really nicely encapsulted.

packages/auth/src/tenancy/FunctionContext.js
mike12345567: This was an idea put forward by @Rory-Powell based on his previous experience implementing tenancy systems, I think it works quite well for things that are supposed to be hidden!

Issue #2077, opened 2021-07-23
Unable to create a new role without inheriting from another one


mslourens:
**Describe the bug**
Creating a role without inheriting an existing role is possible in the frontend but throws a 400 Bad Request in the backend.

**To Reproduce**
1. Go to the users table
2. Click on 'Edit roles'
3. Create a role without selecting an inherited role
4. Click on save
5. A generic server error is thrown: "Error saving role"

**Expected behavior**
I expected a clear client side error message when you don't select an inherited role, since this field is required on the server side, like the name field.

**Desktop (please complete the following information):**
 - OS: Linux (Pop!_OS 21.04)
 - Browser: tested on Vivaldi, Firefox
 - Version: master branch

mike12345567:
Hi @mslourens thanks for the report! I've raised a PR for this here: #2086

mike12345567:
This has been merged to develop.

PR #2086, opened 2021-07-25
Fixes for issue #2077

mike12345567:
## Description
Roles always must inherit from another role, it would be very problematic if we had floating roles. The edit roles modal didn't really show the user what was required and what wasn't allowed, so I made some changes here to make it more obvious.

## Screenshots
Editing builtin role, is now read only
image

Creating/editing a custom role, by default has a basic role to inherit and a permission, the button will also be disabled if all the required options have not been input
image


shogunpurple (approved):
LGTM!

Issue #2079, opened 2021-07-24
Logout button action stopped working


sergey-dme:
Hey all,

Thanks for the fixes arrived with v0.9.79.

But currently Logout action does not work.
Kindly ask you to fix it.. please

Thanks.

mike12345567:
Hi @sergey-dme - this is being addressed in #2087 and merged to `develop`, which is to be released this week.

PR #2087, opened 2021-07-25
Logout button handler

mike12345567:
## Description
Log out button handler was no longer implemented, added back this functionality for backwards compatibility, the frontend also still supports this as an action which can be created so it should be supported by the client.

This was described in issue #2079.

### DIFF
--- a/packages/client/src/store/app.js
+++ b/packages/client/src/store/app.js
@@ -6,8 +6,15 @@ const createAppStore = () => {
 
   // Fetches the app definition including screens, layouts and theme
   const fetchAppDefinition = async () => {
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
Nice one! The original bug here was actually leaving the logout action in the front end at all, because client apps aren't concerned with logging in or out any more. But I guess it's not a huge deal to allow people to still do this.

In terms of my commits there, I pulled this down and tested it and found that the client crashes if you have a public home page and log out because of a bug with refreshing the app package. So I fixed that, but then still thought it's weird that if you log out you can still be in the app just as a public user after logging out. So I think it's better to just clean the cookie and do a full page reload back to the login page.

PR #2078, opened 2021-07-23
display role names instead of _id

mslourens:
## Description
Fixes #2062: display role names instead of ids. 

## Screenshots
image

image

PR #2107, opened 2021-07-25
Add dedicated renderer for roles tags

Rory-Powell:
## Description
A regression was added in https://github.com/Budibase/budibase/pull/2078 that resulted in the group tag being rendered as `undefined`. This was caused by the tags renderer behaviour being changed to require an object instead of an array of values. The scenario where an object is required (nested `name` field in roles) has been brought out into its own component to perform custom logic and delegates to the basic tag renderer. 

## Screenshots
Before

Screenshot 2021-07-25 at 21 17 20
Screenshot 2021-07-25 at 21 17 26

After
Screenshot 2021-07-25 at 21 17 33
Screenshot 2021-07-25 at 21 17 38


shogunpurple (approved):
LGTM!

PR #2114, opened 2021-07-26
Clear form action

### LINKED ISSUE #2102: Clear form action
When setting up button actions, I want to clear my form values by attaching an action to a button, so that I can start the form from scratch.

**COS**
- "Clear form" button action, which clears all inputs in the current form context

mslourens:
Addresses https://github.com/Budibase/budibase/issues/2102

## Description
Add the `Clear Form` action to the button actions. Clicking this button will clear the form.  I used the ValidateForm as an example how to create new actions for forms. 

## Screenshots
image

### DIFF
--- a/packages/standard-components/src/forms/InnerForm.svelte
+++ b/packages/standard-components/src/forms/InnerForm.svelte
@@ -119,20 +119,19 @@
 
     const clearValue = () => {
       const { fieldState } = fieldMap[field]
+      const newValue = initialValues[field] ?? defaultValue;
       fieldState.update(state => {
-        state.value = defaultValue
+        state.value = newValue
         state.error = null
         return state
       })
 
       formState.update(state => {
-        state.values = { ...state.values, [field]: defaultValue }
+        state.values = { ...state.values, [field]: newValue }
         delete state.errors[field]
         state.valid = Object.keys(state.errors).length === 0
         return state
       })
-
-      return true
     }
 
     return {
--- a/packages/standard-components/src/forms/InnerForm.svelte
+++ b/packages/standard-components/src/forms/InnerForm.svelte
@@ -119,7 +119,7 @@
 
     const clearValue = () => {
       const { fieldState } = fieldMap[field]
-      const newValue = initialValues[field] ?? defaultValue;
+      const newValue = initialValues[field] ?? defaultValue
       fieldState.update(state => {
         state.value = newValue
         state.error = null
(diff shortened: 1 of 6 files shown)


packages/standard-components/src/forms/InnerForm.svelte
```diff
@@ -108,8 +116,28 @@
 
       return !newError
     }
+
+    const clearValue = () => {
+      const { fieldState } = fieldMap[field]
```
aptkingston, line 121: I would maybe declare a new variable here for the value we're going to revert to, to save duplication down below. We need to check if there's an initial value here, so something like:
`const newValue = initialValues[field] ?? defaultValue`

packages/standard-components/src/forms/InnerForm.svelte
```diff
@@ -108,8 +116,28 @@
...
+      fieldState.update(state => {
+        state.value = defaultValue
```
aptkingston, line 123: Using the new variable declared above, we can update this to:
`state.value = newValue`
to accound for initial values and default values.

packages/standard-components/src/forms/InnerForm.svelte
```diff
@@ -108,8 +116,28 @@
...
+        state.error = null
+        return state
+      })
+
+      formState.update(state => {
+        state.values = { ...state.values, [field]: defaultValue }
```
aptkingston, line 129: Using the new variable declared above, we can update this to:
`state.values = { ...state.values, [field]: newValue }`
to accound for initial values and default values.

packages/standard-components/src/forms/InnerForm.svelte
```diff
@@ -108,8 +116,28 @@
...
+        delete state.errors[field]
+        state.valid = Object.keys(state.errors).length === 0
+        return state
+      })
+
+      return true
```
aptkingston, line 135: We don't need to return anything here since we never need the result of this function, we could just remove this line.

aptkingston (changes requested):
Thanks so much for the contribution @mslourens! This is pretty much perfect to merge as it is, but I've just spotted a couple of tiny things which would be good to address before merging. Awesome work! 👍

aptkingston:
Those test failures just look like random failures, so I've ran them again for you. Should pass this time!

aptkingston (approved):
Looks good to me 👍

joebudi:
@all-contributors please add @mslourens for tests and code

PR #2115, opened 2021-07-26
Conditional UI

### LINKED ISSUE #1644: Conditional UI - Bindings
When using the design UI, I want to be able to control all component properties with bindings, so that I have the ability to control conditional visibility.

**Conditions** 
- Move all possible component props to handlebars to allow binding based on data

aptkingston:
## Description
This PR adds "conditional UI" - the concept of being able to dynamically show, hide or update parts of your UI in response to state. State in this sense refers to the client app context.

Design doc: https://docs.google.com/document/d/1uxH5_jYpHXO0-92q1hwPdHprFJ_HLX9k4pb0_uShaBc/edit

Closes #1644.
As discussed in https://github.com/Budibase/budibase/discussions/1617, this allows all current form component settings to be redefined conditionally, and is fully generic and flexible so that as we add form enhancements, all those new settings will also be conditionally redefinable.

### Oveview
Conditions are configured by a new settings section available on every single component:
image

The core concepts of conditions are just "if" statements.
`Do [action] if [value] [operator] [reference value]`.
image

The field for `value` is always a bindable input, as 99% of the time this will want to be bound to a value in state. This could be a form field, some data, a HBS expression or some current user property; any bindable property.
The `operator` field provides all the options that our filters provide - equals, not equals, less than, more than, starts with, is empty, is not empty - all depending on the field type being compared.
Both the `operator` and `reference value` fields are type aware. There are a few reasons that we need to be type aware here:
numerical comparison is different than string comparison, only certain operators are relevant for any given data type and most crucially - the `value` field output by HBS is always a string (unless the literal helper is used), which means it cannot be properaly evaluated against the reference field unless the types are the same. The impact of type awareness is that the user must simply pick what type they want to compare against. Strings and numbers offer bindable input fields to configure, while dates and booleans offer date pickers and a true/false select respectively. This is also more friendly for the user, as otherwise a condition like "if the date is later than 1st January" would require the user to hand type an ISO date string.

### Configuration
Conditional UI is broken into 3 discrete actions - "show component", "hide component" and "update setting".
The order of conditions matters, which is why they can be dragged to be reordered. If a "show" action followed by a "hide" action are both found to be true, the latter action will win. Similary if 2 updates are made to the same setting, the last one will be the one applied.

- Show component: if a "show component" action is defined, the component is hidden unless the condition is met, in which case it is shown. If multiple "show component" actions are defined, only one of them needs to be active to show the component.
- Hide component: if a "hide component" action is defined, the component is shown unless the condition is met, in which case it is hidden. If multiple "hide component" actions are defined, only one of them needs to be active to hide the component". If both "show component" and "hide component" actions are defined, the component is hidden by default.
- Update setting: this action requires 2 parameters - which setting to update and what new value to update it to if the condition is met. All settings are configurable. This means even things like data provider sources, button actions and data provider filtering is all conditionally redefinable. Some cool examples are shown below of what you can do with this. The component rendered to allow the choosing of the new value is the same component that would normally be shown in the settings panel. This means that for updating colour you get a colour picker, for updating button actions you get the same "Define actions" button - it's seamless for the user.

## Screenshots
Here's an example drawer showing a bunch of conditions. These are the conditions in the button shown in the first video below.
image

Here's conditional UI in action. This is made with a single button - every setting (including the click action) is redefined depending if the "todo" is done or not.
46Qlq1

Here's another video showing a data provider being conditionally shown, with the source depending on the value entered.
z6v1oY


joebudi:
That was quick! Great work and nice write-up.

kevmodrome (approved):
LGTM

mike12345567:
This has to be one of my favourite features to go into Budibase in a while, can't wait to really put it through its paces! Such a big feature 🚀

shogunpurple (approved):
LGTM, great job. Great context as well in the description - it's so cool to see more interactive elements like this going into budibase. 🚀

Issue #2119, opened 2021-07-26
[Feature request] Pressing Delete should remove the selected component


mslourens:
Instead of having to click the three options dots next to the component and select the delete option from the context menu, it would me much quicker to use the delete button.

PR #2120, opened 2021-07-26
Delete component with keyboard

mslourens:
## Description
Fixes issue #2119: as a user I want to delete a component using the keyboard (backspace and delete button).


aptkingston:
Hey @mslourens, thanks again for the contribution! The bit about cleaning up the event listeners by itself is actually a good catch. I like this functionality, but since it's a global event listener on the iframe it prevents using either backspace or delete keys in the preview - which is something I use quite often when typing in input fields in the preview. Maybe binding it to only the delete key would be an option, but then I use the delete key too to clear date pickers in the preview. I'll discuss this with the team 👍

mslourens:
ah, I didn't think of that. Maybe we could detect if an input has focus. If not, the delete confirmation popups up.

aptkingston (approved):
LGTM! Thanks for the changes.

PR #2131, opened 2021-07-27
convert automation icons into toggle

### LINKED ISSUE #2021: Make automation activation/deactivation more clear
A common issue many users have with the automation system is that its not clear that it needs to be activated for it to work in production.

The small pause and play buttons often go un-noticed, due to their size and coloration, we could probably do with some sort of warning to state that an automation hasn't been activated. We could even perhaps have the automation be activated by default as it is running within the development app, not the production app, so there isn't really any risk.

mslourens:
## Description
Partly Fixes #2021. The issue is about a full redesign of the Automate page, this issue just addresses the tiny play/pause buttons, which it converts into a toggle.

## Screenshots
image

### DIFF
--- a/packages/bbui/src/Form/Core/Switch.svelte
+++ b/packages/bbui/src/Form/Core/Switch.svelte
@@ -6,15 +6,18 @@
   export let id = null
   export let text = null
   export let disabled = false
+  export let dataCy = null
 
   const dispatch = createEventDispatcher()
   const onChange = event => {
     dispatch("change", event.target.checked)
   }
+
 </script>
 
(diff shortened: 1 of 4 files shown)


Rory-Powell (approved):
LGTM - this is a more intuitive UX. 

Looks to be a linting issue with the build, you may be able to fix that with 
```
yarn lint:fix
```

Rory-Powell:
Cc: @joebudi for UX input

joebudi:
Looks good to me. Nice work.

I would like to highlight, in the next few releases we will redesign automations and this change may be superseded.

PR #2135, opened 2021-07-27
rename an app

### LINKED ISSUE #1876: Renaming an App
### Discussed in https://github.com/Budibase/budibase/discussions/1868

<sup>Originally posted by **edc1970** June 29, 2021</sup>
Is there a way to rename an app? </div>

mslourens:
## Description
Fixes #1876: add the ability to rename an app. Like the theme property we just update the name property of the app. The name is checked against previous published apps, like in the CreateAppModal. Only when you changed the name will the validation kick in, otherwise you'll always see the error that another app with the same name exists

### The app context menu in the portal page
image

### The update modal, which contains just the name:
image

### After changing the name to the same name:
image

### DIFF
--- a/packages/builder/src/components/start/AppCard.svelte
+++ b/packages/builder/src/components/start/AppCard.svelte
@@ -15,6 +15,7 @@
   export let exportApp
   export let viewApp
   export let editApp
+  export let updateApp
   export let deleteApp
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM! Nice work

PR #2175, opened 2021-07-29
Sync profile picture when using SSO

Rory-Powell:
## Description
Sync profile picture from SSO when logging in

## Screenshots
When `pictureUrl` is present
Screenshot 2021-07-29 at 14 48 59

When `pictureUrl` is not present
Screenshot 2021-07-29 at 15 38 55

### DIFF
--- a/packages/builder/src/pages/builder/apps/index.svelte
+++ b/packages/builder/src/pages/builder/apps/index.svelte
@@ -54,7 +54,11 @@
             </Layout>
             <ActionMenu align="right">
               <div slot="control" class="avatar">
-                <Avatar size="M" initials={$auth.initials} />
+                <Avatar
+                  size="M"
+                  initials={$auth.initials}
+                  url={$auth.user.pictureUrl}
+                />
                 <Icon size="XL" name="ChevronDown" />
               </div>
               <MenuItem icon="UserEdit" on:click={() => userInfoModal.show()}>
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
Nice! This is a nifty wee feature 💯

packages/builder/src/pages/builder/apps/index.svelte
```diff
@@ -54,7 +54,11 @@
             </Layout>
             <ActionMenu align="right">
               <div slot="control" class="avatar">
-                <Avatar size="M" initials={$auth.initials} />
+                <Avatar
+                  size="M"
+                  initials={$auth.initials}
+                  url={$auth.user.pictureUrl}
```
kevmodrome, line 60: Is there always a pictureUrl key on the user object? Might want to add a wee `?.` there

packages/builder/src/pages/builder/portal/_layout.svelte
```diff
@@ -100,7 +100,11 @@
         <div />
         <ActionMenu align="right">
           <div slot="control" class="avatar">
-            <Avatar size="M" initials={$auth.initials} />
+            <Avatar
+              size="M"
+              initials={$auth.initials}
+              url={$auth.user.pictureUrl}
```
kevmodrome, line 106: Same as above

kevmodrome (approved):
LGTM - just two potential nits

packages/builder/src/pages/builder/apps/index.svelte
Rory-Powell: there might not be, it can evaluate to `undefined` and will be ignored by the `Avatar` component
https://github.com/Budibase/budibase/blob/feature/sso-sync-picture/packages/bbui/src/Avatar/Avatar.svelte#L18

Updated the screenshots to show what happens when a user doesn't have the `pictureUrl` attribute

PR #2182, opened 2021-07-29
Add watchtower hooks

Rory-Powell:
## Description
Add scripts that hook into the watchtower lifecycle when checking/updating a container. 
More details: https://containrrr.dev/watchtower/lifecycle-hooks/

Write to a new file `watchtower-hook.json` to indicate which hook has been executed most recently.
In future these scripts could evolve to do more or become node scripts. 

This enables follow up work to read the hook file via the api and provide a more responsive updates page.

### DIFF
--- a/packages/server/.gitignore
+++ b/packages/server/.gitignore
@@ -7,3 +7,4 @@ public/
 db/dev.db/
 dist
 coverage/
+watchtower-hook.json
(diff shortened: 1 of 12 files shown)


shogunpurple (approved):
LGTM!

PR #2183, opened 2021-07-29
POC: Tenancy Context Abstraction 

Rory-Powell:
## Description

Changes:
- Add a module for interacting with multitenancy
   - `packages/auth/src/tenancy`
- Add tenancy context
   - Use nodes async hooks (provided by cls-hooked library)
   - Stores a tenant id across a chain of calls
   - Integrate tenancy context with retrieving global db
- Add tenancy middleware
   - Pre-auth / request
      - Set tenant context using the incoming request (url parameter) 
   - Post-auth / session
      - Set tenant context using the session
   - Ensures tenant id is always set in context before entering application code
   - Accept a non tenancy list to bypass (see below)
- Revert global -> admin name change
   - Keep global for global non-tenant related calls
      - e.g. flags, self 
 - Integrate middleware with `worker` module. `server` is TBD.


Rory-Powell:
@mike12345567 here's where I got to on a tenancy abstraction. the main ideas are outlined in the description - still some work tbd around the server package

mike12345567:
I think this is interesting, it does hide away the tenancy stuff quite well, although it does make it a little bit harder to work out what is worried about tenancy information and what is not. Hiding the tenancy stuff away though is probably for the best, we probably shouldn't have to deal with it too much in the future all being well.

Not sure I agree with the flip back to to `admin` instead of the global namespace, because now all the authentication endpoints and so on fall under an admin namespace, this was very confusing to begin with and we definitely need to solve that pre Budibase as a backend, admittedly it didn't need to be thrown in here, but there was already so many changes occurring to the endpoints that it seemed like a good time to do it. I think the global namespace is really good for users etc, especially if we are trying to hide the tenancy stuff away, as its a global DB all this stuff is stored within and it gives a concept of global information and app level information. Flags and self I think happily fall within that domain too.

Overall I think this is a pretty nice cleanup of the feature, it hides the stuff away pretty well. I wouldn't like to use this for anything beyond something like multi-tenancy as I feel it abstracts the issue almost too much, to the point at which the code is no longer self explanatory, but thats likely a good thing for multi-tenancy as its supposed to be hidden away under the hood.

Rory-Powell:
Thanks mike! 

Agreed this type of approach shouldn't be our norm, here it works well because it's a similar problem to access control and authentication that lends itself well to middleware. I think the most important things we can get from this style is defining rules, where we can say what is / isn't tenant aware, plus enforce those rules before requests make it near application code - it should then be really difficult for future maintainers to violate tenancy. 

RE: Endpoints, I get you now. The `global` in your changes refers to anything in the `{tenant}_global` db. I think that makes sense. I was approaching it as global = system level, admin = inside a tenant. 

This is the approach I was working with for the middleware rules in worker

- **Not Tenant Aware**
   - Does not require and ignores tenant attributes in the session / request
   - May or may not require authentication
   - Currently:
      - `/api/global/*`
      - `/api/admin/users/self`

- **Public** (currently a subset of  'admin' and 'global' in this pr)
   - Does not require authentication
   - May or may not be tenant aware
   - Currently
      - `/api/admin/auth/*`
      - `/api/admin/configs/public`
      - `/api/global/flags`
      - `/api/admin/configs/checklist`
      - `/api/admin/users/init`
      - `/api/admin/users/invite/accept`

- **Private**
   - Must be authenticated
   - Enforces that a `tenantId` is set in the session
   - Currently
      - Every endpoint in the worker that is not in the public definitions 
      - Every endpoint in the worker that is not in the no tenant definitions

An idea

- revert my `global` -> `admin` change
- use `system` in place of my usage of global, basically just the flags endpoint

public = `/api/public/*` (requires tenant in request)
private = `/api/private/*` (requires tenant in session)
system  = `/api/system/*` 
no tenant = [`/api/system/*`, `/api/admin/users/self` ]

ideally also aligning by package as well, although not required
```
api/routes/
 - public/
    - auth.js
    - configs.js
 - private/
   - configs.js
 - system/
    - flags.js
```

mike12345567:
sounds good to me, will have a look at creating those namespaces, its not hard to do in worker but would be a nightmare for the server due to the issues we discussed around it being very grey there what is required, so I'll maybe leave that for now and we can look at it in a future update, I'd like to get the tenancy stuff into develop asap and I don't think the namespace updates is really required, it extends a lot further than tenancy stuff, its really about how things are authenticated as well.

mike12345567:
closing this as rather than merging this decided to integrate some of the ideas it provided into the multi-tenancy codebase directly as this was the quickest way to get it up and running correctly. Most of the tenancy code added to the `auth` library has been moved over with some modifications to get it working.

Issue #2167, opened 2021-07-29
SQL relationship does not seem to work


mjashanks:
I may be misunderstanding something here, but when I create a relationship of "one", it does not seem to save my foreign key column properly.

Then, when I try to browse the table it looks for the key that I originally chose, but on the wrong table.

Here is my table structure:

![image](https://user-images.githubusercontent.com/3524181/127461929-d1c287be-d827-4cf0-ad9c-347452bd8d01.png)

Here is a demo:

https://user-images.githubusercontent.com/3524181/127461867-d41d8967-45fc-46ab-b200-ddda360d55e6.mp4

Using postgres.

shogunpurple:
@mjashanks do you have display columns set on both of your tables? Me and @mike12345567 were just looking into this and it seems to work if the display columns are set.

jheiselman:
I'm having the same issue that I was before. I have an external MySQL database I'm connected to. Two tables have a fairly basic structure:

host table

id | name
-- | --
1234 | testhost

host_blackout table with foreign key constraint to host.id from host_blackout.host_id

id | host_id | inserted
-- | -- | --
9876 | 1234 | 2021-07-30 10:20:00


The two tables in question have their display columns set. When I create the relationship in Budibase, I use the following settings:

Relationship type: One
From table: host
To table: host_blackout
Foreign Key: host_id
From table column: maintenance_since
To table column: host_name

Once I save the relationship, if I open it again, it appears that Budibase changed the foreign key to the "id" column in the host_blackout table. When I load up the host_blackout table view, the host_name column is null for each record and when I try to load up host table view, it just hangs.

Looking into the stdout in the container, the SQL is clearly using host_blackout.id in the left join statement. I can't seem to get it to accept the host_id column for that foreign key.

I'm running the latest image tag pulled as of this morning.

mike12345567:
Hi @jheiselman - just to check, have you updated your display columns? This is a bit confusing and we plan to update the UI to make this more obvious, but Budibase needs a display column for it to display the relationship between the two different tables.

The fact the foreign key is switching to the `id` field does sound like there is another bug, but just wondering if this is the case.

You can set it with the button here:
![image](https://user-images.githubusercontent.com/4407001/127688947-1ed3a6cd-e633-4fe4-a2fc-49bf423e4493.png)

jheiselman:
@mike12345567 yes, they are set

mike12345567:
[...]

PR #2199, opened 2021-07-30
Fix for SQL relationship issue

mike12345567:
## Description
Fix for issue #2167, the foreign key was not being set correctly in all scenarios for the from table in SQL relationships, our test data happened to hide this fact due to the foreign key and the key in from table having the same name.

Also included an explanation of the somewhat confusing structure that has to be used to replicate the structure used for Budibase internal relationships.


shogunpurple (approved):
Great job - LGTM. Think this is why we shouldn’t just use test data in future 😅

mike12345567:
Yeap, we need to get some good datasets that we can use to test this repeatedly!

PR #2220, opened 2021-08-02
Fix colour picker inconsistencies when app theme has never been changed

### LINKED ISSUE #2219: Color picker is inconsistent with app themes if the theme has never changed
**Describe the bug**
The color picker component is inconsistent with app themes if the theme has never changed.

**To Reproduce**
1. Create an app
2. Choose a text component
3. Notice the color in the settings bar is different from the one shown in the settings panel

This can be fixed by changing the app theme and changing it back,

aptkingston:
## Description
Fixes an issue with color pickers being inconsistent in new apps by ensuring a default app theme of "light" is globally set.
Closes #2219.

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -70,7 +70,7 @@ export const getFrontendStore = () => {
         url: application.url,
         layouts,
         screens,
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM!

Issue #2103, opened 2021-07-25
Open Screen In Modal


shogunpurple:
When I route to another screen, I want the option to open that screen in a modal on the same page, so that I do not have to navigate back and forth.

**COS**
- Some options on the "Navigate" action to allow for the opening of a screen in a modal, including modal title and width
- Svelte SPA router needs to account for the modal opening in real apps

Mitch-Budibase:
From testing this I am noticing an issue with nested iframes (opening a modal within an opened modal)

Explanation:

- I have a Home Screen, a second screen, and a third screen

- Home Screen has one button - opens the second screen in a modal

- Second screen has two buttons - 1 opens Home Screen in a modal, the other opens screen 3 in a modal

When I run my app, I click the button on the Home Screen and all is well (the second screen opens within a modal).
When I click either of the buttons within the second screen modal - It opens another modal but does not display anything. This second modal is slightly further down the screen than the first one.

A few other things to note:

- Clicking the full screen or close options within the second modal will close it and keep the first modal open

- Clicking outside the modals will cause both to close

I've attached a screenshot to illustrate the modal behaviour
![Screenshot 2021-08-24 at 13.35.51.png](https://images.zenhubusercontent.com/60fed9d54a973614368dfdfe/41dd3cd4-571f-4717-a61f-708c1622f6de)

aptkingston:
As we discussed, I'll update this so that opening another screen modal inside a modal updates the URL of the first modal instead. Should be easy to do using the same technique I've used for other aspects of the modal, by broadcasting an event to the real window. Hopefuly no one ever does this cursed design, but we should definitely handle it at least.

PR #2233, opened 2021-08-03
Open screen in modal

aptkingston:
## Description
This PR adds the ability to open any screens in modals. This is attached to the existing "Navigate To" button action, by way of a new checkbox setting which allows users to choose to open the screen in a modal rather than navigating to the screen.

### Opening a screen in a modal
Check the new checkbox in the button action to open the screen URL in a modal. This works for both internal and external URLs.
image

No other settings are needed. In the issue (#2103) we mentioned things like modal title or size, but after discussion with Joe about design and UX we've decided to keep the modal as clean as possible, with no title. The size is also opionated and narrow, highlighting its primary use case for detail screens.

### Using the screen modal
Clicking on this button action in an app will open a modal and load in the configured screen URL. When loading screens inside a modal the navigation section of the layout is hidden, simplifying the screen view and focussing it more on the screen content.
image

There are 2 buttons which float to the top right of the modal: one for going "full screen" which essentially means navigating to the modal URL in the real app, and one for closing the modal.

### Closing the screen modal
The screen modal can be closed in 3 different ways:
- Clicking outside the modal
- Using the dedicated close button
- Using the new "Close Screen Modal" action

A new button action has been added to explicitly close a screen modal, allowing buttons on detail screens to automatically close the modal after performing their actions. This feels very natural for things like saving a details page, or deleting a record. This is totally optional, and the screen modal will continue to run a full version of the app (just without the navigation bar) until it is closed.

image

### Modal awareness
The screen modal functions by iframing in the client app URL and appending a query param to let the new app know that it is inside a modal. Being aware that the app is inside the modal is important for a number of things
- Sharing datasource invalidations, i.e. allowing hot reloading of data in the parent window. If you change data in the screen modal iframe, that data will also be magically refreshed in the parent app. That means when you close a details page modal, the parent view is already up to date with the changes.
- Proxying notifications. All notifications dispatched by the screen modal iframe will instead by displayed in the parent app. This is important as otherwise notifications fired before closing the modal instantly disappear.

## Other features
- Converted client app notifications to spectrum themed ones which look much better
- Client apps will only ever show 1 notification at a time, fixing the issue of notification spam when doing lots of actions quickly


joebudi:
Super work @aptkingston. From an end-user UX perspective, this is a real game changer.

shogunpurple (approved):
Great work! This is such a killer feature for things like auto screens and CRUD workflows.

Issue #2209, opened 2021-08-02
Group by doesn't show values for formula columns


mpodshivalin:
**Describe the bug**
Group by doesn't show values for formula columns, it shows an empty string

**To Reproduce**
1. Create a table, e.g. "Contracts"
2. Create Payment Date column (datetime), and a Price column (number)
3. Create an auto column Payment Month with a formula `{{#if Payment Date}}{{date Payment Date "MM.YYYY" }}{{/if}}` (shows only MM.YYYY part of payment)
4. Create a View, which calculates price by sum, and groups by Payment month

It will show group as empty:
![image](https://user-images.githubusercontent.com/16121630/127861219-86cdc5a8-696d-4608-a17d-fe854a62c4b2.png)

**Expected behavior**
It should show Payment month, e.g. 07.2021, 08.2021, etc.
That way it would be possible to make an aggregated statistics about monthly sales and show it in graps. Or am I doing something wrong?

**Desktop (please complete the following information):**
 - OS: Fedora 34
 - Browser Firefox 90
 - Latest Budibase (docker-compose version)

mpodshivalin:
I've noticed that if you go to the row and explicitly save it, group by will have a value. So the problem may be narrowed down to the situation when the formula column was added after the initial row was created

mike12345567:
Hi @mpodshivalin this is something that is being looked at over the next while, theres a few things you can already do which can provide much deeper functionality and there is some work coming in the future as well.
1. Right now you can use helpers in our bindings, these are called Handlebars helpers, you can find some information about the functionality supported here: https://github.com/Budibase/handlebars-helpers
2. We are currently working on introducing JS everywhere, meaning that everywhere bindings are supported right now it will be possible to write a Javascript function to transform the data - making it more developer friendly.
3. If you are using an external SQL data source it is already possible to write custom SQL queries into this data, allowing you to produce very specific data formats through SQL. If you are using Postgres/MySQL you will get a very similar experience to what you have today, but with the ability to write SQL queries as well.
4. In terms of filtering data, you have the data provider filter functionality, available for internal and external data sources - with the help of bindings you can retrieve and filter your data in pretty much any format.
5. We are also planning to introduce JS based data transformers, this will allow you to essentially write a JS function which the data will be put through on the way out of the API, transforming it in whatever way you desire. This is similar to what views provide today, but it won't be as performant (as it won't be done in database) but comes with the benefit of having access to all Budibase features.

[...]

PR #2260, opened 2021-08-04
Formulas and relationships in views

mike12345567:
## Description
Fixes for issue #2209:

1. Formula field values were being saved, causing them to show up in views, didn't really matter as the value was normally overwritten on way out, but shouldn't save these values as its confusing.
2. Disallowing relationship and formula fields in view calculations
3. Disallowing formulas fields in view group by


shogunpurple (approved):
LGTM! Nice fix

Issue #2100, opened 2021-07-25
Default value on form inputs


shogunpurple:
When building forms in budibase, I want to be able to have a default value for each input field in the form, so that I can reduce the amount of steps required to fill out the form.

**COS**
    - Default value is a bindable HBS/JS field
    - Default value is applied on first render
    - Default value is not reapplied if the field is specifically cleared, evaluates as null etc

PR #2261, opened 2021-08-04
Default field values

aptkingston:
## Description
This PR adds a setting to specify default values for form fields. The setting to enter the default value is a bindable input which may present problems for certain input types (booleans and dates for example), so the form field components themselves try to be clever and parse a variety of inputs.

- `string`, `options` and `longform` default values are treated as they are
- `number` values are attempted to be parsed as floats
- `boolean` values are checked for real booleans (e.g. the `literal` helper) or compared to string `"true"`
- `datetime` values are checked for being real dates, numeric or string timestamps, or otherwise parsed as any sort of date string

Default values are not available on attachment or relationship fields as they do not make sense there. 
The default value is different than just the undefined or `null` value, so fields can still be manually cleared.
When using the "clear form" button action, forms are reset to their initial values so this will set fields back to their default values rather than empty every field. In my opinion this is the correct behaviour - it should reset the form back to the way you first found it.

Addresses #2100.

Few other fixes:
- Fixes a couple of issues with the latest merge with master
- Updated vite plugin svelte and its usage
- Fixed an issue with svelte global style rules in apex charts

## Screenshots
The new setting on fields for default values:
image

A form with default values:
image


shogunpurple (approved):
LGTM! Don't think there's anything we can do about the need to parse, so it's necessary

Issue #2193, opened 2021-07-30
Export doesnt work


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/2184

Export from and internal table fails: `{"message":"Missing/invalid DB name","status":500}`

PR #2272, opened 2021-08-04
Fix exporting data from table

mike12345567:
## Description
Fixing a few issues related to exporting data from a table, as follows:
1. #2193 exporting didn't work due to it not knowing the app ID, fixing this
2. Fixing the name of the exporting file, CSV files were always just named `export` or `export.txt` - making them `export.csv`
3. Fixing an issue with relationships being exported giving `[object Object]` - JSON stringifying it instead


shogunpurple (approved):
LGTM!

PR #2278, opened 2021-08-05
Fix/pin dev versions

shogunpurple:
## Description
Pins development versions when pushing to `develop`, so that pushes to master do not break the staging environment when the 2 get out of sync

- As part of the develop build, the `pinVersions` script will run, stripping the caret from all BB deps in the `packages` dir
- The docker images will then always pull the prerelease versions, even though a later master version may exist

## Screenshots


packages/builder/package.json
```diff
@@ -65,10 +65,10 @@
     }
   },
   "dependencies": {
-    "@budibase/bbui": "^0.9.99-alpha.2",
-    "@budibase/client": "^0.9.99-alpha.2",
+    "@budibase/bbui": "0.9.99-alpha.2",
```
mike12345567, line 68: Might need to unpin versions, looks like a few have pinned during testing!

mike12345567:
Love this, great approach, solves the problem without having to resort to permanent pinning! LGTM!

Rory-Powell:
Looks great marty. 

An additional step I think is needed here to ensure branches are up to date before merging to master, or else the pin could override the released version e.g. develop is 1 version behind because of a hotfix. 

Screenshot 2021-08-08 at 14 47 00

(this might already be the case, just checking as I can't see the branch configuration on the repo)

Issue #2174, opened 2021-07-29
400 bad rquest when trying to delete relationship column


j2l:
**Describe the bug**
I wanted a "one user -> many books" relationship, it's not present in books table but in Users.
So I left the "many to many" relationship column and created a second relationship from users to books.
Then I wanted to delete the first relationship

Hitting Delete for "Checked-out-by" I got errors:
```
POSThttp://0.0.0.0:10001/api/tables
400 Bad request

{"_id":"ta_6ebefebfb0864bb5912ec78c1fda56ef","_rev":"13-fb72e274d658073099155b24d088d515","type":"internal","views":{},"name":"Books","schema":{"Auto ID":{"name":"Auto ID","type":"number","subtype":"autoID","icon":"ri-magic-line","autocolumn":true,"constraints":{"type":"number","presence":false,"numericality":{"greaterThanOrEqualTo":"","lessThanOrEqualTo":""}},"lastID":2},"Created By":{"name":"Created By","type":"link","tableId":"ta_users","fieldName":"Books-Created By","relationshipType":"many-to-many","autocolumn":true},"Created At":{"name":"Created At","type":"datetime","subtype":"createdAt","icon":"ri-magic-line","autocolumn":true,"constraints":{"type":"string","length":{},"presence":false,"datetime":{"latest":"","earliest":""}}},"Updated By":{"name":"Updated By","type":"link","tableId":"ta_users","fieldName":"Books-Updated By","relationshipType":"many-to-many","autocolumn":true},"Updated At":{"name":"Updated At","type":"datetime","subtype":"updatedAt","icon":"ri-magic-line","autocolumn":true,"constraints":{"type":"string","length":{},"presence":false,"datetime":{"latest":"","earliest":""}}},"Title":{"type":"string","constraints":{"type":"string","length":{"maximum":null},"presence":false},"fieldName":"Books","name":"Title"},"Author":{"type":"string","constraints":{"type":"string","length":{"maximum":null},"presence":false},"fieldName":"Books","name":"Author"},"Status":{"type":"options","constraints":{"type":"string","presence":false,"inclusion":["Available","Non Available"]},"fieldName":"Books","name":"Status"},"Users":{"name":"Users","type":"link","tableId":"ta_users","fieldName":"checked out by","relationshipType":"one-to-many"}},"sourceId":"bb_internal","primaryDisplay":"Title","indexes":[]}
```
```
GEThttp://0.0.0.0:10001/api/undefined/rows
[...]

PR #2286, opened 2021-08-05
Some fixes for internal and external relationships

mike12345567:
## Description
A variety of fixes after doing some testing around internal and external relationships:
1. When creating an SQL relationship an error would flash in the modal before it closed, getting rid of this
2. Some external relationships wouldn't work if there were overlapping columns, as per example in #2167 
3. Fixing an issue where internal relationships couldn't be deleted if they were made against the user table #2174 
4. Fixing an issue where many to many relationships didn't get created occasionally

Some of these fixes were required due to some of the changes that had to be made for SQL relationships previously.

### DIFF
--- a/packages/server/src/integrations/mysql.ts
+++ b/packages/server/src/integrations/mysql.ts
@@ -242,7 +242,7 @@ module MySQLModule {
       const input = this._query(json, { disableReturning: true })
       let row
       // need to manage returning, a feature mySQL can't do
-      if (operation === "awdawd") {
+      if (operation === operation.DELETE) {
         row = this.getReturningRow(json)
       }
       const results = await internalQuery(this.client, input, false)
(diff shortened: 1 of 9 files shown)


shogunpurple (approved):
Great job on this. Some good fixes here!

packages/server/src/integrations/mysql.ts
```diff
@@ -242,7 +242,7 @@ module MySQLModule {
       const input = this._query(json, { disableReturning: true })
       let row
       // need to manage returning, a feature mySQL can't do
-      if (operation === "awdawd") {
```
shogunpurple, line 245: I chuckled at this lol

mike12345567: yeap that must have been missed at some point haha

PR #2295, opened 2021-08-06
scarf support

shogunpurple:
## Description
Integrating with scarf - I've updated the domain for our images to point to the scarf gateway

## Screenshots

### DIFF
--- a/hosting/docker-compose.yaml
+++ b/hosting/docker-compose.yaml
@@ -5,7 +5,7 @@ version: "3"
 services:
   app-service:
     restart: always
-    image: budibase/apps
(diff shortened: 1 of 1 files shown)


mike12345567:
This is class, love it! LGTM!

aptkingston (approved):
Awesome! Is scarf a straight up proxy or does it cache images itself? Just thinking if it would add any latency to releases.

shogunpurple:
It has a gateway that it runs in all different AWS regions and it just does a simple redirect to the docker registry, so I don't think it's too much of an issue in terms of latency. You can find more info about the scarf architecture here:

https://docs.scarf.sh/gateway.html

Issue #2283, opened 2021-08-05
Get Started Checklist does not recognise my first app


mjashanks:
I have created and published my first app, but still see this (even after page refresh):

![image](https://user-images.githubusercontent.com/3524181/128360568-71ed512e-ab88-4435-8fb7-f0a5a2c3c4cf.png)

mike12345567:
Fixed in latest release.

Issue #2282, opened 2021-08-05
Platform url defaults to port 1000 on fresh instal


mjashanks:
I created a fresh install of budibase, and ran from localhost:10000

However, my platform URL defaults to `http://localhost:1000` - I haven't tried directly, but I'm pretty sure this will cause issues somewhere!

mike12345567:
Fixed in latest release.

Issue #2195, opened 2021-07-30
New apps do not appear in user record role assignment


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/2194

1. Create a brand new app
2. Go to `Home > Users > User record`
3. Try to assign role your new app

The new app does not appear in the list.

If I restart my containers, the app appears.

mike12345567:
Fixed as of #2296 - was an issue with `allDbs` refreshing.

PR #2296, opened 2021-08-06
Fixing various issues caused by PouchDB allDbs call

mike12345567:
## Description
Fixes for a lot of issues generated by the use of the pouchdb allDb call, which is not designed for multi-client DB setups like ours, using CouchDB method instead.

The issues fixed by this work are namely:
1. #2283 - the checklist did not update correctly as different services were creating/deleting apps, meaning the services didn't know about other DBs.
2. #2195 - as with the previous issue, no service knew about what other services were doing, so databases would go missing.
3. #2282 - simple fix, just the wrong URL specified in the builder codebase.

### DIFF
--- a/packages/auth/src/db/utils.js
+++ b/packages/auth/src/db/utils.js
@@ -1,5 +1,8 @@
 const { newid } = require("../hashing")
 const Replication = require("./Replication")
+const env = require("../environment")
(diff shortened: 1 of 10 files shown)


shogunpurple (approved):
LGTM! nice one

PR #2317, opened 2021-08-09
Makes the NPS Feedback Form options and comments optional

kevmodrome:
## Description
Makes the NPS Feedback Form options and comments optional


shogunpurple (approved):

kevmodrome:
Unsure why this would fail - does it fail on other branches as well?

shogunpurple:
@kevmodrome seems like the develop branch is green. Might just be a flaky cypress test perhaps

kevmodrome:
Re-running to see if it passes 🙏

Issue #2139, opened 2021-07-27
auto columns not populated / deleted rows reappearing / and more


shogunpurple:
### Discussed in https://github.com/Budibase/budibase/discussions/2132

<div type='discussions-op-text'>

<sup>Originally posted by **projestra** July 27, 2021</sup>
Dear all,

when I came across budibase a couple of days ago I was excited by the whole concept. Brilliant - many thanks! 

I've installed budibase as instructed in die manuals via the CLI on an old iMac running Ubuntu 20.04.1. All components from node.js to docker where installed in their most current version. So far so good ... up and running smoothly.

Actually everything woks fine for as long as I work from a single user (which in my case is the budibase admin user). As soon as data rows are added from a second user account, database behaviour becomes very strange (to be on the safe side, I even ditched the whole install incl. containers and volumes, of course but the behaviour stays the same). 

As this is best illustrated with a series of screenshots, I have attached a PDF illustrating the system behaviour. I don't have the faintest idea what may be going on here and would greatly appreciate some insight.

Many thanks in advance and kind regards,
Christian
[The Problem.pdf](https://github.com/Budibase/budibase/files/6885942/The.Problem.pdf)
</div>

mslourens:
I investigated the issue. As far as I can see the Created By is empty, because the new user is not known in the database of the app. When you first create an app, then you create a new user, and finally you open the created app with the new user and add a new row with a 'Created By' field, the system will try to find the user to create a link between the user and the created row. It will not find this new user, because it only has a reference document to the user who originally created the app. 

In order to fix this issue we need to add a reference to the new user (`ro_ta_users_us_<ID>`) in the app this user is working on. We could do this when creating a new user for all apps, or we could check if the user is known by the app when the app is opened, and if not, create the reference to the user table.

Mitch-Budibase:
I have followed the steps within the produced problem.pdf (linked above).

I am seeing an issue still.

Table created and populated by one user - Created By & Updated By columns populated accordingly.
I sign in as another user(admin user) and add a row to the table - Created By & Updated By columns are **NOT** populated correctly.
I sign in as another user (basic user) and add a row to the table - Created By & Updated By columns are **NOT** populated correctly.

![Screenshot 2021-08-18 at 15.46.34.png](https://images.zenhubusercontent.com/60fed9d54a973614368dfdfe/abc22c57-f499-4003-b512-b6228cada44e)

Issue #2108, opened 2021-07-25
Wrong sorting order by Date/Time type column 


sergey-dme:
Hey all,

Sorting order works incorrectly with Date/Time type columns, across whole platform.

Few examples ("Updated" is Date/Time type filed I created):

<img width="270" alt="Screenshot 2021-07-26 at 02 00 07" src="https://user-images.githubusercontent.com/53024832/126916160-b10a0b08-3adc-438a-a091-9df646f9fa69.png">

<img width="236" alt="Screenshot 2021-07-26 at 02 00 12" src="https://user-images.githubusercontent.com/53024832/126916166-70bd7188-d589-48a3-b278-7cccc2509006.png">

<img width="259" alt="Screenshot 2021-07-26 at 02 00 44" src="https://user-images.githubusercontent.com/53024832/126916151-11266756-0be9-42aa-9797-85c622dab710.png">

<img width="186" alt="Screenshot 2021-07-26 at 10 38 28" src="https://user-images.githubusercontent.com/53024832/126951164-e7c4993c-3a15-4696-a830-95ee470462b8.png">

<img width="182" alt="Screenshot 2021-07-26 at 10 38 35" src="https://user-images.githubusercontent.com/53024832/126951171-00520140-77cd-4770-af3c-0844a2b266d1.png">

 Current Version: 0.9.79

sergey-dme:
Maybe it will help to investigate: my rows containing "00:00" were created using Date Picker. Others - using` {{now}}`

sergey-dme:
Hey all,

I've just checked using Repeater unformatted Date/Time values:

Values saved using` {{now}}` stored as:
`2021-07-22T19:13:14.244Z`

Values from Data Picker:
`Fri Feb 01 2019 00:00:00 _GMT+XXX (Time Zone Name)_`

That's why sorting order does not work..

Could you please fix it?

Thank you.

sergey-dme:
Hi @shogunpurple @aptkingston please take a look at the following example of Flatpicker value behavior:

1. Create a internal table "**Table**"

2. Create a two columns: **DateA**, **DateB** - both Date/Time type filed

3. Setup the form for adding new row, keeping only **DateA** filed in the form

![DatteForm](https://user-images.githubusercontent.com/53024832/130256471-2eb1479a-0ba9-4ac3-a86f-b9f59ab187b6.PNG)

[...]

Issue #2020, opened 2021-07-19
Created By and Modified By Columns are not populated.


medouederni:
Using Budibase  0.9.73 Docker deployment on Windows
I added a new internal table (Files) with all auto columns and the related screens.
When I add a new row, the **Created By** column is empty.
I expected that the **Created By** column will be set with the current user.

Am I missing something?

_Originally posted by @medouederni in https://github.com/Budibase/budibase/issues/1022#issuecomment-878185141_

mjashanks:
Can confirm that this is an issue - whether creating data via the builder or via the app. Have tried a few different users.

dashrandom:
@shogunpurple for some reason I'm still experiencing this issue despite the merge. Is it because it hasn't been pushed to a release yet?

shogunpurple:
@dashrandom are you using an existing app? The fix above will not backfill old entries.

If this occurs when you are creating a new row, is there any errors in the browser console? 

What Budibase version are you on?

shogunpurple:
Oh also @dashrandom did you upload your data with a CSV? We noticed an issue with this and there’s a PR up to fix it:

https://github.com/Budibase/budibase/pull/2451

dashrandom:
@shogunpurple I'm using an existing app, but updated it to version 0.9.114 of Budibase. I'm not looking to backfill old entries, I tested with new entries but it doesn't work.

Nope, not uploading with CSV either. I created actual entries both from backend and via the app and both ways the created by and updated by do not populate. Let me know if I can provide some logs or further details.

dashrandom:
![image](https://user-images.githubusercontent.com/9287237/130762503-0cee9b43-cccb-4092-ae89-0c06485df0e7.png)

Image for reference

shogunpurple:
@dashrandom thanks - does the issue still occur in a new app? I'm having trouble recreating it, just trying to identify how it's happening

dashrandom:
@shogunpurple I just tested a new app and I confirm it doesn't happen with the new app! Does this mean that the upgrade process for existing apps does not move the merge fixes?

I'm going to try exporting the bugged app and reimporting to see if it fixes it and report back.

dashrandom:
Exporting and importing the app with the issue, the same issue is carried over. I've provided the export here for you to verify on your instance.
[export (1).txt](https://github.com/Budibase/budibase/files/7046487/export.1.txt)

dashrandom:
[...]

Issue #1922, opened 2021-07-05
Can't assign other users in my app


shogunpurple:
### Discussed in https://github.com/Budibase/budibase/discussions/1706

<div type='discussions-op-text'>

<sup>Originally posted by **hshifaz** June 12, 2021</sup>
Need help in figuring this out!

Lets consider 2 Internal tables, Users (default) and Projects. In this App, each project will have a Manager and it is a Relationship field linked to Users table.

![cant assign new users 3](https://user-images.githubusercontent.com/14042347/121782574-05870b00-cb5f-11eb-9fd1-5ebc7bdd5424.PNG)

In the Data table Project, when creating or updating a record, I can select the default user (user which was created as part of the installation of Budibase). However, other users I've created can't be assigned. The message shows "Row Saved Successfully", but as you can see on the screenshot, it is not.

![cant assign new users 1](https://user-images.githubusercontent.com/14042347/121782590-1cc5f880-cb5f-11eb-9093-42fe1d990d05.PNG)
![cant assign new users 2](https://user-images.githubusercontent.com/14042347/121782601-26e7f700-cb5f-11eb-96d8-6899fce7b8d8.png)

_**I've blur or cut out some personal info.**_

I've checked the following
1- Whether User is active
2- Gave Admin access also, changed to Public to test.

Could anyone help me figure this out, please.

Thank you in Advance.</div>

joebudi:
FYI, I've run into this issue twice now. It's pretty essential for assigning tasks/tickets to users. It's a blocker for me now.

PR #2323, opened 2021-08-10
ensure user metadata is created before link documents are created

shogunpurple:
## Description
Ongoing fixes branch:

Addresses: 
- https://github.com/Budibase/budibase/issues/2020
- https://github.com/Budibase/budibase/issues/1922
- https://github.com/Budibase/budibase/issues/2139
- https://github.com/Budibase/budibase/issues/2108

## Screenshots

### DIFF
--- a/packages/bbui/src/Form/DatePicker.svelte
+++ b/packages/bbui/src/Form/DatePicker.svelte
@@ -13,9 +13,9 @@
 
   const dispatch = createEventDispatcher()
   const onChange = e => {
-    value = e.detail
-    console.log(value.toISOString())
-    dispatch("change", e.detail)
+    const isoString = e.detail.toISOString()
+    value = isoString
+    dispatch("change", isoString)
   }
 </script>
 
--- a/packages/bbui/src/Form/DatePicker.svelte
+++ b/packages/bbui/src/Form/DatePicker.svelte
@@ -14,6 +14,7 @@
   const dispatch = createEventDispatcher()
   const onChange = e => {
     value = e.detail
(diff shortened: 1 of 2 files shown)


packages/bbui/src/Form/DatePicker.svelte
```diff
@@ -13,8 +13,9 @@
 
   const dispatch = createEventDispatcher()
   const onChange = e => {
-    value = e.detail
-    dispatch("change", e.detail)
+    const isoString = e.detail.toISOString()
```
aptkingston, line 16: Was this the cause of the bad date sorting issue - that it was storing actual JS objects sometimes and strings other times? I know Flatpickr emits JS date objects (and I hate Flatpickr) but there could be consequences of changing this somewhere that might have been written to expect the date objects, so it just might need checked well.

shogunpurple: Yep, there was a mismatch between the way dates are stored in the system, between handlebars representation (which is ISO) and the way flatpickr stores it, which is just `new Date().toString()` from the looks of it. Updating this still seems to work with flatpickr and store the right format in the DB.

kevmodrome:
Can this be merged? Would love to get #1922 fixed so I can make the tutorials look a bit better 👍

aptkingston (approved):

packages/bbui/src/Form/DatePicker.svelte
aptkingston: Cool, all good then.

PR #2339, opened 2021-08-11
Custom form validation

### LINKED ISSUE #2281: Fleshing out the constraint functionality in the data section
I ran into a situation where I want to restrict a user from selecting a date that is in the past. Currently that isn't possible. One solution is to allow the user to write HBS in the constraints. Another would be the ability to write full-on JS.

In addition to the simple use-case of a date there are a ton of other situations where you might be able to use it. For example, this might not make much sense but the functionality does I think: let's say you have a table for a CRM Opportunities. The table has `Name`, `Stage` and `Attachments`. Let's say we have an opportunity that is at the `Unqualified lead` stage, and we only want the user to be able to add an attachment when the stage is `Negotiating`.

@aptkingston pointed out that this is very much linked to #2096. When the design section gets this functionality we wouldn't need this.

@mike12345567 on the other hand pointed out that we might need this for backend validation for the backend API that is in the pipeline.

Could the constraints fields be the place where we handle validations for the data/backend? Could the design section inherit these validations somehow (less work for the user)?

aptkingston:
## Description
This PR adds the ability to add custom validation rules to form fields. Validation on form fields is built of both the table-level schema constraints and any additional custom rules. Validation rules can use normal values, HBS expressions and values from context. This means you can reference other fields in your rules.

image

**Edit**: I've added a duplicate button and resized some inputs to maximize space efficiency and be more in line with the other drawers:
image

Some example rules:
`billing country must be the same as shipping country` (using `must equal` constraint and referencing a field from context)
`delivery date cannot be in the past` (using `min value` constraint and a HBS expression for `{{ now }}`)
`phone number cannot contain any letters` (using `must not match regex` constraint with something like `[a-zA-Z]+`)

Combining with conditional UI is even more powerful, allowing things like:
`if age > 18 then ticket type must not equal child`

As usual, the validator options depend on the type of field you are validating. Strings allow regex while numbers and dates allow things like min and max values. Attachments can only be required or not. Validation even works on relationships, so you can enforce validation on the number of relationships, as well as forcing inclusion or exlcusion of a specific record.

## Configuring validation rules
Validation rules are made of 4 parts - the constraint, the value type (value or binding), the value, and an error message.

The constraint options are different for each field type and are selected from a list. When we add JS, we can add another option here for custom JS validation. Here are the options for strings for example:
image

And here are the options for relationships, as another example:
image

The value type option allows you to use the native component for the field type to enter a value, such as a datepicker for dates, if you choose the `Value` option. This is the same strategy used elsewhere to support both bindings and native components.
image

The value option is unique to each constraint, and all constraints except for `Required` need a value parameter to operate.

The error message is the error that will be shown if validation fails.
image

Closes #2281.


shogunpurple (approved):
LGTM! Nice work

PR #2353, opened 2021-08-13
Bindings for device type

aptkingston:
## Description
This is a quick feature to add global data bindings for device type. Combining this with conditional UI finally lets users change the settings of their components depending on screen resolution, so should help massively with mobile support. This is related to #2263.

2 new bindings have been added - `Device.Mobile` and `Device.Tablet`. After discussion with Joe, the breakpoints are as follows:
Mobile: 0-767px
Tablet: 768-1279px
Desktop: 1280px+

Everything is desktop by default, so there is no binding for desktop. You design apps for desktop, and can add mobile or tablet compatibility rules as required.

Here are the new bindable properties:
image

They evaluate to true or false if you bind to them:
image

Here's an example of using conditional UI to change a setting to make a button larger on mobile:
image

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -17,7 +17,13 @@ export const getBindableProperties = (asset, componentId) => {
   const contextBindings = getContextBindings(asset, componentId)
   const userBindings = getUserBindings()
   const urlBindings = getUrlBindings(asset)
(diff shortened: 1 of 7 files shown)


joebudi:
A very welcomed addition for a lot of our users. Super work!

shogunpurple:
@aptkingston this is awesome. Implementation with the ResizeObserver is 👌 too. Nice work

shogunpurple (approved):

PR #2354, opened 2021-08-13
Some tweaks for Joe

aptkingston:
## Description
This is a collection of tweaks which @joebudi requested:
- Add XS, XL, XXL and XXXL size options to parapraph and heading components
image
- Add static white text color (I added static black too). I also updated the first heading to "Theme" instead of "Grays" to make it more obvious that the gray palette is dynamic, but the colours are not. We can add more to the theme section when we implement custom theme colours.
image
- Add "over background" button variant
image
image
- Add 12px options for margin and padding
image

### DIFF
--- a/packages/bbui/src/ColorPicker/ColorPicker.svelte
+++ b/packages/bbui/src/ColorPicker/ColorPicker.svelte
@@ -19,7 +19,7 @@
   const dispatch = createEventDispatcher()
   const categories = [
     {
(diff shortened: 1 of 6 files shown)


shogunpurple (approved):
LGTM!

aptkingston:
How many times will the release bot generate commits that cause conflicts 🙄 second time pulling in develop already!

PR #2370, opened 2021-08-16
Return validation errors

### LINKED ISSUE #2355: Updating rows in Data section does not show correct validation messages
**Describe the bug**
Editing and adding rows results in a generic "Bad Request" error when something goes wrong. This should use the actual message to inform the user of what is happening. In the reproduction case below it should say something along the lines of `Column can't be blank`

**To Reproduce**
1. Add table and a column, make it required
2. Add row without anything in the column and try to save
3. See "Bad Request" error

**Expected behavior**
I expect the proper error messages to be shown, not a generic `Bad Request` one.

mslourens:
## Description
Fixes #2355 - return validation errors to client and display them

## Screenshots
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
@@ -31,7 +31,10 @@
         .flat()
       // Prevent modal closing if there were errors
       return false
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM! Thanks @mslourens

PR #2372, opened 2021-08-16
Autocomplete component

### LINKED ISSUE #2300: Searchable Pickers
As a user i want to be able to search within a pickers form field in order to select the option fast and avoid long scrolling.
- Options Picker
- Relatioinship Picker

**Screenshots**
searchable_pickers

mslourens:
## Description
Fixes #2300 
- Adds an autocomplete component and makes the optionsField and relationshipsField optionally searchable 
- Adds a new optionsType to the optionsField, called 'autocomplete'
- Adds a new boolean option to the relationshipsField, called 'autocomplete'

## Discussion
I am not quite sure if we should create a complete new component (which is a duplicate of the CoreSelect component, with added properties) or re-use the CoreSelect component. I choose the former, because then we can customize it more, if needed. 

### Optionsfield optionsType
image

### RelationshipsField option
image

### Autocomplete component
image


aptkingston:
Hey @mslourens - thanks for this! Regarding the discussion, I actually think it would be much better to make `autocomplete` a prop on the core `Select` and `Multiselect` components. Because they both share the same `Picker` underneath, if we just pass on the `autocomplete` prop down to the picker then we can have the full implementation there. Having autocomplete on the multiselect would be really useful as well. I understand why you went with the duplicated component, but I don't think it will need any more customisation. 

To keep the props interface even more simple, I don't even think we need to worry about the autocomplete placeholder - just something very generic like `Search` would be perfect.

We could also make autocomplete a checkbox setting for the options field then as well, similar to the relationship field.

mslourens:
I will revert my changes to the autocomplete component and move them to the CoreSelect and CoreMultiselect. But I'm not sure if you want to change the OptionsField autocomplete option into a checkbox, because optionType radio button and autocomplete is not possible, it's either autocomplete or radio buttons. Therefore I added the autocomplete option to the optionsType select. Do you agree to keep it there?

aptkingston:
You could actually use a new feature that was recently merged into develop. In `manifest.json` we have the option of `dependsOn` to conditionally show a setting only if another setting has been chosen. We recently improved this feature so that you can conditionally show a setting only if another setting equals a particular value - so in this case we could only show the autocomplete checkbox if the type is "select".

The syntax for the autocomplete setting would be something like:
```
"key": "autocomplete",
"type": "boolean",
...
"dependsOn": {
  "setting": "optionsType",
  "value": "select"
}
```
This would mean that we can reduce the options type back down to only 2 options, and only show the checkbox if type "select" is picked - which keeps it consistent with the relationship picker.

This is a brand new feature though so I totally understand what you mean, and without this feature I would probably agree with you! You'll need to pull down the latest `develop` to be able to use the `dependsOn` property like this.

mslourens:
Thanks for your suggestion. I implemented the change, it seems to work well.

aptkingston:
Awesome. I'll quickly check this out and test it and then approve 👍

aptkingston (changes requested):
Looks great! Couple of points which I think need changed, and then it's good to go. I'd also like to improve the look of the search bar, so if you could stick this CSS at the bottom of the core Picker file as well then that would be super!

```css
.spectrum-Popover :global(.spectrum-Search) {
  margin-top: -1px;
  margin-left: -1px;
  width: calc(100% + 2px);
}
.spectrum-Popover :global(.spectrum-Search input) {
  height: auto;
  border-bottom-left-radius: 0;
  border-bottom-right-radius: 0;
  padding-top: var(--spectrum-global-dimension-size-100);
  padding-bottom: var(--spectrum-global-dimension-size-100);
}
.spectrum-Popover :global(.spectrum-Search .spectrum-ClearButton) {
  right: 1px;
  top: 2px;
}
.spectrum-Popover :global(.spectrum-Search .spectrum-Textfield-icon) {
  top: 9px;
}
```

That CSS affects the search bar and removes the double border, straightens the divider line and increases the height so that it looks a bit better when embedded inside the popover.

image

aptkingston (approved):
LGTM! Great job, thanks @mslourens.

PR #2441, opened 2021-08-23
Multi-step forms

### LINKED ISSUE #2434: Detail screen pulls all rows in SQL table, in builder preview
When i use a SQL (postgres) table to create a detail screen, the Data Provider pulls all rows in the table. This only happens in the builder preview, not when running the app.

image

Filter is set correctly:

image

I can set `Limit = 1` to solve this. However, since it's not set by default, it will cause someone pain if they have a lot of rows in their table.

aptkingston:
## Description
This PR adds the concept of multi-step forms to client apps. Forms can have multiple steps which can be independently validated and navigated through, building up the full form payload.

Design doc here: https://docs.google.com/document/d/1iyBqQPewCk6IBN-i7dSRSklf5B9si3NJxPGOc5REGDc/edit

### Adding form steps
Form steps are a new component, which lives under the "Form" category, and can only be placed inside forms.
image

Form steps have a single setting - the step number. This is a number that represents the position of the step in the form, starting from 1. Steps are automatically named and configured with a default step number depending on the existing steps - basically they will automatically increment for you by default when you add them, but you can always rename or change the number if you want to.

Automatic naming:
image

### Using form steps with other content
When building a form, you can combine form steps with other form components - such as if you wanted to always show a heading, buttons or even some other form fields regardless of what step you're on.

Here's an example of a form with 2 steps, and the full component structure required. This form has a header, footer and buttons at the bottom which are always displayed. The form steps only contain the fields required for each step, and an additional text field for a question title.
image
image

Notice the use of conditional UI to update the buttons when the form is on step 2.

### Navigating between form steps
A new button action has been created to change the form step. This step takes 2 setting - the form component you're targeting, and the step to go to. There are a few built in helpers to make it really easy to navigate without having to duplicate buttons on each step. There are helpers for "next step", "pervious step" and "first step". The other option is navigating to a specific step, which takes an additional setting of which step number you'd like to go to.

image
image

### Validating form steps
Form steps can be validated independently, allowing you to validate your form as you progress through it. The `Validate Form` button action has been expanded with a new checkbox, for whether you just want to validate the current step or not. Leaving this unchecked will validate the entire form, while checking it will only validate the current step you're on.

image

### Form context extensions
The `Form` component has been extended with a couple of extra properties to make working with steps and conditional UI really easy. The new fields are:
- `Valid` (bool): whether the form is valid or not
- `Current Step` (number): the current step that the form is on (or 1 if there are no steps)
- `Current Step Valid` (bool): whether the current step is valid (only considers fields in the current step)

Combining these new fields with conditional UI make it simple to create dynamic forms that share the same base components, reducing boilerplate.

### Editing form steps in the builder
Logic has been added to the client library so what whatever form step you have selected (or if you have a component inside a form step selected) then that form step will always be the one displayed inside the builder preview. This feels really natural, and it means you just need to select whatever component you want to edit and it will be shown. You don't need to manually adjust any settings to say which form step you would like to see inside the builder.

### Behaviour clarification
- Form steps are just empty containers, so you can add whatever components you want inside
- Form steps are optional, you do not need to use them and forms will work as they are today without any
- Form steps do not need to be "top level" inside the form - you can have headers at the top of your form and common buttons at the bottom, and fire the form steps in the middle
- Form steps are automatically named and numbered for convenience, but you can change the default settings
- Validation can be done on each step
- You can use the same step number for multiple steps, and every matching step will be shown unless you hide some with conditional UI. This is one way to configure optional steps depending on previous values.
- You can go to any step number you want, but if you don't have a matching form step component then nothing will be displayed

## Other features
- Added a spectrum stepper component for numeric inputs
- Fix bug where preview would crash if it encountered an incorrect node type while applying borders around the selected component
- Entirely rewrote client form library to handle steps and derive as much state as possible rather than duplicate it
- Fix form validation ending early when encountering an error. All fields will now be validated
- Default picker sorting of options to off, and only default it to on for client apps. Any pickers which should be sorted in the builder can do so by adding the `sort` prop
- Add placeholder to detail autoscreens for when no rows exist
- Ensure limit setting on dataproviders is set to 1 for external tables. Fixes #2434
- Always show a component if it is selected, even if conditional UI is hiding it. Otherwise it's impossible to edit a component unless you temporarily disable conditional UI


shogunpurple (approved):
Awesome work cheeks - this is such a killer feature 🚀

PR #2442, opened 2021-08-24
Replace 'replaceAll' with 'replace' and a regular expression

mslourens:
## Description
Older browser don't support replaceAll yet, so it's better to use a regular expression to replace all occurrences in a string, to support the older browsers.
So, instead of `string.replaceAll(" ", "-")` we use `string.replace(/ +/g, "-")`


shogunpurple (approved):
Thanks @mslourens !

PR #2443, opened 2021-08-24
Add resizable dividers in design tab

### LINKED ISSUE #2159: Expandable sidebar
### Discussed in https://github.com/Budibase/budibase/discussions/2158

<sup>Originally posted by **GuilhermeHirsch** July 28, 2021</sup>
Hello! It would be nice if it was possible to expand or if it had a side scroll on the left bar. This affects my experience when I have a lot of components.

Sem título
</div>

mslourens:
## Description
Fixes #2159 
Add the ability to resize the side panels in the design tab. The position of the dividers is not (yet) (re)stored. The component used comes from the Svelte [REPL project](https://github.com/sveltejs/svelte-repl/blob/d6a99e702d42eb18142dccbd6b766b56b69245e4/src/SplitPane.svelte)
When hovering over the divider, the cursor changes and the border becomes blue (to indicate that it can be changed, see screenshot). 

## Screenshots
image


aptkingston:
Hey @mslourens. Am I right in thinking that this absolutely positions the side panels? I can see in the screenshot that the app preview is not centered any more, making me thinking that you could extend the side bar and it would go on top of the preview rather than shrink it, is that correct?

I'd be keen to see what happens on smaller screen sizes too, where we would normally shrink down the app preview. Could you shrink down the window to something like 1200px and check what the behaviour is like?

Thanks.

mslourens:
There are two (nested) SplitPanes involved here. The parent pane is divided in two panes: the left of the outer SplitPane is 85% of the width and contains the second SplitPane. The left of the inner SplitPane is 25% and contains the Screens panel. The right of the inner SplitPane is 75% and contains the Preview panel. 
The right of the outer SplitPane is 15% and contains the Settings Panel. 

A picture speaks more than a thousand words, so here is what I mean:
image

This is how it looks like on a small screen:
image

The user is able to adjust the width of all three panels, so is it really that import that the initial position of the dividers is such that the preview panel will be exactly in the middle? I could tweak the values a bit so it looks better, if you want.

PClmnt:
If you pull up a binding panel, i.e, the `Configure Conditions` in the right panel, how does it look after the dividers have been re-sized? Does the panel re-size correctly?

mslourens:
the binding panel will in the center of the page, independant of the position of the dividers:
image

mslourens:
Hi @aptkingston @PClmnt what is the status of this PR? Do I still need to change anything?

shogunpurple:
We've discussed this internally and decided that we are going to take a new approach to this - with a possible redesign. I'm going to close this PR, as it won't fit in with our binding drawers and other UI elements. Thanks a lot for the contribution and for helping us realise what is possible with this!

PR #2446, opened 2021-08-24
Filter users table out of DataSource component and NewScreenModal

### LINKED ISSUE #2018: Prevent user row creation from the UI
**Describe the bug**

Hello Budibaser!

1. I created a new row [user table] from user UI, it's work as picture below, there is total 3 rows then.

image

2. but data UI does show the new row, as it's still 2 rows there.

image

mslourens:
## Description
Fixes #2018 - filter out Users table from NewScreenModal template selector and DataSourceSelect.

### Discussion
I'm not sure if this is the desired solution for preventing creating users from the UI, because now it's not possible to bind the Users table to anything, not even for viewing only. As fas as I know the DatasourceSelect is not aware if a form is selected or a dataprovider, so it cannot filter out the Users table conditionally. Is that a correct assumption?

### NewScreenModal
image

### DataSourceSelect
image

### DIFF
--- a/packages/builder/src/components/design/NavigationPanel/NewScreenModal.svelte
+++ b/packages/builder/src/components/design/NavigationPanel/NewScreenModal.svelte
@@ -1,7 +1,6 @@
 <script>
   import { store, allScreens, selectedAccessRole } from "builderStore"
-  import { tables } from "stores/backend"
-  import { roles } from "stores/backend"
+  import { tables, roles } from "stores/backend"
   import { Input, Select, ModalContent, Toggle } from "@budibase/bbui"
   import getTemplates from "builderStore/store/screenTemplates"
   import analytics from "analytics"
@@ -16,7 +15,7 @@
   let createLink = true
   let roleId = $selectedAccessRole || "BASIC"
 
-  $: templates = getTemplates($store, $tables.list)
+  $: templates = getTemplates($store, tables.getDataSources())
   $: route = !route && $allScreens.length === 0 ? "*" : route
   $: {
     if (templates && templateIndex === undefined) {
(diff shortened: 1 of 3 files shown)


aptkingston (approved):

shogunpurple (approved):
Yep, on board with this too. Thanks for this @mslourens - it's tripped quite a few people up

kevmodrome:
Sorry just saw this.

I don't think this is the correct way of solving this, we'll lose a bunch of functionality by just hiding the Users table. For example, imagine a situation where you have some metadata attached to your users and you want to show a list of all the users with said data, could be anything from number of sales made, last time they logged in, etc. 

If we remove the users table we should expose it in some other way IMO.

Removing it from the create new screen modal does make sense to me though.

aptkingston:
> 
> 
> Sorry just saw this.
> 
> I don't think this is the correct way of solving this, we'll lose a bunch of functionality by just hiding the Users table. For example, imagine a situation where you have some metadata attached to your users and you want to show a list of all the users with said data, could be anything from number of sales made, last time they logged in, etc.
> 
> If we hide the users table we should expose it in some other way IMO.
> 
> Removing it from the create new screen modal does make sense to me though.

Yea I get where you're coming from. The one (common?) case I can see that we lose is the idea of an admin user being able to see info about other users. Like an admin being able to check Bob's sales - because this would require a data provider to target the `Users` table with a filter by ID. Bob's ID would already be found because it exists in relationships of other app data. So maybe there is a case to be able to select it as a data provider source?

I hadn't thought of this Kev, nice catch.

shogunpurple:
We could add the metadata only as a datasource, maybe. This would allow you to show relationships on non-user table fields. The only problem with that is that you lose their email address (and name) if you don't assign custom metadata fields to the user to identify them

PR #2470, opened 2021-08-25
Multi-Option Data Type

PClmnt:
## Description
This PR adds a new data type, presented to the user as a `List` when selecting it, but represented internally by an array. 

This touches a few sections within the codebase. 

- Data screen
  -  Creating a column now has an option to select `List` as the type and provide it some values
  -  Creating a row will expose a multiselect picker composed of the column values
  -  The array is exposed in the table by way of badges, limited to 5, in the same style as a relationship field

- Design screen
  -  Adds a new form option called `MultiField Picker` (n.b I this sounds terrible, what should we actually call it? Open to suggestions)
  -  New constraints added to the validation drawer that allow the user to check if the picker `contains one of` a number of user selected value, or `does not contain one of` 
  -  Filter in a data provider to a value of the list

- Indexes
  - Couch does not allow indexing of arrays, so in order to support filtering, each element in the stored array needs to be indexed separately which is handled here: packages/server/src/db/views/staticViews.js#105
 
 ## Screenshots
Screenshot 2021-08-25 at 18 35 20

Screenshot 2021-08-25 at 18 35 31

Screenshot 2021-08-25 at 18 54 27

Screenshot 2021-08-25 at 18 57 02

Screenshot 2021-08-25 at 18 57 21


shogunpurple (commented):
Just a few questions - I think the equality thing might be a little bug. Apart from that the logic looks good. Good job!

aptkingston (commented):
Super job @PClmnt! This was down as a 3 point task but was more like 13 haha.
Few questions and suggestions made in the review comments.

The other thing I'd like to discuss is the addition of the "contains one of" and "not contains one of" expressions, and the changing of the existing "contains" and "not contains" expressions. Those existing expressions are generic ones to check if an array contains a value, and in my mind were already perfect candidates to use with this new type. While I understand that "contains one of" offers unique functionality, I don't think it's worth the trade off for the large increase in complexity in both form validation and filtering that's required to handle the array value of the expression. So personally I would like to keep things consistent with our other array type (relationships) and just stick to the existing "contains" and "not contains" expressions which already work with array types.

aptkingston (commented):
Few other questions, but we're getting there!

aptkingston (approved):
LGTM! Let's get this merged. Any other wee things should become clear soon from using it, but looks good now I think.

PR #2472, opened 2021-08-26
Revert "Filter users table out of DataSource component and NewScreenModal"

shogunpurple:
Reverts Budibase/budibase#2446

We are reverting this for now due to the comments made in the original PR. 

The way we solve this requires more thought, before removing potentially useful functionality for users.


kevmodrome (approved):
LGTM

PR #2502, opened 2021-08-31
App state

aptkingston:
## Description
This PR adds the ability to set arbitrary values in app "state". This is just a generic store of values that can be used to easily pass values around and persist them between navigation and even page reloading.

Any values in state are bindable from anywhere - component order **does not matter** for state. You can set a value from any button, and bind to it anywhere. You don't need to have your components nested like you need to do with normal bindings.

### Updating state
A button action has been created for updating state. There are 2 different types of update - setting a value and deleting a value.
image

Setting a value requires a name (what the value is called) and a value (what the actual value is). Deleting a value just requires the name.

The name field is a combobox and has options for all existing state names, so you don't need to type out the same names every time as that would risk introducing typos (and you'd have to remember them all). So you can pick from the dropdown instead, once you've used a certain name at least once.

image

### Persisting values
By default, values are set in a global store which will survive in-app navigation, but will be lost on page reload or a real browser navigation. When setting a value, you can choose to persist the value - meaning that it will be stored in the brower's local storage and survive page reloads and even closing the browser. This allows quite a few new features, like keeping filters persisted when going back from a details page.

Clarification of behaviour:
- If you set a value and persist it and then set the same name again but don't persist it, then the persisted value is cleared and only the non-persitent one remains
- Deleting a value deletes it regardless of whether the value was persisted or not

### Referencing state values
State values can be reference just like any other bindable property. The available state properties are automatically determined from the names of the values that you set - so if you set a value called `Name`, then `State.Name` will be an option in your bindings.

image

### Other stuff
- Updated the client side implementation of lucene searching to better match the real thing, by not matching on `equals` or `not equals` if using a nullish comparison value


shogunpurple (approved):
This is such a great feature - the persistence part is super powerful too, adding persistent state between refreshes on the client.

I would hold off on merging this until after the release as it's incredibly fundamental, changing the `ClientApp` file etc.

Great work 🚀

aptkingston:
> 
> 
> This is such a great feature - the persistence part is super powerful too, adding persistent state between refreshes on the client.
> 
> I would hold off on merging this until after the release as it's incredibly fundamental, changing the `ClientApp` file etc.
> 
> Great work 🚀

Yep for sure! Didn't mean to sneak this in before the release or anything - definitely hold off until after. And cheers 🚀

PR #2525, opened 2021-09-02
Fix indexing of rows with attachments

### LINKED ISSUE #2523: Adding an attachment to a row makes the row disappear
**Describe the bug**
If you have an attachment that is added to a row it will not be indexed and not show up in your application.

The issue has been identified and a fix is being worked on. Adding this bug here anyway in case someone is running into this and is looking through the issues.

**To Reproduce**
1. Create Table
2. Add attachment column to table
3. Create row and add something to the attachment
4. Add a data provider pointing to the table
5. No row will show up

**Environment**
Budibase Version: 0.9.117

aptkingston:
## Description
This fixes an issue where having an attachment on a row means the row will not be indexed correctly by lucene, and therefore never be pulled back inside data providers. The problem was that we added the capability to index array values in order to implement the multi-select options type, but we did not check the type of the items inside the array. Attempting to index full objects is what causes it to break.

Fixes #2523.

PR'ing straight to master as this is a pretty significant bug which is currently live.

## Screenshots
Before (only the row without an attachment is returned)
image

After (all rows are returned, including one with an attachment)
image

### DIFF
--- a/packages/server/src/db/views/staticViews.js
+++ b/packages/server/src/db/views/staticViews.js
@@ -100,11 +100,11 @@ exports.createAllSearchIndex = async appId => {
           let idxKey = prev != null ? `${prev}.${key}` : key
           idxKey = idxKey.replace(/ /, "_")
           if (Array.isArray(input[key])) {
-            for (let val in input[key]) {
-              // eslint-disable-next-line no-undef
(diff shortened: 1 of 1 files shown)


kevmodrome (approved):
LGTM

Issue #2530, opened 2021-09-02
Unable to create Views with the letters 'stat' in the name


joebudi:
**Describe the bug**
When you create a View and name it 'stat', the view does not populate with data.

**To Reproduce**
1. add data to a table
2. Click add view
3. Name it stat

**Expected behavior**
View should work

![Screenshot 2021-09-02 at 16 49 12](https://user-images.githubusercontent.com/49767913/131878413-03f51ab4-bc70-46bc-aee4-e86b81bb919c.png)

mike12345567:
This issue has been resolved as of latest bug fixes PR.

Issue #2435, opened 2021-08-23
On detail screen, SQL Relationships don't save unless re-selected


mjashanks:
To Repro:
1. Create SQL datasource (ds plus)
2. Create tables with One-Many relationship, and add some data
3. Create a "detail" screen for the "Many" table. 
4. Run your app and open the detail screen
5. Change nothing, just press save

I got the following error:
```
{"message":"error: invalid input syntax for type integer: \"{\"primaryDisplay\":\"michael\",\"_id\":\"%5B9%5D\"}\"","status":400}
```

here is my request body for the save:

```
{"id":null,"name":"New Deal","customerId":9,"value":1000,"_id":"%5B2%5D","tableId":"datasource_plus_32a51e0dfccb41058080147904f898d8__deals","_rev":"rev","Customer":[{"primaryDisplay":"michael","_id":"%5B9%5D"}],"Customer_text":"michael"}
```
Note the `Customer` field - this is my relationship (the parent/"one"). It's `_id` is "%5B9%5D". 

I'm guessing that this is because
1. The foreign key is an integer
2. The supplied value is `%5B9%5D`

@AndrewJRamay reported this in https://github.com/Budibase/budibase/discussions/2396

AndrewJRamay:
Hi, @mjashanks any chance this issue is resolved in the next feature release I hear that is coming in the next few days?

mike12345567:
This issue has been resolved as of latest bug fixes PR.

kengeorge2:
Kennymiles Tech
0 seconds ago
Hello, thanks for the demo, can you kindly make a video on a master detail  form, lets say like an invoice with invoice items, or requisition with request items. Would be great.

Issue #2404, opened 2021-08-18
Users are not indexed until some metadata column has been added to the table


kevmodrome:
### Discussed in https://github.com/Budibase/budibase/discussions/2389
It's not possible to access the rows in the Users table using the Data Provider

<div type='discussions-op-text'>

<sup>Originally posted by **erakastrati** August 17, 2021</sup>
Hi guys,

Do you know a way to add the Users table(that is generated itself from Budibase) in the development section or it is possible just from the tables part. 
I want to use the Users as Employees of my application without creating a new table named Employees?

Thanks a lot,
Era.

</div>

aptkingston:
Can't seem to reproduce this - it works fine for me.
![image](https://user-images.githubusercontent.com/9075550/129866577-5cb29f13-308f-4ae2-9692-4d85d6dd5aed.png)

kevmodrome:
That is odd indeed! I'm seeing this:

<img width="1779" alt="CleanShot 2021-08-18 at 10 40 14@2x" src="https://user-images.githubusercontent.com/534488/129866827-cf0f5207-da58-4026-8ab2-158dda55cafe.png">

kevmodrome:
Is your application a new one? Maybe there's some issue with it working in old ones? 🤔

aptkingston:
Looks like the user rows aren't indexed until some app-level metadata is created. Try adding a column to the users table in that app, and then edit a row and fill in the new column. Every row you edit and add data to should appear in the table.

kevmodrome:
That's interesting. I'll update the title of the issue so it makes more sense.

mike12345567:
This issue has been resolved as of latest bug fixes PR.

Issue #2402, opened 2021-08-17
SQL Relationships do not allow for the editing of the table that holds the FK


AndrewJRamay:
Hi Guys after playing with this most of the day I believe there is a bug that is not allowing me to edit a row once it is created in the table that holds the FK. I have it documented here #2396

when I remove all of the relationships I then can use the table properly to add any SQL Relationships I can no longer edit a row in that table. I can save a new row but not edit one.

AndrewJRamay:
to get the row to update I have to go and reselect every relationship box again. then it will save 
![image](https://user-images.githubusercontent.com/80058238/129810021-a494276f-ee43-421c-8f11-70227f1dcb83.png)

mike12345567:
This issue has been resolved as of latest bug fixes PR.

PR #2531, opened 2021-09-02
Relationship bug fixes (SQL and internal) and a few other bits

mike12345567:
## Description
Fixing a few issues that have been found while I was away:

1. #2402 - some issues with SQL relationships, especially when setting from the many side, this went a bit deeper than the issue suggested, but should have resolved most of these issues.
2. #2435 - was related to the previous issue.
3. #2530 - Joe found this issue, quick fix for it.
4. #2404 - this issue was a bit more complicated than issue suggested, basically the search endpoint didn't handle users correctly (enriching with global user information) but this is now fixed.
5. Fixed an issue where automations wouldn't be ran correctly when an SQL row was written/updated/deleted, the trigger wasn't working as expected.
6. Upgraded to Node 14 as Node 12 is set to go into EOL next year and wanted to move to the real LTS, this caused some issues around circular dependencies, versions of a few packages etc - upgraded and fixed any issues encountered. Still works perfectly fine with Node 12 and have checked this, but now works properly with Node 14 as well.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -138,7 +138,7 @@
     "supertest": "^4.0.2",
     "ts-jest": "^27.0.3",
     "ts-node": "^10.0.0",
-    "typescript": "^4.3.4",
+    "typescript": "^4.3.5",
     "update-dotenv": "^1.1.1"
   },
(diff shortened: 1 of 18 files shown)


shogunpurple (approved):
Great job @mike12345567 - some great fixes in here.

Issue #2378, opened 2021-08-17
Relationship breaks when primary display column is the same name as the table


mjashanks:
I have internal tables setup like: `one Requests has many Notes`

The `Notes` table has a primary display column called `Notes`, which contains the text of the note.

On the Request Detail screen, I want to display the all notes that belong to that request. So, I setup the structure...

`Data Provider - Requests` > `Repeater` > `Data Provider - Repeater.Requests.Notes`.

The `enrich` call returns the following for the `Requests.Notes` field:

(see the inner `Notes` field? - it should be returning a string)

```
Notes: [
  {
    _id: "ro_ta_0d5d6fddce624bedbe45cbfee676976e_ed38f83774b44eafa198005b6d69b327",
    _rev: "3-34fe49ee8ff54baa083649d9c2150684",
    tableId: "ta_0d5d6fddce624bedbe45cbfee676976e",
    Auto ID: 3,
    Created At: "2021-08-13T21:59:26.641Z",
    Updated At: "2021-08-13T21:59:26.641Z",
    type: "row",
    Notes: [
      { },
      { },
      { },
      { }
    ],
    Help Desk Requests: [
      {
      _id: "ro_ta_d6ceeef7ef8844559cf1df6a23a56177_43efdb6aae81472b94c44ffe2ecfb397",
      _rev: "3-52c359070867aef3e6de555b93c80b3e",
      tableId: "ta_d6ceeef7ef8844559cf1df6a23a56177",
      Auto ID: 7,
      Created At: "2021-08-13T21:58:17.497Z",
      Updated At: "2021-08-17T08:47:00.634Z",
      type: "row",
      Completed : false,
      Attachment: [ ],
      Description : "Mike"
      }
    ]
  }
]
```

This is immediately fixed if I rename the `Notes` column to something else.

mike12345567:
This issue has been resolved on develop.

PR #2539, opened 2021-09-03
Fixing issue with relationships that was uncovered by having same primary display and table name

mike12345567:
## Description
Issue is described here: https://github.com/Budibase/budibase/issues/2378

This is a simple fix for this issue, when using the enrich call rows may not all be from the same table, for whatever reason this is uncovered by the issue.

### DIFF
--- a/packages/client/src/api/rows.js
+++ b/packages/client/src/api/rows.js
@@ -121,7 +121,7 @@ export const enrichRows = async (rows, tableId) => {
       rows.forEach(row => {
         for (let key of keys) {
           const type = schema[key].type
-          if (type === "link") {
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):

aptkingston (approved):
Nice one! Happy to merge this for now, but I'll try and sort the root cause at some stage soon as well. For reference, the root cause is that when fetching data from a relationship, the `tableId` param that's passed in is for the parent row's table, rather than the other side of the relationship. So it fetches the incorrect schema and then assumes that some fields exist. I can see why this would crash if both tables had the same column name, but they were of different types.

PR #2549, opened 2021-09-06
fix exporting internal tables

### LINKED ISSUE #2526: Exporting a table results in a file with a url as content
**Describe the bug**
When exporting a CSV it does not successfully export the rows in the table. Instead the content is a single line: `/api/views/export?view=ta_users&format=csv`. No error is shown from on the client or server.

**To Reproduce**
1. Export Users table
2. Open file

**Expected behavior**
To be able to export a proper CSV file when exporting a table.

 - OS: Macos Big Sur
 - Browser Chrome
 - Version Develop

mslourens:
## Description
Fixes #2526 - exporting internal tables. The issue was caused by downloadjs: when passing a filename argument to downloadjs, it expects a payload instead of getting the data from the passed url. 

### Discussion
The downside from this is that the filename is automatically set by downloadjs and cannot be passed to the download function (it uses the last part of the path). So every export will be called `export` (without an extension). I'm not sure if this is what we want.

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/ExportModal.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/ExportModal.svelte
@@ -18,12 +18,10 @@
   let exportFormat = FORMATS[0].key
 
   async function exportView() {
-    const filename = `export.${exportFormat}`
     download(
       `/api/views/export?view=${encodeURIComponent(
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
Thanks again @mslourens !

PR #2558, opened 2021-09-06
Fixing issue with backwards compatibility for multi-tenancy

mike12345567:
## Description
Quick fix for backwards compat, enabling multi-tenancy on a previously single tenancy environment.


shogunpurple (approved):
LGTM!

PR #2561, opened 2021-09-06
Custom theming

aptkingston:
## Description
This PR adds the ability to easily theme some parts of your Budibase apps. Themes apply to the whole app, but can still be overridden by normal style settings or by custom CSS. In future we'll expand this to be org-wide and to be able to share themes across apps, and we'll also add additional theming options.

### Configuring theming
Theme settings are available by clicking the new `Theme` button, where the theme dropdown used to be. Note - if you haven't yet updated your app then you'll continue to the the old theme dropdown until you do.

image
image

The available theme settings are:
- Theme: the base theme, the same option as we offered previously
- Button roundness: how round you would like all your buttons to be
- Primary colour: the primary accent colour of your app. This is used for "Action" buttons, form fields (icons, checkboxes, radios, borders, datepickers), icon colours etc.
- Primary colour (hover): the same as above, but when hovering instead.
- Navigation bar background colour: the background colour of your navigation bar. This also includes the drawer menu when on mobile.
- Navigation bar text colour: the text colour in your navigation bar. It's important this is customisable since the background colour is.

Theme settings are updated in real time, so you can actually preview the changes behind the modal. They're saved automatically (like component settings) so there's no save button - just close the modal when you're done.

All theme settings have defaults, which are displayed initially. You can clear a colour setting at any time using the `X` icon in the lower left of the colour picker popup. You can also reset everything to defaults using the `Reset` button in the theme editor modal.

### Examples
These might only seem like a few settings, but they're enough to add some personality to apps. Here are a few examples:
theme1
theme2
theme3
theme4

### Other stuff
- The `standard-components` package has been removed, instead merging the client app components into the client library (under `src/components/app`)
- The link in the navigation bar is now highlighted for the current page (Sales is active here)
image
- Links in the navigation bar now get highlighted and underlined when hovering over them (Sales is being hovered here)
image

I think I've accounted for the removal of standard-components everywhere, but it might still lead to some strange behaviour in dev environments until you kill the dev server and freshly install packages again.


joebudi:
This is a crowd pleaser 🙌. Great work

packages/client/manifest.json
```diff
@@ -2,7 +2,9 @@
   "features": {
     "spectrumThemes": true,
     "intelligentLoading": true,
-    "deviceAwareness": true
+    "deviceAwareness": true,
+    "state": true,
```
shogunpurple, line 6: Out of curiosity, how exactly does this work? We attach certain features to client lib versions so that we know if they are compatible or not when you upgrade an old app?

packages/server/src/api/controllers/static/index.js
```diff
@@ -109,36 +104,3 @@ exports.serveClientLibrary = async function (ctx) {
     root: join(NODE_MODULES_PATH, "@budibase", "client", "dist"),
   })
 }
-
-exports.serveComponentLibrary = async function (ctx) {
```
shogunpurple, line 113: Nice to get rid of this. Makes life way easier

shogunpurple (approved):
LGTM! This feature is incredible - I pulled it down and had a play around. Really cool 🚀

packages/client/manifest.json
aptkingston: Yep exactly! So I introduced this concept a while back to be able to know (in the builder) if the current app supports certain features. A lot of features involve a configuration part in the builder (which is always the latest version) and the corresponding implementation in the client lib (which is versioned). So these flags are used by the builder to know whether to show or hide stuff.
With this feature, the new flag is used to know whether or not to show the full theme editor (in updated apps) or just show the old theme dropdown (in apps that haven't been updated yet).
I also snuck a feature flag in for the recently merged state feature, because I realised that I hadn't put in a flag for that - so you could configure state bindings in the builder but they would never show up in the app.

PR #2562, opened 2021-09-07
Account portal auth changes 

Rory-Powell:


packages/worker/src/index.js
```diff
@@ -14,6 +15,14 @@ const redis = require("./utilities/redis")
 
 const app = new Koa()
 
+// TODO: Remove this when using envoy / nginx
+const accountPortalCors = {
+  origin: "http://localhost:3001",
+  credentials: true,
+}
+
+app.use(cors(accountPortalCors))
```
shogunpurple, line 24: Since this is just for dev, we should use the `isDev()` check from `environment.js` to avoid the middleware being attached in prod.

shogunpurple (approved):
LGTM - should just conditionally include the cors middleware as it won't be relevant in prod.

packages/worker/src/index.js
Rory-Powell: I think it will have to remain in some form, at least in the current implementation so that the account-portal web will be able to talk to budibase ui. 

I'll update this with environment specific values for now

PR #2574, opened 2021-09-09
Add new redis key for email verification & pre-hashed password option

Rory-Powell:
For in progress here
https://github.com/Budibase/account-portal/compare/feature/onboarding-backend...feature/onboarding-backend-email?expand=1


mike12345567:
LGTM! Like the idea, nice simple and secure solution!

Rory-Powell:
@mike12345567 yes that's it, the verifications key is used by the account portal here
https://github.com/Budibase/account-portal/blob/master/packages/server/src/utilities/redis.js

arguably you could just use a string constant there instead, but it's nice to keep all our key spaces together and use the API consistently

PR #2575, opened 2021-09-09
Issue with public apps and multi-tenancy

mike12345567:
## Description
This issue was discovered by @aptkingston and @joebudi in production environment.

Fixing issue with multi-tenancy and public apps, when the tenant isn't necessarily known, it can be found in the app ID, building a middleware to manage this.


aptkingston:
🙏 cheers for looking into it!

PR #2576, opened 2021-09-09
Responsive preview

aptkingston:
## Description
This PR adds the ability to preview apps in the builder for different device sizes - desktop, tablet and mobile.

The preview size is controlled using a new control positioned beside the theme control.
image

Desktop size (full screen):
image

Tablet preview (1024x768):
image

Mobile preview (390x844):
image

Resizing the actual window will always take priority, and will further reduce the size of the preview if required. For example:
image

This was not just as simple as resizing the iframe, as doing that means that previewing on mobile leaves no room for the settings bar, which we want to use more in future. Therefore the iframe is actually full screen, but the client app is in control of rendering the actual app in a certain sized frame, while rendering additional components like the settings bar in as much space as they require. Due to the client app being in control, availing of this feature requires an app update to use, but this is controlled with a feature flag so that the builder will not show options for preview size until the client version supports it.

Here's what the settings bar looks like on mobile, rendering itself outside of the actual preview and using the spare space correctly:
image

### Other things
- Added a slight border radius to the preview
- Improved how the component selection borders are applied so that they will clip sightly better to the viewable area
- Added a `meta` tag to scale the app to the full viewport on devices that require it. This may lead to issues down the line, and we'll probably need to account for the "safe" viewable area and apply padding to the client apps, but for now we will let the user be in control of this
- Fixed inconsistencies between when the app used mobile styles, and when the context value for device size would update

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -43,6 +43,7 @@ const INITIAL_FRONTEND_STATE = {
     deviceAwareness: false,
     state: false,
     customThemes: false,
+    devicePreview: false,
   },
   currentFrontEndType: "none",
   selectedScreenId: "",
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -56,6 +56,7 @@ const INITIAL_FRONTEND_STATE = {
   clientLibPath: "",
   theme: "",
   customTheme: {},
+  previewDevice: "desktop",
 }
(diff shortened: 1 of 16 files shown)


mike12345567:
This is an amazing feature! Its really nice to be able to just see what it will look like in the preview, makes it so easy to develop a responsive app!

shogunpurple (approved):
This is great - looks so slick as well with the buttons/icons. Great job!

Rory-Powell (approved):
Wow! amazing

Issue #2579, opened 2021-09-09
Google SSO Not working properly


mohadel1990:
**Google SSO Redirect URI is invalid**
When setting up Google Auth using BudiBase Builder settings. `Callback URL` is a read-only field and set to
`/api/global/auth/google/callback`
The resulting URL link 
```sh
http://localhost:10000/api/global/auth/google/callback
```

![chrome_nbFKmbFQCK](https://user-images.githubusercontent.com/64827602/132718550-0e333c50-d5fa-4a2f-92f5-a0befebee349.jpg)

Even that the request URL is actually
![chrome_RCZwgKnX9z](https://user-images.githubusercontent.com/64827602/132717582-4223b2ae-0249-44e9-9564-88781a2601d7.jpg)

Returns a page of "Not Found" after google redirects back to BudiBase

![image](https://user-images.githubusercontent.com/64827602/132718174-53663e9e-00d7-4455-824e-ec1cb936a592.png)

By adding /default/ before google in the address bar to the redirected link making it 
```sh
http://localhost:10000/api/global/auth/default/google/callback=4%2F0AX4XfWgxD-QmUmCen...etc
```
Seems to be redirecting properly, and the user can sign in

Trying to set the `Google OAuth Client IDs redirect URIs` to the working redirect link will result in
```sh
Error 400: redirect_uri_mismatch
```
![chrome_abNIyMu3B7](https://user-images.githubusercontent.com/64827602/132719292-c179e181-41a7-4e77-a48d-deba7aaf5df8.jpg)

KernOil:
I am having this same problem with Azur AD using OpenID Connect I keep getting the error 
![image](https://user-images.githubusercontent.com/80058238/132722744-c4ff4ceb-b029-4b8b-ac5a-4fe01c4d8d18.png)

no matter what I change my URL to

mike12345567:
Thanks for reporting this! I've merged a fix for this, if you update your Budibase installations you'll be able to use Google/OIDC SSO now.

KernOil:
@mike12345567 Thank you very much!

mohadel1990:
@mike12345567
Thanks for the quick fix
I can confirm it is working in the docker installation

When should we expect it to reflect in the helm installation? upgrading still yields 0.9.122
And the issue still persists there

shogunpurple:
Hey @mohadel1990 - you shouldn't need to update your helm chart at the moment, the easiest way is to just pull the latest images by running:

[...]

PR #2582, opened 2021-09-09
Fix SSO callback URL in single tenancy mode

mike12345567:
## Description
Fixing issue #2579. In single tenancy mode needs callback URL endpoints without tenant ID.


shogunpurple (approved):

Rory-Powell (commented):
👍

PR #2584, opened 2021-09-09
Add horizontal bar chart setting

aptkingston:
## Description
This PR adds a horizontal option for the bar chart. This was far tricker than it should have been.... we should really attempt to replace Apex Charts at some stage.

This was requested by @joebudi (and proxied on from a user request I believe).

image


joebudi:
Awesome @aptkingston . The quick turnaround is appreciated.

PR #2587, opened 2021-09-10
Fix checklist spinner and progress

### LINKED ISSUE #2583: Minor graphical issue - checklist progress has become a spinner rather than progress bar
**Describe the bug**
As title, picture can be seen below, although its more obvious when its animating, instead of showing progress it is simply going round and round now.

image

**To Reproduce**
1. Login
2. Go to main app screen
3. Observe it spinning

**Expected behavior**
Should show progress instead of spinning.

aptkingston:
## Description
This PR fixes a couple of issues with the getting started checklist that were introduced in #2556.

The first issue was that the ProgressCircle component uses a falsey check to see whether or not a value had been provided. This meant a progress of 0 was treated as "no value provided" and therefore cause it to spin. This issue was not introduced by the recent PR, but just highlighted it as the second bug caused a value of 0 to always be passed in.

The second issue was that the array `reduce` function that was used to calculate the progress of steps was incorrect, and was always returning 0. It was iterating over the `keys` of the JSON and expecting to see the values inside, which are not there. There's no need for a `reduce` there anyway so I've replaced it with a simple filter.

Closes #2583.

### DIFF
--- a/packages/bbui/src/ProgressCircle/ProgressCircle.svelte
+++ b/packages/bbui/src/ProgressCircle/ProgressCircle.svelte
@@ -17,8 +17,6 @@
   export let minValue = 0
   export let maxValue = 100
 
-  $: console.log(value)
-
   let subMask1Style
   let subMask2Style
   $: calculateSubMasks(value)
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM!

Issue #2436, opened 2021-08-23
Automations: UI Redesign


shogunpurple:
https://xd.adobe.com/view/19fce9a0-fe56-4986-81b5-dbc395646a5f-4caf/screen/bc4520ab-e2f2-46c0-b2fb-8dddda460964/

mike12345567:
Work completed in PR #2610.

Issue #2412, opened 2021-08-18
Created by not present when using stringified JSON body with webhooks


shogunpurple:
### Discussed in https://github.com/Budibase/budibase/discussions/2366

<div type='discussions-op-text'>

<sup>Originally posted by **burnoutberni** August 16, 2021</sup>
Whenever a new row is created in a Budibase table, I push the whole row to an API via the webhooks feature:

```
{
  "posting": {{ JSONstringify trigger.row }},
  …
}
```

Unfortunately the resulting JSON string does not contain the `Created By` field, but I also need this information. Is there any way to do this?</div>

mike12345567:
Fixed in recent PRs.

Issue #2410, opened 2021-08-18
Automations: Query Step


shogunpurple:
When I'm setting up automations, I want to have the ability to add a "query" step, which would allow me to utilise the data provider interface to query a set of records (internal, external or relationships) and pass these records along to other steps in my automation, so that I can query the data I need based on values in my automaton blocks.

**COS**
- Add "Query rows" block definition
- Provide an identical interface to the data provider in the Design UI, including filters

![Screenshot 2021-08-18 at 15.23.57.png](https://images.zenhubusercontent.com/5fd773ddf3d67d4ceebe466b/55dfc7ba-b02d-4faf-ba82-83e395bb72c9)

More context here: https://github.com/Budibase/budibase/discussions/2024

mike12345567:
Added in most recent PR.

Issue #2409, opened 2021-08-18
Automations: Show Data Flow


shogunpurple:
When testing automations on the platform, I want to be able to see the data flow from one block to the other so that I can identify the inputs/outputs of each block and debug with more ease.

**COS**
- Automations restructured to run block by block
- Register automation failures and where exactly they occurred
- Allow configuration of seed data so you can test with different inputs

mike12345567:
Feature work completed as part of @PClmnt PR and my own on automation redesign and rework.

PR #2598, opened 2021-09-13
Automation backend rework

mike12345567:
## Description
Just a note initially, this PR is being raised as it is functional and tested, but not complete. There are some issues which are not quite done in this, like #2410 - but everything needed for the automation re-design is already in this so raising now to get it merged ahead of @PClmnt work.

This feature solves the following issues:
1. #2409 - it provides a test mechanism which takes inputs from the UI and then generates a report on the various inputs and outputs of each stage. This links up with @PClmnt work on the new test UI for automations.
2. #2408 - this introduces the concept of "metadata" to the server, metadata can be stored against an entity (such as for storing automation test inputs). The system can also generate metadata internally, such as automation test history. endpoints are `/api/metadata/:type/:entityId/` on POST (save), DELETE and GET, with types of `automationTestInput` and `automationTestHistory`.
3. Backend for #2436 - this provides all the necessary components for the UI re-design.
4. Cron jobs in production - cron appeared to work correctly in development, but once the app was published it did not start the new repeat job when in production - this was fixed.
5. Automations can no longer run in development, the only way to check an automation is working is to use the test system as per the design requirements. This introduced a few complexities, such as automations that would loop (e.g. trigger themselves) and I didn't want there to be a discrepancy between what would happen in production and what happens in development, so an automation can be "flagged" to be in test, when this occurs the trigger mechanism will be active again. This however only happens for the length of the test API call for that automation, so it is very unlikely the user will be able to trigger it while this is occurring (unless they add a large delay).
6. #2412 - now when rows are pushed to the automation system during triggers they will contain the full enriched rows.
7. Removed the concept of live automations, setting this on an automation will no longer do anything.
8. Removed the concept of downloadable automation steps, this added confusion and complexity which we do not need as most of our automation steps will be built in functions, with only a small number of external options.
9. Introduces a Slack integration.
10. Removes the Sendgrid automation step as this was not really used and it is not something we want to support.

The issues which have not been resolved by this PR:
1. #2410 - a query step has not been added yet, this will come in a following PR.


packages/server/src/api/controllers/automation.js
```diff
@@ -229,35 +144,34 @@ exports.find = async function (ctx) {
...
 }
 
 exports.getTriggerList = async function (ctx) {
-  ctx.body = triggers.BUILTIN_DEFINITIONS
+  ctx.body = triggers.TRIGGER_DEFINITIONS
 }
 
 exports.getLogicList = async function (ctx) {
-  ctx.body = logic.BUILTIN_DEFINITIONS
+  ctx.body = logic.LOGIC_DEFINITIONS
```
PClmnt, line 167: Hey @mike12345567 I see here that Logic is still an object of its own, but in the designs it seemed like they're now actions (https://xd.adobe.com/view/e4f6d4aa-86f3-48ca-a92a-b6d3d70b5a75-b1a5/screen/341effa1-ad27-42a8-af01-e68f677282a0?fullscreen) and i think @joebudi might have mentioned this as well to me at some point.

mike12345567: Ah, do you want me to represent that in the backend as well? I had just left it as it was because the API returns them altogether anyway, didn't want to break the API endpoint for you!

PClmnt: I think that would maybe be cleaner, because I would have to stitch them into the actions object in the frontend anyways. But it's no big deal either way, if it's hassle in any way at all I can easily just bring them together myself.

packages/server/src/api/controllers/automation.js
```diff
@@ -34,119 +53,6 @@ function cleanAutomationInputs(automation) {
   return automation
 }
 
-/**
- * This function handles checking of any cron jobs need to be created or deleted for automations.
- * @param {string} appId The ID of the app in which we are checking for webhooks
- * @param {object|undefined} oldAuto The old automation object if updating/deleting
- * @param {object|undefined} newAuto The new automation object if creating/updating
- */
-async function checkForCronTriggers({ appId, oldAuto, newAuto }) {
```
shogunpurple, line 43: Nice refactor on this, this was a bit nasty before and didn't fit the flow of the other automation blocks.

packages/server/src/api/controllers/automation.js
```diff
@@ -268,15 +176,37 @@ module.exports.getDefinitionList = async function (ctx) {
...
+    {
+      ...ctx.request.body,
+      appId,
+    },
+    { getResponses: true }
+  )
+  // save a test history run
+  await updateTestHistory(ctx.appId, automation, {
+    ...ctx.request.body,
+    occurredAt: new Date().toISOString(),
```
shogunpurple, line 208: NAB: Might be better to store this as a timestamp in the DB

packages/server/src/automations/triggerInfo/cron.js
```diff
@@ -0,0 +1,31 @@
+exports.defintion = {
```
shogunpurple, line 1: wee typo here

packages/server/src/automations/triggerInfo/index.js
```diff
@@ -0,0 +1,15 @@
...
+const rowUpdated = require("./rowUpdated")
+const webhook = require("./webhook")
+
+exports.definitions = {
+  ROW_SAVED: rowSaved.definition,
+  ROW_UPDATED: rowUpdated.definition,
+  ROW_DELETED: rowDeleted.definition,
+  WEBHOOK: webhook.definition,
+  APP: app.definition,
+  CRON: cron.defintion,
```
shogunpurple, line 14: actually it will still work because it's been imported correctly 😂 but the same typo here

packages/server/src/automations/triggers.js
```diff
@@ -255,14 +33,22 @@ async function queueRelevantRowAutomations(event, eventType) {
...
+    // don't queue events which are for dev apps, only way to test automations is
+    // running tests on them, in production the test flag will never
+    // be checked due to lazy evaluation (first always false)
     if (
-      !automation.live ||
-      !automationTrigger.inputs ||
-      automationTrigger.inputs.tableId !== event.row.tableId
+      !env.ALLOW_DEV_AUTOMATIONS &&
+      isDevAppID(event.appId) &&
+      !(await checkTestFlag(automation._id))
```
shogunpurple, line 42: 👌 that this has been prevented on the backend and not just in the UI

packages/server/src/automations/utils.js
```diff
@@ -0,0 +1,205 @@
...
+const newid = require("../db/newid")
+const { updateEntityMetadata } = require("../utilities")
+const { MetadataTypes } = require("../constants")
+
+const WH_STEP_ID = definitions.WEBHOOK.stepId
+const CRON_STEP_ID = definitions.CRON.stepId
+
+let workers = workerFarm(require.resolve("./thread"))
+
+function runWorker(job) {
```
shogunpurple, line 18: Is `worker-farm` necessary anymore because we can assume redis locally and just use bull? IIRC `worker-farm` was a threading workaround for the electron app locally

shogunpurple (approved):
LGTM! Some nice abstractions in here around appending data to metadata documents which is very useful - I can see us doing a lot more of that in future. I think it's much better to keep all that stuff in one document as well, because it would become very hard to manage in separate ones and makes it easy for us to poke around in fauxton. Good job 🚀

packages/server/src/api/controllers/automation.js
mike12345567: I thought about using a timestamp here, the only thing is that our system uses ISO strings for everything right now, do we want to make a decision on that?

Personally I think using ISO strings for users dates is good as it includes they selected timezone etc, but I think for storing dates/times that things happened internally we should just use timestamps as they are easier to work with in the DB etc.

PR #2610, opened 2021-09-15
Automation Redesign

PClmnt:
## Description
This PR completely redesigns the automations screen. 

- "Splash Screen" added when you first access the page
- New automation blocks
  - block setup is moved within the block
  - minimise blocks when they are completed
  - User is essentially guided through the flow. i.e adding another block is not possible until the current one is completed
  - delete specific blocks or the entire automation
- Ability to enter test data
  - validation on required test data based on schema
  - test data button only available on the last block
  - Add JSON instead of filling out the form yourself
- View test success / failure and results
- old icons replaced with spectrum icons

## Screenshots

image
image
image
image
image
image
image
image
image
image


joebudi:
Great work @PClmnt and @mike12345567 

A few design points:
- In the triggers modal - change the row created and row deleted icons to (https://spectrum.adobe.com/page/icons/  - search row and use 'table row add bottom and table row remove center)
- in the main interface - the xl icon is too big. Reduce by a few pixels - maybe 6 for height and width
- The text for the final CTA should read 'Finish and test automation'
- the tabs in the 'add test data' modal are indented somehow - this may be their design but it looks out of line
- the JSON input box is too small - maybe double its height
- the slack icon in the action panel is too small (I'll send you a diff icon on Discord)
- in re to actions - make the same changes as I mentioned above for the trigger
- test automations response modal - the error icon is not aligned properly. If this is not possible, just use the icon and change its background to red.
- test automations response modal - increase the size of the response boxes

I am happy to run through these minor UI changes with you on a call if you like.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ActionModal.svelte
```diff
@@ -0,0 +1,149 @@
...
+  import IntegromatLogo from "assets/integromat.png"
+  import SlackLogo from "assets/slack.svg"
+  import n8nlogo from "assets/n8n.png"
+
+  import { database } from "stores/backend"
+  $: instanceId = $database._id
+
+  let selectedAction
+  let actionVal
+  export let blockComplete
```
shogunpurple, line 15: NAB: We should probably have a lint rule or something for this, but generally the order of the svelte files are:

```
const atTheTop

export let propsAfterThat

let thenVariables

$: reactiveStuff

function utilityFuncsAtTheEnd() {}
```

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ActionModal.svelte
```diff
@@ -0,0 +1,149 @@
...
+
+  let externalActionsPredicate = [
+    { name: "zapier", logo: ZapierLogo },
```
shogunpurple, line 18: Looks like we could just spread these in from `ExternalActions.js`

shogunpurple (approved):
LGTM! This is awesome, automations looks so much cleaner and more functional now

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ExternalActions.js
```diff
@@ -0,0 +1,13 @@
+import DiscordLogo from "assets/discord.svg"
+import ZapierLogo from "assets/zapier.png"
+import IntegromatLogo from "assets/integromat.png"
+import SlackLogo from "assets/slack.svg"
+import n8nlogo from "assets/n8n.png"
+
+export const externalActions = {
```
mike12345567, line 7: Two things on this:
1. Right now we don't have an n8n action
2. I added a `internal: false | true` to the action definitions (was one of the last things that I did) - might be useful for determining the external actions!

PClmnt: Not quite sure why I added that haha, maybe because it was in the design, it won't show up in the UI as there was no backend for it, but i'll remove it anyway. And ah that;s very nice I hadn't noticed that flag, that'll actually simplify that area a good bit.

aptkingston (approved):
Awesome job! @joebudi has already gone over the few design tweaks so I haven't duplicated those.

One minor thing I can see in the screenshots that I haven't seen mentioned - the text for every action seems to say "When this happens", but shouldn't that be only for the trigger? I think the actions used to say "Do this" or something along those lines.

Great job though overall! 🚀

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -12,15 +12,34 @@
...
   $: bindings = getAvailableBindings(
-    block,
+    block || $automationStore.selectedBlock,
     $automationStore.selectedAutomation?.automation?.definition
   )
+  $: instanceId = $database._id
+
+  $: inputData = testData ? testData : block.inputs
+
+  const onChange = debounce(async function (e, key) {
```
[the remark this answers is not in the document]
PClmnt: I hadn't noticed that prop, I think I agree with sticking with debounce, there's a few common components used in the block setup, like ModalBindableInput for example. Debounce just makes it that little bit more generic.

PClmnt:
@aptkingston 

In some of the designs it has 'Do this' others have 'When this happens'.

What is the correct behaviour here @joebudi? Should we only should 'Do This' for the last block?

joebudi:
> Awesome job! @joebudi has already gone over the few design tweaks so I haven't duplicated those.
> 
> One minor thing I can see in the screenshots that I haven't seen mentioned - the text for every action seems to say "When this happens", but shouldn't that be only for the trigger? I think the actions used to say "Do this" or something along those lines.
> 
> Great job though overall! 🚀

joebudi:
> @aptkingston
> 
> In some of the designs it has 'Do this' others have 'When this happens'.
> 
> What is the correct behaviour here @joebudi? Should we only should 'Do This' for the last block?

Triggers have 'When this happens'
Actions have 'Do this'

mike12345567:
Been playing with this redesign locally, noticed a few issues, just putting them here for now:
1. The Binding drawer doesn't stretch across the screen to handle the area in which the setup panel used to appear:
image

2. This is a slightly difficult one to solve, but for the test modal when selecting a table ID it labels the table choice as `row` - I realise this is what it is called in the output definition of the automation trigger, but its a little confusing, this should probably be re-mapped to `Table` instead.
image

3. Getting the test modal to be considered valid is a bit difficult, it appears the validity of the form is based on the `id` field in the below example, if you unfocus the id field quickly after input it won't enable the test button (as seen below).
image

mike12345567 (approved):

PR #2618, opened 2021-09-16
Automation query functionality

mike12345567:
## Description
NOTE: this PR is to follow @PClmnt PR for automation redesign as it has been built atop it.

This solves issue #2410 by adding a new query rows step. This allows retrieving data within an automation run in a similar manner to that of a data provider in the frontend. It makes use of the search endpoint internally so it allows filtering the rows.

## Screenshots
The definition and setup of the new step:
image

The filtering drawer setup for automation:
image

The output from testing this new step:
image


shogunpurple (approved):
LGTM! Nice work - this will solve for a ton of use cases

PR #2630, opened 2021-09-17
Sanitise urls after change

### LINKED ISSUE #2611: Creating a screen with invalid route breaks entire app
Create a new screen and set it's route to `main` (i.e. no leading `/`).

This causes the whole app to break, the in-builder preview and the app itself fails to load, for any screen.

We should handle this better.

mslourens:
## Description
Fixes #2611 - Adds forward slash to route (for new screen and in settings section) and sanitises the rest of the url.


packages/builder/src/components/design/PropertiesPanel/ScreenSettingsSection.svelte
```diff
@@ -7,13 +7,17 @@
...
 
   export let componentInstance
   export let bindings
 
   function setAssetProps(name, value, parser) {
     if (parser && typeof parser === "function") {
+      console.log("before", value)
       value = parser(value)
+      console.log("after", value)
+      console.log("deepGet", deepGet($currentAsset, name))
```
PClmnt, line 20: do we need these logs in here?

mslourens: no, I forgot to remove them, I updated my PR.

PClmnt (approved):

PR #2668, opened 2021-09-20
Update app ID parsing to handle tenant ID's

aptkingston:
## Description
Due to tenant ID's potentially now being present in app IDs, the parsing logic needs to be updated to account for this. Currently creating an app with a tenant ID will cause the apps list screen to show no apps at all, as the parsing logic will silently fail.


shogunpurple (approved):
LGTM! Great job

PR #2672, opened 2021-09-20
fix the example of the avg helper

### LINKED ISSUE #2671: Math functions in handlebars that use arrays are broken
**Describe the bug**
Math functions in handlebars that use arrays always return 0 as value.

**To Reproduce**
1. Go to a number field
2. Set "Placeholder" to "{{ avg [1,2,3,4,5] }}" (without the quotes)

**Expected behavior**
Placeholder shows value "3".

**Actual behavior**
Placeholder shows value "0"

**Screenshots**
Inside editor:

image

Runtime:
image

 - OS: Windows 10 Pro 2004 (OS Build 19041.1237)
 - Browser: Google Chrome
 - Version: 93.0.4577.82 (Official Build) (64-bit)

mslourens:
## Description
Fixes #2671 - The example for the AVG helper is incorrect. The code shown is not working, the unit test was using a different syntax, which I now use for the example.

## Screenshots
image

### DIFF
--- a/packages/string-templates/manifest.json
+++ b/packages/string-templates/manifest.json
@@ -22,7 +22,7 @@
         "array"
       ],
       "numArgs": 1,
-      "example": "{{ avg [1,2,3,4,5] }} -> 3",
+      "example": "{{ avg 1 2 3 4 5 }} -> 3",
       "description": "<p>Returns the average of all numbers in the given array.</p>\n"
     },
     "ceil": {


mike12345567:
Hi @mslourens - thanks for raising this, I think it would probably be good for us to modify the average function to work with arrays as well as lists of values passed through rather than adjusting the example.

Also, the `manifest.json` file is actually generated, from comments written in the codebase, we have a fork of the library: https://github.com/Budibase/handlebars-helpers

mslourens:
handlebars doesn't understand arrays in the template, just separate (primitive) arguments and a context (which can contain arrays). If I pass `{{avg [1 2 3 4]}}` in the handlebars.compile without a context, the arguments are:
```javascript
[
   undefined,
  {
    lookupProperty: [Function: lookupProperty],
    name: 'avg',
    hash: {},
    data: { root: undefined },
    loc: { start: [Object], end: [Object] }
  }
]
```
I can update the comment in the branch you mentioned.

mike12345567:
Ah I see what you mean now, yes we would simply need to update the comment in the handlebars template branch in that case! If you're able to do that thank you, the process is a little complicated but the idea was to use the existing comments from the library.
To update you update the comments on the `budibase/handlebars-helpers`, get that merged and it'll be released, then update the library in the `string-templates` and run `yarn manifest` command to generate the new manifest.

mslourens:
I created a [PR](https://github.com/Budibase/handlebars-helpers/pull/4) in the handlebars-helpers repo

mslourens:
When I update the library to 0.11.5 in the string-templates, the changes are not reflected in the node_modules. According to the commits in the [tag](https://github.com/Budibase/handlebars-helpers/releases/tag/0.11.5) my changes should be in there. 
 Do I have to do something special to get it released?

mike12345567:
Sorry @mslourens - I forgot to bump the version, will do that now!

mslourens:
@mike12345567 could you have a look at this PR, when updating the string-templates with version 0.11.5 of the handlebar-helpers package, the manifest remains unchanged.

mike12345567:
Hi @mslourens - sorry for the problems with it, I've gone ahead and pushed an update for everything directly in commit 485d7045866b88b2df409b29910c74f278999f18 - will close this PR.

PR #2677, opened 2021-09-20
Drag and drop

aptkingston:
## Description
This PR adds drag and drop functionality to the visual app preview within the builder. Components can be dragged before, inside or after other components, depending on what is valid behaviour. Dragging to the top or left edges represent "before", while the bottom and right edges of a component represent "after".

By far the easiest way to get a feel for this is by using it, so please do check out this branch and try it. Drag and drop is sometimes quite subjective so I'd be keen to get other opinions on this.

Here's a quick video overview:
https://user-images.githubusercontent.com/9075550/134028220-793a44de-774f-45e6-ad3d-b4843cb73da7.mp4

### Usage
There are a few rules built in to make the experience better, and a few rules built in that technically need to be there.

- To move a component inside another one, drag the component into the empty component. The "inside" function is only available for empty components - otherwise you need to drag before or after an existing child component.
- To move a component before another one, drag the component near the top or left edges of the other component
- To move a component after another one, drag the component near the bottom or right edges of the other component
- For components that support before, inside and after targets, dragging near the top/left edges means before, the bottom/right edges means after, and the center of the component means inside

### Visual feedback
To make the experience simple and easy, there is visual feedback when dragging and dropping. A green line represents where the current action is targetting and is positioned to the correct side of the component being targetted, or fully bordering the target component for and "inside" target. There is a label stating exactly what position is being targetted, which removes any ambiguity.

Dragging above or below:
image
image

Dragging left or right:
image
image

Dragging inside:
image

The hover border and settings bar are both hidden while dragging, to reduce visual clutter. The component being dragged becomes selected, and the selected component styles remain visible on it to make it clear which component is being dragged.


joebudi:
This is amazing. Looking forward to testing it out.

mslourens:
Very nice! Is it intended behavior that you can only drop in a column? I have a container in a row direction. The left side contains the form, the right side an image. I wanted to drag the image to the left side of the form, but was not able to do that. Then I tried to drag the image inside the form, below a control, which worked, but I didn't like the overall look, so I wanted to move the image back to the right container, which was not possible. 
Here is a movie where you can see what I mean:

https://user-images.githubusercontent.com/1907152/134055292-c2758312-6ee1-4f7e-b88a-0b9caf52bda5.mp4

aptkingston:
> 
> 
> Very nice! Is it intended behavior that you can only drop in a column? I have a container in a row direction. The left side contains the form, the right side an image. I wanted to drag the image to the left side of the form, but was not able to do that. Then I tried to drag the image inside the form, below a control, which worked, but I didn't like the overall look, so I wanted to move the image back to the right container, which was not possible.
> Here is a movie where you can see what I mean:
> dnd-horizontal.mp4

Thanks for checking it out @mslourens! You're right that it's only the top and bottom edges that control positioning at the moment. Unfortunately it's pretty much impossible to work out if we're in a horizontal or vertical container because the tree could be many layers deep.

You can technically still use this at the moment by dragging a component above another to position it to the left of it, or the bottom for right, but it just looks wrong to do that at the moment.

One idea I had was using the side edges to also mean before and after. So both the top and left edges would represent "before" and both the right and bottom edges would mean "after". This way it would look a bit better, since you could drag a component to whatever edge makes logical sense for you. I think I'll implement this idea and then we can see if it feels better!

aptkingston:
@mslourens if you get a chance to try this after my latest update then I would love to hear your feedback. I've added the ability to target the left and right edges of components which should make things feel much more natural when working inside horizontal layouts.

mslourens:
@aptkingston here are some of my findings:

- you cannot drop a container/card after a form, it will always drop it after the last form control inside the form
- you cannot drop a container/card after a repeater, it will always drop it inside the repeater
- you cannot drop a container after a dataprovider, it will drop it inside the dataprovider
- you cannot drag a whole form at once, it will always drag a control
- if you create a container with a row direction and you drag one item inside that container below another, it shows as if that is possible, but when you release the item, it will be moved next to the item (because of the row direction of the container). It's probably too much to ask to disable the indicator that the item can be dropped below the other one.
- dragging a table is a bit difficult, because it's hard to see where you can drop it (the table blocks the visible indicators). You could make the dragged item a bit more transparent, in order to see the drop indicators.
- it's also hard to drag a dataprovider with a table. Nine times out of ten I only drag the table, and leave the dataprovider behind (which results in an empty table, which is not draggable anymore). A workaround is to grap the table near the bottom, next to the pagination (which is only possible if pagination is enabled).
- when you have dataprovider with a table and want to drag, for example, a card next to the table, the card ends up inside the dataprovider, above the table. Maybe dataprovider should be used as drop targets? 
- it is not possible to drag a repeater without a gap which is inside a dataprovider without pagination 

That's it for now. My overall impression is that this is very user friendly and much easier to use than the drag/drop in the ComponentNavigationTree.

aptkingston:
> 
> 
> @aptkingston here are some of my findings:
> 
>     * you cannot drop a container/card after a form, it will always drop it after the last form control inside the form
> 
>     * you cannot drop a container/card after a repeater, it will always drop it inside the repeater
> 
>     * you cannot drop a container after a dataprovider, it will drop it inside the dataprovider
> 
>     * you cannot drag a whole form at once, it will always drag a control
> 
>     * if you create a container with a row direction and you drag one item inside that container below another, it shows as if that is possible, but when you release the item, it will be moved next to the item (because of the row direction of the container). It's probably too much to ask to disable the indicator that the item can be dropped below the other one.
> 
>     * dragging a table is a bit difficult, because it's hard to see where you can drop it (the table blocks the visible indicators). You could make the dragged item a bit more transparent, in order to see the drop indicators.
> 
>     * it's also hard to drag a dataprovider with a table. Nine times out of ten I only drag the table, and leave the dataprovider behind (which results in an empty table, which is not draggable anymore). A workaround is to grap the table near the bottom, next to the pagination (which is only possible if pagination is enabled).
> 
>     * when you have dataprovider with a table and want to drag, for example, a card next to the table, the card ends up inside the dataprovider, above the table. Maybe dataprovider should be used as drop targets?
> 
>     * it is not possible to drag a repeater without a gap which is inside a dataprovider without pagination
> 
>     * dragging an empty section drags the whole page, although the section is not the parent of the rest of the page.
> 
> 
> That's it for now. My overall impression is that this is very user friendly and much easier to use than the drag/drop in the ComponentNavigationTree.

Wow that's a detailed a list! Nearly all of these are the same thing - when there is no padding around a component it's impossible to target it for dragging or dropping and you'll always target a child component. I don't currently know if that can be improved.

> if you create a container with a row direction and you drag one item inside that container below another, it shows as if that is possible, but when you release the item, it will be moved next to the item (because of the row direction of the container). It's probably too much to ask to disable the indicator that the item can be dropped below the other one.

Dragging and dropping only moves components, but it doesn't change any settings. So although it might look like you can drag something to an explicit side, the only options are actually "before" and "after". Where it goes is determined by whatever styles are applied to the container you're in. It's (pretty much) impossible to know where the component would actually go at the moment because of how flexible everything is, but I wish it possible!

> dragging an empty section drags the whole page, although the section is not the parent of the rest of the page.

I can't seem to reproduce this but I'll keep trying.

> dragging a table is a bit difficult, because it's hard to see where you can drop it

I think this must depend on the browser (in Firefox for me it is very transparent). I'll see if it's possible to manually control this, because yea we definitely want it very transparent.

Really appreciate the feedback @mslourens, thank you!

mslourens:
This is the issue with dragging the section:
https://user-images.githubusercontent.com/1907152/134161376-85827a8c-73b5-4ce7-8ab0-f7181d3fe587.mov

The drag item is much more transparant in Firefox, this is how it looks in Vivaldi (Chromium based browser):
https://user-images.githubusercontent.com/1907152/134161932-61931d94-c7cf-40d3-8ef8-de77c8a54724.mov

shogunpurple (approved):
Awesome work cheeks - svelte makes it so clean to implement this kind of stuff, which is usually pretty nasty in other frameworks or even in pure vanilla JS 🚀

PR #2678, opened 2021-09-20
Automation improvements

PClmnt:
## Description

- Adds datetime / boolean / multi-select types to automations
- Show confirmation modal when deleting automation / change colour of delete icon hover
- Fix app actions, previously you were unable to select a type and adding / removing was temperamental.  

## Screenshots
image

image


aptkingston (approved):
Nice one. FYI the BBUI `Icon` component has a `hoverable` prop that will change colour for you on hover! And you can also just apply the `on:click` handler directly to the `Icon` as well.

PR #2686, opened 2021-09-21
replace all handlebars bindings, not just the first

### LINKED ISSUE #2679: Handlebars only seems to expand variables one time
**Describe the bug**
Handlebars only seem to expand variables one time. If the variable is used twice, the second time it is ignored.

**To Reproduce**
1. Create a new form with 2 number input fields named "a" and "b"
2. Set the placeholder for "b" to "{{ sum nf.Fields.a nf.Fields.a }}" (without the quotes)
3. Check the result in "b" for any given input of "a". When "20" is entered into "a", you would expect "40" as a result in field "b", but the actual result is "20".

**Expected behavior**
Variables are expanded as many times as they occur inside the handlebar string.

**Screenshots**
image

OS: Windows 10 Pro 2004 (OS Build 19041.1237)
Browser: Google Chrome
Version: 93.0.4577.82 (Official Build) (64-bit)

mslourens:
## Description
Fixes #2679 - When using bindings in handlebars helpers, only the first binding is replaced with the real value. This PR fixes that by replacing all bindings inside the string.

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -443,7 +443,10 @@ function bindingReplacement(bindableProperties, textWithBindings, convertTo) {
     for (let from of convertFromProps) {
       if (shouldReplaceBinding(newBoundValue, from, convertTo)) {
         const binding = bindableProperties.find(el => el[convertFrom] === from)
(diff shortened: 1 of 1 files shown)


mike12345567:
Thanks for this!

PR #2698, opened 2021-09-21
Cloud View replacement

mike12345567:
## Hacktastic Description
This is a PR for the work to replace CouchDB views with in memory versions for the Cloud. This is required due to some of the CPU spikes we were experiencing at view creation.

The first step is that when `SELF_HOSTED` env variable is falsy views will be written to a separate document rather than to the design database. I played around with the idea of leaving the view definitions in the design DB, as if they aren't queried the view isn't built, but I believe that the next time a row is written it will start to index that row, which creates a lot of un-necessary performance cost, considering we're never using the index.

After this we make use of the in memory view, this is found in the `packages/server/src/db/inMemoryView.js` file which does the following:
1. reads all the rows from the specified table
2. Creates an in memory PouchDB instance, which it then writes the table to via a `bulkDocs` call
3. Takes the view definition which was written to a doc in CouchDB, converts the map function string to a real function using `eval` (this is a bit hacky, but the easiest way to do this)
4. Uses the function to carry out a temporary PouchDB query
5. Destroys the in memory database

I did a lot of performance testing around this and found this approach to be the fastest, without writing the view indexing, reduction and grouping functionality from scratch. For a large number of large docs it can take 100-200ms to carry out the writing of the docs, the query and the deletion, but reading those docs takes seconds so its very small in the grand scheme of things.

This branch also comes with a few performance improvements that I made when I was testing, namely around relationships and creating tables from large CSVs. Due to the created by and updated by relationships this would lead to hundreds/thousands of relationships being created to the builder user, which was wasted time as I couldn't think of a good argument for doing this. Therefore I disabled relationship creation during CSV import.

## Notes
Reading the docs out takes a fair amount of time, however if we were to implement a caching layer for requests this time would only be spent the first time it was queried (until the cache entry was invalidated). I think it would be worthwhile to consider a cached responses system for raw database rows, which we could invalidate when someone writes to the table, this would be a simple but effective way to reduce database load significantly. My thinking on this would be that instead of calling `db.allDocs` or `db.get` directly, we would call it through a cache layer, which looks to see if the keys in question have been queried and the results are stored in Redis, in which case it returns the rows from a cache entry in Redis rather than CouchDB. When a `row:save`, `row:update` or `row:delete` event is fired we delete all the cache entries for the tableId and for the rowId. If the relationship system was updated to use this as well then we could see a massive improvement in relationship performance as well.


aptkingston:
While I'm extremely disappointed to not see the word `hacktastic` in the description even once, this was a really quick turnaround and a nice solution for a tricky problem, so nice work. The `eval` solution for map functions is particularly wonderfully hacky yet elegant and I love it.

mike12345567:
> While I'm extremely disappointed to not see the word `hacktastic` in the description even once, this was a really quick turnaround and a nice solution for a tricky problem, so nice work. The `eval` solution for map functions is particularly wonderfully hacky yet elegant and I love it.

I'm sorry, really screwed up on the description, it has been fixed!

aptkingston:
> 
> 
> > While I'm extremely disappointed to not see the word `hacktastic` in the description even once, this was a really quick turnaround and a nice solution for a tricky problem, so nice work. The `eval` solution for map functions is particularly wonderfully hacky yet elegant and I love it.
> 
> I'm sorry, really screwed up on the description, it has been fixed!

🙏 much appreciated!

shogunpurple (approved):
LGTM! Nice work on this, it worked out well that you were able to leverage pouch and a whole bunch of mature code to handle the views, rather than reimplementing the API in JS. I also like the migration path from self host to in memory views too. 🚀

PR #2703, opened 2021-09-21
Feature/analytics

shogunpurple:
## Description
This PR contains code for all the analytics we will run in the cloud.

- Refactored the current analytics setup to use a class based approach, with different clients that get injected into the main "hub". This is because certain events get sent to more than one place, so putting a common interface in front of them allows us to send data to several of these sources utilising the same analytics client.
- The analytics hub can be turned on and off, and if it's disabled calls to the clients will simply be noops - the same applies if a client has not been configured properly too.
- `analytics` can be turned on and off per tenant in the organisation settings
- Added intercom integration into the budibase builder. There's still a bit to do here once we have the accont portal onboarding completely wired up - we only want to show the intercom bubble to certain orgs
- Turned off analytics completely for self hosting. It's not really relevant any more and we do not want to track user data for self hosted installations.

## Screenshots


joebudi:
Great work Marty. An essential piece of work that will benefit us greatly in the short and long term.

Rory-Powell (approved):

aptkingston (approved):
Awesome. Super clean implementation too 👍 Nice one Marty.

PR #2706, opened 2021-09-22
Fix home screen importing + extras

### LINKED ISSUE #2298: Home screen replacement on app import
Hi all,

Finally I investigated (caught) home screen replacement on app import.

**Describe the bug**
After app import existing home screen replaced by empty one.

**To Reproduce**
1. Import an BB App w/ Home screen customised (not default/empty one)
2. RIGHT AFTER APP IMPORT -> Click on App Dev Preview (play icon at top right corner)
3. Your Home screen is gone - "Welcome to budibase" instead of ex.home screen

The only way to retain ex. home screen is do not click on play button right after app import.
Currently, the right way after app import - click on Design Tab. And home screen will be appeared. Then you can click on play button.

aptkingston:
## Description
This PR fixes the issue of home screen replacement on app import. Closes #2298.
This was actually really simple. The poorly named `createEmptyAppPackage` actually only creates the default home screen and layouts, and we were always calling this regardless of whether we were importing an app or not, so all we need to do is not call this when importing an app.

This also fixes a couple of issues with the expandable datasource work in the backend UI that was introduced recently:
- If a table is selected, the corresponding datasource will be expanded initially and always stay expanded
- If the internal table data source is initially selected, actually select it in state
- Don't reset datasource selection when fetching datasource options unless the selected datasource no longer exists


mike12345567:
Good find! This seemed like a very hard to find bug, amazing getting it so quickly!

PR #2707, opened 2021-09-22
Fix formula constraints check

### LINKED ISSUE #2643: Use Formula Field as table display column
### Discussed in https://github.com/Budibase/budibase/discussions/2641

it's possible to set a formula field as a display column. However, Budibase will not allow us to create new rows - the error is "<Fomula field name> cannot be empty" - even when the fields that make up the formula are entered.

Either
- Formulas as display columns should not be allowed
OR
- This should work, provided the formula result is not empty

mslourens:
## Description
Fixes #2643 - Check if the formula validates against its constraints by processing the formula string and validate it with validate.js. When the fields used in the formula are all empty, the formula is invalid, otherwise the formula is valid.


shogunpurple (approved):
Nice one @mslourens!

Issue #2673, opened 2021-09-20
Save Row (insert) action to postgres fails, but row still inserts


mjashanks:
I have 2 postgres tables
- Vehicles
- ServiceLog

One Vehicles has Many ServiceLog

Create scripts for tables, with data are attached.

Now, I ....
1. create a "New Row" form, for "ServiceLog"
2. Remove unnecessary fields - VehicleId  and ID
3. Run my application, and open my screen
4. Fill out form and press save

RESULT: 
- the request to `POST /api/table_id/rows` fails with `400 - Bad Request`. There is no response body.
- The row is saved to the database anyway.

**My Request**

```json
{
  "Vehicle": [
    "%5B'7'%5D"
  ],
  "Mileage": 444444,
  "ServiceDate": "2021-09-29T23:00:00.000Z",
  "Description": "hello",
  "__valid": true,
  "__currentStep": 1,
  "__currentStepValid": true,
  "Category": "Brakes",
  "tableId": "datasource_plus_238519b3a39b417cb2e58646988e1d67__ServiceLog"
}
```

**This request works...**
I got this by inserting a row via the builder

```json
{
  "Category": "cat",
  "Description": "description",
  "ServiceDate": "2021-09-08T11:00:00.000Z",
  "VehicleId": "",
  "Mileage": 66666,
  "id": null,
  "Vehicle": [
    "%5B'3'%5D"
  ],
  "tableId": "datasource_plus_238519b3a39b417cb2e58646988e1d67__ServiceLog"
}
```

**Create script for tables...**

```sql
CREATE TABLE IF NOT EXISTS public."Vehicles"
(
    id bigint NOT NULL GENERATED ALWAYS AS IDENTITY ( INCREMENT 1 START 1 MINVALUE 1 MAXVALUE 9223372036854775807 CACHE 1 ),
    "Registration" text COLLATE pg_catalog."default",
    "Make" text COLLATE pg_catalog."default",
    "Model" text COLLATE pg_catalog."default",
    "Colour" text COLLATE pg_catalog."default",
    "Year" smallint,
    CONSTRAINT "Vehicles_pkey" PRIMARY KEY (id)
)

TABLESPACE pg_default;

CREATE TABLE IF NOT EXISTS public."ServiceLog"
(
    id bigint NOT NULL GENERATED ALWAYS AS IDENTITY ( INCREMENT 1 START 1 MINVALUE 1 MAXVALUE 9223372036854775807 CACHE 1 ),
[...]

PR #2709, opened 2021-09-22
Fix for issue #2673

mike12345567:
## Description
Fixing issue #2673 - discovered issues with some column types not being worked out to be auto-columns in postgres.

This was blocking @mjashanks so needed fixed asap.


mjashanks:
Get in! Nice work @mike12345567

Issue #2708, opened 2021-09-22
Queries should also part of the expanded panel in the DatasourceNavigator


mslourens:
**Describe the bug**
Datasources can be expanded and collapsed, but queries are not part of the expanded panel, which looks weird. I think queries belonging to datasources should be part of the expanded/collapsed panel. 

**Screenshots**
![image](https://user-images.githubusercontent.com/1907152/134375449-13f5e8ff-ae39-4a32-a04f-0fb1ebfbd85f.png)
- getPosters is a MongoDb query and should be collapsed with the MongoDb datasource
- updateBooks is a Postgress query and should be collapsed with the Postgres datasource

**Desktop (please complete the following information):**
 - OS: MacOS 11.6
 - Browser: Vivaldi
 - Version: develop branch

aptkingston:
Thanks @mslourens!

Issue #2155, opened 2021-07-28
[Feature request] Collapse/Expand database connections


mslourens:
I have a database with a lot of tables. It would be nice to be able to collapse the tables per database connection in the data tab. Currently it's difficult to work with multiple connections which contain multiple tables each.

PR #2714, opened 2021-09-23
Collapse queries with their datasource

mslourens:
## Description
With issue #2155 datasources can be collapsed and expanded. But queries belonging  to those datasources were left untouched, which results in a strange looking DatasourceNavigator, where datasources are collapsed, but their queries are not. This PR fixes that (issue #2708)

## Screenshots
image
_Collapsed_

image
_Expanded_


aptkingston (approved):
LGTM! I would definitely call this a bug - this behaviour looks much better now. Thanks Maurits.

Issue #2427, opened 2021-08-20
Refreshing datasources on an interval and manually


aptkingston:
Discussed in https://github.com/Budibase/budibase/discussions/1609.

We should allow data sources to be manually refreshed both on an interval and on a button click. We could add 2 settings to the data provider component - a checkbox to say that you want it to refresh on an interval, and then a condtional number field for the duration in minutes to refresh.

We should also add a button action for refreshing a data provider on click.

bobhenkel:
My first 10 minutes with budibase I setup a button to execute my data source query to see if the populated visual table would refresh, with no luck.  Looking forward to this.  Hitting refresh on the browser kind of breaks the app usage flow.

shogunpurple:
Hi @bobhenkel - which data source did you use? It would be good to get reproduction steps so we can help debug. Thanks!

bobhenkel:
Hello @shogunpurple - I'm using postgres  with a table as the data source.  Here's a short walk through of what I was trying to do.
[https://odysee.com/@OpenSourcererSystems:d/chicken-nugget-strikes-again:d](https://odysee.com/@OpenSourcererSystems:d/chicken-nugget-strikes-again:d)
Thanks!

PR #2730, opened 2021-09-23
Add button action for manually refreshing a data provider

aptkingston:
## Description
This PR adds a button action which can be configured to manually refresh a certain data provider. This has been requested a few times. The core functionality for this was already implemented and is used for the automatic refreshing that's built in to the client library, so this PR simply adds another entrypoint to use that functionality.

Worth nothing that this shouldn't normally need to be used - it's specifically for cases where the automatic refreshing fails, which is for things like REST datasources where each query has a different datasource ID that breaks the automatic refreshing.

Addresses part of #2427, but the automatic refreshing part still needs to be done. @mjashanks was after this today which is why I'm just implementing the manual part for now, and it's a decent standalone feature.


mjashanks:
Thank you @aptkingston ! REST Api guide can be completed now :) - and a feature I've needed a few times in the past.

aptkingston:
Gonna go ahead and merge this in since it's a pretty small-scoped change (and it lets Mike get his hands on it).

PR #2735, opened 2021-09-24
Datasource wizard

PClmnt:
## Description

(When I say wizard, I mean 2 modals)

This PR changes the functionality of the data screen somewhat, in that now the user has to go through a couple of modals in order to create a datasource. 

The datasource select modal has been re-designed, and then the configuration part of the datasource setup has been added to the following modal. So you enter your config there and click save and you're brough to the existing datasource page. If it's a datasource plus then it automatically fetches the schema / tables for you.

This modal also triggers upon first entry of the data screen, i.e encouraging the user to setup a datasource or internal table before doing anything else.  

## Screenshots
image

image

image

image

image

## Questions
If the user clicks `Start from scratch`, we never want to show them this modal automatically upon navigation to the data page, right? If so, I will need to add a flag that denotes that when they click that button. But just getting confirmation on that first


joebudi:
Hey @PClmnt 

I have a few points:

Initial modal
- Instead of 'Add data' change the title to 'Data'
- remove the text 'it's simple!'
- the 'connect to data sources' label looks small, and has too much spacing underneath it. Reduce spacing, and increase the size / change type to 'detail'

Data source modal

- The title of the modal should be Connect to {{ datasource.name }} or something
- The modal should have a description under the heading  - "Connect your database to Budibase using the config below"
- SSL label does not have an input ? Should it? If not, take it off.
- I'm not sure about 'add' button? It provides no insight into what it does. Is this linked to SSL? I think this is an issue with the actual config - the plus button should be aligned with SSL if it is linked.

Your question
Change the 'start from scratch' button to 'Create a new table with Budibase DB' - then the next modal is the same modal they would see if they select Budibase DB as their datasource.

aptkingston:
Nice one! Few points (in addition to Joe's, I agree with that he said):
- I don't think we need the "no prior data required" label
- What does start from scratch actually do? Does it just skip the creation of a datasource?
- Does the modal currently check the validity of the connetion? I know this is probably a pre-existing issue, but it would be awesome if we actually tried to connect and didn't close the modal if connecting failed, and provide feedback either way. The modal supports an async callback, and if you return `false` then it won't close, so it would be awesome to add this.

PClmnt:
@joebudi should be all of those fixed now

@aptkingston All of yours fixed as well, it's now impossible to save your config unless it has fetched the schema successfully.

aptkingston:
Changes look good 👍 last question - why is there the second button for "Create a new table with Budibase DB"? Can you not just choose "Budibase DB" as the source and press continue?

PClmnt:
@aptkingston I think @joebudi mentioned having this instead of 'start from scratch' in the above comment, it's slightly redudant, but sort of like a failsafe for a user that isn't sure exactly what they should do?

aptkingston:
> 
> 
> @aptkingston I think @joebudi mentioned having this instead of 'start from scratch' in the above comment, it's slightly redudant, but sort of like a failsafe for a user that isn't sure exactly what they should do?

Ah ok. Yea I wasn't sure what the "start from scratch" button did either, but this new text also confuses me haha. Personally I really don't think it's needed and it just causes confusion, but if other people feel it adds some value then that's fine.

packages/server/src/api/controllers/datasource.js
```diff
@@ -80,13 +75,20 @@ exports.update = async function (ctx) {
 }
 
 exports.save = async function (ctx) {
+  console.log(ctx.request.body)
```
aptkingston, line 78: Wee log statement left in here accidentally.

aptkingston (approved):
LGTM! Nice job 🚀 Just a wee log statement left in but that's it.

joebudi:
@PClmnt remove the secondary button as @aptkingston requested. I can run an ab test later on to see if adding the button has any effect on adoption.

PR #2738, opened 2021-09-24
Fix regex binding replacement issue due to not being escaped

aptkingston:
## Description
This PR fixes an issue introduced in https://github.com/Budibase/budibase/pull/2686, where bindings were showing up totally mangled due to the fact that the new regex expressions being created were not escaped.

Bindings were being generated that looked like this:
image

A quick regexr example shows what was happening - using this as a literal regex expression was replacing some random characters in the string with the full expression.
image

This mostly affected the runtime to readable conversion, so sometimes the saved version in the DB would work correctly but just be displayed incorrectly.

I've reverted the functionality to a simple string replace, but now just calling it repeatedly until every string instance has been replaced. This way we don't need to worry about trying to escape every regex character.


mike12345567:
Thanks for this, I was having issues with it inserting mangled bindings into Postgres!

Issue #2612, opened 2021-09-15
Postgres update query fails when data contains newline character \n


MarkOSullivan94:
**Describe the bug**
Unable to successfully execute update query when sending data via parameter which contains a new line character `\n`

After clicking save, I receive back a 500 error code response.

The request payload sent in the POST request

```
{"parameters":{"new_biography":"This is some text\nThis is some more text","new_age":"30","new_name":"Mark","new_id":"1","old_id":"1"}}
```

I have tested editing with jsonb data which does not contain newline characters and it updates successfully, only time I have an issue is whenever I encounter data which contains `\n` and I'm guessing this might be the same for other characters like `\r` which I've seen in our data too.

**To Reproduce**
I've been able to successfully replicate this in a separate test table. Outlined below are the steps to replicate.

1. Have a Postgres DB source setup
2. Create a table containing `id` field and a `data` field which stores jsonb (see below for data structure used)
3. Ensure the first record has, within the `data` column a value which is using new line character `\n`
4. Create a view which will be used to display the data in the Budibase app - in this case I called mine `cms_test` and you can see how it was created below
5. Tap `fetch tables from database` to ensure you have access to the new view
6. Add the detail screen for this new view
7. Adjust your navigation links so it will open up the record stored in your table
8. Hook up the `Save` button so that on click it executes your update query (shown below is the values I'm using for the query parameters) 
9. Update one of the fields (but keep field with new line character)
10. Click save

**View creation**
```
CREATE VIEW "public"."cms_test" AS SELECT
	x.id,
	x.age,
	x.name,
	x.biography
FROM
	test c,
	LATERAL jsonb_to_record(c.data) x (id character varying(32), 
		age BIGINT,
		name TEXT,
		biography TEXT
		);
```

**Update query**
```
UPDATE test SET data = '{"id": {{new_id}}, "name": "{{new_name}}", "age": {{new_age}}, "biography": "{{new_biography}}"}'::jsonb, id = '{{new_id}}' WHERE id = '{{old_id}}';
```

**Expected behavior**
I expected to get a success message and the data to be updated successfully with no issues.

**Screenshots**

**1. Data structure**
<img width="771" alt="Screenshot 2021-09-15 at 10 31 32" src="https://user-images.githubusercontent.com/6950843/133408839-ed0e0b18-e927-4177-9669-bd5cb0cc8137.png">

**2. Example data**
[...]

PR #2756, opened 2021-09-27
Allow newlines in Postgres JSON inputs

mike12345567:
## Description
This issue is quite specific to Postgres, when carrying out queries that include JSON there is an opportunity for the a newline to cause a break in the query. This issue is documented in #2612.

I've created an example database that can be used to show the symptom of this problem for testing (found in the `server/scripts/integrations` directory) and I have fixed the issue by putting together a regex for Postgres queries which includes JSON. It will find the JSON input in the query and make sure that newlines have been escaped correctly.

This is a simple and quick fix for Postgres - rather than doing an expansive fix which could break other things I felt a narrow fix would be best.

### DIFF
--- a/packages/server/scripts/integrations/pg-json/docker-compose.yml
+++ b/packages/server/scripts/integrations/pg-json/docker-compose.yml
@@ -1,0 +1,28 @@
+version: "3.8"
+services:
+  db:
+    container_name: postgres-json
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM! Nice fix 🚀

PR #2771, opened 2021-09-28
adding subdomain checking to tenancy middleware

shogunpurple:
## Description
Set at tenantId from the subdomain in the URL if the other methods fail.

Allows `mytenant.budibase.app` in cloud.

## Screenshots

### DIFF
--- a/packages/auth/src/tenancy/context.js
+++ b/packages/auth/src/tenancy/context.js
@@ -53,6 +53,11 @@ exports.setTenantId = (
   // processed later in the chain
   tenantId = user.tenantId || header || tenantId
 
+  // Set the tenantId from the subdomain
+  if (!tenantId) {
+    tenantId = ctx.subdomains && ctx.subdomains[0]
+  }
+
   if (!tenantId && !allowNoTenant) {
     ctx.throw(403, "Tenant id not set")
   }


Rory-Powell:
For the corresponding part to this we'll have to get account portal redirecting to `{tenantid}.budibase.app` instead of `budibase.app` as it does atm

packages/auth/src/tenancy/context.js
```diff
@@ -53,6 +53,11 @@ exports.setTenantId = (
   // processed later in the chain
   tenantId = user.tenantId || header || tenantId
 
+  // Set the tenantId from the subdomain
+  if (!tenantId) {
+    tenantId = ctx.subdomains && ctx.subdomains[0]
```
Rory-Powell, line 58: do you think we need a check here for when there is a user and tenant id but it doesn't match the tenant being requested? e.g. I'm logged in to `abc.budibase.app` but I navigate on over to `xyz.budibase.app`. I think atm it will silently ignore the `xyz` part and serve me data from `abc`. We'd probably want to redirect them away or fail the request

Rory-Powell (approved):

packages/auth/src/tenancy/context.js
shogunpurple: Will it even reach this branch if that's the case? It should be set earlier on in this middleware, I think

Rory-Powell: I think it's still an issue, here's an example where my session is scoped to `rpowell`
Screenshot 2021-09-29 at 11 09 54

But I nav to `disney` and budibase doesn't complain and still serves me `rpowell` data
Screenshot 2021-09-29 at 11 10 06

If the tenant doesn't match we should probably kick them back to account portal (which should send them to the correct tenant if they don't have account portal access)

Rory-Powell: probably easiest to prevent this in the svelte app, and just do a sanity in the backend by throwing a 403

PR #2780, opened 2021-09-29
Fix bug with filter editor preventing some types being filtered

aptkingston:
## Description
This fixes a bug I spotted which currently prevents filtering on `datetime` or `boolean` fields. The syntax for checking if the field type is `options` or `array` is invalid and is instead evaluating `if "array"` which is always true, therefore preventing the further conditions from ever being evaluated.

The scope of this is just that if you were trying to filter a data provider on a date or boolean field then you be shown a combobox to enter your value rather than a datepicker or a checkbox.

This does currently exist live on master as well. It looks like a small oversight when the recent work on the multi-options data type was merged in.


shogunpurple (approved):
LGTM

PR #2789, opened 2021-09-29
Tenant deprovisioning

Rory-Powell:
Delete all tenant data when the root account is deleted
See: https://github.com/Budibase/account-portal/pull/32


shogunpurple (approved):
LGTM! Nice work

PR #2790, opened 2021-09-29
Data UI pagination and searching

aptkingston:
## Description
This PR adds pagination and full searching to the tables in the data UI page.

### Pagination
All internal tables and "plus" datasources are now paginated in pages of 10. A new core implementation of the lucene searching has been added to the builder and we will eventually use this in the client as well, but currently the logic is duplicated in the `DataProvider` component to avoid the risk of additional changes before sandbox launch.

image

Fetching pages correctly handles loading states and can be spammed as quickly as you like, it won't cause any issues.

### Searching
All internal tables and "plus" datasources can now be fully searched using the exact same powerful system that we have for the design section. A new "Filter" button now appears for tables as well as views, which opens the filter editor.

image

The filter editor feels familiar as it is identical to the one available in the design section, with the exception of not being able to use bindings here.

image

The filter button is highlighted when a filter is applied.

image

### Sorting
Previously, clicking a table column sorted the data client side which simply sorted the data you could see. This is useless when pagination is enabled, as it will only sort the data in each page. New functionality has been added to the table to enable it to broadcast sort events and disable client side sorting, and this has been integrated with the new data UI tables. The result is that clicking on a table column header now actually refetches the data with a new server sorting order applied. The interaction from the user is the same, but now it just works as expected for paginated data. This will eventually be built into the client library too, but this has not been done yet.

## Other stuff
- Removed whole route section under `.../datasource/bb_internal/table...` which was never used and was a dudplicate of the normal table routes
- Cleaned up refetching of data whenever columns or rows are edited
- Centralised some core lucene logic in the builder, but not the actual table searching logic yet
- Fixed a large amount of bugs with the data UI section URL parameters, state selection and URL/state sync. Expanding datasources and clicking tables should just work now, and refreshing the page should always keep you on the same page, with the same selected item.
- Svelte runtime warnings have been supressed. This is configurable in `main.js` but I've supressed the worst offenders for now. This makes everything feel much faster to me, without the browser being clogged by hundreds (or thousands) of warnings.
- Added slight fade transitions and delays to tables loading in and out in order to hide flashes and jitters that were caused by loading new data and schemas. The result is that it looks perfectly smooth now when changing table.


shogunpurple (approved):
This is super nice - makes the backend UI so much better and more performant. Good job 🚀  We will consolidate the data fetching logic between the client and builder fairly soon  👍

PR #2801, opened 2021-09-29
Respect tenant in url in UI app. Reject tenant and session mismatch

Rory-Powell:
## Description
- Force logout users who try to access a tenant that does not belong to their session
- Set the UI app tenant id from the URL
   - Currently used `default` when there was no session, breaking sso-non redirect flow

### DIFF
--- a/packages/builder/src/pages/builder/_layout.svelte
+++ b/packages/builder/src/pages/builder/_layout.svelte
@@ -9,10 +9,31 @@
   $: hasAdminUser = $admin?.checklist?.adminUser?.checked
   $: tenantSet = $auth.tenantSet
   $: cloud = $admin.cloud
+  $: user = $auth.user
+
+  const validateTenantId = async () => {
+    // set the tenant from the url in the cloud
+    const tenantId = window.location.host.split(".")[0]
+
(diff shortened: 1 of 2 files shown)


packages/builder/src/pages/builder/_layout.svelte
```diff
@@ -9,10 +9,31 @@
...
   $: cloud = $admin.cloud
+  $: user = $auth.user
+
+  const validateTenantId = async () => {
+    // set the tenant from the url in the cloud
+    const tenantId = window.location.host.split(".")[0]
+
+    if (!tenantId.includes("localhost:")) {
+      // user doesn't have permission to access this tenant - kick them out
+      if (user && user.tenantId && user.tenantId !== tenantId) {
```
shogunpurple, line 20: NAB can be `user?.tenantId !== tenantId`

shogunpurple (approved):
LGTM!

packages/builder/src/pages/builder/_layout.svelte
Rory-Powell: Added in https://github.com/Budibase/budibase/pull/2801/commits/8de6cd35850973505a587733757a76d31b948e29

PR #2804, opened 2021-09-29
fix firefox redirect loop

shogunpurple:
## Description
Prevent a problem in firefox where a redirect loop was occurring. This was happening due to `onMount` being called constantly, which caused a ton of redirect calls. This seems to be an issue specifically with routify and firefox. Moved to an inline `<script/>` which will give us the same effect, but solves this problem.

Also updated the self host URL to match the account portal install link.

## Screenshots

### DIFF
--- a/packages/builder/src/components/upgrade/UpgradeModal.svelte
+++ b/packages/builder/src/components/upgrade/UpgradeModal.svelte
@@ -4,7 +4,7 @@
   let upgradeModal
 
   const onConfirm = () => {
-    window.open("https://accounts.budibase.com/install", "_blank")
+    window.open("https://account.budibase.app/install", "_blank")
   }
 </script>
 
--- a/packages/builder/src/components/upgrade/UpgradeModal.svelte
+++ b/packages/builder/src/components/upgrade/UpgradeModal.svelte
@@ -4,7 +4,7 @@
   let upgradeModal
 
   const onConfirm = () => {
-    window.open("https://account.budibase.app/install", "_blank")
+    window.open("https://account.budibase.app/portal/install", "_blank")
   }
 </script>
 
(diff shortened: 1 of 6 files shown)


packages/builder/src/components/upgrade/UpgradeModal.svelte
```diff
@@ -4,7 +4,7 @@
   let upgradeModal
 
   const onConfirm = () => {
-    window.open("https://accounts.budibase.com/install", "_blank")
+    window.open("https://account.budibase.app/install", "_blank")
```
Rory-Powell, line 7:
```suggestion
    window.open("https://account.budibase.app/portal/install", "_blank")
```

Rory-Powell (approved):

aptkingston:
This is really strange - I have no idea how an `onMount` can be called more often than an inline statement. onMount is only fired once per instantiation and surely the inline statement has to be fired *at least* once per instantiation? Unless svelte is doing something funky due to the store subscriptions `$` being used, and invalidating and calling the `onMount` handler again when required? That would be strange behaviour too, but swapping the `$` for manual `get`'s might fix it.

I've never really been sure of the actual behaviour of inline code like that.

shogunpurple:
@aptkingston it's super weird. I'm still not sure about the root cause to be honest, but what seemed to happen was that there were several `redirect` calls happening in `onMount` of certain pages. Ie -> `/builder` to `/builder/auth` etc. This actually seemed to affect it, because it was only the second redirect that got stuck in the loop.

My understanding of inline vs `onMount` is basically running script inline on the page vs waiting for the DOM content to be there for operations that require it. Thankfully in this case we don't actually need that page to be mounted, because we already have the state we need from the store initialisations and can just do the redirect straight away. I generally use `onMount` for almost everything because the difference is negligible

PR #2815, opened 2021-09-30
Cloud export -> self host import

mike12345567:
## Description
This feature introduces the capability to export all apps (and users) from the Cloud and import it to a self hosted installation in a single click. It removes all tenancy information from the users during the import phase, creating a new environment which can be directly logged into.

An export apps button has been added to the builder portal and an import apps from cloud button has been added to the admin creation screen.

The DBs that we export are the users tenant `global-db` and then all of their *development* apps - this is key because otherwise we would have ended up with duplicates of each published app. This method does mean that when the user logs into their self hosted environment none of their apps will be published yet, but I believe this is probably the correct behaviour anyway, they may not have correctly secured their environment yet so it is probably best to leave this step up to the user. The development app is exported instead of the published one as it will contain all the published data (due to the replication between prod -> dev DB) and it will have the most up to date version of the apps development.

The above point is open for discussion, we could possibly mark certain app exports as "published" and then after importing them in the self hosted environment we could publish them straight away.

Once we have all of the exported app DBs and the global DB we build a JSON object structure with each of the DBs denoted like: `{ "global-db": "db-export-string", "appName": "app-export-string" }` - the app names are used as keys as the import functionality requires a new name, this is the easiest way to access the old name. Once the databases are imported we iterate through the users and remove the tenancy information.

I have also updated the view system to have di-directional migration, meaning that it can migrate views to the cloud and it can migrate views to real CouchDB views in self host.

As part of this work I also carried out two little side tasks that I needed to test/get everything working:
1. I re-worked some of the pre-auth routing in the builder, between the `builder/src/pages/builder/index.svelte` and `builder/src/pages/builder/_layout.svelte` there was some duplication of routing (like two different ways to end up at login if no user found) which kept causing me to get into loops pre-auth and crashing the browser. I found the easiest way to fix this was to re-work the routing to make sure that it couldn't redirect to multiple pages at the same time risking a loop scenario.
2. I added an environment variable to the server/worker `DISABLE_ACCOUNT_PORTAL` which simply allows setting the server/worker into a Cloud environment but disables the account portal access which makes it impossible to login/use without an account portal and AWS access - this simplified testing quite a bit.

## Screenshots
Export apps button on portal:
image

Import apps button on admin creation page:
image


packages/builder/src/pages/builder/index.svelte
```diff
@@ -5,11 +5,9 @@
   auth.checkQueryString()
 
   $: {
-    if (!$auth.user) {
-      $redirect(`./auth`)
-    } else if ($auth.user.builder?.global) {
+    if ($auth.user?.builder?.global) {
```
Rory-Powell, line 8: was the `(!$auth.user)` redundant here?

Rory-Powell: specifically around the redirect
```
$redirect(`./auth`)
```

is that taken care of somewhere else now?

Rory-Powell: re-read the description, all good here 👍

Rory-Powell (approved):

packages/builder/src/pages/builder/index.svelte
mike12345567: Yeah, there was already somewhere that it caught the login, with much better rules and checks around it to stop loops, specifically:
```js
    // Redirect to log in at any time if the user isn't authenticated
    else if (
      loaded &&
      (hasAdminUser || cloud) &&
      !$auth.user &&
      !$isActive("./auth") &&
      !$isActive("./invite") &&
      !$isActive("./admin")
    ) {
      const returnUrl = encodeURIComponent(window.location.pathname)
      $redirect("./auth?", { returnUrl })
    }
```
in the `pages/builder/_layout.svelte`

shogunpurple (approved):
LGTM! This is probably one of my favourite features in a while, being able to migrate from self-host to cloud and even back is awesome. 

The migration of views is a good one as well, super nice to have that option before implementing the "real" view solution going forward!

PR #2816, opened 2021-09-30
Remove global info users on user deletion

Rory-Powell:
## Description
- Fixing a bug where the `global_info` representation of a user doesn't get deleted when a user is deleted
- This was preventing the user from being created again now that extra checks are in place


shogunpurple (approved):
LGTM

Issue #2696, opened 2021-09-21
MongoDB Read requests seem to be broken


mjashanks:
I get an error when I try run a MongoDb `find` : 

![image](https://user-images.githubusercontent.com/3524181/134178790-093aff77-cc4d-4d19-9d6c-d33bc4f00636.png)

Looking at the source, I think that we are missing an `await` call on all of these requests: https://github.com/Budibase/budibase/blob/62307b0b55f731d8dade780cd7acd0d07d8d03ce/packages/server/src/integrations/mongodb.ts#L115

The Mongo/Nodejs docs show that a `cursor` is returned. Then `await cursor.forEach(..)` must be called. https://docs.mongodb.com/drivers/node/current/usage-examples/find/

mslourens:
The cursor returned by find is an object with a method called `toArray`, which converts the result cursor into an array of results: [https://docs.mongodb.com/manual/reference/method/cursor.toArray/](https://docs.mongodb.com/manual/reference/method/cursor.toArray/). This method returns a promise, which in turn is returned by the Mongo integration. 
Somehow your connection got closed before the data was read.I am not sure why, because I just created a MongoDb connection and was able to run a read request from the query page of the Mongodb datasource.

mjashanks:
I pushed this to a branch: https://github.com/Budibase/budibase/commit/3c2b21abecca0d25288cdd0def503312ceda059b

Not sure how it's broken for me, but not for you. It does seem strange that it has not been raised before.

Will look into the other commands as well.

mjashanks:
Could it just be that your result returns before `await this.client.close()` - its in the `finally` statement, so always gets run directly after the return.

My DB is in Mongo Atlas, so there is plenty of latency there to be delayed.

mslourens:
that's probably why you get that message, it closes the connection before the data is loaded from Atlas. I run a local MongoDb instance, so that's much faster. I thought the promise from the returned query is always resolved in order to get the results, but apparently the finally is run first. 
At least this will fix your issue and any other timing issues.

spmurrayzzz:
Just to add some additional context here for future posterity (as I ran into the same problem). The issue is not around timing at all, I can repro this 100% of the time using local mongo with large datasets. Its just an async issue.

[...]

PR #2823, opened 2021-09-30
Fix for #2696 - adding awaits to mongodb connector

mjashanks:
See #2696 

There was a race condition happening, between a promise resolution and a `finally` statement. If we connect to a localhost MongoDB, we do not see this issue as the promise resolves before the `finally` is executed.

I was using Mongo Atlas, thus saw some latency. Thus, the mongo connector never worked me.


shogunpurple:
Good catch - odd that this ever worked?

shogunpurple (approved):

PR #2824, opened 2021-09-30
end user ping for client apps

shogunpurple:
## Description
This PR introduces simple analytics for end-users. It means we can at least track in the cloud how many active users we have across not just the builder but the apps themselves.

The analytics happen server side so that we aren't sending events directly from end users browsers.

## Screenshots


aptkingston (approved):
LGTM! Nice positioning of the ping too, so that it'll ping when loading a real app but not in the builder preview.

Issue #2796, opened 2021-09-29
Rows that come from enrich command do not have their formula fields calculated


mjashanks:
Rows that originate from `/enrich` do not show formula fields. 

Steps to repro
1. Create a table, with a "Created By" column
2. Add a formula column
3. Create a screen
4. Add Dataprovider, with datasource = `CurrentUser.YourTable-CreatedBy`
5. Add a table
RESULT - formula field is not populated

PR #2830, opened 2021-09-30
Fix for formula fields in enrich call

mike12345567:
## Description
Fix for #2796 - rows in the enrich call didn't have the correct table schema being passed to the `outputProcessing`, this groups the linked rows together with their correct tables for enrichment.

This also includes some fixes for the `dataBinding` readable -> runtime, runtime -> readable functions in the builder, there were some circumstances I found (specifically when building formula) where it could get into a loop as it tried to find all the instances to replace, as well as situations like `{{ sum hello hello2 }}` where you would get `{{ sum [hello] [hello]2] }}` after conversion to runtime, to fix this use a search string now, where each instance of bindings that have been replaced get removed from the search string, so that they can't be replaced again. We work from the longest binding replacements to the shortest to the best match is always found first.

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -1,12 +1,12 @@
 import { cloneDeep } from "lodash/fp"
 import { get } from "svelte/store"
 import {
+  findAllMatchingComponents,
   findComponent,
   findComponentPath,
-  findAllMatchingComponents,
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
The new binding replacement looks suitably complex for me to approve this 😆 I trust your logic! Nice one though, I understand why certain edge cases would have failed before.

PR #2840, opened 2021-10-01
remove unnecessary deployment poll

shogunpurple:
## Description
Remove deployment polling as it is no longer needed. It was here to maintain state between refreshes, but the deployment should not take a very long time, and we can rely on the single API call.

## Screenshots


aptkingston (approved):
Very glad to see this go 👏

PR #2877, opened 2021-10-04
Fix email link from internal bb onboarding. Prevent account holder from being deleted

Rory-Powell:
## Description
- Honour `PLATFORM_URL` for sending emails within budibase
- Audit other platform url usage, no changes needed only tidied a few lines in `WebhookDisplay`
- Copy changes in upgrade modal, remove mention of SSO being OS feature only
- Prevent account holder from being deleted within budibase


shogunpurple (approved):
LGTM

Issue #2835, opened 2021-10-01
Redirect to login on direct app request not authenticated


tobehn:
**Describe the bug**
As a user i want to directly access app via its short link _example.ie/app/<app-name>_. In case i am logged in beforehand this is absolutley possible. In the other case i just get a 404 Database is missing response. Whereas i would expect a redirect to the login.

**To Reproduce**
1. log in
2. access an application via your-domain.ie/app/<app-name>
3. access should be successful
4. log out
5. access the same app again 
6. you should get a 404 - Database missing

**Expected behavior**
In case not logged in redirect to login.

**Screenshots**
![image](https://user-images.githubusercontent.com/40893107/135583241-08a3d5e6-8e3a-4ace-bace-d1f1723849d8.png)

**Desktop (please complete the following information):**
 - OS: [ubuntu 20.04, iOS, macOS and other]
 - Browser [e.g. Firefox 92]
 - Version [e.g. 0.9.136]

**Additional context**
See a discussion on this topic #2428

PR #2880, opened 2021-10-04
Fix for shorthand app URL not re-directing if the user isn't logged in

mike12345567:
## Description
Fixing issue with direct app request in self host not working when un-authenticated (doesn't re-direct to login).

Fix for #2835


shogunpurple (approved):
LGTM! Nice fix

PR #2882, opened 2021-10-04
Deny only when usage quota is exceeded

Rory-Powell:
## Description
The quota exceeded notification was misleading; when creating a fourth app the message was
> You have exceeded your usage quota of 4 apps

Screenshot 2021-10-04 at 14 04 39

Change the logic to use `>` only, so that this message is seen when trying to create the fifth app
Screenshot 2021-10-04 at 14 05 28


shogunpurple (approved):
Good catch 👍

PR #2902, opened 2021-10-06
Some UX fixes for datasources and automations

PClmnt:
## Description
Fixes:

Datasources:
- Datasource config modal was persisting the previously entered password after completion

Automations:

- Changed size of delete icon and run test action button, also changed bg color on delete button 
- fixed the block background colour
- Added functionality to the plus icon, so now if you clock it you can add a block in between existing blocks, instead of having to delete and re-add in the correct order, it's animated so the block flys in from the right

## Screenshots
image


shogunpurple (approved):
LGTM, some nice updates here!

PR #2905, opened 2021-10-06
Don't redirect when tenant is missing, if unauthenticated

Rory-Powell:
## Description
- Previous fix for sso redirects broke the internal user onboarding flow
- Don't redirect when tenant is missing, if unauthenticated


shogunpurple (approved):
LGTM!

Issue #2788, opened 2021-09-29
Inconsistencies in "Define Filters", when no comparison value is set


mjashanks:
When using filters on a data provider, the various operators behave differently, depending on what data source is being used.

**Like**
On a custom query, when the comparison value is not set, all results are returned
On a SQL table (used postgres), when the comparison value is not set, no results are returned
On an internal table, we don't have the option of Like (which i think is fine)

**Equals**
On a custom query and SQL Table, when the comparison value is not set, no results are returned
On an internal table, when the comparison value is not set, all results are returned (filter not applied)

Note: By "Comparison Value" I mean "Volkswagen" in the statement "Car.Make = 'Volkswagen' "

I actually prefer "when the comparison value is not set, all results are returned" - i.e. when no options is selected from the dropdown, the condition is not applied. Otherwise, we have to start using conditions to setup filters, which gets messy, quickly.

Context: https://github.com/Budibase/budibase/discussions/2778

aptkingston:
Yea I think we should update the SQL to have the same results as lucene returns - which in this case would be returning all results if an equal expression does not have a value. Consistency is probably the most important thing like you say though, but there be differences in how the different databases themselves treat certain queries which might be outside of our control.

PR #2909, opened 2021-10-06
Various minor fixes

### LINKED ISSUE #2277: SQL unable to revert value to NULL
### Discussed in https://github.com/Budibase/budibase/discussions/2276

<sup>Originally posted by **hardc0l2e** August  5, 2021</sup>
in SQL (PostgreSQL/MySQL) If the row already have an existing value (text, numbers, etc), then wishes to revert it back to NULL, it won't allow you and the system restore it old value back.

Hho1pOo4dM
7p4uhwVMtE

But when you use the Internal/CSV table it works well, you can easily revert/remove value back to NULL

dDX9GyiCMA

</div>

mike12345567:
## Description
Just a few fixes after playing around a little bit and testing:
1. Fixing an issue with a redirect loop in the builder if a user went to an app that is locked, or tried to access an app that wasn't locked when they first reached the app screen (happened to me in the cloud when I had been sitting on the main page for a while).
2. Fixing issue #2788 - makes client-side and SQL based queries consistent with that of the internal tables, if no value is provided to the filter then no filter is applied.
3. `Like` as an operator for SQL tables, previously this did nothing because we don't have a fuzzy SQL search, but this caused confusion (especially since like is an SQL operator) so now it operates the same as a contains filter (`ilike %search value%`)
4. Fixing an issue with data source creation, when creating any data source plus you would need to refresh the builder to be able to access any of the tables UI.
5. Fixes #2277 means that SQL row values can be reverted to null.

### DIFF
--- a/packages/auth/src/redis/index.js
+++ b/packages/auth/src/redis/index.js
@@ -191,6 +191,12 @@ class RedisWrapper {
     }
   }
 
+  async getTTL(key) {
+    const db = this._db
+    const prefixedKey = addDbPrefix(db, key)
+    return CLIENT.ttl(prefixedKey)
+  }
+
   async setExpiry(key, expirySeconds) {
     const db = this._db
     const prefixedKey = addDbPrefix(db, key)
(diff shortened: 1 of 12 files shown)


shogunpurple (approved):
Awesome, some good ones in here 👍

PR #2920, opened 2021-10-07
Fixing multi-tenancy public app access

mike12345567:
## Description
As we have discussed, this solves the issues with multi-tenancy and logged in users not being able to access other tenants public apps.

This is less aggressive about rejecting users from the wrong tenant when they try to access an app, but instead it revokes all of their RBAC permissions, meaning that they cannot access any API endpoints or retrieve any non-public screens.

For all intent and purposes the user is simply a public user and they get exactly the same experience as a public user.

This also makes sure that the current app cookie is deleted on logout and login, which is logically a bit cleaner.

### DIFF
--- a/packages/auth/src/db/constants.js
+++ b/packages/auth/src/db/constants.js
@@ -1,5 +1,20 @@
 exports.SEPARATOR = "_"
 
+const PRE_APP = "app"
+const PRE_DEV = "dev"
+
+exports.DocumentTypes = {
+  USER: "us",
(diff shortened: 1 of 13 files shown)


shogunpurple (approved):
LGTM! Nice clean fix for this 👍

Rory-Powell (approved):
LGTM - just one comment about adding the cookie removal to account portal. This should just work because all cookies share a domain between the two apps in prod

PR #2937, opened 2021-10-08
Fixing apps import from cloud (removing user export)

mike12345567:
## Description
Fixing some issues with cloud export/import, removing the ability to export and import your users as this was dangerous and didn't really work with passwords/SSO.

The main adjustments are that the export doesn't include any users details, the reason being two fold:
1. Users passwords are hashed with a secret key that obviously cannot be shared, password hashes can't be exported from the cloud and therefore we don't really have an easy way for users to get back into their accounts when they are imported, apart from going through a forgotten password flow, it just seems easier to not export the users and have the users be re-invited to the new self hosted platform.
2. This was even more awkward if SSO was enabled in the cloud as this isn't really exportable and would be confusing to users on the new platform that they have to use a forgotten password flow for their previously SSO configured app.

### DIFF
--- a/packages/builder/src/pages/builder/admin/_components/ImportAppsModal.svelte
+++ b/packages/builder/src/pages/builder/admin/_components/ImportAppsModal.svelte
@@ -1,6 +1,7 @@
 <script>
   import { notifications, ModalContent, Dropzone, Body } from "@budibase/bbui"
   import { post } from "builderStore/api"
+  import { admin } from "stores/portal"
 
   let submitting = false
 
@@ -20,8 +21,8 @@
       if (!importResp.ok) {
         throw new Error(importJson.message)
       }
-      // now reload to get to login
-      window.location.reload()
+      await admin.checkImportComplete()
+      notifications.success("Import complete, please finish registration!")
     } catch (error) {
       notifications.error(error)
       submitting = false
--- a/packages/builder/src/pages/builder/admin/index.svelte
+++ b/packages/builder/src/pages/builder/admin/index.svelte
@@ -15,6 +15,7 @@
   import PasswordRepeatInput from "components/common/users/PasswordRepeatInput.svelte"
   import ImportAppsModal from "./_components/ImportAppsModal.svelte"
   import Logo from "assets/bb-emblem.svg"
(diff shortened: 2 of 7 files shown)


Rory-Powell (approved):
Nice straight forward solution to this problem 👍

Rory-Powell (commented):
The main thing that I'm concerned about; the deletion of users that may be linked to other entities in the system. I'm not sure if it's an issue right now, but I see it becoming a problem if we ever maintain that relationship. Maybe groups will run into it, in which case the group would need to be deleted also and so on. There might also be tenants where only the admin user has a password and everyone else is sso and we are removing data needlessly. 

A potential future change may be to mark all users as inactive and provide a flag to `forceAdminReset` which starts the admin init screen, any admin email / password can be supplied and that users password is reset, skipping provisioning. It would then be up to the admin to take care of resetting passwords / sending emails for the other users

The approach now is solid, just a thought that could save some headaches for larger installations when we get there

mike12345567:
> The main thing that I'm concerned about; the deletion of users that may be linked to other entities in the system. I'm not sure if it's an issue right now, but I see it becoming a problem if we ever maintain that relationship. Maybe groups will run into it, in which case the group would need to be deleted also and so on. There might also be tenants where only the admin user has a password and everyone else is sso and we are removing data needlessly.
> 
> A potential future change may be to mark all users as inactive and provide a flag to `forceAdminReset` which starts the admin init screen, any admin email / password can be supplied and that users password is reset, skipping provisioning. It would then be up to the admin to take care of resetting passwords / sending emails for the other users
> 
> The approach now is solid, just a thought that could save some headaches for larger installations when we get there

In terms of relationships in apps, I considered this, but the relationship simply doesn't exist, this is something that our link system manages, if a relation isn't present then it won't be returned from the API, the next time the row is updated the link docs will be cleared up.

However you're right, exporting users in the future would be a good idea, but we'd need to do a bit of work around building some UI for getting the email setup in self host and then sending out an email to everyone to set their new password/setup their SSO again (or if no email, export some sort of file full of passwords for each user to be passed out by the admin). However as you say I think this is something to tackle for larger deployments.

PR #2960, opened 2021-10-11
fix issue where automation was not being added in the correct order

PClmnt:
## Description
The blockIdx wasn't being passed to the addBlock function. Not sure how I missed this in testing, think i might have popped a previous stash before I merged. 

But this fixes it - I tested:

- Adding an action using the plus icon between two blocks
- Adding an action using the button between two block
- Adding an action to the end using the plus icon
- Adding an action to the end using the button

Also fixes an issue with where the 'Add Action' wouldn't be enabled if there was a following block, but this was inconsistent with how the 'plus' icon worked as we can add blocks at any stage now. it is now enabled as long as long as the inputs have been completed.
## Screenshots

https://user-images.githubusercontent.com/5665926/136774544-b6ec92e5-0468-4986-9e57-37fd3d47357a.mp4

### DIFF
--- a/packages/builder/cypress/support/commands.js
+++ b/packages/builder/cypress/support/commands.js
@@ -5,7 +5,7 @@
 // ***********************************************
 //
 
-Cypress.on('uncaught:exception', (err, runnable) => {
+Cypress.on("uncaught:exception", () => {
   return false
 })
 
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM!

aptkingston (approved):
LGTM!

Issue #2503, opened 2021-08-31
"No Access" Role / Public by default


shogunpurple:
**COS**
- Users are assigned a PUBLIC role by default
- Don't  actually show the PUBLIC role badge, just show nothing to indicate the user has no role for an app
- Account for non-home public routes

KernOil:
Another great feature that would be awesome! keep up the good work!

Rory-Powell:
Potential mock ups of how this might look
![](https://cdn.discordapp.com/attachments/735083509202419778/867785747159711744/Screenshot_2021-07-22_at_16.10.05.png)

Discussed here
https://discord.com/channels/733030666647765003/735083509202419778/867772469061353472

PR #2975, opened 2021-10-12
No access role / public by default

Rory-Powell:
## Description

https://github.com/Budibase/budibase/issues/2503

- Make app access public by default
- Split app roles view into two tables
   - Apps with role assigned
   - Apps with no role assigned
- Filter apps list in portal
   - Only show apps where user has a role
- Re-add the Public role option in update role modal

One difference in this approach from the original discussion, originally we said we could infer if an app is public by scanning all of it's routes and always showing it in the app list if so. This implementation takes a more explicit stance and requires the user to be assigned the public role for the app to appear. This allows for public apps to be hidden from logged in user portals if desired, and also allows for a logged in user to be assigned the public role for an app if desired.
 
Ticket: https://app.zenhub.com/workspaces/budibase-product-5fd773eeaec59f001117ad65/issues/budibase/budibase/2503

## Screenshots

**Video**

https://user-images.githubusercontent.com/8755148/136928153-eba62e09-2b27-443a-8198-30d73a87f5e7.mp4

**All apps**
Screenshot 2021-10-11 at 22 04 11

**No roles assigned**
Screenshot 2021-10-11 at 22 04 24

**Some roles assigned**
Screenshot 2021-10-11 at 22 08 03

**User granted development access**
Screenshot 2021-10-11 at 22 17 30

**Role selection**
Screenshot 2021-10-11 at 22 13 53


shogunpurple (approved):
Great job 👍 

It seems like this is migration friendly as well because it removes the current role if they have one

Rory-Powell:
> would love to see this on a user level. right now all of my users can see all of my published apps. would love to see a no access for users on an app by app user by user basis

@KernOil I might have misunderstood your request, but this implemention should give you exactly that! The screenshots above are from editing an individual user, so you can configure per app permissions, per user.

PR #2984, opened 2021-10-12
Fixes for automation issues

PClmnt:
## Description
This fixes a number of issues with automations including:

- Within the test data modal, we now add the schema to the json immediately so the users know what they're working with
- Fixed bug with app actions not displaying or handled correctly 
- Fixed webhooks - this is handled quite specifically due to the schema that it comes back with, it requires a `body` but returns a `body` then whatever the user has POST'ed as part of the schema, so we just add those values to the body in the json view. 
- Allow the user to select relationships within an automation block / test data
- Added _id and _rev to the autocolumns, this was handled solely at the front-end due to concerns with how migrations would be handled etc if we added them for real on the back-end, it's quite hacky due to the fact the schema of the auto-columns gets re-populated with every tables fetch, so we have to specifically check for `_id` and `_rev` and then render a copy button which copies the value to your clipboard

## Screenshots
image

image

image

image


shogunpurple (approved):
LGTM

aptkingston (approved):
LGTM!

PR #2992, opened 2021-10-13
Query transformers

mike12345567:
## Description
Note: this is being merged to @aptkingston `js-binding-drawer` branch as it originally based on some of his work towards the new JS code editor.

This introduces the ability to add a JavaScript transformer function to a query, which is carried out on the backend after the data has been retrieved. I've also written some documentation around a basic example of using this, which should help users get up and running with their transform functions.

Documentation can be found [here](https://app.gitbook.com/o/-LrDgq6yzTKOIMH5iQRQ/s/-LoJeodOP1kHMdukJ-o3/c/DI0WO9Onu9NbLhxHijMW/building-apps/data/transformers) for Gitbook users.

There isn't too much to say about this, it does what it says on the tin, the functionality is roughly based on what we used for the JS script block in automations, however I've re-worked it a little bit to use the `vm2` module instead of `vm` itself. I discovered a few issues with `vm` that allowed the user to get access to `process.X` information, which is properly blocked and secured by vm2.

I've also added a small help icon which can be seen in the screen shots below which takes the user directly to the transformer documentation. The editor is also re-sizeable vertically as I found myself often wanting to expand it for my more complex use cases.

## Screenshots
The initial state of the transformer code editor:
image

Adding a complex function:
image

Building a basic app with the output:
image


joebudi:
This is one of those changes that takes us from no code to low code, and will certainly be a crowd pleaser. Great work @mike12345567 ♥️

Rory-Powell (approved):
So cool. Great work on this!

shogunpurple (approved):
This is **the** feature. So excited to start using this. Great work! 🚀

aptkingston (approved):
Awesome stuff! I would actually remove the `Transformer` label above the editor though, since to me it looks unnecesary and just a duplication of the title above. Think it's fine to leave the `label` prop in the code editor, but I think the label in this usage isn't needed. Super work 🚀

Issue #2089, opened 2021-07-25
JS Scripting In Binding Drawer


shogunpurple:
When I am interacting with the binding drawer, I want to be able to write JavaScript inside the binding drawer with all the available context bindings, so that I can introduce logic and dynamic functionality to my apps

**COS**
- JS bindings in design panel UI - the code needs to be executed client side
- Code in the expression will be evaluated at the same time that HBS expressions are

mike12345567:
bump

PR #3002, opened 2021-10-14
JavaScript bindings

aptkingston:
## Description
This PR adds the ability to use JavaScript for component settings, with support for data bindings. Included as well is a new JS code editor, full support for safely executing JS in both Node and browser environments and a couple of bug fixes.

This PR also includes @mike12345567's recent PR for query transformers.

Addresses #2089.

## JS code editor
A new generic editor based on CodeMirror has been added, which include support for automatic bracket insertion, bracket highlighting, hint insertion and some other IDE-feeling features.

image

Pressing `ctrl+space` opens the hint menu, which is used for data bindings when in the binding drawer but can be easily used for any other hints.

image

## Bindings
Components which accept data bindings via the binding drawer now have the option to use JS instead of Handlebars. As shown in the screenshot above, this can be toggled using tabs. JS bindings use a jQuery style syntax, using the `$` symbol as a function, with the full readable text as the parameter. This keeps bindings looking clean and consistent with HBS.

HBS: `{{ Current User.name }}`
JS: `$("Current User.name")`

Behaviour clarification:
- When re-opening the drawer, the correct tab will automatically be selected, depending on whether the value is a HBS or JS value.
- If a JS expression is used, the setting will show the text `(JavaScript function)` in the right panel, to signify that it is JS code and not a raw text or HBS value.
image
- Bindings on the left of the drawer can be clicked regardless of whether you are using JS or HBS, and will insert the correct syntax.
image
image
- You can safely edit content in both tabs, and toggling between the tabs will not remove the value in either - until you press "Save"
- When you press "Save" in the drawer, whatever tab is currently selected will be saved as the value, and the content of the other tab will be lost
- JavaScript expressions must return a value. Writing `"foo"` will do nothing - you need to write `return "foo"`. There is a helper message to explain this below the editor.
- Functions that return `undefined`, `null`, `NaN` or `""` will be treated as if there is no value and be empty
- If the function throws an error while evaluating, this will be displayed to the user
image
- There is a 1 second timeout on JS code execution
- Async code is not supported
- HBS helpers are not available to insert when writing JS
- You can technically use JS expressions in HBS if you read the implementation details below and use the base64 encoded JS, but please don't do this...
image

## Implementation details
JS support has been added as a Handlebars helper, and works in Node and the browser. This means that any Budibase module can easily support JS by just using string templates with the same API that we have today.

The JS helper is defined by the `js` prefix, followed by the base64 encoded JS code to execute in quotes.
For example, let's say we want to run `return 1 + 2`:
- Base 64 encode that to `cmV0dXJuIDEgKyAy`.
- Then wrap it in a HBS expression to get `{{ js "cmV0dXJuIDEgKyAy" }}`

The base64 encoding and quote wrapping is required to allow the statement to be valid Handlebars.

## Security and bundling
Safely sandboxing JS code is tricky, and depends on the platform. [vm2](https://github.com/patriksimek/vm2) is the best choice for safely sandboxing JS in Node, but does not support the browser. There are fewer safety concerns when running in the browser anyway, as the code is being executed on the user's own machine.

The `string-templates` package now has 2 discrete entrypoints: one for CJS and one for ES6. We actually already had part of this in place so that we could bundle `string-templates` with rollup as an ES6 bundle for the client and builder. This has just been extended, because importing vm2 in any file will result in rollup being unable to bundle the code at all - so now the separate entrypoints import and configure the JS execution strategy for their respective environments.

- When running in Node a `vm2` sandbox is used with no external access
- When running in the browser, a `vm` polyfill is used with certain browser API's restricted. We have total control over this, but for now I've only blocked `alert`, `setInterval` and `setTimeout`. The `vm` polyfill already executes the code inside a iframe sandbox so accessing the `window` or `location` globals is not vulnerable. I haven't performed a full pen-test on this though, so I can't say that it's not possible to escape the sandbox

## Misc bug fixes
- Fix button roundness for buttons other than medium. Closes #2996 
- Fix potential issue with table pagination. Closes #2989 
- Fix tabs not showing the selected tab when being rendered inside a svelte transition


mike12345567 (approved):
> You can technically use JS expressions in HBS if you read the implementation details below and use the base64 encoded JS, but please don't do this...

this made me chuckle, such a beautiful JS function 🤣 

This looks great, really like the CJS/ES6 method of deciding which module to use for security purposes, should be easy enough for us to implement some server-side JS bindings when we need them! This introduces so much power to the design phase, you can do so much with this now! 

Lastly all I can say is `{{ js cmV0dXJuICJHcmVhdCBqb2IgQ2hlZWtzISI= }}`

shogunpurple (approved):
Amazing work! Also have to say I'm a fan of the jQuery!  Pure nostagia 😂

Actually surprised how little code was required, given the complexity of this feature!

Rory-Powell (approved):
Great work as always, two amazing features incoming 💥

PR #3010, opened 2021-10-14
Handle longform types within Automations

PClmnt:
## Description
We've never handled a longform type correctly within the automations, so this just displays a TextArea for that specific case

## Screenshots

image


aptkingston (approved):
LGTM

PR #3014, opened 2021-10-14
Don't show the create datasource popup in dev environments

PClmnt:
## Description
Turns off the create datasource modal **_in dev_** so that A) you don't have to see it every time you set up a new app and B) it was slowing down the automation tests quite a lot.


shogunpurple (approved):
Nice idea, LGTM!

Issue #2981, opened 2021-10-12
Options Picker validation issues


Rory-Powell:
**Describe the bug**
Seeing some inconsistency around the validation performed on the options picker specifically when using sources other than schema. 

**To Reproduce**

- Create an options field
- Don't specify schema options
- Change the `Options Source` to `Custom`
- Add custom options
- Select a custom option
- See output validation message
![Screenshot 2021-10-12 at 16 01 44](https://user-images.githubusercontent.com/8755148/136982195-16491256-12b2-463f-b589-a6d193ecb59f.png)
- Change the `Options Source` to a `Data Provider`
- Select a data provider option
- See output validation message
![Screenshot 2021-10-12 at 16 02 02](https://user-images.githubusercontent.com/8755148/136982299-49e014ac-109d-462f-840d-41389a3e1c68.png)
- Add some schema options
![Screenshot 2021-10-12 at 16 12 22](https://user-images.githubusercontent.com/8755148/136982637-971bc073-546b-47da-84f4-f41d124b498b.png)
- Change the `Options Source` to `Schema`
- Only now can a value be chosen
![Screenshot 2021-10-12 at 16 01 29](https://user-images.githubusercontent.com/8755148/136982809-08634a63-f35d-450b-a834-e4da8fe09463.png)
- Visit the `Configure Validation` panel
- Notice that no schema validation rules are shown
![Screenshot 2021-10-12 at 16 14 33](https://user-images.githubusercontent.com/8755148/136983006-c8d48b61-c7ad-4946-ac7d-32e32421eb95.png)
- Add a validation rule to match a custom option
![Screenshot 2021-10-12 at 16 15 42](https://user-images.githubusercontent.com/8755148/136983181-6e9ba371-63a6-4ef7-844a-f9346e2e9bac.png)
- Switch back to `Custom` for `Options Source`
- Select the matching option
- Validation does not take effect
![Screenshot 2021-10-12 at 16 16 46](https://user-images.githubusercontent.com/8755148/136983371-9c863e2c-c6a0-4107-b412-6a2f5fcbef32.png)
- Switch back to `Schema` for `Options Source`
- Select an option
- Validation takes effect
![Screenshot 2021-10-12 at 16 17 38](https://user-images.githubusercontent.com/8755148/136983517-089d67f2-8491-4862-8b22-fcb1e1014acf.png)

It appears that validation rules only work correctly when the `schema` source is chosen

**Expected behavior**
[...]

Issue #2674, opened 2021-09-20
Cannot use an options picker on a text column (SQL or internal table)


mjashanks:
I have a postgres `text` column. I would like to make this into an options picker, on the create form.

So, I
1. Remove the default text input
2. Add an options picker
3. Try to choose the field ...
**the field is not listed in the dropdown** 

4. Manually type the field name
**I get an error on the component preview:**
![image](https://user-images.githubusercontent.com/3524181/134018891-a3e81827-df7f-4161-81fb-8854fb4dc9c7.png)

The workaround is to call the field something else, e.g. `Category_picker`, and then assign that to the category column in the `Save Row` action.

mjashanks:
PS - this is also an issue with internal tables. 

If I have a text column, but I want to use an options picker to set it. Use case for this: I want "Options", on an internal table, where the options are loaded from an API, or another Internal Lookups Table.

PR #3015, opened 2021-10-14
Allow options picker to be used on text columns

Rory-Powell:
## Description

Allow the Options Picker form component to be used on text fields as well as the built-in options column type. 

### Options on a text picker
https://github.com/Budibase/budibase/issues/2674

#### Previously
- Postgres `product` table
Screenshot 2021-10-14 at 16 58 00
- We want to set `type` via a picker
- Generate a form 
- Add the options picker component
- The `Field` dropdown is empty
Screenshot 2021-10-14 at 16 59 07
- Set the field manually
- An error message is displayed
Screenshot 2021-10-14 at 17 00 13
- A workaround using the on save action is required

#### Now
- The `Field` dropdown is populated with text fields
- Custom values can be added and saved to text fields
Screenshot 2021-10-14 at 17 04 49

### Validation on non-schema options sources
https://github.com/Budibase/budibase/issues/2981

#### Previously
- Validation would always fail when a `Custom` or `Data provider` options source is chosen
- Unless the options matched what was defined in the options schema, or would always fail when schema is empty
- Validation is no longer coupled to the options schema when using other options sources on text fields
- e.g. one custom string field and one schema options field acting independently:
Screenshot 2021-10-14 at 17 16 23
- e.g. Custom validation acting independently
Screenshot 2021-10-14 at 17 25 08
- e.g. Pulling values from a datasource passing validation
Screenshot 2021-10-14 at 17 20 26

### DIFF
--- a/.github/workflows/release-develop.yml
+++ b/.github/workflows/release-develop.yml
@@ -9,7 +9,7 @@ env:
   POSTHOG_TOKEN: ${{ secrets.POSTHOG_TOKEN }}
   INTERCOM_TOKEN: ${{ secrets.INTERCOM_TOKEN }}
   POSTHOG_URL: ${{ secrets.POSTHOG_URL }}
-  SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+  SENTRY_DSN: ${{ secrets.SENTRY_DSN_SELF_HOST }}
       
 jobs:
   release:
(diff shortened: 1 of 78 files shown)


aptkingston (commented):
This will be a great feature to have and make pickers a lot more useful, particularly for SQL tables external datasources! Added a couple of questions about refactoring, but looks great. 👍

PR #3019, opened 2021-10-15
Fix performance of importing large CSVs

### LINKED ISSUE #1840: Large size CSV import generates 504 Gateway timeout error
Importing a CSV of 30000 entries and 3MB size to create a DB table generates a 504 gateway timeout error and no error is shown on the Budibase interface (I could check the error from the Network tab of the Chrome Dev Tools). The "Create" button just stays grayed out and no table is added.

 - OS: Linux Manjaro
 - Browser: Chrome 86.0
 - Budibase Version: 0.9.65

aptkingston:
## Description
This PR is an extremely small change but totally fixes performance issues when importing large CSVs. We were treating rows in CSV imports as if they may contain relationships and it was this checking which was making it very slow and timing out with big CSVs. CSV imports cannot contain relationships so we can safely remove this check.

Closes #1840.

### DIFF
--- a/packages/server/src/api/controllers/table/utils.js
+++ b/packages/server/src/api/controllers/table/utils.js
@@ -93,19 +93,10 @@ exports.handleDataImport = async (appId, user, table, dataImport) => {
         }
       }
 
-      // make sure link rows are up to date
-      finalData.push(
-        linkRows.updateLinks({
-          appId,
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
Nothing better than when it's a small change! great job

mike12345567 (approved):
Lovely, solves the problem perfectly!

PR #3025, opened 2021-10-15
Fixing refresh on wrong password and setting hard SQL limits

mike12345567:
## Description

This has been discussed a bit, @Rory-Powell found an issue with very large SQL datasets causing issues with the SQL plus querying system, making sure its impossible to cause the server to lock up by setting a hard limit to the number of rows which can be returned. This hard limit is set as part of the relationship addition as there was already a limit for the base rows, but no limit was set for the related rows that could be returned.

Right now this is a semi-basic fix for the SQL issue as it essentially lets the database decide which rows will be returned as part of the relationship (for very large data sets). It sets the initial limit first of how many to get from the parent table, and once this is met it simply fetches to the overall limit as many joined rows as it can. This isn't particularly great as if you could get a random number of joined rows back for each row.

I attempted to put a more elegant solution in place with the idea of sub-queries, performing a subquery for data from within the joined table up to a limit and then joining using that, but this proved very very complex and I don't believe the performance of it would be very good, so I feel its best to just leave a hard limit in place and for use cases that need to work with very large data sets, they will need to setup their own queries.

Just for awareness the more "elegant" solution might have looked something like this:
```js
      query = query.leftJoin(
        knex(toTable).select(fieldsToGet).where({[`${toTable}.${to}`]: `${fromTable}.${from}`}).limit(100).as(`${toTable}`),
        `${fromTable}.${from}`,
        `${toTable}.${to}`
      )
```
This would be done for every joined table and as you can imagine this would create a very complicated query, attempting to debug it would become nearly impossible. I don't think this is a path worth pursuing, unless someone has an issue with the more basic approach taken. 

Also fixed an issue where the 403 on wrong password/email would cause the builder to refresh, meaning the user couldn't see the issue/errors.

### DIFF
--- a/packages/builder/src/builderStore/api.js
+++ b/packages/builder/src/builderStore/api.js
@@ -15,7 +15,7 @@ const apiCall =
     if (resp.status === 403) {
       removeCookie(Cookies.Auth)
       // reload after removing cookie, go to login
-      if (!url.includes("self")) {
+      if (!url.includes("self") && !url.includes("login")) {
         location.reload()
       }
     }
--- a/packages/server/src/integrations/base/sql.ts
+++ b/packages/server/src/integrations/base/sql.ts
@@ -98,7 +98,9 @@ function addFilters(
 }
 
 function addRelationships(
+  knex: Knex,
   query: KnexQuery,
+  fields: string | string[],
   fromTable: string,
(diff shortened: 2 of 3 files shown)


shogunpurple (approved):
LGTM! That 403 fix is actually a big one as well, good job.

Issue #3031, opened 2021-10-17
Cannot read property 'hasOwnProperty' of undefined after changing column


juliangeiges:
**Describe the bug**
Having a Table with Rows. After Adding or Changing a column i'am not able to create or edit rows anymore.

Error log shows:

```
ERROR [1634488322087] (29 on f3babc9e46c2): Cannot read property 'hasOwnProperty' of undefined

    TypeError: Cannot read property 'hasOwnProperty' of undefined

        at Object.exports.coerce (/app/dist/utilities/rowProcessor/index.js:153:34)

        at exports.inputProcessing (/app/dist/utilities/rowProcessor/index.js:190:38)

        at Object.<anonymous> (/app/dist/api/controllers/row/internal.js:181:30)

        at Generator.next (<anonymous>)

        at fulfilled (/app/dist/api/controllers/row/internal.js:5:58)

        at runMicrotasks (<anonymous>)

        at processTicksAndRejections (internal/process/task_queues.js:97:5)

```

**To Reproduce**
1. Create a new table
2. create a column
3. create a row 
4. create a new column
5. try to update that row of 3. or create a new one

**Expected behavior**
Also after changing or adding columns i expect that i can insert rows or update existing ones

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - Version 0.9.166

**Additional context**
I tried several releases. It breaks starting with version 0.9.165.
On 0.9.164 it works normal.

PR #3032, opened 2021-10-17
Potential fix for issues with updating and saving rows and columns

PClmnt:
## Description
So this issue (https://github.com/Budibase/budibase/issues/3031) came out of this PR (https://github.com/Budibase/budibase/pull/2984/files) and the addition of the _id and _rev fields to autocolumns, I didn't realise, and obviously didn't fully understand the reprocussions of adding these, in that I didn't realise the columns would then actually get saved. Totally my bad and hold my hands up for it. 

The crux of the issue is essentially that if changes have been made to a document,  it means they are in a bad state, in that the `_id` field has become undefined.

This has led to some serious issues in which users can't save or edit columns and rows. However a fix was pushed to master tonight that stops this from happening, but obviously some users may have made some changes over the weekend. 

This PR seems to fix it (but that's with some brief testing and some very hacky code) but I have more testing to do and thought I would get this up so people could see what the issue was and get some input on potential fixes for users who have edited or made changes to their data. 

Outstanding issues I can see:

- other autocolumns aren't visible for the table that I am looking at (new tables are fine)
- saving an attachment causes the row to dissapear and not return in any further calls to the /search endpoint (but I am not sure if this related, as i reverted the changes from that PR and still seeing that issue

### DIFF
--- a/packages/bbui/src/Table/CellRenderer.svelte
+++ b/packages/bbui/src/Table/CellRenderer.svelte
@@ -5,7 +5,6 @@
   import RelationshipRenderer from "./RelationshipRenderer.svelte"
   import AttachmentRenderer from "./AttachmentRenderer.svelte"
   import ArrayRenderer from "./ArrayRenderer.svelte"
-  import InternalRenderer from "./InternalRenderer.svelte"
 
   export let row
   export let schema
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
Great job with this fix - I think the surface area of problems from the incident should be fairly small, but this will cover us in any case. LGTM!

aptkingston (approved):
Was part of the problem with this just that the real schema object was being mutated? If we just copied and passed a different schema to the table (just to get it to diplay the columns) but didn't mutate the real schema the way we currently do, then that should avoid the issue. Although the table is responsible for rendering the modals for creating and editing rows and modal, so whatever schema is passed to the table will ultimately end up being used for those.

Separating the table config from the actual data schema is probably the right way to go to implement this again once we sort this out. Nice job catching this though!

PR #3041, opened 2021-10-18
Fix/helm make multitenancy configurable

fabiankaestner:
## Description
This PR addresses the following issues with the helm chart:
1. Make helm chart self-hosted by default
2. Set a correct default value for the minio service
3. Expose multi-tenancy in the values.yaml

Also, apologies for some unneccessary diffs due to my autoformatter. If those are undesired I can update the PR.


shogunpurple (approved):
LGTM! Thanks @fabiankaestner 😃

Update: I’ll merge this one, since it supports multi tenancy as well.

Issue #1718, opened 2021-06-16
App Export Performance Issues


shogunpurple:
In the staging environment, it takes well over 30-40 seconds to export an app with around 8000 rows.

It does seem as though the `pouchdb-replication-stream` package suffers from some performance bottlenecks, preventing us from reaching the full throughput of what should be a fairly lightweight operation.

- Test locally with a large app
- Monitor how long export takes
- Identify performance bottlenecks and fix

sergey-dme:
Bad bot

PR #3042, opened 2021-10-18
Improve app export experience

aptkingston:
## Description
After looking in to #1718, there are a couple of issues with how we handle app exports, especially for large apps. One is that the dumping the app DB is fundamentally slow, and I don't believe we can speed that up. We use [PouchDB replication stream](https://github.com/pouchdb-community/pouchdb-replication-stream) which uses  [PouchDB dump CLI](https://github.com/pouchdb-community/pouchdb-dump-cli) to actually do the dump, and that seems like the only/best tool for the job already.

The other (and bigger, IMO) issue is that we wait for the backup to complete serverside before returning anything to the client. This causes timeouts because the default envoy config is a 15 second timeout on the proxy. For any big apps, 15 seconds is not enough for the server backup to complete the backup and therefore the request is terminated.

This PR introduces the following changes:
- Increase envoy timeout to 120s for `/api` routes. This matches the default Koa config.
- Refactor the app export functionality to add a new option for streaming app exports. This means we can now export an app to file, to a string, or as a stream which can be immediately sent to the client.
- App exports are now treated like file downloads, and handled by the browser. This means that when exporting an app, you will immediatly get a popup asking where you want to save the export. The download will then commence and the PouchDB dump will be streamed in real time, so you can see the progress of the download and get immediate feedback. Large exports can still take upwards of 30 seconds to complete, but because it is clearly an ongoing download, this feels totally fine. The only caveat is that there is no known file size (since we're streaming an unknown size app DB in real time), so it will say "unknown time remaining" on your download.
- As soon as a download is started, you can continue using the builder normally. You don't need to hang around and wait for the download (like we used to have to) since it's just a normal browser download that's being handled separately from your Budibase tab.
- Small bugfix for the app export filter. I believe we were incorrectly including user relationship data in our exports, when they should have been filtered out.
image

## Video
Here's a video of exporting a massive (20MB) app. You can see that it takes a while to stream the download, but since it looks like a normal download this is totally fine. You can see that afterwards I download a small app which completes instantly. I sit and wait for the download to complete to show the functionality, but I could have immediately opened an app or left the page - the download will still continue in the background like a normal download.

https://user-images.githubusercontent.com/9075550/137721611-2abd053e-675a-4ff8-9c29-a33c05fac9d7.mp4

## Future work
This same strategy is probably required for the "export all aps" functionality, since that is just a much larger version of this same feature. It could also potentially be used for table exports, since they could probably also be large. This PR, with the timeout increase, will allow for much bigger exports of both of those other use cases, but the downside is that you will need to sit on an unresponsive page for however long is required before you get the prompt to save your completed dump.

### DIFF
--- a/hosting/envoy.dev.yaml.hbs
+++ b/hosting/envoy.dev.yaml.hbs
@@ -41,6 +41,7 @@ static_resources:
               - match: { prefix: "/api/" }
                 route:
                   cluster: server-dev
+                  timeout: 120s
 
               - match: { prefix: "/app_" }
                 route:
--- a/hosting/envoy.yaml
+++ b/hosting/envoy.yaml
@@ -58,6 +58,7 @@ static_resources:
               - match: { prefix: "/api/" }
                 route:
                   cluster: app-service
+                  timeout: 120s
 
               - match: { prefix: "/worker/" }
                 route: 
--- a/packages/builder/src/pages/builder/portal/apps/index.svelte
+++ b/packages/builder/src/pages/builder/portal/apps/index.svelte
@@ -112,16 +112,8 @@
 
   const exportApp = app => {
     const id = app.deployed ? app.prodId : app.devId
-    try {
-      download(
-        `/api/backups/export?appId=${id}&appname=${encodeURIComponent(
-          app.name
-        )}`
-      )
-      notifications.success("App exported successfully")
-    } catch (err) {
-      notifications.error(`Error exporting app: ${err}`)
-    }
+    const appName = encodeURIComponent(app.name)
+    window.location = `/api/backups/export?appId=${id}&appname=${appName}`
   }
 
   const unpublishApp = app => {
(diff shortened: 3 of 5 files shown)


shogunpurple (approved):
Great work cheeks - this feels a lot more smooth now just having it as a regular download.

We'll need to push another envoy image with the updated config once this is merged :+1:

mike12345567 (approved):
LGTM - Love this, such a good improvement, realistically we could probably consider doing something similar for the tables export too, this totally covers the problem in a nice and simple way!

aptkingston:
@shogunpurple you happy if I merge this now? I guess we need to wait until we merge to master to push the envoy image rather than develop do we? Although it would be totally safe to push the image whenever since it's just adding a couple of lines actually. In the staging env with the old image this new code will still work fine, but any downloads which take longer than 15 secs will just get cut off and fail. But that's the restriction we have today anyway, so I think it's fine to merge this now and push the envoy image whenever.

PR #3046, opened 2021-10-18
Add back the ability to copy the ID and Revision (safely)

PClmnt:
## Description
This adds back the ability to copy the ID and Revision from an internal table. 

I took @aptkingston's suggestion and copied the table schema, instead of mutating it like last time...

Also fixes an issue where a Formula column couldn't be created due to the fact i'd added another type by accident. It needed to be added to the exclusions list as it could potentially be sitting in some users DBs.


aptkingston (commented):
Couple of questions but yea in principal the copy approach should be safe 👍

aptkingston (approved):
LGTM now! 👍

Issue #2417, opened 2021-08-19
URL parameter is not what I expect, when using SQL


mjashanks:
Copied from this comment: https://github.com/Budibase/budibase/discussions/2043#discussioncomment-1205027

I have a `Customers` table, in postgres, and corresponding `/customers` and `/customers/:id` screens.

The `View` link on my customer table reads `/customers/{{ customers Table.customers._id }}`. However, at runtime, this translates to `/customers/%5B10%5D`. So, the `{{ URL.id }}` parameter decodes as `[10]` (the id of the record is `10`).

This is inconsistent with internal tables, which do not wrap the id in brackets `[ ]`.

If I manually browse to `/customers/10` the page works just fine, so there doesn't seem to be a purpose for this.

There is a very sensible workaround, it's just confusing.

aptkingston:
This is by design. Since there's no guarantee that external SQL tables have an `_id` field (which we depend on), we enrich them with one automatically for rows in external tables. This field is an array because SQL rows may have composite keys, so the only way to uniquely identify them is to generate this `_id` field using all the attributes of the composite key. In the case where there is just a single primary key then there's just one entry. We need to do this in order to be able to unique identify the row when we want to update it.

What the key is shouldn't matter though, because the API should be able to retrieve the correct row using this ID. Was it not retrieving the row correctly when clicking the link and using that generated ID?

mjashanks:
It was retrieving it ok.

However, when tried to add a Note (child record) inline, it wouldn't work - I assume because it was trying to set the foreign key to `[10]` (see below).

The workaround is to use the id from the repeater, rather than the URL - so not an issue really. It just makes the usage of `{{ URL.id }}` inconsistent.

![image](https://user-images.githubusercontent.com/3524181/130065482-e79c39fc-0968-499f-bb39-caad7786a69d.png)

aptkingston:
Ah fair enough. I can't image it would be hard to update it so that any schema columns for foreign keys just extract the value inside the generated `_id` array if present, which would make that work. I can see people are bound to trip up on that. At the same time if people don't have an `_id` column in their table they may also be confused as to why they're setting a foreign key to a column that doesn't exist.

I also have no idea what would happen there if the primary key actuallty was a composite key, because that would require totally different SQL to be generated if you wanted to reference it as a foreign key.

sergey-dme:
Hey  @mjashanks just as idea - use Handlebars:
[...]

PR #3072, opened 2021-10-18
Handling URL.id usage with SQL tables

mike12345567:
## Description
As #2417 demonstrates the usage of `URL.id` with SQL tables could be a little complicated/confusing due to the way we build our `_id` field for SQL table rows. Due to the nature of primary keys and the fact they could be built out of a composite we had to handle the mixture of ways unique identifiers could be built. We handled this by aggregating the keys into an array, then serialising the array into a string. This technique did not fit well when using the `URL.id` field to input values directly to the SQL database (like using the ID for a foreign key field).

Due to the way we query SQL databases all queries are built from one central part of the code, which means we can manipulate the values being input. The configuration used to run a query is defined as follows (in our Typescript definitions):

```typescript
interface RunConfig {
  id?: string
  filters?: SearchFilters
  sort?: SortJson
  paginate?: PaginationJson
  row?: Row
}
```

The important parts that the `_id` field could have been used as an input for are `filters` and `row`. The sorting and pagination properties don't take a value, they are simply linked to previous query and table schema information. The ID field already handles the ID mechanism we have in place (as can be seen when querying a route like `/customers/10` or `/customers/[10]`).

The way I have implemented this is a clean up function which looks at the table schema, to see if any of the input fields could contain a key field (e.g. they are of type like integer or string which could be a foreign key/key field) and then looking for strings like `[10]`. If these are seen then the brackets are stripped and the raw value is used. There are two negatives to this approach:
1. Using a string like `[Hello there!]` as a raw row value in SQL will have the brackets removed, we perhaps could solve this if it is a serious issue by looking through our relationship configuration to see if the field we are trying to manipulate is a foreign key for any table, but then users have to fully setup their relationships for this fix to apply, a more all encompassing fix felt better here.
2. Currently it can't technically handle composite keys, as the user has input the `_id` field for a single row value, if the foreign key is actually a composite key then this won't work. Again if this is a massive issue we can use the table schema information + provided relationship info to work out where the `_id` components should be inserted, but this would require we support composite keys for relationships (currently unsupported).

## Screenshots
My test scenario, I created a child form on a detail page where you could create a related row to the parent using the `{{ URL.id }}` binding.
image
image
image

### DIFF
--- a/packages/server/src/api/controllers/row/ExternalRequest.ts
+++ b/packages/server/src/api/controllers/row/ExternalRequest.ts
@@ -15,8 +15,9 @@ import {
 import {
   breakRowIdField,
   generateRowIdField,
+  isRowId,
+  convertRowId,
 } from "../../../integrations/utils"
-import { RelationshipTypes } from "../../../constants"
 
 interface ManyRelationship {
   tableId?: string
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM, nice one

PR #3094, opened 2021-10-19
Fixing issue with SQL tables and automation update row action

mike12345567:
## Description
Fixing issue with SQL tables and automations updating a row, also making error handling better across automations to make sure some sort of error message is always returned.

Previously our update endpoint didn't specifically require a table ID to be specified, however for external tables we need to know this information up front to decide whether to use internal or not. Our automation action already knows the table ID anyway so just passing it down to the row controller.


Rory-Powell (approved):

shogunpurple (approved):
LGTM

PR #3104, opened 2021-10-21
Handle nulls / empty in views and tables

### LINKED ISSUE #2042: View filters do not handle nulls
First described in https://github.com/Budibase/budibase/discussions/2030.

When a column value is `null` (or possibly `undefined` ?) we cannot account for this in a view filter.

To reproduce a `null` value
- Add a column of any type, when you have existing data in the table. All existing values have blank values for this column.
- Create a new row via the builder and do not enter a value for Number or Date 

### Suggested solution

- Filters should have an option for `Is not set`
- When values are not set, we should explicitly say in the rows `Not Set` rather than having a blank cell, or unchecked box. This should probably apply to both the builder and the table component within the app
- We could also set a "Value when not set" option on the column so that it can work nicely with formulas and view calculations.
- For bonus points, it would be really nice to be able to run batch updates inside the builder. This has other uses, but very much a nice-to-have if the above points are covered

Rory-Powell:
## Description
Fixes https://github.com/Budibase/budibase/issues/2042

- Add support for `Is Set` and `Is Not Set` to view filters
   - Check for empty, null and undefined to be not set
- Add corresponding to logic to table rendering to show `Not Set` for matching columns

## Screenshots
Using the example in the bug ticket, it was difficult to filter out columns which has no value (created before a column was added) in this case a boolean. 

Dataset

Screenshot 2021-10-21 at 14 22 00

Configuring a `Not Set` filter
Screenshot 2021-10-21 at 14 22 33

Results
Screenshot 2021-10-21 at 14 23 10

## Example 2 

Show where number is set
Screenshot 2021-10-21 at 14 25 09

Screenshot 2021-10-21 at 14 24 44

vs not set

Screenshot 2021-10-21 at 14 25 35

Screenshot 2021-10-21 at 14 25 44

### Misc

Example json from a row with undefined values
```json
{
  "_id": "ro_ta_75345deb2cf749bb8d7323b647e03c27_8aa42f23be9f4029873701d69fa2a702",
  "_rev": "1-b34a2c3e6e5a6bf66cd98afce2b043d6",
  "tableId": "ta_75345deb2cf749bb8d7323b647e03c27",
  "Auto ID": 1,
  "type": "row"
}
```
Example json from a row with empty values
```json
{
  "_id": "ro_ta_9259c4b7baaa4995a6929e7cd454ccb5_4def4e997733437592065e6ea0437f00",
  "_rev": "1-1b165944ab358e6e57899a8bf2c470a3",
  "attach": [],
  "date": null,
  "boolean": false,
  "number": null,
  "multi": [],
  "options": null,
  "long_form_text": "",
  "text": "",
  "tableId": "ta_9259c4b7baaa4995a6929e7cd454ccb5",
  "Auto ID": 1,
  "Created At": "2021-10-20T13:24:46.931Z",
  "Updated At": "2021-10-20T13:24:46.931Z",
  "type": "row"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,18 +8,19 @@
 </h1>
 
 <h3 align="center">
-  Build, automate and self-host internal tools in minutes
+  The low code platform you'll enjoy using
 </h3>
 <p align="center">
-  Budibase is an open-source low-code platform, helping developers and IT professionals build, automate, and ship internal tools on their own infrastructure in minutes.
+  Budibase is an open source low-code platform, and the easiest way to build internal tools that improve productivity.
 </p>
 
 <h3 align="center">
  🤖 🎨 🚀
 </h3>
+<br>
 
 <p align="center">
-  <img alt="Budibase design ui" src="https://i.imgur.com/5BnXPsN.png">
+  <img alt="Budibase design ui" src="https://res.cloudinary.com/daog6scxm/image/upload/v1633524049/ui/design-ui-wide-mobile_gdaveq.jpg">
 </p>
 
 <p align="center">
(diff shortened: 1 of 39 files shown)


aptkingston (commented):
Awesome stuff Rory. 

My only question would be whether we should use the same terminology we already use for filtering data providers - `Is empty` and `Is not empty`. Here's a screenshot for reference:

image

Might just be good to keep things consistent. I also think that if using the word "empty" then it would probably be fine to leave empty values as empty, rather than rendering the "Not Set" placeholder text. I'll let @joebudi weigh in on that one though.

PR #3105, opened 2021-10-21
Disable action if SMTP has not been setup

PClmnt:
## Description
Adds a tooltip and disables the action button if SMTP has not been configured. 

I've added the tooltip at the bottom, I know the top is probably a little bit more intuitive but there's something really weird going on with the tooltip component and specifically how adding it to the top it interacts with the CSS for this. 

I can go back and spend some more time to try and figure it out if needed, but didn't want to get too bogged down on this so thought i'd get a PR up at least. 

## Screenshots
image


Rory-Powell (approved):
LGTM!

PR #3142, opened 2021-10-25
Scope view filters to table always

### LINKED ISSUE #3141: OR filter in view can query multiple tables
**Describe the bug**
The JS function produced by an OR query on a view can query multiple tables 

**To Reproduce**
- Create table1 with name column = value1
- Create table2 with name column = value2
- Create view for name = value1 or name = value2
- Notice results from both tables are brought back

**Additional context**
- If a non table document is matched the entire view will break

Rory-Powell:
## Description
Fixes https://github.com/Budibase/budibase/issues/3141


shogunpurple (approved):
LGTM, nice fix

PR #3145, opened 2021-10-25
Prevent non builder from accessing dev apps

### LINKED ISSUE #3095: Non developers can view app preview
**Describe the bug**
A non developer / builder may view the dev preview of an app by visiting the url directly. 

**To Reproduce**
- Log in as developer
- Create an app
- Assign a user access to the app
- Open dev preview in a new tab
- Log out of budibase
- Log in as other user
- Visit the dev preview tab
- User can view dev preview

**Expected behavior**
- Only developers should be able to see preview

Rory-Powell:
## Description
Fixes https://github.com/Budibase/budibase/issues/3095

- add check on app id to current app middleware
- refactor tests to create a production app and use that for non builder requests
- remove unused email / password from server tests
- remove some unused run configs
- remove info logging from jess run to reduce noise in test output


shogunpurple (approved):
LGTM!

PR #3155, opened 2021-10-26
Ignore case when finding user by email

### LINKED ISSUE #3110: OIDC SSO may mismatch existing email address due to case sensitivity
**Describe the bug**
When logging in using sso using the same email address as an existing user, case sensitivity should be ignored so that the user is not duplicated

**To Reproduce**
- create user USER@test.com
- configure sso
- log in using SSO with user@test.com
- notice a new user is created

**Expected behavior**
- mismatched casing should not result in a new user

Rory-Powell:
## Description
Fixes https://github.com/Budibase/budibase/issues/3110

The view function for getting a user by email used an exact match, update this to use a case insensitive match to fix a bug where sso emails had a different casing than the pre-created user. Should also protect against multiple account abuse. 

There was no clean way to achieve this in a backwards way, short of creating an entirely new view and leaving the old one to remain unused. Opted for adding in a small migration lib that can run a piece of code conditionally based on a value in the database. At this time the update view migration is ran on demand, in future we could look at running migrations at startup for every tenant.


shogunpurple (approved):
LGTM, although I think we should maybe have a test for the migration lib directly, just to ensure everything is working as expected. If it passes we have fairly high confidence it will work in prod too given that we use an actual couchDB integration in our tests.

mike12345567 (approved):
Looks good to me! Good idea around the migrations, I don't actually think there is a problem with running them on demand as needed, although when enough of them build up handling them from a central location might be worthwhile just so we can easily see all migrations that can be carried out!

PR #3159, opened 2021-10-26
allow opening a template directly from a URL

shogunpurple:
## Description
This PR contains code that allows users to select a template from an external website, and it will open straight into the Create App Modal in the builder if you are logged in.

### URL Format
`<budibase-url>/builder/portal/apps?template=app/lead-generation`

## Screenshots

https://user-images.githubusercontent.com/11256663/138887135-533d60ba-5ca9-46c9-93d1-96effea35913.mov

**Warning message on invalid template key**

Screenshot 2021-10-26 at 15 20 49

### DIFF
--- a/packages/builder/src/pages/builder/portal/apps/index.svelte
+++ b/packages/builder/src/pages/builder/portal/apps/index.svelte
@@ -201,9 +201,10 @@
     await apps.load()
     loaded = true
     // if the portal is loaded from an external URL with a template param
-    const templateKey = $auth.initTemplate
-    if (templateKey) {
-      createAppFromTemplateUrl(templateKey)
+    const initInfo = await auth.getInitInfo()
+    console.log(initInfo)
+    if (initInfo.init_template) {
+      createAppFromTemplateUrl(initInfo.init_template)
     }
   })
 </script>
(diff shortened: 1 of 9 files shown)


Rory-Powell (approved):
LGTM 👍 

One suggestion would be to say on the app modal `Using template: XYZ` somewhere, there's a little bit of lack of feedback once you go from website to BB that all has worked

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -184,9 +184,27 @@
...
+    }
+  }
+
   onMount(async () => {
     await apps.load()
     loaded = true
+    // if the portal is loaded from an external URL with a template param
+    const initInfo = await auth.getInitInfo()
+    if (initInfo.init_template) {
+      createAppFromTemplateUrl(initInfo.init_template)
```
Rory-Powell, line 206: I know we have the request already in the app modal, it might read better to put it here as well (instead of?)
`await auth.setInitInfo({})`
just so we know it's being cleared right after the init is done

Rory-Powell (approved):
LGTM 👍

joebudi:
Awesome work @shogunpurple

PR #3160, opened 2021-10-26
Add timeout to node mailer, lower than jest default

Rory-Powell:
## Description
- The node mailer has a default timeout of 2 minutes
- This was causing our flaky test condition to never been caught due to exceeding the jest default
- Set the node mailer to use a 1 second timeout and revert the 30 second jest exception to a default of 5 seconds


shogunpurple (approved):
Nice one!

mike12345567 (approved):
Nice catch, should solve it well!

Issue #2820, opened 2021-09-30
SQL connectors should fail gracefully when a table does not have primary key


mjashanks:
When we create a table without a primary key, Budibase only ever returns the first row, when listing records.

I think it's right that we should require a primary key, when using the auto "Fetch tables from db" feature.

However, we should 
1. Not import tables without an identity.
2. Warn users about this.

You can probably log this as an enhancement if you like :)  - however, it causes confusion and is time-consuming for me to investigate this issue.

mjashanks:
I would add (on a slightly unrelated note) - maybe we should only support identity columns for "Fetch tables from db" - i.e. no composite primary keys. 

I feel like composite primary keys are rare enough to allow us to avoid them. This would wipe out this issue: https://github.com/Budibase/budibase/issues/1738

To compensate, we should improve custom queries  - thus people can still work with composite key tables.

Issue #2718, opened 2021-09-23
Disable handlebars helpers for custom queries to prevent clashes


mjashanks:
When adding a custom datasource query (could be REST, SQL or anything), some parameter names do not work.

I think it's because they clash with Handlebars internal methods/names. One that I have just encountered:
- title
- content

I have experienced others in the past.

The result - when I use `{{ title }}` it puts `Invalid Binding` into the string. This always causes me to check and recheck my query, trying to figure out what is wrong.

I think we should scope parameters, e.g. `{{ parameters.title }}`. To avoid a breaking change, we could fallback `{{ my_parameter }}` to `{{ parameters.my_parameter }}`

PR #3165, opened 2021-10-26
Not importing SQL tables without a primary key and displaying an error, disabling query HBS helpers

mike12345567:
## Description
This resolves the following issues:

1. #2718 - custom queries for data sources no longer have access to handlebars helpers as these aren't particularly useful and the conflicts kept causing issues.
2. #2820 - no longer importing SQL tables that don't have a primary key, an error will be displayed like:
image

This makes it clear what the problem is and is dis-missable, this is preferable to a toast as it can't really be missed. To do this I added the `InlineAlert` component to BBUI which can be used by others if desired.


Rory-Powell (approved):
LGTM!

aptkingston (approved):
LGTM! Just some tiny style points.

PR #3177, opened 2021-10-27
Relationship aware data provider automatic reloading + extras

aptkingston:
## Description
This PR enables automatic reloading of data providers which contain data related to other data which has changed. This means that things like creating rows related to `A` will ensure that the details of `A` are also reloaded, to avoid an issue where immediately saving `A` from a separate details form after creating linked rows would result in stale data being saved.

This also fixes a couple of issues with data provider automatic reloading from modal screens, as well as an issue with data provider cleanup which may have resulted in unnecessary API calls.

This PR also includes the ability for the `disabled` setting of form fields to be correctly toggled by conditional UI.

Closes #3158.
Closes #3135.


shogunpurple (approved):
LGTM

PR #3187, opened 2021-10-28
Fix budibase builder preview on safari

### LINKED ISSUE #2945: spinning circle in design mode with safari not chrome
### Discussed in https://github.com/Budibase/budibase/discussions/2933

<sup>Originally posted by **Bogey** October  8, 2021</sup>
For some weird reason Safari won't display the Screens and Layouts of the Design mode.
It only shows the spinning 'loading...' circle (see screen shot).

On Chrome, it works as usual.
I've disabled content blockers, emptied caches, cleared cookies and what not.
Unfortunately no working design mode on Safari.

It's Safari 15.0 on MacOS 11.6

Thx,
Eddie</div>

shogunpurple:
## Description
- Closes https://github.com/Budibase/budibase/issues/2945

This PR contains code that makes the budibase preview work on safari.

The issue that was happening was that safari does not allow you to pass custom events between frames directly for security reasons, so you have to make use of the `postMessage` and `receiveMessage` APIs for cross frame communication. We already use these APIs for communicating between preview and iframe, so there's nothing new being introduced here that would break a browser that didn't support it.

I've updated all of the inner iframe events to fire a `postMessage` to the parent, which is handled and runs the correct logic in the builder. All communication between inner iframe and builder is now reliant on the messaging API.

## Screenshots
Screenshot 2021-10-28 at 19 41 28

### DIFF
--- a/packages/builder/src/components/design/AppPreview/CurrentItemPreview.svelte
+++ b/packages/builder/src/components/design/AppPreview/CurrentItemPreview.svelte
@@ -32,6 +32,16 @@
     .component("@budibase/standard-components/screenslot")
     .instanceName("Content Placeholder")
     .json()
+  
+  // Messages that can be sent from the iframe preview to the builder
+  // Budibase events are and initalisation events
+  const MessageTypes = {
+    IFRAME_LOADED: "iframe-loaded",
+    READY: "ready",
+    ERROR: "error",
+    BUDIBASE: "type",
+    KEYDOWN: "keydown"
+  }
 
   // Construct iframe template
   $: template = iframeTemplate.replace(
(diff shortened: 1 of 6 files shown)


Rory-Powell (approved):

aptkingston (commented):
The `postMessage` technique looks much better 👍 nice one! There are a couple of other places in the builder where we use `CustomEvent` at the moment, for communicating between the screen modal iframe and real app. The two instances are:
- Invalidate datasource event in `dataSource.js`
- Notification event in `notification.js`
Would it be worth updating those instances (and their listeners) in this PR too just to ensure everything works in safari again?

PR #3194, opened 2021-10-29
Added troubleshouting for 'authenticated worker issue' #3193

tluyben:
## Description

Small troubleshooting header for people who accidentally visit the wrong port 4001 (which looks like it's working fine) including a start for adding other troubleshooting issues + solutions.


Rory-Powell:
Agree with @shogunpurple on this one, it's already stated to go to 10000 so we might be duplicating ourselves a bit here

PR #3224, opened 2021-11-02
SQL table manipulation and some external table UI rework

mike12345567:
## Description
This introduces the SQL table creation, updating and deletion from within Budibase to bring data source plus tables close to the same level of functionality as internal tables (minus attachment fields).

I'd like to add this extends a little bit past the lab day work, as I did a bit of UI re-work around SQL tables, modals and the datasource UI as we had discussed to try and make it a bit easier/less confusing to use. For example the "Create relationship" being renamed to "Define existing relationship" and making it possible to trigger this from the tables UI.

To achieve this i've updated the tables API to a similar style as the Rows API, having a main `index.js` which splits between an `internal.js` and `external.js` implementation. When an operation is requested on a table check the table ID to see if it is an external table and then go to the right API based on this. This was actually quite simple because the tables API is a lot more limited than the Rows API, the only endpoints that really needed different implementations was the `save` and `destroy` endpoints.

Under the external tables API is a new SQL base class, `sqlTable.ts` which the existing `sql.ts` class extends. It introduces three new operation types, `CREATE_TABLE`, `UPDATE_TABLE` and `DELETE_TABLE` - which will be controlled by the new class. The `sqlTable.ts` is again based on Knex, as with our current SQL rows system, using the [Knex Schema Builder](https://knexjs.org/#Schema) to generate the DDL used to manage the queries; this means that it should be portable between any future SQL databases we implement (that Knex supports). I've tested this with MySQL and Postgres (the currently implemented ones) and didn't need to make any changes to support either, due to the relatively simplified nature of `CREATE TABLE`, `ALTER TABLE` and `DROP TABLE/COLUMN`.

This introduces a few pieces of functionality which didn't exist previously for SQL tables, namely:
1. Options and multi-select data types - these can be supported as new columns
2. Formula data type - this is post-processed, not carried out in the database, previously it was just not possible to add this to the schema.

### Relationship creation
Most importantly this adds a controlled mechanism for creating completely new relationships between SQL tables, altering them to include foreign keys as required. This is done using an identical modal to what we use for internal tables, it is very opinionated and creates keys as required.

For example the following setups:
1. One test row -> many tasks rows
image
will produce a foreign key on the tasks table as below:
image
The inverse of this will occur for the one task, many test.

2. Many to many scenario will generate a junction table, as such:
image
The constraints will be applied to the junction table, which has a composite key to ensure uniqueness of entries.

Once these are setup they will function exactly as Budibase internal relationships do.

### Further work
There are a few things that I haven't currently implemented/could be a target for future iterations:
1. Foreign key/junction table cleanup if the relationship column is deleted. Right now the foreign keys and junction table are created when the relationship column is added, as its impossible to achieve the relationship column type without adding them - however I wasn't sure that deleting them on cleanup was the best idea. The issue with automatically deleting these is data loss and the fact that if an existing relationship was deleted it would remove references, could cause people problems. The column/junction table can easily be deleted through the UI if the user is ok with the data loss.

2. Some type mis-matches especially around generating relationships, right now everything is quite opinionated and if you have existing tables which you attempt to relate to, it will assume the foreign keys that need to be generated are simply unsigned integers. In the future we could attempt to garner more of this information when fetching the tables and perhaps match the type of the generated foreign keys as required. This only really matters for MySQL (number types mis-matches are common on constraints) and if a UUID/string has been used as a primary key.

3. Adjusting column types - this isn't something that can be done for the internal tables right now, but being able to adjust a string type to options, or longform, or a number to boolean, would fix some issues that people have had with SQL tables.

4. Updating the "Define existing relationship" modal to match the style of our relationship column creation modal, with the three options like `one table row -> many other table row`. Just to note I did a bit of work on this modal, to make it a little easier to use, but nothing major, just pre-filling everything so that you don't need to enter anything for most common scenarios.

## Screenshots
The data source menu, the "update display column" button has been removed as it was confusing and isn't really needed anymore, "fetch tables from database" and "Define existing relationship" have been made a lot more quiet, with the intention of bringing more inline with the table UI. Table creation button is the cta and has been made the same as the internal datasource (button is identical).
image

A define existing relationship button has been added to the tables UI for external tables.
image

When this is clicked the same UI is brought up as in the datasource UI, except now the table is pre-selected (when can bring this modal in line with how our relationship column creation works, with the options like `One person has many Tasks`).
image


shogunpurple (approved):
This is awesome - nice work. The abstractions over SQL -> knex schema building DSL make it so much easier, such a great library. 

The UI rework is much simpler too. Splitting between new and existing relationships is much easier to reason about.

aptkingston (approved):
Absolutely huge feature! Budibase is the best new SQL management UI around! UI changes definitely look good too, we always want to keep things as simple and accessible as possible. Nice one 👍

PR #3238, opened 2021-11-03
Fix: Prevent user updates in multi tenant mode from deleting user password

Rory-Powell:
## Description
- Prevent user updates in multi tenant mode from deleting user password
- Forward the authentication error from the backend to the login page to warn when an sso user is trying to log in with a password when one is not present
- Prevent bcrypt error and hanging login request when a password is missing. Show expired user error

## Screenshots
SSO with no password error message
Screenshot 2021-11-03 at 13 04 45

Expired user error message
Screenshot 2021-11-03 at 15 44 20

Incorrect credentials error message
Screenshot 2021-11-03 at 15 48 10

### DIFF
--- a/packages/builder/src/pages/builder/auth/login.svelte
+++ b/packages/builder/src/pages/builder/auth/login.svelte
@@ -44,7 +44,7 @@
       }
     } catch (err) {
       console.error(err)
-      notifications.error("Invalid credentials")
+      notifications.error(err.message ? err.message : "Invalid Credentials")
     }
   }
 
--- a/packages/auth/src/middleware/passport/local.js
+++ b/packages/auth/src/middleware/passport/local.js
@@ -10,6 +10,7 @@ const { getTenantId } = require("../../tenancy")
 
 const INVALID_ERR = "Invalid Credentials"
 const SSO_NO_PASSWORD = "SSO user does not have a password set"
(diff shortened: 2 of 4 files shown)


mike12345567 (approved):
LGTM! One question I have, for previous users that have had their password deleted and try to login, should we add an error message which lets them know they need to reset they're password?

EDIT - scratch my comment actually, theres no way to tell between an SSO user and an affected user so I don't think we could do a message like that.

Rory-Powell:
It's a good question @mike12345567 . I've left a comment on the discord channel about it. We should be able to detect if we do want to, it would be the same logic as we use for showing the new message for "SSO user does not have a password set"

shogunpurple (approved):

packages/builder/src/pages/builder/auth/login.svelte
```diff
@@ -44,7 +44,7 @@
       }
     } catch (err) {
       console.error(err)
-      notifications.error("Invalid credentials")
+      notifications.error(err.message ? err.message : "Invalid Credentials")
```
shogunpurple, line 47: NAB:

`err.message || "Invalid Credentials"` is equivalent here.

Rory-Powell: 👍  thanks

Issue #3237, opened 2021-11-03
REST integration and Outbound Webhook don't handle 204 response code.


pavel-orlov:
**Describe the bug**
Servers may respond with empty body to HTTP requests, ideally with 204 code but 200 too.
REST integration does 't seem to handle response codes and response with empty body result in exception "unexpected end of input" in `response.json()` from `fetch()`.

Outbound Webhook considers 200 is the only successful response code, 204 response leads to unsuccessful output status of the action.

Issue #3235, opened 2021-11-03
outgoingWebhook ignores custom headers.


pavel-orlov:
**Describe the bug**
Outgoing webhook action ignores custom headers for GET and DELETE methods, or for every method if body is null or empty.

**To Reproduce**
1. Create Outgoing Webhook action in Automate with custom headers.
2. Test the automation with GET/DELETE method or any method with empty body.

**Expected behavior**
Custom headers are included in the request.

**Additional context**
packages/server/src/automations/steps/outgoingWebhook.js. The if statement below should probably be taken out of its parent if.

```javascript
if (headers) {
      try {
        const customHeaders =
          typeof headers === "string" ? JSON.parse(headers) : headers
        request.headers = { ...request.headers, ...customHeaders }
      } catch (err) {
        return {
          success: false,
          response: "Unable to process headers, must be a JSON object.",
        }
      }
    }
```

Issue #3227, opened 2021-11-02
REST integration PUT method issue 


pavel-orlov:
**Describe the bug**
REST data source sends POST instead of PUT

**To Reproduce**
1. Add REST data source
2. Add PUT query
3. Run query

**Expected behavior**
PUT request is invoked

**Additional context**
packages/server/src/integrations/rest.ts uses POST in the "update" function.

Issue #3199, opened 2021-10-30
Budibase builder hangs trying to Save an App Action automation without typing a Name


cverdes:
**Describe the bug**
Budibase builder hangs trying to create an App Action automation without typing a name

**To Reproduce**
1. Open a Application with no automations
2. Click on the 'Create automation' button
3. Click on the 'App Action' button with the Name field empty
4. Click on the 'Save' button (with the Name field empty)
4. Budibase builder hangs. I found no action to recover the control, only refreshing the navigator page.

**Expected behavior**
The Save button should not be enabled if the Name field is empty

**Screenshots**
![Screenshot from 2021-10-30 14-29-47](https://user-images.githubusercontent.com/6440942/139543201-1d697f8d-2dc4-4fd7-84cf-6cfcbb7bbaf5.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.3 LTS
 - Browser: Chrome
 - Version: Version 95.0.4638.69 (Official Build) (64-bit)

**Additional context**
The application has no connection to any external data sources, and the only table on Budibase DB is 'users'

PR #3239, opened 2021-11-03
Various fixes for recent issues

mike12345567:
## Description
Fixing a number of issues that have been raised recently:
1. #3237 - allowing REST and webhooks to function with a range of 200-206 HTTP codes.
2. #3235 - applying the custom headers on outgoing webhooks consistently across all methods.
3. #3227 - in a rest datasource, making the PUT call actually use PUT rather than POST.
4. #3199 - fixing an issue causing the builder to crash when creating an automation.
5. #3143 - fixing the issue with webhooks and using the dev app ID vs prod app ID (displays the prod app ID for the trigger now, and updates dev apps to prod app IDs in the trigger endpoint).

### DIFF
--- a/packages/builder/src/components/automation/AutomationPanel/CreateAutomationModal.svelte
+++ b/packages/builder/src/components/automation/AutomationPanel/CreateAutomationModal.svelte
@@ -8,10 +8,13 @@
 
   let name
   let selectedTrigger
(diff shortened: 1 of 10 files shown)


shogunpurple (approved):
LGTM! Nice fixes

Issue #3139, opened 2021-10-25
Block: Table with Search


shogunpurple:
A simple table with a filterable group of results controlled through a search box with free search

**Setup Wizard**
- Table selection
- Filter fields

https://xd.adobe.com/view/a6cc4c4a-568b-4296-91b9-4a1556806870-f39f/screen/c7ce244a-ea53-4885-9dbd-2d6a6d1a08a6/

**Designs** 
https://xd.adobe.com/view/a6cc4c4a-568b-4296-91b9-4a1556806870-f39f/screen/b9cf4625-a13e-45a4-902d-e37fe14bfab0/

**Props**
![Screenshot 2021-10-25 at 13.39.31.png](https://images.zenhubusercontent.com/5fd773ddf3d67d4ceebe466b/e332e80a-83c0-4924-9ba7-71d0821f6f39)

PR #3249, opened 2021-11-04
Core blocks implementation + table with search block

### LINKED ISSUE #3216: Checkbox tick is out of place when the checkbox size is changed
CleanShot 2021-11-01 at 14 17 03

To recreate the issue:
Add checkbox, change size

aptkingston:
## Description
This PR adds the core implementation for developing block components. Blocks are simply prebuilt composite components which provide a lot of functionality out of the box, with minimal configuration. I've also added the initial "table with search" and "card list with search" blocks. 

As an extra feature, I've also added support for table sorting to actually update the server side sorting. When you click a table column to sort, it will actually update the parent data provider and perform a new server side search with new sorting. This is in line with how the new Data UI pagination works.

I've also added the ability to click directly on rows within tables. This removes the need to add a child link or button into table rows, and it's now far faster to configure your table.

Addresses #3139.

## Blocks
Blocks are a new component category that appears first in the list. At the moment there is only the one new block - table with search.

image

Blocks appear like any other component, and have settings. Block settings can be split into multiple sections, as usually blocks will have more settings than simple components. Here's an example of the new settings panel for a "table with search" block. You can see that the settings have been split into multiple sections to make them more readable.

image

Blocks appear as a single element in the component tree, and can be moved, copied, deleted, drag-and-dropped and otherwise treated as any other single component.

image

Blocks do not accept child components (but they could down the line if we need to support this).

## Table with search block
This block contains a table which can be searched and filtered. A title and button can be also added. The block supports any data source - an internal table, views, SQL table, queries, non-plus datasources like a REST API and nested data providers.

image

One advantage of blocks is that we can ensure they are fully mobile compatible, as this one is:
image

To get up and running with a paginated, server-side searched table is now only about 3 clicks. You can just pick a data source, pick your search columns and that's it.

Minimal settings to get the block to work:
image

When you pick what columns you would like to search via, an appropriate field component will be rendered. This could be a string field, an options picker, a checkbox or a date field. You can't currently search on attachments, multi-selects, relationships or formula fields.

## Card list with search block
This block contains a list of cards which can be searched and filtered. The majority of the settings are the same as the table block, with the exception of the card specific settings. This block was much harder to implement as the settings for things like card title and subtitle need to be able to reference bindable properties for your data model. Both the provider and consumer of these bindings is the same component, which cannot happen with normal components. This has required some new low level concepts to be introduced to handle this, as well as handle binding enrichment and data context management with blocks.

image

## Table linking
New settings have been added to the table component to support clicking on rows within the table to link to other screens. There are 3 main settings to control this, and one optional setting under a new "Advanced" section. The advanced setting is for non-internal and non-sql datasource (i.e. those that don't use `_id`) and lets the user pick whatever column identifies their rows. This value is what will be appended to the chosen URL. 

image

Table linking settings have been pulled into the new table with search block as well.

## Other stuff
- Refactored and consolidated how datasource schema are fetched in the client library, as there were certain components which would not work properly if configured with certain datasource types
- Fix spectrum bug with large and extra-large checkboxes. Closes #3216
- Fixed an issue when saving docs that contain multi-options types


shogunpurple (approved):
LGTM! Refactor much cleaner as well - good job

aptkingston:
This will still require a bit of testing before actual release, but merging this into develop now so that it's available for video recording using the staging env.

Issue #3215, opened 2021-11-01
Handling user metadata within apps


mike12345567:
**Describe the bug**
An issue with the in app user metadata system is that when global users are created, or given permission to use that app their metadata doc is not created until they actually access the app. This is a problem as if a builder wants to build in some information specifically for that user, they cannot do so until the user has used the app.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create an app
2. Create another user
3. Make sure the other user has permission to use the app
3. Attempt to assign something to the user within the app (they will not appear in the user metadata table)

**Expected behaviour**
They should appear within the metadata table in an app once they are created, I think how we should manage this is by reversing how we get users. Right now the way this works is that within an app we lookup the list of user metadata docs that exist in the app, then enrich them with details from the global users. Instead we should retrieve the list of global users, reduce this down to users which can access the app (have a role that isn't public) then enrich the users with whatever metadata docs that exist. This will work fine as when a metadata doc is saved that doesn't exist it will simply be created (as needed).

There is still an issue around what should happen when the user no longer has access to an app, e.g. they've been deleted or their role has been revoked - but this may be a discussion for a separate issue as its more of a feature than a bug. For solving this @Rory-Powell had the idea to put users into a `deactivated` state when the metadata doc exists, but there is no corresponding global user (they don't have access, or they simply don't exist anymore).

PR #3251, opened 2021-11-04
Syncing user metadata - allow builder to interact with user metadata in app

mike12345567:
## Description
This issue arose when testing #3215 - which was about being able to manipulate/use metadata about users within apps before they have logged into the app. This feature makes it so that when the user is created they will be added to the metadata database, assuming they have access to the app.

Originally I had been going to implement this by just reversing the way we looked up users, looking to the global set of users and then enriching with metadata but this doesn't work for one big reason: we now use Lucene to lookup users. As we are actually searching for the users, the call needs to support pagination, filtering and sorting; to do this across two different databases (one of which doesn't support searching/isn't indexed) was far too complicated. Instead I decided that keeping the relevant elements of the user metadata within apps up to date would be easier to achieve and will simplify a few things moving forward. This has the added benefit that we can now search the metadata docs on things like `email` or `roleId`, something which we couldn't do before.

There are a few core features to mention about this:
1. When the user `save` endpoint is hit in the worker it makes a call to the server to alert that a user has been updated. This call syncs the user across all apps that they have access to (if they are a builder then that means all apps). The sync is quite simple, it has two paths it can take:
    * The user is not a builder, it will retrieve their `roles`, remove props that aren't of importance (`password`, `forceResetPassword`) then use the roles object to find the production app IDs which they have access to. It will iterate through the production and development databases and update the metadata within.
    * They are a builder, in which case it simply retrieves a list of production app IDs and proceeds with the above step, but instead their role ID will be ADMIN for them all.
2. The sync call will also occur when the user is deleted, but rather than deleting the metadata it sets the user's status in the app to `inactive` - meaning they are disabled but no data is lost.
3. When a builder opens an app for the first time as part of the production -> development sync it also makes sure all users are up to date.

This also introduces a new environment variable to the worker `APPS_URL` so that it can communicate with the apps service, i've made it so that if this isn't filled out it will be injected with a the best value for it, depending on whether the service is being run in development or production.

### DIFF
--- a/packages/server/src/api/controllers/user.js
+++ b/packages/server/src/api/controllers/user.js
@@ -6,25 +6,58 @@ const {
 const { InternalTables } = require("../../db/utils")
 const { getGlobalUsers } = require("../../utilities/global")
 const { getFullUser } = require("../../utilities/users")
+const { isEqual } = require("lodash")
+const { BUILTIN_ROLE_IDS } = require("@budibase/auth/roles")
 
-function removeGlobalProps(user) {
-  // make sure to always remove some of the global user props
-  delete user.password
-  delete user.roles
-  delete user.builder
-  return user
-}
-
-exports.fetchMetadata = async function (ctx) {
-  const database = new CouchDB(ctx.appId)
-  const global = await getGlobalUsers(ctx.appId)
-  const metadata = (
-    await database.allDocs(
+exports.rawMetadata = async db => {
+  return (
+    await db.allDocs(
       getUserMetadataParams(null, {
         include_docs: true,
       })
     )
   ).rows.map(row => row.doc)
+}
+
+exports.syncGlobalUsers = async appId => {
+  // sync user metadata
+  const db = new CouchDB(appId)
+  const [users, metadata] = await Promise.all([
+    getGlobalUsers(appId),
+    exports.rawMetadata(db),
+  ])
+  const toWrite = []
+  for (let user of users) {
+    // skip users with no access
+    if (user.roleId === BUILTIN_ROLE_IDS.PUBLIC) {
+      continue
+    }
+    delete user._rev
+    const metadataId = generateUserMetadataID(user._id)
+    const newDoc = {
+      ...user,
+      _id: metadataId,
+      tableId: InternalTables.USER_METADATA,
+    }
+    const found = metadata.find(doc => doc._id === metadataId)
+    // copy rev over for the purposes of equality check
+    if (found) {
+      newDoc._rev = found._rev
+    }
+    if (found == null || !isEqual(newDoc, found)) {
+      toWrite.push({
+        ...found,
+        ...newDoc,
+      })
+    }
+  }
+  await db.bulkDocs(toWrite)
+}
+
+exports.fetchMetadata = async function (ctx) {
+  const database = new CouchDB(ctx.appId)
+  const global = await getGlobalUsers(ctx.appId)
+  const metadata = await exports.rawMetadata(database)
   const users = []
   for (let user of global) {
     // find the metadata that matches up to the global ID
@@ -52,7 +85,9 @@ exports.updateSelfMetadata = async function (ctx) {
 exports.updateMetadata = async function (ctx) {
   const appId = ctx.appId
   const db = new CouchDB(appId)
-  const user = removeGlobalProps(ctx.request.body)
+  const user = ctx.request.body
+  // this isn't applicable to the user
+  delete user.roles
   const metadata = {
     tableId: InternalTables.USER_METADATA,
     ...user,
--- a/packages/server/src/api/controllers/user.js
+++ b/packages/server/src/api/controllers/user.js
@@ -8,11 +8,7 @@ const { getGlobalUsers, getRawGlobalUser } = require("../../utilities/global")
 const { getFullUser } = require("../../utilities/users")
 const { isEqual } = require("lodash")
 const { BUILTIN_ROLE_IDS } = require("@budibase/auth/roles")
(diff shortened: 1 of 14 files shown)


packages/server/src/api/controllers/user.js
```diff
@@ -4,27 +4,141 @@ const {
...
-  // make sure to always remove some of the global user props
+exports.syncUser = async function (ctx) {
+  let deleting = false,
+    user
+  const userId = ctx.params.id
+  try {
+    user = await getRawGlobalUser(userId)
+  } catch (err) {
+    user = {}
+    deleting = true
```
Rory-Powell, line 82: This might give us some bother if there's ever a non 404 error here, accidentally marking the user as inactive. 

We could either 
- check for the error code to infer delete (easiest / fastest)
- send the action (create, update, delete) as a part of the request from worker -> server 
- send the user in the payload and don't read from the db (probably most future proof so that the database isn't coupled between services)

Probably option 1 is most suitable here, I'm mostly curious about: is removing the dependency on the same db entities across apis a longer term goal for us? Or are we fine with always giving multiple apis access to the same data. From experience it becomes an issue eventually and produced inconsistencies in the db.

packages/server/src/api/routes/user.js
```diff
@@ -34,5 +34,10 @@ router
     authorized(PermissionTypes.USER, PermissionLevels.WRITE),
     controller.destroyMetadata
   )
+  .post(
+    "/api/users/sync/:id",
```
Rory-Powell, line 38: Should it be `/api/users/metadata/sync/:id`? to be consistent with the rest of the routes in the api

packages/worker/src/utilities/appService.js
```diff
@@ -0,0 +1,26 @@
...
+  if (env.isTest()) {
+    return
+  }
+  const request = { headers: {} }
+  request.headers[Headers.API_KEY] = env.INTERNAL_API_KEY
+  if (isTenantIdSet()) {
+    request.headers[Headers.TENANT_ID] = getTenantId()
+  }
+  request.headers["Content-Type"] = "application/json"
+  request.body = JSON.stringify({})
```
Rory-Powell, line 17: Might be worth pulling this request init logic out into it's own function so that it can be reused for future requests

packages/worker/src/utilities/appService.js
```diff
@@ -0,0 +1,26 @@
...
+  request.method = "POST"
+  const response = await fetch(
+    checkSlashesInUrl(env.APPS_URL + `/api/users/sync/${userId}`),
+    request
+  )
+  if (response.status !== 200) {
+    throw "Unable to sync user."
```
Rory-Powell, line 24: I see a lot of `throw "string"` vs `throw new Error("string")`. Do we have a preference? 
from what I've seen the latter is more preferable as it results in a well formatted json message in response bodies, not sure if that's applicable here in this client call. I have sonar lint running in VS Code and it warns me about not wrapping these with `Error`

packages/server/src/api/controllers/user.js
```diff
@@ -4,27 +4,141 @@ const {
...
   delete user.password
+  delete user.forceResetPassword
   delete user.roles
-  delete user.builder
-  return user
+  // run through all production appIDs in the users roles
+  let prodAppIds
+  // if they are a builder then get all production app IDs
+  if ((user.builder && user.builder.global) || deleting) {
+    prodAppIds = (await getAllApps(CouchDB, { idsOnly: true })).filter(
```
Rory-Powell, line 93: Might be nice to have a new function in the auth lib `getAllDeployedApps`

packages/worker/src/utilities/appService.js
mike12345567: Yeah, I usually just use the `throw "string"` because its just easier to use and expecting all errors to be wrapped is an error in an error waiting to happen (theres always going to be some libraries etc that don't wrap) - so my preference is to not wrap since the formatted version really isn't that useful.

packages/server/src/api/controllers/user.js
mike12345567: Yeah, I'll add a utility function for this.

packages/server/src/api/controllers/user.js
mike12345567: Good catch, i'll go with option 1 as I think that makes the most sense, just look for a 404, otherwise ignore. 

In terms of consistency, me and Marty discussed this, having to duplicate data across all of the databases isn't the nicest option but the only other option would be to remove access to the data within the apps, which I think would be a pretty breaking/brutal change. The issue comes down to searching/filtering, we need the data to be in the index, along with the actual user metadata (stored in the app) so that it can be used in a filter together.

I think it makes sense that some of this data should be duplicated, we also still do what we did before, going to the central DB to access the data when retrieving, so the data actually stored per app database is really just used for filtering, meaning there could be a filtering inconsistency, but when its actually retrieved it should be up to date.

Rory-Powell: Chatted a bit about this offline. The main concerns is about reading and writing data from the `global` db in the app service, it seems we don't ever read from `app` databases in worker so it might be a good logical split to prevent cross domain direct db access like this in case service logic diverges, as it locks us in to using a shared lib which couples deployments of services.

Nothing required for this PR in particular, we'd like to start outlining the bigger ticket tech debt items to factor into future iterations. 

Some additional things we discussed:
- Write / read locking databases to particular services
- Moving apis into a single service (makes the above redundant)
- Moving automations into their own service

Rory-Powell (approved):
LGTM!

Issue #3277, opened 2021-11-07
The revert button once the error occur so the entire app get deleted


nnasoke:
**Describe the bug**
At the app builder screen: I try to undo the previous action by clicking the 'Revert' button at the top-right corner of the app builder screen but it gives me an error and then my app gets deleted.

**To Reproduce**
1. Go to 'https://[my domain].budibase.app/builder/portal/apps'
2. Click the button 'Create App' and choose 'Start from scratch', naming the app as 'Test Crash'.
3. Follow the steps of creating a data source. Then, create a table naming 'table1', leave all tick-boxes as default.
4. Once the app has been created, click the 'Design' tab at the top-left corner.
5. Click on page '/table1/new/row' and make a selection at the 'Save' button located in the app's design screen.
6. Click on the toolbar 'Elements -> Icon' to add a new icon.
7. Undo the previous action by clicking the revert button at the top-right corner. Confirm 'Revert', then you will see error 'Error reverting changes: Unable to revert. [object Object]'.
8. Then go back to ''https://[my domain].budibase.app/builder/portal/apps', the 'Test Crash' is missing.

**Expected behavior**
The 'Revert' should be undoing the previous action. But the consequence of error causes the app to get deleted.

**Screenshots**
I have a video capture for the entire steps but no idea how to put it into the ticket. Please let me know if the reproduce steps just don't work somehow.

**Desktop (please complete the following information):**
 - OS: OSX Monterey
 - Browser: Chrome
 - Version: 95.0.4638.69 (Official Build) (x86_64)

**Additional context**
The issue happens to Budibase cloud.

HyperCrowd:
I've identified a bit more details on this

#3278

aptkingston:
Hi @nnasoke. Thanks for raising this - this is absolutely a valid bug. The "revert" functionality lets you revert your apps to the last published state. This is so that if you have an application that has been published an in use, but then you've made some changes to your app and haven't published those changes yet, then you can revert your app back to the latest published version.

The issue here is that there is no published version to revert to, because your app has not been published. This is absolutely an issue and we should not even show that button if the app has not yet been published.

We will get this fixed as a priority.

@HyperCrowd I will respond to your issue separately but it is not related to this.

nnasoke:
Thank you Andrew @aptkingston . The response is so fast, appreciated that.

PR #3289, opened 2021-11-08
fix revert on unpublished app

shogunpurple:
## Description
Fixes https://github.com/Budibase/budibase/issues/3277

This PR contains code that prevents reverting an app when it hasn't been published. The problem wasn't actually the logic in the controller, it was the fact that the **production app database gets created on app creation** so when you click revert, it just reverts your app to an empty app with no metadata.

We can't rely on the prod app database not being there anymore without doing a fairly large backfill job, because so many apps have been created, but we can check when they click deploy to make sure that the app has no deployments, and throw accordingly. I also considered adding logic to add a `published` field to the app metadata or pull the number of deployments in the `appPackage` call to determine this, but this would have to be kept up to date on deploy and felt a little nasty.

### DIFF
--- a/packages/server/src/api/controllers/application.js
+++ b/packages/server/src/api/controllers/application.js
@@ -337,7 +337,7 @@ exports.sync = async ctx => {
   } catch (err) {
     // the database doesn't exist. Don't replicate
     ctx.body = {
-      message: "App sync completed successfully.",
+      message: "App sync not required, app not deployed.",
     }
     return
   }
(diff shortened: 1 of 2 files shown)


mike12345567:
LGTM! Good catch, this might actually be a consequence of the new sync feature, since sync is now attempted when the builder is opened, I think attempting to trigger a replication from a database which doesn't exist doesn't actually throw an error in Couch, it instead creates the missing database and then replicates nothing (might be related to our PouchDB config having `skip_setup: false` set by default).

We could probably add a check to the sync to see if the production database exists, if it doesn't then simply don't try to replicate. Although its probably safer to make sure the production database contains something worth reverting for this issue (such as an existing deployment).

shogunpurple:
Added a check to make sure that the prod DB isn't created on sync, so future users won't run into the same issue 👍

Issue #3182, opened 2021-10-28
Sort by relationship does not work


mjashanks:
If we sort by relationship, either in a data provider or in the `Data` section, the sort order is random. The expectation is that it sorts by the display name.

I think that this will likely be impossible for internal tables (?). If so, we should disable sorting by relationship.

Having said that, maybe allowing it to sort by the relationship `_id` is slightly better than nothing at all - this will allow us to group rows together that have the same relationship, even if not sorted by the display name.

mike12345567:
Hi @tluyben - I think we would need to implement a more complex sort system as a separate feature, for now we will disable it as this is the quickest option to protect against problems arising.

Right now our sort is implemented at the database level so that it plays nicely with pagination, I believe how a JS/complex sort would be implemented is client-side; this is problematic for any sort of paginated data however so I think we'd need to put a bit more thought around this.

PR #3291, opened 2021-11-08
MS SQL Server data source plus

mike12345567:
## Description
This PR introduces Microsoft SQL Server (MS-SQL) as a new data source plus option. This includes the following features:
1. Fetching tables and generating schemas
2. Managing joins as Budibase relationships
3. CRUD over existing and new rows within tables
4. Creation, altering and deletion of SQL tables (as with the new SQL feature)

This feature should function identically to Postgres and MySQL, with only a few minor caveats, these are as follows:
1. Pagination in SQL Server works a little differently, namely it cannot function without a sort column. If no sort column is provided for SQL Server read queries (with pagination) sorting will be performed on the primary key.
2. Pagination is not implemented in the most efficient method for SQL Server (it currently uses the same method as MySQL and Postgres), it will work just fine, but there is technically more efficient ways for very very large data sets - but re-implementing this specifically for SQL Server didn't feel necessary at this point and time.
3. I implemented this in such a way that the master database can be used and all of the system tables will not be exposed (such as `MSreplication_options` or `spt_monitor`). These tables are simply ignored because updating them could cause problems.
4. All row creation calls have a `SELECT SCOPE_IDENTITY() AS id;` added to them, this was the best way I could find to get the primary key of the row that was just created if it is using a computed or identity field. Normally the inserted ID is just returned, or in the case of Postgres the `returning` syntax exists so we can simply get back what we just changed, but my MS-SQL its a bit of a mix between using the create/update body to find the primary keys and using the `SCOPE_IDENTITY` to look up computed.

Schema fetching in SQL Server is a little more complex than our other datasources, but it basically does it all through the `INFORMATION_SCHEMA` view which seems to be the most forward and backwards compatible method. The process works something like:
* Fetching a list of table names and some very basic type information
Then for each table we get the following:
* Fetching a list of constraints, such as which column is the primary key
* Fetching a list of computed and identity columns, we consider these to be auto columns (this may not be the perfect solution, but has worked well for Postgres and MySQL, stops the user from trying to input over a column that is supposed to be generated)

One thing that I needed to do to make this work was to rework how the SQL Server datasource actually makes requests, as sometimes our plus data sources will make multiple queries per datasource instances (when just using custom queries there is only ever one single SQL statement sent). When the `connect()` call is made it now just gets a connection from the pool, then when a query is made is gets an instance of the `client.request()` rather than at connection time. This is required because we use `request.inputs()` for the various user inputs to protect against SQL injection, but if there is only ever a single request instance then input variables are shared across all requests made.

I also centralised our type mappings for SQL to Budibase types into one gigantic list, as over-compensating for each database seemed like a better choice (since we were generally just copying the list of each new SQL datasource anyway).

The changes may look quite large but a lot of this was refactoring, as now there are three SQL datasource plus types and there was a bit of repetition between their implementations. I tried to merge as much as possible to the central SQL base class so that it could all be managed cleanly.

This also includes a fix for #3182 - simply because when playing with SQL databases these fields did not play nice and attempting to sort by formula, relationships, multi-selects or attachments doesn't work for external or internal tables.

One final note is that the docker compose file that I added to the server scripts for SQL Server isn't that great, sadly its not that simple to spin up a SQL Server database with compose, so I had to build my own image which generates tables at startup with a shell script - if you change the startup init SQL then you need to rebuild the image for it to take affect.

## Screenshots
Fetching tables:
image

Interacting with some tables, including one that has been created newly, as well as showing a many to many relationship where a junction table has been generated.
image


shogunpurple (approved):
LGTM! Awesome job consolidating the DB types. It's all looking a lot more consistent now. To be honest it doesn't seem like we have had to do many crazy things to get the schema definitions, I thought it might be much more complex to do so.

mike12345567:
> LGTM! Awesome job consolidating the DB types. It's all looking a lot more consistent now. To be honest it doesn't seem like we have had to do many crazy things to get the schema definitions, I thought it might be much more complex to do so.

Yeap its actually not that bad, the hardest part about introducing a new datasource plus isn't really the implementation, its just testing the specific elements which are different between the databases and adding cases for them, but once thats done its pretty plain sailing!

PR #3294, opened 2021-11-09
Fix/feature flag safari

shogunpurple:
## Description
Adding feature flag support for message passing so that old versions of the client library still work inside the builder. This will allow us to release master with the latest bug fixes and updates.

I've also fixed an issue where the call to fetch templates failing prevents users from creating an app. The "Start from scratch" and "Import app" buttons will appear regardless as to whether the template call fails or not. I'll make the alert look better by using the `InlineAlert` component added in `develop` as part of the next work I do off `develop`.

## Screenshots


aptkingston (approved):
LGTM. Good stuff, should be good to go now.

Issue #3152, opened 2021-10-26
Full attachment objects in bindings


shogunpurple:
When I'm rendering attachments in the budibase builder, I want to have access to the full attachment object rather than just the URL, so that I can render it with a repeater and use the file name, file size and other information in the attachment objects.

**COS**
- Revert change to render attachment URLs only and return the full objects
- Update dependent components such as the dropzone to pluck the URLs from the attachment objects

Issue #3151, opened 2021-10-26
Full Bindings in Relationships


shogunpurple:
When I'm using bindings with relationships in the budibase builder, I want to have access to the full relationship object so that I can use the ID for links, datasource queries and more.

**COS**
- Revert the frontend changes to parse out the displayName of the relationship by default, and return the object
- Update UI components that depend on the displayName parsing

gvlx:
Hi, trying to understand this issue.

I can access the full relationship object by including a new data provider inside the main one, using the relationship as the source.

This allows me to use repeaters or tables to show all the details of the related data.

aptkingston:
Hi @gvlx. You're correct, that's exactly how to do it at the moment if you want to pull relationship data.

This issue concerns whenever you use a relationship binding in something like a text field. Currently, if you use a binding like `{{ Current user.Dogs}}` (where `Dogs` is a relationship field) - then we automatically just convert that binding into some readable text, by pulling out all the "display column" values and joining them. For example, if that field contained a relationship to multiple dogs and the primary display column of the dog table was their name, then we would display the text `Rover, Barney, ...` etc. This is useful because it allows relationships to display readable text.

What we actually store in those columns is an object with the `_id` of the relationship and the `primaryDisplay` value. Sometimes it would be useful to be able to access the stored `_id` directly, without having to nest another data provider. This change removes the previous behaviour of making readable text, and instead allows the real array of objects to be used. When combining this with JavaScript bindings, we can now let users do more advanced things by using the objects directly.

The current method of using data providers to get relationship data will not be changed, and is still the correct way to do things for most use cases. This feature just lets you use relationship `_id` columns a bit easier, without having to use another data provider - in case you just want to do something simple like create a link to related record's detail page.

gvlx:
OK. I understand now.
Thank you for your explanation.

gvlx:
You should probably check if this approach would still work or, in fact, would not be needed anymore after your change.

See ➡️  https://github.com/Budibase/budibase/discussions/3365#discussioncomment-1636793

PR #3299, opened 2021-11-09
Full bindings for relationships and attachments

aptkingston:
## Description
This PR removes the binding transformers that we used to have in order to make relationships and attachments easily bindable. Previously we used to append a suffix to runtime bindings for data fields of certain types, and then created these artificial data fields in the client library.

Addresses #3151, #3152.

This PR removes that feature, and means that using a binding for a relationship or an attachment will instead be served the full real array. It's much easier to work with this in JS than HBS.

I've done this in a non-breaking way, by ensuring that any apps that are using the existing transformed bindings will continue to function. The client library side of this feature is not being removed yet, so although it will not be possible to create new bindings like these, existing ones will continue to work as they always have.

I've also fixed a small bug with links, where clicking on links in the builder preview would actually change the iframe location.

Things like dropzones and the way tables handle displaying relationships did not need updated as these do not use bindings - they already directly use the full array instance from the row.

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -217,18 +217,8 @@ const getProviderContextBindings = (asset, dataProviders) => {
       keys.forEach(key => {
         const fieldSchema = schema[key]
 
-        // Make safe runtime binding and replace certain bindings with a
-        // new property to help display components
-        let runtimeBoundKey = key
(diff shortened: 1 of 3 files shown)


shogunpurple (commented):
LGTM! Nice to remove a little bit of complexity here too.

Rory-Powell (approved):

Issue #3256, opened 2021-11-05
Template bugs


Rory-Powell:
**Describe the bug**
Template are now live at https://budibase.com/templates

Some initial fixes needed:

**Cancellation**
- If I cancel the template creation, I'm presented with the pop every time I reload the page
   - Cancelling template creation should update the `budibase:init` cookie so that template creation is no longer active

**Logout**
- If I log out and log thinking that will fix it I still see the pop up
   - Logging out should clear the init cookie so to give the user a path for fixing existing issues  

**Account Redirect**
- If I'm logged in as a root account holder I get sent to account portal instead
   - When budibase.app is navigated to directly, we send to account portal 
   - For non-root account holders they are then redirected to their budibase tenant as they don't have portal access
   - For root account holders they remain in account portal and the template creation flow ends unless they visit their tenant
      - We should add a mechanism to this redirect flow that will indicate to the account portal that the user should be sent back to budibase 
      - @shogunpurple suggested detecting if the init cookie is present and redirecting to tenant if so
      - An alternative approach would be to add an additional param `?return=budibase`. This might be useful for future scenarios and could be added [here](https://github.com/Budibase/budibase/blob/master/packages/builder/src/pages/builder/_layout.svelte#L33)

PR #3300, opened 2021-11-09
Fix/templates

shogunpurple:
## Description
- Fix issue where `ready` event was still required for safari feature flagging
- Fixes for https://github.com/Budibase/budibase/issues/3256
- Added an input for users to have to enter their app name to revert - it's a destructive operation and should have safety
- Apply tenant ID from user rather than redirect to account portal to fetch it for account owners

## Screenshots

Screenshot 2021-11-09 at 18 01 41


mike12345567:
LGTM! Nice to get that check in so quickly for the revert as well, we should probably add one for app deletion too, although its a much harder operation to do accidentally.

mike12345567 (approved):

Rory-Powell (approved):
LGTM

shogunpurple:
Added the same check for deleting an app as well 👍

Issue #3119, opened 2021-10-22
Oracle Datasource (Custom)


shogunpurple:
When I’m selecting a datasource from the datasource options in the backend UI, I want to be able to have the custom datasource experience with an oracle database.

**COS**
- Add oracle data source with custom query functionality

PR #3301, opened 2021-11-09
Oracle datasource

Rory-Powell:
## Description
https://github.com/Budibase/budibase/issues/3119

Adding oracle as a datasource with custom query functionality. 

More details can be found in the oracle readme:
https://github.com/Budibase/budibase/blob/oracle-datasource/packages/server/scripts/integrations/oracle/oracle.md

## Screenshots
Screenshot 2021-11-09 at 17 34 10
Screenshot 2021-11-09 at 17 34 15
Screenshot 2021-11-09 at 17 34 26


packages/server/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM node:12-alpine
+FROM node:12-slim
```
Rory-Powell, line 1: The oracle client relies on several system binaries that are not present in alpine linux. Even when added manually there are still issues and inconsistencies with the regular linux install script, while creating an image almost the same size. The cleanest approach was to bump the node base image to slim. 

This brings the total size of the base image from `89.5MB` to `142MB`.
Coupled with the installation of the instant client this brings the total image size of the server  from `1.22GB` to `1.45GB`

shogunpurple: All good - to be honest our app server image could probably be smaller in general. It does quite a lot. I saw this the other day, which might be worth a try at some point to see if it can reduce our images down. It would also make deployments etc much faster.

https://github.com/docker-slim/docker-slim

packages/server/src/integrations/oracle.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    constructor(config: OracleConfig) {
+      super("oracle")
+      this.config = config
+    }
+
+    private query = async (query: SqlQuery): Promise<Result<any>> => { 
+      let connection
+      try {
+        connection = await this.getConnection()
```
shogunpurple, line 78: According to the oracle-node docs for `getConnection`:

> Obtains a connection from a pool in the connection pool cache or creates a new, standalone, non-pooled connection.

It doesn't seem like we have a pool here, but other data sources tend to use a pooled approach. We had some problems in the past with lots of connections being created, and pooling made things a bit more efficient. 

We can still use `getConnection`, but on the pool object rather than `oracledb` directly.
https://oracle.github.io/node-oracledb/doc/api.html#connpoolcache

mike12345567 (approved):
LGTM! A pain that it requires an external client to interact with Oracle DB, but to be honest I expected it to be much worse than that! Also I really like the addition of the `README.md` in the `scripts/integration` directories, I should really look at adding one to document each of the testing databases.

shogunpurple (approved):
image

LGTM! Just that comment about pooling, but it might not be necessary as it looks like we are doing the appropriate cleanup in your `finally` block.

packages/server/src/integrations/oracle.ts
Rory-Powell: I'm not 100% sure if the pooled approach translates well to our application. If we were a traditional web app that connects to one database it would make a lot of sense, for us it's not as straightforward as we have many connection configurations. Normally you will create your pool once as a part of application initialization as early as possible. We would need to maintain a pool per tenant which would vastly increase the number of connections. Oracle actually handles this quite well as they provide a pool cache layer behind an alias as a part their api, but I'm not sure it's right for this use case and is probably intended for managing a small number of pools.

There's a part in the docs about the correlation between the connection pool sizes and nodes own thread pool https://oracle.github.io/node-oracledb/doc/api.html#-162-connections-threads-and-parallelism that tells us:
> The thread pool size should be equal to, or greater than, the maximum number of connections.

This sounds like maintaining lots connections can run into quite a bit of bother, I'd be more confident in the fire and forget method that we use currently. 

I had a look at the postgres integration for a reference on how we do pooling there, I'm not sure it's functioning as expected. Given the pool is `static`, I expected a global variable being overwritten for every new config being queried which sounds like there could be a lot of contention. When I debugged there I see that a new pool is created for every query being made and then thrown away e.g.

https://user-images.githubusercontent.com/8755148/141772894-72753971-b7f2-4f54-bb62-20de8d5730cb.mov

If we wanted to support pooling we could consider doing some dependency injection instead of creating the pool internally each time, but it's probably something we only want to do in single tenant mode

Rory-Powell: > We would need to maintain a pool per tenant 

It might actually be a pool *per config, per tenant

shogunpurple: Nice one - great detail digging into that. Yeah it looks like it's "working"  for postgres in that it throws away the pool afterwards, which seems to give the same behaviour as fire and forget. Probably fine for now but since fire and forget works we should probably just make them all like that for consistency.

Actually now that I think about it - a lot of this was done before we even supported multi tenancy or cloud, where self-hosted users were reporting that they were blowing their maximum number of connections from using bb against a single database.

The key thing is that we just don't blow the database quota for connections, so think we are good here 👍 .

PR #3331, opened 2021-11-11
bust cache when app versions are different

shogunpurple:
## Description
Adding a version flag to the client library fetch with the app version to ensure that we do not get cached versions of `budibase-client.js` between versions.

Credit to @aptkingston for the idea 👍 

## Screenshots


aptkingston:
Awesome. We should update the usage of `clientLibraryPath` in `controllers/application.js` too since that's where the builder loads its app definition from, and ultimately where the URL comes from that gets used in the client preview which we also need to cache bust.

aptkingston (approved):

PR #3334, opened 2021-11-11
Feature/auto screen UI

PClmnt:
## Description
New Autoscreen UI

The main change here is that you now only select one table, and the screens are autogenerated from that. i.e you no longer select just one screen. detail / list / create screens all get added

image
Default screen when nothing has been added, you can open up the new modals by clicking the cta or the plus icon as normal

 You're then brought to this screen where you can select multiple autoscreens or just a blank screen, selecting an option from one selection disables the other. i.e if you select Blank then you can't select an autoscreen.

image

If you click on autoscreen you're brought to this screen where you can pick the position of your nav bar, selecting this mutates the navigation layout on save of the modal.

image
 However if you selected the "Blank" option, you get an intermediary screen asking for what name and url. 

image
If you've selected a table, then the 3 screens will be created, and a link to the List screen will be added also. 

Questions:

- What do we want to do with the first modal if the user already has created a set of autoscreens using the Sales table for example? Create a new set but with a new route and name? i.e they've created /`sales` so create a new route with   /`sales-1`? 
- @joebudi not sure about the fonts? they look alright to you? the mockup fonts didnt match what our typography components have


joebudi:
@PClmnt great work.

_What do we want to do with the first modal if the user already has created a set of autoscreens using the Sales table for example? Create a new set but with a new route and name? i.e they've created /sales so create a new route with /sales-1?_
Exactly - they can change the route in settings if they like

_@joebudi not sure about the fonts? they look alright to you? the mockup fonts didnt match what our typography components have_
The detail font type is off. Reduce its color to #2C2C2C and make sure it is semi-bold and change the size to M. See how that looks.

packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte
```diff
@@ -16,13 +18,25 @@
...
+  import Logo from "assets/bb-space-man.svg"
 
   // Cache previous values so we don't update the URL more than necessary
   let previousType
   let previousAsset
   let previousComponentId
   let hydrationComplete = false
 
+  // Manage the layout modal flow from here
+  let modal
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte:33]: It might be a good idea at this stage to create a new file that's in charge of the screen creation process, and remove all these modals from the `_layout.svelte` file. That file can be in control of showing and hiding modals and it can contain the various state properties that are currently defined here. Basically just cut and paste all this screen creation logic and modals to a single file.

Since there are now multiple modals involved in the screen creation process, I don't think it's very good practice to pass around references to all the modals to each of them and let them be in control of showing and hiding each other. It makes maintaining it very difficult since there's no obivous way to see the control flow between them, and also couples them together strongly.

If it's too much work to refactor to remove the modal references being passed in to each other then that's alright I guess, but I think it would be a much cleaner solution if the modals themselves didn't have to be aware of each other, and the controlling file was the one in charge, and could show and hide the correct modal and use callbacks from the other modals to keep track of state.

packages/builder/src/components/design/NavigationPanel/NewScreenModal.svelte
```diff
@@ -1,117 +1,117 @@
...
 
-  let name = ""
-  let routeError
-  let baseComponent = CONTAINER
-  let templateIndex
-  let draftScreen
-  let createLink = true
-  let roleId = $selectedAccessRole || "BASIC"
+  const blankScreen = "createFromScratch"
+  $: blankSelected = selectedScreens.find(x => x.id === blankScreen)
```
aptkingston, line 12: Can this logic just be simplified to "is there one screen or not"? If you choose blank, then there's only one screen - if you chose one or more tables then there's always at least 3 screens right? I just saw this in a few places and thought we could simplify it.

packages/builder/src/components/design/NavigationPanel/ScreenNameModal.svelte
```diff
@@ -0,0 +1,52 @@
...
+      screen =>
+        screen.routing.route.toLowerCase() === url.toLowerCase() &&
+        screen.routing.roleId === roleId
+    )
+  }
+</script>
+
+<ModalContent
+  size="M"
+  title={"Enter Details"}
```
aptkingston, line 38: We should probably lowercase the "Details" here to be consistent with the other modals.

aptkingston (commented):
Good stuff! Few comments added.

On the design side of things - the "add screens" modal looks like its got a bit too much padding to me. Everything just looks a bit big and spacious. I think we need to reduce the spacing between the titles ("blank screens", "autogenerated screens") and reduce the spacing between the table options themselves. The other modals look good to me.

At some point (not right now) we should make a BBUI spectrum component for this style of selection list if we're going to keep using it, because that's now 2 places (here and the template selection modal when creating an app) that we use it.

packages/builder/src/components/design/NavigationPanel/NavigationSelectionModal.svelte
```diff
@@ -0,0 +1,180 @@
...
+    for (let screen of screens) {
+      await saveScreens(screen)
+    }
+
+    let navLayout = cloneDeep(
+      $store.layouts.find(layout => layout._id === "layout_private_master")
+    )
+    navLayout.props.navigation = selectedNav
+
+    await store.actions.routing.fetch()
```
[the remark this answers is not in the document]
PClmnt: Yea, i think that fetches the routes so they can be clicked on within the builder itself? It was present previously and appears in other areas where the routes are being updated.

packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte
Comment (PClmnt) [reply packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte]: updated this to be controlled from ScreenWizard.svelte, it's much cleaner.

PClmnt:
@aptkingston Let me know if i've missed anything here, would be good to get this in and used for a couple of days in staging, as its pretty front and centre of things.

aptkingston:
Nice improvements Peter 👍 I'm not sure how I feel about mutating the navigation layout whenever autoscreens are created, since there's no indication that doing it will affect all your other screens too, but I think the upcoming changes to the design screen layout (role changes, nav bar position as just a setting of every screen) will help that anyway. So I'm happy to get this in now and we can start having a play with it!

aptkingston (approved):

shogunpurple (approved):
LGTM!

This looks really nice - much cleaner than before and it will make it so much faster to create autoscreens.

PR #3338, opened 2021-11-11
Query RBAC, query and automation threading/timeouts

mike12345567:
## Description
This provides functionality to run queries and automations in a separate thread, with the goal being that we can control runaway processes - for example a user makes a very large data request, this can cause lag in the main API thread. It also provides RBAC control for queries.

_EDIT - this also includes a fix for the Redis issue we've seen where CRON events cause the redis store to fill up (by default `completed` events aren't removed from Bull) so I added a cleanup loop that occurs every minute to check if there are any events in the queue that have completed and remove them._ 

The RBAC has been introduced as a dropdown option when creating a query, this will default to `BASIC`.

This also includes two minor bug fixes:
1. When querying SQL tables for a date it was possible to get an error scenario with dates, as the lucene searching used ISO dates that were invalid when converted to JS dates - it catches these in the SQL implementation and converts them to something valid.
2. Fixes an issue where deleting a datasource wouldn't delete the queries, meaning they'd show up in automations as an option that could be selected for querying.

To do this I have re-introduced the `worker-farm` dependency that we used to make use of for automations, it was removed because it was deemed too complex to use for running tests, however after some testing I've found that it works quite well when configured differently.

Previously it was configured to use as many threads as it saw fit to complete automations and it would kill/restart them whenever it wanted, this caused some initialisation cost. Instead now we instance the worker farm and auto start threads immediately on startup of the server - we also limit the number of threads to a specific count. Right now we simply use a single thread for queries and a single thread for automations with the `autoStart: true` parameter. This gives us an awful lot more control over the resource usage of them and makes sure there is no init cost other than at startup.

There is only caveat to this right now, which is that with the introduction of Typescript threads in development can be sort of slow to start, if the server has *just* started the first call can take a few seconds to complete, however after this there will never be any init carried out again; this also only applies to development as in production we are running the built, production JS bundle which starts instantly.

Right now I haven't set a timeout for the automation runs (due to the fact that they could have a delay in them) but I have set a timeout for queries, at 10 seconds. The main focus of this feature however is making sure that the main server thread is never locked up trying to process something needlessly large/complex, automations and queries are the only two sources where a user can have the system make a request for a silly amount of data.

## Screenshots
RBAC control for queries
image


shogunpurple (approved):
LGTM! Some great stuff in here on top of the RBAC - this will make automations way more efficient. I think we should see a pretty nice drop in resource usage in prod from this

Issue #3310, opened 2021-11-10
e-mail invitation from `<tenant host>.budibase.app` points to `localhost:10000`


gvlx:
**Describe the bug**
In the budibase cloud app, user onboarding using e-mail sends a link to `localhost:10000` instead of the budicloud host address `<tenant host>.budibase.app`.

**To Reproduce**
1. Go to 'bb' (upper-left icon)
2. Click on 'Manage -> Users -> Add user' or `https://<tenant host>.budibase.app/builder/portal/manage/users`
3. Choose 'Email onboarding' and provide the user's e-mail.
4. Click 'Send'
5. The user will receive an invitation e-mail with a link to `http://localhost:10000/builder/invite?code=<code>&tenantId=<tenant name>`
6. Replacing `localhost:10000` with `<tenant host>.budibase.app` works.

**Expected behavior**
The link should point to `http://<tenant host>.budibase.app/builder/invite?code=<code>&tenantId=<tenant name>`

Rory-Powell:
Thanks @gvlx

I've tried to reproduced this but I get an email with the correct link:
<img width="639" alt="Screenshot 2021-11-10 at 11 46 34" src="https://user-images.githubusercontent.com/8755148/141107813-e18b606c-a271-421d-a972-a22784d84803.png">

To rule something out, do you have custom SMTP enabled?

Rory-Powell:
@gvlx what is the tenant id you are using in budibase cloud? I can check to see if some global settings are being overwritten

Rory-Powell:
I see the issue, making changes to the organization config (updating name, logo etc) is creating a config with the default value. I'll get a fix prepped

gvlx:
Thank you.

FYI, yes, I've setup a custom SMTP for this, the GoogleApps SMTP for my organization.

Issue #3295, opened 2021-11-09
Google OAuth behind proxy


jwillmer:
I'm using the docker-compose deployment of budibase but additionally I have a reverse proxy in front which handles certificates. Budibase is reachable via `https` and the https URL is configured as the platform URL. But when setting up Google OAuth I can't set the full Budibase return URL so that Budibase is requesting a redirect to http. Google does not like that and returns `Error 400: redirect_uri_mismatch`

![image](https://user-images.githubusercontent.com/1503577/140915850-4319a045-84b0-495c-b6f4-652754e9b0c1.png)

![image](https://user-images.githubusercontent.com/1503577/140915915-0e1dd20f-92d4-4b2a-97d8-48dc26c59c39.png)

Rory-Powell:
@jwillmer does the url configured in google for the callback use http or https? I can't see any other reason for the http redirect, given that you have access to budibase over https. The base url you use in google should match the url you visit budibase on the web

jwillmer:
This is the config from the Google side:
![image](https://user-images.githubusercontent.com/1503577/140957788-6a1a279f-c80b-4f53-aef1-3d10e87c5a86.png)

jwillmer:
Fyi I have another instance with the same setup working normal on version `Current Version: 0.9.119`. The only visual difference is that I see the full callback url on the old version but not on the new.

![image](https://user-images.githubusercontent.com/1503577/140960266-2d9e10f2-16e6-4e20-abe7-e9017cbf7693.png)

Rory-Powell:
I believe the issue is due to the url being inferred by the request from the proxy to the budibase api, which is being made over http. I will prep some work to honour the platform url when it is set, in the meantime you should be able to use http in the google config instead as a workaround

jwillmer:
Google does not support `http` callbacks: 
![image](https://user-images.githubusercontent.com/1503577/140999425-80453efa-3977-47f2-afe2-94ede6257e4f.png)

Rory-Powell:
@jwillmer a fix has been added to this PR: https://github.com/Budibase/budibase/pull/3354
In the meantime you will be able to override the `http` callback by editing the config in the database. 

Visit http://localhost:10000/db/_utils/#database/global-db/config_google and update the `callbackURL` field to be the full https url you wish to redirect to

[...]

Issue #3234, opened 2021-11-03
Cannot update logo in Budibase Cloud


mjashanks:
I am unable to update the logo on my Budibase Cloud account. It seems to save correctly, but doesn't change the logo and doesn't retain the image in `Organisation` on page refresh.

It is working ok on self-host.

`POST api/global/configs/upload/settings/logoUrl` returns a `500`, with message:

```json
{"message":"Cannot read property 'replace' of undefined","status":500}
```

pavel-orlov:
I cannot change the logo on my self-hosted deployment v0.9.178. A while ago I changed the logo from the default "bb" and now I'm unable to change it to any different. I can only reset it to the default one but it doesn't seem to delete the custom one from the backend.

 **Behavior:**
- Deletion of the custom logo, resets the logo to default.
- Uploading any different logo, sets the logo to the initial custom logo, not the one I upload.
- Changing custom logo has no effect.

```POST api/global/configs/upload/settings/logoUrl``` returns `200`, with message:
```{"message":"File has been uploaded and url stored to config.","url":"/global/settings/logoUrl"}```

**Tried:**
- Updating `envoy.yaml` with `docker-compose restart proxy-service`
- `budi hosting --update` with yaml files update
- Restarting the host
- Making full page refresh in the browser

Rory-Powell:
@pavel-orlov I was able to reproduce your scenario. It looks to be caused by browser caching, can you try emptying your cache?

e.g. done like this on chrome
![Screenshot 2021-11-12 at 13 29 27](https://user-images.githubusercontent.com/8755148/141474876-7313c8d4-2aa2-4aef-84d0-0ff975577f7d.png)

pavel-orlov:
@Rory-Powell, thank you, that worked.

PR #3354, opened 2021-11-12
Fixes for google sso, cloud email url and cloud logo updates

Rory-Powell:
## Description
Fixes for:
- https://github.com/Budibase/budibase/issues/3234
- https://github.com/Budibase/budibase/issues/3310
- https://github.com/Budibase/budibase/issues/3295

## Cloud logo upload
The problem here was that the bucket referenced by cloud uploads was `undefined` so the upload was always failing. 
This has been updated to point to a newly created bucket `prod-budi-tenant-uploads`. This bucket works the same way as the minio `global` folder, with the addition of tenancy path prefixes to avoid name collisions. The bucket is referenced in the same way as the app uploads using a cloudfront distribution and a new dns name: tenants.cdn.budi.live

Bucket: https://s3.console.aws.amazon.com/s3/buckets/prod-budi-tenant-uploads?region=eu-west-1&tab=objects
Distribution: https://console.aws.amazon.com/cloudfront/v3/home#/distributions/E26JN515KBF91A

## Email containing localhost in the cloud
This was caused by the default value http://localhost:10000 being saved to the public settings for cloud tenants, this occurs when the name of a tenant is updated. This handling of the Platform URL has been updated to always return the tenant aware cloud url if the environment is cloud based, or to fallback to the set value / env var in self hosted. The most failsafe way of integrating this into the current and future usages was to add some overriding behaviour when fetching `Configs.SETTINGS` document to always honour the new logic. 

## Google sso issues
Currently it's unclear on the UI of the exact URL that will be used by the google callback. Previously this was a free text field that could be edited, which was then updated to be a more opinionated default path. The hostname is inferred by the incoming request hostname / protocol which is appended to the path. This causes problems for https hosts as the proxy -> worker communication always takes place over http and results in an incorrect url being used. The fix is to link the platform url to the callback urls (also done for oidc, which followed a similar behaviour), additionally the urls are now fetched from the server instead of being populated on the UI so there can be no difference and we can guarantee on the UI what will be redirected to. 

The google callback url has now been enabled for edits when there is an existing config, so that the user may delete their existing value to fallback to the new behaviour. After this takes place the URL is no longer editable. 

## Label Tooltips
Tooltip support has been added to the `Label` component to help explain the link between platform url and the callback plus behaviour when there is an existing.

## Screenshots

Platform url / tooltips on google and oidc 

https://user-images.githubusercontent.com/8755148/141477946-45b07c81-fe7e-4922-8ada-b04e80648ed4.mov

Behaviour when there is an existing callback url

https://user-images.githubusercontent.com/8755148/141477918-76c6d534-575c-42fd-9285-db124ba44ea0.mov

### DIFF
--- a/packages/auth/src/db/utils.js
+++ b/packages/auth/src/db/utils.js
@@ -1,6 +1,6 @@
 const { newid } = require("../hashing")
 const Replication = require("./Replication")
-const { DEFAULT_TENANT_ID } = require("../constants")
+const { DEFAULT_TENANT_ID, Configs } = require("../constants")
 const env = require("../environment")
 const { StaticDatabases, SEPARATOR, DocumentTypes } = require("./constants")
 const { getTenantId, getTenantIDFromAppID } = require("../tenancy")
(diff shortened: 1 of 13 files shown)


mike12345567 (approved):
LGTM! Nice changes, there's probably still something to be done around self hosted multi-tenant installs as these might not direct correctly to the right tenant - but we'd need to use a query string param like `?tenantId=${tenantId}` and that couldn't be done in the `getPlatformUrl` function, we'd either need to refactor some stuff or add that to the email template building work.

I think there's a fair amount of work in it for a very niche use case, so we're probably better to be aware of it and handle it when a problem arises.

shogunpurple (approved):
LGTM! Nice to have that backwards compatibility in there too for older installations.

Rory-Powell:
@mike12345567 thanks for pointing that out. At the minute I can see that a param is added to the self hosted / multi tenant platform url when sending the emails, the ui doesn't seem to honour it though and sends me back to the org picker. Given that this is an unsupported feature atm I think we're good here, our staging environment should be the only example of this. 

For the auth callbacks I think we are good because the tenant is always added to the path and that is honoured. If we run into a scenario where this function is being called outside tenancy context we'll get an error when calling `getTenantId` which seems like the right behaviour, it'll be up to us to ensure the context is supplied correctly then

Issue #3173, opened 2021-10-27
Allow Data Provider to render Array columns


shogunpurple:
When I'm using a nested data provide rcomponent, I want to be able to select an array based column of the parent record as a binding for the data provider, so that I can iterate over them.

**COS**
- Allow array based bindings to be available as sources for data providers, such as attachments, and multi options
- Hard code the schemas in the bindings, because we know what they are and will not change

aptkingston:
So I've tried this multiple different ways and it's not looking great. There is no nice way to get an array-like field to pretend to be a data provider, which is effectively what needs to be done. Implementing this as it's defined would require really badly hacking at the client library and how a lot of the internals work and is absolutely not worth it, I guarantee.

Can I clarify the use case here in case there is something better we can do. Like for attachments, would a gallery component not make more sense?

The only way this would work is if we were ok with adding another data provider layer, much like how it works for relationship columns. For example, lets say you want to iterate over the current user `pictures` (attachment field). Your tree would be:
> Pictures provider (data provider, with a source of `Current user.Pictures`)
>> Pictures repeater (repeater)
>>> Whatever component you want, bindings for `{{ Pictures repeater.Pictures.URL }}` etc are available

The extra data provider layer does not fetch any data fom the server, so doesn't slow anything down. But it's essential for managing the schema and other important state properties. The upside for this is that this will allow array like columns to be used with *any* component that uses data providers. e.g. you could render a chart with multi options data, or list attachment names and URL's in a table.

aptkingston:
Just to be clear, at the moment there is no way to iterate over array columns even with a data provider (other than relationships). So this is still a new feature, but the part about not needing an extra data provider is the part that needs to change.

mjashanks:
I think doing this via data provider is completely fine.

A gallery component is also good, but the other use case is when iterating a relationship that has been enriched already. I think it will be even more important when we start to handle deeper JSON objects.

Is it an easy change to also modify the data provider to also have a `[ x ]  Repeater` checkbox, that automatically iterates below it?

aptkingston:
> 
> 
> I think doing this via data provider is completely fine.
> 
> A gallery component is also good, but the other use case is when iterating a relationship that has been enriched already. I think it will be even more important when we start to handle deeper JSON objects.
> 
> Is it an easy change to also modify the data provider to also have a `[ x ] Repeater` checkbox, that automatically iterates below it?

[...]

PR #3355, opened 2021-11-12
Data block + array fields as data sources

aptkingston:
## Description
This PR introduces a new data block component and allows array-type fields to be used as data sources.

Addresses #3173.

Other small stuff:
- Drag and drop should work better with blocks
- Fixed an issue with the `literal` helper in string templates when the value isn't defined
- Cleaned up the data source list when categories have no available data sources
- Fixed settings bar not working with blocks

## Data block
The data block is basically a combined data provider + repeater. It's designed to simplify the common coupling of a data provider + repeater that's needed to fetch and iterate over some data. It also combines well with the new array field data sources, allowing you to iterate over and render contents from array fields with just one component.

Data block in action, fetching and rendering sale rows:
image

The data block is the first block that takes child components, since it needs something to render. This is just an implementation detail though, it feels and acts like a single component as all other blocks do.

image

It contains data settings and layout settings, basically the combination of the data provider settings and repeater settings.

image

## Array fields as data sources
There's a new category of data source now just called "Fields". These are any array type fields which can be iterated over. Relationships would fall into this category, but since they're already handled separately (and need enriched) they are excluded from this new section. That means that attachments and multi-select fields are the only ones which will appear here.

(In this screenshot I have a multi-options field called "Array", and an attachment field called "Pictures")
image

The schema for these datasources is fixed - attachments have `url` and `name` properties, while multi-select values are not objects at all but just plain strings. This breaks a lot of stuff, so they are treated as if they have a single property called `value` and this is what the binding will be called.

Here's an example combining the new data block with an attachment field on the current user. I'm rendering some cards and using both the name and url properties.

image
image

Here's another example using a multi-options field. I'm just printing the value of the options. Below that you can see what the binding looks like to use this. Remember "Array" is just the name of my field here.

image
image

## Combining both
Here's a quick example of combining data blocks with array types (and even nesting data blocks). In this example there is a parent data block iterating over my `Sales` table, and a child data block iterating over the `attachments` field of each sale. I've limited it to 3 rows for the sake of the example.

You can see that the tree remains extremely compact, and there are very few settings required to link everything together.

image


aptkingston:
Not sure if "Data block" is the correct name for this block. Would "Repeater block" be better? Or something else? Suggestions are welcome.
Data provider + Repeater = ?

shogunpurple (approved):
LGTM! Nice work and super obvious too because of the hard-coded schema. Agree on the "Data Block" - tbh is "Repeater Block" a bad name? Given it's just a pre-configured repeater? I think it makes sense.

joebudi:
Does this mean the repeater can be removed? If so, just call it repeater.

Issue #3118, opened 2021-10-22
Import into existing table with CSV


shogunpurple:
When I'm importing a CSV in budibase, I want to be able to select an existing table so that I can import several CSVs over time.

**COS**
- Only match existing columns with the CSV, and insert the data. Chuck unknown columns and don’t import those values

PR #3356, opened 2021-11-12
Import CSV to existing table from builder

mike12345567:
## Description
This implements the required work for #3118 - specifically for the builder. It introduces a new button that will appear on internal and data source plus tables, which opens a modal with a file drop zone for uploading the CSV (the same as internal table creation modal).

The UI changes are relatively minor for this, beyond a new button and modal and some better error handling for when the CSV doesn't provide any usable columns.

On the backend I've updated the CSV parser to look at an existing table and make sure all the schema generated is valid for the table, as well as being relatively loose about the keys in the CSV (doesn't worry about casing, matches it to the casing of the table). In the UI I've updated it so that the type can't be selected for each column, it is simply inferred based on what the column is in the table.

When the rows are being collected from the CSV the `transform` function iterates through the existing table schema and makes sure all of the rows columns are the correct casing and anything which doesn't exist is deleted.

The only piece of new work here was the external table CSV import, right now you can't create an external table from a CSV (although this could be supported with some UI work) but the import button is present for external tables and will work. To do this for SQL i'm using the Knex alternate `insert` implementation, where if supplied with an array of rows it will generate a single insert command which inserts them all to the same table like: `insert into "tasks" ("taskname") values ("assembling"), ("testing")`. This isn't perfect for very large datasets as its requesting the database to do quite a lot, for example with Postgres its recommended to use the `COPY` command instead, but there is no other better consistent method that works across all SQL implementations.

One final note, this is designed to be used in the builder right now, the one big thing that would need to be solved before it could be implemented for a client component is what to do with automations, right now this doesn't signal any `create:row` events in the backend, which would trigger automations in a production app - we probably need to have a think about this before just implementing a spam of events, realistically we probably need a `create:row:bulk` event with its own trigger.

## Screenshots
The table UI, with the new `Import` button - note the change to the export button so that it matches with the export.
image

The import modal, with some information to hopefully help explain how it works.
image

I've added an error which can occur, specifically if there are no columns which can be used, this is highly unlikely for table creation (although if a totally invalid file is provided then it will show with a different error).
image


shogunpurple (approved):
LGTM! This is great, didn't expect for it to be done for external tables too! 

For future on the client app import, I think we only have 3 choices:
- Don't fire the `row:created` event on bulk import at all as current
- Fire a specific `row:created:bulk` event as you say and account for it in the UI
- Possibly arrayify the `row:created` event and have some kind of backoff mechanism, so that we aren't just blasting automations. We'd probably need some kind of "iteration" automation block for this too though to perform an action on each record

mike12345567:
> LGTM! This is great, didn't expect for it to be done for external tables too!
> 
> For future on the client app import, I think we only have 3 choices:
> 
> * Don't fire the `row:created` event on bulk import at all as current
> * Fire a specific `row:created:bulk` event as you say and account for it in the UI
> * Possibly arrayify the `row:created` event and have some kind of backoff mechanism, so that we aren't just blasting automations. We'd probably need some kind of "iteration" automation block for this too though to perform an action on each record

Have to agree, this feels like the correct method/approach, we would need to have a play with exactly how to implement this, but making `row:created` always an array with some sort of built in automation loop seems like a good approach, but the back off is definitely necessary, could cause some real problems otherwise.

Issue #3385, opened 2021-11-15
Resource RBAC falls back to base permissions


mike12345567:
### Discussed in https://github.com/Budibase/budibase/discussions/3367

<div type='discussions-op-text'>

<sup>Originally posted by **SantoshSrinivas79** November 14, 2021</sup>
I made a simple app to test out automation and role-based access. The roles don't seem to be getting applied in the data access.

tldr: a user with `basic` role is able to see data that required `admin` access

A quick recording of the observation is shown below and also attached is the export file.

Please clarify if I've understood RBAC incorrectly. Thx!

https://www.loom.com/share/9b505b36ceac48f887ffa7c08e15ca2c
/div>

PR #3388, opened 2021-11-15
Fixing issue with RBAC always defaulting to base permissions

mike12345567:
## Description
Fixing an issue discovered in #3385 - RBAC roles worked for applying lower levels of roles, but they didn't revoke access correctly, it would always fallback to the base permissions if higher permissions were set.

All in all this change removes more than it adds because it changes the way in which the `authorized` middleware approaches RBAC.

There is basically four steps to how the middleware works:
1. Check if the user is a builder and is building an app, if they are then simply proceed (as builders can currently do anything)
2. Check the endpoint being called has a `resourceId` or `subResourceId` - if it does then retrieve all role information and see if a specific role has been assigned this resource.
3. If no resource, then check if there are any base permissions for all resources of its type
4. If no base permissions then reject users that aren't logged in, any logged in users are allowed through (they have some sort of non-public role for the app and the endpoint they are hitting has no specific authorization rules).

The big change here lies around step 2 and 3, managing resources and base permissions - to make sure if a resource has some specific permissions its not possible to fallback to the base ones.

This removes a bunch of code because the previous method worked by retrieving a list of permissions the user had and then comparing them against a list of permissions for the resource and then the base permissions, but it was better to just flip this, retrieve the list of permissions required to access the resource and then check if they existed in the users hierarchy of roles, based on the role and which ones they have inherited.

### DIFF
--- a/packages/auth/src/security/permissions.js
+++ b/packages/auth/src/security/permissions.js
@@ -131,38 +131,14 @@ exports.getBuiltinPermissionByID = id => {
   return perms.find(perm => perm._id === id)
 }
 
-exports.doesHaveResourcePermission = (
-  permissions,
-  permLevel,
-  { resourceId, subResourceId }
-) => {
-  // set foundSub to not subResourceId, incase there is no subResource
-  let foundMain = false,
-    foundSub = false
-  for (let [resource, levels] of Object.entries(permissions)) {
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM! Good job getting this fix out so quickly.

PR #3392, opened 2021-11-15
Caching app metadata to improve application fetch performance

mike12345567:
## Description
When playing around with the test/staging environment I noticed that the get `/api/applications?status=all` call was taking at least 2 seconds to complete. This is compounded by the fact that you might access the apps quite a bit and the fact that the `getAllApps` call is beginning to be used more and more.

As this data is frequently accessed, but infrequently updated (its only really updated by a builder when they are actually editting an app) it makes sense for it to be cached. The real slow down here was the fact that everytime the call was made it had to run through each of the Couch databases to retrieve a single doc, the metadata doc - which is known to be a very slow process.

I didn't change the logic much, instead making this call retrieve from a Redis cache instead - its hard to see the benefits locally as my system is quite quick, but it did go from a 400ms call to a less than 100ms call and I believe these benefits will be larger against staging/test envs.

There are five key locations that the cache state is busted for an app, these are as follows:
1. The updated at timestamp middleware, this will be the heaviest but it only occurs when a builder is working on an app - the rest of the apps won't need invalidated.
2. When an app client is reverted
3. When an app is deployed
4. When an app metadata is specifically updated
5. When an app is deleted

Even if the metadata cache is out of sync it is really only used for fetching a list of apps, where the data isn't super critical, it can be a few seconds out of data and not cause problems. The cache will also be bust on a time expiry, so there shouldn't be any chance of stale data causing problems - I believe the performance gain will outweigh anything here.

### DIFF
--- a/packages/auth/src/db/utils.js
+++ b/packages/auth/src/db/utils.js
@@ -54,6 +54,9 @@ exports.isProdAppID = appId => {
 }
 
 function isDevApp(app) {
+  if (!app) {
+    return false
+  }
   return exports.isDevAppID(app.appId)
 }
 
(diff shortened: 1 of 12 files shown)


packages/auth/src/db/utils.js
```diff
@@ -53,6 +54,9 @@ exports.isProdAppID = appId => {
 }
 
 function isDevApp(app) {
+  if (!app) {
```
Rory-Powell, line 57: What's going in with this one? Can't see a usage in the pr for `isDevApp`.

Rory-Powell (approved):
LGTM! Nice change, I noticed it was getting slow on staging as well. I added a usage of `getAllApps` recently so great to see that we'll start to get speed improvements across the board

packages/auth/src/db/utils.js
mike12345567: I added this because it was causing an error when `app` was null (cache sometimes returned a null value without an error) - but now I think about it its probably better to have an error here, since its not a production app or a dev app, its just nothing.

Rory-Powell: I agree on error here, better to handle getting into this scenario from the outside

PR #3400, opened 2021-11-16
Update for app metadata caching to improve long term performance

mike12345567:
## Description
Quick update to the app caching to improve performance even further, cache when an app doesn't have metadata/is invalid, meaning we don't need to poll the database everytime to see if the metadata doc exists.

The need for this likely needs to be investigated, we know that sometimes the app databases stick around a while after they have been deleted, usually due to the `_all_dbs` call returning them despite the fact they don't exist. Staging has this problem in particular because its been running through so many development updates, its probably seen a pile of bugs which would cause these databases to be replicated.

In any case, if an app database is totally invalid/doesn't have any metadata/useful docs within it then we shouldn't need to poll the database everytime to find this out, instead we can cache a doc in Redis to tell us its invalid and if it becomes valid through a revert/package update, then the cache will be busted and we will be made aware of this.

### DIFF
--- a/packages/auth/src/cache/appMetadata.js
+++ b/packages/auth/src/cache/appMetadata.js
@@ -67,7 +67,19 @@ exports.getAppMetadata = async (appId, CouchDB = null) => {
   return metadata
 }
 
-exports.invalidateAppMetadata = async appId => {
+/**
+ * Invalidate/reset the cached metadata when a change occurs in the db.
+ * @param appId {string} the cache key to bust/update.
+ * @param newMetadata {object|undefined} optional - can simply provide the new metadata to update with.
+ * @return {Promise<void>} will respond with success when cache is updated.
+ */
+exports.invalidateAppMetadata = async (appId, newMetadata = null) => {
+  if (!appId) {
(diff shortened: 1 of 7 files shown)


aptkingston:
Probably totally irrelevant, but my local dev version has had zombie apps for a few months. I think something went wrong when there were initially deleted, and so now they are listed like normal apps but can never be deleted (the deletion requests succeeds and 200s, but it doesn't actually get deleted). They can't be used either, as it 404's when trying to get the app definition if you open them. Don't know if this is the sort of thing you're talking about, but happy to use my bugged env as a test env if that helps.

mike12345567:
> Probably totally irrelevant, but my local dev version has had zombie apps for a few months. I think something went wrong when there were initially deleted, and so now they are listed like normal apps but can never be deleted (the deletion requests succeeds and 200s, but it doesn't actually get deleted). They can't be used either, as it 404's when trying to get the app definition if you open them. Don't know if this is the sort of thing you're talking about, but happy to use my bugged env as a test env if that helps.

Yeah staging is in the same scenario, we should clean these up, they won't do any harm but could hurt performance for very large environments.

PR #3403, opened 2021-11-16
Inline text editing + perf. enhancements + preview enhancements

aptkingston:
## Description
This PR was my last lab day project, which I've fleshed out to include a few extra improvements.

I don't want this to be merged immediately because it includes some core changes to the client library which I'll continue to test. It's worth testing though, because there's some significant gains for both app speed and designing speed.

TLDR:
- You can edit the text of headings, paragraphs, links, form field labels and buttons by double clicking them in the preview
- Performance enhancements for client apps (better memoisation of settings)
- Snappier design interface (components don't remount when settings are changed)

## Inline text editing
Various components can now be edited directly in the preview by double clicking. This changes the border to green, and focuses the field so you can begin typing. You can simply click off the field again and the text will be saved.

image

Components which can be edited inline should define `"editable": true` in the manifest, and then use the `editing` component context value to correctly render an editable version of themselves when required.

## Performance enhancements
There already was good memoisation of component settings implemented in the client library, but there was a bit of a bug with how the memoised settings were actually passed to the component instance, which resulted in props changing more than they should have (and therefore triggering a lot more svelte reactive statements, and in turn things like API calls). This has been fixed now and improved further with block support, and it should result in a decent reduction of redundant API hits (and renders, but the performance gain there is reasonably negligible).

## Preview enhancements
The client library has features built in to support running inside the builder preview. One of these features was that whenever any setting of a component was changed, the client library fully remounted the component. This had the advantage of ensuring that things never got into weird states when changing settings which the components never expect to change at runtime. It had the side effect of causing slower designing, flashing, and excess loading indicators. This was particularly noticable to me in the new promo video, where every time any setting of the card block is changed (even something simple like the title), the full block remounts, the loading indicator appears, and the data is loaded again. It's a bit jarring and seems unecessary most of the time, unless you actually changed the data setting.

This component remounting feature wasn't actually needed for most components. Most were already written to never depend on static props at runtime, and those that did have now been updated to not depend on them. This has let me remove the component remounting feature, meaning no more flashing or refetching data when editing settings. It feels much, much snappier to design apps with this removed.

I'm going to continue to test this branch for a while because these are changes to the core of the client library and could introduce bugs for some edge cases (think weird stuff like changing the schema of a form with multiple form steps at runtime via conditional UI).

## Other stuff
- Refactored the way key events were handled by the preview iframe (handle them in the client lib instead). This was already done on lab day before Marty's recent work on fixing Safari support, so I've just kept this and merged it.
- Removed the unused `iframe-loaded` event
- Don't change component settings if the value is identical. This seems to actually save quite a few API requests and helps reduce 409s.
- Fixed a temperamental race condition crash when sometimes loading the data page after a refresh


packages/client/src/components/app/forms/Field.svelte
```diff
@@ -55,16 +59,23 @@
...
     <label
+      bind:this={labelNode}
+      contenteditable={$component.editing}
+      on:blur={$component.editing ? updateLabel : null}
       class:hidden={!label}
       for={fieldState?.fieldId}
       class={`spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Form-itemLabel ${labelClass}`}
     >
-      {label || ""}
+      {label || " "}
```
mike12345567, line 78: Just wondering, why the `" "` - not sure what that does?

mike12345567 (approved):
LGTM! This is super exciting to see, love the inline editable elements, can't wait to see the performance improvements too and the reduced saving of stuff, I think this will make a big difference! Also like the `KeyboardManager` - nice central place to see what all can be done with the keyboard/what isn't in use right now.

packages/client/src/components/app/forms/Field.svelte
aptkingston: Haha well spotted. If there is no text then it basically doesn't take up any space, which screws up the table layout that fields use. It means that whenever you double click to start editing text in the preview, the label is misaligned and right up in top left of the field rather than centered. Making sure there is always at least a space seems to fix this so that's why I added it.

mike12345567: Makes sense!

PR #3418, opened 2021-11-17
Dependency audit/upgrade

mike12345567:
## Description
Thought we were long over-due for a general tidy up and upgrade of packages as some of the things we have been using have gotten quite out of date, I added a top level script `scripts/audit.js` which will use `npm audit fix` (there is no yarn alternative) to run through all the packages and update our locks throughout to try and remove as many issues as possible. There are a few things I learnt while doing this:

1. `@sveltejs/vite-plugin-svelte` currently cannot be upgraded past `1.0.0-next.19`, this is a dev dependency but I was doing a general upgrade of packages as a lot of the things we use are so bleeding edge that bugs are to be expected. If you upgrade this package to anything between `1.0.0-next.20 - 30` you'll run into an issue with `error when starting dev server: Error: Failed to resolve force included dependency:` and some specific package (usually in BBUI) breaking. This could be something to do with our config, but I couldn't find any info online and was unable to resolve it, so for now I've pinned us at the best we can use.
2. We previously had an express server running inside our koa server just to handle bull board, with the latest versions there are koa adapters which makes this no longer required - express has been removed and we purely use koa now.
3. We cannot upgrade any packages within BBUI, the frontend breaks quite a bit if we try this, generally a pile of issues will crop up with spectrum. I assume we'd need to handpick all of the versions of spectrum we're using as there is a ton of different spectrum packages, with different versions, all dependent on the underlying `@spectrum-css/vars` version.
4. With upgraded versions of vite I found that the `--host 0.0.0.0` was required, as we didn't have vite pinned it was likely that people attempting to run Budibase from a fresh install could have run into issues if they'd used most up to date versions.

In general there aren't any high/critical vulnerabilities left, apart from the odd one which is coming from dev packages where it doesn't matter (not used in prod).


Rory-Powell (approved):
Nice one. Great to remove express

Issue #3120, opened 2021-10-22
Oracle Datasource (Plus)


shogunpurple:
When I’m selecting a datasource from the datasource options in the backend UI, I want to be able to have the datasource plus experience with an oracle database.

**COS**
- Support datasource plus and fetching of tables from oracle databases

PR #3437, opened 2021-11-18
Oracle datatasource plus

Rory-Powell:
## Description
https://github.com/Budibase/budibase/issues/3120

Adding plus support for oracle
Based on https://github.com/Budibase/budibase/pull/3301

Due to the linting changes it is easiest to view the majority of the changes in this commit range
https://github.com/Budibase/budibase/pull/3437/files/1be2d006db751890ce799639d9ebcebd4fadb210
with linting and boolean support being added here
https://github.com/Budibase/budibase/pull/3437/commits/cb3495c831288a2f307697fdd70babd908ca4948

## Screenshots
Quick demo of configuration + screen gen:
https://user-images.githubusercontent.com/8755148/142425744-e93fc594-3ea3-4c77-8126-c1e4d2ae511b.mov

Relationships example:
Screenshot 2021-11-18 at 12 25 35

Screenshot 2021-11-18 at 12 25 45


packages/server/src/integrations/oracle.ts
```diff
@@ -63,64 +79,361 @@ module OracleModule {
...
+              name: tableName,
+              columns: {},
+            }
+            oracleTables[tableName] = table
+          }
+
+          let column = table.columns[columnName]
+          if (!column) {
+            column = {
+              name: columnName,
```
mike12345567, line 216: One thing that's missing here is auto-columns, e.g. any column which is generated via a computed value (like an incrementing ID column) - in the case of SQL Server I had to perform a lookup for identity and computed columns which I then used to determine if it was an auto column (https://github.com/Budibase/budibase/blob/a013c51e2bd27b69985e3b2d67147bfca81aafb7/packages/server/src/integrations/microsoftSqlServer.ts#L142) may not be possible for Oracle, but just thought I'd add the comment!

mike12345567: Ignore this sorry, missed the private function further down!

mike12345567:
Might be worth merging the Oracle PR first, making this one a bit easier to review - right now its a mixture of the two and might be easier to just see the changes required to generate the Oracle schema!

packages/server/src/integrations/oracle.ts
Rory-Powell: @mike12345567 this block is an intermediate transition format between the raw sql output and an object format for the `OracleX` interfaces above. The output `OracleTable` is used in the `buildSchema` function below which does the conversion to the budibase domain including auto column

Rory-Powell:
@mike12345567 this pr only has the new stuff, the other noise is coming from a lint that was done. 
To make it easier you can check this commit range that has everything minus the linting:
https://github.com/Budibase/budibase/pull/3437/files/1be2d006db751890ce799639d9ebcebd4fadb210
the only other change post that commit was support for booleans. I'll add this to the description

mike12345567:
> @mike12345567 this pr only has the new stuff, the other noise is coming from a lint that was done. To make it easier you can check this commit range that has everything minus the linting: https://github.com/Budibase/budibase/pull/3437/files/1be2d006db751890ce799639d9ebcebd4fadb210 the only other change post that commit was support for booleans. I'll add this to the description

Cool that makes it easier, thanks!

packages/server/src/integrations/oracle.ts
```diff
@@ -63,22 +68,250 @@ module OracleModule {
...
+              // match the order of the columns in the db
+              .sort((c1, c2) => c1.id - c2.id)
+              .forEach(oracleColumn => {
+          const columnName = oracleColumn.name
+          let fieldSchema = table.schema[columnName]
+          if (!fieldSchema) {
+            fieldSchema = {
+              autocolumn: this.isAutoColumn(oracleColumn),
+              name: columnName,
+              type: convertType(oracleColumn.type, TYPE_MAP),
```
mike12345567: One thing I've been trying to do is just compress all of the SQL types into a single type map definition, as per the `convertSqlType` and `SQL_TYPE_MAP` in: https://github.com/Budibase/budibase/blob/a013c51e2bd27b69985e3b2d67147bfca81aafb7/packages/server/src/integrations/utils.ts#L10

Rory-Powell: no worries! I will get this rebased from the latest changes in the mssql stuff, thanks for pointing that out

mike12345567 (changes requested):
Looks good - really like the use of interfaces for the Oracle responses, makes it very clear what to expect! Just a couple of comments related to some of the stuff I changed around SQL Server, might need to address the point about returning from Oracle otherwise I think automations/relationships might break with it.

germanykurd (approved):

aptkingston:
Awesome stuff Rory. The pain of spinning up an oracle instance was worth it 😄

shogunpurple (approved):
LGTM - great work on this one. Seems like oracle is a bit of a beast to work with.

packages/server/src/integrations/oracle.ts
```diff
@@ -104,24 +337,39 @@ module OracleModule {
...
     }
 
     async delete(query: SqlQuery | string) {
-      const response = await this.query(getSqlQuery(query))
+      const response = await this.internalQuery(getSqlQuery(query))
       return response.rows && response.rows.length ? response.rows : [{ deleted: true }]
     }
+
+    async query(json: QueryJson) {
+      const operation = this._operation(json).toLowerCase()
```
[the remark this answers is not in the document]
Rory-Powell: @mike12345567 thanks! will take a look - going to merge into the base query pr and do modifications / testing there

PR #3441, opened 2021-11-18
Assorted bug fixes

PClmnt:
## Description

Fixes a couple of bugs:

- We now save user test data within the automations screen. So if you navigate away and come back, you are able to pick up with your testing where you left off
- fixed an issue where the automation blocks were breaking because of svelte animation issues
- in the data tab it was possible to edit a datasource plus table, you shouldn't be able to do that, and it put the table into a weird state.


Rory-Powell:
> in the data tab it was possible to edit a datasource plus table, you shouldn't be able to do that, and it put the table into a weird state.

Is this anything to do with the changes you put in @mike12345567 or unrelated?

mike12345567:
> > in the data tab it was possible to edit a datasource plus table, you shouldn't be able to do that, and it put the table into a weird state.
> 
> Is this anything to do with the changes you put in @mike12345567 or unrelated?

Sort of, the way that renaming tables is implemented is using some very specific property on the body of a table save (we should probably break it out into a separate endpoint or something like that) - previously saving an external table would just break, but with my changes it just didn't know what to do with the properties so would ignore it/try to update the table name in Couch schema which would cause problems. Spoke to @PClmnt about it and thought it was best to just disable the option for external tables as it would be difficult to implement a table rename across the board - should probably also add a check on the backend to reject it if a request comes in trying to do it.

packages/builder/src/components/backend/TableNavigator/popovers/EditTablePopover.svelte
```diff
@@ -66,7 +66,9 @@
   <div slot="control" class="icon">
     <Icon s hoverable name="MoreSmallList" />
   </div>
-  <MenuItem icon="Edit" on:click={editorModal.show}>Edit</MenuItem>
+  {#if internal}
+    <MenuItem icon="Edit" on:click={editorModal.show}>Edit</MenuItem>
+  {/if}
   <MenuItem icon="Delete" on:click={showDeleteModal}>Delete</MenuItem>
```
aptkingston, line 72: Should we also hide the delete button? In my testing there, trying to delete a datasource plus table does nothing but says it's been deleted. It actually sends a request to a URL ending in `/undefined` which the server 500's on, but the front end says success.... I guess the correct thing to do would be to remove it from the list (like un-import it) but we aren't actually deleting it.

mike12345567: Deletion should work, that is implemented for data source plus tables, I can look at the specific scenario that caused a 500 - the `/undefined` is because of the way our delete endpoint works, its `/:id/:rev` and plus tables don't have a `_rev` so they just send undefined, but it does still work (usually).

PClmnt: ^ Lol i'll revert the commit I just pushed

PClmnt: @aptkingston @mike12345567 Is this okay now then?

mike12345567: Yeap looks good to me!

mike12345567 (approved):

packages/builder/src/components/backend/TableNavigator/popovers/EditTablePopover.svelte
aptkingston: Fire away!

Issue #2420, opened 2021-08-19
Issues displaying data from table when using _id as column name & using foreign keys


MarkOSullivan94:
**Describe the bug**

We are having an issue displaying data from table in the `Data` tab section of the Budibase app

The table data structure can be seen below
![Screenshot 2021-08-19 at 23 00 30](https://user-images.githubusercontent.com/6950843/130150802-85fee030-0142-42d4-9257-73539b5a4693.png)

Only one record is shown as you can see from the screenshot below

![Screenshot 2021-08-19 at 23 20 14](https://user-images.githubusercontent.com/6950843/130151817-da86c347-2aeb-4c19-ad62-738f692e6e2b.png)

For comparison here is how our table looks in the DB within TablePlus

![Screenshot 2021-08-19 at 23 21 44](https://user-images.githubusercontent.com/6950843/130151941-03b36658-8fd8-4ba5-86c6-1cb2e2ffc15c.png)

**To Reproduce**
1. Create a table within your Postgres DB with a similar data structure to the one shown above which uses foreign keys being the `_id` column in other tables & populate it with more than 2 records
2. Follow the steps needed to add Postgres DB & connect datasource to Budibase 
3. Click on the button `Fetch tables from database`
4. Click on the newly created table from the sidebar
5. Now the table should open and not be showing all the records

**Expected behavior**

I expect to be able to see all the records from our table

**Desktop (please complete the following information):**
 - OS: Mac OS 11.2.3
 - Browser: Brave
 - Budibase Version: 0.9.107

mike12345567:
Hi @MarkOSullivan94 - this is due to the fact that `_id` and `_rev` fields are reserved within Budibase right now, the way we have implemented SQL data sources into Budibase is to replicate patterns used by our CouchDB system, where `_id` always contains primary key attributes.

I will see if there is someway to get around this, but for now if its possible to change the column name to something other than `_id` this will likely begin to work for you.

PR #3444, opened 2021-11-18
Disallow _id, _rev and tableId fields for SQL tables

mike12345567:
## Description
Fix for #2420 - don't allow columns which are considered invalid/not allowed - tables are ignored and error is shown.

## Screenshots
image


aptkingston (approved):
LGTM! Nice and clean. Should we use `\n` rather than ` - ` as the error separator in `updateError` if the error already exists? That would maybe pretty print it slightly better in the front end since they'd be on separate lines.

mike12345567:
As per @aptkingston updated to use newlines, the inline alert component handles newlines as different message elements.

image

Issue #3174, opened 2021-10-27
Multi-Filter Component


shogunpurple:
When I'm creating components in the design section of the builder, I want to be able to have access to a multi-filter component that allows me to apply multiple controllable filters at once, so that I have more control over how my data is filtered in the UI, rather than setting it all up manually in the design UI by binding data provider arguments.

**COS**
- Add multi filter component that has a similar UI to the data provider configuration UI
- Allow data providers to be bound to it directly

**Notes**
- This will be part of a filtering block in future

PR #3476, opened 2021-11-22
Dynamic filter component

aptkingston:
## Description
This PR adds a new dynamic filter component. This component behaves similarly to how filters are configured in the builder - except it's inside client apps. It lets you update data provider filters and search for data on the fly.

Addresses #3174.

## Dynamic filter component
The dynamic filter component lives inside a new component section called "Data". It is a button which opens a modal when clicked and allows filters to be defined. The settings bar is supported, as is inline editing of the button text. The button text shows how many filters have been configured, so you know whether the data you're looking at is filtered or not.

The component takes the following settings:
- Which data provider to filter
- Which columns are allowed to be filtered (this is important, as you may not want some columns to be filtered on as this overrides any filters defined in the builder)
- Button settings (text, variant, size, quiet theme)

Dynamic filter settings:
image

When clicked, the filter modal is opened. In here you can define filters the same as you would define them in builder settings. The main difference is that handlebars and data bindings are not available, since it wouldn't make sense at this level.

The filter modal:
image

Technical note: the modal is its own component, and not just imported from the builder. Importing the builder one directly leads to too many builder dependencies being pulled in and doesn't build due to the rollup alias plugin. I don't think this is a bad thing anyway, as they have some significant differences like fully disabling HBS and data binding support.

The button text after adding filters:
image

## Other stuff
- Component structure reorganised. New sections for "Data" and "Layout" have been added. Button has been moved into "Elements"
image
- Add support in general for modals in client apps
- Fix issue with client apps resetting too much state when settings are changed
- Fix issue with some undefined manifest properties not being coerced into boolean types
- Fix issue with inline text editing getting into a bad state after deleting all text while editing inline
- Trim text when editing inline to fix newlines sometimes appearing when they shouldn't
- Enhance the internal client library API for extending data provider filters to support multiple filter extensions by the same component


shogunpurple (approved):
LGTM! Good job. I thought originally that the multi-filter component would be inline - but this is actually a much better idea - it's way more flexible given the fact you bind it to a data provider and use it to control the filtering from inside. Nice one

aptkingston:
> LGTM! Good job. I thought originally that the multi-filter component would be inline - but this is actually a much better idea - it's way more flexible given the fact you bind it to a data provider and use it to control the filtering from inside. Nice one

Cheers! Since the core component is done now, we can do stuff like include in a block however we want. Having something like a "table with dynamic filter" block would probably be a good call, since that will effectively feel like just a table with an inline filter.

Rory-Powell (approved):
Looks great!

joebudi:
This is awesome. 

Did you consider using a filter icon instead of a button? 

Using a filter icon would mean you don't need the additional button options making the settings look more cleaner. From an end-user perspective, this is sort of what i expect to see, and if there are additional ctas, it won't look as busy.

aptkingston:
> This is awesome.
> 
> Did you consider using a filter icon instead of a button?
> 
> Using a filter icon would mean you don't need the additional button options making the settings look more cleaner. From an end-user perspective, this is sort of what i expect to see, and if there are additional ctas, it won't look as busy.

I had an icon in the button initially, but an icon by itself would be a bit lost would it not? I'll make some pics here of the different options.

joebudi:
> > This is awesome.
> > Did you consider using a filter icon instead of a button?
> > Using a filter icon would mean you don't need the additional button options making the settings look more cleaner. From an end-user perspective, this is sort of what i expect to see, and if there are additional ctas, it won't look as busy.
> 
> I had an icon in the button initially, but an icon by itself would be a bit lost would it not? I'll make some pics here of the different options.

https://spectrum.adobe.com/page/icons/ - properties icon is a nice icon for this

aptkingston:
Here are a few different variations. Last one is the properties icon size XL.

Untitled-1

Personally I see the properties icon as more like a settings icon (but I do like it), but that might just be me. I don't really like the filter icon which is why I didn't include it in the button for this PR. The other thing to consider is that if it's just an icon we can't append the extra text to show filters are in use. We could probably do something different like change the icon to blue if filters are in effect?

joebudi:
My thinking is there will be more options, and it will get a little busier. A quiet button with an icon will probably suffice - what do you think?

Also, when filtering tables, cards, charts, the data changes as I filter and I can quickly see the filtered data, change it or remove the filter. Does the modal allow this?

aptkingston:
> My thinking is there will be more options, and it will get a little busier. A quiet button with an icon will probably suffice - what do you think?
> 
> Also, when filtering tables, cards, charts, the data changes as I filter and I can quickly see the filtered data, change it or remove the filter. Does the modal allow this?

The data updates as soon as you close the modal yea. You can open the modal, quickly add, change or remove filters than then save the filters and the data will update. It doesn't matter if it's tables or cads or charts - the filter is done  at the data provider level so it will work with all of them.

About the filter button again - it's important that the core component has settings on it to allow it to be quiet or not, and to choose the text or variants, because we always allow customisation at the core component level. If we use this in a block, then at that stage we can make it opinionated and always quiet for example. I don't mind making it quiet by default though if you want to do that. If we add an icon to it then I will have to disable being able to edit the text inside the preview, so that's the only downside.

aptkingston:
Keen to get this merged so please let me know what way you want to go. Am I right in thinking default to quiet, add the icon ("Properties" one?) and then disable the inline editing of text as a consequence.

joebudi:
Sounds good.

aptkingston:
The button is now always a quiet, secondary button with the "Properties" icon. The only setting I left was size (and button text of course).

It now looks like this:
image

When filters are applied, I've updated it to turn blue (or whatever the primary color is) when active, rather than adding text in brackets. When active, it looks like this:
image

Inline editing of button text has had to be disabled due to these changes.

@joebudi you happy enough with these changes?

joebudi:
Looks good to me sir. Good job. If it were up to me, I would remove the free type button text.

PR #3477, opened 2021-11-22
Table/column deletion checks and disabling fetched external table deletion

mike12345567:
## Description
This PR brings an update to the UI to make sure the user is aware of the risks when deleting a table or a column, making sure that the user is aware of the risk attached with deleting them. For now I have also disabled the ability to delete tables which weren't created by Budibase in external SQL tables as this has quite a bit of risk attached to it as well. If we want to enable this again it is a very simple change, but this just makes sure that we are aware of which tables were created through Budibase and which ones were imported.

## Screenshots
Table deletion:
image

Column deletion:
image


aptkingston (approved):
LGTM! Nice one, much safer now 👍

Issue #3487, opened 2021-11-23
Issues with screen creation wizard


aptkingston:
Spotted a couple of issues when trying the new screen creation wizard:
- Some colours are static, so some text doesn't work in dark mode
![image](https://user-images.githubusercontent.com/9075550/142994632-086144d8-fa3f-4414-b5d3-9c4b66a08523.png)
- Once you select blank, even if you unselect it you can never select autoscreens, and vice versa
- If you select blank, then unselect and try to click autoscreens (which fails as described above), you can never click anything again
- If you select an autoscreen table, then unselect it and select blank, it will fail and also cause a console error `Uncaught TypeError: table is undefined`
- You can't choose more than one autoscreen table (was this by design? I thought we could create screens for multiple tables at once)
- If you select something, then close the modal and reopen it again, the selection is still there and hasn't been cleared
- We should probably hide the users table from the list of tables. We don't really want users to be directly edited in client apps

Most of these just seem to stem from some state issues when selecting and deselecting.

Mitch-Budibase:
I've reviewed and verified the above points with one exception:

- If you select something, then close the modal and reopen it again, the selection is still there and hasn't been cleared

This is still the case. Will this be updated?

PR #3488, opened 2021-11-23
Remove nav selection modal from screens UI (for now)

PClmnt:
## Description
Talking to @joe and @aptkingston about this, and we decided that it's probably best for now to remove the navigation selection modal until the refactor of the design UI has been done, because the only way to set the navigation currently is to mutate the main layout which isn't really that great.. i.e it will change the layout for any screen using it. 

Also refactored the save functions into the controller file which tidies things up a good bit.


packages/builder/src/components/design/NavigationPanel/ScreenWizard.svelte
```diff
@@ -32,17 +110,12 @@
...
+  <ScreenDetailsModal bind:screenName bind:url {save} {chooseModal} />
 </Modal>
 <Modal bind:this={navigationSelectionModal}>
-  <NavigationSelectionModal
-    bind:url
-    bind:screenName
-    bind:selectedScreens
-    {chooseModal}
-  />
+  <NavigationSelectionModal {chooseModal} {save} />
```
aptkingston, line 120: We may as well just fully remove this code that references `NavigationSelectionModal` at the moment since if we include it like this then the file will still be bundled with the builder even though it isn't needed.

aptkingston (commented):
Nice changes! I opened an issue about the same time that this PR was opened, detailing some bugs with the wizard stuff, so let's just double check that all those issues are fixed before we merge this PR in.
https://github.com/Budibase/budibase/issues/3487

Looks like the core issue is fixed here though - the reassignment of `templates` looks like it was probably an issue previously.

PClmnt:
> Nice changes! I opened an issue about the same time that this PR was opened, detailing some bugs with the wizard stuff, so let's just double check that all those issues are fixed before we merge this PR in. #3487
> 
> Looks like the core issue is fixed here though - the reassignment of `templates` looks like it was probably an issue previously.

@aptkingston Haha you're too quick for me was trying to get those fixed in this before anybody noticed :D But yea, the templates re-assignment was causing pretty much all of these. You should be able to create screens from multiple tables now etc.

aptkingston (approved):
Cool, lets get this in then if that's those various issues fixed.

Issue #3485, opened 2021-11-23
SQL - Relationships need ability to specify where to pull data


lcslouis:
**Describe the feature request**
In 1 of my projects I have multiple tables that are lookup tables static tables that contain a selection choice.

Depending on the field I will have my static table have columns 
ID, OptionValue, OptionName

Or they are just ID, LongTitle, ShortTitle

In the case of the First option I have to create a view in mysql so that the OptionValue is used instead of the ID field. 

the realtionships settings need the ability to designate a different column Other than the ID Primary Key.

marco-santulli:
I would like to piggyback on this enhancement for a request.  At the moment it is not possible to specify a relationship on the same table, the UI prevents you from specifying the same table as source and target.  However it is quite normal to have father-son relationship on the same object (like contractor / subcontractor relationship).

mike12345567:
Hi @marco-santulli - this likely needs to be a separate issue as first we need to make sure Budibase internal tables are capable of it, then we can make other data sources capable.

marco-santulli:
Hi Mike,

see what you mean, I will create a separate one, thanks

Marco

PR #3491, opened 2021-11-23
Quick change for SQL existing relationships - allow picking both keys

mike12345567:
## Description
Quick fix for #3485 - default the key to the primary key but allow another field to be selected, just making the UI flexible enough to represent all types of SQL relationships.

## Screenshots
Adding a field just below the from table selection to pick the key in the from table, this defaults to the primary key of the table.
image

### DIFF
--- a/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
+++ b/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
@@ -71,6 +71,9 @@
     if ($touched.toCol && !toRelate.name) {
       errors.toCol = colNotSet
     }
+    if ($touched.primary && !fromPrimary) {
+      errors.primary = "Please pick the primary key"
+    }
     // currently don't support relationships back onto the table itself, needs to relate out
     const tableError = "From/to/through tables must be different"
     if (fromTable && (fromTable === toTable || fromTable === throughTable)) {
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -95,6 +98,12 @@
     return errors
   }
 
+  $: isManyToMany =
+    fromRelationship?.relationshipType === RelationshipTypes.MANY_TO_MANY
+  $: isManyToOne =
+    fromRelationship?.relationshipType === RelationshipTypes.MANY_TO_ONE
+  $: fromPrimary =
+    !fromPrimary && fromTable ? fromTable.primary[0] : fromPrimary
```
aptkingston, line 106: I have no idea how svelte handles this lol. Is this not like a recursive assignment? Since the reactive statement references itself? And then it's a reactive assignment but also bound to a form field. I'm guessing it works since the PR is up but the look of it scares me 😆 

Personally I would make `fromPrimary` just be a normal variable, then you can have a reactive snippet to set it to something else when required.

So like:
```js
let fromPrimary
$: {
  if (!fromPrimary && fromTable) {
    fromPrimary = fromTable.primary[0]
  }
}
```

aptkingston (commented):
Nice change, but help the svelte compiler out and do my wee refactor 😆

shogunpurple (approved):
LGTM! Agree with cheeks suggestion too - just to make things a bit more easy to understand 👍 I think it works because you can actually reassign reactive variables in svelte and your condition catches it with the `!fromPrimary`

aptkingston:
> LGTM! Agree with cheeks suggestion too - just to make things a bit more easy to understand 👍 I think it works because you can actually reassign reactive variables in svelte and your condition catches it with the `!fromPrimary`

Was curious so did some testing in the svelte REPL. Statements don't cyclically invalidate themselves, and if you make a cyclic reassignment between 2 variables then the svelte compiler catches it and warns you. Pretty cool.

packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
mike12345567: Thanks for suggestion, yeah wasn't sure how this would work but it seemed to work fine so left it in - will change to your suggestion.

Issue #3140, opened 2021-10-25
JS Support for Formula Fields


shogunpurple:
When I’m adding a formula column to the data UI, I want to be able to use JavaScript so that I can perform more sophisticated calculations.

**COS**
- Implement JS Binding drawer in formula UI
- Leverage existing JS Binding support
- Refactor binding drawer implementation by consolidating the drawers

PR #3508, opened 2021-11-23
JS for formula fields + binding panel refactor

aptkingston:
This PR actually addresses one of the pieces of technical debt which I wanted to address - refactoring the "binding panels" into a single component that can be used for both client bindings and "server" bindings (formula fields, automations etc).

Addresses #3140.

I've then also enabled JS for formula fields. Going forward, we can now enable JS for *any* place we use HBS/data bindings by simply passing in the `allowJS` prop. This means we could enable full JS for all automation parameters if we wanted. I haven't done that in this PR, but it's now just a setting change if we want to.

As usual, code hints work and can be enabled with ctrl+space.

Writing JS for a formula field:
image

The column config shows the same helper text of `(JavaScript function)`:
image

And here's the resulting data working correctly for the above expression:
image

### DIFF
--- a/packages/bbui/src/Form/Combobox.svelte
+++ b/packages/bbui/src/Form/Combobox.svelte
@@ -6,6 +6,7 @@
   export let value = null
   export let label = undefined
   export let disabled = false
+  export let readonly = false
   export let labelPosition = "above"
   export let error = null
   export let placeholder = "Choose an option or type"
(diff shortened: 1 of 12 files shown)


mike12345567 (approved):
LGTM! So happy to see the binding panels cleaned up, was really a nightmare updating anything to do with them - very happy to see them merged!

PClmnt (approved):
This is brilliant @aptkingston  

As @mike12345567 said that whole area was so disjointed, low key one of the best PRs!

PR #3524, opened 2021-11-24
Improving error output of create/edit column modal

mike12345567:
## Description
Adding more error tracking around column creation, to reduce confusion/users not understanding required settings. Previously there was no error handling here, it would simply check if it was valid and disable the button if it wasn't - but with no feedback. Also outputting errors when saving the column from the backend, rather than just dumping the user back to the table selection page.

This is a fix for #3512.

## Screenshots
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -276,7 +276,6 @@
       const relatedTable = $tables.list.find(
         tbl => tbl._id === fieldInfo.tableId
       )
-      console.log(relatedTable)
       if (inUse(relatedTable, fieldInfo.fieldName)) {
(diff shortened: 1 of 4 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -258,6 +257,32 @@
...
+        ", "
+      )} are not allowed as column names`
+    } else if (inUse($tables.draft, fieldInfo.name, originalName)) {
+      newError.name = `Column name already in use.`
+    }
+    if (fieldInfo.fieldName && fieldInfo.tableId) {
+      const relatedTable = $tables.list.find(
+        tbl => tbl._id === fieldInfo.tableId
+      )
+      console.log(relatedTable)
```
Rory-Powell, line 279: looks like a wee missed log statement

Rory-Powell (approved):

Issue #3531, opened 2021-11-24
Saving form with a related row with a different type of primary key results in an error


Rory-Powell:
**Describe the bug**
Using the following DDL:

```sql
CREATE TABLE REGIONS (
	REGION_ID INTEGER,
	REGION_NAME VARCHAR(25),
	CONSTRAINT REG_ID_PK PRIMARY KEY (REGION_ID)
);


CREATE TABLE COUNTRIES (
	COUNTRY_ID CHAR(2),
	COUNTRY_NAME VARCHAR(40),
	REGION_ID INTEGER,
	CONSTRAINT COUNTRY_C_ID_PK PRIMARY KEY (COUNTRY_ID)
);

ALTER TABLE COUNTRIES ADD CONSTRAINT fkregions FOREIGN KEY (REGION_ID) REFERENCES regions(REGION_ID);


INSERT INTO REGIONS
(REGION_ID, REGION_NAME)
VALUES(1, 'Europe');
INSERT INTO COUNTRIES
(COUNTRY_ID, COUNTRY_NAME, REGION_ID)
VALUES('UK', 'United Kingdom', 1);
```

With the corresponding relationship
<img width="514" alt="Screenshot 2021-11-24 at 17 14 41" src="https://user-images.githubusercontent.com/8755148/143284748-4554d53a-22d9-4d22-be7b-c206859d1578.png">

Produces this error on save / edit of the countries entity

https://user-images.githubusercontent.com/8755148/143284797-1c3616e5-3f40-4b99-aa19-97b1d27d86f7.mov

This is caused by a misconfiguration in how we are querying for relations here
https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/controllers/row/ExternalRequest.ts#L442

e.g. the following config 
![Screenshot 2021-11-24 at 17 17 12](https://user-images.githubusercontent.com/8755148/143285413-68a3df15-bbf9-4a7a-b614-a2257ffec058.png)

Results in a query being made with a mismatched primary key to value. 
![Screenshot 2021-11-24 at 17 21 14](https://user-images.githubusercontent.com/8755148/143285731-f99d9404-265e-471c-8c57-b2fe3bb5fa73.png)

Note that the value 'UK' is being provided for the 'region_id'. 
The 'UK' primary key belongs to the 'countries' table and not the 'regions' table. 
'UK' is a char and 'region_id' is an integer, producing the error. 

[...]

Issue #3523, opened 2021-11-24
Adding relationships breaks postgres / oracle sorting


Rory-Powell:
**Describe the bug**
When a relationship is added to a postgres table the sorting no longer works

**To Reproduce**
see:

https://user-images.githubusercontent.com/8755148/143261817-d59428c4-b724-4718-ab88-cb42436854d9.mov

vs

https://user-images.githubusercontent.com/8755148/143261832-816ac283-e0fd-4e82-b4fd-1160e398ef92.mov

Rory-Powell:
This one also applies to oracle. 
The root cause is the returned order in the database determined by the join queries used. 

Here's a video showing the different behaviour across the sql database: oracle -> pg -> mssql -> mysql


https://user-images.githubusercontent.com/8755148/143271377-52116b21-e010-426c-b982-35599d9f3764.mov

Issue #3327, opened 2021-11-11
Getting error on connecting Mysql data source.


aamirbinabd:
**Error statement:**
_Error saving datasource: Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'Team Member' at line 1_

I am adding a DB through localhost where Mysql is running on default port 3306. Budibase is running on docker. 

This DB is working as expected on multiple applications, no such error has ever been received. 

**Screenshot**
![IMG-20211111-WA0003](https://user-images.githubusercontent.com/81005721/141261286-ac206357-91fa-4bf7-a2ad-14e9406eb19a.jpg)

**Machine info:**
Ubuntu 20.04.3 LTS
Codename: focal
60GB Disk
4GB Memory

onofalt:
Dame issue here with Mysql and mariadb

marco-santulli:
same issue with mysql, in fact it only works if you chose sys as the databse

aamirbinabd:
I should add that my setup is based on MySQL and MariaDB (not just MySQL), like @onofalt has mentioned.

shogunpurple:
What version of MySQL/MariaDB are you all using?

marco-santulli:
I have tried various versions, I get this error on all databases excluding the sys database:
<img width="370" alt="image" src="https://user-images.githubusercontent.com/82715690/143236619-167775c1-f61f-4889-bcee-4d8fd8a77058.png">

marco-santulli:
I think I found the reason for my bug, if the database name has upper cases, it doesn't work... hopefully it's a simple fix adding double quotes to the budibase driver

PR #3541, opened 2021-11-25
Fixing various SQL issues

mike12345567:
## Description
1. #3327 - MySQL databases could not contain upper case letters due to the table lookup lowercasing the table name (MySQL db/tables are case sensitive depending on operating system/version).
2. #3523 - Sorting for Postgres/Oracle broke under certain conditions, added sorting before and after pre-query setup so that sorting plays nice in all environments with pagination.
3. #3531 - Fixed an issue with some relationship setups, also fixed some issues with many to many relationships and setting up a bunch of different relationships on the same table and then trying to update them.
4. Fixed a UI issue with deleting a datasource when a plus table was selected, this wouldn't navigate the user away - leaving them looking at a stale database which can no longer be accessed.


Rory-Powell (approved):
LGTM, thanks for going over these changes with me. And bigger thanks for getting them in so quickly 🚀

Issue #3337, opened 2021-11-11
Card block - horizontal mode too wide for mobile


joebudi:
When creating a block with the card block, I want to use the horizontal format. The horizontal block looks great but it's a little too wide for mobile.
![CleanShot 2021-11-11 at 13 23 58](https://user-images.githubusercontent.com/49767913/141305781-f77ccffd-095f-4589-a384-2be12c098ffe.png)

Issue #3262, opened 2021-11-06
Table component not showing rows for query based on a external Rest API


fseixas:
**Describe the bug**
When configuring a table component to show data from a query based on an external REST API, the table component always show "No rows found"

**To Reproduce**
1. Create a data source based on a REST API and create a new Query based on this API. I used http://content.sof.to:1337/Articles
2. Create a new page
3. Add a Data Provider component and set de "Data" property to the query you created on step 1
4. Add a Table component inside the Data Provider you created on step 3
5. The Table component will always show "No rows found"
6. If you add a Repeater component under the Data provider, and add a Card component to this repeater, configure the card title to any column from the data provider, it will show the rows from the provider correctly. 

**Expected behavior**
The Table component should show the data from the data provider

**Screenshots**
![2021-11-05 20_51_22-Window](https://user-images.githubusercontent.com/2001438/140590833-45516fa4-2115-4c48-825e-dd214c0712d0.png)

![2021-11-05 20_51_50-Window](https://user-images.githubusercontent.com/2001438/140590840-7653198a-3515-43ac-9752-ac14e31fc1c1.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome Version 95.0.4638.54 (Official Build) (64-bit)
 - Version 95.0.4638.54 (Official Build) (64-bit)

aptkingston:
Hey @fseixas. I've just replicated this exact setup (using the API your provided) and it works perfectly for me.
![image](https://user-images.githubusercontent.com/9075550/140958464-9c911dc1-44ca-4264-b37f-e859f38141f0.png)

I've even duplicated what settings I can see from your screenshots. Are there any errors if you open the console in developer tools?

The only thing I can think is that somehow the table has settings that is stopping it from displaying rows, but everything in your screenshot looks fine. Can you try deleting the table and adding a new one, and see if that makes your rows appear?

callsoft:
I'm having a similar issue but with an external SQL query.  Same result, table shows no data yet cards in repeater do.   

Deleting and re-adding the table didn't work.  

![image](https://user-images.githubusercontent.com/3580655/141651513-5c17b4fa-076d-49db-8b7c-6dda32607760.png)

aptkingston:
Would either one of you be able to provide me with an export of your app so I could debug this? Like I said I've tried to reproduce this using an identical setup but can't reproduce the issue. An export would really help if either of you could provide that.

fseixas:
[...]

PR #3542, opened 2021-11-25
Query schema fixes + card fixes

aptkingston:
## Description
Couple of fixes for some issues:
- Fix horizontal spectrum cards in the card block being too wide on mobile #3337
- Fix query definitions using the wrong logic to determine whether sensitive fields should be stripped (thanks @mike12345567)
- Fix query definitions removing the schema from the definition as this is required and not sensitive #3262


shogunpurple (approved):
LGTM! Nice one

PR #3554, opened 2021-11-26
API Docs, generation, tables and rows and Swagger definitions

mike12345567:
## Description
As per lab day demo, this adds the ability to generate some API docs (right now just tables and rows) along with the scripts needed to do this. I've added the following to make this a bit easier:
1. `yarn build:docs` a lerna command which will build and open the docs in the browser.
2. `server/scripts/docs/generate.js` which generates the docs, if `open` is specified at the end it will open in browser.
3. The beginnings of a swagger definition generator (2.0) based on https://github.com/fsbahman/apidoc-swagger - it can be found in `server/scripts/docs/toSwagger.js` which is generated when the API docs are.
4. For now all generated assets can be found in `server/builder/assets/docs` - the main files are `apidoc.json`, `project.json` and `swagger.json`.

There is still some work to be done before the swagger definitions are fully formed, but once they are it should be possible to import them into something like Postman (or Budibase!) to access the API.

In the future we could also import the docs into the builder as the JSON definitions are quite easy to use, e.g:
```
{
    "type": "post",
    "url": "/api/:tableId/rows",
    "title": "Creates a new row",
    "name": "Creates_a_new_row",
    "group": "rows",
    "permission": [
      {
        "name": "table write access"
      }
    ],
    "description": "<p>This API will create a new row based on the supplied body. If the body includes an &quot;_id&quot; field then it will update an existing row if the field links to one. Please note that &quot;_id&quot;, &quot;_rev&quot; and &quot;tableId&quot; are fields that are already used by Budibase tables and cannot be used for columns.</p>",
    "parameter": {
      "fields": {
        "Parameter": [
          {
            "field": "tableId",
            "description": "<p>The ID of the table to save a row to.</p>"
          }
        ],
        "Body": [
          {
            "field": "_id",
            "description": "<p>If the row exists already then an ID for the row must be provided.</p>"
          },
        ],
        "success": {
          "fields": {
            "Success 200": [
              {
                "field": "_id",
                "description": "<p>The ID of the row that was just saved, if it was just created this is the rows new ID.</p>"
              }
            ]
          }
        }
      }
    },

```

## Screenshots
The docs, shown on the basic templated page.
image


packages/server/scripts/docs/toSwagger.js
```diff
@@ -0,0 +1,320 @@
+let _ = require("lodash")
+let { pathToRegexp } = require("path-to-regexp")
+
+/********************************************************
+ * Based on: https://github.com/fsbahman/apidoc-swagger *
```
Rory-Powell, line 5: Probably a good idea just copying this, considering how out of date all the swagger generator libs for apidoc are! Good starting point for changes we need to deal with going forward

Rory-Powell (approved):
Nice one, brilliant start to this

packages/server/scripts/docs/toSwagger.js
mike12345567: Its MIT licensed and I thought this method would give us way more control when we come to getting the swagger definitions fully formed, means we can alter the generated API Doc JSON before converting it which should be a fair bit easier to do.

PR #3555, opened 2021-11-26
Improve returning logic for oracle

Rory-Powell:
## Description
Adding an improvement for returning logic to oracle. There were some scenarios where the primary key wasn't available in queries being made when a row being updated had relationships attached. Use the built in `lastRowId` identifier and a corresponding lookup that will always give us a row that was created or updated when there are no other rows returned in the result set. 

The `ROWID` is an identifier for a row that exists in every table in oracle. The value is usually something like `*BAEAqeMCUlD+` which is an encoded mechanism of telling oracle exactly where to go to get the row it needs.

### DIFF
--- a/packages/server/scripts/integrations/oracle/docker-compose.yml
+++ b/packages/server/scripts/integrations/oracle/docker-compose.yml
@@ -4,7 +4,7 @@
 version: "3.8"
 services:
   db:
-    container_name: oracle-xe
+    restart: always
     platform: linux/x86_64
     image: container-registry.oracle.com/database/express:18.4.0-xe
     environment:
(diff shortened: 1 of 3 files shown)


mike12345567 (commented):
Nice, looks good, makes sense to not use the common returning functionality because Oracle is so custom/specific - just one question, how is the delete handled? We still need the deleted row to be returned from oracle for row deleted automations, usually that needs to be retrieved before the delete operation.

Rory-Powell:
Very valid thank you. Added here https://github.com/Budibase/budibase/pull/3555/commits/5aa9d95c460c7b1026569781a479f5f1b7822070

mike12345567 (approved):
LGTM! Fantastic keeping all the changes for Oracle nicely encapsulated in its own layer - its all nicely hidden away at the deep layers; where it should be!

PR #3558, opened 2021-11-26
Fixing SQL enrichment issue

mike12345567:
## Description
Fixing issue found with enriching rows in SQL - the system was assuming the array of entries contained a row ID that needed to be broken down.


Rory-Powell (approved):

shogunpurple (approved):
LGTM! Good job getting on that so quickly 👌

Issue #3505, opened 2021-11-23
Navigation does not scroll


Rory-Powell:
**Describe the bug**
When using a vertical nav that has items that extend the window, the nav does not scroll

**To Reproduce**
See: 

https://user-images.githubusercontent.com/8755148/143056908-d9b537dc-785c-47e6-ba22-f4821e6f5c13.mov

aptkingston:
You monster with that nav! Valid bug though, will get it sorted.

Issue #3361, opened 2021-11-13
Duplicated element has all bindings replaced with `Invalid binding`


gvlx:
**Describe the bug**
The duplicate of an element which uses some bindings (in its value, or conditions or actions), has all its bindings replaced with `Invalid binding`.

This is odd as all references would certainly still work (the duplicated element is _always placed as the next sibling of the original_).

**To Reproduce**
1. Create an element (for instance, a paragraph) and set its value using some bindings (for a data provider or form)
2. Click on 'Duplicate'
3. The bound values in the duplicate are replaced with the text `Invalid binding`.

**Expected behavior**
No changes in the bound values.

**Screenshots**
![imagem](https://user-images.githubusercontent.com/67311/141651970-827d7789-7687-4484-86b9-7d7673c71ae1.png) 
![imagem](https://user-images.githubusercontent.com/67311/141652008-c4d96a44-3d06-4dba-90cc-64b5fd51b7bf.png)

**Desktop (please complete the following information):**
 - OS: W10
 - Browser Firefox 94.0.1
 - Version 0.9.185

**Additional context**
See feature request #

gvlx:
Hi @mjashanks,

Would you mind checking this?

This bug makes the `copy` and `duplicate` feature almost useless.
I'm forced to correct all validations, behaviors, code and references each and every time I make a copy of _any_ element.

Mitch-Budibase:
The duplicate behaviour surrounding bindings has been fixed and tested.

However, this is not the case for copy behaviour.
Copying and pasting a component with a binding will result in the copied component containing 'Invalid binding'

@aptkingston could you take a look please?

aptkingston:
@Mitch-Budibase yep that's correct. I was specifically addressing the "Duplicate" action because when duplicating we can guarantee that all bindings are valid. This isn't the case for copying and pasting, so we still remove the bindings for those.

PR #3576, opened 2021-11-29
Misc bug fixes

aptkingston:
## Description
A few bug fixes from the board:
- Preserve bindings when duplicating components #3361
- Allow layout navigation links to scroll #3505
- Don't mutate the real component structure when editing navigation links. This was causing navigation link settings to not save when editing.


Rory-Powell (approved):
LGTM!

PR #3614, opened 2021-12-01
Add fix for user metadata sync when deleting

Rory-Powell:
## Description
Fixing issue with user deletion. 
The sync endpoint in app service is throwing an error here 
https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/controllers/user.js#L100
due to empty roles object for deleted user


mike12345567 (approved):
Nice catch LGTM!

PR #3618, opened 2021-12-01
Fixes for button actions, block contexts and nester data provider settings

aptkingston:
## Description
Few fixes for some issues I discovered:

- Data providers were able to reference any other component that provided context as an available source. This doesn't make sense, and it should only show other data providers. This is fixed and now only shows data providers in this list.
- Fix button actions not saving as they were directly mutating the real component structure, preventing them from saving due to the original and new component definitions being the same
- Fix "save row" action not using the correct runtime context path when combined with a block. The "save row" action will now also only use either "schema" or "form" context types as available sources which filters out invalid options that we used to show.

We should hotfix this to master as well as I'm pretty sure button actions cannot be saved at the moment. It will look like everything works, but it won't actually save the definition to the server until another change is made, which will then effectively send up both changes.

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -61,7 +61,7 @@ export const getComponentBindableProperties = (asset, componentId) => {
 /**
  * Gets all data provider components above a component.
  */
-export const getContextProviderComponents = (asset, componentId) => {
+export const getContextProviderComponents = (asset, componentId, type) => {
   if (!asset || !componentId) {
     return []
   }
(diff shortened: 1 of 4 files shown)


Rory-Powell:
@aptkingston I'm pushing out a hotfix today, do you want to get a PR for these changes at master or happy to wait?

Rory-Powell:
Normally for hotfix I target the pr at master, merge, then merge the latest master back into develop so I can test in staging. 
For develop first targeted prs what's the approach, cherrypick the changes into a follow up pr for master?

aptkingston:
> Normally for hotfix I target the pr at master, merge, then merge the latest master back into develop so I can test in staging. For develop first targeted prs what's the approach, cherrypick the changes into a follow up pr for master?

Yep my bad, should have targeted master. I'll update.

aptkingston:
> @aptkingston I'm pushing out a hotfix today, do you want to get a PR for these changes at master or happy to wait?

Would be keen to include this yea.

aptkingston:
@Rory-Powell feel free to merge this whenever you get your hotfix up and merged, then we can release both.

Rory-Powell (approved):

PR #3628, opened 2021-12-01
Completely automate digitalocean release with hashicorp packer

shogunpurple:
## Description
Providing a totally reproducible version of the digitalocean one click offering built with hashicorps `packer` - which means that the digitalocean image can be built and snapshotted with one command, and should just instantly spin up the latest version of budibase whenever someone deploys from DO. With this, we should hopefully rarely need to upgrade the DO image (if ever).

Some other things:
- The current DO image was using one set of credentials, this will now generate random, new credentials for every instance spun up on DO, and write them to an `.env` file.
- Unrelated - Removed recording key from cypress ahead of upgrading to paid plan.

## Screenshots

### DIFF
--- a/hosting/digitalocean/README.md
+++ b/hosting/digitalocean/README.md
@@ -9,7 +9,7 @@ You must have the `DIGITALOCEAN_TOKEN` environment variable set, so that packer 
 ## Building
 Just run the following command:
 ```
-./build.sh
+yarn build:digitalocean
 ```
 
 ## Uploading to Marketplace
--- a/hosting/digitalocean/README.md
+++ b/hosting/digitalocean/README.md
@@ -1,0 +1,19 @@
+# Budibase DigitalOcean One Click
+You will find in this directory configuration for packaging and creating a snapshot for the Budibase 1 click Digitalocean build. We use this configuration to have an immutable and reproducible build package for Digitalocean, that rarely needs updated.
+
+## Prerequisites
+You must install Hashicorps `packer` to build the snapshot for digitalocean. Follow the instructions to install packer [here](https://learn.hashicorp.com/tutorials/packer/get-started-install-cli)
+
+You must have the `DIGITALOCEAN_TOKEN` environment variable set, so that packer can reach out to the digitalocean API for build information.
+
+## Building
+Just run the following command:
+```
+./build.sh
+```
+
+## Uploading to Marketplace
+You can upload the snapshot to the Digitalocean vendor portal at the following link (Requires vendor account):
+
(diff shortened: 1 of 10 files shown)


hosting/digitalocean/files/var/lib/cloud/scripts/per-instance/001_onboot
```diff
@@ -0,0 +1,22 @@
...
+
+# fetch envoy and docker-compose files
+wget https://raw.githubusercontent.com/Budibase/budibase/master/hosting/docker-compose.yaml
+wget https://raw.githubusercontent.com/Budibase/budibase/master/hosting/envoy.yaml
+wget https://raw.githubusercontent.com/Budibase/budibase/master/hosting/hosting.properties
+
+# Create .env file from hosting.properties using bash and then remove it
+while read line; do
+		uuid=$(uuidgen)
+		echo $line | sed "s/budibase/$uuid/g" | sed "s/testsecret/$uuid/g" >> .env
```
Rory-Powell, line 14: What does the `testsecret` mean here?

hosting/digitalocean/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Prerequisites
+You must install Hashicorps `packer` to build the snapshot for digitalocean. Follow the instructions to install packer [here](https://learn.hashicorp.com/tutorials/packer/get-started-install-cli)
+
+You must have the `DIGITALOCEAN_TOKEN` environment variable set, so that packer can reach out to the digitalocean API for build information.
+
+## Building
+Just run the following command:
+```
+./build.sh
```
Rory-Powell, line 12: Should we add a `yarn` command for this?

Rory-Powell (approved):
This is fantastic, really good job

aptkingston:
This is class Marty. Really clever idea to just always pull and run the latest version dynamically, and pretty much totally removes any maintenance workload for keeping the DO image valid.

hosting/digitalocean/files/var/lib/cloud/scripts/per-instance/001_onboot
shogunpurple: it's the default value for the `JWT_SECRET` value, so it just replaces it with a UUID.

hosting/digitalocean/README.md
shogunpurple: I'll add it for completeness - I considered making this whole thing a CI job, but there's not really a lot of point since it happens so infrequently.

PR #3644, opened 2021-12-02
Rest import

Rory-Powell:
## Description
REST Query import

Currently supported:
- OpenAPI 2.0 (Swagger)
   - Generate multiple queries
   - Populate params where possible 
- Curl
   - Generate single query 
   - Populates raw request with no params

OpenAPI 3.0 should be easy enough to add now that the foundation is in for 2.0

Libs added
- https://github.com/APIDevTools/swagger-parser
   - NPM: https://www.npmjs.com/package/@apidevtools/swagger-parser 
   - Parses json object and resolves all `$ref` fields for an easier to work with object
   - Manually convert yaml to json first to support yaml files
   - Supports 2.0 and 3.0
   - Returns typescript objects from `openapi-types`
- https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-types
   - NPM: https://www.npmjs.com/package/openapi-types 
   - TS definitions, makes working with openapi documents easier / safe
- https://github.com/nodeca/js-yaml
   - NPM: https://www.npmjs.com/package/js-yaml
   - Convert yaml file -> json
- https://github.com/curlconverter/curlconverter  
   - NPM: https://www.npmjs.com/package/curlconverter
   - Parse curl string into json 

## Screenshots
Screenshot 2021-12-02 at 13 27 56
Screenshot 2021-12-02 at 13 28 00
Screenshot 2021-12-02 at 13 28 06
Screenshot 2021-12-02 at 13 28 15
Screenshot 2021-12-02 at 13 29 32
Screenshot 2021-12-02 at 13 29 39


packages/server/src/api/controllers/query/import/sources/base/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  url: string
+  name: string
+}
+
+export interface QueryParameter {
+  name: string
+  default: string
+}
+
+export interface Query {
```
mike12345567, line 11: This is great, only thing I'd say is could we move the interfaces to the common `src/definitions` directory? Theres a place for datasource definitions in there already. keeping them all centralised like that means that if anyone goes looking for a definition they'll find it there and not think that it needs written etc.

mike12345567 (approved):
LGTM! Looks great, lovely feature, can't wait to use the cURL importer! Also, great test coverage, great to see this coming in with such a nice suite of tests!

packages/server/package.json
```diff
@@ -108,6 +111,7 @@
     "mysql2": "^2.3.1",
     "node-fetch": "2.6.0",
     "open": "^8.4.0",
+    "openapi-types": "^9.3.1",
```
mike12345567, line 114: This should be moved to devDependencies so that production builds don't include it, typing is only needed for development (`yarn add --dev openapi-types` and remove from `dependencies`).

packages/server/src/api/controllers/query/import/sources/base/index.ts
Rory-Powell: Nice suggestion, I wasn't sure on the best place for this. This is ideal - added in https://github.com/Budibase/budibase/pull/3644/commits/9d402d03560f9bbbc4d3106d82aa14904bfba79d

packages/server/package.json
Rory-Powell: Great shout! added here https://github.com/Budibase/budibase/pull/3644/commits/9d402d03560f9bbbc4d3106d82aa14904bfba79d

PR #3702, opened 2021-12-07
Duplicate queries

Rory-Powell:
## Description
Duplicate a datasource query. Use the same logic as postman:
```
     name           all names          result
    ------        -----------         --------
  ("foo")       ["foo"]              "foo (1)"
  ("foo")       ["foo", "foo (1)"]   "foo (2)"
  ("foo (1)")   ["foo", "foo (1)"]   "foo (2)"
  ("foo")       ["foo", "foo (2)"]   "foo (1)"
```

## Screenshots

https://user-images.githubusercontent.com/8755148/145028921-d917a512-dcb3-438d-a731-578062849e8f.mov

### DIFF
--- a/packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
+++ b/packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
@@ -14,7 +14,7 @@
 
   async function duplicateQuery() {
     try {
-      await queries.duplicate(query, queries.save)
+      await queries.duplicate(query)
     } catch (e) {
       notifications.error(e.message)
     }
--- a/packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
+++ b/packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
@@ -11,6 +11,14 @@
     await queries.delete(query)
     notifications.success("Query deleted")
   }
+
+  async function duplicateQuery() {
+    try {
+      await queries.duplicate(query, queries.save)
+    } catch (e) {
+      notifications.error(e.message)
+    }
+  }
 </script>
 
 <ActionMenu>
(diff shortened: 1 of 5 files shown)


packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
```diff
@@ -11,13 +11,22 @@
     await queries.delete(query)
     notifications.success("Query deleted")
   }
+
+  async function duplicateQuery() {
+    try {
+      await queries.duplicate(query, queries.save)
```
aptkingston, line 17: We could refactor how the actions in the `queries` store are defined, so that they can reference each other. This would remove the need to pass in the save function as a param here.

Basically defining the queries store something like this:
```js
...
const { subscribe, set, update } = store
const actions = {
  ...
  save: () => { ... },
  duplicate: () => {
    ...
    actions.save()
    ...
  }
}
return {
  subscribe,
  set,
  update,
  ...actions
}
```

It would just feel more natural to me to be able to call `queries.duplicate(query)` without having to pass in that second param, and I assume the current inability to reference `save` from `duplicate` was the main motivator for that. If there's another reason that I'm not thinking of at the moment though then let me know.

aptkingston (approved):
Awesome Rory. Super slick and very useful. I like how it selects the new duplicated query too.

packages/builder/src/components/backend/DatasourceNavigator/popovers/EditQueryPopover.svelte
Rory-Powell: Yep spot on, I wasn't sure how to reference another store action short of pulling it out to an outer function which didn't feel right to do for one and not the others. Your suggestion is great and works like a charm, I didn't think an inner object function could reference the variable it was being defined in 🤯 

Added here https://github.com/Budibase/budibase/pull/3702/commits/3bd93f275585bc0560198b347c499c5f6b1ff536

PR #3737, opened 2021-12-09
Home Screen Redesign

PClmnt:
## Description
This PR updates the home screen to the new design to include templates on the main screen. it also adds a new modal / popover selection that allows the user to change their icon / colour for a chosen app. It defaults to the 'Apps' icon when an app is first created. 

Also:

- simplified the app creation modal so you just need to add the name
- added  a quiet option to the search component
- removed the old grid view

## Screenshots
image

image

image


packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -21,7 +15,6 @@
 
   export let template
   export let inline
-
   const values = writable({ name: null })
```
Rory-Powell, line 18: NAB: normally I'm a fan of the gap between the exported and internal variables

packages/builder/src/stores/portal/apps.js
```diff
@@ -65,6 +65,27 @@ export function createAppStore() {
     }
   }
 
+  async function updateIcon(appId, icon) {
```
Rory-Powell, line 68: We could probably combine this with the `update` below so that we have a single function for logic around app updates, the implementations look pretty similar. 

Just a question, does the `put` endpoint act as a `patch`? Seems like it's fine if we send partial data by looking at usages of the two functions

Rory-Powell (approved):
Looks great! Just the one comment about possible reusing some logic around updating an app

packages/builder/src/components/start/ChooseIconModal.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+<Modal bind:this={modal} on:hide={onCancel}>
+  <ModalContent
+    title={"Edit Icon"}
+    confirmText={"Save"}
+    onConfirm={() => save()}
+  >
+    <div class="scrollable-icons">
+      <div class="title-spacing">
+        <Label>Select an Icon:</Label>
```
aptkingston, line 69: NAB: Would probably case this as just "Select an icon" (without the colon as well).

packages/builder/src/components/start/ChooseIconModal.svelte
```diff
@@ -0,0 +1,125 @@
...
+            on:click={() => (selectedIcon = item)}
+          >
+            <Icon name={item} />
+          </div>
+        {/each}
+      </div>
+    </div>
+    <div class="color-selection">
+      <div>
+        <Label>Select a Color:</Label>
```
aptkingston, line 85: NAB: same as above, would just case this as "Select a color" (without the colon as well)

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -299,35 +357,19 @@
...
       </div>
-      <div
-        class:appGrid={layout === "grid"}
-        class:appTable={layout === "table"}
-      >
+      <div class="appTable">
         {#each filteredApps as app (app.appId)}
           <svelte:component
-            this={layout === "grid" ? AppCard : AppRow}
+            this={AppRow}
```
aptkingston, line 370: Since this is hardcoded now to just be one component, we can remove the `<svelte:compoennt>` tag and just render a `<AppRow ... />` directly.

packages/builder/src/components/start/AppRow.svelte
```diff
@@ -18,14 +18,26 @@
   export let deleteApp
   export let unpublishApp
   export let releaseLock
+  export let editIcon
+
+  $: color =
+    $apps.find(filtered_app => app?.appId === filtered_app.appId)?.icon
```
aptkingston, line 24: Does the passed in `app` prop not contain this info already in `app.icon`? Or is there a reason that we need to find the app definition from the list again?

packages/builder/src/components/start/AppRow.svelte
```diff
@@ -18,14 +18,26 @@
...
+      ?.color || ""
+  $: name =
+    $apps.find(filtered_app => app?.appId === filtered_app.appId)?.icon?.name ||
```
aptkingston, line 27: Same as above about just accessing the value directly - I think we can just use `app.name` directly.

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -260,8 +266,14 @@
     }
   }
 
+  async function fetchTemplates() {
+    const response = await api.get("/api/templates?type=app")
```
aptkingston, line 270: NAB: It might be a nice idea to save templates in a store. On initial load we need to load them so we can't avoid a loading spinner there, but if we cache them in a store (and still refresh them on mount here) it means that at least after the initial load there won't be a flashing loading spinner! I haven't checked out the code at ran it though so maybe that's a non-issue - it might look totally fine already.

packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -207,16 +190,12 @@
...
-    <Body size="S">
-      Give your new app a name, and choose which groups have access (paid plans
-      only).
-    </Body>
     <Input
       bind:value={$values.name}
       error={$touched.name && $errors.name}
       on:blur={() => ($touched.name = true)}
       label="Name"
+      placeholder={`${$auth.user.firstName}'s first app`}
```
aptkingston, line 198: `firstName` might not be defined - so we might need to account for that and render some fallback like `Your first app` if that's the case.

packages/builder/src/components/start/AppRow.svelte
```diff
@@ -18,14 +18,26 @@
   export let deleteApp
   export let unpublishApp
   export let releaseLock
+  export let editIcon
+
+  $: color =
+    $apps.find(filtered_app => app?.appId === filtered_app.appId)?.icon
+      ?.color || ""
```
aptkingston, line 25: I'm guessing if `color` is an empty string like this then the inline style later down is ignored and it renders the icon with the defaullt spectrum color?

aptkingston (commented):
Looks great Peter! Just a few questions noted.

packages/builder/src/components/start/CreateAppModal.svelte
Rory-Powell: good call

packages/builder/src/components/start/AppRow.svelte
PClmnt: yep exactly

packages/builder/src/pages/builder/portal/apps/index.svelte
PClmnt: Seemed fine, but added a templates store anyways, slow internet you'd maybe run into this.

packages/builder/src/stores/portal/apps.js
PClmnt: Yea it does, updated the function to handle both.

aptkingston (approved):

Issue #3509, opened 2021-11-23
Request Body not being processed when using REST API (Put/Post)


developer-uhwi:
**Describe the bug**
I have configured a data source to post to a REST api endpoint, although I have added the params correctly it sometimes fail with missing params. I have now remove all the params except one and have manually entered the test data in the field, this action works and a new item is created but the data in the field is not passed to the query.

I have check my endpoint and it says to values were received.

**EDITED**
**I have done further checks and realize that the Request Body is not being processed. The values have to be passed instead using the query string.**

**To Reproduce**
1. Go to Query Rest API
2. Click and create a query name and use POST as the function.

**Expected behavior**
A new entry should be created, but this does not happen.

**Screenshots**
<img width="698" alt="Screenshot 2021-11-23 at 2 06 55 PM" src="https://user-images.githubusercontent.com/10439242/143088581-f621a38c-2c99-4e54-86f8-90eb83bb23e8.png">
<img width="754" alt="Screenshot 2021-11-23 at 2 07 03 PM" src="https://user-images.githubusercontent.com/10439242/143088602-7933c9cd-32fc-421a-92ca-2c7fcc2dc659.png">

**Return from Browser Console**
{
  "fields": {
    "headers": {},
    "queryString": null,
    "path": "contacts",
    "requestBody": "{\n  \"full_name\": \"{{full_name}}\",\n  \"first_name\": \"{{first_name}}\",\n  \"last_name\": \"{{last_name}}\",\n  \"external_identifier\": \"{{external_identifier}}\",\n  \"record_type_id\": {{record_type_id}}\n}"
  },
  "queryVerb": "create",
  "transformer": "return data",
  "parameters": {
    "full_name": "Jon Doe",
    "first_name": "Jon",
    "last_name": "Doe",
    "external_identifier": "TL111",
    "record_type_id": "1"
  },
  "datasourceId": "datasource_1490adf186e6417ca57a8c82437d6617"
}
**Desktop (please complete the following information):**
 - OS:  macOS Monterey version 12.0.1
 - Browser  : chrome
 - Version Version: 96.0.4664.55 (Official Build) (x86_64)

**Additional context**
It seems the request body is not being sent to the API endpoint, I have checked and the endpoint indicates no data being received.

shogunpurple:
Hi @developer-uhwi - this is indeed a bug - I just tried it.

[...]

Issue #3504, opened 2021-11-23
REST Connector V2: Enable Handlebars bindings in URL/Query Params


shogunpurple:
When I'm configuring URL Params and the URL itself for a REST query, I want to be able to use my handlebars arguments so that I can perform more dynamic queries using values from the frontend.

**COS**
- Run URL and URL Params fields through handlebars interpolation with the arguments in the query

Issue #3496, opened 2021-11-23
REST Connector V2: Headers Updates


shogunpurple:
When I'm using the new REST connector, I want to have more control over my header keys, so that I can create headers faster.

**COS**
- Add relevant HTTP headers from https://www.iana.org/assignments/message-headers/message-headers.xml#perm-headers
- Add auto complete to values for common headers such as Content-Type
- Add option to enable/disable headers for testing, rather than delete them completely

Issue #3494, opened 2021-11-23
REST Connector V2: UI Redesign


shogunpurple:
**Spec**
https://docs.google.com/document/d/1IC9fs9SapinXtNGMAxwXVtn1V51-PFFwif7hzeuHTzc/edit

**COS**
- [x] Redesign the current REST connector Query UI (not the datasource config UI) by putting the form sections into tabs. Parameters, Fields, Headers, Transformer
- [x] Add response status, response time and response size as stats above the result/schema UI
- [x] Add HTTP Verb Identifiers beside the queries on the left hand side navbar
- [x] "Raw" view in schema/results UI, to show the full, pretty printed and scrollable output of the API call
- [x] "Headers" view in schema/results UI, to show the response headers from the request
- [x] Key/Value configuration for JSON request body
- [x] Fix broken link to https://github.com/Budibase/budibase/discussions/ from REST connector and point to docs on REST
- [x] Create more documentation and examples on the REST connector
- [x] Add placeholder fields in all REST URL fields just to show some examples of the URL format
- [x] Allow the use of a full URL with protocol in queries, which will override the baseURL

PR #3741, opened 2021-12-09
Rest UI Redesign

mike12345567:
## Description
This redesign falls inline with Joe's design work on the matter, around building a custom experience for REST that fits with other industry standard tools.

The following issues are covered by this work:
1. #3504 - all elements of the REST request now support handlebars expressions.
2. #3509 - moved where the request body was being parsed, so that it is always parsed after the handlebars expression evaluation, solving this issue (also helping to lead the way for different body types).
3. Partial #3496 - completed the active/inactive toggles for headers, still needs work for the autocomplete/common values.
4. Completes #3494 - bar two issues, the ability to create JSON fields with a key value setup as this isn't part of the designs and could be quite limiting if it doesn't support nesting and the improvement of docs - docs can now be worked on with the new UI.

Its hard to cover everything this introduces in one PR, but I'll give it a go!
* Everything mentioned in #3494
* A banner component with learn more button, with a flags endpoint for a user that allows them to hide something permanently.
* Body options, these are not implemented yet but I've put a bit of the ground work in behind the scenes to make this a bit easier.
* Params which relate to query string params, as you add to the key value store it will automatically update the URL to make it obvious how it works.
* Renaming "Parameters" throughout datasources to bindings as well as cleaning up the layout of some drawers. After discussion with @Rory-Powell we decided renaming this made sense because the word "parameters" has different meanings in different datasources, its better to be explicit and just call bindings what they are.
* A tooltip on the bindings key value component which describes bindings.
* The question mark button in bottom right of screen now shows a message "Need help? Go to our forums" on hover to explain its not a docs link.
* The name of the rest request is automatically generated from the URL, until it is edited. On hover an edit icon will appear and allow the user to overwrite the default name to whatever they desire.
* Extra info and data returned by the REST query, such as the HTTP stats and the response headers.
* Access control is now in the top right of screen.

## Unfinished
There's only a few things that still need to be done:
1. The key/value option for filling in JSON, not sure how this would work with nesting.
2. Improvement of REST docs.
3. Different request body types - right now none and JSON are implemented, but need to implement form, files etc - wanted to get the REST UI merged so that we can begin using it and testing it as its quite a large change.

## Screenshots
New HTTP verbs instead of icons for REST queries
image

New datasource UI, buttons have been changed (if a button is in the top right of section it is a standard button, action buttons are only used for adding to sections) and headers have been broken out into their own section with an optional tag.
image
This also bleed over into other datasources, tables are now used instead of custom components in SQL plus datasource UI.
image

A general look at the UI, showing generally all of the supported features
image

Screen with learn more banner
image


packages/server/src/integrations/rest.ts
```diff
@@ -5,8 +5,52 @@ import {
...
+    enum: Object.values(BodyTypes),
+  },
+}
+
 module RestModule {
   const fetch = require("node-fetch")
+  const { formatBytes } = require("../utilities")
+  const { performance } = require("perf_hooks")
+
+  interface RestQuery {
```
Rory-Powell, line 44: Just realising now that this base query I have added isn't very generic
https://github.com/Budibase/budibase/blob/develop/packages/server/src/definitions/common.ts#L121-L125

We could link these two by renaming this one you have added as `RestQueryFields` and replacing fields block in the common definition with
```js
fields: RestQueryFields
```

I'm not sure if the fields key is used by other types of query, if so we can type them and expand the definition e.g.
```js
fields: RestQueryFields |  PostgresFields | S3Fields
```

Should we move this one out into the definitions folder as well?

Rory-Powell:
Looks and works great! I had a play about with this based on the auth stuff and found some things I think we should look at before merging
- bug in the request / response tabs - always goes to first one on double click 
- new headers should be enabled by default
- show full url in query when there is just a path e.g. sub in the existing config url
- we should use `disabledHeaders` instead of `enabledHeaders` so that existing query headers still work
- can't add body to get requests - I know this isn't officially recommended in the rest spec, but a lot of api servers support it and there will be requests people want to add with this

probably unrelated to this pr
- deleting last datasource breaks page

packages/server/src/integrations/rest.ts
mike12345567: yeah we probably should, it is used by basically all query types, so it might just need an `any` on the end for now as there isn't interfaces for everything, it really can contain anything.

joebudi:
Wonderful work @mike12345567 . Looking forward to playing around with this.

packages/server/src/integrations/rest.ts
Rory-Powell: aye right enough, for now we could do
```js
fields: RestQueryFields | any
```

during the import stuff I came across `is` functions for typescript for safe casting where we can have a function like:
```
const isRestQuery(query: Query, fields: any) fields is RestQueryFields => {
   query.type === "REST"
} 
```
and then get typesafe usage on a multi type field by 
```
const fields = query.fields
if (isRestQuery(query, fields)) {
 // fields is now type checked to 'RestQueryFields
}
```

aptkingston:
(realise I'm too late and this has been merged)

Firstly this is awesome! It looks so good, seriously.
The only point in terms of design I was going to say, was should we make the query verb a fixed length so we could align the queries?

i.e. changing this:
image

to this:
Capture

Might just make it easier to read the queries.

PR #3785, opened 2021-12-14
REST body types

mike12345567:
## Description
Small piece of functionality to provide the following:
1. Handle REST body form-data and URL encoded form data
2. allow text and JSON based responses
3. Allow sending XML (editor switched into XML mode)
4. Receive XML and parse it to JSON
5. Allow switching the editor mode reactively without needing to remount 
6. Disables the outgoing webhook action (in the frontend) so that it cannot be added to new automation definitions, instead the external data connector action can be used to fire a REST request instead
7. Adding back check to not allow bodies on GET requests, as node fetch doesn't support this


packages/server/src/integrations/rest.ts
```diff
@@ -150,7 +177,51 @@ module RestModule {
       return complete
     }
 
-    getAuthHeaders(authConfigId: string): { [key: string]: any }{
+    addBody(bodyType: string, body: string | any, input: any) {
```
Rory-Powell, line 180: NAB: good candidate for unit tests

packages/server/src/integrations/rest.ts
```diff
@@ -110,15 +114,38 @@ module RestModule {
...
     async parseResponse(response: any) {
       let data, raw, headers
-      const contentType = response.headers.get("content-type")
-      if (contentType && contentType.indexOf("application/json") !== -1) {
-        data = await response.json()
-        raw = JSON.stringify(data)
-      } else {
-        data = await response.text()
-        raw = data
+      const contentType = response.headers.get("content-type") || ""
```
Rory-Powell, line 117: NAB: good candidate for unit tests

Rory-Powell (approved):
Very nice!

shogunpurple (approved):
Some really nice updates here. Agree that a little unit test coverage might be good just to cover the conditionals. Some snapshot tests even

Issue #3781, opened 2021-12-14
App state set value not working while open as Modal


ggs331:
**Describe the bug**
App state set value not working while open as Modal, but works on Full Screen mode.

Please find the screen video. If the page open as modal, the app state set value on the modal page didn't work.
https://user-images.githubusercontent.com/29649651/145947452-519625f3-02cb-4b23-9344-8168ddfadd48.mov

ggs331:
BTW, everything work fine on v0.9.173, this issue just happened  after upgraded to 1.0.x.

ggs331:
When I trying to reproduce this issue, I found something strange. Then `Update State` will not work in modal if it's ran after `Execute Query`, but works in modal if I put the `Update State` before `Execute Query`

If `Update State` as the below screenshot, the `Update State` not work while open as modal.
![image](https://user-images.githubusercontent.com/29649651/145950204-a34181e0-88d4-4cb3-a4fe-cf69dc921e0d.png)

If I put the `Update State` on top of the actions, then it works on both modal and full screen mode.
![image](https://user-images.githubusercontent.com/29649651/145950627-8e4e3778-a64b-4bab-a6f2-5067d87d3f29.png)

aptkingston:
Thank you for the detailed description and the video @ggs331! It makes things much easier to reproduce and fix. I know the reason for this already and think it should be an easy fix.

ggs331:
Thank you @aptkingston . Appreciate if you can let me know when this fix released as I was stuck on 0.9.173 due to this issue.

ggs331:
@aptkingston Thanks for the update, I have upgraded to 1.0.24 and seems the issue mentions before has been fixed. 

But after upgraded just for few hours, I found another issue is: when button click and updated app state, it may caused other button `Navigate To > Open screen in modal` just didn't work(no any response at all), and this open modal not working issue not just happened on a single page, but all of other pages. I still not find how to reproduce this issue. But it happens on my apps. Now I have to disable `Open screen in modal` option temporary. I am wondering this issue maybe something related to this bug fix?

I will update you if I find more details.

aptkingston:
Hi @ggs331. Could you take a screenshot of your button actions? I have tried to reproduce this but everything works.

I have tried:
- Update state only
- Update state, then navigate to a screen in a modal
- Navigate to a screen in a modal, then update state
- Navigate to a screen in a modal only

[...]

PR #3791, opened 2021-12-15
Screen modal fixes for state and datasource invalidation

aptkingston:
## Description
Fixes app state not being shared with screen modals. #3781
- When a screen modal is opened, the current state is passed into the new iframe modal by serialising it and passing it as a query param
- When a state update action is executed inside a screen modal, this event is proxied back to the parent window so that it also updates

Also some other bug fixes:
- Fix datasource invalidation (for hot reloading) inside screen modals not being proxied back to the parent window
- Fix persisted app state using an incorrect local storage key

Merging to master as this addressesa couple of bugs which are currently live.


shogunpurple (approved):
LGTM! Nice one

PR #3795, opened 2021-12-15
Home screen UI tweaks

aptkingston:
## Description
This is just a couple of small design tweaks to the new home screen UI after checking it out and using it.
Super job on the rework by @PClmnt and this is only building on top of that! The diff looks bigger than it actually is because indentation changed.

Tweaks:
- Improve mobile compatibility and wrapping of buttons and elements
- Increase gap between sections but reduce gap between section title and content, to give better a indication of separation between the header, templates and app list sections
- Remove the quiet prop from the search input and "open" button in the apps list
- Move the search/sort UI over inline with the title rather than below
- Slightly increase size and reduce thickness of section titles
- Adjust grid settings on templates so that they stretch to fill empty space
- Slightly increased font size of main title
- Swap order of search and sort picker
- Change selected icon color (in icon editor) to blue, rather than the users chosen color. This is because initially, when no color is selected, (or when choosing a grey color) it was impossible to tell what icon was selected.
- Mark the default app icon of "Apps" as visually selected in the icon editor

## Screenshots
Desktop:
image

Mobile:
image

Icon picker for a brand new app, showing the icon highlighted in blue, and the default "Apps" icon selected:
image

### DIFF
--- a/packages/bbui/src/Layout/Layout.svelte
+++ b/packages/bbui/src/Layout/Layout.svelte
@@ -63,6 +63,9 @@
   .gap-L {
     grid-gap: var(--spectrum-alias-grid-gutter-medium);
   }
+  .gap-XL {
+    grid-gap: var(--spectrum-alias-grid-gutter-large);
+  }
   .horizontal.gap-S :global(*) + :global(*) {
     margin-left: var(--spectrum-alias-grid-gutter-xsmall);
   }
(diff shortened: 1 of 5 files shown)


PClmnt (approved):

shogunpurple (approved):
LGTM!

PR #3796, opened 2021-12-15
Infra/ci pipelines

shogunpurple:
## Description
This PR contains code to improve the existing CI pipelines and to add new ones. Note that some are still in progress, but I need to merge this to master to test it properly before moving on.

#### Release Pipeline (WIP)
Code has been added to pull `values.preprod.yaml` from `budibase-infra` and deploy it to the preprod environment running in AWS as part of the release process. All pushes to master will be reflected in the pre-prod env, and when we are happy with them we can kick off the `cloud-deploy` and `release-selfhost` job to roll things out to our users. 

#### Release Selfhost Pipeline (WIP)
The release-selfhost pipeline previously built the latest docker images from master, but has been updated to simply pull the latest image from docker hub, tag it with `latest`, and push it back up. This should dramatically speed up this job, as the docker build is the slowest part.

The part that is in progress is the automated release of the budibase helm chart, and a subsequent github release to indicate to github users that a new version of budibase is available. In order to ensure that the self-host pipeline is still functional for now, I've just commented this out until the other pipelines are tested.

#### Cloud Deploy Pipeline
The cloud deploy pipeline is a manual `workflow_dispatch` pipeline that will allow anyone to deploy the code on the latest master to our kubernetes cloud platform, with one click. The action is also parameterised and takes a version, which can be used to perform rollbacks in production. If no version is specified, the latest `lerna.json` version is used by default.

It pulls the `values.yaml` from `budibase-infra` and injects it into the `helm upgrade` call, which will deploy directly to production. This means any configuration can be done over there, then you can kick off the cloud deploy job to deploy to the K8S environment. 

This means nobody should have to deploy directly from their machines anymore using `helm`, and everyone has the ability to deploy to our cloud platform without configuring their setup to interact with our EKS cluster.

## Screenshots


Rory-Powell (approved):
Looks good so far! For testing this stuff we might hit a wall unless we actually start releasing new versions and deploying stuff.  For testing the lerna prerelease stuff in the past I setup a dummy repo and changed the package.json / npm token so that I could test ci stuff in isolation. It might be nice to make a fork of budibase and do something similar so we could test our workflows end to end on a different npm / docker hub account. Because the k8s preprod env isn’t used yet that should be fine to remain as the target, we could probably configure the cloud deploy job to use that env too until everything is right and we could switch everything over for the final test. 

This is all just a suggestion and could take some time to setup which might not be worth the time sink, it would be nice though to have a way of testing this stuff without affecting real artifacts and environments which will become more important as we go along

shogunpurple:
@Rory-Powell that's a pretty good suggestion. The releasing versions thing is fine, since it doesn't really do anything other than increment the version number (another thing we should probably look to resolve for cleaner releases, as we discussed) but the deployment part makes me a bit nervous. 

Between the pre-prod and prod environments, the deployment steps aren't the same because one uses EKS and the other one doesn't. The only way I'll really be able to test against production is by deploying to either an exact replica EKS cluster, or by trying production and rolling back manually if it doesn't work.

In future I think we should look at platforms that allow you to run your CI totally locally, like act. There's got to be a way that we can do that and it would be able to mock the env and let us test it from our machines at least

PR #3901, opened 2021-12-31
add DingTalk & WeChat Bot webhook for automations

listeng:
## Description
Add [DingTalk](https://www.dingtalk.com/) and [WeChat(Weixin)](https://work.weixin.qq.com/) Group Message Bot for automation apps.


listeng:

shogunpurple (commented):
Hi @listeng - thank you for the contribution.

I'm not sure if these automations are necessary - you can do the same thing with the "Outgoing webhook" automation already included in budibase.

Screenshot 2021-12-31 at 14 29 03

Is the outgoing webhook not sufficient for your use case?

listeng:
Hi @shogunpurple ,Thanks for your review. But the "Outgoing webhook" had removed by #3785  - "Disables the outgoing webhook action (in the frontend) so that it cannot be added to new automation definitions, instead the external data connector action can be used to fire a REST request instead", and the REST datasource is very complicated……

shogunpurple:
Apologies @listeng - I was under the impression that you were self hosting, and you were just trying to get this to work with wechat/dingtalk.

You are right - the outgoing webhook automation is being replaced with the new REST connector. To clear up the confusion - here's a video showing how you can model the same thing as the outgoing webhook connector in an automation, with the new REST connector.

https://user-images.githubusercontent.com/11256663/147859581-55ccb042-6a4d-42a1-b846-c35a145673f2.mov

We will add docs for this around the time we release the new REST connector, which will be in a week or two. The reason we removed the outgoing webhook connector and opted for the REST connector, is because the REST connector is much more flexible. You can add auth, variables, different request types and much more to the REST connector, so it didn't make sense to have 2 "versions" of an outgoing webhook, when you can just use the REST connector instead.

I'm going to close this one for now as it won't be required when the new REST connector is available - but we will make sure to update the docs to guide users through how to set this up.

Thanks again!

Issue #3489, opened 2021-11-23
Unable to create new role without touching 'Base Permissions' field


Rory-Powell:
**Describe the bug**
The permissions dropdown must be used even when there is a value present already

**To Reproduce**
See: 

https://user-images.githubusercontent.com/8755148/143002163-11f4c5a5-761e-4983-a3f0-59fe54786973.mov

![Screenshot 2021-11-23 at 09 47 08](https://user-images.githubusercontent.com/8755148/143002337-a477fc92-c631-4eaa-9f8a-2c92f755a23e.png)

Caused by:

```json
{
  "message": "Invalid body - \"permissionId\" must be one of [public, read_only, write, admin, power]",
  "status": 400
}

```

**Expected behavior**
The role should be able to be saved without touching the dropdown

PR #3913, opened 2022-01-04
Bug fixes for #3399 and #3489 

### LINKED ISSUE #3399: Filter column type number with value 0 not working
**Describe the bug**
Filter column with type number and set value to 0 will not affect the result.

**To Reproduce**
1. In Data design, create table and set new column type to number
2. Create multiple new rows and key-in any number in the column previously created Eg: 1 , 9 , 10
3. Create filter and choose value of the column to 0

**Expected behavior**
Result of the table should return empty record but this bug return all records.

**Screenshots**
image

image

**Additional context**

PClmnt:
## Description
Just a very small PR that fixes #3399 and #3489


shogunpurple (approved):
LGTM!

Issue #3915, opened 2022-01-04
Short App Links for Regular Users


serkonyc:
**Describe the bug**
Short app links don't work for regular users until you set up development/administration access level on user level. Please find some screenshots below.

The BB system is deployed to Ubuntu instance via Docker container using BB CLI.

**To Reproduce**
1. Create an app, e.g., named 'TIMS'
2. Publish this and open using short link like http://domain.com:10000/app/tims
3. Check that home page is shown
4. Open incognito window and go to the same URL, e.g., http://domain.com:10000/app/tims
5. See the "too many redirects" error
6. Go to the full automatically generated URL, e.g., http://domain.com:10000/app_fda6032376cd4f7788bf1574acb93583
7. Login page is successfully shown

**Expected behavior**
Login screen must be shown.

**Screenshots**

Under admin/dev user (**works**, short link):

<img width="958" alt="TIMS 2022-01-04 14-54-29" src="https://user-images.githubusercontent.com/6172067/148055383-f1d86bf3-b1ea-4225-8e66-543be5c44833.png">

Under regular/guest user (**fails**, short link):

<img width="1009" alt="TIMS 2022-01-04 14-55-24" src="https://user-images.githubusercontent.com/6172067/148055509-9fba120c-4ffd-49f1-acbc-453440cc9165.png">

Under unauthenticated guest (**works**, full link):

<img width="628" alt="Budibase 2022-01-04 15-00-44" src="https://user-images.githubusercontent.com/6172067/148056142-437daf80-1b2c-48f6-a498-fbc15d1651a6.png">

**Desktop:**
 - OS: macOS
 - Browser: any
 - Version: 12.0.1

serkonyc:
UPD: this doesn't depend on level of user role. Tried basic/power/admin access level – works only when user is updated with admin/dev flag using main builder window.

PR #3931, opened 2022-01-05
Fixing some issues with short hand URL for prod apps - using it by default in self host

mike12345567:
## Description
Fixing issue #3915 so that when using the short hand URL for apps you will be redirected correctly if not logged in, allows all users to make use of it and updating the builder to send the user to the short form factor URL if not in cloud.

The feature has been supported for a while but isn't really known about, I've seen a lot of comments recently about being able to use a prettier URL to access the app, thought making the app name the default was better than the app ID as that is something which no one will ever be able to remember.

After recent changes we could actually support using the short hand URL in the cloud as well (within the scope of a tenant) but thought we should make it the default for self host first, make sure no issues there and then we could roll it out to cloud down the line too.

Also removed a component that was a repeat of another but unused.


packages/builder/src/pages/builder/apps/index.svelte
```diff
@@ -34,12 +34,16 @@
...
-
+  $: isCloud = $admin.cloud
   $: userApps = $auth.user?.builder?.global
     ? publishedApps
     : publishedApps.filter(app =>
         Object.keys($auth.user?.roles).includes(app.prodId)
       )
+
+  function getUrl(app) {
+    return !isCloud ? `/app/${encodeURIComponent(app.name)}` : `/${app.prodId}`
```
Rory-Powell, line 45: How come we need `/app` prefix for the short name but not the id?
Would you be able to paste an example link for the short and long form?

Rory-Powell:
Awesome that this exists, agree with your plan for rolling it out to cloud after some initial testing self hosted 👍

packages/builder/src/pages/builder/apps/index.svelte
mike12345567: the `/app` is to make it easier to direct to the server instance through envoy, the same way `/db` or `/global` go to CouchDB/worker respectively.

We had a special case for the app ID in Envoy, directing anything from `/app_` to the server instance, so that it could work out how to serve the client etc, but if we are using the app name we need some consistent way to direct to the server instance (the default is MinIO because it didn't play nice with a path).

examples would be:
long: `localhost:10000/app_9d739324a920463facbf13bd3a4560b7`
short: `localhost:10000/app/test`

The only way we could get rid of the `/app` is if we made it so that the server was the default and somehow got MinIO to behave under these conditions (probably would need to proxy everything through the server) and then we'd need to make sure that is backwards compatible. Personally I think its probably safer to have everything going to a server under a path like `/app` as it gives us a lot of consistent control. The `/app_` special case does essentially the same thing, its just that we know for certain all app IDs will start with it.

Rory-Powell: Got it, thanks! I actually prefer the /app prefix, it’d be nice to just drop the special logic for /app_ in envoy and have the routing be consistent, keeping both approaches is fine too and probably required for backwards compatibility

mike12345567: Have to agree I'd like to drop it too, the only reason it was there was for backwards compatibility, it was built that way during the transition from the original electron builder to what it is today as the OG server handled routing that way. I think we leave the rule in place for a little while until the short hand is fully adopted and then we could get rid of it.

PR #3932, opened 2022-01-05
Fixes for new rest datasource

Rory-Powell:
## Description
Fixes for new rest datasource

### DIFF
--- a/packages/bbui/src/Tabs/Tabs.svelte
+++ b/packages/bbui/src/Tabs/Tabs.svelte
@@ -102,7 +102,7 @@
     border-bottom: none !important;
   }
   .onTop {
-    z-index: 1;
+    z-index: 20;
   }
   .spectrum-Tabs {
     padding-left: var(--spacing-xl);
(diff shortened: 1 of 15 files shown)


packages/bbui/src/Tabs/Tabs.svelte
```diff
@@ -98,7 +101,9 @@
   .quiet {
     border-bottom: none !important;
   }
-
+  .onTop {
+    z-index: 1;
```
mike12345567, line 105: Is this z-index going to be high enough consistently? If its for dropdowns etc it might need to be something like `20`.

mike12345567 (approved):
LGTM! Good catch on the invalidation of dynamic vars when a query is deleted.

packages/bbui/src/Tabs/Tabs.svelte
Rory-Powell: Sounds good, added in https://github.com/Budibase/budibase/pull/3932/commits/c0006b4d8f6ff1d5d03734d71a5b94e57d84db60

Issue #3740, opened 2021-12-09
Adding column text type to a SQL table when long form text


diogenesbrussels:
**Describe the feature request**
When adding a Long Form Text to a MySQL table, budibase creates a column typed VARCHAR(255) maybe it should be a TEXT type instead?
I changed it manually but it would be nice if it was done when creating the column.

Issue #3667, opened 2021-12-03
"Skip" Button for fetching internal table on fetch datasources


mjashanks:
Comes from this: https://github.com/Budibase/budibase/discussions/3650#discussioncomment-1739001

To create a SQL datasource, we must "Fetch Tables from Database"

![image](https://user-images.githubusercontent.com/3524181/144598882-8888bfc7-1594-43f2-b3ae-146a1928471a.png)

If we have an enormous amount of tables, then it times out. The workaround is to point it to a smaller database, and then change the connection details.

An easier workaround would be to add a "Skip" button. 

I think this is an unusual edge case that we have a workaround for, so probably not a massive issue. I will let you all decide.

Issue #2843, opened 2021-10-01
Attachments are not deleted when a row is deleted


mjashanks:
When we delete a row that has an attachment, the attachment is not removed from MinIO.

It should be. 

However - how should this work for `dev` vs `published` apps? Don't they share the same attachment?

shogunpurple:
They do share the same bucket to avoid creating 2 and having to upload all the "dev" attachments into prod when you deploy. I would say it makes sense to:

- if you delete the row in production, it deletes the row, and also the attachment from the bucket
- if you delete the attachment row in dev, delete the row but keep the attachment, so you can revert if you need to

PR #3970, opened 2022-01-07
Various fixes

mike12345567:
## Description
Fixing the following:

1. #3667 - allow skipping table fetch when creating a datasource incase working with a very large number of tables.
2. #2843 - when working with attachments, they will be deleted when they are not in use, handled during row update, deletion and bulk deletion. The rules around this are a little weird because the bucket is shared for production and development, when an app is not deployed attachments will automatically be deleted, but when an app is deployed attachments will only be deleted if they are deleted from the production app. This means if you want to delete attachments from the builder for a prod app you need to unpublish it, delete and then re-publish.
3. #3740 - updating SQL table/column creation to use text type rather than a 255 character limit varchar.

### DIFF
--- a/packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
+++ b/packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
@@ -5,22 +5,28 @@
   import { IntegrationNames } from "constants/backend"
   import cloneDeep from "lodash/cloneDeepWith"
   import { saveDatasource as save } from "builderStore/datasource"
+  import { onMount } from "svelte"
 
   export let integration
   export let modal
 
   // kill the reference so the input isn't saved
   let datasource = cloneDeep(integration)
+  let skipFetch = false
 
   async function saveDatasource() {
     try {
-      const resp = await save(datasource)
+      const resp = await save(datasource, skipFetch)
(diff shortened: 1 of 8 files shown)


packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
```diff
@@ -5,32 +5,45 @@
...
 <ModalContent
   title={`Connect to ${IntegrationNames[datasource.type]}`}
   onConfirm={() => saveDatasource()}
   onCancel={() => modal.show()}
   confirmText={datasource.plus
-    ? "Fetch tables from database"
+    ? "Save and fetch tables"
     : "Save and continue to query"}
   cancelText="Back"
+  showSecondaryButton={datasource.plus}
```
Rory-Powell, line 40: Glad to see this is getting used 😃  wasn't sure if it was overboard adding yet another button but was the best place for doing the rest import

Rory-Powell (approved):
Looks great. Clean and lots of good comments 🚀

shogunpurple (approved):
Some really nice stuff here. Especially the attachments! I like the way the utility function is in its own module too - makes it pretty easy to change that behaviour through an env variable or something - like if we wanted to archive attachments instead of delete them or similar.

PR #3989, opened 2022-01-11
Invalidate dynamic variables when they are removed from datasource

Rory-Powell:
## Description
Similar follow up fix to https://github.com/Budibase/budibase/pull/3932/commits/6f53b25116568a4928158a3f265fb329c6e8568e
- invalidate variables when they are removed from a datasource
- tests for the above
- tests for the existing logic to remove and invalidate datasource variables when the linked query is deleted


mike12345567 (approved):
LGTM!

Issue #3228, opened 2021-11-03
Access to parameters in query transformer


phaseOne:
Great work on #2992! I've got a suggestion for an elaboration of this feature:

In an external data source JS transformer, it would be lovely if I could access the value of a parameter that's passed in by the component that's running the query on the data source.

This is especially important for data sources where you can't pass parameters into the query, like Airtable, or where you don't control the API and need to perform additional filtering.

For example, if I'd like to extract all books by a given author from the data that's returned by the data source, the JS function needs the `author` parameter so that it can run a filter function.

<img width="667" alt="CleanShot 2021-11-02 at 17 44 10@2x" src="https://user-images.githubusercontent.com/1310985/139970787-05fd344e-fe05-4575-9d40-57a0af2a8f99.png">

As point of reference for the design, Airtable has [a scripting "action" in their automation feature](https://www.airtable.com/developers/scripting/api/input#config), and they use an `input` object that contains information from user-defined parameters when you call `input.config()`.

<img width="1089" alt="CleanShot 2021-11-02 at 17 56 45@2x" src="https://user-images.githubusercontent.com/1310985/139971947-d0b96990-ff0e-4fb1-af6d-37ad0cfdca63.png">

PR #3990, opened 2022-01-11
Allow access to query parameters in transformer

mike12345567:
## Description
Fixing #3228 - allowing access to parameters in the transformer.

This was a small change, noticed it wasn't mentioned in the docs, then realised that it hadn't actually been enabled.


shogunpurple (approved):
LGTM!

Rory-Powell (approved):

PR #3991, opened 2022-01-11
Fix: Load apps when using the default multi tenant

Rory-Powell:
## Description
- Apps weren't loading when using the default tenant in multi tenant mode 
- Just load all apps in this case and rely on subsequent application side filtering to produce the correct list of apps
- Would not be performant if there were a giant number of tenants and the default tenant was being accessed, but this is a) not supported in cloud (we don't use the default tenant) and b) not supported in self hosted (we don't officially support multi tenant self host)


mike12345567 (approved):
LGTM!

Issue #3722, opened 2021-12-08
Only the first Bash/JS Action is executed


JMLX42:
**Describe the bug**

When creating an Automation, if two (or more?) Bash Actions are chained, the second (and following?) actions are not executed or have the same output has the 1st one.

**To Reproduce**

1. Go to Automate
2. Create a new Automation
3. Select Row Created as a trigger
4. Add a Bash Action with the following code: `echo first`
5. Add 2nd a Bash Action with the following code: `echo second`

**Expected behavior**

The output of the 2nd Bash Action is :

```json
{
  "stdout": "second\n",
  "success": true
}
```

**Actual Behavior**

```json
{
  "stdout": "first\n",
  "success": true
}
```

**Screenshots**

![image](https://user-images.githubusercontent.com/998703/145244231-67c650b0-188a-4f76-848a-312c883618e3.png)

![image](https://user-images.githubusercontent.com/998703/145244292-4f7490df-4303-4b31-8cf5-176484d9698a.png)

**Desktop (please complete the following information):**

 - OS: Fedora 34
 - Browser: Firefox
 - Version: 94.0

yeoldegrove:
Not only the "bash action" is affected...
I have a use case where I use chained "update row" and "external data connector" and only the "update row" action is executed.

PR #3998, opened 2022-01-11
Fixing issue with multiple automation steps of the same type showing same response

mike12345567:
## Description
There was a frontend visual issue denoted in #3722 - when automation steps of the same type were added to the same automation the test results would show the same for each step - this was due to how the frontend was selecting the test result for each step (it was just using the `stepId` which is the type of step, rather than the unique `id` property).

Also performed some quick visual updates to the automation frontend, there was a custom button used which was easily replaced with an `ActionButton` and used the same action button in the automation code editor which had all bindings displayed within it, allowing collapsing of the bindings to make the editor a bit cleaner to look at.

## Screenshots
image
Setup was previously a div with some custom stuff going on to flip the chevron, replaced with an `ActionButton`.

image
Automation code editor, with bindings closed.

image
Automation code editor, bindings opened.


shogunpurple (approved):
LGTM!

PR #4004, opened 2022-01-12
Refactoring core library usage

mike12345567:
## Description
Refactoring core library usage throughout backend services, make it a bit cleaner/easier to search/more standardised. The previous usage was a little non-standard and also not used consistently.

I believe our core library has become too large to be entirely exported from a single file and then broken down, this isn't so important for it being a backend library, but when a frontend library comes around for code-splitting purposes we should build it in an modular manner and it'd be good to follow one convention throughout the entire codebase.

Ideally node libraries should only have a singular export and the idea is that once its grown to such a large scope it should be broken into lots of smaller scoped libraries/modules, however I think that would introduce an awful lot more maintenance and headache for us, given its simply a backend core library breaking it up into modules this way is a lot of easier for us to maintain without needing to create 10 new packages.

There is one significant benefit for the backend which this brings out, mocking of the core libraries is significantly easier when it is broken up. Previously if you wanted to mock a set cookie function in the core library, you'd have to mock out the entire core library, which could cause a lot of issues if you mock something that is required for the test to complete (say some other utility function) now you can simply mock the component of the library that you want to mock, rather than the whole thing - see example below.

```
// before
jest.mock("@budibase/backend-core", () => ({ constants: { Cookies: {} } })
// now
jest.mock("@budibase/backend-core/constants", () => ({ Cookies: {} })
```

I also tried out a few different methods of providing the actual export files as it would be great if they didn't need to be at the root of the library, however any other method (for example the `exports` property of `package.json`) didn't seem to work correctly and would often break code assistance and so it felt like a pointless step simply to clean up a directory.


shogunpurple (approved):
LGTM! Agree on the import paths - it's much cleaner to have those subfolders for sure rather than just a huge top level import.

Issue #4010, opened 2022-01-13
SQL DB Column 'Text' Made as Data Type 'Options' on the Platform keep disappearing when re-fetching database tables


superbolt-x:
Hi guys!
I love your project!

I know you fixed the disappearing relationships when refetching the database tables.
I wonder if it's possible also to keep the configuration changed on the platform for the columns. I have some text columns on my PostgresDB that i need to be treated as 'Options' on BB but when i re fetch tables my Options List disappear and the columns are Text type again

**To Reproduce**
1. Connect a Database (Postgres in my case)
2. Change a Text column of the DB table to the BB data type 'Options' and list your Options in 'Options (one per line)' TextArea
3. Save and confirm that your DB Col is being interpreted as 'Options' and not 'Text'
4. Now go to the DB connection section and refetch the tables
5. Go back to the table having the column of the type 'Options' mentioned above and check the type - it should have been reverted to 'Text' by the fetching operation

**Expected behavior**
ReFetching tables should remember the configuration of the columns modified on BB - at least the 'required' and/or the 'data type'

 - OS: using docker-compose on a AWS EC2
 - Browser : Brave

**Additional context**
Am needing this for a option filter of a form

Thanks!

Lorenzo

mike12345567:
Hi @superbolt-x - we've just put out a release that should solve this issue for you.

PR #4019, opened 2022-01-13
Maintaining options and boolean types when fetching SQL schemas

mike12345567:
## Description
Fixing issue #4010 - options and boolean types can be maintained - checks the same way we do for relationships.

Also ran a TS lint which hadn't been run in a while, accounts for most of the changes.


shogunpurple (approved):
LGTM!

PR #4022, opened 2022-01-13
Fix return url

Rory-Powell:
## Description
Fix return urls by introducing a new `budibase:returnurl` cookie. This cookie is set in three places
- Prod apps: By the client when there is no session before redirecting to login (change existing logic)
- Dev apps: By the server in a dev app when we there is no session before redirecting to login (we already intercept the client in this case)
- Portal: When there is no session and we are not in a pre-login page (change existing logic)

Why cookie based?
- cross cutting way of handling redirects
  - don't need to worry about passing url params around auth pages 
- keeps all redirect logic inside budibase
  - don't need to forward params to/from sso
  - don't need to forward params to/from account portal

Known caveats
- return urls don't override in the builder after logout
  - e.g. if I logout and paste a url but don't then complete login, if I then paste another url I will be redirected to the original url on login
  - this doesn't affect direct links to apps because the client always overrides the return url 
  - this happens because the return url cookie is only set when it doesn't exist, this is required so that we don't always override it with the base route `/builder` which triggers before the redirect takes place

Tested:
- in chrome, firefox, edge, safari, using the various self hosted / account portal / custom sso flows. 
- using a fresh browser, an expired browser and a browser post logout
- in prod apps, dev apps and in builder

## Description
To app after logout
return-url-app

To builder after expiry
return-url-builder-expiry


shogunpurple (approved):
LGTM, this is a nice experience 👍

aptkingston (approved):
Awesome stuff Rory 👍 Few `console.log` left in the builder top level `_layout.svelte` file. We're all guilty of this from time to time haha.

Rory-Powell:
@aptkingston thanks for the heads up 👍 Removed in https://github.com/Budibase/budibase/pull/4022/commits/5f65c704014b36aabc6702ddc559058486937beb

Issue #3675, opened 2021-12-04
Validation for Date Pickers does not work correctly if more than one rule set


sergey-dme:
Hi All,

After the latest update validation & Custom validation for Date Pickers started working strange. 
I always have a Form validation as a first action on the button before it saving. 
So, if a date picker has wrong value or empty (depending on the rules set) previously, the red label with txt appeared below form component once value/date chosen or when button with Form Validation step pressed. Now red labels don't appear on date pickers. Only sometimes when the button pressed once again. Very strange...
So, the rules are still working (still unable to proceed to the next action/step until value match the rule) just no red labels shown. 
For other Form components it works okay.
Current version 1.0.6

Please help.

sergey-dme:
![TestValidation](https://user-images.githubusercontent.com/53024832/145196188-6a5a2c35-e571-49b0-a231-92fd7fb27952.PNG)

I hope this will help with reproducing the issue. if leave only one condition (required), it works. With two like on the scree it does not work. no red labels with error txt

sergey-dme:
v. 1.0.35 the issue is still persists.
Hey @aptkingston I know the team is working hard and I apologize for mentioning you, but could you please take a look at this?

aptkingston:
Hey @sergey-dme. Thanks for your patience. I have a PR up to fix this 👍

PR #4033, opened 2022-01-14
Fix dynamic validators not correctly validating form components

aptkingston:
This fixes an issue reported in #3675, where sometimes validation did not seem to work.

The error was actually caused by having `{{ now }}` inside a validation rule - because any time this is enriched it is a new timestamp, therefore causing the validation rules to change and the field to re-register itself. There was an oversight in field registration in the form logic that always wiped errors when re-registering an existing field.

This has been updated so that whenever an existing field is re-reigstered and it used to have an error, the new validator function is invoked and an initial error can be set if required.

### DIFF
--- a/packages/client/src/components/app/forms/InnerForm.svelte
+++ b/packages/client/src/components/app/forms/InnerForm.svelte
@@ -141,8 +141,18 @@
         return
       }
 
+      // Create validation function based on field schema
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):
Nice, LGTM!

PR #4027, opened 2022-01-13
sameSite and secure cookie settings

shogunpurple:
## Description
Adding `sameSite: none` and `secure` flags to cookies so they work on the latest version of firefox when running over HTTPS.

## Screenshots
_If a UI facing feature, some screenshots of the new functionality._

PR #4061, opened 2022-01-17
Revert "sameSite and secure cookie settings"

shogunpurple:
Reverts Budibase/budibase#4027

Reverting this as its not working correctly on pre-prod. The use of `samesite` and `secure` together expects HTTPS, and even in preprod with HTTPS, it still doesn't work correctly and throws errors. 

Will re-evaluate solution so it works across browsers and in dev etc.

### DIFF
--- a/packages/backend-core/src/environment.js
+++ b/packages/backend-core/src/environment.js
@@ -6,13 +6,6 @@ function isTest() {
   )
 }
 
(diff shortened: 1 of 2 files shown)


mike12345567:
Great to see pre-prod already showing its value - perfect for these sort of test cases.

PR #4073, opened 2022-01-17
SQL relationship filtering

mike12345567:
## Description
This feature introduces the ability to define static and dynamic filters on related SQL tables, allowing a syntax like `WHERE table.property = value` on a joined SQL statement.

The backend changes required for this were fairly minimal, most of the work had already been done with supporting this in mind, it simply required moving a few things around but for all intent and purposes it was just adding a few more statements to the where clause.

The majority of the work needed was required in the frontend, thanks to @aptkingston here for the help on this as some of the client work required needed a bit of thought/discussion. The main issue here was that the schema used for searching needed to be different to the schema used for everything else, currently the data provider has a single schema which it works from, but when filtering a expanded schema needed to be used for the options.

The following work was required in the builder/client:
1. In the builder a new settings component was added a `SearchFieldSelect` which is like the `FieldSelect` component but for SQL tables is expanded to include the related fields.
2. The `FilterDrawer` was updated to work out the related fields for SQL tables and allow these as an option when using the `Define filters` button on a data provider/block.
3. The client implementation of the dynamic filtering, table block and card block components had to be updated to make them aware of the schema of related fields.
4. This has required a few extra network requests from the client, when working with related tables the client needs to request the definitions of the related tables as well to use their schemas, the client is not able to request all tables at once as that is a builder only feature.
5. A minor update was required to the syntax of the handlebars used to pair up automated search column fields generated by search capable blocks - this work was centralised into the `fetchDataSourceSchema` function of the SDK context, so that it could be re-used by other future components that are search enabled. The update was required due to the presence of a full top in the field name.

It should be noted that due to the join configuration used the filtering occurs after all joins have been completed, meaning it isn't possible to filter on a specific foreign keys joins. There is an edge case scenario like:
1. Table exists with two foreign keys relating to the same table, e.g. the table is `product` and is related to the `staff` table, with two keys `qaAssessor` and `assignedTo`.
2. The user wants to filter only on the `assignedTo` field, to find the staff that are assigned to the product, but doesn't want that to include the staff that are QA assessing the product.

In our case, the scenario above would be very difficult to implement as it requires the user to make us aware when they are building the filter that they want to perform one of the joins with a filter. It would require some fairly complex changes to the backend to be able to support a scenario like this as we couldn't use a generic join method anymore, we'd have to join based on the filter clauses. Its possible for us to implement this down the line, but for now it felt better to use what the system supports today. For a scenario like this we always have the fallback of queries, which can be used sparingly in the scenarios that we cannot support generically.

## Screenshots
Showing new options in the filter drawer.
image

Adding a related search column to a table block.
image

Table block filtered (using the display field to make it obvious what is happening):
image

Lastly showing the dynamic filter component supporting it as well.
image
image

An example containing multiple SQL relationship filters, with one being one to many and one being many to many.
image


packages/server/src/integrations/utils.ts
```diff
@@ -127,7 +126,12 @@ export function isSQL(datasource: Datasource): boolean {
   if (!datasource || !datasource.source) {
     return false
   }
-  const SQL = [SourceNames.POSTGRES, SourceNames.SQL_SERVER, SourceNames.MYSQL]
+  const SQL = [
+    SourceNames.POSTGRES,
+    SourceNames.SQL_SERVER,
+    SourceNames.MYSQL,
+    SourceNames.ORACLE,
```
Rory-Powell, line 133: 👍

packages/server/src/api/controllers/table/index.js
```diff
@@ -49,6 +49,7 @@ exports.fetch = async function (ctx) {
       ...entity,
       type: "external",
       sourceId: row.doc._id,
+      sql: isSQL(row.doc),
```
Rory-Powell, line 52: Is the variable naming right here? I see above we query datasources and call it `externalTables`, should that be `datasources`? would make more sense here as `isSql` expects a datasource and not a table. Might help if we rename the `row` here to `datasource` as row can be confused with a table row which was my first thought

Rory-Powell (approved):
LGTM 👍

packages/server/src/api/controllers/table/index.js
mike12345567: Yeah, this is just a very old function, dates back to before data was ever called rows, I'll fix it!

shogunpurple (approved):
LGTM! Can't believe how little code this took - thought that it would require a ton more. Good job

PR #4076, opened 2022-01-17
don't check for hydrationComplete

### LINKED ISSUE #4075: Empty screens tab not visible after selecting layouts tab
**Describe the bug**
After creating a new app, the screens tab becomes active, and a Create Screen button is visible. After selecting the layouts tab and back to the screens tab, the content of the layouts tab stays active.

**To Reproduce**
1. Create a new app
2. The screens tab becomes active with a new screen button and the bb-space-man. 
3. Click the layouts tab
4. The layouts tab content is visible
5. Click the screens tab to create a new screen
6. Nothing happens. 

**Expected behavior**
The new screens tab should become active.

**Screenshots**
image

 - OS: Windows 11
 - Browser: Edge
 - Version: develop branch

**Additional context**
- There are no errors in the console.
- After clicking the Automate tab at the top and then back to Design, the New Screens page loads again (until you click the layouts tab again)

mslourens:
## Description
I removed the check for `hydrationComplete` since that boolean was set to true the first time the screen was rendered. When clicking the layouts tab and back to the screens tab, the state was not updated by the url and didn't change anymore. After removing the `hydrationComplete` boolean, the state was updated correctly.

Fixes #4075

## Screenshots
image

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/[assetType]/_layout.svelte
@@ -25,7 +25,6 @@
   let previousType
   let previousAsset
   let previousComponentId
-  let hydrationComplete = false
 
   // Manage the layout modal flow from here
   let showModal
@@ -41,12 +40,6 @@
   )
 
   const hydrateStateFromURL = (params, leftover) => {
-    if (hydrationComplete) {
-      return
-    } else {
-      hydrationComplete = true
-    }
-
     // Do nothing if no asset type, as that means we've left the page
     if (!params.assetType) {
       return


shogunpurple (approved):
LGTM!

aptkingston:
Hey @mslourens! Sorry it took me so long to get round to looking at this.

While you're right that this does solve the problem, the `hydrationComplete` check in that layout is actually very important as it prevents endless cycles of `update state > update url > update state > update url > ...`.

I've pushed up a PR here with a safer way to fix this: https://github.com/Budibase/budibase/pull/4681

It's a bit confusing because we have logic in quite a few different places to update state, but the logic that needed updated in this case was the `index.svelte` file for this route. This file contains a `selectValidAsset` function which normally selects an asset of a certain type whenever landing on a URL like `.../screen` or `.../layout` (without any ID's after it).

The logic in that function is good, but it doesn't account for when no assets of a certain type exist. In this case, that's highlighted by a new app. If we don't have any screens, then we never updated state, and the builder (and preview) always think that a layout is still selected. That's why the preview app and the builder didn't properly update, as described in your issue.

Hopefully that makes some sense! It took me a while to get to the bottom of this - it's a complicated issue.

PR #4077, opened 2022-01-17
fix schema constraint mapping to validation rule

### LINKED ISSUE #4072: Text length validation does not work on forms
**Describe the bug**
Text length validation does not work on forms

**To Reproduce**
1. Create a new text column
2. Call it validation and set the text length to 2
3. Create a form and add the field and ensure the validation has been set automatically
4. Press play and fill out the form
5. Put in more than two characters in the field
6. Click Save
7. A 400 error happens

**Expected behaviour**
Pressing the save button would prompt the user to correct the field or the form would disallow any input that is greater than the desired length of the field.

**Screenshots**
v1
v3
v4
v5
v9

 - OS: Windows
 - Browser Chromium
 - Version 97.0.4692.71

mslourens:
## Description
Fixes #4072 - mapping between `schemaConstraints.length?.maximum` and validation `handlerMap` was incorrect for `maxLength` constraint. 

## Screenshots
image

### DIFF
--- a/packages/client/src/components/app/forms/validation.js
+++ b/packages/client/src/components/app/forms/validation.js
@@ -37,7 +37,7 @@ export const createValidatorFromConstraints = (
       const length = schemaConstraints.length.maximum
       rules.push({
         type: "string",
-        constraint: "length",
+        constraint: "maxLength",
         value: length,
         error: `Maximum length is ${length}`,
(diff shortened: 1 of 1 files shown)


Rory-Powell (approved):
LGTM 👍

PR #4083, opened 2022-01-18
make empty array valid when not required

### LINKED ISSUE #4041: An empty Multi-select field causes a 400 error when saving the form when it is not set as required
**Describe the bug**
400 error occurs when trying to submit an empty Multi-select field when it is not set as required

**To Reproduce**
1. Create a text field and a Multi-select field with some values in a Budibase DB table
2. Generate a form in design and press the Update form field button on the field group
3. Enter a value in the text field and leave the Multi-select field untouched
4. Press the save button
5. "An Error Occured" will flash at the top of the screen
6. In the browser console there will be a error. The error response shows: `MultiSelect	[ " is not included in the list" ] 0	" is not included in the list"`

**Expected behavior**
The form saves all other data input on the form and any Multi-select field that has not been assigned a value is empty

**Screenshots**
image
image
image
image

 - OS: Windows 10
 - Browser Firefox
 - Version 96.0.1

mslourens:
## Description
When saving a form with a MultiSelect which is not required, the value of that field is an empty array, which is currently not allowed by validation.js. This PR adds this case to the custom array validation. 

Fixes #4041

### DIFF
--- a/packages/server/src/api/controllers/row/utils.js
+++ b/packages/server/src/api/controllers/row/utils.js
@@ -61,12 +61,17 @@ exports.validate = async ({ appId, tableId, row, table }) => {
     let res
 
     // Validate.js doesn't seem to handle array
-    if (type === FieldTypes.ARRAY && row[fieldName] && row[fieldName].length) {
-      row[fieldName].map(val => {
-        if (!constraints.inclusion.includes(val)) {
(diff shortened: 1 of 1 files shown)


aptkingston (approved):
Looks great, thanks @mslourens!

Issue #4067, opened 2022-01-17
Multi-select picker - cannot set default value


mjashanks:
I am trying to set a default value on the multi-select picker, but cannot.

I have tried to set it to

```javascript
return ["Value1"];
```

`Value1` is a valid option.

When I do this, I see the console error

```
uncaught (in promise) TypeError: e.forEach is not a function
    at budibase-client.js?v=1.0.39:1:617582
```

Originally raised here: https://github.com/Budibase/budibase/discussions/4048

bastiaanterhorst:
@aptkingston I looked at this just now, but missed that you had assigned this to yourself. I hope you hadn't spent too much time on it yet, and that my solution is acceptable!

aptkingston:
@bastiaanterhorst no problem at all! The changes I made were almost identical to yours, so I'm very happy to get your PR merged in and add you as a contributor. Thanks again!

bastiaanterhorst:
This issue is resolved and can be closed.

shogunpurple:
Thanks @bastiaanterhorst!

PR #4084, opened 2022-01-18
Update MultiFieldSelect to accept a defaultValue

bastiaanterhorst:
This changes the multi select control to accept defaultValues. As these are passed in as strings (flattened arrays in the form of "1, 2, 3") they need to be split into an array to be accepted by the control.

This solves #4067

### DIFF
--- a/packages/client/src/components/app/forms/MultiFieldSelect.svelte
+++ b/packages/client/src/components/app/forms/MultiFieldSelect.svelte
@@ -20,6 +20,7 @@
   let fieldSchema
 
   $: flatOptions = optionsSource == null || optionsSource === "schema"
+  $: expandedDefaultValue = expand(fieldState?.defaultValue)
   $: expandedValue = expand(fieldState?.value)
   $: options = getOptions(
     optionsSource,
@@ -39,7 +40,7 @@
       return values
     }
 
-    return values.split(",")
+    return values.split(",").map(value => value.trim())
   }
 </script>
 
@@ -57,6 +58,7 @@
   {#if fieldState}
     <CoreMultiselect
       value={expandedValue}
+      defaultValue={expandedDefaultValue}
       error={fieldState.error}
       getOptionLabel={flatOptions ? x => x : x => x.label}
       getOptionValue={flatOptions ? x => x : x => x.value}
--- a/packages/client/src/components/app/forms/MultiFieldSelect.svelte
+++ b/packages/client/src/components/app/forms/MultiFieldSelect.svelte
@@ -20,6 +20,7 @@
   let fieldSchema
 
   $: flatOptions = optionsSource == null || optionsSource === "schema"
+  $: expandedValue = expand(fieldState?.value)
   $: options = getOptions(
     optionsSource,
     fieldSchema,
@@ -28,6 +29,18 @@
     valueColumn,
     customOptions
   )
+
+  const expand = values => {
+    if (!values) {
+      return []
+    }
+
+    if (Array.isArray(values)) {
+      return values
+    }
+
+    return values.split(",")
+  }
 </script>
 
 <Field
@@ -43,7 +56,7 @@
 >
   {#if fieldState}
     <CoreMultiselect
-      value={fieldState.value || []}
+      value={expandedValue}
       error={fieldState.error}
       getOptionLabel={flatOptions ? x => x : x => x.label}
       getOptionValue={flatOptions ? x => x : x => x.value}
--- a/packages/client/src/components/app/forms/MultiFieldSelect.svelte
+++ b/packages/client/src/components/app/forms/MultiFieldSelect.svelte
@@ -20,8 +20,7 @@
   let fieldSchema
 
   $: flatOptions = optionsSource == null || optionsSource === "schema"
-  $: expandedDefaultValue = expand(fieldState?.defaultValue)
-  $: expandedValue = expand(fieldState?.value)
+  $: expandedDefaultValue = expand(defaultValue)
   $: options = getOptions(
     optionsSource,
     fieldSchema,
@@ -49,7 +48,7 @@
   {label}
   {disabled}
   {validation}
-  {defaultValue}
+  defaultValue={expandedDefaultValue}
   type="array"
   bind:fieldState
   bind:fieldApi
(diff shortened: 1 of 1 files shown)


bastiaanterhorst:

packages/client/src/components/app/forms/MultiFieldSelect.svelte
```diff
@@ -20,6 +20,7 @@
   let fieldSchema
 
   $: flatOptions = optionsSource == null || optionsSource === "schema"
+  $: expandedValue = expand(fieldState?.value)
```
aptkingston, line 23: The best idea here would be to expand the `defaultValue` prop rather than the `value`, and then pass that in to the `Field` component, so that it gets registered with the form properly. We could just call it `expandedDefaultValue`.

packages/client/src/components/app/forms/MultiFieldSelect.svelte
```diff
@@ -28,6 +29,18 @@
...
+  const expand = values => {
+    if (!values) {
+      return []
+    }
+
+    if (Array.isArray(values)) {
+      return values
+    }
+
+    return values.split(",")
```
aptkingston, line 42: To make this setting a little more user friendly, we could trim the values in case someone enters a default value with spaces after the commas.
```js
return values.split(",").map(value => value.trim())
```

packages/client/src/components/app/forms/MultiFieldSelect.svelte
```diff
@@ -43,7 +56,7 @@
 >
   {#if fieldState}
     <CoreMultiselect
-      value={fieldState.value || []}
+      value={expandedValue}
```
aptkingston, line 59: Like mentioned above, if we leave `value` as it was and instead pass in our `expandedDefaultValue` to the `defaultValue` prop of the `Field` component instead, then we should be good to go 👍 
```js
defaultValue={expandedDefaultValue}
```

aptkingston (changes requested):
Hey @bastiaanterhorst! Thanks so much for the contribution. Your solution here is almost perfect, but rather than change the value that gets passed to the underlying `CoreMultiselect`, we need to make sure the value is actually passed in to the form itself. Otherwise the form state and actual component value won't be in sync.

I've added a few comments about how to do that, and then let's get it merged 👍

packages/client/src/components/app/forms/MultiFieldSelect.svelte
bastiaanterhorst: I have expanded both now -- just expanding defaultValue gave me the same error as before. Value is also expected to be enumerable, but unless we expand it, it is a string.

packages/client/src/components/app/forms/MultiFieldSelect.svelte
bastiaanterhorst: as mentioned above, I have expanded both value and defaultValue now.

bastiaanterhorst:
Thanks for your review @aptkingston! I made the changes, but did find that only updating defaultValue did not work -- the error I originally encountered (passing a string into forEach) resurfaced.

In this commit I expand both value and defaultValue. I hope this meets your needs, in my local test everything works as it should.

aptkingston:
Thanks @bastiaanterhorst! I can see in the code that the `defaultValue` is being passed to the `CoreMultiselect` component rather than the parent `Field` component (which would also explain why it still broke in your testing). 

Could you try just passing `defaultValue` to the `Field` component and see if that does the trick?

Although there's no harm in also expanding `value`, it shouldn't be needed because there should be no way for the value to be a string (as long as the default value is not a string).

bastiaanterhorst:
Ah, I misunderstood your comment. I'm new to this ecosystem and slightly overwhelmed by how it all works. Thanks for bearing with me.

Unfortunately I still could not make this work. When I change the field component like so:
```
<Field
  {field}
  {label}
  {disabled}
  {validation}
  {expandedDefaultValue}
--snip--
```

And restore the CoreMultiselect component to:
```
    <CoreMultiselect
      value={fieldState.value || []}
      error={fieldState.error}
--snip--
```

Updates do not work. I can select options using the dropdown, but when I programatically change options using the binding, the field does not update to reflect those changes. I am probably missing something obvious -- again apologies for being somewhat out of my depth here :)

aptkingston:
No problem at all! Trust me - it's already impressive that you've managed to dive in and contribute already. Like you said, it's a big project!

I can see the small issue - it's just how you're passing in the default value. You just need to update it to:
```js
<Field
  ...
  defaultValue={expandedDefaultValue}
--snip--
```
Doing just `{expandedDefaultValue}` like in the example you showed is short hand in svelte for `expandedDefaultValue={expandedDefaultValue}`, but the prop name is just `defaultValue`. Easy mistake to make.

Fingers crossed it works this time. And absolutely no need to apologise - even finding the right file to edit was impressive enough in a first PR haha.

bastiaanterhorst:
Thanks Andrew!

I actually tried that, but it did not work. It made my chrome pages
unresponsive when I wanted to test the result.

I got the impression that I might have inadvertently introduced some
circularity seeing that fields defaultValue is set also set using the
fields defaultValue, with:

‘’’
expand(fieldState?.defaultValue)
‘’’

So we have a field value that is being changed based on that same field
value? I could be missing something here haha!

aptkingston:
I completely missed that in the code - my bad! You're completely right, that's a loop that's locking it up. It should just be:
```js
$: expandedDefaultValue = expand(defaultValue)
```
We just want to expand the prop that gets passed in, not the default value in the field state.

bastiaanterhorst:
I've made that change and indeed now it works! Its a impressive how much 'magic' Svelte has going on 😅 , though I suppose that once you're familiar it makes it very fast to work with.

One 'issue' that I've found (I think it is a design choice, not a bug) is that there is a disctinction between value and defaultValue.
- Say I bind the state of the picker to a state variable, using the defaultValue
- When I set the state, the picker changes. Awesome.
- I can change the selection in the picker. Now the state does not update (so the binding is one way). Now the binding and the picker are out of sync.

Other than that: thank you for all the hand holding :D

aptkingston (approved):
Looks perfect 👍 

You're right that svelte has a lot of magic going on haha. In the example you mentioned, you're correct that it's a design decision rather than a bug. Our form component library in the BBUI package (which includes the `Multiselect` component used here) don't use svelte bindings to propagate changes, but instead use event dispatching (`on:change`, which you can see us passing in as a prop in this file). Part of the reason for that is to have more control over the flow of data and actually try to reduce some of that svelte "magic"!

PR #4086, opened 2022-01-18
Fixes for multi-selects and data fetching

### LINKED ISSUE #4068: Save row action - column override cannot set multi-select value
In the save row action, I want to hard-code a multi-select value, using the column override function... "Meds" in my screenshot below:
image

The Javascript function is 

```javascript
return  ["Covid"];
```

The `POST /api/rows` request body submits the value

```
{
...
    Meds: "Covid"
...
}
```

The API returns with the error

```json
{"message":"row[fieldName].map is not a function","status":400}
```

### LINKED ISSUE #4069: Can no longer create a Data Provider, that has a Data Provider as a source
We used to be able to create the following structure

```
> Data Provider 1: Data = People
   > Data Provider 2  : Data = Data Provider 1, with filter
   > Data Provider 3  : Data = Data Provider 2, with filter
```

This no longer works. In the Design preview, the child data provider displays with a spinner:
image

In the app, nothing appears.

There are no console errors or failed network requests in the app.

FYI, this was sort of raised here: https://github.com/Budibase/budibase/discussions/4050 - although they were wanting to transform the data.

aptkingston:
## Description
This PR updates string temlpates so that the result of JS bindings are the actual parsed types, rather than strings. This is important so that you can do things like use JS bindings to insert arrays for the value of a multi-select field in a button action when saving a row, or when specifying a default value.

It also fixes an issue when fetching data from certain datasource types.

- Fixes #4069 
- Fixes #4068
- Fixes part of #4067 (other part is fixed by another PR)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.27-alpha.22",
+  "version": "1.0.44-alpha.0",
   "npmClient": "yarn",
   "packages": [
     "packages/*"
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@budibase/backend-core",
-  "version": "1.0.27-alpha.22",
+  "version": "1.0.44-alpha.0",
   "description": "Budibase backend core libraries used in server and worker",
   "main": "src/index.js",
   "author": "Budibase",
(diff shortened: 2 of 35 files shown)


shogunpurple (approved):
LGTM! Nice fixes here

PR #4087, opened 2022-01-18
Fixes for rows quotas

Rory-Powell:
## Description
Based on the initial work done in https://github.com/Budibase/budibase/pull/4000

Changes:
- Decrement rows quota when table is deleted
- Increment rows quota in data import (new / existing table csv import)
- Increment / decrement rows on app actions (import, template, delete)
   - Added small lifecycle framework to quota middleware to do this to hook into:
      - pre create
      - post create
      - pre delete
      - post delete
- Run quota updates only after entity has successfully been created 
- Add corresponding dry run flag to run quota checks before creation
- Update table creation modal to notify when quota has been exceeded
- Generify quota migrations for backfills in future
- Prevent quotas from ever going negative on deletion
- Simplify local dev by:
   - Move quota usage logic out to env vars
   - Move quota exclusion orgs out to env vars
- Centralise logic around checking if quotas are enabled


packages/server/src/automations/steps/deleteRow.js
```diff
@@ -74,9 +73,7 @@ exports.run = async function ({ inputs, appId, emitter }) {
   })
 
   try {
-    if (env.isProd()) {
-      await usage.update(usage.Properties.ROW, -1)
-    }
+    await usage.update(usage.Properties.ROW, -1)
```
Rory-Powell, line 76: Note to self: should go after the delete

charts/budibase/values.yaml
```diff
@@ -93,6 +93,8 @@ globals:
   logLevel: info
   selfHosted: "1" # set to 0 for budibase cloud environment, set to 1 for self-hosted setup
   multiTenancy: "0" # set to 0 to disable multiple orgs, set to 1 to enable multiple orgs
+  useQuotas: "0"
+  excludeQuotasTenants: "" # comma seperated list of tenants to exclude from quotas
```
shogunpurple, line 97: 👌 super useful

shogunpurple (approved):
LGTM - its great to have finality on this, and the migration piece is really well done. The sync aspect is really useful in future too, if someones quota gets out of whack, we can run it again to reset them.

Not one for now but for these kinds of jobs, that may run again in future, we could keep them somewhere centralised so that we can run them directly from the app server, either through SSH or HTTP. I know we have `background-jobs`, but it requires a lot more env setup to replicate the same thing over there

Issue #3721, opened 2021-12-08
Impossible to delete an invalid MySQL relationship


diogenesbrussels:
**Describe the bug**
I'm unable to delete a budibase relationship from a deleted table.

**To Reproduce**
1. Connect a MySQL database
2. Fetch the tables
3. Create a relationship between 2 tables
4. Delete one of the tables
5. Try to delete the relationship (now named 'Cannot build name' and 'to undefined')

**Expected behavior**
We should be able to delete the relationship manually (preferred) or while fetching new tables delete old non-valid relationships.

**Screenshots**

**Desktop (please complete the following information):**
 - Self-hosted Docket
 - Firefox 94
 - Budibase 1.0.6

**Additional context**
Console output 
`Uncaught (in promise) TypeError: s is undefined
    B http://localhost:8080/builder/assets/index.9f18a315.js:442
    nX http://localhost:8080/builder/assets/vendor.470042a4.js:4
    nX http://localhost:8080/builder/assets/vendor.470042a4.js:4
    b http://localhost:8080/builder/assets/index.9f18a315.js:14
    jZ http://localhost:8080/builder/assets/vendor.470042a4.js:1
    cp http://localhost:8080/builder/assets/vendor.470042a4.js:1
    m http://localhost:8080/builder/assets/index.9f18a315.js:14
    I8 http://localhost:8080/builder/assets/vendor.470042a4.js:4
    m http://localhost:8080/builder/assets/index.9f18a315.js:442
    m http://localhost:8080/builder/assets/index.9f18a315.js:442
    m http://localhost:8080/builder/assets/index.9f18a315.js:21
    m http://localhost:8080/builder/assets/index.9f18a315.js:21
    I8 http://localhost:8080/builder/assets/vendor.470042a4.js:4
    m http://localhost:8080/builder/assets/index.9f18a315.js:442
    I8 http://localhost:8080/builder/assets/vendor.470042a4.js:4
    m http://localhost:8080/builder/assets/index.9f18a315.js:1104
    m http://localhost:8080/builder/assets/index.9f18a315.js:21
    m http://localhost:8080/builder/assets/vendor.470042a4.js:6
    I8 http://localhost:8080/builder/assets/vendor.470042a4.js:4
    m http://localhost:8080/builder/assets/index.9f18a315.js:21
    p http://localhost:8080/builder/assets/index.9f18a315.js:21
    Bx http://localhost:8080/builder/assets/vendor.470042a4.js:4
[...]

PR #4090, opened 2022-01-18
Fixing issue with existing SQL relationships and deleting tables externally to Budibase

mike12345567:
## Description
Fix for #3721 - deleting invalid relationships if tables have been removed external to Budibase - otherwise these could not be removed without deleting the datasource.

There was also some unrelated Cypress files that needed linted/eslint disables added to.

### DIFF
--- a/packages/server/src/integrations/utils.ts
+++ b/packages/server/src/integrations/utils.ts
@@ -143,16 +143,41 @@ export function isIsoDateString(str: string) {
   return d.toISOString() === str
 }
 
+/**
+ * This function will determine whether a column is a relationship and whether it
+ * is currently valid. The reason for the validity check is that tables can be deleted
+ * outside of Budibase control and if this is the case it will break Budibase relationships.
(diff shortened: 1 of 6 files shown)


packages/server/src/integrations/utils.ts
```diff
@@ -143,11 +143,21 @@ export function isIsoDateString(str: string) {
   return d.toISOString() === str
 }
 
+function shouldCopyRelationship(column: { type: string, tableId?: string }, tableIds: [string]) {
```
Rory-Powell, line 146: Can we add some doc to both these new functions? On the surface all looks good but I'm really not sure why / what they are for 😅

Rory-Powell (approved):

packages/server/src/integrations/utils.ts
Rory-Powell: Thank you! much easier to understand now

Issue #4093, opened 2022-01-18
Updating existing app causes "Database does not exist" after being published


funkymule:
**Describe the bug**
When I updated existing apps to the 1.0.43 version, then publish the app, it displays
`{"message":"Database does not exist.","status":404}`
upon opening.

**To Reproduce**
1. Open an app created in an older version (mine was 1.0.25).
2. Click the `App Version` button in the upper right corner (looks like a refresh button...is quite confusing to users).
3. Accept the update to 1.0.43.
4. (Re)publish the application.
5. Go back to Budibase home screen.
6. Click `View published app` on the newly updated app
7. A white screen displays `{"message":"Database does not exist.","status":404}`

**Expected behavior**
The newly published app opens

**Screenshots**
![image](https://user-images.githubusercontent.com/4054389/150002038-f571092d-ea0c-4174-9644-a48a4b2d1aa1.png)

mike12345567:
Hi @funkymule - I gave this a go myself, setup a Budibase environment running `v1.0.25`, created an app, added some screens, published it - then updated the environment to `v1.0.43` then upgraded the app and re-published it, when I viewed the app it was still functioning correctly.

Could you let me know a few things - 
1. We changed the way apps are accessed in the latest versions Budibase, it now accesses them via `/app/<app_name>` as the URL - I'm wondering if something went wrong with this making the app inaccessible - could you share the resource component of the URL?
2. How did you perform the update, did you do it through the command line, or did you use the update section of Budibase to update via Watchtower?
3. Just wondering if theres any console errors, or which network request is specifically failing, might help diagnose.

funkymule:
Actually, this error occurs for several published apps that were not upgraded or republished. :(
Every app that has a Data Source defined gives the same error.

funkymule:
> 

The URL is `/app/App%20Name`

sergey-dme:
I faced with the same problem after update to v1.0.42 yesterday. Unpublish-Publish-Unpublish-Publish (two times) helped in my case.

mike12345567:
Hi @funkymule - what type of datasource is it that seems to be linked to the issue? Or is it just any datasource existing seems to cause it?

funkymule:
We only use SQL Server data sources, so I can't tell you if it's due to a SQL Server data source, or just because it has a data source in general.

And...to make it more confusing...of all the applications with SQL Server data sources...ONE of them is working. :o

funkymule:
[...]

PR #4095, opened 2022-01-18
Fix for case sensitivity in client app URLs

mike12345567:
Quick fix for #4093 - make sure the App ID/App URL returned for deployed apps is always lowercase, as URLs are not case sensitive and should not be compared as such.

Previously if the app was named App Name then a URL of App%20Name would be generated in the server for comparison, which is wrong as the URL will always be case insensitive.

Merging to master to get quick fix out.


shogunpurple (approved):
LGTM!

Rory-Powell (approved):
👍

PR #4105, opened 2022-01-19
Fix log out action

### LINKED ISSUE #3968: Log Out Action does not work
I have reproduced the issue below. It actually seems that `Log out` does not work at all. When this action is triggered, I get redirected to the App portal and can continue working as normal (logged in).

### Discussed in https://github.com/Budibase/budibase/discussions/3967

<sup>Originally posted by **micebrain** January  7, 2022</sup>
Hi,

I have a public page with a button that navigate to a second page that required permission and this works perfectly fine and I can see that after login it gets redirected to the return url.
On this second page, I have a button to logout and I see that it redirects back to `/builder/app/login`. I would like to have it redirect back to the app public page.
I tried setting a `Navigate to` action this in the button action, but it doesn't seems to work. Is there a way around this?

Screenshot 2022-01-07 at 14 47 46

Thanks!

 </div>

**COS**
- Make call to worker to log out
- Don't redirect
- Add property to logout action for URL you want to navigate to

aptkingston:
## Description
This PR fixes the logout button action, and also adds a setting for a return URL, so that users can be redirected to any URL after logging out. This URL can be an app URL or an external URL.

Also fixes a crash that could happen in the server when trying to log out twice in a row.

Fixes #3968.

New log out button action settings:
image

### DIFF
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/ButtonActionEditor/actions/LogOut.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/ButtonActionEditor/actions/LogOut.svelte
@@ -13,11 +13,11 @@
       If you don't enter a value, you'll be redirected to the login screen.
     </Body>
     <div class="content">
-      <Label small>Return URL</Label>
+      <Label small>Redirect URL</Label>
       <DrawerBindableInput
         title="Return URL"
-        value={parameters.returnUrl}
-        on:change={value => (parameters.returnUrl = value.detail)}
+        value={parameters.redirectUrl}
+        on:change={value => (parameters.redirectUrl = value.detail)}
         {bindings}
       />
     </div>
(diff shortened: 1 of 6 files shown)


packages/builder/src/components/design/PropertiesPanel/PropertyControls/ButtonActionEditor/actions/LogOut.svelte
```diff
@@ -1,13 +1,38 @@
...
+
+  export let parameters
+  export let bindings = []
 </script>
 
 <div class="root">
-  <Body size="S">This action doesn't require any additional settings.</Body>
+  <Layout noPadding gap="M">
+    <Body size="S">
+      Please enter what URL you would like to be redirected to after logging
```
Rory-Powell, line 12: minor minor point I think it would read better as 
```
Please enter the URL you...
```

Rory-Powell (approved):
LGTM 👍 Any screenshots?

aptkingston:
> LGTM 👍 Any screenshots?

Only visual change is that the log out action has that one text input, but I'll throw a screenshot in the description.

PR #4112, opened 2022-01-19
Home Screen UI fixes

PClmnt:
## Description
This PR:

- Fixes the templates section so it only displays two rows no matter what viewport you are using
- Adds a "splash screen" to the home page, like automations and the design tab. 
- Removes the create app / import app modal that was shown whenever clicking `Create App` - it was no longer needed. We already have a specific `Import App button` there
 - update button sizing
 
## Screenshots

image


Rory-Powell:
Nice one peter, good to see the new empty state in as well. One thing I've noticed is a bit of a jitter when loading the app screen where the empty state loads briefly see:

https://user-images.githubusercontent.com/8755148/150290913-89d2aac3-a4db-4351-8114-b28a1fd717ce.mov

Could we use a conditional around this so that it doesn't show that state until the apps are loaded and verified as empty?

shogunpurple (approved):
LGTM! Good job. Just wondering if this may make sense as an accordion or you can click to see them all in a modal? So people can see all the templates if they want. Maybe I'm wrong but feel like it's a shame not to show everything we have. Thoughts @joebudi?

PClmnt:
@shogunpurple I did mention to @joebudi about this. He had mentioned this would be okay until we had a proper templates section. Short of that, what about randomising them everytime? I think a Modal is a decent solution as well.

mike12345567:
This looks great, much more usable! One thing I noticed is this is currently set to go into `develop` - I know the Cloud/self host is kinda difficult to use right now with the number of templates, would it be a good idea to put this to master so we can get a bug fix release out?

joebudi:
> LGTM! Good job. Just wondering if this may make sense as an accordion or you can click to see them all in a modal? So people can see all the templates if they want. Maybe I'm wrong but feel like it's a shame not to show everything we have. Thoughts @joebudi?

I agree. I didn't want to steal more of @PClmnt time on this so just left it. Eventually, a template modal will be great, but limiting the templates to 1 or 2 rows is fine until we have time to implement a template / new app / import modal

shogunpurple:
Sounds good @joebudi - let's not let it block this one, and I'll create another issue for the templates modal as there might be a few design elements involved before we actually build it.

shogunpurple:
https://github.com/Budibase/budibase/issues/4117

Issue #4126, opened 2022-01-20
Form Fields no longer work as data provider bindings


Gomeyy:
**Describe the bug**
I can no longer use a form field as a data provider binding after updating to 1.0.44, they also don't show up as options when using handlebars when setting a binding.

**To Reproduce**
1. Create a form with any field
2. Create a data provider in that form
3. Try to use that field as a binding in the data provider

**Expected behavior**
Being able to use form variables in data provider bindings

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 97.0.4692.71

**Additional context**

sergey-dme:
I guess this is why a tab with my app completely freezes after update to 1.0.44, it happens on some pages with similar structure and even complicated tree (lots and data providers. repeaters and forms inside of each other). I rolled back to v1.0.35

Gomeyy:
@mike12345567 Using a form field as a query binding not a filter 
![image](https://user-images.githubusercontent.com/22477051/150364645-a022217b-a1c7-4a0a-9f4f-1228a75a604c.png)
This is what I see when using a data provider, no bindable values 
![image](https://user-images.githubusercontent.com/22477051/150364804-629860de-065e-4c42-a4bf-afd13c8dc11d.png)
When using a headline at the same position in the hierarchy is acts as I would expect
![image](https://user-images.githubusercontent.com/22477051/150365345-c127e28b-6fe2-458f-ada8-7d438c13be42.png)

mike12345567:
Thanks @Gomeyy - I can see that happening as well - will get a fix out for it.

mike12345567:
Hi @Gomeyy - I've triggered `v1.0.45` to release which should resolve your issue.

@sergey-dme - if possible could you check if it resolves your issue as well? If not it'd be good to try and diagnose what went wrong in your app too.

Gomeyy:
@mike12345567 Okay, so the bindable values are appearing now for data providers.

But now when running a rest query on the design tab any bindings assigned for that rest query become empty?

I have a company binding on one of my rest queries and when I run it from the data tab, it uses the default value and runs fine.
![image](https://user-images.githubusercontent.com/22477051/150395433-450dcd78-2353-4be4-a379-f21278bbad49.png)

When I run the same query from a data provider/table block it doesn't use the default value, or any value I attempt to assign and seems to be running the query with that binding as an empty string.
[...]

PR #4128, opened 2022-01-20
Fix query binding drawer bindable options

mike12345567:
## Description
Fix for #4126 - wrong bindings were being passed to the drawer, making sure correct ones are available.

### DIFF
--- a/packages/builder/src/components/integration/QueryBindingBuilder.svelte
+++ b/packages/builder/src/components/integration/QueryBindingBuilder.svelte
@@ -7,27 +7,26 @@
   import DrawerBindableInput from "components/common/bindings/DrawerBindableInput.svelte"
 
   export let bindable = true
+  export let queryBindings = []
   export let bindings = []
-  export let bindableOptions = []
   export let customParams = {}
 
   function newQueryBinding() {
-    bindings = [...bindings, {}]
+    queryBindings = [...queryBindings, {}]
   }
 
+  $: console.log(bindings)
+
   function deleteQueryBinding(idx) {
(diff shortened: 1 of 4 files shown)


aptkingston:
Was just looking at this file - for some reason it has props for both `bindableProperties` (the old name) and `bindings` (the new name). I think we should just remove `bindableProperties` altogether, and just update the usages to always pass in the prop called `bindings`.

packages/builder/src/components/integration/QueryBindingBuilder.svelte
```diff
@@ -72,7 +72,7 @@
             bindableOptions,
             customParams?.[binding.name]
           )}
-          {bindableOptions}
+          bind:bindings={bindableOptions}
```
aptkingston, line 75: We don't need the `bind:` here, it's just passing them in one way so we can remove that.

mike12345567: The `bindings` in this case are different to the `bindableOptions` - the `bindings` are the Query bindings that need to have values provided for - not really sure what this can be renamed to.

aptkingston: Oh right yea, the generic naming caught me out. That's probably why `bindableProperties` didn't get refactored to `bindings` in this file in the first place.

shogunpurple (approved):
LGTM! Nice getting that fix up so quickly.

Rory-Powell (approved):

aptkingston (approved):
LGTM! Renaming makes things much more consistent.

PR #4131, opened 2022-01-20
Quick fix for data section filtering

mike12345567:
## Description
Minor fix, when testing the filtering the data section noticed some stuff didn't work, like filtering by just a number (which is actually text) we fixed this previously but had to version the API to fix it, builder needed to send up the API version it desires to use.


aptkingston (approved):
Nice one. I've fixed this too in my refactoring and core API client work but this is definitely a safer merge so we can get it in sooner!

PR #4134, opened 2022-01-20
Multiple fixes for v1.0.45

mike12345567:
## Description
Fixing an issue where old apps did not send up the values of the default parameters, the server should be able to handle this, enrich them in if they aren't there.

Discovered in #4126.

UPDATE: I've added the necessary components from https://github.com/Budibase/budibase/pull/4086 to this PR as well - to try and get a fix out for the data provider using a data provider as a source problem.


shogunpurple (approved):
LGTM!

Issue #4136, opened 2022-01-20
Default values for Option Pickers do not work correctly (being ignored)


sergey-dme:
Hello again,

I have a Form which is a filter for Data Provider/Table. There is a regular Option Picker (not multi), like a Status filed with values "Open, Closed, Reserved, etc.." The Option Picker has default value "Open". When I open a page with this form - it's okay, value "Open" is shown in Picker, but Data Provider filter just ignores it (returns all rows with Open, Closed, etc..) until I select any other option and then again "Open". The same story across entire app.

It works in 1.0.35 but does not work in 1.0.44-46
Im still on 1.0.35. 

Thanks.

aptkingston:
Thanks Sergey, will get this fixed.

PR #4142, opened 2022-01-21
Fix rare prop enrichment issue in client library

aptkingston:
This PR fixes a rare issue that could occur when component props were enriched multiple times before a single component render had taken place. This was highlighted in #4136 by way of default values for form fields not correctly updating a data provider filter.

Merging to master as it's addressing a small bug that is live.


shogunpurple (approved):
LGTM!

PR #4206, opened 2022-01-26
Revert "Fix dev app preview return url"

Rory-Powell:
This reverts commit e76ea10fc228a3da7c12da3bc9e54d3331d86440.

## Description
To fix an issue highlighted by @aptkingston:

lets say you were logged in and your session expires and you're sitting on the app list screen
- you click on an app name
- the first request the builder does is a POST to the application sync endpoint
- the server 403s it and responds with a set cookie for the return URL of the API endpoint
- builder reloads
- log in again
- builder now pushes you to the return URL, which is an API endpoint so not a valid URL to push the browser to

Fix is to revert any setting or return url on the server, with the known limitation that visiting a dev preview app without a session does not return you to the correct url after login

### DIFF
--- a/packages/backend-core/src/constants.js
+++ b/packages/backend-core/src/constants.js
@@ -8,7 +8,6 @@ exports.Cookies = {
   Auth: "budibase:auth",
   Init: "budibase:init",
   OIDC_CONFIG: "budibase:oidc:config",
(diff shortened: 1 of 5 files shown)


aptkingston (approved):
LGTM. Thanks for doing this 👍

Issue #4127, opened 2022-01-20
Cannot filter by formula field outcome


FlaminWrap:
**Describe the bug**
Cannot filter formula fields.

**To Reproduce**
1. Create a new app
2. Create a BB DB data source
3. Create a new BB DB table
4. Create two columns, Name (text) and Formula (Formula)
5. For the formula field set this as the formula `return $("Name")` this will set the formula field as the name field
6. Create at least two rows make sure at least one row has the word test in the name field
7. Create a new view and call it test
8. Add a filter
9. Where `Formula Equals Test`
10. No records appear

**Expected behavior**
The rows are filtered by the filter criteria set on the outcome of the formula column. (e.g. only showing rows with the word test in the formula column)

**Screenshots**
![image](https://user-images.githubusercontent.com/97764630/150370050-02aa6fa8-f38b-4a5a-b7e9-4b7aa83ebddc.png)
![image](https://user-images.githubusercontent.com/97764630/150368603-7bdad84c-dcc5-483d-b1fe-0e86a24696ff.png)
![image](https://user-images.githubusercontent.com/97764630/150369029-46a7fc3c-690d-4eca-9c53-afb68140ea40.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chromium
 - Version 97.0.4692.71
 - BudiBase version 1.0.44

**Additional context**
Selecting `Is Empty` returns all rows
![image](https://user-images.githubusercontent.com/97764630/150369511-cd62b2a3-5fc6-4bf7-a290-a5249210ddc5.png)
![image](https://user-images.githubusercontent.com/97764630/150369642-44d7b228-38a9-4d29-b733-f1ad74979efc.png)

FlaminWrap:
Ah I see, thank you for the speedy response

Issue #2316, opened 2021-08-09
Select relationships in data provider filters


kevmodrome:
**Describe the feature request**
As a user it would be a much greater experience if you could filter on relationships somehow. I'm finding myself running into this when building different things. It's particularly nice to have in apps where an admin assigns tasks/jobs/etc. to other users. Ideally you just want to list those rows that do not have an "owner" if that makes sense.

I tried a workaround where I added a formula field that checks the length of the relationship but that doesn't work either (probably for the same reason it's not possible to filter on relationships?).

mjashanks:
I'd just like to note that this is different from: https://github.com/Budibase/budibase/issues/2876

This, I believe, is a lot simpler. 

In the example `One Project has Many Tasks`, I want to be able to
- Set up a `Data provider` for `Tasks`
- Filter the dataprovider to only show Tasks that belong to the project that is selected in a options/relationship picker. 

I feel like we can already do this on the backend, however, it is not possible to apply this filter on the frontend. Relationships are not selectable in filters.

gvlx:
Hi, this is similar to #3344.

yeoldegrove:
Are "formulas" also covered by this issue or would this be easier to implement already?

mjashanks:
Just writing a note here as I had an idea that I don't want to forget.

Filtering by relationships on internal tables is a problem because relationships exist as a separate document in CouchDB. The reason that it is done this way is so that we can support Many-To-Many relationships - every relationship internally is a many-to-many underneath.

It seems that we could simplify filtering (though it still might be a big change), by changing 1-N relationships to be stored as a field in the child document.... i.e. `MyParentTable_id : "row_1234abcd"`. This structure would make 1-N filtering trivial (I think).

I believe that this would cover the majority of uses. We could decide not to support filtering on many-to-many for internal tables (or deprecate many-to-many for internal tables entirely).

Another benefit of this is that it's an easy-to-explain limitation of internal tables, making it easier to describe when it's a good time to choose SQL.

mike12345567:
Hi @mjashanks sorry I didn't see this until now, had a discussion with @shogunpurple about this early this evening, this is an approach we considered before ever building the relationship system, but there are a few troubling issues with it. While it would definitely solve the filterability of internal relationships, it could open up some serious problems we'd have to solve.

[...]

PR #4214, opened 2022-01-26
Feature/formula filtering

mike12345567:
## Description
This introduces a relatively useful feature which should help with #2316 and adds a solution for #4127. This allows creating a "static" formula, which is calculated when the row is saved - meaning that the value of related rows can also be stored into the row.

There are a few key features to note about this:
- There is now an option when creating internal table formula columns to select between "Dynamic" and "Static". A dynamic formula is what we have today, it will be calculated on the way out but cannot be filtered - static is the new feature. The reason for this is that it would be very difficult to migrate all existing formula to this new format and it would also be breaking, if someone used a formula like `{{ now }}` it will only be the value each time it is saved, whereas with dynamic it will always be the now time (this isn't a great example, but the idea is it will always be re-calculated with dynamic).

- The formula aren't truly static, there are a few situations in which the formulas will be updated (namely around relationships) - if a value is updated in the same row it will automatically be recalculated, but the more advanced feature is that when related rows are updated formulas will be calculated. The way this works is that when a formula column is added to a table we find any relationships used in it and then make a mark against the related tables that they are in use, via the `relatedFormula` property that has been added to tables. When a row is saved, if it has a relationship column and the `relatedFormula` contains the same table ID it will iterate through all the rows which are related and update their static formula - this makes sure they are always in sync. This is a bit of a complex flow so I'll try to explain it a bit better:
  - Two tables are related, table A and table B - on column `relationship`. Table A has a formula column, called `formula`
  - A row is written to table B, it is related to two rows within table A
  - After the rows have been saved and relationships updated for table B the row controller checks if any formula updates are required
  - It finds that the table B definition contains the ID of table A in its `relatedFormula` array
  - It pulls the rows from table A which the saved row is related to and enriches them
  - It runs the formula calculation over them as if the rows from table A had just been saved
  - When the rows from table A are fetched the formulas will contain the in-sync values from the related table B rows

- The way in which static formulas are recalculated on relationships is not as perfectly optimised as it could be, it has been written with consistency in mind. There are a few edge case scenarios in which the recalculation could be cut short, for example it could check if the specific properties used in the formula were updated in the saved related row, but right now it will just always try to enrich the related rows then recalcuate the formulas. I have made one small improvement to reduce the write usage - if the row after calculation has not changed it will not attempt to write it back to Couch.

- There are a few scenarios in which table operations will also update static formulas - these are primarily to do with when the formula is changed, or when it is made invalid. Right now if one of the columns used in a formula, from the same table or related, is removed it will clear the formula column. The column will be left in place so that the value is similar to how dynamic formulas would be - in their case if the property isn't available it will just be empty.

- When a table is updated to include a static formula column, or when the static formula is changed, it will attempt to calculate a value for every row in the table. This is a very intense operation when working with a very large number of rows, I've tested this locally against a table with 10,000 rows in it, I found that it took anywhere between 10-20 seconds to complete the formula calculations. This feels like the correct functionality and I believe in self host it should be allowed, if the user desires to let the calculation run they should be allowed to do so - if however this proves problematic in the cloud we can limit it so that a static formula can only be added to an empty table.

- I've limited this feature to internal tables only, while it is possible to add dynamic formulas to datasource plus tables I don't feel these static formula make sense; it feels like the wrong approach to use for something like Postgres or MySQL. In these cases if the user desires to have a calculated column then they should make use of computed/derived/virtual columns in the database directly - at some point we could support adding them through Budibase directly but I think this is a niche requirement.

- One other note is that when back-filling a large number of rows I found that the JS formula are much slower to respond than the handlebars based ones, this makes sense as for each formula we have to re-configure the sandbox, but again this could be an area for optimisation if we desire - also proof that transformers/backend JS are a CPU heavy operation.

## Screenshots
A video showing the flow of setting up a static formula on a small amount of data (with relationships) then adding a table block and filtering by the static formula which is configured on the relationship. Also shows that updating relational data affects formulas in other tables.
https://www.loom.com/share/bbecbbb2669f4a3c92640c11a8f48020

Standard config for a static formula column - not there is a tooltip describing the difference between static and dynamic.
image

Tooltip contents.
image

Response times for a 10,000 row table having a static formula added to it, not that the search also takes some time to respond immediately after the bulk write, this is due to the search re-indexing all of the row data.
image


packages/bbui/src/Tooltip/TooltipWrapper.svelte
```diff
@@ -0,0 +1,60 @@
...
+  import Icon from "../Icon/Icon.svelte"
+
+  export let tooltip = ""
+  export let size = "M"
+
+  let showTooltip = false
+</script>
+
+<div class:container={!!tooltip}>
+  <slot />
```
Rory-Powell, line 12: Good shout 👍

packages/server/src/api/controllers/row/staticFormula.js
```diff
@@ -0,0 +1,156 @@
...
+              relatedRows[tableId].map(related =>
+                exports.finaliseRow(appId, relatedTable, related, {
+                  updateFormula: false,
+                })
+              )
+            )
+          }
+        }
+      } catch (err) {
+        // no error scenario, table doesn't seem to exist anymore, ignore
```
Rory-Powell, line 64: Is there anything more specific we can do here to ensure this is why the error has been thrown? There's a fair bit of logic in `finaliseRow` that could go wrong and give us a false ignore here.

If there is an error here that we shouldn't ignore I wonder what's the best approach, perhaps to continue with the updates as much as possible but flag that the operation was 'completed with some errors' and flag that to the user?

packages/server/src/api/controllers/table/bulkFormula.js
```diff
@@ -0,0 +1,184 @@
...
+    formula.push(column.name)
+  }
+  return formula
+}
+
+/**
+ * This functions checks for when a related table, column or related column is deleted, if any
+ * tables need to have the formula column removed.
+ */
+
```
Rory-Powell, line 40: whitespace

Rory-Powell (commented):
Would be great to see the long running operation like the formula / relationship re-calculation get moved out to async jobs that we can monitor, would be good to start recording use cases like these. I'm not sure if we have a ticket or docs around this, for the proposed 'make worker a real worker' task. @shogunpurple does anything exist for this?

Rory-Powell (approved):
LGTM 👍  Just the point on error handling I think we need to address

packages/server/src/api/controllers/row/staticFormula.js
mike12345567: Good catch about the `finaliseRow` errors, I think in this error scenario it will need to just error as we have no real way to alert the user to a partial completion. This is an issue that occurs in a few places, like bulk deletions since we have no way to perform database ACID transactions.

This is something that we could address with the Budibase bulk/long running/transaction system, with bulk updates have a consistent approach and API response for when errors occur.

Rory-Powell: Alerts aside, is there anything we can do in the mean time to check this error was caused by the table not existing anymore, do we get a 404 response anywhere?

Rory-Powell: For alerts in the mean time, it would be possible to allow the iteration to complete and mark non expected errors in an array. Then do a final throw if there are present, so that we can return those errors to the UI

mike12345567: For now I'll just move the finalise rows out of the try catch, that way it'll only catch the error of the table not existing - its an edge case scenario that shouldn't occur but if it does then we can just move on, the relationship isn't present anymore.

As for alerts, I think its probably not the best idea to do something about this specific case, as its a problem in multiple places, its probably best to try and resolve the issue of bulk operations having error scenarios in one go, this applies to lots of things, like adding/updating relationships, bulk deleting rows, adding/removing columns from a table etc.

Rory-Powell: That should cover it then as it'll error out on any other scenario, we shouldn't need the alert then. 

All for a global strategy for error handling in bulk ops, was just mentioning above so we can get some form of protection in the mean time. Not even sure it needs to be coupled to the worker piece, we could even add some wrapper behaviour like:

```js
/**
* Perform an operation that impacts many rows. 
* Because CouchDB does not support ACID transactions there
* needs to be some allowance for failures, in which case we will notify the user 
* that the operation was not completely successful 
*/
bulkOperation = async (ctx, operation) => {
   try {
      await operation() 
   } catch (err) {
      if (err.partialError) {
         // don't completely fail
      }
      if (err.errors) {
         // add to body
      }
      if (err.status) {
        // set the desired status
      }
      if (err.message) {
        // add to body
      }
   }
```

or something similar. I can move this out to a separate task for now and reference this PR as an example of somewhere we could benefit.

Rory-Powell: Added: https://github.com/Budibase/budibase/issues/4270

PR #4220, opened 2022-01-27
Migrations 2.0

Rory-Powell:
## Description
Add a migration framework
- Env variable `HTTP_MIGRATION` to turn on http migrations
   - This is the default in cloud where we will run migrations manually
   - This is not the default in self hosted where migrations will run on startup
      - If migrations fail on startup the server will shutdown
- Add API for running migrations 
   - `POST api/migrations/run`
      - Runs the migrations, accepts force flags for individual migrations and tenant filtering
   - `GET api/migrations/definitions`
      - Return the migration definitions (for building a UI)   
- Add two types of migration
   - `global`
      - Run once per tenant,  migration is recorded in the global db
   - `app`
      - Run once per app,  migration is recorded in the app db
- Convert existing on demand migrations to use new framework
   - usage quotas
   - user email view casing
- Add new migration to backfill app urls   

## Screenshots
https://user-images.githubusercontent.com/8755148/151345942-ec103987-3708-40fd-9965-ab014000bd29.mov


packages/server/src/api/routes/migrations.js
```diff
@@ -0,0 +1,14 @@
+const Router = require("@koa/router")
+const migrationsController = require("../controllers/migrations")
+const router = Router()
+const { internalApi } = require("@budibase/backend-core/auth")
+
+router
+  .post("/api/migrations/run", internalApi, migrationsController.migrate)
+  .get(
+    "/api/migrations/definitions",
+    internalApi,
```
mike12345567, line 10: Should these be runnable in someway from the actual app? Right now looks like they'll be restricted to only being used on requests that have access to the API key.

packages/server/src/migrations/functions/quotas1.ts
```diff
@@ -0,0 +1,20 @@
+import { runQuotaMigration } from "./usageQuotas"
```
mike12345567, line 1: Just to check, why is this file called `quotas1.ts`? is there another file somewhere?

mike12345567 (approved):
LGTM! Comments are just questions of interest!

packages/server/src/api/routes/migrations.js
Rory-Powell: Right now that's the intention, so that authenticated users can't run migrations which is especially important for us in the cloud. 

I do like the idea of eventually moving stuff like this into the product though, similar to how updates can be invoked from the UI. A 'Server Management' page in the admin portal would be cool. Ideally you'd want to be able to see the live stream from the migration log there too, I'm not sure how we'd achieve it exactly in a multi clustered setup, might have to use s3 / minio as a log store which in itself would be a decent feature.

Rory-Powell: If we wanted to do it without the API key we might need some other notion of a 'cluster admin' user that can perform tenant wide operations

packages/server/src/migrations/functions/quotas1.ts
Rory-Powell: It was just to put in place a pattern as we are going to need more quota migrations in future as we backfill new quotas (or fix existing ones)

shogunpurple: Not a big deal - but generally this structure is a bit more easy to grok:

`quotas_11_04_2021`

shogunpurple (approved):
Great job - this is crucial going forward. Nice work!

PR #4223, opened 2022-01-27
Frontend core

aptkingston:
## Description
This PR adds a new `frontend-core` package, which contains frontend code shared between the builder and the client library.

The new package contains:
- a configurable API client
- data fetching logic for searching/filtering/sorting rows from datasources
- utilities and helpers (including some constants, cookie utilities, local storage svelte store etc)

**This needs a lot of testing, because hundreds of files and every single API interaction in the builder has been refactored. I've tested this of course but I'm sure some endpoints remain untested. We definitely shouldn't merge this until after the next release. I'll keep this up to date with develop. This is pretty much un-reviewable due to the nature of the refactor, but the most useful thing by far would be checking out the code and running it and just confirming everything works as intended.**

## Core API client
The core API client requires configuration, so the `frontend-core` package exports a constructor called `createAPIClient`. This is invoked with configuration options, and returns an API client with all endpoints defined top level. The configuration options that can be passed in are:
- `enableCaching` used to allow caching of certain cacheable endpoints that return data that does not change at app runtime. This is disabled by default and enabled only for the client library.
- `attachHeaders` used to configure which headers should be attached to every request. We get the app ID from a store in the builder and from the window object in client apps, for example, and then attach it to every request.
- `onError` a hook for handling errors. This is called before the error is thrown back to the calling code, allowing a chance to do something first, such as log out if a 403 is ever received.

## API refactor
The client library has been refactored to use the new API client from `frontend-core`. It already had its own API client properly defined so this was not a large refactor.

The builder, however, needed a massive overhaul. Previously in the builder codebase, API calls were made by using a shallow wrapper around `fetch` and passing in URLs directly. This has all been refactored now and every endpoint has been added to the core API client. There should be no more magic string URLs inside the builder (one or two magic string URLs do still exist because they use an old `fetchData` helper that Kev wrote and I have not removed).

## Error handling
The builder was terribly inconsistent in how it handled API responses, and particularly API errors. Sometimes errors were caught and displayed, sometimes errors were thrown but a success message was shown, and sometimes errors just weren't handled altogehter.

Errors will always happen, and so they always need to be handled. The new core API client has utilities built in to make error handling easier. If a request fails for any reason, or an error HTTP status code (>= 400) is received, it will throw a JS error which it has constructed. This means every single usage of any API endpoint now needs wrapped in a try/catch (this should have been the case anyway, as unexpected errors can always happen).

I've refactored the entire builder codebase to properly handle these API errors. One question which needed answered was at what stage should we handle errors. I think the correct answer here is that errors should not be caught until they have surfaced to the top level. This is the best way for the calling code to be in control of the flow of execution.

If we caught API errors in svelte stores (where we often use the API) it would mean that the calling code (in a svelte component for example) would still need to somehow know whether the result was successful or not, so it can know whether or not to proceed with further actions. This means we would end up either proxying on some other form of error, or even worse, returning success booleans or some other manual construct with no consistency. Store functions also often call through to each other, and if every store function was swallowing API errors then it would becomed very difficult to keep track of a success/failure path.

The best solution is to allow errors to propagate right up to the top level of code (typically in svelte components themselves). This is the only way that happy paths properly stop executing (e.g. an error was thrown from the API, so you don't want to do a store update, or the store threw an error so we don't want to close the modal). I've updated the whole codebase to use this approach and I think we should try our best to maintain this strategy going forward.

## Future work
We could think about potentially logging API errors to sentry if we wanted, as the new core API client can be configured with an `onError` hook. This hook is used to log errors to the console, but it could also be a central place to do other things with these errors like log then to sentry.

## Builder jest tests
I've disabled the `test` command in the builder package. The jest configuration there is outdated and absolutely awful, and it will not work after the refactor as it has never properly handled having to import another package (such as `frontend-core` now). I've spoken to Marty and we will rethink builder jest tests at some stage, but it needs created from scratch.


Rory-Powell (approved):
Loads of good work. Great shout on the error handling stuff

shogunpurple (approved):
LGTM! The new structure is super clean. Good job 🚀

PR #4228, opened 2022-01-27
Fix not showing pagination of query on production

pueding:
## Description
#3962 adds a feature of pagination for REST query but there is no pagination component showing on production.
It's because the server doesn't want to send 'fields' attribute in production when finding the definition of the query.
So I've attached the attribute on the front side and it can handle the definition?.fields?.pagination value.

## Screenshots
image

### DIFF
--- a/packages/client/src/utils/fetch/QueryFetch.js
+++ b/packages/client/src/utils/fetch/QueryFetch.js
@@ -16,7 +16,10 @@ export default class QueryFetch extends DataFetch {
     if (!datasource?._id) {
       return null
     }
-    return await fetchQueryDefinition(datasource._id)
+    const definition = await fetchQueryDefinition(datasource._id)
+    // After getting the definition of query, it loses "fields" attribute because of security reason from the server. However, this attribute needs to be inside of defintion for pagination.
+    definition.fields = datasource.fields
+    return definition
   }
 
   async getData() {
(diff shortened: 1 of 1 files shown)


pueding:

packages/client/src/utils/fetch/QueryFetch.js
```diff
@@ -16,7 +16,10 @@ export default class QueryFetch extends DataFetch {
     if (!datasource?._id) {
       return null
     }
-    return await fetchQueryDefinition(datasource._id)
+    const definition = await fetchQueryDefinition(datasource._id)
+    // After getting the definition of query, it loses "fields" attribute because of security reason from the server. However, this attribute needs to be inside of defintion for pagination.
+    definition.fields = datasource.fields
```
aptkingston, line 21: Could we first check if `fields` exist? In dev apps the fields will be present inside `definition`, and will be up to date. The ones stored here in `datasource` are potentially out of date, as the value of the datasource is saved at the moment you select it in the builder settings.

So we should just change this to:
```js
if (!definition.fields) {
  definition.fields = datasource.fields
}
```

aptkingston (changes requested):
Thanks @pueding! I'm happy to accept this PR if you could make the change I requested.

We need to solve this issue properly (by returning the pagination configuration from the server), but for now we may as well merge this in to fix the issue!

Well spotted, and thanks for the contribution.

packages/client/src/utils/fetch/QueryFetch.js
pueding: I've changed it. Thanks for your feedback. It helps me to understand the logic.

aptkingston (approved):
Thanks @pueding!

PR #4237, opened 2022-01-28
Fix app update endpoint

Rory-Powell:
## Description
A bug introduced in https://github.com/Budibase/budibase/pull/4137 prevented anything but the name and url from being updated in the app update endpoint.


aptkingston (approved):
LGTM! Thanks for getting this fixed quickly 👍

PR #4261, opened 2022-01-31
Fix CSRF token not present using local auth

Rory-Powell:
## Description
- Update fetch current user to always use the `self` call
- No longer set user information as a result of the `login` call
- This results in consistent experience for SSO and password login
- Prevents small differences between the responses
   - e.g. current login response https://github.com/Budibase/budibase/blob/develop/packages/worker/src/api/controllers/global/auth.js#L75-L81
   - e.g. current self response https://github.com/Budibase/budibase/blob/develop/packages/worker/src/api/controllers/global/users.js#L172-L175 

## Screenshots

### DIFF
--- a/packages/builder/src/pages/builder/_layout.svelte
+++ b/packages/builder/src/pages/builder/_layout.svelte
@@ -61,7 +61,7 @@
       await auth.setInitInfo({ init_template: $params["?template"] })
     }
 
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM!

Issue #4198, opened 2022-01-25
Budibase Backend API: Rate Limiting


shogunpurple:
**COS**
  - Limit budibase rest API requests to 10 requests per second per user
  - Put a wider rate limit on the entire budibase system to prevent spam

Issue #4135, opened 2022-01-20
Index field in DynamoDB Query should not be mandatory


farzadpanahi:
**Describe the bug**
When creating a query for DynamoDB connector, `index` field under `Fields` cannot be left empty. 
```
Query Error: Cannot convert undefined or null to object.
```

**To Reproduce**
1. Add a DynamoDB source
2. Click on Add Query
3. Under Fields section populate `table` but leave `index` empty
4. Write a simple query in `Query` section
5. Click on Run Query

**Expected behavior**
`index` field should not be mandatory to fill. We can have a table without any extra index on it. We should be able to leave this field empty and query DynamoDB without any error.

**Screenshots**
![image](https://user-images.githubusercontent.com/3793563/150417233-4057af21-ea98-4bcf-8520-4f67211486ac.png)

**Desktop (please complete the following information):**
 - OS: `Linux pop-os 5.15.11-76051511-generic #202112220937~1640185481~21.10~b3a2c21 SMP Wed Dec 22 15:41:49 U x86_64 x86_64 x86_64 GNU/Linux`
 - Browser: Chromium
 - Version: 97.0.4692.71 (Official Build) (64-bit)

Issue #2274, opened 2021-08-04
Replace Envoy with NGINX


shogunpurple:
Right now we ship an envoy proxy that sits in front of the budibase architecture.

It works well, however NGINX is a much more well known and mature solution, providing a full web server and more options for serving files etc. Providing a configurable NGINX config out of the box gives users total flexibility, as well as the ability to easily set up things like certbot for HTTPS.

**COS**
- Replace envoy.yaml with an nginx.conf and port it over
- Replace proxy service docker image with swag or standard nginx docker container
- Ensure docs are updated to reflect no longer relevant envoy links

aptkingston:
Swag would be awesome to allow out of the box HTTPS by just configuring env variables - something we could easily configure the CLI to handle.

mike12345567:
I think pre-configured SWAG is one of the top reasons to do it, since it'd provide so much value to self hosters.

PR #4268, opened 2022-01-31
Feature/envoy nginx

shogunpurple:
Addresses: 
- https://github.com/Budibase/budibase/issues/2274 
- part of https://github.com/Budibase/budibase/issues/4198
- https://github.com/Budibase/budibase/issues/4135

## Description
This PR replaces all the references to envoy in the budibase platform with NGINX. The dev, docker-compose prod and kubernetes proxy configurations have all been updated to use NGINX. Some other things have been changed to accommodate this:
- Replaced `hosting.properties` with `.env` to make the configuration method more consistent across the board. The CLI uses `.env` and the manual docker-compose setup uses `hosting.properties` this also makes the `docker-compose` command easier to run, and will support older docker-compose versions that make use of the `--env-file` flag. This won't break old installations given they can just keep using `--env-file`, but should solve a lot of issues for new users
- Changed `MINIO_URL` to the actual URL of the MinIO instance in dev/prod. This is due to the way NGINX handles the `/` route. NGINX doesn't make a difference between an empty root URL and one with a trailing slash. MinIO also does not work when you do path rewrites, so even adding a specific route like `/minio` to the reverse proxy config does not work with MinIO. File paths (such as for attachments, will still work the same way, and the default fallback URL is to route to MinIO.
- The budibase production proxy now runs as part of the build process - so it will ship along with the app/worker service images, and users will not have to pull the latest `nginx.conf` each time.

### Rate Limiting
I've added a 10 request per second rate limit also reflected in the account portal (https://github.com/Budibase/account-portal/pull/76) with bursting and an extra limit of 10 requests. This should be a reasonable rate limit that will avoid spamming. We can always adjust as time goes on.

### Security Updates
I've added some security based headers that will make budibase overall more secure:
```
    # Protects from clickjacking and abusing iframes from different domains
    X-Frame-Options SAMEORIGIN always;
    # Prevent MIME type sniffing
    X-Content-Type-Options nosniff always;
    # Legacy header for XSS - just for older browsers
    X-XSS-Protection
    # Ensure that we know all the domains and sources that we retreive assets from. Further layer of security for XSS and other attacks
    Content-Security-Policy
```

## Screenshots


Rory-Powell:
Question on the rate limiting, do we have any visibility into when rate limiting is in effect? 
Similar to the account portal question, 10 requests per seconds seems really low

shogunpurple:
@Rory-Powell since it's per client, I don't believe so unless we look at the nginx logs - in which case it will show that someone has been rate limited.

hosting/digitalocean/files/var/lib/cloud/scripts/per-instance/001_onboot
```diff
@@ -3,7 +3,7 @@
 # go into the app dir
 cd /root
 
-# fetch envoy and docker-compose files
+# fetch nginx and docker-compose files
 wget https://raw.githubusercontent.com/Budibase/budibase/master/hosting/docker-compose.yaml
 wget https://raw.githubusercontent.com/Budibase/budibase/master/hosting/envoy.yaml
```
Rory-Powell, line 8: Should we still fetch `envoy.yaml`?

packages/server/scripts/dev/manage.js
```diff
@@ -24,8 +24,8 @@ const Commands = {
 async function init() {
   // generate envoy file, always do this incase it has changed
   const hostingPath = path.join(process.cwd(), "..", "..", "hosting")
-  const envoyHbsPath = path.join(hostingPath, "envoy.dev.yaml.hbs")
-  const envoyOutputPath = path.join(hostingPath, ".generated-envoy.dev.yaml")
+  const envoyHbsPath = path.join(hostingPath, "nginx.dev.conf.hbs")
+  const envoyOutputPath = path.join(hostingPath, ".generated-nginx.dev.conf")
```
Rory-Powell, line 28: Rename these variables?

Rory-Powell (approved):

datdamnzotz:
Nice!  One suggestion, include some best practices
```
http {
	charset utf-8;
	sendfile on;
	tcp_nopush on;
	tcp_nodelay on;
	server_tokens off;
	types_hash_max_size 2048;
	client_max_body_size 16M; 
        ##buffer policy
        client_body_buffer_size 1K;
        client_header_buffer_size 1k;
        client_max_body_size 1k;
        large_client_header_buffers 2 1k;
        ##end buffer policy
...
}
```
Not sure if budibase is using trace, or options but might want to add
```
if ($request_method !~ ^(TRACE|OPTIONS)$ ) 
{
return 405; 
}
```

hosting/digitalocean/files/var/lib/cloud/scripts/per-instance/001_onboot
shogunpurple: Nope - updated all the leftover references to `envoy` now - thanks!

shogunpurple:
Thanks @datdamnzotz - added some of those options to our configuration. Since the entire configuration is held there in that file - it's worth covering ourselves. Thanks!

aptkingston (approved):
This is super work Marty! Very glad to shift over to nginx. Checked this out myself and had a play and it seems great.

Only pain point I had in upgrading is that I had to bump my docker-compose version to get it to run - just because of the classic issue with resolving the internal docker IP on linux (the `extra_host` of `host.docker.internal:host-gateway` for the nginx container). I upgraded from docker-compose 1.28.6 to 2.2.3 (seems like a huge jump, but 1.28.6 was from 2021) and that solved the issue. This is only related to dev envionments anyway.

Only 2 questions I have:
- Have you checked that the `X-Frame-Options SAMEORIGIN` header works alright with the client app iframe preview in prod?
- Should we consider increasing the burst size on the rate limiting to more than 10? On an initial load of the data UI in the builder I count 28 requests which will all go through nginx. That's over about 2.8 seconds, but it just sounds like cutting it very close to the limit (there are loads of duplicates here which is really cleanup we should do in the builder at some point, but that's another issue).
image

shogunpurple:
Thanks @aptkingston - in the case of the rate limiting - there's a base of 10, with a burst of 10 too and only to `/api` endpoints, so a lot of those calls there won't be affected.

I'm going to test the frames in pre-prod just to make sure everything works as expected ahead of security review tomorrow too.

aptkingston:
> Thanks @aptkingston - in the case of the rate limiting - there's a base of 10, with a burst of 10 too and only to `/api` endpoints, so a lot of those calls there won't be affected.
> 
> I'm going to test the frames in pre-prod just to make sure everything works as expected ahead of security review tomorrow too.

Didn't realise it was only to the `/api` endpoints - that's perfect then, shouldn't be an issue.

Issue #4195, opened 2022-01-25
Refactor PouchDB usage - use a CLS function context instead


mike12345567:
## Description
Been thinking lately about our usage of `new CouchDB(...)` and whether or not it could be improved.
The TLDR of this issue is to shift to a mechanism like we use for tenancy, with the `cls-hooked` function context for the app ID and the app database. The idea of this is to use a `getDevAppDB()` or `getProdAppDB()` call, as in the server 99% of the database access is to either the prod app, dev app or global databases. There should never be cross app communicate either, so this should be restrictive in a good way.

There are two key benefits to doing this:
1. Code quality - right now there is a bit of an anti-pattern that exists around some of the very expansive APIs, where we'll sometimes pass the db instance around and sometimes we'll pass the appId around - which can be painful if you are trying to integrate parts of the API and they don't do it the same way, you have to go refactor one to follow a single pattern so that they can talk. If we moved to a method where a single contextual method is used to retrieve the DB everyone would be able to speak the same language easily.
2. Performance - it turns out there is a negative to our current approach, everytime a new instance is created of Pouch it carries out an `info()` call to see if the database exists or not. For very heavy endpoints there could be a large number of excess HTTP calls, which serve no value.

Specific examples of APIs that would be affected by this (performance wise) are things like say the link controller, view functions, or bulk operations called by table ops, where it'll pass the app ID down to the functions to construct a new DB accessor.

All in all I think this could be worthwhile, we could expose three things that we use commonly within the context:
1. The app ID, so that it doesn't need to be passed around everywhere, simplifying function signatures.
2. The prod app DB
3. The dev app DB

I think the way the DBs should work is that the context functions should check if it exists in the context when the `getProdAppDB` or `getDevAppDB` functions are called, if they don't create them and add to context, that way per API call there will be max 2 DB instances created (a prod and a dev one - e.g. for replication).

## Investigation
I did some checks about the performance implications of this to check if its true or not (the performance issue) - when I wiresharked the connection between the database and the server I noticed numerous calls like:
```json
HTTP/1.1 200 OK
Cache-Control: must-revalidate
Connection: close
Content-Length: 481
Content-Type: application/json
Date: Tue, 25 Jan 2022 15:26:23 GMT
Server: CouchDB/3.1.1 (Erlang OTP/20)
X-Couch-Request-ID: aec96cf81c
X-CouchDB-Body-Time: 0
[...]

PR #4276, opened 2022-01-31
Refactoring app DB usage - one DB load per request

mike12345567:
## Description
The main reason for this feature was detailed in issue #4195 - which really discusses the reason for all of this.

The main focus of this work was to perform as many updates as possible to the server, there are a few places that haven't been updated, namely around the middlewares and then some of the application endpoints (where it might need to access multiple different app DBs, e.g. getting a list of all apps) but the idea was to reduce the direct usage of `new CouchDB` down to only edge cases.

The only middleware that still needs updated to follow this pattern is the `currentapp` middleware, as it is before any sort of CLS context has been configured, I wanted to avoid changing too much and I didn't like the idea of stacking context, rather than adjusting how context was loaded I left the middleware to access the app DB as it previously did - however this can be updated at some point in the future by refactoring the general context setup so that the context gets set from the current app middleware, rather than from a later one.


packages/backend-core/src/db/conversions.js
```diff
@@ -0,0 +1,46 @@
...
+const { APP_DEV_PREFIX, APP_PREFIX } = require("./constants")
+
+exports.isDevAppID = appId => {
+  if (!appId) {
+    throw NO_APP_ERROR
+  }
+  return appId.startsWith(APP_DEV_PREFIX)
+}
+
+exports.isProdAppID = appId => {
```
Rory-Powell, line 11: While we're touching this stuff, could we either make this `isDeployedAppID` or update the `getDeployedAppID` to be `getProdAppID`. Any preference here? kind of annoying that we use deployed and production interchangeably

packages/backend-core/src/tenancy/context.js
```diff
@@ -1,6 +1,25 @@
...
+const ContextKeys = {
+  TENANT_ID: "tenantId",
+  APP_ID: "appId",
+  // whatever the request app DB was
+  CURRENT_DB: "currentDb",
+  // get the prod app DB from the request
+  PROD_DB: "prodDb",
+  // get the dev app DB from the request
+  DEV_DB: "devDb",
+  DB_OPTS: "dbOpts",
```
Rory-Powell, line 21: Doesn't feel quite right for this app stuff to live in `tenancy/context.js`. Because this stuff has the potential to grow over time (user context, licensing context etc), it might make sense to refactor this a bit now to set in place a more modular structure e.g. 
```
src/context
- index.js // the context keys, shared functions
- app.js
- tenancy.js
```

packages/server/src/api/controllers/application.js
```diff
@@ -45,11 +44,13 @@ const { getTenantId, isMultiTenant } = require("@budibase/backend-core/tenancy")
...
 const { app: appCache } = require("@budibase/backend-core/cache")
 const { cleanupAutomations } = require("../../automations/utils")
+const context = require("@budibase/backend-core/context")
 
 const URL_REGEX_SLASH = /\/|\\/g
 
 // utility function, need to do away with this
-async function getLayouts(db) {
+async function getLayouts() {
+  const db = context.getAppDB()
```
Rory-Powell, line 53: We should probably be consistent with the usage of tenancy context and the other `getAppDB` usages in this pr by importing / using the function directly rather than going through the `context` object

Rory-Powell (approved):
Love this. Great work 👍

packages/backend-core/src/db/conversions.js
mike12345567: yeah you're right, may as well update it now, think we should move to `prod` term, as deployment/deployed isn't used much really.

packages/server/src/api/controllers/application.js
mike12345567: in this case I had wanted it to be clear in the application controller when it was getting the DB from context and when it was getting a new/raw one (such as creating an app DB) - for ever other controller this doesn't really matter since they aren't in an app context, so this use case here was specifically to differentiate it from everywhere else it was used.

packages/backend-core/src/tenancy/context.js
mike12345567: Thought about this, might just change the directory from `tenancy` to `context` as tenancy should live under the context directory anyway.

packages/builder/cypress/setup.js
```diff
@@ -29,22 +26,20 @@ process.env.ALLOW_DEV_AUTOMATIONS = 1
 // Stop info logs polluting test outputs
 process.env.LOG_LEVEL = "error"
 
-async function run() {
+exports.run = (
+  serverLoc = "../../server/dist",
```
Rory-Powell, line 30: Just checking this doesn't actually fix the issue we had with needing to build, or does it?

mike12345567: It does, there's a `ts/setup.ts` as you can't just run `ts-node`, you actually have to run it out of a typescript file for it to recognise it correctly, when ts-node is called it re-directs to the actual typescript server file instead.

Rory-Powell: ah nice I missed that one good shout. Now that the worker TS has been merged can we update to include it here as well?

Issue #4256, opened 2022-01-30
Formula column as table display column is not shown when displaying relationship column 


lhtrieu87:
**To Reproduce**
1. Create the table `users` `(id, first_name, last_name, full_name)` with `full_name` being formula = `{{first_name}} {{last_name}}`
2. Set `full_name` as table display column
3. Populate some rows into the `users` table
4. Create the `billings` table with a column relationship `creator_id` to the `users` table
5. Create a row in the `billings` table with a valid `creator_id` value
6. The `creator_id` column of the `billings` table is empty

**Expected behavior**
The `creator_id` column should display the corresponding `full_name` column of the `users` table.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Chrome
 - Version 97.0.4692.99 (Official Build) (64-bit)

**Additional context**
- Use Budibase Cloud
- Remote PostgreSQL 14 on Google Cloud
- Set `first_name` as table display column, `first_name` is displayed in the billings table, but not with `full_name` as table display column

Issue #4255, opened 2022-01-30
Rows are not displayed when there are multiple relationships between 2 PostgreSQL tables


lhtrieu87:
**Describe the bug**
Rows are not displayed when there are multiple relationships between 2 PostgreSQL tables.

**To Reproduce**
1. Use Budibase Cloud
2. Use PostgreSQL as the data source for all below steps
3. Create a users table (id, first_name, last_name)
4. Populate some data into the users table
5. Create a billings table (id, creator_id, billed_user_id) with both creator_id and billed_user_id being the foreign key relationships to the users table
6. Create a new row for the billings table
7. The newly created billings row is **NOT** displayed in both Budibase data builder and deployed view

**Expected behavior**
The newly created billings row should be displayed in both Budibase data builder and deployed view.  

**Screenshots**
Rows in PostgreSQL:
![Screenshot from 2022-01-30 10-58-41](https://user-images.githubusercontent.com/2397432/151685002-1a75ef8b-d5d8-48f4-8ee8-ea232924b0a9.png)

No rows in Budibase data builder:
![Screenshot from 2022-01-30 10-59-38](https://user-images.githubusercontent.com/2397432/151685030-69230dff-e4ff-4b41-8286-d98fbc92f715.png)

No rows in the deployed version:
![Screenshot from 2022-01-30 11-00-01](https://user-images.githubusercontent.com/2397432/151685035-1d8c2520-4378-47ce-bff2-1beb81230497.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Chrome
 - Version 97.0.4692.99 (Official Build) (64-bit)

**Additional context**
- Use Budibase Cloud
- Use PostgreSQL version 14 on Google Cloud

maxschrems:
Maybe related: 

* Cannot define a second relationship between two tables in postgresql - "save" button is not clickable.

![Unbenannt-2](https://user-images.githubusercontent.com/7262828/152132549-01865daf-b328-40a6-b9b2-ff2f42d885db.png)

PR #4297, opened 2022-02-02
Some small SQL core fixes

mike12345567:
## Description
Two issues were raised recently, features that weren't supported by Budibase SQL core but should have been:

1. Multiple relationships between the two same tables - this would break Budibase previously, supporting it by changing join syntax - #4255 
2. A quick fix for using formula columns added to SQL tables as the primary display field, making sure it is enriched onto relationships before they are turned into links - #4256.


shogunpurple (approved):
Awesome - some really core fixes here that affect the critical path when using SQL. Good job

PR #4309, opened 2022-02-03
Rich text field + markdown viewer component

aptkingston:
# Update
The following changes have been made after discussions with the team:
- A flag has been added to the current long form data type to specify whether it is rich text or not:
image
image
- A setting has been added to the long form field component, to allow manually specifying whether you want a plain text or rich text editor. This defaults to "Auto", which will automatically render the correct editor depending on the schema:
image
- Autoscreens and autogenerated form fields will default to auto, and render the correct editor:
image
- The heights of both editors have been standardised to 150px by default and this can be controlled by the height setting on the component
- Rich text editor works properly in fullscreen and side-by-side mode in any resolution and in any layout navigation type (top, left, none)

# Original PR
## Description
This PR replaces the current "long form text" field component with a rich text editor. The type of the field has remained the same, and it is still called "Long form text", but it will now render a rich text markdown editor instead of a text area. This is fully backwards compatible and will not cause any issues with existing fields.

## Rich text editor
The new rich text editor is displayed in the data UI when editing rows and displayed in client apps when adding a long form text field (or with auto generated fields). The editor uses [`easy-markdown-editor`](https://github.com/Ionaru/easy-markdown-editor). This is a great hybrid between a raw markdown editor and a WYSIWYG editor - you're editing and storing markdown, but it styles the markdown in real time.

Here's an example:
image

There is a preview button built in to the toolbar (the eye icon) which will show you how the markdown actually looks:
image

The editor supports width and height styling in the builder, so can be sized as desired. Pretty much all the standard markdown syntax is supported, as well as some extended syntax.

## Full screen editing
The new rich text editor supports a new full screen mode.This opens the editor full screen for easy editing. 
Full screen editor enabled by clicking the full screen icon:
image

Even better than this is the ability to preview the markdown side-by-side with the editor when full screen.
This is enabled by clicking the side-by-side icon:
image

## Feature list
 A quick list of features are:
- Bold
image
- Italics
image
- Headers (6 sizes)
image
- Block quotes
image
- Unordered lists
image
- Ordered lists (with auto numbering)
image
- Links
image
- Images
image
- Horizontal rules
image
- Tables
image
- Code blocks
image
- Inline code
image
- Nesting most of these things together
image

## Theming
The rich text editor is fully supported by spectrum themes. The primary colour is also used for the color of links.

Lightest:
image
Light:
image
Dark:
image
Darkest:
image

## Markdown viewer component
Since we're storing markdown, it's important to be able to use it. It's a pain to have to build a form just to display your markdown, so I've added a new markdown viewer component, which simply takes in a markdown string and displays it. This isn't coupled to the new rich text field - you could pull in external markdown and display it with this if you wanted to.

The markdown component takes a single, bindable text setting which is just the markdown to display. By default, the viewer will take up only as much height as is needed to fully display the markdown without scrolling. If you want to set a fixed height then this is also supported, but it wil scroll the content. It is styled differently from the editor so that the content looks natural on a page. *Starts planning editable wiki app template*.

The viewer component:
image


shogunpurple (approved):
I love it! This is awesome. Especially the full screen and side by side views. Great job

joebudi:
This is awesome @aptkingston . I really like the addition of the markdown viewer component. My first thought was, I can't wait to create internal docs with this! Also, I love the simplicity of it, the full-screen mode, and of course how it's themeable. 

The preview button is critical for end-users. I do think there is value in having a plain text long-form alt. For example, one of our users has 700,000,000 end-users. Do we expect the end-users and general public to complete long-text inputs in markdown? I think with some use cases a simple long-form plain text component suffices. Therefore an option to swap the editor with a plain text alternative might be nice? I hope I am making sense here.

Also, how does this look within a form on mobile?

aptkingston:
> This is awesome @aptkingston . I really like the addition of the markdown viewer component. My first thought was, I can't wait to create internal docs with this! Also, I love the simplicity of it, the full-screen mode, and of course how it's themeable.
> 
> The preview button is critical for end-users. I do think there is value in having a plain text long-form alt. For example, one of our users has 700,000,000 end-users. Do we expect the end-users and general public to complete long-text inputs in markdown? I think with some use cases a simple long-form plain text component suffices. Therefore an option to swap the editor with a plain text alternative might be nice? I hope I am making sense here.
> 
> Also, how does this look within a form on mobile?

About the plain text - I think this is where we need to make a distinction between plain text and rich text. I think ideally what we do is add a new schema option to text for whether it is multiline or not - and that determines whether you get a text area or an input. Rich text is always rich text, so would stay as is. What are your thoughts on that? In terms of naming then we could have either "Text" or "Pain text" for the normal one, and "Rich text" or even "Markdown" for the rich text (instead of the current "Long form text").

The toolbar wraps on mobile:
image

Fullscreen mode on mobile has a slight issue (not my fault, it's the underlying editor) that means if icons wrap on mobile fullscreen then they look weird:
image

I can't really fix that... But it's a minor issue and doesn't stop anything working, it just looks weird. I doubt anyone will ever see this anyway. The options are really
- Don't allow fullscreen on mobile
- Remove some icons from the toolbar in general so that it is smaller and never wraps
- Do nothing since it's not much of a problem

mjashanks:
Also - not a blocker, but can we view the text in read-only mode, without editing? I.e. Formatted text, with no toolbar - just readable content.

aptkingston:
Yep this is what the markdown viewer component is for (the last part of the post). It just takes a markdown string and displays it, formatted. No forms required even.

joebudi:
> > This is awesome @aptkingston . I really like the addition of the markdown viewer component. My first thought was, I can't wait to create internal docs with this! Also, I love the simplicity of it, the full-screen mode, and of course how it's themeable.
> > The preview button is critical for end-users. I do think there is value in having a plain text long-form alt. For example, one of our users has 700,000,000 end-users. Do we expect the end-users and general public to complete long-text inputs in markdown? I think with some use cases a simple long-form plain text component suffices. Therefore an option to swap the editor with a plain text alternative might be nice? I hope I am making sense here.
> > Also, how does this look within a form on mobile?
> 
> About the plain text - I think this is where we need to make a distinction between plain text and rich text. I think ideally what we do is add a new schema option to text for whether it is multiline or not - and that determines whether you get a text area or an input. Rich text is always rich text, so would stay as is. What are your thoughts on that? In terms of naming then we could have either "Text" or "Pain text" for the normal one, and "Rich text" or even "Markdown" for the rich text (instead of the current "Long form text").
> 
> The toolbar wraps on mobile: image
> 
> Fullscreen mode on mobile has a slight issue (not my fault, it's the underlying editor) that means if icons wrap on mobile fullscreen then they look weird: image
> 
> I can't really fix that... But it's a minor issue and doesn't stop anything working, it just looks weird. I doubt anyone will ever see this anyway. The options are really
> 
> * Don't allow fullscreen on mobile
> * Remove some icons from the toolbar in general so that it is smaller and never wraps
> * Do nothing since it's not much of a problem

I like the multi-line option.

I like the naming conventions:
- text area
- rich text editor

I feel we should remove the side by side and help icon? That should sort it?

aptkingston:
I meant naming conventions for the actual data types - the components will probably just be named "{data type} field".

Can definitely remove the help icon, but the side by side is really powerful and I wouldn't want to remove it. I'd probably say remove either the quote or the numbered list (since the numbered list is exactly how you would type it anyway).

joebudi:
Ah, sorry, I meant to remove the icons from the mobile view just?

aptkingston:
> Ah, sorry, I meant to remove the icons from the mobile view just?

👍 I've hidden the side-by-side and help icons for mobile. I think the library itself actually hides fullscreen mode entirely for some viewport dimensions - it seems to check exactly for iphone dimensions for some reason... But this is still a good addition - it'll just stop the overflowing icons.

Rory-Powell (approved):
Looks great 👍

shogunpurple (approved):
Great job on those changes @aptkingston 🚀

Issue #4308, opened 2022-02-03
Rest API cannot use more than one binding in URL 


mjashanks:
When adding a REST Api query....

`https://api.github.com/repos/budibase/{{ name }}` WORKS
`{{ baseUrl }}/repos/budibase/budibase` WORKS
`{{ baseUrl }}/repos/budibase/{{ name }}` FAILS with `400: only absolute URLs are supported`

PR #4313, opened 2022-02-03
Fixing query multiple handlebars statements enrichment

mike12345567:
## Description
Fix for #4308 - triple brace conversion was not working correctly, wrote this into the string templates instead - also fixing an issue with the RBAC for Rest.

### DIFF
--- a/packages/builder/src/components/integration/AccessLevelSelect.svelte
+++ b/packages/builder/src/components/integration/AccessLevelSelect.svelte
@@ -26,6 +26,7 @@
 
   async function getPermissions(queryToFetch) {
     if (fetched?._id === queryToFetch?._id) {
+      loaded = true
       return
     }
     fetched = queryToFetch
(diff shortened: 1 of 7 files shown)


aptkingston (approved):
LGTM! Can I just check what this is actually doing - the naming is confusing me a little. If you pass in `noEscaping: true` then it will actually escape all HBS expressions and treat them as string literals instead?

Or what happens when a triple brace is enountered in the rest of the processing.

mike12345567:
> LGTM! Can I just check what this is actually doing - the naming is confusing me a little. If you pass in `noEscaping: true` then it will actually escape all HBS expressions and treat them as string literals instead?
> 
> Or what happens when a triple brace is encountered in the rest of the processing.

Yeah have to admit the functionality is a little weird here, basically what this does is it converts the double braces to triple braces when `noEscaping: true` is specified, the issue is that with the double braces Handlebars under the hood will escape certain characters which breaks queries - the triple brace is a string literal as you say and fixes this.

As you say its a bit weird that you're essentially escaping the double braces to triple braces, so that it disables the internal handlebars escaping, although when you think about it the user of the string templates API isn't worried about what the library actually does under the hood to disable handlebars escaping.

Issue #4320, opened 2022-02-04
Default value for Relationship Picker


emrahcetiner:
All forms components have a default value feature but the relationship picker doesn't. I want a form for creating a new row but some fields should fill with last time choices. Before saving the form I write its value to app state then read from app state and set it as default value.

How can I do this for the Relationship Picker without the default value property?

aptkingston:
Hey @emrahcetiner. I've just merged a PR to add this! It will be included in our next release which will be soon.

peterhuppertz:
I have a problem setting the default value of the relationship picker:
1) If I set the default value with a string it's selected fine: e.g. %5B'us_c81da6f7f3f04e70bcb46cbd69ea44d0'%5D
2) If I set the default value including a variable, it does not work. E.g. %5B'{{db.Rows.0.fk_users}}'%5D

I tested and the resulting string is EXACTLY the same, so I don't understand why the "normal" string selects the correct value, while the string including the result from the DB does not. Any idea?

BernardTeske:
Hello! The problem is that the data is loaded asynchronously. I have used a javascript in my default value.

```
const serverID = String($("serversDataProvider.Rows.0._id"));
console.log(serverID);
return serverID;
```

I also have this output via console.log. I get first "undefined" and then the correct value.

In my case it helped to use the ID from the request string. I pass the id=3 in the URL (as "server"). The default value for the field is now:

`%5B{{URL.server}}%5D`

I don't know if this always works, but for me this workaround worked.

PR #4323, opened 2022-02-04
Relationship field default value

aptkingston:
## Description
This PR adds a default value setting for relationship fields as raised in https://github.com/Budibase/budibase/issues/4320. It also fixes the clipboard functionality for copying IDs to the clipboard in the data UI, which previously depended on the clipboard API which depends on a secure context.

The relationship field default value works for both internal tables and SQL.
The relationship field default value needs to use Budibase row IDs rather than actual SQL IDs. This might trip people up. 

The value format can be any of:
- (HBS or JS) a single ID
- (HBS or JS) multiple IDs separated by commas (spaces optional)
- (JS) a literal array of IDs

New default value setting for a SQL datasource with 2 rows pre-selected:
image

### DIFF
--- a/packages/bbui/src/Table/InternalRenderer.svelte
+++ b/packages/bbui/src/Table/InternalRenderer.svelte
@@ -8,10 +8,35 @@
     copyToClipboard(value)
   }
 
-  function copyToClipboard(value) {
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM! Nice to have this.

PR #4324, opened 2022-02-04
Reset the builder store whenever unmounting the app root layout

aptkingston:
## Description
Resets the builder store whenever the root app layout is unmounted. This means whenever changing apps in the builder, we properly reset state and don't hang on to any extra stale values.

The problem with stale data was highlighted by the fact we were sending up the app ID of the previously opened app in the call to fetch the app package,


shogunpurple (approved):
LGTM! Nice one

Issue #3928, opened 2022-01-05
MS SQL Server connector not working correctly with MSSQL v2012


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3883

```
 ERROR [1640887454758] (30 on b3d416c9bd8e): RequestError: Invalid usage of the option next in the FETCH statement.
bbapps                |     Error: RequestError: Invalid usage of the option next in the FETCH statement.
bbapps                |         at /app/dist/integrations/microsoftSqlServer.js:85:23
...
```

I believe that Knex does not support some fairly recent versions of MS SQL Server: https://github.com/knex/knex/issues/1527. Some comments in there mention that error happens with v2017. 

We need to work out
1. Which versions of SQL Server do we currently support?
2. Which versions do we intend to support?

The maintainers of Knex have made it clear that they are happy to accept contributions to support earlier versions. If we want to support these versions we should look into contributing to Knex, since it's such a great project.

mike12345567:
Currently we test against the latest image from the Microsoft MCR (https://mcr.microsoft.com/v2/mssql/server/tags/list) - this should be MS SQL 2019. As of right now this is the only version we've tested extensively against - it would be easy enough for us to run tests against older versions and check what changes we'd need to make to support it.

Right now the big issue seems to be around pagination (`OFFSET`) in older versions, which we might be able to fix by detecting the version of MS SQL in use and recording the version against the datasource, allowing us to adjust our pagination efforts depending on version. The bigger issue here is we'd need to make a decision about what versions we want to test (and support) for MS-SQL and realistically other DBs as well.

datdamnzotz:
Confirmed - same error on Microsoft SQL Server 2016 (SP3) (KB5003279) - 13.0.6300.2 (X64)
```
ERROR [1643643870316] (29 on a6305281b83f): RequestError: Invalid usage of the option next in the FETCH statement.
    Error: RequestError: Invalid usage of the option next in the FETCH statement.
        at /app/dist/integrations/microsoftSqlServer.js:85:23
```
I tried changing SQL Db compatibility modes as well (from 2012 up to 2016)
![image](https://user-images.githubusercontent.com/3913859/151826239-7d7f57b5-a14c-4783-b246-6b0b928caa12.png)

Same error

pbjorklund:
Same with Azure Sql Server

pbjorklund:
Still getting this error with azure sql server

```
[...]

PR #4328, opened 2022-02-04
Fix for SQL Server - fix pagination issue

mike12345567:
## Description
This is a quick fix for the pagination issue in SQL Server, after testing this in a few different versions I found that we could simply force Knex to search on the primary key, which would allow pagination to work correctly for older SQL Server versions.

This resolves the issue that many people have seen such as `invalid use of "FETCH" ...`.

Issue #3928
and
Discussed #3883


shogunpurple (approved):
LGTM

Rory-Powell (approved):

Issue #4267, opened 2022-01-31
Admin role cannot see screen whose access is set to a custom role


mjashanks:
I have an app with
- one screen `/contacts`
- a custom role `New Role`
- `/contacts` access is set to `New Role`

When I run my app (Dev or Published), I cannot access the `/contacts` screen. I am a full admin on my tenant.

The same issue does not happen if I set the screen's role to `Basic`

PR #4332, opened 2022-02-04
Various fixes for RBAC and user administration

mike12345567:
## Description
Did some manual testing around the user administration section, creating users, deleting etc as part of fixing #4267 - found some issues:
1. Fixed the primary issue, making sure admins have access to all screen routes
2. Fixed an issue that was preventing user logout, was also causing a problem with user password reset
3. Fixed some issues around user sync to various DBs, this was an edge case that wasn't caught in my refactor of the app DB, as it accesses many different app databases as part of the sync.
4. Made sure that when a user accesses the system via a pretty app URL that the app ID is correctly set into the context.

### DIFF
--- a/packages/server/src/api/controllers/static/index.js
+++ b/packages/server/src/api/controllers/static/index.js
@@ -16,7 +16,7 @@ const { clientLibraryPath } = require("../../../utilities")
 const { upload } = require("../../../utilities/fileSystem")
 const { attachmentsRelativeURL } = require("../../../utilities")
 const { DocumentTypes } = require("../../../db/utils")
-const { getAppDB } = require("@budibase/backend-core/context")
+const { getAppDB, updateAppId } = require("@budibase/backend-core/context")
 const AWS = require("aws-sdk")
 const AWS_REGION = env.AWS_REGION ? env.AWS_REGION : "eu-west-1"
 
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM! Nice work

packages/server/src/api/controllers/static/index.js
```diff
@@ -49,11 +49,9 @@ async function getAppIdFromUrl(ctx) {
     a => a.url && a.url.toLowerCase() === possibleAppUrl
   )[0]
 
-  if (app && app.appId) {
-    return app.appId
-  } else {
-    return ctx.params.appId
-  }
+  const appId = app && app.appId ? app.appId : ctx.params.appId
+  updateAppId(appId)
```
Rory-Powell, line 53: I think we should do this from `serveApp` instead, that seems the more appropriate as this is more of a helper, it's also used in migrations so we can't guarantee it's the right thing to do to set context here

Rory-Powell (changes requested):

packages/server/src/api/controllers/static/index.js
Rory-Powell: apologies it's `getAppUrl` that is used in migrations, but the comment above I believe is still valid

mike12345567: Since this is a locally scoped and not exported function, I thought it would be best to move the functionality here, as I wouldn't want anyone to try and use the pretty app URL and not understand that the app ID needs to be updated in the context. I felt like that was a more likely scenario than the one of getting the pretty URL but not actually wanting that app ID to be in the context, since this is just to be used in the local scope of serving apps/clients etc.

Rory-Powell (approved):

Issue #3739, opened 2021-12-09
Disable timeout on toast for errors


diogenesbrussels:
**Describe the feature request**
When an import fails, the error message is displayed for only a few seconds. It is difficult to read or understand the details.
Maybe this type of notification should persist until the user clicks to hide it? And add a way to copy the content of an error message to the clipboard?
Thank you!

**Screenshots**
![image](https://user-images.githubusercontent.com/91942877/145391882-7e72bdb8-7aee-41e4-9b25-282f39a9d563.png)

PR #4372, opened 2022-02-07
Remove automatic dismissal of error notifications

aptkingston:
## Description
This PR prevents error notifications from being automatically dismissed, both in the builder and in client apps.

In the builder we can show multiple notifications as once, in a vertical list. Error notifications behave normally and move as normal, but have an X icon to manually dismiss them and they will not be automatically removed. Non-error notifications do not have this X icon, and can not be manually dismissed.

Client apps can only show one notification at once, but behave exactly the same as the new builder notifications.

The new error notifications look like this - the only change being the X icon indicating it needs to be manually dismissed:
image

Addresses #3739.

### DIFF
--- a/packages/bbui/src/Notification/Notification.svelte
+++ b/packages/bbui/src/Notification/Notification.svelte
@@ -1,7 +1,12 @@
 <script>
+  import { createEventDispatcher } from "svelte"
+
   export let type = "info"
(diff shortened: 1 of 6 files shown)


Rory-Powell (approved):
LGTM 👍

PR #4390, opened 2022-02-08
Adding ability to toggle between Automation input types. 

PClmnt:
## Description
This PR adds the ability for the user to have more control over how they handle automations and the binding within them. 

There are now two modes which can be changed via a dropdown that's present in each block that allows you to manipulate rows. 

image

Using that you can toggle between what we have now which is:

image

and the new control which is like this:

image

You can toggle each block to use bindings or normal values, and it will remember per block what option was selected. 

This also now means that we can use Javascript within automation bindings now. 

n.b also tidied up how we were selecting the current block and removed some uneeded code.


packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -32,23 +36,43 @@
     })
   }
 
+  $: currentAutomation = $automationStore.selectedAutomation.automation
+  $: console.log(currentAutomation)
```
aptkingston, line 40: Extra `console.log` left in.

aptkingston (commented):
Nice work Peter! This will make some features much easier when working with automations. Noted down a few questions which I think we should address.

In general, I think you're right when you say that the toggle to enable this should be built into the normal content area, like beside the bin icon as you mentioned.

I think the terminology of "row control" is a bit confusing too, so I think we should either update the naming (although I'm not sure what to), or add a tooltip somewhere to explain what it means. Maybe even both.

Issue #4431, opened 2022-02-11
Something not working on database joins (postgresql)


martino-zito:
**Describe the bug**
i have just upgraded from 1.0.46 version to 1.0.57 and the relationship between tables have stopped working. When a relationship between two tables exist if you browse the related entry and the foreign key is missing (null on database) the related row is not displayed

es.

**artist** table
id | name | surname | role_id
1   | john  | smith       |  null

**band** table
id | band name 
1  | Smiths&sons

in this case if you set an existing relationship on band table with artist table in the artist data section you will see no rows at all.

It seems to me that instead of a left outer join a inner join is used.

mike12345567:
Hi @martino-zito, thanks for reporting this, I can reproduce, will get a fix out asap.

mike12345567:
Hi @martino-zito 1.0.59 has been released (https://github.com/Budibase/budibase/releases/tag/v1.0.59) you should be able to update to the latest image now.

martino-zito:
@mike12345567  i think that something has gone wrong after the fix. Or that the revert is partial.

I've updated to 1.0.59 and as a side effect i'm now unable to create any query on postgres. The preview api is failing and complains for a validation error. i think that the client is sending parameters (name and schema) that are not allowed on the server side. removing them and curling the right payload the preview response comes back

`ERROR [1644644299532] (51 on 6d431d844748): Invalid body - "schema" is not allowed

    BadRequestError: Invalid body - "schema" is not allowed

        at Object.throw (/app/node_modules/koa/lib/context.js:97:11)

        at /app/dist/middleware/joi-validator.js:17:22

        at dispatch (/app/node_modules/@koa/router/node_modules/koa-compose/index.js:44:32)

        at next (/app/node_modules/@koa/router/node_modules/koa-compose/index.js:45:18)

        at /app/node_modules/@budibase/backend-core/src/middleware/csrf.js:55:14

        at /app/dist/middleware/authorized.js:90:16

        at Generator.next (<anonymous>)

        at fulfilled (/app/dist/middleware/authorized.js:5:58)

        at processTicksAndRejections (internal/process/task_queues.js:95:5)

Trace: BadRequestError: Invalid body - "schema" is not allowed

    at Object.throw (/app/node_modules/koa/lib/context.js:97:11)

    at /app/dist/middleware/joi-validator.js:17:22

    at dispatch (/app/node_modules/@koa/router/node_modules/koa-compose/index.js:44:32)

[...]

PR #4432, opened 2022-02-11
Fix for joins in SQL tables

mike12345567:
## Description
Fix for #4431 - converting SQL joins back to left joins, syntax specifying join type as string is old which doesn't work in latest versions of Knex.


shogunpurple (approved):
LGTM!

martino-zito:
Ty, i'll check it when available on dockerhub.

PR #4464, opened 2022-02-13
Remount button action components fully

aptkingston:
## Description
This PR fixes a UI issue where button action components can sometimes show stale data due to not fully remounting when changing step.

This is descibed in https://github.com/Budibase/budibase/discussions/4461, but the actual issue was only visual and the data being saved to the rows was always correct.


shogunpurple (approved):
LGTM!

Issue #4449, opened 2022-02-12
"No rows found" when column name has more than one space


gerhardmpl:
**Describe the bug**
When clicking on a three word column header (e.g. "process owned by") of a table in the data section to sort the rows, the sorted table is empty and instead "No rows found" is displayed.

**To Reproduce**
1. Go to data section of an app
2. Create a Budibase DB table (e.g. "processes")
3. Add a new column (name e.g. "process owned by") as text
4. Add some rows with text content
5. Click on column header to sort the rows and instead of sorting the rows "Now rows found" is displayed
6. Change the column name to "process owner" and sorting works

**Expected behavior**
The tables content is sorted when clicking on the column header.

**Screenshots**

**Desktop (please complete the following information):**
 - Windows 10 Desktop
 - Browser: Firefox
 - Version 97 (64bit)

**Additional context**
Same result when self hosted (app version 1.0.59) and cloud (app version 1.0.57)

aptkingston:
Hey @gerhardmpl. Thanks for the clear description and reproduction steps. I've found the issue and will get a fix pushed 👍

PR #4472, opened 2022-02-14
Replace all spaces with underscores in column names

aptkingston:
## Description
This PR fixes an issue reported in https://github.com/Budibase/budibase/issues/4449, where only the first space in column names was being replaced for the search index, due to us not using a global regex.


mike12345567 (approved):
LGTM!

PR #4475, opened 2022-02-14
Table column config

aptkingston:
## Description
This PR adds a new component for configuring table columns. This feature is available on both the table component and table block.

The new available features are:
- Add and remove columns
- Drag to reorder columns
- Set custom labels for columns
- Quickly reset to add all columns

Some clarification of functionality:
- Initially all columns are added, as before
- You cannot remove all columns, and the delete icon will be disabled when you are down to one column remaining
- The component does correctly handle if no columns are selected though, for example if the datasource has no columns available

This change is fully backwards compatible. The previous setting value was a flat array of column names. The new component can handle this format, and will transform it in to the new structure the first time it is saved.

## Screenshots
image

### DIFF
--- a/packages/bbui/src/Drawer/Drawer.svelte
+++ b/packages/bbui/src/Drawer/Drawer.svelte
@@ -74,4 +74,12 @@
     align-items: flex-start;
     gap: var(--spacing-xs);
   }
+
+  .buttons {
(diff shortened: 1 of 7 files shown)


mike12345567 (approved):
LGTM!

Issue #4337, opened 2022-02-04
Automation query step - cannot filter by _id


mjashanks:
Create a new automation "Query" block.

Now try to filter by `_id`. There is no `_id`. This is pretty essential.

Shyru:
Yeah it is, otherwise you cannot access data in related records within an automation.
See here for more background information:
https://github.com/Budibase/budibase/discussions/3673

PR #4478, opened 2022-02-14
Various fixes for release, HBS editor, _id filtering on automations, Safari issues

mike12345567:
## Description
Fixing issue #4337, allows internal tables to search by the _id or _rev properties in automation query block. The way this is implemented for the design side vs the automation side is a little different, there is a function in `builder/src/builderStore/dataBinding.js` called `getSchemaForDatasource` which is used in the design section to generate the schema used for the filter drawer, but this can't be easily used for the automation side as it depends on the component/asset structure that of the design, rather than re-working all of this for a bug fix went for the easiest route of just detecting an internal table in the automation schema building side and adding the `_id` and `_rev` fields. These can't be used for SQL tables as they don't actually exist in the table and therefore cannot be searched or updated.

Ideally we might want to take the functionality that is defined in `dataBinding.js` and produce some more generic helpers which could be moved into the frontend core for determining the schema of a structure, based on its datasource/how it is used.

Also fixed a visual issue where the buttons in the drawer slot were stacking, rather than being horizontally laid out.

Also includes a fix for buttons, after the addition of the tooltip option some buttons weren't respecting their old full width, instead defaulting to the size of their text, this caused problems around the auth pages.

Also fixes an issue with HBS editor, where curly braces weren't always added at the correct times.

Also includes fixes for Safari, removing the use of regex lookbehinds which are not supported.

## Screenshots
Screenshot showing it working.
image

### DIFF
--- a/packages/bbui/src/Button/Button.svelte
+++ b/packages/bbui/src/Button/Button.svelte
@@ -68,6 +68,10 @@
 </div>
 
 <style>
+  button {
+    width: 100%;
+  }
+
   .container {
     display: flex;
     align-items: center;
--- a/packages/bbui/src/Button/Button.svelte
+++ b/packages/bbui/src/Button/Button.svelte
@@ -17,65 +17,55 @@
   let showTooltip = false
 </script>
 
-<div class:container={!!tooltip}>
-  <button
-    class:spectrum-Button--cta={cta}
-    class:spectrum-Button--primary={primary}
-    class:spectrum-Button--secondary={secondary}
-    class:spectrum-Button--warning={warning}
-    class:spectrum-Button--overBackground={overBackground}
-    class:spectrum-Button--quiet={quiet}
-    class:active
-    class="spectrum-Button spectrum-Button--size{size.toUpperCase()}"
-    {disabled}
-    on:click|preventDefault
-    on:mouseover={() => (showTooltip = true)}
-    on:mouseleave={() => (showTooltip = false)}
-  >
-    {#if icon}
+<button
+  class:spectrum-Button--cta={cta}
+  class:spectrum-Button--primary={primary}
+  class:spectrum-Button--secondary={secondary}
+  class:spectrum-Button--warning={warning}
+  class:spectrum-Button--overBackground={overBackground}
+  class:spectrum-Button--quiet={quiet}
+  class:active
+  class="spectrum-Button spectrum-Button--size{size.toUpperCase()}"
+  {disabled}
+  on:click|preventDefault
+  on:mouseover={() => (showTooltip = true)}
+  on:mouseleave={() => (showTooltip = false)}
+>
+  {#if icon}
+    <svg
+      class="spectrum-Icon spectrum-Icon--size{size.toUpperCase()}"
+      focusable="false"
+      aria-hidden="true"
+      aria-label={icon}
+    >
+      <use xlink:href="#spectrum-icon-18-{icon}" />
+    </svg>
+  {/if}
+  {#if $$slots}
+    <span class="spectrum-Button-label"><slot /></span>
+  {/if}
+  {#if !disabled && tooltip}
+    <div class="tooltip-icon">
       <svg
         class="spectrum-Icon spectrum-Icon--size{size.toUpperCase()}"
         focusable="false"
         aria-hidden="true"
-        aria-label={icon}
+        aria-label="Info"
       >
-        <use xlink:href="#spectrum-icon-18-{icon}" />
+        <use xlink:href="#spectrum-icon-18-InfoOutline" />
       </svg>
-    {/if}
-    {#if $$slots}
-      <span class="spectrum-Button-label"><slot /></span>
-    {/if}
-    {#if !disabled && tooltip}
-      <div class="tooltip-icon">
-        <svg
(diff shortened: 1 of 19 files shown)


shogunpurple (approved):
LGTM! Nice fix

packages/bbui/src/Button/Button.svelte
```diff
@@ -68,6 +68,10 @@
 </div>
 
 <style>
+  button {
+    width: 100%;
+  }
```
aptkingston, line 73: Not sure about this - this could cause a few issues with client apps or elsewhere in the builder, unless this used to be here and was removed?

packages/bbui/src/Drawer/Drawer.svelte
```diff
@@ -44,6 +44,11 @@
 {/if}
 
 <style>
+  .buttons {
+    display: flex;
+    gap: var(--spacing-m);
+  }
```
aptkingston, line 50: Very glad to see this fixed 😄

packages/builder/src/components/design/PropertiesPanel/PropertyControls/FilterEditor/FilterDrawer.svelte
```diff
@@ -39,17 +45,30 @@
...
+    }
+    return base
+  }
+
+  function removeFilter(id) {
     filters = filters.filter(field => field.id !== id)
   }
 
-  const duplicateFilter = id => {
+  function duplicateFilter(id) {
```
aptkingston, line 65: We have an endless battle of updating functions in svelte components to be either arrow functions or normal functions lol.

packages/builder/src/components/common/bindings/BindingPanel.svelte
```diff
@@ -51,8 +51,9 @@
   })
   $: codeMirrorHints = bindings?.map(x => `$("${x.readableBinding}")`)
 
-  const updateValue = value => {
-    valid = isValid(readableToRuntimeBinding(bindings, value))
+  const updateValue = val => {
+    valid = isValid(readableToRuntimeBinding(bindings, val))
+    value = val
```
aptkingston, line 56: Just wondering was this fixing an issue? Normally whenever we update the value here with the `change` event, the new `value` prop will be passed in from the parent component again.

This is updating the value even it's not valid, which I don't think will cause any issues... but just wondering what prompted the change.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -174,7 +173,7 @@
             slot="body"
             bind:filters={tempFilters}
             {bindings}
-            {schemaFields}
+            {table}
```
aptkingston, line 176: Would it be cleaner here to just add the ID and rev fields to the `schemaFields` value that we already work out in this file, and then avoid having to make any changes to the filter drawer?

The function you put inside the filter drawer is exactly the function we would just need to put in here, and then we could keep the interface of filter drawer the same.

It's probably much of a muchness, but I'd lean towards keeping the filter drawer prop interface the same rather than having 2 possible props passed in to determine the schema.

The other option which might actually be even cleaner, would be actually just using `getSchemaForDatasource` inside here. You're right that it's sort of tailored towards the client, but the invocation to use it here would just be:
`$: schemaFields = getSchemaForDatasource(null, { type: "table", tableId })`

The only issue with using `getSchemaForDatasource` is that it adds `_id` and `_rev` for all tables, not just internal. Technically `_id` is in the schema of all tables, but `_rev` should only be internal, so that's a change we should probably make in that file. This use case is different though because this is like the "searchable" schema, which means we don't want to include `_id` for non-internal tables.

I think overall then my preference would be to pull the function from the filter drawer inside here, and just work out the full schema fields inside here, and make no changes to the filter drawer. And then we won't use `getSchemaForDatasource` at all, since we don't want the readable schema.

What are your thoughts? Do you prefer adding it to the filter drawer?

packages/builder/src/components/common/bindings/utils.js
```diff
@@ -1,7 +1,10 @@
 export function addHBSBinding(value, caretPos, binding) {
   binding = typeof binding === "string" ? binding : binding.path
   value = value == null ? "" : value
-  if (!value.includes("{{") && !value.includes("}}")) {
+
+  const left = caretPos?.start ? value.substring(0, caretPos.start) : ""
+  const right = caretPos?.end ? value.substring(caretPos.end) : ""
+  if (!left.includes("{{") || !right.includes("}}")) {
```
aptkingston, line 7: Nice one 🙏

aptkingston (commented):
Added a few comments! Nice work, this is very much needed 👍

packages/builder/src/components/common/bindings/BindingPanel.svelte
mike12345567: Yeah so the issue I was getting around here was some weirdness around clearing the state and then adding helpers and such, the `addHelper`/`addBinding` functions both depend on the value of `value` to be able to be up to date so that you don't end up in scenarios where some old value comes back, because the state is a bit out of sync with what is in the text area.

Around the a validity, I did wonder about this myself, since anything bound to the component could pull out an invalid expression.

PR #4482, opened 2022-02-14
Try to fix safari

aptkingston:
## Description
This just removes the `jsdom` and `fs-extra` dependencies from the client library. Anyone know why they were there...?

It looks like it was `punycode` which was bundling the regex that broke safari, and a `yarn why punycode` revealed it was a deep dependency of `jsdom`. So I'm hopeful this will fix it.

Would be awesome if someone with a mac could pull this branch down, `rm -r node_modules` in the client package, do a top level `yarn install && yarn run bootstrap` and then see if the builder now works in safari.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -41,8 +41,6 @@
     "@spectrum-css/vars": "^3.0.1",
     "apexcharts": "^3.22.1",
     "dayjs": "^1.10.5",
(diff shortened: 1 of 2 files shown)


mike12345567:
Going to close this PR as its contents are now in the master release PR https://github.com/Budibase/budibase/pull/4478

PR #4501, opened 2022-02-15
Fix vite warnings

aptkingston:
## Description
This is a quick PR to fix the spam of `[vite] Failed to load source map for /node_modules/.vite/...` when running the builder in dev. This was just caused by not actually including sourcemaps in the vite config, so I've instructed vite to use sourcemaps when not running in production.

This change would probably also have helped us diagnose the recent issue with safari faster, as it would probably have pointed us to the source of the JS error being thrown due to the look behind regex.

I also addressed the other couple of vite warnings which were printed in the console - namely removing some empty style blocks and fixing some A11y issues.

Before:
image

After (no logs at all from vite in the builder):
image

### DIFF
--- a/packages/bbui/src/Button/Button.svelte
+++ b/packages/bbui/src/Button/Button.svelte
@@ -30,6 +30,7 @@
     {disabled}
     on:click|preventDefault
     on:mouseover={() => (showTooltip = true)}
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM - nice one @aptkingston

PR #4511, opened 2022-02-15
Pro package basic usage 

Rory-Powell:
## Description
- A basic usage of the pro package for testing
- Update `packages/bbui/src/SideNavigation/Item.svelte` with onClick handler for work in https://github.com/Budibase/account-portal/pull/78


packages/bbui/src/SideNavigation/Item.svelte
```diff
@@ -7,13 +7,14 @@
...
   export let disabled = false
+  export let onClick = undefined
 </script>
 
 <li
   class="spectrum-SideNav-item"
   class:is-selected={selected}
   class:is-disabled={disabled}
-  on:click
+  on:click={onClick}
```
aptkingston, line 17: Any reason why the `on:click` event which was already here didn't work?

This syntax of just `on:click` means it will propagate a `click` event automatically, without the need for a custom handler. So you can already just do `<Item on:click={foo} />` and it will work.

Just in case you weren't aware of what that syntax means.

aptkingston (commented):
Looking forward to pro! Awesome stuff Rory.

packages/bbui/src/SideNavigation/Item.svelte
Rory-Powell: Thank you for this wonderful comment. I had tried to use the native `on:click` initially but was having some trouble with the href being overridden, I think I was using it incorrectly as I've tried now with you snippet and it's fine.

Rory-Powell: Updated here
https://github.com/Budibase/account-portal/pull/78/commits/d7b2c128d56646907a2b93e7931bacd70b0bc6c5
and here
https://github.com/Budibase/budibase/pull/4525/commits/0d25fcf7352f4a75168a2bbf6f07bae0adde270c

Rory-Powell:
closing in favour of https://github.com/Budibase/budibase/pull/4525

PR #4526, opened 2022-02-16
Checklist API performance on default tenants and email invitations

### LINKED ISSUE #4522: Error inviting user but the user is emailed with an invite anyway
**Describe the bug**
When inviting a user to Budibase you sometimes get a "Error inviting user" error message. But the user gets the invitation email but cannot sign up.

**To Reproduce**
1. Setup SMTP
2. Go to Users
3. Click on Add user
4. Fill out their email address
5. Click Add user
6. "Error inviting user" appears

**Expected behavior**
The user does not receive the email if there is an error.

**Additional context**
If you create the user through basic on boarding, the user can reset their password and then gain access.

mike12345567:
## Description
There was an issue with the checklist call taking upwards of two seconds to respond. This is a relatively simple call, it shouldn't have any heavy database calls.

This also fixes #4522 - there was some issues introduced by the frontend-core that needed fixed for this flow to work correctly.

Can see the call here:
image

It appeared to be tied to the default tenant, as there was an edge case which meant that the default tenant couldn't use the more efficient app tenant lookup, instead it has to lookup the full list of app IDs. I added an efficient flag to the `getAllApps` function so that we can state that we want it to be as fast as possible, perhaps sacrificing some accuracy in the default tenant. This is fine for the checklist call because it only needs a list of apps in the tenant. This should speed up the login process for the default tenant fairly significantly.

### DIFF
--- a/packages/builder/src/stores/portal/users.js
+++ b/packages/builder/src/stores/portal/users.js
@@ -11,7 +11,7 @@ export function createUsersStore() {
   }
 
   async function invite({ email, builder, admin }) {
-    await API.inviteUser({
+    return API.inviteUser({
       email,
       builder,
       admin,
@@ -19,7 +19,7 @@ export function createUsersStore() {
   }
 
   async function acceptInvite(inviteCode, password) {
-    await API.acceptInvite({
+    return API.acceptInvite({
       inviteCode,
       password,
     })
--- a/packages/backend-core/src/db/utils.js
+++ b/packages/backend-core/src/db/utils.js
@@ -179,8 +179,11 @@ exports.getStartEndKeyURL = (base, baseKey, tenantId = null) => {
 /**
  * if in production this will use the CouchDB _all_dbs call to retrieve a list of databases. If testing
  * when using Pouch it will use the pouchdb-all-dbs package.
+ * opts.efficient can be provided to make sure this call is always quick in a multi-tenant environment,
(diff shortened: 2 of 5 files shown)


shogunpurple (approved):
Nice one! Good catch

packages/builder/src/stores/portal/users.js
```diff
@@ -11,15 +11,15 @@ export function createUsersStore() {
   }
 
   async function invite({ email, builder, admin }) {
-    await API.inviteUser({
+    return API.inviteUser({
```
aptkingston, line 14: Should be `return await` shouldn't it?

packages/builder/src/stores/portal/users.js
```diff
@@ -11,15 +11,15 @@ export function createUsersStore() {
...
       email,
       builder,
       admin,
     })
   }
 
   async function acceptInvite(inviteCode, password) {
-    await API.acceptInvite({
+    return API.acceptInvite({
```
aptkingston, line 22: Should be `return await` shouldn't it?

aptkingston (commented):
Knew there would be at least one issue with all the renaming in frontend core! I'm glad it's taken us this long to find one though haha.

packages/builder/src/stores/portal/users.js
mike12345567: Same as above.

PR #4549, opened 2022-02-17
Fix role being stale when creating screens

aptkingston:
## Description
The value of `roleId` is determined initially in `ScreenWizard` and never updated, but because `ScreenWizard` never actually remounts - the value is stale. This PR just swaps it to a reactive statement to ensure it is always up to date.

This affects the uniqueness validation as well as the role attached to new screens. This affects the patch earlier today to honour the selected role when creating screens, but this bug has always existed and was not introduced by that change.


mike12345567:
I've added a fix to this PR for an issue with query bindings, if there was multiple of the same binding e.g. `select * from {{ name }} where name = {{ name }}` it would not correctly escape the HBS bindings. Added a test case to the string templates for this issue as well.

Issue #4569, opened 2022-02-18
NGINX URL rewriting for couch db decodes URL 


Rory-Powell:
**Describe the bug**
Our NGINX configuration for accessing couch db does path rewriting on `/db/` 
The default behaviour in NGINX will decode the url before sending it on when rewriting takes place
https://serverfault.com/questions/459369/disabling-url-decoding-in-nginx-proxy

> Note that URL decoding, commonly known as [$uri](http://nginx.org/r/$uri) "normalisation" within the documentation of nginx, happens before the backend IFF:
either any URI is specified within proxy_pass itself, even if just the trailing slash all by itself,
or, URI is changed during the processing, e.g., through rewrite

This can result in data loss when sending encoded information. For example, creating a new document with an id `user+test@budibase.com`

The following URL is used is sent to NGINX:
http://localhost:10000/db/global-info/user%2Btest%40budibase.com

NGINX decodes the encoded url and couch receives:
` PUT /global-info/user+test@budibase.com`
Which couch doesn't understand, resulting in a document that looks like

```json
{
  "_id": "user test@budibase.com",
  "_rev": "1-967a00dff5e02add41819138abb3284d"
}
```

Notice the `+` has been removed.

Rory-Powell:
Note: this doesn't affect production where we talk to couch directly, but can affect local development

Rory-Powell:
This is no longer an issue due to changes in how we talk to couch locally

PR #4571, opened 2022-02-18
Deprovision and local couch db url fix

Rory-Powell:
## Description
- Fix for deprovisioning 
- Updating development couch db urls to bypass nginx - fixes https://github.com/Budibase/budibase/issues/4569


shogunpurple (approved):
LGTM! I think it makes more sense to talk to the DB directly anyway.

Issue #4547, opened 2022-02-17
Public screens no longer work


mjashanks:
Repro steps
1. Create a new app
2. Create a public screen (doesn't need to connect to any data sources)
3. Publish app
4. Browse to screen in private browsing

Page Displays `{"message":"Database does not exist.","status":404}`

### Discussed in https://github.com/Budibase/budibase/discussions/4541

<div type='discussions-op-text'>

<sup>Originally posted by **stellarfinn** February 17, 2022</sup>
I found topics covering this in the previous discussions but they were closed.

I have an app and database set to public and I cannot view it unless logged in to Budibase.

App at:
HTTPS://ciaoimports.budibase.app/app/ciaob2b

Also, I setup an app template and set the screens and database to public and this app also gives the same 404 error.

Any help much appreciated!</div>

mjashanks:
User reported that it only affects Budibase Cloud: https://github.com/Budibase/budibase/discussions/4541#discussioncomment-2209279

PR #4599, opened 2022-02-21
Forward Host header to worker and server through NGINX

Rory-Powell:
## Description
Fixes https://github.com/Budibase/budibase/issues/4547

- Forward the Host header through nginx
- This fixes tenancy parsing for subdomains
- Which fixes app list retrieval for matching custom urls
- Did not affect self host due to single tenancy

### DIFF
--- a/hosting/nginx.dev.conf.hbs
+++ b/hosting/nginx.dev.conf.hbs
@@ -34,6 +34,7 @@ http {
 
     location ~ ^/api/(system|admin|global)/ {
       proxy_pass      http://{{ address }}:4002;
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM! Good catch - I think you could probably set this inside the `http` block, and then you don't need to set it inside each `location` - I could be wrong

Rory-Powell:
Good call @shogunpurple . Setting at the http level works for all sub locations. Added in https://github.com/Budibase/budibase/pull/4599/commits/3ee7997daea49dcbef4286d36d7d0cbaaadf9220

PR #4608, opened 2022-02-21
Table V2

aptkingston:
# Description
This PR introduces a completely rewritten table component. This improves upon the previous work done to add extra configuration options for table columns. With this work, column widths, alignment, background color, text color and value can all be edited.

image

# Table rewrite
The largest chunk of work in this feature was actually having to completely rewrite the table component from scratch. Previously we used an actual HTML `table` component, but the requirement for custom widths meant that this would not be compatible. The new table is written entirely with divs, using a grid layout. There's quite a lot of custom logic to support dynamic widths of any size, row hover states and custom colors, but the result is that the new table works really well, and has a lot of big advantages over the old table.

Some of the new table features:
- Columns will only take up as much room as required, which means a massive reduction in whitespace
- Columns which can be long are capped by default at a maximum width of 200px
- If a column is given a custom width then it will always be exactly that size
- If there is empty space in the table, columns will evenly share the empty space
- Borders have been tidied up and have been made sharp
- Loading state has been fixed
- Child components added inside tables in the first column are autosized and will never wrap or overflow
- A new `Compact` setting has been added to tables, to increase the density of data further. It reduces padding and row height, but still supports the `large` table option. Here's an example of a table with this new setting enabled:
image

Here's how the new table looks with the class sales CSV compared to the old table, showing the massive reduction of empty space.
Old table:
image

New table:
image

# Column configuration
The only settings visible in the top level of the column settings drawer are the column name and column label. The column order can also be adjusted here by dragging and dropping. There are a few buttons here to add or remove columns:

- Add column adds a single column. The available columns from the dropdown will only show columns which have not been used so far.
- Add all columns will add all columns that are not already displayed in the table, but without resetting the state of any columns you have already added and customised
- Reset columns will remove all customisation of columns and show the empty state again, which will render all columns by default

By default, all table columns will be shown (as is currently the case).
Table column configuration drawer empty state:
image

Column configuration drawer with some columns added:
image

To configure further column options, the settings icon to the right of each row can be clicked. This will open a new drawer for configuring the specified column:
image

Within this new drawer, we have the following options available:
- Width: the width of the column. Any normal CSS measurments can be used (`200px`, `30%`, `10vw`, `15rem` etc). The special value `auto` can also be used, which will make the column take up as much space as needed so that all content is visible. This can lead to very wide columns if you have long strings, but can be useful for some column types.
- Alignment: how to align content inside the cell. Cell contents can be aligned to the left, center or right hand side of each cell.
- Value: how to format the value inside the cell. HBS and JS bindings are supported, but no other data bindings are available to use. The only available binding is the `{{ Value }}` of the cell. This field is intended to be used to format the value - for example adding a currency symbol to numbers, rounding values or otherwise making them display friendly. In the example above, here's the JS binding used to format the numbers in thousands, and add a £ symbol:
image
- Background color: the cell background color
- Text color: the color of the text in each cell

The new column configuration setting and compact setting are available on both the table component and table block.

# Other stuff

- Added a new custom svelte slide transition to drawers, because I hate the overshooting that happens currently when we open drawers caused by the default svelte transition
- Fixed navigating with a drawer open crashing the app because of svelte transitions on drawers trying to fire
- Fixed styling of rendering attachments in tables


packages/builder/src/components/design/PropertiesPanel/PropertyControls/ColumnEditor/ColumnEditor.svelte
```diff
@@ -18,22 +18,30 @@
...
   $: options = Object.keys(schema || {})
   $: sanitisedValue = getValidColumns(value, options)
   $: updateBoundValue(sanitisedValue)
 
+  const getSchema = (asset, datasource) => {
+    const schema = getSchemaForDatasource(asset, datasource).schema
+
+    // Don't show ID and rev in tables
+    if (schema) {
+      delete schema._id
```
shogunpurple, line 31: Hopefully not an issue but just thought I'd check - is there a chance here that the schema gets mutated, and re-saved to the DB? I remember something like this happened before, and it caused some issues.

shogunpurple (approved):
LGTM! This is such an awesome feature that many have asked for. Great job 🚀

packages/builder/src/components/design/PropertiesPanel/PropertyControls/ColumnEditor/ColumnEditor.svelte
aptkingston: Well remembered and yea it makes sense to double check! In this case we're all good, because `getSchemaForDatasource` always ensures to deep clone any real table schemas to ensure they aren't mutated.

The resulting schema that we delete these columns from is then only ever used in these components to determine which column options to make available, but is never saved or used for anything else.

Good call on double checking though.

joebudi:
@aptkingston this is even better than I expected. Wonderful work!

packages/bbui/src/Drawer/Drawer.svelte
```diff
@@ -21,11 +22,27 @@
     }
     visible = false
   }
+
+  const easeInOutQuad = x => {
+    return x < 0.5 ? 2 * x * x : 1 - Math.pow(-2 * x + 2, 2) / 2
```
PClmnt, line 27: Could you explain what is going on here, just for my benefit haha

PClmnt (approved):
This is class, definite game changer!

packages/bbui/src/Drawer/Drawer.svelte
aptkingston: Haha of course! The purpose for replacing the default svelte `slide` transition in the first place is because it has an overshoot built in (goes up too far then jumps down a bit) and I think it looks gimmicky, so I wanted to replace it with our own slide transition.

This is just a quadratic function to smooth the new slide in animation for the drawer. Svelte provides a param `t` for custom animations, which goes from 0 to 1. If we linearly interpolated the animation relative to `t`, we would have a completely linear and smooth animation. This doesn't actually look particularly natural though - which is why CSS offers things like `ease-in` or `ease-out`, which slowly accelerates or decelerates the animation.

Since we're using a JS transition we can't use CSS helpers like those, so this is just a quadratic maths functions to minic what `ease-in-out` does. Plotted on a graph it looks like this:
image

It basically starts off slightly slower, speeds up in the middle, and then slows down again towards the end. This makes the transition feel much smoother.

joebudi: Haha this is great. A true perfectionist.

PR #4629, opened 2022-02-22
Convert roleId prop to reactive prop

### LINKED ISSUE #4506: Can't create screens with different roles
**Describe the bug**
When I try to create the same screen for 2 different roles, one overrides the other.

**To Reproduce**
1. Go to design
2. Create a home screen under the Basic role (Name: Home, Url: /)
3. Go to "Power" role, and do the same (Name: Home, Url: /)
4. Notice power screen overrides the Basic screen

**Expected behavior**
2 Screens should exist, one for each role

**Screenshots**

**Additional context**
App export: 
[Role Based Screens-export-1644929422707.txt](https://github.com/Budibase/budibase/files/8069197/Role.Based.Screens-export-1644929422707.txt)

mslourens:
## Description
The prop roleId is set after the creation of the screen, which always results in 'BASIC'. When it's reactive, it will always re-evaluate it's value when used.
Fixes #4506

### DIFF
--- a/packages/builder/src/components/design/NavigationPanel/ScreenWizard.svelte
+++ b/packages/builder/src/components/design/NavigationPanel/ScreenWizard.svelte
@@ -6,13 +6,14 @@
   import { store, selectedAccessRole, allScreens } from "builderStore"
   import analytics, { Events } from "analytics"
 
+  $: roleId = $selectedAccessRole || "BASIC"
+
   let newScreenModal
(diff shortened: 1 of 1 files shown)


aptkingston:
Hey @mslourens! We actually already have a PR up for this:
https://github.com/Budibase/budibase/pull/4549

I should have linked it to the original issue - sorry that I didn't.

PR #4632, opened 2022-02-22
Fix dependencies

aptkingston:
## Description
- Fix builder lockfile
- Revert server lockfile
- Fix client lib dependencies and lockfile from bad merges
- Add nuxt config file to finally actually silence the vite spam of:
image
- Fix A11y issue with button again because it was deleted in merges
- Fix vite warning about `className` usage


shogunpurple (approved):
LGTM! Nice fixes

Rory-Powell (approved):
Nice one

PR #4637, opened 2022-02-22
Setting Video flag to False

Mitch-Budibase:
There seems to be an issue sometimes with video processing and test runs 

--Setting this to false for now while i investigate


shogunpurple (approved):
LGTM

PR #4638, opened 2022-02-22
Table Row Selection

PClmnt:
## Description
This PR adds the ability to select / multi-select a number of rows in a table and use them in bindings. This will allow cool stuff like the exporting of specific table in a data, or you could select a number of rows and pass them to automations using a button action etc. Stuff like that. 

You can enable this feature by clicking this checkbox in the Table settings

image

A couple of changes have been made to the BBUI table component in order to enable this. There was an already the ability to select rows in the data tab, but you lost your selection every time you navigated to a new page, this PR fixes that and your selection is maintained over whatever amount of pages you have in the table. 

Also new is the Checkbox used to select all rows on your current page (just added to the top left corner like here: https://spectrum.adobe.com/page/table/). This Checkbox will _**only**_ select the rows on that page, not for the entire table. There's a little bit of hacky code in here as well, that basically ensures the Checkbox deselects when you navigate to a new page with no rows. The functuionality around this is never going to be perfect, essentially because the table has no idea what page it is as Pagination is a separate area altogether.

You can also see in the below screenshot that a row count has been added, just to keep track of what size your currrent selection is. 
image

To allow for the handling of multiple tables, each selection is shown in the binding drawer like so 

## Screenshots
Screenshot 2022-02-22 at 15 16 30

Each of these are just an array of Row IDs. 

## Updates

A few quick updates made after review:
- Readable bindings are now called "Selected rows"
- Row selection bindings are added for table blocks too
- The text saying "X record(s) selected" has been updated to "X rows selected" (with optional "s") and is only shown now when rows are actually selected, and is not absolutely positioned to avoid being hidden due to overflow
- Data bindings are only generated for tables in the current asset


aptkingston (approved):
Great stuff Peter! Trying to get these cypress tests to pass and then let's get this merged.

PR #4649, opened 2022-02-23
Cypress port conflicts / local dev fix 

Rory-Powell:
## Description
- Prevent port conflicts when running cypress locally alongside account portal
   - Account portal nginx runs on `10001`
   - Change `10001 / 10002` in cypress to be `4100 / 4200` for server and worker respectively
   - 1000X gives the impression we are using a proxy, which cypress does not. Use the above to mimic 4001 / 4002
- Update yarn commands to fix local dev


shogunpurple (approved):
LGTM

PR #4684, opened 2022-02-24
Add status banner + fix templates 403 redirect loop

Rory-Powell:
### Banner
- Add status banner
- Banner appears in builder only 
- For more details see: https://github.com/Budibase/account-portal/pull/84

### Templates 403 loop
- When the `currentapp` cookie is present for an app in another tenant the `builder` and `admin` properties are incorrectly removed from the current user 
- When the builder is accessed, remove the currentapp cookie if the app belongs to another tenant.

## Screenshots

Screenshot 2022-02-24 at 14 55 42


shogunpurple (approved):
LGTM! Will be so helpful to have this during an outage👌

aptkingston (commented):
Love this Rory 🚀 

How does the banner interact with normal notifications? Basically just wondering if it would be better to add a feature to notifications to show buttons (which I think is the only difference) and the just use the normal BBUI `notifications` module to show it. I think at the minute this will render above (depth wise) normal notifications will it?

Either that, or make the banner a full width header that actually shifts the page down or something, so it wouldn't affect notifications.

Rory-Powell:
> How does the banner interact with normal notifications?
> I think at the minute this will render above (depth wise) normal notifications will it?

@aptkingston you're spot on it renders above normal notifications. Making this a normal notification would be a nice way to integrate them better, the only other usage of the banner is on the rest transformer here:

Screenshot 2022-02-28 at 07 18 45

Which we could probably convert as well, although it might be strange to have a notification going outside of the normal usage and being embedded like that. I like the idea of making the banner display full width, which I think is probably the correct ux here as well - I'll take a look

packages/builder/src/App.svelte
```diff
@@ -1,13 +1,16 @@
...
-  import { NotificationDisplay } from "@budibase/bbui"
+  import { NotificationDisplay, BannerDisplay } from "@budibase/bbui"
   import { parse, stringify } from "qs"
   import HelpIcon from "components/common/HelpIcon.svelte"
 
   const queryHandler = { parse, stringify }
 </script>
 
+<div class="banner-container" />
+<BannerDisplay />
```
aptkingston, line 12: This seems a bit strange to me - rendering a DOM node and then immediately after rendering a svelte portal which targets that DOM node. Why don't we just rename `BannerDisplay` to `StatusBanner`, remove the portal altogether and the `.banner-container` div, and just have the `StatusBanner` render a normal banner with status info in it's current location? That would achieve the same result but be a bit less complicated, and more explicit that it's intended to be used for status messages rather than a generic banner.

aptkingston (commented):
Looks great @Rory-Powell! I much prefer the full width approach. Just left one suggestion about simplifying things and then it should be good to go :+1:

packages/builder/src/App.svelte
Rory-Powell: Sounds like a good suggestion @aptkingston . I had a go at this but hit a roadblock around pointer events, the pointer isn't able to interact with with banner without the portal. Attached is the diff of what I tried, if you have any other suggestions of how to work around this I'd be up for trying them out. Gonna merge this one for now so that we can release the two fixes in the release

Issue #4175, opened 2022-01-24
REST Api UX - Default endpoint naming rarely makes sense


mjashanks:
When using the REST api connector, I always rename my endpoint, because it always looks like this...

![image](https://user-images.githubusercontent.com/3524181/150819370-33909f01-88c3-485f-94a1-4bdcd708fb37.png)

I.e. when using the full host, I always see an ellipsis - which blocks the actual endpoint name. Maybe we should always be using variables like `{{baseURL}}/api/path`, or maybe it would be better to default to `/api/path` ?

Rory-Powell:
I like the path idea because it will work in all situations

shogunpurple:
Yep, sounds good - let’s strip the host from the display in the nav items on the left to make it easier to read. The only issue is that since the host is overridable per query now, the paths could all be for different hosts. Are we ok with that? 

/api/something could appear twice, but point to 2 different APIs

mjashanks:
I think 1 host + many paths is much more common than many hosts.

Just need the 80% case

Rory-Powell:
>  /api/something could appear twice, but point to 2 different APIs

I'd be alright with this, because the user is still free to rename one if they want to distinguish. It's unlikely to happen during import because you wouldn't have an api definition from multiple hosts

PR #4700, opened 2022-02-25
Improve REST query naming in navigation

aptkingston:
## Description
This PR updates the displayed names of REST queries in the navigation.
Described in #4175.

## Behaviour
- Protocol is trimmed
- Double leading slash is trimmed
- Trailing slash is trimmed
- Domain is trimmed if path exists
- Domain is shown if no path exists

Example of REST datasource:
image

## Examples of new naming

| Query URL | Displayed URL |
| --- | ---| 
| foo.bar | foo.bar |
| foo.bar/ | foo.bar |
| /foo/bar | /foo/bar |
| //foo.bar | foo.bar |
| //foo.bar/baz/ | /baz |
| http://foo.bar | foo.bar |
| http://foo.bar/ | foo.bar |
| http://foo.bar/baz | /baz |
| {{variable}}/foo/bar/ | /foo/bar |
| {{variable}} | {{variable}} |
| /{{variable}} | /{{variable}} |
| http://foo.{{variable}}.bar/foo/bar/ | /foo/bar |
| https://{{subdomain}}.{{domain}}.com/{{foo}}/bar/{{baz}}/ | /{{foo}}/bar/{{baz}} |

### DIFF
--- a/packages/builder/src/helpers/data/utils.js
+++ b/packages/builder/src/helpers/data/utils.js
@@ -109,6 +109,36 @@ export function customQueryIconColor(datasource, query) {
   }
 }
 
+export function customQueryText(datasource, query) {
+  if (!query.name || datasource.source !== IntegrationTypes.REST) {
+    return query.name
+  }
+
+  // Remove protocol
+  let name = query.name
+  if (name.includes("://")) {
+    name = name.split("://")[1]
+  }
+
+  // If no path, return the full name
+  if (!name.includes("/")) {
+    return name
+  }
+
+  // Remove trailing slash
+  if (name.endsWith("/")) {
+    name = name.slice(0, -1)
+  }
+
+  // Only use path
+  const split = name.split("/")
+  if (split[1]) {
+    return `/${split.slice(1).join("/")}`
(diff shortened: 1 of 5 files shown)


packages/builder/src/helpers/data/utils.js
```diff
@@ -109,6 +109,36 @@ export function customQueryIconColor(datasource, query) {
   }
 }
 
+export function customQueryText(datasource, query) {
```
Rory-Powell, line 112: Great to see the example usage of this in the pr description. Could we add unit tests for this function?

aptkingston: Good call actually, it's a perfect candidate. Will add some in.

aptkingston: Never mind actually - I forgot that we nuked jest from the builder since it was so horribly broken. This is just something we'll need to remember to add in once we've got a working jest setup again.

shogunpurple (approved):
LGTM! URLs look super clean - love the way it accounts for all different configurations.

PR #4703, opened 2022-02-25
Budibase Command Palette

shogunpurple:
## Description
This PR introduces the budibase command palette. This allows users to hit the ctrl key when they are inside an app in the builder. They will then be presented with a panel that allows you to type "commands", which you can then select with the arrow key and hit enter to run. This is a feature that should dramatically speed up workflows in budibase, and allow users to comfortably use the platform without the use of the mouse, in some cases.

Here's the commands that the command palette currently supports:
- Deploy an app
- View published app, preview
- Navigate to automations and specific automations
- Navigate to design and specific screens
- Navigate to data and specific datasources or queries
- Open a discussion on github
- Report a bug on github
- Change the theme of the builder
- Navigate to Portal

When a command is run, the panel is closed. You can also hit ctrl again to dismiss the palette.

**Future work**
- Allow for completely configurable keyboard shortcuts for each command
- Adding more commands based on requests from the community

## Screenshots

https://user-images.githubusercontent.com/11256663/155741196-9229912e-0973-4009-9df0-158cc47e95d5.mov


aptkingston (commented):
Gimme gimme gimme! This is so awesome.

Not sure what to do about the issue I mentioned though during the demo. How we handle URL and state synchronisation isn't consistent across the app. Basically we only hydrate state from the URL once on the initial load - although we do this at various depths typically inside the routify `_layout.svelte` or `index.svelte` files. 

If we navigate to a URL that's different enough from the current one, then things will probably work fine as the remounting of the routify layouts will trigger a fresh hydration from the URL.

If we navigate to a similar URL (like changing between two tables in the demo) then this won't work, because we won't update state from the URL again and so the new table won't actually get selected.

In general I much prefer for state to the source of truth, with hydration only carried out once, and state dynamically updating the URL as required - which is how the design section works. The data section hydrates once but when clicking around it will both update state and the URL which is very error prone and can lead to inconsistencies and broken URLs. That preface is just to say that I don't really want to refactor everything to make the URL the source of truth (which would solve the issue here).

The best solution overall would be writing a really clean middleware to keep state and URL in sync, and support dynamic changes to both. I've done this before for react and react router. That's a big job of course and it's probably not an urgent refactor considering we thankfully don't have too many bugs from our current solution.

To get this merged in (because it's awesome) I think we could try a solution like adding an optional query param `?hydrate` which is supported on any URL, and will force the hydration logic to run, regardless of whether it's been run before.

Required changes would then be:
- Update the command palette to always append the `hydrate` query param to generated URLs
- Update the routify files so that anywhere we update state from the URL, we check for the prescence of this param and run the hydration regardless of initial load or not

I don't think the scope of this is too bad because I know the location of most of those routify hydration functions already. I'd be happy to jump on this branch and do that if you want.

Sorry for the huge post. Other ideas welcome!

joebudi:
This is an incredibly powerful feature and will have a major say in how we signpost actions in the future. It would be good to chat more about this @shogunpurple

shogunpurple:
@joebudi happy to - it could do with some design/UX input, and was mainly an experiment to speed up the budibase workflow.

@aptkingston great points - the final goal is to just completely refactor the way state is managed in the data UI so all this stuff just works. The hydrate parameter is a good solution to trigger the correct state updates even when changing to a sibling route. I guess with this feature given that it’s only doing this in a few places - it might even be possible to add the state hydration logic inline in the CommandPalette function handlers too. Regardless I will create a ticket for the refactor of data UI state - having a consistent way to do this throughout budibase will serve us well as we inevitably add more screens at various depths in the UI. 

Don’t think it will be a huge effort to get this specific piece of work over the line - as mentioned it would be good to get some feedback on the design front as to wel

aptkingston:
Just a quick update to my above points - I've added a generic utility for syncing state to the URL as part of the new design UI work. Once that is done, we can easily use this utility everywhere else in the builder. This would then let us get this awesome feature merged, because then navigating directly to any route would correctly update state.

aptkingston:
Updated to new designs and added categories. The keybind is now ctrl+k/cmd+k.
image
image
image

aptkingston:
Closing in favour of https://github.com/Budibase/budibase/pull/9942 which includes this work.

PR #4708, opened 2022-02-25
[feature] select all datatable rows

conorw:
## Description
In the admin data screens I would like to select all rows by clicking a "select all" button.
The checkbox should auto check if all the rows are manually selected.
The checkbox should auto un-check if any of the rows are manually deselected.
The checkbox should auto un-check when new rows are loaded.

N.b. I was unable to run the cypress tests to determine if the cypress test change was valid.

## Screenshots
image
image

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -5,6 +5,9 @@
   "trailingComma": "es5",
   "arrowParens": "avoid",
   "jsxBracketSameLine": false,
-  "plugins": ["prettier-plugin-svelte"],
+  "endOfLine": "auto",
+  "plugins": [
+    "prettier-plugin-svelte"
+  ],
   "svelteSortOrder": "options-scripts-markup-styles"
 }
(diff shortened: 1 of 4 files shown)


.prettierrc.json
```diff
@@ -5,6 +5,9 @@
   "trailingComma": "es5",
   "arrowParens": "avoid",
   "jsxBracketSameLine": false,
-  "plugins": ["prettier-plugin-svelte"],
+  "endOfLine": "auto",
```
shogunpurple, line 8: Why has this been added?

shogunpurple (approved):
LGTM apart from the prettier update - just not sure why it's there.

Thanks for the contribution!

.prettierrc.json
conorw: this is for windows machines where it flags as a prettier error on every page if not present

PR #4732, opened 2022-02-28
Budibase public API

mike12345567:
## Description
This PR contains the work to implement a public facing API for the backend of Budibase.

Currently this supports five endpoints (CRUD + search) for 4 resources, (tables, users, rows and applications) and then two endpoints for queries (search and execute).

The API is defined using the [OpenAPI 3.0 specification](https://swagger.io/specification/) - I've built a generator for this so that we can provide the spec in both YAML and JSON formats.

You can find the current definitions here: [YAML](https://github.com/Budibase/budibase/blob/feature/budibase-api/packages/server/specs/openapi.yaml) and [JSON](https://github.com/Budibase/budibase/blob/feature/budibase-api/packages/server/specs/openapi.json).

There's a few parts to this work so I'll break it down quickly here into the important components.

### New controllers
This work updates the controllers/routes of the server to have `controllers/public` and `routes/public` - they implement their routes, middlewares and controllers somewhat independently to the internal API. While the internal API provides all this functionality I wanted to make sure we had complete control over the public endpoints - most of the work carried out by the controllers of these public endpoints is making sure the format of the input is ok and then formatting the output to be consistent across the whole public API. There are a few decisions that have been made to simplify the API/make it more upgrade compatible than our internal API.

1. By default all the public routes are under `/api/public/v1` meaning that we can increment the version easily.
2. All responses from the API are wrapped objects now, a lot of our internal API simply returns the output but this means that if we ever want to expand the response, e.g. including a bookmark for pagination or quota usage info, we have to revision the API. By wrapping the output (e.g. `{ row: {... } }`) we can easily expand upon the API.
3. The API is defined a bit more dynamically, rather than each route file generating its own Koa router, which is then merged with the primary router, I've built an `Endpoint` class - each of the routes builds a list of `read` and `write` endpoints. This means that middlewares can be applied in a central location, e.g. for authorization, param lookup and the likes.

### Generation
The usual method for building OpenAPI specs is to handwrite the full file - it is realistically required as the descriptions, enumerations etc need to be defined and all this may not be defined in code. Rather than writing this as one fully manual file I wanted to bring the definitions a bit closer to the code, so after trying a few methods, I settled on the following toolset:
1. [swagger-jsdoc](https://www.npmjs.com/package/swagger-jsdoc) - this allows us to generate the full spec from JSDoc comments.
2. [jest-openapi](https://www.npmjs.com/package/jest-openapi) - this allows us to test the final generated spec against the API.

If you look at the public route files you'll see definitions like:
```js
/**
 * @openapi
 * /queries/{queryId}:
 *   post:
 *     summary: ....
 */
```

This defines the schema + examples for the API endpoint as a comment right beside it in the code, making it easier to see what needs updated if changes are being made to the API.

We also make relatively heavy use of open API references for schemas and examples. You can find all of the examples/schemas defined in `packages/server/specs/resources` - they've been broken up by the resource they are associated with e.g. tables. This makes the process of defining examples and specs significantly easier as things can be re-used in ways that open API references don't support and examples etc can be generated dynamically in JS.

### Testing
Due to the way the spec is generated/written its very possible for it to be inaccurate. I wanted to make sure that what the schema/examples state will be returned by the API is infact correct, this is where `jest-openapi` comes in. This package allows us to compare the response of an endpoint to the schema specified for it - I've written some basic tests for now in `packages/server/src/api/routes/public/tests/compare.spec.js`.

### Release
I've updated the Github CI release action to generate the open API specs and include them as part of the release resources. This should mean that the specs that show up in the release will always be accurate.

### API keys
This utilises a new mechanism which allows a user to generate an API key for themselves. This API key is encrypted, it contains information about the tenant and their user ID. Users can view their API key from the Budibase portal - if this is used to hit the API the `authenticated` middleware will lookup the user by the API key, confirm it is all valid and then from that point forward it will appear to the server/worker as a totally standard user session. A user can only have one API key active at a time, if they want to invalidate their API key them simply generate a new one.

## Screenshots
Showing the developer info menu item.
image

Showing the API key modal.
image

Showing the API spec being loaded into Swagger editor.
image


aptkingston (approved):
image

Literally unreviewable but I'm approving because I love it 💪 Amazing work!

packages/builder/src/components/settings/UpdateAPIKeyModal.svelte
```diff
@@ -0,0 +1,37 @@
...
+      notifications.success("New API key generated.")
+    } catch (err) {
+      notifications.error("Unable to generate new API key")
+    }
+    // need to return false to keep modal open
+    return false
+  }
+
+  onMount(async () => {
+    apiKey = await auth.fetchAPIKey()
```
aptkingston, line 21: We just need to wrap this in a try/catch too so we don't miss any errors.

packages/builder/src/stores/portal/auth.js
```diff
@@ -172,6 +172,14 @@ export function createAuthStore() {
         resetCode,
       })
     },
+    generateAPIKey: async () => {
+      const info = await API.generateAPIKey()
+      return info?.apiKey ? info.apiKey : null
```
aptkingston, line 177: Super minor but we can just shorten this to `return info?.apiKey` (since it's not like it can return literal `false` or anything so a falsey check falling back to `null` won't do anything extra).

Or if this call will only ever return an API key and not anything else, we should move this into `frontend-core` so that the actual API call just returns the string directly.

packages/builder/src/stores/portal/auth.js
```diff
@@ -172,6 +172,14 @@ export function createAuthStore() {
...
+    },
+    fetchAPIKey: async () => {
+      const info = await API.fetchDeveloperInfo()
+      return info?.apiKey ? info.apiKey : null
```
aptkingston, line 181: Same as above above oprtional chaining, but this one looks like it will return more info than just an API key so that part doesn't apply here.

packages/builder/src/components/settings/UpdateAPIKeyModal.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import { auth } from "stores/portal"
+  import { onMount } from "svelte"
+  import CopyInput from "components/common/inputs/CopyInput.svelte"
+
+  let apiKey = null
+
+  async function generateAPIKey() {
+    try {
+      apiKey = await auth.generateAPIKey()
+      notifications.success("New API key generated.")
```
aptkingston, line 12: NAB: If you remove the full stop it will make me smile 🙏

joebudi:
Great work @mike12345567 . I have a strong feeling this is going to put a smile on a lot of developers' faces. One small possible update I'd propose is making changing the label 'View developer information' to 'View API key'. This is, unless, you plan to add more 'developer information' to the modal.

packages/frontend-core/src/api/user.js
```diff
@@ -15,7 +15,7 @@ export const buildUserEndpoints = API => ({
    */
   fetchBuilderSelf: async () => {
     return await API.get({
-      url: "/api/global/users/self",
+      url: "/api/global/self",
```
aptkingston, line 18: We should probably move these endpoints into the new `self` file then shouldn't we? If we have a file for just the `self` routes then these 3 endpoints (`fetchSelf`, `fetchBuilderSelf` and `updateSelf`) probably belong in there rather than in users.

aptkingston (commented):
Joking about being unreviewable, I did review it 👌 Huge amount of work gone into this.

packages/builder/src/components/settings/UpdateAPIKeyModal.svelte
mike12345567: Would it be an idea to adjust the notification to strip full stops if they are at the end? Will fix it in code too but would be good if component made sure that a convention was always followed too!

packages/server/src/api/routes/public/applications.ts
```diff
@@ -0,0 +1,155 @@
...
+  new Endpoint("post", "/applications", controller.create).addMiddleware(
+    applicationValidator()
+  )
+)
+
+/**
+ * @openapi
+ * /applications/{appId}:
+ *   put:
+ *     summary: Update an existing application by its ID.
```
Rory-Powell, line 44: We can probably update the terminology across the summary statements to be a bit more consistent. 

e.g: 
- PUT /applications/{appId} Update an existing application by its ID.
- PUT  /tables/{tableId}/rows/{rowId}  Update a single row within a specified table.
- PUT /tables/{tableId} Update the specified table.

Each have some slight differences in sentence structure. I also noticed these are a bit lengthy on the readme screenshot that rene posted, if we could shorten them all down to a consistent format like:
```
- POST    /path                Create an/a X
- PUT     /path/id             Update an/a X
- GET     /path                Retrieve all X
- GET     /path/id             Retrieve an/a X
- POST    /path/search         Search Xs
```

packages/server/src/api/routes/public/applications.ts
```diff
@@ -0,0 +1,155 @@
...
+ *               application:
+ *                 $ref: '#/components/examples/application'
+ */
+read.push(new Endpoint("get", "/applications/:appId", controller.read))
+
+/**
+ * @openapi
+ * /applications/search:
+ *   post:
+ *     summary: Search for an application based on its app name.
```
Rory-Powell, line 120: Less might be more here given the search endpoint can be expanded over time. Simply `Search Applications` would be a decent descriptor. 

If you're dead set in calling out the name we could remove the `app name` bit as it's implied and just go `name`

packages/frontend-core/src/api/self.js
```diff
@@ -0,0 +1,54 @@
...
+      url: "/api/global/self/api_key",
+    })
+    return response?.apiKey
+  },
+
+  /**
+   * retrieves the API key for the logged in user.
+   * @return {Promise<object>} An object containing the user developer information.
+   */
+  fetchDeveloperInfo: async () => {
```
Rory-Powell, line 18: Should it be just `fetchAPIKey` to match the above / the endpoint?

packages/server/src/api/controllers/public/applications.ts
```diff
@@ -0,0 +1,71 @@
...
+    app._id = params.appId
+  }
+  return app
+}
+
+async function setResponseApp(ctx: any) {
+  if (ctx.body && ctx.body.appId && (!ctx.params || !ctx.params.appId)) {
+    ctx.params = { appId: ctx.body.appId }
+  }
+  await controller.fetchAppPackage(ctx)
```
Rory-Powell, line 21: Is there being returned in the response here than the docs suggest?
in `fetchAppPackage`
```js
  ctx.body = {
    application,
    screens,
    layouts,
    clientLibPath: clientLibraryPath(ctx.params.appId, application.version),
  }
```

packages/server/src/api/controllers/public/rows.ts
```diff
@@ -0,0 +1,75 @@
...
+    limit,
+    query,
+  }
+  await rowController.search(ctx)
+}
+
+export async function create(ctx: any) {
+  ctx.request.body = fixRow(ctx.request.body, ctx.params)
+  await rowController.save(ctx)
+  ctx.body = { row: ctx.body }
```
Rory-Powell, line 44: I really like the idea of wrapping the raw response in an object. I think there's room here to generify the wrapping by using a `data` key instead of `row`, `table` etc, as we already know the resource type by the path and this will make it easier for clients to consume. 

e.g. stripe api 
```js
{
  "object": "list",
  "url": "/v1/products",
  "has_more": false,
  "data": [
    {
      "id": "prod_LEuBjywZYxGfuH",
      "object": "product",
      "active": true,
      "created": 1646120649,
      "description": null
   }]
}
```

Usually you only see this type on object wrapping on collections endpoints btw, for single object responses it's fine to just return the resource directly

Rory-Powell:
Really great work on this, I particularly the documentation generation framework you have put in place which will make maintaining this 10x easier.

Rory-Powell (changes requested):
There are two changes that I think are required before this becomes shippable
- Decouple database entities from API responses
- Decouple internal API controllers from public api controllers

I'll elaborate a bit further below

Rory-Powell:
**Decouple database entities from API responses**
- Use DTOs 
   - e.g. AppDto.ts or AppResource.ts vs the db entity (document) App.ts 
- Convert entity to DTO at web layer using a `Converter` helper

The benefits here are numerous, but the main ones are 
- no accidental data leakage
- freedom to make internal changes without affecting api responses

 This would look something like: 

Screenshot 2022-03-01 at 09 09 49

Rory-Powell:
**Decouple internal API controllers from public api controllers**

The approach to use would be to copy-paste shared logic out of the internal controllers into a `sdk` package. 

The main benefits here:
- freedom to make changes to the internal api without affecting public api resources
- step away from coupling our application layer to koa `ctx`
- introduce better type safety across the board, which in genera will drive quality up and bugs down 

This would look something like:
```
/src
   /api
      /controllers
         /public
            row.ts // this parses web information from koa and passes an object to sdk
         row.ts // same as above 
   /sdk
      row.ts // core logic for CRUD rows
```

packages/frontend-core/src/api/self.js
mike12345567: Wanted to have a single endpoint to generate, but a fetch that could fetch the whole doc of developer info. In the future if there is any other developer info can get it all from a single endpoint, but we might have multiple endpoints to re-generate components of it.

packages/server/src/api/controllers/public/applications.ts
mike12345567: Yes, there is, but I didn't want to document the screens, layouts and client lib stuff as its not useful to the user right now. I thought though it would be good to return everything that we when we get an app through our internal API, as it means its consistent at least.

In the future we can document it, but I think we'd need to expose the ability to alter the screens/layouts through the public API as well before we expose/explain that extra stuff.

packages/server/src/api/controllers/public/rows.ts
mike12345567: Personally I'm not a fan of using a generic wrapper, I also really like wrapping in general, theres no negative to the user experience and it allows us to add on anything extra in the future, I've had too many issues with single object responses returning directly in the past.

packages/server/src/api/routes/public/applications.ts
mike12345567: Was thinking about doing this, will add a description with any other info that I'd like it to have and keep the summary small.

mike12345567:
I've made some of the changes discussed above that were within the scope of this PR/issue. I've added the following:
1. A new package, [openapi-typescript](https://www.npmjs.com/package/openapi-typescript) which generates Typescript definitions from our OpenAPI specification - this is now part of the spec generation.
2. Mapping functions, which take the `ctx` and map over the body to make sure that the response type is correctly within the Typescript definition.
3. A middleware which uses the URL resources to determine which mapping function should be used - determining if it is a search/expected to return an array or whether it should just be returning a single object, then maps on the wrapper, as well as mapping the objects themselves.
4. All public controller endpoints have been updated to continue the middleware chain so that output middlewares can be utilised.
5. The summaries of all endpoints have been updated to follow a simplified common format.
6. Descriptions have been added to some endpoints where information was missing that was previously in the summary.
7. Some updates to the OpenAPI spec for queries, previously there was no definition for the query output but some of this could be defined (like pagination info).
8. A common wrapper is now used of `data` this is applied at the mapper level for consistency.

Parts that have been left out for now:
1. The SDK/ORM update, this is outside the scope of this issue and needs further design, its a major overhaul of the internal server, it would allow us to decouple not only the public API, but also the automation actions from the internal API. This will be designed and put forward for review so that we can refactor the various components which are coupled to the internal API to utilise a SDK instead.
2. Currently the mapper doesn't understand the structure of the input, meaning its usefulness is somewhat degraded, right now it isn't possible for it to detect at compile time an inaccuracy between the internal API output and the expected output API format. This will again come with the SDK when we create types for everything.

packages/server/src/api/controllers/public/mapping/applications.ts
```diff
@@ -0,0 +1,32 @@
+import { Application, ApplicationOutput } from "./types"
```
Rory-Powell, line 1: It could be nice if these were inverted, where `Application` is the full resource (the output) and the input one becomes `ApplicationInput`

packages/server/src/definitions/openapi.ts
```diff
@@ -0,0 +1,964 @@
...
+    };
+    /** @description The table to be created/updated. */
+    table: {
+      /** @description The name of the table. */
+      name: string;
+      /** @description The name of the column which should be used in relationship tags when relating to this table. */
+      primaryDisplay?: string;
+      schema: {
+        [key: string]:
+          | {
```
Rory-Powell, line 633: It would be nice if we could get these `|` blocks as named types, atm it's a bit unclear what each one represents

Rory-Powell:
These latest changes are 🔥 having the public types generated from the openapi spec is incredible for keeping the api  compliant with the documentation. Brilliant as well for when we introduce the budibase javascript client as we can re-use them there really easily. Great job!

Rory-Powell (approved):
LGTM 👍

packages/server/src/definitions/openapi.ts
mike12345567: I've removed the type definition, apparently in OpenAPI the defined way to say that its any type is with no type!

shogunpurple (commented):
Apologies that I only got round to reviewing this now after merge. Honestly there's nothing I have to add that hasn't been highlighted/addressed already. This is a fantastic (and huge) piece of work. 

In addition of the actual API and the custom automation that you have put together to generate specs from docs, there's a **ton** of by-product stuff in here that increases the type-safety, security (encryption), documentation (in the specs/resources folders) and just overall quality of the backend of the entire platform. Great job 🚀

PR #4737, opened 2022-02-28
Improve parsing of REST query strings to support equals symbol in values

### LINKED ISSUE #4667: REST Query Params with "=" in them get truncated at first "="
**Describe the bug**
When a Query is added for a REST data source, and a parameter is added with a value that has an "=" in it, the query will work on first try but after saving and going back into the query, the value is truncated to the first "="

**To Reproduce**
1. Go to the REST datasource.
2. Click on '+ Add Query'
3. Add your query URL
4. Add a param.  Enter a key name. (eg. 'conditions')
5. Enter a value:  eg.  board/name="Service" and summary like "The system %"
6. Save it.
7. Send it and it works.
8. Exit out of the query screen and then go back into it.
9. The "value" for the key is now truncated.  eg.  board/name

**Expected behavior**
Value should retain the full value, even if it contains "="

On version 0.11.3 of Budibase

aptkingston:
## Description
This PR updates the builder parsing of REST queries to support an equals symbol `=` in values, as previously this was not accounted for. While I would highly discourage using an `=` in a query string with multiple values, it is technically a valid URL.

Fixes https://github.com/Budibase/budibase/issues/4667.

### DIFF
--- a/packages/builder/src/helpers/data/utils.js
+++ b/packages/builder/src/helpers/data/utils.js
@@ -30,8 +30,8 @@ export function breakQueryString(qs) {
   const params = qs.split("&")
   let paramObj = {}
   for (let param of params) {
-    const [key, value] = param.split("=")
-    paramObj[key] = value
+    const split = param.split("=")
+    paramObj[split[0]] = split.slice(1).join("=")
   }
   return paramObj
 }


Rory-Powell (approved):
I can't believe it's a valid url, but good fix 👍

PR #4750, opened 2022-03-01
Fixes for 409s and option picker values

### LINKED ISSUE #4744: Option Picker does not show all rows from Data Provider (with `0` values being part of the issue)
**Describe the bug**
The option picker does not show all rows from a bbdb table used as a Data Provider for the picker (basically second problem from #4721).

A `priority` table has three columns: (id: number 0-9, name: text, name_display: formular, {{id}} - {{name}}). Adding rows (0, none) (1, low), (2, medium) and (3, high) should result in four options to choose from, but the option picker only shows 1 and higher.

**To Reproduce**
1. Create a table `priority` with three columns (id: number 0-9, name: text, name_display: formular, {{id}} - {{name}}).
2. Add rows with the value (0, none) (1, low), (2, medium) and (3, high)
3. Create an Data Provider for table `priority` on a screen
4. Add a Option Picker that uses the Priority Data Provider with _name_display_ as `Label Column` and _id_ as `Value Column`
5. Run the screen and open the Option Picker.
6. Only values _1 - low_, _2 - medium_, and _3 - high_ are displayed. _0 - none_ is missing.

Step 4 will also throw the `Error updating component prop` error described in #4721.

**Expected behavior**
The Option Picker should show all values from the Data Provider

**Example**
See the file attached to #4721

 - OS: Windows 10
 - Browser Firefox
 - Version 97.0.1
 - App Version 1.0.74 (BB cloud)

**Additional context**
- Zero (`0`) values are displayed in the tables of the Data and Design section, but not on the edit/update screen (field is empty).
- When saving the row in the Data section without reentering the `0` for _id_ results in an empty _id_ field.

aptkingston:
## Description
This PR addresses a couple of issues raised in https://github.com/Budibase/budibase/issues/4744 and https://github.com/Budibase/budibase/discussions/4721.

- Update option pickers so that we use a nullish check rather than a falsey check to filter out empty options. This was causing things like numeric 0 to be hidden
- Add a generic utility function for making an async function sequential. This is then used for saving component settings so that we will never get 409s due to multiple updates on the same component, as we will always wait for one update to complete before sending the next one. This can't be used in stores as queued invocations would already contain stale data, but it's perfect for this use case as component prop updates only fetch the full definition to update when they are invoked

Fixes https://github.com/Budibase/budibase/issues/4744.


PClmnt (approved):
This is such a nice change! Those 409s could be really annoying.

shogunpurple (approved):
LGTM!

PR #4764, opened 2022-03-02
Public API IP rate limiting

mike12345567:
## Description
Adding public API rate limiting, with env variable option, defaults to 120 requests per minute. The environment variable `API_RATE_LIMITING` can be set to any number to set the limit of requests per minute. For now this is entirely limited to the public API.

This is achieved with the use of [koa2-ratelimit](https://www.npmjs.com/package/koa2-ratelimit) - this allows rate limiting per IP and use of a Redis store for managing the limits across a cluster of services.


shogunpurple (approved):
LGTM! Really nice that we can just plug this library into existing redis infra. Saw that `koa2-ratelimit` supports other backends too - seems like a nice library 👌

packages/server/package.json
```diff
@@ -126,6 +127,7 @@
     "pouchdb-all-dbs": "1.0.2",
     "pouchdb-find": "^7.2.2",
     "pouchdb-replication-stream": "1.2.9",
+    "redis": "4",
```
Rory-Powell, line 130: in backend-core we use `io-redis` but here we need `redis`. I don't see any usages of the `redis` dependency here, how come it's needed?

packages/server/src/api/routes/public/index.ts
```diff
@@ -8,18 +8,46 @@ import authorized from "../../../middleware/authorized"
 import { paramResource, paramSubResource } from "../../../middleware/resourceId"
 import { CtxFn } from "./utils/Endpoint"
 import mapperMiddleware from "./middleware/mapper"
+import env from "../../../environment"
+// below imports don't have declaration files
 const Router = require("@koa/router")
+const RateLimit = require("koa2-ratelimit").RateLimit
+const Stores = require("koa2-ratelimit").Stores
```
Rory-Powell, line 15: We can remove the duplicate imports by doing
```js
const { RateLimit, Stores } = require("koa2-ratelimit")
```
Or even better would be to use ES syntax
```js
import { RateLimit, Stores } from "koa2-ratelimit"
```
And add the types with 
```
yarn add --dev @types/koa2-ratelimit
```

packages/server/src/api/routes/public/index.ts
```diff
@@ -8,18 +8,46 @@ import authorized from "../../../middleware/authorized"
...
 const {
   PermissionLevels,
   PermissionTypes,
 } = require("@budibase/backend-core/permissions")
+const { getRedisOptions } = require("@budibase/backend-core/redis").utils
 
 const PREFIX = "/api/public/v1"
+const DEFAULT_API_LIMITING = 120
```
Rory-Powell, line 23: Can we make this more explicit to what it represents? i.e. `DEFAULT_API_LIMITING_REQUESTS_PER_MINUTE`
It might make sense to model this as requests per second instead at that is a more useful time key when it comes to api load. Can update the internal unit below to `sec` to do this

packages/server/src/environment.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
   INTERNAL_API_KEY: process.env.INTERNAL_API_KEY,
   MULTI_TENANCY: process.env.MULTI_TENANCY,
   HTTP_MIGRATIONS: process.env.HTTP_MIGRATIONS,
+  API_RATE_LIMITING: process.env.API_RATE_LIMITING,
```
Rory-Powell, line 48: Similar to above, would be nice if we mentioned `API_LIMITING_REQUESTS_PER_MINUTE` or `API_LIMITING_REQUESTS_PER_SECOND` as suggested

packages/server/package.json
mike12345567: Its required for the koa2 redis store, it is based on the Redis client rather than the IORedis client that we use. Weirdly the `koa2-ratelimit` library doesn't include its own store dependencies.

Rory-Powell (approved):

PR #4781, opened 2022-03-03
Replacing mysql2 with mysql library

mike12345567:
## Description
After running into https://github.com/sidorares/node-mysql2/issues/939 had a look at solutions for it, it appears that the recommended solution is to use pooling, however due to the nature of Budibase data sources pooling could cause even more problems. MySQL2 started originally based on on the larger https://github.com/mysqljs/mysql library, which is also the library of choice for use with Knex. The two libraries are supposed to be API compatible, we also have a very minimal use (get a connection, make a request, shut down connection) so I don't really see us running into too many issues with transitioning to it. Tested a few different scenarios without any code changes and everything just seemed to work.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -114,7 +114,7 @@
     "memorystream": "^0.3.1",
     "mongodb": "3.6.3",
     "mssql": "6.2.3",
-    "mysql2": "^2.3.1",
+    "mysql": "^2.18.1",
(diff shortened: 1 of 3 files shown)


mike12345567:
Closing due to the need to support MySQL 8 pluggable auth, only node-mysql2 does this.

PR #4782, opened 2022-03-03
Updating usage of MySQL2 to use promise methods

mike12345567:
## Description
Attempting to fix mysql issue by changing our usage of mysql2 to use the promise version, making sure disconnection always occurs correctly and using a slightly different syntax/approach.

Hopefully this should help with the connection problems, if not then we can try switching over the connect method to use pools, which may cause other problems but should solve our current issue.


shogunpurple (approved):
LGTM! Great job on this fix.

PR #4817, opened 2022-03-07
Fix S3 upload component

aptkingston:
## Description
Somehow the S3 upload component stopped working due to being unable to resolve a sibling API call. I'm not sure what caused this to start failing all of a sudden, but in any case I've updated the API definitions to fix this.

I've also updated the error message so that it won't show `[Object object]` and fixed the loading state which previously would never be unset if errors happen.


shogunpurple (approved):
LGTM!

Issue #4797, opened 2022-03-04
UI error spam


glm-lgauthier:
**Describe the bug**
When we work with the Design part of our app, errors begin to spam in our screen everytime we move, edit, create or delete them.

**To Reproduce**
1. Go to Design
2. Play with UI components, either add, edit, remove
4. Errors spam and we have to close them one by one

**Expected behavior**
Components should be either added, remoned, edited without receiving an error modal. Plus it would be nice if the error could disapear on it's own instead of blocking the whole screen and having to close them one by one.

**Screenshots**
![MicrosoftTeams-image](https://user-images.githubusercontent.com/65235805/156826896-d98a8751-e591-4de6-96be-f566c75cb7f8.png)

**Desktop (please complete the following information):**
 - OS: Pop!_OS 21.10
 - Browser: Brave
 - Version: Cloud based version

**Additional context**
We were a team of 3 persons doing 1 app each and it happened quite often.

aptkingston:
Could you please open your browser dev console (F12) and either screenshot or paste the errors you're getting? The real errors will be logged in there.

aptkingston:
Thanks for the detail @GlitchWitch. That error means either the browser or the server is being slow. I've actually already got a PR merged which will prevent these errors in future:
https://github.com/Budibase/budibase/pull/4750

This will be included in our next release which should be soon.

GlitchWitch:
Thanks @aptkingston, the error happens only on certain apps and regardless of how slowly I drag items. I'll keep you posted if the issue persists after that release.

![Peek 2022-03-07 09-47](https://user-images.githubusercontent.com/32590246/157007315-a3f7312f-18ce-475d-9e82-6cf9547bde5b.gif)


Also further to @glm-lgauthier's point though, I think it would be good to have these errors automatically fade away or disappear after a set time. Having to manually close each one is a bit tedious.

aptkingston:
Thanks for the video @GlitchWitch. That actually looks much more different than I thought. There shouldn't be anywhere near that many requests attempted when just moving one component. I'll look in to this because that must be something else.

As for dismissing errors - that was the way they used to behave, but we deliberately removed that because it meant important errors were disappearing before they could be read or copied and pasted. If we fix the cause of this spam then it won't be a problem again.

PR #4820, opened 2022-03-07
Properly remove window event listener when unmounting component preview

aptkingston:
## Description
This PR fixes an issue reported in https://github.com/Budibase/budibase/issues/4797#issuecomment-1060402155. Thanks to @GlitchWitch for providing a video which helped me diagnose the issue.

We weren't properly removing an event listener when unmounting the component preview, which mean that if you toggled between the data and design section multiple times, new listeners would be created. This meant that whenever an even was sent from the iframe, it was handled multiple times. This causes lots of conflicts, and may have been the source of multiple other bugs as well.


shogunpurple (approved):
LGTM!

Issue #3825, opened 2021-12-17
Display component type in setting panel


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3369

<div type='discussions-op-text'>

<sup>Originally posted by **itsamemarkus** November 14, 2021</sup>
Maybe I'm missing something or you expect a perfect naming convention :) , but for me it's really hard to figure out which components I'm interacting with when I have a lot of screen elements.

![image](https://user-images.githubusercontent.com/34944988/141676238-51f128e5-b1e2-44d5-be42-84fa510af17c.png)

Afaik the only way to be sure which component I'm using is by trying to delete it, so it shows what it is without inspecting the component options.
![image](https://user-images.githubusercontent.com/34944988/141676261-ce609f21-45a2-4f90-a7d1-34a08fa9ae33.png)

Suggestion: Individual symbol for each component.

If not possible / too much clutter: Info which component I'm interacting with in the options menu.

![image](https://user-images.githubusercontent.com/34944988/141676375-46adb6ca-3b33-4fbe-aa03-24b703522d75.png)
</div>

Issue #3818, opened 2021-12-17
Duplicate a Screen + Improve copy and paste


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3768

<div type='discussions-op-text'>

<sup>Originally posted by **lxnunes** December 12, 2021</sup>
Hi!

I found myself creating several versions of the same screen, sometimes to be used as a regular edit screen, other times as a modal or even as a read-only screen. 

I know most (if not all) of these aspects can be controlled using Conditions, but sometimes it would be easier to just duplicate a screen for testing - just like duplicating screen objects. 

Is there a way to achieve this?

Best</div>

- Ask the user to enter the URL of the new screen when you copy it.
- Preserve data bindings by generating new IDs on the new screen

gvlx:
Copy/Paste/Duplicate a screen or layout would be useful.

ggs331:
Copy/Paste/Duplicate a screen or layout currently works, but cannot keep the bindings. It will be really helpful if copy/past/duplicate can keep bindings, then we can easily to reuse the components.

PR #4821, opened 2022-03-07
Copy paste improvements + extras

aptkingston:
## Description

Edit: a few more QOL improvements:
- Added the option to paste inside a screen (previously you couldn't do this, and had to add a component first)
image
- Added a duplicate icon to the inline settings bar
image

This PR adds a few QOL improvements to working with screens and components:
- Added the ability to duplicate screens
- Copying and pasting or duplicating components (and now screens) will keep all data bindings functional (including JS)
- Added the ability to delete all screens underneath a route folder
- Added the selected component's type to the settings panel
- Adjust table height calculation slightly to avoid scrolling when not needed
- Refactored some of the new screen modal logic to make it cleaner, fix some rare bugs and be reusable (for example for duplicating screens)

Addresses #3818 and #3825.

## Working data bindings for copying/pasting/duplicating components
When you copy and paste components that use data bindings, everything will now work as you would expect:

How this actually works:
- Recurse the component tree, and change the ID of every component we enounter to a new UUID
- Search the entire tree for any references to the old ID inside HBS data bindings and replace them with the new UUID
- Search the entire tree for any JS bindings, decode them, find any references to the old UUID, replace them and re-encode the JS binding

Here's an example of duplicating a details form with a data provider, repeater and form:
image
image

## Duplicating screens
There is a new "Duplicate" option in a screen context menu:
image

Clicking this will bring up a modal asking for the new name and URL for the duplicated screen. The values are prefilled with the current screen details, so you can quickly edit them to something slightly different rather than starting from scratch:
image

You'll then have a fully functional dupe of your screen with all data bindings working:
image

## Delete all screens under a route
There is now a context menu for route folders, and an option to delete all screens under that route:
image

This will bring up an informative confirmation dialogue detailing exactly what will be deleted:
image

## Selected component type
The type of the selected component is now displayed in the settings panel, under the section that was previously called "General":
image


mike12345567:
Such an exciting feature to add, massively improves the usefulness of copy paste!

joebudi:
Wow, I never knew this was coming! Brilliant. @aptkingston **please** can we have the duplicate function in the component action bar!

shogunpurple (approved):
This is killer - amazing work 👌highly requested as well!

PR #4824, opened 2022-03-07
Export selected rows from button action

PClmnt:
## Description
Previously users had no way to export data from their application. 

This PR allows the user to create a button action, select a table that they would to export rows from and export those rows when their created button is clicked. 

This works for both internal and external tables. 

n.b I had to slightly modify the row selection store and client table component to store the Table ID, for use in the exportRows API call. The table ID is needed for output processing etc. I've therefore updated dataBinding.js to take account of this, it still will only output the selected row IDs
image

image


packages/client/src/components/app/table/Table.svelte
```diff
@@ -42,6 +42,7 @@
   $: {
     rowSelectionStore.actions.updateSelection(
       $component.id,
+      selectedRows.length ? selectedRows[0].tableId : "",
```
PClmnt, line 45: Setting the table id here, this seems to be fine because the store gets reset when the data provider above it gets changed. Still unsure of why this is the case, but it does.

shogunpurple (approved):
LGTM! Nice work 🚀 Also really cool the way you can do this directly through the budibase API now.

aptkingston (commented):
Nice one Peter! This looks good and will be really useful :+1: 

The only issue I can see is around configuring the action. By fetching selected rows by table ID, having 2 table components referencing the same table will now overwrite each other.  I think it's also a bit confusing that this isn't linked to the table components (where you actually do the row selection).

I would suggest:
- Update the button action so that you select a table component, not a backend table
- Don't store `tableId` in the selection store, just fetch the selected rows by table component ID
- When executing the action. determine the `tableId` by looking at the selected rows
- Update the button action to be more clear that it's linked to table components, and not just a generic data export action

PR #4826, opened 2022-03-07
Minor OpenAPI spec change

mike12345567:
## Description
I was playing around with the spec in Readme, noted there was no easy way to set the tenant ID/org ID when dealing with the Cloud, added it as a variable to the server spec. Also updated the local server spec so that the protocol and host can be set as variables as well - this is important for being able to make calls directly from Readme/copy and pasting the cURL requests directly.

Looks like this in Readme:
image

### DIFF
--- a/packages/server/specs/generate.js
+++ b/packages/server/specs/generate.js
@@ -17,14 +17,8 @@ const options = {
     },
     servers: [
       {
-        url: "https://{organisationId}.budibase.app/api/public/v1",
+        url: "https://budibase.app/api/public/v1",
         description: "Budibase Cloud API",
-        variables: {
-          organisationId: {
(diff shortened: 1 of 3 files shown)


Rory-Powell:
Do we need the tenant in the url at all for public api? It should always be derived from the api key. Is this for cosmetic purposes?

mike12345567:
> Do we need the tenant in the url at all for public api? It should always be derived from the api key. Is this for cosmetic purposes?

I was thinking for the future, incase we ever decide to support any public API endpoints, like allowing user login it might be good to support this - however we could just say that we aren't going to support any public access with the public API.

Rory-Powell (approved):

Issue #4811, opened 2022-03-07
Renaming a MySQL/Postgres column clears the data from that column.


mslourens:
**Describe the bug**
When changing the name of a column in MySQL or Postgres, the data from that column gets cleared. 

**To Reproduce**
1. Go to the Data tab
2. Create a new MySQL or Postgres data connection
3. Click on a table inside the created data connection
4. Enter some data (if not already present)
5. Change the name of one the columns (which contain data)
6. Notice: the data from the renamed column disappears
7. Refresh the page to retrieve the data again
8. Notice: the data is still gone.

**Expected behavior**
The column should be renamed without changing the content of the column

**Screenshots**

https://user-images.githubusercontent.com/1907152/156993348-7c7fc75d-e591-42f0-a42e-b58e7c46ae09.mov

**Desktop (please complete the following information):**
 - OS: MacOS 12.2.1
 - Browser: Vivaldi
 - Version: develop

**Additional context**
If this is somehow expected behavior, we should add a warning dialog saying that renaming a column will delete all data from that column.

mike12345567:
Resolved as of #4834.

PR #4834, opened 2022-03-07
Renaming SQL columns - don't drop old column

mike12345567:
## Description
Fixing issue #4811 - when renaming SQL columns use an actual rename alteration rather than dropping the column.


shogunpurple (approved):
LGTM!

PR #4838, opened 2022-03-08
show correct error and keep modal open

### LINKED ISSUE #4812: Data source modal closes on error
**Describe the bug**
When adding a new data source with incorrect data (database does not exist or user credentials are incorrect), the data source modal closes with a very generic error message: `Error saving data source`. This message is not completely correct: the data source can be saved, but the fetch tables query cannot be executed. 

**To Reproduce**
1. Go to the Data tab
2. Click on the '+' button
3. Enter incorrect connection details for Postgres (a non-existing table, for example)
4. Click the `save and fetch tables button`
5. The modal closes and a generic error is displayed: `Error saving the data source`

**Expected behavior**
The error should be shown, but it should say: `database x does not exist`, because this is the actual error Postgres is throwing.  The modal should remain open when an error is thrown, so the user can fix the incorrect details.

 - OS: MacOS 12.2.1
 - Browser: Vivaldi
 - Version: develop branch

mslourens:
## Description
When the connection to a data source fails, a very generic error is thrown and the modal is closed. This should not happen, the user should be able to fix the issue and know what's wrong. This PR fixes #4812 

## Screenshots
image

### DIFF
--- a/packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
+++ b/packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
@@ -20,7 +20,9 @@
       $goto(`./datasource/${resp._id}`)
       notifications.success(`Datasource updated successfully.`)
     } catch (err) {
-      notifications.error("Error saving datasource")
+      notifications.error(err?.message ?? "Error saving datasource")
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
Nice fix here! Thanks @mslourens

PR #4842, opened 2022-03-08
Update how button action for exporting works

PClmnt:
## Description
Updates how the button action works for the export data action. 

- It now takes a table component id, instead of a backend table ID. 
- Also update some of the UX for the button action, to make it a little tidier. 

image


aptkingston (approved):
LGTM!

shogunpurple (approved):
LGTM!

PR #4852, opened 2022-03-08
Quick fixes for OpenAPI search response schema

mike12345567:
## Description
As I've been working on the NextJS examples I noticed that there wasn't much of a definition available as a reference for the search schemas of the resources. This simply makes the search responses available as components that can be referenced, which importantly means that they can easily be used through our Typescript definitions that are generated from the OpenAPI spec - which makes the examples a bit more interesting.


Rory-Powell (approved):
LGTM 👍

Issue #4401, opened 2022-02-09
Google Firebase Integration


shogunpurple:

mesqueeb:
@shogunpurple can we re-open this ? : )

PR #4864, opened 2022-03-09
Google firebase integration

mslourens:
## Description
Implements #4401 - Firestore integration. 
This is my first attempt to create an integration with an external data source, so any tips for improvements are welcome.

Some small comments:

- authentication is done using a service account. You can either enter the used details separately or copy the whole json string and enter that in the serviceAccount input. I tried using a CodeMirrorEditor, which is easier to read the JSON file, but for some reason that didn't work.
- I am not really comfortable with bindings yet, so I used the MongoDb integration as an example and got it working somehow. 
- For some reason I get an error after creating/updating/removing a document, when the `buttonAction.js` is checking if the query is readable, and then it invalidates the datasource, which in turn will call `fetchTableDefinition`. This call fails for the Firestore datasource. This `readable` property is set for all read queries in `enrichQueries`. Should that not happen for Firebase/MongoDb queries (because they don't have tables, but collections of documents) or should the method `fetchTableDefinition` return the collection?

### Config
image

### Selection
image


packages/server/src/api/routes/public/index.ts
```diff
@@ -29,7 +29,7 @@ function getApiLimitPerSecond(): number {
   return parseInt(env.API_REQ_LIMIT_PER_SEC)
 }
 
-if (!env.isTest()) {
+/*if (!env.isTest()) {
```
mike12345567, line 32: Noticed this commented out, this will need added back in, not sure if you're still experiencing issues with this, might be worth trying pulling the latest stuff and seeing if its still not connecting.

mike12345567 (changes requested):
Hi @mslourens - this is an amazing piece of work, thank you for it, I'm sure many people in the community will be very happy to see this integration being introduced!

The only thing that needs fixed before this can be merged is the commented out lines around the Koa rate limiting connecting to Redis.

Generally we would also write up a basic test suite for new integrations as well, like we have in `server/src/integrations/tests/postgres.spec.js` - we would mock out the actual connection to the integration and then just check that everything on our side works as expected; however happy to get this merged and we can add tests ourselves/if you have any time could add them as a separate PR.

packages/server/src/api/routes/public/index.ts
mslourens: I am still experiencing issues with this part of the code, but I shouldn't have committed it. I reverted the change.

mike12345567 (approved):
Thanks for the changes @mslourens - approved!

PR #4871, opened 2022-03-10
Date filtering logic

aptkingston:
## Description
This PR attempts to fix date filtering by updating how SQL queries are generated from `range` lucene filters.

## The current problem
Our current syntax for performing a search using the `less than` or `more than` operators generates a SQL `between` statement, because that's how the lucene search structure is formatted. For example, if we want to search for `less than 10`, that will actually send up a structure of:
```js
[field]: {
  low: Number.MIN_SAFE_INTEGER,
  high: 10
}
```
As we use a between statement. we need to fill in the other limit of the range. For numbers there are the min and max safe integers which are generally also valid integers for most SQL databases, as long as a small bit range integer isn't used.

For dates the frontend sends up 2 magic strings to represent the lowest and highest possible date range: `0000-00-00T00:00:00.000Z` and `9999-00-00T00:00:00.000Z`. While generating a SQL statement from filters, the backend needs to parse these date strings into real JS dates, and there are edge cases hardcoded in to transform these date strings into the actualy minimum and maximum JS dates. These are constructed using `new Date(-8640000000000000)` and `new Date(8640000000000000)` respectively. Looking at the minimum date in particular, this date represents `Tuesday, April 20th, 271,821 BCE`, and therefore this is the date that will be inserted into a SQL query.

For example, doing a filter for `less than 10th March 2022` would generate SQL of:
```sql
...
between "-271821-04-20T00:00:00.000Z" and "2022-03-10T00:00:00.000Z"
...
```

The problem with this is that certain SQL databases and data types consider that date invalid. Using Postgres and a `date` type for example, will throw an error saying:
```
error: date out of range: "271822-04-19T23:58:45.000-00:01 BC"
```

Taken from the Postgres docs, the earliest supported dates are 4713 BCE:
image

I'm sure other databases have different limits and it would be a nightmare to support them all.

## The solution
The best way to deal with these different limits is to not have to worry about them at all. If the user only wants to filter `less than 10th March 2022` then we should simply query for exactly that instead of using a `between` statement and needing a limit.

This PR updates the between logic so that if only one operand is given, it will generate a `where` clause with a single operand rather than a `between` clause with two. This should hopefully then support any SQL database.


mike12345567 (approved):
LGTM! Good idea, why bother trying to use a between when the user hasn't specified a full range!

shogunpurple (approved):
Great work - super simple and clean solution. Also good to know that there's a limit in postgres if I ever want to work with dates years before modern life as we know it existed. 😂

PR #4872, opened 2022-03-10
Add css print styles to ensure multi-page print is possible

### LINKED ISSUE #4870: multi-page tables not printable
**Describe the bug**
When having a table with many entries (that do not fit on a single page), they cannot be printed as **no new page is created**.
In print view (`ctrl+p`), budibase seems only to be able to render a single page. The page is not wrapped.
This seems also to be true for other design elements, not only tables.

In general, the print view is not very optimized.
It can be worked around a lot by, e.g. hiding elements like the headers via a "print view button".
But not wrapping/creating any new pages is a major issue here.

**To Reproduce**
1. create a simple page with a table
2. fill table with enough rows to not fit on a single page

**Expected behavior**
Budibase should be able to render multiple pages in print view and wrap pages.

 - OS: Linux
 - Browser: chromium/firefox
 - Version: 97/96

**Additional context**
Tested with latest docker containers `v1.0.82`

aptkingston:
## Description
This PR adds `print` media queries to ensure the app looks good when printed. It ensures content spans multiple pages when required and also removes the nav bar.

Unfortunately as our table uses a grid layout we cannot control the page break behaviour, which means that sometimes breaks might occur in the middle of rows. But apart from this edge case things work well.

Fixes https://github.com/Budibase/budibase/issues/4870.

### DIFF
--- a/packages/client/src/components/ClientApp.svelte
+++ b/packages/client/src/components/ClientApp.svelte
@@ -227,6 +227,7 @@
     border-radius: 4px;
   }
 
+  /* Print styles */
   @media print {
     #spectrum-root,
     #clip-root,
--- a/packages/client/src/components/ClientApp.svelte
+++ b/packages/client/src/components/ClientApp.svelte
@@ -226,4 +226,12 @@
     border: 1px solid var(--spectrum-global-color-gray-300);
     border-radius: 4px;
   }
+
+  @media print {
+    #spectrum-root,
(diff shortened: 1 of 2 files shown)


yeoldegrove:
@aptkingston When will this land in `develop` and be released?

shogunpurple (approved):
LGTM!

PR #4873, opened 2022-03-10
NextJS public API example

mike12345567:
## Description
This adds the first example of external integration with Budibase, using Next.JS to build a simple sales app, utilising Budibase as a backend for the project.

There is a walkthrough which has been written up which is attached to this example, it will be made part of the Budibase documentation.

The example includes a Postgres database that generates some sales data, with relationships, the walkthrough then guides users through setting up an app, connecting to the Postgres database then setting up the Next.JS example and running through it a bit. It also describes the integration between the app and Budibase through the API, as well as the typescript definitions used.

## Screenshots
The list of sales in the Next.js app:
image

The create a new sale page:
image

### App export
If you are running in dev and want to try this out, simply startup the postgres database from the `examples/nextjs-api-sales/db` directory using docker-compose, then import this app and call it `sales`. Lastly all you need to do is get your API key and update the `examples/nextjs-api-sales/next.config.js` with your API key.
[sales-export-1646916536517.txt](https://github.com/Budibase/budibase/files/8223467/sales-export-1646916536517.txt)


joebudi:

PR #4880, opened 2022-03-10
Quick fixes for Public API and spec

mike12345567:
## Description
Making the URL optional for application creation/update in OpenAPI spec, removing unused parameter and fixing `getRedisOptions` function to work with fully qualified URLs.


Rory-Powell (approved):
LGTM 👍

PR #4888, opened 2022-03-11
Fixing SQL constant character usage in queries with bindings

mike12345567:
## Description
After our change to use vars in queries, there was some issues with them being used within character constants, e.g. `'{{binding}}'` would become `'$1'` which is not valid SQL syntax - the variable will break and throw an error.

 I've had to account for a few scenarios here because not only could you do something like `'{{binding}}'` but you could also do something like `'hello there {{binding}} welcome!'` which needs to be converted into some form of concat statement.

The way the concat is handled is dialect specific, so I had to add new methods to all the datasource plus implementations.

For most it will become either `concat('hello there ', $1,  'welcome!')` or `'hello there ' || $1 || ' welcome!'`. I've tested this across all of the differnet SQL datasource plus databases that we currently support, with a variety of use cases and haven't run into any issues.


shogunpurple (approved):
LGTM! Great job finding this one.

PR #4911, opened 2022-03-15
fix button action export for csv

PClmnt:
## Description
CSV data wasn't being exported correctly, so added the csv headers and converted the data


aptkingston (commented):
This looks like it'll work just fine, but I think it would be better to do this on the server side like we do our other exports - the server already has the utilities needed. Check out the view controller for an example.

aptkingston (approved):
Nice one, definitely better doing it on the back end :+1:

Issue #4914, opened 2022-03-15
API Response for Missing App Id Header not ideal


melohagan:
**Describe the bug**
Whenever the **x-budibase-app-id** header is missing / incorrect, a _500_ is returned with message: _Missing/invalid DB name_

**To Reproduce**
1. Omit x-budibase-app-id header
2. Post request to _/tables/search_
3. See generic 500

**Expected behavior**
Should probably return a 400 - Invalid app id header

**Screenshots**
<img width="449" alt="Screenshot 2022-03-15 at 15 54 23" src="https://user-images.githubusercontent.com/101575380/158418429-f91f9502-376e-4588-b5fd-dfb1c0f3f450.png">

PR #4918, opened 2022-03-15
Fix for public API header error warnings

mike12345567:
## Description
Quick fix for #4914 - adding some checks in API middleware to confirm headers have been set correctly.


Rory-Powell (approved):
Looks good!

PR #4925, opened 2022-03-16
Button action inputs + map component actions

aptkingston:
Base branch is currently set to `feature/map-component` but will update to `develop` when that gets merged.

## Description
This PR adds a new concept of input parameters for button actions, which are passed in by the event trigger. An example of what this means, which is included in this PR, are the event actions for interacting with the map component.

The scope of this new concept goes beyond the map component and could be useful for lots of different applications. Couple of examples of things that are possible now:
- On click event for a row in a table (table row is provided as a binding)
- On change event for form fields (the value is provided as a binding)

The map component has two different actions - when you click a marker, and when you create a marker. Both of these actions required some input context:
- When clicking a marker, you need to know which marker was clicked
- When creating a new marker, you need to know the desired latitude and longitude of the new marker so you can use them

These new inputs are easily defined in the component manifest, under the `context` key for any `event` settings. Here's the excerpt for the manifest for the "clicking on a marker" action:
```json
{
  "type": "event",
  "label": "On Click Marker",
  "key": "onClickMarker",
  "context": [
    {
      "label": "Clicked marker",
      "key": "marker"
    }
  ]
}
```
Data bindings are generated for any fields specified here, and are listed top level whenever using the binding panel inside the button actions drawer. For example, if I wanted to navigate to a details page for the marker I clicked then this is the flow:
image
image
image

In this example I have manually added the `_id` field, as the data binding represents the full object which could have any structure.

Components simply need to invoke any button action settings with the values specified in the manifest, and the client library will take care of the rest.

## New map actions
As shown above, there is a new button action setting for "on click marker" which is fired on a single left click of a marker. This event gets passed in the full object that was clicked.

The other new setting is the ability to create markers. There is initially a checkbox to control where creating new markers is possible at all. If it is enabled then the button action setting appears:
image

## Creating new map markers
When creating new markers has been enabled, left clicking on the map will now begin the process of creating a new marker. This will place a candidate new marker on the map, highlighted in the theme primary color to be different from the red normal markers. It also has a permanent tooltip indicating that it is the new marker:
image

Continuing to click on the map will move around the location of the new marker. Clicking directly on the new marker will remove it.

When a candidate new marker is positioned on the map, 2 new buttons will be displayed below the map. These are a cancel button which will remove the new marker, and a "create marker" button which will invoke the button action setting. Typically this setting will be linked up to a "save row" or external query action.

When the "create marker" button is pressed, the map will await the successful execution of the button actions before removing the blue marker and replacing it with a normal red one. If something goes wrong then the candidate blue marker will remain where it was, but if everything works then the new mark state is reset and the new marker will be shown on the map.
image

The input values provided to the "on create marker" setting are the desired latitude and longitude. Here's the "save row" action I was using in this example to create the markers, making use of the new bindings:
image


shogunpurple (approved):
LGTM! Great work adding these new actions - the map is going to be super powerful with these.

PR #4930, opened 2022-03-16
don't use _id for finding selected rows

### LINKED ISSUE #4895: Selecting a row in a table with Allow row selection enabled using a REST API data source appears to select all rows
**Describe the bug**
Selecting a row in a table with Allow row selection enabled using a REST API data source appears to select all rows. You also cannot parse the row for data as it is undefined.

**To Reproduce**
1. Create a REST API data source
2. Add a query
3. The query should be set a GET `https://gorest.co.in/public/v2/users` (This is a public REST API with dummy data)
4. Send query
5. Save query
6. Create a Screen
7. Add the REST API data source as a data provider component on the screen
8. Create a table component underneath the REST API data provider component
9. You should see data appear in the table
10. Press the play icon at the top right of the page
11. Navigate to the screen you created
12. Click on a table row
13. All table rows are now selected while saying 1 row selected

**Expected behaviour**
Each table row is able to be selected.

**Screenshots**
REST API query
image
Screen layout
image
Table setting
image
Table before selecting a row
image
Table after selecting a row
image

**Additional context**
BB version 1.0.82
This only happens with REST API data sources.

mslourens:
## Description
FIxes #4895

Table.svelte uses `_id` property to check whether a row is selected or not. This property is not present when using data sources other than the internal database. Therefore I now check for the whole object instead of just `_id`. 

## Screenshots

### DIFF
--- a/packages/bbui/src/Table/Table.svelte
+++ b/packages/bbui/src/Table/Table.svelte
@@ -71,9 +71,7 @@
 
   // Deselect the "select all" checkbox when the user navigates to a new page
   $: {
-    let checkRowCount = rows.filter(o1 =>
-      selectedRows.some(o2 => o1._id === o2._id)
-    )
+    let checkRowCount = rows.filter(o1 => selectedRows.some(o2 => o1 === o2))
     if (checkRowCount.length === 0) {
       checkboxStatus = false
     }
@@ -209,10 +207,8 @@
     if (!allowSelectRows) {
       return
     }
-    if (selectedRows.some(selectedRow => selectedRow._id === row._id)) {
-      selectedRows = selectedRows.filter(
-        selectedRow => selectedRow._id !== row._id
-      )
+    if (selectedRows.some(selectedRow => selectedRow === row)) {
+      selectedRows = selectedRows.filter(selectedRow => selectedRow !== row)
     } else {
       selectedRows = [...selectedRows, row]
     }
@@ -223,15 +219,13 @@
     if (select) {
       // Add any rows which are not already in selected rows
       rows.forEach(row => {
-        if (selectedRows.findIndex(x => x._id === row._id) === -1) {
+        if (selectedRows.findIndex(x => x === row) === -1) {
           selectedRows.push(row)
         }
       })
     } else {
       // Remove any rows from selected rows that are in the current data set
-      selectedRows = selectedRows.filter(el =>
-        rows.every(f => f._id !== el._id)
-      )
+      selectedRows = selectedRows.filter(el => rows.every(f => f !== el))
     }
   }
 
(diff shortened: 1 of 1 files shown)


aptkingston (commented):
Will pull this down and test it out. The only potential issue I can see is that if a data provider was refreshed - even if it received the exact same data again - the actual objects will be different, which would mark everything as unselected.

There's no easy way around this though, without doing something like hashing the entire row and make our own unique property.

I'll check this out and see what happens!

aptkingston:
I pulled this down and tried it out, but it seems like this doesn't match selected columns at all for me. The selected row count goes up, and the objects are added to the `selectedRows` array, but the obsolute equality check seems to fail and the checkbox never actually gets selected. I can also keep adding the same row then because it doesn't realise it was already selected. Did this work for you? Just wondering if somehow other changes in develop have affected it.

mslourens:
yes, it does work when I check out this branch:

image

aptkingston:
Ah I was testing this in a table inside an app, where it doesn't work as we again use the `_id` field again there. This does work like you said in the builder UI though, like the data section.

This works fine in the data section as every time the data changes (for example if you change the sort order) we reset the selected rows. If we didn't reset the selected rows, this would not work as the object references would change, which would stop the object equality checks from working as the objects themselves have changed.

In client apps though, we do need to persist the selection between multiple data updates so we can't use this approach unfortunately. The 2 options are either:
- Stringify the entire row and use that string
- Disable row selection for non-plus datasources. So row selection would only work for internal tables and SQL tables, all of which have an `_id`. I'm actually in favour of this approach as you cannot use the "export rows" action for non-plus datasources anyway.

Was just thinking out loud there - I don't mean you have to implement that haha.

hjbout:
Same issue with a query based on MariaDB tables. Renaming my existing column "id" to "_id" does help. Is this workaround save to use?

aptkingston:
> Same issue with a query based on MariaDB tables. Renaming my existing column "id" to "_id" does help. Is this workaround save to use?

Hey @hjbout. Renaming your current ID column to `_id` will let you select rows in the table, but you won't be able to do anything with those rows since they come from a custom query - so you won't be able to use the "Export rows" button action. This is why I'm in favour of disabling row selection for datasources that won't support it (the ones which support it are internal tables and most SQL databases - MySQL, Postgres, SQL server etc).

aptkingston:
Since row selection (and export from client apps) depends heavily on an `_id` property, we've updated tables so that you won't be able to use row selection for custom queries or other datasources like REST. This solves this issue while also allowing us to continue to use `_id` for the sources that support it.

I'm going to close this one off as it's solved by those changes which are already deployed.

If there is a large demand for being able to use row selection for custom queries then we can reevaluate how to do it and implement a proper change at that stage.

PR #4950, opened 2022-03-18
fetch the table which contains info about the primaryDisplay

### LINKED ISSUE #4910: Display column validation not enforced
This applies to both SQL and internal tables.

Display columns are always "Required" - this is not editable.

However, this is not actually enforced when using an app: https://github.com/Budibase/budibase/discussions/4843#discussioncomment-2363629

Workaround is to manually add validation on the form itself.

mslourens:
## Description
All code for using the primaryDisplay column as a required validation constraint was already present in `createValidatorFromConstraints`, but it needed a table object, which was never passed, so when I retrieved the table and passed it to the `createValidatorFromConstraints` method, it added a required constraint. 
I am not sure if I used the best method to retrieve the table definition. I copied it from the `RelationShipField.svelte` component.

Fixes #4910

### DIFF
--- a/packages/client/src/components/app/forms/Form.svelte
+++ b/packages/client/src/components/app/forms/Form.svelte
@@ -17,6 +17,7 @@
   let table
 
   $: fetchSchema(dataSource)
+  $: fetchTable(dataSource)
 
   // Returns the closes data context which isn't a built in context
   const getInitialValues = (type, dataSource, context) => {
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
LGTM! thanks @mslourens

aptkingston (approved):
LGTM! I'm going to open another PR to fix a bug where we don't  actually attach the required constraint if the column if the first one in the table, which is why we even need this frontend validation in the first place!

This is a good catch though and should be fixed, as this extra frontend validation will help validate tables even if the contraints aren't properly set up.

Basically if a column is the `primaryDisplay`, then
```js
field.constraints.presence = { allowEmpty: false }
```
should also always be set - but we don't always do this currently.

Thankfully the form validator manually checks for `primaryDisplay` to get around this limitation, but that wasn't working due to this bug that you've just fixed.

Thanks!

PR #4953, opened 2022-03-18
don't sent searchableSchema to getSchemaForDatasource

### LINKED ISSUE #4948: Columns not appearing in Define filters
**Describe the bug**
Unable to select columns from datasource schema for Define filters.

**To Reproduce**
1. Add REST datasource with columns present in the schema
2. Add Table Block
3. Click Define filters. Column options are not present

**Expected behavior**
Should be able to select columns from the datasource schema for filtering.

**Screenshots**
filter_missing

mslourens:
## Description
the rest schema doesn't have query parameters which define the schema, so don't sent `searchableSchema` option.

Fixes #4948


aptkingston (approved):
LGTM!

PR #4976, opened 2022-03-21
Handle error when running a a Query Rows automation with a deleted column

PClmnt:
## Description
Running a Query Rows automation that had a deleted column as it's sort was causing repeated errors and affecting performance.


shogunpurple (approved):
LGTM!

PR #4986, opened 2022-03-21
Remove automation usage increment

Rory-Powell:
## Description
- Fix an issue where `getTenantId` was failing because of the quota usage update not being wrapped within the `doInTenant` statement
- Remove the increment as total automation will no longer be tracked, this is changing to a monthly quota


shogunpurple (approved):
LGTM! Great job fixing.

Issue #4978, opened 2022-03-21
SQL query with bindings for `IN` values no longer work


mjashanks:
A change in the way SQL parameters work has made the SQL `... WHERE something IN {{ values }}` impossible.

Original poster reported in MySQL, and I reproduced in Postgres

### Discussed in https://github.com/Budibase/budibase/discussions/4963

<div type='discussions-op-text'>

<sup>Originally posted by **micebrain** March 19, 2022</sup>
Hi,
I updated my app to v1.0.90 and found that SQL queries that has value binding to integer column is giving error:
```
error: invalid input syntax for type integer: "1,2"
```
I figured out that this only happened to queries that uses the `in` expression with multiple values. And example of the query is something like

```
SELECT * FROM customer WHERE id in ({{ customer_ids }});
```
with the `customer_ids = 1,2`

My expectation of the template output would be 
```
SELECT * FROM customer WHERE in (1,2);
```
But the error message seems to indicate that it is now rendering as
```
SELECT * FROM customer WHERE in ('1,2');
```

How can I change this behaviour?</div>

micebrain:
Hi, 
I updated to the latest version v1.0.99, however this it didn't seems to fix the issue.
I still get the same error in my test.

<img width="381" alt="Screenshot 2022-04-01 at 20 34 18" src="https://user-images.githubusercontent.com/7136209/161322737-c362025b-3d7e-4574-9596-fb9ec65ad358.png">

mike12345567:
Hi @micebrain - could you provide an example of your query to test?

micebrain:
Hi @mike12345567 sure, it's actually pretty simple. Here is the screenshot of the configured query

<img width="617" alt="Screenshot 2022-04-03 at 08 18 46" src="https://user-images.githubusercontent.com/7136209/161414607-1d7a969a-fb8d-46d0-b143-46ac26864cb9.png">

mike12345567:
Hi @micebrain - can you try this query without the parentheses around the binding e.g. `WHERE id in {{ customer_ids }};` - we've changed how Budibase manages these kind of statements and it will now automate the inclusion of the braces. 

I will see if Budibase can detect that they've already been included in the query and not add them if they are already there - but I think removing them should fix your query.

micebrain:
Hi @mike12345567 
Yes that solved the problem. Thank you!

PR #4991, opened 2022-03-21
Fixing SQL in queries

mike12345567:
## Description
Fixing issue #4978 - fixing an issue with using the keyword 'in' as part of an SQL query.

The fix for this has to make some changes to the SQL query itself, switching to a `where id in ($1, $2, $3)` syntax, this utilises some of the work that was put in place around string concatenation to find instances that need swapped. Currently this will look for the keyword `in` (any casing) before a binding, if found it will switch it to use this syntax. This is also dependent on the inserted value being a string of comma separated values that were to be used as in parameters, however this should always be true as that is the format that would have been used if the bindings were directly interpolated into the SQL query itself.

I've tested this with multiple in queries as well, as well as using in for a single string value (no comma separation). If there are no commas it will just generate a syntax like `in ($1)` where the string value is the only one used.

I've also added a `SQL_MAX_ROWS` env variable, so that self hosters can set the max number of rows returned by SQL queries (greater than 5000) if they desire to export large SQL tables.

### DIFF
--- a/packages/server/src/threads/query.js
+++ b/packages/server/src/threads/query.js
@@ -37,7 +37,7 @@ class QueryRunner {
       arrays = []
     for (let binding of bindings) {
       // look for array/list operations in the SQL statement, which will need handled later
-      const listRegex = new RegExp(`(in|IN|In|iN) ${binding}`)
+      const listRegex = new RegExp(`(in|IN|In|iN)( )+${binding}`)
       const listRegexMatch = sql.match(listRegex)
       // check if the variable was used as part of a string concat e.g. 'Hello {{binding}}'
       const charConstRegex = new RegExp(`'[^']*${binding}[^']*'`)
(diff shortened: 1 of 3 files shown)


packages/server/src/threads/query.js
```diff
@@ -33,10 +33,12 @@ class QueryRunner {
...
     }
     const bindings = findHBSBlocks(sql)
-    let variables = []
+    let variables = [],
+      arrays = []
     for (let binding of bindings) {
-      let variable = integration.getBindingIdentifier()
-      variables.push(binding)
+      // look for array/list operations in the SQL statement, which will need handled later
+      const listRegex = new RegExp(`(in|IN|In|iN) ${binding}`)
```
shogunpurple, line 40: NAB: Might be worth doing `(in|IN|In|iN)\s+${binding}` here - as extra spaces are valid SQL as well.

shogunpurple (approved):
LGTM - great job getting this sorted so quickly.

packages/server/src/threads/query.js
mike12345567: Good catch, didn't think of that!

PR #5001, opened 2022-03-22
Licensing

Rory-Powell:
## Description
Integrate with pro bundle 
- Licensing middleare
- Usage quotas
	- published apps
	- apps
	- rows
	- automations
	- queries

API
- Add license key api
- Add quota usage api 

UI
- License key activate page

Misc
- Add error handling framework
- Add feature flagging lib based on environment variables
- Various bbui updates to support styling
- Refactor touched files to TS
- Split tsconfig for better editor test support

See:
https://github.com/Budibase/account-portal/pull/107
https://github.com/Budibase/budibase-pro/pull/12

## Screenshots

Self host pre activation
Screenshot 2022-03-22 at 12 27 28

Self host activated
Screenshot 2022-03-22 at 12 30 31


charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -98,10 +98,6 @@ spec:
           value: http://worker-service:{{ .Values.services.worker.port }}
         - name: PLATFORM_URL
           value: {{ .Values.globals.platformUrl | quote }}
-        - name: USE_QUOTAS
```
shogunpurple, line 101: Don't forget to bump the version in `Chart.yaml` to cut a new release of the helm chart when this goes out 👍

packages/backend-core/src/errors/base.js
```diff
@@ -0,0 +1,11 @@
+class BudibaseError extends Error {
```
shogunpurple, line 1: 👌 we could do a lot with this - grafana, sentry etc

packages/server/src/api/controllers/application.ts
```diff
@@ -280,13 +288,41 @@ exports.create = async ctx => {
...
+const appPostCreate = async (ctx: any, appId: string) => {
+  // app import & template creation
+  if (ctx.request.body.useTemplate === "true") {
+    const rows = await getUniqueRows([appId])
+    const rowCount = rows ? rows.length : 0
+    if (rowCount) {
+      try {
+        await quotas.addRows(rowCount)
+      } catch (err: any) {
+        if (err.code && err.code === errors.codes.USAGE_LIMIT_EXCEEDED) {
```
shogunpurple, line 303: NAB: Would be good to send something to posthog here.

`budibase:row_limit_exceeded`

shogunpurple (approved):
Such a huge and complete piece of work - migrations etc accounted for too. Great job - can't wait to get this up and running with our first premium feature. 🚀

packages/server/src/api/controllers/application.ts
Rory-Powell: good idea! this could go in the usage checker itself so we get it for all of them. could then use a namespace like
`budibase:quota_exceeded:row`
`budibase:quota_exceeded:apps`
etc etc

aptkingston (commented):
This is awesome Rory - so much work here. Need to pull this down and check it out properly, but one thing that just jumps out is there are quite a few big deletions to the yarn locks (beyond what would just be minor version bumps) and no corresponding changes to the package.jsons themselves. Wondering what caused that - getting flashbacks of all the client dependencies being removed again haha.

Edit: I also get these huge lockfile changes when adding any new dependencies. So it's probably fine and our lockfiles are just a bit derped.

charts/budibase/templates/app-service-deployment.yaml
Rory-Powell: 👍

Rory-Powell:
@aptkingston the good majority of the lock files changes are related to removing the `@budibase` dependencies once again that get done by lerna. The only other outlier is lots of patch version increments in server which possibly are related to the jest and typescript version bumps. p.s. sorry for just getting back to you on this

Issue #4117, opened 2022-01-20
Templates Section on Home Screen


shogunpurple:
When I land on the initial home screen of budibase, I will see our featured templates. I want to be able to click a button that will present a modal/section showing all of the templates that budibase has on offer.

More context here: https://github.com/Budibase/budibase/pull/4112

Design scope:

- Introducing a new empty state
- Introducing a new Create new app section
- Introducing a new Temples section
- Use thumbnail images from marketing site
- Adding a way for users to preview a working template

https://docs.google.com/document/d/10hGZo_rX6YwjfypubUX3wEjjwrsBP46E0fzFip9DVro/edit?usp=sharing

PR #5002, opened 2022-03-22
Feature/templates home screen

deanhannigan:
## Description

Addresses https://github.com/Budibase/budibase/issues/4117

- Reworked Home Screen
    - With no apps, you are presented with all available templates or the option to create from scratch.
    - With one or more applications, the 'My Apps' table will display as usual.
    - The UI elements for exporting, sorting or filtering apps only appear when more than one app is present.
    - If the user has entered their first name, it will display in the 'Welcome [UserName]!' messaging. It will say 'Welcome back!' otherwise.

- Template Cards
    - Cards will display the remote image configured in the template definition along with the associated background colour.
    - Images have been uploaded for all currently available templates.
    - If a template image is unavailable/down, the template icon will take its place.
    - The 'Details' button links to template overview on the marketing site.
    - The 'Use Template' button opens the create app modal with default information derived from the template name.
    - Templating updates here: https://github.com/Budibase/templates/pull/34

- New Template Screen
    - A new screen to allow users to browse templates at their leisure.

- New 'Create App' Screen
    - A dedicated screen for the app creation workflow.
    - Messaging is slightly different and more focused to the task.

## Screenshots

**Reworked Home Screen - No Apps**
Screenshot 2022-03-22 at 11 48 47

**Reworked Home Screen - With Apps**
Screenshot 2022-03-22 at 12 53 24

**Updated create app modal with template display**
Screenshot 2022-03-22 at 11 49 01

**New App Create Screen**
Screenshot 2022-03-22 at 11 50 04

**New Template Browsing Screen**
Screenshot 2022-03-22 at 11 50 40


deanhannigan:
@joebudi Yes, that's right.

packages/builder/src/components/common/TemplateCard.svelte
```diff
@@ -0,0 +1,123 @@
...
+        width="26px"
+        height="26px"
+        class="spectrum-Icon"
+        style="color: white"
+        focusable="false"
+      >
+        <use xlink:href="#spectrum-icon-18-{icon}" />
+      </svg>
+    </div>
+    <div class={overlayEnabled ? "template-thumbnail-action-overlay" : ""}>
```
aptkingston, line 40: I love these cards! One improvement I think we could make would be to slightly transition the appearance of the overlay to be in line with the rest of the spectrum feel.

I would recomment:
- Set overlay opacity to 0, and leave display as flex always
- On hover, set the overlay opacity to 1
- Add `transition: opacity var(--spectrum-global-animation-duration-100) ease` to match how spectrum components do it

aptkingston (commented):
Amazing work @deanhannigan! Works perfectly, including on mobile. Love the new template card component too.

Being extremely picky here just because it's the home screen, but it would be good to slightly increase the spacing here:
image

So that it lines up with the rest of the screens which tend to have more spacing before the divider:
image

Awesome job overall and just some minor comments added :rocket:

aptkingston:
p.s. welcome to the world of cypress :laughing:
image

Changing the home screen was destined to make that happen!

packages/builder/src/pages/builder/portal/apps/create.svelte
```diff
@@ -0,0 +1,143 @@
...
+    <span>
+      <Link
+        quiet
+        secondary
+        on:click={() => {
+          $goto("../")
+        }}
+      >
+        &lt;&nbsp;Back
+      </Link>
```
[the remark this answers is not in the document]
deanhannigan: Yep, no problem, I'll switch them out.

packages/builder/src/components/common/TemplateDisplay.svelte
```diff
@@ -0,0 +1,147 @@
...
+  const stopAppCreation = () => {
+    template = null
+  }
+</script>
+
+<div class="template-category-filters">
+  <Layout noPadding gap="S">
+    <Heading size="S">Templates</Heading>
+
+    <div class="template-category-filters spectrum-ActionGroup">
```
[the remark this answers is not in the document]
deanhannigan: I think I was looking over the spectrum css docs and took the lead from there. I'll rework them using the existing components.

packages/builder/src/components/common/TemplateCard.svelte
deanhannigan: Sound nice, I'll give it a go!

aptkingston (approved):
Super work Dean, all looks good to me now :+1:

packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -9,17 +9,53 @@
...
   const validation = createValidationStore()
   $: validation.check($values)
 
   onMount(async () => {
+    $values.name = resolveAppName(template, $values.name)
+    nameToUrl($values.name)
     await setupValidation()
   })
 
+  $: appUrl = `${window.location.origin}${
```
Rory-Powell, line 28: I think this will need an `/app` prefix in here. e.g. https://bb.budibase.app/app/agency%20c%20portal%20test

aptkingston:
image
:clap:

packages/builder/src/components/start/CreateAppModal.svelte
deanhannigan: Yes, that's correct. I had a chat with @aptkingston to confirm the expected behaviour regarding the url display here.  A fix has been applied and merged.

Issue #4996, opened 2022-03-22
NGINX `502` bad gateway when accessing `localhost:10000`


daawaan:
**Describe the bug**
I can access `localhost:4001` and it will show the budibase login page but this isn't the intended route. Accessing `localhost:10000` on the other hand will return `502`. The logs will print the following whenever the page is refreshed again:

```
2022/03/22 06:54:25 [error] 30#30: *25 connect() failed (111: Connection refused) while connecting to upstream, client: 172.30.0.1, server: _, request: "GET / HTTP/1.1", upstream: "http://172.17.0.1:4001/", host: "localhost:10000"
```

**To Reproduce**
1. git clone budibase
2. yarn setup

**Expected behavior**
Accessing `localhost:10000` should bring me to the login page for the builder

**Screenshots**
![image](https://user-images.githubusercontent.com/53972232/159422894-2b015570-0095-4b94-938d-858d776ecde9.png)

**Desktop (please complete the following information):**
 - Debian through WSL in Windows 10
 - Chrome
 - 1.0.90

daawaan:
I tried cloning and setting up budibase in a fresh install of Debian, I still can't get it to work

daawaan:
We've tried directly using `host.docker.internal` but the issue still persists for some reason. We're still getting 502 Bad Gateway unfortunately.

shogunpurple:
Hi @daawaan - You will need to restart your NGINX container for the changes to take effect. Could you run a:

`docker restart budi-nginx-dev`

In dev, NGINX uses a file generated by us called `.generated-nginx.dev.conf`. Could you confirm that this is the file that you updated?

daawaan:
@shogunpurple yes i can confirm, we actually ran `yarn nuke:docker` then `yarn setup` to ensure a fresh restart

shogunpurple:
Ok, thanks @daawaan, could you then  run the following command?

`docker exec -it budi-nginx-dev cat /etc/nginx/nginx.conf`

Thanks!

daawaan:
@shogunpurple after all containers have been started or shall i force exit the others first beforehand?

shogunpurple:
@daawaan after they have all started

daawaan:
@shogunpurple okay, thanks for the response! I'm not at work right now so I'll get back to you by tomorrow for the dev logs

daawaan:
@shogunpurple

```nginx
user                    nginx;
error_log               /var/log/nginx/error.log debug;
[...]

PR #5040, opened 2022-03-23
Check for WSL when configuring dev environment on Windows

mike12345567:
## Description
Adding a fix for #4996 - checks if WSL is in use when configuring NGINX.


shogunpurple (approved):
LGTM!

Windows 😭

Issue #4947, opened 2022-03-18
Visting published app in tenant1 while logged into tenant2 produces 404


Rory-Powell:
**Describe the bug**
Create and publish app in tenant1
Log out
Log in to tenant2
Visit url of app in tenant1
See: 
![Screenshot 2022-03-18 at 12 04 28](https://user-images.githubusercontent.com/8755148/158999929-9755aae2-cb34-4373-9bac-6e93a55becd8.png)

Mitch-Budibase:
Testing steps:

login to t1.budipreprod.live
publish -> t1.budipreprod.live/app/123
logout
login to t2.budipreprod.live
go to  t1.budipreprod.live/app/123
go back to t2.budipreprod.live

After following the above steps, I am logged out of t2.budipreprod.live

No 404 error being thrown, but rather the user is always logged out instead.

Mitch-Budibase:
Further testing highlighted the issue I was seeing was related to permissions.

All good on my end, setting this to Done

PR #5044, opened 2022-03-23
Fix/cross tenant apps with session

Rory-Powell:
## Description
Fix: https://github.com/Budibase/budibase/issues/4947

Background:

Viewing a public tenant app in tenant1 while logged into tenant2 previously produced a 404 error.
Viewing with no session did not produce an error.

This was caused by the apps lookup to match the published app url, which used the tenant of the logged in user. 
This resulted in a no-match and a 404 on the db call to get app metadata. 

Fix:
- Shift app url resolution up the stack into `getAppId` from `backend-core/src/utils.js` this is the primary function used to determine the app id from the incoming `ctx` before setting: cookie, cls, ctx.appId
- Rename the above function to `getAppIdFromCtx` to distinguish it from `getAppId` in `backend-core/src/context` which reads from cls
- Remove the url re-write on `/app/` from NGINX so that we don't over eagerly fetch all apps to determine app urls, now we can match on `/app/` like we do with dev apps and `dev_`
- When in the cloud, always honour the tenant id in the URL prior to app url resolution, so we match the correct list of apps


shogunpurple (approved):
LGTM! Nice fix

packages/server/src/definitions/openapi.ts
```diff
@@ -9,273 +9,273 @@ export interface paths {
       parameters: {
         header: {
           /** The ID of the app which this request is targeting. */
-          "x-budibase-app-id": components["parameters"]["appId"];
-        };
-      };
+          "x-budibase-app-id": components["parameters"]["appId"]
```
Rory-Powell, line 12: @mike12345567 not sure why these changes have come in, will this break anything?

mike12345567: Looks like Prettier has got its hands on the openAPI Typescript definitions again, they are supposed to be excluded from prettier but it seems it has picked them up somehow!

In terms of will it break anything, no it doesn't cause any issues, in terms of usage it should be identical - its just non-ideal as its a generated file, each time it is generated again prettier will then need to be ran again etc.

Rory-Powell: Figured this one out, it was happening because we were doing an additional es-lint inside server and worker for typescript files, and in these cases we weren't re-ignoring the openapi.ts file. 

I've moved everything to the top level so we don't need to keep multiple sets of ignore / config files in future, see this commit:
https://github.com/Budibase/budibase/pull/5044/commits/70c54e983d6e3645d14aa200091a5cc01b08aa84

I noticed that some lint violations were being detected this time, which seems to be a result of the previous eslint configs in worker/server not doing anything due to the `files` array being empty. For now I have turned the violations off in the eslint typescript config to reduce noise here, and I'll turn them back on and address in a follow up pr.

PR #5057, opened 2022-03-23
watchtower should restart if it crashes

shogunpurple:
## Description
- Small fix to have watchtower restart on a crash, fixing some update issues from users.

## Screenshots


shogunpurple:
Nope, I think we talked about it but seems like it was never done. May as well take this PR to do it - will update

shogunpurple:
Updated 👍

PR #5067, opened 2022-03-24
Allow deep-getting of pagination params from REST responses

### LINKED ISSUE #5066: REST Connector: Allow Nested Values In Pagination Keys
**Describe the bug**
When you use a nested value in a REST connector as your pagination response body parameter name, it will not actually use the nested value. This is because of the following piece of code in the rest connector:

`packages/server/src/integrations/rest.ts`
```
      let nextCursor = null
      if (pagination?.responseParam) {
        nextCursor = data?.[pagination.responseParam]
      }
```

**To Reproduce**
- Create a REST datasource that relies on pagination
- Choose a nested object key for your pagination response key
- Try to paginate a table with that datasource in the UI

**Expected behavior**
- Pagination should work as expected for cursor and page based.

**Screenshots** (Thanks @melohagan)
image
image
image

aptkingston:
## Description
Allow deep getting of pagination params from REST responses.

Making any changes to the package.json (in this case adding `@types/lodash`) causes this huge change to the lock for some reason. Not sure why.

Fixes https://github.com/Budibase/budibase/issues/5066.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -149,6 +149,7 @@
     "@types/jest": "^26.0.23",
     "@types/koa": "^2.13.3",
     "@types/koa-router": "^7.4.2",
+    "@types/lodash": "4.14.180",
     "@types/node": "^15.12.4",
     "@types/oracledb": "^5.2.1",
     "@types/redis": "^4.0.11",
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM!

Rory-Powell (approved):
Looks good 👍 For the yarn lock only thing I can think is maybe a recent pr has made some changes but not pushed lockfile changes

Issue #5049, opened 2022-03-23
Many-to-many relationships broken on Postgres


esprssoo:
**Describe the bug**
An error occurs when creating rows for 2 tables with a many-to-many relationship between them. This seems to happen for tables created via Budibase.

**To Reproduce**
1. Go to a Postgres data source and create 2 tables (companies and sectors)
2. For each table add a "name" text column.
3. Add a column to the companies table called sectors that is a many to many relationship to sectors. An association table is also created.
4. Add 2 rows to the sectors table
5. Add 1 row to the companies table with the 2 sectors selected.
6. The following error occurs in console

```
[Builder] HTTP 400 on POST:https://<URL>/api/datasource_plus_d0b976b949574f2fa094d7c3d9a0d588__companies/rows
	error: insert or update on table "jt_companies_sectors_sectors_companies" violates foreign key constraint "jt_companies_sectors_sectors_companies_companiesid_foreign" [index.afc77e6b.js:1:2553](https://<URL>/builder/assets/index.afc77e6b.js)
```

Then, it seems a row was actually created with just a single sector instead of 2. This now renders my 2 tables unusable, because now I can't delete rows since there's foreign key constraints. Deleting the relationship does nothing and the association table is kept. Deleting the association table manually from the Database will work, but it does still appear in the table list in Sources. I can do "Fetch tables" but this means that my data types for the rest of the tables will change (Text will become Long Form Text) which might be another unrelated bug.

**Expected behavior**
Expected to be able to add rows and select multiple associations for the second table in a many-to-many relationship. Deleting one column from the relationship would also remove the defined relationship and delete the automatically created association table.

**Screenshots**

**Additional context**
I created all PostgreSQL tables and columns inside Budibase, nothing was added externally.

mike12345567:
Hi @esprssoo thanks for the report - was able to reproduce. I've pushed a fix in #5073 that will mean all new many-to-many relationships made in Budibase will work correctly.

For now, if you remove both of the relationship columns in your `companies` and `sectors` tables, then go to the `companies` table and hit `Define existing relationship` and run through all the options as below:
![image](https://user-images.githubusercontent.com/4407001/159933902-91763945-0402-4e8c-99fe-f98393ae6e9d.png)

It'll fix the relationship. Once the PR is merged relationships will correctly be built when adding a many to many + junction table to existing tables.

PR #5073, opened 2022-03-24
Fixing an issue with many-to-many SQL relationships generated in SQL tables

mike12345567:
## Description
Fixing a mis-match of fields when creating many-to-many relationships on tables from within Budibase - #5049.


Rory-Powell (commented):
Without knowing this domain fully it's hard to give a solid review here, but I think a unit test for the `generateManyLinkSchema` and `generateRelatedSchema ` functions could help future maintainers understand the expected inputs / outputs

Rory-Powell (approved):

PR #5077, opened 2022-03-24
fix: incorrect parameters for public rows editing

dqbd:
When calling `PUT /api/public/v1/tables/{tableId}/rows/{rowId}`, a new row is created instead of patching an existing row. I believe the bug is caused by passing an incorrect parameter to `fixRow` function


dqbd:

dqbd:
recheck

dqbd:
recheck

mike12345567:
Hi @dqbd thanks for finding the issue and a fix! LGTM.

PR #5086, opened 2022-03-25
'Continue if' button action

aptkingston:
## Description
This PR adds a new button action which allows aborting following button actions if certain conditions are met.

image

The expression  is made up of the following parts:
- A type of either "Continue if" (which only continues if the condition is true) or "Stop if" (which only stops if the condition is true)
- A value, which supports data bindings (including previous button action outputs)
- An operator, either "Equals" or "Not equals"
- A reference value, which supports data bindings (including previous button action outputs)

The comparisons support objects, which will e deeply compared by stringifying them.

Placeholders have been added so that even a brand new action makes semantic sense, removing the need for labels:
image


shogunpurple (approved):
LGTM - this is really nice the way it's bindable on both key and value too. Nice work

PR #5087, opened 2022-03-25
ignore AWS config when using endpoint field

shogunpurple:
## Description
Fix for local `endpoint` in dynamo connector

## Screenshots

### DIFF
--- a/packages/server/src/integrations/dynamodb.ts
+++ b/packages/server/src/integrations/dynamodb.ts
@@ -131,7 +131,9 @@ module DynamoModule {
 
     constructor(config: DynamoDBConfig) {
       this.config = config
-      this.connect()
+      if (!this.config.endpoint) {
+        this.connect()
+      }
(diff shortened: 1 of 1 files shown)


packages/server/src/integrations/dynamodb.ts
```diff
@@ -131,7 +131,9 @@ module DynamoModule {
 
     constructor(config: DynamoDBConfig) {
       this.config = config
-      this.connect()
+      if (!this.config.endpoint) {
+        this.connect()
```
Rory-Powell, line 135: Do we not need to call the connect if there is an `endpoint`?
How does that work?

shogunpurple: Nope, it doesn't need the AWS configuration and passes the endpoint straight into the `DynamoClient`

Rory-Powell (approved):

Issue #5127, opened 2022-03-28
Refused to load the image 'data:image/svg+xml,%3Csvg.. because it violates the following Content Security Policy directive: "img-src http: https: data


sergey-dme:
Hi All,

There are lots of SVGs and images (base64) are used in Custom CSS of my app, for branding, UI and design purposes. 
Now after the update to v1.0.98 the elements are not loaded.  

Console returns: `Refused to load the image 'data:image/svg+xml,%3Csvg.. because it violates the following Content Security Policy directive: "img-src http: https: data"`

Would it be possible to make this working? Please... The design is a huge part of my app.

I'm still on v 1.0.35 on my main server

Thank you

aptkingston:
Hi @sergey-dme. Looks like we made a typo - the CSP should end with `data:` rather than `data. We don't mean to block images from data URI's. I'll get that fixed!

sergey-dme:
Hey @aptkingston Thank you for your attention and help! Looking forward to the PR! Appreciated

sarbjit-rnd:
I have configured stripe payment method with test mode in shopify app. But payment is reflected in  stripe dashboard. Can any one has idea where may be the issue.
![image](https://user-images.githubusercontent.com/33019564/228501725-3bce240f-2f0f-48b1-bda5-601ff28a7ff2.png)


Thanks

PR #5131, opened 2022-03-28
Allow data URI's for image sources and font sources

aptkingston:
## Description
This PR fixes some of the CSP schemes to allow local `data:` URI's. These were lacking the colon suffix which meant images using data URI's (and fonts, but don't think they're used) would not work.

Fixes https://github.com/Budibase/budibase/issues/5127.


shogunpurple (approved):
LGTM!

aptkingston:
You say that but I do remember really old exploits with certain TTFs on windows haha. So it's not 100% safe. But in practice yea, I think that would also be a good idea.

Issue #5103, opened 2022-03-25
Issues with templates that include a public form


mjashanks:
All templates that include a public form seem to be broken right now, in 3 separate ways

1. The table that is being inserted into has Basic write access. It should be Public. I assume that the template's table was public when saved
2. The public form's schema is set to the "Create" and the schema is set to the correct table. However, this causes an API cal to `/api/tables/<id>` which fails due to a permissions error. Changing the schema to `Custom` allows the `Save Row" to work however...
3. Right after the row has saves, a  call gets made to `/api/tables/<id>` even though there are no other actions in the sequence
4. If the form has an attachment (e.g. the Applicant Tracker), then initial uploading of the attachment just fails. It's just not possible.

aptkingston:
I tried the applicant tracker template there, and it correctly has public write and basic read. Were you seeing basic access for both?

Copying and pasting message from discord:
technically i actually think basic read with public write is correct - you don't want any old public user being able to see all applicants (using applicant tracker as an example) 
but then yea i understand of course why that would cause the 403 when trying to fetch the schema
with both public read and write, any public user could read all applicants (eg get their emails and contact details) and then be able to even update them
i sometimes wonder if schemas should be under protection or not... i guess it is slightly sensitive info sometimes
updating the form schema to custom fixes the error from popping up since it doesnt need the schema, and theres no downside to that (for creating) other than its harder to maintain in the builder since the form fields arent linked to the available fields from the table

aptkingston:
Just realised that tables that do need public read do not have it - e.g. the vacancies table for the job applicant tracker. Missed that initially.

mjashanks:
Honestly, I think the way we do public forms is all wrong. I think it should be simply a form with a completely custom schema, that sends all of its data to an automation, which then writes to the tables.

shogunpurple:
@mjashanks I like that idea - it effectively removes the concept of completely public read/write. We also don't account for public, tenant based read/write in the budibase API either so it wouldn't affect that.

> But.... building this flow is a bit of a pain.
This says to me that the flow there should just be automated. If you have a form with a (or even regular) schema, you should be able to create a new automation from the button action of the form, and check an option to have it read the schema of the form to create an app action automation with a`Create Row` step.. with all the values filled in. This is a common flow regardless, so it might be nice to introduce a feature to accommodate it

mike12345567:
[...]

PR #5134, opened 2022-03-28
Fix for RBAC on apps built from templates containing public screens

mike12345567:
## Description
Fix for #5103 - some templates are built on an older version that stored permissions differently, we can't migrate these as they will keep being added, easiest to just support the old method (apply the old rule and convert to the new format when retrieving roles).

This fixes the templates which contain public roles, converted the roles which are stored as:
```
"ta_a889aa4c78b14eedaaf2765e30558054": "write"
```
to
```
"ta_d0eaf94963834567ba4fc083d7072118": [
      "write",
      "read"
    ]
```
Which is the new method. Previously "write" as a string implied "read" but that was updated a little while ago. Its easy enough to convert this when reading roles and resolve the issue where used, making use of the older rule that write is a higher level than read.

This also introduces the ability to set a public write, but have a basic read, and the table schema will use the lower of the two roles which provides it access - table schemas see if either of the levels would provide access using the `{ schema: true }` option on the authorized middleware.

### DIFF
--- a/packages/backend-core/src/security/roles.js
+++ b/packages/backend-core/src/security/roles.js
@@ -1,5 +1,5 @@
 const { cloneDeep } = require("lodash/fp")
-const { BUILTIN_PERMISSION_IDS } = require("./permissions")
+const { BUILTIN_PERMISSION_IDS, PermissionLevels } = require("./permissions")
 const {
   generateRoleID,
   getRoleParams,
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM! Great job getting that fix in - should solve a lot of issues.

PR #5136, opened 2022-03-28
Some automation bug fixes

aptkingston:
## Description
Fixes a couple of issues I experienced with automations.

- Fix automation section crashing after deleting a table that is being used
- Fix awful look of `SchemaSetup` (used for app action triggers)
Before:
image
After:
image

I encountered multiple other issues with automations too but those are too big to quickly address. The main one being that the whole idea of debouncing server updates is flawed and will not truly debounce updates because there is no local state to reference. Multiple updates will simply override each other and abandon the previous one. It also makes everything feel very laggy.

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -162,7 +162,7 @@
         <Select
           on:change={e => onChange(e, key)}
           value={inputData[key]}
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM - also experienced the lagginess issue. There's a ticket in there to address some automation bugs for next iteration so we'll get it tidied up then 👍

PClmnt (approved):
Agreed, the debounce thing definitely not the right way to go about it.

PR #5145, opened 2022-03-28
Foreach Automation Block

PClmnt:
## Description
This PR adds an automation block that allows for looping of a connected block. This will allow users to perform much more complex operations with their automations, and hopefully allow for much more flexibility. 

The block can be added by clicking the `Add Looping` button shown here: 

image

This then adds a loop block to the automation, which is stored as a separate block, but linked to the selected block using an ID, that we use to attach both visually and relationally. 

image

As shown in the image above, there are a number of different configuration options that can be provided to the option. And bindings are available for all options. 

- Array / String dropdown denotes what type of binding should be provided to the loop, and if a different type is passed we throw back an error at runtime stating this. 
`{
  "items": [
    {
      "success": true
    },
    {
      "success": true
    }
  ],
  "iterations": 2,
  "success": false,
  "status": "INCORRECT_TYPE"
}`

- Binding / Value is the actual value that you want to loop over
- Max Iterations is a soft limit that is set by the user that defines how many times that automation will run (this is superseded by a hard limit that we or the self-hosting user sets)
`{
  "items": [
    {
      "success": true
    }
  ],
  "iterations": 1,
  "success": false,
  "status": "MAX_ITERATIONS_REACHED"
}`
- Failure Condition stops the loop whenever the `currentItem` equals the value provided in the failure condition, then loop is then broken at that particular point and the results up to then provided
`{
  "items": [
    {
      "success": true
    },
    {
      "success": true
    }
  ],
  "iterations": 2,
  "success": false,
  "status": "FAILURE_CONDITION_MET"
}`

This leads into how we provide the bindings for each step. A block that is being looped gets provided the following binding (in addition to any previous blocks):

image

Any blocks after the loop block get provided with the entire output such as: 

image

Having the full items obviously provides even more flexibility, and hopefully in the future will allow us to do some other cool things with the result, i.e reduce the array down into one etc (shoutout to @mike12345567 for that idea).


charts/budibase/values.yaml
```diff
@@ -103,6 +103,7 @@ globals:
   google:
     clientId: "" 
     secret: ""
+  automationMaxIterations: "0"
```
mike12345567, line 106: should we give this some sort of default in the `values.yaml`? My thinking is for self hosting Kubernetes users - not just BB cloud.

PClmnt: Yep you're right, what do you think is a reasonable max? 500? More? I'm really not sure.

mike12345567: I think 500 would be reasonable for now, since this can perform async operations I think it'd be dangerous to allow too large a limit.

packages/server/src/automations/steps/loop.js
```diff
@@ -0,0 +1,50 @@
...
+          title: "Failure Condition",
+        },
+      },
+      required: ["type", "value", "iterations", "failure"],
+    },
+    outputs: {
+      properties: {
+        items: {
+          customType: "item",
+          description: "the item currently being executed",
```
mike12345567, line 35: Capitalise the t on the start of description, just to make it the same as the rest of descriptions

packages/server/src/automations/steps/loop.js
```diff
@@ -0,0 +1,50 @@
...
+        success: {
+          type: "boolean",
+          description: "Whether the message loop was successfully",
+        },
+        iterations: {
+          type: "number",
+          descriptions: "The amount of times the block ran",
+        },
+      },
+      required: ["success"],
```
mike12345567, line 46: Should `items` and `iterations` be required here as well? I would assume these will always be returned by the loop block - even if they are `items: []` and `iterations: 0`.

packages/server/src/threads/automation.js
```diff
@@ -74,50 +78,200 @@ class Orchestrator {
...
+          }
+
+          // The "Loop" binding in the front end is "fake", so replace it here so the context can understand it
+          // Pretty hacky because we need to account for the row object
+          for (let key in originalStepInput) {
+            if (key === "row") {
+              for (let val in originalStepInput["row"]) {
+                originalStepInput["row"][val] = originalStepInput["row"][
+                  val
+                ].replace(/loop/, `steps.${loopStepNumber}`)
```
mike12345567, line 164: is there a risk here of someone doing a very odd binding like `this is my loopy binding: {{ loop.values }}` and it replacing the constant string rather than the binding value? Might want to search for `loop` inside of HBS statements, we have a function to find HBS statements in `@budibase/string-templates` - its called `findHBSBlocks`.

packages/server/src/threads/automation.js
```diff
@@ -74,50 +78,200 @@ class Orchestrator {
...
+            })
+            loopSteps = null
+            loopStep = null
+            break
+          }
+
+          // The "Loop" binding in the front end is "fake", so replace it here so the context can understand it
+          // Pretty hacky because we need to account for the row object
+          for (let key in originalStepInput) {
+            if (key === "row") {
```
mike12345567, line 160: should we handle other types here than just rows? Just wondering if it'd be safer here to do a
```
for (let [key, value] of Object.entries(originalStepInput)) {
  if (typeof value === "object") {
    for (let [innerKey, innerValue] of Object.entries(originalStepInput[key])) {
      originalStepInput[key][innerKey] = innerValue.replace(...)
```

packages/server/src/threads/automation.js
```diff
@@ -74,50 +78,200 @@ class Orchestrator {
...
+              )
+            }
+          }
+
+          if (
+            index === parseInt(env.AUTOMATION_MAX_ITERATIONS) ||
+            index === loopStep.inputs.iterations
+          ) {
+            this.updateContextAndOutput(loopStepNumber, step, tempOutput, {
+              status: "MAX_ITERATIONS_REACHED",
```
mike12345567, line 179: Could we enumerate the status codes in the `src/constants.js` or somewhere like that for loop error codes, or perhaps even a wider automation error code? Might be something we want to expand upon.

mike12345567 (changes requested):
Just some minor comments, might be worth looking at the comment about the typing and the risk of calling `.replace` without a quick type check, also the regex for searching for `/loop/` and the risk of a larger, concat-ed string might be problematic, worth a check there too.

Otherwise it looks great, looking forward to getting to play with this feature!

PClmnt:
@mike12345567 How's that now?

packages/server/src/automations/steps/loop.js
```diff
@@ -43,7 +43,7 @@ exports.definition = {
           descriptions: "The amount of times the block ran",
         },
       },
-      required: ["success"],
+      required: ["success, items, iterations"],
```
mike12345567, line 46: Minor issue here, this should be `["success", "items", "iterations"]` - right now its just an array of one string.

mike12345567:
@PClmnt - looks great, just one minor issue to fix then LGTM!

mike12345567 (approved):

packages/server/src/automations/steps/loop.js
PClmnt: Whoops! Fixed!

PR #5151, opened 2022-03-29
invalidate sessions before login

### LINKED ISSUE #5055: Logging in on 2 separate browsers on the same account is allowed
**COS**
- Invalidate all sessions on the current user ID when you log in
- Prevent the ability for several people to share a login

mslourens:
## Description
Before creating a new session when logging in, invalidate all other sessions by the same user.
Fixes #5055

### DIFF
--- a/packages/backend-core/src/middleware/passport/local.js
+++ b/packages/backend-core/src/middleware/passport/local.js
@@ -5,10 +5,7 @@ const env = require("../../environment")
 const { getGlobalUserByEmail } = require("../../utils")
 const { authError } = require("./utils")
 const { newid } = require("../../hashing")
-const {
-  createASession,
-  invalidateSessions,
-} = require("../../security/sessions")
+const { createASession } = require("../../security/sessions")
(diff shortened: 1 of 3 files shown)


Rory-Powell (commented):
Thanks for picking this up @mslourens! One suggestion around where the invalidation could go, rather than adding to each login mechanism we could instead add directly into `createASession` so that every time a session is created, no matter the source, we will invalidate the existing

Rory-Powell (commented):
We should hold off on merging this one until we can give it a run through with the account portal as well which also uses the backend-core auth library

mslourens:
> Thanks for picking this up @mslourens! One suggestion around where the invalidation could go, rather than adding to each login mechanism we could instead add directly into `createASession` so that every time a session is created, no matter the source, we will invalidate the existing

I was looking for the best place to put this code, and ended up with this one. I'll update the PR

shogunpurple (approved):
LGTM @mslourens - looks like you maybe just need to pull latest `develop` into your branch and we are good to merge when CI passes.

Issue #5152, opened 2022-03-29
problem with Webhook , error 500 TypeError: Cannot convert undefined or null to object


medvejonok79:
Hi,

I am trying to set up a web hook between Espo CRM and Budibase. I have checked and the CRM works fine, does what it is meant to do. However there is an error 500 on the budibase side.
At first, it looked as though this was because some values in the JSON that was posted by the CRM were empty, but even filling every field with values, I still get the same error.

Here is the relevant section of the app log (run with docker-compose):

```
ERROR [1648544512399] (29 on XYZ): Cannot convert undefined or null to object
    TypeError: Cannot convert undefined or null to object
        at Function.keys (<anonymous>)
        at /app/dist/api/controllers/webhook.js:73:33
        at Generator.next (<anonymous>)
        at fulfilled (/app/dist/api/controllers/webhook.js:5:58)
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
INFO  [1648544412319] (29 on XYZ): request errored
    res: {
      "statusCode": 500,
      "headers": {
        "vary": "Accept-Encoding",
        "set-cookie": [
          "budibase:currentapp=XYZ**************XYZ; path=/; expires=Tue, 19 Jan 2038 03:14:07 GMT"
        ],
        "content-type": "application/json; charset=utf-8",
        "content-length": "69"
      }
    }
    err: {
      "type": "Error",
      "message": "failed with status code 500",
      "stack":
          Error: failed with status code 500
              at ServerResponse.onResFinished (/app/node_modules/pino-http/logger.js:77:38)
              at ServerResponse.emit (events.js:412:35)
              at ServerResponse.emit (domain.js:475:12)
              at ServerResponse.emitted (/app/node_modules/emitter-listener/listener.js:134:21)
              at onFinish (_http_outgoing.js:792:10)
              at callback (internal/streams/writable.js:513:21)
              at afterWrite (internal/streams/writable.js:466:5)
              at afterWriteTick (internal/streams/writable.js:453:10)
              at processTicksAndRejections (internal/process/task_queues.js:81:21)
    }
    responseTime: 37
    req: {
      "id": 179,
      "method": "POST",
      "url": "/api/webhooks/schema/app_dev_06d58e316fc64a88a9a5955c2a298dcb/wh_5f258f7919ff48fbb1b8bcaeba41a5a9",
      "headers": {
        "connection": "upgrade",
[...]

PR #5153, opened 2022-03-29
try converting query parameters to numbers

### LINKED ISSUE #5149: Custom SQL limit & offset params issue
Hello,

I have just upgraded to 1.0.90, then many of my page got error `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''6' OFFSET '0'' at line 1` due to I am using custom SQL with limit and offset for pagination. 

Custom SQL with `limit` and `offset` params will got error, but it works if I use a number instead
image

It works if use  `limit` and `offset` number.
image

It seems Budibase is doing SQL syntax verification, and somehow it found the syntax was wrong.

mslourens:
## Description
Values like offset and limit should be numbers, so we should try to parse numbers when replacing bindings in the query. 
Fixes #5149

### DIFF
--- a/packages/builder/src/stores/backend/queries.js
+++ b/packages/builder/src/stores/backend/queries.js
@@ -79,7 +79,9 @@ export function createQueriesStore() {
       const parameters = query.parameters.reduce(
         (acc, next) => ({
           ...acc,
-          [next.name]: next.default,
+          [next.name]: isNaN(next.default)
+            ? next.default
+            : Number(next.default),
         }),
         {}
       )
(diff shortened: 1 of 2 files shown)


mike12345567:
Hi @mslourens - I noticed issue #5152 - I had been planning to put a change into the MySQL `internalQuery`, have it look for strings that are numbers with a regex like `/^\d*/` - converting them with `parseFloat` when found. Was going to update it to look at the `query.bindings` structure for this.

While this change limits its benefits to that of pagination in MySQL, I think its probably more dangerous to convert any number-like parameter at the query level as there may be plenty of valid reasons that you want to send `"20"`. There are examples that this may have edge case scenarios in, for example `isNaN("               1.2")` returns `false` - suggesting this should be converted to a number, which may not always be the correct thing to do, across all datasource queries.

mslourens:
No problem

PR #5162, opened 2022-03-29
Moving assign/unassign user role type tests to test env only

Mitch-Budibase:
The assign roles test seems to be quite flakey.

I am moving it and the unassign roles test to only run in the test environment while i isolate and fix the issue


shogunpurple (approved):
LGTM!

Issue #5161, opened 2022-03-29
Typing directly into a paragraph element results in double text.


cfepdan:
**Describe the bug**
When I type content directly into a paragraph element (i.e., rather than typing it into the Text field on the right pane), the WYSIWYG text that I type gets duplicated, and visible HTML tags get added.

**To Reproduce**
1. Create a blank screen.
2. Click Elements --> Paragraph
3. Double click on the New Paragraph element
4. Type some words, followed by hitting enter.
5. Click off the element.

In the New Paragraph element, you will see what you typed, followed by a copy of what you typed, followed by <div></div>

This gets a little worse if you use ctrl-b, ctrl-i to add italics while typing. The `<b>` and `<i>` tags will be replicated as visible tags. 

This appears to go better if you double click again, delete the offending text, and start over in the paragraph element.

**Expected behavior**
I expect only one copy of what I typed to appear, with no HTML tags.

**Screenshots**
![image](https://user-images.githubusercontent.com/28074021/160668209-6a85fb6e-703a-4419-8bd5-41cbea05d6ba.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 99.0.4844.84 (64-bit)

PR #5172, opened 2022-03-30
Fix in-preview text editing bugs

aptkingston:
This fixes a couple of issues with editing text in the builder preview. This has been flagged by Joe before and has recently been raised again in https://github.com/Budibase/budibase/issues/5161.

There are 2 problems here:
- Editing text inline allows inserting HTML, so if you copy and paste from the page it messes up
- Chrome has issues with dynamically toggling the contenteditable attribute

This PR fixes these issues. Changelog:
- Only the text content is saved when editing. The downside is that new lines / empty lines are no longer possible.
- Fully remount editable components when changing the editable state. The downside is that buttons will flash briefly when starting editing but this is a very minor UI thing
- Removed some `editable` attributes from the component manifest which should not have been there
- Updated buttons so that if you add a button and don't change the text, it will actually use the default fallback text displayed in the preview rather than show an empty button

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -2529,7 +2529,6 @@
     "name": "Embedded Map",
     "icon": "Location",
     "styles": ["size"],
-    "editable": true,
     "draggable": false,
     "illegalChildren": ["section"],
     "settings": [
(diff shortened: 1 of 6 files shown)


shogunpurple (approved):
LGTM! Noticed some of these ones myself using the platform.

PR #5178, opened 2022-03-30
Fix for MySQL Limits and offsets (numbers in bindings)

mike12345567:
## Description
Fix for #5153 - doing it at the mysql level as it seems to be affected by incorrect types in a way that other SQL databases aren't - limits the risk of the type coercion being applied too widely.

Also, also added a change to the ethereal real email test cases in the worker, they've been failing a little bit due to Jest timeouts when Ethereal is being slow, increased the timeout and made sure that the CI flow isn't considered a failure if Ethereal is being slow.


Rory-Powell (approved):
LGTM 👍 Should this target master?

mike12345567:
> LGTM 👍 Should this target master?

Woops, yeap mis-clicked!

PR #5179, opened 2022-03-30
Fix for MySQL Limits and offsets (numbers in bindings)

mike12345567:
## Description
Fix for https://github.com/Budibase/budibase/pull/5149 - doing it at the mysql level as it seems to be affected by incorrect types in a way that other SQL databases aren't - limits the risk of the type coercion being applied too widely.

Also, also added a change to the ethereal real email test cases in the worker, they've been failing a little bit due to Jest timeouts when Ethereal is being slow, increased the timeout and made sure that the CI flow isn't considered a failure if Ethereal is being slow.

### DIFF
--- a/packages/worker/src/api/routes/tests/realEmail.spec.js
+++ b/packages/worker/src/api/routes/tests/realEmail.spec.js
@@ -3,6 +3,9 @@ const { EmailTemplatePurpose } = require("../../../constants")
 const nodemailer = require("nodemailer")
 const fetch = require("node-fetch")
 
+// for the real email tests give them a long time to try complete/fail
+jest.setTimeout(30000)
+
 describe("/api/global/email", () => {
   let request = setup.getRequest()
   let config = setup.getConfig()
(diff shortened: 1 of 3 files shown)


packages/worker/src/api/routes/tests/realEmail.spec.js
```diff
@@ -3,6 +3,9 @@ const { EmailTemplatePurpose } = require("../../../constants")
 const nodemailer = require("nodemailer")
 const fetch = require("node-fetch")
 
+// for the real email tests give them a long time to try complete/fail
+jest.setTimeout(30000)
```
Rory-Powell, line 7: 👍

Rory-Powell (approved):

packages/worker/src/api/routes/tests/realEmail.spec.js
mike12345567: Noticed it failed on yours and it had failed on one of mslouren's PRs as well - may as well throw a fix in!

shogunpurple (approved):
LGTM!

PR #5201, opened 2022-04-01
Redis Integration

shogunpurple:
## Description
This PR introduces a redis integration to budibase.

**Methods**
- Create: maps to redis `SET`
- Read maps to redis `GET`
- Delete: maps to redis `DEL`
- Command: allows users to pass commands directly to redis, and return the response.

**Note**
- Will hold off on merging this until after the release.

## Screenshots


packages/server/src/integrations/redis.ts
```diff
@@ -0,0 +1,146 @@
...
+      password: {
+        type: "password",
+        required: false,
+      },
+    },
+    query: {
+      create: {
+        type: QueryTypes.FIELDS,
+        fields: {
+          key: {
```
mike12345567, line 44: Should we allow setting a TTL as part of the creation/update? Might be super super useful to be able to set a TTL, if they don't provide a value then just don't set any TTL.

mike12345567 (approved):
LGTM! I like the fact you have added the command option, means you can basically do whatever you want which is amazing!

Sorry for the feature creep etc - but I do think it might be nice to cover the common use case of create with a TTL, just allow the user to specify a number in seconds and pass that to the `.expire(key, seconds)` call when setting a key value pair.

packages/server/src/threads/utils.js
```diff
@@ -72,7 +72,9 @@ exports.hasExtraData = response => {
   return (
     typeof response === "object" &&
     !Array.isArray(response) &&
+    response &&
     response.data != null &&
+    response &&
```
Rory-Powell, line 77: do we need this double `response` check?

Rory-Powell (approved):
Looks great!

packages/server/src/integrations/redis.ts
shogunpurple: Great shout - really nice for caching use cases etc and prevents you needing another delete. Will add!

packages/server/src/threads/utils.js
shogunpurple: Yep, it's really just instead of a `response?.data` - it caught me out at runtime with the redis connector so just added some safety to it.

Rory-Powell: I mean more around we're already doing `response &&` prior to the data check, and then again after. Is the second check here not redundant?

shogunpurple: Ah - yep it is, will remove the second one.

PR #5206, opened 2022-04-01
Updating tests to work with test Env

Mitch-Budibase:
The CI test run is ran with a clean slate (no apps)

The test environment has several existing apps. I have updated the tests to reflect both states.
Also some changes that affect the general flow of the tests


shogunpurple (approved):
LGTM! Once merged I'll update budibase-infra to use the test env, and add the cypress key back in here so we can see the runs in their dashboard.

Issue #5202, opened 2022-04-01
Single Docker Image Deployment


shogunpurple:
**COS**
- Allow budibase to be run in a single docker image

**Notes**
- Assess impact a single docker image has on our deploy targets
- What does this mean for compose
- Needs tested on ARM

mike12345567:
Still required for full feature completion:
1. Adding single image build into CI pipeline, as following:
    * adding `build:docker:single` to
        * `docker:build`
        * `docker:build:develop` (Be good to have develop versions of this image for QA)
    * updating `release-to-dockerhub.sh` to include the Budibase image
    * updating `release-selfhost.yml` to push the latest tag
2. Merging all `nginx.conf` files, right now there is a separate image for the K8S/Docker compose setup which is built with HBS, would be good to merge the single docker file NGINX conf into this build process as well.
3. Exposing more environmental options - right now you can override any of the supplied variables on startup, it would be good to provide these in an .env file as we do with compose and document how to use this in a command.
4. Image size reduction and optimisation, for both the single image and for the `apps` and `worker` images used in the compose/K8S installs.
5. Volume management, some planning around how this will work in the single docker file deployment.

aderici:
I spent quite some time to get this dockerfile for single image working. all went well but i'm stuck at one specific point. would really appreciate any help you could provide : 

<img width="1045" alt="image" src="https://user-images.githubusercontent.com/6830693/167260634-b8ec14fd-db7c-404e-a6eb-b85e028cd3b5.png">

@mike12345567

shogunpurple:
Some futher context on this - we have decided on the extent of the single image and where it will be used. 

- The current docker-compose option is still the preferred option due to the ability to run everything separately, with a cleaner separation of volumes
- We will add a section to the docs for users to run a single docker image, along with all of the options that it provides
- We will replace the digitalocean one click deployment with the single image
- The single image will be the standard method for all future self hosted PaaS like templates - azure, heroku, linode, etc

Timeline for completion on the single image is around the next 2-3 weeks, as we set up the CI pipelines and work on the tasks defined by @mike12345567 above.

aderici:
Thanks @shogunpurple really happy to see a timeline for this. For Azure Appservice there are 2 requirements

1) all important data should be written into somewhere under /home folder within the docker image. Azure Appservice will only persist that folder. This is really important so that we can easily make azure persist our data for couchdb and minio. (I assume there's no need to persist redis data between restarts?)

[...]

PR #5209, opened 2022-04-01
Single docker file lab-day

mike12345567:
## Description
This includes the beginnings of #5202 - a method for running Budibase that is not reliant on a cluster, instead working from a single image. While this isn't scalable, its useful for lots of smaller deployment types that Budibase may be used for, such as deploying a single app, or simply testing self hosting in your setup.

This method should also be ARM compliant, as it does not make use of the IBM CouchDB 3.0 image, instead building the CouchDB + search functionality from scratch using methods similar to that of our Ansible configuration.

This also still makes use of a proxy, running NGINX with a more limited configuration - right now it does not used the templated version, it instead runs a more cut down version - as no network is required for the sake of lab day I just modified the generate NGINX config to map to what was needed in a fully local environment.

This also utilises two Docker volumes, one for MINIO and one for CouchDB + Search indexes, ideally we would provide a run command for users that maps these volumes to a named volume, making it easier for them to reload + reconnect those volumes onto images when an update occurs.

Work that would still need to be done to finish this off:
1. Image optimisation/size reduction - right now the image is very large, roughly 3.5GB compared to a standard CouchDB image which would be anywhere in the 200MB range. I started to diagnose this and see where we could cut down, around 1.5-2GB of this comes from our `node_modules` in the worker/server, due to the fact that we build within the Dockerfile, rather than on the host. Ideally we could strip this down significantly by building on the host and only inserting the distributions for both, along with the required production node modules. I believe it should be possible to get the image size down to 1.5GB range with this change alone - also to note this problem applies to our existing app/worker images, I have just copied over the process that they use so that everything conforms for now.
2. Exposing more environmental options - right now you can override any of the supplied variables on startup, it would be good to provide these in an `.env` file as we do with compose and document how to use this in a command.
3. Volume management - providing some information about how to successfully persist between image upgrades etc - with a single image there is more risk of volumes being lost/incorrectly maintained.
4. CI - including this as part of the CI workflows so that it gets generated as part of releases, this will be very easy but need to setup dockerhub for the publish etc. Also need to decide when/how it should be released, there is a world in which we could release an image per commit or PR, this seems overboard but it could help immensely with testing having lots of easy to use, ephemeral versions of Budibase.
5. As the issue states, generally planning around what this means for our deployment strategies etc before we can release this fully.


hosting/single/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+ENV MINIO_SECRET_KEY=budibase
+ENV SELF_HOSTED=1
+ENV CLUSTER_PORT=10000
+ENV REDIS_PASSWORD=budibase
+ENV ARCHITECTURE=amd
+ENV APP_PORT=4001
+ENV WORKER_PORT=4002
+
+RUN apt-get update
+RUN apt-get install software-properties-common wget nginx -y
```
Rory-Powell, line 22: Could move the `nginx` install to the section near the bottom where the nginx conf is done, or move that conf up here

hosting/single/nginx.conf
```diff
@@ -0,0 +1,116 @@
...
+error_log            /etc/nginx/logs/error.log;
+pid                  /etc/nginx/logs/nginx.pid;
+worker_processes     auto;
+worker_rlimit_nofile 8192;
+
+events {
+  worker_connections  1024;
+}
+
+http {
```
Rory-Powell, line 11: Can we look into having this run through the prod hbs nginx file? That'll give us a win for maintenance so that we don't and can't accidentally forget changes to this file. This looks pretty much like the prod config file ran through the processor so it should be achievable enough

Rory-Powell: Just that having _yet another_ nginx config floating about doesn't sound like a good idea

hosting/single/runner.sh
```diff
@@ -0,0 +1,16 @@
...
+pushd app
+pm2 start --name app "yarn run:docker"
+popd
+pushd worker
+pm2 start --name worker "yarn run:docker"
+popd
+sleep 10
+URL=http://${COUCHDB_USER}:${COUCHDB_PASSWORD}@localhost:5984
+curl -X PUT ${URL}/_users
+curl -X PUT ${URL}/_replicator
```
Rory-Powell, line 15: what's going on with these curl commands, does this mimic what the couch init does?

hosting/single/nginx.conf
mike12345567: Yeah I was hoping to do that at some point, just ran out of time on lab day and didn't want to start spreading the work out across multiple days - was going to try and merge all the configs into one when taking on the full issue itself! Could be a bit of a fiddly task, needs a good bit of testing since it has potential to break all self hosting/cloud deployments.

Rory-Powell (commented):
Great job on this, very impressive amount of configuration sorted out in just one day! I think the only thing remaining would be to actually publish this to dockerhub now by 
- adding `build:docker:single` to
   - `docker:build` 
   - `docker:build:develop` (Be good to have develop versions of this image for QA)
- updating `release-to-dockerhub.sh` to include the `budibase` image
- updating `release-selfhost.yml` to push the latest tag

hosting/single/nginx.conf
Rory-Powell: Sounds good! be great to get them all merged

Issue #5223, opened 2022-04-03
Table block search field flagged 'Required'


bjornclauw:
**Describe the bug**
When clearing a search field in a table block, the frontend flags the searchfield, indicating this value is required.

**To Reproduce**
1. Create new internal Budibase DB table
2. Set one column as 'Required'
3. Add a new 'Table block' linked to our new table
4. Add 'Search Columns' for our 'Required' column
5. Search for *something*
6. Delete value *something* and press Enter

**Expected behavior**
Search fields are not subject to the state 'required'. When clearing them no error is given.

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/32811735/161422429-68121e22-43df-4e7d-85d5-148168f8ac16.jpg)

**Desktop (please complete the following information):**
 - OS: [e.g. Android & Windows]
 - Browser [chrome v99.0.4844.84, samsung browser]
 - Budibase 1.0.99 (selfhosted docker)

Issue #5212, opened 2022-04-01
Update State does not trigger before Save Row action


melohagan:
**Describe the bug**
If I have update state as my first action, I cannot use the new updated state in my second 'Save Row' action

**To Reproduce**
1. Add a button with update action to setup a state that is an array with 3 elements. Have a paragraph to show the length of this state array.
2. Add another button with action > Update State > Set Value (Elements) > JavaScript binding that returns an array of two elements.
3. Add a second binding: Save Row. Return the state array to a column in a table. Save your actions.
4. Run the app. Click the first button to setup 3 elements in the state array.
5. Click the second button to replace the state with a 2 element array.
6. The data table will save 3 elements, whereas the paragraph will update to show the new state of 2 elements.

**Expected behavior**
If I have update state as my first action, then the second action 'Save Row' should use the updated state from Action 1. Or there could be a new action state type for passing and transforming data along an action chain.  

**Screenshots**
| Create 3 Button | Replace With 2 Button |
| ----- | ----- |
| ![Screenshot 2022-04-01 at 14 56 11](https://user-images.githubusercontent.com/101575380/161277980-c4d13aa1-0c89-40c0-a9d6-ebfe5781a8b8.png) *return [1, 2, 3]* | ![Screenshot 2022-04-01 at 14 56 52](https://user-images.githubusercontent.com/101575380/161278092-108a500d-4c06-46ff-86c7-a55fbe8e4f84.png) *return [1, 2]* ![Screenshot 2022-04-01 at 14 57 19](https://user-images.githubusercontent.com/101575380/161278185-a3a111f0-6532-4aa4-b792-0c92c2bee7fd.png) |

**Paragraph**
![Screenshot 2022-04-01 at 15 00 38](https://user-images.githubusercontent.com/101575380/161278791-391c3960-e3ab-4623-83c5-7fad08a57060.png)

**Step 1 - Click Create 3**
![Screenshot 2022-04-01 at 15 00 07](https://user-images.githubusercontent.com/101575380/161278709-047fdb0c-e257-4411-8140-668ed1a1c17b.png)

**Step 2 - Click Replace With 2**
| State | Table |
| ----- | ----- |
[...]

Issue #5199, opened 2022-04-01
Charts: visual accessibility of data labels


andrewleith:
**Describe the bug**
When cycling through the chart palettes, I noticed many label colors are not accessible.  Not only are they not accessible, many can't be read by anyone no matter how good their eyesight is :)

**Screenshots**
A couple of examples:
- In palette 1 
- #505050 on #319cfb
- Contrast of 2.79:1, fails WCAG AA and AAA
  <img width="59" alt="image" src="https://user-images.githubusercontent.com/823749/161183943-73df8507-d6a0-4f0d-b29b-beebf68daf0f.png">

- In palette 4
- #505050 on #667c88
- Contrast of 1.84:1, fails WCAG AA and AAA
  <img width="118" alt="image" src="https://user-images.githubusercontent.com/823749/161183842-b8c4a69d-6171-489f-97fd-52760fc05db5.png">

**Additional context**
Do you have the palette colors documented somewhere?  Do the palettes change based on theme (light/lighter/dark/darker)?

aptkingston:
Hey @andrewleith. The issue here is that we set the text color (for the whole chart) depending on the theme used. So for light themes we set dark text, which is currently also used for those labels as you've pointed out.

I've had a quick look through the palette options and setting the label color to white should work quite well for them all, so I think that's the fix we'll go for.

andrewleith:
Thanks @aptkingston.  That will probably work in most cases.  I see at least one exception in my current app, in palette 4.  The first two label backgrounds are #61d1c9 and #ccf374 - these won't pass contrast with white text.  If you can only set one font color for all, the background colors may need to be revisited.

Issue #4962, opened 2022-03-19
Navigating to newly duplicated row via Navigate To doesn't work properly


Grayda:
**Describe the bug**
When using the "Duplicate Row" action on a button, `{{ Action 1.Duplicated row._id }}` doesn't get updated when you click it more than once

**To Reproduce**
1. Create two tables, `Items` and `Templates`. Add a text field called `Title` to each
2. Create a row in the `Templates` table with the title `Test 1`
3. Generate screens for the two tables
4. On `/templates`, add a button to the Templates Table
5. Define these actions for the button: `Duplicate Row` (Data Source = `Templates Table`, Duplicate To Table = `Items`), and `Navigate To` (Screen = `/items/{{ Action 1.Duplicated row._id }}`, Open Screen in Model = Ticked)
6. Run or publish the app. Go to the templates screen and click the button next to "Test 1".
7. Note that the row in the modal is "Test 1". Change the title of the row from "Test 1" to "Hello" and save it
8. Close the modal using the "X Close" link in the top right-hand corner
9. Click the button next to "Test 1" again
10. **You will be taken to "Hello", instead of "Test 1"**
11. Go to the `/items` screen and note that the row has been successfully duplicated (both "Hello" and "Test 1" will be there), but `{{ Action 1.Duplicated row._id }}` didn't get updated, so you were redirected to the existing "Hello" row, instead of the newly duplicated "Test 1" row

**Expected behavior**
The row displayed in the modal should be the newly duplicated row, each and every time the button is clicked. Right now the `_id` only gets updated once for each button, and you have to refresh the page to clear out the remembered `_id` before you will be taken to the correct page.

I assume this doesn't happen if "Open Screen in Modal" is unticked because it's a full page reload, not just a modal.

**Screenshots**
![budibase_issue](https://user-images.githubusercontent.com/4891563/159123872-89a65b5d-a60a-4acf-baef-4830cf26769b.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10 (64 bit)
 - Browser Chrome
 - Version 99.0.4844.51 (Official Build) (64-bit)

**Additional context**
This bug applies to each individual button, and only until you reload the page. To confirm this:

1. Create a second row in the `Templates` table called "Test 2"
2. Click on the button next to "Test 1" and call the row "Hello 1"
3. Click on the button next to "Test 2". It'll take you to "Test 2" (NOT "Hello 1"). Call the second row "Hello 2"
[...]

PR #5225, opened 2022-04-04
Misc fixes

aptkingston:
## Description
Collection of various fixes.

Fixes:
- Add validation to prevent changing a screen URL to an in-use URL #4840
image
- Add validation to prevent changing a screen role to one that is already in use for that URL #4840
image
- Fix UI not updating properly to display screen list when changing the role for a screen
- Fix flashing empty states when moving around home screen pages as we already have the required data
- Fix flashing of template pictures when loading templates list by just setting images that fail to load to `display: none` rather than defaulting to this state
- Pull in the latest global state context on each button action enrichment to allow for state updates made during the button action execution chain to be referenced by later actions #5212
- Disable validation for search fields used inside blocks #5223
- Update chart data labels to white #5199
image
- Update home screen sizes #5228
image
- Fix chart tooltip color and size
image
- Fix not resetting button action context on each execution of button actions #4962
- Update spectrum card word wrapping again to not break on words that don't need broken
- Fix endless encoding loop when entering screen URLs https://github.com/Budibase/budibase/issues/5291
- Fix issue with automations with no inputs in an app action trigger https://github.com/Budibase/budibase/issues/5300

### DIFF
--- a/packages/builder/src/builderStore/store/screenTemplates/utils/sanitizeUrl.js
+++ b/packages/builder/src/builderStore/store/screenTemplates/utils/sanitizeUrl.js
@@ -3,13 +3,14 @@ export default function (url) {
     .split("/")
     .map(part => {
       part = decodeURIComponent(part)
+      part = part.replace(/ /g, "-")
 
       // If parameter, then use as is
       if (!part.startsWith(":")) {
         part = encodeURIComponent(part)
       }
 
-      return part.replace(/ /g, "-")
+      return part
     })
     .join("/")
     .toLowerCase()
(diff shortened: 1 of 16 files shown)


shogunpurple (approved):
Great fixes here - I think we can probably get this one into `master` to be honest, and just have QA verify it in pre-prod.

aptkingston:
Double checked there but can't just update the base branch since this was originally off develop so contains a bunch of other stuff. I guess cherry picking every individual commit into another branch based off master is the only other way?

Rory-Powell:
> I guess cherry picking every individual commit into another branch based off master is the only other way?

Another method could be to checkout a copy of this branch `cheels-fixes-master` and run a `git rebase master` on that

aptkingston:
Actually think I'd be more comfortable getting this into develop given the amount of changes here. Plus I think it would just lead to conflicts on develop with the other work going in this month.

Issue #5102, opened 2022-03-25
[BUG] Time picker very inconsistent behavior.


ItaloCunhaBueno:
**Describe the bug**

The date picker is very inconsistent when used just to pick time.
In the video attached is possible to see values constantly disappearing and reseting (the value actualy gets written in the field, just the UI goes blank)
The AM/PM button needs several pressings to actualy change. An option to set 24 hours mode would be nice.

**To Reproduce**

1. Add a date picker and set it to "time only"
2. Try to pick a time

**Expected behavior**

The picker should not reset the value and the value should not disappear from the UI.

**Screenshots**

screenshot of the widget config:
![image](https://user-images.githubusercontent.com/47254894/160189108-4770ae1b-61eb-45a4-9291-36c95202fdcc.png)

Video attached

**Desktop (please complete the following information):**
 - OS: docker-compose in Fedora 35
 - Browser used Firefox in Windows 10
 - Version 1.0.90

**Additional context**

Video of the problem:
https://user-images.githubusercontent.com/47254894/160188583-1893ce48-dffa-4785-a3be-ab42b2ad6acd.mp4

yeoldegrove:
I can confirm that the time picker also does not load "default values" defined via a javascript funtion on a firefox 96 on linux.
On chromium 97 this is working fine by the way.

sergey-dme:
Hey all, 

In my case, when binding `{{ now }}` is set as Default value it's unable to change the date in UI - picker always jumps to the current Date, but when I set JS `return Date.now() ` instead, as @aptkingston showed, it crashes everything, even builder freezes and after the change when I close frozen tab and re-open the app in builder mode then navigate to this screen it freezes again, fortunately I was able to I revert changes back and restored my app

aptkingston:
Hi @sergey-dme. All this will be fixed in the next release coming out soon - both Handlebars like `{{ now }}` and JS expressions will be fine and not cause any issues.

Issue #3758, opened 2021-12-10
MySQL time DataType does not work in Budibase


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3752

If we have a `time` column, in MySQL it fails to read into Budibase  - in multiple places:
1. The `Data` section - displayed as `invalid date`
2. The table component - displayed as `invalid date`
3. When loaded into `Date Picker` on a details page, no error - but date picker is empty

If I use a `Data Provider > Repeater > Paragraph bound to time` - the time displays correctly.

It is possible to **save** the time correctly, via a date picker. This works both in
1. Date Picker inside an app
2. Edit row, in the data section

PR #5232, opened 2022-04-04
Fixes for date/time picker issue.

deanhannigan:
## Description
Resolves the behavioural issues raised in the ticket. When setting a date/time field to "Time only", the value will remain

Fixes - https://github.com/Budibase/budibase/issues/5102
Related to - https://github.com/Budibase/budibase/issues/3758

### DIFF
--- a/packages/bbui/src/Form/Core/DatePicker.svelte
+++ b/packages/bbui/src/Form/Core/DatePicker.svelte
@@ -31,8 +31,6 @@
     }
   }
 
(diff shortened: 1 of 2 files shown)


aptkingston (approved):
LGTM! 
Worth noting for future reference (as this caught me out while testing) that a "time only" data type exists in MySQL which is what this issue is really addressing. Using the "time only" setting on a date picker on any other date field doesn't make sense.

PR #5238, opened 2022-04-04
fix CSP variables

shogunpurple:
## Description
Fixing issues with the CSP policy and variables - this has been tested locally in dev also and works as expected.

## Screenshots

### DIFF
--- a/hosting/nginx.prod.conf.hbs
+++ b/hosting/nginx.prod.conf.hbs
@@ -43,7 +43,7 @@ http {
     ignore_invalid_headers off;
     proxy_buffering off;
     
-    set $csp_default "default-src 'self'"
+    set $csp_default "default-src 'self'";
     set $csp_script "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.budi.live https://js.intercomcdn.com https://widget.intercom.io";
     set $csp_style "style-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net https://fonts.googleapis.com https://rsms.me https://maxcdn.bootstrapcdn.com";
     set $csp_object "object-src 'none'";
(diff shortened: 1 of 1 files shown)


hosting/nginx.prod.conf.hbs
```diff
@@ -60,7 +60,7 @@ http {
     add_header X-Frame-Options SAMEORIGIN always;
     add_header X-Content-Type-Options nosniff always;
     add_header X-XSS-Protection "1; mode=block" always;
-    add_header Content-Security-Policy ${csp_default}; ${csp_script}; ${csp_style}; ${csp_object}; ${csp_base_uri}; ${csp_connect}; ${csp_font}; ${csp_frame}; ${csp_img}; ${csp_manifest}; ${csp_media}; ${csp_worker};" always;
+    add_header Content-Security-Policy "${csp_default}; ${csp_script}; ${csp_style}; ${csp_object}; ${csp_base_uri}; ${csp_connect}; ${csp_font}; ${csp_frame}; ${csp_img}; ${csp_manifest}; ${csp_media}; ${csp_worker};" always;
```
Rory-Powell, line 63: Can we add these to the dev config as well? There have been a few issues with CSP in the past, mainly because we don't test out this configuration before deploying to production. Or is there something blocking us from adding to dev?

shogunpurple: There's nothing blocking, the dev one just has to account for all the different URLs that we use locally, and it gets in your way

Rory-Powell (approved):
Can we target this one at master and get it out as a hotfix, some self hosted people might run into this

shogunpurple:
@Rory-Powell this one isn't out in the release - I merged it after we rolled out to master so that we could test it first

PR #5240, opened 2022-04-04
Fix for IN SQL query parsing - handling parentheses and string interpolation

mike12345567:
## Description
Extension of fix for 4978 - fixing an issue where parentheses are added to the IN query, causing the query system to not recognise the need to switch out the binding.

This is caused by some forms of IN query requiring the data be wrapped in parentheses, like `WHERE something IN (1, 2, 3, 4)` - Budibase naturally handles this when coercing the query into the right format to handle SQL variables, however it didn't detect any queries that had been written as `WHERE something IN ({{ list }});` - making sure that if the parentheses have already been added to not add more and to catch the instance in the query so that it can be updated. This was raised in #4978.

Also includes a fix for #5242 - string interpolation was casting too wide a net, for very large queries this was breaking as if you had a lot of bindings and const char strings it would combine some of them, finding bindings in the areas between them and in-correctly interpolating them. To fix this changed it to a two part regex, looking first for the const char strings, then looking within the matches for bindings to switch out.

### DIFF
--- a/packages/server/src/threads/query.js
+++ b/packages/server/src/threads/query.js
@@ -9,6 +9,8 @@ const {
 const { doInAppContext, getAppDB } = require("@budibase/backend-core/context")
 const { isSQL } = require("../integrations/utils")
 
+const CONST_CHAR_REGEX = new RegExp("'[^']*'", "g")
+
 class QueryRunner {
   constructor(input, flags = { noRecursiveQuery: false }) {
     this.datasource = input.datasource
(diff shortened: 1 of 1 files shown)


phil-lange (approved):

crisisdice:
Hey @mike12345567, I tested it locally and it works! Thanks so much for the quick response and looking into this.

shogunpurple (approved):
LGTM!

PR #5258, opened 2022-04-05
Add text input padding when FlashOn icon shown

### LINKED ISSUE #4988: Text is hidden in Bindable Text Field for Long Strings
**Describe the bug**
When typing into the text field with the lightning icon button, you cannot see what you are typing behind it. 

**To Reproduce**
1. Add a Table Block
2. Go to Link Screen, and type in: _This is some long text that gets hidden_
3. As you type near the end, you won't be able to see what characters you are typing / removing

**Expected behavior**
The left hand side of the bindings button should be the end of the text field

**Screenshots**
text_too_long

melohagan:
## Description
Text is now not hidden in Bindable Text Field for Long Strings
https://github.com/Budibase/budibase/issues/4988

### DIFF
--- a/packages/builder/src/components/common/bindings/DrawerBindableCombobox.svelte
+++ b/packages/builder/src/components/common/bindings/DrawerBindableCombobox.svelte
@@ -51,7 +51,7 @@
   }
 </script>
 
-<div class="control">
+<div class="control" class:disabled>
   <Combobox
     {label}
     {disabled}
@@ -120,5 +120,9 @@
     color: var(--spectrum-alias-text-color-hover);
     background-color: var(--spectrum-global-color-gray-50);
     border-color: var(--spectrum-alias-border-color-hover);
+  }
+
+  .control:not(.disabled) :global(.spectrum-Textfield-input) {
+    padding-right: 40px;
   }
 </style>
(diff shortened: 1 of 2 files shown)


melohagan:
recheck

melohagan:

aptkingston (commented):
Good job on your first code change! While this approach works perfectly, we could do it another way which would be a bit cleaner. The core `Input` and `Comboox` components don't need to know anything about data bindings, so it would be best to keep that concept out of them.

What I would do instead is inside `DrawerBindableCombobox` and `DrawerBindableInput` add a global style rule which targets the inner textfield element and adds padding. We can add a class name for the disabled state so that we don't do this when the control is disabled.

First let's add a class name to indicate being disabled, on the top level `div`. `class:disabled` is shorthand for `class:disabled={disabled}` which means in svelte "only apply the class "disabled" if `disabled` is true".
```html
<div class="control" class:disabled>
```

Then we can add a global style rule like this to add padding to the input component when the class `disabled` does not exist:
```css
.control:not(.disabled) :global(.spectrum-Textfield-input) {
  padding-right: 40px;
}
```

The code should be exactly the same for both the combobox and input.

melohagan:
Updated to use global style rule

aptkingston (approved):
LGTM!

PR #3468, opened 2021-11-21
Issue:2993 Copy RoleID

sean-brydon:
## Description
As per #2993 the ability for users to copy the roleID of custom roles.

## Screenshots

CleanShot 2021-11-21 at 20 26 49@2x

CleanShot 2021-11-21 at 20 26 58@2x

PR #5259, opened 2022-04-05
Add role IDs as bindings

aptkingston:
Adds bindings for all roles. This was actually extremely simple to implement as we don't need a corresponding client library change - we can provide the role IDs immediately in the builder as we know them at that stage.

image

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -7,7 +7,11 @@ import {
   getComponentSettings,
 } from "./componentUtils"
 import { store } from "builderStore"
-import { queries as queriesStores, tables as tablesStore } from "stores/backend"
+import {
+  queries as queriesStores,
+  tables as tablesStore,
+  roles as rolesStore,
+} from "stores/backend"
 import {
   makePropSafe,
   isJSBinding,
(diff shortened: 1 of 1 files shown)


upgreydd (approved):

packages/builder/src/builderStore/dataBinding.js
```diff
@@ -391,6 +397,16 @@ const getUrlBindings = asset => {
   }))
 }
 
+const getRoleBindings = () => {
```
Rory-Powell, line 400: awesome this was straight forward. just to check, do we get the custom roles from the `roleStore`?

aptkingston: I think we removed the idea of custom roles... At least I couldn't find them in the UI when I went to check it haha. Either way I do think they are in this role store yea.

Rory-Powell: I ran into this the other day, it wasn't so clear but they are there. You need to add the role in-app to the users table using `Edit Roles`. That role can then be assigned on the edit user page in the portal when assigning a role for that app with the custom role.

Rory-Powell: Could be worth a quick check just to see how it plays, as custom roles were the rationale for the original copy/paste role id issue. We can probably just close this pr https://github.com/Budibase/budibase/pull/3468 when we merge this new change instead

aptkingston: Haha you'd think I'd know where to find it.

Yep works with custom roles:
image

aptkingston: And yea agree about closing that PR once this is in.

Rory-Powell: nice! this is great

Rory-Powell (approved):

Issue #5243, opened 2022-04-04
[Client] HTTP 403 on GET:/api/global/self Forbidden (App with Public access after update)


sergey-dme:
Hey all, 

I tried to access my app in public mode after updating to 1.0.103, the app page and data are shown but there is a red error message `Forbidden` displayed on the page and also in the console:
`[Client] HTTP 403 on GET: ...api/global/self Forbidden`
API returns: `Unauthorized - no public worker access`

v. 1.0.103

mike12345567:
Hi @sergey-dme - I believe you may be suffering from issue: https://github.com/Budibase/budibase/issues/5158 due to the Envoy -> NGINX migration, can you follow the steps in https://github.com/Budibase/budibase/discussions/5169.

sergey-dme:
Hi @mike12345567 thanks for replying.  
I always respond `Y`to update my `docker-compose.yaml` when asked during `hosting --update`
My current proxy section as follows:
```
  proxy-service:
    restart: unless-stopped
    ports:
      - "${MAIN_PORT}:10000"
    container_name: bbproxy
    image: budibase/proxy
    depends_on:
      - minio-service
      - worker-service
      - app-service
      - couchdb-service
```
 The error message in public mode app I got after the update. But It was working in 1.0.98-99

sergey-dme:
Glad to know that I'm not alone with this

sergey-dme:
Hi @aptkingston is it possible that the issue related with this changes https://github.com/Budibase/budibase/commit/bd93f150b5456459eb5f60d13421bd026517109f ?

aptkingston:
> 

Hey @sergey-dme. Yes that's exactly the change that caused this. I had raised the issue internally that it didn't work for public apps and I thought it had been changed, but it looks like it hasn't. We'll get this fixed soon.

FlaminWrap:
Fixed for me in 1.0.108. Thank you!

PR #5263, opened 2022-04-06
Don't show an error for the  endpoint `/api/global/self`

aptkingston:
## Description
Adds the `/api/global/self` endpoint to a list of URLs for which the client app won't show errors. It was easier to fix this issue this way rather than open up the endpoint to public usage, as there are multiple places in the codebase which expect a non-200 error code for this endpoint to indicate being logged out.

Raised in #5243.

Usage of ` characters in my commit seem to have ruined the commit message... TIL.


Rory-Powell (approved):
LGTM 👍

shogunpurple (approved):

Issue #5015, opened 2022-03-22
New Screen Addition UI


shogunpurple:
As a user creating a new screen, I want to be able to determine between creating a blank screen, and creating several CRUD screens.

Spec: https://docs.google.com/document/d/1QLCLczXlAZkQ14hWDrgZiP4skgbiM6SaU8YfDj_QZ50/edit#heading=h.cwqflae0tq9n

**COS**
- Split blank screen and auto generated screens into 2 sections
- Always show the autogenerated screens option
- Disable the autogenerated screens option when there is no data source
- Show a list of data sources under the autogenerated section 

#### CRUD Screens
- When you click a datasource, you can then choose a table under that datasource and generate CRUD screens from it.

**Events**
- `budibase:changed_role_on_screen_creation` - indicate whether people change the role when they create a screen, or we should remove that and just let them change it after
- `Screen Created` - add a "datasource" field to indicate if the CRUD screens have been generated from a particular datasource

**Designs**
Walkthrough - https://www.loom.com/share/62256da462bb456f99f889edf00e985f
XD - https://xd.adobe.com/view/76a3927f-aae5-4944-8686-c5b8872ac013-23f0/?fullscreen

deanhannigan:
@shogunpurple Is there any preference on the contents of the "datasource" column? 

The below shows examples in posthog of either the string name or the type of the source being used.
![Screenshot 2022-04-08 at 11 15 13](https://user-images.githubusercontent.com/5913006/162418452-70b15b69-9d0d-4850-97e4-136155684249.png)

shogunpurple:
@deanhannigan - yep, we should just use the name of the datasource itself, given you could have a few of the same type.

PR #5266, opened 2022-04-06
Screen workflow updates

deanhannigan:
Addresses: https://github.com/Budibase/budibase/issues/5015

## Description
Updated workflow for creating screens in the application.

- Modal UI split into 2 flows for creating screens.
    - Blank -  No data source involved, allowing the user to start from scratch. The name option has been removed as it is no longer used in the application.
    - Autogenerated - All available datasources in the system are listed. One or more can can be selected to automatically generate crud screens
- The screen access level can now be set via a dropdown menu. All screens generated with the flows outlined above will reflect the chosen screen access level.

## Screenshots

**Add screen split into two workflows**
Screenshot 2022-04-25 at 09 10 47

**Auto generate options grouped by source with the option to set screen access level**
Screenshot 2022-04-25 at 09 12 24

**Set screen access levels at the point of creation**
Screenshot 2022-04-25 at 09 12 53


aptkingston:
Nice one Dean! Pulled this down and had a play with it and it works well. There are a few inconsistencies with the design and a few visual things I noticed myself which would be good to correct and I'll note them down here. Apologies for being a stickler but with this being one of the most common flows people will use, I think it's important that it's really polished.

Functionality wise it works perfectly though!

### Initial modal
image
image
- Put the text inside the options like shown in designs
- Update the text of the of the options like shown in the designs
- Update the button text to "Continue"

###  CRUD screens modal
image
image
- Reduce padding between datasource type and datasource options
- Increase padding between different datasources to make the grouping clear
- Might be nice to add the green checks to the options on this screen (to match the initial modal) rather than just highlighting them black
image

### CRUD screen access
image
- The access for CRUD screens was designed to be pulled out into a separate modal to make the CRUD screens modal less busy. If that's not too much work, I do think it would be good to stick to that design.

shogunpurple (approved):
LGTM! Nice work 🚀

PR #5272, opened 2022-04-06
DevTools

aptkingston:
## Description
Thought it was worthwhile actually getting this in since it's been sitting ready for ages, and no point in letting it get stale! I also keep finding that having this would be really useful for developing new components and for the role preview feature.

image

## DevTools overview
This PR adds devtools into dev apps. These will only be visible when viewing your dev app - they will not appear when inside the builder preview or when viewing real published apps.

The features of the devtools are:
- Allow selecting a role to view your app as. This lets you easily see what your app looks like as a public user for example, without having to log out or switch accounts.
image
- Statistics about your app, to help diagnose bugs and performance issues
image
- The ability to inspect and debug a component, letting you see raw (non-enriched) settings, the enriched settings and the value of all available data bindings in real time. The values of any settings or data bindings are copied to the clipboard when clicking on them.
image
image
image

Some refactors as part of this PR:
- Pull out some component state and utilities into a new client libary component store
- Add new core implementation of copying a value to a clipboard and dedupe usages


PClmnt (approved):
This is an incredibly impressive piece of work, just pulled it down and had a play with and it's going to make it so much easier to make changes, especially when adding new bindings stuff. Amazing!

packages/bbui/src/helpers.js
```diff
@@ -106,3 +106,29 @@ export const deepSet = (obj, key, value) => {
 export const cloneDeep = obj => {
   return JSON.parse(JSON.stringify(obj))
 }
+
+/**
+ * Copies a value to the clipboard
+ * @param value the value to copy
+ */
+export const copyToClipboard = value => {
```
shogunpurple, line 114: Never seen this implementation of copy to clipboard before. Love the textarea workaround if the browser API isn't available 👌

packages/client/src/components/ClientApp.svelte
```diff
@@ -55,8 +59,22 @@
...
+
+          // If using devtools, find the first screen matching our role
+          if ($devToolsStore.role) {
+            const roleRoutes = $screenStore.screens.filter(
+              screen => screen.routing?.roleId === $devToolsStore.role
+            )
+            firstRoute = roleRoutes[0]?.routing?.route || "/"
+          }
+
+          // Otherwise just use the fist route
```
shogunpurple, line 72: NAB: typo

shogunpurple (approved):
Love this - it's so feature rich already. I think we will identify lots of other nice dev related features that we can add to this as time goes on too. LGTM!

PR #5294, opened 2022-04-07
Refactored the invalidate session functionality.

deanhannigan:
## Description
Issue caused by merge of: https://github.com/Budibase/budibase/pull/5151

After pulling the above,  the server will no longer accepts requests to create a new session and become unresponsive.

### DIFF
--- a/packages/backend-core/src/security/sessions.js
+++ b/packages/backend-core/src/security/sessions.js
@@ -14,25 +14,7 @@ function makeSessionID(userId, sessionId) {
   return `${userId}/${sessionId}`
 }
 
-exports.createASession = async (userId, session) => {
-  // invalidate all other sessions
-  await this.invalidateSessions(userId)
-
-  const client = await redis.getSessionClient()
-  const sessionId = session.sessionId
-  if (!session.csrfToken) {
-    session.csrfToken = uuidv4()
-  }
-  session = {
-    createdAt: new Date().toISOString(),
-    lastAccessedAt: new Date().toISOString(),
-    ...session,
-    userId,
-  }
-  await client.store(makeSessionID(userId, sessionId), session, EXPIRY_SECONDS)
-}
-
-exports.invalidateSessions = async (userId, sessionIds = null) => {
+async function invalidateSessions(userId, sessionIds = null) {
   let sessions = []
 
   // If no sessionIds, get all the sessions for the user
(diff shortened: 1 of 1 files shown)


Rory-Powell:
Can you explain what has actually changed here? 😅 it looks like just a move

packages/backend-core/src/security/sessions.js
```diff
@@ -14,25 +14,7 @@ function makeSessionID(userId, sessionId) {
...
-    createdAt: new Date().toISOString(),
-    lastAccessedAt: new Date().toISOString(),
-    ...session,
-    userId,
-  }
-  await client.store(makeSessionID(userId, sessionId), session, EXPIRY_SECONDS)
-}
-
-exports.invalidateSessions = async (userId, sessionIds = null) => {
+async function invalidateSessions(userId, sessionIds = null) {
```
Rory-Powell, line 17: Preference would be to keep the ES6 syntax as backend is all going that way, esp with TS refactor of this package pending

aptkingston:
> Can you explain what has actually changed here? sweat_smile it looks like just a move

It's the `this.invalidateSessions` - attempting to reference the export using `this` is a reference either. The options are either pull it out as a separate function and reference it by plain function name (which is what is done here to be consistent with the other function `getSessionsForUser` in this file) or reference it using `exports.invalidateSessions`.

packages/backend-core/src/security/sessions.js
aptkingston: That was my recommendation to keep it consistent with the other function declarations in this file which are just plain functions too (the others that aren't directly exported) - but glad to hear the backend is adopting ES6 syntax more, so we could just update the other functions to all be ES6 syntax too.

deanhannigan:
> Can you explain what has actually changed here? 😅 it looks like just a move

There was a reference to `await this.invalidateSessions(userId)` on line 19 that threw an exception. The use of 'this' was the root of the issue. A minor refactor to declare the invalidateSessions function was needed to achieve the desired outcome. The syntax  was used to keep it in line with `getSessionsForUser`

Rory-Powell (approved):
Looks good! NAB: Could keep it as `exports.invalidateSessions` and just replace `this.` with `exports.`, as this will need refactored to TS eventually

aptkingston (approved):

PR #5298, opened 2022-04-07
Display correct selector label for JS Scripting

### LINKED ISSUE #3926: Update automation JS scripting data "label" to use array access rather than dot notation
**Describe the bug**
JS Script block in automation tab won't work with steps due to invalid syntax. Objects can't begin with a number. "steps.1" / "steps.2" etc

**To Reproduce**
1. Go to 'automation' tab
2. Create a new automation using app action
3. Add a step which returns some data
4. Add a JS Script block after it
5. Add some code and try to return data from the step, eg: "steps.1.response[0]" <-- from a external data source

**Expected behaviour**
Return the first response in the first step

**Screenshots**
chrome_Jt6oySYRlG
chrome_oAGJ232Q6t

 - OS: Windows 10
 - Browser: Chrome
 - Version: 96.0.4664.110
 
 **Suggested fixes?**
 1. return the steps as an array? steps[0] / steps[1]
 2. have each step it's own object? step1 / step2

melohagan:
## Description
See issue: https://github.com/Budibase/budibase/issues/3926
The correct label and path is now shown for JS Scripting bindings.

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -104,9 +104,9 @@
       )
       bindings = bindings.concat(
         outputs.map(([name, value]) => {
-          const stepsLabel = block.name.toLowerCase().includes("bash")
-            ? `steps.${idx}.${name}`
-            : `steps[${idx - 1}].${name}`
+          const stepsLabel = block.name.startsWith("JS")
+            ? `steps[${idx}].${name}`
+            : `steps.${idx}.${name}`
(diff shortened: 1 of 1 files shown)


aptkingston:
I'm sure this is totally correct but could you add a screenshot or 2 of the new behaviour just to show it works? It's always good to see some visual confirmation of changes in PRs if possible. 

Also why is the bash step an edge case that needs the old syntax?

melohagan:
@aptkingston Updated the code to correct the indexing.

**JavaScript code block**:
Screenshot 2022-04-13 at 13 23 12

**Bash code block**:
Screenshot 2022-04-13 at 13 26 01

shogunpurple (approved):
LGTM!

Issue #5014, opened 2022-03-22
Improved Empty States for Backend Data UI


shogunpurple:
As a user who has not added any data to my budibase data UI, I want to have more suggestive CTAs to add columns and rows so that I know what to do next.

Spec: https://docs.google.com/document/d/1lDiTyUA4PMcQ2c6lq7WpF5dHp6e-BAdyKnXTf8T1sqc/edit#

**COS**
- Disable the option to create new row, create view, filter and export as none of these options are useful until they have data
- Add new empty state content with a CTA to create columns 
- Add a new empty state with a CTA to add rows, if the table has columns but no data

**Designs**

- Show empty state when there are no columns created and highlight the columns button
- Show second empty state when there are columns but no rows created and highlight the create row button

Other changes:
- Reposition some of table options
- Change icon for auto columns to visible / visibility off
- Add header to table for the column labels
- Update sources panel to match the panels from the design section
 - Title, CTA, Full width selection like component tree

XD - https://xd.adobe.com/view/31a48354-bf44-438a-a8cd-fe121c18608f-64f7/

PR #5315, opened 2022-04-08
Data UI empty states

aptkingston:
## Description
This PR adds new empty states to tables in the data UI, updates the position and style of table buttons and adjusts some aspects of the data UI size and padding. Spec written up in #5014.

## Table changes
image

- Buttons have been split so that the most common actions are on the left, and the less common actions are on the right
- Table title size increase
- Table title and button padding increased
- Border added around table headers
- Action button alignment fixed

## Empty states and button states
- When no columns exist a new placeholder message is shown, "Create column" is highlighted and "Create row" is disabled
image
- When no rows exist, but columns do, a new placeholder message is shown and both "Create column" and "Create row" are highlighted
image
- "Create view", "Export" and "Filter" are disabled when no rows exist

### DIFF
--- a/packages/bbui/src/ActionButton/ActionButton.svelte
+++ b/packages/bbui/src/ActionButton/ActionButton.svelte
@@ -80,8 +80,4 @@
   .active svg {
     color: var(--spectrum-global-color-blue-600);
   }
-
-  .spectrum-ActionButton-label {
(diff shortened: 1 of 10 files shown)


shogunpurple (approved):
LGTM - nice bit of guidance for users with nothing setup yet.

PR #5316, opened 2022-04-08
QUERY_THREAD_TIMEOUT is string instead of int

upgreydd:
## Description
`QUERY_THREAD_TIMEOUT` is passed to application as a string and can't be used for options which requires integer as a parameter. docker-compose pass all environment variables as strings so declaration like:
```
...
    environment:
      QUERY_THREAD_TIMEOUT: 120000
      ...
```
becomes to:
```
...
    environment:
      QUERY_THREAD_TIMEOUT: "120000"
      ...
```

It's used in 2 places:
- packages/server/src/api/controllers/query/index.js
- packages/server/src/automations/steps/bash.js
In both cases default value is integer.

### DIFF
--- a/packages/server/src/environment.js
+++ b/packages/server/src/environment.js
@@ -68,6 +68,7 @@ module.exports = {
   SENDGRID_API_KEY: process.env.SENDGRID_API_KEY,
   DYNAMO_ENDPOINT: process.env.DYNAMO_ENDPOINT,
   POSTHOG_TOKEN: process.env.POSTHOG_TOKEN,
+  QUERY_THREAD_TIMEOUT: parseIntSafe(process.env.QUERY_THREAD_TIMEOUT),
   // old - to remove
   CLIENT_ID: process.env.CLIENT_ID,
   BUDIBASE_DIR: process.env.BUDIBASE_DIR,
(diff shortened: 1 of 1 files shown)


Upgreydd:
@aptkingston can you please look at this. Thank you in advance

packages/server/src/environment.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
   DEPLOYMENT_CREDENTIALS_URL: process.env.DEPLOYMENT_CREDENTIALS_URL,
   ALLOW_DEV_AUTOMATIONS: process.env.ALLOW_DEV_AUTOMATIONS,
   DISABLE_THREADING: process.env.DISABLE_THREADING,
-  QUERY_THREAD_TIMEOUT: process.env.QUERY_THREAD_TIMEOUT,
+  QUERY_THREAD_TIMEOUT: parseInt(process.env.QUERY_THREAD_TIMEOUT),
```
Rory-Powell, line 74: One thing we might run into here is `undefined` being passed to parse int, resulting in `NaN` as the value for `QUERY_THREAD_TIMEOUT`. This isn't so big of a deal given that `NaN` is falsey, but it might be nice to use a helper function so that we retain `undefined` for the value when it's unset. Something like:

```js
const parseIntSafe(number) = () => {
   if (number) {
      return parseInt(number)
   }
}
```

and use that instead

shogunpurple:
@Upgreydd please do not tag team members directly. We will review PRs as soon as we can.

shogunpurple (approved):

packages/server/src/environment.js
Upgreydd: Added

Rory-Powell (approved):
LGTM

PR #5325, opened 2022-04-08
Added validation for options editor

### LINKED ISSUE #5264: Adding options on a custom schema without specifying values doesn't throw error
**Describe the bug**
When adding custom options to a custom schema form, you need to enter labels and values. When omitting values and pressing save, there's no error, and the options don't get saved.

**To Reproduce**
1. Add a form and keep custom schema
2. Add options picker
3. Define custom schema, and click the "define options" field
4. Enter multiple options, only fill label, keep values empty
5. Click save
6. No error, options not saved

**Expected behavior**
Values should be defaulting to labels, or an error should be thrown values are omitted.

 - Browser: Firefox
 - Version cloud 1.0.106

melohagan:
## Description
As described here (https://github.com/Budibase/budibase/issues/5264) you could click save on invalid options not realising that they had not been saved.
The fix adds a bit of validation that requires at least one option to be filled in before save is successful.

## Screenshots
Screenshot 2022-04-08 at 16 44 32

### DIFF
--- a/packages/bbui/src/Drawer/Drawer.svelte
+++ b/packages/bbui/src/Drawer/Drawer.svelte
@@ -6,7 +6,8 @@
 
   export let title
   export let fillWidth
-  export let visible = false
+
+  let visible = false
 
   export function show() {
     if (visible) {
--- a/packages/bbui/src/Drawer/Drawer.svelte
+++ b/packages/bbui/src/Drawer/Drawer.svelte
@@ -6,8 +6,7 @@
 
   export let title
   export let fillWidth
-
-  let visible = false
+  export let visible = false
 
   export function show() {
     if (visible) {
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
@@ -7,55 +7,60 @@
 
   export let value
 
-  let drawerVisible = false
   let drawer
-  let valid = true
   let tempValue = value || []
 
   const saveFilter = async () => {
-    // Filter out incomplete options, but if all are incomplete then show error
-    let filteredOptions = tempValue.filter(
-      option => option.value && option.label
-    )
-    if (filteredOptions.length > 0 || tempValue.length === 0) {
-      tempValue = filteredOptions
-      valid = true
-      dispatch("change", tempValue)
+    let hasError = false
+    for (let i = 0; i < tempValue.length; i++) {
+      let option = tempValue[i]
+      if (!(option.label && option.value)) {
+        option.error = {
+          label: option.label ? undefined : "You must provide a label.",
+          value: option.value ? undefined : "You must provide a value.",
+        }
+        tempValue[i] = option
+        hasError = true
+      }
+    }
+    if (!hasError) {
       drawer.hide()
-    } else {
-      valid = false
+    }
+    dispatch("change", tempValue)
+  }
+
+  const clearOptionErrors = () => {
+    for (let i = 0; i < tempValue.length; i++) {
+      let option = tempValue[i]
+      option.error = undefined
+      tempValue[i] = option
     }
   }
 
-  //If the drawer is hidden or error corrected, reset the validation
   $: {
-    if (
-      !drawerVisible ||
-      tempValue.some(option => option.label && option.value)
-    ) {
-      valid = true
+    for (let i = 0; i < tempValue.length; i++) {
+      let option = tempValue[i]
+      if (option.error?.label && option.label) {
+        option.error.label = undefined
+      }
+      if (option.error?.value && option.value) {
+        option.error.value = undefined
+      }
+      tempValue[i] = option
     }
   }
 </script>
 
-<ActionButton on:click={drawer.show}>Define Options</ActionButton>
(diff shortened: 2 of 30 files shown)


packages/bbui/src/Drawer/Drawer.svelte
```diff
@@ -6,8 +6,7 @@
 
   export let title
   export let fillWidth
-
-  let visible = false
+  export let visible = false
```
aptkingston, line 9: Not sure about exposing this as a prop, because it allows control by both this prop and also the existing show/hide API that we export.

packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
```diff
@@ -7,22 +7,55 @@
 
   export let value
 
+  let drawerVisible = false
```
aptkingston, line 10: Rather than keeping track of this and resetting the validation whent he drawer is hidden, it would be better to instead just wipe valiation when showing the drawer. We could update the handler for clicking on the "define options" button to both reset the validation and then show the drawer. That's a bit cleaner and means we don't need to change the drawer itself.

aptkingston (commented):
Nice one Mel! Personally I would probably rather add the errors to the actual fields instead of a top level error, because that would also quickly highlight which fields are the problem. I'd also rather prevent saving the drawer if there any errors rather than only stopping saving if everything is an error.

The best way to do this is probably set a flag whenever they try to save the fields, which turns on showing the errors. We can then validate in real time and just use that flag for whether to show the errors or not. Quick snippet would be:
```html
<Input
  placeholder="Value"
  bind:value={option.value}
  label="Value"
  error={showErrors && !option.value ? "Required" : null}
/>
```

Unrelated but the layout of the options drawer looks terrible too (not your fault - it's like that already but I just didn't realise). Not sure how it ended up looking like this. Might be good to correct that at the same time. I think we just need to remove the `label` and `labelPosition` props from the inputs, and then render the labels separately above all the fields. Might need a few other tweaks to align everything properly if we start showing errors in the actual fields too. I can easily help you with that if you need me to :+1:

Topener:
Love it, would it be possible to use the label as value by default? That way less errors and more UX

melohagan:
@Topener @aptkingston Replaced error handling with defaulting options instead:
- use label as value if no value
- use value as label if no label
- remove totally empty options

melohagan:
@Topener @aptkingston Replaced error handling with defaulting options instead:
- use label as value if no value
- use value as label if no label
- remove totally empty options

packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
```diff
@@ -10,9 +10,15 @@
...
-    // Filter out incomplete options
-    tempValue = tempValue.filter(option => option.value && option.label)
+  const saveOptions = async () => {
+    // Filter out incomplete options, default if needed
+    tempValue = tempValue.filter(option => option.value || option.label)
+    for (let i = 0; i < tempValue.length; i++) {
+      let option = tempValue[i]
+      option.label = option.label ? option.label : option.value
+      option.value = option.value ? option.value : option.label
+      tempValue[i] = option
```
aptkingston, line 20: NAB: you don't need to change anything, but just letting you know that you don't need to reassign `option` back to `tempValue[i]` here. When using objects JS will assign by reference rather than value, which means that after doing `let option = tempValue[i]`. both `option` and `tempValue[i]` are referencing the exact same object and mutating either will update the same value.

packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
```diff
@@ -10,9 +10,15 @@
...
-  const saveFilter = async () => {
-    // Filter out incomplete options
-    tempValue = tempValue.filter(option => option.value && option.label)
+  const saveOptions = async () => {
+    // Filter out incomplete options, default if needed
+    tempValue = tempValue.filter(option => option.value || option.label)
+    for (let i = 0; i < tempValue.length; i++) {
+      let option = tempValue[i]
+      option.label = option.label ? option.label : option.value
+      option.value = option.value ? option.value : option.label
```
aptkingston, line 19: NAB: you don't need to update this, but just letting you know that you can use shorthand here to make this even terser
`option.label = option.label || option.value`

There is actually something called a "logical OR assignment" which is never used in the whole project so please don't introduce this, but you can make this *even shorter* by doing
`option.label ||= option.value`
which basically means "only assign `option.value` to `option.label` if `option.label` is falsey"

Alright that's my JS lesson over for the day :laughing:

aptkingston (approved):
LGTM!

packages/builder/src/components/design/PropertiesPanel/PropertyControls/OptionsEditor/OptionsEditor.svelte
melohagan: Nice! I forgot about that in JS. I had tried using '??' but that doesn't work for falsey values.

aptkingston: Yep `??` only works for literal `null` or `undefined` whereas `||` works for anything falsey.

Issue #5339, opened 2022-04-10
Internal Error - Datasource is not configured fully on a page displaying REST data


oboudry:
**Describe the bug**
While reproducing the example provided in the documentation for transfomers https://docs.budibase.com/docs/transformers, I managed to create the REST request with the appropriate transformer. Create a basic page and cards showing the brewery data that is used as an example. But when displaying it, also the content is there and all looks fine, a big ugly red button showing "Internal Error" pops up. 

Server error log can be found in following Gist
https://gist.github.com/oboudry/4ac60ef884cfa271541ac63b98a26606 

Error message is "Datasource is not configured fully" so I suspect it could be due to the fact that my datasource only implements the GET functionality, as it's a REST request with only that option set. But for the use case here I don't see a need for more.

**To Reproduce**
1. Create a new app with a REST datasource.
2. Follow the steps in https://docs.budibase.com/docs/transformers, I put my transformer code here: https://gist.github.com/oboudry/c01769ad77f8574759a3985d2115f41e
3. Create a blank page and add a cards block using the REST query, and map fields (name = state, description = count, image = flags)
4. All looks good but the big red "Internal Server Error" button

**Expected behavior**
No internal server error.

**Screenshots**
![Budibase_REST_Transformer_ERR500](https://user-images.githubusercontent.com/2891956/162619197-72eeaee5-7db6-4fe6-b52a-b1f394570673.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 2004
 - Browser Firefox
 - Version 98.0.2

**Additional context**

that-one-tom:
+1, trying out budibase for the first time today and building a simple frontend for some REST data. No params/headers/auth need to be set. I am getting this error despite the REST API responding as expected and budibase processing and rendering its data.

Running latest version via docker using the docker compose file provided [here](https://docs.budibase.com/docs/docker-compose).

Issue #5318, opened 2022-04-08
Internal Server Error - Budibase Builder using Airtable as datasource


melohagan:
**Describe the bug**
When using Airtable as a datasource, an internal server error message is displayed. It doesn't appear to prevent the data being loaded, it's just not a nice UX.
There's no 500 in the network tab, just in the console. 

**To Reproduce**
1. Add a valid Airtable datasource with query
2. Add a table block with the query as a datasource
3. Data loads, but see error message.

**Expected behavior**
Whenever airtable is used as a datasource, no "Internal Server Error" message should be shown. 

**Screenshots**
![Screenshot 2022-04-08 at 14 38 40](https://user-images.githubusercontent.com/101575380/162447164-2fc69e57-3ddf-4412-9be5-1b685ff40d77.png)
![Screenshot 2022-04-08 at 14 39 50](https://user-images.githubusercontent.com/101575380/162447356-011ce004-f381-4440-ac2b-7573dc599739.png)

```
...,onError:e=>{const{status:t,method:r,url:n,message:a,handled:i}=e||{},o=["analytics","/api/global/self"];if(i){if(a){let e=!1;for(let t of o)if(n?.includes(t)){e=!0;break}e||WI.actions.error(a)}
```

melohagan:
Fixed by: https://github.com/Budibase/budibase/pull/5369

Issue #5144, opened 2022-03-28
Allow multi-select as a search column for table blocks


aidenmitchell:
**Describe the feature request**
This was mentioned here: https://github.com/Budibase/budibase/pull/3249
> You can't currently search on attachments, multi-selects, relationships or formula fields.

Ideally, searching on multi-selects could be added, as it helps with tag-based organization.

tailorck:
Is anyone else getting an issue where the search column drop down is not populated?
<img width="242" alt="image" src="https://user-images.githubusercontent.com/7613002/162985233-4342d171-65ef-453b-a41a-f4fff4582928.png">

aptkingston:
> Is anyone else getting an issue where the search column drop down is not populated? <img alt="image" width="242" src="https://user-images.githubusercontent.com/7613002/162985233-4342d171-65ef-453b-a41a-f4fff4582928.png">

This is technically a bug, but you're using a custom query for your datasource so it means that we can't do any searching for you. I'll push a fix though that will list the "readable" schema here because we can do fake client-side searching on your data at least, even if you're using a custom query.

shogunpurple:
Fixed in https://github.com/Budibase/budibase/pull/5369

PR #5369, opened 2022-04-12
Fixes for custom queries

aptkingston:
## Description
Fixes a couple of issues with handling custom queries and attempting to search.
- Use the readable schema rather than the searchable schema when generating lists of fields to search by (mentioned here https://github.com/Budibase/budibase/issues/5144#issuecomment-1096802409)
- Remove table ID from datasource definitions for custom queries to prevent 500 errors being thrown when attempting to fetch a table definition and sending up a query ID https://github.com/Budibase/budibase/issues/5318 https://github.com/Budibase/budibase/issues/5339
- Fix issue with search field select not being able to parse schemas for datasources that don't include a table ID

### DIFF
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/DataSourceSelect.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/DataSourceSelect.svelte
@@ -52,7 +52,6 @@
     .map(query => ({
       label: query.name,
       name: query.name,
-      tableId: query._id,
       ...query,
       type: "query",
     }))
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/SearchFieldSelect.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/SearchFieldSelect.svelte
@@ -22,7 +22,7 @@
   function getOptions(ds, dsSchema) {
     let base = Object.values(dsSchema)
     if (!ds?.tableId) {
-      return base
+      return base.map(field => field.name)
     }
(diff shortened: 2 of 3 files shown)


mike12345567:
Hey @aptkingston - just wanted to check, I was looking into issue https://github.com/Budibase/budibase/issues/5339 - would this be solved by fix 2 of yours?

Also would it be possible to switch this to go to master since its just two small commits?

mike12345567 (approved):
LGTM - just the comment about whether this can go to master instead of develop since we're quite far from the next release - there's been a couple of issues raised around it I believe.

shogunpurple (approved):
LGTM!

aptkingston:
> Hey @aptkingston - just wanted to check, I was looking into issue #5339 - would this be solved by fix 2 of yours?
> 
> Also would it be possible to switch this to go to master since its just two small commits?

Yep that's indeed this error, hadn't seen that issue so thanks for pointing it out.

mike12345567:
> > Hey @aptkingston - just wanted to check, I was looking into issue #5339 - would this be solved by fix 2 of yours?
> > Also would it be possible to switch this to go to master since its just two small commits?
> 
> Yep that's indeed this error, hadn't seen that issue so thanks for pointing it out. To be honest I'm slightly wary of going to master because removing the table ID from the datasource definition for queries may have side effects that I haven't spotted yet...

Excellent, that's good to get that issue solved as well then. If we put this to master then it'll go onto pre-prod for QA checks before it goes out, I'm just concerned about leaving this for 3-4 weeks in a bugged state, there will likely be quite a few issues raised.

PR #5370, opened 2022-04-12
Backend Events API

Rory-Powell:
## Description
Changes:
- Centralise couchdb configuration in backend-core, rather than having individual configs in server, worker, tests. 
- Remove remaining direct usages of `CouchDB`
- Add custom couch `put` function that adds `createdAt` and `updatedAt` to all documents (enabled by above)
- Event definitions
- Event handlers + usage across the codebase / tests
- Add timekeeper to base test configurations to freeze dates at `2020-01-01T00:00:00.000Z`
- Move user creation out of backend core to resolve cyclic dependencies between worker, pro and core. 

Remaining bits will be tidied up when doing the backfill which will go with this:
- Event properties
- Event Identification
- License events
- High volume events: Automation / Query / Row 
- Enable / Disable analytics on UI
- Remove `endUserPing`


mike12345567 (approved):
This is a huge piece of work, amazing to get all the events setup on all the API endpoints, as well as getting it all well tested from the get go. Also good to get the last of the direct CouchDB references removed, its really nice to have a centralised config/entrypoint for PouchDB, LGTM!

PR #5371, opened 2022-04-12
Smoke Build Test Changes

Mitch-Budibase:
Changes to tests affecting the nightly smoke build.
This build runs with the `test_env` enabled - but differs from the daily test env.

### Updated deleteApp function
- No longer tries to delete an app that does not exist

### Table Pagination Tests
- Updated to include a page reload, page navigation not possible otherwise

### createApp.spec.js
- Test fix for `should provide filterable templates`. Checks if apps already exist and acts accordingly


shogunpurple (approved):
LGTM!

Issue #5322, opened 2022-04-08
DynamoDB stopped working after upgrading BB


superbolt-x:
All my DynamoDB queries stopped working and result in a "worker timeout". Tried to generate another IAM set of keys but still nothing. Any idea why?

mike12345567:
Hi @superbolt-x - could you provide a bit more info on this - like the rough form of the query, which version of Budibase, etc? I've tried this on the currently released version against our local DynamoDB test environment and did not run into any issues.

superbolt-x:
Hey!

sure thing! so I upgraded via the Update button on the UI from BB 1.0.90 to 1.0.110 (docker-compose on an EC2). I increased as u said the limit of the timeout so now i can see a new error - but cannot understand why since am using the us-east-1 default:

<img width="563" alt="Screen Shot 2022-04-09 at 11 49 38 AM" src="https://user-images.githubusercontent.com/67989957/162581590-900a7d49-dd17-4357-a01e-39e402b20a96.png">


Everything was working properly before and even though i was doing some "Scan" queries - thery were taking max 1 sec.
Already tried to refresh credentials - not sure if you added some scope to be enabled in the IAM for the dynamoDB IAM user

mike12345567:
Hi @superbolt-x i've been able to replicate this - i'll get a fix out for it, for now if you want to get it working (until fix is released) you can remove the DynamoDB endpoint, having both a region and DynamoDB endpoint set seems to be causing an issue.

superbolt-x:
Thanks Mike!!

PR #5376, opened 2022-04-12
Fix for DynamoDB queries - connecting with endpoint set

mike12345567:
## Description
Fixing issue #5322 - when endpoint is specified for DynamoDB and not in a the region that the whole instance expects it will throw an odd credentials error - making sure connection is commenced everywhere other than localhost.


shogunpurple (approved):
LGTM! This one is my bad 🙏

PR #5378, opened 2022-04-12
Airtable 500 Fix: Only fetch the table definition if id is not query

### LINKED ISSUE #5318: Internal Server Error - Budibase Builder using Airtable as datasource

melohagan:
## Description
Airtable 500 bug: https://github.com/Budibase/budibase/issues/5318

Made a change to prevent a call being made to the fetch table definition if it's a query datasource.

### DIFF
--- a/packages/client/src/api/patches.js
+++ b/packages/client/src/api/patches.js
@@ -96,7 +96,9 @@ export const patchAPI = API => {
   // handlebars enrichment
   const fetchTableDefinition = API.fetchTableDefinition
   API.fetchTableDefinition = async tableId => {
-    const definition = tableId.startsWith("ta_") ? await fetchTableDefinition(tableId) : undefined
+    const definition = !tableId.startsWith("query")
+      ? await fetchTableDefinition(tableId)
+      : undefined
     Object.keys(definition?.schema || {}).forEach(field => {
       if (definition.schema[field]?.type === "formula") {
         delete definition.schema[field].formula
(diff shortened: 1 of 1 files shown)


aptkingston:
Well found Mel! This will stop the symptom, but the root cause will be fixed by this PR: https://github.com/Budibase/budibase/pull/5369 which we're going to pull into master today to fix the issue in production. There's no need to add both, so while I'm impressed you managed to find the issue here we'd better close this one off.

melohagan:
@aptkingston Brilliant! I had a feeling there'd be a better way as I couldn't figure out how that method was called. Closing.

Issue #4349, opened 2022-02-06
Design Selection sometimes covering the Object


NanoClass:
STEP TO REPRODUCE:

If you create a button on the top of the layout

![image](https://user-images.githubusercontent.com/37381079/152667546-8886cf4b-ae65-476d-882b-9d9a9194ef73.png)

and then select the button to edit it,
The Design modal will cover the button

![image](https://user-images.githubusercontent.com/37381079/152667533-799d2d0f-b8f1-4ce8-8fce-2eff5fa0a26f.png)

Maybe you can implement some "Drag" feature for the Modal.
So we can move the modal somewhere else.

Desktop (please complete the following information):

OS: any
Browser any
Version any

aptkingston:
Thanks for reporting @NanoClass! We can hopefully detect when this is happening and show the bar underneath the component instead.

PR #5399, opened 2022-04-13
Move settings bar below element if at very top [TWEAK]

melohagan:
## Description
Only apply this to small components (e.g. button, headline)
Issue: https://github.com/Budibase/budibase/issues/4349


shogunpurple (approved):
LGTM! Nice fix.

Issue #5392, opened 2022-04-13
Allow assignment of field values, on change of other inputs


mjashanks:
Recently, we changed the way `Default Value` works, so that it only executes once - on load.

Whilst this was the correct decision, and fits better with the term "Default Value", it was a breaking change.

Previously, people were using `Default Value` to trigger updates on other fields. I have recommended this method many times. Discussions have been raised about this already.

So, I am proposing to fill this gap with the following changes.

### 1. Add "Define Actions" for  `On Change` of all form inputs  

This is something that has been requested many times, and is a useful feature in it's own right. For example, people want to trigger `Execute Query` on change of a dropdown.

### 2. Extend the `Update State` action to allow assignment of form fields

![image](https://user-images.githubusercontent.com/3524181/163145387-979ab08b-b47b-4b01-a698-d51ef24d159d.png)

- In the "Type", add a "Set Form Field"
- The "Key" dropdown would then be populated with all available form fields

We could also implement this as a separate action.

### SUMMARY

So, the overall flow to update a text input, based on a select input:
1. Add a "Define Action" for "On Change" of the Select
2. Add and "Update State" action
3. Set the `Form.Fields.MyTextColumn` to whatever value we defined

I think this makes sense overall, and is how I would normally approach this problem as a programmer anyway.

dubbya:
Theres also oninput to capture text field changes without losing focus.

aptkingston:
> @mjashanks @aptkingston Can this issue be closed - is there anything else that needs done?

Yep this can be closed off!

PR #5417, opened 2022-04-14
Form actions

aptkingston:
## Description
This PR adds actions to provide much better control of form fields, and to react to changes in form fields. This fixes the issue of being unable to dynamically update form field values, which was a consequence of recently having to update how field default values are handled.
Mentioned in https://github.com/Budibase/budibase/issues/5392.

### On change settings
All fields have an "on change" setting which will fire when the user chages the value of the field.
image

This action has an input binding of the value of the field.
image

You can conduitionally do actions depending on the new value if you combine this action with the "continue if" action. For example you could only proceed if the new value is "foo", or stop if the value is "bar".

Clarification of behaviour:
- Inputs only update their value when the field is blurred, not on every keystroke
- Options, datepickers etc will update their value immediately when a new value is picked
- If you have a "on change" setting for field A, and then you use the new update field action to programatically set the value of field A, the "on change" setting will *not* trigger. It is only triggered by user interaction.

### Update field value action
A new button action has been added to allow updating the value of a form field. You can either set an explicit value for a form field, or reset a field back to its default value.
image

This action takes 4 settings:
- The form that contains the field to update
- The type of update (either setting an explicit value, or resetting the field back to the default value)
image
- The field to update (automatically populated in a dropdown)
image
- The value to set (if choosing to set an explicit value - hidden if choosing to reset the field to the default value)

If setting a value for a field that takes a non-primitive value (e.g. a multiselect field, a relationship picker or an attachment field) then you need to use a JS binding to set an actual array or object. For example, to update a multi-select field to select 2 options, your binding might look like this:
image

### Other stuff
- Update the sizing or the text and buttons on the create new app screen to match the home screen components which were recently resized


shogunpurple (approved):
LGTM - this is a really nice feature 👍

Issue #5411, opened 2022-04-14
SQL query for IN state with string is not working


Kanary159357:
A change in the way SQL parameters work has made the SQL ... WHERE something IN {{ values }} impossible.

Discussed in https://github.com/Budibase/budibase/issues/4978
cc. https://github.com/Budibase/budibase/issues/3863

Hi, I'm using cloud version of budibase that is 1.0.110

It seems that budibase currently work on IN hotfixs, and there is problem when using IN with string value

```
select * from tableA
where method in {{array}}
```
and array is     
```Alex, Jin```

And then I have this error
```Query results empty. Please execute a query with results to create your schema.```

when I do a query with direct value

```
select * from tableA
where method in ('Alex', 'Jin')
```
It works well

Is there anything I miss?? Or is this budibase's error?

mike12345567:
Hi @Kanary159357 - I believe this maybe due to whitespace in the binding, we split the binding into variables that can be injected into the SQL query, so `Alex, Jin` will split into `'Alex'` and `' Jin'` - if you could remove the whitespace from your query to confirm the issue that would be great.

I'll also update Budibase to trim whitespace on the variables as that's a better/less confusing default.

PR #5419, opened 2022-04-14
Fix for SQL IN queries, trimming the inputs to remove excess whitespace

mike12345567:
## Description
Fix for #5411 - trimming the various binding components for SQL in queries. Also moving SQL interpolation out into a Typescript - prep for further unit testing/improvements.

Also added a minor change to make the query timeout error more descriptive than `Worker call timed out!` - it now states specifically that the query timed out and what the timeout period is currently set to. Fixed the REST UI query error notification output.


shogunpurple (approved):
Great job getting this fix in. Nice one 👍

PR #5443, opened 2022-04-15
Fix many-many multi select options. 

### LINKED ISSUE #5204: Setting MANY relations removes displayed rows from parent table and multiselect dropdown
**Describe the bug**
When setting multiple "MANY" relations from a parent table to child tables, the number of entries shown in the parent table and in the multiselect dropdown is greatly reduced for no apparent reason (the parent_child_mapping table has entries for all parent_row/child_row relations). 

Using MySQL as the configured datasource.

Tables: `PARENT_TABLE` > `PARENT_CHILD_MAPPING_TABLE`  (id, parent_id, child_id) > `CHILD_TABLE`

**Expected behavior**
All relations entries are shown in the respective budibase tables as well as in the multiselect dropdown menus.

 - OS: Mac OS (running docker-compose)
 - Browser [chrome]
 - Version [latest docker-compose]

Thanks in advance!

melohagan:
## Description
Issue here: https://github.com/Budibase/budibase/issues/5204
When the mutli-select value was an array of objects, the *optionsSet* was treating unique relationship combinations as the same, e.g. `[object], [object], [object]`

This change creates a unique id by joining the ids of the array elements into one string. 

## Screenshots
**FIXED**
Screenshot 2022-04-15 at 15 07 01

**BROKEN**
Screenshot 2022-04-15 at 14 24 32

### DIFF
--- a/packages/client/src/components/app/forms/optionsParser.js
+++ b/packages/client/src/components/app/forms/optionsParser.js
@@ -29,9 +29,12 @@ export const getOptions = (
   if (optionsSource === "provider" && valueColumn) {
     let optionsSet = {}
     dataProvider?.rows?.forEach(row => {
-      const value = row?.[valueColumn]
+      let value = row?.[valueColumn]
+      const label = row[labelColumn] || value
+      if (Array.isArray(value)) {
+        value = value.map(item => item._id).join()
+      }
       if (value != null) {
-        const label = row[labelColumn] || value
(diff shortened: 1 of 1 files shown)


melohagan:
Actually I don't this makes any sense. 
I think the solution here is to not show relationship arrays as selectable value columns in a multi-select. It's confusing when you have the Relationship Picker (which there are no docs for).

PR #5447, opened 2022-04-16
Allowing budibase to connect to oidc authentication providers

StarZeus:
## Description
Current budibase self hosted behind corporate proxy is not able to connect to oidc provider hosted in internet since the node-fetch package used in packages/backend-core is not respecting proxy setting set environment (http_proxy and https_proxy env variables).  Therefore I had to make 2 changes
1. Installed `node-fetch-with-proxy` npm package in `packages/backend-core` and used that in `packages/backend-core/src/middleware/passport/oidc.js`
2. Replaced npm package `@techpass/passport-openidconnect` with `@starzeus/passport-openidconnect-proxy` that allows us to get access token from internet hosted oidc provider via proxy set in env variables.

## Screenshots
It is a backend enhancement so there is no screenshot

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.105-alpha.14",
+  "version": "1.0.105-alpha.19",
   "npmClient": "yarn",
   "packages": [
     "packages/*"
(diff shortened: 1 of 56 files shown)


StarZeus:

PR #5475, opened 2022-04-19
Collect hosting information in bug report template

Rory-Powell:
## Description
Update the `bug_report.md` template to collect information about the users hosting method


shogunpurple (approved):
Great idea. This will help a lot!

PR #5492, opened 2022-04-19
Add k8s cron job to periodically rollout restart deployment

Rory-Powell:
## Description
Adds a cron job to optionally restart app-service and worker-service every 60 minutes


shogunpurple (approved):
LGTM! The only thing I would say is that I think this should be in `budibase-infra`, rather than the main repo. We should just be able to deploy this directly into our cluster from there, and remove it once we don't need it anymore.

PR #5498, opened 2022-04-20
Remove Row Rev field for Actions and Automations

### LINKED ISSUE #5326: Documentation for Delete Row: What is Row Rev?
I auto-generated a CRUD screen for a table in my MySQL database.

When looking at the actions performed by the "Delete" button,   the source of Row ID is intuitively `{{ Repeater.contacts._id }}`

The source of Row Rev is less intuitive... `{{ [c691d93a5f02f486c900fa198e3736633].[_rev] }}`

Displaying this in a paragraph reveals that this in fact contains the string "rev"

I have not found Row Rev referenced in the docs for actions... https://docs.budibase.com/docs/actions

What's it mean?

melohagan:
## Description
A few users were confused about Row Rev, and it was decided it should be removed as it has a very niche use-case. Deletes from the internal DB will now behave similar to those in SQL. 
The latest Row Rev is taken when the row data is retrieved. 

## Screenshots
Screenshot 2022-04-20 at 12 15 06

Screenshot 2022-04-20 at 12 15 40

### DIFF
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/ButtonActionEditor/actions/DeleteRow.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/ButtonActionEditor/actions/DeleteRow.svelte
@@ -26,14 +26,6 @@
     on:change={value => (parameters.rowId = value.detail)}
   />
 
-  <Label small>Row Rev</Label>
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM! This will simplify things a lot and clear it up for users. Great job 🚀

PR #5510, opened 2022-04-20
PouchDB memory leak protection

mike12345567:
## Description
This PR introduces a fix for the memory leak that we've been experiencing in our app and worker services, which appears to be linked to our usage of PouchDB.

Generally PouchDB isn't designed for a use case like ours, were we open a database per connection and we never *technically* close this database. This doesn't cause any issues for CouchDB as there isn't specifically a connection (only HTTP requests) but it leaves around some useless event emitters, which explains the gradual growth in memory usage. The `close()` function (documented [here](https://pouchdb.com/api.html#close_database)) is used to close the event emitters and consider the database connection "finished".

To accomplish this I utilised a lot of the work we've done around storing the database to the function context during an API call - essentially caching it until the API controller completes all of the necessary operations. Once the operations are complete it will call the close function and this should free any used memory.

I've also upgraded our PouchDB usages to [7.3.0](https://pouchdb.com/2022/04/13/pouchdb-7.3.0.html) which supposedly fixes one of the memory leaks around failed replications (unlikely to have affected us massively, but every little helps). I've pinned this version for now so that once we resolve issues with this we can depend on it to remain consistent until we need to perform another upgrade.

Thanks to @Rory-Powell for his work on the final consolidation of the `new PouchDB` calls - removing them all from the rest of the server/worker code base and replacing them with the `getDB` calls. I've renamed the `getDB` to `dangerousGetDB` so that we can be very specific about where we need to do dangerous operations.

There are only a few places this needs done now:

1. As part of the context for the app DBs
2. As part of the context for the tenancy, storing a global DB instance
3. As part of the `doWithDB` - which is similar to our `doInTenant` functionality, it wraps the provided task in a function which will open the DB and then close it when the task is complete.
4. Replication - the replication functionality we have built into the `backend-core` is class based and therefore it is a little more difficult for us to manage the DB connections using closures. Instead in this case I've added a `close` function to the replication (as there are no destructors in JS [please please please](https://github.com/tc39/proposal-weakrefs#finalizers)) which can be called when all of the replication functionality is complete - i've added to this to the three places that we currently utilise the replication functionality (app deployment,  dev app revert and the app sync).

To confirm that this covers everywhere I expected, I added a check at the core level, around the `dangerousGetDB` function, the only place from which databases can now be created, storing the database reference when created and then removing it upon deletion.

Can see here after a few hundred requests that all databases that have been created have successfully been closed.
image

I also inspected memory usage again, while Pouch wasn't consuming zero memory, it was down from 30% of the overall usage in dev (likely more in production) to 4%.

Ideally we will want to create some sort of leaky test case if this proves to be the issue, some way to catch if anything ever goes wrong in this area again.


packages/backend-core/src/context/index.js
```diff
@@ -140,6 +197,22 @@ exports.setTenantId = (
   if (tenantId) {
     cls.setOnContext(ContextKeys.TENANT_ID, tenantId)
   }
+  return tenantId
+}
+
+exports.setGlobalDB = tenantId => {
```
Rory-Powell, line 203: Nice to see the global dbs being added to context as well, another optimisation

packages/backend-core/src/db/index.js
```diff
@@ -1,13 +1,64 @@
...
-  return new Pouch(dbName)
+// we have to use a callback for this so that we can close
+// the DB when we're done, without this manual requests would
+// need to close the database when done with it to avoid memory leaks
+exports.doWithDB = async (dbName, cb, opts) => {
+  const db = exports.dangerousGetDB(dbName, opts)
+  // need this to be async so that we can correctly close DB after all
+  // async operations have been completed
+  const resp = await cb(db)
+  await exports.closeDB(db)
```
Rory-Powell, line 57: Should this do something like:
```js
try {
   const resp = await cb(db)
   return resp
} finally {
   await exports.closeDB(db)
}
```

Similar to how the closing is done in the `doInAppContext` and `doInTenant`,  we'd get all the manual usages covered then as well.

packages/server/src/api/controllers/deploy/index.js
```diff
@@ -94,18 +94,16 @@ async function initDeployedApp(prodAppId) {
...
-    const productionAppId = getProdAppID(appId)
-
-    const replication = new Replication({
-      source: devAppId,
-      target: productionAppId,
-    })
+  const appId = getAppId()
+  const devAppId = getDevelopmentAppID(appId)
+  const productionAppId = getProdAppID(appId)
+  const replication = new Replication({
```
Rory-Powell, line 100: could use `let replication` outside the try and do the rest inside the try like we had before so we can catch any of exceptions there, then in the finally wrap the close with `if (replication)`

packages/server/src/db/inMemoryView.js
```diff
@@ -1,48 +1,49 @@
...
+        continue
+      }
+      const found = data.find(possible => possible._id === row._id)
+      if (found) {
+        row._rev = found._rev
+      }
     }
+    return response
+  } finally {
+    await db.destroy()
```
Rory-Powell, line 46: edge case: the destroy throws an exception and the close never gets invoked
could wrap with an inner try-catch here to be safe 
```js
} finally {
   try {
      await db.destroy()
   } catch (e) {
      throw e
   } finally {
      await closeDB(db)
   }
}
```

packages/server/yarn.lock
```diff
@@ -995,30 +995,6 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
-"@budibase/backend-core@^1.0.97":
```
Rory-Powell, line 998: Annoying how the lerna dependencies keep getting re-added then removed. Chatted with cheeks about this before, looks like they come back in when a `yarn install` is done inside one of the packages. 

Trying to spread awareness that we should always run a `yarn bootstrap` to manage any installation of packages which will prevent the budibase dependencies being added to the lock file. We could maybe add something to CI or a pre-commit hook to check if a `.lock` file contains another dependency from the mono repo and fail just so we don't keep seeing these diffs

Rory-Powell (commented):
Looks great! Brilliant getting this in so quickly. 

All minor comments, only thing might be worth looking at is the exception handling in `doWithDB`

packages/server/src/db/inMemoryView.js
mike12345567: in this case I think `destroy` closes the DB as well, I put the close in just as an extra precaution, if it throws an error then I don't think there is a way it can be closed as well anyway - destroy is basically the same as close.

packages/server/yarn.lock
mike12345567: ah interesting, I didn't realise that, what happened here was I installed a new version of `pouchdb` - right now I don't think we have a way to do this through lerna easily, perhaps we need to provide some tooling to add new packages like `yarn add pouchdb`

Rory-Powell: you're right doing a yarn add will bring them back too. Normally I run a `yarn clean && yarn bootstrap` to re-sanitise the lock files after that

mike12345567: I will actually add a top level operation which runs through all of the repos and runs a yarn clean, then runs a bootstrap to update.

Rory-Powell (approved):

PR #5527, opened 2022-04-21
Fix QUERY_THREAD_TIMEOUT out of range

xnohat:
## Description
environment.QUERY_THREAD_TIMEOUT must be an unsigned integer, but environment.QUERY_THREAD_TIMEOUT now return a string
lead to error: The value of "timeout" is out of range. It must be an unsigned integer. Received '10000'"

This is minor fix for [#5267](https://github.com/Budibase/budibase/pull/5267) of @shogunpurple 

## Screenshots
image


xnohat:

Issue #5530, opened 2022-04-22
Add the RejectUnauthorized switch to MySQL DB Integration SSL


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5515

<div type='discussions-op-text'>
For the mysql2/promise library I can see that RejectUnauthorized is defaulted to true:
https://github.com/sidorares/node-mysql2/blob/fce1a49b97b259b3926101fd1128835e0aa0bcc4/lib/connection_config.js
</div>

As a user I want to be able to disable **RejectUnauthorized** for MySQL DB integrations.

Issue #5495, opened 2022-04-20
Field from Relationship Unavailable with Budibase defined relationship between SQL tables


zimmra:
## **Hosting**

- Self
  - Method: <method>docker compose
  - Budibase Version: <version>1.0.110
  - Database: MS SQL (also confirmed on PG SQL)

## **Describe the bug**
When attempting to use a field from a related table, the object is only populated with `primaryDisplay` and `_id`. 

 ***I believe budibase is either is improperly loading the related object, or the formula helpers are incorrectly recommending these fields as being available as a relationship/in the formula***

## **To Reproduce**
1. Create 2 MSSQL Tables, with an ID column for use as key and additional columns to reference 
2. Create Relationship with ID as key 
3. Create new 'Formula' column on one table
4. Try to use a formula to reference the related table (e.g. `{{ fromtable.0.examplecolumn }}`
5. previous step yields no results, but `{{ fromtable }}` has `primaryDisplay` and `_id` as the only fields in the object

## **Expected behavior**
For the value from related table `examplecolumn` to be displayed. 

## **Screenshots**
> ### Formula Helpers indicating these fields can be used
>>![image](https://user-images.githubusercontent.com/28514085/164147439-f5b6e492-ec2f-46c9-a274-35baacb7b1af.png)

**Additional context**
Info on my table structure below:

>#### Table: TimeSheets
>>primary key: Id
Created Relationship Column: TimeSheetEntries
Desired Related Columns: TimeSheetEntries.StartTime, TimeSheetEntries.EndTime, etc

>#### Table: TimeSheetEntries
>>primary key: Id (_does not equal TimeSheets.Id_)
bb relationship foreign key: TimeSheetID (_does equal TimeSheets.Id_)
Created Relationship Column: TimeSheets

From discussion #5494

zimmra:
For what it's worth, I have since recreated this scenario with PGSQL, so it is not limited to just MSSQL

zimmra:
@mike12345567, I spun up a second stack running branch v1.0.117 and I am able to properly reference related table columns!

Do you guys prefer for these issues to stay open until it's in the production release or should I go ahead and close this?

zimmra:
I may have spoken a bit too quickly

Worked fine at a glance, but started getting some errors when flipping through the tables pages (this was thrown once I got to page 9)

```
ERROR [1650653771214] (29 on 4933b8f7488a): Cannot convert undefined or null to object
    TypeError: Cannot convert undefined or null to object
[...]

PR #5542, opened 2022-04-22
Fix for SQL relationship based formulas

mike12345567:
## Description
Fixes for:
1. #5495 - there was an issue where SQL tables wouldn't have access to relational data - the relational data was being squashed to the `{ primaryDisplay: ..., _id: ... }` structure before the formulas were being calculated which meant they could not access any of the relational properties. Also updated the frontend binding generation, when you clicked on a relationship value it was just adding the relational column name, not the path to it e.g. `relatedColumn.0.someColumn` - fixed this.
2. #5530 - Quick fix for the MySQL `RejectUnauthorized` flag.

## Screenshots
The new Reject unauthorized flag for MySQL SSL connections.
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/formula.js
+++ b/packages/builder/src/components/backend/DataTable/formula.js
@@ -60,6 +60,7 @@ export function getBindings({
     )
 
     const label = path == null ? column : `${path}.0.${column}`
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM! Great job getting that wee SSL fix in there too. 👏

PR #5545, opened 2022-04-22
Fixing CouchDB issue

mike12345567:
## Description
Fixing an issue we've experienced with couch, using a base64 cookie that we control manually.


shogunpurple (approved):
LGTM! This is much better.

PR #5558, opened 2022-04-24
Static Formula number data type filtering

### LINKED ISSUE #5541: [BUDI-7138] Allow filtering on different data types for Formulas
**Describe the bug**
At the moment you can only filter on string data types, unless you use a JavaScript binding to return the comparison in the desired type.

**To Reproduce**

1. Create a static formula column with a JavaScript formula that returns the number `100`
2. Define filters on your Data Provider
3. *Static formula column Equals Value 100*
4. The value does not show in the table
5. Edit the filter, change the type to *Binding* and return `100` in a JavaScript function
6. The value now shows

**Expected behavior**
I should not need a JavaScript binding to filter on numbers, booleans, etc. Have the define filter match the data type drop down on the Conditions block.

**Screenshots**

### Add these value types from the Condition block

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7138](https://linear.app/budibase/issue/BUDI-7138/allow-filtering-on-different-data-types-for-formulas)</sub>

melohagan:
## Description
You can now filter on Number types returned from JavaScript Static Formulas.
I didn't include Date types as these appear to be compatible with Binding. 
See issue here: https://github.com/Budibase/budibase/issues/5541

## Screenshots
Screenshot 2022-04-24 at 16 49 12

### DIFF
--- a/packages/builder/src/components/design/PropertiesPanel/PropertyControls/FilterEditor/FilterDrawer.svelte
+++ b/packages/builder/src/components/design/PropertiesPanel/PropertyControls/FilterEditor/FilterDrawer.svelte
@@ -31,10 +31,6 @@
       value: "number",
       label: "Number",
     },
-    {
-      value: "datetime",
-      label: "Date",
-    },
   ]
 
   $: enrichedSchemaFields = getFields(schemaFields || [])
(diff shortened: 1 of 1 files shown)


aptkingston:
Hey Mel. I see where you're coming from with this change, but I don't think this is the best route to go down. At its core, the real "issue" here is that HBS values are strings, and that's a problem that exists everywhere in the platform.

Specifically with formula fields I can see the problem - they can return any type, while a filter expression of "value" (or a HBS binding) will return return a string by default. I think a proper solution for this would involve adding metadata to formula columns to define what type it returns. For every other column type, we know exactly what data type it is and can handle them appropriately. Formula columns are the one exception in that they can return anything. I think if we added schema metadata so say what data type they return (text, number, boolean, date, object etc) then we could automatically handle that when filtering. This would have the added benefit of working for any data type, not just numbers.

As an aside, I actually think it's alright to expect users to write a JS binding to return a number here for filtering, since they've already written a JS expression to create the formula that returns the number - so it's actually just the same experience. I totally appreciate that needs highlighted as a limitation at the moment though, since everyone would expect you could fire a number into that field and it would just work.

melohagan:
That's fair. What I had did feel a bit messy, and the schema metadata definitely sounds better! I'll be interested to see how this works when it's done. 

Not too confident about how many users will think 'let me write a JS function to return a constant' but it's an easy workaround when you know about it.

PR #5568, opened 2022-04-25
adding larger limits to the NGINX ingress contoller

shogunpurple:
## Description
Fix for https://github.com/Budibase/budibase/discussions/5556

The NGINX ingress controller did not have the correct `client_max_body_size` and `proxy_body_size` settings, causing a 413 when someone tried to import a large file (an app import).

## Screenshots

### DIFF
--- a/charts/budibase/Chart.yaml
+++ b/charts/budibase/Chart.yaml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/Budibase/budibase
   - https://budibase.com
 type: application
-version: 0.2.8
+version: 0.2.9
 appVersion: 1.0.48
 dependencies:
   - name: couchdb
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
LGTM! This still will hit the inner server limit of 10mb I believe, we will need to add an env variable to increase this within the server as well.

shogunpurple:
Cheers @mike12345567 - is this just for the public API? Our NGINX configuration in prod should allow for `1000M`

mike12345567:
Its in https://github.com/Budibase/budibase/blob/develop/packages/server/src/app.ts - as part of the `koaBody` setup, although now that I look at it I'm not sure that any of those set limits would affect multi-part uploads, so as long as it isn't sent up as a single body I don't think it would matter!

PR #5569, opened 2022-04-25
Some small automation fixes

### LINKED ISSUE #4922: Automations does not save multiple relationship
I have the following table structure

`One Contact has many Deals`

I then set up an automation to update a contact with multiple deals. For the `Deals` column, I put the following.

```javascript
return [
  "ro_ta_20165c6196654925a9cc74dfec158c3f_2485b71e8e714764b35d890fc63e60bb",
  "ro_ta_20165c6196654925a9cc74dfec158c3f_f3775cf97f774f9a9533c939f163a0e1"
];
```
When I do this, the the Deals column is completely wiped out of any previous data.

The `Input` to the `Update Row` step is below. As you can see, my array is nested inside an array, like: `[  [ "id1", "id2" ] ]`. I have tried a few different methods and can't find a way to assign multiple relationships at once.

```json
{
  "row": {
    "tableId": "ta_5047dfad5b24426ab3fd0b350fe1780c",
    "Contact": [],
    "Deals": [
      [
        "ro_ta_20165c6196654925a9cc74dfec158c3f_2485b71e8e714764b35d890fc63e60bb",
        "ro_ta_20165c6196654925a9cc74dfec158c3f_f3775cf97f774f9a9533c939f163a0e1"
      ]
    ]
  },
  "rowId": "ro_ta_5047dfad5b24426ab3fd0b350fe1780c_dc9b54563ebe4921a68e066a9877868f"
}
```

PClmnt:
## Description
Fixes a couple of issues

- Fixed the bad debounce code that made changing inputs in automation blocks act strangely
- Removed limitation that disallowed you from adding a new block if the old one hadn't been completed, this is handled when testing an automation as you'll get a "invalid inputs" response
-  Also fixed #4922 
- Added an array field to the app action trigger (@Topener)


packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -43,6 +43,11 @@
   }
 
   const coerce = (value, type) => {
+    var re = new RegExp(/{{([^{].*?)}}/g)
```
shogunpurple, line 46: Does this need to be function scoped? Could probably just use `const` here.

shogunpurple (approved):
LGTM!

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
PClmnt: whoops, you're right!

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -54,7 +53,7 @@
   $: schema = getSchemaForTable(tableId, { searchableSchema: true }).schema
   $: schemaFields = Object.values(schema || {})
 
-  const onChange = debounce(async function (e, key) {
+  const onChange = async (e, key) => {
```
aptkingston, line 56: This should probably work fine as long as we don't have rapid saves in succession. In the design section I made a wrapper function that ensures multiple calls to async functions are fired in sequence, letting the first one complete before continuing. The reasoning was that if you had a bad ping, and made 2 changes quickly, the second one would be sent up referencing stale data as the change in the first request hadn't completed yet. Might not be a problem here at all - but if it ever becomes a problem then the change is as easy as:
```js
import { Utils } from "@budibase/frontend-core"

const onChange = Utils.sequential(async (e, key) => {
...
})
```

aptkingston (approved):
LGTM!

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
PClmnt: Updated to use this! Thanks!

PR #5583, opened 2022-04-26
BB logo on free plan

Rory-Powell:
## Description
- Add existing `MadeInBudibase` component to all layouts. 
- Change positioning to fixed so that it always sticks to bottom right 
- Update link styling by using link component
- Integrate showing the logo with the current license
- Only show in dev preview and published apps, not in the builder
- Clicking logo redirects to https://budibase.com/?utm_source=budibase-apps-public-screens&utm_medium=badge&utm_campaign=made-in-budibase

## Screenshots
https://user-images.githubusercontent.com/8755148/165281785-14ed95cb-d3c8-43d0-92c6-a0aae7638616.mov


shogunpurple (approved):
The first premium feature! So good to see. LGTM

daawaan:
Hello, does this show in self-hosted instances or only in the Budibase Cloud?

Rory-Powell:
Hello @daawaan, the logo will show in cloud and self host. When this goes live users will have the option to upgrade their plan to remove the logo, increase usage quotas (cloud) and access future premium features.

aptkingston (commented):
Nice one Rory!

NAB, but the only thing I would say is that the new licensing data sent down looks like a perfect candidate for a derived svelte store. We could condense that entire new `licensing` directory into a single `src/stores/licensing.js` derived svelte store, which simply takes in the auth store and derives licensing features - which allows us to safely specific fallbacks like defaulting to the free plan, as well as additional fields like `logoEnabled`.

It would also have the benefit of then actually being a svelte store with reactivity (not that this is likely to change at runtime anyway), and would follow the pattern of the rest of the client stores for consistency.

Also NAB but should we make the text black rather than blue?

Rory-Powell:
Sounds good @aptkingston! all great suggestions. I think as well it would be good for us to move this into frontend-core so we can re-use across builder and client.

I had a go with the black and it didn't look at good imo, I'll post comparisons when I come back to this

daawaan:
Hello @aptkingston @Rory-Powell, the Budibase logo being present in the self-hosted version which can only be removed by buying a plan feels a bit concerning to me. Although it seems the logo can be removed by modifying, building, and deploying Budibase directly from source, I'm more concerned as to what does this mean to the future of Budibase since it claims itself as an open-source low-code platform?

aptkingston:
> Sounds good @aptkingston! all great suggestions. I think as well it would be good for us to move this into frontend-core so we can re-use across builder and client.
> 
> I had a go with the black and it didn't look at good imo, I'll post comparisons when I come back to this

Yea frontend-core probably does make sense. There are probably a few different stores that could be shared, but at the moment everything is still separate. The stores usually do end up being different between the client and builder, but perhaps licensing is a good candidate to be first shared store.

shogunpurple:
Hi @daawaan - I know you deleted your original comment but I just wanted to get your thoughts. You said this:
> I'm more concerned as to what does this mean to the future of Budibase since it claims itself as an open-source low-code platform?

I'm interested as to what you mean by being concerned about budibases future as an "open source low code platform".  Budibase is, and always will be, open source - however open source does not mean totally free, and we plan to introduce paid features and plans in the future. I'm just wondering what exactly you meant by that comment, out of curiosity. I just wanted it to be clear that budibase will not stop being open source, if that was your concern.

mjashanks:
Hi @daawaan 

> I was wondering if Budibase as an open-source project will eventually turn into a partially open-source low-code platform with some closed-source paid features

This will indeed be our path. We will not be pulling any existing features out of the open-source platform, and we will continue to develop open-source features. 

However, we will also be building closed-source (paid) features.

PR #5586, opened 2022-04-26
add storageClass to redis and objectStore PVCs

atalakey4work:
## Description
Problem: PVC storageClass option not available in helm chart
Solution: modified redis and objectStore PVC templates to avail option to set storageClassName


shogunpurple (approved):
Thanks for the contribution! This is a useful one. 👍

Issue #5269, opened 2022-04-06
Adding custom column to Users table results in blank Created By & Updated By columns in linked table


melohagan:
**Describe the bug**
Discussion here: https://github.com/Budibase/budibase/discussions/5260
Adding custom columns in the user table breaks the auto-generated Created By / Updated By fields in other internal DB tables.

**To Reproduce**
1. Create new app with a new table with one text column and all of the auto-generated columns
2. Create the default screens for this table
3. Use the app to create a new row. You will see the *Created By* field is populated.
4. Now add a custom text column to the Users table
5. Use the app to create a new row. You will see the *Created By* field is NOT populated.
6. Delete the custom text column from the Users table
7. Use the app to create a new row. You will see the *Created By* field is still NOT populated.
8. You need to delete the new table and re-create it.

**Expected behavior**
Created By / Updated By should still work even when custom columns have been added to the Users table

**Screenshots**
*Created By* field present in the */rows* response when it works.
![Screenshot 2022-04-06 at 11 18 00](https://user-images.githubusercontent.com/101575380/161954275-8dfb4d8e-c102-4085-b75e-464eb2382e3b.png)

*Created By* field missing in the */rows* response when it does not work.
![Screenshot 2022-04-06 at 11 19 28](https://user-images.githubusercontent.com/101575380/161954420-b07b6867-c594-4796-a848-26a475ffc9eb.png)

xnohat:
Hi everyone,

Does we have any way to workaround or tempolary fix for this bug ?
Or any maintainer can guess what part of code make this problem ? 
I will try to contribute a fix for this. I have dig into code of POST /api/tables but not found the root cause of this bug
It's just change table schema from
```
"Created By":{
               "name":"Created By",
               "type":"link",
               "subtype":"createdBy",
               "icon":"ri-magic-line",
               "autocolumn":true,
               "constraints":{
                  "type":"array",
                  "presence":false
               },
               "tableId":"ta_users",
               "fieldName":"testtable-Created By",
               "relationshipType":"many-to-many"
            },
```
To
```
"Created By":{
               "name":"Created By",
               "type":"link",
               "tableId":"ta_users",
               "fieldName":"testtable-Created By",
               "relationshipType":"many-to-many",
               "autocolumn":true
            },
```
and delete all Relationship from TableName:Created By to Users: TableName-Created By column

PR #5592, opened 2022-04-26
Fixing created by/updated by columns after updating user table

mike12345567:
## Description
Fix for #5269 - the sub type was being lost in some scenarios related to the user table being updated - making sure it is consistently kept when dealing with either of the related tables.


shogunpurple (approved):
LGTM, nice fix!

xnohat:
Thanks for more completed fix

PR #5595, opened 2022-04-26
Small changes for Smoke Build

Mitch-Budibase:
- using data-cy id for createScreen command
- added 1 second wait for selectExternalDatasource command - assists with timing on smoke build
- Correct usage for HOST within addDatasourceConfig command


shogunpurple (approved):
LGTM!

PR #5597, opened 2022-04-26
Fix: Update global db context when tenant is updated + configurable couch integration in tenancy middleware

Rory-Powell:
## Description
- Add `setGlobalDB` to `updateTenantId` to make sure the correct global db is being accessed (fixes login)
- Add option `USE_COUCH` to turn off global db initialisation when using tenancy middleware (for account portal)


shogunpurple (approved):
LGTM! Like the change to disable env vars too.

Issue #5522, opened 2022-04-21
App and Role Assignment lost on Unpublish


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5508

<div type='discussions-op-text'>

<sup>Originally posted by **fwit-carson** April 20, 2022</sup>
If I create users while the app is published, when I unpublish the app, all the roles are lost. When I publish the app again, roles do not come back.</div>

Issue #5219, opened 2022-04-01
Text fields created in Postgres break if 'Fetch Tables' is used


sudpuzzer:
**Describe the bug**
When creating columns in Budibase on an external Postgres 13.6 database, a normal text column (not long form text) is created in Postgres with the TEXT type. If the Fetch Tables button is used on the data source configuration page, Budibase reads the TEXT column in as a Long Form Text field, and it is not possible to change it back to a normal text field. This breaks any forms and field groups that are expecting a the text field type.

**To Reproduce**
1. Create a new column in a Postgres table in the Budibase table editor.
2. Select the 'Text' type.
3. Use the 'Fetch Tables' button in the data source configuration page for PostgresSQL
4. Verify that the newly created column is now a 'Long Form Text' type, and it is not possible to change.

**Expected behavior**
The normal text type should be created in Postgres as a VARCHAR column, which Budibase correctly identifies as the normal text type.

**Screenshots**
Create column:
![image](https://user-images.githubusercontent.com/10662820/161323922-5a9f17a0-d362-4e64-b139-72b07011ce20.png)

Verify TEXT type:
![image](https://user-images.githubusercontent.com/10662820/161323975-4e0dde64-6779-4636-a995-07723c34a7aa.png)

Fetch tables:
![image](https://user-images.githubusercontent.com/10662820/161324000-6692dee4-6d44-4801-b9ab-98ec44a89af0.png)

Column is now Long Form Text:
![image](https://user-images.githubusercontent.com/10662820/161324030-659e1c8b-0607-4790-945a-70c0fd6f1fc7.png)

Change column to varchar(255) in database:
![image](https://user-images.githubusercontent.com/10662820/161324100-b8957563-120c-4c49-a6c2-1f4b547db29c.png)

Fetch tables again, column back to text:
![image](https://user-images.githubusercontent.com/10662820/161324168-432bc2a1-722c-4d51-a149-26c87214f77a.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 99.0.4844.84
 - PostgresSQL: 13.6
 - Budibase: 1.0.98

**Additional context**
Specifying a max length for the does not make a difference either, when the Fetch Tables button is used, the Max Length value is discarded and behaves as above. A text field with a Max Length should certainly be created as a VARCHAR(255) in the PostgresSQL database.

mike12345567:
[...]

PR #5616, opened 2022-04-27
Various April release fixes

mike12345567:
## Description
1. Fixes the issue that caused deletion of apps/unpublish to throw an error (close DB was being called after a destroy and the error wasn't captured).
2. Fix for #5522 - making sure that app roles aren't removed unless the development app is being deleted (all trace of that app ID is being removed from the system) - unpublish no longer does anything to roles.
3. Fix for #5219 - most people expect all text types to be our standard text type, making it so that our SQL type conversion converts all string types to our basic text type. Made it so that you can switch a text type to longform if desired (as well as options).
4. Fixing an issue with the new table UI, filter/search button was disabled if the search returned no rows, requiring a page refresh to search again.
5. Fixing some issues with SQL fetch tables wiping changes made to the schema within Budibase, e.g. adding a formula column (https://github.com/Budibase/budibase/discussions/5539#discussioncomment-2629322).

### DIFF
--- a/packages/backend-core/src/db/pouch.js
+++ b/packages/backend-core/src/db/pouch.js
@@ -1,31 +1,53 @@
 const PouchDB = require("pouchdb")
 const env = require("../environment")
 
-exports.getCouchInfo = () => {
-  let url = "http://localhost:4005"
-  if (env.COUCH_DB_URL && env.COUCH_DB_URL.includes("@")) {
-    url = env.COUCH_DB_URL
-  } else if (env.COUCH_DB_URL) {
-    const [protocol, ...rest] = env.COUCH_DB_URL.split("://")
-    url = `${protocol}://${env.COUCH_DB_USERNAME}:${env.COUCH_DB_PASSWORD}@${rest}`
-    if (!env.COUCH_DB_USERNAME || !env.COUCH_DB_PASSWORD) {
-      throw new Error(
-        "CouchDB configuration invalid. You must provide a fully qualified CouchDB url, or the COUCH_DB_USER and COUCH_DB_PASSWORD environment variables."
-      )
-    }
-  }
+function getUrlInfo() {
+  let url = env.COUCH_DB_URL
+  let username, password, host
   const [protocol, rest] = url.split("://")
-  const [auth, host] = rest.split("@")
-  let [username, password] = auth.split(":")
-  if (!username && env.COUCH_DB_USERNAME) {
-    username = env.COUCH_DB_USERNAME
+  if (url.includes("@")) {
+    const hostParts = rest.split("@")
+    host = hostParts[1]
+    const authParts = hostParts[0].split(":")
+    username = authParts[0]
+    password = authParts[1]
+  } else {
+    host = rest
   }
-  if (!password && env.COUCH_DB_PASSWORD) {
+  return {
+    url: `${protocol}://${host}`,
+    auth: {
+      username,
+      password,
+    },
+  }
+}
+
+exports.getCouchInfo = () => {
+  const urlInfo = getUrlInfo()
+  let username
+  let password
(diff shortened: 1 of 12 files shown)


packages/backend-core/src/db/pouch.js
```diff
@@ -1,35 +1,36 @@
...
-    return env.COUCH_DB_URL
-  }
-
-  const [protocol, ...rest] = env.COUCH_DB_URL.split("://")
-
-  if (!env.COUCH_DB_USERNAME || !env.COUCH_DB_PASSWORD) {
-    throw new Error(
-      "CouchDB configuration invalid. You must provide a fully qualified CouchDB url, or the COUCH_DB_USER and COUCH_DB_PASSWORD environment variables."
-    )
+exports.getCouchInfo = () => {
```
Rory-Powell, line 4: Great that everything now uses a consistent connection config 👍 

We could probably make this function more readable, I found the current flow (construct auth aware url -> destruct auth aware url -> fallback to env vars) a bit hard to understand, but I get why it's done like that so the lower half of the function operates on the same url structure. 

Rather than constructing / destructing the url, we could prefer the auth vars first and fallback to the url auth if provided. I found this to be more readable and could look like:

```js
/**
 * Return the couch db connection config. 
 * Optionally include the username and password if present.
 */
const getConectionConfig = () => {
  let url = env.COUCH_DB_URL
  let username
  let password
  let host
  const [ protocol, ...rest] = url.split("://")

  if (url.includes("@")) {
    const hostParts = rest.split("@")
    host = hostParts[1]
    const authParts = hostParts[0].split(":")
    username = authParts[0]
    password = authParts[1]
  } else {
    host = rest
  }

  return {
    url: `${protocol}://${host}`,
    auth: {
      username,
      password
    }
  }
}

/**
 * Construct the couch db connection configuration. 
 * Always prefer the authentication provided from environment variables, 
 * fallback to authentication provided by the url if not present.
 */
exports.getCouchInfo = () => {
  const url = getConectionConfig()

  let username
  let password

  if (env.COUCH_DB_USERNAME) {
    // set from env 
    username = env.COUCH_DB_USERNAME
  } else if (url.auth.username) {
    // set from url
    username = url.auth.username
  } else {
    throw new Error("CouchDB username not set")
  }

  if (env.COUCH_DB_PASSWORD) {
    // set from env
    password = env.COUCH_DB_PASSWORD
  } else if (url.auth.password) {
    // set from url
    password = url.auth.password
  } else {
    throw new Error("CouchDB password not set")
  }

  const authCookie = Buffer.from(`${username}:${password}`).toString("base64")

  return {
    url: url.url,
    auth: {
      username: username,
      password: password,
    },
    cookie: `Basic ${authCookie}`,
  }
}
```

Could also move the default of the localhost url out to the env like 
```js
COUCH_DB_URL: process.env.COUCH_DB_URL | "http://localhost:4005"
```

packages/worker/yarn.lock
```diff
@@ -321,12 +321,12 @@
     uuid "^8.3.2"
     zlib "^1.0.5"
 
-"@budibase/pro@1.0.124-alpha.0":
-  version "1.0.124-alpha.0"
-  resolved "https://registry.yarnpkg.com/@budibase/pro/-/pro-1.0.124-alpha.0.tgz#6287a51fa7c19754e44374c209c4aa3480fc3ac9"
-  integrity sha512-EgMuh+XSd/9tb3Ej9EZa4Y8hgiS6fHG+tuUwUcTuP6zvHbTijQGPb9075yImUbSc10bS3o41AP2qa2/ZdZKV2w==
+"@budibase/pro@1.0.127":
```
Rory-Powell, line 324: Looks like these aren't getting bumped in the release job, I'll get it updated to include the lock file in the commit

Issue #5365, opened 2022-04-12
Validation in form can't be disabled by conditions


Upgreydd:
**Describe the bug**
I have conditional field with YES/NO values:
If NO - render second field and make it required
if YES - hide field and do not require it
I'm able to set require flag on if condition but I'm not able to unrequire after that and so if I check NO, YES then field is required.

![image](https://user-images.githubusercontent.com/579481/162939654-f46ad537-8470-4a1b-aee3-2f757b6b0ed6.png)

Second validation was added to try clear all validation conditions and is empty but it doesn't work cause form is only appending requirements.

PR #5637, opened 2022-04-28
Deregister field validation when unmounting form fields

aptkingston:
## Description
Field validation is now removed when fields are unmounted (e.g. due to conditional UI). We don't want to actually remove fields from form state when they're unmounted - so this is the minimum amount required to just prevent hidden fields from causing validation errors.

Fixes https://github.com/Budibase/budibase/issues/5365.


upgreydd (approved):
LGTM :+1:

mike12345567 (approved):

PR #5638, opened 2022-04-28
Templates Details Tests

Mitch-Budibase:
Added tests which check the details button for each template.
- Split per category (Will add more specific template tests under each category)
- Checks template name matches url (There are a few exceptions)
- Checks status of url - expects 200 back each time

Also updated the DeleteApp Function
- Fixed a smoke test issue associated with deleting the tests app


shogunpurple (approved):
LGTM! Awesome work getting these templates tested 👍

PR #5642, opened 2022-04-28
(draft) backend-core typescript building successfully

Rory-Powell:
## Description
- backend-core typescript support
- convert some js files to ts, where the auto generated d.ts files were failing to build with server / worker
   -  packages/backend-core/src/db/Replication.ts
   - packages/backend-core/src/db/utils.ts
   - packages/backend-core/src/objectStore/index.ts

### DIFF
--- a/packages/backend-core/src/db/utils.ts
+++ b/packages/backend-core/src/db/utils.ts
@@ -1,47 +1,26 @@
-const { newid } = require("../hashing")
-const Replication = require("./Replication")
-const { DEFAULT_TENANT_ID, Configs } = require("../constants")
-const env = require("../environment")
-const {
-  StaticDatabases,
-  SEPARATOR,
-  DocumentTypes,
-  APP_PREFIX,
-  APP_DEV,
-} = require("./constants")
-const { getTenantId, getGlobalDBName } = require("../tenancy")
-const fetch = require("node-fetch")
-const { doWithDB, allDbs } = require("./index")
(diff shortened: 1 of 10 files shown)


packages/backend-core/src/db/utils.ts
```diff
@@ -1,47 +1,26 @@
...
   USER_BY_EMAIL: "by_email",
   BY_API_KEY: "by_api_key",
   USER_BY_BUILDERS: "by_builders",
 }
 
-exports.StaticDatabases = StaticDatabases
-
-exports.DocumentTypes = DocumentTypes
-exports.APP_PREFIX = APP_PREFIX
-exports.APP_DEV = exports.APP_DEV_PREFIX = APP_DEV
```
Rory-Powell, line 38: @mike12345567 not sure if I've converted this properly, I don't follow what the double `=` is doing. Below I've just re-exported conversions and contants for handiness. Probably we move `db/utils.ts` (and associated  Replication, constants, conversions) into -> `db/utils` and introduce `db/utils/index.ts` which will export things at the `utils` module level

mike12345567: The double equals is basically just equivalent to `a = b = 1` which means that `a = 1, b = 1`, essentially setting them all in one flow to the same value.

Rory-Powell: ah great, thanks!

Rory-Powell:
closing this - finished on the backfill branch

Issue #5487, opened 2022-04-19
Allow budibase object store bucket names to be configurable through environment variables


shogunpurple:
### Discussed in https://github.com/Budibase/budibase/discussions/5430

<div type='discussions-op-text'>

<sup>Originally posted by **dougwettlaufer** April 14, 2022</sup>
In the helm chart, under `objectStore` I see references to `region` and `AWS_ACCESS_KEY` but I'm a little lost on setting the bucket. From the looks of things it seems like the bucket names are hardcoded in utils.js

```js
exports.ObjectStoreBuckets = {
  BACKUPS: "backups",
  APPS: "prod-budi-app-assets",
  TEMPLATES: "templates",
  GLOBAL: "global",
  GLOBAL_CLOUD: "prod-budi-tenant-uploads",
}
```

Since a bucket in S3 must be globally unique this won't work. Please excuse my lack of JS knowledge but from grepping through the code it doesn't seem like there's a way to override the bucket name or add a unique prefix. Can someone please confirm that's the case?</div>

MartinPicc:
I tried using newly created ENV variables to set custom names for buckets, without success.

I'm using docker compose install, here is what I did:
- In `.env` file, I have defined `APPS_BUCKET_NAME` to a value different from the default (default value is `prod-budi-app-assets`)
- then I start the stack and create a new app
- when I browse the APPS_BUCKET_NAME bucket, I can see the subpath `app_<appid>` for the newly created app with `budibase-client.js` file inside
- however, when I open my app and then go to the "Design" tab, I see the error "App preview failed to load":
![image](https://user-images.githubusercontent.com/35298861/194719250-02c7fe1a-cc2d-491f-98e6-e119a45bf40f.png)
and if I open the console of my browser, I have HTTP 403 Forbidden error for the following request: `GET https://fra1.digitaloceanspaces.com/prod-budi-app-assets/app_5047077b270842688fc618119725067d/budibase-client.js?v=2.0.3`

It tries to download the file from the bucket `prod-budi-app-assets`, despite the env variable `APPS_BUCKET_NAME` having a different value. So it seems that this bucket name (and maybe others) is still hard-coded somewhere in the code, hence it tries to download the file from the wrong bucket. Especially I found this suspect line: https://github.com/Budibase/budibase/blob/270836b30ef08c89f172c0394ed712fc2e28d439/packages/server/src/constants/index.js#L177

[...]

Issue #4944, opened 2022-03-18
Add Firebase test cases


mike12345567:
**Describe the feature request**
With the introduction of the standard datasource for Firebase, we should add test cases for it like we have for all other datasources - these will likely require mocking out Firebase, they are simply a test of the datasource performing its expected actions towards to Firebase API.

Firebase PR - #4864

Issue #4790, opened 2022-03-04
Inviting user email field error & incorrect label


AndyWhann:
### Problem

1. On the option to invite a basic user, the label says `Username` when it should say `Email address`

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/ea1ce519-16d5-46b9-b460-fcd768cdbe9b)

2. We currently show an error for an invalid email before the user tries to submit an invalid email on both user invite options. It's not a major issue but just not a nice UX.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/0eefea6a-da2e-4208-a006-a1b8a51a9400)

### Solution

Fix the label
Don't show this error until the user submits an invalid email

mslourens:
The onboarding label will be changed in PR #4639

Issue #3707, opened 2021-12-07
Postgres connection pooling issues


marco-santulli:
Author: @Rory-Powell 

## Description
The postgres integration currently creates a new pool with each request due to how the integration object is constructed. 
See debugging details here: https://github.com/Budibase/budibase/pull/3301#discussion_r749229022

## COS
In single tenant environments the postgres datasources pools should be created outside the scope of the postgres integration. The number of connections to a database should not exceed the configured pool size. 

Investigate how this may be feasible in a cloud environment where potentially thousands of pools would have to exist. 

## Context

Author: @marco-santulli

Hi,

does budibase not use connection pooling to the database?  I got an error on the number of transactions in the log, I upped the number of transactions to 100 (note there is just one developer at the momento on it), and got this report from the DB:

<img width="685" alt="image" src="https://user-images.githubusercontent.com/82715690/145048035-17f13d22-3e68-40a5-9596-e76ee262cb33.png">

to give context, I am running a simple query that selects one column from one table with two rows in it.  It getting called multiple times (every time a row in a table is rendered).  Maybe there is some setting to use connection pools?

M

Rory-Powell:
Hi @marco-santulli, which database datasource are you using?

marco-santulli:
> Hi @marco-santulli, which database datasource are you using?

sorry I should have mentioned that, it is postgres

Rory-Powell:
Thanks marco. We do have pooling with our postgres integration but there is an issue at the moment which came up recently in this thread https://github.com/Budibase/budibase/pull/3301#discussion_r749229022

The bigger issue here seems to be that the connections aren't being released, I'll rename this issue towards getting fixes in for postgres on this

marco-santulli:
> Thanks marco. We do have pooling with our postgres integration but there is an issue at the moment which came up recently in this thread [#3301 (comment)](https://github.com/Budibase/budibase/pull/3301#discussion_r749229022)
> 
> The bigger issue here seems to be that the connections aren't being released, I'll rename this issue towards getting fixes in for postgres on this

Hi Rory,

sorry I think that diagram was misleading, I apologize.. That is the total number of new connections requested.  Connections are getting released, look at this:
total number of active connections:
<img width="719" alt="image" src="https://user-images.githubusercontent.com/82715690/145057344-b17e8f6d-5700-4b1d-b128-61337bc0b150.png">
 
total number of "succeeded" connections:
[...]

PR #5643, opened 2022-04-28
Feature/small april tickets

shogunpurple:
## Description
This PR addresses a few smaller tickets for April.

- https://github.com/Budibase/budibase/issues/4982
Our row processing code was relying on a truthy value, so this actually threw away `false` booleans on our rows, and returned them as `undefined`. I've updated `ExternalRequest` to allow for `false` and only chuck `null` or `undefined` values in the rows.

- https://github.com/Budibase/budibase/issues/4944
Unit tests for completeness in firebase integration.

- https://github.com/Budibase/budibase/issues/5487
Added environment variables for S3 bucket names to allow users who are using their own S3 and not MinIO to configure it in self host.

- https://github.com/Budibase/budibase/issues/4790
Small copy change to user onboarding.

- https://github.com/Budibase/budibase/issues/3707
Removed pooling from the postgres connector, and just went back to manual connection management with a `try` `catch` `finally` block.

## Screenshots


mike12345567 (approved):
LGTM - good fixes and nice to expand the testing, also good to move away from a pooled approach, I think this is the right decision for Budibase as a whole, ephemeral singular connections are the only way it can work for Cloud, maybe sometime in the future we can provide variables for self hosters to re-use connections, but its totally non-critical; just a performance enhancement.

packages/server/src/integrations/postgres.ts
```diff
@@ -171,16 +166,17 @@ module PostgresModule {
       } catch (err) {
         // @ts-ignore
         throw new Error(err)
+      } finally {
+        await this.client.end()
```
Rory-Powell, line 170: should the `buildSchema` call end as well? 

As far as I know a new instance of this class is initialized with every request, so we might create the client, set schema, connect and get the schema but not end. 

We could nearly take a similar approach to what @mike12345567 did recently with the pouch connections, where we could have:

```js
const doWithClient async (fn) {
   const client = new Client(config) 
   client.connect()
   client.query(`SET search_path TO ${this.config.schema}`)
   try {
         await fn(client)
   } finally {
      client.end()
   }
}
```

Additionally removing `this.client` to prevent any accidental misuse. Then wrap the bodies of `buildSchema` and `internalQuery` with `doWithClient`

mike12345567: Good catch, didn't notice this, there is a chance of the same client being used after the `buildSchema` call - it'd probably be better if we wrapped where we used it with a connect then disconnect, rather than just connecting as part of the constructor (`setSchema` function).

packages/server/yarn.lock
```diff
@@ -1091,12 +1119,128 @@
     svelte-flatpickr "^3.2.3"
     svelte-portal "^1.0.0"
 
-"@budibase/pro@1.0.127":
-  version "1.0.127"
-  resolved "https://registry.yarnpkg.com/@budibase/pro/-/pro-1.0.127.tgz#a8bcffb8ccc6afde64370b3a3dc22e2d0dd04f46"
-  integrity sha512-dj0SFTmO8JuMQ97/Ik6jVPQsh9AW7U5Wkgpa4yeNfwWw3DvSoktCxpeZ9mND6BR/DWTaeZljFKsQ6uk6nimzdQ==
+"@budibase/bbui@^1.0.131":
```
Rory-Powell, line 1122: think you need to run `yarn clean && yarn bootstrap` to remove these lerna dependencies. did you run a `yarn add` or do a manual `yarn install` in server? seems that is the scenario that brings them in

packages/server/src/integrations/postgres.ts
shogunpurple: Good catch - it should. Decided not to refactor to the `doWithClient` as it's just a bit of an additional wrapper in this case and the client gets initialised as part of the `setSchema` call - so just added a `close` to `buildSchema`

packages/server/yarn.lock
shogunpurple: Yeah, I ran it to solve the merge conflicts. This is something we've ran into quite a bit recently - might be nice for us to automate it in some way before merging

Issue #5649, opened 2022-04-29
Automate,  Drawer not overlaying properly


MihailHadzhiev2022:
Env: PreProduction

Steps to reproduce: 

1. Create an app, using Budibase BD
2. Add 1 column, and 1 row
3. Click automate and select Row created, add test data

Short video : https://www.loom.com/share/2dc61a1c562f4372b398f5d4a16525c1

expected result: The drawer should pop up and overlay properly.
actual result: The drawer not overlaying properly

shogunpurple:
Looks like https://github.com/Budibase/budibase/pull/5652 addresses this one, closing

PR #5652, opened 2022-04-29
Automation binding drawer for test data was not showing correctly

PClmnt:
As described in: #5649 the binding panel for test data was not showing correctly, and was using the `DrawerBindableInput` instead of `ModalBindableInput` which meant the user wasn't able to use handlebars within the test modal. 

Also refactored a dodgy if statement in the same area to use `svelte:component` instead.


shogunpurple (approved):
LGTM!

PR #5684, opened 2022-05-03
Pass through type when creating internal table

### LINKED ISSUE #5444: Relationship options are empty after initial table creation
**Describe the bug**
When you create a new Internal Table, you cannot select relationships. You have to click onto another table and then come back to add your relationship column

**To Reproduce**
1. Create an app, and add a table
2. Create a second table, and create a new *Relationship* column
3. There are no table options available in the Table dropdown
4. Click onto a different table, and back to the second table.
5. You can now add a relationship column

**Expected behavior**
I should be able to create a new table with relationship columns straight away.

**Screenshots**
Screenshot 2022-04-15 at 15 57 46

melohagan:
## Description
On initial create, Internal DB did not show relationship options: https://github.com/Budibase/budibase/issues/5444

Now passing through a default type and sourceId for the *CreateTableModal* because as far as I can see, this modal is only ever used for internal tables.

### DIFF
--- a/packages/builder/src/components/backend/TableNavigator/modals/CreateTableModal.svelte
+++ b/packages/builder/src/components/backend/TableNavigator/modals/CreateTableModal.svelte
@@ -45,6 +45,8 @@
       name,
       schema: addAutoColumns(name, dataImport.schema || {}),
       dataImport,
+      type: "internal",
+      sourceId: "bb_internal",
     }
 
     // Only set primary display if defined


melohagan:
Without this, the type defaults to "table" and the sourceId *undefined*, so there is no match.
Is this comment out of date?

```
// in the case of internal tables the sourceId will just be undefined
  $: tableOptions = $tables.list.filter(
    opt =>
      opt._id !== $tables.draft._id &&
      opt.type === table.type &&
      table.sourceId === opt.sourceId
  )
```

PR #5688, opened 2022-05-03
Add test:e2e:ci:record variant

Rory-Powell:
## Description
- Record cypress runs on smoke tests only


shogunpurple (approved):
LGTM - makes sense to separate them.

PR #5689, opened 2022-05-03
Query preview - sort by number for number types

### LINKED ISSUE #5347: Query Preview sorting does not update with chosen data type
**Describe the bug**
Whenever you create a MySQL query and pull back data for a numeric column, you can change the value data type between text and number. This effects the sorting in the builder, however the preview tab does not update to reflect the new sorting. 
Discussion here: https://github.com/Budibase/budibase/discussions/5301#discussioncomment-2538730

**To Reproduce**
1. Create a MySQL table with Int column, and Varchar column, both with number values
2. Create a query in budibase that pulls back this data
3. Value data type defaults to Text, but in the Preview the sorting will adhere to the data type as determined in the MySQL schema
4. Change the data types to Number, the preview sorting will not change.  

**Expected behavior**
The preview tab sorting for a query should reflect the data types of the schema tab in budibase, not the database schema. 

**Screenshots**
Should be Number based sorting:
Screenshot 2022-04-11 at 09 59 52

Uses Text based sorting in the Preview (the builder sorting is correct however):
Screenshot 2022-04-11 at 10 00 31

melohagan:
## Description
Sorting in the query preview did not respect the number typing of the schema. Updated to cast the query field to number type if needed.
https://github.com/Budibase/budibase/issues/5347

### DIFF
--- a/packages/builder/src/components/integration/QueryViewer.svelte
+++ b/packages/builder/src/components/integration/QueryViewer.svelte
@@ -44,6 +44,21 @@
   $: readQuery = query.queryVerb === "read" || query.readable
   $: queryInvalid = !query.name || (readQuery && data.length === 0)
 
+  //Cast field in query preview response to number if specified by schema
+  $: {
+    for (let i = 0; i < data.length; i++) {
+      let row = data[i]
+      for (let fieldNameIndex in Object.keys(fields)) {
+        let fieldName = Object.keys(fields)[fieldNameIndex]
+        if (fields[fieldName] === "number" && !isNaN(Number(row[fieldName]))) {
+          row[fieldName] = Number(row[fieldName])
+        } else {
+          row[fieldName] = row[fieldName]?.toString()
+        }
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):
LGTM!

packages/builder/src/components/integration/QueryViewer.svelte
```diff
@@ -44,6 +44,21 @@
   $: readQuery = query.queryVerb === "read" || query.readable
   $: queryInvalid = !query.name || (readQuery && data.length === 0)
 
+  //Cast field in query preview response to number if specified by schema
+  $: {
+    for (let i = 0; i < data.length; i++) {
+      let row = data[i]
+      for (let fieldNameIndex in Object.keys(fields)) {
```
mike12345567, line 51: Just a minor comment, you can do `for (let fieldName of Object.keys(fields))` to remove the need for the next line.

Issue #5669, opened 2022-04-30
Form Steps freeze Builder in V.1.0.132


poirazis:
**Hosting**

- Self
  - Method: Self, Docker, Mac Silicon
  - Budibase Version: 1.0.132

**Describe the bug**
Add a new blank screen
Add a form
Add a form Step.
Builder Preview window no longer responds and the Preview doesn't t get rendered.
Navigating to another screen also stops working.
( Also opening an existing screen with a form and multiple steps causes the same freeze on the preview panel)

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOSX Monterey
 - Browser Safari
 - Version Latest

**Additional context**

Rory-Powell:
Hi @poirazis, thanks for raising this. `1.0.132` is not yet a publicly supported version, the current `latest` tag is `1.0.110`. Can you try with this version?

poirazis:
With 110 works fine. 

Just wanted to give you guys a heads up with 132 and 133 ( problem persists ) during your testing

Rory-Powell:
very much appreciated @poirazis, thanks for the heads up here!

poirazis:
happy to help @Rory-Powell . Also have a look at application Export in .133. It seems to be broken.

poirazis:
Hwy @Rory-Powell , @mike12345567 i think i spotted another one, again in .133

if you try to set a filter on a table column that contains a space you get an error (both in design and runtime).

<img width="1101" alt="Screenshot 2022-05-03 at 6 14 30 PM" src="https://user-images.githubusercontent.com/80512170/166482560-7532b9c9-a317-4705-95ad-c3248b638c07.png">

mike12345567:
Hi @poirazis - thanks for the report, I'm unable to replicate this, could you give me a bit more instruction/a walk-through of what you did to create the error scenario?

I've done as follows:
![image](https://user-images.githubusercontent.com/4407001/166484961-7f10e24e-8e15-4b9e-8103-6cac19d9c6ba.png)

Table has column `space name`, I've created a form, created a text field and then created a filter in the data provider for the `space name` column equalling that of the text field value from the form.

poirazis:
Hey @mike12345567 
it seems to happen only when the condition is "Like" not "Equals".
Delete your Filter and recreate the filter with the condition "Like"

Another thing while trying to recreate this, is that updating the type of the filter from Equals to Like doesn't take effect.
You have to remove the filter, save, and recreate.

[...]

Issue #5612, opened 2022-04-27
Texts with "A-1-1" structure is automatically converted into date type in MySQL data source


apsonandtesla:
**Describe the bug**
Texts with "A-1-1" (alphabetic-number-number) structure is automatically converted into date time type when writing to a text field in MySQL data source.

**To Reproduce**
1. Create a data table in MySQL with a text column (varchar).
2. Use the MySQL data table created in _**Step 1**_ to Budibase as a data source.
3. Add a row in Budibase Data View with the text field set to "A-1-1".
4. You can see that the "A-1-1" has been converted to "2001-01-01 00:00:00.000" in Budibase record (as well as the record in MySQL database).

**Expected behavior**
The text "A-1-1" entered in text field should never be converted into date time type, it should remains "A-1-1" instead.

**Screenshots**

![Enter "A-1-1" to the text field](https://user-images.githubusercontent.com/42469315/165517645-db28762d-afca-4fdd-81ae-2e65de2176de.png)

![The "A-1-1" is converted to a date time type](https://user-images.githubusercontent.com/42469315/165517617-50d5c304-ba67-45e6-97de-b14c0b261d07.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 99.0.1

**Additional Information**
If you are using a Budibase DB data source, the behaviors are correct.

melohagan:
@mike12345567 I think this is related as well: https://github.com/Budibase/budibase/discussions/5639

mike12345567:
Hi @melohagan - thanks yes I solved for that as part of the issue thanks for referencing it!

Issue #5594, opened 2022-04-26
MySQL datasource does not work with table names containing __


deltasquare4:
**Hosting**
- Self
  - Method: k8s
  - Budibase Version: 1.0.108

**Describe the bug**
Importing MySQL datasource that contains table names with double underscores (`__`) - for example, `dev__users` leads to 404 error with the following response: `{"message":"missing","status":404}`. I believe it's because double underscores are used internally by budibase to separate datasource id and table name from the URL. See screenshot below:

**To Reproduce**
1. Open DevTools -> Network in the browser window.
2. Create a MySQL data source with a table named `dev__users` (or anything with `__` really)
3. Fetch all tables and click on the table `dev__users` from the source browser sidebar. 
4. You should have gotten the silent error, and should be able to observe the request in browser DevTools Network tab.

**Expected behavior**
The table should load the data like the rest of them.

**Screenshots**
![2022-04-26_19-58](https://user-images.githubusercontent.com/205111/165323646-4860433b-6e69-49df-b005-b6e67aa2c10a.png)

**Desktop (please complete the following information):**
 - OS: Linux Desktop
 - Browser: Chrome
 - Version: 91

Issue #5553, opened 2022-04-23
MySQL DATE data type is imported incorrectly


pbleic:
**Hosting**

**Describe the bug**
Budibase adds 20:00 to DATE data type when importing from MySQL

**To Reproduce**
1. Create a DATE column in a MySQL database.
2. Confirm that a SELECT query on the DATE column doesn't contain any information about time.
3. Confirm that a SELECT DATE_FORMAT(date_column_name, '%Y-%m-%d %H:%i:%s') FROM test_table gives output with 00:00:00 appended
4. Use the database in a Budibase data source.
5. Create a table based on the MySQL table containing a date.
6. Date field displays with a 20:00 timestamp.
7.  NOTE - Date-filtering and date choosing with Date Picker, etc. do NOT work with columns imported in this way.

**Expected behavior**
Importing a date field should NOT have a time appended to it at all.  
ADD:  See comment below.  DATE is being imported as DATE/TIME.  This would also likely fix my enhancement, #5548, which is, in fact, a bug.

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]: Ubuntu (pop-os)
 - Browser [e.g. chrome, safari]: Chrome

pbleic:
Here is the fundamental problem.  A DATE column in MySQL is imported as a DATE/TIME column in Budibase.  See screenshot from MySQL and same table in Budibase:
![Screenshot from 2022-04-23 09-18-03](https://user-images.githubusercontent.com/26202394/164896355-9b288cf3-4139-4107-a256-551f3162a6ba.png)
![Screenshot from 2022-04-23 09-17-34](https://user-images.githubusercontent.com/26202394/164896367-c83270bc-f1bd-498c-8689-02062360ee3e.png)

PR #5690, opened 2022-05-03
April release fixes - MySQL dates and various form/automation fixes

mike12345567:
## Description
Various fixes for release as follows:
1. https://github.com/Budibase/budibase/issues/5594 - minor fix, allow using `__` in SQL table names.
2. https://github.com/Budibase/budibase/issues/5612 - edge case, some formats of string being passed to MySQL were causing mis-processing, making sure this can't happen for data source plus tables and reducing the risk of date mis-processing for queries.
3. https://github.com/Budibase/budibase/issues/5553 - some work towards handling this, importing SQL tables will look for `date` and `time` types, appending a property to our `datetime` type (as we don't have a specific type for each) of `dateOnly` or `timeOnly` respectively. This is handled in the frontend, making sure the date picker only displays what it needs to. This should likely be added to the internal tables as well, to allow them to be date or time only at a data level, but for now just handling it for SQL tables. This isn't a total fix as some of the query engines still return time with the data, this may need to be stripped rather than just removed from the display, but it appears to fix issues with filtering etc.
4. Fixing an issue with exporting apps, streaming of the app export meant that our middlewares have finished execution but the DB connection needs to remain open, adding an edge case which opens the database outside of context and closes it when the stream finishes.
5. Fix for https://github.com/Budibase/budibase/issues/5669 - form step was erroring when loaded in an empty state.
6. Automation and query fix, there was an issue with context being closed prematurely as the threading uses callbacks, meaning that the `doInContext` functions would return before the automation/query has completed. Adjusted to use async functions and simply perform the callback at the end of the async function, meaning that the context function can wait for completion before closing out context.

### DIFF
--- a/packages/bbui/src/Table/DateTimeRenderer.svelte
+++ b/packages/bbui/src/Table/DateTimeRenderer.svelte
@@ -2,17 +2,22 @@
   import dayjs from "dayjs"
 
   export let value
+  export let schema
 
   // adding the 0- will turn a string like 00:00:00 into a valid ISO
   // date, but will make actual ISO dates invalid
   $: time = new Date(`0-${value}`)
-  $: isTime = !isNaN(time)
+  $: isTimeOnly = !isNaN(time) || schema?.timeOnly
+  $: isDateOnly = schema?.dateOnly
+  $: format = isTimeOnly
+    ? "HH:mm:ss"
+    : isDateOnly
+    ? "MMMM D YYYY"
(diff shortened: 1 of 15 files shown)


Rory-Powell (approved):
Nice fixes 👌

shogunpurple (approved):
Awesome work here - after this is tested we should hopefully be in good shape for release.

PR #5696, opened 2022-05-04
Foreach block fixes

PClmnt:
Some fixed issues: 

- String values weren't always being parsed correctly
- Delete button on front end should have deleted only the loop part of the block, not the attached block as well

### DIFF
--- a/packages/server/src/threads/automation.js
+++ b/packages/server/src/threads/automation.js
@@ -107,6 +107,7 @@ class Orchestrator {
     let loopSteps = []
     for (let step of automation.definition.steps) {
       stepCount++
+      let input
       if (step.stepId === LOOP_STEP_ID) {
         loopStep = step
         loopStepNumber = stepCount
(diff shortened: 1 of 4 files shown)


packages/server/src/threads/automation.js
```diff
@@ -107,25 +107,28 @@ class Orchestrator {
...
       }
 
       if (loopStep) {
-        input = await processObject(loopStep.inputs, this._context)
+        // lets first of all handle the input
+        if (
+          typeof loopStep.inputs.binding === "string" &&
+          loopStep.inputs.option === "String"
+        ) {
+          loopStep.inputs.binding = loopStep.inputs.binding.split("\n")
```
mike12345567, line 122: Talked about this offline - think this should handle CSV data as well as newline separated, might also be worth adding a tooltip to the input for bindings/values to tell the user the format it will split on.

mike12345567 (approved):
LGTM - just the one comment about CSV data and tooltip, once thats sorted should be good to go!

mike12345567 (approved):
LGTM!

PR #5699, opened 2022-05-04
Minor fix to string-templates for JSONparse helper

mike12345567:
## Description
Fixing an issue with JSONparse helper always being considered invalid due to a lack of input.


Rory-Powell (approved):
Looks good 👍

PR #5702, opened 2022-05-04
Account for object in failure condition

PClmnt:
- Account for objects when checking for the failure condition in a loop block. Before we were only checking equality on a flat comparison, now we check objects as well.


Rory-Powell (approved):
Looks good 👍

PR #5726, opened 2022-05-05
Assign user roles test fix

Mitch-Budibase:
Changing the conditionals for creating user and assigning roles. Should assist with the test issues


Rory-Powell (approved):
LGTM 👍

PR #5730, opened 2022-05-05
Job Application Tracker Template Tests

Mitch-Budibase:
Tests utilise pre-populated data already associated with the template

Tests associated with the Job Application Tracker template:
Add active/inactive vacancies
- Adds active and inactive vacancies to the relative locations
- Verifies tables are updated accordingly
- Random options picked for dropdown values

Filter applications by stage
- Each stage filter option is selected and the relative results verified

Edit an application
- Change application from not hired to hired
- Confirm relative sections update

Delete an application
- Currently skipped as there seems to be an issue with deleting an application within the template


shogunpurple (approved):
LGTM!

PR #5735, opened 2022-05-05
Content Security Policy fix for S3 upload

upgreydd:
CSP allows wildcard only at first element of FQDN, so policy `*.s3.*.amazonaws.com` is invalid cause it contains second wildcard in the middle of FQDN.

nginx.prod.conf.hbs is not parsed by real handlebarsJS but manual way with regex and replacements. It's impossible to use `{{#each}}` so I've hardcoded all regions available for S3 basing on documentation under: https://docs.aws.amazon.com/general/latest/gr/s3.html

## Description
https://github.com/Budibase/budibase/discussions/5608


shogunpurple (approved):
Thanks @Upgreydd for the contribution!

Issue #5778, opened 2022-05-09
App creation fails when tenant names contain the word `app` 


Rory-Powell:
**Hosting**
- Any multi tenancy environment

**Describe the bug**
The current logic for adding tenancy to an app id should be updated to support tenant names containing the word `app`. The root cause being how we grep for `app_` and `app_dev` in app db names. 

**To Reproduce**
- Create a tenant `testapp`
- Try to create an app

Related to
- https://github.com/Budibase/budibase/issues/5738
- https://github.com/Budibase/budibase/discussions/5774

PR #5783, opened 2022-05-09
Fixing app creation when tenant ID ends in app

mike12345567:
## Description
Fixing issue #5778 - making sure to only remove the first app_ element of the app ID.
 
Doesn't appear to be to do with contains, only affected tenants that ended in `app_`.

App created successfully in local cloud/multitenant environment (with tenant ID `testapp`).
image

This does not migrate existing apps which are in a bad state, but it will allow anyone with a tenant like this to create apps going forward - the migration can be a separate issue.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.146",
+  "version": "1.0.147",
   "npmClient": "yarn",
   "packages": [
     "packages/*"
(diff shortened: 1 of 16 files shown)


shogunpurple (approved):
LGTM - might be nice to have a little unit test just to cover it.

Rory-Powell (approved):
Looks good, thanks for fixing the pro version also. I was looking at having the bootstrap not fail when `1.0.46` exists on the filesystem and is yarn linked which should now be the case in CI. Using this fix should also work as there are no updated in 1.0.46 needed. Worth mentioning that the release script will overwrite your change to 
`"@budibase/pro": "1.0.47"` on the next release

Issue #5547, opened 2022-04-22
App Overview: Backups Section


shogunpurple:
When I'm viewing my app overview panel, I want to be able to roll back and manage my versions of my app from the dashboard so that I can roll back a bad version or revert back to a previous version of my app.

Spec: https://docs.google.com/document/d/1lhWFsRGVnuBdsN8Yt7ePe66dE1WzRD6EyIQM8jini6A/edit#

**COS**
- Show "Backups" panel under app overview
- Every revision should have the names of the screens, automations and datasources it currently contains attached to it for identifying revisions
- Backups are auto saved every hour (can be turned off)
- App publishes are contained in the backup history
- Users can restore their app back to any version in the backup history
- Users can name their backups
- Users can rename their backups
- Add backup floppy disk icon into builder to save a backup, once clicked, prepopulate a name in a modal for the backup, and allow them to edit or save.
- Every publish automatically saves a named backup
- When you restore a backup, another backup entry is created at the top of the list so context isn't lost
- Add "No Rows" option to export

**Designs**

Design scope and XD links inside Figjam file 
https://www.figma.com/file/Wl2EO1w4fB6uRJbZkHpXCK/Version-History?node-id=0%3A1

**Notes**
Revision history totally ignores rows, and only applies to everything else.

Issue #4977, opened 2022-03-21
Automation Run Log


shogunpurple:
As a user, I want to be able to see a history of my automation runs so that I can debug problems that arise from their use.

Spec: https://docs.google.com/document/d/1Nst-lib7CpyZe0nLp1vJ5U6vdrt3tGVCpA28SChr9wQ/edit

**COS**
- Add "Automation History" Section under the app overview section in the portal
- Show whether the runs failed or succeeded, and for failed runs show the error message
- Allow users to go to the relevant automation from a link in the history entry, so that they can go straight to the automation and configure it. 
- Automations are ordered by date and are paginated
- Automation runs are searchable by name
- Automation runs can be filtered by automation, date range (Last 30 days, Last 90 days, All time) and status
- Self hosted users should be able to turn automation logging off for both production and test
- if an app has a broken automation, it should be displayed in the app list
- Allow users to "ignore" each automation error to make the error status go away from the app list
- Details sidebar shows the results of the automation run step by step

**Designs**
Design scope and XD links all in Figjam file 

https://www.figma.com/file/rquXp9z75JqJMc6NpBHsGG/Automation-run-log?node-id=0%3A1

PR #5786, opened 2022-05-09
Feature/app overview section

deanhannigan:
Addresses: https://github.com/Budibase/budibase/issues/5543

## Description
An application overview section has been added to the builder. To access it, load up the application list and either click on the application name or the "View" button.

The Overview Section:
Meta information for the application is displayed here.
- Overview Tab 
  - Publish status 
  - Last Edited By
  - App Version
  - (TBD) Automation Summary - https://github.com/Budibase/budibase/issues/4977
  - (TBD) Backups Summary - https://github.com/Budibase/budibase/issues/5547 
- (TBD) Automation History
  - https://github.com/Budibase/budibase/issues/4977
- (TBD) Backups
  - https://github.com/Budibase/budibase/issues/5547 
- Settings Tab
  - Edit - the application name and URL can be altered here
  - Application Version - a summary of the current budibase version used by the app with actions to perform an update, if available
  - Self Host - a link to the self host documentation for budibase.

App list updates:
- There has been a minor reflow of the apps list. The "Create new application" and "Templates" buttons have been moved down to the app table heading. 
- The app lock status will now only display if the application is locked. The status is now represented by a button. Upon clicking the button, the user will be presented with a summary of the lock status.
- The "Spaceman" image has been removed.

## Screenshots

**Updated app list**
Screenshot 2022-05-12 at 09 40 52

**App Overview Page, overview tab**
Screenshot 2022-05-12 at 09 41 02

**App Overview Page, settings tab**
Screenshot 2022-05-12 at 09 41 32

**Edit your app icon**
Click on the icon in the app overview header for icon customisation options.
Screenshot 2022-05-12 at 09 45 22

**Application unlocking**
Application unlocking will now reflect the name of the user who holds the lock. If the user hasn't specified their full name, the email will be displayed instead
Screenshot 2022-05-12 at 09 52 43


packages/builder/src/components/common/AppLockModal.svelte
```diff
@@ -0,0 +1,146 @@
...
+  $: lockedByYou = $auth.user.email === lockedBy?.email
+
+  $: lockIdentifer = `${
+    lockedBy && lockedBy.firstName ? lockedBy?.firstName : lockedBy?.email
+  }`
+
+  $: lockedByHeading =
+    lockedBy && lockedByYou ? "Locked by you" : `Locked by ${lockIdentifer}`
+
+  $: lockExpiry = getExpiryDuration(app)
```
aptkingston, line 31: If we cache this in a reactive statement then the value will get stale over time, if the user doesn't refresh the page. I think we should just call `getExpiryDuration(app)` directly in the render code to ensure the label generated is always correct.

packages/builder/src/components/common/DashCard.svelte
```diff
@@ -0,0 +1,63 @@
...
+  export let actionIcon
+  export let action
+  export let dataCy
+
+  $: actionDefined = typeof action === "function"
+</script>
+
+<div class="dash-card" data-cy={dataCy}>
+  <div
+    class={actionDefined ? "dash-card-header active" : "dash-card-header"}
```
aptkingston, line 14: NAB: this is equivalent to just
```js
<div
  class="dash-card-header"
  class:active={actionDefined}
  ...
>
```
Just avoids repetition of class names.

packages/builder/src/components/common/DashCard.svelte
```diff
@@ -0,0 +1,63 @@
...
+    on:click={() => {
+      if (actionDefined) {
+        action()
+      }
+    }}
```
aptkingston, line 19: NAB: This can just be `on:click={action}`. If `action` is nullish, then nothing will happen. Just a bit terser!

packages/builder/src/components/common/DashCard.svelte
```diff
@@ -0,0 +1,63 @@
...
+  >
+    <span class="dash-card-title">
+      <Detail size="M">{title}</Detail>
+    </span>
+    <span class="dash-card-action">
+      {#if actionDefined}
+        <Icon name={actionIcon ? actionIcon : "ChevronRight"} />
```
aptkingston, line 26: NAB: This can be reduced to just `actionIcon || "ChevronRight"`.

aptkingston (changes requested):
Really nice work @deanhannigan! Functionally it all looks good to me - my comments in the review are mostly just terser and simpler ways of writing the same expressions.

I think we need to make a few visual changes though, to get this in line with the designs. It's important we stick to those designs to ensure our UI delivers a consistent and friendly experience for users, and so that it looks really polished. 

I pulled down the branch and played around and I've noted down anywhere I found a deviation from the designs that I think is worth addressing. Some of these might come across as being pedantic, but I really do think every little change is significant for providing a polished feel.

### App list
- More space needed below main buttons, and less space above the first app. The spacing on either side should be roughly the same:
image
- When no apps are locked, the app list shows an empty column which makes things look a bit odd. This is just a consequence of hiding "open" status. I think we should check if any app locks exist, and if they don't then update the grid layout here to remove this empty column. This is what I mean:
image
- The lock expiry time in the lock modal doesn't get updated in real time. I've noted this one as an actual review comment in the code.

### App overview
- The style of the back button needs updated to match designs. All of the back buttons actually need updated to match this new design (which was outside the scope of this PR) so this could maybe be done as a separate task later, just before deploying.
- The view app button should be a quiet button
- We need to make the tabs full width. To do this we can either remove the padding around the page as whole, or just apply a negative margin to the tab container. 
- The "blocks" on the overview tab shoudn't have a border round them
- The header section needs a background colour added - this will look better once the tabs are full width, properly dividing the page.
- The "more" menu (3 dots) needs copied in from the app list screen to beside the other buttons on this page. We should remove the "edit" and "edit icon" options from this menu as these are both possible now as standalone features on the overview page.
- We need to adjust the spacing slightly in the overview blocks so that the content looks a bit more centered when the blocks are at their minimum height.
- Change spacing in "app version" and "self-host budibase" sections in the settings tab to match the "Name and Url" section - less spacing after divider, less spacing before button
- Add more spacing between the sections in the settings tab
- "Name and Url" -> "Name and URL" (case change)
- "App Version" -> "App version" (case change)
- "Update App" -> "Update app" (case change)
- "Self-host Budibase" button should be a secondary button rather than a CTA

packages/bbui/src/Tabs/Tab.svelte
```diff
@@ -27,14 +27,30 @@
...
+  $: {
+    if ($selected.title === title && tab_internal) {
+      if ($selected.info?.left !== tab_internal.getBoundingClientRect().left) {
+        $selected = {
+          ...$selected,
+          info: tab_internal.getBoundingClientRect(),
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
deanhannigan: Yep, exactly. When you click on a tab, the underline offset is automatically updated in the Tabs component (calculateIndicatorOffset). If you wish to navigate to another tab without clicking on one directly, you have to trigger the recalculation of the tab underline manually. 

If you click on the "App Version" card header, it will update the selected tab key only. At that point, the selected tab will determine that it is now the current tab and ensure that its own position be pushed into the 'tab' context. Updating the context in this way triggers the repositioning in the Tabs parent component with the correct data.  Hopefully than makes sense!

packages/builder/src/components/common/AppLockModal.svelte
deanhannigan: Yep, that's fair enough. I've replaced the value in the modal.

packages/builder/src/components/common/DashCard.svelte
deanhannigan: Updated this to follow the convention.

packages/builder/src/components/common/DashCard.svelte
deanhannigan: This has been updated

PR #5787, opened 2022-05-09
Allow to use select for boolean field type

upgreydd:
## Description
Like in title

just parse strings to a boolean when a source of data is custom and you declare 2 options with "true" and "false" values.


shogunpurple (approved):
LGTM

PR #5793, opened 2022-05-10
Adds tooltips to icons

upgreydd:
## Description
Allow set specific tooltip for the icon - useful for naming actions with icons
Based on https://github.com/Abreu00/svelte-tooltip

## Screenshots
image
image

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -1011,6 +1011,33 @@
         "key": "color"
       },
       {
+        "type": "text",
+        "label": "Tooltip text",
+        "key": "tooltip"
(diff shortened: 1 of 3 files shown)


shogunpurple (commented):
Thanks for putting this together @Upgreydd

We aren't going to merge this one, unfortunately.

This PR doesn't really make a lot of sense:
- we already have a tooltip component that is part of spectrum CSS. 
- There's no issue for this, or it hasn't been requested
- It is not something that is going to be that useful for the rest of the community, in its current format

Thanks again for taking the time though!

PR #5800, opened 2022-05-10
Misc Picker Fixes

### LINKED ISSUE #5317: Form components bugs & improvements
I went through the form components and noted a lot of bugs, quick wins and UX improvements, as well as overhaul and suggestions for some.

The spec is available here: https://docs.google.com/document/d/1GPhTAfBwIsF0OO-5mcOyN7HqBivB4vPB_60rJEY3opo/edit?usp=sharing

melohagan:
## Description
Misc fixes for pickers.

1. Pickers are now closed when you click on the list box.
2. Multi-select Pickers can now be checkboxes (similar to RadioGroup)

----
I did try to address the focus on search box for autocomplete, but I couldn't figure it out, although it should be possible.
Another suggestion was changing Long Form Field to Text Area. In my opinion that makes sense, but I left it.
The other issues raised in that document (see linked issue) are mostly covered in other issues out there.

Addresses: 
- https://github.com/Budibase/budibase/issues/5317

## Screenshots
Screenshot 2022-05-10 at 16 01 51

Screenshot 2022-05-10 at 16 02 06

Screenshot 2022-05-10 at 16 02 52


packages/bbui/src/Form/Core/Picker.svelte
```diff
@@ -71,105 +71,73 @@
...
-    <span class="icon-Placeholder-Padding">
-      <img src={fieldIcon} alt="icon" width="20" height="15" />
-    </span>
-  {/if}
-
-  <span
-    class="spectrum-Picker-label"
-    class:is-placeholder={isPlaceholder}
-    class:auto-width={autoWidth}
+<div use:clickOutside={() => (open = false)}>
```
aptkingston, line 74: I was worried that adding another wrapper div around the core Picker could have consequences for controlling the width, but I checked out the branch and everything still seems to work. If we see any instances of pickers not using the full width any more then we might need to add some styles here to ensure the inner button takes up the full width of this new wrapper div, but it seems fine for now.

packages/client/manifest.json
```diff
@@ -2346,6 +2346,43 @@
...
+          },
+          {
+            "label": "Vertical",
+            "value": "vertical"
+          }
+        ],
+        "dependsOn": {
+          "setting": "optionsType",
+          "value": "checkbox"
+        }
```
aptkingston, line 2384: NAB: Perfect usage of "dependsOn" :+1: One other setting which we need to adjust now is the "autocomplete" setting, which now is only relevant for the "select" variant.

packages/client/src/components/app/forms/MultiFieldSelect.svelte
```diff
@@ -61,17 +63,31 @@
...
-      getOptionLabel={flatOptions ? x => x : x => x.label}
-      getOptionValue={flatOptions ? x => x : x => x.value}
-      id={fieldState.fieldId}
-      disabled={fieldState.disabled}
-      on:change={handleChange}
-      {placeholder}
-      {options}
-      {autocomplete}
-    />
+    {#if !optionsType || optionsType === "select"}
```
aptkingston, line 66: Well done for remembering to make select the fallback to account for all existing apps.

aptkingston (approved):
Awesome work @melohagan. One tiny change noted below.

packages/client/manifest.json
melohagan: Good catch! I've added that to the manifest.

PR #5801, opened 2022-05-10
Fix CLS hooked memory leak

mike12345567:
## Description
After fixing the memory leak we found in our PouchDB usage we were still experiencing a slow leak, I found issues:
1. https://github.com/Jeff-Lewis/cls-hooked/issues/67
2. https://github.com/Jeff-Lewis/cls-hooked/issues/66

These two issues demonstrate a memory leak that is related to the async hooks not being cleared up when the context is finished. We also had an issue with using a single static and global namespace which could not be destroyed. First I changed how our context functions work, so that there are two contexts, a single global one which stores the ID of a namespace which is created per request, then the request namespace which is created and destroyed as part of the run. I tested this by taking periodic snapshots of the heap state and comparing them while the app service was under heavy load, before these changes I would see a few structures grow in count in a linear fashion compared to the number of requests, e.g. PouchDB objects left over after being closed.

Once these changes were implemented all of these structures were no longer present, the memory usage would remain constant across all of the snapshots.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "build": "rimraf dist/ && tsc -p tsconfig.build.json && mv dist/src/* dist/ && rimraf dist/src/ && yarn postbuild",
+    "debug": "yarn build && node --expose-gc --inspect=9222 dist/index.js",
     "postbuild": "copyfiles -u 1 src/**/*.svelte dist/ && copyfiles -u 1 src/**/*.hbs dist/ && copyfiles -u 1 src/**/*.json dist/",
     "test": "jest --coverage --maxWorkers=2",
     "test:watch": "jest --watch",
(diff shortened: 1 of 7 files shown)


packages/server/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "build": "rimraf dist/ && tsc -p tsconfig.build.json && mv dist/src/* dist/ && rimraf dist/src/ && yarn postbuild",
+    "debug": "yarn build && node --expose-gc --inspect=9222 dist/index.js",
```
Rory-Powell, line 13: What relevance does the `expose-gc` flag have here? the only details I can find on it are that it will allow us to call `global.gc()` from within the program, is that something you did while attached in a debugger?

Rory-Powell (approved):
Great work diagnosing and fixing this issue 💯 / 💯

packages/server/package.json
mike12345567: yes, while I was debugging I was manually calling the `global.gc()` to make sure that the garbage collector had ran before I collected a snapshot, it just made sure that my testing of the solution was accurate.

PR #5806, opened 2022-05-10
Add query quota increment to rows CRUD

Rory-Powell:
## Description
Targeting this at master so we can do an earlier release to get some statistics

Addresses: 
- https://github.com/Budibase/account-portal/issues/117

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -71,7 +71,7 @@
     "@apidevtools/swagger-parser": "^10.0.3",
     "@budibase/backend-core": "^1.0.151",
     "@budibase/client": "^1.0.151",
-    "@budibase/pro": "1.0.151",
+    "@budibase/pro": "1.0.150",
     "@budibase/string-templates": "^1.0.151",
     "@bull-board/api": "^3.7.0",
     "@bull-board/koa": "^3.7.0",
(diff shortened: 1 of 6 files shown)


Rory-Powell:
Hi @shogunpurple @mike12345567, do you think there's anything else worth adding to here? I wasn't sure if it's something we want to add to table crud as well

mike12345567:
Are datasource queries included in this in anyway? The datasource queries have their own endpoint through the `api/routes/query.js` - the `executeV1` and `executeV2` controllers (V1 is deprecated, we can likely remove it soon, was to support anything pre the REST query update).

shogunpurple:
@Rory-Powell everything should be included - queries for all datasources (including internal tables) and even in dev

Rory-Powell:
@mike12345567 yep datasource queries are already being counted. It's done here:
https://github.com/Budibase/budibase/blob/crud-quotas/packages/server/src/api/controllers/query/index.ts#L168
https://github.com/Budibase/budibase/blob/crud-quotas/packages/server/src/api/controllers/query/index.ts#L126

@shogunpurple This will cover the dev experience as it'll use the rows api. 

I was wondering are rows the logical separation here, or would we want to increment query count when e.g. getting back the table definition itself

shogunpurple:
Awesome @Rory-Powell - yeah I wouldn't worry about getting the table definitions, it's just when you make a query for rows essentially. Don't want to be too stingy lol

mike12345567 (approved):
LGTM!

Issue #5809, opened 2022-05-11
Trying to duplicate an auto generated screen throws error


aderici:
{
    "message": "Invalid body - \"props._instanceName\" is required",
    "status": 400
}

i'm using self hosted version. I'm on the latest version. Am i missing something?

aptkingston:
I already have this fixed on a dev branch but I'll PR a bug fix to master. We just need to remove instance name from the screen validation as we don't use it any more for screen.

PR #5825, opened 2022-05-11
Fix duplicate screens

aptkingston:
## Description
Removes `_instanceName` as a required setting for screens, as we don't show this field for them any more.

Addresses #5809.


mike12345567 (approved):
LGTM

Issue #3821, opened 2021-12-17
create a clickable container


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/2256

<div type='discussions-op-text'>

<sup>Originally posted by **oswaldB** August  4, 2021</sup>
Hello,

It could be very convenient and very useful to have the capacity to create a clickable container :)

Is it something hard for you to design?

Have a nice day</div>

PR #5829, opened 2022-05-11
Containers support onClick

melohagan:
## Description
You can now add onClick events to Containers. 

Addresses: 
- https://github.com/Budibase/budibase/issues/3821
- https://github.com/Budibase/budibase/discussions/5821#discussioncomment-2729810

## Screenshots
Screenshot 2022-05-11 at 14 05 18


aptkingston (commented):
You've implemented this completely correctly - but whether or not we want to add this feature is another question haha. Personally I'm actually against it because I just don't like the ambiguity of any container potentially being a click handler, especially since the only indicator is a pointer cursor.

I think if we wanted to add and ship this then we would need to solve the issue of nesting other clickable components inside containers. I wouldn't want the container click event to fire if I click a button, link, form field or indeed any other interactive component inside the container.

Happy to open this up to the entire team for feedback though about how/if we should add this.

As for the code though - 100% looks good to me :+1:

melohagan:
It has been requested a few times, so I think the use case is there: https://github.com/Budibase/budibase/discussions/2256

It would be good to be able to disable nested click events. I was trying to think of a reason why you might want this to be a toggle, but nothing comes to mind.

shogunpurple:
LGTM!

Im not a huge fan of the idea of clickable containers either - but the use case that it generally supports is being able to build completely custom buttons/clickable images etc, which is valid and requested often. 

On the bubbling - I would have to agree. Even though it’s not the default behaviour in the browser, I feel like its easier to understand for newcomers if we prevent it by default. Especially when it comes to onClick/onChange events. 

This would probably mean however, that we would have to add config (a Boolean or whatever) to the handler setting in the builder to allow it if it was desired.

shogunpurple (approved):

packages/client/src/components/app/Container.svelte
```diff
@@ -10,22 +10,30 @@
...
+    clickableClass,
   ].join(" ")
 </script>
 
-<div class={classNames} use:styleable={$component.styles} class:wrap>
+<div
+  class={classNames}
+  use:styleable={$component.styles}
+  class:wrap
+  on:click={onClick}
```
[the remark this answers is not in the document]
melohagan: Handy!

packages/client/src/components/app/Container.svelte
```diff
@@ -25,7 +26,13 @@
   ].join(" ")
 </script>
 
-<div class={classNames} use:styleable={$component.styles} class:wrap>
+<div
+  class={classNames}
+  class:clickable={onClick}
+  use:styleable={$component.styles}
+  class:wrap
+  on:click={onClick}
```
aptkingston, line 34: If we use `on:click|self={onClick}` here then I think that will sort the bubbling events :+1:

aptkingston: Mel pointed out that this prevents clicks on *any* child component from firing - even things like a paragraph for which you would still expect the container on click event to fire for. Let's leave this open for now and I'll jump on at some stage and disable propagating of events from our other components, which is the proper solution to this.

I'd rather get that in before we ship this.

melohagan:
@aptkingston I had another go at this. Buttons are disabled if the parent container has an onClick event. 

disablebuttonclick

aptkingston:
Started implementing handlers to stop propagation of all clicks on all components and realised it's far too much work to be feasible. 

However I had a flash of inspiration and realised we can easily solve this with some CSS. What I've done is just use CSS to disable pointer events for every single component inside a clickable container. This means we don't need any additional code, and clicking anywhere inside the container will only trigger the container click event.

While this prevents being able to use something like a button inside a clickable container, I think that's actually a good thing. It enforces good design, and prevents nesting multiple ambiguous click handlers.

aptkingston (approved):

Issue #5188, opened 2022-03-31
Pre-fill app setup modal


joebudi:
**Describe the feature request**
I feel we should pre-fill the app creation modal with a sensible default (that increments)

For example:

Name:
`My first app`

Url:
`my-first-app`

Twilio and others do something like this, and it makes onboarding faster and removes potential obstacles. 

**Screenshots**
<img width="1676" alt="CleanShot 2022-03-31 at 11 04 42@2x" src="https://user-images.githubusercontent.com/49767913/161030674-638cda1c-8147-4f10-aedf-b84c35b12a1e.png">

<img width="1679" alt="CleanShot 2022-03-31 at 11 05 05@2x" src="https://user-images.githubusercontent.com/49767913/161030755-32ca821e-2760-45c6-bd85-004346586066.png">

<img width="1920" alt="CleanShot 2022-03-31 at 10 57 40@2x" src="https://user-images.githubusercontent.com/49767913/161030475-6deb172b-a1ea-4020-8154-9a67e273f47a.png">

**Notes/further work**
I would add to this, with the new designs, we will display the app name within the app-building UI. The user should be able to change the app/domain of the app here in my opinion (@AndyWhann ) what do you think?

AndyWhann:
Yeah we could do this but the field needs to have the text highlighted and in focus by default so the user can delete this suggestion easily too.

Would also maybe suggest using the name of the user e.g 

Andy's app (1)

aptkingston:
This is already done for templates at least, Dean added it. Name is focused automatically too.

e.g. when clicking on a template:
![image](https://user-images.githubusercontent.com/9075550/161039666-48072dbf-90f4-41d4-bfe3-53b6455c690f.png)

PR #5855, opened 2022-05-12
Feature/fill app name on create

deanhannigan:
## Description
When creating an application, the name will now auto-fill with either "My app" or with users first name, if it has been configured on their account.

This behaviour is the same for apps created from scratch and those created via import.

The behaviour for apps created with templates is unchanged.

Addresses: 
https://github.com/Budibase/budibase/issues/5188

## Screenshots
**Default to "My app" when no first name is set in the system. Focus is forced to the app name by default.**
Screenshot 2022-05-12 at 17 14 58

**If the user has specified their first name in the user information for their account, it will use that instead.**
Screenshot 2022-05-12 at 17 14 46


packages/bbui/src/Form/Core/TextField.svelte
```diff
@@ -13,10 +13,18 @@
...
 
   const dispatch = createEventDispatcher()
+
+  let field
   let focus = false
 
+  $: if (autofocus === true && field) {
+    focus = true
+    field.focus()
+  }
```
aptkingston, line 26: If we move this to a svelte `onMount` callback then we won't need to worry about whether the reference exists or not (it always will after mounting). I think semantically as well it makes sense to just explicitly focus once when mounting is complete.

packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -22,7 +22,10 @@
   $: validation.check($values)
 
   onMount(async () => {
-    $values.name = resolveAppName(template, $values.name)
+    const defaultName = $auth.user?.firstName
+      ? `${$auth.user.firstName}s app`
```
aptkingston, line 26: I wish we could put an apostrophe here since we're showing possession.... but app names can't have special characters so we can't. This will also look weird if their name ends in an "s" - e.g. "Chriss app". I think we should account for names ending in "s" and add some extra logic here to not append the second "s".

packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -159,6 +162,7 @@
     />
   {/if}
   <Input
+    autofocus={true}
```
aptkingston, line 165: NAB: just an FYI - but this is equivalent to just `autofocus`. You can skip the `={true}` part when just passing in a boolean prop like this.

aptkingston (approved):
Nice one @deanhannigan! LGTM but just suggested a couple of tiny changes.

packages/bbui/src/Form/Core/TextField.svelte
deanhannigan: Initially this was intended to allow the value to continue to be changed after loading, but in this scenario that isn't required. The behaviour has been updated.

packages/builder/src/components/start/CreateAppModal.svelte
deanhannigan: Agreed. I've added accommodation now for user names that end in 'S'.  The tests were updated to reflect this use case.

packages/builder/src/components/start/CreateAppModal.svelte
deanhannigan: In this instance, there is no variable in the parent component called 'autofocus'. If I don't explicitly set the value to 'true', it'll end up undefined in the Input component

aptkingston: Just passing `autofocus` is the same as `autofocus={true}` - I think you're thinking of `{autofocus}` (wrapped in curlies) which passes in the parent value like you say. But if you remove the curlies then the behaviour is different :+1:

aptkingston: You don't need to change this or anything lol but it's just a tip for the future!

Issue #5841, opened 2022-05-12
User with basic role permission can not call Trigger Automation


MSrubar:
User with basic role permission can not call Button - Define actions - Trigger Automation after update to build 1.0.154.

 Basic User get error message "User does not have permission" after pushing button.
![canvas1](https://user-images.githubusercontent.com/41183598/168020777-11f39f01-9c21-471f-971a-adb99e5227cb.png)

![canvas](https://user-images.githubusercontent.com/41183598/168019794-b7c16b51-9847-4085-8c96-92a0bb114a95.png)

Rory-Powell:
Deployed in `1.0.158`

MSrubar:
Thank you

PR #5858, opened 2022-05-12
Add execute automation to basic user / write permission set

Rory-Powell:
## Description
Add execute automation to basic user / write permission set

Addresses: 
- https://github.com/Budibase/budibase/issues/5841


shogunpurple (approved):
LGTM!

Issue #5367, opened 2022-04-12
Cannot set role from Admin to No Access 


melohagan:
**Describe the bug**
Related to this discussion: https://github.com/Budibase/budibase/discussions/5361
I am trying to configure roles for a dev user, to remove access to some apps. Whenever I select **No Access** a toast pops up with success, but the role does not change from *Admin*. I can change the role to *Power* or *Basic*.

**To Reproduce**
1. Add a new development access user
2. Go to Configure roles for this new user
3. Select an app, and select Role: No Access
4. Green "Role Updated" toast appears, but the role in the table remains *Admin*

**Expected behavior**
As an admin user, I should be able to restrict the access of dev users for any number of apps.

sohail-hussain:
just ran into same bug - is the only workaround

1. export application
2. delete application
3. re-import the application
?

aptkingston:
I think this is because if the user has the development priviledge then they implicitly have access to all apps and you can't revoke it. Developers can edit all apps.

You can unset admin as long as they aren't a developer so I don't think this is a bug, but just by design.

melohagan:
I see, that makes sense. Maybe remove the 'No Access' option from the dropdown (or disable it) for users that have developer set.

Issue #5156, opened 2022-03-29
Webhook Schema and Trigger URLs do not include domain


melohagan:
**Describe the bug**
When creating a Webhook automation, the domain (e.g. budibase.app/) is not being added to the URL

**To Reproduce**
1. Create a Webhook automation

**Expected behavior**
The URLs shown should include the domain

**Screenshots**
![image](https://user-images.githubusercontent.com/9000976/159528450-4d647950-24ae-4a39-91b0-8a1b88745410.png)

![Screenshot 2022-03-22 at 16 45 52](https://user-images.githubusercontent.com/101575380/159531824-1ecb2bf3-1481-4a9d-801c-1ad0c6563539.png)

PR #5861, opened 2022-05-12
webhook URL paste, hiding no access role

shogunpurple:
## Description
Just a few small ones.

Addresses:
Re-added the host when using webhooks:
- https://github.com/Budibase/budibase/issues/5156

Disable "No Access" role for admins and devs, as it doesn't do anything:
- https://github.com/Budibase/budibase/issues/5367

## Screenshots
Screenshot 2022-05-12 at 19 52 14

Screenshot 2022-05-12 at 19 50 55

Screenshot 2022-05-12 at 19 51 09

### DIFF
--- a/packages/builder/src/pages/builder/portal/manage/users/_components/UpdateRolesModal.svelte
+++ b/packages/builder/src/pages/builder/portal/manage/users/_components/UpdateRolesModal.svelte
@@ -6,8 +6,6 @@
   export let app
   export let user
 
-  let devAccess = user?.admin?.global || user?.builder?.global
-
   const NO_ACCESS = "NO_ACCESS"
 
   const dispatch = createEventDispatcher()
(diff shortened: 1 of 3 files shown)


packages/builder/src/pages/builder/portal/manage/users/_components/UpdateRolesModal.svelte
```diff
@@ -6,6 +6,8 @@
   export let app
   export let user
 
+  let devAccess = user?.admin?.global || user?.builder?.global
```
Rory-Powell, line 9: I think this should apply to builder only, an admin user doesn't automatically get access in the portal and when trying to view the app directly

Screenshot 2022-05-13 at 11 34 03
Screenshot 2022-05-13 at 11 34 10

shogunpurple: Good catch - we can revoke dev access on the admin so that makes sense

Issue #5731, opened 2022-05-05
The date stored in MySQL is always wrong


apsonandtesla:
**Hosting**

- Self
  - Method: docker compose (via portainer stack)
  - Budibase Version: 1.0.142

**Describe the bug**
For fields with DATE datatype in MySQL, the Date Picker in Budibase always stores the wrong date.

**To Reproduce**
1. Create a data table in MySQL/MariaDB with a field named "Date" and set the datatype to "DATE"
2. Connect the data table created in _Step 1_ to Budibase
3. Click "Create row" to add a new row to the data table
4. Select a date (for example, "May 6, 2022")
5. Click "Create Row" to save the entry
6. You can now see that the date stored in MySQL (and as shown in Budibase) is wrong ("May 5, 2022")

**Expected behavior**
The date stored in MySQL should be as the same as what we select in Date Picker.

**Screenshots**
![create_row_budibase_date](https://user-images.githubusercontent.com/42469315/166990373-c10ffe36-4636-4712-befd-99a933da2b1b.png)
![check_row_budibase_date](https://user-images.githubusercontent.com/42469315/166990467-21bdb320-67a7-4dba-8ae5-30bee85a17b5.png)
![check_row_mariadb_date](https://user-images.githubusercontent.com/42469315/166991022-93588c44-e562-40b2-bab9-f07c3a4271a1.png)

apsonandtesla:
I think it could because that the date picker isn't really just picking a date even when the datatype is "DATE" in MySQL/MariaDB (and also we can see that in Budibase the "DATE" type is shown as "Date/Time"), but instead the picker is just hiding the time selector but still it's getting a value. Budibase will then use that value (with time) and convert it to a certain timezone (possibly UTC) and therefore creating the wrong date which is stored into the database.

apsonandtesla:
Thanks @mike12345567

PR #5873, opened 2022-05-13
Fix for external DB date only columns in various timezones

mike12345567:
## Description
Addresses an issue with Date only columns from external DBs, offset the UTC offset in ISO strings to allow dateonly columns in external DBs to save the correct date when in a timezone differing from UTC.

Addresses: 
- #5731


aptkingston (approved):
LGTM!

PR #5875, opened 2022-05-13
Misc MongoDB Connector Fixes

melohagan:
## Description
Added support for ObjectIds. Made some fixes for the CRUD operations.

Addresses: 
- ObjectId bug: https://github.com/Budibase/budibase/issues/5613
- General Query issues: https://github.com/Budibase/budibase/issues/5387

There's still a number of improvements to be done in the V2 connector, but this should help a bit in the meantime.

## Screenshots
**Insert**
mongodb_insert

**Update**
mongodb_update

Issue #5387, opened 2022-04-13
Improve the MongoDB First Query Experience


MarcusSorealheis:
**Describe the feature request**

Connecting to MongoD is very simple if you run MongoDB on localhost.  Querying it is tough. 

Furthermore, to preserve resources, I almost exclusively use Atlas to host my MongoDB. In many cases, I have too much data to store locally. 

**Screenshots**

Today, the experience is not bad as seen below for localhost:

<img width="628" alt="image" src="https://user-images.githubusercontent.com/2353608/162145575-ab1cf9ec-9f49-4058-b26c-89e9d75e936a.png">

If you are using Atlas, it is not clear what users should include.

Once to the query phase, things get a bit more confusing. Here, I am wondering, are these the only db commands I can use?

<img width="761" alt="image" src="https://user-images.githubusercontent.com/2353608/162147287-3730cb39-96eb-4270-9f25-a7a4aea55bb7.png">

I assume these options are not aggregation pipeline commands. Perhaps we should clarify it, or consider an `aggregate` command.

shogunpurple:
Hi @MarcusSorealheis - thanks very much for typing up this detailed issue. We are planning on doing a full revamp to the MongoDB connector in June - https://github.com/Budibase/budibase/issues/4362

It still needs a bit more specification, but some of the features it definitely will have is:
- Support for every query method in the mongo node driver, rather than a select few
- Support for aggregations
- Much improved UX and the ability to allow you to test a connection
- Improved mongoDB query language support

Are there other features/UX updates that you would like to see included in a mongoDB connector? I noticed that you actually work at Mongo, so you are in a pretty good position to provide feedback or feature requests!

Thanks!

MarcusSorealheis:
Here are the three things I think would be most beneficial for the most compelling swathe of Budibase users:

1. An easy path/button to connect to an existing Atlas Cluster. This will facilitate easy migration of existing apps or for testing purposes. Also, MongoDB has a lot more feature when in Atlas so it is often preferred. 
2. Support for Client-Side Field Level Encryption. Companies building internal apps with sensitive data, which is probably many, would really benefit from Budibase supporting encrypted fields. Here are the Node [docs](https://www.mongodb.com/docs/drivers/node/current/fundamentals/csfle/). CSFLE requires Atlas or Enterprise. 
[...]

PR #5877, opened 2022-05-13
Misc MongoDB Connector Fixes 

### LINKED ISSUE #5613: Can't figure out how to do a delete request using the mongoDB connector
I'm using your hosted version, I've succesfully connected to mongoDB and can display everything that I need. The problem is that I can't delete something

These are the querries I've tried
{ "_id" : "{{id}}" } this one seems to work but return a delte count of 0, and the Id I'm using is 100% in the default input the same.
{ "_id" : ObjectId("{{id}}") } this one returns a mistake

**Screenshots**
image
image
image
image

mike12345567:
## Description
Rebase of #5875 for @melohagan 

Added support for ObjectIds. Made some fixes for the CRUD operations.

Addresses: 
- ObjectId bug: https://github.com/Budibase/budibase/issues/5613
- General Query issues: https://github.com/Budibase/budibase/issues/5387

There's still a number of improvements to be done in the V2 connector, but this should help a bit in the meantime.

## Screenshots
**Insert**
mongodb_insert

**Update**
mongodb_update

### DIFF
--- a/packages/server/src/integrations/mongodb.ts
+++ b/packages/server/src/integrations/mongodb.ts
@@ -85,12 +85,12 @@ module MongoDBModule {
 
     createObjectIds(json: any): object {
       const self = this
-      function replaceObjectIds(json: any) {
+      function interpolateObjectIds(json: any) {
         for (let field of Object.keys(json)) {
           if (json[field] instanceof Object) {
             json[field] = self.createObjectIds(json[field])
           }
-          if (field === "_id") {
+          if (field === "_id" && typeof json[field] === "string") {
             const id = json["_id"].match(
               /(?<=objectid\(['"]).*(?=['"]\))/gi
             )?.[0]
@@ -104,11 +104,11 @@ module MongoDBModule {
 
       if (Array.isArray(json)) {
         for (let i = 0; i < json.length; i++) {
-          json[i] = replaceObjectIds(json[i])
+          json[i] = interpolateObjectIds(json[i])
         }
         return json
       }
-      return replaceObjectIds(json)
+      return interpolateObjectIds(json)
     }
 
     async create(query: { json: object; extra: { [key: string]: string } }) {
--- a/packages/server/src/integrations/mongodb.ts
+++ b/packages/server/src/integrations/mongodb.ts
@@ -112,6 +112,33 @@ module MongoDBModule {
       return interpolateObjectIds(json)
     }
 
+    parseQueryParams(params: string, mode: string) {
+      let queryParams = params.split(/(?<=(},)).*{/g)
+      let group1 = queryParams[0]
+      let group2 = queryParams[2]
+      let group3 = queryParams[4]
+      if (group1) {
+        group1 = JSON.parse(group1.replace(/,+$/, ""))
+      }
+      if (group2) {
+        group2 = JSON.parse("{" + group2.replace(/,+$/, ""))
+      }
+      if (group3) {
+        group3 = JSON.parse("{" + group3.replace(/,+$/, ""))
+      }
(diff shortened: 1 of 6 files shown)


packages/server/src/integrations/mongodb.ts
```diff
@@ -6,13 +7,19 @@ import {
...
 
   interface MongoDBConfig {
     connectionString: string
     db: string
   }
 
+  interface UpdateDoc {
+    filter: object
+    update: object
+    options: Object
```
Rory-Powell, line 20: If the `Object` vs `object` usage above isn't intentional we could make the options an `object` as well. I have always used the upper case variant without much thought, after seeing this I had a look at this great article:
https://2ality.com/2020/01/typing-objects-typescript.html#object-vs.-object-in-typescript

The tldr is `Object` allows for primitives, which we would use defined types for anyway like `string` `number` so we can just use `object`  here and probably in most cases.

packages/server/src/integrations/mongodb.ts
```diff
@@ -141,18 +183,27 @@ module MongoDBModule {
       }
     }
 
-    async update(query: { json: object; extra: { [key: string]: string } }) {
+    async update(query: { json: UpdateDoc; extra: { [key: string]: string } }) {
```
Rory-Powell, line 186: Instead of defining our own type here for `UpdateDoc` we can use the official mongo types. To do this there would only be a few changes needed

1. Add the dependency
```
yarn add --dev @types/mongodb@3.6.3
```

2. Update the mongo import, instead of `require` use `import`. Move this to the top level of the file outside of `MongoDBModule`
```
import { MongoClient, ObjectId } from "mongodb"
```

I tried this locally and there are a few req squiggly lines around our usage of the mongo client, similar to using the `UpdateDoc` fix, maybe there could be more

packages/server/src/integrations/mongodb.ts
melohagan: Good catch! I'd also accidentally brought in an Object class from an aws library, so I removed that import as well.

packages/server/src/integrations/mongodb.ts
melohagan: I brought in the official types. 

Doing this made me realised I'm probably doing this wrong, but it might be okay for now? 

To do filter/updates with this implementation you need to do:
```
{
  "filter": { "name": "A" },
  "update": { "$set": { "name": "B" } }
}  
```
but if you were doing this as a mongodb query it should really be:
**Update One Selected**
```
{ "name": "A" },
{ "$set": { "name": "B" } }
```

Rory-Powell: I see what you mean, having the two different dialects isn't ideal here, this will add translation from the bb query domain to the mongo one that'll need extra docs, etc. Probably the right way here is to have fields rendered based on the query type selected, that sounds like a v2 thing though. 

The simplest thing to do for now might be to allow them to paste an identical mongo query and try to determine which part is the operator object in the update queries `findOneAndUpdate`, `updateOne` and `updateMany`. Could be a simple check for `$` in the key, and fallback to the second object which seems to be the case in all the example docs.

shogunpurple (approved):
Nice! This is actually a very significant enhancement and turns on a lot of Mongos features in this connector.

packages/server/src/integrations/mongodb.ts
melohagan: I've updated it to parse the params if needed. So both of the options above will now work.

Rory-Powell: sounds great, good job mel 👍

PR #5931, opened 2022-05-17
Improve parseQueryParams regex

melohagan:
## Description
Improved the Regex for parsing query params.


Rory-Powell (approved):
Can't say I'm deciphering this regex but I trust you 👍

PR #5953, opened 2022-05-18
Porting usage UI to builder

mike12345567:
## Description
This is a PR for issue https://github.com/Budibase/account-portal/issues/111 - introducing the same usage UI that is present in the account portal.

This also adds a `DISABLE_DEVELOPER_LICENSE` environment variable which can be applied to the worker/server so that usage quotas are returned properly, not just assuming unlimited for everything.

Also refactored the lists of env variables in both the server and worker to make them a bit easier to look at, as well as removing some of the ancient env variables from the server which haven't been used since the electron days.

Addresses: 
- https://github.com/Budibase/account-portal/issues/111

## Screenshots
UI if you have the `accountPortalAccess` flag set on your user account.
image

UI if you are not the account holder.
image


packages/builder/src/components/billing/Usage.svelte
```diff
@@ -0,0 +1,56 @@
+<script>
```
Rory-Powell, line 1: maybe we could remove `billing` from the file path in budibase

packages/builder/src/pages/builder/portal/settings/usage.svelte
```diff
@@ -0,0 +1,140 @@
...
+  import { admin, auth, licensing } from "stores/portal"
+  import Usage from "components/billing/Usage.svelte"
+
+  let staticUsage = []
+  let monthlyUsage = []
+  let loaded = false
+
+  $: quotaUsage = $licensing.quotaUsage
+  $: license = $auth.user?.license
+  $: console.log($auth.user)
```
Rory-Powell, line 20: _remove me_

packages/builder/src/pages/builder/portal/settings/usage.svelte
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+</script>
+
+{#if loaded}
+  <Layout>
+    <Heading>Usage</Heading>
+    <Body
+      >Get information about your current usage within Budibase.
+      {#if $admin.cloud}
```
Rory-Powell, line 87: Good shout on re-using this copy, it fits perfectly

Rory-Powell (commented):
Looks great, nice work. I wonder should we feature flag this page until we get sign off on the licensing strategy, quotas in particular are a bit up in the air atm. The corresponding quotas page in account portal is locked behind the `LICENSING` tenant feature flag

mike12345567:
Yeah its probably worth putting this behind feature flag if changes are still on the way for quotas, I'll throw that in!

Rory-Powell (approved):
Great work Michael 👍  Thanks for the help on this one

PR #5965, opened 2022-05-18
Cypres Testing - Datasource fetch table intercept

Mitch-Budibase:
Changing structure for `addDatasourceConfig` in commands.js
- Intercepting the request that is made when fetching tables from a datasource
- There is a specific issue with this in regards to the smoke build. This will allow me to see deeper into the potential timing issues that are occurring


MihailHadzhiev2022 (commented):
Good job!

Issue #5364, opened 2022-04-12
Account for QA Templates in Templates Repository


shogunpurple:
**COS**
- Add a "qa" folder to the budibase templates directory
- Update the `/templates` endpoint to return the QA templates when the `TEST_ENV` environment variable is set
- Update test environment docker-compose file to use the `TEST_ENV` variable

PR #5966, opened 2022-05-18
Return QA templates when environment variable is set

PClmnt:
## Description
QA need to have their own templates folder in which to pull from, specifically limited to their test environment. So this PR introduces a variable (set in the test environments docker-compose) that when set, will ensure that the templates are pulled from a new QA folder (created in the templates repo)

Addresses: 
#5364 

## Screenshots
Not much to demo, but tested with the `TEST_ENV` variable set, and successfully returning the one template I had added to the new QA folder. 

image

### DIFF
--- a/packages/server/src/environment.js
+++ b/packages/server/src/environment.js
@@ -78,6 +78,7 @@ module.exports = {
   ALLOW_DEV_AUTOMATIONS: process.env.ALLOW_DEV_AUTOMATIONS,
   DISABLE_THREADING: process.env.DISABLE_THREADING,
   SQL_MAX_ROWS: process.env.SQL_MAX_ROWS,
+  TEST_ENV: true,
   _set(key, value) {
     process.env[key] = value
     module.exports[key] = value
--- a/packages/server/src/environment.js
+++ b/packages/server/src/environment.js
@@ -78,7 +78,7 @@ module.exports = {
   ALLOW_DEV_AUTOMATIONS: process.env.ALLOW_DEV_AUTOMATIONS,
   DISABLE_THREADING: process.env.DISABLE_THREADING,
   SQL_MAX_ROWS: process.env.SQL_MAX_ROWS,
-  TEST_ENV: true,
+  TEST_ENV: process.env.TEST_ENV,
   _set(key, value) {
     process.env[key] = value
     module.exports[key] = value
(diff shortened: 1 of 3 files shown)


packages/server/src/environment.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
   ALLOW_DEV_AUTOMATIONS: process.env.ALLOW_DEV_AUTOMATIONS,
   DISABLE_THREADING: process.env.DISABLE_THREADING,
   SQL_MAX_ROWS: process.env.SQL_MAX_ROWS,
+  TEST_ENV: process.env.TEST_ENV,
```
Rory-Powell, line 81: It might be nice if we can create a more multi purpose flag so that we can do similar changes in other environments. e.g. a var called `DEPLOY_ENV`  with the value `qa`, this could be extended to `dev`, `staging` if we wanted to use custom templates there in future.

The usage could then look more like the below, so that we don't need to add a new env var each time. 
```js
let type = "app"
if (env.DEPLOY_ENV === "qa") {
   type = "qa"
}
if (env.DEPLOY_ENV === "dev") {
   type = "dev"
}
```

Another option would be to just go very specific and use a variable `TEMPLATE_REPOSITORY` that defaults to `app` when not set and it's value would be used directly in determining the type. Both would work, and the current solution is equally fine, I'm just not a fan of introducing a single flag for indicating if it is test env or not.

PClmnt: Yea totally agree, that's a good point! Went with the latter option, which I think is a bit nicer than having to add a new conditional for every environment.

packages/server/src/environment.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
   ALLOW_DEV_AUTOMATIONS: process.env.ALLOW_DEV_AUTOMATIONS,
   DISABLE_THREADING: process.env.DISABLE_THREADING,
   SQL_MAX_ROWS: process.env.SQL_MAX_ROWS,
+  TEMPLATE_REPOSITORY: process.env.TEMPLATE_REPOSITORY,
```
Rory-Powell, line 81: something we want to promote more in future is default undefined values in this file directly so it's clear to see the state of the env in one place. This could become
```js
TEMPLATE_REPOSITORY: process.env.TEMPLATE_REPOSITORY || "app"
```
and then the var can just be referenced in the consumer

Rory-Powell (approved):
Left one other comment about default values, otherwise looks great! 👍

Issue #5794, opened 2022-05-10
Cypress Nightly Build Detail Breakdown


shogunpurple:
**COS**
- Add link to cypress run in nightly build discord notification
- Post table of results in discord notification, passed, failed, how long to run etc

PR #5970, opened 2022-05-18
Qa/smoke test report

shogunpurple:
## Description
This PR contains improvement and a much better breakdown of the nightly cypress test reports.

The nightly run will now use `mochawesome` to record a breakdown of the runs and write the result as JSON as well as generating a full HTML report. There is now a node script that will fire after the tests have run and the reports have been generated, to push the notification to discord.

I've updated the `smoke_test.yaml` CI job to stop using the old discord notification, and to make use of this. I've also added an `upload-artifact` action, which takes the HTML report and uploads it to the github actions run, so anyone can pull the `.zip` and look at the report in their browser. In future we could actually publish these files directly from the job and have them hosted somewhere (S3 or similar) and just visit a URL to have them served directly.

Addresses: 
- https://github.com/Budibase/budibase/issues/5794

### HTML Report
Screenshot 2022-05-18 at 20 00 43

### Discord Notification
Screenshot 2022-05-18 at 20 03 32


packages/builder/package.json
```diff
@@ -13,11 +13,12 @@
     "cy:setup:ci": "node ./cypress/setup.js",
     "cy:open": "cypress open",
     "cy:run": "cypress run",
-    "cy:run:ci": "xvfb-run cypress run --headed --browser chrome",
+    "cy:run:ci": "cypress run --headed --browser chrome --spec  cypress/integration/createApp.spec.js",
```
shogunpurple, line 16: The `createApp.spec.js` is just temporary here - will remove after I'm done testing on `develop`

Rory-Powell (approved):
This is awesome, nice job

PR #5978, opened 2022-05-19
Lead Generation Form Template Test + Refactor

Mitch-Budibase:
New Lead Generation Form template test
- Verifies app creation
- Successful Publish
- Published app opens when button clicked

Template Testing Refactor
- Created a command to publish app and either click view or done button afterwards (Refactored into some template tests)
- Also directly visiting the templates URL across all template tests

### DIFF
--- a/packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
+++ b/packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
@@ -8,13 +8,13 @@ filterTests(["all"], () => {
     before(() => {
         cy.login()
         cy.deleteApp(templateName)
-        cy.visit(`${Cypress.config().baseUrl}/builder`, {
+        // Template navigation
+        cy.visit(`${Cypress.config().baseUrl}/builder/portal/apps/templates`, {
             onBeforeLoad(win) {
                 cy.stub(win, 'open')
             }
         })
         cy.wait(2000)
-        cy.templateNavigation()
         })
 
     it("should create and publish app with Job Application Tracker template", () => {
--- a/packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
+++ b/packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
@@ -35,19 +35,10 @@ filterTests(["all"], () => {
             cy.get(".spectrum-Button").contains("Create app").click({ force: true })
         })
 
-        // Publish App
+        // Publish App & Verify it opened
         cy.wait(2000) // Wait for app to generate
-        cy.get(".toprightnav").contains("Publish").click({ force: true })
-        cy.get(".spectrum-Dialog-grid").within(() => {
-            cy.get(".spectrum-Button").contains("Publish").click({ force: true })
-        })
-
-        // Verify Published app
-        cy.wait(2000) // Wait for App to publish and modal to appear
-        cy.get(".spectrum-Dialog-grid").within(() => {
-            cy.get(".spectrum-Button").contains("View App").click({ force: true })
-            cy.window().its('open').should('be.calledOnce')
-        })
+        cy.publishApp(true)
+        cy.window().its('open').should('be.calledOnce')
     })
 
     it("should add active/inactive vacancies", () => {
--- a/packages/builder/cypress/integration/templates/marketing/leadGenerationForm.spec.js
+++ b/packages/builder/cypress/integration/templates/marketing/leadGenerationForm.spec.js
@@ -1,0 +1,44 @@
+import filterTests from "../../../support/filterTests"
+
+filterTests(["all"], () => {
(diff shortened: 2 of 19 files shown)


MihailHadzhiev2022 (approved):

packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
```diff
@@ -8,13 +8,13 @@ filterTests(["all"], () => {
...
         cy.login()
         cy.deleteApp(templateName)
-        cy.visit(`${Cypress.config().baseUrl}/builder`, {
+        // Template navigation
+        cy.visit(`${Cypress.config().baseUrl}/builder/portal/apps/templates`, {
             onBeforeLoad(win) {
                 cy.stub(win, 'open')
             }
         })
         cy.wait(2000)
```
Rory-Powell, line 17: It could be worth checkout out timeouts instead adding constant wait times, would mean tests will complete faster as the system performs well and vice versa, giving us a good indication of any sluggishness at times. 

https://stackoverflow.com/questions/60063879/how-to-wait-for-element-to-be-visible

If we started using these a bit more across the board we might be able to bring the build times down quite a bit

packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
```diff
@@ -35,19 +35,10 @@ filterTests(["all"], () => {
...
-            cy.get(".spectrum-Button").contains("Publish").click({ force: true })
-        })
-
-        // Verify Published app
-        cy.wait(2000) // Wait for App to publish and modal to appear
-        cy.get(".spectrum-Dialog-grid").within(() => {
-            cy.get(".spectrum-Button").contains("View App").click({ force: true })
-            cy.window().its('open').should('be.calledOnce')
-        })
+        cy.publishApp(true)
```
Rory-Powell, line 40: Is it a bit redundant to do publish tests here? As long as the app can be created I think that's more than enough, the publish shouldn't be affect by the import source

packages/builder/cypress/integration/templates/marketing/leadGenerationForm.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+        // Create App
+        cy.get(".spectrum-Dialog-grid").within(() => {
+            cy.get(".spectrum-Button").contains("Create app").click({ force: true })
+        })
+
+        // Publish App & Verify it opened
+        cy.wait(2000) // Wait for app to generate
+        cy.publishApp(true)
+        cy.window().its('open').should('be.calledOnce')
```
Rory-Powell, line 41: Similar to above - having the publish here feels like it might be unnecessary and add extra time to the builds

Rory-Powell (commented):
I noticed that all these tests are quite similar and there's a lot of files here. We could probably be accomplished in a different way by having a single test that has a config i.e. the expected result from `api/templates`, and have that test iterate over each config and apply the necessary assertions. 

TBH I don't think we should be doing this work via cypress tests, or at least by using the ui to do it. It's fairly redundant doing this for so many templates, we could just fire the api request and apply assertions on the results of the template configs. Although it would be good to have a single test that does visit the templates page, picks a template and creates an app from it, but just one.

packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
Mitch-Budibase: I agree, I have done this in a couple of places across the tests. There will be a refactor including these timeouts. They will be particularly useful for the nightly smoke build since it has the most issues with timing.

packages/builder/cypress/integration/templates/HR/jobApplicationTracker.spec.js
Mitch-Budibase: Yeah I can understand where you are coming from. In terms of these templates we should just ensure the app is being created. The idea was to ensure that clicking 'View app' after publishing opened the app in a new tab. I will instead have one test that does this for a generic app and refactor those template tests

Mitch-Budibase:
> I noticed that all these tests are quite similar and there's a lot of files here. We could probably be accomplished in a different way by having a single test that has a config i.e. the expected result from `api/templates`, and have that test iterate over each config and apply the necessary assertions.
> 
> TBH I don't think we should be doing this work via cypress tests, or at least by using the ui to do it. It's fairly redundant doing this for so many templates, we could just fire the api request and apply assertions on the results of the template configs. Although it would be good to have a single test that does visit the templates page, picks a template and creates an app from it, but just one.

You are correct in that the api is the way to go here. Attempting this via Cypress has become bulky and unreliable (there are test issues using the UI of the templates that would likely not occur otherwise).

There are existing tests which cover the existence of templates and creates an app using one of them. I will keep those and refactor the rest

Issue #5910, opened 2022-05-16
Make 'use bindings' persist in automation


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: v1.0.145
  - App Version: v1.0.145

**Describe the bug**
Automation 'Use bindings' defaults to 'Use values' when step is collapsed and re-opened.
Raised in this discussion: https://github.com/Budibase/budibase/discussions/5892

**To Reproduce**
1. Add an automation with 'Update Row' step
2. Select 'Use bindings'
3. Collapse the step
4. Open the step
5. The options picker has reset to 'Use values'

**Expected behavior**
If 'Use bindings' is chosen for an automation step then that choice should be saved, even when the step is collapsed and reopened.

PR #5980, opened 2022-05-19
Make 'use bindings' persist in automation

melohagan:
## Description
rowControl (Use values/Use bindings) is now stored on a per step basis, rather than per automation.

Addresses: 
- https://github.com/Budibase/budibase/issues/5910

## Screenshots
bug


PClmnt:
Talked offline about this but I had pushed up a fix for this already, so think we can close this PR out. Should have assigned the ticket to myself. Sorry!!

Issue #5491, opened 2022-04-19
Table row selection not working with Data Provider


nilpntr:
**Hosting**

- Self
  - Method: k8s
  - Budibase Version: 1.0.113

**Describe the bug**
When creating a table with a data provider (to use a custom postgres query), I can't use the table "Allow row selection" feature properly. When I select a row every row is getting selected. Maybe this is because the table doesn't know which field from the query result it needs to use?

**Expected behavior**
Should be able to select each row individually.

**Screenshots**
<img width="1369" alt="Screenshot 2022-04-19 at 18 09 20" src="https://user-images.githubusercontent.com/53516395/164047932-724f4309-713d-4837-9a10-c071462dd6a6.png">

melohagan:
Another report of the same issue from this discussion: https://github.com/Budibase/budibase/discussions/5516

aptkingston:
I knew this would be raised eventually yea. Row selection only works with internal tables or SQL tables, not custom queries. We shouldn't allow row selection (or the export data action) if using a custom query.

PR #5983, opened 2022-05-19
Misc bug fixes

PClmnt:
## Description

Addresses: 
#5491 - Partially addresses, adds some context that informs the user that row selection can only be used with internal tables and MySQL.
#5980 - Ensures binding / values selection persists within automations

Also fixes an issues with automation looping where the user inputted max iterations wasn't being parsed correctly. 

image

image

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItem.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItem.svelte
@@ -28,7 +28,6 @@
   let blockComplete
   let showLooping = false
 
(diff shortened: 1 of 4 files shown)


aptkingston (commented):
Nice one Peter! The only other thing I would add as part of the row selection fix is actually disabling row selection (in the client table component) if they still try to use it with an unsupported datasource.

Should be as easy as:
```js
$: table = dataProvider?.datasource?.type === "table"
...
<Table
  ...
  allowSelectRows={allowSelectRows && table}
  ...
>
```
in `client/src/components/app/table/Table.svelte`.

aptkingston (approved):

PR #6007, opened 2022-05-20
Postgres connection - fixing sporadic issues

mike12345567:
## Description
Issue found by @melohagan - in local development both he and I were experiencing issues connecting to Postgres, it could occasionally pull the data but would experience issues with the connection closing in some form.

I've updated the connection management to be within the async functions it is used in, rather than as part of the constructor, as well as managing the errors that can come out of the connection closing incorrectly. Using a callback for the connection close seems to offer more control in this manner.

Using a function to check if the connection is open pre-query, then opening and waiting for it to open successfully, as well as waiting for it to close successfully, seems to stop these issues from occurring.

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -41,8 +41,7 @@
   "devDependencies": {
     "ioredis-mock": "^5.5.5",
     "jest": "^26.6.3",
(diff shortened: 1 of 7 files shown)


shogunpurple (approved):
LGTM, should sort a lot of those connection issues out 👍

PR #6021, opened 2022-05-22
Hotfixes/perf bg

shogunpurple:
## Description
Performance optimisations for some issues discovered by a large customer. The main bottleneck was calls to `/checklist`, which happen a lot and query all the users in the database. This causes a considerable slowdown on large instances with many users.

Also includes a bulk fill script written by @mike12345567, which allows us to quickly write users to couch in ⚡ speed. 

Addresses: 
- ...more if required

## Screenshots


mike12345567 (approved):
LGTM!

Two notes for what needs done after this for completeness:
1. The user UI in the portal doesn't use pagination, its basically un-loadable at this scale, would be easy to fix this, just worth remembering.
2. The sync, realistically we should make it possible run a sync in the background via button in the UI somewhere, its useful for the builder to have some control here.

Issue #5621, opened 2022-04-27
Deprecate Layouts


shogunpurple:
**COS**
- In app, show something that tells the user that layouts will be deprecated by the end of June, and point them to some documentation around the changes that we are planning on making

Issue #5195, opened 2022-03-31
UI redesign: Add screen switcher above preview


AndyWhann:
**COS**

- Add a dropdown option above the preview that allows the user to switch to another screen
- The dropdown options should include the icon and URL path
- The user can switch to another screen whilst in any other section (Components, theme, navigation)
- When the user is in another section e.g components, when they use this screen to switch screens, they should remain in components but the preview etc will update to show the selected screen

XD -https://xd.adobe.com/view/76a76a26-9390-446c-a6eb-f083ce3865d8-23f4/

Issue #5092, opened 2022-03-25
UI Redesign: Screen slot


AndyWhann:
## COS
- Separate width for the screen slot from navigation
- Remove screen slot display settings from the screen level and place a top level component in the tree
- Create an empty state for when there are no components in the screen slot with a CTA to add component
- Rename "Screen slot" to "Screen"

XD - https://xd.adobe.com/view/40ff803c-3ca2-4c0f-9d45-d3be4fa5a843-15ac/

Issue #5053, opened 2022-03-23
UI Redesign: Navigation Configuration Panel


shogunpurple:
**COS**
- Remove old layouts UI
- Add new navigation panel when you click the navigation icon in the left side nav
- Keep an existing Layouts tab (read/delete only) to access the old UI, to ease migration
- Only show this UI to users before a certain version.

#### Navigation Options
- Configure navigation links button (opens drawer)
- Add conditional dropdown to only show links to certain roles
- Top/Left Nav
- Sticky Header
- Logo
- Title
- Width
- BG Colour (Moved from theme)
- Text Colour (Moved from theme)

#### Per Screen Config
- In screen settings, add toggle setting to hide the nav
- add toggle setting to hide the footer

**Events**
- `budibase:layout_created` - add to current UI, to track if people create layouts
- `budibase:navigation_setting_updated` - When links or other navigation settings are updated. Include metadata for the setting and its value
- `budibase:footer_setting_updated` - When links or other footer settings are updated. Include metadata for the setting and its value

**Designs**

- User can access the nav settings by clicking on the nav in the preview
- Hide and show relevant settings depending on the options the user has set 
(e.g if logo is disabled, don't show the logo URL field)
  - Logo field
  - Title field
  - Width
  - Sticky header
- User can open links drawer my clicking configure links

XD - https://xd.adobe.com/view/89669b15-221a-4081-a51d-c40e82329fad-dc35/

PR #6046, opened 2022-05-24
New design UI

aptkingston:
## Description
This PR adds the brand new design interface. This includus a lot of significant changes to how apps are built, as well as just a very different visual experience. Everything is fully backwards compatible with old apps.

Todo:
- [x] Add role option to links drawer, as we no longer have layout level conditional UI to control this
- [x] Add an "add component" CTA to empty screens
- [x] Update component settings panel to be more like the preview settings bar (styles part of this shelved for now, can be a separate feature later)
- [x] Decide how to bundle and activate this for certain cloud hosted tenants (Martin)
- [x] Show blocks in components section when searching?

A brief TLDR of the major changes:
- New top navigation bar for the whole builder
- New side navigation for the design section
- New sections for managing screens, components, theme and navigation
- No more layouts for new apps. A migration strategy exists for existing apps.
- Routing totally rewritten. URL and state are always in sync, and there are no more redirect loops. Navigating forward or backwards always works properly.
- New component drag and drop experience
- New add component experience, with search and keyboard shortcuts

## Code cleanup
The code structure for all the design components has been cleaned up a lot. Everything is properly sorted beside the routes where they are used, and some things have been renamed to be more appropriate.

- All design components have been properly sorted into `_components` directories in their appropriate routify routes
- A lot of the existing components have either been refactored into new ones or deleted
- Only a few shared components exist in `builder/src/components/design` where previously there was everything
- A lot of the new sections have been given sensible names so it should be much easier to find where changes need to be made - things like `ComponentListPanel`, `ScreenListPanel`, `NavigationSettingsPanel` etc.

## New top navigation bar
image

- App name is listed in the top left
- Back arrow takes you back to the app list
- Main menu is centered (always centered relative to the screen - a long app name will not shift this off center)
- Tooltips for right icons
- Eye icon for previewing apps

## New design section
image

### New design section side nav
image

- Icons for Screens, Components, Theme and Navigation - in that order
- Layouts icon is added only when an existing app with layouts is opened. New apps will not have this.

### Screens section
Some navigation happens automatically while using the new design section.

- The app preview is shown all the time, for all sections
- Clicking on a component in the preview will automatically direct you to the component settings for that component
- Clicking on the navigation in the preview will automatically direct you to the navigation settings
- Changing the screen picker above the preview will never change the section you're on, but just update the screen you're previewing
- The "add component" button above the preview exists at all times, and will always take you to the new UI for adding a component to the current screen

#### Screen list panel
image

- Screens are searchable by route
- Screens are filterable by role
- All screens are shown by default
- Screens are sorted by role, then by home screens, then alphabetically by folder
- A status light representing the colour of the access level is displayed on the right side of each screen
- Home screens have their icons set to a home icon
- The "more" 3 dots menu appears instead of the role color when hovering

#### Screen settings panel
image

- The screen route is shown as the title of the settings panel
- New home screen checkbox to set a screen as the home screen for a certain role. Checking this box will unset any other home screen for this role.
- Route and access level settings are as before, but the role colours have been added to the access picker.
- Showing the navigation can be turned on or off for each screen using the toggle
- The width of each screen can now be controlled independently of the layout width. The screen width can be chosen here using the width picker.

### Components section
- When a component is selected, both the component tree and the client preview will scroll to reveal it

#### Component list panel
image

- Component list panel is expandable using the arrow icon. This makes it wider, so more components can be displayed for deep trees.
image
- The add button opens the new component UI
- Icons have been added beside each component in the tree
- Arrows for parent components are only shown when hovering over the panel, and are absolutely positioned so they do not cause the panel content to reflow
image

#### New drag and drop experience
Drag and drop has been rewritten to be more like the app preview drag and drop experience.

image

- A green line indicates the target position of the component when dragging above or below another component
- A green border around a component indicated that the component will be dropped inside the target component
- When dropping above or below another component, the parent component is highlighted, giving a clear indication to the user of the exact depth they are dropping the component at
- All indicators are positioned absolutely and do not cause reflow
- When hovering over an invalid component (e.g. itself or one of its own children) then no indicators are shown at all. This makes it much more clear because the indicator will always be accurate if it is shown.

#### Component settings panel
Not many changes for now. This will eventually be updated to use buttons similar to the component settings bar.

- Component name and icon added as the title of the settings section

#### New add component experience
The new add component UI can be managed entirely by the keyboard. An example of adding a typical component might look like:
- Press either "add component" button
- The search input is focused automatically, so type "button" to search
- The first search result, which is the button component, is automatically focused
- Press enter to add the button component

image

- There are 2 sections - one for components and one for blocks
- All components are displayed in new tiles, with their icon
- Components can be tabbed through using the keyboard
- The focused component can be added by pressing enter
- Components can be searching by typing in the search box
- The first search result is automatically focused
- When opening the new add component UI, the search field is immediately focused
- The blocks section looks slightly different and is not searchable, as there are very few blocks:
image
- When adding a component, the settings panel on the right updates to say exactly where the new component will be placed:
image

### Theme section
image

- The existing theme settings from the old theme modal have been pulled into this new panel
- The old theme picker has been refactored and is now a slider from darkest to lightest
- The old button roundness picker has been refactored and now is represented by 4 clickable buttons, each showing their roundness setting
- Navigation background and text colour has been removed as this is now in the navigation section
- The right side panel updates to offer some descriptive text about themes. This might be updated to include some more information about what accent colour means.
image

### Navigation section
image

As custom layouts have now been removed, the navigation section is the place where your application links are configured. Other layout related settings have also been moved here.

- Links can be configured here in the same drawer as always
- The navigation position setting has been updated to be 2 visual buttons rather than a picker
- The sticky header and width settings are only shown for top navigation
- The layout width can be configured independently of the screen width here
- Navigation background colour and text colour can be configured here rather than in theme settings
- The right panel shows some information about navigation settings while in this section
image

### Backwards compatibility
There are a number of measures in place to ensure that apps created before this update will continue to function as normal, and indeed still be updated without being forced to migrate to this new structure with no layouts.

There are 3 main migrations in place. 

#### Application migration (automatic)
One automatic migration is at the application level. The first time an old app is opened in the builder, we intelligently try to copy as many settings as we can from their main navigation layout into the new navigation settings. We copy over the links from the navigation layout, the width, the theme settings, sticky header and position. All of these should be immediately visible the first time the user navigates to the navigation section. 

#### Screen migration (manual)
Each screen must be individually detached from legacy custom layouts. This operation detaches the screen from the custom layout and enabled the usage of the new navigation bar. When an old screen is selected, this migration is highlighted by a warning box in the screen settings panel. Until this migration is completed, the new screen settings will be disabled.

image

When editing navigation settings and previewing a legacy screen that uses a custom layout, this warning will also be displayed in the right side panel. This is important as otherwise it would be confusing as to why the navigaiton is not updating in the preview when you change settings (the reason being that the preview is showing a custom layout, rather than the new navigation component which respects the new settings).

image

#### Custom layout migration (manual)
As users may have been using custom layouts, we need to allow users to keep their work. When custom layouts exist, a new side navigation section for layouts will be displayed.

image

This panel lists all layouts which exist in the app. They can be selected which displays the layout in the app preview, and they can be deleted. These are the only 2 actions which are possible in this panel.

The right side panel contains another warning about custom layouts, informing users that custom layouts have been deprecated. We need to decide on the actual copy to display in this warning.

Below the warning is an explanation of the migration strategy. There is a "copy components" button which will copy the content of the entire layout. This can then be pasted into whatever screens are desired, and then can be managed as normal components. Individual components in layouts cannot be selected or edited - they must first be copied and pasted into a screen using this utility. A notification is displayed indicating that the components have been copied sucessfully.

image

Addresses: 
- #5195 
- #5621 
- #5023 
- #3814 
- #5022 
- #5025 
- #5024 
- #5028 
- #5053 
- #5092
- #5027


PClmnt:
This is incredible, the migration for layouts is so slick, nice one @aptkingston!!

Issue #4552, opened 2022-02-17
Export App Without Rows


shogunpurple:
**COS**
- Allow export of apps without row records
- Use filtered couchDB replication to achieve this
- This should be configurable on the app export as either a toggle or separate modal

dougwettlaufer:
This feature would be really helpful for us. We'd like to build out an app in stage and then export just the app so it can be imported into our production instance. Grafana does something similar where you export the dashboard and it parameterizes the datasources. That way when you import it somewhere else it prompts you to fill in the datasources it needs with what you have set up.

geertvanheusden:
It seems that this useful feature is not yet part of the self hosted version. Will it be integrated there as well?

PR #6059, opened 2022-05-25
Export app without rows

PClmnt:
## Description
This PR adds the ability for a user to export their app without any of the rows in the data section. This is handy, as mentioned in the related issue, for people who are developing on specific environments and want to migrate their app, or even for the sharing of apps without data being included. 

This works by using Couch's filter function, I added the 'ro_ta' prefix to the filter that is called whenever the user unselects the toggle. 

### Addresses: 
#4552 

### Video
https://user-images.githubusercontent.com/5665926/170279771-2f01dd87-61b2-4efa-af6a-7867a446d56d.mp4

### Questions

- Should we add this same functionality to the 'Export Apps' function in the cloud? applying the filter to every app that gets exported?
- Would it be useful to add other filter options here? i.e allow the user to not export automations, would be pretty trivial by just using the `au` prefix in it.


shogunpurple (approved):
LGTM! Although, it seems like this would be simpler if you flipped the boolean and used `excludeRows`, rather than made the current behaviour the one modelled by the boolean.

packages/server/src/utilities/fileSystem/index.js
```diff
@@ -109,21 +110,34 @@ exports.apiFileReturn = contents => {
   return fs.createReadStream(path)
 }
 
+exports.defineFilter = includeRows => {
```
shogunpurple, line 113: Not one for now - but think a constants file with different export mechanisms could be good here - as we will likely add more in future and we would have to keep extending this conditional logic

Issue #6057, opened 2022-05-25
Overview tab, Edit Icon and colour change not getting saved


MihailHadzhiev2022:
**Describe the bug**
After clicking on the icon and selecting changing the colour or select an icon, changes are not saved. 

**To Reproduce**
1. Go to https://test.budi.live/builder/portal/apps
2. Click View on any of the apps(3629)
3. Click on the icon top left corner by the app name(3629)
4. Select a an icon and a colour and click save button 

**Expected behavior**
Changes are saved

**Screenshots**
https://www.loom.com/share/0c8bbb74c2d347a1951d661f8d1a69eb

PR #6060, opened 2022-05-25
Explicitly passing requested appId when calling updateAppPackage

deanhannigan:
## Description
During the update process, the application context was resolving the dev app Id instead of the explicitly requested prod app Id. This change ensures that the requested appId is used if it is supplied.

Addresses: 
- https://github.com/Budibase/budibase/issues/6057


packages/backend-core/src/context/index.js
```diff
@@ -297,6 +297,10 @@ function getContextDB(key, opts) {
       toUseAppId = getDevelopmentAppID(appId)
       break
   }
+
+  //If the appId is specified, override context
+  toUseAppId = opts?.appId ? opts?.appId : toUseAppId
```
Rory-Powell, line 302: I think this might have the potential to cause side effects where a subsequent call to `getAppDB` would now be using the supplied app id instead of the current, dirtying the current context.  It might be safer to wrap the implementation of `updateAppPackage` with `doInAppContext`, which would create a new context stack for the logic being wrapped, this would also make sure the correct tenantId is set according to the appId supplied.

Rory-Powell: best to check with @mike12345567

Rory-Powell: Something like
```js
const updateAppPackage = async (appPackage: any, appId: any) => {
   doInAppContext(appId, async () => {

   }
})

```

deanhannigan: Thanks for looking at this. Yes, I see what you're getting at. I've pushed an update that follows the approach that you suggested. I retested the areas that make calls to `updateAppPackage` and the behaviour appears to be as expected.

packages/server/src/api/controllers/application.ts
```diff
@@ -340,6 +341,7 @@ export const update = async (ctx: any) => {
   const data = await updateAppPackage(ctx.request.body, ctx.params.appId)
   ctx.status = 200
   ctx.body = data
+  console.log(JSON.stringify(data))
```
Rory-Powell, line 344: wee log line

packages/server/src/api/controllers/application.ts
```diff
@@ -502,22 +504,24 @@ export const sync = async (ctx: any, next: any) => {
 }
 
 const updateAppPackage = async (appPackage: any, appId: any) => {
-  const db = getAppDB({ appId })
-  const application = await db.get(DocumentTypes.APP_METADATA)
+  return doInAppContext(appId, async () => {
+    const db = getAppDB()
```
Rory-Powell, line 508: What was the bug related to this out of curiosity?

packages/server/src/api/controllers/application.ts
deanhannigan: Yep, removed. Cheers!

packages/backend-core/src/context/index.js
mike12345567: Sorry just seeing this, yes wrapping the whole API endpoint in another app context is the safest option.

One question I have is how there is a mis-match in app IDs to begin with, does the request contain both the production and development app IDs?

packages/server/src/api/controllers/application.ts
deanhannigan: The app context in the overview screen is set to the dev app, so when the update request is made there it modifies the dev app icon. For published applications, this gives the impression that no change was made.
This change ensures the same update behaviour as the app list. When updating the icon in the app list, the context is resolved to the prod application if the app is published.

packages/backend-core/src/context/index.js
deanhannigan: Hey, thanks for taking a look! Yes, in this instance there was a mis-match. When updating the icon for a published app in the app list, it resolved the context as the prod application. In the overview screen, the context resolves itself as the dev application. 
In order to mirror the same behaviour as the app list I needed to explicitly set the context.

mike12345567: ah I see, it might be worth seeing why the overview screen resolves to the dev app, I'm assuming either the `x-budibase-app-id` header isn't being set correctly and its falling back to the cookie app ID. Sometimes app IDs within the request get ignored for the context management as well, but the easiest way to get around this is to always declare the correct app ID in the `x-budibase-app-id` header.

Rory-Powell (approved):

PR #6062, opened 2022-05-25
Swapping bcryptjs for bcrypt and load testing script

mike12345567:
## Description
This swaps the [bcryptjs](https://www.npmjs.com/package/bcryptjs) package for the C++ [bcrypt](https://www.npmjs.com/package/bcrypt) package for hashing. This automatically provides threading and a faster implementation which stops thread lockup.

## Screenshots
Using the load script, I was able to generate a large spam of login requests which would bog down the worker to the point at which the UI became unresponsive:
image

Once I swapped out the packages under the exact same load the worker was 100% responsive and functional:
image


shogunpurple (approved):
LGTM! Huge improvement 💯

Issue #6019, opened 2022-05-22
Error when changing between values / bindings in automations in specific circumstance


FlaminWrap:
**Hosting**
- Self
  - Method: BudiCLI
  - Budibase Version: 1.0.166
  - App Version: 1.0.166

**Describe the bug**
When changing between Values and Bindings in the update row module of an automation can cause the app to become unresponsive when navigating to the automations section.

**To Reproduce**
1. Create a new app
2. Create two BBDB tables called Advice and AdviceType
3. Switch to the AdviceType table
4. Create a Text column called Type in AdviceType
5. Create a Number column called Saving in AdviceType
6. Switch to the Advice table
7. Create a relationship of One AdviceType row > Mandy Advice rows in the Advice table (call the field AdviceType)
8. Create a Number column called Saving in Advice
9. Create an automation called UpdateAdvice which is triggered by row creation
10. Create a Query Rows module and query the AdviceType table
11. Filter the data with `trigger.row.AdviceType`, Sort Column: AutoID, Sort Order: Ascending, Limit: 1
12. Create an Update Row module
13. Select the Advice table
14. Enter `trigger.row.AdviceType` for AdviceType
15. Enter 100 in the Saving field
16. Switch to Bindings
17. Reload the page
18. Switch to Values
19. See error

**Expected behavior**
To be able to switch between using bindings or values when updating a row.

**Screenshots**
![error](https://user-images.githubusercontent.com/97764630/169699586-00fb1868-cc36-4037-b937-b904befe8478.gif)
![image](https://user-images.githubusercontent.com/97764630/169699796-d5fa6b9d-39e6-4fa0-8250-a1ef754019ec.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 101.0.4951.64

**Additional context**
Exported app
[Test-export-1653229024754.txt](https://github.com/Budibase/budibase/files/8749625/Test-export-1653229024754.txt)

melohagan:
I think this will be fixed as part of this PR: https://github.com/Budibase/budibase/pull/5983

But I can verify that once merged.

PR #6068, opened 2022-05-26
Make sure linkedRows is an array before map

melohagan:
## Description
Unrecoverable automation bug when swapping between relationship binding and value dropdown is now fixed.

Addresses: 
- https://github.com/Budibase/budibase/issues/6019


PClmnt (approved):
LGTM! Nice one Mel!

PR #6074, opened 2022-05-26
App/resource ID breakdown of quotas

mike12345567:
## Description
This is the counterpart PR to the update within the pro package: https://github.com/Budibase/budibase-pro/pull/65

The details of this can be found there, this is simply an update to a few test cases, as well as inputting the IDs where needed for the pro package to be resource aware (for apps it will retrieve the app ID from context).

The only major change required within the main repo was within `syncRows` which is now `syncAppRows` - this migration now handles updating the app quotas as well as the total row quotas.


shogunpurple:
Think this one will need to be updated with the new licensing model

PR #6075, opened 2022-05-26
Fix provisioning in multi tenant environment

Rory-Powell:
## Description
Wrap admin user creation in tenancy context
Fixes issue where cache busting wasn't tenant aware


mike12345567 (approved):
LGTM!

Issue #5346, opened 2022-04-11
Automation trigger not saving multiple values to multi select


melohagan:
**Describe the bug**
Creating a row with a multi select column can only save one value from an array. See this discussion: https://github.com/Budibase/budibase/discussions/5224

**To Reproduce**
1. Create a table with a multi-select column called Destinations. Options: North America, South America, Asia
2. Create a webhook trigger automation
3. Connect the trigger to a Create Row -> Destinations: {{ trigger.Destinations }}
4. Test with payload `{
  "Destinations": [
    "North America",
    "South America"
  ]
}` >> Error. Field not in list
5. Test with payload `{
  "Destinations": [
    "South America"
  ]
}` >> PASS

**Expected behavior**
You should be able to pass multiple array values into a multi select column when creating a row in automation

PR #6109, opened 2022-05-30
Automation validation - convert array string to array type

melohagan:
## Description
When field type for validation is array, then convert string to array if needed.

Addresses: 
- https://github.com/Budibase/budibase/issues/5346


PClmnt (approved):
LGTM! Nice one!

Rory-Powell (approved):

PR #6120, opened 2022-05-30
Graceful server shutdown and cleaning up hooks which keep instance alive

mike12345567:
## Description
Fixing issue with server not shutting down correctly when an error occurs, making sure that everything clears up gracefully.

After testing a bit I found the following components where keeping the server from shutting down:
1. The rate limiting store for the public Koa API (Redis connection)
2. The Bull automation queue and its Redis connection
3. The threads running in the background

I did some testing around this and found that it was safe to finish the shutdown process with a `process.exit()` due to the fact that the bull shutdown waits all jobs that are currently running to finish. This means that all shut downs that follow a graceful process using `SIGTERM` will now gracefully wait for everything to stop within the instance before actually restarting it.

The `process.exit` at the end isn't technically required, but the automation queue is already cleaned up and all operations have ended, the server will stop naturally anyway without the `exit` call, but my concern is around things like stray `setInterval` calls, any timeout or interval could cause the server to not shutdown correctly again - in environments outside of Kubernetes this could be problematic.

In the future if there are any background processes which need to be gracefully shutdown they should be added to the server/workers close operation, so that under any circumstance if the process is ending it will correctly clear up the background operations.

### DIFF
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -12,6 +12,7 @@ const destroyable = require("server-destroy")
 const koaBody = require("koa-body")
 const koaSession = require("koa-session")
 const { passport } = require("@budibase/backend-core/auth")
+const { logAlert } = require("@budibase/backend-core/logging")
 const logger = require("koa-pino-logger")
 const http = require("http")
 const api = require("./api")
@@ -61,7 +62,8 @@ if (env.isProd()) {
 const server = http.createServer(app.callback())
 destroyable(server)
 
-let shuttingDown = false
+let shuttingDown = false,
+  errCode = 0
 server.on("close", async () => {
   if (shuttingDown) {
     return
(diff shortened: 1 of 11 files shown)


shogunpurple (approved):
LGTM! Great work 🚀

packages/worker/src/index.ts
```diff
@@ -62,25 +61,34 @@ if (env.isProd()) {
...
 module.exports = server.listen(parseInt(env.PORT || 4002), async () => {
   console.log(`Worker running on ${JSON.stringify(server.address())}`)
   await redis.init()
 })
 
 process.on("uncaughtException", err => {
   console.error(err)
-  server.close()
-  server.destroy()
+  shutdown()
```
Rory-Powell, line 89: Could we add a “bb-alert:” log line here, so we can send notifications when this happens

Rory-Powell: It might be worth passing an exit code to the shutdown function, so it defaults to 0 but is set to 1 here

PR #6155, opened 2022-06-01
Snowflake Integration ❄️ 

melohagan:
## Description
Added Snowflake as a query datasource

Addresses: 
- https://github.com/Budibase/budibase/issues/4405

## Screenshots
Screenshot 2022-06-01 at 10 21 05

Screenshot 2022-06-01 at 10 21 35

Screenshot 2022-06-01 at 10 20 47

Screenshot 2022-06-01 at 10 22 35

Screenshot 2022-06-01 at 10 24 30


packages/server/src/integrations/snowflake.ts
```diff
@@ -0,0 +1,122 @@
...
+    private client: Connection
+
+    constructor(config: SnowflakeConfig) {
+      this.client = createConnection(config)
+    }
+
+    async connectAsync() {
+      return new Promise((resolve, reject) => {
+        this.client.connect(function (err: any, conn: any) {
+          if (err) reject(err)
```
mike12345567, line 76: Small change, but with `reject`, `resolve` like this I prefer to have braces around everything as its a little less ambigious when reading, e.g.
```
if (err) {
  reject(err)
}
resolve()
```
that way theres no risk of say the `reject(err)` being removed and suddenly the error state now resolves the call.

Also, if you wanted to avoid the need to wrap the SDK with promises, you could use https://www.npmjs.com/package/snowflake-promise

packages/server/src/integrations/snowflake.ts
```diff
@@ -0,0 +1,122 @@
...
+      let response: any = await new Promise((resolve, reject) =>
+        this.client.execute({
+          sqlText: query.sql,
+          streamResult: false,
+          complete: (
+            err: SnowflakeError | undefined,
+            stmt: Statement,
+            rows: any[] | undefined
+          ) => {
+            if (err) reject(err?.message.split(":")[1] || err?.message)
```
mike12345567, line 93: same as above comment.

mike12345567 (changes requested):
Looks good! Just need the fix around the packages and where the SDK is specified as a dependency, as this will break the docker build.

package.json
```diff
@@ -17,6 +17,7 @@
     "prettier-plugin-svelte": "^2.3.0",
     "rimraf": "^3.0.2",
     "rollup-plugin-replace": "^2.2.0",
+    "snowflake-sdk": "^1.6.10",
```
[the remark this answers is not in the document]
melohagan: Removed. Using snowflake-promise instead.

packages/server/package.json
```diff
@@ -158,6 +158,7 @@
     "@types/node": "^15.12.4",
     "@types/oracledb": "^5.2.1",
     "@types/redis": "^4.0.11",
+    "@types/snowflake-sdk": "^1.6.2",
```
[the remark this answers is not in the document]
melohagan: Using snowflake-promise instead.

packages/server/src/integrations/snowflake.ts
melohagan: snowflake-promise is much nicer! Wish I'd seen it in the first place!

packages/server/src/integrations/snowflake.ts
melohagan: Using snowflake-promise instead.

mike12345567 (approved):
LGTM!

Issue #6161, opened 2022-06-01
App dashboard - Move unpublish action 


AndyWhann:
At the moment, unpublish is placed inside the more option in the app dashboard.

We want to remove it from here and place a text link on the App Status card when the app has been published.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/e4549e5a-b853-42ad-803e-07b30f71af98)

PR #6179, opened 2022-06-02
6161 - move the unpublish link

mslourens:
## Description
Move the unpublish link from the dropdown menu to the publish app status panel
Note: I used the default link component which is not grey but blue

Addresses: 
- #6161
 
## Screenshots
image

image

### DIFF
--- a/packages/builder/cypress/integration/appPublishWorkflow.spec.js
+++ b/packages/builder/cypress/integration/appPublishWorkflow.spec.js
@@ -21,7 +21,6 @@ filterTests(['all'], () => {
 
       cy.get(interact.APP_TABLE_ROW_ACTION).eq(0)
       .within(() => {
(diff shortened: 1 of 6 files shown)


packages/builder/cypress/integration/appPublishWorkflow.spec.js
```diff
@@ -20,10 +21,10 @@ filterTests(['all'], () => {
 
       cy.get(interact.APP_TABLE_ROW_ACTION).eq(0)
       .within(() => {
-        cy.get(interact.SPECTRUM_BUTTON_TEMPLATE).contains("Preview")
+        //cy.get(interact.SPECTRUM_BUTTON_TEMPLATE).contains("Preview")
```
shogunpurple, line 24: We should just delete this if it's not required 👍

shogunpurple (approved):
LGTM, just that one comment about the cypress test. Thanks again @mslourens !

Issue #6124, opened 2022-05-31
Document Update Conflict - Internal DB - Static Formulas


aderici:
I have 2 tables with a relationship in between and static formulas in both. I'll call them parent and child. 

Having a static formula field on child makes saving parent rows impossible. Always errors with update conflict. Same behavior in data tab and my app itself. 

Converting the formula field on child to a dynamic one removes the problem but i need it to be static. 

Is this a bug? Any ideas on how can i make it work? 

Using latest version as of today,

aderici:
More diagnostic info : the problem only occurs when child table has 2 static formula fields and only happens when you try to update a parent row. If you change one of the static formulas to dynamic, leaving only one static formula column problem goes away

yeoldegrove:
I have the very same issue as @aderici described. As soon as I get rid of the multiple static formulas columns, the update conflict is gone.

PR #6183, opened 2022-06-03
fix: document update conflict couchdb error

actus-wirtenberger:
## Description
Updating the same document multiple times without renewing the revision after each update caused couchdb to throw an error.

Addresses:
- [Document Update Conflict - Internal DB - Static Formulas #6124](https://github.com/Budibase/budibase/issues/6124)


actus-wirtenberger:
Not sure if making the db updates sequential is the best solution. Is there any semantic reason why the same rows are updated multiple times in the first place?

Issue #1504, opened 2021-05-19
Unable to use datepicker with an external update query


mjashanks:
**Describe the bug**
Originally raised here: https://github.com/Budibase/budibase/discussions/1500

I am trying to use a Budibase datepicker to update a postgres date column.

**My Setup**
- Postgres date column called `created`
- Query, that takes a parameter of `created` and runs and `update` 
- In design, a `Form`, `Field Group`, `DatePicker`. The datepicker's `Field` property is `created` 
- A save button, which executes the query, and passes `{{ Form.Fields.created }}`. Have also tried `{{ date Form.Fields.created  "YYYY-MM-DD" }}` and various other bindings

The request that gets sent by `Execute Query` sends
- `[Object object]` for `{{ Form.Fields.created }}`. 
- `Invalid Date` for `{{ date Form.Fields.created  "YYYY-MM-DD" }}`

aptkingston:
The value of actual date picker (thanks to Flatpickr) is a full JS date object. Not sure how to handle that in Handlebars, but it's not a string so it might not be trivial. We could maybe do something similar to what happens with relationships, where if you bind to a relationship field it sneakily replaces it behind the scenes with the length of the array, rather than the array itself. We could use this same idea and replace date bindings with either the UTC string or the timestamp, making it easy to work with.

shogunpurple:
Bump

mjashanks:
I still have an issue with this. The date picker is timezone aware - I don't think it should be right now.

For example. I set my timezone to UTC -5:00 choose the date `2nd November 2021 17:00`. When I click save, the request sends the date `2nd November 2021 12:00`.

I think it should send `2nd November 2021 17:00` - i.e. we want the ISO format, but no conversion to UTC.

aptkingston:
But if you pick `2nd November 2021 12:00` in your time zone, and then we send `2021-11-02T12:00:00.000Z` then we are not sending what you picked, we're sending a time 5 hours in the past. Reading that date back into Budibase will show it as 5 hours ago, while if we send the correct `2021-11-02T17:00:00.000Z` and read it back into Budibase then it should show exactly the time you picked of `2nd November 2021 12:00`.

What's the issue you're having with it at the moment?

mjashanks:
So - I can see what you mean here. It works because it gets correctly converted on the way in and on the way out. What we are doing is the most correct thing to do.

[...]

PR #6199, opened 2022-06-06
Dates without timezones

aptkingston:
## Description
This PR adds the ability to use dates/timestamps without timezones. This is a really messy topic so there are some caveats and limitations here. This should be fully backwards compatible. There should be no change to existing apps unless you enable this new feature.

Targetting develop as this definitely requires some more manual testing before release. I've tested this as much as possible with multiple server/client/database timezones and multiple database types but I doubt I have accounted for all permutations.

Addresses #1504.

## Other new features
- When generating auto screens or generating form fields inside a field group, all date field metadata (such as "date only", "time only", and the new "ignore time zones" flag) are correctly honoured and passed into the date form fields
- When creating dates through Budibase, we know inform knex whether we wish to use timezone awareness or not. This lets knex pick more accurate underlying types in external databases when creating columns.

## Limitations
- This only works with datasource plus tables (i.e. internal tables, or SQL tables)
- For external SQL tables, this only works with Postgres and MySQL. SQL server and Oracle are not supported. You can read more about why at the bottom of this PR.
- This only works automatically when using date pickers. If you use JS bindings to edit dates before sending them to the server, you must ensure you send up the correct ISO 8601 string *without a timezone suffix*.
- It is possible, but not recommended, to change the "ignore timezones" flag on tables that already have data. Changing this flag does not change existing data, but it will change how your existing data is presented in the browser.
- Not a limitiation as much as a design decision, but this new flag is never set when importing external tables. If you are using non-timezone-aware types, you must edit your columns after import and mark them as such.

## New ignore timezones flag
When creating a new date field in Budibase, or editing an existing one, a new option called "Ignore time zones" is available.

image

When this option is checked, the values selected in date pickers will be represented exactly as they are in the target database. There is no UTC conversion performed by Budibase - but there may be (transparent) UTC conversion done at the database level in certain circumstances (e.g. a `timestamp` in MySQL). Regardless of the timezone the browser is in, you will always see the same exact timestamp when this option is selected. 

As an example, here is a snippet of a table where I changed my browsers locale to 3 different timezones and picked the time `12:00` in each. You can see that the column "BB no timezone" (which has this new option set) will always show the same time time, whereas the "BB normal" column (a normal date/time without this flag) is using UTC conversion and relative dates that do change as the browser timezone changes.

image

### Frontend support
The reason frontend support is required for this feature is that the server is not aware of the client's timezone. If the server was aware of the client's timezone then the frontend could be entirely agnostic of this feature, and could continue to send up UTC timestamps, which could then be converted as required on the server. Since we do not know the client's timezone. we depend on the client sending up a timestamp string which is already formatted in their timezone. We then must ensure this string is preserved right until it is written into the database, and never converted into a date object during server processing.

### SQL server
SQL server is not supported as the `mssql` node library (the underlying driver used by knex when connecting to a SQL server DB) does not support custom type casting. In order for this feature to work we need to be able to retrieve dates as strings from databases. By default knex parses all date types into JS date objects, but this prevents us from being able to preserve the original timezone as that data is effectively lost as soon as it is parsed into a date object.

As we cannot do custom type casting with `mssql`, we cannot support ignoring timezones for SQL server databases as knex will always parse dates as objects.

### Oracle
Oracle is not supported as we cannot write date values as strings through knex into Oracle databases. Oracle requires a specific date format when writing dates. This format can be customised and specifed per connection, and this is handled by knex. As we have no awareness above knex of the date format being used, we must pass dates as real date objects into knex, which will itself correctly serialise these into the correct string format before saving them to the DB.

As we cannot control and do not know this format, we cannot support ignoring timezones as the standard ISO 8601 format without timezone that we use to support this feature is considered an illegal format. And once again, we cannot pass in literal date objects to knex as it will either convert them to UTC or parse them in the servers timezone - but neither will preserve the original timezone.

## Future work
- "Time only" should really be removed as an option
- "Date only" should be an exposed schema-level setting rather than a presentational setting on date pickers. It's already a schema-level property, but only when importing external tables

### DIFF
--- a/packages/bbui/src/Form/Core/DatePicker.svelte
+++ b/packages/bbui/src/Form/Core/DatePicker.svelte
@@ -66,7 +66,8 @@
     else if (!enableTime) {
       const year = dates[0].getFullYear()
       const month = `${dates[0].getMonth() + 1}`.padStart(2, "0")
-      const day = `${dates[0].getDay() + 1}`.padStart(2, "0")
+      const day = `${dates[0].getDate()}`.padStart(2, "0")
+      console.log(year, month, day)
       newValue = `${year}-${month}-${day}T00:00:00.000`
     }
 
--- a/packages/bbui/src/Form/Core/DatePicker.svelte
+++ b/packages/bbui/src/Form/Core/DatePicker.svelte
@@ -15,7 +15,7 @@
   export let placeholder = null
   export let appendTo = undefined
   export let timeOnly = false
-  export let disableTimezone = false
+  export let ignoreTimezones = false
 
   const dispatch = createEventDispatcher()
   const flatpickrId = `${uuid()}-wrapper`
@@ -51,7 +51,7 @@
 
   const handleChange = event => {
     const [dates] = event.detail
-    const noTimezone = enableTime && !timeOnly && disableTimezone
+    const noTimezone = enableTime && !timeOnly && ignoreTimezones
     let newValue = dates[0]
     if (newValue) {
       newValue = newValue.toISOString()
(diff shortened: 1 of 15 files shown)


shogunpurple (approved):
LGTM! Will be awesome to finally put this one to bed. Just one thing - it is probably a better idea to put this one into master, and we can test it in the pre-prod env as it is going to be released in isolation.

Issue #6013, opened 2022-05-20
Automation -> Query Rows does not apply the filter at all


aderici:
Hi,

In a self hosted environment i have 2 internal tables like parent and child. I did not create a typical relationship between them, instead i'm saving parents auto id into child. 

I have a need of deleting children when parent row is deleted. Firstly is there a recommended way of achieving this? 

I have tried an automation, whenever a row is deleted from parent, query related child rows and delete them in a loop. 

The problem is, "filter" is not applied in query rows block. even though it is properly entered and value is properly populating for filter criteria. 

This caused all my rows to get deleted instead of only filtered ones. 

Can you check this bug? I believe it is a serious one,

Best regards,

PClmnt:
Hi @aderici. Sorry that you're having issues with this! Can you give a little more information about what the filter in your Query Rows step looks like?

aderici:
Hey thanks for the quick response, very happy to see that, let me share some screenshots : 


This is the general look, Order is the parent object Sample is the child object : 

<img width="547" alt="image" src="https://user-images.githubusercontent.com/6830693/169623096-bd6c28ec-7b57-4c71-804d-61553a99dfb9.png">

This is my filter for querying samples table : 

<img width="1092" alt="image" src="https://user-images.githubusercontent.com/6830693/169623123-864453b9-227c-4add-868c-fb3347eb53b7.png">

And this is the JS code : 

<img width="1258" alt="image" src="https://user-images.githubusercontent.com/6830693/169623165-7c5a7584-cc62-44aa-9d61-57908559142f.png">

aderici:
Also my version number : 
<img width="195" alt="image" src="https://user-images.githubusercontent.com/6830693/169623236-97c69cad-76bf-4f9e-92db-a25336421c2d.png">

PClmnt:
So I had a quick look at this and tried to reproduce. I _think_ the problem is that `Auto ID` doesn't actually come back as the response from the `Delete Row` automation. It will just return the `tableId` and the user added columns. Auto columns aren't returned. So this meant the rows weren't filtered. 

Could you potentially use your own ID column as an identifier instead of the `Auto ID` column?

aderici:
If you want to use an auto increment it would again become an auto column right?

Do you think it is the right behavior to let all data pass unfiltered if one side of the equation is undefined? rather throw an error and dont pass anything i guess. Otherwise you'd let people see unauthorized data or the whole table deleted like my case.

[...]

PR #6208, opened 2022-06-07
Handle null values when filtering rows

PClmnt:
## Description
Users who tried to use a filter, that included a null or empty value would end up returning every value in that specific table. 

i.e a filter like this 
image

Where the value is empty, or the handlebars or JS expression returns a null value this will return every row in the queried table, this was previously an issue, but has became even more so with the introduction of the loop block, where you can do some serious damage by including a delete row block after a query block whose filter returns null. i.e, delete an entire table. 

This addresses it by returning an empty array if any of the filters in the lucene expression are null. 

Addresses: 
#6013 

n.b also snuck a change in here to increase the size of the automation test panel text box sizes.

### DIFF
--- a/packages/server/src/api/controllers/row/internalSearch.js
+++ b/packages/server/src/api/controllers/row/internalSearch.js
@@ -304,6 +304,26 @@ const recursiveSearch = async (query, params) => {
   if (rows.length > params.limit - 200) {
     pageSize = params.limit - rows.length
   }
+
+  /*
+  If a null value is passed to a filter expression, we currently return 
+  every row, this can and has resulted in data loss via automation loops etc.
+  So here we are checking if any null values exist in the filters, and if they do
+  just return an empty rows array.
+  */
+  let nullValuesExist = false
+  Object.keys(query).forEach(key => {
+    Object.values(query[key]).forEach(innerVal => {
+      if (innerVal === "" || innerVal === null) {
+        nullValuesExist = true
+      }
+    })
+  })
+
+  if (nullValuesExist) {
+    return rows
+  }
+
   const page = await new QueryBuilder(query)
     .setVersion(params.version)
     .setTable(params.tableId)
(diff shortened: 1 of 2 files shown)


packages/server/src/api/controllers/row/internalSearch.js
```diff
@@ -304,6 +304,26 @@ const recursiveSearch = async (query, params) => {
...
   }
+
+  /*
+  If a null value is passed to a filter expression, we currently return 
+  every row, this can and has resulted in data loss via automation loops etc.
+  So here we are checking if any null values exist in the filters, and if they do
+  just return an empty rows array.
+  */
+  let nullValuesExist = false
+  Object.keys(query).forEach(key => {
```
aptkingston, line 315: Since this loops over every operator, this means that if you wanted to do "not equals empty string", or "not equals null" that those are now impossible too since you'll just get back no results. Some operators don't have values either (like "is empty" or "is not empty" - but I think since those should be `undefined` and you're doing an explicit `=== null` that hopefully those still work. Could you double check that to make sure?

PClmnt: Yea I see your point on the first one, I don't really see another way to handle this other than potentially only doing the null check on specific operators, maybe `string` `fuzzy` `equal`?

aptkingston: It's a tough one because I see what you're saying that when doing things like looping in automations with a delete row action or something, that you want to be conservative. But 95% of the time you're just pulling data, for which you would want to do the opposite and pull back more data than necessary rather than less. What was the original issue that led to data loss? I'm wondering if we can do something more specific around whatever edge case caused that.

PClmnt: You can see the original issue here: https://github.com/Budibase/budibase/issues/6013 using the IDs of a query block in automations in a delete block loop.

Issue #6215, opened 2022-06-07
Upload to s3 - CSP error


PClmnt:
### Discussed in https://github.com/Budibase/budibase/discussions/6204

<div type='discussions-op-text'>

<sup>Originally posted by **SantoshSrinivas79** June  7, 2022</sup>
I am trying to upload to AWS s3 in `us-east-1` on a self-hosted instance. Please advise

Getting the following errors in my console:

```
budibase-client.js?v=1.0.194:243 Refused to connect to 
**********************************
because it violates the following Content Security Policy directive: "connect-src 'self'
```

```
budibase-client.js?v=1.0.194:243 Refused to connect to 
***************************************
 because it violates the document's Content Security Policy.
```

<img width="701" alt="Screenshot 2022-06-07 at 9 39 42 AM" src="https://user-images.githubusercontent.com/1036163/172294710-41de3ee8-8330-4e10-8a5d-a733a847bd14.png"></div>

We need to add a new wildcard to the nginx config to handle `tweeterati.s3.amazonaws.com`

*COS*

- CSP is updated with the correct url and wildcard

PR #6217, opened 2022-06-07
Update csp to account for certain s3 addresses

PClmnt:
## Description
Issue raised from a discussion where someone mentioned they were running into a CSP violation while trying to use the following address `https://tweeterati.s3.amazonaws.com`. This PR updates the CSP to allow for URLs of that type. 

Addresses: 
- #6215


shogunpurple (approved):
LGTM!

Issue #6209, opened 2022-06-07
Clicking table action buttons will also select checkbox unintentionally


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6203

<div type='discussions-op-text'>

<sup>Originally posted by **magickaito** June  7, 2022</sup>
![Screenshot 2022-06-07 at 11 36 06 AM](https://user-images.githubusercontent.com/1195573/172290446-91b9eafc-2079-476c-99c8-40a82c40484b.jpg)

When clicking the action button in a table row (like the checkmark button), the row checkbox will also be selected, which feels like a bug. 
Is there a way to prevent this?</div>

PR #6220, opened 2022-06-07
Don't allow row selection for buttons in rows 

melohagan:
## Description
Clicking a row button no longer causes the row to be selected.
I did try using the *self* event modifier, but it prevented the row from being clicked at all. 

Addresses: 
- https://github.com/Budibase/budibase/issues/6209

## Screenshots
clickrowbug

### DIFF
--- a/packages/bbui/src/Table/Table.svelte
+++ b/packages/bbui/src/Table/Table.svelte
@@ -132,7 +132,7 @@
       style += " auto"
     }
     fields?.forEach(field => {
-      const fieldSchema = schema[field]
+      const fieldSchema = schema[field.name]
       if (fieldSchema.width) {
         style += ` ${fieldSchema.width}`
       } else {
@@ -204,7 +204,6 @@
         return nameA < nameB ? a : b
       })
       .concat(autoColumns)
-      .map(column => column.name)
   }
 
   const editColumn = (e, field) => {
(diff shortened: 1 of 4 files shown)


packages/bbui/src/Table/Table.svelte
```diff
@@ -28,7 +28,7 @@
   export let rowCount = 0
   export let quiet = false
   export let loading = false
-  export let allowSelectRows = true
+  export let allowSelectRows
```
aptkingston, line 31: Any reason in particular you changed the default value for this? We would just need to be careful there are no usages anywhere in the platform that depend on the existing default value.

packages/bbui/src/Table/Table.svelte
```diff
@@ -373,6 +369,16 @@
                 class="spectrum-Table-cell"
                 class:spectrum-Table-cell--divider={!!schema[field].divider}
                 style={cellStyles[field]}
+                on:click={() => {
+                  if (!field.startsWith("custom-")) {
```
aptkingston, line 373: Could we add a schema property to control this rather than depending on field name (which has the potential to cause false positives as unlikey as that is)? 

I was thinking we could add a `unselectable` property to the field schema, and if that new property is true then we skip the row selection here.

All we would need to do to implement that is in the client table component, add in `unselectable: true` when we construct the schema for the custom column, around line 85, and then check here
```js
if (!schema[field]?.unselectable) { ... }
```

There's also no reason for this to be 2 separate on:click handlers any more -  we can just execute both functions in the same handler now to save duplication,

aptkingston (commented):
Nice one Mel! Just left a comment about another way we could do this that would be a little cleaner than depending on field names.

packages/bbui/src/Table/Table.svelte
melohagan: That was a separate bug fix, forgot to mention. This was a fix for tables that were defaulting to showing the selection checkboxes, despite the fact that 'Allow row selection' was not checked.

The manifest contains a default value of false.

packages/bbui/src/Table/Table.svelte
aptkingston: Yea that name is fine.

Not sure about the refactor around the fields though... what motivated you to do that? There's a bit of a strange mixture of `field` being the full field schema now, but yet there are many usages of `schema[field.name]` which are then redundant because that's just the same value as `field`. I would just have kept it as it was unless there's a reason to refactor it? Generally it's best to avoid changes to really core components like the table unless it's necessary.

aptkingston: The `preventSelectRow` was in the schema though - so you could have accessed it like I mentioned in my original suggestion by just `schema[field]?.preventSelectRow`. We already access the field schema to pass into the `CellRenderer` just below that by `schema={schema[field]}`.

melohagan: No you're absolutely right - you even provided the code snippet in the first place which I managed to miss, apologies!
I'll get that sorted.

aptkingston (approved):
LGTM! 👍

PR #6269, opened 2022-06-10
Fix newline bug for Firebase private key

melohagan:
## Description
Authentication was failing because the newline character was being escaped. 
Fixed by replacing '\\\n' with '\n'.
Also removed **ServiceAccount** as this is a confusing property and doesn't save much effort for three fields.

## Screenshots
Screenshot 2022-06-10 at 14 17 58

Screenshot 2022-06-10 at 14 18 29


shogunpurple (approved):
LGTM! Nice work Mel

Issue #6163, opened 2022-06-01
Apps page - CTA buttons


AndyWhann:
- Increase the size to medium
- Make Edit the primary action
- Add 3rd CTA for dashboard
- Remove more option

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/8c26d785-83bd-48e3-8229-841726027664)

Issue #5819, opened 2022-05-11
"Basic user onboarding", The password is not visible enough 


MihailHadzhiev2022:
**To Reproduce**
1. Click "Users" and add a new user using "Basic user onboarding"

**Expected behavior**

- The password should be visible and clear

![Screenshot 2022-05-11 at 14 01 16](https://user-images.githubusercontent.com/102024164/167834698-c6fa8e8e-99ee-4d3b-b128-1646277503c3.png)

PR #6286, opened 2022-06-13
App list updates + small stuff

aptkingston:
## Description
Few small features and issues fixes:
- Update basic user onboarding password to be readonly rather than disabled to be visible. Also just make disabled states slightly more readable in general #5819
image
- Update app list with new button styles and layout #6163. These don't match the issue description because I went through a few additional iterations with Andy.
image
- Removed excess padding and context menu from app list
- Clicking on the app name in the app list defaults to editing the app
- Update app list text to be all styled the same
- Add `newStyles` prop to BBUI button to allow styling buttons with the "new" spectrum styles for primary and secondary
- Fix a crash when running automations with an app action trigger that has no inputs


shogunpurple (approved):
LGTM!

Issue #4776, opened 2022-03-03
Guide user to required fields for components


AndyWhann:
### Problem

When adding an element like a `text field` to an app, the first thing we present is placeholder text that says "Add the Field setting to start using your components". 

The reason this is a problem is for new users who aren't familiar with the settings, they have to go looking for that settings.

### Solution

1. Highlight the field with a border until the user sets the field
2. Auto focus the required field after the user adds the element
3. On the placeholder text, highlight the setting label so it stands out from the rest of the placeholder text
4. Add a link to the placeholder text that the user can click to find the field and focus it for them

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/f1476103-0288-4a24-9e02-8a1349c5f470)

PR #6309, opened 2022-06-14
Builder focus updates

aptkingston:
## Description
These are a few updates to the new builder focus feature after some conversations with @deanhannigan and @AndyWhann.

Original issue: https://github.com/Budibase/budibase/issues/4776
Original PR: https://github.com/Budibase/budibase/pull/5937

Highlighted settings now have a different background colour and a blue border. This generic approach works for any setting type and does not require changes to the underlying components. They are now also scrolled into view if required.

image
image

- Highlighted fields are scrolled into view if required
- The idea of focusing fields has been replaced with highlighting fields
- The core `Component.svelte` wrapper is in charge of determining whether required settings are missing, meaning there is no need for components themselves to be aware of required settings
- Highlighted settings are controlled by the builder, removing the requirement for all settings types to be aware of the focus concept
- Removed focus specific logic from core BBUI components
- Added default text for buttons
- Fixed default values not being passed properly to `PropertyControl`
- Prevent inline editing of components which are missing required settings

### DIFF
--- a/packages/bbui/src/ActionButton/ActionButton.svelte
+++ b/packages/bbui/src/ActionButton/ActionButton.svelte
@@ -1,8 +1,7 @@
 <script>
   import "@spectrum-css/actionbutton/dist/index-vars.css"
-  import { createEventDispatcher, getContext } from "svelte"
+  import { createEventDispatcher } from "svelte"
   const dispatch = createEventDispatcher()
-  const context = getContext("builderFocus")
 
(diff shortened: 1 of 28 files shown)


aptkingston:
Going ahead and merging this in as it's blocking me being able fix tests on other branches, and it's pretty safe. I've also already been over this with @deanhannigan privately and @AndyWhann for the designs.

Issue #3884, opened 2021-12-29
Expose Current User in data sources


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3879

We should allow the use of `{{ Current User }}` inside datasource custom queries, REST calls.

Dominic-Wagner:
Why got this closed?
Current User is not implemented for Custom SQL Queries

Issue #4139, opened 2022-01-21
Use context bindings in REST API (and other custom queries)


mjashanks:
We should add the ability to use the following bindings in all custom queries. 

- `{{ Current User....}}` : logged here already https://github.com/Budibase/budibase/issues/3884
- Access headers on incoming requests: https://github.com/Budibase/budibase/discussions/3934
- Properties from the SSO token: https://github.com/Budibase/budibase/issues/3827
- Eventually, we would like to put environmental things in here - e.g. config and secrets for dev/staging/prod

I thought it would be better to collect these into one issue.

We would need to think about how these work in automations - i.e. is the "User" available, or does it makes sense?

Commented on here: https://github.com/Budibase/budibase/discussions/4120#discussioncomment-2007650

mike12345567:
Just a point, the first one using `Current User` in datasources/automations could be quite difficult as they are server-side, its possible we could enrich it from the user session for datasources, but automations are ran outside the context of a request so won't have a current user for context.

thehowl:
Bumping this so that stalebot doesn't close this, as I think it's quite important to have.

While writing this comment, I've realised many others have commented on the importance of a similar feature on #3827. So I'll just collapse it here:

<details>
<summary>Explanation of use case and argument on the importance of this feature</summary>
<p>So is there currently no way of ensuring some kind of "workspace" access to a data source server-side?
My use case is building an app for a service which is being sold to different clients. We want each client to have access to their data exclusively, without being able to see data of other clients. However, from what I understand, this is currently only possible to send as a parameter of a query to a data source? So client-side?<p>

<p>It seems to me that the only current way would be to duplicate the app and making sure that the requests to the backend always specify the correct workspace, which is of course undesirable as I don't want to maintain an app for each client and have to export/import/change the datasource queries each time. I'd like to make the case that this is quite critical in any scenario where we want access to data to be secured in some kind of way which is outside of the "access level" system, because otherwise the queries to datasources are just waiting for any techie to come by and realise that access is unsecured server-side.</p>
</details>

I understand you have your own priorities and roadmap for the project where this has not yet come up as an important issue -- judging from the fact that this issue has been open for the past 6 months and was about to be closed in five days' time.
I'd be willing to work on this to address at least the `{{ Current User }}` variable in the REST datasource - which is my own itch - and properties from the SSO token which I see has been highly requested and without having seen the source code I'd intuitively think is a trivial change to add if I'm also adding data of the current user.

[...]

Issue #3827, opened 2021-12-17
Access the SSO token and use in REST requests


mjashanks:
Spec: https://docs.google.com/document/d/1U_XY-A2vxRub0SMk7vU6at15GA1LSyz9XUyNL9mC3QQ/edit#

When I'm using the REST connector in budibase, I want to have access to the currently logged in users SSO token so that I can authenticate with my API.

**COS**
- Any SSO enabled orgs developer users will be able to use the SSO token in their REST Connector
- A dynamic variable named budibase_user_sso_token will be automatically injected into every instance of a REST API connector in budibase
- SSO providers that return a refresh token must be refreshed in the background to ensure they don’t expire

brooth:
Budibase looked as a very good product until I saw this 😐 I expected to see that it refreshes the tokens automatically but turns out that I even cannot access id/access tokens to make requests.  Not good

mjashanks:
Also, raised in https://github.com/Budibase/budibase/discussions/3934, it'd be useful to be able to bind to incoming request headers.

jlangree:
Is this something that's being worked on? Are there any updates on when it might be ready? This feature would make a huge difference.

jlangree:
@mjashanks Are there any updates on this?

AsherDeVries:
This feature is absolutely mandatory for our company (and I assume for others as well). It's the only way to integrate with our existing api's in a safe and secure way. 

@mjashanks is this still on your roadmap?

niggydotcom:
Another upvote for this please. Our REST API uses Azure SSO exclusively now so the project has stalled

trialsin:
Same here! We need that too!

mlomax:
Hi 
I'm not a coder but am trying to instruct the development of a budibase app.
It will be using REST APIs to connect to our datasource.
The REST APIS use a microsoft SSO token to authenticate - there is no option to use anything different.

I want to log on to BB using my microsoft SSO and use the returned token in my datasource 

@shogunpurple @mjashanks Is this development work what is scheduled for June as per #4139 

As this capability doesn't yet exist is there a current work around that people are using?

Many thanks

PR #6315, opened 2022-06-14
REST query bindings for Current User

### LINKED ISSUE #6366: REST API Query - raw XML body not parsing correctly.
**Describe the bug**

Raw XML body text is failing to parse correctly resulting in a malformed body being sent to external REST APIs.

Sample raw XML content:
```
<?xml version="1.0" encoding="UTF-8"?>
<note>
  <email>{{Current User.email}}</email> 
  <code>1234</code>
  <ref>6789</ref>
  <somestring>{{sample_binding}}</somestring>
</note>
```

Sample content as it is sent to an external API:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>\n<root/>
```

**To Reproduce**

1. Create a rest datasource
2. Click 'Add query' and set the action to 'POST'
3. Update the API endpoint, click on the 'Body' tab and select the 'raw (XML)' option
4. Enter some valid sample content and send out a test request.
5. A consuming API will see one empty 'root' node in the body

**Expected behavior**

Entering valid XML markup should produce equally valid markup with any expressions properly rendered.

deanhannigan:
## Description
Allows the binding of information relating to the user in REST datasource requests. At runtime, the binding expressions will be replaced with information derived from the authenticated user initiating the request.

Addresses: https://github.com/Budibase/budibase/issues/3884
Fixes: https://github.com/Budibase/budibase/issues/6366

Available values: 
`_id, email, firstName, lastName, roleId, status, _rev`

Usage:
Add one of the above keys to a 'Current User' binding expression and place them into the REST datasource configuration.
- e.g. `{{Current User.email}}`

OAuthToken:
The oAuth2 token has been exposed with the following mapping: `{{Current User.OAuthToken}}`.  The mapping is also listed as an option when configuring `bearer` token authentication for a REST datasource.

The bindings can be used when configuring:
- Datasource Headers
- Request Bindings
- Request Parameters
- Request Headers
- Request Body

Binding updates:
The binding drawer has been added to expose all available bindings when configuring a rest query. These were previously available, but they are now listed in the binding drawer and have been assigned their own descriptive format.

- Query level bindings -> `{{Binding.myQueryBinding}}`. 
- Datasource static variables -> `{{Datasource.static.apiGroup}}`. Static variables exposed from the REST datasource
- Dynamic variables -> `{{Dynamic.etag}}` -> Dynamic variables built by the query.

**Note:** You do not have to explicitly use these formats when declaring your bindings. This format is automatically applied to make it easier to visualise a binding in the context it is being used. For example, `{{Binding.myQueryBinding}}` is still saved as `myQueryBinding` in the database.

## Screenshots

**Datasource Headers**
query_headers

**Datasource Authentication**
datasource_bearer

**Request Bindings**
bindings

**Request Parameters**
query_params

**Request Headers**
query_headers

**Request Body**
You can mix and match request parameters and user bindings in the body of your requests.
query_body

**Builder Provider Binding**
You can override the elements in your data providers in the builder itself.
builder_mapping


ashwamegh:
@deanhannigan Does this also adds Logged in user contexts in Automations. So alongside `Triggers` data, we can also access` Current User` data?

deanhannigan:
Hey @ashwamegh! This ticket focuses on adding Current User bindings to REST datasource queries. There's more information and further discussion around extending the functionality to automations here: https://github.com/Budibase/budibase/issues/4139

deanhannigan:
@aptkingston Thanks for taking a look! Regarding the bindings, that's correct, they are enriched from a global user. 
Also yes, the `userBindings` are those currently exposed in the builder. My assumption was that this would mean parity between the builder and rest queries, but as you point out, for advanced functionality this could be rather limiting. 
To that end, should a specific set of bindings be created explicitly for the global user? Utilising the same syntax `{{ Current User.[query]}}` but a set of context specific bindings?

Some of the elements on the global user have nested elements like `oauth2.accessToken` or `thirdPartyProfile`. There is currently a suggestion to specifically add {{ OAuthToken }} in place of `oauth2.accessToken` in its own right for https://github.com/Budibase/budibase/issues/3827.

As for the `DrawerBindableInput`, that's a fair point. I hadn't considered it as an option when working on the UI elements. As you point out, that would absolutely remove the guess work for the user and certainly remove some of the parsing back-and-forth.

aptkingston:
> @aptkingston Thanks for taking a look! Regarding the bindings, that's correct, they are enriched from a global user. Also yes, the `userBindings` are those currently exposed in the builder. My assumption was that this would mean parity between the builder and rest queries, but as you point out, for advanced functionality this could be rather limiting. To that end, should a specific set of bindings be created explicitly for the global user? Utilising the same syntax `{{ Current User.[query]}}` but a set of context specific bindings?
> 
> Some of the elements on the global user have nested elements like `oauth2.accessToken` or `thirdPartyProfile`. There is currently a suggestion to specifically add {{ OAuthToken }} in place of `oauth2.accessToken` in its own right for #3827.
> 
> As for the `DrawerBindableInput`, that's a fair point. I hadn't considered it as an option when working on the UI elements. As you point out, that would absolutely remove the guess work for the user and certainly remove some of the parsing back-and-forth.

Just following up on this - is it possible to enrich using the app user rather than the global user? I definitely feel that would enable a lot more use cases since they could actually use custom data relating to the user. Maybe it's not possible with the implementation but just wanted to check, and I wasn't sure if you were doing any more work on this or not.

shogunpurple (approved):
LGTM! This is a really solid piece of work 👌

PR #6335, opened 2022-06-15
Use s3 region from datasource config

### LINKED ISSUE #6332: CORS error in S3 File Upload - Non eu region
### Discussed in https://github.com/Budibase/budibase/discussions/6324

<sup>Originally posted by **czarnndz** June 15, 2022</sup>
Budibase Cloud 1.0.199 

I'm trying to use a S3 File Upload component in Form, but I'm getting CORS error

image

I set S3 Source without Endpoint, also I tried with and without Región but I saw that Request always use eu-west-1, maybe that could be the error idk.
image

This are the config I'm using
image

And I'm using the "Upload File to S3" Action in "On change" event
image

The AWS User I'm using has administrator policy,
Below is the CORS bucket configuration in S3 I'm using

`[
    {
        "AllowedHeaders": [
            "*"
        ],
        "AllowedMethods": [
            "PUT",
            "POST",
            "DELETE",
            "GET"
        ],
        "AllowedOrigins": [
            "https://hic.budibase.app"
        ],
        "ExposeHeaders": []
    },
    {
        "AllowedHeaders": [
            "*"
        ],
        "AllowedMethods": [
            "PUT",
            "POST",
            "DELETE",
            "GET"
        ],
        "AllowedOrigins": [
            "*"
        ],
        "ExposeHeaders": []
    }
]`

Hope that someone can help me with this.

Thanks and Regards
</div>

melohagan:
## Description
CORS issue caused by region of S3 datasource not being passed through.

Addresses: 
- https://github.com/Budibase/budibase/issues/6332


packages/server/src/api/controllers/static/index.js
```diff
@@ -115,6 +114,10 @@ exports.getSignedUploadURL = async function (ctx) {
   // Determine type of datasource and generate signed URL
   let signedUrl
   let publicUrl
+  const AWS_REGION =
+    datasource?.config?.region?.length > 0
+      ? datasource?.config?.region
+      : "eu-west-1"
```
Rory-Powell, line 120: I don't think we should be doing any defaulting, there's no real basis for it. If there is no region in the config we should probably throw an error. Maybe best to check with @aptkingston, was there ever not a region present on the s3 datasource config?

aptkingston: I originally just actually copied the convention from the existing dynamo client, which was the other place that used the AWS SDK. I agree though in principle that if no region is specified then we should just error out instead.

Currently there's no validation forcing a region to be picked, so it's possible to have an undefined region. Even if we add validation now, there's the possibility that existing datasources are out there which are empty and are currently defaulting to `eu-west-1` and working, so it could technically be a breaking change to remove the default...

Maybe best to just keep the fallback for now, but also add validation to prevent empty regions going forward.

As an aside @melohagan - you can shorten this expression to just:
```js
const AWS_REGION = datasource?.config?.region ?? "eu-west-1"
```

melohagan: Nice thanks! Made the change suggested.

Rory-Powell (approved):

PR #4815, opened 2022-03-07
use _id instead of name for identifying tables

mslourens:
## Description
When you have two data sources with a table with the same name, you cannot select one of the two tables when creating a new screen, you'll always select all, because the table is identified by name, not by _id. This PR fixes that.
Fixes #4809

## Screenshots
image

Issue #4809, opened 2022-03-07
Unable to select table when another table with the same name exists


mslourens:
**Describe the bug**
When you have two data connections, each with a table called 'products', both table get selected when adding a new screen

**To Reproduce**
1. Go to the Data tab.
2. Create two data connections, each with a table called 'Products' (or any other name, as long as they are the same)
3. Go to the design tab and click the 'add' button
4. Try selecting one of the 'Products' tables.
5. Notice: Both tables get selected
6. Click the Add Screens button
7. 6 screens are added

**Expected behavior**
A single table should become selected and 3 screens should be created

**Screenshots**
<img width="257" alt="image" src="https://user-images.githubusercontent.com/1907152/156992002-69e7ef97-2046-4643-a3ff-696885f87192.png">

<img width="647" alt="image" src="https://user-images.githubusercontent.com/1907152/156991839-526eed1a-6c88-446b-84ee-49db75bdbd31.png">

**Desktop (please complete the following information):**
 - OS: MacOS 12.2.1
 - Browser: Vivaldi
 - Version: develop branch

PR #6352, opened 2022-06-16
Fix autoscreen and autoscreen link creation and deletion

aptkingston:
## Description
This PR builds on top of @mslourens PR https://github.com/Budibase/budibase/pull/4815 (thanks again Maurits!), fixing additional issues and also fixing the merge conflicts from the PR being stale.

Addresses #4809.

Fixes:
- Use table ID rather than table name when creating autoscreens
- Fix autoscreen table ID not being attached to created autoscreens
- Fix autoscreens not being deleted when deleting underlying table
- Fix links to autoscreens not being deleted when deleting underlying table
- Reduce 409 conflict frequency when working with autoscreens
- Remove deprecated logic around creating and deleting links

Enhancements:
- When any screens are deleted, any navigation links to those screens will automatically also be deleted

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -190,6 +190,7 @@ export const getFrontendStore = () => {
 
         // Build array of promises to speed up bulk deletions
         const promises = []
+        let deleteUrls = []
         screensToDelete.forEach(screen => {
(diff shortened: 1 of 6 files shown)


shogunpurple (approved):
Really cool that all these fixes actually resulted in a real simplification of the logic! LGTM

Issue #5431, opened 2022-04-14
Error exporting data from table component


GiovanniBaldo:
I've budibase on premise in docker. version 1.0.110.
Using new export to csv I always receive error as shown in image. 
I had the same problem on previous version so first thing was to update.
Data comes from a sql server table.
Cant give additional info cause there's nothing more than what seen in the image.
![image](https://user-images.githubusercontent.com/70816346/163464743-d9ca74a5-6573-47ab-bb1f-7a19e1b1d51d.png)

aptkingston:
> @GiovanniBaldo it looks like the error comes from this `try {} catch {}` block:
> 
> https://github.com/Budibase/budibase/blob/9423347278daf1cde47b3d3cef49920640ba797d/packages/client/src/utils/buttonActions.js#L255-L264
> 
> For some reason it doesn't log the error, if you don't mind building from source you could add something like `console.error(error)` after the `notificationStore.actions.error` then you could see the error in the console.

You could also check your browser dev tools, because if this error is being generated by the server then you'll see the real error in the API response (and also in the browser console as we have middleware that logs errors to the console).

bolshoytoster:
@aptkingston
> we have middleware that logs errors to the console

That's probably easier.

GiovanniBaldo:
thanks for your answer. I'm not smart enough to build from source. for what I can see from chrome developer console the error is "Cannot convert undefined or null to object" . 
I removed the columns that may contain null values but nothing change.
![image](https://user-images.githubusercontent.com/70816346/164471253-20b0bd9c-97f2-4714-a773-58aa8664183f.png)

melohagan:
Related discussion: https://github.com/Budibase/budibase/discussions/5982

PR #6354, opened 2022-06-16
Bug fix for exporting selected rows from External Datasource

melohagan:
## Description
The filter was incorrect. Updated to grab all matching selected rows for external datasources.
Because of the encoding, the id has to be decoded to work with the query as an Integer is expected.

Addresses: 
- https://github.com/Budibase/budibase/issues/5431


aptkingston (approved):
LGTM! Nice one Mel. Let's get this merged in as there are some other issues that I can't look at until this fix is in.

PR #6361, opened 2022-06-16
Fix for single image build

mike12345567:
## Description
This contains a fix for the single image build, which pins the versions of Budibase packages so that development versions can build correctly and packages can be properly deduped.

There is still some underlying issues that need to be discussed with this, but this is a solution for now.

Also this does some optimisation to reduce the image size down from 4GB to 1.8GB - still large, but a lot better.

This is required by @jonnymccullagh for his work, as the image he was building locally did not work, this should solve the problem.


packages/backend-core/package.json
```diff
@@ -56,7 +57,6 @@
     ]
   },
   "devDependencies": {
-    "@budibase/types": "^1.0.200-alpha.3",
```
Rory-Powell, line 59: How come the move from dev to real dependency?

Rory-Powell: Similar for worker and server, where we already have them in dev and now in both?

scripts/pinVersions.js
```diff
@@ -3,35 +3,44 @@ const path = require("path")
...
+}
 
 function pinDeps(dependencies) {
-	for (let dependency in dependencies) {
-		if (dependency.startsWith("@budibase")) {
-			dependencies[dependency] = dependencies[dependency].replace("^", "")
-		}
-	}
+  for (let dependency in dependencies) {
+    if (dependency.startsWith("@budibase")) {
```
Rory-Powell, line 18: I wonder if we pinned the bb dependencies in our package file, would lerna respect that when it does a release and also use a pinned version. If so it could do away with the need for pinning these altogether

packages/backend-core/package.json
mike12345567: it was failing to run when retrieving dependencies with `NODE_ENV=production yarn` - this appears to be an actual dependency of the running code, I didn't look into it, just moved it as it was required to run.

scripts/pinVersions.js
mike12345567: I wondered that, I think pinning as part of the build is a good start here, ideally I think it'd be good if we just didn't need to do this stuff, as per how we discussed, building externally and then passing the build + Budibase dependencies into the image and only fetching the non-Budibase dependencies, that way pinning isn't an issue anymore.

packages/backend-core/package.json
Rory-Powell: It shouldn't be, it's only ever used as types. Given that it doesn't fail in the develop / release builds, could it be something that needs changed on the single image build rather than in the source?

Rory-Powell: It could have been related to this being based off develop too, maybe an older version that didn't exist or something. Could try again with these moved back into dev after the merge

mike12345567: Yeah, I'll give it another go, it was something to do with Analytics, I thought it shouldn't be needed, but wondered if there was some enumeration that was used in release.

Rory-Powell (approved):

Issue #6175, opened 2022-06-02
CSV Export


PaulMcF1987:
I am planning to open the app I have created up to my customers.

Budibase works relatively well as an Inhouse app.

But for opening the app up to clients I feel there could be some improvements made and this is one of the areas...

I am looking to set Budibase up, so that a customer can view only their jobs. I have customers who ask for exports of their jobs every month and the ability to let them export their own at their leisure would be massive.

However... If I use the standard table that I save all of the jobs to... If the customer choses to export, although the table they are viewing only shows them the information that I want them to see, the CSV exports all of the information form that table which gives them information what should not be shared with them...

If I set it up so that when they log in, the data is shown from a different data source, It then does not allow Row selection.

------------------------------------------------------------------------------------------------------------------------------------------------

Further to the above, I feel the following would be massively beneficial
Currently we will only be able to export the number of records that are showing on the table.
If for example the table is set to display 20 records and someone needs to export 100 records then this would mean doing 5 separate exports and then copying them all onto one document.

In another low code environment that I use, when I chose to export data, it gives me options (similar to the filter function on Budibase) so that I can chose to export depending on column data, this can be set within dates ranges either from dates in a certain table column or the date created/edited etc...

So the ability in the back end to configure the columns which are eligible for export and then in the front end, the ability to further filter the export to select all columns which match the critera (across all rows/pages and not just the rows currently displayed on the table)

PaulMcF1987:
@shogunpurple 

> you copied and pasted https://github.com/Budibase/budibase/issues/6173#issuecomment-1145204969 into this issue too, and it's been mentioned in other issues - so was just making sure there was less duplication.

Sorry, yeah, I thought the issue from the other post had been sorted so I closed it and opened this feature request and added a bit from that. However, I then realised that the issue was not sorted and I am still not able to export to CSV (I have not tried JSON)

I also noticed yesterday when the export did work (For around 10 mins or so) that if you change page you can select more and it will add to selections from other pages which definitely helps.

> Sounds interesting, which platform is this?

The other platform I used was called Corteza. It is more CRM based, Salesforce-esque but it also has an area to create business apps.

> In terms of full filtering functionality and having much more control over this, I can understand why that would be useful.

In terms of filtering... The export issue I have is that, the SQL table is used for my company to note all details of a job. 
So in that table will be details, for example... `how much I bought a product for and then how much I sold it to my customer and the profit made`
Some of my customers want to be able to log in and keep tabs on the jobs done.
[...]

PR #6392, opened 2022-06-20
Add option to customise which columns are exported in export data action

aptkingston:
## Description
This PR adds an option to the `Export data` action to allow choosing which columns should be exported. Leaving this setting blank defaults to exporting all columns.

Selecting specific columns:
image
image

Defaut to all columms:
image
image

Addresses: 
- #6175


shogunpurple (approved):
LGTM

Issue #6398, opened 2022-06-20
SQL Server - Filter More Than / Less Than on int causes validation failure


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.206
  - App Version: 1.0.206

**Describe the bug**
More than / less than filter broken in data provider when using SQL Server source on int type column.
Discussed here: https://github.com/Budibase/budibase/discussions/6375

**To Reproduce**
1. Create SQL Server table with an `int` type field
2. Add a row with the number 10
3. Use this as a data provider -> add filter
4. Equals works, but more than or less than a simple number causes validation error

**Expected behavior**
I should be able to filter more than / less than on integers for SQL Server connector

**Screenshots**
![Screenshot 2022-06-20 at 15 22 23](https://user-images.githubusercontent.com/101575380/174622853-026b5f87-0268-4efd-8268-e01450f3a773.png)

PR #6402, opened 2022-06-20
External data source plus - Integer filter range validation

melohagan:
## Description
All of the external SQL data sources were throwing validation errors when filtering by integer range, because the JavaScript max/min values exceeded for example Int32, and small int.

Now passing through the external data types to the filter so that the range limits can be determined from a map if needed.

Addresses: 
- https://github.com/Budibase/budibase/issues/6398

### DIFF
--- a/packages/frontend-core/src/constants.js
+++ b/packages/frontend-core/src/constants.js
@@ -67,7 +67,7 @@ export const ApiVersion = "1"
 /**
  * Maximum minimum range for SQL number values
  */
-export const SQL_NUMBER_TYPE_RANGE_MAP = {
+export const SqlNumberTypeRangeMap = {
   integer: {
     max: 2147483647,
     min: -2147483648,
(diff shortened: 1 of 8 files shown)


packages/frontend-core/src/constants.js
```diff
@@ -63,3 +63,25 @@ export const TableNames = {
  *   - Coerce types for search endpoint
  */
 export const ApiVersion = "1"
+
+/**
+ * Maximum minimum range for SQL number values
+ */
+export const SQL_NUMBER_TYPE_RANGE_MAP = {
```
aptkingston, line 70: NAB: could we use camel case here rather than block caps to stick with the convention?

aptkingston (approved):
LGTM! Nice work @melohagan.

Obviously this will just work for newly created columns / newly imported tables - so we just need to keep that in mind when communicating the fix to users. Reimporting their SQL tables will be the quick fix to fill in the external type metadata and allow this fix to work.

Issue #6135, opened 2022-05-31
Screen Actions "Before Load" / "On Load" / "Before Close"


poirazis:
The user should be able to define Actions on Screen events

Issue #5019, opened 2022-03-22
default "on page load" action


dubbya:
There is a need to fire an action or actions, without the user clicking a button and as the page loads. E.g. run a query to compare local db data to a rest endpoint.

PR #6409, opened 2022-06-21
"On screen load" actions & query params binding

aptkingston:
## Description
Couple of useful features which have been requested before and were raised again recently.
The upstream branch is the new design UI as these features would otherwise conflict a bit too much.

### "On load" actions for screens
This is a new screen level setting where you can configure actions that will be executed as soon as the screen mounts.
These actions will not fire when working in the builder preview to avoid doing things like saving empty rows.

image

### Binding for query params
A new data binding has been added for query params. As we don't have the "schema" for query params, this binding is just an object, so you need to manually edit the binding to extract the relevant param.

image
image

Both features mentioned in https://github.com/Budibase/budibase/discussions/6388#discussioncomment-2990638.
"On load" requested in #6135, #5019

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -694,6 +694,13 @@ export const getAllStateVariables = () => {
     })
   })
 
+  // Add on load settings from screens
+  get(store).screens.forEach(screen => {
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM, all of these features are big QoL improvements 💪

PR #6415, opened 2022-06-21
ARM support - single image

mike12345567:
## Description
Single image ARM support (Oracle instant client) and adding ARM commands. This uses `docker buildx` to manage the multi-arch build, there is a new command at the top level `yarn build:docker:single:multiarch` which will build both an amd64 image and an arm64 image as well.

To do this on Linux you'll need to install QEMU as a multi-arch emulator, this can be done with the packages `qemu` and `qemu-user-static`. I've add a script which will setup your environment to install these packages (for aptitude based distros) and create a docker-container builder which can be used for multi platform building.

This will need to be added to the Github actions pipeline, we can likely use [this action](https://github.com/docker/setup-buildx-action#with-qemu) to get it setup in our environment, if we run this before the above yarn command then it should just work, we will then have both arch images available to upload.

### DIFF
--- a/hosting/single/Dockerfile
+++ b/hosting/single/Dockerfile
@@ -41,7 +41,6 @@ ENV DEPLOYMENT_ENVIRONMENT=docker \
   SELF_HOSTED=1 \
   CLUSTER_PORT=10000 \
   REDIS_PASSWORD=budibase \
-  ARCHITECTURE=amd \
   APP_PORT=4001 \
   WORKER_PORT=4002
 
@@ -89,7 +88,7 @@ ADD hosting/single/vm.args ./etc/
 
 # setup minio
 WORKDIR /minio
-RUN wget https://dl.min.io/server/minio/release/linux-${ARCHITECTURE}64/minio && chmod +x minio
+RUN wget https://dl.min.io/server/minio/release/linux-${TARGETARCH}/minio && chmod +x minio
 
 # setup runner file
 WORKDIR /
--- a/hosting/single/Dockerfile
+++ b/hosting/single/Dockerfile
@@ -20,6 +20,8 @@ RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
 
 FROM couchdb:3.2.1
 
+ARG TARGETARCH
+
 COPY --from=build /app /app
 COPY --from=build /worker /worker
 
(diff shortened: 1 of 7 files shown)


shogunpurple (approved):
LGTM! Amazing to see bb supporting ARM!

aptkingston:
This is awesome! Many self hosters will be very happy :+1:

Rory-Powell (approved):

hosting/single/Dockerfile
```diff
@@ -20,6 +20,8 @@ RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
 
 FROM couchdb:3.2.1
 
+ARG TARGETARCH
```
jonnymccullagh, line 23: Can we default this to to x86-64 ? And it can be over-ridden for arm64. Also maybe add a comment as to the available architectures.

mike12345567: By default this will be amd64, this is an argument that is passed in my `buildx` defining which build it is performing, I don't think we have control over it really.

For Oracle etc we have an install script which works out based on the `TARGETARCH` (it just checks if it starts with ARM, then uses the ARM binaries, otherwise it uses the x86/64 bit binaries) for MinIO we could have a similar script which does this - you are right that using the arch directly could be a little unsafe.

jonnymccullagh (approved):

package.json
```diff
@@ -62,6 +62,7 @@
     "build:docker:develop": "node scripts/pinVersions && lerna run build:docker && npm run build:docker:proxy:compose && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh develop && cd -",
     "build:docker:airgap": "node hosting/scripts/airgapped/airgappedDockerBuild",
     "build:digitalocean": "cd hosting/digitalocean && ./build.sh && cd -",
+    "build:docker:single:multiarch": "docker buildx build --platform linux/arm64,linux/amd64 -f hosting/single/Dockerfile -t budibase:latest .",
     "build:docker:single:image": "docker build -f hosting/single/Dockerfile -t budibase:latest .",
```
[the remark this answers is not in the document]
mike12345567: Just seeing this - by specifying the `--platform` for buildx, this automatically sets the `TARGETARCH` arg, its one of the options it passes per platform.

Issue #6421, opened 2022-06-22
Filtering Budibase DB columns with spaces


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6400

<div type='discussions-op-text'>

<sup>Originally posted by **theetrain** June 20, 2022</sup>
In my case, I want to filter data using a static value, but the _column_ contains spaces. From what I can gather, here are the cases:

* When a column name contains 1 or 0 spaces, it is filterable in data providers and the 'Data' screen
* When a column name contains 2 or more spaces, it is NOT filterable

See the below screenshot that illustrates my findings. In this test page, every table is wrapped with individual data providers, and the headings describe what filters are applied.

![Screenshot of test page illustrating how some data providers containing columns spaces may or may not be filterable](https://user-images.githubusercontent.com/12798751/174852863-2c8d3cf6-a1e6-4878-8d78-ccfa7b140459.png)</div>

aptkingston:
I'm pretty sure I fixed this exact issue a few months ago in this commit: https://github.com/Budibase/budibase/commit/d6e568dbe3364603cf6a4a2805c861185cfca726. It involved updating the index function though and I'm not sure when/if existing DB instances use this updated search function or not. I have a feeling it needs to be manually triggered... @mike12345567 could shed some light on this.

mike12345567:
@melohagan - did you create a new app in the latest version of Budibase to confirm this? The problem should only persist with older apps, a migration wasn't built for the index as it would require a very large amount of DB work to update all indexes for a minor change, all new apps however should have the problem resolved.

For old apps exporting and re-importing the app will trigger re-creation of the index. It may be worth looking at a migration if this keeps coming up, but all apps made since Andrew's [commit](https://github.com/Budibase/budibase/commit/d6e568dbe3364603cf6a4a2805c861185cfca726) should just work fine.

aptkingston:
Ah I've found the issue - it's another string `replace` regex that doesn't use the global flag, so only replaces one space.
https://github.com/Budibase/budibase/blob/cb153c7da6ce3f85989f02a2e5817a7fda6f7eed/packages/server/src/api/controllers/row/internalSearch.js#L148

fredhp:
Guys, it still happens with special characters á, ó, ã, etc.

PR #6426, opened 2022-06-22
Replace all column spaces with underscores when performing internal searches

aptkingston:
Fixes an issue where searching would not work when using a column containing more than one space. We fixed this issue recently but either overlooked this one usage or introduced a regression at some point.

Addresses https://github.com/Budibase/budibase/issues/6421.


Rory-Powell (approved):
👍

Issue #5708, opened 2022-05-04
'money' Data Type Incompatible with Form Number Field


zimmra:
**Hosting**

- Self
  - Method: Docker Compose 
  - Budibase Version: 1.0.137 
  - App Version: 1.0.137 
  - Database: Postgresql

**Describe the bug**
When using a `money` data type, it is properly detected as a number column in Budibase and displays it in tables properly, but when attempting to use it in a Form/Number Field the existing value is not displayed

**To Reproduce**
1. Have a PGSQL table with two columns, a `numeric` column and a `money` column, populate row(s) with values
2. Define/fetch this table into Budibase
3. In designer, create a screen and populate a Form and Field Group (and needed components)
4. Preview app, see on sceen that Number Field associated with the `numeric` column populates fine, where the Number Field associated with the `money` column does not. 

**Expected behavior**
For the Number Field to display the contents of the associated row from the column that is the `money` data type

**Screenshots**
---
Data in PostgreSQL:
![image](https://user-images.githubusercontent.com/28514085/166717832-14290cee-6c8f-4e11-b892-486e1206c0ba.png)
---
Data in Budibase table:
![image](https://user-images.githubusercontent.com/28514085/166717993-ae5030f9-7b83-48e8-a8ab-6c7752b27b86.png)
---
Data in Budibase Number Field:
![image](https://user-images.githubusercontent.com/28514085/166717384-32dfd7b7-3d98-4e2e-aa47-24e8c336e6f9.png)
![image](https://user-images.githubusercontent.com/28514085/166718481-7650553c-dc07-43be-bd0d-6f40b74cdb19.png)
![image](https://user-images.githubusercontent.com/28514085/166718508-210cc0b0-cec2-416d-97c8-4932da852740.png)
---

PR #6432, opened 2022-06-22
Cast Money as numeric for Postgres

melohagan:
## Description
Money types in MS SQL Server come back as numeric unless otherwise specified.
In Postgres it's the opposite, so you have tell Postgres to return money types as numeric.

Addresses: 
- https://github.com/Budibase/budibase/issues/5708

### DIFF
--- a/packages/server/src/integrations/base/sql.ts
+++ b/packages/server/src/integrations/base/sql.ts
@@ -93,13 +93,14 @@ function generateSelectStatement(json: QueryJson, knex: Knex): any[] {
   const { resource, meta } = json
   const schema = meta?.table?.schema
   return resource.fields.map(field => {
-    const shortFieldName = field.match(/(?<=\.).*$/g)?.[0]
-    if (shortFieldName && knex.client.config.client === "pg") {
-      const externalType = schema?.[shortFieldName].externalType
+    const fieldNames = field.split(/\./g)
+    const tableName = fieldNames[0]
+    const columnName = fieldNames[1]
+    if (columnName && knex.client.config.client === SqlClients.POSTGRES) {
+      const externalType = schema?.[columnName].externalType
       if (externalType?.includes("money")) {
-        const fieldName = field.split(/\./g)
         return knex.raw(
-          `"${fieldName?.[0]}"."${fieldName?.[1]}"::money::numeric as "${field}"`
+          `"${tableName}"."${columnName}"::money::numeric as "${field}"`
         )
       }
     }
--- a/packages/server/src/integrations/base/sql.ts
+++ b/packages/server/src/integrations/base/sql.ts
@@ -89,6 +89,24 @@ function parseFilters(filters: SearchFilters | undefined): SearchFilters {
   return filters
 }
 
+function generateSelectStatement(json: QueryJson, knex: Knex): any[] {
+  const { resource, meta } = json
+  const schema = meta?.table?.schema
+  return resource.fields.map(field => {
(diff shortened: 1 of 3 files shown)


packages/server/src/integrations/base/sql.ts
```diff
@@ -89,6 +89,24 @@ function parseFilters(filters: SearchFilters | undefined): SearchFilters {
   return filters
 }
 
+function generateSelectStatement(json: QueryJson, knex: Knex): any[] {
```
mike12345567, line 92: Can we be more specific than `any` for the type here? I believe this is either a string or a knex raw statement, but this should have a type and it would make it more obvious what this is doing.

I think knex raw may actually return a string, if this is a casting issue/typescript being unaware, you can do something like `return knex.raw(`....`) as string`

packages/server/src/integrations/base/sql.ts
```diff
@@ -89,6 +89,24 @@ function parseFilters(filters: SearchFilters | undefined): SearchFilters {
...
+  const { resource, meta } = json
+  const schema = meta?.table?.schema
+  return resource.fields.map(field => {
+    const shortFieldName = field.match(/(?<=\.).*$/g)?.[0]
```
mike12345567, line 96: Would be good to move this regex to a global constant at the top of the file, something like 
```
const PURPOSE_REGEX = /(?<=\.).*$/g
```
I'll be honest I'm not quite sure what the purpose of the regex is here and this makes it obvious for anyone coming and looking at it, since regex is always shrouded in mystery!

mike12345567 (changes requested):
Just a few minor comments which might help to make the change more obvious in future - probably worth implementing them for our sakes down the line, otherwise LGTM!

packages/server/src/integrations/base/sql.ts
melohagan: Knex.raw returns a 'Knew.Raw' type. I tried casting as a string / using `.toSQL().sql` but that didn't work.

I realised afterwards anyway that when you pass a string into the `select` method, it does the client specific column wrapping, e.g. for Postgres it wraps the column name in double quotes.

This was a problem because it would return `"TableName.ColumnName::money::numeric"` 
You need the raw type to tell knex not to do any additional quote wrapping to the string.

packages/server/src/integrations/base/sql.ts
melohagan: I was able to remove that regex statement altogether!
The split on the '.' was enough to get both the tableName and columnName.

packages/server/src/integrations/base/sql.ts
```diff
@@ -89,6 +89,24 @@ function parseFilters(filters: SearchFilters | undefined): SearchFilters {
...
+    if (shortFieldName && knex.client.config.client === "pg") {
```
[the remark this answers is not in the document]
melohagan: Using the enum

packages/server/src/integrations/base/sql.ts
mike12345567: Ah, in that case we can just `(string | Knex.Raw)[]` to define an array of mixed types - its helpful in the SQL layer to try to apply types where possible to these statements, there's quite a bit of complexity and it'll help us catch any weirdness at build time (since Knex is already pretty well typed).

melohagan: I didn't know you could do that - cool! 

That's done. In the process, I also removed a couple of unused parameters in `addRelationships` just so you're aware.

mike12345567 (approved):
LGTM! Nice fix.

Issue #5962, opened 2022-05-18
@ symbol in environment password var causes docker app-service to fail


melohagan:
**Hosting**
- Self
  - Method: docker compose on VPS
  - Budibase Version: v1.0.150
  - App Version: v1.0.150

**Describe the bug**
Using an '@' symbol for the `COUCH_DB_PASSWORD` env var for the **app-service** causes an invalid URL error.
See discussion: https://github.com/Budibase/budibase/discussions/5939#discussioncomment-2774829

**Expected behavior**
I should be able to use special characters in my passwords without causing the **app-service** to fail.

PR #6440, opened 2022-06-23
Improve Couch DB URL parsing to handle edge cases and special characters

aptkingston:
## Description
This PR improves the parsing logic for Couch DB URLs, which are passed in as an env variable. The parsing now supports multiple different URL formats and edge cases, and also supports using special characters such as `:/@` in the password.

Addresses #5962.
Mentioned in #5939.


Rory-Powell (approved):
Great job, really love the tests which made reviewing much easier

Issue #6268, opened 2022-06-10
Add Relational/Non-Relational labels to datasources


shogunpurple:
**COS**
- Add labels under each datasource to indicate the database type

**Designs**
![image.png](https://images.zenhubusercontent.com/5fd773ddf3d67d4ceebe466b/73d19f26-008b-4b65-9222-91c93ac2eedd
)

PR #6443, opened 2022-06-23
Datasource labels

aptkingston:
## Description
This PR adds a type label to each data source, describing what type it is. I've also updated the "add data source" modal in general to better match designs.

image

Addresses #6268.


Rory-Powell (approved):
Nice one

PR #6447, opened 2022-06-23
fixes for single image build

jonnymccullagh:
## Description
Fixes for the single docker image build with Lets Encrypt support including:
- reloading nginx may not be sufficient to pick up the SSL conf so using a restart
- fix quotes in sed command to ensure CUSTOM_DOMAIN variable is expanded
- add nginx config test to the healthcheck
- comment out the CUSTOM_DOMAIN variable by default
- move port 80 config into the ./sites-enabled directory to be a little more what people might expect


jonnymccullagh:
This version has also been tested on AlmaLinux 8 (Red Hat/CentOS/Rocky family) and has also been tested using podman in place of docker.

jonnymccullagh:
Additionally I have now amended the single image build to have a build target of "aas" for Azure App Service. This build type prepares the image with the Azure App Service requirements, specifically:
- persisted data needs to be under /home so config paths for minio and couch amended
- SSH needs to be installed and listening on port 2222

hosting/single/default
```diff
@@ -0,0 +1,94 @@
+server {
```
mike12345567, line 1: could we name this file something with nginx in it, just make it clear its an nginx configuration file? like `nginx-default.conf` and then just name it as required when being passed into the build? Just thinking the directory is full of a few different config files and its probably good to make their purpose obvious without the need to look through them.

mike12345567 (approved):
LGTM! Very minor comment just about the file `hosting/single/default` - filename doesn't really make it obvious what itdoes!

shogunpurple (approved):
LGTM, amazing to get bb running on so many platforms - great docs too.

PR #6448, opened 2022-06-23
Fix issue with loop item in javascript bindings

PClmnt:
## Description
Fixes an issue where using the `loop.currentItem` binding in javascript would result in undefined. 

n.b not an issue but used this PR to slightly update how we handled the automation max iterations
Addresses: 
- #6387

## Screenshots
As you can see in the video, the loop now successfully runs and logs the results

https://user-images.githubusercontent.com/5665926/175335981-799e821e-e359-4d8b-8adc-a1d77bc36fea.mp4


shogunpurple (approved):
LGTM! Nice fix

PR #6452, opened 2022-06-23
Redis update for pro usage

mike12345567:
## Description
This addresses some of the problems with the `usage_quota` doc, where the write rate is so large for a single document that a hot shard is created and causes snowball CPU usage.

The `writethrough.ts` file provides wrappers for the `put` and `get` functionality provided by PouchDB, essentially you pass a database to it and then attempt to put or get items and it will use the cache as needed, based on the write rate that is specified.

This acts as a "debounce", reducing the rate at which a document/key is updated to that of once per specified update period. It will also handle when no document is present and will read from redis when it is available.

I've added some basic test cases to prove it out using the `timekeeper` package to control the value of `Date.now()` and confirm it works as expected.

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -63,6 +63,7 @@
     "@types/koa": "2.0.52",
     "@types/node": "14.18.20",
     "@types/node-fetch": "2.6.1",
+    "@types/pouchdb": "^6.4.0",
     "@types/redlock": "4.0.3",
     "@types/semver": "7.3.7",
     "@types/tar-fs": "2.0.1",
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -63,7 +63,7 @@
     "@types/koa": "2.0.52",
     "@types/node": "14.18.20",
     "@types/node-fetch": "2.6.1",
-    "@types/pouchdb": "^6.4.0",
+    "@types/pouchdb": "6.4.0",
     "@types/redlock": "4.0.3",
     "@types/semver": "7.3.7",
     "@types/tar-fs": "2.0.1",
--- a/packages/backend-core/src/environment.ts
+++ b/packages/backend-core/src/environment.ts
@@ -16,7 +16,7 @@ if (!LOADED && isDev() && !isTest()) {
   LOADED = true
 }
 
-const env: any = {
+const env = {
   isTest,
   isDev,
   JWT_SECRET: process.env.JWT_SECRET,
@@ -66,6 +66,7 @@ const env: any = {
 for (let [key, value] of Object.entries(env)) {
   // handle the edge case of "0" to disable an environment variable
   if (value === "0") {
+    // @ts-ignore
     env[key] = 0
   }
 }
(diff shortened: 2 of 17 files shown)


packages/backend-core/src/environment.ts
```diff
@@ -66,6 +66,7 @@ const env: any = {
 for (let [key, value] of Object.entries(env)) {
   // handle the edge case of "0" to disable an environment variable
   if (value === "0") {
+    // @ts-ignore
     env[key] = 0
   }
```
Rory-Powell, line 71: Nice one, this is a better fix that the any I had put in which was a real bother

packages/backend-core/yarn.lock
```diff
@@ -291,6 +291,11 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@budibase/types@^1.0.209":
```
Rory-Powell, line 294: This can probably be removed with a yarn clean and bootstrap

packages/backend-core/package.json
```diff
@@ -63,6 +63,7 @@
     "@types/koa": "2.0.52",
     "@types/node": "14.18.20",
     "@types/node-fetch": "2.6.1",
+    "@types/pouchdb": "^6.4.0",
```
Rory-Powell, line 66: Can we pin this dependency?

Rory-Powell (commented):
The .209 version, that might just be a merge conflict?

shogunpurple (approved):
LGTM! This is a really useful one for performance and super reusable too - definitely some other places we could leverage this.

Nice bit of future proofing with the numbered/selectable DBs too.

mike12345567:
> LGTM! This is a really useful one for performance and super reusable too - definitely some other places we could leverage this.
> 
> Nice bit of future proofing with the numbered/selectable DBs too.

Thanks this should help us with improving performance of stuff in future!

Just for reference (i've added this as comments to the Redis utils file where these are defined) heres the difference between a Redis `Database` and a `SelectableDatabase` in our system:

`Database`
> These Redis databases help us to segment up a Redis keyspace by prepending the specified database name onto the cache key. This means that a single real Redis database can be split up a bit; allowing us to use scans on small databases to find some particular keys within. 
> If writing a very large volume of keys is expected (say 10K+) then it is better to keep these out of the default keyspace and use a separate one - the SelectableDatabases can be used for this.

`SelectableDatabase`
> These define the numeric Redis databases that can be access with the SELECT command - (https://redis.io/commands/select/). By default a Redis server/cluster will have 16 selectable databases, increasing this count increases the amount of CPU/memory required to run the server. 
> Ideally new Redis keyspaces should be used sparingly, only when absolutely necessary for performance to be maintained. Generally a keyspace can grow to be very large is scans are not needed or desired, but if you need to walk through all values in a database periodically then a separate selectable keyspace should be used.

Rory-Powell (approved):
Ignore that previous .209 version comment, apparently the GitHub app shows different diffs than GitHub web 🤷‍♂️

Issue #3863, opened 2021-12-22
Add "IN" and "NOT IN" conditions in data filter


micebrain:
**Describe the feature request**
Current filter does not allow the "in" condition and for some queries it is difficult to emulate with the existing available conditions.
e.g.  in SQL a statement like `select something from table where type in ('warning', 'error')`

**Screenshots**

<img width="514" alt="Screenshot 2021-12-22 at 21 10 11" src="https://user-images.githubusercontent.com/7136209/147150406-1c8a3e53-78d9-45a5-9aed-d3d49b8e708e.png">

melohagan:
This would be pretty handy. Same goes for Automation Conditions.

PR #6455, opened 2022-06-24
Add "IN" conditions in data filter

melohagan:
## Description
You can now filter to check if a value is contained within a provided array.
*The "NOT IN" case is not done as part of this PR*

Addresses: 
- https://github.com/Budibase/budibase/issues/3863

## Screenshots
Screenshot 2022-06-24 at 14 45 57

**Result**
Screenshot 2022-06-24 at 14 48 12


aptkingston:
What's the support like for this Mel? I didn't realise `oneOf` and `notOneOf` were valid lucene operators - pretty cool! I can see there's a client side implementation so I'm assuming it works for custom queries, but have you also confirmed it works for internal tables and SQL tables?

melohagan:
Ah @aptkingston I've misunderstood how this works. I thought all filters went through this client side code, but it makes sense that it would do the filtering server side for SQL etc!

With that said, by sheer fluke, *oneOf* does work for SQL 😂 

**Postgres**
Screenshot 2022-06-24 at 16 39 46
Screenshot 2022-06-24 at 16 40 59
**JS**: `return [ "Apple", "Banana", "Carrot" ]`
 
**MySQL**
Screenshot 2022-06-24 at 16 44 22
Screenshot 2022-06-24 at 16 45 23
**JS**: `return [ "11.3453", "234", "line1\nline2\nline3" ]`

----
**notOneOf** is not valid at all, so I could remove it for the time being as I don't need it. Looking to use this for handling relationships from Airtable queries - but it's been requested as well of course. 

**oneOf** is also not working for Internal Tables, but looking at `internalSearch.js` hopefully I can figure out an implementation there.

melohagan:
@aptkingston I've added support for Internal tables. From what I could find online, the only way to do an "IN" with Lucene queries is to concatenate OR statements. Apparently there's a cap of 1024 ORs by default - it'll be up to the user to use this within reason. I can slip that somewhere into the docs.

I also removed the **Is not in** filter option, but the **Is in** should now work for internal searches, external searches and custom queries.

packages/frontend-core/src/utils/lucene.js
```diff
@@ -211,6 +212,11 @@ export const runLuceneQuery = (docs, query) => {
     return docValue == null || docValue === ""
   })
 
+  // Process an includes match (fails if the value is not included)
+  const oneOf = match("oneOf", (docValue, testValue) => {
+    return !testValue?.includes(docValue)
```
aptkingston, line 217: Looks great @melohagan! Could we copy the logic here to support splitting strings into arrays if the value is a string? The same way I see that it gets handled server side. Would be nice to keep support consistent, otherwise comma separated strings will work with internal/SQL but not with custom queries.

aptkingston (approved):
LGTM!

PR #6471, opened 2022-06-27
Infra/deliver new UI

shogunpurple:
## Description
This PR allows beta users to trial and use the new design UI. When their login email matches one in the beta program, they will be presented with a banner, like below:

Screenshot 2022-06-27 at 12 48 11

This will: 
- Set a cookie to say they are using the new experience
- pull the code from `new-design-ui` (bundled in S3) and add it to their filesystem, which then serves the new UI when they access the builder

Some changes have also been made on the `new-design-ui` branch, such as adding UI for opting out of the beta program and returning to the original experience.


aptkingston (approved):
LGTM! Nice one Marty. Can't wait to get the new design UI out there and get feedback.

Have we disabled all caching for these new files served from the CDN? Just thinking that we might we have issues with updates to the new design UI bundle otherwise.

Only other thing that comes to mind is that we don't mention anywhere how apps created in the new design UI aren't backwards compatible. We don't want anyone to edit their production apps with the new UI then revert to the old one again. There's no nice way to do it on a per app basis since the server needs to serve the builder bundle before it's in an app context, so maybe we just need to at least say somewhere about that.

shogunpurple:
@aptkingston same here - will be great to get it out there!

We haven't actually - I've just been doing a manual invalidation when the bundle gets pushed. I'll something into the job there, to do it as part of the upload to s3.

Another good point - I'll add that into the banner, and say that it's not recommended for existing apps.

Issue #6433, opened 2022-06-22
Format column date in a table does not directly work


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6423

<div type='discussions-op-text'>

<sup>Originally posted by **NicolasFabre81** June 22, 2022</sup>
My timestampz "Date début" is 2022-06-21 00:00:00.000 +0200

If I add a tag in my table with the syntax {{ date tbl_ruptures.RUPTURES.Date début "YYYY-MM-DD" }}  I get the correct answer : "2022-06-21" 👍 

Now in my table if I put {{ date Value "YYYY-MM-DD" }} I get "June 21 2022,00:00"  👎 </div>

You can use formulas to work around this, but it's still a bug.

PR #6473, opened 2022-06-27
Fix dates with custom templates

aptkingston:
## Description
This PR updates table cell renderer to use a string renderer for date values that use a custom expression. This means that if you have a date column and use a custom JS or HBS expression to alter the date, you'll now actually see the custom value, rather than having the value being parsed again by the date cell renderer. Also fixes a small issue with layout in the column editor component.

Here's a date field with a custom value, correctly being rendered in a table. Before this change, this would have displayed as `Invalid date` as this string would have been attempted to be parsed as a date.
image

Addresses #6433.


shogunpurple (approved):
Small but important fix - LGTM!

Issue #6485, opened 2022-06-28
Changing from JSON field type to Multi-select breaks table


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.206
  - App Version: 1.0.206

**Describe the bug**
When you add a JSON column and edit to a Multi-select column it breaks your Budibase internal table

**To Reproduce**
1. Add table with JSON column.
2. Use the schema `{ "tags": [ { "label": "string" } ] }`
3. Add a row with some valid tags
4. Edit the JSON columns and select Multi-select type. Save (spinning wheel)
5. Refresh the page and witness error from screenshot. 

**Expected behavior**
I should be able to change between JSON columns and Multi-select column types without my table breaking. It forces you to recreate the table which is nasty. 

**Screenshots**
![Screenshot 2022-06-28 at 12 09 55](https://user-images.githubusercontent.com/101575380/176164828-4a70c31c-dba2-4ceb-a6da-745fb6cac526.png)

aptkingston:
We shouldn't allow this at all - it makes no sense to convert a column from JSON to a multi-select field. If you create a column as a JSON field I don't think you should be able to change type to anything else.

PR #6487, opened 2022-06-28
Json to multi-select field type crash

melohagan:
## Description
When changing a field from JSON to multi-select when the row value is type `object` the app would become unusable.
Type check added to handle this gracefully.

Addresses: 
- https://github.com/Budibase/budibase/issues/6485

### DIFF
--- a/packages/bbui/src/Table/ArrayRenderer.svelte
+++ b/packages/bbui/src/Table/ArrayRenderer.svelte
@@ -5,8 +5,12 @@
 
   const displayLimit = 5
 
-  $: badges = value?.slice(0, displayLimit) ?? []
+  $: badges = isValidType() ? value?.slice(0, displayLimit) : []
   $: leftover = (value?.length ?? 0) - badges.length
+
+  function isValidType() {
+    return Array.isArray(value) || typeof value === "string"
+  }
 </script>
 
 {#each badges as badge}
--- a/packages/bbui/src/Table/ArrayRenderer.svelte
+++ b/packages/bbui/src/Table/ArrayRenderer.svelte
@@ -5,12 +5,8 @@
 
   const displayLimit = 5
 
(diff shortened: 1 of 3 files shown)


packages/bbui/src/Table/ArrayRenderer.svelte
```diff
@@ -5,8 +5,12 @@
 
   const displayLimit = 5
 
-  $: badges = value?.slice(0, displayLimit) ?? []
+  $: badges = isValidType() ? value?.slice(0, displayLimit) : []
```
aptkingston, line 8: By not passing in a parameter to `isValidType` here, the result will not correctly update when the `value` prop changes. Svelte reactive statements don't look inside functions to see what variables are used, but only inspect the variables in the statement. In this very specific case you'll actually get away with this, because you reference `value` later on in the reactive statement, so the function will correctly be re-evaluated when `value` changes. In general though, it's safer to add a parameter to your functions and pass in a value in the reactive statement. e.g.
```js
$: badges = isValidType(value) ...

const isValidType = value => {
  ...
}
```

I wouldn't worry about checking whether the value is a string either as the value should always be an actual array here. You could skip the separate function entirely and refactor to just:
```js
$: badges = Array.isArray(value) ? value.slice(0, displayLimit) : []
```

aptkingston (commented):
This change makes sense to avoid the crash if a non-array value somehow makes it in to the array renderer, but as I mentioned in the linked issue I think we should prevent conversion from a JSON column to a multi-select column as well.

packages/bbui/src/Table/ArrayRenderer.svelte
melohagan: Thanks - that's really useful to know!

I've also made the suggested refactor.

aptkingston (approved):
LGTM!

PR #6498, opened 2022-06-29
Don't allow json columns to swap to multi-select

melohagan:
## Description
JSON should not be convertible to Multi-select as JSON will always be an object.

Addresses: 
- https://github.com/Budibase/budibase/issues/6485

## Screenshots
Screenshot 2022-06-29 at 14 16 52


shogunpurple (approved):
LGTM - makes sense!

Issue #5851, opened 2022-05-12
Environment variables for setting up the first admin user 


shogunpurple:
When I'm setting up budibase for self host in either docker or kubernetes, I want to be able to pass several environment variables to configure my first admin user and "Create admin user" page.

**COS**
- Add `BB_ADMIN_USER_EMAIL` and `BB_ADMIN_USER_PASSWORD` as environment variables to the worker service to provision an admin user automatically on startup

PR #6511, opened 2022-06-30
Add env vars to automatically create initial admin user on first run

aptkingston:
## Description
This PR adds environment variables to allow automatically creating an admin user on first run.

If the `BB_ADMIN_USER_EMAIL` and `BB_ADMIN_USER_PASSWORD` variables are provided to the server (app service) then it will create an admin user with the given credentials in the `default` tenant.

Requirements:
- Self hosted only
- Single tenant only (is this needed? should we still seed the `default` tenant admin user for self hosted multi tenant?)
- Email and password must be provided
- Only runs if no users exist yet

Addresses #5851.


hosting/hosting.properties
```diff
@@ -19,3 +19,7 @@ COUCH_DB_PORT=4005
 REDIS_PORT=6379
```
mike12345567, line 19: probably shouldn't update this file anymore since its deprecated, plan to remove soon!

packages/server/src/app.ts
```diff
@@ -18,7 +18,9 @@ const { logAlert } = require("@budibase/backend-core/logging")
 const { Thread } = require("./threads")
 import redis from "./utilities/redis"
 import * as migrations from "./migrations"
-import { events, installation } from "@budibase/backend-core"
+import { events, installation, tenancy } from "@budibase/backend-core"
+import { createAdminUser, getChecklist } from "./utilities/workerRequests"
+import { DEFAULT_TENANT_ID } from "@budibase/backend-core/dist/src/constants"
```
mike12345567, line 23: Can remove the `DEFAULT_TENANT_ID` import as its unused (and replaced by later comment).

packages/server/src/app.ts
```diff
@@ -110,6 +112,32 @@ module.exports = server.listen(env.PORT || 0, async () => {
...
+    env.BB_ADMIN_USER_EMAIL &&
+    env.BB_ADMIN_USER_PASSWORD
+  ) {
+    const checklist = await getChecklist()
+    if (!checklist?.adminUser?.checked) {
+      try {
+        await createAdminUser(
+          env.BB_ADMIN_USER_EMAIL,
+          env.BB_ADMIN_USER_PASSWORD,
+          "default"
```
mike12345567, line 128: should use the `getTenantId()` function here, ideally even within the `createAdminUser` function, it will always make sure that the correct tenant ID is provided, such as in self host/non-multitenant environments passing down the `"default"` string. Means that if we want to see anything that uses the tenancy logic we can look for usages of that function and don't need to search for magic strings, consts etc.

Can get function from `import { getTenantId } from "@budibase/backend-core/tenancy"`

mike12345567 (approved):
LGTM! Just a couple of minor comments about retrieving the tenant ID, safer to use the `getTenantId` function everywhere and it'll work everything out.

packages/server/src/app.ts
Rory-Powell: Just to highlight a different import syntax, to get type support you can use
```js
import { tenancy } from "@budibase/backend-core"
tenancy.getTenantId()
```
The `/backend-core/something` imports use the `module.d.ts` file in server which won't resolve to a real implementation in the IDE

Rory-Powell: although upon removing the module file entry for `declare module "@budibase/backend-core/tenancy"` I do now get support locally, looks like this works because of the new `exports` block in the package.json of backend-core

Rory-Powell: however the actual build will still fail, so I'd say for now it's best to use the backend-core root import to get type support

Rory-Powell (commented):
What about adding this logic to worker instead? Where worker is the owner of the user domain currently. It would prevent introducing additional server -> worker api requests

packages/server/src/app.ts
```diff
@@ -18,7 +18,9 @@ const { logAlert } = require("@budibase/backend-core/logging")
 const { Thread } = require("./threads")
 import redis from "./utilities/redis"
 import * as migrations from "./migrations"
-import { events, installation } from "@budibase/backend-core"
+import { events, installation, tenancy } from "@budibase/backend-core"
+import { createAdminUser, getChecklist } from "./utilities/workerRequests"
+import { tenantSucceeded } from "@budibase/backend-core/dist/src/events/publishers/backfill"
```
Rory-Powell, line 23: This might cause problems in local dev, we shouldn't have a `dist` import in here. This looks unused so could be removed

aptkingston: > What about adding this logic to worker instead? Where worker is the owner of the user domain currently. It would prevent introducing additional server -> worker api requests

Yea I originally had it in the worker, but the problem with that is that it depends on the server being up. The worker is usually ready first so it was failing when trying to run this new stuff. Can't remember exactly which API endpoint it was trying to call but yea it depends on the server being up. I didn't really like the idea of just waiting around and pinging the server on an interval to check if it was up, so I figured it would be cleaner to instead put this in the server.

aptkingston:
> What about adding this logic to worker instead? Where worker is the owner of the user domain currently. It would prevent introducing additional server -> worker api requests

Yea I originally had it in the worker, but the problem with that is that it depends on the server being up. The worker is usually ready first so it was failing when trying to run this new stuff. Can't remember exactly which API endpoint it was trying to call but yea it depends on the server being up. I didn't really like the idea of just waiting around and pinging the server on an interval to check if it was up, so I figured it would be cleaner to instead put this in the server.

Rory-Powell:
Gotcha, I'd bet it was for syncing users in apps. Not a very big deal - just an odd circumstance where we are going server -> worker -> server. Someday we will likely combine the core api services and this will no longer be needed

mike12345567:
Yeah it was my recommendation to keep everything in the server, due to the fact that the sync depended on the server being online and also the fact that it involves the user setting some environment variables, I think its ideal if we try to keep user options/flags to the server where possible, as when we combine the APIs this should be the only service the user needs to worry about (with other services being workers/internal they shouldn't be of concern to the self hosters). Removes the need for us to migrate environment variables etc.

Rory-Powell (approved):

PR #6515, opened 2022-06-30
User pagination

mike12345567:
## Description
Addresses an issue we've experienced with the user list not being paginated, making it inaccessible/unusable when a large number of users is added.

This update adds a new endpoint `POST /api/global/users/search` which allows searching the users as well as paginating them. The old endpoint for fetching all users is still intact but deprecated. It had to be left in place due to its use in the public API, which currently allows retrieving all users.

## To do
There are still a couple of issues that need to be resolved, but will come in a separate update as they are a bit less pressing.
1. When a custom role is deleted right now it retrieves all global users to see if any of them have the role assigned. This is a relatively poor way of doing this, it only applies to custom roles right now but there is definitely a better way to achieve this, involving searching the app database on the lucene index and then using this to update global users.
2. The public API could do with an update but it would be breaking as it would then apply pagination. A better approach for the time being would be to apply limits within it that stop retrieval past a certain number of users and allow these to be changed via environment variables. This should be carried out for all searching structures in the API as currently all of the endpoints bar rows just query everything out of couch and search in memory.


aptkingston (approved):
LGTM! Random question, but does everything work alright if you spam previous/next page? Might be hard to test on a local server with not many users, but just wondering if state would mess up if you pressed next page multiple times during a request for example. That was one of the things I had to account for when handling pagination with app data.

mike12345567:
> LGTM! Random question, but does everything work alright if you spam previous/next page? Might be hard to test on a local server with not many users, but just wondering if state would mess up if you pressed next page multiple times during a request for example. That was one of the things I had to account for when handling pagination with app data.

Very good point, I'll add something to handle the loading state as well - locally its fast enough to respond too quick, but it could break in a real environment.

PR #6535, opened 2022-07-03
SSO rest request refresh

deanhannigan:
## Description
Systems with OAuth authentication configurations will attempt to refresh a users access tokens if a 401 is encountered during a query request.

There will only be 1 attempt made to refresh the token per query request. 

Upon a successful refresh response, the users accessToken will be updated and the query will be reattempted.
The refreshToken will be updated if one is sent back by the authentication provider.

Addresses:  [3827](https://github.com/Budibase/budibase/issues/3827)


packages/backend-core/src/db/utils.ts
```diff
@@ -416,11 +422,11 @@ export const getPlatformUrl = async (opts = { tenantAware: true }) => {
       platformUrl = platformUrl.replace("://", `://${tenantId}.`)
     }
   } else if (env.SELF_HOSTED) {
-    const db = getGlobalDB()
+    const dbx = db ? db : getGlobalDB()
```
Rory-Powell, line 425: I think the `db` here will always be global db, and it's no extra work to retrieve the db instance using the `getGlobalDB` function as that's initialised as a part of the tenancy middleware.

To keep the method signature of `getPlatformUrl` more clean we can remove the passed in `db` arg.

packages/backend-core/src/middleware/authenticated.js
```diff
@@ -128,6 +128,8 @@ module.exports = (
       }
       if (!user && tenantId) {
         user = { tenantId }
+      } else {
+        delete user.password
```
Rory-Powell, line 132: looks like this is the second time we're doing this in this function. In the first case we delete when
- there is a session
- there is a matched user

then we proceed to populate the user again if there is an api key and we're not already authenticated, the default public api flow. So is that the case we're covering here? I do like where you've added this near the end, I think we could even remove the other usage as this should over all cases

packages/backend-core/src/middleware/passport/utils.js
```diff
@@ -1,14 +1,39 @@
...
+
 exports.authError = function (done, message, err = null) {
   return done(
     err,
     null, // never return a user
     { message: message }
   )
 }
+
+exports.ssoCallbackUrl = async (db, config, type) => {
```
Rory-Powell, line 21: Looks like this is a copy paste from the existing function in worker, we should remove the worker code and re-use the backend-core function to remove the duplication / potential for future bugs

packages/worker/src/api/controllers/global/auth.ts
```diff
@@ -70,7 +70,7 @@ async function authInternal(ctx: any, user: any, err = null, info = null) {
 export const authenticate = async (ctx: any, next: any) => {
   return passport.authenticate(
     "local",
-    async (err: any, user: User, info: any) => {
+    async (err: any, user: any, info: any) => {
```
Rory-Powell, line 73: How come the removed type reference?

packages/backend-core/src/middleware/passport/google.js
```diff
@@ -57,5 +59,19 @@ exports.strategyFactory = async function (config, callbackUrl, saveUserFn) {
     )
   }
 }
+
+exports.fetchStrategyConfig = async function (googleConfig) {
```
Rory-Powell, line 63: I think we won't want to fallback to the environment variables here, as they will be set in the cloud for datasource integration for google sheets using our own google creds. I think that behaviour should only be used for the datasource integration and not for the sso process, so best to just read the config here and can probably remove this function

packages/backend-core/src/middleware/passport/local.js
```diff
@@ -55,6 +55,20 @@ exports.authenticate = async function (ctx, email, password, done) {
   if (await compare(password, dbUser.password)) {
     const sessionId = newid()
     const tenantId = getTenantId()
+
+    if (dbUser.provider || dbUser.providerType || dbUser.pictureUrl) {
+      delete dbUser.provider
+      delete dbUser.providerType
+      delete dbUser.pictureUrl
```
Rory-Powell, line 62: We should go the other way and prevent certain users from setting / logging in using a password if they are a sso user. There are some tickets in the backlog to do this. This might be a risky change to make at the moment as it'll lock someone out from sso if they do log in with their existing password

packages/backend-core/src/middleware/passport/oidc.js
```diff
@@ -89,11 +91,24 @@ function validEmail(value) {
  * from couchDB rather than environment variables, using this factory is necessary for dynamically configuring passport.
  * @returns Dynamically configured Passport OIDC Strategy
  */
-exports.strategyFactory = async function (config, callbackUrl, saveUserFn) {
+exports.strategyFactory = async function (config, saveUserFn) {
```
Rory-Powell, line 94: It's slightly confusing that the `config` here isn't the config loaded from the db, rather it's an enriched form of the config that's aware of the oidc elements from the provider. Maybe a different name could help here like `strategyConfig` which makes it more clear that this argument is the result of the below function

Rory-Powell (commented):
Really nice job on this! great to see the new library being used seamlessly with our existing configurations etc. For the case you mentioned where the `refreshToken` wasn't set did you get the the bottom of it?

deanhannigan:
Hi @Rory-Powell , thanks for taking a look! Regarding the issue where the refresh token wasn't being retrieved, this was specifically related to the Google OAuth flow. The config property `accessType: "offline"` wasn't being passed to passport, so the value was never received. If anyone has used the google auth option and wishes to also use the authentication within their applications, they would have to revoke the application access and re-authenticate. Google only issue the refreshToken on the first authentication.

packages/backend-core/src/db/utils.ts
deanhannigan: Fair enough, I've set this to a fixed call to the global db and cleared out any references to the old signature.

packages/backend-core/src/middleware/authenticated.js
deanhannigan: Yes, the second delete was intended to cover users verified via an API key. As you point out, the second one would actually cover both auth flows.

packages/backend-core/src/middleware/passport/utils.js
deanhannigan: Yep, I shifted this out and should have deleted the original. Update incoming.

packages/worker/src/api/controllers/global/auth.ts
deanhannigan: I think I changed this to accommodate updates that have since been removed.

packages/backend-core/src/middleware/passport/google.js
deanhannigan: Update incoming.

packages/backend-core/src/middleware/passport/local.js
deanhannigan: No problem. When I was shifting between providers and basic auth, I could see these elements were hanging around in the DB. Backing this out for now.

packages/backend-core/src/middleware/passport/oidc.js
deanhannigan: Update incoming

packages/worker/src/api/controllers/global/auth.ts
```diff
@@ -197,7 +170,9 @@ export const googlePreAuth = async (ctx: any, next: any) => {
   const strategy = await google.strategyFactory(config, callbackUrl, users.save)
 
   return passport.authenticate(strategy, {
-    scope: ["profile", "email"],
+    scope: ["profile", "email", "https://www.googleapis.com/auth/spreadsheets"],
```
Rory-Powell, line 173: I think we can remove the spreadsheets addition on this one, it's added to it's own passport integration in `/middleware/passport/datasource/google.js`

packages/worker/src/api/controllers/global/auth.ts
```diff
@@ -1,49 +1,22 @@
...
-    callbackUrl += `/${getTenantId()}`
-  }
-  callbackUrl += `/${type}/callback`
-
-  return `${publicConfig.platformUrl}${callbackUrl}`
-}
-
 export const googleCallbackUrl = async (config: any) => {
-  return ssoCallbackUrl(config, "google")
+  return ssoCallbackUrl(getGlobalDB(), config, "google")
```
Rory-Powell, line 15: We don't need to pass the global db like this if we call `getGlobalDB()` inside `ssoCallbackUrl` so we could remove db from the method signature completely

Rory-Powell (approved):
LGTM 👍  just left one other comment about the db usage in `ssoCallbackURL`

Issue #6350, opened 2022-06-16
Conditions - Add a clear indicator to settings panel


AndyWhann:
Currently it is difficult to tell when conditions have been set on the component. All we do to highlight this is add a `*` to the label when there are conditions.

## Suggestion

I suggest we add a heading to the conditions section that will show:
-  `No conditions set`
- `X conditions set`

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/64658722-fa4c-4536-a1e9-8f8dccb36116)

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/9d2fe082-b3f0-462a-9ad2-beba281c052b)

PR #6558, opened 2022-07-05
Feature/component condition count

deanhannigan:
## Description
When display conditions are set for a component in the builder, the number of conditions set will be displayed to the user.

Addresses: https://github.com/Budibase/budibase/issues/6350

## Screenshots

**No conditions**
none

**One condition**
Screenshot 2022-07-05 at 10 59 42

**Multiple Conditions**
Screenshot 2022-07-05 at 10 59 52


aptkingston (approved):
LGTM! Being extremely pedantic but could we move the text up about 5px? I think it's just the gap in the detail summary at the moment that's spacing it, so maybe try a -5px top margin?

Issue #6560, opened 2022-07-05
App crashes if you delete a column that was being used by a Dynamic Filter


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.214
  - App Version: 1.0.214

**Describe the bug**
App crashes if you delete a column that was being used by a Dynamic Filter.

**To Reproduce**
1. Add an internal table with columns, including a static formula column
2. In design add a dynamic filter to a data provider for that table. Include multiple columns and the static formula column.
3. Delete the static formula column
4. Run the app.
5. App crashes when you click on Dynamic Filter

**Expected behavior**
Deleting a column that is being used by a dynamic filter should also be removed from the schema fields

**Screenshots**
![Screenshot 2022-07-05 at 14 33 40](https://user-images.githubusercontent.com/101575380/177340150-a2399f7d-02fa-4d33-b52e-78bd8302ae7c.png)
![Screenshot 2022-07-05 at 14 34 08](https://user-images.githubusercontent.com/101575380/177340253-e241e29c-39a3-48ee-91bc-f96bfd5bf7df.png)

PR #6561, opened 2022-07-05
Only add defined schema fields to dynamic filter

melohagan:
## Description
Deleting a column used in a dynamic filter was causing an app crash. 
Only add a dynamic filter schema field if it is not **undefined** (i.e. the column no longer exists)

Addresses: 
- https://github.com/Budibase/budibase/issues/6560


mike12345567 (approved):
LGTM!

PR #6563, opened 2022-07-05
Don't try to check externalType of related tables

melohagan:
## Description
No rows were being returned for external datasources when using relationships.
This was because it was trying to find the columns of the related table in the parent schema. 

Addresses: 
- https://github.com/Budibase/budibase/issues/5708

## Screenshots
Screenshot 2022-07-05 at 17 04 57

Screenshot 2022-07-05 at 17 05 19


mike12345567 (approved):
LGTM!

PR #6571, opened 2022-07-06
User Groups and User Management redesign

PClmnt:
### User groups and Users UI redesign

This PR introduces the concept of User Groups as well as delivering an updated experience for the Users UI and how they are managed. 

User Groups will be a Pro feature will allow admins to manage how users are managed and added to their application, as well as the access and control of them.

## User Groups

A new user groups tab has been added to the portal section of the builder, from there, users can manage their groups and add users from and to them. 

New groups can be added through the following modal, I've also added a new icon picker fopr spectrum icons (maybe we could use this in the app icon section as well?)
image

Once a group is created, it can be viewed, along with the count of assosicated apps and users from the list screen:

### User groups and Users UI redesign

This PR introduces the concept of User Groups as well as delivering an updated experience for the Users UI and how they are managed. 

User Groups will be a Pro feature will allow admins to manage how users are managed and added to their application, as well as the access and control of them.

## User Groups

A new user groups tab has been added to the portal section of the builder, from there, users can manage their groups and add users from and to them.  It is gated by a check for a `Pro` license and the screen looks like the following whenever a user is on a free license: 
image

New groups can be added through the following modal, I've also added a new icon picker fopr spectrum icons (maybe we could use this in the app icon section as well?)
image

Once a group is created, it can be viewed, along with the count of assosicated apps and users from the list screen:

image

Clicking into a group leads to the group management screen, where users can be added to the group, as well as what apps are associated with it can be viewed. You can search users in order to add them to your group

image

When no users or apps are assosicated with the group, we just show empty lists with some text:
image

## Users UI

The User management UI has been refactored heavily to provide a better and easier experience to manage users

The first thing to notice is how there are new Budibase roles - App User / Developer / Admin. These correlate under the hood to what we had before, but just in this case we've added App User, to make it more explicit that the only action a user with this level can do, is view / access the application. The BBUI table component has been slightly updated here, to follow Andy's designs, the header border has been removed, i'm passing a prop to the component in order to do that. You can also bulk delete users now by clicking the checkboxes and then clicking the delete button. 

image

After clicking into a user, you can manage different aspects such as the groups (gated by a pro plan license, nothing is shown for it if you don't have one), the same UX is used here as in the group page for adding a new group, you can search by group and add a user directly to one here. You can also see what app and associated role the user has for it. You can switch the users role on this screen as well through the dropdown.  Clicking on the app will take you to the app overview screen

image

The adding of users has also been refactored quite heavily, you can add multiple users and assign their roles from this screen: 

image

Clicking through then gives you the option of what way to create the user:

image

Clciking the Generate Passwords prompt will then take you to a modal which gives you the option to download the passwords via CSV or just copy them directly

image

the same flow can be done with the importing of users via CSV. 

image

## App Management

The ability to manage users and groups on a per app basis has also been added, going into the app overview screen you can get a summary of how many users have been assigned to an app, as well as a count of the total. the access tab also provides the ability to manage groups and users on a per app basis. 

image

image

Clicking on the role dropdowns for groups or users will update the role, and new users / groups can be added by clicking on the assign users button

image

this is a new component specifically for this work that takes a prop that is an array of objects (users object and groups object) and presents it alongside a picker for the role. you can search through and select. You can add as many users and groups as you want through this interface

image

For the actual User Group roles / access handling, it's currently set up ls follow

- An individual user's role will always override the role of the group it is in, i.e if a user has `BASIC` access but the group has `POWER` access the user will **_not_** be able to access any Power screens within the application.
- The same is true for the opposite, if a user has `POWER` access but the group only has `BASIC` access, the user **_will_** be able to access `POWER` as well as `BASIC` screens in the application. 
- If a user has no specific access assigned, then it will always use the access of the group it is in.


packages/bbui/src/Form/InputDropdown.svelte
```diff
@@ -0,0 +1,55 @@
+<script>
```
PClmnt, line 1: Need to move this out of BBUI into builder components section

packages/bbui/src/Form/PickerDropdown.svelte
```diff
@@ -0,0 +1,125 @@
+<script>
```
PClmnt, line 1: Need to move this out of BBUI into builder components section

packages/builder/src/components/common/RoleSelect.svelte
```diff
@@ -0,0 +1,24 @@
...
+  {quiet}
+  bind:value
+  on:change
+  options={$roles}
+  getOptionLabel={role => role.name}
+  getOptionValue={role => role._id}
+  getOptionColour={role => RoleUtils.getRoleColour(role._id)}
+  {placeholder}
+  {error}
+/>
```
aptkingston, line 24: Now that the new design UI is merged, we have this `RoleSelect` component already in `src/components/design/settings/controls/RoleSelect.svelte`. I coped that component into here since we expected this PR to be merged before the new design UI was. I'll jump on this branch and sort out this usage, and the role utils which I also copied from that new branch.

NEOLPAR:

packages/builder/src/components/common/RoleSelect.svelte
PClmnt: Cool, thanks @aptkingston

packages/backend-core/src/db/conversions.js
```diff
@@ -50,3 +50,8 @@ exports.getProdAppID = appId => {
   const rest = split.join(APP_DEV_PREFIX)
   return `${APP_PREFIX}${rest}`
 }
+
+exports.extractAppId = id => {
```
mike12345567, line 54: what does this extract the app ID from? We would usually call the app ID the full `app_<uuid>` - might be worth naming this if it is for extracting the UUID from an app ID.

packages/backend-core/src/users.js
```diff
@@ -13,12 +13,32 @@ exports.getGlobalUserByEmail = async email => {
...
+    response = []
+  }
+  return Array.isArray(response) ? response : [response]
+}
+
+exports.getGlobalUserByAppPage = (appId, user) => {
+  if (!user) {
+    return
+  }
+  return `${getProdAppID(appId)}${SEPARATOR}${user._id}`
```
mike12345567, line 41: Might want to build this sequence in the `server/src/db/utils` file - generally try to hide the nastiness of ID building etc with the `SEPARATOR` down there, makes it easier to re-use/find later on too.

packages/server/src/db/utils.js
```diff
@@ -28,6 +28,7 @@ const DocumentTypes = {
   TABLE: "ta",
   ROW: "ro",
   USER: "us",
+  GROUP: "gr",
```
mike12345567, line 31: This appears to be doubled up, its in the backend core and here in the server, overwriting it could be dangerous as it could end up being changed in the server but not the backend core.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -22,6 +22,65 @@ export const save = async (ctx: any) => {
...
+      let oldGroup = await db.get(groupId)
+      groupsToSave.push(oldGroup)
+    })
+  }
+
+  try {
+    let response: any[] = []
+    for (const user of newUsers) {
+      response.push(
+        await users.save(user, {
```
mike12345567, line 53: Ideally, this bulk save functionality could be built with the `bulkDocs` API - https://pouchdb.com/guides/bulk-operations.html, if you are updating say 100 users here this is very inefficient as lots of individual gets and writes, vs extracting everything you need upfront, building all the docs and then inserting them all in a single fast operation.

Just thinking if someone comes along and tries to upload their entire companies worth of users, they'll run into problems/very slow response times, it might feel broken.

A method to achieve this could be breaking out the synchronous component of the `users.save` core functionality, calling it to gather up all the docs you need to write, getting them into an array and then calling the `bulkDocs` functionality after this for loop.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -22,6 +22,65 @@ export const save = async (ctx: any) => {
...
+  const newUsers: any[] = []
+  const db = tenancy.getGlobalDB()
+  const currentUserEmails =
+    (await users.allUsers())?.map((x: any) => x.email) || []
+  for (const newUser of newUsersRequested) {
+    if (
+      newUsers.find((x: any) => x.email === newUser.email) ||
+      currentUserEmails.includes(newUser.email)
+    )
+      continue
```
mike12345567, line 37: This continue is just floating, it should be wrapped with curlys for safety (hard to see immediately that the if statement only applies to the continue).

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -22,6 +22,65 @@ export const save = async (ctx: any) => {
...
+
+    // delete passwords and add to group
+    newUsers.forEach(user => {
+      delete user.password
+    })
+
+    if (groupsToSave.length) {
+      groupsToSave.forEach(async (userGroup: UserGroup) => {
+        userGroup.users = [...userGroup.users, ...response]
+        await db.put(userGroup)
```
mike12345567, line 69: Same as above, this one is less important as I assume there couldn't be that many groups, but when you are doing a raw `db.put` in a loop of some sort you could extract it to an array and then perform a `db.bulkDocs` after the loop for efficiency reasons.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +143,60 @@ export const adminUser = async (ctx: any) => {
...
+
+  // Remove asssosicated groups
+  if (groups) {
+    for (const groupId of groups) {
+      let group = await db.get(groupId)
+      let updatedUsersGroup = group.users.filter(
+        (groupUser: any) => groupUser.email !== user.email
+      )
+      group.users = updatedUsersGroup
+      await db.put(group)
```
mike12345567, line 160: Same as above, could do this with bulk docs, but again not critical.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +143,60 @@ export const adminUser = async (ctx: any) => {
...
+  const { userIds } = ctx.request.body
+  const db = tenancy.getGlobalDB()
+
+  let deleted = 0
+
+  for (const id of userIds) {
+    let user: User = await db.get(id)
+    let groups = user.userGroups
+
+    await users.destroy(id, ctx.user)
```
mike12345567, line 179: Same as all above comments, this could be achieved with bulk operations, you can perform a bulk read of all the IDs initially (perform a lookup on the `allDocs` and use the `keys` attribute, e.g. `db.allDocs({ keys: userIds, include_docs: true })` this retrieves all of the users in a single speedy operation, then you can build up all the documents that need to be deleted, and attach a `_deleted: true` property on them during the `bulkDocs` call to delete them.

Ideally after seeing this, the logic for bulk saving and deletion of users could be extracted to the users core in the backend-core library, that way you could split out the synchronous portions of `users.destroy` and `users.save` to use in `users.bulkSave` and `users.bulkDestroy`.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +143,60 @@ export const adminUser = async (ctx: any) => {
...
+
+    if (groups) {
+      for (const groupId of groups) {
+        let group = await db.get(groupId)
+        let updatedUsersGroup = group.users.filter(
+          (groupUser: any) => groupUser.email !== user.email
+        )
+        console.log(updatedUsersGroup)
```
mike12345567, line 187: Remove console log.

mike12345567 (changes requested):
Looks great overall, I've a few comments about the bulk user save/destroy functionality, these won't scale well into importing hundreds/thousands of users, ideally these should be performed with bulk database operations rather than lots of individual calls.

If you need a hand with this just let me know, there are some examples of bulk APIs (internal rows API) were we do similar things!

packages/worker/src/api/controllers/global/users.ts
PClmnt: You mentioned backend-core here, is that right? Is there a reason for that, over putting the functionality into the users sdk within the worker? That seems to be where most of this type of functionality seems to go.

packages/worker/src/api/routes/index.js
```diff
@@ -1,3 +1,4 @@
+const { api } = require("@budibase/pro")
```
Rory-Powell, line 1: Nice one 👍

packages/worker/src/api/controllers/global/users.ts
Rory-Powell: We need to be very careful around this WRT events, in the current `bulkCreate` impl it will bypass user save events as we effectively use the save as a validator / object generator instead and defer the save. The events are used for
- analytics 
- licensing

It'll also skip user uniqueness checks. I think we should remove this flag as it's totally unsafe and interferes with application logic in a few ways. 

There's an option for this going forward

**Asynchronous saves**
- Return the bulk api call immediately
- Process saves as a background job, one by one
- Notify the user when the job is done

This relies on what we have been calling just 'background jobs' for our application, which still needs spec'd out and implemented properly. The only other potential option would be something like

**Synchronous saves with back populated events**
- Refactor the `save` logic to have a shared `newUser` function
- Use this in both regular `save` calls as well as a `saveBulk` implementation
- Retroactively make sure that we handle events after the db save has been done in the bulk impl

However it still have a background job element, just not something that needs to be communicated to the user.
I don't think it's feasible though, especially in cloud as we need to perform these uniqueness checks:
- check budibase users inside the tenant
- check budibase users in other tenants
- check root account users in account portal

Skipping this even temporarily could be very bad. 

I think until we have background jobs, handling user saves will need to be a synchronous operation. In future we can detect if size is greater than X and create the user in batches in the background.

PClmnt: I pushed some changes just as you commented Rory, is this closer to what you meant in regards to the synchronous saves?

packages/worker/src/sdk/users/users.ts
```diff
@@ -203,6 +228,131 @@ export const save = async (
...
+          requirePassword: user.requirePassword,
+          bulkCreate: false,
+        },
+        tenantId
+      )
+    )
+  })
+
+  const usersToBulkSave = await Promise.all(usersToSave)
+  await quotas.addDevelopers(() => db.bulkDocs(usersToBulkSave), builderCount)
```
Rory-Powell, line 293: Are there changes to push to the pro repo around this? it's `addDeveloper` currently without the `s`

packages/worker/src/sdk/users/users.ts
```diff
@@ -203,6 +228,131 @@ export const save = async (
...
+
+  // Post processing of bulk added users, i.e events and cache operations
+  for (const user of usersToBulkSave) {
+    delete user.password
+    await eventHelpers.handleSaveEvents(user, null)
+    await apps.syncUserInApps(user._id)
+    await addTenant(tenantId, user._id, user.email)
```
Rory-Powell, line 300: It's probably ok to leave this out, you shouldn't be able to bulk create without having first created the admin user. We should probably do that check like we do with the initial user creation to make sure. Just would be unnecessary db work to do the tenant creation logic for every user

packages/worker/src/api/controllers/global/users.ts
Rory-Powell: The approach you've put in peter is pretty much bang on the second option peter 👍 

Yeah you have called out the main issue here michael, the account check is the inefficient part here. Going forward as a part of the active user pricing there is also now a user representation in the account portal dynamo table. Whenever we create a user there will be a minimal representation there, and when we delete a user we will be setting a `deletedAt` attribute. These will be run as a part of the user created and user deleted events so they will be covered in this implementation, but the advantage gained here in terms of fast response time with a bulk write is now gone as we will be using a network call to account portal even if we are skipping the root account exists checks and looping db inserts. 

Traditionally a microservice architecture would use a message queue for this, where we could do the creates very quickly up front and dump all the users on a queue for dependant processing in account portal. This is more or less the same approach used for what we will call background jobs. Still there's no getting away from the uniqueness check up front here. We could in theory remove the account uniqueness check, as the budibase user would also be created in the db and should meet the points you said about checking up front michael, however there are still some issues:

- Self hosted account email address could be used for a cloud internal user (breaks login)
- Unverified cloud account user added to different tenant as internal user. This isn't something we've tested but will for sure mess with tenancy when the user verifies. Would be a common scenario when a few new people are trying to get off the ground with budibase and are creating accounts, trying to add each other to their accounts etc. 

We're in an unfortunate situation around multiple user stores. If we had a single source of truth for all users which would be a more common scenario we wouldn't have problems like this

packages/worker/src/sdk/users/users.ts
PClmnt: yep, pushing the pro changes up shortly

packages/builder/src/pages/builder/portal/manage/users/_components/ImportUsersModal.svelte
```diff
@@ -0,0 +1,157 @@
...
+
+  $: invalidEmails = []
+  $: hasGroupsLicense = $auth.user?.license.features.includes(
+    Constants.Features.USER_GROUPS
+  )
+
+  const validEmails = userEmails => {
+    if (userEmails.length > MAX_USERS_UPLOAD_LIMIT) {
+      notifications.error(
+        `Max limit for upload is 1000 users. Please reduce file size and try again.`
```
PClmnt, line 32: @shogunpurple @mike12345567 @Rory-Powell Checking for max length in these locations. this seem ok?

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -22,6 +24,62 @@ export const save = async (ctx: any) => {
   }
 }
 
+export const bulkCreate = async (ctx: any) => {
+  let { users: newUsersRequested, groups } = ctx.request.body
+
+  if (!env.SELF_HOSTED && newUsersRequested.length > MAX_USERS_UPLOAD_LIMIT) {
+    ctx.throw(
+      400,
+      "Max limit for upload is 1000 users. Please reduce file size and try again."
```
PClmnt, line 33: @shogunpurple @mike12345567 @Rory-Powell Checking for max length in these locations. this seem ok?

PClmnt:
Just trying to fix up this failing test, and then hopefully should be good to go. Something to do with the user not syncing properly to the app metadata.

packages/backend-core/src/errors/licensing.js
```diff
@@ -12,6 +13,11 @@ const context = {
       limitName: err.limitName,
     }
   },
+  [codes.FEATURE_DISABLED]: err => {
+    return {
+      limitName: err.limitName,
```
Rory-Powell, line 18: I think this should be `featureName` instead of limit

packages/builder/src/pages/builder/portal/manage/users/_components/ImportUsersModal.svelte
Rory-Powell: It would probably be beneficial to do it on the backend as well so that this can't be circumvented

Rory-Powell: Nevermind, just seeing your second comment below - all good!

packages/worker/src/api/controllers/global/users.ts
mike12345567: @PClmnt - yeap blocking on the controller should work well.

packages/backend-core/src/events/publishers/group.ts
```diff
@@ -0,0 +1,64 @@
...
+  await publishEvent(Event.USER_GROUP_DELETED, properties)
+}
+
+export async function usersAdded(emails: string[], group: UserGroup) {
+  const properties: GroupUsersAddedEvent = {
+    emails,
+    count: emails.length,
+    groupId: group._id as string,
+  }
+  await publishEvent(Event.USER_GROUP_USER_ADDED, properties)
```
Rory-Powell, line 41: Could match the plurality here with `USER_GROUP_USERS_ADDED`. We also shouldn't be sending the emails in the event

packages/backend-core/src/events/publishers/group.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export async function usersDeleted(emails: string[], group: UserGroup) {
+  const properties: GroupUsersDeletedEvent = {
+    emails,
+    count: emails.length,
+    groupId: group._id as string,
+  }
+  await publishEvent(Event.USER_GROUP_USER_REMOVED, properties)
```
Rory-Powell, line 50: Similar point in the plurality

packages/backend-core/src/events/publishers/group.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export async function createdOnboarding(groupId: string) {
+  const properties: GroupsAddedOnboarding = {
```
Rory-Powell, line 54: Should be `GroupAddedOnboardingEvent`?

packages/backend-core/src/events/publishers/group.ts
```diff
@@ -0,0 +1,64 @@
...
+    groupId: groupId,
+    onboarding: true,
+  }
+  await publishEvent(Event.USER_GROUP_ONBOARDING, properties)
+}
+
+export async function permissionsEdited(roles: UserGroupRoles) {
+  const properties: UserGroupRoles = roles
```
Rory-Powell, line 62: Similar to above, we should have an event interface for this one too. Is there anything sensitive in the roles?
Looks like just key value. Is that just appId -> roleId?

packages/server/src/api/routes/tests/user.spec.js
```diff
@@ -23,19 +23,6 @@ describe("/users", () => {
   })
 
   describe("fetch", () => {
-    it("returns a list of users from an instance db", async () => {
```
Rory-Powell, line 26: what was the reason for this being removed?

packages/types/src/documents/global/user.ts
```diff
@@ -13,7 +13,9 @@ export interface User extends Document {
   providerType?: string
   password?: string
   status?: string
-  createdAt?: number // override the default createdAt behaviour - users sdk historically set this to Date.now()
+
+  createdAt?: number
+  userGroups?: string[] // override the default createdAt behaviour - users sdk historically set this to Date.now()
```
Rory-Powell, line 18: comment on wrong line?

packages/server/src/api/routes/tests/user.spec.js
PClmnt: I think it needs to be re-added again, i'm trying to fix a bug around this area and thought for a moment that because now uses have no access by default (you need specifically added to an app, we didn't need this test), but now i think that's unrelated, so I will add it back

packages/types/src/documents/global/userGroup.ts
```diff
@@ -0,0 +1,15 @@
+import { Document } from "../document"
+import { User } from "./user"
+export interface UserGroup extends Document {
+  name: string
+  icon: string
+  color: string
+  users: User[]
+  apps: any
```
Rory-Powell, line 8: It would be nice to have an indication here of some of the structure, is it full apps or just ids, is it an object or an array etc

packages/types/src/sdk/events/event.ts
```diff
@@ -150,6 +150,15 @@ export enum Event {
   TENANT_BACKFILL_FAILED = "tenant:backfill:failed",
   INSTALLATION_BACKFILL_SUCCEEDED = "installation:backfill:succeeded",
   INSTALLATION_BACKFILL_FAILED = "installation:backfill:failed",
+
+  // USER
+  USER_GROUP_CREATED = "user_group:created",
+  USER_GROUP_UPDATED = "user_group:updated",
+  USER_GROUP_DELETED = "user_group:deleted",
+  USER_GROUP_USER_ADDED = "user_group_user:added",
```
Rory-Powell, line 158: I think the second `user` bit should be on the other side here. For existing events it generally is `entity:action`. So in this case `user_group:user_added`. Same for the events below

packages/types/src/sdk/events/userGroup.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface GroupUpdatedEvent extends BaseEvent {
+  groupId: string
+}
+
+export interface GroupDeletedEvent extends BaseEvent {
+  groupId: string
+}
+
+export interface GroupUsersAddedEvent extends BaseEvent {
+  emails: string[]
```
Rory-Powell, line 16: Mentioned above, this should be removed. Could be replaced with `userIds`, or could just go away if we don't have any intention for tracking specific users in groups in analytics

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -22,6 +24,62 @@ export const save = async (ctx: any) => {
...
+  }
+
+  try {
+    let response = await users.bulkCreate(newUsersRequested, groups)
+
+    if (groupsToSave.length) {
+      let groupsPromises: any = []
+      groupsToSave.forEach(async (userGroup: UserGroup) => {
+        userGroup.users = [...userGroup.users, ...response]
+        groupsPromises.push(db.put(userGroup))
```
Rory-Powell, line 54: The save of the user group should probably go through the groups sdk from pro

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +142,67 @@ export const adminUser = async (ctx: any) => {
...
+  // Remove asssosicated groups
+  if (groups) {
+    let groupsPromises = []
+    for (const groupId of groups) {
+      let group = await db.get(groupId)
+      let updatedUsersGroup = group.users.filter(
+        (groupUser: any) => groupUser.email !== user.email
+      )
+      group.users = updatedUsersGroup
+      groupsPromises.push(db.put(group))
```
Rory-Powell, line 160: Similar to above, should consider using the groups sdk from pro to do any modifications to a group

packages/worker/src/sdk/users/users.ts
```diff
@@ -203,6 +229,130 @@ export const save = async (
...
+    const afterCreateTenant = () =>
+      migrations.backPopulateMigrations({
+        type: MigrationType.GLOBAL,
+        tenantId,
+      })
+    await tenancy.tryAddTenant(tenantId, _id, email, afterCreateTenant)
+  }
+}
+
+export const bulkCreate = async (newUsersRequested: any[], groups: any) => {
```
Rory-Powell, line 247: Should use the available types here for both args

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +142,67 @@ export const adminUser = async (ctx: any) => {
...
+  }
+
   ctx.body = {
     message: `User ${id} deleted.`,
   }
 }
 
+export const bulkDelete = async (ctx: any) => {
+  const { userIds } = ctx.request.body
+  const db = tenancy.getGlobalDB()
```
Rory-Powell, line 173: The main implementation here could be moved into the user sdk

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +142,67 @@ export const adminUser = async (ctx: any) => {
 
 export const destroy = async (ctx: any) => {
   const id = ctx.params.id
+  const db = tenancy.getGlobalDB()
```
Rory-Powell, line 145: The changes here could be moved into the implementation of `users.destroy` so that we don't have split logic. 
Or based on the below comment about using pro for the group modifications it could read like
```ts
await users.destroy(id, ctx.user)
await groups.removeUser(groups, userId)
```
I would vote for putting the group updates inside `users.destroy` so that it's always captured.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +142,67 @@ export const adminUser = async (ctx: any) => {
...
+
   await users.destroy(id, ctx.user)
+
+  // Remove asssosicated groups
+  if (groups) {
+    let groupsPromises = []
+    for (const groupId of groups) {
+      let group = await db.get(groupId)
+      let updatedUsersGroup = group.users.filter(
+        (groupUser: any) => groupUser.email !== user.email
```
Rory-Powell, line 157: Should probably use `_id` here for additional safety for when we roll out ability to update email addresses. Group users would be another thing we need to update it we didn't use id here

packages/worker/src/sdk/users/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./users"
+export * from "./events"
```
Rory-Powell, line 2: See comments about moving some logic from controllers into sdk. Would mean we don't need to expose the user events outside of the user sdk which sounds right

Rory-Powell (changes requested):
I think there's some test coverage that we'll have to add here. 

The most value we could get from the fewest tests would be
- Any new user management endpoints
- New group endpoints (these should go into the pro repo)

packages/backend-core/src/events/publishers/group.ts
PClmnt: yep it's literally appId -> roleId

packages/types/src/documents/global/userGroup.ts
```diff
@@ -0,0 +1,15 @@
+import { Document } from "../document"
+import { User } from "./user"
+export interface UserGroup extends Document {
+  name: string
+  icon: string
+  color: string
+  users: User[]
```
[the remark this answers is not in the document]
PClmnt: Yea, we need it in numerous locations in the new groups frontends

packages/server/src/api/routes/tests/user.spec.js
PClmnt: added back and test is functioning correctly.

packages/server/src/utilities/global.js
```diff
@@ -89,6 +106,8 @@ exports.getGlobalUsers = async (users = null) => {
   if (!appId) {
     return globalUsers
   }
+  console.log("maybe??")
```
Rory-Powell, line 109: log line

packages/types/src/documents/global/user.ts
```diff
@@ -13,7 +13,9 @@ export interface User extends Document {
   providerType?: string
   password?: string
   status?: string
+
```
Rory-Powell, line 16: probably don't need this newline

packages/types/src/documents/global/userGroup.ts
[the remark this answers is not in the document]
Rory-Powell: Gotcha. tbh it's not ideal this way as we will get out of date information e.g. if a user's name is updated, their email changes etc, then we have to make sure that user is updated inside every group. Are those scenarios catered for?

An improvement would be to only store the id's, and in the front end load the users via an api call i.e. `/api/global/groups/:groupId/users`, ideally paginated. This could mean we could page through the list of users in a group rather than returning potentially thousands of nested, out of date users.

packages/types/src/sdk/events/event.ts
```diff
@@ -150,6 +150,15 @@ export enum Event {
   TENANT_BACKFILL_FAILED = "tenant:backfill:failed",
   INSTALLATION_BACKFILL_SUCCEEDED = "installation:backfill:succeeded",
   INSTALLATION_BACKFILL_FAILED = "installation:backfill:failed",
+
+  // USER
+  USER_GROUP_CREATED = "user_group:created",
+  USER_GROUP_UPDATED = "user_group:updated",
+  USER_GROUP_DELETED = "user_group:deleted",
+  USER_GROUP_USERS_ADDED = "user_group:user_added",
+  USER_GROUP_USERS_REMOVED = "user_group_:users_deleted",
```
Rory-Powell, line 159: some extra `_` here and below

packages/types/src/documents/global/userGroup.ts
```diff
@@ -0,0 +1,15 @@
...
+  apps: any[]
```
Rory-Powell, line 8: What does this structure look like? a type would be useful here

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -84,12 +116,28 @@ export const adminUser = async (ctx: any) => {
...
   ctx.body = {
     message: `User ${id} deleted.`,
   }
 }
 
+export const bulkDelete = async (ctx: any) => {
+  const { userIds } = ctx.request.body
+  try {
+    let { groupsToModify, usersResponse } = await users.bulkDelete(userIds)
+    await groupUtils.bulkDeleteGroupUsers(groupsToModify)
```
Rory-Powell, line 131: I think it would be better that the group handling be done inside the users sdk, that's now an important part of the users logic. Should just be a move and can then simplify the response needed

packages/worker/src/sdk/users/users.ts
```diff
@@ -15,11 +15,12 @@ import {
   accounts,
   migrations,
 } from "@budibase/backend-core"
-import { MigrationType } from "@budibase/types"
+import { MigrationType, User } from "@budibase/types"
+import { groups as groupUtils } from "@budibase/pro/"
```
Rory-Powell, line 19: extra `/` on this import

packages/worker/src/sdk/users/users.ts
```diff
@@ -166,28 +199,21 @@ export const save = async (
     // save the user to db
     let response
     const putUserFn = () => {
-      return db.put(user)
+      return db.put(builtUser)
     }
-    if (eventHelpers.isAddingBuilder(user, dbUser)) {
+    console.log(builtUser)
```
Rory-Powell, line 204: console log

packages/backend-core/src/events/publishers/group.ts
Rory-Powell: cool cool. It's still worth giving this it's own event property definition as all other events do this, it's odd to have one that doesn't which could start a pattern

PClmnt:
Added tests and resolved the remaining comments (hopefully).

I have reduced the user object within the groups document to `{_id: string, email: string}`, I know this has implications later if we allow a user to update their email etc, but I would really like to get this PR in, even if I put a solution for that particular scenario in place before the licensing work goes out.

PR #6575, opened 2022-07-06
Properly invalidate the cached user

deanhannigan:
Authentication tokens were not properly being refreshed on the cached user when new ones were acquired. 

The user cache is now properly invalidated and updated with the correct values in the DB.

### DIFF
--- a/packages/server/src/threads/query.js
+++ b/packages/server/src/threads/query.js
@@ -8,6 +8,7 @@ const {
   refreshOAuthToken,
   updateUserOAuth,
 } = require("@budibase/backend-core/auth")
+const { user: userCache } = require("@budibase/backend-core/cache")
 const { getGlobalIDFromUserMetadataID } = require("../db/utils")
 
 const { isSQL } = require("../integrations/utils")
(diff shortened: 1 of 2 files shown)


packages/server/src/threads/query.js
```diff
@@ -189,6 +183,11 @@ class QueryRunner {
     if (!resp.error) {
       const globalUserId = getGlobalIDFromUserMetadataID(_id)
       await updateUserOAuth(globalUserId, resp)
+      this.ctx.user = await userCache.getUser(globalUserId)
+    } else {
+      // In this event the user may have oAuth issues that
+      // could require re-authenticating with their provider.
+      throw new Error("OAuth2 access token could not be refreshed")
```
Rory-Powell, line 190: Probably worth adding in the `resp.error` or `resp.error.message` (not sure if error is an `Error` or a string) here so that we have some insight into what went wrong with the provider

Rory-Powell (approved):
Just one comment about adding some context into provider errors, otherwise LGTM 👍

PR #6584, opened 2022-07-06
CLI backups functionality

mike12345567:
## Description
Overtime there has always been a need for people to be able to backup their entire self-hosted system and re-create it somewhere else, or over the top of an existing installation.

This provides a first pass attempt at this, using the following mechanism:
1. The CLI now has a LevelDB built into it, which it can utilise through PouchDB
2. The CLI can replicate a real running CouchDB system to its local levelDB - which then writes files to the file system.
3. We then iterate through all of the known/used buckets of Budibase in MinIO/S3 compliant file store and copy these to the local filesystem along with the data from Pouch.
4. Lastly, we tarball and gzip all of these files together to create the final backup file.

To import we essentially reverse these steps, expanding the tarball, uploading to MinIO and replicating from the local Pouch into the CouchDB system specified.

All of this is ran off of a `.env` file or a set of questions asked by the CLI if it cannot find an appropriate environment file to work from.

The size of the CLI has increased a bit due to these changes as it now needs to include prebuilds for levelDB.

## Screenshots
Exporting:
image

With questions:
image

Import:
image


shogunpurple (approved):
This is awesome - really cool to see the CLI start to stand on its own!

packages/cli/package.json
```diff
@@ -9,28 +9,43 @@
...
+      "node_modules/@budibase/backend-core/dist/**/*",
+      "prebuilds/**/*"
     ],
     "outputPath": "build"
   },
   "dependencies": {
+    "@budibase/backend-core": "^1.0.212-alpha.7",
     "axios": "^0.21.1",
     "chalk": "^4.1.0",
+    "cli-progress": "^3.11.2",
```
Rory-Powell, line 32: Given that we've pinned dependencies now in server, worker and backend-core, it could be worth pinning here too. it just solves so many problems around knowing exactly which versions are in use and keeping versions in sync across packages

Rory-Powell (approved):
Nice one!

hosting/docker-compose.dev.yaml
```diff
@@ -11,10 +11,11 @@ services:
       - minio_data:/data
     ports:
       - "${MINIO_PORT}:9000"
+      - "9001:9001"
     environment:
       MINIO_ACCESS_KEY: ${MINIO_ACCESS_KEY}
       MINIO_SECRET_KEY: ${MINIO_SECRET_KEY}
-    command: server /data
+    command: server /data --console-address ":9001"
```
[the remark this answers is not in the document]
mike12345567: In this case, I've just made the change to the dev version, so it shouldn't affect the production docker compose in anyway!

packages/cli/package.json
Rory-Powell: We do, but I've noticed cases where the same non-pinned version diverges across the lockfiles of backend packages. e.g. server and worker using different versions of koa middlewares. Pinning also adds some security improvements. A big example where this was useful was when we needed to purge the lockfiles for a CI problem when backend-core went to TS

PR #6596, opened 2022-07-07
Cypress: User Details Test Fix

Mitch-Budibase:
It seems that at times, Cypress is running so fast that it does not type entire words (it sometimes misses the first couple of letters.

I've put a small change in the userManagement test file to try and revolve this.
- Adding a wait and clear in association with the type command


MihailHadzhiev2022 (approved):
Fingers crossed :)

PR #6601, opened 2022-07-07
Helm chart: allow specifying image pull secrets

panos--:
## Description

Allow specifying image pull secrets on relevant deployments. This is relevant for example to avoid hitting rate limits on Docker Hub.


Rory-Powell (approved):
Thanks for the contribution @panos--

Issue #6538, opened 2022-07-04
Config validation does not trigger when creating a new datasource


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.214
  - App Version: 1.0.214

**Describe the bug**
The datasource config in the codebase includes a **required** property for each field, however it does not seem to do anything.

**To Reproduce**
1. Click add new datasource
2. Select for example Redis or Firestore
3. Clear out any default values and do not fill out any config fields.
4. Click save and continue
5. None of the required fields are highlighted in red.

**Expected behavior**
I should not be allowed to *Save and Continue* unless the required fields have been populated.

**Screenshots**
<img width="235" alt="Screenshot 2022-07-04 at 09 56 59" src="https://user-images.githubusercontent.com/101575380/177120162-c678701f-8128-44c2-8990-9643c2ed0d4a.png">

![Screenshot 2022-07-04 at 09 57 18](https://user-images.githubusercontent.com/101575380/177120233-ff283806-492c-4a44-a041-ee600645545c.png)

![Screenshot 2022-07-04 at 09 57 35](https://user-images.githubusercontent.com/101575380/177120278-2c71d4f6-e4d2-478a-802e-4e4b2fd8c12f.png)

PR #6602, opened 2022-07-07
Adds validation to datasource config modal

mslourens:
## Description
Add yup validation to the datasource config modal fields

Addresses: 
- #6538

## Screenshots
image

### DIFF
--- a/packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
+++ b/packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
@@ -10,10 +10,14 @@
   import KeyValueBuilder from "components/integration/KeyValueBuilder.svelte"
   import { capitalise } from "helpers"
   import { IntegrationTypes } from "constants/backend"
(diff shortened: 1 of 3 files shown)


melohagan:
Hey @mslourens 

I pulled this down and had a play around with it. This is a nice step in the right direction, but there's a couple of things I noticed. 

1. This should really just say *Port is a required field*
Screenshot 2022-09-12 at 10 27 35

2. Validation doesn't immediately kick-in on the configuration screen (after initial setup)
Screenshot 2022-09-12 at 10 28 35

3. Save button is still active when there are validation errors in the config screen
Screenshot 2022-09-12 at 10 30 34

mslourens:
Thanks for the review!
1. I am not sure how to fix this, because Yup will first test the type, and if it's empty it will try to parse null to a number, which it cannot do. It will throw this error before testing the required validation
2. The validation is only run when the config changes, but I can change it so it will run whenever the config is present on mount
3. This is probably caused by the weird number error. I hope it will be fixed when I found a solution for issue 1. I am open for suggestions.

melohagan:
Closed in favour of: https://github.com/Budibase/budibase/pull/8371

PR #6609, opened 2022-07-08
Fixed expression mutation

deanhannigan:
## Description
I found an issue where upon retrying a rest query after updating the accessToken, any expressions used to embed the token had already been mutated by the previous attempt. This fix ensures the process preserves the original.

Addresses: https://github.com/Budibase/budibase/issues/3827


Rory-Powell (approved):
LGTM

Issue #6611, opened 2022-07-08
Automations External Data Connector action not working


PClmnt:
@melohagan  discovered this - 

Output of the automation logs
![image](https://user-images.githubusercontent.com/5665926/177967208-4284819e-2c4b-48ca-b6ab-de5b1bdafda6.png)

The issue seems to be within this file: https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/controllers/query/index.ts#L139 the correct auth context doesn't exist when coming from an automation action.

PClmnt:
This is fixed, closing.

PR #6612, opened 2022-07-08
Fix Automation auth issue - External Data Connector not working

PClmnt:
## Description
Seems to be an auth issue introduced by the recent rest request changes.  

I'm not sure this is the most graceful way to handle this error, I have tested and it does fix the issue but i'm just trying to understand if we actually need the `getAuthConfig` call when running an automation?

Addresses: 
#6611

### DIFF
--- a/packages/server/src/api/controllers/query/index.ts
+++ b/packages/server/src/api/controllers/query/index.ts
@@ -166,12 +166,15 @@ export async function preview(ctx: any) {
   }
 }
 
-async function execute(ctx: any, opts = { rowsOnly: false, isAutomation: false }) {
+async function execute(
+  ctx: any,
+  opts = { rowsOnly: false, isAutomation: false }
+) {
   const db = getAppDB()
 
   const query = await db.get(ctx.params.queryId)
   const datasource = await db.get(query.datasourceId)
-  
+
   let authConfigCtx: any = {}
   if (!opts.isAutomation) {
     authConfigCtx = getAuthConfig(ctx)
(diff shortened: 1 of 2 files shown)


packages/server/src/api/controllers/query/index.ts
```diff
@@ -214,11 +219,11 @@ async function execute(ctx: any, opts = { rowsOnly: false }) {
 }
 
 export async function executeV1(ctx: any) {
-  return execute(ctx, { rowsOnly: true })
+  return execute(ctx, { rowsOnly: true, isAutomation: false })
 }
 
-export async function executeV2(ctx: any) {
-  return execute(ctx, { rowsOnly: false })
+export async function executeV2(ctx: any, isAutomation?: any) {
```
mike12345567, line 225: Preferably all boolean parameters should be wrapped in objects, options objects like `ctx: any, { isAutomation }: { isAutomation?: boolean } = {}`

Avoids situations with calls like `executeV2(ctx, false)` - difficult to see what this does vs `executeV2(ctx, { isAutomation: false })`

mike12345567 (approved):
LGTM, this is probably the best way to handle this anyway, other than adding it as a property to the `ctx` e.g. `ctx.automation = true`. I'd say this is a slightly larger refactor however, as I would wrap all controller calls from automations with a function that applies this parameter to the `ctx`, making it a consistent and dependable parameter that any controller could use.

deanhannigan:
Hey @PClmnt, yes at the minute the automations have no need for access to the authentication information. This is currently only used for rest requests and discerning which OIDC config to use when refreshing an access token.

deanhannigan (approved):
LGTM, great catch!

Issue #6633, opened 2022-07-10
Can't filter on query results when filtering for an integer column


danielbolt:
**Hosting**
- Self
  - Method: Docker
  - Budibase Version: 1.0.218
  - App Version: 1.0.218

**Describe the bug**
I've build a view within budibase which is on top of a SQL database. I'm using the data provider on top of this view. I'm able to pass the URL ID to the filter section in the data provider. The field i'm using for filtering is an integer. When clicking the link the data provider throws the error 'No rows found'.

Is budibase not able to filter data based on a view?

<img width="579" alt="image" src="https://user-images.githubusercontent.com/60649319/178145738-669f412b-e950-41c7-b523-56d0bb710c64.png">

I'm really sure the the 'IndelingenID' = 24 is available within the view.

I'm using this link to the next screen:

/overzicht/{{ New Repeater.VIEW_indelingen.indelingenID }}

mike12345567:
Hi @danielbolt - just to confirm a few details:
1. By view, did you create the view within your SQL database?
2. How are you accessing your view from Budibase, are you using a query or are you using the Budibase built in table generation? I believe that currently Budibase doesn't support importing views as tables.
3. If you are using a query, what happens if you remove the filter?

Currently, when filtering queries, this is carried out in memory rather than sending a `WHERE` statement to the view (due to how generic queries are, we cannot account for every possible filtering method). There could be an issue with this which needs to be diagnosed.

danielbolt:
> Hi @danielbolt - just to confirm a few details:
> 
> 1. By view, did you create the view within your SQL database?
> 2. How are you accessing your view from Budibase, are you using a query or are you using the Budibase built in table generation? I believe that currently Budibase doesn't support importing views as tables.
> 3. If you are using a query, what happens if you remove the filter?
> 
> Currently, when filtering queries, this is carried out in memory rather than sending a `WHERE` statement to the view (due to how generic queries are, we cannot account for every possible filtering method). There could be an issue with this which needs to be diagnosed.

1. I've added the query of the view within Budibase, i did not added the view to the SQL database.
2. See answer 1
3. If i remove the filter i see all the results of the view. I've noticed that i can filter the view based on string field, but not on a number field....

mike12345567:
Ah, this is likely something to do with type casting within the in-memory filtering, I'll have a look into this.

PR #6638, opened 2022-07-11
July release fixes

mike12345567:
## Description
Some general fixes for the release which has just gone out.

Addresses: 
- Issue with in memory filtering of queries #6633 
- Issue with saving email config #6681 
- Fix for ARM build - don't enable the Oracle libraries on ARM


shogunpurple (approved):
LGTM! Nice fixes - especially oracle.

Issue #6362, opened 2022-06-17
Truncated text appears in title attribute when reading from table relationship options


theetrain:
**Hosting**

- Self
  - Method: docker compose/swarm
  - Budibase Version: 1.0.196
  - App Version: 1.0.196

**Describe the bug**

When creating a new column for a Budibase DB table of type "relationship", and the table names are so long that they appear truncated, it turns out both the label copy and `title` attribute copy (the text that appear when hovering a radio option) are truncated. This is problematic for a number of reasons:

1. Sighted users cannot read the truncated `title` text
2. Screen reader users cannot read the truncated label
3. Screen magnifier users may experience difficulty reading `title` text ([see explanation of title attribute and its impact on screen readers](http://simplyaccessible.com/article/title-attributes/))

**To Reproduce**
1. Create two new Budibase DB tables with long names
2. Create a new column in one of the Budibase DB table
3. Select the "Relationship" type
4. Observe the radio options have visually-truncated text
5. Hover one of the options, observe the same truncated text

**Expected behavior**

Ideally, try to come up with a visual appearance that avoids truncating all text as it saves the effort in making full text visible only on hover.

Prefer CSS `text-overflow` over lodash `truncate` if absolutely necessary.

**Screenshots**

<img width="509" alt="screenshot of database create column dialog, illustrating illegible use of truncated text" src="https://user-images.githubusercontent.com/12798751/174214370-439f2547-99c0-402d-a3fa-97972372c4d9.png">

**Desktop (please complete the following information):**
 - OS: macOS 12.4
 - Browser: Chrome 102.0.5005.115 (Official Build) (arm64)

PR #6640, opened 2022-07-11
Pass a title property with untruncated text to the RadioGroup

mslourens:
## Description
Pass an untruncated string to the radiogroup, which can be used as title attribute for accessability.

Addresses: 
- #6362

## Screenshots
image


aptkingston (changes requested):
LGTM @mslourens, thanks!

There's just one other place we need to update, which is the client component where we also use the core radio group component directly. If we don't update that, then the titles for radio groups in client apps will be `[Object object]`.

All we need to do is add
```js
getOptionTitle={flatOptions ? x => x : x => x.label}
```
to `packages/client/src/componens/app/forms/OptionsField.svelte` at line 81.

Issue #6210, opened 2022-06-07
Improve button actions UI


AndyWhann:
# Problem

There are too many button actions for the current UI. 

There is no way to see how many actions are set for your button without clicking `Define actions`

# Solution

### Component settings panel

Add a section for actions that shows the number of actions set

### Actions panel

Open a new menu when clicking `Add Action`

Add the ability to go `Back` from the new menu

Add search option

Divide the actions into categories

![button-actions.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/42a5d062-78aa-41b9-a14a-19f073408195)

![button-actions-with-back-button.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/6a2409ee-ab01-4dbf-aea9-994a11873ffc)

PR #6676, opened 2022-07-13
Feature/app action UI updates

deanhannigan:
## Description
The button action UI has been updated to include:

- A search bar allowing the user to filter the action list by name. This will make the list easier to navigate as it continues to grow.
- A categorised list of button actions:
  - Form - relating to builder forms and behaviours like clearing and validation.
  - Data - relating to any query, datasource and data workflows in general.
  - Application - application operations like logging out and navigating to other screens.
  - Logic - options that influence the execution flow of button actions.

- In the component settings section, the number of actions set on a button is now displayed. This allows the user the gauge which buttons have actions defined at a glance.

Addresses:  https://github.com/Budibase/budibase/issues/6210

## Screenshots

**Filter actions by name**

Screenshot 2022-07-13 at 10 27 01

**Action count displayed with the button**

Screenshot 2022-07-13 at 12 39 53


aptkingston (approved):
LGTM! Nice work Dean.

Only 2 minor style things from the designs which I think are worth implementing:
- Slightly reduce padding below action category titles and slightly increase above titles, to better indicate a split between sections:
image
- Update actions to have backgrounds rather than borders:
image

PR #6685, opened 2022-07-13
Fix for ethereal test case failures

mike12345567:
## Description
Quick fix for realEmail test which was causing problems.

The saving of the email configuration also requires contacting ethereal to verify the settings, which was breaking.


shogunpurple (approved):
LGTM! Nice use of `race` as well 👍

PR #6687, opened 2022-07-13
Add version modal back into builder

aptkingston:
## Description
Adds the version modal back into the builder. This is a temporary measure until we properly address the app update UX. Currently app updates are very hidden, yet updating the client app version is an important step in enabling full use of the new design UI features.

Also adds a tooltip to this icon to be in line with the rest of the new design UI top icons.

Functionally this is identical to the version modal which was historically here.

## Screenshots
image


shogunpurple (approved):
LGTM!

PR #6688, opened 2022-07-13
Fix for single image build

mike12345567:
## Description
Fix for single image build, make sure predocker commands always ran.

Due to using buildx action now instead of running command from `package.json` we were missing some predocker steps.


package.json
```diff
@@ -65,7 +66,7 @@
     "build:digitalocean": "cd hosting/digitalocean && ./build.sh && cd -",
     "build:docker:single:multiarch": "docker buildx build --platform linux/arm64,linux/amd64 -f hosting/single/Dockerfile -t budibase:latest .",
     "build:docker:single:image": "docker build -f hosting/single/Dockerfile -t budibase:latest .",
-    "build:docker:single": "lerna run build && lerna run predocker && npm run build:docker:single:image",
+    "build:docker:single": "yarn build:docker:pre && npm run build:docker:single:image",
```
shogunpurple, line 69: NAB: small edge case but best to just use `npm run` instead of `yarn` in NPM scripts so they work on machines without `yarn`

shogunpurple (approved):
Great job catching this quickly 👍

Issue #5803, opened 2022-05-10
Usage quota doc and user documents have an overlap in ID


mike12345567:
**Describe the bug**
While running test cases locally I noticed in the logs the following:
![image](https://user-images.githubusercontent.com/4407001/167674575-ec910fe4-b8cb-4eba-91eb-0095a1be0956.png)

This appears to be an error coming from PouchDB, it is unable to process one of the documents which is being passed through to a view. This error comes from [this view](https://github.com/Budibase/budibase/blob/7172690bddc64df091c36b3d7fa7398d8f28d42a/packages/backend-core/src/db/views.js#L24) which looks up user IDs by their email address. In this view we have a check `if (doc._id.startsWith("${DocumentTypes.USER}")) { ...` to assert whether or not the document in the global DB is a user document and should therefore be emitted by the view.

The issue is that the `DocumentTypes.USER` is equal to `"us"` and the usage quotas document has an ID of `usage_quota` therefore it overlaps with the user documents and each time it is updated the view attempts to index it and fails.

This isn't critical as Couch will be able to handle this error, but it is likely causing some messages to be logged about it each time it occurs, it won't affect the contents of the view but it should likely be fixed.

As both of these things are in production we would have two options for this:
1. Change the view and come up with a good way to migrate all existing views - realistically the view should be searching for `us_` at the start instead. This is something that we've thought about in the past but haven't had to do yet, the process of migrating all views at once could be very difficult/stressful for the Couch services, it would probably be better to detect the difference when the view is used and fix it if it is not equal.
2. Change the usage quota document ID and migrate existing documents to use the same format, personally I'd rather fix the view as while this is a smaller migration, it doesn't fix the underlying issue. The one reason we might want to do this is to migrate all static documents to use a format like `_usage_quota` so that static documents which don't use the `<document type>_` prefix have their own prefix so that the issue can be avoided in future, as this issue could occur on other views as well, e.g. if you had a static document called `row_count` it would be indexed as a row.

**To Reproduce**
Run any of the test cases in the worker, e.g. `auth.spec.js` and you'll see the error being logged.

PR #6690, opened 2022-07-13
Fixing issue with user email view

mike12345567:
## Description
Adding a mechanism to deprecate a view and replace it with a new one. This issue has been present a while and doesn't actually present as any kind of direct bug, but it could be having an impact on performance at scale, as well as causing erroneous logs during testing and production environments.

Addresses: 
- #5803


NEOLPAR (approved):
LGTM

shogunpurple (approved):
LGTM!

Issue #6594, opened 2022-07-07
Replace PouchDB client library with standard couch calls


shogunpurple:
We've seen some memory issues with our usage of `pouchdb`. We need to replace this library with hand written couch client code.

**COS**
- Write a custom `pouchdb` module that makes direct couchdb calls rather than the abstractions provided by pouch
- Maintain same business logic, just change the core module and mirror the pouch API.

PR #6705, opened 2022-07-14
Memory leak fix

mike12345567:
## Description
This constitutes the work that @Rory-Powell put together to fix the context/clear up of DBs when they are no longer used. This should provide a good hold over until we can do the full work of removing PouchDB (#6594) and fixing the memory leak issue once and for all.

I've tried to simplify the context a little bit in the process as most of the problems that were left simply related to duplication and various functions not following the exact pattern they should be for context updating - however all of this should go away shortly so I tried to do it in as quick a way as possible.

I've left in some functionality at the very core (inside the DB opening/closing end) to allow counting the DBs which are currently open and closed. Left this in for now, it'll be stripped out as part of the larger rework.


packages/backend-core/src/context/index.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (!opts.existing) {
+      updateTenantId(tenantId)
+    }
+
+    try {
+      // invoke the task
+      return await task()
+    } finally {
+      await closeWithUsing(ContextKeys.TENANCY_IN_USE, () => {
+        return closeTenancy()
```
NEOLPAR, line 83: I think we should await here for closeTenancy, just in case it takes longer for any reason. does it make sense?

mike12345567: here this isn't an async function within, its just straight returning the `closeTenancy` function, so the promise will simply be returned for the parent (wrapper) to handle.

packages/backend-core/src/context/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  internal: (opts: { existing: boolean }) => Promise<any>
+) {
+  const using = cls.getFromContext(usingKey)
+  if (using && existing) {
+    cls.setOnContext(usingKey, using + 1)
+    return internal({ existing: true })
+  } else {
+    return cls.run(async () => {
+      cls.setOnContext(usingKey, 1)
+      return internal({ existing: false })
```
NEOLPAR, line 27: this and the internal call in line 23 are async function, should we await for them?

NEOLPAR (approved):
LGTM, I just added a couple of comments about async functions now being awaited, just to be aware of them, I am not sure if we could not wait for them safely

packages/backend-core/src/context/utils.ts
mike12345567: returning from an `async` function replaces the async promise return with the returned promise call, same functionality as with the other commented line!

packages/backend-core/src/context/index.ts
NEOLPAR: Oh I see! thanks for the details :)

packages/backend-core/src/context/utils.ts
mike12345567: Just to be clear, `async () => { return await promise() }` is exactly the same as `async () => { return promise() }` - all the `async` keyword does is enforce the function returns a promise, but if you return a promise specifically from an async function then it is automatically included in the main async function response. Its a quirk of the async definition which can be used for benefit sometimes, like in these situations where I don't want to handle any error scenarios from these functions myself.

shogunpurple (approved):
LGTM! Great job @Rory-Powell and @mike12345567. Also nice to know it's just temporary until we move away from pouch - this will be getting done next month.

packages/backend-core/src/db/index.js
```diff
@@ -1,10 +1,19 @@
 const pouch = require("./pouch")
 const env = require("../environment")
 
+const MEMORY_LEAK_CHECK = 0
+const openDbs = []
 let PouchDB
 let initialised = false
 const dbList = new Set()
 
+if (MEMORY_LEAK_CHECK) {
```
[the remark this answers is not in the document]
mike12345567: Sounds good, I thought about that as well!

PR #6724, opened 2022-07-15
only add account portal people to intercom

shogunpurple:
## Description
Found a better way to do this - that ensures that only account portal users will be shown the bubble.

Addresses: 
- ...more if required

## Screenshots

### DIFF
--- a/packages/builder/src/stores/portal/auth.js
+++ b/packages/builder/src/stores/portal/auth.js
@@ -54,23 +54,24 @@ export function createAuthStore() {
     })
 
     if (user) {
-      const adminStore = get(admin)
       analytics
         .activate()
         .then(() => {
           analytics.identify(user._id)
-          analytics.showChat({
-            email: user.email,
-            created_at: (user.createdAt || Date.now()) / 1000,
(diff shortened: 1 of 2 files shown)


packages/builder/src/stores/portal/auth.js
```diff
@@ -54,23 +54,24 @@ export function createAuthStore() {
...
+              created_at: (user.createdAt || Date.now()) / 1000,
+              name: user.account?.name,
+              user_id: user._id,
+              tenant: user.tenantId,
+              admin: user?.admin?.global,
+              builder: user?.builder?.global,
+              "Company size": user.account?.size,
+              "Job role": user.account?.profession,
+            },
+            !!user?.account
```
Rory-Powell, line 73: Know this is already merged, but one thing I wanted to point out of the potential for future non-account holders to have the account field populated. 

The safer option might be to check that the `accountPortalAccess` field is set which is added to the session for account holders. Even further protection could be compare the `email` to `account.email`, or this could be done in the backend and enriched as a field `accountHolder`.

Issue #6715, opened 2022-07-14
Single Docker Launch Issue


alec-hs:
**Hosting**

- Self
  - Method: docker single image
  - Budibase Version: latest

**Describe the bug**
When launching using single docker conatiner mode on portainer additional volumes are created and app does not launch. 

**To Reproduce**
1. Create Docker Stack in Portainer (example compose file below)
2. Launch the Stack and wait for the container to attemp to launch
3. Observer additional volumes created

***Example Compose File***
```
version: "3.8"

networks:
  kn-core:
  default:
    name: kn-core

volumes:
  budibase:
    name: budibase
    driver_opts:
      type: "nfs"
      o: ${NFS_TARGET}
      device: "${NFS_DOCKER}/budibase"

services:
  budibase:
    container_name: "budibase-kn"
    image: budibase/budibase:latest
    ports:
      - 80
    restart: unless-stopped
    volumes:
      - budibase:/data
    environment:
      - JWT_SECRET=${BUDIBASE_SECRET}
      - MINIO_ACCESS_KEY=${BUDIBASE_SECRET}
      - MINIO_SECRET_KEY=${BUDIBASE_SECRET}
      - REDIS_PASSWORD=${BUDIBASE_SECRET}
      - COUCHDB_USER=admin
      - COUCHDB_PASSWORD=${BUDIBASE_SECRET}
      - INTERNAL_API_KEY=${BUDIBASE_SECRET}
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.budibase.entrypoints=https"
      - "traefik.http.routers.budibase.rule=Host(`budibase.kestrel.network`)"
      - "traefik.http.routers.budibase.tls.certresolver=cloudflare"
```

**Expected behavior**
No additional volumes to be created and for app to launch. 

**Screenshots**
Additional Volumes
![image](https://user-images.githubusercontent.com/6503411/179102306-0ac18b60-6b05-4d18-8afe-bf34a174f176.png)

Contianer Health Issue
![image](https://user-images.githubusercontent.com/6503411/179102809-52d3eb2d-4eb5-42eb-bfc4-758d2c9bb420.png)

**Server**
Debian 11, latest version of Docker and Portainer Business Edition

**Additional Info**
Logs - [_budibase-kn_logs.txt](https://github.com/Budibase/budibase/files/9116016/_budibase-kn_logs.txt)

mike12345567:
[...]

PR #6725, opened 2022-07-15
Quick fix for single image volumes

mike12345567:
## Description
Removing the volume tags from the old directories, instead focusing on the single data directory that is now used - #6715.

Previously we had two separate volumes that had to be mounted, I merged these into a single, simple directory to make use of the image easier, but I hadn't removed the old volume references that things like portainer picks up.


shogunpurple (approved):
LGTM

PR #6731, opened 2022-07-15
Frontend store improvements for conflicts and performance

aptkingston:
## Description
This PR refactors most of the frontend store methods to apply a fake "patch" to screen docs, in an attempt to reduce the frequency of 409 conflicts.

TLDR:
- Hopefully far fewer 409s
- Better performance
- Bugfix for cutting and pasting
- Fixes for messy cases of updating state via derived stores rather than via real source of truth store

~Targeting develop as changing the core frontend store is slightly risky in general, and some manual testing would be good.~
Maybe I'll rebase to master to get this out sooner as develop might not be merged for a while.

### 409s
When designing an app, almost all changes saved (component settings, screen settings) are stored in screen documents. This means that when we change a setting for a component, the full screen definition is sent up and saved. Depending on the frequency of these updates, and the latency between client and server (and performance of the server), multiple successive calls could be made before the first one had responded - leading to a stale revision ID in the screen document being sent up in subsequent in-flght calls.

This refactor creates a core "patch" function for saving screen documents. This function is queued and executed sequentially, and is the underlying mechanism now for all screen and component changes. A patch function is supplied as a parameter, which is applied to the latest version of the screen doc. As the screen docs are only pulled at the moment of execution, and all executions are queued and executed sequentiually, we can always be sure that the version of the screen doc being used is the latest. We take this latest version of the screen doc, clone it and then allow full mutation by the patch function before saving this to the server.

Going forward, the new `store.actions.screens.patch` and `store.actions.components.path` functions should always be used to acutally mutate any component or screen settings.

In my testing I've been unable to reproduce any 409 conflicts, even with introducing massive latency and queuing up 10+ actions. I've also added more protection around every store method - so if you do things like queue a deletion and then queue settings changes to the deleted component, no errors will be throw and it will fail gracefully.

### Performance
I've also combined store updates where possible to reduce unecessary recomputation of derived state, and UI paints.

Going forward we would benefit heavily from breaking this monolith store into multiple smaller stores, as Svelte is not very clever at checking for store changes and updating any property of a store triggers a full recomputation of derived state. Breaking our store into smaller chunks of related state will allow us to optimise this much further.

### Cutting and pasting
Our old version of cutting and pasting was extremely fragile. We were actually removing the cut component from the `selectedScreen` store - so if we had ever used the `selectedScreen` derived store to save a screen doc change while a cut was in progress, we would actually have deleted the component being cut.

I've rewrote this so that cutting a component hides it from the UI entirely as if it was removed, but does not actually remove it from the definition until it is pasted.

### DIFF
--- a/packages/builder/src/builderStore/index.js
+++ b/packages/builder/src/builderStore/index.js
@@ -2,7 +2,6 @@ import { getFrontendStore } from "./store/frontend"
 import { getAutomationStore } from "./store/automation"
 import { getThemeStore } from "./store/theme"
 import { derived } from "svelte/store"
-import { LAYOUT_NAMES } from "../constants"
 import { findComponent, findComponentPath } from "./componentUtils"
 import { RoleUtils } from "@budibase/frontend-core"
 
@@ -27,6 +26,10 @@ export const selectedComponent = derived(
     return findComponent($selectedScreen?.props, $store.selectedComponentId)
   }
 )
+
+// For legacy compatibility only, but with the new design UI this is just
+// the selected screen
+export const currentAsset = selectedScreen
 
 export const sortedScreens = derived(store, $store => {
   return $store.screens.slice().sort((a, b) => {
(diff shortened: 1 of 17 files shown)


aptkingston:
Moved to master in https://github.com/Budibase/budibase/pull/6733.

Issue #6729, opened 2022-07-15
Selected form step not showing


laromana:
**Hosting**
- Self
  - Method: docker single image
  - Budibase Version: 1.1.4

**Describe the bug**
Selected form step not showing in the design area unless page is refresh.

**To Reproduce**
1. Go to design
2. Select screen
3. Open component panel
4. Select form>from step number
5. Select different step number

**Expected behavior**
Selected form step to show within the design are

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: chrome
 - Version 103.0.5060.114 (Official Build) (64-bit)

aptkingston:
Hi @laromana. Could you check if you're seeing any errors in your browser console when this behaviour happens? Probably something that looks like:
![image](https://user-images.githubusercontent.com/9075550/179224525-8e86064b-f5ea-4060-aa65-0cdc8a6bdee0.png)

laromana:
Yes I am getting the error below:

![budibase_error](https://user-images.githubusercontent.com/79430611/179230045-4578dd40-b8ac-4233-8191-adf586e3c5b2.JPG)

192.168.1.196/:1 Unchecked runtime.lastError: The message port closed before a response was received.
budibase-client.js?v=1.1.4:243 Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'scrollMargin')
    at budibase-client.js?v=1.1.4:243:303415
    at e.$$.update (budibase-client.js?v=1.1.4:243:303493)
    at _e (budibase-client.js?v=1.1.4:1:6391)
    at Le (budibase-client.js?v=1.1.4:1:6141)
(anonymous) @ budibase-client.js?v=1.1.4:243
e.$$.update @ budibase-client.js?v=1.1.4:243
_e @ budibase-client.js?v=1.1.4:1
Le @ budibase-client.js?v=1.1.4:1
Promise.then (async)
we @ budibase-client.js?v=1.1.4:1
(anonymous) @ budibase-client.js?v=1.1.4:1
(anonymous) @ budibase-client.js?v=1.1.4:1
(anonymous) @ budibase-client.js?v=1.1.4:243
i @ budibase-client.js?v=1.1.4:1
set @ budibase-client.js?v=1.1.4:243
window.loadBudibase @ budibase-client.js?v=1.1.4:243
receiveMessage @ VM985 about:srcdoc:86
postMessage (async)
I @ index.91ff43ed.js:904
[...]

PR #6735, opened 2022-07-15
Replace incorrect usages of element.childNodes with element.children

aptkingston:
## Description
This PR replaces a few usages of `element.childNodes` with `element.children` throughout the builder and client library. This was just an error as I didn't fully appreciate the differences between an HTML `node` and an `element`. In all cases, we actually just want elements, which is what the array of `children` provides.

Addresses: 
- #6729 

May have been other random crashes while using drag and drop which this will fix.


aptkingston:
Going to merge this in because it's small, safe and fixes some bugs that are live.

Issue #6709, opened 2022-07-14
No longer able to trigger automations from Public screens (Session not authenticated)


bmcbride:
**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 1.1.4
  - App Version: 1.1.4

**Describe the bug**
I have a Public form that triggers an email automation when submitted. After upgrading from version 1.0.196 to 1.1.4 the automation no longer runs and a "Session not authenticated" warning is displayed. 

**To Reproduce**
1. Create a Public screen with a button to trigger an automation.
2. Try triggering an automation while logged out of Budibase.

**Expected behavior**
Previously automations could be triggered from Public screens.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: 103.0.5060.53

bmcbride:
I was able to reproduce the `Session not authenticated` error in my hosted Budibase account on version 1.1.7. Is it no longer possible to trigger automations from Public screens or is this a regression? Thanks!

bmcbride:
@mike12345567 thanks for working on addressing this! Any idea when these fixes might make it into a release? Thanks again, much appreciated!

mike12345567:
Hi @bmcbride - this should be released early next week, with the change you'll need to configure your app actions as public, you can see an example of this done in the PR!

bmcbride:
Sounds good. Thanks again and have a great weekend.

Issue #6702, opened 2022-07-14
Remove user table as an option for automation triggers and create row action


Elliot-Construct:
When I set an automation on the User table, it doesn't trigger when you activate a user on an application.

Any thoughts?

mike12345567:
Hi @Elliot-Construct - this isn't really a supported path for automations, realistically we need to hide the users table as an option for automation row created/deleted triggers. The users table is quite different to other tables under the hood and doesn't really play very well with automations, other than being read, or updating custom columns on an existing user row.

Issue #6501, opened 2022-06-29
Error publishing app - status 500 when using CRON Trigger


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6459

<div type='discussions-op-text'>

<sup>Originally posted by **applicationuser** June 24, 2022</sup>
Error publishing app 

![image](https://user-images.githubusercontent.com/42922833/175696707-eb47d60c-683e-416f-825f-08224356ac30.png)
</div>

https://github.com/Budibase/budibase/discussions/6459#discussioncomment-3042440

mike12345567:
I have been unable to recreate this issue in the latest releases, might be worth re-testing.

melohagan:
Some further information here: https://github.com/Budibase/budibase/discussions/6459#discussioncomment-3154616

I'll test it out when I get a chance.

mike12345567:
@melohagan - I was able to recreate, have added a fix for it.

PR #6739, opened 2022-07-15
Various automation fixes

mike12345567:
## Description
Some fixes for automations as have been raised recently.

Addresses: 
- https://github.com/Budibase/budibase/issues/6702 - the users table cannot be used for triggers due to how it works, remove it as an option.
- https://github.com/Budibase/budibase/issues/6501 - empty CRON expressions broke app deployments, this makes it so that there is an error and deployments will still go through, but no CRON will ever fire.
- https://github.com/Budibase/budibase/issues/6709 - after some fixes to roles, it wasn't possible trigger automations from a screen when it was a public user/screen. This makes it so that you can make an automation accessible from public.

## Screenshots
Role selection on app action:
image

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowChart.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowChart.svelte
@@ -12,6 +12,7 @@
     notifications,
     Modal,
   } from "@budibase/bbui"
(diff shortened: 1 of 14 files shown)


shogunpurple (approved):
LGTM! Not one for now but we should cover the app action case with a cypress/Newman test

PR #6787, opened 2022-07-18
Create App Tests: Timeout Changes

Mitch-Budibase:
This test spec file fires off against every PR. There is some timing issues currently.

I am making a few changes in relation to timeout values (a small restructure too) in hopes this helps with the PR builds.

Tests pass locally for me, so there may be more changes required against this spec file as I check its performance


MihailHadzhiev2022:
Good Job Mitch !

Issue #6244, opened 2022-06-09
Cards - Make full component clickable 


AndyWhann:
When using cards, I expect the full card to be clickable when I have added a link. In adobe spectrum it also suggests that the full card is clickable.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/88fdcaf5-53c6-4eb2-bbdc-aef3cc8fadb3)

Shadowfita:
Yes please.

PR #6788, opened 2022-07-18
Allow full spectrum cards to be clickable

aptkingston:
## Description
This PR allows for cards to be entirely clickable. You can define a click action for cards and control whether or not the full card is used (default) or a button is used (using the `Use button for click action` setting).

- Both cards and card blocks support being fully clickable
- Button text setting only shown if using a button
- Compatible with title link setting (title link takes priority if both are in use)
- Hover state added to clickable cards

Addresses: 
- #6244 

## Screenshots
New settings labels:
image

Full card hover state:
image

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -3509,17 +3509,18 @@
       },
       {
         "type": "boolean",
-        "label": "Show button",
+        "label": "Use button for click action",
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM! Nice wee update

PR #6794, opened 2022-07-18
Fix for automations and context stack up

mike12345567:
## Description
Fixing an issue with automations being unable to access app databases due to new context stack up.

This is only a very temporary fix, it is difficult to come up with a proper fix without an actual refactor due to the complexity of the core - in this case I've just removed the cause of the problem.

### DIFF
--- a/packages/backend-core/src/context/index.ts
+++ b/packages/backend-core/src/context/index.ts
@@ -67,6 +67,7 @@ export const getTenantIDFromAppID = (appId: string) => {
 
 // used for automations, API endpoints should always be in context already
 export const doInTenant = (tenantId: string | null, task: any) => {
+  tenantId = tenantId || DEFAULT_TENANT_ID
   // the internal function is so that we can re-use an existing
   // context - don't want to close DB on a parent context
   async function internal(opts = { existing: false }) {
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
LGTM!

packages/backend-core/src/context/index.ts
```diff
@@ -67,6 +67,7 @@ export const getTenantIDFromAppID = (appId: string) => {
 
 // used for automations, API endpoints should always be in context already
 export const doInTenant = (tenantId: string | null, task: any) => {
+  tenantId = tenantId || DEFAULT_TENANT_ID
```
Rory-Powell, line 70: I'd be very wary of shipping this, if there is every a scenario where we supply a null or undefined tenant id (which really should be addressed) we will silently go to the default db instead. In a single tenant environment this is ok, but in cloud this could cause havoc.

packages/server/src/threads/automation.js
```diff
@@ -264,14 +261,11 @@ class Orchestrator {
...
           // appId is always passed
-          let tenantId = app.tenantId || DEFAULT_TENANT_ID
-          const outputs = await doInTenant(tenantId, () => {
-            return stepFn({
-              inputs: inputs,
-              appId: this._appId,
-              emitter: this._emitter,
-              context: this._context,
-            })
+          const outputs = await stepFn({
```
Rory-Powell, line 264: It's safe to remove the `doInTenant` as we're already doing a `doInAppContext` futher up the chain?

Issue #6795, opened 2022-07-18
Single Image healthcheck fails if REDIS_PASSWORD is not set


alphafalcon:
**Hosting**
- Self
  - Method:  docker single image

**Describe the bug**
When environment vars are not set in the compose file or on the docker command line, the healthcheck script fails because the REDIS_PASSWORD variable is not set in the script.

in a swarm environment this leads to constantly restarting the container.

Adding the snippet
`if [ -f "/data/.env" ]; then
    export $(cat /data/.env | xargs)
fi`
from the runner.sh script at the beginning of healthcheck.sh loads the variables and fixes this.

mike12345567:
Hi @alphafalcon - this issue should be resolved in the latest versions of the single image.

PR #6812, opened 2022-07-19
Quick healthcheck update for single image

mike12345567:
## Description
Moving the single image healthcheck script to be within the single image directory and updating it to look for the environment variable file.

There was an issue where the environment variables wouldn't be loaded before the healthcheck script runs, meaning that the state of Redis couldn't be confirmed (without password).

Addresses: 
- #6795


jonnymccullagh (approved):
lgtm

Issue #5995, opened 2022-05-20
[BUDI-6904] Dynamic Filter - CONTAINS Option


Hajelarohit:
**Describe the feature request**
Current Dynamic Filter doesn't have the capability to search partial strings. 
**Screenshots**
My Data contains multiple first and last names of Vendor Managers. 
![image](https://user-images.githubusercontent.com/104535363/169449580-56a487fb-85b9-487e-abc5-af5c3feb1249.png)

The filter should be able to search with only first name of last name or even partial strings. I tried Like, but it doesn't work for this.
![image](https://user-images.githubusercontent.com/104535363/169449373-c4e94c38-35c7-44b5-8b6c-be374adf7e4e.png)

<sub>[BUDI-6904](https://linear.app/budibase/issue/BUDI-6904/dynamic-filter-contains-option)</sub>

JeroenvdBurg:
Definitly a +1 for this enhancement. FYI: If you use a SQL or Postgres connection the Like operator is behaving as a Contains query.

`
 if (this.client === SqlClients.POSTGRES) {
          query = query[fnc](key, "ilike", ``%${value}%``)
`

JeroenvdBurg:
@mike12345567 any thoughts on this one ?

aricooperdavis:
I don't think this should have been marked as completed - the referenced issue is for multi-select types only, not string types as this issue raises.

aricooperdavis:
@shogunpurple I'm still not seeing a `contains` option for filtering string type fields in 2.3.0. Can you confirm that this has been completed for `string` type fields, not just multiselect?

jeanole:
would be highly appreciated!!

Issue #5585, opened 2022-04-26
Improved filtering


joebudi:
**Describe the feature request**
On Youtube, someone asked:
> Quick question (which is a deal breaker for me), If I had a list of videos (for example), can I add tags to them. (480p, catvid, etc), and then search. Is the search limited to one field, or can I do 'name like XXX' with the tags X,Y ? Do you support tagging and search based on the tags along with other fields?

My answer:

> Hi. If you have a list of tags, you can use the dynamic filter component to solve your query: https://docs.budibase.com/docs/adding-a-search-field#3-using-a-table-and-dynamic-filter---the-filtersearch-hybrid-approach

I attempted to provide alternative solutions specifically for custom tags, but when testing BB I couldn't:
- filter by relationship(sep tags table)
- filter by json field type

I imagine this will be popular (filter by tags, categories, etc), and is something we should support IMO. Apologies if we already support this and I wasted your time. @melohagan and I could not find a solution.

I actually think a tags field type might be nice - it would marry up with the on-change event nicely.

HeronCS:
We're migrating away from budibase for this sole reason of not being able to filter by relationships.

melohagan:
![Screenshot 2022-07-15 at 11 36 51](https://user-images.githubusercontent.com/101575380/179207225-b5d147dc-01fc-4d9d-acd4-464283b2eb5f.png)

I noticed you can add a json array to the Dynamic Filter, but I'm assuming this is a bug.

The new release has added the 'Is in' filter but this is the inverse of what's needed here. The code should hopefully be reusable though, at least for the filter by json array type. 

One 'Is in' Many 👍 **Done**
Many 'Contains' One 🤌 **Needed**

Grayda:
+1 for this, as I just discovered that if I create a Dynamic Filter component in Design, it won't let me filter by my relationship field. I was hoping to use a one-to-many relationship as a tags field, then filter the table to show all rows that have Tag 1 AND Tag 2 (or Tag 1 OR Tag 2, or whatever) 

Some additional thoughts on this: When I was building apps with Laravel Nova, I was using a third party tags package. It had a few different modes, including:

 * No new tags could be added. If the tag was not in the list, you couldn't add it. You could either click into the field to pick a tag, or start to type one, and just press comma to confirm
 * New tags could be added by just typing them in, but they wouldn't get saved anywhere, so if you wanted to use that tag again, you'd need to retype it every time.
[...]

PR #6815, opened 2022-07-19
Contains filter for Multi-select types

melohagan:
## Description
You can now add a filter for Multi-select data types to see if they contain a single value.
I had a look at making the contains work for string types, but couchdb lucene queries don't support substring searches. There is a *startsWith* filter which somewhat works for this case.

**Next Step**
- Add JSON.array column types to the list of filter drawer field options. This contains filter should work for it.
- Could this potentially open up filter by relationship for internal tables?

Addresses: 
- https://github.com/Budibase/budibase/issues/5995
- https://github.com/Budibase/budibase/issues/5585

## Screenshots
**Internal DB**
internaldb_multiselect_filter

**External DBs (SQL Server, MySQL, Postgres)**
externaldb_multiselect_filter

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.15-alpha.3",
+  "version": "1.1.18-alpha.0",
   "npmClient": "yarn",
   "packages": [
     "packages/*"
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@budibase/backend-core",
-  "version": "1.1.15-alpha.3",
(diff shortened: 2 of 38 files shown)


mike12345567 (approved):
LGTM! The only change I would make is perhaps adding a test case to the https://github.com/Budibase/budibase/blob/develop/packages/server/src/integrations/tests/sql.spec.js test spec, just to check if an SQL filter with a contains generates something you expect, helps maintain this down the line!

Might want to test both the MySQL and the Postgres cases.

melohagan:
@mike12345567 Unit tests added for MS-SQL, MySQL and Postgres.

mike12345567:
Thanks @melohagan !

Issue #6820, opened 2022-07-20
Remove Validate only current step checkbox


melohagan:
**Describe the bug**
The 'Validate only current step' checkbox should be removed. This is because in the latest versions it is always checked.
It should be removed to avoid confusion.

**Expected behavior**
As a user, I should not see the 'Validate only current step' checkbox, because it is always enabled now. 

**Screenshots**
![Screenshot 2022-07-20 at 09 10 14](https://user-images.githubusercontent.com/101575380/179931328-c137cab5-433d-4345-87dc-b680d65df7d4.png)

Issue #6356, opened 2022-06-16
Make it clear that required data table fields have not been provided when using Save Row action


melohagan:
**Describe the feature request**
Related to this discussion: https://github.com/Budibase/budibase/discussions/6348#discussioncomment-2964293

I've come across this myself, alongside a couple of other users. 

1. Add a table with columns to the **Data** tab and make one of the fields *Required*
2. Add a screen with a form
3. Add a Save button with a **Save Row** action
4. Click save without filling out the required field
5. See the *Bad Request* error message

You have to look inside the Network tab to see the validation error. This validation error should be displayed to the user.

**Screenshots**
![Screenshot 2022-06-16 at 14 39 18](https://user-images.githubusercontent.com/101575380/174082631-ff4b75fc-47e0-45fb-9b4a-2aa302e0a0d4.png)
*Not helpful/clear*

![Screenshot 2022-06-16 at 14 40 00](https://user-images.githubusercontent.com/101575380/174082746-1d6c078a-31d2-409d-800d-5970f2ddf057.png)
*Display these validation errors in the message*

aptkingston:
Did you add the `Validate form` action? If you don't perform a validation step then the button actions will always just send up whatever values are in the fields. `Validate form` exists to prevent submitting fields when there are errors, so I would always recommend adding validation to your form if you're saving rows from it.

melohagan:
I definitely agree, this is more to tell the developer to do that - so that you're aware of what the problem is immediately. 
Or maybe they have a required field on the DB table that was accidentally set, etc.

aptkingston:
The client displays anything in the `message` property of error responses - so if we stringified those errors and returned them in `message` instead of `Bad Request` then the client will display them. That's probably the easiest way to do this.

PR #6828, opened 2022-07-20
Remove validate current step checkbox and show validation error messages

melohagan:
## Description
1. Remove validate current form step validation (it's always checked)
2. Display the validation error messages instead of generic 'Bad Request'

Addresses: 
- https://github.com/Budibase/budibase/issues/6820
- https://github.com/Budibase/budibase/issues/6356

## Screenshots
**Validation checkbox removed**
Screenshot 2022-07-20 at 13 58 46

**Validation error messages**
Screenshot 2022-07-20 at 13 59 29


melohagan:
@aptkingston I thought the single giant error notification would look weird because the toasts are designed to be single lines. It also encourages the user to acknowledge each error. 

These errors are also specific to validation, so ideally only the app developer should be seeing them, and then fix it by adding the validation action to their form so that invalid submissions can't take place. 

Maybe I could combine the old and new code. So if it's a validation error then stack. If it's any other type of error then clear the stack - as was the case before. 

Happy to change it in whatever way suits!

aptkingston:
No need to change! Happy to leave it as is especially since you'd already considered that option :+1:

packages/client/src/components/overlay/NotificationDisplay.svelte
```diff
@@ -39,7 +41,8 @@
     display: flex;
     flex-direction: column;
     justify-content: flex-start;
-    align-items: center;
+    align-items: end;
```
aptkingston, line 44: After testing it out myself, this is the only thing I think we need to change. Aligning notifications to the right looks strange to me, as it even means single notifications aren't actually centered in the screen. I think we should change this back to `center`.

melohagan: Changed back to center align with fixed toast widths.

packages/client/src/stores/notification.js
```diff
@@ -37,26 +39,32 @@ const createNotificationStore = () => {
...
+          dismissable: !autoDismiss,
+          delay: get(store) != null,
+        },
+      ]
     })
-    clearTimeout(timeout)
     if (autoDismiss) {
-      timeout = setTimeout(() => {
-        store.set(null)
+      const timeoutId = setTimeout(() => {
```
aptkingston, line 57: As above, we don't need to worry about storing a reference. We can just call `setTimeout`.

packages/client/src/stores/notification.js
```diff
@@ -37,26 +39,32 @@ const createNotificationStore = () => {
...
+        dismiss(_id)
       }, NOTIFICATION_TIMEOUT)
+      timeoutIds.add(timeoutId)
```
aptkingston, line 60: As above, we can just remove this line entirely.

aptkingston (commented):
Tested the changes myself there Mel, and while I really like the fixed width notifications, I experienced some problems due to the use of `timeoutIds` and the store subscription callback. Removing all of that logic both tidies up the logic and fixes the problems thankfully!

melohagan:
That's what I get for copy/pasting - thanks for having look!

I've removed the timeoutIds and tested it. Seems to work from what I can see.

aptkingston (approved):
LGTM!

PR #6845, opened 2022-07-21
Fixing issue with app DB not being in context for fetch.

mike12345567:
## Description
Fix for public applications API, the single app read was failing with a `DB closed` error due to some issues with context, switching over to using a better method of configuring the context.

Raised by @MihailHadzhiev2022.


shogunpurple (approved):
LGTM, going to merge and release this one as it could affect quite a few users.

Issue #6842, opened 2022-07-21
Cannot drop on the root "Screen" component


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
Whenever I drag a form into a container, I cannot drag the form out of the container.
This also applies to two containers, etc.

**To Reproduce**
1. Add new screen
2. Add form component
3. Add container component to screen
4. Drag the form into the container
5. Try to drag the form component into the screen. It wont allow you.

**Expected behavior**
I should be able to unnest a form (or any other component) from a container. 

**Screenshots**
![container-bug](https://user-images.githubusercontent.com/101575380/180236011-801b601e-476d-4ddd-8607-cfb0ba9c8a4f.gif)

aptkingston:
What browser and OS are you on? This works as expected for me. Might be specific to some environment.

aptkingston:
Ah I see what it is - hover slightly further down, like around where the line is, and it works. It does nothing if you hover higher like on top of the word "Screen". I'll have a look!

Issue #6833, opened 2022-07-20
Enabling a search column on a date field greys out the search box


aidenmitchell:
**Hosting**
- Cloud
  - Tenant ID: bspit

**Describe the bug**
Enabling a search column on a date field greys out the search box, rendering it unusable.

**To Reproduce**
1. Create a DB with dates in it
2. Create a table with data from the DB, and add a search column on the date field.

**Expected behavior**
Search box is not greyed out.

**Screenshots**
![image](https://user-images.githubusercontent.com/30846409/180094516-80db44c0-d95f-44e5-9580-79a974217071.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 103.0.5060.114 (Official Build) (arm64)

aidenmitchell:
@aptkingston When will this be pushed out to the cloud tenants? Thank you!

shogunpurple:
@aidenmitchell The release goes out on monday :)

Issue #6142, opened 2022-05-31
Table filter not functional for date


purplepatrick:
**Hosting**

- Self
  - Method: <method> 
  - Budibase Version: <version> 
  - App Version: <version> 
- Cloud
  - Tenant ID: <tenantId> 

**Describe the bug**
When trying to apply any date filter, results are always empty

**To Reproduce**
1. Filter a date table header by date

**Expected behavior**
Filtering should filter results based on the chosen parameter

**Desktop (please complete the following information):**

Issue #6141, opened 2022-05-31
Table filter forces timestamp


pjfrey:
**Hosting**

- Self
  - Method: <method> 
  - Budibase Version: <version> 
  - App Version: <version> 
- Cloud
  - Tenant ID: <tenantId> 

**Describe the bug**
When filtering a table, the only option to do this by time is using a date-time combination.  There is no feature to remove the time from a date filter.

**To Reproduce**
1. Filter any table header by time

**Expected behavior**
Users should be able to filter by date independent of time. In other words, time and date should be separate filter option.

**Desktop (please complete the following information):**

PR #6858, opened 2022-07-22
Drag and drop fix + date filtering improvement

aptkingston:
## Description
This PR is a couple of small improvements.

### Drag and drop update update
Allow dropping on root "Screen" component in the component tree so that drag and drop works as expected when targeting this node. The root screen can't be dragged though, of course.

### Make searching on date fields in blocks useful
Previously, when selecting a date field as a search column in a block it was not particularly useful. This was due to us always doing an exact match on the value specified, meaning that only other dates which matched the exact millisecond selected would be returned.

I've updated this to actually do a range search for dates, spanning the entirety of the selected day. This means that any dates falling within the range of the date you pick will be matched. This is how most users would expect this to work, and actually makes it useful.

Here's the date filter working and returning rows with different times on a given day:
image

This works for both table block and cards block. I've refactored some of the core utilities used in these blocks to avoid duplication. 

Addresses: 
- #6842
- #6142, #6141, #5548, #3612,  
- Autocolumns are now also editable when used as search fields #6833


shogunpurple (approved):
LGTM

PR #6867, opened 2022-07-22
Export issues and fix for automation app action permissions spamming

mike12345567:
## Description
This addresses two issues with the export functionality:
1. Each export carries deleted documents with it as well, which affects the overall size of the export.
2. The user didn't have a choice when working with the a published app to export the dev version.

Also fixes an issue pointed out by @NEOLPAR about permission spamming in the automation UI, flow items weren't correctly making the decision to only process roles/permissions for app actions.

## Screenshot
Updated menu with option to export published (only shown when app is published)
image

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItem.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItem.svelte
@@ -52,8 +52,9 @@
       x => x.blockToLoop === block.id
     )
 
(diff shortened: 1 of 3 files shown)


Rory-Powell (approved):

Rory-Powell (commented):
Might be nice to have consistent length labels for exports like
- Export current (or 'latest')
- Export published

just to make it more clear what the current `Export` does.

shogunpurple (approved):

Issue #6691, opened 2022-07-13
Document Update Conflict - Issue after latest update


JayP718:
Issue after latest update. Random error where when I refresh the page it states document update conflict. 

Some previous threads state that this is due to nested data providers. However, this only happened recently after the latest update.

![image](https://user-images.githubusercontent.com/42209052/178846315-6eab77d0-b497-47d1-b047-ede5302cf83c.png)

aptkingston:
Hi @JayP718. Does this happen just when you're using the builder, or when you're actually using your app as well? Are you saving data as soon as the page loads?

poirazis:
Hey @aptkingston 
I can confirm it happens while designing.

JayP718:
Hi @aptkingston,

Same happens in both. We aren't saving anything as soon as the page loads.

@poirazis Glad it's not just me. Thanks for chiming in.

JayP718:
Does this help?
```

Trace: {
  error: 'conflict',
  reason: 'Document update conflict.',
  status: 409,
  name: 'conflict',
  message: 'Document update conflict.',
  stack: 'Error\n' +
    '    at generateErrorFromResponse (/app/node_modules/pouchdb/lib/index.js:694:18)\n' +
    '    at /app/node_modules/pouchdb/lib/index.js:6816:19\n' +
    '    at runMicrotasks (<anonymous>)\n' +
    '    at processTicksAndRejections (internal/process/task_queues.js:95:5)',
  docId: 'screen_f3e3d025f4c748ad93659ec41d687a0e'
}
    at /app/dist/api/index.js:74:21
    at Generator.throw (<anonymous>)
    at rejected (/app/dist/api/index.js:6:65)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

PaulMcF1987:
@aptkingston @JayP718 @poirazis 
I have the same issue, I opened a bug issue on it as well - https://github.com/Budibase/budibase/issues/6670

JayP718:
@PaulMcF1987 

Thanks for chiming in, glad it's not just me. I'm getting 400 errors too I sent part of the docker logs.

Everything was working smoothly prior to this update, hopefully it's not a tough fix.

poirazis:
Hey @aptkingston 
any progress on this ? Should we continue building with v.1.1.4 or do we risk corrupting our app ?

shogunpurple:
[...]

PR #6883, opened 2022-07-25
Ignore harmless 409s in write-through cache

aptkingston:
## Description
This is a small PR to simply ignore 409s which occur in our write through cache. This is safe to ignore as if a 409 occurs here, it simply means that this particular workload has already been handled by someone else. Redis itself will still continue to hold the correct and up to date information.

Added in a small log so we can at least detect the frequency with which these are happening. Didn't think it was worth running this through the full log processor, and the only info we have available to us here is doc IDs which I also didn't think was worth logging.

Addresses #6691


mike12345567 (approved):
LGTM!

Issue #6778, opened 2022-07-18
Bindings issue with MongoDB


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6771

<div type='discussions-op-text'>

<sup>Originally posted by **cyl1234567** July 18, 2022</sup>
Hi, Every one

I can run a MongoDB update directly. But when I use bindings, the system always toast a query error.
("ID" is a field in numeric format)

Is this a bug or am I missing a specific escape character?

app version: 1.1.7

screen shot:
- query directly
  - <img width="527" alt="image" src="https://user-images.githubusercontent.com/109498463/179445455-7bf5ee52-2fa1-4e37-9677-f4e3234753f6.png">
- use bindings
  - <img width="522" alt="image" src="https://user-images.githubusercontent.com/109498463/179445831-bae8d99a-e4a6-4d8d-b526-1d6a0920cf95.png">
  - <img width="321" alt="image" src="https://user-images.githubusercontent.com/109498463/179446305-e37ab1c4-975d-4144-b83c-5c166a8ffb1b.png">

</div>

Replicated for **updateOne**

melohagan:
I think this is actually an issue with the bindings.

![bindings_cleared](https://user-images.githubusercontent.com/101575380/179492678-b5523fcd-99bc-457a-9742-623caa4515a1.gif)

![Screenshot 2022-07-18 at 11 25 48](https://user-images.githubusercontent.com/101575380/179492766-83b88ade-1482-4bc5-a705-eaec4d9b99f1.png)

deanhannigan:
In the second example  the binding needs to be changed from

`{  "id" : { "$eq" : {{ id }} }` to 
`{  "id" : { "$eq" : "{{ id }}" }`

The binding expressions must be added a strings to any query body.

The bug here was related to a UI change that caused the value field to clear.

cyl1234567:
> In the second example the binding needs to be changed from
> 
> `{ "id" : { "$eq" : {{ id }} }` to `{ "id" : { "$eq" : "{{ id }}" }`
> 
> The binding expressions must be added a strings to any query body.
> 
> The bug here was related to a UI change that caused the value field to clear.

Hi @deanhannigan,

"id" is a numeric field in the second example.
Is it possible to use binding expressions without double quote? otherwise, I thinks the update can't match any document.

mongo db document:
```
{
	"_id" : ObjectId("62d54a93aaee5e087b5a90d5"),
	"id" : 11,
[...]

PR #6895, opened 2022-07-26
Support ObjectId in MongoDB operators

melohagan:
## Description
Fix to recognise object ids inside MongoDB operators such as `$eq`

Addresses: 
- https://github.com/Budibase/budibase/issues/6778
- https://github.com/Budibase/budibase/issues/7079 
- https://github.com/Budibase/budibase/discussions/7056


Rory-Powell (approved):
Looks good mel. It would be nice to see a test as well where we can supply some variations of json queries and assert that mongo is called with the right data

melohagan:
@Rory-Powell I realised that the regex wasn't handling scenarios where there were braces `{}` within strings (double quotes). 
I removed the regex and did it programmatically because it was starting to get too complicated for my regex skills. 
Added a unit test to cover this scenario as well!

Rory-Powell (approved):
LGTM 👍

PR #6897, opened 2022-07-26
Nord theme + Midnight theme

aptkingston:
## Description
This PR adds 2 new theme options to Budibase. These options are only available for the actual Budibase platform - they are not available to use (and therefore will not affect) client apps.

Both these new themes are available to select from the Theming section in settings.
image

Going forward, the process for adding a new theme is very simple:
- Think of a great name, e.g. `foo`
- Save the CSS file as `frontend-core/src/themes/foo.css` using a class decorator of `.spectrum--foo`. It's only really the grey palette which needs changed (see `midnight.css` for an example).
- Import the CSS file in `packages/builder/src/main.js`
- Add the name of the theme to `ThemeOptions` in `packages/frontend-core/src/constants.js`
- Select your new theme in settings

All custom themes are presumed to be based off `darkest`. We'll need to do more work to allow custom light themes.

## Nord theme
An arctic, blueish, mid brightness theme. Accent colours have been customised to suit the palette.
image
image
image

## Midnight theme
A classy dark theme with a slight blue tint, shifting it away from the default neutral palette and giving the UI more personality.
image
image
image


Rory-Powell (approved):
🌈

shogunpurple (approved):
Discounting the actual theme CSS - it's pretty incredible how we are able to enable this functionality with like 15 lines of code. Great work 👌  LGTM

Issue #6468, opened 2022-06-26
Deactivate schedule based automation after too many failures


shogunpurple:
**COS**
- When a schedule based automation (cron) fails 5 times in a row, deactivate the automation and stop running the cron job
- Notify the user in the automation logs panel that their automation was stopped due to failures
- Notify user they need to re publish their app to reactivate the automations

Ideally we can reactivate without publishing by turning on the bull processing again.

PR #6898, opened 2022-07-26
CRON automation stop after a number of errors

mike12345567:
## Description
Implements a change to the CRON system so that it can be automatically disabled after 5 failures. The backend of this doesn't utilise the the automation logs as there is a risk that the CRON job is on a long timer, like once a day, if it were a free user then it would never be disabled as there would only be one error log per day (before the rest were deleted).

We now had a metadata document that is stored only for CRON automations (although this can easily be extended if it were useful to some other automation type) to store the number of errors that have occurred. When this number reaches a maximum of 5, the CRON job will be disabled and a log created stating that this has happened (a specific status of `stopped_error`).

The UI will show that the CRON has been stopped and display a specific error to make the user aware that they need to re-publish their app to get their CRON running again.

Ideally in the future it would be good to have a system to email an administrator/creator of the app to alert them this has happened, but it is a significantly larger piece of work on top of the spec for this feature (which is mostly a performance optimisation).

The error count will reset to zero if at anytime during the count a success occurs, meaning if you have 4 errors, then a success, it will not be disabled on the next error.

Lastly, when the app is published all automation metadata documents will be wiped as the app has been re-published, we want to make sure that they have the chance to re-run the CRON 5 times if needed.

This also requires a small budibase-pro change which can be found here: https://github.com/Budibase/budibase-pro/pull/41

Addresses: 
- #6468

## Screenshots
Showing how it appears when CRON disabled (in this scenario the full list of errors isn't shown as I've faked the status for speed).
image


NEOLPAR (approved):
LGTM

shogunpurple (approved):
LGTM! The use of the automation metadata documents is a pretty nice solution. Good job

Issue #6090, opened 2022-05-27
Automation App Action Input not processed (correctly) (for Query Row)


bjv:
**Describe the bug**
If starting/testing a Query in an App Action Based Automation, the trigger field input is not processed correctly

**To Reproduce**
1. Create an App Action Based Automation
2. Created an input field 'id'
3. Create a step: "Query Rows"
4. Set filtering to a column which filters your table like
![image](https://user-images.githubusercontent.com/1166771/170705921-a109a4fe-a4ec-4d60-881c-e2e04138ef58.png)
5. Run Test with JSON like: `{ "fields": { "id": 1234 } }`
6. Create another next step like "SMTP": `{{ steps.1.rows }}` contains the WHOLE table, unfiltered

**Expected behavior**
{{ steps.1.rows.length }} only contains filtered rows

R2bEEaton:
Relatedly, I am getting the following error in the JavaScript console when trying to test a filtered query. My automation doesn't even complete, just stops there and the GUI becomes unresponsive.
![image](https://user-images.githubusercontent.com/34921506/175524431-76f82963-56dd-44f8-b984-3232972c3e5d.png)

Shadowfita:
> 

I am having the same issue when trying to create an app action automation.

Shadowfita:
> Relatedly, I am getting the following error in the JavaScript console when trying to test a filtered query. My automation doesn't even complete, just stops there and the GUI becomes unresponsive. ![image](https://user-images.githubusercontent.com/34921506/175524431-76f82963-56dd-44f8-b984-3232972c3e5d.png)

I am having the same issue when trying to create an app action automation.

NEOLPAR:
Found an issue where the UI isn’t mapping to what the app action automation expects. It will be solved as part of this issue.

Behaviour:
When you test an automation, the Form view won't map it properly, being impossible to test it through the form and forcing the use of the JSON.

COS
The form should be mapped and show the form properly.

mike12345567:
@bjy @Shadowfita - in terms of your issues, we've not been able to find any issues with creating app action automations, or filtering using app actions - could you try to recreate your issues in the latest version of Budibase?

If you can recreate please provide detailed reproduction steps to help us diagnose.

bjv:
> @bjy @Shadowfita - in terms of your issues, we've not been able to find any issues with creating app action automations, or filtering using app actions - could you try to recreate your issues in the latest version of Budibase?
[...]

PR #6904, opened 2022-07-26
fixed automation app:trigger not mapping test modal form

NEOLPAR:
Description
In automation, App Actions, the testing modal wasn't mapping properly the form which forced users to use the JSON if they wanted to test their automation. 

Addresses: 
- This task was included as part of issue [#6090](https://github.com/Budibase/budibase/issues/6090) 

Before
image

After
Screenshot 2022-07-26 at 15 36 43

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
@@ -15,8 +15,6 @@
   let trigger = {}
   let schemaProperties = {}
 
-  $: trigger
-  $: schemaProperties
   $: {
     // clone the trigger so we're not mutating the reference
     trigger = cloneDeep(
@@ -27,7 +25,7 @@
     let schema = Object.entries(trigger.schema?.outputs?.properties || {})
 
     if (trigger?.event === "app:trigger") {
-      schema = Object.entries({ fields: { customType: "fields" } })
+      schema = [["fields", { customType: "fields" }]]
     }
 
     schemaProperties = schema
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
@@ -15,16 +15,22 @@
   let trigger = {}
   let schemaProperties = {}
 
-  // clone the trigger so we're not mutating the reference
-  $: trigger = cloneDeep(
-    $automationStore.selectedAutomation.automation.definition.trigger
-  )
+  $: trigger
+  $: schemaProperties
+  $: {
+    // clone the trigger so we're not mutating the reference
+    trigger = cloneDeep(
+      $automationStore.selectedAutomation.automation.definition.trigger
+    )
 
-  // get the outputs so we can define the fields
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -15,16 +15,22 @@
   let trigger = {}
   let schemaProperties = {}
 
-  // clone the trigger so we're not mutating the reference
-  $: trigger = cloneDeep(
-    $automationStore.selectedAutomation.automation.definition.trigger
-  )
+  $: trigger
```
shogunpurple, line 18: These shouldn't be needed as reactive variables initially - you can just define them as `let trigger` and `let schemaProperties` below.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -15,16 +15,22 @@
...
+      $automationStore.selectedAutomation.automation.definition.trigger
+    )
+
+    // get the outputs so we can define the fields
+    let schema = Object.entries(trigger.schema?.outputs?.properties || {})
 
-  // get the outputs so we can define the fields
-  $: schemaProperties = Object.entries(trigger?.schema?.outputs?.properties)
+    if (trigger?.event === "app:trigger") {
+      schema = Object.entries({ fields: { customType: "fields" } })
```
shogunpurple, line 30: It would be more explicit here to just write the actual value you need rather than use `Object.entries`:

```
[[ "fields", { customType: "fields" } ]]
```

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -41,13 +43,25 @@
...
   $: stepId = block.stepId
   $: bindings = getAvailableBindings(
     block || $automationStore.selectedBlock,
     $automationStore.selectedAutomation?.automation?.definition
   )
-  $: inputData = testData ? testData : block.inputs
+
+  $: getInputData(testData, block.inputs)
+  const getInputData = (testData, blockInputs) => {
+    let newInputData = testData ? testData : blockInputs
```
shogunpurple, line 56: NAB: can be `testData || blockInputs`

shogunpurple (approved):
Just a few minor comments - LGTM and is a nice improvement for those using app actions.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
NEOLPAR: Removed! thanks for that

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
NEOLPAR: You're right, it's simpler, thanks

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
NEOLPAR: done! thanks

Issue #6905, opened 2022-07-26
Filter drawer points to the wrong table when duplicating table blocks / data providers


melohagan:
**Hosting**
Self
  - Method: docker compose
  - Budibase Version: 1.1.7
  - App Version: 1.1.7

**Describe the bug**
When you add a filter to a data provider or table block, and then duplicate the data provider, the filter drawer will only show the last data providers filters. 

**To Reproduce**
1. Add a table block with a filter on one value
2. Duplicate that table block and change the filter value
3. Click define filters on the original table block
4. Notice that it shows the filter of the second instead

**Expected behavior**
When I duplicate table blocks / data providers, the filters should be untied / unique to each table block.

**Screenshots**
![filterbug](https://user-images.githubusercontent.com/101575380/181061300-5419684b-5786-4cc3-801d-63830b69af1f.gif)

aptkingston:
Hey @melohagan. Pretty sure this is just an issue with not remounting settings properly. I reckon if you refresh your page, you'll find that the filters actually are unique. I think we're basically just showing the same filter drawer here. Selecting a different component (that isn't a table block) before selecting the other block would probably fix this.

I'll put in a proper remounting fix though.

melohagan:
Selecting a different component helps, although you have to do it every single time you want to switch between filters of the same component type. 
E.g. Edit Table Block A filters, click container, Edit Table Block B filters.

aptkingston:
Yep that just confirms that the bug is what I thought it was - I don't mean swapping components is a valid solution long term haha. I've opened a PR to fix it properly.

PR #6914, opened 2022-07-27
Fix remounting component settings

aptkingston:
## Description
Fixes an issue where some component settings were stale when selecting another component of the same type. This change ensures the component settings panel fully remounts when changing component.

Addresses: 
- #6905


melohagan (approved):
LGTM!

Issue #6896, opened 2022-07-26
Can't open PDF inside the table block


Klience:
**Hosting**
- Cloud
  - Version 1.1.21

**Describe the bug**
When a PDF attachment gets opened from a "table" or "table block" component, nothing happens.

**To Reproduce**
1. Create a Budibase internal table
2. Add an attachment field
3. Upload PDF to attachment field
4. Set up a data provider with a "table" or "table block" component inside
5. Configure attachment column to be shown inside the component's rows
6. Click on the PDF attachment inside the row
7. Nothing happens

**Expected behavior**
PDF should be opened in another browser tab like PNG attachments do.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 103.0.5060.134, 64 bit

**Additional context**
Please move this to **enhancement** if the inability to open PDF attachments in another window (browser tab) like PNG attachments is intentional.

PR #6915, opened 2022-07-27
Fix attachment renderer

aptkingston:
## Description
Fixes an issue where non-image files were not working inside tables due to our tooltip component. The tooltip component causes a lot of issues and needs rewritten at some point.

This affected not only client app table and table block components, but also tables inside the data section on the builder.

Addresses: 
- #6896

Working renderer:
image

Attachment open in new tab after clicking:
image


shogunpurple (approved):
LGTM

Issue #6813, opened 2022-07-19
Query row in automation always return empty array


Promphut:
**Hosting**
- Self
  - Method: k8s
  - Budibase Version: 1.1.11-1.1.17
  - App Version: <version> 

**Describe the bug**
if filter in data return a row but when filter same column same value in "Query row" on automation return empty array.

**To Reproduce**
1. Go to 'Automation'
2. Click on 'Add new automation -> webhook'
3. add "Query row" 
4. Query row with column and value that existing.
5. See error

**Expected behavior**
It should return a row because query is exist.

**Screenshots**
![image](https://user-images.githubusercontent.com/19907663/179763702-e3228bd7-b415-479c-bbec-e75e94e93ac1.png)
![image](https://user-images.githubusercontent.com/19907663/179763953-38137327-c0d2-4665-b330-eebd0fb600fd.png)
![image](https://user-images.githubusercontent.com/19907663/179764000-38eeea35-8766-4151-b1ee-eedea87a72d8.png)

 - Browser [Edge Version 103.0.1264.51]

Promphut:
@mike12345567 Hi, Thank you for reply. Sorry about replication steps. It's occure when "Row Created" or "Row Updated".

1. Go to 'Automation'
2. Click on 'Add new automation -> Row Updated'
3. add "Query row"
4. Query row with column and value that existing.
5. See error

Promphut:
@mike12345567 I try to create a new simple app, The problem isn't occur. And I try to create simple workflow on the old app, The problem still occur. 
What could i do?
My app is so big. My data is more than 50k rows.

mike12345567:
Hi @Promphut - it could be worth trying exporting your app and re-importing it incase something has gone wrong, as well as making sure your app is up to date.

If your app is not up to date then you'll see this icon in the top right:
![image](https://user-images.githubusercontent.com/4407001/180179385-f38699c3-5370-433b-898e-7a739a1e9c05.png)

If you can supply us with an exported app this is exhibiting this behaviour it would greatly speed up the diagnose process as well, you can email exports through to support@budibase.com with a title of the issue number.

Promphut:
@mike12345567 I have already sent my exported app to [support@budibase.com](mailto:support@budibase.com) 

title: Query row in automation always return empty array #6813

thank you

Promphut:
@mike12345567 Did you find anything wrong in my app?

PR #6928, opened 2022-07-27
Various July fixes

mike12345567:
## Description
Some minor fixes to go out.

Addresses: 
- #6813 fix for query rows not being identical to data table experience.
- Fix for issue with automation UI, running a new test didn't update the side panel, meaning that it looked like the test hadn't run, had to close panel to get new test results through.
- Fix for cloud, where the `authenticated` middleware was throwing an undefined error when deleting the user password, when no user was found (signup).

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.25",
+  "version": "1.1.26",
   "npmClient": "yarn",
   "packages": [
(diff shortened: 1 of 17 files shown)


shogunpurple (approved):
LGTM!

Rory-Powell (approved):

Issue #5352, opened 2022-04-11
Allow for ILIKE search of an ANY or ALL array of multiple values so that OR can be used


bamboowonder:
Hi there,

It would be a fantastic time saver, if we could just use the new Dynamic search everywhere instead of hand building search.

For that to be possible, we need to be able to ILIKE search using OR.

For Postgres, this is possible by using an ANY/ALL array.   (which interestingly, the pg planner breaks many queries down to this format anyway)

for a single line case-insensitive CONTAINs that allows for multiple values that use OR
UI:    field -> contains -> value1,value2
DB:   WHERE field ILIKE ANY('{%value1%,%value2%}')

and the opposite for a single line NOT
UI:    field -> doesn't contains -> value1,value2
DB:   WHERE field NOT ILIKE ALL('{%value1%,%value2%}')

I'd so appreciate this, and I think your other users would too!

Thank you,

mike12345567:
Hi @bamboowonder - just to confirm, you'd like to be able to do this on the internal, Budibase tables, or would you like this specifically for SQL based tables?

bamboowonder:
This is for external postgres primarily.  Though I imagine we'd use internal table for adhoc apps, but much less so.

Issue #3773, opened 2021-12-13
Cannot use multi-select dropdown as filter


mjashanks:
Not sure if this is a bug or an enhancement!

However, mutti-select pickers don't really work as filter components. So, say I have:
- A BB table with a mutli-select column - `Tags`
- A component structure of
```
> Form (for filtering)
   > Multi-select: with values `feature`, `bug`, `data`, `design`
   > Data Provider: with filter of `Tags Contains Form.Multi Select`
       > Table
```
- a record, with tags `feature`, `data`

When I select `feature` from the filter, my record is included in the results (correct)
When I select `feature` and `data` from the picker, my record is **not** included in the result (incorrect)

Ideally, when filtering a multi-select field, we would want operators of
- `Has any of`
- `Has all of`
- `Has none of`
- `Has only` (only the selected values, no more, no less)

aptkingston:
Yep that's just an enhancement then - we'd need to add those new operators exactly like you said. Not all of those might be possible though so we'd need to check what we can do. In fact I know right away that `Has any of` would be impossible, but I think the others could work. `Contains` is just a single value lookup like you discovered. The fact it's a multi select makes it feel like you should be able to filter on multiple though so I get why this might be expected behaviour.

joebudi:
I'm just flagging this here. When using a block (card and table) it's not possible to use a multi-select as a filter. If we don't/can't offer this, we should consider providing some feedback to the user.

geekrambo:
I'm looking for a way to show fields in a form through conditions of a multi-select drop-down. It would be nice to have a "Contains" condition along with Equals, Not Equals, Starts with, Like, Is empty, and Is not empty.

PR #6931, opened 2022-07-27
Multi-select dropdown filters

melohagan:
## Description
Added the following filters for Multi-select data types:
- Contains
- Not Contains (Inverse of contains)
- Has any

A multi-select picker was added to the filter drawer to choose the range of values to compare against. 
I had a look at exact equals, but I don't think you can do this in Lucene because you'd need to know the NOT cases (which could be unknown if using a dynamic binding).

Support:
- InternalDB
- SQL (Postgres, MySQL, SQL Server)
- Bindings parse comma separated strings as arrays, or can return an array in JavaScript
- Custom queries (bindings only)

SQL Server has a *CONTAINS* function, but apparently full text search is turned off by default (as was my case). So using LIKE to compromise.
 
Addresses: 
- https://github.com/Budibase/budibase/issues/3773
- https://github.com/Budibase/budibase/issues/5352

## Value filter
Screenshot 2022-07-27 at 18 05 54

## Binding filter
Screenshot 2022-07-27 at 18 06 20

## Custom query (MongoDB array)
Screenshot 2022-07-27 at 18 03 24

## Internal DB
Screenshot 2022-07-27 at 18 05 16

## MySQL
Screenshot 2022-07-27 at 18 20 02

## PostgreSQL
Screenshot 2022-07-27 at 18 24 57

<h2>SQL Server</h2>

Screenshot 2022-07-28 at 09 18 26


mike12345567 (approved):
LGTM - nice addition of test cases! The only thing that would be good to do is update the internal table Lucene test cases as well just to add some stuff for the new `containsAny` and `notContains` operations - here: https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/routes/tests/internalSearch.spec.js

PR #6982, opened 2022-07-29
Joe's lab day

joebudi:
## Description
I've added a new theme to the builder called the Great Wave, named after the painting '[The Great Wave off Kanagawa'.](https://en.wikipedia.org/wiki/The_Great_Wave_off_Kanagawa). On top of that, I've also updated several minor design inconsistencies throughout the builder.

**Theme screenshots**
CleanShot 2022-07-29 at 11 55 52@2x

CleanShot 2022-07-29 at 11 57 34@2x

**Design inconsistencies**
Throughout the builder, I found inconsistencies regarding titles, so I followed [Spectrums design guide ](https://spectrum.adobe.com/page/grammar-and-mechanics/#Capitalization)and went with sentence case.

CleanShot 2022-07-29 at 12 03 17@2x

**Improved sizing and spacing**
I updated the sizing of some components to make the items within the component easier to read. I've also improved the space across multiple layouts.

CleanShot 2022-07-29 at 11 56 41@2x

**Increased number of app icons**
Users can now select from over 50 icons.

CleanShot 2022-07-29 at 11 58 39@2x


packages/builder/src/pages/builder/portal/overview/_components/SettingsTab.svelte
```diff
@@ -27,18 +28,24 @@
 <div class="settings-tab">
   <Page wide={false}>
     <Layout gap="XL" paddingY="XXL" paddingX="">
+      <Layout gap="XS" noPadding>
+        <Heading size="M">App settings</Heading>
```
melohagan, line 32: NAB: A subjective point, but imo this heading is redundant because you can see that the settings tab is highlighted and selected.
I like the 'App details' header though.

melohagan (approved):
The new theme looks fantastic - great job! 🏄 

The other little UI tweaks are great too.
As discussed, I reverted the changes to `QueryBindingBuilder` and double checked the titles that needed changed to match spectrum.

aptkingston:
Love the color palette and the tweaks Joe! :rocket: 

The one thing I think we do need to change though is using the system font - on Linux for example this ends up looking pretty awful (probably on Windows too), but I appreciate it looks alright on Mac. I think it's really important that we specify a font family to be used. The contrast is also extremely low on my screen (still a good IPS panel) but probably looks stronger on liquid retina Mac screen. The font issue is highlighted by the fact that the spectrum line height is still used but the actual font size will depend on whatever the client default is, in my case making the font height look quite off:
image

I've pulled in develop to highlight that some of the new user groups stuff clashes a little with the palette. Here's the user details page for example. Without being able to see the borders, the list of apps and groups stands out as being too dark I think, since it all blends together.
image

Tooltips were the only other thing I found quite hard to read:
image

joebudi:
Hey

> The one thing I think we do need to change though is using the system font - on Linux for example this ends up looking pretty awful (probably on Windows too), but I appreciate it looks alright on Mac. I think it's really important that we specify a font family to be used. The contrast is also extremely low on my screen (still a good IPS panel) but probably looks stronger on liquid retina Mac screen. 
The fonts are not consistent in the theme as I couldn't override certain font variables. That linux font is horrible. Windows uses Arial (I think), so it should look fine and clean.  The blue CTA could be made a smidge darker.

> Tooltips were the only other thing I found quite hard to read:
Once again, I couldn't change this as it was based on a dark theme.

I'll try and have another look at this on Friday.

Issue #6971, opened 2022-07-29
Import handlebars helper functions into JS context 


shogunpurple:
**COS**
- Import the full helpers library that is available in handlebars into the JS binding experience under a `helpers.functionName` namespace
- Document

PR #6986, opened 2022-07-29
JS helpers, development only HBS -> JS conversion

mike12345567:
## Description
As part of my lab-day I built a very basic transpiler which can convert Handlebars functions into JavaScript functions. For now this PR keeps it only to development use, it is not in the production system, allowing us to play with this a bit and improve it/finalise a UI if we want to make it generally available.

This also addresses the latest issue #6971 - importing the handlebars helpers into the JavaScript context, meaning that dates, durations, maths, array, string etc helpers we have in HBS can now be used in any JS function.

The transpiler still has a little bit of work needed around HBS conditional statements, but for now it can translate the vast majority of HBS statements, including deeply nested ones.

One proof of its level of functionality today is that when I was adding the helpers list to the JS binding menu, I realised all the examples were in Handlebars, as such:
image

It would have been a huge job to add JS and HBS examples for everything, and the syntax is already understood by the conversion system, so:
image

Addresses: 
- #6971 - adds JS helpers and their descriptions.


Rory-Powell (approved):
Very cool 👍

packages/string-templates/src/conversion/index.js
```diff
@@ -0,0 +1,129 @@
...
+      .join(", ")
+  }
+  if (!value) {
+    return parts.length > 1 ? `${build()}` : build()
+  } else {
+    return parts.length === 0 ? value : `${value}, ${build()}`
+  }
+}
+
+function splitBySpace(layer) {
```
aptkingston, line 63: See I knew building a transpiler was completely trivial and not complicated at all. This file proves it once and for all!

packages/builder/src/components/common/bindings/BindingPanel.svelte
```diff
@@ -62,15 +68,24 @@
...
-  const addHelper = helper => {
-    hbsValue = addHBSBinding(hbsValue, getCaretPosition(), helper.text)
-    updateValue(hbsValue)
+  // Adds a JS/HBS helper to the expression
+  const addHelper = (helper, js) => {
+    let value
+    const pos = getCaretPosition()
+    if (js) {
+      const decoded = decodeJSBinding(jsValue)
+      value = jsValue = encodeJSBinding(
```
aptkingston, line 77: The only thing here is that previously I didn't immediately update `value`, because when we call `updateValue` it first checks validity, and then dispatches the change event. That will eventually then result in the new `value` prop being passed back in here. By directly changing value it means we could be setting `value` to something that hasn't passed a validity check yet. Not saying that will definitely cause a problem, but I'd prefer to avoid updating it directly here unless it's needed.

aptkingston (commented):
Awesome feature and looking forward to playing around with it!

packages/builder/src/components/common/bindings/BindingPanel.svelte
mike12345567: Makes sense, I did it this way originally as I was having problems getting the helper to be added to codemirror correctly, but I'll try and find another way!

PR #6987, opened 2022-07-29
Minor automation fixes

mike12345567:
## Description
A minor fix uncovered by recent changes to the automation `queryRows` action - now when working with number/string type differences it would break (couldn't specify a HBS statement for a number filter).

Also added a minor output the `serverLog` action, so that what it actually output can be seen in the automation run log.

Addresses: 
- #6090 - discovered the type issue.


shogunpurple (approved):
LGTM

PR #6989, opened 2022-07-29
User Management UI fixes

PClmnt:
## Description
Just some fixes for issues found in testing and some more optimisation of specific parts of the code:

- Hides the User Groups tab from the portal using a feature flag so we can show it when licensing is ready
- Fixes validation on new user emails
- Update the searching of user in app assigment to use the new users api which is much faster and more efficient
- uses the users by app view in order to count total users per app which improves the efficiency of that app overview page a lot
- Fix User ID page name and avatar

### DIFF
--- a/packages/worker/src/sdk/users/users.ts
+++ b/packages/worker/src/sdk/users/users.ts
@@ -31,9 +31,7 @@ export const allUsers = async () => {
 }
 
 export const countUsersByApp = async (appId: string) => {
-  let response: any = await usersCore.searchGlobalUsersByApp(appId, {
-    include_docs: true,
-  })
+  let response: any = await usersCore.searchGlobalUsersByApp(appId)
   return {
     userCount: response.length,
   }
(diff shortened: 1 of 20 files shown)


packages/worker/src/sdk/users/users.ts
```diff
@@ -30,6 +30,15 @@ export const allUsers = async (newDb?: any) => {
   return response.rows.map((row: any) => row.doc)
 }
 
+export const countUsersByApp = async (appId: string) => {
+  let response: any = await usersCore.searchGlobalUsersByApp(appId, {
+    include_docs: true,
```
shogunpurple, line 35: We shouldn't need to `include_docs` here if it's just a count

packages/worker/src/sdk/users/users.ts
```diff
@@ -56,7 +65,7 @@ export const paginatedUsers = async ({
     userList = await usersCore.searchGlobalUsersByEmail(email, opts)
     property = "email"
   } else {
-    // no search, query allDocs
+    // no search, query allDcso
```
shogunpurple, line 68: Small typo

shogunpurple (approved):
LGTM

Issue #6975, opened 2022-07-29
Add 'Is In' filter to Option data type


melohagan:
**Describe the feature request**
At the moment you can use 'Is In' for string and number types. I should also be able to do this for Option types.
Discussion: https://github.com/Budibase/budibase/discussions/6934

Pulsdesign:
Cool - thx fot that! We will wait until this is working - thx a  lot

PR #6990, opened 2022-07-29
Add 'Is In' filter to Option data type

melohagan:
## Description
Added the 'Is in' filter for options type. 
Includes value matching multiselect, or binding that returns array or string with commas to delimit the values.

Addresses: 
- https://github.com/Budibase/budibase/issues/6975

## Screenshots
Screenshot 2022-07-29 at 16 12 35

Screenshot 2022-07-29 at 16 16 43

Screenshot 2022-07-29 at 16 17 08

Screenshot 2022-07-29 at 16 17 26

### DIFF
--- a/packages/bbui/src/Form/Multiselect.svelte
+++ b/packages/bbui/src/Form/Multiselect.svelte
@@ -27,7 +27,7 @@
     {error}
     {disabled}
     {readonly}
-    {value}
+    value={Array.isArray(value) ? value : [value]}
     {options}
     {placeholder}
     {sort}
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM!

packages/bbui/src/Form/Multiselect.svelte
```diff
@@ -27,7 +27,7 @@
     {error}
     {disabled}
     {readonly}
-    {value}
+    value={Array.isArray(value) ? value : [value]}
```
aptkingston, line 30: I see this is already merged so we can address this in the other branch -  but this expression will coerce nullish values into non-nullish values which can have unintended consequencies. `null` becomes `[null]` which is now not nullish, and the same for `undefined`.

It would be a better idea to use the existing `onOperatorChange` callback to ensure the value is parseable as an array one this operator is selected, and clear the value if it isn't. That's the purpose of those callbacks when changing fields or operators - to ensure the other properties of the expression are still sensible.

melohagan: Makes sense!

melohagan: I've made that change in the other PR/branch!

PR #7015, opened 2022-08-01
Keyboard shortcuts for components

aptkingston:
## Description
This PR adds keyboard shortcuts for components, greatly improves the logic of moving components up and down and improves UX around using modals.

TLDR:
- All component actions in context menu have a corresponding keyboard shortcut
- All keyboard shortcuts work in the preview or in the tree - you no longer need to focus the preview for delete to work.
- `Ctrl+Enter` opens the new component panel for adding a component. `Escape` closes this again if you want to cancel.
- You can use the up and down arrows to scale the tree and change selected components
- After deleting a component, the "next" component is selected rather than the parent. This feels much more natural.
- When a modal opens, you can immediately dismiss it with `Escape` or immediately confirm it with `Enter`
- Custom themes can be used in client apps
- Fixed button roundness not applying to large buttons, and updated control for button roundness
- Modal background matches theme (optional, Nord and Midnight have opted in)
- Fix icon colours being static grey in automation blocks

### Keyboard shortcuts
All actions in the component context menu now show a legend for what the key combination is:
image

Pressing up arrow and down arrow (without `Control` held) will also change the selected component.
`Ctrl+Enter` opens the new component panel, and `Escape` closes it, if it's open.

When moving components up and down, or changing selected components using the up and down keys, you can now properly move up and down the tree - even going inside and outside of other components. This lets you pick the exact position of a component with far greater accuracy.

All keyboard shortcuts work no matter where you have focused. Whether you're clicking in the preview or clicking in the component tree, all the shortcuts will work. This is a big improvement over the current implementation, where the `Delete` key only works if you have focused the preview.

Due to the combination of keyboard shortcuts with the modal UX improvements, it's now extremely easy to delete multiple components sequentially, or move things around, without ever touching the mouse.

### Modal focus
When a modal opens, focus will be automatically set on the most useful component.
- If a modal contains an input, the input is focussed
- If it does not, the confirm button is focussed so that `Enter` will immediately submit it

Pressing `Escape` with a modal open will always dismiss it, without having to click on it first like we do at the moment.

Modal underlay background colour is now themable and respects the theme on both Nord and Midnight. This looks much better than using a flat translucent black.

### Theme settings update
You can now use any theme (including the new Nord and Midnight) in client apps:
image

Button roundness is now controlled by a slider with an example button beside it:
image


shogunpurple (approved):
LGTM - this is such an awesome feature. Although:

- We shouldn't merge until after release
- This needs pretty extensive documentation in readme

Once the 2 above are sorted, we can go ahead 👍

joebudi:
Nice work @aptkingston 

I have a couple of queries in re to keyboard shortcuts:
CTRL - is that the control button for Mac, or the CMD button? Will this update to match the user's OS?
Also, in other designs re keyboard shortcuts, I've always liked how they've made the keyboard actions look a little like muted buttons. You might need to increase the width of the dropdown if you took this approach.
CleanShot 2022-08-01 at 11 02 03@2x

aptkingston:
> Nice work @aptkingston
> 
> I have a couple of queries in re to keyboard shortcuts: CTRL - is that the control button for Mac, or the CMD button? Will this update to match the user's OS? Also, in other designs re keyboard shortcuts, I've always liked how they've made the keyboard actions look a little like muted buttons. You might need to increase the width of the dropdown if you took this approach. CleanShot 2022-08-01 at 11 02 03@2x

It's just CTRL at the minute for everyone - there's no difference between OS. Should be able to also listen for CMD and allow that too, but i'll just need to check that all the shortcuts still work on Mac and that Mac doesn't force capture any of them. Should be fine though!

That little button design looks nice - I haven't seen anything like it but it looks good so I'll update to that.

aptkingston:
Updated to use new style for key binds:
image

On MacOS the `Cmd` key can be used instead of `Ctrl`, but both work. On MacOS the keybind symbol updates to show the `Cmd` button symbol:

shogunpurple:
@aptkingston backspace is pretty commonly accepted for this kind of thing - could possibly use that?

joebudi:
This looks great!

aptkingston:
> @aptkingston backspace is pretty commonly accepted for this kind of thing - could possibly use that?

Yep sounds good - will go for that.

PR #7019, opened 2022-08-01
User groups fixes

aptkingston:
## Description
Collection of fixes from testing out the new user groups feature.

Fixes:
- Prevent adding a group to an app multiple times - this was causing the number of apps in a group to be artificially large as you could keep adding the same group since it wasn't being properly filtered out
- Fix display of role names in user details and group details pages. They will now show the role name rather than ID, and custom roles will show as "Custom role".
image
- Remove all instances of truly global CSS which was removing border radius from all selects everywhere. When using `:global` - always prefix it with the outer container of your component or else the styles are truly global
- Update all instances of role status lights to be square to match new designs
- Add secondary placeholder to assignment modal to fix text overflow
- Fix styles of search component embedded in `PickerDropdown`
- Fix "create new user group" modal showing the previously saved data when opened. It now resets back to the default new group values.
- Fix stale user data showing when navigating to user details page and then flashing
- Improve display of user name and email as some cominbations still resulted in `undefined` being shown
- Fix role colours in group details always showing pink due to missing `app_` prefix
- Update style of groups list to be the same size as table rows and list items
- Update button casing of multiple buttons to remove extra caps

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.29-alpha.1",
+  "version": "1.1.32-alpha.3",
   "npmClient": "yarn",
   "packages": [
     "packages/*"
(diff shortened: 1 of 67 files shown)


shogunpurple (approved):
LGTM

PClmnt (approved):
Great fixes, thanks so much @aptkingston !!

PR #7023, opened 2022-08-01
Fix add component button in empty screens

aptkingston:
## Description
This PR fixes the add component button in empty screens which currently doesn't work.


shogunpurple (approved):
LGTM

Issue #6737, opened 2022-07-15
Fetch tables does not remove deleted column - PostgreSQL


melohagan:
**Hosting**
Self
  - Method: docker compose
  - Budibase Version: 1.1.4
  - App Version: 1.1.4

**Describe the bug**
If I delete a column from PostgreSQL in pgadmin, and refetch tables, Budibase does not remove the deleted column.
Discussion: https://github.com/Budibase/budibase/discussions/6586

**To Reproduce**
1. Fetch a PostgreSQL table
2. Within **pgadmin** delete a column from the table
3. You'll now see an error in the data table view (and no rows will load)
4. Refetching does not help

**Expected behavior**
Deleting a column from a PostgreSQL table outside of Budibase should not break your Budibase builder.

**Screenshots**
![Screenshot 2022-07-15 at 14 52 43](https://user-images.githubusercontent.com/101575380/179237097-78fed648-90da-4985-86e3-a81d68b88328.png)

insomniacj93:
@melohagan Not only in PostgresSQL.
I'm using MySQL/docker and I'm Having the same issue, after noticing that my newly created tables weren't showing (but the changes to the existing tables did show) I decided to try and fetch the tables again. Now I've got my newly created tables, but some columns that no longer exist in MySQL still appear in budibase. On delete the action says it's done, yet the columns won't get removed.

melohagan:
Likely related discussion: https://github.com/Budibase/budibase/discussions/6846

PR #7028, opened 2022-08-01
Fix for SQL column removal (external to Budibase)

mike12345567:
## Description
Fixing issue with being unable to re-fetch SQL table definitions after a column has been removed - making sure that column removal does not require totally refreshing the datasource. Also adding a way to see the error that the server returns as often it is not obvious what has caused no data to be returned.

Addresses: 
- #6737 

## Screenshots
error scenario:
image


shogunpurple (approved):
LGTM

PR #7037, opened 2022-08-02
Add updatedAt and createdAt as optional to all joi middleware

Rory-Powell:
## Description
Make sure that the `createdAt` and `updatedAt` fields that exist on all documents are optional on all requests


mike12345567 (approved):
LGTM!

PR #7038, opened 2022-08-02
Reduce event frequency

Rory-Powell:
## Description
- Add exclusion list to posthog processor to remove some `X:UPDATED` events which aren't as useful


shogunpurple (approved):
Great work rory - even a wee test in there to cover👌

PR #7039, opened 2022-08-02
Fix/app row render

deanhannigan:
## Description
When the apps were sorted, the unique identifier for the row was not updated meaning internal elements got themselves into a strange render state. 
The row index and app id now uniquely identify the row and will trigger a redraw for any nested elements on sort.

Addresses: https://github.com/Budibase/budibase/issues/6952

### DIFF
--- a/packages/builder/src/components/start/AppRow.svelte
+++ b/packages/builder/src/components/start/AppRow.svelte
@@ -6,7 +6,6 @@
   export let app
   export let editApp
   export let appOverview
-  export let idx
 </script>
 
 <div class="title" data-cy={`${app.devId}`}>
--- a/packages/builder/src/components/start/AppRow.svelte
+++ b/packages/builder/src/components/start/AppRow.svelte
@@ -30,7 +30,7 @@
     Never updated
   {/if}
 </div>
-<div class="desktop check">
+<div class="desktop">
   <span><AppLockModal {app} buttonSize="M" /></span>
 </div>
 <div class="desktop">
--- a/packages/builder/src/components/start/AppRow.svelte
+++ b/packages/builder/src/components/start/AppRow.svelte
@@ -6,6 +6,7 @@
   export let app
   export let editApp
   export let appOverview
+  export let idx
 </script>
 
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/start/AppRow.svelte
```diff
@@ -6,6 +6,7 @@
   export let app
   export let editApp
   export let appOverview
+  export let idx
```
aptkingston, line 9: This new prop doesn't seem to be used anywhere.

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -381,8 +381,13 @@
           </div>
 
           <div class="appTable" class:unlocked>
-            {#each filteredApps as app (app.appId)}
-              <AppRow {app} {editApp} {appOverview} />
+            {#each filteredApps as app, idx}
```
aptkingston, line 384: This is interesting, because by removing the key (app.appId) we're actually removing the ability for app rows to fully remount themselves when the order changes, which in my mind would actually create the sort of issues we're trying to solve here.

aptkingston (commented):
Interesting one Dean! Could you descibe how to cause the issues that this PR is addressing? I'd be keen to try and reproduce myself because if I saw these changes without the description I would bet that these were going to actually introduce issues whenever apps were sorted differently!

Edit: literally just saw that it addresses an open issue which described the problem. Ignore that request!

packages/builder/src/pages/builder/portal/apps/index.svelte
deanhannigan: I was confused by this too. My assumption was that the sorting of the filtered apps was somehow causing svelte issues when performing its diff on the rows. I thought that passing a prop that combined both the array index and the appId would solve this.

I've now altered the AppLockModal to add a `key` tag around the modal component inside. This appears to have the desired effect and the contents display without issue. I also replaced the previous key expression which, as you point out, wasn't the root of the issue.

shogunpurple (approved):
LGTM!

PR #7045, opened 2022-08-02
Prevent self deletion

aptkingston:
## Description
This PR prevents a user from deleting themselves in both the users list and the user details pages. They also cannot force reset their own password, but they can always change their password from the top nav popover.

### DIFF
--- a/packages/builder/src/pages/builder/portal/manage/users/[userId].svelte
+++ b/packages/builder/src/pages/builder/portal/manage/users/[userId].svelte
@@ -237,18 +237,21 @@
             </div>
           </div>
         </div>
-        <div>
-          <ActionMenu align="right">
-            <span slot="control">
-              <Icon hoverable name="More" />
-            </span>
-            <MenuItem on:click={resetPasswordModal.show} icon="Refresh"
-              >Force Password Reset</MenuItem
-            >
-            <MenuItem on:click={deleteModal.show} icon="Delete">Delete</MenuItem
-            >
-          </ActionMenu>
-        </div>
+        {#if userId !== $auth.user._id}
+          <div>
+            <ActionMenu align="right">
+              <span slot="control">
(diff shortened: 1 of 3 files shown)


shogunpurple (approved):
LGTM! ⏩

Rory-Powell (commented):
Looks good cheeks. I think we should also prevent this on the API side as `DELETE /users/id` is an available API operation. 

There is already some validation we do around not being able to delete the root account holder, so this could slot in nicely there as well:
https://github.com/Budibase/budibase/blob/master/packages/worker/src/sdk/users/users.ts#L376

We might also want to make sure we filter out the current user id from the `bulkDelete` function. Similarly we could do an `accounts.getAccountByTenantId` to get the account owner and filter them out of the bulk delete.

Edit: on the last point, the `budibaseUserId` is available on the account so we could use that as we don't have the email in bulk user deletion 
https://github.com/Budibase/budibase/blob/master/packages/types/src/documents/account/account.ts#L54

aptkingston:
> Looks good cheeks. I think we should also prevent this on the API side as `DELETE /users/id` is an available API operation.
> 
> There is already some validation we do around not being able to delete the root account holder, so this could slot in nicely there as well: https://github.com/Budibase/budibase/blob/master/packages/worker/src/sdk/users/users.ts#L376
> 
> We might also want to make sure we filter out the current user id from the `bulkDelete` function. Similarly we could do an `accounts.getAccountByTenantId` to get the account owner and filter them out of the bulk delete.
> 
> Edit: on the last point, the `budibaseUserId` is available on the account so we could use that as we don't have the email in bulk user deletion https://github.com/Budibase/budibase/blob/master/packages/types/src/documents/account/account.ts#L54

Totally agree - and Mike has already messaged me saying he's gonna push API protection to this branch too :+1:

mike12345567:
API protection has been pushed!

Issue #2585, opened 2021-09-09
Filter only applies final condition


mjashanks:
I am trying to create a filter with 2 x "Not Equal" conditions. Only the final filter is applied:

![image](https://user-images.githubusercontent.com/3524181/132753858-89b46a3b-f884-4d12-9509-66a81a5d4681.png)

It doesn't matter how many conditions I add - only the final one is applied. Same issue for "Equals"

aptkingston:
This is because of the structure that we use for the search payload - it's an object and therefore will only be able to store one type of condition for each column. I thought this was alright, because you'd never search for `name = bob` and `name = bill`, that wouldn't make sense. But for not equals it's a more valid thing to do. We'll have to refactor things a fair bit to support this.

magickaito:
This should be resolved. Thanks!

PR #7049, opened 2022-08-02
Updating filters to allow multiple uses of the same property and exposing allOr option

mike12345567:
## Description
Updating the filter system to allow adding multiple filter properties of the same name at once, as well as enabling the use of the `allOr` property from within the UI.

To support the multiple properties I've assigned every single filter condition key a number, so for the screenshot below the query structure would be:
```
equal: {
  "1:productname": "Computers",
  "2:productname": "Chairs",
}
```
This is easy to remove on the backend and could be helpful down the line for providing order, something we will need to support complex logic.

I've added a toggle which updates the filter structure with the `allOr` property so that your conditions can be executed that way, this is not a perfect solution as there is still the issue of complex statements like `((this || that) && somethingElse)` essentially supporting parentheses in our logic, but to support this we will need a major rework of the UI.

Addresses: 
-  #2585 - one of the oldest bugs in our system, this has always been an issue which gets brought up a lot on discussions.

## Screenshots
Multiple properties of the same name being filtered, as well as the OR condition toggle, that allows using OR logic rather than the default AND.
image

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
@@ -9,19 +9,26 @@
     Input,
     Layout,
     Select,
+    Toggle,
+    Label,
   } from "@budibase/bbui"
   import DrawerBindableInput from "components/common/bindings/DrawerBindableInput.svelte"
   import ClientBindingPanel from "components/common/bindings/ClientBindingPanel.svelte"
   import { generate } from "shortid"
   import { LuceneUtils, Constants } from "@budibase/frontend-core"
   import { getFields } from "helpers/searchFields"
+  import { createEventDispatcher } from "svelte"
+
+  const dispatch = createEventDispatcher()
 
   export let schemaFields
   export let filters = []
   export let bindings = []
   export let panel = ClientBindingPanel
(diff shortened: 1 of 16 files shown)


shogunpurple (approved):
LGTM! this is a really nice feature. I think we should add some good unit test coverage in here though, to ensure that the filters are working as expected and the queries are being built in the correct way.

mike12345567:
Hey @shogunpurple - good idea, the only issue is that due to the way we run our tests we can't really execute Lucene queries (no real Couch/Clouseau) but we could build some tests to confirm that the input filter structure outputs expected lucene syntax.

shogunpurple:
@mike12345567 yeah going to be hard to get a real integration test in that case, but as long as we ensure the structure is all good and the query generated is valid we should have some coverage.

mike12345567:
@shogunpurple - I've added in the following unit tests: https://github.com/Budibase/budibase/blob/2bb0dbdc06ef301718d9d29c0d0bbce474111cbd/packages/server/src/api/routes/tests/internalSearch.spec.js - `node-fetch` and the `_search` endpoint is mocked by default in our test environment, so we just get back what it tried to send to Clouseau, we can then compare that syntax to the input.

shogunpurple:
Nice one - that's as good as we will get without full integration 👌

PR #7065, opened 2022-08-03
Allow proxy upstream URLs to be configured at runtime

### LINKED ISSUE #7062: Kubernetes: Allow deployment to custom namespace

Currently on Kubernetes Budibase can only be installed into a namespace called _budibase_. It would be nice if the installation would support any namespace. This would for example allow to run staging and production deployments in different namespaces on the same cluster, so that updates can be applied to and tested in the staging deployment without affecting the production deployment in any way.

The only thing that prevents this is the proxy service that has hard-coded upstream URLs in its config which reference the namespace name.

When changing those upstream URLs accordingly (in our case using a forked Helm chart and an initContainer that manipulates the static config) Budibase runs fine in namespaces named other than _budibase_.

panos--:
## Description

This PR adds support to proxy to have its upstream URLs configured at runtime. The main driver for this is to allow running Budibase on Kubernetes in namespaces named other than _budibase_.

To achieve this the proxy's Dockerfile is updated so that relevant upstream URLs can be passed via environment variables.

The Helm chart and the docker-compose.yml are updated accordingly, but since all deployments always fetch the latest proxy docker image it is necessary to make the changed image compatible with existing deployments. This is done by leaving the old static configuration as is and putting the new dynamic configuration behind a feature switch that is enabled in the updated Helm chart and docker-compose.yml only.

The changes have been tested successfully in a Helm chart deployment as well as a docker-compose deployment. The updated proxy docker image has also been tested successfully in old Helm chart and docker-compose deployments (i.e. deployments that don't incorporate the changes made here).

Addresses: 
- Fixes #7062


Rory-Powell (commented):
Hey @panos-- , this is a great addition. Thanks for taking the time to improve this configuration. We've introduced a similar change in this PR around support for `envsubst`
https://github.com/Budibase/budibase/pull/7121

The changes to the `Dockerfile` in this PR could be updated to work in the same way, we're using the default behaviour of the nginx image with templates rather than using a custom script. I'd also be happy to use the dynamic approach as the default rather than introducing a separate `hosting/nginx-dyn.prod.conf.hbs` as having copies leads to a maintenance overhead. 

The changes to get this merged  think are then
- Remove the `ENABLE_DYNAMIC_CONFIG` variable
- Remove the custom entrypoint script
- Update dockerfile to work with latest changes
   - Should be a case of just removing the entrypoint and template file changes, keeping the environment variables 
- Update the base `nginx.prod.config.hbs` instead,  and the corresponding removals in `generateProxyConfig.js` - can now remove the `apps`, `worker`, `minio`, `resolver` handlebars substitutions in favour of env vars. 
- For completeness we should also add `COUCH_DB_UPSTREAM_URL` so we can remove that from the generate script also

The only real reason we would need the generation script at all is for the watchtower `{{#if watchtower}}` block, we could think of just leaving that on by default as it'll be unused in k8s environments and removing the script, but we may require the hbs functionality in future so it's worth keeping. 
 
I'll work on getting the corresponding changes made to our preprod and release environments in a separate PR (private repo). If you want to work on the above changes just give us a shout, that'd be great. Alternatively you can just merge this into a new branch instead and we can take on the above, up to you!

panos--:
Hi @Rory-Powell, thanks for the feedback. I'll give it a try and see if can incorporate your suggestions in the next few days.

panos--:
Hi @Rory-Powell, I incorporated the changes you suggested.

This means

* `ENABLE_DYNAMIC_CONFIG` is removed
* the custom shell script is removed
* dockerfile is updated
* `nging.prod.config.hbs` and `generateProxyConfig.js` are updated
* the `{{#if watchtower}}` is left in, essentially for demonstration purposes
* the couchdb upstream is now resolved from env var `$COUCHDB_UPSTREAM_URL`

Tested the docker-compose and Helm deployments locally. Both looking good.

The only thing is that these changes are not backwards-compatible for existing Helm chart deployments, since as soon as the new proxy docker image is published, existing deployments will pick it up (on next pod re-create). But the new docker image defaults to upstream URLs as used by the docker-compose deployment. So it will break.

If this is a concern (it should, I think), it would be possible to pass env defaults in as build arguments and this way having different defaults for the k8s and the docker-compose images. Like in the following diff. What do you think?

```diff
diff --git a/hosting/proxy/Dockerfile b/hosting/proxy/Dockerfile
index 97738c398..e18490b1a 100644
--- a/hosting/proxy/Dockerfile
+++ b/hosting/proxy/Dockerfile
@@ -1,5 +1,12 @@
 FROM nginx:latest
 
+ARG APPS_UPSTREAM_URL=http://app-service:4002
+ARG WORKER_UPSTREAM_URL=http://worker-service:4003
+ARG MINIO_UPSTREAM_URL=http://minio-service:9000
+ARG COUCHDB_UPSTREAM_URL=http://couchdb-service:5984
+ARG WATCHTOWER_UPSTREAM_URL=http://watchtower-service:8080
+ARG RESOLVER=127.0.0.11
+
 # nginx.conf
 # use the default nginx behaviour for *.template files which are processed with envsubst
 # override the output dir to output directly to /etc/nginx instead of /etc/nginx/conf.d
@@ -11,9 +18,9 @@ COPY error.html /usr/share/nginx/html/error.html
 
 # Default environment
 ENV PROXY_RATE_LIMIT_WEBHOOKS_PER_SECOND=10
-ENV APPS_UPSTREAM_URL=http://app-service:4002
-ENV WORKER_UPSTREAM_URL=http://worker-service:4003
-ENV MINIO_UPSTREAM_URL=http://minio-service:9000
-ENV COUCHDB_UPSTREAM_URL=http://couchdb-service:5984
-ENV WATCHTOWER_UPSTREAM_URL=http://watchtower-service:8080
-ENV RESOLVER=127.0.0.11
+ENV APPS_UPSTREAM_URL=$APPS_UPSTREAM_URL
+ENV WORKER_UPSTREAM_URL=$WORKER_UPSTREAM_URL
+ENV MINIO_UPSTREAM_URL=$MINIO_UPSTREAM_URL
+ENV COUCHDB_UPSTREAM_URL=$COUCHDB_UPSTREAM_URL
+ENV WATCHTOWER_UPSTREAM_URL=$WATCHTOWER_UPSTREAM_URL
+ENV RESOLVER=$RESOLVER
diff --git a/package.json b/package.json
index 4c24e0025..b536dcde6 100644
--- a/package.json
+++ b/package.json
@@ -56,10 +56,11 @@
     "build:docker": "lerna run build:docker && npm run build:docker:proxy:compose && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh $BUDIBASE_RELEASE_VERSION && cd -",
     "build:docker:pre": "lerna run build && lerna run predocker",
     "build:docker:proxy": "docker build hosting/proxy -t proxy-service",
+    "build:docker:proxy:k8s": "docker build hosting/proxy -t proxy-service --build-arg APPS_UPSTREAM_URL=http://app-service.budibase.svc.cluster.local:4002 --build-arg WORKER_UPSTREAM_URL=http://worker-service.budibase.svc.cluster.local:4003 --build-arg MINIO_UPSTREAM_URL=http://minio-service.budibase.svc.cluster.local:9000 --build-arg COUCHDB_UPSTREAM_URL=http://budibase-svc-couchdb.budibase.svc.cluster.local --build-arg RESOLVER=kube-dns.kube-system.svc.cluster.local",
     "build:docker:proxy:compose": "node scripts/proxy/generateProxyConfig compose && npm run build:docker:proxy",
-    "build:docker:proxy:preprod": "node scripts/proxy/generateProxyConfig preprod && npm run build:docker:proxy",
-    "build:docker:proxy:release": "node scripts/proxy/generateProxyConfig release && npm run build:docker:proxy",
-    "build:docker:proxy:prod": "node scripts/proxy/generateProxyConfig prod && npm run build:docker:proxy",
+    "build:docker:proxy:preprod": "node scripts/proxy/generateProxyConfig preprod && npm run build:docker:proxy:k8s",
+    "build:docker:proxy:release": "node scripts/proxy/generateProxyConfig release && npm run build:docker:proxy:k8s",
+    "build:docker:proxy:prod": "node scripts/proxy/generateProxyConfig prod && npm run build:docker:proxy:k8s",
     "build:docker:selfhost": "lerna run build:docker && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh latest && cd -",
     "build:docker:develop": "node scripts/pinVersions && lerna run build:docker && npm run build:docker:proxy:compose && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh develop && cd -",
     "build:docker:airgap": "node hosting/scripts/airgapped/airgappedDockerBuild",
@@ -85,4 +86,4 @@
     "install:pro": "bash scripts/pro/install.sh",
     "dep:clean": "yarn clean && yarn bootstrap"
   }
-}
\ No newline at end of file
+}
```

Rory-Powell:
Hi @panos--, very good point - great forward thinking. I think what we can do instead is no longer use the static `k8s` tag for the proxy, the reason we used this was that we have additional tags `k8s-preprod` and `k8s-release` that we use for our other internal environments. Now with the above changes we can sync the proxy to the app version instead as all configuration can be injected as environment variable. e.g. `1.2.47` tag will be used now as we do with app and worker service - this has a big benefit of being able to ensure the correct proxy version is used with the services and facilitates rollbacks. 

I've pushed the change to this branch and updated the base for this PR. 
https://github.com/Budibase/budibase/pull/7413

I'll merge this PR now into that branch and I'll take care of testing and rolling out on our side. I'll update you when this has been released to self hosters. Thank you very much for contributing this change, it'll be a big benefit!

Issue #7050, opened 2022-08-03
Error getting user groups


aderici:
version 1.2.4, running through cli, the "users" screen wont allow me to click on a user to edit its details. It makes a request to api/roles endpoint and gets a 500 with this message : 

{
  "message": "Missing/invalid DB name",
  "status": 500
}

any way to fix this?

aderici:
magically started working, in the meanwhile what i did was, i created my first app, and i logged in through an openid provider

aptkingston:
Hey @aderici. As you noticed - the fix is creating at least one app. I have done a real fix for this issue though which will be deployed soon, and thanks for reporting :+1:

Issue #7043, opened 2022-08-02
Groups issues


aptkingston:
I'm going to start working on these straight away but just opening for visibility in case anyone else runs into these issues. I did a full sweep of the new features (functionality and design) with Andy and this is a full list of issues which need addressed. Some are minor but some are quite significant bugs.

## (important) Lack of data clean up
When deleting structures, data is never cleaned up. This leads to inconsistencies everywhere.
- [x] When deleting apps, app IDs are not removed from groups
- [x] When deleting groups, the group ID is not removed from the user docs in that group
- [x] When deleting users, the user ID and email address are not removed from the group doc

## Duplication of user/group source of truth
When adding users to groups, there are 2 sources of truth:
- [x] User IDs (and emails) are stored inside group docs
- [x] Group IDs are stored inside user docs

This means that both docs must be manually kept up to date with each other, and it is a certainty that this will eventually lead to inconsistency and therefore must be changed. This will create other work for things like listing all users in a group, as we will likely need to implement new views to make operations like that performant.

User email addresses should also not be stored in groups as this is duplication. We can get away with this for now as users cannot currently change their email address, but it is still bad practice.

## Access logic needs returned to user
When we determine the overall permissions for a user using both their groups and app overrides, we need to return this to the user. Currently the user UI does not accurately display the "real" app access for users, and only displays app overrides. We need to return their full permissions from the user API so we can present them in the UI properly.

This also causes data in the users table to be incorrect, as it will show 0 apps even though they may have full access to 100 apps via groups.

- [ ] return to user

## Client app permissions are broken
If I provide access to an app via group membership, then the client app does not think I have any role. The `roleId` property returned from `/api/self` always shows `null`, which results in the client library thinking there is no role set, but that they are logged in.
![image](https://user-images.githubusercontent.com/9075550/182865940-5b31929f-ce28-4be9-8dd8-4231decd4cd7.png)

The `roles` property of the user is also empty, linking in with the above issue around needing to actually return full permissions to the client.

![image](https://user-images.githubusercontent.com/9075550/182867410-f42d309c-546c-437b-8803-cfa48e8c811a.png)

- [ ] return correct permissions from `/api/self`

## Global
- [x] (important) Multiple mentions of groups are still available when groups are disabled at the platform level, but enabled in the users license

## Implementation issues
[...]

PR #7066, opened 2022-08-03
Prod user fixes

aptkingston:
## Description
This PR is a subset of the issues raised in #7043, targetting just those that affect users rather than groups - as user issues are currently live in production.

Overview of changes:
- Fix internal server error when attempting to fetch roles and app DB doesn't exist
- Fix bulk user invite not being able to assign roles
- Fix showing parts of groups UI when groups is disabled at the platform level
- Fix adding users to apps multiple times
- Fix UX issues with new dropdown picker components
- Update user list and user detail pages to match latest designs

Full change list visible in #7043 (the completed tasks).

Addresses:
- #7043
- #7050

Screenshots:
User list page:
image

User details page:
image


packages/frontend-core/src/api/user.js
```diff
@@ -141,20 +141,18 @@ export const buildUserEndpoints = API => ({
...
    * Invites multiple users to the current tenant.
-   * @param email An array of email addresses
-   * @param builder whether the user should be a global builder
-   * @param admin whether the user should be a global admin
+   * @param users An array of users to invite
    */
-  inviteUsers: async ({ emails, builder, admin }) => {
+  inviteUsers: async users => {
     return await API.post({
       url: "/api/global/users/inviteMultiple",
```
mike12345567, line 148: I didn't actually notice this in the original PR for groups, but this URL isn't great, camel casing in resources is always a bit weird since URLs aren't case sensitive, ideally I think this should probably be `/api/global/users/multi/invite` - incase there are other multi actions we want to perform around groups of users in the future.

mike12345567 (approved):
LGTM - just one quick change to the API endpoint to make it a bit more consistent! Also amazing work getting all these problems discovered and fixed out so quickly!

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -214,31 +214,33 @@ export const invite = async (ctx: any) => {
...
+    let userInfo = users[i].userInfo
+    if (!userInfo) {
+      userInfo = {}
+    }
+    userInfo.tenantId = tenancy.getTenantId()
+    const opts: any = {
+      subject: "{{ company }} platform invitation",
+      info: userInfo,
+    }
+    console.log(userInfo)
```
shogunpurple, line 242: remove `log`

shogunpurple (approved):
LGTM! Great job 👌

Issue #7081, opened 2022-08-04
Login does not work for new users


Rory-Powell:
**Hosting**
- Cloud

**Describe the bug**
- New root account holders cannot login after their initial session expired
- New internal users cannot login

Rory-Powell:
Applied migration to affected users created during the bug period:
https://github.com/Budibase/budibase/pull/7086

Further details:
- Account holder SSO users resolved
- Account holder password users: **password reset required** 
- Internal users (non account holders) resolved

PR #7086, opened 2022-08-04
Sync global-info users to fix login + prevent double password hashing

Rory-Powell:
## Description
Related to: https://github.com/Budibase/budibase/issues/7081

Double hashing: 
- Honour the params sent in the api request to create a user instead of always defaulting to `true` for hashing passwords

Fix `global-info` users:
- Add migration support to worker
- Add migration to re-create the user documents in `global-info` db
- Don't enable migrations on startup, this is only needed for cloud and will only be run against a subset of tenants using the migrations api

### DIFF
--- a/packages/worker/src/api/index.js
+++ b/packages/worker/src/api/index.js
@@ -106,7 +106,10 @@ router
     if (ctx.publicEndpoint) {
       return next()
     }
-    if ((!ctx.isAuthenticated || !ctx.user.budibaseAccess) && !ctx.internal) {
+    if (
+      (!ctx.isAuthenticated || (ctx.user && !ctx.user.budibaseAccess)) &&
+      !ctx.internal
+    ) {
       ctx.throw(403, "Unauthorized - no public worker access")
     }
     return next()
--- a/packages/worker/src/migrations/index.ts
+++ b/packages/worker/src/migrations/index.ts
@@ -1,0 +1,74 @@
+import { migrations, redis } from "@budibase/backend-core"
+import { Migration, MigrationOptions, MigrationName } from "@budibase/types"
+import env from "../environment"
+
+// migration functions
+import * as syncUserInfo from "./functions/globalInfoSyncUsers"
+
+/**
+ * Populate the migration function and additional configuration from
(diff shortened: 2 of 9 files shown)


packages/worker/src/api/index.js
```diff
@@ -106,7 +106,10 @@ router
     if (ctx.publicEndpoint) {
       return next()
     }
-    if ((!ctx.isAuthenticated || !ctx.user.budibaseAccess) && !ctx.internal) {
+    if (
+      (!ctx.isAuthenticated || (ctx.user && !ctx.user.budibaseAccess)) &&
```
mike12345567, line 110: The removal of the brackets around `(!ctx.isAuthenticated || !ctx.user.budibaseAccess)` is this desired? Is it not `(!ctx.isAuthenticated || (ctx.user && !ctx.user.budibaseAccess)) && ...`?

mike12345567 (approved):
LGTM - just the one comment about the change the public access control to the worker.

packages/worker/src/migrations/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export const buildMigrations = () => {
+  const definitions = migrations.DEFINITIONS
+  const workerMigrations: Migration[] = []
+
+  for (const definition of definitions) {
+    switch (definition.name) {
+      case MigrationName.GLOBAL_INFO_SYNC_USERS: {
+        // only needed in cloud
+        if (!env.SELF_HOSTED) {
```
shogunpurple, line 20: NAB: Might be easier to add this stuff to the definition rather than check for the env type in here.

shogunpurple (approved):
LGTM 👍

packages/worker/src/api/index.js
mike12345567: ignore!

packages/worker/src/migrations/index.ts
Rory-Powell: agreed, it would be nice to have a condition in the def itself. Will consider that for future improvements to get this one out

PR #7093, opened 2022-08-04
Allow builders that don't have admin access to delete apps

mike12345567:
## Description
Allow builders (not just admins) to delete apps. Very simple fix, the app destroy endpoint needs to be able to remove roles, builders and admins should be able to use these endpoints.


Rory-Powell (approved):
LGTM 👍 
NAB: Might be nice to move that (or copy so we don't need to refactor all usages now) `builderOrAdmin` function into backend-core beside the existing `adminOnly`

Issue #7112, opened 2022-08-05
OnChange of Options Picker to Set Value of Relationship Picker causes app to crash


melohagan:
**Hosting**
Self
  - Method: docker compose
  - Budibase Version: 1.2.14
  - App Version: 1.2.14

**Describe the bug**
Setting the value of a relationship picker via an options picker causes the app to crash

**To Reproduce**
1. Add an options picker with custom options source (Label with email address value)
2. Add a relationship picker (users table email field)
3. Add an onchange to the options picker to set the value of the relationship picker

**Expected behavior**
If I select an option to set the value of a relationship picker, it should wrap the value in an array if needed.

**Screenshots**
<img width="1194" alt="Screenshot 2022-08-05 at 14 48 28" src="https://user-images.githubusercontent.com/101575380/183090760-1a100b35-2e96-4065-98ac-e00be499d291.png">

aptkingston:
This is probably just because you're setting the value to a non-array value. If it's the "many" side of a reltionship then the component is a multi select which needs an array of values.

melohagan:
It was related to this discussion (forgot to link): https://github.com/Budibase/budibase/discussions/7070

It is because the value isn't being set to an array value. The problem is that it crashes the app. So it needs to be handled some way - do you think it should show a notification error toast instead and return an empty array? E.g. "Cannot set relationship picker value to non-array type"

PR #7113, opened 2022-08-05
Options picker set value for relationship picker - handle non-array values

melohagan:
## Description
If a non-array value is passed into a Relationship picker, this will now be handled and not crash the app.

Addresses: 
- https://github.com/Budibase/budibase/issues/7112


aptkingston (approved):
I'm generally not a fan of these sort of changes where we're just allowing malformed data to be used - but in this case since it actually crashes with invalid data, it makes sense to fix it. Especially since we already have the `flatten` function which is basically sanitising the value.

LGTM!

PR #7114, opened 2022-08-05
Increasing time between updating session TTLs

mike12345567:
## Description
Updating authenticated middleware to typescript and updating the TTL once per minute rather than every API request.


packages/backend-core/src/middleware/authenticated.ts
```diff
@@ -1,28 +1,39 @@
...
-  ctx,
-  { authenticated, user, internal, version, publicEndpoint } = {}
-) {
-  ctx.publicEndpoint = publicEndpoint || false
-  ctx.isAuthenticated = authenticated || false
-  ctx.user = user
-  ctx.internal = internal || false
-  ctx.version = version
+function timeMinusOneMinute() {
+  return new Date(Date.now() - ONE_MINUTE).toISOString()
```
shogunpurple, line 25: NAB: Do you think this might be nice to control through an env var with 1 min as the default? Lets us tweak it in different environments

shogunpurple (approved):
Great work @mike12345567

Rory-Powell (approved):
LGTM!

packages/backend-core/src/middleware/authenticated.ts
mike12345567: Good shout!

Issue #7115, opened 2022-08-05
Pagination on `search` endpoint always thinks there is another page


aptkingston:
Noticed this in self host and replicated on cloud.

The `search` endpoint will always return `hasNextPage: true`, leading the frontend to believe that there is another page of data. You can continue cycling through pages forever and see no results.

This isn't particularly severe as pagination does work to return the full set of results, even if they span multiple pages, but it will just never stop.

Easily reproducable by looking at any table in the data UI - they will always have a next page available even when they have 0 rows.
e.g. here is a brand new table with no data:
![image](https://user-images.githubusercontent.com/9075550/183100570-d80b3f45-f8b8-4b95-a1c0-df828fea52a0.png)

and the API response:
![image](https://user-images.githubusercontent.com/9075550/183100384-4d1b7041-b61e-4d99-b17f-d3da89b76c18.png)

PR #7117, opened 2022-08-05
Insert table ID back into query for follow-up pagination query

aptkingston:
## Description
Fixes an issue, which is live, where pagination always appears active. This includes even when there is no data in tables. The fix replaces the table ID param in the lucene query for the follow up pagination search, as the table ID is now stripped from the original query to facilitate more advanced searching with `or`.

Addresses: 
- #7115


shogunpurple (approved):
LGTM

PR #7121, opened 2022-08-05
Add separate rate limiting config for webhooks + environment variable support

Rory-Powell:
## Description
- Add separate `/api/webhooks` location + rate limiting config
- Change default rate limit status from 503 to 429
- Add support for environment variables in proxy-service image

### DIFF
--- a/hosting/nginx.prod.conf.hbs
+++ b/hosting/nginx.prod.conf.hbs
@@ -9,7 +9,11 @@ events {
 }
 
 http {
+  # rate limiting
+  limit_req_status 429;
   limit_req_zone $binary_remote_addr zone=ratelimit:10m rate=20r/s;
+  limit_req_zone $binary_remote_addr zone=webhooks:10m rate=${PROXY_RATE_LIMIT_WEBHOOKS_PER_SECOND}r/s;
+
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM!

hosting/nginx.prod.conf.hbs
```diff
@@ -126,6 +130,25 @@ http {
       proxy_pass      http://$apps:4002;
     }
 
+    location /api/webhooks/ {
```
Rory-Powell, line 133: Open to suggestions on how to reduce duplication between this location and the above `location /api/`. Nesting doesn't seem to work.

shogunpurple: You can nest locations in NGINX, but I think it's honestly just cleaner here to have the other route.

shogunpurple (approved):
LGTM! Just make sure we give this a good test with a locally built image 👍

hosting/nginx.prod.conf.hbs
mike12345567: I would agree, duplication in this case is probably easier to understand from a readability point of view.

Issue #4901, opened 2022-03-14
Session time inconsistencies


Rory-Powell:
**Describe the bug**
In budibase cloud we sometimes see sessions being invalidated and 403 responses seen prior to the 1 week expiry on the session. 

**COS**
Sessions should live for at least 1 week unless a logout has taken place.
Browser auth session cookies should not be expired and only manually deleted when a 403 is received

joebudi:
I think a month makes sense.

mike12345567:
Been looking into this issue a bit, wondering if the issue is related to something else throwing a 403, which in turn invalidates the cookie. I believe any network request which 403s from the Budibase frontend will trigger the cookie being deleted and the user being sent back to the login screen.

Was thinking that perhaps there is an issue with the template retrieval, maybe it occasionally throws a 403 for some reason, or something similar, which is triggering the session invalidation? @aptkingston wondering if you have any thoughts around this?

shogunpurple:
This issue seems to happen with non-builder/developer users

PR #7133, opened 2022-08-08
Bulk session wipe fix + logging

Rory-Powell:
## Description
Adding a fix to prevent all sessions being wiped when the requested userId is undefined. Ideally we wanted to put this fixed lower into the redis client `scan` function, however there are a number of places that currently rely on the argument being undefined to return all results. Instead move the fix up one call into the `getSessionsForUser` function. 

Addresses: 
Fixes: https://github.com/Budibase/budibase/issues/4901


aptkingston (approved):

mike12345567 (approved):

shogunpurple (approved):
LGTM!

PR #7151, opened 2022-08-08
Fixing CLI CI build

mike12345567:
## Description
Updating self host release to fix issues with CLI build - with the update to support backups via the CLI we have imported usage of the backend-core, however this means that the whole monorepo needs to be bootstrapped and setup, including installing pro (used by backend-core).

This will be necessary for dev experience CLI features as we will need the backend-core for them too - so updating the self host build to allow building the full monorepo is necessary.


shogunpurple (approved):
LGTM!

PR #7154, opened 2022-08-08
Rate limit served:X events

Rory-Powell:
## Description
Rate limiting the following events to a maximum of one per day 
- `served:builder`

Additionally limit to once per day per app
- `served:app`
- `served:app:preview`


shogunpurple (approved):
LGTM - nice work rory. Should keep the events down significantly now 👍

PR #7176, opened 2022-08-09
Further redis TS conversion

Rory-Powell:
## Description
- Add additional TS conversions for redis
- Add `ioredis` types and integrate with `RedisWrapper`


Rory-Powell (commented):
@mike12345567 the only change I'm concerned about in this one is the rename of `delete` to `deleteKey` in `cache/generic`. This had to be done as `delete` is a restricted function name. I think I've covered the usages but I'm not 100% sure, would be grateful if you could take a quick look to make sure there aren't any missed ones

packages/backend-core/src/cache/generic.ts
```diff
@@ -0,0 +1,53 @@
+import BaseCache from "./base"
+
+const GENERIC = new BaseCache()
+
+export const CacheKeys = {
```
mike12345567, line 5: Could possibly convert this and the TTL into real TS enums.

mike12345567 (approved):
LGTM!

mike12345567 (changes requested):
Actually - had a think about the `deleteKey` changes - this will be a pretty huge change as the redis clients are returned from the `init` file and used all over the place.

For example have a look at `packages/backend-core/src/cache/user.js` - line 61 has a cache delete using the `redis.getUserClient()` function. Not sure how to account for this - changing usage of this could be quite difficult.

mike12345567:
Closing as superseded by complete typescript refactor.

PR #7194, opened 2022-08-10
Trigger served events on ping endpoint

Rory-Powell:
## Description
- Update served events to use an endpoint trigger instead

Why
- Triggering on static files served is prone to redirect loops / spamming
- Adding any additionally information to the event via the client side is impossible 

Changes
- Remove emit based event handling - currently has a bug where context is being wiped after the initial request finishes
- Re-add `/api/bbtel/ping` endpoint
   - Trigger from builder and client + `source` field in body to distinguish


shogunpurple (approved):
LGTM - just one thing, what does this mean for old events? Are they essentially not relevant anymore because they have been skewed by over-emitting?

Rory-Powell:
> LGTM - just one thing, what does this mean for old events? Are they essentially not relevant anymore because they have been skewed by over-emitting?

For the served:builder events they don't align with the pageview events so I'm not sure how useful they are, better to rely on the pageview event for that data. 

The app / app preview events should be accurate for both typed of hosting. However given that we are now rate limiting the served events the historical data will never match up. One thing that will be accurate on the historical events is unique user filter which brings the counts down to consistent and accurate looking levels

PR #7197, opened 2022-08-10
Add timezone to served events

Rory-Powell:
## Description
- Pageview events have been removed due to volume
- To some awareness of location, add timezone to served events instead 
- Note: this is not perfect, e.g. belfast based timezone is `Europe/London`, however it is good enough to give us an indication of where requests are coming from in the cloud


shogunpurple (approved):
LGTM

Issue #6279, opened 2022-06-11
csv export table/view doesn't include relationship columns


HeronCS:
**Hosting**
- Cloud
  - Tenant ID: I'd rather not disclose

**Describe the bug**
csv export table/view doesn't include relationship columns

**To Reproduce**
csv export any table/view with a relationship column

**Expected behavior**
to have all data included in csv export

Shadowfita:
Bump. Primary display column of relationship would be ideal.

NEOLPAR:
Unfortunately, I wasn't unable to replicate this issue. However, I fixed an issue where it wasn't possible to export a table from a datasource different than budibase in cloud mode.

If you could still experience this issue, could you please give us more details to replicate it? Thanks!

PR #7200, opened 2022-08-10
fixed pouchdb unable to find the view throwing error

NEOLPAR:
## Description
Exporting datasource other than budibase in cloud mode was throwing an error due to pouch couldn't find the view.
Exporting datasource includes the relationship column, couldn't replicate the original issue.

Addresses: 
- Fixing error #6279 

**BEFORE**
Screenshot 2022-08-08 at 13 52 36

**AFTER**
"helloWorldId" column is a relationship with "helloworld" table, and it shows the id related.
Screenshot 2022-08-10 at 13 23 19


packages/server/src/api/controllers/view/utils.js
```diff
@@ -20,8 +20,13 @@ exports.getView = async viewName => {
       return null
     }
 
-    const viewDoc = await db.get(generateMemoryViewID(viewName))
-    return viewDoc.view
+    try {
+      const viewDoc = await db.get(generateMemoryViewID(viewName))
+      return viewDoc.view
+    } catch (err) {
```
Rory-Powell, line 26: In these scenarios I always like to check the `err.status === 404` to be sure that this was the cause, in case there are any other unknown errors around reading from couch at the time

packages/server/src/api/controllers/view/utils.js
```diff
@@ -20,8 +20,15 @@ exports.getView = async viewName => {
...
+      // Return null when PouchDB doesn't found the view
+      if (err.status === 404) return null
```
Rory-Powell, line 28: Small style one, we usually always use 
```js
if (something) {
  // do stuff
}
```
style for if statements

Rory-Powell (approved):
Looks good, just a point on `if` block styling

PR #7210, opened 2022-08-10
onEmptyFilter selection for automations

shogunpurple:
## Description
This PR contains code that allows you to specify the behaviour of the query rows automation step whenever a `null` filter is encountered. The reasoning for this is that couch will return all results of the table if you filter on a value of `null`, rather than nothing. This is particularly dangerous if you plan to delete rows or perform actions on a loop. 

**Empty Filter Setting**
I've opted for a "When Filter Empty" option in the query rows automation. I felt this was more descriptive and clear than just a checkbox.

Screenshot 2022-08-10 at 15 52 40

The 2 options work as follows:

**Return all table rows**
When a user passes no filters, or a value is null, just retain the old behaviour. This maintains backwards compatibility and still lets you account for use cases such as `not equals null`. 

**Return no rows**
When no filters are passed, or any filter is passed that has a `null` value - simply return an empty array, and a success value of `false`. This is a safeguard for users who want to ensure that all their filters are populated before performing a query to the database. This is a good option to use if your next step is `Delete Row` or any other destructive operation. This should probably be the default - but I didn't want to break any backwards compat for existing automation users.

Addresses: 
- https://github.com/Budibase/budibase/issues/6013

## Screenshots

### DIFF
--- a/packages/server/src/automations/steps/queryRows.js
+++ b/packages/server/src/automations/steps/queryRows.js
@@ -14,6 +14,16 @@ const SortOrdersPretty = {
   [SortOrders.DESCENDING]: "Descending",
 }
 
+const EmptyFilterOptions = {
+  RETURN_ALL: "all",
+  RETURN_NONE: "none",
+}
+
+const EmptyFilterOptionsPretty = {
+  [EmptyFilterOptions.RETURN_ALL]: "Return all table rows",
+  [EmptyFilterOptions.RETURN_NONE]: "Return no rows",
+}
+
 exports.definition = {
   description: "Query rows from the database",
   icon: "Search",
@@ -51,6 +61,12 @@ exports.definition = {
           type: "number",
           title: "Limit",
           customType: "queryLimit",
+        },
+        onEmptyFilter: {
+          pretty: Object.values(EmptyFilterOptionsPretty),
+          enum: Object.values(EmptyFilterOptions),
+          type: "string",
+          title: "When Filter Empty",
         },
       },
       required: ["tableId"],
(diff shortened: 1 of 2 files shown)


packages/server/src/automations/steps/queryRows.js
```diff
@@ -103,6 +119,9 @@ function typeCoercion(filters, table) {
   return filters
 }
 
+const hasNullFilters = filters =>
+  filters.length === 0 || filters.some(filter => filter.value === null)
```
aptkingston, line 123: The main reason for the current behaviour is that empty strings are the biggest problem - so those are important ones to filter out here. We should update this to `filter.value == null || filter.value === ""` to ensure we account for those.

aptkingston (commented):
Just one change needed to account for empty strings but otherwise looks good :+1:

packages/server/src/automations/tests/queryRows.spec.js
```diff
@@ -48,4 +48,71 @@ describe("Test a query step automation", () => {
     expect(res.rows.length).toBe(2)
     expect(res.rows[0].name).toBe(NAME)
   })
+
+  it("Returns all rows when onEmptyFilter has no value and no filters are passed", async () => {
+    console.log(rows)
```
melohagan, line 53: console log

packages/server/src/automations/steps/queryRows.js
```diff
@@ -127,9 +146,21 @@ exports.run = async function ({ inputs, appId }) {
...
+
+    if (
+      inputs.onEmptyFilter === EmptyFilterOptions.RETURN_NONE &&
+      inputs["filters-def"] &&
+      hasNullFilters(inputs["filters-def"])
+    ) {
+      rows = []
+    } else {
+      await rowController.search(ctx)
+      rows = ctx.body ? ctx.body.rows : []
```
melohagan, line 159: could this be `ctx.body?.rows || []`

melohagan (approved):

packages/server/src/automations/steps/queryRows.js
shogunpurple: Unfortunately we can't use `body?` on the backend due to our node version

PR #7143, opened 2022-08-08
filtering datasources for execute query actions

NEOLPAR:
## Description
Fixed execute query returning all datasources instead of the allowed PostgreSQL and MongoDB.

Addresses: 
- fixing issue #6342 

**BEFORE**
image

**AFTER**
image

PR #7213, opened 2022-08-10
fixing wrong filter, excluding just bb_internal

NEOLPAR:
## Description
Fixing wrong filter in case #7143 , now it is just excluding budibase datasource


Rory-Powell (approved):
LGTM 👍

Issue #7206, opened 2022-08-10
Webhook Automation - Finish and test, error saving automation


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7132

<div type='discussions-op-text'>

<sup>Originally posted by **poirazis** August  7, 2022</sup>
Self Hosted V.1.2.20 on Latest Debian / multi image docker.

Scenario : I have a contact form in Wordpress i want to submit to a budibase app.

I have the form ready in WP.
I create a new Webhook.
Click SetUp Webhook
Copy the URL , update my WP form and send a submission. 
Budibase detects the submission telling me there are 13 bindable values

Next add a Create Row action to the chain.
I choose my (my_sql) table
I set fields to bindings i.e name => {{ trigger.name }}

I click "finish and test"
Once I start entering values into the fields, i get a toast saying "Error saving automation" 
<img width="541" alt="Screenshot 2022-08-07 at 10 03 33 PM" src="https://user-images.githubusercontent.com/80512170/183306993-3d0f1f0f-d787-49ba-8618-0199e08dca81.png">

If you publish after this , the automation doesn't get triggered on calls to the webhook.

this setup used to work btw , already caught a few submissions. 
</div>

PR #7215, opened 2022-08-10
Webhook Automation - Finish and test, error saving automation

melohagan:
## Description
Error saving caused by null pointer

Addresses: 
- https://github.com/Budibase/budibase/issues/7206


Rory-Powell (approved):
Nice one mel

PR #7226, opened 2022-08-11
single image build data path for azure

jonnymccullagh:
The aim of this PR is to assist with the single image build for Azure App Service (AAS). AAS requires that data is stored under /home rather than /data. It also requires SSH server installation listening on port 2222.
So these changes add variables into config files to replace DATA_DIR with either /data or /home.

### DIFF
--- a/hosting/single/Dockerfile
+++ b/hosting/single/Dockerfile
@@ -20,10 +20,10 @@ RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
 
 FROM couchdb:3.2.1
 # TARGETARCH can be amd64 or arm e.g. docker build --build-arg TARGETARCH=amd64
-ARG TARGETARCH amd64
+ARG TARGETARCH=amd64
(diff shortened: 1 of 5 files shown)


hosting/single/Dockerfile
```diff
@@ -20,10 +20,10 @@ RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
 
 FROM couchdb:3.2.1
 # TARGETARCH can be amd64 or arm e.g. docker build --build-arg TARGETARCH=amd64
-ARG TARGETARCH amd64
+ARG TARGETARCH=amd64
```
jonnymccullagh, line 23: this is the correct syntax for ARG - it should have an equals sign

mike12345567 (approved):
LGTM - nice to get all of this dynamically controlled through that `DATA_DIR` var!

shogunpurple (approved):
LGTM, nice work on this. Seems quite a few people are using this image.

PR #7227, opened 2022-08-11
General @budibase/types update

mike12345567:
## Description
For the purposes of the dev experience we need a version of the `@budibase/types` library which has all the interfaces that we use internally exposed in it.

Ideally we shouldn't need anymore types than what we use for our internal integrations (as its fairly thoroughly defined and well used etc) - I don't think its worthwhile going down the path of actually integrating the skeleton repos into the mono-repo, or integrating the mono-repo with the skeletons as this has some licensing concerns (keeping MIT licensed code in our main codebase which integrates fully with GNU licenses) as well as the skeletons not releasing regularly, they need to be relatively separate from the monorepo.

For this reason we just need to get a release of the `@budibase/types` library to NPM which we can use for the skeleton - this shouldn't need updated very regularly as we don't change these types very often (due to the breaking changes it would involve).

Once the types are defined they should be relatively constant for the plugins - changing these regularly would be very problematic and introduce a ton of breaking changes so I think its best if we just use the types as an actual NPM library, versioned and pinned as we need it.


shogunpurple (approved):
LGTM! Nice work 👍

Rory-Powell (approved):
Nice work!

Issue #6232, opened 2022-06-08
Custom colour palette for charts


melohagan:
**Describe the feature request**
I should be able to select 'Custom' from the Colours dropdown.
Then I should be given the option to pick a colour for each label - there are a total of 5 colours in a palette. 
Disable the transparent option.  

**Screenshots**
*Add 'Custom' to the top of this list?*
![Screenshot 2022-06-08 at 08 56 28](https://user-images.githubusercontent.com/101575380/172563271-0425defe-dc4b-49c6-a916-5d029b975359.png)

*Could have 5 of these circles in a row, labelled Colour 1 through 5*
![Screenshot 2022-06-08 at 09 00 18](https://user-images.githubusercontent.com/101575380/172564100-7f723df4-c86f-413d-a4ee-c73912b974aa.png)

PR #7237, opened 2022-08-11
Custom colour palette for charts

melohagan:
## Description
You can now select five custom colours instead of using a pre-set palette.
All five colours must be set to render the theme. 

Addresses: 
- https://github.com/Budibase/budibase/issues/6232

## Screenshots
Screenshot 2022-08-11 at 18 02 48

Screenshot 2022-08-11 at 18 03 10

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -1466,10 +1466,11 @@
       },
       {
         "type": "select",
-        "label": "Colours",
+        "label": "Colors",
         "key": "palette",
         "defaultValue": "Palette 1",
         "options": [
+          "Custom",
           "Palette 1",
           "Palette 2",
           "Palette 3",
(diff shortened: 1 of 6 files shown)


packages/client/manifest.json
```diff
@@ -1482,6 +1483,56 @@
           "Palette 10"
         ]
       },
+      {
+        "type": "color",
+        "label": "C1",
+        "key": "c1",
+        "barSeparator": false,
```
aptkingston, line 1490: NAB: not an issue but more an FYI - `barSeparator` only needs specified if you define `showInBar: true` which means use the settings bar that floats in the app preview. If you enable the bar, then `barSeparator: true` means render a divider after a certain setting. So basically all these usages of `barSeparator` here do nothing and so can be deleted.

aptkingston (approved):
LGTM! Nice one Mel.

packages/client/manifest.json
melohagan: Cheers, good to know! I've removed them.

Issue #5564, opened 2022-04-25
Uploaded files stay behind orphaned in minio bucket


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5552

<div type='discussions-op-text'>

<sup>Originally posted by **bjornclauw** April 23, 2022</sup>
Hi,

I upload files through REST api to a custom service.
For this I use the 'attachment' component.

>new form (type: CREATE // schema: CUSTOM)
>>new attachment

After getting my `Image.url` from my component I use a REST API query to send my image.
After the query is finished I clear my form.

Turns out the attachments stay behind in the minio database. </div>

Whenever you upload attachments, there does not seem to be a way to delete them via the UI. Would it be possible to use a checksum on the file to prevent the same file being uploaded more than once? 

**Steps to reproduce**
1. Have a form with an attachment component
2. Drag an image into the upload component, clear with the X
3. The image is still in the minio instance
4. You can continuously drag and clear the same image into the attachment component which will clog up the minio bucket

PR #7243, opened 2022-08-12
Delete attachments on field clear from minio bucket

melohagan:
## Description
Clearing an attachment field will now remove the file from MinIO

Addresses: 
- https://github.com/Budibase/budibase/issues/5564

## Screenshots
**Two files uploaded**
Screenshot 2022-08-12 at 11 41 45
Screenshot 2022-08-12 at 11 41 57

**Clear one file**
Screenshot 2022-08-12 at 11 42 26
Screenshot 2022-08-12 at 11 42 42

**Clear the last file**
Screenshot 2022-08-12 at 11 43 05
Screenshot 2022-08-12 at 11 43 16

### DIFF
--- a/packages/bbui/src/Form/Core/Dropzone.svelte
+++ b/packages/bbui/src/Form/Core/Dropzone.svelte
@@ -17,6 +17,7 @@
   export let disabled = false
   export let fileSizeLimit = BYTES_IN_MB * 20
   export let processFiles = null
+  export let deleteAttachments = null
   export let handleFileTooLarge = null
   export let handleTooManyFiles = null
   export let gallery = true
@@ -95,6 +96,7 @@
       value.filter((x, idx) => idx !== selectedImageIdx)
     )
     selectedImageIdx = 0
+    await deleteAttachments(value.map(item => item.key))
   }
 
   function navigateLeft() {
--- a/packages/bbui/src/Form/Core/Dropzone.svelte
+++ b/packages/bbui/src/Form/Core/Dropzone.svelte
@@ -95,9 +95,11 @@
       "change",
       value.filter((x, idx) => idx !== selectedImageIdx)
     )
-    await deleteAttachments(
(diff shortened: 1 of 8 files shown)


packages/bbui/src/Form/Core/Dropzone.svelte
```diff
@@ -94,6 +95,9 @@
       "change",
       value.filter((x, idx) => idx !== selectedImageIdx)
     )
+    await deleteAttachments(
```
aptkingston, line 98: We need to wrap this in a `if (deleteAttachments) {...` as otherwise this will crash when the prop doesn't exist - such as anywhere within the builder.

aptkingston (commented):
Nice one Mel! We should account for this inside the builder too, though. This PR only has an implementation when dropzones are used inside client apps, but doesn't account for the dropzones used in the builder data section.

Accounting for this anywhere a dropzone is used is a bit of a pain though, because the endpoint depends on a table ID. That means we can't build this functionality into the actual component, meaning we have to duplicate the logic for deleting attachments everywhere dropzones are used, or use another component layer on top.

The builder uses `components/common/Dropzone` to avoid duplication as much as possible, but we also use an API endpoint to upload attachments which doesn't use a table ID (`/api/attachments/process`).

To account for this in the builder we would probably need to add a builder version of the deletion endpoint which does not use a table ID, and then add usage of that endpoint in the builder dropzone file.

Even still this will only delete files when the actual dropzone component is used so it won't work for files removed by bindings or automations, but it's still much better than never deleting anything!

packages/bbui/src/Form/Core/Dropzone.svelte
melohagan: If statement added!

melohagan:
Thanks for pointing that out! I've added an endpoint for file attachments in the data section.

aptkingston (approved):

PR #7251, opened 2022-08-12
roll out release services after deployment

shogunpurple:
## Description
Some small ones just:
- Disabling "Import" on users metadata table. It's pointless and it's tripped up some users thinking that is how you upload more users to the platform
- Performing a manual rollout of release services on deployment so that they are updated until we have proper tagging and private images with ECR

Addresses: 
- ...more if required

## Screenshots


Rory-Powell (approved):
Looks good! It would also be nice if we could roll the proxy after a prod deploy, atm I think that's still manual

Issue #7257, opened 2022-08-13
The attachment name should be the same both when uploaded and downloaded


Klience:
**Describe the feature request**
The attachment name should be the same both when uploaded and downloaded.

Right now if we upload an attachment named e.g. "**Sales Agreement.docx**" it shows it's correct name inside the "Attachment" form element (sales agreement.docx).

But when clicked on that name to download, the name of the downloaded file is the internal id of that attachment, e.g. "**c73cc0bf-8610-4075-9fa3-ae9d34037ab5.docx**"

That way we have to rename it manually ourselves when downloading attachments.

aptkingston:
Great suggestion! Just leaving this here for reference, but in order to support this we would need to proxy attachment downloads through the server, so that we can internally fetch the file, then rename it and push it to the client. Proxying all attachments through the server could lead to a fair bit more load, especially for large attachments.

The better choice would be storing attachments in more uniquely named buckets so that we could just use their original file names. Including the row ID in the name of the bucket would not only let us use the original names, but also let us do things like cleaning up attachments when rows are deleted much easier.

aderici:
@aptkingston is this an option? For the frontend : 

```
<a href="path/to/file" download="renamed.txt">Download</a>
<a href="downloads/export.xlsx" download="Data-Export.xlsx">Download Export</a>
```

aptkingston:
> @aptkingston is this an option? For the frontend :
> 
> ```
> <a href="path/to/file" download="renamed.txt">Download</a>
> <a href="downloads/export.xlsx" download="Data-Export.xlsx">Download Export</a>
> ```

You learn something new every day! I had no idea that the `download` attribute can rename files. This would be perfect :+1:

Klience:
Noticed that #7270 was pushed to master. Does that mean the problem was fixed?

I updated the cloud Budibase app to 1.2.43, but the files still have internal IDs when downloaded.

aptkingston:
It's in master but we just haven't deployed it yet, but it should go out soon!

Klience:
Thanks for the heads up. I'm not a developer, so I'm not familiar with how you deploy things.

Thanks for the clarification once again!

aptkingston:
This has been deployed now, although I've just realised that the problem is only fixed for self-host, but not for our cloud. I hadn't realised that the `download` attribute only works for URLs in the same origin, and because we use a different domain for attachments in our cloud, the issue has not been fixed there. 

[...]

Issue #7249, opened 2022-08-12
Time Only Date Picker Bug


Franosa:
**Hosting**
- Cloud
  - Tenant ID: chesschristian

**Describe the bug**
When selecting a time in a date picker, the date picker decrements once after clicking to increment or decrement for the first time. It functions correctly afterwards.

**To Reproduce**
1. Click on a time only date picker
2. Increment or Decrement the time by 1

**Expected behavior**

**Screenshots**
Before Clicking Date Picker
<img width="1440" alt="1" src="https://user-images.githubusercontent.com/10144553/184395174-d6aed452-2e20-4d41-8025-42d1040ee485.png">
Clicked Date Picker
<img width="1440" alt="2" src="https://user-images.githubusercontent.com/10144553/184395192-999d346d-e5e1-4333-8525-4a14bc960c7d.png">
Clicked Down on Date Picker 1 Time
<img width="1440" alt="3" src="https://user-images.githubusercontent.com/10144553/184395220-96b95934-ce16-4e8c-8b67-d9256f3cdd35.png">
The Date Picker Decrements Another Hour Without Clicking
<img width="1440" alt="4" src="https://user-images.githubusercontent.com/10144553/184395233-1ac53c22-8f69-4890-8498-31042dc20c77.png">
After this the date picker decrements correctly

**Desktop (please complete the following information):**
 - OS: Monterey 12.5
 - Browser: Chrome 104.0.5112.79 (Official Build) (arm64)

**Additional context**
This does not happen in a Date Picker that has the calendar to choose from with time. It only happens when it is a time only date picker

Issue #7199, opened 2022-08-10
Export Data action exports all columns even when export columns selected.


zoell:
**Hosting**
- Self
  - Method: docker
  - Budibase Version: 1.2.9

**Describe the bug**
When I add the Export Data action onto a button and select columns to export in csv it is ignored, all columns are exported, including generated columns.
Not sure if this is for all data sources as currently export with DynamoDB is not working due to bug #7198 but for Budibase DB data source this is definitely the case.

**To Reproduce**
1. Add Budibase data source
2. Add some data to the table
3. Add Data Provider
4. Add Table
5. Add button
6. Add Export Data action to button
7. Select columns in the Export Data action

**Expected behavior**
When selecting columns for export in Export Data action it exports only those columns and nothing more.

aptkingston:
Thanks for reporting, I can confirm that I can see this issue too.

PR #7270, opened 2022-08-15
Various fixes

aptkingston:
## Description
A few fixes for some recent issues.

Addresses: 
- Add component requires 2 clicks if a component is not selected
- Time only datepickers are offset by 1 hour initially https://github.com/Budibase/budibase/issues/7249
- Exporting only certain columns doesn't work for internal tables https://github.com/Budibase/budibase/issues/7199
- Use original file names when downloading attachments rather than UUIDs https://github.com/Budibase/budibase/issues/7257
- Fix considering 0 to be a nullish value https://github.com/Budibase/budibase/issues/7357


packages/server/src/api/controllers/row/internal.js
```diff
@@ -375,14 +375,28 @@ exports.exportRows = async ctx => {
...
       keys: rowIds,
     })
   ).rows.map(row => row.doc)
 
-  let rows = await outputProcessing(table, response)
+  let result = await outputProcessing(table, response)
+  let rows = []
+
+  // Filter data to only specified columns if required
+  if (columns && columns.length) {
```
shogunpurple, line 390: We should cover this with a unit test - it's quite a fundamental path in the code.

shogunpurple (approved):
LGTM, and the flatpickr comment made me chuckle 😂 just need to add a unit test here to cover the columns functionality and cover us in future.

packages/server/src/api/controllers/row/internal.js
aptkingston: This was just the exact same logic that I already added to the external controller and has been deployed for quite a while - I just didn't add it to the internal controller as well. I don't think any of the export stuff is currently covered by unit tests, but I guess I'll write some. I agree though it should have had tests written in the first place.

Issue #7296, opened 2022-08-16
It's possible to create a condition that checks if a value Is In another value of type number - this breaks component preview


andz-bb:
### Discussed in https://github.com/Budibase/budibase/discussions/7248

<div type='discussions-op-text'>

<sup>Originally posted by **mmalessa80** August 12, 2022</sup>
One of my pages lists my components in the left hand list, however they don't show up on the page. If I preview the app and navigate to that page they don't show up there either.

![Screenshot 2022-08-12 9 00 30 AM](https://user-images.githubusercontent.com/106758851/184384126-1c6cda0b-4902-4597-9c51-395e2b206204.png)

</div>

**Hosting**
- Self
  - Method: docker-compose
  - Budibase Version: 1.2.26
  - App Version: 1.2.26

**Describe the bug**
With this you can create a condition where you can use an `Is in` operation against values that are not arrays. Everything generally works OK when the value you are checking against is a Binding value, but for values of type Number `Is in` is still selectable and when you do it errors in the console and then breaks your component preview on a refresh.

```
budibase-client.js?v=1.2.26:218 Uncaught (in promise) TypeError: i.split is not a function
    at budibase-client.js?v=1.2.26:218:82008
    at Array.forEach (<anonymous>)
    at Dk (budibase-client.js?v=1.2.26:218:81756)
    at budibase-client.js?v=1.2.26:243:300240
    at Array.filter (<anonymous>)
    at budibase-client.js?v=1.2.26:243:299745
    at oe (budibase-client.js?v=1.2.26:243:300280)
    at e.$$.update (budibase-client.js?v=1.2.26:243:303134)
    at _e (budibase-client.js?v=1.2.26:1:6565)
    at Le (budibase-client.js?v=1.2.26:1:6315)
```

**To Reproduce**
1. Add a component - in this case we had a button
2. Hit Configure conditions and add a new condition3. 
3. Set the condition to do something if x (number) Is in y (number)4. 
4. Refresh and get whitescreened

**Expected behavior**
I would expect the available operations to filter down to more simple ones when a value of type Number or Date is selected, the same way it does for a boolean value.
![boolean](https://user-images.githubusercontent.com/110921612/184841930-8fc52fc2-217d-4e16-8203-086ccdf39d20.png)

**Screenshots**
![broken, using Is in](https://user-images.githubusercontent.com/110921612/184838532-c0315073-6802-4518-baee-49295ff156c2.png)

[...]

PR #7303, opened 2022-08-16
Only allow bindings for 'Is In' operator in conditional UI

melohagan:
## Description
When using the 'Is In' operator within conditional ui, you can no longer select primitive types as the comparison , e.g. number, date, etc. 
You can only select binding, which will either parse a comma separated string, or accept an array returned by a JS function binding. 

Addresses: 
- https://github.com/Budibase/budibase/issues/7296

## Screenshots
Screenshot 2022-08-16 at 13 47 12

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
@@ -206,7 +206,7 @@
                 on:change={e => onOperatorChange(condition, e.detail)}
               />
               <Select
-                disabled={condition.noValue}
+                disabled={condition.noValue || condition.operator === "oneOf"}
                 options={condition.operator === "oneOf"
                   ? [
                       {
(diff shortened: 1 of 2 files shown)


packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
```diff
@@ -112,7 +112,7 @@
       Constants.OperatorOptions.NotEmpty.value,
     ]
     condition.noValue = noValueOptions.includes(newOperator)
-    if (condition.noValue) {
+    if (condition.noValue || condition.operator === "oneOf") {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte:115]: While `condition.operator` might always be up to date here (I'm not sure whether svelte will update a `bind:value` before calling an `on:change` or not) it would be safer to instead check `newOperator === "oneOf"`.

aptkingston (commented):
LGTM! Just a couple of minor changes.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte]: Good catch! Using newOperator.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
```diff
@@ -206,8 +206,15 @@
                 on:change={e => onOperatorChange(condition, e.detail)}
               />
               <Select
-                disabled={condition.noValue}
-                options={valueTypeOptions}
+                disabled={condition.noValue || condition.operator === "oneOf"}
+                options={condition.operator === "oneOf"
```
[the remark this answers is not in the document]
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte]: Good catch! I had this in before I added the disable code - helps to keep things clean! Removed.

aptkingston (approved):

PR #7310, opened 2022-08-16
fix sed apostrophe error and couchdb path

jonnymccullagh:
- sed wasn't expanding an env variable due to single apostrophe
- couch data path needed updated in local.ini
- add /home/.env to healthcheck for aas

I still have an issue when building. The app server is reporting:
```
ENOENT: no such file or directory, stat '/app/builder/index.html'
```
Possibly another unrelated build issue. But I will test again in a few days on the develop branch.


mike12345567 (approved):
LGTM! Good fixes!

Issue #4651, opened 2022-02-23
Selecting tables in general (not database specific)


kasgithub8:
We would like to select tables to fetch from a new or updated database.

With which release could this be possible?

shogunpurple:
Hello @kasgithub8 

I’m not sure what you are asking for here. 

Please provide more detail or a reproducible example?

dewdeded:
I think he want's to select which tables to fetch.

Example: you can't a database with 100 tables to Budibase, but you only want 5 or 10 of these tables in your Budibase app.

kasgithub8:
Any news?

aptkingston:
No news as we didn't really understand the issue. Can I clarify that @dewdeded is correct in that you would like the ability to choose which tables are imported when fetching tables from a database, rather than importing them all?

melohagan:
I think this is the same issue here: https://github.com/Budibase/budibase/discussions/5539

pbarmak:
Is there any update on this issue? I see it's marked as stale. We would love to be able to select which tables we fetch from an external MySQL database. Issue #5539 can also be somewhat resolved by this being fixed and refers to this issue. Thanks for any update.

melohagan:
I'm re-opening under these requirements:

As a user I should be able to list the specific tables that I want to fetch, rather than the default ALL tables found.
![Screenshot 2022-08-17 at 17 13 51](https://user-images.githubusercontent.com/101575380/185190193-b0b9ba18-e353-4566-ad26-14a450949102.png)

PR #7333, opened 2022-08-17
Allow list of named tables to be fetched

melohagan:
## Description
Formula columns were being cleared on fetch tables. By being able to specify the tables you want to fetch, you can now prevent that from happening. 
In other cases you may only want to pull a subset of your tables, although this could be solved with database level schemas and multiple datasources in Budibase.

Addresses: 
- https://github.com/Budibase/budibase/discussions/5539
- https://github.com/Budibase/budibase/issues/4651

## Screenshots
1. Fetch all tables.
2. Add a formula column to the **Owners** table.
3. Within postgres add new column to the **Appointments** table
Screenshot 2022-08-17 at 18 24 30

4. Fetch tables (Appointments, Pets, Vehicles, etc)
Screenshot 2022-08-17 at 18 27 40

5. See that **Appointments** has been updated
Screenshot 2022-08-17 at 18 26 06

6. **Owners** has been preserved
Screenshot 2022-08-17 at 18 26 30

**Toggle is off by default**
Screenshot 2022-08-17 at 18 27 03

### DIFF
--- a/packages/server/src/api/controllers/datasource.js
+++ b/packages/server/src/api/controllers/datasource.js
@@ -58,7 +58,9 @@ exports.buildSchemaFromDb = async function (ctx) {
       datasource.entities = {}
     }
     for (let key in tables) {
-      if (tablesFilter.includes(key)) {
+      if (
+        tablesFilter.some(filter => filter.toLowerCase() === key.toLowerCase())
+      ) {
         datasource.entities[key] = tables[key]
       }
     }
@@ -237,7 +239,7 @@ const buildSchemaHelper = async datasource => {
   await connector.buildSchema(datasource._id, datasource.entities)
 
   // make sure they all have a display name selected
-  for (let entity of Object.values(datasource.entities)) {
+  for (let entity of Object.values(datasource.entities ?? {})) {
     if (entity.primaryDisplay) {
       continue
     }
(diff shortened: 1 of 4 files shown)


packages/server/src/api/controllers/datasource.js
```diff
@@ -50,9 +50,21 @@ exports.fetch = async function (ctx) {
...
 
-  const { tables, error } = await buildSchemaHelper(datasource)
-  datasource.entities = tables
+  let { tables, error } = await buildSchemaHelper(datasource)
+  if (tablesFilter) {
+    if (!datasource.entities) {
+      datasource.entities = {}
+    }
+    for (let key in tables) {
+      if (tablesFilter.includes(key)) {
```
mike12345567, line 61: this should probably be case insensitive - usually users don't expect to need to define a table name exactly (as some SQL databases treat identifiers as case insensitive).

mike12345567 (approved):
LGTM - I think the one change mentioned to make this a case insensitive search could be quite helpful.

packages/server/src/api/controllers/datasource.js
melohagan: The only risk is you could technically pull down two tables when you only want one, e.g. "appointments":

Screenshot 2022-09-13 at 10 37 34

With that said maybe it's very unlikely users would have two tables with the same name except the casing.

melohagan:
Case insensitive table fetch added.

Screenshot 2022-09-13 at 10 52 45

Screenshot 2022-09-13 at 10 53 06

Issue #7267, opened 2022-08-15
OneLogin OIDC integration failing


andyburgessmd:
**Hosting**
- Cloud
  - Tenant ID: https://mistraldata.budibase.app

**Description**
When the OneLogin OIDC integration has been configured and enabled according to the instructions 
And a user tries to login via the 'Sign In with OneLogin' button on the /auth/login page
Then the browser redirects round and shows a {"message":"Unauthorized","status":403} error message and the url shows the error 'error=invalid_scope&error_description=some%20of%20requested%20scopes%20are%20not%20whitelisted'

**To Reproduce**
1. Configure an OIDC connection to OneLogin
2. Click on 'Sign In with OneLogin'

**Expected behavior**
The user should be preseneted with the OneLogin authentication (if not logged in) and returned to the redirect url set with a valid OIDC JWT.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 104

aderici:
This is probably because of the "offline_access" scope requested by budibase. Onelogin does not allow this on auth code or implicit flows.

I think you should review how you are refreshing the access token for the user. offline access scope is for the cases where you need to access users resources when the user is not using the app (when they are sleeping). I think there is another method to refresh when the user is actively using the app

andyburgessmd:
I see the https://github.com/Budibase/budibase/pull/7349 feature has been merged to develop. Do you have any eta on when it might make it to prod/live?

Rory-Powell:
hey @andyburgessmd! Those changes will become live in the next week or two, pending or QA cycle.

mlomax:
@deanhannigan we were very excited that you'd added scope. It hasn't worked for us though and we get 
![image](https://user-images.githubusercontent.com/105858177/189922930-5f276d51-cd3b-48bd-896d-d9274a8de0b3.png)
 We can log on with Azure SSO when just using the default openid scope but not when putting our bespoke one in. Is there anything we can help on with the testing? thanks Mary

aderici:
@mlomax Hi, unfortunately that 403 error is a bit generic and you can get more descriptive error message about what's wrong by logging into the inspect window or the pm2 logs in the docker container if self hosting. For microsoft, which scope are you looking to get access to? Some scopes require "admin consent" so a regular user won't be able to login before an active directory admin logs in first.

mlomax:
[...]

PR #7349, opened 2022-08-18
Feature/configurable sso auth scopes

deanhannigan:
## Description
Allows customisation of the OIDC OAuth scopes. Any modifications will take effect on a users next login.

- The 'openid' scope will remain as a minimum
- The user can revert to the core scopes that we currently send as a default : `openid`, `profile`, `offline_access`, email
- Bear in mind, depending on the vendor, `offline_access` is normally used to request a refresh token during authentication. If we do not hold a refresh token for a user, we will not be able to update their access tokens.

Addresses: 
- [OneLogin OIDC integration failing](https://github.com/Budibase/budibase/issues/7267)

## Screenshots
Screenshot 2022-08-18 at 15 03 00


packages/builder/src/pages/builder/portal/manage/auth/index.svelte
```diff
@@ -392,15 +408,135 @@
...
+                  e.code == "Space" ||
+                  e.keyCode == 32
+                ) {
+                  let scopes = providers.oidc.config.configs[0]["scopes"]
+                    ? providers.oidc.config.configs[0]["scopes"]
+                    : [...defaultScopes]
+
+                  let update = scopesFields[0].inputText.trim()
+
+                  if (/[\\"\s]/.test(update)) {
```
shogunpurple, line 465: NAB: I'd pull this regex into a constant to just make it more descriptive

packages/builder/src/pages/builder/portal/manage/auth/index.svelte
```diff
@@ -392,15 +408,135 @@
...
+                    scopesFields[0].error =
+                      "Auth scopes cannot contain spaces, double quotes or backslashes"
+                    return
+                  } else if (scopes.indexOf(update) > -1) {
```
shogunpurple, line 469:
```suggestion
                  } else if (scopes.includes(update)) {
```

packages/builder/src/pages/builder/portal/manage/auth/index.svelte
```diff
@@ -392,15 +408,135 @@
...
+                    return
+                  } else if (!update.length) {
+                    scopesFields[0].inputText = null
+                    scopesFields[0].error = null
+                    return
+                  } else {
+                    scopesFields[0].error = null
+                  }
+
+                  if (scopes.indexOf(update) == -1) {
```
shogunpurple, line 480:
```suggestion
                  if (!scopes.includes(update)) {
```

packages/builder/src/pages/builder/portal/manage/auth/index.svelte
```diff
@@ -392,15 +408,135 @@
...
+      <Layout gap="XS" noPadding>
+        <Heading size="XS">
+          <div class="auth-scopes">
+            <div>Advanced</div>
+            <Button
+              secondary
+              newStyles
+              size="S"
+              on:click={() => {
+                providers.oidc.config.configs[0]["scopes"] = [...defaultScopes]
```
shogunpurple, line 427: Why do we always need to use the `0` index here? Just wondering if there's a potential for a breakage here since it's just accessing the first element regardless

shogunpurple (approved):
LGTM - just some small suggestions and a question about the indexing. Didn't realise this would include a nice tags UX as well - great job 👌

packages/builder/src/pages/builder/portal/manage/auth/index.svelte
deanhannigan: We have potential support for having multiple OIDC configs at once, but the screen itself seems to be intended to allow only one at the minute. My assumption was that work to display multiple configs was parked as some point.

Rory-Powell: That's right, the original backend was designed to be able to add multiple oidc configs if desired, mostly for backwards compatibility on the document schema if someone ever did want to add multiple auth sources. The UI is coded to work with one, but it would be possible to have multiple in future with some refactoring.

Issue #7323, opened 2022-08-17
Automations run only with Significant Delay


3699n:
### Discussed in https://github.com/Budibase/budibase/discussions/7293

<div type='discussions-op-text'>

<sup>Originally posted by **3699n** August 16, 2022</sup>
I have created several on "New Row" and "Update Row" automations. 

- One automation sends an email when a new row is added
- The second automation makes a REST GET request when a new row is added and updates the row with additional data
- The third automation does the same as the first, but on "row update"

When running the tests, the automations work flawlessly. However, once published, the automations only actually execute hours after the respective rows are added or updated. What could be causing this issue? 

I'm using BudibaseDB
When I look at my automation history, all of the automations run succesfully, but only a day later at 07:30
</div>

adamjdavies:
I've noticed this today too. My automation runs instantly when testing, but has yet to actually run when triggered on the published version of my app.

kasgithub8:
Same problem here, my automations get executed at 14.40 Central European Summer Time.

Rory-Powell:
Hi all, could you each confirm if you are using budibase cloud or self hosting please

kasgithub8:
> Hi all, could you each confirm if you are using budibase cloud or self hosting please

My team is self hosting.

adamjdavies:
> Hi all, could you each confirm if you are using budibase cloud or self hosting please

I'm using Budibase cloud. 

My automation triggers when a new row is created in a Budibase table. It then executes an external data connector to push this new data to an airtable base.

3699n:
I'm using Budibase cloud

Rory-Powell:
@3699n @adamjdavies could you confirm the tenant id of your budibase cloud instances?
that'll be the part in the url like <tenantid>.budibase.app

adamjdavies:
@Rory-Powell  - mine is **tern**.budibase.app

kasgithub8:
Any updates on this matter? It just came out of nowhere and Automations are really important!

Rory-Powell:
Hi @kasgithub8, we're looking into the issue. Thanks for your patience. The issue stems from a delay in automations being processed from the main redis queue.

kasgithub8:
> Hi @kasgithub8, we're looking into the issue. Thanks for your patience. The issue stems from a delay in automations being processed from the main redis queue.

Oh okay. Thanks for the answer tho!

Rory-Powell:
[...]

PR #7374, opened 2022-08-19
Add logging to all event listeners for automations + stalled handling

Rory-Powell:
## Description
- Add logging listeners for all bull lifecycle events
- Add stalled handling to cancel stalled cron jobs using the same mechanism we use for crons that fail more than 5 times
   - This is commented out until the content of the stalled event data can be confirmed using the logging above
 
Supporting: https://github.com/Budibase/budibase/issues/7323

### DIFF
--- a/packages/server/src/automations/listeners.ts
+++ b/packages/server/src/automations/listeners.ts
@@ -1,0 +1,78 @@
+import { Queue, Job, JobId } from "bull"
+import { AutomationEvent } from "../definitions/automations"
+import * as automation from "../threads/automation"
+
+export const addListeners = (queue: Queue) => {
+  logging(queue)
+  // handleStalled(queue)
+}
+
+const handleStalled = (queue: Queue) => {
+  queue.on("active", async (job: Job) => {
+    await automation.removeStalled(job as AutomationEvent)
+  })
+}
+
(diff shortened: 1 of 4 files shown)


packages/server/src/automations/listeners.ts
```diff
@@ -0,0 +1,78 @@
+import { Queue, Job, JobId } from "bull"
+import { AutomationEvent } from "../definitions/automations"
+import * as automation from "../threads/automation"
+
+export const addListeners = (queue: Queue) => {
+  logging(queue)
+  // handleStalled(queue)
```
Rory-Powell, line 7: Temp commented out - awaiting logging results

packages/server/src/threads/automation.ts
```diff
@@ -465,3 +472,15 @@ export function execute(input: AutomationEvent, callback: WorkerCallback) {
     }
   })
 }
+
+export const removeStalled = (input: AutomationEvent) => {
```
Rory-Powell, line 476: This isn't thread related but this is the easiest way to integrate with the existing cron disabling. We should consider moving the orchestrator / automation specific logic out of the thread class in future into the top level automations package

PR #7383, opened 2022-08-20
Serve builder preview via server to fix dependency on third party cookies

aptkingston:
## Description
This PR updates how the builder preview is served. Previously the builder contained an HTML template which was passed to the iframe by way of the `srcdoc` attribute. This caused the iframe to be unable to properly share the same origin as the page, causing all cookies to be treated as third-party.

This PR now serves the builder preview via the server, which allows the preview to share the domain of the parent page and therefore properly share cookies and treat them as first-party.

Addresses: 
- https://github.com/Budibase/budibase/discussions/7324


shogunpurple (approved):
Awesome work @aptkingston - this is a nice dependency to remove. I was just thinking too as I read through this - we could probably solve the airgapped problem here too, by just bundling all the dependencies that are currently fetched from the internet:
- Inter font
- remix icon (do we still use this?)
- Source sans pro

If we load these from the local env and just have NGINX serve them, then budibase should, in theory, work inside an airgapped environment. You won't be able to access templates but apart from that it should at least work

aptkingston:
Yep it hopefully wouldn't be too hard to do that, at least for those assets.

Both the builder and client currently load Source Sans Pro from Google Fonts.
Remix icon is used by the icon picker component only... It's pulled from `jsdelivr.net` for client apps, and the builder bundles it itself. Would love to just remove this and make the icon picker use spectrum icons instead, but that's a breaking change.
Pretty sure inter is never used so could be stripped.

Airgap support is probably worthy of its own PR though, since I'm sure other issues will crop up with it. Are you happy to merge this as a standalone PR and we can do a separate one for airgap support? I just think there will be more to it than bundling a couple static assets.

shogunpurple:
All good @aptkingston - makes sense. Let's not blow the scope of this one up and we can address it later on 👍

Issue #7398, opened 2022-08-22
Import Users with CSV - Problem with Carriage Return


R2bEEaton:
When importing users via CSV file, you need to create a CSV file that includes a valid email on each line. If you do this in Windows or any text editor that automatically adds the carriage return `\r` character to newlines (`\r\n`), this will result in a problem when splitting. On the user's end, it makes it look like a trailing space has been included on each line, which is not the case.

More discussion at #7397.

PR #7399, opened 2022-08-22
Import Users CSV Carriage Return Support #7398

R2bEEaton:
## Description
The ImportUsersModal takes in a CSV file expecting each line to have a valid email address. When this file is created in Microsoft Excel, Windows in general, or a text editor that uses the carriage return character `\r`, the import fails. This is because the CSV string is split into an email list by `csvString.split("\n")` which fails to account for `\r\n` newlines. This PR changes the split to be the regex `/\r?\n/` which will split on either `\n` or `\r\n`.

Addresses: 
- Issue #7398
- Discussion #7397


R2bEEaton:

PClmnt (approved):
LGTM @R2bEEaton. Thanks for the contribution!

Issue #5818, opened 2022-05-11
[Builder / Core] Allow Custom CSS to be set based on Conditional UI


poirazis:
It would be most helpful, if we could set the Custom CSS property of an element in a Conditional UI filter.

aptkingston:
I agree! We should definitely add this.

poirazis:
Erhm ... Please let not this one stale!

PR #7407, opened 2022-08-22
Update CSS conditional UI

melohagan:
## Description
You can now set the custom css of a component conditionally.

Addresses: 
- https://github.com/Budibase/budibase/issues/5818

## Screenshots
Initial custom css:
Screenshot 2022-08-22 at 16 48 44

Condition:
Screenshot 2022-08-22 at 16 49 57

Update state on button click:
Screenshot 2022-08-22 at 16 50 26

Outcome:
update_state

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ConditionalUIDrawer.svelte
@@ -56,7 +56,11 @@
   ]
 
   let dragDisabled = true
-  $: settings = getComponentSettings($selectedComponent?._component)
+  $: settings = getComponentSettings($selectedComponent?._component)?.concat({
+    label: "CSS",
+    key: "css",
+    type: "text",
+  })
   $: settingOptions = settings.map(setting => ({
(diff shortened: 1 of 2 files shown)


aptkingston (commented):
I'm actually quite surprised that this works! Updating the raw `instance` property here won't trigger any svelte invalidation, and therefore won't update the component store, meaning that `$component.styles` should be unchanged. Something else must be triggering the invalidation - probably the set state action.

The other thing is that we shouldn't update the instance directly here, as it is the source of truth for the raw settings. Applying the update like this means that there's no going back - like if you cleared state again, this style would not be cleared.

The way we handle component settings is that we generate a set of additional overrides needed by conditional UI and apply them on top of the normal settings, but without changing the normal settings. That way we can always just not apply them whenever the condition is not met.

We would need to do something similar here, where we would generate a separate field for conditional UI custom CSS and then override it on top of the base CSS when required.

Might be better for me to jump on and do this since we're right in the heart of the client library here and there are a few extra considerations!

melohagan:
Definitely, thank you! I look forward to trying to wrap my head around the final solution 😀

aptkingston:
Finally got round to taking a look at this and opened a new PR (https://github.com/Budibase/budibase/pull/7644) because I extended this to make custom CSS able to use data bindings as well!

PR #7411, opened 2022-08-23
Add preview endpoint to prod nginx conf

aptkingston:
## Description
Forgot to add the new `/preview` endpoint for serving the builder preview to the prod nginx conf.

### DIFF
--- a/hosting/nginx.prod.conf.hbs
+++ b/hosting/nginx.prod.conf.hbs
@@ -85,6 +85,10 @@ http {
       proxy_pass      http://$apps:4002;
     }
 
+    location /preview {
+      proxy_pass      http://$apps:4002;
+    }
+
     location = / {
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
LGTM - we should really consolidate these configs 😅

aptkingston:
> LGTM - we should really consolidate these configs sweat_smile

Yea absolutely! The vast majority of the content is the same and we could just use more conditionals like we do for watchtower to add dev/prod-only endpoints.

Rory-Powell:
> > LGTM - we should really consolidate these configs sweat_smile
> 
> Yea absolutely! The vast majority of the content is the same and we could just use more conditionals like we do for watchtower to add dev/prod-only endpoints.

This PR will go a long way in helping us achieve this: https://github.com/Budibase/budibase/pull/7065
I'll follow up with this one today

PR #7413, opened 2022-08-23
Allow proxy upstream URLs to be configured at runtime

Rory-Powell:
## Description
Supporting:
https://github.com/Budibase/budibase/pull/7065
https://github.com/Budibase/budibase/pull/7065#issuecomment-1218132153

> The only thing is that these changes are not backwards-compatible for existing Helm chart deployments, since as soon as the new proxy docker image is published, existing deployments will pick it up (on next pod re-create). But the new docker image defaults to upstream URLs as used by the docker-compose deployment. So it will break.

> If this is a concern (it should, I think), it would be possible to pass env defaults in as build arguments and this way having different defaults for the k8s and the docker-compose images. Like in the following diff. What do you think?

In the above PR we can define all environmental changes to proxy upstreams as environment variables, keeping environment specific tags such as `k8s`, `k8s-release` and `k8s-preprod` are no longer required. This solves the problem highlighted in the comment above.


panos--:
Hi there! Any news on this one?

Rory-Powell:
> Hi there! Any news on this one?

Hi @panos--. this should be going out with the january release

hosting/proxy/.generated-nginx.prod.conf
```diff
@@ -0,0 +1,186 @@
+user                    nginx;
```
Rory-Powell, line 1: Should this file be deleted?

Issue #7060, opened 2022-08-03
Nest Bindings Contextually in Binding List in Drawer


shogunpurple:
**COS**
- Nest the bindings under their respective parent in the bindings list in drawer. Eg. My Data Provider and children such as .Rows

PR #7414, opened 2022-08-23
Feature/binding ux update

deanhannigan:
## Description
Group all available options in the binding drawers by category and allow the display of detailed descriptions for each with a popover.

Addresses: 
- https://github.com/Budibase/budibase/issues/7060

## Screenshots

Main Categories
Screenshot 2022-08-23 at 09 38 32

Schema types shown where available
Screenshot 2022-08-23 at 10 29 56

Filtering results grouped
Screenshot 2022-08-23 at 10 32 29

Mouseover Prompts for the Helpers.
Screenshot 2022-08-23 at 10 27 44


joebudi:
I didn't realize how much I wanted this until I saw this PR! Nice one @deanhannigan

aptkingston (commented):
Super work Dean and a massive improvment to that binding list! Having types available is a huge win too.

The main thing I'd like to change here (and I think the original design) is grouping bindings under their individual components that provide bindings, rather than group all context bindings under a single "Dataprovider" header. I've detailed that more in my other comments.

PR #7422, opened 2022-08-23
Mongodb objectid relationship

melohagan:
## Description
The objectid parsing/casting was too specific.
Using *startsWith* "objectid" to handle all cases of ObjectId including relationship fields.

Addresses: 
- https://github.com/Budibase/budibase/discussions/7375


PClmnt (approved):
LGTM! Much cleaner way to match, nice one.

Issue #7152, opened 2022-08-08
Importing app does not preserve Navigation


B-z-F:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.1.17

**Describe the bug**
When I create an app, modify Navigation (Position: Vertical; Logo: off; custom Background and Text color; add some links) then after export and "Create new app" -> "Import app" the new app has default Navigation (Position: Horizontal; Logo: on; default Background and Text color; no links)

aderici:
Yeah, i can confirm after having to restore a backup into a new system. I hope thats the only thing that will be lost

PR #7450, opened 2022-08-25
App navigation and screen migration for imports and templates

aptkingston:
## Description
This PR properly migrates app navigation settings and legacy screen custom layouts when importing apps or using templates. Currently all our templates were made before the new design UI, meaning they have custom layouts and old navigation settings. As soon as you use a template you're shown a deprecation notice and you have to update all your screns, which is poor UX.
This also fixes a few other issues with imports and templates, where things like theme settings were always lost on import.

TLDR:
- Apps are automatically migrated, if required, when imported/created from template
- New navigation settings are automatically guessed from legacy navigation layouts
- All screens are automatically "detached from custom layouts". No more deprecation notice as soon as you import an old app or use a template. (Note: deprecation notice can still exist for existing apps which have not been migrated)
- All this logic is done server side now rather than in the frontend
- Updated some types for app metadata which were missing

### Navigation setting migration
When creating a new app, the default navigation settings are added.
When creating an app from a template or importing an app, we attempt to import the `navigation` metadata if it exists. If it does not exist, then the template/import is for an app built before the new design UI and needs migrated.
The server will then migrate navigation settings from the existing navigation layout into the new app metadata navigation structure.

### Legacy screen migration
Currently in the new design UI, if you are looking at an "old" screen that uses a custom layout you will be greeted with a deprecation notice:
image

When importing old apps or using a template, all your screens will show this button and you'll need to go through every one and update it manually. This is poor UX, so this PR now performs this migration server-side for all screens. 

Addresses: 
- #7152


shogunpurple (approved):
LGTM! Although I think a unit test (even a snapshot test) here to test the mapping would be beneficial

aptkingston:
Yea I totally agree about the tests. I checked the current app tests because I wanted to add one to cover this, but they're pretty much empty tests. Maybe I can test the migration function in isolation though.

PR #7475, opened 2022-08-26
Update NGINX access log to show upstream_response_time, proxy_host and upstream_addr

Rory-Powell:
## Description
Append to the nginx access log:
```
 response_time=0.090 proxy_host=host.docker.internal:4002 upstream_addr=192.168.65.2:4002
```
To aid debugging


shogunpurple (approved):
Nice! We can make some really nice graphs with this in grafana/ELK

PR #7507, opened 2022-08-30
Conditionally enable elastic apm

Rory-Powell:
## Description
Add configuration support for elastic apm in server and worker

### DIFF
--- a/packages/server/src/environment.js
+++ b/packages/server/src/environment.js
@@ -13,10 +13,7 @@ function isJest() {
 }
 
 function isDev() {
-  return (
-    process.env.NODE_ENV !== "production" &&
-    process.env.BUDIBASE_ENVIRONMENT !== "production"
-  )
+  return process.env.NODE_ENV !== "production"
 }
 
 function isCypress() {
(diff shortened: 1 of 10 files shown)


packages/server/src/environment.js
```diff
@@ -13,10 +13,7 @@ function isJest() {
 }
 
 function isDev() {
-  return (
-    process.env.NODE_ENV !== "production" &&
-    process.env.BUDIBASE_ENVIRONMENT !== "production"
```
Rory-Powell, line 18: As far as I can tell this is our only usage of `BUDIBASE_ENVIRONMENT`. Removing this from the `isDev` check so that we can use this flag to differentiate between prod, preprod and release

Rory-Powell: This enabled tools monitoring tools like elastic to consume metrics from our preprod env when testing out integrations while not polluting the production dataset

mike12345567: Agree with this - its weird having a named env var like this has zero function, good to give it one!

mike12345567 (approved):

Issue #7451, opened 2022-08-25
S3 add additional params 


melohagan:
**Describe the feature request**
At the moment the S3 connector is fairly basic; you can only return a list of objects from bucket name.
You should be able to specify a range of parameters as outlined here: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjects-property 

Discussion: https://github.com/Budibase/budibase/discussions/7432

**Screenshots**
<img width="367" alt="Screenshot 2022-08-25 at 09 00 18" src="https://user-images.githubusercontent.com/101575380/186609305-79f0f65b-b37f-439e-b0a7-e758d27b8bd3.png">

PR #7521, opened 2022-08-30
S3 bucket connector improvements

melohagan:
## Description
- Added Create method for creating new buckets
- Added Read CSV for reading returning a CSV file as JSON
- Added Delete method for deleting objects
- Added additional params for the existing Read method

Addresses: 
- https://github.com/Budibase/budibase/issues/7451

## Screenshots
**Read**
Screenshot 2022-08-30 at 20 02 12

Screenshot 2022-08-30 at 20 02 40

**Read CSV**
Screenshot 2022-08-30 at 20 03 35

Screenshot 2022-08-30 at 20 03 59

**Create**
Screenshot 2022-08-30 at 20 04 53

**Delete**
Screenshot 2022-08-30 at 20 05 22


mike12345567 (approved):
LGTM! In future it would probably be nice if we could add an `optional` tag to an attribute and have a badge or something labelled beside the input to make users aware they don't need to fill everything out, if you've only limited experience with S3 this could add more confusion to its configuration.

melohagan:
> LGTM! In future it would probably be nice if we could add an `optional` tag to an attribute and have a badge or something labelled beside the input to make users aware they don't need to fill everything out, if you've only limited experience with S3 this could add more confusion to its configuration.

That's a great idea! Although there is also the standard convention of simply putting an asterisk beside the required field labels.

Rory-Powell (approved):

Issue #7533, opened 2022-08-31
Form Text Field Fires OnChange events even if the value hasn't changed.


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7499

<div type='discussions-op-text'>

<sup>Originally posted by **poirazis** August 29, 2022</sup>
Server : Latest self hosted, Debian, 
Client : Mozilla Firefox, MacOSX Monterrey, M1

It seems that the Form Text Field fires its OnChange event sequence on OnBlur, even if the value hasn't changed.
So focus / unfocus the control will fire the events again.

Shouldn't the event fire only when there is an actual change in the value or should the event be renamed to "OnBlur" rather than "onChange" ?</div>

In this gif I'm just pressing the enter key:
![update_state](https://user-images.githubusercontent.com/101575380/187647332-0aeff90c-4a6e-479d-bc02-41d8fea75c26.gif)

aptkingston:
Pressing enter is just an explicit action we listen for that forces a value update (and therefore the change event). If you just click in and out of the field I think you should see that the change event does not fire?

melohagan:
No it still triggers onBlur. So if I click in and out it does fire, or if I press the tab key it also fires.

PR #7535, opened 2022-08-31
Only fire onchange events from form fields when values change

aptkingston:
## Description
This PR updates field components to only fire change events when their value has changed to a different value. It will now only fire if the value is valid as well - so if you set a field value to an invalid value (and get a validation error displayed) then the change event will not fire for this invalid value.

Addresses: 
- #7533


shogunpurple (approved):
LGTM

Issue #7059, opened 2022-08-03
Custom Datasources: Builder Usage


shogunpurple:
**COS**
- Pull the latest version of each datasource plugin on startup
- Load the modules into BB so that the latest version is the one that is called by the datasource itself
- Show the new datasources under a tab in the data section of the builder
- Should work and behave like any other datasource

Issue #7022, opened 2022-08-01
Budibase CLI: Validate Plugin


shogunpurple:
Specs: https://budibase.atlassian.net/wiki/spaces/PROD/pages/45940739/Custom+Datasources, https://budibase.atlassian.net/wiki/spaces/PROD/pages/46792705/Custom+Components

When I'm using the budibase CLI, I want to be able to validate the structure of the component/plugin that I am currently developing.

Commands:

`budi lint`
The CLI will check if you are currently in the directory of a plugin/component (look for package.json) and if not error.

**COS**
- precursor to `budi build` 
- `budi lint` - validate structure of schema, check `package.json` for correct deps, check types (possibly use Joi/TS for this)

Issue #6956, opened 2022-07-28
Custom Components: Builder Usage


shogunpurple:
When I'm using the builder, I want to be able to use the custom components that I have added to my plugins tab.

**COS**
- Custom components are displayed in their own section in the component selection menu

Issue #6945, opened 2022-07-28
Budibase CLI - Create Plugin/Component


shogunpurple:
Specs: https://budibase.atlassian.net/wiki/spaces/PROD/pages/45940739/Custom+Datasources, https://budibase.atlassian.net/wiki/spaces/PROD/pages/46792705/Custom+Components

When I'm using the budibase CLI, I want to be able to generate a new skeleton of a component or datasource in the directory I am currently in.

Commands:
- `budi plugin --init <name>`
- `budi component --init <name>`

The CLI will check if you are currently in the directory of a plugin/component (look for package.json) and if not error.

**COS**
- `budi plugin --init`  - creates a new plugin skeleton in the current directory.
- `budi component --init` - creates a new component skeleton in the current directory.

mike12345567:
Some notes around plugin creation:
GOALS
1. Single file that exports the functionality and the schema - plugin.js/main.js
2. Skeleton comes with schema.json which default exports schema for datasource/component
3. Single JS file for datasources/components with schema etc in it (built and minified) but we include the package.json and schema.json + the js file in a tarball (support prebuilds down the line)
4. Skeleton has both TS and JS base impls - CLI asks which one user wants, deletes the other file - build needs to account for this
5. There is a specific datasource.js/ts and component.svelte
6. Build step should be the same for both datasources and components, this is needed for datasources to handle Typescript and to make sure that the node_modules get included correctly. Rollup initially, Esbuild down the line perhaps
7. CJS bundle for datasources, IIFE  for components - build can run this based on the schema.json
8. We handle the importing of tarballs and the usage of it for datasources and components
9. Verify, a post build step - built in, this isn't a separate CLI command, its just part of the build, use JOI to verify the schema.
10. Enrichment feature, Typescript in svelte, typecast the stuff gotten from context 
11. Should we include BBUI - is there a good case for using BBUI components inside a custom component, should components not be granular and add BBUI components to a block inside Budibase

PR #7546, opened 2022-08-31
Base plugin CLI and integration

mike12345567:
## Description
This is the work that myself and @aptkingston have been working on to implement the following components:
1. The CLI for init/creation of a new plugin project (component or datasource)
2. The skeleton of the plugin projects, found [here](https://github.com/Budibase/budibase-skeleton) - this is in a separate repo for licensing purposes, as well as the fact that they aren't real JS projects - they are HBS skeletons that the CLI will enrich into ready to build plugins
3. The validation/build process for datasource and component plugins, utilising rollup and a custom tar rollup action
4. The basic upload API within the server, allowing the plugin tarballs to be sent up to the server (as a file upload) and have the server extract them, send them to MinIO/the object store, then make them available for use by storing a plugin document to CouchDB which contains its schema (so that we can supply this in the builder for datasources/components)
5. The integration of components into the builder (so that they can be seen in their own custom section) and into the client, so that they can actually be used, we use SSR to update the client app HTML with script tags for each of the used custom components
6. The integration of datasources into the server, so that when one is used it has the JS available to actually run the various datasource query types
7. Moving around datasource definitions so that they could be made accessible through the `@budibase/types` package, which contains all the various enumerations that are needed to make this work
8. The handling of custom components in a "smart" fashion, meaning that components will only be SSR'd into the client app HTML if they are used
9. The implementation of a watch mode on the server and the CLI - so that plugins can be automatically reloaded by adding them to a mounted directory which the Budibase app service is aware of. By default the app service looks within the `/plugins` directory within the image (controlled by the `PLUGINS_DIR` env var) - all the user needs to do is mount this directory to their local filesystem in a location that contains plugin projects

## further work
While this covers quite a bit there is still some further work which is to come:
1. Better error handling - right now if a bad bundle is pushed into the server it will crash the client lib/the server - this is something that needs a bit of research to see what can be done to invalidate bad JS bundles
2. Use of a websocket server to make the client aware that a plugin has been updated - this allows live reload of the component in the builder preview, improving the development experience dramatically
3. Documentation - a large piece of work is still needed to document everything, how to write the schemas, how to build the components, how to use the SDK and so on. Right now it is possible to build components/datasources but it requires the developer to understand quite a bit about Budibase - we need to make a decision about how we are going to go about documenting this
4. Further upload methods - right now this just supports the file system mount and file upload methods, however we planned to support Github, NPM, URL and so on
5. Statistics - we will likely want to record some information to posthog about the usage of custom components
6. Testing - we need to come up with some plans on how we will test this functionality - a separate Jest PR will come which tests some of the bits of the server that can be tested easily however we likely need to come up with some good tests for the CLI as right now it is completely un-tested and will be the main interface for this feature.

This is being merged as is right now so that it doesn't block the movement of other plugin projects, a lot of the above components are in development but this is already a very large PR and provides a good amount of functionality - once the develop -> master PR has gone through and we begin prep for the larger release having this in develop should smooth the development of all other features related to plugins.

Addresses: 
- #7059
- #6941
- #7022
- #6945
- #6943
- #6956
- #6942


shogunpurple (approved):
Amazing piece of work - the auto bundling from a bind mount, and the CLI work is really nice on this as well. LGTM 👌

aptkingston:
Added error handling for component plugins. For obvious compilation errors, rollup will fail and so you won't build a bad bundle in the first place which is good, but we needed to handle runtime/svelte level errors ourselves.

Now if your plugin errors while running you'll get an error state rendered as your component which even shows the internal error itself. Because it won't crash now, it means that the client continues to work, hot reloading continues to work and you can keep developing.

Here's an example of a bugged component:
image

Issue #7565, opened 2022-09-01
Alphabetically Ordered Automation List


R2bEEaton:
**Describe the feature request**
The automation list is currently ordered by the internal ID of the automation, making finding automations pretty annoying. You also cannot use your own naming scheme to group related automations as the naming is ignored for sorting.

**Screenshots**
This problem is exacerbated as number of automations increases:
![image](https://user-images.githubusercontent.com/34921506/187889951-8bcf55b1-a21c-4d98-960c-864addcf6b6b.png)

poirazis:
Also a grouping mechanism would be good ( various parts of BB ).
Like grouping tables / queries in the DataSource, grouping Automations and an incremental Search.

Say you have 100 Automations, to able to split them into groups ( Invoicing, AutoSocialMediaPosting etc ) so you can navigate better. 

just an idea!

R2bEEaton:
Totally. I feel Automations need a pretty big overhaul in general. Here are some more ideas I compiled recently: #7553.

ConorWebb96:
Going to close this as this is done.

Will open a separate feature request for categorising and grouping automations. Will link back to your comments.

PR #7566, opened 2022-09-01
Automation list sort alphabetically by name #7565

R2bEEaton:
This PR addresses issue #7565 (mine). Automations are currently sorted by internal ID and not by name, which makes it slightly harder to find an automation from a user's point of view.

## Description
The Automation list in the BB builder was sorted by an internal ID instead of the Automation name, making it confusing for users with many Automations. This PR changes the `each` loop to work off the `$automationStore.automations` list, but sorted by `name`, ascending.

Addresses: 
- #7565

## Screenshots
Old:
image

New:
image

### DIFF
--- a/packages/builder/src/components/automation/AutomationPanel/AutomationList.svelte
+++ b/packages/builder/src/components/automation/AutomationPanel/AutomationList.svelte
@@ -5,7 +5,6 @@
   import NavItem from "components/common/NavItem.svelte"
   import EditAutomationPopover from "./EditAutomationPopover.svelte"
   import { notifications } from "@budibase/bbui"
-  import _ from "lodash"
 
   $: selectedAutomationId = $automationStore.selectedAutomation?.automation?._id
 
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/automation/AutomationPanel/AutomationList.svelte
```diff
@@ -23,7 +24,7 @@
 </script>
 
 <div class="automations-list">
-  {#each $automationStore.automations as automation, idx}
+  {#each _.orderBy($automationStore.automations, ["name"], ["asc"]) as automation, idx}
```
shogunpurple, line 27: Can be done without lodash and simplified to:
`$automationStore.automations.sort(aut => aut.name)`

shogunpurple (changes requested):
Thanks @R2bEEaton, I would just remove the use of lodash here and use the standard JS API.

Issue #5733, opened 2022-05-05
Formula from relation causes API error when relation is removed


Upgreydd:
**Hosting**
- Self
  - Method: docker-compose
  - Budibase Version: 1.0.142
  - App Version: 1.0.142

**Describe the bug**
I've check that in Postgres 11.
Can't remove formula column if relation doesn't exist anymore. 

**To Reproduce**
1. Create Table A
2. Create Table B
3. Declare relation between tables A and B
4. Create Formula column in table A with relation to table B
5. Remove relation between tables
6. You have an error - you can't remove Formula column
```
{"message":"error: column \"Employment start date\" of relation \"person\" does not exist","status":500}
```

**Expected behavior**
Column is removed with no issues

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

PR #7574, opened 2022-09-01
Exclude formula columns from Sql drop column

melohagan:
## Description
SQL was incorrectly trying to delete a Formula column which only exists within internal Budibase DB

Addresses: 
- https://github.com/Budibase/budibase/issues/5733


mike12345567 (approved):
LGTM - we'll get this in before code cut off!

PR #7588, opened 2022-09-02
Hot reloading of custom components via websocket

aptkingston:
## Description
This PR adds realtime hot reloading of custom components inside the builder preview, using websockets.
Opening this as a separate PR to address any feedback about websocket changes separately.

## Websocket overview
This is the first usage of a websocket inside the platform, excluding the vite HMR websocket in dev, but it probably won't be the last. Different websockets will need different paths to connect on (e.g. one from client app, and one from the builder) so nginx is now configured to pass everything on `/socket/*` to the server, attatching the necessary headers. `/socket/client` is the path in use here.

The websocket uses socket.io which natively supports Koa, meaning we don't actually have to listen on a separate port - we can use our existing HTTP server instance and simply listen on certain paths.

I've added a simple wrapper class around the socket.io websocket class, providing a very simple API for now for just emitting events to all connected clients. All communication is 1 way and we only have 1 event handler for now - plugin update notifications sent from the server to clients.

As websockets go through our main proxy, this is of course supported in any sort of environment (e.g. a production cluster). There's no dependency on a dev stack.

## How hot reloading works
The full flow for hot reloading components is as follows:
- Rollup watches your plugin directory and generates a new tarball with any changes. These include a hash for uniqueness
- The server has a volume mounted watching your plugins directory, and sees a new tarball
- The new plugin is imported, and the server emits a plugin update event over a websocket to connected clients
- When the client library was rendered, all plugin components were registered in a central store for use later
- The client recieves the websocket message and dynamically updates the `script` tags in the document head to trigger downloading the new bundle
- Bundle request URLs have hashes appended as a query param, which is ignored but serves to bust caches
- When the new bundle is downloaded, the IIFE bundle is executed and registered at runtime inside the client library
- The client library then remounts any custom components which use this bundle

## Demo
In combination with some improvements to node-cholidar, the file watcher, the speed is now quite impressive. Going from making a change on the local filesystem now, to seeing the updated component inside the preview, takes around 500ms.

https://user-images.githubusercontent.com/9075550/188128194-26d58abd-af12-4842-b002-c36c8c2c1d57.mp4

As error handling for custom components is now included, this also works for hot reloading error states.

https://user-images.githubusercontent.com/9075550/188128876-54ee284c-8d5b-49fa-a687-80df92624926.mp4

It's worth noting however that we cannot catch all runtime errors - only those during component initialisation. If errors are thrown during callbacks at runtime then it is up to the plugin author to handle these.


packages/server/src/utilities/websocket.js
```diff
@@ -0,0 +1,9 @@
+const { Server } = require("socket.io")
+
+const io = new Server(4000)
+
+io.on("connection", socket => {
+  socket.on("foo", arg => {
```
shogunpurple, line 6: Seems like this is just some left over test code.

shogunpurple (approved):
LGTM! Super simple impl as well.

packages/server/src/utilities/websocket.js
aptkingston: Nice catch thanks, you're exactly right.

aptkingston:
This is now feature complete and ready to merge, as soon as I work out why random tests are failing after pulling in develop due to `get "/api/:tableId/:rowId/enrich": "middleware" must be a function, not "undefined"`. Can't see anything obviously wrong, but this will miss the code freeze.

PR #7591, opened 2022-09-02
Fix/platform email casing

shogunpurple:
## Description
This PR solves some issues with email casing on logins reported by a user over email. If you invited a user with an email address consisting of different casing, eg. `TeSt@TEST.com` they would have to match this exact casing when logging in. Rather than enforcing a certain structure of `_id` in the database and performing a migration, I've added a new view to the `global-info` DB that lowercases the `_id`s of all the elements, and updated the relevant code to read from this view, rather than accessing the documents directly. This also ensures when inviting new users, they will not be duplicated based on email casing, causing a user that will not be able to log in.

Addresses: 
- ...more if required

## Screenshots


packages/backend-core/src/tenancy/tenancy.ts
```diff
@@ -8,6 +8,7 @@ import {
   getTenantIDFromAppID,
 } from "../context"
 import env from "../environment"
+import { queryPlatformView } from "../db"
```
Rory-Powell, line 11: Can add this to the existing `../db` dependency above

packages/worker/src/sdk/users/users.ts
```diff
@@ -340,8 +332,10 @@ export const bulkCreate = async (
 
   for (const newUser of newUsersRequested) {
     if (
-      newUsers.find((x: any) => x.email === newUser.email) ||
-      existingEmails.includes(newUser.email)
+      newUsers.find(
+        (x: any) => x.email.toLowerCase() === newUser.email.toLowerCase()
+      ) ||
+      existingEmails.includes(newUser.email.toLowerCase())
```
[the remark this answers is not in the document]
Rory-Powell: The lowercasing should be applied to all three look ups for 
- `getExistingTenantUsers`
- `getExistingPlatformUsers`
- `getExistingAccounts`

We should be sure that the `emails: string[]` passed to each of those functions should do 
`emails = emails.map(e => e.toLowerCase)`. It's a bit tedious to do it in each but safest in case they would ever be called elsewhere. 

We would also want to make sure the emails returned from `searchExistingEmails` are lowercased as the comparison here on 338 is comparing the requested email with the email returned from the underlying document.

Rory-Powell (commented):
The number of changes to the cli lock file is strange, any idea what's going in there?

shogunpurple:
@Rory-Powell yeah, I'm not too sure what happened there - I haven't made any changes on that side, so I've just reset it to the current state of `master`

packages/worker/src/api/routes/global/tests/users.spec.ts
```diff
@@ -242,6 +242,26 @@ describe("/api/global/users", () => {
...
     })
+
+    it("should not be able to create a user with the same email and different casing", async () => {
+      const user = structures.users.user()
+      await api.users.saveUser(user)
+
+      user.email = user.email.toUpperCase()
+      await api.users.saveUser(user)
+
+      expect(events.user.created).toBeCalledTimes(1)
```
Rory-Powell, line 253: Should this be `toBeCalledTimes(0)`? Given the user already exists, it seems like it's not matching to the existing one

packages/worker/src/sdk/users/users.ts
```diff
@@ -164,7 +164,7 @@ const buildUser = async (
 const validateUniqueUser = async (email: string, tenantId: string) => {
   // check budibase users in other tenants
   if (env.MULTI_TENANCY) {
-    const tenantUser = await tenancy.getTenantUser(email)
+    const tenantUser = (await tenancy.getTenantUser(email)) as User | undefined
```
Rory-Powell, line 167: I think this should be  `as PlatformUserByEmail | null`, which is a new document that was introduced as a part of the user fixes. Best thing would be to add that directly to the function signature of `getTenantUser` like:
```ts
export const getTenantUser = async (identifier: string): Promise<PlatformUserByEmail | null> => {
```

Couple of side points:

1. The method signature isn't actually validated inside this function due to the usage of `doWithDB` which isn't TS. So inside the db function we could still say `return "hello"` and it wouldn't complain. If we converted `backend-core/src/db/index.js` to typescript we'd get validation that the function actually returns what we say it does. Something to do in a future PR. 

2. I'm more a fan of returning `undefined` than returning `null` in most cases like you have done with this cast, so we could update the exception handling in `getTenantUser` to just `return` instead of `return null` and change the signature to use `undefined` instead. In either case we can update the if statement below to
```ts
if (tenantUser && ...)
```
rather than the explicit null check. 

3. it would be most accurate to introduce an additional type inside `types/src/documents/users.ts`:
```ts
/**
 * doc id is user id
 */
export interface PlatformUserById extends Document {
  tenantId: string
}
```
and have the `getTenantUser` function look like:
```ts
export const getTenantUser = async (identifier: string): Promise<PlatformUserByEmail | PlatformUserById | null> => {
```
I try to stay away from using multi return types as much as possible as it puts additional logic on the consumers to determine which return type they have received. Usually it's easier to favour dedicated signatures. e.g. 
```ts
const getTenantUserByEmail = (): Promise<PlatformUserByEmail | null>  => {}
const getTenantUserById = (): Promise<PlatformUserById | null>  => {}
```

packages/worker/src/api/routes/global/tests/users.spec.ts
shogunpurple: it generates a UUID so it should always be called once, but the next one shouldn't be called because it's been uppercased.

packages/worker/src/sdk/users/users.ts
shogunpurple: Yeah, we also need to just finish the conversion and we will get everything out of the box.

On actually viewing the interface for `PlatformUserByEmail`, tbh `PlatformUserByEmail` and `PlatformUserById` don't feel right as types - for me this is where it starts to get to a point where we are just creating types to satisfy typescript and people unfamiliar with the codebase have no idea what they are - it's not actually mapped to our domain. 

The entity is a `PlatformUser`, where the fields like `tenantId` are realistically optional - the only difference was how we fetched them (by email/tenantId) they return the same object. Obviously a matter of preference but IMO you want to use unions/multi types to put more on the consumer, and make them code more defensively for values that can actually be `null`/`undefined`, unless you account for it in the function itself and guarantee that it cannot return those values.

shogunpurple: > rather than the explicit null check

Since this is loose equality it will catch both, which is nice

Issue #7592, opened 2022-09-02
Preview and live apps don't render any components after 1.3.1 update


Klience:
**Hosting**
- Cloud v.1.3.1

**Describe the bug**
Just updated from 1.2.52 to 1.3.1. All screens are empty.

**To Reproduce**
1. Update an app to 1.3.1
2. Click on "Preview" or live app
3. See blank screens

**Expected behavior**
All components are shown.

**Screenshots**
Note how builder renders everything ok. And **preview** and **live** apps don't.

![builder screens](https://user-images.githubusercontent.com/94982318/188197127-8c4fb4be-1284-4edf-afe8-749eebb549a0.png)
![blank screens](https://user-images.githubusercontent.com/94982318/188197153-bf98e3a4-881b-4828-bbf6-38cc850d09ea.png)
  
**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 104.0.5112.102 (Official Build) (64-bit)

mike12345567:
Hi @Klience - just to check did you re-publish your app after updating?

I am unable to recreate this behaviour, if possible can you provide an export an app which is exhibiting this behaviour?

Klience:
Hi @mike12345567 Yes I did re-publish the app. Closed the browser. Tried incognito. Updated the browser to Version 105.0.5195.54 (Official Build) (64-bit). Tried the same steps. Nothing changed.

Klience:
App was updated 10 minutes before I opened the issue.

![update app](https://user-images.githubusercontent.com/94982318/188198506-171d0576-9a95-436d-8f87-60a2099e1b62.png)

mike12345567:
Hi @Klience - if you try updating your client library now this should resolve the issue. We've released 1.3.2 which should resolve this.

Klience:
@mike12345567 Just updated to 1.3.2. It does solve the issue. Thanks for the quick response and fix!

PR #7593, opened 2022-09-02
Solving screen load issue in client library

mike12345567:
## Description
Adding in a check for when there are no screen actions that need to be performed - this is still partially unexplained as it does not occur on every screen however this appears to solve the issue.

Addresses: 
- #7592


shogunpurple (approved):
LGTM - good job on the fix.

Issue #7530, opened 2022-08-31
User Groups, When I click the 3 dots, and select "Delete", we should have one more extra confirmation step, before we delete the user group 


MihailHadzhiev2022:
When I click the 3 dots, I wanted to edit my User group, but I clicked by mistake the delete button and my userGroup was deleted. 
We should add a second confirmation step. 
<img width="804" alt="Screenshot 2022-08-31 at 11 09 15" src="https://user-images.githubusercontent.com/102024164/187636431-39b2c175-9d65-4cf1-8f13-75b22c40447a.png">

PR #7596, opened 2022-09-02
Fix/delete group confirmation

NEOLPAR:
## Description
Showing a modal window before removing a user group

Addresses: 
- fix issue #7530 

## Screenshot

With issue fixed
Screenshot 2022-09-02 at 19 48 39


aptkingston:
Nice one! Might be better to merge this into my `group-fixes` branch though because a lot of groups changes are already on that branch :+1: We don't need to worry about deploying these changes since groups aren't released yet.

PR #7620, opened 2022-09-04
Datasource adldap

vevedh:
## Description
A datasource connector to connect AD LDAP v0 (without tls actually) .
Query enter fields bucket sample:
(mail=adresse@mail.com)

Addresses: 
- (https://github.com/vevedh/budibase/tree/datasource-adldap)

## Screenshots

image

image

image


shogunpurple:
See other PRs - unfortunately this one is not in a state to be merged and is incomplete. We will be releasing a direct AD/LDAP integration at some point in 2023 - but for now it's on hold. Thanks for the contribution!

Issue #7608, opened 2022-09-03
"Too many requests" on Public API (self-hosted)


yasinormin:
Hi,
I have budibase installed on my localhost. I try to add thousands of row using public api, but got "Too many requests" error. I read #5679 and add theses but error still exist. 

- Added `API_REQ_LIMIT_PER_SEC: 1000` to  `docker-compose.yaml`
- Added `API_REQ_LIMIT_PER_SEC=1000` to `.env` 
- Updating docker `docker-compose up -d`

Any idea?

.env
```
# Use the main port in the builder for your self hosting URL, e.g. localhost:10000
MAIN_PORT=10000

# This section contains all secrets pertaining to the system
# These should be updated
JWT_SECRET=testsecret
MINIO_ACCESS_KEY=budibase
MINIO_SECRET_KEY=budibase
COUCH_DB_PASSWORD=budibase
COUCH_DB_USER=budibase
REDIS_PASSWORD=budibase
INTERNAL_API_KEY=budibase
API_REQ_LIMIT_PER_SEC=1000

# This section contains variables that do not need to be altered under normal circumstances
APP_PORT=4002
WORKER_PORT=4003
MINIO_PORT=4004
COUCH_DB_PORT=4005
REDIS_PORT=6379
WATCHTOWER_PORT=6161
BUDIBASE_ENVIRONMENT=PRODUCTION

```

docker-compose.yaml
```
version: "3"

# optional ports are specified throughout for more advanced use cases.

services:
  app-service:
    restart: unless-stopped
    image: budibase.docker.scarf.sh/budibase/apps
    container_name: bbapps
    environment:
      SELF_HOSTED: 1
      COUCH_DB_URL: http://${COUCH_DB_USER}:${COUCH_DB_PASSWORD}@couchdb-service:5984
      WORKER_URL: http://worker-service:4003
      MINIO_URL: http://minio-service:9000
      MINIO_ACCESS_KEY: ${MINIO_ACCESS_KEY}
      MINIO_SECRET_KEY: ${MINIO_SECRET_KEY}
      INTERNAL_API_KEY: ${INTERNAL_API_KEY}
      BUDIBASE_ENVIRONMENT: ${BUDIBASE_ENVIRONMENT}
      PORT: 4002
      JWT_SECRET: ${JWT_SECRET}
      LOG_LEVEL: info
      SENTRY_DSN: https://a34ae347621946bf8acded18e5b7d4b8@o420233.ingest.sentry.io/5338131
[...]

PR #7622, opened 2022-09-05
Configurable api rate limit through nginx

Rory-Powell:
## Description
Currently there is a flag for controlling the public api rate limit `API_REQ_LIMIT_PER_SEC`. However this is superseded by the NGINX configuration for the `/api` route which covers the `/api/public/xxx` location for public api endpoints. 

Add configuration similar to how webhook rate limiting to allow this to be configured at runtime via environment variables. 

Addresses: 
https://github.com/Budibase/budibase/issues/7608


shogunpurple (approved):
LGTM!

PR #7632, opened 2022-09-05
Prevent default browser events from firing when using keyboard shortcuts

aptkingston:
## Description
Fixes an issue where browser events for certain keybaord shortcuts were firing when they shouldn't be.


melohagan (approved):
LGTM!

Issue #7603, opened 2022-09-03
Can't build single container image for the latest master branch (v1.3.4)


taoyuan:
Can not build the single container image for the latest master branch (v1.3.4). It will throws error below when building: 

```
Cannot find module '@shopify/jest-koa-mocks' or its corresponding type declarations.
```

Just run `yarn build:docker:single` or run `Deploy Budibase Single Container Image to DockerHub` action in github actions. You will see the error.

aderici:
Facing the same issue here! please help! also happens on development branch

aderici:
did you also just change couchdb persist folder from couch to couchdb? Upgraders won't be happy

taoyuan:
No, I just keep the default settings. 

However, when upgrading, I encountered that the design UI could not be displayed, any project (original or new) always kept loading, but the preview was fine.

After I restarted the database related docker services, it worked. 

When I was about to try the single docker image, I encountered the situation described in this issue.

aderici:
That comment was for the budibase team. I believe the persisted files folder name changed for couchdb so it didnt automatically pick up my old files, i had to change local persisted folders name to get it to work. 

I could run the single docker by adding "yarn add @shopify/jest-koa-mocks" to dockerfile just before worker build command.

I have also faced the same issue that builder can not display pages, unable to develop anything. I tried restarting the docker container, no luck. do you remember what exactly you restarted?

taoyuan:
Run `yarn dev:stack:down` and `yarn dev:stack:up` in `packages/server` to re-up docker services. After that, restart budibase apps. Good luck.

jonnymccullagh:
> That comment was for the budibase team. I believe the persisted files folder name changed for couchdb so it didnt automatically pick up my old files, i had to change local persisted folders name to get it to work.

That is on me. I will revert that after Mike addresses the build issue

aderici:
@mike12345567 thanks for the fix. Have you seen #7630 ? Even if you do build the app, can not view screens as a developer. /preview endpoint keeps returning 403

taoyuan:
@mike12345567 Still can't build the single container image. The same error throws.

Rory-Powell:
@taoyuan we're working on a change now that should resolve the issue

geertvanheusden:
Would it be possible to have an automated build setup for this Single container image in Github Actions and let it automatically push to the Docker Hub? That would be really great and helpful. 😊

Rory-Powell:
[...]

PR #7633, opened 2022-09-05
Fixing backend-core build, as raised in.

mike12345567:
## Description
Two part issue, a dependency was not included as required in the backend-core and also the typescript build for the worker was including scripts and tests, which should not be used outside of development and therefore do not need to built.

Addresses: 
-  #7603

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -60,7 +60,7 @@
     ]
   },
   "devDependencies": {
-    "@shopify/jest-koa-mocks": "^5.0.1",
+    "@shopify/jest-koa-mocks": "5.0.1",
     "@types/jest": "27.5.1",
     "@types/koa": "2.0.52",
     "@types/lodash": "4.14.180",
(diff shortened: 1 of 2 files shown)


packages/backend-core/package.json
```diff
@@ -60,7 +60,7 @@
     ]
   },
   "devDependencies": {
-    "@shopify/jest-koa-mocks": "3.1.5",
+    "@shopify/jest-koa-mocks": "5.0.1",
```
Rory-Powell, line 63: How come this one needed bumped? It looks like it wouldn't cause issues

mike12345567: unsure, was experiencing build issues with current version, bumped it and it helped solve it - didn't seem to cause issues bumping it up.

Rory-Powell: It was probably because the `tests` dir was being included, which required a usage on this dev dependency as mentioned on the issue description. We could probably get away with not bumping, but it shouldn't hurt. 

There should be a lockfile change associated with this bump however?

PR #7644, opened 2022-09-05
Custom CSS upgrades

aptkingston:
## Description
This PR adds 2 improvements to custom CSS:
- Custom CSS is now a bindable field and accepts all usual bindings
- Custom CSS can be overridden via conditional UI

Extending work started by @melohagan.

Addresses:
- https://github.com/Budibase/budibase/issues/5818

### Bindable custom CSS
Custom CSS still opens in a drawer like before, but now it uses the binding drawer to allow inserting data bindings. This is full backwards compatible and, as always, using bindings is optional.

For example, here's setting the background color of a container to a state variable:
image

Now we can control the colour by setting a value in state:

https://user-images.githubusercontent.com/9075550/188489207-04ac7642-9420-4cd7-8ce5-3542e00fa035.mp4

### Custom CSS via conditional UI
There's now an option to update custom CSS via conditional UI. This behaves like any other setting.

For example, here's a container which will change background colour due to conditional UI custom CSS whenever a state variable is set:
image

Now we can control the colour by setting a value in state:

https://user-images.githubusercontent.com/9075550/188489555-726932a9-f4a2-48b3-b73e-b0a170afed94.mp4


joebudi:
This is awesome. Do you think it might confuse people if they see handles and JavaScript as the tabs?

aptkingston:
> This is awesome. Do you think it might confuse people if they see handlebars and JavaScript as the tabs? Apologies if I'm missing something here - I'm positive I am.

Thanks! I don't think it will be confusing - this is just the normal binding drawer that they see everywhere else in the platform when they edit a setting that accepts bindings, so they should be very familiar with it.

packages/client/src/components/Component.svelte
```diff
@@ -372,6 +381,10 @@
       initialSettings = cachedSettings
     } else {
       Object.keys(allSettings).forEach(key => {
+        if (key === "_css") {
+          console.log(enrichedSettings._css)
```
melohagan, line 385: Console log.

melohagan (approved):
LGTM! Being able to use bindings directly in the custom CSS is very nice! 😄

poirazis:
your next 10 pints on me @aptkingston !

PR #7646, opened 2022-09-05
QA Core repo, and Jest API tests running end to end

shogunpurple:
## Description
This PR contains code essentially rewriting/remodelling the current newman API test suite to be written from scratch with jest and typescript instead. There's a few reasons for this:
- We are going to need jest eventually to model more complicated logic, which newman is not good at
- There's quite a few differences between running newman with the JS API vs inside postman. It's a more consistent experience having one way to run the tests
- We can use the `@budibase/types` repo and openAPI spec and have a consistent interface for all the types across our domain, and provide some compile time checking when writing API tests.

You will notice a new top level dir `qa-core`, which will be the eventual home of every test that is not a unit test in the budibase platform, including cypress. 

This test structure follows a fairly simple pattern - there are tests, and fixtures (for generating data). A flexible `generate` function has been provided for all entities, so you can generate one with completely random data, leveraging the `chance` library. The `TestConfiguration` files rely on API classes that have been created for each entity, and return type safe tuples such as the following:

```js
  async create(body: CreateApplicationParams): Promise<[Response, Application]> {
    const response = await this.api.post(`/applications`, { body })
    const json = await response.json()
    return [response, json.data]
  }
```

This allows the tests to make assertions on both the actual API response (Headers, statuscode etc) and also to directly pluck out the JSON object from the call - it's a bit clearer that they are two separate things, and we don't have to merge them into a bigger object - which is less explicit.

The public tests are pretty much completely type safe, given that they use the types directly generated by the open API spec. This means a change to those types or the public API will automatically break the integration tests at compile time, giving us a lot more safety.

This **is not ready** to merge yet, I'm just getting it up for some early feedback on the structure and approach.
**TODO**
- [x] Better type usage using `@budibase/types`
- [x] Further abstraction to move all setup code into API client and `TestConfiguration` to make tests less verbose
- [x] Run the tests in CI and make sure they work as part of the `budibase_ci.yml` build
- [x] Add linting, prettier etc
- [x] Better tests - just adding some checking on bodies etc
- [x] Map long relative paths to a cleaner format for easier imports
- [x] Spin down newman once we have this up and running


qa-core/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@types/node-fetch": "^2.6.2",
+    "chance": "^1.1.8",
+    "jest": "^28.0.2",
+    "prettier": "^2.7.1",
+    "start-server-and-test": "^1.14.0",
+    "timekeeper": "^2.2.0",
+    "ts-jest": "28.0.8",
+    "ts-node": "^10.9.1",
+    "tsconfig-paths": "^4.1.0",
+    "typescript": "^4.8.2"
```
Rory-Powell, line 47: Pinning all of these dependencies would be good so we don't get drift between the versions used here and those used inside the rest of the backend packages. 

It might also be good to sync the TS version to `4.7.3`, looking at the search result below it's the most commonly used version across all our packages 
Screenshot 2022-09-06 at 09 00 40

qa-core/src/tests/public-api/PublicAPIClient.ts
```diff
@@ -0,0 +1,58 @@
+import env from "../../environment"
+import fetch from "node-fetch"
+
+interface HeaderOptions {
```
Rory-Powell, line 4: Should it be `Options` given it's more than just headers?

qa-core/src/jest.extends.ts
```diff
@@ -0,0 +1,22 @@
...
+  namespace jest {
+    interface Matchers<R> {
+      toHaveStatusCode(code: number): R;
+    }
+  }
+}
+
+// Expect extensions
+expect.extend({
+  toHaveStatusCode(received, code) {
```
Rory-Powell, line 14: Nice extension 👌

Rory-Powell, line 14: Does this work?

```ts
import { Response } from "node-fetch"

expect.extend({
  toHaveStatusCode(received: Response, code: number) {
...
```

qa-core/tsconfig.json
```diff
@@ -0,0 +1,36 @@
...
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "incremental": true,
+    "types": ["node", "jest"],
+    "outDir": "dist",
+    "skipLibCheck": true,
+    "paths": {
+      "@budibase/types": ["../packages/types/src"],
+      "@budibase/backend-core": ["../packages/backend-core/src"],
+      "@budibase/backend-core/*": ["../packages/backend-core/*"]
```
Rory-Powell, line 18: Do we need backend-core at all for the tests?

qa-core/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "api:server:setup:ci": "node ../packages/builder/cypress/setup.js",
+    "api:test:ci": "start-server-and-test api:server:setup:ci http://localhost:4100/builder test",
+    "api:test": "start-server-and-test api:server:setup http://localhost:4100/builder test"
+  },
+  "jest": {
+    "preset": "ts-jest",
+    "testEnvironment": "node",
+    "moduleNameMapper": {
+      "@budibase/backend-core/(.*)": "<rootDir>/../packages/backend-core/$1",
+      "@budibase/backend-core": "<rootDir>/../packages/backend-core/src",
```
Rory-Powell, line 25: Similar to question on tsconfig, would we need backend-core here?

qa-core/src/tests/public-api/tables/tables.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    })
+    const json = await response.json()
+    config.testContext.table = json.data
+    expect(response).toHaveStatusCode(200)
+  })
+
+  it("POST - Search tables", async () => {
+    const response = await api.post(`/tables/search`, {
+      body: {
+        name: config.testContext.table.name
```
Rory-Powell, line 30: Will this work if this test is run independently? It seems reliant on `config.testContext.table`
If so, should that be done as a part of the `beforeAll` instead?

qa-core/src/tests/public-api/tables/tables.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    })
+    expect(response).toHaveStatusCode(200)
```
Rory-Powell, line 33: Should we be matching on the response content in these tests? Or should we try to start enforcing that at the unit test level instead? It seems like something we are missing, we assert on the status only in the majority of the unit tests inside worker and server

qa-core/src/tests/public-api/applications/applications.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  afterAll(async () => {
+    await config.afterAll()
+  })
+
+  it("POST - Create a application", async () => {
+    const response = await api.post(`/applications`, {
+      body: generateApp()
+    })
```
Rory-Powell, line 20: In this case you could used the `seedApp` function. Having utility functions for all the api operations could be very useful, something similar to: https://github.com/Budibase/budibase/blob/develop/packages/worker/src/tests/api/index.ts

For example below in the row tests we have
```ts
await api.post(`/tables/${config.testContext.table._id}/rows`, {
      body: require("./fixtures/row.json")
    })
```

Which could instead be used as 
```ts
await config.api.rows.create(tableId, row)

// with signature
const create = async (tableId: string, body: Row) => {
   return api.post(`/tables/${tableId}/rows`, { body })
}
```

This would then give us a place to hook into the public api TS definitions produced by the openapi spec build. 
The only problem is maintaining fixtures in json, if we maintained fixtures as a const we could attach the appropriate type from the openapi spec build. Here's an example scenario then: 
1) Update public api / docs
2) Types auto updated
3) Test data integrity enforced at build time

Rory-Powell: The hardest part of this will be getting this moved into `@budibase/types` instead of server, some build tweaking needed https://github.com/Budibase/budibase/blob/master/packages/server/src/api/controllers/public/mapping/types.ts

qa-core/tsconfig.json
shogunpurple: Nope, I've removed it in my branch - this was just copied over from existing config.

qa-core/src/tests/public-api/PublicAPIClient.ts
shogunpurple: Yeah tbh i was thinking about getting rid altogether and having it fall back on the native `node-fetch` types.

Rory-Powell: Would be even better 👍

qa-core/src/tests/public-api/tables/tables.spec.ts
shogunpurple: I think we should - I was just mapping these to be exactly like newman so we could get jest integrated

qa-core/src/tests/public-api/tables/tables.spec.ts
shogunpurple: All tests have been updated to seed in the `beforeAll` now 👍 So can be run in isolation

PR #7670, opened 2022-09-07
Update preview endpoint

aptkingston:
## Description
This PR updates the preview endpoint to be prefixed with `/app` so that it works with older nginx configs.


mike12345567 (approved):
LGTM!

Issue #7683, opened 2022-09-07
API REST Queries: URI param encoding


MentalGear:
**Describe the feature request**
When using the Query UI, it would be great if the input would be automatically URI component converted, if there are special characters in the input.
This would make it much easier to scan the parameter values in the UI and change them.

**Screenshots**
<img width="971" alt="Screenshot 2022-09-07 at 18 12 09" src="https://user-images.githubusercontent.com/2837147/188928294-361e0164-6197-4632-8c3c-1d09a8a8ad8f.png">

MentalGear:
Instead of having to write it like this:

<img width="964" alt="Screenshot 2022-09-07 at 18 15 24" src="https://user-images.githubusercontent.com/2837147/188928493-f9f4099e-cfe4-4071-b5ac-c79ab39607ba.png">

MentalGear:
PS: Happy to submit a PR if someone can point me where I can find the relevant files in the src.

mike12345567:
Hi @MentalGear - thanks for the suggestion - I think this is more of a bug than a feature - it really should do this already! I've made a change on one of my bug squashing branches.

MentalGear:
Thx for the quick reply, Mike!
I'm also experiencing a bug on 1.3.2: when trying to save a modification to the query, the green save toast is displayed, yet no changes are saved on reload. Is this already known ?

mike12345567:
Hi @MentalGear - this is likely due to trying to modify the parameters through the URL, rather than one of the options in the Params tab. Some changes to the URL currently will not be reflected in the saved query - I've made a change on my branch so that changes to the URL will update parameters correctly.

FocusDesigner:
I'm not sure if this is related, but I have this problem the other way around. 

If I use a binding as variable for the params the handlebars are getting encoded in the url as well. 
This way the variable isn't translated correctly, which leads to an error in the request.
According to this guide [https://docs.budibase.com/docs/rest-bindings](https://docs.budibase.com/docs/rest-bindings) that shouldn't be the case, or am I missing something?

charlesDannerSpg:
Hi @mike12345567 I am also facing similar problem to @FocusDesigner, where the bindings are encoded into the url.

mike12345567:
Hi guys, I'll get a fix out for this today.

danielband:
[...]

Issue #7679, opened 2022-09-07
bug(UI, z-index): Dropdown below other inputs


MentalGear:
**Hosting**

- Cloud hosted
  - Budibase Version:  1.3.2,

**Describe the bug**
When setting up a Query under the API REST element, the dropdown menu for previously set Auth underlaps other input elements. Furthermore, the dropdown menu is a bit hard to find initially (I first searched within the {{ bindings }} of the Auth parameters.

**To Reproduce**
1. Go to 'Data > Sources > REST API'
2. Click on a 'Query' in the sidebar, e.g. GET query.
3. Click the 'Headers' Tab
4. Open the Auth dropdown.

**Expected behavior**
Dropdown topmost

**Screenshots**
<img width="270" alt="Screenshot 2022-09-07 at 15 46 45" src="https://user-images.githubusercontent.com/2837147/188895213-8759774c-d926-4677-8ba5-c973e88223f3.png">

  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: macOS 15
 - Browser Safari
 - Version: 15.6.1

**Additional context**

Issue #7617, opened 2022-09-04
handlebars  ellipsis is not work


touchsky2018:
handlebars  ellipsis is not work

{{ ellipsis  "adfasdfasdfasf" 7 }}      return  ""
{{ ellipsis  "adfasdfasdfasf", 7 }}     syntax error

melohagan:
Hey @touchsky2018 

Please provide reproduction steps as well as a detailed description of the problem.

mike12345567:
Hi @touchsky2018 - going to close this issue until better reproduction steps can be provided as in the latest versions of Budibase this appears to work in the core library (tests passing) and works in the design/data/automation sections.

![image](https://user-images.githubusercontent.com/4407001/188861326-b8b0597b-eecb-4343-984b-2f951b55930d.png)

Issue #7583, opened 2022-09-02
Exporting an application with the "excludeRows" set to true, still includes the rows


geertvanheusden:
**Hosting**
- Self
  - Method: docker
  - Budibase Version: 1.2.52

**Describe the bug**
Although I wasn't able to find the button to initialise the export that excludes the rows (from this ticket #4552), I tried using the URL directly `http://localhost:10000/api/backups/export?appId=app_24a8b63182544893adc5e6140167247e&appname=My%20app&excludeRows=true`. It generates a backup but my rows are still part of the export.

**To Reproduce**
1. Create an app, add a table and a row
2. Execute the export URL (see above) with the "excludeRows" parameter set to true
3. Search for the value you entered in the row

**Expected behavior**
The values you entered in the rows should not be in the backup.

**Desktop (please complete the following information):**
 - OS: Maco
 - Browser: chrome
 - Version: latest

**Additional context**
As mentioned above, I cannot find the button in the application that shows me the modal with the toggle to exclude the rows. There is a video in the PR that shows how it should work: https://github.com/Budibase/budibase/pull/6059

geertvanheusden:
Any thoughts/feedback on this?

geertvanheusden:
Thx @mike12345567 ! 👏

PR #7685, opened 2022-09-07
Fix/sept various fixes

mike12345567:
## Description
Various fixes gathered over the last few days, discussed in the addresses section.

Addresses: 
- Adds a test case for #7617 to prove that it works as expected.
- Adds compaction in before replication, this should help with replication time in the case that a lot of rows have been deleted right before replication.
- #7431 the reboot option for CRON simply did not work, I've removed it as an option in the Cloud as it doesn't really make sense there and I've made it work in self host (events are triggered as part of the automation init at startup). There are still some issues, like this occurring multiple times in a clustered environment each time an app service starts up, but at least it functions now. Also stopped it from breaking app publishing (which means there are likely no apps using this today since they couldn't be published).
- #7683 URI parameters in a REST query were not being correctly encoded, this meant parameters containing special characters (say spaces) would not work consistently.
- Fixing a z-index issue uncovered in #7679.
- Fixed an issue I found with MySQL time only fields being unloadable after recent changes to support timezones better.
- #6980 fixed an issue with SQL relationships and foreign keys with mis-matching foreign key labels, depending on the way round the update was occurring sometimes this would fail but not alert the user - looking like updates have just been ignored.
- Another issue discussed in #7683, if you changed a query string parameter in the URL bar rather than through the params tab for a REST query, it would look like it has changed but simply be ignored. I've changed how URL params are handled so that you can update them through the URL bar and have them correctly reflected in the params tab as well.
- #7583 bring back the modal that allows exporting apps with/without internal table data.
- Fixed an issue with deleting very large apps - it would stack overflow if you had too many (200K) internal table rows.

# Screenshots
Fixed modal for exporting apps with/without internal table data.
image

### DIFF
--- a/packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
+++ b/packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
@@ -23,6 +23,7 @@
   const dispatch = createEventDispatcher()
   let bindingDrawer
   let valid = true
+  let currentVal = value
 
   $: readableValue = runtimeToReadableBinding(bindings, value)
   $: tempValue = readableValue
@@ -30,11 +31,17 @@
 
   const saveBinding = () => {
     onChange(tempValue)
+    onBlur()
     bindingDrawer.hide()
   }
 
   const onChange = value => {
-    dispatch("change", readableToRuntimeBinding(bindings, value))
+    currentVal = readableToRuntimeBinding(bindings, value)
+    dispatch("change", currentVal)
+  }
+
+  const onBlur = () => {
+    dispatch("blur", currentVal)
   }
 </script>
 
@@ -45,6 +52,7 @@
     readonly={isJS}
     value={isJS ? "(JavaScript function)" : readableValue}
     on:change={event => onChange(event.detail)}
+    on:blur={onBlur}
     {placeholder}
     {updateOnChange}
   />
--- a/packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte
@@ -28,25 +28,25 @@
   import { onMount } from "svelte"
   import restUtils from "helpers/data/utils"
   import {
-    RestBodyTypes as bodyTypes,
-    SchemaTypeOptions,
     PaginationLocations,
     PaginationTypes,
+    RawRestBodyTypes,
+    RestBodyTypes as bodyTypes,
+    SchemaTypeOptions,
   } from "constants/backend"
   import JSONPreview from "components/integration/JSONPreview.svelte"
   import AccessLevelSelect from "components/integration/AccessLevelSelect.svelte"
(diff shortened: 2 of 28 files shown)


packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
```diff
@@ -23,18 +23,25 @@
...
   }
 
   const onChange = value => {
-    dispatch("change", readableToRuntimeBinding(bindings, value))
+    currentVal = readableToRuntimeBinding(bindings, value)
+    dispatch("change", currentVal)
+  }
+
+  const onBlur = () => {
+    dispatch("blur", currentVal)
```
aptkingston, line 44: Could you quickly explain this change? It looks like it's for the key value builder below but I wonder if there is some other bug at work if using the normal `on:change` handlers wasn't working.

packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte
```diff
@@ -314,6 +314,27 @@
     }
   }
 
+  $: console.log(breakQs)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte:317]: Cheeky log left in.

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -33,14 +32,16 @@
...
   let creatingApp = false
   let loaded = $apps?.length || $templates?.length
   let searchTerm = ""
   let cloud = $admin.cloud
   let creatingFromTemplate = false
   let automationErrors
   let accessFilterList = null
+  let welcomeHeader, welcomeBody
+  let createAppButtonText
+  let enrichedApps, filteredApps, lockedApps, unlocked
```
aptkingston, line 44: Since these are all declared as reactive statements, we don't need to define them again as local variables. I don't think doing this will cause bugs, but it might affect how well svelte is able to hoist dependencies since these are now local variables rather than svelte reactive statement declarations.

Basically at best it does nothing, at worst it might cause edge case issues.

aptkingston (commented):
Nice fixes :ok_hand: Just would like a little clarification around the drawerbindableinput change in case there's some other issues there.

packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
mike12345567: `on:change` does work - but I wanted to use blur instead of change in the REST UI, as I'm parsing the end result - it involves bi-directional parsing and getting it working with on change would have been a nightmare, keeping track of the state + re-parsing constantly didn't make sense, it felt better to wait until the user input was complete.

packages/builder/src/pages/builder/portal/apps/index.svelte
mike12345567: I might need to look into my version of Webstorm svelte as it seems to see this as a bug - hoisting them helps to make the files parse-able, as without this every file I look at is just a sea of red lol

shogunpurple (approved):
Lots of great fixes here - good job 👍

PR #7686, opened 2022-09-07
Package update from node-sass to sass in Next.js 

Chukwu3meka:
## Description

[Node Sass](https://www.npmjs.com/package/node-sass) is deprecated and since we're running a newer version of Next.js which has built-in [Sass support](https://nextjs.org/docs/basic-features/built-in-css-support#sass-support) and has the same benefits and restrictions as the built-in CSS support, we don't need to the node-sass as there are no plans to add additional features or compatibility with any new CSS or Sass features.

```diff
 +    "sass": "^1.52.3",
 -     "node-sass": "^7.0.1",
```

Addresses: 

> You will see it is too basic to warrant an issue. I don't want to give you more work.

##Screenshots

image


Chukwu3meka:

Issue #7123, opened 2022-08-05
SSL rejectUnauthorized for REST API


adamkingsbury:
**Describe the feature request**
Hi, I would like to request that you add rejectUnauthorized functionality to the REST Api integration in a similar fashion to how you have implemented for PostgreSQL and MYSQL integrations. 

I would like the ability to integrate Web services on my self hosted platform that utilise self signed certs for https comms.

FocusDesigner:
plus 1

would be awesome to have this feature as most of my REST APIs use self signed certificates and I am not able to connect as described here already

[https://github.com/Budibase/budibase/issues/4615](https://github.com/Budibase/budibase/issues/4615)

If anyone has a temporary solution to fix this, would be super helpful. Thanks

dwesthuis:
Please help, same problem here:

Query Error: request to https://xx.xx.xx.xx:xxxx/xxxxxxx? failed, reason: self signed certificate

adamkingsbury:
I've submitted a pull request to resolve this issue.....first time I've attempted so we'll see how it goes:
/pull/7716

PR #7716, opened 2022-09-09
Reject unauthorized

adamkingsbury:
## Description
This is an enhancement to provide the ability to accept self-signed SSL certificates when utilising the REST API.
A "rejectUnauthorised" toggle was added to the API configuration screen. It defaults to true for security. Users must actively disable the feature to allow self signed certificates.

In order to enable the ability to substitute in an alternate https agent with rejectunauthorised set to false, instead of using the defalut global agent, the following has also been added to packages/server/src/index.ts
`process.env.GLOBAL_AGENT_FORCE_GLOBAL_AGENT = "false"`

Additionally I also corrected an issue where some rest API calls would fail as budibase would encode the "?" for query paramaters on the URL even when there were no query parameters supplied.

Addresses: 
This is related to the following feature request: [SSL rejectUnauthorized for REST API #7123
](https://github.com/Budibase/budibase/issues/7123)

## Screenshots
The reject unauthorised toggle set to false to allow self-signed certificate communications:
image


adamkingsbury:

Gamechiefx:
@adamkingsbury  thanks for this, i wonder how long it takes for them to commit this

Rory-Powell:
Thanks for taking the time to contribute this @adamkingsbury. I've resolved the conflict and we're going to merge this now. We have some concerns around disabling `GLOBAL_AGENT_FORCE_GLOBAL_AGENT` globally, which we'll look at separately by scoping this configuration to the thread executing the rest request

Issue #6624, opened 2022-07-08
Creating automation from button action without field causes error when creating automation


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
Creating automation from button action without field errors when creating automation. It is fine when you add a field, even if it's empty.
You should be able to create an app action automation without any fields.

**To Reproduce**
1. Add a button -> Define actions -> Trigger Automation.
2. Select Create a new automation
3. Enter a name and click save. 
4. The automation is created, but an error message also shows.
5. Go back to the Trigger Automation action. Enter a new name.
6. This time click add field (you can leave the contents blank). Click Save.
7. The automation is successfully created.

**Expected behavior**
I should be able to add a new automation from a button trigger automation action without needing to specify any fields.

**Screenshots**
![Screenshot 2022-07-08 at 16 32 18](https://user-images.githubusercontent.com/101575380/178024001-df973592-05c6-4c0e-ad77-411acbb90f25.png)

![Screenshot 2022-07-08 at 16 34 02](https://user-images.githubusercontent.com/101575380/178024331-de1e9b1b-e9af-4494-9361-251db0c44f9f.png)

**Added but Empty**
![Screenshot 2022-07-08 at 16 35 12](https://user-images.githubusercontent.com/101575380/178024522-38c8442f-960f-4e42-b02f-2b728aa80293.png)

**Add a blank field for a new automation**
![Screenshot 2022-07-08 at 16 36 36](https://user-images.githubusercontent.com/101575380/178024733-ea40a832-139b-44b1-9dbe-59d173ef095a.png)

![Screenshot 2022-07-08 at 16 37 07](https://user-images.githubusercontent.com/101575380/178024813-eab4d5b7-97fd-4641-82d5-7a312dd72bc0.png)

PR #7727, opened 2022-09-12
Null safety when creating automation from design

melohagan:
## Description
When creating a new app action automation from the Design section, an error would be thrown if no fields were added.
You can now create an app action automation from the Design section without any fields.

Addresses: 
- https://github.com/Budibase/budibase/issues/6624


shogunpurple (approved):
LGTM! 👍

Issue #7691, opened 2022-09-08
Nested data providers with filter not working anymore 


mvhaen:
**Hosting**
- Cloud
  - Tenant ID: strobbo

**Describe the bug**

For an internal dashboard application we wanted to show a bunch of pie charts on 1 page, but putting a single data provider in place for all these charts resulted in timeouts.

We then found a workaround by creating 1 query that returned all the results and then creating child data providers that filtered data from the main one as seen below.

<img width="294" alt="image" src="https://user-images.githubusercontent.com/157020/188501509-11b7f402-ce78-4f0c-a79f-8654cf58cc69.png">

We now wanted to expand this page with some extra graphs. For some reason every new nested data provider starts returning zero results when we add a filter (expect if we add a filter on is empty then we get all results). 

I've started a new page + query from scratch and it shows the same results. The pie charts we set up before still work, when I duplicate the nested provider and the chart under it, it also works but it breaks when I start changing the filter (even if I change it to a filter value which I know returns results on another chart).

**To Reproduce**
1.  Create a new screen
2. Add a new data provider component and select 'Users' as source and call it 'Users Data Provider'
3. Add a new data provider underneath the first one and select 'Users Data Provider' as data source and call it 'Active Users Data Provider'
4. Filter the second data provider on status active by using 'Define filter'
5. add a table underneath the second data provider and select 'Active Users Data Provider' as data source
6. check that the table shows no users
7. switch the data source on the table to 'Users Data Provider', this should show users and some of the users will be active
8. switch the data source on the table back to 'Active Users Data Provider'
9. modify the filter on  'Active Users Data Provider' to filter on status is empty or e.g. remove the filter
10. the table shows data, but the users don't all have an empty status

**Expected behavior**
The filter on the second nested data provider should work correctly

**Screenshots**
![image](https://user-images.githubusercontent.com/157020/189055402-c3ad504e-d9ea-4094-b751-abfbb8e9b61e.png)
  
**App Export**

[Test nested filters app-export-1662620980982.txt](https://github.com/Budibase/budibase/files/9524524/Test.nested.filters.app-export-1662620980982.txt)

**Desktop (please complete the following information):**
 - OS: macos
 - Browser: chrome, brave
 - Version: brave: Version 1.42.88 Chromium
 
**Additional context**

aptkingston:
Thanks for the detailed description and write up @mvhaen! I've got a PR up to fix this :+1:

aptkingston:
[...]

PR #7729, opened 2022-09-12
Fix client side lucene filtering

aptkingston:
## Description
This PR fixes client side lucene searching, which was currently broken. Client side filtering was not working due to recent changes in lucene filtering to allow multiple expressions on the same field (using numeric prefixes) not being accounted for in client side logic.

This also fixes a couple of minor display issues with component empty states (most notably the double empty state border around an empty data provider).

Addresses: 
- #7691

## Screenshots
Quick examples of empty state fixes:

Before:
image

After:
image

Before:
image

After:
image


shogunpurple (approved):
LGTM!

Issue #6491, opened 2022-06-28
Date Range is in the wrong component list - should be Data not Form. 


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.206
  - App Version: 1.0.206

**Describe the bug**
Date Range is in the wrong component category/list.

**Expected behavior**
Date Range should not be in the Form component list/category because it can be used outside of a Form and has nothing to do with forms. It is related to Data - it's a Data Provider filter!

**Screenshots**
![Screenshot 2022-06-28 at 18 08 31](https://user-images.githubusercontent.com/101575380/176242040-e9223318-d4c0-4dc5-bf34-a89f2b1e506b.png)

PR #7731, opened 2022-09-12
Move date range picker into data

melohagan:
## Description
Moved Date Range Picker into the *Data* section as it is a Data Provider filter and is not a form field. 

Addresses: 
- https://github.com/Budibase/budibase/issues/6491

## Screenshots
Screenshot 2022-09-12 at 09 21 43


mike12345567 (approved):
Good change!

Issue #6000, opened 2022-05-20
Implicit casts on foreign keys with different column types


mshiner:
**Describe the feature request**
When setting up relationships the foreign keys do not resolve if the column types do not match.  The specific use case was where the master table had a primamry key of type bigint but in the detail table the linking column was defined as type integer.

In postgreSQL the catalogs handle this and the FKs resolve gracefully with no problem.

As per the comment @melohagan in this [discussion](https://github.com/Budibase/budibase/discussions/5913#discussioncomment-2766221) it would be great if implicit and/or explicit casts could be handled

**Screenshots**
See [discussion](https://github.com/Budibase/budibase/discussions/5913) for more information.

mshiner:
Any movement on this ? Its a big issue for to use Budibase against our uzERP database

Martyn

melohagan:
I've changed this to a bug. As it turns out, the problem is that Knex returns a string for bigint types (probably because the max number size in JavaScript isn't large enough), and returns a number of smallint / int types.

When it comes to the relationship match, because it's a precise equality, it is not picking up the relationship, e.g. `1 !== '1'`

I'll get a PR out for the fix.

mshiner:
Mel

Great stuff - we've been doing some stuff with BB but TBH this was a major block for us. Good news that there may be a fix on the horizon.

Cheers
M

PR #7737, opened 2022-09-12
Postgres (and other external DBs) support foreign key bigint -> smallint

melohagan:
## Description
Currently relationships between bigint and smallint don't work because bigint is returned as type *string* whereas smallint is returned as type *number*.
When comparing relationships, the values are now cast as strings so that the equality check passes. 

Addresses: 
- https://github.com/Budibase/budibase/issues/6000

## Screenshots
**Owners ID**
pgadmin_bigint

**Pets fk_owner**
pgadmin_smallint

**Define relationship**
define_relationships

**BEFORE**
owners_nopets

**AFTER**
owners_success


shogunpurple (approved):
LGTM! good job @melohagan

Issue #7736, opened 2022-09-12
Azure App Service PORT variable


jonnymccullagh:
**Hosting**
Hosting Budibase via the Azure App Service using the Single Image build with TARGETBUILD=aas

**Describe the bug**
While deploying the container image budibase/budibase-aas in Azure App Service it appears as though Azure set many [environment variables](https://docs.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cpython). Those variables include a PORT variable which defaults to '80'. When the APP and WORKER try to start they are using the Azure provided PORT=80 variable before the custom APP_PORT=4001 or WORKER_PORT=4002.
The PM2 logs output that port 80 is already in use (by nginx):
```
/root/.pm2/logs/app-error.log last 15 lines:
0|app      | ction.Module._load (node:internal/modules/cjs/loader:839:12)","message":"listen EADDRINUSE: address already in use :::80","code":"EADDRINUSE","errno":-98,"syscall":"listen","address":"::","port":80}
0|app      | Unable to cleanup automation queue initially - Error: Connection is closed.
0|app      | error Command failed with exit code 255.
0|app      | bb-alert: Uncaught exception. : {"stack":"Error: listen EADDRINUSE: address already in use :::80\n
```
Separately there are some other issues with Azure App Service (SSH config) which I am addressing on branch name: ssh-for-aas.  

**To Reproduce**
To reproduce it is necessary to deploy the budibase/budibase-aas image in Azure App Service then SSH into the container to check the PM2 logs. 

**Expected behavior**
It would be good if we could check for the existence of the APP_PORT and WORKER_PORT variables to prioritise them over the PORT variable.

aderici:
very happy to see azure appservice getting first hand support!

mike12345567:
This is now released as of `v1.3.15` of the single/aas images.

PR #7739, opened 2022-09-12
Fix test case and port issue for AAS build

mike12345567:
## Description
Fix an intermittent test issue and fix a port issue that is affecting AAS images.

Addresses: 
- #7736 - worker/app port specific env var should be preferred to generic port env var.


shogunpurple (approved):
LGTM!

Rory-Powell (approved):

PR #7755, opened 2022-09-13
Azure App Service customisations

jonnymccullagh:
## Description
Deploying Budibase via Azure App Service (AAS) requires a few customisations. Notably settings for the SSH server and some tweaking was required to environment variables as the AAS defines some environment variables that can override our own variables. 
I have pulled master, added the changes from this PR and built and tagged an image. I then pushed the image to an Azure Container Registry and from there I deployed an App Service based on that image as shown below. 

## Screenshots
image


joebudi:
Nice work @jonnymccullagh

hosting/single/runner.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/bash
 declare -a ENV_VARS=("COUCHDB_USER" "COUCHDB_PASSWORD" "DATA_DIR" "MINIO_ACCESS_KEY" "MINIO_SECRET_KEY" "INTERNAL_API_KEY" "JWT_SECRET" "REDIS_PASSWORD")
-
+declare -a DOCKER_VARS=("APP_PORT" "APPS_URL" "ARCHITECTURE" "BUDIBASE_ENVIRONMENT" "CLUSTER_PORT" "DEPLOYMENT_ENVIRONMENT" "MINIO_URL" "NODE_ENV" "POSTHOG_TOKEN" "REDIS_URL" "SELF_HOSTED" "WORKER_PORT" "WORKER_URL")
+# Check the env vars set in Dockerfile have come through, AAS seems to drop them
+[[ -z "${APP_PORT}" ]] && export APP_PORT=4001
```
mike12345567, line 5: As these are set in the runner now, do we still need to set them in the Dockerfile? Just considering the double up here - theres a good chance people won't realise defaults need changed in two places.

mike12345567 (approved):
Looks good - only concern is the duplication between the runner and the Dockerfile - if we can do away with one of them that would be good - if we only set any environment variables which are needed within the Dockerfile process then we can perhaps keep most of the defaults to the runner.

jonnymccullagh:
> Looks good - only concern is the duplication between the runner and the Dockerfile
Yes good thinking I'll have another play with that tomorrow

jonnymccullagh:
Env vars now removed from Dockerfile to prioritise them in the `runner.sh` script. Tested:

- Non AAS normal build working
- AAS build tagged, pushed and re-deployed on Azure

Both appear to be working fine.

mike12345567 (approved):
LGTM!

PR #7758, opened 2022-09-13
Quick fix for CLI build

mike12345567:
## Description
Fixing CLI build - prebuilds required for hashing can be disabled via environment variable. This was an issue introduced in the plugin UI branch - hashing requires some prebuilds and there was some more hashing work added to the backend-core which caused it to break - allow swapping out for a pure JS build if needed.

Also included a small feature to disable parent directory creation - if the user is creating within an already empty directory, say a new git repo, they might not necessarily want to move all the files from the parent into the git repo, allowing disabling of this option is beneficial. By default this will still be done, and if any of the files exist in the current working directory then it will not move files to the cwd.

By default it will not ask this question if it realises it is in an empty git project (a directory with nothing but a `.git` directory within it).

### DIFF
--- a/packages/cli/src/utils.js
+++ b/packages/cli/src/utils.js
@@ -3,6 +3,7 @@ const fs = require("fs")
 const axios = require("axios")
 const path = require("path")
 const progress = require("cli-progress")
+const { join } = require("path")
 
 exports.downloadFile = async (url, filePath) => {
   filePath = path.resolve(filePath)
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils.js
```diff
@@ -67,3 +68,19 @@ exports.progressBar = total => {
 exports.checkSlashesInUrl = url => {
   return url.replace(/(https?:\/\/)|(\/)+/g, "$1$2")
 }
+
+exports.moveDirectory = (oldPath, newPath) => {
+  const files = fs.readdirSync(oldPath)
+  // check any file exists already
+  for (let file of files) {
+    if (fs.existsSync(file)) {
```
aptkingston, line 76: This always just checks the CWD, but since `newPath` is a param, we probably want to check instead if `join(newPath, file)` exists here.

aptkingston (commented):
LGTM! Just one tiny change required I think.

PR #7764, opened 2022-09-14
Allow component plugins in cloud

aptkingston:
## Description
This PR allows component plugins to be uploaded in cloud, but still blocks other types of plugins.


shogunpurple (approved):
LGTM!

Issue #7242, opened 2022-08-12
Postman, I was able to create an app with a special character '@' in the name, but I can't create one from apps portal 


MihailHadzhiev2022:
Navigate to Postman and create an application with a name Test@budi.com, we get status 200.
Navigate to Apps portal and tried to create an app with name Test1@budi.com, we get an error message "App name must be letters, numbers and spaces only" ![Screenshot 2022-08-12 at 10.38.44.png](https://images.zenhubusercontent.com/623866329f3f8e7223d95e00/051b5012-4719-4115-9973-7370a2beac84)![Screenshot 2022-08-12 at 10.42.43.png](https://images.zenhubusercontent.com/623866329f3f8e7223d95e00/9fef684a-052f-437f-bb9f-d8c03c55a479)

Issue #6836, opened 2022-07-21
Unable to filter date columns with Bindings in Automation


chaoticefx:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.1.17
  - App Version: 1.1.17
  
**Describe the bug**
When attempting to use a binding to filter a query in Automations, a console error is thrown as the binding does not match a datetime value, preventing bindings from being used in filtering.

**To Reproduce**
1. Create a table with a datetime column
2. Create a row in a table with any date
3. Create a new automation
4. Create a Query Rows step and select the table created in (1)
5. Press Define Filters, select any Datetime column, select a comparison with Binding and from the Handlebars or JS modal select Trigger.Timestamp
6. Press "Save" once back in the "Filtering" modal, console error is logged each time "Save" is pressed

**Expected behavior**
Binding should be accepted and provide a timestamp value at query time. Also, an error alert would be helpful if the "Save" button were to fail in such scenarios!

**Screenshots**
![Directus-2022-07-21-16-15-19](https://user-images.githubusercontent.com/79575739/180167390-3a3d97e4-c3d2-43e2-b638-6b01f3b35e92.jpg)
![Directus-2022-07-21-16-15-38](https://user-images.githubusercontent.com/79575739/180167406-7e890e49-51c6-426e-b2a3-da3f88325ee8.jpg)
![Directus-2022-07-21-16-16-02](https://user-images.githubusercontent.com/79575739/180167416-328d7c45-5e0f-4fdb-a197-3b973da1f31e.jpg)
![Directus-2022-07-21-16-16-20](https://user-images.githubusercontent.com/79575739/180167429-da43c960-22dc-417d-a732-814e74dde8a4.jpg)
![Directus-2022-07-21-16-17-01](https://user-images.githubusercontent.com/79575739/180167295-12cb1c25-af26-4639-81f2-274b693f8f8b.jpg)

**Desktop:**
 - OS: Ubuntu 18.04
 - Chrome
 - Version 103.0.5060.53 (Official Build) (64-bit)

lolcabanon:
Having the exact same bug right now!

Automate => Query rows => Define filters

Also, it is not possible to save "Match **any** of the following filters" in the `Define filters` screen. I can select it, then save, but if I reopen `Define filters` it is back to "Match **all** the following filters"

App version `1.3.11`

Hosted on budibase.app

PR #7766, opened 2022-09-14
Some bug fixes

PClmnt:
## Description
Bug fixes

Addresses: 
#6836  - Correct lucene filter wasn't being built up when filter was datetime with a binding, added a HBS check so we can avoid the specific parsing for datetime and build the lucene query as normal

#7242  - Added backend validation for the name when creating an app. Used the same regex as we do on the front-end in the creation modal, and added this to `applicationValidator` which is used by both our internal and public application API.  
Tested this by attempting to create an application via Postman with a number of special characters which were rejected with the correct error message. Creating an application without these worked correctly.


shogunpurple (approved):
LGTM! Nice fixes.

PR #7770, opened 2022-09-14
Data source -> datasource throughout code - plus friendly plugin name displayed

mike12345567:
## Description
Two minor changes requested by @melohagan:

1. Consistency in the use of the word `datasource` - throughout the UI/docs we sometimes would use `datasource`, sometimes `data source` - generally the combined single word is preferred and so I have sed replaced any usage with a space in it to get rid of the inconsistency.
2. In the plugin UI you couldn't easily see the friendly name of a plugin (that is used inside the Builder) displayed this in the table and the modal.

## Screenshots
image

image


melohagan (approved):
LGTM! Thanks for this 👍

PR #7774, opened 2022-09-14
Plugin init, imported and deleted events

mike12345567:
## Description
Adding types for plugin events, specifically for plugin init, import and deletion, this is across the CLI and the server.

Also includes an update to datasource events to attach whether or not the datasource is custom, via a boolean flag. This should help with looking up custom datasource usage through posthog.

Specifically the events will contain the following:
`plugin:init`
```
  type: "datasource" | "component"
  name: string
  version: string
  description: string
```

`plugin:imported`
```
  pluginId: string
  type: "datasource" | "component"
  source: "github" | "url" | "file" | "npm"
  name: string
  version: string
  description: string
```

`plugin:deleted`
```
  pluginId: string
  type: "datasource" | "component"
  name: string
  version: string
  description: string
```

### Further work
Down the line we might want to monitor the usage of custom components, this will need some more thought as today we don't track the usage of our own components. We may need to do this as a singular batch job from a read replica of Couch, where we run through and pick up component usage across all screens in all databases.


packages/types/src/documents/plugin/index.ts
```diff
@@ -14,4 +16,17 @@ export interface FileType {
   name: string
 }
 
+export interface Plugin extends Document {
```
Rory-Powell, line 19: Is this a global-db document? If so we should move the plugins folder into the global documents folder

mike12345567: Good point, missed that!

Rory-Powell (approved):
Just a wee note from the description
>  this will need some more thought as today we don't track the usage of our own components

We actually do track these events atm, they're the only ones sent from the client side 

LGTM 👍

PR #7789, opened 2022-09-15
Dev stack improvement + organisation page fixes

aptkingston:
## Description
This PR updates the dev stack to always restart the proxy service container when running the `dev` command. This ensures that any nginx conf changes are properly reflected, as currently swapping between branches with nginx conf changes does not work as the container does not restart, meaning it ignores changes in the mounted config.

This also fixes some incorrect usages of react style `className` attributes on the organisation settings page.

### DIFF
--- a/packages/builder/src/pages/builder/portal/settings/organisation.svelte
+++ b/packages/builder/src/pages/builder/portal/settings/organisation.svelte
@@ -88,14 +88,14 @@
       <Heading size="S">Information</Heading>
       <Body size="S">Here you can update your logo and organization name.</Body>
     </Layout>
-    <div className="fields">
-      <div className="field">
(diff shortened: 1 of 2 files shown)


packages/builder/src/pages/builder/portal/settings/organisation.svelte
```diff
@@ -88,14 +88,14 @@
       <Heading size="S">Information</Heading>
       <Body size="S">Here you can update your logo and organization name.</Body>
     </Layout>
-    <div className="fields">
```
shogunpurple, line 91: React! Blasphemy

shogunpurple (approved):
LGTM! This is a handy one for dev

PR #7795, opened 2022-09-15
Fix for REST query string URL encoding

mike12345567:
## Description
Fixing issue introduced by fix for #7683 - encoding the query string caused handlebars statements to break, this rectifies that.

Also minor fix for z-index issue with a tooltip.

Addresses: 
- Issue introduced by #7683.


packages/builder/src/helpers/data/utils.js
```diff
@@ -31,7 +32,8 @@ export function breakQueryString(qs) {
   let paramObj = {}
   for (let param of params) {
     const split = param.split("=")
-    paramObj[split[0]] = split.slice(1).join("=")
+    console.log(split[1])
```
shogunpurple, line 35: wee log left in

packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte
```diff
@@ -347,6 +347,7 @@
     const datasourceUrl = datasource?.config.url
     const qs = query?.fields.queryString
     breakQs = restUtils.breakQueryString(qs)
+    console.log(breakQs)
```
Comment (shogunpurple) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[selectedDatasource]/rest/[query]/index.svelte:350]: same here

packages/builder/src/helpers/data/utils.js
```diff
@@ -46,7 +48,19 @@ export function buildQueryString(obj) {
       if (str !== "") {
         str += "&"
       }
-      str += `${key}=${encodeURIComponent(value || "")}`
+      const bindings = findHBSBlocks(value)
```
shogunpurple, line 51: It would be really good to have a test for this, if possible

shogunpurple (approved):
LGTM - just a few log lines left in - but the actual parsing logic with HBS would be well covered with a test just for safety.

packages/builder/src/helpers/data/utils.js
mike12345567: Did my best to get jest tests working in the builder, had to remove some of the old (disabled) jest tests as they just don't work anymore and probably need re-thought.

This is probably something we need to extend out further - cypress is great but jest tests are nice for little unit test type things. I've made the (limited) builder tests part of CI so that they hopefully offer some value from here forward.

shogunpurple (approved):
LGTM! Thanks for putting together the jest setup 👌

PR #7801, opened 2022-09-16
Datasource hot reloading fix

mike12345567:
## Description
Minor fix - datasource hot reloading was working apart from the the CJS require caching, need to clear down the cache entry before reloading.

Thanks to @melohagan for finding this issue.


melohagan (approved):
LGTM! Thanks!! 👍

PR #7857, opened 2022-09-20
Add 'extra' validation for plugin datasource

melohagan:
## Description
Wanted to add the 'extra' field to my custom datasource but couldn't because the validation was missing.


mike12345567 (approved):
LGTM!

Issue #6847, opened 2022-07-21
Define relationship for external datasources allow you to select non-matching data types


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.1.20
  - App Version: 1.1.20

**Describe the bug**
When you **define relationship** for an external datasource, it allows you to select a non-matching field type for the foreign key. This will never work and is a little misleading for the user. 

**To Reproduce**
1. Create two Postgres tables. Both with an int id primary key. Add additional text fields (no foreign keys).
2. Fetch tables
3. Define relationship
4. Select the int id primary key
5. Select a text field for the foreign key
6. Save
7. Select one of the tables. No rows will be shown. See error in console tab `error: operator does not exist: integer = text`
 
**Expected behavior**
You should not be able to select a foreign key field when the types do not match.

**Screenshots**
![Screenshot 2022-07-21 at 16 46 13](https://user-images.githubusercontent.com/101575380/180257147-34c11a10-ef37-47d2-a6b8-5aa9f413bf91.png)

PR #7864, opened 2022-09-21
Do not allow non-matching types for primary/foreign keys - External DB

melohagan:
## Description
Previously you could select an Integer primary key against a Text foreign key and save successfully. This would cause issues fetching data later.
Added validation to prevent primary/foreign keys of non-matching types from being saved.

Addresses: 
- https://github.com/Budibase/budibase/issues/6847

## Screenshots
Screenshot 2022-09-21 at 09 13 17

Screenshot 2022-09-21 at 09 13 36

### DIFF
--- a/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
+++ b/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
@@ -90,16 +90,13 @@
     if (inSchema(toTable, toRelate.name, originalToName)) {
       errObj.toCol = colError
     }
-    if (
-      fromPrimary &&
-      fromRelate.fieldName &&
-      plusTables.filter(table => table.name === fromTable?.name)[0]?.schema[
-        fromPrimary
-      ]?.type !==
-        plusTables.filter(table => table.name === toTable?.name)[0]?.schema[
-          fromRelate.fieldName
-        ]?.type
-    ) {
+
+    let fromType, toType
+    if (fromPrimary && fromRelate.fieldName) {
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -90,6 +90,19 @@
     if (inSchema(toTable, toRelate.name, originalToName)) {
       errObj.toCol = colError
     }
+    if (
+      fromPrimary &&
+      fromRelate.fieldName &&
+      plusTables.filter(table => table.name === fromTable?.name)[0]?.schema[
```
mike12345567, line 96: This conditional is quite hard to look at, would it be possible to break this up?

Also, if you're looking for a single entry, generally its better to use `find` rather than `filter`.

Thinking could have:
```javascript
const fromTable = plusTables.find(table => table.name === fromTable?.name)
const toTable =  plusTables.find(table => table.name === toTable?.name)
let fromType, toType
if (fromPrimary && fromRelate.fieldName) {
  fromType = fromTable?.schema[fromPrimary]?.type
  toType = toTable?.schema[fromRelate.fieldName]?.type
}
if (fromType && toType && fromType !== toType) {
  errObj.foreign = "Column type of the foreign key must match the primary key"
}
```

melohagan: Good shout. I noticed as well that on line 120 there was already a binding var for **fromTable** and **toTable** which was handy!

mike12345567 (approved):

Issue #6960, opened 2022-07-28
Budibase Public API SDK


shogunpurple:
As a developer I want to have access to a budibase SDK that contains client code calling the budibase public API.

**COS**
- `@budibase/sdk` package that contains all client methods from public API
- Ship this by default in components built with the CLI
- Typescript stubs for .d.ts
- OpenAPI integration preferable

PR #7866, opened 2022-09-21
Public API SDK

aptkingston:
## Description
This PR adds a JS SDK for our public API. This SDK is entirely generated by [swagger-codegen](https://github.com/swagger-api/swagger-codegen) using the OpenAPI specs we generate.

The generated SDK uses promises and the generated methods have pretty names by way of the new `operationId` tag added to all endpoints.

### Publishing
The swagger-codegen generator is a Java app. Since adding java as a dependency to dev stacks (and release workflows) would be a nightmare, the generator instead runs inside a docker container - meaning docker is the only dependency. The generated code is ignored and not checked in to the repo, but we have a custom entrypoint in `src/index.js` which exposes only the relevant APIs and a handy configuration method.

The SDK is not used by any other Budibase packages, so we can exclude it from the usual `build` and `dev` scripts. There's a new top level yarn script called `build:sdk` which will generate and bundle the SDK. This script has been added to only those release jobs which publish to NPM. I think they already have `docker` available to them inside the `ubuntu-latest` runner.

### Stucture
The SDK exports a default class which is instantiated to form an instance of the SDK. It takes configuration parameters via it's constructor, for example:
```js
import Budibase from "@budibase/sdk"

const budibase = new Budibase({
  apiKey: "my-api-key",
  host: "https://bb.budibase.app",
})
```

### Code completion / IDE intellisense
As we aren't wrapping the generated endpoints, IDEs can take full advantage of the excellent generated docs and provide great code completion prompts:
image

### Configuration
The SDK is configured with only 2 properties:
- API key (acquired from the Budibase portal or programmatically inside plugins)
- Host (the server where your API is running - defaults to the current host)

> **The current public API does not properly configure CORS, so using a different `host` will prevent the SDK from working. I'm assuming this is just an oversight and we should add this? I'll follow up on this.**

### Usage inside component plugins
There is a new utility provided by the client SDK to make authentication easier inside plugins. There's a simple `getAPIKey` utility which will return the users API key, which can then be passed into the `configure` method.

A quick example would look like this. Note that we *do not* need to specify the `host` configuration parameter, as we expect the public API to be running on the same host that this component is being served from (which will always be the case for component plugins, unless you want to contact a different Budibase install from the one you're running on).
```js
import Budibase from "@budibase/sdk"

const { getAPIKey } = getContext("sdk")

const apiKey = await getAPIKey()
const budibase = new Budibase({ apiKey })
await budibase.rows.search()
```

The public API SDK is *not* shipped by the client library be default. This means it is opt-in to use it, and plugin authors can version it themselves.

### Questions
- The generated SDK includes `WithHttpInfo` suffixed versions of all API endpoints. This is unavoidable and a consequence of generating a promisified SDK (as opposed to a callback based one, which is the default). We cannot remove these functions as the normally named SDK functions call through to these. Is that an acceptable tradeoff that we just acknowledge they exist and say to just ignore them? You can see an example in the first screenshot above, where there is `createRow` and `createRowWithHttpInfo`.
- With this being released, I think the current client library context (called `sdk`) needs to be renamed. The double usage of `sdk` is confusing. I would propose we rename the client library context to just `client` - i.e. `const { styleable } = getContext("client)"`. Any thoughts on that?

### Addresses
- #6960


packages/sdk/src/index.js
```diff
@@ -0,0 +1,22 @@
+import * as BudibaseApi from "../sdk"
```
Rory-Powell, line 1: This is great cheeks, so good to have the api docs generating our client like this. 

I was wondering would there be a more fluent way we could use this client, from the docs above we have:
```js
import { configure, ApplicationsApi } from "@budibase/sdk"
```
I think it would be really nice if we could import it like 
```js
import Budibase from "@budibase/sdk"
```


Similarly for the usage, atm we have this:
```js
await ApplicationsApi.applicationsSearchPost
```

It could be more fluent to use a single import for budibase and not then need to pull in the individual api clients. So it could look something like this:
```js
await Budibase.applications.search()
await Budibase.applications.delete()
await Budibase.queries.execute()
```

I'm a fan of this approach as it requires less manual intervention with imports when I want to use different parts of the sdk and seems to be more consistent with api clients i've used from npm. It would reduce some duplication in calling code 

e.g. **Applications**Api.**applications**SearchPost could be condensed by a single `applications` namespace on the client. 


What are your thoughts on this type of usage?

aptkingston: Yea it would be super easy to do a default export of everything, which solves the first point. You could do it already with `import * as Budibase from "@budibase/sdk"` but a default import is probably nicer. I'll add that. 

For the naming of the different APIs themselves, that's also totally in our control. Happy to go for a terser naming convention here so I'll update that.

The one I'm a little more on the fence about is how explicit we are with the actual endpoint names. Personally I don't mind being explicit with these, but I agree that shorter looks nicer of course. Just an FYI that we do at least have them renamed to `searchRows` for example rather than `rowsSearchPost` which is the default! But I'm happy to go either way on this and let other people weigh in too.

mike12345567:
On the comment:
> The generated SDK includes WithHttpInfo suffixed versions of all API endpoints. This is unavoidable and a consequence of generating a promisified SDK (as opposed to a callback based one, which is the default).

Makes sense these can't be removed due to them being called through to my the SDK itself - I was just wondering if there is anyway we could remove the `export` of them however? If they weren't exported I *think* most IDEs will realise they aren't available functions and wouldn't be part of the code completion. If its not possible I think the benefit of being able to generate the SDK far outweighs the negative of a few poorly named/useless functions being exposed, we can just explain that in the documentation, certain functions ending in `withHttpInfo` are not to be used as they are for internal use only.

shogunpurple (approved):
It's mind blowing that this works almost out of the box, and will allow us to generate SDKs for other programming languages as well in future.

>The generated SDK includes WithHttpInfo suffixed versions of all API endpoints. 

It would be good to hide these if possible, if they provide no value. However are they definitely useless? Is there absolutely no chance you would ever use them? Even some weird case?

> With this being released, I think the current client library context (called sdk) needs to be renamed.

Yeah to be honest it's not really an SDK as was originally planned - it's more just the internal API of the client library now so think this makes sense. `client` works, or maybe we could be more explicit and call it `core` - as that's really what it is.

packages/sdk/src/index.js
shogunpurple: 👍 for the single, top level import. Since we have a single call to `configure`, it's easier to reason about IMO if we have one top level object that everything hangs off. This also very closely maps to a lot of our internal stuff as well. That said I'm also not against:

`Budibase.searchApplications()`
`Budibase.deleteApplication()`
`Budibase.executeQuery()`

aptkingston:
> It's mind blowing that this works almost out of the box, and will allow us to generate SDKs for other programming languages as well in future.
> 
> > The generated SDK includes WithHttpInfo suffixed versions of all API endpoints.
> 
> It would be good to hide these if possible, if they provide no value. However are they definitely useless? Is there absolutely no chance you would ever use them? Even some weird case?
> 
> > With this being released, I think the current client library context (called sdk) needs to be renamed.
> 
> Yeah to be honest it's not really an SDK as was originally planned - it's more just the internal API of the client library now so think this makes sense. `client` works, or maybe we could be more explicit and call it `core` - as that's really what it is.

> 

They're definitely useless yea. They accept the exact same params as their equivalent, normal named counterparts. The only difference is that the `WithHttpInfo` suffix version actually does the logic, and the short-named variant literally just calls through to the long variant and follows it with a `then(x => x.data)`. So they basically exist to just strip the `.data` property out of the response, which is completely pointless. They only do this when you generate a promisified JS SDK though, not the callback version.

I'll keep trying to see if there's a way to keep IDE code completion while hiding these (and keeping the logic generic so we don't need to ever update this as we update the public API), but so far I don't see a way to have it all.

aptkingston:
I've updated the description to show the improved syntax, following some of @Rory-Powell's suggestions. The new example usage would look like:
```js
  import Budibase from "@budibase/sdk"

  const budibase = new Budibase({
    apiKey: "my-api-key",
    host: "https://bb.budibase.app",
  })

  await budibase.rows.create()
  await budibase.applications.search()
  await budibase.queries.execute()
```

This is easier to use, allows multiple clients, and follows the example of a lot of similar public SDKs, while still being zero maintenance over time. The only time we'll need to touch this SDK code is if we add a new module to the public API entirely (e.g. a groups API). Adding/updating/deleting endpoints within existing APIs requires no changes.

packages/sdk/src/index.js
Rory-Powell: That `delete` is so annoying, we've ran into it a few times. The `call` bit might be troublesome for people relying on intelisense, as if they were to type the characters `de` looking for `delete` they're less likely to come across the right function.

Airtable uses a `destroy` function, and stripe uses `del`. I assume they've been hit by the same problem.

`destroy` might be the best alternative here as `del` might look strange in anything other than a js client e.g. generated docs.  I'm also aware that I've asked you for many changes at this point and would like to apologise in advance

aptkingston: `destroy` sounds good to me :ok_hand: and don't worry - this is exactly the kind of feedback/suggestions I was after!

mike12345567 (approved):
This is incredible work, its so cool to see the specs running all the way for the JSDoc comments around API endpoints to a real, complete and highly usable SDK at the end of it!

Will be interesting to try generating say the Python SDK in the future as well, it would be very interesting to know if out of the box its basically correct off the back of this work - again its amazing that we can even talk about just *generating a Python SDK*!

packages/sdk/src/index.js
aptkingston: > That `delete` is so annoying, we've ran into it a few times. The `call` bit might be troublesome for people relying on intelisense, as if they were to type the characters `de` looking for `delete` they're less likely to come across the right function.
> 
> Airtable uses a `destroy` function, and stripe uses `del`. I assume they've been hit by the same problem.
> 
> `destroy` might be the best alternative here as `del` might look strange in anything other than a js client e.g. generated docs. I'm also aware that I've asked you for many changes at this point and would like to apologise in advance

Updated to use `destroy` for delete operations :+1:

aptkingston:
This should be ready to merge whenever now - all the comments have been addressed :+1: More issues may come out of using it but we'll get a better idea once this is published to npm for the first time, since it'll be a lot easier to test in different environments (like inside a custom component).

Issue #7878, opened 2022-09-21
Error using data provider filter binding with 'Is In' condition


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7712

<div type='discussions-op-text'>

<sup>Originally posted by **applicationuser** September  9, 2022</sup>
Hello,

Error applying filter 'is in' condition for a data provider with Binding option. When i click on Bolt icon error in console and cannot continue.  need to refresh the page. 

We want to filter data provider based on user selected options from a multi select control 

Budibase cloud
app version: 1.3.11
table source : budibase 

Console error screenshot : 
![image](https://user-images.githubusercontent.com/42922833/189411229-ccb4dc11-e831-4b12-b5b9-7a147aace871.png)

let me know you need any other info

Thank You</div>

![Screenshot 2022-09-21 at 15 19 35](https://user-images.githubusercontent.com/101575380/191529402-ecde0c00-d1ff-483e-8f1a-1d929926eaae.png)

![Screenshot 2022-09-21 at 15 19 52](https://user-images.githubusercontent.com/101575380/191529468-19a42a0d-e4f3-4589-a341-83931f855955.png)

PR #7879, opened 2022-09-21
Is in filter - only call startsWith if string type

melohagan:
## Description
Crash caused by trying to call `startsWith` on array.

Addresses: 
- https://github.com/Budibase/budibase/issues/7878


shogunpurple (approved):
LGTM!

aptkingston (approved):
I'm confused how this could break, because how can a value not be a string here? Perhaps with other operators it's possible to have a value that's an array, if we were letting them pick options from a multi select. If that's the case then I feel we should be resetting the expression value whenever the binding type changes (e.g. a binding type can only be a string), since there might be similar issues to this elsewhere. 

But in any case, this one is good to go in for now!

aptkingston:
> @aptkingston I think it broke because of this commit: [350e023](https://github.com/Budibase/budibase/commit/350e0233e201e2441b8e4face337f13e3338640b)
> 
> Basically the **Is in** filter always expects the expression value to be an array.

Yep that would do it! We should really follow up that `if` statement with something like
```js
else if (operator !== Constants.OperatorOptions.In.value && Array.isArray(expression.value)) {
  expression.value = null
}
```
so that we clear the array again for any other operator. That would also fix this bug, but the current fix is totally fine to merge too. This additional check would just prevent other cases like this from happening.

melohagan:
> > @aptkingston I think it broke because of this commit: [350e023](https://github.com/Budibase/budibase/commit/350e0233e201e2441b8e4face337f13e3338640b)
> > Basically the **Is in** filter always expects the expression value to be an array.
> 
> Yep that would do it! We should really follow up that `if` statement with something like
> 
> ```js
> else if (operator !== Constants.OperatorOptions.In.value && Array.isArray(expression.value)) {
>   expression.value = null
> }
> ```
> 
> so that we clear the array again for any other operator. That would also fix this bug, but the current fix is totally fine to merge too. This additional check would just prevent other cases like this from happening.

I made this change, but I also had to add in an **onChange** handler for the value type changing to get this to work without the original bug fix.
With that said, I left in the string type check for good measure.

aptkingston:
The logic here is becoming a bit gnarly haha. I might jump on and see if it could be refactored a bit to reduce duplication and simplify things, just because this is now the sort of file that you could come back to in a few months and have no idea what's going on anymore!

melohagan:
> The logic here is becoming a bit gnarly haha. I might jump on and see if it could be refactored a bit to reduce duplication and simplify things, just because this is now the sort of file that you could come back to in a few months and have no idea what's going on anymore!

That is true; thanks for taking a look! Handling array types within the filters has been a bit awkward.

aptkingston:
@melohagan I've refactored the validation logic here which should make things much simpler to read and maintain over time! Do you want to have a look and see what you think? The only thing I didn't keep (deliberately) was the automatic conversion of comma-separated strings into real array values, as that can end up generating invalid selected options in the multi-select picker if the user just happens to have a comma inside a normal string.

melohagan:
That looks great @aptkingston much cleaner, thank you!

NAB: The methods are called *validateX* e.g. `validateValue` but really it's cleaning or sanitising.

aptkingston:
> That looks great @aptkingston much cleaner, thank you!
> 
> NAB: The methods are called _validateX_ e.g. `validateValue` but really it's cleaning or sanitising.

Sanitize sounds accurate to me! Have updated it.

Issue #4385, opened 2022-02-08
SMTP Automation Block Updates


mjashanks:
The SMTP automation step should allow sending of multiple recipients.

I suggest that we allow comma-separated emails, e.g.  `bill@microsoft.com, steve@microsoft.com, {{ Step 1.Row.EmailAddress }}`

**COS**
- Allow the SMTP block to take a comma separated list of recipients
- Allow "CC" field in SMTP automation, which can take a comma separated list of recipients also

PR #7881, opened 2022-09-21
Add CC and BCC (with multiple email support) to Send Email automation action step

melohagan:
## Description
You can now use CC and BCC in email automations.

Addresses: 
- https://github.com/Budibase/budibase/issues/4385

## Screenshots
Screenshot 2022-09-21 at 16 10 00

Screenshot 2022-09-21 at 16 10 26

Screenshot 2022-09-21 at 16 11 01

### DIFF
--- a/packages/worker/src/api/routes/global/email.js
+++ b/packages/worker/src/api/routes/global/email.js
@@ -12,7 +12,13 @@ function buildEmailSendValidation() {
   return joiValidator.body(Joi.object({
     email: Joi.string().email({
       multiple: true,
-      }),
+    }),
+    cc: Joi.string().email({
+      multiple: true,
+    }),
+    bcc: Joi.string().email({
+      multiple: true,
+    }),
     purpose: Joi.string().valid(...Object.values(EmailTemplatePurpose)),
     workspaceId: Joi.string().allow("", null),
     from: Joi.string().allow("", null),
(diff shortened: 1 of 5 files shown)


packages/worker/src/api/routes/global/email.js
```diff
@@ -12,7 +12,13 @@ function buildEmailSendValidation() {
...
     email: Joi.string().email({
       multiple: true,
-      }),
+    }),
+    cc: Joi.string().email({
+      multiple: true,
+    }),
+    bcc: Joi.string().email({
+      multiple: true,
+    }),
```
Rory-Powell, line 21: should these have `.optional()` appended?

melohagan: Nice catch! I tried optional but it doesn't change the behaviour. Changed to `allow("", null)` instead to match the other 'optional' fields.

Rory-Powell: just to confirm, can we send a request to this endpoint with those fields missing completely?
that's odd on optional, it usually works. Did you try it like this?
```ts
Joi.string().email().optional()
```

melohagan: Yeah that's how I tried it, but it was already optional without it. The allow `null` means that null values can be sent through.

The following scenarios tested to work, the request went through:
 - Blank fields
 - JavaScript returning `undefined`
 - JavaScript returning `null`

Rory-Powell: sorry just to confirm again on Blank fields, do you mean with the value `""` or with the field omitted completely?

melohagan: both!

Rory-Powell (approved):
LGTM

Issue #7877, opened 2022-09-21
Error when importing plugin - have to leave Plugins page and return


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.4.4
  - App Version: 1.4.4

**Describe the bug**
Importing a datasource plugin throws an error. You have to leave the plugins page and return to see it listed.

**To Reproduce**
1. Import plugin via Github URL (Tarball URL also has the same problem)
2. See error
3. Navigate away from plugins page and back to see it listed

**Expected behavior**
I should see no error when importing a plugin

**Screenshots**
![plugin](https://user-images.githubusercontent.com/101575380/191522740-8ae64d2b-118d-4d77-b460-97a5465fa281.gif)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome

**Additional context**
I have the `.env` var set, and had deleted all previously listed plugins before importing from Github.

PR #7883, opened 2022-09-21
Fix for error when adding a plugin

PClmnt:
## Description
The server now returns a response like `plugin: {...}` when creating a plugin, so needed to adjust the svelte store to handle this

## Address
 - https://github.com/Budibase/budibase/issues/7877


melohagan (approved):
LGTM! Thanks! 👍

Issue #7372, opened 2022-08-19
Clicking on an attachment inside a table component checks the "row selected" checkbox.


Klience:
**Hosting**
- Cloud v.1.2.45

**Describe the bug**
Clicking on an attachment inside a table component (underlined link) checks the "row selected" checkbox.

**To Reproduce**
1. Create a Budibase table with an attachment column
2. Upload an attachment (any)
3. Setup a data provider with a table component
4. Click the underlined link inside the table component (that leads to downloading or viewing the attachment)
5. See how the row gets selected.

**Expected behavior**
Clicking an attachment download / view link inside the table component should not automatically select the row. As this has nothing to do with the row selection.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 104.0.5112.81 (Official Build) (64-bit)

**Additional context**
I believe it tracks the cursor position and selects the row when the mouse cursor is over the row while clicking the link OR the code thinks that you click on the row while clicking on the attachment link.

Note that this doesn't happen when we add a "link" component inside the "table" component. In this case the link gets spawned to the left of the row, between the checkbox and the row. And clicking on the link doesn't select the row.

PR #7891, opened 2022-09-22
Fix/table cell attachment click propagation

deanhannigan:
## Description
Minor fixe to stop event propagation when clicking on an attachment link in a table row. This avoids row selection in tables with the selection enabled.

Addresses: 
- https://github.com/Budibase/budibase/issues/7372

### DIFF
--- a/packages/bbui/src/Table/AttachmentRenderer.svelte
+++ b/packages/bbui/src/Table/AttachmentRenderer.svelte
@@ -20,6 +20,9 @@
       target="_blank"
       download={attachment.name}
       href={attachment.url}
+      on:click={e => {
+        e.stopPropagation()
+      }}
     >
       <div class="center" title={attachment.name}>
         <img src={attachment.url} alt={attachment.extension} />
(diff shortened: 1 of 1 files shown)


packages/bbui/src/Table/AttachmentRenderer.svelte
```diff
@@ -20,6 +20,9 @@
       target="_blank"
       download={attachment.name}
       href={attachment.url}
+      on:click={e => {
```
shogunpurple, line 23: I think `on:click|stopPropagation` may be equivalent here in svelte!

shogunpurple (approved):
LGTM! worth trying the shorthand just to see if it works

Rory-Powell (approved):

packages/bbui/src/Table/AttachmentRenderer.svelte
deanhannigan: Yep, I thought the same but the shorthand is reserved for native elements.

Issue #7903, opened 2022-09-22
Budibase slowly grinds to a halt on design section


mjashanks:
**Hosting**
- Cloud
- Tenant ID: repro'd on both `bb.budibase.app` and `bbdemo.budibase.app`

**Describe the bug**
When I start using the design section for an extended period, it gets slower and slower. Over the course of 5-10 minutes it grinds to halt and becomes totally unusable. 

A page refresh temporarily fixes

**To Reproduce**
1. Go to the design section
2. Open your dev tools, watch the network tab
3. Look out for a call to `/socket/client/?EIO=4&transport=polling&t=xxxx`

The call times out after a minute. Eventually, I have lots of failed and lots of pending requests. After a few minutes, I am unable to add / delete / update components.

**Screenshots**
![image](https://user-images.githubusercontent.com/3524181/191762084-d86ea9a5-86d5-4efa-89b7-39e60dc639b0.png)

 **Desktop:**
 - OS: Ubuntu
 - Browser: Chromium
 - Version: 105.0.5195.102

**Additional context**
My app is a very simple app - the standard generated screens.

aptkingston:
These are requests from the websocket library - the polling indicates that the socket is not connecting properly (which is an issue - it should be and it does in self host so this looks like an env issue) so it's attempting to poll for changes, but this shouldn't cause any sort of slowdown of the browser.

shogunpurple:
Fix: https://github.com/Budibase/budibase/pull/7907

shogunpurple:
Should now be resolved!

maiducan:
My design section is still very lagging and slow after the newest update :/

PR #7907, opened 2022-09-22
turn off websockets in client lib in cloud

shogunpurple:
## Description
Turn off initialisation of websocket client in cloud as it does not add anything and is causing issues:

Addresses: 
- https://github.com/Budibase/budibase/issues/7903#issuecomment-1255053348

## Screenshots

### DIFF
--- a/packages/client/src/websocket.js
+++ b/packages/client/src/websocket.js
@@ -1,12 +1,13 @@
-import { builderStore } from "./stores/index.js"
+import { builderStore, environmentStore } from "./stores/index.js"
 import { get } from "svelte/store"
 import { io } from "socket.io-client"
 
 export const initWebsocket = () => {
   const { inBuilder, location } = get(builderStore)
+  const { cloud } = get(environmentStore)
 
   // Only connect when we're inside the builder preview, for now
-  if (!inBuilder || !location) {
+  if (!inBuilder || !location || cloud) {
     return
   }
 


Rory-Powell (approved):

aptkingston (approved):
Probably still worth investigating why the websocket wasn't connecting in cloud (because it should) but I agree it adds nothing in cloud so makes sense to disable it.

Rory-Powell:
I wonder is this related to multi instances behind the load balancer. Maybe the connection is being opened initially to one instance, with subsequent communication is being routed through the generic address of bb.budibase.app finding the connection no longer is alive

aptkingston:
> I wonder is this related to multi instances behind the load balancer. Maybe the connection is being opened initially to one instance, with subsequent communication is being routed through the generic address of bb.budibase.app finding the connection no longer is alive

Yea could definitely be that. I assume we haven't needed to enable any sort of session stickiness until now.

aptkingston:
> @aptkingston it's weird because as I showed you it semi-works for me, I never fully lose connection to the websocket, just some requests.
> 
> Screenshot_2022-09-22_at_14 57 17

That maybe supports the idea though that the requests which make it through to the original node work, and the rest don't? Although strange that it fails 100% of the time for me.

shogunpurple:
> 
I wonder is this related to multi instances behind the load balancer. Maybe the connection is being opened initially to one instance, with subsequent communication is being routed through the generic address of bb.budibase.app finding the connection no longer is alive

This is a good shout - I think we need to repro this in a clustered env in one of our test environments and see if we can get websockets working across pods

Rory-Powell:
good advice on multi nodes / stick sessions
https://socket.io/docs/v4/using-multiple-nodes

there's steps in here for nginx, not sure about how that will also work with amazons ELB. I assume there is similar configuration

PR #7909, opened 2022-09-22
Reduce websocket polling frequency and cap reconnection attemps

aptkingston:
## Description
This PR reduces the frequency of polling whenever the client websocket fails to connect.

The new config limits it to:
- 10 reconnection attempts total
- 5 second initial reconnection, followed by every 10 seconds after that
- Timeout each attempt after 5 seconds to stop any request stacking

After 95 seconds total (1x5 sec reconnection attempt and 9x10 second attempts) it will give up attempting to connect and do no more requests.

Quick example showing it stopping after 10 attempts total (and timed to ensure config works):
image


Rory-Powell (approved):
LGTM 👍

Issue #6532, opened 2022-07-03
Dragging \ clicking a form step within a form step crashes the browser tab


FlaminWrap:
**Hosting**

- Self
  - Method: budi-cli
  - Budibase Version: 1.0.214
  - App Version: 1.0.214

**Describe the bug**
Dragging \ clicking a form step within a form step crashes the browser tab

**To Reproduce**
1. Create a form
2. Create two form steps
3. Drag one form step so it is within the other form step
4. Browser tab will become unresponsive

**Expected behavior**
A form step is not allowed to be placed within a form step

**Screenshots**
![image](https://user-images.githubusercontent.com/97764630/177033076-56e2c0d3-111e-4563-afdc-03c9f02c6611.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 103.0.5060.66

**Additional context**
You can delete the form step if you click the ... menu to the right of the form step and delete it.
Exported test app
[Form Step-export-1656839347480.txt](https://github.com/Budibase/budibase/files/9034412/Form.Step-export-1656839347480.txt)

PR #7913, opened 2022-09-22
Drag/drop in the builder navigation enforces illegal children config in the component definitions

deanhannigan:
## Description
Ensure that `illegalChildren` config is enforced when dragging and dropping components in the builder component tree.

Addresses: 
- https://github.com/Budibase/budibase/issues/6532

## Screenshots
slim_clip_AdobeExpress (1)


packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/navigation/ComponentTree.svelte
```diff
@@ -68,11 +70,34 @@
...
+  const onDrop = async (e, component) => {
     e.stopPropagation()
     try {
+      const compDef = store.actions.components.getDefinition(
+        $dndStore.source?._component
+      )
+      const compTypeName = compDef.name.toLowerCase()
+      const path = findComponentPath(currentScreen.props, component._id)
+
+      path.forEach(pathComp => {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/navigation/ComponentTree.svelte:82]: If we replace this with `for (let pathComp of path) {... }` then we can escape the loop and function with a `return` statement which lets my suggestion below work.

aptkingston (commented):
LGTM! Just one small suggested refactor :+1: 

Also great idea taking a look at this - it makes the `illegalChildren` attribute much more useful.

Next we will have to deal with the copy/pasting edge cases and moving up/down :laughing: but this is a great start and the most common flow!

Issue #7906, opened 2022-09-22
Automations - Unable to bind to the result of a JS Scripting block


mjashanks:
**Hosting**
Cloud

**Describe the bug**
When using the JS Scripting block in automations, I am unable to bind to the result. This only happens when the previous step has "Add Looping"

The output of a JS script block is

```
{ "value": "...", "success": true }
```

When I test my automation, the JS correctly outputs this structure - exactly what I expect.

However, when I try to use the binding drawer on a subsequent step, I'm presented with the options:
- `Steps.3.Rows`
- `Steps.3.Success`

If I manually type the binding `steps.1.value` - it does not work.

**To Reproduce**
1. Create a new automation
2. Add a `Query Rows` block
3. Add Looping on Query Rows - just loop on a random string for ease. Happens with Array too
4. Add a JS block - `return "test";`
5. Add any other block - I used a conditional.
6. Try and bind to the javascript block - at this point you should notice that you are not offered the JS Block schema
7. Manually type `{{ steps.2.value }}`  - which is the correct binding of the JS block result 
9. Test your automation 

You will see that the `{{ steps.2.value }}` does not bind correctly.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chromium
 - Version: 105

PR #7944, opened 2022-09-23
Automate fixes

PClmnt:
## Description
Fixes a few automation issues. 

- Bindings were not being shown correctly when a loop block was added
- When opening the binding drawer from the filter drawer in a Query Rows step, the Panel was not the correct width.
- Some incorrect text in the queryRows automation definition that mentioned deleting of rows
- Fixed a type error that was appearing in the console when accessing the bindings drawer. 

Addresses: 
#7906


shogunpurple (approved):
LGTM!

PR #7945, opened 2022-09-23
Fix to ensure licensing flags display correctly

deanhannigan:
## Description
Some fixes and refactoring for the feature flags used when displaying licensing overlays.

### DIFF
--- a/packages/builder/src/components/portal/licensing/LicensingOverlays.svelte
+++ b/packages/builder/src/components/portal/licensing/LicensingOverlays.svelte
@@ -8,7 +8,6 @@
   import { ExpiringKeys } from "./constants"
   import { getBanners } from "./licensingBanners"
   import { banner } from "@budibase/bbui"
-  import { FEATURE_FLAGS, isEnabled } from "../../../helpers/featureFlags"
 
   const oneDayInSeconds = 86400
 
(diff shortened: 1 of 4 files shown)


packages/builder/src/components/portal/licensing/LicensingOverlays.svelte
```diff
@@ -82,31 +81,25 @@
...
-      }
-    })
-
-    licensing.subscribe(state => {
-      if (state.usageMetrics && !licensingLoaded) {
-        licensingLoaded = true
-      }
-    })
-    loaded = true
+    domLoaded = true
```
Rory-Powell, line 102: Out of curiosity how come we need the additional `domLoaded` flag?

Rory-Powell (approved):
LGTM 👍 are there still changes to go into this one?

packages/builder/src/components/portal/licensing/LicensingOverlays.svelte
deanhannigan: I removed flag initially and then, when attempting to show the licensing modals, it would periodically fail because they hadn't fully initialised. I changed the name to domLoaded just for clarity.

PR #7949, opened 2022-09-23
Tenancy awareness for plugin bucket

mike12345567:
## Description
Adding tenancy awareness for plugin bucket - most of the work for this is handled on the pro side.


shogunpurple (approved):
LGTM!

Issue #7952, opened 2022-09-24
Custom component development causes `UsageLimitError: Licensed Plugins of 10 has been exceeded`


andreasbhansen:
**Hosting**
- Self
  - Method: docker compose
  - Node: v16.13.2 
  - Budibase Version: CLI v1.4.7 (globally installed)
  - App Version: Installed v1.4.7

**Describe the bug**
I'm working on a custom component using `yarn watch` and the live-reload feature. After a few saves and live-reloads, the component in the builder is no longer updated with the latest code changes. In the Docker `bbapps` logs window, I see the following error:

```
Importing plugin: /plugins/budi-datatable/dist/budi-datatable-1.0.0.tar.gz

Error updating usage quotas for plugins UsageLimitError: Licensed Plugins of 10 has been exceeded
at /app/node_modules/@budibase/pro/dist/index.js:1:14818
at Generator.next (<anonymous>)
at r (/app/node_modules/@budibase/pro/dist/index.js:1:2449)
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  code: 'usage_limit_exceeded',
  type: 'license_error',
  status: 400,
  limitName: 'Plugins'
}

Failed to import plugin: Licensed Plugins of 10 has been exceeded
```

It doesn't matter if I'm working with a template or a project made from scratch, the error is the same.
I only have this one custom plugin/component, and it is the only component on the Budibase Portal's Plugin page.

**To Reproduce**
1. Setup everything related to Docker and Budibase CLI.
2. Create a new custom component and drag it into an app's designer/builder view.
3. Make sure to have `yarn watch` running. 
4. Do an unknown amount of saves to the component, which at some point triggered the above error for me, and I can no longer visually see changes to the custom component I'm working on.

**Expected behavior**
Having the custom component live-reload as expected, and not see the licensing errors described above. 

**Screenshots**
Nothing else to show than the error stack above. 

**Desktop (please complete the following information):**
 - OS: MacOS 12.3.1
 - Browser: Chrome
 - Version: 105.0.5195.102 (Official Build) (x86_64)

mike12345567:
Hi @andreasbhansen this issue should have been resolved with the recent releases of Budibase.

PR #7953, opened 2022-09-24
Fix for plugin quotas

mike12345567:
## Description
Adding a migration for the plugin quotas to make sure they are always accurate in self host at startup. The actual fix for this in on the pro side, PR: https://github.com/Budibase/budibase-pro/pull/61

Addresses: 
- #7952 - plugin quotas were not being correctly evaluated when updating existing plugins.

### DIFF
--- a/packages/server/src/migrations/functions/pluginCount.ts
+++ b/packages/server/src/migrations/functions/pluginCount.ts
@@ -1,0 +1,17 @@
+import { tenancy, logging } from "@budibase/backend-core"
+import { plugins } from "@budibase/pro"
+import env from "../../environment"
+
+export const run = async () => {
+  // only a self hosted op
+  if (!env.SELF_HOSTED) {
+    return
+  }
+  try {
+    await tenancy.doInTenant(tenancy.DEFAULT_TENANT_ID, async () => {
+      await plugins.checkPluginQuotas()
+    })
+  } catch (err) {
+    logging.logAlert("Failed to update plugin quotas", err)
+  }
+}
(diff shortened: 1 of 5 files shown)


shogunpurple (approved):
LGTM!

packages/server/src/migrations/functions/pluginCount.ts
```diff
@@ -0,0 +1,17 @@
+import { tenancy, logging } from "@budibase/backend-core"
+import { plugins } from "@budibase/pro"
+import env from "../../environment"
+
+export const run = async () => {
+  // only a self hosted op
+  if (!env.SELF_HOSTED) {
```
Rory-Powell, line 7: Is it possible there are inaccurate plugin counts in the cloud also? Or why would we not run there?

Rory-Powell: If we did want to retain this check we should probably do it outside the migration, in the case statement for building migrations. 

That means the record of the migration run will never be recorded, which is better in case we do want to then run this in cloud.

mike12345567: no plugins are currently unavailable in the cloud due to another bug.

Rory-Powell: Cool np. Still the same point about not actually running the migration,we probably shouldn't populate the record unless the migration actually runs.

Rory-Powell: Is it possible the quota got incorrect set for any tenants while plugins were available in cloud?

mike12345567: plugins have never actually been available in the cloud - there was an error meaning that it was impossible to add plugins - which will be fixed as part of this release.

PR #7970, opened 2022-09-26
Fix issues with multi tenant app IDs in groups UI

aptkingston:
## Description
This PR fixes a couple of issues with the groups UI which were only happening in multi tenant environments, due to app ID structure being different.

Fixes:
- Group details pages always says "no apps assigned"
- Role colours not being accurate (if apps were appearing)
- Already assigned user groups not being filtered out in group assignment modal
- Remove console log

Caught by @MihailHadzhiev2022 :+1:


mike12345567 (approved):

shogunpurple (approved):
LGTM!

PR #7981, opened 2022-09-26
Add apps to a group from within groups interface

mike12345567:
## Description
Feature to allow adding apps from within the groups interface, rather than having to go to apps individually to do this.

Everyone that had tried out the groups experience suggested this as a missing feature, it felt odd that users could be added but apps could not.

This is a two step process, first picking the app, then picking the role within the app.

Also updates plugin badge to "New" rather than "Beta".

## Screenshots
Groups UI:
image

Adding an app (step 1):
image

Adding an app (step 2):
image

App added:
image

### DIFF
--- a/packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
+++ b/packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
@@ -1,0 +1,51 @@
+<script>
+  import { Body, ModalContent, Select } from "@budibase/bbui"
+  import { apps, groups } from "stores/portal"
+  import { roles } from "stores/backend"
+  import RoleSelect from "components/common/RoleSelect.svelte"
+
+  export let group
+
+  $: appOptions = $apps.map(app => ({
+    label: app.name,
+    value: app,
+  }))
+  $: prodAppId = selectedApp ? apps.getProdAppID(selectedApp.appId) : ""
+  $: confirmDisabled =
+    (!selectingRole && !selectedApp) || (selectingRole && !selectedRoleId)
+  let selectedApp, selectedRoleId
+  let selectingRole = false
+
+  async function appSelected() {
+    if (!selectingRole) {
+      selectingRole = true
+      await roles.fetchByAppId(prodAppId)
+      // return false to stop closing modal
+      return false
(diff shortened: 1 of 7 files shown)


joebudi:
Incredible effort to get this in. The UI looks great, and this will provide admins with extra security and less work when onboarding new users.

packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
```diff
@@ -0,0 +1,51 @@
...
+  size="M"
+  title="Add app to group"
+  confirmText="Next"
+  showSecondaryButton
+  secondaryButtonText="Back"
+  secondaryAction={() => (selectingRole = false)}
+  disabled={confirmDisabled}
+>
+  {#if !selectingRole}
+    <Body>Select the app to add to the <i>"{group.name}"</i> group.</Body>
```
shogunpurple, line 42: Since users are the things in the groups, I would update this copy here to be more explicit.

`Select an app the assign permissions for <Group name>`

packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
```diff
@@ -0,0 +1,51 @@
...
+    <Select bind:value={selectedApp} options={appOptions} />
+  {:else}
+    <Body
+      >Pick the role this group will appear as in the <i>"{selectedApp.name}"</i
```
shogunpurple, line 46: `Select the role that all users in the <group name> will have for <app name>.`

shogunpurple (approved):
Code looks great - just some product suggestions about the copy to be more explicit.

packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
mike12345567: image

packages/builder/src/pages/builder/portal/manage/groups/_components/AppAddModal.svelte
mike12345567: image

PR #7983, opened 2022-09-26
Fix cypress accountPortals.spec

Rory-Powell:
## Description
- Fix for failing cypress test
- Add optional chaining to `$licensing.usageMetrics?.dayPasses` when loading the user portal


shogunpurple (approved):
LGTM

Issue #5629, opened 2022-04-28
Redirect after closing modal


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5524

<div type='discussions-op-text'>
... In particular because the 'add specialty' screen opens in a modal, I couldn't find a way to trigger a navigation after closing the modal. 
This would have effectively allowed a refresh of the page, with the default values of the form saved in App State: https://docs.budibase.com/docs/app-state

It might be nice to have an optional redirect URL for the close modal action. I've suggested that here: https://github.com/Budibase/budibase/discussions/1176#discussioncomment-2614757
</div>

Another discussion with the same issue: https://github.com/Budibase/budibase/discussions/5626

Need a way to update state from a modal screen, or as suggested a redirect option after the modal has been closed.

aptkingston:
> Need a way to update state from a modal screen

State updates are already proxied back from screens in modals - so that's usually the best solution if you need to pass back some value from the modal that isn't saved as data. i.e. if you use update state from a screen in a modal, the state is updated on the parent immediately as well.

Can keep this open as a feature request to add an option for redirecting on closing a screen modal though.

melohagan:
That's cool! I think then it's the on change feature that's needed so that updating the state updates the fields. 

The redirect idea was really just another work around for this, but it may have some other use cases, but not a priority. Haven't seen anyone else ask for it specifically.

melohagan:
Actually, you can't simply use the on change, because the field you're trying to update is in a different screen. The same issue has popped up here: 

https://github.com/Budibase/budibase/discussions/6181

adamjdavies:
I'm also interested in this. I want to close a modal and for it to trigger the 'Refresh Data Provider' action. Can't seem to figure out a way to do this at the moment natively in Budibase. 

Thanks, 
Adam

aptkingston:
@adamjdavies if you've edited data inside the modal that is part of the same datasource as the parent page, then you shouldn't need to manually refresh the data provider - it should do it automatically. This won't work if you're using a rest API as your datasource but should work for anything else (including custom queries). 

Maybe you have other reasons for needing a refresh, but if you're editing data in the same datasource then this should work automatically and would be a bug if it isn't.

adamjdavies:
@aptkingston - all of my data is being pulled from an Airtable query, so I need to update the datasource each time a change is made. Therefore, I need some way of triggering 'Refresh Data Provider' action when either the modal is closed or based off of some trigger.

[...]

PR #8016, opened 2022-09-28
Allow screen redirect after modal close

melohagan:
## Description
Some scenarios require screen navigation after modal close. 

For example: updating a form field in the parent window from a modal screen via URL variable. (App state can't be used in this scenario). Use case here: https://github.com/Budibase/budibase/discussions/6181

I couldn't get the navigation to work through the routeStore. I think it's because the router, even after the modal is closed, is still in the context of the "peeked" screen. That's why I'm setting the `location.href` of the parent window. 

Addresses: 
- https://github.com/Budibase/budibase/issues/5629

## Screenshots
**No screen**
Screenshot 2022-09-28 at 14 07 10
modal

**Different screen**
Screenshot 2022-09-28 at 14 09 18
modal

**Reload same screen**
Screenshot 2022-09-28 at 14 10 33
modal

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/CloseScreenModal.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/CloseScreenModal.svelte
@@ -1,16 +1,31 @@
 <script>
-  import { Body } from "@budibase/bbui"
+  import { Label, Body } from "@budibase/bbui"
+  import DrawerBindableInput from "components/common/bindings/DrawerBindableInput.svelte"
+
+  export let parameters
+  export let bindings = []
 </script>
 
+<Body size="S">Navigate To screen, or leave blank.</Body>
+<br />
 <div class="root">
-  <Body size="S">This action doesn't require any additional settings.</Body>
-  <Body size="S">
-    This action won't do anything if there isn't a screen modal open.
-  </Body>
+  <Label small>Screen</Label>
+  <DrawerBindableInput
+    title="Destination URL"
+    placeholder="/screen"
+    value={parameters.url}
+    on:change={value => (parameters.url = value.detail)}
+    {bindings}
+  />
 </div>
 
 <style>
   .root {
+    display: grid;
+    align-items: center;
+    gap: var(--spacing-m);
+    grid-template-columns: auto 1fr;
+    max-width: 400px;
     margin: 0 auto;
   }
 </style>
--- a/packages/client/src/components/overlay/PeekScreenDisplay.svelte
+++ b/packages/client/src/components/overlay/PeekScreenDisplay.svelte
@@ -45,6 +45,9 @@
       },
       [MessageTypes.CLOSE_SCREEN_MODAL]: () => {
         peekStore.actions.hidePeek()
+        if (message.data?.url) {
+          routeStore.actions.navigate(message.data.url)
+        }
       },
       [MessageTypes.INVALIDATE_DATASOURCE]: () => {
         proxyInvalidation(message.data)
--- a/packages/client/src/utils/buttonActions.js
+++ b/packages/client/src/utils/buttonActions.js
@@ -226,28 +226,9 @@ const changeFormStepHandler = async (action, context) => {
 
 const closeScreenModalHandler = action => {
   let { url } = action.parameters
-  if (url) {
-    window.parent.addEventListener("message", event => {
-      const location = event.target.location
-      //remove any trailing slash
-      if (url.charAt(url.length - 1) === "/") {
-        url = url.substring(0, url.length - 1)
-      }
-      //need to reload if hash route has not changed
-      let shouldReload =
-        `#${url.substring(0, url.lastIndexOf("/"))}` ===
-        location.hash?.substring(0, location.hash.lastIndexOf("/"))
-
-      window.parent.location.href = `${location.origin}${location.pathname}#${url}`
-      if (shouldReload) {
-        window.parent.location.reload()
-      }
-    })
-  }
-
   // Emit this as a window event, so parent screens which are iframing us in
   // can close the modal
-  window.parent.postMessage({ type: "close-screen-modal" })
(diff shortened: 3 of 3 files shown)


aptkingston:
Great idea implementing this Mel! Did you try including the URL as data in the `window.parent.postMessage` call for the `close-screen-modal` event? That's handled by the parent, meaning you could simply do `routeStore.actions.navigate(url)` in the `receiveMessage` function in `PeekScreenDisplay.svelte`. You could just add it as a `proxyNavigation` function and call it just after `peekStore.actions.hidePeek()`. 

You also don't need to worry about validation or parsing the URL then, as the routeStore will handle everything and will simply update the URL via pushstate, so checking for changed hash routing params etc isn't necessary.

melohagan:
Hey @aptkingston I gave that a go, however I still found myself needing the page reload:

```js
window.parent.postMessage({ type: "close-screen-modal", url })
```

```js
[MessageTypes.CLOSE_SCREEN_MODAL]: () => {
        peekStore.actions.hidePeek()
        routeStore.actions.navigate(message.data.url)
      },
```

modal

aptkingston:
> Hey @aptkingston I gave that a go, however I still found myself needing the page reload:
> 
> ```js
> window.parent.postMessage({ type: "close-screen-modal", url })
> ```
> 
> ```js
> [MessageTypes.CLOSE_SCREEN_MODAL]: () => {
>         peekStore.actions.hidePeek()
>         routeStore.actions.navigate(message.data.url)
>       },
> ```
> 
> modal [ modal ](https://user-images.githubusercontent.com/101575380/192814551-8c0bad30-10fe-4c22-847f-59b996adb074.gif) [ ](https://user-images.githubusercontent.com/101575380/192814551-8c0bad30-10fe-4c22-847f-59b996adb074.gif)

What's the URL you're actually using there Mel? The navigation action is working because the URL is updating, but it's just serving you the same screen - maybe because of the way the URL params are set up, or because the new URL is bad (maybe doesn't start with a slash or something).

I tried this setup too and it does work perfectly for me. I just used basic examples like `/modal` inside the moda, going to `/after` afterwards, so just wondering if something else is going on with more complicated URLs.

melohagan:
> What's the URL you're actually using there Mel? The navigation action is working because the URL is updating, but it's just serving you the same screen - maybe because of the way the URL params are set up, or because the new URL is bad (maybe doesn't start with a slash or something).
> 
> I tried this setup too and it does work perfectly for me. I just used basic examples like `/modal` inside the modal, going to `/after` afterwards, so just wondering if something else is going on with more complicated URLs.

Screenshot 2022-09-29 at 08 44 28
Screen: `/data-entry/new/row/{{ Product Table.Product.Product ID }}`

Screenshot 2022-09-29 at 08 45 46

Console log:
Screenshot 2022-09-29 at 08 47 33

Here's the app export: 
[Modal Sample-export-1664437722103.txt](https://github.com/Budibase/budibase/files/9672337/Modal.Sample-export-1664437722103.txt)

aptkingston:
> Screenshot 2022-09-29 at 08 44 28 Screen: `/data-entry/new/row/{{ Product Table.Product.Product ID }}`
> 
> Screenshot 2022-09-29 at 08 45 46
> 
> Console log: Screenshot 2022-09-29 at 08 47 33
> 
> Here's the app export: [Modal Sample-export-1664437722103.txt](https://github.com/Budibase/budibase/files/9672337/Modal.Sample-export-1664437722103.txt)

Thanks for the export! Your app actually works perfectly with that change haha. The only thing that might be misleading you is that the text input for Product ID does not update after the navigation happens because we have deliberately removed the feature of resetting field values when the default value changes.

So the navigation does work, and the URL param does update (you can verify by just putting it in some text component) but your field value doesn't update because a change of the "default value" setting no longer updates the live value in a field, if one is set.

melohagan:
> Thanks for the export! Your app actually works perfectly with that change haha. The only thing that might be misleading you is that the text input for Product ID does not update after the navigation happens because we have deliberately removed the feature of resetting field values when the default value changes.
> 
> So the navigation does work, and the URL param does update (you can verify by just putting it in some text component) but your field value doesn't update because a change of the "default value" setting no longer updates the live value in a field, if one is set.

Ah yes this is what I understood to be happening, but the reload is needed to update the default value. For this use case the reload is needed because you can't use the **Update Field Value** action across different screens. 
From my perspective if I am navigating to a page, then the default values should re-trigger because I'm "entering" that screen. 

Maybe as a work around there could be a "Refresh parent screen" checkbox? At least that way I wouldn't need to do the URL substring checks. Or is there another way to handle this scenario?

aptkingston:
Yea this setup is a bit of an edge case, but this problem would also exist even without thinking about modals - like if you had a button on a page to navigate to the same page and just update a URL param, any fields on the page using a URL param as a default value also wouldn't update. So I think we should consider that a separate issue and think of a separate solution for that.

I would prefer to go ahead and merge this in it's simple form, since it does work and solves the original problem of navigating when closing the modal.

In order to address the specific case here, we could either add options to all navigation actions to fully refresh the page (but I would prefer to avoid that as it's basically an escape hatch for other issues), or maybe add some sort of setting to basically allow fields to be reset if their default value changes.

melohagan:
> Yea this setup is a bit of an edge case, but this problem would also exist even without thinking about modals - like if you had a button on a page to navigate to the same page and just update a URL param, any fields on the page using a URL param as a default value also wouldn't update. So I think we should consider that a separate issue and think of a separate solution for that.
> 
> I would prefer to go ahead and merge this in it's simple form, since it does work and solves the original problem of navigating when closing the modal.
> 
> In order to address the specific case here, we could either add options to all navigation actions to fully refresh the page (but I would prefer to avoid that as it's basically an escape hatch for other issues), or maybe add some sort of setting to basically allow fields to be reset if their default value changes.

That makes sense. I've renamed the original issue and pushed up the simple changes.

shogunpurple (approved):
LGTM, nice work Mel

PR #8046, opened 2022-09-29
CLI hosting improvements

mike12345567:
## Description
After some discussion around improvements that could be made to the CLI/plugin development experience I've extended the CLI hosting capabilities significantly.

1. A `--single` flag, which allows initing a new hosting environment utilising the single docker image rather than the multi-image composed setup.
2. A `--watch-plugin-dir` flag, which looks for a `docker-compose.yaml` in the current directory and updates it to give the app-service/single service access to the supplied directory.
3. A `--gen-user` flag, which adds the environment variables to generate an admin user automatically to the compose setup.
4. A new plugin option, `plugins --dev`, which essentially performs `budi hosting --init quick --gen-user --single --watch-plugin-dir ./`. This is a pretty crazy looking command, but basically it generates an admin user, using the single image, it does it in a single command while selecting a free port and it watches the current directory. The user has the choice when running the command to change the directory.

All of the existing commands, start, update etc updated to work for the single image as well.

## Screenshots
Dev command:
image

New hosting help:
image


shogunpurple (approved):
LGTM! This is a really nice dev exp now locally - looking forward to trying it out

PR #8047, opened 2022-09-29
Allow removing custom datasource CRUD operations

mike12345567:
## Description
As the title says - this is a valid configuration that the validator didn't allow - i've confirmed it doesn't break anything, Budibase can handle this fully. Also this means that custom datasource query verbs are correctly validated.


melohagan (approved):
LGTM! Just to double check - the `queryValidator` doesn't require the **fields** property? 

Some methods only have the **type**, `"type": "json"`

mike12345567:
Hey @melohagan Unless the `.required()` is specified for a property in a joi schema then it is not required - it will only validate that field if it is present.

Issue #8023, opened 2022-09-28
Oracle DB connector is missing


mjashanks:
**Hosting**
Self v2.03 and Cloud

**Describe the bug**
OracleDB connector is not listed when creating a new datasource

**To Reproduce**
1. Add a new datasource
2. Look for OracleDB 

**Screenshots**
Self-host:
![image](https://user-images.githubusercontent.com/3524181/192833108-86852f34-01e8-4f2a-8d61-ceb5a8ef0434.png)

Cloud:
![image](https://user-images.githubusercontent.com/3524181/192833166-66f0fb30-7110-4371-a1d0-614def6d4698.png)

PR #8055, opened 2022-09-30
Oracle DB fix

mike12345567:
## Description
Fix for #8023 - moving Oracle integration to ESM import rather than CJS.

Addresses: 
- #8023


shogunpurple (approved):
LGTM!

PR #8058, opened 2022-09-30
Custom datasource icons

mike12345567:
## Description
Adds the ability for custom datasources to have a `icon.svg` uploaded with them if desired. Currently SVG is the only supported image style as we need to be able to scale it up and down as required.

The backend has been updated to automatically upload the `icon.svg` with the correct content-type to MinIO and I've updated the frontend with a mechanism to display these custom SVGs. I've also added a new component for displaying custom SVGs as they may have classes/CSS applied to them which ends up applying to other SVGs, such as common id tags like `clip0` which are exported by most common SVG rendering applications.

## Screenshots
image


shogunpurple (approved):
LGTM!

PR #8060, opened 2022-09-30
Fetch environment info early in the client app initialisation process

aptkingston:
## Description
This PR fixes a couple of issues in cloud. Currently there is a race condition which means that plugin bundle URLs and the plugin websocket are both not correctly used if the environment information has not been properly loaded, which is usually has not.

This PR fetches the env info early in the client initialisation, allowing those depdendent features to be properly aware of their environment.


shogunpurple (approved):
LGTM! Pesky `await`

PR #8062, opened 2022-09-30
OracleDB development machine fix

mike12345567:
## Description
Quick fix for development machines, when running Budibase development stack on systems that are not oracle compatible it would fail to start due to the lack of dependency.


shogunpurple (approved):
LGTM

PR #8064, opened 2022-09-30
Send more logs to stdout

jonnymccullagh:
## Description
The aim of these changes is to send more logs to stdout/stderr rather than log files within the container. 
- Nginx was fairly easy by adding the www-data user to the tty group and amending the nginx conf files to log to /dev/stdout 
- For redis, clouseau and minio I added `> /dev/stdout 2>&1 &` in `runner.sh` but somewhat difficult to test
- PM2 should take the parameters `-o` and `-e` but I don't see anything from that when tailing the docker logs for the container.


shogunpurple (approved):
Nice work @jonnymccullagh - will be good to get some more logs in the single image.

Issue #8053, opened 2022-09-30
Lab day plugin bugs + improvements


aptkingston:
Opening this for visibility as I work through some of these changes. Feel free to add any aditional feedback that anyone else has!

Bugs:
- [x] Custom components don't work with generated bindings for things like field settings or action inputs
- [x] `fetchRow` in the internal API is currently broken (frontend implementation, the actual endpoint is fine)
- [x] `table` setting type is broken

Improvements:
- [ ] BBUI could be bundled better so that individual components could be imported, e.g. `Import Button from "@budibase/bbui/button"`, without needing to bundle the entirety of BBUI
- [x] Use the plugin friendly name (rather than the component name) for names of new plugin components, e.g. `New Custom Component` instead of `New custom-component`
- [x] Use a separate rollup config when developing plugings (when watch mode is active) which adds inline sourcemaps and disables terser, for easier debugging
- [x] Show `Plugin X reloaded` notification
- [ ] Document use of `"type": "event"` + usage with actions / bindings
- [x] Notify builder from client when plugins are hot reloaded, so the builder can fetch component definitions again (cheap hack to basically get hot reloading of component settings)
- [x] Ensure only one instance of the websocket can be constructed

PR here: https://github.com/Budibase/budibase/pull/8054

melohagan:
Something I came across today with the latest Budibase CLI and master branch was the need to include all four CRUD methods for the datasource plugin to import.
Ideally I shouldn't need to provide all of these.

PR #8066, opened 2022-09-30
Plugin fixes and improvements

aptkingston:
Same PR but destined for master. Accidentally merged the other into develop which is being reverted.

## Description
Collection of fixes and improvements noted in https://github.com/Budibase/budibase/issues/8053.

### Friendly names for component plugins:
image

### Notification when hot reloading plugins:
image

### Use inline sourcemaps when use watch mode to build your plugins:
image
image
(note - this change is in the skeleton repo, but included here for reference)

### Custom component settings are now hot reloaded:

https://user-images.githubusercontent.com/9075550/193256616-6c1b0163-8504-4538-8de8-4cb4646ff781.mp4

### Ensure only one instance of the websocket can exist
Before - every instantiation of the client app would create a new socket, causing a massive stack up of requests:

https://user-images.githubusercontent.com/9075550/193257127-b1b9b37f-9de2-41e6-bb1b-0b7fa46d283b.mp4

After - the same socket is reused, so no more requests are needed:

https://user-images.githubusercontent.com/9075550/193257170-4bcb632b-8837-4d8c-90c0-981fee0a483f.mp4

### DIFF
--- a/packages/builder/src/builderStore/componentUtils.js
+++ b/packages/builder/src/builderStore/componentUtils.js
@@ -143,7 +143,10 @@ export const getComponentSettings = componentType => {
   }
 
   // Ensure whole component name is used
-  if (!componentType.startsWith("@budibase")) {
+  if (
+    !componentType.startsWith("plugin/") &&
+    !componentType.startsWith("@budibase")
+  ) {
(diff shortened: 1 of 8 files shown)


shogunpurple (approved):
LGTM!

PR #8067, opened 2022-09-30
CLI plugins dev command logging improvement

mike12345567:
## Description
After @melohagan tried out some of the new CLI commands - found that the logging output was a bit lacking/confusing. All this does is try to compact the important logs and make them more obvious/lay them out in a way in which they cannot be ignored.

Admin credentials and where to go were easy to miss/get lost in the spam - relatively quick fix.

## Screenshots
Appears as such now:
image


shogunpurple (approved):
LGTM - this output is really helpful

PR #8080, opened 2022-10-01
/application endpoints API tests

shogunpurple:
## Description
Further fleshing out the QA Core suite, so that it has the ability to create, publish and check if apps can render properly.

Also added some documentation on getting started and configured `@budibase/backend-core` to work with Typescript imports so backend core utilities can be used inside the jest tests.

Addresses: 
- ...more if required

## Screenshots


qa-core/src/module.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module "@budibase/backend-core/tenancy"
```
Rory-Powell, line 1: Are these needed? I think with the file mapping you have put in place they shouldn't be e.g. worker doesn't have thus. Introducing it might further promote the legacy approach of reaching directly into backend-core files (which we don't get proper type support for) rather than going through the index file.

shogunpurple: I'll take another look at the worker config and see if there's something I missed - I had to add this to get my `backend-core` imports working inside the tests

shogunpurple: Yep - that works, after I added the wildcard mapping from worker and imported from the top level `index`, `module.d.ts` isn't relevant anymore.

Issue #8030, opened 2022-09-29
Automation: Update row to make a boolean Value "FALSE" does not work


shumpeiKishi:
**Hosting**
- Cloud

**App Ver**
- 2.0.5

**Describe the bug**
When Automation feature Update Row is applied to a boolean column (True/False) in a row on Budibase DB, the boolean column cannot be updated as False. It can be updated to True though. 

**To Reproduce**
1. Create a table in Budibase DB that has a boolean value column.
2. Insert a row in the table setting the boolean value to True.
3. Create an automation to update to make the boolean value False. (Cron Trigger is convenient.)
4. Run a test automation.
5. Go back to Data and see the boolean value is not updated to False.

**Expected behavior**
Boolean values can be updated to False as it can be done to True.

PR #8088, opened 2022-10-03
Automation: Update row to make a boolean Value "FALSE" does not work #8030

shumpeiKishi:
## Description
Fix for "Automation: Update row to make a boolean Value "FALSE" does not work https://github.com/Budibase/budibase/issues/8030" by changing the data validation to accept FALSE but filter null or empty strings "".

Addresses: 
Automation: Update row to make a boolean Value "FALSE" does not work https://github.com/Budibase/budibase/issues/8030

### DIFF
--- a/packages/server/src/automations/steps/updateRow.js
+++ b/packages/server/src/automations/steps/updateRow.js
@@ -66,9 +66,9 @@ exports.run = async function ({ inputs, appId, emitter }) {
   }
   const tableId = inputs.row.tableId
 
-  // clear any falsy properties so that they aren't updated
+  // clear any null or empty string properties so that they aren't updated
   for (let propKey of Object.keys(inputs.row)) {
-    if (!inputs.row[propKey] || inputs.row[propKey] === "") {
+    if (inputs.row[propKey] === null || inputs.row[propKey] === "") {
       delete inputs.row[propKey]
     }
   }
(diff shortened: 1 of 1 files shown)


shumpeiKishi:

melohagan (commented):
Brilliant find - thank you! Can you please add an `undefined` check as well.

shumpeiKishi:
@melohagan Thanks for your feedback. I'll work on it.

packages/server/src/automations/steps/updateRow.js
```diff
@@ -66,9 +66,9 @@ exports.run = async function ({ inputs, appId, emitter }) {
   }
   const tableId = inputs.row.tableId
 
-  // clear any falsy properties so that they aren't updated
+  // clear any null or empty string properties so that they aren't updated
   for (let propKey of Object.keys(inputs.row)) {
-    if (!inputs.row[propKey] || inputs.row[propKey] === "") {
+    if (inputs.row[propKey] === null || inputs.row[propKey] === "") {
```
Rory-Powell, line 71: We can use the loose equality operator `==` instead of `===` here to check for both null and undefined at the same time as @melohagan has suggested. i.e:
```js
inputs.row[propKey] == null || inputs.row[propKey] === ""
```

melohagan: Nice! TIL.

melohagan (approved):
LGTM! Thanks for the fix! 👍

packages/server/src/automations/steps/updateRow.js
shumpeiKishi: Thanks, @Rory-Powell for the advice!

Issue #6880, opened 2022-07-25
Public API 404 - updating/deleting table rows


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6832

<div type='discussions-op-text'>

<sup>Originally posted by **smitsialis** July 20, 2022</sup>
I'm trying to use the OpenAPI to manage the data in my tables.  I'm using Postman to test this and having issues with the delete row and update row.  I am able to create new rows, search for rows and retrieve a row using the id.
Example:
`GET {{baseUrl}}/tables/datasource_plus_8bf0184b1acd4f6a8940ae60eebbc1a4__Contracts/rows/7 `
 Works fine:
```
 {
    "data": {
        "id": 7,
        "Name": "contract_YVES5",
        "_id": "%5B7%5D",
        "tableId": "datasource_plus_8bf0184b1acd4f6a8940ae60eebbc1a4__Contracts"
    }
}
```
I am trying to edit this using:
```
PUT {{baseUrl}}/tables/datasource_plus_8bf0184b1acd4f6a8940ae60eebbc1a4__Contracts/rows/7 
{
  "_id": "%5B6%5D",
  "type": "row",
  "tableId": "datasource_plus_8bf0184b1acd4f6a8940ae60eebbc1a4__Contracts",
  "Name": "New_data"
}
```
but get back:
```
{
    "message": "missing",
    "status": 404
}
```

Same error when I try to delete:
`DELETE {{baseUrl}}/tables/datasource_plus_8bf0184b1acd4f6a8940ae60eebbc1a4__Contracts/rows/7`

The headers are the same for all 3 methods and not the problem.  The table has a single column 'Name'.  What I'm I missing?</div>

----
From the logs it appears that it's trying and failing to find the row in the couchdb-service:
```
budibase-couchdb-service-1 | [notice] 2022-07-25T10:20:21.764450Z couchdb@127.0.0.1 <0.13237.0> 1f94b118f6 couchdb-service:5984 172.18.0.8 yjlaYbkEQdTtKIqVwKVUjOizzv3aLOyN GET /app_dev_513fe5ef178b48c787b81b57e4cf13e9/1? 404 ok 5
bbapps | ERROR [1658744421767] (not_found/55 on ddbeba4e9677):
bbapps |     error: "not_found"
bbapps |     reason: "missing"
bbapps |     status: 404
bbapps |     message: "missing"
[...]

PR #8108, opened 2022-10-03
Minor fix for public API + SQL table rows

mike12345567:
## Description
Fix for #6880 - allow SQL table rows to be updated/deleted by their _id field from the public API.


shogunpurple (approved):
LGTM

PR #8114, opened 2022-10-04
Update exec.js

rosnerdev:
## Description
I had an issue where the datasource plugin wouldn't work because of node engine incompatibility, and after a quick search on google I found that --ignore-engines is the option to use for ignoring this kind of error and it worked! So I want to make this a built-in feature so there's no need for me to manually do this.

## Screenshots
screenshot


rosnerdev:

rosnerdev:
recheck

mike12345567 (approved):

PR #8121, opened 2022-10-04
Fallback to push branch name for pro install script

melohagan:
## Description
If the branch name is missing, fallback to the Github ref name.

### DIFF
--- a/scripts/pro/install.sh
+++ b/scripts/pro/install.sh
@@ -19,6 +19,7 @@ if [[ -d "budibase-pro" ]]; then
   cd budibase-pro
 
   if [[ -z "${BRANCH}" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/pro/install.sh
```diff
@@ -18,6 +18,10 @@ git clone https://$PERSONAL_ACCESS_TOKEN@github.com/Budibase/budibase-pro.git
 if [[ -d "budibase-pro" ]]; then
   cd budibase-pro
 
+  if [[ -z "${BRANCH}" ]]; then
+    export BRANCH=$GITHUB_REF_NAME
```
Rory-Powell, line 22:
```suggestion
    echo Using GITHUB_REF_NAME: $GITHUB_REF_NAME
    export BRANCH=$GITHUB_REF_NAME
```

Issue #8051, opened 2022-09-29
Performing a search on a row with the REST API, doesn't work if the text field only contains numbers


geertvanheusden:
**Hosting**
- Self
  - Method: docker single image
  - Budibase Version: 1.3.15

**Describe the bug**
If you use the internal database and create a table with a column of the type text, it's not possible to search on it using the REST API with the equal method if that text field only contains numbers. Adding one text character to it, makes it searchable again.

**To Reproduce**
1. Create an internal database in your app
2. Create a table with a text column
3. Insert a number (e.g. 123) in the text column
4. Publish the app
5. Perform an REST API row search query with the following body: `{"query":{"equal":{"field":"123"}}}`

**Expected behavior**
The query should return the row. 
Changing the field value to "123a" will make it searchable via the API.

geertvanheusden:
By using the "string" query (`{"query":{"string":{"field":"123"}}}`) instead of "equal", I do get result. 
Although the query is different (equal vs starts with), it could work for me as a workaround.

PR #8123, opened 2022-10-04
Public api equals number text type fix

melohagan:
## Description
Support text type number queries

Addresses: 
- https://github.com/Budibase/budibase/issues/8051


mike12345567 (approved):
LGTM - good test!

PR #8124, opened 2022-10-04
Plugin url updates

aptkingston:
## Description
Corresponding pro PR: https://github.com/Budibase/budibase-pro/pull/69

This PR updates how the URLs for fetching plugins are generated. The URLs are now generated at runtime when requested by the client. This ensures we do not need to migrate any existing plugin or app docs. As the URLs are now fully generated server-side, including accounting for the CDN when running in the cloud, all frontend logic for determining URLs has been removed.

A utility function has been handed to handle the enrichment of these JS URLs as it is required in multiple places.

There might be other locations where enrichment is required which I've missed, so please point any out.

### DIFF
--- a/packages/server/src/utilities/plugins.js
+++ b/packages/server/src/utilities/plugins.js
@@ -1,0 +1,20 @@
+import env from "../environment"
+import { plugins as ProPlugins } from "@budibase/pro"
+
+export const enrichPluginURLs = plugins => {
+  if (!plugins || !plugins.length) {
+    return []
+  }
+  return plugins.map(plugin => {
+    const cloud = !env.SELF_HOSTED
+    // In self host we need to prefix the path, as "plugins" is not part of the
+    // bucket path. In cloud, "plugins" is already part of the bucket path.
+    let jsUrl = cloud ? "https://cdn.budi.live/" : "/plugins/"
+    jsUrl += ProPlugins.getBucketPath(plugin.name)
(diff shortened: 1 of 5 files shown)


packages/server/src/utilities/plugins.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const enrichPluginURLs = plugins => {
+  if (!plugins || !plugins.length) {
+    return []
+  }
+  return plugins.map(plugin => {
+    const cloud = !env.SELF_HOSTED
+    // In self host we need to prefix the path, as "plugins" is not part of the
+    // bucket path. In cloud, "plugins" is already part of the bucket path.
+    let jsUrl = cloud ? "https://cdn.budi.live/" : "/plugins/"
```
mike12345567, line 12: I believe there is a `PLUGIN_BUCKET_NAME` environment variable in backend-core - the `ObjectStoreBuckets` enumeration exported from `import { objectStore } from "@budibase/backend-core` automatically utilises this environment variable if it is specified. We probably need to account for this - in the cloud the bucket name doesn't matter as cloudfront makes this decision but in self host we should probably respect the env var.

mike12345567 (approved):
LGTM - just one minor comment!

packages/server/src/utilities/plugins.js
aptkingston: Thanks for pointing this out - didn't know it was configurable!

PR #8125, opened 2022-10-04
Fixing options type import with CSV

mike12345567:
## Description
Fixing a regression of CSV table creation - normally if a CSV was used to create a table with an options column all of the options would be filled in but this had been broken.

This was an issue noted by @aptkingston who utilised this feature regularly.


aptkingston (approved):
LGTM! My sales CSV will be very happy to see this fixed

PR #8133, opened 2022-10-05
Fix issues with looping in automations

PClmnt:
## Description
Fixes two issues with loop blocks in automations

- Issue where the loop's current item was't being parsed correctly when passed into a query rows filter
- When the loop block was inserted in a particular position, the bindings in the automation backend weren't in the correct position, therefore making thje bindings incorrect (i.e `steps.3.value` would have needed to be `steps.4.value` on the front-end to work correctly

Addresses: 
#7906


Rory-Powell (approved):
Unfamiliar with the looping logic, but confident in your analysis / fix from the description 👍

Issue #8131, opened 2022-10-05
Data Provider User Created-By


morphace:
**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.0.12

**Describe the bug**
I'm trying to limit records in a Data Provider by user that has created the record, as described here: https://docs.budibase.com/docs/fetching-data
However the Data drop down of the Data Provider only contains Tables, not the relationships as described.

I'm using BudibaseDB as the database.

melohagan:
I'm also seeing this as early as **v1.0.15**.

morphace:
Ah, intersting @melohagan - thanks for the feedback. Maybe there's an alternative approach to implement this?

morphace:
I also tried to implement a filter on the Data Provider that would filter the Created By column by the Current User, but the Created By column is not provided as filter column.

PR #8137, opened 2022-10-05
Removed filter excluding links from user bindings.

deanhannigan:
## Description
Removed filtering of user relationships applied to user bindings. This should not have been applied globally.

Addresses: 
- https://github.com/Budibase/budibase/issues/8131


PClmnt (approved):
LGTM, nice one!

Issue #5994, opened 2022-05-20
there was an error loading your app 


robert0825:
budibase version:1.0.158
client :iphone IOS13.5  or under  IOS13.5
when i try to open in IPHONE, there was an error message like this:
"The Budibase client library could not be loaded,Try republishing your app"
![budibase](https://user-images.githubusercontent.com/1121187/169446055-772ff02f-e59f-4975-86ee-aff33e988af6.jpg)

i have tried a lot of device  (android ,iphone etc..) , iphone under IOS13.5 will display this error.

shogunpurple:
Does it get the same error on desktop?

If so, could you post the dev console and show the error? You can open the dev console with:

Ctrl + Shift + J on windows
Ctrl + Option + J on Mac

robert0825:
only get error in mobile device( iphone with IOS13.5 or under IOS13.5).

chooimail:
Same thing here.

I happen to try a publicly published app using an old iPad mini with iOS 12.5...
Same page loaded.
- Didn't matter if it was using Safari or Chrome browser

Tried updating to the latest budibase version 1.0.184
No changes either.
Could be an iOS version problem.

Tried my own same page/app on iPhone device.
- iOS 15.4 does not have this problem

geertvanheusden:
<img width="1574" alt="CleanShot 2022-10-05 at 22 34 07@2x" src="https://user-images.githubusercontent.com/1923423/194158176-27c224fd-d26e-4cdb-950d-8a7e1021a111.png">

Just experienced the same issue in Chrome. I think I got into this state by doing the following:

* Delete an app that was published (not sure why I was able to do this without unpublishing the app first)
* Create an app with the same name
* Publish the app and open it

Unpublishing and deleting the app followed by creating the same app with the same name resolved this issue for me.

miches87:
I just got this feedback from some of your clients as well. 
Is there any better fix know except from rebuilding the whole app as @geertvanheusden did?

PR #8141, opened 2022-10-05
Some iOS 13 Support

melohagan:
## Description
Users were having issues using Budibase in iOS12 and 13. 

Basically this was caused by JavaScript ES2020 support issues below iOS15. **Long story short - upgrade to iOS15.**

I played around with reducing the `es2020` in the tsconfig files, but it was causing issues so I abandoned that, but it means I wasn't able to get the builder working for iOS13 as I came across this same issue: https://stackoverflow.com/questions/68484843/angular-error-promisereactionjob-in-ios

**In the end, the only thing I could do was to allow public apps to be viewed within iOS13. 
iOS12 is a no go.** 

Addresses: 
- https://github.com/Budibase/budibase/issues/5994

## Screenshots
**Before**
Screenshot 2022-10-05 at 15 51 40

**Public app after:**
Screenshot 2022-10-05 at 15 57 28

### DIFF
--- a/packages/client/src/components/app/charts/ApexOptionsBuilder.js
+++ b/packages/client/src/components/app/charts/ApexOptionsBuilder.js
@@ -1,37 +1,39 @@
 export class ApexOptionsBuilder {
-  formatters = {
-    ["Default"]: val => (isNaN(val) ? val : Math.round(val * 100) / 100),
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
Nice @melohagan ! This is a pretty big deal.

Issue #4379, opened 2022-02-07
Barcode/QR Scanner Component


shogunpurple:
Spec: https://budibase.atlassian.net/wiki/spaces/PROD/pages/12615691/QR+Barcode+Scanner

When I'm creating new rows in budibase, I want to be able to use the camera on my device to scan a QR code or barcode, and insert the encoded information into a string type field in the form.

**COS**
- Add barcode scanner component to any input requiring a QR code to be scanned 
- Add barcode/QR type to form inputs section

Designs:
https://www.figma.com/file/NNLgDGmQxM96b5i0OKwkax/QR-%2F-barcode-scanner?node-id=0%3A1

Grayda:
Don't know if you're planning to use a webcam for this (like a QR code scanner), or if it's just a field that would validate barcodes fed in via a USB barcode scanner or something, but if it's webcam based, could that component be split up to also handle taking photos and uploading them, like an Attachment field?

I'm building a video game collection tracker, and it'd be nice to be able to take a photo and upload it in-app, without having to take a photo via my phone or webcam, save it to my PC, then attach it.

mike12345567:
Thoughts on this mentioned in: #4671.

nev-21:
Great, it is a necessary component!

artisleilands:
As you are involving the phone camera already (right?), could you also consider the possibility to add a photo as an attachment? I.e. to enable taking photos directly from the BB app.

kasgithub8:
Are there any updates on this?

PR #8153, opened 2022-10-06
Feature/qr barcode reader

deanhannigan:
## Description
Adding a Code Scanner form component for scanning QR codes and a variety of barcodes. 

 - `Barcode/QR` has been added to datasources as a new column type. 
 - The `Barcode/QR Scanner` component is available under the 'Form' sections in the component list.
 - Any field configured as a `Barcode/QR` will display as a `Barcode/QR Scanner` in the builder and be persisted as a string.
 - Datasource fields currently configured as `Text` or `Long Form Text` or `Options` can safely switch to `Barcode/QR` if required

The full list of supported barcodes is available here : https://github.com/mebjas/html5-qrcode

Addresses: 
- https://github.com/Budibase/budibase/issues/4379

## Screenshots

Scanning a QR Code on mobile.

QR_reader_AdobeExpress

Scanning a Code123 Barcode on mobile.

barcode_clip_AdobeExpress

You can also allow manual entry on the field. This feature can be toggled in the builder

Screen_Recording_2022-10-06_at_10_11_54_AdobeExpress

'Barcode/QR' column type db configuration. You can safely switch between all text types.

Screenshot 2022-10-07 at 10 39 49

If the user hasn't granted camera permissions on their device.

Screenshot 2022-10-06 at 09 32 49


Rory-Powell:
Nice one dean! One suggestion around the name of the column type, `code` might be confused with a code field as in entering html, js etc. Maybe could use something more distinctive like `Scanner Code` or `Barcode`

shogunpurple (approved):
This is a really nice little component, love the UX. Great job 🚀

joebudi:
Nice work @deanhannigan . The UX is great and the write-up was brilliant - loved the videos. I'm with Rory on this. Scanner code is fine, but I'd prefer 'Barcode / QR code' - it is probably the most informative... but a tad long.

shogunpurple:
@deanhannigan We should be able to use slashes in the type names anyway for the friendly names - we could go with `QR/Barcode` even for a bit more brevity. Having that in there makes things much more explicit

deanhannigan:
@joebudi @shogunpurple I've updated the column type to display as `Barcode/QR` and the component itself is now labelled as `Barcode/QR Scanner` in the component list.

When binding, the field type was previously `Scannedcode` but will now display as the following to fit in with the column name better:

Screenshot 2022-10-07 at 10 43 10

Let me know what you think.

shogunpurple:
@deanhannigan thanks for making the updates. Just one change - if it's not possible to have the `Barcode/QR` in the type list, I'd just make that one `Barcode`. Right now it's not clear that it's `Barcode/QR` in the type list, it just looks like one word.

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
+<script>
+  import { ModalContent, Modal, Icon, ActionButton } from "@budibase/bbui"
+  import { Input, Button, StatusLight } from "@budibase/bbui"
+  import { Html5Qrcode } from "html5-qrcode"
+
+  export let value
+  export let disabled = false
+  export let allowManualEntry = false
+  export let scanButtonText = "Scan Code"
```
aptkingston, line 9: NAB: Change casing to "Scan code" for consistency with other buttons

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
...
+    >
+      {scanButtonText}
+    </ActionButton>
+  {/if}
+</div>
+
+<div class="modal-wrap">
+  <Modal bind:this={camModal} on:hide={hideReaderModal}>
+    <ModalContent
+      title={scanButtonText}
```
aptkingston, line 134: `scanButtonText` will default to "Scan code" if the value is undefined, but if the user has entered a value in the setting field and then cleared it, the text will always be undefined as it does not account for empty strings. 

2 ways to solve this would be:
`$: scanButtonText = scanButtonText || "Scan code"` 
or every time you use `scanButtonText`, do
`scanButtonText || "Scan code`

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
...
+
+      <div slot="footer">
+        <div class="footer-buttons">
+          {#if allowManualEntry}
+            <Button
+              group
+              secondary
+              newStyles
+              on:click={() => {
+                manualMode = !manualMode
```
aptkingston, line 170: NAB: since this is just a toggle, if you've enabled manual mode already and then press this again, it will unset it which feels a bit strange. I reckon we should probably just hide the button if we're already in manual mode, so `{#if allowManualEntry && !manualMode}`, and then the onclick handler here can just set it explicitly to `true`.

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
...
+                camModal.hide()
+              }}
+            >
+              Enter Manually
```
aptkingston, line 174: NAB: change casing to "Enter manually" to be consistent with other buttons.

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
...
+        </div>
+      </div>
+    </ModalContent>
+  </Modal>
+</div>
+
+<style>
+  #reader :global(video) {
+    border-radius: 4px;
+    border: var(--border-light-2);
```
aptkingston, line 196: This is a hardcoded light border so it looks off in dark themes. Let's use
```css
border: 2px solid var(--spectrum-global-color-gray-300);
```
instead so that it looks better in any theme.

packages/client/src/components/app/CodeScanner.svelte
```diff
@@ -0,0 +1,235 @@
...
+      var(--spectrum-global-dimension-size-100)
+    );
+    margin-bottom: var(--spacing-m);
+  }
+  .camera-placeholder {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border-radius: 4px;
+    border: var(--border-light-2);
```
aptkingston, line 222: This is a hardcoded light border so it looks off in dark themes. Let's use
```css
border: 2px solid var(--spectrum-global-color-gray-300);
```
instead so that it looks better in any theme.

packages/client/manifest.json
```diff
@@ -3157,6 +3157,57 @@
...
+        "required": true
+      },
+      {
+        "type": "text",
+        "label": "Label",
+        "key": "label"
+      },
+      {
+        "type": "text",
+        "label": "Button",
```
aptkingston, line 3184: NAB: let's rename this to "Button text" to make it more obvious that's what it is

packages/client/manifest.json
```diff
@@ -3157,6 +3157,57 @@
       }
     ]
   },
+  "codescanner": {
+    "name": "Barcode/QR Scanner",
+    "icon": "Camera",
+    "styles": [
+      "size"
+    ],
+    "draggable": true,
```
aptkingston, line 3166: NAB: setting draggable to true here is redundant as it's only a flag used to disable DND if set to false. Setting illegal children is also redundant since this component doesn't take child components anyway.

aptkingston (commented):
Great work @deanhannigan! It works perfectly and well done for getting it to work nicely with other text type fields too :ok_hand:.

Left a few comments but it's mostly minor stuff.

One other suggestion is that I'd like to move the `CodeScanner` component inside the `forms` directory too (beside the field), since it is not exported to use as a standalone component, and the rest of the component inside the `app` directory are for usage inside apps rather than just parts of other components.

One thing which is a bit unique with this implemenation is that none of it is inside BBUI, where all our other fields are. That's fine, but it just means we can't use the code scanner field inside the builder (which is probably grand since the builder isn't mobile friendly so you'd be entering it manually anyway).

Issue #8042, opened 2022-09-29
S3 - CSP problem configuring S3 Component in latest version 2.0.3


reichert:
**Hosting**
- Self
  - Method: docker-compose
  - Budibase Version: 2.0.3

**Describe the bug**
I'm receving a CSP - Content Security Policy problems when using south america region of S3.
![image](https://user-images.githubusercontent.com/1287032/193044986-b88cd653-07d6-45e1-babc-1269fcd3f895.png)
- After reading issue #6215 I checked and *.s3.sa-east-1.amazonaws.com is included in NGINX.
- I'm using postgres database and a noted that:
  - In the FIELD property of the S3 component: if it is empty the component does not work. Also no options are shown of available fields;
  - If I type manually a field (existent or not) the component is shown correctly in the preview app but the error above occurs.

**To Reproduce**
1. Create a bucket in S3 region and respective iam credentials with write permissions
2. Configure a datasource of type S3 in budibase
3. Inside a form put a S3 component and configure it accordinly
4. Run the app and drop a image inside the componente. The error above will be raised

**Screenshots**
I'll paste the way I configured our app, just in case.
![image](https://user-images.githubusercontent.com/1287032/193047048-e93778ed-f9ee-4380-a4bf-1a50bdff6772.png)
![image](https://user-images.githubusercontent.com/1287032/193048337-9fe53ede-fa98-4612-b27b-a3170cf18665.png)

**Desktop (please complete the following information):**
 - OS: linux
 - Browser brave and chrome

reichert:
Thanks @melohagan for your help!
Yes, my bucket is public readable and my cors policy exists as well. I changed a little to me more like yours but no effect.
My current cors policy is:
```
[
    {
        "AllowedHeaders": [
            "*"
        ],
        "AllowedMethods": [
            "GET",
            "PUT",
            "POST",
            "DELETE",
            "HEAD",
            "OPTIONS"
        ],
        "AllowedOrigins": [
            "*"
        ],
        "ExposeHeaders": []
    }
]
```
I'm still receiving the same CSP error:
![image](https://user-images.githubusercontent.com/1287032/193825202-831016f2-41ff-43a2-a6dd-3f35ed9c8eb5.png)

melohagan:
I think potentially the issue could be that we're using `https://*.s3...` 
Maybe we have to add the root domain as well without the asterisk, i.e. `https://s3...`

[...]

PR #8158, opened 2022-10-06
Update CSP for S3

PClmnt:
## Description
As per this issue: #8042 it seems like our CSP was blocking access to this specific S3 URL 

I've updated the NGINX conf to allow this through: `https://s3.sa-east-1.amazonaws.com`


hosting/nginx.prod.conf.hbs
```diff
@@ -55,7 +55,7 @@ http {
...
...rcom.io https://nexus-websocket-b.intercom.io  https://uploads.intercomcdn.com  https://uploads.intercomusercontent.com https://*.s3.amazonaws.com https://*.s3.us-east-2.amazonaws.com https://*.s3.us-east-1.amazonaws.com https://*.s3.us-west-1.amazonaws.com https://*.s3.us-west-2.amazonaws.com https://*.s3.af-south-1.amazonaws.com https://*.s3.ap-east-1.amazonaws.com https://*.s3.ap-southeast-3.amazonaws.com https://*.s3.ap-south-1.amazonaws.com https://*.s3.ap-northeast-3.amazonaws.com https://*.s3.ap-northeast-2.amazonaws.com https://*.s3.ap-southeast-1.amazonaws.com https://*.s3.ap-southeast-2.amazonaws.com https://*.s3.ap-northeast-1.amazonaws.com https://*.s3.ca-central-1.amazonaws.com https://*.s3.cn-north-1.amazonaws.com https://*.s3.cn-northwest-1.amazonaws.com https://*.s3.eu-central-1.amazonaws.com https://*.s3.eu-west-1.amazonaws.com https://*.s3.eu-west-2.amazonaws.com https://*.s3.eu-south-1.amazonaws.com https://*.s3.eu-west-3.amazonaws.com https://*.s3.eu-north-1.amazonaws.com https://*.s3.sa-east-1.amazonaws.com https://s3.sa-east-1.amazonaws.com https://*.s3.me-south-1.amazonaws.com https://*.s3.us-gov-east-1.amazonaws.com https://*.s3.us-gov-west-1.amazonaws.com";
```
Rory-Powell, line 58: would it be possible to just add `https://*.s3.*.amazonaws.com`
we already wildcard `https://*.s3.amazonaws.com` so it doesn't seem like a big security stretch. Otherwise we're going to keep having to add new subdomains for every s3 region 

@shogunpurple what's your thoughts here?

PClmnt: Good point, happy to change to that but thought it was better to be specific, but you are right, it could end up being a constant additions down the line.

shogunpurple: Yep let's wildcard it but make sure we test it in preprod first, as I'm pretty sure we had some issues with wildcarding in the past - it's amazons domain so the risk is relatively low. If I'm honest I've been thinking a lot recently about stripping CSP back in quite a big way. It makes life hard for some use cases - such as custom components that make API calls. I'm not actually sure how much it adds from a security perspective and we are always updating it

Rory-Powell: It's quite difficult to work with you are right, I think what we do get is XSS protection from it so it's quite powerful. 
It's probably something we should make more easily configurable via env vars for self host, or have a process for the cloud where a domain can be requested to be added. 

Agree that we should test the wildcard first and foremost.

JayP718: https://github.com/Budibase/budibase/issues/9252

Hi, 

I just came across this thread after raising the issue above. Not sure if there's a plan to merge this but just a +1 for this one.

"such as custom components that make API calls. " experiencing this now

JayP718:
not stale

hosting/nginx.prod.conf.hbs
shogunpurple: @JayP718 it defeats the purpose of the CSP if we allow calls to anywhere directly from our cloud. 

The whole idea is to prevent calls to URLs we do not trust. I'm going to close the other issue as it's not a bug but a security measure.

PClmnt:
Finally updated this to use `https://*.s3.*.amazonaws.com` instead of all the separate regions.

PClmnt:
So finally got around to updating the CSP, we now have two wildcards

`https://*.s3.*.amazonaws.com`
`https://s3.*.amazonaws.com`

Hopefully that should cover all usecases for S3

Rory-Powell (approved):
LGTM

Rory-Powell (commented):
Should target `develop`?

PClmnt:
> Should target `develop`?

Now targeting develop

Issue #8079, opened 2022-09-30
Minio fails to start on ARM64


MarioGK:
**Hosting**

- Self
  - Method: docker-compose
  - Budibase Version: latest
  - Image: https://hub.docker.com/r/budibase/budibase

**Describe the bug**
On arm64 minio fails to start
```
budibase_1  | /runner.sh: line 70: /minio/minio: cannot execute binary file: Exec format error
```

**To Reproduce**
Use the provided docker-compose.yaml file and run docker-compose up on an arm64 machine

melohagan:
Hi @MarioGK 

Can you try adding `platform: linux/amd64` under the **minio-service** in the docker-compose file.

MarioGK:
Hey @melohagan Thanks for trying to help, i tried to add it to the service but i am getting an error

docker-compose.yaml
```yaml
version: "3.6"

services:
  budibase:
    platform: linux/amd64
    restart: unless-stopped
    image: budibase/budibase:v2.0.3
    ports:
      - "9101:80"
    volumes:
      - ./data:/data
    environment:
      JWT_SECRET: 123
      MINIO_ACCESS_KEY: 123
      MINIO_ROOT_USER: 123
      MINIO_SECRET_KEY: 123
      MINIO_ROOT_PASSWORD: 123
      REDIS_PASSWORD: 123
      COUCHDB_USER: 123
      COUCHDB_PASSWORD: 13
      INTERNAL_API_KEY: 123
```

Error:
```ERROR: The Compose file './docker-compose.yaml' is invalid because:
Unsupported config option for services.budibase: 'platform'
```

melohagan:
Oh sorry this is the single image. I don't think the single image supports ARM at the moment @mike12345567

shogunpurple:
@melohagan The single image does support ARM. The `platform` option shouldn't be required. Looks like this might be something that we need to look into. I'll try debugging on my ARM mac.

RixzZ:
Same problem on my RaspberryPi.

Seems like you use [this script](https://github.com/Budibase/budibase/blob/develop/scripts/install-minio.sh) to download the correct version of minio, after checking the TARGETARCH env variable.

If I'm not missing something, o n [the Dockerfile](https://github.com/Budibase/budibase/blob/develop/hosting/single/Dockerfile#L23) for the single container, you seem to have the env variable fixed to 'amd64'.

shogunpurple:
[...]

PR #8160, opened 2022-10-06
Fix for ARM64v8 single image build

mike12345567:
## Description
Fixing issue #8079 - making sure TARGETARCH gets correctly exported to environment to pick correct MINIO bundle.

Addresses: 
- #8079


Rory-Powell (approved):
LGTM

Issue #7341, opened 2022-08-18
Need to clear state of all modals when we close them


MihailHadzhiev2022:
Video: https://www.loom.com/share/66c58fcb1c434fc2a5a80510f4304cfd

Test env: https://test.budi.live/

Steps: 

1. Create an app and add two columns(fruits=options, Multi  = Multiselect)
2. Add few rows 
3. Go back to Multi and click edit, delete or add some date and click somewhere on the screen

Expected result: The changes are not saved or deleted
Actual result: The changes are saved and I have to reflesh the page to fix the issue

PR #8162, opened 2022-10-06
Clear edit column modal on close/hide

melohagan:
## Description
Editing a multi-select options and not saving was causing issues. Now clearing state when closing/hiding the edit column modal.

Addresses: 
- https://github.com/Budibase/budibase/issues/7341

## Screenshots
modal

### DIFF
--- a/packages/builder/src/components/backend/DataTable/Table.svelte
+++ b/packages/builder/src/components/backend/DataTable/Table.svelte
@@ -8,6 +8,7 @@
   import CreateEditRow from "./modals/CreateEditRow.svelte"
   import CreateEditUser from "./modals/CreateEditUser.svelte"
   import CreateEditColumn from "./modals/CreateEditColumn.svelte"
+  import { cloneDeep } from "lodash/fp"
   import {
     TableNames,
     UNEDITABLE_USER_FIELDS,
@@ -110,7 +111,7 @@
   }
 
   const editColumn = field => {
-    editableColumn = schema?.[field]
+    editableColumn = cloneDeep(schema?.[field])
     if (editableColumn) {
       editColumnModal.show()
     }


packages/bbui/src/Modal/ModalContent.svelte
```diff
@@ -18,6 +18,7 @@
   export let onCancel = undefined
   export let disabled = false
   export let showDivider = true
+  export let cancelOnClose = false
```
Rory-Powell, line 21: Would we be able to just supply the `onCancel` function?
Or what problem is preventing it? If that doesn't work for some reason we should probably fix that rather than introduce the new field. I think we would always want to run the `onCancel` logic if the modal is exited by the `Cancel` or the `x` button

melohagan: I was worried about changing the behaviour in other modals, the idea of the flag was to make it optional, but I'm definitely happy to remove the flag if you think that's the way to go?

aptkingston:
Sorry I'm only seeing this now @melohagan! I'm pretty sure the core issue here is just that we are mutating the real table definition rather than a clone of the definition. If we just deep clone the field definition before editing, I'm pretty sure we won't need any other changes.

I'm pretty sure that's just the `editColumn` callback in `src/components/backend/DataTable/Table.svelte`:
```js
  const editColumn = field => {
    editableColumn = schema?.[field]
    if (editableColumn) {
      editColumnModal.show()
    }
  }
  ```
  I think all we need to do is update that slightly:
  ```js
...
editableColumn = cloneDeep(schema?.[field])
  ```
That will just ensure that every time you start editing a column it will clone fresh state from the table which can be freely mutated without side effect.

melohagan:
That worked a charm @aptkingston 
Much better! 👌

aptkingston (approved):
Awesome, glad that worked! LGTM!

PR #8175, opened 2022-10-07
Removing Redis SCAN operation from app services

mike12345567:
## Description
Switching from `scan` for app locks to `mget` - which is a fast O(N) operation that only retrieves what we need.

The only scan operation left in the system is handled by the worker around sessions - I would like to remove this ideally but due to the way sessions are keyed it is a little difficult currently (sessions utilise a UUID which we cannot know without searching through Redis - however it is a much less common operation the scan for sessions).

### DIFF
--- a/packages/server/src/api/controllers/application.ts
+++ b/packages/server/src/api/controllers/application.ts
@@ -32,7 +32,7 @@ const {
 import { USERS_TABLE_SCHEMA } from "../../constants"
 import { removeAppFromUserRoles } from "../../utilities/workerRequests"
 import { clientLibraryPath, stringToReadStream } from "../../utilities"
-import { getAllLocks } from "../../utilities/redis"
+import { getLocksById } from "../../utilities/redis"
 import {
   updateClientLibrary,
   backupClientLibrary,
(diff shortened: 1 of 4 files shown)


packages/server/src/api/controllers/application.ts
```diff
@@ -172,16 +171,16 @@ export const fetch = async (ctx: any) => {
...
+    const locks = await getLocksById(appIds)
     for (let app of apps) {
-      if (app.status !== "development") {
-        continue
-      }
-      const lock = locks.find((lock: any) => lock.appId === app.appId)
+      const lock = locks[app.appId]
       if (lock) {
-        app.lockedBy = lock.user
+        app.lockedBy = lock
```
Rory-Powell, line 183: Has the structure of what's in the lock changed?

Rory-Powell (approved):
LGTM, just a question about lock vs lock.user

packages/server/src/api/controllers/application.ts
mike12345567: Yeah - previously the `getAllLocks` return an array, of just all the locks by `{ key: appId, user }` - instead now the Redis system returns an object map of appId to user - may as well do it is way for bulk get of keys rather than array.

PR #8180, opened 2022-10-09
remove move on /

tomamplius:
## Description
original script move folder temporary on linux root fs / 
I think is not a good practice.
in some case, it's create lot of io (if / and sub folder are not in the same filesystem)

### DIFF
--- a/scripts/cleanup.sh
+++ b/scripts/cleanup.sh
@@ -1,16 +1,4 @@
 #!/bin/bash
-dir=$(pwd)
-declare -a keep=("dist" "package.json" "yarn.lock" "client" "builder" "build" "pm2.config.js" "docker_run.sh")
-for moveDir in "${keep[@]}"
-do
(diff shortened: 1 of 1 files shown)


tomamplius:

jonnymccullagh:
Thank you for your contribution @tomamplius I think your code could be further simplified and we could also add some output. I would suggest the following:
```
#!/bin/bash
KEEP="dist|package.json|yarn.lock|client|builder|build|pm2.config.js|docker_run.sh"
echo "Removing unneeded build files:"
ls | egrep -v $KEEP
ls | egrep -v $KEEP | xargs rm -rf
NODE_ENV=production yarn
```

tomamplius:
@jonnymccullagh tkx i don't know egrep before

jonnymccullagh (approved):
Approved - easier to read, better output and less IO.

PR #8203, opened 2022-10-10
API Testing - Screens

Mitch-Budibase:
### Screens API Testing

**Coverage includes:**
- Create Screen
- Fetch Screen
- Delete Screen

###  Create Screens
- Creates a screen for each available role type
- Confirms screen exists against expected roleID

### Fetch Screen
- I am checking `routes` to confirm the created screen exists 
- Confirmation of screen existing within the route provided - In this case the route is `/test` (as determined when the screen is created)

### Delete Screen
- Deletes a screen and confirms statusCode is 200

Each test requires some layering. By that I mean we must include code to create an app before creating a screen. This may need updated/improved in the future


qa-core/src/tests/internal-api/screens/screens.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  })
+
+  it("POST - Create a screen with each role type", async () => {
+    // Create app
+    const [appResponse, app] = await appConfig.applications.create(generateApp())
+    expect(appResponse).toHaveStatusCode(200)
+    expect(app._id).toBeDefined()
+    
+    // Create Screen
+    const roleArray = ["BASIC", "POWER", "ADMIN", "PUBLIC"]
```
Mitch-Budibase, line 29: An alternative for this could be to select a random role when a screen gets created (via `/fixtures/screens.ts`)

In theory, with this test running frequently, we would always get the same coverage of all role types without having to specify anytime we call `generateScreen`

Just curious of any thoughts on this?

shogunpurple (changes requested):
Good job! Just a few bits worth updating, with the proper types.

I think these ones are pretty good as some basic tests to hit the screens endpoints. What we really want to be able to test with the screens and roles is critical functionality such as:
- If I create a screen of a particular role ("admin"), and I try to view it as a "basic" user, will I be blocked?
- If I create a public screen, can I see it as a non-logged in user? What happens if I hit it when logged in, and there's no logged in screens - am I shown public then? 

These kinds of scenarios will likely need user/screen/app creation all rolled into one - I wouldn't let it block this PR but covering things like this will allow us to say things like "Public apps are working" or "basic RBAC works"

qa-core/src/config/internal-api/fixtures/screens.ts
```diff
@@ -0,0 +1,34 @@
+import generator from "../../generator"
+
+const randomId = generator.guid()
+
+const generateScreen = (roleId: string): any => ({
```
Rory-Powell, line 5: I think it would be a really good idea for us to be verifying the response content as well as the status. 

We could add another function here `expectedScreen(roleId, _id, _rev)` and call it after the response completes like:
```ts
expect(response.body).toEqual(expectedScreen(roleId, screen._id, screen._rev)
```

It's likely it will share a lot of fields with the payload as in a lot of cases the api request / response is a direct passthrough to / from databases, so there could be some re-use in the `expectedScreen` function and the `generateScreen` function.

qa-core/src/tests/internal-api/screens/screens.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    // Create app
+    const [appResponse, app] = await appConfig.applications.create(generateApp())
+    expect(appResponse).toHaveStatusCode(200)
+    expect(app._id).toBeDefined()
+    
+    // Create Screen
+    appConfig.applications.api.appId = app.appId
+    const [screenResponse, screen] = await config.screen.createScreen(generateScreen("BASIC"))
+    expect(screenResponse).toHaveStatusCode(200)
+    expect(screen._id).toBeDefined()
```
Rory-Powell, line 65: I would say we could do without these lines in all the tests*
```ts
    expect(appResponse).toHaveStatusCode(200)
    expect(app._id).toBeDefined()

    expect(screenResponse).toHaveStatusCode(200)
    expect(screen._id).toBeDefined()
```
*except for the create screen test which should verify the 200 on screen creation

As we already have dedicated tests for this behaviour. It adds more verbosity to these tests, and we can be sure that the other functionality works if our end result of the screen deletion in this case comes back successful.

qa-core/src/tests/internal-api/screens/screens.spec.ts
```diff
@@ -0,0 +1,71 @@
+import TestConfiguration from "../../../config/internal-api/TestConfiguration"
+import { App } from "@budibase/types"
+import InternalAPIClient from "../../../config/internal-api/TestConfiguration/InternalAPIClient"
+import generateApp from "../../../config/internal-api/fixtures/applications"
+import { Screen } from  "@budibase/types"
+import generateScreen from "../../../config/internal-api/fixtures/screens"
+
+
+describe("Internal API - /screens endpoints", () => {
```
Rory-Powell, line 9: A suggestion for how we could use a very predictable jest layout and naming convention for this and other test going forward:

```ts
  // this tells me the whole root endpoint, and tells me it's internal api
  describe("/api/screens", () => {

    // we incorrectly use post on the root endpoint for both create and update
    // so we could add a distinguishing suffix in some cases
    it("POST - create", () => {})
    it("POST - update", () => {})

    it("GET", () => {})

    // this tells me I am working on a particular screen
    describe("/api/screens/:id/:rev", () => {
      it("DELETE", () => {})
    })

  })
```

The basic idea here being that each unique endpoint has it's own `describe` block and the `it` blocks within define the http method used. This actually highlights that our screen update behaviour is incorrect and should really be nested under `/api/screens/:id`

qa-core/src/config/internal-api/fixtures/screens.ts
Pedro-Revez-Silva: The fields that differ between payload and response can only be verified that they exist. Besides the `createdAt`, how could we know the `_id` to expect a value, for example?

qa-core/src/tests/internal-api/screens/screens.spec.ts
Pedro-Revez-Silva, line 29: In this case there should be tests to verify the access of each role. This can be done in this PR or addressed in another for Roles.

Pedro-Revez-Silva (approved):

qa-core/src/config/internal-api/fixtures/screens.ts
Rory-Powell: fair point, this would only really work for updates. would probably need to check that the fields are first defined, then delete
```ts
delete screen._id
delete screen._rev
delete screen.createdAt
delete screen.updatedAt
```
this would get tedious over time doing for many responses, so you could have a function
```ts
expectDocument = (document: Document) => {
   expect(document._id).tobeDefined()
   delete document._id

   expect(document_rev).tobeDefined()
   delete document._rev

   expect(document.createdAt).tobeDefined()
   delete document.createdAt

   expect(document.updatedAt).tobeDefined()
   delete document.updatedAt
}

expectedScreen = {} // some screen object

expectScreen(screen: Screen) => {
   expectDocument(screen)
   expect(screen).toEqual(expectedScreen)
}
```

Issue #8043, opened 2022-09-29
Custom field in user cannot be updated


reichert:
**Hosting**
- Self
  - Method: docker-compose
  - Budibase Version: 2.0.6

**Describe the bug**
In an app, if I create a new custom field in the User table and try to update that value I receive an error message: document update conflict.

**To Reproduce**
1. Create a custom field in the Users table of a given app
2. Click on edit any user (yourself or other) as admin
3. try to edit the field
4. Save
5. You will see the error
![image](https://user-images.githubusercontent.com/1287032/193050452-cad3a31f-4961-41f5-b17d-98088226e82b.png)

**Expected behavior**
- Have the user column updated

dominik-bsr:
Related: https://github.com/Budibase/budibase/discussions/8081

validlogin:
Here is the error I am getting from the console when trying to update a user and receive this error, maybe it could be helpful in shedding some light at the situation:

`ERROR [1665218212850] (30 on 575fea6ec3a6): Document update conflict.
bbapps                |     Error
bbapps                |         at generateErrorFromResponse (/app/node_modules/pouchdb/lib/index.js:694:18)
bbapps                |         at /app/node_modules/pouchdb/lib/index.js:6816:19
bbapps                |         at runMicrotasks (<anonymous>)
bbapps                |         at processTicksAndRejections (internal/process/task_queues.js:95:5)
bbapps                | Trace: Error
bbapps                |     at generateErrorFromResponse (/app/node_modules/pouchdb/lib/index.js:694:18)
bbapps                |     at /app/node_modules/pouchdb/lib/index.js:6816:19
bbapps                |     at runMicrotasks (<anonymous>)
bbapps                |     at processTicksAndRejections (internal/process/task_queues.js:95:5) {
bbapps                |   message: 'Document update conflict.',
bbapps                |   error: 'conflict',
bbapps                |   reason: 'Document update conflict.',
bbapps                |   docId: 'ro_ta_users_us_0633739477ad4659a21d496d5f81241a'
bbapps                | }
bbapps                |     at /app/dist/api/index.js:74:21
bbapps                |     at Generator.throw (<anonymous>)
bbapps                |     at rejected (/app/dist/api/index.js:6:65)
bbapps                |     at runMicrotasks (<anonymous>)
bbapps                |     at processTicksAndRejections (internal/process/task_queues.js:95:5)
bbapps                | INFO  [1665218212870] (30 on 575fea6ec3a6): request completed
[...]

PR #8212, opened 2022-10-11
Fix user metadata update

Rory-Powell:
## Description
Previously the global user _rev was overriding the metadata _rev, producing a conflict on save. 

Had made an attempt at typing `utilities/users.js` and `utiltiies/global.js`. The main difficulties are the differences between the standard `User` document and the augmented in memory global user after app role processing, where we omit roles and populate a roleId field. I had started to introduce an `AppUser` type to make it clear to consumers what are available in return types, this isn't a simple refactor due to uses of `user` across the above files as well as `controllers/auth.ts` and the groups sdk inside pro, these files sometimes expect the user in both forms at once. This would result in additional refactoring that seemed out of scope for this fix - shelving these changes for a future pr. 

Addresses: 
- https://github.com/Budibase/budibase/issues/8043


aptkingston (approved):
LGTM!

Issue #8209, opened 2022-10-11
User access not paging or counting properly


mjashanks:
**Hosting**

- Self - OSS
- Cloud - pro

**Describe the bug**
Despite #8208, I was able to add more than 8 users, using the API.

However:
1. I cannot move to the second page on the user access.
2. The user count in the overview screen does not count beyond 8.

No.1 above is particularly serious because it means that you cannot change or remove access to users beyond the first page.

Unfortunately you cannot see my mouse pointer in the video below, but take note
1. Overview screen counts 8 users
2. Users list has 8 users, with the "Next Page" arrow active
3. I click the next page button and it does nothing
4. I remove a user at the end to prove that there is actually an additional page

[Screencast from 11-10-22 10:43:08.webm](https://user-images.githubusercontent.com/3524181/195056741-fae34bcd-31fd-4f25-9957-777141da45a1.webm)

Issue #8208, opened 2022-10-11
Cannot add more than 8 users to an app


mjashanks:
**Hosting**
Tested on both
- Self - OSS
- Cloud - Pro

**Describe the bug**
I have imported 25 users into Budibase. When I try to add them all, it's impossible to add more than 8 users (I noted that 8 is also the page size of the users list). 

[Screencast from 11-10-22 10:28:02.webm](https://user-images.githubusercontent.com/3524181/195053182-f10fbfc0-2492-4dac-b253-b7169263ff54.webm)

PR #8218, opened 2022-10-11
Fix user pagination not working when viewing users with access to an app

aptkingston:
## Description
Fixes a bug preventing pagination working in the table showing users assigned to an app.

Addresses: 
- #8209

The other related issue (#8208) will be solved seprately as it's a larger frontend refactor, and in the meantime all functionality is still possible if you enter a search string.


shogunpurple (approved):
LGTM!

PR #8230, opened 2022-10-12
remove ipv6 conf from nginx proxy if ipv6 not enabled

jonnymccullagh:
## Description
During the docker-compose deployment the nginx proxy fails to start when IPv6 is not available as the nginx config includes a line to listen on IPv6 interfaces.
This PR replaces the existing script `10-listen-on-ipv6-by-default.sh` that is provided with the nginx image. That vendor provided script focuses on changing `/etc/nginx/conf.d/default.conf`, however the changes in this PR focus on removing IPv6 conf lines from both potential conf files:
- /etc/nginx/conf.d/default.conf
- /etc/nginx/nginx.conf
The default provided `10-listen-on-ipv6-by-default.sh` aimed to add a listen directive for ipv6 on port 80. Since we appear to have the ipv6 listen in both conf files we are replacing the default script to only remove ipv6 lines if ipv6 is not available. 
We also delete the line regardless of the port number used.


jonnymccullagh:
I'd like to test the image built by this change before announcing on the forums so we know the change is good.

shogunpurple (approved):
LGTM - great work on this one @jonnymccullagh.

Rory-Powell (approved):
LGTM!

Issue #7067, opened 2022-08-03
Navigating the same route doesnt fire screen "On Load" Actions


poirazis:
Self Hosted / Latest / Debian

assume we have a route /members/:id, and we are viewing /members/20.
If we navigate to /members/21 the "On Load" actions of the /members screen will not fire.

Is this a bug or is this the expected behavior ?

PR #8240, opened 2022-10-13
Page load actions now fire on parameter change as well as on mount

Ghrehh:
## Description
Navigating to the same screen with a different set of params e.g. `/resource/1` -> `/resource/2` was not triggering "on screen load" actions as expected.

We were relying on `svelte-spa-router` to remount the screen component, but the router doesn't remount a component if only the parameters change, as mentioned [here](https://github.com/ItalyPaleAle/svelte-spa-router/issues/148#issuecomment-706716065).

This change fixes the issue by making the function that triggers the "on screen load" actions react to changes in `svelete-spa-router` params instead of relying on the `onMount` lifecycle function.

Addresses: 
- https://github.com/Budibase/budibase/issues/7067

### DIFF
--- a/packages/client/src/components/Screen.svelte
+++ b/packages/client/src/components/Screen.svelte
@@ -1,5 +1,6 @@
 <script>
   import { screenStore, routeStore, builderStore } from "stores"
+  import { get } from "svelte/store"
   import Component from "./Component.svelte"
   import Provider from "./context/Provider.svelte"
   import { onMount, getContext } from "svelte"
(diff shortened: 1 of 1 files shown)


Ghrehh:

packages/client/src/components/Screen.svelte
```diff
@@ -15,24 +16,30 @@
...
-    if ($screenStore.activeScreen?.onLoad && !$builderStore.inBuilder) {
-      const actions = enrichButtonActions($screenStore.activeScreen.onLoad, {
-        ...$context,
+    if (screenState.activeScreen?.onLoad && !get(builderStore).inBuilder) {
+      const actions = enrichButtonActions(screenState.activeScreen.onLoad, {
+        ...get(context),
         url: params,
       })
-      if (actions != null) {
+      if (actions !== null) {
```
aptkingston, line 32: Let's keep this as `!=` so we account for both `null` and `undefined` values.

aptkingston (approved):
LGTM! Excellent work @Ghrehh and really impressive to be making your first contribution (and a client library bug no less) so soon!

packages/client/src/components/Screen.svelte
Ghrehh: Ah apologies, missed this, bad copy and paste.

shogunpurple (approved):
LGTM! First PR in the books 👌

Rory-Powell (approved):
Nice fix 👍

joebudi:
Awesome work @Ghrehh .

PR #8241, opened 2022-10-13
Add additional timeout configuration to nginx.dev.conf.hbs

Rory-Powell:
## Description
Update the nginx dev config to use keepalive and larger timeouts. This configuration significantly helped my local dev setup which was frequently getting hanging request, making the builder almost unusable at times. 

References:
- https://stackoverflow.com/questions/18740635/nginx-upstream-timed-out-110-connection-timed-out-while-reading-response-hea
- https://stackoverflow.com/questions/54004946/dockerized-nginx-hangs-after-thousands-of-requests

### DIFF
--- a/hosting/nginx.dev.conf.hbs
+++ b/hosting/nginx.dev.conf.hbs
@@ -24,6 +24,21 @@ http {
     default     "upgrade";
   }
 
+  upstream app-service {
+    server {{address}}:4001;
+    keepalive 32;
+  }
+
+  upstream worker-service {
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
LGTM - although is it possible to specify a `keepalive` higher? In the `server` or `http` block?

aptkingston (approved):
LGTM! Weirdly I have never had issues in firefox, but with chrome my requests get stalled 100% of the time.

PR #8243, opened 2022-10-13
nfs for cloud run

jonnymccullagh:
GCP provides NFS storage as a service (filestore). This PR adds a conditional to the single image runner.sh so that if the NFS variables exist at runtime then we mount the NFS share to the /data directory.
This could be re-used if Azure ever provide NFS as a service.


mike12345567 (approved):
LGTM!

shogunpurple (approved):
LGTM!

Issue #8225, opened 2022-10-12
Plugins, search menu, if I press spacebar all available plugins disappeared 


MihailHadzhiev2022:
Navigate to Plugins and click in the search menu. Press spacebar 
https://www.loom.com/share/2661635d35994cd98b3d3914a49df831

Issue #7342, opened 2022-08-18
Prevent non-static formula columns being selected for datasource sorting


Rory-Powell:
**Hosting**
All

**Describe the bug**
See steps here: https://github.com/Budibase/budibase/issues/6630

**Expected behavior**
- Non static formula columns shouldn't be able to be selected as a datasource sorting column
- Update messaging on `Formula type` tooltip to also mention that sorting is not support (currently mentions filtering only)

PR #8247, opened 2022-10-13
Misc bug fixes

aptkingston:
## Description
A couple of fixes for some bugs.

- Fix bricking the plugins screen when entering a search string with no results #8225
- Prevent invalid sort columns being used for sorting data. Only sortable types are now listed for sort column settings. #7342


Rory-Powell (approved):
LGTM

Issue #4351, opened 2022-02-06
MongoDB w/ a self-signed certificate


davext:
**Describe the bug**
Trying to connect to a Mongo Database with a self-signed certificate, results in the following error:
```
{
    "message": "failed to connect to server [mongo.ondigitalocean.com:27017] on first connect [Error: self signed certificate in certificate chain\n    at TLSSocket.onConnectSecure (_tls_wrap.js:1515:34)\n    at TLSSocket.emit (events.js:400:28)\n    at TLSSocket.emit (domain.js:475:12)\n    at TLSSocket._finishInit (_tls_wrap.js:937:8)\n    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:709:12) {\n  name: 'MongoNetworkError'\n}]",
    "status": 400
}
```

**To Reproduce**
1. Add the self-signed MongoDB using the `mongodb+srv://` link
2. Click on add query and fill out the query with any query
3. Click run query

**Expected behavior**
Either accept the path of the certificate, or accept the certificate’s contents, and pass the certificate to the connector on the backend.

Metabase had the same issue a while ago, and this is what it looks like now: https://user-images.githubusercontent.com/1447303/84531524-0b94a280-ace5-11ea-8056-4a106a9ac22b.png

**Screenshots**

**Desktop (please complete the following information):**
 - OS: any
 - Browser any
 - Version any

**Additional context**

satanch:
Bumping this. Any news?

PR #8260, opened 2022-10-14
Add fields for MongoDB SSL cert file paths

melohagan:
## Description
Users want to be able to log into MongoDB using certificates. The client doesn't offer certificates as a string value, only a file path can be provided. 

Addresses: 
- https://github.com/Budibase/budibase/issues/4351

## Screenshots
Screenshot 2022-10-14 at 11 58 27

`SELF_HOSTED=0`
Screenshot 2022-10-14 at 14 05 11

`SELF_HOSTED=1`
Screenshot 2022-10-14 at 14 08 19

---

## Accordion

Added `fieldGroup` type to allow accordion groups of fields:
Screenshot 2022-10-17 at 13 57 38

It is closed by default when all of the settings within the accordion are empty:
Screenshot 2022-10-17 at 13 59 03

It is open by default when at least one of the accordion field group settings has been entered:
Screenshot 2022-10-17 at 14 22 32

**Creating a new datasource**
Screenshot 2022-10-17 at 14 03 23

*Expanded*
Screenshot 2022-10-17 at 14 05 00


shogunpurple (changes requested):
Code looks good, but this will only work in self host, since it relies on a filepath. I think we should hide the cert fields when running in cloud, until we have some kind of way to upload certs.

For self host, we will also need to add some docs to show how to set up a volume/bind mount for the certificate before using this connector.

melohagan:
> Code looks good, but this will only work in self host, since it relies on a filepath. I think we should hide the cert fields when running in cloud, until we have some kind of way to upload certs.
> 
> For self host, we will also need to add some docs to show how to set up a volume/bind mount for the certificate before using this connector.

Now only showing the certificate fields for self-host

Rory-Powell (commented):
There's quite a lot of configuration here for a likely only a small subset of users. I think we should use a collapsed section

`> Configure SSL`

for these three new fields

melohagan:
> There's quite a lot of configuration here for a likely only a small subset of users. I think we should use a collapsed section
> 
> `> Configure SSL`
> 
> for these three new fields

Accordion added!

packages/server/src/integrations/mongodb.ts
```diff
@@ -25,24 +30,54 @@ interface MongoDBQuery {
...
-      type: DatasourceFieldType.STRING,
-      required: true,
-      default: "mongodb://localhost:27017",
-    },
-    db: {
-      type: DatasourceFieldType.STRING,
-      required: true,
-    },
-  },
+  datasource: environment.SELF_HOSTED
```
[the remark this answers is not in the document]
melohagan: Env check moved into a function for readability.

Rory-Powell (approved):

packages/server/src/integrations/mongodb.ts
```diff
@@ -26,291 +31,330 @@ interface MongoDBQuery {
...
   private config: MongoDBConfig
   private client: any
 
   constructor(config: MongoDBConfig) {
     this.config = config
-    this.client = new MongoClient(config.connectionString)
+    const options: MongoClientOptions = {
+      tlsCertificateFile: config.tlsCertificateFile || undefined,
+      tlsCertificateKeyFile: config.tlsCertificateKeyFile || undefined,
+      tlsCAFile: config.tlsCAFile || undefined,
```
Rory-Powell, line 355: I wonder would it be possible for us to write a text input to temp dir if it doesn't exist, and use that to init the mongo client. This would make the mongo integration consistent with the other ssl configs, e.g. this latest PR for elastic 
https://github.com/Budibase/budibase/pull/8386/files

Have you any thoughts on this @shogunpurple @mike12345567

Rory-Powell:
Should we call the UI configuration `SSL` rather than `TLS`, to be consistent with the other datasource configs?

packages/server/src/integrations/mongodb.ts
mike12345567: I'm not sure if that is safe - writing creds to a file could cause a leak.

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
```diff
@@ -97,6 +123,45 @@
...
+              </button>
+              <svg
+                class="spectrum-Icon spectrum-UIIcon-ChevronRight100 spectrum-Accordion-itemIndicator"
+                focusable="false"
+                aria-hidden="true"
+              >
+                <use xlink:href="#spectrum-css-icon-Chevron100" />
+              </svg>
+            </h3>
+            <div class="spectrum-Accordion-itemContent" role={configKey}>
```
mike12345567, line 149: This should probably be a component in BBUI - rather than configured in the builder.

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
```diff
@@ -10,6 +10,7 @@
   import KeyValueBuilder from "components/integration/KeyValueBuilder.svelte"
   import { capitalise } from "helpers"
   import { IntegrationTypes } from "constants/backend"
+  import "@spectrum-css/accordion"
```
mike12345567, line 13: Inserting these directly into the client/builder is generally a bad idea - if this were a component within BBUI it would fit better with the rest of our spectrum component design.

mike12345567 (changes requested):
This contains some good improvements, would love to get this in - just think that the accordion should be moved into BBUI, rather than created directly in the builder, better for re-usability.

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
melohagan: Moved into BBUI!

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
melohagan: Import removed!

mike12345567 (approved):
LGTM!

shogunpurple (approved):
LGTM!

Issue #5765, opened 2022-05-07
Multi-picker, setting default value makes changing the picker value impossible


mjashanks:
**Hosting**
- Self
  - Method: budi CLI
  - Budibase Version: 1.0.145
  - App Version: 1.0.145

**Describe the bug**
If i set the Default Value of a multi-picker, then its impossible to change the value of the multi-picker in-app

**To Reproduce**
1. Add a multi picker and set it's value to a valid json array
2. Preview your app and browse to screen
3. try to change the picker
**Expected behavior**
the picker should change

**Desktop (please complete the following information):**
 - OS: ubuntu
 - Browser: chromium

amircloner:
Hello 
i have a same issue

for string value , above comment is true 
but for Number values 
you should in Default value > JavaScript > paste this code

```
return $("").split(",").map(Number);
```

aptkingston:
> Hello i have a same issue
> 
> for string value , above comment is true but for Number values you should in Default value > JavaScript > paste this code
> 
> ```
> return $("").split(",").map(Number);
> ```

Could you provide more details? Where do you get your options from - schema, data provider or custom? If a data provider, what database are you using and what is the underlying data type of the value field?

The binding you provided is not valid, but I tested with numbers and this still works.

Using actual numbers as a default value:
![image](https://user-images.githubusercontent.com/9075550/180040869-b8491abe-90ca-4f06-a810-869fb207ffb1.png)

Still works for a default value and allows changing the value:
![image](https://user-images.githubusercontent.com/9075550/180040935-73496db5-edaa-4450-91d6-b9473c13fd92.png)

PR #8282, opened 2022-10-17
Parse multi-select values/defaults to avoid unusable options.

deanhannigan:
## Description
When setting values/defaults for the multiselect component, it was possible to pass in values that could not be removed.

The array of values passed to the picker is now parsed to ensure that:

- it contains valid members of the pickers' potential options
- an attempt will be made to convert the option to a string

Addresses: 
- https://github.com/Budibase/budibase/issues/5765


melohagan (approved):
LGTM!

Issue #7552, opened 2022-08-31
Automations save row - JSON field type not listed 


mjashanks:
**Hosting**
Cloud

**Describe the bug**
I have a BudibaseDB table, which has a JSON column. Using an automation, I am trying to
1. Hit a REST API
2. Save the response into the JSON column

In the Automation "Save Row" step, my JSON column is not listed. So, I can't update it :(

mjashanks:
Changed this to a severity level 5, because I found an easy workaround.

Switch the automation step to "Use Bindings", and the JSON field will appear:

![image](https://user-images.githubusercontent.com/3524181/187920378-bca7e4a6-e183-4e1f-8334-a18327fa3fb8.png)

PR #8283, opened 2022-10-17
Add JSON field to automation step

deanhannigan:
## Description
The JSON field is now included when configuring datasource update steps for automations.

Addresses: 
- https://github.com/Budibase/budibase/issues/7552

## Screenshots
Screenshot 2022-10-17 at 12 30 59


packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -14,6 +14,11 @@
   export let block
   export let isTestModal
 
+  $: parsedBindings = bindings.map(binding => {
+    binding.icon = "ShareAndroid"
+    return binding
+  })
```
aptkingston, line 20: It might be better here to return a new object inside the map function, just so we avoid mutating the real bindings array we pass around. Adding an icon like that is innocent enough but would still be better to avoid mutation.

e.g. just
```js
$: parsedBindings = bindings.map(binding => ({
  ...binding,
  icon: "ShareAndroid"
}))
```

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -18,6 +20,11 @@
   export let bindings
   export let isTestModal
 
+  $: parsedBindings = bindings.map(binding => {
+    binding.icon = "ShareAndroid"
+    return binding
+  })
```
aptkingston, line 26: Same as above just about avoiding mutation.

aptkingston (approved):
LGTM! One minor comment just. Nice to have this available in automations now :+1:

PR #8301, opened 2022-10-18
Clicking the Toggle Button of an Open Action Menu Does Not Close It

Ghrehh:
## Description
Clicking the toggle button of an open action menu causes it to rapidly close then open. This is because both the "outside click handler" of the popover and the open action of the toggle button are firing. This change makes it so that both these actions share a single DOM event and we can call `stopPropagation` after the "outside click handler" action.


aptkingston (approved):
LGTM! Glad this worked, turned out very clean.

PR #8302, opened 2022-10-18
Disable ipv6

jonnymccullagh:
The scripts under `/docker-entrypoint.d` are numbered so they run with 
`10-listen-on-ipv6-by-default.sh` running first and removing any mention of IPv6, then:
`20-envsubst-on-templates.sh` runs and puts an nginx.conf in place that contains an IPv6 line.
So I believe the solution is moving the `10-listen-on-ipv6-by-default.sh` so that it runs later i.e. `80-listen-on-ipv6-by-default.sh`


shogunpurple (approved):
LGTM

PR #8307, opened 2022-10-18
Stop Screen Store Double Updating on Screen Change

Ghrehh:
## Description
The screenStore and the routerStore which partially composes it are updating twice on a screen change. This is due to two separate actions being used to update the active route and the route params of the routerStore. 

Moving the route params update action from the `Screen` component to the same callback where the active route update action is called in the `Router` component fixes this issue as Svelte seems to batch the updates.


aptkingston (approved):
LGTM!

PR #8319, opened 2022-10-18
API tests for tables, columns and rows

Pedro-Revez-Silva:
## Description
Added the scenario to test operations on Tables, Columns and Rows. Also made some small refactors to previous code.


qa-core/src/tests/internal-api/applications/create.spec.ts
```diff
@@ -185,11 +171,72 @@ describe("Internal API - /applications endpoints", () => {
...
-    const [response, app] = await config.applications.create(generateApp())
-    expect(response).toHaveStatusCode(200)
-    expect(app.appId).toBeDefined()
+    const app = await config.applications.create(generateApp())
 
     const [deleteResponse] = await config.applications.delete(<string>app.appId)
     expect(deleteResponse).toHaveStatusCode(200)
   })
+
+  it("Operations on Tables", async () => {
```
Rory-Powell, line 180: Doesn't seem quote right for this to live under `applications`. Given all the assertions here are specific to a table, this should probably live in the table suite

qa-core/src/config/internal-api/fixtures/table.ts
```diff
@@ -0,0 +1,60 @@
+import { Row } from "@budibase/types"
+
+type Table = {
```
Rory-Powell, line 3: We can import the existing definition for a table from budibase. If there is data missing from that definition we should add to the existing one so we aren't repeating ourselves

qa-core/src/config/internal-api/TestConfiguration/tables.ts
```diff
@@ -0,0 +1,64 @@
...
+        return [response, json]
+    }
+
+    async update(body: any): Promise<[Response, Table]> {
+        const response = await this.api.put(`/tables`, { body })
+        const json = await response.json()
+        return [response, json]
+    }
+
+    async getRows(id: string): Promise<[Response, Row[]]> {
```
Rory-Powell, line 46: These row operations should probably live in a `RowsApi`

qa-core/src/config/internal-api/TestConfiguration/tables.ts
```diff
@@ -0,0 +1,64 @@
...
+        return [response, json]
+    }
+
+    async getTableById(id: string): Promise<[Response, Table]> {
+        const response = await this.api.get(`/tables/${id}`)
+        const json = await response.json()
+        return [response, json]
+    }
+
+    async create(body: any): Promise<[Response, Table]> {
```
Rory-Powell, line 28: should probably call this `save`, given it's used for the create and the update

qa-core/src/config/internal-api/fixtures/table.ts
Pedro-Revez-Silva: Done

qa-core/src/config/internal-api/TestConfiguration/tables.ts
Pedro-Revez-Silva: Done

qa-core/src/config/internal-api/TestConfiguration/tables.ts
Pedro-Revez-Silva: Done

qa-core/src/config/internal-api/TestConfiguration/rows.ts
```diff
@@ -0,0 +1,28 @@
...
+import InternalAPIClient from "./InternalAPIClient"
+
+export default class RowsApi {
+    api: InternalAPIClient
+
+    constructor(apiClient: InternalAPIClient) {
+        this.api = apiClient
+    }
+
+    async getAll(id: string): Promise<[Response, Row[]]> {
```
Rory-Powell, line 12: Might make sense to make this explicit with `tableId`

packages/types/src/documents/app/table.ts
```diff
@@ -49,4 +49,10 @@ export interface Table extends Document {
   sourceId?: string
   relatedFormula?: string[]
   constrained?: string[]
+  _id?: string
+  _rev?: string
```
Rory-Powell, line 53: We already get these base fields `_id`, `_rev`, `createdAt` and `_updatedAt` from the base `DocumentType`

Pedro-Revez-Silva: Solved

qa-core/src/config/internal-api/TestConfiguration/rows.ts
Pedro-Revez-Silva: Artifact from when this method was part of tables class. Thank you for pointing that out

Mitch-Budibase (approved):
LGTM - Nice work Pedro!

shogunpurple (approved):
LGTM! Nice work Pedro

Issue #7996, opened 2022-09-27
App Backups (API)


shogunpurple:
When I access the app overview screen, I should be able to see all the backups for that particular app, so that I can restore to a particular version of that application.

COS
- API for CRUD on backups
- API to allow users to restore to a particular version of the backup
- Ensure APIs are behind `pro`
- Implementation of the backup process, export and restore
- Includes events for restoring backup
- Licensing/tier updates for backup management
- Pagination for fetch of backup metadata

Issue #7994, opened 2022-09-27
Backup on Publish


shogunpurple:
When I publish an application, I want an app backup to be saved with the revision name of `<app-name>-publish-<timestamp>` and the type of `publish` so that I can view my published app versions from the UI.

**COS**
- Show a tickbox in the publish screen (on by default) to publish a revision
- Provide an input box (defaulted to `<app-name>-publish-<timestamp>` that can be changed
- On publish, write a backup of the app to object storage
- Make sure this op is async as to not slow down deployments

Issue #3378, opened 2021-11-14
`Export apps` does not include attachments


gvlx:
**Describe the bug**
The exported app file does not have all the attachments included in the database.

**To Reproduce**
1. Create an app with a table with attachment fields
2. Add a few records with attachments
3. Export the app to a file
4. Delete the app
5. Import the app back into bb using the file created in the third step.
6. The attachments in the database are lost.

**Expected behavior**
Importing the database also restores the attachments.

**Desktop (please complete the following information):**
 - OS: W10
 - Browser Firefox 94.0.1
 - Version 0.9.185

**Additional context**
The attachments could be exported to a zip file, and after re-creating the database, the zip file could be imported and its contents re-linked.

PR #8354, opened 2022-10-21
App backups backend

mike12345567:
## Description

This coincides with the pro work in: https://github.com/Budibase/budibase-pro/pull/71

This contains the main work for two big pieces of work:

1. Including attachments in app exports
2. The ability to import an app over an existing app

Before now it wasn't possible to import an app to an existing database, app imports would always generate an entirely new app. Now there is an option to import over an existing app using the new pro backups feature.

The main OSS feature being added here is the ability to export apps with their attachments. This was the main bulk of the work as it also needed to include handling of the larger full app list export from the Cloud for self host seeding. Now when an export occurs we generate a tarball, containing a list of attachments, with the app database file included within. When we import we can open the tarball, import the files and then import the database afterwards. One large issue this generated was the need to update all of the attachment URLs - to handle this we are now using direct Mango queries. Basically, we need to iterate through all rows and find ones that have an attachment column that needs updated, rather than building an index to do this once we can use a one time temporary Mango query. This internally does scan the database, but as this is an async operation it doesn't matter how long it takes to complete.

I've also started building out an SDK within the server, I wanted to centralise a lot of logic that utilises a lot of different underlying DB logic and doesn't really fall into its own sub-system, overtime hope to expand on this and add more useful functions that get used throughout all the other components of the server.

This does not currently have an tests added to it - these will be added during the code freeze as it is the best time to fill these out.

Addresses: 
- #3378 - exports now contain the attachments as well as the DB information. This took a bit longer than expected as the export/import process is quite a bit more complicated, as well as supporting the export all apps for a new self hosted instance function that we have in the Cloud.
- #7994 - As the title says, this now triggers a backup as part of the publishing of an app - meaning users should have the ability to rollback to each publish.
- #7996 - The main API for the backups - this is mostly handled in the pro, but the processing for export/import is handled from the OSS codebase.

UI is being developed separately on the `feature/backups-ui` branch.

## Refactors
There were some hefty refactors that took place during this update, to allow pro to access some features that previously were only available inside deep components of the server. These include:
1. Refactoring the bull system to make it more generic and move it into backend core. This included a fair amount of Typescript conversion, trying to improve our type checking around data being queued/de-queued allowing for better static analysis of the consumers of queues.
2. Breaking the app export/import logic out into its own separate SDK, so that it can be re-used in a few places.
3. Combining the `DocumentType` enum in the backend-core with the one in server, ideally in the future want to simplify all of our document params/ID generation down into a single library in backend-core, meaning that it can be more easily shared between pro, server and worker.
4. Making it easier to perform direct queries to CouchDB, as part of support Mango syntax queries.
5. Various typescript expansions, type improvements, conversions.


packages/server/package.json
```diff
@@ -159,12 +159,12 @@
...
     "@types/apidoc": "0.50.0",
     "@types/bson": "4.2.0",
-    "@types/bull": "3.15.8",
     "@types/global-agent": "2.1.1",
     "@types/google-spreadsheet": "3.1.5",
+    "@types/ioredis": "4.28.10",
     "@types/jest": "27.5.1",
-    "@types/koa": "2.13.4",
-    "@types/koa__router": "8.0.0",
+    "@types/koa": "2.13.5",
```
Rory-Powell, line 166: Would be best using `2.13.4` to match the version of koa exactly

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -98,13 +101,25 @@ async function initDeployedApp(prodAppId: any) {
...
-async function deployApp(deployment: any) {
+async function deployApp(deployment: any, userId: string) {
   let replication
   try {
     const appId = getAppId()
     const devAppId = getDevelopmentAppID(appId)
     const productionAppId = getProdAppID(appId)
 
+    // can't do this in test
+    if (!env.isTest()) {
```
Rory-Powell, line 112: I think we should start straying away from the `isTest` function, as it causes our test and production paths to diverge.
We might be better mocking the `triggerAppBackup` function instead, so our test can assert this is called with the right params.

packages/server/src/api/controllers/cloud.js
```diff
@@ -55,6 +34,28 @@ async function getAllDocType(db, docType) {
   return response.rows.map(row => row.doc)
 }
 
+exports.exportApps = async ctx => {
+  // if (env.SELF_HOSTED || !env.MULTI_TENANCY) {
+  //   ctx.throw(400, "Exporting only allowed in multi-tenant cloud environments.")
```
Rory-Powell, line 39: Should this comment be re-enabled?

packages/backend-core/src/queue/queue.ts
```diff
@@ -0,0 +1,51 @@
...
+
+async function cleanup() {
+  for (let queue of QUEUES) {
+    await queue.clean(CLEANUP_PERIOD_MS, "completed")
+  }
+}
+
+export function createQueue<T>(
+  jobQueue: JobQueue,
+  removeStalled?: StalledFn
```
Rory-Powell, line 21: This feels like a good candidate for an `opts` param

packages/server/src/sdk/app/backups/backup.ts
```diff
@@ -0,0 +1,145 @@
...
+import fs from "fs"
+import env from "../../../environment"
+
+type BackupOpts = {
+  doc?: { id: string; rev: string }
+  createdBy?: string
+}
+
+async function removeExistingApp(devId: string) {
+  const devDb = dbCore.dangerousGetDB(devId, { skip_setup: true })
```
Rory-Powell, line 22: Could use a `try`-`finally` here to ensure the destroy is called

packages/server/src/sdk/app/backups/backup.ts
```diff
@@ -0,0 +1,145 @@
...
+  await devDb.destroy()
+}
+
+async function runBackup(
```
Rory-Powell, line 26: There's a lot of pro backup specific logic here, we could probably hide this all behind pro if we had a function for initialising the backups sdk like:

```ts
backups.init({
  importFn,
  exportFn,
  statisticsFn
})
```

This is similar to how we have the `init` function on backend core. We could make it generic so that we could say

```ts
pro.init({
   backups: {...} // as above
})
```


Pro could then setup the processors directly, and we could call `runBackup` directly from the consumer in app-service without this wrapper file.

packages/types/src/core/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./db"
```
Rory-Powell, line 1: There is already the top level `sdk` package inside types. Functionally that is the `core` module, for everything that isn't an api request or a document

packages/types/src/sdk/koa.ts
```diff
@@ -7,7 +7,15 @@ export interface ContextUser extends User {
...
-  body: any
+  status?: number
+  request: {
+    body: any
+  }
+  params: any
+  body?: any
+  redirect?: any
+  attachment: any
+  throw: any
```
Rory-Powell, line 20: We shouldn't need any of these, we `extend Context` so that we get the types provided directly from `koa`.

packages/worker/src/sdk/users/users.ts
```diff
@@ -77,6 +79,9 @@ export const paginatedUsers = async ({
   } else if (email) {
     userList = await usersCore.searchGlobalUsersByEmail(email, opts)
     property = "email"
+  }
+  if (userIds) {
+    // TODO: search users by userIds
```
Rory-Powell, line 84: I think we need to implement this for the backups-ui to be able to search for the users associated with the backups loaded via the `createdBy` field

mike12345567: Ah I'll remove this - we don't need to do this because its done on the backend, enriched instead, much more efficient that way.

packages/types/src/sdk/koa.ts
mike12345567: I had to remove the extending of Context, I was getting errors all over the place actually using the `BBContext` - for some reason it was baffling the TS transpiler.

packages/server/src/api/controllers/deploy/index.ts
mike12345567: Sadly this would be a nightmare - I think its better to keep using the `isTest` for the time being, its a great marker for stuff that we need to update as well! Mocking out the entirety of the object store would be painful as it involves file system access, and would all just be ripped out next month anyway.

packages/server/src/sdk/app/backups/backup.ts
mike12345567: Sorry - not quite sure what you mean, if an error occurs in the `dbCore.dangerousGetDB` then there won't be anyway to destroy the db anyway - do you mean in the import processor?

Rory-Powell: Ah sorry, I mean for a `db.close`. Do we need to add one here, or will destroy auto close?

packages/server/src/api/controllers/deploy/index.ts
Rory-Powell: I don't think we need to mock object store here, just mock pro by doing 

```ts
const pro = jest.requireActual("@budibase/pro")
jest.mock("@budibase/pro", () => ({
  ...pro,
  backups: {
    triggerAppBackup: jest.fn(),
  },
}))
```
jest.mock(

packages/types/src/sdk/koa.ts
Rory-Powell: In this case I think you should switch out the `BBContext` for an `any` for now. We should jump on a call to try to resolve why it wasn't compiling for you, as I'm able to get full access to the correct koa types locally

packages/server/src/sdk/app/backups/backup.ts
mike12345567: it auto closes in these scenarios, generally we don't need to worry about it unless you were to open it in a global context! But it'll close automatically when garbage collected.

packages/worker/src/sdk/users/users.ts
Rory-Powell: Yep far better, good shout 👍

packages/server/src/api/controllers/deploy/index.ts
Rory-Powell: Added in: https://github.com/Budibase/budibase/pull/8354/commits/0274b6755a20ec330c8a6321108f851e615ac43c

packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+    // use the CouchDB Mango query API to lookup rows that have attachments
+    const params: CouchFindOptions = {
+      selector: {
+        _id: {
+          $regex: `^${TABLE_ROW_PREFIX}`,
+        },
+      },
+    }
+    attachmentCols.forEach(col => (params.selector[col] = { $exists: true }))
```
shogunpurple, line 51: Nice work getting this logic together - I'm thinking that when we start to look at "global media" we should rethink how URLs are stored, and probably not store them in the DB like this - possibly another part of using better bucket keys as well.

shogunpurple (approved):
Excellent piece of work @mike12345567 - really complete. Solid type conversions in here as well. Looking forward to having a play around with it.

packages/server/src/sdk/app/backups/imports.ts
mike12345567: Agreed, I would rather we didn't store URLs in the documents and instead just enrich something on the way out, as right now it means migrating them if the app ID changes etc etc.

PR #8355, opened 2022-10-21
Fix Picker Interaction with the Outside Click Action

Ghrehh:
## Description
A previous change I made in #8301 changed the functionality of the outside click action, which inadvertently broke some picker components. This should hopefully fix the issue, I tested all the components I could see/think to be relevant in both the builder and app views. Would appreciate some more manual testing by folks more experience with the site. 

I also added the fix from #8301 to the other pickers that weren't closing properly when toggled, like the icon and color picker.


melohagan (approved):
LGTM! 👍

Issue #8357, opened 2022-10-21
Icon component color picker causes Screen settings to anchor on left


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
Using the color setting on the Icon component causes a weird UI glitch in the builder where the screen settings panel will stick to the left.
In addition, a similar issue exists for the color picker within a container, although it flashes and does not stick (see gif).

**To Reproduce**
1. Add an Icon to a screen
2. Open the color picker
3. Click within the screen area
4. The screen settings panel will appear on the left anchored (see screenshot).

**Expected behavior**
Component settings and screen settings should not appear on the screen at the same time.

**Screenshots**
![Screenshot 2022-10-21 at 15 53 44](https://user-images.githubusercontent.com/101575380/197225384-f0fadde2-85e6-4e06-bb07-6efbd8bc8b66.png)

![glitch](https://user-images.githubusercontent.com/101575380/197225493-99ff7db9-e99b-4bf3-8436-ed8cd75f625a.gif)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome

PR #8358, opened 2022-10-21
Don't apply color picker transition to the screen panel

melohagan:
## Description
Color picker no longer causes the screen panel to appear.

Addresses: 
- https://github.com/Budibase/budibase/issues/8357


shogunpurple (approved):
LGTM!

PR #8373, opened 2022-10-24
Backups UI

PClmnt:
## Description

**Targeting this aghainst Michael's branch until his is merged in**

This PR implements the UI for the backups feature. 
image

The Main UI with associated columns, you can see here that restore events don't have counts for datasources / screens etc. They just denote that a restore took place and when it happened. The tags in the status column denote what stage a backup is at, as some backups may take a while to complete, are pending and completed states. This table is also fully paginated. 

image

image

On that page you can also filter by the type of backup, or for backups between a date range, some slight changes tp the DatePicker component where made to acount for this ability. Selecting a range works right down to the time during a day, so you can be quite specific.

image

The actions available to each row are as above. Update shows a modal allowing the user to edit the name of the modal, while clicking restore takes the user through the flow of restoring a backup: 

image

image
These modals show what backup is to be restored and the second modal is to ensure that when a user restores, they also create a backup. In case the restore goes wrong or they want to revert to the previous version. 

image
When a user does not have a pro plan

image
And when a user has a pro plan but no backups created


packages/builder/src/stores/portal/backups.js
```diff
@@ -0,0 +1,57 @@
...
+
+  async function deleteBackup({ appId, backupId }) {
+    return API.deleteBackup({ appId, backupId })
+  }
+
+  async function createManualBackup(appId, name) {
+    return API.createManualBackup(appId, name)
+  }
+
+  async function downloadBackup({ appId, backupId }) {
```
Rory-Powell, line 37: Is this used?

packages/builder/src/components/portal/overview/backups/ActionsRenderer.svelte
```diff
@@ -0,0 +1,115 @@
...
+<ConfirmDialog
+  bind:this={updateDialog}
+  disabled={!name}
+  okText="Confirm"
+  onOk={onClickUpdate}
+  title="Update Backup"
+  warning={false}
+>
+  <Input onlabel="Backup name" placeholder={row.name} bind:value={name} />
+  <ConfirmDialog />
```
Rory-Powell, line 104: Should this nested confirm dialog be here?

Rory-Powell (approved):
LGTM - minor comments just

packages/bbui/src/Form/DatePicker.svelte
```diff
@@ -14,11 +14,17 @@
...
-
+  export let range = false
   const dispatch = createEventDispatcher()
 
   const onChange = e => {
-    value = e.detail
+    if (range) {
+      // Flatpickr cant take two dates and work out what to display, needs to be provided a string.
+      // Like - "Date1 to Date2". Hence passing in that specifically from the array
+      value = e.detail[1]
```
shogunpurple, line 24: is there potential for this to be `undefined`?

packages/builder/src/components/portal/overview/backups/StatusRenderer.svelte
```diff
@@ -0,0 +1,15 @@
+<script>
+  import { Badge } from "@budibase/bbui"
+
+  export let value = "started"
+  $: status = value?.charAt(0).toUpperCase() + value?.slice(1)
```
shogunpurple, line 5: NAB: Since ES5 you can just use array indexes instead of `charAt` - `value[0]`

shogunpurple (approved):
LGTM! Great work

packages/bbui/src/Form/DatePicker.svelte
PClmnt: I don't believe so, when the range variable is passed to flatpickr it will always return an element in that position, but I will check it just to be sure, might be an edge case etc

joebudi:
Looks good @PClmnt . Might be an idea to go with sentence case for the dialog titles:
https://spectrum.adobe.com/page/alert-dialog/

Also, could we not auto-name the backup for the rollback and save the user from entering text? If this was the case, you could potentially combine the two alerts.

shogunpurple:
@joebudi the reason that manual backups are named is that they are more likely to be "checkpoints" - ie. version numbers, or "Pre migration backup". I do think that it might be nice to pre-fill the text though, and allow the user to edit it if desired.

joebudi:
> These modals show what backup is to be restored and the second modal is to ensure that when a user restores, they also create a backup.

Ah. I thought when someone wants to restore a version, they have to manually backup too. I thought, in this specific scenario, you could just auto-name it.

packages/bbui/src/Form/Core/DatePicker.svelte
```diff
@@ -64,9 +65,8 @@
     if (newValue) {
       newValue = newValue.toISOString()
     }
-
     // If time only set date component to 2000-01-01
-    if (timeOnly) {
+    else if (timeOnly) {
```
aptkingston, line 69: What was the reason for adding this `else`? Adding that in will prevent the `timeOnly` logic from ever running if there's an existing value.

aptkingston (commented):
Just adding some quick UI comments - I started this last night but I see it's been merged already. Can't test the main backups UI because I can't seem to create any backups so these are just some minor things that I was able to see.

packages/builder/src/components/portal/overview/backups/CreateBackupModal.svelte
```diff
@@ -0,0 +1,21 @@
...
+</script>
+
+<ModalContent
+  onConfirm={() => createManualBackup(name)}
+  title="Create new backup"
+  confirmText="Create"
+  ><Input
+    label="Backup name"
+    bind:value={name}
+    placeholder={"test"}
```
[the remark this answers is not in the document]
aptkingston: Think you've already fixed this one.

aptkingston:
All works really well - nice work! :+1: 

Few more comments from using the main feature:
- [ ] Server side sorting is not supported, so we should disable sorting by clicking on table headers as this simply sorts the page you're viewing client side. Everything should always just be ordered by time as returned by the server
- [ ] If you change your filters to return no results (either by changing type, or date range) you get shown the empty state as if you have no backups. This makes it impossible to clear the filter or go back unless you change the page. We need to distinguish a filtered empty state, from a true empty state, and just show an empty table here
- [ ] We could space the table out a bit as the content is fairly squished, but there's loads of empty space on the right side
image
- [ ] The deletion confirmation modal is missing a question mark. The text just flows into a single sentence
image

Issue #7896, opened 2022-09-22
Grid Layout


shogunpurple:
When I'm laying out a screen, I want to be able to control the width, height and spanning of elements with a drag and drop grid based system.

**COS**
- Allow users to drag elements into the grid
- Do not move other elements out of the way when dragged over
- Allow users to control the rows/column span of grid items from the individual items themselves
- Allow users to drag the items within the grid to resize them

PR #8376, opened 2022-10-24
Grid component + builder performance improvements

aptkingston:
## Description
This PR adds a grid component which can be used to easily create custom layouts without needing to nest containers or understand flex. There are also multiple improvements to performance and user experience when using DND (both grid, and non-grid).

## Grid component
The grid component lives in the layout category:
image

By default it will render a 12x12 grid that is 400px tall. Grids have axis lines when inside the builder, but these are not rendered inside real apps:
image

Grids have configurable numbers of rows and columns, and these are the only settings available for the grid:
image

Grid size can go as low as 1x1 (mostly to support things like 3x1), and as high as 32x32. Attempting to enter numbers outside this range will instead just lock it to our limits:
image

"Normal" drag and drop works for moving component between grid children, for adding component inside grid children, for dragging components outside of grid children and for dragging components into the grid itself:

https://user-images.githubusercontent.com/9075550/197703909-a1e6726e-08e7-4ab6-a3cf-bf877eeef19b.mp4

The only type of "normal" drag and drop that does not work is moving direct grid children out of the grid (as this uses the new handler, and drags the component around inside the grid instead).

### Grid children
As we still depend on flex heavily and use that as our primary layout technique, the main caveat of using this new grid component is that only containers and blocks can be added to grids directly. Any components can then be placed inside a container - but the immediate child of a grid must be a container, or a block. Although that might seem weird, it's very important. It also ensures that you can lay things out properly, because you can set styles on your containers to center or position child elements as your see fit, ensuring your componants are correctly aligned in the grid cells that you desire.

Adding a component to the grid will (by default) make the new component occupy a quarter of available grid space:

https://user-images.githubusercontent.com/9075550/197528652-b9de2075-5f22-4042-8118-00d582841440.mp4

Direct children of grids (containers and blocks, as previously mentioned) have a completely new drag and drop experience. These children can be dragged around and positioned anywhere inside the grid, automatically locking to the grid axis lines.

https://user-images.githubusercontent.com/9075550/197529013-8128c25b-db8a-4aec-ba71-2bb1a3a6cfc4.mp4

Grid children also have resize anchors on all corners and edge midpoints, allowing you to resize them as you see fit.

https://user-images.githubusercontent.com/9075550/197529346-07bc34a0-7252-40b8-a88a-12a36208fd74.mp4

**Components inside grids do not need to be selected to be dragged or resized. You can drag or resize any component, no matter which component is currently selected.**

Quick example of laying out 3 columns of text:

https://user-images.githubusercontent.com/9075550/197530608-824c3459-4c33-455f-942f-3851cbc085d3.mp4

## Performance improvements
- Fixed issue where any builder change caused the app preview to reinstantialise twice due to the location object changing
- Improve DND experience by removing flashing dropped components in their old positions before saving
- Improve DND smoothness by reducing page reflows and scoping context changes better

## Component nesting validation rewrite
Since this new grid component is the strictest component in terms of what children it can take, the nesting validation had to be improved to ensure there were no ways no get around our `illegalChildren` and new `legalDirectChildren`. The validation has been fully rewritten and applied at the screen level, meaning before any screen doc is saved, the entire tree is recursively checked to ensure no illegal component orders are present. 

This properly valdiates all cases - like tree DND, preview DND, copy and paste, DND from new component panel, keyboard shortcuts etc.

If you now try to illegally nest components (like a grid inside a grid) it will fail and give you a notification:
image

This new validation is also respected in the new component panel, where only valid components will be shown. For exampe, the container is the only component allowed inside grids (and blocks, in the other tab):
image

### Message passing rewrite
Communication between the app preview and builder is crucial for many functions, but is currently implemented as synchronous message passing without response. This is responsible for many instances of "jank" - like flashing of components in weird or stale positions.

This has now been rewritten so that the builder now basically sends an ack back down when events are handled and completed. The client library interface has been promisified, so from client logic you can `await` an async event happening in the builder (like saving a screen), which will resolve whenever the operation has been completed and the builder has sent back an ack. This is responsible for the majority of the improvements listed above.

### Legacy code removal
Code for handling iframe communication using the legacy listener has been removed. This was shipped over a year ago, so it is safe to assume everyone has upgraded since then.

### New `legalDirectChildren` entry in `manifest.json`
This new flag allows specifying a component whitelist, representing the entire set of components that are allowed as direct children. Any components not in this list are not allowed as direct children (but may still be nested deeper in the tree). This currently just facilitates the grid component, but can be used in future for other restricted components.

## FAQ
- Can columns/rows be different sizes?
> No. If you need more granular control of position or sizing, increase the resolution of your grid and span more rows/columns. (Technically you can use custom CSS to accomplish this, but the axis guidelines will not match up visually, and drag and drop will not work as expected).
- Why is there no gap setting on the grid?
> There is no gap setting on the grid so that the axis guidelines and snapping to axis lines works properly. To add spacing between components, either add padding to the components in your grid or, even easier, leave an empty row/column between them.
- Can I overlap components? Will my other components move?
> You can overlap components as much as you want. Other components will *not* move. Components are rendered in order, so the last component in your tree is on top. You can set custom CSS z-index or reorder your components to change your depth. While dragging, the dragged component is automatically brought on top of everything else.
image

Addresses: 
- #7896


joebudi:
Nice work @aptkingston . I'm curious how you managed to default components to the size/width of the grid? Also, if I add 5 tables in a row, will it increase the width of the grid, or decrease the size of the tables, or wrap the tables onto a new row?

aptkingston:
> Nice work @aptkingston . I'm curious how you managed to default components to the size/width of the grid? Also, if I add 5 tables in a row, will it increase the width of the grid, or decrease the size of the tables, or wrap the tables onto a new row?

The grid wins basically - so it will shrink anything inside it down to the size you configure it to be. If you added 5 tables in a row (in a 10 column grid, for example) and resized all to just use 2 columns, then they will stay that small. It will look awful and be unusable of course, but basically whatever size you choose it will stick with.

It just means the person designing the app has complete control and components will respect the size you pick.

packages/client/manifest.json
```diff
@@ -4566,6 +4566,45 @@
       }
     ]
   },
+  "grid": {
+    "name": "Grid",
```
shogunpurple, line 4570: We should make this Grid (Beta) as we discussed, just to make it more clear.

shogunpurple (approved):
LGTM! Just the little (Beta) thing beside grid as we look to understand usage.

aptkingston:
Updated description to add missing video of normal drag and drop working with the grid for reference:

https://user-images.githubusercontent.com/9075550/197704147-7a64d26a-67f5-4fe8-bb48-88df1c21b5b6.mp4

packages/client/manifest.json
aptkingston: Will do. Wasn't sure what way we had left it after the few changes to the scope and spec.

PR #8386, opened 2022-10-25
Custom ssl config for Elasticsearch datasource

chaoticefx:
## Description
Implements fields to allow for customization of the SSL object passed to the Elasticsearch javascript client in the Elasticsearch datasource.

This allows Budibase to connect to Elasticsearch instances running using the default docker-compose.yml file from Elastic which generates a self-signed CA and certificate upon start-up.

## Screenshots
elastic

### DIFF
--- a/packages/server/src/integrations/elasticsearch.ts
+++ b/packages/server/src/integrations/elasticsearch.ts
@@ -9,6 +9,9 @@ const { Client } = require("@elastic/elasticsearch")
 
 interface ElasticsearchConfig {
   url: string
+  ssl?: boolean
+  ca?: string
+  rejectUnauthorized?: boolean
 }
 
 const SCHEMA: Integration = {
@@ -22,6 +25,21 @@ const SCHEMA: Integration = {
       type: DatasourceFieldType.STRING,
       required: true,
       default: "http://localhost:9200",
+    },
+    ssl: {
+      type: DatasourceFieldType.BOOLEAN,
+      default: false,
+      required: false,
+    },
+    rejectUnauthorized: {
+      type: DatasourceFieldType.BOOLEAN,
+      default: true,
+      required: false,
+    },
+    ca: {
+      type: DatasourceFieldType.LONGFORM,
+      default: false,
+      required: false,
     },
   },
   query: {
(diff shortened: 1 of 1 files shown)


chaoticefx:

packages/server/src/integrations/elasticsearch.ts
```diff
@@ -81,7 +99,25 @@ class ElasticSearchIntegration implements IntegrationBase {
...
   constructor(config: ElasticsearchConfig) {
     this.config = config
-    this.client = new Client({ node: config.url })
+
+    let newConfig = { 
+      node: this.config.url,
+      ssl: this.config.ssl 
+        ? {
+          rejectUnauthorized: this.config.rejectUnauthorized,
+          ca: this.config.ca ? this.config.ca : undefined
```
melohagan, line 108: This could be an or statement, e.g. `ca: this.config.ca || undefined`

packages/server/src/integrations/elasticsearch.ts
```diff
@@ -81,7 +99,25 @@ class ElasticSearchIntegration implements IntegrationBase {
...
+        }
+        : undefined,
+    }
+
+    if (newConfig.ssl && !newConfig.ssl.ca)
+    {
```
melohagan, line 114: Please wrap the brace on the same line as the statement, e.g.
```
if (newConfig.ssl && !newConfig.ssl.ca) {
...
}
```
same for the *else if* below.

melohagan (approved):
LGTM! Thanks for the contribution! 👍

packages/server/src/integrations/elasticsearch.ts
chaoticefx: Thanks for the tip! I’ve amended that line.

packages/server/src/integrations/elasticsearch.ts
chaoticefx: Whoops! Got it, I’ve fixed the styling differences.

packages/server/src/integrations/elasticsearch.ts
```diff
@@ -81,7 +99,23 @@ class ElasticSearchIntegration implements IntegrationBase {
 
   constructor(config: ElasticsearchConfig) {
     this.config = config
-    this.client = new Client({ node: config.url })
+
+    let newConfig = {
```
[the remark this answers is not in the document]
chaoticefx: Hm! I initially attempted this approach which I do agree looks cleaner, but I came up against type errors.

This is the output I get based on edits against my current branch. Perhaps I missed some sort of declaration in another part of the code would allow me to avoid the type errors?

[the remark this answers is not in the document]
Rory-Powell: In this case we can update the import from the elasticsearch dependency from:
```ts
const { Client } = require("@elastic/elasticsearch")
```
To:
```ts
import { Client, ClientOptions } from "@elastic/elasticsearch"
```

This now gives us access to the types provided by the elasticsearch dep. 

Now we can use the `ClientOptions` type by assigning the type to `clientConfig`

```ts
    const clientConfig: ClientOptions = {
      node: this.config.url
    }
```

This should resolve the type error and also make our configuration more type safe.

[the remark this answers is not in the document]
chaoticefx: Ahhh, got it! I've cleaned up the constructor and re-tested the various possible configurations.

Rory-Powell (approved):
LGTM

PR #8392, opened 2022-10-25
Startup fixes + remove bucket default names

Rory-Powell:
## Description

Fixes for

**Migrations not running**
- The `initPro` call results in a never ending promise, as with automations init. 
- This prevented migration and the rest of the startup not running
- Add this to a promises array instead, and await it along with the automations init at the very end of startup

**Reboot trigger tenancy error**
- When running in `SELF = 1` and `MULTI_TENANCY = 1`, the `getAllApps` call produces an error that crashes startup

**Migrations lock error**
- When running in `SELF = 1` and `MULTI_TENANCY = 1`, the migrations fails to determine tenant id and produces an error that crashes startup. 
- Configure this lock as a `systemLock` instead to bypass tenancy 

**Default bucket names (k8s)**
- Remove the default bucket names in k8s. These are defaulted application side where the correct apps bucket name is used

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -79,11 +79,11 @@ spec:
         - name: MINIO_URL
           value: {{ .Values.services.objectStore.url }}
         - name: PLUGIN_BUCKET_NAME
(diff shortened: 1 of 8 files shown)


mike12345567 (approved):
LGTM!

Issue #8202, opened 2022-10-10
Automation email - HTML contents gets replaces by binding value when using looping


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - App Version: 2.0.26

**Describe the bug**
If I use a binding in the HTML contents of an email, it does not include the additional content.
Related discussion: https://github.com/Budibase/budibase/discussions/8191#discussioncomment-3840455

**To Reproduce**
1. Create an app action automation
2. Add an execute query step to return many rows
3. Add a send email step **with looping** over the `{{ steps.1.rows }}`
4. Fill the *HTML Contents* with `test {{ steps.1.rows }}`
5. **Finish and test** -> The emails only include the binding value in the body.
6. Delete the looping
7. **Finish and test** -> The emails now include the word "test" before the binding value

**Expected behavior**
I should be able to include non-binding email content when using looping.

**Screenshots**
**With looping**
![Screenshot 2022-10-10 at 17 10 42](https://user-images.githubusercontent.com/101575380/194909803-387a4ce3-edc4-47d7-abde-b95384e0fd10.png)

![Screenshot 2022-10-10 at 17 09 06](https://user-images.githubusercontent.com/101575380/194909537-e333a70d-fba6-4e23-9b07-e672064e72d2.png)

**Without looping**
![Screenshot 2022-10-10 at 17 11 39](https://user-images.githubusercontent.com/101575380/194909941-1b2536f6-7233-4924-b250-17f307feca7d.png)

![Screenshot 2022-10-10 at 17 11 24](https://user-images.githubusercontent.com/101575380/194909907-d0556128-9022-460c-ae96-eb0dc1ce0fad.png)

PR #8394, opened 2022-10-25
Automations substitute 'loop' in step bindings fixes

melohagan:
## Description
Originally there was an issue with correctly outputting HTML contents to email body, but this issue applied to any bindings that use looping.
In addition, I noticed that the binding substitutions only applied to the last bindings and not all bindings in the statement. 
Fixes added with a couple of unit tests.

Addresses: 
- https://github.com/Budibase/budibase/issues/8202


PClmnt (approved):
This is a superb fix, and really nice to get some test coverage in this area as well as it's quite complicated. LGTM!!

PR #8395, opened 2022-10-25
Assorted Backups UI fixes

PClmnt:
## Description
Most of these are addressing comments left by @aptkingston as we closed it while he was still reviewing 😮 

Fixes:
- `Backup to Back up` where needed
-  Question mark in deletion modal
- Disable backup creation until a name has been etered
- Remove useless space in table
- Disable sort
- Fixed empty states so that when filters are selected, it won't incorrectly show the "splash screen" and instead, shows an empty table. 
- Removed rogue `else if` in DatePicker

### DIFF
--- a/packages/builder/src/components/portal/overview/backups/BackupsTab.svelte
+++ b/packages/builder/src/components/portal/overview/backups/BackupsTab.svelte
@@ -62,24 +62,31 @@
   const schema = {
     type: {
       displayName: "Type",
+      width: "auto",
     },
     createdAt: {
       displayName: "Date",
+      width: "auto",
     },
     name: {
       displayName: "Name",
+      width: "auto",
     },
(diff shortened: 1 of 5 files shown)


packages/builder/src/components/portal/overview/backups/BackupsTab.svelte
```diff
@@ -154,7 +161,10 @@
 
   onMount(() => {
     fetchBackups(filterOpt, page, startDate, endDate)
+    loaded = true
   })
+  $: console.log(loaded)
+  $: console.log(backups.length)
```
Rory-Powell, line 167: Few logs to remove

aptkingston (approved):
LGTM! Good job sorting those :+1:

PR #8399, opened 2022-10-26
Ensure tmpDir/.budibase always exists

Rory-Powell:
## Description
Ensure the `budibaseTempDir` always exists on application startup


PClmnt (approved):
LGTM, nice fix!

PR #8404, opened 2022-10-26
Fix Input Dropdown Outside Click

Ghrehh:
## Description
Another Issue with the #8301, this dropdown wouldn't open because of changes to the outside click handler, should be working now.


aptkingston (approved):
LGTM!

Issue #6048, opened 2022-05-24
Handlebars expressions do not work when an automation is used with MySQL


FlaminWrap:
**Hosting**
- Self
  - Method: BudiCLI
  - Budibase Version: 1.0.178 
  - App Version: 1.0.178 

**Describe the bug**
Handlebars expressions do not work when an automation is used with MySQL datasource and returns a URI malformed error.

**To Reproduce**
1. Create a new app with a MySQL data source
2. Create two tables Record and RecordType
3. Create two columns in RecordType: Type/Text , Value/Number
4. Create at least one row in the with both columns filled in the RecordType table
5. Create the following relationsip between the tables: One RecordType row > mandy Record rows
6. Create a column in the Record table called Value and make it a number type
7. Create an automation to launch when creating a row
8. Create a Query Rows step in the automation to query the RecordType table and filter it by the foreign key from the trigger row
9. Create an Update Rows step in the automation to update the Record Table
10. Set the RecordType field to `{{ trigger.row.RecordType }}`
11. Set the Fk_RecordType_Record field to `{{ trigger.row.fk_RecordType_Record }}`
12. Set the Value field to `{{ steps.1.rows.0.Value }}`
13. Set the RowID field to `{{ trigger.id }}`
14. Press Run test at the top of the automation page
15. There will be an error on the Update Row module looking something like below:

Update Row Input:

`
{
  "row": {
    "tableId": "datasource_plus_fc2c66277ba54f2caac0dc36d5efc531__Record",
    "RecordType": "{{ trigger.row.RecordType }}",
    "fk_RecordType_Record": "{{ trigger.row.fk_RecordType_Record }}",
    "Value": "{{ steps.1.rows.0.Value }}"
  },
  "rowId": "{{ trigger.id }}"
}
`

Update Row Output:

`
{
  "success": false,
  "response": "URIError: URI malformed"
}
`

**Expected behavior**
The values update the record

**Screenshots**
![image](https://user-images.githubusercontent.com/97764630/170086792-1a582c7f-c4b2-4e5f-a847-5336451cf4be.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 101.0.4951.67

**Additional context**
Passing the same values as javascript works. (e.g. `return $("trigger.id");`)
TestApp Export (Not sure how well this will work):
[...]

PR #8428, opened 2022-10-27
Handle mysql id decoding in relationships for automations

melohagan:
## Description
ID can be encoded string, so need to decode if needed.
Also added a minor fix when trying to patch a non-existent row. 

Addresses: 
- https://github.com/Budibase/budibase/issues/6048

### DIFF
--- a/packages/server/src/api/controllers/row/index.ts
+++ b/packages/server/src/api/controllers/row/index.ts
@@ -38,7 +38,7 @@ export async function patch(ctx: any): Promise<any> {
       }
     )
     if (!row) {
-      ctx.throw(404, "Row not found!")
+      ctx.throw(404, "Row not found")
     }
     ctx.status = 200
     ctx.eventEmitter &&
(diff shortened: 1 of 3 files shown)


Rory-Powell (approved):
Looks good. Maybe waiting for a review from @mike12345567 would be best before shipping this one

packages/server/src/api/controllers/row/index.ts
```diff
@@ -30,16 +30,24 @@ export async function patch(ctx: any): Promise<any> {
...
-      datasourceId: tableId,
+  try {
+    const { row, table } = await quotas.addQuery(
+      () => pickApi(tableId).patch(ctx),
+      {
+        datasourceId: tableId,
+      }
+    )
+    if (!row) {
+      ctx.throw(404, "Row not found!")
```
mike12345567, line 41: wouldn't put an exclamation on this - just keep the errors basic.

mike12345567 (approved):
LGTM - on small comment.

PR #8454, opened 2022-10-31
Add Row Explorer Block

Ghrehh:
## Description
Adds the Row Explorer block.

The Row Explorer block when provided with a data source will provide a card per row of that table, when clicked that card will expand into a view allowing the user to see every column of that row.

There weren't any styles given for the mobile view, so I just did something that seemed sensible, see the gif below.

Mobile styles are achieved via custom css and handlebars, I tried using the conditional display functionality but upon ejecting the component, only the currently visible components would be ejected.

The functionality of the search bar is a little strange, currently we don't have a true onChange handler for the input component, so it only resubmits upon blur or hitting enter.

Gifs:
|desktop|mobile|
|---|---|
|desktop|mobile|


shogunpurple (commented):
Code looks good - although I think we should hold this one back from the release. The reasoning is that I think this would actually be better implemented using a form - where you have a selection of cards (searchable) and when you click a card, it opens the row inside a form on the right hand panel. 

This way, we can reuse the form block, and you can edit/save the fields of the element that you have selected. It also makes it way easier to eject. You get all the benefits of being able to view the data, but also to edit it as well. I think this block will be much more useful in this format, and probably easier to implement as well.

shogunpurple:
@Ghrehh pulled this down to have a play around - looking great, love the mobile styles as well. This component is awesome. 

Just one thing that seems to be missing - after editing the form, I can't save the record that I'm currently editing. Not sure if I'm doing something wrong but can't see anything in the code to allow you to save the record.

shogunpurple:
Ahh - I see what's wrong. The `showSaveButton` prop hasn't been passed to the formblock inside the row explorer. When I turn that prop on everything works as expected. Pushed that update to the branch 👍 
Screenshot 2022-11-08 at 01 00 42

Ghrehh:
@shogunpurple Ah nice one, oversight on my part. Thanks for pushing up the change.

shogunpurple (approved):
LGTM! Great work

deanhannigan:
Hey Gerard, the component came together really well, nice job! Just some minor issues to look into:

- The background and border properties in the custom CSS are replacing the global defaults for the spectrum themes.



Removing these styles from your containers will ensure the global theme values are not overridden 
  ```
  background-color: white;
  border: 1px solid rgb(225, 225, 225);
  ```

- The variable `detailsHeading` isn't currently used anywhere

- The form currently doesn't display in the builder as a result of the custom CSS in place. It would be nice to see the structure of the form in the builder as it being configured. The builder has a store variable that can be used to determine any builder specific behaviour `$builderStore.inBuilder`


- Is the `Select a row to view its fields` container necessary? It displays in the builder but doesn't seem to be used in the app itself.

Any issues, let me know!

Ghrehh:
Me and @deanhannigan talked about adding a binding that could be used by components to see whether they were currently in the builder or not. This could be used to always show the non-zero-state of the Row Explorer as seeing the form is more useful to the user than just seeing the zero-state. Users could also use this in their custom components to do something similar perhaps.

I did some poking around but I think I need someone with more domain knowledge to show me how to add a binding, I could figure out where the binding definitions are provided, but I'm not sure where the actual data stems from. Perhaps this is better left as a follow up PR anyway. What do you think Dean? I addressed the other points you raised though.

deanhannigan (approved):
LGTM! Yes, as it stands, I think the component is good to go. The display of the form in the builder would be nice but I wouldn't consider it a blocker.

aptkingston:
Will give this a review myself this morning! Looks great.

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -0,0 +1,247 @@
...
+
+  let listDataProviderId
+  let listRepeaterId
+</script>
+
+<Block>
+  <BlockComponent
+    type="container"
+    props={{
+      direction: "row",
```
aptkingston, line 29: I think it would look better with some spacing between the card list and the form block, so maybe we could add `gap: "M"` here.

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -0,0 +1,247 @@
...
+    }}
+    styles={{
+      custom: `
+        height: ${height} !important;
```
aptkingston, line 33: I was wondering why height was implemented as a setting rather than a style, but then realised it's because you'd have to pass the style down to the correct component and the other blocks don't actually do this either. Let's leave this as is for now, but I'll do an update at some stage to make blocks actually respect styles applied to them.

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -0,0 +1,247 @@
...
+      type="dataprovider"
+      order={0}
+      bind:id={listDataProviderId}
+      props={{
+        dataSource,
+        paginate: true,
+        limit: 10,
+        filter: [
+          {
+            id: generate(),
```
aptkingston, line 47: We can just hard code this to something (e.g. `0` or `1`) since these only need to be unique within the set of filters. By using a randomly generated ID here we're changing the filter prop on every update, meaning we're going to do a new search request on every render of the block which is unecessary.

Some other blocks actually leave out the `id` property altogether (probably an oversight) but thankfully they still work post ejection as they only contain a single filter and therefore an `undefined` ID is still unique! But we should still update the other blocks to also contain IDs at some stage (again not for now, just thought this was worth mentioning while I thought of it).

The same applies to all the usages of IDs inside filters and button actions here - we can update them all to just be hard coded and reduce prop changes.

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -0,0 +1,247 @@
...
+            field: cardSearchField,
+            operator: "fuzzy",
+            type: "string",
+            value: `{{ [state].[${stateKey}-search] }}`,
```
aptkingston, line 51: While it's usually fine to manually escape HBS expressions by just wrapping them in square brackets, it's better to always use the makePropSafe (imported here as safe) util to be sure.

I think it would be good to update all references of manual escaping to use that util. I'll not add comments for every usage since there are loads - but I think it's best to get them all updates.

It also means we can upate that util in future and know that all our bindings will be correctly escaped still.

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -0,0 +1,247 @@
...
+              },
+              "##eventHandlerType": "Update State",
+              id: generate(),
+            },
+          ],
+        }}
+        styles={{
+          custom: `
+            align-self: flex-end;
+            margin-bottom: 5px;
```
aptkingston, line 226: I think we need a little more spacing here since it looks a bit cramped on mobile - maybe 16px?

As an aside - I'll be updating the form block soon to allow executing actions when you press buttons, which will actually help this block a lot too. It means we could automatically close the form on mobile when saving changes.

packages/client/manifest.json
```diff
@@ -5186,5 +5186,88 @@
...
+      },
+      {
+        "section": true,
+        "name": "Details",
+        "settings": [
+          {
+            "type": "text",
+            "key": "detailTitle",
+            "label": "Title",
+            "nested": true
```
aptkingston, line 5257: We need to remove the `nested` property here - the `schema` context is provided by the repeater, which is on the left side and is not a parent of the title, so attempting to use any nested bindings here will fail and evaluate to empty strings.

packages/client/manifest.json
```diff
@@ -5186,5 +5186,88 @@
         "suffix": "repeater"
       }
     ]
+  },
+  "rowexplorer": {
+    "block": true,
+    "name": "Row explorer block",
```
aptkingston, line 5192: Could we use camel case here for the block name? All components use camel case, but there's inconsistency among the blocks. I think they should all be camel case. (Feel free to quickly update the other block titles to be camel case too if you want to, or we can do it separately).

packages/client/manifest.json
```diff
@@ -5186,5 +5186,88 @@
...
+    "icon": "PersonalizationField",
+    "size": {
+      "width": 100,
```
aptkingston, line 5195: We should increase the size here, as this block is usually pretty large. This key is used to render the size of the placeholder when dragging the component from the "new component" panel. I think we should use 600x400, same as the table block.

Also noticed chart block uses 100x100 and it should also use something bigger, so feel free to quickly update that while you're at it!

packages/client/manifest.json
```diff
@@ -5186,5 +5186,88 @@
...
+      {
+        "type": "dataSource",
+        "label": "Data",
+        "key": "dataSource",
+        "required": true
+      },
+      {
+        "type": "text",
+        "label": "Height",
+        "key": "height"
```
aptkingston, line 5208: I think we should use a default value here, as otherwise the block is massive as soon as you add data. `426px` seems to perfectly fit 3 cards for me.

packages/client/manifest.json
```diff
@@ -5186,5 +5186,88 @@
...
+    "block": true,
+    "name": "Row explorer block",
+    "icon": "PersonalizationField",
+    "size": {
+      "width": 100,
+      "height": 100
+    },
+    "settings": [
+      {
+        "type": "dataSource",
```
aptkingston, line 5200: We should update this to be of type "table" (and name it "Table"), since this will only work with tables.

aptkingston (commented):
Awesome work @Ghrehh! This was a tricky block to make and actually includes a number of "firsts"! First usage of nested blocks (very glad to see this worked out of the box), and first usage of using bindings inside custom styles to make a block responsive even after ejection. Some really clever ideas here.

Had a good play around with this and there were no major issues, just some minor points which I've commented on.

Great job.

packages/client/src/components/app/blocks/RowExplorer.svelte
Ghrehh: I wrapped everything including things that weren't variables in safe, not sure if that's correct or not.

packages/client/manifest.json
Ghrehh: Ah good catch, the title now has no access to bindings unfortunately, I don't know if there's a way to give it access to the context within the form block.

packages/client/manifest.json
Ghrehh: Do you literally just mean the name? like "Row Explorer block"? That's what I did, let me know if that's not what you meant 👍

Ghrehh:
@aptkingston Thanks for the in-depth review, I pushed up the changes you requested. Could you give this another look when you get a chance please?

packages/client/manifest.json
aptkingston: We can technically do this by adding another context of type schema, but we'll need to create a weird double suffix because the form block itself needs a specific context key, and then the form block also already adds another layer of suffix to the repeater inside it... And the bigger issue here anyway is that if we add another context entry in the manifest that is also of type `schema` then we will end up generating duplicate bindings that look identical for all those schema fields, and it will be impossible to tell in the frontend which pertains to the list on the left, and which pertains to the form block on the right.

TLDR is I think we just have to accept that we can't use nested here and we can't allow this field to access the schema bindings.

aptkingston (approved):

Issue #8469, opened 2022-11-01
Filter by access is broken in the apps portal


MihailHadzhiev2022:
If you navigate to apps portal and click on filter by access, the filter is broken. 
<img width="1510" alt="Screenshot 2022-11-01 at 17 16 21" src="https://user-images.githubusercontent.com/102024164/199268531-cf826aff-08f7-473b-b33c-49b0249c706a.png">

PR #8482, opened 2022-11-02
General fixes to app list group filtering

deanhannigan:
## Description
Updates to the access filter picker to restore filtering apps by assigned user groups.

Addresses: 
- https://github.com/Budibase/budibase/issues/8469


Ghrehh (approved):
LGTM

Issue #8369, opened 2022-10-23
SQL Server Handlebars are now Global?


kerbersystems:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.0.33
  - App Version: 2.0.33

**Describe the bug**
When handlebars are created in a na individual custom query, it will be populated to every other custom query under the same SQL Server.

kerbersystems:
It looks like a page refresh brings the right bindings. Could this be a browser issue?

gmentzik:
Have same behavior with Mysql. It started happening in on of most recent updates. Binding part does not refresh when you select another query. You have to refresh the whole page.

skoch-inform:
I can confirm this also for the dynamob data configuration.

melohagan:
Hey @kerbersystems 

Thanks for drawing attention to potential bug, but could you please fill out the bug template. 
Provide detailed reproduction steps and screenshots please!

kerbersystems:
@melohagan 

**Hosting**
Self
Method: docker compose
Budibase Version: 2.0.33
App Version: 2.0.33

**Describe the bug**
When handlebars are created in a na individual custom query, it will be populated to every other custom query under the same SQL Server.

**To Reproduce**
Steps to reproduce the behavior:
1. Go to a specific APP. (it effects all of my APPS)
2. Click on 'Data' > 'SQL Server'
3. Which ever Custom Query is clicked on first, the 'Bindings' section will follow to every other custom query page. 
4. Binding are corrected after a refresh of the page

**Expected behavior**
On each query the bindings are correct for that query.

**Screenshots**

-First query clicked on will set the bindings.
![image](https://user-images.githubusercontent.com/96125948/199340018-826a3afb-406e-44fb-b42f-dd194658d55b.png)

-Every other query after will have the same bindings showing
![image](https://user-images.githubusercontent.com/96125948/199340203-9acb6387-8adc-4e33-b8a1-ac7642a9032a.png)

-Upon refresh, bindings are correct
![image](https://user-images.githubusercontent.com/96125948/199340279-3bb63902-ba16-4ab9-8439-e080f2413e62.png)

melohagan:
Bug introduced in **v1.3.20**

PR #8486, opened 2022-11-02
Fix query viewer bindings

melohagan:
## Description
Bindings were not updating when switching between datasource queries.
Added a **key** on the *query.parameters* to refresh the binding builder when the query parameters change.

Addresses: 
- https://github.com/Budibase/budibase/issues/8369

## Screenshots
**Fixed**
query_viewer_binding

### DIFF
--- a/packages/builder/src/components/integration/QueryViewerBindingBuilder.svelte
+++ b/packages/builder/src/components/integration/QueryViewerBindingBuilder.svelte
@@ -2,7 +2,6 @@
   import { Body, Button, Heading, Layout } from "@budibase/bbui"
   import KeyValueBuilder from "components/integration/KeyValueBuilder.svelte"
   import { getUserBindings } from "builderStore/dataBinding"
-  import { createEventDispatcher } from "svelte"
   export let bindable = true
   export let queryBindings = []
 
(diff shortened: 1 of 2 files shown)


packages/builder/src/components/integration/QueryViewerBindingBuilder.svelte
```diff
@@ -45,12 +47,15 @@
       bindings={[...userBindings]}
       bindingDrawerLeft="260px"
       on:change={e => {
-        queryBindings = e.detail.map(binding => {
-          return {
-            name: binding.name,
-            default: binding.value,
-          }
-        })
+        dispatch(
```
deanhannigan, line 50: NAB: Here you can simplify the change event handling here by replacing it with just `on:change`. Then you can push your update behaviour up to the `QueryViewer` on:change event.

deanhannigan (approved):
LGTM!

packages/builder/src/components/integration/QueryViewerBindingBuilder.svelte
melohagan: Nice suggestion - that's cleaner now!

Issue #8468, opened 2022-11-01
Drag and drop, after ejected a form block, I'm having issues to drag and drop components (save button) 


MihailHadzhiev2022:
https://www.loom.com/share/e863ad3ffdda41efbd56639a8d90ca55

1. Create an app with 1 column and 1 row, Names
2. Navigate to design and create screens
3. Click on names/new/row
4. Eject the table block
5. Try to drag and drop "Save button" 

App froze and error message in the console is displayed

PR #8489, opened 2022-11-02
Fix Form Block Create Mode Bindings Causing Error When Ejected

Ghrehh:
Looking this [issue](https://github.com/Budibase/budibase/issues/8468). Bit of a weird one, actually nothing to do with DnD I think:

[Line throwing the error](https://github.com/Budibase/budibase/blob/develop/packages/builder/src/builderStore/dataBinding.js#L1006)

Moving or highlighting the repeater of an ejected create mode form block causes the page to crash.

I believe some component only visible to the user when the component is ejected, possibly the provider selector of the repeater, is trying to parse the object passed to the repeater as a binding; this is causing an error since it's not a string. 

The block works fine when not ejected because this field isn't visible, and the component itself accepts objects just fine.

Changing the form block to pass a binding string instead of an object to the repeater fixes the issue, my knowledge of handlebars isn't great so there might well be a much cleaner way to handle this, but for the life of me I couldn't find a way to pass an object literal 🤷 

Addresses:

https://github.com/Budibase/budibase/issues/8493
https://github.com/Budibase/budibase/issues/8468

### DIFF
--- a/packages/client/src/components/app/blocks/FormBlock.svelte
+++ b/packages/client/src/components/app/blocks/FormBlock.svelte
@@ -85,13 +85,8 @@
       valueType: "Binding",
     },
   ]
-  // If we're using an "update" form, use the real data provider. If we're
-  // using a create form, we just want a fake array so that our repeater
-  // will actually render the form, but data doesn't matter.
-  $: dataProvider =
-    actionType !== "Create"
-      ? `{{ literal ${safe(providerId)} }}`
-      : { rows: [{}] }
+
+  $: dataProvider = `{{ literal ${safe(providerId)} }}`
   $: renderDeleteButton = showDeleteButton && actionType === "Update"
   $: renderSaveButton = showSaveButton && actionType !== "View"
   $: renderButtons = renderDeleteButton || renderSaveButton


deanhannigan (commented):
You can return the default state if you add the following to [this line](https://github.com/Budibase/budibase/blob/876ef762b8fda75fec89b741f9f56047524e968c/packages/builder/src/builderStore/dataBinding.js#L239) in `dataBinding.js`. It'll ensure that the datasource state is returned as an object:

```
if (settingValue && Array.isArray(settingValue.rows)) {
  return settingValue
}
```
The modification you made does work, as it always resolves `providerId` to `undefined`. Eventually the provider will be resolved in `getProviderContextBindings`.

I tested the above and it appears to behave as expected for FormBlocks with the `create` action type.

melohagan:
@Ghrehh Thanks for having a look at this one! I didn't realise you had it covered when I raised the issue: https://github.com/Budibase/budibase/issues/8493

Make sure it gets merged into master branch!

Ghrehh:

Ghrehh:
Myself an @deanhannigan talked through this, and it seems like my original solution was possibly working accidentally, and we're not really too sure of Cheeks' original intention with the ternary operator we're working on. 

Just always passing the data provider key seems to work for now and we'll ask Cheeks what his intentions were when he gets back from PTO.

deanhannigan (approved):
LGTM

aptkingston:
The reason for the original implementation is that a create form does not fetch a row - so if the real data provider value is provided to the repeater then the form will never render as there are no rows to iterate over. 

I can see how this would break after ejection though as this is not a string value. It would need updated to be HBS or JS binding that evaluates to this fake single row provider. 

However I'm assuming that this fix works and still renders the form - so I'm a bit puzzled. The repeater for a create form should now be iterating over an empty provider... 

I don't have the code available to me at the moment so can't debug this but that's just an explanation for the original implementation.

Issue #8491, opened 2022-11-02
Autogenerated screens - navigate back to table after deleting a row 


AndyWhann:
##Problem

When I create autogenerated screens and delete a row, I am left looking at a blank screen rather than being navigated back to the table.

##Solution 

Add navigate to action on the delete option

PR #8506, opened 2022-11-03
Fix for auto-generated screens - buttons navigating correctly

mike12345567:
## Description
Fix for #8491 - auto generated screens use the form block now which had an issue with the close screen event not providing a URL.

Addresses: 
- #8491


shogunpurple (approved):
LGTM!

Issue #8478, opened 2022-11-01
Cannot add any components to a screen


kerbersystems:
After updating to 2.0.37 I can no longer add, delete, or move components on a few of my screens. The error I get is "You can't place a Section here". I am not trying to place a section.

mike12345567:
Hi @kerbersystems - this sounds a bit like #8484 - would you be able to provide an app export exhibiting this behaviour?

kerbersystems:
Hi @mike12345567, I'm new to the scene, how do I create an export for you? I'm using compose.

kerbersystems:
Still persistent in 2.1.8

mike12345567:
Hi @kerbersystems - you can export an app from Budibase by going to `Manage` your app from the portal, then selecting export latest from the three dot menu beside the edit button like so:
![image](https://user-images.githubusercontent.com/4407001/199575855-f2ff2c81-8544-4fc9-ab88-c08e984b0850.png)

kerbersystems:
[NIMO-export-1667425964980.tar.gz](https://github.com/Budibase/budibase/files/9924374/NIMO-export-1667425964980.tar.gz)

mike12345567:
Thanks @kerbersystems - I am able to replicate this now!

sergey-vin:
When this will be released?

mike12345567:
Hi @sergey-vin - we are currently running the release for this - with it being resolved in version 2.1.13.

sergey-vin:
let me test on cloud

sergey-vin:
hm. the latest version available on cloud is 2.1.11. can't see 2.1.13

mike12345567:
@sergey-vin - it might take a little while for all instances to fully update. I can see it running on some of the nodes.

sergey-vin:
I can confirm the fix is working! Thanks!

kerbersystems:
I can also confirm this is working in my environment. 

Version 2.1.13

mike12345567:
Going to close this out as it seems to be resolved!

PR #8508, opened 2022-11-03
Disable screen validation

deanhannigan:
## Description

Temporarily disabling screen validation. Current approach locks users with invalid screen configuration to alter their screens.

Addresses: 
- https://github.com/Budibase/budibase/issues/8478


mike12345567 (approved):
LGTM!

Issue #7977, opened 2022-09-26
Configure columns - conditional background and text colour


melohagan:
**Describe the feature request**
As a user, I want to set the colour of the cell background / text based on conditional UI.

**Screenshots**
![Screenshot 2022-09-26 at 16 39 43](https://user-images.githubusercontent.com/101575380/192320389-ce343099-24a2-44f2-8d17-3cc1c4d32c2e.png)

At the moment the change affects all rows, not a specific row. Add a toggle, e.g. `apply to all` or `apply to matching`
![Screenshot 2022-09-26 at 16 43 51](https://user-images.githubusercontent.com/101575380/192321380-52840871-ed15-4b37-9f29-a430579d81d0.png)

PR #8512, opened 2022-11-03
Conditional table cell - background and text color

melohagan:
## Description
Initially the idea was to have this as an option in the **Configure columns** drawer, however this was very cluttered, so it was moved into its own table setting.

You can now set the background and text colours of table cells conditionally, and you can also select *Row* to apply the change to the entire row if the column condition is met.

Addresses: 
- https://github.com/Budibase/budibase/issues/7977

### Dependencies
I was having issues with circular dependencies between the `helpers.js` file in **bbui** and the `LuceneUtils` of **frontend-core**. 
To resolve this, I swapped the dependencies in the `package.json` files.
Originally **frontend-core** was *dependent* on **bbui**.
Now **bbui** is *dependent* on **frontend-core**. 

I'm not sure if this makes sense, I just did it to see if I could get the build to work. Maybe this is massive overkill / is the wrong dependency setup. With that said I think **frontend-code** is a lot smaller than **bbui** (roughly 10% the size?).
So in that sense it seems like a good thing to swap a dependency on **bbui** with a dependency on **frontend-core** instead? 

## App Export
[Cell conditional-export-1667494995685.tar.gz](https://github.com/Budibase/budibase/files/9931436/Cell.conditional-export-1667494995685.tar.gz)

## Screenshots
Screenshot 2022-11-03 at 17 02 58

Screenshot 2022-11-03 at 17 04 18
*It's a bit confusing to look at, but that colour scheme is not a likely use-case obviously!*


melohagan:
Closing in favour of the **Super Table** custom component: https://github.com/Budibase/budibase/discussions/10204

Issue #8526, opened 2022-11-04
Migration to 2.1 failed because of multi-select picker error


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/8521

<div type='discussions-op-text'>

<sup>Originally posted by **dannyshin76** November  4, 2022</sup>
I exported my app from Budibase 1.3.15 (local deployment) and imported it into 2.1.14 (cloud version), and my app started crashing with this browser console error:
![image](https://user-images.githubusercontent.com/93232582/199976243-f8bf6978-9870-45ee-b35e-9bf5c8e54114.png)

After some investigation, I was able to pinpoint the problem to one of my multi-select picker with a default value that uses Javascript:
![image](https://user-images.githubusercontent.com/93232582/199976582-5f65a6d5-bf53-4418-9da5-59bf1c11ad97.png)

Options source is "Custom", and it crashes no matter if I define options or not.
The very same picker works perfectly fine with BB 1.3.15. Could anyone help me out?

</div>

melohagan:
@dannyshin76 This is now fixed in **v2.1.15**

PR #8527, opened 2022-11-04
Check schema exists for custom options multi picker

melohagan:
## Description
Null pointer was caused when the user selected **Custom** as the options source in a multi-picker.
Fixed! 

Addresses: 
- https://github.com/Budibase/budibase/issues/8526


deanhannigan (approved):
LGTM! Nicely spotted!

Issue #8594, opened 2022-11-08
Table sorting breaks when column has special characters


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.15
  - App Version: Version 2.1.15

**Describe the bug**
Thanks to Joe Moss for finding this one!
Table sorting breaks when the column name has a special character in it, e.g. `% : / ` 

**To Reproduce**
1. Create an internal Budibase DB table with a column with forward slash (or other character `% :`
2. Click on the column heading to sort.
3. The rows will disappear depending on the sort order
4. Rename all columns to not have special characters in them
5. The sorting will now be fine

**Expected behavior**
I should be able to include special characters in my internal DB columns without it breaking sorting.

**Screenshots**
![Screenshot 2022-11-08 at 15 20 48](https://user-images.githubusercontent.com/101575380/200604093-37667d87-b3c2-4060-82af-301d495787b6.png)

![Screenshot 2022-11-08 at 15 21 06](https://user-images.githubusercontent.com/101575380/200604159-1087bc60-6124-4a83-bc6f-8f1097d948cd.png)
  
**App Export**
[Procurement_Approval_2-export-1667919085704.tar.gz](https://github.com/Budibase/budibase/files/9962570/Procurement_Approval_2-export-1667919085704.tar.gz)

PR #8597, opened 2022-11-08
Prevent special characters being used in new internal DB column

melohagan:
## Description
Joe created a row with a forward slash in it, which buggered the table rows upon sorting.
The couchdb index could be updated to handle this, however it is easier to prevent new internal DB columns being created with illegal lucene characters.

Addresses: 
- https://github.com/Budibase/budibase/issues/8594
- https://github.com/Budibase/budibase/issues/8621

## Screenshots
Screenshot 2022-11-08 at 17 16 48

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -304,7 +304,7 @@
     const newError = {}
     if (!external && fieldInfo.name?.startsWith("_")) {
       newError.name = `Column name cannot start with an underscore.`
-    } else if (fieldInfo.name?.match(/[-!*+?^"{}()~/[\]\\]/g)) {
+    } else if (fieldInfo.name?.match(/[-!*+?:^"{}()~/[\]\\]/g)) {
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):
LGTM! It might also be worth adding some validation server-side in the table save controller - as this is actually quite important and will become more important as we move forward with future plans for internal tables.

melohagan:
@mike12345567 I had a go at adding some validation to the **save** method like so:

```js
for (const columnName in table.schema) {
    if (!columnName.match(/^[a-zA-Z0-9\s]*$/g)) {
      ctx.throw(400, "Illegal character in column name; must be alpha-numeric.")
    }
  }
```

The issue I found is that if someone already has multiple columns with illegal characters, then they wont be able to fix them.
For example there are two invalid fields in this table:
Screenshot 2022-11-09 at 20 52 39

*Remove the illegal character and save*
Screenshot 2022-11-09 at 20 53 25

*Error thrown by the other column that has an illegal character, thus it's impossible to fix the table*
Screenshot 2022-11-09 at 20 53 52

mike12345567:
Ah - yes I can see how that would be an issue - we'll need to come up with a migration in future to deal with illegal characters and remove them from tables - for now client-side is as good as we're gonna get I suppose!

Issue #8589, opened 2022-11-08
Export app - 504 Gateway timeout


maiducan:
Hello,

i cannot export my app from budibase portal (self-hosted), my app is large (more than 50MB) and it exceeds the timeout during export even though i increased the limit in nginx. So it must be somewhere inside of the budibase. Its always 2 minutes.

maiducan:
@mike12345567 Hello Mike, thank you for your fast reaction. However it is still not doing the trick for me. I have updated to the newest version and tried to export, and there is still 2 minutes threshold. :/ Could you please look at it once again? Thank you alot!

marian001:
Although different, it may be related..   I can't export even small app from self-hosted portal. I'm getting error 502 Bad gateway.

maiducan:
yes its related .. i have the same issue now, it immediately through 502 after clicking export app  :/ .. @mike12345567

maiducan:
@mike12345567 

1. I am using docker-compose / Budi CLI
2. I am udpating via budibase UI

mike12345567:
Hi @maiducan - updating through the UI can be quite flaky - its something we're probably going to remove in the future as it doesn't work particularly well. Can you please try updating using the Budibase CLI, using the `budi hosting --update` command.

mike12345567:
Hi @maiducan - do you know what version this started happening at? I'm currently testing an installation based on docker-compose, running 2.1.24 and I cannot replicate it at all.

mike12345567:
Hi @marian001 - I'm not 100% sure what image is used on Digital Ocean at the minute, but if it is docker single image, or docker-compose based you can still use the CLI to update it.

maiducan:
For me it started at version 2.1.21 or 24, after you made changes in those nginx config files

marian001:
BTW, I'm running my instance with HTTPS and on own domain. I tried accessing the portal through HTTP on the IP address and port 10000, but the export is still failing.

mike12345567:
Hi guys - I've played around with this, tried upgrading through a few versions etc - but I'm still unable to replicate. Perhaps there is something you can find in your logs that would help explain the issue - or if you can see if it happens with all apps, vs a small set of apps (does it happen with newly created apps, or only with older apps, or is it just a total installation problem).

maiducan:
I have two separate servers, each one has unique app (very big one > 50mb, and the second is tiny) .. And this occurs on both servers.

marian001:
[...]

PR #8599, opened 2022-11-08
Fix for groups access

mike12345567:
## Description
Typescript re-write of the roles layer, this is the backbone of a lot of our security features, and I believe the issue was generally to do with a lack of handling of null-ish inputs.

This was discussed in discussion #8578.

Also includes a fix for #8589 - removing the 2 minute timeout for exporting apps.

I found this going through the logs and noticed:
image


shogunpurple (approved):
LGTM! Good job - TS would definitely have saved us there 😄

Issue #8523, opened 2022-11-04
Save Query button: Show why it is disabled


AlejandroFNadal:
**Describe the feature request**
When creating a new query, if the name is not written, the Save Button is disabled. This makes sense, but perhaps it would be better if the button was enabled but a little red message was shown saying "The query has no name". 
Perhaps unnecessary, but it has happened to me that I got frustrated because I could not save the query and I didn't know why until I started checking everything and saw that the name field was empty. 

![image](https://user-images.githubusercontent.com/32724827/200532033-123f4688-bcd1-42e2-b2e9-c40c8b7bb48b.png)

MistaZidane:
@AlejandroFNadal  Please can you send a screenshot so I could figure out the button in question? Thanks

AlejandroFNadal:
> @AlejandroFNadal Please can you send a screenshot so I could figure out the button in question? Thanks

@MistaZidane thanks for the reply, I edited the issue report adding a screenshot

MistaZidane:
@AlejandroFNadal  I have got it working now. There is a problem when the function is set read. 
[https://github.com/Budibase/budibase/pull/8602](https://github.com/Budibase/budibase/pull/8602)

AlejandroFNadal:
Great, thanks :-D !

aptkingston:
This will be addressed as part of upcoming work on the data section.

PR #8602, opened 2022-11-08
Bug: Fixed save query button not showing when function is equal read

MistaZidane:
## Description
When creating a new query, if the name is not written, the Save Button is disabled. This sounds good but when the selected function is read, even if the query name is written, the save query button still appears disabled.

Addresses: 
-  [https://github.com/Budibase/budibase/issues/8523](https://github.com/Budibase/budibase/issues/8523)

## Files touched 
- packages/builder/src/components/integration/QueryViewer.svelte

## Screenshots
Before:
Screen Shot 2022-11-08 at 11 54 15 PM
After: 
Screen Shot 2022-11-08 at 11 54 02 PM

### DIFF
--- a/packages/builder/src/components/integration/QueryViewer.svelte
+++ b/packages/builder/src/components/integration/QueryViewer.svelte
@@ -41,7 +41,7 @@
   $: integrationInfo = datasourceType ? $integrations[datasourceType] : null
   $: queryConfig = integrationInfo?.query
   $: shouldShowQueryConfig = queryConfig && query.queryVerb
-  $: readQuery = query.queryVerb === "read" || query.readable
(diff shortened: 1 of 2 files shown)


MistaZidane:
recheck

MistaZidane:

melohagan:
Hey @MistaZidane 

The reason the `Save Query` button is disabled in the *Before* screenshot is because you haven't **Run Query**. You have to run the query so that the schema is pulled down, otherwise it wont work when reading data into your table.

I think the solution here is to add validation to highlight the query name in red if it has not been filled in, and to prevent it being red by default, maybe provide a default value.

MistaZidane:
hello @melohagan  thanks for the clearance on this. I will add a default query name to it.

Issue #8585, opened 2022-11-08
Ejecting a date picker from a form has the wrong initial 'Show time' setting


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.15
  - App Version: 2.1.15

**Describe the bug**
Ejecting a date picker from a form has the setting 'Show time' ticked, but the time doesn't actually show. You have to untick and tick the setting again to get the time to show.

**To Reproduce**
1. Create a table with a Date/Time field
2. Create auto-generated screens
3. In the edit screen, see that the date-picker has no time
4. Eject the form block
5. The date field has 'Show time' ticked, but see that no time is shown
6. Untick/tick to get the setting to apply.

**Expected behavior**
When I eject a form block, the date picker settings should match what they were in the form block.

**Screenshots**
*Ejected form field has 'Show time' ticked but no time included*
![Screenshot 2022-11-08 at 11 10 03](https://user-images.githubusercontent.com/101575380/200549481-0b2bcbce-3d2e-4a7e-8b1f-0ad7e932b78a.png)

*Toggle the 'Show time' setting. Time is now shown*
![Screenshot 2022-11-08 at 11 10 47](https://user-images.githubusercontent.com/101575380/200549599-4826b200-67a2-4ec9-a330-8bd8a8798e10.png)

aptkingston:
This is actually just a bug with the date time field itself - the component has a default value of false for showing time, but the manifest defines a default value of true. I'll push that change to the form block fixes branch anyway since hopefully it'll get merged soon.

Issue #8577, opened 2022-11-08
When there are no rows, the 'Create' type form block has no fields


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.15
  - App Version: 2.1.15

**Describe the bug**
When creating autogenerated screens, the Create `/new/row` screen will have no rows if the table has no rows. 

**To Reproduce**
1. Add a Budibase DB table without any rows
2. Auto generate screens
3. Got to `/new/row` screen
4. See that there are no form fields to fill out
5. Add a row manually to your table
6. The `/new/row` screen will now show form fields

**Expected behavior**
When my table has no rows, I should see the fields in a form block when it is in 'Create' mode.

**Screenshots**
![Screenshot 2022-11-08 at 09 41 55](https://user-images.githubusercontent.com/101575380/200530323-ab07ab7d-c85a-47f9-8b62-4c89839491b9.png)

![Screenshot 2022-11-08 at 09 42 23](https://user-images.githubusercontent.com/101575380/200530423-c88f1d8c-6696-40f3-82d4-84562ae4bd0f.png)

---

**Now manually add a row to the table**
![Screenshot 2022-11-08 at 09 43 13](https://user-images.githubusercontent.com/101575380/200530681-6a5a3359-8a84-4515-afcd-1ea090a22575.png)

![Screenshot 2022-11-08 at 09 42 58](https://user-images.githubusercontent.com/101575380/200530585-35cea97c-2907-4198-bf9a-6ffa86c961da.png)

chrismb23:
@melohagan I’m experiencing this issue also on the hosted (cloud) version of Budibase.

avivkri:
@chrismb23,

As a temporary workaround for New row screen, you may eject the Form block, keep only the Form intact and remove the Data provider and the Repeater components.

aptkingston:
I'm pretty sure this is because of the hot fix that was released recently - I commented on that issue that I expected that it would prevent create forms from working at all as it passes a reference to the real data provider, which contains 0 rows when using a create form.

Original issue: https://github.com/Budibase/budibase/issues/8493
Hotfix: https://github.com/Budibase/budibase/pull/8489
My input on the fix: https://github.com/Budibase/budibase/pull/8489#issuecomment-1305547508

Rather than revert the hot fix, I'll push a fix to fix both the original issue and this new regression.

Issue #8522, opened 2022-11-04
Ejecting a form block should not include a Data provider and repeater if the type is 'Create'


melohagan:
**Describe the feature request**
Whenever you created auto-generated screens in the past, the `/new/row` screen would add a Form component without a Data provider and Repeater, as creating a new row does not require a datasource.

Now the Form Block is added with type *Create*, however when you eject the block it includes an unnecessary Data provider and Repeater

**Screenshots**

**Before eject**
![Screenshot 2022-11-04 at 12 58 42](https://user-images.githubusercontent.com/101575380/199978123-539b9b56-eae5-40e7-bd58-b04aa2bf4037.png)

**After eject**
![Screenshot 2022-11-04 at 12 57 18](https://user-images.githubusercontent.com/101575380/199977862-5018259f-a310-4a13-9857-044889aa6592.png)

philsonchoi:
I came across this issue today while auto-generating screens. For any "Create" forms, if the database is empty, then the form fields do not load as the repeater cannot find any rows. It would be helpful to remove this repeater from the autogenerated create forms in the future.

aptkingston:
@philsonchoi this has already been fixed and will be released next week :+1:

Issue #8493, opened 2022-11-02
Autogenerated ejected form block crashes app


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
Ejecting a form block and selecting the Form component causes the app to enter a broken state.

**To Reproduce**
1. Create auto-generated screens from a table (either internal or external SQL)
2. Select the `/new/row` screen
3. Eject the form block
4. Click on the form component
5. The app crashes, and the console shows: `TypeError: value?.match is not a function at extractLiteralHandlebarsID`
6. Refresh the page - **the screen is still unworkable**

**Expected behavior**
Ejecting a form block should not crash the app for auto-generated screens.

**Screenshots**
![Screenshot 2022-11-02 at 17 22 38](https://user-images.githubusercontent.com/101575380/199558432-2a34a946-1d09-426d-a680-02a7f69d75ae.png)

![Screenshot 2022-11-02 at 17 23 18](https://user-images.githubusercontent.com/101575380/199558552-16eb3941-0c46-4044-9707-47adb8f3e615.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome

Ghrehh:
Merged this directly into master like you asked Mel https://github.com/Budibase/budibase/pull/8489

melohagan:
That looks fixed now, thanks @Ghrehh ! Closing.

PR #8608, opened 2022-11-09
Fix form block issues

aptkingston:
## Description
This PR fixes the original issue with ejecting "create" mode form blocks, and also fixes a regression as part of the original hotfix. I've also addressed https://github.com/Budibase/budibase/issues/8522 while doing this, so that ejecting a "create" form block will not include the redundant data provider and repeater.

TLDR:
- Split form block into 2 files
- The wrapper of the shared core components changes depending on mode - "update" has a form inside a data provider and repeater, while "create" just has a form inside a container
- This split also enabled trivially solving the original issue which was due to the problem of handling the data provider and repeater for "create" mode forms, so this is 2 birds with 1 stone
- The original issue of apps breaking when ejecting a "create" mode block is also fixed since it will not longer include the data provider/repeater which was causing the issue

Addresses: 
- https://github.com/Budibase/budibase/issues/8577
- https://github.com/Budibase/budibase/issues/8522
- https://github.com/Budibase/budibase/issues/8493
- https://github.com/Budibase/budibase/issues/8643
- https://github.com/Budibase/budibase/issues/8585
- Fixes an issue where unmounted components are still included in ejected blocks
- Fixes an issue where action parameters can become null, freezing the page

## Screenshots
Ejecting a "create" form:
image

Ejecting an "update" form:
image


packages/client/src/components/app/blocks/form/FormBlock.svelte
```diff
@@ -0,0 +1,101 @@
...
+  export let rowId
+  export let actionUrl
+
+  const { fetchDatasourceSchema } = getContext("sdk")
+
+  let schema
+  let providerId
+  let repeaterId
+
+  $: fetchSchema(dataSource)
```
melohagan, line 26: dataSource appears to be an unused parameter

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
```diff
@@ -0,0 +1,194 @@
...
+  export let fields
+  export let labelPosition
+  export let title
+  export let showSaveButton
+  export let showDeleteButton
+  export let schema
+  export let repeaterId
+
+  const FieldTypeToComponentMap = {
+    string: "stringfield",
```
melohagan, line 20: Having a look at the component map, it seems this type is missing: `"field/sortable": SortableFieldSelect,`
Just wondering if that matters?

melohagan (approved):
Pulled down the branch and the various bugs raised all appear to be fixed to me! Awesome 😄 

LGTM!

NAB: I did notice this tiny bug:
1. Create a data table with no rows
2. Auto generate screens, go to the edit screen `/:id`
3. Eject the form block
4. There is no form component:
Screenshot 2022-11-17 at 09 21 38

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
aptkingston: This is a map for generating form fields from schema - its not the map for component settings which is what you're thinking of! That map is defined in `componentSettings.js`.

packages/client/src/components/app/blocks/form/FormBlock.svelte
aptkingston: It's not a param, but needs to be referenced in the reactive statement so that svelte knows to execute this function again whenever dataSource changes. Making it a param would actually not change the behaviour at all since the function already uses dataSource, which will just reference the latest version from props.

It's a little confusing perhaps, so it may as well be made a parameter for the sake of comprehension.

melohagan: Ah yes I get it, that's good to know!

melohagan:
> The bug you're describing here is just because the form block (before ejection) will have the "fields" setting empty. If there are no fields, it won't render a form. I'm pretty certain if your table has a schema (even if it has no rows) and the form block has a "fields" setting, then there will be a form after ejection. It probably won't be rendered since you don't have a row to edit, but it would be present in the ejected structure.

The table had three columns: Name, Age, DOB (types text, number, date/time) 
All three fields were selected in the fields dropdown before ejection.

aptkingston:
> > The bug you're describing here is just because the form block (before ejection) will have the "fields" setting empty. If there are no fields, it won't render a form. I'm pretty certain if your table has a schema (even if it has no rows) and the form block has a "fields" setting, then there will be a form after ejection. It probably won't be rendered since you don't have a row to edit, but it would be present in the ejected structure.
> 
> The table had three columns: Name, Age, DOB (types text, number, date/time) All three fields were selected in the fields dropdown before ejection.

OK, I'll move back on this branch then and investigate. I feel like we're slightly clutching at straws here and I don't want specific rare edge cases to hold up these important fixes for long, but I'll have another look.

melohagan:
> OK, I'll move back on this branch then and investigate. I feel like we're slightly clutching at straws here and I don't want specific rare edge cases to hold up these important fixes for long, but I'll have another look.

Like I said NAB, if you wanted to merge this one in first it's probably worth doing that because there's a lot of good fixes in there already.

aptkingston:
> > OK, I'll move back on this branch then and investigate. I feel like we're slightly clutching at straws here and I don't want specific rare edge cases to hold up these important fixes for long, but I'll have another look.
> 
> Like I said NAB, if you wanted to merge this one in first it's probably worth doing that because there's a lot of good fixes in there already.

My bad - didn't clock the NAB. Yea it's because without any data in the table, the repeater has nothing to iterate over. That means that a form component never actually gets rendered inside the block, so one doesn't exist after ejection either.

The only way I can think of changing this (which will affect other things, but maybe for the better) would be to update the dataprovider so that inside the builder, it will always render at least one row. Basically if no data is found, just use an empty object as a fake row so that anything using the data provider will still render. That would prevent empty repeaters etc, inside the builder only.

Probably worth trying that at some stage but leaving it out of this branch to avoid other implications.

Issue #8646, opened 2022-11-11
Logout can redirect to account portal instead of login page 


Rory-Powell:
**Hosting**
- Cloud

**Describe the bug**
A race condition in the logout flow can result on the user being stuck on the account portal instead of the login screen

**To Reproduce**
See: https://cdn.discordapp.com/attachments/735083509202419778/1040567831017816135/sign_out_broken.gif

**Expected behavior**
Logout should always redirect correctly to login page

PR #8647, opened 2022-11-11
Fix logout race condition

Rory-Powell:
## Description
Fix logout race condition

Addresses: 
- https://github.com/Budibase/budibase/issues/8646


shogunpurple (approved):
LGTM!

Issue #7762, opened 2022-09-14
"Prompt User" Action


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7741

<div type='discussions-op-text'>

<sup>Originally posted by **poirazis** September 12, 2022</sup>
Scenario : 
I have a button that does something. 
I would like to prompt the user, before continuing. Similarly to the Delete Row or Execute Query.
Maybe there should be an Action "Prompt User" which can then be followed by a Continue If / Stop If.
The "Prompt User" action, should have a bindable "message" property.

Example : 
Button -> OnClick
  -> Action "Prompt User". Message "Are you sure you want to add all {{ State.totalIngedients }} ingredients to the Order ?"
  -> Action "Continue If". Continue If Step 1.Response === "OK" 
  -> Action "Execute Query". ( Would be nice to be able to disable / customize default notifications for success or failure )
  -> Action "Notification" -> "Ingredients Added" ( Or "Operation Failed. Ingredients were not added to the order" )

Thoughts ?</div>

PR #8676, opened 2022-11-14
Button action for user prompt

FlaminWrap:
## Description
Added Prompt User action for buttons which acts as a dialog prompting the user to confirm or cancel before continuing.

Addresses: 
#7762

## App Export
[PromptUserApp-export-1668389698129.tar.gz](https://github.com/Budibase/budibase/files/9998824/PromptUserApp-export-1668389698129.tar.gz)

## Screenshots
Showing in the list of actions
image

After selecting prompt user
image

Custom title and message
image

Video
Link to video


melohagan (approved):
LGTM! Thanks for your contribution!

packages/client/src/utils/buttonActions.js
```diff
@@ -429,12 +440,14 @@ export const enrichButtonActions = (actions, context) => {
...
                   const newContext = { ...context, actions: buttonContext }
 
-                  // Enrich and call the next button action
-                  const next = enrichButtonActions(
-                    actions.slice(i + 1),
-                    newContext
-                  )
-                  resolve(await next())
+                  // Enrich and call the next button action if there is more than one action remaining
+                  if (actions.length > 1) {
```
aptkingston, line 444: We don't actually need to add this check - `enrichButtonActions` won't do anyting when an empty array is passed in.

packages/client/src/utils/buttonActions.js
```diff
@@ -327,6 +327,11 @@ const showNotificationHandler = action => {
   notificationStore.actions[type]?.(message, autoDismiss)
 }
 
+const promptUserHandler = action => {
```
aptkingston, line 330: We can just make this a completely empty function:
`const promptUserHandler = () => {}`

Doesn't affect functionality of course but may as well keep it as tidy as possible!

aptkingston (approved):
Thanks for the contribution @FlaminWrap! It's clever as well to just hook in to the existing confirmation prompt flow to handle this - keeps the change super tiny and safe.

Couple of suggestions just to remove redundant code, but functionality-wise it's perfect!

PR #8686, opened 2022-11-15
Fix for app export 502 issue

mike12345567:
## Description
Updating proxy configuration for docker-compose, k8s and single image - correctly setting up a route in the proxy which will work on all platforms. Previously this only worked on some linux hosts due to the way the route was specified. Also updated the single image to allow exporting larger apps as well.

Addresses: 
- #8589

### DIFF
--- a/hosting/nginx.prod.conf.hbs
+++ b/hosting/nginx.prod.conf.hbs
@@ -117,12 +117,21 @@ http {
     }
 
     location /api/backups/ {
+      # calls to export apps are limited
+      limit_req zone=ratelimit burst=20 nodelay;
+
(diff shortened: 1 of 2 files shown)


hosting/nginx.prod.conf.hbs
```diff
@@ -117,12 +117,21 @@ http {
     }
 
     location /api/backups/ {
+      # calls to export apps are limited
```
Rory-Powell, line 120: Could just update this to be more general, as it affects all backup endpoints and not just the export

Rory-Powell (approved):

Issue #8309, opened 2022-10-18
Default data providers to the first datasource available


shogunpurple:
When I'm using a data provider, I want the first datasource available to be the default data source so that I can see the data on the screen straight away.

**COS**
- Default all data providers to the first element in the data sources array when added to the screen
- Hierarchy is as follows - test DB, internal table (not users table), then SQL/data source plus table, then custom

Issue #8308, opened 2022-10-18
Add Budibase Test DB


shogunpurple:
When I create a new app with budibase, I want to have a test datasource already configured so that I can start building straight away.

**COS**
- Ship budibase test DB as a separate datasource
- Use a simplified version of the northwind database or another simple dataset with 4 or more tables and at least 1 relationship
- Allow deletion of the test database if required

PR #8695, opened 2022-11-15
Feature/auto select dataprovider source

deanhannigan:
## Description

Data providers will now automatically select a datasource when added to a screen for the first time. 

The order of preference for datasource auto-selection is as follows:

1. Test DB (https://github.com/Budibase/budibase/issues/8308)
2. Budibase DB (Excludes the Users table)
3. External Source. (MySQL, Postgres etc.)

When the component instance is created, the first table in the selected datasource will appear in the provider.

Addresses: 
- https://github.com/Budibase/budibase/issues/8309

## Screenshots

Components like the Table block will load immediately with the pre-determined datasource.

output


packages/builder/src/builderStore/store/frontend.js
```diff
@@ -481,6 +481,57 @@ export const getFrontendStore = () => {
           return null
         }
 
+        let dataSourceField = definition.settings.find(
```
aptkingston, line 484: For this first half - we actually have a util called `getComponentSettings` in `componentUtils.js` which will get the full list of settings for  component type, including flattening and merging settings inside section. We could use that here to reduce a lot of this code!

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -490,6 +541,15 @@ export const getFrontendStore = () => {
           }
         })
 
+        // Set a default datasource
+        if (dataSourceField && defaultDatasource) {
+          props["dataSource"] = {
```
aptkingston, line 546: We can't assume the name of the setting is `dataSource` here - we should reference the `key` property of the datasource/table setting found earlier.

aptkingston (commented):
Couple of small suggestions to refactor but looks good! Great QoL improvement.

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -481,15 +481,60 @@ export const getFrontendStore = () => {
...
+        let defaultDatasource
+        if (dataSourceField) {
+          const _tables = get(tables)
+          const filteredTables = _tables.list.filter(
+            table => table._id != "ta_users"
+          )
+
+          const internalTable = filteredTables.find(
+            table =>
+              table.sourceId === "bb_internal" && table.type == "internal"
```
shogunpurple, line 500: NAB: Might be nice to use constants here

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -481,15 +481,60 @@ export const getFrontendStore = () => {
...
+          const defaultSourceTable = filteredTables.find(
+            table =>
+              table.sourceId !== "bb_internal" && table.type == "internal"
+          )
+
+          const defaultExternalTable = filteredTables.find(
+            table => table.type == "external"
+          )
+
+          if (defaultSourceTable) {
```
shogunpurple, line 512: Possible simplification:

`const defaultDatasource = defaultSourceTable || internalTable || defaultExternalTable`

shogunpurple (approved):
LGTM! Optional comments.

PR #8704, opened 2022-11-16
PouchDB replacement and context simplification

mike12345567:
## Description
Opening PR for review - comments/explanation will be added shortly!

Pro counterpart: https://github.com/Budibase/budibase-pro/pull/85

Key points:
1. CLS is gone - replaced by just pure `async_hooks` usage
2. PouchDB mostly replaced by the new `PouchLike` class, which utilises couchdb-nano
3. Lots more Typescript

Addresses: 
- #6594


packages/server/src/api/controllers/auth.ts
```diff
@@ -24,7 +25,7 @@ export async function fetchSelf(ctx: any) {
   }
 
   const appId = context.getAppId()
-  const user = await getFullUser(ctx, userId)
+  const user: ContextUser = await getFullUser(ctx, userId)
```
Rory-Powell, line 28: Can we type the `getFullUser` fn to also return `ContextUser` to make sure these line up. 

We should probably better define the forms the user can take, that's a lot of different concepts in play like
- user built by app-service middleware
- user built by worker-service middleware
- user cached in redis
- db user
- app user
- /api/self user
- /api/global/self user

I'd be up for doing a session where we outline the properties used in each case, make sure we've got the correct types in place and plan for where they'll be used.

packages/server/src/db/utils.ts
```diff
@@ -1,90 +1,64 @@
...
-exports.generateRoleID = generateRoleID
-exports.getRoleParams = getRoleParams
-exports.getQueryIndex = getQueryIndex
-exports.getDocParams = getDocParams
-exports.getRowParams = getRowParams
-exports.generateRowID = generateRowID
-exports.getUserMetadataParams = getUserMetadataParams
-exports.generateUserMetadataID = generateUserMetadataID
-exports.getGlobalIDFromUserMetadataID = getGlobalIDFromUserMetadataID
+export const SEPARATOR = dbCore.SEPARATOR
```
Rory-Powell, line 24: The mix of constants and utils in this block can be pretty confusing, especially when you're writing a new constant and deciding where to put it. It's not very concrete if it should go into the service/constants, this utils file, or the backend core. 

Maybe we should deprecate this block, in favour of the core version directly if it exists. Ideally we could namespace out constants in core, like `Constants.DB.X`

packages/server/src/middleware/selfhost.ts
```diff
@@ -1,7 +1,9 @@
-const env = require("../environment")
+import env from "../environment"
+import { BBContext } from "@budibase/types"
+
 // if added as a middleware will stop requests unless builder is in self host mode
 // or cloud is in self host
-module.exports = async (ctx, next) => {
+module.exports = async (ctx: BBContext, next: any) => {
```
Rory-Powell, line 6: I was gonna say to use `export =` here so it can be imported by other TS files. Had a look and I can't find where this is used, just a test

packages/server/src/threads/automation.ts
```diff
@@ -120,7 +120,7 @@ class Orchestrator {
   }
 
   async getMetadata(): Promise<AutomationMetadata> {
-    const metadataId = generateAutomationMetadataID(this._automation._id)
+    const metadataId = generateAutomationMetadataID(this._automation._id!)
```
Rory-Powell, line 123: Not feedback just a thought - The constant usage of `!` on doc ids make me feel like we've done something wrong. The id should always be available, it's only in the cases of initial creation where the doc fields may be undefined. 

The correct thing is probably for us to make the `Document` fields non-optional to start using a different pattern like:

```ts
interface CreateCar {
  ...
}

interface Car extends CreateCar, Document {
  ...
}
```

packages/backend-core/src/types.ts
```diff
@@ -0,0 +1 @@
+export { PouchLike } from "./db"
```
Rory-Powell, line 1: Should this be exported as `typeof PouchLike`?

packages/worker/src/sdk/users/users.ts
```diff
@@ -479,7 +479,7 @@ export const bulkDelete = async (
     (user: RowResponse<User>) => {
       return user.doc
     }
-  )
+  ) as User[]
```
Rory-Powell, line 482: I think this can be removed, due to the const already having the type assigned. Tried locally and it builds ok

Rory-Powell (approved):
Looks great michael, just some very minor comments. This is so much better, easier to read and reason about, and 🤞 will help with some of the leaking problems we have been seeing.

packages/backend-core/src/context/tests/index.spec.js
```diff
@@ -1,18 +1,9 @@
...
-import env from "../../environment"
-
-// must use require to spy index file exports due to known issue in jest
-const dbUtils = require("../../db")
-jest.spyOn(dbUtils, "closeDB")
-jest.spyOn(dbUtils, "dangerousGetDB")
+require("../../../tests")
+const context = require("../")
+const { DEFAULT_TENANT_ID } = require("../../constants")
+const env = require("../../environment")
```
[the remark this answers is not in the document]
mike12345567: this had to be converted to JS - I think JS for tests is fine, there was no real type support here and the mocking is easier in some cases in JS compared to TS.

packages/backend-core/jest.config.ts
```diff
@@ -12,6 +12,7 @@ if (!process.env.CI) {
   // use sources when not in CI
   config.moduleNameMapper = {
     "@budibase/types": "<rootDir>/../types/src",
+    "^axios.*$": "<rootDir>/node_modules/axios/lib/axios.js",
```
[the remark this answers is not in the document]
mike12345567: Issues with Jest + CouchDB-nano, it utilises Axios which in the case of TS Jest tests causes some seriously weird errors. I think its a combo of Axios believing that generally using `import` means you're in the browser, and Jest not correctly informing the library that its TS imports, not browser imports.

packages/server/src/api/controllers/auth.ts
mike12345567: Yeap, I think handling this as a separate PR and defining it all is a great idea - I have noticed this is a bit of an issue in a few places!

packages/server/src/threads/automation.ts
mike12345567: Yeah, I'd been thinking that as well - we should probably have a think about refactoring this.

packages/server/src/db/utils.ts
mike12345567: Yeah I think ideally I would like to move all of this into the backend-core, its a WIP, all of this is just exposed here for now rather than taking on the massive task of getting everything properly exposed from backend-core!

Issue #6336, opened 2022-06-15
Make duplicate and delete available on every component in the action bar


AndyWhann:
##Suggestion

Not every component currently displays the action bar. 

The suggestion would be to show the duplicate and delete options inside the action bar for every component. 

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/a601bd45-31e8-4804-969d-7b1d4f8b2a1f)

PR #8714, opened 2022-11-17
Show settings bar for all components

deanhannigan:
## Description

The component settings bar will now appear when selecting any component in your screen. The minimum set of actions that will appear in the bar is, `delete` and `duplicate`

The property `showSettingsBar` has been cleared from the manifest but will still be honoured if explicitly set to `false`

Also, when selecting the screen itself, the `delete` and `duplicate` component actions are now hidden in the menu

Addresses: 
- https://github.com/Budibase/budibase/issues/6336

## Screenshots

output

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -84,7 +84,6 @@
     "description": "This component contains things within itself",
     "icon": "Selection",
     "hasChildren": true,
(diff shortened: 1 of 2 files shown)


aptkingston (approved):
LGTM!

PR #8717, opened 2022-11-17
App backup triggered event

mike12345567:
## Description
Adding an event for when a backup is triggered.

Links with pro: https://github.com/Budibase/budibase-pro/pull/86

### DIFF
--- a/packages/backend-core/src/events/publishers/backup.ts
+++ b/packages/backend-core/src/events/publishers/backup.ts
@@ -11,6 +11,7 @@ import { publishEvent } from "../events"
 export async function appBackupRestored(backup: AppBackup) {
   const properties: AppBackupRestoreEvent = {
     appId: backup.appId,
+    restoreId: backup._id!,
     backupCreatedAt: backup.timestamp,
   }
 
@@ -19,11 +20,13 @@ export async function appBackupRestored(backup: AppBackup) {
 
 export async function appBackupTriggered(
   appId: string,
+  backupId: string,
   type: AppBackupType,
   trigger: AppBackupTrigger
 ) {
   const properties: AppBackupTriggeredEvent = {
     appId: appId,
+    backupId,
     type,
     trigger,
   }
--- a/packages/backend-core/src/events/publishers/backup.ts
+++ b/packages/backend-core/src/events/publishers/backup.ts
@@ -1,4 +1,9 @@
-import { AppBackup, AppBackupRestoreEvent, Event } from "@budibase/types"
+import {
+  AppBackup,
+  AppBackupRestoreEvent,
+  AppBackupTriggeredEvent,
+  Event,
+} from "@budibase/types"
 import { publishEvent } from "../events"
 
 export async function appBackupRestored(backup: AppBackup) {
@@ -9,4 +14,13 @@ export async function appBackupRestored(backup: AppBackup) {
   }
 
   await publishEvent(Event.APP_BACKUP_RESTORED, properties)
+}
+
+export async function appBackupTriggered(appId: string, name?: string) {
+  const properties: AppBackupTriggeredEvent = {
+    appId: appId,
+    backupName: name,
+    backupCreatedAt: new Date().toISOString(),
+  }
+  await publishEvent(Event.APP_BACKUP_TRIGGERED, properties)
 }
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/events/publishers/backup.ts
```diff
@@ -10,3 +15,12 @@ export async function appBackupRestored(backup: AppBackup) {
 
   await publishEvent(Event.APP_BACKUP_RESTORED, properties)
 }
+
+export async function appBackupTriggered(appId: string, name?: string) {
+  const properties: AppBackupTriggeredEvent = {
+    appId: appId,
+    backupName: name,
+    backupCreatedAt: new Date().toISOString(),
```
Rory-Powell, line 23: Posthog will already have a time stamp for this. If we wanted to add our own timestamp I'd suggest adding it to all events in the common code where tenancy, identity etc is added and add a new field to the base when type

packages/backend-core/src/events/publishers/backup.ts
```diff
@@ -10,3 +15,12 @@ export async function appBackupRestored(backup: AppBackup) {
 
   await publishEvent(Event.APP_BACKUP_RESTORED, properties)
 }
+
+export async function appBackupTriggered(appId: string, name?: string) {
+  const properties: AppBackupTriggeredEvent = {
+    appId: appId,
+    backupName: name,
```
Rory-Powell, line 22: Can we add a trigger field as well? So we can distinguish between manual, on publish or on schedule in future

Rory-Powell (changes requested):

packages/backend-core/src/events/publishers/backup.ts
mike12345567: Makes sense, will remove it - in the case of the import it is the time the backup that is being imported was created so its different anyway!

packages/backend-core/src/events/publishers/backup.ts
mike12345567: Makes sense, worth knowing!

packages/backend-core/src/events/publishers/backup.ts
```diff
@@ -10,3 +16,16 @@ export async function appBackupRestored(backup: AppBackup) {
...
 }
+
+export async function appBackupTriggered(
+  appId: string,
+  trigger: AppBackupTrigger,
+  name?: string
+) {
+  const properties: AppBackupTriggeredEvent = {
+    appId: appId,
+    backupName: name,
```
Rory-Powell, line 27: The name is probably something we don't need in here, generally anything user specific that isn't defined in our own domain like an id or a constant we should leave out of these

Rory-Powell, line 27: Should we add the `AppBackupType` as well? So we can distinguish real backups from restores

mike12345567: Is it really a negative to have this information if the user specified a specific name for the backup? Do we need to be optimising for event size?

Rory-Powell: Not that we need to optimise for event size, more that we don't pull any unnecessary personal data into analytics. 

The rationale I normally use when adding a field is
- Will I use this to do a filter?
- Will I use this to do a group by?
- etc

User entered input doesn't add any valuable insights in most cases. If there was a specific condition we wanted, like "has the user entered a custom name?", it'd be better to work that out application side and set a boolean `customName`, that would be easier to work with.

packages/types/src/sdk/events/backup.ts
```diff
@@ -1,7 +1,13 @@
...
+import { AppBackupTrigger, AppBackupType } from "../../documents"
 
 export interface AppBackupRestoreEvent extends BaseEvent {
   appId: string
-  backupName: string
   backupCreatedAt: string
 }
+
+export interface AppBackupTriggeredEvent extends BaseEvent {
+  appId: string
```
Rory-Powell, line 10: We should actually add the backup id as well, most other events have the primary id. It's always useful to have the id of the event triggering, as down the line we may use these events to trigger webhooks. On the consuming side if you have the id of the thing that triggered the event you can query for it over the API.

Issue #7760, opened 2022-09-14
Deleting selected row of table - remains selected after deletion


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7707

<div type='discussions-op-text'>

<sup>Originally posted by **eyduh** September  9, 2022</sup>
Hi,

I have a simple table view of a form and would like to allow users to download a CSV of the entire table, as well as make some simple editing of the rows such as deleting duplicate entries. 

Exporting works fine, but if I delete a row I think it stays selected somehow and interferes with other operations:

I set up a table like so:
![image](https://user-images.githubusercontent.com/29815625/189352702-f16001f4-7b06-4132-9cb7-50c6baa8610d.png)

and the delete button like so:
![image](https://user-images.githubusercontent.com/29815625/189352971-a3e88f9a-dfac-4a6f-85fd-937ba152c427.png)

if I go into the exported app and delete a row and then try to delete another row I get `missing` in red:

https://user-images.githubusercontent.com/29815625/189354941-2e483084-f608-4921-b448-c82ec4898076.mp4

when I reload the page and try to delete the same row that gave me "missing" it works fine.
</div>

---

I was able to confirm this - the row stays selected (i.e. an item in the Selected rows binding value) even after deletion.

amit-meshram:
@melohagan Is this issue resolved?

melohagan:
> @melohagan Is this issue resolved?

Not yet - I had looked at it, but it wasn't quite right.

melohagan:
Appears to be fixed

PR #8718, opened 2022-11-17
Deselect row in client table after deletion

melohagan:
## Description
You can now delete a selected row in the client table without it remaining selected and causing errors in deleting subsequent rows.

This was trickier than I initially thought it would be. The problem I ran into was that the *bbui* table was keeping track of the selected rows, but I wanted to use the rowSelectionStore to keep track of the selected rows so that I could clear it upon the **Delete Row** action. 
So I moved the toggle row functionality into the client table, but left the original toggle functionality for other types of tables such as data table.

Addresses: 
- https://github.com/Budibase/budibase/issues/7760

## App Export
[ConditionalUI-export-1668706178904.tar.gz](https://github.com/Budibase/budibase/files/10034061/ConditionalUI-export-1668706178904.tar.gz)

## Screenshots
delete_bug

### DIFF
--- a/packages/bbui/src/Table/Table.svelte
+++ b/packages/bbui/src/Table/Table.svelte
@@ -37,6 +37,7 @@
   export let autoSortColumns = true
   export let compact = false
   export let customPlaceholder = false
+  export let customToggleRowHandlers = false
   export let showHeaderBorder = true
   export let placeholderText = "No rows found"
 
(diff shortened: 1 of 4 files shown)


aptkingston:
I see where you're coming from with this, but I think we should try to avoid duplicating the row selection logic. There's a couple of other issues with this implementation, including stuff like caching table ID's at the moment of saving the delete row action, which won't account for any tables added after the delete row action is saved.

Off the top of my head, the best way to do this would probably be to treat the row selection store as the source of truth, and I think the way to do that is probably by using a derived store inside the table component to observe it's own selection state from the row selection store and update the locally bound `selectedRows` value if that derived value changes. The only gotcha is to avoid endless update loops by only updating when the length of the selected rows changes, for example.

That technique would allow the value inside the row selection store to be updated, and then for that change to propogate right down into the BBUI table. All we would need to do is then build in a row deletion handler to the row selection store to wipe a certain ID from all selected row states. This could then be invoked from the delete row action handler. This would also avoid the need to cache table IDs inside component settings.

I'm not saying you need to go off and build it this way - but just leaving my thoughts as I would prefer to avoid merging this in the current state.

melohagan:
> I see where you're coming from with this, but I think we should try to avoid duplicating the row selection logic. There's a couple of other issues with this implementation, including stuff like caching table ID's at the moment of saving the delete row action, which won't account for any tables added after the delete row action is saved.
> 
> Off the top of my head, the best way to do this would probably be to treat the row selection store as the source of truth, and I think the way to do that is probably by using a derived store inside the table component to observe it's own selection state from the row selection store and update the locally bound `selectedRows` value if that derived value changes. The only gotcha is to avoid endless update loops by only updating when the length of the selected rows changes, for example.
> 
> That technique would allow the value inside the row selection store to be updated, and then for that change to propogate right down into the BBUI table. All we would need to do is then build in a row deletion handler to the row selection store to wipe a certain ID from all selected row states. This could then be invoked from the delete row action handler. This would also avoid the need to cache table IDs inside component settings.
> 
> I'm not saying you need to go off and build it this way - but just leaving my thoughts as I would prefer to avoid merging this in the current state.

That's fair, it did feel a bit 'hacky' handling the row toggling in a different way just for the client table. I'll not work on the suggested fix myself because it probably will be a fairly involved bit of work. 

I'll close the PR, but thanks very much for taking a look!

Issue #8313, opened 2022-10-18
Progressive loading states for data components


shogunpurple:
When I'm loading a data related element - such as a table, form, card or repeater, I want the element to load in a progressive fashion, so that the loading of the page is smoother and less janky.

**COS**
- Add progressive loading states to all data elements. eg. Load the container of the table without rows, then load the data inside, or the container for cards, with skeleton cards.
- Add an option to data providers to delay the showing of the spinner by a ms value for more control.

PR #8719, opened 2022-11-17
Skeleton Loading States

Ghrehh:
## Description
Currently while a DataProvider is loading we will show a spinner. These spinners are of a fixed size and so don't accurately represent the content they are placeholders for, which can be of various sizes. This leads to a significant jump in the placement of components once they're loaded.

Also with our current set up, any child of a DataProvider isn't rendered until said DataProvider has finished fetching its content. This means that if DataProviders are nested in any way it can lead to especially janky staggered loading spinners.

This PR replaces loading spinners with loading skeletons, which aim to more accurately represent the size of the content they are placeholders for. 

This PR also makes changes to the DataProvider and Repeater components to render their children while loading data, meaning nested DataProviders will also load skeleton components on page load.

If anyone has any particularly gnarly apps locally they could test this on it would be greatly appreciated.

## Examples

Note that gif compression makes the skeleton loading shimmer look weirder than it is in person, though I did just wing this without design input, so ofc very open to suggestions.

Description|Spinner|Skeleton
|---|---|---|
Cards Block|cards-nested|cards-skeleton
Charts|charts-spinner|charts-skeleton
Row Explorer|spinner-explorer|explorer-skeleton
Nested DataProviders|spinner-nested|nested-skeleton
Fixed Width Paragraph|paragraph-spinner|paragraph-skeleton
Todo List|todo-spinner|todo-skeleton

Addresses: 
https://github.com/Budibase/budibase/issues/8313

### DIFF
--- a/packages/bbui/src/Skeleton/Skeleton.svelte
+++ b/packages/bbui/src/Skeleton/Skeleton.svelte
@@ -1,0 +1,56 @@
+<div class="skeleton">
+  <div class="children">
+    <slot />
+  </div>
+</div>
+
+<style>
+  .skeleton {
+    height: 100%;
+    width: 100%;
+    opacity: 0;
+    background-color: var(--spectrum-global-color-gray-300) !important;
+    border-radius: 7px;
+    overflow: hidden;
+    position: relative;
+    animation: fadeIn 130ms ease 0s 1 normal forwards;
+  }
+
+  .children {
+    pointer-events: none;
+    opacity: 0;
+  }
+
+  .skeleton::after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    transform: translateX(-100%);
+    background-image: linear-gradient(
+      90deg,
+      rgba(255, 255, 255, 0) 0,
+      rgba(255, 255, 255, 0.2) 20%,
+      rgba(255, 255, 255, 0.5) 60%,
+      rgba(255, 255, 255, 0)
+    );
+    animation: shimmer 2s infinite;
+    content: "";
+  }
+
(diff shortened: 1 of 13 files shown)


JayP718:
This is awesome, we have a lot of components and it looks a little funky with all the spinners. This is much cleaner in my opinion, as long as i'm understand correctly

shogunpurple (approved):
This is awesome @Ghrehh - looks great. Insane how much faster and more peformant it seems just from the visual update.

aptkingston (commented):
Firstly - great work! I reckon we should try fading in the skeleton... In the majority of cases data from internal tables loads so quickly that I find the flashing of the grey skeleton more jarring than the flashing of a spinner - it stands out a lot more (and is a lot bigger of course). I reckon if we faded in the skeleton over a short duration then fast loads (like internal tables) would be solved since you wouldn't see the skeleton by the time the data loads.

Ghrehh:
@aptkingston I faded in the skeleton as you asked, you can see kind of see it in the attached gifs. I also tried to transition in the skeleton > actual content but the `display: contents` div we wrap components in doesn't accept any sort of animation, so no dice on that front.

I also fixed the skeletons for forms and tables as pointed out by Dean in the Avalanche standup. I did this by adding a "skeleton" boolean property to component entries in the manfiest, if this is set to `false` then the component will be allowed to handle it's own loading state and a skeleton won't be rendered for it by `Component.svelte`.

With this we then handle skeletons for the two outlined cases in a more specific way. 

new-form-skeleton
new-table-skeleton

joebudi:
Great work @Ghrehh . You're on a roll sir!

Issue #5366, opened 2022-04-12
Dynamic filter with custom column labels


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5358

<div type='discussions-op-text'>

<sup>Originally posted by **gerhardmpl** April 12, 2022</sup>
Allowing the use of configurable labels instead of default column names will increase user experience (like the _config columns_ feature for tables, maybe even using them for dynamic filter).</div>

EuanRW:
This would be a great feature, any idea of when this will be added?

PR #8726, opened 2022-11-18
Dynamic filter custom labels

melohagan:
## Description
Users wanted to be able to give the columns in their dynamic filter custom names to match their table column custom labels.
A new type of *ColumnEditor* was added that does the same as the column config, just without the cell editing.

You can:
 - Change the label names of your dynamic filter columns
 - Change the order in which the names appear in the dropdown

Addresses: 
- https://github.com/Budibase/budibase/issues/5366

## Screenshots
Screenshot 2022-11-18 at 16 23 38

Screenshot 2022-11-18 at 16 24 00

Screenshot 2022-11-18 at 16 24 30

Screenshot 2022-11-18 at 16 24 46


packages/builder/src/components/design/settings/controls/ColumnEditor/BasicColumnEditor.svelte
```diff
@@ -0,0 +1,83 @@
+<script>
```
aptkingston, line 1: Could this entire file not just be a tiny wrapper around the existing `ColumnEditor` that simply passes on down the value of a new `allowCellEditing` prop?

e.g.
- Add `allowCellEditing` prop to `ColumnEditor`, defaulting to true
- `allowCellEditing` is always passed down from `ColumnEditor` to `ColumnDrawer`
- New `BasicColumnEditor` is a tiny component that simply passes down `{ ...$$props }` and `allowCellEditing={false}`

This would then mean all this duplication could be removed.

The only other difference I can see is the text: "table" => "dynamic filter", so we could just add another optional prop to the `ColumnEditor` to allow customising this text. In my opinion that's still much better than duplicating the file. 

Were there any other differences between the files that I'm missing?

packages/client/manifest.json
```diff
@@ -3940,10 +3940,11 @@
...
       },
       {
-        "type": "multifield",
-        "label": "Allowed filter fields",
+        "type": "allowedFields",
+        "label": "Allowed filter columns",
         "key": "allowedFields",
-        "placeholder": "All fields"
+        "dependsOn": "dataProvider",
+        "nested": true
```
aptkingston, line 3947: `nested` is only used in the context of blocks, and means that settings should not be enriched by the `string-templates` library. We should remove it from here as it serves no purpose.

packages/client/src/components/app/dynamic-filter/DynamicFilter.svelte
```diff
@@ -61,8 +61,9 @@
       clonedSchema = schema
     } else {
       allowedFields?.forEach(field => {
-        if (schema[field]) {
-          clonedSchema[field] = schema[field]
+        if (schema[field.name]) {
```
aptkingston, line 64: As we have changed the type of this setting from an array of strings to an array of objects, we need to ensure that the `DynamicFilter` component handles both. The value of the setting is only migrated whenever the setting component is rendered, so there is no guarantee that all instances are migrated.

We need to update this logic to account for both flat arrays of strings (old structure) and arrays of objects (new structure).

packages/client/src/components/app/dynamic-filter/DynamicFilter.svelte
```diff
@@ -61,8 +61,9 @@
...
+          clonedSchema[field.name] = schema[field.name]
+          clonedSchema[field.name].displayName = field.displayName
```
aptkingston, line 66: Same as above, this needs to account for the old structure of flat arrays of strings.

packages/builder/src/components/design/settings/componentSettings.js
```diff
@@ -42,6 +43,7 @@ const componentMap = {
   filter: FilterEditor,
   url: URLSelect,
   columns: ColumnEditor,
+  allowedFields: BasicColumnEditor,
```
aptkingston, line 46: I would prefer to use a more explicit name for this new settings type that's more in line with our naming convention - maybe something like `columns/basic`.

aptkingston (changes requested):
Think there's a few changes we need to make to get this shippable, but good work so far :+1:

packages/builder/src/components/design/settings/controls/ColumnEditor/BasicColumnEditor.svelte
melohagan: Good call; removed duplication!

packages/builder/src/components/design/settings/controls/ColumnEditor/BasicColumnEditor.svelte
```diff
@@ -1,83 +1,13 @@
...
-    slot="body"
-    bind:columns={boundValue}
-    {options}
-    {schema}
-    allowCellEditing={false}
-  />
-</Drawer>
+<ColumnEditor
+  {...$$props}
+  on:change={e => dispatch("change", e.detail)}
```
aptkingston, line 10: Svelte has a great shorthand way of proxying events - you can literally just have `on:change` and that's it! That basically just says "proxy any `change` events up another level".

aptkingston (approved):
LGTM now! One tiny suggestion but it's not a blocker - just a shorthand trick which is useful to learn.

packages/builder/src/components/design/settings/controls/ColumnEditor/BasicColumnEditor.svelte
melohagan: Fantastic! Made that change

PR #8742, opened 2022-11-21
Require params on db.remove

Rory-Powell:
## Description
- Require params on db.remove
- Fix table usage

## Addresses
- https://github.com/Budibase/budibase/issues/8743

Related: 
https://github.com/Budibase/budibase-pro/pull/87

### DIFF
--- a/packages/types/src/sdk/db.ts
+++ b/packages/types/src/sdk/db.ts
@@ -50,7 +50,7 @@ export interface Database {
   exists(): Promise<boolean>
   checkSetup(): Promise<Nano.DocumentScope<any>>
   get<T>(id?: string): Promise<T | any>
-  remove(id?: string, rev?: string): Promise<Nano.DocumentDestroyResponse>
+  remove(id: string, rev: string): Promise<Nano.DocumentDestroyResponse>
   put(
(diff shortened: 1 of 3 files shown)


packages/types/src/sdk/db.ts
```diff
@@ -50,7 +50,7 @@ export interface Database {
   exists(): Promise<boolean>
   checkSetup(): Promise<Nano.DocumentScope<any>>
   get<T>(id?: string): Promise<T | any>
-  remove(id?: string, rev?: string): Promise<Nano.DocumentDestroyResponse>
+  remove(id: string, rev: string): Promise<Nano.DocumentDestroyResponse>
```
mike12345567, line 53: I think it would actually be better to provide an overload like the PouchDB version - as there could be a lot of other usages of the `remove(document)` syntax, they might be in the JS side of things where we wouldn't be able to find it easily.

Rory-Powell: Pushed up the overloaded version. While it's not the approach I like, and would prefer that we had the strict id and rev requirement, it does make sense for the sake of backwards compatibility.

PR #8764, opened 2022-11-22
Render view groupBy array options as tags

deanhannigan:
## Description

When applying a `Group by` option to a view using an array field type, the options will display in the `Group` field as a string.

## Screenshots

The view below calculates a total for the `Cost` column and groups by `Expense Tags`. The options display as a string

After passing an updated table schema, the array options display as tags.


melohagan (changes requested):
Hey @deanhannigan 

I pulled this down and had a play around - looks great; much nicer!

I did notice some issues with the filtering in the **Design** section though. (Although I think the filtering of a view in the design section may not be working at the moment, it would be good to fix if possible.)

*My view*
Screenshot 2022-12-16 at 10 11 42

*No options in the dropdown within the Design section*
Screenshot 2022-12-16 at 10 12 38

*If I just use the table without the view:*
Screenshot 2022-12-16 at 10 16 06

Also there is this issue too, which is somewhat related to this work: https://github.com/Budibase/budibase/issues/8964
But if it's scope creep, that's fine.

deanhannigan:
Hi @melohagan, as you mentioned above regarding the view filtering, this is a separate issue with the filtering of a grouped view in the Designer. Definitely create a separate issue for it. The above info and screenshots are perfect!

melohagan (approved):

shogunpurple (approved):
LGTM!

PR #8769, opened 2022-11-22
Fix view creation

Rory-Powell:
## Description
- Add couch specific error handling to view creation to fix provisioning (and general first time view creation)


shogunpurple (approved):
LGTM!

Issue #8767, opened 2022-11-22
Dragging a button out of a table can cause the page to error and require a page refresh


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.31
  - App Version: 2.1.31

**Describe the bug**
Moving a button out of a table component and dragging it around can crash the builder page.

**To Reproduce**
1. Add a Data provider with a nested table
2. Click on the table and add a button into the table
3. Drag the button out of the table to the top of the data provider
4. Drag the button from the top of the data provider to the bottom (below the table)
5. Observe crash. Need to refresh page.

**Expected behavior**
Moving a button out of a table and dragging it around should not crash the builder.

**Screenshots**
<img width="1450" alt="Screenshot 2022-11-22 at 13 47 01" src="https://user-images.githubusercontent.com/101575380/203329883-28502333-b190-413c-88f8-c20347fc407d.png">

aptkingston:
I had intermittently seen this issue myself but it seemed hard to reproduce and it's an internal svelte error being thrown which makes things nearly impossible to debug.

I've had another look, and (while I'm completely guessing here) I think that it's something to do with the fact that we don't key table rows, so something is going wrong when the number of table columns reduces and svelte unmounts things. I tried remounting the entire BBUI table component when the number of columns changes and this does appear to fix the issue, so I'll go with that fix.

Issue #8312, opened 2022-10-18
Side Panel: Table Block


shogunpurple:
When I'm working with the table block, I want to have the ability to edit and manage my data from a side panel with one click, so that I can perform full table CRUD more easily.

**COS** 
- Add button action to table block to show side panel
- Once added, the side panel comes with form block inside and a button to create a new record through a form also
- When the table block is ejected, it ejects the side panel component and form blocks underneath for further editing
- A screen must support multiple side panels
- When a side panel is opened, it closes any existing side panels on the screen.
- The side panel allows you to cycle through records using arrow buttons (research/stretch)

**Designs**
https://www.figma.com/file/08ybYXvfhujH3pl5VrP0Hs/Side-panel?node-id=0%3A1

Issue #8311, opened 2022-10-18
Side Panel


shogunpurple:
## Description
The ability to show a side panel for the table block element, for easier inline data management.

## Spec
https://budibase.atlassian.net/wiki/spaces/PROD/pages/116916281/Side+Panel

## Teams and Stakeholders
- Avalanche

## Workflow
- [x]  Spec Created and pasted above
- [x]  Product Review
- [x]  Designs created
- [x]  Individual Tasks created and assigned to Epic

Issue #8310, opened 2022-10-18
Side Panel Component


shogunpurple:
When I'm viewing my screens, I want to be able to configure my global side panel, so that I can edit the settings of all side panels shown in my app.

**COS**
- Add side panel component alongside the screen element
- Add settings for size (S/M/L)

PR #8788, opened 2022-11-23
Side panels

aptkingston:
## Description
This PR adds a side panel component, and updates table blocks to use this new component (in combination with form blocks) to offer full CRUD functionality in a single component. I've also used a migration to move all table and table block settings from our existing link settings into the new action settings. Autoscreen generation has been updated so that now a single screen is generated with just a table block on it.

TLDR:
- New side panel component. Side panels are opened via button action, and closed via button action, clicking outside or by the close icon. You can have multiple side panels. Nesting doesn't matter.
- Table component and table blocks used to have link settings (e.g. on row click, on title button click). These have been replaced with actions. A migration updates all components in all apps.
- Autoscreen generation just generates 1 screen now, with a table block. You still get full CRUD via the new functionality.
- Couple of fixes for random crashes, and some small UI tweaks

### Overview
The side panel component is a container that slides in from the right side of the screen. It overlays the page and smoothly transitions in. The side panel is basically a single column container. You can nest whatever you want inside it if you require custom layouts.

image

Clarification of side panel functionality:
- You can have multiple side panels
- Nesting does not affect functionality
- Only one side panel can be open at once
- Opening a new side panel will replace any current one that is open

### Adding the side panel
The side panel component lives under the layout category. It has no settings, only a name. It is opinionated - it's a fixed width, it always has padding, it's always on the right side and it always has a close icon at the top right. We do not want this to be customised as we want to ensure a consistent, polished experience. You can always use custom CSS if you really need to.

image

### Opening and closing the side panel
The side panel is shown and hidden via actions.

image
image

The side panel can also be hidden by clicking anywhere outside the side panel, or by clicking the dedicated close button. 
image

### Side panels inside the builder
When working inside the builder, side panels will always be hidden unless you select them, or a component nested anywhere inside them. Side panels will not animate when inside the builder, but they do in real apps.

### Mobile support
Side panels work exactly the same on mobile as they do on desktop platforms. The side panel will not take up the full screen width, meaning it's always easy to close the side panel by clicking in the empty space.

image

## Table blocks with side panels
Table blocks now come with the ability to automatically show side panels with form blocks inside, providing full CRUD functionality in a single component. This functionality is only compatible with internal tables or SQL tables of course, so it is not the default. The default settings on a table block are running actions on row click.

New table block settings for clicking rows:
image

New table block settings for title button:
image

Info text has been added to highlight that this new functionality is only compatible with internal or SQL tables. If you choose the side panel options when using a custom query, nothing will happen. You will not get a broken side panel.

Here's a video showing the new table block functionality. This is full CRUD by just picking a couple of settings on a single component. Also notice how it is possible to cycle through rows without closing the side panel every time.
https://user-images.githubusercontent.com/9075550/203596883-c6477584-41eb-4cf2-8015-18d25d7bb676.mp4

## Table component and table block settings migration
As described above for table blocks, all link settings have been migrated to be button actions. Table settings for clicking on a row have also been migrated.

As we previously used an "advanced" section with a field picker to support linking to rows (due to this being a unique concept where we can't use a binding because single row context does not exist), this has now been refactored into a much cleaner (and standard) approach of an on click event, with the clicked row being a parameter.

This means that if you previously had table settings of
- Link screen = `/sales/:id`
- Link column = `_id`

That this will now be migrated to a single action setting of `Navigate To` with a URL of `/sales/{{ Clicked row._id }}`. Because the table block supports custom queries, we cannot assume that `_id` exists. If we updated table blocks to only work with plus datasource (like some other blocks do) then we could provide a binding for `Clicked row ID` and just use that.

New binding for clicked row:
image

The migration for these component settings is non breaking. Because we cannot guarantee that the client library will be updated, we need to migrate settings in such a way that they will work with both the old and new version of the client library. To support this, the new settings are under entirely different keys, and the migration script will simply add the new settings *without removing the old settings*. This means that app    
s will continue to function normally with the old client library version, but will also magically have the new settings filled in as soon as the client library is updated.

## Autoscreens
Autoscreens will now only generate a single screen (the old list screen). This screen has a table block which is preconfigured to use side panels to offer full CRUD. This keeps autoscreens extremely simple and the list of screens smaller.

As autoscreens are only compatible with tables, we can guarantee that they will also be compatible with this new CRUD functionality.

## Other stuff
- Fixed a random flatpickr issue which was being thrown when using a date field inside a form block sometimes
- Center aligned the login screen prompt text
- Improved DND logic to support side panels and prevent some rare crashes
- Rewrote the `makeComponentUnique` util to properly update all IDs regardless of nesting position. This was bugged previously and only replaced instances nested deeper in the tree than the parent.
- Remount table component when column length changes to fix DND bug with tables
- "Eject block" button is now in its own section for all blocks, rather than at the bottom of the general section

Addresses:
- #8310 
- #8311 
- #8312 
- #8767

## TODO
Found a few limitations while testing this, so these are still being worked on:
- ~~Using a datepicker inside a side panel will cause the "click outside" handler to fire, and close the side panel~~
- ~~Using a confirmation modal triggered from inside the side panel will cause the "click outside" handler to fire, and close the side panel~~
- ~~Side panel content is not reset once it is hidden. This means using a "new row" side panel to create a row, then opening the side panel again, will show the same stale values. We can't unmount the content as this breaks ejection, so it needs to just be remounted.~~ (form blocks now automatically clear themselves when a "create" form is submitted)
- Work out a nice way to add the unique instruction text when viewing side panels. From designs:
image


aptkingston:
@Rory-Powell would appreciate a quick review of the migration as this is my first time using one. Loved using it by the way - it's so powerful. Nice work.

packages/server/src/migrations/functions/tableSettings.ts
```diff
@@ -0,0 +1,143 @@
...
+      console.log(
+        `Screen ${screen.routing?.route} contained table settings which were migrated`
+      )
+    }
+  })
+}
+
+// Recursively searches and mutates a screen doc to migrate table component
+// and table block settings
+const migrateTableSettings = (component: any, screen: any) => {
```
Rory-Powell, line 58: Could we type `Screen` here. 

NAB: 
Would be awesome to have a base `Component` type. I could see this extending down the line where we have type safety on the components for doing migrations like this. 
```ts
interface Component {}
interface TableComponent extends Component {}

// with useful type assertion functions
const isTableComponent = (component: Component): component is TableComponent => {
   return component._component.endsWith("/table")
}
```

packages/server/src/migrations/functions/tableSettings.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Legacy "linkRows", "linkURL", "linkPeek" and "linkColumn" settings on tables
+ * and table blocks are migrated into a "Navigate To" action under the new
+ * "onClick" setting.
+ *
+ * Migration 2:
+ * Legacy "titleButtonURL" and "titleButtonPeek" settings on table blocks are
+ * migrated into a "Navigate To" action under the new "onClickTitleButton"
+ * setting.
+ */
+export const run = async (appDb: any) => {
```
Rory-Powell, line 23: A general requirement of migrations are that they are idempotent. Would it be safe if this migration was to be run twice on the same app?

It could happen where get through some screens successfully, but another screen fails. The entire migration would not succeed and would start again on all screens the next run.

Rory-Powell, line 23: Given the high amount of logic in this migration it would be great to see a unit test

packages/server/src/migrations/functions/tableSettings.ts
```diff
@@ -0,0 +1,143 @@
+import { getScreenParams } from "../../db/utils"
```
Rory-Powell, line 1: Unrelated to migrations but from something in the description
> If we updated table blocks to only work with plus datasource (like some other blocks do) then we could provide a binding for Clicked row ID and just use that.

Would it be possible for us to auto provide that binding only when it is a datasource plus? 
And exclude it when it's not a datasource plus?

Generally using any sort of `row._id` binding in the builder feels yucky to me when not using custom queries, potentially raising the barrier to entry for non technical folks.

Rory-Powell:
> @Rory-Powell would appreciate a quick review of the migration as this is my first time using one. Loved using it by the way - it's so powerful. Nice work.

Thanks cheeks! Great to see you were able to jump in and get a new one setup using the existing patterns. If you need any help running the migrations in prod see this doc or give me a shout

https://budibase.atlassian.net/wiki/spaces/EN/pages/9896119/Migrations

packages/server/src/migrations/index.ts
```diff
@@ -73,6 +74,14 @@ export const buildMigrations = () => {
         })
         break
       }
+      case MigrationName.TABLE_SETTINGS: {
+        serverMigrations.push({
+          ...definition,
+          appOpts: { all: true },
```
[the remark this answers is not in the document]
aptkingston: It won't have an affect since it's only adding new values, as opposed to changing or deleting existing values, but yea there's no reason to run it on prod apps since it's impossible for prod app client libraries to update without first being published again from the dev app. So I guess we can filter them out to save time.

packages/server/src/migrations/functions/tableSettings.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  // Recursively update any relevant components and mutate the screen docs
+  screens.forEach((screen: any) => {
+    migrateTableSettings(screen.props, screen)
+
+    // Save screen if we updated it
+    if (screen.touched) {
+      delete screen.touched
+      appDb.put(screen)
```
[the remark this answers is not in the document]
aptkingston: Oops my bad - just left it out. Well spotted!

packages/server/src/migrations/functions/tableSettings.ts
aptkingston: I had it typed originally, but then because I needed to use a temporary `touched` property on the screen to only save updated screens, that of course broke the typing since it doesn't exist on the type. It doesn't use the screen for anything other than setting the `touched` property anyway so we aren't missing out on anything.

I agree about the component typing - it would be nice to have.

packages/server/src/migrations/functions/tableSettings.ts
aptkingston: Yep - the migration will only change components that do not have the new settings defined. Once the migration has run and the new settings are defined, running the migration again would do nothing.

packages/types/src/sdk/migrations.ts
```diff
@@ -44,6 +44,7 @@ export enum MigrationName {
   EVENT_GLOBAL_BACKFILL = "event_global_backfill",
   EVENT_INSTALLATION_BACKFILL = "event_installation_backfill",
   GLOBAL_INFO_SYNC_USERS = "global_info_sync_users",
+  TABLE_SETTINGS = "table_settings",
```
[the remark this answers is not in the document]
aptkingston: Sure thing.

packages/server/src/migrations/functions/tableSettings.ts
aptkingston: I'm not a fan either but sadly we can't - the context provided to the action handler is defined in the manifest which is static and unaware of any sort of logic like that, so it can't be conditional.

packages/server/src/migrations/functions/tableSettings.ts
aptkingston: Caught :sweat_smile: Yea I should probably add one. I did mean to at one point but the scope of this PR kept growing and I must have forgotten about it. I'll add one :+1:

packages/server/src/migrations/functions/tableSettings.ts
aptkingston: I can refactor this to just use tail end recursion and return a bool stating whether or not a change needs made - then I don't need to pass in the screen at all actually, and it's probably cleaner.

aptkingston:
> 

Will do, cheers Rory. I'll touch base again regarding running the migration in production when it comes to deploying this, which could be January I think.

Appreciate the detailed review by the way!

packages/server/src/migrations/functions/tableSettings.ts
Rory-Powell: Ah makes sense. Another option that could have been used is a locally defined utility type

```ts
type MigratingScreen = Screen & { touched: boolean }
```

But your new approach sounds good 👍

packages/server/src/migrations/functions/tableSettings.ts
Rory-Powell: That's sad 😔

joebudi:
Awesome work @aptkingston . I think this addition is much more important than it may first look. Looking forward to trying it!

packages/bbui/src/Form/Core/DatePicker.svelte
```diff
@@ -23,6 +23,15 @@
   let open = false
   let flatpickr, flatpickrOptions
 
+  // Another classic flatpickr issue. Errors were randomly being thrown due to
+  // flatpickr internal code. Making sure that "destroy" is a valid function
+  // fixes it. The sooner we remove flatpickr the better.
```
shogunpurple, line 28: 😅

shogunpurple (approved):
LGTM! Great work. 

Just a few things on the product side - it would be really good to get some events in here for usage of the side panel. I know we ship it by default in the screens, but I'd like to understand the usage of the side panel based controls when since it's non-default in the table block. Could we add events for the settings in the table block for using the side panel?

packages/server/src/migrations/functions/tests/tableSettings.spec.ts
```diff
@@ -0,0 +1,132 @@
+import { App, Screen, ScreenProps } from "@budibase/types"
+
+const { db: dbCore } = require("@budibase/backend-core")
+const TestConfig = require("../../../tests/utilities/TestConfiguration")
+const migration = require("../tableSettings")
```
Rory-Powell, line 5: Because this is a `spec.ts` you can use `import` here instead of require. We can then benefit from better type and build checking in this test for the imports as they will no longer resolve to `any`.

packages/server/src/migrations/functions/tests/tableSettings.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+// Local type for allowing any child components inside screens
+type MigratingScreenProps = ScreenProps & { _children: any[] }
```
Rory-Powell, line 8: Should `_children` be added to the base `ScreenProps` type? if it's not something that exists only in this migration that should probably be the case

aptkingston: `ScreenProps` already has `_children` defined as `ScreenProps[]`, which I'm guessing is because `ScreenProps` is at least more specific than just calling it `any`, but it's not actually correct as it's basically stating that every child of a screen has container settings (`size`, `gap`, `direction`) etc and no others.

I've just changed it to `any` for the test here since I'm adding tables and table blocks to the screen. I guess we don't use this type in the real app otherwise it would break.

This basically comes back to what you said before about having actual component types would be good - but I would prefer not to scope creep this PR by including the introduction of component types since I'm sure that would require quite a few changes.

I'm happy up date `ScreenProps` to change the `_children` property to `any[]` since that's technically correct for now, if you think that's better than just using a temporary type here.

Rory-Powell: Sounds good!

JeroenvdBurg:
Wow! This is an amazing feature for quick row editing. Especially because the parent grid refreshes when changes are made in the side panel. This does not happen with the opening detail view in a popup. Looking forward to use this.

packages/bbui/src/Form/Core/DatePicker.svelte
aptkingston: Maybe next lab day I'll do something about this :joy:

aptkingston:
> LGTM! Great work.
> 
> Just a few things on the product side - it would be really good to get some events in here for usage of the side panel. I know we ship it by default in the screens, but I'd like to understand the usage of the side panel based controls when since it's non-default in the table block. Could we add events for the settings in the table block for using the side panel?

Have added a new feature to optionally capture events for any settings changes. Opt-in of course, to prevent spam. I've enabled it just for the table block "on row click" setting for now.

Issue #8794, opened 2022-11-24
Updating dynamic REST datasource variables causes the consuming requests to fail


deanhannigan:
**Describe the bug**

The User/Auth request context is not being passed to requests updating dynamic REST variables.

**To Reproduce**
1. Create a REST datasource
2. Create a REST query (R1) and define a dynamic variable from the response.
3. Create a second REST query (R2) and add the newly created dynamic variable to a part of the request e.g 
add `{{ Dynamic.your_variable }}` to the params or headers.
4. Run the request. _If_ R2 fails and the dynamic variable needs to be refetched again, R1 will fail as the Auth/User binding context info is missing.

**Expected behavior**
- When a REST request determines that it must update its dynamic variables, the associated REST requests should run without issue.
- The dynamic variable will be updated and the consuming REST request will run without issue.

**Screenshots**
![dynavars](https://user-images.githubusercontent.com/5913006/203759873-e7224990-757b-4ba0-8d8d-25db2a7f6181.png)

PR #8795, opened 2022-11-24
User/Auth context now being passed to dynamic variable request updates

deanhannigan:
## Description

Auth and User context is now being passed to REST requests when updating dynamic variables.

Requests depending on retrying requests in order to update dynamic variables were failing as a result.

Addresses: 
- https://github.com/Budibase/budibase/issues/8794


melohagan (approved):
LGTM!

Rory-Powell (approved):

Issue #8796, opened 2022-11-24
Home Screen selection is lost after updating actions onLoad screen


AlejandroFNadal:
**Hosting**
- Self
  - Method:  docker compose
  - Budibase Version: 2.1.31
  

**Describe the bug**
I choose one of my screens as a home screen. Whenever I update the on Load actions of a screen, the selection of the Home is lost

**To Reproduce**
1. Create an empty app
2. Create two screens
3. Select one of the screens as home screen
4. Pick the other screen, and add a on Load action
5. The home screen icon in the previous screen is lost, and the home behavior is lost.

**Expected behavior**
I expect the home screen selection to remain as chosen.

**Screenshots**
**Set as homescreen**
![Screenshot 2022-11-24 at 12 22 19](https://user-images.githubusercontent.com/101575380/203783511-823cdfc2-21fe-4389-a6d2-d7195c23e63d.png)

**On the other screen define on screen load action. Home screen is reset.**
![Screenshot 2022-11-24 at 12 23 27](https://user-images.githubusercontent.com/101575380/203783700-62f89081-b681-4180-a9e0-4786c9c3289a.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version 107.0

AlejandroFNadal:
Thanks for such a fast fix 👍

PR #8798, opened 2022-11-24
Only uncheck other homescreens if homescreen set

melohagan:
## Description
Setting onLoad actions was resetting other homescreens. Added check to make sure that you only unset other homescreens if the updated screen is set as home.

Addresses: 
- https://github.com/Budibase/budibase/issues/8796


aptkingston (approved):
LGTM! It was actually any screen setting change at all was resetting home screens, not just on load actions, so this was definitely a bad bug! Not sure how it went unnoticed for so long...

PR #8829, opened 2022-11-26
Complete backend Typescript conversion

mike12345567:
## Description
This PR finishes converting our backend repos (worker, server and backend-core) to Typescript. There isn't much to really document for this PR, beyond the fact that I have now disabled the `allowJs: true` flag on all of our Typescript builds, so that all future work stays built in Typescript.

There is still work that can be done to tighten up our definitions in places, but this is already a fairly large improvement to our build step, throughout the conversion I found a heap of edge cases, calls to functions that don't exist, usages of poorly defined enums etc.

Also need to convert the names of some older enums, but wanted to get all the bug fixes this brings and get this in a good bit before the code freeze incase it surfaces any issues with other branches.


Rory-Powell (approved):
Incredible effort 👏 

Would suggest testing this with account-portal and pro just to make sure the backend-core changes play nicely. Could probably even flip there `allowJs` flags there now too 

Great job

mike12345567:
@Rory-Powell - I've pushed updates to both account-portal and pro after testing and disabling the `allowJs` option - now no service allows the use of JS.

shogunpurple (approved):
Seriously impressive piece of work - will solve a ton of build/module import quirks too. Great job 👏

Issue #8258, opened 2022-10-14
Allow developer Users to give access to apps 


MihailHadzhiev2022:
When I tried to give access to an app as a Developer users I got 403 error.

https://www.loom.com/share/bbb05f097be14310bba507b928af220b

mielune:
Same here

PR #8844, opened 2022-11-28
Allow developers to set user access

melohagan:
## Description
Developers can now set user app access.

Addresses: 
- https://github.com/Budibase/budibase/issues/8258

## Screenshots
Screenshot 2022-11-28 at 16 05 34

Screenshot 2022-11-28 at 16 05 48

Screenshot 2022-11-28 at 16 06 07

Screenshot 2022-11-28 at 16 06 35

### DIFF
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -23,11 +23,10 @@ const MAX_USERS_UPLOAD_LIMIT = 1000
 
 export const save = async (ctx: any) => {
   try {
-    if (
-      !ctx.request.body._id &&
-      !ctx.internal &&
-      (!ctx.user || !ctx.user.admin || !ctx.user.admin.global)
-    ) {
+    const body =  ctx.request.body
+    const isCreate = !body._id
+    const isAdmin = !!ctx.user.admin?.global
+    if (isCreate && !isAdmin) {
       ctx.throw(403, "Only admin user can create new user.")
     }
     ctx.body = await sdk.users.save(ctx.request.body)
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -23,7 +23,7 @@ const MAX_USERS_UPLOAD_LIMIT = 1000
 
 export const save = async (ctx: any) => {
   try {
-    const body =  ctx.request.body
+    const body = ctx.request.body
     const isCreate = !body._id
     const isAdmin = !!ctx.user.admin?.global
     if (isCreate && !isAdmin) {
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -23,12 +23,6 @@ const MAX_USERS_UPLOAD_LIMIT = 1000
 
 export const save = async (ctx: any) => {
   try {
-    const body = ctx.request.body
-    const isCreate = !body._id
-    const isAdmin = !!ctx.user.admin?.global
-    if (isCreate && !isAdmin) {
-      ctx.throw(403, "Only admin user can create new user.")
-    }
     ctx.body = await sdk.users.save(ctx.request.body)
   } catch (err: any) {
     ctx.throw(err.status || 400, err)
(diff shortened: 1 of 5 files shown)


packages/worker/src/api/routes/global/users.js
```diff
@@ -51,20 +51,20 @@ function buildInviteAcceptValidation() {
...
     "/api/global/users",
-    adminOnly,
+    builderOrAdmin,
     users.buildUserSaveValidation(),
     controller.save
   )
   .post(
     "/api/global/users/bulk",
-    adminOnly,
+    builderOrAdmin,
```
Rory-Powell, line 60: It's not ideal for us to allow developers to effectively manage all user endpoints around creation and deletion. It's more of an unfortunate consequence of us not using the correct rest semantics, where we could allow builder on a `PUT` request only. 

I think the correct thing to do here is to 
- allow the `builderOrAdmin` on the two `POST` requests
- remove from the `DELETE`
- add request validation at the controller layer to check:
   - is this a create - i.e. no `_id` specified
   - is the `ctx.user` not an admin
   - if so, throw a manual error, to mimic the `builderOrAdmin` middleware behaviour

Rory-Powell: It would also be good to add some RBAC tests for each of the changed endpoints to ensure the desired user type can access the endpoint, and will be rejected if not. i.e.

```ts
describe("POST /api/global/users, () => {
   it("rejects non-admin on create", () => { })
   it("allows builder or admin on edit", () => { })
})

describe("POST /api/global/users/bulk, () => {
   it("rejects non-admin create", () => { })
   it("allows builder or admin on edit", () => { })
})
```

packages/worker/src/tests/api/users.ts
```diff
@@ -91,11 +91,11 @@ export class UserAPI {
 
   // USER
 
-  saveUser = (user: User, status?: number) => {
+  saveUser = (user: User, status?: number, headers?: any) => {
```
Rory-Powell, line 94: I'd like to start promoting this pattern for the optional args of status and headers:
https://github.com/Budibase/budibase/blob/develop/packages/worker/src/tests/api/users.ts#L135-L140

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -129,23 +146,24 @@ class TestConfiguration {
...
       email: "tenant1@test.com",
       password: "test",
     })
-    this.tenant1User = await this.createUser(user)
+    this.tenant1User = await this.createUser(user, { force: true })
   }
 
   async createSession(user: User) {
-    await sessions.createASession(user._id!, {
+    const session: any = {
```
Rory-Powell, line 161: Can we use the `CreateSession` interface here

Rory-Powell: In fact if you just remove the `any` we should get type inference here

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -23,6 +23,13 @@ const MAX_USERS_UPLOAD_LIMIT = 1000
 
 export const save = async (ctx: any) => {
   try {
+    if (
+      !ctx.request.body._id &&
+      !ctx.internal &&
+      (!ctx.user || !ctx.user.admin || !ctx.user.admin.global)
```
Rory-Powell, line 29: I find this block quite hard to read, could it be simplified to:

```ts

const body =  ctx.request.body // re-use this for sdk.users.save
const isCreate = !body._id
const isAdmin = !!ctx.user.admin?.global

// this is a create and we are not admin
if (isCreate && !isAdmin) {
   // throw
}
```

Rory-Powell: if we want to retain the ability for the internal api key to create users no matter what, we could do:

```ts
if ((isCreate && !isAdmin) && !ctx.internal) {
   // throw
}
```

Rory-Powell: should also add a test to ensure the internal api key headers can be used

packages/worker/src/tests/api/users.ts
melohagan: Not going to implement that at the moment as I'm working off master branch, and that pattern is currently on develop.

Rory-Powell (approved):
LGTM!

PR #8845, opened 2022-11-28
Skeleton Fixes

Ghrehh:
## Description

Some changes to fix edge cases where skeleton loaders didn't show up on page load.

There are still some issues with Skeletons remounting as parent data providers finish their requests, which disrupts the skeleton shimmer animation somewhat. As far as I can tell this shouldn't be happening, but I don't have enough svelte/domain knowledge to figure out this one on my own.

I assume removing these guards is fine now because of the loading changes, it seems to work okay anyway. Otherwise I can change this to leverage the existing loading provider also. Any input welcome.

### DIFF
--- a/packages/client/src/components/Component.svelte
+++ b/packages/client/src/components/Component.svelte
@@ -479,6 +479,7 @@
     definition.name !== "Screenslot" &&
     children.length === 0 &&
     !instance._blockElementHasChildren &&
+    !definition.block &&
     definition.skeleton !== false
 </script>
 
--- a/packages/client/src/components/app/blocks/CardsBlock.svelte
+++ b/packages/client/src/components/app/blocks/CardsBlock.svelte
@@ -36,7 +36,6 @@
   let dataProviderId
   let repeaterId
   let schema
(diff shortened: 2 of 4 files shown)


aptkingston:
I'm a little bit apprehensive about removing all these loading guards just because they exist to prevent null references all over the place, but I guess the skeletons will still prevent the real components being rendered until loading is complete. There's bound to still be edge cases which cause errors after these changes but we can fix those whenever they happen.

I guess most stuff has already been written to always assume anything can be nullish since the majority of settings are fully user controllable.

If nothing is erroring with these changes then it's good to go I guess :+1:

aptkingston:
Although these table block changes will cause major conflicts with the new side panels PR that I was about to merge, so if you wouldn't mind holding off until I've merged that one - then just take the new table block and remove the loading guards again?

Ghrehh:
@aptkingston Could you review this when you get a moment please?

aptkingston (approved):
LGTM! The diffs are so awful - you'd think it would be able to detect things like just removing one layer of indendation from everything...

Issue #8837, opened 2022-11-28
Problem deleting an application


KevinHamonApp:
**Hosting**

- Cloud
  - Tenant ID: `hamon `

**Describe the bug**
I deleted my app from the API, from the interface here: https://docs.budibase.com/reference/delete_applications-appid 
The request went well. But then, I still have my app in the apps list on my account. I can't edit / manage it anymore (404 not found).
But I can't add a new app as I reached the 4 limit from free cloud. 

**To Reproduce**
1. Create an app
2. Go to this url https://docs.budibase.com/reference/delete_applications-appid 
3. Specify the app id and api-key
4. Launch the request
5. Come back to your account
6. See error

**Expected behavior**
I would have expected to have deleted my app from the list.

**Screenshots**
![image](https://user-images.githubusercontent.com/74555017/204258668-aba8f5b4-d30e-4593-ba47-bf86485a3b05.png)
![image](https://user-images.githubusercontent.com/74555017/204258867-593796b1-c8c9-4f8b-a3f5-cd6195176451.png)

shogunpurple:
Hi @KevinHamonApp 

I've just looked into this and I'm able to reproduce. You can fix your issue in the meantime by deleting the **dev app**, in your case the appID is:
`app_dev_hamon_6a89ca6a030e47c0b65f0f5f62eaa709`

The problem here is that the minIO bucket is shared between the dev and production apps, and was deleted when you deleted the published app. I'll add some code to prevent this when the app is published, so that the dev app is still accessible.

Issue #8661, opened 2022-11-12
Cannot create a table and UI getting stuck in google sheets settings


smad-bro:
**Hosting**

- Self
  - Method: docker single image
  - Budibase Version: <version> 2.1.24
  - App Version: <version> 2.1.24

**Describe the bug**
The UI get stuck and create button is disabled in the "Create new table" popup window. Even if I type any name, the create button is still in disabled condition and also the Ui is stuck (The close and cancel button is not working). I have to refresh the window to close it. Only I can fetch a table if that sheet has some header field values. 
The google sheet API is connected and its logged in successfully.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the Data tab and click on Add Datasource button.
2. Click on google sheets and log in to your Google account. (Google API is already setup in Auth tab)
3. Input the google sheet URL and click on Save.
4. Click on New Table and Enter a table name.

**Expected behaviour**
The create button should be enabled when there is a value in the field and that popup window should be able to close.

**Screenshots**
![Screenshot (197)](https://user-images.githubusercontent.com/67038642/201470938-f06436b7-83f0-4b08-a829-b610a938e525.png)

**Desktop :**
 - OS: Windows 10
 - Browser : Chrome

Issue #6307, opened 2022-06-14
Admin checklist shows for dev user


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 1.0.192-alpha.6
  - App Version: 1.0.192-alpha.6

**Describe the bug**
As a dev user, I can still see the admin checklist.

**To Reproduce**
1. Don't complete all admin checklist items
2. Log-in with Dev account
3. You can see the admin checklist

**Expected behavior**
Admin checklist should not appear for dev account

**Screenshots**
![Screenshot 2022-06-14 at 15 06 48](https://user-images.githubusercontent.com/101575380/173597298-392c70b5-dd52-47df-9e46-bb2a0e54071a.png)

PR #8846, opened 2022-11-28
fixes for google sheets, admin checklist, and deleting an app from API

shogunpurple:
## Description
Just a few smaller ones while on call. Google sheets one was actually quite bad, because you couldn't create a new sheet without crashing the builder due to a lack of a header row.

Addresses: 
- https://github.com/Budibase/budibase/issues/6307
- https://github.com/Budibase/budibase/issues/8837
- https://github.com/Budibase/budibase/issues/8661

## App Export

## Screenshots


melohagan (approved):
LGTM!

packages/server/src/integrations/googlesheets.ts
```diff
@@ -286,7 +286,7 @@ class GoogleSheetsIntegration implements DatasourcePlus {
   async createTable(name?: string) {
     try {
       await this.connect()
-      return await this.client.addSheet({ title: name })
+      return await this.client.addSheet({ title: name, headerValues: ["test"] })
```
Rory-Powell, line 289: Is this a real value used by google, or something added for testing in dev? would an empty array produce the same fix?

packages/server/src/api/controllers/application.ts
```diff
@@ -470,7 +470,7 @@ const destroyApp = async (ctx: any) => {
   }
 
   /* istanbul ignore next */
-  if (!env.isTest() && !isUnpublish) {
+  if (!env.isTest() && !isUnpublish && isDevAppID(appId)) {
```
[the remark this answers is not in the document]
Rory-Powell: we should add a test

```ts
describe("destroyApp", ()=> {
   it("prevents deletion of prod app without unpublish", () => {})
})
```

packages/server/src/integrations/googlesheets.ts
shogunpurple: This is just part of the `google-spreadsheet` library - we have to put a header in the new sheet, or we are unable to fetch the schema for the new table (sheet) and it breaks the builder

packages/server/src/api/controllers/application.ts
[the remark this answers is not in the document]
shogunpurple: The issue here is that the production (only) bucket is deleted when you delete an app through the API . So if you delete the prod app ID, the dev app is still there, but you can't access it in the builder and it's essentially bricked until you delete it as well. This just makes sure that we don't blow away the bucket when the published app is deleted. Thought it easier to just make sure we don't delete the bucket on prod appIDs.

Will add a test 👍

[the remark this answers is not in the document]
Rory-Powell: Makes sense! I think we should still add the validation, even if we keep the `isDevAppID` check on the `deleteApp` call, so that we are protected against any future issues like this.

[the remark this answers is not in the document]
shogunpurple: Actually yes - it is better to just set the unpublish flag in the public API endpoint - added protection to the endpoint to prevent this happening again and make sure that every delete call to that endpoint for a production appID will fail without the unpublish flag, then added a middleware to the public API application `DELETE` endpoint to append the unpublish to it automatically.

packages/server/src/api/routes/public/applications.ts
```diff
@@ -92,7 +93,19 @@ write.push(
...
  */
-write.push(new Endpoint("delete", "/applications/:appId", controller.destroy))
+write.push(
+  new Endpoint(
+    "delete",
+    "/applications/:appId",
+    controller.destroy
+  ).addMiddleware((ctx, next) => {
+    // set unpublish if the appId is production, so we don't delete the object store bucket
+    if (isProdAppID(ctx.params.appId)) {
```
Rory-Powell, line 103: Nice one!

The only thing about this may be confusion on the users side, where they delete with a prod app id and still see the dev app. Our public api docs don't document anything around unpublishing.

To make the api as easy to use as possible, should we just do the unpublish and the delete in one step if the prod app id is provided?

i.e. we could have a wrapping public api controller that calls `destroy` two times
or we could just update the base functionality in the existing controller to do this

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -218,10 +205,10 @@ const _deployApp = async function (ctx: any) {
...
 
-  let app = await deployApp(deployment, ctx.user._id)
+  let app = await publishApp(deployment, ctx.user._id)
 
   await events.app.published(app)
   ctx.body = deployment
 }
 
-export { _deployApp as deployApp }
+export { _publishApp as publishApp }
```
Rory-Powell, line 214: Could just `export function publishApp` instead of the re-exporting here

packages/server/src/api/routes/application.ts
```diff
@@ -33,6 +34,16 @@ router
     authorized(BUILDER),
     controller.revertClient
   )
+  .post(
+    "/api/applications/publish",
```
Rory-Powell, line 38: it's really rubbing me the wrong way how we don't use the app id in the path here while we do for the unpublish. Can we make them consistent and include the `appId` in both

packages/server/src/api/controllers/public/applications.ts
```diff
@@ -68,10 +69,32 @@ export async function destroy(ctx: any, next: any) {
   })
 }
 
+export async function unpublish(ctx: any, next: any) {
+  await doInAppContext(ctx.params.appId, async () => {
+    // get the app before deleting it
+    await controller.fetchAppPackage(ctx)
```
Rory-Powell, line 75: Do we need this line? Like like we're ignoring the result

Rory-Powell: I think a `204 - No Content` would be fine here, rather than the returning the app. 
Returning the published state of the app might not be correct, the status will show as published

packages/server/src/api/controllers/application.ts
```diff
@@ -450,37 +451,51 @@ export const revertClient = async (ctx: any) => {
   ctx.body = app
 }
 
-const destroyApp = async (ctx: any) => {
```
Rory-Powell, line 453: For peace of mind around dev vs prod app ids, it'd be nice to see tests like:

```ts
describe("DELETE /api/applications/:id", () => {
  it("deletes prod and dev apps using dev app id", () => {})
  it("deletes prod and dev apps using prod app id", () => {})
})

describe("POST /api/applications/:id/publish", () => {
  it("publishes with dev app id", () => {})
  it("publishes with prod app id", () => {})
})

describe("POST /api/applications/:id/unpublish", () => {
  it("unpublishes with dev app id", () => {})
  it("unpublishes with prod app id", () => {})
  it("rejects when app is not published", () => {})
})
```

shogunpurple: Yep, they are in progress - with each of the different scenarios. Just been pushing changes incrementally to make sure I'm not breaking the existing endpoints

Rory-Powell: 👍 👍

Rory-Powell (approved):

PR #8849, opened 2022-11-28
Add namespace to Helm chart templates

tomberget:
## Description

We are trying to use this Helm chart for deployment using *ArgoCD*. However, since all templates are lacking `metadata.namespace`, we are required to change how we define namespaces for the deployment of this chart alone.

Following best practice as visible in the [ingress-nginx](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx) and [kube-prometheus-stack](https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack) template sections (though *kube-prometheus-stack* actually uses a helper), adding the `metadata.namespace` reference everywhere required will ensure that the default way of deploying *Budibase* will work, also using ArgoCD.

The only possible issue I've found, is that it seems that the dependency `couchdb` does not take the `namespace` provided. I assume that is something that needs to be fixed upstream, if at all necessary. The `ingress-nginx` dependency takes the namespace like a champ :) 

Addresses: 
- #7062

## How it is tested:

For a dry-run test, to verify that the manifests can be generated:

```bash
helm install -n anewbudins budibase charts/budibase --set ingress.nginx=true,services.couchdb.enabled=true --dry-run
helm install -n anewbudins budibase charts/budibase --set ingress.nginx=false,services.couchdb.enabled=false --dry-run
```

Linting:

```bash
helm lint charts/budibase --with-subcharts
```

## Side note

I also saw that the `charts/budibase/README.md` file was using one reference to install budibase, and a different reference to remove it. I also fixed that, so that the two name references match.

### DIFF
--- a/charts/budibase/Chart.yaml
+++ b/charts/budibase/Chart.yaml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/Budibase/budibase
   - https://budibase.com
 type: application
-version: 0.2.11
+version: 0.2.12
 appVersion: 1.0.214
 dependencies:
   - name: couchdb
--- a/charts/budibase/Chart.yaml
+++ b/charts/budibase/Chart.yaml
@@ -11,8 +11,8 @@ sources:
   - https://github.com/Budibase/budibase
   - https://budibase.com
 type: application
-version: 0.2.12
-appVersion: 1.0.214
+version: 0.0.0
+appVersion: 0.0.0
 dependencies:
   - name: couchdb
     version: 3.6.1
--- a/charts/budibase/Chart.yaml
+++ b/charts/budibase/Chart.yaml
@@ -11,7 +11,9 @@ sources:
   - https://github.com/Budibase/budibase
   - https://budibase.com
 type: application
+# populates on packaging
 version: 0.0.0
+# populates on packaging
 appVersion: 0.0.0
 dependencies:
   - name: couchdb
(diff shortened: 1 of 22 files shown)


tomberget:

brunowego:
Hi @tomberget can you fix the conflicts? Thanks!

tomberget:
> Hi @tomberget can you fix the conflicts? Thanks!

Ive reset the versions to `0.0.0`, and added the commented text. The conflict seems to be resolved :)

mzac:
@brunowego Any chance this can be merged soon? I'm also trying to deploy Budibase with Helm & ArgoCD into a different namespace.  It doesn't work since the proxy service is trying to talk to minio in the 'budibase' namespace that doesn't exist.

@tomberget Good work!

Thanks!

shogunpurple:
> The only possible issue I've found, is that it seems that the dependency couchdb does not take the namespace provided. I assume that is something that needs to be fixed upstream, if at all necessary. The ingress-nginx dependency takes the namespace like a champ :)

@tomberget do you have any issues running on ArgoCD due to the above problem?

tomberget:
@shogunpurple, since this PR has not been merged and released, I have not actually tried. We deploy budibase as part of an applicationset where namespace was templated to "appname-environment". We had to remove that template, and just go by the `helmCharts.namespace` from the individual `Kustomization` files instead.

Currently, that namespace is set to `budibase`, and it is deploying. With the changes in this PR, it should also work with using the templated namespace from the applicationset. But due to time constraints, I have not tested whether CouchDB will be a problem.

However, I am in favor of having control of all my third party applications, so I mainly want to control the CouchDB chart myself anyways (along with Redis and Ingress-Nginx), and as such it should not be that big of a deal (unless I then get in the exact same problem 🤪 )

PR #8857, opened 2022-11-29
Remove ability to edit the 'Sample Data' datasource name

deanhannigan:
## Description

The ability to edit the name of the datasource in the builder has been removed. The implementation was causing table entity information to be populated into the default Sample Data datasource. This was leading to the scenario where the datasource could eventually no long be deleted.

This behaviour was extended to the backend to ensure that this wouldn't reoccur if the functionality were reenabled in the builder.

**Note** 
Any applications created with a `Sample Data` datasource that had its name changed would need to be deleted entirely

Addresses: 
- https://github.com/Budibase/budibase/issues/8820


shogunpurple (approved):
LGTM!

Issue #8031, opened 2022-09-29
Public accessible app - navigation menu is missing


hheedfeld:
### Discussed in https://github.com/Budibase/budibase/discussions/7922

<div type='discussions-op-text'>

<sup>Originally posted by **hheedfeld** September 23, 2022</sup>
Hi,

I've created a public accessible app, which is working very well, but I'm not able to make the navigation menu visible for unauthenticated users. What am I missing?

Thank for any hints,
Henning</div>

melohagan:
![Screenshot 2022-10-03 at 14 40 39](https://user-images.githubusercontent.com/101575380/193592020-27fe7eb0-2d07-410d-a644-f1056d28d9d6.png)

![Screenshot 2022-10-03 at 14 41 06](https://user-images.githubusercontent.com/101575380/193592118-9c782dc9-f759-4a78-9ba7-54da805d75fe.png)

**Incognito**
![Screenshot 2022-10-03 at 14 41 41](https://user-images.githubusercontent.com/101575380/193592268-0e99ec8c-56d6-48fd-8556-531b763327e0.png)

**Authenticated**
![Screenshot 2022-10-03 at 14 42 23](https://user-images.githubusercontent.com/101575380/193592422-7374a037-c11e-440f-b187-15954ed2110b.png)

hheedfeld:
@shogunpurple Should it be fixed in 2.0.26 already?

cheers,
Henning

melohagan:
Hey @hheedfeld This was closed by accident. In future could you please add reproduction steps and screenshots (I've provided this time).

PR #8870, opened 2022-11-29
Public App Missing Nav Menu

Ghrehh:
## Description
Public apps weren't displaying the nav menu. This seems to be because the default role applied to un-authed users is 0, which doesn't correspond to any defined role.

Addresses: 
https://github.com/Budibase/budibase/issues/8031

### DIFF
--- a/packages/client/src/stores/derived/currentRole.js
+++ b/packages/client/src/stores/derived/currentRole.js
@@ -1,4 +1,5 @@
 import { derived } from "svelte/store"
+import { Constants } from "@budibase/frontend-core"
 import { devToolsStore } from "../devTools.js"
 import { authStore } from "../auth.js"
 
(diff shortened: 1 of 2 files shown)


aptkingston:
Had a quick look at this myself and the root issue is that the `currentRole` store is returning `undefined` for public users. If we just add a quick ` || Constants.Roles.PUBLIC` to the end of `client/src/stores/derived/currentRole.js` (and `import { Constants } from "@budibase/frontend-core"`) then that should fix the issue!

Changing these priorities would mess up screen sorting for users that use custom roles since it would consider public and custom roles the same, so it would be better to leave that untouched :+1:

Ghrehh:
@aptkingston Done, thanks for the explanation. Could you take another look at this when you get a chance please?

aptkingston (approved):
LGTM!

Issue #8781, opened 2022-11-23
MySQL BigInt Primary Key


andr-04:
**Hosting**
- Self
  - Method: <method> docker single image
  - Budibase Version: <version> 2.1.28

**Describe the bug**
MySQL as a data source distorts BigInt datatype because of int53 problem of JavaScript. I've found the distortion comes from a backend (for example, https://portal/api/***/search entrypoint returns the wrong data). There are only 2 options for a column type of BigInt field: Number and Boolean.

**To Reproduce**
1. Create MySQL database.
2. Create the table for the MySQL database.
3. Create a field with the type BigInt(20).
4. Add a record with the value 12345678901234567890.
5. Create Budibase application.
6. Attach MySQL database as a datasource to the Budibase application.
7. See 12345678901234567000 instead of 12345678901234567890.

**Expected behavior**
The safest way is to use strings for BigInt values.

**Desktop:**
 - OS: macOS 13.0
 - Browser: Chrome 107.0.5304.110 (arm64)

melohagan:
There is also this other bug to take into consideration: https://github.com/Budibase/budibase/issues/6000

Users wanted the ability to establish relationships between integer columns and big integer columns. 
If the type of big integer was treated as a string, then it may not allow you to select the columns of different types when creating the relationship.

andr-04:
It looks like the problem with BigInt is still here. I try to use a "Repeater block" component with a filter by BigInt field and was stuck with the problem that no rows found in some cases. Is it the problem with the "Repeater block" component or with something else?

@melohagan can you check it, please?

melohagan:
Hey @andr-04 

I've checked the latest version **v2.6.9** and I can't find any issues myself. If you can reproduce a specific problem please feel free to raise another bug issue!

andr-04:
@melohagan, #10703. Just now I checked it on v2.6.11, and it's the same.

melohagan:
Hey @andr-04 Could you provide the schema for your MySQL table (the CREATE SQL script) as well as a video recording demonstrating the problem please?

PR #8871, opened 2022-11-30
MySQL BigInt as string

andr-04:
## Description
MySQL as a data source distorts BigInt datatype because of int53 problem of JavaScript. The patch interprets BigInt as a string.

Addresses: 
- #8781

### DIFF
--- a/packages/server/src/integrations/mysql.ts
+++ b/packages/server/src/integrations/mysql.ts
@@ -146,7 +146,8 @@ class MySQLIntegration extends Sql implements DatasourcePlus {
         if (
           field.type == "DATETIME" ||
           field.type === "DATE" ||
-          field.type === "TIMESTAMP"
+          field.type === "TIMESTAMP" ||
+          field.type == "LONGLONG"
         ) {
           return field.string()
         }


andr-04:

melohagan:
Hey @andr-04 

Should it not be `field.type === "BIGINT"` ?
https://dev.mysql.com/doc/refman/8.0/en/integer-types.html

I don't see the LONGLONG date type mentioned in the docs.

andr-04:
> Hey @andr-04
> 
> Should it not be `field.type === "BIGINT"` ? https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
> 
> I don't see the LONGLONG data type mentioned in the docs.

No. https://github.com/sidorares/node-mysql2/blob/master/lib/constants/types.js

andr-04:
@melohagan I've just tried to apply your [suggestion](https://github.com/Budibase/budibase/issues/8872#issuecomment-1331796511) instead of and with `LONGLONG` both, but it doesn't help. It doesn't break anything (on the first view), but filtering requests from the front-end go to the server as distorted numbers. So #8872 probably affects other things and should be done separately.

melohagan:
> @melohagan I've just tried to apply your [suggestion](https://github.com/Budibase/budibase/issues/8872#issuecomment-1331796511) instead of and with #8871 both, but it doesn't help. It doesn't break anything (on the first view), but filtering requests from the front-end go to the server as distorted numbers. So #8872 probably affects other things and should be done separately. At least, PostgreSQL doesn't have #8781 issue, but has #8872 one.

Thanks for taking a look at this. I don't want to merge this one in because the id column in SQL is incorrectly showing **Number** as its type in Budibase despite being cast as a string.
I am working on this now though, and the code you have here is part of the fix.

melohagan:
Closing in favour of: https://github.com/Budibase/budibase/pull/8877

PR #8875, opened 2022-11-30
Fixes for table and view editing

deanhannigan:
## Description

Some fixes to correct some issues around editing table and view names. Typically centred around stale or mutated state.

- Renaming a view would create a duplicate with the updated name. This would continue with every name change
- Renaming a view back to its original name would throw and exception.
- Altering the table name in the form would mutate the table name held in state. If a user cancelled a name change, they could unintentionally persist any changes if they were to alter any views.

## Screenshots

Duplicating views

duplicatingViews

Rename Causing Exception

renameCrashViews

Table Name Mutated

tableNameMutate


shogunpurple (approved):
LGTM! nice fix

PR #8877, opened 2022-11-30
Treat bigint as string

melohagan:
## Description
Continuing work started by @andr-04 

As pointed out, it is best to treat BigInt as string type because there is no guarantee that JS max int will be large enough.
 
As some users required relationships to work between Integer and Big Integer columns, I updated the create relationship validation to handle this.

Also fixed a bug with `convertSqlType`. Previously **bigint** was matching on **int** via the includes and was exiting too early. It now performs all matches and chooses the longest match, i.e. 

> 'bigint(20)' includes both 'int' and 'bigint' so choose the longest: 'bigint'  

Sorting and filtering of BigInt columns will now be treated as text instead of number. This may affect some users, but it can't be assumed that the BigInt value will be within the JS max int range, so it is safer to treat it as text.

Addresses: 
- https://github.com/Budibase/budibase/issues/8781


andr-04:
JS has also `BigInt` type. Will it help solve the sorting problem?

melohagan:
> JS has also `BigInt` type. Will it help solve the sorting problem?

I had absolutely no idea that JavaScript actually had a BigInt type. TIL! 

I don't think it would be easy to use within Budibase currently, because there are various places where the field type is checked against "number" and "string". I'd have to find all of those occurrences and add an if statement for BigInt type which could be a lot of work.

mike12345567 (approved):
LGTM - Agree with you @melohagan - don't think we can use the built in `BigInt` because we have no support for it anywhere else - it would be a large rework.

PR #8884, opened 2022-12-01
Multifield Input Autoselect All Fields

Ghrehh:
## Description
Adds an optional flag to auto populate multifield inputs with columns. As well as being generally handy this helps resolve a UX issue Mihail pointed out where the Row Explorer block seems broken unless the user selects fields to display, and it's not necessarily obvious that needs to be done.

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -25,7 +25,7 @@ import {
   DB_TYPE_INTERNAL,
   DB_TYPE_EXTERNAL,
 } from "constants/backend"
-import { getSchemaForDatasource } from 'builderStore/dataBinding'
+import { getSchemaForDatasource } from "builderStore/dataBinding"
 
 const INITIAL_FRONTEND_STATE = {
   apps: [],
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -1042,15 +1042,19 @@ export const getFrontendStore = () => {
           if (component[name] === value) {
             return false
           }
-          if (name === "dataSource") {
+
+          const settings = getComponentSettings(component._component)
+          const updatedSetting = settings.find(setting => setting.key === name)
+
+          if (
+            updatedSetting.type === "dataSource" ||
+            updatedSetting.type === "table"
+          ) {
             const { list: allTables } = get(tables)
             const { schema } = allTables.find(
               ({ _id }) => _id === value.tableId
             )
             const columnNames = Object.keys(schema)
-
-            const settings = getComponentSettings(component._component)
-
             const multifieldKeysToSelectAll = settings
               .filter(setting => {
                 return setting.type === "multifield" && setting.selectAllFields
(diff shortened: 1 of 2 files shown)


packages/builder/src/builderStore/store/frontend.js
```diff
@@ -523,7 +523,9 @@ export const getFrontendStore = () => {
         // Generate default props
         let props = { ...presetProps }
         settings.forEach(setting => {
-          if (setting.defaultValue !== undefined) {
+          if (setting.type === "multifield" && setting.selectAllFields) {
+            props[setting.key] = Object.keys(defaultDatasource.schema)
```
aptkingston, line 527: We might want to change this to `Object.keys(defaultDatasource?.schema || {})` just since we can't guarantee a datasource was set, and this will throw an error otherwise.

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1040,6 +1042,30 @@ export const getFrontendStore = () => {
...
+          const settings = getComponentSettings(component._component)
+          const updatedSetting = settings.find(setting => setting.key === name)
+
+          if (
+            updatedSetting.type === "dataSource" ||
+            updatedSetting.type === "table"
+          ) {
+            const { list: allTables } = get(tables)
+            const { schema } = allTables.find(
+              ({ _id }) => _id === value.tableId
```
aptkingston, line 1055: It's possible to set a `table` type setting back to `undefined` - so we'll need optional chaining here on `value.tableId`.

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1040,6 +1042,30 @@ export const getFrontendStore = () => {
...
+
+          const settings = getComponentSettings(component._component)
+          const updatedSetting = settings.find(setting => setting.key === name)
+
+          if (
+            updatedSetting.type === "dataSource" ||
+            updatedSetting.type === "table"
+          ) {
+            const { list: allTables } = get(tables)
+            const { schema } = allTables.find(
```
aptkingston, line 1054: Settings of type `dataSource` don't need to be tables, so searching the tables list here will return `undefined` for any non-table datasource. That mean's we won't update the schema for these (which is probably fine) but it will throw an error here when it attempt to destructure out `schema` of `undefined.

A better thing to do here would actually be using the existing `getSchemaForDatasource` function in `dataBinding.js`, and just pass in the current value. This will get the correct schema for any type of data source!

We don't need to use this up above at component creation since we know that all tables have a `schema` property directly on them, so we can just use that.

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1040,6 +1042,30 @@ export const getFrontendStore = () => {
...
+              ({ _id }) => _id === value.tableId
+            )
+            const columnNames = Object.keys(schema)
```
aptkingston, line 1057: We should use `schema || {}` to avoid errors being throw if `schema` was nullish.

aptkingston (changes requested):
Few safety checks needed to avoid errors, but logic looks good! Suggested one refactor to make life easier using an existing utility as well.

Ghrehh:
@aptkingston Should be all good now if you could take another look when you get time please.

aptkingston (approved):
LGTM!

PR #8890, opened 2022-12-01
Add env var for tls-reject-unauthorised to helm chart

jonnymccullagh:
## Description
Problem reported by @simoncramer who needed an environment variable set for `NODE_TLS_REJECT_UNAUTHORIZED` to ignore intermediate certificate errors when connecting a BB app to an API endpoint. The variable is optional in the helm chart but can be added by adding ` tlsRejectUnauthorized: 0` line into the `services:` section of the `values.yaml`
e.g.

```
services:
  budibaseVersion: latest
  dns: cluster.local
  tlsRejectUnauthorized: 0
```

Addresses: 
- https://github.com/Budibase/budibase/discussions/8808#discussioncomment-4262077

### DIFF
--- a/charts/budibase/templates/minio-service-deployment.yaml
+++ b/charts/budibase/templates/minio-service-deployment.yaml
@@ -42,6 +42,11 @@ spec:
             secretKeyRef:
               name: {{ template "budibase.fullname" . }}
               key: objectStoreSecret
+        {{ if .Values.services.tlsRejectUnauthorized }}
+        - name: NODE_TLS_REJECT_UNAUTHORIZED
+          value: {{ .Values.services.tlsRejectUnauthorized }}
+        {{ end }}
+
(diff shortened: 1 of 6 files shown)


charts/budibase/templates/minio-service-deployment.yaml
```diff
@@ -42,6 +42,11 @@ spec:
             secretKeyRef:
               name: {{ template "budibase.fullname" . }}
               key: objectStoreSecret
+        {{ if .Values.services.tlsRejectUnauthorized }}
```
shogunpurple, line 45: Don't think we need this in the proxy, redis or minio deployments since it's Node specific

shogunpurple (commented):
Just need to remove the variable from proxy, redis and minio deployments then we are good to go!

PR #8891, opened 2022-12-01
add optional annotations to pods for logging format

jonnymccullagh:
## Description
For sending logs to elastic we need to tell filebeat about the format of our log files. For k8s this is done via annotations on the pods i.e.
```
   co.elastic.logs/multiline.type: pattern
   co.elastic.logs/multiline.pattern: '^[[:space:]]'
   co.elastic.logs/multiline.negate: false
   co.elastic.logs/multiline.match: after
```
To cater for this we need to allow these values to be set within the `values.yaml` file for our helm chart. If `logAnnotations` exists in the `values.yaml` file then the annotations are added to the app, worker and proxy deployments as shown below:
```
{{ if .Values.globals.logAnnotations }}
{{ toYaml .Values.globals.logAnnotations | indent 4 }}
{{ end }}
```
Opinions welcome and should we add to redis and minio pods?

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -4,6 +4,9 @@ metadata:
   annotations:
     kompose.cmd: kompose convert
     kompose.version: 1.21.0 (992df58d8)
+{{ if .Values.globals.logAnnotations }}
+{{ toYaml .Values.globals.logAnnotations | indent 4 }}
+{{ end }}
   creationTimestamp: null
   labels:
     io.kompose.service: app-service
--- a/charts/budibase/templates/proxy-service-deployment.yaml
+++ b/charts/budibase/templates/proxy-service-deployment.yaml
@@ -4,6 +4,9 @@ metadata:
   annotations:
     kompose.cmd: kompose convert
     kompose.version: 1.21.0 (992df58d8)
+{{ if .Values.globals.logAnnotations }}
+{{ toYaml .Values.globals.logAnnotations | indent 4 }}
+{{ end }}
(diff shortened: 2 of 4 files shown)


shogunpurple (approved):
LGTM! Worth a test on the dev cluster just to make sure.

Rory-Powell (approved):
LGTM 👍 For the minio and redis pods question, we won't run them in prod. So if we were to add the log annotations there it would be for the benefit of the QA, Dev and preprod envs. Probably the same for nginx and couch. 

I'd like to have all those logs in elastic, it'll just likely be a bit different for prod vs the other envs on how we get them there

jonnymccullagh:
> LGTM! Worth a test on the dev cluster just to make sure.

Could be difficult as these changes are needed in the published helm chart - kinda what we were discussing a few days ago :-(

jonnymccullagh:
> minio and redis pods question, we won't run them in prod
Yeah I was more thinking about people using our chart who are possibly using minio (not S3) and redis (not elasticache). I don't want to limit logging options for others.

jonnymccullagh:
> like to have all those logs in elastic
The logs are already being sent - just the formatting sucks for multiline traces so this formatting change is to address that

Rory-Powell:
> Could be difficult as these changes are needed in the published helm chart

We should still be able to rest this with the existing release env no? When we deploy to that env we use the charts from src

jonnymccullagh:
> We should still be able to test this with the existing release env no?

Yeah I'll have to amend the EKS stuff in dev to get it to pull the budibase repo and use the Helm charts on a branch

Issue #7810, opened 2022-09-16
[BUDI-7194] Add filtering to Relationship Picker


martkaczmarek:
When I add options picker or multi-select picker with a dataset above it, I can filter a data provider to limit the options. 
But when I add a relationship picker, I can't do that - I have to choose from all the options pulled from a respective table. 
This becomes problematic for a few reasons:

1. End user can be overwhelmed by amount of options to choose from, many of which may not be necessary for a particular form
2. Let's say the app architect would like to limit the choice range, for example based on end user's role (basic user has 3 options to select from, power user has 5) - they can't
3. Because of bug #7773 the form will just stop working properly on large datasets, and filtering would be a way to get around it (limit amount of options to choose from)

There are workarounds described in #4596 and #4579, but they only work on single-option picker, and also I feel like if we have a specific input field for dataset relations, it should be done in a way that doesn't need workarounds. 

<sub>[BUDI-7194](https://linear.app/budibase/issue/BUDI-7194/add-filtering-to-relationship-picker)</sub>

rafikApp:
I have the same issue

gerickt:
same problem here

martkaczmarek:
Resolved in #8897

PR #8897, opened 2022-12-01
Relationship picker filter

melohagan:
## Description
Users wanted to be able to add a filter to the relationship picker because previously all possible relationships were being returned, which could be a very long list.

By default there is no filter and all relationships as determined by the field name will be returned.
If you select a relationship datasource, the option to add a filter will appear. 

Tested on both SQL Server and Internal DB.

Addresses: 
- https://github.com/Budibase/budibase/issues/7810

## App Export
[Relationship filter-export-1669914030877.tar.gz](https://github.com/Budibase/budibase/files/10134461/Relationship.filter-export-1669914030877.tar.gz)

## Screenshots
*No filter: names A-Z*
Screenshot 2022-12-01 at 17 01 47

*Only relationship datasources are listed*
Screenshot 2022-12-01 at 17 02 42

*Add filter*
Screenshot 2022-12-01 at 17 03 05

*Result*
Screenshot 2022-12-01 at 17 03 57


mike12345567 (approved):
This looks good to me - its an important improvement in terms of UX and also performance - although I believe that @aptkingston would be able to best say if it can be merged.

aptkingston:
Had a look at this one Mel since it's a really important feature.

The main issue I have is that you have to define the relationship datasource, which doesnt make sense. You have already picked a relationship field to bind to, so you know what it is...

Off the top of my head, some issues I can see:
1. its redundant to pick a datasource because we can already work it out
2. you could pick a different datasource to the field you're actually bound to which is a mind melt and should never be done
3. the only reason you need to pick a datasource is to get the correct fields listed when defining the filter, but a unique settings component would fix this

Cleaner implementation:
- split the existing FilterEditor into 2 components: maybe a DatasourceFilterEditor and BaseFilterEditor (just splitting the reusable html into its own component)
- use a new, unique settings type here relationshipFieldFilter which uses a new component RelationshipFieldFilterEditor
- use custom logic to determine the form schema, get the relationship info, get the related table schema
- pass the schema into the reusable BaseFilterEditor

Are you happy if I jump on this branch and implement this? I can build on top of what you already have here. Don't want to step on your toes but I can have a clear implementation path in my head and the work you've already done doesn't need to be wasted.

I might also investigate adding pagination and/or search to the picker as well, since they would help. Pagination would actually be essential as otherwise you'll only ever get one page of data due to using the search endpoint, meaning your relationship options would be very limited.

melohagan:
> The main issue I have is that you have to define the relationship datasource, which doesn't make sense. You have already picked a relationship field to bind to, so you know what it is, but I can see why you've had to do that - to get the filter editor to work.
> 
> Are you happy if I jump on this branch and implement this? I can build on top of what you already have here. Don't want to step on your toes but I can have a clear implementation path in my head and the work you've already done doesn't need to be wasted.
> 
> I might also investigate adding pagination and/or search to the picker as well, since they would help. Pagination would actually be essential as otherwise you'll only ever get one page of data due to using the search endpoint, meaning your relationship options would be very limited.

I had realised after the fact that the implementation wasn't ideal, so if you want to go ahead and implement the improvements that would be amazing - definitely not stepping on any toes! 😀

melohagan:
Closing as the work has been done here: https://github.com/Budibase/budibase/pull/10329

Issue #7754, opened 2022-09-13
Changing a Query (MySQL) doesnt update Data Providers bound to it


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7682

<div type='discussions-op-text'>

<sup>Originally posted by **poirazis** September  7, 2022</sup>
Reproduce : 

Create a MySQL connection.

(assume the below tables / fields are defined)

Create a Query with a binding -> "SELECT * FROM members WHERE dept_id = {{ dept_id }} "
Create a Screen
Insert Data Provider
Set Data Provider Source to above query

Go back to "Data"
Change the Query, example "SELECT name, surname FROM members WHERE dept_id = {{ dept_id }} "
Go back to the Screen
Check Data Provider Source 
In the Panel that will open below, you will see the previous definition of the query.
In order to update, you need to select the query again in the Data Providers source drop down.
</div>

aptkingston:
There's always an aspect of this that requires manual updating - for example if you added, removed or changed the name of a parameter. I'm guessing the results are correctly from the updated query, but it's just the params (and this text displayed) that aren't?

melohagan:
Exactly, the functionality is perfectly fine, it's just the query text that doesn't update so it's a little confusing.

![Screenshot 2022-09-13 at 16 12 35](https://user-images.githubusercontent.com/101575380/189939217-05728067-2382-4d1b-93fd-083f2645a02f.png)

*I then save this*
![Screenshot 2022-09-13 at 16 12 59](https://user-images.githubusercontent.com/101575380/189939314-a9be9051-d55e-4bdc-86a4-8ae321afce01.png)

Then back to the design, the correct query is executed, but the query text (from clicking on the cog icon) looks the same as it did before in the first screenshot

mahxds:
yeah this issuse was quite confusing in the beginning. After changing the Query i always reselect the query in all dataproviders/querys in the design.

would be nice if this wasnt necessary

PR #8911, opened 2022-12-02
Update custom sql query in design when changed in the data section

melohagan:
## Description
SQL and JSON query boxes now update in the Design section when they are changed in the Data section.
I also noticed that the MongoDB Aggregate pipeline needed to be disabled within this view.

Addresses: 
- https://github.com/Budibase/budibase/issues/7754


shogunpurple (approved):
LGTM! Nice work

Issue #8882, opened 2022-11-30
Constraint violation on many to many join table


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.41
  - App Version:  2.1.41

**Describe the bug**
Consider the following setup: Many employees -> Many skills
Two employees have many skills. Some overlap, some don't.
Try to change an employees skills through an auto-generated screen will result in constraint violation on the join table.

**To Reproduce**
1. Setup an **Employee**, **Skills** and **Join** table for a many to many relationship in Postgres. (scripts below)
2. Fetch tables in Budibase
3. Setup relationships (Many -> Many)
4. Create auto-generated screens for the **Employee** table
5. Preview app, and unselect a skill for an employee and save
6. See error

**Expected behavior**

**Screenshots**
*Remove 'Docker' and save*
![Screenshot 2022-11-30 at 16 51 05](https://user-images.githubusercontent.com/101575380/204858798-fe789a0d-d214-4a83-9e1b-30f22dfeb9e5.png)

**SQL Scripts**
```sql
CREATE TABLE IF NOT EXISTS public."Employee"
(
    id integer NOT NULL,
    name text COLLATE pg_catalog."default",
    CONSTRAINT "Employee_pkey" PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS public."Employee"
    OWNER to postgres;

INSERT INTO public."Employee" ("id", "name") VALUES (1, 'Alice');
INSERT INTO public."Employee" ("id", "name") VALUES (2, 'Bob');
```

```sql
CREATE TABLE IF NOT EXISTS public."Skills"
(
    id integer NOT NULL,
    name text COLLATE pg_catalog."default",
    CONSTRAINT "Skills_pkey" PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS public."Skills"
    OWNER to postgres;

INSERT INTO public."Skills" ("id", "name") VALUES (1, 'Docker');
INSERT INTO public."Skills" ("id", "name") VALUES (2, 'Microservices');
INSERT INTO public."Skills" ("id", "name") VALUES (3, 'Kubernetes');
INSERT INTO public."Skills" ("id", "name") VALUES (4, 'Spring');
```

```sql
[...]

PR #8915, opened 2022-12-02
Fix/8882

mike12345567:
## Description
This brings a fix for editting many to many relationships, as well as adding a lot more checking around what can and can't be a junction table. For Budibase to be opinionated about generating many to many relationships, the table needs to be a well defined junction table - this is discussed in the issue.

The majority of this issue was updating table definitions to include the presence constraint, which means that Budibase is now able to pre-populate the `required` attribute of a column based on a `NOT NULL` option in the SQL DDL of a table schema.

I've also added the configuration used to create this bad table scenario to the integration scripts in the server, for future use in a test case for this.

Addresses: 
- #8882 

## Screenshots
How the error appears for the table configuration described in the issue.
image

### DIFF
--- a/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
+++ b/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
@@ -45,6 +45,23 @@
 
   const touched = writable({})
 
+  function invalidThroughTable({ through, throughTo, throughFrom }) {
+    // need to know the foreign key columns to check error
+    if (!through || !throughTo || !throughFrom) {
+      return false
+    }
+    const throughTable = plusTables.find(tbl => tbl._id === through)
+    const otherColumns = Object.values(throughTable.schema).filter(
+      col => col.name !== throughFrom && col.name !== throughTo
+    )
+    for (let col of otherColumns) {
+      if (col.constraints?.presence && !col.autocolumn) {
+        return true
+      }
+    }
+    return false
+  }
+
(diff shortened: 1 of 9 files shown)


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -59,6 +76,10 @@
     if ($touched.through && isMany && !fromRelate.through) {
       errObj.through = tableNotSet
     }
+    if ($touched.through && invalidThroughTable(fromRelate)) {
+      errObj.through =
+        "Table contains non-nullable columns which aren't generated"
```
shogunpurple, line 81: I would potentially flip this to make it easier to understand:

`Table PK must have auto generated` or similar.

shogunpurple (approved):
LGTM - just think the error message could be a little clearer. Good job

packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
mike12345567: In this case its specifically about non-nullable fields, its not to do with the PK. For example you could have a whole bunch of non-nullable fields in your junction table, Budibase will never be able to fill them in - creation of rows will always fail.

shogunpurple: Gotcha, in that case maybe:
`Ensure all columns in table are nullable or auto generated`

Issue #8896, opened 2022-12-01
Automation history(logs) disappear after unpublishing app


MihailHadzhiev2022:
1. Create an app
2. Set an automation, CronJob 1min, Row created
3. Publish the app
4. Wait 1 min, and navigate to automation history.
5. 1 automation is displayed
6. Unpublishe the app 
7. Navigate back to Automation History

Expected result:  Automation history logs should be displayed
Actual result: Automation history disappear 

https://www.loom.com/share/8897cece43a84359b2724e9def9c9a89

PR #8937, opened 2022-12-05
Automation logs availability after unpublish

mike12345567:
## Description
Fix for #8896 - the automation logs were already being sync'd to the dev database, but when unpublished they are in-accessible. Some minor updates to make sure that before unpublishing there is a sync, and then most of the changes are in pro.


packages/server/src/sdk/app/applications/general.ts
```diff
@@ -0,0 +1,19 @@
+const URL_REGEX_SLASH = /\/|\\/g
+
+export function getAppUrl(ctx: {
+  request: { body: { name?: string; url?: string } }
```
Rory-Powell, line 4: There's a pattern introduced here: https://github.com/Budibase/budibase/pull/8883/files#diff-8e575ce9e55cd6e844436a203115309c30a986350847a75c080c5735ae526ada

For defining the body type using generics. This can be ported here to allow for saying

```ts
export function getAppUrl(ctx: BBContext<SomeAppType>) { }
```
Where `ctx.request.body` will then have the fields from the types available. 

However I don't think that's what we should do here, the `SomeAppType` doesn't really exist, it's partial fields from:
- POST /api/applications
- PUT /api/applications/:appId
- Constructing the ctx for the migration


As a general rule I think It's best we keep `ctx` out of the sdk so it's framework agnostic and typed as much as possible. Instead changing the method signature here to:

```ts
export function getAppUrl(opts: { name?: string, url?: string) { }
```

We could then even remove the manual construction of context from the migration, which points to this being the right way to go

packages/server/src/sdk/app/applications/general.ts
```diff
@@ -0,0 +1,19 @@
+const URL_REGEX_SLASH = /\/|\\/g
```
Rory-Powell, line 1: The `general` term stands out a bit, it's not a name we've used a lot. `utils` feels more familiar. Or this file could become the beginning on the applications sdk itself, i.e. the main file we use for create, read, update etc, when it then would be named more like `/sdk/applications/applications.ts`

Rory-Powell: or a few other suggestions

`/sdk/applications/`:
- `appService.ts`
- `appSdk.ts` 
- `service.ts`
- `sdk.ts` 

we should probably agree something and get a doc going on confluence for explaining project structure and naming conventions for new hires

packages/server/src/sdk/app/applications/sync.ts
```diff
@@ -0,0 +1,53 @@
+import env from "../../../environment"
+import { syncGlobalUsers } from "../../../api/controllers/user"
```
Rory-Powell, line 2: The sdk package reaching out into controllers is something we'll want to avoid. I think the right thing to do here is to populate a very minimal `sdk/users` package with the `syncGlobalUsers` added to it

Rory-Powell (approved):

PR #8943, opened 2022-12-06
single image nginx logs to stdout

jonnymccullagh:
## Description
I had this change in another [PR](https://github.com/Budibase/budibase/pull/8064/) that is now obsolete so re-adding the missing bits here and I will close the old PR.
- Nginx user needs to be added to the `tty` group to allow permissions to write to /dev/stdout 
- Set nginx log location to /dev/stdout /dev/stderr


Rory-Powell (approved):
🧙‍♂️

PR #8954, opened 2022-12-06
Api tests user management

Pedro-Revez-Silva:
## Description
Tests User creation, updates, deletion, changed of role and permissions(roles) for apps.

## Notes:
There is a flaky test this PR fixed in the side of the applications. However I left a comment that further work is needed in order to make the test better.


qa-core/src/tests/internal-api/userManagement/userManagement.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+        await config.userManagement.changeUserInformation(body)
+
+        const [changedUserInfoResponse, changedUserInfoJson] = await config.userManagement.getUserInformation(createUserJson.created.successful[0]._id)
+        expect(changedUserInfoJson.roles[app.appId?.toString() || ""]).toBeDefined()
+        expect(changedUserInfoJson.roles[app.appId?.toString() || ""]).toEqual("BASIC")
+
+    })
+
+    it("Add ADMIN user to app", async () => {
```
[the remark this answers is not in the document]
shogunpurple: Really? All of the above should be locked down at the API level. Are there some examples?

qa-core/src/config/internal-api/fixtures/userManagement.ts
```diff
@@ -0,0 +1,81 @@
...
+        },
+        builder: {
+            global: false
+        }
+    })
+
+}
+
+export const generateInviteUser = (): Object[] => {
+    //const randomId = generator.guid();
```
shogunpurple, line 52: remove comment

qa-core/src/config/internal-api/TestConfiguration/userManagement.ts
```diff
@@ -0,0 +1,117 @@
...
+import { responseMessage } from "../fixtures/types/responseMessage"
+
+export default class UserManagementApi {
+    api: InternalAPIClient
+
+    constructor(apiClient: InternalAPIClient) {
+        this.api = apiClient
+    }
+
+    async searchUsers(): Promise<[Response, User[]]> {
```
shogunpurple, line 13: We could update these method names to be more inline with the rest of the API clients:
- `search`
- `fetchAll`
- `deleteMultiple`

etc, because they will be used in the tests like so:

`users.search` instead of `users.searchUsers`, which is a bit redundant.

qa-core/src/config/internal-api/TestConfiguration/applications.ts
```diff
@@ -32,13 +33,21 @@ export default class AppApi {
...
     const response = await this.api.get("/routing/client")
+    expect(response).toHaveStatusCode(200)
     const json = await response.json()
     const publishedAppRenders = Object.keys(json.routes).length > 0
-    expect(response).toHaveStatusCode(200)
     expect(publishedAppRenders).toBe(true)
     return [response, publishedAppRenders]
   }
 
+  async failRender(): Promise<[Response, responseMessage]> {
```
shogunpurple, line 43: Doesn't seem like this is called anywhere

shogunpurple (approved):
LGTM!

qa-core/src/config/internal-api/TestConfiguration/applications.ts
Pedro-Revez-Silva: Thank you for catching that.

Issue #8950, opened 2022-12-06
Very long screen names cause side panel to be unaccessible


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.42
  - App Version: 2.1.42

**Describe the bug**
Given a very long screen name (which could occur if using a lot of URL variables for example), the screen details panel is not visible - it is pushed out of sight!

**To Reproduce**
1. Add a screen with a very long name
2. See that you cannot edit screen details - importantly you can't shorten the name!

**Expected behavior**
If a screen name is too long, it should use an ellipsis and a max width.

**Screenshots**
![Screenshot 2022-12-06 at 14 32 11](https://user-images.githubusercontent.com/101575380/205939574-18d99c23-3dee-419d-8b4a-5c918f4cb295.png)

shogunpurple:
Nope all good, I’d only started it!

Issue #8947, opened 2022-12-06
Table Filtering Issues


Mitch-Budibase:
There are a few issues associated with internal table filtering:
1. More than
2. Less than
3. Match any of the following filters

Repro steps:
1. Create app
2. Add table
3. Populate table (e.g. Dogs, with name, age, and breed)
4. Add More than filter to the table for age
**- The selected number is included in the more than filter**
**- e.g. If i select more than age of 5, dogs with an age of 5 still appear**
5. Remove previous filter
6. Add less than filter
**- The selected number is included in the less than filter**
**- e.g. If i select less than age of 5, dogs with an age of 5 still appear**
7. Add another filtering (2 total)
8. Set behaviour to 'Match any of the following filters'
9. Save and review filtered table - **The filter has not worked correctly, not all results are shown**
10. Click filter option to open modal - **The behaviour option has changed to 'Match all of the following filters'**

aptkingston:
In terms of less than and more than - this is correct. The actual behaviour is "more than or equal to" and "less than or equal to". We can update the labels, but can't update the functionality.

As for the match any - I believe that doesn't work at all in the data section like you've noticed, so I think we just need to remove that as an option.

Mitch-Budibase:
Thanks for looking into this Cheeks! I'd say it would be worthwhile to update the labels for the more than and less than options, just to remove any potential confusion.

Issue #8946, opened 2022-12-06
Components: Cannot Edit Component Name


Mitch-Budibase:
There is a problem with changing the name of a component after adding it

Reproduction steps:
1. Create app
2. Add any component to a screen
3. Attempt to change the name of the component - Error occurs
4. Repeat above steps for several other components - Error occurs for each component name change

The error is :

> Error updating component prop

Here is a screenshot showing the error:
![Screenshot 2022-12-06 at 12.44.25.png](https://images.zenhubusercontent.com/60fed9d54a973614368dfdfe/e0f8f68f-ac01-45a0-81b2-8f18d27af799)

Here is a loom video for additional context:
https://www.loom.com/share/5cf1448de8604dcd9373338688f8575a

aptkingston:
This sometimes happen when you get into a bad state, and then it happens continuously - but if you refresh it will be gone.

PR #8965, opened 2022-12-07
Misc fixes

aptkingston:
## Description
Collection of fixes. Targetting develop because some of these affect the side panel and the other fixes affect core things like filtering which need tested well.

It's after code freeze but my preference would still be to get this merged in and included in the ongoing testing.

Fixes:
- Fix rare crash when attempting to determine datasource schema for non string types
- Fix selected rows bindings not having a category in the binding drawer
image
- Add sensible min widths to fields in both the conditional UI drawer and filter drawer to fix being unusable on small resolutions
- Fix extremely long screen names not being cut off properly #8950
image
- Fix other issues with extremely long options labels not breaking. Now "auto width" selects cut off at a sensible max width
image
- Fix issue with side panels where content from multiple panels would be shown due to lack of a `destroy` callback in the svelte action
- Fix not being able to update component names to due recent component settings changes #8946
- Fix filter drawer not working properly in the data section or automation section due to the "match any" functionality on being implemented in design, but the component was reused. Filtering with "match any" now works anywhere, and I've tidied up some of that logic #8947
Example of filtering on "any" in data section
image
image
- Renamed "more than" and "less than" to include "or equal to" to reflect the actual functionality #8947
- Fix side panel ejection not working due to previous fixes attempting to prevent stale data being shown

### DIFF
--- a/packages/bbui/src/Form/Core/Picker.svelte
+++ b/packages/bbui/src/Form/Core/Picker.svelte
@@ -205,7 +205,10 @@
     width: 100%;
   }
   .spectrum-Popover.auto-width :global(.spectrum-Menu-itemLabel) {
+    max-width: 400px;
     white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
   .spectrum-Picker {
     width: 100%;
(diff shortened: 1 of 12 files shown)


mike12345567 (approved):
LGTM! Great fixes, should definitely get this in before release.

PR #8968, opened 2022-12-07
Minor update - stop users updating their own admin/global flags

mike12345567:
## Description
Extension on fix for user self assignment, don't allow users to change their admin/builder status.

Discovered by @MihailHadzhiev2022.


Rory-Powell (commented):
LGTM - Could use some unit tests for these scenarios

Issue #8878, opened 2022-11-30
Refactor data section routing


aptkingston:
In order to finish work on the command palette feature, we first need to refactor the data section to properly sync the URL with state. This has been done in the design section already by way of a new `syncURLToState` helper.

melohagan:
Just to add to this one, I pulled down the data-routing-refactor branch, but I'm still seeing a little issue; not sure if it is related or not.

1. Click on design tab
2. Add component -> Container
3. Click on the data tab
4. See that it does not navigate to data.
5. Click on design and back to data to get it to navigate.

![tabs_bug](https://user-images.githubusercontent.com/101575380/212673932-275dfb9f-ca9e-4256-a7c0-58c06896e88f.gif)

PR #8996, opened 2022-12-09
Builder data section routing refactor

aptkingston:
## Description
This PR refactors the data section in the builder, updating routing and properly syncing URL to state. This will fix a lot of current weird issues in the data section with things like stale URLS, blank screens, weird navigation glitches when deleting entities and more. All routes now properly handle invalid and missing parameters too, and will redirect you as needed rather than crash the builder or throw errors when using invalid or missing route params.

I've also standardised all of the backend stores, removed a load of wasted API calls which were causing needless server load, sped everything up by pruning many wasted state updates, fixed the logic for showing what is selected in the datasource navigator and unified the routing for REST queries with other queries so they don't have different URLs any more. I've also updated the styles of the datasource navigator to match the design section.

This also lays the foundation to get the command palette feature merged.

Some of these diffs look gnarly but it's a lot of reordering rather than rewriting for most of the stores.

Don't merge until after the December release next week.

Addresses: 
- #8878


packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_layout.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
+  import { params } from "@roxi/routify"
+  import { datasources } from "stores/backend"
+  import { syncURLToState } from "helpers/urlStateSync"
+  import * as routify from "@roxi/routify"
+  import { onDestroy } from "svelte"
+
+  const stopSyncing = syncURLToState({
```
Comment (shogunpurple) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_layout.svelte:8]: This change alone will solve a ton of bugs 👌

shogunpurple (approved):
LGTM! This lays a nice foundation in the data section for future work. Any issues that potentially arise from this should hopefully be caught by our E2E automation

Issue #8766, opened 2022-11-22
Filter or selection is not applied when exporting data


diogenesbrussels:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.28
  - App Version: 2.1.28
  - 
**Describe the bug**
The filter or selection made in a table in the Data section of an app isn't applied to the rows exported.

**To Reproduce**
1. Go to a table in Data section of an app
2. Click on Filter, use a criteria to filter the data and save filter
3. Export your data
4. The data exported is the complete table (without filter)

**Expected behavior**
The export function doesn't use the filter selected neither the selection of rows made by checking boxes in the table.
The export should reflects the filter and selection so only the visually visible/selected rows are exported.

**Desktop (please complete the following information):**
 - OS: Win11
 - Browser Firefox
 - Version 106

PR #9025, opened 2022-12-13
Filtering and sort on data row export

deanhannigan:
## Description

Some enhancements to the export row functionality in the data section:

- Export only rows that have been selected on the current page.
- Apply a custom sort to the exported rows.
- Export a distinct set of rows with multiple filters applied.
- If no custom filtering or sorting has been applied, all rows will be exported as normal.

In any of the above scenarios, the user will be given a summary of how their choices will shape their export.

Addresses: 
- https://github.com/Budibase/budibase/issues/8766

## Screenshots

Any filtering or sorting applied to the export is now listed before download.

Screenshot 2022-12-13 at 10 23 04

A confirmation of the row count will be displayed if the user has decided to export only selected rows.

Screenshot 2022-12-13 at 10 24 47


packages/builder/src/components/backend/DataTable/modals/ExportModal.svelte
```diff
@@ -15,8 +22,65 @@
...
+  let exportOpDisplay = []
+
+  $: luceneFilter = LuceneUtils.buildLuceneQuery(filters)
+
+  $: if (filters) {
+    exportOpDisplay = filterDisplay()
+  }
+
+  $: if (sorting) {
+    exportOpDisplay.push({
```
aptkingston, line 40: Running an array.push in a reactive statement like this could be bug prone, because if the reactive statement runs more than once then it will end up pushing additional duplicates onto the array. I would either passing `sorting` as a param to the `filterDisplay` function which creates `exportOpDisplay` and optionally append it, or use a different reactive variable which spreads `exportOpDisplay` and optionally appends another object.

Either way, you just want to ensure that it can't create duplicates from further invocations of the reactive statement.

aptkingston (commented):
Few svelty things to double check, but LGTM apart from that! Nice QoL improvement 👍

aptkingston (approved):

Issue #9015, opened 2022-12-12
Is in filter no longer directly supports relationship pickers


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.2.1
  - App Version: 2.2.1

**Describe the bug**
A common use-case that was somewhat simple has now become impossible without confusing code. 
This is the tutorial here: https://docs.budibase.com/docs/filter-by-relationships

Originally you could use the **Is in** filter as shown in the docs screenshot, i.e.
![image](https://user-images.githubusercontent.com/101575380/207101536-f0084453-acd5-45c4-b734-4a2c34f3da2e.png)

Now the following JavaScript binding is required:
```js
return $("Details Form.Fields.Pets").map(pet => parseInt(decodeURI(pet)?.replace(/[\[\]]/g, "")));
``` 
It's difficult to put that in a tutorial 😓 

**To Reproduce**
1. Follow the steps in the docs tutorial. See that it doesn't work.
2. Replace the **Is in** filter with the JavaScript binding above. It will now work.

**Expected behavior**
The **Is in** filter should be somewhat flexible and not require exact typings and decoding. Maybe the filter needs more context as to whether the field is a relationship and if it is external/internal. 

**SQL Scripts**

**Owners SQL**
```sql
CREATE TABLE IF NOT EXISTS public."Owners"
(
    id integer NOT NULL,
    "Name" text COLLATE pg_catalog."default",
    "Address" json,
    CONSTRAINT "Owners_pkey" PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS public."Owners"
    OWNER to postgres;
```

**Pets SQL**
```sql
CREATE TABLE IF NOT EXISTS public."Pets"
(
    id integer NOT NULL,
    "Name" text COLLATE pg_catalog."default",
    "Animal" text COLLATE pg_catalog."default",
    "Breed_test" text COLLATE pg_catalog."default",
    "Date of birth" timestamp with time zone,
    fk_owner integer,
    "Age" integer,
    CONSTRAINT "Pets_pkey" PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS public."Pets"
    OWNER to postgres;
```

**Appointments SQL**
```sql
CREATE TABLE IF NOT EXISTS public."Appointments"
(
    id integer NOT NULL,
[...]

Issue #8334, opened 2022-10-20
External datasource fails when filtering with _id


gmentzik:
**Hosting**
- Self
  - Method:  docker compose
  - Budibase Version: 2.0.32
  - Mysql version 8.0.30 , 8.0.32 (both dockerized and stand alone tested)

**Describe the bug**
Data provider connected with Mysql datasource table cannot filter by _id and generates error "
Unknown column '{mysqltablename}._id' in 'where clause'", when trying to load page with and  _id  at the url bar at the following cases:
- After modifying filters at auto generated page provider filter 
- Creating a new provider and filter by _id. 
The only way you can filter by _id is creating an auto generate page and using the provider generated taking care not touching  "Define filters".

**To Reproduce**
1. Install a clean budibase 
2. Install mysql 8.0.  Both stand alone at host and dockerized version included as service in budibase docker compose were tested)
3. Create database and user for budibase to connect with. Create a simple table eg "testtable" with an int AI PK 'tid' column and a text column as value.
4. At budibase create a new app , at mysql datasource and autogenerate pages for "testtable".
5. Duplicate the testtable/:id page to something like testtable2/:id
6. Check preview of application and use view button at testtable page to go to testtable/:id. Change url to testtable2/:id. Now both should work.
7. At editor testtable2/:id page try the following scenarios that should all fail:
- Add a 2nd filter to provider.
- Delete/Modify 1st filter save and modify again to _id.
- Delete auto generated provider and add new with _id filter.

**Expected behavior**
Filter by _id should always work if auto generated Mysql table provider's filters are modified or new user added data Mysql table provider with _id filter is added.

**Screenshots**
![image](https://user-images.githubusercontent.com/65217733/196893800-103341b5-4731-45f3-9733-2402560674ce.png)
  

**Desktop (please complete the following information):**
 - OS: Linux Mint 20.3 (Ubuntu) , CentOS 7.8
 - Browser Chrome, Firefox

melohagan:
As a note, the screen duplication is not required to reproduce. Ejecting the form block and adding a second filter to the data provider will cause an issue on the original `/:id` screen.

melohagan:
@gmentzik The work around at the moment is to change the filter from `_id` to `product_id` or whatever your primary key is named within your MySQL table:

*This will break*
[...]

Issue #3386, opened 2021-11-15
`Like` operator does not work for internal tables


mjashanks:
When using a filter on a `Data Provider`, with internal tables, the `Like` operator does not work. It works fine on SQL.

I seem to remember discussing this before and finding that it won't be possible with internal tables.

Let's remove it, for internal tables.

![image](https://user-images.githubusercontent.com/3524181/141793280-30476112-e77c-4a0c-8680-b12b31e3f451.png)

mike12345567:
Like should work, but differently than you'd expect for internal tables, its actually a fuzzy search in the case of internal tables, which is not hugely useful beyond a free text search input.

aptkingston:
Do we need to keep this around for external SQL tables? It's actually quite a pain to only have it for external, because everywhere that previously only needed to know schema types (e.g. from data bindings) now needs to know the actual type of the datasource as well. We've relied on having the same searching API across the board since that was intended to be a feature.

For what it's worth, "Like" does work as intended when I use it. It's a fuzzy search and I think allows for 2 character corrections. For "Iran" for example, "oran", "irab" and even "xxan" all work. @mjashanks could you provide more details on the setup here that didn't work? I would rather investigate and fix a bug if one exists around the usage for internal tables than remove it, because it is supported and does work in the testing that I've done.

mjashanks:
I think it's a misunderstanding of the term. I was expecting it to work like `startsWith` or `contains` - probably because I come from SQL, and that's the most common use of a SQL `LIKE` .... 

1. `WHERE country LIKE 'Ir%'`. matches `Iran`, `Ireland`
2. `WHERE name LIKE '%john%'` matches `John Smith`, `Michael Johnson`

If it does a fuzzy search, then maybe we should rename it `Fuzzy Like`, in the builder.

Long term, I would have thought that this would be an item on the "Feature List" of a datasource - i.e. the list of filter operators it supports. But I know we are not there yet.

xnohat:
Hi, this bug still available ?
Currently at version 1.0.90, Like operator still not working for internal tables, also not working for fuzzy search

mike12345567:
Hi @xnohat - due to technical restrictions in place around the current version of CouchDB we are not able to offer a search along the lines of how an SQL `LIKE` operator works. You can use the string search for a starts with search, but there is no way to do a contains search.

bjornclauw:
[...]

PR #9030, opened 2022-12-13
Fix SQL table `_id` filtering

aptkingston:
## Description
This PR fixes some issues with filtering. Previously the `_id` field was removed from filtering altogether as it was believed that it did not work. This was incorrect, and the issue at that time was actually caused by the previous changes to prefix filter fields with numbers to allow multiple filters on the same field.

This PR:
- Adds back `_id` as a filter option for SQL tables
- Fixes filtering so that `_id` filters work for SQL tables. This supports multiple filters as well, and the "match any" behaviour.
- Restricts which operators are available for `_id` in SQL tables (only equals, not equals)
- Removes the "like" operator for internal tables (it is actually just "starts with" at the moment so is redundant)
- Updates data, design, automate and client to all respect these rules. There are still some differences between available fields depending where you filter because we have little consistency between how schemas are generated.

Addresses: 
- https://github.com/Budibase/budibase/issues/8334
- https://github.com/Budibase/budibase/issues/3386

This does not fix https://github.com/Budibase/budibase/issues/9015, in the sense that it is still impossible to do an "is in" with `_id` on a SQL table, but we cannot support that functionality without substantial additional work. This PR at least removes the "is in" operator to make it clear that it is not supported.

### Screenshots
SQL table with a working ID filter, showing only the available operators:
image

Internal table with "like" removed:
image

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -232,6 +232,7 @@
             {filters}
             {bindings}
             {schemaFields}
+            datasource={{ type: "table", tableId }}
             panel={AutomationBindingPanel}
             fillWidth
             on:change={e => (tempFilters = e.detail)}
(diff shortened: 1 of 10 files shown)


mike12345567 (approved):
LGTM! Thanks for getting this fix together!

PR #9031, opened 2022-12-13
Ensure query results are re-rendered when switching between db queries.

deanhannigan:
## Description

Ensures that the db query result tabs for json, preview and schema are fully reloaded when moving from one query to the next.

This also addresses a routing issue when navigating from a `DB query` to a `REST datasource`. Doing so used to navigate the app to an empty REST query rather than the root REST datasource view.

## Screenshots

The contents of the results tab updating correctly

schemaRender


aptkingston:
This mayyy conflict heavily with the data routing refactor I've had up for a while. I potentially had also fixed this issue (although maybe not specifically the query results part). Do you want to check out my branch and see if the issue still exists?

deanhannigan:
@aptkingston I ran through the same scenarios on your branch and the refactor has fixed them both. I'll close this

Issue resolved here : https://github.com/Budibase/budibase/pull/8996

Issue #9029, opened 2022-12-13
couch returned 400 - users table


andyburgessmd:
### Discussed in https://github.com/Budibase/budibase/discussions/9016

<div type='discussions-op-text'>

<sup>Originally posted by **applicationuser** December 12, 2022</sup>
Hi,

Error when navigating screen in published app and table design access.

Table page not showing data rows.  Table has data before publishing app and new rows added in published app.

Budibase Cloud
App Version : 2.2.1

published app error on navigated to screen
![image](https://user-images.githubusercontent.com/42922833/207116729-fc9f3d0d-b62f-4307-ae46-8efeece61e8d.png)

![image](https://user-images.githubusercontent.com/42922833/207116775-b2f4641e-fa1a-4cda-83be-ab5b72e356f6.png)

Error when table accessed
![image](https://user-images.githubusercontent.com/42922833/207118336-9797cdda-92e1-4ba1-85f1-ad8cf872ce81.png)

</div>

mike12345567:
Hi @andyburgessmd - would it be possible for you to provide an export of an app which is exhibiting this behaviour?

andyburgessmd:
@mike12345567 I've just emailed it to you using the address from your github profile. Hope that's ok?

shogunpurple:
Hi folks - the latest change has now been deployed to budibase cloud and self host. We will close the issue out here if everyone is back up and running. Apologies for the service interruption.

PR #9038, opened 2022-12-13
Fix for user enrichment call (updating to @budibase/nano fork)

mike12345567:
## Description
Fix for #9029 - this should fix the issue users have been experiencing with user enrichment calls in apps, essentially it utilises a fork of the nano library we use to interact with CouchDB, which has been updated to use a POST request rather than a GET request as it supports a larger set of data being sent as query parameters.

Post mortem will follow going into greater detail around the diagnose of this. This is also the most minimal fix possible to get this out - but will need to perform some further updates to `@budibase/nano` to improve its error outputs and fix any other functions which also utilise `GET` requests with a query string, that could instead use `POST` requests.

Addresses: 
- #9029

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@budibase/types": "^2.2.2",
+    "@budibase/nano": "^10.1.0",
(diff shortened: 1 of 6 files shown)


shogunpurple (approved):
LGTM! great work on that fix.

PR #9047, opened 2022-12-14
Adding price duration to BB account and redlock updates

shogunpurple:
## Description
Supporting work for hubspot events in this PR: https://github.com/Budibase/account-portal/pull/196

Adding the plan duration (Monthly/Annually) to the billing account - this allows us to determine if users have churned and find the price for the stripe plan that they are moving to, which in turn enables us to calculate the deal total in hubspot.

Also includes some redlock changes from the free trial work so that we can use it to lock the license sync call, preventing a race condition with several stripe webhooks firing in close succession.

Addresses: 
- ...more if required

## App Export

## Screenshots

### DIFF
--- a/packages/types/src/documents/account/account.ts
+++ b/packages/types/src/documents/account/account.ts
@@ -3,6 +3,7 @@ import {
   Hosting,
   MonthlyQuotaName,
   PlanType,
+  PriceDuration,
   Quotas,
   StaticQuotaName,
 } from "../../sdk"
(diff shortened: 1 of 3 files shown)


packages/types/src/documents/account/account.ts
```diff
@@ -46,6 +47,7 @@ export interface Account extends CreateAccount {
   tier: string // deprecated
   planType?: PlanType
   planTier?: number
+  planDuration?: PriceDuration
```
Rory-Powell, line 50: Mentioned on the other PR. I think we should just rename this type `Duration` now

Rory-Powell (approved):

Issue #8415, opened 2022-10-27
Allow disabling of specific dates in Date Picker


melohagan:
**Describe the feature request**
As a user, I want to specify a list of dates that should be disabled in the date picker.
Related discussion: https://github.com/Budibase/budibase/discussions/8402

https://flatpickr.js.org/examples/#disabling-specific-dates

melohagan:
Hey @daanvdberg12 at some stage we are going to replace the current date picker component, but there is no timeline on this for now.

Closing in favour of: https://github.com/Budibase/budibase/issues/11894

Gtchik:
Hey,

I really needed this feature in my project, and since it isn't currently supported natively, I came up with a workaround inspired by the discussion mentioned here #6839 

Admittedly, the solution is not the cleanest, but it gets the job done for my use case. Here's the code I integrated into an `embed` block:  

```
<img style="display: none;" src onerror="
  if (document.readyState === 'complete' || document.readyState === 'interactive') {
    initializeObserver();
  } else {
    document.addEventListener('DOMContentLoaded', initializeObserver);
  }

  function initializeObserver() {
    const targetDates = ['Friday, December 13, 2024', 'Tuesday, December 17, 2024']; 

    const observer = new MutationObserver(() => {
      targetDates.forEach((targetTitle) => {
        const cells = document.querySelectorAll(`td[title='${targetTitle}']`);
        
        cells.forEach((cell) => {
          if (!cell.dataset.disabled) {
            // Remove existing events by replacing the element with a clone
            const clone = cell.cloneNode(true);
            clone.addEventListener('click', (event) => {
              event.stopPropagation();
              event.preventDefault();
            });

            // Apply a style to indicate it is disabled
            clone.setAttribute('aria-disabled', 'true');
            clone.dataset.disabled = 'true';
            clone.style.opacity = '0.5';  // Make the element visually dimmed
            clone.style.pointerEvents = 'none';  // Disable user interactions

            // Strike through the textual content
            const dateSpan = clone.querySelector('.spectrum-Calendar-date');
            if (dateSpan) {
              dateSpan.style.textDecoration = 'line-through';
            }

            // Replace the old element with the clone
            cell.parentNode.replaceChild(clone, cell);
          }
        });
      });
    });

    observer.observe(document.body, {
      childList: true,
      subtree: true, // Also observe child elements
    });
  }
">
```

[...]

PR #9059, opened 2022-12-15
Disable specific dates and support ranges

melohagan:
## Description
Attempting to solve this use-case: https://github.com/Budibase/budibase/discussions/8402
When you want to allow users to choose a *Start* and *End* date, but not allow them to create a range that intersects disabled dates.

### Disabled dates
Consider a table with holidays that you want to disable from the date picker calendar:
Screenshot 2022-12-15 at 14 10 41

You can use the *Disabled dates* field to disable these dates. In this case the following handlebars expression is being used: `{{ pluck DisabledDates Data Provider.Rows 'Holiday' }}`

Screenshot 2022-12-15 at 14 12 12

Screenshot 2022-12-15 at 14 12 31

### Date range
There was some code already in here to handle a little bit of the flatpickr range mode, however I wasn't able to find anywhere that actually used it, but if there is some other use of the functionality I need to test then let me know!

Added a check box for *Range mode*. When ticked, you can select a date field as the *To* (end) date.
Screenshot 2022-12-15 at 14 15 17

Screenshot 2022-12-15 at 14 16 06

This is probably a bit unorthodox as it involves registering two form fields in one component, but here's an example form:
Screenshot 2022-12-15 at 14 20 47

Screenshot 2022-12-15 at 14 19 35

Screenshot 2022-12-15 at 14 21 42

Unticking the range mode will revert to using the *Start* date only.

Addresses: 
- https://github.com/Budibase/budibase/issues/8415

## App Export
[Date Picker Range-export.tar.gz](https://github.com/Budibase/budibase/files/10237642/Date.Picker.Range-export-1671113331331.tar.gz)

### DIFF
--- a/packages/bbui/src/Form/Core/DatePicker.svelte
+++ b/packages/bbui/src/Form/Core/DatePicker.svelte
@@ -18,6 +18,7 @@
   export let ignoreTimezones = false
   export let time24hr = false
   export let range = false
+  export let disabledDates = []
   const dispatch = createEventDispatcher()
   const flatpickrId = `${uuid()}-wrapper`
   let open = false
@@ -53,6 +54,7 @@
     mode: range ? "range" : "single",
     appendTo,
     disableMobile: "true",
+    disable: disabledDates,
     onReady: () => {
       let timestamp = resolveTimeStamp(value)
       if (timeOnly && timestamp) {
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -3200,6 +3200,18 @@
         "defaultValue": false
       },
       {
+        "type": "boolean",
+        "label": "Range mode",
+        "key": "range",
+        "defaultValue": false
+      },
+      {
+        "type": "field/datetime",
+        "label": "To field",
+        "key": "toField",
+        "dependsOn": "range"
+      },
+      {
         "type": "text",
         "label": "Default value",
         "key": "defaultValue"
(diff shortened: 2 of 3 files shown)


mike12345567:
Hey @melohagan had a look through this - I think its an important change but I'm not sure this is the correct way to go about it. We've a relatively large dependence on `flatpickr` right now - which we don't like - this will increase our dependency on the package and make it harder to replace.

I also think from our technical point of view breaking this out into a separate component, so we have a `Date Picker` and a `Range Picker` would be preferable - I had a chat with @aptkingston about this - there already is a component we use in the builder, it would just need to be exposed to the client library.

Going to close this for now as I think we'll need to think further about this - the date picker is already quite problematic and I wouldn't want to worsen the issue.

melohagan:
> Hey @melohagan had a look through this - I think its an important change but I'm not sure this is the correct way to go about it. We've a relatively large dependence on `flatpickr` right now - which we don't like - this will increase our dependency on the package and make it harder to replace.
> 
> I also think from our technical point of view breaking this out into a separate component, so we have a `Date Picker` and a `Range Picker` would be preferable - I had a chat with @aptkingston about this - there already is a component we use in the builder, it would just need to be exposed to the client library.
> 
> Going to close this for now as I think we'll need to think further about this - the date picker is already quite problematic and I wouldn't want to worsen the issue.

That's fair - thanks for having a look!

Issue #7537, opened 2022-08-31
Unable to clear date in Date/Time column in PostgreSQL


bdelcamp:
**Hosting**

- Self
  - Method: docker
  - Budibase Version: 1.2.52

**Describe the bug**
Unable to remove date from column.

**To Reproduce**
1. Create column with Date
2. Set date
3. Edit, and highlight date, press Delete.
4. Date clears in dialog, but not in actual database row. This breaks formulas to check if the date is set.

**Expected behavior**
Deleting date in picker dialog should clear date.

 - OS: MacOS
 - Browser Edge, Safari

shogunpurple:
@bdelcamp could you please send an app export with a reproduction?

radtkedev:
@shogunpurple It seems like this bug only happens when using PostgreSQL, I could not recreate this issue when using the internal Budibase database. Using selfhosted single-container docker variant on version 2.1.28.

Steps to recreate this bug:
- Connect PostgreSQL
- Create a new table with the following columns:
  - SomeText (required, Type: Text)
  - SomeDate (optional, Type: Date/Time)
- Create a row with any text or date
- Publish an autogenerated screen
- Edit, highlight the date and press the delete key
- After saving, the date field does not get deleted, no errors are shown

PR #9078, opened 2022-12-16
Allow postgres to clear date

melohagan:
## Description
When making an external DB save, null values are treated as un-entered, thus are ignored.
The field clear returns an empty string so that external DBs can delete a date value from the DB. The external request converts the empty string to a **null** before saving.

Addresses: 
- https://github.com/Budibase/budibase/issues/7537


aptkingston (approved):
LGTM!

Issue #8322, opened 2022-10-18
"budibase-infra" Docker Image with couch, minio and redis


shogunpurple:
**COS**
- Build and package a docker image with couch, minIO and redis exposed on different ports
- Use this image in `qa-core` and in the budibase unit tests 
- Use in CI using the github actions `services` key
- Remove all references to `isTest()` that were relying on the mocked services (bull, file system minio and pouchdb)
- Use for local dev

```
services:
    bbtest:
      image: budibase/test
      ports:
        - 6379:6379
        - 4567:5984
        - 9001:9001
```

PR #9079, opened 2022-12-16
Breaking out CouchDB image from single image and test image

mike12345567:
## Description
This PR provides the basis of the test image - it does this by breaking out the CouchDB component of the single image into its own image (https://hub.docker.com/r/budibase/couchdb) and then utilised this for creating the test image (https://hub.docker.com/r/budibase/tests).

I've also started to integrate the Jest Docker pre-runner (https://github.com/Trendyol/jest-testcontainers) package, so that we can make sure the Docker image has started before we run the test suite.

Will flesh out with further details - this is just to show how this will work - working on converting tests currently.

Addresses: 
- #8322


packages/backend-core/src/db/db.ts
```diff
@@ -7,11 +7,11 @@ const dbList = new Set()
...
-  if (env.isTest()) {
-    dbList.add(dbName)
-    // @ts-ignore
-    return getPouchDB(dbName, opts)
-  }
+  // if (env.isTest()) {
+  //   dbList.add(dbName)
+  //   // @ts-ignore
+  //   return getPouchDB(dbName, opts)
+  // }
```
Rory-Powell, line 14: Should this be removed now?

hosting/docker-compose.dev.yaml
```diff
@@ -3,24 +3,23 @@ version: "3"
...
 services:
-  minio-service:
-    container_name: budi-minio-dev
+  dev-service:
+    container_name: budi-dev
     restart: on-failure
-    image: minio/minio
-    volumes:
-      - minio_data:/data
+    image: budibase/tests
```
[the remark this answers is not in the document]
mike12345567: I wanted to name it specifically for what it does - contain functionality that is useful for testing Budibase. If it were named dev then it would suggest that is the only way to develop with Budibase.

I'm still not settled on whether or not to change the dev environment as it will break everyone's local dev envs - i'm not sure theres enough of a benefit to do this really - I had been thinking of reverting this change.

packages/backend-core/src/db/db.ts
mike12345567: Still a WIP - should have marked as draft!

PR #9130, opened 2022-12-20
Only build pro after bootstrapping and after backend-core build

Rory-Powell:
## Description
This protects against the scenario of a breaking change in backend-core
that is relied on by the pro package.

Introducing to fix the current build as the published backend-core package is corrupted.
This change ensures the local filsystem version of core will be used.


packages/backend-core/package.json
```diff
@@ -15,6 +15,8 @@
     "prebuild": "rimraf dist/",
     "prepack": "cp package.json dist",
     "build": "tsc -p tsconfig.build.json",
+    "build:pro": "sh ../../scripts/pro/build.sh",
```
shogunpurple, line 18: is it necessary to run this inside an `sh`?

shogunpurple (approved):
LGTM!

PR #9139, opened 2022-12-21
QueryLevelTransformers Test Update

Mitch-Budibase:
Updated selectExternalDatasource command

Altered Query Level Transformers to include a page reload. Not having this reload was occasionally affecting the test run


Pedro-Revez-Silva (commented):
LGTM

Issue #9138, opened 2022-12-21
Incrementing hour with time only Date Picker


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/8916

<div type='discussions-op-text'>

<sup>Originally posted by **kellis5137** December  3, 2022</sup>
I'm finding a weird issue with the Date picker. If you choose to have a "time only" picker, it seems the picker automatically advances the time by 1 hour when you click into it, then click out of it. Or if you click into it, advance the time by 1 hour, then step out... it goes up one. A user actually found this issue.

Screen shots in order...
<img width="1412" alt="Screenshot 2022-12-02 at 7 39 42 PM" src="https://user-images.githubusercontent.com/1899734/205414018-7e1b0c0d-e0e5-4906-ba07-6cf74fe19361.png">
<img width="1388" alt="Screenshot 2022-12-02 at 7 39 54 PM" src="https://user-images.githubusercontent.com/1899734/205414021-eb122f55-e9c7-477e-a3b6-ef7e593d776e.png">
<img width="1406" alt="Screenshot 2022-12-02 at 7 40 05 PM" src="https://user-images.githubusercontent.com/1899734/205414025-ee3eda5a-c098-48ab-860b-d01b1cb8ca1c.png">

Keiths app-export-1670027839074.tar.gz</div>

koneubauer:
We have this exact issue in 2.11.41

insomniacj93:
Hello, i have the same issue at 2.13.15. After the Summer hour change users that select 10:00 for example, the date picker updates to 09:00. Same issue happens with my SQL table entries, they are saved with 15:00 for example and in budibase we get 14:00 although the table rows are all correct with 15:00!

PR #9140, opened 2022-12-21
Undo time offset

melohagan:
## Description
Time was being intentionally offset in the Date Picker code, but it's causing an unwanted affect.
@aptkingston Do you remember the scenario that was previously not working?

Addresses: 
- https://github.com/Budibase/budibase/issues/9138

bug


aptkingston:
It was added for a reason - as commented in the code. Flatpickr was probably just ignoring timezone and always using daylight savings. Guessing it works fine with this removed now that the hour has changed?

melohagan:
> It was added for a reason - as commented in the code. Flatpickr was probably just ignoring timezone and always using daylight savings. Guessing it works fine with this removed now that the hour has changed?

I did try setting the hour forward and backwards using the Time Shift plugin. Wasn't able to recreate the original issue described by the code comment.

melohagan:
I tested the Date Picker on **1.2.2** and wasn't able to recreate the original issue (this version was before the original bug fix was added).

You can try it out on: http://singleimage.qa.budibase.net/

aptkingston:
No problem, happy to remove this then! This bug will probably reappear when daylight savings ends again, but we can just keep toggling this until we remove the scourge that is flatpickr.

aptkingston (approved):

PR #9145, opened 2022-12-21
Ensure view names are properly encoded to handle certain special chars

aptkingston:
## Description
This PR fixes a few issues with special characters in view names. This is not a proper solution for view names in general as certain special characters are just incompatible with couch view names altogether (`?` being the most broken it appears), but this at least handles the most common use cases of characters like spaces.

This at least brings develop back up to par with master in terms of handling spaces etc, and makes the encoding/decoding of view names more consistent in general.

- Always encode view names when sending them to the server as URL params
- Always decode view names on the server when handling requests
- View names are always encoded when used in URLs in the builder
- View name URL params are always decoded before being written to state in the builder
- `$isActive` from routify seems to use a regex under the hood, and actually throws an error if you attempt to check if a URL is active which contains unescaped regex special chars such as `(`. This is really a bug on their part, but I've accounted for it by updating our usages of `$isActive` on view pages.
- Trim whitespace when creating view names to avoid one bricking scenario at least

Future work should be done to either enforce an alphanumeric character set for view names or, much better again, generate IDs and use those as view identifiers under the hood.

Addresses: 
- https://github.com/Budibase/budibase/issues/9141

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateViewModal.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateViewModal.svelte
@@ -10,6 +10,7 @@
   $: views = $tables.list.flatMap(table => Object.keys(table.views || {}))
 
   const saveView = async () => {
+    name = name?.trim()
     if (views.includes(name)) {
       notifications.error(`View exists with name ${name}`)
       return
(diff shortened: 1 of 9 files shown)


shogunpurple (approved):
LGTM!

PR #9154, opened 2022-12-22
Cypress views testCypress: Create Views test update

Mitch-Budibase:
Updating a couple of tests within createView
- Timing issues
- Ensuring we get the correct element for testing

Also updated QueryLevelTransformers test
- Timing issue
- I've extended the timeout value for when checking the nav item content (passes locally but timed out on the CI run)

Increased timeout values within the following commands:
- applicationInAppTable
- createTable


Pedro-Revez-Silva (approved):
LGTM

Issue #8869, opened 2022-11-29
API Test Coverage: Rename App (Replace Remaining Cypress Tests)


Mitch-Budibase:
We currently have a set of API tests which cover renaming an application.

The next step is to completely replace our associated Cypress tests.
Here is the remaining tests:
- Create two applications with the same name
- Validate application names

There is already partial coverage of these tests within the current API automation suite. A few changes or additional tests is desired to fully cover what Cypress was testing

Issue #8866, opened 2022-11-29
API Test Coverage: Tables (Replace Remaining Cypress Tests)


Mitch-Budibase:
There is existing API automation coverage for tables. We need to extend this to replace the remaining Cypress tests associated with tables.

Here are the tests:
- Add rows and check pagination
- Delete rows and check pagination

Issue #8865, opened 2022-11-29
API Test Coverage: App Overview (Replace Remaining Cypress Tests)


Mitch-Budibase:
There is already a set of API tests which cover Application Overview.

We need to extend this set of tests to replace the remaining Cypress tests associated with Application Overview.

Here are the tests:
- Allow App unlocking
- Edit app icon and colour

PR #9188, opened 2023-01-02
Extend Coverage of API tests to replace Cypress tests

Pedro-Revez-Silva:
## Description
Add more test coverage for some features to match Cypress tests and also has some improvements in the naming of tests, methods and comments.

Addresses: 
- #8865
- #8866
- #8869


qa-core/src/tests/internal-api/tables/tables.spec.ts
```diff
@@ -86,4 +86,61 @@ describe("Internal API - Application creation, update, publish and delete", () =
...
+
+    //Table was added
+    await config.tables.getAll(3)
+
+    //Get information about the table
+    await config.tables.getTableById(<string>createdTableData._id)
+
+    //Add Column to table
+    const newColumn = generateNewColumnForTable(createdTableData)
+    const [addColumnResponse, addColumnData] = await config.tables.save(
```
shogunpurple, line 112: should we have a utility method for `addColumn`?

qa-core/src/tests/internal-api/tables/tables.spec.ts
```diff
@@ -86,4 +86,61 @@ describe("Internal API - Application creation, update, publish and delete", () =
...
+    )
+
+    //Add Row to table
+    let newRow = generateNewRowForTable(<string>addColumnData._id)
+    await config.rows.add(<string>addColumnData._id, newRow)
+
+    //Search single row
+    await config.rows.searchNoPagination(<string>createdTableData._id, searchBody(<string>createdTableData.primaryDisplay))
+
+    //Add 10 more rows
```
shogunpurple, line 124: NAB: Worth doing this in a bulk operation

qa-core/src/config/internal-api/TestConfiguration/rows.ts
```diff
@@ -36,4 +36,21 @@ export default class RowsApi {
     expect(response).toHaveStatusCode(200)
     return [response, json]
   }
+
+  async searchNoPagination(tableId: string, body: any): Promise<[Response, Row[]]> {
```
shogunpurple, line 40: Why do we have 2 methods for searching with and without pagination? Is there ever a point where we absolutely don't want pagination and need to assert on it? Seems easier just to have one `search` method

Pedro-Revez-Silva: We need to check that pagination is working, both by having multiple pages when there are more than 10 rows and only one page when there are 10 or less rows. We could have an `if` inside that method, but this is against best practices because in order to check both scenarios we would need to add logic inside the method and that would require its own tests. Each method does its own simple assertion and it will fail if the result is unexpected, as it should.

qa-core/src/tests/internal-api/tables/tables.spec.ts
Pedro-Revez-Silva: how do bulk operations work for rows?

Mitch-Budibase (approved):

PR #9195, opened 2023-01-03
Ensure the Minio UI port is always set to 9001

PClmnt:
## Description
Ensures that Minio exposes a specific port on startup, so that users who are using the single image can bind to it


melohagan (approved):
Thanks!

shogunpurple (approved):
LGTM!

Rory-Powell (approved):

Issue #9011, opened 2022-12-12
Create column internal DB issues


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb

**Describe the bug**
I am incorrectly able to create duplicate auto columns in an internal DB table.

Also I am able to enter a column name with a variation of capitalisation, but clicking `Create Column` doesn't save it. Validation should pick this up, i.e. do a comparison on a `toLowerCase()`.

**To Reproduce**
1. Create a new BudibaseDB table with all auto columns selected
2. Click `Create column` and select *Auto Column* type.
3. **I am allowed to create duplicate auto columns**

1. Create a new column: `Name`
2. Create another new column called: `NAME`. 
3. **You are allowed to save, but the column isn't actually created.** 

1. Add a row of data to the newly created table
2. Create an auto-generated screen
3. **See that the header is a JSON string**

**Expected behavior**
I should not be able to create duplicate auto columns in an internal DB table.
Validation should prevent case variations of the same column name being entered.

**Screenshots**
![Screenshot 2022-12-12 at 09 14 12](https://user-images.githubusercontent.com/101575380/207006892-b82cc4c3-92ea-40fd-b274-aec9b0608991.png)

*Duplicate Created By data*
![Screenshot 2022-12-12 at 09 17 00](https://user-images.githubusercontent.com/101575380/207007483-8aebe37c-e72c-4b20-9cb5-f9dde53cf164.png)

*Consequences of this bug in side panel*
![Screenshot 2022-12-12 at 11 12 45](https://user-images.githubusercontent.com/101575380/207031315-c9c1ce08-4446-407f-becd-ab4263f20f79.png)

PR #9205, opened 2023-01-03
Create/Edit Column refactoring and validation updates

deanhannigan:
## Description

- Ensures column names with mixed casing do not clash with existing columns of the same name. Names are also trimmed to ensure whitespace is ignored on validation.
- Refined the autocolumn subtype dropdown to only show the remaining subtypes, if any. The option will be disabled when none remain.
- Refactored the edit UX for autocolumns to exclude unnecessary UI and include the selected subtype.
- Added a "Schema Invalid" warning to the head of the database table if the user has previously created duplicate auto-columns. It will direct them to problematic columns and inform them that a future update will enforce the validation on the backend.

Addresses: 
- https://github.com/Budibase/budibase/issues/9011

## Screenshots

Name validation taking into account casing and surrounding white space

Subtype is now required with its own validation messaging. The field also will now only list the available autocolumn types.

If **all** subtypes have already been created for the table, the option to create another autocolumn is disabled.

Updated the Edit column UX for autocolumn fields to remove unnecessary field type validation and UI. Added in the explicit subtype to the bottom of the modal.

<div>

</div>
<br />
New schema warning to alert the user to the fact that duplicates autocolumn types exist in the table. It points out both the name and type of the columns
<br />


aptkingston (approved):
LGTM! Nice work, I know the logic here is a bit confusing.

Should we maybe swap out that validation warning message to use a `Banner` instead? I think we can just use `type="warning"` to make it a yellow warning banner, which might look more appropriate for a warning and also encourage them to take action!

aptkingston (approved):

PR #9209, opened 2023-01-03
Doing sanitizeKey when doing deleteFile(s)

xnohat:
Doing sanitizeKey when doing deleteFile(s). 
In case of backup file name in backups bucket backup-2023-01-03T185512.686Z.tar.gz but in global-db store backup filename is backup-2023-01-03T18:55:12.686Z.tar.gz sanitizeKey on deleteFile(s) method will prevent all other case related to this bug

### DIFF
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -331,10 +331,7 @@ export const deleteFile = async (bucketName: string, filepath: string) => {
     Bucket: bucketName,
     Key: sanitizeKey(filepath),
   }
-  return objectStore.deleteObject(params, function(err, data) {
-    if (err) console.log(err, err.stack); // an error occurred
-    else     console.log(data);           // successful response
-  });
+  return objectStore.deleteObject(params).promise()
 }
 
 export const deleteFiles = async (bucketName: string, filepaths: string[]) => {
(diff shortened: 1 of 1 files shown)


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -329,9 +329,12 @@ export const deleteFile = async (bucketName: string, filepath: string) => {
   await makeSureBucketExists(objectStore, bucketName)
   const params = {
     Bucket: bucketName,
-    Key: filepath,
+    Key: sanitizeKey(filepath),
   }
-  return objectStore.deleteObject(params)
+  return objectStore.deleteObject(params, function(err, data) {
+    if (err) console.log(err, err.stack); // an error occurred
+    else     console.log(data);           // successful response
```
Rory-Powell, line 336: I think we'll want to remove this, logging the delete response isn't necessary here. 

Instead of adding a callback function we should use `objectStore.deleteObject(params).promise()`, perhaps this is why the error wasn't being logged in the exception handler before, is that why you've added this?

Rory-Powell: > deleteObject without callback function cannot delete object successful 

Good catch! I've been able to replicate as well. 

Appending `.promise()` here fixed it for me, which is consistent with how we used the rest of the S3 SDK

xnohat: I have changed to .promise() instead callback for this pull request

Rory-Powell (approved):

PR #9218, opened 2023-01-04
Prefer app url resolution over referer

Rory-Powell:
## Description
Currently there is an issue with viewing a published app from within the builder. When this redirect takes place the `referer` header was previously used before app url resolution to get the app id. This resulted in the dev app id being embedded in the served app. 

This was introduced in https://github.com/Budibase/budibase/commit/7ef9e9dc51e7e90b4e30816422005c609ac77e17 where the referer resolution was fixed. 

In this changeset:
- Always use app url resolution first
   - filter out `/app/preview` - this would result in unnecessarily loading all apps when viewing the builder to search for the app with a custom url of `/preview` 
- Restrict referer resolution to `/builder/app/` paths only - the only reliable time we want to use the referer to determine the app id is when using the builder. Any other path is unreliable and should be ignored. 
- Add unit tests fro `getAppIdFromCtx`
- Linting on qa core     

Addresses: 
- No ticket - bug in develop only


shogunpurple (approved):
LGTM!

PR #9244, opened 2023-01-05
Fix looping issue with subsequent loop steps

### LINKED ISSUE #9123: In automations a second looped action returns each loop run as its own step
**Hosting**
- Self
  - Method: dev setup
  - Budibase Version: 2.2.4-alpha.7
  - App Version: 2.2.4-alpha.7

**Describe the bug**
In automations a single looped action will run and return as expected, with each loop response returned under one step, but on subsequent looped actions in the automation each loop returns from the server as its own step. This causes problems with the steps binding object and the results preview sidebar where the bad loop response will pollute subsequent step outputs. 

**To Reproduce**
1. Create a new automation
2. Add an action like server log and have it loop (this one will be OK)
3. Add another similar looped action (this one will return each loop as an individual step in the response)
4. Add a final step after the second looped action to demonstrate that a looped response from the previous step displays for this step.
5. Test the automation :)

**Expected behaviour**
All looped actions should be represented properly as a single step that has run multiple times with the items array for each looped response. The green portion in the screenshot below shows the expected behaviour for every looped action.

**Screenshots**
SCR-20221216-gid annotated

  
**App Export**
[dodgy automation-export-1671475925538.tar.gz](https://github.com/Budibase/budibase/files/10261937/dodgy.automation-export-1671475925538.tar.gz)

 - OS: macOS 16.2
 - Browser: Chrome
 - Version: 108

**Additional context**
[found](https://discord.com/channels/733030666647765003/1022503693867823104/1052892320493748286) while helping Discord user Sourceae who has had big problems with this.

PClmnt:
## Description
When there were two loop steps subsequent to the other, the executionOutput of the automation would be incorrect, and loop outputs would be recorded as a step, instead of an item within a step. 

Fixes: #9123 

image

you can see that the final non looped step is represented correctly in the test output. Compared to the example in the linked issue.

### DIFF
--- a/packages/server/src/threads/automation.ts
+++ b/packages/server/src/threads/automation.ts
@@ -419,7 +419,7 @@ class Orchestrator {
 
       // Delete the step after the loop step as it's irrelevant, since information is included
       // in the loop step
-      if (wasLoopStep) {
+      if (wasLoopStep && !loopStep) {
         this._context.steps.splice(loopStepNumber + 1, 1)
         wasLoopStep = false
       }
@@ -438,8 +438,7 @@ class Orchestrator {
         })
         this._context.steps[loopStepNumber] = tempOutput
 
-        loopSteps = undefined
-        wasLoopStep = true
+        loopSteps = []
       }
     }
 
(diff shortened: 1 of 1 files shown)


melohagan (approved):
LGTM!

Issue #9249, opened 2023-01-05
Importing apps with a large number of attachments causes some to not correctly be updated


mike12345567:
**Describe the bug**

**To Reproduce**
Discussion where this was discovered: https://github.com/Budibase/budibase/discussions/9041

The way that we retrieve rows which require the attachment URL to be updated has a limit, need to run through this in a paginated way.

**Expected behavior**
It should correctly update all attachment URLs when importing an app.
  
**App Export**
[NFA_PB_PJ-export-1672755068552.tar.gz](https://github.com/Budibase/budibase/files/10354397/NFA_PB_PJ-export-1672755068552.tar.gz)

PR #9251, opened 2023-01-05
Fix for attachment URL update on import - large number of attachments

mike12345567:
## Description
Fix for #9249 - there was an issue with the updating of attachment URLs when importing an app/backup, this resolves it by utilising pagination in the CouchDB find API, making sure to iterate through all documents when there is a larger number.

Addresses: 
- #9249

### DIFF
--- a/packages/server/src/sdk/app/backups/imports.ts
+++ b/packages/server/src/sdk/app/backups/imports.ts
@@ -43,7 +43,7 @@ function rewriteAttachmentUrl(appId: string, attachment: RowAttachment) {
   }
 }
 
-async function updateAttachmentColumns(prodAppId: string, db: Database) {
+export async function updateAttachmentColumns(prodAppId: string, db: Database) {
   // iterate through attachment documents and update them
   const tables = await sdk.tables.getAllInternalTables(db)
   let updatedRows: Row[] = []
--- a/packages/server/src/sdk/app/backups/imports.ts
+++ b/packages/server/src/sdk/app/backups/imports.ts
@@ -1,5 +1,5 @@
 import { db as dbCore, objectStore } from "@budibase/backend-core"
-import { Database } from "@budibase/types"
+import { Database, Row } from "@budibase/types"
 import { getAutomationParams, TABLE_ROW_PREFIX } from "../../../db/utils"
 import { budibaseTempDir } from "../../../utilities/budibaseDir"
 import { DB_EXPORT_FILE, GLOBAL_DB_EXPORT_FILE } from "./constants"
@@ -17,6 +17,9 @@ import {
 const uuid = require("uuid/v4")
 const tar = require("tar")
 
+// default limit - seems to work well for performance
+const FIND_LIMIT = 25
+
 type TemplateType = {
   file?: {
     type: string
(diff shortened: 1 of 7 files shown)


packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -39,44 +62,45 @@ async function updateAttachmentColumns(prodAppId: string, db: Database) {
...
+            // add new app ID
+            urlParts.unshift(prodAppId)
+            const key = urlParts.join("/")
+            return {
+              ...attachment,
+              key,
+              url: `/${prefix}/${key}`,
+            }
+          })
+        }
```
Rory-Powell, line 97: Could we pull the mapping out into a a function with a dedicated name?
It'd be a bit easier to see the intended purpose then like:

```ts
reWriteAttachmentUrl(appId: string, attachment: RowAtachment): RowAttachment
```

Rory-Powell: There's an alternative we can make to this sort of behaviour, where we use the app id in context to dynamically enrich the url, instead of embedding the app id into the row itself - something we might consider in future.

packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -25,9 +28,29 @@ type TemplateType = {
...
+    limit: FIND_LIMIT,
+  }
+  attachmentCols.forEach(col => (params.selector[col] = { $exists: true }))
+  if (bookmark) {
+    params.bookmark = bookmark
+  }
+  return params
+}
+
 async function updateAttachmentColumns(prodAppId: string, db: Database) {
```
Rory-Powell, line 50: Can we `export` this function so we can add a unit test for it? 

A db aware test that can:
- Create table
- Add attachment rows 
- Call `updateAttachmentColumns`
- Read back rows and ensure urls is in correct format

Rory-Powell: Preferably we could decouple the pagination from `updateAttachmentColumns`. In the ideal scenario the pagination would live in the rows sdk:

```ts

interface PageOpts {
   limit: number,
   bookmark: string
}

interface PageResponse<T> {
   bookmark: string,
   rows: T[] 
}

// specific function for creating `CouchFindOptions` for attachments search
// uses `getPage`
sdk.rows.getAttachmentRows(appId: string, opts: PageOpts): PageResponse<Row>

// generic pagination function
// defaults for `FIND_LIMIT` live here
sdk.rows.getPage(opts: CouchFindOptions): PageResponse<Row>
```

This is beneficial for a few resons:
- Enable us to unit test pagination in isolation
- Enable us to unit test attachment rows retrieval in isolation
- Allow us to supply an arbitrary page size to result in faster test execution
- Remove the concerns of pagination from the test for `updateAttachmentColumns`
- Encapsulate db logic in rows sdk and build up more re-usable functions

Rory-Powell (commented):
I think we'll need some unit tests around this, there's a few different pieces here around
- pagination
- attachment filtering
- url re-writing

that could all benefit from individual tests

mike12345567:
@Rory-Powell  - Sadly, unit testing is difficult until we have the new test system in place - as the `_find` API utilised is not easily available through PouchDB.

packages/server/src/sdk/app/backups/imports.ts
mike12345567: Utilising the context in this case could be a little dangerous, as it is around the creation of the app - the ID may not yet be in context as the app doesn't really exist yet.

Rory-Powell: Ah no for this I mean on read. So we never store the appId in the attachment url, we can enrich it on read, like we do for tenancy and presigned urls

Rory-Powell:
Makes sense! I think it's not that much of a problem, we can still do the suggested refactors and get almost full test coverage. We'll just have to go without the integration on the `getPage` test. This aligns with the goal of the refactor, as we can test all the other parts in isolation using mocks:

- `getAttachmentRows` -> assert that `getPage` is called with the expected pagination params
- `updateAttachmentColumns` -> mock `getAttachmentRows` to return rows, end result and assertions remain unchanged

Issue #9104, opened 2022-12-19
Default value no longer being set on Navigate To action. Causes problems with filtering in version 2.2.3


ShivrajJoshi1996:
Hosting

Internal Budibase Hosting

  - App Version: 2.2.3

**Describe the bug**
Filtering the table using the data from other table is not working.

**To Reproduce**
1. Create two internal tables with one to many relations. 
2. Use the value of relationship column to new column (so that we can use the new column for filtering )
3. On screen 1 click on table row and navigate to other screen
4. On screen 2 show the selected/clicked information on top. (like column value). Take data provider>new repeater>text field to show the value in text field. For data provider, specify the condition as below " Screenshot for Meta data"
5. Based on meta data, provide filtering to table 2. As per "Screenshot to filter below table"

**Expected behavior**
I should be able to filter screen 2 table based on text field which provide data from table 1.

**Screenshots**
1.
![Table 1](https://user-images.githubusercontent.com/121036387/208453775-545b8943-b1cf-414b-8897-6f3f73dbc6f1.png)

2.
![Table 2](https://user-images.githubusercontent.com/121036387/208453800-3fa895fd-a8a2-4b8a-af05-dfd4806a0986.png)

3.
![Screenshot 1 navigate](https://user-images.githubusercontent.com/121036387/208453822-70d525dd-56a8-4b10-8d9c-a799baa95fbb.png)

4.
![Screenshot for Meta data](https://user-images.githubusercontent.com/121036387/208453852-e9aeca02-1c56-44a2-bb5d-6fe787f94712.png)

5.
![Screenshot to filter below table](https://user-images.githubusercontent.com/121036387/208453880-54792095-8aa8-4a91-9b7a-7c43921725aa.png)

  
**Additional context**
For earlier version filtering was achieved using above steps, but since the last update filtering is not possible.

melohagan:
Hey @ShivrajJoshi1996 

I can't reproduce any issue here - I don't think I understand what you're trying to do in steps 4 and 5. 

What data type is the NEW_RELATION field, and why do you need it?

ShivrajJoshi1996:
Hi @melohagan 

Thanks for your replay. let me give you little briefing on each screenshot that I have added.

On 1 and 2 screenshot: 
I created two tables as both table are interdependent means one best practice is having multiple strategy so its 1:Many relationship. Here you will see on both tables we have relationship columns highlighted with grey background .

[...]

Issue #9064, opened 2022-12-15
Skeleton loader a little sluggish, increasing the time before the user can interact with the page


hawkinsjon:
**Hosting**
- Self
  - Method: digital ocean
  - Budibase Version: 2.2.3
  - App Version: 2.2.3

I think the point of the skeleton loader addition was to help show a single loading animation until all of the queries of a given view are completed. 

The way the skeleton loader is set up now, I see some queries speed past the animation, the data fully renders, and then a short moment later the skeleton animation takes over masking all items that had previously rendered, then when the final queries complete, the page is fully rendered.

My feeling after flipping around to a few query heavy pages is that the skeleton loader is being a little greedy and slowing down the speed in which the page is finally rendered and useable for the user.

As a comparison I reverted my app from 2.2.3 back to 2.1.42, refreshed the page and I get snappy "feeling" page loads again.

I'm guessing for a lot of budibase users, this Skelton loader feature will be just fine as it is now. For those with greater queries per page, especially if those queries are from different sources or APIs, I think this experience as it sits now is not good enough. Feels a little clunky. 

If there's a way to tweak or disable this feature and revert back to the old set up which seemed to be an individual spinner per area querying data, that would be great. Maybe in the setting per app?

melohagan:
Hey @hawkinsjon 

I'm not seeing any particular issue myself:
![add_comp](https://user-images.githubusercontent.com/101575380/208111758-7ff9cb58-a3e5-4e41-98ac-c723b4910a9c.gif)

If there's a particular problem you want to showcase, a recording would be very helpful. 

As a bug there really needs to be reproduction steps, otherwise it may be best to raise feedback in either the GitHub discussions, or in our discord channel: https://discord.gg/zPH2EYKQ

I believe that you may have created this thread in Discord? `Skeleton Loader feels slow`

hawkinsjon:
Sure thing, here's two clips, one on version 2.1.42 and the next on version 2.2.3 where you can see the experience a bit.


https://user-images.githubusercontent.com/3512179/208146207-07f14917-72c7-429a-9112-7515a3111ded.mov


https://user-images.githubusercontent.com/3512179/208146220-dc21c55f-cd7f-44cf-a20e-5ffa8e934b9a.mov

melohagan:
Thanks for that @hawkinsjon I can see the visual issue! 

I think seeing your component tree, i.e. the composition of your Data providers and repeaters would also be useful. 

[...]

PR #9256, opened 2023-01-06
Optimise client library performance with skeleton loaders

aptkingston:
## Description
This is a collection of fixes and improvements which attempt to restore some client library optimisations which we sacrificed in order to develop the skeleton loader feature. As that feature fundamentally changes how everything is rendered (and specifically, renders data providers and form components before their data is ready) it introduced a number of edge case bugs depending on how your components were ordered.

These fixes will hopefully fix the majority of issues and reduce some of the exessive API load which was being generated.

I'd like to stress that this will only fix the majority of issues and not fully restore the optimisations we had prior to the skeleton loaders. If we are rendering everything before data is ready then there are just some optimisations that will no longer work, so redundant API calls will just have to be something we accept if we want this feature.

Brief changelog:
- Update form field default value determination so that resetting forms after their schema loads does not cause a field value update unless required
- Update data providers to wait for parent data providers to load before loading their own data. This prevents wasted API calls in the event that nested providers use data from parent providers, such as for filtering which is common
- Use the new `loading` context inside forms, so that components nested inside forms will only render skeletons until the form schema as loaded. This is required as forms cannot dynamically update their schema and require a full reset, so currently we are fully resetting data providers nested inside forms, causing bugs as well as wasted API calls
- Optimise data providers to avoid updates as much as possible
- Included a fix for an issue with button actions

We won't be able to fully remove all redundant API calls without introducing debouncing (which introduces lag to the UI of course), so there are still some cases which will be inefficient. An example is when using a form field as a filter on a data provider, and providing a default value setting to that field. In this scenario, the data provider will initially render without a value for this field, but will then immediately refresh itself as soon as the form field mounts and notifies the form of its existence and default value. This scenario is highlighted in one of the issues raised recently.

Addresses: 
- #9064
- #9104
- #9164
- #9166
- #9069
- #9258

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/ButtonActionDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/ButtonActionDrawer.svelte
@@ -13,7 +13,6 @@
   import { generate } from "shortid"
   import { getEventContextBindings } from "builderStore/dataBinding"
   import { currentAsset, store } from "builderStore"
-  import { selectedScreen } from "../../../../../builderStore"
 
   const flipDurationMs = 150
   const EVENT_TYPE_KEY = "##eventHandlerType"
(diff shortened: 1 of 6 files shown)


Ghrehh (approved):
LGTM

shogunpurple (approved):
LGTM!

PR #9262, opened 2023-01-06
move helm log annotations

jonnymccullagh:
I was unable to see annotations added on our containers so I'm having a go at putting them under the template section. 
Feedback welcome.


Rory-Powell (approved):
LGTM

Issue #9150, opened 2022-12-22
Search Fields on table block fails with `Unknown column '<mytablename>' in 'where clause'`


ttben:
**Hosting**
- Self
  - Method:  
  - Budibase Version: 2.2.3
 
**Describe the bug**
When using `Search Fields` on a table block, 

only the filter on boolean type works, others on Text/Long text type fails with `Unknown column 'sources' in `where clause'`.

I have to specify that "`sources`" is the external MySQL _table name_, but that my data provider for this screen is called "sources-dev".

I don't  know if it is linked to #8787 or #8334 as they were routing related.

Looks like the way budibase generate the query can not be interpreted as mentioned in [https://stackoverflow.com/questions/153598/unknown-column-in-where-clause](https://stackoverflow.com/questions/153598/unknown-column-in-where-clause)

**Screenshots**
The search fields are set on the table block
![image](https://user-images.githubusercontent.com/6167712/209137703-63ed0dd6-34cb-468a-b164-b77dd09df488.png)

They appear correctly on the UI (omit css issue) 
![Screenshot from 2022-12-22 13-46-59](https://user-images.githubusercontent.com/6167712/209137721-3efb9fa8-2ba8-44e1-bd84-4affeae7d33b.png)

But when using filter on name or schedule (respectively Text and Long Text type)  this error appears. The filter works for "enabled" boolean field
![image](https://user-images.githubusercontent.com/6167712/209137368-12ea0719-6c0b-43c2-8329-b4613e78b302.png)

ttben:
Sure !

Here is the SQL script to recreate the DB, thank you for your help!

```
SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO";
SET AUTOCOMMIT = 0;
START TRANSACTION;
SET time_zone = "+00:00";

CREATE TABLE `sources-dev` (
  `id` int(11) NOT NULL,
  `name` varchar(255) NOT NULL,
  `enabled` tinyint(1) NOT NULL,
  `schedule` varchar(12) NOT NULL COMMENT 'CRON FORMAT SCHEDULE',
  `url` varchar(255) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;

ALTER TABLE `sources-dev`
  ADD PRIMARY KEY (`id`),
  ADD UNIQUE KEY `name` (`name`);

ALTER TABLE `sources-dev`
  MODIFY `id` int(11) NOT NULL AUTO_INCREMENT;
[...]

PR #9278, opened 2023-01-09
SQL filtering: Always escape key in LIKE statements 

melohagan:
## Description
LIKE statements were failing for tables with dashes in the name. This bug was reproducible in SQL Server and MySQL.

This is because of this snippet in the **likeKey** function:
```js
if (!key.includes(" ")) {
  return key
}
```
I simply removed this because as far as I understand there is no drawback to always escaping key. 
E.g. `table.column` is valid, but `[table].[column]` is also valid and also supports `[table-name].[column]`

Addresses: 
- https://github.com/Budibase/budibase/issues/9150

### DIFF
--- a/packages/server/src/integrations/base/sql.ts
+++ b/packages/server/src/integrations/base/sql.ts
@@ -232,7 +232,9 @@ class InternalBuilder {
         } else {
           const rawFnc = `${fnc}Raw`
           // @ts-ignore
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
Makes sense - this complicates the output queries a bit more - but they are already essentially un-readable so it doesn't really matter.

PR #9279, opened 2023-01-09
publish helm tarball from develop branch

jonnymccullagh:
I'd like to get helm charts published from the `develop` branch so they can be tested before arriving on the master branch. 

We currently publish helm charts to github pages from our budibase repo but from the `gh-pages` branch. To avoid interfering with the existing helm tarballs, this PR action creates a folder named helm-develop and places the tarballs and helm index in that folder. 

I'm hoping this will allow us to use the following address when pulling helm charts for testing/dev/qa:
```
https://budibase.github.io/budibase/helm-develop
```

### DIFF
--- a/.github/workflows/release-develop.yml
+++ b/.github/workflows/release-develop.yml
@@ -93,6 +93,30 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_API_KEY }}
           RELEASE_TAG: k8s-release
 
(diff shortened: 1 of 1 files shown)


Rory-Powell (commented):
This is already done now here: https://github.com/Budibase/budibase/blob/develop/.github/workflows/release-develop.yml#L149-L179

PR #9284, opened 2023-01-09
Renamed proxy docker image tag from proxy-service to budibase/proxy

deanhannigan:
## Description

Discovered this with @aptkingston. The `build:docker:proxy` step was mistakenly naming the docker proxy image `proxy-service` instead of `budibase/proxy`.

The dev compose was recently updated to allow users to specify their proxy address and to now use `budibase/proxy` as the container image.

With this change the proxy image will now be rebuilt with the correct tag and the new config changes will work as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "build:specs": "lerna run specs",
     "build:docker": "lerna run build:docker && npm run build:docker:proxy && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh $BUDIBASE_RELEASE_VERSION && cd -",
     "build:docker:pre": "lerna run build && lerna run predocker",
-    "build:docker:proxy": "docker build hosting/proxy -t proxy-service",
+    "build:docker:proxy": "docker build hosting/proxy -t budibase/proxy",
     "build:docker:selfhost": "lerna run build:docker && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh latest && cd -",
     "build:docker:develop": "node scripts/pinVersions && lerna run build:docker && npm run build:docker:proxy && cd hosting/scripts/linux/ && ./release-to-docker-hub.sh develop && cd -",
     "build:docker:airgap": "node hosting/scripts/airgapped/airgappedDockerBuild",


shogunpurple (commented):
I think this may potentially cause an issue with `hosting/scripts/linux/release-to-docker-hub.sh`

```
docker tag proxy-service budibase/proxy:$tag
docker tag app-service budibase/apps:$tag
docker tag worker-service budibase/worker:$tag
```
That script provides the release tag to the built images, which is based on their previous tag. The reason we don't tag them directly with just `budibase/proxy`/`budibase/apps` etc is that pushing those images would then update the `latest` tag.

What exactly was the problem you were seeing with the current setup?

aptkingston:
Yea the main issue was that building the image locally didn't actually update the image used by the dev stack, since they're tagged differently. 

Without updating the tagging then, another solution would just be to pull the latest budibase proxy image as part of bringing up the dev stack? Since we're refering the published image, I think it makes sense that we should ensure we're always running the latest proxy when developing, as otherwise issues like this will happen over time as the local proxy image becomes stale.

shogunpurple:
Ah - that makes sense now. Running the manage script will just spin up the docker stack, and won't actually update it. You can of course run: `docker compose pull` in the `docker-compose.dev.yml` dir and it will get the latest. However, with the setup that now exists in the dev file:
```
  proxy-service:
    container_name: budi-nginx-dev
    restart: on-failure
    image: budibase/proxy:latest
    volumes:
      - ./nginx.dev.conf:/etc/nginx/templates/nginx.conf.template
      - ./proxy/error.html:/usr/share/nginx/html/error.html
    ports:
      - "${MAIN_PORT}:10000"
    depends_on:
      - minio-service
      - couchdb-service
    extra_hosts:
      - "host.docker.internal:host-gateway"
    environment:
      - PROXY_ADDRESS=host.docker.internal
```
The bind mount exists with the `nginx.dev` config so any local NGINX changes should take effect. I think this is more just a one time problem, in that we made significant changes to `budibase/proxy`, which doesn't happen very often. 

The only time you would actually need to update this image now is if we do something similar again, otherwise the bind mount pointing to the nginx config should take care of it. I think this is probably:
- safe to close if everyone just updates to the latest proxy image
- If we really want to be safe we could add a pull of `budibase/proxy` to the manage script to ensure the latest is always there. However, it will pull new versions quite often given we roll a new proxy image with every release

PR #9288, opened 2023-01-09
JSON import changes

Ghrehh:
## Description
Bit of a chunky one, so apologies in advance. Also my first time working with the backend, so sorry if I've made some major faux pas there.

# Changes
- Updates our import functionality to also accept JSON in two formats: either an array of objects representing a row, or an object with two keys specifying the schema and the data which can be used to autofill the correct fields for the schema during import
- Updates the export functionality to allow users to export in the above specified json formats.
- Importing to an existing table now specifies if any columns you're attempting to import aren't accepted by the schema

Screenshot 2023-01-09 at 16 14 24

# Implementation details

- The modals to import data to a new and existing tables share some functionality and were previously one component, but in my opinion they are different enough to warrant being separate for now, rule of threes and all that.
- I used type predicates and runtime type checks on data passed up from the client, I have no idea if this is correct/our style, please let me know.


deanhannigan:
Nice work! Had a play around with this and it appears to be working nicely.

Some minor UI suggestions:

- The text in the import modal alluding to `CSV` would have to be updated to include mention of JSON support.
- Could the UI for the unrecognised columns reflect the UI of the successful elements above? I believe we already do this when creating a new table from a CSV import, e.g. 
Screenshot 2023-01-10 at 10 54 35
Maybe state that an ignored column will be `Skipped`, `Ignored` or `Invalid`

packages/types/src/documents/app/table.ts
```diff
@@ -69,7 +69,7 @@ export interface Table extends Document {
   constrained?: string[]
   sql?: boolean
   indexes?: { [key: string]: any }
-  dataImport?: { [key: string]: any }
```
mike12345567, line 72: Just had a look through here - noticed the type change, if this means that the API has changed from accepting the `dataImport` property, then might need to update the API doc comments in the `routes/table.ts` file, theres still a few mentions of it there!

Ghrehh: Ah yes sorry I was meant to ask someone about that, also, I notice there were comments above each route on the server describing the payload etc, are those written by hand or generated?

mike12345567: Those were written by hand, as part of plans for an APIDoc/OpenAPI generator for the internal API, although the docs are barely finished, its just useful for the table/rows APIs since they are so core to the whole system!

Ghrehh:
As @deanhannigan pointed out, the Auto ID behaviour is a little broken at the moment, but it's currently broken for the existing imports functionality also.

I believe the issue is that when you import a bunch of rows, say 100, it will respect whatever the current Auto ID is set to in the schema. If there are four user created rows, and these 100 are imported their IDs will range from 5-105 which is correct. However if you imported those same rows again, their IDs will also be set to 5-104, instead of 105-204 as expected. This seems to be because the `db.bulkDocs()` function doesn't update the tables schema to have the correct Auto ID, but I could be wrong.

Either way it seems to not be caused by this PR so i'll punt on it for now.

packages/backend-core/src/featureFlags/index.ts
```diff
@@ -36,7 +35,7 @@ export function isEnabled(featureFlag: string) {
 }
 
 export function getTenantFeatureFlags(tenantId: string) {
-  const flags = []
+  const flags = ["USER_GROUPS"]
```
aptkingston, line 38: Was this a hack to get groups working locally by any chance? :laughing: Just forgot to remove it.
The easier way to do this is add the tenant feature flags to your `.env` file in the worker package like so:
```
TENANT_FEATURE_FLAGS=*:LICENSING,*:USER_GROUPS
```

aptkingston (commented):
LGTM! Just need to remove the feature flags hack.

One general question though - it looks like `TableDataImport` has basically been duplicated and changed a bit to create `ExistingTableDataImport`. Was there no way to keep it as one component and customise it with props? Maybe their functionality is so different that it needs to be split which is fine, but at a quick glance it looks like a decent chunk of duplication. You don't need to go and refactor this if that's time consuming, but was just wondering about the motivation for it!

packages/backend-core/src/featureFlags/index.ts
Ghrehh: I'm a spoon, aye I'll remove this. I'm pretty sure I copied this from dev docs somewhere but I could be wrong.

Ghrehh:
@aptkingston I mentioned it a bit in the main body of the PR but yeah in my opinion they're just about different enough to warrant being separate components, there's UI that's exclusive to both forms of it. Could definitely be argued that they should be DRYed up, but IMO a bit of repetition is better than the abstraction. Open to changing it though, not married to it by any means.

aptkingston (approved):

PR #9296, opened 2023-01-10
adjust service annotation locations

jonnymccullagh:
My previous attempts at this were lacking. This time I have tested with the `helm template` command.
I moved the annotations under the deployment > spec > template (with an indent of 8) where I think they belong. `helm template` shows this substitution is into the correct place but I still need to test against QA when this is on the develop branch.

Similarly I moved the annotations location in the `values.yaml` file so they appear under: services > servicename > annotations 
which feels like a more intuitive place to have them.

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -20,8 +20,8 @@ spec:
       annotations:
         kompose.cmd: kompose convert
         kompose.version: 1.21.0 (992df58d8)
-{{ if .Values.globals.appServiceAnnotations }}
-{{ toYaml .Values.globals.appServiceAnnotations | indent 4 }}
+{{ if .Values.services.apps.appServiceAnnotations }}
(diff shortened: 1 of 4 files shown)


charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -20,8 +20,8 @@ spec:
       annotations:
         kompose.cmd: kompose convert
         kompose.version: 1.21.0 (992df58d8)
-{{ if .Values.globals.appServiceAnnotations }}
-{{ toYaml .Values.globals.appServiceAnnotations | indent 4 }}
+{{ if .Values.services.apps.appServiceAnnotations }}
```
Rory-Powell, line 23: Given the namespacing we not have under `services.app` we could shorten the variable name to just `annotations`  i.e. 

```
Values.services.apps.annotations
Values.services.proxy.annotations
Values.services.worker.annotations
```

This is a little easier to read and more consistent

Rory-Powell (approved):
LGTM 👍  Just a comment about shortening the annotations paths

PR #9312, opened 2023-01-11
Default source to Github

melohagan:
## Description
Github URL is used 99% of the time when adding plugins. Changed to be the option selected by default.

## Screenshots
Screenshot 2023-01-11 at 15 18 19


shogunpurple (approved):
LGTM, makes sense.

mike12345567 (approved):

PR #9330, opened 2023-01-12
Fix local dev hosting and allow primary foreign keys

melohagan:
## Description
As discussed, needed to cherry-pick PROXY_ADDRESS dev environment changes.


Rory-Powell:
Proxy changes all good. 

Primary key changes I would recommend we run through develop, and get some dedicated QA.

shogunpurple (approved):
LGTM - agree with @Rory-Powell, think the NGINX change and this could be separate PRs with one going into `develop`

Issue #8879, opened 2022-11-30
Portal redesign


aptkingston:
There wasn't an existing issue for this so thought I should create one. We're hopefully going to implement this in December, or at least finish it the current implementation and gather more feedback

Link to new designs:
https://www.figma.com/file/T57nnTekuyKgZhxdBCbgGM/Portal-re-design?node-id=0%3A1

This is probably over 50% completed on a lab day branch of mine already (`cheeks-lab-day-portal-redesign`).

MihailHadzhiev2022:
![Screenshot 2023-01-19 at 11.46.18.png](https://images.zenhubusercontent.com/623866329f3f8e7223d95e00/288ff699-59bc-4063-8b61-ebda05b35b60)

PR #9336, opened 2023-01-13
Portal redesign

aptkingston:
## Description
This PR refactors the entire portal to give it a fresh new look and make it entirely responsive. The routing structure has also been updated slightly to have a more natural hierarchy.

This is pretty unreviewable given the scope of the changes, but you can see the new designs here:
https://www.figma.com/file/T57nnTekuyKgZhxdBCbgGM/Portal-re-design?node-id=0%3A1

The portal should feel faster, look much better, be less buggy, be more intuitive to use and just generally be a nicer experience.

Many pages have also been refactored/rewritten (basically just addressing longstanding tech debt) to do things like:
- Improve loading states
- Reduce redundant API calls
- Fix various UI bugs
- Break some monolithic pages into reusable components

Some changes not within the remit of just redesign:
- Developer users now have readonly access to the users and groups pages
- Developer users now have access to the usage page
- Theme and API key are now managed from the user dropdown, rather than settings
- The app overview section is now fully routed rather than just using a URL param and tabs
- Pickers (used for selects/action menus) have been rewritten to use popovers (which have also been rewritten) to now use a svelte portal targetting the root DOM node. This has a very large benefit of no longer requiring overflow to be visible for containers that use these components. The largest benefit of this is that dropdowns basically work inside tables now, and tables can also hide their overflow to be properly responsive.

Addresses: 
- #8879

## Screenshots
There are too many pages to screenshot them all, but here's a few on both desktop and mobile.

### Apps page
image
image

### Templates page
image
image

### Users page
image
image

### User details page
image
image

### Plugins page
image
image

### Settings page (auth subpage)
image
image

### App overview page
image
image

### App access page
image
image

### Automation history page
image
image
image
image

### User dropdown
image

### Responsive tables

https://user-images.githubusercontent.com/9075550/212284152-e76bdb08-d138-4984-b50a-c85c2f686747.mp4

### Mobile menu

https://user-images.githubusercontent.com/9075550/212284303-b1bb1f3d-fa9e-4a57-848c-b7ed744ffbfd.mp4


shogunpurple (approved):
LGTM - Think we are better to just get this one in given it's size and the fact we can E2E test it

Issue #7220, opened 2022-08-10
Deleting User from Main Users Table Does Not Delete App User Row


pbarmak:
If I create a user in the main, system-wide Users portal table, then assign that user to an app, the system creates a row in the app's Users budibasedb table. I can then add custom columns to the apps's Users table and update that user row as usual. That works.

If I then delete the user from the system-wide Users table in the portal, it deletes that main user, but it does not delete the user row from each of the apps Users tables. 

Since the apps' Users tables are somehow tied to the system-wide one, I cannot clean up the specific app's User table. It won't let me delete a row there. I can't even update any custom columns in that table (it gives me a "Document update conflict." error).

I think, when a main user is deleted, the system needs to either a) delete all apps' Users table rows for that user or b) unlink the user and allow us to manually delete an orphaned user row from the app Users table.

In the meantime, anyone know how I can clean up the app's Users table without destroying my app?

aderici:
Within your app you can use delete row action in a button and delete an apps user row. I'm not sure if this is supported behavior but it does delete

pbarmak:
@aderici That's interesting, I'll try it, thank you. Though that maybe is concerning, if we can delete a row from the Users table via a button action while we can't from the actual Data interface. But it's good to know, thanks for pointing it out.

mlomax:
> 
We have this problem too.
The lack of cleanup also seems to have an effect on re-adding a user.

I was having issues with SSO so deleted a user from the main screen so I could run a test by re-adding. 
The email was deleted from the user portal but when I come to re-add it I get a reason of "unavailable" and the email isn't added.

On checking Budi hasn't deleted the user from the app-user tables, although it has set to inactive. But I think this is why I can't re-add the email.

michal-81:
Having the same issue. Are they any plans to fix this?

PR #9337, opened 2023-01-13
Bug - #7220 removing user from auth does not remove from app

adrinr:
## Description
Removing a user from the access tab does not remove the user within the app

Addresses: 
- #7220: This bug was discovered while investigating a related bug, not 100% the one described in the bug ticket

## App Export
- Not required, it can be reproduced with an empty app

## Screenshots
https://user-images.githubusercontent.com/15987277/212298953-af265e5b-60e2-4842-84a6-fc06903279be.mov

### DIFF
--- a/packages/server/src/api/controllers/user.ts
+++ b/packages/server/src/api/controllers/user.ts
@@ -189,16 +189,11 @@ export async function removeUserFromApp(ctx: BBContext) {
       try {
         metadata = await db.get(metadataId)
       } catch (err) {
+        console.warn(`User cannot be found in the app`, { userId, appId })
         return
       }
 
-      let combined = {
-        ...metadata,
-        status: constants.UserStatus.INACTIVE,
-        metadata: rolesCore.BUILTIN_ROLE_IDS.PUBLIC,
-      }
-
-      await db.put(combined)
+      await db.remove(metadata)
     })
   }
   ctx.body = {
--- a/packages/server/src/api/controllers/user.ts
+++ b/packages/server/src/api/controllers/user.ts
@@ -28,9 +28,10 @@ export async function syncUser(ctx: Ctx) {
     }
   }
 
-  let previousApps = isUser(previousUser)
-    ? Object.keys(previousUser.roles).map(appId => appId)
-    : []
+  let previousApps =
+    previousUser && isUser(previousUser)
+      ? Object.keys(previousUser.roles).map(appId => appId)
+      : []
 
   const roles = deleting ? {} : user.roles
   // remove props which aren't useful to metadata
--- a/packages/server/src/api/controllers/user.ts
+++ b/packages/server/src/api/controllers/user.ts
@@ -4,19 +4,21 @@ import { getGlobalUsers, getRawGlobalUser } from "../../utilities/global"
 import { getFullUser } from "../../utilities/users"
 import {
   context,
-  constants,
   roles as rolesCore,
   db as dbCore,
 } from "@budibase/backend-core"
-import { BBContext, Ctx, User } from "@budibase/types"
+import { BBContext, Ctx, isUser, User } from "@budibase/types"
 import sdk from "../../sdk"
 
-export async function syncUser(ctx: BBContext) {
+export async function syncUser(ctx: Ctx) {
   let deleting = false,
     user: User | any
   const userId = ctx.params.id
+
+  const previousUser = ctx.request.body?.previousUser
+
   try {
-    user = await getRawGlobalUser(userId)
+    user = (await getRawGlobalUser(userId)) as User
   } catch (err: any) {
     if (err && err.status === 404) {
       user = {}
@@ -25,6 +27,11 @@ export async function syncUser(ctx: BBContext) {
       throw err
     }
   }
+
+  let previousApps = isUser(previousUser)
+    ? Object.keys(previousUser.roles).map(appId => appId)
+    : []
+
   const roles = deleting ? {} : user.roles
   // remove props which aren't useful to metadata
   delete user.password
@@ -40,8 +47,9 @@ export async function syncUser(ctx: BBContext) {
       .filter(entry => entry[1] !== rolesCore.BUILTIN_ROLE_IDS.PUBLIC)
       .map(([appId]) => appId)
   }
-  for (let prodAppId of prodAppIds) {
+  for (let prodAppId of new Set([...prodAppIds, ...previousApps])) {
     const roleId = roles[prodAppId]
+    const deleteFromApp = !roleId
     const devAppId = dbCore.getDevelopmentAppID(prodAppId)
     for (let appId of [prodAppId, devAppId]) {
       if (!(await dbCore.dbExists(appId))) {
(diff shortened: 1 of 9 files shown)


adrinr:

adrinr:
recheck

packages/worker/src/sdk/users/users.ts
```diff
@@ -265,6 +280,11 @@ export const save = async (
     await eventHelpers.handleSaveEvents(builtUser, dbUser)
     await addTenant(tenantId, _id, email)
     await cache.user.invalidateUser(response.id)
+
+    for (const appId of appsToRemove) {
+      await apps.removeUserFromApp(_id, appId)
+    }
```
Rory-Powell, line 286: The looping network call might have some performance problems if this request is ever used with more than one app. I think we could get the server to perform this logic, as a result of the existing `syncUserInApps` call. It would be nice to contain all the user -> app user sync in one place like this. We already have some logic in there for when the user is actually being deleted. 

I assume it is not working in this case as the roles being removed from the user are already saved to the db at the time of the request being sent and so the server never knows to delete the relationship. We could update the existing sync to receive a body: 
```
{
   "previousUser": {} // supply the `dbUser` if it exists
}
```

Then in the sync code here:
https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/controllers/user.ts#L14
we can perform a diff to get the list of removed apps and action on them, like is being done above.

packages/worker/src/sdk/users/users.ts
```diff
@@ -265,6 +280,11 @@ export const save = async (
     await eventHelpers.handleSaveEvents(builtUser, dbUser)
```
Rory-Powell, line 280: Just a heads up for learning, we have some similar logic to the above being triggered inside this call:
https://github.com/Budibase/budibase/blob/develop/packages/worker/src/sdk/users/events.ts#L17-L52
see: `unassignAppRoleEvents`

If we were going to continue on the path of sending multiple requests from worker -> server I would recommend to integrate here. However the alternate approach of a single api call to sync as mentioned already I think would be better.

packages/worker/src/sdk/users/users.ts
```diff
@@ -257,6 +261,17 @@ export const save = async (
...
   }
 
+  let appsToRemove: string[] = []
+  if (dbUser && instanceOfUser(user)) {
+    const newRoles = Object.keys(user.roles)
+    const existingRoles = Object.keys(dbUser.roles)
+
+    appsToRemove = existingRoles.filter(r => !newRoles.includes(r))
+    if (appsToRemove.length) {
+      console.log("Deleting access to apps", { appsToRemove })
```
mike12345567, line 271: Generally we try not to leave console logs in API requests, if there is useful information here that could be logged in the form of a warning or error then we would use our logging functionality (which we can detect easily in Elastic) under `import { logging } from "@budibase/backend-core"` e.g. `logWarn` or `logAlert`.

packages/server/src/api/controllers/user.ts
```diff
@@ -173,3 +173,30 @@ export async function getFlags(ctx: BBContext) {
...
+      continue
+    }
+    await context.doInAppContext(appId, async () => {
+      const db = context.getAppDB()
+      const metadataId = generateUserMetadataID(userId)
+      let metadata
+      try {
+        metadata = await db.get(metadataId)
+      } catch (err) {
+        console.warn(`User cannot be found in the app`, { userId, appId })
```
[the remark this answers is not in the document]
adrinr: I kept is as a warn and return because of the similar functionality on the sync function above. Do you think we should throw a 400 in there as well? Or might we have a different use case?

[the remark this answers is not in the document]
mike12345567: I think for the sync endpoints we decided just not to error, as its nots a specific user action, I think in this case I would just replace the `console.warn` with a `logWarn` instead.

packages/server/src/api/controllers/user.ts
```diff
@@ -173,3 +173,30 @@ export async function getFlags(ctx: BBContext) {
...
+        return
+      }
+
+      await db.remove(metadata)
+    })
+  }
+  ctx.body = {
+    message: `User ${userId} deleted from ${prodAppId} and ${"devapp"}.`,
```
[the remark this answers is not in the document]
adrinr: Good catch, I forgot to add the TODO when I did that 💩

packages/worker/src/sdk/users/users.ts
adrinr: I like the idea :)

packages/worker/src/sdk/users/users.ts
adrinr: I implemented the single app approach

packages/types/src/sdk/koa.ts
```diff
@@ -41,7 +41,7 @@ export interface UserCtx<RequestBody = any, ResponseBody = any>
 }
 
 /**
- * Deprecated: Use UserCtx / Ctx appropriately
+ * @deprecated: Use UserCtx / Ctx appropriately
```
adrinr, line 44: Changing this to the keyword, so the IDEs pick it up

mike12345567 (approved):
LGTM!

packages/server/src/api/controllers/user.ts
```diff
@@ -25,6 +27,12 @@ export async function syncUser(ctx: BBContext) {
       throw err
     }
   }
+
+  let previousApps =
+    previousUser && isUser(previousUser)
```
Rory-Powell, line 32: We can remove the need for the `isUser` here by using parameters on the `ctx: Ctx`. 

Now that the sync user request has a body, we can add a new type here: https://github.com/Budibase/budibase/blob/develop/packages/types/src/api/web/user.ts

Like:

```ts
export interface SyncUserRequest {
   previousUser: User
}
```

we can then update the function signature to look like:
```ts
export async function syncUser(ctx: Ctx<SyncUserRequest>) {
```

Now our `ctx.request.body` will automatically infer the `previousUser` as a `User` object

packages/worker/src/utilities/appService.ts
```diff
@@ -20,11 +21,11 @@ async function makeAppRequest(url: string, method: string, body: any) {
   return fetch(checkSlashesInUrl(env.APPS_URL + url), request)
 }
 
-export async function syncUserInApps(userId: string) {
+export async function syncUserInApps(userId: string, previousUser?: User) {
   const response = await makeAppRequest(
     `/api/users/metadata/sync/${userId}`,
     "POST",
-    {}
+    { previousUser }
```
Rory-Powell, line 28: With the new type definition from above we can add

```ts
const body: SyncUserRequest = {
   previousUser
}

  const response = await makeAppRequest(
    `/api/users/metadata/sync/${userId}`,
    "POST",
    body
```

So that this api request will always align with future changes to the type definition

Rory-Powell (commented):
Thanks for addressing the endpoint change 👍 

Before we merge this one, we should add a test case for the new scenario here:
https://github.com/Budibase/budibase/blob/develop/packages/server/src/api/routes/tests/user.spec.js#L163

The existing test logic is a bit light, we should really be doing some assertions on the results of the sync. Any improvements you can make there would be great, particularly for the new scenario of sending the previous user with some deleted app roles

adrinr:
> 

Sure, I was looking for an existing test and I could not find it :)

Issue #8960, opened 2022-12-07
Relationship cannot be deleted


yuppox:
**Hosting**

- Self
  - Method: <method> Docker single image
  - Budibase Version: 2.1.42

**Describe the bug**
A relationship cannot be deleted.

I started an app, using a MySQL database. I tried to put in some relationships and then I added a table. The relationship I was using didn't look right. I went back to change the relationship. I couldn't get it configured the way I wanted so I tried to delete it. It didn't go away.

I deleted all the relationships except this one. It continues to tell me it has been deleted - `Datasource updated successfully` - but nothing happens. If I modify it, a new relationship is created and I can delete that one.

EDIT:
It looks like the columns Budibase originally created are still there. That could be the reason it won't delete the relationship. I tried to edit it to match those and I get `Cannot build name` in the Tables column.

**To Reproduce**
1. Create an app with a MySQL database.
2. Add a valid relationship between two tables.
3. Click the relationship to edit it.
4. Clear the "To table column" field and click "Delete".

The relationship is now broken and cannot be deleted. The old "To table column" name is in use (but not visible) and cannot be deleted.

![broken_relationship](https://user-images.githubusercontent.com/39724098/206521623-67a81a65-72e1-442c-9bb2-a7153708c070.png)

**Expected behavior**
The relationship can be deleted.

**Screenshots**
![no_delete](https://user-images.githubusercontent.com/39724098/206078331-340be69d-3a59-4887-a244-587d080c935c.png)
  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox
 - Version 107.0

yuppox:
> Is unhelpful.

I understand but I don't know what happened. I'll try to work on it and see if I can reproduce the issue.

yuppox:
Updated with repro steps.

Issue #6029, opened 2022-05-23
MySQL: support primary foreign keys 


palsivertsen:
**Hosting**

- Self
  - Method: docker compose (testing locally)
  - Budibase Version: 1.0.169
  - App Version: N/A
- Cloud
  - Tenant ID: N/A

**Describe the bug**
I have two tables, "a" and "b", with a one to one relationship where table "b" has a foreign key ("a_id") that references the primary key in table "a". While defining this relationship the foreign key is missing as an option from the dropdown menu.

![image](https://user-images.githubusercontent.com/11890143/169812665-581d9bb6-a7a0-42ed-94d4-9c8b1255cefa.png)

**To Reproduce**
1. Setup a MySQL 5.6 instance with the following tables:
```
CREATE TABLE `a` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `name` varchar(50) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
CREATE TABLE `b` (
  `a_id` bigint(20) NOT NULL,
  `additional_field_1` varchar(50) DEFAULT NULL,
  `additional_field_2` int(11) DEFAULT NULL,
  PRIMARY KEY (`a_id`),
  CONSTRAINT `b_ibfk_1` FOREIGN KEY (`a_id`) REFERENCES `a` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
2. Add the database as a datasource in Budibase
3. Select "Define relationship" under "Relationships"
4. Select table "a" as from table
5. Select "id" column as Primary key
6. Select table "b" as to table
7. Open drop down menu for foreign key and see that the column "a_id" is missing

**Expected behavior**
I should be able to see and select any column from "b" table as foreign key.

**Screenshots**
See above

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser Google Chrome Version 101.0.4951.64 (Official Build) (64-bit)

palsivertsen:
Issue seems to be related to the fact that `b.a_id` is both a primary and foreign key at the same time.

palsivertsen:
This is still an issue in version `1.0.195`

palsivertsen:
Still an issue in `1.2.52`

palsivertsen:
I suspect the `b.a_id` column is filtered out from the view since it's both a primary and foreign key.

palsivertsen:
[...]

PR #9338, opened 2023-01-13
CreateEditRelationship modal fixes

melohagan:
## Description
I found a UI bug with the recent define relationship modal change. 
The relationship had been setup correctly with the appropriate keys (in this case `dept_no`).
However the code was only taking the first primary key it found within the edit modal. 

Added some code to get the primary key from the relationship rather than just taking the first primary field from the array.

I also noticed the validation in the define relationships picker was quite gnarly and broken in some places (i.e. it would allow you to save cleared fields that should be required).
See the gif for the fix. 

Added a bug fix for relationship deleting. Now using original names instead of the selected names for deletion.

### Validation refactor
I was having a lot of issues with the reactive validation getting into weird states. Every bug I fixed seemed to introduce another one. I've changed the validation to only fire on click of the save button like a traditional form. Should help to reduce the buggy-ness, and imo it's a nicer UX.  

I have also swapped the *from column name* and *to column name*. This one is subjective, but personally I find it more congruent to match the from column name with the from table name.

Addresses: 
- https://github.com/Budibase/budibase/issues/6029
- https://github.com/Budibase/budibase/issues/8960

## Screenshots
| ISSUE | FIX |
| --- | --- |
| Screenshot 2023-01-13 at 11 45 43 | Screenshot 2023-01-20 at 10 50 24 |


mike12345567 (approved):
This looks good, pulled it down and had a play around - any improvements in this area are just great!

Going to merge as I think this has implications for other PRs.

Issue #9351, opened 2023-01-16
Unable to deactivate Google/OIDC auth when the other fields are empty


andz-bb:
### Discussed in https://github.com/Budibase/budibase/discussions/9348

<div type='discussions-op-text'>

<sup>Originally posted by **sweekark** January 16, 2023</sup>
![image](https://user-images.githubusercontent.com/3829871/212606849-6411dfcb-03f5-472f-8d14-f2cfe3fc38d6.png)
although the toogle is turned off for google auth 

its not getting saved.</div>

## Checklist
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: dev stack
  - Budibase Version: 2.2.12-alpha.18
  - App Version: n/a
- Cloud
  - Tenant ID: ddconst

**Describe the bug**
When a user tries to deactivate an auth method and then save the settings, the save button does nothing unless the other fields are filled. No error message is shown.

**To Reproduce**
1. Go to Portal > Manage > Auth
2. Click the Activated toggle for Google or OIDC
3. Click save. If the other fields for that auth method are empty, the save button does nothing

**Expected behavior**
You should be able to deactivate Google Auth without having to enter other config for it. At the bare minimum a validation error message should be displayed.

**Screenshots**
https://www.loom.com/share/717f799c184b4eb68ad1a84935e97b92
  
**Desktop (please complete the following information):**
 - OS: macOS 13.1
 - Browser: Chrome
 - Version: 108

**Additional context**
User also reported the issue here: https://discord.com/channels/733030666647765003/1064430959736393789

PR #9361, opened 2023-01-16
fix auth config validation for saving google/oidc auth deactivation

andz-bb:
## Description
If the config fields for Google/OIDC auth were empty the save button would do nothing on click. This was a problem for users wanting to deactivate those methods of authentication while the config fields were empty. I updated the behaviour so that it only requires the config fields to have values if the activation toggle is set to true. This means that if you want to disable the method of authentication you can do so without changing any other fields.

Addresses: 
#9351 - raised here: #9348

## Screenshots
OIDC config validation


andz-bb:
I have read the CLA Document and I hereby sign the CLA :)

andz-bb:
recheck

shogunpurple (approved):
LGTM! Nice work to flatten some of that nested conditional logic, and to also lock this down on the backend with Joi validation. Good job 👍

PR #9368, opened 2023-01-17
Test - #9339 sqlpostgres row api test suite

### LINKED ISSUE #9339: SQL/Postgres Row API test suite
At the minute we have very limited coverage of our rows API when it comes to external datasources (plus). While we have tested the internal table row API fairly well (apart from the `search` endpoint) we have essentially zero test cases for external tables - a feature that is heavily utilised and will likely see some fairly large changes in the near future. There are a few endpoints of the rows API that we would like to test a bit more fully:
- [x] `POST /api/:tableId/rows` - creating a new row in an external table
- [x] `PATCH /api/:tableId/rows` - updating an existing row
- [x] `DELETE /api/:tableId/rows` - deleting an existing row
- [x] `GET /api/:tableId/rows/:rowId` - getting an individual row
- [x] `POST /api/:tableId/search` - performs a search request based on the `QueryJson` structure
- [x] `GET /api/:tableId/:rowId/enrich` - enriches the supplied row with all expected/attached related rows - this only performs one layer of relationship enrichment.
- [x] `GET /api/:tableId/rows` - gets all rows within a table, this endpoint is unpaginated and does not perform well, but is still used in a few places - this requires testing until it is removed fully (in future it will be replaced by the `search` endpoint).

These are the main endpoints that need to be tested, there are other rows within the Rows API, but they are either deprecated or aren't really much of a test case (like the `exportRows` which just fetches all rows and returns as a CSV/JSON file).

The process of performing these tests will involve the following:
1. Creating a Postgres datasource, which connects to a test database, pre-seeded with some data and tables - the test case could create the tables/a different schema to work within, but its probably easier to just pre-seed the database. This will use the `POST /api/datasources` to create the datasource and configure it to point at the Postgres instance.
2. Fetching the table schemas into Budibase to work with, using the `POST /api/datasources/:datasourceId/schema` - the test should confirm these schemas conform with the expected structure. In the `datasource` there will be an `entities` property, which can be validated in someway.
3. Now that the tables are created, relationships between them can be tested, we should confirm throughout the tests that the many-to-many (with a junction table) and one-to-many work as expected. The relationships can be added using the `PUT /api/datasource/:datasourceId` endpoint to update the table structures of the two tables that are related, this will make Budibase aware of the relationship. The easiest way to do this is likely using the Budibase UI to create the relationships against your test data, copying the structures through Fauxton and having Budibase simply add this to the table structures.
4. Lastly, run through the test cases, paying particular attention to the `/search` endpoint, running through a few different options of searching, including filtering on some related fields.

Example relationship structures to add to table (point 3) below.
```
Tasks table:
	"persons": {
		"fieldName": "personid",
		"foreignKey": "qaid",
		"name": "persons",
		"relationshipType": "one-to-many",
		"tableId": "datasource_plus_de9ccb65cd2448a68bba284180c4d650__persons",
		"type": "link"
	}
Persons table:
	"tasks": {
		"fieldName": "qaid",
		"foreignKey": "personid",
		"main": true,
		"name": "tasks",
		"relationshipType": "many-to-one",
		"tableId": "datasource_plus_de9ccb65cd2448a68bba284180c4d650__tasks",
		"type": "link"
	}
```

adrinr:
## Description
Adding integration tests for the row api using postgres

Addresses: 
- https://github.com/Budibase/budibase/issues/9339

## App Export
- Not required

## Screenshots


hosting/docker-compose.test.yaml
```diff
@@ -45,3 +45,11 @@ services:
       - 6379
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
+
+  postgres:
```
adrinr, line 49: This will be added to all the tests, I am looking for a way to spin up this in a smarter way

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -360,6 +368,14 @@ class TestConfiguration {
       [constants.Header.CSRF_TOKEN]: this.defaultUserValues.csrfToken,
       ...extras,
     }
+
+    if (!isInternal) {
```
Rory-Powell, line 372: It looks like the below cookies are a repeat of the default?

packages/server/src/integrations/base/sql.ts
```diff
@@ -89,13 +89,26 @@ function parseFilters(filters: SearchFilters | undefined): SearchFilters {
 
 function generateSelectStatement(
   json: QueryJson,
-  knex: Knex
-): (string | Knex.Raw)[] {
+  knex: Knex,
+  excludeJoinColumns = false
```
Rory-Powell, line 93: I think we should use `opts` here:
```ts
opts: { excludeJoinColumns: boolean } = { excludeJoinColumns: false }
```

So that the usage of the function is easier to read and reason about. 
```ts
generateSelectStatement(json, knex, true) // before
generateSelectStatement(json, knex,  { excludeJoinColumns: true }) // after
```

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -0,0 +1,663 @@
...
+          const rowData = makeRandomRow()
+          return {
+            rowData,
+            row: await createRow(rowData, tableId || postgresTable._id),
+          }
+        })
+    )
+  }
+
+  test("validate table schema", async () => {
```
Rory-Powell, line 207: Should use `it` instead of `test` to be consistent

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -0,0 +1,663 @@
...
+      type: "datasource",
+      _id: expect.any(String),
+      _rev: expect.any(String),
+      createdAt: expect.any(String),
+      updatedAt: expect.any(String),
+      entities: expect.any(Object),
+    })
+  })
+
+  describe("create a row", () => {
```
Rory-Powell, line 237: Personal preference when writing api `describe` blocks is to use the verb / endpoint combination. I find it's easier to grep the test file for the block I want and sets a easy to extend standard. ie:

```ts
describe("POST /api/:tableId/rows")
describe("DELETE /api/:tableId/rows")
```

This isn't very common across our tests, but it's something new I've found useful.

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -0,0 +1,663 @@
...
+    })
+  })
+
+  describe("delete a row", () => {
+    const deleteRow = (
+      tableId: string | undefined,
+      body: Row | { rows: Row[] }
+    ) => makeRequest("delete", `/api/${tableId}/rows`, body)
+
+    test("Given than a row exists, delete request removes it", async () => {
```
Rory-Powell, line 314: I find breaking up the `Given` statements into their own `describe` blocks can help with producing re-usable scenarios / highlight negative and error test cases. e.g.

```ts
describe("delete a row", () => {
    it("produces 404", () => {
    })

    describe("rows exists", async () => {
      
      let rows: Row[]
      
      beforeEach(async () => {
         rows = await populateRows(numberOfInitialRows)
      })
        
      it("deletes a row", () => {})
      it("deletes multiple rows", () => {})
    })
})
```

Rory-Powell (commented):
Awesome work on this, great to start getting this sort of coverage 👍 

A higher level question to this would be that given these are higher level API tests, should we be running this as a part of the `qa-core` integration suite instead?

packages/server/src/integration-test/postgres.spec.ts
adrinr: I think it's a good idea for API tests!

adrinr:
> Awesome work on this, great to start getting this sort of coverage 👍
> 
> A higher level question to this would be that given these are higher level API tests, should we be running this as a part of the `qa-core` integration suite instead?

@Rory-Powell  It currently takes around 22 seconds to run. If you think it's worth to move to qa, happy to move it in there

mike12345567 (approved):
Brilliant to have this tested - I think we should keep this in the same location for now - there has been a lot of on-going discussion around `qa-core` and how the server/worker tests factor into our integration suite, but I think that is a separate issue to this test case as the vast majority of our tests are integration tests, so any sort of move should be a separately spec'd item.

This is a massive improvement to our testing so happy to finally have this, amazing work @adrinr !

PR #9370, opened 2023-01-18
fix couchdb image tag to 3.1.1

rooso:
## Description
couchdb image version was set to 3.2.1 in values.yml but should be 3.1.1 according to the bug described in https://github.com/apache/couchdb-helm/issues/33


rooso:

PR #9371, opened 2023-01-18
New Help Menu

Ghrehh:
Putting this up early for some feedback:

The designs use font awesome, we have a few options to achieve this:

- Use substitute icons from spectrum, I couldn't find icons for discord or github however, so this might not be viable. 
- Embed the require icons as SVGs.
- Include font awesome using a script tag, which I opted to do for now, seems like you can created scoped kits that include only the desired icons also.
- Use our package manager, which I'm a little apprehensive about. At my previous role we used this in a React app and it bloated the bundle size massively even with tree shaking. Unsure if that's still an issue.

Any opinions very welcome, I'll add some additional questions to the code too.

I also need to clean up some css to use css variables also.

Screenshot 2023-01-18 at 10 14 59


packages/bbui/src/Popover/Popover.svelte
```diff
@@ -16,6 +16,9 @@
   export let direction = "bottom"
   export let showTip = false
 
+  let clazz
+  export { clazz as class }
```
Ghrehh, line 20: This is to support passing down a class from a parent component, is there a better way to do this? Also I had to mark the passed down class as `global` to get this to work, is that correct?

packages/builder/src/components/common/HelpIcon.svelte
```diff
@@ -1,41 +1,163 @@
...
+    class="helpMenuPopoverOverride"
+    bind:show
+    bind:hide
+    anchor={popoverAnchor}
+  >
+    <nav>
+      <div class="header">
+        <Heading size="XS">Help resources</Heading>
+        <button on:click={hide} class="closeButton"
+          ><FontAwesomeIcon icon="fa-solid fa-xmark" /></button
```
Ghrehh, line 33: Odd formatting here by prettier 🤷

packages/client/src/components/ClientApp.svelte
```diff
@@ -96,6 +96,9 @@
       <script src={`${plugin.jsUrl}`}></script>
     {/each}
   {/if}
+  <script
```
Ghrehh, line 99: Don't think I actually need this tag, I don't believe this menu is visible on the client.

packages/builder/src/components/common/HelpIcon.svelte
aptkingston: Yea prettier sometimes formats like this in the first instance, but if you fix it (or structure it this way yourself in the first instance) then it won't touch it. I can't count the amount of instances of these I have fixed... They annoy me so I always ensure it's structured properly. I would encourage you to do the same in general going forward - it does help readability.

packages/client/src/components/ClientApp.svelte
aptkingston: Yep this can be removed - it's builder only.

packages/bbui/src/Popover/Popover.svelte
aptkingston: Is this definitely needed? I can see that the reason you're doing this is to apply custom styles to the popover to effectively hardcode it to use a dark background, but can we not just let it respect themes as everything else does?

aptkingston (commented):
Nice work! Left a few comments.

About pulling in font awesome - I think my preferred way of using it would be:
- The script tag for pulling in FA works well since we can choose what icons we include, but it's hard to maintain and probably requires a login or something to edit the icon list? I've detailed what I think is the best way to include it below.
- In terms of file structure, I think I'd prefer to not put it in BBUI since we don't pull in any icons in BBUI, so standalone it wouldn't work. I think we should move the `FontAwesomeIcon` to the builder common components directory
- We should move the font awesome import into the `FontAwesomeIcon` file - makes sense to keep them together

In terms of adding font awesome - the way I've done it before which only includes the icons you want and is also easy to maintain is basically installing the SVG version via NPM in a tree-shakeable way. The gist of that is:
- Install `@fortawesome/fontawesome-svg-core` and `@fortawesome/free-solid-svg-icons`
- Import the icons you need from `@fortawesome/free-solid-svg-icons`
- Add them to the SVG library
- You can now use that icon

The way I see that in practice (following on from the file structure stuff above) is a font awesome icon component inside the builder rather than BBUI. This file would basically have:
```html
<script context="module">
  import { dom, library } from "@fortawesome/fontawesome-svg-core"
  import { faPlus } from "@fortawesome/free-solid-svg-icons"

  library.add(faPlus)
  dom.watch()
</script>

<script>
  export let name
</script>

<i class="fa-solid {name}" />
```

That example above will only work for the `fa-plus` icon. If you wanted to add `fa-coffee`, you would simply add it as an additional import, then add `library.add(faCoffee)`, then you would be able to use `<FontAwesomeIcon name="fa-coffee" />`.

This implementation is fully tree shakable and the only stuff bundled it the SVG JS code itself (60KB, 16KB gzipped) and whatever icons you need which is basically 0. It's also very easy to maintain over time.

I'd say if/when we start using FA in BBUI components, then we can move that into BBUI, but until then we keep it inside the builder to avoid bundling it inside the client lib etc.

Edit: updated the example code to use `context="module"` and this ensures the library and dom watcher are only initialised once.

Ghrehh:
Thanks for the detailed response @aptkingston, that way makes sense and I'll give implementing it a go.

aptkingston (commented):
Perfect, glad that idea worked well :+1:  Pulled this down and had a play - think there's just a couple of changes we need:
- Remove the premium lock beside email support when you have a paid plan
- Don't show the help menu when you aren't logged in - it should only be authenticated users
- Don't show the help menu when inside the builder - just the portal
- Don't show the help menu when inside the regular user app list page - just inside the portal for devs and admins

AndyWhann:
@Ghrehh @aptkingston could Just one small adjustment, could we get a set width on the icons so all the labels are left aligned?

It was my bad, I didn't have it like this in design and only noticing now.

Here's what I mean:
image

aptkingston (approved):
The lock diff is huge - usually happens when running an `install` directly inside the builder package. Normally better to just add the new deps to `package.json` directly, then run a top level bootstrap and let it sort out the deps.

Given the time scenario I'm happy to merge this in this case though. Looks like mostly minor package bumps so hopefully it won't cause any issues.

PR #9373, opened 2023-01-18
Fix Table Creation With No Import

Ghrehh:
Oversight by me on the previous JSON import PR, the validation on the button to create the table basically always assumed you were importing data; this PR fixes that.


melohagan (approved):
LGTM! Nice fix!

aptkingston (approved):

PR #9400, opened 2023-01-20
Create edit relationship modal

melohagan:
## Description
Needed to swap a couple of the to/from columns I had the wrong way around.

Also made a change to prevent the same column being selected on a through table.

## Screenshots
tabs_bug


mike12345567 (approved):
LGTM!

Issue #9225, opened 2023-01-04
Onboarding: Builder Tour


shogunpurple:
When I land into the builder for the first time on my first app creation, I want to see a tour of the builder so that I can get more familiar with the main areas of budibase.

**Designs**
https://www.figma.com/proto/AzrkyCDKmhQ9BQHqHdD9Yx/Onboarding?node-id=33%3A1139&scaling=scale-down&page-id=0%3A1&starting-point-node-id=1%3A457

**COS**
The builder tour is forced per user - we will store some info against them in the database, and any user who opens the builder and has not gone through the tour will see it until they complete it. This includes existing users.
- Show data, design, automate etc and navigate user to each one with a first time tour.

**Events**
All events for builder tour available here.
- https://budibase.atlassian.net/wiki/spaces/PROD/pages/164429833/Onboarding+Revamp+Q1+2023

**Notes**
This is a pretty good tool for these types of experiences: https://shepherdjs.dev/

PR #9409, opened 2023-01-23
Feature/user onboarding overlays

deanhannigan:
## Description
Added the Onboarding workflow

- A new `onboardedAt` flag has been added to the user. Not having this set will trigger the onboarding tour to load.
- The tour `builder-onboarding` has been defined with 4 steps. Each step directs users through the core features of the Builder.
- Steps can be defined with:
  - `route` - if specified, when the next button is clicked, the builder will navigate to this path. An example route would be the design page `/builder/app/:application/design`
  - `title` - The header for the currently displayed step
  - `layout` - An optional component reference. More complex
  - `body` - If a layout is not particularly complex, you can simply define text to appear in tour popover
  - `query` - The explicit css query to find the element to anchor tour popovers.
  - `onLoad` - An action to be carried out when the step loads. In this instance, firing off onboarding events.
  - `onComplete` - An action that currently concludes onboarding but can contain any custom terminating action
- *All* new and existing users will be prompted with this Tour once.
- Once complete, the user `onboardedAt` is updated and this Tour will not be seen again.
- It should be possible to define further tours/steps relating to other areas of the product.

Addresses: 
- https://github.com/Budibase/budibase/issues/9225

## Screenshots

Step 1 - Onboarding - Data

Screenshot 2023-01-22 at 23 06 02

Step 2 - Onboarding - Design

Screenshot 2023-01-22 at 23 06 28

Step 3 - Onboarding - Automations

Screenshot 2023-01-22 at 23 06 38

Step 4 - Onboarding - Publish

The defined workflow will terminate here

Screenshot 2023-01-22 at 23 06 47


aptkingston (commented):
Nice work Dean! The functionality here is great - resuming the tour, stepping forwards and back, generic implementation to be reusable - loads of stuff. Really nice work.

On the design side there are a few things to update, but it should actually be a simplification. The designs don't have the tip above the popover at all, so we can actually fully revert `position_dropdown.js`. We should left align the popovers for data, design and automate, and right align the popover for publish. To do this in a generic way, could we include an `align` property inside the tour step definition then just pass that into the root `Popover` in `TourPopover`? That way we can dynamically align left or right based on what we want, and not worry about the tip at all.

deanhannigan:
Hi @aptkingston thanks for taking a look! Regarding the tooltip, that's no problem at all. I thought it would help direct the eye to the target but as you say, it does add a bit more complexity. I'll implement the changes as you've laid out and get the PR updated.

A couple things I would keep in `position_dropdown.js` would be the accommodation for the popovers 20px margin on the left. This change should only affect popovers that would extend off-screen. Secondly, the observer on the document body stops the popovers from being marooned if the browser window is resized. Would you be happy to keep those two elements?

aptkingston:
> 

I had actually checked this out locally and was playing with the code around that stuff - the 20px margin left that appears when you have tooltip shouldn't even be there in the first place (this is the pre-existing style in the popover I mean). It only accounts for right alignment and the tooltip postition isn't dynamic which I know is what some of the changes you made were to address. I tinkered with it and while I had it working nice and generically, it ultimately isnt needed as we only have 1 usage of a tooltip in the entire codebase (the deployment modal). So I just left it.

If you want to then you can keep the adjustment to the position to account for the margin left, but it needs to only be applied when the alignment is right as well. My preference would proabably be delete the tip logic entirely and just remove the ability to even have the tip, to standardise popovers everywhere. If we ever need it then we can add it in properly with support for both left and right alignment.

As an aside, you can actually delete the "right-side" alignment as well - it's the same as "right" now. You'll just need to change the `BindingPanel` (which is the only place that uses "right-side") to "right".

aptkingston (approved):
LGTM!

Issue #6134, opened 2022-05-31
global-info > tenants is prone to 409 conflicts 


Rory-Powell:
**Hosting**
- cloud

**Describe the bug**
When concurrent accounts are created or deleted there is the possibility for the `tenants` document to receive conflicts. 

**To Reproduce**
- Create two accounts at the same time
- Delete two accounts at the same time
- Create and delete an account at the same time 

**Expected behavior**
- No errors

**Additional context**
The suggested fix here would be to use a document per tenant, and introduce a view for querying across all tenants similar to how users are queried by email. This introduces a nice place to store tenant wide information in future, e.g. the globally unique tenant id which is currently stored in the settings document

mike12345567:
This issue cropped up during the conversion of the worker tests, as all creation of users happening nearly simultaneously - for some reason this wasn't an issue with PouchDB but is with real CouchDB (likely timing related). For now we have added a `env.isTest` wrap to the tenancy functionality as it doesn't affect testing until this issue can be rectified overall.

PR #9419, opened 2023-01-23
Chore - use testcontainers

adrinr:
## Description
This PR provides the basis of the test image - it does this by breaking out the CouchDB component of the single image into its own image (https://hub.docker.com/r/budibase/couchdb) and then utilised this for creating the test image (https://hub.docker.com/r/budibase/tests).

I've also started to integrate the Jest Docker pre-runner (https://github.com/Trendyol/jest-testcontainers) package, so that we can make sure the Docker image has started before we run the test suite.

Will flesh out with further details - this is just to show how this will work - working on converting tests currently.

Addresses: 
- #8322


packages/server/.env.test
```diff
@@ -0,0 +1,7 @@
+JWT_SECRET=testsecret
```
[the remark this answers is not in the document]
adrinr: File removed

adrinr:
recheck

adrinr:
recheck

packages/server/src/tests/jestSetup.ts
```diff
@@ -15,3 +17,24 @@ if (!process.env.CI) {
   // 100 seconds
   jest.setTimeout(100000)
 }
+
+function overrideConfigValue(key: string, value: string) {
```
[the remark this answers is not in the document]
adrinr: This is required to have it in here are is setting up the variables related to testcontainers. Sadly they are not available until after the load

packages/server/jest.config.ts
```diff
@@ -19,21 +18,4 @@ const config: Config.InitialOptions = {
   },
 }
 
-if (!process.env.CI) {
```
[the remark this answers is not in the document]
adrinr: Added back

adrinr:
This PR requires https://github.com/Budibase/budibase-pro/pull/104 to be merge

packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -83,7 +83,14 @@ export class DatabaseImpl implements Database {
       throw new Error("DB does not exist")
     }
     if (!exists) {
-      await this.nano().db.create(this.name)
+      try {
+        await this.nano().db.create(this.name)
+      } catch (err: any) {
+        // Handling race conditions
+        if (err.statusCode !== 412) {
```
adrinr, line 90: @Rory-Powell @mike12345567 thoughts on this?

packages/backend-core/src/auth/auth.ts
```diff
@@ -9,16 +9,8 @@ import {
...
-  auditLog,
   tenancy,
-  authError,
-  ssoCallbackUrl,
   csrf,
-  internalApi,
-  adminOnly,
-  builderOnly,
-  builderOrAdmin,
-  joiValidator,
```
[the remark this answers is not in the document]
adrinr: Do you mean this should be added back?

packages/backend-core/src/cache/user.ts
```diff
@@ -61,7 +62,12 @@ export async function getUser(
...
-export async function invalidateUser(userId: string) {
+export async function invalidateUser(userId: string, tenantId?: string) {
+  tenantId = tenantId || getTenantId()
+  const cacheKey = getCacheKey(tenantId, userId)
   const client = await redis.getUserClient()
-  await client.delete(userId)
+  await client.delete(cacheKey)
 }
+
+const getCacheKey = (tenantId: string, userId: string) =>
```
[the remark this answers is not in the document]
adrinr: Yes, it was due conflicting ids between tenants. But it can probably be undone as now everything should be using randoms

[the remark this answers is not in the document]
adrinr: Undone

packages/server/src/tests/jestSetup.ts
```diff
@@ -1,9 +1,6 @@
-import { mocks } from "@budibase/backend-core/tests"
-
-// mock all dates to 2020-01-01T00:00:00.000Z
-// use tk.reset() to use real dates in individual tests
-const tk = require("timekeeper")
-tk.freeze(mocks.date.MOCK_DATE)
```
Rory-Powell, line 6: Can we re-add the date freezing here? I find frozen dates by default are useful, especially around making tests and data more predictable. It's also consistent with how it's done in the other packages. 

We can still turn them off when required by using `tk.reset()` in a particular test

packages/server/src/db/tests/linkController.spec.js
```diff
@@ -10,6 +10,7 @@ describe("test the link controller", () => {
   let table1, table2, appId
 
   beforeEach(async () => {
+    config.modeCloud()
```
Rory-Powell, line 13: Can't see any obvious requirement for cloud hosting here, or was the intention something else around multi tenancy?

packages/server/src/tests/jestEnv.ts
```diff
@@ -3,11 +3,7 @@ import { tmpdir } from "os"
 
 env._set("SELF_HOSTED", "1")
```
Rory-Powell, line 4: Would be nice if we made this file consistent with:

https://github.com/Budibase/budibase/blob/develop/packages/worker/src/tests/jestEnv.ts

And used `process.env`

packages/server/src/tests/jestEnv.ts
```diff
@@ -3,11 +3,7 @@ import { tmpdir } from "os"
...
 env._set("NODE_ENV", "jest")
-env._set("JWT_SECRET", "test-jwtsecret")
-env._set("CLIENT_ID", "test-client-id")
+env._set("MULTI_TENANCY", "1")
```
Rory-Powell, line 6: 👌

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -20,22 +20,45 @@ import {
...
 import { CreateUserResponse, User, AuthToken } from "@budibase/types"
 import API from "./api"
+import sdk from "../sdk"
 
 enum Mode {
   CLOUD = "cloud",
   SELF = "self",
 }
 
+async function retry<T extends (...arg0: any[]) => any>(
```
Rory-Powell, line 36: Similar to the `retry` function in server? 
Could be moved out to `backend-core`

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -20,22 +20,45 @@ import {
...
+  }
+}
+
 class TestConfiguration {
   server: any
   request: any
   api: API
   defaultUser?: User
   tenant1User?: User
+  #tenantId?: string
```
Rory-Powell, line 61: What's up with the `#` notation? It's not something I've come across before

packages/worker/src/api/routes/global/tests/auth.spec.ts
```diff
@@ -40,36 +40,43 @@ describe("/api/global/auth", () => {
 
     describe("POST /api/global/auth/:tenantId/reset", () => {
       it("should generate password reset email", async () => {
-        const { res, code } = await config.api.auth.requestPasswordReset(
-          sendMailMock
-        )
-        const user = await config.getUser("test@test.com")
+        await tenancy.doInTenant(config.tenant1User!.tenantId, async () => {
+          const userEmail = `${utils.newid()}@test.com`
```
Rory-Powell, line 44: Can use `structures.email()` which will also use a new uuid in the first half of the address

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -20,22 +20,45 @@ import {
   auth,
   constants,
   env as coreEnv,
+  utils,
+  DEFAULT_TENANT_ID,
```
Rory-Powell, line 24: It seems no longer safe to use `DEFAULT_TENANT_ID` and `TENANT_ID` constants from below, they have the value of `default` and will be inherently racey. 

Do you think it would be possible to use an autogenerated tenant id instead like we are using for the `#tenantId` here? That might help with the `retry` logic required in this file.

packages/worker/src/api/routes/global/tests/email.spec.ts
```diff
@@ -6,7 +6,7 @@ import { EmailTemplatePurpose } from "../../../../constants"
 describe("/api/global/email", () => {
   const config = new TestConfiguration()
 
-  beforeAll(async () => {
+  beforeEach(async () => {
     await config.beforeAll()
```
Rory-Powell, line 10: Is this intentional , `beforeEach` calling `beforeAll`? The beforeAll is generally a test level setup, that does any heavier lifting.

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -129,10 +160,19 @@ class TestConfiguration {
...
+  createTenant = async ({
+    addTenantToGlobalDb,
+  }: {
+    addTenantToGlobalDb: boolean
+  }): Promise<User> => {
     // create user / new tenant
     const res = await this.api.users.createAdminUser()
 
+    // This needs to be added because it was disabled for bulk testing: // https://github.com/Budibase/budibase/issues/6134
+    if (addTenantToGlobalDb) {
```
Rory-Powell, line 172: The `globalDb` part here is a little confusing, is this the `global-info` db or the `tenant_global-db`? 

We really should have called the `global-info` db something else, such as `static-db` or `platform-db` to signify there is only one.

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -129,10 +160,19 @@ class TestConfiguration {
 
   // TENANCY
 
-  createTenant = async (): Promise<User> => {
+  createTenant = async ({
+    addTenantToGlobalDb,
+  }: {
+    addTenantToGlobalDb: boolean
```
Rory-Powell, line 166: Default to `true`? It should be a fairly niche case where you don't want to do this

Rory-Powell (commented):
GREAT work Adri, amazing you were able to work through all of this. There's a lot of concepts here you're unpacked and understood really quickly - nice job!

There are a lot of comments here, but they are mostly very small points and hints towards using some helpers we have in place in a lot of cases with test data.

packages/backend-core/src/cache/tests/writethrough.spec.js
```diff
@@ -1,15 +1,17 @@
-require("../../../tests")
+const{generator}=require("../../../tests")
```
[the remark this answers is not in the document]
adrinr: Is there a reason why linting is ignoring the js files?

hosting/dependencies/runner.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+redis-server --requirepass $REDIS_PASSWORD > /dev/stdout 2>&1 &
+/bbcouch-runner.sh &
+/minio/minio server ${DATA_DIR}/minio --console-address ":9001" > /dev/stdout 2>&1 &
```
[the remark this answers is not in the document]
adrinr: It should not be a problem. This will happen in the container itself, docker will only expose it when required. And testcontainers will provide a random port

hosting/docker-compose.dev.yaml
```diff
@@ -42,25 +42,30 @@ services:
...
     environment:
       PUT_CALL: "curl -u ${COUCH_DB_USER}:${COUCH_DB_PASSWORD} -X PUT couchdb-service:5984"
     depends_on:
       - couchdb-service
-    command: ["sh","-c","sleep 10 && $${PUT_CALL}/_users && $${PUT_CALL}/_replicator; fg;"]
+    command:
+      [
+        "sh",
+        "-c",
+        "sleep 10 && $${PUT_CALL}/_users && $${PUT_CALL}/_replicator; fg;",
```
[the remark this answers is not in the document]
adrinr: You are right. @mike12345567 already used that in our image, not need to run this :)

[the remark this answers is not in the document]
adrinr: We should probably use the same couch image for https://github.com/Budibase/budibase/blob/develop/hosting/docker-compose.yaml#L99
Not sure if the volume will be persisted properly or if there is a risk of losing data. Thoughts, @mike12345567 ?

packages/worker/src/tests/TestConfiguration.ts
adrinr: I think this is required as we are using "default" by default in the code itself. If we put something random, all the pieces of code that run `if tenant == 'default'` would fail. And this is still needed for tests such as migrations, plugins... global stuff like that

packages/worker/src/api/routes/global/tests/email.spec.ts
adrinr: It does not look intentional here, as there is only one test in the file 😅 In some other files it was required, as the `.beforeAll` is creating a new tenant, and there were conflicts with tests that required a clean instance

packages/worker/src/tests/TestConfiguration.ts
adrinr: I needed to add this because of this issue: https://github.com/Budibase/budibase/issues/6134
When running [these tests](https://github.com/Budibase/budibase/blob/68d147b0f75e0746b5c7347bb81ba87063a1cd6a/packages/worker/src/middleware/tests/tenancy.spec.ts) in parallel with the rest, we kept getting conflicts.
I think this might not be an issue anymore as we handle the 412 in the db level. I will double check it

packages/backend-core/src/cache/tests/writethrough.spec.js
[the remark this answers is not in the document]
Rory-Powell: Aye you're right it's ignored in `.prettierignore`. No good reason, best to remove that line and re-lint!

packages/worker/src/tests/TestConfiguration.ts
adrinr: As mentioned [here](https://github.com/Budibase/budibase/pull/9419#discussion_r1095598847) this is not needed anymore! 🎉

packages/worker/src/tests/TestConfiguration.ts
adrinr: Yup, not needed anymore. This might have solve the issue as well

packages/worker/src/tests/TestConfiguration.ts
Rory-Powell: In general we only check for `default` when `MULTI_TENANT` is false, and use that as a shortcut instead of accessing the tenant id in context. See: https://github.com/Budibase/budibase/blob/develop/packages/backend-core/src/context/mainContext.ts#L172-L174

As the worker has multi tenant set to true it should be safe, and it should save a lot of trouble with conflicting tests going forward. Each tenant would have it's own migrations and plugins in their respective global-db then.

hosting/docker-compose.dev.yaml
```diff
@@ -42,25 +42,30 @@ services:
...
+    image: budibase/couchdb
     environment:
       - COUCHDB_PASSWORD=${COUCH_DB_PASSWORD}
       - COUCHDB_USER=${COUCH_DB_USER}
     ports:
       - "${COUCH_DB_PORT}:5984"
     volumes:
-      - couchdb3_data:/opt/couchdb/data
-  
+      - couchdb_data:/data
```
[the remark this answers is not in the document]
mike12345567: This will not be possible sadly.

packages/backend-core/src/db/couch/DatabaseImpl.ts
adrinr: It was happening in multiple places, I actually needed retries because of this mainly. But the biggest conflicting one was `global-db`

hosting/docker-compose.dev.yaml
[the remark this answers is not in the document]
mike12345567: There will be data loss sadly if we switch image - this is something that actually needs discussed with the team.

packages/backend-core/src/cache/tests/writethrough.spec.js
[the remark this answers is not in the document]
adrinr: I will do this in another PR, I will try to un-ignore it and see how many file changes we have

hosting/docker-compose.dev.yaml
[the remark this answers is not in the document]
adrinr: To be tackled separately

packages/worker/src/tests/TestConfiguration.ts
adrinr: Some tests required multi_tenant to be false, as they were part of the test. I did not want to change the logic of these tests. Maybe worth to note it down and tackling it separately, as this was already like this, https://github.com/Budibase/budibase/blob/develop/packages/worker/src/tests/TestConfiguration.ts#L155

Rory-Powell: I think some tests require there to be `SELF_HOST`, which should be a fine configuration we can allow for a self hosted multi tenant test. 

Do you know which ones require single tenancy? It shouldn't be the case.

Given that worker runs tests as `MULTI_TENANCY=1`, using the constant `default` as the tenant id in all tests undoes a lot of the benefit. Changing it to a random should help us a lot with conflicting test state.

packages/worker/src/api/routes/global/tests/email.spec.ts
Rory-Powell: I think we should try to be more clear on the intent in this case, as it can set a misleading pattern to new devs that we must use the `beforeAll` helpers in `beforeEach` blocks. 

An alternative might be to add a helper to config, `reset`, and we can use as
```ts
beforeEach(async () => {
   await config.reset()
})
```

This can under the hood just call `beforeAll`, it's just a bit more clear on the reason why.

packages/worker/src/tests/TestConfiguration.ts
adrinr: We have the helper `modeCloud` that was used in multiple places
 https://github.com/Budibase/budibase/blob/feature/dependencies-image/packages/worker/src/tests/TestConfiguration.ts#L101

packages/backend-core/tests/utilities/structures/tenants.ts
```diff
@@ -0,0 +1,5 @@
+import { newid } from "../../../src/newid"
+
+export function name() {
```
Rory-Powell, line 3:
```suggestion
export function id() {
```
Just as we already use `tenantId` everywhere, there is no mention of name outside of the account portal

adrinr: True 😅

packages/backend-core/src/middleware/passport/tests/third-party-common.seq.spec.js
```diff
@@ -6,6 +6,8 @@ const { DEFAULT_TENANT_ID } = require("../../../constants")
 const { generateGlobalUserID } = require("../../../db/utils")
 const { newid } = require("../../../utils")
 const { doWithGlobalDB, doInTenant } = require("../../../tenancy")
+const { default: environment } = require("../../../environment")
+environment._set("MULTI_TENANCY", 'TRUE')
```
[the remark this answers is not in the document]
adrinr: Not required anymore, just deleted. Probably we could have all the tests running in multitenant, but probably we will need to refactor something

packages/backend-core/src/middleware/passport/tests/third-party-common.seq.spec.js
```diff
@@ -6,6 +6,8 @@ const { DEFAULT_TENANT_ID } = require("../../../constants")
 const { generateGlobalUserID } = require("../../../db/utils")
 const { newid } = require("../../../utils")
 const { doWithGlobalDB, doInTenant } = require("../../../tenancy")
+const { default: environment } = require("../../../environment")
```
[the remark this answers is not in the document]
adrinr: For some reason does not work :/

Rory-Powell (approved):

Issue #9401, opened 2023-01-20
MySQL Not unique table/alias: Multiple many-to-many relationships between the same from and to tables


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.2.16
  - App Version: 2.2.16

**Describe the bug**
When attempting to setup multiple many-to-many relationships between the same from and to tables, the SQL statement fails because the alias columns use the table names instead of the relationship column names.

**To Reproduce**
1. Define a Many to Many relationship for Many Employees -> Many Departments (through dept_emp)
2. Define a Many to Many relationship for Many Employees -> Many Departments (through dept_manager)
3. View Departments table, view error: `Not unique alias 'employees'`

**Expected behavior**
You should be able to setup multiple many-to-many relationships between two tables given you have multiple joining tables.
E..g.
Many Employees -> Department Employees -> Many Departments
Many Employees -> Department Managers -> Many Departments

The SQL query should use the column name and not the table name in the column alias. 

**Screenshots**
![Screenshot 2023-01-20 at 15 40 36](https://user-images.githubusercontent.com/101575380/213741338-cc62c7f9-f4c0-4ff8-b0af-0230ea4aeaba.png)

![Screenshot 2023-01-20 at 15 41 32](https://user-images.githubusercontent.com/101575380/213741529-c2ed197b-591b-4811-adef-83522b5cf49e.png)

![Screenshot 2023-01-20 at 15 41 06](https://user-images.githubusercontent.com/101575380/213741437-f12ce5f0-6961-4576-9563-1faf2cd316fe.png)

**Additional Context**
Using this dataset: https://github.com/datacharmer/test_db

melohagan:
For further context, the joins of the SQL have to change from:

```sql
 left join `dept_emp` on `employees`.`emp_no` = `dept_emp`.`emp_no`
 left join `departments` on `departments`.`dept_no` = `dept_emp`.`dept_no` 
 left join `dept_manager` on `employees`.`emp_no` = `dept_manager`.`emp_no` 
left join `departments` on `departments`.`dept_no` = `dept_manager`.`dept_no` order by `employees`.`emp_no` asc limit 5000;
```

to this:

```sql
left join `dept_emp` on `employees`.`emp_no` = `dept_emp`.`emp_no`
 left join `departments` on `departments`.`dept_no` = `dept_emp`.`dept_no` 
[...]

PR #9421, opened 2023-01-23
Prevent many-many relationships with same tables

melohagan:
## Description
I struggled to stop One->Many relationships being duplicated between tables because the linking tableId is not included in the schema. Using the keys isn't necessarily reliable.

The bug scenario however is fixed, i.e. preventing two Many->Many relationships between two tables that already have a Many-Many relationship.  

Addresses: 
- https://github.com/Budibase/budibase/issues/9401

## Screenshots
*Existing Many-Many relationship with dept_emp successfully created*
Screenshot 2023-01-23 at 23 10 43

*Cannot create a duplicate Many-Many relationship using the dept_managers table*
Screenshot 2023-01-23 at 23 12 24


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -271,6 +277,39 @@
...
+    }
+    let match = false
+    let fromThroughLinks = Object.values(
+      datasource.entities[fromTable.name].schema
+    ).filter(value => value.through)
+    let toThroughLinks = Object.values(
+      datasource.entities[toTable.name].schema
+    ).filter(value => value.through)
+
+    fromThroughLinks.forEach(fromVal => {
```
mike12345567, line 297: Logically this is quite difficult to follow, wondering if something like:
```js
const compareAgainstUserInput = (fromTableId, toTableId) => (fromTableId === fromId && toTableId === toId) || (fromTableId === toId && toTableId === fromId)
fromThroughLinks.find(from => 
  toThroughLinks.find(to => from.through === to.through && compareAgainstUserInput(from.tableId, to.tableId)))
```

Might need some formatting, but it breaks it up a little better and remove the double looping which isn't technically exited fully (the inner `toThroughLinks` is stopped, but the external `fromThroughLinks` will keep iterating currently which could cause issues down the line if this function was ever updated).

melohagan: That does look a bit cleaner - thanks!

mike12345567 (approved):

Issue #8989, opened 2022-12-09
Incorrect looping array binding will cause test to crash for external SQL database query rows


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.42
  - App Version: 2.1.42

**Describe the bug**
When you enter an incorrect binding value for the looping array, such as `{{ steps.1.rows.0.id }}` then the automation test hangs and a page refresh is required.
Whilst the binding is wrong, it should not crash the page. Does not crash for the internal DB query rows. 
Related discussion: https://github.com/Budibase/budibase/discussions/8974

**To Reproduce**
1. Add a Postgres datasource and fetch a table
2. Add an automation with the trigger *Row Created*
3. Add a **Query Rows** step. Then add an **External Data Connector** step.
4. Add *looping* to the EDC, and for Input type select Array. Use the incorrect binding `{{ steps.1.rows.0.id }}`
5. Finish and test. See error: `TypeError: Cannot read properties of undefined (reading 'filter')` and page hangs.
6. Refresh page and enter correct binding: `{{ steps.1.rows }}`. Works fine.

**Expected behavior**
Even if the binding value is incorrect for a looping array, the app should not crash.

**Screenshots**
![Screenshot 2022-12-09 at 11 21 04](https://user-images.githubusercontent.com/101575380/206691407-6ec14223-1428-4625-9ef0-abb848a8cd23.png)

*Stuck spinning*
![Screenshot 2022-12-09 at 11 21 34](https://user-images.githubusercontent.com/101575380/206691486-ec2a5de3-f686-4365-b5aa-8c24aee1ba1a.png)

PR #9434, opened 2023-01-24
Some extra checks for automation loop binding type safety

mike12345567:
## Description
Quick updates for #8989 - this issue appeared to be fixed already but just adding a few extra bits of security to make sure that looping only occurs when valid data is found in the binding.


Rory-Powell (commented):
Would be great to see a unit test proving out this behaviour. This should also help reviewers with understanding why this changed was required / the input condition that triggers it. 

The looping block in the `execute` function is also fairly huge and hard to read. By adding a test here it should help us naturally break up the logic a bit into more readable / testable chunks.

Issue #9206, opened 2023-01-03
In-Builder Env Variables Bindings (UI)


shogunpurple:
When I'm configuring a datasource or using the REST connector, I want to be able to bind and create new environment variables to my configuration.

**COS**
- On paid, dropdowns in the datasource configuration UI for each field, to allow you to use the global secrets and environment variables. Includes REST datasource bindings. When the value is injected, it will say `{{ env.secretname }}`. We should consider not showing the binding in the datasources text field as this would indicate you can use any bindings here.

- Add "Environment" binding to the bindings drawer where you have access to all of your environment variables

- Option is also available to "create a new environment variable" in which you will be presented with a modal and asked to create the prod/dev values for your environment variable, and on save, the variable binding will be injected into your configuration. This will use the same modal as we use in the variables configuration UI in the portal.

- Free users will see a dropdown on the environment variables with a single option to upgrade their plan. This will take them to the upgrade page.
With the lock beside it. On clicking this lock, you are taken to the upgrade page.
In the top level portal we also show a lock beside the “Environment Variables” tab to show that it’s a premium feature.

- Enable support for environment variables in automations

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/0a2c6988-be43-4f16-8eca-862fad8f7f8c)

Issue #8317, opened 2022-10-18
Portal Variables Management Panel (UI)


shogunpurple:
When I'm building budibase apps in different environments, I want to be able to use both staging/production credentials, so that I don't have to swap them based on dev/prod or expose credentials in the UI.

**COS**
- Only admins and builders can view the environment variables management UI
- UI to edit, add and control secrets
- Environment Variables are always encrypted and the value cannot be retrieved, only updated or deleted.
- Free users cannot see the management UI at all 
- Business plans or above will be able to edit and add environment variables
- Provide a warning when you export an app that says the following:
`Do not share your budibase application exports publicly as they may contain sensitive information such as database credentials or secret keys.`
- Includes the screen that users will see when they are free. **Need to include copy to actually describe what the feature is**

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/b374b41c-539c-4651-ade1-36f72d502e06)
![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/a5ecad58-7819-4b6a-b8b6-315ba6a1bb28)

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/f0ca4623-6407-4fbd-9d39-1be68b0cf79c)

PR #9453, opened 2023-01-26
Environment Variables

PClmnt:
## Description
This PR adds the UI, and some backend work, for the environment variables feature. 

This Feature will allow Business tier users to configure Development and Production variables, that can then be used in the following parts of the product:

- Datasource configuration
- REST query configuration
- Automations

Addresses: 
#9206 
#8317 

All PRs:
https://github.com/Budibase/account-portal/pull/243
https://github.com/Budibase/budibase/pull/9453
https://github.com/Budibase/budibase-pro/pull/103

### Portal
image
image

Environment variables can be added in this screen within the settings section portal. You can create a new variable by clicking on the `Add variable` button. You are then presented with the options for `Production` and `Development`. Development can either be different to Production, or be mirrored by the use of the `Use production value` button.

Once a variable has been created, it's value(s) can no longer be viewed, however, you can edit these by clicking on the `Edit` button within the table and adding new values. 

### Datasource Config
image
image
image

The first place that the feature can be used is within the datasource configuration screen. Envrionment Variables can be added to all configuration options, and will appear as a handlebars within the input (which is then disabled). You can still free type within the input, if there are no environment variables added. 

Clicking the `Add variable` button within this screen will also open the variable creation modal, and you can create environment variables directly from this screen, and they will be automatically inserted into the input box. 

Also on this screen are states for when
- the user has no environment variables
- the user has no license for envrionment variables

You can also add Environment Variables within the REST config screen for headers, as well as authentication.

### REST Config and Automations

image

image

In addition to this, environment variables will be aavailable in the REST Query viewers as well as automations. Once again, these are gated to only show when a license is found. 

Video: https://www.loom.com/share/8ba3fb9fe62f4daa997820ef02c5d005 (automation run failed at the end because i was running in development)

## Documentation


aptkingston:
Looks like something is wrong with the branch given all these conflicts - did you rebase or something to try and remove the commits from the new portal UI branch? If it's not easily fixable at this stage you might be better branching freshly off develop and just cherrypicking in your own commits.

Normally the squashing would just cause loads of commits numbers but the diff would be clean, but in this case the diff shows loads of the new portal UI changes too.

mbadan:

mike12345567:
recheck

packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -362,6 +363,13 @@
       notifications.error("Error getting datasources")
     }
 
+    try {
+      // load the environment variables
+      await environment.loadVariables()
+    } catch (error) {
+      notifications.error("Error getting environment variables")
```
mike12345567, line 370: Can we present a bit more of a useful error here? This should never happen, would be good to include some of the error message e.g. `Error getting environment variables - ${error}`.

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/rest/auth/RestAuthenticationModal.svelte
```diff
@@ -28,7 +41,16 @@
   let hasErrors = false
   let hasChanged = false
 
-  onMount(() => {
+  let createVariableModal
+  let formFieldkey
+
+  onMount(async () => {
+    await environment.loadVariables()
```
mike12345567, line 48: The other `loadVariables` calls have a try catch wrapper, if this or the licensing init were to fail it would likely cause issues on the page load.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -196,6 +216,10 @@
     onChange({ detail: tempFilters }, defKey)
     drawer.hide()
   }
+
+  onMount(async () => {
+    await environment.loadVariables()
```
mike12345567, line 221: Like other comment, the other calls to `environment.loadVariables` have an error try catch, might be worth adding here.

packages/builder/src/components/backend/DatasourceNavigator/TableIntegrationMenu/IntegrationConfigForm.svelte
```diff
@@ -70,6 +78,30 @@
...
+    createVariableModal.show()
+  }
+
+  async function handleUpgradePanel() {
+    await environment.upgradePanelOpened()
+    $licensing.goToUpgradePage()
+  }
+
+  onMount(async () => {
+    await environment.loadVariables()
```
mike12345567, line 99: Same as other comments, try catch for loading.

packages/builder/src/components/portal/environment/CreateEditVariableModal.svelte
```diff
@@ -0,0 +1,100 @@
...
+  const saveVariable = () => {
+    try {
+      save({
+        name,
+        production: productionValue,
+        development: developmentValue,
+      })
+      notifications.success("Environment variable saved")
+    } catch (err) {
+      notifications.error("Error saving environment variable")
```
mike12345567, line 44: Like other comments/the delete above, might be worth including the actual `err.message`

packages/builder/src/pages/builder/portal/settings/environment/_components/EditVariableColumn.svelte
```diff
@@ -0,0 +1,35 @@
...
+<ActionButton size="S" on:click={editVariableModal.show}>Edit</ActionButton>
+
+<Modal bind:this={editVariableModal}>
+  <CreateEditVariableModal {row} {save} />
+</Modal>
+
+<ConfirmDialog
+  bind:this={deleteDialog}
+  onOk={() => {
+    environment.deleteVariable(row.name)
```
mike12345567, line 27: Same as above, should there be an async element to wait for deletion.

packages/builder/src/pages/builder/portal/settings/environment/index.svelte
```diff
@@ -0,0 +1,140 @@
...
+
+  let modal
+
+  const customRenderers = [{ column: "edit", component: EditVariableColumn }]
+
+  $: noEncryptionKey = $environment.status?.encryptionKeyAvailable === false
+  $: schema = buildSchema(noEncryptionKey)
+
+  onMount(async () => {
+    await environment.checkStatus()
```
mike12345567, line 27: As other comments, one had a try catch, this doesn't.

packages/builder/src/pages/builder/portal/settings/environment/index.svelte
```diff
@@ -0,0 +1,140 @@
...
+        width: "auto",
+        borderLeft: true,
+        displayName: "",
+      }
+    }
+    return schema
+  }
+
+  const save = data => {
+    environment.createVariable(data)
```
mike12345567, line 48: as with other comments, should this include async handling.

mike12345567 (commented):
I've reviewing the `frontend-core` and `builder` changes made, looks good, just a few comments that might need addressed.

packages/frontend-core/src/api/environmentVariables.js
```diff
@@ -0,0 +1,36 @@
...
+  },
+  deleteEnvironmentVariable: async varName => {
+    return await API.delete({
+      url: `/api/env/variables/${varName}`,
+    })
+  },
+
+  updateEnvironmentVariable: async data => {
+    return await API.patch({
+      url: `/api/env/variables/${data.name}`,
```
shogunpurple, line 32: Is this a `PATCH`? Do we send partial data?

packages/server/src/sdk/app/datasources/datasources.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export async function get(
+  datasourceId: string,
+  opts?: { enriched: boolean }
+): Promise<Datasource> {
+  const appDb = context.getAppDB()
+  const datasource = await appDb.get(datasourceId)
+  if (opts?.enriched) {
+    return (await enrichDatasourceWithValues(datasource)).datasource
```
shogunpurple, line 38: NAB: `await` not needed here

packages/types/src/sdk/datasources.ts
```diff
@@ -1,5 +1,7 @@
 import { Table } from "../documents"
 
+export const PASSWORD_REPLACEMENT = "--secret-value--"
```
shogunpurple, line 3: I think we should not show this in the UI when a user adds a secret value - it should be pretty hidden from the user.

shogunpurple (approved):
LGTM! Nothing blocking, just a few small comments

packages/frontend-core/src/api/environmentVariables.js
mike12345567: We're counting it as a patch because each environment variable is added to an overall document, there isn't a document for each single variable, therefore you don't get an ID etc for each one. It is simply added to a map within an existing document, so we decided the best way to denote that was as a patch.

packages/types/src/sdk/datasources.ts
mike12345567: Sure, do you think it would be better to just not return the property so that it always appears as an empty field in the UI?

Issue #9228, opened 2023-01-04
[Builder] Login and onboarding UI


shogunpurple:
When I accept an invite from my budibase admin, I want to be onboarded smoothly so that I can get started quickly with budibase. The various login pages should also be updated to be identical to the account portal ones. 

**Designs** - See "New Developer" and "App User"
https://www.figma.com/file/AzrkyCDKmhQ9BQHqHdD9Yx/Onboarding?node-id=0%3A1&t=25vxgAnmcp6B8Pg9-0

**COS - Developer User**
When a user is invited with developer/builder privileges.
- Fill in name (optional)
- Fill in password
- Take to portal with help menu shown
- When they create their first app or open an existing app, they are shown the first time builder tour

**COS - App User**
When a user is invited with app user privileges.
- Fill in name (optional)
- Fill in password
- Take to standard app user portal

**COS - Login**
- Update self hosted login page, forgot password, reset password to look the same as the account portal

PR #9467, opened 2023-01-27
Selfhost onboarding UX/UI updates

deanhannigan:
## Description

Updating existing onboarding  self host UI/UX to utilise the new form and layout components design.

Screens updated: 

- Login
- Request password reset
- Reset Password
- Accept User Invite
  - We now allow users to set their name immediately and upon completing their acceptance, they will now be taken directly into the application
- Create admin

Addresses: 
- https://github.com/Budibase/budibase/issues/9228

## Screenshots

Login with multiple SSO configured

Request Password reset

Reset password

Accept User Invite

Accept User Invite - acceptance to login flow

Create Admin - the screen you see when provisioning a new budibase install.

### DIFF
--- a/packages/builder/src/pages/builder/auth/login.svelte
+++ b/packages/builder/src/pages/builder/auth/login.svelte
@@ -5,7 +5,6 @@
     Button,
     Divider,
     Heading,
-    Input,
     Layout,
     notifications,
     Link,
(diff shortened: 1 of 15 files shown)


shogunpurple (approved):
LGTM!

packages/builder/src/pages/builder/auth/login.svelte
```diff
@@ -57,75 +64,96 @@
...
+        <FancyInput
+          label="Your work email"
+          value={formData.username}
+          on:change={e => {
+            formData = {
+              ...formData,
+              username: e.detail,
+            }
+          }}
+          validate={() => {
```
aptkingston, line 94: Same as before, all you need to do is just return error strings, then invoke form.validate() and use the result to know wether the form is valid. You don't need to keep track of any error variables or pass any `error` props into the fields unless they're custom errors (such as returned from an API or something).

This is my bad - I musn't have explained that clearly enough.

packages/builder/src/pages/builder/auth/reset.svelte
```diff
@@ -37,47 +49,92 @@
...
+          value={formData.password}
+          type="password"
+          on:change={e => {
+            formData = {
+              ...formData,
+              password: e.detail,
+            }
+          }}
+          validate={() => {
+            let fieldError = {}
```
aptkingston, line 79: Confirmation passwords are a bit of a different case so it's probably fine to use custom errors here, but again you probably don't need to.

aptkingston (approved):
The form validation could be done in a much neater way using the built in validation, but  that's my bad for obviously not explaining it properly! I can run you through that and we can refactor this to use it properly at a later stage, but I'm happy to get this merged now. No need to make it a blocker if this implementation also works.

I started writing comments for each usage but then saw that all the validation is done that same way so I stopped - but like I said we can jump on a call and refactor it all some other time.

Let's get this merged!

Issue #9375, opened 2023-01-18
Set value to Relationship from a multiple picker on Saving row


juangbm:
I found in a previous issue a workaround to filter information from a Relationship using an option picker (https://github.com/Budibase/budibase/discussions/4579), and I was able to adapt it to use a multiple picker for a many relationship. I can display the different options filtered from the data provider that I have added, but when I select them and try to save, I get an error "row[key].map is not a function". What should I use on Action Save Row to update this relationship with the information selected in the multiple picker?

Any idea?

Thanks!

_Originally posted by @juangbm in https://github.com/Budibase/budibase/discussions/4579#discussioncomment-4670395_

PR #9475, opened 2023-01-30
Handle encoded many relationships

melohagan:
## Description
**row[key].map is not a function** error was being caused by the fact that the relationship ID was sometimes an encoded string, and not always an array.
Now checking if the id needs to be decoded first. Previously it handled One->Many relationships, but I had missed the Many->Many cases.

Addresses: 
- https://github.com/Budibase/budibase/issues/9375

## App Export
*Let me know if you want access to it*


shogunpurple (commented):
It would be really good to have a wee unit test for this to cover.

Issue #7972, opened 2022-09-26
Background image Element has same comportment as image, and can not be set to background


KevinHamonApp:
**Hosting**
Same comportement in cloud solutions and hosting.

**Describe the bug**
When I use background image Element, I cannot put the image in background. The component cannot receive any childrens. I tried putting in a container or section, and does have any comportment giving me the ability to put it in background, or to put something else in foreground. 

**To Reproduce**
I tried to put an Background Image in a container, and other elements following in the same container.
I also tried to give childrens to Background Image. 

**Expected behavior**
I expect the background Image Component to be able to host childrens just as a container, or to be in the background of following components for example. 
In the end, it works just as the Image Element. For me it is a bug, as we expect the Background Image to be put in the background of other elements. 

**Additional context**

Is that a normal comportment or a bug ?
If not, how to use it as a background of other elements ? 

Thanks a lot for any help :)

PR #9477, opened 2023-01-30
Fix/background image

melohagan:
## Description
I was going to document the background image component, but it didn't appear to work as users' might expect.

Personally I felt that it made more sense for this component to be removed and a container setting to be added instead.

Otherwise, it may still be best to remove the background image component and allow users to create a custom component that can take children.

Addresses: 
- https://github.com/Budibase/budibase/issues/7972

## Screenshots
Screenshot 2023-01-30 at 12 18 01

*In this case, the screen has one background image, and the container has another.*

## Documentation

*I will update the docs if this change is approved.*


packages/client/manifest.json
```diff
@@ -998,70 +998,6 @@
       }
     ]
   },
-  "backgroundimage": {
```
shogunpurple, line 1001: For backwards compat, we might be better just to keep this around in the manifest but not expose the component in the builder - seems like over 100 are created monthly.

Screenshot 2023-01-30 at 16 21 58

packages/client/src/utils/styleable.js
```diff
@@ -6,7 +6,9 @@ import { builderStore } from "stores"
 const buildStyleString = (styleObject, customStyles) => {
   let str = ""
   Object.entries(styleObject || {}).forEach(([style, value]) => {
-    if (style && value != null) {
+    if (style === "background-image-url" && value != null) {
+      str += `background-image: url("${value}"); background-repeat: no-repeat; background-size: cover; `
```
shogunpurple, line 10: Hm, this seems like quite a dangerous place to do this - is it not possible to add this directly into the container component to encapsulate a bit?

melohagan: I had a go at moving this code into Container.svelte but I wasn't able to get it working. I'm not sure of the order in which the styling is set.

Maybe @aptkingston has an opinion?

packages/client/manifest.json
melohagan: Manifest entry restored.

packages/client/src/utils/styleable.js
aptkingston: Sorry I was completely sure I had replied to this...

I would prefer not to add in this sort of edge case, but it's reasonably safe if we need to. I would probably rather extend the style definitions to support something like a `format` key, so for URL you could define it as:
```js
    {
      control: Input,
      label: "Image URL",
      key: "background-image",
      format: url => url ? `url("${url}"); background-repeat: no-repeat; background-size: cover;` : ""
    },
```
Which would then require no changes to `styleable.js` and also be reusable in future.

melohagan:
Closing in favour of: https://github.com/Budibase/budibase/pull/9981

Issue #9480, opened 2023-01-30
REST API Authentication secret value replacement


mike12345567:
With the addition of the environment variable feature we have introduced a password replacement feature for datasources. This feature does not extend into REST queries and therefore API authentication is not covered.

This issue tracks the work required to make this possible as it is relatively separate to the datasource system and was no specified as part of the main environment variable work.

PR #9481, opened 2023-01-30
Password replacement for REST authentication basic type

mike12345567:
## Description
Update to cover the authentication for REST, replacing it with secret value and converting to password field, as well as minor update to env dropdown to only be a password field when not containing an environment variable - #9480.

Very minor update overall, just a specific set of functions to cover this edge case, all other fields should be within the config as password types ideally.

Addresses: 
- #9480


PClmnt (approved):
LGTM, nice change!

PR #9489, opened 2023-01-31
New App Onboarding

Ghrehh:
Our new App onboarding experience, currently shows up when a user has no apps, even if they've previously created apps but deleted them all. We don't want that secondary behaviour, but that'll require some backend changes I believe, so maybe best to do that in a second pass.

Further context on some changes provided as comments below.

Screenshot 2023-01-31 at 16 04 11
Screenshot 2023-01-31 at 16 04 24
Screenshot 2023-01-31 at 16 04 40


packages/builder/src/components/backend/DatasourceNavigator/icons/GoogleSheets.svelte
```diff
@@ -182,3 +182,9 @@
     </g>
   </g>
 </svg>
+
+<style>
```
Ghrehh, line 186: Some of these icons weren't centered, changing this doesn't affect the look of the other places they're consumed negatively.

packages/builder/src/components/common/FontAwesomeIcon.svelte
```diff
@@ -0,0 +1,15 @@
+<script context="module">
+  import { dom, library } from "@fortawesome/fontawesome-svg-core"
+  import {
+    faFileArrowUp,
+    faChevronLeft,
+  } from "@fortawesome/free-solid-svg-icons"
+  library.add(faFileArrowUp, faChevronLeft)
+  dom.watch()
+</script>
```
Ghrehh, line 9: I needed to add Font Awesome icons here also, this'll probably cause conflicts with the menu PR

packages/builder/src/pages/builder/portal/_layout.svelte
```diff
@@ -45,7 +45,9 @@
   })
 </script>
 
-{#if $auth.user && loaded}
+{#if $page?.path?.includes("/builder/portal/apps/create")}
```
Ghrehh, line 48: This is actually wrong, and I'll need to fix this, will look in a moment

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/DatasourceConfigPanel.svelte
```diff
@@ -0,0 +1,108 @@
...
+  import PanelHeader from "./PanelHeader.svelte"
+
+  export let title = ""
+  export let onBack = null
+  export let onNext = () => {}
+  export let fields = {}
+
+  let errors = {}
+
+  const formatName = name => {
```
Ghrehh, line 13: Function to format certain fields with irregular casing.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/ExampleApp.svelte
```diff
@@ -0,0 +1,254 @@
...
+      email: "francis.castro@example.com",
+      address: "3970 High Street",
+      city: "Wells",
+      postcode: "X12 6QA",
+      phone: "017684 23551",
+    },
+  ]
+</script>
+
+<div tabindex="-1" class="exampleApp">
```
Ghrehh, line 45: I chose to just reimplement a dummy version of the app, drawer and table here. There were issues with themes and such and this seemed simpler.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/index.svelte
```diff
@@ -0,0 +1,198 @@
...
+    data.append("name", name.trim())
+    data.append("url", url.trim())
+
+    if (useSampleData) {
+      data.append("sampleData", true)
+    }
+
+    const createdApp = await API.createApp(data)
+
+    // TODO I'm unsure if all this stuff is necessary, app creation seems to work without it but
```
Ghrehh, line 43: Anyone any context on if these methods are necessary?

packages/frontend-core/src/components/SplitPage.svelte
```diff
@@ -25,6 +25,7 @@
     overflow-y: auto;
   }
   .right {
+    overflow: hidden;
```
Ghrehh, line 28: Was having trouble with elements not respecting 100% width here and overflowing until I added this. I wonder if it's because this is the child of a grid parent?

aptkingston: Yea that's fine to add it. I have added position relative in a separate commit which is what is conflicting here. That might actually also solve this problem, but it's fine to have both.

packages/builder/src/components/common/FontAwesomeIcon.svelte
aptkingston: Yea in terms of this, I think we should merge the help menu PR, then update this one to fix any FA conflicts and (more importantly) fully revert any lockfile changes.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/ExampleApp.svelte
aptkingston: Yep that's totally fine. Makes sense to hardcode everything here to keep it as safe as possible.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/index.svelte
aptkingston: If it's the same code that runs when you normally create an app then I would keep it. I'm not entirely sure either why it's all required but I'd rather play it safe.

I agree it looks entirely redundant and we have so many wasted API calls but we don't have time to investigate it fully, so just go for whatever the normal app creation flow does.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/PanelHeader.svelte
```diff
@@ -0,0 +1,55 @@
...
+    class="budibaseLogo"
+    src={"https://i.imgur.com/Xhdt1YP.png"}
+  />
+  <div class="headingAndBack">
+    {#if onBack}
+      <button on:click={onBack}>
+        <FontAwesomeIcon name="fa-solid fa-chevron-left" />
+      </button>
+    {/if}
+    <h1>{title}</h1>
```
aptkingston, line 21: We should use a BBUI Heading component here to ensure consistent colour and font.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/PanelHeader.svelte
```diff
@@ -0,0 +1,55 @@
...
+  </div>
+  <p>{subtitle}</p>
```
aptkingston, line 23: We should use a BBUI Body component here to ensure consistent colour and font. You can use global CSS rules if you need to customise with margin bottom etc.

aptkingston (changes requested):
Code looks good. Checked this out and had a play as well. Works great but just spotted a couple of things we need to fix:

- The top nav disappears when you create an app that is not your first app
image
- The app created via the first app onboarding doesn't have the default home screen. Maybe you're just missing the call to create it because I'm fairly certain it's a separate API call after creating the app

packages/builder/src/components/backend/DatasourceNavigator/icons/GoogleSheets.svelte
aptkingston: No probs.

packages/builder/src/pages/builder/portal/apps/_FirstAppOnboarding/DatasourceConfigPanel.svelte
aptkingston: Nice improvement actually over the actual data section!

aptkingston (approved):

PR #9491, opened 2023-01-31
Removing Cypress

Mitch-Budibase:
We no longer need Cypress.

I am removing the Cypress directory and making any other necessary changes


Pedro-Revez-Silva (approved):
LGTM

MihailHadzhiev2022 (approved):

packages/builder/setup.js
```diff
@@ -1,14 +1,14 @@
-const cypressConfig = require("../cypress.json")
```
shogunpurple, line 1: I would possibly rename this file as `testSetup.js` or similar as it's only for testing.

packages/builder/ts/setup.ts
```diff
@@ -0,0 +1,4 @@
+// @ts-ignore
```
shogunpurple, line 1: Same as above.

shogunpurple (approved):
Code LGTM - just a minor naming comment and looks like the qa-core tests are failing due to this change. Once fixed it's good to go!

aptkingston:
The day has finally come :pray: 

I'll hope on this branch and remove all traces of `data-cy` in our components. Makes sense to include that removal in this PR too.

packages/builder/ts/setup.ts
Rory-Powell: We could probably move this out of the `builder` now all together, as it's only api related. It should probably live it qa-core itself

packages/builder/setup.js
Rory-Powell: Similar to other comment, we can move this directly into qa-core now

Issue #8236, opened 2022-10-13
Query for CouchDB datasource does nothing.


Cityjohn:
**Self hosted version 1.4.7**

Maybe I don't understand the workings of CouchDB at all but,

When I run this query in Fauxton:

![image](https://user-images.githubusercontent.com/34970833/195496516-4c9fe782-7893-4343-8883-a8bc7771579e.png)

It returns the request:

![image](https://user-images.githubusercontent.com/34970833/195496637-d3d02ee5-59da-41a5-bc34-cb160a0c6a92.png)

But when I run the same query in budibase:

![image](https://user-images.githubusercontent.com/34970833/195496668-56785be2-7a5d-46ad-b7b2-1a9a282406b9.png)

It still returns the entire database no matter what I put in the query box, it has no effect what so ever.

![image](https://user-images.githubusercontent.com/34970833/195496716-ce8dd9bc-0ddf-49ac-b21e-1a86466e4381.png)

I've tried all sorts of variations.

melohagan:
Hey @Cityjohn 

The current read function in the CouchDB connector uses `allDocs` which is documented here: https://pouchdb.com/api.html#batch_fetch

You can filter using the 'key' attribute like so: 
![Screenshot 2022-10-18 at 11 13 18](https://user-images.githubusercontent.com/101575380/196402805-59579ead-e1ff-4320-9040-c23fa37fb2df.png)

I need to write up some documentation for the CouchDB connector!

Cityjohn:
Hmmm, I only recently became familiar with NoSQL data storage. I have to be honest, I still don't know how I can use the query to return all users.

I am currently using this Javascript code to filter based on users that are in the table 'ta_users' and it seems to be doing the trick.

![image](https://user-images.githubusercontent.com/34970833/196717326-862fa369-855e-4716-9f09-c99f6e1c8dc5.png)

`var result = data.filter(x => x.tableId === 'ta_users');
return result`

melohagan:
Docs added: https://docs.budibase.com/docs/couchdb

PR #9493, opened 2023-01-31
CouchDB integration fixes

mike12345567:
## Description
Updates to use our new Nano layer for CouchDB integration rather than PouchDB.

This previously wasn't a very useful integration because it used PouchDB and broke fairly often, now we can use our own CouchDB integration based on Nano which improves performance and functionality.

Addresses #8236


melohagan (approved):
LGTM!

Issue #9502, opened 2023-02-01
App state binding does not work as expected


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.2.23

**Describe the bug**
Following through this tutorial: https://docs.budibase.com/docs/tag#removing-tags

There is this **Update State** action:
![image](https://user-images.githubusercontent.com/101575380/216049074-b130d2ca-d73c-4cfe-bfd8-e5e1551d79f0.png)

With this JavaScript binding:
```js
let removedTags = $("State.RemovedTags") || {};
removedTags[$("Tags Repeater block.Tags._id")+$("Repeater.Movies._id")] = true;
return removedTags;
```
*I also tried* `let removedTags = $("State.RemovedTags") ? { ...$("State.RemovedTags") } : {};` *just incase, but same issue*.

I would expect the `removedTags` to append a field, and not be a brand new object:
![fast_close](https://user-images.githubusercontent.com/101575380/216051210-878b0e3a-7d78-4045-9efc-426368ae877b.gif)

**To Reproduce**
1. Follow through the tutorial (or you can use the app export provided below).
2. Preview the app and see that two actions cannot be removed.
3. Click `Define actions` of the Tag component
4. In **Update State**, edit the JS binding. Remove the `$("State.RemovedTags")` reference and re-add it via the bindings helper menu. Click `Save`.
5. Preview the app again. Both tags can now be removed, despite the binding code being identical.

**Expected behavior**
It appears to me that because I am setting the **RemovedTags** state variable at the same time as trying to reference it, the actions are saved without the binding mapping. 
The save button should figure out if a state variable is being created, and make sure any references are pointing to that state variable. 
  
**App Export**
[Tag tutorial-export-1675255767246.tar.gz](https://github.com/Budibase/budibase/files/10557237/Tag.tutorial-export-1675255767246.tar.gz)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome

melohagan:
@aptkingston The reason that worked for you was because the state variable was already created at that point. I have a PR out for the bug fix, which includes a GIF that should hopefully highlight the problem more clearly!

aptkingston:
The state variable is undefined actually initially (here's a video to show):

https://user-images.githubusercontent.com/9075550/216089104-a03fe254-296b-4d14-b095-2a047f07c7be.mp4

[...]

PR #9505, opened 2023-02-01
Refresh Update state bindings before save complete

melohagan:
## Description
Needed to refresh the update state bindings before saving the actions to make sure newly created state variables were referenced correctly.

Addresses: 
- https://github.com/Budibase/budibase/issues/9502

## App Export
See issue.

## Screenshots
fast_close

## Documentation

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/ButtonActionEditor.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/ButtonActionEditor.svelte
@@ -5,6 +5,10 @@
   import ButtonActionDrawer from "./ButtonActionDrawer.svelte"
   import { automationStore } from "builderStore"
   import { cloneDeep } from "lodash/fp"
+  import {
+    readableToRuntimeBinding,
+    runtimeToReadableBinding,
+  } from "builderStore/dataBinding"
 
   const dispatch = createEventDispatcher()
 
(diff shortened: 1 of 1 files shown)


aptkingston (commented):
I understand what you mean - the state binding doesn't exist when you're creating the state binding for the first time, so it's impossible to get it to update itself because the readable to runtime conversion will fail.

You solution here works, but am I correct in thinking that this will allow that initial conversion to work, but still won't provide the binding? So you would have to manually type out the state binding for itself.

I had a quick idea and tested it out and I've got it actually adding a binding for itself now - so not only does the conversion work, but you get the benefit of having a real binding listed in the side panel, even when you're creating the state value for the first time.

Would having the binding listed be even nicer here? If so, check out the diff for that branch here:
https://github.com/Budibase/budibase/compare/add-own-state-binding?expand=1
It's basically just pulling out the state binding generation logic to be reusable, then using that to append a binding for it's own state key if required.

melohagan:
@aptkingston That would be nice to have the binding available! Would that expose the state binding for both the **Update State** action and the **Update Field Value** actions, as per this example?

aptkingston:
Ah no - I was only attempting to solve the issue of a state binding not being able to reference itself. Yea I can see that there's the other use case of all unsaved field actions being able to reference newly created state bindings. I'll update that branch to move the binding generation up a level and account for that too!

PR #9507, opened 2023-02-01
New Onboarding URL Validation

Ghrehh:
I was trying to be fancier with this and see if `new URL(theUrl)` threw an error instead of using a regex, but that seems to escape things like spaces and special characters automatically, so it didn't work. Opted to just use the existing regex that only rejects spaces.

Could have made it reject `/`s also, but the current app creation flow accepts it, so I assume they just escaped on the backend along with other special characters.

One thing I didn't copy from the current url validation was that it seems to prevent the user from using certain strings like `http`, this seems a little odd, as http could be included in a valid path or something 🤷 Happy to add it again if folks think it should stay, or if I'm missing something.

Also didn't add the validation that checks if the app already exists, as the user shouldn't have any apps.

### DIFF
--- a/packages/builder/src/components/start/CreateAppModal.svelte
+++ b/packages/builder/src/components/start/CreateAppModal.svelte
@@ -26,7 +26,15 @@
 
   const values = writable({ name: "", url: null })
   const validation = createValidationStore()
-  $: validation.check($values)
+
+  $: {
+    const { name, url } = $values
+
+    validation.check({
+      name,
+      url: url?.[0] === "/" ? url.substring(1, url.length) : url,
+    })
+  }
 
   onMount(async () => {
     const lastChar = $auth.user?.firstName
@@ -87,7 +95,11 @@
     appValidation.url(validation, { apps: applications })
     appValidation.file(validation, { template })
     // init validation
-    validation.check($values)
+    const { name, url } = $values
+    validation.check({
+      name,
+      url: url?.[0] === "/" ? url.substring(1, url.length) : url,
+    })
   }
 
   async function createNewApp() {
--- a/packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
+++ b/packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
@@ -1,6 +1,7 @@
 <script>
   import { Button, FancyForm, FancyInput } from "@budibase/bbui"
   import PanelHeader from "./PanelHeader.svelte"
+  import { APP_URL_REGEX } from "constants"
 
   export let name = ""
   export let url = ""
@@ -24,6 +25,10 @@
   const validateUrl = url => {
     if (url.length < 1) {
       return "URL must be provided"
+    }
+
+    if (!APP_URL_REGEX.test(url)) {
+      return 'Invalid URL'
     }
   }
 </script>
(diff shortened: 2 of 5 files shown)


packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
```diff
@@ -20,6 +21,10 @@
     if (url.length < 1) {
       return "URL must be provided"
     }
+
+    if (!APP_URL_REGEX.test(url)) {
+      return "Invalid URL"
```
aptkingston, line 26: Might be better to say "No spaces are allowed" or something, since the only thing this regex tests for is whitespace and it would be nice to inform the user why the URL is invalid to help them fix it.

aptkingston: To be honest I think we should use a regex that allows `[0-9a-zA-Z-_]` and nothing else... Paths (`/`) aren't actually allowed in app URLs either - they would probably break it.

Ghrehh: Done, could you give the most recent commit a look when you get a chance?

packages/builder/src/components/start/CreateAppModal.svelte
```diff
@@ -26,7 +26,15 @@
...
   const values = writable({ name: "", url: null })
   const validation = createValidationStore()
-  $: validation.check($values)
+
+  $: {
+    const { name, url } = $values
+
+    validation.check({
+      name,
+      url: url?.[0] === "/" ? url.substring(1, url.length) : url,
```
aptkingston, line 35: Could we not move this logic (of stripping a leading forward slash) into the validation store? Then we would only have to write it once, rather than duplicate this code fragment in a bunch of different places. Maybe there's a reason for it though, but it just looks a bit more verbose now that I though it needed to.

I notice you removed some of those checks from the validation store though in favour of adding them here so I presume there's a reason.

Ghrehh: I've done it this way so that this could be shared between the onboarding creation flow and the existing new app creation flow. Since the existing new app creation flow uses a leading slash and we don't want to allow the new onboarding to use that (presumably) this seemed like the best way. I don't think I removed anything else except what pertains to this, but I could be wrong.

aptkingston: OK, I still don't think I understand the harm in putting it in the store since it's conditionally removing it and the validation should really be identical anyway since it's both just creating an app, so using the same endpoint and would expect the same format of URL - there shouldn't be any differences. But this one has been open long enough so I'm happy just to get it in since it solves the original issue.

aptkingston (approved):

packages/builder/src/components/start/CreateAppModal.svelte
Ghrehh: Unless I'm misunderstanding it, it's because if we conditionally allowed leading slashes in the store then the onboarding flow would also allow users to enter urls with leading slashes and still have them be valid. Which might go against the idea of the UX of the new flow 🤷 

Happy to come back to this, but I'll merge for now since you approved.

PR #9508, opened 2023-02-01
Generate bindings for unsaved "Update State" actions

aptkingston:
## Description
This PR ensures that bindings are generated for all state entries, including newly created and unsaved ones, and makes them available to all button actions.

Good write up is documented in https://github.com/Budibase/budibase/pull/9505.

Addresses: 
- #9502


melohagan (approved):
LGTM! Pulled it down to try out and it works well - thank you! 😄

Issue #9512, opened 2023-02-02
Many to Many random constraint failures


royaljeffr:
I am receiving random errors when checking and unchecking a Dropdown box that pulls from a join table.  Here is my Postgres schema.

```
CREATE TABLE IF NOT EXISTS public.titles
(
    title character varying COLLATE pg_catalog."default",
    id integer NOT NULL GENERATED BY DEFAULT AS IDENTITY ( INCREMENT 1 START 112 MINVALUE 1 MAXVALUE 2147483647 CACHE 1 ),
    CONSTRAINT titles_pkey PRIMARY KEY (id)
)

CREATE TABLE IF NOT EXISTS public.titles_attributes
(
    type character varying(50) COLLATE pg_catalog."default",
    name character varying(100) COLLATE pg_catalog."default",
    description text COLLATE pg_catalog."default",
    id integer NOT NULL GENERATED BY DEFAULT AS IDENTITY ( INCREMENT 1 START 396 MINVALUE 1 MAXVALUE 2147483647 CACHE 1 ),
    CONSTRAINT titles_attributes_pkey PRIMARY KEY (id)
)

CREATE TABLE IF NOT EXISTS public.titles_attributes_join
  (
        title_id integer NOT NULL DEFAULT,
        title_attribute_id integer NOT NULL DEFAULT,
        id integer NOT NULL GENERATED BY DEFAULT AS IDENTITY ( INCREMENT 1 START 4061 MINVALUE 1 MAXVALUE 2147483647 CACHE 1 ),
        CONSTRAINT uq_title_attribute_ids UNIQUE (title_id, title_attribute_id)
        INCLUDE(title_id, title_attribute_id)
)
```

And then in Budibase I have this for my JOIN
![image](https://user-images.githubusercontent.com/49763873/216198850-a7530aa6-29d1-4463-b93a-a9154fbb7da4.png)

Then when I go to the Autogenerated screen and change the values from the picklist I get error message - error: duplicate key value violates unique constraint "uq_title_attribute_ids".  I can verify at the database level there no duplicates.  And when I get that message it usually deletes all the record associated with that title.
![image](https://user-images.githubusercontent.com/49763873/216199657-0fb16b4a-53d1-4118-a0d5-360d3e7d0f3e.png)

melohagan:
Hey @royaljeffr 

The next release which will be within the next week or so, will improve the relationship modal.

In this case the validation should have prevented you from saving without setting foreign keys.

Issue #6921, opened 2022-07-27
Multiple Many-to-one Relationships between the same two Tables breaks 


Debugnium:
**Hosting**
- Self
  - Method: docker compose
  - App Version: 1.1.21
  - Database: MS SQL Server
 
**Describe the bug**

Having multiple many-to-one relationships from Table A to Table B seems to break relationships:
![image](https://user-images.githubusercontent.com/77417343/181231846-05dc5e29-b9ba-4726-929b-4fa88849b39b.png)

**Expected behavior**

"Employees" relationships to load an Employee when viewing "Customers" data.

**Additional context**

In my case, a legacy database has 3 relations between customers and employees representing different roles.

The SQL seen in the XEvent Profiler uses an `or` in the left join, and in the select statement, it includes all of the `[Employees]` columns three times...
`left join [Employees] on [Customers].[Employee ID] = [Employees].[Employee ID] or [Customers].[OrgEmpID] = [Employees].[Employee ID] or [Customers].[Service ID] = [Employees].[Employee ID] `

This results in 3 rows with the same "Employees" data repeated 3 times in each row, rather than a single row with data for each employee.

_Removing two of the three relationships fixes the issue, but then two relationships are missing._

marcin9806:
Also, having many-to-many relationship on tables A-B-C prevents us from creating one-to-many relationship on tables A-B (even though it's the same thing, it is sometimes useful in forms building)

melohagan:
Possibly related issue in MySQL. 

Message from Discord:
> if I create/define an relationship with Table A to Table B, then do it Table A to Table C, the the A<->B goes away. IF I create a relationship Table B to Table A, then Table C to Table A, it works fine,

melohagan:
Closing as multiple Many->Many relationships with the same tables will no longer be supported: https://github.com/Budibase/budibase/pull/9421

This is because as well as the break, it would be substantially detrimental on performance. Better to use [Custom SQL queries](https://docs.budibase.com/docs/custom-queries) if you need to perform a lot of joins.

Issue #4604, opened 2022-02-21
SQL many-to-many breaks when through table has composite primary key


mjashanks:
I have the tables

```
Customer
- customer_id : PK

Category
- category_id : PK

Customer_category
- customer_id
- category_id
PK = composite of the above 2
```

I have fetched the tables and setup the relationship in Budibase OK.

The relationship works when I am adding values to the relationships. I.e. checking more values in the relationship picker.

However, if I try and remove (uncheck values) - it does not save the removed values.

Now, if I add an `id` to my `customer_category` table and set it to the PK (as below) - then everything works.

```
- customer_id
- category_id
- id : PK
```

mjashanks:
Also, its possible that I led you down the wrong path here. I was able to create the issue by the steps above - however, someone else seemed to encounter this issue and fix it themselves by recreating the relationship: https://github.com/Budibase/budibase/discussions/4408

v6admin:
Unfortunately, it's still happening to me, in multiple tables with a connection to an external postgresdb. I also tried to create several new (simple) apps with another external postgresql, mariadb and mysql db on a different host, but it still won't save removed relationship values (uncheck values in a relationship picker). It does save them. 

It doesn't work in the builder backend and it doesn't work in the app front-end. I don't have this problem when using the internal bb database, there I can remove relationships.

v6admin:
I created a new database and started from scratch. Instead of using the budibase database editor, I created and updated the schemes from within pgAdmin 4. That did the trick. The relationships and the associated tables are still being created from within budibase. It seems when you create tables and relationships from budibase, the database gets invalid.

PR #9516, opened 2023-02-02
SQL relationship modal

mike12345567:
## Description
This attempts to fix some of the issues with the modal which had shown up since the last re-work, this is a complex modal and could really do with more in-depth automated testing.

Issues found:
1. Random UX errors, like when you open the modal immediately showing an error
2. Errors weren't reactive, they would only occur when save was pressed
3. Errors weren't always accurate
4. Didn't correctly work when used from the table menu, rather than datasource menu
5. Pre-filling as much known information as is possible

### Addresses
 - https://github.com/Budibase/budibase/issues/9512
 - https://github.com/Budibase/budibase/issues/6921
 - https://github.com/Budibase/budibase/issues/4604

## Screenshots
When used from a table it auto-fills with as much as it can:
image

### DIFF
--- a/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
+++ b/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
@@ -176,7 +176,6 @@
       fromPrimary,
       fromForeign
     )
-    console.log(errObj)
     errors = errObj
     return getErrorCount(errors) === 0
   }
--- a/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
+++ b/packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
@@ -65,16 +65,8 @@
   $: isManyToOne = relationshipType === RelationshipTypes.MANY_TO_ONE
   $: toRelationship.relationshipType = fromRelationship?.relationshipType
 
-  function getFromTable() {
-    return plusTables.find(table => table._id === fromId)
-  }
-
-  function getToTable() {
-    return plusTables.find(table => table._id === toId)
-  }
-
-  function getThroughTable() {
-    return plusTables.find(table => table._id === throughId)
+  function getTable(id) {
+    return plusTables.find(table => table._id === id)
   }
 
   function invalidThroughTable() {
@@ -97,16 +89,16 @@
     if (
       originalFromTable &&
       originalToTable &&
-      originalFromTable === getFromTable() &&
-      originalToTable === getToTable()
+      originalFromTable === getTable(fromId) &&
+      originalToTable === getTable(toId)
     ) {
       return false
     }
     let fromThroughLinks = Object.values(
-      datasource.entities[getFromTable().name].schema
+      datasource.entities[getTable(fromId).name].schema
     ).filter(value => value.through)
     let toThroughLinks = Object.values(
-      datasource.entities[getToTable().name].schema
+      datasource.entities[getTable(toId).name].schema
     ).filter(value => value.through)
 
     const matchAgainstUserInput = (fromTableId, toTableId) =>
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -116,93 +90,90 @@
...
+    )
   }
 
-  function isColumnNameBeingUsed(table, columnName, originalName) {
-    if (!table || !columnName || columnName === originalName) {
-      return false
+  function getErrorCount(errors) {
+    return Object.entries(errors)
+      .filter(entry => !!entry[1])
+      .map(entry => entry[0]).length
```
melohagan, line 125: Why do we need the **map** if we're just getting the length?

packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -116,93 +90,90 @@
...
+
+  function validate() {
+    if (!allRequiredAttributesSet() && !hasValidated) {
+      return
     }
-    const keys = Object.keys(table.schema).map(key => key.toLowerCase())
-    return keys.indexOf(columnName.toLowerCase()) !== -1
+    hasValidated = true
+    errorChecker.setType(relationshipType)
+    const errObj = {}
```
melohagan, line 143: Could we remove the **errObj** variable and just use an object here like:

```js
errors = {
  relationshipType: errorChecker.relationshipTypeSet(relationshipType),
  fromTable: errorChecker.tableSet(fromTable),
  etc..
}
```

packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -342,6 +283,35 @@
     await tables.fetch()
     close()
   }
+
+  function changed(fn) {
+    if (fn && typeof fn === "function") {
```
melohagan, line 288: I think this could simply be: `if (typeof fn === "function")`

packages/builder/src/components/backend/Datasources/relationshipErrors.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  differentTables(table1, table2, table3) {
+    // currently don't support relationships back onto the table itself, needs to relate out
+    const error = table1 && (table1 === table2 || (table3 && table1 === table3))
+    return error ? mustBeDifferentTables : null
+  }
+
+  columnBeingUsed(table, column, ogName) {
+    return isColumnNameBeingUsed(table, column, ogName) ? columnCantExist : null
```
melohagan, line 78: NAB: Could change the `columnCantExist` variable to simply be `columnBeingUsed` ?

melohagan (approved):
LGTM, although I noticed a few minor issues while testing it out:

**Validation triggered for foreign keys in existing relationship, but not in a new relationship.**


**From and To column validation is triggered one by one and not both together.**



**Error state in One->Many is not cleared when switching to Many->Many. Cannot save even if now valid state.**



#### Feature requests

Foreign keys of through table don't check the primary key types of the From/To tables:


*In this case the foreign key of from_date should be invalid because it does not match the integer type of the employees primary key*.

Rather than use the *through* text with ellipsis, it would be better to use a double headed arrow for the Many->Many relationships, i.e. departments < - > employees

melohagan (changes requested):
I just noticed this fix has been undone: https://github.com/Budibase/budibase/pull/9421

packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
mike12345567: Unsure - will remove it, just moved that code around!

melohagan (approved):

PR #9517, opened 2023-02-02
Ensure tour popover disappears when finished

aptkingston:
## Description
Fixes an issue where the tour popover does not disappear when it is finished.

Diff looks weird but it's literally just wrapping an `{{#if...}}` block around the tour popover.


deanhannigan (approved):
LGTM!

PR #9519, opened 2023-02-02
New Onboarding Loading States

Ghrehh:
Adds a loading indicator for when app creation is in progress. Previously it would just hang, which is confusing for the user.


shogunpurple (approved):
LGTM!

aptkingston:
Visually all good - but noticed if you fail to connect toa datasource then it gets bricked in the loading state:
image

So might just need to unset the loading flag when failing to connect. Also should we show a better error message here, like failed to connect to datasource?

aptkingston (approved):
Changes look good!

PR #9522, opened 2023-02-02
Chore - vscode setup

adrinr:
## Description
Updating contributor docs and vscode setup

1. Updating docs to add python dependency
2. Adding a script that installs all the dependencies using asdf
3. Updating documentation accordingly
4. Setting up recommended extensions

This PR suggests using asdf as a recommended choice. The main reasons are:
1. It manages node, python, and many more dependencies with a single tool
2. It automatically loads the required version for the specific folder, so the developer does not need to worry about changing environments


scripts/install-contributor-dependencies.sh
```diff
@@ -0,0 +1,27 @@
+# Install brew
+if ! command -v brew &> /dev/null
+then
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+fi
```
Rory-Powell, line 5: Should this be wrapped in a check for `uname == "Darwin"`?

Rory-Powell: Similar to below, maybe we should just rename this file as macos specific script and note that from the contributing guide

docs/CONTRIBUTING.md
```diff
@@ -73,36 +72,56 @@ A component is the basic frontend building block of a budibase app.
...
+#### 1. Prerequisites
+
+- NodeJS version `14.x.x`
+- Python version `3.x`
+
+#### 1.1 Using asdf (recommended)
+
+Asdf is a package manager that allows managing multiple dependencies.
+
+1. Install using script (recommended):
```
Rory-Powell, line 100: Very minor: the `1.` and `2.` implies to me this is a set of instructions to follow. So reads like first install using this script then do these manual steps. Might be clearer as two bullets instead

.tool-versions
```diff
@@ -0,0 +1,2 @@
+nodejs 14.19.3
+python 3.11.1
```
Rory-Powell, line 2: very nice! for the people already on nvm (most of the team) it might be good to add an `.nvmrc` also. Not sure if there is an equivalent with pyenv

adrinr: I think there is a pyenv. I'll check it out

docs/CONTRIBUTING.md
adrinr: Good shout

.tool-versions
adrinr: Done

Rory-Powell (approved):
LGTM!

Issue #9523, opened 2023-02-02
Cannot assign Barcode/QR field in form


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/9471

<div type='discussions-op-text'>

<sup>Originally posted by **muksidin** January 30, 2023</sup>
Hello, 
I try to create data type Barcode/QR and than i create form manually. Then, i add component Barcode/QR and at field i can not find database with type Barcode/QR. And then i try create data type Text, at field of component Barcode/QR found it. This is bug or not?

Other problems is Barcode can not access camera, i already give it permission and grant but still **Your camera is disabled**.</div>

![](https://cdn.discordapp.com/attachments/978670708966162471/1070728175308517466/Screenshot_2023-02-02_at_15.31.03.png)

![](https://cdn.discordapp.com/attachments/978670708966162471/1070728175597916251/Screenshot_2023-02-02_at_15.31.27.png)

PR #9528, opened 2023-02-03
Parse field type configuration for invalid characters

deanhannigan:
## Description

The field type configuration for the barcode reader was not being parsed correctly. The field `type` string is now stripped of any invalid characters.

Addresses: 
- https://github.com/Budibase/budibase/issues/9523

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FormFieldSelect.svelte
+++ b/packages/builder/src/components/design/settings/controls/FormFieldSelect.svelte
@@ -25,16 +25,21 @@
   const getOptions = (schema, type) => {
     let entries = Object.entries(schema ?? {})
     let types = []
-    if (type === "field/options" || type === "field/barcode/qr") {
+    if (type === "field/options") {
       // allow options to be used on both options and string fields
       types = [type, "field/string"]
     } else {
       types = [type]
     }
 
-    types = types.map(type => type.slice(type.indexOf("/") + 1))
+    types = types.map(type => {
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/design/settings/controls/FormFieldSelect.svelte
```diff
@@ -25,16 +25,21 @@
...
       types = [type, "field/string"]
     } else {
       types = [type]
     }
 
-    types = types.map(type => type.slice(type.indexOf("/") + 1))
+    types = types.map(type => {
+      let fieldTypeRaw = type.slice(type.indexOf("/") + 1)
+      let fieldTypeParsed = fieldTypeRaw.replace("/", "")
+      return fieldTypeParsed
```
aptkingston, line 38: Was the problem here that the field type was `field/barcode/qr` but the actual datatype is `barcodeqr`?

aptkingston: The reason I ask is just that if it is, we could actually just update the manifest to change the setting type to `field/barcodeqr` couldn't we? Which would just be a one line change and not require and other changes (I think).

deanhannigan: That's essentially it. My only concern in modifying the type was any potential side-effect that I might miss. I've run through some more testing and it does seem a manifest update is all that's required here.

aptkingston (approved):
Nice, glad that worked. LGTM!

PR #9546, opened 2023-02-06
Fix Import Validations

Ghrehh:
We previously weren't updating the constraints of the schema when a new field type was selected in the import modal. This meant that every field had the string type validation applied to it. This should fix that issue.


aptkingston (approved):

shogunpurple (approved):
LGTM! Although we should merge this one into `master` as a hotfix

shogunpurple:
I cherry picked this one into `master` too - as it's blocking a few E2E tests

Issue #8770, opened 2022-11-22
Budibase CLI memory usage improvements for large backup procedures


maiducan:
Hello, I cannot backup my full environment with budi backup command due to the fact, that my container (couchdb - 2.5GB) is a little bigger and the backup process exceeds the RAM limit (using all of the 12GB) and then crashes. 

- Self hosted, using Budi CLI / docker compose
- 12GB of RAM

Is there a possibility to update the command so that it utilise the memory more efficiently? Thank you

mike12345567:
Hi @maiducan - sadly its very difficult to make this more memory efficient easily, however you could try adding the `--max-old-space-size` when running the CLI, by default Node can only use 1.5GB of memory, you could set this to say 5GB with `--max-old-space-size=5120` and try this again.

mike12345567:
@maiducan - I see, I'll convert this issue into an enhancement request, i'm not sure what we can do to improve this - for now I would recommend looking into direct CouchDB replication as this will be much more performant in larger deployments.

PR #9550, opened 2023-02-06
CLI backup + app export memory usage improvements

mike12345567:
## Description
After seeing that some people have been having issues with the CLI backup functionality and high memory usage, I attempted to investigate the issue a bit further. After some help from @andz-bb I found that some large apps with a large number of revisions could cause the CLI to use very large amounts of memory. To counter this I updated the export functionality to use `style: "main_only"` which makes sure that the exports of apps do not include revision history, something that isn't really useful to use as an app export.

To do this I also had to fork one of the libraries we utilise for dumping databases, however this turned out to remove a deprecation warning we've had issues with - which proves to be a double benefit (every export of app caused a spam of warnings).

Forked repo: 
- https://github.com/Budibase/pouchdb-replication-stream
- Change: https://github.com/pouchdb-community/pouchdb-replication-stream/compare/master...Budibase:pouchdb-replication-stream:master

Addresses: 
- https://github.com/Budibase/budibase/issues/8770


Rory-Powell (approved):
LGTM 👍 Could you also link the change in the forked repo for visibility on the PR?

PR #9561, opened 2023-02-07
Support Chrome autofill

aptkingston:
This PR updates the new input fields to support chrome autofill a bit better. The old form fields didn't have any special handling for this either, so this is a flat out improvement.

Fields are now highlighted when autofill is used and don't use the awful built-in chrome styles.

Autofill looks like this:
image

And here's a video showing the fields being populated on page load:

https://user-images.githubusercontent.com/9075550/217244699-f966349a-538b-48bd-8c6b-3fd21c820529.mp4


shogunpurple (approved):
LGTM!

PR #9589, opened 2023-02-08
Datasource plugin app error

mike12345567:
## Description
Minor fix for the plugin datasource error that occurred when opening apps in an environment where the datasource no longer exists.

This error was reported on Discord appeared as such:
image

### DIFF
--- a/packages/server/src/sdk/app/datasources/datasources.ts
+++ b/packages/server/src/sdk/app/datasources/datasources.ts
@@ -63,8 +63,12 @@ function useEnvVars(str: any) {
 
 export async function removeSecrets(datasources: Datasource[]) {
   const definitions = await getDefinitions()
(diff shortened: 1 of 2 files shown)


packages/server/src/sdk/app/datasources/datasources.ts
```diff
@@ -65,6 +65,9 @@ export async function removeSecrets(datasources: Datasource[]) {
   const definitions = await getDefinitions()
   for (let datasource of datasources) {
     const schema = definitions[datasource.source]
+    if (!schema) {
```
Rory-Powell, line 68: Looks like a nice simple fix. Would we be able to add a unit test?

shogunpurple (approved):
LGTM!

PR #9629, opened 2023-02-09
Reactivity fix for the Create/Edit column modal.

deanhannigan:
## Description

Addresses an issue with the initialisation statement firing too many times when editing a data column. Some edits to the column configuration were being overwritten as a result. The offending statement will now only run once on load.


shogunpurple (approved):
LGTM!

Issue #9406, opened 2023-01-21
Screen freezes when filtering by date.


kzkondo:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.2.19
  - Postgres 15.1

**Describe the bug**
Screen freezes when filtering by date.

**To Reproduce**
1. Create a data provider.
2. Press Define Filter.
3. Press Add Filter and select Column.
4. For the column, select the date type of the relation destination.

**Screenshots**

https://user-images.githubusercontent.com/93172658/213859321-ddfcc932-2d83-44fc-93b9-7bc577076894.mov

**Desktop (please complete the following information):**
 - OS: MacOS 12.6.2
 - Browser chrom
 - Version 109.0.5414.87

melohagan:
Hey @kzkondo 

Could you provide your Postgres table schemas please?  For tables **groups** and **kouji**

kzkondo:
Hi! @melohagan Thanks for your cooperation!
Table schema is public.
The relationship between kouji and groups is one-to-one.
groups.id is the primary key and is connected to kouji.group_id.

melohagan:
Re-opening because it's still an issue for custom SQL queries:

![Screenshot 2023-02-20 at 10 20 04](https://user-images.githubusercontent.com/101575380/220078213-45d193e2-8286-4dd5-adec-ab99edfca715.png)

PR #9635, opened 2023-02-09
Date filter freeze fix

melohagan:
## Description
When doing a SQL filter on a date, the app was crashing because the schema fields were not including the relationship fields.
In addition, found a bug were `{}` was being passed as a value into the range. This was causing query issues. 
Wasn't able to find exactly what/were the invalid date (0000 or 9999) was being converted to an empty object - think it may have been in Knex. 
I updated the `sql.ts` code to handle empty objects. 

Addresses: 
- https://github.com/Budibase/budibase/issues/9406

## Screenshots
Screenshot 2023-02-09 at 19 26 02

Screenshot 2023-02-09 at 19 26 20

Screenshot 2023-02-09 at 19 26 47

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
@@ -101,7 +101,7 @@
   }
 
   const getSchema = filter => {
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
LGTM!

PR #9649, opened 2023-02-10
Fix for environment variable MS-SQL ports

mike12345567:
## Description
Fix issue with MS-SQL, port needs to be a number for the `node-mssql` package, with env vars it isn't anymore.

To fix this, added to the datasource SDK get functionality to check the types against the required integration types.


PClmnt (approved):
LGTM!

Issue #9405, opened 2023-01-20
Relationship picker + SQL improvements


mike12345567:
This is in regards to this spec item: https://budibase.atlassian.net/wiki/spaces/EN/pages/177242117/Encoded+Row+ID+filtering+issues

There has been some discussion around how to handle this, between @mike12345567 and @aptkingston - after some thought we have decided a good option would be to include the actual primary key fields in the relationship objects.

Right now when a row is returned, the relationships are denoted like: `{ primaryDisplay: "some value", _id: "encodedvalue" }` - this works well for our system as it can use the `_id` very easily, but for users that want to get the actual values of the primary keys (for further searching/updating or queries) they need to decode the `_id` field.

An example of this would be if the table has a composite key, made up of the `emp_no` and `dept_no` then the output object would look something like: `{ primaryDisplay: "some value", _id: "encodedvalue", emp_no: "key value", dept_no: "key2 value" }`.

This should make the binding experience a lot more pleasant for users, assuming that the relationship pickers and so on return the full binding (rather than just the `_id` value) then this could be quite simple to implement.

I've allocated to both firestorm and avalanche as work could be required with expertise from both sides.

poirazis:
Hey @mike12345567 
If it carries the related table name too, would we be able to do recursive relationship drilling !
( unless that's in the field schema )

mike12345567:
Further discussion has been had about this issue - the relationship picker currently returns the value of the `_id` field and therefore this is harder to update.

@aptkingston is going to look at whether there is an edge case approach we can take to this problem that will solve the problem for relationship pickers when they are used to return values outside the scope of a form.

PR #9653, opened 2023-02-10
Add Fetch Row Button Action

Ghrehh:
[Solves #9405](https://github.com/Budibase/budibase/issues/9405)

We discussed as a team the most of the problems outlined in the above ticket could be solved by nesting data providers.

There were certain scenarios where the hydrated relationship data might be needed within a button event or something similar, and to solve this use case we added an action that when provided with an `_id` will fetch a row.


aptkingston (approved):
LGTM!

shogunpurple (approved):
LGTM! We will likely need to update the docs for this to show people how to use it!

PR #9669, opened 2023-02-13
Configurable test log levels and common error handling

Rory-Powell:
### Error handling

- Add common middleware `errorHandling` 
   - Add to the top of router configuration
   - This ensures all http errors will have a consistent structure
- By default log errors for any error that:
   - Does not have a status
   - is less than 499
- For debugging, add additional flag `LOG_4XX` which will log 404s, 400s, 403 etc.  

### Test log levels
- Add better integration with the existing `LOG_LEVEL` environment variable
- Apply standard rules for log levels to visibility of `console.error`, `console.warn` etc. 
- Set test runs to `error` by default   

Addresses: 
- https://linear.app/budibase/issue/BUDI-6558/configurable-test-log-levels-and-common-error-handling

See also: 
- https://github.com/Budibase/budibase-pro/pull/110

### DIFF
--- a/packages/backend-core/src/environment.ts
+++ b/packages/backend-core/src/environment.ts
@@ -82,6 +82,7 @@ const environment = {
   SESSION_UPDATE_PERIOD: process.env.SESSION_UPDATE_PERIOD,
   DEPLOYMENT_ENVIRONMENT:
     process.env.DEPLOYMENT_ENVIRONMENT || "docker-compose",
+  LOG_4XX: process.env.LOG_4XX,
   _set(key: any, value: any) {
     process.env[key] = value
     // @ts-ignore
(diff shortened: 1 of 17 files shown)


packages/backend-core/src/environment.ts
```diff
@@ -82,6 +82,7 @@ const environment = {
   SESSION_UPDATE_PERIOD: process.env.SESSION_UPDATE_PERIOD,
   DEPLOYMENT_ENVIRONMENT:
     process.env.DEPLOYMENT_ENVIRONMENT || "docker-compose",
+  LOG_4XX: process.env.LOG_4XX,
```
mike12345567, line 85: This naming makes sense, but could be confusing to self hosters and it could be a very useful feature, a slightly more verbose name could be beneficial? It doesn't describe if it enables/disables etc.

mike12345567 (approved):

packages/backend-core/src/environment.ts
Rory-Powell: Added here: https://github.com/Budibase/budibase/commit/91e63c8ec41fdc865b98eb114eedd26ea3e289a4

PR #9670, opened 2023-02-13
Enable higher concurrency and resiliency in worker tests

Rory-Powell:
### Simplify worker test configuration
- Remove retries for setup
- No longer use constant `default` for tenant id - use a random instead
- Remove default user + tenant 1 user in favour single unique user in single unique tenant

### Locking

- Add a lock to tenant creation for manipulation of the `tenants` doc
- This partially addresses: https://linear.app/budibase/issue/BUDI-5239/global-info-tenants-is-prone-to-409-conflicts
- However this isn't the ideal solution, as a high lock retry limit is needed specifically for tests due to high contention
- We should leave this ticket open in favour of using individual tenant docs - this will speed up test runs also
- Also fix bug with `system` locks still trying to use a tenant id 

### Separate platform documents 
- Separate the creation of platform user docs from writing / update the tenants doc in the global-info db
- Bulk user creation runs through platform user creation in a loop - if left combined, in addition to the new tenancy lock, user creation could become very slow
- Only try to create tenant when admin user is created instead of on every user 

### Misc
- Remove remaining usages of `doWithGlobalDB`, now that pouch is gone we no longer need the functional style of accessing the database for a given tenant id
- Run worker and backend-core tests with CPU default threads instead of 2, this produces a faster test run and doesn't seem to affect overall performance of a full `yarn test`
- Remove sequential `seq` tests - these are no longer needed
- Add db test configuration for backend-core
   - The existing code was still referencing in-memory configuration to no affect 

### Address some circular dependencies 
- After some of the changes from above, circular dependencies became an issue producing `undefined` packages in some test runs. To fix:
- Don't re-export context from tenancy package inside backend-core
- Remove deprovisioning package, in favour of tenants sdk inside worker
   - Much of the logic in deprovisioning that caused circular deps wasn't needed in backend-core  

This reduces circular deps from 38:
Screenshot 2023-02-13 at 12 54 02

To 21:
Screenshot 2023-02-13 at 12 58 22

Related PR for introducing real redis integration to enable per-tenant locks to work:
- https://github.com/Budibase/budibase/pull/9671

Addresses: 
- https://linear.app/budibase/issue/BUDI-6559/enable-higher-concurrency-and-resiliency-in-worker-tests
- https://linear.app/budibase/issue/BUDI-5239/global-info-tenants-is-prone-to-409-conflicts

Also see:
- https://github.com/Budibase/budibase-pro/pull/111


mike12345567 (approved):

mike12345567:
Changes to platform DB are a good improvement!

PR #9671, opened 2023-02-13
Enable use of redis container in worker tests

Rory-Powell:
## Description

This is required as a part of: https://github.com/Budibase/budibase/pull/9670
Without using a shared redis container the per tests locks in memory will achieve nothing. 

- Add new environment variable `MOCK_REDIS` to enable `ioredis-mock` instead of `env.isTest()`
- Add integration with test containers to add redis url / port / password to env
- Move reading of the above inside functions instead of at sources load time
   - This allows for the environment to be updated after the container starts
   - Otherwise the default credentials would be read instead  

Addresses: 
- https://linear.app/budibase/issue/BUDI-6560/enable-use-of-redis-container-in-worker-tests

Also see:
- https://github.com/Budibase/account-portal/pull/256


shogunpurple (approved):
LGTM!

mike12345567 (approved):

Issue #3754, opened 2021-12-10
Prevent SSO users from setting / resetting a password


Rory-Powell:
**Describe the feature request**

If an SSO user:
- Prevent in app password reset 
- Prevent password recovery 

Introduce new environment flag to prevent lockout:
- `SSO_MAINTENANCE_MODE` 
   - When false, the above rules apply
   - When true, an sso user may set / reset and login with a password

PR #9672, opened 2023-02-13
Prevent SSO users from setting / resetting a password

Rory-Powell:
## Description

Prevent an sso user from performing any password based operation
- In-app update
- Request password reset email
- Perform update from password reset email
- Reset as admin

### Backend Core 
- Rename and restructure vague `/src/cloud` package into `src/accounts`
- Move the `platformLogout` function out of `src/utils` into `src/auth`
- Rename `third-party-common` middleware to `sso` and nest google and oidc related modules in `sso` package
- Move common operations from password, google and oidc auth into the auth sdk inside worker to reduce duplication:
     - creating a  session
     - signing a jwt
     - adding to user 
- Tests
   - Replace the manual mocks for the accounts package with an auto mock
   - Build out sso and user structures

### Types
- API
   - Add auth types e.g. password request, login
   - Add additional user types around email invites and creating admin user
- Documents 
   - Updates to accounts for sso
   - Updates to configs for oidc and google 
   - Updates to users for separating the `ThirdPartyUser` (Renamed `UserSSO`) from the base user
- SDK
   - Build out sso helpers for profile and authentication
   - Add dedicated type for updating self instead of runtime protection

### Worker
- Better type the `authInternal` function - renamed `passportCallback` for clarity
- Integrate above callback with session creation instead of requiring each middleware to do this
- Updates the self api 
   - Use new types
   - Integrate with users sdk instead of using DB and events directly  
- Build out auth sdk
   - login, logout,  password reset / update
   - Integrate sso password check into above
   - Update auth controller to call auth sdk
- Integrate accounts sdk with core accounts package to provide unified access to account operations
- Update users sdk:
   - Add `isPreventSSOPasswords` for common check for:
      - sso maintenance mode / admin user check
      - db user has sso fields
      - check root account for sso fields
   - Integrate user save with `isPreventSSOPasswords`  

Tests
   - API integration tests for sso users performing password operations
   - Users sdk tests for more specific scenarios for `isPreventSSOPasswords` 

Addresses: 
- https://github.com/Budibase/budibase/issues/3754

## Screenshots

Remove update password option for self
Screenshot 2023-02-20 at 16 55 35

Remove force password reset option 
Screenshot 2023-02-20 at 16 59 08

Error message on password reset request:
Screenshot 2023-02-20 at 17 01 46

Error message on login:
Screenshot 2023-02-20 at 16 28 16

## Documentation


packages/backend-core/src/accounts/accounts.ts
```diff
@@ -5,9 +5,14 @@ import { CloudAccount } from "@budibase/types"
 
 const api = new API(env.ACCOUNT_PORTAL_URL)
 
+const EXIT_EARLY = env.SELF_HOSTED || env.DISABLE_ACCOUNT_PORTAL
```
adrinr, line 8: What is the logic behind this exit early. It might be confusing, as any consumer of the service will not be aware of these functions returning always undefined. This "undefined" seems more of a business requirement, could it be moved to a higher implementation level?

packages/backend-core/src/auth/auth.ts
```diff
@@ -188,3 +192,32 @@ export async function updateUserOAuth(userId: string, oAuthConfig: any) {
...
+  const ctx = opts.ctx
+  const userId = opts.userId
+  const keepActiveSession = opts.keepActiveSession
+
+  if (!ctx) throw new Error("Koa context must be supplied to logout.")
+
+  const currentSession = getCookie(ctx, Cookie.Auth)
+  let sessions = await getSessionsForUser(userId)
+
+  if (keepActiveSession) {
```
adrinr, line 209: What is the use case of keeping the session active on logout?

packages/backend-core/src/events/identification.ts
```diff
@@ -166,7 +167,10 @@ const identifyUser = async (
   const type = IdentityType.USER
   let builder = user.builder?.global || false
   let admin = user.admin?.global || false
-  let providerType = user.providerType
+  let providerType
+  if (isSSOUser(user)) {
+    providerType = user.providerType
```
adrinr, line 172: If it's a non sso user, is the provider type undefined?

packages/backend-core/src/middleware/passport/local.ts
```diff
@@ -33,52 +28,31 @@ export async function authenticate(
...
 
-  // check that the user is currently inactive, if this is the case throw invalid
+  // ACTIVE
   if (dbUser.status === UserStatus.INACTIVE) {
+    console.info(`user=${email} is inactive`, dbUser)
     return authError(done, INVALID_ERR)
   }
 
-  // check that the user has a stored password before proceeding
+  // HAS PASSWORD
```
adrinr, line 44: Are these comments required? The code looks verbose enough

packages/backend-core/src/middleware/passport/sso/sso.ts
```diff
@@ -0,0 +1,163 @@
...
+) => Promise<any>
+
+/**
+ * Common authentication logic for third parties. e.g. OAuth, OIDC.
+ */
+export async function authenticate(
+  details: SSOAuthDetails,
+  requireLocalAccount: boolean = true,
+  done: any,
+  saveUserFn?: SaveUserFunction
```
adrinr, line 26: This should not allow `undefined` if then we throw an error. This should be typed as required and use to the validation in build time

packages/backend-core/src/middleware/passport/sso/sso.ts
```diff
@@ -0,0 +1,163 @@
...
+) {
+  if (!saveUserFn) {
+    throw new Error("Save user function must be provided")
+  }
+  if (!details.provider) {
```
adrinr, line 31: Same as above, can this be done removing the undefined option on typing?

packages/backend-core/src/middleware/passport/sso/sso.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  const ssoUser = await syncUser(dbUser, details)
+  // never prompt for password reset
+  ssoUser.forceResetPassword = false
+
+  // create or sync the user
+  try {
+    await saveUserFn(ssoUser, { hashPassword: false, requirePassword: false })
```
adrinr, line 90: Should `hashPassword` not be undefined, instead of a false boolean?

packages/backend-core/src/middleware/passport/sso/sso.ts
```diff
@@ -0,0 +1,163 @@
...
+    return authError(done, "Error saving user", err)
+  }
+
+  // now that we're sure user exists, load them from the db
+  dbUser = await users.getById(dbUser._id!)
+
+  return done(null, dbUser)
+}
+
+async function syncProfilePicture(user: User, details: SSOAuthDetails) {
```
adrinr, line 101: Is this a `sync` or a `get`?

packages/types/src/documents/account/account.ts
```diff
@@ -79,14 +79,24 @@ export const isSelfHostAccount = (account: Account) =>
...
-export interface SSOAccount extends Account {
-  pictureUrl?: string
-  provider?: string
-  providerType?: string
+export enum AccountSSOProviderType {
+  GOOGLE = "google",
+}
+
+export enum AccountSSOProvider {
+  GOOGLE = "google",
```
adrinr, line 87: What is the difference between this and the type? Should these field not be related?

packages/worker/src/sdk/users/events.ts
```diff
@@ -84,6 +84,10 @@ export const handleSaveEvents = async (
     ) {
       await events.user.passwordForceReset(user)
     }
+
+    if (user.password != existingUser.password) {
```
adrinr, line 88:
```suggestion
    if (user.password !== existingUser.password) {
```

packages/backend-core/src/accounts/accounts.ts
Rory-Powell: This is a cloud vs self hosted consideration. We only want to use this client in the case of being cloud deployed. 
The current usages always have to wrap the call with their own checks. e.g:
- https://github.com/Budibase/budibase/blob/1dc3d44a44e5207c63cc1ec6a8e3ea247e5d0765/packages/backend-core/src/cache/user.ts#L17
- https://github.com/Budibase/budibase/blob/8d4b59126afc99cebad4bfcafa3faf923f7d51a0/packages/worker/src/sdk/users/users.ts#L567
-  https://github.com/Budibase/budibase/blob/8d4b59126afc99cebad4bfcafa3faf923f7d51a0/packages/worker/src/sdk/users/users.ts#L188


I've added this check internally now so everything is centralised here instead. 

The type signature for `getAccount` and `getAccountByTenantId` already tell us undefined is an potential response, I've updated `getStatus` as well to show the response format and the potential for undefined. 

see: https://github.com/Budibase/budibase/pull/9672/commits/6750a035b96510c586d796301ace93c2a32a7365

packages/backend-core/src/auth/auth.ts
Rory-Powell: The only current use case is updating your own password from inside the builder

https://github.com/Budibase/budibase/blob/c4ef5b30bf56e7cca987116def5aa24afeb31ba7/packages/worker/src/api/controllers/global/self.ts#L137-L142

packages/backend-core/src/events/identification.ts
Rory-Powell: yep that's the intention, it's an optional field on the identity

packages/backend-core/src/middleware/passport/local.ts
Rory-Powell: I thought the original comments were slightly too verbose which is why I shortened to the simpler statements. With the new log lines they probably are a bit overkill to have them at all now - removed in https://github.com/Budibase/budibase/pull/9672/commits/46f98f76122f88058b9e1231e3a8ff14dda59829

packages/backend-core/src/middleware/passport/sso/sso.ts
Rory-Powell: Good call on this - it's likely the source of a bug with datasource auth and user token refresh - updated in: https://github.com/Budibase/budibase/pull/9672/commits/a8ab00ca0eba9cc3e13d7de170d8a64d43ab0aa5

packages/backend-core/src/middleware/passport/sso/sso.ts
Rory-Powell: yep - doesn't need to be there anymore. this is a hangover from when we were pure js - removed in: https://github.com/Budibase/budibase/pull/9672/commits/e7461e180dc1ee447ff23f3b885feca732fb2710

packages/backend-core/src/middleware/passport/sso/sso.ts
Rory-Powell: In this case we have to be specific with `false` as the default is `true` in the save function

packages/backend-core/src/middleware/passport/sso/sso.ts
Rory-Powell: yeah better as get - updated: https://github.com/Budibase/budibase/pull/9672/commits/a4a6d0d7c762df4258d7ce945f891a39c67054f3

packages/types/src/documents/account/account.ts
Rory-Powell: The sso fields on the account were modelled around the same fields on the budibase users, where: 
- providerType = the auth mechanism, in budibase this is either `oidc` or `google` 
- provider = the provider for the mechanism, always `google` for google type but can be any external identity provider when using oidc auth. 

The account fields are currently limited to google only as it's the only one we support.

packages/backend-core/src/utils/utils.ts
```diff
@@ -38,7 +30,7 @@ export async function resolveAppUrl(ctx: Ctx) {
   const appUrl = ctx.path.split("/")[2]
   let possibleAppUrl = `/${appUrl.toLowerCase()}`
 
-  let tenantId: string | null = tenancy.getTenantId()
+  let tenantId: string | null = context.getTenantId()
```
mike12345567, line 33: Something I've been thinking about - should really get rid of all context based stuff from tenancy, could be a worthwhile refactor moving it all over at some point.

packages/types/src/documents/global/config.ts
```diff
@@ -45,7 +47,7 @@ export interface OIDCConfiguration {
   callbackURL: string
 }
 
-export interface OIDCInnerCfg {
```
mike12345567, line 48: Good improvement, always better to be verbose in type names!

packages/worker/scripts/dev/manage.js
```diff
@@ -29,6 +29,7 @@ async function init() {
       SERVICE: "worker-service",
       DEPLOYMENT_ENVIRONMENT: "development",
       TENANT_FEATURE_FLAGS: "*:LICENSING,*:USER_GROUPS,*:ONBOARDING_TOUR",
+      ENABLE_EMAIL_TEST_MODE: 1,
```
mike12345567, line 32: Should this be on in all dev environments?

mike12345567 (approved):
LGTM!

packages/backend-core/src/utils/utils.ts
Rory-Powell: It was done more out of necessity because of some circular deps in backend-core. It's just externally now that the context package is re-added to tenancy for backwards compatibility:
https://github.com/Budibase/budibase/blob/bd09914e4ef41da73884817d79e27b3891bc2882/packages/backend-core/src/index.ts#L33-L36

It wouldn't be that big of a job now to just remove that, build and fix the fallout.

packages/types/src/documents/global/config.ts
Rory-Powell: As long as we don't get the to point of `OIDCStrategyFactoryProducerWrappingBean` I agree 😂

packages/worker/scripts/dev/manage.js
Rory-Powell: Should be good, all the email sending is nicely encapsulated in the worker. I'll add something similar to the account portal when I move onto the sso updates there too

Issue #8560, opened 2022-11-07
Trigger binding not adding on click


melohagan:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.1.15
  - App Version: 2.1.15

**Describe the bug**
Trigger bindings do not get added to handlebars expression in automation bindings drawer.

**To Reproduce**
1. Add an automation with **Row Created** trigger
2. Add an action, e.g. **Create Row**
3. Click the flash icon for a binding, and try to add a trigger binding with the mouse, e.g. `Trigger.Row`
4. See that it does not add the binding
5. Weirdly, if you click twice it will add the binding (once)

**Expected behavior**
Within automations, I should be able to add a Trigger binding with the mouse (without double click). 

**Screenshots**
![Screenshot 2022-11-07 at 14 53 58](https://user-images.githubusercontent.com/101575380/200340682-96f137f6-3a96-46cc-9059-1ca72dd4b17b.png)
*Clicking these bindings from the side panel does not add them into the input*
  
**App Export**
[My app-export-1667832794642.tar.gz](https://github.com/Budibase/budibase/files/9952779/My.app-export-1667832794642.tar.gz)

PR #9688, opened 2023-02-14
Fix Automation Bindings Panel Requiring a Double Click

Ghrehh:
## Description
The popover outside click handler was eating the initial click a user made to select a binding suggestion, since were using the Popover component as a tooltip here it seems fair to add an option to make Popovers un-dismissable which avoids the issue.

Addresses: 
[#8560](https://github.com/Budibase/budibase/issues/8560)


packages/bbui/src/Popover/Popover.svelte
```diff
@@ -18,6 +18,7 @@
   export let useAnchorWidth = false
   export let dismissible = true
```
aptkingston, line 19: Do we not already have this `dismissible` prop to do exactly this - by setting it to false? Or does that do something else?

Ghrehh: Ooof, yeah. Good shout.

aptkingston (approved):

Issue #9654, opened 2023-02-10
[BUDI-6545] Many to many relationship broken?  


MaxBernstetter:
Budibase 2.3.5 (self-hosted)
I have 3 tables in a psql- datasource

* managers (pk mail -> text)
* positions (pk id -> integer)
* relation_manager_position
* 
  * managermail (text)
* 
  * positionid (integer)

I want to define a many to many relationship but  cannot do so

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/dd47cc21-84a3-4bfc-8e91-1cbfdebfaac5/7ab564a6-7fe7-4b9e-acd9-fe63d43f5509)

<sub>[BUDI-6545](https://linear.app/budibase/issue/BUDI-6545/many-to-many-relationship-broken)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6545](https://linear.app/budibase/issue/BUDI-6545/many-to-many-relationship-broken)</sub>

oscargws:
Having trouble with the same issue here, following

shogunpurple:
This one should now be resolved. Closing

PR #9689, opened 2023-02-14
Fix for SQL many-to-many configuration via relationship modal

mike12345567:
## Description
Fix for #9654 - the error checking was looking at the through fields flipped, which caused many to many relationships to throw an error. Also I've updated the dev compose file on master so that switching between develop and master is smoother.

I've also included the SQL that was used to exhibit this error as it is an example of something that we haven't tested before now.

Addresses: 
- #9654


shogunpurple (approved):
LGTM! Thanks for the quick fix.

PR #9698, opened 2023-02-15
Null safety

melohagan:
## Description
`TypeError: Cannot read properties of undefined (reading 'name')`. Should be fixed with the null-safety. The *definition* in other parts of the file were already using this null check.

Addresses: 
- https://github.com/Budibase/budibase/discussions/9365

## Screenshots
image


shogunpurple (approved):
LGTM!

PR #9706, opened 2023-02-15
Re-enabling nightly tests 

shogunpurple:
## Description
This PR re-enables the QA nightly tests - which will run every morning at 5AM on our private self hosted github runners. This CI job points to our QA environment (hence the self hosted runners as the network is private), so we shouldn't even need to spin up the budibase server as part of nightly which will speed startup significantly.

I've also moved and updated the discord webhook which should ping **only if the tests fail** with a breakdown of the test results.

I also removed the "searchIndex" concept here from the UI while I was at it - while unrelated to the tests, the feature no longer does anything, and it can be removed from the UI.

Addresses: 
- https://linear.app/budibase/issue/QA-2/qa-job-for-nightly-api-test


MihailHadzhiev2022 (approved):

Mitch-Budibase (approved):

Pedro-Revez-Silva (approved):
LGTM

PR #9725, opened 2023-02-17
Fix tenant creation on user save

Rory-Powell:
## Description
Fix a bug that prevents
- platform user doc creation
- tenants doc update
on admin user creation. 

The tests do start to fail when this is re-added however. Since the tests are currently disabled I'd like to add this to fix the integrity of the DB in qa - when the following is merged to `develop` the tests should work again: https://github.com/Budibase/budibase/pull/9670


PClmnt (approved):
LGTM. Nice fix

Issue #9369, opened 2023-01-17
[BUD-3240] Backups UX Improvements


shogunpurple:
**COS**
- Add polling to backup screen for live updates
- Remove backup "name" from rows and the concept of naming backups altogether
- Add friendly dates - instead of the raw date use time ago
- Remove "Scheduled backup" from options
- Replace "Name" with users avatar and tooltip in table
- Remove names from manual restores - not required

<sub>[BUD-3240](https://linear.app/budibase/issue/BUD-3240/backups-ux-improvements)</sub>

PR #9726, opened 2023-02-17
Backups UI Changes

Ghrehh:
## Description
- Removes the name field for manual backups and related UI.
- Adds more easily readable dates
- Removes "scheduled backups" from the filter options, which I assume is what was meant by the ticket.
- Not 100% sure what is meant by adding a user column in place of the removed "name" column, as this already exists.
- Adds functionality to poll in progress backups.

Budibase Pro Changes that are required for this to work: https://github.com/Budibase/budibase-pro/pull/112

Addresses: 
[Backups UI Changes](https://github.com/Budibase/budibase/issues/9369)


aptkingston (commented):
Nice work! I think the ticket wasn't properly specced, but the idea was to remove the concept of names altogether from backups, although from what I can tell the only remnant left is when you restore a backup, it asks you to name the pre-restore backup. So we can just remove this too:
image

Again the ticket wasn't properly worded but for the user avatar instead of name, it means replace the users email address with an avatar circle with their initials, similar to how we do it in the users section. That component is hopefully reusable.
image

Everything you've done so far looks spot on though!

One other thing I spotted was that we attempt to reference name when deleting a backup, so it currently shows as `undefined`:
image

So we just need to update that text.

Ghrehh:
@aptkingston Fixed the two bugs relating to names you pointed out, as well as removing the user's email address and replacing it with an avatar. I also made the avatar display the email address on hover.

Screenshot 2023-02-21 at 15 26 02

aptkingston (approved):

shogunpurple:
My bad folks - there was a design made up for this that wasn't attached to the ticket. Apologies for the confusion

Ghrehh:
@shogunpurple np, does how it currently looks in the above screenshot look okay? Or should we make additional changes?

shogunpurple:
Nope, LGTM. Nice work @Ghrehh

Issue #9327, opened 2023-01-12
Improve add screen modal to make autogenerated screen clearer


AndyWhann:
## Problem 

Autogenerated screens were updated to only one screen with a table block, so it's no longer "Screens" 

The side panel is also still difficult to discover because it's part of the table block

## Solution

As a quick win we want to improve the create screen modal to give more information to the user about the autogenerated screen.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/e554bb88-e386-4473-ab88-eb9d70143382)

https://xd.adobe.com/view/6850f509-141c-43bc-8ad8-9fb9608e5754-3d03/

PR #9759, opened 2023-02-21
Improve Add Screen Modal

Ghrehh:
## Description
Updates the new screen modal for a better user experience, as outlined in the ticket

Addresses: 
https://github.com/Budibase/budibase/issues/9327


packages/builder/src/pages/builder/app/[application]/design/[screenId]/screens/_components/NewScreenModal.svelte
```diff
@@ -23,71 +25,84 @@
...
           style="color: var(--spectrum-global-color-green-600); float: right"
         >
-          {#if selectedScreenMode == blankScreenModeKey}
-            <div class="checkmark-spacing">
-              <Icon size="S" name="CheckmarkCircle" />
-            </div>
-          {/if}
+          <div
+            class={`checkmark-spacing ${
+              selectedScreenMode == blankScreenModeKey ? "visible" : ""
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/screens/_components/NewScreenModal.svelte:54]: You don't need to change this, but just a FYI that the usual (and quite nice) way to conditionally apply classes like this in svelte is:
```html
<div
  class="checkmark-spacing"
  class:visible={selectedScreenMode == blankScreenModeKey}
  ....others
/>
```
Keeps it readable and works perfectly for multiple classes.
NAB but just letting you know for the future!

aptkingston (approved):
Code changes look good! It would be great to get a few screenshots for PR's like this that are heavily UI focused, just to help review it, but all good in this case since it's reasonable small anyway.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/screens/_components/NewScreenModal.svelte
Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/screens/_components/NewScreenModal.svelte:54]: Ah, wasn't aware you could combine the two, thanks for the heads up.

PR #9777, opened 2023-02-22
Bug - budi-6619 - Cannot update relationships in postgresql many to one

adrinr:
## Description
Fixing issue when editing many to one and many to many relationships on a postgres table

Addresses: 
- https://linear.app/budibase/issue/BUDI-6619/cannot-update-relationships-in-postgresql-many-one

## App Export
[test 6619-export-1677147062653.tar.gz](https://github.com/Budibase/budibase/files/10812436/test.6619-export-1677147062653.tar.gz)

## Screenshots
The screenshots are using the following schema:
```
Owner-pet: one to many
Owner-pet_toy: many to many
Owner-location: many to one
```

1. Creating a row without relationships:

https://user-images.githubusercontent.com/15987277/220877017-5507ee2c-d9a8-43a5-a933-be575c9b2d33.mov

2. Creating a row with all the relationships populated:

https://user-images.githubusercontent.com/15987277/220877159-c22ab8d5-e16c-4f8c-a59a-bd5004182bb6.mov

3. Populating a row relationships

https://user-images.githubusercontent.com/15987277/220877331-7a1698e5-7874-45b5-8a62-f11e47b4c570.mov

4. Modifying relationships

https://user-images.githubusercontent.com/15987277/220877788-371103b9-9d6b-48f3-b1bd-da3ad07bbc17.mov

## Documentation
- [X] I have reviewed the budibase documentatation to verify if this feature requires any changes. If changes or new docs are required I have written them.


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -379,6 +384,15 @@ export class ExternalRequest {
       ) {
         continue
       }
+
+      if (
```
adrinr, line 388: This is ensuring we don't map wrong fields when both primary and foreign tables have the same field names

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -572,11 +586,18 @@ export class ExternalRequest {
       if (!linkTable || !linkPrimary) {
         return
       }
+
+      const linkSecondary =
```
adrinr, line 590: This is solving an issue that occurred when reassigning many-to-many relationships

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -27,7 +27,9 @@ describe("row api - postgres", () => {
   let makeRequest: MakeRequestResponse,
```
adrinr, line 27: This test file became really big, with the `beforeEach` actions hard to follow. I will work in a refactor to use better helpers to make it more verbose

.tool-versions
```diff
@@ -1,2 +1,2 @@
 nodejs 14.19.3
-python 3.11.1
\ No newline at end of file
+python 3.10.0
```
melohagan, line 2: Just double checking this change was intentional?

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -293,7 +302,7 @@ export class ExternalRequest {
         // we're not inserting a doc, will be a bunch of update calls
         const otherKey: string = field.throughFrom || linkTablePrimary
         const thisKey: string = field.throughTo || tablePrimary
-        row[key].map((relationship: any) => {
```
melohagan, line 296: Nice one! I thought that map was odd.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -379,7 +388,8 @@ export class ExternalRequest {
       ) {
         continue
       }
-      let linked = basicProcessing(row, linkedTable)
+
+      let linked = basicProcessing(row, linkedTable, true)
```
melohagan, line 392: Can we change this to an object instead? You can't see what the parameter name is.

I.e. 
```js
{
  row,
  linkedTable,
  isLinked: true
}
```

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -572,11 +582,18 @@ export class ExternalRequest {
       if (!linkTable || !linkPrimary) {
         return
       }
+
+      const linkSecondary =
+        linkTable?.primary &&
```
melohagan, line 587: I don't think you need the `linkedTable?.primary` check if you have `linkedTable?.primary?.length > 1`

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -572,11 +582,18 @@ export class ExternalRequest {
...
+        linkTable?.primary?.length > 1 &&
+        linkTable?.primary[1]
+
       const rows = related[key]?.rows || []
       const found = rows.find(
```
melohagan, line 592: I'm struggling to understand this section of the code. Can the variable name be improved or some of this moved into a function? I'm not sure what has been "found".

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -629,10 +646,7 @@ export class ExternalRequest {
    * Creating the specific list of fields that we desire, and excluding the ones that are no use to us
    * is more performant and has the added benefit of protecting against this scenario.
    */
-  buildFields(
-    table: Table,
-    includeRelations: IncludeRelationship = IncludeRelationship.INCLUDE
-  ) {
+  buildFields(table: Table, includeRelations: boolean) {
```
melohagan, line 649: Any reason this is a boolean now?

packages/server/src/api/controllers/row/external.ts
```diff
@@ -58,7 +58,7 @@ export async function patch(ctx: BBContext) {
   return handleRequest(Operation.UPDATE, tableId, {
     id: breakRowIdField(id),
     row: inputs,
-    includeSqlRelationships: IncludeRelationship.EXCLUDE,
+    includeSqlRelationships: IncludeRelationship.INCLUDE,
```
melohagan, line 61: Why has this changed?

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -27,7 +27,9 @@ describe("row api - postgres", () => {
...
     postgresDatasource: Datasource,
     primaryPostgresTable: Table,
-    auxPostgresTable: Table
+    o2mInfo: ForeignTableInfo,
```
melohagan, line 30: NAB: `o2m` is a bit odd. Personally I'd prefer `oneToManyInfo`

Same applied to `m2m`, etc

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -151,30 +197,105 @@ describe("row api - postgres", () => {
     value: number
   }
 
+  type ForeignTableInfo = {
+    table: Table
+    fieldName: string
+    relationshipType: RelationshipTypes
```
melohagan, line 203: Property is singular but the type is plural - which one is it?

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -198,7 +319,9 @@ describe("row api - postgres", () => {
   async function populatePrimaryRows(
     count: number,
     opts?: {
-      createForeignRow?: boolean
+      createOne2Many?: boolean
```
melohagan, line 322: NAB: Replace '2' with 'To'

melohagan (changes requested):
Pulled down the code and saw that the bug appears to be fixed. Thanks for that! :)

I've left a few small comments, but otherwise LGTM pending a look over from @mike12345567

packages/server/src/integration-test/postgres.spec.ts
mike12345567: This enumeration needs refactored - enums should all be singular (just an old one that wasn't that way). `RelationshipTypes` should really just be `RelationshipType`.

packages/server/src/integration-test/postgres.spec.ts
mike12345567: Would agree - more verbose variable names are preferred.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -142,7 +142,11 @@ function cleanupConfig(config: RunConfig, table: Table): RunConfig {
   return config
 }
 
-function generateIdForRow(row: Row | undefined, table: Table): string {
+function generateIdForRow(
```
mike12345567, line 145: A personal preference around boolean parameters, it would be better if it was like
```
generateIdForRow(row: Row | undefined, table: Table, opts?: { isLinked?: boolean })
```

Idea being that boolean parameters are confusing when looking at a call of the function, e.g.
`generateIdForRow(row, table, true)` - its not obvious immediately the purpose of the bool parameter, vs `generateIdForRow(row, table { isLinked: true })`

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -174,18 +181,20 @@ function getEndpoint(tableId: string | undefined, operation: string) {
   }
 }
 
-function basicProcessing(row: Row, table: Table): Row {
+function basicProcessing(row: Row, table: Table, isLinked: boolean): Row {
```
mike12345567, line 184: Same boolean parameter comment.

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: Sorry - didn't see this comment, I made a similar comment further up.

packages/server/src/integrations/base/sql.ts
```diff
@@ -415,9 +415,7 @@ class InternalBuilder {
     if (opts.disableReturning) {
       return query.insert(parsedBody)
     } else {
-      return query
-        .insert(parsedBody)
-        .returning(generateSelectStatement(json, knex))
```
mike12345567, line 420: Is this safe? I remember there being a valid reason for using a select statement rather than the total `*` - in theory there will never be any joined data here so we can't have column overlap, but selecting the columns the same way we do when we are actually fetching data could be safer.

mike12345567 (changes requested):
A few minor comments around use of boolean parameters and the preferred use of object specifiers, only other thing is about the change to the `returning` syntax in the SQL generation and any risks that might be attached to it, although it really only applies to Postgres - otherwise looks good, just worth addressing @melohagan other comments.

.tool-versions
adrinr: Yes and no. For some reason, some dependency fails with 3.11. Not really related to this PR change, I might add it in a different PR to make it clear

packages/server/src/integrations/base/sql.ts
melohagan: This is actually the change required to fix the bug. It's just reverting the code to what was previously there: https://github.com/Budibase/budibase/commit/14beca3dde33114a0c87876229596bb5e60f5e56#diff-4c4ab33e19e10728e2cad3fa664f4f4af0b02e2c5bae0eb6eecc4d1421edc13c

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: Yup, I totally agree with this kind of best practices. I just did want to touch as little things as possible on the code, as I don't want to make a bug hard to PR. But happy to change it :)

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: You need, if not you have `undefined > 1`, and ts will complain

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: Changed, I agree with this approach. I did not want to add many changes on the code and in a lot of places we don't use this approach. But I think we all agree is probably the best approach

mike12345567: If there is anywhere you see this approach we should mark them with a comment or something like that and change it - it isn't a great approach and I know there are some old hold outs which could do with being converted.

packages/server/src/integrations/base/sql.ts
mike12345567: Interesting this caused the issue - what is the exact problem occurring here?

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: I think it's much more verbose. The IncludeRelationships is just an enum, and getting a field `IncludeRelationships` I would expect either true or false. I even found some possible bug where the code was checking `if(includeRelationships)`, treating it as a boolean

packages/server/src/api/controllers/row/ExternalRequest.ts
melohagan: Ah it's TypeScript not JavaScript!

packages/server/src/api/controllers/row/external.ts
adrinr: This is one of the causes of the bug. Because we were not fetching the SQL relationships, any edition request was trying to recreate them, given that if could "not find any existing relationship" matching the incoming request

packages/server/src/integration-test/postgres.spec.ts
adrinr: Always the fight between verbosity and not ultra-long variables. But I'll change it :)

packages/server/src/api/controllers/row/external.ts
melohagan: Makes sense, thanks!

packages/server/src/integration-test/postgres.spec.ts
adrinr: I will not change this here, as this enum is used in a lot of places, svelte included

packages/server/src/integrations/base/sql.ts
adrinr: We changed this because this was causing a small bug, returning unexpected relationships in returns when the relationship was null and the both primary column name of the foreign table was being used for the primary table as well. It was a small bug that affected the return of `patches` and `inserts`, probably never affected anybody as it's not common to listen to these responses. But this causes the UI to fail in some cases.
All the changes on this PR are to be able to get it working in both scenarios and getting sure that all type of relationships are supported

mike12345567: Thanks for the explanation!

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -174,18 +181,20 @@ function getEndpoint(tableId: string | undefined, operation: string) {
...
   const thisRow: Row = {}
   // filter the row down to what is actually the row (not joined)
-  for (let fieldName of Object.keys(table.schema)) {
+  for (let field of Object.values(table.schema)) {
+    const fieldName = field.name
+
     const pathValue = row[`${table.name}.${fieldName}`]
-    const value = pathValue != null ? pathValue : row[fieldName]
+    const value = pathValue != null || isLinked ? pathValue : row[fieldName]
```
Rory-Powell, line 199: I find the ternary in combination with the `||` quite difficult to read. Could we extract this logic to `if/else` instead?

Question: Is this the same as the logic from `generateRowId`? If so, we could extract out to a shared function?

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: Just one minor comment about this - I understand why boolean might be preferred to enum here, but can we use an object specifier for it as well? There might be some older code that doesn't use object specifiers, but we should strive to remove as many of them as possible.

mike12345567 (approved):
Just one minor changed needed, then LGTM!

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -572,14 +593,37 @@ export class ExternalRequest {
       if (!linkTable || !linkPrimary) {
         return
       }
+
+      // @ts-ignore
```
melohagan, line 597: Sorry just noticed this. Best not to use a ts-ignore. (It's caused issues before!)

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -572,14 +593,37 @@ export class ExternalRequest {
...
+        const matchesSecondayLink = row[linkSecondary] === body?.[linkSecondary]
+        return matchesPrimaryLink && matchesSecondayLink
+      }
+
+      const existingRelationship = rows.find((row: { [key: string]: any }) =>
+        relationshipMatchPredicate({ row, linkPrimary, linkSecondary })
       )
       const operation = isUpdate ? Operation.UPDATE : Operation.CREATE
-      if (!found) {
+      if (!existingRelationship) {
```
melohagan, line 626: This helps a lot - thank you! :)

melohagan (approved):
LGTM, just one last comment about a `ts-ignore` I spotted.

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: I just added this to relace the old `if(primary &...)`. I don't like using the @ts-ignore, but this otherwise we need either the full check or disabling this https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess.
In javascript `['a','b'][4]` will return undefined, so this is what we would expect. This `@ts-ignore` pattern is being used multiple times in this file, I just replicate it to match this use case. This is a problem with the typings that would need a better understanding of these keys. I don't think we can have a relationship without the primary, but currently the types support it :(

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: Changed! Good spot

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: Anyways, I just cleaned the one I introduced

Rory-Powell (commented):
Thanks for making that change. I'll defer to the other approvals here as I'm not very familiar with this area of the code.

Rory-Powell:
https://github.com/ts-ignore lol 
Screenshot 2023-02-24 at 10 31 31

PR #9779, opened 2023-02-22
metrics api endpoint

jonnymccullagh:
## Description
This PR adds a /api/public/v1/metrics endpoint that can be scraped by Prometheus or other tools supporting OpenMetrics.
A getAllUserIds function was added to the backend core to support counting the number of users.

## Screenshots
Selection_016

## Documentation
Docs have been added to readme at https://docs.budibase.com/docs/metrics


packages/server/src/api/routes/public/index.ts
```diff
@@ -120,6 +121,7 @@ function applyRoutes(
 }
 
 applyRoutes(appEndpoints, PermissionType.APP, "appId")
+applyRoutes(metricEndpoints, PermissionType.APP, "appId")
```
Rory-Powell, line 124: Do we need the `PermissionType.APP, "appId"` here?  There's no `appId` in the `"/metrics"` path. 
We should lock the new endpoint down as an builder or admin only operation

packages/server/src/api/routes/public/metrics.ts
```diff
@@ -0,0 +1,23 @@
...
+ *     description: Output metrics in OpenMetrics format compatible with Prometheus
+ *     tags:
+ *       - metrics
+ *     responses:
+ *       200:
+ *         description: Returns tenant metrics.
+ *         content:
+ *           text/plain:
+ */
+read.push(new Endpoint("get", "/metrics", controller.fetch))
```
Rory-Powell, line 21: We should get corresponding unit test in for this new api

Rory-Powell (commented):
Can we unpublish this for now: https://docs.budibase.com/docs/metrics

packages/server/src/api/routes/tests/metrics.spec.js
```diff
@@ -0,0 +1,49 @@
+const { roles, utils } = require("@budibase/backend-core")
+const { checkBuilderEndpoint } = require("./utilities/TestFunctions")
+const setup = require("./utilities")
+const { BUILTIN_ROLE_IDS } = roles
+
+jest.setTimeout(30000)
+
+// jest.mock("../../../utilities/workerRequests", () => ({
```
mike12345567, line 8: Can remove commented out blocks.

mike12345567 (approved):

Issue #9658, opened 2023-02-11
[BUDI-6547] Console error - add bindings to 'Execute Query' action 


s6dediss:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method:  docker compose
  - Budibase Version: 2.3.10
  - App Version: 2.3.10

**Describe the bug**
I can't add a binding an 'Execute Query' action. When i try to save any binding to any property i got an error in the console and i can't save the binding.

**To Reproduce**
1. Add any 'Execute Query' action to a button 
2. Try to add a binding for this action

**Expected behavior**
The binding should be saved.

**Screenshots**
![consoloError](https://user-images.githubusercontent.com/97907418/218250016-1008db75-2a79-4d23-b571-b71293f52e10.png)

**Desktop (please complete the following information):**
 - OS: Linux Ubuntu
 - Browser chrome

<sub>[BUDI-6547](https://linear.app/budibase/issue/BUDI-6547/console-error-add-bindings-to-execute-query-action)</sub>

PR #9782, opened 2023-02-23
Fix error being thrown when attempting to change query parameters

aptkingston:
## Description
Fixes an error being thrown when attempting to change parameters for a query using the execute query button action.

Addresses: 
- #9658


shogunpurple (approved):
LGTM

PR #9785, opened 2023-02-23
Configs /  Server tests / Failing real email tests updates 

Rory-Powell:
## Description
- Started some of this while updating public settings config for enforced sso - breaking into own changeset 

Configs
- Move all config (google, oidc, settings, smtp etc) reading and writing out of utils and into a dedicated module
- Add type support for all functions
- Consolidate defaulting logic for when config wasn't available which was done in multiple places
- Remove circular dependencies between analytics and configs
- Update usages inside budibase with new configs module
- Removes a lot of direct access to context and db that has now been abstracted away 
- Add new types to configs highlighted by refactor
- Add api types for oidc and settings responses

Server
- Remove default tenant id from server tests to increase reliability and reduce conflicts

Email tests
- An error was being swallowed causing the tests to never run properly
- Update the test with new credentials that work + integrate with configs module

Reduces circular deps from 21
image

to 16:
Screenshot 2023-02-24 at 08 16 13

### DIFF
--- a/packages/backend-core/src/auth/auth.ts
+++ b/packages/backend-core/src/auth/auth.ts
@@ -2,25 +2,33 @@ const _passport = require("koa-passport")
 const LocalStrategy = require("passport-local").Strategy
 const JwtStrategy = require("passport-jwt").Strategy
 import { getGlobalDB } from "../context"
-const refresh = require("passport-oauth2-refresh")
-import { Config, Cookie } from "../constants"
-import { getScopedConfig } from "../db"
+import { Cookie } from "../constants"
 import { getSessionsForUser, invalidateSessions } from "../security/sessions"
 import {
+  authenticated,
+  csrf,
+  google,
   jwt as jwtPassport,
   local,
-  authenticated,
-  tenancy,
-  csrf,
   oidc,
-  google,
+  tenancy,
 } from "../middleware"
+import * as userCache from "../cache/user"
 import { invalidateUser } from "../cache/user"
-import { PlatformLogoutOpts, User } from "@budibase/types"
+import {
+  ConfigType,
+  GoogleInnerConfig,
+  OIDCInnerConfig,
+  PlatformLogoutOpts,
+  User,
+} from "@budibase/types"
 import { logAlert } from "../logging"
 import * as events from "../events"
-import * as userCache from "../cache/user"
+import * as configs from "../configs"
 import { clearCookie, getCookie } from "../utils"
+import { ssoSaveUserNoOp } from "../middleware/passport/sso/sso"
+
+const refresh = require("passport-oauth2-refresh")
 export {
   auditLog,
   authError,
@@ -33,7 +41,6 @@ export {
   google,
   oidc,
 } from "../middleware"
-import { ssoSaveUserNoOp } from "../middleware/passport/sso/sso"
 export const buildAuthMiddleware = authenticated
 export const buildTenancyMiddleware = tenancy
 export const buildCsrfMiddleware = csrf
@@ -63,11 +70,10 @@ _passport.deserializeUser(async (user: User, done: any) => {
 })
 
 async function refreshOIDCAccessToken(
-  db: any,
-  chosenConfig: any,
+  chosenConfig: OIDCInnerConfig,
   refreshToken: string
-) {
-  const callbackUrl = await oidc.getCallbackUrl(db, chosenConfig)
+): Promise<RefreshResponse> {
+  const callbackUrl = await oidc.getCallbackUrl()
   let enrichedConfig: any
   let strategy: any
 
@@ -90,7 +96,7 @@ async function refreshOIDCAccessToken(
 
   return new Promise(resolve => {
     refresh.requestNewAccessToken(
-      Config.OIDC,
+      ConfigType.OIDC,
       refreshToken,
       (err: any, accessToken: string, refreshToken: any, params: any) => {
         resolve({ err, accessToken, refreshToken, params })
@@ -100,11 +106,10 @@ async function refreshOIDCAccessToken(
 }
 
 async function refreshGoogleAccessToken(
-  db: any,
-  config: any,
+  config: GoogleInnerConfig,
   refreshToken: any
-) {
-  let callbackUrl = await google.getCallbackUrl(db, config)
+): Promise<RefreshResponse> {
+  let callbackUrl = await google.getCallbackUrl(config)
 
   let strategy
   try {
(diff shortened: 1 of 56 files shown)


packages/backend-core/src/configs/configs.ts
```diff
@@ -0,0 +1,224 @@
...
+  opts: { tenantAware: boolean; config?: SettingsInnerConfig } = {
+    tenantAware: true,
+  }
+) {
+  let platformUrl = env.PLATFORM_URL || "http://localhost:10000"
+
+  if (!env.SELF_HOSTED && env.MULTI_TENANCY && opts.tenantAware) {
+    // cloud and multi tenant - add the tenant to the default platform url
+    const tenantId = context.getTenantId()
+    if (!platformUrl.includes("localhost:")) {
```
adrinr, line 89: Could not this be part of the URL set on config, instead of a check in here? It sounds strange to have "dev" code checks in production

packages/backend-core/src/configs/configs.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  return platformUrl
+}
+
+export const analyticsEnabled = async (opts?: {
+  config?: SettingsInnerConfig
+}) => {
+  // cloud - always use the environment variable
```
adrinr, line 108: Why don't we use the same mechanism (either env or config) in both cases?

packages/backend-core/src/middleware/authenticated.ts
```diff
@@ -115,7 +115,8 @@ export default function (
           authenticated = true
         } catch (err: any) {
           authenticated = false
-          console.error("Auth Error", err?.message || err)
+          console.error(`Auth Error: ${err.message}`)
+          console.error(err)
```
adrinr, line 119: Do we need this second error log?

packages/types/src/api/web/global/configs.ts
```diff
@@ -0,0 +1,23 @@
+import { SettingsConfig, SettingsInnerConfig } from "../../../documents"
+
+/**
+ * Settings that aren't stored in the database - enriched at runtime.
+ */
+export interface PublicSettingsInnerConfig extends SettingsInnerConfig {
+  google: boolean
+  oidc: boolean
```
adrinr, line 8: Will we ever have google integrations without oidc?

packages/types/src/documents/global/config.ts
```diff
@@ -5,32 +5,42 @@ export interface Config extends Document {
...
-    analyticsEnabled?: boolean
-  }
+  config: SettingsInnerConfig
 }
 
 export interface GoogleInnerConfig {
   clientID: string
   clientSecret: string
   activated: boolean
+  // deprecated / read only
```
adrinr, line 42:
```suggestion
  /**
   * @deprecated read only
   */
```

adrinr: Using the `@deprecated` attribute will highlight usages on GUIs

packages/worker/src/tests/structures/configs.ts
```diff
@@ -55,8 +55,8 @@ export function smtpEthereal() {
       host: "smtp.ethereal.email",
       secure: false,
       auth: {
-        user: "don.bahringer@ethereal.email",
-        pass: "yCKSH8rWyUPbnhGYk9",
+        user: "wyatt.zulauf29@ethereal.email",
```
adrinr, line 58: Can we use randomised values?

packages/worker/src/utilities/email.ts
```diff
@@ -36,24 +36,24 @@ const FULL_EMAIL_PURPOSES = [
   EmailTemplatePurpose.CUSTOM,
 ]
 
-function createSMTPTransport(config: any) {
+function createSMTPTransport(config?: SMTPInnerConfig) {
   let options: any
-  let secure = config.secure
+  let secure = config?.secure
   // default it if not specified
   if (secure == null) {
```
adrinr, line 43: Should this handle undefineds too now?

Rory-Powell: the `==` will handle null and undefined here

packages/backend-core/src/middleware/authenticated.ts
Rory-Powell: If we want to capture the stack trace yes, I was seeing log messages with just
```
"Auth Error": 
```

packages/backend-core/src/configs/configs.ts
Rory-Powell: In self host we want the config stored in the db to take precedence. The rules being

- Have a config in the db? 
   - yes - use the value from the config
   - no - continue
- Env var is explicitly set to false?
   - yes - return false
   - no -  return true (on by default)

Rory-Powell: Self host
Screenshot 2023-02-27 at 11 27 06

Cloud
Screenshot 2023-02-27 at 11 27 20

packages/backend-core/src/configs/configs.ts
Rory-Powell: The platform url has two forms
- system - the base platform url, no tenancy information - just where dns will resolve to. this never changes
- tenant aware - in a multi tenant environment, this url is dynamic based on the tenant being accessed. we enrich this here

We are able to mimic the production url format locally, see: 
https://github.com/Budibase/account-portal/blob/develop/docs/local_development/url_tenancy.md

We want to perform perform the tenant enriching in production and in development environments (when the above url tenancy is active). The only distinguishing factor that we have for this right now is `localhost`. Unless we were to add another flag for `ENABLE_PLATFORM_URL_TENANT_ENRICHMENT` instead, set to false in local development, unless url tenancy is being tested. We could do that, but I'd prefer not to change it right now as this was a copy paste from another location. 

Another thing we could do is default the value of `PLATFORM_URL` in the core env file instead, which is the pattern we are tending towards instead of doing it inside application logic. There are only a few usages of the env field: 
Screenshot 2023-02-27 at 11 45 10
The rest of the application uses the value of this helper function. 

If you think it's worth it, could raise a follow up task to address both of these

packages/backend-core/src/auth/auth.ts
```diff
@@ -133,41 +138,37 @@ async function refreshGoogleAccessToken(
...
-    refreshResponse = await refreshGoogleAccessToken(
-      db,
-      chosenConfig,
-      refreshToken
-    )
+    return refreshGoogleAccessToken(config, refreshToken)
   }
 
-  return refreshResponse
+  throw new Error(`Unsupported configType=${configType}`)
```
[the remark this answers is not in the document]
Rory-Powell: Good spot, the incorrect enum was actually being used here. We want `SSOProviderType` however `ConfigType` was used - they happen to overlap so this worked. 

Added to the other pr that is based off this one here: https://github.com/Budibase/budibase/pull/9787/commits/aeb02102419134f8e83da1fbc4610e5b299d0254

packages/types/src/documents/global/config.ts
Rory-Powell: Nice one

PR #9787, opened 2023-02-23
Enforceable SSO

Rory-Powell:
## Description
- Add the option to auth config for disabling password login (business and enterprise)
- Remove the password login fields when the setting is enabled
- Prevent password actions at the API level (login, email reset, in-app reset) when the setting is enabled
   - This extends the existing behaviour for SSO users to all users 
- Disable all of the above if:
   - `ENABLE_SSO_MAINTENANCE_MODE` is set
   - License does not have the required feature

Changes:

### Backend-core
- Move maintenance mode to core env file for use in pro package

### Types
- Add `isSSOEnforced` to settings config
- Add new feature `enforceableSSO`

### Worker
- Rename `isPreventSSOPasswords` to `isPreventPasswordActions` 
   - Integrate is enforced check   
- Always set `isSSOEnforced` on read of settings config  
   - Facilitates downgrades and maintenance mode flag

Related PRs:
- https://github.com/Budibase/budibase-pro/pull/117

Addresses: 
- https://linear.app/budibase/issue/BUDI-4821/enforceable-sso-on-business-and-enterprise

### Auth Settings

Free plan:
Screenshot 2023-02-23 at 16 13 55

Business or Enterprise plan:
Screenshot 2023-02-23 at 16 15 28

### Login Page

With enforced sso:

Screenshot 2023-02-23 at 16 18 03

## Documentation


.github/workflows/budibase_ci.yml
```diff
@@ -10,7 +10,8 @@ on:
  pull_request:
     branches:
       - master
-      - develop 
+      - develop
+      - configs-refactor-and-server-test-fixes
```
adrinr, line 14: This should be removed before merging

packages/backend-core/src/environment.ts
```diff
@@ -91,6 +93,14 @@ const environment = {
...
   SMTP_PORT: parseInt(process.env.SMTP_PORT || ""),
   SMTP_FROM_ADDRESS: process.env.SMTP_FROM_ADDRESS,
+  /**
+   * Enable to allow an admin user to login using a password.
+   * This can be useful to prevent lockout when configuring SSO.
+   * However, this should be turned OFF by default for security purposes.
+   */
+  ENABLE_SSO_MAINTENANCE_MODE: selfHosted
+    ? process.env.ENABLE_SSO_MAINTENANCE_MODE
+    : false,
```
adrinr, line 103: What about cloud users? Will this never be needed?

packages/worker/src/api/controllers/global/auth.ts
```diff
@@ -61,8 +61,8 @@ export const login = async (ctx: Ctx<LoginRequest>, next: any) => {
   const email = ctx.request.body.username
 
   const user = await userSdk.getUserByEmail(email)
-  if (user && (await userSdk.isPreventSSOPasswords(user))) {
-    ctx.throw(400, "SSO user cannot login using password")
+  if (user && (await userSdk.isPreventPasswordActions(user))) {
+    ctx.throw(400, "Password login is disabled for this user")
```
adrinr, line 65: Didn't we say we wanted to have a more generic text that did not mention sso?

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -126,10 +127,10 @@ export async function save(ctx: UserCtx<Config>) {
   const existingConfig = await configs.getConfig(type)
   let eventFns = await getEventFns(ctx.request.body, existingConfig)
 
-  // Config does not exist yet
-  if (!existingConfig) {
-    body._id = configs.generateConfigID(type)
+  if (existingConfig) {
+    body._rev = existingConfig._rev
```
adrinr, line 131: What is the use case of this? If the `_rev` value is used to avoid unexpected overrides while multiple processes/users working in parallel, will this not invalidate this safety check?

packages/worker/src/api/controllers/global/auth.ts
Rory-Powell: Yep I'll get these updated today

PR #9793, opened 2023-02-24
Remove deploy to release env

Rory-Powell:
## Description
- This env is no longer used 
- This job has been failing


shogunpurple (approved):
LGTM! Looks like we can finally bring this env down

Issue #6466, opened 2022-06-26
Audit Logs: Backend


shogunpurple:
Spec: https://budibase.atlassian.net/wiki/spaces/PROD/pages/12615698/Audit+Logs+Enterprise

When I'm viewing the budibase portal as an admin user, I want an interface to view the audit logs for my budibase instance, so that I can track everything that happens within my installation.

**COS**
- Add admin-only audit logs tab to portal
- Allow audit log export
- Users should be able to filter their audit logs by appId, timestamp, event name user email, and user ID
- Enhance event/analytics system to write events to audit log if enabled for user
- All audit logs must have a timestamp, user and event attached to them
- The process that handles the writing of audit logs must be monitored for health. 

**Designs**
https://www.figma.com/file/JeAMNqn0zhnXZh55MKklQq/Audit-logs?node-id=0%3A1

Issue #6465, opened 2022-06-26
Audit Logs: UI


shogunpurple:
Spec: https://budibase.atlassian.net/wiki/spaces/PROD/pages/12615698/Audit+Logs+Enterprise

When I'm viewing the budibase portal as an admin user, I want an interface to view the audit logs for my budibase instance, so that I can track everything that happens within my installation.

**COS**
- Add admin-only audit logs tab to portal
- Allow audit log export
- Table view and drilldown panel for viewing raw logs
- If an event cannot be tied to an individual user, then we mark the events user as “System”
- The user should see no real reference to app Id by default, instead seeing the app name, that will be mapped in the frontend.
- Allow users to ignore app name altogether and show the appId

**Designs**
https://www.figma.com/file/JeAMNqn0zhnXZh55MKklQq/Audit-logs?node-id=0%3A1

Issue #6464, opened 2022-06-26
Audit Logs


shogunpurple:

datapolitical:
What's the most recent update in terms of expected launch date for this?

shogunpurple:
@datapolitical Q4

PR #9795, opened 2023-02-24
Audit Logs

PClmnt:
## Description
This PR adds the Audit Logs feature, which allows Business  plan users to view and filter events that have happened on their system.

Addresses: 
- #6465 
- #6466 
- #6464 

### Main Audit logs screen
image

The entrypoint to Audit logs accessed via Account then Audit logs. This table is paginated so it only retrieves the next set of results when you navigate to the next page

### Side panel
image

Side panel with metadata. This is opened on clicking `Details` or the row in the table. 

### Filters
image

There are a number of different filters that can be applied to the table, and filters are applied in the order that they are selected
#### User

Filter by a single or multiple users. Use the search box within the dropdown to search through all users on the system

#### App

Filter by a single or multiple apps. All apps available on the system will be displayed in the dropdown and can be searched

 #### Event
 Filter by single or multiple events. 

 #### Date Range
 You can filter by any date range. The default is set to 30 days, but any range is permissible. Selecting one date will display all results from then until the current date. 

 #### Search
 The search input allows a full text search on all of the properties within an audit log. For example, you could search for an App ID such as `app_dev_uuid1` and any audit logs that have that as a value within any property will be returned. Or you could search for `BASIC` and that would return Screens that were created with that Role. 

## Documentation
- [X] I have reviewed the budibase documentatation to verify if this feature requires any changes. If changes or new docs are required I have written them.


packages/backend-core/src/events/events.ts
```diff
@@ -15,6 +45,22 @@ export const publishEvent = async (
   // no backfill - send the event and exit
   if (!backfilling) {
     await processors.processEvent(event, identity, properties, timestamp)
+    if (auditLogsEnabled && isAudited(event)) {
```
Rory-Powell, line 48: Best to break this out into an `AuditLogsProcessor` so we don't need to make any changes to the core events publishing - we can just implement a new processor and act there instead. 

We can then move the associated `isAudited`, `configure`, `auditLogsEnabled` and `auditLogQueue` fields there instead so this file is completely agnostic of audit logs.

packages/backend-core/src/middleware/downloadBody.ts
```diff
@@ -0,0 +1,22 @@
+import { Ctx } from "@budibase/types"
+
+/**
+ * Expects a standard "query" query string property which is the JSON body
+ * of the request, which has to be sent via query string due to the requirement
+ * of making an endpoint a GET request e.g. downloading a file stream.
```
Rory-Powell, line 6: There's no downloading happening here which is a bit misleading for the name of the middleware. Could we rename it some more clear, `queryToBody`? Might also want to add the restriction that this only affects `GET` requests, as there is the potential to misuse and overwrite the body of a post request.

packages/backend-core/src/users.ts
```diff
@@ -10,24 +10,52 @@ import { BulkDocsResponse, User } from "@budibase/types"
 import { getGlobalDB } from "./context"
 import * as context from "./context"
 
-export const bulkGetGlobalUsersById = async (userIds: string[]) => {
+type GetOpts = { cleanup?: boolean }
+
+function cleanupUsers(users: User | User[]) {
```
Rory-Powell, line 15:
```suggestion
function removeUserPassword(users: User | User[]) {
```
more specific?

packages/backend-core/src/users.ts
```diff
@@ -10,24 +10,52 @@ import { BulkDocsResponse, User } from "@budibase/types"
...
   const db = getGlobalDB()
-  return (
+  let users = (
     await db.allDocs({
       keys: userIds,
       include_docs: true,
     })
   ).rows.map(row => row.doc) as User[]
+  if (opts?.cleanup) {
+    users = cleanupUsers(users) as User[]
```
Rory-Powell, line 42: The way we handle user passwords is a bit scary atm, it's feels too easy to read a user doc and accidentally delete the password by writing it again to the db without going through the proper sdk. 
We also do a lot of this in various places:
```ts
delete user.password
```

We'd be nearly better moving passwords out of the main user doc altogether, into the platform doc where we only write once. Could then have dedicated functions for password read / write, more protection around misuse by decoupling password updates from user saves.

packages/backend-core/src/utils/utils.ts
```diff
@@ -214,6 +214,36 @@ export async function getBuildersCount() {
   return builders.length
 }
 
+/**
+ * Logs a user out from budibase. Re-used across account portal and builder.
+ */
+export async function platformLogout(opts: PlatformLogoutOpts) {
```
Rory-Powell, line 220: This has been moved to the `auth` package now. Must have been re-added in a merge.

packages/types/src/sdk/auditLogs.ts
```diff
@@ -0,0 +1,20 @@
...
+  timestamp?: string | number
+  userId?: string
+  hostInfo?: HostInfo
+}
+
+export type AuditLogFn = (
+  event: Event,
+  metadata: any,
+  opts: AuditWriteOpts
+) => Promise<any>
```
Rory-Powell, line 14: `any` or `void`?

packages/types/src/sdk/events/app.ts
```diff
@@ -3,50 +3,61 @@ import { BaseEvent } from "./event"
 export interface AppCreatedEvent extends BaseEvent {
   appId: string
   version: string
+  name: string
```
Rory-Powell, line 6: Similar case as emails re: PII. Doing a blanket `delete properties.name` feels more risky though. Do we need the name here in the event at all, given we're already enriching the whole app into the payload on read?

packages/types/src/sdk/events/auth.ts
```diff
@@ -7,10 +7,12 @@ export type SSOType = ConfigType.OIDC | ConfigType.GOOGLE
 export interface LoginEvent extends BaseEvent {
   userId: string
   source: LoginSource
+  email: string
```
Rory-Powell, line 10: Similar as above, we already enrich the user on read, do we need email here at all?

packages/worker/src/api/routes/global/tests/auditLogs.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  beforeAll(async () => {
+    await config.beforeAll()
+  })
+
+  afterAll(async () => {
+    await config.afterAll()
+  })
+
+  it("should be able to fire some events (create audit logs)", async () => {
```
Rory-Powell, line 26: Can we add an inner describe wrapping all the below `it` blocks:
```ts
describe("POST /api/global/auditlogs/search", () => {...})
```

packages/worker/src/tests/api/auditLogs.ts
```diff
@@ -0,0 +1,26 @@
...
+    const res = await this.request
+      .post("/api/global/auditlogs/search")
+      .send(search)
+      .set(this.config.defaultHeaders())
+      .expect("Content-Type", /json/)
+      .expect(200)
+    return res.body as SearchAuditLogsResponse
+  }
+
+  download = (search: AuditLogSearchParams) => {
```
Rory-Powell, line 20: unused?

packages/types/src/sdk/events/app.ts
Rory-Powell: Same for all the `name` additions to the other events below. We're getting into risky territory by re-using all the posthog property payload types for the audit logs. I get the intention with wanting to put real data into the audit event, it gives a more accurate view of history and prevents you needing to enrich things on read, particularly with things that have since been deleted. 

We might want to consider splitting the event types, e.g. 
```ts
// this can contain as much info as we want, can be written for audit
interface AppCreatedEvent extends BaseEvent {} 

// we must map specifically to this event in the posthog processor
interface AppCreatedAnalyticsEvent extends BaseEvent, AnalyticsEvent {}
```

e.g. in the posthog processor we can add a new integration for getting the analytics representation of an event
```ts

import * as mapping from "./mapping"

async processEvent(
    event: Event,
    identity: Identity,
    properties: BaseEvent,
    timestamp?: string | number
  ): Promise<void> {

  const properties = mapping.mapEvent(event, properties)
}

// ---- mapping.ts ---

const MAPPERS: { [key in Event]: () => AnalyticsEvent } = {
   [Event.APP_CREATED]: mapAppAnalyticsEvent
}

export function mapEvent(event: Event, properties: BaseEvent) {
   return MAPPERS[event](properties)
}

// APPS

function mapAppAnalyticsEvent(event: AppCreatedEvent): AppCreatedAnalyticsEvent {
   return {
      // pull out the specific fields required 
   }
}
```

Rory-Powell (changes requested):

packages/worker/src/tests/api/auditLogs.ts
mike12345567: Not finished testing yet!

packages/backend-core/src/context/mainContext.ts
```diff
@@ -30,14 +30,23 @@ export function getGlobalDBName(tenantId?: string) {
   return baseGlobalDBName(tenantId)
 }
 
+export function getAuditLogDBName(tenantId?: string) {
+  if (!tenantId) {
+    tenantId = getTenantId()
+  }
+  if (tenantId === DEFAULT_TENANT_ID) {
+    return StaticDatabases.AUDIT_LOGS.name
```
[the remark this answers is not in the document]
mike12345567: I'd prefer to keep consistency for now - originally I had it like that, but the inconsistency with the existing `global-db` concerned me - considering we don't know if/when multi-tenancy will occur for self host (there is no strict requirement for us to change this functionality).

packages/types/src/sdk/events/event.ts
```diff
@@ -180,6 +180,190 @@ export enum Event {
...
+  AUDIT_LOG_DOWNLOAD = "audit_log:download",
+}
+
+// all events that are not audited have been added to this record as undefined, this means
+// that Typescript can protect us against new events being added and auditing of those
+// events not being considered. This might be a little ugly, but provides a level of
+// Typescript build protection for the audit log feature, any new event also needs to be
+// added to this map, during which the developer will need to consider if it should be
+// a user facing event or not.
+export const AuditedEventFriendlyName: Record<Event, string | undefined> = {
```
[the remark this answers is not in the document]
mike12345567: Do you think this is strictly necessary? IMHO this makes the list of audited logs much harder to read/use, a direct map, with undefined telling us un-audited is far easier to maintain and read, the comment explains it well enough.

packages/backend-core/src/users.ts
mike12345567: I agree, there were some issues that Gerard found around this - I would like to do something about this in the future, as this issue has come up a few times. I just created a slightly nicer layer to make it easier to remove it when returning users as an enrichment to handle this for the time being.

packages/types/src/sdk/events/auth.ts
mike12345567: It is used for generating the proper friendly names, as per the requirements - if we did not do this then the email address and app names would not be searchable.

packages/server/src/db/index.ts
```diff
@@ -12,5 +12,5 @@ export function init() {
     dbConfig.allDbs = true
   }
 
-  coreInit({ db: dbConfig })
+  core.init({ db: dbConfig })
```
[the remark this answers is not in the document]
mike12345567: Sadly can't move it in here as it needs to return a promise - would prefer that the core init just does sync stuff and can be called/used from anywhere.

packages/server/src/startup.ts
```diff
@@ -17,6 +17,7 @@ import * as pro from "@budibase/pro"
 import * as api from "./api"
 import sdk from "./sdk"
 const pino = require("koa-pino-logger")
+import { sdk as proSdk } from "@budibase/pro"
```
Rory-Powell, line 20: Heads up we already have pro imported above

packages/backend-core/src/auth/auth.ts
```diff
@@ -225,6 +225,12 @@ export async function platformLogout(opts: PlatformLogoutOpts) {
 
   const sessionIds = sessions.map(({ sessionId }) => sessionId)
   await invalidateSessions(userId, { sessionIds, reason: "logout" })
-  await events.auth.logout()
+  let user: User | undefined
+  try {
+    user = await userCache.getUser(userId)
```
Rory-Powell, line 230: `ctx.user.email` should be available to us already

packages/backend-core/src/events/events.ts
```diff
@@ -1,8 +1,12 @@
-import { Event } from "@budibase/types"
+import { Event, AuditedEventFriendlyName } from "@budibase/types"
 import { processors } from "./processors"
 import identification from "./identification"
 import * as backfill from "./backfill"
 
+export function isAudited(event: Event) {
```
Rory-Powell, line 6: Can move this into `packages/backend-core/src/events/processors/AuditLogsProcessor.ts`

Issue #9286, opened 2023-01-09
Add the New Tab checkbox to the Navigate To action


melohagan:
**Describe the feature request**
The link component allows you to open a link in a new tab. I would like to do the same using the **Navigate To** action. 

Discussion: https://github.com/Budibase/budibase/discussions/9268

**Screenshots**
Add this setting:
![Screenshot 2023-01-09 at 15 44 33](https://user-images.githubusercontent.com/101575380/211348419-38903316-6d5b-46ba-8a2b-e3114caadec2.png)

Here:
![Screenshot 2023-01-09 at 15 45 06](https://user-images.githubusercontent.com/101575380/211348542-cf1c1265-d696-4eb5-a873-5fb433a7d392.png)

Issue #6243, opened 2022-06-09
Button actions - Show list of screens in a dropdown for "Navigate to"


AndyWhann:
When configuring links there is normally a dropdown which lists all of your screens to let you easily select which one you want to navigate to. 

On the button actions, when using "Navigate to" these options are not presented and you have to type or copy the URL from your screen. 

This would be nice to have here.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/e1f06863-0ad9-427f-aa83-fc52c3e2441b)

PR #9800, opened 2023-02-24
Support navigate to link in new tab

melohagan:
## Description
Users wanted to be able to use the **Navigate To** action to open an external URL in a new tab.

Also fixed a couple of minor issues:
 - external URLs can no longer be opened within a modal (it never worked)
 - if the **http** protocol is missing from an external URL then add it (this tripped up a user in the past)

Addresses: 
- https://github.com/Budibase/budibase/issues/9286
- https://github.com/Budibase/budibase/issues/6243

## Screenshots
Screenshot 2023-02-28 at 11 49 35

Screenshot 2023-02-28 at 11 49 55

Screenshot 2023-02-28 at 11 50 05

## Documentation

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
@@ -39,7 +39,6 @@
     on:change={() => (parameters.url = "")}
   />
   {#if parameters.type === "screen"}
-    <Label small>Screen</Label>
     <DrawerBindableCombobox
       title="Destination"
       placeholder="/screen"
@@ -52,7 +51,6 @@
     <div />
     <Checkbox text="Open screen in modal" bind:value={parameters.peek} />
   {:else}
-    <Label small>URL</Label>
     <DrawerBindableInput
       title="Destination"
       placeholder="/url"
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
@@ -1,26 +1,68 @@
 <script>
-  import { Label, Checkbox } from "@budibase/bbui"
+  import { store } from "builderStore"
+  import { onMount } from "svelte"
+  import { Label, Checkbox, Select } from "@budibase/bbui"
   import DrawerBindableInput from "components/common/bindings/DrawerBindableInput.svelte"
+  import DrawerBindableCombobox from "components/common/bindings/DrawerBindableCombobox.svelte"
 
   export let parameters
   export let bindings = []
+
+  $: urlOptions = $store.screens
+    .map(screen => screen.routing?.route)
+    .filter(x => x != null)
+
+  const typeOptions = [
+    {
+      label: "Screen",
+      value: "screen",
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
```diff
@@ -7,24 +7,28 @@
 </script>
 
 <div class="root">
-  <Label small>Screen</Label>
+  <Label small>Screen or URL</Label>
```
Rory-Powell, line 10: The changes here look good. 

The experience here I think might be confusing for users though for these reasons:
1. The form options suggest you can click both open in new tab and in modal, but that seems incompatible
2. The open in new tab mentions external link only - opening a screen in a new tab could be a valid case

I think ideally you have separate actions for: 
- `Navigate To Screen`
   - Screen input as is today 
   - Open in: (radio selection)
      - Modal 
      - New tab

- `Navigate To External URL`
   - URL input, with validation for protocol
   - Open in new tab (checkbox) 

Alternatively you could combine them into the existing action as is the case at the minute, but provide a picker at the top for destination:
- Screen
- External URL

Rory-Powell: Unrelated: I think the navigate to screen is a point of confusion for first time users, particularly around needing to build up the binding yourself e.g. `/fleet/{{ New Repeater.Fleet._id }}`. 

By splitting screen from url we give ourselves a bit more opportunity to improve the screen select e.g. by listing all the screens in a dropdown and rendering inputs for the individual path variables or adding params

melohagan: I updated the action to include a dropdown for the destination type. Updated the screenshots in the OP.

melohagan: Screenshot 2023-02-28 at 13 11 20

When there are path variables, it just adds the text directly. The user would need to update these. It possibly could be nice to parse the path variables out and allow the user to enter bindings if you think it's worth doing that, although getting what's already done merged in would still be an improvement over what's there now.

Also a small UI question. Do you think I should remove the label from the binding box. The *Destination* dropdown above already tells the user which type of input it is.

Rory-Powell: All good, thanks for clarifying. I think what you have there is great - a big improvement on what we have now. Updating the binding yourself should be more clear now. 

I think the label above the box is fine, but removing it might also be fine lol. What does it look like without it?

Rory-Powell (approved):
Awesome mel 👍

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/NavigateTo.svelte
melohagan: Screenshot 2023-02-28 at 13 52 42

Screenshot 2023-02-28 at 13 53 03

Looks a bit neater imo. What do you think?

Rory-Powell: no label LGTM

Issue #9778, opened 2023-02-22
[BUDI-6638] 403 Forbidden (Invalid CSRF token) thrown on Requests to the Public API


annafohlmeister:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: smatch (smatch.budibase.app)

**Describe the bug**
When attempting to access data from Budibase's internal data storage via the Public API a 403 Forbidden (Invalid CSRF token) is thrown whenever query parameters are set. The issues does not occur with URL parameters (I have not attempted, but would assume, that it would also occur for requests with bodies, ie POST / PUT). 

**Example #1: Searching for Table Rows**
GET https://budibase.app/api/public/v1/tables/$tableId/rows/search?param=xxx
403 Status with `{ message: 'Invalid CSRF token', status: 403 }` in the response body. 

**Example #2: Fetching a Table by ID**
GET https://budibase.app/api/public/v1/tables/$tableId
200 Status with `{ data: <tableData> }` in the response body.

I would not expect a CSRF token to be required on a stateless API. As Example #2 is working, and I can find no reference to CSRF tokens in the API documentation, I am confident that this is a bug and not a feature. 

**To Reproduce**
I'm working in nodeJS, therefore specific steps to replicate, but I believe this issue would arise anywhere:
 
1. Create a new NodeJS project and add the Axios Open API Client NPM Package (https://github.com/anttiviljami/openapi-client-axios/blob/master/packages/openapi-client-axios/README.md). 
2. Create an index.js file with:
```
const { OpenAPIClientAxios } = require('openapi-client-axios');

const openApiUrl = 'https://raw.githubusercontent.com/annafohlmeister/budibase/develop/packages/server/specs/openapi.json';

const config = {
    headers: {
        'x-budibase-api-key': 'xxx',
        'x-budibase-app-id': 'xxx'
    }
};

const tableId = 'xxx';

const api = new OpenAPIClientAxios({ definition: openApiUrl });

var test = async function () {

    const client = await api.getClient();

    /*
    // Working example 
    const table = await client.getTableById({tableId: tableId}, null, config);
    console.log(table);
    */

    // Failing example - feel free to add search parameters: { tableId: tableId, search: 'param' }
    const rows = await client.searchTableRows({ tableId: tableId }, null, config);
    console.log(rows);
}

var init = function () {
    api.init();
    test();
}

init();
```
4. Run the script. 
5. See error in console. 

[...]

PR #9828, opened 2023-02-28
Fix for operationId in OpenAPI specs

mike12345567:
## Description
Fix for #9778 - there was an issue with `operationId` in the OpenAPI spec, these need to be unique to be compliant with the spec as brought up in the issue. Also re-writing the spec generation to Typescript to be compliant with how backend code is now written.

Addresses: 
- #9778


Rory-Powell (approved):
LGTM

PClmnt (approved):
LGTM, nice to get these switched over.

Issue #9602, opened 2023-02-09
[BUDI-6551] MySQL - Options (and Long Form Text) field resets itself to text type (and cannot be changed) after fetching tables into new app 


Darojax:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.2.23
  - MariaDB 10.10.2

**Describe the bug**
When fetching tables into a new app 'Options' (and 'Long Field' Text) Columns are displayed as 'Text' type columns. The Options that had been entered previously and still work in the original app from where the column was created still works, and running a query shows that it's still an Enum column. But in the new app where the tables were fetched it shows as 'text' type. It's not possible to change the type shown in Budibase, nor is it possible to change the name of this column. I've tried deleting the column and recreating it as Options type from scratch, but still when fetching this column into a new app, the problem returns. Is this a bug or something I am doing wrong?

**To Reproduce**
1. Create an Options type column and enter some options.
2. Create a new App and fetch the tables with the Option column.
3. In the new App the Options column is now a text column, which cannot be name changed or type changed, and all the options are gone.
4. Deleting and recreating the column does not help. Both apps are running on the latest Budibase version 2.2.23

**Expected behavior**
Fetching the table should retain the Options type column.

**App Export**
Apps attached. the Recruit Status app is where the Options column was created and works fine, it's the personnel.status column. The ORBAT app is where those tables have been fetched and the personnel.status Options column no longer works as expected, it's text type and cannot be changed.

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Browser Chrome
 - Version 109.0.5414.120

[Recruit Status-export-1675903251875.tar.gz](https://github.com/Budibase/budibase/files/10691354/Recruit.Status-export-1675903251875.tar.gz)
[ORBAT-export-1675903276200.tar.gz](https://github.com/Budibase/budibase/files/10691355/ORBAT-export-1675903276200.tar.gz)

<sub>[BUDI-6551](https://linear.app/budibase/issue/BUDI-6551/mysql-options-and-long-form-text-field-resets-itself-to-text-type-and)</sub>

mike12345567:
Hi @Darojax - in the latest version I've been unable to re-create this issue - can try updating to `2.3.12`.

Darojax:
Hello @mike12345567 , thanks for getting back on this.

[...]

PR #9829, opened 2023-02-28
Query schema - keep changes when running query again

mike12345567:
## Description
Fixing an issue shown in #9602 where the schema type will revert each time you re-run a query, this will hold any previously selected schema information in the query.

Also - re-selecting the JSON tab each time the query is re-ran which allows the schema to re-render in the background.

Addresses: 
- #9602


PClmnt (approved):
LGTM

PR #9834, opened 2023-03-01
Fix to clear empty strings from user invite lastName fields.

deanhannigan:
## Description

When entering details on the accept invitation screen, if a user entered a last name and then cleared the field, an empty string would be passed to the api causing it to fail.

This fix ensures that empty strings are cleared before submission.


shogunpurple (approved):
LGTM! Although the bug fixes should go into `master`.

Issue #8063, opened 2022-09-30
Usage quota document conflicts can cause Writethrough to fail 


Rory-Powell:
**Hosting**
- Self
- Cloud

**Describe the bug**
When the usage quota Writethrough attempts to flush to couch there is the potential for multiple instances to try the write at once when under high load. e.g. many queries fired concurrently. The can cause couch db to conflict when trying to reconcile the document across instances. 

Suggested fix: 

**Locked Writes**
- Introduce a write lock using Redlock
- When a write is required, try to acquire write lock
- Ignore the write if the write lock cannot be acquired immediately 
- Upon acquiring the lock perform the write as below

**Conflict free writes**
1. Read the latest document 
2. Add the latest revision to the current write
3. Write the document
- If there is a conflict: 
   - Request conflicts with `GET /db/doc?conflicts=true`
   - Delete the conflicts `bulkDocs(resp._conflicts.map(conflict => ({ _id, _rev: conflict, _deleted: true })`
   - Perform steps 1-3 again

PR #9835, opened 2023-03-01
Bug - budi-5901 usage quota document conflicts can cause

adrinr:
## Description
Using redlock on WriteThrough persist. This will prevent multiple instances persisting the cache value to couchdb.

Also changing the `.put` to use force, to prevent issues where the redis rev value is behind couchdb.
This might be caused because of the absence of the redlock mechanism. Example:
```
1. Two processes running in parallel (more likely).
2. Process 1 updates couch successfully.
3. Process 2 updates couch, conflicting.
4. Process 1 updates redis, having the correct data.
5. Process 2 updates redis (even if we get a 409), so we revert the change to the previous version
```

But also because of some issues related to the redis update and the non atomic update between redis and couch. Examples:
```
a. Redis update failed, so it kept the old version
b. The process crashed between couch and redis update
```

Addresses: 
- https://github.com/Budibase/budibase/issues/8063

## App Export
- Not required

## Screenshots
No UI cahnges

## Documentation


packages/backend-core/src/cache/writethrough.ts
```diff
@@ -27,44 +28,61 @@ function makeCacheItem(doc: any, lastWrite: number | null = null): CacheItem {
...
-    } catch (err: any) {
-      if (err.status !== 409) {
-        throw err
-      } else {
-        // Swallow 409s but log them
-        logWarn(`Ignoring conflict in write-through cache`)
+    const lockResponse = await locks.doWithLock(
+      {
+        type: LockType.TRY_ONCE,
+        name: LockName.PERSIST_WRITETHROUGH(key!),
```
Rory-Powell, line 48: We can use `nameSuffix` here instead:
https://github.com/Budibase/budibase/blob/6befad8c658b5fd6483e04b3e024bb6b4eb8853d/packages/types/src/sdk/locks.ts#L34

That's the reason for the change to use constructor right?

packages/types/src/sdk/locks.ts
```diff
@@ -8,11 +8,22 @@ export enum LockType {
   DELAY_500 = "delay_500",
 }
 
-export enum LockName {
-  MIGRATIONS = "migrations",
-  TRIGGER_QUOTA = "trigger_quota",
-  SYNC_ACCOUNT_LICENSE = "sync_account_license",
-  UPDATE_TENANTS_DOC = "update_tenants_doc",
+export class LockName {
```
Rory-Powell, line 11: Mentioned above. we can revert this in favour of `nameSuffix`

packages/backend-core/src/redis/redlockImpl.ts
```diff
@@ -92,7 +107,7 @@ export const doWithLock = async (opts: LockOptions, task: any) => {
         // don't throw for try-once locks, they will always error
         // due to retry count (0) exceeded
         console.warn(e)
-        return
+        return { executed: false }
```
Rory-Powell, line 110: Nice 👍

packages/backend-core/src/cache/tests/writethrough.spec.ts
```diff
@@ -17,43 +20,80 @@ describe("writethrough", () => {
...
     })
   })
 
   describe("get", () => {
     it("should be able to retrieve", async () => {
       await config.doInTenant(async () => {
-        const response = await writethrough.get("test")
-        expect(response.value).toBe(3)
+        const response = await writethrough.get(docId)
+        expect(response.value).toBe(4)
```
Rory-Powell, line 96: Not really a fan of how this test relies on another test to setup it's data. Can we update to first write a unique doc id that we can assert on here?

packages/backend-core/src/cache/tests/writethrough.spec.ts
```diff
@@ -17,43 +20,80 @@ describe("writethrough", () => {
   const writethrough = new Writethrough(db, DELAY)
   const writethrough2 = new Writethrough(db2, DELAY)
 
+  const docId = structures.uuid()
```
Rory-Powell, line 23: Can we generate a new doc id in the `beforeEach`? So we can ensure these tests won't interfere with each other in future. I could see something like this producing flaky test runs in future

adrinr: These tests are written in a sequential way, so they are already dependent on each other.

packages/backend-core/src/cache/writethrough.ts
adrinr: Not really. Is will be a suffix of the key, but the key here does not need to be aware of it.
My change is to have dynamic keys. Until now we had redlocks for processes, and now we need it for documents. Because of this, we need dynamic keys. Hence this change

packages/backend-core/src/cache/tests/writethrough.spec.ts
adrinr: Same as above. I could refactor the tests, but I am not a big fan of many changes while fixing a bug

packages/backend-core/src/cache/tests/writethrough.spec.ts
Rory-Powell: Is that desirable though? Say I want to run / debug a particular test, it's useful to be able to run it in isolation rather than relying on the suite to pre-populate data for me. It can also be hard to read and reason about a test case where data is being setup in other test executions

mike12345567: I think in this case - this is the way the test was written (by me admittedly) - to be honest I think this is a perfectly fine test strategy, as I rarely, if ever run a single test case in isolation, as for very small test cases, I want to check that I haven't broken anything else in the process of trying to fix something. If a test file were to be very large, say taking a minute or so to run then it would be beneficial to be able to run parts of it in isolation, but for tests that are sub-second completion times, and will likely always be that way - I see no real harm in test suites like this.

mike12345567 (approved):
LGTM!

packages/backend-core/src/cache/tests/writethrough.spec.ts
Rory-Powell: It's more around using the debugger, I run individual tests all the time - when you have 5 or 6 tests all hitting your breakpoint and you need to proceed them all to get to your individual test case, it's tedious. For this reason and the reasons above I think it makes a lot of sense to not have tests cases rely on data from other tests. It's not a safe way of writing them given the order may change in future with refactors. 

I'm fine with leaving this as is given it's existing, but I don't think we should consider it an ok approach for future tests

packages/types/src/sdk/locks.ts
```diff
@@ -29,9 +30,9 @@ export interface LockOptions {
    */
   ttl: number
   /**
-   * The suffix to add to the lock name for additional uniqueness
+   * The resource to the lock name
```
Rory-Powell, line 33: Can we update this doc to be more descriptive? e.g.
> The individual resource to lock. This is useful for locking around very specific identifiers, e.g. a document that is prone to conflicts

Rory-Powell (approved):

PR #9842, opened 2023-03-01
Prevent showing user exists for password disabled actions

Rory-Powell:
## Description
Mask the existence of a user when they are prohibited from performing a password action 

Addresses: 
- Feedback from demo

### DIFF
--- a/packages/worker/src/sdk/auth/auth.ts
+++ b/packages/worker/src/sdk/auth/auth.ts
@@ -59,7 +59,7 @@ export const reset = async (email: string) => {
 
   // exit if user has sso
   if (await userSdk.isPreventPasswordActions(user)) {
-    throw new HTTPError("Password reset is disabled for this user", 400)
+    return
   }
 
   // send password reset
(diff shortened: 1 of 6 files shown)


PClmnt (approved):
LGTM

packages/worker/src/sdk/auth/auth.ts
```diff
@@ -59,7 +59,7 @@ export const reset = async (email: string) => {
 
   // exit if user has sso
   if (await userSdk.isPreventPasswordActions(user)) {
-    throw new HTTPError("Password reset is disabled for this user", 400)
+    return
```
adrinr, line 62: Will this not hide some unexpected behaviour?

Rory-Powell: It'll result in requesting a reset email to succeed, effectively means "if you have a valid account you will receive an email". In reality the email will never be sent as we exit early here. Same as above when we exit if the user is not found

PR #9843, opened 2023-03-01
Only allow admins to access audit logs screen

PClmnt:
## Description
Fixes an issue where the audit logs menu item was visible to admin users. Once they clicked it they got logged out as they had no permissions.


mike12345567 (approved):
LGTM!

PR #9844, opened 2023-03-01
Audit log multi-tenancy fix

mike12345567:
## Description
Minor fix for audit log system - after the switch to use an async queue for handling of audit log storage, context was lost - in multi-tenant environments need to carry the tenant ID into the queue job so that context can be applied.


Rory-Powell (approved):
LGTM

Issue #9739, opened 2023-02-19
[BUDI-6595] Export from Cloud, import to Self-hosted doesn't load the pictures / attachments.


sergey-vin:
**Hosting**
- Cloud

**Describe the bug**
Export from Cloud, import to Self-hosted doesn't load the pictures / attachments.

**To Reproduce**
1. Create an app
2. Create a BB table with Attachment column
3. Upload some pictures there
4. Export the app with data (don't set "Exclude rows")
5. Import to self-hosted version
6. Open Data section, observe the error: files are somewhat uploaded - they are present as data entries but content is broken
<img width="576" alt="image" src="https://user-images.githubusercontent.com/1080493/219939963-729c3f0e-407d-4954-9d73-a7aeea16d39b.png">
7. Try replacing one of them with a picture (re-upload it), and observe that it works: 
<img width="634" alt="image" src="https://user-images.githubusercontent.com/1080493/219940037-83860734-7f3a-4316-ae1a-20e31095a1ef.png">

This means that self-hosted storage works well, but the import fails.

**Expected behavior**
To see the pictures from the dump file, not broken, in self-host version right after the import.

<sub>[BUDI-6595](https://linear.app/budibase/issue/BUDI-6595/export-from-cloud-import-to-self-hosted-doesnt-load-the-pictures)</sub>

mike12345567:
Hi @sergey-vin - this should be resolved as of the latest release.

Issue #9738, opened 2023-02-19
[BUDI-6596] App Export includes attachments in a dump file even if "exclude rows" is selected.


sergey-vin:
**Hosting**
- Cloud

**Describe the bug**
There are 2 options the you try to export an app - include or exclude the data rows. But regardless of this toggle, at attachment folder gets added to the .tgz dump file anyway.

It's not affecting the "import" functionality - if rows were excluded, they won't be imported, but the dump file contains the attachments anyway - which makes it potentially very big.

**To Reproduce**
1. Create a new app from "Professional Directory Template", 
2. Click "exit to portal" and create 2 exports with and without data rows <img width="387" alt="image" src="https://user-images.githubusercontent.com/1080493/219939093-46c2dbd4-f4b0-47f4-9ade-137abe9877bb.png">
3. Import the 2 dumps as 2 new apps
4. Both apps work as expected - either show or not show rows in the Data section  
<img width="981" alt="image" src="https://user-images.githubusercontent.com/1080493/219939609-df32a976-7e25-467b-ab47-3fc1e8626838.png">
<img width="962" alt="image" src="https://user-images.githubusercontent.com/1080493/219939621-ee0d814f-d15c-4e51-90b5-5c4f3a18d1b3.png">
5. But the export files are of the same size
<img width="588" alt="image" src="https://user-images.githubusercontent.com/1080493/219939644-e70d26e6-38d7-4b45-867e-4807b7284ee2.png">
6. Then go to the original app, add 1 more column of type Attachment to the Directory table, add some pictures there
<img width="735" alt="image" src="https://user-images.githubusercontent.com/1080493/219939693-b7ce895f-f6b0-4b2e-82a6-d76c747370c8.png">
7. Repeat steps 2-5, and observe the same results, but now the dump files are both bigger: 
<img width="591" alt="image" src="https://user-images.githubusercontent.com/1080493/219939752-c5fd45ea-7ac3-4384-b57d-94233e04d116.png">
and both contain the "attachments" folder, which adds extra size: 
<img width="456" alt="image" src="https://user-images.githubusercontent.com/1080493/219939789-dfcb01a6-5283-4c51-b757-69681ec3560e.png">

**Expected behavior**
No attachment folder in the dump with "Exclude rows set to ON"

<sub>[BUDI-6596](https://linear.app/budibase/issue/BUDI-6596/app-export-includes-attachments-in-a-dump-file-even-if-exclude-rows-is)</sub>

[...]

PR #9849, opened 2023-03-01
Fix for exporting/importing apps with different attachment columns

mike12345567:
## Description
Fix for #9739 - there was an issue with the mango syntax, when working with multi attachment columns it was using an AND comparator instead of OR, it should be searching for rows that contain any attachment column, not all attachment columns.

Also includes a fix for #9738 - exclude rows could be specified but attachments would still be included, this was an oversight, have flipped it so that it only extracts the required static app files from object store when data is not required.

Addresses: 
- #9739
- #9738


shogunpurple (approved):
LGTM!

PR #9855, opened 2023-03-01
small UX updates to improve user addition and perceived loading of builder

shogunpurple:
## Description
Doesn't need to go in for release - just a couple of small ones I noticed when doing an end to end of the february features, and a little pet peeve I've wanted to fix for a while.

### Progressive builder loading
The entire builder loads from the outside in based on the response from an HTTP request that it doesn't need to wait on. This causes a little bit of a disjoint when you are navigating between portal and builder.

We can render the shell of the builder instantly, and wait for the inner content of the builder to load after that. This improves the initial loading performance and makes the builder look faster - all I've done here is moved some code to render the outer shell and updated the state logic to not `Update available` on first load.

### New User Side Panel
- Autofocus the modal when the side panel opens
- Add an enter keydown handler to allow adding a user with the keyboard

### Before
https://user-images.githubusercontent.com/11256663/222285700-d26d9fe9-8a82-4287-afc8-799a457812c7.mov

### After
https://user-images.githubusercontent.com/11256663/222285355-d1ce449c-3263-4e95-9fdd-f9593ae73ff5.mov

## Documentation

### DIFF
--- a/packages/builder/src/components/deploy/VersionModal.svelte
+++ b/packages/builder/src/components/deploy/VersionModal.svelte
@@ -24,7 +24,10 @@
   let updateModal
 
   $: appId = $store.appId
-  $: updateAvailable = clientPackage.version && $store.version && clientPackage.version !== $store.version
+  $: updateAvailable =
+    clientPackage.version &&
+    $store.version &&
+    clientPackage.version !== $store.version
   $: revertAvailable = $store.revertableVersion != null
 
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
LGTM! The loading change makes it feel way faster 👌

deanhannigan (approved):
LGTM! The autofocus and keypress behaviour in the side panel are great improvements to the UX

PR #9870, opened 2023-03-03
Chore - esbuild for server and builder

adrinr:
## Description
Using esbuild for worker and server. This required multiple steps as:
1. Changing pro to remove the package.json dependencies for versioning
2. Changing the build pipeline to actually check types when required
3. Adding an option to build the docker images locally (for dev proposes)
4. Optimise docker image builds
5. Change local references for files such as templates and the client.json outputs
6. Using nx/esbuild for the worker
7. Using esbuild for the server (not using `nx/esbuild` as it conflicts with the svelte files)
8. Splitting tests between apps and packages to improve pipeline flakiness

The benefits of these changes are the following:

- Much faster build times

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image | 

- Optimising docker image creation, as all the dependencies are in a single dist folder

Next steps:
- Using esbuild for bbui and client, as they are now the slower ones


packages/builder/src/pages/builder/portal/settings/auth/scim.svelte
```diff
@@ -115,37 +115,13 @@
     align-items: center;
   }
 
-  input[type="file"] {
```
mike12345567, line 118: what are these changes about out of interest? Did linting change in someway?

packages/cli/yarn.lock
mike12345567, line 1: Needs removed

packages/server/yarn.lock
mike12345567, line 1: Needs removed

.github/workflows/budibase_ci.yml
```diff
@@ -44,9 +44,31 @@ jobs:
...
+      - run: yarn build --projects @budibase/shared-core,@budibase/string-templates
+      - run: yarn test --ignore=@budibase/worker --ignore=@budibase/server --ignore=@budibase/pro
+      - uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
+          name: codecov-umbrella
+          verbose: true
 
-  test:
+  test-apps:
```
mike12345567, line 71: I do worry the name of this could be confusing - everytime I see it I assume its just the `server` package being tested due to it generally being called the `apps` service!

packages/cli/src/index.ts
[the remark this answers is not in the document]
adrinr: I don't see any change in here, not sure what GH is detecting as a change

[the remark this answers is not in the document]
mike12345567: looks like the file has changed permissions, from 644 to 755 - not sure why/how exactly

packages/builder/src/pages/builder/portal/settings/auth/scim.svelte
mike12345567: Good to know thanks!

.github/workflows/budibase_ci.yml
adrinr: This "apps" refer to the apps vs packages in monorepos. What do you think would be the best name to refer to the "deployables" vs npm-like packages?

packages/cli/src/index.ts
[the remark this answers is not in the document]
adrinr: I undid whatever change was that

.github/workflows/budibase_ci.yml
mike12345567: I feel like `services` is a better name - suggesting it is actual runnable services, rather than libraries.

Rory-Powell: +1 on `services`

hosting/docker-compose.build.yaml
```diff
@@ -0,0 +1,118 @@
+version: "3"
```
Rory-Powell, line 1: aXj57Bg_460s

Rory-Powell: Rather than having another locally running:
- `minio-service`
- `proxy-service`
- `couchdb-service`

I would greatly prefer to be able to run the dockerised server / worker alongside my existing infrastructure in `docker-compose.dev.yaml`. 

Often it'll be the case that you'll want to test something that you've been testing in dev and already have data setup for. If we could:

- `docker-compose.infra.yaml` -> replacement for `dev`, name change only
- `docker-compose.services.yaml` -> this file renamed, remove infra, keep server / worker

Then it'd be simpler to be able to 
- Run services via ide during development
- Stop those
- Run the docker services instead
- Pickup where I left off 


Cc: @mike12345567 @shogunpurple for thoughts here.

package.json
```diff
@@ -64,6 +69,7 @@
     "build:docker:single:image": "docker build -f hosting/single/Dockerfile -t budibase:latest .",
     "build:docker:single": "npm run build:docker:pre && npm run build:docker:single:image",
     "build:docker:dependencies": "docker build -f hosting/dependencies/Dockerfile -t budibase/dependencies:latest ./hosting",
+    "build:docker:run": "yarn build && docker-compose -f hosting/docker-compose.build.yaml up --build",
```
Rory-Powell, line 72: Alternative: `dev:docker`

packages/server/Dockerfile
```diff
@@ -15,27 +15,35 @@ ENV POSTHOG_TOKEN=phc_bIjZL7oh2GEUd2vqvTBH8WvrX0fWTFQMs6H5KQxiUxU
...
 
 # Install client for oracle datasource
 RUN apt-get install unzip libaio1
+COPY scripts/integrations/oracle/ scripts/integrations/oracle/
 RUN /bin/bash -e scripts/integrations/oracle/instantclient/linux/x86-64/install.sh 
 
+COPY package.json .
+RUN yarn \
+    && yarn cache clean 
+RUN apt-get remove -y --purge --auto-remove g++ make python \
```
Rory-Powell, line 32: Could consider merging the `yarn cache clean` line into this one then we could add a comment like
```dockerfile
 # Remove unneeded data from file system to reduce image size
RUN yarn cache clean  \
&& apt-get remove -y --purge --auto-remove g++ make python \
&& rm -rf /tmp/* /root/.node-gyp /usr/local/lib/node_modules/npm/node_modules/node-gyp
```

packages/server/package.json
```diff
@@ -10,22 +10,23 @@
...
+    "check:types": "tsc -p tsconfig.build.json --noEmit",
     "build:dev": "yarn prebuild && tsc --build --watch --preserveWatchOutput",
     "debug": "yarn build && node --expose-gc --inspect=9222 dist/index.js",
-    "postbuild": "copyfiles -u 1 src/**/*.svelte dist/ && copyfiles -u 1 src/**/*.hbs dist/ && copyfiles -u 1 src/**/*.json dist/",
     "test": "bash scripts/test.sh",
     "test:memory": "jest --maxWorkers=2 --logHeapUsage --forceExit",
     "test:watch": "jest --watch",
-    "predocker": "copyfiles -f ../client/dist/budibase-client.js ../client/manifest.json client",
-    "build:docker": "yarn run predocker && docker build . -t app-service --label version=$BUDIBASE_RELEASE_VERSION",
+    "predocker": "copyfiles -f ../client/dist/budibase-client.js ../client/manifest.json client && yarn build  --configuration=production",
```
Rory-Powell, line 21: Just to note for CI, we'll have already ran `yarn build --configuration=production` as a part of the pipeline definition. I'd be in favour of what you've done here, by keeping it inside the scripts definition instead. The only complication is the npm publish that done between the pipeline build and this docker build. Not sure what the best approach would be to prevent the need for this. 

Would the nx cache behaviour result in an almost instant build on this second step, or will the npm release mean it'll re-do a full build?

packages/server/scripts/build.js
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/node
+const { join } = require("path")
+const fs = require("fs")
+const coreBuild = require("../../../scripts/build")
+
+const dir = join(__dirname, "..")
+const entryPath = join(dir, "src")
+const outfilePath = join(dir, "dist")
+
+const ignoredFiles = ["definitions", "index", "utils"]
```
Rory-Powell, line 10: Can we add some comments for why we are doing this ignoring / more details on what we're doing with threading here

packages/server/scripts/test.sh
```diff
@@ -5,8 +5,8 @@ if [[ -n $CI ]]
 then
   # --runInBand performs better in ci where resources are limited
   export NODE_OPTIONS="--max-old-space-size=4096"
-  echo "jest --coverage --runInBand --forceExit"
-  jest --coverage --runInBand --forceExit
+  echo "jest --coverage --forceExit"
+  jest --coverage --forceExit
```
Rory-Powell, line 9: Will the changes in this pr make a difference such that we no longer need to runInBand in CI?

packages/server/tsconfig.build.json
```diff
@@ -10,7 +10,15 @@
     "incremental": true,
     "types": ["node", "jest"],
     "outDir": "dist/src",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "baseUrl": ".",
+    "paths": {
```
Rory-Powell, line 15: What's the implication of paths being moved into the build ts config? Do we use the `-p` now somewhere?

.github/workflows/budibase_ci.yml
```diff
@@ -44,9 +44,31 @@ jobs:
           node-version: 14.x
           cache: "yarn"
       - run: yarn
-      - run: yarn build
+      - run: yarn build --projects=@budibase/string-templates
```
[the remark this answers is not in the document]
adrinr: The "check types" is actually doing so. But I just realised it is not really verbose. I will change that.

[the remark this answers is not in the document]
adrinr: I was attempting to run only build server and worker, as all the other packages come as a dependency and esbuild analyses it. But I realised I was missing the frontend builds, so reverted to build everything everytime

packages/worker/project.json
```diff
@@ -0,0 +1,41 @@
...
+          }
+        ],
+        "format": ["cjs"],
+        "esbuildOptions": {
+          "outExtension": {
+            ".js": ".js"
+          },
+          "sourcemap": true
+        },
+        "minify": true,
```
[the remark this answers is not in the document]
adrinr: Yes, I left it like this as the debug works fine thanks to the sourcemap. We can change it if it causes trouble in other IDEs

hosting/docker-compose.build.yaml
adrinr: @mike12345567 you built this one are you are closer to the different configurations. Do you mind having a look at this?

packages/server/scripts/test.sh
adrinr: I did split the test to run both worker and server separately (as they are the flaky ones). With this change, it looks like the `runInBand` is not required (at least they are not failing) and hopefully will speed the tests a little bit

packages/server/scripts/build.js
adrinr: @mike12345567 ^^

packages/server/scripts/test.sh
Rory-Powell: Just FYI, the `runInBand` was much faster from previous testing. It can help in environments with limited resources. 

Could be worth trying a before and after commit in this pr just to see

packages/server/package.json
adrinr: As you mentioned the two build are required for different reasons, but no change should happen between both build. Nx should make it instant, so it should not be a problem.
The only change on the code should happen before the first build (Update versions)

package.json
adrinr: I like it, easier to understand and remember

packages/server/scripts/test.sh
adrinr: I checked and adding it runs in a similar time. I will add it back to avoid further issues until we fix the overall issue with the tests being slow

packages/server/scripts/build.js
mike12345567: The reasoning for this is that now our built version is simple `dist/index.js` - any kind of threaded approach in Node.js requires a runner file to work from - I played around with a lot of different methods, but we really want to be able to use forks.

Rather than trying to rewrite so that forks run the whole system, I instead went down a path of building the individual threads so that we have runner files for each of them e.g. `dist/automations.js` and `dist/query.js` - these can be ran totally independently and then the parent process can pass down data for processing to them.

The ignoring is simply to remove the files which really don't need to be built - they could be built and it wouldn't cause any issues, but this just means if any further threads are added in future they will naturally work (rather than including, which would mean adjustments to the build files).

packages/worker/project.json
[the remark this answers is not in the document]
mike12345567: This proves to be faster as well in terms of nodemon restart times, compared to our registered TS runner.

Rory-Powell (approved):

mike12345567 (approved):
Looks great - amazing work 🥳

Issue #9389, opened 2023-01-19
Dynamic filter - selecting fields from relationship table


phaethon:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists. There is #6550, but it seems to be obsolete as filtering by relationship fields is working

**Hosting**
- Self
  - Method: Google Cloud Run
  - Budibase Version: 2.1.42

**Describe the bug**
When I add a Dynamic filter for a data provider with a table with relationships (using PostgreSQL as the data source), I can filter by fields from relationship table. For Dynamic filter component there is a property Allowed Filter Fields. If I select any of those fields, then relationship fields disappear from the Dynamic filter. I cannot select relationship fields from Allowed Filter Fields component. Thus it is either all fields (with relationship fields) or selected fields from direct data source without relationship fields.

**To Reproduce**
1. Create 2 tables with a relationship between them in PostgreSQL
2. Create a screen with Data provider (and set source to a table), Dynamic filter (set provider to added provider, Table (set source to data provider)
3. Open Allowed Filter Fields property to see that relationship fields are not there.
4. Open Preview and see that relationship fields are visible
5. Open Allowed Filter Fields property and select any field(s)
6. Open Preview and see that relationship fields are not visible

**Expected behavior**
It should be possible to select relationship fields as part of Allowed Filter Fields of Dynamic filter.

Nicoautoxp:
Hi, 
I'm having the exact issue that is being described here. I see that it is marked as closed but i can't seem to find way to do the dynamic filtering. Could someone point me in the right direction?
Regards, 
Nicolas

PR #9877, opened 2023-03-03
Dynamic filter: Enrich columns

melohagan:
## Description
Users wanted to use the dynamic filter to filter related columns.

**Note**: I also added a minor fix to REST query URL cleaner to check for `https` before checking for `http`.

Addresses: 
- https://github.com/Budibase/budibase/issues/9389

## Screenshots
Screenshot 2023-03-03 at 16 11 42

Screenshot 2023-03-03 at 16 12 43

## Documentation

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnEditor.svelte
+++ b/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnEditor.svelte
@@ -22,7 +22,9 @@
 
   $: datasource = getDatasourceForProvider($currentAsset, componentInstance)
   $: schema = getSchema($currentAsset, datasource)
-  $: options = enrichedSchemaFields?.map(field => field.name)
+  $: options = allowCellEditing
+    ? Object.keys(schema || {})
+    : enrichedSchemaFields?.map(field => field.name)
   $: sanitisedValue = getValidColumns(value, options)
   $: updateBoundValue(sanitisedValue)
   $: enrichedSchemaFields = getFields(Object.values(schema) || [], {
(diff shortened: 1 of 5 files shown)


packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnEditor.svelte
```diff
@@ -21,9 +22,14 @@
...
   $: datasource = getDatasourceForProvider($currentAsset, componentInstance)
   $: schema = getSchema($currentAsset, datasource)
-  $: options = Object.keys(schema || {})
+  $: options = allowCellEditing
+    ? Object.keys(schema || {})
+    : enrichedSchemaFields?.map(field => field.name)
   $: sanitisedValue = getValidColumns(value, options)
   $: updateBoundValue(sanitisedValue)
+  $: enrichedSchemaFields = getFields(Object.values(schema) || [], {
+    allowLinks: true,
```
mike12345567, line 31: Should this check whether the table is SQL or not? If not SQL then we cannot allow links (not filterable for internal table).

melohagan: I did test this scenario, and related fields do not appear for internal tables.

It is actually already accounted for in the `getFields` function:

```js
if (!table || !table.sql) {
    return []
  }
```

See `searchFields.js`.

mike12345567: ah - is that not the `getTableFields` function? as the `getFields` function isn't aware of the table - its just the schema.

Would it be beneficial to change this to its table counterpart rather than using the raw `getFields` to avail of this functionality?

melohagan: As discussed, `getFields` calls `getTableFields` so it should be fine as is!

mike12345567 (approved):

PR #9885, opened 2023-03-06
Fix/gsheets test

shogunpurple:
## Description
Gsheets tests failing due to the following code to fetch the gsheets config, where the google env variables have not been set up:
```
export function getDefaultGoogleConfig(): GoogleInnerConfig | undefined {
  if (environment.GOOGLE_CLIENT_ID && environment.GOOGLE_CLIENT_SECRET) {
    return {
      clientID: environment.GOOGLE_CLIENT_ID!,
      clientSecret: environment.GOOGLE_CLIENT_SECRET!,
      activated: true,
    }
  }
}
```
Added the option to switch these variables on in the `TestConfiguration` so that the test returns the config.

### DIFF
--- a/packages/backend-core/src/configs/configs.ts
+++ b/packages/backend-core/src/configs/configs.ts
@@ -179,7 +179,7 @@ export async function getGoogleDatasourceConfig(): Promise<
 }
 
 export function getDefaultGoogleConfig(): GoogleInnerConfig | undefined {
-  if (env.isTest() || (environment.GOOGLE_CLIENT_ID && environment.GOOGLE_CLIENT_SECRET)) {
+  if (environment.GOOGLE_CLIENT_ID && environment.GOOGLE_CLIENT_SECRET) {
     return {
       clientID: environment.GOOGLE_CLIENT_ID!,
       clientSecret: environment.GOOGLE_CLIENT_SECRET!,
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/configs/configs.ts
```diff
@@ -179,7 +179,7 @@ export async function getGoogleDatasourceConfig(): Promise<
 }
 
 export function getDefaultGoogleConfig(): GoogleInnerConfig | undefined {
-  if (environment.GOOGLE_CLIENT_ID && environment.GOOGLE_CLIENT_SECRET) {
+  if (env.isTest() || (environment.GOOGLE_CLIENT_ID && environment.GOOGLE_CLIENT_SECRET)) {
```
adrinr, line 182: Should we not add the values `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` in the test config instead? If it's not there, we will be returning `undefined!`

shogunpurple: For the gsheets tests I think it's ok given that they are just mocked anyway - but it would be cleaner for sure to do that and not need to add conditional logic to the code here. Will update

adrinr (approved):

Issue #8557, opened 2022-11-07
Filters do not work for Google Sheets


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
Filters are not working for Google Sheets.
Discussed here: https://github.com/Budibase/budibase/discussions/8536

**To Reproduce**
1. Create a Google spreadsheet with 'Created At' column date string value `2021-09-01T15:33:00.133Z` and another row with an empty value
2. Fetch tables in Budibase
3. Add filter to one of the spreadsheets: 'Created At' *Is empty*
4. See that all rows are still shown
5. **Try any other filter with any other data type and operator - filters are not applied.**

**Expected behavior**
Filters should work for Google Sheets within the **Data** tab.

**Screenshots**
![image](https://user-images.githubusercontent.com/101575380/200325398-692c32cf-0567-48d9-b29d-0e83876d1cbd.png)

![image](https://user-images.githubusercontent.com/101575380/200325417-dd44b427-b71e-4697-82f7-eef5d7e874b0.png)

  
**App Export**
Mel Test on bb.budibase.app

PR #9886, opened 2023-03-06
Bug - BUDI-6068 filters do not work for google sheets

adrinr:
## Description
Enabling filtering and sorting on google sheet datasources:
1. Create `@budibase/shared-core` to share code between ts and js
2. Enable filtering on google sheet search
3. Enable sorting on google sheet search

TODO: display loading state while fetching on sort/filter. Otherwise, it looks broken

Addresses: 
- https://github.com/Budibase/budibase/issues/8557

## App Export
[My first app-export-1678203133867.tar.gz](https://github.com/Budibase/budibase/files/10911126/My.first.app-export-1678203133867.tar.gz)

## Screenshots
Functionality tested:
1. Filter by string value
2. Filter by numeric value
3. Sort by string value asc/desc
5. Sort by numeric value asc/desc
6. Filter and sort by numeric value

https://user-images.githubusercontent.com/15987277/223469131-79953672-095f-4026-ade5-57539e1ab701.mov

## Documentation


packages/server/src/integrations/googlesheets.ts
```diff
@@ -345,14 +349,20 @@ class GoogleSheetsIntegration implements DatasourcePlus {
...
+    sheet: string
+    filters?: SearchFilters
+    sort?: SortJson
+    paginate?: PaginationJson
+  }) {
     try {
       await this.connect()
       const sheet = this.client.sheetsByTitle[query.sheet]
       const rows = await sheet.getRows()
+      const filtered = runLuceneQuery(rows, query.filters)
```
adrinr, line 362: Making use of the new package via typescript

packages/frontend-core/src/constants.js
```diff
@@ -1,56 +1,7 @@
...
-  In: {
-    value: "oneOf",
-    label: "Is in",
-  },
-  ContainsAny: {
-    value: "containsAny",
-    label: "Has any",
-  },
-}
+export { OperatorOptions, SqlNumberTypeRangeMap } from "@budibase/data-utils"
```
adrinr, line 4: Making use of the new package via javacript

packages/server/src/integrations/googlesheets.ts
```diff
@@ -345,18 +349,35 @@ class GoogleSheetsIntegration implements DatasourcePlus {
...
         response.push(
           this.buildRowObject(headerValues, row._rawData, row._rowNumber)
         )
       }
+
+      if (query.sort) {
+        const sortInfo = Object.entries(query.sort)[0]
+        response = dataFilters.luceneSort(
+          response,
+          sortInfo[0],
```
melohagan, line 375: NAB: Is there a way to make this a little bit more readable / use types instead of `sortInfo[0]` and `sortInfo[1]` ?

packages/shared-core/src/filters.ts
```diff
@@ -106,12 +110,35 @@ const removeKeyNumbering = key => {
   }
 }
 
+type Filter = {
+  operator: keyof Query
+  field: string
+  type: any
+  value: any
+  externalType: keyof typeof SqlNumberTypeRangeMap
```
melohagan, line 118: NAB: What does `keyof typeof SqlNumberTypeRangeMap` do?

packages/shared-core/src/filters.ts
```diff
@@ -204,12 +235,29 @@ export const buildLuceneQuery = filter => {
   return query
 }
 
+const deepGet = (obj: { [x: string]: any }, key: string) => {
```
melohagan, line 238: There's already a `deepGet` method in **bbui** Helpers. Unless that's changed. Still it seems generic enough to be moved outside of the `filters.ts` file?

packages/shared-core/src/filters.ts
```diff
@@ -204,12 +235,29 @@ export const buildLuceneQuery = filter => {
...
+
 /**
  * Performs a client-side lucene search on an array of data
  * @param docs the data
  * @param query the JSON lucene query
  */
-export const runLuceneQuery = (docs, query) => {
+export const runLuceneQuery = (
+  docs: any[],
+  query?: { [x: string]: any; sheet?: string }
```
melohagan, line 259: NAB: Seems a little strange to me to have a property called `sheet` in a method as generic as this.

packages/shared-core/src/filters.ts
```diff
@@ -329,20 +397,28 @@ export const runLuceneQuery = (docs, query) => {
...
-    }
-  })
+  const parse =
+    sortType === "string" ? (x: any) => `${x}` : (x: string) => parseFloat(x)
+  return docs
+    .slice()
+    .sort((a: { [x: string]: any }, b: { [x: string]: any }) => {
+      const colA = parse(a[sort])
+      const colB = parse(b[sort])
+      if (sortOrder.toLowerCase() === "descending") {
```
melohagan, line 416: Case change in `"descending"`. Should a constant / enum be used here for safety?

packages/shared-core/src/filters.ts
```diff
@@ -351,7 +427,7 @@ export const luceneSort = (docs, sort, sortOrder, sortType = "string") => {
  * @param docs the data
  * @param limit the number of docs to limit to
  */
-export const luceneLimit = (docs, limit) => {
+export const luceneLimit = (docs: string | any[], limit: string) => {
```
melohagan, line 430: This seems a bit strange. How could docs by a string or an array of objects?

Should limit not be a number?

melohagan (changes requested):
LGTM, it'll be really nice to have filtering and sorting working for Google Sheets!

There were just a few minor questions I had.

packages/server/src/integrations/googlesheets.ts
adrinr: Cleaned

packages/shared-core/src/filters.ts
adrinr: It says the value should be a key of the selected type. Given the type
image
the `externalType` must be the following:
image

melohagan: Ah yes I see - you only need `keyof Query` because Query is a **type** already. Thanks!

packages/shared-core/src/filters.ts
adrinr: This new package will be able to get all these use cases to write ts code shareable in both ts and js. I did not want to move many things now, but this was an easy one. Done

packages/shared-core/src/filters.ts
adrinr: This was autogenerated, not sure why? 😅

packages/shared-core/src/filters.ts
adrinr: We have a mix of usages between js and ts, so we can not be type-safe. For this I check case insensitive

mike12345567 (approved):
LGTM - Thanks to @melohagan for the in-depth review!

melohagan (approved):

mike12345567:
One thing to add - when we merge this PR - can we `Squash and merge` it? We might want to cherry-pick this one for a minor release before the big end of month release - so squashing into a single commit would make life a lot easier for that.

Issue #9705, opened 2023-02-15
[BUDI-6582] Collapsible Settings Panels


budibot:
**COS**

* Condense the existing actions, size, margin, custom CSS properties into different tabs in the right hand panel
* Move required settings to the top of the settings panel, then non-required settings after that
* Can ignore the "conditions" part of this here - will be addressed in a separate ticket along with actions. Just keep actions and conditions button under the "General" tab until then

[Design - Component setting tabs (Figma)](https://www.figma.com/proto/Vdl0tM5Xh1SoNersW4FJ50/Component-setting-tabs?page-id=0%3A1&node-id=2%3A205&viewport=272%2C24%2C0.14&scaling=scale-down&starting-point-node-id=2%3A1883)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6582](https://linear.app/budibase/issue/BUDI-6582/collapsible-settings-panels)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6582](https://linear.app/budibase/issue/BUDI-6582/collapsible-settings-panels)</sub>

shogunpurple:
@AndrewBastin.whann @aptkingston 

* updated this with the figma with tabs
* Updated ticket description
* Updated to 3 points

melohagan:
It would be cool to potentially do this feature as part of this work? https://github.com/Budibase/budibase/issues/10043

They want to be able to drag the width of the panel out if needed.

PR #9895, opened 2023-03-06
Component settings section tabs

deanhannigan:
## Description

The settings UI for components in the builder has been separated into 3 areas: `General`, `Styles` and `Conditions`.

New tabs have been added to the settings header which can be used to show or hide the desired configuration options.

Addresses: 
- [BUDI-6582](https://github.com/Budibase/budibase/issues/9705)

### DIFF
--- a/packages/builder/src/components/design/Panel.svelte
+++ b/packages/builder/src/components/design/Panel.svelte
@@ -14,38 +14,42 @@
   export let borderRight = false
 
   let wide = false
+  $: customHeaderContent = $$slots["panel-header-content"]
 </script>
 
 <div class="panel" class:wide class:borderLeft class:borderRight>
-  <div class="header-wrap">
-    <div class="header">
-      {#if showBackButton}
-        <Icon name="ArrowLeft" hoverable on:click={onClickBackButton} />
-      {/if}
-      {#if icon}
-        <Icon name={icon} />
-      {/if}
-      <div class="title">
-        <Heading size="XXS">{title || ""}</Heading>
-      </div>
-      {#if expandable}
-        <Icon
-          name={wide ? "Minimize" : "Maximize"}
-          hoverable
-          on:click={() => (wide = !wide)}
-        />
-      {/if}
-      {#if showAddButton}
-        <div class="add-button" on:click={onClickAddButton}>
-          <Icon name="Add" />
-        </div>
-      {/if}
-      {#if showCloseButton}
-        <Icon name="Close" hoverable on:click={onClickCloseButton} />
-      {/if}
+  <div class="header" class:custom={customHeaderContent}>
+    {#if showBackButton}
+      <Icon name="ArrowLeft" hoverable on:click={onClickBackButton} />
+    {/if}
+    {#if icon}
+      <Icon name={icon} />
+    {/if}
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/design/Panel.svelte
```diff
@@ -76,9 +80,12 @@
     justify-content: space-between;
     align-items: center;
     padding: 0 var(--spacing-l);
-    border-bottom: var(--border-light);
     gap: var(--spacing-l);
   }
+  .header-wrap {
+    border-bottom: var(--border-light);
+    padding: var(--spacing-l) 0px;
```
aptkingston, line 87: This is adding additional padding into the core Panel component, so will affect the layout of all other panels - could we remove this padding and instead add it to node we insert inside this slot, so it doesn't affect all panels?

I know it probably doesn't affect layout because the padding is probably less than what spacing is already there due to the fixed flex rule, but I'd prefer to leave it up to whatever gets rendered inside the slot, so it's always in control.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ComponentSettingsPanel.svelte
```diff
@@ -25,32 +27,75 @@
...
+  let componentInstanceId = null
+
+  const tabs = ["general", "styles", "conditions"]
+  const updateSection = () => {
+    if (componentInstanceId !== $selectedComponent?._id) {
+      section = tabs[0]
+      componentInstanceId = $selectedComponent?._id
+    }
+  }
+  $: updateSection(componentInstance)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ComponentSettingsPanel.svelte:41]: NAB, but a nice refactor of the check to reset the section every time the component changes would be:
```js
$: id = $selectedComponent?._id
$: id, (section = tabs[0])
```
The first line is required to ensure that it memoizes a primitive to avoid firing the reactive statement every time any property inside the selected component property changes.

aptkingston (commented):
Nice work Dean!

Just as a duiscussion point, would "Settings" be a better title for the first tab instead of "General"? I know it's "General" in the designs so this isn't to do with implementation, but I think "Settings" would be a much better fit.

packages/builder/src/components/design/Panel.svelte
deanhannigan: Yep, that makes sense, I'll make the change here and ensure that it can't interfere with the padding of the component in general. As for renaming the `General` tab, I think you're right in that it gives a little bit more context.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ComponentSettingsPanel.svelte
Comment (deanhannigan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/components/[componentId]/_components/settings/ComponentSettingsPanel.svelte:41]: Nice, this is much neater!

aptkingston:
Did a quick run through with Andy and we agreed a slightly bigger button and lighter font weight looked better, so I've just pushed that tiny change up!

aptkingston (approved):
LGTM! Nice one Dean :+1:

PR #9911, opened 2023-03-07
User side panel UI fixes

deanhannigan:
## Description

Some UI fixes for the user side panel addressing:

- The panel header size not maintaining it's full height.
- User emails not truncating causing overflow.


shogunpurple (approved):
LGTM!

PR #9912, opened 2023-03-07
CLI dependency issue

mike12345567:
## Description
Fixing an issue with the CLI dependencies, the version of docker-compose we were using didn't support ES6 imports, despite appearing as if it did, upgrading to a version this is properly supported.


shogunpurple (approved):
LGTM!

PR #9915, opened 2023-03-07
fix roleId being undefined leading to group role overriding incorrectly

PClmnt:
## Description
Fixes an issue where when a user was in a user group, it would set the roleId to undefined - hence the group role would always take precedence. A user role will now always override a group role. 

The removed branch was leftover from the initial groups work, where role assignment worked slightly differently.


shogunpurple (approved):
Nice - I thought this one would have been a much larger fix!

mike12345567:
@PClmnt - can we move this change to `master` instead of `develop`? Would be good to get this change out.

PClmnt:
> @PClmnt - can we move this change to `master` instead of `develop`? Would be good to get this change out.

Done

PR #9918, opened 2023-03-07
Add release helm chart to release job

Rory-Powell:
## Description
- Publish chart under `0.0.0-master` so we can deploy to preprod in `budibase-deploys`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,6 +90,41 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_API_KEY }}
           BUDIBASE_RELEASE_VERSION: ${{ steps.previoustag.outputs.tag }}
 
+  release-helm-chart:
+    needs: [release-images]
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -90,14 +90,49 @@ jobs:
...
+
   # Trigger deploy to new EKS preprod environment
   trigger-deploy-to-preprod-env:
     needs: [release-helm-chart]
     runs-on: ubuntu-latest
     steps:
       - uses: passeidireto/trigger-external-workflow-action@main
         env:
-          PAYLOAD_VERSION: ${{ env.RELEASE_VERSION }}
+          PAYLOAD_VERSION: ${{ steps.previoustag.outputs.tag }}
```
jonnymccullagh, line 141: in the deploys repo we are looking for:
```
VERSION: ${{ github.event.client_payload.VERSION }}
```
So should this line read:
```
VERSION: ${{ steps.previoustag.outputs.tag }}
```
Or we change in the other repo

Rory-Powell: the `PAYLOAD_` prefix is required by the `trigger-external-workflow-actio` action. It uses that identifier to add to the payload of the request

PR #9919, opened 2023-03-07
Fix S3 CSP

PClmnt:
## Description
Fixes an issue introduced in https://github.com/Budibase/budibase/pull/8158 where I attempted to solve the problem of having many S3 regional domains in the CSP by using two wildcards like `https://*.s3.*amazonaws.com`. What I didn't realise is that two wild cards is actually invalid: https://wiki.mozilla.org/Security/CSP/Specification#Hostname_Wildcards 
I think this was missed in the testing of the bug as it was tested in an environment which wasn't using the production nginx config. 

I'm not sure if there's a more elegant solution without drastically changing the CSP around this, so I have added all the custom S3 regional domains back for now. 

If there is another way to handle this I am open to suggestions - I might be missing something simple. 

Addresses: 
- #9902


Rory-Powell (approved):
😢

Issue #9790, opened 2023-02-23
[BUDI-6697] Calculated columns lost on "fetch tables" in SQL


lukuser:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 2.2.19
  - App Version: 2.2.19

**Describe the bug**
Using SQL database, the "fetch tables" erases all calculated, formula-based javascript columns, even though the tables containing those columns have not changed at the source db.

**To Reproduce**
create new column -> function -> javascript
add code.
save
fetch tables.
is the column still there?

**Expected behavior**
I used calculated columns on several tables. I had to add something on one table, not even the one that had any of those formula columns. When doing "fetch tables", all the formula columns in all tables were gone and I had to recreate them. Luckily I had the backup of the code. Possibly any added column would disappear, not just a formula column.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser [chrome
 - Version 110.0.5481.104 (Official Build) (64-bit)

<sub>[BUDI-6697](https://linear.app/budibase/issue/BUDI-6697/calculated-columns-lost-on-fetch-tables-in-sql)</sub>

melohagan:
Hey @lukuser 

There's a related discussion here: https://github.com/Budibase/budibase/discussions/5539

A partial workaround was added with the following change: https://docs.budibase.com/docs/sql-datasource#fetching-specific-tables

This allows you to update specific tables, which may allow you to pull in an updated schema of one table without removing the formulas from another unrelated table.

PR #9923, opened 2023-03-07
SQL formulas - keeping after fetching

mike12345567:
## Description
Fix for #9790 - formulas never exist in the database and so were always being considered deleted, handling this.

Addresses: 
- #9790


shogunpurple (approved):
Nice work on the quick fix 👌

Issue #9934, opened 2023-03-07
API: Unable to save user - Password must be specified.


marian001:
**Checklist**
- [v ] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: digital ocean
  - Budibase Version: 2.4.2
  - App Version: 2.4.2

**Describe the bug**
I'm trying to update user through API. This is a new user registered and authenticated via Auth0 with no app assigned, so I'm trying to assign him a BASIC role in one of my apps. This assignment used to work in previous versions, but now the API returns response status 400 with message Unable to save user - Password must be specified.
I also can't assign the user to an app from within Access screen of app's web UI. When I try to do that (i.e. click Assign user, select user email, select Basic role and click Done), the wheel keeps turning and does nothing...

**To Reproduce**
1. API: go to https://docs.budibase.com/reference/put_users-userid and try executing the call...
2. UI: go to app's Access screen, click Assign user, select user email, select Basic role and click Done

**Expected behavior**
User gets BASIC role enabling access to my app

 
**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome
 - Version 110

shogunpurple:
This one is resolved in the latest release - which is currently still running for self host. The image will be deployed in around 10 minutes. Closing for now

marian001:
just upgraded - the same problem on 2.4.4

shogunpurple:
Hm it could be something specific with the public API - I will investigate and see if I can repro

shogunpurple:
@marian001 we are working on the fix - best thing to do for now would be to revert to the previous version if it's critical.

marian001:
Since I upgraded directly from 2.4.2 (without reverting to version where this wasn't a problem) to 2.4.4, I can't now return to any working version (as far as I know..). We'll wait for the fix..

Rory-Powell:
Hi @marian001, could you try again with the latest version

marian001:
Fixed! Thank you very much!

PR #9941, opened 2023-03-08
Don't require password on update if user doesn't have one

Rory-Powell:
## Description
Don't require password on update if user doesn't have one - this is required for updating SSO users. 

Previously we didn't perform the password check if the db user existed at all. This was corrected in the previous release, however caused a bug where users who don't have a password  weren't able to update. 

Addresses: 
- https://linear.app/budibase/issue/BUDI-6689/unable-to-update-profile-for-sso-user
- https://github.com/Budibase/budibase/issues/9934


shogunpurple (approved):
LGTM! Nice work

mike12345567 (approved):

PR #9944, opened 2023-03-08
Revert "pinning MinIO versions"

mike12345567:
Reverts Budibase/budibase#9943

This cannot be added for now - need to have a think about how to support this, discussed on Discord, we cannot pin a specific version of docker-compose as any installs after October 2022 will be broken.


shogunpurple:
I think we may as well just remove this from everywhere it's used now - from dev also

shogunpurple (approved):
LGTM!

PR #9948, opened 2023-03-08
Allow Is in for _id SQL

melohagan:
## Description
Allow **Is in** filter on **_id** field so that Relationship Picker can be used as the input of a Data Provider filter. (See tutorial)

Addresses: 
- https://github.com/Budibase/budibase/issues/9015


mike12345567:
Closing this, the lucene layer has moved to `shared-core` in the latest version - this PR while it looks like it can be merged will cause conflicts/not work - I'll add it in a separate PR.

mike12345567:
Actually decided to merge this - then will commit separately to resolve the issue it will cause (the problem hasn't been merged to master yet - will fix it in the develop -> master merge)

PR #9950, opened 2023-03-08
es6 and commonjs versions of packages

adrinr:
## Description
Creates an es6 and a commonjs version of the types and the shared-core. Required to allow the consumption of both frontend and backend services, via node imports, rollup, vite and jest.

### DIFF
--- a/packages/client/rollup.config.js
+++ b/packages/client/rollup.config.js
@@ -20,6 +20,19 @@ const ignoredWarnings = [
   "a11y-click-events-have-key-events",
 ]
 
+const devPaths = production
+  ? []
+  : [
+      {
+        find: "@budibase/shared-core",
+        replacement: path.resolve("../shared-core/dist/mjs/src/index"),
+      },
+      {
+        find: "@budibase/types",
+        replacement: path.resolve("../types/dist/mjs/index"),
+      },
+    ]
+
 export default {
   input: "src/index.js",
   output: [
@@ -69,6 +82,7 @@ export default {
           find: "sdk",
           replacement: path.resolve("./src/sdk"),
         },
+        ...devPaths,
       ],
     }),
     svelte({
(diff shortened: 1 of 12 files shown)


packages/client/rollup.config.js
```diff
@@ -20,6 +20,19 @@ const ignoredWarnings = [
   "a11y-click-events-have-key-events",
 ]
 
+const devPaths = production
+  ? []
+  : [
+      {
+        find: "@budibase/shared-core",
+        replacement: path.resolve("../shared-core/dist/mjs/src/index"),
```
aptkingston, line 28: Is this still needed, even with both `shared-core` and `types` now specifying a default ESM `import` export? The client should be able to consume it now without these additional aliases, like it does for `string-templates` which also gets built to both CJS and ESM bundles.

packages/shared-core/package.json
```diff
@@ -2,21 +2,28 @@
   "name": "@budibase/shared-core",
   "version": "2.4.8-alpha.1",
   "description": "Shared data utils",
-  "main": "dist/src/index.js",
-  "module": "dist/src/index.js",
-  "types": "dist/src/index.d.ts",
+  "main": "dist/mjs/src/index.js",
```
aptkingston, line 5: Normally I would say that the `main` key should probably point to the CJS build, but when this module is intended to be consumed via Typescript from node environments maybe that changes things.

aptkingston (approved):
LGTM, but I think we can remove the client aliases :+1:

packages/client/rollup.config.js
adrinr: It's needed for the local dev for now, as the published package does not have it, and rollup does not know about the local package.

adrinr: This only runs on development, for production it will always read from the npm repo as it does now

Issue #9751, opened 2023-02-20
[BUDI-6613] Horizontal align right hand settings panel fields


shogunpurple:
When I'm viewing the settings panel on the right hand side of the builder, I want the titles and form fields to be horizontally aligned so that I can read them more easily line by line.

**COS**

* Align titles and inputs horizontally in right hand panel
* Increase width of the right hand panel

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/ca1c41f0-229b-42e4-95f1-6709ad944aee/7ef1ef94-bb66-4bb5-a411-8b73c8371956)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6613](https://linear.app/budibase/issue/BUDI-6613/horizontal-align-right-hand-settings-panel-fields)</sub>

PR #9951, opened 2023-03-09
Settings panel layout update

aptkingston:
## Description
This PR updates the layout of the component settings panel to use left aligned labels rather than top aligned labels. I've also gone through every setting of every component and standardised the casing, as well as reducing the length of all labels where required so that they fit in the new layout.

Left aligned labels are more compact, allowing more settings to be displayed without scrolling, and are much easier to scan visually. The settings panel width has been increased slightly to accommodate this change. The new width is 310px, which is identical to the width of the left side panel combined with the side nav. So despite this panel being wider than the component tree, the page now looks symmetrical because of the left icon nav.

Width symmetry:
image
image

The styles section has also been updated to match this new design, as it uses the same underlying components.

I've also had to slightly adjust the client library mobile breakpoint to avoid considering the app preview on a 14 inch macbook screen with default zoom as mobile.

Addresses: 
- #9751

## Screenshots
Some example components with the new alignment:
image
image
image

Style section, now looking the same as settings:
image

Full screen, effective res of 14 inch macbook pro (default zoom, 1400px):
image

Full screen, effective res of 16 inch macbook pro (default zoom, 1728px):
image

Full screen, 1080p:
image

## Documentation

### DIFF
--- a/packages/bbui/src/Drawer/Drawer.svelte
+++ b/packages/bbui/src/Drawer/Drawer.svelte
@@ -7,7 +7,7 @@
   export let title
   export let fillWidth
   export let left = "314px"
-  export let width = "calc(100% - 576px)"
+  export let width = "calc(100% - 626px)"
 
   let visible = false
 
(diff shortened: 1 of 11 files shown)


shogunpurple (approved):
LGTM! Solid improvement

PR #9963, opened 2023-03-09
QA-6 Api tests public api key generation

Pedro-Revez-Silva:
## Description
Public API tests require an API key to work and currently this API key is a static environment variable. This PR fixes this by creating a new account, tenant and application before the tests, all necessary steps to be able to retrieve an public API key. 

Addresses: 
- https://linear.app/budibase/issue/QA-6/refactor-public-api-tests-to-handle-environments


qa-core/scripts/jestSetup.js
```diff
@@ -1,10 +1,10 @@
-const env = require("../src/environment")
+//const env = require("../src/environment")
 
-env._set("BUDIBASE_SERVER_URL", "http://localhost:4100")
-env._set(
-  "BUDIBASE_PUBLIC_API_KEY",
-  "a65722f06bee5caeadc5d7ca2f543a43-d610e627344210c643bb726f"
-)
+//env._set("BUDIBASE_SERVER_URL", "http://localhost:4100")
```
shogunpurple, line 3: remove commented code

qa-core/src/config/internal-api/TestConfiguration/InternalAPIClient.ts
```diff
@@ -49,14 +49,17 @@ class InternalAPIClient {
...
+      ) {
         console.error("Error in apiCall")
         console.error("Response:")
         console.error(response)
+        const json = await response.json()
         console.error("Response body:")
-        console.error(response.body)
-        console.error("Request body:")
-        console.error(requestOptions.body)
+        console.error(json)
```
shogunpurple, line 62: `console.error` here doesn't seem right - might have been left in from debugging just.

qa-core/src/config/public-api/TestConfiguration/PublicAPIClient.ts
```diff
@@ -11,39 +11,64 @@ interface ApiOptions {
...
+
   apiCall =
     (method: APIMethod) =>
     async (url = "", options: ApiOptions = {}) => {
       const requestOptions = {
         method,
         body: JSON.stringify(options.body),
         headers: {
-          "x-budibase-api-key": this.apiKey,
+          "x-budibase-api-key": this.apiKey ? this.apiKey : null,
```
shogunpurple, line 47:
```suggestion
          "x-budibase-api-key": this.apiKey || null,
```

qa-core/src/config/public-api/TestConfiguration/PublicAPIClient.ts
```diff
@@ -11,39 +11,64 @@ interface ApiOptions {
...
-      if (response.status !== 200) {
+      const response = await fetch(`https://${process.env.TENANT_ID}.${this.host}${url}`, requestOptions)
+
+      if (
+        response.status == 404 ||
+        response.status == 500 ||
+        response.status == 403
+      ) {
+        console.error("Error in apiCall")
+        console.error("Response:")
```
shogunpurple, line 68: You can do:
```js
console.error("Response:", response)
console.error("Response body:", response.body)
```

qa-core/src/config/public-api/TestConfiguration/accountsAPIClient.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      // @ts-ignore
+      const response = await fetch(`${this.host}${url}`, requestOptions)
+      if (
+        response.status == 404 ||
+        response.status == 500 ||
+        response.status == 400
+      ) {
+        console.error("Error in apiCall")
```
shogunpurple, line 50: same as above

qa-core/src/config/public-api/TestConfiguration/accountsAPIClient.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type APIMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
+
+interface ApiOptions {
+  method?: APIMethod
+  body?: object
+  headers?: HeadersInit | undefined
+}
+
+class AccountsAPIClient {
```
shogunpurple, line 12: Is it necessary to create a completely new API client here? It seems quite similar to the one that lives in `PublicAPIClient`. NAB for this PR, but we should think about combining them in some way.

qa-core/src/config/public-api/TestConfiguration/index.ts
```diff
@@ -3,19 +3,79 @@ import ApplicationApi from "./applications"
...
+
+  async setupAccountAndTenant() {
+    const account = generateAccount()
+    await this.accounts.validateEmail(<string>account.email)
+    await this.accounts.validateTenantId(<string>account.tenantId)
+    process.env.TENANT_ID = <string>account.tenantId
+    await this.accounts.create(account)
+    await this.updateApiClients(<string>account.tenantName)
+    await this.auth.login(<string>account.email, <string>account.password)
+    const body = {
```
shogunpurple, line 57: I'd add a comment in here just to explain why this is necessary.

qa-core/src/config/public-api/fixtures/accounts.ts
```diff
@@ -0,0 +1,21 @@
+import { NewAccount } from "./types/newAccount"
+
+import generator from "../../generator"
+import { Hosting } from "@budibase/types"
+
+export const generateAccount = (): Partial<NewAccount> => {
+  const randomGuid = generator.guid()
+  let tenant: string = "a" + randomGuid
```
shogunpurple, line 8: what is the `"a"` for here? Is it necessary?

qa-core/src/config/internal-api/TestConfiguration/InternalAPIClient.ts
Pedro-Revez-Silva: @shogunpurple `console.log` doesn't work in the pipeline and its useful to have this information in case of failures. However and as you pointed in another comment, these logs are done poorly. Will change.

qa-core/src/config/public-api/TestConfiguration/accountsAPIClient.ts
Pedro-Revez-Silva: @shogunpurple Yes because the URL they are calling is different. I have plans to refactor the API tests because there is code repetition in both internal and public API configurations and different services (with different urls) require different API client classes to be defined, but they share all the same functionality, including error handling. So it seems we need a base API client class.

PR #9980, opened 2023-03-10
BUDI-6683 - SCIM usergroup provisioning endpoints skeleton

adrinr:
## Description
Creating the skeleton for SCIM endpoints. Changes:
1. Create endpoints structure in pro
2. Enable auth by bearer token
3. Implement mock results for get and create endpoints


packages/types/src/api/web/global/scim/users.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export interface ScimListResponse {
+  schemas: ["urn:ietf:params:scim:api:messages:2.0:ListResponse"]
+  totalResults: number
+  Resources: ScimUser[]
+  startIndex: number
+  itemsPerPage: number
+}
+
+export interface ScimUserRequest {
```
Rory-Powell, line 34: Could we add the verb for this request, like `ScimCreateUserRequest`, or if it's the same for create and update could go with `ScimSaveUserRequest`. 

For `ScimListResponse`, given we have users and groups this could conflict, the adding in user might help: `ScimUserListResponse`

packages/types/src/api/web/global/scim/users.ts
```diff
@@ -0,0 +1,83 @@
...
+      type: "work"
+      primary: true
+    }
+  ]
+}
+
+export interface ScimListResponse {
+  schemas: ["urn:ietf:params:scim:api:messages:2.0:ListResponse"]
+  totalResults: number
+  Resources: ScimUser[]
```
Rory-Powell, line 29: Could we do away with `ScimUser` interface and use `ScimUserResponse` here instead?
Looks like they have the same properties. 

Alternatively could do 
```ts
interface ScimUserResponse extends ScimUser {}
```
but preference would be on using the single interface, I think it's fine to have the response type be re-used in another response. I actually think that makes a lot of sense so the contracts line up for the singular vs the `Resources`.

packages/worker/src/api/routes/index.ts
```diff
@@ -39,4 +39,5 @@ export const routes: Router[] = [
   accountRoutes,
   restoreRoutes,
   eventRoutes,
+  ...api.scimRoutes,
```
Rory-Powell, line 42: Not directly related to these changes but could we update the naming here to:
```ts
import { api as pro } from "@budibase/pro"
```

Also looks like we pull the `userGroupRoutes` and `auditLogRoutes` out into their own variables. Could we be consistent with the `scimRoutes` and either
- use a variable, or
- use the spread pattern for the others too

packages/types/src/api/web/global/scim/users.ts
adrinr: Totally agree, I actually changed that in a further PR. Changing it here

packages/types/src/api/web/global/scim/users.ts
adrinr: Totally agree, I actually changed that in a further PR. Changing it here

packages/types/src/api/web/global/scim/users.ts
adrinr: Same, already modifies in a further PR. Changing it here

packages/worker/src/api/routes/index.ts
```diff
@@ -33,10 +30,11 @@ export const routes: Router[] = [
...
   licenseRoutes,
-  userGroupRoutes,
-  auditLogRoutes,
+  pro.groups,
+  pro.auditLogs,
   migrationRoutes,
   accountRoutes,
   restoreRoutes,
   eventRoutes,
+  ...pro.scimRoutes,
```
mike12345567, line 39: preferred if the exported prop is `scim` instead of `scimRoutes` to line up with the rest of the pro API exports.

adrinr: Should the other ones not be `groupsRoutes`, to be consistent with the worker?

mike12345567 (approved):

PR #9982, opened 2023-03-10
Rotatable secrets

Rory-Powell:
## Description
Ability to rotate internal environment based secrets safely with no downtime

## Documentation
The strategy on how to use the new env vars for rotation can be seen here:
- https://budibase.atlassian.net/wiki/spaces/EN/pages/214335493/Secret+Rotation


packages/backend-core/src/security/encryption.ts
```diff
@@ -8,7 +8,7 @@ const RANDOM_BYTES = 16
 const STRETCH_LENGTH = 32
 
 export enum SecretOption {
-  JWT = "jwt",
+  API = "api",
```
mike12345567, line 11: How will this work in self host? Given the secret doesn't exist there - we probably need to consider using the `JWT_SECRET` as a fallback right? We should probably also update this for new installs.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -332,8 +331,8 @@ class TestConfiguration {
         roleId: roleId,
         appId,
       }
-      const authToken = auth.jwt.sign(authObj, env.JWT_SECRET)
-      const appToken = auth.jwt.sign(app, env.JWT_SECRET)
+      const authToken = auth.jwt.sign(authObj, coreEnv.JWT_SECRETS[0])
```
mike12345567, line 334: I think it would be preferential if we had a `latest` property and then maybe an `old` array rather than just treating it as an array - comparing it to other environment variables its very odd that this is an array - its not well explained.

packages/server/src/utilities/workerRequests.ts
```diff
@@ -15,7 +16,7 @@ export function request(ctx?: BBContext, request?: any) {
     request.headers = {}
   }
   if (!ctx) {
-    request.headers[constants.Header.API_KEY] = env.INTERNAL_API_KEY
+    request.headers[constants.Header.API_KEY] = coreEnv.INTERNAL_API_KEYS[0]
```
mike12345567, line 19: Same as comment about `JWT_SECRET`.

packages/backend-core/src/security/encryption.ts
Rory-Powell: That's what is does atm, always falls back to the JWT secret value. I'll add it for new installs too just incase someone does ever want to change their JWT secret

mike12345567 (changes requested):
I think we'll need to address the implications this has for self host - especially for API keys - with the introduction of a new env variable, we should really have it fallback to the old secret for existing self host installs (perhaps all self host installs and only use the new one in the Cloud?)

packages/server/src/tests/utilities/TestConfiguration.ts
Rory-Powell: I was toying with a get latest function as well, I'll add that

Rory-Powell: And similar for api key

Issue #9937, opened 2023-03-08
[BUDI-6706] Google authentication enabled as default after upgrade even it is not activated/configured


MDXDave:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.4.4
  
**Describe the bug**
After upgrading from v2.3.15 to v2.4.4, the _Sign-In with Google_ button is shown even the google authentication is not enabled nor configured.

To disable it, you need to enter something in the clientId and clientSecret field, enable the authentication with Google, submit the the form and than disable the authentication with Google and submit the form again.

**To Reproduce**
1. Upgrade from v2.3.15 to v2.4.4
2. Open the login screen

**Expected behavior**
The _Sign-in with Google_-button should not be visible if the authentication is not enabled.

<sub>[BUDI-6706](https://linear.app/budibase/issue/BUDI-6706/google-authentication-enabled-as-default-after-upgrade-even-it-is-not)</sub>

mike12345567:
This issue should be resolved in the latest version.

PR #9983, opened 2023-03-10
Disable google auth if no config found

mike12345567:
## Description
Fix for #9937 - Google config was pre-activated always, even if no google config had been created.

Addresses: 
- #9937


Rory-Powell (approved):
LGTM

PR #9985, opened 2023-03-10
Fix does not contain filter in Lucene, when using match any filter

mike12345567:
## Description
Raised on Linear: https://linear.app/budibase/issue/BUDI-6720/bug-filters-match-any-filter-doesnt-work-as-expected

This was a weird issue - specifically the `NOT` keyword has weird interactions with `OR` syntax - the fix is relatively minor, essentially combining the different does not contain operations where possible, then utilising de-morgans law to invert the condition, e.g. rather than `NOT contains(...) OR NOT contains(....)` do `NOT (contains(...) AND contains(...))`. The issue is that the `NOT` operator only removes entries from your results, it does not perform a search, whereas performing the actual contains searches, then filtering those out of all the results for the table gives a better result.

### DIFF
--- a/.github/workflows/deploy-preprod.yml
+++ b/.github/workflows/deploy-preprod.yml
@@ -1,0 +1,53 @@
+name: "deploy-preprod"
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  deploy-to-legacy-preprod-env:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Get Previous tag'
+        id: previoustag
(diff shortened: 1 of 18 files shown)


aptkingston (approved):
De Morgan! There's a name I haven't heard in a while. LGTM :+1: 
I'm guessing with the tests this is safe enough to go straight to master yea? Just thinking that this is pretty core logic.

mike12345567:
Yeap @aptkingston - to be fair I can merge it to develop, its not really that beneficial out in prod now anyway!

PR #9995, opened 2023-03-13
Update package versions

adrinr:
## Description
Fixes the issue that keeps the @types package not aligned. A breaking change has been added and not using the latest version for all of them causes issues on the UI.
A fix getting sure all the packages are on sync is required, but this PR will fix the existing breaking UI


shogunpurple (approved):
LGTM!

Issue #10001, opened 2023-03-13
[BUDI-6734] Onboarding tour not loading


deanhannigan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Describe the bug**
Minor timing issue caused by recent updates to the builder loading flow. The tour nodes load before the tour has been fully initialised. 

The tour eventually tries to run, but the nodes are empty.

**To Reproduce**

1. Create a new developer/admin user.
2. Log in as the new user and open/create an application
3. No onboarding tour will load

**Expected behavior**
If the user is new or has never gone through onboarding, they should get the full onboarding tour.

**Additional context**
Currently only affects develop

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6734](https://linear.app/budibase/issue/BUDI-6734/onboarding-tour-not-loading)</sub>

PR #10003, opened 2023-03-13
Fix for onboarding initialisation.

deanhannigan:
## Description

Recent changes to the builder loading flow meant that tour nodes were attempting to register before the tour configuration had been built and set.

The tour system has been updated to be responsive to changes in the builder tour state. Tour nodes will now wait until they have been rendered on screen, a builder tour key has been set and the current tour step is selected. 

Addresses: 
- https://github.com/Budibase/budibase/issues/10001


aptkingston (approved):
LGTM!

Issue #10017, opened 2023-03-14
Helm Chart - Provide ability to configure custom schedulerName


StarZeus:
**Describe the feature request**
schedulerName is not part of the deployment templates. So it's forced to use default-scheduler. We need a way to configure schedulerName from values.yaml file(s)

PR #10018, opened 2023-03-14
Ability to configure schedulerName for helm charts

StarZeus:
## Description
This adds an optional schedulerName in helm chart deployment allows users to configure it through values.yaml file.

Addresses: 
- Issue https://github.com/Budibase/budibase/issues/10017

## Documentation

### DIFF
--- a/charts/budibase/templates/minio-service-deployment.yaml
+++ b/charts/budibase/templates/minio-service-deployment.yaml
@@ -72,9 +72,9 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- with .Values.schedulerName }}
+      {{ if .Values.schedulerName }}
       schedulerName: {{ .Values.schedulerName | quote }}
-      {{- end }}
+      {{ end }}
       {{ if .Values.imagePullSecrets }}
       imagePullSecrets:
       {{- toYaml .Values.imagePullSecrets | nindent 6 }}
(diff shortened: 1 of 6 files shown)


charts/budibase/templates/minio-service-deployment.yaml
```diff
@@ -72,6 +72,9 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+      {{- with .Values.schedulerName }}
```
Rory-Powell, line 75: I think using `with` here might not be the correct syntax. From having a look at: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

I see that the `with` will rewrite `.` to mean the value of `.Values.schedulerName`. In this case that would mean the inner usage of `.Values.schedulerName` wouldn't exist. 

Two options
- We can switch the usages in this pr over to `if` instead
- Update the inner `.Values.schedulerName` in each case to just be `.`

Rory-Powell (changes requested):
Comment on usage of `with`

charts/budibase/templates/minio-service-deployment.yaml
StarZeus: `with` is not required in these cases. So changed it to `if`. It should be good now.

Rory-Powell (approved):

Rory-Powell:
@StarZeus I've merged this to develop to bring in your changes to our development cycle. If you wish, cherry-pick the commit and raise as a new pr to master - we can get it released a bit sooner then

PR #10019, opened 2023-03-14
BUDI-6683 - Delete user

adrinr:
## Description
Add delete scim user endpoint

### DIFF
--- a/packages/worker/src/initPro.ts
+++ b/packages/worker/src/initPro.ts
@@ -6,7 +6,7 @@ export const initPro = async () => {
     scimUserServiceConfig: {
       functions: {
         saveUser: userSdk.save,
-        removeUser: (id: string) => userSdk.destroy(id, undefined),
+        removeUser: (id: string) => userSdk.destroy(id),
       },
     },
   })
--- a/packages/worker/src/initPro.ts
+++ b/packages/worker/src/initPro.ts
@@ -6,6 +6,7 @@ export const initPro = async () => {
     scimUserServiceConfig: {
       functions: {
         saveUser: userSdk.save,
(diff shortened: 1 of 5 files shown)


packages/worker/src/initPro.ts
```diff
@@ -6,6 +6,7 @@ export const initPro = async () => {
     scimUserServiceConfig: {
       functions: {
         saveUser: userSdk.save,
+        removeUser: (id: string) => userSdk.destroy(id, undefined),
```
mike12345567, line 9: what is the `undefined` in this call? Its likely more of an issue with the `userSdk` than anything else.

mike12345567 (approved):

packages/worker/src/initPro.ts
adrinr: I was actually checking this, it is used only to check if the user is deleting himself:
```if (!env.SELF_HOSTED && !env.DISABLE_ACCOUNT_PORTAL) {
    // root account holder can't be deleted from inside budibase
    const email = dbUser.email
    const account = await accounts.getAccount(email)
    if (account) {
      if (email === currentUser.email) {
        throw new HTTPError('Please visit "Account" to delete this user', 400)
      } else {
        throw new HTTPError("Account holder cannot be deleted", 400)
      }
    }
  }```

Not sure if this can be refactored to not be needed

mike12345567: ideally - would be good if we could just refactor this so that opts an optional property in an object specifier, make it a little less confusing looking - not sure if its possible to be removed altogether.

adrinr: Changed it to use the current identity instead of passing it as params

PR #10039, opened 2023-03-15
Improve invalid API key error messaging

mike12345567:
## Description
Making it obvious that API key is invalid - error otherwise is quite cryptic. This is a very minor change - based on @jonnymccullagh experience with it, the system will output all sorts of weird errors based on it believing the user is public if they have an invalid API key, which isn't very useful.


packages/backend-core/src/middleware/authenticated.ts
```diff
@@ -48,22 +48,31 @@ async function checkApiKey(apiKey: string, populateUser?: Function) {
...
       return {
         valid: true,
         user: await getUser(userId, tenantId, populateUser),
       }
     } else {
-      throw "Invalid API key"
+      throw {
+        message:
+          "Invalid API key - may need re-generated, or user doesn't exist",
+        name: "InvalidApiKey",
```
Rory-Powell, line 74: I think this would be a great candidate to build out the errors framework. I was gonna suggest how to do this, but realised that the existing framework was unnecessarily complex and hard to understand around `type`, `code` and `context`. It wasn't that easy to extend. 

I've updated it here: https://github.com/Budibase/budibase/pull/10042

This updat makes it much easier to just drop in a new section:
```ts
// AUTH

export class InvalidAPIKey extends BudibaseError {

  constructor(message: string) {
    super(message, ErrorCode.INVALID_API_KEY)
    this.status = httpStatus
  }
}
```

The catch block can then be updated to
```ts
} else if (err?.code ===  ErrorCode.INVALID_API_KEY) {
```

PR #10042, opened 2023-03-15
Simplify errors framework

Rory-Powell:
## Description
Added to aid https://github.com/Budibase/budibase/pull/10039

Changes:
- Restructure and simplify the backend core errors lib
- It was unnecessarily complex with the concept of error `type` and error `code`
- Just use error code
- Combine errors into single file, there were each very small and scattered
- Single file makes it more clear to the purpose of `getPublicError` and easier to extend


shogunpurple (approved):
LGTM - not usually a fan of classic OOP in JS but this fits the bill well.

PR #10046, opened 2023-03-16
Budi-6683 - Implement search filters

adrinr:
## Description
Allow filtering on user get.
1. Change the static view for a lucene index
2. Add tests 

Addresses: 
- https://linear.app/budibase/issue/BUDI-6683/scim-usergroup-provisioning-endpoints


packages/backend-core/src/db/searchIndexes/searchIndexes.ts
```diff
@@ -0,0 +1,52 @@
...
+  const db = getGlobalDB()
+  let designDoc
+  try {
+    designDoc = await db.get("_design/database")
+  } catch (err: any) {
+    if (err.status === 404) {
+      designDoc = { _id: "_design/database" }
+    }
+  }
+  // this is a very specific function given that it is only for audit logs
```
Rory-Powell, line 14: comment references audit logs

packages/backend-core/src/db/searchIndexes/searchIndexes.ts
```diff
@@ -0,0 +1,52 @@
...
+  const fn = function (user: User) {
+    if (user._id && !user._id.startsWith("us_")) {
+      return
+    }
+    const ignoredFields = ["_id", "_rev", "password"]
```
Rory-Powell, line 19: Some things come to mind that can be bulky on the user objects, on some historical docs we have session state saved into the user which shouldn't be there. This could be quite a few docs. Any of these fields could be present:
https://github.com/Budibase/budibase/blob/fb7e81fbc907eb6856044ba9c9d1ae565c5d7874/packages/worker/src/api/controllers/global/self.ts#L87

The `license` and `account` in particular could be very bulky - can we exclude these session attributes for safety.

Rory-Powell: Would it be better to take a whitelist approach instead? So we can only index the expected fields we know about. There's no reason to index many things such as third party sso profile, roles, the above mentioned session attrs.

Rory-Powell (approved):
Mentioned on the other pr, I think a unit test for the new skipping behaviour lucene would be great. Other than that LGTM

Issue #10051, opened 2023-03-16
[BUDI-6761] Trigger field bindings in automation not working


joebudi:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
tenant - bb.budibase.app
app name - visitor booking app

**Describe the bug**
I had a button with an onClick action to trigger an automation (passing a rowid). 

My automation: 
- trigger (app action with the form field - rowid. 
- action - update row (see screenshot). 

When I clicked the finish and test automation button, nothing happened - no feedback and the builder crashed.

<img width="1189" alt="CleanShot 2023-03-16 at 10 14 06@2x" src="https://user-images.githubusercontent.com/49767913/225586251-46e66888-4b89-46d9-8a86-8c7ceece6034.png">

<sub>[BUDI-6761](https://linear.app/budibase/issue/BUDI-6761/trigger-field-bindings-in-automation-not-working)</sub>

PR #10052, opened 2023-03-16
Re-add code that was removed to fix App Action fields null-pointer

melohagan:
## Description
Recent code change removed a cloneDeep, but it's needed!

Addresses: 
- https://github.com/Budibase/budibase/issues/10051


PClmnt (approved):

aptkingston:
Is it actually the clone that's needed, or just always an assignment to block inputs? As we've discussed before, we shouldn't have these clones all over the place - the goal is to remove them. If the clone is actually needed then it means there's other mutation going on that we need to fix.

aptkingston:
I'll have a look at the original issue and root cause it.

melohagan:
Thanks @aptkingston !

I was going to go ahead and deploy this fix anyway just because it is a fairly bad breaking change.

aptkingston:
Yep makes sense, no problem!

PR #10064, opened 2023-03-17
BUDI-6683 - Create group

adrinr:
## Description
Create initial endpoints for groups: search and create. It follows the same logic/patterns as the scim groups

### DIFF
--- a/packages/worker/src/api/routes/global/tests/scim/groups.spec.ts
+++ b/packages/worker/src/api/routes/global/tests/scim/groups.spec.ts
@@ -1,0 +1,183 @@
+import tk from "timekeeper"
+import _ from "lodash"
+import { mocks, structures } from "@budibase/backend-core/tests"
+import { ScimCreateGroupRequest, ScimGroupResponse } from "@budibase/types"
+import { TestConfiguration } from "../../../../../tests"
+
+mocks.licenses.useScimIntegration()
+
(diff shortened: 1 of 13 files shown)


packages/worker/src/api/routes/global/tests/scim/groups.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+  return group
+}
+
+describe("/api/global/scim/v2/groups", () => {
+  let mockedTime = new Date(structures.generator.timestamp())
+
+  beforeEach(() => {
+    tk.reset()
+    mockedTime = new Date(structures.generator.timestamp())
+    tk.freeze(mockedTime)
```
Rory-Powell, line 38: we have a constant mocked timestamp already that could be used instead? it's applied by default for all tests so could remove this new freezing and import the constant from structures instead. or is there another reason for using a new timestamp for each test?

I'd like to keep data handling as as noninvasive as possible which is why I always question date manipulation in individual tests

Rory-Powell (approved):
Magnificent!

packages/worker/src/api/routes/global/tests/scim/groups.spec.ts
adrinr: Update it to use the generic mocked time and overriding it when actually needed

PR #10077, opened 2023-03-20
Handle application/pdf content type in rest integration

robert4dt:
## Description
The problem you have here is probably due to the way the code handles different content types. In the code, the content type of the API response is checked and the response is handled accordingly. There are special handlings for JSON and XML content types, while all other content types are treated as text.

In the case of a PDF file returned in raw PDF format, the content type should be application/pdf in most cases. Since the code has no special handling for this content type, it is treated as text, which results in incorrect processing of the binary code.

To fix this problem, you can add a condition to specifically handle application/pdf content type and save the binary code as a buffer. Modify the appropriate part of the code in the parseResponse method as follows.

### DIFF
--- a/packages/server/src/integrations/rest.ts
+++ b/packages/server/src/integrations/rest.ts
@@ -150,7 +150,10 @@ class RestIntegration implements IntegrationBase {
           data = data[keys[0]]
         }
         raw = rawXml
-      } else {
+      } else if (contentType.includes("application/pdf")) {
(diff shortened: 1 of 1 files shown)


robert4dt:

Rory-Powell (approved):
LGTM

Rory-Powell:
@robert4dt looks like there is a lint failure, you can fix with `yarn lint:fix`

mike12345567:
Merging to separate branch so can lint and merge to `develop`.

PR #10086, opened 2023-03-20
Update logging to support dd trace attributes

Rory-Powell:
## Description
- Updating our default console logger to integrate with `pino`
- This enables us to inherit automation trace log injection from datadog
- Add a new environment variable `HTTP_LOGGING` to control whether http requests should be logged, as previously this was controlled by the `LOG_LEVEL` attribute in combination with `pino-http` module - now we control general application logging levels with this var hence the new one. 
   - This defaults to true for the benefit of useful logs from self hosters
   - Turned off in any environment where we pay for log storage (qa / prod currently) and don't need this as much due to datadog 
- Remove lingering elastic apm refernces
- Cleanup older log lines that manually added data like tenant id to the log line
- Turn off `pino-pretty` in non development environments for easier log ingestion
- Update `bb-alerts` to no longer need to handle single-line stack traces as per above
- Add small api in server for testing in deployed environments
   -  `POST /api/ops/log` - log each of the supported log types
   - `POST /api/ops/error` - force an http error
   - `POST /api/ops/alert` - force a `bb-alert`
- Remove log overrides that were done for unit tests - now controlled by log levels only
- Add attributes to all log lines as pino [merging objects](https://getpino.io/#/docs/api?id=mergingobject)
   - correlationId
   - tenantId
   - appId
   - identityId
   - identityType
   - Note: only correlationId is added to `pino-http` logging 
- Use custom serializers for http `res` / `res` logging to reduce the giant payloads of data, making the logs more useful and preventing leaking sensitive security headers 

Related: 
- https://github.com/Budibase/budibase-pro/pull/141
- https://github.com/Budibase/budibase-deploys/pull/41
- https://github.com/Budibase/account-portal/pull/289
- https://github.com/Budibase/budibase-infra/pull/163

Addresses: 
- https://linear.app/budibase/issue/BUDI-6767/update-logging-to-support-dd-trace-attributes

## Screenshots

`/api/ops/log` with the following body:
```json
{
   "message": "hey!",
   "data": { "foo": "bar" }
}
```
Dev:
Screenshot 2023-03-21 at 23 04 23

Prod:
Screenshot 2023-03-21 at 23 11 27

`/api/ops/error` with the following body:
```json
{
   "message": "something went wrong"
}
```
Screenshot 2023-03-21 at 23 07 18

HTTP Logging

Prod:
```json
{"level":"INFO","timestamp":"2023-03-21T23:14:26.720Z","req":{"method":"GET","url":"/api/applications?status=all","correlationId":"d061a58c-d54d-4a4d-831b-fad6fa6a0efc"},"res":{"status":200},"responseTime":48,"msg":"request completed"}
```
Dev:
Screenshot 2023-03-21 at 23 21 10


packages/backend-core/package.json
```diff
@@ -39,6 +39,7 @@
     "joi": "17.6.0",
     "jsonwebtoken": "9.0.0",
     "koa-passport": "4.1.4",
+    "koa-pino-logger": "4.0.0",
```
adrinr, line 42: Is there a reason why we fix the npm versions?

packages/backend-core/src/environment.ts
```diff
@@ -36,6 +36,15 @@ function getAPIEncryptionKey() {
     : process.env.JWT_SECRET // fallback to the JWT_SECRET used historically
 }
 
+function httpLogging() {
+  if (process.env.HTTP_LOGGING === undefined) {
+    // on by default unless otherwise specified
+    return true
+  } else {
+    return process.env.HTTP_LOGGING
+  }
```
adrinr, line 45:
```suggestion
  }
  
  return process.env.HTTP_LOGGING
```

packages/backend-core/src/logging/alerts.ts
```diff
@@ -0,0 +1,26 @@
+const NonErrors = ["AccountError"]
+
+function isSuppressed(e?: any) {
+  return e && e["suppressAlert"]
+}
+
+export function logAlert(message: string, e?: any) {
```
adrinr, line 7: Can these `e` be typed?

packages/backend-core/package.json
Rory-Powell: Just as a best practice for security mostly. We also had some problems with drift between services as well, where we want worker, server and account portal to be on the same version of some lib, but the locked version have diverged

packages/backend-core/src/logging/alerts.ts
Rory-Powell: We tend to use `any` for errors, but I suppose we could use `Error` from here on out 👍

packages/backend-core/src/middleware/errorHandling.ts
```diff
@@ -9,9 +8,10 @@ export async function errorHandling(ctx: any, next: any) {
     const status = err.status || err.statusCode || 500
     ctx.status = status
 
-    if (status > 499 || env.ENABLE_4XX_HTTP_LOGGING) {
```
mike12345567, line 12: Is the `ENABLE_4XX_HTTP_LOGGING` env var gone or have I just missed its usage now? If its gone, what does that mean for self hosters utilising it

mike12345567 (approved):
Looks great - good work on this, great to finally consolidate our logging! Just one minor comment about how this could affect self hosters.

PR #10092, opened 2023-03-21
BUDI-6685 - SCIM usergroup lockdown

adrinr:
## Description
Blocking user and group edition when SCIM feature is enabled. UI changes described here: https://www.figma.com/file/jA3KI85k4eFqfoQMMtLkly/SCIM?node-id=1-111&t=c9nm6utdA3fRJ1h6-0

Addresses: 
- https://linear.app/budibase/issue/BUDI-6685/scim-usergroup-lockdown

## App Export
Not required

### Users list
image

### Users detail
image

### Groups list
image

### Group detail
image

### DIFF
--- a/packages/builder/src/pages/builder/portal/users/_components/SCIMBanner.svelte
+++ b/packages/builder/src/pages/builder/portal/users/_components/SCIMBanner.svelte
@@ -1,0 +1,15 @@
+<script>
+  import { Icon } from "@budibase/bbui"
+</script>
+
+<div class="scim-banner">
+  <Icon name="Info" size="S" />
+  Users are synced from your AD
+</div>
+
+<style>
+  .scim-banner {
+    display: flex;
+    gap: var(--spacing-s);
+  }
+</style>
--- a/packages/builder/src/pages/builder/portal/users/groups/[groupId].svelte
+++ b/packages/builder/src/pages/builder/portal/users/groups/[groupId].svelte
@@ -14,7 +14,7 @@
   } from "@budibase/bbui"
   import UserGroupPicker from "components/settings/UserGroupPicker.svelte"
   import { createPaginationStore } from "helpers/pagination"
-  import { users, apps, groups, auth } from "stores/portal"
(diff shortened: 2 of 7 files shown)


PClmnt:
image I wonder if for SCIM should we update GroupNameTableRenderer to not show an icon? Because that gap where an icon would normally be looks a bit strange. So for SCIM it would just be the name.

adrinr:
> > I wonder if for SCIM should we update GroupNameTableRenderer to not show an icon? Because that gap where an icon would normally be looks a bit strange. So for SCIM it would just be the name.
> 
> Would be nice if we could get it to create the group with an icon, just the default one we see when creating a new group would be good

This is set on the creation process, in the backend. I had it as a TODO. Let me update the screenshots

adrinr:
> > > I wonder if for SCIM should we update GroupNameTableRenderer to not show an icon? Because that gap where an icon would normally be looks a bit strange. So for SCIM it would just be the name.
> > 
> > 
> > Would be nice if we could get it to create the group with an icon, just the default one we see when creating a new group would be good
> 
> This is set on the creation process, in the backend. I had it as a TODO. Let me update the screenshots

PR adding the icon as expected: https://github.com/Budibase/budibase-pro/pull/145
@Rory-Powell @PClmnt

PClmnt:
> > > > I wonder if for SCIM should we update GroupNameTableRenderer to not show an icon? Because that gap where an icon would normally be looks a bit strange. So for SCIM it would just be the name.
> > > 
> > > 
> > > Would be nice if we could get it to create the group with an icon, just the default one we see when creating a new group would be good
> > 
> > 
> > This is set on the creation process, in the backend. I had it as a TODO. Let me update the screenshots
> 
> PR adding the icon as expected: [Budibase/budibase-pro#145](https://github.com/Budibase/budibase-pro/pull/145) @Rory-Powell @PClmnt

Awesome @adrinr! LGTM

PClmnt (approved):

PR #10108, opened 2023-03-24
allow arrow key navigation in the components panel

shogunpurple:
## Description
Just one I run into a lot. I open the component selection panel, and try to use the arrow keys to navigate, which is not possible.

I feel like `Tab` isn't that intuitive compared to using the arrow keys to navigate up and down the component selection menu.

## Documentation


Rory-Powell (approved):

aptkingston (approved):
Yep I run into this too, even though I'm the one who initially made it use tabs :roll_eyes:  nice one!

PR #10109, opened 2023-03-24
BUDI-6686 - SCIM events and auditing

adrinr:
## Description
Extending the auditing messages and events to capture when a user/group has been created/edited/removed using an SCIM endpoint.
This PR also merges both routers (users and groups) into a single one, so we don't have duplicated middlewares set up

Addresses: 
- https://linear.app/budibase/issue/BUDI-6686/scim-events-and-auditing

## App Export
- Not required

## Screenshots
image
image


packages/types/src/sdk/events/user.ts
```diff
@@ -4,20 +4,23 @@ export interface UserCreatedEvent extends BaseEvent {
   userId: string
   audited: {
     email: string
+    viaScim?: boolean
```
Rory-Powell, line 7: We could probably pull this out of the `audited` block into the main body, this would be useful to have in posthog too. The `audited` fields are removed for privacy reasons.

adrinr: 👍🏽

adrinr: I think the audited are used for creating the messages. Is there a reason on why we only use these `audited` fields? We could use the full event to populate the handlebars message, so we could move this new field into a higher level.
Is there any specific reason why it was not done like that? @mike12345567

mike12345567: The `audited` props are simply there so that certain properties can be passed to the auditing system, but removed before sending to Posthog. Some private information like say `email` should not be passed up to Posthog as it is PII.

In the case of SCIM flag, it should be top level as we want to use this for auditing (the whole structure and the `audited` section) but also want to send it to Posthog (the structure minus the `audited` section).

adrinr: Ok. This means that `AuditedEventFriendlyName` will ingest the full event and the handlebars will use `user {{audited.email}}` instead of `user {{email}}`

packages/backend-core/src/context/types.ts
```diff
@@ -6,4 +6,5 @@ export type ContextMap = {
   appId?: string
   identity?: IdentityContext
   environmentVariables?: Record<string, string>
+  scimCall?: boolean
```
adrinr, line 9: Not sure if this is the best name for it. Any ideas to make something more explicit, or is it clear enough?

packages/types/src/sdk/events/event.ts
```diff
@@ -194,9 +194,9 @@ export enum Event {
 // a user facing event or not.
 export const AuditedEventFriendlyName: Record<Event, string | undefined> = {
   // USER
-  [Event.USER_CREATED]: `User "{{ email }}" created`,
-  [Event.USER_UPDATED]: `User "{{ email }}" updated`,
-  [Event.USER_DELETED]: `User "{{ email }}" deleted`,
+  [Event.USER_CREATED]: `User "{{ email }}" created{{#if viaScim}} via SCIM{{/if}}`,
```
adrinr, line 197: This is quite dirty, and it requires a handlebars execution with empty objects to display the title used in the dropdown. If this needs to extend further, I would recommend changing the string for an object `{title:string, template:string}`. Otherwise it will become quite hard to follow

packages/backend-core/src/context/types.ts
Rory-Powell: adding `is` in front might help like `isScimCall` or just `isScim`

packages/backend-core/src/events/publishers/group.ts
```diff
@@ -25,6 +26,7 @@ async function created(group: UserGroup, timestamp?: number) {
 async function updated(group: UserGroup) {
   const properties: GroupUpdatedEvent = {
     groupId: group._id as string,
+    viaScim: context.isScimCall(),
```
Rory-Powell, line 29: personal pref
```suggestion
    isScim: context.isScim(),
```

packages/backend-core/src/context/mainContext.ts
```diff
@@ -214,6 +214,13 @@ export function doInEnvironmentContext(
   return newContext(updates, task)
 }
 
+export function doInScimContext(task: any) {
```
Rory-Powell, line 217: Could add a small unit test

```ts
  describe("doInScimContext", () => {
    it("returns true when set", () => {
      context.doInScimContext(() => {
        const isScim = context.isScimCall()
        expect(isScim).toBe(true)
      })
    })
    it("should false when not set", () => {
      const isScim = context.isScimCall()
      expect(isScim).toBe(false)
    })
  })
```

Rory-Powell (approved):
small comments - otherwise LGTM

packages/backend-core/src/events/publishers/group.ts
adrinr: That was my first approach, but we could run into some case that we are deleting a SCIM user manually. The field here should be false, even if the user is scim sync. And it might be confusing

Rory-Powell: The above suggestion was just for the naming, I don't think the settings on the user would impact it

PR #10120, opened 2023-03-27
Fix for user groups adding users to app user metadata tables

mike12345567:
## Description
There was an issue that caused user's added to groups to be added to all app user metadata tables - this includes a small fix for it.

The majority of the work in this PR was around testing, there has been issues with the server testing before now, I've updated Jest and I believe this should resolve the issue we've been seeing where the server tests haven't been running correctly/haven't been showing coverage. Updating jest also broke a few test cases (namely snapshot tests) which required updates.

Also ran into some issues with Node 16.10+ versions, it seems that we have introduced an issue with the switch to `swc` - I spent a lot of time looking into this, our tests were leaking memory, but it turns out this is a known issue with Node 16.10+ and Jest + SWC. https://github.com/facebook/jest/issues/7874#issuecomment-1057890147

There are also some other various fixes/improvements due to the way Jest expect changed a bit.

NOTE: most of the LoC changed here are from the yarn.lock being massively updated due to all the Jest library updates.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6657/a-user-in-a-group-is-incorrectly-added-to-an-apps-user-table
- #9816


packages/server/src/api/routes/tests/appSync.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      },
+    })
+  })
+
+  async function getUserMetadata() {
+    const { rows } = await config.searchRows(dbCore.InternalTable.USER_METADATA)
+    return rows
+  }
+
+  it("make sure its empty initially", async () => {
```
Rory-Powell, line 27: Does the test name need updated? Looks like it used to model a test where the results were `[]`

Rory-Powell: I would be a fan of that approach, to have an empty state + a test after sync

Rory-Powell (approved):

packages/server/src/api/routes/tests/appSync.spec.ts
mike12345567: There is no empty state anymore - the app starts in a sync'd state.

PR #10124, opened 2023-03-27
BUDI-6683 - Honor configuration flag

adrinr:
## Description
Honoring the new configuration that allows activating and deactivating the SCIM endpoints

### DIFF
--- a/packages/backend-core/src/configs/configs.ts
+++ b/packages/backend-core/src/configs/configs.ts
@@ -245,6 +245,8 @@ export async function getSMTPConfig(
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/backend-core/src/configs/configs.ts
```diff
@@ -242,3 +243,7 @@ export async function getSMTPConfig(
     }
   }
 }
+
```
Rory-Powell, line 246:
```suggestion

// SCIM

```

Rory-Powell (approved):
Minor comments - LGTM

PR #10134, opened 2023-03-28
Lowercase values for Oracle LIKE

melohagan:
## Description
Oracle LIKE statements are case sensitive, so needed to lowercase the comparison value as well. 

Addresses: 
- https://linear.app/budibase/issue/BUDI-6793/oracle-db-starts-with-filter-always-requires-lower-case-value


mike12345567 (approved):
Nice tests 👍

Issue #7016, opened 2022-08-01
add finally block in action 


wenzhixue:
![图片](https://user-images.githubusercontent.com/1578423/182130932-76fdc8be-f6b5-4ded-b42a-25e4d6d57b28.png)

wenzhixue:
I use "continue if" to check if one field is equal to true， then "execute query". if the condition triggered , the 5th action "Navigate to" will never be processed.

wenzhixue:
so two methods can solve this problem:
1: "continue if /stop if " only work on the next action
2: add "finally" block to ensure "navigate to " is always wokring

ryzam:
Any update on this issue. Also think about conditional else statement execution

IchthysMaranatha:
Please have a look here
https://github.com/Budibase/budibase/discussions/10123

PR #10135, opened 2023-03-28
Add Goto If Action

IchthysMaranatha:
Hi, have also added changes for restricting loops according to this post
https://github.com/Budibase/budibase/discussions/10123#discussioncomment-5448456
Have not built and tested after that

Address:
 - https://github.com/Budibase/budibase/issues/7016

Simple *Finally* use-case:
Screenshot 2023-03-31 at 08 45 46

goto

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ContinueIf.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ContinueIf.svelte
@@ -14,6 +14,10 @@
       label: "Stop if",
       value: "stop",
     },
+    {
+      label: "Goto if",
+      value: "goto",
+    },
   ]
   const operatorOptions = [
     {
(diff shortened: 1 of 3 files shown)


IchthysMaranatha:

melohagan (changes requested):
Add a Goto If / Else (see comment above).

IchthysMaranatha:
It's a jump only if condition is met, if not it will go to the next action by default. No need of else here. You have to configure steps accordingly to get the if then else construct. https://github.com/Budibase/budibase/discussions/10123#discussion-5009639

IchthysMaranatha:
Because if both are jumps what would be the point of the action just after the go to if action?

IchthysMaranatha:
In your case you've jumped to next step with goto if, which it will go to anyway even if condition is not met. For Goto if, go to step must always be something that is not in the normal progression, just like in any programming language that supports goto.

melohagan:
> Because if both are jumps what would be the point of the action just after the go to if action?
> In your case you've jumped to next step with goto if, which it will go to anyway even if condition is not met. For Goto if, go to step must always be something that is not in the normal progression, just like in any programming language that supports goto.

That makes sense! It probably is grand then, but I would like the product team to have a look over this as well just to be sure they're happy with the flow.

melohagan:
Hey @IchthysMaranatha 

I had a chat with the team and our product manager, and we have decided for now that this feature isn't quite what we want for the core product, but there is no time frame for when we would implement this. Possibly we could allow flexible scripts for users who need advanced action flows, but that's just an idea! 

With that said I really appreciate the contribution effort! To help us in future, what specific use-case were you trying to solve using 'Go To' actions?

Issue #9754, opened 2023-02-20
[BUDI-6615] Component Error States - Auto Bind to Data


shogunpurple:
**COS**

* When dragging a table or other data component into a tree, automatically bind that data component to the next data provider parent in the tree
* When adding a text field into a form component, automatically bind the text field to the next available text column. If you add another field, bind to the next column. Automatically use the label from this column in the form

[Automatically choose data providers when dragged into a data provider - Error states (Figma)](https://www.figma.com/file/V50fSGM2LP5iHGjJfnSMQO/Error-states?node-id=13%3A4316&t=OUjz4Zqv3VdTK2e3-0)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6615](https://linear.app/budibase/issue/BUDI-6615/component-error-states-auto-bind-to-data)</sub>

Issue #9753, opened 2023-02-20
[BUDI-6614] Component Error States - Validation UI


shogunpurple:
When I'm using the design section of the builder, I want to ensure that the builder guides me to configure my components properly, and prevent me from shipping a broken app.

### **Error States**

**Required Settings**

When any component is missing a required setting, show it as errored in the design UI. When the error is clicked, so the missing settings in the right hand setting panel (highlighted in red).

**Parent components with no children**

Examples are:

* Form with no fields
* Data provider with no children

With both of the configurations above - we do not show them as an "error" but rather guide the user to add components to them with the action bar. This is to prevent constant errors while building out apps.

**Children components missing required parents**

Form fields have to be placed inside a form component in order to be useful.

These components require a data provider as a parent to function:

* Repeater
* Table
* Dynamic Filter
* Date Range

When any component is missing a required parent, show it as errored in the design UI. Allow users to wrap the component in the required parent from the component action bar.

**COS**

* Show error state when component is misconfigured (a required setting is missing, or a missing parent such as form or data provider.)
* Show error state icon in component tree for that component
* Show warning when you try to publish an app with a "broken" component
* Add options to the action bar to direct the user to the next step - adding a component or wrapping a parent.

**Designs**

[Page 1 - Error states (Figma)](https://www.figma.com/file/V50fSGM2LP5iHGjJfnSMQO/Error-states?node-id=0%3A1&t=OUjz4Zqv3VdTK2e3-0)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6614](https://linear.app/budibase/issue/BUDI-6614/component-error-states-validation-ui)</sub>

PR #10136, opened 2023-03-28
Component error state improvements

aptkingston:
## Description
This PR adds multiple improvements that help to make it harder to get invalid component states. There are a few features which help populate required settings and some better error messages (with actions) to help get you out of error states.

Designs: https://www.figma.com/file/V50fSGM2LP5iHGjJfnSMQO/Error-states?node-id=12-3888&t=mvzTDyxko1kzFzP6-0
Note: implementation is slightly different to designs due to technical restrictions:
- Actions for wrapping components and adding component are links inside the components, rather than in the action bar
- There are no error icons in the component tree. This can maybe be solved at a later date but is not trivial due to the iframe boundary. We can duplicate the logic in future to add this.

## New features
- Components which accept child components, but have none, now have a prettier info message, which is clickable to immediately open the new component panel
image
- When adding or moving any component that has a data provider setting, this setting will be filled in if empty or validated if non empty. If you move something to a different data provider then the data provider setting will automatically update. This makes it much easier working with data heavy components.

https://user-images.githubusercontent.com/9075550/228224478-29ba1c4c-4291-4c40-83a9-488410759b2c.mp4

- When adding a form, or anything with a `schema` setting, the schema will be automatically filled in to the first available datasource, and custom if none exist. This is similar to what already happens for datasources.
- When adding form fields, the field name will be automatically filled in to the first available field that matches the component type that is not already in use. Continuing to add more fields will continue to use different fields every time, where possible. The label setting will be filled in as well. This does *not* get validated when you change form or change schema, as you are always allowed arbitrary field names - so this is never an error state.

https://user-images.githubusercontent.com/9075550/228224691-31182351-ba5c-462f-b44b-cc8e52c0e9dd.mp4

- When a component is missing a required setting, an error message will be shown allowing the user to highlight the missing setting. This functionality already existed but was broken.
image

https://user-images.githubusercontent.com/9075550/228225065-5222dd1a-3db5-44a8-b0e6-6ea8d4243976.mp4

- When a component depends on another component, a new error message will be displayed showing the user that their component needs wrapped in this other component. There is also the functionality to automatically wrap this component in that required component type, instantly solving their problem. Here's an example of adding a table before a data provider:
image

https://user-images.githubusercontent.com/9075550/228225417-ea012043-c606-4cda-8f0f-6f9e7a38b09c.mp4

Addresses: 
- #9753
- #9754

## Documentation


shogunpurple (approved):
LGTM - although would be good to see some unit tests on the store logic once we have Gerards work merged up.

PR #10137, opened 2023-03-28
Chore: Account deletion via API

Rory-Powell:
## Description

Updates: 
- Allow tenant deletion api to work when internal api key is supplies

See: https://github.com/Budibase/account-portal/pull/292

Addresses: 
- [BUDI-6794](https://linear.app/budibase/issue/BUDI-6794/account-deletion-via-api)

### DIFF
--- a/packages/worker/src/api/controllers/system/tenants.ts
+++ b/packages/worker/src/api/controllers/system/tenants.ts
@@ -5,7 +5,7 @@ export async function destroy(ctx: UserCtx) {
   const user = ctx.user!
   const tenantId = ctx.params.tenantId
 
-  if (tenantId !== user.tenantId) {
+  if (!ctx.internal && tenantId !== user.tenantId) {
     ctx.throw(403, "Tenant ID does not match current user")
   }
 


packages/worker/src/api/controllers/system/tenants.ts
```diff
@@ -5,7 +5,7 @@ export async function destroy(ctx: UserCtx) {
   const user = ctx.user!
   const tenantId = ctx.params.tenantId
 
-  if (tenantId !== user.tenantId) {
+  if (!ctx.internal && tenantId !== user.tenantId) {
```
adrinr, line 8: Are we using user tokens for internal calls?

Rory-Powell: we use the `INTERNAL_API_KEY`

shogunpurple (approved):
LGTM!

packages/worker/src/api/controllers/system/tenants.ts
Rory-Powell: In this case we want to only allow the action if the current session matches the users tenant id, or the api key is the internal api (system) api key

PR #10139, opened 2023-03-28
Chore: Update livenessprobe timeout, add readinessprobe

Rory-Powell:
## Description

- Update the liveness probe timeout to 3s to try to resolve restarts 
- Add readiness probe to prevent an unready serviced being sent http traffic to prevent 504s. 

Addresses: 
- Unticketed

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -215,8 +215,21 @@ spec:
           httpGet:
             path: /health
             port: {{ .Values.services.apps.port }}
+          initialDelaySeconds: 10
+          periodSeconds: 5
+          successThreshold: 1
+          failureThreshold: 3
+          timeoutSeconds: 3
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: {{ .Values.services.apps.port }}
           initialDelaySeconds: 5
           periodSeconds: 5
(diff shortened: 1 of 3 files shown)


charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -212,11 +212,24 @@ spec:
         image: budibase/apps:{{ .Values.globals.appVersion }}
         imagePullPolicy: Always
         livenessProbe:
+          httpGet:
+            path: /health
+            port: {{ .Values.services.apps.port }}
+          initialDelaySeconds: 10
```
shogunpurple, line 218: Is this too long? A pod will need at least 10 seconds to start essentially? Would `5` work here even?

shogunpurple (approved):

adrinr (approved):

charts/budibase/templates/app-service-deployment.yaml
Rory-Powell: This is for the liveness, which affects only killing a container after it's started. 
The readinessprobe below is what determines when to start sending traffic. I intentionally made the delay of the liveness to be longer than the readiness, which is set to 5

shogunpurple: super, all good then!

PR #10140, opened 2023-03-28
GSheet logout possible fix

mike12345567:
## Description
Adding error to check why logout might have occurred as well as a possible fix.

### DIFF
--- a/packages/builder/src/pages/builder/_layout.svelte
+++ b/packages/builder/src/pages/builder/_layout.svelte
@@ -41,13 +41,13 @@
         return
       }
 
-      if (user.tenantId !== urlTenantId) {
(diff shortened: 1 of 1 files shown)


Rory-Powell (approved):

packages/builder/src/pages/builder/_layout.svelte
```diff
@@ -41,13 +41,13 @@
         return
       }
 
-      if (user.tenantId !== urlTenantId) {
+      if (urlTenantId && user.tenantId !== urlTenantId) {
```
Rory-Powell, line 44: Should be safe on the `urlTenantId` check, given the above block will return if it doesn't exist

PR #10149, opened 2023-03-29
Budicloud logout on googlesheet auth

mike12345567:
## Description
This is a fix for the googlesheets issue with QA-wolf and QA have been experiencing in the budicloud environment.

This works by checking if the tenant ID which it has retrieved is valid, if it is not then it doesn't execute any logic based on it, specifically around logging the user out for not being in the correct tenant.

This was an issue which stemmed from environments with hosts like `something.qa.host.net` - where we would consider the `something` to be the tenant ID.

This method is safer than utilising the `platformUrl` as it must be setup, if it isn't set the user will experience weird log outs like this, checking tenant validity is the safest way to confirm if it is a good idea to logout the user or not.


shogunpurple (approved):
LGTM! Nice work on the fix - glad we can verify in QA and preprod as well.

packages/worker/src/sdk/tenants/tenants.ts
```diff
@@ -74,3 +74,10 @@ async function removeTenantUsers(tenantId: string) {
     throw err
   }
 }
+
+export async function tenantInfo(tenantId: string) {
+  const globalDbName = tenancy.getGlobalDBName(tenantId)
+  return {
+    exists: await dbCore.dbExists(globalDbName),
```
Rory-Powell, line 81: This might give us a false positive. e.g. in qa, because of the db auto creation we end up with a db `budicloud` 

Screenshot 2023-03-30 at 10 31 56

Rory-Powell: nevermind, just seen the follow up pr 👍

PR #10153, opened 2023-03-29
BUDI-6803 - Remove users that should not be there anymore when syncGlobalUsers

adrinr:
## Description
We have some actions (such as deleting a group from an app) that are meant to delete users permissions from app, but they are never removed. This means that the users are still assigned to the app on their internal tables.

This PR is modifying the `syncGlobalUsers` function that runs on app load to not just add any missing user, but to remove them if required.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6803/bug-caught-on-mishoiskambudicloudqabudibasenet


packages/server/src/sdk/users/tests/utils.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  })
+
+  it("admin and builders users are synced", async () => {
+    const user1 = await config.createUser({ admin: true })
+    const user2 = await config.createUser({ admin: false, builder: true })
+    await config.doInContext(config.appId, async () => {
+      await syncGlobalUsers()
+
+      const metadata = await rawUserMetadata()
+      expect(metadata).toHaveLength(3)
```
Rory-Powell, line 38: Might be nice to have an assertion before and after the sync

packages/server/src/sdk/users/tests/utils.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        expect(metadata).toContainEqual(
+          expect.objectContaining({
+            _id: db.generateUserMetadataID(user2._id),
+          })
+        )
+      })
+    })
+  })
+
+  it("app users are removed when user removed from user group", async () => {
```
Rory-Powell, line 101: would this be more accurate as "when app is removed from user group"

Rory-Powell: it would be nice to have a test for the current description though, where we remove users from the group instead of removing apps

Rory-Powell (approved):

packages/server/src/sdk/users/tests/utils.spec.ts
adrinr: Good catch. Done

PR #10162, opened 2023-03-30
Bug - BUDI-6782 - Internal table all row ids for

adrinr:
## Description
Fixing an issue with importing internal data. The current behaviour is that the auto ids are all set to 1, causing multiple problems. This PR fixes to assign these rows consecutive ids.

This issue was affecting both test data creation and importing data

Addresses: 
- https://linear.app/budibase/issue/BUDI-6782/internal-table-all-row-ids-for-relationship-row-in-table-are-1

## App Export
- Any exported and reimported data set can reproduce the issue

## Screenshots
Not required

## Documentation

### DIFF
--- a/packages/server/src/api/controllers/table/utils.ts
+++ b/packages/server/src/api/controllers/table/utils.ts
@@ -99,16 +99,17 @@ export function makeSureTableUpToDate(table: any, tableToSave: any) {
   return tableToSave
 }
 
-export function importToRows(data: any, table: any, user: any = {}) {
+export function importToRows(data: any, table: Table, user: any = {}) {
   let finalData: any = []
   for (let i = 0; i < data.length; i++) {
     let row = data[i]
-    row._id = generateRowID(table._id)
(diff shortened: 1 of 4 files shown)


packages/server/src/api/controllers/table/utils.ts
```diff
@@ -99,32 +105,35 @@ export function makeSureTableUpToDate(table: any, tableToSave: any) {
...
     let row = data[i]
-    row._id = generateRowID(table._id)
+    row._id = generateRowID(table._id!)
     row.tableId = table._id
-    const processed: any = inputProcessing(user, table, row, {
+    const processed = inputProcessing(user, table, row, {
       noAutoRelationships: true,
     })
     row = processed.row
+    table = processed.table
```
adrinr, line 122: This is the actually required line to fix, the rests are tests and typings

mike12345567 (approved):
Great test additions, great work @adrinr !

PR #10165, opened 2023-03-30
Explictly check for google datasource configured

Rory-Powell:
## Description
Use the specific helper for `getGoogleDatasourceConfig` for determining the existence of the config, rather than relying on the db google config only. 

also remove `isDev` check so that local dev mimics production in this case. 

Addresses: 
- Adhoc issue


shogunpurple (approved):
LGTM! Looks like the tests just need updated to reflect.

PR #10172, opened 2023-03-30
Googlesheets onboarding - fix in cloud, disable self host

mike12345567:
## Description
This addresses:
https://linear.app/budibase/issue/BUDI-6797/using-google-sheets-in-onboarding-breaks-onboarding

The onboarding flow did not account for the requirement of googlesheets to authorize with the user the ability to access the sheet.

The solution to this is two fold:
1. On self host I've disabled the ability to onboard with a google sheet as its impossible for the user to have configured the required Google SSO settings at that point. Rather than walking the user through a separate flow we simply hide the option.
2. In the cloud I've updated it so that instead of a continue button, the user sees the google authentication button, which will open in a new tab, go through the flow and eventually send the user back to the app with the sheet configured.

We have to open the authorization in a new tab as that is the expected flow, however the onboarding tab at this point becomes useless - I've a screen to simply tell the user what the next steps are, we can't really direct them to the app without a refresh cycle, but the app will open directly in the screen which they have authorized in, so it doesn't seem like it matters too much. Open to feedback on this however.

## Screenshots
Disabled button
image

Enabled button
image

Continue screen
image

Authorizing
image

App ready
image


Rory-Powell:
> We have to open the authorization in a new tab as that is the expected flow

Any reason we can't just go straight to google in the current tab?

packages/shared-core/src/helpers/integrations.ts
```diff
@@ -0,0 +1,5 @@
+import { SourceName } from "@budibase/types"
+
+export function isGoogleSheets(type: SourceName) {
```
Rory-Powell, line 3: Such a good example of having the types package available in the client

packages/builder/src/pages/builder/portal/apps/onboarding/index.svelte
```diff
@@ -74,6 +78,10 @@
       const newPlusIntegrations = {}
 
       Object.entries($integrations).forEach(([integrationType, schema]) => {
+        // google sheets not available in self-host
+        if (helpers.isGoogleSheets(integrationType) && !cloudHosted) {
```
Rory-Powell, line 82: Could change the `cloudHosted` check to use the new field for `googleDatasourceConfigured` from the organisation store. 

It won't change much considering it's v unlikely sheets will be configured but it'd be nice to be consistent. Examples where it would produce a more accurate result:

- A self host install that's running for the first time and has google env vars - likely QA 
- A cloud install that doesn't have the google env vars - likely QA or feature branch.

Rory-Powell (approved):
minor comments - LGTM

mike12345567:
> > We have to open the authorization in a new tab as that is the expected flow
> 
> Any reason we can't just go straight to google in the current tab?

@Rory-Powell - I tried this, doing a `window.open` without setting a new tab target, but it seems to me ignored, it might be a built in chromium etc feature to detect when you are going to Google auth page and automatically new tab it, not sure how it works but I couldn't force it in same tab.

PR #10173, opened 2023-03-30
Builder Unit Test PoC

Ghrehh:
## Description
Adds some unit tests to the following parts of the builder:

- History store
- Editable Label component
- Admin store
- Backups store

I opted to use [Vitest](https://vitest.dev/) as it works mostly out of the box by consuming our existing Vite config. If we instead used something like Jest we'd probably need to replicate some portion of our transpilation pipeline in Webpack, so this is greatly preferable. Vitest also has a more or less identical API to Jest, so it should look very familiar if you're more used to the later.

Similar to Jest and other unit testing frameworks, Vitest has the ability to mock certain imports used by a file. This is useful as you can end up testing the functionality of a 3rd party package. or worse: duplicating the test logic of a file you're consuming that you've already written tests for. 

We're extensively using mocks in the history store tests, especially for things relating to svelte stores. In hindsight It might have been better to just not mock out the functionality of the base svelte stores, as the tests are rather fragile to change at the moment and are full of mocked out implementation details. However taking this route can have its own issues in my experience with making it hard or impossible to test certain things atomically. If there's a series of side effects, for instance loading flags, without mocks you can often only test the state a function or class _settles_ in after your calls, instead of the series of steps it takes to get to that final state. The history store is a very complicate case however.

Simpler are the tests for the backups store, which more or less just test that the provided parameters are passed through to API calls as expected. 

The admin store is fairly complex, but not as bad as the history store, and also makes use of mocking, particularly of note is that is derives the admin store, and without mocking, the logic of this store would also need to be tested here, as well as that file possibly having it's own tests in future.

Finally there is also a test for a svelte component, `EditableLabel` I picked this one as it has some basic conditional display logic that makes it a good test case. To facilitate testing Svelte we included [Testing Library](https://testing-library.com/). The tests are fairly straight forward, but highlight another issue: there's no easy, out of the box way to mock svelte components. Meaning that the issue highlighted above, duplicating test logic, is a major cause for concern here. Mocking components in React is trivial, but it's seemingly a more complex issue for svelte. Even copying a compiled component and trying to stub using that didn't seem to work, and would probably be very fragile anyway. In short, if we plan on writing more Svelte component unit tests in future this is something we should probably aim to figure out or it will lead to issues.

The next best step in my opinion is probably to add tests for helpers, these functions are normally pretty straight forward and don't have many side effects, meaning that the tests should be resilient to change and fairly easy to write.

Addresses: 
https://linear.app/budibase/issue/BUDI-6557/research-spike-investigate-frontend-unit-tests

### DIFF
--- a/packages/bbui/src/Label/Label.svelte
+++ b/packages/bbui/src/Label/Label.svelte
@@ -9,6 +9,7 @@
 
 <TooltipWrapper {tooltip} {size}>
   <label
+    data-testid="label"
     class:muted
     for=""
     class={`spectrum-FieldLabel spectrum-FieldLabel--size${size}`}
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -56,6 +56,7 @@ jobs:
         run: yarn install:pro $BRANCH $BASE_BRANCH
       - run: yarn
       - run: yarn bootstrap
+      - run: yarn build
       - run: yarn test
       - uses: codecov/codecov-action@v3
         with:
--- a/packages/bbui/src/Typography/Body.svelte
+++ b/packages/bbui/src/Typography/Body.svelte
@@ -9,6 +9,7 @@
 </script>
 
 <p
+  data-testid="typography-body"
   style={`
     ${weight ? `font-weight:${weight};` : ""}
     ${textAlign ? `text-align:${textAlign};` : ""}
--- a/packages/bbui/src/Typography/Heading.svelte
+++ b/packages/bbui/src/Typography/Heading.svelte
@@ -9,6 +9,7 @@
 </script>
 
 <h1
+  data-testid="typography-heading"
   style={textAlign ? `text-align:${textAlign}` : ``}
   class:noPadding
   class="spectrum-Heading spectrum-Heading--size{size} spectrum-Heading--{weight}"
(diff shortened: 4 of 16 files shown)


packages/bbui/src/Label/Label.svelte
```diff
@@ -9,6 +9,7 @@
 
 <TooltipWrapper {tooltip} {size}>
   <label
+    data-testid="label"
```
Rory-Powell, line 12: It strikes me as a bit strange that `testing-library` out of the box uses it's own attribute for selection instead of giving you the option to query by `id`, which I've seen other ui testing libraries use. 

If we did want to use `id` instead of adding an attributes dedicated to tests if seems possible using this: https://testing-library.com/docs/queries/bytestid/#overriding-data-testid

aptkingston (approved):
Super work Gerard! Great foundation for being able to write additional frontend tests on top of. I think going forward we should focus on building out store tests and keeping the majority of our logic there, but it's great to be able to write DOM level tests as well if we want to.

History story is a great example of a pure, atomic store to test (and contains complicated logic) so nice one for writing tests for it.

packages/bbui/src/Label/Label.svelte
Ghrehh: I think a lot of the more "modern" testing approaches recommend this, or at least I've seen Jest and Cypress recommend it also. I think the idea is that having a totally separate attribute to query on is the most resilient to change, as ids actually have a use outside of the test for stuff like inputs so it's technically relying on an implementation detail 🤷

PR #10181, opened 2023-03-31
Minor googlesheets onboarding improvement

mike12345567:
## Description
Minor improvement - opening the google authorization in the same tab as onboarding rather than creating a dead tab.

Also updated the login page to have the OIDC and Google buttons send the user to the same page, rather than opening a new tab, as the login page becomes a dead tab otherwise.

Suggested by @Rory-Powell


shogunpurple (approved):
Ohh this is a nice one! Best to leave this out for release IMO as it could cause a few issues. Let's get it rolled next week :)

Rory-Powell (approved):
LGTM!

PR #10194, opened 2023-04-03
Chore - Update lerna and use nx

adrinr:
## Description
Updating Lerna to the latest version and configuring NX to cache `build` and `test`.
Note that the new Lerna version doesn't display outputs by default. This requires adding the `--stream` flag.

This will allow setting up NX cloud, providing a shared cache that can drastically reduce pipeline times

## Screenshots
Output of first `build` run:
image

Second run, when everything has been cached:
image


.python-version
```diff
@@ -1 +1 @@
-3.11.1
+3.10.0
```
adrinr, line 1: 3.11 is causing some strange node-gyp issue on M1s

packages/server/package.json
```diff
@@ -99,7 +99,7 @@
     "node-fetch": "2.6.7",
     "open": "8.4.0",
     "pg": "8.5.1",
-    "pino-pretty": "4.0.0",
+    "pino-pretty": "5.1.3",
```
adrinr, line 102: This was causing a strange build issue. The "breaking change" between 4 and 5 is that 5 adds TS types

Rory-Powell (approved):

PR #10195, opened 2023-04-03
Chore - Use yarn workspaces

adrinr:
## Description
Budibase relies on lerna to manage the dependencie between package, using `lerna bootstrap` and `lerna link`. These are old commands that are going to be deprecated in favour of the yarn/npm/pnpm equivalent. https://github.com/lerna/lerna/discussions/3410. Because of this, we will be facing some issues when trying to use some modern tools like esbuild, and even with submoduling.

This PR changes the way that the packages are link with each other, and the way that the npm dependencies are handled.
Using the Lerna's feature `"useWorkspaces": true` with `"npmClient": "yarn"` we will be using yarn workspaces instead of the lerna links.
This means that the new node_module hoisting structure will be used.


package.json
```diff
@@ -23,11 +22,10 @@
   },
   "scripts": {
     "setup": "node ./hosting/scripts/setup.js && yarn && yarn bootstrap && yarn build && yarn dev",
-    "bootstrap": "lerna bootstrap",
-    "postbootstrap": "lerna link && ./scripts/link-dependencies.sh",
+    "bootstrap": "./scripts/link-dependencies.sh",
```
adrinr, line 25: This is still needed to link the pro and account-portal packages. It will not be needed when everything is a proper submodule and managed via yarn workspace

packages/cli/package.json
```diff
@@ -9,7 +9,7 @@
   "author": "Budibase",
   "license": "GPL-3.0",
   "scripts": {
-    "prebuild": "rm -rf prebuilds 2> /dev/null && cp -r node_modules/leveldown/prebuilds prebuilds",
+    "prebuild": "rm -rf prebuilds 2> /dev/null && cp -r ../../node_modules/leveldown/prebuilds prebuilds",
```
adrinr, line 12: This might brake in the future if the hosting changes. TODO: check how we can use the actual used package

package.json
```diff
@@ -23,11 +22,10 @@
   },
   "scripts": {
     "setup": "node ./hosting/scripts/setup.js && yarn && yarn bootstrap && yarn build && yarn dev",
-    "bootstrap": "lerna bootstrap",
-    "postbootstrap": "lerna link && ./scripts/link-dependencies.sh",
+    "bootstrap": "lerna link && ./scripts/link-dependencies.sh",
```
adrinr, line 25: The `link-dependencies` is still needed to link the pro and account-portal packages. It will not be needed when everything is a proper submodule and managed via yarn workspace.
The `lerna link` is still required because the function `getComponentLibraryManifest` is still hardcoding the `node_modules` path.

We will remove these commands in further iterations

package.json
```diff
@@ -23,11 +22,10 @@
...
     "build": "lerna run --stream build",
     "build:dev": "lerna run --stream prebuild && tsc --build --watch --preserveWatchOutput",
-    "backend:bootstrap": "./scripts/scopeBackend.sh 'lerna bootstrap' && yarn run postbootstrap",
+    "backend:bootstrap": "./scripts/scopeBackend.sh && yarn run bootstrap",
```
Rory-Powell, line 28: We can probably remove this. I added it while trying to get the qa core suite to succeed with _only_ the backend packages. However I ended up going with the full bootstrap as parts of the API don't work without everything.

Rory-Powell (approved):
LGTM! 

Could you explain this comment from the description:
> This means that the new node_module hoisting structure will be used.

Not familiar with how workspaces changes things.

package.json
adrinr: You mean the full command?

adrinr:
> LGTM!
> 
> Could you explain this comment from the description:
> 
> > This means that the new node_module hoisting structure will be used.
> 
> Not familiar with how workspaces changes things.

I will do a demo or similar about all these changes, but in short:
- package/A has a dependency of dep1@^1.0.0
- package/B has a dependency of dep1@1.2.0
- package/C has a dependency of dep1@0.6.2

Without workspaces, you will have the following structure when you run `yarn install`:
- package/A/node_modules/dep1, with any version ^1.0.0
- package/B/node_modules/dep1, version 1.2.0
- package/C/node_modules/dep1, version 0.6.0

Yarn workspaces move the dependencies to the root. In this case, we will have:
- node_modules/dep1, version 1.2.0
- package/C/node_modules/dep1, with any version 0.6.0

Because node tries to resolve the dependencies first at ./node_modules, then ../node_modules, ../../node_modules and so on, packages A and B will be "sharing" the dependencies.

The main benefit is a reduced node_module size, but because of this hoisting I needed to fix some dependencies, specially around jest

PR #10209, opened 2023-04-05
Run integration suite in CI again / auto detect tenancy / refactors  

Rory-Powell:
## Description

**Test target**

Use a development env instead of using the custom env that was put together for cypress. This has a few benefits:
- Brings in minio, nginx and redis for the integration run
- Allows for running single tests locally against a local dev env 
- Less variations of how we run budibase

To do this, we:
- Bootstrap and build the backend packages only
- Use a new command `dev:built` which will run the built code for server and worker from the repo root - this uses the `run:docker` scripts in each package. This command also runs the same setup files that we use in local dev to spin up docker services and create environment files. 

**QA core**

A number of changes have been made to the configuration elements of the suites:
- Move jest config out to `jest.config.ts` 
- **Commands**: Rewrite the package.json scripts to be more re-use friendly, all ultimately using a single command `test` and building on it. See below. 
- **Restructure**: change from `config/foo, tests/foo` to `foo/config, foo/tests` - this makes imports significantly shorter and easy to work with. e.g:
```
├── account-api
│   ├── api
│   └── index.ts
├── internal-api
│   ├── api
│   ├── config
│   ├── fixtures
│   ├── index.ts
│   └── tests
├── public-api
│   ├── api
│   ├── config
│   ├── fixtures
│   ├── index.ts
│   └── tests
```
- **Consistent naming**: For all files that export a default class, update the filename to use that class name and use consistent casing. e.g. `default class AppAPI` lives in `AppAPI.ts`
- **API wrappers**: Decouple apis from configs by introducing a dedicated api class for each service. There are now reusable apis: `AccountInternalAPI`, `BudibaseInternalAPI`, `BudibasePublicAPI`
- **Shared test config**: Add shared test configuration `BudibaseTestConfiguration`
   - This is inherited by the test configuration in each api suite
   - Common beforeAll to remove duplicated setup logic in each suite
- **State**: Introduce the concept of `State`. This is an object that lives in the base configuration and a reference is passed to each api client. This removes all lines that manually update the state of each individual api client. e.g. setting the table id on the rows api manually is no longer needed - just set the state on your config. Most state manipulation is done automatically for apps and users in the base setup. 
- **Auto detect tenancy**: Detect the environment of the target - we now hit the `/api/system/env` at setup time to determine of the target is multi or single tenant, switching the authentication approach appropriately. 

~~**Global setup**: `globalSetup` / `globalTeardown` - these files run once in the entire suite. The global setup creates a **single account** when in multi tenant mode, previously we created 16. The teardown is currently a no-op, but there is room for expansion if we want to do test cleanup to delete the account - this would be useful if we run against production.~~
- Removed temporarily to allow for bypassing app limit in qa new accounts - will be re-added when licensing is added
 
**QA core package commands**
- `setup` - install deps, write a (now ignored) `.env` file with the default configuration for ci
   - the `.env` command is loaded by the suite the same way as our services do in dev - we no longer need  the `env-cmd`.
- `test` - run all tests, write results 
   - This command can be used in local dev to run against the locally running env
   - The only change needed will be to add user credentials to the `.env` file if running against self host
- `test:watch` - unchanged
- `test:debug` - unchanged
- `test:smoke` - runs tests excluding `dataSources`
- `test:ci` - starts the local services using `dev:build` and then runs `test:smoke` 
- `dev:built` - starts the local services using the built js


Rory-Powell:
@Pedro-Revez-Silva @shogunpurple there are a lot of changes but it's mostly from the file moves, the PR description should give context of all the main changes

Pedro-Revez-Silva (commented):
The changes to the scripts within qa-core are fine, if it makes it easier to run that is ok.
I don't understand the changes to the test structure at all. The folder structure changed, but otherwise seems the same, just with different file names.

Pedro-Revez-Silva (approved):

Issue #10155, opened 2023-03-29
[BUDI-6812] 400 errors are swallowed in data UI connectors


shogunpurple:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Happening in cloud and self host

**Describe the bug**
When a validation rule is broken at the database level and I get a 400 error from the backend, the budibase data UI swallows the error and shows no output.

**To Reproduce**
1. Go to data UI and connect to a postgres database.
2. Save or edit a row that breaks either referential integrity or a schema rule on the DB (not NULL etc)
3. No error shown

**Expected behavior**
An error is shown on the field that caused the error. 

**Screenshots**

https://user-images.githubusercontent.com/11256663/228615142-42465c9b-787d-4413-b391-d26462160430.mov

  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 OS: OSX
Browser Chrome
Version 109.0.5414.119 (Official Build) (arm64)

**Additional context**
I think the issue here may be that validation errors in joi or the frontend are related to the fields that they are in, whereas the 400 error from the backend is not associated with a particular field, and isn't caught by the `fetch` call.

<sub>[BUDI-6812](https://linear.app/budibase/issue/BUDI-6812/400-errors-are-swallowed-in-data-ui-connectors)</sub>

PR #10227, opened 2023-04-05
Fixing #10155 - some options errors were being swallowed and in gener…

mike12345567:
## Description
Fixing #10155 - some options errors were being swallowed, as well as errors coming from the underlying database that we can't handle being swallowed.

In general the UI assumed that we handled all data errors, so they were all marked as handled, but in reality a lot of them didn't have any case for being handled, so should just present a error notification instead.

## Screenshots
image
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
@@ -27,21 +27,22 @@
       notifications.success("Row saved successfully")
       dispatch("updaterows")
     } catch (error) {
-      if (error.handled) {
-        const response = error.json
-        if (response?.errors) {
-          errors = response.errors
-        } else if (response?.validationErrors) {
-          const mappedErrors = {}
-          for (let field in response.validationErrors) {
-            mappedErrors[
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
```diff
@@ -27,21 +27,22 @@
...
-          const mappedErrors = {}
-          for (let field in response.validationErrors) {
-            mappedErrors[
-              field
-            ] = `${field} ${response.validationErrors[field][0]}`
-          }
-          errors = mappedErrors
+      const response = error.json
+      if (error.handled && response?.errors) {
+        console.error("FIRST")
```
shogunpurple, line 32: remove

shogunpurple (approved):
LGTM - good to have this in!

packages/builder/src/components/backend/DataTable/modals/CreateEditRow.svelte
mike12345567: Woops sorry - some debugging left, thanks!

Issue #5302, opened 2022-04-07
Handling deletion of display columns ( 'Use as table display column' )


melohagan:
**Describe the feature request**
It's a little frustrating when trying to delete a display column. 

The app allows you to Click **Delete**, type in the column name for confirmation, click **Delete Column** and then tells you that it can't be deleted. 
Maybe hide the delete button when *Use as table display column* is checked?

Or alternatively it could allow you to delete, and automatically set the next column over as the display column?

PR #10234, opened 2023-04-06
Assign a new display column when deleting field

melohagan:
## Description
Previously an error toast would display when trying to delete the display column. This was annoying because you would have to untick the display column, save, then edit again to delete.

Now you can delete display columns and a new display column will automatically be chosen.

I also fixed a minor bug caused by a typo `name !== field` was always true. Updated to `name !== field.name`. This fixes a bug were the very first column could not be unselected as the display column (because it was always the first column).

Addresses: 
- https://github.com/Budibase/budibase/issues/5302

## Screenshots
**BEFORE**
del_disp_before
*What a hassle!*

**AFTER**
del_disp


packages/builder/src/stores/backend/tables.js
```diff
@@ -90,6 +90,23 @@ export function createTablesStore() {
     await fetch()
   }
 
+  const assignDisplayColumn = ({
```
shogunpurple, line 93: Now that we have `vitest`, I think it would be good to see a unit test for this.

shogunpurple (commented):
Functionality looks good to me once a unit test is added!

packages/builder/src/stores/backend/tables.js
melohagan: Unit tests added for the tables store

mike12345567 (commented):
Brilliant set of tests, above and beyond what was required - great work!

Just one question about the way a random column is picked, I think we need some sort of type checking, so that we don't pick a display column which isn't valid.

packages/builder/src/stores/backend/tables.js
```diff
@@ -90,6 +90,23 @@ export function createTablesStore() {
...
+    draft,
+    field,
+    originalName,
+  }) => {
+    if (primaryDisplay) {
+      draft.primaryDisplay = field.name
+    } else if (draft.primaryDisplay === originalName) {
+      const fields = Object.keys(draft.schema)
+      // pick another display column randomly if unselecting
+      draft.primaryDisplay = fields.filter(
```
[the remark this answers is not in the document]
melohagan: Added a check to ignore attachment, json and link types.

scripts/releaseHelmChart.js
```diff
@@ -1,28 +1,34 @@
-const yaml = require("js-yaml")
```
mike12345567, line 1: I assume these changes are just linting?

mike12345567 (approved):
LGTM - nice test cases :)

scripts/releaseHelmChart.js
melohagan: Yes just linting

PR #10245, opened 2023-04-08
Type updates for loading new plans

Rory-Powell:
## Description
Corresponding type updates for: https://github.com/Budibase/account-portal/pull/301

### DIFF
--- a/packages/types/src/sdk/licensing/quota.ts
+++ b/packages/types/src/sdk/licensing/quota.ts
@@ -55,7 +55,13 @@ export const isConstantQuota = (
   return quotaType === QuotaType.CONSTANT
 }
 
-export type PlanQuotas = { [ key in PlanType]: Quotas | undefined }
+export interface Minimums {
+  users: number
+}
+
+export type PlanMinimums = { [key in PlanType]?: Minimums}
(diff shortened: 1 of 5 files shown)


packages/types/src/sdk/licensing/quota.ts
```diff
@@ -54,7 +55,13 @@ export const isConstantQuota = (
   return quotaType === QuotaType.CONSTANT
 }
 
-export type PlanQuotas = { [ key in PlanType]: Quotas | undefined }
+export interface Minimums {
+  users: number
+}
+
+export type PlanMinimums = { [key in PlanType]?: Minimums}
```
adrinr, line 62: Will we have type plans without minimums? If not, the `?` should be removed to use type safety

Rory-Powell: Yeah we will, the existing pro and team plans that are day pass based. They have no concept of minimum. There is some protection around retrieving the minimum for any plan type, where we'll throw if the minimum isn't found for the given plan type. We shouldn't ever hit this code path for those plans

adrinr (approved):

PR #10257, opened 2023-04-11
Fix Help Menu

Ghrehh:
@AndyWhann pointed out there was an issue with the help menu cutting off the email button.

This is caused by a hard max-height being enforced by the popover, so I added an override. I'm unsure if there's a cleaner way to do this, if so please let me know.

### DIFF
--- a/packages/bbui/src/Actions/position_dropdown.js
+++ b/packages/bbui/src/Actions/position_dropdown.js
@@ -10,7 +10,7 @@ export default function positionDropdown(element, opts) {
 
   // Updates the position of the dropdown
   const updatePosition = opts => {
-    const { anchor, align, maxWidth, useAnchorWidth, offset = 5 } = opts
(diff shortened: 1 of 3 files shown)


aptkingston (commented):
Should we make this a bit more generic and reusable? The max height (240px) is applied by the `positionDropdown` action, designed to limit the max height of popovers when we flip them and render them above the anchor component due to lack of screen space.

We could instead:
- Add a `maxHeight` param to `positionDropdown`, and use it in place of the fixed `240px` max height when inverting the dropdown
- Add a `maxHeight` param to `Popover.svelte` and pass it in to the `positionDropdown` action
- Specify some large max height inside for the popover inside the help menu

This would make this feature usable by other components in case there are other situations in future when we need similar functionality.

aptkingston (approved):
LGTM!

PR #10264, opened 2023-04-11
Qa core datasource api tests

Pedro-Revez-Silva:
## Description
Implements CRUD and query tests for datasources (MySQL, PostgresSQL, MongoDB, REST API).

It requires new environment variables for the datasources' information and we may need to specify the CI commands so we make sure to only run the datasource's tests when the environment has access to these datasources.

```
BB_ADMIN_USER_EMAIL=
BB_ADMIN_USER_PASSWORD=
COUCH_DB_URL=
COUCH_DB_USER=
COUCH_DB_PASSWORD=
JWT_SECRET=
BUDIBASE_SERVER_URL=
BUDIBASE_URL= 
ACCOUNT_PORTAL_URL=

# PostgreSQL
POSTGRES_HOST=
POSTGRES_PORT=
POSTGRES_DB=
POSTGRES_USER=
POSTGRES_PASSWORD=

# MongoDB
MONGODB_CONNECTION_STRING=
MONGODB_DB=

# REST API
REST_API_BASE_URL=
REST_API_KEY=

#MariaDB
MARIADB_HOST=
MARIADB_PORT=
MARIADB_DB=
MARIADB_USER=
MARIADB_PASSWORD=
```

Addresses: 
- https://linear.app/budibase/issue/QA-15/qa-core-sql-server-tests
- https://linear.app/budibase/issue/QA-12/qa-core-mysql-tests
- https://linear.app/budibase/issue/QA-14/qa-core-mongo-tests
- https://linear.app/budibase/issue/QA-7/qa-core-postgres-tests


qa-core/src/internal-api/api/apis/DatasourcesAPI.ts
```diff
@@ -0,0 +1,92 @@
...
+  async getIntegrations(): Promise<[Response, any]> {
+    const [response, json] = await this.client.get(`/integrations`)
+    expect(response).toHaveStatusCode(200)
+    const integrationsCount = Object.keys(json).length
+    expect(integrationsCount).toBe(16)
+    return [response, json]
+  }
+
+  async getAll(): Promise<[Response, Datasource[]]> {
+    const [response, json] = await this.client.get(`/datasources`)
```
Rory-Powell, line 22: I think every root api path should be it's own api, which helps with more consistent naming e.g.
```ts
config.api.datasources.getAll()
config.api.integrations.getAll()
config.api.queries...
config.api.permissions...
```
etc.

qa-core/src/internal-api/fixtures/datasources.ts
```diff
@@ -0,0 +1,57 @@
...
+      name: "MariaDB",
+      plus: true,
+      source: "MYSQL",
+      type: "datasource",
+      config: {
+        database: process.env.MARIADB_DB,
+        host: process.env.MARIADB_HOST,
+        password: process.env.MARIADB_PASSWORD,
+        port: process.env.MARIADB_PORT,
+        schema: "public",
```
Rory-Powell, line 51: `process.env.MARIA_DB_SCHEMA`?

qa-core/src/internal-api/fixtures/datasources.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const postgresSQL = () => {
+  return {
+    datasource: {
+      name: "PostgresSQL",
+      plus: true,
+      source: "POSTGRES",
+      type: "datasource",
+      config: {
+        database: process.env.POSTGRES_DB,
```
Rory-Powell, line 26: Might be a good idea to add these new env vars to the root `environment.ts` file so we can see all env vars used by the suite at a glance

qa-core/src/internal-api/tests/dataSources/postgresSQL.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    // Update data source
+    const newDataSourceInfo = {
+      ...dataSourceJson.datasource,
+      name: "PostgresSQL2",
+    }
+    const [updatedDataSourceResponse, updatedDataSourceJson] =
+      await config.api.datasources.update(newDataSourceInfo)
+
+    const dataSourceQuery = {
```
Rory-Powell, line 31: For this and other `dataSourceQuery` vars we could probably pull out to a `fixtures.queries` helper as there looks to be some overlap. We could then use like

```ts
fixures.queries.query(datasourceId, fields)
```

could even add some nice helpers like
```ts
fixures.queries.sqlQuery(datasourceId, "SELECT * FROM categories;")
```

qa-core/src/internal-api/tests/dataSources/mariaDB.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      name: "Query 1",
+      parameters: {},
+      queryVerb: "read",
+      schema: {},
+      transformer: "return data",
+    }
+    // Query data source
+
+    const [queryResponse, queryJson] =
+      await config.api.datasources.previewQuery(dataSourceQuery)
```
Rory-Powell, line 45: Can we do some assertion on the `queryJson`? Like an expected response from this query.

Mitch-Budibase (approved):

qa-core/src/internal-api/api/apis/IntegrationsAPI.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  constructor(client: BudibaseInternalAPIClient) {
+    this.client = client
+  }
+
+  async getAll(): Promise<[Response, any]> {
+    const [response, json] = await this.client.get(`/integrations`)
+    expect(response).toHaveStatusCode(200)
+    const integrationsCount = Object.keys(json).length
+    expect(integrationsCount).toBe(16)
```
shogunpurple, line 15: could be flaky as we add more integrations - prob not worth checking the count

qa-core/src/internal-api/api/apis/DatasourcesAPI.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  async getQuery(queryId: string): Promise<[Response, any]> {
+    const [response, json] = await this.client.get(`/queries/${queryId}`)
+    expect(response).toHaveStatusCode(200)
+    return [response, json]
+  }
+
+  async getQueryPermissions(queryId: string): Promise<[Response, any]> {
+    const [response, json] = await this.client.get(`/permissions/${queryId}`)
```
shogunpurple, line 83: Not a blocker, but we now have a lot of boilerplate for this:

```
const [response, json] = await this.client.get(`/queries/${queryId}`)
expect(response).toHaveStatusCode(200)
return [response, json]
```

We could probably wrap this in an abstraction at this point - because it does 3 things:
- Fetches the API call and returns the response and json
- Makes an assertion
- returns the response and json in a "tuple"

Just a rough (untested) example here, of how we could probably maintain this, make it more composable and reduce the boilerplate significantly without sacrificing anything.
```js
function testResponseWrapper(url, { method, body, assertions}) {
   const [response, json] = await this.client[method](url)
   if (assertions) { 
       assertions(response, json)
   } else {
       expect(response).toHaveStatusCode(200)
   }
  return [response, json]
}

async getQuery(queryId: string): Promise<[Response, any]> {
    return testResponseWrapper(`/queries/${queryId}`, { method: "get" }))
 }

async saveQuery(body: any): Promise<[Response, any]> {
    return testResponseWrapper(`/queries/${queryId}`, { method: "post", body }))
}
```

shogunpurple (approved):
LGTM! Just some stylistic comments but looks good to merge

PR #10267, opened 2023-04-11
Chore - use pro as submodule

adrinr:
## Description
This PR changes the way we interact with budibase-pro. Instead of being an independent repository published via npm or consumed via yarn links, now the pro will be treated as any other package in the budibase monorepo. The main difference is that this code lives in a different repo, as it's private code, but consuming it as a submodule, it will act as any other existing package.
Doing so we have the benefits of having all the packages together, and using properly build tools such as nx and esbuild, and dev tools such as proper debugging and hot reloading.

The main changes in this PR are:
1. Creating the reference to the submodule, it will be checked out in `packages/pro`
2. Changing all the relative references to use this new path
3. Deleting all the scripts and pipelines that were doing the checkout and link "manually"
4. Changing the repo `bootstrap` to link properly the pro package for `account-portal`
5. Removing all the `references` in the `tsconfig.json` files, as now they are not required anymore. Nx handles these dependencies for us, so we don't need to worry. This has a benefit in that we don't need to check if we have pro loaded or not.
6. Adding an extra pipeline step that checks that the updated submodule has been merged properly. This will ensure that we don't commit references to pro that has not been merged and reviewed

### Contributor experience
_**What happens to open source contributors or developers that have the module not loaded?**_
Nothing, for them the experience will be the same as it is now. When we run `yarn install`, the workspace will detect that the package does not exist locally, so it will use the npm repo as it's doing now. No symlinks needed, `yarn workspace` handle this for us out of the box

image

Addresses: 
- https://linear.app/budibase/issue/BUDI-5831/research-spike-npm-less-deployments

## Screenshots
image

## Documentation


package.json
```diff
@@ -23,7 +23,7 @@
   },
   "scripts": {
     "setup": "node ./hosting/scripts/setup.js && yarn && yarn bootstrap && yarn build && yarn dev",
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap && cd packages/pro && lerna bootstrap && yarn setup",
```
Rory-Powell, line 26: I think the bootstrap shouldn't run the setup, is that an infinite loop?

scripts/link-dependencies.sh
```diff
@@ -34,35 +34,11 @@ yarn unlink
...
-
-  echo "Linking pro to worker"
-  cd packages/worker && yarn link '@budibase/pro'
-  cd -
-
-  echo "Linking pro to server"
-  cd packages/server && yarn link '@budibase/pro'
-  cd -
-fi
+echo "Linking pro"
```
Rory-Powell, line 37: Do we need to do this at all anymore? I'd have thought with it now being a package it should be auto symlinked like the other packages.

Aside, if that's not possible for some reason we should keep the `if [ -d "../budibase-pro" ]; then` check (updated for the correct path) so that open source contributors don't run into trouble with this script.

.github/workflows/budibase_ci.yml
```diff
@@ -71,8 +67,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 14.x
-      - name: Install Pro
-        run: yarn install:pro $BRANCH $BASE_BRANCH
```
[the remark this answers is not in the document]
Rory-Powell: looks like `actions/checkout` supports submodules https://github.com/actions/checkout/blob/main/README.md?plain=1#L92

lerna.json
```diff
@@ -1,7 +1,21 @@
...
+    "packages/cli",
+    "packages/client",
+    "packages/frontend-core",
+    "packages/sdk",
+    "packages/server",
+    "packages/shared-core",
+    "packages/string-templates",
+    "packages/types",
+    "packages/worker",
+    "packages/pro/packages/pro"
```
Rory-Powell, line 17: We could consider bringing in pro as a direct dependency and ditching the lerna setup in that repo, it was setup that way for future expansion but that never materialised. Could save us from having to do any custom bootstrapping logic. 

The only thing that stands out is the offline license generation script which rely on the `offline-keys` dir at the top level. Arguably this script should be pulled out of the pro repo altogether and used in a standalone repo that pulls secrets from secrets manager instead. There's a bit of risk having it anywhere near the source code when it comes to packaging and releasing to npm.

packages/server/jest.config.ts
```diff
@@ -16,15 +16,10 @@ const baseConfig: Config.InitialProjectOptions = {
     "@budibase/backend-core": "<rootDir>/../backend-core/src",
     "@budibase/shared-core": "<rootDir>/../shared-core/src",
     "@budibase/types": "<rootDir>/../types/src",
+    "@budibase/pro": "<rootDir>/../pro/packages/pro/src",
```
Rory-Powell, line 19: This may break open source contributors who don't have access to the pro repo. Without the source mapping they will get the types only build of pro from npm + the minified bundle. If we include this they'll likely be in a broken state

packages/worker/jest.config.ts
```diff
@@ -16,15 +16,9 @@ const config: Config.InitialOptions = {
     "@budibase/backend-core/(.*)": "<rootDir>/../backend-core/$1",
     "@budibase/backend-core": "<rootDir>/../backend-core/src",
     "@budibase/types": "<rootDir>/../types/src",
+    "@budibase/pro/(.*)": "<rootDir>/../pro/packages/pro/$1",
+    "@budibase/pro": "<rootDir>/../pro/packages/pro/src",
   },
 }
 
-// add pro sources if they exist
-if (fs.existsSync("../../../budibase-pro")) {
```
Rory-Powell, line 23: As above

scripts/pro/release.sh
```diff
@@ -1,95 +0,0 @@
-#!/bin/bash
```
Rory-Powell, line 1: I can't tell you how glad I am to see this deletion

package.json
```diff
@@ -48,7 +46,6 @@
     "dev:server": "yarn run kill-server && lerna run --stream --parallel dev:builder --concurrency 1 --scope @budibase/backend-core --scope @budibase/worker --scope @budibase/server",
     "dev:built": "cd packages/server && yarn dev:stack:up && cd ../../ && lerna run --stream --parallel dev:built",
     "test": "lerna run --stream test --stream",
-    "test:pro": "bash scripts/pro/test.sh",
```
Rory-Powell, line 51: Out of curiosity what happens when we run test, build etc with an empty pro dir? 
Does lerna just ignore the package in that case?

Rory-Powell: It was kind of nice that we were able to have a dedicated job in the pipeline for the pro tests (especially as they are somewhat flaky at present) - might be able to replace that with `--scope` definitions instead

lerna.json
adrinr: The current approach is:
1. Use the pro repo as it is now as a sumodule
2. Get sure everything runs fine (build, tests, pipelines, etc)
3. Publish this changes

At this point we will know that this approach works fine. So then, when the time is convenient and in sync on whatever other work is happening in pro, we will refactor pro to ditch lerna and have it as a single package

Otherwise, I will be having plenty of conflicts in pro while setting everything up in budibase repo

package.json
adrinr: Do you think it's worth having pro as a separate task? As you say, we can use `--scope` and `--exclude` for it.
On the first question, yarn workspace will be checking for valid packages in `package/*`, so if it does not exists, nothing happen. If this is the case, it will fetch `@budibase/pro` from the repo.
At least, this is what I hope it will happen :D Still testing in progress

packages/server/jest.config.ts
adrinr: Yup, when everything runs for us I will be checking what is the 'no-pro' experience 👍🏽

scripts/link-dependencies.sh
adrinr: This will be needed if `account-portal` needs pro

Rory-Powell: Sure 👍

.github/workflows/budibase_ci.yml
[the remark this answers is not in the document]
adrinr: Yes, already done using the flag!

package.json
adrinr: Added back

package.json
adrinr: It runs the bootstrap in pro. For the time being is still required because pro uses lerna as well. This will be removed when the pro repo gets rid of the lerna dependency to become a single package

packages/server/jest.config.ts
adrinr: Done

package.json
```diff
@@ -82,12 +79,25 @@
     "mode:account": "yarn mode:cloud && yarn env:account:enable",
     "security:audit": "node scripts/audit.js",
     "postinstall": "husky install",
-    "install:pro": "bash scripts/pro/install.sh",
-    "dep:clean": "yarn clean && yarn bootstrap"
+    "dep:clean": "yarn clean && yarn bootstrap",
```
Rory-Powell, line 82: Maybe could add the `-y` option to clean to automatically handle the question prompt - if you're running `dep:clean` its likely intentional

tsconfig.json
```diff
@@ -1,9 +0,0 @@
-{
```
Rory-Powell, line 1: I think this was needed for the `build:dev` command - I'm not sure if it's widely used by the team, it was something I added when doing the backend-core conversion to TS that allows for running ts build in watch mode across all backend dependencies - possibly not applicable anymore with nx.

package.json
```diff
@@ -82,12 +79,25 @@
...
+    "submodules:load": "git submodule init && git submodule update && yarn && yarn bootstrap",
+    "submodules:unload": "git submodule deinit --all"
```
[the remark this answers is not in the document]
adrinr: Added some more docs in pro about this

tsconfig.json
Rory-Powell: Nice!

Rory-Powell (approved):

PR #10276, opened 2023-04-12
Automation step plugins

mike12345567:
## Description
This work from lab-day brings in a few improvements to the automation system, as well as support (in self host installations) for automation plugins.

Action and logic steps can be created using this system, using the standard schema of our automation steps and a small runner function. These are largely similar to the datasource system, simply with a different schema and a different place that they are ran.

Most of the work required for this was between two places:
1. Defining the skeleton plugin: https://github.com/Budibase/budibase-skeleton/tree/master/automation
2. Updating all of the type definitions so that the joi validator could be accurate - note that the vast majority of the changes that have occurred in this PR stem from this, simply improving the typing around the automation schemas as previously it was all strings, wanted to use more enums.

This comes with all of the audit logs/event handling that plugins has, as it is simply an extension of the existing system.

This also includes some improvements to the CLI, the new logger can cause the CLI output to become a little garbled, so added a mechanism to disable it.

There are some limitations to this feature which should be discussed:
1. This can only run in self host, like datasources, due to the risk to the cloud environment.
2. These have the same issue with packages that require binaries/pre-builts to be used, at the minute there is no way to package these pre-builds into the bundle for use with automations or datasources.
4. The steps are confined to the limitations of what a single function call can do, there is no way to perform say branching logic with these.
5. Triggers are not supported, triggers in the Budibase automation system are very unique and individual, with a lot of hooks into different components of the system; for the forseeable the only way new automation triggers can be created is through PRs to the core automation system to add a new trigger.

## Screenshots
CLI - automation plugin creation
image

Adding the plugin - like other ones
image

Automation plugin appearing in automation add step modal
image

Automation step in use
image


PClmnt (approved):
LGTM, nice work getting types in for the steps etc!

packages/backend-core/src/logging/pino/logger.ts
```diff
@@ -31,6 +42,15 @@ if (env.isDev()) {
 
 export const logger = pino(pinoOptions)
 
+export function disableLogger() {
+  console.log = BUILT_INS.log
```
adrinr, line 46: How come this is disabling the logs? Is the BUILT_INS.log not the console.log?

packages/backend-core/src/plugin/utils.ts
```diff
@@ -82,6 +90,55 @@ function validateDatasource(schema: any) {
   runJoi(validator, schema)
 }
 
+function validateAutomation(schema: any) {
```
adrinr, line 93: These validations are quite heavy. Do we usually write unit tests for it?

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ActionModal.svelte
```diff
@@ -26,15 +26,15 @@
 
   const external = actions.reduce((acc, elm) => {
```
adrinr, line 27: Will custom not always be internal?

packages/backend-core/src/logging/pino/logger.ts
mike12345567: By default, all of the loggers have been overwritten with the pino logger, if you load backend-core this is the way it works. I didn't want to change this API, but I need the `console.log` to be the standard built in Node.js one *purely* for the CLI, all other usages of backend-core should have their logger replaced by the pino logger.

packages/backend-core/src/plugin/utils.ts
mike12345567: We haven't in the past, but could easily add something in to run through the datasource/automation validations with our internal ones, the internal ones should all pass so would be an easy test case to write.

Issue #10216, opened 2023-04-05
[BUDI-6754] When users are updated, user metadata should be reflected in all the apps


shogunpurple:
When I'm viewing the user section of the builder, I will only see users in my users table when I publish the application and trigger the user sync.

**COS**

* Trigger the user metadata sync on app creation
* Trigger the user metadata sync when the app is in dev
* Look at ways to parallelise user sync using bull queue or jobs

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6754](https://linear.app/budibase/issue/BUDI-6754/when-users-are-updated-user-metadata-should-be-reflected-in-all-the)</sub>

shogunpurple:
@michael.drury create issue around user invites:

* Users being added to apps from drawer but not available in the users metadata table

PR #10287, opened 2023-04-13
Improving user sync into app user metadata tables

mike12345567:
## Description
This PR introduces a re-work of the user sync into app user metadata tables. There are a few major improvements which need to be considered:

1. This process is now asynchronous, depending on a Bull queue to carry the events which trigger various syncs into apps - this means that in environments with a lot of users and apps, there is no longer a large delay when updating roles/groups
2. The process now removes users which no longer have access to apps - this can in theory cause data loss, the reason we didn't originally do this was because a user might have some extra data held against the metadata table for them, but in reality the users continuing to exist after they no longer have access to an app is confusing
3. There still is a sync process when opening the app, just to confirm that everything is in correct order, this was the only reasonable way to cleanup user metadata tables which have gotten into a bad state (say the same user repeated multiple times, this could happen through funky automations)
4. This also means that users can be updated by an admin while another user is building an app and the updates will be propagated correctly into the app metadata tables
5. This also introduces a new system for tracking document updates, I've tied this to the event system, so that when an event like `user:updated` is triggered, we can add events to a Bull queue, which can be picked up, processed and handled. This mechanism means that the worker can update users/groups, the event system registers those on the Bull queue and then the server can pick them up and process the app user metadata table updates.

I have also added a new unit test case for this system, triggering events and checking what updates have occurred within an app. I found a variety of 409 issues related to view creation and groups which have now been resolved, so hopefully an improvement in that area as well.

Addresses: 
- https://github.com/Budibase/budibase/issues/10216
- https://linear.app/budibase/issue/BUDI-6754/when-users-are-updated-user-metadata-should-be-reflected-in-all-the


packages/backend-core/src/events/processors/DocumentUpdateProcessor.ts
```diff
@@ -0,0 +1,42 @@
+import { Event, Identity, Group, DocumentUpdateEvents } from "@budibase/types"
+import { EventProcessor } from "./types"
+import * as docUpdates from "../../docUpdates"
+import { getTenantId } from "../../context"
+
+export default class DocumentUpdateProcessor implements EventProcessor {
```
Rory-Powell, line 6: I think we can do a bit of refactor around the difference between event publisher and event processor. 

Here we are using an event _processor_ as a _publisher_, with a separate concept of processing inside `docUpdates/updates`.

Ideally we would have a unified interface around event processing and publishing and slot the document processing into that. 

For example we can move publishing out of `DocumentUpdateProcessor` and place it here instead:
https://github.com/Budibase/budibase/blob/develop/packages/backend-core/src/events/events.ts#L6

We could still keep the `DocumentUpdateEvents` as a whitelist (but maybe change it to `AsyncEvents`) for when to do asynchronous job based processing (although ultimately I think every event should be processed this way). Could then use it like 

```ts
interface EventPayload {
    event: Event,
    identity: Identity,
    properties: any,
    timestamp?: string | number
}

if (AsyncEvents.indexOf(event) !== -1) {
   // async - as fast as possible
   const data: EventPayload = { event, identity, properties, timestamp }
   queue.add(data)
} else {
   // sync - slows down response
   processors.processEvent(...)
}
```

packages/server/src/sdk/app/applications/sync.ts
```diff
@@ -1,6 +1,147 @@
...
+  const failed = resp.filter(promise => promise.status === "rejected")
+  if (failed.length > 0) {
+    const reasons = failed.map(fail => (fail as PromiseRejectedResult).reason)
+    logging.logAlert("Failed to sync users to apps", reasons)
+  }
+}
+
+export function initUserGroupSync(updateCb?: (docId: string) => void) {
+  const types = [constants.DocumentType.USER, constants.DocumentType.GROUP]
+  docUpdates.process(types, async update => {
```
Rory-Powell, line 118: I think this is maybe the wrong place to put this, with the event intg being nested inside the SDK. It feels like this could be a candidate to pull out to a top level `events` package in server. 

Ideally we could init event processing in a single place in the server for all future events. 
e.g. 

```ts
class BullEventsProcessor implements EventProcessor {

  processors: { [key in Event]: EventProcessor[] }

  constructor(processors) {
      this.processors = processors
  }

  init() {
      // hand off to the event processor interface
     queue.process(job: EventPayload, () => { await processEvent (job.event, job.identity... })
  }
 
  // main process impl to hand off to individual processors
  async processEvent(
    event: Event,
    identity: Identity,
    properties: any,
    timestamp?: string
  ): Promise<void> {
    // process the event with any registered sub-processors
     const processors =  processors[event]
     for (processor in processors) {
       processor.processEvent(event, identity, properties, timestamp)
     }
  }
```

We can then create a dedicated processor that can handle user group sync and register it with the above 
```ts
class UserGroupSyncEventProcessor implements EventProcessor {}
```

This will set us up for really easy and clear expansion, by just implementing `EventProcessor` and registering it with the main processor.

packages/backend-core/src/events/processors/DocumentUpdateProcessor.ts
Rory-Powell: See below comment about integration on the read side

mike12345567: I understand the logic of this - but after some thought I don't think its a great idea to move to a half an half model for this. I'm going to refactor this a bit to make the async processing possible, using a standard event payload, but I don't like the idea of cutting over some events to async and some not.

Personally I believe this is quite outside the scope of this issue, as theres quite a risk of breaking the whole event processing pipeline by adding Bull inbetween it.

The goal of using the queue was to disconnect the producer of the event from the consumer, and allow an intensive process to become a background one, rather than re-building the event system.

I've updated it so that the `publishEvent` now fires off onto an async events queue, with a generic event payload, but it will also continue and finish the actual synchronous event processing.

I fear that disconnecting the producer of the event to the actual Posthog updates and all other parts of the event processing pipeline will come with a lot of risks - something I'd prefer to avoid when it doesn't appear to be causing any major issues.

We can open a separate issue if we want to move completely to a async model for events, with the testing required to see what impact this will have to our pipeline; where I'll leave it will be easily converted to that if desired.

packages/backend-core/src/events/processors/async/DocumentUpdateProcessor.ts
```diff
@@ -0,0 +1,44 @@
+import { EventProcessor } from "../types"
+import { Event, Identity, DocUpdateEvent } from "@budibase/types"
+import { DocumentType, SEPARATOR } from "../../../constants"
+import { doInTenant } from "../../../context"
+import { getDocumentId } from "../../documentId"
+import { shutdown } from "../../asyncEvents"
+
+export type Processor = (update: DocUpdateEvent) => Promise<void>
```
Rory-Powell, line 8: We could remove:
- `DocUpdateEvent`
- `Processor`
In favour of:
- `EventPayload`
- `EventProcessor` 


I don't see a lot of benefit in implementing a sub event type system when we could simplify and re-use the top level ones.

packages/backend-core/src/events/processors/async/DocumentUpdateProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  async processEvent(
+    event: Event,
+    identity: Identity,
+    properties: any,
+    timestamp?: string | number
+  ) {
+    const tenantId = identity.tenantId
+    const docId = getDocumentId(event, properties)
```
Rory-Powell, line 25: Could this potentially trigger a lot of redundant user syncs? 
e.g. if a user is created we would send:
- `user:created`
- `user:builder:assigned`
- `user:admin:assigned`
- `role:assigned`

The same for any update and onboarding completed. It might be simpler / more lightweight if we add a function to `EventProcessor`: 
```ts
supports(event: Event)
```
The `DocumentUpdateProcessor` can return `true` for everything. 
Then the sub `processors` e.g. `UserSyncProcessor` can have:

```ts

const supported  = [
  Event.USER_CREATED,  
  Event.USER_UPDATED, 
  Event.USER_DELETED, 
  Event.GROUP_CREATED, 
  Event.GROUP_UPDATED,
  Event.GROUP_DELETED,
]

supports(event: Event) {
   return supported.includes(event)
}
```

It then becomes a lot easier / more specific for when we run the user sync, rather than relying on the presence of a `userId` or `groupId` field in the event properties. 

As we add more event types to the async event processing we may run into more unintended consequences of events with a user or group field running the sync. This problem can only become more compounded over time.

Rory-Powell (commented):
Overall this is a fantastic change and definitely going in the best direction for future async event handling in budibase. 

I left some comments around the specific of `docUpdates` - I think there are some side affects that we might run into with bundling multiple events into buckets by doc type and running the processor for all of them. I think the concept is a bit over complicated given we have interfaces and types for event processing already. 

Preferably we could be specific in processors about which events we want to react to. 

Left suggestions around using a more native integration with `Event` and `EventPayload` by implementing `EventProcessor` in named files in the server, effectively each file acting as a 'job handler'. This should be easier to extend and test.

packages/backend-core/src/events/processors/DocumentUpdateProcessor.ts
Rory-Powell: Maybe I wasn't clear, I wasn't suggesting that we cutover to async processing for everything right away - that's definitely another piece of work. What you've done around integrating into the publisher and having a generic way to init and extend events in server is exactly what I was looking for. Thanks for making the change

Rory-Powell: I see now why the above might have been misleading, it should have been an `if` only and not an `if-else`

packages/backend-core/src/events/processors/async/DocumentUpdateProcessor.ts
mike12345567: This is the same as previous comment - I really prefer having the specific processors which aren't aware of the event system - in the actual usage you just want a simple, functional interface where you pass a function that happens when a document type is updated - completely separate from event processing (a somewhat complex and very un-related system).

packages/backend-core/src/events/processors/DocumentUpdateProcessor.ts
mike12345567: This makes a lot more sense now!

PR #10303, opened 2023-04-14
Permissions test

mbadan:
## Description
Added unit tests for backend-core permissions, covering all of the main functions, such as:
levelToNumber,
getAllowedLevels,
doesHaveBasePermission,
isPermissionLevelHigherThanRead,
getBuiltinPermissions,
getBuiltinPermissionByID

Got the coverage of the file to: 98.36% Statements,95.65% Functions,100% Lines


packages/backend-core/src/security/tests/permissions.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    expect(isPermissionLevelHigherThanRead(PermissionLevel.WRITE)).toBe(true)
+  })
+
+  it("should return false if level is read or lower", () => {
+    expect(isPermissionLevelHigherThanRead(PermissionLevel.READ)).toBe(false)
+  })
+})
+
+describe("getBuiltinPermissions", () => {
+  test("returns a clone of the builtin permissions", () => {
```
mike12345567, line 114: should use `it` rather than `test`

packages/backend-core/src/security/tests/permissions.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    const builtins = getBuiltinPermissions()
+    expect(builtins).toEqual(cloneDeep(BUILTIN_PERMISSIONS))
+    expect(builtins).not.toBe(BUILTIN_PERMISSIONS)
+  })
+})
+
+describe("getBuiltinPermissionByID", () => {
+  test("returns correct permission object for valid ID", () => {
```
mike12345567, line 122: same as above

packages/backend-core/src/security/tests/permissions.spec.ts
mbadan: Fixed

packages/backend-core/src/security/tests/permissions.spec.ts
mbadan: Fixed

packages/backend-core/src/security/tests/permissions.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  isPermissionLevelHigherThanRead,
+  PermissionLevel,
+  PermissionType,
+  levelToNumber,
+  getAllowedLevels,
+  BuiltinPermissionID,
+  getBuiltinPermissions,
+  BUILTIN_PERMISSIONS,
+  Permission,
+} from "../permissions"
```
Rory-Powell, line 14: Generally prefer whole module imports  i.e. `* as permissions` so we can just use the single import in all the tests like
```ts
expect(permissions.levelToNumber(...))
```

Reads slightly better in the test as I know the subject under test rather than what could be a local function definition

mbadan: Done, had to change some things along with the function declaration, but its done and working

mike12345567 (approved):
Looks good!

PR #10309, opened 2023-04-14
Chore - Remove unnecessary second build in pipeline

adrinr:
## Description
Remove conflicting second build step in the release pipeline

### DIFF
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -75,7 +75,6 @@ jobs:
       - name: Build/release Docker images
         run: | 
           docker login -u $DOCKER_USER -p $DOCKER_PASSWORD
-          yarn build
           yarn build:docker
         env:
           DOCKER_USER: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-master.yml
```diff
@@ -75,7 +75,6 @@ jobs:
       - name: Build/release Docker images
         run: | 
           docker login -u $DOCKER_USER -p $DOCKER_PASSWORD
-          yarn build
           yarn build:docker
```
Rory-Powell, line 78: The only thing now might be the `backend-core` module which has a similar problem as pro, where it uses a nested inner structure in dist:
```
dist/
   src/
   tests/
   package.json // this is now behind 
package.json
```

It's a bit more complicated to solve this one. It might involve moving tests inside src and updating the exposed modules definition in the `package.json` to point to the inner path instead.

adrinr:
This should be good to be merged now that we changed it in here: https://github.com/Budibase/budibase/pull/10324/

.github/workflows/release-master.yml
adrinr: Solved in here: https://github.com/Budibase/budibase/pull/10324

Rory-Powell (approved):

Issue #7400, opened 2022-08-22
Update table data via csv import


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7353

<div type='discussions-op-text'>

<sup>Originally posted by **yewkay** August 19, 2022</sup>
It seems that there is only an import function in the database CSV import function. At the moment I didn't find a function within budibase to update a table with existing data using csv. Importing a csv into a table with data will repeat the data. 

I would like to propose a way of matching the data to be updated eg: using a user-defined column as the ID, so each row item in that column will be matched and all the other columns will be updated with the new CSV being imported.

Additional suggestions to build on the import function:
- it could also contain mapping of columns, eg: DB columns A, B, C.. etc are matched to CSV's Columns E,F,G
- Additional columns that are not matched in the CSV can have the option to either ignore or create a new column 
</div>

stellarfinn:
@melohagan  Has this been abandoned?

melohagan:
![Screenshot 2023-09-20 at 15 44 59](https://github.com/Budibase/budibase/assets/101575380/5493ddbf-44c7-4953-bfde-11d307199a9b)

Hey @stellarfinn 

This has already been released for the internal Budibase DB!

It was skipped for external databases because there would have been many implementations to do, and you can write your own update queries for external datasources.

PR #10313, opened 2023-04-14
Update table data via CSV import 

melohagan:
## Description
Added a toggle for internal DB CSV imports that allows a user to update existing rows.
You choose one or more fields to match for the update. This could easily be changed to a single options picker if that's easier to understand for the user.

I choose to only make this option available for the Internal DB because to be honest I wasn't able to figure out how to do an upsert with Knex - the blogs I saw were for a specific DB implementation and not a generic one.
Also an upsert would also need to be implemented for Google Sheets, so it may be worth adding this feature just for the Internal DB to start with. 

There is also no import verification for uniqueness. As it is, it's up to the user to pick fields appropriately, but a check could potentially be added. 

Addresses: 
- https://github.com/Budibase/budibase/issues/7400

## Screenshots

**Insert mode**
Screenshot 2023-04-14 at 22 56 08

Duplicate row:
Screenshot 2023-04-14 at 22 57 19

**Upsert mode (one key field)**
Screenshot 2023-04-14 at 23 01 18

**Upsert mode (multiple key fields)**
Screenshot 2023-04-14 at 23 03 40

Original row updated:
Screenshot 2023-04-14 at 23 01 45


mike12345567 (commented):
EDIT - failed to notice the table type being passed around - all good.

mike12345567 (approved):

Issue #3408, opened 2021-11-17
Table creation column ordering


shogunpurple:
### Discussed in https://github.com/Budibase/budibase/discussions/3097

<div type='discussions-op-text'>

<sup>Originally posted by **tluyben** October 20, 2021</sup>
I am just recently trying some things out with Budibase and I was wondering, when I create a table in the Data view, the fields jump to the end after you edit them; I know I can order them in the Design view, but you want consistent ordering in the Data view as well. Is there no way to leave them ordered as I put them in drag/drop/change some metafield 'ordering' ? 

  </div>
  
**COS**
- Table columns in data UI are draggable and can be rearranged
- This stores a `sortingOrder` key on the field in the schema in the database
- If no `sortingOrder` applied, sort columns alphabetically to prevent them moving after editing

gvlx:
another suggestion: #3290

adamjdavies:
Would be great to have this!

melohagan:
Now possible with the Grid view!

PR #10329, opened 2023-04-18
Grid UI in data section

aptkingston:
## Description
This PR adds an editable spreadsheet to the data UI, replacing the existing table. The spreadsheet enables editing data inline, as well as a long list of other features.

This is a complete monster and I'm still testing it, so may continue to make small changes as I find them.
Todo:
- [x] Use unique edit modal for users
- [x] Disable websocket integration server-side since we aren't using it for now
- [x] Get maze link for feedback button
- [x] Rename to Grid
- [x] Add `currentApp` middleware to websocket chain (still disabled for now)

image
image

## Infinite scroll
Rows are automatically loaded as you scroll down. There's no need to manually page through data.

https://user-images.githubusercontent.com/9075550/232727532-bdf94786-6bd3-4b2d-90dd-d6f8be999dc1.mov

## Multi-user collaboration
The spreadsheet supports multiple user collaboration via websockets but I've disabled this for now, since only one user can edit an app at once.

## Editing data inline
All data types can be edited inline:
- Text / Barcode
image
- Numbers
image
- Options
image
- Long form text
image
- Multi select
image
- Attachments
image
- Booleans
image
- JSON
image
- Dates
image
- Relationships (the new relationship picker supports searching for the row you want to add, using the primary display column of the other table). You can also click on a relationship badge to open the related row in a modal.
image
- Formula (readonly)
image
- Autocolumns are readonly versions of their respective types

## UI features
The spreadsheet is very flexible in how it presents data.

**Al UI customisation features are persisted as part of table metadata, so will be saved and used next time you view the same table.**

### Sticky column
The primary display field of your table is sticky, and will always stay pinned to the left of your sheet as you scroll.

https://user-images.githubusercontent.com/9075550/232718641-0438faa3-0f50-4011-9fd6-a31a9b6bbd7f.mov

### Column sizes
Columns can be resized by dragging.

https://user-images.githubusercontent.com/9075550/232718098-b23983bc-48b2-4b57-9846-183b2f753476.mov

### Column order
Columns can be reordered by dragging (after a very short click and hold), or by using the header cell context menu.

https://user-images.githubusercontent.com/9075550/232719205-5a3bd20e-5717-405a-92ee-0749e0d828a7.mov

### Column visibility
Columns can be hidden or shown as desired, using the dedicated button above the sheet.

image

### Row height
Row height can be customised using the 3 preset options. Using a larger row height will give more space for things like long text fields, relationships, attachments and multi-select fields.

Row height can be configured using the button above the sheet:
image

Small (default):
image

Medium:
image

Large:
image

### Sorting
Sort column and order can be chosen using the button above the sheet, and can also be updated using the header cell context menu:
image

The sort column currently in use is highlighted (note the darker background and icon in the header cell):
image

### Filtering
Filtering uses the existing filter modal - this has not been changed. It can be opened using the button above the table.

## Context menus
Context menus have been added, both to normal cells and to header cells.

For header cells, the context menu offers the following options:
image

- Edit column (opens in the existing edit column modal)
- Use as display column (immediately updates the table definition to change display column, and pins the column to the left)
- Sort A-Z
- Sort Z-A
- Move left (swaps position one place to the left)
- Move right (swaps position one place to the right)

For regular cells, the context menu offers the following options:
image

- Copy
- Paste
- Edit row in modal (opens the row for editing in the existing edit row modal, if you prefer that view for certain data)
- Duplicate row (creates a duplicate row immediately below the selected one)
- Delete row

## Gutter
The sheet gutter contains 2 functions - row selection and "expand" arrows to open rows in a modal for editing. By default, the gutter contains the row numbers. Hovering over a row reveals a checkbox to select the row, and an icon to open the row in a modal.

### Row selection
All rows can be selected using the checkbox in the header row. When rows are selected, they can be bulk deleted using the button in the top right.

https://user-images.githubusercontent.com/9075550/232722281-43ce33d4-8f68-4eee-9451-b3c85e13e42e.mov

### Editing rows in modal
Using the gutter icon, you can open rows for editing inside the existing row modal.

https://user-images.githubusercontent.com/9075550/232722517-3cd5be75-36e7-4c2b-9ea8-28170bca202b.mov

## Keyboard shortcuts
The sheet supports keyboard shortcuts for navigation and editing data. It basically works as you would expect, the same as something like google sheets or excel.

- Arrow keys can be used to navigate around
- Tab can be used to go to the next cell
- Enter focuses a cell for editing and confirms changes for text cells
- When editing more complicated cells like options or relationships, you can search, change options and add data all by the keyboard
- Delete or backspace clears a cell value
- Cmd/Ctrl+C/V does copy and paste
- The spreadsheet will always keep the focused cell in view and handle scrolling for you

## Validation
If saving a row fails, validation errors will be shown on the offending fields. If the offending fields are not visible then they will be shown, and the sheet will automatically scroll to the first error.

Row changes are cached as long as a row is selected. For example, if you attempt to change `Name` but the change fails because a required field `Age` is blank, then your new value for `Name` will be persisted in the sheet until you either fix the error (in which case all pending changes are saved) or change row, in which case all pending changes are discarded.

There is no client side validation - the sheet attempts to make changes and allows the server to reply with validation errors, but this is all handled gracefully in the UI.

Here's a quick example of making a change, receiving an error, correcting the error, then having all changes persisted:

https://user-images.githubusercontent.com/9075550/232724824-e15864e9-d4c9-4490-8b7f-3bba3b8207de.mov

## Row and column creation
For now, row and column creation are handled using the existing modals. We cannot easily allow adding rows by clicking the bottom of the sheet, as there may not be a bottom (e.g. we may have 1 million rows in 100,000 pages of data, and we don't know where the end is). I've made a PoC example of inserting new rows inline as the first row, but it's disabled for now. We can have a think about how we want to tackle this.

## Performance
The sheet is extremely fast. It uses virtual rendering techniques and CSS transforms to fake the appearance of scrolling. The largest dataset I have is 10,000 rows but I can say with confidence that there is no upper limit on how many rows it can handle, as long as your browser can handle loading them all into memory.

Quick example of scrolling over 10,000 rows (note, in this example I manually kept scrolling and loading more rows into memory until I reached 10,000, but obviously by default we do not pull back so many rows and they are fetched incrementally as you scroll down):

https://user-images.githubusercontent.com/9075550/232726111-90ecd358-f38b-4dd1-b5e3-11463b77c8fb.mov

## Addresses
- #3408
- #7810

## Documentation


melohagan:
This is amazing in so many ways! Very excited for this! 🤩

Issue #7299, opened 2022-08-16
[BUDI-6903] Add "empty text" property to all data components


Klience:
**Describe the feature request**
Right now only Repeater has "Empty text" property that lets you specify what text to show when data is empty.

Other data components don't have that feature. E.g. if a table is empty it says "no rows found".

This is also a small step towards letting us custom-localize hard-coded messages (like "rows not found") into other languages besides English.

<sub>[BUDI-6903](https://linear.app/budibase/issue/BUDI-6903/add-empty-text-property-to-all-data-components)</sub>

PR #10332, opened 2023-04-18
Add empty text property to all blocks

melohagan:
## Description
Users wanted to customise the *No rows founds* message in the following:

 - Row Explorer
 - Form Block
 - Table Block
 - Table
 - Cards Block

Addresses: 
- https://github.com/Budibase/budibase/issues/7299

## App Export
[Data Provider message-export-1681816189015.tar.gz](https://github.com/Budibase/budibase/files/11261013/Data.Provider.message-export-1681816189015.tar.gz)

## Screenshots

**Row Explorer**
Screenshot 2023-04-18 at 12 11 43

Screenshot 2023-04-18 at 12 12 06

**Cards Block**
Screenshot 2023-04-18 at 12 13 24

**Table Block**
Screenshot 2023-04-18 at 12 14 04

**Form Block**
Screenshot 2023-04-18 at 12 14 52

*Setting only shows when Update/View is selected*

## Documentation

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -4966,6 +4966,17 @@
         "defaultValue": "spectrum--medium"
       },
       {
+        "type": "text",
+        "label": "Empty text",
+        "key": "noRowsMessage",
+        "defaultValue": "We couldn't find a row to display",
+        "dependsOn": {
+          "setting": "actionType",
+          "value": "Create",
+          "invert": true
+        }
+      },
+      {
         "section": true,
         "name": "Fields",
         "settings": [
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -4547,6 +4547,12 @@
         "key": "paginate"
       },
       {
+        "type": "text",
+        "label": "Empty text",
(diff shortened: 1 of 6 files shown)


aptkingston:
Svelte default values for props only apply when the value is nullish, so I'm curious what happens when the value is an empty string here. Could you try setting a custom error message, then clearing the field (which should reset the setting to an empty string) and see what message is shown?

The way I normally prefer to do this would be just pass the prop in as normal, but then do a loose falsey check at usage, which accounts for both nullish and empty strings.

e.g.
```html
...
export let noRowsMessage
...
<Component
  placeholder={noRowsMessage || "Default no rows message"}
/>
...
```

I generally would reserve default values for cases when you're explicitly targeting `null` or `undefined` values.

This might be a non issue because I'm not sure if the builder already intelligently sends up empty strings as null by the way!

melohagan:
Good catch @aptkingston The default doesn't work when the field is cleared. I'll implement your suggestion now.

aptkingston (approved):
LGTM! Nice one Mel.

Issue #10270, opened 2023-04-12
Unable to fully release the lock on resource "lock:default_persist_writethrough_usage_quota".


FlaminWrap:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 2.5.2
  - App Version: 2.5.2

**Describe the bug**
An error message appears with the text `Unable to fully release the lock on resource "lock:default_persist_writethrough_usage_quota".` randomly when loading application pages

**To Reproduce**
Unsure how to replicate this.

**Expected behavior**
This message not to appear

**Screenshots**
![image](https://user-images.githubusercontent.com/97764630/231412483-1a8093e2-774c-43f1-b201-9a5cadf33c45.png)
  
**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version 111.0.1

**Additional context**
This happens with every app on the tenant.

TheFunkyBiskit:
Also having this issue fairly frequently. Refreshing the page or reloading it in the editor seems to cause the query to re-execute, but it is definitely an issue that started with a recent release of Budibase.

yeoldegrove:
Also seeing this issue after my last update from 2.2.14 to 2.5.2.

kasgithub8:
Same issue on version 2.5.0

FlaminWrap:
This has seemed to have disappeared for me now. I noticed that the budibase host CPU was really high for a few days while it was producing this error.

FlaminWrap:
> This has seemed to have disappeared for me now. I noticed that the budibase host CPU was really high for a few days while it was producing this error.

Spoke too soon, it has returned

PR #10350, opened 2023-04-19
Hotfix - Increase writethrough cache timeout

adrinr:
## Description
Increase the redlock TTL of writethrough cache. It's currently set at 1 second, and some users are reporting that it's timing out. This might be occurring in instances running on slow machines or with heavy CPU usage.

The issue was just "cosmetic", as the current usages of `writethrough` have eventual consistency and the right data will be retried.

This started happening because of this [PR](https://github.com/Budibase/budibase/pull/9835), configuring the redlock.

Addresses: 
- https://github.com/Budibase/budibase/issues/10270

### DIFF
--- a/packages/backend-core/src/cache/writethrough.ts
+++ b/packages/backend-core/src/cache/writethrough.ts
@@ -47,7 +47,7 @@ async function put(
         type: LockType.TRY_ONCE,
         name: LockName.PERSIST_WRITETHROUGH,
         resource: key,
-        ttl: 1000,
+        ttl: 15000,
       },
       async () => {
         const writeDb = async (toWrite: any) => {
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):

packages/backend-core/src/cache/writethrough.ts
```diff
@@ -47,7 +47,7 @@ async function put(
         type: LockType.TRY_ONCE,
         name: LockName.PERSIST_WRITETHROUGH,
         resource: key,
-        ttl: 1000,
+        ttl: 15000,
```
adrinr, line 50: This TTL is only in place to ensure that the lock does not get stuck forever. It can virtually be anything, but 15s should be more than enough for the current operations

Rory-Powell (approved):
LGTM

shogunpurple (approved):
LGTM! Nice turnaround on this fix @adrinr

Issue #10349, opened 2023-04-19
Table display column with a formula referencing related table doesn't display properly in related tables


manwithpowers:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud

**Describe the bug**
A display column in a table from our MySQL database is a formula column that references the name column of a related table. That column works as expected in that table and displays the related name. In related tables, the corresponding column doesn't properly display the name reference.
This seems similar to #4256 but clearly distinct as the issue differs somewhat and 4256 has been fixed & closed.

**To Reproduce**
1. Create the table `courses` (`id`, `name`) with `name` as the display column
2. Create the table `schedule` (`id`, `internal_ref`, `courses_id`, `instance_name` + others)
3. Create one to many relationship between `courses` and `schedule` Tables (`schedule.courses_id`->`courses.id`)
4. Make the `instance_name` column in `schedule` a formula column with `{{ courses.0.name }} - {{ internal_ref }}` and make it the display column
5. The column displays as expected in the `schedule` table but not in the `courses` table (nor in any other table related to `schedule`) - the referenced `name` is missing while the `- {{internal_ref}}` is there.

**Expected behavior**
The related columns in related tables should properly display the referenced column.

**Screenshots**
Attached
  
**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 112.0.5615.121 

**Additional context**
![Screenshot 2023-04-19 at 13 02 15](https://user-images.githubusercontent.com/5508635/233058562-2be948c9-9034-44a1-a657-39bf31be498d.png)
![Screenshot 2023-04-19 at 13 02 34](https://user-images.githubusercontent.com/5508635/233058565-a7bedf46-b6b2-491c-a9e9-119b86c3bbdc.png)

PR #10352, opened 2023-04-19
Allow SQL formulas using related row information to be used as display columns

mike12345567:
## Description
Fixing issue with SQL tables and using formulas that contain relationships as display columns. Also cleaning up imports in ExternalRequests a bit.

Also fixes an issue with creating relationships between existing tables, previously the SQL column generation always assumed a unsigned integer for foreign key types, this tries to match up with an existing primary key types.

Addresses: 
- https://github.com/Budibase/budibase/issues/10349

### DIFF
--- a/packages/server/src/api/controllers/row/ExternalRequest.ts
+++ b/packages/server/src/api/controllers/row/ExternalRequest.ts
@@ -1,31 +1,32 @@
 import {
+  Datasource,
+  FieldSchema,
+  FieldType,
   FilterType,
   IncludeRelationship,
   Operation,
   PaginationJson,
   RelationshipsJson,
+  RelationshipTypes,
+  Row,
   SearchFilters,
   SortJson,
-  Datasource,
-  FieldSchema,
-  Row,
-  Table,
-  RelationshipTypes,
-  FieldType,
   SortType,
+  Table,
 } from "@budibase/types"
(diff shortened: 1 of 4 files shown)


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -382,10 +383,18 @@ export class ExternalRequest {
       }
       const display = linkedTable.primaryDisplay
       for (let key of Object.keys(row[relationship.column])) {
-        const related: Row = row[relationship.column][key]
+        let relatedRow: Row = row[relationship.column][key]
+        // add this row as context for the relationship
+        for (let col of Object.values(linkedTable.schema)) {
+          if (col.type === FieldType.LINK && col.tableId === table._id) {
+            relatedRow[col.name] = [row]
```
adrinr, line 390: Is this right? Is the foreign row an array even if it's a simple link?

mike12345567: Yeap - it always uses an array, even for a simple link, thats just our relationship system, for a one relation, you still need to do `relationship[0]`.

This could change - but its all based on the way we handle relationships in our internal tables.

adrinr (approved):

PR #10357, opened 2023-04-20
allow ssl-policy annotation in helm chart

jonnymccullagh:
Allow the optional specifying of SSL policy to use on the ALB for Kubernetes.
The default policy (if none is specified) supports old TLS versions (e.g. v1).
We would like to specify a more updated policy in our values files e.g. `ELBSecurityPolicy-TLS13-1-2-2021-06`


shogunpurple (approved):
LGTM!

Issue #10358, opened 2023-04-20
[BUDI-6898] Builder - CSV Import Data - Invalid file type error


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/9551

<div type='discussions-op-text'>

<sup>Originally posted by **applicationuser** February  6, 2023</sup>
Hello,

Receiving error while importing CSV file in builder, see below screenshot.  No error in browser console.

CSV file with 4 columns 

Even export of row data and import same file (excluding auto columns)  also NOT working.

Budibase app cloud version : 2.5.2

**OS: Windows 11**

**Browser: Firefox 111**

![image](https://user-images.githubusercontent.com/42922833/217054358-453aaf82-5bd6-453e-b4fc-bc4cbff871a5.png)
</div>

<sub>[BUDI-6898](https://linear.app/budibase/issue/BUDI-6898/builder-csv-import-data-invalid-file-type-error)</sub>

melohagan:
I cannot reproduce this issue on MacOS, however it may be related to this issue here: https://github.com/thoughtbot/paperclip/issues/2170

In our code we explicitly check for `file.type === "text/csv"`
https://github.com/Budibase/budibase/blob/develop/packages/builder/src/components/backend/TableNavigator/utils.js#L46

To support Windows, we may need to have a more generic check that also tests for `application/octet-stream` and `application/vnd.ms-excel`

melohagan:
@applicationuser Could you please try importing a CSV in Windows+Firefox within the cloud environment now please?

Your app version should be **2.5.6**. Please let us know if the issue is resolved for you.

applicationuser:
@melohagan 
With app version 2.5.6,  Windows+Firefox csv import works fine as long as we have csv header names match with table column names ( case sensitive match). 

Thanks for the fix.

melohagan:
That's great to hear @applicationuser 

Thanks for providing the info that helped us locate the issue!

PR #10359, opened 2023-04-20
CSV parsing on Windows/Firefox

mike12345567:
## Description
Fix for #10358 - making sure that client-side we don't check the content type for CSV, check if it might be JSON, if not assume we can try it as a CSV - this is a fix for an issue which occurs on Windows, in Firefox.


melohagan (approved):
LGTM!

PR #10360, opened 2023-04-20
Chore - Pipeline npm version updates

adrinr:
## Description
Changing pipelines to allow "npm-less" builds. After this PR, the pipeline flow will be the following:

1. On develop/master push, tag it as vX.Y.Z-alpha.R and vX.Y.Z respectively
2. Change the develop/master release pipeline to trigger on tag created instead of running on every commit
3. Change the develop/master release pipeline to use the version on the tag as dependencies, to publish the npm and to build the docker images. This action does not pushes any git changes anymore
4. Changing all the committed versions as 0.0.1. On the local development time lerna workspaces will take care of the links, at pipeline time the modified actions will
5. Add some scripts to handle the dev process, depending on the access rights to local pro

Why these changes:
1. Using yarn workspaces we don't need to keep tracking the versions in the code. Constantly updating these versions keep causing conflicts between develop and master, making almost impossible adding extra environments
2. Not having a tag and release as a single action will allow us to break down the pipeline steps, easily rolling back to a specific version without rebumping
3. Not having the package versions will prevent any race conditions between builds and bumps, specially around pro


mike12345567 (approved):
LGTM! So nice to see all those `0.0.1`!

PR #10363, opened 2023-04-20
Removing "export all apps" functionality from cloud

mike12345567:
## Description
Removing export all apps functionality from cloud - this was a very niche feature and often confusing, now there is a consistent flow for starting out in self host.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6900/remove-download-all-apps-cloud-functionality

## Screenshots
Super admin page (self host)
image

Cloud apps page
Download button removed in cloud implementation (same as self host now)
image


melohagan (approved):
LGTM!

PR #10366, opened 2023-04-20
Dynamic filter contains option

melohagan:
## Description
Updated the **Like** filter for the Internal DB lucene search to use regex that matches the equivalent syntax of the like statements in SQL.

There were issues raised in the past and it was removed in the end:
 - https://github.com/Budibase/budibase/issues/3386
 - https://github.com/Budibase/budibase/pull/9030

But I can't see any reason not to use regex here?

Addresses: 
- https://github.com/Budibase/budibase/issues/5995

## Screenshots

**Table block filters**
Screenshot 2023-04-20 at 18 19 22

Screenshot 2023-04-20 at 18 19 47

**Dynamic filter**
Screenshot 2023-04-20 at 18 23 20

## Documentation

### DIFF
--- a/packages/backend-core/src/db/lucene.ts
+++ b/packages/backend-core/src/db/lucene.ts
@@ -309,7 +309,7 @@ export class QueryBuilder<T> {
         return null
       }
       if (!Array.isArray(value)) {
-        return `${key}:/${value?.toLowerCase()}/`
+        return `${key}:/.*${value?.toLowerCase()}.*/`
       }
       let statement = `${builder.preprocess(value[0], { escape: true })}`
       for (let i = 1; i < value.length; i++) {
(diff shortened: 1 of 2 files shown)


aptkingston:
Does it still work alright if you use special characters in your filter value? Just wondering if things are properly escaped before being passed into the regex. If it works then I don't see any reason to not merge this - it's a lot closer to a real SQL "like".

melohagan:
Tested with special characters and it works fine @aptkingston 

I also fixed a bug where forward slash wasn't being escaped, so the existing starts with filter for '/' will now work.

I had tried to add lucene test cases, but I spent ages on them before I realised that the behaviour in the app was different to what was happening in the tests. 
Maybe there's another layer of escaping somewhere that isn't obvious when only looking at `lucene.spec.ts`

aptkingston (approved):
LGTM!

PR #10369, opened 2023-04-20
Chore - use pro as submodule

adrinr:
## Description
This PR changes the way we interact with budibase-pro. Instead of being an independent repository published via npm or consumed via yarn links, now the pro will be treated as any other package in the budibase monorepo. The main difference is that this code lives in a different repo, as it's private code, but consuming it as a submodule, it will act as any other existing package.
Doing so we have the benefits of having all the packages together, and using properly build tools such as nx and esbuild, and dev tools such as proper debugging and hot reloading.

The main changes in this PR are:
1. Creating the reference to the submodule, it will be checked out in `packages/pro`
2. Changing all the relative references to use this new path
3. Deleting all the scripts and pipelines that were doing the checkout and link "manually"
4. Changing the repo `bootstrap` to link properly the pro package for `account-portal`
5. Removing all the `references` in the `tsconfig.json` files, as now they are not required anymore. Nx handles these dependencies for us, so we don't need to worry. This has a benefit in that we don't need to check if we have pro loaded or not.
6. Adding an extra pipeline step that checks that the updated submodule has been merged properly. This will ensure that we don't commit references to pro that has not been merged and reviewed

### Contributor experience
_**What happens to open source contributors or developers that have the module not loaded?**_
Nothing, for them the experience will be the same as it is now. When we run `yarn install`, the workspace will detect that the package does not exist locally, so it will use the npm repo as it's doing now. No symlinks needed, `yarn workspace` handle this for us out of the box

image

Addresses: 
- https://linear.app/budibase/issue/BUDI-5831/research-spike-npm-less-deployments

## Screenshots
image

## Documentation

Re-reverting https://github.com/Budibase/budibase/pull/10267


mike12345567 (approved):
Previously approved - merging into central `chore/npmless-builds` branch for integration with other work.

PR #9647, opened 2023-02-10
Generate option inclusion constraint

aptkingston:
## Description
This PR fixes a longstanding issue where options inclusion constraints (i.e. the actual "options" for an options field) are not generated when creating a table from CSV. Currently if you do this, the options field has no "options" in its schema - meaning it's useless and you need to add them all manually.

Also snuck in a tiny fix mentioned by a user on Discord, fixing 0 being treated as null for number setting min and max constraints.

## Technical stuff
The diff might look weird and you might wonder how this fixes anything. The issue was the line `table = processed.table`. This was overriding the real table param and preventing mutation in place from working, so the solution was to delete this line. It was not needed anyway as all table schema mutation is done in place, so the `table` reference was already pointing to the most up to date version. 

The other additions are simply preventing null/empty values going into the schema, and sorting the schema. 

## Screenshots
Video showing the current functionality (no options are generated):

https://user-images.githubusercontent.com/9075550/218050548-cd331ec9-441e-41cb-b121-08a67546d0e9.mp4

Video showing new functionality (options are generated and sorted):

https://user-images.githubusercontent.com/9075550/218051292-7c9ec0dd-e54d-4b81-b17c-b5a3a58b3b93.mp4

## Documentation
- [x] I have reviewed the budibase documentatation to verify if this feature requires any changes. If changes or new docs are required I have written them.

PR #10371, opened 2023-04-20
Fix options inclusion constraint regression

aptkingston:
## Description
This PR fixes a recent regression where options inclusions constraints were not being generated when importing data. This meant that importing data with options columns was generating empty schemas.

I originally fixed this bug a few months ago here: https://github.com/Budibase/budibase/pull/9647. This PR had the side effect of breaking auto IDs not incrementing properly when importing data, causing the sample data to always have IDs of 1.

A change was made to restore auto ID functionality here: https://github.com/Budibase/budibase/pull/10141 (master) and here: https://github.com/Budibase/budibase/pull/10162 (develop). Both of these changes effectively undid my original fix, causing inclusion constraints to no longer be generated.

This PR fixes both the original issue while also still allowing auto ID incrementing to work.

The real outcome of all these PRs is the lesson that the way the `importToRows` function works, in combination with a mutated table clone being returned from `inputProcessing` is both confusing and very bug prone.

## Proof
Newly generated app with sample data, showing correct auto IDs:
image

Newly imported CSV, using an options column, showing the inclusion constraint has been generated successfully:
image

### DIFF
--- a/packages/server/src/api/controllers/table/utils.ts
+++ b/packages/server/src/api/controllers/table/utils.ts
@@ -110,19 +110,25 @@ export function importToRows(
   table: Table,
   user: ContextUser | null = null
 ) {
+  let originalTable = table
   let finalData: any = []
   for (let i = 0; i < data.length; i++) {
     let row = data[i]
     row._id = generateRowID(table._id!)
     row.tableId = table._id
+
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):
Looks good, we should probably do something about the mutation, it is confusing!

PR #10379, opened 2023-04-21
Incoming user limits warnings

PClmnt:
## Description
This PR:

1. adds a banner that shows a user that is on the free plan that they are going to be user limited in X number of days. 
2. Adds an inline alert to the users page saying the same thing and linking them out to the plans page

N.B startDate hasn't been added to the license object yet so I tested with an epoch timestamp in the future, (which is where the `in X days ` comes from). That should be added before this PR is merged. 

## Screenshots
Screenshot 2023-04-21 at 13 44 29

image

### DIFF
--- a/packages/builder/src/components/portal/licensing/licensingBanners.js
+++ b/packages/builder/src/components/portal/licensing/licensingBanners.js
@@ -155,7 +155,7 @@ const buildUsersAboveLimitBanner = EXPIRY_KEY => {
     key: EXPIRY_KEY,
     type: BANNER_TYPES.WARNING,
     criteria: () => {
-      return userLicensing.warnUserLimit && Date.now() < startDate
+      return userLicensing.warnUserLimit
     },
     message: `${capitalise(
       userLicensing.license.plan.type
--- a/packages/builder/src/components/portal/licensing/licensingBanners.js
+++ b/packages/builder/src/components/portal/licensing/licensingBanners.js
@@ -4,6 +4,8 @@ import { admin, auth, licensing } from "stores/portal"
 import { get } from "svelte/store"
 import { BANNER_TYPES } from "@budibase/bbui"
 import { Constants } from "@budibase/frontend-core"
+import { capitalise } from "helpers"
+
 import dayjs from "dayjs"
 import relativeTime from "dayjs/plugin/relativeTime"
 dayjs.extend(relativeTime)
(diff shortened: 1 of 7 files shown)


packages/builder/src/components/portal/licensing/licensingBanners.js
```diff
@@ -141,6 +145,36 @@ const buildPaymentFailedBanner = () => {
   }
 }
 
+const buildUsersAboveLimitBanner = EXPIRY_KEY => {
+  const userLicensing = get(licensing)
+  return {
+    key: EXPIRY_KEY,
+    type: BANNER_TYPES.WARNING,
+    criteria: () => {
+      return (
```
PClmnt, line 154: Important logic.

packages/builder/src/pages/builder/portal/users/users/index.svelte
```diff
@@ -81,6 +86,13 @@
       width: "1fr",
     },
   }
+
+  $: showUserLimitAlert =
```
PClmnt, line 90: Same important logic in here as well.

packages/builder/src/components/portal/licensing/licensingBanners.js
PClmnt: (i removed the negation but forgot to remove the comment. Will sort.

packages/builder/src/pages/builder/portal/users/users/index.svelte
```diff
@@ -237,6 +251,25 @@
     <Body>Add users and control who gets access to your published apps</Body>
   </Layout>
   <Divider />
+  {#if showUserLimitAlert}
+    <InlineAlert
+      type="error"
+      onConfirm={() => {
+        window.open("https://budibase.com/pricing/", "_blank")
```
PClmnt, line 258: need to change this to check for account portal access and link to the upgrade page if so

packages/builder/src/components/portal/licensing/licensingBanners.js
```diff
@@ -141,6 +145,35 @@ const buildPaymentFailedBanner = () => {
...
+        get(admin).cloud &&
+        userLicensing.license.plan.model === Constants.PlanModel.PER_USER &&
+        userLicensing.isFreePlan &&
+        userLicensing.quotaUsage.usageQuota.users > 5
+      )
+    },
+    message: `Free plan changes - Users will be limited to ${
```
Rory-Powell, line 161: Can just sub in `utils.capitalise(user.license.plan)` to make this banner agnostic of the free plan and determined completely by the backend

packages/builder/src/pages/builder/portal/users/users/index.svelte
```diff
@@ -81,6 +88,14 @@
       width: "1fr",
     },
   }
+
+  // Per user check
+  $: showUserLimitAlert =
+    $admin.cloud &&
+    $licensing.isFreePlan &&
+    $licensing.license.plan.model === Constants.PlanModel.PER_USER &&
+    $licensing.quotaUsage.usageQuota.users > 5
```
Rory-Powell, line 97: Can we combine this logic with the logic for the banner in the store, so we can say
```js
$licensing.warnUserLimit
```

packages/builder/src/components/portal/licensing/licensingBanners.js
```diff
@@ -141,6 +145,35 @@ const buildPaymentFailedBanner = () => {
...
+      userLicensing.license.quotas.usage.static.users.value
+    } users ${dayjs(
+      userLicensing.license.quotas.usage.static.users.value.startDate
+    ).fromNow()}`,
```
Rory-Powell, line 165: What would be shown when we go past the start date?
Probably we should hide the banner at that point, as we'll start to deactivate then

packages/builder/src/components/portal/licensing/licensingBanners.js
PClmnt: Ah right, I misunderstood when startDate was going to be added. Updated and simplified this!

Rory-Powell (approved):
Brilliant work peter!

PR #10396, opened 2023-04-24
Grid tweaks

aptkingston:
## Description
This is a collection of improvements and fixes to the Grid UI in the data section after initial feedback and further development.

### Row creation changes
Row creation now happens inside the grid by default, instead of via a modal.
image

Where possible, this inline new row functionality will be rendered at the bottom of the grid, where you would naturally expect a new row to be appended. We do this whenever we can tell that we're actually at the end of your data - i.e. when there are no more pages.

Where this is not possible (i.e. when we have more data waiting to be loaded, and we don't know where the "end" is) this new row component will be rendered at the very top, and the new row will be inserted as the very first row rather than the last row:
image

In either case the experience is identical - errors are handled, an underlay covers the rest of the grid to help you focus on the creation experience, and buttons are required to save or cancel the change.

Row creation can still be done via the modal by clicking the expand icon inside the new row component:
image

### Keyboard shortcut changes
- Start adding row: Cmd/Ctrl+Enter
- Save new row: Cmd/Ctrl+Enter
- Abandon row creation: Escape
- Deselect focused cell: Escape
- Bulk delete rows: Delete or Backspace whenever rows are selected

### Small changes
- New column width button to quickly change the size of all columns. Columns can still always be dragged and resized individually.
image
- Update buttons above grid to shrink text and make more compact
image
- Fix copy and paste being able to edit readonly cells
- Fix immediate typing circumventing readonly cell protection
- Fix bulk delete rows button not correctly being cleared after deletion
- Use different background colour for readonly cells to highlight differences
- Use column friendly names in sort dropdown (e.g. users table)
- Remove background colour from sorted column header
- Use shift+scroll instead of ctrl+scroll for horizontal mouse wheel scrolling
- Make bulk delete button more prominent
image
- Use a grab cursor when hovering on column header cells to help drag-to-reorder discovery
- Performance improvements
- Columns can be created inline via a new icon:
image


packages/frontend-core/src/components/grid/stores/ui.js
```diff
@@ -141,18 +146,18 @@ export const initialise = context => {
   })
 
   // Reset selected rows when selected cell changes
-  focusedCellId.subscribe(id => {
-    if (id) {
-      selectedRows.set({})
-    }
-  })
+  // focusedCellId.subscribe(id => {
+  //   if (id) {
```
shogunpurple, line 150: Should we remove the commented code here?

shogunpurple (approved):
LGTM! Some really nice updates here - especially the best of both worlds (top/bottom) with row addition based on whether we can see the bottom or not.

packages/frontend-core/src/components/grid/stores/ui.js
aptkingston: Nice one, forgot to remove it after testing with them commented out!

PR #10404, opened 2023-04-24
Form Block Improvements

Ghrehh:
Implements [this ticket](https://linear.app/budibase/issue/BUDI-6605/form-block-updates)

- Adds a new field configuration component that borrows heavily from the existing input used to configure form fields.
- Users now have the ability to add validation to the fields in the form block and table block side panel as if they were interacting with an ejected form.
- Users can edit the labels and order of fields in the above mentioned blocks also.
- Ability for users to specify the text of the save and delete buttons on form blocks.
- Ability to hide the delete button on table blocks side panel.

- Still need to conditionally hide some config on the form block, will tackle that in a follow up.
- The field configuration component needs to be dried up some, as it was directly copied from the table columns component, will also tackle that in a follow up 

|Form Ordering and Label Editing|Form Validation Editing|Table Block Side Panel Controls|
|---|---|---|
|Screenshot 2023-04-24 at 18 34 24|Screenshot 2023-04-24 at 18 34 31|Screenshot 2023-04-24 at 18 34 48|

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -4435,6 +4435,48 @@
                 "key": "row"
               }
             ]
+          },
+          {
+            "label": "Side panel fields",
+            "type": "fieldConfiguration",
+            "key": "sidePanelFields",
+            "nested": true,
+            "dependsOn": {
+              "setting": "clickBehaviour",
+              "value": "details"
+            }
+          },
+          {
+            "label": "Show side panel delete button",
+            "type": "boolean",
+            "key": "sidePanelShowDelete",
+            "nested": true,
+            "dependsOn": {
+              "setting": "clickBehaviour",
+              "value": "details"
+            }
+          },
+          {
+            "label": "Side panel save label",
+            "type": "text",
+            "key": "sidePanelSaveLabel",
+            "defaultValue": "Save",
+            "nested": true,
+            "dependsOn": {
+              "setting": "clickBehaviour",
+              "value": "details"
+            }
+          },
+          {
+            "label": "Side panel delete label",
+            "type": "text",
+            "key": "sidePanelDeleteLabel",
+            "defaultValue": "Delete",
+            "nested": true,
+            "dependsOn": {
+              "setting": "sidePanelShowDelete",
+              "value": true
+            }
           }
         ]
       },
@@ -4979,7 +5021,7 @@
         "name": "Fields",
         "settings": [
           {
-            "type": "multifield",
+            "type": "fieldConfiguration",
             "label": "Fields",
             "key": "fields",
             "selectAllFields": true
@@ -5017,6 +5059,20 @@
         "section": true,
         "name": "Buttons",
         "settings": [
+          {
+            "type": "text",
+            "key": "saveButtonLabel",
+            "label": "Save Button Label",
+            "nested": true,
+            "defaultValue": "Save"
+          },
+          {
+            "type": "text",
+            "key": "deleteButtonLabel",
+            "label": "Delete Button Label",
+            "nested": true,
+            "defaultValue": "Delete"
+          },
           {
             "type": "boolean",
             "label": "Show save button",
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -4437,7 +4437,7 @@
             ]
           },
           {
-            "label": "Side panel fields",
+            "label": "Fields",
             "type": "fieldConfiguration",
             "key": "sidePanelFields",
             "nested": true,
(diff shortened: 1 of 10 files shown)


packages/client/manifest.json
```diff
@@ -4435,6 +4435,48 @@
...
+            }
+          },
+          {
+            "label": "Side panel delete label",
+            "type": "text",
+            "key": "sidePanelDeleteLabel",
+            "defaultValue": "Delete",
+            "nested": true,
+            "dependsOn": {
+              "setting": "sidePanelShowDelete",
```
aptkingston, line 4477: NAB: We actually should update this to depend on clickBehaviour = details as well. We technically want it to depend on 2 conditions (both details behaviour, and also sidePanelShowDelete) but we can't do both. Choosing the behaviour setting prevents this setting being shown when the behaviour is "run actions" which is the better choice in my opinion. 

Having a setting for the delete button text always present (even if you choose to hide the button) isn't as strange as seeing the setting when you aren't even using a side panel.

packages/client/manifest.json
```diff
@@ -5017,6 +5059,20 @@
         "section": true,
         "name": "Buttons",
         "settings": [
+          {
+            "type": "text",
+            "key": "saveButtonLabel",
+            "label": "Save Button Label",
```
aptkingston, line 5065: NAB: We should try to use consistent casing for settings, and the standard is to only capitalise the first letter of the first word, rather than every word. Would also be preferable to shrink this slightly to fit into a single line if possible.

packages/client/manifest.json
```diff
@@ -5017,6 +5059,20 @@
...
+            "nested": true,
+            "defaultValue": "Save"
+          },
+          {
+            "type": "text",
+            "key": "deleteButtonLabel",
+            "label": "Delete Button Label",
```
aptkingston, line 5072: NAB: casing and length as above.

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
```diff
@@ -188,14 +196,14 @@
...
+              type={getComponentForField(field.name)}
               props={{
-                field,
-                label: field,
-                placeholder: field,
-                disabled,
+                validation: field.validation,
+                field: field.name,
+                label: field.displayName,
+                placeholder: field.name,
```
aptkingston, line 206: We should use the display name for the placeholder as well.

packages/client/manifest.json
```diff
@@ -4435,6 +4435,48 @@
...
+            "type": "fieldConfiguration",
+            "key": "sidePanelFields",
+            "nested": true,
+            "dependsOn": {
+              "setting": "clickBehaviour",
+              "value": "details"
+            }
+          },
+          {
+            "label": "Show side panel delete button",
```
aptkingston, line 4450: NAB: There are the sort of labels that I think we can shrink - e.g. this one could just be "Show delete button".

aptkingston (changes requested):
Code reads well!

I pulled this down to review in depth and encountered a few issues though.

I opened an existing table block in an app I had, pressed the new configure fields button and attempted to add one but the builder immediately crashed:
image
This seems to only happen with the "add column" button, but the "add all columns" button works.

There are issues with the button settings for existing table blocks (works for new table blocks):
- I noticed that the setting value for the save button text displayed as "Save" but in my app the save button was actually labelled "Button"
- The same issue happens with the delete button - when you show the delete button the setting appears with a default value of "Delete" but the real button shows "Button"
- The reason is that any table blocks created before this PR do not honour the default value of the new setting as their definitions do not contain it. I've commented in the code how to fix this.

NAB: We should shrink a few of the setting labels so that they fit into a single line - the "side panel" prefixes are probably redundant since the settings are already nested inside the side panel category.

NAB: As you mentioned yourself it would be good to update some of the drawer text to mention "fields" instead of "columns".

Given the time pressure, let's ignore anything I've flagged as NAB for now and just focus on fixing the couple of bigger issues.

packages/client/manifest.json
Ghrehh: Ah hadn't considered this, good catch, sucks there can't be two conditions.

Ghrehh:
@aptkingston Thanks for the review, that should all be sorted now.

aptkingston (changes requested):
Thanks for the quick updates!

Tested this again and still 2 major showstoppers which we still need to fix:
I noticed in that commit last night you removed the logic from `FormBlock` that handled converting old arrays of plain string field names into the new structure, but we still need it. I've tested this and currently all existing form blocks are broken without this migration.

The default behaviour for table blocks is now show no fields at all in the side panel. This is an issue as they should default to all fields, which is the current behaviour. Don't even worry about accounting for this inside the settings component, as long as the table block component itself knows to default to all fields (if no fields are configured) in the implementation.

aptkingston (approved):

Issue #10403, opened 2023-04-24
[BUDI-6920] Automation history fails in cloud with HTTP 500


aptkingston:
**Checklist**
- [x] I have searched Budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: cheeks, bb

**Describe the bug**
Viewing automation history fails with a HTTP 500, responding with `{"message":"Database name cannot be undefined.","status":500}`. Tested across multiple tenants with multiple apps.

The call that 500s is GET `/api/automations`. 

**To Reproduce**
1. View automation history for any app
2. See HTTP 500 in console

**Other info**
The error only occurs the first time the page loads. If you change to another tab in the overview, then go back to automation history, then it will work fine. Refreshing the page will reveal the error again.

Off the top of my head, it looks like the request is firing before something else happens which is required in order to identify the tenant. This is potentially the `sync` request, judging by the network log.

<img width="634" alt="image" src="https://user-images.githubusercontent.com/9075550/234071773-29a1c59c-9612-4d83-9c8b-28865824c4ba.png">

<sub>[BUDI-6920](https://linear.app/budibase/issue/BUDI-6920/automation-history-fails-in-cloud-with-http-500)</sub>

aptkingston:
@mike12345567 let me know if this is indeed frontend related as the symptoms suggest! We can take a look at it if so. Weird that this is the only request that seems susceptible to this issue though - if the sync call is firing late I would expect loads of endpoints to be struggling to identify tenants.

PR #10406, opened 2023-04-24
Fix automation history first load in multi-tenant environments

mike12345567:
## Description
Fix for #10403 - when app overview loads, sometimes old store doesn't get a valid app ID.

Addresses: 
- https://github.com/Budibase/budibase/issues/10403
- https://linear.app/budibase/issue/BUDI-6920/automation-history-fails-in-cloud-with-http-500

### DIFF
--- a/packages/builder/src/api.js
+++ b/packages/builder/src/api.js
@@ -5,12 +5,16 @@ import {
 } from "@budibase/frontend-core"
 import { store } from "./builderStore"
 import { get } from "svelte/store"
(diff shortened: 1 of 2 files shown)


aptkingston (commented):
Nice one! If we start using the overview store here too then we just need to make sure that the app ID is removed from the store whenever we leave the overview, otherwise we'll continue sending up a wrong app ID header. 

I'll jump on here and just push that up.

mike12345567:
Thanks @aptkingston !

PR #10410, opened 2023-04-25
Chore - Npmless builds

adrinr:
## Description
Branch containing already approved PRs to fix the build and pipeline errors related to npm.

- https://github.com/Budibase/budibase/pull/10369
- https://github.com/Budibase/budibase/pull/10360

Summarising, these changes will:
1. change the budibase-pro repo to be consumed as a submodule
2. change the way we reference packages, making use of the yarn workspaces linking instead of constantly updating all the package versions
3. change the pipelines to keep publishing the versions as they are

Related:
- https://github.com/Budibase/budibase-pro/pull/171
- https://github.com/Budibase/account-portal/pull/318


lerna.json
```diff
@@ -1,8 +1,22 @@
 {
   "version": "2.5.10-alpha.0",
   "npmClient": "yarn",
+  "packages": [
+    "packages/backend-core",
```
mike12345567, line 5: I assume we can change this back once we get account-portal/pro into a standard shape?

adrinr: This is because of the shape of the pro only. Without this will treat as a package the path in `pacakge/pro`, that is not a valid package. Once this out we can refactor pro repo to be a single package, changing this back to `packages/*`

scripts/bumpVersion.js
```diff
@@ -0,0 +1,46 @@
...
+const lernaData = JSON.parse(fileData)
+
+// Get current version and split into major, minor, patch, and alpha components
+const currentVersion = lernaData.version
+const [versionWithoutPrerelease, alpha] = currentVersion.split("-")
+const [major, minor, patch] = versionWithoutPrerelease.split(".").map(Number)
+
+// Calculate new version based on specified version bump
+let newVersion = currentVersion
+switch (versionBump) {
```
Rory-Powell, line 17: Could use `semver` here: https://www.npmjs.com/package/semver
It has lots of useful utilities, but I think the `increment` would be most useful for what we want to achieve

mike12345567 (approved):
LGTM!

scripts/bumpVersion.js
adrinr: Good find, I was actually looking for an existing library... Not sure how I did not find it 😓

scripts/bumpVersion.js
```diff
@@ -7,36 +8,9 @@ const versionBump = process.argv[2] || "patch"
...
-      newAlphaVersion = +alphaVersion + 1
-    }
-    newVersion = `${versionWithoutPrerelease}-alpha.${newAlphaVersion}`
-
-    break
-  default:
-    console.error(`Invalid version bump '${versionBump}' specified.`)
-    process.exit(1)
-}
+const newVersion = semver.inc(currentVersion, versionBump, "alpha")
```
Rory-Powell, line 13: Is the `alpha` ignored if the `versionBump` is not `prerelease`, or how does that work?

adrinr: Yes, it's quite confusing, but it works like this:
```
Prerelease Identifiers

The method .inc takes an additional identifier string argument that will append the value of the string as a prerelease identifier:

semver.inc('1.2.3', 'prerelease', 'beta')
// '1.2.4-beta.0'

command-line example:
```

It's only used for prereleases.

Rory-Powell: That's not clear at all lol, but good that it works!

packages/backend-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@budibase/backend-core",
-  "version": "2.5.10-alpha.3",
+  "version": "0.0.1",
```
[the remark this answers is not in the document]
adrinr: Not needed. On the code we don't bump any other version than the one from lerna, but at release pipeline time we are actually bumping them, as documented in the docs: https://docs.google.com/document/d/1Rvx8dRlG2mmUYYvQE-B0uTU3c0petjkZcO_M14DLB8E/edit
The actual package.json and all references will be updated before the build, so all the versions should be aligned

.github/workflows/release-develop.yml
```diff
@@ -1,21 +1,16 @@
...
+on:
+  push:
+    tags:
+      - v*-alpha.*
+  workflow_dispatch:
+    inputs:
+      tags:
+        description: "Release tag"
+        required: true
+        type: boolean
```
Rory-Powell, line 13: Is this input used?

adrinr: If we want to trigger existing tagged releases. It might be useful for rollbacks, for example

Rory-Powell: I mean more so that it doesn't look to be used in the pipeline? I could be missing something

adrinr: True, this will actually not work. The action already allows us selecting a tag:
image

Rory-Powell (approved):
LGTM!

PR #10413, opened 2023-04-25
Enrich the upgradable version property when fetching the app package

deanhannigan:
## Description

Instead of using the client package.json to determine the current upgradable app version, enrich the app package with the available version upon retrieval. This utilises the updated version resolution from backend-core.

This should alleviate any issues related to querying the client package.json file from the client directly.


aptkingston (approved):
LGTM! Nice one 👌 I think we can actually remove `@budibase/client` as a dependency of the builder now too can't we? It was only used for this I'm pretty sure.

Rory-Powell (approved):
Nice work! Good to have the backend integration.

PR #10417, opened 2023-04-25
Add feature and quota overrides to qa core for cloud tests

Rory-Powell:
## Description

Related:
- https://github.com/Budibase/account-portal/pull/320

QA core:
- Add api for updating license overrides
- Add all features and unlimited quotas to all api tests
- Update logging to better integrate with log levels and pretty printing
- Remove global setup in each test - this is now truly global setup i.e. one account per test run

Types:
- Move the update license request from account portal

Backend-core:
- Update pino library, to bring in better shutdown handling


qa-core/src/account-api/api/AccountInternalAPIClient.ts
```diff
@@ -39,6 +43,13 @@ export default class AccountInternalAPIClient {
         credentials: "include",
       }
 
+      if (options.internal) {
+        requestOptions.headers = {
+          ...requestOptions.headers,
+          ...{ "x-budibase-api-key": env.ACCOUNT_PORTAL_API_KEY },
```
shogunpurple, line 49: shouldn't need to spread here - just add the `"x-budibase-api-key":  env.ACCOUNT_PORTAL_API_KEY` directly

shogunpurple (approved):
LGTM! Awesome that we get the features as a side effect

qa-core/src/account-api/api/AccountInternalAPIClient.ts
Rory-Powell: I had it already like that but it complained because of there is no type definition and the original var doesn't have this one

Issue #10274, opened 2023-04-12
[BUDI-6876] Big Budibase Logo in E-mails


SICUFI:
**Checklist**
- [ X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: Docker compose
  - Budibase Version: 2.5.2

**Describe the bug**
Every email I receive from Budibase has the Budibase logo abnormally large

**To Reproduce**
1. Just trigger an e-mail from an automation for example

**Expected behavior**
I know that brand removing is a premium feature, but I suppose the logo this bis is a bug and not an intended behaviour

**Screenshots**
![image](https://user-images.githubusercontent.com/113527716/231445169-4b06870f-b9a0-4b21-b994-8d6061751902.png)
  

<sub>[BUDI-6876](https://linear.app/budibase/issue/BUDI-6876/big-budibase-logo-in-e-mails)</sub>

SICUFI:
This is the template for Automations in Settings - Email : 
`<tr>
  <td class="email-body" width="570" cellpadding="0" cellspacing="0">
    <table class="email-body_inner" align="center" width="570" cellpadding="0" cellspacing="0" role="presentation">
      
      <tr>
        <td class="content-cell">
          <div class="f-fallback">
					<div style="border: 1px red solid; border-radius: 5px; align:center; vertical-align: middle;"><img src="obscured"><b><i>	{{ company }} - Intranet  </b></i></div><br><br>
            {{ contents }}
          </div>
        </td>
      </tr>
    </table>
  </td>
</tr>`

PR #10419, opened 2023-04-25
Add height to BB logo in core email template

mike12345567:
## Description
Quick attempt at fix for #10274 - some email clients don't respect width, add both height and width. This shouldn't cause any issues in other clients as it is a square image and cannot distort.

Also some clients don't like when a unit is specified for width/height so removing the px: https://stackoverflow.com/questions/20989897/image-style-height-and-width-not-taken-in-outlook-mails.

I've had the dis-pleasure and writing email HTML before so have come against similar issues in the past.

### DIFF
--- a/packages/worker/src/constants/templates/core.hbs
+++ b/packages/worker/src/constants/templates/core.hbs
@@ -16,8 +16,8 @@
               cellspacing="0"
             >
               <img
-                width="32px"
-                height="32px"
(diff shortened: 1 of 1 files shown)


deanhannigan (approved):
LGTM!

shogunpurple (approved):
LGTM!

Issue #10233, opened 2023-04-06
[BUDI-6836] App Status does not immediately display Published info


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: plugins.budibase.app

**Describe the bug**
When managing an app, it does not show the Published status straight away. You have to click a different tab and navigate back.

**To Reproduce**

1. Manage a published app
2. See in Overview there is no published info
3. Click Access tab
4. Click Overview. You can now see *Last published* info

**Expected behavior**
App Status in the app management screen should update in real-time.

**Screenshots**

![Screenshot 2023-04-06 at 15.31.51.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/be6b0dd5-dd4e-44a6-ad2f-e7cd2b1791e2/127c789c-7f9d-46d5-83a5-7b783bd9c485)

![Screenshot 2023-04-06 at 15.32.10.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/efc7d14f-9550-4c4c-8017-61c944d3ec73/b67ccde5-8fe7-4721-8590-2546c661814f)

**Got this after clicking the Access tab, and back again**

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6836](https://linear.app/budibase/issue/BUDI-6836/app-status-does-not-immediately-display-published)</sub>

dewdeded:
Yeah, we have the same issue. Since over > 1 year.

PR #10420, opened 2023-04-25
Overview published status updating after app sync

deanhannigan:
## Description

The request to fetch the application deployments was not waiting for the app package to be fetched in the page layout.
Depending on the timing, this would either result in a successful request with the db context having the appropriate app, or the appId would be missing and the db request would fail silently. 

The page now waits until the application context is fully loaded and the context error will now be surfaced if this were to happen again.

Addresses: 
- https://github.com/Budibase/budibase/issues/10233


mike12345567 (approved):
Looks good - do we want to rebase this to `develop` as the release is going out soon - I'm not sure this is critical enough for it to go into master and require a patch release.

PR #10423, opened 2023-04-26
Standardise width of design section side panels to fix drawer width issues

aptkingston:
## Description
Currently only the component settings panel is flagged as "wide" whereas the screen, theme and navigation panels are not. Not only does this cause the preview width to jump when swapping between, it causes a gap between the action drawer and the screen settings panel.

This PR ensures all panels are marked as wide to fix that issue and keep things consistent.


shogunpurple (approved):
LGTM!

PR #10425, opened 2023-04-26
Feature/chatgpt automationblock

shogunpurple:
## Description
A ChatGPT/OpenAI Automation block, allowing users to pass prompts to chatGPT and use the response in automations. Requires an OpenAI API key to work - we are still on the waitlist for GPT4 so only support 3.5 turbo at the moment. But it's a configurable option in the automation so easily uncommented and turned on when we gain GPT4 access.

## Screenshots
Screenshot 2023-04-26 at 16 12 07

### Demo Video - translating text to 3 different languages and adding to a BB table
https://user-images.githubusercontent.com/11256663/234620577-b916fce9-d61d-4f93-8b83-a32dc1ac4283.mov

## Documentation


packages/server/src/automations/steps/openai.ts
```diff
@@ -0,0 +1,105 @@
...
+  AutomationStepType,
+  AutomationIOType,
+} from "@budibase/types"
+import * as automationUtils from "../automationUtils"
+import environment from "../../environment"
+
+enum Model {
+  GPT_35_TURBO = "gpt-3.5-turbo",
+  // will only work with api keys that have access to the GPT4 API
+  // GPT_4 = "gpt-4",
```
mike12345567, line 15: Should we allow it incase people have access to it? Considering it can't be long until its generally available and we might forget about it.

mike12345567 (approved):
LGTM!

Rory-Powell (approved):

aptkingston (approved):
LGTM! Nice one Marty 👌 is there something in here that will disable this in cloud? Since I'm guessing we aren't going to pay for an OpenAI API key for everyone to use haha.

packages/server/src/automations/steps/openai.ts
shogunpurple: Yep, no reason not to! I was planning on holding off merging this until we can provide GPT-4 in cloud anyway.

PR #10429, opened 2023-04-27
More grid improvements

aptkingston:
## Description
This is a few new improvements to the grid after further review and feedback:

- It's much easier to see if multi select and relationship cells have more options can be displayed. Previously cells were wrapped and hidden so you didn't know if that was the end of them. Now, when the grid is using "small" row height, they won't wrap. This means you'll see a cutoff edge which tells you that there are more options. Subtle change but does provide a lot of value:
image
- Always properly ignore keypresses which originate from modals or the user side panel. This fixes issues noticed by @melohagan where certain keypresses originating outside the grid (like a modal) were still being handled by the grid when they shouldn't be
- Readonly cells (i.e. autocolumns) or the uneditable user fields, now have a grey border when selected, indicating that they cannot be edited
image
- When creating rows, autocolumns show placeholder text (and a grey border) to explain that they do not need filled in for row creation
image
- When creating rows, both the cells in the new row and the save button are disabled while the request is in flight. There are no loading spinners, just a disabled overlay which gently fades in

https://user-images.githubusercontent.com/9075550/234802632-09553cf5-c2a0-4715-9843-77373aa38128.mov

- Added a keybind indicator for adding new rows, helping discovery of keyboard shortcuts. This show up as Ctrl/Cmd depending on platform.
image
- Added keybind indicators to the save and cancel buttons when adding a row, following on from the keyboard shortcut that opened the new row in the first place
image


melohagan (approved):
LGTM!

shogunpurple (approved):
LGTM! Nice work. Happy enough for this one to go in ahead of code freeze as it's fairly low risk.

PR #10432, opened 2023-04-27
Remove automation logs from app exports

mike12345567:
## Description
Removing automation logs from app exports as they are not useful - they take up space and the imported app won't have ran any automations yet. I have added it as a flag so that this does not affect backups (as you would want to maintain logs for backups).


shogunpurple (approved):
LGTM! nice work

PR #10433, opened 2023-04-27
Stop Multiple onClick Events from Firing

Ghrehh:
Fixes: https://linear.app/budibase/issue/BUDI-6349/multiple-clicks-on-form-submit-button-fires-multiple-duplicated

### DIFF
--- a/packages/client/src/components/app/Button.svelte
+++ b/packages/client/src/components/app/Button.svelte
@@ -19,12 +19,12 @@
   export let active = false
 
   const handleOnClick = async () => {
-    if (handlingOnClick) {
-      return
+    handlingOnClick = true
+
+    if (onClick) {
+      await onClick()
     }
 
-    handlingOnClick = true
-    await onClick()
     handlingOnClick = false
   }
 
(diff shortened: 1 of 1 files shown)


melohagan (approved):
LGTM!

packages/client/src/components/app/Button.svelte
```diff
@@ -16,6 +18,16 @@
   export let icon = null
   export let active = false
 
+  const handleOnClick = async () => {
+    if (handlingOnClick) {
+      return
+    }
+
+    handlingOnClick = true
+    await onClick()
```
aptkingston, line 27: When there are no actions specified, `onClick` will be `null` so I'm fairly sure this will throw an error. We should check if it's null, or add in optional chaining.

packages/client/src/components/app/Button.svelte
```diff
@@ -16,6 +18,16 @@
   export let icon = null
   export let active = false
 
+  const handleOnClick = async () => {
+    if (handlingOnClick) {
```
aptkingston, line 22: Instead of returning here, could we not use the `handlingOnClick` flag to disable the button instead? Then you also get visual feedback that something is happening, and it won't fire any more onclick events while disabled. I'm fairly certain all you need to do is pass in the actual `disabled` attribute and spectrum CSS should have styles for it.

aptkingston (changes requested):
One small bug which I think needs fixed, and one small improvement suggestion!

Ghrehh:
@aptkingston Done, please review this again when you get a chance.

aptkingston (approved):
LGTM!

PR #10435, opened 2023-04-27
Fix number cell issue with grids and external SQL tables

aptkingston:
## Description
Fixes an issue raised by @melohagan where number cells inside SQL tables could not be edited due to being sent up as strings. The datasource plus API accepts strings for numbers, and it works fine for internal tables, but with a SQL table the request will go on to fail with a SQL type error.

This change parses number cells as numerical values before sending them up.

I've also added click outside handlers to cells which grow when selected so that they can be easily closed when creating new rows.


melohagan (approved):
LGTM!

PR #10451, opened 2023-04-28
is empty lucene query with OR fix

melohagan:
## Description
As per StackOverflow, we were missing the `*:*` from our query, although this was only an issue in a query with an OR statement.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6956/match-any-filters-not-working-as-expected

## Screenshots
Screenshot 2023-04-28 at 16 22 01


shogunpurple (approved):
LGTM!

Issue #10400, opened 2023-04-24
[BUDI-6919] Use trash can icon instead of "X" in dropzone component for deletion


shogunpurple:
[https://discord.com/channels/733030666647765003/872822618238095360/1098890279014121512](https://discord.com/channels/733030666647765003/872822618238095360/1098890279014121512)

**COS**

* Update dropzone to use a trashcan icon instead of an "X" for deletion of attachments.

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6919](https://linear.app/budibase/issue/BUDI-6919/use-trash-can-icon-instead-of-x-in-dropzone-component-for-deletion)</sub>

PR #10459, opened 2023-05-01
Fix/10400 use trashcan icon

mslourens:
## Description
Replaces the close icon with a trash  can in the dropzone

Addresses: 
- #10400 

## Screenshots
image

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "2.5.6-alpha.41",
+  "version": "2.5.6-alpha.42",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "packages": ["packages/*"],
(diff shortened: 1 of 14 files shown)


aptkingston:
Hey @mslourens, and thanks! I see you've used a remix icon - the remix icons are actually a legacy leftover (I think they're only used for the `Icon` app component, otherwise we would remove the dependency altogether) and everywhere else we use spectrum icons (via the `Icon` BBUI component).

Would you mind updating this to use to a spectrum icon instead? The previous syntax here is fine, just the icon name needs changed.
https://spectrum.adobe.com/page/icons

Thanks!

mslourens:
I wanted to use the Icon component from BBUI project, but couldn't find a trashcan icon. The spectrum css icon library doesn't contain a bin or trashcan icon, as far as I can see. That's why I opted for this icon instead.

aptkingston:
It can be tricky to find the correct name in spectrum sometimes! They're called "delete" in this case :)
image

Issue #10131, opened 2023-03-28
[BUDI-6789] Import Data: Empty date is not allowed


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.4.31
  * App Version: 2.4.31

**Describe the bug**
When importing into an existing table, if a Date column is empty, the modal shows **Failure** against that column, even when importing the same data that was exported.

**To Reproduce**

1. Create new app with sample data
2. Export Employees
3. Import into Employees, selecting export
4. See that *End Date* shows failure because *End Date* is empty in the CSV.

**Expected behavior**
Empty dates should be allowed in a CSV import.

**Screenshots**

![Screenshot 2023-03-28 at 08.15.35.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/7e816331-4153-4192-a013-6736c6c20cbb/281d74fc-66a7-4ec7-83e3-b829bf8fc0fa)

![Screenshot 2023-03-28 at 08.15.07.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/4fd5cd2b-2f87-42e1-b989-7e6410d9b371/3c723ad4-b753-4fd7-9b56-c5a7e1975547)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6789](https://linear.app/budibase/issue/BUDI-6789/import-data-empty-date-is-not-allowed)</sub>

melohagan:
I found another issue. Try importing data into the **Expenses** sample table (1 row):

[expenses.csv](https://github.com/Budibase/budibase/files/11086738/expenses.csv)

You get an error caused by the attachment field. The value in the CSV for the attachment field is: `[]`

![Screenshot 2023-03-28 at 08 48 26](https://user-images.githubusercontent.com/101575380/228165874-f77bf906-58af-4346-bd2a-4702a1d7dd77.png)

![Screenshot 2023-03-28 at 08 48 52](https://user-images.githubusercontent.com/101575380/228166002-2e9d333a-0162-489c-9492-f7cc885fbc0d.png)


Could be related to this issue as well: https://github.com/Budibase/budibase/issues/9142

PR #10462, opened 2023-05-02
Fix - Budi 6789 import data empty date is not allowed

adrinr:
## Description
Fix import and export CSV to handle empty values. New behaviour:
1. On exporting empty values, the csv will contain empty values instead of empty strings
2. On importing CSVs with empty values, these will be respected instead of casting them as empty strings

This casting to empty string was causing issues when casting optional types values, fe. datetimes

Addresses: 
- https://github.com/Budibase/budibase/issues/10131

### DIFF
--- a/packages/server/src/utilities/csv.ts
+++ b/packages/server/src/utilities/csv.ts
@@ -1,6 +1,19 @@
 import csv from "csvtojson"
 
 export async function jsonFromCsvString(csvString: string) {
-  const result = await csv().fromString(csvString)
+  const castedWithEmptyStrings = await csv().fromString(csvString)
+  if (!castedWithEmptyStrings.length) {
+    return []
+  }
+
+  // By default the csvtojson library casts empty values as empty strings. This is causing issues on conversion.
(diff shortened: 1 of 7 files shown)


packages/server/src/utilities/csv.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function jsonFromCsvString(csvString: string) {
+  const castedWithEmptyStrings = await csv().fromString(csvString)
+  if (!castedWithEmptyStrings.length) {
+    return []
+  }
+
+  // By default the csvtojson library casts empty values as empty strings. This is causing issues on conversion.
+  // ignoreEmpty will remove the key completly if empty, so creating this empty object will ensure we return the values with the keys but empty values
+  const emptyObject = Object.keys(castedWithEmptyStrings[0]).reduce(
```
adrinr, line 11: This is quite hacky but it is the cleanest way I found to support it, without migrating to another library

mike12345567 (approved):
LGTM!

PR #10468, opened 2023-05-02
Designer user side panel UX updates.

deanhannigan:
## Description

Without focus on the user search bar, the panel will display all `active users`, `pending users` and `groups` that have been granted an access level. This addresses the scenario where all pending tenant invites were displayed in the access panel, whether they have been granted explicit access to an application or not.

When the focus is placed in the search input within the user invite panel
- Up to 100 pending tenant invitations will be listed
- All available groups will be displayed (remains invisible if feature is locked)
- Displays the first page of users currently registered in the system.

Addresses: 
- BUDI-6878
- Fixes an issue where entities that had their access removed would display "X Remove" instead of returning to a blank state


packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
```diff
@@ -52,15 +55,32 @@
...
+    if (!prodAppId) {
+      return
+    }
+
     appInvites = await getInvites()
-    if (!query || query == "") {
-      filteredInvites = appInvites
+
+    //On Focus behaviour
+    if (!filterByAppAccess && (!query || query == "")) {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:65]: NAB but just a quick FYI that `(!query || query == "")` is redundant - `!query` by itself will already account for empty strings.

packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
```diff
@@ -481,6 +509,8 @@
                   </div>
                 </div>
                 <div class="auth-entity-access">
+                  <!-- {#key filteredInvites} -->
+                  <!-- {invite.info.apps?.[prodAppId]} -->
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:513]: Some commented out code left over.

aptkingston (approved):
LGTM! Just at a high level, is there any reason we can't just use the already existing `searchFocus` variable to determine whether to filter by access or not?

deanhannigan:
> LGTM! Just at a high level, is there any reason we can't just use the already existing `searchFocus` variable to determine whether to filter by access or not?

I wanted to separate the focus and data filtering flags in the hopes that `filterByAppAccess` would convey a little more context around the state we are trying to achieve across the user, group and invite searches. 

I do see what you mean about the flags being, essentially, interchangeable at this point. I'm happy to swap it to the use the focus flag instead, if you'd prefer?

PR #10470, opened 2023-05-02
CLI - fully disable pino logger

mike12345567:
## Description
Re-writing the disabling of pino/logging - it seems that the pino logger is causing a variety of issues in the built CLI version - easier to offer an environment variable for backend-core which completely removes the logger.


Rory-Powell (approved):
LGTM

PR #10481, opened 2023-05-04
Fixes around user limit banner reappearing and copy adjustment

PClmnt:
## Description
As title, this fixes:

- An issue where the User above limit warning banner would reappear even if the close button was clicked
- Fixed copy in the users section


Rory-Powell (approved):
🙌

PR #10482, opened 2023-05-04
Fix for inclusion parsing for arrays and options.

deanhannigan:
Addresses: 
- Array/Opts Inclusion parsing was not correctly building a valid set of unique values when importing rows to the DB. This manifested during CSV imports and the Sample Data table generation.
- Table schema mutation after importing rows was not being persisted with:
  - `Array fields` or `Options`. Any of the inclusion mutations/additions above were lost.
  -  Rows with `autoId` type fields. The `lastId` element on the field schema was not saved after processing all the new rows. If a user imported 50 rows, the autoId fields would correctly reflect the incrementing row value for the whole batch. However, if the lastId was 20, every subsequent import would start the autoId increments from 20.
- BUDI-6828 - Array/Option row values could not be re-imported correctly. The value would be stored as a string representation instead.
- Added tests to cover the scenarios above.
- View UI updated to hide the `Edit` button.


packages/server/src/api/controllers/table/utils.ts
```diff
@@ -129,17 +129,19 @@ export function importToRows(
...
     for (const [fieldName, schema] of Object.entries(originalTable.schema)) {
+      const rowVal = Array.isArray(row[fieldName])
+        ? row[fieldName]
+        : [row[fieldName]]
       if (
         (schema.type === FieldTypes.OPTIONS ||
           schema.type === FieldTypes.ARRAY) &&
-        row[fieldName] &&
-        (!schema.constraints!.inclusion ||
-          schema.constraints!.inclusion.indexOf(row[fieldName]) === -1)
```
mike12345567, line 137: Just to confirm - what happens here if there are no `constraints` or `inclusion` properties - I suppose it just creates an array with the `rowVal` in it, then gets the uniques from that?

mike12345567 (approved):
Looks good - just one clarification!

aptkingston (approved):
LGTM but just added another suggestion to Mikes existing question.

PR #10483, opened 2023-05-04
Fix for row import failure caused by empty attachment parsing.

deanhannigan:
## Description

When we export rows with no attachments, the field value contains a string representation of an empty array `"[]"`.
On import back into the system, this value is not accommodate and it will cause the row import to fail.

Addresses: 
- Empty attachment string value causing row import to fail.


mike12345567 (approved):
Just one minor comment about the use of `logAlert` over `console.error` but all good!

PR #10487, opened 2023-05-04
Potential fix for automation log issue

mike12345567:
## Description
Fixing some un-identified listeners, queue messages not being correctly printed, also improving the mechanism for detecting if in the main thread or not.

We were getting a few `undefined` messages as part of the queue listeners, this attempts to rectify that, as well as adding a new environment variable that can be used to detect if a thread is a fork. The default already was `process.env` so spreading it in and adding the extra environment variable shouldn't cause a problem.


Rory-Powell (approved):
LGTM!

Issue #10490, opened 2023-05-05
Can't edit write access


ben-biber:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: <method> docker compose
  - Budibase Version: <version> 2.6.1
  - App Version: <version> 2.6.2
 
**Describe the bug**
Can't edit write access to a data table

**To Reproduce**
1. Go to 'Data'
2. Click on 'Add Source'
3. Create an 'BudiBase DB' table
4. Click on 'Access'
5. You can only edit read access

**Expected behavior**

I earlier versions it was possible to edit the write access
  
**App Export**
[Sommerfreizeit 2023-export-1683272448680.tar.gz](https://github.com/Budibase/budibase/files/11404166/Sommerfreizeit.2023-export-1683272448680.tar.gz)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: safari
 - Version 16.4

shogunpurple:
Hi @ben-biber - we are currently looking into this one and will have a fix out ASAP.

PR #10491, opened 2023-05-05
Grid improvements and fixes

aptkingston:
## Description
This PR contains further improvements to the grid based on feedback. It also addresses some unintended functionality where keyboard shortcuts could circumvent readonly cells.

Targeting master since I'm keen to get these out sooner rather than later, and some of these are indeed bug fixes rather than just improvements.

### Accessibility improvements for small screens
A significant improvement is the ability to work on much smaller screen sizes. The grid will now function on resolutions as small as 1280x720 (and even smaller):
image

To support this, some buttons collapse to icons which show tooltips on hover (on small screens only):
image

Row bulk deletion has been moved inline, instead of the button in the top right. Now when rows are selected, a delete icon will appear beside each row. Clicking any of these icons will prompt for bulk deletion:
image

### Auto column discoverability
2 changes have been made to support discovery of auto columns.

All auto columns now use the same icon in the grid header cells:
image

Icons have been added to the "hide columns" dropdown, showing the data type of each column. With the new dedicated icon for auto columns, this makes it much faster to tell which columns are auto columns:
image

### Other changes
- Fix issues with not honouring the delete rows flag
- Fix keyboard shortcuts being able to add rows when you shouldn't be able to
- Fix selecting rows with the keyboard when you shouldn't be able to
- Hide add column icon when you shouldn't be able to add columns
- Improve z-index of various elements to fix edge cases
- Move new row save and cancel buttons more left to avoid cases of overlapping dropdowns 
- Fix not being able to export selected rows, as selected rows were cleared when clicking the export button
- Improve other aspects of the "click outside" handler to fix issues with download-js, and improve click handling with modals

### DIFF
--- a/packages/frontend-core/src/components/grid/controls/HideColumnsButton.svelte
+++ b/packages/frontend-core/src/components/grid/controls/HideColumnsButton.svelte
@@ -1,6 +1,7 @@
 <script>
   import { getContext } from "svelte"
-  import { ActionButton, Popover, Toggle } from "@budibase/bbui"
+  import { ActionButton, Popover, Toggle, Icon } from "@budibase/bbui"
+  import { getColumnIcon } from "../lib/utils"
 
   const { columns, stickyColumn, compact } = getContext("grid")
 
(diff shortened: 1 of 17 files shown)


packages/frontend-core/src/components/grid/controls/HideColumnsButton.svelte
```diff
@@ -47,25 +48,34 @@
...
+        <div class="column">
+          <Icon size="S" name={getColumnIcon($stickyColumn)} />
+          {$stickyColumn.label}
+        </div>
         <Toggle disabled size="S" value={true} />
-        <span>{$stickyColumn.label}</span>
       {/if}
       {#each $columns as column}
+        <div class="column">
+          <!--{#if column.schema.autocolumn}-->
```
shogunpurple, line 69: Can remove comment

shogunpurple (approved):
LGTM!

mike12345567:
Also fixes: https://github.com/Budibase/budibase/issues/10490

shogunpurple:
@aptkingston nice work on the quick fix. Is there any way for us to cover this one with a unit test for regression?

aptkingston:
> @aptkingston nice work on the quick fix. Is there any way for us to cover this one with a unit test for regression?

Don't think it's unit testable - it would literally just be asserting that a svelte store works. I moved around how certain props were handled and this was just one reference I forgot to update (but I've searched and there aren't any more references to the old syntax now).

It would probably fit the criteria for a QA wolf integration test though, to verify all the basic features of the grid (and buttons above it).

mike12345567 (approved):
LGTM!

PR #10508, opened 2023-05-08
BUDI-6158 - Extract components

adrinr:
## Description
Refactoring edit group component to split into smaller components. Currently we have all of them in a single components, with different user fetches and other actions that make it quite confusing, error prone and hard to test.

This PR does not contain any change rather than moving and reorganising code around. That's done to make the required iterations easier and safer

Screenshot 2023-05-08 at 09 44 14


aptkingston (approved):
LGTM! Always happy to see these big pages broken down into more modular components 👍

PClmnt (approved):

Issue #10498, opened 2023-05-05
[BUDI-6977] Google sheets should support pagination


melohagan:
Google sheets is currently missing pagination:

![paginate.gif](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/ad3a5ca0-944c-4fcb-b4a0-f2102bf4b1a0/8e152dbe-df17-4cf3-8571-64f5171ebd01)

We pass pagination information into the `read` method of the connector, so we just need to pass that onto the `getRows` method of the client library which has pagination support built in:

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/3d37d1f6-3c9e-4f13-8365-5ec1bdd8633e/fd7d4ca4-284a-47b0-af0f-a710b26aef9b)

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/c064a3f3-7f6c-4ffa-b1d4-97072451dcf8/ae3a9074-8b4a-4122-9d5b-3505339674b8)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6977](https://linear.app/budibase/issue/BUDI-6977/google-sheets-should-support-pagination)</sub>

PR #10525, opened 2023-05-09
Add pagination to Google Sheets

melohagan:
## Description
Added pagination for Google Sheets - need to test in release to get access to **https** Google SSO callback.

Addresses: 
- https://github.com/Budibase/budibase/issues/10498

### DIFF
--- a/packages/server/src/integrations/googlesheets.ts
+++ b/packages/server/src/integrations/googlesheets.ts
@@ -15,7 +15,7 @@ import {
 } from "@budibase/types"
 import { OAuth2Client } from "google-auth-library"
 import { buildExternalTableId, finaliseExternalTables } from "./utils"
-import { GoogleSpreadsheet } from "google-spreadsheet"
+import { GoogleSpreadsheet, GoogleSpreadsheetRow } from "google-spreadsheet"
 import fetch from "node-fetch"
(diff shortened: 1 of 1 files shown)


packages/server/src/integrations/googlesheets.ts
```diff
@@ -434,7 +434,19 @@ class GoogleSheetsIntegration implements DatasourcePlus {
...
       const sheet = this.client.sheetsByTitle[query.sheet]
-      const rows = await sheet.getRows()
+      let rows: GoogleSpreadsheetRow[] = []
+      if (query.paginate) {
+        let page: number =
+          typeof query.paginate.page === "number"
+            ? query.paginate.page
+            : parseInt(query.paginate.page || "1")
+        rows = await sheet.getRows({
+          limit: query.paginate.limit,
```
aptkingston, line 444: Should we fall back to some sensible value for limit in case it isn't specified? Something like 50?

aptkingston (approved):
LGTM! Just one small comment.

packages/server/src/integrations/googlesheets.ts
melohagan: Good call. I went for 100 as this matches the pagination for the Internal DB

PR #10540, opened 2023-05-10
Onboarding Improvements

Ghrehh:
## Description
Some shared packages changes that facilitate the onboarding changes in [this ticket](https://linear.app/budibase/issue/BUDI-6948/adapt-sign-up-onboarding-flow)

- Some changes to the button component to allow it to be used in forms as the submit input without swallowing inputs.
- Exposed the fancyform error as a separate component.
- Broke the testimonial out into a separate component.


packages/bbui/src/Button/Button.svelte
```diff
@@ -32,7 +43,7 @@
   class:disabled
   class="spectrum-Button spectrum-Button--size{size.toUpperCase()}"
   {disabled}
-  on:click|preventDefault
```
Ghrehh, line 35: Preventing default on click here breaks form submission if you try and use this button as the submit input, this work around is an attempt to allow it to be used with forms

packages/bbui/src/Button/Button.svelte
```diff
@@ -2,6 +2,8 @@
   import "@spectrum-css/button/dist/index-vars.css"
   import Tooltip from "../Tooltip/Tooltip.svelte"
 
+  export let type
+  export let preventDefault = true
```
aptkingston, line 6: Unused export here.

aptkingston (approved):
Just one unused export leftover from PR feedback but LGTM apart from that.

PR #10542, opened 2023-05-11
Adds account locking if user limit is exceeded

jvcalderon:
## Description

Once the new pricing plans was applied, some users will need to update their accounts if the new user limit is exceeded. This PR adapts the UI to this behaviour.

Addresses: 

https://linear.app/budibase/issue/BUDI-6863/take-apps-offline-when-user-limit-reached

## App Export

## Screenshots

If limit is not exceeded, the behaviour should be normal. Once the limit is exceeded...

For an 'app user':

https://github.com/Budibase/budibase/assets/2486322/efe72c42-e80a-4273-a69d-bcc0b24890f0

For a developer/admin/owner:

https://github.com/Budibase/budibase/assets/2486322/0c55602b-7bd4-47fd-8d87-ad88e1e1b7a7

## Documentation


packages/builder/src/stores/portal/licensing.js
```diff
@@ -198,11 +198,11 @@ export const createLicensingStore = () => {
...
         const userLimit = userQuota?.value
         const userCount = usage.usageQuota.users
-        const userLimitReached = willReachUserLimit(userCount, userLimit)
-        const userLimitExceeded = willExceedUserLimit(userCount, userLimit)
+        const userLimitReached = usersLimitReached(userCount, userLimit)
+        const userLimitExceeded = usersLimitExceeded(userCount, userLimit)
         const days = dayjs(userQuota?.startDate).diff(dayjs(), "day")
         const userLimitDays = days > 1 ? `${days} days` : "1 day"
-        const warnUserLimit = userQuota?.startDate && userLimitExceeded
+        const errUserLimit = userQuota?.startDate && userLimitExceeded
```
jvcalderon, line 205: Some doubts about `warnUserLimit` var:

1. I've assumed that the proper name should be now `errUserLimit`; it is not a warning anymore; if limit is exceeded, account is directly locked. Rigth?
2. What's the point to check `userQuota?.startDate`? Do we still needing that? Why?

packages/builder/src/stores/portal/licensing.js
```diff
@@ -130,11 +130,11 @@ export const createLicensingStore = () => {
...
     },
-    willReachUserLimit: userCount => {
-      return willReachUserLimit(userCount, get(store).userLimit)
+    usersLimitReached: userCount => {
+      return usersLimitReached(userCount, get(store).userLimit)
     },
-    willExceedUserLimit(userCount) {
-      return willExceedUserLimit(userCount, get(store).userLimit)
+    usersLimitExceeded(userCount) {
+      return usersLimitExceeded(userCount, get(store).userLimit)
```
jvcalderon, line 137: Correct me if I'm wrong, but I've assumed that the prefix 'will...' is a reference to a future event, but now if limit is exceeded, is exceeded now.

packages/builder/src/stores/portal/licensing.js
```diff
@@ -198,11 +198,11 @@ export const createLicensingStore = () => {
         const userQuota = license.quotas.usage.static.users
         const userLimit = userQuota?.value
         const userCount = usage.usageQuota.users
-        const userLimitReached = willReachUserLimit(userCount, userLimit)
-        const userLimitExceeded = willExceedUserLimit(userCount, userLimit)
+        const userLimitReached = usersLimitReached(userCount, userLimit)
+        const userLimitExceeded = usersLimitExceeded(userCount, userLimit)
```
jvcalderon, line 202: What is the difference between this check:

`usersLimitReached(userCount, userLimit)`

And this other one in the same file:

`usersLimitReached(userCount, get(store).userLimit)`

Why do we need to check with static value and the value in the store? Do we still needing this? What cases?

packages/builder/src/components/portal/licensing/licensingBanners.js
```diff
@@ -146,23 +145,19 @@ const buildUsersAboveLimitBanner = EXPIRY_KEY => {
...
-      userLicensing.license.plan.type
-    )} plan changes - Users will be limited to ${
-      userLicensing.userLimit
-    } users in ${userLicensing.userLimitDays}`,
+    message: "Your Budibase account is de-activated. Upgrade your plan",
     ...{
-      extraButtonText: "Find out more",
+      extraButtonText: "View plans",
       extraButtonAction: () => {
         defaultCacheFn(ExpiringKeys.LICENSING_USERS_ABOVE_LIMIT_BANNER)
```
jvcalderon, line 159: Not sure about this. As far as I see, the `ExpiringKey` is a way to avoid to show the banner again and again, but it doesn't seems to be adapted to be shown for each user; once its shown to the first one, the rest admins will receive it?

packages/builder/src/stores/portal/licensing.js
Rory-Powell: The `will` was added here as the users might not actually be exceeded. This is mostly used whenever we are in the process of adding new users via the portal, so that we can say 'the limit will be reached if you add 1 more user', so that we can show / hide the inputs for creating additional users. Same for CSV import.

packages/builder/src/stores/portal/licensing.js
Rory-Powell: the main idea is that we expose some functions from the store:

`willReachUserLimit`
`willExceedUserLimit` 

These both take an argument of a user count. This is compared against the user limit, and the appropriate return value given. 

These are used by the components that create users
- `AddUserModal`
- `ImportUsersModal`

Where these modals can supply a new count of what the total will be, so that they can pre-emptively show warnings / disable inputs. 

Also in the store we have the usage of the above functions, for creating constants for 
- `userLimitReached`
- `userLimitExceeded`

Rather than being pre-emptive calculations, these are real indicators if this user limit has been reached with the real users count. These are used for
- Showing the upgrade modals for adding new users
- Showing the user limit banner

The functions just default to grabbing the current user limit from the store
```ts
    willReachUserLimit: userCount => {
      return willReachUserLimit(userCount, get(store).userLimit)
    },
```
while the creation of the constants supply the limit they have already have access to as the store is being populated for the first time
```ts
const userLimitReached = willReachUserLimit(userCount, userLimit)
```

packages/builder/src/components/portal/licensing/licensingBanners.js
Rory-Powell: The expiration is stored in local storage, so that if you dismiss it you will see it again in 24 hours. If you log into another browser you should see it again.

packages/builder/src/stores/portal/licensing.js
Rory-Powell: 1. Makes sense
2. We only set the `startDate` in certain circumstances:

https://github.com/Budibase/account-portal/blob/87c1dbe29c6a1f49a62ab2c472c557a1a76e58da/packages/server/src/sdk/licensing/quotas.ts#L295-L298
```ts
    account.hosting === Hosting.CLOUD &&
    plan.model === PlanModel.PER_USER && 
    quotas.usage.static.users.value !== UNLIMITED &&
    env.USER_LIMIT_START_DATE_ISO
 ```
 i.e.
- only for cloud
   - never shows in self host 
- only for per user plans 
   - never shows for people on existing plans that use day pass pricing 
- only when it's not unlimited
   - never show for people on unlimited users 

The start date acts as a whitelist flag for us here, basically saying 'the quota is only applicable under these circumstances'. In hindsight it could have been split into two fields, one for the actual date and another for if the date is applicable to the current user. But just negating the date altogether seemed like a simple way to achieve this. 

I think it would be nice if we could update the code to now be date aware as well, so that we can deploy prior to the date being passed and have the UI automatically react, rather than us being locked down to deploying on that date. 

if we can update to this 

```ts
const warnUserLimit = userQuota?.startDate && userLimitExceeded
const errUserLimit = warnUserLimit && Date.now() >= new Date(userQuota.startDate).getTime()
```

Then we can safely use the `errUserLimit` for showing the new modal / messaging, while keeping in the warnings in the current code. 

We can then either
- Target this PR at master, and bypass the main release process so that our current date of 1st june is correct, or
- Target at develop instead, and update account portal to change the date to 5th june, when the release is going out

packages/builder/src/components/portal/licensing/AccountLockedModal.svelte
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  export function hide() {
+    modal.hide()
+  }
+</script>
+
+<Modal bind:this={modal} on:hide={modal}>
+  <ModalContent
+    title="Your account is currently locked"
```
Rory-Powell, line 19: Think it's a bit confusing how we mix in two terminologies here
- account is locked
- account has been de-activated

Might be worth asking andy on a unified terminology to use here

packages/builder/src/components/portal/licensing/licensingBanners.js
jvcalderon: Ahh. Great. 👍

packages/builder/src/stores/portal/licensing.js
Rory-Powell: Alternatively we can keep as is, as I see you've already updated the warning banners. Doesn't make that much sense to keep in the old code given it will become immediately redundant when the date passes and we'll have to remove it. 

We'll just need to update the `USER_LIMIT_START_DATE_ISO` env var in account portal to reflect the next release date instead of june 1st

As for the `startDate`, we could either
1) Keep checking against the `startDate` flag
2) Move the logic required for showing the banner completely to the UI - i.e. 
   - cloud
   - user limit is not `-1` (unlimited)
   - plan.model == per user

I think my preference is #2 as it's really a frontend concern, we don't use the `startDate` on the backend

packages/builder/src/pages/builder/portal/apps/index.svelte
```diff
@@ -294,6 +314,10 @@
 </Modal>
 
 <AppLimitModal bind:this={appLimitModal} />
+<AccountLockedModal bind:this={accountLockedModal} onConfirm={() =>
+    isOwner
+      ? $licensing.goToUpgradePage()
+      : window.open("https://budibase.com/pricing/", "_blank")} />
```
Rory-Powell, line 320: Could update this to be `$licensing.goToPricingPage()` so we can abstract this away, there are now a number of places we do this

jvcalderon: Done.

packages/builder/src/stores/portal/licensing.js
jvcalderon: From my point of view the usage of `startDate` as a flag to decide if a user is in a whitelist is confusing, because it is not an explicit behaviour; you just need to know it. Is very difficult to infer this by reading the code. On the other hand I wouldn't change this now; `startDate` is being used to calculate the remaining days for the user so we already have that value in the file; we can just use it for the moment. Maybe in a separated issue, we can make a little refactor to treat:

1. Make explicit that check (instead directly use startDate).
2. Move the responsibility to the front.

I can create a backlog issue and just move forward with this one. What do you think @Rory-Powell ?

packages/builder/src/stores/portal/licensing.js
jvcalderon: In your opinion, should I keep the 'will'? I mean, in the end the name of the function shouldn't reflect the external usage in use cases but the internal behaviour of the function. Just to confirm, what is the returning result of the function by itself? Forgetting the outside context, when that function returns `true` it means ('the limit is exceeded now' or 'the limit is not exceeded now but will be exceeded in the future')?

packages/builder/tsconfig.json
```diff
@@ -6,7 +6,6 @@
     "sourceMap": true,
     "baseUrl": ".",
     "paths": {
-      "@budibase/types": ["../types/src"],
```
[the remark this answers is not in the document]
Rory-Powell: I believe the paths here allowed us to references sources directly, rather than relying in a `yarn build` before the differences would be picked up locally. When we add esbuild integration (currently in progress) we may be able to remove this. @adrinr might be able to give a better explanation. 

With that said I'm not sure why we have the `backend-core` dependencies here, they can probably be removed.

packages/builder/src/stores/portal/licensing.js
```diff
@@ -205,9 +204,10 @@ export const createLicensingStore = () => {
         const userCount = usage.usageQuota.users
         const userLimitReached = usersLimitReached(userCount, userLimit)
         const userLimitExceeded = usersLimitExceeded(userCount, userLimit)
-        const days = dayjs(userQuota?.startDate).diff(dayjs(), "day")
-        const userLimitDays = days > 1 ? `${days} days` : "1 day"
-        const errUserLimit = userQuota?.startDate && userLimitExceeded
+        const errUserLimit = license.account.hosting === Hosting.CLOUD &&
```
Rory-Powell, line 207: `account` doesn't exist on the license, I think we will need to check the hosting that comes back from the `/environment` request here as an alternative. We should be able to import the `admin` store here which exports the `cloud` property like:

```ts
import { admin } from "stores/portal"
import { get } from "svelte/store"

async function isCloud() {
   let adminStore = get(admin)
   if (!adminStore.loaded) {
       await admin.init()
      adminStore = get(admin)
   }
  
   return adminStore.cloud

}

```

packages/builder/package.json
```diff
@@ -62,6 +62,7 @@
     "@budibase/frontend-core": "0.0.1",
     "@budibase/shared-core": "0.0.1",
     "@budibase/string-templates": "0.0.1",
+    "@budibase/types": "0.0.1",
```
Rory-Powell, line 65: I think this is a great thing for us to start doing. A lot of our constants are duplicated - there's no real reason not to bring in the types package. Especially if we're ever gonna implement TS in the parts of the builder that integrate with APIs etc. 

I believe we already pull it in transiently anyway via the `shared-core` dependency. Better to call it out explicitly like you have here.

packages/builder/src/stores/portal/licensing.js
```diff
@@ -198,11 +202,12 @@ export const createLicensingStore = () => {
...
-        const userLimitReached = willReachUserLimit(userCount, userLimit)
-        const userLimitExceeded = willExceedUserLimit(userCount, userLimit)
-        const days = dayjs(userQuota?.startDate).diff(dayjs(), "day")
-        const userLimitDays = days > 1 ? `${days} days` : "1 day"
-        const warnUserLimit = userQuota?.startDate && userLimitExceeded
+        const userLimitReached = usersLimitReached(userCount, userLimit)
+        const userLimitExceeded = usersLimitExceeded(userCount, userLimit)
+        const errUserLimit = license.account.hosting === Hosting.CLOUD &&
+          license.plan.model === PlanModel.PER_USER &&
+          userQuota.usage.static.users.value !== UNLIMITED &&
```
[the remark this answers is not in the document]
jvcalderon: Done.

Rory-Powell (approved):
LGTM 🚀

PR #10558, opened 2023-05-12
Add status check to license api

Pedro-Revez-Silva:
## Description
Add a small check to stop tests from proceeding if the license is wrong.

### DIFF
--- a/qa-core/src/account-api/api/apis/LicenseAPI.ts
+++ b/qa-core/src/account-api/api/apis/LicenseAPI.ts
@@ -22,7 +22,9 @@ export default class LicenseAPI {
     )
 
     if (response.status !== 200) {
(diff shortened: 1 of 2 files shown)


qa-core/src/account-api/api/apis/LicenseAPI.ts
```diff
@@ -20,6 +20,10 @@ export default class LicenseAPI {
         internal: true,
       }
     )
+
+    if (response.status !== 200) {
+      throw new Error(`License Error: ${response.status}`)
```
Rory-Powell, line 25: just a small suggestion on more specific logging 
```suggestion
      throw new Error(`Could not update license for accountId=${accountId}: ${response.status}`)
```

Rory-Powell (approved):

PR #10568, opened 2023-05-12
Synchronous automation for App actions and webhook

PClmnt:
## Description
This PR adds the ability for users to trigger automations from app actions, synchronously.  We are doing this by adding a new block called `Collect data` in the automations tab, when added to an automation (it can only be added once and as the final step) it will "collect" the given binding or value and ensure the automation runs synchronously. Now, whenever selecting this automation from the Button Actions drawer in the Design section, you will have access to a binding presenting the value in the following steps which you can do with as you want. 

It also provides the same functionality for webhooks, in that you can make a call to a webhook automation, and have it return the Collect block's value - this will be a premium feature. 

Button actions demo: 

https://github.com/Budibase/budibase/assets/5665926/0ee4e0b8-2ab3-4deb-b32b-ef6f5bdfec06

Webhook Demo:
https://github.com/Budibase/budibase/assets/5665926/57226ce5-1778-43be-af32-5be0df4a8372


packages/server/src/automations/utils.ts
```diff
@@ -14,13 +14,16 @@ import sdk from "../sdk"
...
 const CRON_STEP_ID = definitions.CRON.stepId
-const Runner = new Thread(ThreadType.AUTOMATION)
 
 const jobMessage = (job: any, message: string) => {
   return `app=${job.data.event.appId} automation=${job.data.automation._id} jobId=${job.id} trigger=${job.data.automation.definition.trigger.event} : ${message}`
 }
 
-export async function processEvent(job: any) {
+export async function processEvent(job: any, timeout?: { timeout?: number }) {
+  const Runner = new Thread(ThreadType.AUTOMATION, {
```
PClmnt, line 23: @mike12345567 I've moved the thread initialisation inside this function as it now needs access to the timeout which is coming through, does this make sense?

mike12345567: hmm I think this could be pretty dangerous - this technically means its creating a thread per event - we are trying to avoid that with total initialisation.

I think a better approach would be to avoid the whole thread mechanism, and then just use a promise timeout (`Promise.race`) - so essentially load the `execute` function directly for sync runs instead - if you need more info/thoughts are this give me a shout and we can have a call!

packages/frontend-core/src/constants.js
```diff
@@ -70,6 +70,7 @@ export const Features = {
   ENFORCEABLE_SSO: "enforceableSSO",
   BRANDING: "branding",
   SCIM: "scim",
+  SYNC_WEBHOOKS: "syncWebhooks",
```
Rory-Powell, line 73: Given that the only type of webhook that we have is an automation webhook, I'd be in favour of scoping this more narrowly to be `syncAutomations`. I'm also really against the 'webhook' terminology in general around automations, as it's not a webhook at all. Where a webhook is a request sent by budibase to an external system. What we have is a http trigger for an automation. 

If we do introduce the concept of real webhooks in future, e.g. sent upon system events, I wouldn't want this name to be confused there. 


*sorry for saying webhook so many times

PClmnt: No that makes a lot of sense, happy to change it - just terminology that kind of stuck in my head. Not wedded to the idea!

packages/server/src/threads/automation.ts
```diff
@@ -495,6 +495,32 @@ export function execute(job: Job, callback: WorkerCallback) {
...
+    setTimeout(() => {
+      reject(new Error("Timeout exceeded"))
+    }, job.data.event.timeout || 12000)
+  })
+
+  return context.doInAppContext(appId, async () => {
+    const envVars = await sdkUtils.getEnvironmentVariables()
+    // put into automation thread for whole context
+    return context.doInEnvironmentContext(envVars, async () => {
+      const automationOrchestrator = new Orchestrator(job)
```
PClmnt, line 514: @mike12345567 I've split the execute synchronously out into a new function and used promise.race here - I know it was slightly different than the flag within execute that we discussed but i thought this kind of ensured we separated sync code from async code. Maybe it makes more sense to it discussed way? I did test and this works well and throws back the `Timeout exceeded` error to the client correctly etc

packages/server/src/api/controllers/automation.ts
```diff
@@ -257,13 +264,34 @@ export async function getDefinitionList(ctx: BBContext) {
...
+      },
+      { getResponses: true }
+    )
+
+    let collectedValue = response.steps.find(
+      step => step.stepId === AutomationActionStepId.COLLECT
+    )
+    ctx.body = collectedValue?.outputs
+  } else {
+    if (ctx.appId && !isProdAppID(ctx.appId)) {
```
PClmnt: @mike12345567 Instead of adding a second app endpoint, is this a bit cleaner? Remove the middleware from the route, but in the controller for async automations check if it's a Production ID or not.

packages/server/src/api/controllers/automation.ts
```diff
@@ -257,13 +264,34 @@ export async function getDefinitionList(ctx: BBContext) {
...
   let automation = await db.get(ctx.params.id)
-  await triggers.externalTrigger(automation, {
-    ...ctx.request.body,
-    appId: ctx.appId,
-  })
-  ctx.body = {
-    message: `Automation ${automation._id} has been triggered.`,
-    automation,
+
+  let hasCollectStep = sdk.automations.utils.checkForCollectStep(automation)
```
mike12345567, line 268: It would be good to have a test for this - just to make sure the new logic is maintained and doesn't regress in future.

packages/server/src/automations/steps/collect.ts
```diff
@@ -0,0 +1,53 @@
...
+  },
+  schema: {
+    inputs: {
+      properties: {
+        collection: {
+          type: AutomationIOType.STRING,
+          title: "What to Collect",
+        },
+      },
+      required: ["text"],
```
mike12345567, line 30: There is no `text` input - so I don't believe this is required.

packages/server/src/automations/steps/collect.ts
```diff
@@ -0,0 +1,53 @@
...
+          description: "Collected data",
+        },
+      },
+      required: ["success", "value"],
+    },
+  },
+}
+
+export async function run({ inputs }: AutomationStepInput) {
+  return {
```
mike12345567, line 49: should we have a `success: false` scenario if the collection could not be found in some manner?

packages/server/src/automations/steps/delay.ts
```diff
@@ -14,6 +14,7 @@ export const definition: AutomationStepSchema = {
   description: "Delay the automation until an amount of time has passed",
   stepId: AutomationActionStepId.DELAY,
   internal: true,
+  canLoop: false,
```
mike12345567, line 17: Just thinking about this - I would argue this is a bit of a feature flag, it might be useful to do something like we've done with the datasources and actually have a `features` array that we add to automations, e.g.

`features: [AutomationFeature.LOOPING]` - this way if the feature is not present we can assume it doesn't support looping. This is also quite a bit more re-usable if we ever want to be able to turn off features for other blocks as well.

packages/server/src/threads/automation.ts
mike12345567: Yes - this is probably acceptable, although flagging to the orchestrator in someway, like through a `cancel` function, just to make sure the orchestrator is aware that it doesn't need to continue.

The concern with this approach is that the sync operation will simply continue, and the response will be ignored, which uses up CPU time. Its not very major, but making the orchestrator aware that the response is no longer needed means we could exit early.

packages/types/src/documents/app/automation.ts
```diff
@@ -122,6 +123,7 @@ export interface AutomationStepSchema {
     outputs: InputOutputBlock
   }
   custom?: boolean
+  canLoop?: boolean
```
mike12345567, line 126: As with other comment, this could be adjusted to `features: AutomationFeature[]` and then create the automation feature enum.

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/TriggerAutomation.svelte
```diff
@@ -1,12 +1,15 @@
 <script>
-  import { Select, Label, Input, Checkbox } from "@budibase/bbui"
+  import { Select, Label, Input, Checkbox, Icon } from "@budibase/bbui"
   import { automationStore } from "builderStore"
   import SaveFields from "./SaveFields.svelte"
   import { TriggerStepID } from "constants/backend/automations"
+  import { AutomationActionStepId } from "../../../../../../../../types/src/documents"
```
Rory-Powell, line 6: Looks like a mistaken import

mike12345567: image

packages/server/src/automations/utils.ts
PClmnt: Added a flag inside the orchestrator so we exit successfully.

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/TriggerAutomation.svelte
PClmnt: Haha, sorted.

mike12345567 (approved):
LGTM! ready to go :)

mike12345567 (approved):
LGTM! ready to go :)

PR #10586, opened 2023-05-15
Disable revert button when not applicable, fix tooltip clipping

deanhannigan:
Addresses: 
- Application revert button was active when the application hadn't been published. Attempting to use the button without a published app resulted in a 400 and an error state. `App has not yet been deployed` is visible in the server errors.
- Also added a z-index fix for `ActionButton` tooltips. The tooltips in the main menu in the builder were being clipped by the new grid layout.

## Screenshots

UI after initiating the revert behaviour without a published application.

Tooltips clipped by the new grid layout
Screenshot 2023-05-15 at 10 43 33

Tooltips with their z-index bumped to 100
Screenshot 2023-05-15 at 10 43 54

### DIFF
--- a/packages/bbui/src/ActionButton/ActionButton.svelte
+++ b/packages/bbui/src/ActionButton/ActionButton.svelte
@@ -128,5 +128,6 @@
     max-width: 150px;
     transform: translateX(-50%);
     text-align: center;
+    z-index: 100;
   }
 </style>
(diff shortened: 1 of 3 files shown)


aptkingston (changes requested):
Would prefer to change how the z index issue is handled, but the revert stuff LGTM!

PR #10587, opened 2023-05-15
Removing version error breaking CLI

mike12345567:
## Description
Fix for CLI startup complaining about version not being found.

Current CLI version when installed via CLI shows the following message:
```
Error - Failed to run CLI command - please report with the following message:
Error - Error: Cannot find a valid version in its package.json
```

Removed this error from backend-core as it makes the library much harder to import/use.


melohagan (approved):
LGTM - thanks!

Issue #6049, opened 2022-05-24
[BUDI-6993] Query Rows does not work with bindings when using a MySQL datasource


FlaminWrap:
**Hosting**
- Self
  - Method: BudiCLI
  - Budibase Version: 1.0.178 
  - App Version: 1.0.178 

**Describe the bug**
Query Rows does not work with bindings when using a MySQL datasource. Instead it returns all rows until the row limit is hit.

**To Reproduce**
1. Create a new app with a MySQL data source
2. Create two tables Record and RecordType
3. Create two columns in RecordType: Type/Text , Value/Number
4. Create at least two rows in the with both columns filled with different values in the RecordType table
5. Create the following relationsip between the tables: One RecordType row > many Record rows
6. Create a column in the Record table called Value and make it a number type
7. Create an automation to launch when creating a row
8. Create a Query Rows step in the automation to query the RecordType table and filter it by the foreign key from the trigger row
9. Create an Update Rows step in the automation to update the Record Table
10. Set the RecordType field to `return $("trigger.row.RecordType");`
11. Set the Fk_RecordType_Record field to `return $("trigger.row.fk_RecordType_Record");`
12. Set the Value field to `return $("steps.1.rows.0.Value");`
13. Set the RowID field to `return $("trigger.id");`
14. Press Run test at the top of the automation page and select anything but the first record in the RecordType table created earlier
15. The first value will have been pulled through instead of the correct value.

Query Row Input:

`
{
  "tableId": "datasource_plus_fc2c66277ba54f2caac0dc36d5efc531__RecordType",
  "filters": {
    "string": {},
    "fuzzy": {},
    "range": {},
    "equal": {
      "id": null
    },
    "notEqual": {},
    "empty": {},
    "notEmpty": {},
    "contains": {},
    "notContains": {}
  },
  "filters-def": [
    {
      "id": "Aa5fR_PiF",
      "field": "id",
      "operator": "equal",
      "value": "{{ js \"cmV0dXJuICQoInRyaWdnZXIucm93LmZrX1JlY29yZFR5cGVfUmVjb3JkIik7\" }}",
      "valueType": "Binding",
      "type": "number"
    }
  ],
  "sortColumn": "id",
  "sortOrder": "ascending",
  "limit": "50"
}
`

Query Row Output:

`
{
  "rows": [
    {
      "id": 1,
      "Type": "Type1",
      "Value": 1,
      "_id": "%5B1%5D",
[...]

PR #10591, opened 2023-05-15
Fix automation filtering on number fields

aptkingston:
## Description
This PR fixes an issue where the "query rows" automation step does not work when filtering on number fields using a binding. This was broken because it did not account for the numerical key prefixing we use in our filter expression structures.

Addresses: 
- #6049 

I went to update the tests to account for this, but then realised the tests are actually pointless because the actual search operation is completely mocked, so an additional test for this edge case would provide no value.


melohagan (approved):
LGTM! Nice neat fix.

PR #10593, opened 2023-05-15
Fix for autoid issue on create row

deanhannigan:
## Description

Fix for the autoId increment behaviour on internal tables. The fix for https://github.com/Budibase/budibase/pull/10482 fixed the issue for importing new rows, but broke the existing behaviour for create.

Addresses: 
- https://github.com/Budibase/budibase/discussions/10585
- #10599

### DIFF
--- a/packages/server/src/api/controllers/row/internal.ts
+++ b/packages/server/src/api/controllers/row/internal.ts
@@ -118,8 +118,11 @@ export async function patch(ctx: UserCtx) {
     combinedRow[key] = inputs[key]
   }
 
+  // need to copy the table so it can be differenced on way out
+  const tableClone = cloneDeep(dbTable)
+
   // this returns the table and row incase they have been updated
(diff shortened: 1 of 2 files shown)


aptkingston (approved):
LGTM! We may discover at some point that there are other usages of `inputProcessing` elsewhere in the codebase that also depend on the clone, but at least we know how to quickly fix them if we find them.

mike12345567 (approved):
Looks good to me - theres only really two places this is used:
```
packages/server/src/api/controllers/table/utils.ts:122:    const processed = inputProcessing(user, table, row, {
packages/server/src/api/controllers/row/internal.ts:122:  let { table, row } = inputProcessing(ctx.user, dbTable, combinedRow)
packages/server/src/api/controllers/row/internal.ts:166:  let { table, row } = inputProcessing(ctx.user, dbTable, inputs)
```
These are all covered now so should be good - but might need to refactor this a bit in future.

PR #10601, opened 2023-05-16
Verify google sheets connection

adrinr:
## Description
Implement google sheets datasource verification


mike12345567 (approved):
Looks good - forgot we would need to retrieve the config for verification of GSheets!

PR #10602, opened 2023-05-16
Further CLI build fixes (NPM)

mike12345567:
## Description
Fix for 'TypeError: Cannot read properties of undefined (reading 'command')' error that was occurring - program wouldn't load successfully with an undefined version number.


melohagan (approved):
LGTM - thanks again!

PR #10606, opened 2023-05-16
Verify airtable connection

adrinr:
## Description
Verify airtable connections

### Valid config
image

### Invalid key
image


mike12345567 (approved):
Very good idea for detecting connection - wouldn't have thought of it!

PR #10620, opened 2023-05-17
Oracle connection checking

mike12345567:
## Description
Checking connections for Oracle databases.

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -131,10 +131,8 @@ export async function verify(
 ) {
   const { datasource } = ctx.request.body
   let existingDatasource: undefined | Datasource
-  try {
-    existingDatasource = await sdk.datasources.get(datasource._id!)
-  } catch (err) {
-    // doesn't exist - can ignore, first creation
+  if (datasource._id) {
+    existingDatasource = await sdk.datasources.get(datasource._id)
   }
   let enrichedDatasource = datasource
   if (existingDatasource) {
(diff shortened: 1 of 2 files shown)


packages/server/src/api/controllers/datasource.ts
```diff
@@ -127,13 +130,19 @@ export async function verify(
...
   const { datasource } = ctx.request.body
-  const existingDatasource = await sdk.datasources.get(datasource._id!)
-
-  const enrichedDatasource = sdk.datasources.mergeConfigs(
-    datasource,
-    existingDatasource
-  )
-
+  let existingDatasource: undefined | Datasource
+  try {
```
adrinr, line 134: Do we need the try/catch? Can we not check if the `datasource._id` exists instead?

packages/server/src/integrations/oracle.ts
```diff
@@ -322,6 +323,30 @@ class OracleIntegration extends Sql implements DatasourcePlus {
...
+      connection = await this.getConnection()
+      response.connected = true
+    } catch (err: any) {
+      response.connected = false
+      response.error = err.message
+    } finally {
+      if (connection) {
+        try {
+          await connection.close()
+        } catch (err: any) {
```
adrinr, line 341: Why is this? Should the connected:`true/false` not be already setup in the catch above?

mike12345567: It seems some Oracle issues don't crop up until you actually try to close the connection - doing this means we safely catch all errors that could occur (start and end of connection).

adrinr (approved):

PR #10622, opened 2023-05-17
Chore - Improve pipeline run time

adrinr:
## Description
1. Add caching to yarn installs

| Before | After |
|--------|--------|
| image | image |

2. Remove redundant bootstraps
Got rid of unnecessary steps: 
image

3. Remove unnecessary build steps on testing

| Before | After |
|--------|--------|
| image |  image |

### DIFF
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -59,6 +59,7 @@ jobs:
           node-version: 14.x
           cache: "yarn"
       - run: yarn
+      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
       - run: yarn test --ignore=@budibase/pro
         env:
           JEST_MAX_WORKERS: ${{ steps.cpu-cores.outputs.count }}
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -61,9 +61,6 @@ jobs:
       - run: yarn
       - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
       - run: yarn test --ignore=@budibase/pro
-        env:
-          JEST_MAX_WORKERS: ${{ steps.cpu-cores.outputs.count }}
-
       - uses: codecov/codecov-action@v3
         with:
           token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -59,7 +59,7 @@ jobs:
           node-version: 14.x
           cache: "yarn"
       - run: yarn
-      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates --scope=@budibase/client
+      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
       - run: yarn test --ignore=@budibase/pro
       - uses: codecov/codecov-action@v3
         with:
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -59,7 +59,7 @@ jobs:
           node-version: 14.x
           cache: "yarn"
       - run: yarn
-      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
+      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates --scope=@budibase/client
       - run: yarn test --ignore=@budibase/pro
       - uses: codecov/codecov-action@v3
         with:
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -97,7 +97,8 @@ jobs:
           cache: "yarn"
       - run: yarn
       - run: yarn build
-      - run: |
+      - name: Run tests
+        run: |
           cd qa-core
           yarn setup
           yarn test:ci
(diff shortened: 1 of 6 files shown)


mike12345567 (approved):

.github/workflows/budibase_ci.yml
```diff
@@ -22,44 +22,51 @@ jobs:
...
       - run: yarn
-      - run: yarn bootstrap
-      - run: yarn build
+      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
+      - name: Get number of CPU cores
+        id: cpu-cores
+        uses: SimenB/github-actions-cpu-cores@3e824c4e1e0f0880c1e1023cbd7a418f5929d358
       - run: yarn test --ignore=@budibase/pro
+        env:
+          JEST_MAX_WORKERS: ${{ steps.cpu-cores.outputs.count }}
```
Rory-Powell, line 68: The default behaviour for jest should be to use the cpu core available as max workers. We should omit the `max-workers` arg completely to enable this. 

We also have `scripts/test.sh` scripts in worker and server that always runs in band when in CI. this brought down time in CI significantly. This might conflict with the above. 

From past experiments, the server and worker tests perform very badly when using max available cores, given that so many of our tests start the whole service, it causes resource utilisation problems in CI.

.github/workflows/budibase_ci.yml
```diff
@@ -22,44 +22,51 @@ jobs:
...
       - name: Use Node.js 14.x
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: 14.x
+          cache: "yarn"
       - run: yarn
-      - run: yarn bootstrap
-      - run: yarn build
+      - run: yarn build --scope=@budibase/types --scope=@budibase/shared-core --scope=@budibase/string-templates
```
Rory-Powell, line 62: We didn't used to have to build to run the tests, where we should be able to use source mappings instead. Is there a way we can revert to this behaviour? looks like it was added here

https://github.com/Budibase/budibase/commit/80002873aaed2f4ea7933efe2487bf4f9196b96d
https://github.com/Budibase/budibase/commit/96f2e830728c12ee7ad5ac2ff17c96fe69d83e3c

.github/workflows/budibase_ci.yml
Rory-Powell: See you have updated the test scripts 👍 Other points remain

adrinr: That was an attempt to see any improvement, it did not work

.github/workflows/budibase_ci.yml
adrinr: This is required to build the shared-core for the js projects (builder and others) to run. I don't think we can use jest paths for this with js to ts

packages/builder/tsconfig.json
```diff
@@ -4,12 +4,7 @@
     "composite": true,
     "declaration": true,
     "sourceMap": true,
-    "baseUrl": ".",
-    "paths": {
-      "@budibase/types": ["../types/src"],
-      "@budibase/backend-core": ["../backend-core/src"],
-      "@budibase/backend-core/*": ["../backend-core/*.js"]
-    }
+    "baseUrl": "."
```
adrinr, line 7: This is not used, removed to avoid confusion

packages/worker/src/tests/structures/groups.ts
```diff
@@ -18,7 +18,7 @@ export const UserGroup = () => {
     apps: [],
     color: generator.color(),
     icon: generator.word(),
-    name: generator.word({ length: 2 }),
+    name: generator.word(),
```
adrinr, line 21: This was causing flakiness, hopefully it will remove all the need rerun issues

.github/workflows/budibase_ci.yml
Rory-Powell: I see! One option could be to run the ui tests with ts-jest using the swc or esbuild plugins, would be very fast then. 
Maybe when we move to esbuild everywhere it's not much of a problem though

packages/worker/src/tests/structures/groups.ts
Rory-Powell: nice!

.github/workflows/budibase_ci.yml
```diff
@@ -22,43 +22,44 @@ jobs:
...
+      - uses: actions/checkout@v3
         with:
           submodules: true
           token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
       - name: Use Node.js 14.x
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: 14.x
+          cache: "yarn"
```
Rory-Powell, line 45: Could you elaborate in the description what this will achieve for us? around what exactly is being cached / when does the cache invalidate

e.g. 
- When the pr first runs
- When the pr re-runs
- When the package.lock changes in subsequent commits 
etc.

Rory-Powell (approved):
Looks great nice work! Such an improvement. Just a question on the caching behaviour

.github/workflows/budibase_ci.yml
adrinr: This does not cache any node_modules or similar, it caches the yarn cache. Every time that `yarn install` runs, the machine will download the packages and keep them locally (see `yarn cache dir`). Then these packages are then copied over to your node_modules, making subsequent `yarn install` much faster. This was the main selling point when `yarn` overtook `npm`, as it was much faster. You can test this by running `yarn cache clean` and rerunning yarn with a fresh cache. By default, GitHub action will run every time with a fresh cache. This is because it runs in dockerised instances.
Adding the cache key, these folders will be persisted between runs, pulling more from the local cache and less from the repos. https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

The performance increase can be seen in the description message. You can see that the setup takes longer (as it needs to load the cache, but the yarn install is much quicker: 
image

Rory-Powell: Thanks! Sounds like a great usage of the cache. So every run will start to benefit from this right away

PR #10623, opened 2023-05-17
Change the nightly tests to be more explicit

adrinr:
## Description
Currently, we have smoke tests for QA that exclude specific test files, that then run at night on chron bases. This change make the tests to run as "only nightly" more explicit, allowing us to create more of these tests across the directories


mike12345567 (approved):
👌

mike12345567 (approved):
👌

PR #10626, opened 2023-05-17
Automation logging updates

Rory-Powell:
## Description
Updates to logging for better pino integration to aid debugging via datadog. 

Additionally add some better type definitions around bull jobs specific to automations.


packages/backend-core/src/logging/pino/logger.ts
```diff
@@ -159,6 +161,16 @@ if (!env.DISABLE_PINO_LOGGER) {
     return appId
   }
 
+  const getAutomationId = () => {
+    let appId
+    try {
+      appId = context.getAutomationId()
+    } catch (e) {
+      // do nothing
```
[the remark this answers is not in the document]
Rory-Powell: The reason is that this may be invoked anywhere a console statement is done. We may or may not be in automation context at this point, or in any context at all. The safest approach is to try-catch so that we don't break logging.

packages/backend-core/src/queue/listeners.ts
```diff
@@ -31,77 +32,135 @@ function handleStalled(queue: Queue, removeStalledCb?: StalledFn) {
...
+      await doInJobContext(job, () => {
+        console.error(...getLogParams(eventType, BullEvent.STALLED, { job }))
+      })
+    })
+    .on(BullEvent.ERROR, (error: any) => {
+      // An error occurred.
+      console.error(...getLogParams(eventType, BullEvent.ERROR, { error }))
+    })
+
   if (process.env.NODE_DEBUG?.includes("bull")) {
```
Rory-Powell, line 114: I'm planning on turning off `bull` debug logging by default to bring down logging costs, hence the `stalled` and `error` logs above being promoted to logging by default

mike12345567 (approved):
LGTM! Would be good after getting this merged to pull it back to development as some of the types changes will have implications for sync automations - theres a few changes around `Job` types already!

packages/backend-core/src/context/mainContext.ts
```diff
@@ -104,6 +104,22 @@ async function newContext(updates: ContextMap, task: any) {
   return Context.run(context, task)
 }
 
+export async function doInAutomationContext(
+  appId: string,
```
adrinr, line 108: Can we use an object as a parameter to make it more verbose in its usage?
```
export async function doInAutomationContext(params: {
  appId: string,
  automationId: string,
  task: any
})
```

adrinr (commented):
Small comment, but LGTM

PClmnt:
> LGTM! Would be good after getting this merged to pull it back to development as some of the types changes will have implications for sync automations - theres a few changes around `Job` types already!

+1 to this, some of the stuff you've done around the triggers and automation thread code overlap quite heavily with some work I was doing for sync automations. LGTM though!

Issue #10594, opened 2023-05-15
[BUDI-7012] Formula Columns Ignore Order in Table on Firefox


shawnweeks:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.6.8
  - App Version: 2.6.8

**Describe the bug**
When adding a formula based column to a table component the formula column ignores column order and gets placed in the far left side before any buttons. Everything renders in the correct order in Chrome.

**To Reproduce**
Attached application export

**Expected behavior**
Formula columns should behave like any other columns in a table component and respect the order and should not be placed before the buttons.

**Screenshots**
<img width="1267" alt="Screenshot 2023-05-15 at 1 25 41 PM" src="https://github.com/Budibase/budibase/assets/9768773/f29790ac-c424-4f6d-af08-e84e0dc89ad8">

  
**App Export**
[Volunteer-export-1684175287280.tar.gz](https://github.com/Budibase/budibase/files/11480908/Volunteer-export-1684175287280.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Firefox
 - Version 112.0.2

**Additional context**

<sub>[BUDI-7012](https://linear.app/budibase/issue/BUDI-7012/formula-columns-ignore-order-in-table-on-firefox)</sub>

shawnweeks:
Here is the same place in Chrome.

<img width="1291" alt="Screenshot 2023-05-15 at 1 46 45 PM" src="https://github.com/Budibase/budibase/assets/9768773/5ada78ab-9233-4647-96f8-5ac7fa934a5d">

shawnweeks:
Just tested with 2.6.11 as well, same behavior.

shawnweeks:
@melohagan there are still some issues with column ordering, for some reason in this example the column order is getting reversed.

<img width="1294" alt="Screenshot 2023-06-12 at 12 33 34 PM" src="https://github.com/Budibase/budibase/assets/9768773/72fa5671-fbc4-494c-949c-078aea9e005d">

[Column Order Bug-export-1686591266376.tar.gz](https://github.com/Budibase/budibase/files/11726284/Column.Order.Bug-export-1686591266376.tar.gz)

melohagan:
Hey @shawnweeks Thanks for highlighting that. There's an open issue for it here: https://github.com/Budibase/budibase/issues/10689

PR #10627, opened 2023-05-17
Make sure divider fields are left-most

melohagan:
## Description
The column order defined in `Configure columns` was working in Google Chrome, but only because the sort appeared to not be functioning correctly.

The sort order from `Configure columns` was not applied in Firefox because the table component was overriding that order with this field sort.
The field sort has been removed to ensure the user defined order is maintained, however a check is in place to make sure custom fields (components added to a table) appear left-most.

Addresses: 
- https://github.com/Budibase/budibase/issues/10594

## App Export
- See original issue

## Screenshots
Screenshot 2023-05-17 at 14 04 42

Screenshot 2023-05-17 at 14 05 16


PClmnt (approved):

aptkingston (commented):
Removing the ability to honour the order property at all seems a bit dangerous - I'm sure it was added for a reason. Putting the divider leftmost is a good change, but I think I'd prefer to simply update this sorting logic so that it considers `divider` to be the most significant sorting factor, then falling back to the order property.

Issue #10609, opened 2023-05-16
[BUDI-7001] MySQL TIME type column breaks time-only date picker in forms and grid, crashes builder


andz-bb:
**Checklist**

* [X] I have searched budibase discussions and GitHub issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.6.11
  * App Version: 2.6.11

**Describe the bug**
When trying to load a screen with a time-only date picker present, the app preview freezes and the following error appears in the console:

```
Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'tabIndex')
    at buildMonthSwitch (index.js?v=19f05e52:580:47)
    at jumpToDate (index.js?v=19f05e52:341:13)
    at Object.setDate (index.js?v=19f05e52:1684:9)
    at $$self.$$.update (Flatpickr.svelte:29:7)
    at update (index.mjs:1294:12)
    at flush (index.mjs:1258:17)
```

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/5bdabaae-afad-48c0-9166-1aa2127139b1/7adecdec-1e7c-4b15-b675-0c6ee2ce5136)

The same issue occurs with the grid view in the data section when trying to work with these MySQL Time type columns.

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/262fa1de-3df4-4a9a-bddb-2c812f4e637b/3763e806-6427-433b-a468-57e281fede49)

This is a significant issue as in both cases the app preview or builder freezes following the error.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a MySQL table with a TIME type column
2. Add the MySQL DB as a datasource with the new table
3. Go to the grid view and try to add a record with a value for time
4. The builder will freeze when trying to interact with the time picker. Invalid date is shown.
5. (the time picker control has the same issue in design view)
6. Create a form with a Date picker field linked to the time column
7. Make sure the Date picker is set to Time only
8. Try to load the component or interact with it - the screen freezes with the same error

**Expected behaviour**
Date/time picker component should load for this data type and be usable.

**Additional Context**

[Raised by community member here](https://github.com/Budibase/budibase/discussions/10566)

[...]

PR #10631, opened 2023-05-17
hotfix for time-only date picker problems (reverting flatpickr to previous version)

andz-bb:
@aptkingston already fixed this yesterday but only in develop (#10612). This fix is to add the critical part to master for a quick fix to prevent screens affected by the bug from completely breaking. The rest of the cheeks grid time picker changes are in develop and will be included in the next iteration.

Addresses: 
#10609 

## Documentation


aptkingston (approved):
LGTM!

PR #10633, opened 2023-05-17
Production app user updates

mike12345567:
## Description
Fixing an issue with the tenant ID not being found in self host (default) for identification purposes, stopping doc update queue from activating and working as expected.

Also - making the fallback sync cover both the production app (if exists) and the development app - this means that the cleanup should occur for any users that aren't supposed to be there (eventually this will be removed, but for now need this for cleaning).


shogunpurple (approved):
LGTM! Nice fix

PR #10635, opened 2023-05-18
Automation panel does not scroll

### LINKED ISSUE #10535: Automation panel does not scroll
**Checklist**

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.6.8
  - App Version: 2.6.8

**Describe the bug**
You cannot scroll in the automation side panel

**To Reproduce**
1. Create an app
2. Create at least 5 automatons
3. Resize the screen so some of the automatons aren't visible
4. You cannot scroll down to see them

**Expected behavior**
I am able to scroll

**Screenshots**
panel

 - OS: Windows 10
 - Browser Edge
 - Version 113.0.1774.35

gitstart:
Fixes https://github.com/Budibase/budibase/issues/10535
#### Demo

https://github.com/GitStartHQ/client-budibase/assets/68722497/1f4f9a7b-c0b6-480c-af3f-bfedb4179f1c

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### DIFF
--- a/packages/builder/src/components/design/Panel.svelte
+++ b/packages/builder/src/components/design/Panel.svelte
@@ -58,6 +58,7 @@
     justify-content: flex-start;
     align-items: stretch;
     transition: width 130ms ease-out;
+    overflow: hidden;
   }
   .panel.borderLeft {
     border-left: var(--border-light);
--- a/packages/builder/src/pages/builder/app/[application]/automate/_layout.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/automate/_layout.svelte
@@ -32,9 +32,7 @@
 
 <!-- routify:options index=3 -->
 <div class="root">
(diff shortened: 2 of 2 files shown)


aptkingston:
Hi! Thanks for the contribution. Your change does fix the issue, but there's actually a simpler change which fixes a few similar scroll issues as well.

Inside `packages/builder/src/components/design/Panel.svelte` we just need to add `overflow: hidden` to the `.panel` selector. Our `Panel` component is intended to scroll (and sets `overflow: auto` on the `.body` selector), but without hiding overflow in the parent selector it doesn't work as intended, so that's the real bug here!

If you'd like to update this PR to just include that change instead then I'm still happy to merge your PR rather than making that change internally.

gitstart:

PR #10659, opened 2023-05-22
Datasource+ table fetching API

mike12345567:
## Description
This introduces a new API endpoint for datasource+ types, which allows retrieving the list of tables which are available within the datasource.

This does not attempt to retrieve any information about the schema of each of the tables, as this is strictly for determining which tables are of interest - we should not retrieve any information beyond what is required to confirm with the user the tables they need.

I've also introduced a new feature flag which can be used for determining whether or not the datasource supports the table fetching feature - this can be assumed for all datasource+ databases, but this could change in future (some DS+ might not be able to do this).

I've also added some more test cases for this, using the new API endpoint `GET /api/datasources/:datasourceId/tables` as well as some basic test cases for the connection verification API `/api/datasources/verify` - these will run as part of CI against Postgres.

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -21,7 +21,7 @@ import {
   CreateDatasourceRequest,
   VerifyDatasourceRequest,
   VerifyDatasourceResponse,
-  FetchTablesDatasourceResponse,
+  FetchDatasourceInfoResponse,
   IntegrationBase,
   DatasourcePlus,
 } from "@budibase/types"
@@ -154,8 +154,8 @@ export async function verify(
   }
 }
 
-export async function fetchTables(
-  ctx: UserCtx<void, FetchTablesDatasourceResponse>
+export async function information(
+  ctx: UserCtx<void, FetchDatasourceInfoResponse>
 ) {
   const datasourceId = ctx.params.datasourceId
   const datasource = await sdk.datasources.get(datasourceId, { enriched: true })
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -154,7 +154,7 @@ export async function verify(
   }
 }
 
-export async function tables(
+export async function fetchTables(
   ctx: UserCtx<void, FetchTablesDatasourceResponse>
 ) {
   const datasourceId = ctx.params.datasourceId
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -163,9 +163,9 @@ export async function fetchTables(
   if (!connector.getTableNames) {
     ctx.throw(400, "Table name fetching not supported by datasource")
   }
-  const tables = await connector.getTableNames()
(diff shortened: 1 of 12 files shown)


packages/server/src/api/controllers/datasource.ts
```diff
@@ -21,6 +21,7 @@ import {
   CreateDatasourceRequest,
   VerifyDatasourceRequest,
   VerifyDatasourceResponse,
+  FetchTablesDatasourceResponse,
```
Rory-Powell, line 24: `FetchDatasourceTablesResponse`? As tables belong to datasource

packages/server/src/integrations/googlesheets.ts
```diff
@@ -240,6 +243,11 @@ class GoogleSheetsIntegration implements DatasourcePlus {
     }
   }
 
+  getTableNames(): Promise<string[]> {
+    // TODO: implement
```
Rory-Powell, line 247: Remove todo?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -153,6 +154,21 @@ export async function verify(
   }
 }
 
+export async function fetchTables(
```
Rory-Powell, line 157: Now we will have:

```ts
/api/datasources/:id:tables -> datasourceController#fetchTables
```
and
```ts
/api/tables -> tableController#fetch
```

Personally I feel like these entities are so similar it's not worth splitting like this, it'll likely be confusing why both exist.


Preference would be to augment the existing tables endpoint instead, to make it smarter so that we can query by datasource.
```ts
/api/tables?datasourceId=<id>
```

 If the `schema` is a problem we can make it optional. or use another param `noSchema` to make it more obvious.

adrinr: It's actually confusing, but I think they are different and maybe we should rename some. AFAIK, `/tables` are the the tables that a specific datasource has in budibase, while this new `/datasource/tables` are the tables available on the actual source

packages/server/src/integrations/oracle.ts
```diff
@@ -323,6 +326,17 @@ class OracleIntegration extends Sql implements DatasourcePlus {
     this.schemaErrors = final.errors
   }
 
+  async getTableNames() {
+    const columnsResponse = await this.internalQuery<OracleColumnsResponse>({
+      sql: this.COLUMNS_SQL,
+    })
+    if (!columnsResponse.rows) {
```
adrinr, line 333: Is this `if` not redundant?

packages/server/src/api/controllers/datasource.ts
mike12345567: Yes these are very different endpoints - the fetch tables really is just getting the table names - I wasn't really sure how to demonstrate that.

Perhaps a better name for this would be `information` - which retrieves initial information about the datasource+ connection, like what is available.

mike12345567: This has been updated to an `info` endpoint - to reflect its relatively unrelated to all other table endpoints.

adrinr (approved):

PR #10671, opened 2023-05-23
[Fixed] Solve a problem with joi validator middleware

jvcalderon:
## Description

The current implementation of joi validator has a problem; with the current implementation the error code and message are not given in the validation.

Without the change in this PR, we get `404` instead `400`:

```
Error: expect(received).toBe(expected) // Object.is equality

Expected: 400
Received: 404
    at (...)/tests/stripe.spec.ts:63:31
Expected :400
Actual   :404
```

I've added `400` error and body to the validation.

Addresses: 

https://linear.app/budibase/issue/BUDI-6925/update-plan-inside-account-portal

## Documentation

### DIFF
--- a/packages/backend-core/src/middleware/joi-validator.ts
+++ b/packages/backend-core/src/middleware/joi-validator.ts
@@ -29,10 +29,12 @@ function validate(
 
     const { error } = schema.validate(params)
     if (error) {
-      ctx.throw(400, `Invalid ${property} - ${error.message}`)
-      return
+      ctx.status = 400
+      ctx.body = `Invalid ${property} - ${error.message}`
+      ctx.throw(ctx.status, ctx.body)
+    } else {
+      return next()
     }
-    return next()
   }
 }
 


Rory-Powell (commented):
Do we need this change? I think the diff was applicable in the account-portal version of the middleware, but now that we're using the backend-core version, it looks safe as is.

jvcalderon:
It is safe indeed, but it returns `404` instead `400` and an empty body instead the validation message. That's the point for this PR @Rory-Powell.

Rory-Powell:
> It is safe indeed, but it returns `404` instead `400` and an empty body instead the validation message. That's the point for this PR @Rory-Powell.


From trying the API locally, in this case `POST /api/queries` which uses the validator from backend-core, it's producing the correct result. 

Screenshot 2023-05-24 at 14 45 13


I believe it's only an issue in the older implementation that had been used in account portal
https://github.com/Budibase/account-portal/pull/342/files#diff-5c11ee4f5dc7ccaddeaba491726cb2aa671b3d892360cdc3fed264206f05579f

Which you have since removed. Hence I think these changes are unneeded. 

The current impl in backend-core already adds the required `return` statement and formats the response correctly. I think in this changeset we might change the contract of the response body to be a single string, which could cause problems. Or that may be overridden in the subsequent `ctx.throw` I'm not sure.

jvcalderon:
@Rory-Powell Take a look at the following test in the other PR: https://github.com/Budibase/account-portal/pull/342/files#diff-655a81086349b5e26527b581148c72a8a052711bdb42a75f0f840e217ba3b1baR60-R70

Without the current change, it just fails; it doesn't return `400`; just `404`. Maybe it is working in the context of Budibase project but seems like it is not working inside `account-portal`:

Screenshot 2023-05-25 at 12 23 18
Screenshot 2023-05-25 at 12 22 25

But out of the test context, the validator is working properly. For example this call...

```
curl --location --request POST 'http://localhost:4003/api/stripe/checkout-session' \
--header 'Cookie: ...' \
--header 'Content-Type: application/json'
```
...returns:

```
{
    "message": "Invalid body - \"priceId\" is required",
    "status": 400
}
```

Seems like a different behaviour between the request in test environment and the 'up and running' code, so what do you think? Should we keep this change? Should we try to adapt the test? (if not possible, would you remove it?)

jvcalderon:
We decided not merge this change. In the end this is not the desired behavior. It is more related with tests catching the exception.

PR #10680, opened 2023-05-24
Verify data source connection before saves

adrinr:
## Description
Use the new functionality to validate if a datasource configuration is valid before creating/updating it.
These checks have been added to:
- Creating new datasource
- Updating an existing datasource
- Creating a datasource while onboarding

Addresses: 
- https://linear.app/budibase/issue/BUDI-6933/verify-data-source-connection-during-configuration

### Creation
image

### Update
image

### Onboarding
image

### DIFF
--- a/packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
+++ b/packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
@@ -31,6 +31,18 @@
       return "Invalid URL"
     }
   }
+
+  $: urlManuallySet = false
+
+  const updateUrl = event => {
+    const appName = event.detail
+    if (urlManuallySet) {
+      return
+    }
+
+    const parsedUrl = appName.toLowerCase().replace(/\s+/g, "-")
+    url = encodeURI(parsedUrl)
+  }
 </script>
 
 <div>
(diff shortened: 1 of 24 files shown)


packages/builder/src/pages/builder/portal/apps/onboarding/_components/NamePanel.svelte
```diff
@@ -31,6 +31,18 @@
       return "Invalid URL"
     }
   }
+
+  $: urlManuallySet = false
```
adrinr, line 35: This is not related to this ticket, changing the onboarding app to URL sync for an better dev/user experience

deanhannigan (approved):
LGTM!

packages/server/src/integrations/airtable.ts
```diff
@@ -20,7 +20,9 @@ const SCHEMA: Integration = {
     "Airtable is a spreadsheet-database hybrid, with the features of a database but applied to a spreadsheet.",
   friendlyName: "Airtable",
   type: "Spreadsheet",
-  features: [DatasourceFeature.CONNECTION_CHECKING],
+  features: {
+    [DatasourceFeature.CONNECTION_CHECKING]: true,
```
mike12345567, line 24: What is the reason for the conversion from an array of features supported, to a map of booleans? Is there a benefit to doing it this way?

Only reason I ask is that this style of feature management (via an array) has been implemented into automations as part of Peter's work towards sync automations - just want to check if we should be refactoring it as well if there is a benefit.

adrinr: The main reason I changed that is for the way we consume it. This features will be mainly a lookup, and `features[Feature.X]` is more efficient and clearer than `features.find(x=>Feature.X)`.
If that's a problem we can keep it as an array and reduce it when required

PR #10684, opened 2023-05-24
Chore - Explicit type checks

adrinr:
## Description
Make type checks of the esbuild built projects more explicit

### DIFF
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           node-version: 14.x
           cache: "yarn"
-      - run: yarn
+      - run: yarn install --frozen-lockfile
       - run: yarn lint
 
   build:
@@ -43,7 +43,7 @@ jobs:
         with:
           node-version: 14.x
           cache: "yarn"
-      - run: yarn
+      - run: yarn install --frozen-lockfile
       # Run build all the projects
       - run: yarn nx run-many -t=build
       # Check the types of the projects built via esbuild
@@ -61,7 +61,7 @@ jobs:
         with:
           node-version: 14.x
           cache: "yarn"
-      - run: yarn
+      - run: yarn install --frozen-lockfile
       - run: yarn test --ignore=@budibase/worker --ignore=@budibase/server --ignore=@budibase/pro
       - uses: codecov/codecov-action@v3
         with:
@@ -81,7 +81,7 @@ jobs:
         with:
           node-version: 14.x
           cache: "yarn"
-      - run: yarn
+      - run: yarn install --frozen-lockfile
       - run: yarn test --scope=@budibase/worker --scope=@budibase/server
       - uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/budibase_ci.yml
```diff
@@ -44,7 +44,10 @@ jobs:
           node-version: 14.x
           cache: "yarn"
       - run: yarn
-      - run: yarn nx run-many -t=build --configuration=production
+      # Run build all the projects
+      - run: yarn nx run-many -t=build
```
Rory-Powell, line 48: `types`, `backend-core` etc, looks like they still use `tsc`. Does that mean we'll still have potentially slower build times on those packages?

adrinr: Yes, these packages still use tsc, so we have the same performance as before. The `configuration=production` has only effect if using `nx/esbuild`, for this I removed it into the package itself as the pipeline should not be aware of this things, otherwise we will run into issues. I actually already got some issues that I missed adding the checks to `server`, as I needed to remove the `nx/esbuild` from there because of vite.

package.json
```diff
@@ -33,6 +33,7 @@
     "bootstrap": "./scripts/bootstrap.sh  && ./scripts/link-dependencies.sh",
     "build": "yarn nx run-many -t=build",
     "build:dev": "lerna run --stream prebuild && yarn nx run-many --target=build --output-style=dynamic --watch --preserveWatchOutput",
+    "check:types": "lerna run check:types",
```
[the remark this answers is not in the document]
adrinr: The normal build already checks its types, so I did not want to duplicate steps. Maybe it would be cleaner to add the `check:types` to all the packages, but then it would be incoherent with the "non-typed" packages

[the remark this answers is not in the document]
Rory-Powell: However that might not be possible, the type checking for a module that relies on another module would always require that `.d.ts` files have been produced

.github/workflows/budibase_ci.yml
```diff
@@ -44,7 +44,10 @@ jobs:
...
+      # Check the types of the projects built via esbuild
+      - run: yarn check:types
```
Rory-Powell, line 50: Should we run the type check before the build, in case any artifact is ever produced by the type checking that would interfere with the `dist` dir

adrinr: `yarn check:types` might also produce dist folders, as it will need packages like `string-templates` to be built

package.json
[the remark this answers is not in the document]
Rory-Powell: One option might be to use a top level file `tsconfig.json`

```json
{
  "references": [
    { "path": "./packages/types" },
    { "path": "./packages/backend-core" },
    { "path": "./packages/server" },
    { "path": "./packages/worker" }
   ...etc
  ],
  "files" :[]
}
```

As the target for `tsc -p`. This would run the typescript build in order for all the dependencies listed. Provided that the `references` of each sub `package.json` includes the packages that are depended on. 

This file is currently in master, although it looks to have been removed in `develop`. This is actually a required file for `yarn build:dev`

.github/workflows/budibase_ci.yml
adrinr: Yes, we will be moving them over now that everything works fine. The preference was worker and server, because doing this we can now achieve npm-less docker images

adrinr:
Some of these changes have already been done via other PRs

PR #10687, opened 2023-05-24
ESbuild single image fixes + nodemon changes

mike12345567:
## Description
Fixes the single image build, as well as fixing the nodemon process for the server.


aptkingston (approved):

hosting/single/Dockerfile
```diff
@@ -1,22 +1,22 @@
...
 # install node-gyp dependencies
 RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-recommends apt-utils cron g++ make python
 
 # add pin script
 WORKDIR /
-ADD scripts/pinVersions.js scripts/cleanup.sh ./
+ADD scripts/cleanup.sh ./
 RUN chmod +x /cleanup.sh
 
 # build server
```
adrinr, line 11: We will be already using the built version, right? If so, this comment needs to be removed and we can change this to be `yarn install --frozen-lockfile --production=true`

hosting/single/Dockerfile
```diff
@@ -1,22 +1,22 @@
...
 WORKDIR /app
 ADD packages/server .
-RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
+RUN yarn && /cleanup.sh
 
 # build worker
 WORKDIR /worker
 ADD packages/worker .
-RUN node /pinVersions.js && yarn && yarn build && /cleanup.sh
+RUN yarn && /cleanup.sh
```
adrinr, line 19: Same here

hosting/single/Dockerfile
adrinr: Actually, we need to do the same to the worker/server dockerfile

mike12345567: Fixed - added for server/worker as well.

adrinr (approved):

PR #10694, opened 2023-05-25
Chore - Respect esbuild config paths

adrinr:
## Description
We found an issue that if a package (lets say `backend-core`) was built, when building the server via esbuild, that dist folder was used instead of using the tsconfig.paths defined. This broke our dev experience.
This PR is solving this by using the `tsconfig-paths` esbuild plugin

### DIFF
--- a/packages/server/tsconfig.build.json
+++ b/packages/server/tsconfig.build.json
@@ -8,7 +8,7 @@
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "incremental": true,
-    "types": ["node", "jest"],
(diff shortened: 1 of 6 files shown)


packages/server/tsconfig.build.json
```diff
@@ -8,7 +8,7 @@
     "esModuleInterop": true,
```
adrinr, line 8: These tsconfig.json are not really related to the issue, I just tidied them up while analysing the issue

mike12345567 (approved):

PR #10695, opened 2023-05-25
Grid enhancements from user feedback

aptkingston:
## Description
This is a collection of new features and improvements to the grid based on user feedback. There are also some small fixes.

New features:
- Add ability to stripe rows alternate colours. This is disabled at the moment, but the functionality is there so that we can offer it as a setting in the app component when we ship that
image
image
- Automatically scroll grid left/right when dragging columns and near table edges. This makes it much easier to reorder columns in large grids

https://github.com/Budibase/budibase/assets/9075550/c2d62f8e-88e9-47ba-8349-2385fc1297e2

- Center align boolean fields
image

Fixes:
- Fix escape keybind being unreadable in light theme
- Fix issue with bulk deletion if selecting then unselecting a row
- Disable ability to hide the sticky column via the header context menu
- Fix z-index of add column button

Addresses:
- (Linear) BUDI-6996
- (Linear) BUDI-6998 
- (Linear) BUDI-6998


aptkingston:
Merging this in because I want to pull it into the data section multi-dev and get it all up on QA for manual testing.

PR #10701, opened 2023-05-25
Post Sign Up Onboarding Changes

Ghrehh:
## Description
- Removes the data portion of the existing onboarding experience, instead pushing users directly to the new data creation screen.
- Users can now add sample data to existing apps.
- Use the new data screen instead of the existing modal for creating new data sources.
- The last onboarding step will now deposit the user back on the new data screen

Addresses: 
https://linear.app/budibase/issue/BUDI-6949/adapt-post-sign-up-flow
https://linear.app/budibase/issue/BUDI-6931/remove-data-source-modal-and-add-empty-state-for-data-source-page

## Screenshots
|New Page|New Page With Datasource Modal|New Page with Close Button|
|---|---|---|
|Screenshot 2023-05-25 at 16 14 51|Screenshot 2023-05-25 at 16 14 58|Screenshot 2023-05-25 at 16 15 10|


packages/builder/src/components/backend/DatasourceNavigator/modals/GoogleDatasourceConfigModal.svelte
deanhannigan, line 1: If you leave out the variable `onCancel` entirely, won't it achieve the same result? If it's undefined, it does nothing anyway.

melohagan:
Looks awesome! Just wondering if there could be any sorting or grouping added. The datasources don't seem to presented in any particular order in the screenshots.

packages/builder/src/components/backend/DatasourceNavigator/modals/GoogleDatasourceConfigModal.svelte
Ghrehh: Doh, good catch.

Ghrehh:
@melohagan Good point! I've copied the existing sorting code now, so it should be the same sorting as it was previously.

aptkingston (changes requested):
I haven't been through the code yet, but I pulled this down to try it and ran into a few issues straight away:

- The colours are static and don't respect the users theme
image
- ~~Oracle DB is missing~~ (edit - this is just my env)
- There should be 4 columns of datasources rather than unlimited. You can check the figma for reference
- Clicking on any external datasource throws an error and does nothing
image
- When hovering over a datasource, the cursor flips between the default cursor and `pointer` multiple times. This should consistently be `pointer`
- The notification when creating a table says "table foo created"
image

Let's get those sorted first and then I'll review the code 👍

Ghrehh:
@aptkingston Ah, my bad particularly on the external data sources not working, but update when implementing sorting.

All your feedback should be sorted now.

aptkingston (approved):
Code LGTM! Just those few other changes mentioned in my previous comment from manual testing, then should be good to go 👍

deanhannigan (approved):
LGTM! This looks fantastic!

Ghrehh:
@aptkingston Should be good to review again if you want to have a gander

PR #10713, opened 2023-05-26
Fix for fetching Internal DB Views in QA cloud

deanhannigan:
## Description

The indirect call to `eval` in `inMemoryViews.ts` was causing view fetches in the QA cloud system to fail.
Reverting back to a direct call fixes the issue.

Addresses: 
- BUDI-7060


mike12345567 (approved):
Looks good!

PR #10714, opened 2023-05-26
Build worker using esbuild script

adrinr:
## Description
Using nx/esbuild is causing a few issues, for example outputting different extensions depending on the configuration. This PR gets rid of the nx/esbuild to use our custom script that is anyways required for the server build


mike12345567 (approved):
Looks good - thanks for getting this sorted quickly!

PR #10717, opened 2023-05-26
Reset pro dependencies before the npm push

adrinr:
## Description
Fixing an issue caused by running the monorepo as open source contributor.
They will be pulling the pro packages that will have versions of @budibase/whatever as dependency. Because locally we have 0.0.1, it will pull these versions. This will cause conflicts between both versions of packages being located in different folders.

These changes will ensure that pro can only work within the monorepo, linking with the expected packages.


mike12345567 (approved):
Looks good! Good to know that `*` won't pick the local version

PR #10736, opened 2023-05-30
Test/qa 20 add relationship tests to datasources

Pedro-Revez-Silva:
## Description
This PR adds a test to the creation of a relationship from BB in a MySQL database. 

Addresses: 
- https://linear.app/budibase/issue/QA-20/add-test-scenarios-for-relationships-in-datasources-tests


packages/types/src/documents/app/datasource.ts
```diff
@@ -12,7 +12,7 @@ export interface Datasource extends Document {
   }
   plus?: boolean
   entities?: {
-    [key: string]: Table
+    [key: string]: any
```
Rory-Powell, line 15: Is there a difference between an entity and a table? If so, would be good to type whatever `Entity` is rather than `any`. Any thoughts here @mike12345567 ?

qa-core/package.json
```diff
@@ -17,7 +17,8 @@
     "test:smoke": "yarn run test --testPathIgnorePatterns=/.+\\.integration\\.spec\\.ts",
     "test:ci": "start-server-and-test dev:built http://localhost:4001/health test:smoke",
     "serve": "start-server-and-test dev:built http://localhost:4001/health",
-    "dev:built": "cd ../ && yarn dev:built"
+    "dev:built": "cd ../ && yarn dev:built",
+    "test:datasources": "jest --runInBand --json --outputFile=testResults.json --testPathPattern='datasources/.*\\.spec\\.(ts|js)$'"
```
Rory-Powell, line 21: We can probably remove the `test:smoke` now? I take it this is the replacement. How does it affect the commands for the test that run against QA env?

qa-core/src/internal-api/tests/dataSources/postgresSQL.integration.spec.ts
```diff
@@ -37,7 +37,7 @@ describe("Internal API - Data Sources: PostgresSQL", () => {
       fixtures.queries.postgres(updatedDataSourceJson.datasource._id!)
     )
 
-    expect(queryJson.rows.length).toEqual(91)
+    expect(queryJson.rows.length).toEqual(92)
```
Rory-Powell, line 40: what happened here lol

qa-core/src/internal-api/fixtures/datasources.ts
```diff
@@ -70,3 +73,50 @@ export const restAPI = (): DatasourceRequest => {
...
 }
+
+export const generateRelationshipForMySQL = (
+  updatedDataSourceJson: any
+): Datasource => {
+  const entities = updatedDataSourceJson!.datasource!.entities!
+  const datasourceId = updatedDataSourceJson!.datasource!._id!
+  const relationShipBody = {
+    ...updatedDataSourceJson.datasource,
+    entities: {
```
mike12345567, line 84: If there is an issue with typing here - it would be better to cast this structure through `as any as Datasource`  rather than updating the core types - although it would be good to work out what is not valid in this structure for `entities` to not read as a `Table` type.

Mitch-Budibase (approved):

shogunpurple (approved):
LGTM!

Rory-Powell (approved):

PR #10737, opened 2023-05-30
Fix performance issue with looping in automations

PClmnt:
## Description
This PR fixes a performance big that was happening whenever looping was used in tandem with a large object. The object was being passed into `processObject` every single loop of the block resulting in an exponentially growing chunk being stored in memory. 

To give an example, trying to iterate over a 10mb piece of data was resulting these sorts of times:

Screenshot 2023-05-25 at 16 48 03

Nearly 27 seconds per loop. 

The fix was actually extremely simple - there was no need to be calling `processObject` within the loop, as it was actually being done outside the loop as well. The problematic `newInput` declaration was essentially doing the same thing but in a less performant matter due to being inside the loop.  It just took me longer than expected to realise this second call was totally unnecessary! 

With the code updated it is now taking around 5 seconds to perform the full automation - with 200 iterations (max limit set by an env var) 

Screenshot 2023-05-30 at 14 23 05

n.b Made a slight change to the `getLoopIterations` method within the file as well, but this has nothing to do with the fix. Just tidying things.


Rory-Powell (approved):

packages/server/src/threads/automation.ts
```diff
@@ -479,32 +475,54 @@ class Orchestrator {
       }
     }
 
+    const end = performance.now()
+    const executionTime = end - start
+
+    console.log(`Execution time: ${executionTime} milliseconds`)
```
Rory-Powell, line 481: Should remove

packages/server/src/threads/automation.ts
```diff
@@ -479,32 +475,54 @@ class Orchestrator {
       }
     }
 
+    const end = performance.now()
+    const executionTime = end - start
```
Rory-Powell, line 479: Should remove these timers

PR #10738, opened 2023-05-30
Automation investigation logging + max doc size fix for automation log

Rory-Powell:
## Description
- Add `pid` to logs
- Add check for max size on automation log document
- Add name of forked process to environment + corresponding logging
- Add automation context (just the automation id currently) and integrate with logging
- Update timeout error message on threads to not mention queries 

Addresses: 
- Partial part of a wider fix for ongoing issues with automation performance


PClmnt (approved):
LGTM

PR #10740, opened 2023-05-30
Add teardown logic for multitenancy

Pedro-Revez-Silva:
## Description
Adds a function to the teardown process that calls the endpoint that will delete the account & tenant for a test. The intention is to be able to run tests in production environments without adding too much fake data. 

Addresses: 
- https://linear.app/budibase/issue/QA-23/add-test-clean-up


shogunpurple (approved):
LGTM

shogunpurple (approved):
LGTM

Issue #10741, opened 2023-05-30
Error opening apps when using single container, and CUSTOMDOMAIN


gcglinton:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker single image
  - Budibase Version: 2.6.11

**Describe the bug**
When trying to open an app in either preview, or published mode, the browser would throw an error:
> There was an error loading your app
> The Budibase client library could not be loaded. Try republishing your app.

When checking things with the Developer tools, I can see that there's some URLs throwing 403 errors, all of which are under `/files`.

**To Reproduce**
- Deploy Budibase with Docker, using image `budibase-aas`, and set a custom domain with CUSTOMDOMAIN envar
- Create new application
- View the preview, or published application

**Expected behavior**
Application loads successfully.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Edge
 - Version: 113.0.1774.50 

**Additional context**
I'm running in Azure Container Instance, but I suspect this problem would exist in any and all deployments.

PR #10742, opened 2023-05-30
Make SSL and non-SSL nginx configs match

gcglinton:
The nginx config for SSL didn't match the one for non-SSL, so things didn't work as expected. 
Copied the default one to the SSL one, re-adding the few directives/parameters required to make SSL work.

Also standardized indentation for the default nginx config; some blocks had only two spaces, but everywhere it was 4. Now it's 4 everywhere.

## Description
The nginx config that's used when enabling SSL is missing some `location` blocks, and it causes the apps to not load.
Addresses: 
- #10741 

## Documentation


gcglinton:

gcglinton:
recheck

PR #10744, opened 2023-05-31
Update logging for automations

PClmnt:
## Description
Add some performance logging within the automation thread.

### DIFF
--- a/packages/server/src/threads/automation.ts
+++ b/packages/server/src/threads/automation.ts
@@ -252,7 +252,6 @@ class Orchestrator {
         return
       }
     }
-    const start = performance.now()
     for (let step of automation.definition.steps) {
       if (timeoutFlag) {
         break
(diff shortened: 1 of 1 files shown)


Rory-Powell (commented):
It might actually be useful to keep this in! If we format it differently we can pull it out as a metric in dd. Will comment on the line

packages/server/src/threads/automation.ts
```diff
@@ -474,11 +473,6 @@ class Orchestrator {
       }
     }
 
-    const end = performance.now()
-    const executionTime = end - start
-
-    console.log(`Execution time: ${executionTime} milliseconds`)
```
Rory-Powell, line 480:
```ts
console.info(`Execution time: ${executionTime} milliseconds`, {  _logKey: "automation",  executionTime })
```

PClmnt: Done!

Rory-Powell (approved):

PR #10748, opened 2023-05-31
Update concurrency settings for release and prerelease

Rory-Powell:
- Unique concurrency groups for each tag / release job
- Specifically set `cancel-in-progress: false` to try to fix apparent bug


adrinr (approved):
🎉

PR #10753, opened 2023-05-31
Websocket redis rewrite for multi-user collaboration

aptkingston:
## Description
This PR rewrites entire backend portion of multi-user collaboration to hopefully work properly in multi-instance environments. The new implementation relies on redis to keep track of the state of all sessions, allowing all server instances to be fully aware of all connected users at any given time.

The socket.io redis adapter is still being used to proxy messages between server instances, but this was completely insufficient to actually enable multi-user collaboration as we need to be able to know the full connection state at any time, and proxying messages is not enough to accomplish this.

A lot of this has been run past @mike12345567 so hopefully follows best practices.

Changes:
- The `selectDb` flag in our redis client wrapper was not actually being used to select a separate database. As this was probably deliberate, all current redis keys have been kept inside the same default database, but a new one has been added for usage with websocket state. The `selectDb` flag now does actually use the selected database provided.
- Each session is persisted in redis with a 60 second TTL
- Each "room", or group of sessions relating to the same entity, has a separate redis entry which acts as a lookup to find all sessions for a given room, removing the need for any scan operations. These maps do not have any TTL.
- A heartbeat has been added to keep active connections alive in redis
- Websocket events are enumerated
- If we "miss" the websocket disonnection event for any reason, leaving stale data in redis, then it will be pruned the next time a user connects to that particular entity that was in use by the missing user

### Technical stuff - the challenge of multi-instance environments
The initial design for this simply used the redis adapter provided by socket.io to proxy messages to all instances. This works, and when a user connects or disconnects we can successfully inform users connected to any server instance.

The issue with this design was that when you initially connect to a socket, we pass back the list of all connected users (to render the avatars for example). This didn't work as intended, because each server instance doesn't know who's connected to the others, and clients aren't aware of others unless they were already connected, in which case they would have received the message.

Each server instance basically needs to know the state of all connections, and redis is the best place to persist this info as server crashes or restarts would lead to inconsistent state if the servers stored info in memory themselves.

The next design used redis to persist the state of all connected clients inside a single key per room, which meant it was trivial to get all clients for a room. These keys had no TTLs, because there would be no way to update the TTL as clients don't perform actions on the websocket often. The fundamental flaw with this approach was that without a TTL in redis there's pretty much no way to avoid stale data which would then persist indefinitely. There's guaranteed to be edge cases with server restarts, or crashes, that would circumvent the normal websocket disconnection cleanup logic and skip the clearing of redis. We tried to design something that could just wipe this whole redis database on a restart, but every design had some sort of issue with getting out of sync, or breaking other server instances when one restarted..

The current design has a TTL, which means that no matter what we won't have stale data in redis. It's a bit more complicated to reason about, but I'm hopeful it will work as expected in any sort of multi-instance environment, as well as properly handle any sort of crash, restart or rolling updates etc. The worst case scenario is that stale data is persisted in the frontend for users until a new user connects, which triggers a prune of missing sessions in redis.

The flow is basically:
- User connects to a room
- That room is pruned, removing any session IDs from the redis lookup map which no longer exist in redis due to TTL expiry
- The room lookup map is fetched from redis, containing all session IDs in that room
- A bulk request gets all entries for those keys
- Session info is sent back to the user
- User starts a heartbeat to persist their own session


packages/backend-core/src/redis/init.ts
```diff
@@ -6,17 +6,19 @@ let userClient: Client,
...
+  lockClient: Client,
+  socketClient: Client
 
 async function init() {
   userClient = await new Client(utils.Databases.USER_CACHE).init()
   sessionClient = await new Client(utils.Databases.SESSIONS).init()
   appClient = await new Client(utils.Databases.APP_METADATA).init()
   cacheClient = await new Client(utils.Databases.GENERIC_CACHE).init()
   lockClient = await new Client(utils.Databases.LOCKS).init()
-  writethroughClient = await new Client(
```
mike12345567, line 17: Just for awareness - this is no longer in a database as the database selection was not working - so all writethrough data is currently in the default key space.

We could probably do with moving sessions out into a separate database to reduce scan load but its not an issue currently so no rush.

packages/client/src/websocket.js
```diff
@@ -18,7 +18,7 @@ export const initWebsocket = () => {
   }
 
   // Initialise connection
-  socket = createWebsocket("/socket/client")
+  socket = createWebsocket("/socket/client", false)
```
mike12345567, line 21: can we update this boolean parameter to an object specifier? Looking at this in PR i'm not sure what purpose the `false` has/does as part of the websocket creation.

packages/frontend-core/src/utils/websocket.js
```diff
@@ -1,6 +1,7 @@
 import { io } from "socket.io-client"
+import { SocketEvent, SocketSessionTTL } from "@budibase/shared-core"
 
-export const createWebsocket = path => {
+export const createWebsocket = (path, heartbeat = true) => {
```
mike12345567, line 4: I see it now - thats cool to have an option for web socket creation with heartbeat - I imagine we'll use this for other stuff.

packages/server/src/websockets/websocket.ts
```diff
@@ -86,10 +87,184 @@ export default class Socket {
...
+      return []
+    }
+  }
+
+  // Detects keys which have been pruned from redis due to TTL expiry in a certain
+  // room and broadcasts disconnection messages to ensure clients are aware
+  async pruneRoom(room: string) {
+    const sessionIds = await this.getRoomSessionIds(room)
+    const sessionsExist = await Promise.all(
+      sessionIds.map(id => this.redisClient?.exists(this.getSessionKey(id)))
```
mike12345567, line 173: I'm fairly sure the `exists` operation can take N number of keys and perform the exists check all at once - this might be quite a bit more performant.
image

mike12345567 (approved):
LGTM!

packages/server/src/websockets/websocket.ts
mike12345567: Scrap this - exists isn't useful in this scenario - doesn't tell us which keys exist.

packages/frontend-core/src/utils/websocket.js
aptkingston: Yep the generic `BaseSocket` in the server, which all the others are based off, handles this automatically. So we'll get multi-instance user management handled for free going forward which should make things much easier.

packages/server/src/websockets/websocket.ts
aptkingston: As discussed privately, can't use this since it only returns the number of keys which exist rather than which exact keys. Unfortunately we need to know the exact keys.

packages/client/src/websocket.js
aptkingston: I will update this at some stage, but CI is passing and I'm 90% sure I'll need another PR after this anyway, so merging for now!

PR #10755, opened 2023-05-31
Datasource info endpoint update

mike12345567:
## Description
This PR includes two changes:
1. Updating datasource information endpoint to POST which allows sending up an unfinished/unsaved datasource for fetching information with. Also changing how verification and information endpoints work so that enrichment is used and therefore env vars can also be used.

EDIT: Decided to not add the removal of `ioredis-mock` to this PR as it complicates things and it appears we have issues with the pro package passing tests when using real Redis - this requires more research and isn't strictly necessary as we don't have any tests for the websockets currently.


packages/backend-core/src/index.ts
```diff
@@ -21,6 +21,7 @@ export * as context from "./context"
 export * as cache from "./cache"
 export * as objectStore from "./objectStore"
 export * as redis from "./redis"
+export { Client as RedisClient } from "./redis"
```
aptkingston, line 24: Probably wise to be explicit about this export anyway, but just an FYI that it can already be consumed as:
```ts
import { redis } from "@budibase/backend-core"
let socket: redis.Client
```
But I definitely prefer the explicit syntax!

packages/server/src/websockets/index.ts
```diff
@@ -3,15 +3,19 @@ import Koa from "koa"
...
+let clientAppSocket: ClientAppSocket | undefined
+let gridSocket: GridSocket | undefined
+let builderSocket: BuilderSocket | undefined
 
 export const initialise = (app: Koa, server: http.Server) => {
-  clientAppSocket = new ClientAppSocket(app, server)
-  gridSocket = new GridSocket(app, server)
-  builderSocket = new BuilderSocket(app, server)
+  // currently we use ioredis-mock which doesn't work with the websocket system for testing
+  if (!env.isTest()) {
```
aptkingston, line 14: Is this still needed even with ioredis-mock removed now? Just since the comment mentions it.

aptkingston (approved):
LGTM!

packages/backend-core/src/index.ts
mike12345567: In this case I'm explicitly exporting the type, we've had issues with this in the past, but you need to export from the index of backend-core for the types to be consumable.

packages/server/src/websockets/index.ts
mike12345567: It is not - good catch!

packages/server/src/api/routes/datasource.ts
```diff
@@ -20,8 +20,8 @@ router
     authorized(permissions.BUILDER),
     datasourceController.verify
   )
-  .get(
-    "/api/datasources/:datasourceId/info",
+  .post(
```
adrinr, line 23: Shouldn't this be kept as a get?

packages/server/src/websockets/builder.ts
```diff
@@ -14,9 +14,9 @@ export default class BuilderSocket extends BaseSocket {
     super(app, server, "/socket/builder", [authorized(permissions.BUILDER)])
   }
 
-  async onConnect(socket: Socket) {
+  async onConnect(socket?: Socket) {
```
adrinr, line 17: This should probably be handled at the caller level. Will it ever need to be called `onConnect` without a socket?

packages/server/src/api/routes/datasource.ts
mike12345567: No, we need a post body for this - so we can send up the datasource contents.

packages/server/src/websockets/builder.ts
mike12345567: I didn't write this - just fixed the typescript definition!

PR #10767, opened 2023-06-02
Chore - Update node versions

adrinr:
## Description
Updating node versions from 14.x to 18.x


packages/server/Dockerfile
```diff
@@ -27,7 +27,7 @@ COPY scripts/integrations/oracle/ scripts/integrations/oracle/
 RUN /bin/bash -e scripts/integrations/oracle/instantclient/linux/x86-64/install.sh 
 
 COPY package.json .
-RUN yarn install --frozen-lockfile --production=true
+RUN yarn install --production=true
```
Rory-Powell, line 30: I see that we don't copy the lock file across, how does that work?
Wouldn't it mean that the versions we have in our lock don't reflect what will be in the image?

Rory-Powell, line 30: Just a general question on the `install` - In local dev how does this work with dependencies? 
I take it now that we're building a bundled dist file the bb deps will be included there. Does the install just pull down a fixed version from npm that's ignored?

PR #10772, opened 2023-06-02
Chore - Fix OS contributor dev experience

adrinr:
## Description
Fix esbuild issue causing pro not to load properly when running as external contributor


mike12345567 (approved):
Brilliant - thank you for handling the ongoing pain 😆

PR #10775, opened 2023-06-02
Collaboration fixes

aptkingston:
## Description
This PR contains a few fixes after testing collaboration with many developers.

Changes:
- Ensure everything works as intended even when websockets are disabled
- Fix new tables not appearing without websocket enabled
- Fix other users changes causing you to lose focus from a cell when creating new rows
- Fix relationships not appearing after making changes (thanks @mike12345567)
- Fix first row creation in empty tables creating 2 rows
- Fix duplicating row action creating duplicate IDs
- Fix JSON editor overflow row edit modal
- Fix grid and backend store state getting out of sync

A lot of those fixes were enabled by refactoring websockets so that clients now ignore all events triggered by themselves. This is not trivial because we initiate changes by an API request rather than sending data up the websocket, otherwise it would be trivial to emit to everyone else other than yourself. The motivation for this was that the websocket was so fast that the websocket message would be received before the client even thought that the API request it just sent had completed.

This feature works as follows:
- All instances of the builder have a "session ID", which is just a random ID that is generated in memory and constant until you refresh the page. The ID is unique for each tab opened.
- This ID is sent up with all requests as a header
- When websocket messages are triggered by API events, the originating session ID is sent back down in the message
- Clients have a new utility function `socket.onOther` rather than `socket.on` which will only trigger on events made by others, by ignoring messages you originate from their own session ID

### DIFF
--- a/packages/backend-core/src/constants/misc.ts
+++ b/packages/backend-core/src/constants/misc.ts
@@ -16,6 +16,7 @@ export enum Header {
   LICENSE_KEY = "x-budibase-license-key",
   API_VER = "x-budibase-api-version",
   APP_ID = "x-budibase-app-id",
+  SESSION_ID = "x-budibase-session-id",
   TYPE = "x-budibase-type",
   PREVIEW_ROLE = "x-budibase-role",
   TENANT_ID = "x-budibase-tenant-id",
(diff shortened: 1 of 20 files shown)


mike12345567 (approved):
Love the `onOther` very handy abstraction!

PR #10777, opened 2023-06-02
Fix/google sheets issues

mike12345567:
## Description
Issues covered in: https://linear.app/budibase/issue/BUDI-7080/inconsistencies-with-rows-api-when-using-google-sheets
`_id` field was not filterable within Google Sheets and the deletion endpoint for Google sheets did not return the `_id` of the field which had been deleted.


aptkingston (approved):
LGTM!

PR #10783, opened 2023-06-05
Refactor Datasource Creation Modal

Ghrehh:
# Addresses (partially)
https://linear.app/budibase/issue/BUDI-6936/change-configuration-display-in-data-source-details

# Explanation
Refactors the datasource configuration modal and surrounding code in preparation for using this modal to edit datasources in future.

# Changes

- Previously the `DatasourceConfigModal`, the `IntegrationConfigForm` and some surrounding code used a slightly confusing series of objects to go from an integration definition to creating a datasource. This PR changes that so that the only things passed around are an unchanging integration definition, which we pull directly from the server, as well as an object containing the fields and their values. These are then provided to the datasources store's new `create` method which validates and creates the datasource. These changes should make it easier to consume these components for use in the above mentioned ticket.
- Added a `yarn test:watch` command to the package.json for convenience, oversight on my part to leave this out to begin with.
- Removed some unused components that were in a `_components` directory, and moved the only component there still in use to alongside these new components, which is the only place it's still used.
- Added a sortedIntegrationsStore the derives from the base integrations store, this could possibly replace the default integrations store, but that'll require a bit more refactoring.
- This directly copies the `IntegrationConfigForm` this is only temporary, and the old version will be deleted once the above ticket is fully finished.
- We previously had the components for creating a google sheets datasource entirely separate from the components for creating other datasources, despite there being a lot of commonality. These changes unify them a great deal, and the logic for the different potential stages of the datasource creation flow are now handled by a single [store](https://github.com/Budibase/budibase/blob/refactor-datasource-creation-modal/packages/builder/src/pages/builder/app/%5Bapplication%5D/data/_components/CreateExternalDatasourceModal/stores/datasourceCreation.js) that should hopefully make it as easy as possible to understand different screens that are presented for each integration at a glance.

#Notes
- There's still a bunch of vestigial code existing that is used by the datasource editing/updating flows, this will be cleaned up in a follow up.


aptkingston (commented):
Code LGTM! Will pull this down and test locally as well.

packages/builder/src/stores/backend/datasources.js
```diff
@@ -1,16 +1,23 @@
...
     list: [],
     selectedDatasourceId: null,
     schemaError: null,
   })
+
   const derivedStore = derived(store, $store => ({
     ...$store,
     selected: $store.list?.find(ds => ds._id === $store.selectedDatasourceId),
+    hasDefaultData:
+      $store.list.findIndex(
```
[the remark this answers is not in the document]
Ghrehh: Beaut, TIL, changed this.

Ghrehh:
@aptkingston Had to work around some changes to google integration that got merged in under me, I shimmed some fixes in to get the google code working but will properly address it in a branch/pr off this I'm working on. Could you give this another look when you get a chance please?

packages/builder/src/pages/builder/auth/login.svelte
```diff
@@ -11,7 +11,7 @@
   } from "@budibase/bbui"
   import { goto } from "@roxi/routify"
   import { auth, organisation, oidc, admin } from "stores/portal"
-  import GoogleButton from "./_components/GoogleButton.svelte"
+  import GoogleButton from "components/common/GoogleButton.svelte"
```
deanhannigan, line 14: This button needs to be switched back to the one at the previous location `./_components/GoogleButton.svelte`. The updated button is used in Google authorisation flows within the builder and cannot function properly/display in the login page. 

Screenshot 2023-06-12 at 14 32 40

Ghrehh: Jeepers, my bad. Dodgy find and replace I guess 🤦

aptkingston (changes requested):
Nice one Gerard! Nice work adding some tests in there as well.

Playing around with this locally to test it, and these are the things I noticed:

- We seem to enforce a password length when connecting to a datasource, which doesn't make sense. It's probably just down to automatically applying "password" validation to a password field, but since we're entering an existing password rather than creating a new one, we shouldn't have length validation here:
image
- Could we avoid showing validation errors initially? Ideally they'd only show after being touched or after the user attempts to connect, but any change so that they aren't shown immediately would be great
image
- When there's an error connecting, could we show that? Thanks to the new verification feature we have quite descriptive errors (as shown) so it would make sense to show these to the user
image

aptkingston (approved):
LGTM! Really went the extra mile here, defining a full new set of components for editing the config. Hopefully will make this a lot more understandable and easier to maintain in future. Nice one!

Issue #8198, opened 2022-10-10
[BUDI-7082] Send meeting invite via automation email SMTP


melohagan:
**Describe the feature request**
As a user I would like to send email meeting invites.

Related discussion: https://github.com/Budibase/budibase/discussions/7980
Related SO post: https://stackoverflow.com/questions/24323452/how-to-send-a-meeting-request-correctly-with-nodemailer

<sub>[BUDI-7082](https://linear.app/budibase/issue/BUDI-7082/send-meeting-invite-via-automation-email-smtp)</sub>

andyburgessmd:
Hopefully this will reopen the FR as I think this would be really useful to have.

PR #10787, opened 2023-06-05
Send meeting invite via automation email smtp

melohagan:
## Description
Users wanted the ability to send an email meeting invite. 

Using the standard `.ics` attachment approach, with **ical-generator** building the content.

Addresses: 
- https://github.com/Budibase/budibase/issues/8198

## Screenshots

unchecked

checked

email

## Documentation


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -317,210 +324,233 @@
 
 <div class="fields">
   {#each deprecatedSchemaProperties as [key, value]}
-    <div class="block-field">
```
mike12345567, line 320: Do you know what formatting change occurred here for this file to be so different?

packages/server/src/automations/steps/sendSmtpEmail.ts
```diff
@@ -48,6 +48,35 @@ export const definition: AutomationStepSchema = {
           type: AutomationIOType.STRING,
           title: "HTML Contents",
         },
+        addInvite: {
+          type: AutomationIOType.BOOLEAN,
+          title: "Add calendar invite?",
```
mike12345567, line 53: We don't generally make these questions, I think it should just be `Add calendar invite` to be consistent with other checkbox driven options in the system.

packages/server/src/utilities/workerRequests.ts
```diff
@@ -65,15 +65,35 @@ async function checkResponse(
...
-  bcc: string,
+export async function sendSmtpEmail({
+  to,
+  from,
+  subject,
+  contents,
+  cc,
+  bcc,
+  automation,
+  addInvite,
```
mike12345567, line 76: could we compress the invitation options into a single object? Right now its not very obvious which options apply to the invitation.

packages/worker/src/api/controllers/global/email.ts
```diff
@@ -14,6 +14,12 @@ export async function sendEmail(ctx: BBContext) {
     cc,
     bcc,
     automation,
+    addInvite,
```
mike12345567, line 17: Same as other point, would be good if this was an invite structure with all invitation properties within - don't need the `addInvite` property if there is just a:
```
invite: {
  startTime,
  endTime,
  summary,
  location,
  url,
}
```

Can simply detect that the invite structure is present.

packages/worker/src/utilities/email.ts
```diff
@@ -24,6 +25,12 @@ type SendEmailOpts = {
   cc?: boolean
   bcc?: boolean
   automation?: boolean
+  addInvite?: boolean
```
mike12345567, line 28: would be:
```
invite?: {
  startTime: Date
  endTime: Date
  summary: string
  location: string
  url: string
}
```

improves typing as the required properties for a meeting can now properly be required.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
melohagan: I think it's because of the new line `{#if canShowField(key, value)}` which wraps the entire field block, so all the child divs etc got shifted inwards.

mike12345567: Ah that makes sense, good old git freaking out lol

packages/server/src/automations/steps/sendSmtpEmail.ts
melohagan: Question mark removed.

packages/server/src/utilities/workerRequests.ts
melohagan: compressed into invite object

packages/worker/src/utilities/email.ts
```diff
@@ -24,6 +25,13 @@ type SendEmailOpts = {
   cc?: boolean
   bcc?: boolean
   automation?: boolean
+  invite?: {
```
mike12345567, line 28: If we break this out into its own type we could move it into `types` and make it available for the server to use.

packages/server/src/utilities/workerRequests.ts
```diff
@@ -65,15 +65,25 @@ async function checkResponse(
...
+}: {
+  to: string
+  from: string
+  subject: string
+  contents: string
+  cc: string
+  bcc: string
   automation: boolean
-) {
+  invite?: object
```
mike12345567, line 85: As with the comment about moving the email types into `types` - we could use that type definition here rather than `object` - as it provides very little insight.

mike12345567 (approved):
Looks good to me! Just would be good to move the email send types into the `@budibase/types` package so that we can use them in both the worker and server packages.

packages/worker/src/utilities/email.ts
melohagan: EmailInvite type added

mike12345567 (approved):

Issue #10786, opened 2023-06-05
[BUDI-7085] Multiple dev collab: User has to refresh when the last user leaves the app to get design/automation access


shogunpurple:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud - qa env
  * Tenant ID: martin

**Describe the bug**
When a user is collaborating with another in the data section, and the user that currently has design/automation privilege leaves, the other user has to refresh to get access to design/automation.

**To Reproduce**

1. Log in with 2 accounts and access the same app
2. Leave the app with the first account
3. The other user is the only user in the "room" but needs to refresh to get the design/automation access

**Expected behavior**
Should update automatically through the websocket when the first user leaves the room.

**Screenshots**
[https://jam.dev/c/39010384-1309-4dc8-a523-e5f0bbc712ab](https://jam.dev/c/39010384-1309-4dc8-a523-e5f0bbc712ab)

**App Export**
N/A

**Desktop (please complete the following information):**

* OS: OSX Ventura 13.0.1
* Browser Chrome
* Version 113.0.5672.126 (Official Build) (arm64)

**Additional context**

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7085](https://linear.app/budibase/issue/BUDI-7085/multiple-dev-collab-user-has-to-refresh-when-the-last-user-leaves-the)</sub>

aptkingston:
Is this a sev1? This was known and I assumed was acceptable.

aptkingston:
I had written down in my notes that notifying a new user via the websocket would be a nice-to-have to save them refreshing, but I'll get to work on adding this if it's crucial.

shogunpurple:
@aptkingston I believe so - I think people will probably know what to do, but it looks like a bug IMO given when someone joins the room, it's automatic. I'm just not sure why we would omit it, unless there were technical reasons to do so

aptkingston:
We just never scoped any sort of lock transfer. I built in instant lock disconnection, removing the need the for 10 min cooldown we currently have, but then assumed that was enough, because the app list instantly reflects that no one is editing it any more. Transferring ownership of the "primary user" lock without leaving the builder again was just something I didn't prioritise.

[...]

PR #10789, opened 2023-06-05
Transfer lock to next oldest session when disconnecting from the builder

aptkingston:
## Description
This PR adds a new feature to collaboration, so that whenever the "primary" user disconnects (the person who can edit screens and automations), this ability is automatically transferred to the user who has also been editing the same app the longest.

The prevents the need to either refresh your browser or exit the app and enter it again in order to be able to edit screens.

The user who gets control will get a notification letting them know they're able to edit screens and automations now, and their builder will automatically update itself.

This PR also slightly updates the websocket API to ensure that all payloads are wrapped in objects.

Addresses: 
- #10786 

See it in action here:

https://github.com/Budibase/budibase/assets/9075550/063a8a57-4ed2-4612-b2d9-a04a2f4da244


mike12345567 (approved):
LGTM! Actually quite a valuable change, does look like a very clean experience

shogunpurple (approved):
This is awesome, really slick. Thanks for making the update 🙏

Issue #10782, opened 2023-06-05
[BUDI-7083] Can't update branding logo after the first upload


shogunpurple:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: bb

**Describe the bug**

**To Reproduce**

1. go to branding page
2. Upload a logo and save config
3. Click "Save"
4. Upload another logo to update the current
5. Click "Save"
6. New logo not applied.

You can also reproduce this by clearing the first logo, saving, and uploading a new one. Seems like the first logo is kept around forever.

**Expected behavior**
Logos overridden as expected

**Screenshots**

**App Export**
N/A

**Desktop (please complete the following information):**

* OS: OSX Ventura 13.0.1
* Browser Chrome
* Version 113.0.5672.126 (Official Build) (arm64)

**Additional context**
N/A

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7083](https://linear.app/budibase/issue/BUDI-7083/cant-update-branding-logo-after-the-first-upload)</sub>

aptkingston:
Closing as this has been deployed and verified by QA.

PR #10790, opened 2023-06-05
Fix/logo urls

deanhannigan:
## Description

Fixes and refactoring for the Logo Url and Favicon Url branding configuration options.

Addresses: 
- https://github.com/Budibase/budibase/issues/10782


packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -352,6 +360,15 @@ export async function publicSettings(
       config: {
         ...config,
         ...branding,
+        ...(branding.faviconUrl && branding.faviconUrl !== ""
```
Rory-Powell, line 363: Can we extract this to a variable? I think it's a bit hard to read inline

Rory-Powell (approved):

PR #10794, opened 2023-06-06
Fix release pipeline

adrinr:
## Description
Changing the pipeline to check that the tag is in master instead of checking the sha ref, given the workflow changes. Having the release triggered by the tag does not populate the git sha

### DIFF
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -10,11 +10,6 @@ on:
       # Exclude all pre-releases
       - "!v*[0-9]+.[0-9]+.[0-9]+-*"
   workflow_dispatch:
-    inputs:
-      tags:
-        description: "Release tag"
-        required: true
-        type: boolean
 
 env:
   # Posthog token used by ui at build time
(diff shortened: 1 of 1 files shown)


.github/workflows/release-master.yml
```diff
@@ -33,12 +28,13 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Fail if branch is not master
-        if: github.ref != 'refs/heads/master'
+      - name: Fail if tag is not master
         run: |
-          echo "Ref is not master, you must run this job from master."
-          // Change to "exit 1" when merged. Left to 0 to not fail all the pipelines and not to cause noise
-          exit 0
+          if ! git merge-base --is-ancestor ${{ github.sha }} origin/develop; then
```
Rory-Powell, line 33: `develop`?

.github/workflows/release-master.yml
```diff
@@ -10,11 +10,6 @@ on:
       # Exclude all pre-releases
       - "!v*[0-9]+.[0-9]+.[0-9]+-*"
   workflow_dispatch:
```
Rory-Powell, line 12: Could probably remove this, given we only want to run the release upon tag

adrinr: This was added having in mind that maybe we can use it for rollbacks. Not 100% sure how is this handled now. But maybe we can have a different pipeline for it if needed, that does not need to rebuild everything

Rory-Powell: Yeah we can use the existing deploy-cloud job to push out any version we want so I'd say it's safe / desirable to remove the dispatch

Rory-Powell (approved):

PR #10797, opened 2023-06-06
Change release pipelines to rely on tags

adrinr:
## Description
Until now we had the release process relying on branch push hook. Because of this, we were pulling at some point to ensure we got the latest code.
Now these pipelines have been split into 2 steps, and the release will always run from a particular tag. These `git pulls` are unnecessary and causing issues.

Changing pipelines to rely on the tags and secure them

### DIFF
--- a/.github/workflows/deploy-preprod.yml
+++ b/.github/workflows/deploy-preprod.yml
@@ -1,17 +1,30 @@
 name: "deploy-preprod"
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: Budibase release version. For example - 1.0.0
-        required: false
   workflow_call:
 
 jobs:
   deploy-to-legacy-preprod-env:
     runs-on: ubuntu-latest
     steps:
+      - name: Fail if not a tag
+        run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-preprod.yml
```diff
@@ -16,12 +16,7 @@ jobs:
       - name: Get the latest budibase release version
         id: version
         run: |
-          if [ -z "${{ github.event.inputs.version }}" ]; then
-            git pull
-            release_version=$(cat lerna.json | jq -r '.version')
-          else
-            release_version=${{ github.event.inputs.version }}
-          fi
+          release_version=$(cat lerna.json | jq -r '.version')
```
Rory-Powell, line 19: what about rollbacks via the workflow_dispatch?

adrinr: If we use `workflow_dispatch` we will be selecting the tag we want to run on. This tag will be pulling the code that has the version in the `lerna.json`.
So, if you manually trigger the tag `v1.2.3` it will checkout the tagged code, which should have the version `1.2.3` hardcoded in there

Rory-Powell (approved):

PR #10802, opened 2023-06-06
Docker yarn.lock tweaks

adrinr:
## Description
Adding yarn.lock to dockerfiles to fix the `yarn install --frozen-lockfile`


.github/workflows/release-develop.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           node scripts/syncLocalDependencies.js $version
           echo "Syncing yarn workspace"
           yarn
-      - run: yarn build --configuration=production
+      - run: yarn build
```
adrinr, line 56: This is not required anymore and it can be confusing. That was added for @nx/esbuild but not used anymore

Rory-Powell (approved):

Issue #9627, opened 2023-02-09
[BUDI-6523] Exported BudibaseDB rows are not UTF-8 encoded 


aricooperdavis:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists
   - A related issue, #3844, has been closed as completed.

**Hosting**

- Cloud
  - Tenant ID: ari 

**Describe the bug**
When exporting rows from a BudibaseDB table that contain "special" characters, the resulting `export.csv` file is not UTF-8 encoded.

**To Reproduce**
1. Create a new BudibaseDB table
2. Create a new text type column
3. Create a test row containing some "special" characters, e.g. `This is a tést 😎`
4. Export that row to CSV
5. The [export.csv](https://github.com/Budibase/budibase/files/10698709/export.csv) is not UTF-8 encoded

**Expected behavior**
The resultant `export.csv` should be UTF-8 encoded.

**Screenshots**
![budibase](https://user-images.githubusercontent.com/8947634/217857001-ed5fc393-5894-448a-9080-59c1be865802.png)
![notepad++](https://user-images.githubusercontent.com/8947634/217857035-3267592f-a6b8-48c3-a018-0d6f017b22c2.png)

**App Export**
n/a

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise 19045.2486
 - Browser: Firefox
 - Version: 109.0.1 (64-bit)

memarius:
The following workaround has worked for me and might work for those experiencing this issue aswell:

I believe the current export is encoded with the `windows-1252` encoding. After opening the `export.csv` in VSCode, you can reopen it with a different encoding:

1. Klick on the encoding at the bottom
![image](https://user-images.githubusercontent.com/14908727/227448902-7c3a2a9a-e59c-49f5-a9c3-25e594c7ce02.png)

2. Select `Reopen with encoding`
![image](https://user-images.githubusercontent.com/14908727/227448949-34a872ba-6f56-4245-a239-268e3fbe7151.png)

3. Select `Windows 1252`
![image](https://user-images.githubusercontent.com/14908727/227449058-7ea46b86-707a-495a-9d6f-0cb8854b1186.png)

Then the special chars should be displayed correctly and you can use the same process, just selecting `Save with Encoding` and then using `UTF-8` to save the file in UTF-8.

It obviously doesn't fix the bug but makes it possible to work with the current export.
Thanks in advance to whoever will fix it

mike12345567:
[...]

PR #10803, opened 2023-06-06
Fix CSV encoding issue

gitstart:
## Description
When exporting rows from a BudibaseDB table that contain “special” characters, the resulting export.csv file is not UTF-8 encoded.

Addresses: 
- https://github.com/Budibase/budibase/issues/9627

## App Export

## Screenshots
No UI changes here

## Documentation
- [x] I have reviewed the budibase documentation to verify if this feature requires any changes. If changes or new docs are required, I have written them.

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/ExportModal.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/ExportModal.svelte
@@ -130,7 +130,10 @@
         notifications.error("Export Failed")
       }
       if (response) {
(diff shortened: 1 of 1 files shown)


gitstart:
The issue is with the `downloadjs` package, it seems not to handle “special” characters well in plain text even for other file types other than CSV. Wrapping the text in a `Blob` however seems to fix the issue.

aptkingston (approved):
Looks good to me. Thanks @gitstart!

PR #10814, opened 2023-06-07
Revert single image back to node 14

adrinr:
## Description
Reverting the single image as node16 is causing issues

### DIFF
--- a/hosting/single/Dockerfile
+++ b/hosting/single/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16-slim as build
+FROM node:14-slim as build
 
 # install node-gyp dependencies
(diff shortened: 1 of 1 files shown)


mike12345567 (approved):

hosting/single/Dockerfile
```diff
@@ -12,13 +12,15 @@ RUN chmod +x /cleanup.sh
 WORKDIR /app
 ADD packages/server .
 COPY yarn.lock .
-RUN yarn install --production=true && /cleanup.sh
+RUN yarn install --production=true
```
adrinr, line 15: Just splitting it for more verbosity on the build. This was failing and it was hard to see if it was because the yarn install or the cleanup. It should have no impact on the performance, as it was already 2 different steps

PR #10819, opened 2023-06-07
Display indeterminate when few checkbox selected

adrinr:
## Description
Changing UI to support indeterminate checkboxes on checkbox group

### All selected
image

### Indeterminate
image

### None selected
image

### DIFF
--- a/packages/bbui/src/Form/Core/Checkbox.svelte
+++ b/packages/bbui/src/Form/Core/Checkbox.svelte
@@ -20,11 +20,10 @@
 </script>
 
 <label
-  class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass} {indeterminate
-    ? 'is-indeterminate'
-    : ''}"
+  class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass}"
   class:is-invalid={!!error}
   class:checked={value}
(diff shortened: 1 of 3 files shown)


packages/bbui/src/Form/Core/Checkbox.svelte
```diff
@@ -19,7 +20,9 @@
 </script>
 
 <label
-  class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass}"
+  class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass} {indeterminate
+    ? 'is-indeterminate'
+    : ''}"
```
aptkingston, line 25: Not a blocker, but just letting you know that svelte has a shorthand way of applying classes - you can simply do:
```html
class:is-indeterminate={indeterminate}
```
And you can have multiple class tags, so this can simply be:
```html
...
class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass}"
class:is-indeterminate={indeterminate}
...
```

adrinr: Oh! That's handy :)

mike12345567 (approved):
Good improvement, very nice :)

PR #10822, opened 2023-06-07
Fix single image top level path

mike12345567:
## Description

By default in our app/worker images the top level path generally will resolve to just `/` - the mechanism we use for this generally works fine, however within the single image it is better to just hardcode the top level path to `/app` for the server to avoid any issues.


adrinr (approved):
Is this running fine locally?

adrinr (approved):

PR #10837, opened 2023-06-08
Relationship cell fix

mike12345567:
## Description
When testing a customer app export ran into an issue with this, quick fix to make sure the each is always iterating over an array.


aptkingston:
Nice, yea I'll take a look at it but I'm happy to merge this in the meantime!

aptkingston (approved):

mike12345567:
Closing PR as @aptkingston has looked into this and found the deeper root cause.

PR #10844, opened 2023-06-09
Preview apps inside builder

aptkingston:
## Description
This PR updates how apps are previewed, so that they open in a modal inside the builder rather than a new tab. This lets you quickly preview your app without losing your current place.

image

Also fixed a few longstanding issues:
- The blue preview header would disappear in your other tab if you reloaded the builder. This was due to sharing local storage context between tabs.
- Changing your role inside the preview tab would affect the preview inside the builder as well

Clarification of functionality:
- The preview will open at whatever screen you have selected in the design section
- The preview can be opened from anywhere in the builder
- The preview includes a theme-aware skeleton loader to increase how fast loading feels
- Nice animations when loading, and background blur on the parent window
- The "old" preview still works - you can still manually enter the preview URL and it will function as before, so this is not breaking in any way
- The preview can be closed by clicking outside it (in the blurred area) or by pressing the dedicated exit preview button

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -74,6 +74,7 @@ const INITIAL_FRONTEND_STATE = {
   propertyFocus: null,
   builderSidePanel: false,
   hasLock: true,
+  showPreview: false,
 
   // URL params
   selectedScreenId: null,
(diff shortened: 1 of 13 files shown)


melohagan:
Just checking - do the navigation links along the top work in the preview modal?

shogunpurple (approved):
LGTM! Such a better experience

melohagan (approved):

PR #10850, opened 2023-06-09
Datasource plus onboarding V2

mike12345567:
## Description
This PR covers the work to address the improved onboarding experience for datasource plus integrations. Specifically this includes work to improve the experience of using SQL and Googlesheet datasources.

This makes use of the API built earlier by firestorm to get information about the external datasource being connected to. This PR is simply about presenting that information (with a few minor changes to the API to best suit the designed flows).

The efforts within this PR are combined efforts of @adrinr and @mike12345567 - implementing the two different sides Googlesheets and SQL.

This also adds a new useful component to BBUI, the `FancyCheckboxGroup` which is used for selecting from a list of options.

# Screenshots
Connecting to a Postgres database, listing the tables found within
image

Fetching tables from within a SQL datasource
image

Connecting to a Googlesheet, showing the sheets within
image

Fetching sheets from existing
image

Addresses: 
- https://linear.app/budibase/issue/BUDI-6935/[blocked]-data-source-fetch-tables-changes
- https://linear.app/budibase/issue/BUDI-6934/data-source-fetch-tables-changes
- https://linear.app/budibase/issue/BUDI-6945/google-sheets-fetch-all-worksheets-when-we-have-spreadsheet-url


packages/bbui/src/FancyForm/FancyField.svelte
```diff
@@ -61,7 +65,7 @@
 
 <style>
   .fancy-field {
-    max-width: 400px;
+    max-width: var(--fancy-field-max-width);
```
adrinr, line 68: Should we keep the 400x as default?

packages/bbui/src/FancyForm/FancyField.svelte
```diff
@@ -91,8 +101,8 @@
   }
   .content {
     position: relative;
-    height: 64px;
-    padding: 0 16px;
+    height: var(--fancy-field-height);
```
adrinr, line 104: Same here, this can break any current usage

packages/bbui/src/FancyForm/FancyForm.svelte
```diff
@@ -22,9 +25,17 @@
     })
     return valid
   }
+
+  const styles = () => {
```
adrinr, line 29: Why is this a function? Could this not be an object?

packages/builder/src/components/backend/DatasourceNavigator/icons/index.js
```diff
@@ -44,6 +44,9 @@ export default ICONS
 
 export function getIcon(integrationType, schema) {
   const integrationList = get(integrations)
+  if (!integrationList) {
```
adrinr, line 47: Why is this supported? Should the integration not be required?

packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
```diff
@@ -15,11 +22,25 @@
   // kill the reference so the input isn't saved
   let datasource = cloneDeep(integration)
   let isValid = false
+  let fetchTableStep = false
+  let selectedTables = []
+  let tableList = []
 
+  $: console.log(selectedTables)
```
adrinr, line 29: Log 🔊

packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
```diff
@@ -15,11 +22,25 @@
...
+  $: datasourcePlus = datasource?.plus
+  $: title = fetchTableStep ? "Fetch your tables" : `Connect to ${name}`
+  $: confirmText = fetchTableStep
+    ? "Continue"
+    : datasourcePlus
+    ? "Connect"
+    : "Save and continue to query"
 
   async function validateConfig() {
+    if (!integration.features[DatasourceFeature.CONNECTION_CHECKING]) {
```
adrinr, line 41: integration.features undefined will break the js (as the live bug yesterday)

packages/builder/src/stores/backend/datasources.js
```diff
@@ -57,14 +57,18 @@ export function createDatasourcesStore() {
     return updateDatasource(response)
   }
 
-  const save = async (body, fetchSchema = false) => {
+  const save = async (body, { fetchSchema, tablesFilter } = {}) => {
+    if (fetchSchema == null) {
```
adrinr, line 61: When will this be null? This means that we actively marked it as null?

packages/bbui/src/FancyForm/FancyField.svelte
mike12345567: it is default - the variable in the background defaults to 400px

packages/bbui/src/FancyForm/FancyForm.svelte
mike12345567: it was just for ease of readability - but it could probably be an array thats just joined, just didn't want to create a massive string all in one go thats hard to understand

packages/builder/src/components/backend/DatasourceNavigator/modals/DatasourceConfigModal.svelte
mike12345567: ah yes - good catch!

packages/builder/src/stores/backend/datasources.js
mike12345567: it will be null/undefined if the option isn't passed in, wanted to maintain similar parity to the old way where everything was totally optional. At the minute if you were to pass `{ tablesFilter: [] }` then `fetchSchema` would be null. This is a bit of a weird scenario, but wanted to maintain parity with `fetchSchema` always being either `true` or `false`.

packages/builder/src/components/backend/DatasourceNavigator/icons/index.js
mike12345567: I ran into an issue with the `integrationList` not being loaded initially when this was called and bricking the app due to failing to retrieve anything, just wanted to avoid this scenario going forward!

packages/bbui/src/FancyForm/FancyField.svelte
mike12345567: Also has a default option set in the variable

aptkingston (approved):
LGTM!

Issue #10838, opened 2023-06-08
Unable to Open Binding Editor Outside the Side Panel on Design Tab


MarleTangible:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker single image
  - Budibase Version: v2.7.7 also repeatable on budibase.app online.
  - App Version: v2.7.7

**Describe the bug**
When adding a filter to a table block, clicking on the thunderbolt symbol fails to show the binding editor. This issue only present itself when the bottom pane is used. Buttons on the right pane works as expected.

**To Reproduce**
1. Select any table block and click on define filters on right pane. 
2. Filtering panel will be shown at the bottom.
3. Click on the add filter button.
4. Select a column and change the value to binding.
5. Click on the thunderbolt symbol to open the binding editor.
6. Interaction fails and bottom panel navigates back to the main filtering page.
7. The issue can be repeated for any thunderbolt button presented on the bottom pane. 

**Expected behavior**
Binding editor to be shown.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/107574310/2890741a-31c8-4bb2-8dd1-713112d167b9)
  
**App Export**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10 21H2
 - Browser: Chrome and Edge
 - Version: Both 114

**Additional context**
N/A

aptkingston:
Are you sure this one was fixed @shogunpurple? It predates the binding changes and Dean was looking into it.

shogunpurple:
Was chatting to dean there on DM - It happened to me every time before the release, then I couldn't repro after the latest release went out. But it seems like it may actually be intermittent. Will re-open

itopaloglu83:
This looks like the same issue as well.
https://github.com/Budibase/budibase/issues/10851

shogunpurple:
Should now be resolved!

PR #10853, opened 2023-06-09
Filtering Binding Drawer Fix 

deanhannigan:
## Description

Temporarily removing the binding drawer focus behaviour as it was triggering a redraw of the component settings components.

Addresses: 
- https://github.com/Budibase/budibase/issues/10838


shogunpurple (commented):
Best to remove the commented code if we won't be using it. Otherwise LGTM!

aptkingston (approved):

PR #10868, opened 2023-06-12
Use Modal For Editing Datasources

Ghrehh:
## Description
- Consumes the changes made in #10783 to edit datasource config in a modal.
- Adds some selectors for getting data out of stores, will explain underneath the code itself.
- The REST config button looks a little different, I ran this by product and they're okay with it.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6936/change-configuration-display-in-data-source-details

|explaination|screenshots|
|---|---|
|Couch|Screenshot 2023-06-12 at 11 37 53|
|Postgres|Screenshot 2023-06-12 at 11 37 42|
|Sheets|Screenshot 2023-06-12 at 11 37 33|
|Rest|Screenshot 2023-06-12 at 11 37 19|


packages/builder/src/stores/selectors.js
```diff
@@ -0,0 +1,19 @@
+import { IntegrationTypes, DEFAULT_BB_DATASOURCE_ID } from "constants/backend"
+
+export const integrationForDatasource = (integrations, datasource) => ({
```
Ghrehh, line 3: Something like this is pretty trivial to implement in the component itself, but it arguably leaks a fair amount of implementation detail, and some code relies on an integration having a `name` field, which is taken from its key in the object structure it's returned from the backend in. This'd be easy to leave out and potentially introduce bugs, so to me it makes sense to have a helper method to do this.

The most obvious method is adding it to the store itself but the problem is that this method isn't reactive to changes in integrations.

```js
integrations.integrationForDatasource(datasource)
``` 

Another method is adding it to the store data itself, this does work, but seems a little strange to me personally, to have helper methods both on the store itself, and on the store's data.

``` js
$integrations.datasourceForIntegration(datasource)
```

What does everyone think? Happy to implement this some other way if folks dislike it.

Another method is to create a function, get given some sort of parameter, returns a store with the desired data, like this example of a store that returns a single datasource given an id.

``` js
import { datasources } from "./datasources"

// Returns a store that is just a wrapper around the base datasources store
export const createDatasourceStore = (id) => {
  return {
    subscribe: (subscriptionFunction) => {
      const datasourcesSubscriptionFunction = (datasources) => {
        const datasource = datasources.list.find(datasource => datasource._id === id)
        subscriptionFunction(datasource)
      }

      const unsubscribe = datasources.subscribe(datasourcesSubscriptionFunction)

      return unsubscribe
    }
  }
}
```

The issue is that this needs to be consumed like so. which is maybe a little verbose:

``` js
let datasource = createDatasourceStore(id)
$: console.log($datasource)

\\ THIS DOESN'T WORK

$: console.log($createDatasourceStore(id))
```

Which leads me to believe the selector method is best, they're easy to write tests for and must be provided stores that user can automatically subscribe to changes from like so:

``` js
console.log(integrationForDatasource($integrations, $datasources.list[0]))
```

Ghrehh:
Going to close this for now, as the base of this has changed considerably, and on second thoughts I'm just going to pull parts of this into the base anyway.

PR #10869, opened 2023-06-12
Added value prop to the schema deprecation notifications

deanhannigan:
## Description

When checking the schema deprecation status for fields in automations, no value is being passed to `canShowField` when processing string values. In this scenario an exception is thrown which crashes the builder.

- https://github.com/Budibase/budibase/pull/10787


Ghrehh:
Can we leave key out of `canShowField`'s arguments? It doesn't seem to be used, quite possible I'm missing something very obvious though lol

deanhannigan:
Fixed here: https://github.com/Budibase/budibase/pull/10948

PR #10874, opened 2023-06-12
BUDI-7010 - Encrypt app exports

adrinr:
## Description
Allowing encrypting exported app files during both import and export. This is optional on the API, so old exports will keep working as they work now.
The initial approach was to change the tarball to something such as 7z or zip file, but it required much additional work without adding much benefit. Instead, we will only encrypt the files, so we can reuse the current functionality

Addresses: 
- https://linear.app/budibase/issue/BUDI-7010/encrypt-app-exports


mike12345567 (approved):
LGTM - good to have those improvements to the encryption library like being able to control the IV length!

Issue #10438, opened 2023-04-27
[BUDI-7026] When custom role doesn't exist in production app - assigned users should be treated as public


galenvssysadmin:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud

**Describe the bug**
Whenever I create a custom role within an app, then assign it to a particular user, that user's access to the app is broken.

**To Reproduce**
1. Edit any app, and create a custom role. In this case, this is a custom 'Super' role I created:
<img width="395" alt="Screen Shot 2023-03-02 at 06 54 18" src="https://user-images.githubusercontent.com/102808173/222421701-0ffb5339-db38-4e03-a9df-a6f5a5069f91.png">

2. Assign this role to a user:
<img width="849" alt="Screen Shot 2023-03-02 at 06 56 11" src="https://user-images.githubusercontent.com/102808173/222422223-03df19b5-015d-4e67-9d86-8a7332845bfc.png">

3. When this user tries to access the target app, they see this:
![image](https://user-images.githubusercontent.com/102808173/234920103-1b2d2f7f-7dc0-4a51-83ec-ece9a33a7ccf.png)

If I revert this user's access back to any of the default roles (i.e. Basic, Power, Admin), their access is restored.

<sub>[BUDI-7026](https://linear.app/budibase/issue/BUDI-7026/creating-and-assigning-in-app-custom-roles-breaks-access-to-app)</sub>

melohagan:
The custom role does seem to be fine if you make sure to publish the app!

It's only when you don't publish the app after creating the new role that I was able to reproduce.

---

I also noticed another minor UI bug:

1. Assign a user to an app with a custom role
2. Try to delete the custom role -> see error `Error deleting role`
3. Change the access of the user from the custom role to 'Admin'. You can now delete the custom role.

The error message should be updated to explicitly state that the custom role is currently in use, or it should automatically assign the existing users to the inherited role.

I.e. if I have SuperAdmin extends Admin, then all my SuperAdmins should become Admins when I delete the custom role.

PR #10881, opened 2023-06-13
Custom roles breaking app access when not published

mike12345567:
## Description
Fix for custom roles that have not been published causing users to be unable to access an app completely. They should instead be treated as public users as their role isn't valid.

Addresses: 
- https://github.com/Budibase/budibase/issues/10438
- https://linear.app/budibase/issue/BUDI-7026/when-custom-role-doesnt-exist-in-production-app-assigned-users-should


PClmnt (approved):
LGTM!

adrinr (approved):

PR #10894, opened 2023-06-13
BUDI-7010 - Frontend decrypt app imports

adrinr:
## Description
Now that we support exporting encrypted apps, we need to support importing them.
This PR is detecting when an import is encrypted (based on its extension) and asks for the decryption password

Addresses: 
- [BUDI-7010](https://linear.app/budibase/issue/BUDI-7010/encrypt-app-exports)

#### Step 1
image

#### Step 2 (only when app is encrypted)
image

##### Wrong password
image


packages/server/src/api/controllers/application.ts
```diff
@@ -114,7 +114,18 @@ function checkAppName(
   }
 }
 
-async function createInstance(appId: string, template: any) {
+interface AppTemplate {
```
mike12345567, line 117: Nice to get this typed!

mike12345567 (approved):
LGTM!

PClmnt (approved):

Issue #10834, opened 2023-06-08
[BUDI-7099] MySQL - Converting circular structure to JSON


diogenesbrussels:
**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: From 2.5.9 to 2.7.6
  - App Version: From 2.5.9 to 2.7.6

**Describe the bug**
After the update, data requests via the API/table/search on certain tables generate the following error:
Converting circular structure to JSON

**To Reproduce**
1. Have a MySQL table with relationships (here a person have multiple actions rows linked)
2. Open the data tab of this table or use it in a data provider
3. The table/ data provider is empty (error 500)
4. Converting circular structure to JSON --> starting at object with constructor 'Object' | property 'XXXXXXX' -> object with constructor 'Array' | index 0 -> object with constructor 'Object' | property 'primaryDisplay' -> object with constructor 'Array' --- index 0 closes the circle

**Expected behavior**
The data must be sent as before.

**Additional context**
I had to change all the data provider for this table to sql queries to resolve this.
Has the way relationships are resolved changed? Is there a list of breaking changes between releases? The last 3 updates I made created errors in an existing application without any problems. I'd like to keep an up-to-date version as much as(for example, some components display a missing order by column error, even though no sort order has been defined, but if I select one it works again) possible to take advantage of your great work, but this type of problem makes me wonder about this for the future. (for example, some components display a missing order by column error, even though no sort order has been defined previously, but if I select one it works again :/) Thank you!

<sub>[BUDI-7099](https://linear.app/budibase/issue/BUDI-7099/converting-circular-structure-to-json)</sub>

mike12345567:
Hi @diogenesbrussels - I've given this a test, using a MySQL database, with a `Person` and `Tasks` table - configuring the tasks with a `PersonID` foreign key then setting up this relationship with Budibase.

The data is loaded fine as shown below:
![image](https://github.com/Budibase/budibase/assets/4407001/5230de47-58d0-4ccc-9249-f73aea8ce1bc)

I'm wondering if you could provide a bit more information here, such as the schema of two tables or some screenshots of the error as you see it? If there is a network request that fails it would also be good to see a screenshot of the failing request to help with debugging.

diogenesbrussels:
Hey @mike12345567 

Here is the log when opening one of the table in the Data section:
```
[...]

PR #10895, opened 2023-06-13
MySQL cyclic JSON conversion issue with primary display columns

mike12345567:
## Description
Fix for circular issue with primary display fields on SQL tables introduced in most recent update - if somehow the primary display field is set to a relationship field there was a chance of cyclic structure occurring which Koa could not convert to JSON.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7101/custom-mysql-update-with-date-like-value-saves-as-date-instead-of-text
- https://github.com/Budibase/budibase/issues/10834

### DIFF
--- a/packages/server/src/integrations/utils.ts
+++ b/packages/server/src/integrations/utils.ts
@@ -328,3 +328,22 @@ export function finaliseExternalTables(
     .reduce((r, [k, v]) => ({ ...r, [k]: v }), {})
   return { tables: finalTables, errors }
 }
+
+/**
+ * Checks if the provided input is an object, but specifically not a date type object.
+ * Used during coercion of types and relationship handling, dates are considered valid
+ * and can be used as a display field, but objects and arrays cannot.
+ * @param test
+ */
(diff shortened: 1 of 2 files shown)


PClmnt (approved):

packages/server/src/integrations/utils.ts
```diff
@@ -328,3 +328,22 @@ export function finaliseExternalTables(
...
   return { tables: finalTables, errors }
 }
+
+/**
+ * Checks if the provided input is an object, but specifically not a date type object.
+ * Used during coercion of types and relationship handling, dates are considered valid
+ * and can be used as a display field, but objects and arrays cannot.
+ * @param test
+ */
+export function getPrimaryDisplay(test: any): string | undefined {
```
adrinr, line 338: This being in the utils it looks like it's going to be reused. But with the description/name/params it's not really clear what is it used. Could we make the function name and params more verbose?

mike12345567: yeap - will do!

PR #10900, opened 2023-06-14
BUDI-7010 - Export controller as post

adrinr:
## Description
Refactoring the export functionality to work via post, so we can extend it to accept password values in the body, making it safer

Addresses: 
- [BUDI-7010](https://linear.app/budibase/issue/BUDI-7010/encrypt-app-exports)


mike12345567 (approved):
Looks good to me - quite a pain to have to do this, but don't think theres a better way!

Issue #3164, opened 2021-10-26
Support for Offline / Air-gapped deployments


mjashanks:
Discussed in https://github.com/Budibase/budibase/discussions/3162 - please read for full detail

Budibase is currently not working on an Air-Gapped deployment.

mjashanks:
Noting in here that a user finds this to be an issue with slow internet connections as well: https://github.com/Budibase/budibase/discussions/4421

pandel:
Hi!

To circumvent the offline problem, is there any possibility to extend the docker-compose file in a way, that it provides Budibase with a corporate HTTP(S) proxy, so that it can connect to the Internet if run inside a protected LAN environment?

I already tried HTTP(S)_PROXY env vars, but they didn't work...

Regards, Holger

PR #10905, opened 2023-06-14
removing need for external internet assets and self hosting our own f…

shogunpurple:
Addresses:
- https://github.com/Budibase/budibase/issues/3164

This PR strips all reliance on internet based services for our fonts so budibase runs in an airgapped environment without issues (apart from templates, which is a separate piece of work).  It woas actually a reasonably simple change, installing the relevant NPM modules for the font assets and a small vite plugin to copy them over to the `builder` output directory so they could be served.


packages/builder/src/pages/builder/Branding.svelte
```diff
@@ -8,7 +8,7 @@
   $: platformTitle =
     !$auth.user && platformTitleText ? platformTitleText : "Budibase"
 
-  $: faviconUrl = $organisation.faviconUrl || "https://i.imgur.com/Xhdt1YP.png"
+  $: faviconUrl = $organisation.faviconUrl || "/builder/bblogoimgur.png"
```
mike12345567, line 11: Should we remove the `imgur` from this - since we don't need to worry about imgur anymore, might be better to just reference any reference to the fact we ever used it.

shogunpurple: Yeah, it's kind of redundant - mainly just used it to keep myself right when I was replacing the imgur URLs. Will update!

aptkingston (approved):
LGTM! Although should we maybe use the black logo rather than the grey one? I know we use the grey one at the moment, but I've been wanting to swap that out, and this seems like a nice time to do it.

PR #10918, opened 2023-06-15
Budi 5262 - Support pg client cert

adrinr:
## Description
Support Postgres client certificates. Docs: [34.19.2. Client Certificates](https://www.postgresql.org/docs/current/libpq-ssl.html)

Also tweaked the save button to show the loading state while saving:
image

Addresses: 
- [BUDI-5262
](https://linear.app/budibase/issue/BUDI-5262/not-able-to-connect-to-gcp-postgres-ssl-client-cert-client-key-server)

## Screenshots
Having a PG instance with client certificates set up...
### Not setting up SSL
image

### Enabling SSL but missing certificates
image

### All configured
image


mike12345567 (approved):
Nice addition of the spinner here, much better UX!

Issue #10037, opened 2023-03-15
[BUDI-6753] Builder error toast spam


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.4.30
  * App Version: 2.4.30

**Describe the problem**
If a datasource has been deleted, or is configured incorrectly, the builder in the design section will spam you error messages, particularly if there's many Data providers, repeaters, etc.

Dean had the idea of collapsing these with an error count; something like `(11) deleted`

**Screenshots**

даже

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6753](https://linear.app/budibase/issue/BUDI-6753/builder-error-toast-spam)</sub>

PR #10920, opened 2023-06-15
Add count to duplicate errors

melohagan:
## Description
When using repeaters or many components there could be scenarios where a user would be spammed with error toasts.
Duplicate messages are now compressed into one with a count in brackets.

Addresses: 
- https://github.com/Budibase/budibase/issues/10037

## Screenshots

setup_error_spam

result


aptkingston (approved):
LGTM!

PR #10922, opened 2023-06-15
Fix public apps still displaying free footer when licensed

mike12345567:
## Description
Adding a license mechanism to the app package call so that public apps can still return what license the app utilises. This is retrieved based on the tenant of the app - we only attach the license type so as to not expose further information than is necessary.

This now means that generally the client library will determine its licensing from the app package, rather than from the user, as this is always going to be available, while the user may not always be found.

### DIFF
--- a/packages/server/src/api/controllers/application.ts
+++ b/packages/server/src/api/controllers/application.ts
@@ -1,53 +1,54 @@
 import env from "../../environment"
 import {
+  createAllSearchIndex,
   createLinkView,
   createRoutingView,
-  createAllSearchIndex,
 } from "../../db/views/staticViews"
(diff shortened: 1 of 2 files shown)


shogunpurple (approved):
Nice solution to this - LGTM!

packages/server/src/api/controllers/application.ts
```diff
@@ -207,6 +208,7 @@ export async function fetchAppPackage(ctx: UserCtx) {
   let application = await db.get(DocumentType.APP_METADATA)
   const layouts = await getLayouts()
   let screens = await getScreens()
+  const license = await licensing.getLicense()
```
Rory-Powell, line 211: This will actually call to account portal every time, we should use `getCachedLicense`

PR #10923, opened 2023-06-15
Quick fix to app package licensing fetch

mike12345567:
## Description
Quick fix suggested by @Rory-Powell to make use of the cached license when fetching the app package.

### DIFF
--- a/packages/server/src/api/controllers/application.ts
+++ b/packages/server/src/api/controllers/application.ts
@@ -208,7 +208,7 @@ export async function fetchAppPackage(ctx: UserCtx) {
   let application = await db.get(DocumentType.APP_METADATA)
   const layouts = await getLayouts()
   let screens = await getScreens()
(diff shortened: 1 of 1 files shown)


packages/server/src/api/controllers/application.ts
```diff
@@ -208,7 +208,7 @@ export async function fetchAppPackage(ctx: UserCtx) {
   let application = await db.get(DocumentType.APP_METADATA)
   const layouts = await getLayouts()
   let screens = await getScreens()
-  const license = await licensing.getLicense()
+  const license = await licensing.getCachedLicense()
```
Rory-Powell, line 211: Seems like we should have a test here that mocks `licensing` and asserts that the relevant `plan.type` (or free when there is no license) has been returned in the response.

Rory-Powell (approved):

PR #10936, opened 2023-06-16
Handle esbuild dependencies

adrinr:
## Description
Changing esbuild configuration to embed node module dependencies for packages. This will prevent issues with different version between apps and packages in the monorepo, and it will free us for pulling the npm versions of our own packages to ensure we have all the required subdependencies.

The packages will come with the required dependencies, but will exclude the built package one. These will come defined in the package.json, so they are not really required (`esbuild-node-externals` takes care of it)

<table>
    <thead>
        <tr>
            <th>Before</th>
            <th>After</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td colspan=2>Worker bundle</td>
        </tr>
        <tr>
            <td>
image
</td>
            <td>image</td>
        </tr>
        <tr>
            <td colspan=2>Server bundle</td>
        </tr>
        <tr>
            <td>
image
</td>
            <td>
image
</td>
        </tr>
    </tbody>
</table>


package.json
```diff
@@ -48,9 +48,9 @@
     "kill-builder": "kill-port 3000",
     "kill-server": "kill-port 4001 4002",
     "kill-all": "yarn run kill-builder && yarn run kill-server",
-    "dev": "yarn run kill-all  && lerna run --stream --parallel dev:builder --stream",
-    "dev:noserver": "yarn run kill-builder  && lerna run --stream dev:stack:up && lerna run --stream --parallel dev:builder --ignore @budibase/backend-core --ignore @budibase/server --ignore @budibase/worker",
-    "dev:server": "yarn run kill-server && lerna run --stream --parallel dev:builder --scope @budibase/worker --scope @budibase/server",
+    "dev": "yarn run kill-all && lerna run --stream --parallel dev:builder --stream",
+    "dev:noserver": "yarn run kill-builder && lerna run --stream dev:stack:up && lerna run --stream --parallel dev:builder --ignore @budibase/backend-core --ignore @budibase/server --ignore @budibase/worker",
+    "dev:server": "yarn run kill-server && yarn build --projects=@budibase/client && lerna run --stream --parallel dev:builder --scope @budibase/worker --scope @budibase/server",
```
adrinr, line 53: Client is required by the server, but because we are still using lerna to run this command, the `nx depends on` does not work

packages/server/package.json
```diff
@@ -63,6 +63,7 @@
     "airtable": "0.10.1",
     "arangojs": "7.2.0",
     "aws-sdk": "2.1030.0",
+    "bcrypt": "5.1.0",
```
adrinr, line 66: This needs to be excluded from the esbuild, so it needs to be added as external on the esbuild process. Because of this, we need to add it in the final app to ensure the dependency its pulled

scripts/build.js
```diff
@@ -25,32 +25,28 @@ function runBuild(entry, outfile) {
...
-      }),
-    ],
+    plugins: [TsconfigPathsPlugin({ tsconfig }), nodeExternalsPlugin()],
     target: "node14",
     preserveSymlinks: true,
     loader: {
       ".svelte": "copy",
     },
+    metafile: true,
+    external: [
```
mike12345567, line 35: How do we keep this list maintained, and what makes a package external rather than part of the service?

mike12345567 (approved):
Looks good - just one question on the external list!

PR #10953, opened 2023-06-19
BUDI-7115 - Sql schema export

adrinr:
## Description
Adding endpoint to export sql schemas for postgres, myssql and mssql databases

Addresses: 
- [BUDI-7115](https://linear.app/budibase/issue/BUDI-7115/sql-schema-export)

### DIFF
--- a/packages/builder/src/components/start/ExportAppModal.svelte
+++ b/packages/builder/src/components/start/ExportAppModal.svelte
@@ -16,7 +16,7 @@
 
   let password = null
   const validation = createValidationStore()
-  validation.addValidatorType("password", "password", true)
+  validation.addValidatorType("password", "password", true, { minLength: 8 })
   $: validation.observe("password", password)
 
   const Step = { CONFIG: "config", SET_PASSWORD: "set_password" }
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -26,7 +26,6 @@ import {
   IntegrationBase,
   DatasourcePlus,
   SourceName,
-  Ctx,
 } from "@budibase/types"
 import sdk from "../../sdk"
 import { builderSocket } from "../../websockets"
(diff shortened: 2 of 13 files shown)


packages/builder/src/components/start/ExportAppModal.svelte
```diff
@@ -16,7 +16,7 @@
 
   let password = null
   const validation = createValidationStore()
-  validation.addValidatorType("password", "password", true)
+  validation.addValidatorType("password", "password", true, { minLength: 8 })
```
mike12345567, line 19: do we need a min length?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -441,3 +442,18 @@ export async function query(ctx: UserCtx) {
     ctx.throw(400, err)
   }
 }
+
+export async function getExternalSchema(ctx: Ctx) {
```
mike12345567, line 446: This should probably be `UserCtx` rather than `Ctx` as there will be a user available

packages/server/src/api/routes/datasource.ts
```diff
@@ -66,5 +66,10 @@ router
     authorized(permissions.BUILDER),
     datasourceController.destroy
   )
+  .get(
+    "/api/datasources/:datasourceId/external-schema",
```
mike12345567, line 70: not a fan of a dash in the URL - I think in this case it would be fine if it were just `schema`, I don't think we need to mark it as external as the whole datasources API is external (the internal tables never use the datasources API).

mike12345567 (approved):
Looks amazing - great work getting the tests together for this as well!

Only change I think worth making is the URL - we don't need `external` in it as datasources simply are external.

packages/builder/src/components/start/ExportAppModal.svelte
adrinr: This min length was actually added for the encrypted exports. It was no other referecne to this validator type, so I assumed that it was not used. But it's used dynamically based on datasource schemas, so my change was adding a min length requirement to all the passwords. I only wanted it for the exporting encryption

packages/server/src/api/routes/datasource.ts
adrinr: This can be confused by the internal schema. In fact we have a `POST /schema` that sets the internal schema. Having a `POST` that sets the internal a schema and the same `GET` that gets the external schema can be confusing.
I don't really like the hyphen either, but I think it's what we use in other urls.
Another alternative is to have `/api/datasources/:datasourceId/schema/external`. Thoughts?

mike12345567: Sure - `schema/external` works as well!

PR #10960, opened 2023-06-19
Remove the doble deep format references

adrinr:
## Description
Changing the way we consume the pro submodule. Until now the pro submodule was a lerna monorepo with a single package. This added unnecessary complexity that has been removed by keeping the pro repo as a flat package.

In this PR we also change the way we check the validity of the submodule. Until now the process  was checking that the current submodule pointer was actually merged in the main branch. This did not mean that we are actually pointing to the latest commit, or even that this commit was part of the git history. The PR checks now get sure that we are always pointing to the head of develop/master, avoiding any possible crossover.

Example of the current possible problem:
1. Develop is on commit A
2. We branch out feature 1 from the commit A
3. We branch out feature 2 from the commit A
4. We develop feature 1 branch, adding the commits: 1a, 1b and 1c
4. We develop feature 2 branch, adding the commits: 2a and 2b
5. We merge feature 1 to develop, the git history of develop is: A, 1a, 1b, 1c, A2 (being the merge commit). On the pipeline we check that the commit 1c has been merged into develop, all the checks are green
6. We then merge feature 2 to develop. Develop git's history: A, 1a, 1b, 1c, A2, 2a, 2b, A3 (merge commit). Again, the pipelines are green.

The issue is that our monorepo might be actually pointing to the A2 commit but being in a wrong git sub-branch, missing the commits 1a, 1b and 1c


mike12345567 (approved):
Looks good to me - would be worth filling out the explanation in this PR of what it is doing - might also be worth updating our docs on confluence around this so we can send out a message into the engineering channel; simply for those that will be affected by this so they are aware of what they need to do when merging pro changes!

PR #10962, opened 2023-06-19
Grid block

aptkingston:
## Description
This PR adds a grid block for client apps. There are also a lot of improvements to the grid overall, which both this block and the data section will benefit from.

image

It's worth noting that this is actually just a component, not a block, and so cannot be ejected. We are making it appear as a block because that aligns more with its functionality.

The core grid is the same as the one in the data section, but with a few important differences.
- There isn't a bar above the grid with buttons in it, so some actions are unavailable (showing and hiding columns, changing row height)
- Column mutation is fully disabled. You can't add, edit or delete columns
- Any schema mutations (resizing columns, reordering columns etc) still work, but are not saved. Users can customise their view as they see fit on the page, but this will be lost when the refresh the page
- Your accent colour is used, because why not
image

### Collaboration
Multi-user collaboration is fully supported, the same as in the data section.
image

The only difference is that you can't explicitly see avatars of connected users because the header bar is not shown. Everything else is identical.

**Collaboration for public users is fully supported. Public users will appear as an "Anonymous" user.**

Collaboration between the prod app and builder does not work, due to internal tables having the same table ID for both the prod and dev apps. Therefore I've had to include the app ID as part of the room key for collaboration, meaning only users in the same app will see each other. This is unfortunate because it means that if you're using a SQL table it will be possible to overwrite each other without knowing, if someone was to use the same table in the builder and published app at the same time.

### Settings
image

The grid's settings are minimal and simple, especially compared to something like the table block. This is deliberate and should be maintained - this is an opinionated component designed for speed and ease of use.

Settings overview:
- Table: the datasource. DS+, so only tables are listed
- Columns: basic column editor. You can choose which columns to include and assign display names. Column order is ignored as we always use the order defined in the actual table.
- Filtering: normal filter, same as table block etc
- Sort column: same as table block etc (can be changed in the component, this is just the initial value)
- Sort order: same as table block etc (can be changed in the component, this is just the initial value)
- Row height: because you can't change the height at runtime, this setting lets you define for all uses. Defaults to using the value of the actual table
- Add rows: enable adding rows
- Edit rows: enable editing rows
- Delete rows: enable deleting rows
- High contrast: use a different background color for alternate rows. e.g.
image

### Adding rows
As there is no add row button, rows can either be added using the empty row as usual, or by using the new floating add button:
image

This button functions identically to the one in the data section - it will scroll to the empty row at the bottom if available, or otherwise show the new row at the top.

## Changes to grid in data section
This includes a few changes to the grid as a whole. There are a bunch of smaller QoL improvements and fixes, but the primary ones are:
- The add row and add column buttons above the grid have been removed, in favour of the inline buttons
- The row height and column width buttons have been compressed into a single "size" button
- The filter and hide column buttons show the number of active items

Overall the buttons above the grid now look like:
image

It's a much cleaner, reduced set of buttons which makes it easier to use.


aptkingston:
Merging this in as it's been open for a week and is difficult to review. Time for manual testing is needed.

PR #10973, opened 2023-06-20
App settings section

deanhannigan:
## Description

- Designer navigation layout has been updated.
- A new settings section has been added inside the app designer
  - App `Delete` and `Export` have been moved in to the side menu of the app settings page.
- The overview section has been completely removed.
  - User/Group management can still be performed inside the designer.
- The publish workflow has been reworked to consolidate a variety of common app actions into once place.
- A new app `Embed` section has been added under settings.
  - When your app is published you can copy the iframe code provided and embed your application wherever you like!
  - Only `public` screens and/or content will we loaded in an embedded application.
- The `/embed/{app_id}` endpoint has been added to accommodate the embedding of applications.
  - Analytics will reflect when an embedded app is served to a user via the embed path.

Addresses: 
- BUDI-7031
- BUDI-7029
- BUDI-7028
- BUDI-7030
- BUDI-7027

## Screenshots

**Designer Updates**
updated_designer_publish

**App Settings Page**
app_settings

**Embed Section**

embed_page

**Embed App**

iframe_example


aptkingston:
Will do a proper code review, but from using this it seems rock solid! Excellent work.

Only change I think would be useful is updating the copy when there's no public access to the embed. Currently it shows
image

Could we update this to "This Budibase app is not publicly accessible"?

packages/builder/src/pages/builder/app/[application]/settings/name-and-url.svelte
```diff
@@ -10,14 +10,22 @@
...
+  import { store } from "builderStore"
+  import { apps } from "stores/portal"
   import UpdateAppModal from "components/start/UpdateAppModal.svelte"
+  import { API } from "api"
 
   let updatingModal
 
-  $: app = $overview.selectedApp
+  $: filteredApps = $apps.filter(app => app.devId == $store.appId)
+  $: app = filteredApps.length ? filteredApps[0] : {}
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/settings/name-and-url.svelte:21]: NAB: but is there any reason you always pull the current app by searching the portal `apps` store rather than from the main builder `store` where we already have the app? We do always load the full list of apps anyway, but doing it this way means we're trying to keep both the apps store and the builder store updated with the current definition of the app being edited.

The side effect of doing it this way is that we fire the main `initialise` method of the builder store multiple times, and I'm worried that's going to keep creating new websocket connections. It also triggers a large cascade of other API requests which could be avoided if we just updated state in the frontend store.

NAB like I said so no need to refactor this unless my testing here does show we're creating multiple websocket connections, in which case we just might need a small refactor of that initialisation logic to ensure it kills it's previous connection if we re-initialise the store while already editing it.

packages/client/src/components/ClientApp.svelte
```diff
@@ -86,7 +89,10 @@
     if (get(builderStore).inBuilder) {
       builderStore.actions.notifyLoaded()
     } else {
-      builderStore.actions.analyticsPing({ source: "app" })
+      builderStore.actions.analyticsPing({
+        source: "app",
+        ...{ embedded: $appStore.embedded || undefined },
```
aptkingston, line 94: NAB: I don't think the spread here does anything (and we could just do a quick `!!` to cast as a boolean. We also don't actually need the `source: "app"` here because the store function adds it automatically (I know this was pre-existing) - so this could just be:
```js
builderStore.actions.analyticsPing({
  embedded: !!$appStore.embedded
})
```

packages/server/src/api/controllers/static/index.ts
```diff
@@ -100,6 +100,10 @@ export const deleteObjects = async function (ctx: any) {
 }
 
 export const serveApp = async function (ctx: any) {
+  const bbHeaderEmbed = ctx.request.get("x-budibase-embed")
+    ? ctx.get("x-budibase-embed").toLowerCase() == "true"
```
aptkingston, line 104: NAB: may as well use `===` here since we're directly comparing two strings. Could also make this a little terser with optional chaining:
```js
const bbHeaderEmbed = ctx.request.get("x-budibase-embed")?.toLowerCase() === "true"
```

packages/client/src/index.js
```diff
@@ -45,6 +45,9 @@ const loadBudibase = async () => {
   // server rendered app HTML
   appStore.actions.setAppId(window["##BUDIBASE_APP_ID##"])
 
+  // Set the flag used to determine if the app is being loaded via an iframe
+  appStore.actions.setAppEmbedded(window["##BUDIBASE_APP_EMBEDDED##"] == "true")
```
aptkingston, line 49: NAB: should probably use an explicit `===` here since we want to compare against the specific value `"true"`. This will function fine, but I think technically it's safer to be explicit.

packages/server/src/middleware/builder.ts
```diff
@@ -86,12 +86,6 @@ export default async function builder(ctx: UserCtx) {
 
   const referer = ctx.headers["referer"]
 
-  const overviewPath = "/builder/portal/overview/"
```
aptkingston, line 89: Nice catch to spot this.

aptkingston (approved):
LGTM! Super work Dean. Will check how websockets react to multiple store initialisations, but otherwise I think we're good to go. Just hang off merging until I've checked that and reported back.

aptkingston (approved):
Like I thought, the multiple store initialisations were indeed creating new websocket instances (without cleaning up the old one) every time. I've updated this to ensure that the websocket acts as a singleton, and just pushed that change straight up.

The other tiny change I included (just to save time) was adding a little bit of padding to the secondary user message so that it matches the spacing of the tabs.

LGTM!

Issue #10812, opened 2023-06-07
[BUDI-7089] OIDC Custom Icons don't appear when added.


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.6.20
  * App Version: 2.6.20

**Describe the bug**
Custom OIDC icons don't appear whenever added returns a blank/missing image. Previous issue fixed the default icons here; [https://github.com/Budibase/budibase/pull/9826](https://github.com/Budibase/budibase/pull/9826)

**To Reproduce**

1. Go to settings
2. Click on auth
3. Scroll down to OpenID Connect
4. Fill in all relevant information
5. Upload a custom icon 
6. See issue

**Expected behavior**
Uploaded icon should be displayed after upload.

**Screenshots**

![242895318-124532c3-faac-4978-8130-4e79ea8c9cd3.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/955f14d1-3acd-4bf3-a91d-8998705e7e41/c7b816a0-c8ee-4cd1-9354-8285f59ce42d)

**Desktop (please complete the following information):**

* OS: IOS
* Browser Brave
* Version **v1.52.122**

**Additional context**
Related discussion; [https://github.com/Budibase/budibase/discussions/10771](https://github.com/Budibase/budibase/discussions/10771)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7089](https://linear.app/budibase/issue/BUDI-7089/oidc-custom-icons-dont-appear-when-added)</sub>

gmentzik:
Confirming this issue. I have opened discussion for it here: https://github.com/Budibase/budibase/discussions/10771
I think this was working fine before version 2.

koneubauer:
Can replicate this issue in 2.13.20

PR #10978, opened 2023-06-20
Fix for custom OIDC icons

deanhannigan:
## Description

- Fix to the `Picker` component to ensure that when `useOptionIconImage` is set, it will load the selected field option icon as an image, rather than passing it to an `Icon` component.
- Fix to ensure that custom OIDC logos are enriched with valid object store URLs on retrieval.

Addresses: 
- https://github.com/Budibase/budibase/issues/10812

## Screenshots

Custom image displaying correctly as the selected option and in the option list
Screenshot 2023-06-20 at 18 14 06

Custom OIDC Icon displayed in the login page.
Screenshot 2023-06-20 at 18 14 33


packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -280,13 +280,38 @@ export async function save(ctx: UserCtx<Config>) {
   }
 }
 
+const parseScopedConfigFileUrls = (cfg: any, type: string) => {
```
Rory-Powell, line 283: I think it's safe to make this an OIDC by default function, the impl is quite specific. Could go with `enrichOIDCLogos` and update the consumers, also giving us some type safety (we should try to stay from `any` as much as possible)

```ts
function enrichOIDCLogos(config: OIDCLogosConfig) {
   ...
}
```

```ts
function find(ctx: UserCtx) {
   ...
   if (type === ConfigType.OIDC_LOGOS) {
      enrichOIDCLogos(config) // shouldn't need to re-assign here
   }
   ...
}
```

```ts
function publicOidc(ctx: Ctx<void, GetPublicOIDCConfigResponse>) {
   if (scopedConfig) {
      enrichOIDCLogos(scopedConfig)
   }
}
```


I think this is somewhat easier to read. Would also be great if we could add a helper for retrieving the logos doc in `publicOidc`, instead of using `configs.getConfig`. Could add this to here: https://github.com/Budibase/budibase/blob/fe9a91ae7ae2a8882ffa768370935fdd4f2014ee/packages/backend-core/src/configs/configs.ts#L192 meaning we can return a typed config rather than a generic one.

aptkingston (approved):

packages/types/src/documents/global/config.ts
```diff
@@ -79,6 +79,8 @@ export interface OIDCConfigs {
   configs: OIDCInnerConfig[]
 }
 
+export interface OIDCLogosConfig extends Config {}
```
Rory-Powell, line 82: Is there any fields we can add to here?

Rory-Powell: Should add config specific fields to `OIDCLogosInnerConfig` and define this type as 
```ts
export interface OIDCLogosConfig extends Config<OIDCLogosInnerConfig> {}
```

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -28,6 +28,8 @@ import {
   SSOConfig,
   SSOConfigType,
   UserCtx,
+  OIDCLogosConfig,
+  OIDCInnerConfig,
```
Rory-Powell, line 32: Unused?

Rory-Powell (approved):

PR #10992, opened 2023-06-22
Fix duplicate loop block issue

PClmnt:
## Description
Fixes an issue where you could click the "Add Looping" button even if a loop block was already attached, breaking the automation.

Also removed some unneeded code in the same file.


aptkingston (approved):
LGTM! May as well get this in before the freeze.

Issue #10949, opened 2023-06-19
[BUDI-7209] Relationship fields (foreign keys) aren't saved when using relational databases


zedascouves:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: valis

**Describe the bug**

(I've initially reported the issue in [https://github.com/Budibase/budibase/discussions/10937](https://github.com/Budibase/budibase/discussions/10937), and I believe this one also relates to the same problem: [https://github.com/Budibase/budibase/discussions/10935.)](https://github.com/Budibase/budibase/discussions/10935.))

When a form contains a dropdown picker of type "Relationship", the field value contains unparsed characters. This prevents the row from being saved, as the foreign key is invalid.

**To Reproduce**

1. Create two tables in a MySQL/MariaDB database, with a relationship of 1 to many.
2. Import these tables in Budibase
3. Create 1-N relationship
4. Autogenerate a listing page for the table with the foreign key column, with save/update actions
5. Edit a row, changing the option related to the foreign key.
6. The selected option will **NOT** be saved.

**Expected behavior**
The row should be saved with the updated values.

**Desktop (please complete the following information):**

* OS: Windows 11
* Browser: Firefox
* Version 114.0.1

**Additional context**
I was able to workaround the issue by changing the "Save row" action and cleaning up the form field value.

`{{ toInt (remove (remove (remove (remove (stringify Details Form.Fields.VIPS_Country) '%5B') '%5D') '["') '"]') }}`

So, my action looks like:

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/dc71df5f-b8ec-4146-97a9-95d692aa1352/0e0d6455-6ad5-471b-8b32-48fcd66a8a32)

This way I can select my "foreign key" element on the picker, and it'll save the correct ID on the row. Nevertheless, it doesn't fix the underlying problem.

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7209](https://linear.app/budibase/issue/BUDI-7209/relationship-fields-foreign-keys-arent-saved-when-using-relational)</sub>

sklaffa:
Hi, I'm the author of report #10935. I would like to highlight that at least the issues I experienced are reproducible also in other database backends. I tried with PostgreSQL and got the very same behaviour, so I suspect the bug is not really backend-dependent.

mike12345567:
[...]

PR #10995, opened 2023-06-22
Make sure CreateEditRelationship modal correctly sets relationship types

mike12345567:
## Description
Fix for #10949 - the relationship types were always locked to be the same, which means that the foreign key wasn't being updated correctly as the backend was not aware of which side was the one-to-many side.

Also fixing an issue with validation where foreign keys that are required to be not null would throw errors before the backend has worked out what the value for it should be.


melohagan (approved):
LGTM!

Issue #7441, opened 2022-08-24
Ability to have readable custom role ids in API


developer-uhwi:
**Describe the feature request**
Assigning a new role to a user does not show in an API request of the user details. E.g. creating a role name "Accountant", does not show as such, instead it shows as "role_xxxxxx", this is not usable when wanting to do other processing externally especially using the API.

Solution: Display the name of the role "Accountant", or display both roleName and the roleId.

**Screenshots**

PR #11001, opened 2023-06-22
Custom roles - readable IDs

mike12345567:
## Description
Custom roles previously used a UUID based role ID, which would often be displayed/cause confusion within the builder. Going forward the role name must now be unique as it will be used to create the role ID for the role itself.

We have limited to role name to alpha-numeric characters and underscores, so that it can be used as part of the ID with out causing problems.

We can now treat the role like the internal role IDs and strip the `role_` prefix, making them readable by default without any conversion.

To achieve this, I've implemented a role version, which controls whether the role is actually a UUID, or whether it is based on the name. This alerts us as to whether or not we need to handle the prefix when saving - it is important to note that this change only applies to new roles, rather than performing a migration.

Addresses: 
- https://github.com/Budibase/budibase/issues/7441
- https://linear.app/budibase/issue/BUDI-5695/ability-to-have-readable-custom-role-ids-in-api-and-design-section

## Screenshots
image


packages/backend-core/src/security/roles.ts
```diff
@@ -25,18 +25,28 @@ const EXTERNAL_BUILTIN_ROLE_IDS = [
   BUILTIN_IDS.PUBLIC,
 ]
 
+export const RoleVersion = {
```
adrinr, line 28: Could we make it more verbose?

```
export const RoleType/RoleFormat = {
  // original version, with a UUID based ID
  UUID: undefined,
  // new version - with name based ID
  NAME: 'name',
}
```

packages/builder/src/components/backend/DataTable/modals/EditRoles.svelte
```diff
@@ -101,14 +100,28 @@
     }
   }
 
+  const getRoleNameError = name => {
+    const hasUniqueRoleName = !otherRoles
```
adrinr, line 104: Should we worry about casing?

packages/server/src/api/controllers/role.ts
```diff
@@ -14,7 +14,8 @@ const UpdateRolesOptions = {
 async function updateRolesOnUserTable(
   db: Database,
   roleId: string,
-  updateOption: string
+  updateOption: string,
+  roleVersion?: number
```
adrinr, line 18: This should probably be mandatory to get sure we update all the usages

packages/builder/src/components/backend/DataTable/modals/EditRoles.svelte
mike12345567: I think we allow it to be case sensitive, we haven't compared the names before now (we already checked if the role name was "unique" but we allowed it to be case sensitive).

packages/backend-core/src/security/roles.ts
```diff
@@ -157,13 +167,16 @@ export async function getRole(
     role = cloneDeep(
       Object.values(BUILTIN_ROLES).find(role => role._id === roleId)
     )
+  } else {
+    // make sure has the prefix (if it has it then it won't be added)
+    roleId = generateRoleID(roleId)
```
melohagan, line 172: NAB: The name of the function is a little strange to me. It's called `generateRoleID` yet you pass the roleId into it.

Maybe it should be called something like `prefixRoleID`

packages/builder/src/components/backend/DataTable/modals/EditRoles.svelte
```diff
@@ -101,14 +100,28 @@
...
+      ?.map(role => role.name)
+      ?.includes(name)
+    const invalidRoleName = !validRegex.test(name)
+    if (!hasUniqueRoleName) {
+      return "Select a unique role name."
+    } else if (invalidRoleName) {
+      return "Please enter a role name consisting of only alphanumeric symbols and underscores"
+    } else {
+      return null
```
melohagan, line 113: NAB: Is it not a little bit of a code-smell to return null from a function?

packages/server/src/api/controllers/role.ts
```diff
@@ -24,11 +25,15 @@ async function updateRolesOnUserTable(
     if (prop === "roleId") {
       updated = true
       const constraints = schema[prop].constraints
-      const indexOf = constraints.inclusion.indexOf(roleId)
+      const updatedRoleId =
+        roleVersion === roles.RoleIDVersion.NAME
+          ? roles.getExternalRoleID(roleId, roleVersion)
+          : roleId
+      const indexOf = constraints.inclusion.indexOf(updatedRoleId)
```
melohagan, line 32: NAB: `indexOfRoleId`

melohagan (approved):
LGTM!

Issue #10340, opened 2023-04-18
[BUDI-6875] DecodeURI Handlebars helpers are confusing


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.5.2
  * App Version: 2.5.2

**Describe the bug**
When looking at the decodeURI helpers, there are two options and the description is wrong.

**Expected behavior**
Reduce the Decode URI/URL helpers into one and fix the description 

**Screenshots**

![Screenshot 2023-04-18 at 17.26.40.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/2cf5cf00-f8c7-441d-bfcd-63710a4d4540/8127933f-4011-40d5-9835-05e8b4375f71)

States *escape* in the example, but prints *decodeURI*. Includes a duplicate function \*\*Url_decode\*\*.

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6875](https://linear.app/budibase/issue/BUDI-6875/decodeuri-handlebars-helpers-are-confusing)</sub>

Issue #6645, opened 2022-07-11
Duplicate handlebars helper (times and multiply)


mslourens:
**Describe the bug**
Both handlebars helpers `times` and `multiply` are doing the same thing. One can be removed 

**To Reproduce**
1. Go to a bindingspanel and search for times
2. Notice the similarity between times and multiply.
3. Create two columns in a BB table, both with type `number`
4. Create two more columns with formula types `times` and `multiply`
5. Create a row with data for the two number columns
6. Notice the result of both formula fields, they are the same

**Expected behavior**
One Handlebars helper for multiplication. 

**Screenshots**
![image](https://user-images.githubusercontent.com/1907152/178297043-a8d3d23f-e389-422d-b30e-462a2ae2476b.png)

**Desktop (please complete the following information):**
 - OS: PopOS 22.04
 - Browser: Vivaldi
 - Version: develop

PR #11003, opened 2023-06-23
Helpers functions bug fixes

ConorWebb96:
## Description
Duplicate helper for multiply being times which didn't serve a purpose. Additionally url_encode and url_decode didn't have descriptions, opted to remove them. Lastly, encodeURI example was incorrect as it had the escape as its example.

Addresses: 
- [Duplicate handlebars helper (times and multiply)](https://github.com/Budibase/budibase/issues/6645)
- [DecodeURI Handlebars helpers are confusing](https://github.com/Budibase/budibase/issues/10340)


packages/string-templates/manifest.json
```diff
@@ -74,8 +74,8 @@
         "b"
       ],
       "numArgs": 2,
-      "example": "{{ multiply 10 5 }} -> 50",
-      "description": "<p>Return the product of <code>a</code> times <code>b</code>.</p>\n"
+      "example": "{{ times 10 5 }} -> 50",
```
andz-bb, line 77: should this be multiply?

andz-bb:
will this break functionality for people who are currently using the times helper?

aptkingston:
> will this break functionality for people who are currently using the times helper?

I *think* this only controls what gets displayed, but the actual helper will still be there and able to be used if you were to free type it, so I don't think it will break existing apps. Definitely worth double checking this though!

PR #11009, opened 2023-06-23
Fix OIDC Custom Icons

deanhannigan:
## Description

Restored OIDC Icon fix - https://github.com/Budibase/budibase/pull/10978

Addresses: 
- https://github.com/Budibase/budibase/issues/10812

## Screenshots

Custom image displaying correctly as the selected option and in the option list
Screenshot 2023-06-20 at 18 14 06

Custom OIDC Icon displayed in the login page.
Screenshot 2023-06-20 at 18 14 33


packages/builder/src/pages/builder/auth/_components/OIDCButton.svelte
```diff
@@ -29,9 +30,9 @@
     }
   })
 
-  $: src = !$oidc.logo
+  $: src = !oidcLogoCheck
     ? OidcLogo
-    : preDefinedIcons[$oidc.logo] || `/global/logos_oidc/${$oidc.logo}`
+    : preDefinedIcons[$oidc.logo] || oidcLogoCheck
```
melohagan, line 35: The statement `preDefinedIcons[$oidc.logo] || oidcLogoCheck` will always be true here.

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -301,13 +328,20 @@ export async function publicOidc(ctx: Ctx<void, GetPublicOIDCConfigResponse>) {
...
+      enrichOIDCLogos(oidcLogoConfig)
+    }
 
     if (!config) {
       ctx.body = []
     } else {
       ctx.body = [
         {
-          logo: config.logo,
+          logo: oidcLogoConfig
```
melohagan, line 342: Can this be simplified? Something like:

```js
logo: oidcLogoConfig?.config[config.logo] ?? config.logo
```

melohagan (changes requested):
There's a boolean statement that needs double checked, unless I'm mistaken!

packages/worker/src/api/controllers/global/configs.ts
aptkingston: Yep I agree with Mel here, definitely should be simplified into a one liner. On a separate note, the naming here is a little confusing haha.

packages/builder/src/pages/builder/auth/_components/OIDCButton.svelte
```diff
@@ -29,9 +30,9 @@
     }
   })
 
-  $: src = !$oidc.logo
+  $: src = !oidcLogoCheck
```
aptkingston, line 33: What was the purpose of pulling out `$oidc.logo` to a separate variable but then only using it once in the statement? Personally I think that if something can't be simplified to a one liner, then multi-line ternary statements should be broken out into functions to make it easier to understand the logic, and possible to maintain it in future.

packages/builder/src/pages/builder/auth/_components/OIDCButton.svelte
deanhannigan: I've refactored out the area in favour of `oidcLogoImageURL = preDefinedIcons[$oidc.logo] ?? $oidc.logo` which is largely the same.

`$oidc.logo` will either be one of the preset strings (`OidcLogo`,`Auth0` etc),  a custom logo file URL or `null`

If I chose the Auth0 icon in budibase, `$oidc.logo` would be set to 'Auth0' and `preDefinedIcons[$oidc.logo]` would  resolve to `/builder/assets/auth0-logo.png`

If the user selected one of their own images, `$oidc.logo` would resolve to something like `/files/signed/global/logos_oidc/mylogo.png`

In the event that absolutely nothing was set in `oidcLogoImageURL` after that , `OidcLogo` will be used as the default.

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -280,13 +281,39 @@ export async function save(ctx: UserCtx<Config>) {
   }
 }
 
+function enrichOIDCLogos(oidcLogos: OIDCLogosConfig) {
+  if (!oidcLogos) {
+    return
+  }
+  oidcLogos.config = Object.keys(oidcLogos.config).reduce(
```
melohagan, line 288: Can we guarantee that `oidcLogos.config` will always be an object?

I recently fixed a null-pointer where I had to do `Object.keys(x || {})`

melohagan (approved):
LGTM!

packages/worker/src/api/controllers/global/configs.ts
deanhannigan: It's unlikely to ever not be an object once its been created, but definitely better safe than sorry

PR #11013, opened 2023-06-23
Further SQL relationship fixes

mike12345567:
## Description
After some discussion in the listed issue, there was still a few issues with utilising foreign keys that are not-nullable in screens. I've added functionality to switch foreign keys to an autocolumn when in use, and when the relationship is removed revert them back to standard columns. This means that the frontend does not get confused and require the user to enter values for the foreign keys which will be enriched by Budibase.

I've also fixed an issue with fetching tables and the table filter list, while this is going away in the next release needed to get a fix out for the functionality until the new version is released.

Addresses: 
- https://github.com/Budibase/budibase/issues/10949

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -183,9 +183,7 @@ export async function buildSchemaFromDb(ctx: UserCtx) {
 
   let { tables, error } = await buildSchemaHelper(datasource)
   if (tablesFilter) {
-    if (!datasource.entities) {
-      datasource.entities = {}
-    }
+    datasource.entities = {}
     for (let key in tables) {
       if (
         tablesFilter.some(
@@ -200,7 +198,7 @@ export async function buildSchemaFromDb(ctx: UserCtx) {
   }
 
   setDefaultDisplayColumns(datasource)
-  const dbResp = await db.put(datasource)
+  const dbResp = await db.put(sdk.tables.checkExternalTableSchemas(datasource))
   datasource._rev = dbResp.rev
   const cleanedDatasource = await sdk.datasources.removeSecretSingle(datasource)
 
@@ -286,7 +284,9 @@ export async function update(ctx: UserCtx<any, UpdateDatasourceResponse>) {
     datasource.config!.auth = auth
   }
 
-  const response = await db.put(datasource)
+  const response = await db.put(
+    sdk.tables.checkExternalTableSchemas(datasource)
+  )
   await events.datasource.updated(datasource)
   datasource._rev = response.rev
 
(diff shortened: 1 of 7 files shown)


packages/server/src/sdk/app/tables/validation.ts
```diff
@@ -0,0 +1,67 @@
...
+        delete column.autoReason
+      }
+    }
+  }
+
+  return datasource
+}
+
+export function isEditableColumn(column: FieldSchema) {
+  const isAutoColumn =
```
adrinr, line 59: Is a foreign key `column.autocolumn`?

packages/types/src/documents/app/table.ts
```diff
@@ -9,6 +9,10 @@ export enum RelationshipTypes {
   MANY_TO_MANY = "many-to-many",
 }
 
+export enum AutoReason {
+  FOREIGN_KEY = "foreign_key",
```
adrinr, line 13: This is a little bit confusing. Should we not have `primary_key` and other options?

packages/server/src/sdk/app/tables/validation.ts
```diff
@@ -0,0 +1,67 @@
...
+    column.autocolumn && column.autoReason !== AutoReason.FOREIGN_KEY
+  const isFormula = column.type === FieldTypes.FORMULA
+  return !(isAutoColumn || isFormula)
+}
+
+export function checkExternalTableSchemas(datasource: Datasource) {
```
adrinr, line 65: This is not checking, this is setting values. By checking I would expect to be a pure function. Could we rename it to populateExternalTable or similar?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -183,9 +183,7 @@ export async function buildSchemaFromDb(ctx: UserCtx) {
 
   let { tables, error } = await buildSchemaHelper(datasource)
   if (tablesFilter) {
-    if (!datasource.entities) {
```
adrinr, line 186: How come will this not be populated anymore?

packages/server/src/sdk/app/tables/validation.ts
mike12345567: yes it is - when used in a relationship we want to determine it is an auto-column, but also still allow editing of it via API if user desires.

packages/server/src/api/controllers/datasource.ts
mike12345567: This was actually causing a bug with the old table filters, not resetting the entities structure before each filter meant that you couldn't remove already added tables - not super important given this is all being updated shortly anyway.

adrinr (approved):

PR #11026, opened 2023-06-26
Chore - Speed up tag pipeline

adrinr:
## Description
Moving dependencies in scripts to avoid installing everything for a single script run. Currently it's tacking a long time, and 90% of the time is because the yarn install


Rory-Powell (approved):
nice idea

PR #11031, opened 2023-06-26
Chore - Unify pro lint

adrinr:
## Description
Change the lint config to lint pro as well, treating it as any other package https://github.com/Budibase/budibase-pro/pull/202


mike12345567 (approved):

package.json
```diff
@@ -56,6 +56,7 @@
     "lint:eslint": "eslint packages && eslint qa-core",
     "lint:prettier": "prettier --check \"packages/**/*.{js,ts,svelte}\" && prettier --write \"examples/**/*.{js,ts,svelte}\" && prettier --check \"qa-core/**/*.{js,ts,svelte}\"",
     "lint": "yarn run lint:eslint && yarn run lint:prettier",
+    "lint:pro": "yarn eslint packages/pro && yarn prettier --check packages/pro",
```
Rory-Powell, line 59: This is just for local dev?

Rory-Powell: It seems a bit strange to have a command to specifically lint one package, if we don't do it for others like `lint:server`, `lint:worker` etc. Does it have a specific workflow in mind?

Rory-Powell (approved):

PR #11033, opened 2023-06-26
Don't sync automation logs from prod to dev app

mike12345567:
## Description
Adding a test case for the removal of automation logs from app sync, as well as adding to the publish/sync filter a check to not carry over automation logs.

This was causing memory issues in very large apps containing say 300K+ automation logs and there isn't much (if any) benefit to syncing.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7188/remove-syncing-of-automation-history-from-production-to-development

### DIFF
--- a/packages/server/src/api/routes/tests/application.spec.ts
+++ b/packages/server/src/api/routes/tests/application.spec.ts
@@ -325,7 +325,7 @@ describe("/applications", () => {
     })
   })
 
-  describe("app sync", () => {
+  describe("POST /api/applications/:appId/sync", () => {
     it("should not sync automation logs", async () => {
       // setup the apps
       await config.createApp("testing-auto-logs")
@@ -344,19 +344,15 @@ describe("/applications", () => {
         .expect(200)
 
       // does exist in prod
-      await context.doInAppContext(config.getProdAppId(), async () => {
-        const logs = await config.getAutomationLogs()
-        expect(logs.data.length).toBe(1)
-      })
+      const prodLogs = await config.getAutomationLogs()
+      expect(prodLogs.data.length).toBe(1)
 
       // delete prod app so we revert to dev log search
       await config.unpublish()
 
       // doesn't exist in dev
-      await context.doInAppContext(config.getAppId(), async () => {
-        const logs = await config.getAutomationLogs()
-        expect(logs.data.length).toBe(0)
-      })
+      const devLogs = await config.getAutomationLogs()
+      expect(devLogs.data.length).toBe(0)
     })
   })
 })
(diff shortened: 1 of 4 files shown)


packages/server/src/api/routes/tests/application.spec.ts
```diff
@@ -324,4 +324,39 @@ describe("/applications", () => {
       expect(events.app.unpublished).toBeCalledTimes(1)
     })
   })
+
+  describe("app sync", () => {
```
Rory-Powell, line 328: Preference: `describe` blocks cover the endpoint they are testing e.g. 

```ts
describe("POST /api/applications/:appId/sync", () => {})
```

packages/server/src/api/routes/tests/application.spec.ts
```diff
@@ -324,4 +324,39 @@ describe("/applications", () => {
...
+        .expect(200)
+
+      // does exist in prod
+      await context.doInAppContext(config.getProdAppId(), async () => {
+        const logs = await config.getAutomationLogs()
+        expect(logs.data.length).toBe(1)
+      })
+
+      // delete prod app so we revert to dev log search
+      await config.unpublish()
```
Rory-Powell, line 353: This seems like pretty un-deterministic behaviour, is there not a way retrieve the desired logs by id?

packages/server/src/tests/utilities/structures.ts
```diff
@@ -241,6 +243,21 @@ export function collectAutomation(tableId?: string): Automation {
   return automation as Automation
 }
 
+export function basicAutomationLog(automationId: string): AutomationResults {
```
Rory-Powell, line 246: Naming here is somewhat confusing. Is it an `AutomationLog` or `AutomationResults`?

packages/server/src/api/routes/tests/application.spec.ts
mike12345567: No - the issue isn't by ID, its by database, the way the system is designed it retrieves from prod then from dev. I can probably make this more clear in the test as the context shifts aren't really needed.

packages/server/src/tests/utilities/structures.ts
mike12345567: Good point, yes this returns an automation result, mis-typed the function name!

packages/server/src/api/routes/tests/application.spec.ts
Rory-Powell: By "id" I mean the app id vs the prod id - essentially the different databases. 

I think the logic is just unclear? Needing to remove the prod db in order to query the dev db logs seems counter intuitive. Often tests highlight things like this which can be an opportunity to improve the interface. 

NAB = just wondered if there was an easier way to get the dev db logs.

mike12345567: Yeah - there isn't a particularly nice way to do this - without basically writing a bunch of code into the SDK to provide access to something that isn't supposed to be accessible.

PR #11045, opened 2023-06-27
Prompt to create queries

Ghrehh:
## Description
Prompts the user to create a query upon creating a non-datasource+ datasource. It wasn't specified in the ticket, but I opted to exclude REST from this for now, as the copy doesn't seem to take it into account, and the flow is a bit different.

NOTE: Icon differs from the mockups, but design requested it be done this way.

Addresses: 
- https://linear.app/budibase/issue/BUDI-6946/non-ds-modal-and-action-to-create-first-query

## Screenshots

Screenshot 2023-06-27 at 16 24 20

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/PromptQueryModal.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/PromptQueryModal.svelte
@@ -1,0 +1,61 @@
+<script>
+  import { goto as gotoStore, params as paramsStore } from "@roxi/routify"
+  import { Modal, ModalContent, Body, Heading } from "@budibase/bbui"
+  import FontAwesomeIcon from "components/common/FontAwesomeIcon.svelte"
+
+  const handleOpen = (modal, params) => {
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
LGTM!

packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/PromptQueryModal.svelte
```diff
@@ -0,0 +1,61 @@
...
+  const handleOpen = (modal, params) => {
+    if (params["?promptQuery"] && modal?.show) {
+      modal.show()
+      history.replaceState({}, null, window.location.pathname)
+    }
+  }
+
+  let modal
+  $: params = $paramsStore
+  $: goto = $gotoStore
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/PromptQueryModal.svelte:15]: Random question - what's the purpose of reassigning these stores to local variables?

Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/PromptQueryModal.svelte:15]: Good question, for whatever reason, if I'd just reloaded the page, I kept getting an error when I tried to use `get` on them in the callback 🤷 , no idea why

PR #11067, opened 2023-06-29
Use constant for keeping modal open

Ghrehh:
We were previously using the method of returning `false` to keep modals from closing in certain scenarios, it isn't always super explicit why we were doing this, so this was often accompanied by comment explaining why. It seems like a better idea to have a named constant to achieve this instead.


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -159,7 +160,7 @@
   async function saveColumn() {
     savingColumn = true
     if (errors?.length) {
-      return
+      return keepOpen
```
aptkingston, line 163: This particular case didn't return `false` previously - are we still ok to update this to change the return value?

aptkingston (commented):
LGTM! Just one query about one of the updates.

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
Ghrehh: Nope, you're right, mistake on my part.

aptkingston (approved):

PR #11073, opened 2023-06-29
MS-SQL fixes

mike12345567:
## Description
Fixes discovered by MS-SQL validator tests in QA-core, fixing custom ports being used (not default) and allowing the use of self signed certificates. Also updating an error message in line with Peters changes.

### DIFF
--- a/packages/server/src/integrations/microsoftSqlServer.ts
+++ b/packages/server/src/integrations/microsoftSqlServer.ts
@@ -184,14 +184,17 @@ class SqlServerIntegration extends Sql implements DatasourcePlus {
 
   async connect() {
     try {
+      const encrypt = this.config.encrypt
       const clientCfg: MSSQLConfig & sqlServer.config = {
         ...this.config,
         port: +this.config.port,
         options: {
-          encrypt: this.config.encrypt,
+          encrypt: encrypt,
           enableArithAbort: true,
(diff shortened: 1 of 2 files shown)


packages/server/src/integrations/microsoftSqlServer.ts
```diff
@@ -186,10 +186,11 @@ class SqlServerIntegration extends Sql implements DatasourcePlus {
     try {
       const clientCfg: MSSQLConfig & sqlServer.config = {
         ...this.config,
-        port: +this.config,
+        port: +this.config.port,
         options: {
           encrypt: this.config.encrypt,
           enableArithAbort: true,
+          trustServerCertificate: true,
```
adrinr, line 193: Should this always be true? Should it not depend of any config? Any downside of having it always on?

mike12345567: I suppose we could make it that if encryption is enabled we always allow it? I think having it as an option is more of a pain than just allowing self signed certs, they are very common now days afterall.

adrinr: Sounds good, just doublechecking that having it always on does not cause some edge case issues

adrinr (approved):

PR #11099, opened 2023-07-03
Apply column order for Grid block

melohagan:
## Description
Configure column draggable re-order in the Design section is now disabled to avoid confusion.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7234/configured-column-order-ignored-by-grid-block

## Screenshots

Grid block:
Screenshot 2023-07-03 at 11 50 49

Table block for comparison:
Screenshot 2023-07-03 at 11 51 24

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
@@ -111,7 +111,7 @@
             {#each columns as column (column.id)}
               <div class="column" animate:flip={{ duration: flipDurationMs }}>
                 <div
-                  hide={!allowReorder || null}
+                  class:hide={!allowReorder}
                   class="handle"
                   aria-label="drag-handle"
                   style={dragDisabled ? "cursor: grab" : "cursor: grabbing"}
@@ -195,7 +195,7 @@
     display: grid;
     place-items: center;
   }
-  .handle[hide] {
+  .handle.hide {
     visibility: hidden;
   }
   .wide {
--- a/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
@@ -18,6 +18,7 @@
   export let options = []
   export let schema = {}
   export let allowCellEditing = true
+  export let allowReorder = true
 
   const flipDurationMs = 150
   let dragDisabled = true
@@ -110,6 +111,7 @@
             {#each columns as column (column.id)}
               <div class="column" animate:flip={{ duration: flipDurationMs }}>
                 <div
+                  hide={!allowReorder || null}
                   class="handle"
                   aria-label="drag-handle"
                   style={dragDisabled ? "cursor: grab" : "cursor: grabbing"}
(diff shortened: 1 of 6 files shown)


aptkingston (changes requested):
Unfortunately we can't merge this one because it prevents column reordering at all! The grid block allows you to reorder columns as you see fit, but schema overrides are constant and always applied over the top, so this change breaks the innate functionality of column reordering.

I think it also makes sense for the grid to inherit the default order of columns from the table, the same way it inherits other settings like column widths.

If this is a highly requested feature then we can look at building something new which would allow just a "default" order override to be specified, but then still allow customisation after that.

In the meantime the better change might just be to disable the reordering inside this setting, which would at least make it clear that you can't specify a custom order.

aptkingston:
This PR has helped me spot another issue though, so thank you! I just realised that if a column is hidden in the table, it will never be visible in the block. I'll fix that one now - there are definitely some updates needed so I'll see if perhaps there's a way to use the order from this setting too.

aptkingston:
> Thanks @aptkingston I realised after I put up the PR that the user column ordering would be prevented. I have an idea of possibly passing through the order in the schemaOverrides but only applying it if order is undefined, i.e. make the configure column order an initial sort only.
> 
> I'll see how I get on with that.

Just an FYI that the order is unlikely to ever be undefined - as soon as you ever reorder columns in the parent table, the schema order is saved for all columns. So even at initial load in the grid block, the order will have been defined. I don't think there's any nice way to apply custom order while also allowing reordering in the grid block.

I think we'd have to offer some sort of setting which would do what this PR does - disable column reordering in the block, but enables a fixed custom order.

melohagan:
@aptkingston As discussed, I've added a prop to disable column drag and drop re-order for the Grid block to prevent user confusion, i.e. the impression that the column order can be configured in the Design section.
Initial column order comes from the Data section.

packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
```diff
@@ -110,6 +111,7 @@
             {#each columns as column (column.id)}
               <div class="column" animate:flip={{ duration: flipDurationMs }}>
                 <div
+                  hide={!allowReorder || null}
```
aptkingston, line 114: Custom HTML attributes should always be prefixed with `data-` per the HTML5 spec. Querying by custom attributes is also more expensive than class names, so we may as well stick with the normal approach of using classes and just update this to `class:hide={!allowReorder}` (we can remove the `|| null` as well).

packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
```diff
@@ -193,6 +195,9 @@
     display: grid;
     place-items: center;
   }
+  .handle[hide] {
```
aptkingston, line 198: This would just need updated to `.handle.hide` as well, if you update to use a class rather than a custom attribute.

aptkingston (commented):
LGTM but just one small suggested change to follow best practices 👍

packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnDrawer.svelte
melohagan: Good to know - thanks! Updated.

aptkingston (approved):
LGTM!

PR #11103, opened 2023-07-03
Handle errors in grid when adding a new row

PClmnt:
## Description
Handles a case where errors weren't being displayed due to non existence of a Focused Cell ID. We have to show an actual notification in this instance instead of an inline Grid error.

### DIFF
--- a/packages/frontend-core/src/components/grid/stores/rows.js
+++ b/packages/frontend-core/src/components/grid/stores/rows.js
@@ -245,16 +245,7 @@ export const deriveStores = context => {
         focusedCellId.set(`${rowId}-${erroredColumns[0]}`)
       }
     } else {
-      // Some other error - just update the current cell
-      if (get(focusedCellId)) {
-        validation.actions.setError(
-          get(focusedCellId),
-          error?.message || "Error"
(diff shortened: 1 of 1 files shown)


packages/frontend-core/src/components/grid/stores/rows.js
```diff
@@ -246,6 +246,14 @@ export const deriveStores = context => {
...
     } else {
       // Some other error - just update the current cell
+      if (get(focusedCellId)) {
+        validation.actions.setError(
+          get(focusedCellId),
+          error?.message || "Error"
+        )
+      } else {
+        get(notifications).error(error?.message || "An unknown error occurred")
+      }
```
aptkingston, line 256: We don't actually need to check here at all for a focused cell ID - when you press submit it will blur the focused cell, so we'll never have one unless they use they keyboard. I think we just update this to always show a notification 👍 We can also remove the line after this, which I think you just left in by mistake.

PClmnt: Sorted that!

aptkingston (approved):

PR #11106, opened 2023-07-03
BUDI-7240 - Refetch relationships

adrinr:
## Description
Fixing a bug caused by the relationships info and columns not being fetched properly after adding new ones

Addresses: 
- [BUDI-7240](https://linear.app/budibase/issue/BUDI-7240/adding-relationships-and-not-refreshing-keeps-datasources-in-bad-state)

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -186,7 +186,10 @@
         saveColumn.type === LINK_TYPE &&
         saveColumn.relationshipType === RelationshipTypes.MANY_TO_MANY
       ) {
-        dispatch("updatetables")
+        // Fetching the new tables
+        tables.fetch()
+        // Fetching the new relationships
+        datasources.fetch()
       }
       if (originalName) {
         notifications.success("Column updated successfully")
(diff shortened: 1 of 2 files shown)


PClmnt (approved):

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -182,6 +182,15 @@
...
       })
       dispatch("updatecolumns")
+      if (
+        saveColumn.type === LINK_TYPE &&
+        saveColumn.relationshipType === RelationshipTypes.MANY_TO_MANY
+      ) {
+        // Fetching the new tables
+        tables.fetch()
+        // Fetching the new relationships
+        datasources.fetch()
```
aptkingston, line 192: I wish there was a cleaner way to just update state with the new data and avoid hitting the API again, but in this case because it affects both the datasource docs and table docs (and multiple tables at that) I'm happy to just refresh everything.

aptkingston (approved):

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
mike12345567: We could theoretically update the table API to return all the affected entities when a relationship is added (or anything that affects multiple tables) but it might be a fair bit of extra work.

PR #11110, opened 2023-07-03
Add section setting to field group

melohagan:
## Description
As a user, I want a convenient way to layout my form fields. At the moment you need to use multiple containers and play around with width settings if you want form fields to be displayed in a row.
Sections cannot be added within a form, therefore I have updated the field group component to include a section so that you have more quick and easy layout options for your form fields.

Also the CSS was killing me for the left aligned labels. In the end I was able to make it appear nicely for "One Column" mode, but in other modes only the "Above" label is suitable (similar to a form without the field group).

Addresses: 
- https://github.com/Budibase/budibase/discussions/6158#discussioncomment-6343676

## Screenshots
**One Column (default)**
Screenshot 2023-07-03 at 14 02 08

**Two Column**
Screenshot 2023-07-03 at 14 02 29

**Three Column**
Screenshot 2023-07-03 at 14 02 47

**Main with Sidebar**
Screenshot 2023-07-03 at 14 03 06

**Sidebar with Main**
Screenshot 2023-07-03 at 14 03 25

**Left aligned labels**
Screenshot 2023-07-03 at 14 54 51

## Documentation

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -2355,7 +2355,11 @@
         "type": "section",
         "label": "Type",
         "key": "type",
-        "defaultValue": "oneColumn"
+        "defaultValue": "oneColumn",
+        "dependsOn": {
+          "setting": "labelPosition",
+          "value": "above"
+        }
       }
     ]
   },
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -2350,6 +2350,12 @@
             "value": "above"
           }
         ]
+      },
+      {
+        "type": "section",
+        "label": "Type",
+        "key": "type",
+        "defaultValue": "oneColumn"
       }
     ]
   },
--- a/packages/client/src/components/app/forms/FieldGroup.svelte
+++ b/packages/client/src/components/app/forms/FieldGroup.svelte
@@ -1,7 +1,9 @@
 <script>
   import { getContext, setContext } from "svelte"
+  import Section from "../Section.svelte"
 
   export let labelPosition = "above"
+  export let type = "oneColumn"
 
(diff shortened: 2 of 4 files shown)


packages/client/src/components/app/forms/FieldGroup.svelte
```diff
@@ -13,7 +15,18 @@
     class="spectrum-Form"
     class:spectrum-Form--labelsAbove={labelPosition === "above"}
   >
-    <slot />
+    <Section {type}>
```
aptkingston, line 18: Could we instead check first if labels are left or above, and then only render the section if we're using above? That way we can continue to just render a `slot` of left labels, allowing the table layout to work without another form node.

packages/client/manifest.json
```diff
@@ -2350,6 +2350,12 @@
             "value": "above"
           }
         ]
+      },
+      {
+        "type": "section",
+        "label": "Type",
+        "key": "type",
+        "defaultValue": "oneColumn"
```
aptkingston, line 2358: We could make this new setting depend on a label position value of "above". Then users will only see it when they're actually able to use it.

aptkingston (changes requested):
Really nice feature Mel! Left aligned labels use a table layout (thanks Spectrum) which is why it's impossible to use more than one column! I've suggested a couple of tweaks to tidy this one up, but it will be a nice change.

packages/client/src/components/app/forms/FieldGroup.svelte
melohagan: That's nicer - thanks!

aptkingston (approved):
LGTM!

PR #11119, opened 2023-07-03
Linting improvements

Ghrehh:
This is probably best reviewed per commit:

1. Linting config and package changes.
2. Manual code changes.
3. Automatic code changes.

- Eslint warnings now fail the build, so that they stop getting pushed into master, as they probably should be addressed or ignored with comments so that don't make local linting/CI more difficult to read.
- Updates the eslint/babel packages, as we were using several deprecated ones. 
- The newer official svelte linter can detect unused variables, so I removed a bunch of those, as well as fixing various other errors.
- I'll comment some other errors with relevant changes


packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1305,7 +1305,7 @@ export const getFrontendStore = () => {
     links: {
       save: async (url, title) => {
         const navigation = get(store).navigation
-        let links = [...navigation?.links]
+        let links = [...(navigation?.links ?? [])]
```
Ghrehh, line 1308: There's a couple of these around the place:

The linter was complaining that safe navigation wasn't actually safe, which makes sense, because if `navigation` is undefined here, we can't spread it as it's not an interable.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/AutomationBlockTagline.svelte
```diff
@@ -60,6 +60,7 @@
 </script>
 
 <div>
+  <!-- eslint-disable-next-line svelte/no-at-html-tags-->
```
Ghrehh, line 63: There's various places we unsafely inject HTML, but given what our product is that's not really avoidable, so just comment these errors out.

aptkingston (approved):
Nice one @Ghrehh! Changes look good to me. Always happy to make these sort of changes which help maintain quality.

PR #11123, opened 2023-07-04
Fix grid block row height not working

aptkingston:
## Description
Fixes the row height setting being overridden by the table schema in grid blocks. The grid now has a "fixedRowHeight" setting rather than an "initialRowHeight" setting. This disables any other sort of row height change when set.

Addresses: 
- BUDI-7245

### DIFF
--- a/packages/frontend-core/src/components/grid/controls/SizeButton.svelte
+++ b/packages/frontend-core/src/components/grid/controls/SizeButton.svelte
@@ -8,7 +8,8 @@
     SmallRowHeight,
   } from "../lib/constants"
 
-  const { stickyColumn, columns, rowHeight, table } = getContext("grid")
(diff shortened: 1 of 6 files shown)


packages/frontend-core/src/components/grid/controls/SizeButton.svelte
```diff
@@ -86,6 +87,7 @@
       <div class="options">
         {#each rowSizeOptions as option}
           <ActionButton
+            disabled={$fixedRowHeight}
```
melohagan, line 90: NAB: Is this needed? The Grid block doesn't have this button, unless I'm mistaken!

melohagan (approved):
LGTM! Thanks for the fast fix 😄

PR #11131, opened 2023-07-04
Fix google sheets import

Ghrehh:
Google sheets table fetching was failing because we weren't fetching the `entities` property for google sheets datasources. Not fetching this was necessary at some point, but not any more, so we can just remove the check and always fetch for datasource plus.


Ghrehh:
If the release is made before this, this should probably go straight into master, as sheets imports are completely broken atm (at least for me)

mike12345567 (approved):

adrinr:
We should update the description of https://github.com/Budibase/budibase/pulls

PR #11140, opened 2023-07-05
Allow suppressing block row creation alerts

Ghrehh:
## Description
Adds a setting to suppress row creation alerts for the form and table block.

Addresses: 
- https://linear.app/budibase/issue/BUDI-5433/allow-disabling-success-notifications-for-blocks

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -4490,6 +4490,16 @@
               "setting": "clickBehaviour",
               "value": "details"
             }
+          },
+          {
+            "type": "boolean",
+            "label": "Do not display default save notification",
+            "key": "notificationOverride",
+            "defaultValue": false,
+            "dependsOn": {
+              "setting": "clickBehaviour",
+              "value": "details"
+            }
           }
         ]
       },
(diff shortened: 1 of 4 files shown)


packages/client/manifest.json
```diff
@@ -4490,6 +4490,16 @@
               "setting": "clickBehaviour",
               "value": "details"
             }
+          },
+          {
+            "type": "boolean",
+            "label": "Do not display default save notification",
```
aptkingston, line 4496: Does this fit in one line? If not then I think we should reword this slightly so that it doesn't wrap.

aptkingston (approved):
LGTM! Just one question about the label length.

packages/client/manifest.json
Ghrehh: It does not (though neither does the label above it if I remember correctly) I'll try and think of a different wording

Ghrehh: I had to changed it to 'Hide save alert' to get it fit, not sure if that's acceptable. I thought about inverting the toggle and dropping the verb at the start, to get `Save notification`, but the problem is that also sounds like an action so idk.

PR #11145, opened 2023-07-05
Add BigInt type support

melohagan:
## Description
BigInts were numbers, then strings, but for full support they need to be their own type. Luckily it seems knex and lucene handles this well, and I didn't need to cast anything to JavaScripts BigInt type.

- Sort numerically
- Filter numerically
- Display values larger than JavaScrip MAX_NUMBER
- Add BigInt fields to BudibaseDB tables
- Support for BigInt fields in forms

Addresses: 
- https://linear.app/budibase/issue/BUDI-7256/add-bigint-type-support

## Screenshots
Screenshot 2023-07-05 at 17 27 02

Screenshot 2023-07-05 at 17 29 48

Screenshot 2023-07-05 at 17 32 48

## Documentation

*Will update if this is the right way to go*.


shogunpurple (approved):
LGTM! At first I thought it a bit weird to have a specific bigint field, but given it's existence in SQL, this is likely to be the easiest way to solve it rather than overload the other types.

Issue #10884, opened 2023-06-13
[BUDI-7119] Repeated addition of new App State keys in bindings drawer


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.7.14
  * App Version: 2.7.14

**Describe the bug**
When using the **Update State** action and entering a brand new key, it gets added to the bindings drawer list as expected.
If however you type in a value for the key, replace, and replace again, now three values will appear in the bindings drawer instead of one.

**To Reproduce**

 1. Add a button to a screen and `Define actions`
 2. Select **Update State**, *Set Value*
 3. Type a value into the Key field.
 4. Clear
 5. Type a different value into the Key field.
 6. Clear
 7. Type a different value into the Key field.
 8. Clear
 9. Click the lightning bolt icon of the *Value* field.
10. See that the state category of the bindings drawer has multiple options, not just one.

**Expected behavior**
When appending a new App State variable to the bindings drawer, it should only have the last typed value, and not a change history.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/2d6f07aa-d1a0-4060-b66e-40e61eea30ac/6ba18221-ad3f-4d2b-a792-5cce91f8f5a6)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7119](https://linear.app/budibase/issue/BUDI-7119/repeated-addition-of-new-app-state-keys-in-bindings-drawer)</sub>

PR #11162, opened 2023-07-07
Filter duplicate state bindings when editing Update State button action.

deanhannigan:
## Description

When editing `Update State` button actions in the button action editor, the `state` bindings are filtered to avoid duplication. 

Addresses: 
- https://github.com/Budibase/budibase/issues/10884


aptkingston (approved):

aptkingston:
Had a quick look at this locally before merging in case there was an easier way to solve this, and think I've found that it's actually only a tiny change! Hold off on merging this until I confirm.

PR #11172, opened 2023-07-10
Add embed config to single image nginx config

aptkingston:
## Description
Adds the embed route to the single image nginx config, which was left out in the initial PR.


shogunpurple (approved):
LGTM!

PR #11173, opened 2023-07-10
Enabled collaboration behaviour in the automation section

deanhannigan:
## Description

Collaboration has been extended to the automation section. 

- See what everyone is currently working on in the automation section!
- Live broadcast of any changes made by collaborators working in your application automations.

Addresses: 
- BUDI-7151

## Screenshots

Screenshot 2023-07-17 at 11 36 22


packages/builder/src/builderStore/store/automation/index.js
```diff
@@ -248,4 +248,39 @@ const automationActions = store => ({
...
+    }
+    if (!automation) {
+      store.update(state => {
+        // Remove the automation
+        state.automations = state.automations.filter(
+          x => x._id !== automationId
+        )
+        // Select a new automation if required
+        if (automationId === state.selectedAutomationId) {
+          store.actions.select(state.automations[0]?._id)
```
aptkingston, line 263: NAB: won't do anything bad having this here I don't think, but is this needed? I would have expected that the URL state sync util would correctly recognise that the selected automation ID was no longer valid and trigger a redirect to select a new one.
That's what happens with other entities (screens etc).

aptkingston (approved):
LGTM! Will have a play locally too but code looks great.

packages/builder/src/builderStore/store/automation/index.js
deanhannigan: Yep, I did a bit a re-testing there and the state sync redirect stops it from becoming an issue  :thumbsup:

aptkingston: Cool, lets just remove that then and we can merge after.

PR #11174, opened 2023-07-10
set global retry times on QA core tests

shogunpurple:
## Description
Noticed that a majority of the failing QA core CI tests in our scheduled run fail due to issues like network flakiness, rather than the tests themselves. I've added a global retry to run them again one more time if they fail, which ideally should give us more stability and reduce false positives.

### DIFF
--- a/qa-core/src/jest/jestSetup.ts
+++ b/qa-core/src/jest/jestSetup.ts
@@ -1,4 +1,5 @@
 import { logging } from "@budibase/backend-core"
 logging.LOG_CONTEXT = false
 
+jest.retryTimes(2)
 jest.setTimeout(60000)


mike12345567 (approved):

qa-core/src/jest/jestSetup.ts
```diff
@@ -1,4 +1,5 @@
 import { logging } from "@budibase/backend-core"
 logging.LOG_CONTEXT = false
 
+jest.retryTimes(2)
```
Rory-Powell, line 4: The only thing I'm uncertain of is the idempotency of the test run. Given they usually setup some data, set some global state, I'm not sure where the retry will/will not help - but worth a shot

Rory-Powell (approved):

PR #11185, opened 2023-07-10
Self host console log export

adrinr:
## Description
Implementing exporting log tail on self hosted 

Addresses: 
- [BUDI-7113](https://linear.app/budibase/issue/BUDI-7113/self-host-console-log-export)

## Screenshots
image

image


packages/frontend-core/src/api/logs.js
```diff
@@ -0,0 +1,14 @@
+export const buildLogsEndpoints = API => ({
+  /**
+   * Gets a list of datasources.
+   */
+  getServerLogs: async () => {
+    return await API.get({
+      url: "/api/global/system/logs",
```
Rory-Powell: I think the `/global/system` here could be confusing, as we have currently:
- `/api/global` - tenant level endpoints
- `/api/system` - installation level endpoints

given that the logs are an installation level concept, we could probably just remove `global` here

adrinr: They are system logs. Maybe they should be `api/global/systemlogs` or just `api/global/logs` (I am worried that this can be confusing with the auditlogs or other logs)

adrinr: And we log all the installation, I guess it make sense that it's kept under system

Rory-Powell: My point is that it's potentially confusing to have /api/global/system, when "/api/system" is already a term we use to represent non-tenancy related endpoints. Everything under "/api/global" is supposed to be linked to a tenant. I think "/api/system/logs" makes the most sense? Given were it returns all unfiltered logs across all tenants (if tenancy was enabled)

packages/backend-core/src/environment.ts
```diff
@@ -154,13 +161,14 @@ const environment = {
...
     : false,
-  VERSION: findVersion(),
+  ...getPackageJsonFields(),
   DISABLE_PINO_LOGGER: process.env.DISABLE_PINO_LOGGER,
   _set(key: any, value: any) {
     process.env[key] = value
     // @ts-ignore
     environment[key] = value
   },
+  ROLLING_LOG_MAX_SIZE: process.env.ROLLING_LOG_MAX_SIZE || "100M",
```
mike12345567, line 171: is this 100M of log by default? I think we should reduce the size of this a bit for self hosters - thats quite a large log file (for uploading etc as well)

packages/backend-core/src/logging/system.ts
```diff
@@ -0,0 +1,45 @@
+import fs from "fs"
+import path from "path"
+import * as rfs from "rotating-file-stream"
+
+import env from "../environment"
+import { budibaseTempDir } from "../objectStore"
+
+const logsFileName = `budibase.logs`
```
mike12345567, line 8: I think `.log` is a more standard extension for log files

packages/backend-core/src/environment.ts
adrinr: What do you think would be a good size? 20MB?

mike12345567: I think `10MB` would be good!

mike12345567 (approved):
LGTM! Would just drop the default log size down for self hosters + I think anything too big will be hard to upload, we'll likely be able to find the issue in a 10MB log file anyway.

PR #11186, opened 2023-07-10
Show all users editing an app in the app list

aptkingston:
## Description
This PR updates the app list to show all users that are currently editing an app. It uses the newly improved avatar list and tooltips to provide the same high quality experience used elsewhere in the builder.

image
image

This works by using the metadata we store in redis for multi-dev collaboration. We can query this data out of redis efficiently by using bulk get queries, so this is reasonably lightweight and does not assault redis. In total it only uses 2 bulk get queries to get all users in all apps.


packages/server/src/api/controllers/application.ts
```diff
@@ -197,6 +199,30 @@ export async function fetch(ctx: UserCtx) {
     }
   }
 
+  // Get all builder sessions in each app
+  const sessions = await builderSocket?.getRoomSessions(appIds)
```
mike12345567, line 203: Ideally I would break this out into an SDK function, would be useful to have a `sdk.users.sessions.list(appIds)` - that way we wouldn't pollute the application controller as much!

packages/server/src/api/controllers/application.ts
```diff
@@ -197,6 +199,30 @@ export async function fetch(ctx: UserCtx) {
...
+  if (sessions?.length) {
+    let appSessionMap: { [key: string]: SocketSession[] } = {}
```
mike12345567, line 205: Typing for this which is a bit nicer: `Record<string, SocketSession[]>`

packages/server/src/api/controllers/application.ts
aptkingston: The only external call is the builder socket function, the rest is just local logic for enriching apps. `sdk.users.sessions.list(appIds)` would just be a straight proxy through to `builderSocket?.getRoomSessions(appIds)` which then wouldn't reduce the implementation size at all, unless we moved some of the local logic into it as well (i.e. generating the map) which would pull out a few extra lines, but then it's not really a `list` function... Any ideas?

mike12345567: Changed the function name, I think it would be better to do `const appsWithSessions = sdk.users.sessions.enrichApps(apps)`

packages/server/src/api/controllers/application.ts
aptkingston: Nice, TIL that this is just a nicer shorthand way of doing that.

packages/server/src/api/controllers/application.ts
aptkingston: Sounds good!

mike12345567 (approved):
Looks good!

PR #11193, opened 2023-07-11
Settings updates

aptkingston:
## Description
This PR improves the appearance of a few component settings types. I think this should now ensure that all settings types are consistent in their use of full width buttons and more descriptive text.

`event` type (actions) now has a full width button with the number of actions inside the button, rather than above:
image
image

`columns` type now has a full width button and shows how many columns are selected, rather than the static "Configure columns" text:
image
image

`fieldConfiguration` (used for form block and table block) type now has a full width button and shows how many fields are selected:
image
image


melohagan (approved):
Nice update - LGTM!

PR #11211, opened 2023-07-12
Remove view on deletion

adrinr:
## Description
Fixing the frontend to remove the views from the screens when view is deleted


aptkingston (approved):
LGTM!

Issue #8715, opened 2022-11-17
Cannot delete more than one selected row


melohagan:
**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
When using the **Delete Row** action, you cannot delete more than one row without an error being displayed.

**To Reproduce**
1. Create a table with many rows
2. Add a table that allows row selection
3. Add a button with the **Delete Row** action bound to `{{ Table.Selected rows }}`
4. Preview, and try to delete more than one row - see error

**Expected behavior**
Either it should delete all rows selected, or the error message should be explicit that only one selected row can be deleted.

**Screenshots**
<img width="1481" alt="Screenshot 2022-11-17 at 15 39 11" src="https://user-images.githubusercontent.com/101575380/202490578-a4b64408-3810-4192-8516-edd7db32c917.png">

  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**

**Additional context**

aptkingston:
The delete row action only deletes one row, but the error message is a bit cryptic because it thinks it's attempting to delete one row. The HBS binding for selected rows will evaluate to a comma separated string of all the selected IDs, so the shape is correct (a single string).
The servers error message just means it couldn't find a row to delete with that ID.

I'll convert this to a feature request to add a new button action for deleting multiple rows. The new action will have to try and be optimistic and parse strings to attempt to find comma separated IDs, and also handle real JS arrays.

PR #11236, opened 2023-07-14
Delete multiple rows with Delete Row button action 

deanhannigan:
## Description

Updated the `Delete Row` action to allow deleting multiple rows at once. 

You can now configure the action to delete rows using:
- A single row `_id` string. 
  - This was previously the only option available on the `Delete Row` action. This is still supported and no changes are required for any apps currently using this action.
- A single string of multiple row `_id` values separated by a comma (`,`)
  - An example of this can be seen using a Table components `Selected rows` binding e.g. `{{ My Table.Selected rows }}`
- A stringified row object: `"{_id: 'theId', tableId: 'myTable' ... }"`
  - This is typically seen when utilising Handlebar/Text bindings to fetch a row. e.g. `{{ MyProvider.rows.[0] }}`
- A row JS object: `{_id: 'theId', tableId: 'myTable' ... }`
  - Seen when utilising JavaScript bindings to work with rows e.g. `return $("MyProvider.rows")[0]")`
- An array of row objects 
  - You can also configure an entire collection row objects to delete at once.
  - You could pass all the rows currently selected in your table by the `Selected rows` binding e.g. `My Table.Selected rows`.

Addresses: 
- https://github.com/Budibase/budibase/issues/8715

## Screenshots

trimmed

New and existing `Delete Row` actions will now display as `Delete Rows`

Screenshot 2023-07-14 at 10 49 12


packages/server/src/api/controllers/row/index.ts
```diff
@@ -103,7 +105,22 @@ export async function destroy(ctx: any) {
   const inputs = ctx.request.body
   const tableId = utils.getTableId(ctx)
   let response, row
+
   if (inputs.rows) {
+    const targetRows = inputs.rows.map(
+      (row: { [key: string]: string | string }) => {
+        let processedRow = typeof row == "string" ? { _id: row } : row
```
Rory-Powell, line 112: I think some type definitions can help a lot here, as the exact expected format is quite hard to understand declaratively. 

i.e. we should add a DeleteRowsRequest api type that expresses the various forms described in the pr description. That can be integrated into the signature like ctx: Ctx<DeleteRowsRequest>. The ctx.body will then automatically infer that type and the usages in the controller will become type safe

Rory-Powell (changes requested):

packages/types/src/documents/app/row.ts
```diff
@@ -32,3 +32,13 @@ export interface Row extends Document {
...
+
+export interface DeleteRows {
+  rows: (Row | string)[]
+}
+
+export interface DeleteRow {
+  _id: string
+}
+
+export type DeleteRowRequest = DeleteRows | DeleteRow
```
Rory-Powell, line 44: These should be moved to the `src/api/web` package to indicate that it's a web request

packages/server/src/api/controllers/row/index.ts
```diff
@@ -100,25 +100,24 @@ export async function find(ctx: any) {
   })
 }
 
-export async function destroy(ctx: any) {
+export async function destroy(ctx: UserCtx<DeleteRowRequest>) {
   const appId = ctx.appId
-  const inputs = ctx.request.body
+  const inputs: DeleteRowRequest = ctx.request.body
```
Rory-Powell, line 105: Can do without the :`DeleteRowRequest` definition here, it'll automatically be inferred via the usage in the function signature

packages/server/src/api/controllers/row/index.ts
```diff
@@ -98,12 +100,26 @@ export async function find(ctx: any) {
...
+
   const tableId = utils.getTableId(ctx)
   let response, row
-  if (inputs.rows) {
+
+  if ("rows" in inputs && Array.isArray(inputs?.rows)) {
```
Rory-Powell, line 110: I think something that can help the readability here are some type guards

```ts
function isDeleteRows(input: any): input is DeleteRows {
  return input.rows !== undefined && Array.isArray(input.rows);
}

function isDeleteRow(input: any): input is DeleteRow {
  return input._id !== undefined;
}

function isStringRow(row: any): row is string {
  return typeof row === 'string';
}
```

We can then use like
```ts
// multiple delete request
if (isDeleteRows(inputs)) {
  ...
  let processedRow: Row = isStringRow(row) ? 
      { _id: row } : // row id request
      row // row object request
  ...
} 
// singular delete request
else if (isDeleteRow(inputs) {
  ...
}
```

I would advise staying away from things like `"rows" in inputs` as it completely ignores any typed fields that live in the target object and would be missed by refactors. 

The

Rory-Powell: The controller function is getting pretty huge at this point. In an effort to tame this it would help to break out some helper functions as well like:

```ts
function deleteRows(body: DeleteRows) { }
function deleteRow(body: DeleteRow) {}
```

And just life the contents of each section of the `if/else` into them respectively.

packages/server/src/api/controllers/row/index.ts
```diff
@@ -100,49 +107,83 @@ export async function find(ctx: any) {
...
-  const inputs = ctx.request.body
+function isDeleteRows(input: any): input is DeleteRows {
+  return input.rows !== undefined && Array.isArray(input.rows)
+}
 
+function isDeleteRow(input: any): input is DeleteRow {
+  return input._id !== undefined
+}
+
+async function processDeleteRowRequest(ctx: UserCtx<DeleteRowRequest>) {
```
Rory-Powell, line 118: Might be clearer as `processDeleteRowsRequest` as this is always going to be the plural form

packages/server/src/api/controllers/row/index.ts
```diff
@@ -100,49 +107,83 @@ export async function find(ctx: any) {
...
+  ctx.eventEmitter && ctx.eventEmitter.emitRow(`row:delete`, appId, resp.row)
+  gridSocket?.emitRowDeletion(ctx, resp.row._id)
+
+  return resp
+}
+
+export async function destroy(ctx: UserCtx<DeleteRowRequest>) {
+  let response, row
   ctx.status = 200
+
```
Rory-Powell, line 175: Usually best to put `ctx.status = 200` at the very end

Rory-Powell (approved):

packages/server/src/api/controllers/row/index.ts
```diff
@@ -98,35 +107,83 @@ export async function find(ctx: any) {
...
+  for (let row of rows) {
     ctx.eventEmitter && ctx.eventEmitter.emitRow(`row:delete`, appId, row)
     gridSocket?.emitRowDeletion(ctx, row._id)
   }
+
+  return rows
+}
+
+async function deleteRow(ctx: UserCtx<DeleteRowRequest>) {
```
Rory-Powell, line 157: Out of curiosity, is there any reason not to send the single delete through the bulk delete mechanism, to simplify and remove duplication?

deanhannigan: Its largely identical so it shouldn't be a problem to use the bulk delete for both cases. We would just have to make sure that we continue to return a row for the single automation delete step.

PR #11239, opened 2023-07-14
Moving files and functions

adrinr:
## Description
Extracting search functionality to be able to be used for the new views. No new functionallity has been added, just moving code around

Addresses: 
- [BUDI-7189](https://linear.app/budibase/issue/BUDI-7189/[views-v2]-new-views-object)


packages/server/src/sdk/app/rows/search.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function pickApi(tableId: any) {
+  if (isExternalTable(tableId)) {
+    return external
+  }
+  return internal
+}
+
+export async function search(tableId: string, ctx: Ctx) {
+  return pickApi(tableId).search(ctx)
```
mike12345567, line 25: I think I would prefer to keep any sort of `ctx` based functions out of the SDK if possible - these API pickers are really internal/external controllers, I would rather keep this part within the API controllers.

packages/server/src/sdk/app/rows/search/internal.ts
```diff
@@ -0,0 +1,246 @@
...
+import * as userController from "../../../../api/controllers/user"
+import * as inMemoryViews from "../../../../db/inMemoryView"
+import {
+  migrateToInMemoryView,
+  migrateToDesignView,
+  getFromDesignDoc,
+  getFromMemoryDoc,
+} from "../../../../api/controllers/view/utils"
+
+export async function search(ctx: Ctx) {
```
mike12345567, line 29: As above comments, again would prefer if we had common types for these SDK functions.

mike12345567 (changes requested):
Would be good to get the use of the `Ctx` out of the SDK where possible, I know it cannot be avoided in some places, but it would be nice if we had an `internal.search` and `external.search` that the controllers could call, which have common `RowSearchOpts` structures which can be passed to them, with a common response.

I know its a fair bit more refactoring, but I think it would be really worthwhile.

packages/server/src/sdk/app/rows/search.ts
adrinr: Yup, this is the idea. This PR is only moving files around, the next changes are getting rid of these CTX. I wanted to split the moving code vs changing functionality into different PRs, otherwise it would be massive

adrinr:
> Would be good to get the use of the `Ctx` out of the SDK where possible, I know it cannot be avoided in some places, but it would be nice if we had an `internal.search` and `external.search` that the controllers could call, which have common `RowSearchOpts` structures which can be passed to them, with a common response.
> 
> I know its a fair bit more refactoring, but I think it would be really worthwhile.

Yup, this is the idea. This PR is only moving files around, the next changes are getting rid of these CTX. I wanted to split the moving code vs changing functionality into different PRs, otherwise it would be massive.
Having proper params instead of context will be the only way to be able to reuse this functionallity for views, hence these refactor

packages/server/src/sdk/app/rows/search.ts
mike12345567: Ah I see no problem!

mike12345567:
> > Would be good to get the use of the `Ctx` out of the SDK where possible, I know it cannot be avoided in some places, but it would be nice if we had an `internal.search` and `external.search` that the controllers could call, which have common `RowSearchOpts` structures which can be passed to them, with a common response.
> > I know its a fair bit more refactoring, but I think it would be really worthwhile.
> 
> Yup, this is the idea. This PR is only moving files around, the next changes are getting rid of these CTX. I wanted to split the moving code vs changing functionality into different PRs, otherwise it would be massive. Having proper params instead of context will be the only way to be able to reuse this functionallity for views, hence these refactor

This makes a lot of sense! Sorry I didn't realise this was only the first part of it!

mike12345567 (approved):

PR #11247, opened 2023-07-17
New views crud

adrinr:
## Description
Adding scaffolding and tests for the new views. This PR will leave these endpoints ready to extend and to add extra view information


mike12345567 (approved):

packages/server/src/api/routes/view.ts
```diff
@@ -3,15 +3,37 @@ import * as viewController from "../controllers/view"
...
 import authorized from "../../middleware/authorized"
 import { paramResource } from "../../middleware/resourceId"
-import { permissions } from "@budibase/backend-core"
+import { DocumentType, SEPARATOR, permissions } from "@budibase/backend-core"
 
 const router: Router = new Router()
 
+router
+  .get(
+    "/api/views/v2",
```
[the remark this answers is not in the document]
Rory-Powell: +1

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -0,0 +1,40 @@
+import sdk from "../../../sdk"
+import { Ctx, ViewV2 } from "@budibase/types"
+
+export async function fetch(ctx: Ctx) {
```
Rory-Powell, line 4: Can we add the expected response type throughout the controller  i.e.`ctx: Ctx<void, ResponseType>`?
Ideally having `XRequest` and `XResponse` types, so we can decouple the web layer from the sdk

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  let table: Table
+
+  afterAll(setup.afterAll)
+
+  beforeAll(async () => {
+    await config.init()
+    table = await config.createTable(priceTable())
+  })
+
+  const saveView = async (view: ViewV2) => {
```
Rory-Powell, line 38: These api functions can be added to the api test clients, so we can use like `config.api.views.foo()`

packages/server/src/sdk/app/views/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    include_docs: true,
+  })
+
+  return response.rows.map(r => r.doc)
+}
+
+export async function get(viewId: string) {
+  const db = context.getAppDB()
+  const result = await db.get(viewId)
+  return result
```
Rory-Powell, line 39: I believe this will return `any`. Could we type it?

adrinr: Typed! I just realised our types are returning `get<T>(id?: string): Promise<T | any>`. This `any` will always ignore the T, so we should probably fix this. I will do it in another branch, as it might have plenty of changes

adrinr: https://github.com/Budibase/budibase/pull/11259

Rory-Powell: Nice one @adrinr. Initially I had meant to add typing to the functions inside this module only basically casting the `any`, but your other PR is a better way of handling this across the board 👍

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -31,10 +37,15 @@ export async function save(ctx: Ctx<ViewV2>) {
   }
 }
 
-export async function remove(ctx: Ctx) {
+export async function remove(ctx: Ctx<{}, {}>) {
```
Rory-Powell, line 40: I think it could be better to use `Ctx<void, void>` when there is no request / response body, to be the `{}` indicates empty rather than none. It would also be ok just to leave them out and they'll fallback to `any`

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -635,22 +635,25 @@ class TestConfiguration {
...
-      ...config,
-    }
-    return this._req(view, null, controllers.view.v2.save)
-  }
-
-  getViewV2(viewId: string): supertest.Test {
-    return this.request!.get(`/api/v2/views/${viewId}`)
-      .set(this.defaultHeaders())
-      .expect("Content-Type", /json/)
+  api = {
```
Rory-Powell, line 638: Sorry for being unclear, in my original suggestion around using test api clients I meant in similar fashion as we do in worker and account portal e.g. 

https://github.com/Budibase/budibase/blob/9dfd4aa811479a1dda72eff7ba2588d66b132277/packages/worker/src/tests/api/index.ts#L21-L39

The views code would be a good first use case. Meaning we can extract api specific concepts like views and their operations out of the `TestConfiguration` file directly. 

I would also be in favour of not using `controllers.view.v2.save` directly, as any use of this will bypass middlewares and request / response body parsing which is a main concern of the api unit tests.

adrinr: Are you thinking to use the `superttest` to do api calls, or just calling the sdk directly?

Rory-Powell: just the way we do it currently with `request.get`, `request.post` etc

Rory-Powell (approved):
LGTM!

PR #11249, opened 2023-07-17
Make sure a duplicate BlockComponent can't be registered

melohagan:
## Description
Duplicate *Create row* buttons were being registered which was causing a crash. 

I did have a go at changing the `registerComponent` code in **BlockComponent.svelte** to use *onMount* instead of being reactive, but this didn't handle re-ordering/rendering.

The solution I settled on was to ensure that when `registerBlockComponent` is called, any potential duplicates are removed first. 

Addresses: 
- https://linear.app/budibase/issue/BUDI-7280/selecting-and-de-selecting-search-fields-on-a-table-block-before

## Screenshots
fix

### DIFF
--- a/packages/client/src/components/Block.svelte
+++ b/packages/client/src/components/Block.svelte
@@ -13,6 +13,13 @@
     if (!structureLookupMap[parentId]) {
       structureLookupMap[parentId] = {}
     }
+    //If the instance already exists, delete the previous
+    for (const [key, value] of Object.entries(structureLookupMap[parentId])) {
+      if (value?._id === instance?._id) {
+        delete structureLookupMap[parentId][key]
+        break
+      }
+    }
     // Add this instance in this order, overwriting any existing instance in
     // this order in case of repeaters
     structureLookupMap[parentId][order] = instance


aptkingston:
Gonna take a quick look at this one locally myself! This looks to fix the symptom, but the cause will be something else - the unregistration of block components must not be working if we've got into this situation.

melohagan:
@aptkingston Appreciate it! I also noticed later on as well that when unchecking the selected *Search fields*, you could remove the incorrect field. I think this may also be something to do with the order when doing the component registration.

This was in in `develop` branch I noticed that to be clear - I haven't checked master but I imagine it's the same.

Reproduction:

1. Check *Region* search field
2. Check *Country* search field
3. Uncheck *Region* search field -> see that it incorrectly removes the *Country* search field

aptkingston:
> @aptkingston Appreciate it! I also noticed later on as well that when unchecking the selected _Search fields_, you could remove the incorrect field. I think this may also be something to do with the order when doing the component registration.
> 
> This was in in `develop` branch I noticed that to be clear - I haven't checked master but I imagine it's the same.
> 
> Reproduction:
> 
> 1. Check _Region_ search field
> 2. Check _Country_ search field
> 3. Uncheck _Region_ search field -> see that it incorrectly removes the _Country_ search field

Haha thought you might spot that one too! About to open another PR here to fix both issues 👍 I found the culprit!

PR #11250, opened 2023-07-17
Clean ctx dependencies from rows SDK

adrinr:
## Description
Cleaning CTX references from the rows SDK. No business logic changed, just removing the CTX dependencies after moving some files around in https://github.com/Budibase/budibase/pull/11239


mike12345567 (approved):
Looks great, massive improvement, thanks for doing this!

PR #11251, opened 2023-07-17
Fix block component keying

aptkingston:
## Description
Fixes a couple of issues with blocks:
- Blocks identified their child `BlockComponents` by using their `order` as part of an identifier. We depend on this indentifier to properly unregister them from their parent `Block` component instance when unmounting, but `order` can change at runtime, meaning we were not properly unregistering them. This meant that block ejection could sometimes contain duplicate components which would crash the builder. This PR updates keying to only consider the component ID, which is a constant.
- Search fields (used in `TableBlock` and `CardsBlock`) were not keyed by anything. This meant that when changing the setting in the builder, it appeared as though you sometimes removed the wrong field (even though a refresh would reveal the correct fields).

Addresses: 
- https://linear.app/budibase/issue/BUDI-7280/selecting-and-de-selecting-search-fields-on-a-table-block-before

### DIFF
--- a/packages/client/src/components/BlockComponent.svelte
+++ b/packages/client/src/components/BlockComponent.svelte
@@ -23,6 +23,8 @@
   // Create a fake component instance so that we can use the core Component
   // to render this part of the block, taking advantage of binding enrichment
   $: id = `${block.id}-${context ?? rand}`
+  $: parentId = $component?.id
+  $: inBuilder = $builderStore.inBuilder
   $: instance = {
     _component: `@budibase/standard-components/${type}`,
     _id: id,
@@ -38,14 +40,14 @@
   // Register this block component if we're inside the builder so it can be
   // ejected later
   $: {
-    if ($builderStore.inBuilder) {
-      block.registerComponent(id, order ?? 0, $component?.id, instance)
+    if (inBuilder) {
+      block.registerComponent(id, parentId, order ?? 0, instance)
     }
   }
 
   onDestroy(() => {
-    if ($builderStore.inBuilder) {
-      block.unregisterComponent(order ?? 0, $component?.id)
+    if (inBuilder) {
+      block.unregisterComponent(id, parentId)
     }
   })
 </script>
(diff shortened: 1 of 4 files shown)


packages/client/src/components/BlockComponent.svelte
```diff
@@ -23,6 +23,8 @@
   // Create a fake component instance so that we can use the core Component
   // to render this part of the block, taking advantage of binding enrichment
   $: id = `${block.id}-${context ?? rand}`
+  $: parentId = $component?.id
+  $: inBuilder = $builderStore.inBuilder
```
aptkingston, line 27: Just some dereferencing to primitives, to allow svelte to trigger the reactive registration below as infrequently as possible.

packages/client/src/components/app/blocks/CardsBlock.svelte
```diff
@@ -126,7 +126,7 @@
             order={1}
           >
             {#if enrichedSearchColumns?.length}
-              {#each enrichedSearchColumns as column, idx}
+              {#each enrichedSearchColumns as column, idx (column.name)}
```
melohagan, line 129: Just curious what this does / is for?

melohagan (approved):
Using the ID as a key instead of the order makes a lot of sense to me! Thanks for the quick fix 😄

packages/client/src/components/app/blocks/CardsBlock.svelte
aptkingston: This is the fix for my second bullet point! Adding a parameter in parenthesis after the `as` in a svelte `each` statement defines how you'd like to key your items - it's basically saying "when this value changes, fully remount the component".
By using the name property, I'm basically saying that when a certain component instance receives a different name, fully remount it because it's a different field. By default svelte will reuse component instances and simply pass in different props when the number of items in an `each` loop changes, but this is only bug-free if your components are fully atomic and can handle it.
Basically adding this ensures that you never get that issue where you see "stale" search fields. The fields you see will now always be the correct ones. 
Hope that explanation helps!

melohagan: Thanks for the explanation - makes sense!

PR #11254, opened 2023-07-17
Upgrade MongoDB version and quick fix for multi-tenancy mode without account portal

mike12345567:
## Description
Upgrading to latest version of MongoDB to attempt to solve the memory leak, as well as some minor changes to websocket integration to get the development environment to work without the account portal when running in multi-tenancy mode.


Rory-Powell (approved):
LGTM

PR #11255, opened 2023-07-17
Skippable tours

deanhannigan:
## Description

Tours within Budibase are now skippable.

- A `Skip` button has been added to the tour popover and will display if an `onSkip` function has been configured. This replaces the `Previous` button.
- The main onboarding tour has been configured with an `onSkip` behaviour and will redirect to the data section if skipped.
- The feature onboarding tour remains unchanged as it only contains one step for the moment.

Addresses: 
- BUDI-7196

## Screenshots

Screenshot 2023-07-17 at 17 32 08


aptkingston (approved):
LGTM! Nice one 👌  And well remembered to add the skip event in there too.

PR #11259, opened 2023-07-18
Chore - Type couchdb get

adrinr:
## Description
Typing the `db.get` response to respect the T type. Currently we return `T | any`, that is the same as `any`.
Typed the few breaking changes to `<any>` as there are not existing types for these use cases

Dependant: https://github.com/Budibase/budibase-pro/pull/205

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -433,7 +433,7 @@ export async function destroy(ctx: UserCtx) {
 
 export async function find(ctx: UserCtx) {
   const db = context.getAppDB()
-  const datasource = await db.get(ctx.params.datasourceId)
+  const datasource = await sdk.datasources.get(ctx.params.datasourceId)
   ctx.body = await sdk.datasources.removeSecretSingle(datasource)
 }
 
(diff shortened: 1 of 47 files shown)


packages/server/src/api/controllers/datasource.ts
```diff
@@ -433,7 +433,7 @@ export async function destroy(ctx: UserCtx) {
 
 export async function find(ctx: UserCtx) {
   const db = context.getAppDB()
-  const datasource = await db.get(ctx.params.datasourceId)
+  const datasource = await sdk.datasources.get(ctx.params.datasourceId)
```
Rory-Powell, line 436: Can remove the `db` above now?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -449,7 +449,7 @@ export async function query(ctx: UserCtx) {
 
 export async function getExternalSchema(ctx: UserCtx) {
   const db = context.getAppDB()
-  const datasource = await db.get(ctx.params.datasourceId)
+  const datasource = await sdk.datasources.get(ctx.params.datasourceId)
```
Rory-Powell, line 452: Same as above

Rory-Powell (approved):
Nice work on this. Removing the fallback `any` is probably the right choice, it'll make us be more strict with properly typing db reads which I'm all for

adrinr:
> Nice work on this. Removing the fallback `any` is probably the right choice, it'll make us be more strict with properly typing db reads which I'm all for

This will eventually "force" us to move the actual db.gets into the sdk or db functions, where we will cast the objects. Doing so we will have the db layer better abstracted, instead of having db.get around all the code

Issue #10376, opened 2023-04-21
[BUDI-6902] Move 'When Filter Empty' option into Filter Drawer


melohagan:
Currently there is a *When Filter Empty* option for the **Query Rows** action in automations.

In my opinion it would make more sense to put this into the Filter Drawer so that it can be used and be consistent in the Design section as well.

![Screenshot 2023-04-21 at 09.47.07.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/46ccaa6b-8e7a-44d3-ba4e-8b5a0d7b6c9f/270c551a-ef46-4cbd-b494-f34a62f3cb97)

Move the *When Filter Empty* dropdown into:

![Screenshot 2023-04-21 at 09.47.34.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/c522eef6-bf86-4f68-a8c3-106f8fbcbcea/c9ec21f4-8bf4-4d87-bf72-7996f5cb32f7)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6902](https://linear.app/budibase/issue/BUDI-6902/move-when-filter-empty-option-into-filter-drawer)</sub>

melohagan:
Also noticed a bug that is relevant to this:

---

**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.8.10

**Describe the bug**
When filters are blank, *Match all* will return all of the rows (even though technically the filter condition is not met) whereas *Match any* returns no rows (because none of the filter conditions are met).

**To Reproduce**
Steps to reproduce the behavior:
1. Add a filter to a Data provider, **Match all** with two filters bound to form fields
2. See that by default all rows are displayed
3. Change the filter to **Match any**
4. See that by default no rows are displayed

**Expected behavior**
When filters are bound to form fields, and none of the field values are set, I would expect the result from **Match all** and **Match any** to be the same, ideally showing all rows because that seems more useful.

Issue #7508, opened 2022-08-30
Add option to return no rows for empty filters to data providers


aptkingston:
**Describe the feature request**
Discussed in #7502.

The automation section recently got updated to support returning no rows, rather than all rows, for an empty filter expression. We should add this as a data provider setting to allow them to also support this feature.

It's useful for cases where rendering "all rows" would return so much data that it would crash/freeze the browser - easily currently highlighted by doing something like filtering a custom query on a URL param and previewing the screen.

Issue #7502, opened 2022-08-30
Handle preview in the builder better when there is no URL variable present for a filter (all data being returned)


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/7443

<div type='discussions-op-text'>

<sup>Originally posted by **R2bEEaton** August 24, 2022</sup>
I have some complicated pages on my app, and a couple of them require data sources with high limits. In reality though, a normal user would enter the page and the source would be filtered down to a manageable limit.

However, in the builder, the filters don't work because the URL variable isn't set, for example. In this case, the data provider returns all of the data which freezes the client and makes editing nigh on impossible sometimes.

I found four solutions, none of which are ideal:
1. Making the change within 2-3 seconds before the builder loads the preview.
2. Change the data source to be limited (which you have to do within 2-3 seconds before the builder loads the preview).
3. Wait minutes for the page to fully load and become responsive.
4. Not make the edit you want because it's a pain to deal with.

My proposed solution is a global toggle somewhere in the builder where you can disable the site preview. _Another option would be to be able to set URL variables within the builder, for example, or a default value for data source filters. This might be preferable for UX purposes._</div>

melohagan:
Also discussed here: https://github.com/Budibase/budibase/discussions/7479

aptkingston:
Yea being able to set a value for URL vars for the builder preview would be nice.

In the meantime there are multiple solutions:
- The data provider limit should not be so large that it crashes the page, even if we assume a URL variable will always be set.  I would strongly recommend he sets a sensible data provider limit.
- Use conditional UI to change the filters if the URL variable is not set (e.g. just return 1 row)
- When configuring the filter, use a fallback value. So the filter would be `URL.foo or "bar"`

R2bEEaton:
> Yea being able to set a value for URL vars for the builder preview would be nice.

Yeah!

> In the meantime there are multiple solutions:
> 
> * The data provider limit should not be so large that it crashes the page, even if we assume a URL variable will always be set.  I would strongly recommend he sets a sensible data provider limit.
> * Use conditional UI to change the filters if the URL variable is not set (e.g. just return 1 row)
> * When configuring the filter, use a fallback value. So the filter would be `URL.foo or "bar"`

[...]

PR #11262, opened 2023-07-18
Move when filter empty option into filter drawer

melohagan:
## Description

- [x] Looking into another issue in develop relating to the the onEmptyFilter value in automations
- [x] Need to do more testing around automations

---

There are cases when you want to decide if all or no rows should be returned when your filter binding values are empty (similar to in automations).

The *When filter empty* setting has been added into the filter drawer.

The UI in the filter drawer may be a little cluttered now? I am open to layout suggestions! 

Addresses: 
- https://github.com/Budibase/budibase/issues/10376
- https://github.com/Budibase/budibase/issues/7508
- https://github.com/Budibase/budibase/issues/7502

## App Export
[Filter Table Block with Options Pickers-export-1689612248908.tar.gz](https://github.com/Budibase/budibase/files/12079281/Filter.Table.Block.with.Options.Pickers-export-1689612248908.tar.gz)

## Screenshots
Screenshot 2023-07-18 at 11 20 03

https://github.com/Budibase/budibase/assets/101575380/fa04148a-936b-4679-8740-b0338728cd29

## Documentation


packages/shared-core/src/filters.ts
```diff
@@ -111,7 +117,8 @@ const removeKeyNumbering = (key: string) => {
 }
 
 type Filter = {
-  operator: keyof Query
+  operator: keyof Omit<Query, "onEmptyFilter">
+  onEmptyFilter: EmptyFilterOption
```
aptkingston, line 121: This should be optional - it'll only exist on at most one filter.

packages/shared-core/src/filters.ts
```diff
@@ -160,6 +167,7 @@ type QueryFields = {
 
 type QueryConfig = {
   allOr?: boolean
+  onEmptyFilter?: EmptyFilterOption
```
aptkingston, line 170: I don't think we need this in here - this type is mixed with `QueryFields` to represent the available options for `operator`, which is how we pass up the flag for whether we have an AND or and OR query, i.e. `operator = "allOr"` indicates that it's an OR.

For this new field, we set it top level in the filter object (hence it's presence in the `Filter` type), but we don't need to add it into the `QueryConfig` type since we will never send up `operator = "onEmptyFilter"`.

packages/backend-core/src/db/lucene.ts
```diff
@@ -463,6 +474,13 @@ export class QueryBuilder<T> {
       allOr = false
       build({ tableId }, equal)
     }
+    if (allFiltersEmpty) {
+      if (this.#query.onEmptyFilter === EmptyFilterOption.RETURN_NONE) {
```
aptkingston, line 478: Nice work making sure that we're comparing to `NONE`, since that will account for all the existing saved filters which have an `undefined` `onEmptyFilter` filter, meaning we'll keep the same value as before. Would be easy to overlook this!

packages/backend-core/src/db/tests/lucene.spec.ts
```diff
@@ -156,6 +156,76 @@ describe("lucene", () => {
...
+        builder.addNotEqual("number", undefined)
+        builder.addOneOf("number", null)
+        builder.addContains("array", undefined)
+        builder.addNotContains("array", null)
+        builder.addContainsAny("array", null)
+
+        const resp = await builder.run()
+        expect(resp.rows.length).toBe(1)
+      })
+    })
```
aptkingston, line 227: Great job adding some test cases!

aptkingston (commented):
Great work @melohagan! I like how you've considered backwards compatibility here and also added new test cases.

Just raised a couple of queries about the types, but all the core logic looks good to me.

packages/shared-core/src/filters.ts
melohagan: Made optional

packages/shared-core/src/filters.ts
melohagan: As discussed this was needed in the end. I have pushed up further code for handling SQL tables, and in the case of custom queries the empty filter setting is hidden.

melohagan:
@aptkingston I have removed the *onEmptyFilter* property from the automation step schema.

I checked backwards compatibility, and whatever option was selected before the update will be respected going forward (both *all* and *none* tested).
Furthermore, you can still use the *When Filter Empty* setting to update existing automation query rows behaviour. 

The only thing that's a little strange is that whenever you set the *When filter empty* behaviour inside the `Define filters` drawer, then that choice will always override whatever is selected in the *When Filter Empty* dropdown of the automation step block.

All new Query Rows blocks however will not have this old field at all.

aptkingston:
> that choice will always override whatever is selected in the When Filter Empty dropdown of the automation step block.

Should we not then remove this setting from the UI for that automation step? It will still be saved in the automation definition for existing automations, so if we just keep the backend logic as you currently have it now, but remove it from the UI, then backwards compatibility should still work while allowing users to customise the behaviour going forwards using the new location inside the filter drawer.

melohagan:
> Should we not then remove this setting from the UI for that automation step? It will still be saved in the automation definition for existing automations, so if we just keep the backend logic as you currently have it now, but remove it from the UI, then backwards compatibility should still work while allowing users to customise the behaviour going forwards using the new location inside the filter drawer.

The setting has been removed from the `queryRows.ts` schema, so the UI won't reflect the old dropdown going forward.
The tricky thing is that the `AutomationBlockSetup` renders components in a generic way, so you'd need to add code to explicitly check the name of the property and not render it - doesn't seem like the nicest solution.

Here's the code that adds the setting in the GUI to outline what I mean:

```js
{#if value.type === "string" && value.enum && canShowField(key, value)}
          <Select
            on:change={e => onChange(e, key)}
            value={inputData[key]}
            placeholder={false}
            options={value.enum}
            getOptionLabel={(x, idx) => (value.pretty ? value.pretty[idx] : x)}
          />
```

aptkingston (approved):

PR #11263, opened 2023-07-18
Implement search views

adrinr:
## Description
Implement search views

Dependant: https://github.com/Budibase/budibase-pro/pull/206

### DIFF
--- a/packages/server/src/api/controllers/row/index.ts
+++ b/packages/server/src/api/controllers/row/index.ts
@@ -146,6 +146,20 @@ export async function search(ctx: any) {
   })
 }
 
+export async function searchView(ctx: any) {
+  const { viewId } = ctx.params
+  const view = await sdk.views.get(viewId)
+  const tableId = view.tableId
+
+  ctx.status = 200
+  ctx.body = await quotas.addQuery(
+    () => sdk.rows.search({ tableId, query: {} }),
+    {
+      datasourceId: tableId,
+    }
+  )
+}
+
 export async function validate(ctx: Ctx) {
   const tableId = utils.getTableId(ctx)
   // external tables are hard to validate currently
--- a/packages/server/src/api/controllers/row/index.ts
+++ b/packages/server/src/api/controllers/row/index.ts
@@ -2,7 +2,7 @@ import { quotas } from "@budibase/pro"
 import * as internal from "./internal"
 import * as external from "./external"
 import { isExternalTable } from "../../../integrations/utils"
-import { Ctx } from "@budibase/types"
+import { Ctx, SearchResponse } from "@budibase/types"
 import * as utils from "./utils"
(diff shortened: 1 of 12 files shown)


packages/server/src/api/controllers/row/index.ts
```diff
@@ -146,6 +146,23 @@ export async function search(ctx: any) {
...
+export async function searchView(ctx: Ctx<void, SearchResponse>) {
+  const { viewId } = ctx.params
+  const view = await sdk.views.get(viewId)
+  if (!view) {
+    ctx.throw(404)
+  }
+  const tableId = view.tableId
+
+  ctx.status = 200
+  ctx.body = await quotas.addQuery(
```
Rory-Powell, line 158: Maybe should try the body assignment first. In case the query fails in future if we add a limit. I'm fairly certain the exception handling would result in a 400 for that type of exception, but I couldn't say for certain. Having the status after the successful query will play it safe.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -685,4 +685,47 @@ describe("/rows", () => {
...
+
+    it("returns table rows from view", async () => {
+      const table = await config.createTable(priceTable())
+      const rows = []
+      for (let i = 0; i < 10; i++)
+        rows.push(await config.createRow({ tableId: table._id }))
+
+      const createViewResponse = await config.api.viewV2.create()
+
+      const response = await request
```
Rory-Powell, line 719: Would be good to get this request into the new views test client

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -685,4 +685,47 @@ describe("/rows", () => {
...
+            },
+          },
+        },
+      }
+    }
+
+    it("returns table rows from view", async () => {
+      const table = await config.createTable(priceTable())
+      const rows = []
+      for (let i = 0; i < 10; i++)
```
Rory-Powell, line 714: Usually always add {} to any if. Have seen some cases before where someone accidentally adds a second statement inside an unwrapped if that had side effects

packages/server/src/api/controllers/row/index.ts
adrinr: I just tested it, any error will ignore the 200: 
image

packages/server/src/api/routes/tests/row.spec.ts
adrinr: In this case, we are testing this endpoint, for this I use it explicitly instead of using the helpers. Don't you think it's better than using the helpers for it?

adrinr: Changed as discussed

Rory-Powell (approved):
LGTM

PR #11265, opened 2023-07-18
Implement querying views

adrinr:
## Description
Implement querying and sorting views

### DIFF
--- a/packages/server/src/api/routes/tests/row.spec.ts
+++ b/packages/server/src/api/routes/tests/row.spec.ts
@@ -790,7 +790,37 @@ describe("/rows", () => {
         },
         ["Danny", "Charly", "Bob", "Alice"],
       ],
-    ])("allow sorting", async (sortParams, expected) => {
+      [
+        {
+          field: "age",
+          order: SortOrder.ASCENDING,
+          type: SortType.number,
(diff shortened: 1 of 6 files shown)


packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -727,5 +727,131 @@ describe("/rows", () => {
...
+            name: generator.name(),
+            age: 40,
+          })
+        )
+
+      const createViewResponse = await config.api.viewV2.create({
+        query: { equal: { age: 40 } },
+      })
+
+      const response = await request
```
Rory-Powell, line 754: Can go into views test client and some similar cases below

packages/server/src/sdk/app/rows/search.ts
```diff
@@ -11,8 +11,8 @@ export interface SearchParams {
   bookmark?: string
   limit?: number
   sort?: string
-  sortOrder?: string
-  sortType?: string
+  sortOrder?: "ascending" | "descending"
```
Rory-Powell, line 14: Can use the SortOrder type?

Rory-Powell (approved):

PR #11266, opened 2023-07-18
New Design Section Nav Component

Ghrehh:
## Description
Adds the new navigation settings panel to the design section as outlined in the ticket. Also implements some new components that will be used by other changes to these panels in future.

Addresses: 
https://linear.app/budibase/issue/BUDI-7164/design-section-navigation-component


aptkingston:
Nice work!

Just wondering, could you clarify why you opted for a new `Pane` component with a built in custom store rather than using the existing `Panel` component? We already had all the building blocks to make something like this - for example the `ComponentSettingsPanel` is extremely similar in the sense that it also has different tab options for the content of the panel.

There are a few design differences between the existing implementation and this new one (both the panel style and the action button tab style) so we'll need to basically maintain both as duplicates of each other, unless we choose to use one for everything.

Not saying that it's a bad decision, but just wondering what your reasoning was for creating duplicate components instead of reusing (and updating if required) the existing one, in case there's a good reason that I haven't thought of.

My other question was just around the actual navigation component - it doesn't seem to have been added to the tree, but maybe you were planning on doing that in a separate PR?

Ghrehh:
@aptkingston Nope, your correct, that's a much better shout, didn't think to use it. I'll refactor this.

Not sure I understand the second point, the component is visible currently, it's just on the right hand side of the screen, the existing navigation component is still present too, but that'll be replaced in another PR.

aptkingston:
> @aptkingston Nope, your correct, that's a much better shout, didn't think to use it. I'll refactor this.
> 
> Not sure I understand the second point, the component is visible currently, it's just on the right hand side of the screen, the existing navigation component is still present too, but that'll be replaced in another PR.

Cool, no worries.

For the second point I mean the actual "component" in the tree, which is the new way of accessing the navigation settings instead of the existing small icon in the side nav. It would basically just appear as a fake app component, like in this screenshot from the designs:
image

I assumed it would make sense to include that in this PR as well, but maybe you were planning on adding that later.

Issue #11158, opened 2023-07-06
[BUDI-7260] [Builder] Bindings drawer HBS {{ /if }} syntax highlight issue


poirazis:
Self hosted 2.8.1

The HBS expression works, but is not correctly highlighted.

First Screenshot is from editing the HBS on a Formula Column

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/e79df6b5-9d77-43ce-ad98-8b4b8cb13200/99a72d09-82a0-4f05-8c17-9e8ebbb09423?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9lNzlkZjZiNS05ZDc3LTQzY2UtYWQ5OC04YjRiOGNiMTMyMDAvOTlhNzJkMDktODJhMC00ZjA1LThjMTctOWU4ZWJiYjA5NDIzIiwiaWF0IjoxNjg5Njc0MzA3LCJleHAiOjE2ODk3NjA3MDd9.RskJ5M2gnOLyxYO2WoV0K08r0lSk4Iqk4528z0D_32Q)

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/a47f657a-4ac9-40e0-b41a-af8d0404e93c/810caffe-aaf8-4db7-beaa-4556d8b3c522?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9hNDdmNjU3YS00YWM5LTQwZTAtYjQxYS1hZjhkMDQwNGU5M2MvODEwY2FmZmUtYWFmOC00ZGI3LWJlYWEtNDU1NmQ4YjNjNTIyIiwiaWF0IjoxNjg5Njc0MzA3LCJleHAiOjE2ODk3NjA3MDd9.mDtvo37RMysKrtHlzjRuyxwy6fKk8d0Oj8GTq0HfjPs)

[Source image](https://github.com/Budibase/budibase/assets/80512170/fae85dcf-558e-43ad-a377-f01469c96d5c)

Second Screenshot by @melohagan from autogenerated screen bindings

[...]

Issue #11084, opened 2023-06-30
[BUDI-7248] CTRL + Z reverts actions outside the focused component


kasgithub8:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.7.14
  - App Version: 2.7.14

**Describe the bug**
Using CTRL + Z, for example, in a paragraph while writing reverts actions outside the paragraph field.

**To Reproduce**
1. Add a component where u can write in it
2. move the component, for example, in a container
3. Open and write in the component
4. Press CTRL + Z, moves the component back to it's position before the container

**Expected behavior**
Reverts just the actions done inside the open component

<sub>[BUDI-7248](https://linear.app/budibase/issue/BUDI-7248/ctrl-z-reverts-actions-outside-the-focused-component)</sub>

aptkingston:
Hey @kasgithub8. Ctrl+Z maintains as long a history as possible (which is a good thing) but that means that it can revert any change. If we only reverted actions for the selected component it would severely limit it's usefulness, because as soon as you did something like delete a component (which would select a new one) you'd never be able to undo that.

I think there's probably a separately issue here for just disabling our undo/redo functionality while editing bindings, and let the browser default ctrl+z behaviour take over.

kasgithub8:
@aptkingston thanks for the response. 
I understand.
I have another quick question tho, is it normal that I can't scroll in, for example, a embed component?
![image](https://github.com/Budibase/budibase/assets/95284278/90f3c5cb-6833-44f3-a4c6-be972e2b5650)
Im pretty sure that there used to be a scrollbar.

Issue #10561, opened 2023-05-12
[BUDI-6987] Unexpected undo when editing inputs within the design view.


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.6.8
  * App Version:  2.6.8

**Describe the bug**
Undo has issues whenever editing content within an input on the design view. E.g. editing a JS binding, if you hit the shortcut key while working with these it will occasionally undo a random component change/amend or even delete it. Instead of being fixed to the current area you are working in.

**To Reproduce**

1. Go to the design view of any app.
2. Add some components
3. Edit bindable areas for example: default value on a input field
4. Write some text and hit the shortcut key for undo
5. See some components being removed as well as the text you are working on.

**Expected behavior**
Prevent app undos whenever focused inside a bindable areas. This will prevent loss of progress and also prevent frustrating circumstances where certain previously setup features are broken because components have been changed or removed. 

**Desktop (please complete the following information):**

* OS: Mac OS
* Browser Brave
* Version **v1.51.114**

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6987](https://linear.app/budibase/issue/BUDI-6987/unexpected-undo-when-editing-inputs-within-the-design-view)</sub>

aptkingston:
Just to clarify we don't offer any sort of undo functionality while editing text or bindings - if you press our undo shortcut at any time then it'll undo your latest "budibase" change. I imagine this is just doubling up with some browser default functionality which is also undoing some typing with the same shortcut.

I think your suggestion of disabling undo while an input is focused should probably work just fine, because I can't think of anywhere in the design section that you might still want an undo while something is focused. We could also try to swallow key press events that originate from inside text inputs but I reckon checking globally if something is focused should work.

PR #11267, opened 2023-07-18
General fixes for bindings and the undo/redo behaviour

deanhannigan:
## Description

- Builder Undo/Redo suspended when focused on an input in the builder. This will restrict the Undo/Redo behaviour to that of the focused input and stop unintended screen changes.
- Platform issue - Redo key binding bug where the expected key check was in the wrong case.
- Dual `Cancel` buttons in the CSS binding drawer.
- Updated regex pattern for the handlebars formatting in the CodeEditor.

Addresses: 
- https://github.com/Budibase/budibase/issues/11158
- https://github.com/Budibase/budibase/issues/11084
- https://github.com/Budibase/budibase/issues/10561

### DIFF
--- a/packages/bbui/src/Drawer/Drawer.svelte
+++ b/packages/bbui/src/Drawer/Drawer.svelte
@@ -30,6 +30,7 @@
   setContext("drawer-actions", {
     hide,
     show,
+    headless,
   })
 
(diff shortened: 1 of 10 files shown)


aptkingston (approved):
LGTM!

Issue #10543, opened 2023-05-11
[BUDI-7070] Helm chart uses wrong image tags


mauriziobrilli:
**Checklist**
- [*] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: k8s
  - Budibase Version: 2.6.9

**Describe the bug**
Image tags in helm chart lack the initial 'v'. E.g. budibase/budibase:2.6.9 instead of budibase/budibase:v2.6.9. I also checked version 2.6.8, same situation

**To Reproduce**
1. helm install --create-namespace --namespace budibase budibase budibase/budibase
2. kubectl get pods -n budibase
3. 

>     Output:
    NAME                              READY   STATUS             RESTARTS   AGE
    minio-service-785c7bf9f8-9bzw5    1/1     Running            0          2m22s
    redis-service-5dcbf4db55-cb2tv    1/1     Running            0          2m22s
    buditest-couchdb-0                2/2     Running            0          2m22s
    buditest-couchdb-1                2/2     Running            0          2m22s
    buditest-couchdb-2                2/2     Running            0          2m22s
    worker-service-5f7f9cc457-4bhvl   0/1     ImagePullBackOff   0          2m22s
    app-service-75f55fb46-9mdqb       0/1     ImagePullBackOff   0          2m22s
    proxy-service-7796b4644b-svwck    0/1     ImagePullBackOff   0          2m22s

4. describe pods --namespace budibase - for each budibase container Events report failure to pull image (e.g. Pulling image "budibase/proxy:2.6.9")

**Expected behavior**
All services running

<sub>[BUDI-7070](https://linear.app/budibase/issue/BUDI-7070/helm-chart-uses-wrong-image-tags)</sub>

ecarlotti:
I can confirm that bug, I had the same issue. The helm chart is using images names missing the "v" before the version number, e.g. "budibase/proxy:2.6.9" instead of the correct "budibase/proxy:v2.6.9" - That happens on the three services mentioned by @mauriziobrilli

jamesread:
Can confirm also - the helm release is currently broken, and I had to manually patch the deployments to add the "v" in the "image: " spec. 

I was going to PR this as a quick fix, but it seems  that the image version is populated from the chart version (annoyingly), so I'm uncertain if the original authors want to add the "v" to the chart version too, or the "v" just to the image version. Adding the "v" to the chart version too may break some helm things / auto-upgrades etc.

[...]

PR #11274, opened 2023-07-19
Fix/k8s fixes

shogunpurple:
## Description
FIxing some issues with our helm chart. Namely the `v<version>` tags that we use, where the `v` is pointless. From here on in we will just push the version number, and this will be reflected in all our versions and helm charts.

Also reduced the number of couch replicas to 1 in our default helm chart. It's pointless for almost all users to run 3 nodes - and it introduces additional complexity that just isn't required when getting started with our K8S offering.

Addresses: 
- https://github.com/Budibase/budibase/issues/10543

## Documentation


.github/workflows/release-selfhost.yml
```diff
@@ -108,8 +108,8 @@ jobs:
       - name: Perform Github Release
         uses: softprops/action-gh-release@v1
         with:
-          name: v${{ env.RELEASE_VERSION }}
-          tag_name: v${{ env.RELEASE_VERSION }}
+          name: ${{ env.RELEASE_VERSION }}
+          tag_name: ${{ env.RELEASE_VERSION }}
```
Rory-Powell, line 112: Thank's for having a look at this. I was just about to take it on but you've saved me the trouble 👏 
Should we update the default tag behaviour to remove the `v` also then?
https://github.com/Budibase/budibase/blob/a0202a7e48b38b690dd2b013cb3dac367cbe5c3d/scripts/versionCommit.sh#L16

Rory-Powell: I think the github tag actually determines the image tag, so we will have to update this

PR #11279, opened 2023-07-19
Store views2 in tables

adrinr:
## Description
Refactoring new views to be stored in the table object instead of in its own document


packages/server/src/api/routes/view.ts
```diff
@@ -3,23 +3,13 @@ import * as viewController from "../controllers/view"
...
-    viewController.v2.fetch
-  )
-  .get(
-    `/api/v2/views/:viewId`,
-    authorized(permissions.BUILDER),
-    viewController.v2.find
-  )
   .post(
-    "/api/v2/views",
+    "/api/v2/views/:tableId",
```
mike12345567, line 12: should this be `:viewId`

packages/server/src/db/utils.ts
```diff
@@ -277,9 +277,13 @@ export function getMultiIDParams(ids: string[]) {
  * @returns {string} The new view ID which the view doc can be stored under.
  */
 export function generateViewID(tableId: string) {
-  return `${viewIDPrefix(tableId)}${newid()}`
+  return `${tableId}${SEPARATOR}${newid()}`
 }
 
-export function viewIDPrefix(tableId: string) {
-  return `${DocumentType.VIEW}${SEPARATOR}${tableId}${SEPARATOR}`
+export function extractViewInfoFromId(viewId: string) {
```
mike12345567, line 283: I generally prefer to whole uppercase `ID` just so that it stands out

mike12345567 (approved):

packages/server/src/api/routes/view.ts
adrinr: This was meant to be table, as it's creating the view on this table. But it could be inferred from the body

PR #11286, opened 2023-07-20
Return schema from table endpoint

adrinr:
## Description
Add a new endpoint that will return the schema of the new views. It will default to the table schema, or using the defined columns if defined


mike12345567 (approved):

mike12345567 (approved):
Great type improvements, very nice!

mike12345567 (approved):

PR #11293, opened 2023-07-20
Fix release and prerelease tag triggers

Rory-Powell:
## Description
Follow up fix from tag naming change in: https://github.com/Budibase/budibase/pull/11274


shogunpurple (approved):
LGTM!

PR #11297, opened 2023-07-20
Fix duplicate dynamic variables when creating new REST queries

mike12345567:
## Description
Fix for dynamic variables being duplicated when creating new REST queries - also moved some stuff to backend SDK to make a bit re-usable, allowing backend to detect duplicate dynamic variables and error.

Initially did the refactoring as I was looking into the queries, decided to keep it as it was an improvement that was needed anyway.

The duplicate dynamic variables for different queries was causing some to fail.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7304/rest-queries-using-dynamic-variables-returning-400-connection-is

### DIFF
--- a/packages/types/src/documents/app/datasource.ts
+++ b/packages/types/src/documents/app/datasource.ts
@@ -7,9 +7,7 @@ export interface Datasource extends Document {
   name?: string
   source: SourceName
   // the config is defined by the schema
-  config?: {
-    [key: string]: string | number | boolean | any[]
-  }
+  config?: Record<string, any>
   plus?: boolean
   entities?: {
     [key: string]: Table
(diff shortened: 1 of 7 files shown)


packages/types/src/documents/app/datasource.ts
```diff
@@ -7,9 +7,7 @@ export interface Datasource extends Document {
   name?: string
   source: SourceName
   // the config is defined by the schema
-  config?: {
-    [key: string]: string | number | boolean | any[]
-  }
+  config?: Record<string, any>
```
adrinr, line 10: Why did we lose any?

adrinr (approved):
LGTM

packages/types/src/documents/app/datasource.ts
mike12345567: The typing of this was causing issues, it wasn't *really* accurate the type before, it was just a smattering of stuff, but in reality it could be absolutely anything, it depends on the datasource schema what is in the config, its basically impossible to type per datasource - each datasource implementation has an actual definition of the config.

PR #11302, opened 2023-07-20
demote some app metadata related bb-alert messages

shogunpurple:
## Description
App metadata not being present is a valid state (when the app is deleted). Due to this we get a bit of noise in our logs from the `bb-alert` message. I've updated the call not to throw when the app metadata isn't found, and accounted for that state in the calling functions.


packages/backend-core/src/cache/appMetadata.ts
```diff
@@ -61,11 +61,8 @@ export async function getAppMetadata(appId: string) {
     }
     await client.store(appId, metadata, expiry)
   }
-  // we've stored in the cache an object to tell us that it is currently invalid
-  if (isInvalid(metadata)) {
-    throw { status: 404, message: "No app metadata found" }
-  }
-  return metadata as App
+
+  return metadata as App & { state: AppState }
```
Rory-Powell, line 65: I think `state` here is a bit confusing, in that it can only have a single value. 
The naming is a bit confusing as well because what does "invalid" actually mean. 

I prefer how stripe sdk handles this, where the function would either return `App | DeletedApp`. Where the former would look like:

```ts
interface DeletedApp { 
   appId: string,
   /**
    * Always true for a deleted object
    */
   deleted: true
}
```
and the latter
```ts
interface App extends Document {
   // ...other props
   deleted?: boolean
}
```

Rory-Powell: If we were to keep an implementation closer to the current model, I think it should be more like:
```ts
interface DeletedApp {
   state: AppState
}

export async function getAppMetadata(appId: string): Promise<App | DeletedApp> {
   ...
   if (isInvalid(metadata)) {
      return { state: AppState.INVALID }
   }
   ...
   return metadata
}
```

Preference would be on a deleted flag because it's a lot more clear what's going in. In either implementation I don't think we should return `App & { state: AppState }` as that's not what we return, it'll be one or the other.

shogunpurple: Good suggestion - we wouldn't need to cast it then with `as`. Will update to a dedicated `interface` and union type

packages/backend-core/src/db/utils.ts
```diff
@@ -126,7 +128,11 @@ export async function getAppsByIDs(appIds: string[]) {
   )
   // have to list the apps which exist, some may have been deleted
   return settled
-    .filter(promise => promise.status === "fulfilled")
+    .filter(
+      promise =>
+        promise.status === "fulfilled" &&
+        (promise.value as DeletedApp).state !== AppState.INVALID
```
Rory-Powell, line 134: Probably worth exposing a type helper for checking `state`. 
i.e. 

```ts
export function isDeletedApp = (
  app: App | DeletedApp
): App is DeletedApp => {
  if (app.state === AppState.INVALID) {
    return true
  }
  return false
}
```
usage:
```ts
promise.status === "fulfilled" && !isDeletedApp(promise.value)
```
This can then be reused across both checks

Rory-Powell: `isDeletedApp` / `DeletedApp` / `AppState` should probably all be moved to:
`budibase/packages/types/src/sdk/apps.ts`

shogunpurple: Hmm - given this is essentially used in the context of only one function (albeit in 2 places for `pro`) it's a bit of a needless abstraction IMO using the type helper - the cast is valid (and clear) here given the type check here is only for one field and TS is being overly cautious. I'd still need to add the explicit field check in the type helper for TS to compile, which seems a bit overkill given it's usage:

```js
export function isDeletedApp(app: App | DeletedApp): app is DeletedApp {
  if ('state' in app && app.state === AppState.INVALID) {
    return true;
  }
  return false;
}
```

PR #11318, opened 2023-07-21
Pino logging tweaks

adrinr:
## Description
1. Log service name only when running self-host. "service" is being injected by DataDog using the pod names, so we should leave it blank to allow the default behaviour if it's not running self-hosted.
2. Honoring min levels
3. Also setting up the default stdout when not running in dev

### DIFF
--- a/packages/backend-core/src/logging/pino/logger.ts
+++ b/packages/backend-core/src/logging/pino/logger.ts
@@ -35,9 +35,9 @@ if (!env.DISABLE_PINO_LOGGER) {
 
   const destinations: pino.DestinationStream[] = []
 
-  if (env.isDev()) {
-    destinations.push(pinoPretty({ singleLine: true }))
-  }
+  destinations.push(
+    env.isDev() ? pinoPretty({ singleLine: true }) : process.stdout
+  )
 
   if (env.SELF_HOSTED) {
     destinations.push(localFileDestination())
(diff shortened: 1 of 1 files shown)


packages/backend-core/src/logging/pino/logger.ts
```diff
@@ -29,9 +35,9 @@ if (!env.DISABLE_PINO_LOGGER) {
 
   const destinations: pino.DestinationStream[] = []
 
-  if (env.isDev()) {
-    destinations.push(pinoPretty({ singleLine: true }))
-  }
+  destinations.push(
+    env.isDev() ? pinoPretty({ singleLine: true }) : process.stdout
```
Rory-Powell, line 39: I think this works ok without this line? E.g. when I remove pino pretty, I still see logs in my terminal

Rory-Powell: Actually it does not work for budibase services. It worked in account portal, where the logging returned to plaintext and is shown in the console. I'm not sure of why there is a difference

Rory-Powell: ah - it's because of the `localFileDestination` - pino must default to console when no other streams are present

Rory-Powell (approved):

Rory-Powell (approved):

PR #11319, opened 2023-07-21
adding a livenessprobe to the proxy service

shogunpurple:
## Description
Adding a liveness and readiness probe to the proxy service. Experiment to see if this helps with the unhealthy load balancer alerts, and should give us more control over rollouts

Addresses: 
https://linear.app/budibase/issue/OPS-71/service-instability-around-deployments-and-restarts

### DIFF
--- a/charts/budibase/templates/proxy-service-deployment.yaml
+++ b/charts/budibase/templates/proxy-service-deployment.yaml
@@ -40,6 +40,24 @@ spec:
       - image: budibase/proxy:{{ .Values.globals.appVersion | default .Chart.AppVersion }}
         imagePullPolicy: Always
         name: proxy-service
+        livenessProbe:
+          httpGet:
+            path: /health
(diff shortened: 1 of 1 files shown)


charts/budibase/templates/proxy-service-deployment.yaml
```diff
@@ -40,6 +40,24 @@ spec:
       - image: budibase/proxy:{{ .Values.globals.appVersion | default .Chart.AppVersion }}
         imagePullPolicy: Always
         name: proxy-service
+        livenessProbe:
```
Rory-Powell, line 43: I think we shouldn't be experimenting with this in production - would be better to test in QA first then cherry-pick the changes over.

shogunpurple: Was planning on trying this in preprod just - but it makes sense if there's a hotfix or something

PR #11321, opened 2023-07-21
Removed right from field group label select.

ConorWebb96:
## Description
Removed the right label from the field group component as it served no purpose and didn't change anything.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7321/remove-right-from-field-group-labels-select


aptkingston (approved):

melohagan (approved):
LGTM - thanks!

PR #11332, opened 2023-07-24
Allow sort override on views

adrinr:
## Description
Allow overriding the default sort values via API params

### DIFF
--- a/packages/server/src/api/controllers/row/index.ts
+++ b/packages/server/src/api/controllers/row/index.ts
@@ -164,13 +164,13 @@ function getSortOptions(
   | undefined {
   const { sort_column, sort_order, sort_type } = ctx.query
   if (Array.isArray(sort_column)) {
-    ctx.throw(404, "sort_column cannot be an array")
+    ctx.throw(400, "sort_column cannot be an array")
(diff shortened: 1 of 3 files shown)


packages/server/src/api/controllers/row/index.ts
```diff
@@ -146,6 +152,45 @@ export async function search(ctx: any) {
...
+):
+  | {
+      sort: string
+      sortOrder?: SortOrder
+      sortType?: SortType
+    }
+  | undefined {
+  const { sort_column, sort_order, sort_type } = ctx.query
+  if (Array.isArray(sort_column)) {
+    ctx.throw(404, "sort_column cannot be an array")
```
mike12345567, line 167: Should these be a 404? Given that its an incorrect type - the API was found, but invalid inputs.

mike12345567 (approved):
Looks good - might just change the status code on the errors!

packages/server/src/api/controllers/row/index.ts
adrinr: Right, that was meant to be a 400

PR #11334, opened 2023-07-24
Prevent parentheses in query names

aptkingston:
## Description
This PR validates query names. The only characters currently banned are parentheses as they break JS bindings.

Validation has been added to both frontend and backend.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7262/invalid-characters-in-query-names


deanhannigan (approved):
LGTM!

Issue #11326, opened 2023-07-22
Formula dialog won't load and makes the page unresponsive


Bodent81:
**Checklist**
- [ ✔️] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: mhmlu

**Describe the bug**
Having issues with the modal formula dialog not showing and freezing the page. Creating a formula column and pressing the lightning bolt icon in the 'Create Column' dialog and, since a few days the formula dialog won't open. The page becomes unresponsive until I reload it. Same issue if I try and edit on an already existing column. 
I tried Brave Browser version 1.56.11, Safari 16.5.1. Also tried my private computer (same two browsers) to no avail.

If I enter a formula directly in the text field in the 'Create Column' dialog I can save the column and the formula works.

**To Reproduce**
1. Create a formula column.
2. Press the lightning bolt icon to the right of the formula field.
3. The formula dialog doesn't show and page becomes unresponsive until reload.

**Expected behavior**
The formula dialog to open so one can get some assistance when writing expressions.

**Screenshots**
<img width="1105" alt="Formula Dialog" src="https://github.com/Budibase/budibase/assets/140268445/31153ff6-bac2-4456-badf-5b82b9de36fc">

**App Export**

**Desktop (please complete the following information):**
 - OS: Ventura 13.4.1 
 - Browser: Brave Browser version 1.56.11, Safari 16.5.1

aptkingston:
Appreciate the detailed issue @Bodent81, thank you! Will get a fix out for this soon.

aptkingston:
Fix should already be live 👍

PR #11337, opened 2023-07-24
Fix builder crash when editing formula bindings

aptkingston:
## Description
Fixes a builder crash when attempting to edit a formula column binding.

Addresses: 
- https://github.com/Budibase/budibase/issues/11326


deanhannigan (approved):
LGTM!

PR #11338, opened 2023-07-24
Stop data query binding drawer value from being cleared.

deanhannigan:
## Description

A function used in the `DataSourceSelect` was inadvertently clearing the query binding values in the UI.

- The value remains populated on load and when the `cog` icon is clicked next to the `Data` setting.
- The behaviour in the `Execute Query` button action was updated to accommodate the fix

Addresses: 
- BUDI-7090

## Screenshots

A `DataProvider` component configured with a Query data source. Click on the cog to see the affected drawer.
Screenshot 2023-07-24 at 17 51 29


aptkingston (approved):
LGTM!

Issue #10654, opened 2023-05-20
[BUDI-7041] Option picker dropdown not working on mobile


ben-biber:
**Checklist**

- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.6.20
  - App Version: 2.6.20

**Describe the bug**
Option picker dropdown not working on mobile

**To Reproduce**
1. Create Option picker with select type in form 

**Expected behavior**
The app there should open a dropdown on mobile if you click on the field

<sub>[BUDI-7041](https://linear.app/budibase/issue/BUDI-7041/option-picker-dropdown-not-working-on-mobile)</sub>

aptkingston:
Hi @ben-biber. I've tried to reproduce this, but it works as expected.

<img width="368" alt="image" src="https://github.com/Budibase/budibase/assets/9075550/07e9ede0-cc5d-4432-84f8-c35073a470e0">

Would you be able to post a screenshot showing your form?

Do you definitely have some options available? If you don't have any options you'll just get an empty dropdown like this:

<img width="364" alt="image" src="https://github.com/Budibase/budibase/assets/9075550/7a9c7fed-3306-4ed1-a300-cc28a97b7fb9">

ben-biber:
Here is an Export:

![Sommerfreizeit 2023-export-1684775982788.tar.gz](https://github.com/Budibase/budibase/files/11534305/Sommerfreizeit.2023-export-1684775982788.tar.gz)

The Problem ist in the second Form step at the "Klasse" field .
It only accurse on a Mobile Browser

ben-biber:
I tested it on a Pixel 6 Pro running the Vivaldi Browser 6.0.2980.48

ben-biber:
I think is some kind of CSS issue?
I can even recreate this problem on my Mac with Safari if I reduce the width of the browser window

ben-biber:
The problem is on the second form step
https://kjgdossenheim.budibase.app/app/bens-app

arminus:
I have the same problem. Same server version, dropdown works on Windows, but doesn't open on an Android 13 M32 latest Chrome (and some others who just reported in). Had to replace the dropdown with radio buttons as a workaround.

aptkingston:
@ben-biber @arminus sorry it took a while to manage to reproduce this one. I have a PR up to fix this. Here's your app working as expected on mobile now:
[...]

PR #11345, opened 2023-07-25
Fix nullish check bug causing options fields to break on mobile sometimes

aptkingston:
## Description
Fixes a bad nullish check in the dropdown positioning logic which meant that there were certain situations (most often on mobile) where dropdowns did not appear.

Addresses: 
- #10654


deanhannigan (approved):
LGTM!

Issue #10970, opened 2023-06-20
[BUDI-7172] Define relationships button is visible for Google Sheets even though it does not support relationships


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Describe the bug**
> I am trying out a self-hosted budibase installation  with google sheets integration. I am trying to create a simple relationship > between two tables. After fetching the tables successfully and on trying to establish the relationship, I am getting this error: > "Error saving column: Column type: link not allowed for GSheets integration."

> Now, I am somewhat confused because feedback I am getting from the UI is a bit thin. Is the relationship not possible? > Looking at https://budibase.com/blog/data/google-sheets-dashboard/ it seems like it should, but the post is also slightly > outdated and everything looks different. I tried Googling for help but to no avail.

From this discussion: https://ptb.discord.com/channels/733030666647765003/1120666969826394163

**To Reproduce**
1. Add a Google Sheets datasource
2. Fetch some sheets
3. Click on a table -> See that the `Define relationships` button is visible.

**Expected behavior**
Define relationships should be hidden or disabled if the Integration has `relationships: false`

<sub>[BUDI-7172](https://linear.app/budibase/issue/BUDI-7172/define-relationships-button-is-visible-for-google-sheets-even-though)</sub>

jochenstu:
Cool, thanks for the responsiveness and creating the issue right away! (I was the one who brought this thing up)

To add a little onto this: To me, having Google Sheets as a backend in budibase was a super big deal, given that it brings super user friendly backup, rollback and "managed everything". It gels really well with the whole low tech approach, so I want to encourage you to actually include/bring back this option if at all feasible in the future.

mike12345567:
Hi @jochenstu - sadly having relationships in googlesheets is very unlikely to be supported - as there is no concept of relational data within Googlesheets itself it is very difficult for us to provide this.

PR #11346, opened 2023-07-25
Hide the 'Define Relationships' button when an integration does not support relationships

deanhannigan:
## Description

The `Define Relationships` button in data tables will now be hidden when relationships are unsupported by the integration type.

This is a general modification but currently only applies to `Google Sheets` datasources.  

Addresses: 
- https://github.com/Budibase/budibase/issues/10970

### DIFF
--- a/packages/builder/src/components/backend/DataTable/DataTable.svelte
+++ b/packages/builder/src/components/backend/DataTable/DataTable.svelte
@@ -28,15 +28,15 @@
   $: isInternal = $tables.selected?.type !== "external"
 
   $: datasource = $datasources.list.find(datasource => {
-    return datasource._id === $tables.selected.sourceId
+    return datasource._id === $tables.selected?.sourceId
   })
+
+  $: relationshipsEnabled = relationshipSupport(datasource)
 
   const relationshipSupport = datasource => {
     const integration = $integrations[datasource?.source]
     return !isInternal && integration?.relationships !== false
   }
-
-  $: relationshipsEnabled = relationshipSupport(datasource)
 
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/backend/DataTable/DataTable.svelte
```diff
@@ -27,6 +27,17 @@
   $: isUsersTable = id === TableNames.USERS
   $: isInternal = $tables.selected?.type !== "external"
 
+  $: datasource = $datasources.list.find(datasource => {
+    return datasource._id === $tables.selected.sourceId
```
aptkingston, line 31: Would maybe just quickly double check that if you deleted the selected table that this statement doesn't fire with `$tables.selected` being undefined, since that would crash. Could also just add in optional chaining to be extra safe!

packages/builder/src/components/backend/DataTable/DataTable.svelte
```diff
@@ -27,6 +27,17 @@
...
+  })
+
+  const relationshipSupport = datasource => {
+    const integration = $integrations[datasource?.source]
+    return !isInternal && integration?.relationships !== false
+  }
+
+  $: relationshipsEnabled = relationshipSupport(datasource)
```
aptkingston, line 39: NAB: I tend to group reactive statements together for readability, and automatic code hosting will take care of the order (as in it's ok to have the reactive statement before the function definition - it'll still be able to use it).

aptkingston (approved):
LGTM but just one query about safety of `$tables.selected` potentially being null.

PR #11347, opened 2023-07-25
Chore - Jest improvements and clean-ups

adrinr:
## Description
Node 16.11.0+ introduced an [issue](https://github.com/jestjs/jest/issues/11956) around jest memory. As we are planning on upgrading to node 18, we need to sort this out. This PR contains a few improvements and clean-ups around our jest usage:
1. Update and sync all jest libraries
2. Remove the not used ts-jest references
3. Remove other unused libraries
4. Remove the barrel file usage of lodash (known jest conflicts)

Related pro PR: https://github.com/Budibase/budibase-pro/pull/210


mike12345567 (approved):
Looks good! One thing that might be worthwhile is extracting what we need from lodash into a single file and exporting it from backend-core utils or something like that. That way we could remove lodash from the server/worker and hopefully reduce the chance of people importing the whole lodash barrel file.

adrinr:
> 

That is a good idea. I was planning on removing the "lodash" dependency, using the modular ones "lodash.`function`" when required

PR #11352, opened 2023-07-25
Return default row data

adrinr:
## Description
Return mandatory row fields even if the schema is defined. Adding tests around it


mike12345567 (approved):
Nice to have test cases for this - good to confirm it works as expected and stop any regressions!

PR #11354, opened 2023-07-25
Update helm chart to expose OFFLINE_MODE env var

Rory-Powell:
## Description
Add support for setting `OFFLINE_MODE` via kubernetes values.


shogunpurple (approved):
LGTM

PR #11365, opened 2023-07-26
split singleimage build into 3 jobs

jonnymccullagh:
Splitting the single image build into 3 separate jobs rather than 1 single job to see if it helps with the disk space issues experienced when they were all one job
I have tested this on another repo and it appears to work ok.


mike12345567 (approved):
Looks good to me, we could probably run these in parallel as well to speed things up if we wanted too.

calexiou (approved):
looks good to me as well

jonnymccullagh:
> we could probably run these in parallel as well to speed things up if we wanted too.
I believe that parallel is the default unless we specify otherwise so we should get that.

.github/workflows/release-singleimage.yml
```diff
@@ -1,4 +1,4 @@
-name: Deploy Budibase Single Container Image to DockerHub
+name: release-singleimage
```
Rory-Powell, line 1: 👌

.github/workflows/release-singleimage.yml
```diff
@@ -24,17 +24,93 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
         with:
+          repository: Budibase/budibase
+          ref: 'master'
```
Rory-Powell, line 28: Should remove this, as we'll want to checkout with the tag selected in the run

.github/workflows/release-singleimage.yml
```diff
@@ -24,17 +24,93 @@ jobs:
...
           token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
           fetch-depth: 0
-
       - name: Fail if tag is not in master
         run: |
           if ! git merge-base --is-ancestor ${{ github.sha }} origin/master; then
             echo "Tag is not in master. This pipeline can only execute tags that are present on the master branch"
             exit 1
           fi
+      - name: Use Node.js ${{ matrix.node-version }}
```
Rory-Powell, line 38: I think we should have an `_inc` file for the actual build and release, which accepts an arg for the platforms and image name, so that we don't need to duplicate the configuration multiple times as that could be bug prone

jonnymccullagh: I can see pros and cons with that as we may also want to remove some of the buildx steps from some jobs. I'd like to see this running and I'll create a ticket to add refinements.

PR #11368, opened 2023-07-26
Ensure we don't allow views to update other fields

adrinr:
## Description
Implement view2.0 row patch, not allowing editing fields not belonging to the view.
Ideally, the same approach of moving the patch should be done for the internal db, but given its dependencies with other controllers it requires a much bigger refactor. Because of this controller "unit tests" will cover the use cases

### DIFF
--- a/packages/server/src/sdk/app/rows/tests/utils.spec.ts
+++ b/packages/server/src/sdk/app/rows/tests/utils.spec.ts
@@ -1,7 +1,10 @@
 import { generator } from "@budibase/backend-core/tests"
-import { FieldType, Table, ViewV2 } from "@budibase/types"
+import { FieldType, Table } from "@budibase/types"
 jest.mock("../../../../sdk", () => ({
-  views: { get: jest.fn() },
+  views: {
+    ...jest.requireActual("../../../../sdk/app/views"),
+    get: jest.fn(),
+  },
 }))
 
 import sdk from "../../../../sdk"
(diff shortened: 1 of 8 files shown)


adrinr:
@Rory-Powell  First approach of breaking sdks and having unit tests around. Due to our barrel file approach, mockings become really hard. Is this something we worked out before?

mike12345567 (approved):
LGTM - suggested a test for the constant columns, but its minor!

packages/server/src/sdk/app/rows/tests/utils.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        type: FieldType.STRING,
+      },
+    },
+  }
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+  })
+
+  describe("trimViewFields", () => {
```
mike12345567, line 39: A test case which confirms the constant internal/external columns are maintained correctly would be good, as well as making sure that the `_viewId` is always removed as expected.

adrinr: This should not be the job of this trimViewFields. This is a more db logic. It's tested on the patch itself

adrinr:
@mike12345567 thinking about this, we should probably have a different endpoint for this update. There is too much hidden logic within the endpoint, and having flags like the presence of the _viewid to do one thing or another might cause issues and really hard debugging. Thoughts?

aptkingston:
This patch solves the issue of updating rows (securely too), but there are still the other cases of creation and deletion which we need to account for.

For creation, we effectively need to be able to create a row in the parent table but ensure only fields we have access to are set. We should not be able to create a row with fields that are not in the view.

For deletion, we need to be able to delete a row from the parent table. Ideally we would ensure that the row to be deleted actually exists in the view, but this might be too much effort to make it worthwhile.

Just thinking that if you're considering breaking out this endpoint into a dedicated viewsV2 endpoint, then we could also add dedicated endpoints for creation (and potentially deletion).

adrinr:
Closed in favour of https://github.com/Budibase/budibase/pull/11406

PR #11380, opened 2023-07-27
Revert back to use lerna

adrinr:
## Description
While testing the pro load and unload submodules, lerna stopped working as expected around NX. Because of this, we changed these commands to use nx directly.
This is causing some other issues around logging, and we checked on clean installs that the lerna command works fine, so we are reverting back.
Tested commands:
- yarn dev
- yarn dev:server
- yarn dev:noserver

All commands had been tested with clean environments, both with the submodule loaded and unloaded

image


shogunpurple (approved):
LGTM!

Issue #9294, opened 2023-01-10
Add histogram plot


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/9276

<div type='discussions-op-text'>

<sup>Originally posted by **Murali-001** January  9, 2023</sup>
I need histogram plot for my data.</div>

This should be doable with the Apex Charts library used in Budibase.

lcorcodilos:
Hi all,

It seems like this made it into v2.9.25. I upgraded my docker image to this version but don't seem to have the Histogram accessible as either a component or an option in a Chart Block. Is there a trick I'm missing? Can anyone confirm they're able to use the feature in the latest release?

Thanks!

melohagan:
Added a page to the docs: https://docs.budibase.com/docs/histogram-chart

PR #11382, opened 2023-07-27
Added a simple histogram to the supported chart types

deanhannigan:
## Description

Added a simple frequency Histogram component.

- `Histogram Chart` now available in the component list
- `Histogram` now an available type in the `Chart Block`

The `Value column` property should be a number column from your datasource. The histogram chart will take all rows supplied and output a frequency density chart based on configured field.

Increase or decrease the `Bucket Count` to refined the distribution.

Addresses: 
- https://github.com/Budibase/budibase/issues/9294

## Screenshots

Screenshot 2023-07-27 at 13 19 48


aptkingston (approved):
LGTM! Nice one 👌

Issue #11381, opened 2023-07-27
[BUDI-7336] Maintain nesting children validation from apps when imported


shogunpurple:
[https://discord.com/channels/733030666647765003/872822618238095360/1134091946625871983](https://discord.com/channels/733030666647765003/872822618238095360/1134091946625871983)

Users reported breaking changes in apps that used to work due to the introduction of children validation in the design section. This is currently only applied to new apps, but if you export and import an old app the new rules will be applied, effectively breaking the imported application.

**COS**

* maintain the illegalChildren validation setting in the previous app on import - ie if it was off in the previous app, leave it off

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7336](https://linear.app/budibase/issue/BUDI-7336/maintain-illegalchildren-validation-from-apps-when-imported)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7336](https://linear.app/budibase/issue/BUDI-7336/maintain-illegalchildren-validation-from-apps-when-imported)</sub>

aptkingston:
Now been deployed to cloud.

PR #11384, opened 2023-07-27
Disable component validation for old app imports

aptkingston:
## Description
Disable component validation for imports of apps which did not have it enabled.

Current behaviour (validation enabled or not):
Old app => new app
false => true
true => true

New behaviour with this PR (validation enabled or not):
Old app => new app
false => false
true => true

Addresses: 
- https://github.com/Budibase/budibase/issues/11381


shogunpurple (approved):
LGTM

PR #11392, opened 2023-07-28
Per app builders - backend and some frontend

mike12345567:
## Description
This is the 80% of the work needed to support per-app builders, the discussed paid feature. For now this implements all of the backend changes needed to support this, as well as some of the frontend changes needed for a non-global builder to access the builder portal.

The main work here is within the middlewares, as well as adding a new role `GLOBAL_BUILDER` - we are unlikely to change these enumerations so went with what they really are, most of the `BUILDER` APIs have been maintained, but some like app creation and deletion, as well as plugin management have moved to global builders/admins only, while the APIs for building and managing apps have been updated to provide support for app builders.

The authorized middleware has been updated to check the users app ID they are sending up, if it does not match one of the apps they can build in they will be thrown out with a 403 - I've built some tests to confirm this.

The main body of work that was needed was refactoring the users SDK into backend core from the worker so that it could be easily utilised by pro. This wasn't too simple as it would have made backend-core dependent on pro, which is dependent on pro, a circular dependency which we cannot support. Instead I have added a new mechanism for creating an instance of `UserDB` - an ORM like interface over the user objects, which allows performing certain database level tasks in a consistent manner across the server, worker and now pro libraries. To do this, we pass in the components of pro which the user database requires, such as feature checking, quotas and groups. This turned out to work quite well and act as a simple abstraction of the user database functions (like saving and updating).

The other big piece of work was a general refactor in the frontend, I have moved the logic for checking if a user has builder, admin or app builder permissions into `@budibase/shared-core`. This means the same logic can be utilised by the frontend and backend, I have worked through the builder codebase to try and find everywhere that the `.builder` or `.admin` flag on the user was check and instead call through to the utility functions in shared core, making it far more maintainable for the future.

This has meant the PR has turned out quite big, as there has been so many little components that needed refactored to make this a cleaner implementation experience, as well as supporting further changes in the future if required.

Pro PR (API and licensing): https://github.com/Budibase/budibase-pro/pull/211

Addresses: 
- https://linear.app/budibase/issue/BUDI-7052/rbac-allow-per-app-builders-backend
- Partially: https://linear.app/budibase/issue/BUDI-7053/rbac-allow-per-app-builders-frontend

## Screenshots
Main change to UI:
image

Also added a placeholder for when no apps found, app builders cannot create new apps so don't go through onboarding the same way.
image

## Documentation
Not documented yet as frontend is in-complete.


packages/backend-core/src/constants/db.ts
```diff
@@ -1,5 +1,5 @@
-export const SEPARATOR = "_"
-export const UNICODE_MAX = "\ufff0"
+import { prefixed, DocumentType } from "@budibase/types"
+export { SEPARATOR, UNICODE_MAX, DocumentType } from "@budibase/types"
```
adrinr, line 2: Can we namespace or rename the `SEPARATOR` in here to make it more explicit? As a consumer, having a type called `separator` might be pretty confusing

packages/backend-core/src/users/users.ts
```diff
@@ -65,6 +71,27 @@ export const bulkUpdateGlobalUsers = async (users: User[]) => {
   return (await db.bulkDocs(users)) as BulkDocsResponse
 }
 
+export const grantAppBuilderAccess = async (userId: string, appId: string) => {
+  const prodAppId = getProdAppID(appId)
+  const db = getGlobalDB()
+  const user = (await db.get(userId)) as User
```
adrinr, line 77: Can we use the sdk, that already types it?

packages/backend-core/src/users/users.ts
```diff
@@ -65,6 +71,27 @@ export const bulkUpdateGlobalUsers = async (users: User[]) => {
   return (await db.bulkDocs(users)) as BulkDocsResponse
 }
 
+export const grantAppBuilderAccess = async (userId: string, appId: string) => {
+  const prodAppId = getProdAppID(appId)
```
adrinr, line 75: Is this not duplicated in pro? `src/sdk/users/users.ts` addAppBuilder

packages/server/src/app.ts
```diff
@@ -15,7 +15,15 @@ import * as api from "./api"
...
-import { events, logging, middleware, timers } from "@budibase/backend-core"
+import { ServiceType } from "@budibase/types"
+import {
+  events,
+  logging,
+  middleware,
+  timers,
+  env as coreEnv,
+} from "@budibase/backend-core"
+coreEnv._set("SERVICE_TYPE", ServiceType.APPS)
```
adrinr, line 26: This should be in `packages/server/src/environment.ts`

packages/server/src/db/utils.ts
```diff
@@ -3,10 +3,10 @@ import { db as dbCore } from "@budibase/backend-core"
 
 type Optional = string | null
 
-export const AppStatus = {
-  DEV: "development",
-  ALL: "all",
-  DEPLOYED: "published",
+export enum AppStatus {
```
adrinr, line 6: We should use `const enum` when possible

packages/server/src/sdk/app/applications/applications.ts
```diff
@@ -0,0 +1,56 @@
...
+  const roleApps = Object.keys(user.roles || {})
+  if (users.hasAppBuilderPermissions(user)) {
+    appList = user.builder?.apps || []
+    appList = appList.concat(roleApps)
+  } else if (!users.isAdminOrBuilder(user)) {
+    appList = roleApps
+  } else {
+    return apps
+  }
+  const finalApps: App[] = []
```
adrinr, line 19: This could be simplified in a `map`

packages/shared-core/src/sdk/documents/applications.ts
```diff
@@ -0,0 +1,29 @@
+import { DocumentType, prefixed } from "@budibase/types"
+
+const APP_PREFIX = prefixed(DocumentType.APP)
+const APP_DEV_PREFIX = prefixed(DocumentType.APP_DEV)
+
+export function getDevAppID(appId: string) {
+  if (!appId || appId.startsWith(APP_DEV_PREFIX)) {
```
adrinr, line 7: Should `if !appId` not throw?

packages/types/src/sdk/permissions.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+// these are the global types, that govern the underlying default behaviour
+export enum PermissionType {
+  APP = "app",
+  TABLE = "table",
+  USER = "user",
+  AUTOMATION = "automation",
+  WEBHOOK = "webhook",
+  BUILDER = "builder",
```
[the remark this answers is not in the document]
mike12345567: We could rename the enum, but not the value, as the permission is used quite a bit, I think that would be quite a bit more confusing than adding the new permission for global builders (it'll make more sense post frontend update)

packages/types/src/sdk/permissions.ts
```diff
@@ -0,0 +1,19 @@
+export enum PermissionLevel {
```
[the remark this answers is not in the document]
mike12345567: This isn't something new, just moving it, we can't change it as it is built into roles already.

packages/builder/src/stores/portal/menu.js
```diff
@@ -2,8 +2,12 @@ import { derived } from "svelte/store"
 import { isEnabled, TENANT_FEATURE_FLAGS } from "helpers/featureFlags"
 import { admin } from "./admin"
 import { auth } from "./auth"
+import { sdk } from "@budibase/shared-core"
 
 export const menu = derived([admin, auth], ([$admin, $auth]) => {
+  const user = $auth?.user
+  const isAdmin = sdk.users.isAdmin(user)
```
aptkingston, line 9: isAdmin is already available as $auth.isAdmin so we don't need to call the sdk again.

packages/builder/src/stores/portal/menu.js
```diff
@@ -24,19 +28,21 @@ export const menu = derived([admin, auth], ([$admin, $auth]) => {
       title: "Apps",
       href: "/builder/portal/apps",
     },
-    {
+  ]
+  if (sdk.users.isGlobalBuilder(user)) {
```
aptkingston, line 32: Another example of where we could memoize this in the auth store.

aptkingston (approved):
LGTM! No blockers, but I think we could simplify a little but by memoizing the required user permissions in the auth store rather than invoking the SDK at runtime multiple times.

I'd also appreciate a quick explanation of builder vs globalBuilder vs hasBuilderPermissions, when there's no app ID involved.

My guess is:
- global builder is self explanatory
- builder is... builder in any app?
- hasBuilderPermissions... not sure

packages/builder/src/pages/builder/index.svelte
```diff
@@ -1,11 +1,12 @@
 <script>
   import { redirect } from "@roxi/routify"
   import { auth } from "stores/portal"
+  import { sdk } from "@budibase/shared-core"
 
   auth.checkQueryString()
 
   $: {
-    if ($auth.user?.builder?.global) {
+    if (sdk.users.hasBuilderPermissions($auth.user)) {
```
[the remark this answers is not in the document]
mike12345567: So, basically the difference is:
1. `isGlobalBuilder`, as you say its self explanatory, it is the current builder type user
2. `isBuilder`, checks if the user is a global builder, without an app ID it will fail for app builders, with an app ID then app builders can be permitted
3. `hasBuilderPermissions` - this means that the user could be a builder in an app, but you don't have the app ID to confirm which app it is, for example the decision to show them the builder portal. Its less meaningful in frontend, its used heavily in backend for requests that we aren't aware of an app for

packages/backend-core/src/constants/db.ts
mike12345567: This isn't a new thing, its been around since some of the earliest versions of Budibase, the `SEPARATOR` will always be `_`.

packages/server/src/sdk/app/applications/applications.ts
adrinr: const finalApps = apps.filter(app => appList.includes(dbCore.getProdAppID(app.appId))

PR #11401, opened 2023-07-31
Plugin improvements to allow block plugins

aptkingston:
## Description
This PR is a couple of small improvements that will allow plugin authors to make custom blocks.

- `Block` and `BlockComponent` are exported via the client SDK, so can be used in plugins
- `BlockComponent` now supports plugins

The rough structure of a block plugin is as follows:
```html
<script>

  const { Block, BlockComponent } = getContext("sdk")
</script>

<Block>
  <BlockComponent type="container">
    <BlockComponent order={0} type="plugin/component-1" props={{ text: "Hello, world!" }} />
    <BlockComponent order={1} type="plugin/component-2" props={{ text: "Foo" }} />
  </BlockComponent>
</Block>
```

- `Block` must wrap everything
- All components are initialised via passing in a `type` to `BlockComponent`. This can be a plugin (`plugin/xxx`) or a built-in Budibase component (`container` for example).
- If you use a plugin, you need to make sure to add this plugin to Budibase, and to your app! We don't automatically add it for you.

### Supporting ejection
If you want Budibase to see your plugin as an actual block, and therefore offer the "eject" functionality, then there are a few more guidelines you need to follow.
- Be sure to set `"block": true` in your `schema.json`. This tells Budibase that your component plugin is a block.
- One of the core requirements of ejection is that everything is contained inside a single component, so I recommend you wrap all content inside a container as shown in the example above.
- You need to tell Budibase what order you want your components to follow after ejection. This is achieved by using the `order` prop on the `BlockComponent` component. You can reference the source code of our existing blocks for examples, but it basically just means that all sibling components at the same depth should get `order={0}`, `order={1}` passed in sequentially. `order={0}` is not required when there is only a single child.

There may be docs for this at some stage, but for now the best idea is reference our existing blocks for examples of how to build complex blocks.


aptkingston:
Going to go ahead and merge this in - it's basically zero risk and we're already slightly after code freeze, but I'd like to get this out in this release.

PR #11404, opened 2023-07-31
add startupProbe;Probes into values.yaml

jonnymccullagh:
## Description
PR to update the readiness/liveness probes we have in our helm chart by adding a `startupProbe` so that we check the container has fully started before we begin the liveness checks.
- Replaced the existing liveness/readiness lines for app-service and worker service with variables to pull these values (if present) from the `values.yaml` file 
- Added liveness/readiness lines to the proxy service 
- Added a section for startUp probe to the 3 services to give time for the container to become ready to take traffic
- Added examples into the `values.yaml` file

Tested the helm chart with the accompanying `values.yaml` using `helm template` and the output produced is what I expected.

Feedback on the values for the probes is welcome as I will now to moving to create a PR in the deploys repo to test this on QA.

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -201,25 +201,24 @@ spec:
 
         image: budibase/apps:{{ .Values.globals.appVersion | default .Chart.AppVersion }}
         imagePullPolicy: Always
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM!

calexiou (approved):
lgtm 🚢

PR #11406, opened 2023-07-31
CrUD row from views

adrinr:
## Description
Implement create/update/delete rows via views. This PR created and uses a set of middleware to make sure we have the data for both table/view row updates set up correctly. This means trimming the data request when needed based on the endpoint

Addresses: 
- [BUDI-7189](https://linear.app/budibase/issue/BUDI-7189/[views-v2]-new-views-object)


packages/server/src/api/controllers/row/views.ts
```diff
@@ -0,0 +1,86 @@
+import { quotas } from "@budibase/pro"
```
adrinr, line 1: This code had been just moved from `packages/server/src/api/controllers/row/index.ts`

mike12345567 (approved):

PR #11407, opened 2023-07-31
Split columns and UI on the fields

adrinr:
## Description
After discussing the usage for the grid, we are splitting the column security (what fields the view will retrieve) and the UI (which ones are visible, width, etc.).

Both create and update endpoint will be checking the view schemas vs the table schemas, ensuring that every field defined in there matches the table schema, allowing only UI changes

### DIFF
--- a/packages/server/src/api/controllers/view/viewsV2.ts
+++ b/packages/server/src/api/controllers/view/viewsV2.ts
@@ -9,9 +9,40 @@ import {
   RequiredKeys,
 } from "@budibase/types"
 
-export async function create(ctx: Ctx<CreateViewRequest, ViewResponse>) {
-  const view = ctx.request.body
-  const { tableId } = view
+async function parseSchemaUI(ctx: Ctx, view: CreateViewRequest) {
+  if (!view.schema) {
+    return
+  }
+
+  function hasOverrides(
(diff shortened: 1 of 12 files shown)


mike12345567 (approved):

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -2,15 +2,77 @@ import sdk from "../../../sdk"
...
+async function parseSchemaUI(ctx: Ctx, view: CreateViewRequest) {
+  if (!view.schema) {
+    return
+  }
+
+  function hasOverrides(
+    newObj: Record<string, any>,
+    existingObj: Record<string, any>
+  ) {
+    for (const [key, value] of Object.entries(newObj)) {
```
mike12345567, line 21: This is somewhat difficult to follow, would it be easier to use `.find`? E.g.

```
return !!Object.entries(newObj).find(([key, value]) => {
  const isObject = typeof value === "object"
  const existing = existingObj[key]
  return (isObject && hasOverrides(value, existing || {})) || (!isObject && value !== existing) 
})
```

mike12345567 (approved):
Looks good! Good to error check this, hopefully reduce confusion.

PR #11417, opened 2023-08-01
Chore - Sort build and NX issues

adrinr:
## Description
Sorting the way we build and ingest the packages in our monorepo. This simplify its flow, allowing us to remove many of the pre-build steps and the double build for types and shared core.
This also removes the dev errors we were usually seeing, better watching the builds and speeding up script execution times.

Main changes:
1. nx default targets had been removed in favour or specific package target dependencies
2. Removing all the unnecessary nx dependency script runs.
3. Building only ES versions of `types` and `shared-core`. The cjs will never be consumed, only es6 and the TS versions will be used.
4. `--parallel` flags from lerna as they [had been deprecated](https://lerna.js.org/docs/lerna6-obsolete-options#--parallel) on lerna 7
5. Change integration github action to not build unnecessary packages
6. Use `@budibase/shared-core` as module from `@budibase/backend-core`
7. Remove `gitHead` from package.jsons are they are obsolete
8. Change nodemon watch configuration to read all the "new" packages
9. Add `--bail` flag to jest to fail quickly (while handling flaky tests)

Learnings:
Composite is needed in order to run incremental builds. We ran into some issues due the `tsbuildinfo` file being generated in the `/src`. This file is the snapshot of the last build, but because we delete the `/dist` folder multiple times, on the next incremental build, as nothing was there to build, the dist folder was never recreated. Forcing the `tsbuildinfo` to live within the `/dist` folder will make sure that if we delete this folder, everything needs to be rebuilt.

Tested scenarios:
1. Build all packages from a clean start ✅
11. Build each package one at a time from a clean start ✅
12. Run dev script from a clean start ✅
13. Run dev on each package one at a time from a clean start ✅
14. Test all packages ✅


mike12345567 (approved):
Looks great, thank you for the explanation as well!

Issue #11423, opened 2023-08-02
[BUDI-7355] Using a Formula column as the display column in SQL will cause the table not to load data after re-fetching


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.8.14
  * App Version: 2.8.14

**Describe the bug**
I have a SQL Server datasource and I fetch some tables. In one of the tables I create a new Formula column and set it as the display column. When I fetch tables again, that table will no longer return any data.

**To Reproduce**

1. Fetch tables for a SQL datasource
2. Add a formula column to a table
3. Fetch tables again
4. See that the table with the formula column no longer loads any rows

**Expected behavior**
I should be able to set a formula column as the display column in a SQL table, and not have any issues when re-fetching my tables.

**Screenshots**

![Screenshot 2023-08-02 at 13.40.47.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/10f37133-1e50-4f04-a02c-7ed3391b6b38/63a69ae8-deba-45f4-8766-5394f77c38d1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8xMGYzNzEzMy0xZTUwLTRmMDQtYTAyYy03ZWQzMzkxYjZiMzgvNjNhNjlhZTgtZGViYS00NWY0LTg3NjYtNTM5NGY3N2MzOGQxIiwiaWF0IjoxNjkwOTgwMTE0LCJleHAiOjE2OTEwNjY1MTR9.6Ad5p6jULfZDII9FbtMGr_t47M1hBxHYYB_nrqwyQe8)

*Now fetch tables and click on the table*…

[...]

PR #11424, opened 2023-08-02
Fix for formula display columns issue in SQL Server

melohagan:
## Description
When using a formula column in SQL Server, fetch tables was causing the table to break because SQL server needs to have a sort order assigned when doing a **fetch** statement.

Addresses: 
- https://github.com/Budibase/budibase/issues/11423


mike12345567 (commented):
We should rebase this to `develop` ahead of the release next week, otherwise approved!

melohagan:
> We should rebase this to `develop` ahead of the release next week, otherwise approved!

New PR opened: https://github.com/Budibase/budibase/pull/11439

PR #11431, opened 2023-08-03
BUDI-7189 - Type search

adrinr:
## Description
1. Changing the view row search to be a post request, using the same request and response type as we have for a table row search
2. Typing requests


packages/server/src/api/controllers/row/views.ts
```diff
@@ -29,49 +32,29 @@ export async function searchView(ctx: UserCtx<void, SearchResponse>) {
...
-        query: view.query || {},
-        fields: viewFields,
-        ...getSortOptions(ctx, view),
-      }),
-    {
-      datasourceId: view.tableId,
-    }
-  )
+
+  const searchOptions: RequiredKeys<SearchViewRowRequest> &
```
adrinr, line 36: Doing so will make sure that the request object and the SDK parameters match, and adding new fields in the request body will require a specific mapping in here

mike12345567 (approved):
Good to get this typed!

Issue #11420, opened 2023-08-02
[BUDI-7354] When changing the field type of a custom query, the filter options are not immediately updated in the Design filter drawer


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.8.14
  * App Version: 2.8.14

**Describe the bug**
I had a custom query with some fields that were of type 'Text'. In the Design section I added a filter for the 'Text' column **order_date**, and saved.
Realising the data type of the field was wrong, I went back to the Data section and set the **order_date** column type to `Date/time` in the custom query, and saved.
Back to the Design, the filter options still didn't show more than / less than. I had to select a different field, and then back to the original field to get it to update the available filter options.

**To Reproduce**

1. Add a custom query that returns a Date field, but leave the type as *Text*. Save.
2. In the Design section add a data provider with a filter on that field. See that the filter options are for the *Text* type.
3. Back to Data, set the Date field to the data type 'Date/time' and save.
4. Back to Design, the filter options have not updated to include more than / less than.
5. Change the field being filtered, then change it back to the original field
6. You can now see the filter options are available

**Expected behavior**
If changing the data types in custom queries in the Data section, it should update the filter options available in the Design section.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/224105bc-0172-42c3-8fdf-63ee3052ab70/34349f6f-1ab5-46ac-a734-3932cb2519d5?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8yMjQxMDViYy0wMTcyLTQyYzMtOGZkZi02M2VlMzA1MmFiNzAvMzQzNDlmNmYtMWFiNS00NmFjLWE3MzQtMzkzMmNiMjUxOWQ1IiwiaWF0IjoxNjkwOTY1ODA5LCJleHAiOjE2OTEwNTIyMDl9.QYKAV1QQeraolDn2NyPiPgwxpcX86WqVcgnd2Dsq4OI)

*Filter options have not updated*

[...]

Issue #11264, opened 2023-07-18
[BUDI-7350] Filter "equal" for numbers not working in data providers


SufiSR:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker single image> 
  - Budibase Version: 2.8.10
  - App Version: 2.8.10

**Describe the bug**
Filter "EQUAL" on data providers does not work for numbers anymore after System Update to 2.8.10
The app version is irrelevant - it doesn't work on either on the old nor the updated.

**To Reproduce**
I created a datasource that picks up data from mysql with a query.
The schema is properly set (eg. text columns as text, number columns as numbers)
I created a data provider and a table. 
I choose my query as a data provider --> Result: Full table is shown, upon inspection of the response the data source returns the following json:

"data": [
        {
            "firstname": "bla",
            "lastname": "bla2",
            "my_date": "2021",
            "Paid": "154.02",
            "Unpaid": "179.20",
            "percentage_paid": "46.22"
        },
       etc.
]
When now applying the filter my_date equals value 2023 no results are returned.

**Expected behavior**
The Equal to Filter should work again

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 114.0.5735.199

**Additional context**
One can see that the response returns a list of json objects with all numbers in double quotes.
The filters properly identify the field as either text or number (they will provide different comparison options based on the type of field).

**Evil Workaround**: I changed the schema of the column from number to text and then created a new filter based on the text value. Works for now but not cool.

**What I tried to solve the issue**
Query: I explicitly casted the my_date as an integer.
Filter: I changed the filter to equal or more/less --> Results are properly shown
Filter: I changed the filter lastname equals value "bla2" --> Results are properly shown
Filter: I changed the filter Paid equals value 154.02 --> No Data
Filter I changed the filter my_date equals "2023" --> No Data
New Page: I created an entirely new page with my components and the error persists

<sub>[BUDI-7350](https://linear.app/budibase/issue/BUDI-7350/filter-equal-for-numbers-not-working-in-data-providers)</sub>

melohagan:
Hey @SufiSR 

> I created a datasource that picks up data from mysql with a query.

Would you be able to provide the query that you are using in addition to the SQL schema (which you can grab within Budibase: [Troubleshooting SQL](https://docs.budibase.com/docs/troubleshooting-sql).

[...]

PR #11435, opened 2023-08-03
Filter equal for UNSIGNED, DECIMAL, BigInt, etc not working in MySQL custom query 

melohagan:
## Description
This issue turned out to be more niche than I had initially thought. Basically if you have a custom query that uses a number type that is outside of the range of JavaScript numbers, then the database clients will cast these as strings by default to avoid loss of precision, which affects equality checks if the *Number* type is selected in the response schema. 
This could be fixed by adding a where clause filter into the custom query, but it's still a bug.  

I have added an optional `defineTypeCastingFromSchema` function to Integrations that can specify how to handle type casting based on the data types selected by the user in custom queries, however this could in theory also be used when editing data table columns. 
Perhaps Date/time fields could implement removing the timestamp when the *Ignore timezones* flag is set.  

Currently this is only implemented for MySQL. The challenge with Postgres was that the type casting did not have an option to specify a column name, and in the case of SQL Server there's still an open issue for this type casting. In future it could be implemented within Budibase, or custom Budibase data types could be added to handle these various scenarios.

I have also added a fix in for the FilterDrawer not refreshing the operators after the query schema was updated.

Addresses: 
- https://github.com/Budibase/budibase/issues/11264
- https://github.com/Budibase/budibase/issues/11420

## Screenshots

Screenshot 2023-08-03 at 19 59 06

**UNSIGNED as Number**

Screenshot 2023-08-03 at 20 00 43

**Equals matches on number**

Screenshot 2023-08-03 at 20 01 38

**UNSIGNED as Text**

Screenshot 2023-08-03 at 20 02 38

**Equals matches on string**


mike12345567 (approved):
Looks good - it would be interesting to see how we might roll this out to the other SQL sources as well.

I've re-based this to `develop` as I think its just a bit too risky to go out as a hot fix.

PR #11441, opened 2023-08-04
Fix for app action fields and don't clone testData

melohagan:
## Description
A recent bug fix had an unintended side effect of causing an issue with app action triggers with fields not being set.

Worked with @andz-bb to resolve the new bug issue without breaking his original fix.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7359/automation-with-app-action-trigger-breaks-when-it-has-a-field-added

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -108,7 +108,13 @@
   /****************************************************/
 
   const getInputData = (testData, blockInputs) => {
-    let newInputData = cloneDeep(testData || blockInputs)
+    // Test data is not cloned for reactivity
+    let newInputData = testData || cloneDeep(blockInputs)
+
+    // Ensures the app action fields are populated
+    if (block.event === "app:trigger" && !newInputData?.fields) {
+      newInputData = cloneDeep(blockInputs)
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


aptkingston:
I think these various bugs in the same piece of code is shouting to us that the core logic sorely needs refactored, because it's incredibly fragile. All these random clones to avoid mutation (existing stuff I mean, not your changes) are terrible practice and that's why I've tried to undo and remove as much of it throughout the codebase as possible.

Happy to get this important fix in for now of course, but this badly needs a rewrite at some point!

melohagan:
> I think these various bugs in the same piece of code is shouting to us that the core logic sorely needs refactored, because it's incredibly fragile. All these random clones to avoid mutation (existing stuff I mean, not your changes) are terrible practice and that's why I've tried to undo and remove as much of it throughout the codebase as possible.
> 
> Happy to get this important fix in for now of course, but this badly needs a rewrite at some point!

Definitely! Talked to @PClmnt about that, and I think he's going to take a look if he gets a chance.

PClmnt:
> I think these various bugs in the same piece of code is shouting to us that the core logic sorely needs refactored, because it's incredibly fragile. All these random clones to avoid mutation (existing stuff I mean, not your changes) are terrible practice and that's why I've tried to undo and remove as much of it throughout the

Agreed, it would be good to get some time in the future to rework this.

PClmnt (approved):

PR #11451, opened 2023-08-07
Fix fork checks

adrinr:
## Description
Fix fork checks after https://github.com/Budibase/budibase/pull/11449 changes
After merging it we could test it, realising the code was actually not running as expected


adrinr:
Wrong branch

PR #11453, opened 2023-08-07
Change query stored in view2.0

adrinr:
## Description
Changed view2.0 query filters as discussed, to be able to hydrate the frontend


aptkingston (approved):
LGTM! Thanks @adrinr.

PR #11458, opened 2023-08-07
Merge create, update delete view APIs into row API

mike12345567:
## Description
After some discussion with @aptkingston I've merged the view CUD endpoints into the existing row endpoints to simplify the client library implementation.

Utilising different endpoints for this complicates the implementation on the client-side quite a bit, so using one common API avoids issues like this.

The search endpoint has been kept separate due to it being significantly different and this isn't too hard to implement in the frontend, there are already multiple different data fetch pathways.


aptkingston:
image

Love it 👌 thanks!

packages/backend-core/src/security/permissions.ts
```diff
@@ -78,7 +78,6 @@ export const BUILTIN_PERMISSIONS = {
     permissions: [
       new Permission(PermissionType.QUERY, PermissionLevel.READ),
       new Permission(PermissionType.TABLE, PermissionLevel.READ),
-      new Permission(PermissionType.VIEW, PermissionLevel.READ),
```
adrinr, line 81: How are we giving these permissions to the users? Can this break the current views somehow?

packages/server/src/api/routes/row.ts
```diff
@@ -177,7 +175,7 @@ router
     "/api/:tableId/rows",
```
adrinr, line 175: Can this tableId be the viewId from now on? If so, should we call it differently? SourceId or similar? I can see this causing confusion and future bugs

packages/server/src/db/utils.ts
```diff
@@ -284,10 +284,20 @@ export function getMultiIDParams(ids: string[]) {
  * @returns {string} The new view ID which the view doc can be stored under.
  */
 export function generateViewID(tableId: string) {
-  return `${tableId}${SEPARATOR}${newid()}`
+  return `${DocumentType.VIEW}${SEPARATOR}${tableId}${SEPARATOR}${newid()}`
```
adrinr, line 287: This id will not be part of a document. Maybe this shouldn't be a `documenttype`

packages/server/src/utilities/security.ts
```diff
@@ -22,9 +23,6 @@ export function getPermissionType(resourceId: string) {
     case DocumentType.QUERY:
     case DocumentType.DATASOURCE:
       return permissions.PermissionType.QUERY
-    default:
```
adrinr, line 25: Is this not needed for existing views?

packages/server/src/db/utils.ts
mike12345567: I'll add a `VirtualDocumentType` that might be less confusing

packages/server/src/utilities/security.ts
mike12345567: I've switched over existing views to use the table permission, as it is more accurate now - all of this wasn't baked into anything, it shouldn't affect existing permissions.

packages/backend-core/src/security/permissions.ts
mike12345567: we don't need the view permission anymore, it has been swapped to table permission as that is closer based on how the row API is used.

packages/server/src/api/controllers/row/utils.ts
```diff
@@ -45,13 +45,20 @@ export async function findRow(ctx: UserCtx, tableId: string, rowId: string) {
...
+  if (ctx.params?.sourceId) {
+    return ctx.params.sourceId
+  }
+  // check body for a table ID
+  if (ctx.request.body?.tableId) {
     return ctx.request.body.tableId
   }
-  if (ctx.params && ctx.params.tableId) {
+  // now check for old way of specifying table ID
+  if (ctx.params?.tableId) {
```
adrinr, line 57: Should path not be a priority as well?

packages/types/src/documents/document.ts
```diff
@@ -37,6 +37,13 @@ export enum DocumentType {
   USER_FLAG = "flag",
   AUTOMATION_METADATA = "meta_au",
   AUDIT_LOG = "al",
+  VIEW = "awd",
```
adrinr, line 40: Is this expected?

adrinr (commented):
LGTM, just 2 small comments

packages/types/src/documents/document.ts
mike12345567: woops - good catch, was just checking usage - thanks!

packages/server/src/api/controllers/row/utils.ts
mike12345567: true, I suppose should be sourceId then tableId path

Issue #11457, opened 2023-08-07
[BUDI-7369] Error message: "Row does not exist" when trying to update a value in database (Budibase + Google Sheets)


GuiGom83:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: beatly.budibase.app

**Describe the bug**
I'm using Budibase as app to manipulate a simple google sheets database. It was working fine for the last 2-3 months. But now, when I'm trying to update a data pint (row) and hit "save". I get an error message: "Row does not exist."

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/c6b9a080-cf22-4971-a645-7a2f46c94810/dd909157-5e7e-472f-82e7-cda6bf779808?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9jNmI5YTA4MC1jZjIyLTQ5NzEtYTY0NS03YTJmNDZjOTQ4MTAvZGQ5MDkxNTctNWU3ZS00NzJmLTgyZTctY2RhNmJmNzc5ODA4IiwiaWF0IjoxNjkxNDI4MjM4LCJleHAiOjE2OTE1MTQ2Mzh9.LR1QzImeUD1gvIxnOQE6SUYePUgM-L3D3nYue1P-j5Y)

**To Reproduce**

1. Go to any saved result (row)
2. Click on it to open the sidebar view where you can edit the data
3. Edit something for the test
4. Click in "save"
5. See error message

**Expected behavior**
Once I click in "save" the dat is updated in my google sheets without any errors.

**Screenshots**
See above

**Desktop (please complete the following information):**

* OS: MacOS 13.2.1 (22D68)
* Browser: Chrome
* Version: Version 114.0.5735.198 (Official Build) (arm64)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7369](https://linear.app/budibase/issue/BUDI-7369/error-message-row-does-not-exist-when-trying-to-update-a-value-in)</sub>

PR #11460, opened 2023-08-07
Fix for DS+ updating rows

mike12345567:
## Description
Small programming mistake during the views update, should have been `_id` it was fetching from the row body, not `id`.

Addresses: https://github.com/Budibase/budibase/issues/11457


shogunpurple (approved):
LGTM - nice work on that fix

PR #11461, opened 2023-08-07
Fix for CLI building correctly with esbuild

mike12345567:
## Description
Utilising esbuild for the CLI to bundle in backend-core and types as required.


shogunpurple (approved):
LGTM! Nice work

PR #11467, opened 2023-08-08
Use NX affected

adrinr:
## Description
Taking advantage of nx via lerna, using its "[affected](https://nx.dev/concepts/affected)" utils. Affected allows you to run certain commands only if there is any code change in its dependency chain.

This is currently our dependency graph:
image

Any modification in a package will be detected along the chain, running only the required commands.
Examples:
1. Changes on a pipeline or in qa-core. No packages were affected, so no need to run build/test/etc.
2. Changes in the worker. Nothing depends on the worker, so only the worker scripts will run
3. Changes in @budibase/types. Almost everything depends on these packages. All scripts will run except `SDK` and `string-templates` 
image
5. 

This will not only make our pipeline quicker and hopefully help reduce the issues with flakiness, but it will also make us more aware of the full chain. In the future, we can move files across different packages to make our chain more efficient

### How will this affect the current pipelines
These changes will only take effect on PRs, excluding PRs to master. This was set as a "safety" mechanism to catch any issue if we were to "badly" detect affections. Any PR to master will still run all the checks, so these changes are safe. In the future we can revisit this to increase the "affect" usage.

The "affect" will always be compared to the base branch, so chained PRs will have minimal checks to run

- PR pipeline (tests "ran" in 5 seconds): https://github.com/Budibase/budibase/actions/runs/5812795519/job/15758948619?pr=11467
image

- Push pipeline (no affect usage): https://github.com/Budibase/budibase/actions/runs/5818608000/job/15775368610

### DIFF
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -18,7 +18,8 @@ env:
   BRANCH: ${{ github.event.pull_request.head.ref }}
   BASE_BRANCH: ${{ github.event.pull_request.base.ref}}
   PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
-  NX_BASE_BRANCH: origin/${{ github.base_ref || 'master'}}
+  NX_BASE_BRANCH: origin/${{ github.base_ref }}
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
Love this!

Issue #9939, opened 2023-03-08
[BUDI-7517] Allow continuing actions after form validation even if invalid


lolcabanon:
**Describe the feature request**

It would be nice to allow continuing actions after form validation even if invalid.

Example (this is what I am trying to do) :

1. Validate form
2. Send notification : `if (!valid) return 'error: please fill values' else return null;`
3. Stop if (!valid)
4. Save row

Before the step "Validate form" the state is always valid on an empty form, and after "Validate form" the action is stopped if the form is not valid.

Also, if I click "Save" and the form is not valid, then change values to make it valid, the state is still invalid until the step Validate form, so I can't send notification before (see screenshot below). In this case, it shows notification 'error : form invalid' and then validate and save the form.

**Screenshots**

![image](https://user-images.githubusercontent.com/90099349/223626272-7321453a-3dac-44bb-a0df-35df5ea50d47.png)

**Reference**

Also see #9938 bug report about form validation.

<sub>[BUDI-7517](https://linear.app/budibase/issue/BUDI-7517/allow-continuing-actions-after-form-validation-even-if-invalid)</sub>

lolcabanon:
Still the same today. I understand this could be a breaking change tho.

What about an option to "Validate without stopping" in the "Validate form" action? This would preserve the actual default (validate and stop) but would allow to Validate (no-stop) -> Notify user -> Stop if invalid.

On a long form there can be absolutely no direct visual cue if the first field is invalid until user scroll back to top, this is really unintuitive.

PR #11473, opened 2023-08-08
Scroll to enhancements and final actions

melohagan:
## Description
Noticed a couple of small things when working on the recipe tutorials. 

**1. Scroll to top of page**. 

This could be done with a hidden field, but really I just wanted to be able to select the vertical alignment of the scroll to be 'End' so that the scroll would naturally go to the top.

**2. Navigate To, even if condition failed**. 

This has popped up with a few users, and I did run into it myself. In my case it was a form that allowed a user to toggle between creating a new related entry, or selecting an existing one. Regardless of the condition (which determined the save action) I wanted to navigate after submission.
The approach I took with this feature is more of a temporary step, before action branching is added, but it could open up a few use-cases. 
You tick a box to whitelist allowed actions after the **Continue if** action step. If you have multiple conditionals, then only the last logic action has this option enabled. Previous continue if steps will have this setting cleared and disabled.
This is to avoid overly complex flows / unexpected results. 

**3. Scroll to invalid field**. 

With the Scroll To action enabled, it seemed reasonable to implement that with the validation so that off screen invalid fields would be shown to the user. 
*Should this be optional? Is there a scenario you wouldn't want the scrollIntoView?*

Addresses: 
- https://github.com/Budibase/budibase/issues/7016
- https://github.com/Budibase/budibase/issues/9939

## App Export
[Scroll and Final actions-export-1691506016459.tar.gz](https://github.com/Budibase/budibase/files/12292651/Scroll.and.Final.actions-export-1691506016459.tar.gz)

## Screenshots

**Scroll To (vertical alignment and validation)**

https://github.com/Budibase/budibase/assets/101575380/d0b5287a-53fd-47b4-9363-8eee743bc9c4

Screenshot 2023-08-08 at 15 54 35

**Finally block for actions**

Screenshot 2023-08-08 at 15 55 19
*Setting disabled because it is not the last 'Continue if' action*

Screenshot 2023-08-08 at 15 55 46
*Allow 'Navigate To' actions that occur after this action step, even if the condition is false*

https://github.com/Budibase/budibase/assets/101575380/dd3f590c-11ef-4ef9-9f46-b8176d5cffb0

## Documentation
- [x] Will update relevant pages if merged.


aptkingston:
I'll let product maybe weigh in on this before reviewing the code. I understand your use cases, but some of this is a little complicated so it might be best to let them decide if/how we ship features like these 👍

melohagan:
> I'll let product maybe weigh in on this before reviewing the code. I understand your use cases, but some of this is a little complicated so it might be best to let them decide if/how we ship features like these 👍

Definitely! Just wanted to see what it'd look like

PR #11476, opened 2023-08-08
Chore - building packages via esbuild and consume from builder

adrinr:
## Description
Changing the way we build `backend-core`, `shared-core` and `types` to use esbuild with all the bundled dependencies.
Also changing the builder to consume directly `shared-core` and `types` as a module, so it does not need to be compiled in any specific way.

We are building all the packages as cjs, so any external consumer should be able to run it. The changes in the builder will allow us to build using the packages' source code instead of the built version.

Tested scenarios:
1. Build all from a clean install ✅
2. Run dev from a clean install ✅
3. Run dev via docker from a clean install ✅

### DIFF
--- a/packages/bbui/package.json
+++ b/packages/bbui/package.json
@@ -98,8 +98,7 @@
           {
             "projects": [
               "@budibase/string-templates",
-              "@budibase/shared-core",
-              "@budibase/types"
+              "@budibase/shared-core"
             ],
             "target": "build"
           }
--- a/packages/builder/package.json
+++ b/packages/builder/package.json
@@ -133,9 +133,7 @@
         "dependsOn": [
           {
             "projects": [
-              "@budibase/shared-core",
-              "@budibase/string-templates",
-              "@budibase/types"
+              "@budibase/string-templates"
             ],
             "target": "build"
           }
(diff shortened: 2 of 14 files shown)


mike12345567 (approved):
Changes look great, nice to use esbuild for more stuff, removes so much of the headache and confusion, great to simplify our tsconfig as well!

packages/cli/package.json
```diff
@@ -63,20 +63,5 @@
     "renamer": "^4.0.0",
     "ts-node": "^10.9.1",
     "typescript": "4.7.3"
-  },
```
adrinr, line 66: Not needed since it's built using esbuild

packages/builder/package.json
```diff
@@ -133,9 +133,7 @@
         "dependsOn": [
           {
             "projects": [
-              "@budibase/shared-core",
-              "@budibase/string-templates",
-              "@budibase/types"
+              "@budibase/string-templates"
```
adrinr, line 136: Not needed as now vite is using it as an alias path

packages/bbui/package.json
```diff
@@ -98,8 +98,7 @@
           {
             "projects": [
               "@budibase/string-templates",
-              "@budibase/shared-core",
-              "@budibase/types"
```
adrinr, line 102: This is not a bbui dependency

aptkingston: shared-core is imported in `src/helpers.js` so it is a dependency. It's only proxying a module from shared-core though, so it could be refactored to be removed, but right now it is a dependency.

PR #11483, opened 2023-08-09
Handle tableschema changes on views

adrinr:
## Description
Adding new logic while amending the default table schema.
1. Do not allow overriding viewsV2 when updating a table. We will use the view controllers for this, and otherwise, we will be overriding parts of the document that should not be in there.
2. If a new column is added and the schema is defined for the view, the new column will be added as non-visible
3. If a column is deleted and it was defined as a view column, it will be deleted from the view as well
4. If a column is deleted and it was part of the overrides schema UI, it will be deleted from it
5. If a column is renamed and it was not part of the view, it will be handled as a normal addition and deletion
6. If a column is renamed but it was part of the view, the configuration will be kept


aptkingston:
Nice one @adrinr! Does this account for schema updates? So if we had column `X` in a table, then made a view, which now also includes column `X`. Then we rename column `X` to `Y`, or change the type of the column. Are those changes synced to both the view schema and the view column overrides if required?

adrinr:
> Nice one @adrinr! Does this account for schema updates? So if we had column `X` in a table, then made a view, which now also includes column `X`. Then we rename column `X` to `Y`, or change the type of the column. Are those changes synced to both the view schema and the view column overrides if required?

For the schema type changes we should not worry, as the only fields that are being handled are the UI ones.
You have a good point about renamings, as if we are renaming visible columns, we want to keep it visible. I will handle these use cases

PR #11484, opened 2023-08-09
Adding NPM ignores - make sure dist/ pushed to NPM

mike12345567:
## Description
Adding NPM ignore files so that doesn't default to the .gitignore when publishing to NPM.


adrinr:
On this is probably better to whitelist (denying all by default) that blacklisting (example https://github.com/Budibase/budibase/blob/develop/packages/server/.dockerignore)

PR #11500, opened 2023-08-10
Fix python installation in server docker and single image

adrinr:
## Description
Fix python installation in server docker and single image.

Also removing the affected from build. The command `check:types` uses the dist versions, so without this it might fail


mike12345567 (approved):
LGTM! Weird one, but quick fix at least!

PR #11505, opened 2023-08-11
New Left Panel

Ghrehh:
## Description
- Replaces all existing left panels with a split screen/component view.
- New screen search.
- New screen modal/screen now animates upwards, which doesn't cause the preview to reload if closed which is a bit nicer UX imo.
- Navigation is displayed as a component, the navigation ui is highlighted in the preview when it is selected.
- The highlighting in the preview for when the "screen" component is selected has been slightly changed, it now places an outline around the entire preview instead of around the root component of the screen.
- All routes under /design have been removed except for /components, so the user is always directed there when accessing the design section.
- Design requested that the two tabs of the navigation panel be conjoined, so that's done here.

## Known issues to iron out
- Arrow key/other hotkey behavior isn't 100% for the screen and navigation "components" as they both map to the root component for the screen when selected.
- The component tree UI doesn't exactly match what was shown in the mockups, but this wasn't specified in the ticket, so I'll punt on it for now.

Addresses: 
https://linear.app/budibase/issue/BUDI-7163/design-section-search-and-overflow-for-componentsscreens
https://linear.app/budibase/issue/BUDI-7161/design-section-left-panel-changes

|Description|Image|
|---|---|
|Screen search|search|
|New Screen|new screen|
|Navigation|Screenshot 2023-08-11 at 14 02 00|
|Navigation hidden|Screenshot 2023-08-11 at 14 02 05|
|Screen|Screenshot 2023-08-11 at 14 02 13|


aptkingston (changes requested):
Nice work @Ghrehh! Looks great. Went through this with @AndyWhann and listed down some changes we think need done.

- Left panel border is 1px but should be 2px
- Remove gap beside scrollbar for the screen list. They should touch the scrollbar like the data section.
- Show 5 screens instead of 4.5 screens in the screen list
- Border between screens and components is 1px but should be 2px
- Don’t full remount the preview when changing screen. Currently you can swap between screens quickly, but on this branch the preview is fully remounting which slows things down significantly.
- When selecting navigation, the top few pixels get cut off the preview and never come back. Not sure why.
- Remove indentation from top level components so they are in line with navigation and screen components. I think we talked about this already but we just need a more negative indent.
- Update nav items in the screen list to match component list. Too little padding at the moment between the role colours and the text. And too little indentation compared to components.
- Use grey 700 for icons in the left panel (and if you could update the right panel component to use 700 too that would be great)
- Use small size for search icon, keep add medium though. This looks visually more consistent.
- Add hover state for icons
- Wrong font family for search input. It's missing the `var(--font-sans)` font family.
- Undo redo has disappeared. It should be in the top left of the preview.
- Remove show navigation setting from screen settings
- Remove view components button from screen settings
- The right panel component for navigation settings is that unique new panel. I think we talked about updating this to be consistent with the existing panels and use that existing component.
- Configure links button in navigation is broken - clicking it does nothing
- Disabling the navigation should hide the customise section. Currently it doesn't.
- Remove slide up animation from new screen page
- No results text when searching for screens is right at the bottom, it should be at the top
- Collaboration issues (because the screen component and navigation have fixed IDs, selecting them in any screen shows them as active in all screens for other users). Ignore this though and I'll take a look at this. More for my reference.

I'll review the code after these changes are in.

aptkingston:
Nice work on the fixes 👌 I'm just pushing a few final tweaks to this branch directly, rather than going through another round of reviews - just will let you get on with other stuff. Will approve this when I'm done!

aptkingston (approved):

PR #11511, opened 2023-08-13
Improve relationship picker default value

melohagan:
## Description
When adding the default value docs section for the relationship picker, I noticed two main issues that I wanted to improve:

1. Remove the need for the Data provider
2. Remove the need for the JavaScript

An additional setting has been added to the relationship picker that allows you to select if you want the default value to be based on the *Row _id* or the *Display* column. 
You do have to refresh the page to get the new default value to apply - I did look into making this dynamic but didn't have much luck with it.   

Docs page: https://docs.budibase.com/docs/filter-by-relationships#default-value

## Screenshots
Screenshot 2023-08-13 at 14 33 51

#### Internal DB

**Display mode**
Screenshot 2023-08-13 at 14 34 16

**Row _id mode**
Screenshot 2023-08-13 at 14 35 01

**No match when using display values in Row _id mode**
Screenshot 2023-08-13 at 14 35 50

#### SQL

**Display mode (on a Formula column)**
Screenshot 2023-08-13 at 14 36 40

**Row _id mode**
Screenshot 2023-08-13 at 14 37 16


aptkingston:
Just on the use cases for this - if you were planning to hard code in some strings for a default value anyway, why not hard code the row IDs for those rows rather than using the display column? Just wondering what the scenario is that this fixes.

If you're using bindings you're almost certainly going to have IDs available to set the default value, and if you're not using bindings (i.e. you're hard coding something) then can you not hard code the `_id` rather than the display value?

melohagan:
> Just on the use cases for this - if you were planning to hard code in some strings for a default value anyway, why not hard code the row IDs for those rows rather than using the display column? Just wondering what the scenario is that this fixes.
> 
> If you're using bindings you're almost certainly going to have IDs available to set the default value, and if you're not using bindings (i.e. you're hard coding something) then can you not hard code the `_id` rather than the display value?

It is the hard-coding scenario that I'm trying to make a little easier. At the moment you have to know that the relationship picker uses the **_id** field as its value, which is a very Budibase specific thing. For example you may have a City relationship, and you want the default selected city to be "London". The first thing a user thinks is to type "London" into the default value column. 

It's come up a couple of times. Having the docs section obviously helps, but it's not an intuitive thing having Row IDs hard-coded into the default value field. Although it is set to use the *_id* column mode by default to ensure backwards compatibility.  

- https://github.com/Budibase/budibase/discussions/10966
- https://github.com/Budibase/budibase/discussions/7687

PR #11521, opened 2023-08-14
adoptjdk repo has changed to adoptium

jonnymccullagh:
## Description
In the Budibase CouchDB image we had been using AdoptJDK to obtain a JDK version 8. AdoptJDK have now changed to Adoptium and the old repo no longer exists.
As such we need to obtain JDK v8 from their new repo at https://packages.adoptium.net/artifactory/deb

This problem currently affects all builds of the single image as the couchdb image build fails to complete.

We should be able to remove this dependency if we move away from Clouseau in future.

I am also having to make this change for our internal infrastructure.


mike12345567 (approved):

shogunpurple (approved):
LGTM!

Issue #11225, opened 2023-07-13
[BUDI-7269] Data table does not display when table name is matched in a different schema - MS SQL


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.8.10
  * App Version: 2.8.10

**Describe the bug**
When the same table name is used in different schemas for the MS SQL Server connector, the **Data** section does not correctly display either table.

**To Reproduce**

1. Have two schemas that have tables with the same name
2. Create a MS SQL Server datasource for each schema and fetch all tables
3. Click on a non duplicate table name → the table data loads correctly
4. Click on the duplicate table name → the previous table is still showing

**Expected behavior**
When tables share the same name, but exist in different schemas, Budibase should handle each as separate tables.

**Screenshots**

![Linear-hosted image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/6cac07df-3d78-496b-a792-1e8782c2bfd1/1a3759ac-650d-4537-93bc-17b2ba359ef9)
[Image source](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/6cac07df-3d78-496b-a792-1e8782c2bfd1/1a3759ac-650d-4537-93bc-17b2ba359ef9)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7269](https://linear.app/budibase/issue/BUDI-7269/data-table-does-not-display-when-table-name-is-matched-in-a-different)</sub>

PR #11532, opened 2023-08-15
Only get definition for given schema in SQL Server

melohagan:
## Description
Duplicate table names but in different schemas were conflicting when doing a fetch tables. The `getDefinition` method has been updated in the SQL Server connector to only fetch tables for the given schema. 

Addresses: 
- https://github.com/Budibase/budibase/issues/11225


aptkingston (approved):
LGTM but will let Mike confirm 👍

mike12345567 (approved):

PR #11546, opened 2023-08-17
remove enabled from probe values.yaml

jonnymccullagh:
In our example `values.yaml` we had a field `enabled: true` for some of the probes (liveness, readiness). 
This label does not appear in the k8s documentation so it should be removed.


shogunpurple (approved):
LGTM

Issue #11539, opened 2023-08-16
[BUDI-7398] Automation history filtering not working


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud

**Describe the bug**
Automation history filtering is not showing **Stopped** automations.

**To Reproduce**

1. Go the automation history tab of an automation that has an automation that has stopped in the past
2. Filter on Status = 'stopped'
3. No rows are returned

**Expected behavior**
Automation history filter should show stopped automations.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/92de0aa8-dc6b-4714-826b-41ecc374d91b/fc1cf2aa-e56e-48ec-a2fd-0f0b4764c48b?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC85MmRlMGFhOC1kYzZiLTQ3MTQtODI2Yi00MWVjYzM3NGQ5MWIvZmMxY2YyYWEtZTU2ZS00OGVjLWEyZmQtMGYwYjQ3NjRjNDhiIiwiaWF0IjoxNjkyMTkzODE3LCJleHAiOjE2OTIyODAyMTd9.VQJowiaEyxEALOKOhGNpyDB230dCz-t85aBI_ib5A-E)

**Filter on stopped**

[...]

PR #11548, opened 2023-08-17
Automation improvements

mike12345567:
## Description
This is a few minor improvements to the automations system, with the hopes of improvements in performance.

This addresses a couple of minor issues:
1. Issue with filtering and status - make it more obvious what has happened to CRON automations: https://github.com/Budibase/budibase/issues/11539
2. Improved the logging around publishing of apps, to understand the CRON jobs that were stopped and enabled

On top of these minor issues, I've also attempted to add in a feature flagging meaning that the API and automations can be turned on separately. By default everything is enabled so that there is no change, but a new environment variable has been added to the server/worker, `APP_FEATURES` and `WORKER_FEATURES` respectively. For now the worker doesn't have any options, but the `APP_FEATURES` is a comma separated list of the features you wish the instance to support. For example, we can set `automations,api`, which is the current server instance/the default, or we can set `automations` for only automations, `api` for only API.


packages/worker/src/features.ts
```diff
@@ -0,0 +1,26 @@
+import env from "./environment"
+
+enum WorkerFeature {}
+
+const featureList: WorkerFeature[] = processFeatureList()
+
+function processFeatureList() {
```
shogunpurple, line 7: NAB: Might be worth putting this piece of feature flag logic into `backend-core` as it's shared

shogunpurple (approved):
LGTM! This is actually a lot simpler than I anticipated. We could try this out on separate infra in QA/Dev to see if it works

Issue #11549, opened 2023-08-18
[BUDI-7404] Add Bulk Create for SQL datasources


melohagan:
**Describe the feature request**
Whenever I need to bulk insert rows I have done this through automation looping. The performance is very poor with this approach however as it does one insert at a time.

As a user, I would like to be able to add a *BulkInsert* type query for my SQL datasources, so that I do not need to use looping in automations to do a bulk insert.

<sub>[BUDI-7404](https://linear.app/budibase/issue/BUDI-7404/add-bulk-create-for-sql-datasources)</sub>

melohagan:
Requested by Advenir

PR #11556, opened 2023-08-18
Add bulk create for sql datasources

melohagan:
## Description
The current approach of bulk insert is to use automation looping, but this has poor performance. Added a `Bulk Create` query type for:

- MySQL
- Postgres
- SQL Server

It's not very intuitive because you must provide a binding, and then reference that binding in the query, but the performance is much better!

Screenshot 2023-08-18 at 14 55 15

Addresses: 
- https://github.com/Budibase/budibase/issues/11549

## Screenshots

**Before** 👎

https://github.com/Budibase/budibase/assets/101575380/d5789ef8-8a07-405d-a6c4-cd90c37a6de3

**After** 👍 

https://github.com/Budibase/budibase/assets/101575380/f21858df-15da-487a-a5e7-67c453e82903

## Documentation
- [x] https://linear.app/budibase/issue/SUP-98/add-bulkcreate-method-for-sql-datasource


packages/server/src/integrations/base/sql.ts
```diff
@@ -646,6 +646,30 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
...
+
+  async getBulkValues(rowsToInsert: Array<any>, source?: SqlClient) {
+    let finalValues = ""
+    const escapeCharacter = source === SqlClient.POSTGRES ? "E" : ""
+    for (let row of rowsToInsert) {
+      let values = ""
+      for (let value of Object.values(row)) {
+        if (typeof value === "string") {
+          values += `${escapeCharacter}'${value.replace(
+            /[\/#+\-&|!(){}\]^"~'*?:\\]/g,
```
mike12345567, line 658: This regex is a bit hard to understand, what is the purpose here?

packages/server/src/integrations/base/sql.ts
```diff
@@ -646,6 +646,30 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
     }
     return results.length ? results : [{ [operation.toLowerCase()]: true }]
   }
+
+  async getBulkValues(rowsToInsert: Array<any>, source?: SqlClient) {
+    let finalValues = ""
+    const escapeCharacter = source === SqlClient.POSTGRES ? "E" : ""
```
mike12345567, line 652: I know there is some common logic here, but it might be easier to implement this per datasource as it is a query type, rather than here, the `base/sql.ts` is more reserved for datasource plus functionality.

packages/server/src/integrations/base/sql.ts
mike12345567: Is there anyway we could de-duplicate it in that case, at least create a named variable in `shared-core` or `backend-core` that it can be used by Lucene and SQL?

mike12345567 (approved):

melohagan:
After a bit of discussion, this functionality would likely be better as some form of automation action. Closing.

Issue #11554, opened 2023-08-18
[BUDI-7407] Selecting "None" as a data provider's sort column should keep rows in their original order


andz-bb:
**Checklist**

* [X] I have searched Budibase discussions and GitHub issues to check if my issue already exists

**Hosting**

* Self
  * Method: cloud
  * Budibase Version: 2.9.23
  * App Version: 2.9.23

**Describe the bug**
When you choose `None` as the sort column for a data provider, it should leave the data from the server in its original order rather than choosing the first column behind the scenes and sorting by that.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a REST API query with multiple items on the response
2. Use that query in a data provider and ensure that `None` is selected for the sort column
3. See that the data provider is sorting the results by the first column rather than leaving them in the original order from the api response.

**Expected behaviour**
`Default` should be added to the picker with the current functionality where the data is sorted by the first column

`None` should also be an available option for the Sort Column, and when selected, would mean that the rows in the data provider are presented in their original order from the server.

**Screenshots**

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/4448249e-c591-4fde-b3ae-091aefd7b869/2f4a056d-217f-43c0-95c7-abb39eb42402?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC80NDQ4MjQ5ZS1jNTkxLTRmZGUtYjNhZS0wOTFhZWZkN2I4NjkvMmY0YTA1NmQtMjE3Zi00M2MwLTk1YzctYWJiMzllYjQyNDAyIiwiaWF0IjoxNjkyMzY2NzU3LCJleHAiOjE2OTI0NTMxNTd9.i9Xde9lMFAP-Y_YGfGmN7_hOXH2do8eoS4Q0Kc_ZPhk)

**Context**

Raised in [this GitHub discussion](https://github.com/Budibase/budibase/discussions/11543)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7407](https://linear.app/budibase/issue/BUDI-7407/selecting-none-as-a-data-providers-sort-column-should-keep-rows-in)</sub>

aptkingston:
This will be fixed as part of the views V2 PR (it's already fixed on that branch).

[...]

PR #11573, opened 2023-08-21
Views V2

aptkingston:
Apologies for the size of this PR - I realise it's un-reviewable. The scope of this change was absolutely massive and it affects every part of Budibase.

### Description
This PR adds a new implementation of views. A TLDR of a view is that it's an independent view over some table's data, with it's own saved filters, sorting, column selection, access and grid metadata (column order, size, row size, primary display). Views can be updated and saved independently from their parent table.

Data in views is the same data as the table, i.e. updating a row from a view will cause the row in the table to also be updated. It's the same dataset, just viewed differently.

Views can be created for both internal and SQL tables, using the button above the grid.
image
image

Views will be created with whatever the currently applied filters, sorting and table metadata are. Initially it will look like a complete clone of your table, and you can customise it from there.

### Access
Access can be controlled independently from tables. You can have an admin only table, yet create a public view of a subset of that data. 

**Data that is not contained in a view, because you applied filters or hid columns, cannot be retrieved by users. It will not be sent over the API. Views are a secure way of limiting access to table data.**

image

### Apps
Views are drop in replacements for tables inside apps. Anywhere that you can use a table, you can now use a view. They fully support CRUD, and are "datasource plus" in the sense that they support additional server-side filtering, pagination, sorting and limiting.

Views work with data providers and blocks as normal.

You can also create autoscreens from views.

When adding filters to views in an app, these filters are applied *in addition to* the saved filters in the view itself. You cannot use filters to retrieve data that the view does not already contain. Attempting to filter on the same column that a view already contains a saved filter for will do nothing.

Addresses:
- https://github.com/Budibase/budibase/issues/11554
- https://linear.app/budibase/issue/BUDI-7191/%5Bviews-v2%5D-tooltip-when-filters-or-sorting-is-applied-to-a-table
- https://linear.app/budibase/issue/BUDI-7190/%5Bviews-v2%5D-grid-interface-for-views-with-the-ability-to-add-rows


packages/bbui/src/Tabs/Tabs.svelte
```diff
@@ -57,10 +57,8 @@
   function calculateIndicatorLength() {
     if (!vertical) {
       width = $tab.info?.width + "px"
-      height = $tab.info?.height
     } else {
       height = $tab.info?.height + 4 + "px"
-      width = $tab.info?.width
```
aptkingston, line 63: We only need one dimension of sizing for each tab layout. Adding these other sizes caused an issue with tab heights in dev tools.

packages/bbui/src/Modal/Modal.svelte
```diff
@@ -53,6 +54,9 @@
   }
 
   async function focusModal(node) {
+    if (!autoFocus) {
+      return
+    }
```
aptkingston, line 59: Need to add a prop for this because when you have a very long modal this auto focus behaviour scrolls right to the bottom of the modal which is jarring.

packages/builder/src/builderStore/dataBinding.js
```diff
@@ -714,17 +721,25 @@ export const getActionBindings = (actions, actionId) => {
...
  */
-export const getSchemaForTable = (tableId, options) => {
-  return getSchemaForDatasource(null, { type: "table", tableId }, options)
+export const getSchemaForDatasourcePlus = (resourceId, options) => {
+  const isViewV2 = resourceId?.includes("view_")
+  const datasource = isViewV2
+    ? {
+        type: "viewV2",
+        id: resourceId,
+        tableId: resourceId.split("_").slice(1, 3).join("_"),
```
adrinr, line 739: We should use the shared-core to avoid having this code duplicated between front and back

aptkingston: The best match I can find is `extractViewInfoFromID` in the server package - is that the best util to move over to shared-core or is there another util?

packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/DatasourceTemplateRow.svelte
```diff
@@ -0,0 +1,42 @@
...
+    <span class="data-source-check">
+      <Icon size="S" name="CheckmarkCircle" />
+    </span>
+  {/if}
+</div>
+
+<style>
+  .data-source-entry {
+    cursor: pointer;
+    grid-gap: 12px;
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/DatasourceTemplateRow.svelte:22]: Shouldn't we use "space" for gaps like this?

packages/client/src/components/Component.svelte
```diff
@@ -272,12 +272,36 @@
...
+      // Migrate "table" settings to ensure they have a type and resource ID
+      if (setting.type === "table") {
+        const val = instance[setting.key]
+        if (val) {
+          if (!val.type) {
+            val.type = "table"
+          }
+          if (!val.resourceId) {
+            if (val.type === "viewV2") {
+              val.resourceId = val.id || val.tableId
```
adrinr, line 295: When will the id be undefined?

packages/client/src/components/app/blocks/TableBlock.svelte
```diff
@@ -47,7 +47,7 @@
 
   // Accommodate old config to ensure delete button does not reappear
   $: deleteLabel = sidePanelShowDelete === false ? "" : sidePanelDeleteLabel
-
+  $: isDSPlus = dataSource?.type === "table" || dataSource?.type === "viewV2"
```
adrinr, line 50: Again, why is viewv2 considered plus?

packages/frontend-core/src/api/viewsV2.js
```diff
@@ -0,0 +1,72 @@
+export const buildViewV2Endpoints = API => ({
+  /**
+   * Create a new view
```
adrinr, line 3: Wrong comment

packages/frontend-core/src/components/grid/stores/viewV2.js
```diff
@@ -0,0 +1,216 @@
...
+    row.tableId = $datasource?.tableId
+    row._viewId = $datasource?.id
+    return {
+      ...(await API.saveRow(row, SuppressErrors)),
+      _viewId: row._viewId,
+    }
+  }
+
+  const deleteRows = async rows => {
+    // Ensure we delete _viewId from rows as otherwise this throws a 500
```
adrinr, line 36: Not sure if this should be the case anymore

packages/server/src/api/controllers/table/internal.ts
```diff
@@ -44,7 +44,9 @@ function checkAutoColumns(table: Table, oldTable?: Table) {
 export async function save(ctx: UserCtx<SaveTableRequest, SaveTableResponse>) {
   const db = context.getAppDB()
   const { rows, ...rest } = ctx.request.body
-  let tableToSave: TableRequest = {
+  let tableToSave: Table & {
```
adrinr, line 47: Not sure why this change. The old type should be compatible

packages/server/src/api/routes/row.ts
```diff
@@ -269,6 +269,7 @@ router
 
 router.post(
   "/api/v2/views/:viewId/search",
+  paramResource("viewId"),
```
adrinr, line 272: Not sure this will actually work as expected. I will double check with my permissions changes once merged

packages/builder/src/builderStore/dataBinding.js
```diff
@@ -849,12 +876,12 @@ export const getSchemaForDatasource = (asset, datasource, options) => {
 
     // Determine if we should add ID and rev to the schema
     const isInternal = table && !table.sql
-    const isTable = ["table", "link"].includes(datasource.type)
+    const isDSPlus = ["table", "link", "viewV2"].includes(datasource.type)
```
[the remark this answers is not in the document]
aptkingston: As far as client apps are concerned, DS plus basically just means server side searching, sorting and pagination via the rows API. Views support all these, so we can consider it a DS plus to take advantage of those features.

packages/builder/src/components/backend/DatasourceNavigator/DatasourceNavigator.svelte
```diff
@@ -152,10 +163,19 @@
...
     })
-    return table != null
+    if (viewTable) {
+      return true
+    }
+
+    // Check for a matching viewV2
+    const selectedViewV2 = viewsV2.selected?.name
+    const viewV2Table = options.find(table => {
+      return table.views?.[selectedViewV2] != null
```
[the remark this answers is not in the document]
aptkingston: I don't think it will ever be null no, but the loosey `!= null` will account for both null and undefined. But you're right that an explicit `=== undefined` would also be fine.

I think this does need changed though because this would false positive if you had the same view name in 2 different tables. We should use the view's `tableId` property instead.

packages/builder/src/components/backend/DataTable/modals/grid/GridCreateViewModal.svelte
```diff
@@ -0,0 +1,60 @@
...
+  import { Input, notifications, ModalContent } from "@budibase/bbui"
+  import { goto } from "@roxi/routify"
+  import { viewsV2 } from "stores/backend"
+
+  const { filter, sort, definition } = getContext("grid")
+
+  let name
+
+  $: views = Object.keys($definition?.views || {})
+  $: nameExists = views.includes(name?.trim())
```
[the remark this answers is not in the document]
aptkingston: I suppose there's no issue with allowing the same name with different casing, since keys in JS objects are case sensitive. I guess we can add in case checking just for consistency since we do it in other places, but I don't think there will be any bugs due to casing. I'll add this in.

packages/client/src/components/app/blocks/TableBlock.svelte
aptkingston: Answered this above.

packages/frontend-core/src/components/grid/stores/viewV2.js
aptkingston: I'll check and remove this if no longer needed.

packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/DatasourceTemplateRow.svelte
Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/DatasourceTemplateRow.svelte:22]: I actually just moved this into a new file rather than writing it, but in general yes it's better to use those variables. The designs don't usually use those variables though, so it's sometimes easier to use pixels to ensure we're matching the designs. I used to be a real stickler about this myself, but now I don't really mind.

packages/client/src/components/Component.svelte
aptkingston: Should never be undefined! Good catch.

packages/frontend-core/src/components/grid/stores/viewV2.js
aptkingston: No longer needed in the store because the API client has this deletion code, so I can remove it from here indeed - nice catch.

The logic is still needed is general though, as if we do not remove `_viewId` then the rows API will throw a 400 with `Bad special document member _viewId`.

packages/server/src/api/controllers/table/internal.ts
aptkingston: Me neither, but I think it was maybe a commit you made yourself?
image

aptkingston: Unless maybe this changed again with your recent views permission code? According to my git history it was never of type `TableRequest`. This was maybe changed upstream in develop?
image

packages/builder/src/builderStore/dataBinding.js
adrinr: This might require a refactor for now, so better to note it and to tackle it separately in the future

packages/server/src/api/controllers/table/internal.ts
adrinr: Funny 😅 I will have a look at it

packages/builder/src/components/backend/DatasourceNavigator/DatasourceNavigator.svelte
```diff
@@ -171,10 +171,7 @@
     }
 
     // Check for a matching viewV2
-    const selectedViewV2 = viewsV2.selected?.name
-    const viewV2Table = options.find(table => {
-      return table.views?.[selectedViewV2] != null
-    })
+    const viewV2Table = options.find(x => x._id === viewsV2.selected?.tableId)
```
adrinr, line 174: Just to double check, is this meant to be tableId or view.id?

aptkingston: Table - this code is trying to check if a certain datasource (e.g. a postgres datasource) contains the active entity (which could be a table or view or anything). This particular line is trying to find a table in that datasource that contains the selected view. So we're trying to find a table from `options` (which is the list of tables in this datasource) which has the ID of the selected view table ID. If we find a table, it means we contain the active entity.

PR #11581, opened 2023-08-23
Group support for per app builders

mike12345567:
## Description
As title says, this is the backend support for two new API endpoints, allowing groups to be used to specify per app builder permissions.

This was largely implemented in the middlewares, there maybe a need for caching at some point in the future but didn't want to prematurely add this as groups are an underutilised feature today and caching could create weird RBAC bugs allowing people access to apps that have been revoked/not being able to access things they have been given.

I also did a large refactor of the `processUser` function, which is core to how a lot of middlewares work, pulling the user then processing and enriching it based on the app in context. It was previously very hard to follow so I compressed a lot of the excess logic down and then extended it to include the logic for per app builders from groups as well.

This addresses: https://linear.app/budibase/issue/BUDI-7411/per-app-developer-support-for-groups

Some minor changes were needed to pro as I wanted to better contain the logic for this. I've added a new function to pro that allows getting the list of app IDs that a user would have per app builder access to based on their user groups as well as the APIs needed to updated groups (add/remove the per app builder permission). Pro PR: https://github.com/Budibase/budibase-pro/pull/218


PClmnt (approved):
LGTM, that pio

PR #11585, opened 2023-08-24
Allow setting view permissions

adrinr:
## Description
Allow setting view permissions, overriding the table permissions. This behaviour will be under a feature flag: when off, only the table permissions will be 

Addresses: 
- [BUDI-7393](https://linear.app/budibase/issue/BUDI-7393/commercial-updates-views)


shogunpurple (approved):
LGTM!

PR #11596, opened 2023-08-25
Per app builder updates in builder side panel

PClmnt:
## Description
This PR adds on the work done by @mike12345567 in order to allow builders on a per app basis. To enable this, a user can now select a `Creator` from within the role select dropdown in the users side panel. 

I've also updated the side panel somewhat as requested by in the designs, to allow for invites / groups / users to load in immediately, these are then sorted by their access type. With users / groups without access being pushed to the bottom of the list. 

image

When inviting a user there is a new flow

Screenshot 2023-08-29 at 14 54 18

Screenshot 2023-08-29 at 14 55 44

Screenshot 2023-08-29 at 15 02 45
- Admins can invite users as members or admin. But other users can only invite users as Members. 

Screenshot 2023-08-29 at 15 02 49


packages/builder/src/components/common/RoleSelect.svelte
```diff
@@ -15,17 +17,42 @@
...
       return roles.filter(role => allowedRoles.includes(role._id))
     }
     let newRoles = [...roles]
+
+    if (allowCreator) {
+      newRoles = [
+        {
+          _id: CreatorID,
+          name: "Creator",
+          tag: !$licensing.perAppBuildersEnabled && "Business",
```
shogunpurple, line 51: Should probably be a constant

packages/builder/src/components/common/RoleSelect.svelte
```diff
@@ -58,25 +85,61 @@
...
+    {footer}
+    bind:value
+    on:change={onChange}
+    {options}
+    label="Access on this app"
+    getOptionLabel={role => role.name}
+    getOptionValue={role => role._id}
+    getOptionColour={getColor}
+    getOptionIcon={getIcon}
+    isOptionEnabled={option => {
```
shogunpurple, line 111:
```suggestion
    isOptionEnabled={option => option._id !== CreatorID || $licensing.perAppBuildersEnabled}
```

packages/builder/src/components/common/RoleSelect.svelte
```diff
@@ -58,25 +85,61 @@
...
+    {footer}
+    bind:value
+    on:change={onChange}
+    {options}
+    getOptionLabel={role => role.name}
+    getOptionValue={role => role._id}
+    getOptionColour={getColor}
+    getOptionIcon={getIcon}
+    isOptionEnabled={option => {
+      if (option._id == CreatorID && !$licensing.perAppBuildersEnabled) {
```
shogunpurple, line 136: Same here

packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
```diff
@@ -301,11 +357,20 @@
     const payload = [
       {
         email: newUserEmail,
-        builder: false,
-        admin: false,
-        apps: { [prodAppId]: Constants.Roles.BASIC },
+        builder:
+          creationRoleType === Constants.BudibaseRoles.Admin ? true : false,
```
Comment (shogunpurple) [inline packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:361]:
```suggestion
          !!creationRoleType === Constants.BudibaseRoles.Admin,
```

shogunpurple (approved):
LGTM, some minor comments.

packages/builder/src/components/common/RoleSelect.svelte
```diff
@@ -58,25 +88,51 @@
   const onChange = e => {
     if (allowRemove && e.detail === RemoveID) {
       dispatch("remove")
+    } else if (e.detail === CreatorID) {
+      dispatch("addcreator")
```
aptkingston, line 92: NAB: Is there a special reason that choosing creator fires a unique event? Just listening to the normal `change` event and then checking the payload is creator would achieve the same thing wouldn't it? Just thinking along the lines of trying to keep this simple.

packages/builder/src/pages/builder/portal/users/users/[userId].svelte
```diff
@@ -119,7 +120,11 @@
         name: app.name,
         devId: app.devId,
         icon: app.icon,
-        role: privileged ? Constants.Roles.ADMIN : roles[prodAppId],
+        role: privileged
+          ? Constants.Roles.ADMIN
+          : user?.builder?.apps.includes(prodAppId)
+          ? Constants.Roles.CREATOR
+          : roles[prodAppId],
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/portal/users/users/[userId].svelte:127]: NAB: don't need to change this, but in general I think these multi condition ternary statements are very hard to read and understand. In my opinion, if it's anything more than a basic ternary statement then it should be a dedicated function.

packages/frontend-core/src/api/user.js
```diff
@@ -156,13 +156,14 @@ export const buildUserEndpoints = API => ({
...
       body: payload.map(invite => {
-        const { email, admin, builder, apps } = invite
+        const { email, admin, builder, apps, appBuilders } = invite
         return {
           email,
           userInfo: {
             admin: admin ? { global: true } : undefined,
             builder: builder ? { global: true } : undefined,
             apps: apps ? apps : undefined,
+            appBuilders: appBuilders ? appBuilders : undefined,
```
aptkingston, line 166: NAB: doing a conditional with `undefined` won't ever change anything (other than technically turning explicit `null` or `false` into `undefined`, but I doubt that's needed). So this can just be `appBuilders,`.

packages/frontend-core/src/api/user.js
```diff
@@ -175,10 +176,12 @@ export const buildUserEndpoints = API => ({
    * @param invite the invite code sent in the email
    */
   updateUserInvite: async invite => {
+    console.log(invite)
     await API.post({
       url: `/api/global/users/invite/update/${invite.code}`,
       body: {
         apps: invite.apps,
+        appBuilders: invite.appBuilders ? invite.appBuilders : undefined,
```
aptkingston, line 184: Same as above, conditional isn't needed.

aptkingston (approved):

PR #11597, opened 2023-08-25
Form/Table block UX Updates

deanhannigan:
## Description

UX updates to the FormBlock/TableBlock. Adds a new field configuration UX allowing top-level form layout customisation and granular control of all the field controls within.

Addresses: 
- BUDI-7310 - [Form block improvements] Create form
- BUDI-7309 - [Form block improvements] Update-form
- BUDI-7308 - [Form block improvements] View-form

Additional Updates:
- Form block bindings - `grouping by data context`. `Field` and `Row` bindings are now filtered into their own respective categories prefixed with the component name. It was difficult to discern at a glance what they were as they share the same names.
- Form block bindings - `filtering by Form Block Action Type`. There would be a confusing number of bindings that did not relate to the current FormBlock type. They will now filter based on the selected FormBlock  `Action Type` setting:
  - `Create` - Form Fields only
  - `Update` - Form Fields and Row data
  - `View` -  Row data only
- ComboBox component now using core `position_dropdown/click_outside` behaviours due to UX issues with its current implementation.

Table block updates
- Field configuration updates have been extended to the `details side panel form`. 
- Save/Delete button UX has been updated to reflect the same design as the FormBlock.

Bug Fixes:
- Bug - No FormBlock `Field` Bindings available when creating a new FormBlock with no field config. 
- Bug - Invalid FormBlock `Field` Bindings generated. When fields were eventually added to the block, the were not being parsed correctly into the list. No matter what fields you added, only one invalid binding would display.

## Screenshots
The updated form UX
menu

FormBlock UX
formblock

TableBlock UX
tableblock

Bindings Updates
bindings


aptkingston (commented):
Looks absolutely fantastic @deanhannigan! Still in the middle of reviewing this, but when using it I noticed that when you set up "on change" actions, the binding for the new value  (which is `{{ Field Value }}` in a category called `New Text Field`, or whatever the instance name is) seems to be missing. Thought you could maybe a take a look at that in the meantime while I continue reviewing!

packages/builder/src/components/design/settings/controls/FieldConfiguration/EditFieldPopover.svelte
```diff
@@ -0,0 +1,160 @@
...
+  export let field
+  export let componentBindings
+  export let bindings
+
+  const draggable = getContext("draggable")
+  const dispatch = createEventDispatcher()
+
+  let popover
+  let drawers = []
+  let sudoComponentInstance
```
aptkingston, line 19: As much as I love something called `sudoComponentInstance` I think this should probably be `pseudoComponentInstance` should it? Or maybe even `fake` or `tempt` to make it easier to type.

packages/builder/src/components/design/settings/controls/FieldConfiguration/EditFieldPopover.svelte
```diff
@@ -0,0 +1,160 @@
...
+    if ($draggable.selected == field._id) {
+      $draggable.actions.select()
+    }
+  }}
+  {anchor}
+  align="left-outside"
+  showPopover={drawers.length == 0}
+  clickOutsideOverride={drawers.length > 0}
+  maxHeight={600}
+  handlePostionUpdate={(anchorBounds, eleBounds, cfg) => {
```
aptkingston, line 95: Nice work on getting the position to work well for now! I think in future we'll definitely do further work on the position dropdown and build in things like screen edge awareness, but definitely happy to proceed with this for now like we discussed.

packages/bbui/src/Drawer/Drawer.svelte
```diff
@@ -4,27 +4,34 @@
...
+
   let visible = false
+  let drawerId = generate()
 
   export function show() {
     if (visible) {
       return
     }
     visible = true
+    dispatch("drawerShow", drawerId)
```
aptkingston, line 26: Just wondering if you ever use this drawer ID? I can't see it being used, and you have to pass `on:drawerShow` handler to an instance of a drawer anyway, so you'll probably know which drawer this is.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -1,45 +1,74 @@
...
+  export let value
 
   const dispatch = createEventDispatcher()
+  let sanitisedFields
+  let fieldList
+  let schema
+  // let assetIdCache
+  let cachedValue
+  // $: value, console.log("VALUE UPDATED")
+  // $: $currentAsset, console.log("currentAsset updated ", $currentAsset)
```
aptkingston, line 26: Some commented out code which we could remove.

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
```diff
@@ -198,22 +211,19 @@
...
-                field: field.name,
-                label: field.displayName,
-                placeholder: field.displayName,
-              }}
-              order={idx}
-            />
-          {/if}
-        {/each}
-      </BlockComponent>
+      {#key fields}
```
aptkingston, line 214: Do we need to key this whole thing? This will mean every field will remount whenever any field setting changes. We could maybe instead add in the keying param to the `#each` loop, depending on what property requires remounting - e.g. `{#each fields as field, idx (field._component}` or whichever property it is that requires remounting.

aptkingston (approved):
LGTM! Just that one issue with the on change value binding missing that I mentioned previously.

packages/bbui/src/Drawer/Drawer.svelte
deanhannigan: Its picked up and stored in the `EditFieldPopover` component to help discern how many unique drawers have been opened.

PR #11615, opened 2023-08-29
Support views on plus datasources

adrinr:
## Description
Fixing view backend to support plus datasources. Datasource plus tables are stored in the ds document instead of a separate the table doc, so we need to update the documents accordingly


shogunpurple (approved):
LGTM!

Issue #11409, opened 2023-07-31
[BUDI-7362] Expose global Binding which Theme is in use


poirazis:
a global binding , possibly under Device, which will tell us which theme is in use to adjust coloring with conditionals

<sub>[BUDI-7362](https://linear.app/budibase/issue/BUDI-7362/expose-global-binding-which-theme-is-in-use)</sub>

deanhannigan:
PR [https://github.com/Budibase/budibase/pull/11617](https://github.com/Budibase/budibase/pull/11617)

melohagan:
There is a whole section for bindings, so a page could be created in there for the global bindings

PR #11617, opened 2023-08-30
App Theme binding added

deanhannigan:
## Description

Added a new binding, `{{App.Theme}}` to the `Device` binding category. 

All currently supported themes will be prefixed by `spectrum--`

Addresses: 
-  https://github.com/Budibase/budibase/issues/11409

## Screenshots

The binding can be found in the Device section in the binding drawer.

Screenshot 2023-08-30 at 09 28 43

Updating when the theme is changed

bindings


melohagan (approved):
LGTM! The only thought I have is if *Device* is the correct category for the binding?

I.e. if the theme comes from the Budibase app. Don't want to give the impression that the theme comes from the device itself?

aptkingston:
> LGTM! The only thought I have is if _Device_ is the correct category for the binding?
> 
> I.e. if the theme comes from the Budibase app. Don't want to give the impression that the theme comes from the device itself?

We decided on device just because it was the closest matching category we have. If we add more "app" level stuff down the line then I'd definitely break this out into a new dedicated catergory, but it felt unnecessary to create a new category for one binding.

The readable binding is also `App.Theme` to attempt to help with this.

aptkingston:
@deanhannigan is this ready to review? It's marked as a draft at the moment.

aptkingston (approved):
LGTM!

PR #11632, opened 2023-08-31
View bug fixes

deanhannigan:
## Description

- `JSON with Schema` export option now filtered out when a calculation is applied to a view.  This was producing invalid exports.
- Minor check to ensure that when exporting with a schema that only valid columns are included.
- If attempting to use a view with a TableBlock and the option `Show new row side panel` is set, the builder preview will crash.
  - Create a view and add it as the TableBlock data source
  - Ensure `Show button` is enabled
  - Select the `Show new row side panel` radio option
  - There should be an exception in the console and the builder preview will have crashed
  - The only way to recover is to ensure `Show new row side panel` is deselected


aptkingston (approved):
LGTM!

PR #11635, opened 2023-08-31
Fix daysjs import error

jvcalderon:
## Description

The following error was occurring in the development environment:

```
Uncaught SyntaxError: The requested module '/node_modules/@budibase/frontend-core/node_modules/dayjs/dayjs.min.js?v=9d05bfea' does not provide an export named 'default'
```

This PR just replace `import dayjs` by `import { dayjs }` to solve the problem.


shogunpurple (approved):
LGTM

aptkingston:
I'm pretty sure the old syntax was correct. Now on developI 'm getting errors about dayjs importing:
image

jvcalderon:
@aptkingston If I don't remember wrong the problem was being produced while developing in account portal, but to be honest currently it works fine with the old import. We can revert this it is producing problems.

aptkingston:
Yea this was probably just a slight version mismatch with your installed packages. On my next PR I'll set the versions of dayjs to the same one across the board, and then ensure the same syntax is used everywhere.

Issue #11607, opened 2023-08-29
[BUDI-7426] Relationship picker searching enhancement


ConorWebb96:
**Describe the feature request**
Add the ability to search relationships with the relationship picker component, currently the autocomplete option is capped to 100 items. (Optionally allow this to be increased depending on the users needs.)

**Related Discussions**

* [https://github.com/Budibase/budibase/discussions/11558](https://github.com/Budibase/budibase/discussions/11558)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7426](https://linear.app/budibase/issue/BUDI-7426/relationship-picker-searching-enhancement)</sub>

aptkingston:
I assume you mean searching from inside the component itself? Since we did recently add relationship picker filtering as a builder setting.

JCV67:
Thank you @ConorWebb96

Issue #11478, opened 2023-08-08
[BUDI-7375] Relationship picker incorrectly renders selections, and not all records show in the dropdown


kpturner:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: monito

**Describe the bug**
A relationship picker does not render selections properly, and does not provide all the available rows for selection

**To Reproduce**
Created a relationship between two simple tables - `country` and `currency`.  A `country` can have multiple `currency` records associated with it. The data is held in a `country_currency` table that just contains the primary key of `country` and the primary key of `currency`.  The data is correctly rendered in the grid in the "data" section of the designer - for example:

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/25437d62-e8f8-42f6-bc9a-996ab9bb7d40/5d16af49-d7b5-41c8-992b-a5f38b738dae?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8yNTQzN2Q2Mi1lOGY4LTQyZjYtYmM5YS05OTZhYjliYjdkNDAvNWQxNmFmNDktZDdiNS00MWM4LTk5MmItYTVmMzhiNzM4ZGFlIiwiaWF0IjoxNjkxNTcyMTU5LCJleHAiOjE2OTE2NTg1NTl9.VGqQDQqK3M9uCBtx2XnVSVFdeWZ1cLBatM7kYADMdnA)

But within the application where a relationship picker is used, `USD` shows as `%5B'USD'%5D`

[...]

PR #11639, opened 2023-08-31
Relationship picker searching enhancement

melohagan:
## Description
Users ran into some difficulty with the new relationship picker filtering because of the hard-coded limit of 100 records.

Search functionality has now been added (mostly copied from the Data section relationship cell) so that users may find related records outside of the first 100 records.

The UI of the search may need improved - open to suggestions on that!

- bug fixed for the *Update* type forms to ensure that the field value was searched for and selected
- upon search, the picker is opened (and a spinner is shown)
- when a value is selected, it is appended to the end of the list so that a **null** value does not appear
- the *autocomplete* setting has been replaced with the *Search* setting which is enabled by default 
- if a filter is applied, then the search will be client-side against the filtered options

Addresses: 
- https://github.com/Budibase/budibase/issues/11478
- https://github.com/Budibase/budibase/issues/11607

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/cdfa2aab-b93c-40a1-8c23-0f5838883552

**Picker above**
Screenshot 2023-08-31 at 20 38 57

**Picker below**
Screenshot 2023-08-31 at 20 40 30

## Documentation
- [x] Will add if merged


packages/bbui/src/Actions/position_dropdown.js
```diff
@@ -47,7 +48,7 @@ export default function positionDropdown(element, opts) {
         styles.top = anchorBounds.top - elementBounds.height - offset
         styles.maxHeight = maxHeight || 240
       } else {
-        styles.top = anchorBounds.bottom + offset
+        styles.top = anchorBounds.bottom + (offsetBelow || offset)
```
aptkingston, line 51: Could we not just use the existing `offset` property here?

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -77,10 +134,66 @@
...
+    // Reset state if this search is invalid
+    if (!linkedTableId || !searchString) {
+      searchResults = null
+      return
+    }
+
+    // If a filter exists, then do a client side search
+    if (hasFilter) {
+      searchResults = $fetch.rows.filter(option =>
+        option[`${primaryDisplay}`].startsWith(searchString)
```
aptkingston, line 159: NAB: You don't need a JS template string when your property is already a string - so you could just write this as `option[primaryDisplay]`. I've seen quite a few instances of this so I'll not comment on them all, but just letting you know that you could shorten them all to this syntax.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -35,13 +50,55 @@
...
+        tableId: linkedTableId,
+        limit: 100,
+        query: {
+          oneOf: {
+            [`1:${primaryDisplay}`]: fieldState?.value?.map(
+              value => value.primaryDisplay
+            ),
+          },
+        },
+      }).then(response => {
```
aptkingston, line 96: I'm not sure what the purpose of this is... I can see that you're searching the table to retrieve the full rows for each of the selected option, but I don't know why that's needed. The data we already have in the initial row includes the primary display of each option, so we have all the data we need to be able to render the selected values.

Perhaps the bigger limitation here is the fact that it's a multi select and depends upon having "options" available for each selected value.

aptkingston (commented):
This is a complicated feature, so well done for getting it working Mel. If users are waiting for this then I don't mind if we go ahead and merge this in, but in general it would probably be a good call to run changes like this past product or design - it's important we present a consistent experience to users.

I think the few "issue" (not really issues, more like just non-ideal things) that I see with this are mostly down to the fact that it uses a multi select component (which was not your doing, I know, you're working with that you've got here). I think we need to write a brand new relationship picker component that's a copy of the grid one, to get around all those issues.

So while I'm happy to merge this in for now, I think we should definitely schedule in a piece of work to write a brand new relationship picker that doesn't have the limitations of using a multi-select component under the hood.

packages/bbui/src/Actions/position_dropdown.js
melohagan: I did try that initially, but unfortunately it didn't work out because the offset is applied when the popover is above the field, or below the field.
In this case I wanted to use the standard offset when above the field, but a large offset when below the field (because of the search input), as shown in the screenshots.

packages/client/src/components/app/forms/RelationshipField.svelte
melohagan: This is required to fix the bug were the selected option(s) were null if not within the first 100 records. 

It also ensures that the options list include the initially selected options so they can be deselected without having to search for them, as demonstrated in the video.

packages/client/src/components/app/forms/RelationshipField.svelte
melohagan: Good catch - updated1

melohagan:
> If users are waiting for this then I don't mind if we go ahead and merge this in, but in general it would probably be a good call to run changes like this past product or design - it's important we present a consistent experience to users.

I definitely agree with this, and also admit I haven't been running things past product as I should be! Going forward myself and the support team are looking to do a product triage which will give us the chance to discuss changes up front. 

<br />

> I think the few "issue" (not really issues, more like just non-ideal things) that I see with this are mostly down to the fact that it uses a multi select component (which was not your doing, I know, you're working with that you've got here). I think we need to write a brand new relationship picker component that's a copy of the grid one, to get around all those issues.
> 
> So while I'm happy to merge this in for now, I think we should definitely schedule in a piece of work to write a brand new relationship picker that doesn't have the limitations of using a multi-select component under the hood.

The problem is that for a handful of users, the record limit was a breaking change to some extent, in that they had to replace their relationship pickers with data provided multi-selects and/or options pickers. Because they were in the cloud environment, they were unable to rollback.

There's a bit of discussion here: https://github.com/Budibase/budibase/discussions/11558

aptkingston (approved):
Alright, I'm happy to merge this as discussed in order to restore functionality for those users who were affected by the new 100 record limit. Thanks for taking a look at this Mel!

PR #11643, opened 2023-09-01
Add missing grid tooltip for views + form fixes

aptkingston:
## Description
- Adds a tooltip to tables when initially applying filters, to highlight that you can create a view to save your filters
- Fixes a crash when using formula or nested JSON fields in a table then using that table in a form block


packages/builder/src/components/design/settings/controls/DraggableList.svelte
```diff
@@ -33,17 +33,19 @@
...
-    })
+  const buildDraggable = items => {
+    return items
+      .map(item => {
+        return {
+          id: listItemKey ? item[listItemKey] : generate(),
+          item,
+        }
+      })
+      .filter(item => item.id)
```
aptkingston, line 44: Filtering out nullish IDs, as otherwise this causes duplicate keys and crashes the page. They should never be null in the first place, but this is added protection since this is where the crash was occurring when using formula fields.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -99,6 +99,9 @@
     }
 
     const type = getComponentForField(instance.field, schema)
+    if (!type) {
+      return null
+    }
```
aptkingston, line 104: Skip fields which we don't render in forms - e.g. formula and nested JSON.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -116,7 +119,9 @@
   }
 
   $: if (sanitisedFields) {
-    fieldList = [...sanitisedFields, ...unconfigured].map(buildSudoInstance)
+    fieldList = [...sanitisedFields, ...unconfigured]
+      .map(buildSudoInstance)
+      .filter(x => x != null)
```
aptkingston, line 124: Filter out the invalid fields that we just returned `null` for.

deanhannigan (approved):
LGTM!

PR #11644, opened 2023-09-01
Handle inheritance permissions

adrinr:
## Description
Change permissions to support inheritance. From now on, a resource will be able to get the permissions defined in the different ways:
1. Explicit: the permission for the role has been set to something specific
2. Base: the permission to the resource has been defined by the role default access
3. Inherited: the permission to the resource has been defined by the parent resource (that might, at the same time, be defined explicitly, via base or inherited from its parent)

Inherited will only be allowed on certain conditions. For now, it's only for views on paid users.

This required some API changes to be able to expose this new information.

The new functionality is displayed in the following fashion:
### Default permissions for a view (for a paid user)
image
image

### Permissions for a table with views that inherit
image

### Permissions for a table without views or without views that inherit
image

### Permissions for a view when no setting permissions is allowed
image

Addresses: 
- [BUDI-7393](https://linear.app/budibase/issue/BUDI-7393/commercial-updates-views)

## Documentation


packages/builder/src/components/backend/DataTable/modals/ManageAccessModal.svelte
```diff
@@ -9,58 +9,122 @@
...
     } catch (error) {
       notifications.error("Error updating permissions")
     }
   }
+
+  $: computedPermissions = Object.entries(permissions.permissions).reduce(
+    (p, [level, roleInfo]) => {
+      p[level] = {
+        selectedValue:
+          roleInfo.permissionType === "INHERITED" ? "inherited" : roleInfo.role,
```
aptkingston, line 48: NAB: I noticed you have the permission types enumerated in the server as `PermissionSource`. Is it worth moving those somewhere that could be shared and referencing the constant here rather than rely on a magic string?

aptkingston (approved):
LGTM!

packages/builder/src/components/backend/DataTable/modals/ManageAccessModal.svelte
adrinr: Good catch. Done!

mike12345567 (commented):
LGTM! Nice to have some more typing around the permissions.

PR #11655, opened 2023-09-04
Use permission sdk on middleware

adrinr:
## Description
Change middleware to use permission SDK in order to unify logic around permissions, such as inheritance.
Doing so, the permissions endpoint and the auth middleware will be always align on its business logic

Addresses: 
- [BUDI-7393
](https://linear.app/budibase/issue/BUDI-7393/commercial-updates-views)


mike12345567 (approved):
Good improvement, nice to bring this all into the permissions SDK!

PR #11657, opened 2023-09-04
ensure secrets are not replaced on every helm run

shogunpurple:
## Description
Fix for https://github.com/Budibase/budibase/pull/11594

- Prevent the helm chart from updating it's own secrets on upgrade, which leaves the real secrets out of sync with what's in helm
- Reported by a customer


pcnoic (approved):
lgtm

calexiou (approved):

PR #11662, opened 2023-09-05
Per app builder updates / fixes

PClmnt:
## Description
Fixes a few things found in QA

- In some cases Admins weren't being added correctly
- Fix counting of apps in groups
- Update how invite roles are assigned
- Change structure of invite object to more closely match the user object (specifically the apps array)
- Remove invite button when on the invite side panel


packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
```diff
@@ -366,18 +363,21 @@
...
+        admin: { global: creationRoleType === Constants.BudibaseRoles.Admin },
       },
     ]
 
-    if (creationAccessType === Constants.Roles.CREATOR) {
-      payload[0].appBuilders = [prodAppId]
-    } else {
-      payload[0].apps = {
-        [prodAppId]: creationAccessType,
+    if (creationRoleType !== Constants.BudibaseRoles.Admin) {
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:371]: can this be simplified? The if in an if, could it just be:
```
const notCreatingAdmin = creationRoleType !== Constants.BudibaseRoles.Admin
const isCreator = creationAccessType === Constants.Roles.CREATOR
if (notCreatingAdmin && isCreator) {
  payload[0].builder.apps = [prodAppId]
} else if (notCreatingAdmin && !isCreator) {
  payload[0].apps = {
    [prodAppId]: creationAccessType,
  }
}
```
Bit easier to understand what is happening here.

mike12345567 (approved):
LGTM! Just a wee line that was hard to read with a suggestion!

PR #11665, opened 2023-09-05
Fix for datasource reset on formblock. Added 'resetOn' behaviour to the component manifest

deanhannigan:
## Description

Fixes for FormBlock issues:
- Field configuration correctly updates available fields in the component settings panel when switching between data sources.
- Existing field configuration is now cleared after switching datasources
- Missing save button on TableBlock create new row form.

Update: 
- Added a `resetOn` property to the component manifest. The property takes the string name of another property in the component definition. When executing an update, any configured `resetOn` values will be cleared.
- The `resetOn` property also ensures reactive changes to the config create a single entry to the builder undo/redo history.

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -1250,9 +1250,7 @@ export const getFrontendStore = () => {
             setting => name === setting.resetOn
           )
           resetFields?.forEach(setting => {
-            if (component[setting.key]) {
-              component[setting.key] = null
-            }
+            component[setting.key] = null
           })
 
           if (
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -1250,7 +1250,6 @@ export const getFrontendStore = () => {
             setting => name === setting.resetOn
           )
           resetFields?.forEach(setting => {
-            console.log(setting.key, component)
             if (component[setting.key]) {
               component[setting.key] = null
             }
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -1246,6 +1246,16 @@ export const getFrontendStore = () => {
           const settings = getComponentSettings(component._component)
           const updatedSetting = settings.find(setting => setting.key === name)
 
+          const resetFields = settings.filter(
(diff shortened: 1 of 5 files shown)


packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1246,6 +1246,15 @@ export const getFrontendStore = () => {
           const settings = getComponentSettings(component._component)
           const updatedSetting = settings.find(setting => setting.key === name)
 
+          const resetFields = settings.filter(
+            setting => name === setting.resetOn
+          )
+          resetFields?.forEach(setting => {
+            if (component[setting.key]) {
```
aptkingston, line 1253: I would probably just always set this to null, and remove the truthy check - there might be values like `false` or `0` or `""` which we should reset, but we would skip them with this check here since they are loosely falsey.

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
```diff
@@ -211,17 +211,19 @@
...
-          {#if getComponentForField(field) && field.active}
-            <BlockComponent
-              type={getComponentForField(field)}
-              props={getPropsForField(field)}
-              order={idx}
-            />
-          {/if}
-        {/each}
-      </BlockComponent>
+      {#key fields}
```
aptkingston, line 214: I have a feeling I mentioned this before, but what's the purpose of the full remount of all fields whenever any field changes? I think I recommended just keying the `each` statement based on whatever property of the fields requires a full remount whenever they change.

aptkingston (commented):
Nice work Dean! Just one question and one recommended change.

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
deanhannigan: I ran into an issue with the `each` statement where the ordering of the list would update correctly, but some of the form components would still not reflect updates made in the settings popover. They did sometimes update, but it was inconsistent. I had a look at the `DateTimeField` and the `redrawOptions` values would only update sporadically. 

It might be possible to ensure that only the updated fields are re-rendered but keying the form fields with the `fields` object seemed to cover the field ordering and settings updates in one go.

aptkingston (approved):

PR #11689, opened 2023-09-06
Add option to duplicate columns

aptkingston:
## Description
This PR adds the ability to duplicate columns in grids. The name is automatically generated, and will increment as required. All column metadata is duplicated.

image
image

Working with options, copying over metadata successfully:
image

Automatic incrementing of names when duplicating again:
image

Addresses: 
- https://linear.app/budibase/issue/BUDI-7418/duplicate-a-column-in-data-section


aptkingston:
Await release before merging.

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
```diff
@@ -118,6 +120,33 @@
...
   }
 
+  const duplicateColumn = async () => {
+    open = false
+
+    // Generate new name
+    let newName = `${column.name} copy`
+    let attempts = 2
+    while ($definition.schema[newName]) {
+      newName = `${column.name} copy ${attempts++}`
```
melohagan, line 130: What happens when it runs out of attempts, does it just throw an error?

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
```diff
@@ -118,6 +120,33 @@
...
+    const existingColumnDefinition = $definition.schema[column.name]
+    await datasource.actions.saveDefinition({
+      ...$definition,
+      schema: {
+        ...$definition.schema,
+        [newName]: {
+          ...existingColumnDefinition,
+          name: newName,
+          schema: {
+            ...existingColumnDefinition.schema,
```
melohagan, line 143: If we're already spreading the `existingColumnDefinition` on line 140, so we need specify the schema property here?

melohagan (approved):
LGTM - handy feature!

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
aptkingston: Attempts increments, so it never runs out!

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
aptkingston: We need to change the schema from the existing column definition, so this is how you copy and change a nested property without mutating the base! If I didn't spread the base schema then schema would only have whatever properties specified here.

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
melohagan: Ah yes, nice!

PR #11706, opened 2023-09-07
Customize signup flow BB changes

Ghrehh:
Corresponding bb changes to https://github.com/Budibase/account-portal/pull/393

https://linear.app/budibase/issue/BUDI-7302/[p1]-this-first[email-validation]-customize-sign-up-flow

### DIFF
--- a/packages/builder/src/pages/builder/_layout.svelte
+++ b/packages/builder/src/pages/builder/_layout.svelte
@@ -19,26 +19,26 @@
   $: useAccountPortal = cloud && !$admin.disableAccountPortal
   let showVerificationPrompt = false
 
-  const checkVerification = (user) => {
-    console.log('checking')
+  const checkVerification = user => {
     if (!showVerificationPrompt && user?.account?.verified === false) {
-      console.log('show it')
       showVerificationPrompt = true
-      banner.queue([{
-        message: `Please verify your account. We've sent the verification link to ${user.email}`,
-        type: BANNER_TYPES.NEUTRAL,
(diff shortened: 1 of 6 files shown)


packages/builder/src/pages/builder/_layout.svelte
```diff
@@ -16,6 +17,32 @@
...
+              body: JSON.stringify({ email: user.email }),
+            })
+          },
+          extraButtonText: "Resend email",
+        },
+      ])
+    }
+  }
+
+  $: checkVerification(user)
```
aptkingston, line 45: Since this is just running in a reactive statement, if this retriggered (say because a user updated their name before verifying) would this cause a second banner to appear, or will their only be one? As long as multiple invocations don't flash or cause duplicate banners then we're fine. Otherwise, we'd be better to use run this on mount as soon as the user has loaded.

aptkingston (approved):
LGTM as long as that case I mentioned doesn't cause any problem!

packages/builder/src/pages/builder/_layout.svelte
Ghrehh: Yeah that's a good point, I'll just run it on mount.

Ghrehh: Actually, looking at this again, I added a flag that means this is only ran once, and I avoided using the onMount originally because the user object isn't loaded properly when the on mount is ran.

So I think this is fine the way it is 👍

PR #11724, opened 2023-09-08
User table changes

aptkingston:
## Description
This PR makes the users table readonly, in preparation for the user column type. A new button has been added above the users table, explaining this change.

- Users table renamed "App users"
- Rows cannot be added, edited or deleted
- Columns cannot be added, edited or deleted
- Create view button removed for users table

image

Addresses: 
- https://linear.app/budibase/issue/BUDI-7402/%5Busers-column-type%5D-move-users-table-to-read-only-and-add-tooltip


diogenesbrussels:
Hey @aptkingston,
Quick question about this:
How does this handle columns already added to this table?
I'm currently using it to easily get information about the current user with bindings and customize their experience.
Thank you!

aptkingston:
@diogenesbrussels existing columns will remain unchanged and still be able to be used. Going forwards we're introducing a "user" column type as the new way to assign user metadata, but yea any existing columns will remain functional.

shogunpurple (approved):
LGTM!

Issue #11719, opened 2023-09-07
[BUDI-7468] Save Row action will attempt to save the image data of the S3 Upload component


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: bb.budibase.app

**Describe the bug**
**Save Row** action will attempt to save the image base64 data contents of the **S3 Upload** component.

This can cause a `Request Entity Too Large` if a large amount of image data is unintentionally passed to the `/rows` endpoint.

**To Reproduce**

1. Add a form with a **custom** schema
2. Add an S3 File Upload component
3. Add a button with a **Save Row** action
4. Preview the app and upload an image
5. Look in the network tab and see the raw image data being saved against the `/rows` endpoint

**Expected behavior**
The S3 Upload component should be ignored from the form schema to prevent it being saved against the `/rows` endpoint and causing unintended errors such as `Request Entity Too Large`

**Screenshots**

![Screenshot 2023-09-07 at 21.17.32.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/9f7590f4-b946-4b9a-acec-bd93c7dd4d89/7660b2b0-5f99-43b9-a201-1f15a0538ed0?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC85Zjc1OTBmNC1iOTQ2LTRiOWEtYWNlYy1iZDkzYzdkZDRkODkvNzY2MGIyYjAtNWY5OS00M2I5LWEyMDEtMWYxNWEwNTM4ZWQwIiwiaWF0IjoxNjk0MTE3OTU3LCJleHAiOjE2OTQyMDQzNTd9.DXdMuQ9X54FrcnwSvmGcoihCisOKnDdtugVg7ic4fdc)

[...]

Issue #11677, opened 2023-09-06
[BUDI-7453] [BUDI-7454] S3 connection has 'null' in the title when the endpoint is left blank


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.9.36

**Describe the bug**
Small UI issue where the connection details for S3 contain 'null' when the endpoint is blank

**To Reproduce**

1. Add an S3 datasource
2. Leave *Endpoint* blank
3. See 'null' in connection details

**Expected behavior**
Maybe use a `*` to indicate null, might look a little nicer. E.g. `*:us-west-1` instead of `null:us-west-1`

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/f55d2d4c-7940-4cd6-88f0-c46ea7d6a199/9f8d8135-bd8a-46c6-a7c7-9a80ee791987?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9mNTVkMmQ0Yy03OTQwLTRjZDYtODhmMC1jNDZlYTdkNmExOTkvOWY4ZDgxMzUtYmQ4YS00NmM2LWE3YzctOWE4MGVlNzkxOTg3IiwiaWF0IjoxNjkzOTkwMzc4LCJleHAiOjE2OTQwNzY3Nzh9.qO-lVA3SbwxnEvEHKO9YXrUuSuZLJ8PslnSC4-Isu5g)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7453](https://linear.app/budibase/issue/BUDI-7453/[budi-7454]-s3-connection-has-null-in-the-title-when-the-endpoint-is)</sub>

Issue #9547, opened 2023-02-06
[BUDI-6438] S3 File Upload should not need to be in a form


melohagan:
**Describe the feature request**
Currently the S3 File Upload component must be inside a form to work. This doesn't make sense because there is no S3 data-type.

~~Furthermore, the field dropdown shows field suggestions for *Attachment* data-types, which are not supported.~~

**Screenshots**
![Screenshot 2023-02-06 at 14 39 37](https://user-images.githubusercontent.com/101575380/217000865-78cb0221-7d0e-436c-a77d-f4bb9514c8ae.png)

<sub>[BUDI-6438](https://linear.app/budibase/issue/BUDI-6438/s3-file-upload-should-not-need-to-be-in-a-form)</sub>

melohagan:
Related issue: https://github.com/Budibase/budibase/issues/8975

The current endpoint looks like this:
```js
.post(
    "/api/attachments/:datasourceId/url",
    authorized(PermissionType.TABLE, PermissionLevel.READ),
    controller.getSignedUploadURL
  )
```

It assumes the file upload to be an attachment, but in this case it should be handled differently.

melohagan:
> Furthermore, the field dropdown shows field suggestions for Attachment data-types, which are not supported.

Solved this issue, but still need to add support for S3 Upload to be a standalone non-form field component.

PR #11732, opened 2023-09-08
Save row action S3 Upload fix, and other misc S3 fixes

melohagan:
## Description
When doing a **Save Row** action, the image data url was being posted to the `/rows` endpoint which was causing a **413**.

A check has been added to ensure that the image contents of an S3 upload are not saved against the **Save Row** target table.

Other fixes include using `*` in the S3 datasource subtitle instead of the string `null`.
Also changed the S3 Upload component field to use a simple input as you can never select a valid value from the dropdown - it must always be manually entered, which should be a bit more obvious now. 

Addresses: 
- https://github.com/Budibase/budibase/issues/11719
- https://github.com/Budibase/budibase/issues/11677
- https://github.com/Budibase/budibase/issues/9547

## Screenshots
Screenshot 2023-09-09 at 00 04 26

Screenshot 2023-09-09 at 00 10 10

**Before**
https://github.com/Budibase/budibase/assets/101575380/fecae8c7-3ff8-40b1-a398-393abd6ca2aa

**After**
https://github.com/Budibase/budibase/assets/101575380/58970dcd-7adb-4601-a87e-131e55521f39


packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/EditDatasourceConfigButton.svelte
Comment (deanhannigan) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/EditDatasourceConfigButton.svelte:1]: As discussed, could the text for S3 endpoint be switched to `s3.${datasource.config.region}.amazonaws.com`?

deanhannigan (commented):
LGTM! Just the text label for the s3 endpoint and it's good to go.

packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/EditDatasourceConfigButton.svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/EditDatasourceConfigButton.svelte:1]: Screenshot 2023-09-12 at 10 24 36

Looks nicer :)

deanhannigan (approved):

PR #11738, opened 2023-09-11
cherry-picking dev fix to master

mike12345567:
## Description
This is in `develop` - adding to master as makes bug fixes a lot easier.

Updating the yarn dev process to not include a removal of the dist directory as part of the streamed watchers - this seems to create a problem for shared core and moving this up a level to a parallel run before the watchers massively improves stability.


mike12345567:
Not needed as develop being merged to master.

PR #11743, opened 2023-09-11
Postgres sql invalid time value error when saving json as text

melohagan:
## Description
Fix and unit tests for bad regex validation

Addresses: 
- https://linear.app/budibase/issue/BUDI-7476/postgres-sql-invalid-time-value-error-when-saving-json-as-text


shogunpurple (approved):
LGTM

PR #11748, opened 2023-09-12
Some small bug fixes

PClmnt:
## Description
Fixes two small issues

- Remove premium reference from view permissions
- Remove developer reference when importing users


shogunpurple (approved):
LGTM!

Issue #10118, opened 2023-03-27
[BUDI-6781] Dynamic Filtering on 0 for external sources e.g. mysql, postgres


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.4.27-alpha.5.
  * App Version: 2.4.27-alpha.5.

**Describe the bug**
You aren't able to interact with 0 with the dynamic filter within the Data or design view when using an external data source like MySQL or Postgres. When trying to do this it doesn't filter the results at all, they just remain unchanged. A side note the filter will work with 0.1 instead of 0, leading to the next point.

This does work with the Budibase datasource. The reason it doesn't work with the other sources is likely due to a falsey check where 0 is being treated as false. (There is likely a bad falsey check somewhere in the plus datasources which isn't accounting for 0)

**To Reproduce**

1. Setup a table
2. Setup up a number based column
3. Add some values
4. Add a filter for the column created for greater or less than with the value set to 0

**Expected behavior**
Filtering should occur on the data set, to show the numbers which match the filter. Nothing occurs and it remains unchanged when trying to use the filter.

**Screenshots**

Filter being used

![filter being used.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/933d6196-331e-44e7-a147-4a43941b98d0/b6592b11-c9f7-4a93-90e3-8cc101b0a664)

Budibase data source

![budibase source and what should happen.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/f6997e60-dbe1-401c-9847-d81d4fb8cabb/859f5702-1f63-41a8-a90c-6501df05082c)

External datasource - unchanged using the same filter

![external source, nothing changes.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/03d2b286-59db-4db6-821a-9edc4ec11c64/b79a0187-c81a-41f8-bdca-25fc777dc6de)

**App Export**

[d-filtering-0.tar.gz](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/afb38e85-399c-4ec3-88df-db210742d6a2/f7d40700-12d8-4ed7-9a30-b06021aec78f)

[...]

PR #11750, opened 2023-09-12
Fix filtering on 0 in external datasources

mike12345567:
## Description
`0` is a falsy value, we filter out empty filters as sometimes empty inputs such as `""` are sent up from the frontend for filtering, I've made the removal of what we consider "empty" filters more strict, looking for nullish or specifically empty string values.

Addresses: 
- #10118
- https://linear.app/budibase/issue/BUDI-6781/dynamic-filtering-on-0-for-external-sources-eg-mysql-postgres


ConorWebb96 (approved):
LGTM 🔥

PR #11751, opened 2023-09-12
Budi 7481 initial form step binding drawer can crash

melohagan:
## Description
 - Fix to ensure a string is passed into the new EditorView
 - Fix to allow the S3 Upload Field to be cleared

Addresses: 
- https://linear.app/budibase/issue/BUDI-7481/initial-form-step-binding-drawer-can-crash

### DIFF
--- a/packages/client/src/components/app/forms/S3Upload.svelte
+++ b/packages/client/src/components/app/forms/S3Upload.svelte
@@ -17,6 +17,13 @@
   let fieldApi
   let localFiles = []
 
+  $: {
+    // If the field state is reset, clear the local files
+    if (!fieldState?.value?.length) {
+      localFiles = []
+    }
+  }
+
   const { API, notificationStore, uploadStore } = getContext("sdk")
   const component = getContext("component")
 
--- a/packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
+++ b/packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
@@ -235,7 +235,7 @@
     const baseExtensions = buildBaseExtensions()
 
     editor = new EditorView({
-      doc: value,
+      doc: value?.toString(),
       extensions: buildExtensions(baseExtensions),
       parent: textarea,
     })


packages/client/src/components/app/forms/S3Upload.svelte
```diff
@@ -17,6 +17,13 @@
   let fieldApi
```
captain-hoity-toity, line 17: Overall, the code seems to be well-structured and organized. However, there are a few recommendations that could help to improve the code quality and readability.

1. **Code Commenting**: It would be beneficial to add comments for complex code blocks to improve readability and maintainability.

2. **Error Handling**: There seems to be a lack of error handling when processing files. It would be beneficial to add a `catch` block or a `finally` block to handle any errors or to ensure that certain code runs regardless of whether an error was thrown.

3. **Use of Constants**: `MaxFileSize` is defined as a constant at the top of the script, it would be a good practice to define other constants like `"image"` and `"sdk"`.

4. **Code Duplication**: The `loading = false` statement is duplicated in both the `try` and `catch` blocks of the `upload` function. To avoid this, you could place it in a `finally` block that will execute after the `try` and `catch` blocks.

5. **Function Naming**: Function names should be descriptive. The function name `processFiles` could be more descriptive. Consider renaming it to something like `processFileList` or `processUploadedFiles`.

Here's how the code might look after applying these suggestions:

```javascript
// Constants
const IMAGE_TYPE_PREFIX = "image"
const SDK_CONTEXT = "sdk"

// Error handling added in processFiles function
const processUploadedFiles = async fileList => {
    try {
      // existing code
    } catch (error) {
      // handle error
    }
}

// Removed duplicate code in upload function
const uploadFile = async () => {
    loading = true
    try {
      const res = await API.externalUpload({
        datasourceId,
        bucket,
        key,
        data,
      })
      notificationStore.actions.success("File uploaded successfully")
      return res
    } catch (error) {
      notificationStore.actions.error(
        `Error uploading file: ${error?.message || error}`
      )
    } finally {
      loading = false
    }
}

// Renamed function
<Field
  {label}
  {field}
  {disabled}
  {validation}
  type="s3upload"
  bind:fieldState
  bind:fieldApi
  defaultValue={[]}
>
  <div class="content">
    {#if fieldState}
      <CoreDropzone
        value={localFiles}
        disabled={loading || fieldState.disabled}
        error={fieldState.error}
        on:change={handleChange}
        {processUploadedFiles} // renamed function
        {handleFileTooLarge}
        maximum={1}
        fileSizeLimit={MaxFileSize}
      />
    {/if}
    {#if loading}
      <div class="overlay" />
      <div class="loading">
        <ProgressCircle />
      </div>
    {/if}
  </div>
</Field>
```

deanhannigan (approved):
LGTM!

PR #11753, opened 2023-09-12
Test DS plus row unittest

adrinr:
## Description
Testing the row API using a Postgres datasource as well as the internal couchdb. This will ensure that all the endpoints behave equally and it will avoid regression issues

Addresses: 
- [BUDI-7367
](https://linear.app/budibase/issue/BUDI-7367/full-postgres-ds-test-suite)


mike12345567 (approved):
Incredible work! From improving the consistency of loading the Postgres container, massively improving the test utilities around data, cleaning up so many parts of the table/rows tests and then solving a bunch of bugs/issues in the process - while improving the typing around all of the APIs - this is just a general face lift for an area of the product we depend on so much! Amazing work on this and LGTM 🎊

Issue #9764, opened 2023-02-21
[BUDI-6699] Automation's "Update Row" Action does not allow to save rows with "Off" / "False" value of a Boolean field


sergey-vin:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists
https://github.com/Budibase/budibase/issues?q=is%3Aissue+is%3Aopen+update+row+automation

**Hosting**
- Cloud

**Describe the bug**
In Update Row action there's only a toggle with 2 values for boolean fields:
- "Off" leaves the value unchanged.
- "On" sets it to true.
- There's no way to set it to "False".

(Note: That's a more detailed description of https://github.com/Budibase/budibase/issues/9697)

**To Reproduce**
1. Create new app, import from a template "Professional Directory Template"
2. In "Data", add a boolean field to a "Directory" table 
<img width="390" alt="image" src="https://user-images.githubusercontent.com/1080493/220314531-9cf31cc8-c2cd-45f9-813f-22b41c8656dd.png">

3. In "Data", edit "Directory" table to set/remove some values from some fields  (`isTerminated`, `City` and `Zip` of IDs=1, 2, 3)
4. Remember the state of the DB 
<img width="678" alt="image" src="https://user-images.githubusercontent.com/1080493/220314930-8775bd12-6a68-4df2-bbce-3c1b00f5e0ef.png">
5. Create an automation to query all rows and change only `Address`, see that it's not clear what will happen with "isTerminated" (will it be all marked as "false", or kept as is?)
6. Run it
<img width="515" alt="image" src="https://user-images.githubusercontent.com/1080493/220328624-140549cb-b0c9-4c6d-8d8c-7dc3a5ebe96b.png">

7. Observe the DB - address was changed, but `isterminated` - not
<img width="704" alt="image" src="https://user-images.githubusercontent.com/1080493/220328221-eb520f17-e436-413e-8903-912d00c4b0a0.png">

8. Toggle `IsTerminted` to "On" / "True", re-run automation
<img width="549" alt="image" src="https://user-images.githubusercontent.com/1080493/220328841-9447e459-ba92-446e-8f9b-52b01936e81a.png">

9. Observe that `IsTermintated` was set to True in the DB
<img width="657" alt="image" src="https://user-images.githubusercontent.com/1080493/220328995-a6672cc5-f15e-4e95-a2d9-b466803aad07.png">

**Expected behavior**
[...]

PR #11759, opened 2023-09-13
Add bindings support to meeting start / end dates

melohagan:
## Description
Realised that the calendar invite feature is much more useful with bindings.

Also removed code to complete a deprecation for the Zapier and Make automation steps.

I initially had the `Use values / use bindings` toggle but it was a bit overkill for just the two date fields. That's why I added in the `DrawerBindableSlot` which allows you to use bindings directly with the date picker (or any other field in theory).

If the date is not valid, then the input will become text, e.g. handlebars or JavaScript bindings. When a valid date is entered it will automatically switch back to the calendar.

If in "text input mode", clicking on the Cancel icon will clear the field, and switch to calendar mode. See the video below!

Decided to disable the clear button for JavaScript bindings because you can't see what you're clearing, it just says '(JavaScript binding)', and it could potentially be a lot of code that gets accidentally wiped.

Boolean fields in the row selector have also been updated to use a select with three options, and tested for a bug fix around the update row action step:
- Choose an option (will ignore the field if not set)
- True
- False

Addresses: 
- https://linear.app/budibase/issue/SUP-101/add-bindings-to-meeting-startend-dates
- https://github.com/Budibase/budibase/issues/9764

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/f43837bf-241b-4a6e-a9e0-9937436b2438

js

*Field with JavaScript binding and disabled clear button*

<br />
Screenshot 2023-09-15 at 18 05 31

*Date field with picker*

<br />
handlebars

*Field with editable handlebars text and clear button*

bool

*Boolean now has three options to allow it to be ignored*

### Removed the Use values / use bindings dropdown
Screenshot 2023-09-25 at 18 40 34

Screenshot 2023-09-25 at 18 41 47

## Documentation
- [x] https://linear.app/budibase/issue/SUP-71/add-send-meeting-invite-via-email-smtp-docs


melohagan:
Backwards compatibility test:

Screenshot 2023-09-25 at 21 54 06

Screenshot 2023-09-25 at 21 50 18

This is correct behaviour because of the linked bug. Up to now the false toggle has been considered as *ignore*. You can now set this as *False* to overwrite the existing checkbox.

packages/builder/src/components/common/bindings/DrawerBindableSlot.svelte
```diff
@@ -0,0 +1,254 @@
...
+      return !value
+    }
+    return links.every(link => schema?.constraints?.inclusion?.includes(link))
+  }
+
+  const isValidBoolean = value => {
+    return value === "false" || value === "true" || value == ""
+  }
+
+  const isValid = value => {
```
PClmnt, line 94: for this function could we do something like:

```
const validationMap = {
  date: isValidDate,
  link: hasValidLinks,
  array: hasValidOptions,
  longform: value => !isJSBinding(value),
  json: value => !isJSBinding(value),
  boolean: isValidBoolean,
};

const isValid = value => {
  const validate = validationMap[type];
  return validate ? validate(value) : true;
};


```
to make it bit more succinct?

PClmnt (approved):
Just one small suggestion but looks good!!

packages/builder/src/components/common/bindings/DrawerBindableSlot.svelte
melohagan: Nice - I like it!

Issue #11155, opened 2023-07-06
[BUDI-7258] It looks like you can add a new user via the table block, but it fails silently


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.8.1

**Describe the bug**
You can display the users from the users table in a Table block. Checking the `Show new row side panel` option allows you to add a user, and a green success toast is shown, but it obviously doesn't actually save.

**To Reproduce**
1. Add a table block for the Users table
2. Check `Show button` and select `Show new row side panel`
3. Publish and view app
4. Add a new user entry
5. See the green success toast and the user incorrectly appears to have been added

**Expected behavior**
If I try to add a row to the users table, I should see an error message indicating that I need to use user management, or the users API. 
Or the table block should have a read-only mode and enable this if it detects the users table being used.

**Screenshots**
https://github.com/Budibase/budibase/assets/101575380/76ea886e-3c4b-49c1-838e-88f35cb0cdd5

<sub>[BUDI-7258](https://linear.app/budibase/issue/BUDI-7258/it-looks-like-you-can-add-a-new-user-via-the-table-block-but-it-fails)</sub>

Also don't think this should be an option:

![Screenshot 2023-07-06 at 15 54 50](https://github.com/Budibase/budibase/assets/101575380/74b8e8ab-9e8f-4893-93c2-f5a4d8bfcc71)

PR #11769, opened 2023-09-14
Don't allow adding new users to user metadata table within apps

mike12345567:
## Description
Add an error if trying to add new users to the user metadata table within an app - this is invalid and should throw an error.

## Addresses
https://linear.app/budibase/issue/BUDI-7258/it-looks-like-you-can-add-a-new-user-via-the-table-block-but-it-fails
https://github.com/Budibase/budibase/issues/11155


packages/server/src/api/controllers/row/index.ts
```diff
@@ -72,6 +72,11 @@ export const save = async (ctx: UserCtx<Row, Row>) => {
   const tableId = utils.getTableId(ctx)
   const body = ctx.request.body
 
+  // user metadata doesn't exist yet - don't allow creation
```
adrinr, line 75: Could we add some basic tests to make sure this behaviour does not get removed in the future?

PR #11771, opened 2023-09-14
More/less than SQL fix

mike12345567:
## Description
Fix for more than/less than ranges, zeros were ignored when building up ranges, so that it simply acted like an upper limit, rather than a range.

### DIFF
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -25,6 +25,13 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
+        uses: easimon/maximize-build-space@master
+        with:
+          root-reserve-mb: 35000
+          swap-size-mb: 1024
+          remove-android: 'true'
+          remove-dotnet: 'true'
       - name: Checkout repo and submodules
         uses: actions/checkout@v3
         if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'Budibase/budibase'
--- a/packages/server/src/api/controllers/row/external.ts
+++ b/packages/server/src/api/controllers/row/external.ts
@@ -18,7 +18,6 @@ import {
 import sdk from "../../../sdk"
 import * as utils from "./utils"
 import { dataFilters } from "@budibase/shared-core"
-import { removeEmptyFilters } from "./utils"
 
 export async function handleRequest(
   operation: Operation,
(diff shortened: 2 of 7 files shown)


.github/workflows/budibase_ci.yml
```diff
@@ -25,6 +25,13 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
```
melohagan, line 28: This is mentioned in the PR - what is it for?

melohagan (approved):
The fix is good, although I wonder if there's any merit in trying to catch this earlier?

As we're using Typescript, can we not assert the data type of low / high to be a number, and then have some validation lower down the stack

.github/workflows/budibase_ci.yml
mike12345567: Unrelated to the main body of work - hoping this helps with test stability - I merged it to develop and tests were flaky on master, so added it to this PR to see if it helps here.

mike12345567: PR for `develop`: https://github.com/Budibase/budibase/pull/11760

Mentioned it in engineering, won't cause any issues for releases etc - if its a problem on `master` i'll roll it back, but trying to get it more consistent, had to run a master PR 7 times yesterday to get it to pass!

melohagan: Cool - hopefully it does help! 😄

melohagan:
> > The fix is good, although I wonder if there's any merit in trying to catch this earlier?
> > As we're using Typescript, can we not assert the data type of low / high to be a number, and then have some validation lower down the stack
> 
> Agreed it would be better to resolve this through typing, but sadly its a rather large overhaul to improve typing across all filtering, technically these are already sort of typed, but the type has to be:
> 
> ```
> high?: number
> low?: number
> ```
> 
> So there always has to be a check, but it just can't be a falsy check as that excludes `0`.
> 
> The range theoretically could just include one part of it, although it generally doesn't do that today, but we need to handle if just say the `high` was provided for pure a less than check.

It's a little odd that the type includes string:

```ts
range?: {
    [key: string]: {
      high: number | string
      low: number | string
    }
  }
```

I wonder if that was to handle date strings? Could it be changed to `number | Date` and that way you could just do a null check `value.low != null` instead of `isValidFilter(value.low)`

Just some thoughts, PR is good to go as far as I'm concerned!

mike12345567:
No string is included because you can do more than/less than strings in Lucene, testing their encoding against each other, essentially used for sorting strings! It can also be used for dates as well though, and numeric formats represented as strings (like big integers).

PR #11775, opened 2023-09-14
Add user column type to internal tables

adrinr:
## Description
Adding a new column type in internal tables for users. We will be building this as a subtype in the backend, so some refactoring was needed to handle this in the frontend.

The frontend is dummy, handling this field as a regular text field. The frontend will be handled in another PR, for this this field is only allowed for dev users

Addresses: 
- [BUDI-7455
](https://linear.app/budibase/issue/BUDI-7455/[users-column-type]-create-user-column-type)

### Adding a new column
image

### Editing an existing column
image


mike12345567 (approved):
LGTM - might get @aptkingston thoughts on the frontend side as it touches the grid.

PR #11781, opened 2023-09-15
Update dayjs dependencies

aptkingston:
This PR updates the dayjs dependencies across all packages to the same version, and fixes date cells which are currently broken due to import syntax.

As mentioned in the official documentation, dayjs should be consumed as a top level export:
```js
const dayjs = require('dayjs')
import dayjs from 'dayjs'
```

Hopefully making all versions the same will fix the problem that motivated the previous PR.


deanhannigan (approved):
LGTM!

PR #11785, opened 2023-09-15
Populate user refs on get

adrinr:
## Description
Fetch user bb refs on get, replacing the ids for the actual objects from db

Addresses: 
- [BUDI-7457
](https://linear.app/budibase/issue/BUDI-7457/[users-column-type]-retrieving-rows-with-user-columns)


packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
```diff
@@ -39,3 +39,33 @@ export async function processInputBBReferences(
...
+) {
+  if (typeof value !== "string") {
+    return value
+  }
+
+  const result = []
+
+  switch (subtype) {
+    case FieldSubtype.USER:
+      for (const id of value.split(",").filter(x => !!x)) {
```
mike12345567, line 55: Each BB reference type should be a CSV of just different IDs ideally, so we could probably move this to top level in the function, just a `const validIds = value.split(",").filter(id => !!id)`

I assume ideally as well this will use the new user cache function to fetch the users from a user ID array?

mike12345567 (approved):
Looks good - just one minor comment!

packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
adrinr: Yup, I am changing the cache usage in another branch once merged to develop

PR #11796, opened 2023-09-18
Data section frontend for the user column

adrinr:
## Description
Frontend for the data section of the user column type.
The approach of these changes were the following:
1. Reuse the link grid cell type, to avoid duplicating front end logic, but trying not to do any massive refactor to that content. For this I used overrides
2. Make use of the field subtypes of the new field type `bb_reference` on the type selector. For this I created the logic around "composite types"

### Creating a user column
image

### Editing a user cell
image
image

### Editing a user column
image

### DIFF
--- a/packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
@@ -8,6 +8,7 @@
 
   const schema = {
     ...$$props.schema,
+    // This is not really used, just adding some content to be able to render the relationship cell
     tableId: "external",
   }
 
--- a/packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
@@ -1,16 +1,36 @@
 <script>
+  import { getContext } from "svelte"
   import RelationshipCell from "./RelationshipCell.svelte"
   import { FieldSubtype } from "@budibase/types"
-  import { TableNames } from "../../../constants"
 
-  const subtypeToTable = {
-    [FieldSubtype.USER]: TableNames.USERS,
-  }
+  const { API } = getContext("grid")
+  const { subtype } = $$props.schema
 
   const schema = {
     ...$$props.schema,
-    tableId: subtypeToTable[$$props.schema.subtype],
+    tableId: "external",
+  }
+
+  async function searchFunction(searchParams) {
+    if (subtype !== FieldSubtype.USER) {
+      throw `Search for '${subtype}' not implemented`
+    }
+
+    const results = await API.searchUsers({
+      ...searchParams,
+    })
+    return {
+      ...results,
+      data: undefined,
+      rows: results.data,
+    }
   }
 </script>
 
-<RelationshipCell {...$$props} {schema} />
+<RelationshipCell
+  {...$$props}
+  {schema}
+  {searchFunction}
+  primaryDisplay={"email"}
+  relationshipType={"many-to-one"}
+/>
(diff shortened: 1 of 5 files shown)


packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
```diff
@@ -0,0 +1,35 @@
...
+  import { getContext } from "svelte"
+  import RelationshipCell from "./RelationshipCell.svelte"
+  import { FieldSubtype } from "@budibase/types"
+
+  const { API } = getContext("grid")
+  const { subtype } = $$props.schema
+
+  const schema = {
+    ...$$props.schema,
+    tableId: "external",
```
aptkingston, line 11: If we just change this to be the users table ID (or dynamically set based on subtype), then the relationship cell should correctly use this and fetch the definition of the users table and pull out the correct primary display column.

Would that not let us only require one new custom prop (`searchFunction`) rather than 2? I also think that currently the relationship cell will override the value of primary display when it uses `getPrimaryDisplayForTableId` to try and get the correct table ID for the schema, which is currently `external`, and I would expect that to throw an error.

If that works then that would be great since it would be nice to not have to manually override the primary display field.

aptkingston (commented):
LGTM! Just one potential optimisation by not needing to override the primary display field via a new prop.

packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
adrinr: The issue in here is that the relationship currently is looking for app tables, and this will be a global table. Another API is required, for this I did override the API. And the table id is actually not used in here, but before this I was getting some rendering errors when we were checking `if !tableId`

adrinr: Maybe we could modify the external component to have "global vs app data", and we use the global API or the table API. It might cover most of the cases in the future

aptkingston: I guess since we're overriding the search function, it's alright to override primary display too. If we were going to add awareness of this global users table then we would probably allow it to understand how to search that table too, so thats fine.

In that case, we can leave it like it is here 👍

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -259,14 +261,16 @@
       on:wheel={e => (focused ? e.stopPropagation() : null)}
     >
       {#each value || [] as relationship}
-        {#if relationship.primaryDisplay}
+        {#if relationship[primaryDisplay] || relationship.primaryDisplay}
```
aptkingston, line 264: Is this needed because the value for user column types is a different shape than a normal relationship, so we don't have a `primaryDisplay` key? I assume so, but just double checking.

packages/frontend-core/src/components/grid/cells/BBReferenceCell.svelte
adrinr: If we need to extend this in the future we can pass a parameter to differentiate server vs global tables, but I wanted to keep the changes to the RelationshipCell as minimal as possible

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
adrinr: Exactly. We were passing this primaryDisplay but it was never used, as the data was enriched at some point before. In our case we don't have it, so we need to make use of this field.
It could be safer inverting the operation, and always checking the `relationship.primaryDisplay` value first. But I think it's more intuitive to use the requested value in the component.

aptkingston (approved):
LGTM!

PR #11782, opened 2023-09-15
Feature - User column type

adrinr:
## Description
Adding a new column type that will allow referencing internal users. This is just the feature branch PR, a union of the already approved PRs:
- https://github.com/Budibase/budibase/pull/11775
- https://github.com/Budibase/budibase/pull/11778
- https://github.com/Budibase/budibase/pull/11783
- https://github.com/Budibase/budibase/pull/11785
- https://github.com/Budibase/budibase/pull/11796
- https://github.com/Budibase/budibase/pull/11823
- https://github.com/Budibase/budibase/pull/11828
- https://github.com/Budibase/budibase/pull/11829

More detail and screenshots can be found on the other PR descriptions

Addresses: 
- [BUDI-7403 - Users column type](https://linear.app/budibase/issue/BUDI-7403/[users-column-type]-initial-backend-ticket)

PR #11798, opened 2023-09-18
User cache bulk get

adrinr:
## Description
Adding a function to the user cache in order to retrieve users in bulk. If they are in the cache, they will be returned. If some of them are not, they will be retrieved from db in bulk.
This will be used in the user column type work (https://github.com/Budibase/budibase/pull/11782), avoiding performance issues for big requests

Addresses: 
- [BUDI-7403 - Users column type
](https://linear.app/budibase/issue/BUDI-7403/[users-column-type]-initial-backend-ticket)


packages/backend-core/src/cache/user.ts
```diff
@@ -27,6 +28,47 @@ async function populateFromDB(userId: string, tenantId: string) {
   return user
 }
 
+async function populateUsersFromDB(
+  userIds: string[],
+  tenantId: string
+): Promise<{ users: User[]; notFoundIds?: string[] }> {
+  const db = tenancy.getTenantDB(tenantId)
+  const allDocsResponse = await db.allDocs<any>({
```
mike12345567, line 36: if we could add this to the user DB functionality (`UserDB`) as a function for fetching bulk users, that would be good. There might even be one there already!

mike12345567 (approved):
Looks good - just one minor comment about checking the `UserDB`!

packages/backend-core/src/cache/user.ts
adrinr: Good shout, I missed that db utils :) This simplified my tests a lot

PR #11799, opened 2023-09-18
[Added] Creators quota update

jvcalderon:
## Description

Depends on: https://github.com/Budibase/budibase-pro/pull/222

Changes needed to deal with creators tracking.

Addresses: 
https://linear.app/budibase/issue/GROW-43/track-creators

## Screenshots

https://github.com/Budibase/budibase/assets/2486322/d66ac090-908e-4aa3-a33b-284fc8423df0

## Documentation


packages/backend-core/src/cache/writethrough.ts
```diff
@@ -4,7 +4,7 @@ import { logWarn } from "../logging"
 import { Database, Document, LockName, LockType } from "@budibase/types"
 import * as locks from "../redis/redlockImpl"
 
-const DEFAULT_WRITE_RATE_MS = 10000
+const DEFAULT_WRITE_RATE_MS = 5000
```
jvcalderon, line 7: @mike12345567 During some tests in local I've noticed a race condition related with the usage of cache. I've decreased this number and now I don't have that problems but I think it is a weak solution. What's the purpose of this usage? What problems could be derived from this change? Is this change acceptable?

in line 42 we can find this:
```javascript
const cache = await getCache()
  const key = doc._id
  let cacheItem: CacheItem | undefined
  if (key) {
    cacheItem = await cache.get(makeCacheKey(db, key))
  }
  const updateDb = !cacheItem || cacheItem.lastWrite < Date.now() - writeRateMs
```

Quotas for users are not just a sum but a complete recalculation based on persisted values; if I make a change in DB and 8 seconds later I make another one, it will take the value from cache and no update in DB will be performed. This change reduce the changes of this kind of problems but it doesn't avoid them.

Should I modify the code to use `writeRateMs = 0` instead changer the DEFAULT_WRITE_RATE_MS? What impact could have this kind of changes in performance?

packages/shared-core/src/sdk/documents/users.ts
```diff
@@ -52,6 +53,18 @@ export function hasAppBuilderPermissions(user?: User | ContextUser): boolean {
...
+export function hasAppCreatorPermissions(user?: User | ContextUser): boolean {
+  if (!user) {
+    return false
+  }
+  return _.flow(
+    _.get("roles"),
+    _.values,
+    _.find(x => ["CREATOR", "ADMIN"].includes(x)),
+    x => !!x
+  )(user)
```
jvcalderon, line 65: I've used here a functional programming approach; if it is not clear I can change it, just let me know.

packages/shared-core/src/sdk/documents/users.ts
```diff
@@ -52,6 +53,18 @@ export function hasAppBuilderPermissions(user?: User | ContextUser): boolean {
...
 }
 
+export function hasAppCreatorPermissions(user?: User | ContextUser): boolean {
+  if (!user) {
+    return false
+  }
+  return _.flow(
+    _.get("roles"),
+    _.values,
+    _.find(x => ["CREATOR", "ADMIN"].includes(x)),
```
jvcalderon, line 63: AppRole literals are defined in a separated module `frontend-core`. Should we duplicate those literals in `shared-core` too?

packages/backend-core/src/cache/writethrough.ts
```diff
@@ -119,8 +119,8 @@ export class Writethrough {
     this.writeRateMs = writeRateMs
   }
 
-  async put(doc: any) {
-    return put(this.db, doc, this.writeRateMs)
+  async put(doc: any, writeRateMs: number = this.writeRateMs) {
+    return put(this.db, doc, writeRateMs)
   }
```
jvcalderon, line 124: What could be the impact of avoid delays in this point? What is the point for that delays?

jvcalderon:
We need to apply the migration in prod: https://budibase.atlassian.net/wiki/spaces/EN/pages/9896119/Migrations

packages/shared-core/src/sdk/documents/users.ts
jvcalderon: Theoretically speaking can we create direct dependencies between modules in the same project? What's the purpose of `shared-core`? As far as I understand other modules could depend of `shared-core` but `shared-core` must not depend of other modules. So what should we do?

shogunpurple: Makes sense - in that case I think we are probably ok to just duplicate these literals rather than move things around

PR #11813, opened 2023-09-19
Sign up flow changes feedback

Ghrehh:
## Description
Budibase changes to address issues with [this ticket](https://linear.app/budibase/issue/BUDI-7302/[p1]-this-first[email-validation]-customize-sign-up-flow) as pointed out by @aptkingston (in a comment on the same ticket).

- The verification banner as specified is a bit different design wise from the bbui banner component, so instead of adding a bunch of optionality to that component I thought it best to just create a new one for this use case.

- Banners were breaking certain sections of the website, namely the app builder section and the portal as they were mounted outside of the flexbox based layout these pages use. This will remain a problem for any banner displayed using the global bbui banner, but the verification prompt banner is now displayed within these two layouts and they will shrink accordingly.

corresponding account portal pr: https://github.com/Budibase/account-portal/pull/403

### DIFF
--- a/packages/builder/src/components/common/VerificationPromptBanner.svelte
+++ b/packages/builder/src/components/common/VerificationPromptBanner.svelte
@@ -1,0 +1,102 @@
+<script>
+  import { notifications, Button, Icon, Body } from "@budibase/bbui"
+  import { admin, auth } from "stores/portal"
+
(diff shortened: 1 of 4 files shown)


aptkingston (approved):
LGTM!

PR #11823, opened 2023-09-20
Make use of bulk cache for BB references

adrinr:
## Description
Making use of the bulk cache for BB references, to avoid any performance issue


mike12345567 (approved):
Excellent - good expansion of types as well, LGTM!

PR #11824, opened 2023-09-20
Scroll to first invalid field

melohagan:
## Description
Upon form validation, invalid fields will now scroll into view.

Also added different scroll to options. I wanted to be able to select the vertical alignment of the scroll to be 'End' so that the scroll would naturally go to the top of the screen, without the need of a hidden field.

Addresses: 
- https://github.com/Budibase/budibase/issues/9939

## App Export
[Scroll and Final actions-export-1691506016459.tar.gz](https://github.com/Budibase/budibase/files/12292651/Scroll.and.Final.actions-export-1691506016459.tar.gz)

## Screenshots
**Scroll To (vertical alignment and validation)**

https://github.com/Budibase/budibase/assets/101575380/d0b5287a-53fd-47b4-9363-8eee743bc9c4

Screenshot 2023-08-08 at 15 54 35


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ScrollTo.svelte
```diff
@@ -36,6 +36,15 @@
...
   <Combobox bind:value={parameters.field} options={fieldOptions} />
+  <Label small>Vertical alignment</Label>
+  <Select
+    bind:value={parameters.block}
+    options={[
+      { label: "Start", value: "start" },
+      { label: "Center", value: "center" },
+      { label: "End", value: "end" },
+    ]}
+  />
```
aptkingston, line 47: I'd prefer to avoid adding this option if possible - I'm a strong believe that this should always be `nearest` which will scroll as little as possible to show your component. It means it accounts for needing to scroll either up or down to see the field. Picking any of these options will always overscroll for at least one direction. So I think we don't need to provide this option here, and instead we just always use `block: "nearest"` in the button action handler.

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ScrollTo.svelte
```diff
@@ -11,8 +11,8 @@
   export let parameters
 
   onMount(() => {
-    if (!parameters.type) {
-      parameters.type = "top"
+    if (!parameters.block) {
+      parameters.block = "start"
```
aptkingston, line 15: We can remove this onMount handler entirely if we opt for always using nearest.

packages/client/src/components/app/forms/InnerForm.svelte
```diff
@@ -408,16 +408,38 @@
...
+    fieldElement.focus({ preventScroll: true })
+    if (block === "start") {
+      const label = document.querySelector(`label[for="${fieldId}"]`)
+      label.scrollIntoView({ behavior: "smooth" })
+    } else {
+      fieldElement.scrollIntoView({
+        behavior: "smooth",
+        block,
+      })
+    }
```
aptkingston, line 426: As mentioned above, I think we should just use `nearest` for the `block` param here. We can ensure we always have a little bit of padding from the edge of the screen by using the `scrollMargin` propery.

e.g.
```js
fieldElement.style.scrollMargin = "100px"
fieldElement.scrollIntoView({
  behavior: "smooth",
  block: "nearest",
})
```

packages/client/src/components/app/forms/InnerForm.svelte
```diff
@@ -408,16 +408,38 @@
...
+      callback: () => ({
+        fields: fields
+          .filter(field => get(field).step === get(currentStep))
+          .map(field => ({
+            field: get(field),
+            valid: get(field).fieldApi.validate(),
+          })),
+        valid: formApi.validate(),
+      }),
+    },
```
aptkingston, line 442: I think we'd be better just moving the "scroll to the first invalid field" behaviour into the formApi.validate function itself. It would make sense for that to internally handle it. That also means we can remove this custom logic here, and we can also remove the custom return value logic in the button actions file where we manually invoke scrolling.

Let's just automatically scroll to the first invalid field before we return the validation result, inside the InnerForm validate function itself.

packages/client/src/utils/buttonActions.js
```diff
@@ -531,6 +532,19 @@ export const enrichButtonActions = (actions, context) => {
...
+              {
+                [`##eventHandlerType`]: "Scroll To Field",
+                parameters: {
+                  block: "start",
+                  componentId: action.parameters.componentId,
+                  field: result.fields?.find(field => !field.valid)?.field,
+                },
+              },
+              context
+            )
```
aptkingston, line 547: As mentioned above, we can remove this if we just handle scrolling to the first invalid field as part of the formAPI validate function.

aptkingston (commented):
This is a very useful feature so I think we should definitely get it in. I've added a few comments about how this could be handled in a different way which simplifies things. Nice work so far!

packages/client/src/components/app/forms/InnerForm.svelte
melohagan: The scrollMargin property is perfect! It makes sure the title of the form is also shown if the top most field is invalid.

packages/client/src/components/app/forms/InnerForm.svelte
```diff
@@ -408,37 +414,23 @@
...
-      label.scrollIntoView({ behavior: "smooth" })
-    } else {
-      fieldElement.scrollIntoView({
-        behavior: "smooth",
-        block,
-      })
-    }
+    const label = document.querySelector(`label[for="${fieldId}"]`)
+    label.style.scrollMargin = "100px"
+    label.scrollIntoView({ behavior: "smooth", block: "nearest" })
```
aptkingston, line 426: Could we use the field itself rather than the label? There's a chance that there won't be a label, and that will be more likely with the recent (and more upcoming) form block changes.

packages/client/src/utils/buttonActions.js
```diff
@@ -196,6 +196,7 @@ const scrollHandler = async (action, context) => {
     ActionTypes.ScrollTo,
     {
       field: action.parameters.field,
+      block: action.parameters.block,
```
aptkingston, line 199: We can remove this since it's no longer a parameter.

packages/client/src/components/app/forms/InnerForm.svelte
```diff
@@ -409,15 +415,23 @@
...
   }
 
   // Action context to pass to children
   const actions = [
-    { type: ActionTypes.ValidateForm, callback: formApi.validate },
+    {
+      type: ActionTypes.ValidateForm,
+      callback: formApi.validate,
+    },
```
aptkingston, line 434: NAB, but we could revert this to being on one line like the other actions (which is what it was originally).

aptkingston (approved):
2 very minor changes suggested, but all logic looks good to me! Will approve now, but it would be good to quickly do those 2 small changes while fixing the lint issue anyway 👍

packages/client/src/components/app/forms/InnerForm.svelte
melohagan: I tested the scenario for when the label is left blank. The label element still exists and the scroll is successful.

I had to use the label because for some reason it wouldn't scroll to a date picker.

PR #11827, opened 2023-09-20
Small user openAPI update

mike12345567:
## Description
Removing required from roles on user API, as it is now paid.


shogunpurple (approved):
LGTM!

PR #11828, opened 2023-09-20
Fix bb references deletions

adrinr:
## Description
Properly handle deleting bb reference fields, ensuring we save undefined values when empty and undefined values when fetching

### DIFF
--- a/packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
+++ b/packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
@@ -6,20 +6,25 @@ import { InvalidBBRefError } from "./errors"
 export async function processInputBBReferences(
   value: string | string[] | { _id: string } | { _id: string }[],
   subtype: FieldSubtype
-): Promise<string> {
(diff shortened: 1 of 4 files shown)


packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
```diff
@@ -6,20 +6,25 @@ import { InvalidBBRefError } from "./errors"
 export async function processInputBBReferences(
   value: string | string[] | { _id: string } | { _id: string }[],
   subtype: FieldSubtype
-): Promise<string> {
-  const result: string[] = []
+): Promise<string | undefined> {
+  const referenceIds: string[] = []
+
+  if (Array.isArray(value)) {
+    referenceIds.push(...value.map(x => (typeof x === "string" ? x : x._id)))
```
mike12345567, line 13: can we use `id` or `possibleObj` rather than `x`? When looking at a lot of different array functions I do mind it a lot more readable to see something contextual.

mike12345567 (approved):
LGTM!

PR #11829, opened 2023-09-20
Display design BB refs

adrinr:
## Description
Implementing design changes in order to display the new BB reference types.
For the display on the tables, we will use the RelationshipRenderer as the data is designed to come in the same shape.
For the the form elements, we tweaked the Relationship one to accept different datasource types, using the already created `dataFetch`

### Tested scenarios
_All both with a multiuser and a single user field_

- [x] Adding a new row without users
- [x] Adding a new  row with users
- [x] Editing the users of a row without users
- [x] Editing the users of a row that contains users
- [x] Removing the users of a row
- [x] Deleting a row with users
- [x] Deleting a row without users

### Table side form configuration
image

### Adding fields on forms
image

### Form block
image

### Display with data
image

### Display without data
image

### Editing without data
image
image

### Editing with data
image


PClmnt (approved):

aptkingston (approved):
Nice one @adrinr 👌

PR #11830, opened 2023-09-20
Global bindings

aptkingston:
## Description
Opening a PR just so that we can spin this up in an environment and experiment with this feature more. The branch should currently be fully functional and backwards compatible.

All bindings and actions provided by components are available anywhere on the page, regardless of nesting. Repeaters are the only component which still only provides "local" scope for its context.

## Feature branch env
[Feature Branch Link](http://fb-global-bindings.fb.qa.budibase.net)


deanhannigan (approved):
LGTM

PR #11851, opened 2023-09-22
Remove form field default value settings from being updated by conditions

aptkingston:
## Description
The "default value" setting for form fields will no longer appear as an option when updating settings via conditions. This has been changed for all field types.

Setting is now hidden:
image

Also updated the conditions button to be in line with all our other buttons in settings:
image
image

Addresses: 
- https://linear.app/budibase/issue/BUDI-7464/remove-the-option-to-conditionally-update-default-value


melohagan (approved):
LGTM!

Issue #11853, opened 2023-09-22
[BUDI-7530] Relationship picker auto focuses search text


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.10.10

**Describe the bug**
When a relationship picker has search enabled, the search will be auto-focused, which doesn't make sense in the context of a form

**To Reproduce**
1. Add a form with relationship picker that has search enabled
2. Preview the app - see that the search field auto-focused

**Expected behavior**
The search text field of a relationship picker should not auto-focus.

<sub>[BUDI-7530](https://linear.app/budibase/issue/BUDI-7530/relationship-picker-auto-focuses-search-text)</sub>

melohagan:
Will be resolved when the new relationship picker is released.

PR #11854, opened 2023-09-22
Fix relationship picker auto focusing search text

melohagan:
## Description
Removed autofocus on relationship picker search field. 

Also added in a performance fix to ensure a search isn't called when the number of relationships is within the 100 limit.

Addresses: 
- https://github.com/Budibase/budibase/issues/11853


melohagan:
Closing in favour of: https://github.com/Budibase/budibase/pull/11856

PR #11856, opened 2023-09-22
Fix display of field with large relationships

adrinr:
## Description
Fixing several issues and refactor the relationship fields.
1. Remove direct API references from the `RelationshipField` to rely only on the fetchData. This `fetchData` might have different implementations depending on the datasource type, so using `fetchData` will encapsulate this logic
2. Fixing `Picker` and `Select` autocomplete components. The options passed to the object will be growing when new data is retrieved, the filtering will be managed by the picker itself. Doing this, the display values will always be respected

| Before |
|--------|

https://github.com/Budibase/budibase/assets/15987277/0860bc3e-b2d7-41b8-aee8-dd9eaf0ddacf

| After |
|--------|

https://github.com/Budibase/budibase/assets/15987277/75280d86-bd50-47d9-b1bf-46807ac1ea4b

3. Use autocomplete on the Relationship field

| Before | After |
|--------|--------|
| image | image | 

4. Add initial values on the dropdowns, so they can be deselected. In the screenshot above, the item `C25` is not in the first 100 returned items

5. Add infinite scroll loading when relationship has more than 100 elements. Until now it was impossible to select further elements, and there was not messaging around that, so the user would not know that some options were missing

https://github.com/Budibase/budibase/assets/15987277/58370a27-0002-4a24-ac1d-a4faccd1a8db

The refactor of these APIs will fix the issues around https://github.com/Budibase/budibase/pull/11829


packages/builder/src/pages/builder/portal/account/auditLogs/index.svelte
```diff
@@ -86,8 +86,15 @@
   $: userPage = $userPageInfo.page
   $: logsPage = $logsPageInfo.page
 
+  $: userList = {
+    ...(userList || {}),
+    ...$users.data?.reduce((p, c) => {
```
melohagan, line 91: What are letters 'p' and 'c' in this case? Maybe give them full variable names?

packages/builder/src/pages/builder/portal/account/auditLogs/index.svelte
```diff
@@ -86,8 +86,15 @@
...
+      p[c._id] = c
+      return p
+    }, {}),
+  }
   $: sortedUsers = sort(
-    enrich($users.data || [], selectedUsers, "_id"),
+    enrich(Object.values(userList), selectedUsers, "_id"),
```
melohagan, line 97: Potential null pointer - can we be certain userList will always be an array here? Could do `userList || []` to be safe?

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,55 +36,47 @@
...
-            [`1:${primaryDisplay}`]: fieldState?.value?.map(
-              value => value.primaryDisplay
-            ),
-          },
-        },
-      }).then(response => {
-        const value = multiselect ? multiValue : singleValue
-        selectedOptions = response.rows.filter(row => value.includes(row._id))
-      })
+  $: options = Object.values(allOptions)
```
melohagan, line 64: Just confirming that **allOptions** will always be an array here - null pointer risk.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,55 +36,47 @@
...
+    initialValues ||
+    (primaryDisplay &&
+      fieldState?.value?.map(x => ({
+        _id: x._id,
+        [primaryDisplay]: x.primaryDisplay,
+      })))
+
+  $: allOptions = {
+    ...(allOptions || {}),
+    ...[...($fetch.rows || []), ...(initialValues || [])]?.reduce((p, c) => {
```
melohagan, line 58: So many spread operators! Is there a way to make this a bit more readable? Are all these spreads necessary?

Same comment as before about what are 'p' and 'c' ?

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,55 +36,47 @@
...
-        selectedOptions = response.rows.filter(row => value.includes(row._id))
-      })
+  $: options = Object.values(allOptions)
+
+  $: fetchRows(searchTerm, primaryDisplay)
+
+  const fetchRows = (searchTerm, primaryDisplay) => {
+    const allRowsFetched =
+      $fetch.loaded &&
+      !Object.keys($fetch.query.string).length &&
```
melohagan, line 71: Double checking null pointer risk! Will this always be present: `$fetch.query.string` ?

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,55 +36,47 @@
...
+      !$fetch.hasNextPage
+    if (!allRowsFetched) {
+      // Don't request until we have the primary display
+      if (primaryDisplay) {
```
melohagan, line 75: Could be combined into a single if statement rather than nested:

`if (!allRowsFetched && primaryDisplay)`

melohagan (dismissed):
Code is grand, just had a few small comments.

Pulled down the branch and played around with it. Overall looks and feels great to use! 😄 

I do have a couple of observations though.

#### 1. Expanded list

Whenever you expand the list with the load more infinite scrolling, it never resets. So if you close the relationship picker and re-open it, the list is very long.
Also when unchecking multi-select options, the infinite scroll makes this awkward. Previously the options that were outside of the 100 limit were appended to the final list so that you had easy access to both see and uncheck those options. 

Previously:
Screenshot 2023-09-22 at 17 32 31

Now:
https://github.com/Budibase/budibase/assets/101575380/54cbd1e0-fb48-472a-aa2b-09849b300b3d

List stays expanded after being closed and re-opened:
expanded list stays expanded

#### 2. NAB: Search spinner

This is more personal preference, but I preferred the spinner being on the right of the field as its position stayed static, whereas with the spinner at the bottom you get somewhat of a 'bouncing' effect when you are searching:

https://github.com/Budibase/budibase/assets/101575380/be1535cf-2dc1-46be-b889-7c97d37b347d

packages/builder/src/pages/builder/portal/account/auditLogs/index.svelte
adrinr: `previous` and `current`. It's the way the js reduce works

packages/builder/src/pages/builder/portal/account/auditLogs/index.svelte
adrinr: This will never be null, as we are setting it up on the line above with default values already

packages/builder/src/pages/builder/portal/account/auditLogs/index.svelte
adrinr: I renamed it with the full variable names

packages/client/src/components/app/forms/RelationshipField.svelte
adrinr: This should never be null as the default value are empty objects

adrinr:
> Code is grand, just had a few small comments.
> 
> Pulled down the branch and played around with it. Overall looks and feels great to use! 😄
> 
> I do have a couple of observations though.
> #### 1. Expanded list
> 
> Whenever you expand the list with the load more infinite scrolling, it never resets. So if you close the relationship picker and re-open it, the list is very long. Also when unchecking multi-select options, the infinite scroll makes this awkward. Previously the options that were outside of the 100 limit were appended to the final list so that you had easy access to both see and uncheck those options.
> 
> Previously: Screenshot 2023-09-22 at 17 32 31
> 
> Now: https://github.com/Budibase/budibase/assets/101575380/54cbd1e0-fb48-472a-aa2b-09849b300b3d
> 
> List stays expanded after being closed and re-opened: expanded list stays expanded
> #### 2. NAB: Search spinner
> 
> This is more personal preference, but I preferred the spinner being on the right of the field as its position stayed static, whereas with the spinner at the bottom you get somewhat of a 'bouncing' effect when you are searching:
> search.spinner.mov

Good point on these comments. I will have a look at it, even if I am not sure about the resetting the list each time make sense, as we will then be doing unnecessary API calls

melohagan:
> Good point on these comments. I will have a look at it, even if I am not sure about the resetting the list each time make sense, as we will then be doing unnecessary API calls

Thanks! Maybe there's a possibility of caching the results for when you re-open the picker.

aptkingston:
@adrinr I wouldn't worry about resetting the list - like you said, the priority should be avoiding wasted API calls. Personally I think the loading spinner is also fine where it is, but that's probably just personal preference.

I think the one quality of life change that might be useful though would be to sort the options so that all selected options are listed first - that means that even if you have an extremely long list, you'll always be able to see and remove the selected options quickly. Would that be easy enough to implement?

I would also cap the dropdown with a max height so it doesn't span the entire page.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,54 +36,51 @@
...
-          oneOf: {
-            [`1:${primaryDisplay}`]: fieldState?.value?.map(
-              value => value.primaryDisplay
-            ),
-          },
-        },
-      }).then(response => {
-        const value = multiselect ? multiValue : singleValue
-        selectedOptions = response.rows.filter(row => value.includes(row._id))
+  $: fetchRows(searchTerm, primaryDisplay)
```
aptkingston, line 73: Thanks for breaking some of these giant reactive statements out into functions - it's much clearer to read and also gives us much more explicit control over when svelte reactivity will invoke them.

melohagan:
> I think the one quality of life change that might be useful though would be to sort the options so that all selected options are listed first - that means that even if you have an extremely long list, you'll always be able to see and remove the selected options quickly. Would that be easy enough to implement?
> 
> I would also cap the dropdown with a max height so it doesn't span the entire page.

100% on this, that's the main thing for me. Everything else was perfect.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -50,54 +36,51 @@
...
+            _id: value._id,
+            [primaryDisplay]: value.primaryDisplay,
+          }
+          return accumulator
+        }, {}),
+      }
     }
   }
+  $: optionsObj = {
+    ...optionsObj,
```
aptkingston, line 66: Does this work alright... it's basically a recursive declaration but I guess svelte can work out how to handle it?

It might be a bit clearer to refactor this to have an `optionsObj` as you have, then something like
```js
$: enrichedOptions = enrichOptions(optionsObj, $fetch.rows)
```
and break the logic out into a function.

I just think the recursive declaration is a little confusing, especially in future whenever the next dev comes to work on this file.

Or maybe just renaming them would be enough so that the reactive part is actually a new variable with a different name, rather than reactively updating a variable with the same name.

aptkingston (commented):
Super work @adrinr! Great to get some consistency in terms of search, and it functions better as well. The main thing I think it would be good to change is my comment above about sorting the options to list the selected ones first. The only other comment here is just a suggestion to simplify the code a little, but it's not super important.

aptkingston (approved):
LGTM!

Issue #11867, opened 2023-09-23
Allow plugins to contribute DatasourcePlus integrations


Samuel-Martineau:
**Describe the feature request**
Plugins can't currently contribute `DatasourcePlus` integrations, which prevents them from having feature-parity with native data sources. _(I ran into this issue whilst trying to fully integrate [Xata.io](xata.io) into Budibase, https://github.com/Budibase/budibase/discussions/11863#discussion-5660411)_

Fixing this should only require a change to the schema validation for plugins, as they seem to behave the same as native integrations after that step.

Samuel-Martineau:
I have prepared a pull request with what I believe to be the minimal changes required to bring feature-parity to plugins.

Samuel-Martineau:
@mike12345567 I understand your reasons for closing the above pull request. However, even through the lack of documentation, it is quite possible to build a DatasourcePlus plugin based on the Google Sheet and Postgres examples. I am almost done building a FOSS Xata.io integration for my organization and it's quite saddening to know that we will need to run our own fork for now, just because of some schema validation rule, with all the functionality behind working.

Do you think it would be possible to unofficially / experimentally support this? Perhaps throw a warning when a "plus" plugin is detected, whilst still allowing it to function.

mike12345567:
Hi @Samuel-Martineau - I would be interested to see the approach used to build a Xata.io integration - if it is infact possible to do this then we might consider opening up the possibility but yes we would need to add some sort of logic to warn the user that not everything will work.

Again, I should warn you that building DS+ plugins might seem possible initially, but the level to which they integrate makes them very hard to make and keep them stable - it is possible but quite a bit harder to achieve right now. There isn't any guarantee that the plugin will work with future versions of Budibase as the structure and approach used for DS+ could change, which might require a re-write of the plugin.

mike12345567:
That would be fantastic, just to get an idea of the approach you've used, would help to support the possibility of exposing it!

Samuel-Martineau:
@mike12345567 Sorry for the delay, the last few days have been quite busy. Here is a repo containing the code for my Xata.io DatasourcePlus integration. https://github.com/Samuel-Martineau/budibase-datasource-xata

Do you think that we could consider reopening #11868 to make this possible without having to fork the project?

mike12345567:
Hi @Samuel-Martineau - this looks very interesting, thank you for showing us how you've approached this!

[...]

PR #11868, opened 2023-09-23
Allow plugins to contribute DatasourcePlus integrations

Samuel-Martineau:
## Description

Plugins can't currently contribute DatasourcePlus integrations, which prevents them from having feature-parity with native data sources.

Addresses:

- https://github.com/Budibase/budibase/issues/11867#issue-1909956187

## Documentation


Samuel-Martineau:

mike12345567:
Hi @Samuel-Martineau - the reason this wasn't supposed initially was due to the fact that datasource plus plugins would be extremely difficult to build, we haven't really written any documentation around this and exposing all of the functionality/types needed would take quite a bit of time.

While this would make it possible to build a datasource with a plus schema, the plugin itself would need to support all of the schema building, dynamic query functionality, filtering, pagination, sorting etc that all datasource plus are expected to have.

Just as an example, here is the `Postgres` Datasource plus code: https://github.com/Budibase/budibase/blob/develop/packages/server/src/integrations/postgres.ts

If you look at this, you'll see there is a function `query` - as well as the normal CRUD functions, this is what must be implemented to support Datasource plus capabilities - today this `query` function maps to an SQL sublayer: https://github.com/Budibase/budibase/blob/develop/packages/server/src/integrations/base/sql.ts

Our plan was at some point to expose the ability to flesh out full datasource plus plugins in the future, but first we want to turn a few of our other datasources into plus versions to make sure that is easy enough - as well as handling the ability to remove features that some databases may not support (for example turning DynamoDB into a full DS+ would be quite difficult without disabling some features).

I'm going to close this for now as while this allows people to attempt to implement DS+ plugins, I'm not sure it would realistically be possible to have any success with it (due to lack of documentation etc, typing etc).

Issue #11844, opened 2023-09-21
[BUDI-7528] Linked rows, longform text areas, and multi-select arrays do not save when set within an automation


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.10.10

**Describe the bug**
Setting the values of a linked row picker (multi-select or single option) does not save against the automation.

The same bug can be reproduced with:
- longform TextArea
- array Multi-select

**To Reproduce**
1. Create an app with the sample data
2. Add an automation with any trigger, and add a **Create Row** step
3. Select the Employees table
4. Select some jobs
5. Refresh the page - the jobs are not selected, they were not saved against the automation
6. Select some jobs, and change a text field
7. Refresh the page - see that both the jobs and the text field value have now been saved

**Expected behavior**
Linked rows, longform fields, and multi-selects should save against the automation when changed.

**Screenshots**
![automations](https://github.com/Budibase/budibase/assets/101575380/f2e7fbf0-b1fa-42d1-a388-3a4033984713)

**Additional context**
There appears to be an issue with `bind:value`. The components that are using `on:change` appear to be fine.

<sub>[BUDI-7528](https://linear.app/budibase/issue/BUDI-7528/linked-rows-longform-text-areas-and-multi-select-arrays-do-not-save)</sub>

PR #11874, opened 2023-09-25
Add onChange to linked rows, longform text area, and multi select arrays in automations

melohagan:
## Description
Fix for some of the row selector types not saving against the automation on change

Addresses: 
- https://github.com/Budibase/budibase/issues/11844


PClmnt (approved):
LGTM

Issue #11774, opened 2023-09-14
[BUDI-7490] Delete button does not show for table block until after you focus the input


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method:  docker compose
  - Budibase Version: 2.10.6

**Describe the bug**
When you add an auto-generated screen for list view, no delete button is added even though it is shown in the settings. If you click on the 'Delete' input text, then the delete button will be added.

**To Reproduce**
1. Add sample data
2. Create list view screen for the 'Employees' table
3. Preview app - click on a row to see there is no delete button
4. Click on the 'Delete' text in the settings panel of the table block
5. Preview app - click on a row to see the delete button has now been added

**Expected behavior**
When adding an auto-generated list view screen, the delete button should be included in the side panel, as reflected in the settings panel.

**Screenshots**
![Screenshot 2023-09-14 at 15 38 02](https://github.com/Budibase/budibase/assets/101575380/a08ad049-4bf8-419b-b2bd-b15f79c68c87)

![Screenshot 2023-09-14 at 15 38 21](https://github.com/Budibase/budibase/assets/101575380/ce1c2b0e-bd8b-4039-8ddb-3fd1243b815b)
*No delete button*

![Screenshot 2023-09-14 at 15 38 41](https://github.com/Budibase/budibase/assets/101575380/e7e9ab1c-60f9-4b85-a0de-12e9789d5bfd)
*Focus the delete text input*

![Screenshot 2023-09-14 at 15 39 08](https://github.com/Budibase/budibase/assets/101575380/13ba5df5-ff2b-432c-8273-8b5ab5bea7f9)
*Delete button now appears*

<sub>[BUDI-7490](https://linear.app/budibase/issue/BUDI-7490/delete-button-does-not-show-for-table-block-until-after-you-focus-the)</sub>

deanhannigan:
PR [https://github.com/Budibase/budibase/pull/11875](https://github.com/Budibase/budibase/pull/11875)

PR #11875, opened 2023-09-25
Fix for ensure Table Block Delete button displays when creating an auto generated screen.

deanhannigan:
## Description

Fixes an issue where config required to display the `Delete` button in the table block is missing initially. 
- Suitable label defaults are now added when generating auto screens
- In the event that these are missing, the delete button will remain visible until the `Delete button` field is explicitly cleared 

Addresses: 
- https://github.com/Budibase/budibase/issues/11774

### DIFF
--- a/packages/client/src/components/app/blocks/TableBlock.svelte
+++ b/packages/client/src/components/app/blocks/TableBlock.svelte
@@ -249,7 +249,7 @@
             props={{
               dataSource,
               saveButtonLabel: sidePanelSaveLabel || "Save", //always show
-              deleteButtonLabel: deleteLabel, //respect config
+              deleteButtonLabel: deleteLabel == "" ? "" : "Delete",
               actionType: "Update",
               rowId: `{{ ${safe("state")}.${safe(stateKey)} }}`,
               fields: sidePanelFields || normalFields,
--- a/packages/client/src/components/app/blocks/TableBlock.svelte
+++ b/packages/client/src/components/app/blocks/TableBlock.svelte
@@ -45,8 +45,21 @@
   let enrichedSearchColumns
   let schemaLoaded = false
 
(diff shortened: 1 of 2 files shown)


packages/client/src/components/app/blocks/TableBlock.svelte
```diff
@@ -249,7 +249,7 @@
             props={{
               dataSource,
               saveButtonLabel: sidePanelSaveLabel || "Save", //always show
-              deleteButtonLabel: deleteLabel, //respect config
+              deleteButtonLabel: deleteLabel == "" ? "" : "Delete",
```
Ghrehh, line 252: Does the non-strict equality get us anything an impilict truthiness check wouldn't here? Just checking as `==` always confuses me 🙃

Ghrehh (approved):
LGTM but small question

packages/client/src/components/app/blocks/TableBlock.svelte
deanhannigan: Yep, there are two things in play here. There's a backwards compatibility check at the top that looks for the old toggle boolean and ensures that it is taken into account first.

If the legacy value was explicitly `false` we change the label value to `""`. This implies then, on line 252, that a user decision to hide the button was made and so an empty string will now be used for the `Delete button` label, hiding it.

If no value was ever set for the `Delete button` label in the screen config, this value is explicitly `undefined` and considered to have never been set. In that scenario the default behaviour is to show the button with the `Delete` label.

deanhannigan: I had a look over this area and decided to refactor all the delete label code into its own function. It was difficult to follow and the line you highlighted is actually a mistake on my part. Cheers 👍

aptkingston (approved):

PR #11876, opened 2023-09-25
App import to existing app, and public API endpoints

mike12345567:
## Description
This PR solves the following issues:
- https://linear.app/budibase/issue/BUDI-7400/[existing-app-update]-endpoint-to-override-an-existing-app
- https://linear.app/budibase/issue/BUDI-7458/[existing-app-update]-frontend-for-app-update

Also has a pro component, PR: https://github.com/Budibase/budibase-pro/pull/224

This introduces the capability of importing an app to an existing app. This essentially overwrites all of the "app" components, without removing any of the user permissions/internal data from the system. All of the following will be updated:
- datasources
- queries
- table schemas
- views
- screens
- automations/webhooks

Meaning that everything that is "designed" within the builder will be replaced by what is coming from the export. It is important to note that the app export does not need to be from the same app, technically you can import from a totally different apps export, although if published this can cause some parts of the data to become inaccessible (tables/columns don't exist etc).

There isn't much else to this feature, it works about as you would expect - the frontend is quite simple, just an extension of the current "Import" section of the app settings:

image

and the modal:
image

There is also a public API component to this, which requires a business/enterprise license that can be used for automation. It is now possible to export/import apps through the public API - this also includes the OpenAPI documentation of such APIs.


PClmnt (approved):
UI LGTM

packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -151,7 +151,8 @@ export function getListOfAppsInMulti(tmpPath: string) {
 export async function importApp(
   appId: string,
   db: Database,
-  template: TemplateType
+  template: TemplateType,
+  opts: { objStore: boolean } = { objStore: true }
```
adrinr, line 155: What is this `objStore`?

packages/types/src/documents/document.ts
```diff
@@ -39,6 +39,25 @@ export enum DocumentType {
   AUDIT_LOG = "al",
 }
 
+// these are the core documents that make up the data, design
+// and automation sections of an app. This excludes any internal
+// rows as we shouldn't import data.
+export const DocumentsToImport: DocumentType[] = [
```
adrinr, line 45: These are the "allowed to import doc types", right?

packages/types/src/sdk/licensing/feature.ts
```diff
@@ -11,7 +11,7 @@ export enum Feature {
   SYNC_AUTOMATIONS = "syncAutomations",
   APP_BUILDERS = "appBuilders",
   OFFLINE = "offline",
-  USER_ROLE_PUBLIC_API = "userRolePublicApi",
+  EXPANDED_PUBLIC_API = "expandedPublicApi",
```
adrinr, line 14: Is this just a rename? Is expanded public API the business name?

packages/server/src/sdk/app/applications/import.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+    // get a temporary version of the import
+    // don't need obj store, the existing app already has everything we need
+    await backups.importApp(devId, tempDb, template, { objStore: false })
+    // get the documents to copy
+    const documents = await getImportableDocuments(tempDb)
+    // clear out the old documents
+    await removeImportableDocuments(appDb)
+    // now write the import documents
+    await appDb.bulkDocs(documents)
```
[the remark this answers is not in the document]
mike12345567: I could actually delete and write the documents in a single call - I will do that instead as that would remove any chance of deletion then not importing.

packages/server/src/api/controllers/public/applications.ts
```diff
@@ -91,12 +95,22 @@ export async function publish(ctx: UserCtx, next: Next) {
   })
 }
 
+// get licensed endpoints from pro
+export const importToApp = proSdk.publicApi.applications.buildImportFn(
```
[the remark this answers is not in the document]
mike12345567: Responded via DMs.

packages/server/src/sdk/app/backups/imports.ts
mike12345567: Its the object store, basically means whether it should try to import anything to the object store or not, theres a similar parameter on the export - I can rename them.

packages/types/src/documents/document.ts
mike12345567: Yeap, well less allowed and more desired - the ones we want, the rest we have no desire to import.

PR #11877, opened 2023-09-25
Relationship configuration for internal tables

PClmnt:
## Description
Adding a new component that allows a user to more intuitively define their relationships

image

Creating a new column - the current table you are creating the relationship from cannot be changed, so the picker is disabled for it. 

image

Trying to edit an existing relationship - all fields disabled as was the case before 

Addresses: 
#BUDI-7399


mike12345567 (approved):
LGTM - just offered an option for reducing the duplication of the mapping between types and the relationship parts.

PR #11880, opened 2023-09-25
workflow to trigger on PR close to delete feature branch env

shogunpurple:
## Description
Trigger a workflow in budibase deploys when a featurebranch is closed - we can use this to clean up the featurebranch environment when it's done with.


calexiou (approved):
lgtm

PR #11881, opened 2023-09-25
Query UI improvements

Ghrehh:
## Description
Apologies this is a bit late. I'll put some better pictures up for this later. It's mostly good to go with some small known additions that are still needed, these are mentioned down below. 

## Issues
- Access level can't be set on a new query, only while updating one. I'm like 95% sure this is an existing bug/issue, as the code to update the access level requires a resource id which as far as I can tell shouldn't exist until the query is actually created, so I don't think this is a regression. Also, this value is totally detached from the query overall, and will updated upon the dropdown being selected, again this is already now it works and not a regression, but is possibly something we should address.

## Some extra additions
- An asterisk "file changed" indicator next to the query name à la a text editor/word processor. This check is debounced so that it doesn't trigger every time a value is updated. The check before the data is submitted to the server is not debounced however.
- Upon scrolling the main body of the query editor, a border will appear on the bottom of the sticky header to better differentiate if from the background, we did this on the new design section so I think it made sense here.
- The side bar animates it's scrolling out, if folks think it looks bad I can remove it.
- ~~All inputs are disabled while the query is validating or saving, this is the reason for all those disabled props being added all over the place.~~
- The "You have unsaved changes" prompt works properly now, it didn't seem to work before from my testing.
- The auto-populating of your queries schema follows a nicer pattern now in my opinion, previously if you remove an auto-generated key if would keep trying to put it back in every time you tested the query. Now it will only auto populate the schema when the schema is completely empty, which in most use-cases should only be when a user is creating a brand new query.

## To fix

Some things that still need ironed out, but they're small enough that hopefully if this squeezes in under the wire I can merge them directly to master: (edit: should all be sorted now)

- ~~Need to rename and remove some old components.~~
- ~~Some of the icons are incorrect~~
- ~~The copy per datasource query type (sql, json etc) hasn't been entered correctly yet.~~
- ~~Some weirdness with the schema tab not updating correctly.~~
- ~~Side panel width is wrong~~
- ~~Sidebar overflow issue~~

Addresses: 
https://linear.app/budibase/issue/BUDI-6939/query-ui-improvements

Screenshot 2023-09-25 at 19 15 49
Screenshot 2023-09-25 at 19 16 06
Screenshot 2023-09-25 at 19 15 56


aptkingston (changes requested):
Nice work Gerard! Went through this with Andy and just have a few changes that we think would be useful:

- Could we remove loading states or do they serve a purpose? It's quite jarring and not something we do anywhere else, and I can't see why we would need to disable them for a very short duration while saving or running a query
- After running for the first time you can't save the query, it still stays disabled. Running it again then enables saving
- Add a tooltip to say you need to run your query before saving when the save button is disabled. You could just wrap the button in the `AbsTooltip` component to do this.
- Add min widths to the table columns in the preview table because otherwise it's unreadable. Even something small would be useful.
- Access should default to basic when creating a new query. Currently its undefined

deanhannigan:
> ## Issues
> * Access level can't be set on a new query, only while updating one. 

Hey Gerard, I had a look into this and it does appear to be an existing issue on prod. New queries do default to `Basic`, no matter what they are set to.

And for visibility, we already discussed the binding drawer overlap:

aptkingston:
I have no idea what you do to your branches but it's never possible to just pull down changes to re-review them or else it conflicts massively haha. Always need to delete the local branch and check it out again. That's not a complaint, I'm just curious what your local git workflow is that causes it because I never see it anywhere else.

aptkingston (approved):
LGTM!

Ghrehh:
@aptkingston I rebase and force push instead of merging base branch changes, but afaik a git pull should still just pull the branch as long as you have no local changes. If it's causing serious issues I can start merging instead.

deanhannigan (approved):
LGTM!

aptkingston:
> @aptkingston I rebase and force push instead of merging base branch changes, but afaik a git pull should still just pull the branch as long as you have no local changes. If it's causing serious issues I can start merging instead.

@Ghrehh it conflicts with even just git pull unfortunately, so it does make reviewing slightly harder. Up to you though, whatever works for you - don't need to change it if that's what you're used to. But yea even with zero local changes it causes a bunch of conflicts and forces deleting and checking out the branch again.

PR #11882, opened 2023-09-25
Expose date-only setting on date fields

aptkingston:
## Description
Small PR to expose a toggle on date fields which allows them to have the `dateOnly` flag. We already had this concept, but just didn't expose it to users.

Setting this flag will hide the "ignore timezones" option, as timezones should not be relevant to dates without times.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7178/frontend-expose-date-only-setting-for-date-fields

## Screenshots
image


melohagan (approved):
Awesome!

PR #11889, opened 2023-09-26
Fix for to ensure selectedBy Avatar is not obscured in NavItem

deanhannigan:
## Description

If a source name was too long it would push the selectedBy Avatar out of view.
- Source names should now truncate with an ellipsis if too long for the NavItem to display.
- Hover text has been added to ensure users can still see the entire resource title
- The Avatar should always remain in view 

## Screenshots
Screenshot 2023-09-26 at 10 48 39


aptkingston (approved):
LGTM!

PR #11891, opened 2023-09-26
Chore - Clean unused dependencies

adrinr:
## Description
Remove unused dependencies in order to reduce dependencies install time. Also deduplicated the yarn.lock file in order to avoid duplicated packages


mike12345567 (approved):
Looks like a great cleanup - LGTM!

PR #11895, opened 2023-09-26
Fix issue where one to one relationships could be selected

PClmnt:
## Description
Also extracts the logic for defining the relationship options to the parent component, with the benefit of making the RelationshipSelector more reusable.


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -246,10 +270,7 @@
   $: external = table.type === "external"
   // in the case of internal tables the sourceId will just be undefined
   $: tableOptions = $tables.list.filter(
-    opt =>
-      opt._id !== $tables.selected._id &&
-      opt.type === table.type &&
-      table.sourceId === opt.sourceId
+    opt => opt.type === table.type && table.sourceId === opt.sourceId
```
adrinr, line 273: Why did we move the filtering of the current table to the selector component?

adrinr (approved):
LGTM, just a small comment

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
PClmnt: Because both sides of the relationship need different table options and this was filtering out the current table which was needed in the other side. So instead of passing two different tableOptions props - I moved it to the Selector so we can filter it there.

Issue #11807, opened 2023-09-19
[BUDI-7508] Relationship fields disappears from the data area when a new column is added. Internal data source


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.10.10
  - App Version: 2.10.10

**Describe the bug**
Relationship column disappears whenever a new column is created for internal table databases. It also doesn't affect other relationships within the table. It only occurs with the first relationship field, others seem fine.

You also can't delete or recreate the relationship field (As it will display that the column already exists when trying to recreate.). The only way to remove the field is by deleting the tables. 

**To Reproduce**
1. Create an app
2. Create 2 tables with the Budibase internal db
3. Add 2 columns e.g. name
4. Add a relationship column to link the 2 tables
5. Add a row of data to both tables
6. Link these records with the relationship field
7. Add a new column
8. On save see that the relationship field is no longer visible
9. After a page refresh you still can't see it

**Expected behavior**
Relationship fields should remain visible at all times within the data area. 

**Screenshots**
![relationship-internal-table-issue](https://github.com/Budibase/budibase/assets/126772285/69870c0b-27e8-4243-971d-2ca23c40f0ba)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Brave
 - Version v1.58.127

**Additional Info**
There are no console errors/network errors when this happens. There is also no errors after page refresh.

<sub>[BUDI-7508](https://linear.app/budibase/issue/BUDI-7508/relationship-fields-disappears-from-the-data-area-when-a-new-column-is)</sub>

PR #11897, opened 2023-09-26
Relationships disappearing when adding new columns fix

mike12345567:
## Description
When a relationship is added it creates schema changes in both related tables, however the svelte store is only updated for the table you created the relationship in, rather than both tables. I've added some functionality to make sure that when adding/updating tables all related tables are also fetched to make sure the local state is correct.

Addresses:
- https://linear.app/budibase/issue/BUDI-7508/relationship-fields-disappears-from-the-data-area-when-a-new-column-is
- https://github.com/Budibase/budibase/issues/11807

### DIFF
--- a/packages/builder/src/stores/backend/tables.js
+++ b/packages/builder/src/stores/backend/tables.js
@@ -81,12 +81,13 @@ export function createTablesStore() {
     replaceTable(savedTable._id, savedTable)
     select(savedTable._id)
     // make sure tables up to date (related)
-    const tableIdsToFetch = []
+    let tableIdsToFetch = []
     for (let column of Object.values(updatedTable?.schema || {})) {
       if (column.type === FIELDS.LINK.type) {
         tableIdsToFetch.push(column.tableId)
       }
     }
+    tableIdsToFetch = [...new Set(tableIdsToFetch)]
     // too many tables to fetch, just get all
     if (tableIdsToFetch.length > 3) {
       await fetch()
@@ -114,7 +115,6 @@ export function createTablesStore() {
     indexes,
   }) => {
     let draft = cloneDeep(get(derivedStore).selected)
-    console.log(draft)
 
     // delete the original if renaming
     // need to handle if the column had no name, empty string
--- a/packages/builder/src/stores/backend/tables.js
+++ b/packages/builder/src/stores/backend/tables.js
@@ -81,13 +81,18 @@ export function createTablesStore() {
     replaceTable(savedTable._id, savedTable)
     select(savedTable._id)
     // make sure tables up to date (related)
-    const tableUpdates = []
+    const tableIdsToFetch = []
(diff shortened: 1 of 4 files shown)


packages/builder/src/stores/backend/tables.js
```diff
@@ -63,6 +80,19 @@ export function createTablesStore() {
     const savedTable = await API.saveTable(updatedTable)
     replaceTable(savedTable._id, savedTable)
     select(savedTable._id)
+    // make sure tables up to date (related)
+    const tableIdsToFetch = []
+    for (let column of Object.values(updatedTable?.schema || {})) {
+      if (column.type === FIELDS.LINK.type) {
+        tableIdsToFetch.push(column.tableId)
+      }
+    }
```
aptkingston, line 89: Might be worth throwing in a wee:
```js
tableIdsToFetch = [...new Set(tableIdsToFetch)]
```
Just to ensure we don't have any duplicates - since it's possible that you have 2 relationships to the same table I guess.

aptkingston (approved):
LGTM! I'm guessing this same scenario breaks for multi-user collaboration though - since we probably just send down the "primary" updated table, but probably don't notify other users that 2 table schemas actually changed. Or maybe we do, I'm not sure. It just came to mind there when thinking about this issue.

packages/server/src/api/controllers/table/index.ts
```diff
@@ -80,7 +82,7 @@ export async function save(ctx: UserCtx<SaveTableRequest, SaveTableResponse>) {
   ctx.eventEmitter &&
     ctx.eventEmitter.emitTable(`table:save`, appId, { ...savedTable })
   ctx.body = savedTable
-  builderSocket?.emitTableUpdate(ctx, { ...savedTable })
+  builderSocket?.emitTableUpdate(ctx, cloneDeep(savedTable))
```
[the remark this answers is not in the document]
mike12345567: I switched it for a `cloneDeep` as I assumed that was the purpose of the `{ ...savedTable }` - to create a separate copy of the table schema, there wasn't an issue here other than a lack of typing support with the initial approach, `cloneDeep` is type safe.

packages/server/src/integrations/utils.ts
```diff
@@ -184,7 +184,9 @@ export function getSqlQuery(query: SqlQuery | string): SqlQuery {
   }
 }
 
-export const isSQL = helpers.isSQL
+export function isSQL(datasource: Datasource) {
```
adrinr, line 187: What is the reason for this?

mike12345567: For whatever reason I was getting IDE errors with typing around this - it wasn't fully understanding the typing of shared-core utilities. Will try to look into that separately, but this utility is quite important/useful and its nice to be able to see its typing easily without needing to jump into shared-core.

PR #11898, opened 2023-09-26
Splitting worker and server tests runners

adrinr:
## Description
These tests are the flakiest given it's heavy resource usage and conflicting with the limited GitHub runners. Splitting each one into its own pipeline will hopefully fix flakiness and speed up processes


mike12345567 (approved):
Looks great - good idea!

shogunpurple (approved):

PR #11906, opened 2023-09-27
Hide Portal Sidebar When the User Has No Apps

Ghrehh:
## Description
Design doesn't want this popping up during the app onboarding flow.


aptkingston (approved):
LGTM!

PR #11908, opened 2023-09-27
Fixing DS+ save rows

adrinr:
## Description
Fixing DS+ user field type, ensuring that the fields are saved as retrieved properly.
Also refactoring the `CreateEditColumn.svelte` file in order to remove hardcodings and make the fields more extensible and rely only on the field configuration.
Added tests for the row API


mike12345567 (approved):
LGTM!

Issue #11038, opened 2023-06-27
[BUDI-7236] Navigation Link Roles not working as expected


tglynx:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists
This issue may be related to #7917, however, in that case a navigation link is not visible when assigned to a custom role directly. 
Here a navigation link is visible to all members of the Basic role when the link is assigned to a role that inherits from 'Basic' (as an example). This may also be a security issue.

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.6.20
  - App Version: 2.6.20

**Describe the bug**
I have a custom role with the following setup: 
Name: Member
Inherits Role: Basic
Base Permission: Read / Write

As per my understanding the 'Member' role inherits all the permissions of the 'Basic' role which is fine. 
However, if i assign the 'Member' Role to a 'Navigation Link' this link can also be seen by members of the 'Basic' role.
I would assume that just members of the 'Member' role should be able to see the link.

**To Reproduce**
1. Create a custom role based on 'Basic' with Read/write permissions
2. Assign the new role to a navigation link
3. Log in with a member of a lower role (Basic) 
4. See error (link assigned to higher role is visible)

**Expected behavior**
I would expect that a Basic User can only see navigtion links that have been assigned to Basic or Public roles.
I would expect that a navigation link that has a custom role assigned is only visible to members of this custom role or roles that inherit from this custom role.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 114.0.5735.133

<sub>[BUDI-7236](https://linear.app/budibase/issue/BUDI-7236/navigation-link-roles-not-working-as-expected)</sub>

tglynx:
I add some additional information as i really think this should be fixed and to try to display the problem as easy as possible!

I have created a custom role called 'Member' that inherits from 'Power' (same behaviour observed when using Basic)
<img width="346" alt="Bildschirmfoto 2023-08-19 um 11 55 13" src="https://github.com/Budibase/budibase/assets/30650435/0b4fe4fe-44fb-49d0-8cbf-432f777de1ed">

In this example we have look at two navigational links (/profile and /debug): 
<img width="827" alt="Bildschirmfoto 2023-08-19 um 11 54 11" src="https://github.com/Budibase/budibase/assets/30650435/8a6a6e00-93e8-4e8d-8246-25d0e34efa49">
[...]

PR #11911, opened 2023-09-27
Fix navigation bar custom role restrictions

mike12345567:
## Description
The navigation bar within the client apps can have roles restricted behind a certain role level. For example we can create a "Member" custom role and then have two screens:
1. /basic - which is restricted to "BASIC" role - and the link should only be visible to those above basic.
2. /member - which is restricted to "Member" roles - and link should only be visible to those that have the custom member role.

Today this doesn't work - custom roles aren't detected and links with a custom role visibility attached simply appear for everyone. This fixes this issue by requesting a list of accessible roles from the backend, which can then be used for lookups.

This endpoint handles builders/admins, public users and role based users, with a test case to check each of these. From there the frontend only needs to compare against this list to decide what links should appear.

I have also done a bit of work in this PR to make it easier to debug the backend while using client apps - before now if you ran the codebase directly from typescript (with say `ts-node`)  it would be unable to load the client library version and it would not refresh in dev mode - I have fixed this.

## Addresses
- https://linear.app/budibase/issue/BUDI-7236/navigation-link-roles-not-working-as-expected
- https://github.com/Budibase/budibase/issues/11038


packages/client/src/components/app/Layout.svelte
```diff
@@ -61,7 +60,8 @@
   })
   setContext("layout", store)
 
-  $: validLinks = getValidLinks(links, $currentRole)
+  $: accessibleRoles = $roleStore
+  $: validLinks = getValidLinks(links)
```
aptkingston, line 64: We should pass in the value of the role store here as a param, otherwise validLinks won't be reactively recomputed when the value of the role store changes, because it's silently used inside the function.

If we just refactor this to
```js
$: validLinks = getValidLinks(links, $roleStore)
```
Then it will be reactive to both values, which it depends on.

aptkingston (commented):
Just one suggested change to ensure things are properly reactive!

Another thing we might want to consider here is that you're able to change the role you're previewing an app as, and I don't see anything in this that would be reactive to that. The old implementation used the `currentRole` derived store to account for this, but in this case we might need to just rehydrate the roles store whenever the devtools store role changes. 

That, in combination with the suggested change here, should ensure the links are reactively updated when changing the preview role.

PR #11919, opened 2023-09-28
Fix column type on Google integration

adrinr:
## Description
Allowing column type in google integration

## Screenshots
image

image

### DIFF
--- a/packages/server/src/api/controllers/row/ExternalRequest.ts
+++ b/packages/server/src/api/controllers/row/ExternalRequest.ts
@@ -269,13 +269,25 @@ function isEditableColumn(column: FieldSchema) {
   return !(isExternalAutoColumn || isFormula)
 }
 
-export class ExternalRequest {
-  private operation: Operation
-  private tableId: string
+export type ExternalRequestReturnType<T> = T extends Operation.READ
+  ?
+      | Row[]
+      | {
+          row: Row
+          table: Table
+        }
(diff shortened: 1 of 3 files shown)


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -269,13 +269,25 @@ function isEditableColumn(column: FieldSchema) {
...
+      | {
+          row: Row
+          table: Table
+        }
+  : {
+      row: Row
+      table: Table
+    }
+
+export class ExternalRequest<T extends Operation> {
```
adrinr, line 284: This will type the external request runs, so the response will be defined instead of having to cast it everytime

packages/server/src/api/controllers/row/external.ts
```diff
@@ -68,12 +72,10 @@ export async function patch(ctx: UserCtx<PatchRowRequest, PatchRowResponse>) {
     id: breakRowIdField(_id),
     row: dataToUpdate,
   })
-  const row = await sdk.rows.external.getRow(tableId, _id, {
-    relationships: true,
-  })
+  const row = await outputProcessing(table, response.row)
```
adrinr, line 75: We already have the row on the response, we don't need to refetch it

mike12345567 (approved):
Very nice!

PR #11920, opened 2023-09-28
Fixing query row index on google sheets

adrinr:
## Description
Fixing row index while updating/deleting from queries using a google sheets datasource

Addresses: 
- [BUDI-7249](https://linear.app/budibase/issue/BUDI-7249/google-sheet-rowindex-error-off-by-1-or-2)


mike12345567 (approved):
LGTM!

Issue #11878, opened 2023-09-25
[BUDI-7555] Saving pre-filled relation picker doesn't work


arminus:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.8.14

**Describe the bug**
I have a form with a relation picker for a relation in an internal Budibase DB: customer relation from an invoice table to a customer table - which at this stage only has a name column).

My form saves the value of the chosen customer correctly when the customer is changed. But when I load the form with an existing relation (which is shown correctly in the form), the same save action sets the relation to empty in the table

**To Reproduce**
![269007351-50ebc0e2-3540-4792-9854-fd8f84f1c8ad](https://github.com/Budibase/budibase/assets/683680/1efcd2ce-a009-48bd-b69a-0e958c709a4c)

**Expected behavior**
Relation picker should save the value regardless of initialization.

**Screenshots**
So I looked at the post requests for the 2 different saves (with a modified and unmodified form), turns out the the content of _value.Customer and Customers are different:

For the modified form (where the save works):

![modified](https://github.com/Budibase/budibase/assets/683680/56312880-a5be-4e20-a812-480478257fb2)

For the unmodified form (where the save doesn't work):
![unmodified](https://github.com/Budibase/budibase/assets/683680/46c224c7-1949-4a9c-97e0-9280816b7956)

**App Export**
[Invoices-export-1695655142123.tar.gz](https://github.com/Budibase/budibase/files/12716908/Invoices-export-1695655142123.tar.gz)

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: Firefox/Brave
 - Version: 117/1.45 Chromium 107

<sub>[BUDI-7555](https://linear.app/budibase/issue/BUDI-7555/saving-pre-filled-relation-picker-doesnt-work)</sub>

melohagan:
Fix will be deployed as part of the next release

PR #11928, opened 2023-09-29
Saving pre filled relation picker doesnt work

melohagan:
## Description
Address initial values being re-selected after updating to valid format.

Addresses: 
- https://github.com/Budibase/budibase/issues/11878


packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -60,6 +60,9 @@
       // even if they are not in the inital fetch results
       initialValuesProcessed = true
       optionsObj = (fieldState?.value || []).reduce((accumulator, value) => {
+        if (!value._id) {
```
adrinr, line 63: Can we put a comment on why this was added? Or pointing to the linear ticket?

adrinr (changes requested):

adrinr (approved):

PR #11932, opened 2023-09-29
[Changed] Adapt models to per-user/per-creator price

jvcalderon:
## Description

In order to support per-user/per-creator price model we need to modify existen models:

1. Added users and creators to license (amounts and quantities).
2. Added amount and quantities to subscriptions to allow more than one product.

## Addresses

https://linear.app/budibase/issue/GROW-171/per-creator-per-user-license-model


shogunpurple (approved):
LGTM!

jvcalderon:
It still being needed.

PR #11934, opened 2023-09-29
Fix deleting one-to-many relation in SQL

adrinr:
## Description
Fixing deleting one-to-many relation in SQL.
Also extending the row API testing in order to test the relationships as well we did for the BB references


mike12345567 (approved):
LGTM!

PR #11939, opened 2023-10-02
Fix client initialisation and fix navigation links being hidden

aptkingston:
## Description
This PR contains 2 fixes for recent client library regressions.

- Fix initialising client library 3 times when loading an app. This was caused by redundant calls to update the client library active role, and by not skipping changes when the role does not change.
- Navigation links without a role set have always been considered as basic. This seems to have been accidentally changed, meaning those links would now always be hidden (for all users). I've updated this to ensure links without an explicit role are treated as basic and show up as expected.


shogunpurple (approved):
LGTM

adrinr (approved):

PR #11954, opened 2023-10-02
New image for single images

adrinr:
## Description
Updating single image docker in order to be able to be cachable. These changes are for now not applied in the actual single image release pipeline, but they are used in a test pipeline and the same Dockerfile is used for the feature branch environments.
This new image will not have dependencies with the npm registry, so we will be able to test the feature branches as a whole.
It will require some changes on the release side that can be found in this PR: https://github.com/Budibase/budibase-deploys/pull/130

This PR also contains multiple package updates in order to fix some security and performance issues and in order to remove the dependency of pulling down the Budibase dependencies from the registry

- Testing single image build and release: https://github.com/Budibase/budibase/actions/runs/6534924939
- Testing feature branch release: https://github.com/Budibase/budibase-deploys/actions/runs/6534615225/job/17742218626

Feature branch URL: http://fb-budi-7573-use-existing-image-cache.fb.qa.budibase.net/builder/auth/login


scripts/removeWorkspaceDependencies.sh
```diff
@@ -0,0 +1,52 @@
...
+  @budibase/string-templates
+  @budibase/types
+  @budibase/worker
+)
+
+
+root_package_json=$(cat "package.json")
+
+process_package() {
+  local pkg="$1"
```
mike12345567, line 24: How does this work with loading the dependencies of the packages themselves? Like if we remove `@budibase/backend-core` from the `package.json` how do we fetch its dependencies?

I may be mis-understanding this as well - just trying to understand the script!

shogunpurple (approved):
LGTM Adri - I think we should write some documentation in confluence on how this build works though at a high level just for reference.

scripts/removeWorkspaceDependencies.sh
mike12345567: Ok cool - good to understand!

mike12345567 (approved):
Looks good to me! If adding documentation to Confluence should probably also include the note about Node 18 and IPv6 vs v4 (issues with it) as the single image is the only thing that will be affected by this (nothing else needs to access local resources).

adrinr:
Wrong push merged an unexpected node_modules. Removing it in here: https://github.com/Budibase/budibase/pull/12081

PR #11962, opened 2023-10-03
Filtering data providers by the new user column type

adrinr:
## Description
Building filters in order to support the new user column type

Addresses: 
- [BUDI-7583 - Filtering data provider with a `user column` does not work, the type of filter does not show and only blank.](https://linear.app/budibase/issue/BUDI-7583/filtering-data-provider-with-a-user-column-does-not-work-the-type-of)

## Screenshots
image


mike12345567 (approved):
LGTM!

Issue #11569, opened 2023-08-21
[BUDI-7410] Drag and drop not working:  multi-select option config


joebudi:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Budimaster - Cloud

**Describe the bug**
I added a multi-select field to my schema. I then added 6 options. I want to rearrange the options but DnD is not working.

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/d1ebf9bf-b628-4455-bc31-2bd67e2ae382/0fb2d080-c73a-445b-949a-3d53592751cf?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9kMWViZjliZi1iNjI4LTQ0NTUtYmMzMS0yYmQ2N2UyYWUzODIvMGZiMmQwODAtYzczYS00NDViLTk0OWEtM2Q1MzU5Mjc1MWNmIiwiaWF0IjoxNjkyNjE2NjQ3LCJleHAiOjE2OTI3MDMwNDd9.hhP6T-ggBsWWN5H-3EHjP-IkDkBV1syQEVrEbozBcgs)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7410](https://linear.app/budibase/issue/BUDI-7410/drag-and-drop-not-working-multi-select-option-config)</sub>

melohagan:
Same also applies to a newly added options field.

aptkingston:
What browser or OS is this? What happens when you drag?

This is working fine for me in Chrome, although in Firefox when you release drag it closes the popover.

melohagan:
This is so strange, I was sure that I couldn't drag the options, and now I can. I was using Google Chrome.

The last time I tried, the grab hand was definitely displaying, but I couldn't drag.

melohagan:
Ah I've figured out the reproduction steps:

1. Add three options
2. Rename the third option, and without clicking off the text input, try to drag.
3. Now the drag is broken

Page refresh will fix

aptkingston:
> Ah I've figured out the reproduction steps:
> 
> 1. Add three options
> 2. Rename the third option, and without clicking off the text input, try to drag.
> 3. Now the drag is broken
> 
[...]

PR #11964, opened 2023-10-03
Do not re-initialise options if changed

melohagan:
## Description
Fix for not being able to re-order changed option. 

Addresses: 
- https://github.com/Budibase/budibase/issues/11569


PClmnt (approved):
LGTM, nice fix!

PR #11981, opened 2023-10-05
Internal DB enrichment performance improvements

mike12345567:
## Description
This PR really provides one major change to the enrich endpoint and then makes some minor adjustments to the link system to try and improve efficiency of the calls.

In general, this is pretty much as optimised as it can be - this does not make any headway to improving the self call, but there is very little we can do there as we don't know the scope of what is needed from the user structure - if it is a heavily modified user table (with a lot of relationships) we cannot improve the speed of this.

The biggest change here is that when the data provider requests a to enrich a row, it doesn't need all of the columns that come back, it generally only looks at a single field. In fact within the API functions of frontend-core it strips out everything else it gets back, so this is wasted if it is retrieved. I've added a `field` query string parameter to the enrich `GET` call which allows restricting the call down to exactly what the data provider needs, significantly improving the performance of this call.

Below is a before and after on a local network, note the difference in size as well, this makes a big difference in our cloud platform where the size creates extra latency ontop of the actual work done in the request.

## Screenshots
Before (enrichment call)
image

After (enrichment call)
image


shogunpurple (approved):
LGTM! Nice work

PR #11988, opened 2023-10-06
Multi user column

adrinr:
## Description
Rethinking and reenabling multi-user column. This column was changed from being stored as a CSV to being an array, allowing complex filtering operations.

Addresses: 
- [BUDI-7589 - User column - multi-user filtering support](https://linear.app/budibase/issue/BUDI-7589/user-column-multi-user-filtering-support)

### Filtering by user (single column)
image

### Filtering by users (multiple column)
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -62,8 +62,6 @@
     return `${type}${subtype || ""}`.toUpperCase()
   }
 
-  $: console.warn(fieldDefinitions)
-
   let originalName
   let linkEditDisabled
   let primaryDisplay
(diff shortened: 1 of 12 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -659,18 +694,20 @@
     <Button primary text on:click={openJsonSchemaEditor}
       >Open schema editor</Button
     >
-  {:else if editableColumn.type === USER_REFRENCE_TYPE}
-    <!-- Disabled temporally -->
-    <!-- <Toggle
-      value={editableColumn.relationshipType === RelationshipType.MANY_TO_MANY}
+  {:else if editableColumn.type === FieldType.BB_REFERENCE && [FieldSubtype.USER, FieldSubtype.USERS].includes(editableColumn.subtype) && datasource?.source !== SourceName.GOOGLE_SHEETS}
```
mike12345567, line 697: Could we break out the logic here into a named function? Just so we can comment what is happening here - it'll be hard to remember what it is in the future.

mike12345567 (approved):
Looks great, some nice typing improvements as well!

PR #11989, opened 2023-10-06
Current user ID user column search mapping

mike12345567:
## Description
Adding a mapping layer to search queries so that we can map search inputs based on the table schema if desired - primarily for the user column.

By default the `{{ Current User._id }}` will return the table within the app, the user metadata table, which is not the global user ID. We have to map these IDs, this functionality allows us to lookup the schema of a column and then decide if the mapping is needed. This could be expanded to cover other user attributes, like email.

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
@@ -191,7 +191,7 @@
   }
 
   const getValidOperatorsForType = filter => {
-    if (!filter.field) {
+    if (!filter?.field) {
       return []
     }
 
--- a/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
+++ b/packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
@@ -191,7 +191,7 @@
   }
 
   const getValidOperatorsForType = filter => {
-    if (!filter) {
+    if (!filter.field) {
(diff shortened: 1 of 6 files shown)


packages/builder/src/components/design/settings/controls/FilterEditor/FilterDrawer.svelte
```diff
@@ -191,7 +191,7 @@
   }
 
   const getValidOperatorsForType = filter => {
-    if (!filter) {
+    if (!filter.field) {
```
adrinr, line 194:
```suggestion
    if (!filter?.field) {
```

packages/server/src/sdk/app/rows/search/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+    const rowPrefix = DocumentType.ROW + SEPARATOR
+    // TODO: at some point in future might want to handle mapping emails to IDs
+    if (filterValue.startsWith(rowPrefix)) {
+      return dbCore.getGlobalIDFromUserMetadataID(filterValue)
+    }
+    return filterValue
+  })
+}
+
+export function searchInputMapping(table: Table, options: SearchParams) {
```
adrinr, line 42: Should this be called enrich?

mike12345567: Its not really enrichment, we aren't adding anything, we're just mapping any inputs that aren't valid to something that we know is valid.

adrinr (approved):

PR #11990, opened 2023-10-06
Use nx read-write token on CI

adrinr:
## Description
Making usage of the NX read-write token to speed up CI pipelines


shogunpurple (approved):
LGTM!

Issue #11952, opened 2023-10-02
[BUDI-7585] Error on testing "SEND EMAIL (SMTP)" automation


ArthurLERAY:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: handipharm

**Describe the bug**
After creating an automation to send email via SMTP, I get two errors when clicking "Finish and test automation".

The first one is `Cannot read properties of undefined (reading 'primaryDisplay')` on click and then a GET 404 on `https://handipharm.budibase.app/api/.../rows` with a message `deleted`.
After clicking, I can't go to any other tab ("Data", "Design", ...) and nothing seems to work.

**To Reproduce**
1. Go to 'Automation'
2. Create a trigger on row created
3. Create a first step to send email via SMTP
4. Fill fields "Send To" and "Send From" with an email address
5. Fill email subject with some words
6. Fill HTML Contents with some basic HTML or leave empty

**Expected behavior**
Nothing will happen. Then, when looking into the developper console, 2 errors will appear as described earlier.

**Screenshots**
  
<img width="460" alt="Capture d’écran 2023-10-02 à 17 57 05" src="https://github.com/Budibase/budibase/assets/56477022/e4c052f6-3863-4813-b321-0008ab7467e9">

**Desktop (please complete the following information):**
 - OS: macOS Ventura 13.5
 - Browser Google Chrome
 - Version 116.0.5845.187

<sub>[BUDI-7585](https://linear.app/budibase/issue/BUDI-7585/error-on-testing-send-email-smtp-automation)</sub>

melohagan:
Hi @ArthurLERAY 

I am unable to recreate this. Could you please provide more information about the automation trigger please?

Is this Row Created trigger for an internal DB, or an external DB? What fields are involved, can you provide the schema?

Can you share an app export please, and/or a recording demonstrating the issue.

FGRibreau:
- It's a row trigger from an internal DB (see the automation here: https://handipharm.budibase.app/builder/app/app_dev_handipharm_be4376b50e1a409c9c222e61e1d8a767/automation/au_11489051db9c470caa314ffb1da
- ![Sans titre](https://github.com/Budibase/budibase/assets/138050/b5ab6fc2-bee7-4e61-af68-7f4ac35dc397)
0e6a1 )

Regarding the export to what email can we send it?

melohagan:
[...]

PR #11991, opened 2023-10-06
Allow relationship to be deleted even if missing

melohagan:
## Description
A user reported a bug relating to relationships in automations, however the root cause was that they had a relationship column that was pointing to a table that no longer existed.

When trying to delete that column, it was failing due to a `missing` error thrown from couch.

Fix added to not throw an error when attempting to update the column on the related table if it's missing.

Addresses: 
- https://github.com/Budibase/budibase/issues/11952

### DIFF
--- a/packages/server/src/db/linkedRows/LinkController.ts
+++ b/packages/server/src/db/linkedRows/LinkController.ts
@@ -316,6 +316,7 @@ class LinkController {
       }
       await this._db.put(linkedTable)
     } catch (error: any) {
+      // ignore missing to ensure broken relationship columns can be deleted
       if (error.statusCode !== 404) {
         throw error
       }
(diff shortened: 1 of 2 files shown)


packages/server/src/db/linkedRows/LinkController.ts
```diff
@@ -308,12 +308,18 @@ class LinkController {
...
-      delete linkedTable.schema[field.fieldName]
+    try {
+      // remove schema from other table, if it exists
+      let linkedTable = await this._db.get<Table>(field.tableId)
+      if (field.fieldName) {
+        delete linkedTable.schema[field.fieldName]
+      }
+      await this._db.put(linkedTable)
+    } catch (error: any) {
+      if (error.statusCode !== 404) {
```
mike12345567, line 319: worth commenting why the 404 is ignored for future

mike12345567 (approved):
LGTM - one minor comment!

PR #11993, opened 2023-10-06
Fix for saving user metadata IDs - convert them to global IDs in user column

mike12345567:
## Description
Quick fix for saving `{{ Current User._id }}` - make sure it is the correct format of ID.


shogunpurple (approved):
LGTM!

PR #11995, opened 2023-10-06
Fix form block crash on no data tables

melohagan:
## Description
Null pointer fix - the `getDatasourceForProvider` function was returning undefined if an app had no data tables, and only queries. 
Now returning an empty object instead.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7603/page-stops-responding-after-clicking-on-component-in-design

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -949,13 +949,11 @@ export const buildFormSchema = (component, asset) => {
   if (component._component.endsWith("formblock")) {
     let schema = {}
     const datasource = getDatasourceForProvider(asset, component)
+    const info = getSchemaForDatasource(component, datasource)
 
-    // Return if an app has no data tables
-    if (!datasource) {
+    if (!info?.schema) {
       return schema
     }
-
-    const info = getSchemaForDatasource(component, datasource)
 
     if (!component.fields) {
       Object.values(info.schema)
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -948,12 +948,17 @@ export const buildFormSchema = (component, asset) => {
 
   if (component._component.endsWith("formblock")) {
     let schema = {}
-
     const datasource = getDatasourceForProvider(asset, component)
+
+    // Return if an app has no data tables
+    if (!datasource) {
+      return schema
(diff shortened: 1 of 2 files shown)


aptkingston (commented):
I'm a little confused by the changes to `dataBinding.js` -  it's already configured to handle null values for datasources. `getSchemaForDatasource` will simply do nothing if the `datasource` setting is nullish, so I don't think this change is needed.

Is the only change here not just to update `FieldConfiguration.svelte` to handle a null datasource? Which as far as I can tell is just updating
```js
resourceId = datasource.resourceId || datasource.tableId
```
to
```js
resourceId = datasource?.resourceId || datasource?.tableId
```
I think that's a little cleaner than using an empty object as a fallback for the datasource.

melohagan:
> I'm a little confused by the changes to `dataBinding.js` - it's already configured to handle null values for datasources. `getSchemaForDatasource` will simply do nothing if the `datasource` setting is nullish, so I don't think this change is needed.

The null pointer was on `info.schema` because `info` was null, due to the datasource being null. I could always do `Object.values(info?.schema ?? {}) ` instead if you prefer?

I'll make the other suggested change relating to the `resourceId`

aptkingston:
> > I'm a little confused by the changes to `dataBinding.js` - it's already configured to handle null values for datasources. `getSchemaForDatasource` will simply do nothing if the `datasource` setting is nullish, so I don't think this change is needed.
> 
> The null pointer was on `info.schema` because `info` was null, due to the datasource being null. I could always do `Object.values(info?.schema ?? {}) ` instead if you prefer?
> 
> I'll make the other suggested change relating to the `resourceId`

Ah you're totally right - obviously my brain was in Sunday mode. I think it threw me because the return statement was in between `getDatasourceForProvider` and `getSchemaForDatasource`.

But yea you're right that this does need changed - but I think what we should do is check that we have a schema after `getSchemaForDatasource` runs, to be double safe.

So maybe go for:
```js
const datasource = getDatasourceForProvider(asset, component)
const info = getSchemaForDatasource(component, datasource)
if (!info?.schema) {
  return schema
}
```

Just since technically you might have a datasource, but then still end up not getting back `info.schema` from `getSchemaForDatasource`, and then the early return wouldn't have caught this.

aptkingston (approved):
LGTM!

PR #12002, opened 2023-10-09
Fix/users column search mapping

adrinr:
## Description
Tweaking the recent PR https://github.com/Budibase/budibase/pull/11989 in order to handle the multiuser column


mike12345567 (approved):
LGTM!

PR #12003, opened 2023-10-09
Ban user column to be used as display column

adrinr:
## Description
Disabling user columns to be set as a display column. They add plenty of complexity, and there are workarounds using formulas instead

## Screenshots
image

### DIFF
--- a/packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
@@ -1,6 +1,6 @@
 <script>
   import { getContext, onMount, tick } from "svelte"
-  import { FieldType, FieldSubtype } from "@budibase/types"
+  import { FieldType } from "@budibase/types"
   import GridCell from "./GridCell.svelte"
   import { Icon, Popover, Menu, MenuItem, clickOutside } from "@budibase/bbui"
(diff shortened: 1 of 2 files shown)


packages/server/src/db/linkedRows/index.ts
```diff
@@ -208,7 +215,32 @@ export async function squashLinksToPrimaryDisplay(
...
         const obj: any = { _id: link._id }
         if (linkedTable?.primaryDisplay && link[linkedTable.primaryDisplay]) {
-          obj.primaryDisplay = link[linkedTable.primaryDisplay]
+          const fieldSchema = linkedTable.schema[linkedTable.primaryDisplay]
+          switch (fieldSchema.type) {
+            case FieldType.BB_REFERENCE:
+              const subtype = fieldSchema.subtype as FieldSubtype
+              switch (subtype) {
+                case FieldSubtype.USER:
+                  const related = await processOutputBBReferences(
```
mike12345567, line 224: One thing to consider here is performance if we have a lot of rows/links, this is inside a double loop, this await could massively delay things - I wonder if there is a better way we can manage this (like bulking the requests in someway?)

adrinr: As discussed, simplified to use a formula if required

mike12345567 (approved):
LGTM!

PR #12006, opened 2023-10-09
Allow user table changes in existing apps again

aptkingston:
## Description
This PR restores the ability to fully edit the user table inside existing apps, while keeping it readonly inside new applications. Old apps that are exported and imported again keep their existing state, i.e. you will still be able to edit the user table if that app allowed it when exported.

The new flag `disableUserMetadata` has to be negative since we want the non-existing case to mean editing *is* allowed. So the new flag signifies disabling editing by its prescence.

### Existing app
Notice the lack of the "why is my table readonly" button, and the ability to create columns.
image
image

### New app
Notice the users table is now readonly, and the message is displayed.
image
image

### Imported old app
Notice user table editing is still allowed, as it was in the original app when exported.
image

Addresses: 
- https://linear.app/budibase/issue/BUDI-7604/enable-editing-of-user-metadata-for-existing-apps

### DIFF
--- a/packages/builder/src/builderStore/store/frontend.js
+++ b/packages/builder/src/builderStore/store/frontend.js
@@ -64,6 +64,7 @@ const INITIAL_FRONTEND_STATE = {
   },
   features: {
     componentValidation: false,
+    disableUserMetadata: false,
   },
   errors: [],
   hasAppPackage: false,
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM!

mike12345567 (approved):

PR #12007, opened 2023-10-09
Grid support for all datasource types + custom datasources

aptkingston:
## Description
This PR adds support for all datasource to the grid block. It now supports everything the table supports. 
This PR also adds a new type of datasource - "JSON / CSV" which allows specifying an arbitrary blob of JSON (or generating it through any sort of binding) or upload a CSV file of data to use.

### All datasources
The new setting for picking datasources is the same as the table block - it shows everything including things like queries and relationships:
image

Using any sort of non-plus datasource (anything apart from tables and views) will mean the grid block is always fully readonly. You can still use row click actions to make custom interactions with your data.
image

### Manifest enhancements
When using a non DS+ datasource, the CRUD settings will be hidden. To enable this I've had to enhance the `dependsOn` functionality in `manifest.json`. It now suppports nested keys (such as `datasource.type`) and also supports matching against multiple values via an array (such as `["table", "viewV2"]`).

### JSON / CSV datasource
This was originally called "Custom" but I think it's better to be explicit with "JSON / CSV".
This new datasource is the very last option in the datasource picker:
image

Picking this setting shows the settings icon beside the datasource picker, the same as what appears when you pick a query:
image

This icon opens a binding drawer, where you can configure any sort of HBS or JS binding to return JSON, or indeed any sort of string which looks like data.
image

This new datasource intelligently guesses the types of your data and will render them appropriately. It will identify data that looks like it should be an "options" type and render coloured tags as appropriate. The datasource full supports sorting and all other grid customisation features.

### Upload CSV for datasource
Within the drawer shown above, there's a button to let you pick a CSV file. Choosing a CSV file will convert your CSV to JSON and save it as a binding. As soon as you pick a file, everything will be closed and this will save and be used as your datasource.
image
image

Reopening the binding drawer will show the JSON payload that was converted from your CSV file:
image

### Examples of new JSON / CSV datasource
Here's a list of valid examples to try with the new datasource type.

#### HBS
- `a, b, c`
- `[1, 2, 3]`
- `[{ "foo": "bar" }, { "foo": "baz" }]`
- CSV-like strings. Note that header rows are not accounted for if using true CSV syntax, and so will be included as data:
image

#### JS
Basic:
```js
return new Array(10).fill(null).map(() => ({
  name: "Bill",
  number: Math.random(),
}))
```
More data and types:
```js
return new Array(100).fill(null).map(() => ({
  name: "Bill",
  info: "Lorem ipsum dolor sit amet " + Math.random(),
  tags: ["Foo", "Bar", "Baz"],
  metadata: { foo: "bar" },
  number: Math.round(Math.random() * 100)
}))
```

#### CSV
Classic sales CSV:
[100 Sales Records.csv](https://github.com/Budibase/budibase/files/12847807/100.Sales.Records.csv)

Addresses: 
- https://linear.app/budibase/issue/BUDI-7547/grid-block-allow-the-grid-block-to-work-with-non-ds

## Feature branch env
[Feature Branch Link](http://fb-grid-all-datasources.fb.qa.budibase.net)


packages/client/manifest.json
```diff
@@ -5618,29 +5617,37 @@
...
       {
         "type": "boolean",
         "label": "Edit rows",
         "key": "allowEditRows",
-        "defaultValue": true
+        "defaultValue": true,
+        "dependsOn": {
+          "setting": "datasource.type",
+          "value": ["table", "viewV2"]
+        }
```
deanhannigan, line 5640: When testing, this locked row editing for non `["table", "viewV2"]` grids into `true` with no way to disable it. If they cannot disable editing, their onRowClick behaviour will never be accessible. In these scenarios we may need to ensure that editing is resolved to false in the grid component itself

aptkingston: Yep you're correct that this locks it at whatever the current setting is, but that's how all those conditional settings behave. We just need to ensure that we properly handle that inside the components, and we do in this case. If you use a non DS+ inside the grid then the prop is ignored, because it will override certain features based on the datasource type.

So even though the prop will be true, you'll notice that you can't actually edit rows.

aptkingston:
After a chat with @deanhannigan who gave this an in-depth review, the following changes and fixes have been made:
- Fix `onClick` setting not working when the "edit rows" setting is enabled, but hidden due to not using a DS+
- Fix "add rows" setting not working in isolation. It depended upon "edit rows" also being enabled in order to properly add rows
- Hide settings for columns, filtering and sorting when using a custom DS as we don't know the schema
- Fix "depends on" logic for nullish values
- Add ability to specify multiple "depends on" constraints
- Prevent sorting by certain unsortable fields. This is applied in the grid data section, in the grid itself, in tables and in "sort column" settings

deanhannigan (approved):
LGTM!

PR #12008, opened 2023-10-09
Import user column type

adrinr:
## Description
Allowing importing and exporting tables with column types

### Creating a new internal table with CSV data
image

### Importing a CSV into an existing internal table
image

### Importing a CSV into an existing external table
image

### DIFF
--- a/packages/server/src/utilities/schema.ts
+++ b/packages/server/src/utilities/schema.ts
@@ -162,6 +162,8 @@ export function parse(rows: Rows, schema: Schema): Rows {
         parsedRow[columnName] = columnData
           ? new Date(columnData).toISOString()
           : columnData
+      } else if (columnType === FieldTypes.BB_REFERENCE) {
+        parsedRow[columnName] = columnData && parseCsvExport(columnData)
       } else {
         parsedRow[columnName] = columnData
       }
--- a/packages/server/src/utilities/schema.ts
+++ b/packages/server/src/utilities/schema.ts
@@ -1,9 +1,12 @@
+import { FieldSubtype } from "@budibase/types"
 import { FieldTypes } from "../constants"
-import { ValidColumnNameRegex } from "@budibase/shared-core"
+import { ValidColumnNameRegex, utils } from "@budibase/shared-core"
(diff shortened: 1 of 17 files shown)


packages/server/src/utilities/schema.ts
```diff
@@ -155,3 +188,43 @@ export function parse(rows: Rows, schema: Schema): Rows {
     return parsedRow
   })
 }
+
+function isValidBBReference(
```
mike12345567, line 192: Can this be simplified, to just:
```js
const isValid = (userIdArray: string[]) => {
  const found = userIdArray.find(userId => !db.isGlobalUserID(d._id))
  return !found
}
switch (columnSubType) {
  case FieldSubtype.USER:
    if (typeof columnData !== "string") {
      return false
    }
    const userIdArray = parseCsvExport(columnData)
    if (userIdArray.length > 1) {
      return false
    }
    return isValid(userIdArray)
  case FieldSubtype.USERS:
    if (!Array.isArray(columnData)) {
      return false
    }
    return isValid(columnData)
}
```

adrinr: Both types require parsing the `columnData`, as it comes as CSV string

mike12345567 (approved):
LGTM!

packages/server/src/utilities/schema.ts
```diff
@@ -155,3 +188,32 @@ export function parse(rows: Rows, schema: Schema): Rows {
...
+  columnSubtype: FieldSubtype
+): boolean {
+  switch (columnSubtype) {
+    case FieldSubtype.USER:
+    case FieldSubtype.USERS:
+      if (typeof columnData !== "string") {
+        return false
+      }
+      const userArray = parseCsvExport<{ _id: string }[]>(columnData)
+      if (columnSubtype === FieldSubtype.USER && userArray.length > 1) {
```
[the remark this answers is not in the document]
mike12345567: Looks good!

mike12345567 (approved):
Yeap - looks good - gets get it in!

PR #12011, opened 2023-10-09
Fixing null in user column form field if using user metadata IDs

mike12345567:
## Description
Converting user IDs where necessary to global user IDs in the frontend, correcting the null entry when using current user IDs, or other user IDs, as a default value for a user column form type.

Before:
image

After:
image

### DIFF
--- a/packages/client/src/components/app/forms/BBReferenceField.svelte
+++ b/packages/client/src/components/app/forms/BBReferenceField.svelte
@@ -1,17 +1,31 @@
 <script>
   import RelationshipField from "./RelationshipField.svelte"
   import { sdk } from "@budibase/shared-core"
+
+  export let defaultValue
+
+  function updateUserIDs(value) {
+    if (Array.isArray(value)) {
+      return value.map(val => sdk.users.getGlobalUserID(val))
+    } else {
+      return sdk.users.getGlobalUserID(value)
+    }
+  }
+
+  function updateReferences(value) {
+    if (sdk.users.containsUserID(value)) {
+      return updateUserIDs(value)
+    }
+    return value
+  }
 </script>
 
 <RelationshipField
   {...$$props}
   datasourceType={"user"}
   primaryDisplay={"email"}
-  valueConversion={value => {
-    if (Array.isArray(value)) {
-      return value.map(val => sdk.users.getGlobalUserID(val))
-    } else {
-      return sdk.users.getGlobalUserID(value)
-    }
+  defaultValue={updateReferences(defaultValue)}
+  onChange={value => {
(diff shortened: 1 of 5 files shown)


packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -18,6 +18,7 @@
   export let filter
   export let datasourceType = "table"
   export let primaryDisplay
+  export let valueConversion
```
aptkingston, line 21: Could we hide the details of this inside `BBReferenceField` instead of introducing it to the relationship field? By that I mean that we can pass in a custom `onChange` prop which does the conversion, and also pass in a custom `defaultValue` prop, which is already converted.

In my opinion it's just a little cleaner to contain this conversion logic inside the place that it's needed, the reference field, and leave the relationship field untouched.

aptkingston (approved):
LGTM!

packages/client/src/components/app/forms/BBReferenceField.svelte
```diff
@@ -1,9 +1,31 @@
...
 </script>
 
 <RelationshipField
   {...$$props}
   datasourceType={"user"}
   primaryDisplay={"email"}
+  defaultValue={updateReferences(defaultValue)}
+  onChange={value => {
+    return updateReferences(value)
+  }}
```
aptkingston, line 30: Actually wait - this `onChange` function won't do anything. This is currently called when you configure on click actions, so we still need to call the `onChange` prop passed in to the reference field.

```js
export let onChange

...
<RelationshipField
  ...
  onChange={({ value }) => onChange({ value: updateReferences(value) })}
  ...
/>
...
```

should do it I think.

aptkingston (changes requested):

packages/client/src/components/app/forms/BBReferenceField.svelte
aptkingston: Although I don't think we need to worry about onChange at all do we? The values broadcast from the picker should always be the global IDs, since thats where the options are loaded from. I reckon the default value is maybe the only place we need to care about?

aptkingston: Just for clarity too, `onChange` is specifically only the callback for any "on change" settings configured in the design section. It's not used to actually update the value of the field - that's done by `fieldApi.setValue(value)`. But like I said in my comment above, I don't think we need to worry about that since the values set by the picker will always be global IDs since that's where we load the options from.

aptkingston (approved):

adrinr (approved):

PR #12020, opened 2023-10-10
Restrict display on import table

adrinr:
## Description
Only enable valid column types to be selected as display column when importing a table

## Screenshots
image


packages/shared-core/tsconfig.build.json
```diff
@@ -15,7 +15,8 @@
     "skipLibCheck": true,
     "paths": {
       "@budibase/types": ["../types/src"]
-    }
+    },
+    "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo"
```
adrinr, line 19: This fixes the issue with shared-core not being properly rebuilt on dev. We had the `build` and `build:dev` commands generating different tsbuildinfo, and it was conflicting with the incremental builds

mike12345567 (approved):
Amazing - good to finally sort the shared-core issue!

PR #12030, opened 2023-10-11
Fix lint pipeline issues

adrinr:
## Description
Removing conflicting `easimon/maximize-build-space` on lint process. Lint should not have a problem with the disk space, so not really required.
Also, making use of the read/write cloud nx token

## Feature branch env
[Feature Branch Link](http://fb-fix-lint-pipeline-issues.fb.qa.budibase.net)


mike12345567 (approved):
makes sense - linting really doesn't need it.

PR #12049, opened 2023-10-12
Inline searching for grid and grid block

aptkingston:
## Description
This PR adds inline searching to the grid and grid block. Searching is only supported on a subset of column types, which right now is `string`, `options`, `longform`, `array`, `number` and `formula` columns which are of the static type.

The search field itself is always a text field at the moment, which is the main reason searching on dates is currently unsupported. This may be improved upon in future.

The search operator used is the most sensible of the available options for a given type. For most text based columns that is currently a `string` operator (which is a "starts with"). For numbers we use an `equal` operator, and for multi-select columns we use a `contains` operator.

Filters can be cleared by either clearing the text field or pressing the dedicated clear icon.

### Using search
Hover over a header cell, and if search is supported then the left icon will transform into a search icon:
image

Pressing the icon will transform the header into a text input and focus is, ready to type to search:
image

Typing a search value will automatically start filtering the rows in the grid. Searching is automatically performed on keystrokes (debounced by a short delay to avoid spam):
image

Blurring the field will keep the filter active and highlight the header cell with a unique background color, highlighting to the user that this column has a filter applied:
image

Multiple filters can be applied at the same time and results are those that match ALL the specified filters:
image

Addresses: 
- https://linear.app/budibase/issue/BUDI-7542/grid-block-add-search-functionality

## Documentation

## Feature branch env
[Feature Branch Link](http://fb-grid-inline-searching.fb.qa.budibase.net)


packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
```diff
@@ -24,13 +25,19 @@
     definition,
     datasource,
     schema,
+    focusedCellId,
+    filter,
   } = getContext("grid")
 
+  const searchableTypes = ["string", "options", "number", "array", "longform"]
```
adrinr, line 32: Can we use the types from `FieldType`?

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
```diff
@@ -41,6 +48,17 @@
...
     : "Z-A"
+  $: searchable = isColumnSearchable(column)
+  $: searching = searchValue != null
+  $: debouncedUpdateFilter(searchValue)
+
+  const isColumnSearchable = col => {
+    const type = col.schema.type
+    return (
+      searchableTypes.includes(type) ||
+      (type === "formula" && col.schema.formulaType === "static")
```
adrinr, line 59: Same here, if we use the fields from `types` we will have safer refactors code as we will not depend on magic strings

packages/frontend-core/src/components/grid/cells/HeaderCell.svelte
aptkingston: I definitely will now that they're typed! That's great to have 👌

packages/frontend-core/src/fetch/ViewV2Fetch.js
```diff
@@ -35,9 +35,32 @@ export default class ViewV2Fetch extends DataFetch {
...
+
+    // If sort params are not defined, update options to store the sorting
+    // params built in to this view. This ensures that we can accurately
+    // compare old and new sorting params and skip a redundant API call.
+    if (!sortColumn && definition.sort?.field) {
+      this.options.sortColumn = definition.sort.field
+      this.options.sortOrder = definition.sort.order
+    }
+
+    // If sort params are not defined, update options to store the sorting
```
Ghrehh, line 57: Very NAB: duplicate comment here I believe, probably applies to both cases it's pasted above but maybe a single comment above both does the same trick?

Ghrehh (approved):
LGTM, small question though.

A row added while an inline search is in progress will be visible even if it doesn't match the search parameters, at least for string fields, just checking this is intentional, not sure if it being immediately hidden is preferable tbh

aptkingston:
> LGTM, small question though.
> 
> A row added while an inline search is in progress will be visible even if it doesn't match the search parameters, at least for string fields, just checking this is intentional, not sure if it being immediately hidden is preferable tbh

Yea I think that's the better alternative to creating a row just to have it immediately hidden. It's sort of a middle-ground though, because if you re-sort your data or do another search then the new row will be hidden. So it's basically just visible immediately after creation until the next filter/sort gets applied.

packages/frontend-core/src/fetch/ViewV2Fetch.js
aptkingston: Yep good catch, that's a dupe!

deanhannigan (approved):
LGTM. Couple of questions:

- Search case sensitive for MultiSelect column but not Options?
  - When searching for "fuel" in the expenses I get no results.
-  For V1 Views with a `groupBy` defined, the grid loads but its blank
  - Similar issue to the schema issue outlined here.
  - https://github.com/Budibase/budibase/pull/12135

aptkingston:
As discussed, just leaving the comments here for visibility:
> Search case sensitive for MultiSelect column but not Options?

Technical restriction unfortunately due to how the fields are indexed.

> For V1 Views with a `groupBy` defined, the grid loads but its blank

I'll attempt to support these old views by updating the `ViewFetch` to transform "groupBy" views into a valid schema structure.

aptkingston:
I've added support for old views with groupBy specified. Quick screenshot showing that working:
image

Issue #11843, opened 2023-09-21
[BUDI-7525] When fetching external tables - display errors for why certain tables cannot be fetched (e.g. no primary keys)


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase 2.10.10
  - App Version: 2.10.10

**Describe the bug**
Whenever fetching tables that don't have primary keys set, a success notification appears stating that the tables have been successfully fetched. The tables aren't synced across, I believe this is related to no primary keys being set.

**To Reproduce**
1. Your a Budibase app
2. Add a MySQL data source
3. Create a new table in it using the schema provided below
4. Try to fetch the table
5. See issue.

**Expected behavior**
Instead of a success notification, an error notification should be thrown explaining why a table failed to be fetched and synced into Budibase.

**Screenshots**
![Screen Recording 2023-09-21 at 10 59 42](https://github.com/Budibase/budibase/assets/126772285/b1080a79-d8cc-4d9c-b2b6-731ffe310462)
  
**Table create SQL**
```
CREATE TABLE noPrimaryKey (
    column1 INT,
    column2 VARCHAR(50),
    column3 DATE
);
```
**Additional context**
No console errors or network errors are thrown when this happens.

<sub>[BUDI-7525](https://linear.app/budibase/issue/BUDI-7525/when-a-mysql-table-doesnt-have-primary-keys-set-budibase-cannot-fetch)</sub>

**Related discussions**
- https://github.com/Budibase/budibase/discussions/11504
- https://github.com/Budibase/budibase/discussions/11836

jbueter:
@ConorWebb96 Thank you for creating this issue! I just wanted to add that this issue probably exists for all types of SQL databases. At least in our case, we had the issue with PostgreSQL as well.

Issue #11613, opened 2023-08-29
[BUDI-7433] Google Sheets validation won't let you import any sheets if one sheet in the doc is invalid


andz-bb:
**Checklist**

* [X] I have searched Budibase discussions and GitHub issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.9.30
  * App Version: 2.9.30

**Describe the bug**
If there's a sheet in your Google Sheets document without valid headers, it won't let you import any of the otherwise valid sheets as tables.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a Google Sheets doc with a sheet with valid headers and another sheet without
2. Set up the data source and try to fetch the valid sheet
3. See that it doesn't allow you to fetch that table because the invalid sheet exists

**Expected behaviour**
It should only validate sheets for headers if they are actually selected to be fetched

**Screenshots**

![image.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/6d777b36-cbfd-43af-9a83-a85a5dc1e74f/ef382bd9-f241-4385-a339-db033b41324b?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC82ZDc3N2IzNi1jYmZkLTQzYWYtOWE4My1hODVhNWRjMWU3NGYvZWYzODJiZDktZjI0MS00Mzg1LWEzMzktZGIwMzNiNDEzMjRiIiwiaWF0IjoxNjkzMzE1OTE1LCJleHAiOjE2OTM0MDIzMTV9.5XWsJjH98Obeo5ELpRRwkNQw0WVwf_jfScghuCjlllQ)

[...]

PR #12050, opened 2023-10-12
Surface errors when importing tables from a datasource, don't surface errors from tables that have been filtered out.

samwho:
## Description

In investigating https://github.com/Budibase/budibase/issues/11613 I noticed that most of the validation done against the fields of `DatasourcePlus` instances is never surfaced to the user. It's returned through the API, but the frontend does nothing with it.

The practical result of this PR is that the following endpoints:

- `POST /api/datasources`
- `POST /api/datasources/:datasourceId/schema`

No longer return an `error` key containing a string, but an `errors` key containing an object of `{ "tableName": "error message" }`. The idea is to make it possible to have an error message per table that we can filter so that an error on a single table doesn't prevent the user from importing any tables (the crux of https://github.com/Budibase/budibase/issues/11613).

This `errors` key is then surfaced in the UI and looks like this:

CleanShot 2023-10-16 at 17 10 58@2x

CleanShot 2023-10-16 at 17 42 12@2x

Addresses: 
- [BUDI-7356](https://linear.app/budibase/issue/BUDI-7356/google-sheet-errors-not-being-returned-to-ui)
- https://github.com/Budibase/budibase/issues/11613
- https://github.com/Budibase/budibase/issues/11843

## Documentation


packages/server/src/integrations/utils.ts
```diff
@@ -295,42 +292,41 @@ function copyExistingPropsOver(
...
-      errors[name] = BuildSchemaErrors.INVALID_COLUMN
-      continue
-    }
-    // make sure all previous props have been added back
     finalTables[name] = copyExistingPropsOver(name, table, entities, tableIds)
   }
-  // sort the tables by name
-  finalTables = Object.entries(finalTables)
+  // sort the tables by name (TODO(samwho): why?)
+  return Object.entries(finalTables)
```
mike12345567, line 309: on the TODO - really its just for a consistent response from the API, you can expect the tables to come back in a predictable order.

mike12345567 (approved):
Looks great! Nice to have a proper test case for the primary key check as well in Postgres, great work - nice to fix the schema generation to be actually return the results as well!

PR #12051, opened 2023-10-12
RelationshipField - load default and handle proper searching with user API

mike12345567:
## Description
The way that the relationship field expected to be able to search users was broken - it uses Lucene syntax to search which the user API did not support. I have updated the search API to allow this, but only certain supported keys, it will throw a 501 if anything that is unexpected is provided.

It can now search by the `email` and the `_id` of the user - this will be expanded once SQS is finished.

Also the relationship field component did not account for the default value not being in the paginated results, so you would get a `null` in the field if the value was not found. Solving this by on load checking for the default value and if there is one, attempting to search for it using the `equal._id` search operation. Hopefully this should solve this for both relationships and users.

## Feature branch env
[Feature Branch Link](http://fb-fix-user-search-api.fb.qa.budibase.net)

### DIFF
--- a/packages/backend-core/src/users/users.ts
+++ b/packages/backend-core/src/users/users.ts
@@ -14,13 +14,14 @@ import {
 } from "../db"
 import {
   BulkDocsResponse,
+  ContextUser,
+  SearchQuery,
+  SearchQueryOperators,
   SearchUsersRequest,
   User,
-  ContextUser,
 } from "@budibase/types"
-import { getGlobalDB } from "../context"
 import * as context from "../context"
-import { user as userCache } from "../cache"
+import { getGlobalDB } from "../context"
 
 type GetOpts = { cleanup?: boolean }
 
@@ -37,6 +38,27 @@ function removeUserPassword(users: User | User[]) {
     return users
   }
   return users
+}
+
+export const isSupportedUserSearch = (query: SearchQuery) => {
+  const allowed = [
+    { op: SearchQueryOperators.STRING, key: "email" },
+    { op: SearchQueryOperators.EQUAL, key: "_id" },
+  ]
+  for (let [key, operation] of Object.entries(query)) {
+    if (typeof operation !== "object") {
+      return false
+    }
+    const fields = Object.keys(operation || {})
+    const allowedOperation = allowed.find(
+      allow =>
+        allow.op === key && fields.length === 1 && fields[0] === allow.key
+    )
+    if (!allowedOperation && fields.length > 0) {
+      return false
+    }
+  }
+  return true
 }
 
 export const bulkGetGlobalUsersById = async (
@@ -211,8 +233,8 @@ export const searchGlobalUsersByEmail = async (
 
 const PAGE_LIMIT = 8
 export const paginatedUsers = async ({
-  page,
-  email,
+  bookmark,
+  query,
   appId,
 }: SearchUsersRequest = {}) => {
   const db = getGlobalDB()
@@ -222,18 +244,20 @@ export const paginatedUsers = async ({
     limit: PAGE_LIMIT + 1,
   }
   // add a startkey if the page was specified (anchor)
-  if (page) {
-    opts.startkey = page
+  if (bookmark) {
+    opts.startkey = bookmark
   }
   // property specifies what to use for the page/anchor
   let userList: User[],
     property = "_id",
     getKey
-  if (appId) {
+  if (query?.equal?._id) {
+    userList = [await getById(query.equal._id)]
+  } else if (appId) {
     userList = await searchGlobalUsersByApp(appId, opts)
     getKey = (doc: any) => getGlobalUserByAppPage(appId, doc)
-  } else if (email) {
-    userList = await searchGlobalUsersByEmail(email, opts)
+  } else if (query?.string?.email) {
+    userList = await searchGlobalUsersByEmail(query?.string?.email, opts)
     property = "email"
   } else {
     // no search, query allDocs
--- a/packages/backend-core/src/users/users.ts
+++ b/packages/backend-core/src/users/users.ts
@@ -50,11 +50,15 @@ export const isSupportedUserSearch = (query: SearchQuery) => {
       return false
     }
     const fields = Object.keys(operation || {})
+    // this filter doesn't contain options - ignore
+    if (fields.length === 0) {
+      continue
+    }
     const allowedOperation = allowed.find(
       allow =>
         allow.op === key && fields.length === 1 && fields[0] === allow.key
     )
-    if (!allowedOperation && fields.length > 0) {
+    if (!allowedOperation) {
       return false
     }
   }
--- a/packages/client/src/components/app/forms/RelationshipField.svelte
+++ b/packages/client/src/components/app/forms/RelationshipField.svelte
@@ -25,7 +25,6 @@
   let tableDefinition
   let searchTerm
   let open
-  let hasFetchedDefaultValue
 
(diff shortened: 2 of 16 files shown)


packages/backend-core/src/users/users.ts
```diff
@@ -39,6 +40,27 @@ function removeUserPassword(users: User | User[]) {
...
+  for (let [key, operation] of Object.entries(query)) {
+    if (typeof operation !== "object") {
+      return false
+    }
+    const fields = Object.keys(operation || {})
+    const allowedOperation = allowed.find(
+      allow =>
+        allow.op === key && fields.length === 1 && fields[0] === allow.key
+    )
+    if (!allowedOperation && fields.length > 0) {
```
adrinr, line 57: This is a little bit confusing. Will we ever have more than one field? On the line above we are checking that we only have one

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -105,19 +110,26 @@
...
-        query: { string: { [primaryDisplay]: searchTerm } },
+    // Don't request until we have the primary display or default value has been fetched
+    if (allRowsFetched || !primaryDisplay) {
+      return
+    }
+    if (defaultVal) {
+      await fetch.update({
+        query: { equal: { _id: defaultValue } },
       })
+      fetchedDefault = $fetch.rows?.[0]
```
adrinr, line 128: Why do we have this as a different object, instead of adding it to the `optionsObj`?

packages/backend-core/src/users/users.ts
mike12345567: The idea is to support something like `range: {}` where the operator has been supplied but with no search parameters. This is allowable as it technically doesn't nothing, but I could probably do something like stripping them out or skipping them which would make more sense.

packages/client/src/components/app/forms/RelationshipField.svelte
mike12345567: Good call - didn't realise this was an option!

adrinr: Much nicer! Thx :D

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -105,19 +105,29 @@
     }
   }
 
-  $: fetchRows(searchTerm, primaryDisplay)
+  $: fetchRows(searchTerm, primaryDisplay, defaultValue)
```
adrinr, line 108: This `defaultValue` can now be removed, right?

mike12345567: This is used for reactivity - I want to trick the `fetchRows` if the `defaultValue` changes for any reason.

packages/frontend-core/src/fetch/UserFetch.js
```diff
@@ -27,16 +28,25 @@ export default class UserFetch extends DataFetch {
...
+    const { limit, paginate } = this.options
     const { cursor, query } = get(this.store)
+    let finalQuery
+    // convert old format to new one - we now allow use of the lucene format
+    const { appId, paginated, ...rest } = query
+    if (!LuceneUtils.hasFilters(query) && rest.email) {
+      finalQuery = { string: { email: rest.email } }
+    } else {
+      finalQuery = rest
+    }
```
aptkingston, line 40: Did you discover where we use this old query? I'd be happy to update those places to use the normal structure now that we support it, but that can be done after in a separate PR to reduce scope of this one.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -105,19 +105,29 @@
...
       !$fetch.hasNextPage
-    // Don't request until we have the primary display
-    if (!allRowsFetched && primaryDisplay) {
-      fetch.update({
-        query: { string: { [primaryDisplay]: searchTerm } },
+    // Don't request until we have the primary display or default value has been fetched
+    if (allRowsFetched || !primaryDisplay) {
+      return
+    }
+    if (defaultVal) {
```
aptkingston: I think we need to check here if we already have the row we need for the default value. Otherwise we're going to do a search for the default value row every time, which blocks the main search call and means we're spending twice as long waiting.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -105,19 +105,29 @@
...
+      await fetch.update({
+        query: { equal: { _id: defaultValue } },
       })
+      const defaultRow = $fetch.rows?.[0]
+      if (defaultRow) {
+        optionsObj[defaultRow._id] = defaultRow
```
aptkingston, line 126: Do we need to manually set this? I would have expected the normal reactive statement for `enrichOptions` to have triggered here when this fetch returns, and pulled out the row that comes back here.

aptkingston (commented):
Looks good!

We'll need to check that other usages of the `searchUsers` API still works, since we call that from both the `BBReferenceCell` (which directly passes in an `email` so will be broken now I think) and the portal users store.

packages/client/src/components/app/forms/RelationshipField.svelte
mike12345567: Yeap - Adri raised this - fixed now!

packages/client/src/components/app/forms/RelationshipField.svelte
mike12345567: sadly as there is two operations happening in the one call here, two different fetches, the next one overwrites this one before the reaction it seems (fought with this last night).

packages/frontend-core/src/fetch/UserFetch.js
mike12345567: Yeah I'd rather leave the same consistent API to the `UserFetch` for now as fetching by email is nice anyway and really there only is two search options - don't want to expose the "search" until it actually works fully.

packages/client/src/components/app/forms/RelationshipField.svelte
aptkingston: Interesting! I've seen that behaviour before but never with something async like this. But yea that's no problem then. As long as it only fires when needed then extra duration isn't a problem, since it'll just be on the initial load.

aptkingston (commented):
Just one change suggested but otherwise LGTM!

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -105,19 +106,26 @@
...
       !$fetch.hasNextPage
-    // Don't request until we have the primary display
-    if (!allRowsFetched && primaryDisplay) {
-      fetch.update({
-        query: { string: { [primaryDisplay]: searchTerm } },
+    // Don't request until we have the primary display or default value has been fetched
+    if (allRowsFetched || !primaryDisplay) {
+      return
+    }
+    if (defaultVal && !hasFetchedDefaultValue) {
```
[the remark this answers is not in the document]
adrinr: This made me realise that I would expect `defaultVal` to be the value, not the id. Is it just the id?

aptkingston (approved):

adrinr (approved):

packages/client/src/components/app/forms/RelationshipField.svelte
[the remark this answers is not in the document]
mike12345567: Usually - although we can't really know what it will be - it might be nothing.

PR #12053, opened 2023-10-13
Bug warnings stale bot

melohagan:
## Description
- Tag high priority bugs with 'warn' after one month
- Tag medium priority bugs with 'warn' after three months
- Tag any bug with 'stale' after six months, and auto-close after 30 days.

## Feature branch env
[Feature Branch Link](http://fb-chore-stale-bot.fb.qa.budibase.net)

### DIFF
--- a/.github/workflows/stale_bot.yml
+++ b/.github/workflows/stale_bot.yml
@@ -2,7 +2,7 @@ name: Close stale issues and PRs # https://github.com/actions/stale
 on:
   workflow_dispatch:
   schedule:
-    - cron: '*/30 * * * *' # Every 30 mins
+    - cron: "*/30 * * * *" # Every 30 mins
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale_bot.yml
```diff
@@ -2,28 +2,41 @@ name: Close stale issues and PRs # https://github.com/actions/stale
...
-          close-issue-label: closed-stale
-          close-issue-message: This issue has been automatically closed it has not had any activity in 90 days."
           days-before-pr-close: 7
 
-          # exemptions
-          exempt-pr-labels: pinned,security,roadmap
+      - uses: actions/stale@v8
+        with:
+          # stale rules for high priority bugs
+          days-before-stale: 30
```
shogunpurple, line 23: should we put a warning on it, before the SLA expires? Rather than wait for it to happen then scramble to fix it

shogunpurple (approved):
LGTM! I like the solution of using 3 separate actions to assign the labels!

One thing I'm not sure about though is API quotas - essentially that github action just pulls a bunch of issues back from github, but not all so it doesn't hit the rate limits. I'm wondering if 3 separate instances of the job will go over the entire list of issues 3 times, maybe hitting limits. 

Just something worth us testing to see!

shogunpurple (approved):
LGTM! I like the solution of using 3 separate actions to assign the labels!

One thing I'm not sure about though is API quotas - essentially that github action just pulls a bunch of issues back from github, but not all so it doesn't hit the rate limits. I'm wondering if 3 separate instances of the job will go over the entire list of issues 3 times, maybe hitting limits. 

Just something worth us testing to see!

shogunpurple (approved):
LGTM! I like the solution of using 3 separate actions to assign the labels!

One thing I'm not sure about though is API quotas - essentially that github action just pulls a bunch of issues back from github, but not all so it doesn't hit the rate limits. I'm wondering if 3 separate instances of the job will go over the entire list of issues 3 times, maybe hitting limits. 

Just something worth us testing to see!

melohagan:
@shogunpurple If rate limits are a possible problem, I'm sure the cron could be changed from every 30mins to every few hours if you want me to do that?

shogunpurple:
@melohagan you can also adjust the amount of tickets it pulls at once as well - maybe a bit of trial and error involved

.github/workflows/stale_bot.yml
melohagan: Well the discord message is the warning - it doesn't have to be a scramble, it's just more drawing attention to it. I won't put any red siren emojis on the message 😄 

🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨 
Not this ^^

It'll be more: 

🐛 Fix me please! 🐞

shogunpurple: Makes sense! Also if we did that we'd probably need another one to say "ok this is actually past the deadline now". Happy to leave as is

PR #12054, opened 2023-10-13
Stop the `postgres` container after the Postgres integration tests run.

samwho:
## Description

I noticed when running `yarn jest src/integration-test/postgres.spec.ts` from within `packages/server` that `jest` will hang indefinitely after the tests have completed. This appears to be because of the `postgres` container still running, because adding in an `afterAll` block that stops the container explicitly fixes the problem.

## Documentation


adrinr (approved):
Nice catch :)

Issue #7917, opened 2022-09-22
Custom role - Navigation links


applicationuser:
### Discussed in https://github.com/Budibase/budibase/discussions/7849

<div type='discussions-op-text'>

<sup>Originally posted by **applicationuser** September 19, 2022</sup>
Hello ,

Need help with navigation link setup with custom roles.

anything i missed or setup incorrectly? Please help

Here is my setup

Custom role: Regional inherits Basic Role
 
![image](https://user-images.githubusercontent.com/42922833/191123951-fdc4ba85-502d-487c-bbb2-650b2b16b48a.png)

Navigation links - configured Minimum role to access link as Basic.

![image](https://user-images.githubusercontent.com/42922833/191124175-fccf0433-b32f-4794-a8a0-11d78a203031.png)

user setup with app access set to custom role: Regional
![image](https://user-images.githubusercontent.com/42922833/191125168-4599dcf8-ff3c-4a6a-b3f3-dbf0c8066a79.png)

Published app and logged in with user defined at role: Regional 
Navigation links not displayed on Home page 

![image](https://user-images.githubusercontent.com/42922833/191125562-851f7cc5-c228-4839-8365-618f5ef3a180.png)

another issue, with the dev mode preview there is no option to view screen as custom role user 
![image](https://user-images.githubusercontent.com/42922833/191124299-8d35c8b6-12dc-4e71-9f68-0c7dac8a5b66.png)
</div>

insomniacj93:
Same issue here, I wanted to set partitions of the program to be only used by a certain role but also have the lower roles such basic accessable to the higher roles created.

melohagan:
Thank you for raising this issue @applicationuser 

I was able to replicate it. If you set the navigation link role to **Regional** it will work as expected, but I understand that the *Basic* role should work as well because the Regional role is higher.

aptkingston:
This is unfortunately just a limitation of custom roles. We don't really have a proper hierarchy of where custom roles fall, and this is a symptom of that. I can only recommend sticking to the built-in roles to ensure everything works as expected.

JayP718:
Having similar issues are visibility of navigation links using custom roles

mp-strachan:
> This is unfortunately just a limitation of custom roles. We don't really have a proper hierarchy of where custom roles fall, and this is a symptom of that. I can only recommend sticking to the built-in roles to ensure everything works as expected.

[...]

PR #12056, opened 2023-10-13
Custom role navigation links

melohagan:
## Description
The app preview was not correctly reflecting chosen custom user roles. 

Fix to ensure that custom roles in the app preview cannot see content at a higher permission level (above the inherited role).
This matches the behaviour of a published app.

**Note**: There is something funky going on with the UI of the preview navigation links for the custom role screens. Sometimes the link is not highlighted, other times it and another screen role link is highlighted.
This does not affect the actual routing itself, and there is no UI issue for published apps. 
Tried to figure out a fix but it evaded me.

Addresses: 
- https://github.com/Budibase/budibase/issues/7917

## App Export
[Roles test-export-1697205847826.tar.gz](https://github.com/Budibase/budibase/files/12894599/Roles.test-export-1697205847826.tar.gz)

## Screenshots

Screenshot 2023-10-13 at 16 26 15

Screenshot 2023-10-13 at 16 26 28

Screenshot 2023-10-13 at 16 26 45

Screenshot 2023-10-13 at 16 27 03

## Feature branch env
[Feature Branch Link](http://fb-budi-5841-custom-role-navigation-links.fb.qa.budibase.net)


packages/server/src/api/controllers/role.ts
```diff
@@ -143,4 +144,20 @@ export async function accessible(ctx: UserCtx) {
   } else {
     ctx.body = await roles.getUserRoleIdHierarchy(roleId!)
   }
+
+  // If a custom role is provided in the header, filter out higher level roles
+  const roleHeader = ctx.header?.["x-budibase-role"] as string
```
mike12345567, line 149: I think we have an enumeration of headers somewhere - I think in backend-core `src/constants/misc.ts` - should probably move this to types for easy - but if you could use the `PREVIEW_ROLE` from that would be great!

packages/server/src/api/controllers/role.ts
```diff
@@ -3,6 +3,7 @@ import { getUserMetadataParams, InternalTables } from "../../db/utils"
 import { Database, Role, UserCtx, UserRoles } from "@budibase/types"
 import { sdk as sharedSdk } from "@budibase/shared-core"
 import sdk from "../../sdk"
+import { BUILTIN_ROLE_IDS } from "@budibase/backend-core/src/security/roles"
```
mike12345567, line 6: can't pull in references this way - can break builds - can you update this to `import { roles } from @budibase/backend-core`.

mike12345567 (approved):
LGTM - nice tests!

PR #12058, opened 2023-10-13
Fix sort column settings

aptkingston:
## Description
This PR fixes a recent regression that prevented any fields being available to use as the sort column setting for any component.


mike12345567 (approved):
Very nice!

PR #12072, opened 2023-10-16
Prevent the key user being used in rest queries

Ghrehh:
## Description
Users get unexpected behaviour/failures if they attempt to create and consume a binding called "user". This is used internally by budibase on the backend, so anything they define will be overwritten. This change will warn the user that the binding name cannot be used.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7471/prevent-the-key-user-being-used-as-a-binding-in-rest-queries

## Feature branch env
[Feature Branch Link](http://fb-prevent-the-key-user-being-used-in-rest-queries.fb.qa.budibase.net)

### DIFF
--- a/packages/builder/src/components/integration/RestQueryViewer.svelte
+++ b/packages/builder/src/components/integration/RestQueryViewer.svelte
@@ -206,7 +206,7 @@
       throw bindingError
     }
 
-    if (forbiddenBindings.test(query?.fields?.requestBody ?? "")) {
+    if (forbiddenBindings.test(query.fields.requestBody ?? "")) {
       throw bindingError
     }
 
@@ -216,7 +216,7 @@
       }
     })
 
-    Object.values(query?.fields?.headers).forEach(headerValue => {
+    Object.values(query.fields.headers).forEach(headerValue => {
       if (forbiddenBindings.test(headerValue)) {
         throw bindingError
       }
--- a/packages/builder/src/components/integration/RestQueryViewer.svelte
+++ b/packages/builder/src/components/integration/RestQueryViewer.svelte
@@ -196,8 +196,34 @@
     }
   }
 
+  const validateQuery = async () => {
+    const forbiddenBindings = /{{\s?user\s?}}/g
+    const bindingError = new Error("'user' is a protected binding and cannot be used");
+
+    if (forbiddenBindings.test(url)) {
+      throw bindingError;
+    }
+
+    if (forbiddenBindings.test(query?.fields?.requestBody ?? "")) {
+      throw bindingError;
+    }
+
+    Object.values(requestBindings).forEach((bindingValue) => {
+      if (forbiddenBindings.test(bindingValue)) {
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -196,8 +196,36 @@
...
+
+    if (forbiddenBindings.test(url)) {
+      throw bindingError
+    }
+
+    if (forbiddenBindings.test(query.fields.requestBody ?? "")) {
+      throw bindingError
+    }
+
+    Object.values(requestBindings).forEach(bindingValue => {
```
aptkingston, line 213: Can we guarantee `requestBindings` is always defined? Just since otherwise this will throw an error and crash.

packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -196,8 +196,36 @@
...
+      if (forbiddenBindings.test(bindingValue)) {
+        throw bindingError
+      }
+    })
+
+    Object.values(query.fields.headers).forEach(headerValue => {
```
aptkingston, line 219: Same as above - can we guarantee`query.fields.headers` is always defined?

packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -196,8 +196,36 @@
     }
   }
 
+  const validateQuery = async () => {
+    const forbiddenBindings = /{{\s?user\s?}}/g
```
aptkingston, line 200: This will catch `{{ user }}` by itself, but won't catch things like `{{ user.email }}`. Just wanted to check that this was acceptable. Worth checking that something like `{{ user.email }}` doesn't also throw an error if used.

packages/builder/src/components/integration/RestQueryViewer.svelte
Ghrehh: Always defined as far as I can tell

packages/builder/src/components/integration/RestQueryViewer.svelte
Ghrehh: Yup good catch, updated this to handle property access too. It didn't throw any errors but seems improper to leave it possible anyway I guess. A user can still technically subscript `user` I guess, dunno if that's possible in handlebars.

aptkingston (approved):

packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -196,8 +196,36 @@
     }
   }
 
+  const validateQuery = async () => {
+    const forbiddenBindings = /{{\s?user(\.(\w|\$)*\s?|\s?)}}/g
```
melohagan, line 200: Is it possible to unit test this at all?

melohagan (approved):

Issue #12070, opened 2023-10-16
[BUDI-7628] BB freezes after dragging and dropping user column


joebudi:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
budimaster > marketing app > tasks table

**Describe the bug**
BB freezes after scrolling table, then dragging and dropping user column. See video...

**To Reproduce**
1. Scroll down the tasks table within the data section
2. Drag the user column to another location

**Screenshots / video**
https://vimeo.com/874744808/28fc7a8b65?ts=20000&share=copy

**Desktop (please complete the following information):**
Macbook/Arc.

<sub>[BUDI-7628](https://linear.app/budibase/issue/BUDI-7628/bb-freezes-after-dragging-and-dropping-user-column)</sub>

mike12345567:
Appears to be fixed in Budi-cloud, tested in a few different environments.

PR #12076, opened 2023-10-16
User column drag and drop freeze fix

mike12345567:
## Description
Quick fix for drag and drop behaviour of relationship cells, appears empty cells were causing things to break in the re-render.

Appears to affect all relationship column types.

Addresses: https://github.com/Budibase/budibase/issues/12070

## Feature branch env
[Feature Branch Link](http://fb-fix-12070-user-column-dnd-freeze.fb.qa.budibase.net)


aptkingston:
This is a really strange one - the error thrown is a svelte internal error. The cells handle empty values just fine, so I reckon this is some really strange edge case with how svelte is remounting things inside unkeyed `#each` blocks.

I actually dug deep into this one, and the error is actually down to the `Icon` component that displays the `X` to remove a value. Removing this icon fixes this problem.

The change here works almost by accident, because if you move the logic into a reactive statement, something like:
```js
$: foo = Array.isArray(value) ? value : []
...
{#each foo as relationship}
  ...
```
 then the error still happens - so the actually array checking doesn't affect anything, but I think it's the extra reference to `value` is causing svelte to handle things internally differently.

The easy fix here is just to key the `#each` block (something like `{#each value as relationship (relationship._id)`), but I really don't want to do that because remounting every value badge *should* be unnecessary and introduces a big performance hit. The grid deliberately never remounts anything apart from swapping columns, which is why it has fantastic performance with thousands of rows.

aptkingston (approved):

mike12345567:
> After looking at it more I've learned something new about svelte! The grid has always deliberately remounted cells when the column changes, since values between column types are incompatible with each other. Turns out that when a parent component is remounted, child components are potentially _not_ remounted - so all along grid cells have not actually been properly remounting when changing columns.
> 
> There's a larger change there I need to make to the grid to fix that issue, which will in turn properly fix this.
> 
> I'm going to get that sorted now as there may be other consequences, but I'm happy to merge this in the meantime to fix the issue, and I'll revert it when doing that larger change.

Very interesting - yeah it does feel like something deeper is going on, as I was really sure how totally stopping the relationship cell from rendering when it was empty would affect things - this makes a lot more sense!

Issue #12068, opened 2023-10-16
[BUDI-7627] Auto column after selecting no longer pops up with column options.


ConorWebb96:
**Checklist**
- [x] I have searched Budibase discussions and Github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.11.33
  - App Version: 2.11.33

**Describe the bug**
This seems to be a regression. Upgraded from `2.11.16` to `2.11.33` and auto columns no longer allow for selection of `updated at`, `created by` fields anymore. Console error does appear when selecting this, suggests that theres something not being handled correctly.

**To Reproduce**
1. Upgrade to latest -> 2.11.33
2. Create an app
3. Create an internal table with the Budibase datasource.
4. Create a new column using the `Auto columns` option
5. See that no additional select field appears to choose which type of auto column required.

**Expected behavior**
The same behaviour as before the regression, an additional select is shown whenever the auto column is selected.

**Screenshots**
<img width="351" alt="Screenshot 2023-10-16 at 09 04 00" src="https://github.com/Budibase/budibase/assets/126772285/ad0f9681-3cb4-41e9-9e34-7fd829d9af8c">
<img width="1430" alt="Screenshot 2023-10-16 at 09 04 20" src="https://github.com/Budibase/budibase/assets/126772285/0fb2f3ff-11bf-4187-b466-3db259a3f45d">
<img width="455" alt="Screenshot 2023-10-16 at 09 04 50" src="https://github.com/Budibase/budibase/assets/126772285/46f7a4d3-191e-4248-9900-bfee8bad72e9">

**Desktop (please complete the following information):**
Happens on all browsers

**Related Discussion**
https://github.com/Budibase/budibase/discussions/12067

<sub>[BUDI-7627](https://linear.app/budibase/issue/BUDI-7627/auto-column-after-selecting-no-longer-pops-up-with-column-options)</sub>

mike12345567:
Fixed in latest.

PR #12078, opened 2023-10-16
Fix autocolumn addition/editing in data UI

mike12345567:
## Description
Fixes for adding autocolumns, it was not possible to select auto column type and fixing icon, as well as display of column when editing it - this corrects a few issues in this area as auto columns did not have a correct fieldId and also did not have a valid field definition.

Adding:
image
image

Editing:
image

Addresses: #12068

## Feature branch env
[Feature Branch Link](http://fb-fix-12068-autocolumn-data-ui-fixes.fb.qa.budibase.net)


samwho (approved):
Took me a little while to figure out why this change fixes the problem, but I see it now. Creating a field ID out of `type + subtype` for auto columns results in lots of variants of auto that don't appear in the field type definition object.

Good find. This component is a little hard to follow as-is, probably worth refactoring at some point. 🙂

PR #12082, opened 2023-10-17
Remove sentry references

adrinr:
## Description
Remove all sentry references, as it is not used anymore. It was replaced from Posthog, so we can remove all the references

## Feature branch env
[Feature Branch Link](http://fb-chore-remove-sentry-references.fb.qa.budibase.net)


mike12345567 (approved):
LGTM - will just wait for confirmation from @aptkingston and @shogunpurple they are happy to get this merged as well!

Just to add to the reasoning for this - we found some errors which are being thrown on startup due to Sentry not being available, since it is no longer used we can simply get rid of the problem rather than fixing it.

shogunpurple (approved):
LGTM - yeah we get everything we need now generally from datadog. Sentry is a bit too noisy

adrinr:
> LGTM - yeah we get everything we need now generally from datadog. Sentry is a bit too noisy

Just confirming that is not used for the frontend either @aptkingston

packages/builder/src/analytics/index.js
```diff
@@ -23,12 +21,9 @@ class AnalyticsHub {
 
   identify(id) {
     posthog.identify(id)
-    sentry.identify(id)
   }
 
-  captureException(err) {
-    sentry.captureException(err)
-  }
+  captureException(err) {}
```
adrinr: @aptkingston should this not be using posthog?

packages/server/src/db/linkedRows/LinkController.ts
```diff
@@ -434,8 +433,7 @@ class LinkController {
           await this._db.put(linkedTable)
         }
       } catch (err) {
-        /* istanbul ignore next */
-        Sentry.captureException(err)
+        // Ignored error
```
adrinr, line 436: @mike12345567 @shogunpurple  should this be using posthog, or is done automatically?

packages/builder/src/analytics/index.js
```diff
@@ -23,12 +21,9 @@ class AnalyticsHub {
 
   identify(id) {
     posthog.identify(id)
-    sentry.identify(id)
   }
 
-  captureException(err) {
-    sentry.captureException(err)
-  }
+  captureException(_err) {}
```
adrinr, line 26: @aptkingston should this not be using posthog?

packages/server/src/db/linkedRows/LinkController.ts
mike12345567: To catch errors with Posthog, we just need to console.error it, or use the `import { logging } from "@budibase/backend-core"` and `logging.logWarn` to log a message + the err object easily in a way that will be detected by Posthog.

packages/builder/src/analytics/index.js
mike12345567: Posthog isn't really an option in the frontend right now, we don't have the ability to do this. Really we need a new way for detecting this stuff, but any errors that go to Sentry are ignored today, so the above isn't actually a change.

PR #12091, opened 2023-10-17
Fix datasource deletion (when selected) causing builder crash

mike12345567:
## Description
Quick fix for datasource deletion, due to the datasource defaulting to an empty object it appeared like a datasource had been set and the UI would attempt to render with this empty state. When not selected move off of the selected datasource page instead and handle the datasource not being selected.

Addresses: https://linear.app/budibase/issue/BUDI-7630/error-in-console-when-deleting-a-datasource-navigation-then-breaks

## Feature branch env
[Feature Branch Link](http://fb-fix-budi-7630-fix-datasource-deletion.fb.qa.budibase.net)

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte
@@ -12,16 +12,25 @@
   import PromptQueryModal from "./_components/PromptQueryModal.svelte"
   import SettingsPanel from "./_components/panels/Settings.svelte"
   import { helpers } from "@budibase/shared-core"
+  import { goto } from "@roxi/routify"
 
   let selectedPanel = null
   let panelOptions = []
 
   // datasources.selected can return null temporarily on datasource deletion
-  $: datasource = $datasources.selected || {}
+  $: datasource = $datasources.selected
+  $: {
+    if (!datasource) {
+      $goto("./datasource")
+    }
+  }
 
   $: getOptions(datasource)
 
   const getOptions = datasource => {
+    if (!datasource) {
+      return
+    }
     if (datasource.plus) {
       // Google Sheets' integration definition specifies `relationships: false` as it doesn't support relationships like other plus datasources
       panelOptions =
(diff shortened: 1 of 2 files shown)


packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte
```diff
@@ -12,16 +12,25 @@
...
   let panelOptions = []
 
   // datasources.selected can return null temporarily on datasource deletion
-  $: datasource = $datasources.selected || {}
+  $: datasource = $datasources.selected
+  $: {
+    if (!datasource) {
+      $goto("./datasource")
+    }
+  }
```
Comment (samwho) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte:26]: Ooh I didn't know you could do this! I assumed reactive statements had to be variable assignments. Cool!

samwho (approved):
I pulled this down and gave it a try and it fixes the problem I found. The code looks reasonable to me though I will say I'm still quite fresh on the frontend code.

shogunpurple (approved):
Seems fixed on the feature branch!

aptkingston (commented):
The even cleaner way to fix this is simply wrap a guard around the slot inside `[datasourceId]/_layout.svelte` which ensures that we won't ever render a datasource details route without having one selected in state.

A quick test locally confirms that this does fix it (and is more in line with what we do elsewhere):
```js
...
  {#if $datasources.selected}
    <slot />
  {/if}
...
```

We can also remove the fallback code of
```js
  // datasources.selected can return null temporarily on datasource deletion
  $: datasource = $datasources.selected || {}
```
since that isn't needed either with this change.

packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte
Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/index.svelte:26]: You can indeed! You can have any arbitrary code inside a reactive statement like this, and svelte re-run it when any of the variables used inside the statement change. What svelte determines a "change" worth invalidating reactive statements is either a reassignment (if a normal JS variable) or a store value change (if using a svelte store).

aptkingston (approved):
LGTM!

PR #12094, opened 2023-10-18
Refactor grid column handling

aptkingston:
## Description
This PR refactors how grids render columns, removing "virtual" horizontal scroll and instead rendering all columns. This provides one major advantage, which is that we never need to fully remount any cells. There was an issue where svelte was memoising props being passed to cells and resulted in crashes when reordering relationship columns. Fully remounting cells cripples performance, so it's actually far more performant to render all columns.

Performance has still been optimised by taking advantage of the shiny new CSS style `content-visibility: hidden` which is used so that all columns offscreen are not rendered but are still part of the DOM and stay updated.

This is a follow up from the convo in https://github.com/Budibase/budibase/pull/12076 and fixes the core underlying issue.

Horizontal scrolling performance is even faster and smoother with these changes:

https://github.com/Budibase/budibase/assets/9075550/056cf0fa-3d55-49e1-98fa-56f7429ec351

### Other misc changes
It's now much easier to delete columns. You can click on the name of the column in the confirmation prompt to autofill the value into the confirmation input:

https://github.com/Budibase/budibase/assets/9075550/b9865df0-e8e0-4bf3-848f-8e4fdab99b10

There were multiple issues with the error displayed when you have duplicate auto columns, so I've fixed those:
image

Addresses:
- https://github.com/Budibase/budibase/issues/12070

## Feature branch env
[Feature Branch Link](http://fb-refactor-grid-columns.fb.qa.budibase.net)


Ghrehh (approved):
LGTM

MihailHadzhiev2022:
LGTM

PR #12098, opened 2023-10-18
CouchDB extend logout timeout

mike12345567:
## Description
Extending Fauxton timeout from 600s -> 7200s. Allow for 2 hours before logout occurs.

## Feature branch env
[Feature Branch Link](http://fb-fix-couchdb-fauxton-timeout.fb.qa.budibase.net)

### DIFF
--- a/hosting/couchdb/couch/local.ini
+++ b/hosting/couchdb/couch/local.ini
@@ -3,3 +3,6 @@
 [couchdb]
 database_dir = DATA_DIR/couch/dbs
 view_index_dir = DATA_DIR/couch/views
+
+[chttpd_auth]
+timeout = 7200 ; 2 hours in seconds


samwho (approved):
❤️

hosting/couchdb/couch/local.ini
```diff
@@ -3,3 +3,6 @@
 [couchdb]
 database_dir = DATA_DIR/couch/dbs
 view_index_dir = DATA_DIR/couch/views
+
+[chttpd_auth]
+timeout = 7200 ; 2 hours in seconds
```
adrinr, line 8: Could this not be longer? It's only local, so there should not be any security issue to have it as long as possible

mike12345567: This isn't only local, the same image is used in self hosted deployments as well - I wanted to extend it in a way that is useful for users (5 minutes is too short for any real usage IMO) and for our own.

shogunpurple (approved):

PR #12100, opened 2023-10-18
Improving typing of external row API

mike12345567:
## Description
Improving the typing around the `ExternalRequest` object, which has implications throughout the row API and SDK, cleaning up where possible based on it.

The main change here is making sure that `Operation.READ` has a specific response (as it already did) and all other operations have their own specific response (as they already did). The types just did not allow for this knowledge to be fully transferred through the layers, requiring some gymnastics when using the class.

## Feature branch env
[Feature Branch Link](http://fb-fix-improve-external-request-typing.fb.qa.budibase.net)


samwho (approved):
Much clearer, thanks for sorting this Michael! 🙏

PR #12101, opened 2023-10-18
Added description field to the Formblock

deanhannigan:
## Description

Added a new Formblock option for a conditional paragraph called `Description`. 

- The paragraph will display between the header and the form fields. It will provide an area where users can provide a bit of context for the form.
- If no value is specified, the component will not be displayed

Addresses: 
- BUDI-7536

## Screenshots

Screenshot 2023-10-18 at 11 12 13

## Feature branch env
[Feature Branch Link](http://fb-feature-form-block-description.fb.qa.budibase.net)


aptkingston (approved):
LGTM!

PR #12102, opened 2023-10-18
Automatic user relationship to user column migration

samwho:
## Description

Feature branch for https://linear.app/budibase/issue/BUDI-7607/migrate-user-relationship-columns-to-the-new-user-column-type.

Addresses: 
- 

## Documentation


packages/server/src/sdk/app/links/links.ts
```diff
@@ -0,0 +1,32 @@
...
+import { LinkDocument, LinkDocumentValue } from "@budibase/types"
+import { ViewName, getQueryIndex } from "../../../../src/db/utils"
+
+export async function fetch(tableId: string): Promise<LinkDocumentValue[]> {
+  if (!isTableId(tableId)) {
+    throw new Error(`Invalid tableId: ${tableId}`)
+  }
+
+  const db = context.getAppDB()
+  const params: any = { startkey: [tableId], endkey: [tableId, {}] }
```
mike12345567, line 12: This should be typable with the `DatabaseQueryOpts` type

packages/server/src/sdk/app/links/links.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export async function fetchWithDocument(
+  tableId: string
+): Promise<LinkDocument[]> {
+  if (!isTableId(tableId)) {
+    throw new Error(`Invalid tableId: ${tableId}`)
+  }
+
+  const db = context.getAppDB()
+  const params: any = {
```
mike12345567, line 25: Same as above

packages/server/src/sdk/app/links/links.ts
samwho: I did initially try that but `startkey` and `endkey` are both expected to be strings, not arrays. I'm uncertain what their string values should be, or how this works in the place I copied it from.

packages/server/src/sdk/app/rows/search.ts
```diff
@@ -49,10 +49,14 @@ export async function fetch(tableId: string): Promise<Row[]> {
...
 
+export async function fetchRaw(tableId: string): Promise<Row[]> {
+  return pickApi(tableId).fetchRaw(tableId)
+}
+
 export async function fetchView(
   tableId: string,
   viewName: string,
   params: ViewParams
-): Promise<Row[]> {
```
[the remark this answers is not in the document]
samwho: Accidental 🙈 good catch.

packages/server/src/sdk/app/links/links.ts
mike12345567: Ah - I would update them in that case - *usually* they are strings, but realistically it could be a number, or an array of string or number type.

packages/server/src/sdk/app/tables/migration.ts
```diff
@@ -0,0 +1,117 @@
...
+  delete table.schema[oldColumn.name]
+  table = await sdk.tables.saveTable(table)
+  await updateLinks({ eventType: EventType.TABLE_UPDATED, table, oldTable })
+}
+
+interface ColumnMigrator {
+  doMigration(): Promise<void>
+}
+
+function getColumnMigrator(
```
mike12345567, line 39: I like this - very nice implementation!

mike12345567 (commented):
Looks great overall - just wanted to comment to say that really like the approach to the migration here - the column migrator is really nice way of handling this!

mike12345567 (approved):
LGTM - massive improvement to typing included here!

PR #12106, opened 2023-10-18
Remove npm dependency from apps and worker docker images

adrinr:
## Description
This PR is creating a npm-less image for worker and apps. It uses the same approach as https://github.com/Budibase/budibase/pull/11954.
These image are not going to be used in production at the moment, they will be used on the QA environment until we are happy that they are good to go for production.
When the old images are removed, we will be able to ship the new release pipeline that will be npm independant.


mike12345567 (approved):
LGTM!

PR #12110, opened 2023-10-18
Major table controller/SDK refactor

mike12345567:
## Description
TODO: full explanation.

This is an attempt to get a lot of the very useful logic of the table controllers out into the SDK so that it can be used more easily to run through the full process of updating or deleting a table.

### DIFF
--- a/packages/server/src/sdk/app/tables/external/index.ts
+++ b/packages/server/src/sdk/app/tables/external/index.ts
@@ -41,9 +41,10 @@ export async function save(
     ...update,
   }
 
+  const tableId = opts?.tableId || update._id
   let oldTable: Table | undefined
-  if (opts?.tableId) {
-    oldTable = await getTable(opts.tableId)
+  if (tableId) {
+    oldTable = await getTable(tableId)
   }
 
   if (hasTypeChanged(tableToSave, oldTable)) {
@@ -114,6 +115,11 @@ export async function save(
         relatedTable,
         relationType
       )
+      if (fkTable.schema[foreignKey] != null) {
+        throw new Error(
+          `Unable to generate foreign key - column ${foreignKey} already in use.`
+        )
+      }
       fkTable.schema[foreignKey] = foreignKeyStructure(foreignKey)
       if (fkTable.constrained == null) {
         fkTable.constrained = []
@@ -132,7 +138,7 @@ export async function save(
 
   cleanupRelationships(tableToSave, tables, oldTable)
 
-  const operation = oldTable ? Operation.UPDATE_TABLE : Operation.CREATE_TABLE
+  const operation = tableId ? Operation.UPDATE_TABLE : Operation.CREATE_TABLE
   await makeTableRequest(
     datasource,
     operation,
--- a/packages/server/src/sdk/app/tables/external/index.ts
+++ b/packages/server/src/sdk/app/tables/external/index.ts
@@ -7,25 +7,29 @@ import {
   ViewV2,
 } from "@budibase/types"
 import { context } from "@budibase/backend-core"
-import { buildExternalTableId } from "../../../integrations/utils"
-import sdk from "../../index"
+import { buildExternalTableId } from "../../../../integrations/utils"
 import {
   foreignKeyStructure,
   hasTypeChanged,
   setStaticSchemas,
-} from "../../../api/controllers/table/utils"
+} from "../../../../api/controllers/table/utils"
 import { cloneDeep } from "lodash/fp"
-import { FieldTypes } from "../../../constants"
-import { makeTableRequest } from "../../../api/controllers/table/ExternalRequest"
+import { FieldTypes } from "../../../../constants"
+import { makeTableRequest } from "../../../../api/controllers/table/ExternalRequest"
 import {
   isRelationshipSetup,
   cleanupRelationships,
   generateLinkSchema,
   generateManyLinkSchema,
   generateRelatedSchema,
-} from "./externalUtils"
+} from "./utils"
 
-export async function externalSave(
+import { getTable } from "../getters"
+import { populateExternalTableSchemas } from "../validation"
+import datasourceSdk from "../../datasources"
+import * as viewSdk from "../../views"
+
+export async function save(
   datasourceId: string,
   update: Table,
   opts?: { tableId?: string; renaming?: RenameColumn }
@@ -39,7 +43,7 @@ export async function externalSave(
 
   let oldTable: Table | undefined
   if (opts?.tableId) {
-    oldTable = await sdk.tables.getTable(opts.tableId)
+    oldTable = await getTable(opts.tableId)
   }
 
   if (hasTypeChanged(tableToSave, oldTable)) {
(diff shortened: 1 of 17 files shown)


packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -0,0 +1,128 @@
...
+  const internalTables = await db.allDocs(
+    getTableParams(null, {
+      include_docs: true,
+    })
+  )
+  return internalTables.rows.map((tableDoc: any) => ({
+    ...tableDoc.doc,
+    type: "internal",
+    sourceId: tableDoc.doc.sourceId || BudibaseInternalDB._id,
+  }))
```
samwho, line 43: Should this be `return processInternalTables(internalTables)`?

packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export async function getExternalTable(
+  datasourceId: string,
+  tableName: string
+): Promise<Table> {
+  const entities = await getExternalTablesInDatasource(datasourceId)
+  return entities[tableName]
+}
+
+export async function getTable(tableId: any): Promise<Table> {
```
samwho, line 66: Should `tableId` be a `string` here?

packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -0,0 +1,128 @@
...
+    return db.get(tableId)
+  }
+}
+
+export async function getAllTables() {
+  const [internal, external] = await Promise.all([
+    getAllInternalTables(),
+    getAllExternalTables(),
+  ])
+  return [...internal, external]
```
samwho, line 83: `return [...internal, ...external]`?

packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export async function getExternalTablesInDatasource(
+  datasourceId: string
+): Promise<Record<string, Table>> {
+  const datasource = await datasources.get(datasourceId, { enriched: true })
+  if (!datasource || !datasource.entities) {
+    throw "Datasource is not configured fully."
```
samwho, line 91: When would this happen in practice, and what would you be expected to do about it?

packages/server/src/sdk/app/tables/internal/index.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  // if the table obj had an _id then it will have been retrieved
+  let oldTable: Table | undefined
+  if (opts?.tableId) {
+    oldTable = await getTable(opts.tableId)
+  }
+
+  // check all types are correct
+  if (hasTypeChanged(table, oldTable)) {
+    throw new Error("A column type has changed.")
```
samwho, line 47: Would be nice in a follow-up PR to expand on this error. Looking at the implementation in `hasTypeChanged`, it might be worth throwing the error from without `hasTypeChanged` and make its name `checkColumnTypes` or something.

packages/server/src/sdk/app/tables/internal/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  // make sure that types don't change of a column, have to remove
+  // the column if you want to change the type
+  if (oldTable && oldTable.schema) {
+    for (const propKey of Object.keys(table.schema)) {
+      let oldColumn = oldTable.schema[propKey]
+      if (oldColumn && oldColumn.type === FieldTypes.INTERNAL) {
+        oldTable.schema[propKey].type = FieldTypes.AUTO
+      }
+    }
+  }
```
samwho, line 70: The comment doesn't seem to line up with what this code is doing. I also can't quite understand why this code needs to be here, what's the history of it?

packages/server/src/sdk/app/tables/internal/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      eventType: oldTable ? EventType.TABLE_UPDATED : EventType.TABLE_SAVE,
+      table: table,
+      oldTable: oldTable,
+    })
+    if (linkResp != null && linkResp._rev) {
+      table._rev = linkResp._rev
+    }
+  } catch (err) {
+    throw new Error(err as string)
+  }
```
samwho, line 118: Could we get rid of this try-catch?

packages/server/src/sdk/app/tables/internal/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  const tableId = table._id!
+
+  // Delete all rows for that table
+  const rowsData = await db.allDocs(
+    getRowParams(tableId, null, {
+      include_docs: true,
+    })
+  )
+  await db.bulkDocs(
+    rowsData.rows.map((row: any) => ({ ...row.doc, _deleted: true }))
```
samwho, line 152: I've seen writes that set `_deleted: true` before, is this a soft delete or is this how you signal to Couch to do a hard delete?

samwho (changes requested):
A scattering of mostly questions. 🙂

packages/server/src/sdk/app/tables/internal/index.ts
mike12345567: This is how you do a bulk delete, you can either call a specific `DELETE` endpoint for a single document, or you can write to the bulk documents API, with the documents tagged with a `_deleted` field.

packages/server/src/sdk/app/tables/internal/index.ts
mike12345567: Agreed - theres a lot of improved validation and error messaging we could do around the table API.

packages/server/src/sdk/app/tables/getters.ts
mike12345567: This is very unlikely to happen, but it is likely there to protect against some sort of very old app version, where a datasource has been mal-configured. This shouldn't happen in modern versions, but apps can be imported from very old versions and datasources could be in a bad state - I don't think I've ever seen this error occur, but realistically the user would need to delete and recreate their datasource if it were to happen to get into such a bad state.

packages/server/src/sdk/app/tables/getters.ts
mike12345567: ah good catch!

packages/server/src/sdk/app/tables/external/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+          relatedSchema.fieldName === foreignKey
+        ) {
+          delete relatedTable.schema[relatedKey]
+        }
+      }
+    }
+  }
+}
+
+export function otherRelationshipType(type?: string) {
```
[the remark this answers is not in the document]
mike12345567: This is simply due to the poor typing the table used to have, but that is now resolved!

packages/server/src/sdk/app/tables/getters.ts
mike12345567: Good catch!

packages/server/src/sdk/app/tables/external/index.ts
```diff
@@ -0,0 +1,190 @@
...
+        extraTablesToUpdate.push(fkTable)
+      }
+    }
+    generateRelatedSchema(schema, relatedTable, tableToSave, relatedColumnName)
+    schema.main = true
+  }
+
+  cleanupRelationships(tableToSave, tables, oldTable)
+
+  const operation = oldTable ? Operation.UPDATE_TABLE : Operation.CREATE_TABLE
```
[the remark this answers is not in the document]
mike12345567: I've updated this to just depend on the table ID - which is either extracted by the passed in options (if for example the table ID is known from a URL param) or from the `_id` property in the properties provided for updating. We need to have the ID for querying the database to see if it exists, if no ID is provided then it simply will be a new table, it won't error, it just won't be an update.

samwho (approved):
Thanks for addressing my questions. I'm happy with this now, though there are some type-related CI failures still. They don't look too complex.

PR #12113, opened 2023-10-19
Formblock styles/settings tab reflow

deanhannigan:
## Description

The Formblock `Align labels` and `Size` settings have been moved into the `Styles` tab in the component settings panel.

- Added a `tag` attribute to the client component manifest.
- The 2 settings above have both had their `tag` value set to `style` and been shifted from 'Fields' into the root.
- Added a `ComponentSettingSection` to top of the `Styles` panel. This section behaves in exactly the same way as the `Settings` section but it only displays settings/sections with a `tag` value of `style`

Addresses: 
- BUDI-7533

## Screenshots

Screenshot 2023-10-19 at 11 57 08

## Feature branch env
[Feature Branch Link](http://fb-feature-formblock-styles-tab-reflow.fb.qa.budibase.net)


packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -17,13 +17,23 @@
...
 
-  $: sections = getSections(componentInstance, componentDefinition, isScreen)
+  $: sections = getSections(
+    componentInstance,
+    componentDefinition,
+    isScreen,
+    tag
+  )
 
   const getSections = (instance, definition, isScreen) => {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:29]: Although I think this will work since the reactive statement will re-run whenever `tag` changes, I think we should still have `tag` listed as a param here.

aptkingston (approved):
LGTM! Just one tiny change suggested.

aptkingston:
The other thing is that the title "General" looks like its styled differently than the other section headings - it looks white in the screenshot. We just need to make sure it matches the other subheadings.

deanhannigan:
> The other thing is that the title "General" looks like its styled differently than the other section headings - it looks white in the screenshot. We just need to make sure it matches the other subheadings.

Yep, that's an odd one. I think I must have had my mouse over the section header when I took the screenshot. In the builder they are completely identical 😄

PR #12117, opened 2023-10-19
Many-to-many user relationship -> user column migration

samwho:
## Description

Implements the backend code required to migrate many-to-many user relationships.


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "2.11.38",
+  "version": "2.11.39",
```
mike12345567, line 2: not sure why this changed - you might need to pull master into the base branch.

packages/server/src/tests/utilities/api/index.ts
```diff
@@ -27,5 +27,6 @@ export default class API {
     this.datasource = new DatasourceAPI(config)
     this.screen = new ScreenAPI(config)
     this.application = new ApplicationAPI(config)
+    this.global = new GlobalAPI(config)
```
mike12345567, line 30: Just to check - is the `GlobalAPI` in the base branch?

mike12345567 (approved):
Looks good to me - very clear! Just a couple of type errors to check !

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -315,10 +315,7 @@ class TestConfiguration {
       roles,
     })
     await cache.user.invalidateUser(globalId)
-    return {
-      ...resp,
-      globalId,
-    }
+    return resp
```
mike12345567, line 318: Why doesn't think include the `globalId` anymore?

samwho: It wasn't used by anything (tests seem to pass without it there), and it was preventing the `User` type from being the correct return type of this function.

mike12345567: Ah perfect in that case!

mike12345567 (approved):
LGTM!

Issue #11788, opened 2023-09-15
[BUDI-7496] Update table to use grid "order" metadata


Ph1com:
**Checklist**
- [ ] I have searched budibase discussions and github issues to check if my issue already exists - yes, can't find it

**Hosting**

  - Method: <method> I am using Budibase in the cloud
  - Budibase Version: <version>  <-2.10.6 ->
  - App Version: <version>  - Cloud
  - Tenant ID: <tenantId> 

**Describe the bug**
I joined Budibase yesterday and I've just created my first Budibase app and screen using the Budibase database. I've managed to create an inventory app, which scans in barcodes and allows me to take and store photos. All looks good, except that when I try to rearrange the column fields within the data table, it does not reflect these changes on the screen within the design tab. I have republished, but it is still the same. I can change the position of the fields in the side panel, but not the row view.

**To Reproduce**
1. Go to "Data" tab, and choose source table "Inventory"
2. Drag Field labeled "Date Booked In" two positions to the left
3. Clicke "Publish"
4. Go to "Design" tab and choose screen "Inventory"
5. As you can see, the Field labeled "Date Booked In" is not in the same position as per the source table "Inventory"

**Expected behavior**
I expected that when I moved the field columns within the data table, that it would be reflected within the columns shown on the respective screen

**Screenshots**
![Budibase data2 15092023](https://github.com/Budibase/budibase/assets/145127046/58e39189-8425-4779-8e13-9d6d9bb687c9)
![Budibase design screen 15092023](https://github.com/Budibase/budibase/assets/145127046/bf588d0a-7b2c-4dc0-aed5-cd734bc65834)
![Budibase data 15092023](https://github.com/Budibase/budibase/assets/145127046/fdd933a7-7d20-49a7-8088-18493db6938d)
  
**App Export**
[GHL-export-1694794502198.tar.gz](https://github.com/Budibase/budibase/files/12621687/GHL-export-1694794502198.tar.gz)

**Desktop (please complete the following 
 - OS: Win7
 - Browser Firefox
 - Version 115.2.0esr

**Additional context**

<sub>[BUDI-7496](https://linear.app/budibase/issue/BUDI-7496/column-fields-in-data-table-do-not-match-screen-column-fields)</sub>

PR #12123, opened 2023-10-20
Fix table metadata

aptkingston:
## Description
This PR updates tables to be better aware of grid metadata. That's things like column visibility and order, which are configured in grids in the data section.

This PR makes a number of changes to make tables behave more like grids:
- By default, tables will not show hidden columns
- By default, tables will sort columns the same way grids do (using the order from the data section). This includes showing the primary display column as the first column.
- When specifying a custom column configuration for a table, visibility and order are both overridden by this custom config
- Columns widths are specifically *not* used from the grid, as tables render columns differently and cannot be resized. Therefore they will continue to stretch columns by default to use the full width of the table.

This also fixes an issue where custom column config for tables would be completely ignored if your table contained a hidden field.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7496/update-table-to-use-grid-order-metadata
- https://github.com/Budibase/budibase/issues/11788

## Feature branch env
[Feature Branch Link](http://fb-fix-table-metadata.fb.qa.budibase.net)


Ghrehh (approved):
lgtm

deanhannigan:
- Table/Tableblock not displaying any columns by default.
  - Seems to be limited to existing tables without any visibility config specified.
  - If the column visibilty is toggled in the data in the data section, the `All columns` behaviour works as expected.

packages/client/src/components/app/table/Table.svelte
```diff
@@ -55,32 +56,52 @@
...
     }
 
     // Otherwise generate columns
     let columns = []
     let autoColumns = []
     Object.entries(schema).forEach(([field, fieldSchema]) => {
+      if (!fieldSchema.visible) {
+        return
+      }
       if (!fieldSchema?.autocolumn) {
```
deanhannigan, line 76: Would this need to default to true when not set to ensure its visible?

aptkingston: Nice one - I think I actually meant to check against an explicit `false`, which would mean nullish is considered visible.

aptkingston: Pushed that up now - would you mind checking if that now works as expected?

deanhannigan: Reviewed this again, all default visibility checks for the Table/Tableblock working as expected 👍

MihailHadzhiev2022:
LGTM!

PR #12124, opened 2023-10-20
Edit component name from the settings panel title

deanhannigan:
## Description

Make the component name title in the Component settings panel editable.

- `Name` field will no longer appear the component settings panel.
- Click on the title in the settings panel to begin editing. Press `enter` or blur the field to persist the change.
- The name will default to a component type placeholder when left empty.

Addresses: 
- BUDI-7537

## Screenshots
editname

## Feature branch env
[Feature Branch Link](http://fb-feature-component-name-setting-ux-update.fb.qa.budibase.net)


packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsPanel.svelte
```diff
@@ -39,6 +49,20 @@
...
     <Panel {title} icon={componentDefinition?.icon} borderLeft wide>
+      <span class="panel-title-content" slot="panel-title-content">
+        <input
+          class="input"
+          value={title}
+          {title}
+          placeholder={getComponentText(componentInstance)}
+          on:change={e => {
+            onUpdateName(e.target.value)
+            if (e.target === document.activeElement) {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsPanel.svelte:60]: Could you explain what this does? I'm not entirely sure what checking the target (which is surely always the input component) against the document active element would do. If this is to handle an enter keypress we would be better to check if `e.key` is `"Enter"` and then blur the field.

aptkingston (approved):
LGTM! Just one question.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsPanel.svelte
Comment (deanhannigan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsPanel.svelte:60]: As you point out, I think it's just an unnecessary check.  I'll switch it out for a key press event 👍

PR #12126, opened 2023-10-20
Push QA on master commit

adrinr:
## Description
Changing QA process to be commited on master push, instead of waiting for the release to be tagged.
The new images will only be used in QA for now, master is still pushing the old images until we confirm everything works as expected

Depends on: https://github.com/Budibase/budibase-deploys/pull/133


mike12345567 (approved):
LGTM!

PR #12128, opened 2023-10-20
Reduce nx cloud usage

adrinr:
## Description
Don't use nx cloud locally or on CI. We are using nx-affected, so most of the times the cloud cache is not being used, and we can save some costs

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -3,14 +3,16 @@
     "default": {
       "runner": "nx-cloud",
       "options": {
-        "cacheableOperations": ["build", "test", "check:types"],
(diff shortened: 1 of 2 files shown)


nx.json
```diff
@@ -3,14 +3,16 @@
     "default": {
       "runner": "nx-cloud",
       "options": {
-        "cacheableOperations": ["build", "test", "check:types"],
-        "accessToken": "MmM4OGYxNzItMDBlYy00ZmE3LTk4MTYtNmJhYWMyZjBjZTUyfHJlYWQ="
```
adrinr, line 7: Not having the valid access token, it will default to local cache

mike12345567 (approved):

mike12345567:
@shogunpurple - just looping you in - this is disabling NX Cloud - hopefully should reduce bills!

PR #12135, opened 2023-10-20
Parse view v1 schema to fix builder crash.

deanhannigan:
## Description

Navigating to an older view with `groupBy` config would cause the builder to crash.

- The default groupBy schema was causing issues when determining sortable columns.
- `groupBy` schemas are now parsed before being processed and displayed in the DataTable.

## Feature branch env
[Feature Branch Link](http://fb-fix-view1-sortable-column-config.fb.qa.budibase.net)


aptkingston (approved):

aptkingston (changes requested):
Think we actually need to revisit this, because it's only certain types of views with "group by" fail. For example, having a calculation with "stats" will work, while "count" will fail. They all also have different schema, but the solution here is only applicable to "count" (not sure about the other types like min/max/average - it could well be that "stats" is the outlier that's not accounted for).

I think we just need to go through the schema for "group by" views, find any string properties and turn them into `{ foo: { type: schema[foo] }}` which is similar to what is done here, but more generic.

deanhannigan (commented):
Updated the logic so as to not exclude the other calculation types from parsing

aptkingston (approved):
LGTM!

Issue #12088, opened 2023-10-17
[BUDI-7637] Admin/builder user only endpoint


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.11.35
  - App Version: 2.11.35

**Describe the bug**
When a basic user is using the new user type in a create field the select is empty and the application shows a notification "Admin/Builder user only endpoint"

**To Reproduce**
1. Create an application with an internal table and add a user object to that table
2. Create a page (basic access) and add a form block to that page with create type and set the data to use the table created
3. Invite a user with basic access level and ask them to go to that application/pate

**Expected behavior**
The basic user should be able to view and select from the users in the user dropdown

**Screenshots**
![booking1](https://github.com/Budibase/budibase/assets/110813989/3c352c7d-eb29-4916-8211-91ebb02d1e9a)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Edge

<sub>[BUDI-7637](https://linear.app/budibase/issue/BUDI-7637/adminbuilder-user-only-endpoint)</sub>

mike12345567:
Hi @andyburgessmd - this will be fixed in 2.11.41.

PR #12136, opened 2023-10-20
Make user search allow basic users, so that can be used in apps

mike12345567:
## Description
Two fixes here - a quick fix for the builder side panel, making sure it fills up with users correctly (not all, but enough to make it look more pleasant) as well as dropping user search endpoint permissions to allow basic users to access it for user columns.

Just to be clear this isn't really a change in security as today users can be retrieved by any user from the app user table. This just exposes the global users instead, which doesn't contain any different information.

Also just making the builder side panel fill up with users when there are enough - not all users but looks a bit nicer.
image

Addresses: 
- https://github.com/Budibase/budibase/issues/12088
- https://linear.app/budibase/issue/BUDI-7637/adminbuilder-user-only-endpoint

### DIFF
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -206,10 +206,8 @@ export const search = async (ctx: Ctx<SearchUsersRequest>) => {
   }
 
   if (body.paginate === false) {
-    console.log("not paginated")
     await getAppUsers(ctx)
   } else {
-    console.log("paginated")
     const paginated = await userSdk.core.paginatedUsers(body)
     // user hashed password shouldn't ever be returned
     for (let user of paginated.data) {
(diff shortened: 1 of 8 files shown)


packages/worker/src/api/controllers/global/users.ts
```diff
@@ -203,8 +206,10 @@ export const search = async (ctx: Ctx<SearchUsersRequest>) => {
   }
 
   if (body.paginate === false) {
+    console.log("not paginated")
```
adrinr, line 209: This should be cleaned

packages/worker/src/api/routes/global/users.ts
```diff
@@ -72,7 +72,8 @@ router
   )
 
   .get("/api/global/users", auth.builderOrAdmin, controller.fetch)
-  .post("/api/global/users/search", auth.builderOrAdmin, controller.search)
+  // search can be used by any user now, to retrieve users for user column
+  .post("/api/global/users/search", controller.search)
```
adrinr, line 76: Is the authorised middleware added by default?

mike12345567: For the worker - all endpoints require authentication by default - thats why all of the global stuff is in here, while the server doesn't strictly require permissions for everything.

I added a test case to show this as well, gets 403 if not logged in on the search endpoint.

adrinr (approved):

PR #12139, opened 2023-10-20
Package version from env instead of package.json

adrinr:
## Description
Reading package version from environment instead of relying on the package.json for the docker images V2. As we don't have npm repo dependencies anymore, the package.json will always have 0.0.0.
This will only apply to the new images, currently only used in QA

| Before | After |
|--------|--------|
| image | image| 

Related: https://github.com/Budibase/budibase-deploys/pull/134


packages/backend-core/src/environment.ts
```diff
@@ -75,12 +75,12 @@ function getPackageJsonFields(): {
     const content = readFileSync(packageJsonFile!, "utf-8")
     const parsedContent = JSON.parse(content)
     return {
-      VERSION: parsedContent.version,
+      VERSION: process.env.VERSION || parsedContent.version,
```
adrinr, line 78: This will use the env variable version if set

mike12345567 (approved):
LGTM!

PR #12151, opened 2023-10-23
Many-to-one user relationship -> user column migration

samwho:
## Description

I've realised in this PR that many-to-many and many-to-one can use the same code, which led to the realisation that what matters is whether you're migrating to a single user column or a multi user column. So I've renamed the migrators accordingly.


mike12345567 (approved):
Makes a lot of sense - these are the only options for both of them!

PR #12158, opened 2023-10-23
Frontend changes for the user column migration work.

samwho:
This PR hooks up the frontend to the new `POST /api/tables/:table_id/migrate` endpoint. The video demonstrates how the functionality works.

https://github.com/Budibase/budibase/assets/338027/da066897-f83b-4324-8fb7-6368e2024cf5


packages/server/src/integrations/utils.ts
```diff
@@ -87,6 +87,10 @@ export function isExternalTable(tableId: string) {
   return tableId.includes(DocumentType.DATASOURCE)
 }
 
+export function isInternalTable(tableId: string) {
+  return !isExternalTable(tableId)
+}
```
samwho, line 92: Not entirely sure if this is the best implementation of this, but I wanted to avoid writing `if (!isExternalTable(table))` in a conditional because I feel like it's the wrong intent. I wanted to make sure something was an internal table specifically.

packages/server/src/sdk/app/tables/migration.ts
```diff
@@ -105,14 +113,17 @@ function getColumnMigrator(
   throw new BadRequestError(`Unknown migration type`)
 }
 
-class SingleUserColumnMigrator implements ColumnMigrator {
+abstract class UserColumnMigrator implements ColumnMigrator {
```
samwho, line 116: The vast majority of the 2 classes that now inherit from this was identical, so it felt natural to use an abstract class here.

packages/server/src/sdk/app/tables/migration.ts
```diff
@@ -1,53 +1,65 @@
...
-import { isExternalTable } from "../../../../src/integrations/utils"
-import { db as dbCore } from "@budibase/backend-core"
-import { EventType, updateLinks } from "../../../../src/db/linkedRows"
+import { isExternalTable } from "../../../integrations/utils"
+import { EventType, updateLinks } from "../../../db/linkedRows"
 import { cloneDeep } from "lodash"
 
+export interface MigrationResult {
+  tablesUpdated: Table[]
+}
```
samwho, line 22: Created and propagated this type back out of the migrator so that we can fire webhook messages back to the client and ensure the client state is correct. This migration endpoint is unusual in that it affects the current table _and_ another table in the same operation, which the UI doesn't automatically account for.

packages/server/src/tests/utilities/api/table.ts
```diff
@@ -58,8 +58,13 @@ export class TableAPI extends TestAPI {
...
       .set(this.config.defaultHeaders())
-      .expect("Content-Type", /json/)
-      .expect(expectStatus)
+    if (res.status !== expectStatus) {
+      throw new Error(
+        `Expected status ${expectStatus} but got ${
+          res.status
+        } with body ${JSON.stringify(res.body)}`
+      )
+    }
```
samwho, line 67: This made error messages from tests that call the migration endpoint significantly easier to debug. I would like functionality akin to this for all of our supertest expects, but I saw no easy way to do this generally.

packages/server/src/websockets/builder.ts
```diff
@@ -100,11 +102,24 @@ export default class BuilderSocket extends BaseSocket {
...
-      id: table._id,
-      table,
-    })
+  emitTableUpdate(ctx: any, table: Table, options?: EmitOptions) {
+    // This was added to make sure that sourceId is always present when
+    // sending this message to clients. Without this, tables without a
+    // sourceId (e.g. ta_users) won't get correctly updated client-side.
+    if (isInternalTable(table._id!)) {
+      table = processInternalTable(table)
+    }
```
samwho, line 111: Two important changes here:

1. The addition of `EmitOptions` to indicate when we want a websocket update to also be processed by the event originator (necessary to update the state of the user table in the UI).
2. The enriching of the `table` with `processInternalTable`. This adds `sourceId` to the table, without which the UI won't use it to update its state.

aptkingston: Happy to leave this in here, but I think this maybe needs another look. With this enrichment, and a new app, the users table *will not* have a sourceId when fetched from the tables list endpoint (as it does not actually exist in the doc), but then *will* have one when a table change event is broadcast. That means that the original frontend issue of not being able to create a relationship to the users table will still exist, and will only be fixed if the client receives a table change event with this enriched property.

So I think the sourceId issue is a totally separate thing.

@mike12345567 could you confirm if removing the sourceId property from users table in new apps is intended? If so then I'm leaning towards removing this enrichment here, and instead updating the frontend to understand that it's optional.

packages/server/src/websockets/websocket.ts
```diff
@@ -270,10 +278,17 @@ export class BaseSocket {
...
-    this.io.in(room).emit(event, {
-      ...payload,
-      apiSessionId: ctx.headers?.[Header.SESSION_ID],
-    })
+  emitToRoom(
+    ctx: any,
+    room: string | string[],
+    event: string,
+    payload: any,
+    options?: EmitOptions
```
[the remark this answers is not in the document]
aptkingston: Happy to leave it out until there's a use case for it. I'd prefer to reduce cases where broadcasting events back to the originator is required, and instead encourage the originator to make their own state updates where possible. So yea, I'd prefer to keep this new option to only places where it's needed.

packages/frontend-core/src/components/grid/controls/MigrationModal.svelte
```diff
@@ -0,0 +1,69 @@
...
+  import { ValidColumnNameRegex } from "@budibase/shared-core"
+
+  const { API, definition, rows } = getContext("grid")
+
+  export let column
+
+  let newColumnName = `${column.schema.name} migrated`
+  $: error = checkNewColumnName(newColumnName)
+
+  const checkNewColumnName = newColumnName => {
```
aptkingston, line 18: Do we need to check the new column name against the other columns in the table? Just in case you chose a column name that's the same as another column in the table.

aptkingston (commented):
Incredibly impressive for one of your first features @samwho! Fantastic job. Left a few questions but overall this looks great.

packages/server/src/websockets/builder.ts
mike12345567: The `sourceId` was really only designed for external datasources - but internal tables had one thrown on just for consistency - if the frontend expects it to be there for all tables then we can update the typing to make it so that it is non-optional on all tables - right now it is required for external but is optional for internal tables, however we can make that required I think.

@samwho - let me know if you'd like a hand with that!

packages/server/src/websockets/websocket.ts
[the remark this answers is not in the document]
mike12345567: I agree - I think its an edge case that we should handle as and when the backend needs to make this sort of change.

packages/server/src/websockets/builder.ts
samwho: I just tried creating a brand new app and inspecting the return values from some API endpoints and here's what I found:

1. `GET /api/tables` returns a `sourceId` (tacked on, the `sourceId` is not present in Couch) on each table
2. `GET /api/tables/ta_users` does not return a `sourceId`
3. `GET /api/tables/ta_<id>` _does_ return a `sourceId` for another internal table (checking Fauxton, this appears to be because the table document has a `sourceId`)

@mike12345567 I'd appreciate a hand with this, thank you 🙏

packages/server/src/integrations/utils.ts
mike12345567: I think this is fine, I'm a fan of utility functions like this purely for it being easier to read what is going on!

packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -20,11 +20,15 @@ import datasources from "../datasources"
 import sdk from "../../../sdk"
 
 function processInternalTables(docs: AllDocsResponse<Table[]>): Table[] {
```
mike12345567, line 22: I would be tempted to update this function to match the function params of the single `processInternalTable`. The function is only used in two places, and it would be fine for those places to map out the tables first, with a `docs.rows.map(row => row.doc)` then we can update the function param to expect a list of tables instead, `tables: Table[]`.

packages/frontend-core/src/components/grid/controls/MigrationModal.svelte
samwho: Great call, done!

packages/frontend-core/src/components/grid/controls/MigrationModal.svelte
```diff
@@ -0,0 +1,69 @@
...
+    if (column.schema.name === newColumnName) {
+      return "New column name can't be the same as the existing column name."
+    }
+    if (newColumnName.match(ValidColumnNameRegex) === null) {
+      return "Illegal character; must be alpha-numeric."
+    }
+  }
+
+  const migrateUserColumn = async () => {
+    let subtype = "users"
```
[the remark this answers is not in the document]
samwho: I, for whatever reason, made the assumption these were backend-only types. 🙈

packages/server/src/sdk/app/tables/migration.ts
```diff
@@ -1,53 +1,65 @@
...
+  if (newColumn.name in table.schema) {
+    throw new BadRequestError(`Column "${newColumn.name}" already exists`)
+  }
 
   table.schema[newColumn.name] = newColumn
   table = await sdk.tables.saveTable(table)
 
-  await migrator.doMigration()
+  let migrator = getColumnMigrator(table, oldColumn, newColumn)
+  let result: MigrationResult
```
mike12345567, line 37: very minor - in this sort of scenario, we would generally stick to:
```ts
try {
  return (await await sdk.tables.saveTable(table)) as MigrationResult
} catch (e) {
  // If the migration fails then we need to roll back the table schema
  // change.
  delete table.schema[newColumn.name]
  await sdk.tables.saveTable(table)
  throw e
}
```
This avoids the result variable being slightly out of scope and makes the typing a bit easier to understand.

packages/server/src/tests/utilities/api/table.ts
mike12345567: hmm - we might be able to overload the supertest object that the config returns, e.g. add it to what `this.request` does - I'll take a look at this!

mike12345567 (approved):
Amazing work @samwho - fantastic implementation :)

packages/server/src/integrations/utils.ts
samwho: It feels like a small thing but I do think it matters that the intent of the code is clear. I don't want to find non-external tables, I want to find internal tables. That they're functionally equivalent doesn't matter imo.

packages/server/src/sdk/app/tables/migration.ts
samwho: Ahh I think this was leftover from a previous iteration of this function that was trying to do something with `result` after the try-catch. Good spot! 🙏

PR #12163, opened 2023-10-24
Client library fetching through server

mike12345567:
## Description
Moving the client library to be retrieved using the same mechanism as we do in development, through the server instance (same as the builder itself). This should help with the CDN issues that we have been having, although attachments are still currently affected.

Also includes a quick improvement to types as there was still a lot of `any` used throughout the static API, made it a bit harder to interact with when no real types used.


shogunpurple (approved):
LGTM - just make sure we test this one in QA before rolling! Was also thinking we could probably have NGINX even serve this

mike12345567:
> LGTM - just make sure we test this one in QA before rolling! Was also thinking we could probably have NGINX even serve this

that could be a good idea - to be honest we could probably serve both the builder and client library stuff through nginx, lifting a good bit of load off of the server in general.

PR #12172, opened 2023-10-25
[Changed] Reactivate sync quotas migration

jvcalderon:
## Description

Quotas migration is a heavy task; it involves a big number of actions in a long period of time. By changing the migration name we can perform this action incrementally; already migrated accounts won't be processed again.


adrinr:
What is the difference? Why changing the name will allow to run them incrementally?

jvcalderon:
It is because internally it has a cache that relates each tenantId with a migration id. Previous id (sync_quotas_2) has been launched in the past so all the old tenants has that migration assigned, the only way to run it over them is to force it but, by doing this, if process stops, we need to reprocess again. Changing the name, each time that we run the process, already migrated users will be ignored.

adrinr (approved):

Issue #11866, opened 2023-09-23
[BUDI-7548] Automation Step numbers output incorrect when using JS Scripting


JClarke90:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker single image 
  - Budibase Version: 2.10.11 

**Describe the bug**
The automation steps are incorreect when testing, it doesn't seem to affect anything except the JS Scripting output from my testing.

**To Reproduce**

The helper suggests to use steps.4.value, which produces a blank output when called.

![image](https://github.com/Budibase/budibase/assets/12259768/16e4bc41-06ad-4ab7-a159-c66f0dfdea07)
![image](https://github.com/Budibase/budibase/assets/12259768/52b6bbdb-c599-4c19-8aa6-98985d536539)

When changing the input to the correct steps 5 it works correctly

![image](https://github.com/Budibase/budibase/assets/12259768/4a482aa8-624c-4303-b6fa-07866c054750)

This took me a while to figure out as it only seems to affect JS scripting, although I haven't tested every automation task. 
When using something like query rows then you need to use the step number shown on the right column.

![image](https://github.com/Budibase/budibase/assets/12259768/7fc255a7-aaa6-48a1-8c56-2af63f4d649a)
![image](https://github.com/Budibase/budibase/assets/12259768/af596906-fc1d-4d51-94a7-a28af4f8b636)

This just makes things super confusing as step 4 output just doesn't exist at all now.

The main issue with fixing this is that it would likely break a lot of automation tasks, but it should be corrected.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Firefox/Edge
 - Version latest

<sub>[BUDI-7548](https://linear.app/budibase/issue/BUDI-7548/automation-step-numbers-output-incorrect-when-using-js-scripting)</sub>

mike12345567:
dupe: [BUDI-7341](https://linear.app/budibase/issue/BUDI-7341/multiple-loops-in-actions-cause-the-steps-to-become-out-of-sync)

mike12345567:
As of the latest release it is possible to use the automation name in bindings - the bindings are now unique to each step and cannot get into an inconsistent state.

PR #12174, opened 2023-10-25
Automation naming and UX updates

PClmnt:
## Description
This PR adds the ability to name automations as well as some other UX style updates around automations

Naming an automation:

https://github.com/Budibase/budibase/assets/5665926/7dfc7809-aa47-4217-b8e9-a3a8b444d2f3

Error and reset states when adding illegal characters and removing a given name

https://github.com/Budibase/budibase/assets/5665926/21617046-dfba-4c56-84be-b3cce77551d4

There are multiple states around this, the name can also only be alphanumeric which we check for on blur. If the name does not satisfy this requirement, and the user moves away from the input then it is reset to the original step name, either a name or the step type. 

The new name is also shown in the bindings drawer:
image

Some of the styling updates include
- Updating the fields to have proper labels instead to the left of the fields
- Moving the action buttons to the top of the block instead of in the body, and updating tooltips and icons around them
- Updating the undo / redo / run test context menu to be sticky like the data section query view

Other fixes:
- Removes Users table from Create Delete row options: #BUDI-7714
- Automation bindings not working correctly for JS scripting: #11866 

image

https://github.com/Budibase/budibase/assets/5665926/a1c3f514-adc6-4b1c-b17e-e5829b436b4a

## Feature branch env
[Feature Branch Link](http://fb-feat-automation-naming-ux-updates.fb.qa.budibase.net)


packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowChart.svelte
```diff
@@ -20,7 +14,8 @@
 
   let testDataModal
   let confirmDeleteDialog
-
+  let scrolling = false
+  let hasScrolled = false
```
aptkingston, line 18: Looking at the usages of `hasScrolled`, I think it's redundant. We can just swap out any usages of it for `scrolling` and get the same behaviour so we may as well remove it.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -1,27 +1,39 @@
...
+  let validRegex = /^[A-Za-z0-9_\s]+$/
+  let typing = false
 
   const dispatch = createEventDispatcher()
 
+  $: stepNames = $selectedAutomation.definition.stepNames
+  $: automationName = stepNames[block.id] || block.name || ""
+  $: automationNameError = getAutomationNameError(automationName)
+  $: status = updateStatus(testResult, isTrigger)
+  $: isTrigger = isTrigger || block.type === "TRIGGER"
```
aptkingston, line 26: This is a recursive declaration, so I'm not sure how svelte will handle this. I think we'd be better using a new variable name here to distinguish between the prop and this, and it would also help keep things more understandable.

I can see that you just moved this, but still think it's worth addressing since we're using it in other ways now.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -43,10 +55,40 @@
       return { negative: true, message: "Error" }
     }
   }
+
+  const getAutomationNameError = name => {
+    if (name.length > 0) {
+      let invalidRoleName = !validRegex.test(name)
+      if (invalidRoleName) {
+        return "Please enter a role name consisting of only alphanumeric symbols and underscores"
```
aptkingston, line 63: The copy here mentions "role name" but this has nothing to do with roles does it? We could just use "name" instead. Maybe just a wee leftover from copying and pasting!

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -1,27 +1,39 @@
...
   $: {
     if (!testResult) {
       testResult = $automationStore.testResults?.steps?.filter(step =>
         block.id ? step.id === block.id : step.stepId === block.stepId
       )?.[0]
     }
   }
-  $: isTrigger = isTrigger || block.type === "TRIGGER"
-  $: status = updateStatus(testResult, isTrigger)
+  $: loopBlock = $selectedAutomation?.definition.steps.find(
```
aptkingston, line 34: `selectedAutomation` has optional chaining here, but up above we're referencing a nested property without it. I think this should be consistent - it either can be null (in which case we need to update the `stepNames` declaration above) or it can't be null (in which case we don't need optional chaining).

I reckon just try deleting the selected automation and see if it throws an error or not - then we'll know whether this gets rendered one final time when the selected automation is null.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -69,12 +111,32 @@
...
+          placeholder="Enter some text"
+          name="name"
+          autocomplete="off"
+          value={automationName}
+          on:input={e => {
+            automationNameError = getAutomationNameError(e.target.value)
+            automationName = e.target.value
+            if (!automationNameError) {
+              automationNameError = false // Reset the error when input is valid
+            }
```
aptkingston, line 127: This statement doesn't do anything - it's checking if something is falsey, then setting it to false if it is. I think we can just remove this if statement.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -69,12 +111,32 @@
...
-          <Body size="XS">Do this:</Body>
         {/if}
-        <Detail size="S">{block?.name?.toUpperCase() || ""}</Detail>
+        <input
+          placeholder="Enter some text"
+          name="name"
+          autocomplete="off"
+          value={automationName}
+          on:input={e => {
+            automationNameError = getAutomationNameError(e.target.value)
```
aptkingston, line 123: `automationNameError` is automatically derived using the reactive statement above, so we don't need to set it here at all. We can just set `automationName` to the new value, and let the reactive statement run to derive any error.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -69,12 +111,32 @@
...
+            if (!automationNameError) {
+              automationNameError = false // Reset the error when input is valid
+            }
+          }}
+          on:click={startTyping}
+          on:blur={async () => {
+            typing = false
+            if (automationNameError) {
+              automationName = stepNames[block.id]
+              automationNameError = null
```
aptkingston, line 134: We don't need to worry about clearing the error here, as setting `automationName` back to a valid name will automatically clear it for us using the reactive statement above.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
aptkingston, line 123: We also might want to trim the value here, because in `saveName` we depend on checking the length of the name, and we wouldn't want to store a blank name that's just a space for example.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -123,4 +215,47 @@
...
+    font-family: var(--font-sans);
+    color: var(--ink);
+    background-color: transparent;
+    border: none;
+    font-size: var(--spectrum-alias-font-size-default);
+    width: 260px;
+    box-sizing: border-box;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
aptkingston, line 232: It might be nice to add top-left and top-right border radius to the input, to match the actual header, so that it doesn't get squared edges when typing.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -123,4 +215,47 @@
...
+
+  /* Hide arrows for number fields */
+  input::-webkit-outer-spin-button,
+  input::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+
+  .typing {
+    border: 0.5px solid var(--spectrum-global-color-static-blue-500);
```
aptkingston, line 247: We should use 1px here rather than 0.5px to be consistent with the rest of the platform.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -123,4 +215,47 @@
...
+  }
+
+  .typing-error {
+    border: 0.5px solid var(--spectrum-global-color-static-red-500);
```
aptkingston, line 251: Same as above, we should use 1px here.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/FlowItemHeader.svelte
```diff
@@ -123,4 +215,47 @@
...
   }
+
+  .hide-context-actions {
+    display: none;
+  }
+  input {
+    font-family: var(--font-sans);
+    color: var(--ink);
+    background-color: transparent;
+    border: none;
```
aptkingston, line 226: We should use a 1px solid transparent border here. This will mean that whenever we start typing, and show a blue border, that the input doesn't jump because its size changes all of a sudden. It'll just make the experience a little smoother for the user.

aptkingston (commented):
Nice work addressing the feedback!

2 little things I still found:
- When focusing to rename, there's still a jump. Maybe there's still some border differences? Would be great if there was no jump when focusing.
- If you take the default name, like after loading the page, then rename it and use invalid characters on your first attempt, it will revert to the previous value because the current one is invalid. The issue is that when you have no previous value, it will actually insert the string `undefined` as the name, so we should just check if we have a previous value or not, and use the default name if we don't have a previous one to revert to.

PClmnt:
> Nice work addressing the feedback!
> 
> 2 little things I still found:
> 
> * When focusing to rename, there's still a jump. Maybe there's still some border differences? Would be great if there was no jump when focusing.
> * If you take the default name, like after loading the page, then rename it and use invalid characters on your first attempt, it will revert to the previous value because the current one is invalid. The issue is that when you have no previous value, it will actually insert the string `undefined` as the name, so we should just check if we have a previous value or not, and use the default name if we don't have a previous one to revert to.

Sorted those two issues, you were right for the border one, was just missing a transparent border on the outer div, seemed to sort it!

mike12345567:
One issue I found with the add test data modal: 
image

Unsure if this is related - but was an issue I saw when testing, probably worth solving!

mike12345567 (approved):
LGTM! If @aptkingston approves then I'm happy to merge - left one comment about something we should probably solve, appears to be the case on feature branch as well, so isn't must my environment!

PClmnt:
> LGTM! If @aptkingston approves then I'm happy to merge - left one comment about something we should probably solve, appears to be the case on feature branch as well, so isn't must my environment!

Yea fixed that! Just needed to update the modal sizing

aptkingston:
One final issue I found is that you're able to rename multiple steps to the same name. That means that the bindings get confused because the readable to runtime map can't uniquely identify the readable bindings. Here's a quick example where I have 2 steps called `a`:
image

My backend log step doesn't work as expected because both bindings for `{{ a.rows }}` end up being the same value, even though I'm trying to get the outputs from different steps.

Basically I think we just need to enforce step name uniqueness 👍

PClmnt:
> That means that the bindings get confused because the readable to runtime map can't uniquely identify the readable bindings

Sorted, uniqueness is now enforced when adding automation names

PR #12177, opened 2023-10-25
Make sourceId property of table required (for all table types)

mike12345567:
## Description
We have been facing an issue with the user table where the `sourceId` is not always returned with the table, today the backend treats the source ID is required for external tables, but not required for internal tables. Some internal tables will have a `sourceId` of `bb_internal` - but this was not strictly required.

The frontend expects this to be available for every table and has created issues with it being missing. There are a few key changes in this PR which make sure the backend keeps the `sourceId` consistent.

1. The `ExternalTable` type has been removed, and the `sourceId` property of the `Table` has been made required. This required updating a fair number of tests that used the type, but I believe it is correct to represent a table as it should be in this way.
2. After this update there were some small code changes needed to handle this type, the main one was making sure that the base user table schema always includes the `sourceId`.
3. Another small change was needed to the old view API, there was a scenario where if it was unable to fetch a matching table for the view, it returned a mal-formed version of a table, this would have likely lead to weird things happening, so instead just throwing an error if this scenario occurs.


aptkingston:
So just to clarify - with this change, the sourceId property will always be available on all internal tables (including the users table) even if it's not actually part of the couch doc? Old apps seem to have it stored in couch, while new ones don't.

I think the only part of the frontend that actually depends on sourceId for internal tables is the new relationship column UI that was recently refactored, which is why we haven't seen any other issues with it until now. It should work as expected though if we can guarantee that sourceId will always be present on all tables.

samwho (approved):
I don't know if this is all of the places we need to make this change, but what I see in the PR looks good to me. Seems to be some CI failures, though.

mike12345567:
> So just to clarify - with this change, the sourceId property will always be available on all internal tables (including the users table) even if it's not actually part of the couch doc? Old apps seem to have it stored in couch, while new ones don't.
> 
> I think the only part of the frontend that actually depends on sourceId for internal tables is the new relationship column UI that was recently refactored, which is why we haven't seen any other issues with it until now. It should work as expected though if we can guarantee that sourceId will always be present on all tables.

The idea of this PR is to try and make the table returned from the API more consistent, we are now processing all tables before they are returned from the API and as they go into the API to make sure they are saved with the correct typing. This means even if they are now currently correct in CouchDB, when they are returned from the API we will make sure they are consistent and correct.

packages/builder/src/components/backend/DataTable/TableDataTable.svelte
```diff
@@ -46,10 +47,7 @@
     tables.replaceTable(id, e.detail)
 
     // We need to refresh datasources when an external table changes.
-    // Type "external" may exist - sometimes type is "table" and sometimes it
-    // is "external" - it has different meanings in different endpoints.
-    // If we check both these then we hopefully catch all external tables.
-    if (e.detail?.type === "external" || e.detail?.sql) {
+    if (e.detail?.sourceType === DB_TYPE_EXTERNAL || e.detail?.sql) {
```
aptkingston, line 50: We can remove the other part of this statement now (`|| e.detail?.sql)`) since we know that sourceType = external will always catch all external tables. Yay!

aptkingston (approved):
LGTM! Nice work, that's a lot of references to update but definitely worth it. Fantastic to make this more consistent.

packages/server/src/api/controllers/table/index.ts
```diff
@@ -10,27 +10,34 @@ import { events } from "@budibase/backend-core"
...
-import { processInternalTable } from "../../../sdk/app/tables/getters"
 
 function pickApi({ tableId, table }: { tableId?: string; table?: Table }) {
   if (table && !tableId) {
     tableId = table._id
   }
-  if (table && table.type === "external") {
+  if (
+    table?.sourceId &&
+    table.sourceId.includes(DocumentType.DATASOURCE + SEPARATOR)
```
adrinr, line 37: It would be better if we use helpers for this, so we can centralise and test this logic:
`dbUtils.isExternalTable(...)` or similar

packages/server/src/constants/index.ts
```diff
@@ -87,12 +95,10 @@ export const USERS_TABLE_SCHEMA = {
         },
         presence: true,
       },
-      fieldName: "email",
```
adrinr, line 90: How come this `fieldName` are dropped?

packages/server/src/db/defaultData/datasource_bb_default.ts
```diff
@@ -89,9 +90,10 @@ const AUTO_COLUMNS: TableSchema = {
 
 export const DEFAULT_INVENTORY_TABLE_SCHEMA: Table = {
   _id: DEFAULT_INVENTORY_TABLE_ID,
-  type: "internal",
+  type: "table",
```
adrinr, line 93: Will this cause any conflict with the existing apps?

packages/server/src/integrations/mysql.ts
```diff
@@ -317,8 +318,10 @@ class MySQLIntegration extends Sql implements DatasourcePlus {
         }
         if (!tables[tableName]) {
           tables[tableName] = {
+            type: "table",
```
adrinr, line 321: Will we need to migrate all the existing ds?

packages/server/src/constants/index.ts
mike12345567: it really shouldn't there - this wasn't typed previously and the `fieldName` only pertains to relationships. Won't cause any problems - but no need for it to be there.

packages/server/src/db/defaultData/datasource_bb_default.ts
mike12345567: No - I've updated the frontend to no longer use the `type` and the backend doesn't really use it either - they all depend on the `sourceType` instead now. Old apps will have this added onto them as part of enrichment.

packages/server/src/integrations/mysql.ts
mike12345567: This will be handled as part of enrichment of the tables when they are retrieved from the API/saved. Essentially we'll be migrating in place.

I've tested this - we already depended on enrichment for `internal/external` type value - so its not much different what we're doing now, just more consistent.

packages/server/src/db/defaultData/datasource_bb_default.ts
adrinr: So, if nobody is using that, why do we need this field?

mike12345567: We don't massively need it - but other documents have this `type` - like `row`, where it is quite important for the internal DB. I'd like to keep this consistent at least - I hadn't realised it was as inconsistent as it was.

PR #12185, opened 2023-10-26
Replace several instances of `any` with `DatabaseQueryOpts`.

samwho:
This PR replaces `any` with `DatabaseQueryOpts` in a few places it was applicable to do so. The reason we were using `any` was because of a problem in `@budibase/nano` that was fixed in https://github.com/Budibase/couchdb-nano/pull/1. With that in, we can make the typing around this better.

There are instances we can improve in `budibase-pro`, too, which I'll do in a follow-up PR over there.

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -21,7 +21,7 @@
     "test:watch": "jest --watchAll"
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


mike12345567 (approved):
Very nice!

PR #12187, opened 2023-10-26
Fix redirect loop when accessing a group as a global builder but not an admin.

samwho:
## Description

Depends on https://github.com/Budibase/budibase-pro/pull/234.

This fixes an infinite redirect loop if you access the groups page as a global builder without admin permissions.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7681/groups-list-infinite-reload-loop

## Feature branch env
[Feature Branch Link](http://fb-bug-budi-7681-groups-list-infinite-reload-loop.fb.qa.budibase.net)


mike12345567 (approved):
✅

PR #12190, opened 2023-10-26
Fix grids not reflecting changes in related table primary display columns

aptkingston:
## Description
This PR primarily resolves an issue where changing the primary display column in a related table would not be reflected inside the grid. This was because the grid held a cached value for the old schema, and did not know it had changed.

This PR refactors those caches and ensures that they are wiped whenever the grid datasource changes - aka whenever you start using a different table in the data section.

### Other changes
-  Updated grids to support endless scrolling pagination for REST queries which are configured to support pagination. Until now, REST queries with pagination config would not work inside grids and would only ever show a single page of data
- Revert a temporary bandaid for a crash that was happening when reordering relationship columns in grids, now that the larger refactor has been completed in another PR to fix the root cause.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7606/changing-primary-displays-on-relationships-requires-a-page-reload

## Feature branch env
[Feature Branch Link](http://fb-grid-tweaks.fb.qa.budibase.net)


deanhannigan (approved):
LGTM!

PR #12198, opened 2023-10-27
Remove APIDoc comments.

samwho:
## Description

We're planning to favour an OpenAPI schema approach in future and nobody really relies on these APIDoc comments today.


mike12345567 (approved):
Thanks for putting this together!

PR #12200, opened 2023-10-27
Form block field layout

aptkingston:
## Description
This PR allows customisation of fields within a form block. This allows custom layouts, with a mix of 1, 2 or 3 fields per column. 

image

Quick video of changing layout:

https://github.com/Budibase/budibase/assets/9075550/2d96825b-275a-4911-8aa1-e0d505be03c9

Field layout is controlled by a new setting inside the field settings modal:
image
image

The form block will automatically slot fields together, making the best possible use of space.

On mobile, all fields will stretch to take up the full width:
image

This new setting is only available when using a form block. The field components themselves do not expose it, as it will not work with a normal form.

### Removal of left-aligned labels
The one downside of this feature is that it is incompatible with left aligned labels, so that setting is no longer available on form blocks. All form blocks will use labels above the fields.

It's not impossible in future that we could support this, but it would require a significant rewrite of the current field layout structure, as well as reverting all the recent field group component work which now provides layout settings.

Addresses: 
- https://linear.app/budibase/issue/BUDI-7531/form-block-add-layout-options-to-fields

## Feature branch env
[Feature Branch Link](http://fb-form-block-field-layout.fb.qa.budibase.net)


deanhannigan (approved):
LGTM!

shogunpurple (approved):
LGTM

PR #12205, opened 2023-10-27
Fix how attachment URLs are handled

Samuel-Martineau:
## Description
Datasources used to be (as of around three weeks ago) able to provide attachments with remote URLs. This behaviour is necessary for custom Datasource+ plugins (which work since #11868) to handle attachments.

The fix is exceedingly simple: only use the storage bucket's URL if the attachment data doesn't already contain one.

## Documentation

### DIFF
--- a/packages/server/src/utilities/rowProcessor/index.ts
+++ b/packages/server/src/utilities/rowProcessor/index.ts
@@ -241,7 +241,7 @@ export async function outputProcessing<T extends Row[] | Row>(
           continue
         }
         row[property].forEach((attachment: RowAttachment) => {
-          attachment.url = objectStore.getAppFileUrl(attachment.key)
+          attachment.url ??= objectStore.getAppFileUrl(attachment.key)
         })
       }
     } else if (


Samuel-Martineau:
@mike12345567 Is this a voluntary change or an accidental one?

Samuel-Martineau:
Are you certain that there was no change in this area? Because up until I updated my version of Budibase, the system was able to handle attachments in external tables

Samuel-Martineau:
What I noticed is that the current (new?) code ignores the current url attribute on the attachment object and replaces it with the path transformed to a signed S3 url. My idea was to only do that step if there is no current url, in order to have the behaviour that paths can be loaded from S3 and urls from their native system

mike12345567:
> What I noticed is that the current (new?) code ignores the current url attribute on the attachment object and replaces it with the path transformed to a signed S3 url. My idea was to only do that step if there is no current url, in order to have the behaviour that paths can be loaded from S3 and urls from their native system

There were some small changes to align the internal/external row APIs, as we've discussed introducing attachment for external rows, but it would use the same mechanism as our internal DB right now.

Would it be possible for you to give this a test with the internal DB, make sure that attachment columns can still be stored and fetched through the data UI of Budibase? It is possible this won't cause any issues - but it would be good to confirm this first.

Samuel-Martineau:
I just double checked and internal DB attachments are still working, as they don't have an assigned url initially. When that line of code is called, the attachment object is only
```json
{
    "size": 398423,
    "name": "Sans titre.png",
    "extension": "png",
    "key": "app_5111acd9c77843eeaf578a45886670c0/attachments/5cc7e2ad-87b8-4361-b2fb-8e95156d3710.png"
  }
```
The code to generate the signed url if and only if the attachment doesn't already have a url should allow for compatibility for internal and external sources.

I just updated the PR to use nullish coalescing assignment to simplify the code even further.

mike12345567:
> I just double checked and internal DB attachments are still working, as they don't have an assigned url initially. When that line of code is called, the attachment object is only
> 
> ```json
> {
>   "size": 398423,
>   "name": "Sans titre.png",
>   "extension": "png",
>   "key": "app_5111acd9c77843eeaf578a45886670c0/attachments/5cc7e2ad-87b8-4361-b2fb-8e95156d3710.png"
> }
> ```
> 
> The code to generate the signed url if and only if the attachment doesn't already have a url should allow for compatibility for internal and external sources.
> 
> I just updated the PR to use nullish coalescing assignment to simplify the code even further.

Hi @Samuel-Martineau - thanks for looking into this - I think this is safe to merge in that case.

PR #12209, opened 2023-10-29
Grid columns UI

Ghrehh:
## Description

Addresses: 
https://linear.app/budibase/issue/BUDI-7544/grid-block-add-columns-ui-to-the-settings-panel

- Adds a new `labelfield` component that is used by the `FieldConfiguration` component to allow editing just the label of a column.
- Adds some options to the `FieldConfiguration` component that allow us to implement the changes we need to reorder grid fields.
- Replace the existing Grid fields drawer with the `FieldConfiguration`  component.
- Grids always display the primary column of the datasource as a sticky column on the left, the user can't reorder this field, but can still rename the label.

Screenshot 2023-10-29 at 20 55 22

## Feature branch env
[Feature Branch Link](http://fb-grid-columns-ui.fb.qa.budibase.net)


packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -38,20 +43,31 @@
...
+    primaryDisplayColumn = getPrimaryDisplayColumn($currentAsset, datasource)
     options = Object.keys(schema || {})
     sanitisedValue = getValidColumns(convertOldFieldFormat(value), options)
     updateSanitsedFields(sanitisedValue)
     unconfigured = buildUnconfiguredOptions(schema, sanitisedFields)
     fieldList = [...sanitisedFields, ...unconfigured]
       .map(buildPseudoInstance)
       .filter(x => x != null)
+
+    if (resourceId !== previousResourceId) {
```
Ghrehh, line 63: This autoselect all columns when changing datasources which is nicer imo.

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,96 @@
+<script>
```
Ghrehh, line 1: I just duplicated `FieldSetting` to make this, seemed better than a bit of optionally to conditionally display a tooltip imo.

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,96 @@
...
+      on:change
+    >
+      <div slot="header" class="type-icon">
+        <Icon name={componentDef.icon} />
+        <span>{item.field}</span>
+      </div>
+    </EditComponentPopover>
+    <div class="field-label">{readableText}</div>
+  </div>
+  <div
```
Ghrehh, line 52: our tooltip component can't be made to align left, so it kept going off the screen. I just opted to use a html `title` instead.

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,96 @@
...
+    title="The leftmost column is dictated by your datasource's primary column, which can be changed in the data section"
```
Ghrehh, line 53: Copy check please.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -152,6 +172,25 @@
 </script>
 
 <div class="field-configuration">
```
Ghrehh, line 174: Seemed nicer to just have two lists that conditionally hide certain fields rather than injecting a bunch of grid specific logic into the field configuration component, open to suggestions though.

packages/builder/src/components/design/settings/controls/DraggableList/DraggableList.svelte
```diff
@@ -58,15 +65,25 @@
...
+    }
+
+    return update
   }
 
   const handleFinalize = e => {
     updateRowOrder(e)
-    dispatch("change", serialiseUpdate())
+    const foo = serialiseUpdate()
+    console.log(foo)
```
deanhannigan, line 85: An easter egg?

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,96 @@
...
+  import { Icon } from "@budibase/bbui"
+  import { store } from "builderStore"
+  import { runtimeToReadableBinding } from "builderStore/dataBinding"
+  import { isJSBinding } from "@budibase/string-templates"
+
+  export let item
+  export let componentBindings
+  export let bindings
+  export let anchor
+  export let hideToggle = false
```
deanhannigan, line 12: Doesn't seem to be used anywhere

deanhannigan (changes requested):
- Column label bindings don't work
Screenshot 2023-10-31 at 15 13 32

- The previous UX didn't support this and extra work would be required to support it.
  - Safe to remove for now

- Separate the Field Configuration and Column Configuration UX
  - They are separate flows and should be isolated.
  - Create a new component to house the Column config flow.

- Switch the `columns` setting type back to `columns/grid` in the component manifest:
  - `"type": "columns/grid"`
  - A migration would be required to switch from `columns/grid` to `fieldConfiguration`


- Link the existing `columns/grid` to a separate Column Configuration component/ux

packages/builder/src/components/design/settings/controls/DraggableList/DraggableList.svelte
```diff
@@ -39,14 +40,20 @@
...
+    const draggableItems = items
       .map(item => {
         return {
           id: listItemKey ? item[listItemKey] : generate(),
           item,
         }
       })
       .filter(item => item.id)
+
+    if (whitelist) {
```
deanhannigan, line 52: Remove the whitelist behaviour from within the core DraggableList component and replicate the behaviour in a component dedicated to configuring columns.

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
Ghrehh: Would much prefer to use this, but I was having issues with the length of the text + how the bounding rect in the tooltip component is calculated, any advice? 
Screenshot 2023-11-08 at 09 04 46

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
Ghrehh: No longer making changes to this file, but leaving this in seems like a nice upgrade if no one objects 🤷

deanhannigan (changes requested):
Some further updates:
- You don't have to duplicate `DraggableList` and `DragHandle` as the existing elements can be reused entirely within your `GridColumnConfiguration` component.
-  If you follow the pattern from the `ButtonConfiguration` component, you can extract the unique sticky-column behaviour [here](https://github.com/Budibase/budibase/pull/12209/files#diff-abe7ce5290b2369777a408eae489bf260f884df583a8da5cc50055fd94b6cf55R82)  and place it entirely inside `GridColumnConfiguration`. This keeps the UX isolated from the generic components.
- The `FieldSetting` component could also be renamed as the `ColumnSetting` component, to differentiate.

If you want to run through these in more detail, let me know 👍

aptkingston (commented):
I'll do a follow up review of the actual code, but after a quick glance I'd like to agree with what Dean is saying - there are a few duplicated files here which aren't needed, as we can reuse the existing ones.

Did some initial testing of the feature and it works well! Ran into a few things which I think need addressed though:
- Obviously backwards compatibility, but I'm aware that you're already working on that. Had to create a new app to test this since it crashed the builder when using it with existing grids.
- The tooltip for the sticky column could use a real `AbsTooltip` with left alignment. It would be a bit nicer than the title attribute.
- If you set a custom column name, we should use that as the column name in the list. If that's empty or you haven't customised it, then use the real column name like we currently do.
- The popover always shows a text icon at the top, regardless of the type of the column. We should either remove that or render the correct icon for the column type.
- If you reorder columns in the data section and have not touched your grid settings, the actual grid will respect this new order, but the settings list does not. If no custom value is set then we should ensure the order of the columns in the settings UI matches the true order

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
deanhannigan: I'm not sure that this behaviour is necessary as [this line](https://github.com/Budibase/budibase/blob/f0a162c8dc8c793deb72dea26f01e3014beade44/packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte#L83) will set the toggle to active when the resource is reset.

Ghrehh: Not 100% sure, want to say I tested it originally and it didn't seem to behave that way, but at this point I think it's best I just remove it regardless, so I have done so 🧙

packages/client/manifest.json
```diff
@@ -6339,6 +6342,18 @@
...
+      },
+      {
+        "section": true,
+        "name": "Columns",
+        "settings": [
+          {
+            "type": "columns/grid",
+            "key": "columns",
+            "nested": true,
+            "resetOn": "dataSource"
```
deanhannigan, line 6354: Shouldn't this be `resetOn: "table"` as thats this key for the target field? `dataSource` is the type of field

deanhannigan (approved):

packages/builder/src/components/design/settings/controls/FieldConfiguration/PrimaryColumnFieldSetting.svelte
aptkingston: Would just go for "The display column is always shown first". Short and sweet.

packages/builder/src/components/design/settings/controls/GridColumnConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,101 @@
...
+    >
+      <div slot="header" class="type-icon">
+        <Icon name={getIcon()} />
+        <span>{item.field}</span>
+      </div>
+    </EditComponentPopover>
+    <div class="field-label">{item.label || item.field}</div>
+  </div>
+  <div
+    title="The leftmost column is dictated by your datasource's primary column, which can be changed in the data section"
```
aptkingston, line 58: Had replied to your previous comment but that's a bit hidden, so I'll say it again here for visibility.

I would update this to just "The display column is always shown first".

packages/builder/src/components/design/settings/controls/GridColumnConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,101 @@
...
+<div class="list-item-body">
+  <div class="list-item-left">
+    <EditComponentPopover
+      {anchor}
+      componentInstance={item}
+      {parseSettings}
+      on:change
+    >
+      <div slot="header" class="type-icon">
+        <Icon name={getIcon()} />
```
aptkingston, line 51: This won't be reactive, so if the `item` prop changes we wouldn't show the new icon. This is probably fine in this instance because the component probably remounts, but it would be better practice to derive the `icon` in a new reactive statement, then use that here.

packages/builder/src/components/design/settings/controls/GridColumnConfiguration/PrimaryColumnFieldSetting.svelte
```diff
@@ -0,0 +1,101 @@
...
+      .map(setting => {
+        return { ...setting, nested: true }
+      })
+  }
+
+  const getIcon = () => {
+    const component = `@budibase/standard-components/${
+      FieldTypeToComponentMap[item.columnType]
+    }`
+    return store.actions.components.getDefinition(component).icon
```
aptkingston, line 38: This should obviously never be null, but we should still add optional chaining to ensure we wouldn't crash the page if we couldn't find the definition.

packages/builder/src/components/design/settings/controls/GridColumnConfiguration/PrimaryColumnFieldSetting.svelte
Ghrehh: Good catch

Ghrehh:
@aptkingston Thanks for the review, I've addressed your feedback if you could give this another look when you get a chance please.

aptkingston (approved):
Will move this into QA now 👍  think my upstream PR for grid buttons has caused a minor conflict here but should be easy to resolve hopefully.

PR #12212, opened 2023-10-30
Use docker image v2 for production

adrinr:
## Description
This PR changes the current pipelines and scripts to use the new v2 docker images. These ones are npm independent, and they will allow us to unify our pipelines: https://github.com/Budibase/budibase-deploys/pull/140
This change does not have any performance improvement, it just uses the new images.

These new images require building both amd and arm versions. Because of this some changes were required, using `docker buildx`. Pipeline test: https://github.com/Budibase/budibase/actions/runs/6691827922/job/18179827453?pr=12212

### DIFF
--- a/packages/server/Dockerfile
+++ b/packages/server/Dockerfile
@@ -38,7 +38,7 @@ RUN apt update && apt upgrade -y \
 
 COPY package.json .
 COPY dist/yarn.lock .
-RUN yarn install --production=true \
+RUN yarn install --production=true --network-timeout 1000000 \
     # Remove unneeded data from file system to reduce image size
     && yarn cache clean && apt-get remove -y --purge --auto-remove g++ make python \
     && rm -rf /tmp/* /root/.node-gyp /usr/local/lib/node_modules/npm/node_modules/node-gyp
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -18,7 +18,7 @@
     "test": "bash scripts/test.sh",
     "test:memory": "jest --maxWorkers=2 --logHeapUsage --forceExit",
     "test:watch": "jest --watch",
-    "build:docker": "yarn build && docker build . -t app-service --label version=$BUDIBASE_RELEASE_VERSION --build-arg BUDIBASE_VERSION=$BUDIBASE_RELEASE_VERSION",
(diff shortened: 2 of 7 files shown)


packages/server/Dockerfile
```diff
@@ -38,7 +38,7 @@ RUN apt update && apt upgrade -y \
 
 COPY package.json .
 COPY dist/yarn.lock .
-RUN yarn install --production=true \
+RUN yarn install --production=true --network-timeout 1000000 \
```
samwho, line 41: What's the backstory here? 😄

packages/server/package.json
```diff
@@ -18,7 +18,7 @@
     "test": "bash scripts/test.sh",
     "test:memory": "jest --maxWorkers=2 --logHeapUsage --forceExit",
     "test:watch": "jest --watch",
-    "build:docker": "yarn build && docker build . -t app-service --label version=$BUDIBASE_RELEASE_VERSION --build-arg BUDIBASE_VERSION=$BUDIBASE_RELEASE_VERSION",
+    "build:docker": "yarn nx build && docker buildx build ../.. -t app-service --label version=$BUDIBASE_RELEASE_VERSION --build-arg BUDIBASE_VERSION=$BUDIBASE_RELEASE_VERSION -f Dockerfile.v2 --platform linux/amd64,linux/arm64",
```
samwho, line 21: How come the path to the build context has changed here?

packages/server/Dockerfile
adrinr: We are using the default gh runners and we have plenty of packages to download. Sometimes we hit the default yarn timeout, and it needs to retry. Increasing the timeout will ensure we don't have this unnecessary retries

packages/server/package.json
adrinr: The new images (v2) are created to be run from the root of the monorepo, as these require access to other files

samwho: Are you worried at all about the affect this will have on the cacheability of the image?

adrinr: No, we have really strict dockerignore file and the docker image is only copying what is required, instead of using wildcards. So it should not make any difference

samwho (approved):

PR #12214, opened 2023-10-30
Fix pushing images v2

adrinr:
## Description
We attempted using the new images for production just changing the dockerfiles, but as they need to be built via `docker buildx`, the push did not work as expected.
This PR is making use of the caching and `buildx` in order to fix this issue.
Also, removing old scripts (singleimage and develop) not used anymore

The push actions had been mostly copied from the new single pipeline: https://github.com/Budibase/budibase-deploys/pull/140

Testing the push in here: https://github.com/Budibase/budibase/actions/runs/6694095222


.github/workflows/release-master.yml
```diff
@@ -66,14 +67,60 @@ jobs:
...
+          cache-to: type=inline
+        env:
+          IMAGE_NAME: budibase/apps
+          IMAGE_TAG: ${{ inputs.image_tag }}
+          BUDIBASE_VERSION: ${{ steps.currenttag.outputs.version }}
+
+      - name: Build proxy docker
+        uses: docker/build-push-action@v5
+        with:
+          context: .
```
mike12345567, line 115: does the proxy need the platforms as well? It might not given its just the nginx image

mike12345567 (approved):
LGTM!

.github/workflows/release-master.yml
adrinr: Nothing changed in there, so it should be fine. I am testing the actual push images, I will make sure we can pull the image properly

PR #12226, opened 2023-10-31
Removing automation logs from app backups

mike12345567:
## Description
Updating app backup exports to not include automation logs as these bloat the backups. Also includes an improvement to testing, trying to make it a bit easier to check what is happening with backups (still some work needed).

This is an attempt to address: https://linear.app/budibase/issue/BUDI-7586/customers-app-backups-are-failing-in-cloud

Pro side: https://github.com/Budibase/budibase-pro/pull/239


packages/server/src/tests/utilities/api/backup.ts
```diff
@@ -0,0 +1,49 @@
...
+      .expect("Content-Type", /json/)
+      .expect(200)
+    return result.body as CreateAppBackupResponse
+  }
+
+  importBackup = async (
+    appId: string,
+    backupId: string
+  ): Promise<ImportAppBackupResponse> => {
+    tk.freeze(mocks.date.MOCK_DATE_PLUS(1))
```
samwho, line 41: How come this is required?

packages/server/src/sdk/app/backups/exports.ts
```diff
@@ -84,7 +84,11 @@ export async function exportDB(
 }
 
 function defineFilter(excludeRows?: boolean, excludeLogs?: boolean) {
-  const ids = [USER_METDATA_PREFIX, LINK_USER_METADATA_PREFIX]
+  const ids = [
+    USER_METDATA_PREFIX,
+    LINK_USER_METADATA_PREFIX,
+    AUTOMATION_LOG_PREFIX,
```
samwho, line 90: It looks like it's possible for `ids` to contain `AUTOMATION_LOG_PREFIX` twice now if the conditional on line 95 evaluates to true. Is that intended?

packages/server/src/tests/utilities/api/backup.ts
mike12345567: App backups use the time to create backup IDs - in testing we freeze the timestamp which means that all backups happen at exactly the same time - bump it forward a bit to avoid this.

packages/server/src/sdk/app/backups/exports.ts
mike12345567: It is not!

samwho (approved):

packages/server/src/tests/utilities/api/backup.ts
mike12345567: Removed the need for this.

Issue #12216, opened 2023-10-30
Add new user experience


joebudi:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Using budimaster...

When adding a new user (Jamie) I initially set him as a Member - not an admin. This was a mistake so I tried to change it. I initially could not see his name within the users list (portal). Pagination seemed to be broken.  So, I searched for his name and found it. The search input was not behaving as expected - I couldn't retrieve the full list of users by clicking the X icon.

aptkingston:
Pending invites have a dedicated table below the normal users table (in the portal users list) so he wouldn't be listed there until he accepted:
![image](https://github.com/Budibase/budibase/assets/9075550/d0683839-ec37-4663-a566-f6974e56ec99)

Did you initially invite him via the portal or via the side panel in an app?

If via the app side panel, pending invites should be shown at the top of the list:
![image](https://github.com/Budibase/budibase/assets/9075550/f670fdaa-8de1-477e-8b94-fe051a54ea79)

aptkingston:
The portal user list does look broken. I think it was the recent updates around how users can be searched. Pagination is broken and clicking the "x" actually gives a 501 because it tries to search by some illegal property
![image](https://github.com/Budibase/budibase/assets/9075550/59209fe9-1b1e-4fca-b846-a778b1fe526a)

aptkingston:
Have a fix and will get a PR up soon 👍

andz-bb:
duplicate of 7711

PR #12230, opened 2023-10-31
Fix global users endpoint pagination

aptkingston:
## Description
This PR fixes a small issue with the global user search endpoint which meant pagination didn't work. I've also updated the users page in the portal to fix a 501 error being thrown when clearing the search field, and to use the proper lucene query structure.

Also adds some spacing to the buttons on the users list page as there was none.

Addresses: 
- https://github.com/Budibase/budibase/issues/12216
- https://linear.app/budibase/issue/BUDI-7711/user-table-isnt-displaying-all-the-available-users

## App Export

## Screenshots
Pagination working now:
image

### DIFF
--- a/packages/backend-core/src/users/users.ts
+++ b/packages/backend-core/src/users/users.ts
@@ -249,7 +249,8 @@ export const paginatedUsers = async ({
   limit,
 }: SearchUsersRequest = {}) => {
   const db = getGlobalDB()
-  const pageLimit = limit ? limit + 1 : PAGE_LIMIT + 1
+  const pageSize = limit ?? PAGE_LIMIT
+  const pageLimit = pageSize + 1
   // get one extra document, to have the next page
   const opts: DatabaseQueryOpts = {
     include_docs: true,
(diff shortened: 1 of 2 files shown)


packages/backend-core/src/users/users.ts
```diff
@@ -276,7 +277,7 @@ export const paginatedUsers = async ({
     const response = await db.allDocs(getGlobalUserParams(null, opts))
     userList = response.rows.map((row: any) => row.doc)
   }
-  return pagination(userList, pageLimit, {
+  return pagination(userList, pageSize, {
```
aptkingston, line 280: We were previously passing in the same number here as the number of docs we were fetching. Inside here we check if the data length is longer than the desired page limit, but since this was the same number that was always false. We should have been passing in the actual page size, rather than page size + 1.

packages/builder/src/pages/builder/portal/users/users/index.svelte
```diff
@@ -65,10 +68,12 @@
     { column: "role", component: RoleTableRenderer },
   ]
   let userData = []
+  let invitesLoaded = false
+  let pendingInvites = []
+  let parsedInvites = []
```
aptkingston, line 73: Most of the following changes are simply restructuring the file to be more consistent with our usual structure and more readable.

mike12345567 (approved):
LGTM! Thanks for picking this one up.

Also addresses: https://linear.app/budibase/issue/BUDI-7711/user-table-isnt-displaying-all-the-available-users

Issue #12203, opened 2023-10-27
[BUDI-7701] one-to-many relationship not working on external DB


marco-santulli:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: budi cli
  - Budibase Version:  2.11.45.
- Cloud
  - Tenant ID: We replicated the same behaviour on our aevoluta.com account

Creating a relationship between two tables on a posgres db, we define a one-to-many relationship.  After we save it and re-open it, it becomes many-to-one (and the behaviour reflects a many-to-one relationship

this is what we enter:
![image](https://github.com/Budibase/budibase/assets/82715690/6244a19d-e483-47c8-b58e-22ae7dbde2f4)

then we save, reopen it, and this is what we get:

![image](https://github.com/Budibase/budibase/assets/82715690/3bcc14dd-994d-4322-b352-b8183af1349c)

<sub>[BUDI-7701](https://linear.app/budibase/issue/BUDI-7701/one-to-many-relationship-not-working-on-external-db)</sub>

marco-santulli:
I tested the new version, now it works, however there is still a bug with linked objects, that I thought was caused by this bug.  

The bug is as follows on a posgres DB:
1. I have two tables, say owner and pet.  I create a one-to-many relationship from owner to pet. the foreign key is fk_owner in the pet table.
2. I create two screens (automatic creation) to see owners and pets.  In the pets view, I add a search on the owner.name.  The search doesn't work (it returns random rows).

PR #12233, opened 2023-10-31
Fix issue with directional relationship being swapped

PClmnt:
## Description
Fixes the issue described in #12203. 

When doing the original work for this I hadn't accounted for the `otherRelationshipType()` function which swapped the relationship direction, so it was being swapped correctly then swapped back. This PR removes the swapped definition.


shogunpurple (approved):

mike12345567 (approved):
Got round to trying this - played with it in a few different Postgres configurations - LGTM!

PR #12234, opened 2023-10-31
Stop the sample data being identified as an external source.

deanhannigan:
## Description

When attempting to alter the columns and metadata of the `Sample Data` tables, the updates would fail with the following error: `BadRequestError: No datasource implementation found.`

The `Sample Data` source was being mischaracterised as an external source that could not be found.


mike12345567 (approved):

mike12345567:
@deanhannigan - just to let you know, tests were failing due to a cyclic occurring, I rejigged where the table IDs came from so that they could be retrieved from `constants` without importing a lot of different parts of the codebase (as happens when importing `datasource_bb_default.ts`).

deanhannigan:
@mike12345567 ahh right, I just assumed the test was flaking on me. Cheers!

Issue #12125, opened 2023-10-20
[ internal DB ] - Action Menu "Truncate Table"


poirazis:
Add the ability to truncate an internal table via an Action Menu on the Table in Data View.

PR #12235, opened 2023-10-31
Enable truncating of internal database table

mslourens:
## Description
Adds a button to truncate internal databases

Addresses: 
- #12125

## App Export

## Screenshots
**Confirmation modal**
image

**Toolbar button**
image

## Documentation
- [X] I have reviewed the budibase documentatation to verify if this feature requires any changes. If changes or new docs are required I have written them. (I couldn't find docs about the data toolbar buttons)

### DIFF
--- a/packages/builder/src/components/backend/DataTable/TableDataTable.svelte
+++ b/packages/builder/src/components/backend/DataTable/TableDataTable.svelte
@@ -16,6 +16,7 @@
   import GridRelationshipButton from "components/backend/DataTable/buttons/grid/GridRelationshipButton.svelte"
   import GridEditColumnModal from "components/backend/DataTable/modals/grid/GridEditColumnModal.svelte"
   import GridUsersTableButton from "components/backend/DataTable/modals/grid/GridUsersTableButton.svelte"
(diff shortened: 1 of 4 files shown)


mike12345567:
Hi @mslourens - thank you for the submission! I think this is a great idea - but it would probably be better to build out a table API specifically for truncating a table, as I think today this will only delete the rows which have been loaded into the grid. If you have say 10K+ rows in a table, you would need to scroll entirely through the grid to load all rows, then you could delete all of them.

A backend endpoint could instead fetch the IDs of the rows and make a bulk request to delete everything at once - being as efficient as possible.

aptkingston:
> Hi @mslourens - thank you for the submission! I think this is a great idea - but it would probably be better to build out a table API specifically for truncating a table, as I think today this will only delete the rows which have been loaded into the grid. If you have say 10K+ rows in a table, you would need to scroll entirely through the grid to load all rows, then you could delete all of them.
> 
> A backend endpoint could instead fetch the IDs of the rows and make a bulk request to delete everything at once - being as efficient as possible.

Correct - the grid rows store will only contain the rows that have been loaded in, which will increase as you continue to scroll.

shogunpurple:
Closing for now as it won't always fully truncate the table - we will revisit this in future

PR #12243, opened 2023-11-01
Single image build fixes

mike12345567:
## Description
Updating single image build - the old method of installing Node 18 was deprecated, switching to the new mechanism.


adrinr (approved):
LGTM

Issue #10676, opened 2023-05-23
[BUDI-7105] Builder Option for Read Only Form Fields


shawnweeks:
**Describe the feature request**
Add an option in the builder to set form fields to read only by setting the html input tag property `readonly`, this would be similar in nature to the existing `disabled` option and should also be included in the component conditional options. Primary difference is setting read only option would not dim the field but would still prevent normal user modification. This is purely cosmetic but would allow some flexibility to how form fields are presented especially when used in non-traditional ways.

**Screenshots**

<sub>[BUDI-7105](https://linear.app/budibase/issue/BUDI-7105/builder-option-for-read-only-form-fields)</sub>

shawnweeks:
I'm interested in contributing to this if this seems like something appropriate to the project.

melohagan:
Hey @shawnweeks 

This is possible through global styling! https://docs.budibase.com/docs/custom-css#global-styling

Add your embed component with the following content:

```html
<style>
.is-disabled > input:disabled {
  background-color: dimgrey !important;
  color: linen !important;
  opacity: 1 !important;
}
</style>
```

![Screenshot 2023-06-08 at 17 07 35](https://github.com/Budibase/budibase/assets/101575380/2cfae08c-a202-48a8-9ea1-7bbd043c1359)

You can change the style in anyway you like!

shawnweeks:
@melohagan Wouldn't this apply to all disabled form inputs? I'm wanting to set the `readonly` property on the actual input component. CSS doesn't exactly support canceling out an existing setting.

melohagan:
Hey @shawnweeks 

Thanks for the docs link. I gave the readonly setting a go and I wasn't able to notice much of a difference, for example I could copy and paste the content from both a disabled and readonly field - at least on Google Chrome, MacOS.

One difference was the focus, I was able to tab through read-only fields but not disabled fields.

I think @aptkingston would have a better opinion on this, and it may be worth adding.

shawnweeks:
I've spent some time messing with the global styles but I've never gotten it to make a disabled field look like a non disabled one. The behavior for selecting the contents of disabled fields seems to behave different in firefox.
[Fields Example-export-1686312355097.tar.gz](https://github.com/Budibase/budibase/files/11706092/Fields.Example-export-1686312355097.tar.gz)

aptkingston:
[...]

Issue #6979, opened 2022-07-29
Default preview for Rich text (markdown) Long Form Field


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/6911

<div type='discussions-op-text'>

<sup>Originally posted by **kasgithub8** July 27, 2022</sup>
Is there any way to pre-select the "Toggle preview" button?
![image](https://user-images.githubusercontent.com/95284278/181480179-4e5f2534-0d97-40ce-8fbd-81b3bc4b7ec3.png)

Thanks!</div>

Could add a checkbox 'Preview by default'. Alternatively, or in addition, the default could be effected by the form mode:
![Screenshot 2022-07-29 at 11 28 50](https://user-images.githubusercontent.com/101575380/181740521-f9af5be7-bd10-4151-8b0d-387a2393f821.png)
 
IF create THEN not in preview mode by default (you have to add a value)
IF update THEN in preview mode by default (a value is already present)

PR #12244, opened 2023-11-01
Read only form fields

melohagan:
## Description
Various form fields now have a setting to make them *readonly*. 

The two exceptions are attachment fields and the code scanner fields which have the same behaviour regardless of whether the parent form is disabled or readonly.

For form blocks, *View* mode now sets the fields to *readonly* instead of *disabled*, as disabled fields are quite difficult to read, however you can override this with the disabled setting if wanted.

If disabled is checked, then the readonly setting will not display.

Addresses: 
- https://github.com/Budibase/budibase/issues/10676
- https://github.com/Budibase/budibase/issues/6979

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/5a0fdb39-4674-4623-8c9c-24db78920baa

## Feature branch env
[Feature Branch Link](http://fb-budi-7105-builder-option-for-read-only-form-fields.fb.qa.budibase.net)


packages/bbui/src/Form/Core/Checkbox.svelte
```diff
@@ -29,6 +30,11 @@
     checked={value}
     {disabled}
     on:change={onChange}
+    on:click={e => {
+      if (readonly) {
+        e.preventDefault()
+      }
+    }}
```
aptkingston, line 37: Rather than hijacking clicks, for these components that don't natively support readonly, could we use a CSS solution instead?

I'm thinking something like:
```js
<label
  class="spectrum-Checkbox spectrum-Checkbox--emphasized {sizeClass}"
  class:readonly
>
....
<style>
.readonly {
  pointer-events: none;
}
</style>
```

That's more performant in the sense that it reduces additional event handlers, and also avoids any potential side effects to due handling click events.

packages/bbui/src/Form/Core/CheckboxGroup.svelte
```diff
@@ -40,6 +41,11 @@
         >
           <input
             on:change={onChange}
+            on:click={e => {
+              if (readonly) {
+                e.preventDefault()
+              }
+            }}
```
aptkingston, line 48: Same as above regarding click handlers.

packages/bbui/src/Form/Core/RadioGroup.svelte
```diff
@@ -43,6 +44,11 @@
       >
         <input
           on:change={onChange}
+          on:click={e => {
+            if (readonly) {
+              e.preventDefault()
+            }
+          }}
```
aptkingston, line 51: Same as above regarding click handlers.

packages/bbui/src/Markdown/MarkdownEditor.svelte
```diff
@@ -43,7 +45,7 @@
   const debouncedUpdate = debounce(update, 250)
 </script>
 
-{#key height}
+{#key (height, readonly)}
```
aptkingston, line 48: We don't key using `disabled` which is used in the same way, so I'm assuming we either *do* need to key by disabled (if it does not toggle properly) or *do not* need to key by `readonly`.

aptkingston (commented):
Nice work @melohagan! Just left one suggestion about how to better handle click events in some of the fields 👍

packages/bbui/src/Form/Core/Checkbox.svelte
melohagan: https://github.com/Budibase/budibase/assets/101575380/2abf6732-ec42-49fc-bd9e-55ba5242cd10

I gave this a go, but I ran into this problem were the click event wasn't prevented on the label. Not sure why.

packages/bbui/src/Markdown/MarkdownEditor.svelte
melohagan: Nice catch - was able to remove the readonly key.

packages/bbui/src/Form/Core/Checkbox.svelte
melohagan: As discussed, added the readonly class to the parent Field component label.

aptkingston (approved):
LGTM!

PR #12246, opened 2023-11-01
Fix to - Issue running standard Docker Compose instance of bbapps in CLUSTER_MODE

kellis5137:
FYI: I signed an Individual Contributor Agreement and sent it to community@budibase.com. Please let me know if there is anything else I need to do in this regard.

## Description
My problem can be found here - https://github.com/Budibase/budibase/discussions/11814#discussioncomment-7210547. After some very intense diving into Budibase, KOA, and NodeJS code, I determined that the particular method ("server.address()") returns null in the case where the NodeJS Cluster module is used to start Budibase server. This happens whenever CLUSTER_MODE is set to true and PM2 starts the server. The bug appears to be related to NodeJS and enhancements to use "nodelay" with the HTTP server module. In versions 18+, the feature was defaulted to "true", instead of false. I believe the removal of the delay is the problem. 

Please note, it's a little difficult to debug a clustered system. However, by tracing down code and using monitoring software, I was able to determine the issue and create a solution. In my submitted code, you'll find that I have used the existing CLUSTER_MODE environmental variable within the startup code to skip the log statements that fail. I then was able to monitor the use of the cluster with Linux "top" while running a load with JMeter. You'll see that with my change, PM2 clustering is working as expected. Please see attached pic.

In addition to adding the flag, I also tested it to ensure that it would never affect anyone who does not use clustering. They will start as expected and return the existing log statements. Please let me know if you have any questions. I hope this code may be added to the main baseline so that I and others who use cluster mode will be able to update to the latest instance.

Addresses: 
-  https://github.com/Budibase/budibase/discussions/11814#discussioncomment-7210547

## App Export
- Related to the core functionality of the server app

## Screenshots
# Before my fix
image

You can see the full error message in the discussion

#But after my fix
image

## Documentation
- [X] I have reviewed the budibase documentation to verify if this feature requires any changes. If changes or new docs are required I have written them.


kellis5137:

kellis5137:
recheck

kellis5137:
recheck

packages/server/src/startup.ts
```diff
@@ -58,7 +58,7 @@ export async function startup(app?: any, server?: any) {
     return
   }
   STARTUP_RAN = true
-  if (server) {
+  if (server && env.CLUSTER_MODE?.toLowerCase() !== "true") {
```
shogunpurple, line 61: You could just do `&& env.CLUSTER_MODE` here I believe. Usually for env variables we just set them to `1` if it's a boolean

kellis5137:
So change it to

if (server && env.CLUSTER_MODE)

and the I should be able to turn it on with 1? I'll change it.

packages/server/src/startup.ts
kellis5137: changed it accordingly! Let me know if that works

shogunpurple (approved):
LGTM! @mike12345567 wouldn't mind if you run an eye over this just to make sure - seems good from a local test though

mike12345567 (commented):
I'm fine for this to be merged - just need to do a full build locally and make sure the standard docker-compose setup is unaffected by the cluster mode option.

shogunpurple:
Tested this and it's working as expected - waiting for CI then auto-merge should kick in!

Issue #8939, opened 2022-12-06
Add Google Cloud Storage File Upload


s1223:
https://cloud.google.com/storage
Like S3 File Upload but on google cloud storage

phaethon:
+1

vincenewton:
I think this is a fantastic proposal and I'm excited to see it implemented.

braydennjinu:
I fully support the idea presented and believe it to be a great solution.

hazzenut23:
I think the idea is very good, and I am happy to endorse it.

PR #12267, opened 2023-11-03
Google Cloud Storage Integration

deanhannigan:
## Description

Adding Google Cloud Storage support to Budibase. 

Query, create and manage buckets in your Google Cloud Storage and allow users to upload new media via Budibase.

Screenshot 2023-11-03 at 09 38 28

### Query Types

- Create Bucket
  - `Bucket Name` - the name of your new bucket. You will be notified if the bucket name is not unique
  - `Location` - http://g.co/cloud/storage/docs/locations#location-mr
  - `Storage Class` - https://cloud.google.com/storage/docs/storage-classes
  - `Public` (Optional) - this will provide completely unauthenticated access to the bucket. Ensure that this is the desired outcome.

- Delete File By Key
  - `Bucket Name` - the name of the target bucket.
  - `Delete` - the key of the file to be deleted. You can hard code this or use a binding to build a dynamic delete in your application

- Read Bucket
  - List all available media in your bucket and integrate them into your applications.
  - `Bucket Name` - the name of the target bucket.

### Authorization

The integration currently supports  [Service Account](https://cloud.google.com/iam/docs/service-account-overview) authorization by JSON token; the contents of which are used to configure the datasource. The following fields are required from the token.
  - `Project Id`
  - `Private Key`
  - `Client Email` - the service user email.

Privileges granted to the service can be managed in the Google Cloud Console.

Addresses:

- https://github.com/Budibase/budibase/issues/8939

## Screenshots

The `Google Cloud Storage File Upload` component is now available in the components list.

Screenshot 2023-11-03 at 11 15 08

The `Upload File to GoogleCloud` button action

Screenshot 2023-11-03 at 11 17 14

## Feature branch env
[Feature Branch Link](http://fb-labday-google-cloud-storage.fb.qa.budibase.net)


packages/client/src/components/app/forms/GoogleCloudUpload.svelte
```diff
@@ -0,0 +1,155 @@
...
+
+      // Don't read in non-image files
+      data = fileList[0]
+      if (!data.type?.startsWith("image")) {
+        resolve([
+          {
+            name: data.name,
+            type: data.type,
+          },
+        ])
```
samwho, line 49: Maybe a silly question, but do you need a `return` after this `resolve` call? Without it, doesn't the function continue?

samwho: Did a little browser console noodling and I believe you likely do want to return after this:

CleanShot 2023-11-29 at 10 12 59@2x

packages/server/src/api/controllers/static/index.ts
```diff
@@ -269,9 +270,14 @@ export const getSignedUploadURL = async function (ctx: Ctx) {
   const awsRegion = (datasource?.config?.region || "eu-west-1") as string
   if (datasource?.source === "S3") {
     const { bucket, key } = ctx.request.body || {}
+
+    // Ensure we aren't using a custom endpoint
+    if (datasource?.config?.endpoint) {
+      ctx.throw(400, "S3 datasources with custom endpoints are not supported")
```
samwho, line 276: How come?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -287,6 +293,37 @@ export const getSignedUploadURL = async function (ctx: Ctx) {
     } catch (error: any) {
       ctx.throw(400, error)
     }
+  } else if (datasource?.source === "GOOGLE_CLOUD") {
+    const { bucket, key } = ctx.request.body || {}
+
+    const parsedKey = datasource?.config?.privateKey
+      .split(String.raw`\n`)
+      .join("\n")
```
samwho, line 301: What's the purpose of these two lines? Maybe it's me being naive, but don't they cancel each other out?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -287,6 +293,37 @@ export const getSignedUploadURL = async function (ctx: Ctx) {
...
+
+    publicUrl = `https://storage.cloud.google.com/${bucket}/${key}`
```
samwho, line 303: I believe both `bucket` and `key` can be `undefined` here. We likely want to do a check before attempting to make any calls.

packages/server/src/integrations/googlecloud.ts
```diff
@@ -0,0 +1,173 @@
...
+      autoPaginate: false,
+    })
+    return files.map((file: any) => file.metadata)
+  }
+
+  async delete(query: { bucket: string; delete: string }) {
+    return await this.storage.bucket(query.bucket).file(query.delete).delete()
+  }
+
+  async create(query: {
```
samwho, line 150: Generally you want to annotate methods that override something in a base class with the `override` keyword, e.g.

```typescript
  async override create(query: {
```

Doing this allows the TypeScript compiler to check for you that the method is overridden correctly. 🙂

samwho (commented):
I know this isn't quite ready for merging yet but I'm doing a sweep of all PRs and thought I'd take a look. Hope that's okay.

melohagan (changes requested):
It looks good, however I feel like we should try and reduce the duplicate code, actions and components, because Azure Storage can also be added: https://github.com/Budibase/budibase/pull/12530

packages/server/src/api/controllers/static/index.ts
deanhannigan: I ran into [this](https://stackoverflow.com/questions/74131595/error-error1e08010cdecoder-routinesunsupported-with-google-auth-library) issue when trying to authenticate. Turned out that the value from the google key gets escaped when stored in the db.

This code splits on escaped (`\\n`) new lines and replaces them before we use them. `String.raw` actually escapes the newline value passed in, turning `\n` into `\\n`. This is kind of opaque. The other example from that post is definitely clearer.

Example:

```javascript
// An string with an escaped new line 
let test = "Hi\\nthere!"

test.split(String.raw`\n`)
// String.raw escaped `\n` becoming `\\n`
// The string splits on the escaped value ['Hi', 'there!']

test.split(String.raw`\n`).join('\n')
// Finally becomes 'Hi\nthere!'
```

packages/server/src/api/controllers/static/index.ts
deanhannigan: I don't believe our infrastructure can currently support custom endpoints. There would have to be a fairly big update to support file storage and retrieval in the builder. Definitely possible though.

packages/server/src/integrations/googlecloud.ts
deanhannigan: Does this apply to classes that implement an interface? I added the keyword but VS complains that it can't be used as `'GoogleCloudIntegration' does not extend another class`

deanhannigan:
> It looks good, however I feel like we should try and reduce the duplicate code, actions and components, because Azure Storage can also be added: #12530

I replaced the duplicated UI/UX with a common `StoreUpload` component. Both S3 and GCS use it with a type parameter to discern which cloud storage endpoint is being targeted

packages/server/src/integrations/googlecloud.ts
samwho: Ah, sorry, it does not. My mistake!

packages/server/src/api/controllers/static/index.ts
samwho: Ewwww, that's not fun 😞

shogunpurple:
@deanhannigan can we move this to draft if it's not ready for review/merge?

PR #12274, opened 2023-11-03
Eliminate TOCTOU problem in creating bbTmp.

samwho:
## Description

The way this was previously done had a [time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use problem that could cause flakes in CI: https://github.com/Budibase/budibase/actions/runs/6745433029/job/18337235959.

The new method always attempts to create the directly, and ignores the case where the create fails because the directory already exists.


mike12345567 (approved):
LGTM!

Issue #12272, opened 2023-11-03
[BUDI-7725] Auto-generated table screen will add a user field with a multi-select even when the User column is created without the 'Allow multiple users' setting toggled on 


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.12.2

**Describe the bug**
I have a table correctly configured with a user column that only allows a single user to be selected. The problem is that when I create an auto-generated screen, the table block adds a user field that incorrectly supports multi-select.

**To Reproduce**

1. Add a user column to a table, with the 'Allow multiple users' setting *off*.
2. See that within the **Data** section, only one user can be selected.
3. Go to the **Design** section and add an autogenerated 'Table' view
4. Preview the app and see that the user field allows you to select and save multiple users (this is wrong)

**Expected behavior**
When a table view screen is auto-generated, it should take into account the 'Allow multiple users' setting being off.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/b839a4f3-f9c1-4026-9590-3b87a7fadfd4/e0b252ae-10f8-4273-b5fb-91414a7d8c0e?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9iODM5YTRmMy1mOWMxLTQwMjYtOTU5MC0zYjg3YTdmYWRmZDQvZTBiMjUyYWUtMTBmOC00MjczLWI1ZmItOTE0MTRhN2Q4YzBlIiwiaWF0IjoxNjk5MDE4MTUzLCJleHAiOjE2OTkxMDQ1NTN9.8Si8ivCOTEdOexSoec3_WAcvShh6QxmjWRS6GQUb59Q)

[...]

PR #12275, opened 2023-11-03
Set the relationship type

melohagan:
## Description
The user field component makes use of the relationship field component, therefore we need to ensure the relationshipType is set.
Undefined defaults to multi-select.

Addresses:
- https://github.com/Budibase/budibase/issues/12272
- https://github.com/Budibase/budibase/issues/12302

## Screenshots

Screenshot 2023-11-03 at 15 07 27

Screenshot 2023-11-03 at 15 07 47

rel_bug


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -143,6 +144,8 @@
       )?.[0]
       // Set the tableId based on the selected table
       editableColumn.tableId = relationshipTableIdSecondary
+    } else if (editableColumn.subtype === USER_TYPE) {
```
adrinr, line 147: Don't we need the `USERS` type as well?

melohagan: The picker does default to many, but it probably is worth being explicit.

Would multiple users be `many-to-one` or `many-to-many` in this case?

melohagan: users type added

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -143,6 +145,10 @@
...
+      editableColumn.relationshipType = RelationshipType.ONE_TO_MANY
+    } else if (editableColumn.subtype === USERS_TYPE) {
+      editableColumn.relationshipType = RelationshipType.MANY_TO_MANY
```
mike12345567, line 151: Can you update the `BBReferenceFieldMetadata` type in `types/src/documents/app/table/schema.ts` to include the `relationshipType` field as the `RelationshipType` enum - just for future reference, while it isn't used on backend, good to know the structures used.

mike12345567 (changes requested):
One small change - once thats done good to go!

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -287,6 +289,12 @@
     if (saveColumn.type !== LINK_TYPE) {
       delete saveColumn.fieldName
     }
+    if (saveColumn.subtype === USER_TYPE) {
```
mike12345567, line 292: should you confirm it is a bbReference type first just to make sure that it is a user column?

melohagan: Now checking bb_reference type

mike12345567 (approved):
LGTM!

PR #12280, opened 2023-11-03
Restore bull parameter changes + reduce Redis config complexity

mike12345567:
## Description

This PR restores the functionality from: https://github.com/Budibase/budibase/pull/12237

This was reverted due to issues that it was causing in the QA and production environments.

I have tested this now in the QA environment and believe that the issue was due to the `redisProtocolUrl` that we used to use over the separate parameters was causing some problems, such as a password not correctly being set, or the queue parameters defaulting to a localhost Redis connection.

Instead, I have switched the response to always use the `RedisOptions` which we use during development. This means that if you pass in `REDIS_PASSWORD` + `REDIS_URL` or just `REDIS_URL` with the password set via a string like `redis://:<password>@localhost:6379` we will always use the same mechanism for connecting - reducing the difference between the development environment and the Cloud like environments.

As part of this I also had to update the typing around the public API as we use a package called `koa2-ratelimit` and its Redis store for rate limiting the public API. This cannot use the standard RedisOptions for its connection to Redis, so an update was needed.

## Feature branch env
[Feature Branch Link](http://fb-fix-update-bull-queue-parameters.fb.qa.budibase.net)


shogunpurple (approved):
LGTM! Nice work @mike12345567

PR #12283, opened 2023-11-04
[Changed] Posthog's feature flags Ids

jvcalderon:
## Description

In account portal currently we are using surrogated keys that are not constant between environments; this PR changes that to use keys instead.


shogunpurple (approved):
LGTM!

PR #12289, opened 2023-11-06
Frontend/Builder Monolith Refactor

deanhannigan:
## Description

The front end store has been removed and replaced with multiple new context specific stores. This work was initiated in  response to the growing size of `frontend.js` and the performance hit incurred with all stores hitting a single store.

New stores:
- Preview - limited only to builder preview state
- Component - a huge store containing all component data and logic.
- Navigation - app links and branding
- Screens - all screen data and logic. All derived screen stores, including the history, have been consolidated here.
- App - app metadata
- Hover - dedicated store to avoid to much chatter.
- Builder - all builder state
- Layouts

Most of the new stores have complete unit test coverage, with some others in progress.

New stores are now structured as classes that extend `BudiStore`. The new base class contains boilerplate for core store functionality. There is also an options parameter in the super constructor where can you can pass in `{ debug: true }`. This will emit store changes to console prefixed with the class name.

Stores have been reorganised into two categories: `builder` and `portal`.
- `stores/builder` - app related data/functionality needed when editing apps.  This includes a large number of stores that were not part of the frontend.js refactoring. Eventually all of these stores should extend the `BudiStore` class.
- `stores/portal` - Budibase wide stores concerning things like auth, utils and global theme.

Builder initialisation and aliasing has also been centralised into one location: `stores/builder/index.js`

### Address:
- https://linear.app/budibase/issue/BUDI-7427/refactor-monolithic-frontend-store-and-add-tests

## TODO

- More test coverage for all stores moved to `stores/builder`
- Refactor all existing stores that were relocated to `stores/builder` to extend the new `BudiStore` class

## Feature branch env
[Feature Branch Link](http://fb-feature-monolith-js-refactor.fb.qa.budibase.net)


aptkingston:
image
Love to see it 👌 This eclipses anything I've seen before!

aptkingston (approved):
I've merged in my changes. Let's get this updated again 👌

Issue #12291, opened 2023-11-06
[BUDI-7736] Relationship picker required validation fires immediately


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.12.11

**Describe the bug**
When adding required validation to a relationship picker that has no selected value, the field is highlighted red immediately.

**To Reproduce**

1. Add a create form
2. Add a relationship picker
3. Add required validation -> see that validation is triggered immediately

**Expected behavior**
Validation should not trigger until the relationship picker has been touched, or the validation form action is fired.

**Screenshot**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/08389ab5-572e-4e8b-aa9c-5594fc12d712/bc88c47d-4e47-48c8-938a-e47ae4f7ada5?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8wODM4OWFiNS01NzJlLTRlOGItYWE5Yy01NTk0ZmMxMmQ3MTIvYmM4OGM0N2QtNGU0Ny00OGM4LTkzOGEtZTQ3YWU0ZjdhZGE1IiwiaWF0IjoxNjk5Mjc4Mjc2LCJleHAiOjE2OTkzNjQ2NzZ9.\_SYnPmqjhV0ixAC45EE5wj1QJT-lgN_tmvIFUtz8yJU)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7736](https://linear.app/budibase/issue/BUDI-7736/relationship-picker-required-validation-fires-immediately)</sub>

PR #12293, opened 2023-11-06
Don't trigger required validation early

melohagan:
## Description
Fix to make sure the field state isn't updated if there are no selected options to prevent the required field validation being triggered early.

### Addresses:
- https://github.com/Budibase/budibase/issues/12291

## Screenshots
Screenshot 2023-11-06 at 14 00 09

*Initial state*

Screenshot 2023-11-06 at 14 00 38

*State after validation action*


shogunpurple (approved):
LGTM

PR #12297, opened 2023-11-06
Fix `UPDATE_ROW` automation step from blanking out link fields if they haven't changed.

samwho:
## Description

This bug appears to have been fixed in other changes since it was filed (following up with @PClmnt about that), so in the meantime I've added some tests for this behaviour to make sure it doesn't regress.


shogunpurple (approved):
LGTM!

PR #12298, opened 2023-11-06
Allow null strings

melohagan:
## Description
Currently if a field in the front-end is left blank, it will pass a blank string, but this needs to update to null when saving text to a SQL DB.

### Addresses:
- https://ptb.discord.com/channels/733030666647765003/1164884245991399444

## Screenshots
Screenshot 2023-11-06 at 16 44 21

*After change*
Screenshot 2023-11-06 at 16 42 56

*Before change*
Screenshot 2023-11-06 at 16 59 35

## Feature branch env
[Feature Branch Link](http://fb-fix-allow-null-strings.fb.qa.budibase.net)


shogunpurple (approved):
LGTM! makes sense as the default behaviour

aptkingston (approved):

PR #12299, opened 2023-11-06
Fix filtering/sorting of row exports

deanhannigan:
## Description

Filtering and sorting of exported table rows reintroduced. This appears to be a feature regression as this behaviour was previously fully supported for both internal and external datasources.

### Addresses:
- Internal Export - missing sorting functionality when exporting table rows with a sort column applied.
- External Export - missing both filtering and sorting functionality.

## Feature branch env
[Feature Branch Link](http://fb-fix-export-row-data.fb.qa.budibase.net)


Ghrehh (approved):
LGTM 🥚

aptkingston (approved):

PR #12300, opened 2023-11-06
Update nginx-ssl.conf

gcglinton:
## Description
The nginx config for SSL doesn't match the one for non-SSL, so things don't work as expected.
Copied the default one to the SSL one, re-adding the few directives/parameters required to make SSL work.

Per @shogunpurple's comment in #12140, submitting a new PR into `master`.

### Addresses:
The nginx config that's used when enabling SSL is missing some location blocks, and it causes the apps to not load.
- #10741


shogunpurple:
Thanks for the contribution @gcglinton !

Issue #11972, opened 2023-10-04
[BUDI-7593] External datasource formulas don't allow for relationship fields to be outputted in formula columns.


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.11.8
  * App Version: 2.11.8

**Describe the bug**
When trying to display relationship fields values attached to records in an external data source nothing happens. When you select for example `{{ jobs.0.job_title }}` which does have a value for a formula column nothing appears after saving. You can output the `{{ jobs }}` but this only outputs the `primaryDisplay` and the `_id`.

**To Reproduce**

1. Import the app export
2. Use the Postgres dump schema to create the 2 tables
3. Use the insert data to add data to the tables
4. Connect your Postgres data source and fetch the tables
5. Go to the formula field in the users table, edit and try select a job field binding and save
6. Nothing appears even though there is data in that relationship field

**Expected behavior**
You should be able to access the relationship fields within a formula column. It should output and display whichever one is selected. This did work previously for external data sources but seems to not be working anymore.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/4f8cdab3-9a16-4ddb-9896-1f24199edd41/a4f2ebf3-ac7c-42e9-ba55-707080fee9f7?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC80ZjhjZGFiMy05YTE2LTRkZGItOTg5Ni0xZjI0MTk5ZWRkNDEvYTRmMmViZjMtYWM3Yy00MmU5LWJhNTUtNzA3MDgwZmVlOWY3IiwiaWF0IjoxNjk4ODI3MjM2LCJleHAiOjE2OTg5MTM2MzZ9.6mRJHKm1iL6oyogLMkf3Ub6\_-uDWofIcHB7GgQ9vrCo)

[...]

PR #12301, opened 2023-11-06
Datasource plus formulas using relationships

mike12345567:
## Description
Getting external rows to be more consistent with the internal API - the formulas should be processed using the `outputProcessing` as they were being processed, but was causing an overwrite. The problem was that the external system internally processed the formulas/relationships, then squashed the relationships. Once it got to the external API, it would go through normal output processing, which would run over the squashed rows, causing an inconsistent behaviour.

Added test case as well to confirm that relationships can be used in formulas for both external and internal tables.

Addresses: https://linear.app/budibase/issue/BUDI-7593/external-datasource-formulas-dont-allow-for-relationship-fields-to-be and https://github.com/Budibase/budibase/issues/11972

## Feature branch env
[Feature Branch Link](http://fb-fix-11972-external-relationship-formulas.fb.qa.budibase.net)


packages/backend-core/src/utils/utils.ts
```diff
@@ -237,3 +237,17 @@ export function timeout(timeMs: number) {
...
 }
+
+export function hasCircularStructure(json: any) {
+  if (typeof json !== "object") {
+    return false
+  }
+  try {
+    JSON.stringify(json)
+  } catch (err) {
+    if (err instanceof Error && err?.message.includes("circular structure")) {
```
samwho, line 248: This makes me nervous. I was hoping `lodash` would have something that could do this in a more concrete way but they appear to recommend this exact approach instead: https://github.com/lodash/lodash/issues/1548.

Given that there's a test and I imagine the regressive behaviour would be an infinite loop, we'd catch this as soon as it became a problem. Still, gives me the shivers.

packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -233,6 +232,11 @@ export async function outputProcessing<T extends Row[] | Row>(
       })
     : safeRows
 
+  // make sure squash is enabled if needed
+  if (!opts.squash && utils.hasCircularStructure(rows)) {
+    opts.squash = true
+  }
+
```
samwho, line 239: If we can detect when this is needed, should it be an option at all?

packages/server/src/utilities/rowProcessor/utils.ts
```diff
@@ -40,52 +45,50 @@ export function fixAutoColumnSubType(
 /**
  * Looks through the rows provided and finds formulas - which it then processes.
  */
-export function processFormulas(
+export function processFormulas<T extends Row | Row[]>(
```
samwho, line 48: What's the associativity here? Is this saying `T extends Row | T extends Row[]`? Or is it `(T extends Row) | Row[]`?

packages/server/src/utilities/rowProcessor/index.ts
mike12345567: yes - squashing isn't just for removing circular dependencies, it can be used in a variety of different ways - if it was not possible to trigger it we would have to force circular dependencies to be able to squash relationships.

packages/server/src/utilities/rowProcessor/utils.ts
mike12345567: it works like `T extends Row | T extends Row[]` - we've used a format like this before a few times, for something like this where a function can process either type, its nice to be able to make sure it denotes the correct response type as well, so this function is aware that if type `Row` is passed in, you'll get `Row` out, or vice versa for `Row[]`

packages/backend-core/src/utils/utils.ts
mike12345567: I think its pretty safe as `JSON.stringify` has always been protected against this - its impossible for it to generate a string when it contains a circular structure, and IMO keeping track of the state isn't always that easy, I prefer this method for managing rows than just trusting Koa to look after it correctly.

samwho: I'm more worried about the error message changing and we start to incorrectly let objects through.

samwho (approved):

PR #12303, opened 2023-11-07
Fix for 501 on user invitation

mike12345567:
## Description
Fix for user invitations throwing a 501 error due to the way the search was being performed.

## Feature branch env
[Feature Branch Link](http://fb-fix-invitation-user-search.fb.qa.budibase.net)


shogunpurple (approved):
LGTM! Although a test to cover would be nice too

packages/frontend-core/src/fetch/UserFetch.js
```diff
@@ -33,7 +33,7 @@ export default class UserFetch extends DataFetch {
     let finalQuery
     // convert old format to new one - we now allow use of the lucene format
     const { appId, paginated, ...rest } = query
-    if (!LuceneUtils.hasFilters(query) && rest.email) {
+    if (!LuceneUtils.hasFilters(query) && rest.email != null) {
```
samwho, line 36: Is this intentionally a coercing check (`!=` vs `!==`)? What's the specific thing you're changing this for?

mike12345567: yes, its the one singular time I'm ok with a `!=` rather than a specific check, as `!= null` checks for both of the empty states, null and undefined, whereas `!== null` will miss undefined.

PR #12310, opened 2023-11-07
Add buttons to grids

### LINKED ISSUE #11746: [BUDI-7478] Expose state store via client library SDK
Currently there is no awareness of what actions attached to an `on change` or `on click` event when creating custom components. Amend the client library's SDK to expose these elements so they can be pulled into custom plugins and mutated within them.

**Conditions of Satisfaction (COS)**

* Expose additional stores and allow them to be used within custom plugins.

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7478](https://linear.app/budibase/issue/BUDI-7478/allow-users-to-manually-mutate-actions-for-custom-components)</sub>

### LINKED ISSUE #12307: [ SDK ] - Expose binding evaluation function(s) via SDK context
Although not so common, there are cases when a custom component receives the raw binding string and needs to evaluate said string. To avoid bundling "@budibase/string-templates" into the custom component unecessarily, please expose appropriate functions to evaluate bindings via the SDK.

I am currently using `processStringSync` but i am not sure if there are others which i am not aware or havent encoutered their need yet.

aptkingston:
## Description
This PR allows you to add buttons to grids, and configure actions to run using the relevant row. 

image

Buttons go into a column that sticks to the right side of the grid when the grid width exceeds the available space, and simply acts as the last column when there is available space.

image

You can see here that it sticks right, then slots in as the last column when space allows:

https://github.com/Budibase/budibase/assets/9075550/6a74e6ac-9c05-4a77-aae3-eaf812b1f8f2

The buttons use the same settings configuration as the new button group component:
image

The grid will allow a maximum of 3 buttons. When 3 buttons have been added, the ability to add more buttons will be disabled.
image

There are a few new bindings provided to help you reference the specific row that your buttons concern. There's a category of binding called the name of your grid block, and it provides all the fields for the relevant row:
image

Within this category, there's also a special binding provided which represents the entire row - the "Clicked row":
image

The grid also makes itself available as a datasource to use in things like a "Save row" action, so it's extremely simply to do something like update a field in the clicked row. Here's a quick example showing me updating a column using a button, and the corresponding settings to do that:

https://github.com/Budibase/budibase/assets/9075550/bc40b7a4-2009-44fd-84c0-b052975cc130

My save row button has 1 action:
image

Which is a simple save row action, overwriting one field and leaving the rest untouched:
image

### SDK additions
I've also added a few utils to the client library SDK to help plugin authors.
- Added `processStringSync` from string templates
- Added `stateStore` to allow manual control of state

### Other stuff
As part of this I've removed the restriction that previously prevented you from being able to remove your last button. This made sense for the button group component, but for grids we need to be allowed to have 0 buttons. Having 0 buttons in a button group is also totally fine (it's just an empty container in that case) so there's no harm in allowing users to remove all buttons if they really want to.

### Addresses
- https://linear.app/budibase/issue/BUDI-7543/grid-block-add-built-in-buttons-to-the-grid
- Fixes https://github.com/Budibase/budibase/issues/12307
- Fixes https://github.com/Budibase/budibase/issues/11746

## Feature branch env
[Feature Branch Link](http://fb-grid-buttons.fb.qa.budibase.net)


packages/frontend-core/src/components/grid/layout/ButtonColumn.svelte
```diff
@@ -0,0 +1,144 @@
...
+  })
+</script>
+
+<!-- Hidden copy of buttons to measure -->
+<div class="measure" bind:this={measureContainer}>
+  <GridCell width="auto">
+    <div class="buttons">
+      {#each buttons as button}
+        <Button size="S">
+          {button.text || "Button"}
```
deanhannigan, line 52: Bindings aren't currently being applied to the button labels

Screenshot 2023-11-07 at 15 49 01

packages/frontend-core/src/components/grid/layout/ButtonColumn.svelte
```diff
@@ -0,0 +1,144 @@
...
+    visibleColumns,
+    scroll,
+    isDragging,
+    buttonColumnWidth,
+  } = getContext("grid")
+
+  let measureContainer
+
+  $: buttons = $props.buttons?.slice(0, 3) || []
+  $: columnsWidth = $visibleColumns.reduce(
```
deanhannigan, line 24: 👀

deanhannigan (approved):
LGTM! Besides the minor binding issue, everything looks great!

packages/frontend-core/src/components/grid/layout/ButtonColumn.svelte
aptkingston: Don't even 👀 it pained me to use reduce.

packages/frontend-core/src/components/grid/layout/ButtonColumn.svelte
aptkingston: Nice catch! After looking into this, I think we just have to disable bindings here. The setting needs to be marked as nested, but doing so makes it think that it should be able to access the grid row context for the button name, which it actually can't. I'd also have to manually run string templates to do the enrichment, which is less than ideal as it won't properly update automatically.

So to that effect, I've basically had to just disable bindings for the text setting of buttons inside button groups, *only when the buttonConfiguration setting is nested*.

This means that in normal usage of a button group, you can bind as normal:
image

But when used inside a grid block for example, you cannot use bindings for the text setting:
image

This has no impact on the other settings.

deanhannigan (approved):
LGTM! The new `disableBindings` flag now successfully hides the binding UX 👍

poirazis:
Can you also please add `makePropSafe` from @budibase/string-templates to the SDK as we need it for building custom blocks ?

aptkingston:
> Can you also please add `makePropSafe` from @budibase/string-templates to the SDK as we need it for building custom blocks ?

Will do @poirazis!

Issue #12162, opened 2023-10-24
[BUDI-7685] Adding _Update Setting_-Condition to _Section_-Component crashes frontend


LennysHead:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: kubernetes
  - Budibase Version: 2.11.8
  - App Version: v18.18.0

**Describe the bug**
When using the UI-Component **section**, adding a **condition** to **update the setting** _type_ crashes the designer. The page needs to be reloaded afterwards.

**To Reproduce**
1. Open the design of an app
2. Adding a new **section**-component
3. Open menu-tab **Conditions**
4. Edit **Conditions**
5. Add new **Condition**
6. Change behaviour to **Update setting**
7. From the dropdown _Choose an option_ click on _type_
8. The designer crashes with the error:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'name')
    at Array.create_default_slot_3$1g (index.72fb0529.js:1707:136568)
    at create_slot (index.72fb0529.js:1:1002191)
    at create_if_block_1$2k (index.72fb0529.js:27:15498)
    at create_fragment$91 (index.72fb0529.js:27:16667)
    at init$2 (index.72fb0529.js:4:11134)
    at new ActionButton (index.72fb0529.js:27:20213)
    at create_fragment$3S (index.72fb0529.js:1707:139387)
    at init$2 (index.72fb0529.js:4:11134)
    at new SectionSelect (index.72fb0529.js:1707:140879)
    at construct_svelte_component (index.72fb0529.js:1:1007401)
```

**Expected behavior**
The option should change to _type_ and the dropdown of the possible options of _types_ should be displayed.
  
**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome 
 - Version: 118.0.5993.89

**Additional context**
The page needs to be refreshed afterwards. The bug occurs afterwards

<sub>[BUDI-7685](https://linear.app/budibase/issue/BUDI-7685/adding-update-setting-condition-to-section-component-crashes-frontend)</sub>

ConorWebb96:
Hey @LennysHead,

Would you be able to provide a video of the issue you are experiencing. I wasn't able to replicate this behaviour on version `2.11.41` of Budibase.

Can you also try updating to the latest version as it may have indirectly been fixed at some point. 

Thanks ahead of time.

andz-bb:
[...]

PR #12311, opened 2023-11-07
Fix bug in SectionSelect

aptkingston:
## Description
Fixes a bug in the SectionSelect component which manifested itself as a builder crash when using the section with conditional UI.

### Addresses
- https://github.com/Budibase/budibase/issues/12162#issuecomment-1798777993

## Feature branch env
[Feature Branch Link](http://fb-fix-section-crash.fb.qa.budibase.net)


melohagan (approved):
LGTM - thanks for the quick fix!

PR #12314, opened 2023-11-07
Ensure that the DB always returns Documents, and other type changes.

samwho:
## Description

The key change here is that the generic type returned from DB functions now enforces that `T extends Document`. This will make it more difficult to mis-use these functions, and will force us to correctly type information that comes out of the database (currently a source of `any` in our codebase).

## Feature branch env
[Feature Branch Link](http://fb-better-types-in-global-users.fb.qa.budibase.net)

### DIFF
--- a/packages/server/src/api/controllers/layout.ts
+++ b/packages/server/src/api/controllers/layout.ts
@@ -30,12 +30,12 @@ export async function destroy(ctx: BBContext) {
     layoutRev = ctx.params.layoutRev
 
   const layoutsUsedByScreens = (
-    await db.allDocs(
+    await db.allDocs<any>(
       getScreenParams(null, {
         include_docs: true,
       })
     )
-  ).rows.map(element => element.doc.layoutId)
+  ).rows.map(element => element.doc!.layoutId)
   if (layoutsUsedByScreens.includes(layoutId)) {
     ctx.throw(400, "Cannot delete a layout that's being used by a screen")
   }
--- a/packages/server/src/api/controllers/layout.ts
+++ b/packages/server/src/api/controllers/layout.ts
@@ -1,7 +1,7 @@
 import { EMPTY_LAYOUT } from "../../constants/layouts"
 import { generateLayoutID, getScreenParams } from "../../db/utils"
 import { events, context } from "@budibase/backend-core"
-import { BBContext } from "@budibase/types"
+import { BBContext, Layout } from "@budibase/types"
 
 export async function save(ctx: BBContext) {
   const db = context.getAppDB()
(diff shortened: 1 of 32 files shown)


packages/server/src/api/controllers/layout.ts
```diff
@@ -30,12 +30,12 @@ export async function destroy(ctx: BBContext) {
     layoutRev = ctx.params.layoutRev
 
   const layoutsUsedByScreens = (
-    await db.allDocs(
+    await db.allDocs<any>(
```
samwho, line 33: I wasn't sure exactly what type this should be.

packages/server/src/api/controllers/permission.ts
```diff
@@ -79,7 +79,7 @@ async function updatePermissionOnRole(
     ) {
       rolePermissions[resourceId] =
         typeof rolePermissions[resourceId] === "string"
-          ? [rolePermissions[resourceId]]
+          ? [rolePermissions[resourceId] as unknown as string]
```
samwho, line 82: This was a weird one. The type of `rolePermissions[resourceId]` is `string[]` according to `Role.permissions` but I didn't want to remove this explicit check in case it's here for older apps that might have different data to what we expect.

packages/server/src/api/controllers/view/utils.ts
```diff
@@ -53,12 +53,12 @@ export async function getViews() {
     }
   } else {
     const views = (
-      await db.allDocs(
+      await db.allDocs<any>(
```
samwho, line 56: Again, wasn't sure what type this should be.

packages/server/src/api/controllers/permission.ts
mike12345567: This is likely to be what is going on here, its probably something to do with old apps that may have had a slightly different data strategy - we could comment it for removal at some point in the future/if anyone comes across it and has a problem with it they are aware its not critical/linked to something very old/likely no longer needed.

packages/server/src/api/controllers/view/utils.ts
mike12345567: Ah - this will return something like a `DesignDocument` - but it isn't well typed right now, same goes for the `db.get<any>` just above this - i'll jump on the branch and improve this.

packages/server/src/api/controllers/layout.ts
mike12345567: This should be a `Layout` - but they are deprecated objects, not used in the platform anymore.

mike12345567 (approved):
Looks good to me - such a massive improvement to the DB layer!

Issue #12318, opened 2023-11-08
[BUDI-7741] Can't view automation history on latest version of Budibase, crashes the page and requires a hard refresh to fix.


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.13.3
  * App Version: 2.13.3

**Describe the bug**
When trying to click into any of my automation's history, nothing happens and the whole app seems to crash. The only way to fix is by hard refreshing the webpage

**To Reproduce**

1. Go to an app that has automations on the latest version of Budibase.
2. Go into the settings area then automation History.
3. Click on one of the items in the automation History.
4. Notice it will appear in a strange way, half of the side panel is cut off.
5. Refresh the page.
6. Re-click one of the automation history items in the table, then you will see the error appear in the console.
7. See that page crashes.

**Expected behavior**
You should be able to view the automations steps and preferably be able to continue to use the system without having to hard refresh the page.

**Screenshots**

```
index-e184adb4.js:1682 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'definition')
    at n1.$$.update (index-e184adb4.js:1682:75918)
    at init$1 (index-e184adb4.js:4:11078)
    at new FlowItemHeader (index-e184adb4.js:1682:76591)
    at create_if_block_1$J (index-e184adb4.js:1687:9705)
    at create_each_block$n (index-e184adb4.js:1687:15573)
    at create_fragment$2u (index-e184adb4.js:1687:16566)
    at init$1 (index-e184adb4.js:4:11134)
    at new TestDisplay (index-e184adb4.js:1687:18232)
    at create_key_block$1 (index-e184adb4.js:1687:33027)
    at Array.create_default_slot$1I (index-e184adb4.js:1687:33924)
```

**App Export**
[simple automation-export-1699430983399.tar.gz](https://github.com/Budibase/budibase/files/13293995/simple.automation-export-1699430983399.tar.gz)

**Desktop (please complete the following information):**

* OS: MacOS
* Browser Brave
* Version 1.60.110

**Additional context**
No network errors, purely a console error that prevents anything from happening once clicked.

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7741](https://linear.app/budibase/issue/BUDI-7741/cant-view-automation-history-on-latest-version-of-budibase-crashes-the)</sub>

mike12345567:
[...]

PR #12321, opened 2023-11-08
Fix for automation history issues

PClmnt:
## Description
Fixes issues where automation history blocks were cut off from the side of the screen and could potentially error due to a missing null check. #12318 

Also fixes a bug where delete could become enabled for the Trigger in certain scenarios. 

## Feature branch env
[Feature Branch Link](http://fb-fix-automation-history-error.fb.qa.budibase.net)


melohagan (approved):
LGTM

PR #12322, opened 2023-11-08
Convert user.spec.js to user.spec.ts

samwho:
## Description

Converted `user.spec.js` to `user.spec.ts`, implementing the `UserAPI` class in the process. This required giving types to all of the `/api/users/*` endpoints.


mike12345567 (approved):
Very nice - good improvement, nice to get such a core test converted!

Issue #9583, opened 2023-02-08
[BUDI-6461] Parameters Binding keyword duplicated (repeated) in REST Query URL


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.3.0
  * App Version: 2.3.0

**Describe the bug**
When focusing on a REST query URL that has query param bindings, the 'Binding' keyword is added multiple times.

**To Reproduce**

1. Add and save a GET REST query that has params with a binding value.
2. Click on the URL to focus it, and click off. See that the 'Binding' keyword is now duplicated.
3. Do this as much as you like.

**Expected behavior**
Binding keyword should only appear once in the URL for query params.

**Screenshots**

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/8ba32a25-5987-4e39-85d9-b58fcfbb3018/9766e245-b386-431a-8fd2-369545d17714?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC84YmEzMmEyNS01OTg3LTRlMzktODVkOS1iNThmY2ZiYjMwMTgvOTc2NmUyNDUtYjM4Ni00MzFhLThmZDItMzY5NTQ1ZDE3NzE0IiwiaWF0IjoxNjkyODA4NDE1LCJleHAiOjE2OTI4OTQ4MTV9.R2Sk6oIIc2vSaqgwaSv3043hMluP2OQG0RuW7rHsoII)

[...]

PR #12325, opened 2023-11-08
Fix for Parameters Binding keyword duplicated (repeated) in REST Query URL

melohagan:
## Description
If the binding is already in readable form, then don't replace it.

## Addresses
- https://github.com/Budibase/budibase/issues/9583

## Screenshots
bindings_fix


shogunpurple (commented):
Would be good to have a unit test around this to prove it out

shogunpurple (approved):
LGTM! Thanks for adding the test 🙏

PR #12330, opened 2023-11-08
Fix adding/editing relationships via grid modal (and automations)

aptkingston:
## Description
This PR fixes a few bugs with the `LinkedRowSelector` component which meant values were not properly being updated when this component was used.

This was evident by being unable to add or edit relationships when using the modal from inside the grid in the data section. The same issue could be seen when choosing related rows in some automation blocks.

## Addresses
- https://linear.app/budibase/issue/BUDI-7582/%5Bqa-wolf%5D-relationship-column-does-not-have-any-values-after-adding-a

## Feature branch env
[Feature Branch Link](http://fb-fix-modal-relationship-picker.fb.qa.budibase.net)


melohagan (approved):
LGTM!

PR #12334, opened 2023-11-08
Another random assortment of typing improvements

samwho:
## Description

Picked out a few more core places and made sure types were flowing through them correctly and explicitly. Involved introducing some more explicit `any`'s, but I figure this just makes them easier to remove later because we can properly search for them.

## Feature branch env
[Feature Branch Link](http://fb-more-db-typing-improvements.fb.qa.budibase.net)


packages/server/src/automations/triggers.ts
```diff
@@ -94,7 +94,7 @@ export async function externalTrigger(
   automation: Automation,
   params: { fields: Record<string, any>; timeout?: number },
   { getResponses }: { getResponses?: boolean } = {}
-) {
+): Promise<any> {
```
mike12345567, line 97: Would be good to type this one - but the whole async/sync thing does make it a lot more difficult

packages/server/src/threads/automation.ts
```diff
@@ -241,7 +241,7 @@ class Orchestrator {
     })
   }
 
-  async execute() {
+  async execute(): Promise<any> {
```
mike12345567, line 244: Again this would be great to type - but the way it works today its very hard to understand all of the returns - probably suggests it needs a bit of refactoring here.

mike12345567 (approved):
LGTM!

packages/server/src/threads/automation.ts
samwho: I did attempt both this one and the one above but they can return wildly different things, which the callsites don't seem to care about, they just bung the result into a `ctx.body` and call it a day.

Something we should certainly address at some point.

Issue #11741, opened 2023-09-11
[BUDI-7475] Select / deselect all fields in blocks


melohagan:
**Describe the feature request**
When a form has a lot of fields and you want to use a couple of required fields, it is fairly tedious to unselect all fields.

I would like to be able to unselect all fields, and then select the fields that I want. Similar to the fetch tables selection.

**Screenshots**
![Screenshot 2023-09-11 at 13 43 16](https://github.com/Budibase/budibase/assets/101575380/a775cbe6-3ac5-47ca-a219-f505e84c19da)

*This example isn't the worst case, some tables could have 20+ fields*

<sub>[BUDI-7475](https://linear.app/budibase/issue/BUDI-7475/select-deselect-all-fields-in-blocks)</sub>

melohagan:
A `Show all` / `Hide all` button exists in the Data section that could be replicated:

![Screenshot 2023-09-11 at 17 11 16](https://github.com/Budibase/budibase/assets/101575380/08e3286e-95d0-416b-b467-e4d2b1852608)

PR #12336, opened 2023-11-08
Added toggle-all functionality to the FormBlock

deanhannigan:
## Description

All fields in the `FormBlock` can now be toggled at the same time.

- There was a small adjustment to the design as sections are now collapsible 
~~A new `settings` context was added to the `ComponentSettingsSection` where save requests can be tracked. This can be used by any setting type to allow custom behaviour when updating. In this instance, the fields are briefly disabled while the toggle request is processed.~~

## Addresses
- https://github.com/Budibase/budibase/issues/11741

## Feature branch env
[Feature Branch Link](http://fb-feature-toggle-all-formblock-fields.fb.qa.budibase.net)


packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -43,6 +49,10 @@
     cachedValue = cloneDeep(value)
   }
 
+  $: if (typeof $updates.resp == "string") {
```
melohagan, line 52: NAB: How would this differ from:

```js
if ($updates.resp) {
  updating = false
}
```

melohagan (approved):
LGTM - handy feature! 😄 

I was wondering why `CellEditor` and `ColumnDrawer` have been removed?

deanhannigan:
> LGTM - handy feature! 😄
> 
> I was wondering why `CellEditor` and `ColumnDrawer` have been removed?

Hey Mel, they were a hangover from the previous FieldConfiguration UX and they are no longer being used anywhere.

melohagan:
> > LGTM - handy feature! 😄
> > I was wondering why `CellEditor` and `ColumnDrawer` have been removed?
> 
> Hey Mel, they were a hangover from the previous FieldConfiguration UX and they are no longer being used anywhere.

Cool, just double checking 👍

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
deanhannigan: They achieve the same thing, but the `typeof` just communicates the contains of `resp` better

aptkingston:
Looks great! Just was wondering around the new settings context and the flashing disabled state - what was the motivation for adding this? Disabling things while changes are in progress isn't a pattern we use elsewhere in settings, and I'm just not sure if the additional complexity is worth it. It also looks a little bit flaky since it's just checking for the existence of a string value to be able to tell when the transaction completes.

Basically I think I'd just prefer to leave out the disabled state stuff unless there's a good reason to add it, which there very well might be!

deanhannigan:
@aptkingston I added it after testing the toggle-all in slow mode in chrome. As the toggle-all request was pending I was still  able to drag fields around and edit them. As the requests completed, the fields began jumping around and I thought it may be a bit confusing. Bit of an extreme use-case but, as you say though, it's not a current patten so I'll revert the changes. 👍

packages/builder/src/components/design/settings/controls/EditComponentPopover.svelte
```diff
@@ -90,6 +91,7 @@
       open = true
     }
   }}
+  {disabled}
```
aptkingston, line 94: Think we can remove this disabled prop now.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldConfiguration.svelte
```diff
@@ -145,16 +148,35 @@
...
 <div class="field-configuration">
+  <div class="toggle-all">
+    <span />
+    <Toggle
+      on:change={() => {
+        let update = fieldList.map(field => ({
+          ...field,
+          active: selectAll,
+        }))
+        toggleAll(update)
```
aptkingston, line 170: We should either remove the `toggleAll` function now (since its a straight pass-through to `listUpdate`) or better yet in my opinion, move the inline logic here into toggle all and keep it.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldSetting.svelte
```diff
@@ -49,6 +50,7 @@
       {bindings}
       {parseSettings}
       on:change
+      disabled={updating}
```
aptkingston, line 53: We can remove the `disabled` prop now since we can also remove this `updating` prop.

packages/builder/src/components/design/settings/controls/FieldConfiguration/FieldSetting.svelte
```diff
@@ -58,7 +60,13 @@
     <div class="field-label">{readableText}</div>
   </div>
   <div class="list-item-right">
-    <Toggle on:change={onToggle(item)} text="" value={item.active} thin />
+    <Toggle
+      on:change={onToggle(item)}
+      text=""
+      value={item.active}
+      thin
+      disabled={updating}
```
aptkingston, line 68: We can remove this `disabled` prop.

aptkingston (approved):
LGTM!

All comments are just leftover things that can be removed now, but no blockers. Moving this to QA since its functionally complete, but would be good to do that little bit of clean up 👍

PR #12346, opened 2023-11-09
Prevent sending invites to users who have already been invited.

samwho:
## Description

This PR modifies the `/api/global/users/multi/invite`, `/api/global/users/invite`, and `/api/global/users/onboard` endpoints to reject sending invites to an email address if there is already an invite pending.

Additionally, I've moved the functionality in `packages/worker/src/utilities/redis.ts` to `packages/backend-core` so that it could be used in the bulk user create code. A notable change here is lazily initialising the Redis clients, and not having an explicit shutdown for them anymore. I am uncertain if this will cause problems.

## Addresses
- https://linear.app/budibase/issue/BUDI-7008/i-was-able-to-send-two-invitations-to-the-same-user-email-address-one


packages/backend-core/tests/core/utilities/structures/users.ts
```diff
@@ -12,7 +12,7 @@ import { generator } from "./generator"
 import { tenant } from "."
 
 export const newEmail = () => {
-  return `${uuid()}@test.com`
```
samwho, line 15: `test.com` is a private domain owned by who-knows-who. `example.com` is owned by IANA and is specifically reserved for documentation purposes in [RFC2606](https://datatracker.ietf.org/doc/html/rfc2606). I always prefer using these domains, in case anything should happen and we accidentally send emails to these domains.

packages/backend-core/src/redis/invite.ts
```diff
@@ -0,0 +1,96 @@
...
+
+let client: redis.Client
+
+async function getClient(): Promise<redis.Client> {
+  if (!client) {
+    client = new redis.Client(redis.utils.Databases.INVITATIONS)
+    await client.init()
+  }
+  return client
+}
```
samwho, line 23: This Redis client used to get set up in an explicit `init()` function, and torn down in an explicit `shutdown()` function that were part of the `packages/worker` start up and shutdown code. I'm lazily initialising them here and not shutting them down, depending instead on the process death to get rid of them. Not sure if this is problematic.

packages/backend-core/src/users/lookup.ts
```diff
@@ -26,6 +27,9 @@ export async function searchExistingEmails(emails: string[]) {
   const existingAccounts = await getExistingAccounts(emails)
   matchedEmails.push(...existingAccounts.map(account => account.email))
 
+  const invitedEmails = await getExistingInvites(emails)
+  matchedEmails.push(...invitedEmails.map(invite => invite.email))
```
samwho, line 31: This is the key change that fixes the bug. It was wanting to do this here that made me move the invite / password reset functionality into `backend-core`.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -249,59 +251,35 @@ export const tenantUserLookup = async (ctx: any) => {
...
-        password,
-        forceResetPassword: true,
-        roles: invite.userInfo.apps,
-        admin: invite.userInfo.admin,
-        builder: invite.userInfo.builder,
-        tenantId: tenancy.getTenantId(),
-      }
-    })
-    let bulkCreateReponse = await userSdk.db.bulkCreate(users, [])
+export const onboardUsers = async (
```
samwho, line 254: I've simplified this controller substantially. `create` was never a property in the request as far as I can tell.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -328,33 +306,31 @@ export const invite = async (ctx: Ctx<InviteUserRequest>) => {
...
-  const request = ctx.request.body
-  ctx.body = await userSdk.invite(request)
+  ctx.body = await userSdk.invite(ctx.request.body)
 }
 
 export const checkInvite = async (ctx: any) => {
   const { code } = ctx.params
   let invite
   try {
-    invite = await checkInviteCode(code, false)
```
samwho, line 339: I renamed this function to `getInviteCode` to better reflect what it actually does, and removed the functionality within it to delete tokens. It felt dangerous for a function called `checkInviteCode` to, unless you pass it `false` as a second argument, delete data. There's only one place in the codebase where this delete functionality is used, and it now calls a delete function directly.

packages/worker/src/sdk/auth/auth.ts
```diff
@@ -73,7 +73,7 @@ export const reset = async (email: string) => {
  * Perform the user password update if the provided reset code is valid.
  */
 export const resetUpdate = async (resetCode: string, password: string) => {
-  const { userId } = await redis.checkResetPasswordCode(resetCode)
+  const { userId } = await redis.passwordReset.getResetPasswordCode(resetCode)
```
samwho, line 76: This appears to be the only place in the codebase that fetches the password reset code, and it immediately discards the `info` portion. If the `info` isn't used, should we get rid of it?

packages/backend-core/src/redis/invite.ts
mike12345567: Generally a quite like lazy initing of resources like this - the services will always confirm that the redis parameters are correct on startup, so therefore it'll be safe for this connection to occur when needed.

packages/worker/src/api/controllers/global/users.ts
mike12345567: I'm happy enough with this - to be honest the only real thing you need to test is the user onboarding in the builder, if you can go to the user section and add multiple users in one go then the onboarding should be working as expected - it has no other use cases today.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -1,8 +1,9 @@
 import {
-  checkInviteCode,
+  getInviteCode,
+  deleteInviteCode,
   getInviteCodes,
   updateInviteCode,
-} from "../../../utilities/redis"
+} from "@budibase/backend-core/src/redis/invite"
```
mike12345567, line 6: We can't use import statements like this sadly, it generally baffles esbuild and can cause weird issues in production - for that reason we don't really allow import statements like this for monorepo libraries.

You can instead import it through the default import like `import { redis } from "@budibase/backend-core"` then use the commands like `redis.checkInviteCode`

packages/worker/src/index.ts
```diff
@@ -88,7 +86,6 @@ const shutdown = () => {
 export default server.listen(parseInt(env.PORT || "4002"), async () => {
   console.log(`Worker running on ${JSON.stringify(server.address())}`)
   await initPro()
-  await redis.init()
```
mike12345567, line 91: Ah - this could be an issue, as mentioned earlier this will mean that there are no redis parameters being checked until a code is used - meaning that the worker could be mis-configured but it won't be flagged up until a code is needed.

Its sort of off topic, but it could be worth adding redis management to the worker, which purely calls through to init/shutdown commands in the backend-core redis parts that the worker needs - purely so that we can maintain similar functionality to what we have today.

packages/worker/src/sdk/users/users.ts
```diff
@@ -14,11 +18,13 @@ export async function invite(
   const matchedEmails = await usersCore.searchExistingEmails(
     users.map(u => u.email)
   )
-  const newUsers = []
+  const newUsers: InviteUserRequest[] = []
 
   // separate duplicates from new users
   for (let user of users) {
     if (matchedEmails.includes(user.email)) {
+      // This "Unavailable" is load bearing. The tests and frontend both check for it
```
mike12345567, line 26: This is sadly true in a few places - there are a few tests that utilise events to check whether certain outcomes have occurred, rather than using API responses - but luckily there isn't too many cases like this.

packages/worker/src/utilities/email.ts
```diff
@@ -3,7 +3,8 @@ import { EmailTemplatePurpose, TemplateType } from "../constants"
 import { getTemplateByPurpose, EmailTemplates } from "../constants/templates"
 import { getSettingsTemplateContext } from "./templates"
 import { processString } from "@budibase/string-templates"
-import { getResetPasswordCode, getInviteCode } from "./redis"
+import { createResetPasswordCode } from "@budibase/backend-core/src/redis/passwordReset"
```
mike12345567, line 6: The same as the point made before - will need to use `import { redis } from "@budibase/backend-core"` here instead

mike12345567 (changes requested):
Looks good - only real major thing that will need changed is the import statements as these can cause some weird bugs to diagnose - we really should work out a linting rule to deny these type of statements completely - I'll create a ticket for that.

EDIT: ticket: https://linear.app/budibase/issue/BUDI-7749/add-linting-rule-to-deny-direct-import-statements-from-monorepo

packages/worker/src/index.ts
samwho: It's easy enough to re-instate the `init`/`shutdown` functions, there will just be 2 of them (one for invites, one for password reset). If that's no bother, I'll go do it.

mike12345567: yeah - I think we probably should do that - expose the ability to init/shutdown just for feature parity.

mike12345567 (approved):
LGTM - thanks for the changes :)

samwho:
Thanks for the review! 🙇

samwho:
@mike12345567 I don't know exactly why, but the changes I made to bring worker inline with server re: Redis initialisation was causing a bunch of weird failures in tests. I've reverted back to how it was prior to this PR and things are passing now. Will wait until after we're back to merge.

PR #12350, opened 2023-11-09
Settings enhancements

aptkingston:
## Description
This PR adds the ability to better utilise actions and context provided by block components.

With these changes you can now do things like:
- Reference actions provided by components inside blocks
- Configure button actions that depend on certain component types (like forms) to target blocks

Some use cases for these changes are:
- Trigger a refresh of a grid block data by adding a button to it
- Be able to add buttons to a form block, and use all form relation actions

I've gone ahead and added some additional context and actions to the existing blocks. Form blocks now provide all form related actions, and the grid block exposes a "refresh datasource" action.

The primary motivation for these changes was to enable the upcoming feature of adding fully configurable buttons to the form block, which previously was not possible as it was not able to expose bindings or context to itself.

## Addresses
- https://linear.app/budibase/issue/BUDI-7740/research-spike-exposing-block-component-context

## Feature branch env
[Feature Branch Link](http://fb-settings-enhancements.fb.qa.budibase.net)


deanhannigan (approved):
LGTM!

Issue #12154, opened 2023-10-23
[BUDI-7683] There was a problem loading your grid - App users


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.11.41
  - App Version: 2.11.41 & tested with reverted 2.11.6

**Describe the bug**
When I try to view the App users table in the data view of builder it shows the message below and no data is shown
```
There was a problem loading your grid
Cannot read properties of null (reading 'sourceId')
```
![image](https://github.com/Budibase/budibase/assets/110813989/3f0b5916-e76f-4cb5-9555-73c8485385a2)

Additionally in the front end where the a table component previously did show the application users it now shows
![image](https://github.com/Budibase/budibase/assets/110813989/2b48ba58-d024-41db-8218-53ae493dfdfe)

I think this is applicable to an application which did previously have additional fields added to the users table

**To Reproduce**
I can send you an export of the application to reproduce. Just let me know who to send it to.

**Expected behavior**
I should be able to view users in both views.

<sub>[BUDI-7683](https://linear.app/budibase/issue/BUDI-7683/there-was-a-problem-loading-your-grid-app-users)</sub>

aptkingston:
Could you check if the error is returned by the API or not? If you check dev tools you'll be able to see if the network request fails or not. That will help us narrow down where the error is coming from.

Is this also just this one particular app or all apps? I've tested plenty of apps with and without the old users columns and they work fine, so it must be something else unique to this app.

andyburgessmd:
Builder console shows
![image](https://github.com/Budibase/budibase/assets/110813989/ab50e109-127a-40c4-b631-c897ffd8bab6)

And the application console is 
![image](https://github.com/Budibase/budibase/assets/110813989/2a2f4c93-66a4-417c-97c2-67e34aa8bba2)

aptkingston:
Thanks for the extra info!

andyburgessmd:
This may be low priority to you but for me its critical being able to use an application or make any changes to it.

aptkingston:
We're actively testing a new option that will allow you to migrate a relationship to the users table into the new "user type" column, and it's almost ready for release. By marking this low priority we're just saying that we aren't going to make any other changes to the existing user relationship codebase (which is the cause of this problem) because we've nearly got this other solution ready.

[...]

PR #12354, opened 2023-11-09
Fix issues with data UI loading some tables

mike12345567:
## Description
We have added some new mechanisms for fetching tables - in some edge case scenarios it seems that tables are fetched that don't exist. If this occurs, due to say a bad relationship column, we should still allow the tables to be retrieved so that they can be fixed.

I have added a new function to the DB layer `getMultiple` which can be used to bulk fetch documents by their ID - previously this was configured in lots of different places throughout the app, all doing the same job. This improves the typing management of these function calls and allows handling the error that occurs for each of these events. Due to the way the `allDocs` call works, if fetching by an ID isn't successful it gives us an error structure rather than erroring - this new function handles this and only returns the successfully fetched objects - it also makes sure it is only fetching the unique list for performance reasons.

## Addresses
- #12154 

## Feature branch env
[Feature Branch Link](http://fb-fix-12154-invalid-table-fetches.fb.qa.budibase.net)


packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -109,14 +109,35 @@ export class DatabaseImpl implements Database {
...
     const db = await this.checkSetup()
     if (!id) {
       throw new Error("Unable to get doc without a valid _id.")
     }
     return this.updateOutput(() => db.get(id))
   }
 
+  async getMultiple<T extends Document>(
+    ids: string[],
+    opts?: { failIfMissing?: boolean }
```
samwho, line 122: nit: `failIfAnyMissing` would be a slightly clearer name, but I wouldn't block this PR on it.

samwho: I also feel slightly stronger that the default should be `true`. It would surprise me if the majority of places this is used would be fine if they got an empty array back.

samwho (changes requested):

packages/backend-core/src/db/couch/DatabaseImpl.ts
mike12345567: funnily enough - the expectation of everywhere this is used today is to fetch what they can - but I do agree it would be better to be flipped, I'll set that and then add the option to all current calls.

samwho (approved):

PR #12355, opened 2023-11-09
Fix issue in relationship definitions where primary / foreign key selection didn't trigger properly

PClmnt:
## Description
#BUDI-7729

Fixes an issue that when you select a One to Many relationship after having previously selected Many to Many or Many to One the Primary Key and Foreign Key options wouldn't appear. Just a reactivity bug and not accounting for the in the relationship map. 

## Feature branch env
[Feature Branch Link](http://fb-fix-relationship-selection-issues.fb.qa.budibase.net)


packages/builder/src/components/backend/Datasources/CreateEditRelationship.svelte
```diff
@@ -33,6 +33,10 @@
       part1: PrettyRelationshipDefinitions.MANY,
       part2: PrettyRelationshipDefinitions.ONE,
     },
+    [RelationshipType.MANY_TO_ONE]: {
```
mike12345567, line 36: Should this be `ONE_TO_MANY` as many to one is specified above?

mike12345567 (approved):

PR #12356, opened 2023-11-09
Add search capabilities to data section

adrinr:
## Description
Adding the search functionallity to datasources, following the pattern we have for screens and apps.
Few bugs fixed along the way, such as allowing visiblility toggles when a selected datasource if chevron clicked or adding some animations

## Addresses
- [BUDI-7680 - Data section: Add search to data sources
](https://linear.app/budibase/issue/BUDI-7680/data-section-add-search-to-data-sources)

### No filter applied
image

### Filtering (with the users table in the results)
image

### Filtering (without the users table in the results)
Queries, tables and views are displayed
image

### Filtering with whitespace
image

### No results
image

## Feature branch env
[Feature Branch Link](http://fb-budi-7680-data-section-add-search-to-data-sources.fb.qa.budibase.net)


mike12345567 (approved):
LGTM - would check with @aptkingston to confirm as well.

packages/builder/src/components/backend/DatasourceNavigator/DatasourceNavigator.svelte
```diff
@@ -228,6 +129,13 @@
         {/each}
       {/if}
     {/each}
+    {#if showNoResults}
+      <Layout paddingY="none" paddingX="L">
+        <div class="no-results">
+          There aren't any datasource matching '{searchTerm}'
```
aptkingston, line 135: Could we update this copy to be more similar to the design section? Maybe just "There aren't any sources matching that name". We could also change the colour to grey to match the design section empty state too.

aptkingston (commented):
Works perfectly, nice work @adrinr! Would it be easy to update the look of this left panel to match the design section? By that I mean removing the divider (until scrolled) and removing the additional padding.

This is the spacing I mean:
image

Which would make it look more like:
image

Where the first item starts immediately after the header.

In the design section the divider only appears after scrolling, but in the portal app list there is never a divider. So I'm less opinionated on what happens with the divider, but ideally we could at least remove the spacing. If we can add a divider only on scroll then that's a bonus!

adrinr:
> Works perfectly, nice work @adrinr! Would it be easy to update the look of this left panel to match the design section? By that I mean removing the divider (until scrolled) and removing the additional padding.
> 
> This is the spacing I mean: image
> 
> Which would make it look more like: image
> 
> Where the first item starts immediately after the header.
> 
> In the design section the divider only appears after scrolling, but in the portal app list there is never a divider. So I'm less opinionated on what happens with the divider, but ideally we could at least remove the spacing. If we can add a divider only on scroll then that's a bonus!

I'll have a look at it. I did not want to do many design changes, and the screens has a different behaviour as there are 2 different panels in it

packages/builder/src/components/backend/DatasourceNavigator/DatasourceNavigator.svelte
adrinr: Good catch!

adrinr:
Done @aptkingston. I also changed the screens header to use the new generic one

aptkingston (approved):
Thanks @adrinr! Looks great 👌 

If we could still do a minor copy update here
image

from

> There aren't any datasource matching that route

to

> There aren't any datasources matching that name

Then all good from my perspective!

PR #12383, opened 2023-11-16
Field help text

Ghrehh:
## Description
Add a "help text" field to our form fields.

Given the copy of the ticket, I assume this was for the fields in the client, which use slightly different components than what are used in the builder, though the look very similar. I just added a version of this to both the client and builder to be safe.

## Addresses
https://linear.app/budibase/issue/BUDI-7539/add-a-new-help-text-setting-to-form-fields


deanhannigan:
@Ghrehh minor thing, but we may need to confirm the setting field ordering. In the designs, the `Help text` field comes after `Default value`

deanhannigan (approved):
LGTM! I don't see an issue with adding the client fields into the mix.

aptkingston (commented):
This is specifically for client components, as part of the form improvements project, but no harm in having it available in the builder too!

I would agree with @deanhannigan that we should either remove the icon for help text or (better yet) add in an icon for errors now too, to keep things consistent.

aptkingston (approved):
LGTM!

NAB, but could we increase the space between the icons (both help and error) and the text from 4px to 6px to match designs? I think it's important for very common UI like this.

PR #12384, opened 2023-11-16
Disable preview and publish for apps with no screens

Ghrehh:
## Addresses
https://linear.app/budibase/issue/BUDI-7420/disable-both-preview-and-publish-until-you-have-at-least-1-screen


melohagan (approved):
LGTM! 

Just as a sanity check - `$sortedScreens` will never be null/undefined ?

aptkingston:
These changes look good! There's another part to this ticket though which is preventing users from being able to preview the app using the new side panel list of apps in the portal. Right now you can still preview an app that way and it'll say you don't have access if there are no screens.

I think we should show a custom message instead of the iframe preview for those apps, just saying that the app has no screens. Probably go for some similar empty-state placeholder message with a large icon.

Ghrehh:
@aptkingston Good shout, there's a screengrab of what I implemented, you can let me know if it looks alright:
Screenshot 2023-11-17 at 13 55 30

@melohagan Double checked there, and it definitely seems to never be `null`.

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -14,6 +17,18 @@
...
     return `/${app.devId}`
   }
+
+  // Normally fetched in builder/src/pages/builder/app/[application]/_layout.svelte
+  const fetchScreens = async appId => {
+    if (!appId) return
+
+    store.actions.reset()
+    const pkg = await API.fetchAppPackage(appId)
+    await store.actions.initialise(pkg)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:27]: Rather than initialise the store from the portal here, which causes a pretty large cascade of requests and hydration logic, we can just check the length of `pkg.screens` directly. Much simpler and achieves the same thing.

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -45,14 +60,26 @@
...
       Fullscreen
     </ActionButton>
   </div>
-  <iframe src={iframeUrl} title={app.name} />
+  {#if noScreens}
+    <div class="noScreens">
+      <ErrorSVG />
+      <Body>You haven't added any screens to your app yet.</Body>
+      <Body
+        ><a href={$url(`/builder/app/${app.devId}/design`)}>Click here</a> to add
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:76]: We should use the `Link` component here to keep consistent with the rest of the platform. That will also style the link blue rather than indigo, which I think is the safer choice until we properly move our accent color over.

aptkingston (approved):
LGTM! 2 small changes suggested but good to go once those are in.

PR #12388, opened 2023-11-17
Removing references to budibase packages usages without barrel imports

adrinr:
## Description
Adding a rule to ensure that we don't import workspace packages using the relative files. This will break execution, as we are bundling and flattening the packages using esbuild.
There are primarily 2 changes:
1. Installing and using the `import` plugin. This will be used to ensure that a package is not using another one via relative imports
2. Setting up a new custom rule that does not allow importing `@budibase/package/whatever`, forcing to use the root `import @budibase/package` instead. We are using `eslint-plugin-local-rules` for it, as we cannot rely on the `import/no-internal-modules`. We need to allow certain ones (svelte, dayjs, lodash...) but not the budibase ones, with some exceptions. Its configuration is not smart enough to support this easily, so it's easier to create a custom rule instead


packages/client/src/components/app/forms/S3Upload.svelte
```diff
@@ -2,7 +2,7 @@
   import Field from "./Field.svelte"
   import { CoreDropzone, ProgressCircle } from "@budibase/bbui"
   import { getContext, onMount, onDestroy } from "svelte"
-  import { cloneDeep } from "../../../../../bbui/src/helpers"
+  import { cloneDeep } from "@budibase/bbui/src/helpers"
```
mike12345567, line 5: is this one ok?

mike12345567 (approved):
Excellent - fantastic job getting this sorted - great to get a linting rule in place!

packages/client/src/components/app/forms/S3Upload.svelte
adrinr: Yup, as it is not build via esbuild. But I just cleaned it :)

PR #12390, opened 2023-11-17
Fix for cyclics causing test failures in master

mike12345567:
## Description
Moving things around to get rid of cyclics created by moving invite/passwordReset into backend-core - I haven't updated the test case that triggered this - as the test case has helped detect a weird issue, may as well keep it as is.


packages/backend-core/src/cache/invite.ts
```diff
@@ -1,7 +1,10 @@
-import { utils, tenancy, redis } from "../"
+import * as utils from "../utils"
+import { Duration, DurationType } from "../utils"
 import env from "../environment"
+import { getTenantId } from "../context"
+import * as redis from "../redis/init"
 
-const TTL_SECONDS = 60 * 60 * 24 * 7
+const TTL_SECONDS = Duration.fromDays(7).to(DurationType.SECONDS)
```
samwho, line 7: <3

Could we create helpers like `toSeconds` to ease writing this? `Duration.fromDays(7).toSeconds()`?

packages/backend-core/src/redis/init.ts
```diff
@@ -7,7 +7,9 @@ let userClient: Client,
   cacheClient: Client,
   writethroughClient: Client,
   lockClient: Client,
-  socketClient: Client
+  socketClient: Client,
+  inviteClient: Client,
+  passwordResetClient: Client
```
samwho, line 12: <3

packages/worker/src/index.ts
```diff
@@ -89,8 +88,8 @@ const shutdown = () => {
 export default server.listen(parseInt(env.PORT || "4002"), async () => {
   console.log(`Worker running on ${JSON.stringify(server.address())}`)
   await initPro()
-  await redis.invite.init()
-  await redis.passwordReset.init()
+  await redis.clients.getInviteClient()
```
adrinr, line 91: This might be confusing. Could we not keep the init function, and to call the get clients inside? It's strange that on app limit we are getting the clients to then not use it.

samwho (approved):
Thank you for fixing this 🙏

packages/worker/src/index.ts
mike12345567: fair - I'll just expose the init

PR #12398, opened 2023-11-20
Setup OSS checks

adrinr:
## Description
Setting up recurrent checks that will run the CI pipeline simulating they are OSS. This will allow detecting pipeline and build issues for users without the pro submodules. This is setup to run on weekdays at 8am and 4pm, notifying to #engineering if it fails.

image

Also, disabling FB releases for opensource, as the way we handle the action call is making it error


mike12345567 (approved):
LGTM - good idea to check this twice daily!

Issue #12400, opened 2023-11-20
New REST query - undefined param added when entering the URL


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: bb.budibase.app

**Describe the bug**
An **undefined** param is added when typing in a URL.

**To Reproduce**
1. Create a new REST query
2. Type into the URL text bar
3. See that an `undefined` param is added. It can be removed.

**Expected behavior**
When entering a REST query URL, undefined parameters should not be added.

**Screenshots**
![Screenshot 2023-11-20 at 13 50 54](https://github.com/Budibase/budibase/assets/101575380/35b670a6-c9a8-46ec-883b-194b218006f0)

PR #12404, opened 2023-11-20
Don't encode undefined in REST query string

melohagan:
## Description
Fix for query having initial undefined param

## Addresses
- https://github.com/Budibase/budibase/issues/12400

## Feature branch env
[Feature Branch Link](http://fb-fix-rest-query-string-params.fb.qa.budibase.net)


PClmnt (approved):
LGTM!

PR #12411, opened 2023-11-21
Move from an allow list to a block list of file extensions.

samwho:
## Description

@melohagan brought this up as probably not the best way to do this after some customers asked about what types of file are allowed. We've decided to move to a list of specifically disallowed file types, to better represent the intent (to stop people distributing potentially harmful content through Budibase).

He also pointed out the check didn't compensate for different casing, so I've also fixed that.

Extension list is based on https://www.lifewire.com/list-of-executable-file-extensions-2626061, plus some more we thought made sense (e.g. archive types).


packages/shared-core/src/constants.ts
```diff
@@ -96,45 +96,55 @@ export enum BuilderSocketEvent {
...
+  "shs",
+  "u3p",
+  "vb",
+  "vbe",
+  "vbs",
+  "vbscript",
+  "workflow",
+  "ws",
+  "wsf",
+  "wsh",
```
melohagan, line 149: Could you add the archive extensions to this list as well, such as zip and tar

melohagan (changes requested):
One comment, otherwise good to go!

mike12345567:
Where did the invalid list come from? Looking at it there is a lot of file types that are missing that we found within the full list of attachments, for example `.php` or `.js` - we thought that code file types could be considered malicious and end up with us getting into a similar situation as happened before with a virus checker flagging one of our URLs.

mike12345567 (approved):
LGTM!

melohagan (approved):

Issue #11535, opened 2023-08-16
SQL Server: Filtering against Date/time does not work for custom queries


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.9.5
  - App Version: 2.9.5

**Describe the bug**
Discussed here: https://github.com/Budibase/budibase/discussions/11357

Essentially a custom sql server query that has date/time field cannot be filtered via more than/less than in the Design section. 

**To Reproduce**
1. Add a SQL Server custom query that returns a **date** field
2. Add a Data provider/Table in a screen for the read SQL query
3. Add a more than filter against the date field of the query compared against a bound Date Picker binding
4. See that the filter does nothing

**Expected behavior**
Data provider filters should work for SQL Server custom queries against date/time fields.

**Screenshots**
![Screenshot 2023-08-16 at 09 30 31](https://github.com/Budibase/budibase/assets/101575380/512eba76-9625-46ff-a491-3c9b63973ab1)

![Screenshot 2023-08-16 at 09 29 12](https://github.com/Budibase/budibase/assets/101575380/ad390c4f-6abf-4319-b742-16297b5c4d1e)

PR #12413, opened 2023-11-21
Handle dates in query range filter

melohagan:
## Description
The custom query filter has been updated to handle date strings.

## Addresses
- https://github.com/Budibase/budibase/issues/11535

## Screenshots
Screenshot 2023-11-21 at 12 03 58


samwho:
Would it be possible to add a test that prevents this regressing in future? 🙏

packages/shared-core/src/filters.ts
```diff
@@ -302,11 +302,16 @@ export const runLuceneQuery = (docs: any[], query?: SearchQuery) => {
       docValue: string | number | null,
       testValue: { low: number; high: number }
     ) => {
+      if (docValue == null || docValue === "") {
+        return true
+      }
+      if (!isNaN(+docValue)) {
+        return +docValue < testValue.low || +docValue > testValue.high
+      }
+      // if the string is NaN, then assume date
```
adrinr, line 311: Could we actually check if it's a valid date ISO? Otherwise, it should not work and catch the error

samwho (approved):
Superb tests. Thank you so much for writing them 🙏

PR #12416, opened 2023-11-21
Make our helm chart work out of the box

samwho:
## Description

I tried to get our existing Helm chart running in my home Kubernetes cluster and encountered a number of problems. 

1. The bundled nginx-ingress-controller conflicted with my existing ingress controller and prevented the chart from installing.
2. The bundled Clouseau image did not have any ARM builds, so I had to swap it out for one that did.
3. The `_users` database in CouchDB is not created automatically, I had to manually create it after installing the chart.
4. The `by_email2` and `by_link` views that Budibase attempts to create in the onboarding both failed to create because the bundle CouchDB runs version 3.1.1, which uses Spidermonkey 1.8.5 as its JS engine. This JS engine is very old, and couldn't run the JS those views required.
5. The Ingress resource that the CouchDB Helm chart we depend on tries to create uses the networking/v1beta1 Ingress, which is no longer supported in recent versions of Kubernetes.
6. Attempting to update to a newer CouchDB image, e.g. 3.2.1 as used by the `budibase/budibase` image, worked initially but then any attempt to communicate with Clouseau failed because the version of Erlang CouchDB 3.2.1 uses differs too much from what Clouseau 2.17.0 expects.
7. The `hpa.yaml` in this chart doesn't correctly reference any `Deployments`, so functionally doesn't do anything.

This PR is an attempt to make the Helm chart a much more out-of-the-box experience. I have implemented the following changes:

1. Removed the bundled nginx-ingress-controller. I'll need to update the docs to help people install their own ingress controller if they do not have one, but most clusters will already be running one.
2. Updated the version of the CouchDB Helm chart we're using, and switched to using our own `budibase/couchdb` image. This solves the CouchDB Ingress resource being outdated, the lack of an ARM build for the Clouseau container (we bundle Clouseau inside of `budibase/couchdb`), and our version is 3.2.1 which uses a much more recent JS engine, fixing the problems with the views.
3. Removed `hpa.yaml` and replaced it with 3 new `HorizontalPodAutoscaler` definitions targeting the apps, worker, and proxy pods.

Additionally, I've removed a bunch of the kompose annotations that aren't adding value, updated the `values.yaml` to remove some unused config, and also updated the values under the `couchdb:` key to be in line with version 4.3.0 of the CouchDB chart.

With these changes, it's possible to bring up Budibase in my home Kubernetes cluster with only the following `values.yaml`:

```yaml
couchdb:
  persistentVolume:
    enabled: true
    storageClass: nfs-client
ingress:
  className: nginx
  enabled: true
  hosts:
  - host: <domain>
    paths:
    - backend:
        service:
          name: proxy-service
          port:
            number: 10000
      path: /
      pathType: Prefix
services:
  objectStore:
    storageClass: nfs-client
  redis:
    storageClass: nfs-client
```

Which is mostly setting up the Ingress and the persistent volumes to write to my NAS.

## Addresses
- https://linear.app/budibase/issue/BUDI-7754/make-our-helm-chart-work-out-of-the-box
- https://linear.app/budibase/issue/BUDI-7755/improve-helm-chart-documentation

## Feature branch env
[Feature Branch Link](http://fb-budi-7754-make-our-helm-chart-work-out-of-the-box.fb.qa.budibase.net)


charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -200,7 +194,7 @@ spec:
         {{ end }}
 
         image: budibase/apps:{{ .Values.globals.appVersion | default .Chart.AppVersion }}
-        imagePullPolicy: Always
```
shogunpurple, line 203: This might break QA because we push a `develop` tag there - and a rollout restart won't pull the latest then. Tbf we should just fix that

charts/budibase/values.yaml
```diff
@@ -41,14 +36,10 @@ service:
...
   enabled: true
-  aws: false
-  nginx: true
-  certificateArn: ""
   className: ""
   annotations:
-    kubernetes.io/ingress.class: nginx
-    nginx.ingress.kubernetes.io/client-max-body-size: 150M
-    nginx.ingress.kubernetes.io/proxy-body-size: 50m
+    # nginx.ingress.kubernetes.io/client-max-body-size: 150M
```
shogunpurple, line 41: can probably just remove these altogether if there's no NGINX ingress

charts/budibase/values.yaml
```diff
@@ -331,31 +344,65 @@ couchdb:
   ## `OrderedReady`
   podManagementPolicy: Parallel
 
+  ## To better tolerate Node failures, we can prevent Kubernetes scheduler from
```
shogunpurple, line 347: Given that a vast majority of users (I'd say a tentative 99%) won't touch almost any of the options from this point on, I'm wondering:

Should we continue to use the official couchDB helm chart at all? The couchDB team recommend that you don't cluster couch in K8S/docker - and even in our prod we use ansible provisioned EC2 instances. I'm thinking we may actually be better to only provide single node couchDB by default in our chart for simplicity, with the option to scale out the cluster if someone really desires it. Should we just hide all of the complex couch options in another values file, and show users how to pull them in or "merge" them into this chart if needed?

charts/budibase/templates/app-service-deployment.yaml
samwho: Whoops, good catch. I changed this to make iterating on my own cluster faster. Put it back to how it was. 🙏

charts/budibase/values.yaml
samwho: I was in two minds about this. I'm guessing this is here purely for attachment uploading? If so, and Budibase will function just fine without them, I'll remove them.

charts/budibase/values.yaml
samwho: I think we _should_ keep using the official CouchDB chart. It offers a lot of extras that users may want, such as monitoring through Prometheus, placement management, pod disruption budgets, etc. I've set the cluster size to 1 (away from the default of 3) to keep it in line with how the chart works currently. Clustering does work if you want to use it, and I suspect there will be customers out there that may want to.

If we want our values.yaml to be easier to navigate, we can remove anything that's a CouchDB default (which is most of this stuff) and point people over to their chart in our docs. I only put it in here for documentation purposes. It falls back to the default set in their values.yaml anyway, so removing them from here is a no-op.

samwho:
@shogunpurple I've pulled out the Dockerfile changes to this PR https://github.com/Budibase/budibase/pull/12441 so that we can get them in and get some new images built, which will make it a bit easier to do continued testing of the migration paths in my home cluster and the QA cluster.

charts/budibase/values.yaml
shogunpurple: Yep, exactly - and due to the "two NGINXes" you had to reflect that config in both.  Good to remove!

charts/budibase/values.yaml
shogunpurple: Makes sense - they do have a whole bunch of custom logic in there, and if it's simple enough for us to just drop in our image and keep using their chart then no point rewriting it. Let's just simplify the `values.yaml` here then and point to the official chart to make it easier for first timers and the standard use case!

charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -24,12 +21,9 @@ spec:
   template:
     metadata:
       annotations:
-        kompose.cmd: kompose convert
-        kompose.version: 1.21.0 (992df58d8)
 {{ if .Values.services.apps.templateAnnotations }}
 {{- toYaml .Values.services.apps.templateAnnotations | indent 8 -}}
 {{ end }}
-      creationTimestamp: null
```
calexiou, line 32: I believe this was here to mitigate the fact that our images didn't update their tag for each new version, so we needed a way to trick helm into refetching the image. Is the tagging issue resolved now?

calexiou (approved):
LGTM! Great job 🚀

charts/budibase/templates/app-service-deployment.yaml
samwho: We have an `imagePullPolicy` of `Always`, so I don't think this trick is necessary.

shogunpurple (approved):
LGTM! The documentation under every option in `values.yaml` is excellent. Such a big improvement to the current chart!

charts/budibase/values.yaml
```diff
@@ -1,56 +1,32 @@
...
+  # -- What ingress class to use.
   className: ""
-  annotations:
-    kubernetes.io/ingress.class: nginx
-    nginx.ingress.kubernetes.io/client-max-body-size: 150M
-    nginx.ingress.kubernetes.io/proxy-body-size: 50m
+  # -- Standard hosts block for the Ingress resource. Defaults to pointing to the Budibase proxy.
   hosts:
-    - host: # change if using custom domain
+    # @ignore
```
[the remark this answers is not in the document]
samwho: Yes. 🙂

charts/budibase/values.yaml
```diff
@@ -60,361 +36,427 @@ ingress:
...
   enableAnalytics: "1"
+  # @ignore (only used if enableAnalytics is set to 1)
   posthogToken: "phc_bIjZL7oh2GEUd2vqvTBH8WvrX0fWTFQMs6H5KQxiUxU"
-  selfHosted: "1" # set to 0 for budibase cloud environment, set to 1 for self-hosted setup
-  multiTenancy: "0" # set to 0 to disable multiple orgs, set to 1 to enable multiple orgs
-  offlineMode: "0" # set to 1 to enable offline mode
+  # @ignore (should not normally need to be changed, we only set this to "0"
+  # when deploying to our Cloud environment)
+  selfHosted: "1"
+  # -- Whether to enable the multi-tenancy feature or not. "0" means an installation can only
```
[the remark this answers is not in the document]
samwho: Should I @ignore it in that case? If not, what steps does the user need to take to enable it?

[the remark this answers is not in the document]
shogunpurple: They can't, as today it relies on account portal, so it's more of an internal value for us. We will likely open it in future to enterprise customers

PR #12417, opened 2023-11-21
Allowing use of BB_ADMIN environment variables at all times

mike12345567:
## Description
Updating bb admin user creation so that it can be used incase in self host a user gets locked out, the environment variables can be used to create a simple user to access the system.

I have also done a bit of refactoring as part of this around the user layers, nothing major, but just taking some of the core functionality and making it more available - previously this was done through an API but the core functionality needed is quite simple.

This is in response to users asking how to login to their self host installation after say credentials to the main admin account are lost and forgotten password is not configured, or SSO setup fails and login is lost.

The environment variables used for this are `BB_ADMIN_EMAIL` and `BB_ADMIN_PASSWORD`.

## Feature branch env
[Feature Branch Link](http://fb-fix-admin-user-backup.fb.qa.budibase.net)


shogunpurple (approved):
LGTM

PR #12426, opened 2023-11-23
Fix relationship filter setting being ignored, and debounce searches

aptkingston:
## Description
Fixes an issue where the relationship picker filter setting was being ignored. I've also debounced the search requests sent whenever using the inline searching, as there was no debounce at all which meant it was absolutely assaulting the server.

There are many other issues with the relationship picker component - it's a complete mess. This will keep it alive for now, but it really need fully rewritten at some point.

## Addresses
- https://linear.app/budibase/issue/BUDI-7734/filtering-on-a-relationship-picker-component-doesnt-work

## Feature branch env
[Feature Branch Link](http://fb-fix-relationship-picker-filtering.fb.qa.budibase.net)


mike12345567 (approved):
LGTM!

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -124,10 +124,21 @@
...
     }
+
+    // Ensure we match all filters, rather than any
+    const baseFilter = (filter || []).filter(x => x.operator !== "allOr")
     await fetch.update({
-      query: { string: { [primaryDisplay]: searchTerm } },
+      filter: [
+        ...baseFilter,
+        {
+          field: `999:${primaryDisplay}`,
```
mike12345567, line 134: NAB: it would be nice to have a filtering utility for this, to explain why the `999:` is needed. I know this is so that if someone else has searched for the primary display to make sure this filter is added on as well, but having something that explains that reasoning could be helpful.

aptkingston: For sure. To be honest I think the whole filtering experience needs an overhaul. We should design our own search payload, with flags (like "allOr" and "onEmptyFilter"), in combination with an array of filters, and the backend can process that and transform it into the correct lucene structure.

At the moment there's so much random logic across the frontend - some dealing with "filters" (this array structure that the builder generated), some with "query" (the actual lucene object) and a bunch of stuff to convert between them.

adrinr: Could we add this 999 as a variable to remove the magic string? Or add a comment on how this works?

mike12345567: @aptkingston - just to check, by "actual lucene object" you mean the search syntax of the backend, rather than specifically Lucene strings being sent up?

aptkingston: It doesn't need to be 999 - that's just a safe choice since it will never clash with an existing filter. Any numeric prefix is safe as long as it doesn't clash so the more natural choice would be `baseFilter.length + 1`. 999 was just more terse 😂

aptkingston: @mike12345567 yea sorry I tend to use that term incorrectly, but yes the backend search object with `equal`, `string` keys etc.

aptkingston: Added a comment to explain the 999.

PR #12427, opened 2023-11-23
Add Buttongroup configuration support to Formblock/Tableblock

deanhannigan:
## Description

- `Formblock` buttons have been updated to allow complete customisation. Each form type has their own default button configuration, but you can now add/remove/customise buttons as you see fit.
- A new option `Button position` has been added to the `Styles` section. You can now configure your form buttons to appear at either the top or the bottom.  Existing forms will have their buttons remain at the top and new forms will have them at the bottom.
- `Tableblock` forms have been updated in the background to accomodate the `Formblock` updates but their external UX remains the same.
- Legacy config options are still valid and will be transferred to the new button configuration. If you're curious, you can open the configuration for your old forms and things like the `Navigate To` url will still be present if you had configured it before.
- For the `Formblock`, the `resetOn` behaviour for the button config will now trigger if the `Data` OR `Type` are changed. This is to make sure that the button configuration is properly regenerated.
- `resetOn` behaviour has been updated to support a string AND an array of strings.

## Addresses
- https://linear.app/budibase/issue/BUDI-7535/form-block-add-the-new-button-group-to-the-form-block

## Screenshots

The buttons config properties have now been replaced with a button configuration component

---

Previous `Formblock` button configuration layout

Screenshot 2023-11-23 at 11 15 40

---

Updated button configuration layout. The default button configurations have been exposed and any previous config will appear here.

Screenshot 2023-11-23 at 11 34 25

## Feature branch env
[Feature Branch Link](http://fb-feature-add-buttongroup-to-formblock.fb.qa.budibase.net)


packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1271,14 +1299,17 @@ export const getFrontendStore = () => {
...
+          store.actions.components.processNestedSettings(component)
         }
       },
       requestEjectBlock: componentId => {
         store.actions.preview.sendEvent("eject-block", componentId)
       },
       handleEjectBlock: async (componentId, ejectedDefinition) => {
         let nextSelectedComponentId
-
+        console.log("EJECTING")
```
shogunpurple, line 1312: Can remove

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -1247,9 +1271,13 @@ export const getFrontendStore = () => {
           const settings = getComponentSettings(component._component)
           const updatedSetting = settings.find(setting => setting.key === name)
 
-          const resetFields = settings.filter(
-            setting => name === setting.resetOn
-          )
+          // Can be a single string or array of strings
+          const resetFields = settings.filter(setting => {
```
shogunpurple, line 1275: NAB: seems cleaner to return inline here

packages/client/src/components/app/blocks/TableBlock.svelte
```diff
@@ -33,6 +34,7 @@
   export let notificationOverride
 
   const { fetchDatasourceSchema, API } = getContext("sdk")
+  const component = getContext("component")
```
shogunpurple, line 37: NAB: it might be good for us to make these into constants

packages/frontend-core/src/utils/utils.js
```diff
@@ -106,3 +109,135 @@ export const domDebounce = callback => {
     }
   }
 }
+
+/**
+ * Build the default FormBlock button configs per actionType
+ * Parse any legacy button config and mirror its the outcome
```
shogunpurple, line 115: Typo in this comment

packages/frontend-core/src/utils/utils.js
```diff
@@ -106,3 +109,135 @@ export const domDebounce = callback => {
...
+    notificationOverride,
+    actionUrl,
+    showDeleteButton,
+    deleteButtonLabel,
+    showSaveButton,
+    saveButtonLabel,
+  } = props || {}
+
+  if (!_id) {
+    console.log("MISSING ID")
```
shogunpurple, line 133: Remove log

shogunpurple (approved):
LGTM - just a few small ones!

packages/client/src/components/app/blocks/TableBlock.svelte
aptkingston: Would be a good ticket for a refactoring/cleanup project for sure. Would be nice to roll that out to all components/blocks and provide the constants via the SDK for plugin authors too.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -19,6 +19,8 @@
   export let includeHidden = false
   export let tag
 
+  $: store.actions.components.processNestedSettings(componentInstance)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:22]: Is this to try and do the migration just whenever the component is rendered for the first time? I reckon it might be cleaner to move this logic actually into the `selectedComponent` derived store itself, right before it returns the instance - which would mean that by using `$selectedComponent` anywhere we already have the up to date migrated instance. It also escapes any race conditions which could happen here since there will be an initial render of the settings section before this reactive statement runs.

Do you think it might also make sense to name this `migrateSettings` as well? To make it clear that the intention is to migrate legacy settings into new ones.

aptkingston (commented):
Awesome work @deanhannigan. I really like the approach of pulling out the button generation logic and reusing it in both the component itself as well as the builder, since it ensures we're always handling things the same way.

Left one little suggestion around how we might be able to use `processNestedSettings` a little more generically (and how I think it's basically migrating settings and we should call it that) but it works perfectly and looks great 👍

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -601,6 +601,28 @@ export const getFrontendStore = () => {
...
+      processNestedSettings: enrichedComponent => {
+        const componentPrefix = "@budibase/standard-components"
+
+        if (enrichedComponent?._component == `${componentPrefix}/formblock`) {
+          // Use default config if the 'buttons' prop has never been initialised
+          if (!("buttons" in enrichedComponent)) {
+            enrichedComponent["buttons"] =
+              Utils.buildDynamicButtonConfig(enrichedComponent)
+            //Ensure existing Formblocks position their buttons at the top.
+            enrichedComponent["buttonPosition"] =
```
Ghrehh, line 613: NAB: Would `||=` work here? Or maybe even `??=` if the scenario is appropriate?

packages/client/src/components/app/blocks/form/InnerFormBlock.svelte
```diff
@@ -184,42 +113,14 @@
               props={{ text: title || "" }}
               order={0}
             />
-            {#if renderButtons}
+            {#if buttonPosition == "top"}
```
Ghrehh, line 116: Is the non-strict equality intentional here?

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -601,6 +601,28 @@ export const getFrontendStore = () => {
...
       },
+      processNestedSettings: enrichedComponent => {
+        const componentPrefix = "@budibase/standard-components"
+
+        if (enrichedComponent?._component == `${componentPrefix}/formblock`) {
+          // Use default config if the 'buttons' prop has never been initialised
+          if (!("buttons" in enrichedComponent)) {
+            enrichedComponent["buttons"] =
+              Utils.buildDynamicButtonConfig(enrichedComponent)
+            //Ensure existing Formblocks position their buttons at the top.
```
Ghrehh, line 612: Very very very NAB:  Lack of a space after a comment here compared with all the other comments you added.

Ghrehh (commented):
lgtm

packages/builder/src/builderStore/index.js
```diff
@@ -69,7 +69,14 @@ export const selectedComponent = derived(
     if (!$selectedScreen || !$store.selectedComponentId) {
       return null
     }
-    return findComponent($selectedScreen?.props, $store.selectedComponentId)
+    const selected = findComponent(
+      $selectedScreen?.props,
+      $store.selectedComponentId
+    )
+
+    const clone = selected ? Object.assign({}, selected) : selected
```
aptkingston, line 77: `Object.assign` still creates a shallow copy so this wouldn't be safe if we're mutating any nested properties as part of `migrateSettings` (which I don't think we are for now, but obviously could in future). Maybe safer to do a proper deep clone here to ensure we'll always be safe?

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -869,7 +877,13 @@ export const getFrontendStore = () => {
           if (!component) {
             return false
           }
-          return patchFn(component, screen)
+
+          // Mutates the fetched component with updates
+          const updated = patchFn(component, screen)
+          // Mutates the component with any required settings updates
+          const migrated = store.actions.components.migrateSettings(component)
```
aptkingston, line 884: Nice 👌  Good call here by returning a flag to let us know if a migration was required, to let us skip wasted updates.

aptkingston (commented):
Nice one Dean! One final comment about maybe using a deep clone to prevent potential issues, but LGTM after that!

aptkingston (approved):

PR #12436, opened 2023-11-24
Proof of concept for the app migration mechanism

adrinr:
## Description
This is just a proof of concept of the migration system. The current state does the follow:
1. When a new migration is required, a new middleware will detect it and queue it up for processing
2. When loading the builder, the user will be redirected to an "uploading" page. This page will redirect back to the original page when the migrations are done (using a pulling mechanism)
3. When loading the app, if a migration is required, the user will be redirected to an "uploading" page different than the builder
4. When navigating through a loaded app, if a migration is required, the user will be redirected to that "uploading" page

The wider design and discussions can be found in the following doc:
https://docs.google.com/document/d/1aG0WB10ZVNRRdHTd4LdyEVidGvTFzvihRb_u6U6EW-U/edit?usp=sharing

## Feature branch env
[Feature Branch Link](http://fb-poc-budi-7655.fb.qa.budibase.net)


packages/frontend-core/src/api/index.js
```diff
@@ -170,6 +170,14 @@ export const createAPIClient = config => {
 
     // Handle response
     if (response.status >= 200 && response.status < 400) {
+      const migration = response.headers.get("migration-header")
```
mike12345567, line 173: We should get @aptkingston thoughts on best way to handle this, I think an event dispatcher which the builder can attach to could work, but he might have a better idea!

packages/server/src/api/controllers/migrations.ts
```diff
@@ -1,14 +1,28 @@
-import { migrate as migrationImpl, MIGRATIONS } from "../../migrations"
-import { BBContext } from "@budibase/types"
+import { context } from "@budibase/backend-core"
```
mike12345567, line 1: We might want to create a separate app migration API - reason being the migration system isn't totally deprecated yet - we'll probably keep it around for a little while as tenant level migrations still have to be ran this way. A separate app migration API would mean we can keep the two totally separate until we handle tenant level migrations as well.

packages/server/src/api/controllers/static/index.ts
```diff
@@ -126,6 +127,18 @@ export const deleteObjects = async function (ctx: Ctx) {
...
+  // TODO: merge with middleware
+  const appDb = context.getAppDB()
+  const migrationDoc = await appDb.get<MigrationDoc>("_design/migrations")
+  const latestMigration =
+    Object.keys(MIGRATIONS).sort()[Object.keys(MIGRATIONS).length - 1]
+
+  const migrated = migrationDoc.version === latestMigration
+  if (!migrated) {
+    // TODO
+    return ctx.redirect("/app-preview-page")
```
mike12345567, line 139: I don't know if we should re-direct this, as the client library won't necessarily know where to go, it might be easier to server render an app being migrated page.

packages/server/src/middleware/migrations.ts
```diff
@@ -0,0 +1,143 @@
+import { context, locks, queue } from "@budibase/backend-core"
+import { UserCtx, Document, LockName, LockType } from "@budibase/types"
+import Bull, { Job } from "bull"
+
+export const MIGRATIONS: Record<string, { migration: () => Promise<void> }> = {
+  "20231122115100": {
+    migration: async () => {
+      await new Promise<void>(r => {
```
mike12345567, line 8: we will need to work out the migration runner, as discussed yesterday this will likely need to involve a bit of: https://linear.app/budibase/issue/BUDI-7611/db-batching-functionality

packages/builder/src/pages/builder/updating/index.svelte
```diff
@@ -0,0 +1,31 @@
...
+    const response = await API.get({ url: "/api/migrations/status" })
+    if (!response.migrated) {
+      return
+    }
+
+    const urlParams = new URLSearchParams(window.location.search)
+    const returnUrl = urlParams.get("returnUrl")
+
+    window.location = returnUrl
+  }, 1000)
```
samwho, line 15: Probably want to do this with some kind of back off and jitter to avoid lots of users polling at the same time and DoSing ourselves.

packages/server/src/api/controllers/migrations.ts
adrinr: Yup, this is just for the PoC and to get the idea. Then things will need to be moved to their proper place

packages/server/src/middleware/migrations.ts
```diff
@@ -0,0 +1,143 @@
...
+      type: LockType.DEFAULT,
+      resource: appId,
+      ttl: 30000,
+    },
+    async () => {
+      await context.doInAppContext(appId, async () => {
+        const db = context.getAppDB()
+        const migrationDoc = await db.get<MigrationDoc>("_design/migrations")
+
+        const currentVersion = migrationDoc.version || ""
```
samwho, line 49: How come the `|| ""` here?

packages/server/src/middleware/migrations.ts
```diff
@@ -0,0 +1,143 @@
...
+  },
+}
+
+export interface MigrationDoc extends Document {
+  version: string
+}
+
+const appQueues: Record<string, Bull.Queue<any>> = {}
+
+async function processMessage(job: Job) {
```
samwho, line 34: I mentioned it on the design doc, but it would be nice to be able to store migration logs in a way that's easier for us or the user to fetch them to try and debug what went wrong if a migration fails. Not sure how easy that would be today, or if we would want to put something special in place to capture and store them.

packages/server/src/middleware/migrations.ts
```diff
@@ -0,0 +1,143 @@
...
+export default async (ctx: UserCtx, next: any) => {
+  if (!ctx.appId) {
+    return next()
+  }
+
+  try {
+    if (new URL(ctx.headers.referer!).pathname === "/builder/updating") {
+      return next()
+    }
+  } catch {}
```
samwho, line 83: What errors are you expecting here?

packages/server/src/middleware/migrations.ts
```diff
@@ -0,0 +1,143 @@
...
+          appQueues[appId] = queue.createQueue(appId)
+          appQueue = appQueues[appId]
+          appQueue.process(processMessage)
+        }
+      )
+
+      if (!appQueue) {
+        // TODO
+        throw "Error"
+      }
```
samwho, line 129: Under what circumstances would `appQueue` still not be set here?

adrinr:
PoC discussed, proceeding to the actual implementation

Issue #9776, opened 2023-02-22
[BUDI-6630] Table export CSV action includes headers for all columns instead of selected ones


andz-bb:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.3.17
  * App Version: 2.3.17

**Describe the bug**
When choosing specific columns to export from a table, the headers for the unselected columns are still included along with empty strings for each of them. 

**To Reproduce**

1. Create a table
2. Create a button with an Export Data action
3. Select the table and specific columns to export
4. Click the export button to download the file and see unselected column headers in the CSV

**Expected behaviour**

A CSV file should be generated with headers and data for only the selected columns.

**Screenshots**

![Screenshot 2023-02-22 at 15.58.12.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/bc41a412-4d26-49c5-9c87-7c4931bfc23f/465fb6fb-0b41-4692-9166-730864420aba)

![Screenshot 2023-02-22 at 15.49.45.png](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/38be666e-7d19-4314-b78c-28f43d1bed86/ca27500f-5404-41bc-b8e2-371be32b72d6)

**Additional context**

raised by community member here: [https://github.com/Budibase/budibase/discussions/9773](https://github.com/Budibase/budibase/discussions/9773)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6630](https://linear.app/budibase/issue/BUDI-6630/table-export-csv-action-includes-headers-for-all-columns-instead-of)</sub>

zedascouves:
Just to say that I also have this issue. Is there a way to upvote a bug, so there's a way to know if it's affecting many people?

andyburgessmd:
I also have this issue and would upvote

PR #12438, opened 2023-11-24
Only export selected columns

melohagan:
## Description
CSV exports from the **Export Data** action now only include the selected columns

## Addresses
- https://github.com/Budibase/budibase/issues/9776

## Screenshots
Internal DB:
Screenshot 2023-11-24 at 11 54 14

Screenshot 2023-11-24 at 11 51 55

External DB:
Screenshot 2023-11-24 at 11 53 52

Screenshot 2023-11-24 at 11 52 17


packages/server/src/sdk/app/rows/search/external.ts
```diff
@@ -133,32 +133,32 @@ export async function exportRows(
...
 
+  // Filter data to only specified columns if required
   if (columns && columns.length) {
     for (let i = 0; i < result.rows.length; i++) {
       rows[i] = {}
       for (let column of columns) {
         rows[i][column] = result.rows[i][column]
       }
     }
+    headers = Object.keys(schema).filter(header => columns.includes(header))
```
adrinr, line 151: The "columns" should never be bigger than the schema. Could we iterate through "columns" checking that is a valid schema key?

adrinr (commented):
LGTM, just added a small comment. Could we add some unit test for it?

packages/server/src/sdk/app/rows/search/external.ts
melohagan: Realised that this was completely unnecessary, I can just directly use the columns in this case, no need to do the filter.

Unit test added as well

packages/server/src/sdk/app/rows/search/external.ts
```diff
@@ -133,32 +133,32 @@ export async function exportRows(
 
   let result = await search({ tableId, query: requestQuery, sort, sortOrder })
   let rows: Row[] = []
+  let headers
 
-  // Filter data to only specified columns if required
+  if (!tableName) {
+    throw new HTTPError("Could not find table name.", 400)
```
samwho, line 139: nit: please use `BadRequestError` instead of `HTTPError` here. It represents the intent slightly better.

samwho (approved):
Just the one comment, and I don't mind you ignoring it. Generally if I prepend a comment with `nit`, don't feel obliged to do it. 🙂

PR #12441, opened 2023-11-24
Changes to the single image and CouchDB image builds to make them work with new Helm chart changes.

samwho:
## Description

These changes have been extracted from https://github.com/Budibase/budibase/pull/12416.

Main changes:

1. Inlined build-target-paths.sh as it's only used in one place and is highly specific to the CouchDB image.
2. Replaced a `sleep 10` in the CouchDB `runner.sh` with a health checking loop. This fixes a situation where the container can start but CouchDB isn't ready when it attempts to do the initial cluster setup, so the image comes up in a broken state.
3. Makes the CouchDB image aware of some changes it needs to apply when running as part of the single image or in a Kubernetes environment. In the single image, it needs to write its data to the `/data` directory (even in Kubernetes, where we mount a gp2 volume to /data). In Kubernetes but not the in the single image (so when run through our Helm chart) it has to write data to `/opt/couchdb/data` because that's where the official CouchDB chart mounts a persistent volume. It also removes the `-name couchdb@127.0.0.1` argument to the Erlang VM because the CouchDB uses the pod's fully qualified domain name for clustering purposes.
4. `export` the `DATA_DIR` variable so that it gets passed to scripts run by `runner.sh`.

## Feature branch env
[Feature Branch Link](http://fb-update-couchdb-single-image.fb.qa.budibase.net)


adrinr (approved):
Nice one!

PR #12442, opened 2023-11-24
count creators with the new global creator flag

shogunpurple:
## Description
Work to count creators with the new global creator flag

### DIFF
--- a/packages/shared-core/src/sdk/documents/users.ts
+++ b/packages/shared-core/src/sdk/documents/users.ts
@@ -13,7 +13,7 @@ export function isBuilder(user: User | ContextUser, appId?: string): boolean {
   if (!user) {
     return false
   }
-  if (user.builder?.global) {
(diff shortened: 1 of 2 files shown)


jvcalderon (approved):
LGTM!

aptkingston:
I don't think either of these changes are actually needed - I've updated the types in my PR and I've also updated the user SDK to correctly check if a user is a creator. I actually think the logic in this change is slightly wrong too - having the creator flag does not implicitly imply they are a builder for all apps, so I think the `isBuilder` function should remain unchanged. Creators will still have app IDs in their `builder.apps` array, which that function already accounts for without any changes.

The user SDK in my PR has changes to `isCreator` and a new `hasCreatorPermissions` function which both include the logic to determine creator privileges.

shogunpurple:
@aptkingston you're right - your PR accounts for this already. Nice one - closing

PR #12443, opened 2023-11-24
Creator changes

aptkingston:
## Description
Collection of changes to enable new pricing changes.

Corresponding pro PR: https://github.com/Budibase/budibase-pro/pull/245

## Functionality changes
- New "Creator" tenant-level role, specified by the `builder.creator` flag on user objects
- The "Creator" role allows users to create their own apps, as well as have the ability to be given creator access to other apps
- The "Creator" role is available to free users, however being able to give creator access to an app is restricted. This means users can build their own private apps on the free plan, but cannot collaborate with others on the free plan unless all users are admins
- App creator access can only be given to users with the "Creator" tenant role
- Users are automatically promoted to the "Creator" role when given creator access to any app
- Users are *not* ever automatically demoted if that creator access is revoked
- Groups can no longer be used to assign creator access to apps, as that would require that all members of the group have the "Creator" tenant level role
- Demoting a user down to being a global "App user" will revoke all editing capabilities

## Backwards compatibility
- Existing users which have creator access to apps will continue to have access. They will still appear as "App users" in the global users list and will not be global creators. It will no longer be possible to get into this state though, as assigning creator access to any other apps will promote their tenant role to "Creator". They can also safely be promoted to a creator in the user settings page.
- Group creator access does not work and cannot be assigned, but as far as I can tell this never actually worked anyway

## UI changes
- New copy for roles across the board: user details, inviting users, importing users:
image
image
image
- Roles and access updated in user settings. App access now displayed as "Can use as X". Creator access displayed as "Can edit". This applies to tables in user and group list and details pages.
image
image
image
- User side panel updated to use these same copy changes:
image
- Account admins are now displayed as having creator access ("Can edit") rather than admin access
image
- User invitation updated to use new copy and add creator tenant role
image
- User invitation app access also updated with the same copy
image
- Inviting an account admin forces the use of the creator app access
image

## Addresses
- https://linear.app/budibase/issue/BUDI-7645/user-side-panel-inviting-new-users
- https://linear.app/budibase/issue/BUDI-7644/user-side-panel-group-assignment
- https://linear.app/budibase/issue/BUDI-7643/update-user-side-panel
- https://linear.app/budibase/issue/BUDI-7646/portal-invite-new-user-updates

## Feature branch env
[Feature Branch Link](http://fb-creator-changes.fb.qa.budibase.net)


shogunpurple (approved):
LGTM! Nice work @aptkingston

packages/backend-core/src/users/db.ts
```diff
@@ -146,12 +146,12 @@ export class UserDB {
 
   static async allUsers() {
     const db = getGlobalDB()
-    const response = await db.allDocs(
+    const response = await db.allDocs<User>(
```
mike12345567, line 149: 👌

Issue #11629, opened 2023-08-30
[BUDI-7444] Queries use blank string instead of column default value when binding default value is empty


abreland-amd:
**Checklist**

* [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: Docker Compose
  * Budibase Version: 2.9.30
  * App Version: 2.9.30

I am using a MySQL 8.0.34 database and one of my tables has the following columns:

![](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/c62084c1-1be6-49ee-8a49-6960dfe488ec/15558b91-5d8b-42c9-a59b-2e950afde876?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9jNjIwODRjMS0xYmU2LTQ5ZWUtOGE0OS02OTYwZGZlNDg4ZWMvMTU1NThiOTEtNWQ4Yi00MmM5LWE1OWItMmU5NTBhZmRlODc2IiwiaWF0IjoxNjkzNTExNzY2LCJleHAiOjE2OTM1OTgxNjZ9.XZo-3zFmndRNbWbOGXKWUCbV9NSKglBzgP8RcZnpJMc)

But when I create a query and try to insert, using the following SQL/handlebars:

INSERT INTO configs (type, customer_project_id, board_id, vendor_id, build_phase_id, cvl_id)
VALUES ( {{ type }}, {{ customer_project_id }}, {{ board_id}}, {{ vendor_id}}, {{ build_phase_id}}, {{ cvl_id}} );

and the following bindings:

[...]

PR #12447, opened 2023-11-27
FIX: queries use blank string instead of column default value

melohagan:
## Description
Blank query bindings were passing empty strings into the SQL queries, which was causing errors for number types.

Fixed so that blank bindings for SQL queries now use null instead, which is also consistent with the row mapper: https://github.com/Budibase/budibase/blob/master/packages/server/src/utilities/rowProcessor/map.ts
 

## Addresses
- https://github.com/Budibase/budibase/issues/11629


samwho (commented):
This is making my spidey-senses tingle.

What if we're legitimately trying to save an empty string to a column?

What's preventing whatever's upstream of this code from correctly sending null?

melohagan:
@samwho This is a visual of the problem here:

Screenshot 2023-11-29 at 10 30 48

Because the bindings use a text field, leaving the default value blank will pass an empty string.

It's impossible to pass `null`, as it gets converted to a string:

Screenshot 2023-11-29 at 10 31 51

Even in the case of trying to pass `null` via JS in the Design section, it will interoperate `null` as ignore, which will then try to use the invalid default value. 
 
Screenshot 2023-11-29 at 10 34 24

samwho:
That's a pickle.

I'm going to wait to see what Mike says, but with my relatively new-to-the-codebase eyes it feels like coercing `""` to `null` at the query building layer isn't the right solution. It has too much scope to mess with other things.

If it's possible to find a layer higher up to do this coercion, that specifically addresses this problem with bindings, that would be preferable I think.

mike12345567:
Yeap agree with @samwho - I think doing this at the actual datasource implementation level means things will be missed, for example the Oracle datasource is not covered by this.

I believe if we could move this logic in someway up to where we process the bindings into prepared statements for SQL queries then we can remove the repetition and make it a more consistent rule. Where the type coercion function is called from should be a suitable place to carry out these checks. 

I know we've discussed this a few times before, but ideally I would love to schedule a call across teams to decide on some actual rules to follow in regards to what `null`, `undefined` and `""` mean when the frontend sends them up; it probably won't be the easiest fix, but it is obviously a massive source of bugs as I know @melohagan you've already raised a couple of edge case fixes around this.

melohagan:
Thanks for taking a look guys! Will close this PR for now 👍

PR #12453, opened 2023-11-27
Fix Deleting and Remaking Relationships to the User Table

Ghrehh:
Creating, deleting and remaking a relationship to the another budibase table wasn't working correctly until the page is refreshed. This was because we we're correctly refetching child tables after deleting the relationship column.

## Feature branch env
[Feature Branch Link](http://fb-fix-deleting-and-remaking-user-relationship-column.fb.qa.budibase.net)


samwho:
@aptkingston hey is this related to the problem we were debugging back when I was doing the user-column migration work? I swear we had problems when deleting relationships.

aptkingston:
@samwho could well have been! There was another related issue here to do with primary display columns in the grid being cached and going stale, and that one has since been fixed, but this could also have been related.

aptkingston (commented):
I'm surprised this fixes things, because I would expect the builder websocket to receive events for any sort of table mutation (including deleting a column like this) but I assume it isn't if this fix works.

I dislike having to do the nuclear approach like this of just reloading all data whenever you do a tiny change, so while I would much rather investigate why the websockets aren't handling this case, I also don't mind just merging this and moving on.

aptkingston:
Maybe there specific cases where the websocket doesn't work - like the user table for example. If that's the case there will be other bugs and it would be worth root causing this properly. You mention specifically the users table in the title of this PR, but this code affects all tables. Does the issue only happen with the users table? So I'm assuming this is when you delete a relationship in table `Foo` where the other side of the relationship is the users table?

Ghrehh:
@aptkingston Feedback addressed if you could give this another look please 🙏

aptkingston (approved):
LGTM!

PR #12456, opened 2023-11-28
Tweak CouchDB image to better work in Kube

samwho:
## Description

This change removes the `database_dir` and `view_index_dir` from the `local.ini` file we ship with our CouchDB image, but only if it detects that it's running in Kubernetes.

This is to facilitate the smoothest possible transition from the official CouchDB image to our CouchDB image when we update our Helm chart. Today, we use the official CouchDB image which defaults to not setting these values, and they in turn default to the directory `/opt/couchdb/data`. Maintaining the defaults means that we can swap the image out and everything continues to work.

## Feature branch env
[Feature Branch Link](http://fb-tweak-couchdb-image-to-better-work-in-kube.fb.qa.budibase.net)


mike12345567 (approved):
LGTM ✅

PR #12461, opened 2023-11-28
Manually release version 2.13.19 of our Helm chart.

samwho:
## Description

Our Helm chart release process has been broken since November 4th. This PR manually creates the latest version of the Helm chart, v2.13.19, because in the next update we'd like to detach the chart version from the app version in order to release a major update to the chart.


adrinr (commented):
How is this working now? Is the helm chart not being used at all?

samwho:
The Helm chart hasn't been released since November 4th. Because the `--version 0.0.0-master` line, the `docs/index.yaml` never updates.

I don't think many people are using the Helm chart today, but if they are this wouldn't have broken them. They just won't have been able to update past 2.12.11 for the last month or so. I doubt they noticed.

adrinr (approved):

PR #12465, opened 2023-11-29
Hover component tree preview

Ghrehh:
## Description
Hovering the component tree now highlights the component in the app preview and vice versa. I made the hover background colour slightly dimmer as having a highlighted component next to a selected one in the component tree with the same colour is a little strange when the cursor isn't also present.

## Addresses
https://linear.app/budibase/issue/BUDI-7674/design-section-hover-component-tree-reflects-in-preview

## Feature branch env
[Feature Branch Link](http://fb-hover-component-tree-preview.fb.qa.budibase.net)


melohagan (approved):
LGTM! This is a really nice feature 👌 

One small UI thing I noticed was that the menu `...` appears beside the component in the tree, which in my opinion doesn't make sense because you aren't directly hovering over the component for the extra actions:

https://github.com/Budibase/budibase/assets/101575380/3076286e-a17d-491e-aaed-dfb5d4a638aa

Also if you have a parent selected, then you can't see which child is being hovered in the component tree, which may be fine.

packages/builder/src/components/common/NavItem.svelte
```diff
@@ -152,15 +158,17 @@
...
-  .nav-item:hover {
-    background-color: var(--spectrum-global-color-gray-300);
+  .nav-item:hover,
+  .isHovering {
+    background-color: var(--spectrum-global-color-gray-200);
     --avatars-background: var(--spectrum-global-color-gray-300);
   }
-  .nav-item:hover .actions {
+  .nav-item:hover .actions,
+  .isHovering .actions {
```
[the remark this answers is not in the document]
Ghrehh: Ended up removing this one to apply your feedback about removing the "..." on app preview hover, but yeah it's necessary because we want to apply these attributes to two different classes.

packages/builder/src/components/common/NavItem.svelte
```diff
@@ -61,6 +62,7 @@
 
 <div
   class="nav-item"
+  class:hovering
```
adrinr, line 65: Why are we not using the default `hover`?

Ghrehh: Because this class is applied whenever the user is hovering over the corresponding component in the app preview, so it needs to be applied via js, I don't think there's any way we could do this with the `:hover` pseudo-class but I could be wrong.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ComponentList/ComponentTree.svelte
```diff
@@ -109,6 +118,9 @@
         on:dragover={dragover(component, index)}
         on:iconClick={() => toggleNodeOpen(component._id)}
         on:drop={onDrop}
+        hovering={$store.hoverComponentId === component._id}
+        on:mouseover={() => handleMouseover(component._id)}
+        on:mouseout={() => handleMouseout(component._id)}
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ComponentList/ComponentTree.svelte:123]: Using `mouseover` and `mouseout` here thrashes the frontend store extremely hard (which cripples the page because everything depends on the frontend store). `mouseenter` and `mouseleave` are the events we should use here.

`mouseover` fires when you enter a component or its children, while `mouseenter` only fires for the outermost component, which is the behaviour we want. Same in reverse for `mouseleave` vs `mouseout`.

So we should update these events (and therefore the core nav item) to use `mouseenter` and `mouseleave` instead.

The same applies for the other nav item usages but I won't duplicate this comment.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ComponentList/ComponentTree.svelte
```diff
@@ -89,6 +89,15 @@
     }
     return findComponentPath($selectedComponent, component._id)?.length > 0
   }
+
+  const handleMouseover = componentId => {
+    $store.hoverComponentId = componentId
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ComponentList/ComponentTree.svelte:94]: We should check here if the store already has this component hovered, and skip if so. Same applies to the other handlers in other files.

packages/builder/src/components/common/NavItem.svelte
```diff
@@ -71,6 +73,10 @@
   on:dragstart
   on:dragover
   on:drop
+  on:mouseover
+  on:mouseout
+  on:focus
+  on:blur
```
aptkingston, line 79: As mentioned, we should update to use enter and leave instead. I don't see a usage for `focus` or `blur` here - do we need them? Otherwise it would be better to remove an additional 2 event handlers on every nav item.

aptkingston (commented):
Works well, great job! I think all we need to do is refactor this a little to fix the performance issues, then it's good to go.

packages/builder/src/components/common/NavItem.svelte
Ghrehh: Originally added this to silence a linting error complaining about the lack of them when specifying mouseover and mouseout, I assumed if they were unused they wouldn't generate handlers, but perhaps I'm wrong. But since we're using mouseenter and mouseleave now the linter is no longer complaining, so I removed this.

Ghrehh:
@aptkingston Addressed your feedback if you could give this another look please

aptkingston (approved):
LGTM!

Issue #12425, opened 2023-11-23
[BUDI-7760] Relationship lost after adding a new column and add data to it


markusthiel:
**Checklist**

* \[ x \] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker single image
  * Budibase Version: 2.13.12
  * App Version: 2.13.12

**Describe the bug**
Create a table "employee" with two columns. The first column contains a name, the second column is a relationship to another table in which a work schedule with dates can be edited. After an assignment has been made in the work schedule, the respective link, or possibly several, lands in the employee table in the relationship column. If a new column is added to the employee table and an entry is made there, the relationship assignments disappear.

**To Reproduce**

1. Go to Data
2. Create a table "employee" with a column "Name".
3. Create a second table "workschedule" with two columns "Date" and "Name" Name should be a relationship to "employee + Name"
4. Make some assignments in the table "workschedule".
5. Add a new column to "employee" like email or telephone.
6. Add an entree to someone who was previously assigned to some dates.
7. The new column in "employee" which includes the assignments gets lost, looses entrees.

**Expected behavior**
Database entrees should stay

**Screenshots**

![bug1](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/48e97755-a98c-4adb-9237-6f80f115a8ed/4f29e95b-ae08-4f75-a7fa-f1c75cdd6594?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC80OGU5Nzc1NS1hOThjLTRhZGItOTIzNy02ZjgwZjExNWE4ZWQvNGYyOWU5NWItYWUwOC00Zjc1LWE3ZmEtZjFjNzVjZGQ2NTk0IiwiaWF0IjoxNzAwNzM5ODk1LCJleHAiOjE3MDA4MjYyOTV9.Tr9j1T8xPd8hGlED-DmFR3NHjqyhuWkzIIKk7crKrvg)

[...]

PR #12468, opened 2023-11-29
Fix relationship saving when both sides of relationship use same field name (internal)

mike12345567:
## Description
Fix for saving relationships that have the same field name used on both sides, previously this could cause a relationship to be cleared depending on how the relationship schema was configured. There is a chance when saving that this won't happen as which side of the relationship is denoted by doc1 and doc2 is random, so when this happens is random. Using the table to pick the correct side is safer than just using the field name.

Also includes a fix for debugging with `dev:noserver` - the preview wouldn't load due to the difference in how they are run, if you do not run a built esbuild version of the server/worker it would run into problems, getting around that by detecting whether or not the `preview.hbs` has moved or not (as happens with esbuild).

## Addresses
- https://linear.app/budibase/issue/BUDI-7760/relationship-lost-after-adding-a-new-column-and-add-data-to-it
- https://github.com/Budibase/budibase/issues/12425

## Feature branch env
[Feature Branch Link](http://fb-fix-budi-7760.fb.qa.budibase.net)

### DIFF
--- a/packages/server/src/api/controllers/static/index.ts
+++ b/packages/server/src/api/controllers/static/index.ts
@@ -24,7 +24,7 @@ import AWS from "aws-sdk"
 import fs from "fs"
 import sdk from "../../../sdk"
 import * as pro from "@budibase/pro"
-import { App, Ctx, ProcessAttachmentResponse, Upload } from "@budibase/types"
+import { App, Ctx, ProcessAttachmentResponse } from "@budibase/types"
 
 const send = require("koa-send")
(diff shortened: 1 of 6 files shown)


shogunpurple (approved):
A test to cover this scenario would be good if possible.

mike12345567:
> A test to cover this scenario would be good if possible.

Was in the process of adding a test case, actually revealed an issue with external table configurations that needed fixing!

PR #12469, opened 2023-11-29
fix: use /health and remove 301 from the ingress

calexiou:
## Description
Improving the healthcheck endpoint and status codes.

## Feature branch env
[Feature Branch Link](http://fb-fix-correct-ingress-healthchecks.fb.qa.budibase.net)


shogunpurple (approved):
Ah yes, that would be it! We were looking at the wrong directory

Issue #12372, opened 2023-11-13
[BUDI-7773] Query users in automation


andyburgessmd:
**Checklist**

* I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker
  * Budibase Version: 2.13.9
  * App Version: 2.13.9

**Describe the bug**
Since the changes to the user table its not not possible to query the users table in automation. I used this previously to query all application users and remind them to fill in timesheet weekly.

**To Reproduce**

1. Create an automation and add a query rows action
2. Try to select the users table

**Expected behavior**
You should be able to select and query the users table

**Screenshots**

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/6b7bf3d2-218b-43f1-bda0-7570fa02ff09/29acad71-0ca4-4dd5-81b9-df2b7ef8d4c1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC82YjdiZjNkMi0yMThiLTQzZjEtYmRhMC03NTcwZmEwMmZmMDkvMjlhY2FkNzEtMGNhNC00ZGQ1LTgxYjktZGYyYjdlZjhkNGMxIiwiaWF0IjoxNzAwODE4MjkyLCJleHAiOjE3MDA5MDQ2OTJ9.r9cVLBVeqdMJMzj7uI-4chjpD5-P2dnEsXnTJzuE18k)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7773](https://linear.app/budibase/issue/BUDI-7773/query-users-in-automation)</sub>

melohagan:
Fixed in version **2.13.26**

PR #12472, opened 2023-11-30
Allow querying of users table from automation

PClmnt:
## Description
- Fixing an issue where the Users table wasn't shown in the automation Query Rows table selector. 

## Addresses
- #12372 

## Screenshots
Screenshot 2023-11-30 at 08 58 12

## Feature branch env
[Feature Branch Link](http://fb-fix-automations-query-user-table-fix.fb.qa.budibase.net)

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/TableSelector.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/TableSelector.svelte
@@ -22,7 +22,7 @@
 <Select
   on:change={onChange}
   bind:value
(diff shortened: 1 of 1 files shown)


melohagan (approved):
LGTM! Just wondering why it was filtered out in the first place?

melohagan:
> > LGTM! Just wondering why it was filtered out in the first place?
> 
> I made a mistake when I was removing references in the Create Row / Update Row blocks and accidentally filtered it here too.

Ah makes sense - no worries!

PR #12474, opened 2023-11-30
Smart locking mechanism for redis locks without TTL

adrinr:
## Description
Currently, we have some long running processes (such as migrations) that require locking, but we don't know for how long. Because of this we are locking these process with pretty long timeouts, but this can cause the following issues:
1. We might underestimate the task and lock less than the required time
2. If the process crash, the TTL will need to expire in order to be able to be repeated

This PR is allowing an extra configuration to our redis lock implementation that will allow long locks following the steps below:
1. Create the lock with the expected TTL (this could be any in this case)
2. Execute the task asynchronously
3. In parallel, periodically on TTL/2, extend the lock by TTL (it does not really extend the remaining ttl, it sets the TTL based on the current time)

This has the following benefits:

- If the process keeps running, the lock will be extended periodically
- If the process crashes, the lock will be released

## Feature branch env
[Feature Branch Link](http://fb-chore-allow-redis-without-ttl.fb.qa.budibase.net)

### DIFF
--- a/packages/backend-core/src/redis/redlockImpl.ts
+++ b/packages/backend-core/src/redis/redlockImpl.ts
@@ -4,8 +4,7 @@ import { LockOptions, LockType } from "@budibase/types"
 import * as context from "../context"
 import env from "../environment"
 import { logWarn } from "../logging"
-import { Duration } from "../utils"
-import { timers } from ".."
+import { utils } from "@budibase/shared-core"
 
 async function getClient(
   type: LockType,
@@ -14,7 +13,11 @@ async function getClient(
   if (type === LockType.CUSTOM) {
     return newRedlock(opts)
   }
-  if (env.isTest() && type !== LockType.TRY_ONCE) {
+  if (
+    env.isTest() &&
+    type !== LockType.TRY_ONCE &&
+    type !== LockType.AUTO_EXTEND
+  ) {
     return newRedlock(OPTIONS.TEST)
   }
   switch (type) {
@@ -30,13 +33,16 @@ async function getClient(
     case LockType.DELAY_500: {
       return newRedlock(OPTIONS.DELAY_500)
     }
+    case LockType.AUTO_EXTEND: {
+      return newRedlock(OPTIONS.AUTO_EXTEND)
+    }
     default: {
-      throw new Error(`Could not get redlock client: ${type}`)
+      throw utils.unreachable(type)
     }
   }
 }
 
-const OPTIONS = {
+const OPTIONS: Record<keyof typeof LockType | "TEST", Redlock.Options> = {
   TRY_ONCE: {
     // immediately throws an error if the lock is already held
     retryCount: 0,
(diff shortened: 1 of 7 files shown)


packages/backend-core/src/redis/redlockImpl.ts
```diff
@@ -105,13 +106,39 @@ export async function doWithLock<T>(
...
+      // We keep extending the lock while the task is running
+      const extendInIntervals = (): void => {
+        timeout = setTimeout(async () => {
+          let isExpired = false
+          try {
+            lock = await lock!.extend(opts.ttl)
+          } catch (err: any) {
+            isExpired = err.message.includes("Cannot extend lock on resource")
+            if (isExpired) {
+              console.error("The lock expired", { name })
```
mike12345567, line 127: Do we need to throw an error of some sort here? Theres a bit of a risk of undefined behaviour, the process is still running but the lock has timed out/been lost - another process could start running on the resources now and then both break.

packages/backend-core/src/redis/redlockImpl.ts
```diff
@@ -105,13 +106,39 @@ export async function doWithLock<T>(
...
+            } else {
+              throw err
+            }
+          }
+
+          if (!isExpired) {
+            extendInIntervals()
+          }
+        }, opts.ttl / 2)
```
mike12345567, line 136: Does the user still have the ability to set the TTL? Only worry is if a really short TTL was set, like 100ms, then this will end up trying to extend every 50ms, which likely won't be possible/could lead to failures to extend - maybe we should have a minimum TTL for auto-extend?

packages/backend-core/src/redis/redlockImpl.ts
```diff
@@ -67,10 +64,14 @@ const OPTIONS = {
...
   },
+  CUSTOM: {},
+  AUTO_EXTEND: {
+    retryCount: -1,
+  },
 }
 
 export async function newRedlock(opts: Redlock.Options = {}) {
-  let options = { ...OPTIONS.DEFAULT, ...opts }
+  let options = { ...OPTIONS, ...opts }
```
samwho, line 74: How come `.DEFAULT` has been removed here?

packages/backend-core/src/redis/redlockImpl.ts
adrinr: True, remove the try/catch. If there is an error because the lock expired, the lock itself should throw an exception and kill the tread. I will check if I can write a test for it

packages/backend-core/src/redis/redlockImpl.ts
adrinr: Not sure, good catch 😅

mike12345567 (approved):
LGTM - would just need to work out what happened with the `.DEFAULT`!

packages/backend-core/src/redis/redlockImpl.ts
mike12345567: > I was reading https://brooker.co.za/blog/2023/11/27/about-time.html recently and it mentions a "safety period" on locks that lasts after the expiry where the lock cannot be given out again, in order to handle clock drift and so on. Is that something we have / can configure in Redis?

Redlock by default uses a `0.01` factor on the TTL to adjust for drift during locks/unlocking, I believe that is automatically factored into the TTL release time, instead of deleting the entry it sets the final expiry based on the TTL + drift factor.

adrinr:
> LGTM - would just need to work out what happened with the `.DEFAULT`!

I re-added it

adrinr:
@samwho I am not sure why but I can't quote you in this comment: https://github.com/Budibase/budibase/pull/12474#discussion_r1410894315
AFAIK this should not affect us, as we don't use redis instances outside our docker hosts. The docker hosts will automatically sync the clocks between containers, so it should not be an issue. The same with k8s clusters.
Not sure what would happen if somebody is using external redis services

mike12345567 (approved):
LGTM - `jest.useFakeTimers` is very useful!

PR #12478, opened 2023-11-30
Move CODEOWNERS to the root to see if that fixes it.

samwho:
## Description

For some reason having it in the directories themselves wasn't working. Let's see if this fixes it.


adrinr (approved):
🎉

PR #12480, opened 2023-11-30
Update originalIndex on select

melohagan:
## Description
Fix for issue with actions of the same eventHandlerType not re-drawing the selected action content.

## Addresses
- https://linear.app/budibase/issue/BUDI-7787/bug-regarding-2-seperate-save-row-actions


ConorWebb96 (approved):
LGTM!

PR #12482, opened 2023-11-30
Clean up extraneous files in gh-pages branch.

samwho:
## Description

The `gh-pages` branch of this repository is where we store and manage our Helm repository. At some point in the past, it looks like this change came off of `master` or `develop` (or one of those branches was merged into it), so it contains a lot of old and outdated things that aren't actually doing anything.

This PR removes all of that and leaves behind only the stuff that's actively being used (as far as I can tell, will need someone to check me on that).

It notably removes the `charts/` directory, which is used in a GitHub Action to build our Helm chart. It was discovered that this is an old version of the chart, and new versions committed to `master` haven't been picked up by our release process. I'll follow-up with another PR that changes our Helm chart release process to use the `charts` directory from `master` to build the chart from now on.


shogunpurple (approved):
LGTM!

PR #12486, opened 2023-12-01
Create migration context functions

adrinr:
## Description
Creating a new context.doInAppMigrationContext function that will be used for the new backend migrations. This function will be marking the context as migrating, and it will not allow starting any new context. We want to protect ourselves from any migration running in multiple databases or doing multiple db save, so we will be forcing this new function to be used in any migration script.

### DIFF
--- a/packages/backend-core/src/context/mainContext.ts
+++ b/packages/backend-core/src/context/mainContext.ts
@@ -148,12 +148,20 @@ export async function doInAppContext<T>(
   appId: string,
   task: () => T
 ): Promise<T> {
+  return _doInAppContext(appId, task)
+}
+
+async function _doInAppContext<T>(
+  appId: string,
+  task: () => T,
+  extraContextSettings?: ContextMap
+): Promise<T> {
   if (!appId) {
     throw new Error("appId is required")
   }
 
   const tenantId = getTenantIDFromAppID(appId)
-  const updates: ContextMap = { appId }
+  const updates: ContextMap = { appId, ...extraContextSettings }
   if (tenantId) {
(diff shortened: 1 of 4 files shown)


packages/backend-core/src/context/mainContext.ts
```diff
@@ -182,6 +188,24 @@ export async function doInIdentityContext<T>(
   return newContext(context, task)
 }
 
+function guardMigration() {
+  const context = Context.get()
+  if (context?.isMigrating) {
+    throw new Error(
+      "The context cannot be change, a migration is currently running"
```
samwho, line 195:
```suggestion
      "The context cannot be changed, a migration is currently running"
```

packages/backend-core/src/context/tests/index.spec.ts
```diff
@@ -144,4 +149,107 @@ describe("context", () => {
...
+      ],
+    ])(
+      "a nested context.%s function cannot run",
+      async (_, otherContextCall: () => Promise<void>) => {
+        await expect(
+          context.doInAppMigrationContext(db.generateAppID(), async () => {
+            await otherContextCall()
+          })
+        ).rejects.toThrowError(
+          "The context cannot be change, a migration is currently running"
```
samwho, line 250:
```suggestion
          "The context cannot be changed, a migration is currently running"
```

samwho (approved):
2 tiny suggestions, I don't mind if you choose not to apply them.

mike12345567 (approved):
LGTM - I think this is a great way to guard against migrations doing things outside their expected realms.

PR #12497, opened 2023-12-04
[Chore] Remove conflicting typing

adrinr:
## Description
We are having conflicting types with account portal due to its ts configuration. Removing this type will fix the issue


jvcalderon (approved):
LGTM!

PR #12500, opened 2023-12-04
Cross tenant app denial

mike12345567:
## Description
This resolves issues with cross tenant apps, specifically when interacting with public apps.

addresses: https://linear.app/budibase/issue/BUDI-7811/vulnerability-cross-tenant-cookies

### DIFF
--- a/packages/backend-core/src/tenancy/tenancy.ts
+++ b/packages/backend-core/src/tenancy/tenancy.ts
@@ -93,11 +93,19 @@ export const getTenantIDFromCtx = (
   // subdomain
   if (isAllowed(TenantResolutionStrategy.SUBDOMAIN)) {
     // e.g. budibase.app or local.com:10000
-    const platformHost = new URL(getPlatformURL()).host.split(":")[0]
+    let platformHost
+    try {
+      platformHost = new URL(getPlatformURL()).host.split(":")[0]
+    } catch (err: any) {
+      // if invalid URL, just don't try to process subdomain
+      if (err.code !== "ERR_INVALID_URL") {
+        throw err
+      }
+    }
     // e.g. tenant.budibase.app or tenant.local.com
     const requestHost = ctx.host
     // parse the tenant id from the difference
-    if (requestHost.includes(platformHost)) {
+    if (platformHost && requestHost.includes(platformHost)) {
       const tenantId = requestHost.substring(
         0,
         requestHost.indexOf(`.${platformHost}`)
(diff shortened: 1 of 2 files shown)


aptkingston (approved):
LGTM!

packages/server/src/middleware/currentapp.ts
```diff
@@ -69,28 +70,34 @@ export default async (ctx: UserCtx, next: any) => {
...
-      !tenancy.isUserInAppTenant(requestAppId, ctx.user)
-    ) {
-      // don't error, simply remove the users rights (they are a public user)
-      ctx.user = users.cleanseUserObject(ctx.user) as ContextUser
-      ctx.isAuthenticated = false
-      roleId = roles.BUILTIN_ROLE_IDS.PUBLIC
-    }
+  const userId = ctx.user ? generateUserMetadataID(ctx.user._id!) : undefined
 
+  // if the user not in the right tenant then make to wipe their cookie
```
samwho, line 75:
```suggestion
  // if the user is not in the right tenant then make sure to wipe their cookie
```

Bit of a nit but 👼

samwho (commented):
Is there a way to add tests for this?

Also, do we need to worry about any of the other `TenantResolutionStrategy` types?

mike12345567:
> Is there a way to add tests for this?
> 
> Also, do we need to worry about any of the other `TenantResolutionStrategy` types?

Testing this could be quite difficult, we don't really have a test setup for this right now, but its probably one worth putting with QA-wolf!

The other tenant strategies aren't affected, this was just a problem with local debugging, won't be a production issue.

packages/backend-core/src/tenancy/tenancy.ts
```diff
@@ -93,11 +93,19 @@ export const getTenantIDFromCtx = (
   // subdomain
   if (isAllowed(TenantResolutionStrategy.SUBDOMAIN)) {
     // e.g. budibase.app or local.com:10000
-    const platformHost = new URL(getPlatformURL()).host.split(":")[0]
+    let platformHost
+    try {
+      platformHost = new URL(getPlatformURL()).host.split(":")[0]
+    } catch (err: any) {
+      // if invalid URL, just don't try to process subdomain
+      if (err.code !== "ERR_INVALID_URL") {
```
adrinr, line 101: In which case we want to support invalid URLs?

mike12345567: Just local development, it won't happen in multi-tenant environments, but this was blocking the ability to fetch tenant IDs in other ways

Issue #3826, opened 2021-12-17
Enum type in PostgreSQL, MySQL, etc


mjashanks:
### Discussed in https://github.com/Budibase/budibase/discussions/3192

<div type='discussions-op-text'>

<sup>Originally posted by **micebrain** October 29, 2021</sup>
Hi,
I have several enums in my external Postgresql database. Is it possible for this fields with enum type to have dropdown selection instead of text in new record creation?
Also is it possible to configure certain field to have dropdown selection instead of plain text? E.g. dropdown with values like - Yes, Maybe, No. </div>

tmaiaroto:
Or just empty so we can add a list in validation settings? I have a enum string field it forced to datetime and I can't seem to configure it in the data model or change it anywhere.

mike12345567:
Hi @tmaiaroto - just had a look at this, could you provide a little more info around the types used in your enum? I created an example of this like:
```
CREATE TYPE person_job AS ENUM ('qa', 'programmer', 'designer');
CREATE TABLE People (
    PersonID SERIAL PRIMARY KEY,
    Job person_job
);
```

Budibase loaded the Job field as `Text` correctly, which I was then able to change to options and add the enumeration values (currently Budibase doesn't find custom types in your database). If you could provide a bit of an example of what your enum type looks like I can look into why it ended up as datetime, that seems like a very odd bug. If the type had been selected correctly you would be able to add in a list by changing the type to `Options` - however the base type has to be considered string based for Budibase to allow switching to this.

melohagan:
Could be useful: https://stackoverflow.com/questions/9540681/list-postgres-enum-type

PR #12512, opened 2023-12-06
Support enum types in PostgreSQL and MySQL

melohagan:
## Description
Users wanted to import enum types as an option type in Budibase, with the available options already pre-populated.

I did check SQL Server, but there is no enum type for that database.

## Addresses
- https://github.com/Budibase/budibase/issues/3826

## Schemas used

MySQL:
```sql
CREATE TABLE tickets (
    id INT PRIMARY KEY AUTO_INCREMENT,
    title VARCHAR(255) NOT NULL,
    priority ENUM('Low', 'Medium', 'High') NOT NULL
);
```

Postgres:
```sql
CREATE TYPE color AS ENUM ('red', 'green', 'blue', 'yellow');

ALTER TABLE test
ADD COLUMN test_color color;
```

## Screenshots

MySQL:

Screenshot 2023-12-06 at 11 36 22

Screenshot 2023-12-06 at 11 36 43

Postgres:

Screenshot 2023-12-06 at 11 37 28

Screenshot 2023-12-06 at 11 37 39

*Colours added after fetch*

## Feature branch env
[Feature Branch Link](http://fb-budi-1409-enum-type-in-postgresql.fb.qa.budibase.net)

### DIFF
--- a/packages/server/src/integrations/microsoftSqlServer.ts
+++ b/packages/server/src/integrations/microsoftSqlServer.ts
@@ -17,7 +17,7 @@ import {
 import {
   getSqlQuery,
   buildExternalTableId,
-  convertSqlType,
+  generateColumnDefinition,
   finaliseExternalTables,
   SqlClient,
   checkExternalTables,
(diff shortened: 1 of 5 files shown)


mike12345567 (commented):
looks good - not really critical, but if would be nice if we could make this part of the `convertSqlType` function, if we expand the function definition to allow passing meta information e.g. `convertSqlType(type: string, opts?: { options?: string[] })` then the connector can work out the options list from the database, then the `convertSqlType` can handle setting this information on the Budibase schema, just to keep the schema setting relatively together.

This should avoid constraints etc getting overwritten in future if we ever updated the `convertSqlType` to include more constraints.

mike12345567 (approved):
Looks good to me - thanks for doing that, quite the improvement, feature and refactor!

PR #12522, opened 2023-12-07
Fix Grid Block Toggle Causing the label editing menu to open

Ghrehh:
## Description
Clicking the grid block configuration toggle buttons caused the menu to edit the label to pop up. Stopping event propagation fixes this.

https://linear.app/budibase/issue/BUDI-7819/grid-block-toggle-columns-causes-pop-over-to-appear


melohagan (approved):
LGTM!

aptkingston (approved):

PR #12523, opened 2023-12-07
Prevent collapsed component settings sections reverting on component settings changes.

Ghrehh:
## Description
Changing a component setting would uncollapse all collapsed sections, this should prevent that 

## Addresses
https://linear.app/budibase/issue/BUDI-7821/[-builder-]-settings-section-with-collapsed-true-collapses-with-every

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
@@ -19,6 +19,8 @@
   export let includeHidden = false
   export let tag
 
+  let sections = []
+
   $: sections = getSections(
     componentInstance,
     componentDefinition,
@@ -32,10 +34,12 @@
     const generalSettings = settings.filter(
       setting => !setting.section && setting.tag === tag
     )
+
+    // object pointers getting copied here?
     const customSections = settings.filter(
       setting => setting.section && setting.tag === tag
     )
-    let sections = [
+    let newSections = [
       ...(generalSettings?.length
         ? [
             {
@@ -48,12 +52,13 @@
     ]
 
     // Filter out settings which shouldn't be rendered
-    sections.forEach(section => {
-      section.visible = shouldDisplay(instance, section)
-      if (!section.visible) {
+
+    newSections.forEach(newSection => {
+      newSection.visible = shouldDisplay(instance, newSection)
+      if (!newSection.visible) {
         return
       }
-      section.settings.forEach(setting => {
+      newSection.settings.forEach(setting => {
         setting.visible = canRenderControl(
           instance,
           setting,
@@ -61,12 +66,12 @@
           includeHidden
         )
       })
-      section.visible =
-        section.name === "General" ||
-        section.settings.some(setting => setting.visible)
+      newSection.visible =
+        newSection.name === "General" ||
+        newSection.settings.some(setting => setting.visible)
     })
 
-    return sections
+    return newSections
   }
 
   const updateSetting = async (setting, value) => {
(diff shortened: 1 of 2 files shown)


packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -44,29 +49,35 @@
             },
           ]
         : []),
-      ...(customSections || []),
+      ...(JSON.parse(JSON.stringify(customSections)) || []),
```
Comment (Ghrehh) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:52]: Bit of a stinker here, this was copying component definitions by pointer, and the component identities themselves were getting the `collapsed` property applied to them, which caused weirdness if navigating between two components with the same definition.

Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:52]: Err is there no way to avoid this? Doesn't seem right to me at all.

melohagan (changes requested):

Ghrehh:
@aptkingston No that seems fair and is a good idea, I'll implement it that way.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -19,23 +19,28 @@
...
+    sections
   )
 
   const getSections = (instance, definition, isScreen, tag, includeHidden) => {
     const settings = definition?.settings ?? []
     const generalSettings = settings.filter(
       setting => !setting.section && setting.tag === tag
     )
+
+    // object pointers getting copied here?
```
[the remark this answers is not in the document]
Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:39]: Not 100% what you mean by this haha, this was a `\\ TODO` I left in by accident, my b

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -19,23 +19,28 @@
...
+  let sections = []
+
   $: sections = getSections(
     componentInstance,
     componentDefinition,
     isScreen,
     tag,
-    includeHidden
+    includeHidden,
+    sections
```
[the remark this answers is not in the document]
Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:30]: Yup you're right, left this in by mistake, it's not even consumed by the function that's being called.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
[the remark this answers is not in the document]
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte]: Ah okay no worries - didn't realise it was a TODO!

melohagan (approved):

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -19,6 +19,8 @@
   export let includeHidden = false
   export let tag
 
+  let sections = []
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:22]: Do we still need this change? I'm not sure what it does now that we've gone for this other approach to fix this issue.

Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte]: @aptkingston Whoops, nope, left over. Also removed some renaming changes that I'd made from the previous implementation that aren't strictly necessary now. 

Also just in case it comes up, I checked that no other consumers of this use the event based stuff before I removed it, also no other consumers used the "show" prop so they shouldn't be affected by the change from "show" to "initiallyShow"

If you could review this again when you get a chance please

aptkingston (approved):
LGTM!

PR #12524, opened 2023-12-07
Add another special case for the docker-compose variant of our CouchDB image.

samwho:
## Description

This is in preparation for switching our docker-compose users over from `ibmcom/couchdb3` to `budibase/couchdb`. Because `ibmcom/couchdb3` uses the default data directory, and we mount a volume there in the `docker-compose.yaml`, this change makes the `budibase/couchdb` image also use the defaults when it detects it's running under docker-compose.

This will need a corresponding change to `hosting/docker-compose.yaml` that changes the image and also sets the environment variable `TARGETBUILD=docker-compose`.

In the long term, I think this approach to configuring CouchDB isn't going to scale nicely. It's already quite confusing. It would be nicer to find a way to make the image very neutral/default and then inject custom config in each environment.

This will require republishing `budibase/couchdb:v3.2.1`.


mike12345567 (approved):
LGTM

PR #12525, opened 2023-12-07
Fix attachment cleanup when in development mode

mike12345567:
## Description
This solves an issue with column name updating for attachment types, previous to this update, if a column name was changed for an attachment type when the app was in development mode (no published version) the attachments for the rows would be deleted from the object store - this is likely quite an old bug.

Also fixes a small frontend bug where it wouldn't show you the current name when editing a column, it would always try to build a new "Column 0X" type column name.

I have also refactored most of the attachment cleanup logic as there was a single function that handled all the different update types with a massive number of optional inputs, this has been broken down into a more understandable set of functions.

Mocked tests added as our test environment doesn't really allow for object storage testing today.


samwho (changes requested):
Just the one real change suggested. Good work!

samwho (approved):

Issue #12526, opened 2023-12-08
[BUDI-7823] Curl on Minio-Docker-Image not aviailible anymore


darkspiret:
**Checklist**
- [x ] I have searched budibase discussions and github issues to check if my issue already exists
I found a similar one: https://github.com/Budibase/budibase/pull/12279

**Hosting**
- Self
  - Method: docker-compose.yml + portainer
  - Budibase Version: latest
  - App Version: latest

**Describe the bug**
curl is not aviailible on the minio image anymore, the healthcheck is not working.

Error: 
```
OCI runtime exec failed: exec failed: unable to start container process: exec: "curl": executable file not found in $PATH: unknown
```
**To Reproduce**
1. create a new budibase-stack with the docker-compose.yaml : https://raw.githubusercontent.com/Budibase/budibase/master/hosting/docker-compose.yaml || https://github.com/Budibase/budibase/blob/master/hosting/docker-compose.yaml
2. the minio container will be unhealthy

**Solution**

change the healtchek from:
```
old: test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
```
to
```
new: test: timeout 5s bash -c ':> /dev/tcp/127.0.0.1/9000' || exit 1
```

<sub>[BUDI-7823](https://linear.app/budibase/issue/BUDI-7823/curl-on-minio-docker-image-not-aviailible-anymore)</sub>

melohagan:
Thanks @darkspiret but closing as duplicate of: https://github.com/Budibase/budibase/issues/12491

PR #12533, opened 2023-12-08
Update docker-compose.yaml to use our CouchDB image instead of IBM's.

samwho:
## Description

This required a few changes to the CouchDB image which have been made in others PRs. This was in order to make sure that CouchDB is looking in the right places in the filesystem to make this the smoothest transition it can be.

When users upgrade to this using `budi hosting --update` it should Just Work™.

I was also able to get rid of the `couch-init` service because what it does now happens from within the CouchDB container.

I've also set the `pull_policy` for the CouchDB image to "always" because the label has historically been mutated.

## Addresses
- https://linear.app/budibase/issue/BUDI-7799/sqs-investigate-docker-compose-self-host-migration
- https://github.com/Budibase/budibase/issues/12526

### DIFF
--- a/hosting/docker-compose.yaml
+++ b/hosting/docker-compose.yaml
@@ -69,7 +69,7 @@ services:
       MINIO_BROWSER: "off"
     command: server /data --console-address ":9001"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
+      test: "timeout 5s bash -c ':> /dev/tcp/127.0.0.1/9000' || exit 1"
       interval: 30s
       timeout: 20s
       retries: 3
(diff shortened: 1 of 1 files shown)


hosting/docker-compose.yaml
```diff
@@ -98,26 +97,15 @@ services:
 
   couchdb-service:
     restart: unless-stopped
-    image: ibmcom/couchdb3
+    image: budibase/couchdb:v3.2.1
```
mike12345567, line 100: Wondering if we should hard code this - when we release SQS we'll want it to auto upgrade to that version - if we get people to update their docker-compose files sooner then hopefully it should just be an image upgrade. Since this image tag is controlled by us, the latest should always be fine.

mike12345567 (approved):
Very nice! One minor comment, so happy to see the back of `couch-init`!

hosting/docker-compose.yaml
samwho: I did um and ahh about this. Removed. 🙂

PR #12546, opened 2023-12-11
remove pricing banner

shogunpurple:
## Description
Just removing the pricing banner telling folks we are updating our pricing. After the email and other communications it's not really necessary anymore.


jvcalderon (approved):
LGTM!

PR #12550, opened 2023-12-11
Add `isSqsAvailable` key to the environment endpoint.

samwho:
## Description

This PR adds `isSqsAvailable` as a boolean key to the `/api/system/environment` endpoint. It works by attempting to fetch the SQS endpoint and, if it throws an exception, returns false. Any successful HTTP response, including errors, indicates that SQS is available.

I've decided to cache the result of checking for SQS availability indefinitely in memory. This might not be the right choice, but at least doesn't feel too incorrect. 

## Addresses
- This is the backend part of the work for https://linear.app/budibase/issue/BUDI-7664/sqs-self-host-ui-for-detecting-lack-of-sqs-support.


mike12345567 (changes requested):
Nice very quick! Only thing that needs done is making sure this only executes in self host, we don't want to try and collect this information or make the user aware in the Cloud - the infrastructure should always be correct for the service deployed in the cloud. We can add infrastructure information only if the `SELF_HOSTED` flag is set.

Also NAB, just an idea, but I think we could possibly prepare for any other infrastructure updates, like MinIO updates, and add an `infrastructure` section to the environment call, like `infrastructure: { sqs: { available: true } }` - using an object for marking status could be useful if we want to return any other information such as a specific error if something isn't functioning correctly.

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -240,6 +241,34 @@ class TestConfiguration {
     return { message: "Admin user only endpoint.", status: 403 }
   }
 
+  async withEnv(newEnvVars: Partial<typeof env>, f: () => Promise<void>) {
+    let cleanup = this.setEnv(newEnvVars)
+    try {
+      await f()
+    } finally {
+      cleanup()
```
adrinr, line 249: I love this approach!

packages/worker/src/tests/TestConfiguration.ts
```diff
@@ -36,6 +36,7 @@ import {
 } from "@budibase/types"
 import API from "./api"
 import jwt, { Secret } from "jsonwebtoken"
+import { cloneDeep } from "lodash"
```
adrinr, line 39: Can we use `import cloneDeep from "lodash/fp/cloneDeep"` for a better tree shaking and hoisting?

packages/worker/src/api/controllers/system/environment.ts
```diff
@@ -1,6 +1,25 @@
...
+import nodeFetch from "node-fetch"
+
+let sqsAvailable: boolean
+async function isSqsAvailable() {
+  if (sqsAvailable !== undefined) {
+    return sqsAvailable
+  }
+
+  try {
+    await nodeFetch(coreEnv.COUCH_DB_SQL_URL, {
```
adrinr, line 13: Will SQS be always available per instance, and not per tenant? And can we have a DDoS issue on worker restarts, if there are many requests and a slow sqs instance?

samwho: It's my understanding that SQS will be a global thing. We're baking it in to the CouchDB image for self-hosters, unsure what we're doing in Budicloud.

I don't think this poses a DDoS concern. The request to `/` is very fast (it returns a few metrics) and this request is sent once per lifetime of a pod.

packages/worker/src/tests/TestConfiguration.ts
samwho: I understand what tree shaking is and how this would help, how does this help with hoisting? I only have a rough idea of what hoisting is 🙈

packages/worker/src/api/controllers/system/environment.ts
mike12345567: Yeah SQS is global, the whole self host cluster will need SQS support. I don't think this poses a threat of DDoS either as it only runs in self host, and in most installations there is max 1-2 worker pods, if they get stuck in a restart look its still only a few hundred requests max.

mike12345567 (approved):
LGTM

PR #12551, opened 2023-12-11
feat: use /storage as data_dir for aas builds

calexiou:
## Description
This is a workaround for the `budibase/budibase-aas` image in order to avoid the API restriction that Azure App Service has for mounting external storage to `/home` when a service is created through their client API.

### DIFF
--- a/hosting/single/runner.sh
+++ b/hosting/single/runner.sh
@@ -22,7 +22,7 @@ declare -a DOCKER_VARS=("APP_PORT" "APPS_URL" "ARCHITECTURE" "BUDIBASE_ENVIRONME
 
 # Azure App Service customisations
 if [[ "${TARGETBUILD}" = "aas" ]]; then
-    export DATA_DIR="${DATA_DIR:-/home}"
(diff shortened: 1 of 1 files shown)


calexiou:

calexiou:
recheck

calexiou:
It will, since we're using the `:-` operator. 

The point of the PR is that the `DATA_DIR` is only important to a user who needs storage persistence.  Otherwise, any random path in the filesystem, even `/tmp` would accommodate the needs of someone who doesn't care about persistence. 

Setting the default to`/home` is shipping a broken default by design, since Azure App Service restricts mounts to `/home`.

samwho:
It's worth noting that the use of `TARGETBUILD` has become a bit warped over time. It used to just be used as something set during build time, but over time some environments need it setting at runtime as well. For anything in the runner script, you need to make sure `TARGETBUILD` is set at runtime.

shogunpurple:
@calexiou is this still relevant? Can we close?

Issue #12479, opened 2023-11-30
[BUDI-7788] "Can only search by string.email or equal._id"


andyburgessmd:
**Checklist**

* I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker
  * Budibase Version: 2.13.26 - but issue present from around 2.13.18
  * App Version: 2.13.26 - but issue present from around 2.13.18

**Describe the bug**
I have a data provider (internalDB) which has a filter of Owner (user type) equals binding on a form field which is a user object.

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/df24b0ec-3ef7-47f1-828a-bc1f46a9fd3f/73aea97f-29f7-416c-8152-41ebbfe29a04?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9kZjI0YjBlYy0zZWY3LTQ3ZjEtODI4YS1iYzFmNDZhOWZkM2YvNzNhZWE5N2YtMjlmNy00MTZjLTgxNTItNDFlYmJmZTI5YTA0IiwiaWF0IjoxNzAxMzUzODMwLCJleHAiOjE3MDE0NDAyMzB9.0WAGiwfdGOgkj8UrDVncaLl3qGa40FNht1ITQJ6ArY0)

[...]

PR #12553, opened 2023-12-11
Update global users search to account for numeric prefixing

aptkingston:
## Description
This PR updates the global user search endpoint to account for numeric prefixing. Currently it does not account for it, meaning the usual numerically prefixed search fields of something like `1:email` will be rejected as it does not match the whitelist of supported filter fields which are only `email` and `_id`.

With this change, the endpoint will now sanitise the search query by removing the prefixing before proceeding.

I've added a few test cases to account for this new functionality.

## Addresses
- https://github.com/Budibase/budibase/issues/12479


melohagan (approved):
LGTM! Just an import that needs tidied up according to the linter 👍

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -185,9 +186,23 @@ export const getAppUsers = async (ctx: Ctx<SearchUsersRequest>) => {
...
+      if (filters && typeof filters === "object") {
+        for (let [field, value] of Object.entries(filters)) {
+          delete filters[field]
+          filters[removeKeyNumbering(field)] = value
+        }
+      }
+    }
+    // Validate we aren't trying to search on any illegal fields
+    if (!userSdk.core.isSupportedUserSearch(body.query)) {
+      ctx.throw(501, "Can only search by string.email or equal._id")
```
samwho, line 204: 501 isn't the correct status code here. It's for when a specific HTTP verb hasn't been implemented by a route (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501). I think 400 is more appropriate here.

samwho: Also would be great to have a test that hits this condition.

samwho (changes requested):

packages/worker/src/api/controllers/global/users.ts
aptkingston: I didn't write this original feature (I'm just adjusting it to account for the numeric prefixing) but I'm happy to change that status code over.

There already is a test that hits the error case from the original implementation. The 2 tests I've added are just to specifically test the numeric prefixing removal but there already are a few existing tests, including one that hits the 501 path.

aptkingston: Changed to 400 👍

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -185,9 +186,23 @@ export const getAppUsers = async (ctx: Ctx<SearchUsersRequest>) => {
...
+  // TODO: for now only two supported search keys; string.email and equal._id
+  if (body?.query) {
+    // Clean numeric prefixing. This will overwrite duplicate search fields,
+    // but this is fine because we only support a single custom search on
+    // email and id
+    for (let filters of Object.values(body.query)) {
+      if (filters && typeof filters === "object") {
+        for (let [field, value] of Object.entries(filters)) {
+          delete filters[field]
+          filters[db.removeKeyNumbering(field)] = value
```
mike12345567, line 198: Is there a risk of this defeating the purpose of the numeric keying, e.g. if we have sent up `999:email` and `1:email`, this will flatten them in an order depending on how the object gets enumerated (likely deterministic but possibly confusing) we could probably write a quick test case and adjust the behaviour in someway to handle it.

aptkingston: It will yea, but I explained in the comment that the underlying search mechanism here only supports a single filter anyway, so flattening is fine since we would never have used both in the first place.

mike12345567: Should we add a case to prefer the `999:email` over all others? As this is one set by the system this is really the one we want to use.

aptkingston: Hmm the `999` is the one sent by relationship pickers (large just to avoid clashing with other filters), but if you just did a normal filter yourself on the users table then you could easily sent up 2 email filters with a `1` and `2` prefix for example, and then you still wouldn't know which would win. I'm leaning towards just erroring for correctness here I think, since that's the only way to avoid us having to be opinionated in some hidden way.

mike12345567: yeah I think that works, an error makes it easier to understand what is happening if we have to debug it at least!

mike12345567 (approved):
LGTM - just need error as discussed.

PR #12554, opened 2023-12-11
Split out modifying coreEnv in withEnv

samwho:
## Description

`TestConfiguration.withEnv` in `packages/server` up until now has modified both `env` and `coreEnv`. I've realised this isn't actually correct, the two don't strictly overlap.

I have created a new `withCoreEnv` to address this.


adrinr:
What is the use case when we want to keep different settings between core and env?

samwho:
I'm not sure, but they are different types. This indicates to me that it is not always valid to set the same keys in both.

mike12345567 (approved):

PR #12556, opened 2023-12-12
Revert TTL on signed urls

melohagan:
## Description
Some users have reported not being able to access attachments before 29th November. Reverting the TTL change on the signed URLs to see if that fixes this issue.


shogunpurple (approved):
LGTM

PR #12559, opened 2023-12-12
Fix external formulas and issues with external relationship configuration

mike12345567:
## Description

This addresses two issues I found in this area:
1. The backend did not correctly handle the updating of dynamic formulas on external tables, you could add them, but if you tried to update the formula or name (both very valid operations) you would receive errors - this is now resolved.
2. A variety of frontend issues, all related to relationship configuration. Due to the way existing relationships are created for external tables (directly on the datasource) the tables were not correctly getting updated. I've added websocket emissions of these tables to update clients when tables are updated.
3. I have tried to reduce the usage of the datasource tables (`entities`) and move to using the raw table API where possible.
4. Lastly, I have attempted to fix an issue with formula creation in the grid, today when a column is added it doesn't refresh the rows. I have added a refresh for now, but this could be up for discussion.

Addresses: https://linear.app/budibase/issue/BUDI-7760/relationship-lost-after-adding-a-new-column-and-add-data-to-it


packages/builder/src/components/backend/DataTable/buttons/ExistingRelationshipButton.svelte
```diff
@@ -22,13 +27,18 @@
     })
   }
 
-  const afterSave = ({ action }) => {
+  const afterSave = async ({ action }) => {
```
aptkingston, line 30: This doesn't need to be async, but it does no harm either.

packages/builder/src/components/backend/DataTable/modals/grid/GridCreateColumnModal.svelte
```diff
@@ -2,7 +2,13 @@
...
-  const { datasource } = getContext("grid")
+  const { datasource, rows } = getContext("grid")
 </script>
 
-<CreateEditColumn on:updatecolumns={datasource.actions.refreshDefinition} />
+<CreateEditColumn
+  on:updatecolumns={async () => {
+    await datasource.actions.refreshDefinition()
+    // Need to find a better way to handle formula columns than just always refreshing
+    rows.actions.refreshData()
```
aptkingston, line 12: I would prefer to leave this out and try to find some other way. Not only will this increase API load substantially, it will also constantly reset the users current scroll positions and paginated data. We could maybe add a unique event for when formula columns are created/changed and use that to trigger data refreshes?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -245,6 +247,15 @@ export async function update(ctx: UserCtx<any, UpdateDatasourceResponse>) {
...
   builderSocket?.emitDatasourceUpdate(ctx, datasource)
+  // send table updates if they have occurred
+  if (datasource.entities) {
+    for (let table of Object.values(datasource.entities)) {
+      const oldTable = baseDatasource.entities?.[table.name]
+      if (!oldTable || !isEqual(oldTable, table)) {
+        builderSocket?.emitTableUpdate(ctx, table, { includeOriginator: true })
+      }
+    }
+  }
```
aptkingston, line 258: Cool, I wasn't aware we ever made schema updates against datasources directly. Since this is checking for diffs it'll hopefully not send any erroneous or duplicate events.

aptkingston (approved):
LGTM, apart from the refreshing all rows when making any schema change. I'd like to tackle that issue with something else, but happy to merge the rest!

packages/builder/src/components/backend/DataTable/buttons/ExistingRelationshipButton.svelte
mike12345567: woops, don't know why I made it async!

packages/server/src/api/controllers/datasource.ts
mike12345567: Its really only in regards to relationships, due to the way existing relationships are defined the table API can't really handle it today (requires updating multiple tables at the same time) we could probably build a table API specifically for this, but emitting the table events seems like a good fix for now.

PR #12563, opened 2023-12-12
App migration builder frontend

adrinr:
## Description
Adding the frontend view of the app migration. When an app is migrating, the user will be redirected to a loading page. This page will be pulling the migration status until all the migrations are properly applied.

TODO:
- [x] Frontend design

## Addresses
- [BUDI-7654 - (SQS) Build app migration frontend](https://linear.app/budibase/issue/BUDI-7654/sqs-build-app-migration-frontend)

### Updating
image

### Migration timeout
image

### Flow
https://github.com/Budibase/budibase/assets/15987277/0a480daf-fc97-415a-95ba-4262546e41ed

### Light theme
image

### Nord theme
image

## Feature branch env
[Feature Branch Link](http://fb-budi-7654-app-migration-builder-frontend.fb.qa.budibase.net)


packages/builder/src/api.js
```diff
@@ -45,4 +45,15 @@ export const API = createAPIClient({
...
     }
   },
+  onMigrationDetected: appId => {
+    const updatingUrl = `/builder/app/${appId}/updating`
+
+    if (window.location.pathname === updatingUrl) {
+      return
+    }
+
+    window.location = `${updatingUrl}?returnUrl=${encodeURIComponent(
```
aptkingston, line 55: We can use routify here can't we? So we could use `goto` rather than doing a hard window.location navigation. `goto` also will handle params for us.
https://www.routify.dev/docs/helpers#goto

packages/builder/src/pages/builder/app/[application]/updating.svelte
```diff
@@ -0,0 +1,31 @@
...
+  setInterval(async () => {
+    const response = await API.get({ url: "/api/migrations/status" })
+    if (!response.migrated) {
+      return
+    }
+
+    const urlParams = new URLSearchParams(window.location.search)
+    const returnUrl = urlParams.get("returnUrl")
+
+    window.location = returnUrl
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/updating.svelte:14]: We could use routify's `redirect` here to redirect the user and prevent them going back to this page, rather than adding another entry to the browsers history by just navigating away again. Might also be sensible to add a fallback URL in case a return URL doesn't exist, to just push them to the main builder URL for that app or something.

packages/builder/src/pages/builder/app/[application]/updating.svelte
```diff
@@ -0,0 +1,31 @@
...
+  import { API } from "api"
+
+  setInterval(async () => {
+    const response = await API.get({ url: "/api/migrations/status" })
+    if (!response.migrated) {
+      return
+    }
+
+    const urlParams = new URLSearchParams(window.location.search)
+    const returnUrl = urlParams.get("returnUrl")
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/updating.svelte:12]: Routify provides URL and query params via the `$params` store, so we should make use of those. I'd imaging that simply `$params.returnUrl` here is the value we're after.
https://www.routify.dev/guide/introduction/parameters

aptkingston (commented):
Few suggestions around using routify to simplify some of the manual navigation, but logic looks good to me!

adrinr:
@aptkingston I created https://github.com/Budibase/budibase/pull/12577 to separate the builder and client concerns from the actual backend. It will be easier to review and discuss

packages/builder/src/pages/builder/app/[application]/updating.svelte
Comment (adrinr) [reply packages/builder/src/pages/builder/app/[application]/updating.svelte]: Added the usage. For some reason it's not stripping the `?` properly. Any idea why?

packages/builder/src/pages/builder/app/[application]/updating.svelte
```diff
@@ -0,0 +1,44 @@
...
+      const returnUrl = $params.returnUrl || $params["?returnUrl"]
+      $redirect(returnUrl)
+    }, 1000)
+  }
+
+  checkMigrationsFinished()
+
+  function migrationTimeout() {
+    // TODO
+    alert("Something went wrong 💀")
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/updating.svelte:31]: - [ ] TODO

packages/frontend-core/src/components/Updating.svelte
```diff
@@ -0,0 +1,61 @@
...
+  checkMigrationsFinished()
+
+  function migrationTimeout() {
+    timedOut = true
+  }
+
+  $: text = !timedOut ? "System update" : "Something went wrong!"
+  $: subtext = !timedOut
+    ? "Please wait and we will be back in a second!"
+    : "An error occurred, please try again later"
```
samwho, line 33: Is there something more helpful we can say here? Ask them to email support and give them some logs to show them, maybe? I can see users being frustrated here if it happens repeatedly and they're locked out of their app.

packages/frontend-core/src/components/Updating.svelte
```diff
@@ -0,0 +1,61 @@
+<script>
+  export let isMigrationDone
+  export let onMigrationDone
+  export let timeoutMs = 180000 // 3 minutes
```
samwho, line 4: Would it be worth making this `timeoutSeconds` or `timeoutMinutes`? Parsing 180000 milliseconds takes a few seconds for me.

adrinr: I usually like using ms, as it's usually the standard. But I can change it in here, as we will never worry about ms in here

samwho (approved):

mike12345567 (approved):

PR #12567, opened 2023-12-13
Improving writethrough cache typing and removing useless log statement

mike12345567:
## Description
Improving types around the writethrough cache, exposing the proper typing from the database and removing a log statement which is useless, errors are correctly propagated if they matter (and handled above this level with proper context) or in the 99% scenario it is not a real error (such as try once locks already being held) and a log is just spammy.

There is also a pro component to this: https://github.com/Budibase/budibase-pro/pull/246

### DIFF
--- a/packages/backend-core/src/cache/writethrough.ts
+++ b/packages/backend-core/src/cache/writethrough.ts
@@ -7,7 +7,7 @@ import * as locks from "../redis/redlockImpl"
 const DEFAULT_WRITE_RATE_MS = 10000
 let CACHE: BaseCache | null = null
 
-interface CacheItem {
+interface CacheItem<T extends Document> {
   doc: any
   lastWrite: number
 }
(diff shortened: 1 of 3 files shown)


samwho (approved):

packages/backend-core/src/cache/writethrough.ts
```diff
@@ -84,16 +84,16 @@ async function put(
...
   const cacheKey = makeCacheKey(db, id)
   let cacheItem: CacheItem = await cache.get(cacheKey)
   if (!cacheItem) {
-    const doc = await db.get(id)
+    const doc = await db.get<T>(id)
     cacheItem = makeCacheItem(doc)
     await cache.store(cacheKey, cacheItem)
   }
-  return cacheItem.doc
+  return cacheItem.doc as T
```
adrinr, line 96: Is this not typed by default?

adrinr (approved):

packages/backend-core/src/cache/writethrough.ts
mike12345567: No it wasn't, but I've switched from casting it to making the typing part of the `CacheItem` - which is a bit better.

PR #12572, opened 2023-12-13
HBS/JS UUID helper

mike12345567:
## Description
As the title says, this adds a helper which can be used to generate a UUID in both the frontend and backend without needing to rely on the crypto library (which may not be available in all scenarios).

Works in both handlebars (`{{ uuid }}`) and JS (`return helpers.uuid()`).

image


packages/string-templates/src/processors/preprocessor.js
```diff
@@ -71,6 +71,9 @@ module.exports.processors = [
     ) {
       insideStatement = `(${insideStatement})`
     }
+    if (statement === "{{ uuid }}") {
```
melohagan, line 74: Maybe this should be at the top of the function to better show intent, as the `insideStatement` does not need to be calculated for this path.

Also are we sure that this statement couldn't be `{{uuid}}` for example?

melohagan (approved):
LGTM!

shogunpurple (approved):
LGTM - although where did the need for this come from? Is there a ticket or doc anywhere?

mike12345567:
> LGTM - although where did the need for this come from? Is there a ticket or doc anywhere?

was just a quick evening project, seen a few people now asking for now to create UUIDs, would rather they do it through our system than trying to import random libraries!

packages/string-templates/src/processors/preprocessor.js
mike12345567: sorry this wasn't needed - was introduced during testing, removed!

PR #12578, opened 2023-12-14
Allow tagging PRs that require feature branches deployments

adrinr:
## Description
We recently created our feature branch functionality, creating FB branches for any single PR to master. This is causing heavy loads on our infra and many of these FB are not even used. This might be because there are just backend changes, cleanups, documentation, etc.
We will be adding a new `feature-branch` label into the PRs that will mark which ones need to be deployed as FB

## Feature branch env
[Feature Branch Link](http://fb-chore-create-fb-release-label.fb.qa.budibase.net)


.github/workflows/deploy-featurebranch.yml
```diff
@@ -2,12 +2,19 @@ name: deploy-featurebranch
 
 on:
   pull_request:
-    branches:
-      - master
+    types: [
+        labeled,
+        # default types below (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)
+        opened,
+        synchronize,
```
shogunpurple, line 9: Just to check - if we:
- tag a branch as feature branch
- push to that branch again

It will rebuild?

shogunpurple (approved):
LGTM! Just a question to make sure I'm getting the automation

.github/workflows/deploy-featurebranch.yml
adrinr: Yes, it will. The action checks (on created, updated or labelled) if the PR has the `feature-branch` label. If so, it triggers the deploy

PR #12580, opened 2023-12-14
Optimise getUniqueByProp

samwho:
## Description

We've found that `getUniqueByProp` is consuming a lot of CPU time in Budicloud. This PR presents an alternate implementation that is a lot faster.

image

## Runtime comparison

<details>

```
data size: 0
newFunc: 0ms
oldFunc: 0ms

data size: 1000
newFunc: 6ms
oldFunc: 250ms

data size: 2000
newFunc: 8ms
oldFunc: 701ms

data size: 3000
newFunc: 15ms
oldFunc: 1594ms

data size: 4000
newFunc: 15ms
oldFunc: 2788ms

data size: 5000
newFunc: 22ms
oldFunc: 4397ms

data size: 6000
newFunc: 27ms
oldFunc: 6136ms

data size: 7000
newFunc: 29ms
oldFunc: 8382ms

data size: 8000
newFunc: 33ms
oldFunc: 11088ms

data size: 9000
newFunc: 49ms
oldFunc: 14300ms
```

And here's the code I used to do it:

```js
function newFunc(array, prop) {
  const seen = new Set();
  const ret = [];
  for (const item of array) {
    if (!seen.has(item[prop])) {
      seen.add(item[prop]);
      ret.push(item);
    }
  }
  return ret;
}

function oldFunc(array, prop) {
  return array.filter((obj, pos, arr) => {
    return arr.map((mapObj) => mapObj[prop]).indexOf(obj[prop]) === pos;
  });
}

// adapted from https://stackoverflow.com/a/4644364/205770
const bench = function (iterations, f) {
  let time = 0;
  let timer = function (action) {
    var d = Date.now();
    if (time < 1 || action === "start") {
      time = d;
      return 0;
    } else if (action === "stop") {
      var t = d - time;
      time = 0;
      return t;
    } else {
      return d - time;
    }
  };

  let result = [];
  let i = 0;
  timer("start");
  while (i < iterations) {
    result.push(f());
    i++;
  }

  return {
    time: timer("stop"),
    result: result[0],
  };
};

const genData = function (n) {
  let data = [];
  for (let i = 0; i < n; i++) {
    data.push({ a: i });
  }
  return data;
};

const cmpData = function (a, b) {
  if (a.length !== b.length) {
    return false;
  }
  for (let i = 0; i < a.length; i++) {
    if (!Object.is(a[i], b[i])) {
      return false;
    }
  }
  return true;
};

const iterations = 100;

for (let i = 0; i < 10; i++) {
  const data = genData(1000 * i);

  const tNew = bench(iterations, () => newFunc(data, "a"));
  const tOld = bench(iterations, () => oldFunc(data, "a"));
  if (!cmpData(tNew.result, tOld.result)) {
    console.log("ERROR: results do not match");
    console.log(tNew.result);
    console.log(tOld.result);
  }

  console.log(`data size: ${data.length}`);
  console.log(`newFunc: ${tNew.time}ms`);
  console.log(`oldFunc: ${tOld.time}ms`);
  console.log();
}
```

</details>

## Feature branch env
[Feature Branch Link](http://fb-optimise-get-unique-by-prod.fb.qa.budibase.net)


packages/server/src/db/linkedRows/linkUtils.ts
```diff
@@ -99,9 +99,15 @@ export async function getLinkDocuments(args: {
 }
 
 export function getUniqueByProp(array: any[], prop: string) {
-  return array.filter((obj, pos, arr) => {
-    return arr.map(mapObj => mapObj[prop]).indexOf(obj[prop]) === pos
-  })
+  const seen = new Set()
+  const ret = []
```
melohagan, line 103: NAB: Can we rename this variable, or maybe there is a way to return a set as an array in JS?

melohagan (approved):
LGTM - would be nice to see a unit test for good measure

packages/server/src/db/linkedRows/linkUtils.ts
samwho: It's not the set that gets returned, it's a filtered version of the original array passed in. How do you feel about `filteredArray`?

mike12345567 (approved):
Class, would never find stuff like this without the CPU profiling!

samwho:
@melohagan this function is eventually called through the `outputProcessing` function, which is already tested in a lot of different ways through other tests. On the backend side of Budibase we tend to test things through calls to the public API, rather than having granular unit tests.

adrinr (approved):
LGTM! Great improvement 😅
Wondering if we could not do something cleaner with a reduce, but these results are great

packages/server/src/db/linkedRows/linkUtils.ts
melohagan: Yeah got it! `filteredArray` is spot on then 👍

aptkingston (approved):
Lovely 👌 Orders of magnitude faster! That original implementation is very questionable 😂

PR #12581, opened 2023-12-14
App migration app frontend

adrinr:
## Description
Adding the frontend logic for migrating apps. The flow will be the following:
1. A user requests to load an app
2. If no migration is required, the app is served properly
3. Otherwise, we load a "updating" page.
4. This page will be checking in intervals when the migration is done
5. Once done, the browser will reload, serving the app properly

If the app is already loaded when a new migration is uploaded, the browser will run a page refresh. This will trigger the flow above.

TODO:
- [x] Frontend design

## Addresses
- [BUDI-7654 - (SQS) Build app migration frontend](https://linear.app/budibase/issue/BUDI-7654/sqs-build-app-migration-frontend)

### Updating
image

### Migration timeout
image

## Feature branch env
[Feature Branch Link](http://fb-budi-7654-app-migration-app-frontend.fb.qa.budibase.net)

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -37,7 +37,6 @@
     "downloadjs": "1.4.7",
     "html5-qrcode": "^2.2.1",
     "leaflet": "^1.7.1",
-    "regexparam": "^1.3.0",
     "sanitize-html": "^2.7.0",
     "screenfull": "^6.0.1",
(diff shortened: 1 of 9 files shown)


mike12345567 (approved):
LGTM - nice and easy for the client side which is great!

PR #12583, opened 2023-12-14
Allow cluster internal traffic to hit the nginx status endpoint.

samwho:
## Description

DataDog is currently unable to hit the nginx status endpoint because it doesn't originate from the loopback address. Adding `10.0.0.0/8` to the allow list for port 81 on our nginx pod to rectify this.


shogunpurple (approved):
LGTM

PR #12584, opened 2023-12-14
Massively improve performance of hovering components

aptkingston:
## Description
This PR fixes significant performance issues with the [recent PR](https://github.com/Budibase/budibase/pull/12465) to implement synchronisation between the client preview hovered component and the component tree in the builder. This PR had 2 major issues - it used the `builderStore` in the client to store the hovered component ID, and it passed in the hovered component ID via reinitialising the preview.

These both led to massive computation by the client library whenever the hovered component changed. Since that's something that changes rapidly, it could heavily degrade the whole browser performance when quickly moving the mouse between components.

This PR resolves these issues while still keeping the new synchronisation feature. This is accomplished by:
- Using a new tiny store in the client library to store the hovered component ID, that nothing depends on or observes other than what is necessary
- Using events to pass the hovered component ID between the client and builder rather than a reinitialisation of the preview.


samwho:
Sorry, removing myself from the time being. Focusing on the Budicloud stuff atm. 🙏

aptkingston:
@samwho the budibase/backend review request seems forced and automatic - I didn't desire or intend to request a review so apologies!

samwho:
No that's okay, it's automatic on purpose. It's the effect of the new CODEOWNERS stuff we're testing, it'll assign a reviewer from the backend folks if you've modified backend code. 🙂

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -85,14 +85,16 @@ const INITIAL_FRONTEND_STATE = {
...
 
   // Client state
   selectedComponentInstance: null,
 
   // Onboarding
   onboarding: false,
   tourNodes: null,
+
+  // UI state
+  hoveredComponentId: null,
```
deanhannigan, line 97: We could possibly go one step further with this and mirror the dedicated hover store approach you made in the client. The constant updates to the builderStore also triggers quite a bit of computation on the builder side. This could also be addressed in the monolith ticket.

deanhannigan (approved):
LGTM!

PR #12585, opened 2023-12-14
Multi-step form block

aptkingston:
## Description
This PR adds a multi-step form block component. 

With the `Multi-step Form Block` you can now build and configure multistep forms with a single component.

- Customise the `Title`, `Description`, `Fields` and `Buttons` for each step you create.
- Use the step controls in the component settings to browse and edit your form steps 

Screenshot 2024-01-04 at 16 30 58

- Automatic navigation buttons
  - When creating a new step, the component will add in `Back`, `Next` or `Save` buttons where necessary. These can be customised or removed as required. 

  Screenshot 2024-01-03 at 17 14 33

  - By default, if you add a step to the end it will also include a `Save` button with preset `Validate Form`,`Save Row` and `Clear Form` button actions.
  - Default step buttons will be replaced with static buttons if you customise them.
- `Current Step` binding available
  - Step `Title` value uses this binding in the default e.g `Step 2` 

## Addresses
- https://linear.app/budibase/issue/BUDI-7532/form-block-add-multi-step-functionality

## Screenshots

Building a multi-step form
multistep_designer

## Feature branch env
[Feature Branch Link](http://fb-feature-multistep-form-block.fb.qa.budibase.net)


Ghrehh:
Looks great so far, some things I noticed though:

- Dragging from the button list to the fields list in the builder for a multistep component breaks both lists until page refresh and seems to delete the dragged button. Not a huge issue tbh, stupid games stupid prizes, but maybe worth looking in to.

- The "toggle all fields" toggle seems to maintain the state of the previous step. If you toggle all fields off on step 1, then navigate to step 2, the toggle will remain in the "off" position, even if all the fields in step 2 are enabled.

Ghrehh (approved):
LGTM!

PR #12592, opened 2023-12-15
Tweak newly added spans to track formula processing.

samwho:
## Description

After looking at the spans in the QA environment, I added some extra tags and changed the way I'm attaching the app ID to the trace. When creating an new trace at the middleware level, it seems to become orphaned and shows up strangely in DataDog. I want to see if attaching to an existing span works better.

### DIFF
--- a/packages/server/src/middleware/currentapp.ts
+++ b/packages/server/src/middleware/currentapp.ts
@@ -15,109 +15,106 @@ import { UserCtx, ContextUser } from "@budibase/types"
 import tracer from "dd-trace"
 
 export default async (ctx: UserCtx, next: any) => {
-  return tracer.trace("currentapp middleware", {}, async span => {
-    // try to get the appID from the request
-    let requestAppId = await utils.getAppIdFromCtx(ctx)
-    if (!requestAppId) {
-      return next()
-    }
+  // try to get the appID from the request
+  let requestAppId = await utils.getAppIdFromCtx(ctx)
+  if (!requestAppId) {
+    return next()
+  }
 
-    if (requestAppId) {
-      span?.addTags({ app_id: requestAppId })
-    }
+  if (requestAppId) {
+    const span = tracer.scope().active()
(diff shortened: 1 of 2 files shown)


packages/server/src/utilities/rowProcessor/utils.ts
```diff
@@ -77,7 +78,7 @@ export function processFormulas<T extends Row | Row[]>(
           rows[i] = {
             ...row,
             [column]: tracer.trace("processStringSync", {}, span => {
-              span?.addTags({ column })
+              span?.addTags({ table_id: table._id, column, statis: isStatic })
```
adrinr, line 81:
```suggestion
              span?.addTags({ table_id: table._id, column, static: isStatic })
```

packages/server/src/middleware/currentapp.ts
```diff
@@ -15,109 +15,106 @@ import { UserCtx, ContextUser } from "@budibase/types"
...
+  if (!requestAppId) {
+    return next()
+  }
 
-    if (requestAppId) {
-      span?.addTags({ app_id: requestAppId })
-    }
+  if (requestAppId) {
+    const span = tracer.scope().active()
+    span?.addTags({ app_id: requestAppId })
```
adrinr, line 26:
```suggestion
    span.addTags({ app_id: requestAppId })
```

adrinr (approved):
Just 2 small suggestions

shogunpurple (approved):

packages/server/src/middleware/currentapp.ts
samwho: My editor suggested that span may not be present, so I used the ?

How come this suggestion removes it?

samwho: Double checked just to be sure 😅 

CleanShot 2023-12-15 at 13 14 17@2x

PR #12594, opened 2023-12-15
Instrument CouchDB client with DD APM traces.

samwho:
## Description

This PR wraps the existing `DatabaseImpl` with a new `DDInstrumentedDatabase` which also implements `Database` but wraps all of the calls in `tracer.trace` calls so they'll show up in the DataDog APM traces.


packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -38,7 +39,10 @@ export function DatabaseWithConnection(
   if (!connection) {
     throw new Error("Must provide connection details")
   }
-  return new DatabaseImpl(dbName, opts, connection)
+  return new DDInstrumentedDatabase(
```
adrinr, line 42: We use these calls constantly. What is going to be the performance impact of this?

samwho: I'm uncertain. If it's significant it will show  up in the metrics in DataDog and we can roll it back.

For what it's worth, Redis calls are also instrumented this way it's just that DataDog injects that for us at runtime.

adrinr (approved):
All good, but there are some build errors to look at

PR #12597, opened 2023-12-15
Tell a user if their JS failed due to a timeout.

samwho:
## Description

If a JS execution times out, let the user know instead of using the generic "Error while executing JS".


melohagan (approved):
LGTM! Nice improvement :)

PR #12603, opened 2023-12-18
Optimise getLinkDocuments

samwho:
## Description

Took a look at the Budicloud profiling over the weekend and `getLinkDocuments` was a significant portion of CPU usage.

CleanShot 2023-12-18 at 09 22 27@2x

I suspect it's the `foundIds` scan inside the filter loop, so I've replaced that with a `Set`.


melohagan (approved):

adrinr:
🚀

Issue #12606, opened 2023-12-18
Some column data types are incorrectly loading in automation steps


melohagan:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.13.36

**Describe the bug**
Certain automation columns are loading without a field input, but the labels and binding box are still displayed in the **Create Row** action step.

This applies to:

* Barcode
* BigInt
* User

**Expected behavior**
If these fields are not valid for automations, then they should not be displayed at all.

**Screenshots**

![Screenshot 2023-12-18 at 10 53 19](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/f849ce56-c7f2-42e3-a4a8-aaf4951c2c2f/e4c4bd92-7fa9-4788-8019-116d37cc0d13?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9mODQ5Y2U1Ni1jN2YyLTQyZTMtYTRhOC1hYWY0OTUxYzJjMmYvZTRjNGJkOTItN2ZhOS00Nzg4LTgwMTktMTE2ZDM3Y2MwZDEzIiwiaWF0IjoxNzAyODk5NTgxLCJleHAiOjMzMjczNDU5NTgxfQ.iUyQD\_-UP1e5IHjWPqsY3Uvl-llGSphRcbEytWArxvk)

PR #12610, opened 2023-12-18
Support Barcode, BigInt and User column types in automations

melohagan:
## Description
Noticed that the data types mentioned in the title were not supported in automations. This has been fixed.

Also added a fix for the *Clear relationships if empty?* checkbox. Previously it only considered text inputs, but now that we have the pickers, we need to check if the relationship picker is an empty array

## Addresses
- https://github.com/Budibase/budibase/issues/12606

## Screenshots
Screenshot 2023-12-18 at 14 05 53

Screenshot 2023-12-18 at 14 06 19

### DIFF
--- a/packages/server/src/automations/steps/updateRow.ts
+++ b/packages/server/src/automations/steps/updateRow.ts
@@ -84,9 +84,12 @@ export async function run({ inputs, appId, emitter }: AutomationStepInput) {
 
   // clear any undefined, null or empty string properties so that they aren't updated
   for (let propKey of Object.keys(inputs.row)) {
+    const clearRelationships =
+      inputs.meta?.fields?.[propKey]?.clearRelationships
     if (
-      (inputs.row[propKey] == null || inputs.row[propKey] === "") &&
-      !inputs.meta?.fields?.[propKey]?.clearRelationships
+      ((inputs.row[propKey] == null || inputs.row[propKey] === "") &&
+        !clearRelationships) ||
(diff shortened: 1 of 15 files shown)


packages/server/src/automations/steps/updateRow.ts
```diff
@@ -84,9 +84,12 @@ export async function run({ inputs, appId, emitter }: AutomationStepInput) {
 
   // clear any undefined, null or empty string properties so that they aren't updated
   for (let propKey of Object.keys(inputs.row)) {
+    const clearRelationships =
+      inputs.meta?.fields?.[propKey]?.clearRelationships
     if (
-      (inputs.row[propKey] == null || inputs.row[propKey] === "") &&
-      !inputs.meta?.fields?.[propKey]?.clearRelationships
+      ((inputs.row[propKey] == null || inputs.row[propKey] === "") &&
```
adrinr, line 90: This is quite confusing. Why we have different logic for `undefined` and `false` `clearRelationships`s?

samwho:
I have a feeling this yarn.lock change is actually a result of the dd-trace bump I did earlier today. Sorry about that.

packages/server/src/automations/steps/updateRow.ts
melohagan: I think I just got a bit confused - I've refactored to simplify it

adrinr (approved):

PR #12611, opened 2023-12-18
Implement per-request JS execution limiting.

samwho:
## Description

Allow Budibase installations to impose a per-request limit on user-supplied JS execution time.

For a long time, there has been a per-execution limit of 1 second. So for each formula cell that uses JavaScript, if the JavaScript took longer than 1 second to execute the cell would not evaluate, instead returning an error.

With this PR, there's also more broad tracking of overall JS execution. I'm exposing 2 environment variables to control this:

- `JS_PER_EXECUTION_TIME_LIMIT_MS` defaults to 1000, and is the 1 second limit that's been in place for a long time. This is making that timeout configurable instead of hardcoded.
- `JS_PER_REQUEST_TIME_LIMIT_MS` defaults to undefined, so it's off and not enforcing limits. When set to a value, though, if the total time spent executing user-supplied JS exceeds this value within a single request, all subsequent attempts to run JS will fail immediately with timeout errors.


packages/backend-core/src/timers/timers.ts
```diff
@@ -20,3 +20,39 @@ export function cleanup() {
...
+  }
+
+  constructor(private limitMs: number) {}
+
+  private totalTimeMs = 0
+
+  track<T>(f: () => T): T {
+    this.checkLimit()
+    const [startSeconds, startNanoseconds] = process.hrtime()
+    const startMs = startSeconds * 1000 + startNanoseconds / 1e6
```
adrinr, line 40: You could use [`process.hrtime.bigint()`](https://nodejs.org/docs/latest-v18.x/api/process.html#processhrtimebigint) to get this calculation instead

packages/backend-core/src/utils/Duration.ts
```diff
@@ -49,4 +49,8 @@ export class Duration {
   static fromDays(duration: number) {
     return Duration.from(DurationType.DAYS, duration)
   }
+
+  static fromMiliseconds(duration: number) {
```
adrinr, line 53: Should this be `milli`?

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -2086,4 +2086,112 @@ describe.each([
...
+              },
+            },
+          })
+
+          for (let i = 0; i < 10; i++) {
+            await config.api.row.save(table._id!, { text: "foo" })
+          }
+
+          // Run this test 3 times to make sure that there's no cross-request
+          // polution of the execution time tracking.
```
adrinr, line 2168:
```suggestion
          // pollution of the execution time tracking.
```

packages/server/src/startup.ts
```diff
@@ -152,4 +153,6 @@ export async function startup(app?: Koa, server?: Server) {
       }
     })
   }
+
+  javascript.init()
```
adrinr, line 157: Could we call this jsRunners or something more specific?

packages/string-templates/src/helpers/javascript.js
```diff
@@ -56,10 +56,12 @@ module.exports.processJS = (handlebars, context) => {
     const res = { data: runJS(js, sandboxContext) }
     return `{{${LITERAL_MARKER} js_result-${JSON.stringify(res)}}}`
   } catch (error) {
-    console.log(`JS error: ${typeof error} ${JSON.stringify(error)}`)
     if (error.code === "ERR_SCRIPT_EXECUTION_TIMEOUT") {
       return "Timed out while executing JS"
     }
+    if (error.name === "ExecutionTimeoutError") {
```
adrinr, line 62:
```suggestion
    if (error.name === ExecutionTimeoutError.name) {
```

adrinr: Removing some magic strings

samwho: I think doing this would require `string-templates` to have a dependency on `backend-core`, which I suspect we don't want because this gets used in the frontend as well.

samwho:
@adrinr thanks for the review, I've implemented them all apart from 1. Please take another look 🙏

packages/string-templates/src/helpers/javascript.js
adrinr: Does this mean that we are adding reference `string-templates` that should not be part of it? Would it not be a better practice to bubble up the error instead of trying to cast the types?

samwho: It would almost certainly be better to bubble up the error, yes. I was sticking to how it works today (we already catch all errors and return a string here, I don't know the history behind it).

I would prefer to come back to this at a later date though if that's alright. It's an existing problem that we're not making much worse here, and I have a feeling it'll be a tricky change.

adrinr: That's ok, we should revisit this library if we keep extending it. `string-templates` is usually pretty static

adrinr (approved):

PR #12616, opened 2023-12-19
Set `--enable-source-maps` when running Node.

samwho:
## Description

Docs: https://nodejs.org/docs/latest-v18.x/api/cli.html#--enable-source-maps

This will make stack traces much easier to read, and crucially it will allow DataDog to more accurately report on where problems are occurring in our code.

The docs do point out that this has a performance impact when printing out stack traces, and there are some bugs from the Node 14 days complaining that this was a significant performance impact, but I'd like to try it. If it's really bad, we can turn it back off.

### DIFF
--- a/packages/server/nodemon.json
+++ b/packages/server/nodemon.json
@@ -9,5 +9,5 @@
   ],
   "ext": "js,ts,json",
   "ignore": ["src/**/*.spec.ts", "src/**/*.spec.js", "../*/dist/**/*"],
-  "exec": "yarn build && node ./dist/index.js"
+  "exec": "yarn build && node --enable-source-maps ./dist/index.js"
 }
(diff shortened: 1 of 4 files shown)


packages/server/nodemon.json
```diff
@@ -9,5 +9,5 @@
   ],
   "ext": "js,ts,json",
   "ignore": ["src/**/*.spec.ts", "src/**/*.spec.js", "../*/dist/**/*"],
-  "exec": "yarn build && node ./dist/index.js"
+  "exec": "yarn build && node --enable-source-maps ./dist/index.js"
```
adrinr, line 12: Do we need this here? This will run only on local dev

adrinr (approved):
LGTM, let's monitor any impact JIC

packages/server/nodemon.json
samwho: I put it in because I figure it's also useful in local dev, and the performance impact on local dev is less concerning. Happy to take it out if you don't think it's necessary.

Issue #11773, opened 2023-09-14
[BUDI-7489] PostgreSQL validation is not respected in the Design section


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.10.6

**Describe the bug**
When fetching a Postgres table that has NOT NULL columns, the **Data** section correctly shows these fields are **Required**.

Despite this, in the **Design** section the form (with schema selected) does not reflect this validation.

**To Reproduce**
1. Import a Postgres table with NOT NULL TEXT fields
2. See that the **Data** section shows these fields as required
3. In the **Design** section, add a 'Create' type form and add one of the required fields
4. Click on *Validation* - no rules are being pulled from the schema

**Expected behavior**
If a Postgres column is marked as required in the data section, the form validation should reflect that.

**Screenshots**
![Screenshot 2023-09-14 at 15 16 32](https://github.com/Budibase/budibase/assets/101575380/8694913f-8fe6-4c55-973f-c0b0cacac02e)

![Screenshot 2023-09-14 at 15 18 30](https://github.com/Budibase/budibase/assets/101575380/8f039f0c-34a5-40d3-a548-70e2c7df2b38)

<sub>[BUDI-7489](https://linear.app/budibase/issue/BUDI-7489/postgresql-validation-is-not-respected-in-the-design-section)</sub>

PR #12617, opened 2023-12-19
Support both presence constraint variants

melohagan:
## Description
Required fields were not being mapped from PostgreSQL tables. 

I noticed that in the `FieldConstraints` interface (https://github.com/Budibase/budibase/blob/master/packages/types/src/documents/app/table/schema.ts#L120) the presence constraint could either be an object or a boolean, therefore I have updated the validation code to factor in both variants.

## Addresses
- https://github.com/Budibase/budibase/issues/11773

## Screenshots
Screenshot 2023-12-19 at 10 57 21

Screenshot 2023-12-19 at 10 57 55

Screenshot 2023-12-19 at 10 58 17


shogunpurple (approved):
LGTM!

PR #12618, opened 2023-12-19
Manually inject DataDog trace context to log messages.

samwho:
## Description

Following the guide here: https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/nodejs/#manual-injection.

I'm not sure why, but setting `DD_LOGS_INJECTION=true` isn't injecting a trace ID into our log messages, which means we don't have traces and logs linking up together. Having this would give us more information to work with when looking at tracers, so I'd like to get it working.


shogunpurple (approved):
LGTM

Issue #12451, opened 2023-11-27
[ Core Component - Button ] - Enable buttons to have an Icon


poirazis:
Its an extremely commmon usecase to place icons on buttons.
This would also require the buttonConfiguration property editor to allow selecting an Icon for each button.

ConorWebb96:
Closing as this has been done

PR #12624, opened 2023-12-19
Added icon to button component, reworked icon display code.

ConorWebb96:
## Description
This amends the manifest to add an icon selector. Additionally, small code change to allow the displaying of the selected icons.

## Addresses
- [Button enhancement issue](https://github.com/Budibase/budibase/issues/12451)

## App Export

## Screenshots
Screenshot 2023-12-19 at 11 08 12
Screenshot 2023-12-19 at 11 08 19


melohagan (changes requested):
Pulled it down and it works grand!

I just have a few thoughts:

- Spectrum offer an approach to buttons with icons: https://opensource.adobe.com/spectrum-web-components/components/button/#variants
- There's a comment on this line: https://github.com/Budibase/budibase/blob/c1283860beebabc6e866de38b5c7cd21c17fdb43/packages/client/src/components/app/Button.svelte#L17C1-L17C1 (there may be a reason we don't want to allow icons to be added to buttons, might want to ask AK when he gets back next year)
- This didn't work for buttons within a button group.

aptkingston:
Will take a look at this tomorrow @ConorWebb96, sorry it's taken so long. Only just spotted it!

joebudi:
From the image of the icon/button, there needs to be more of a gap. Small enhancement, but a big crowd pleaser IMO. Well done @ConorWebb96

ConorWebb96:
Hey @joebudi,

Added the gap setting that should hopefully resolve that spacing issue. Opted to use gap to keep the consistency with the other components that use this.

Did some additional fixes whilst i was in there.
- Allowing to search for icons via hitting enter.
- Added the ability to clear the icon as there wasn't a way to do this before.
- Added gap to button and buttonGroup components.

packages/client/src/components/app/Button.svelte
```diff
@@ -58,15 +59,7 @@
...
-      <svg
-        class:hasText={componentText?.length > 0}
-        class="spectrum-Icon spectrum-Icon--size{size.toUpperCase()}"
-        focusable="false"
-        aria-hidden="true"
-        aria-label={icon}
-      >
-        <use xlink:href="#spectrum-icon-18-{icon}" />
-      </svg>
+      <i use:styleable={$component.styles} class="{icon} {size}" />
```
aptkingston, line 62: We should remove the `use:styleable` here - that's specifically for the button itself rather than the icon, so could have unintended side effects if also applied here.

aptkingston (changes requested):
LGTM other than one small change!

I hate that our icon picker is so terrible though, because this makes it a lot more discoverable. Would be really nice to make a new icon picker first (using spectrum icons) and then use there here, but if this feature is needed in the meantime then all good.

Nice work though @ConorWebb96 - code looks good.

packages/client/src/components/app/Button.svelte
ConorWebb96: Removed 👍

melohagan (approved):
LGTM! Nice improvements 😄

aptkingston (approved):
LGTM!

Issue #12260, opened 2023-11-03
[BUDI-7722] Relationship Picker filtering doesn't seem to be working


ConorWebb96:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.11.45
  * App Version: 2.11.45

**Describe the bug**
When using the relationship field on external db you have to do a lot of JS logic to get this working. The encoded id of the relationship fields is no longer present meaning you don't have anything which will match the value of what the relationship picker component  outputs. This means you have to do a lot of unnecessary JS logic to get the value to output in the correct format to use against the primary id of the relationship instead.

**To Reproduce**

1. Go to an app with an external db with a relationship the type doesn't matter
2. Generate a table, nested inside a form
3. Add a relationship picker component inside the form
4. Try and hook this up so it filters the data by the selected relationship
5. See that there is no encoded \_id for the relationship to match up against.
6. Proceed to do

**Expected behavior**
You should be able to hook the relationship filter up out of the box without having to do a lot of to decode, parse to array and optionally parse to number depending on what you want to do.

**Additional context**
[Discord discussion thread link](https://discord.com/channels/733030666647765003/1169353483653226586/1169353483653226586)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-7722](https://linear.app/budibase/issue/BUDI-7722/relationship-filtering-doesnt-seem-to-be-working-like-it-used-to-for)</sub>

melohagan:
The filtering as per this example for an internal DB is also broken: https://docs.budibase.com/docs/cascading-dropdown-filters

nasilrgb:
Still not fixed this issue

melohagan:
Hi @nasilrgb There is an open issue here: https://github.com/Budibase/budibase/issues/13893

PR #12625, opened 2023-12-19
If the filter has changed, re-fetch

melohagan:
## Description
Some recent refactors of the Relationship Field didn't account for dynamic filters, i.e. binding values.

Added a reactive method to check for changes in the filter, and re-fetch as needed.

## Addresses
- https://github.com/Budibase/budibase/issues/12260

## App Export
- https://docs.budibase.com/docs/cascading-dropdown-filters#app-export

## Screenshots
Screenshot 2023-12-19 at 13 38 30

Screenshot 2023-12-19 at 13 38 52


melohagan:
Screenshot 2023-12-19 at 13 51 16

I have just noticed this scenario:

1. Select a relationship value
2. Change the dynamic filter value so that the chosen value would not be in the list
3. See that it displays as a `ro_ta...`

I have updated the code to wipe the selection when the filter changes, as I think it's the only way to guarantee you are selecting a valid option.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -108,8 +108,16 @@
     }
   }
 
+  $: forceFetchRows(filter)
   $: debouncedFetchRows(searchTerm, primaryDisplay, defaultValue)
 
+  const forceFetchRows = async () => {
+    // if the filter has changed, then we need to reset the options, clear the selection, and re-fetch
+    optionsObj = {}
```
adrinr, line 116: This would clean the `optionsObj` cache. This was added to fix initial value fetches, so we should never do this. The filters should be happening on memory as it's currently doing

adrinr (changes requested):

packages/client/src/components/app/forms/RelationshipField.svelte
melohagan: The filter object can change though, such as in this example. If the `optionsObj` isn't cleared then it will add an additional page of filtered options rather than override the existing options with the filtered options

adrinr (approved):
After discussing the issue, LGTM :)

PR #12626, opened 2023-12-19
Revert "Set `--enable-source-maps` when running Node."

samwho:
Reverts Budibase/budibase#12616


shogunpurple (approved):

samwho:
Reverting this because we suspect it's linked to massively increased latency in QA.

Issue #6984, opened 2022-07-29
Autofocus custom css panel once drawer opens


AndyWhann:
Once I open the custom css drawer I want the field to be in focus so I can start typing css straight away.

![image.png](https://images.zenhubusercontent.com/620a72941dd7bc25ac548dc4/24468ad1-9ec2-45aa-962b-f01781e4bcf4)

PR #12628, opened 2023-12-19
Focus the code editor on mount

melohagan:
## Description
When opening drawers such as Custom CSS and Bindings, the code editor will now be focused.

The linked issue only mentioned the custom CSS drawer, however I imagine when presented with a code editor, you would most likely want it to be in focus for immediate typing.

----

Out of interest I had a go at making the Custom CSS drawer saveable on `Ctrl+Enter`. Looks a bit odd, but it is handy. 
Could easily be extended to all binding drawers, but also happy to revert:

binding drawer

## Addresses
- https://github.com/Budibase/budibase/issues/6984


deanhannigan:
Hey @melohagan, this has drifted from what was originally requested. Could you pivot this change back to the auto focus behaviour?

I would also not suggest using `editor.focus()` alone in the core of the editor. Add a reactive way to trigger this at the `CodeEditor` component level. That way, instead of applying the behaviour globally for every single instance of the code editor, it can be applied explicitly in the area that it is needed. In this case, the CSS editor.

I 100% agree that a keyboard shortcut is warranted for the code editor and it is a great enhancement, but it shouldn't be bundled with another ticket. It needs it's own PR and some input from Product.

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -239,6 +251,8 @@
       extensions: buildExtensions(baseExtensions),
       parent: textarea,
     })
+
+    editor.focus()
```
deanhannigan, line 255: Add a way to trigger this so that it can be applied where and when its needed.

deanhannigan (changes requested):

packages/builder/src/components/common/bindings/BindingPanel.svelte
```diff
@@ -199,6 +200,7 @@
                   ]}
                   placeholder=""
                   height="100%"
+                  autofocus={autofocusEditor}
```
deanhannigan, line 203: You would need a similar entry for the JS editor around line 303.

deanhannigan (approved):
LGTM! Just that minor addition for the JS editor focus and its all good

packages/builder/src/components/common/bindings/BindingPanel.svelte
melohagan: Good spot! Added.

deanhannigan (approved):
LGTM!

PR #12629, opened 2023-12-19
Add DataDog tracing to automations.

samwho:
## Description

Adds custom tracing to automation execution. This should give us much more visibility into how automations are affecting the performance of Budicloud.


shogunpurple (approved):
LGTM

PR #12632, opened 2023-12-19
Security/patches

shogunpurple:
## Description
Working through security patches for ironbank - the majority of this stuff is all version updates to patch dependencies, but I've made several updates as well.

### vm2 - isolated VM
```
TL;DR The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to [isolated-vm](https://www.npmjs.com/package/isolated-vm).
```
I've migrated our usages of `vm2` to `isolated-vm`. So we should be covered from the vulns in `vm2` in future. `isolated-vm` is a more robust and safer library. It also provides nice options for memory limits per execution, preventing endless loops etc from blowing up BB.

### QoL
A pet peeve I've always had is in the query UI, how you can't use the classic combo of Meta + Enter to run a query. I've added some key handlers for this, so when editing a query in the data section you can use that combination of keys to run your query.

NB: I'm aware this is a fairly high risk change. We don't have to merge it as I can continue to build the image for ironbank off this branch in the meantime. Just wanted to get a draft up for now

## Feature branch env
[Feature Branch Link](http://fb-security-patches.fb.qa.budibase.net)


shogunpurple:
Update: I've updated the ironbank image to node 20 and debian, so it's now using the latest master. Changed the base branch of this to point to `revert-12934-revert-12930-revert-12929-revert-12769-isolated-vm` since it's going to be merged first

adrinr (approved):
Nice to see that the branch has become smaller :)
There are some changes on the submodules, the equivalent PRs will be required

PR #12633, opened 2023-12-19
Set appId, userId and tenantId as baggage items in DataDog APM.

samwho:
## Description

A "baggage item" on a span in DataDog is a tag that will propagate to all child spans. I think for appId, userId, and tenantId, this is worth doing. 

DataDog warn that overusing baggage items can cause an increase in network load, because they do get transferred remotely with distributed tracing, but I think these three shouldn't be too bad.


shogunpurple (approved):
LGTM

PR #12636, opened 2023-12-19
Cache the helpers list for executing JS.

samwho:
## Description

This function shows up prominently in profiles that include processing formulas.

image

It doesn't appear to change after it has been constructed, so I'm proposing we build it, freeze it, and cache it so it can be reused between invocations of `processJS`.


deanhannigan (approved):
LGTM! Great catch!

PR #12637, opened 2023-12-19
Add traces to track running arbitrary JS.

samwho:
## Description

Can't believe I forgot to do this. We still have requests spending a long time in processing formulas, and I want to make sure it's not some obscure bug in the JS runner execution time tracking.

### DIFF
--- a/packages/server/src/jsRunner.ts
+++ b/packages/server/src/jsRunner.ts
@@ -2,35 +2,44 @@ import vm from "vm"
 import env from "./environment"
 import { setJSRunner } from "@budibase/string-templates"
 import { context, timers } from "@budibase/backend-core"
+import tracer from "dd-trace"
 
 type TrackerFn = <T>(f: () => T) => T
 
 export function init() {
   setJSRunner((js: string, ctx: vm.Context) => {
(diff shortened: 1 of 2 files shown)


adrinr (approved):

packages/server/src/jsRunner.ts
```diff
@@ -2,35 +2,44 @@ import vm from "vm"
...
   setJSRunner((js: string, ctx: vm.Context) => {
-    const perRequestLimit = env.JS_PER_REQUEST_TIME_LIMIT_MS
-    let track: TrackerFn = f => f()
-    if (perRequestLimit) {
-      const bbCtx = context.getCurrentContext()
-      if (bbCtx) {
-        if (!bbCtx.jsExecutionTracker) {
-          bbCtx.jsExecutionTracker =
-            timers.ExecutionTimeTracker.withLimit(perRequestLimit)
+    return tracer.trace("runJS", {}, span => {
```
melohagan, line 11: NAB: I don't have a suggestion, but there's quite a lot of nesting here - is there any way to deal with that? In Spring Boot for example this would probably be an annotation on the method.

melohagan (approved):
LGTM!

Issue #12607, opened 2023-12-18
REST PUT Query throws error - Failed to parse response body.


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.13.45

**Describe the bug**
When performing a PUT query, an error is thrown: `Query Error: Failed to parse response body`.

It appears that Budibase is trying to parse the response, however a 204 response will never have a body to be parsed.

**To Reproduce**
1. Create a valid REST query with method PUT
2. Click `Send`: see the error

**Expected behavior**
I should be able to perform a PUT query without an error.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/101575380/bb4c87eb-9de3-4924-9b16-76a671c9ca3a)

![image](https://github.com/Budibase/budibase/assets/101575380/786fe300-ff3f-4c28-ae17-0485ff253a82)

![image](https://github.com/Budibase/budibase/assets/101575380/55ace991-75ac-42fe-b949-c935d7e7e714)

PR #12643, opened 2023-12-20
Do not throw error on 204 no content

melohagan:
## Description
A valid PUT will sometimes return no JSON body in the case of a 204. Do not try to parse the JSON in this case.

## Addresses
- https://github.com/Budibase/budibase/issues/12607

### DIFF
--- a/packages/server/src/integrations/rest.ts
+++ b/packages/server/src/integrations/rest.ts
@@ -131,7 +131,10 @@ class RestIntegration implements IntegrationBase {
     let data, raw, headers
     const contentType = response.headers.get("content-type") || ""
     try {
-      if (contentType.includes("application/json")) {
+      if (response.status === 204) {
+        data = []
+        raw = []
+      } else if (contentType.includes("application/json")) {
         data = await response.json()
         raw = JSON.stringify(data)
       } else if (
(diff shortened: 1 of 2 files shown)


packages/server/src/integrations/rest.ts
```diff
@@ -131,7 +131,10 @@ class RestIntegration implements IntegrationBase {
     let data, raw, headers
     const contentType = response.headers.get("content-type") || ""
     try {
-      if (contentType.includes("application/json")) {
+      if (response.status === 204) {
+        data = []
```
adrinr, line 135: Why an empty array instead of just undefined?

samwho (approved):
Really good catch!

packages/server/src/integrations/rest.ts
melohagan: I was concerned about null pointers, but I can see that it was already handled:
https://github.com/Budibase/budibase/blob/895c9a9ac36532371a3f97327678099887588613/packages/frontend-core/src/fetch/QueryFetch.js#L78

Think I'll leave it as an empty array for good measure though!

adrinr (approved):

Issue #12645, opened 2023-12-20
Some form fields have a undefined conditional setting within conditions.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.13.48
  - App Version: 2.13.48

**Describe the bug**
Whenever setting up conditions for the multi-select picker field and choosing the update setting. Scrolling through the available options I noticed that there was a undefined option which could be selected. I believe this option to be the `customOptions` select found within the manifest.json.

**To Reproduce**
1. Create an app with sample data
2. Go to the design screen add a form block
3. Eject the form block
4. Scroll down to the multi-select picker that has been generated from the sample data.
5. Click the multi-select picker component
6. Click on the conditions tab
7. Add a condition
8. Select update setting
9. Scroll down and you will eventually see undefined.

**Expected behavior**
Show the actual label of whatever one has been selected, i don't believe that this causes any additional errors.

**Screenshots**
<img width="1511" alt="Screenshot 2023-12-20 at 11 44 30" src="https://github.com/Budibase/budibase/assets/126772285/b3bb72f5-a29c-4bff-bd74-641b56a6fa2c">
  
**App Export**
[form-export-1703072822339.tar.gz](https://github.com/Budibase/budibase/files/13726907/form-export-1703072822339.tar.gz)

**Desktop (please complete the following information):**
 - OS: Mac Os
 - Browser Brave
 - Version v1.61.106

PR #12648, opened 2023-12-20
Exclude custom options from update settings

melohagan:
## Description
Custom options was appearing as `undefined` in the update settings list because no label was provided. 
Initially I tried simply adding a label, however the custom options would not conditionally update (similar to default value), therefore I set `supportsConditions` to **false** instead.

## Addresses
- https://github.com/Budibase/budibase/issues/12645

## Screenshots
Screenshot 2023-12-20 at 14 47 55


deanhannigan (changes requested):
I double checked the `Update setting` functionality of the custom field behaviour and it is actually fully functional, albeit with a number of UX issues. So we don't have to add the `supportsConditions` filtering here. It needs `label: "Custom Options"`

This behaviour allows for conditional building/filtering of the options for the Option/MultiSelect types. You can also conditionally customise the option labels.

#### Conditional Filtering and customisation:
Depending on the specified criteria in my condition, the list *may* filter down to just 2 options and I may want to change the names of the them in the UI. `Agency Worker` might be a better fit where this form is used.

Screenshot 2024-01-02 at 13 02 48

#### UX issue 1: A big offset on the left of the control causing display issues.

Screenshot 2024-01-02 at 12 46 00

#### UX issue 2: A similar issue in the component settings section.
Screenshot 2024-01-02 at 12 54 00

There could also be some enhancements to this drawer to allow browsing of available schema options and a reset option too, but these would be beyond the scope of the current ticket.

For now, see if you can resolve the UX issues and this should be good to go.

melohagan:
Screenshot 2024-01-03 at 13 21 32

UI fix.

Screenshot 2024-01-03 at 13 22 03

deanhannigan (approved):
LGTM!

Issue #9812, opened 2023-02-27
[BUDI-6695] Actions on rich-text long form fields are triggered while you are still editing the field


adowla-nnsc:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 2.3.16

**Describe the bug**
I have an action that's meant to update the row once you've finished editing the field. It behaves as expected for plain text long form fields. The action isn't triggered until I have left the field. On rich text long form fields, the action gets triggered with every few characters typed into the field and doesn't wait till you have left the field.

**To Reproduce**
1. Create a text column that is long form and has rich text enabled
2. Create a long form field tied to that column and choose plain text formatting
3. Create any action for that field--the most straightforward one is:
    a. Choose the "Save Row" action
    b. Choose the Data Source and Table of the field
    c. Select Yes to "Require confirmation"
    d. Use "Add Column" to select the field's column, and update the value with {{ Field Value }}
4. Test by opening the app and typing into the field--you should see a confirmation prompt only when you've left the field / field is no longer focused
5. Edit the long form field and change formatting to rich text
6. Retry step 4--the confirmation prompt will interrupt you as you're in the middle of typing
7. I've attempted those same steps but with step 3 changed to have it be rich text from the very beginning

**Expected behavior**
The action should only trigger once you've finished typing in a rich text field and stopped focusing on it.

**Screenshots**
![msedge_kGeLVZM9Hv](https://user-images.githubusercontent.com/122759217/221475114-2b33d185-4875-45f2-87c3-ec8a0b3c3c2c.gif)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Microsoft Edge

<sub>[BUDI-6695](https://linear.app/budibase/issue/BUDI-6695/actions-on-rich-text-long-form-fields-are-triggered-while-you-are)</sub>

PR #12649, opened 2023-12-20
Use blur for change event RichTextField

melohagan:
## Description
Only trigger the On Change actions on blur of the Rich Text Field. This matches the behaviour of the TextArea 

## Addresses
- https://github.com/Budibase/budibase/issues/9812

## Screenshots

Screenshot 2023-12-20 at 16 07 06

TextArea already triggers OnChange when the field is blurred. The RichText variant has been updated to copy this behaviour:

onblur


shogunpurple (approved):

deanhannigan (approved):
LGTM!

PR #12650, opened 2023-12-20
Allow TEMPLATES_BUCKET_NAME to be configured in the Helm chart.

samwho:
## Description

Looks like this was either missed or the bucket was added and the Helm chart forgotten about.

## Addresses
- https://linear.app/budibase/issue/GROW-87/configuration-of-environment-variable-templates-bucket-name-is-missing


melohagan (approved):
LGTM!

PR #12655, opened 2023-12-21
Remove all custom tracing to see if it's the cause of the memory leak.

samwho:
## Description

This should be a no-op other than we will have fewer custom trace spans.


shogunpurple (approved):
LGTM - worth a try I guess so we can rule it out!

PR #12659, opened 2023-12-22
Add multiple relationships dev script

adrinr:
## Description
Adding a new dev script to allow creating multiple rows with multiple relationships for dev proposes. While doing these changes I found some load testing limits, so I needed to apply the following refactor:
1. Removing "too many requests" limits while running in dev
2. Moving the create row logic from the controller to the SDK, making it more robust and testable
3. Fixing the following ssue with multiple updates in parallel:
   1. A new row with an autoid is created
   2. The row checks the table schema to see what is the latest "autoid" assigned
   3. The row gets autoid+1
   4. The table schema updates the latest "autoid" assigned
   5. If there are multiple creations in parallel, we run into race conditions between these reads and writes, causing db 409s. We added a lock to prevent these issues, making the parallel inserts slower but more robust


packages/backend-core/src/cache/generic.ts
```diff
@@ -18,14 +18,15 @@ export enum TTL {
...
-  return (...args: any) => GENERIC[funcName](...args)
-}
-
-export const keys = performExport("keys")
-export const get = performExport("get")
-export const store = performExport("store")
-export const destroy = performExport("delete")
-export const withCache = performExport("withCache")
-export const bustCache = performExport("bustCache")
+export const keys = (...args: Parameters<typeof GENERIC.keys>) =>
```
adrinr, line 21: This will export the actual parameter types, making the consumers typing stronger

packages/server/src/sdk/app/rows/external.ts
```diff
@@ -1,6 +1,13 @@
-import { IncludeRelationship, Operation } from "@budibase/types"
```
adrinr, line 1: This code has been moved from the controller

mike12345567 (approved):
Very nice - good to have this script and some great typing improvements!

packages/server/scripts/load/create-many-relationships.js
```diff
@@ -0,0 +1,192 @@
...
+      })
+      .finally(() => {
+        delete inFlight[itemIndex]
+      })
+
+    inFlight[itemIndex] = promise
+
+    if (Object.keys(inFlight).length >= maxConcurrency) {
+      await Promise.race(Object.values(inFlight))
+    }
```
samwho, line 59: This is a really neat way of limiting concurrency!

packages/server/scripts/load/create-many-relationships.js
```diff
@@ -0,0 +1,192 @@
...
+  return table
+}
+
+async function run() {
+  if (!appId) {
+    const app = appId ? await getApp(apiKey, appId) : await createApp(apiKey)
+    appId = app._id
+  }
+
+  console.log(`App created: http://localhost:10000/builder/app/${appId}`)
```
samwho, line 95: Would be confusing to say "App created" when an app ID gets passed in. Probably worth having a different log message in each scenario.

samwho (approved):
Just one small comment, no need to block on it.

PR #12675, opened 2023-12-29
Fix flaky test

adrinr:
## Description
Fixing flaky scim test. There is a test that it creates 25 scim groups randomly at the setup, and from time to time we are getting 409 as some of these names might be repeated, given that we are using random generated names.
Changing the setup to ensure uniqueness

These failing pipelines will be fixed:

- https://github.com/Budibase/budibase/actions/runs/7345702459/job/19999268757
- https://github.com/Budibase/budibase/actions/runs/7336376609/job/19975601067


samwho (approved):
You are my hero. Also, TIL `generator.unique`.

PR #12676, opened 2023-12-29
Add migration to support SQS

adrinr:
## Description
Adding a new app migration to support the future SQS implementation. This will require a deterministic field to support joins, so we are adding a tableId field to internal links in order to achieve this


packages/server/src/db/linkedRows/LinkDocument.ts
```diff
@@ -17,6 +17,7 @@ import { LinkDocument } from "@budibase/types"
 class LinkDocumentImpl implements LinkDocument {
   _id: string
   type: string
+  tableId: string
```
samwho, line 20: I'm a little uncertain about this name, because it's not actually a table ID. Could be confusing when reading the code to later find out that it's 2 conjoined table IDs.

Perhaps `lookupId`?

packages/server/src/db/linkedRows/LinkDocument.ts
```diff
@@ -54,7 +55,11 @@ class LinkDocumentImpl implements LinkDocument {
       fieldName: fieldName2,
       rowId: rowId2,
     }
+    this.tableId = [this.doc1.tableId, this.doc2.tableId].sort().join(SEPARATOR)
   }
+  _rev?: string | undefined
+  createdAt?: string | number | undefined
+  updatedAt?: string | undefined
```
samwho, line 62: Are these necessary? `LinkDocument` inherits from `Document`, which defines these already.

PR #12684, opened 2024-01-02
Set password limits

adrinr:
## Description
1. Do not accept login attempts with long passwords, preventing possible DDoS attacks.
2. Add password limits on:
- Reset password
- Create the admin user

These restrictions have been added in the controllers/sdk instead of using joi validation in order to be able to provide better user feedback, with custom messages.

Related: https://github.com/Budibase/account-portal/pull/493

## Addresses
- [BUDI-7793 - Vulnerability Report 03: No Password Length Restriction leads to Denial of Service](https://linear.app/budibase/issue/BUDI-7793/vulnerability-report-03-no-password-length-restriction-leads-to-denial)

## Screenshots
image


packages/backend-core/src/security/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+export const PASSWORD_MIN_LENGTH = 8
+export const PASSWORD_MAX_LENGTH = 512
+
+export function validatePassword(
+  password: string
+): { valid: true } | { valid: false; error: string } {
+  if (env.isDev() && !env.isTest() && password) {
+    // We accept any password while on development
+    return { valid: true }
+  }
```
samwho, line 12: This absolutely terrifies me. If this ever makes it to production by mistake it would be catastrophic.

packages/backend-core/src/security/auth.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!password || password.length < PASSWORD_MIN_LENGTH) {
+    return {
+      valid: false,
+      error: "Password invalid. Minimum eight characters.",
```
samwho, line 17: Can we use `PASSWORD_MIN_LENGTH` here, similar to how we use `PASSWORD_MAX_LENGTH` below?

packages/backend-core/src/security/auth.ts
adrinr: Same, I just cleaned it :)

mike12345567 (approved):

samwho (approved):

PR #12686, opened 2024-01-02
Remove CLI PKG build

mike12345567:
## Description
Removing CLI pkg build - this has been a problematic element of our build for a while now and it is rarely utilised - rather than continuing to maintain this, getting rid of it.

Simply - I was running into some problems with builds - which may have been related to the issues that @PClmnt had been experiencing over the Christmas period - I decided to quickly check how much the CLI builds are actually used.

Utilises the Github API: https://api.github.com/repos/Budibase/budibase/releases to determine number of downloads across the last 30 releases (1 month).

Adding up the totality of downloads since the 29th of November, the Github downloads of the asset files are as follows:
```
cli-linux: 162
cli-macos: 43
cli-win.exe: 215
openapi.json: 16
openapi.yaml: 20
```

It should also be noted that across these downloads, quite a few versions I tried were not functional, due to various issues, as well as the Mac version simply not working on most M1/2 platforms due to arch issues.

By comparison - looking at statistics from our NPM build release of the CLI, we can see that it is into 20,000+ downloads in the same period (https://www.npmjs.com/package/@budibase/cli) - with the PKG versions being so under-utilised it feels best to focus on the NPM versions, which are significantly slowed by the pkg functionality that had to be introduced, removing all of this speeds up our builds, removes an area which is very susceptible to errors and should make sure that all users are getting the best experience of the CLI possible.

I have already removed from the documentation the links to the CLI builds on Github, as well as creating a PR in deploys (https://github.com/Budibase/budibase-deploys/pull/228) to remove the assets from the release.


shogunpurple (approved):
Happy with this - I think it introduces a lot more pain than it solves. Node should really be a requirement for using the CLI and keeping it up to date

packages/cli/src/index.ts
```diff
@@ -21,6 +20,23 @@ async function init() {
...
+    if (evt) {
+      console.error(
+        error(
+          "Failed to run CLI command - please report with the following message:"
+        )
+      )
+      console.error(error(evt))
+    }
+  })
+})
```
samwho, line 38: wot dis

mike12345567: Moving something that was done as part of the prebuild code to handle errors up to the top level, as the prebuilds step has been removed.

PR #12687, opened 2024-01-02
Fix for Relationship picker filtering in the Formblock

deanhannigan:
The `RelationshipFilterEditor` component currently attempts to resolve it's datasource from the nearest form in the component tree. In a Formblock, that resolution will fail as its Form is not rendered in the tree.

The picker now will attempt to resolve its datasource from the selected component. In this instance, it will expect a Formblock. With that datasource it is then able to determine the schema for the linked table and populate the fields in the filters list

**Addresses:**
https://linear.app/budibase/issue/BUDI-7864/relationship-filtering-in-the-formblock-does-not-list-the-linked-table


packages/builder/src/components/design/settings/controls/RelationshipFilterEditor.svelte
```diff
@@ -20,8 +23,23 @@
     component => component._component.endsWith("/form")
   )
 
+  const resolveDatasource = (currentAsset, componentInstance, form) => {
+    if (!form && componentInstance._id != $store.selectedComponentId) {
```
Ghrehh, line 27: NAB: Can we use `!==` here?

Ghrehh (approved):
LGTM

PR #12698, opened 2024-01-03
Fix local development of multi-tenancy apps

mike12345567:
## Description
Fix for multi-tenancy issue in local development, couldn't load apps as it requires a tenant ID in the subdomain when in multi-tenancy mode which isn't possible in development - this makes sure that local development can work by using the users cookie instead.


samwho (approved):
Looks good to me, but some tests have been broken by the change.

PR #12699, opened 2024-01-03
Triggering another automation from within an automation

PClmnt:
## Description
Quite a requested feature for a while has been the ability to trigger an automation from within a running automation. This PR implements that.

- It is synchronous, when the other automation is triggered the automation will wait for it to complete before continuning.
- Only available on business tier. 
- Will timeout after 12000ms
- Output of the step is all of the steps of the triggered automation. So any of them can be accessed later in the original automation. 
- The other automation has to be an app action automation
- Can pass in fields that are defined in the trigger of the automation that is being called

- ## Addresses
#BUDI-5799

## Screenshots
image

## Feature branch env
[Feature Branch Link](http://fb-feat-automation-features.fb.qa.budibase.net)


packages/builder/src/builderStore/store/automation/index.js
```diff
@@ -155,6 +155,7 @@ const automationActions = store => ({
     await store.actions.save(newAutomation)
   },
   test: async (automation, testData) => {
+    console.log(testData)
```
mike12345567, line 158: console log needs removed.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ActionModal.svelte
```diff
@@ -19,10 +19,12 @@
   export let lastStep
 
   let syncAutomationsEnabled = $licensing.syncAutomationsEnabled
+  let triggerAutomationsEnabled = true
```
mike12345567, line 22: what does the boolean do - I don't see anywhere that changes it?

packages/server/src/automations/actions.ts
```diff
@@ -102,6 +105,7 @@ export async function getActionDefinitions() {
 
 /* istanbul ignore next */
 export async function getAction(stepId: string) {
+  console.log(stepId)
```
mike12345567, line 108: console log needs removed.

packages/server/src/automations/steps/trigger.ts
```diff
@@ -0,0 +1,77 @@
+import {
```
mike12345567, line 1: it would be good to have some testing around this if possible.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/ActionModal.svelte
PClmnt: I had it set to true for testing, fixed.

packages/server/src/tests/utilities/structures.ts
```diff
@@ -21,7 +21,9 @@ import {
   Table,
   INTERNAL_TABLE_SOURCE_ID,
   TableSourceType,
+  AutomationIOType,
 } from "@budibase/types"
+import { bool } from "joi"
```
mike12345567, line 26: I think this may have been an accidental import!

packages/server/src/automations/steps/trigger.ts
mike12345567, line 1: Not to be nit-picky, but the naming of this is a little confusing, the fact we have a `Trigger` step to an automation, and a step type of `trigger` - I fear this could lead to some confusion.

Would it be possible to rename this to something like `TRIGGER_AUTOMATION`, or perhaps something very different but could be `CASCADE` or `CHAIN` - I'm just worried that the reuse of the word trigger for two types of actions will lead to confusion.

packages/server/src/automations/steps/trigger.ts
```diff
@@ -0,0 +1,88 @@
...
+  AutomationFeature,
+  AutomationResults,
+  Automation,
+  AutomationCustomIOType,
+} from "@budibase/types"
+import * as triggers from "../triggers"
+import { db as dbCore, context } from "@budibase/backend-core"
+
+export const definition: AutomationStepSchema = {
+  name: "Trigger Automation",
```
mike12345567, line 16: `Trigger an automation` might read better

packages/server/src/automations/actions.ts
```diff
@@ -15,6 +15,7 @@ import * as delay from "./steps/delay"
 import * as queryRow from "./steps/queryRows"
 import * as loop from "./steps/loop"
 import * as collect from "./steps/collect"
+import * as trigger from "./steps/trigger"
```
mike12345567, line 18: As I said with my comment on the steps file, this is an example where its a little confusing, the first time I saw this I was very confused why the triggers were being pulled into the action list, but I get whats happening now!

packages/types/src/documents/app/automation.ts
```diff
@@ -61,6 +63,7 @@ export enum AutomationActionStepId {
   LOOP = "LOOP",
   COLLECT = "COLLECT",
   OPENAI = "OPENAI",
+  TRIGGER = "TRIGGER",
```
mike12345567, line 66: Does this need to be updated? Should this be `TRIGGER_AUTOMATION_RUN` or something like that?

mike12345567 (approved):
LGTM - just one small thing, the `AutomationActionStepId` in the types still having a `TRIGGER` enum value should probably be updated!

PR #12701, opened 2024-01-04
Allow ndots to be configurable in the Helm chart.

samwho:
## Description

Allows the DNS `ndots` configuration to be changed per-service in our Helm chart.

### DIFF
--- a/charts/budibase/templates/automation-worker-service-deployment.yaml
+++ b/charts/budibase/templates/automation-worker-service-deployment.yaml
@@ -227,6 +227,7 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{ end }}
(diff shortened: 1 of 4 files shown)


charts/budibase/templates/automation-worker-service-deployment.yaml
```diff
@@ -251,6 +252,11 @@ spec:
       {{ end }}
       restartPolicy: Always
       serviceAccountName: ""
-      {{ if .Values.services.automationWorkers.command }}}
```
samwho, line 254: This was in the complete wrong place. I think we got quite lucky that it didn't break anything.

shogunpurple (approved):
LGTM

PR #12705, opened 2024-01-04
Fix running worker directly from ts

adrinr:
## Description
Configuring ts paths when running the worker via ts-node. Not having this means that it will be picking up the dist folders, forcing us to build them all and to keep them up to date. Using the paths will give us a better dev experience.

Current error:
image


mike12345567 (approved):
Thanks for this!

PR #12706, opened 2024-01-04
[Fix] Keep exisitng views when refetching tables

adrinr:
## Description
Fixing a bug that, on external datasource refetch, the existing views are dropped.

## Addresses
- [BUDI-7571 - Refetching tables deletes pre-existing views](https://linear.app/budibase/issue/BUDI-7571/refetching-tables-deletes-pre-existing-views)

## Screenshots
https://github.com/Budibase/budibase/assets/15987277/37d27969-d53d-490b-890f-4a38348ebce3

### DIFF
--- a/packages/server/src/integrations/utils.ts
+++ b/packages/server/src/integrations/utils.ts
@@ -305,8 +305,8 @@ export function shouldCopySpecialColumn(
 }
 
 /**
- * Looks for columns which need to be copied over into the new table definitions, like relationships
- * and options types.
+ * Looks for columns which need to be copied over into the new table definitions, like relationships,
+ * options types and views.
  * @param tableName The name of the table which is being checked.
  * @param table The specific table which is being checked.
  * @param entities All the tables that existed before - the old table definitions.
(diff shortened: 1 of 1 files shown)


packages/server/src/integrations/utils.ts
```diff
@@ -325,6 +325,9 @@ function copyExistingPropsOver(
     if (entities[tableName]?.created) {
       table.created = entities[tableName]?.created
     }
+
+    table.views = entities[tableName].views
```
adrinr, line 329: This is the only actual required change

samwho (commented):
Is it possible to write an automated test for this, so it doesn't regress? 🙏

adrinr:
I am working on a refactor for this. These integration utils are currently used in each integration individually (pg, MySQL, etc). This should not be the case, as these fields (views, primary column, etc) are part of the datasource logic, so it should live in our datasource and it should be applied to any integration.
I am creating this refactor independently, and I will add all the test suit for that

samwho (approved):

Issue #11418, opened 2023-08-01
[BUDI-7353] The 'Is in' filter is not working in automations for number type columns when using app action trigger bindings


melohagan:
**8Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.8.30
  * App Version: 2.8.30

**Describe the bug**
The **Is in** filter works in automations when using an app action trigger field binding whenever the value being filtered is of type Text, but not of type Number.

I did check the filtering for the Table block in the Design section, and it works as expected, so it does appear to be specific to automations (and number fields) in this case.

I tried a variety of setting the app action field to Text, or Array type, as well as returning arrays from a JavaScript binding instead of handlebars, but it didn't work for any combination of these.

**To Reproduce**

1. Add an automation with an **App Action** trigger. Add a field 'Filter' of type *Text* (or *Array* it doesn't matter)
2. Add a **Query Rows** step for the Expenses sample table, and `Define filters`:
3. Notes *Is in* Binding `{{ trigger.fields.Filter }}` → run the automation with a known note, e.g. 'Work van', and see that it returns correctly
4. Change the filter to Cost *Is in* Binding `{{ trigger.fields.Filter }}` → run the automation now for a known cost, e.g. 45, and see that no rows are returned

**Expected behavior**
The **Is in** filter should work for *number* type fields within automations.

Should double check other non text types, such as Dates.

**Screenshots**

[...]

PR #12708, opened 2024-01-04
Support number array binding filtering in automations

melohagan:
## Description
Check for handlebars was not occurring for `oneOf` filtering.

## Addresses
- https://github.com/Budibase/budibase/issues/11418

## Screenshots
Screenshot 2024-01-04 at 16 00 33

Screenshot 2024-01-04 at 16 00 55

Screenshot 2024-01-04 at 16 01 26


PClmnt (approved):

samwho (commented):
Sorry to be that guy again: could we get a test for this? 😇

packages/server/src/automations/steps/queryRows.ts
```diff
@@ -117,7 +117,13 @@ function typeCoercion(filters: SearchFilters, table: Table) {
           continue
         }
         if (column.type === FieldTypes.NUMBER) {
-          searchParam[property] = parseFloat(value)
+          if (key === "oneOf") {
+            value = value.split(",").map(item => parseFloat(item))
+          } else {
+            value = parseFloat(value)
+          }
+          // @ts-ignore
```
mike12345567, line 125: why is the `@ts-ignore` needed? generally we prefer any typing to removing all type security (if no typing is possible at all). 

Can probably remove all other `@ts-ignore` from this file - the types should be fully solvable now.

packages/shared-core/src/filters.ts
```diff
@@ -174,10 +174,10 @@ export const buildLuceneQuery = (filter: SearchFilter[]) => {
           return
         }
       }
-      if (type === "number" && typeof value === "string") {
+      if (type === "number" && typeof value === "string" && !isHbs) {
```
mike12345567, line 177: this feels like it could do with a test - I don't really track what this logic is trying to achieve and why the `!isHbs` needed to be removed - its likely someone else could be confused by this - a quick unit test could (we already have a filters test set I think) could help avoid a regression here.

melohagan:
Added another fix for when using the **Array** type in an app action trigger. If there are no options in the constraints (which there won't be in this case) then show a text input instead:

action trigger array

add test data

packages/shared-core/src/filters.ts
melohagan: Unit tests added

packages/server/src/automations/steps/queryRows.ts
melohagan: ts-ignore removed!

samwho (approved):
Superb work on this, Mel.

mike12345567:
LGTM :)

PR #12710, opened 2024-01-05
Fix Grid component undefined primary array

deanhannigan:
## Description

Minor issue when the value `undefined` is being concatenated into the columns list. When processing the list entries for the draggable columns list, errors get thrown in the console for the erroneous entry. Allowing the `primary` array to default to `[]` solved the issue.


Ghrehh (approved):
LGTM

PR #12712, opened 2024-01-05
Get rid of docker compose warning for OFFLINE_MODE not set

mike12345567:
## Description
Get rid of a warning when starting up self host stack with docker compose V2 - OFFLINE_MODE is fine to be left as a blank string and warning might be causing some confusion.


melohagan (approved):
LGTM!

PR #12713, opened 2024-01-05
Remove password validation for admin user environment variables

mike12345567:
## Description
Adding an option to disable password validation when creating an admin user - this means that the environment variables used for BB_ADMIN creation can have any length of password (not breaking change).

There was a breaking change that affected plugin development environments and stopped images starting when the `BB_ADMIN...` environment variables were set due to the new password validation rules.

### DIFF
--- a/packages/backend-core/src/users/db.ts
+++ b/packages/backend-core/src/users/db.ts
@@ -48,7 +48,7 @@ type CreateAdminUserOpts = {
   ssoId?: string
   hashPassword?: boolean
   requirePassword?: boolean
-  noPasswordValidation?: boolean
+  skipPasswordValidation?: boolean
 }
 type FeatureFns = { isSSOEnforced: FeatureFn; isAppBuildersEnabled: FeatureFn }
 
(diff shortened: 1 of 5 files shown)


packages/backend-core/src/users/db.ts
```diff
@@ -44,6 +44,12 @@ type GroupFns = {
   getBulk: GroupGetFn
   getGroupBuilderAppIds: GroupBuildersFn
 }
+type CreateAdminUserOpts = {
+  ssoId?: string
+  hashPassword?: boolean
+  requirePassword?: boolean
+  noPasswordValidation?: boolean
```
samwho, line 51: nit: it would be a lot nicer if this was a truthy name, e.g. `validatePassword`. The usage later on, `if (!opts.noPasswordValidation)` makes me pause for a sec when looking at it.

samwho (approved):
One small comment that I'm not doing to dig my heels in over.

packages/backend-core/src/users/db.ts
mike12345567: I had a look at this - I fear that might require a lot more validation, as any call that sets options would then need to set the `validatePassword` - or we would need to do an `undefined` check to default it to true - setting it disable the password validation is a bit more confusing but feels a bit safer as the default is always validate.

adrinr (approved):

packages/backend-core/src/users/db.ts
adrinr: That could be called `skipPasswordValidation` and it sounds less negative :)

PR #12716, opened 2024-01-05
Enable dd-trace logging in QA to try and figure out why user info isn't being added to traces.

samwho:
## Description

Not seeing user info on traces, not sure why. Hoping that enabling the logs in QA will help.

### DIFF
--- a/packages/server/src/ddApm.ts
+++ b/packages/server/src/ddApm.ts
@@ -3,5 +3,9 @@ import apm from "dd-trace"
 // enable APM if configured
 if (process.env.DD_APM_ENABLED) {
   console.log("Starting dd-trace")
-  apm.init()
+  apm.init({
+    // @ts-ignore for some reason dd-trace types don't include this options,
+    // even though it's spoken about in the docs.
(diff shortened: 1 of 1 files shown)


packages/server/src/ddApm.ts
```diff
@@ -3,5 +3,9 @@ import apm from "dd-trace"
 // enable APM if configured
 if (process.env.DD_APM_ENABLED) {
   console.log("Starting dd-trace")
-  apm.init()
+  apm.init({
+    // @ts-ignore for some reason dd-trace types don't include this options,
+    // even though it's spoken about in the docs.
+    debug: process.env.DD_ENV === "qa",
```
adrinr, line 9: Should we not use the node env settings?

samwho: I was in two minds about this. I don't mind either way, personally, but decided to go with the `DD_ENV` here because it's datadog-specific code.

mike12345567: I'm happy enough for it to work this way - we can change it if its a problem.

mike12345567 (approved):

PR #12721, opened 2024-01-05
Remove objects from setUser call. We can add them later if we need them.

samwho:
## Description

The user data on spans looks like this:

CleanShot 2024-01-05 at 18 14 33

I'm just removing the ones that are objects who have had `.toString()` called on them for now. If we find that information would be useful to have in traces we can add it back in the correct way.


samwho:
@adrinr at this moment all I really care about is the user ID and the tenant ID. I was throwing the other stuff in because it could be useful at some point. If there's a use-case you have for them, I can change this PR to keep them.

adrinr (approved):

PR #12724, opened 2024-01-08
Trigger binding suggestions from  trigger fields. 

PClmnt:
## Description
Adds binding suggestions for field trigger bindings.

Previously the user had no visibility in their bindings drawer as to what was in their app actions or rows, nor did they have access to the type of binding. They simply had to select `fields` and then add the bindings themselves like `{{tigger.fields.action1}}`

Now they have access to those bindings in the bindings drawer with their associated types. 

I have also refactored the automation binding generation code in to smaller functions, it was getting to be a pretty large and unwieldy function.
## Addresses
- #BUDI-7557

## Screenshots
image

image

image

## Feature branch env
[Feature Branch Link](http://fb-feat-trigger-binding-suggesti.fb.qa.budibase.net)


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -180,54 +253,45 @@
...
-      } else if (isLoopBlock) {
-        bindingIcon = "Reuse"
-        bindingRank = idx + 1
-      } else {
-        bindingIcon = allSteps[idx].icon
-        bindingRank = idx - loopBlockCount
+
+      if (idx === 0 && automation.trigger?.event === "app:trigger") {
+        schema = Object.fromEntries(
+          Object.keys(automation.trigger.inputs.fields).map(key => [
```
deanhannigan, line 259: One minor change here would be to change this to `automation.trigger.inputs.fields || []`. I ran this through old automations and it seems that `fields` wasn't being defaulted to an empty array on create and the builder crashes. New automations will add in the empty array.

deanhannigan (approved):
LGTM! One minor suggestion but otherwise this is an awesome addition!

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
PClmnt: Sorted! Great catch, thanks Dean!

PR #12738, opened 2024-01-09
Updated lock tags to correct plan type

ConorWebb96:
## Description
Updated lock tags for features which were business and have been moved to other plans after the deprecation of the business tier.


jvcalderon (approved):
LGTM

PR #12755, opened 2024-01-10
Disable eject option for blocks in the component tree when ejectable set to false in manifest

deanhannigan:
## Description

- General fix to remove the `Eject Block` option from the context menu if `ejectable` is set to `false` in the component definition. 
- Added type check to the `ctrl-e` keyboard shortcut to only show the eject modal for ejectable block components. Previously there was no type check and the modal would open for `all types`.

## Addresses
[Original Multistep Ticket](https://github.com/Budibase/budibase/pull/12585)


Ghrehh (approved):
LGTM, good catch.

You can also still use `ctrl-e` to attempt to eject stuff you shouldn't be able to, so we should probably fix that at some point too.

deanhannigan:
> LGTM, good catch.
> 
> You can also still use `ctrl-e` to attempt to eject stuff you shouldn't be able to, so we should probably fix that at some point too.

I'll add a fix for that into this branch too, cheers!

Ghrehh (approved):

melohagan (approved):

PR #12765, opened 2024-01-11
Remove constant CouchDB HEAD requests

mike12345567:
## Description
This removes the need for constant DB HEAD requests to check if a database exists or not. Instead, it tries to make the request, and if it fails it will check if the reason for failure is the database not existing. If it doesn't exist it runs through the same old flow to confirm that it definitely doesn't exist, and if it doesn't then it will create it.

This doesn't strictly need its own test cases, as essentially every single test case runs through a variety of creating databases, removing them and recreating them - a unit test around this would not provide much value.

As confirmation I've tried a few combinations of nuking my Couch installation locally and rebuilding it from scratch, I haven't run into any problems with this.

I have also kept around the `HEAD` request before database creation, just incase it might exist in a different node, this at least leaves two chances for the DB to be replicated/come into existence before we attempt to create it - in future we can try to update this to make sure that DB creation only occurs in the appropriate scenarios.

### DIFF
--- a/packages/backend-core/src/db/couch/DatabaseImpl.ts
+++ b/packages/backend-core/src/db/couch/DatabaseImpl.ts
@@ -19,6 +19,8 @@ import { WriteStream, ReadStream } from "fs"
 import { newid } from "../../docIds/newid"
 import { DDInstrumentedDatabase } from "../instrumentation"
 
+const DATABASE_NOT_FOUND = "Database does not exist."
+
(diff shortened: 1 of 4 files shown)


adrinr (approved):
Will this improve latency and db loads?

mike12345567:
> Will this improve latency and db loads?

This is the hope - today we send a fair amount of HEAD requests to Chesterfield, which sometimes can make up 100ms+ of the latency of the request, this simply gets rid of them!

PR #12769, opened 2024-01-12
Migrate from `vm` to `isolated-vm`

samwho:
## Description
Replacing vm2 usages in favour of isolated VM. There are multiple reasons for this, one of them being a suggestion from the vm2 developers: `TL;DR The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to [isolated-vm](https://www.npmjs.com/package/isolated-vm).`

Moving to `isolated-vm` will add extra benefits, such as:
- Isolation, removing the vulnerability of allowing users to run malicious code
- Memory limits, preventing some DDoS attacks (both voluntary or involuntary ones)

The main incompatibility between our implementation with vm2 vs the isolated-vm one is the way we inject our handlebars helpers. We could not inject the full library as as objects, so we require following the process below:
1. Use esbuild to generate a file with the required helper functions, without any external dependency
2. Create an isolated-vm module with the generated code
3. Register this module in an isolated-vm script
4. Run the script with the user's code

These changes also improve the performance quite a lot when running heavy calculations. Example, running search requests for a table with 2500 rows with 250 formula columns each. The benchmarks are fetching only 1 row. Running in feature-branches.
### Before
image

### After
image

## Addresses
- [BUDI-7884 - Isolated-vm integration
](https://linear.app/budibase/issue/BUDI-7884/isolated-vm-integration)

## Feature branch env
[Feature Branch Link](http://fb-isolated-vm.fb.qa.budibase.net)


packages/server/src/jsRunner.ts
```diff
@@ -1,61 +1,145 @@
...
-            bbCtx.jsExecutionTracker = tracer.trace(
-              "runJS.createExecutionTimeTracker",
-              {},
-              span => timers.ExecutionTimeTracker.withLimit(perRequestLimit)
+      try {
+        const bbCtx = context.getCurrentContext()!
+
+        const isolateRefs = bbCtx.isolateRefs
+        if (!isolateRefs) {
+          const jsIsolate = new ivm.Isolate({ memoryLimit: 64 })
```
samwho, line 30: We should make the `memoryLimit` configurable with an environment variable. I'm not sure what the right value is just yet, and self-hosters may want to significantly increase this.

packages/server/src/jsRunner.ts
```diff
@@ -1,61 +1,145 @@
...
+          const cpuMs = Number(jsIsolate.cpuTime) / 1e6
+          if (cpuMs > perRequestLimit) {
+            throw new ExecutionTimeoutError(
+              `CPU time limit exceeded (${cpuMs}ms > ${perRequestLimit}ms)`
+            )
+          }
+        }
+
+        // We need to warp up the actual run in a `cb` function to be able to extract its value from isolated-vm
+        js = js.replace(/run\(\);$/, "cb(run());")
```
samwho, line 109: This makes me a little bit nervous. It would be possible to break JS handling in the server by making a seemingly-unrelated change in the `string-templates` package. Might be worth adding a comment over there saying that the name `run` is important for the backend.

packages/server/src/jsRunner.ts
```diff
@@ -1,61 +1,145 @@
-import vm from "vm"
+import ivm from "isolated-vm"
```
samwho, line 1: When me you and @mike12345567 chatted about this, I remember Mike having the idea to have a test suite in `package/server` that calls this `jsRunner.ts` code with JS snippets containing calls to every helper. Is that something we could do now we have easy access to the JS helper list from `string-templates`?

packages/server/src/jsRunner.ts
samwho: Thinking a bit more about this, it likely makes the most sense for this to be off by default (no limit), so as to preserve existing behaviour for self-hosters. Then we can tinker with it in cloud to find the best balance of protection for other workloads and not annoying customers.

packages/server/src/jsRunner.ts
mike12345567: We could just tie these two together using a constant somewhere like `const JS_RUNNER_FUNCTION = "run"` then use that in string-templates + here, if someone ever changed it they'd be aware of its impact here as well and ideally it wouldn't be as impacted.

packages/server/Dockerfile
```diff
@@ -82,6 +82,8 @@ EXPOSE 4001
 # due to this causing yarn to stop installing dev dependencies
 # which are actually needed to get this environment up and running
 ENV NODE_ENV=production
+# this is required for isolated-vm to work on Node 20+
+ENV NODE_OPTIONS="--no-node-snapshot"
```
mike12345567, line 86: Its probably worth applying these same changes to the `worker` as well - JS isn't really used in it, but it could be used in regards to email templates etc - theres nothing stopping it from happening so should support it.

packages/string-templates/src/index.cjs
```diff
@@ -36,3 +36,8 @@ if (!process.env.NO_JS) {
     return vm.run(js)
   })
 }
+
+const errors = require("./errors")
```
mike12345567, line 40: Do we still need the `index.cjs`, `index.mjs` and `index.js`? Right now all usage of the library comes through the `index.mjs` - its probably worth clearing this down to just a single endpoint - although likely worth doing in a separate PR.

packages/server/src/jsRunner.ts
adrinr: I thought the same, and this comment made me rethink it. I used a selfinvoking function in string-templates, that should not break anything and make this usage much easier :)

packages/string-templates/src/index.cjs
adrinr: Not sure what is using these indexes. Probably we can clean it, but as you said I will research it in a separated PR :)

packages/server/src/jsRunner.ts
adrinr: Done in here https://github.com/Budibase/budibase/pull/12922

mike12345567 (approved):

Issue #8389, opened 2022-10-25
[BUDI-6818] Array type missing from query schema selector


andz-bb:
**Hosting**

- Self
  - Method: dev stack with MySQL db
  - Budibase Version: 2.0.30-alpha.16
  - App Version: 2.0.30-alpha.16

**Describe the bug**
When running a query on an external datasource there is no proper support for array type columns. The server returns the type as "array" but the type select component doesn't have that as a possible value so it shows as null in the front end.

**To Reproduce**
1. Create a Multi-select column or another array type in an external datasource
2. Create a query in the datasource that pulls that array column back
3. Check the query schema - the array column type will be null

**Expected behavior**
Include an option for Array in the query schema column type options

**Screenshots**
![SCR-20221025-fpl](https://user-images.githubusercontent.com/110921612/197748669-a575a5d9-64ab-4f57-86ec-beb0999cc16c.png)

In my case I'm running MySQL so this is a JSON type column:
<img width="227" alt="image" src="https://user-images.githubusercontent.com/110921612/197750076-2204d6a4-b05e-47fe-b398-99a9c916d410.png">

when previewing the query, the server correctly returns the column type as "array":
![SCR-20221025-fof](https://user-images.githubusercontent.com/110921612/197750504-87ba2aab-edaf-4807-a591-98a7024a10e6.png)

but that type from the response isn't properly supported in the front end
![SCR-20221025-foz](https://user-images.githubusercontent.com/110921612/197750187-f014fcca-6340-4768-8197-87f0bd8fa01d.png)

we need a value for array here:
![SCR-20221025-fp5](https://user-images.githubusercontent.com/110921612/197750334-5da9bac4-03bb-45f2-8178-96cf5db7f4ee.png)

**Desktop (please complete the following information):**
 - OS: macOS 12.6
 - Browser: Chrome
 - Version: 106

**Additional context**
Raised from discussion #8304 

<sub>[BUDI-6818](https://linear.app/budibase/issue/BUDI-6818/array-type-missing-from-query-schema-selector)</sub>

aptkingston:
I think the bug here is that we're trying to handle that column type at all - I don't think we support JSON/array types in external DBs, so we should just be ignoring it when pulling it in.

It would probably be a reasonable large enhancement to add support for them. It would also depend heavily on which type of database you're connecting to. Our JSON column only works with internal tables, but we also can't assume that all JSON columns in external DBs are arrays (since an object is valid JSON).

[...]

Issue #3382, opened 2021-11-15
JSON Support - Postgres/MySQL


shogunpurple:
Depends on https://github.com/Budibase/budibase/issues/3381

When I'm working with a plus datasource that has support for a JSON field, I want to be able to leverage the JSON types available in that database inside budibase, so that I can perform JSON specific CRUD operations against that database without performing any additional processing in JS/Handlebars. 

**COS**
- Account for the JSON type in postgres and mysql data source plus type generation
- Account for JSON type in custom datasources for mysql and postgres
- Ensure that the same UX that occurs for https://github.com/Budibase/budibase/issues/3381 is used for these SQL sources. Note - we will still need to generate a schema from these fields, until we have entirely dynamic JSON support later on

chaoticefx:
Hi! Could I check in to see if this is still in the roadmap or if it's been done but I'm not seeing it somewhere? I see above that it was added to "November 21" in "Roadmap 2022" but I don't see it in "Roadmap 2022". I just started a POC for my company and am really loving Budibase so far, being able to natively interact with the JSON in PG/MySQL would be a great addition!

aderici:
If you don't support this, how can one implement multi file uploads into s3 and save those into a record in mysql? also can you add a native attachment field type so that attachments (saved into s3 or somewhere else) can be nicely previewed in table list view? Like the internal db attachments,

andz-bb:
we'll discuss this with engineers in bug grooming on the 1st

Athik321:
i am using mysql datasource now in a table i want to add attachment in field type column how can i do it

ConorWebb96:
@Athik321,

You might want to look into [custom plugins](https://github.com/Budibase/plugins) there is a plugin that allows you to attach attachments to rows within Budibase/

Athik321:
When I upload an image using the Budibase File Upload plugin and the data gets stored in your MySQL table as a long text (base64 encoded string), it takes up a lot of space. To handle this efficiently, what can i do ?

ConorWebb96:
@Athik321,

Thats just the way the plugin works. Feel free to ask on the plugin's Github repository if you have any questions related to the plugin.

Athik321:
The data being stored is very long; can I shorten it? And where was the uploaded image stored?

PR #12772, opened 2024-01-12
Fix array type missing from query schema selector

melohagan:
## Description
This bug has been around for a while and a user wasn't able to display nested MongoDB documents, so I wanted to have a stab at it. 

This will allow you to use the default schema assigned to nested array objects, so that you can use that data in a nested table block or repeater block. For example accounts with nested product docs, etc.
The bindings also appear in the bindings drawer.

This isn't complete support for JSON/Array type columns in queries and SQL, but it allows you to at least use your data even if you can't edit the nested schemas in the UI.  

The 'Array' type will now appear in the *Schema* tab of the Query Viewer, but you cannot change it (read only).

## Addresses
- https://github.com/Budibase/budibase/issues/3382
- https://github.com/Budibase/budibase/issues/8389

## Screenshots

**BEFORE**
BEFORE_MONGODB_QUERY

**AFTER**

*Read query with nested data*
Read query

*Schema showing 'Array' type*
Screenshot 2024-01-13 at 20 13 10

*Table block using nested data example*
JSON array from query

*Card block example, using the bindings for Title, Description*
cards block example

*Bindings displaying in the drawer*
bindings from query json array

*Much more efficient than making a second nested query for the same data. Now only one POST is done by the outer data provider*
only one post query

*If the array has primitives and not nested objects, it will still work:*
non object array


packages/frontend-core/src/utils/json.js
```diff
@@ -119,3 +119,36 @@ const extractJSONSchemaKeys = (jsonSchema, squashObjects = false) => {
...
 }
+
+const isObject = test => {
+  return (
+    typeof test === "object" &&
+    !Array.isArray(test) &&
+    test !== null &&
+    !(test instanceof Date)
+  )
+}
```
samwho, line 130: This feels incomplete. For example, `isObject(new Map())` would return true.

samwho (commented):
Have given this a once-over and nothing jumps out at me, but I'd recommend getting a review from someone more familiar with the frontend aspects of the code.

packages/frontend-core/src/utils/json.js
samwho: I took a look at some other implementations of `isObject` on StackOverflow and NPM and there doesn't seem to be any canonical implementation that people agree on (the `is-object` npm package has the same `new Map()` problem).

What's the goal of the function? Maybe it's fine that it's incomplete?

adrinr: If passing an empty object returns false, it seems like the function name is wrong. Could it be more verbose? What are you actually trying to get from it?

packages/client/src/components/app/forms/Form.svelte
```diff
@@ -64,7 +64,7 @@
 
   // Fetches the form schema from this form's dataSource
   const fetchSchema = async dataSource => {
-    if (dataSource?.tableId && dataSource?.type !== "query") {
+    if (dataSource?.tableId && !dataSource?.type?.startsWith("query")) {
```
adrinr, line 67: Could we not be more explicit? Could we exclude the specific types, and can we use the typed ones?

packages/frontend-core/src/utils/json.js
melohagan: I think it probably is, specifically it's testing a a property of an API JSON response body is an object, i.e. in the form:

```json
"objProp": {
  "name": "Alice",
  "age": 30
}
```

I don't think we ever cast to an instance of Map.

packages/client/src/components/app/forms/Form.svelte
melohagan: Do you have an example of specific types being used in another svelte component?

packages/frontend-core/src/fetch/QueryArrayFetch.js
```diff
@@ -0,0 +1,25 @@
+import FieldFetch from "./FieldFetch.js"
+import {
+  getJSONArrayDatasourceSchema,
+  generateQueryArraySchemas,
+} from "../utils/json"
+
+export default class QueryArrayFetch extends FieldFetch {
+  async getDefinition(datasource) {
+    if (!datasource?.tableId) {
+      return null
```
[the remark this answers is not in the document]
melohagan: I don't think it's needed, but I didn't want to exclude it because I used the `QueryFetch` class as a template, and it had that check.

packages/frontend-core/src/utils/json.js
melohagan: Renamed to `hasSchema` to better indicate the purpose of the function.

samwho (approved):

packages/types/src/documents/app/row.ts
```diff
@@ -7,6 +7,7 @@ export enum FieldType {
   NUMBER = "number",
   BOOLEAN = "boolean",
   ARRAY = "array",
+  QUERY_ARRAY = "queryarray",
```
mike12345567, line 10: One thing that we have been doing, in `packages/types/src/documents/app/table/schema.ts` - we've been defining the structure of Field schemas, the `QUERY_ARRAY` type - if it has any special properties etc it would be worth building out a specification for what the schema actually looks like/what is expected in the table/query schema definition. Right now these definitions are for tables, however someone could easily try to use this type for a table, so its probably best to define what is expected.

mike12345567 (approved):
LGTM - just one point about improving the typing around this.

melohagan:
Holding off for the time being. Chatted with Michael, and ideally we want to make some typing improvements first: https://linear.app/budibase/issue/BUDI-7916/issue-with-how-the-schema-is-saved-between-rest-and-other-query-types

mike12345567 (approved):

PR #12788, opened 2024-01-15
Update dd-trace version

mike12345567:
## Description
Updating `dd-trace` to 5.0.0 across all packages now that the memory leak is solved, this also removes the erronous message on startup describing DataDog startup problems.

5.0.0 introduces a breaking change which doesn't affect us, simply a Typescript definition change to make sure `tracer.trace` is being used correctly (which we already were doing).

We were waiting for issue: https://github.com/DataDog/dd-trace-js/issues/3887 to be solved.

Account portal PR: https://github.com/Budibase/account-portal/pull/500 (it wasn't using the right version before, now it matches).

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -32,7 +32,7 @@
     "bcryptjs": "2.4.3",
     "bull": "4.10.1",
     "correlation-id": "4.0.0",
-    "dd-trace": "3.13.2",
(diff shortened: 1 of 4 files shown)


adrinr (approved):
Happy to see the startup error message gone! 🚀

samwho (approved):

PR #12789, opened 2024-01-16
[Bug] Incorrect handling of short passwords in user invitation

adrinr:
## Description
We are facing a problem when inviting a user and there is some creation error. The latest password limits are making this occur more. When a user submits the invite form, we mark the invite code as used. If there is a problem during the creation, such as a data validation issue, the same form can not be submitted as the codes are not valid anymore.

This PR does the following:

1. Invalidate the invite code only when the user has been created properly
2. Use a lock to prevent concurrent processes to use the same invite
3. Return the verbose error message, instead of a generic "unable to create the user"
4. Remove some unnecessary db calls

## Addresses
- [BUDI-7896 - Incorrect Handling of Short Passwords in User Invitation Process](https://linear.app/budibase/issue/BUDI-7896/bug-incorrect-handling-of-short-passwords-in-user-invitation-process)

## Screenshots

https://github.com/Budibase/budibase/assets/15987277/8730ee0b-2095-4df4-af67-2e53cced6555

### DIFF
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -425,6 +425,6 @@ export const inviteAccept = async (
       ctx.throw(400, err)
     }
     console.warn("Error inviting user", err)
-    ctx.throw(400, "Unable to create new user, invitation invalid.")
+    ctx.throw(400, err || "Unable to create new user, invitation invalid.")
   }
 }
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -12,6 +12,8 @@ import {
   InviteUserRequest,
   InviteUsersRequest,
   InviteUsersResponse,
+  LockName,
+  LockType,
   MigrationType,
   SaveUserResponse,
   SearchUsersRequest,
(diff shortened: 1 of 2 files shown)


packages/worker/src/api/controllers/global/users.ts
```diff
@@ -380,51 +383,60 @@ export const inviteAccept = async (
...
+    await locks.doWithLock(
+      {
+        type: LockType.AUTO_EXTEND,
+        name: LockName.PROCESS_USER_INVITE,
+        resource: inviteCode,
+        systemLock: true,
+      },
+      async () => {
+        // info is an extension of the user object that was stored by global
+        const { email, info } = await cache.invite.getCode(inviteCode)
```
samwho, line 395: Do we also want to do a check here to make sure a user with this email doesn't already exist?

samwho, line 395: Do we also want to do a check here to make sure a user with this email doesn't already exist?

adrinr: This is already done as part of the userSdk.save

samwho (approved):

PR #12793, opened 2024-01-16
Improve removeWorkspaceDependencies.sh

adrinr:
## Description
We created this script to build our dockerfiles. Now this is needed for ironbank, so it needs to run on all the package.json. We ran into the following problems:

1. Issues when there are escaped quotes in the package.json content
2. All the resolutions are removed, not only our internal packages

This PR simplifies the script and solves these issues


shogunpurple (approved):
Oh nice, this is way cleaner!

PR #12798, opened 2024-01-17
Global Bindings fixes and updates

deanhannigan:
## Description

- Exposed new Global bindings for the Tableblock: `Rows`, `Extra Info`, `Rows Length`, `Schema`, `Page Number`
- `PageNumber` binding fix - General fix for the binding context data in the DataProvider component. An existing issue.
- Added a `scope` prop to the `Repeater` to allow block components to surface their row data globally.
  - In general, repeaters within blocks will only return 1 row so adding it globally makes sense.
  - The default repeater behaviour remains the same. Row data/context is scoped locally to each repeater row.
- Formblock/Multistep Formblock - allow access to internal row data
  - Used the new repeater flag to switch the internal repeater context to global

## Addresses
- https://github.com/Budibase/budibase/pull/11830

### DIFF
--- a/packages/client/src/components/app/Repeater.svelte
+++ b/packages/client/src/components/app/Repeater.svelte
@@ -9,6 +9,7 @@
   export let hAlign
   export let vAlign
   export let gap
+  export let nested = false
 
   const { Provider, ContextScopes } = getContext("sdk")
   const component = getContext("component")
@@ -22,7 +23,10 @@
     <Placeholder />
   {:else if rows.length > 0}
     {#each rows as row, index}
-      <Provider data={{ ...row, index }} scope={ContextScopes.Local}>
+      <Provider
+        data={{ ...row, index }}
+        scope={nested ? ContextScopes.Global : ContextScopes.Local}
+      >
         <slot />
       </Provider>
     {/each}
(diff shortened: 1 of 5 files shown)


packages/client/src/components/app/Repeater.svelte
```diff
@@ -22,7 +23,10 @@
     <Placeholder />
   {:else if rows.length > 0}
     {#each rows as row, index}
-      <Provider data={{ ...row, index }} scope={ContextScopes.Local}>
+      <Provider
+        data={{ ...row, index }}
+        scope={nested ? ContextScopes.Global : ContextScopes.Local}
```
shogunpurple, line 28: Is there any reason this can't always just be global?

shogunpurple: Ah that makes sense - we would need a new way to key the repeater rows in the global context. Is `global` a better name for this property then?

deanhannigan: Yep agreed. It does step on the current nested flag used elsewhere in the block components. I've just replaced it with a configurable scope prop instead.

shogunpurple (approved):
LGTM

Issue #12783, opened 2024-01-15
Validation does not work for search fields of ejected table block


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.14.7

**Describe the bug**
When trying to add validation to a table block search field, it does not work.

**To Reproduce**
1. Add a table block for the *Employees* sample data table
2. Add the search field for 'Email'
3. Eject the table block
4. Add validation to the email field, and add the 'Required' or 'Must match regex' validation
5. See that the validation does not trigger (even if I add a button with the **Validate Form** action)

**Expected behavior**
If I eject a table block with search fields, I should be able to add validation to those search fields.

**Screenshots**
![Screenshot 2024-01-15 at 09 59 16](https://github.com/Budibase/budibase/assets/101575380/3a025474-7c9e-4c42-92eb-12d51e0e5eab)

Validation error messages never show!

PR #12799, opened 2024-01-17
Support custom validation rules for ejected block fields

melohagan:
## Description
As a user I wanted to add validation to search fields of an ejected table block, for a custom searching scenario.
This was not possible because table and card blocks were disabling validation for search fields. This makes sense because you wouldn't normally want search fields to be validated, however once the block is ejected and a user adds a custom validation rule, they would expect that rule to be enforced.

I have changed the `disableValidation` prop to `disableSchemaValidation`. This means that search fields in blocks won't trigger any schema level validation, but if a user ejects the block and adds their own validation rule then that will be respected.

## Addresses
- https://github.com/Budibase/budibase/issues/12783

## Screenshots
Screenshot 2024-01-17 at 10 22 50
*Table block with search fields - no validation errors are triggered from the schema*

Screenshot 2024-01-17 at 10 23 56
*Table block ejected, and custom validation rule added to the 'Email' search field*

Screenshot 2024-01-17 at 10 24 45
*Custom validation on search field is now respected.*

### DIFF
--- a/packages/client/src/components/app/blocks/CardsBlock.svelte
+++ b/packages/client/src/components/app/blocks/CardsBlock.svelte
@@ -88,7 +88,7 @@
     <BlockComponent
       type="form"
       bind:id={formId}
-      props={{ dataSource, disableValidation: true }}
+      props={{ dataSource, disableSchemaValidation: true }}
     >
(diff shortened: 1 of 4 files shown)


PClmnt (approved):
LGTM!

PR #12807, opened 2024-01-18
[Fixed] Calculate creators count when group role changes

jvcalderon:
## Description

Creators by groups role were not being calculated.
See https://github.com/Budibase/budibase-pro/pull/252 to get more info.

## Addresses

https://linear.app/budibase/issue/GROW-286/creators-quota-update-on-groups-change

## Feature branch env
[Feature Branch Link](http://fb-feature-count-creators-in-groups.fb.qa.budibase.net)


packages/shared-core/src/sdk/documents/users.ts
```diff
@@ -70,7 +70,7 @@ export function hasAppCreatorPermissions(user?: User | ContextUser): boolean {
   return _.flow(
     _.get("roles"),
     _.values,
-    _.find(x => x === "CREATOR"),
+    _.find(x => ["CREATOR", "ADMIN"].includes(x)),
```
jvcalderon, line 73: Admin roles was inferred from other places of the model, but I've added this in addition, just in case.

packages/backend-core/src/index.ts
```diff
@@ -2,6 +2,7 @@ export * as configs from "./configs"
 export * as events from "./events"
 export * as migrations from "./migrations"
 export * as users from "./users"
+export * as usersUtils from "./users/utils"
```
adrinr, line 5: Nitpick, but could this be called `userUtils`?

packages/backend-core/src/users/db.ts
```diff
@@ -335,7 +336,7 @@ export class UserDB {
       }
       newUser.userGroups = groups || []
       newUsers.push(newUser)
-      if (isCreator(newUser)) {
+      if (await isCreator(newUser)) {
```
jvcalderon, line 339: Now, in pro submodule, `isCreator` requires a group role check, so it is an async funtion

adrinr: Backend-core is not using pro. This is using the code from shared-core, and as far as I can see, it is still using a sync call. Am I wrong?

packages/shared-core/src/sdk/documents/users.ts
adrinr: Is this "CREATOR" (and now "ADMIN") not typed in some enum we can use, instead of having magic strings?

packages/backend-core/src/index.ts
jvcalderon: Changed 👍

packages/shared-core/src/sdk/documents/users.ts
jvcalderon: That roles are defined in the fronted module, I imagine we don't have to create dependencies from fronted to shared. Maybe we need to move that constant here. Maybe in a separated issue? What do you think?

packages/backend-core/src/users/utils.ts
```diff
@@ -1,22 +1,53 @@
...
+  }
+  return isCreatorByUserDefinition
+}
+
+async function isCreatorByGroupMembership(user?: User | ContextUser) {
+  const userGroups = user?.userGroups || []
+  if (userGroups.length > 0) {
+    const db = context.getGlobalDB()
+    const groups: UserGroup[] = []
+    for (let groupId of userGroups) {
```
shogunpurple, line 34: Would it make sense to use `Promise.all` here?

shogunpurple (approved):
LGTM

packages/shared-core/src/sdk/documents/users.ts
jvcalderon: https://linear.app/budibase/issue/GROW-294/move-role-enum-to-shared-core

PR #12810, opened 2024-01-18
Added function to get correct plan display name

ConorWebb96:
## Description
Fixed an instance were business was being displayed instead of premium. Added helper function to show the correct plan type display name.

## Screenshots
Screenshot 2024-01-18 at 15 01 34


jvcalderon (approved):
LGTM, but we should remove `capitalise` as part of this PR. Pipeline is failing.

Screenshot 2024-01-18 at 16 44 23

Issue #12790, opened 2024-01-16
Error deleting datasource on first attempt at deleting Sample Data


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.14.7

**Describe the bug**
When trying to delete the sample data, I receive an error. Refreshing the page shows that only the *Employees* table was deleted. Then I can delete the sample data fully.

**To Reproduce**
1. Create a new app and add the Sample Data
2. Delete the sample data -> error
3. Refresh page -> see that only the Employees table has been deleted.
4. Delete the sample data again -> now it works.

**Expected behavior**
I should be able to delete the sample data without any errors or page refreshing.

**Screenshots**
![Screenshot 2024-01-16 at 11 09 01](https://github.com/Budibase/budibase/assets/101575380/05d2dac5-c937-4fc0-a31d-b43047abe27b)
*Try to delete sample data*

![Screenshot 2024-01-16 at 11 09 35](https://github.com/Budibase/budibase/assets/101575380/ba6d62b1-6ac1-4bfe-a38d-59079e17d892)

![Screenshot 2024-01-16 at 11 10 06](https://github.com/Budibase/budibase/assets/101575380/fe0077b2-4530-4d52-b3fe-7201512ecf04)
*Refresh page. It only deleted the Employees table*

melohagan:
When creating the sample data, it is not saving "Badge Photo" to the Couch DB. In comparison, the attachment field in the Expenses table correctly saves an empty array.

PR #12819, opened 2024-01-19
Error deleting datasource on first attempt at deleting

melohagan:
## Description
There were two problems that have been solved:
1. Badge Photo was missing from the employee import data
2. When deleting a table with a relationship, it would bump the revision of the related table, however within the loop the revision would still be out of date - change was made to make sure the latest revision was used when deleting internal tables in a loop.

## Addresses
- https://github.com/Budibase/budibase/issues/12790

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/c8b2bea7-09ec-4b23-a9e7-c648ac58a4bf

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -9,8 +9,11 @@ import {
   CreateDatasourceResponse,
   Datasource,
   DatasourcePlus,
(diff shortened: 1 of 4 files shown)


shogunpurple (approved):
LGTM

Issue #12600, opened 2023-12-17
[ Client Library - SDK ] - Expose tools from frontend-core via SDK


poirazis:
It would be awesome, very helpful, and good at standardizing how custom somponents access and fetch data from the backend if tools inside frontend-core were exposed via the SDK.

Primarily, it would be nice to have `fetchData` and `constants` and 'utils' , i think, but you guys are better suited to decide.

Bundling budibase/frontend-core into a custom component, adds a flat 450kb as it brings in some of bbui and string-templates.
Getting it from the SDK would make things lighter and faster and standardized.

PR #12822, opened 2024-01-20
Expose fetchData and LuceneUtils via the client library SDK

poirazis:
## Description
Addresses #12600 

( This is also mostly so i can learn the piping of making a contribution, so please be kind if I messed things up )


aptkingston (approved):
LGTM! Congrats on your first contribution :)

PR #12827, opened 2024-01-22
Configure automation chaining

PClmnt:
## Description
Automation chaining was previously only available in self host. This PR updates the settings page to change Automation History -> Automations and adds a new section allowing cloud users to turn on (and self host users to turn off) the chaining of automations 

## Addresses
BUDI-7058

## Screenshots
image

## Feature branch env
[Feature Branch Link](http://fb-feat-configure-automation-chaining.fb.qa.budibase.net)


packages/builder/src/components/start/ChooseIconModal.svelte
```diff
@@ -49,6 +49,7 @@
       return
     }
     try {
+      console.log(app.instance._id)
```
mike12345567, line 52: console log needs removed

packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
```diff
@@ -35,9 +37,13 @@
   let timeRange = null
   let loaded = false
 
+  $: app = $apps.find(app => app.devId === $store.appId)
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:40]: might be worth doing `$store.appId.includes(app.appId)` as the automation section may need to handle both prod/dev apps soon!

packages/server/src/events/AutomationEmitter.ts
```diff
@@ -24,7 +15,23 @@ class AutomationEmitter {
...
   }
 
-  emitRow(eventName: string, appId: string, row: Row, table?: Table) {
+  async getMaxAutomationChain() {
+    const db = context.getAppDB()
+    const appMetadata = await db.get<App>(DocumentType.APP_METADATA)
+    let chainAutomations = appMetadata?.automations?.chainAutomations
+
+    if (chainAutomations === true) {
+      return 5
```
mike12345567, line 24: Can we make this a constant somewhere - just so the default chaining max is recorded/easy enough to find?

mike12345567 (approved):
LGTM - just some minor comments!

Issue #12476, opened 2023-11-30
[BUDI-7786] Options picker (dropdown) opens above even though it is at the top of the screen


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Budibase Version: 2.13.20

**Describe the bug**
The dropdown is incorrectly popping above the field, even when the field is at the top of the screen.

**To Reproduce**

1. Add two options picker to the top of the screen inside a two column field group
2. Open the second options picker
3. See that it incorrectly opens above the field

**Expected behavior**
Dropdown should drop below when the field is at the top of the screen.

**Screenshots**

![Screenshot 2023-11-30 at 11 21 21](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/8c1674e1-15c8-47cc-835c-a36eca656edd/b1da7b8b-dc91-42a9-9015-2fe86bebda97?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC84YzE2NzRlMS0xNWM4LTQ3Y2MtODM1Yy1hMzZlY2E2NTZlZGQvYjFkYTdiOGItZGM5MS00MmE5LTkwMTUtMmZlODZiZWJkYTk3IiwiaWF0IjoxNzAxMzQzNDg2LCJleHAiOjE3MDE0Mjk4ODZ9.editIcDfPoslFeqEicAqE5W4shnAnauXjVhh26_U8kY)

*First picker is fine*

[...]

PR #12836, opened 2024-01-22
Revert old picker styling from 2.10

melohagan:
## Description
I had added some styling as part of the old searchable relationship picker in **2.10** however since it has been improved, these old styling props are no longer needed, and are causing issues, so reverting.

Original commit: https://github.com/Budibase/budibase/commit/3f1ec54fe99a6ebc26154c707bc8720a9e4c919c

## Addresses
- https://github.com/Budibase/budibase/issues/12476


deanhannigan (approved):
LGTM!

PR #12846, opened 2024-01-23
Fix query preview endpoint to be consistent with rest of query API

mike12345567:
## Description
As part of reviewing https://github.com/Budibase/budibase/pull/12772 with @melohagan - we decided that it would be beneficial if the query API could make use of subtypes, similar to how the table schema works. The query API is mostly capable of this, however the `preview` endpoint returned a flattened schema structure. The preview is the only part of the API which didn't really support returning more than just a string type, so the idea was to fix this.

The majority of this work involved cleaning up the `QueryViewer` and `RestQueryViewer` so that the preview response being a more complex structure (minimum `{ type: "string" }` for example) didn't cause issues.

## Addresses
- https://linear.app/budibase/issue/BUDI-7916/issue-with-how-the-schema-is-saved-between-rest-and-other-query-types


packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -386,6 +394,7 @@
 
   onMount(async () => {
     query = getSelectedQuery()
+    schema = query.schema
```
melohagan, line 397: On line 100 we have the following code:

```js
$: hasSchema =
    Object.keys(schema || {}).length !== 0 ||
    Object.keys(query?.schema || {}).length !== 0
```

because the schema is now set in the onMount, are we able to simplify that to:

```js
$: hasSchema = Object.keys(schema || {}).length !== 0
```

packages/server/src/api/controllers/query/index.ts
```diff
@@ -1,15 +1,21 @@
...
 import sdk from "../../../sdk"
 import { QueryEvent } from "../../../threads/definitions"
-import { ConfigType, Query, UserCtx, SessionCookie } from "@budibase/types"
+import {
+  ConfigType,
+  Query,
+  UserCtx,
+  SessionCookie,
+  QuerySchema,
+  FieldType,
```
melohagan, line 17: Just a sanity check - what is the difference between 'FieldType' and 'FieldTypes' ?

packages/server/src/api/controllers/query/index.ts
```diff
@@ -163,38 +169,42 @@ export async function preview(ctx: UserCtx) {
     }
 
     const { rows, keys, info, extra } = (await Runner.run(inputs)) as any
```
melohagan, line 171: Can we replace the `as any` here?

packages/server/src/api/controllers/query/index.ts
```diff
@@ -163,38 +169,42 @@ export async function preview(ctx: UserCtx) {
...
+    const makeQuerySchema = (type: FieldType, name: string): QuerySchema => ({
+      type,
+      name,
+    })
     if (rows?.length > 0) {
       for (let key of [...new Set(keys)] as string[]) {
         const field = rows[0][key]
         let type = typeof field,
-          fieldType = FieldTypes.STRING
+          fieldMetadata = makeQuerySchema(FieldType.STRING, key)
```
melohagan, line 181: NAB: Why use a function here? Why not:

```ts
let fieldMetadata: QuerySchema
...
fieldMetadata = {
  type: FieldType.STRING,
  name: key
}
```

packages/server/src/api/controllers/query/index.ts
```diff
@@ -163,38 +169,42 @@ export async function preview(ctx: UserCtx) {
...
+        previewSchema[key] = fieldMetadata
       }
     }
     // if existing schema, update to include any previous schema keys
     if (existingSchema) {
-      for (let key of Object.keys(schemaFields)) {
+      for (let key of Object.keys(previewSchema)) {
         if (existingSchema[key]?.type) {
-          schemaFields[key] = existingSchema[key].type
+          previewSchema[key] = existingSchema[key].type
```
melohagan, line 207: This would flatten the `QuerySchema`. Do we not want:

```ts
previewSchema[key] = existingSchema[key]
```

Or is this purely for backwards compatibility with existing schemas?

packages/server/src/sdk/app/queries/queries.ts
```diff
@@ -3,6 +3,27 @@ import { processStringSync } from "@budibase/string-templates"
 import { context } from "@budibase/backend-core"
 import { getQueryParams, isProdAppID } from "../../../db/utils"
 import { BaseQueryVerbs } from "../../../constants"
+import { Query, QuerySchema } from "@budibase/types"
+
+function updateSchema(query: Query) {
```
melohagan, line 8: NAB: should you explicitly provide the return type?

packages/server/src/sdk/app/queries/queries.ts
```diff
@@ -3,6 +3,27 @@ import { processStringSync } from "@budibase/string-templates"
...
+      schema[key] = { type: query.schema[key] as string, name: key }
+    } else {
+      schema[key] = query.schema[key] as QuerySchema
+    }
+  }
+  query.schema = schema
+  return query
+}
+
+function updateSchemas(queries: Query[]) {
```
melohagan, line 24: NAB: should you explicitly provide the return type?

melohagan (approved):
LGTM! Really nice refactor and unit tests 👍 

I left a couple of small comments, but I did pull down the branch and appears to be working well!

yarn.lock
adrinr, line 1: Why these changes? There is no changes on any package.json

packages/builder/src/components/integration/KeyValueBuilder.svelte
```diff
@@ -35,6 +35,7 @@
   export let bindingDrawerLeft
   export let allowHelpers = true
   export let customButtonText = null
+  export let compare = (option, value) => option === value
```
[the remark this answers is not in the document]
mike12345567: This is the default operation, basically wanted to define the default so that it can easily be overridden.

yarn.lock
mike12345567: Unsure, the `yarn.lock` has been changing wildly lately - I'm not quite sure of the cause!

packages/server/src/api/controllers/query/index.ts
mike12345567: `FieldTypes` is the old non-TS way that we did this, it hasn't been fully cleaned up yet - I'll raise a separate PR to finally be done with it.

If an enumeration is plural, not singular then it is an old pre-TS type, which needs to be gotten rid of. It doesn't really impact anything, but just for consistency sake working on getting rid of them.

packages/server/src/api/controllers/query/index.ts
mike12345567: Just for clarity sake, reduces the chance of typos/extra props being added, Typescript does protect against this, but its just a rule I follow since we work in both JS/TS (with the frontend).

packages/server/src/api/controllers/query/index.ts
mike12345567: you're right - sorry missed this!

packages/server/src/api/controllers/query/index.ts
melohagan: Ah the extra props is a good point. Kind of like constructor overloading; create an object with the specified parameters only.

packages/builder/src/components/integration/KeyValueBuilder.svelte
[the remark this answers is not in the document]
adrinr: I meant that is already the default down the component tree

PR #12854, opened 2024-01-24
Modified planTitle function to accommodate old plans

ConorWebb96:
## Description
Added missing logic to accommodate for older plans. This allows the displaying of these older plan names as well as the new plan names.


jvcalderon (approved):
LGTM

Issue #12842, opened 2024-01-23
Button group click actions cannot be set


melohagan:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.15.3

**Describe the bug**
I cannot set actions for buttons within a button group.

**To Reproduce**

1. Create a new app and a new screen
2. Add a button group
3. Click the cog icon on one of the buttons
4. Click the *On click* `No actions set` button
5. See that the actions drawer does not appear!

**Expected behavior**
I should be able to set button actions for the button group.

**Screenshots**

![buttons](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/defdac5d-f26e-4973-a0d7-af7858ff864e/6c72ad19-2a09-422e-a197-2348e241c667?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9kZWZkYWM1ZC1mMjZlLTQ5NzMtYTBkNy1hZjc4NThmZjg2NGUvNmM3MmFkMTktMmEwOS00MjJlLWExOTctMjM0OGUyNDFjNjY3IiwiaWF0IjoxNzA2MDExNDI1LCJleHAiOjMzMjc2NTcxNDI1fQ.ksePo-CZ3JnvFAqfmFD6-n35OnhREJF7fb0PRNBtPeI)

PR #12856, opened 2024-01-24
General builder bug fixes

deanhannigan:
## Description

- Refactored the builder hover behaviour into its own store. Constant updates to the builder store was triggering excessive rendering
- Fix in the `RelationshipField` to ignore initialisation until the fieldApi has been initialised properly. This didn't cause any degradation in the builder but it did cause a number of console errors.
- Added some render mitigation into the `ButtonGroup`. The constant updates to the builder caused the default button instances to re-render constantly and close the context menus. The hover fix will alleviate this almost entirely but this additional check will limit renders to only occur upon changing the associated setting value.

## Addresses
- https://github.com/Budibase/budibase/issues/12842


Ghrehh (approved):
Tested locally and LGTM, sorry about the excessive rendering from the hover behaviour, believe it me who implemented that, ty for fixing

melohagan (approved):
LGTM!

PR #12862, opened 2024-01-24
Refactor - fully replace FieldTypes with newer FieldType

mike12345567:
## Description
A quick refactor to get rid of the old `FieldTypes` enumeration, considering how core it is to all data handling in Budibase, the fact we had both `FieldType` and `FieldTypes` was confusing - righting this fully.


packages/server/src/constants/index.ts
```diff
@@ -1,16 +1,15 @@
 import { constants, objectStore, roles } from "@budibase/backend-core"
 import {
-  FieldType as FieldTypes,
+  FieldType,
   INTERNAL_TABLE_SOURCE_ID,
   Table,
   TableSourceType,
 } from "@budibase/types"
 
 export {
```
adrinr, line 9: Do we actually need these exports? Why not using directly the ones from `types`, if we are already refactoring it?

mike12345567: True - I'll finish getting rid of the exports!

adrinr (approved):

PR #12867, opened 2024-01-25
Fix for relationship save button being enabled when table was missing

PClmnt:
## Description
When editing a relationship if you removed the "to" table the save button would remain enabled, when clicked this would result in a continuous loading spinner 

## Addresses
#BUDI-7923

## Screenshots
image


deanhannigan (approved):
LGTM!

PR #12872, opened 2024-01-25
Configurable session expiry

mike12345567:
## Description
Adding a 'SESSION_EXPIRY_SECONDS' environment variable which can be set on the services to configure how long before an idle user is logged out.


melohagan (approved):
LGTM! 🚀

Issue #12875, opened 2024-01-26
Cannot select relationship field value in Design section


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.15.6 (introduced)

**Describe the bug**
Since updating to the latest version, I can no longer select a relationship field value.

**Screenshots**
![relpicker](https://github.com/Budibase/budibase/assets/101575380/045c4c3e-1ad0-492f-88bd-9944772037cd)

  
**App Export**
[My app-export-1706262644005.tar.gz](https://github.com/Budibase/budibase/files/14063133/My.app-export-1706262644005.tar.gz)

**Desktop (please complete the following information):**

**Additional context**
https://ptb.discord.com/channels/733030666647765003/1200368898831290480

melohagan:
Fixed in version **2.15.7**

PR #12876, opened 2024-01-26
Cannot select relationship field value in design section

melohagan:
## Description
Fix for recent change in **2.15.6** caused relationship field values to no longer be selectable.

## Addresses
- https://github.com/Budibase/budibase/issues/12875


deanhannigan (approved):
LGTM!

PR #12881, opened 2024-01-26
Attempting to fix CI Postgres test failures

mike12345567:
## Description
Attempting to fix issue with Postgres not being fully started in Github CI - previous message the DB was starting, this message means the database is fully started.


samwho (approved):
Oof, great find.

mike12345567:
> Oof, great find.

Sadly just a different error now - yay!

packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -15,7 +16,9 @@ export async function getDsConfig(): Promise<Datasource> {
         .withExposedPorts(5432)
         .withEnv("POSTGRES_PASSWORD", "password")
         .withWaitStrategy(
-          Wait.forLogMessage("database system is ready to accept connections")
+          Wait.forLogMessage(
+            "PostgreSQL init process complete; ready for start up."
+          ).withStartupTimeout(new Duration(2, TemporalUnit.SECONDS))
```
samwho: Could we not instead do a proper healthcheck here? Hit the Postgres port, run a simple query, assert it worked? Very much not a fan of time-based waiting.

mike12345567: Same - I've been trying to work out how to do a proper query healthcheck, its not particularly simple, but might just have to go for it. Annoyingly the database has to restart to be fully init'd the first time - so the started message is only correct the second time - which we can't watch for.

mike12345567: Switched to `@testcontainers/postgres` - their own prebuilt testcontainer for it!

packages/server/scripts/integrations/postgres/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 services:
   db:
     container_name: postgres
-    image: postgres:15-bullseye
+    image: postgres:latest
```
samwho, line 5: Is this strictly necessary? Seems risky.

packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -1,28 +1,23 @@
...
-const isMac = process.platform === "darwin"
+let container: StartedPostgreSqlContainer | undefined
 
 export async function getDsConfig(): Promise<Datasource> {
   try {
     if (!container) {
-      // postgres 15-bullseye safer bet on Linux
-      const version = isMac ? undefined : "15-bullseye"
-      container = await new GenericContainer("postgres", version)
+      const pgContainer = new PostgreSqlContainer()
```
samwho, line 12: We probably want to be explicit with the version we use here, by default it appears to be Postgres 13: https://github.com/testcontainers/testcontainers-node/blob/cce5bccd7f4dee04de3948bdf7dae52e21969883/packages/modules/postgresql/src/postgresql-container.ts#L10.

samwho (changes requested):

packages/server/scripts/integrations/postgres/docker-compose.yml
mike12345567: this is just one of our manual scripts for testing, it was always latest before but had to be changed due to issues with Linux+Docker compose compat issues, those seem to have been resolved so can go back to latest.

samwho (approved):

PR #12888, opened 2024-01-28
Time since published

mikesealey:
## Description
When users navigate to the "Publish" pop-up in the top right, and click "Publish", upon successful publish the green text would say "Published a few seconds ago" but would not change unless the page is refreshed.

This PR tackles this, by adding a `setInterval` which will, once every 60 seconds, check the time since last publish, and update the text accordingly

## Screenshots

image
Under previous circumstances, if left alone this status never changes

With these changes, time since last published is easily observable
image
image

### DIFF
--- a/packages/builder/src/components/deploy/AppActions.svelte
+++ b/packages/builder/src/components/deploy/AppActions.svelte
@@ -57,7 +57,6 @@
   $: canPublish = !publishing && loaded && $sortedScreens.length > 0
   $: lastDeployed = getLastDeployedString($deploymentStore)
 
-  // Fixes "Last Published"
   onMount(() => {
     const interval = setInterval(() => {
       lastDeployed = getLastDeployedString($deploymentStore)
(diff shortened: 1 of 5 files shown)


aptkingston (commented):
Thanks for the contribution @mikesealey!

While this does solve the problem, I think we could do it in an even easier way - just update the `lastDeployed` time whenever the popover is opened. That means that every time the user opens the popover they'll always see the accurate time since the last deployment (and you wouldn't leave this popover open, so we don't need to worry about frequently refreshing it).

We already have an `on:open` callback defined here for the popover which contains this text, so we can simply recompute the `lastDeployed` value inside there.

If you're up for making this change then that would be great!

mikesealey:
Hey Andrew, thanks for your feedback. This is still on my to-do list 

Most of my time using Budibase so far, I've left the pop-over open, but I understand that the typical user might not.

Your suggestion sounds great, but perhaps also there could be an interval after which the popover closes itself? Or would you rule out any uses of SetInterval?

PR #12898, opened 2024-01-29
App Preview Skeleton

Ghrehh:
## Description
- Adds a skeleton loader to improve the experience for previewing you apps.

- Differs a little from the designs outlined in the ticket because the original mock-up didn't account for the dev tools bar, cleared the changes with design.

- Also added a cut out effect to the text of the dev tools bar on the skeleton, folks can let me know if they think this is a little too much, it's most obviously visible on the darker themes.

### Loading logic

The skeleton is inserted into the document returned from the server so that it appears during the potentially sizeable period between initial document load and when svelte is loaded. The client app then removes the skeleton once it's ready to render content.

The builder design section uses a different document than the production apps and the portal preview section, and it has a unique load handler. For this case, the builder renders the skeleton loader until the client is ready to render.

For the portal app list preview, the builder also displays an unanimated skeleton loader until the client app iframe loads its document. This was because, even on fast connections there was a notable flash when selecting another app from your list before the client document itself was loaded. Once the client app document is loaded, the builder then hides its skeleton and the aforementioned skeleton loader placed on the client app takes over.

I did attempt to use svelte ssr hydration to "take over" the skeleton loader in the client and stop rendering it that way, as that seems a little more "sveltey". But it didn't seem to work, and honestly the documentation is very sparse. I'm guessing unless your rendering exactly the same app tree it doesn't know how to hook the client app logic in properly, having the same top level components doesn't seem to be enough.

### Animation

There's a shimmer animation on the skeleton, I used an svg for the "lines" of the skeleton to create a cutout, so that the header and lines share a single background image. This was done so that the "shimmer" effect is continuous for all the elements displayed, instead of each having their own animation that would very in speed based on the size of the element.

### Caveats

The loader doesn't take into account if the default page of the app has completely hidden its nav. This is because the value marking navigation as hidden is set per screen, and this would require making a bunch of additional db requests in the document prerender process with is obviously not ideal for something purely cosmetic.

Another option would be to add some sort of memoization endpoint that caches whether the default page of the app has its nav hidden or not. This proved to be really complicated though, as this needs to take into account things like user auth levels, custom user auth levels, whether home pages are selected or not, and then whether these values have been edited or deleted, so I opted to just punt on it. 

### Misc

Another potentially interesting find is that currently in production, the default selected screen for apps with no home page is different in the builder design section and in the portal/client app proper. Probably not a huge issue, but should maybe be unified in future.

## Addresses
- https://linear.app/budibase/issue/BUDI-7679/apps-list-skeleton-loader-for-viewing-app

|description|gif|
|---|---|
|lightest theme|lightest|
|light theme|light|
|dark theme|dark|
|darkest theme|darkest|
|midnight theme|midnight|
|nord theme|nord|
|no dev tools|published|
|vertical nav|sidenav|
|design section desktop|design desktop|
|design section mobile|design mobile|
|published app|production-client|

## Feature branch env
[Feature Branch Link](http://fb-app-skeleton-2.fb.qa.budibase.net)


packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -28,6 +30,46 @@
...
+  let previousIframeUrl = null
+  let appLoading = true
+  let fontsLoading = true
+  //$: loading = appLoading || fontsLoading;
+  $: loading = appLoading || fontsLoading
+
+  const handleLoad = e => {
+    //setTimeout(() => loading = false, 300);
+    const doc = e.currentTarget.contentDocument
+    console.log(doc.fonts)
```
Comment (deanhannigan) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:43]: 🥚

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -28,6 +30,46 @@
   }
 
   $: fetchScreens(app?.devId)
+
+  let previousIframeUrl = null
+  let appLoading = true
+  let fontsLoading = true
+  //$: loading = appLoading || fontsLoading;
```
Comment (deanhannigan) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:37]: Remove comment

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -28,6 +30,46 @@
...
+  $: loading = appLoading || fontsLoading
+
+  const handleLoad = e => {
+    //setTimeout(() => loading = false, 300);
```
Comment (deanhannigan) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:41]: Remove comment

deanhannigan (changes requested):
While I was testing, the `appLoading` parameter appears to get stuck as true in most cases and the animation wouldn't clear. The `dev-tools-loaded` only fires once so the `appLoaded` flag is never reset.

Also, in the figma there are 3 variants of the loading animation. One for a top navigation bar, one for the left and another for no bar at all. Were these scrapped?

aptkingston:
Holding off reviewing this until the new changes to move it into the client lib are added 👍

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -28,6 +34,20 @@
   }
 
   $: fetchScreens(app?.devId)
+
+  const receiveMessage = async message => {
+    if (message.data.type === "docLoaded") {
```
Comment (Ghrehh) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:39]: I mention why we need a document load handler to begin with in the main body of the PR.


Putting an `on:load` handler on the iframe doesn't actually fire when the document loads, it seems to fire once it thinks "enough stuff is loaded to paint" which seemed to be once most scripts were downloaded. So I added this handler, and a script that fires the event to the client app doc.

packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte
```diff
@@ -78,7 +98,17 @@
       </Body>
     </div>
   {:else}
-    <iframe src={iframeUrl} title={app.name} />
+    <div class:hide={!loading} class="loading">
```
Comment (Ghrehh) [inline packages/builder/src/pages/builder/portal/apps/[appId]/index.svelte:101]: Opting to visually "hide" the loader and the actual iframe so that stuff actually gets painted, hopefully reducing flashes.

packages/client/src/components/ClientApp.svelte
```diff
@@ -103,140 +120,140 @@
...
-                class:preview={$builderStore.inBuilder}
-                class:tablet-preview={$builderStore.previewDevice === "tablet"}
-                class:mobile-preview={$builderStore.previewDevice === "mobile"}
-              >
-                <!-- Actual app -->
-                <div id="app-root">
-                  {#if showDevTools}
-                    <DevToolsHeader />
-                  {/if}
+<div
```
Ghrehh, line 123: Previously this wasn't displayed at all until `dataLoaded` was set to true, then the entire tree was rendered at once, which caused a few milliseconds of choppy visuals. Always rendering the tree in a hidden div seems to fix this, and doesn't through any errors, so I assume its okay 🤷

packages/client/src/components/FreeFooter.svelte
```diff
@@ -13,6 +13,7 @@
 
 <style>
   .free-footer {
+    min-height: 51px;
```
Ghrehh, line 16: This previously flip-flopped in height as fonts loaded in, setting it to a constant min of 51px fixes this.

packages/frontend-core/src/components/ClientAppSkeleton.svelte
```diff
@@ -0,0 +1,346 @@
+<script>
+  export let sideNav = false
+  export let hideDevTools = false
+  export let hideFooter = false
+  export let noAnimation = false
+</script>
+
+<svelte:head>
```
Ghrehh, line 8: Were using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries) here to simulate screen resizing for loaders in the builder when there's no iframe. Unfortunately the svelte preprocessor doesn't support them in the svelte version were on, so they need to be inserted in this way. I put the media query version alongside them as it looked neater also imo.

packages/frontend-core/src/components/ClientAppSkeleton.svelte
```diff
@@ -0,0 +1,346 @@
...
+      }
+    }
+  </style>
+</svelte:head>
+
+<div class:sideNav id="clientAppSkeletonLoader" class="skeleton">
+  <div class="animation" class:noAnimation />
+
+  {#if !hideDevTools}
+    <div class="devTools">
```
Ghrehh, line 66: This dev tools implementation is potentially a little fragile to change, so i'd appreciate thoughts on it.

Importing the actual dev tools into this component to share some of the styles is a potential option, but the way we render svelte components on the server is very bare bones, and it doesn't handle imports at all, so you need to manually compile each component, or have your build step copy them in, so I opted to do it this way.

packages/server/src/api/controllers/static/index.ts
```diff
@@ -24,14 +24,69 @@ import AWS from "aws-sdk"
...
-import { App, Ctx, ProcessAttachmentResponse } from "@budibase/types"
+import { UserCtx, App, Ctx, ProcessAttachmentResponse } from "@budibase/types"
 import {
   getAppMigrationVersion,
   getLatestMigrationId,
 } from "../../../appMigrations"
 
 import send from "koa-send"
 
+const getThemeVariables = theme => {
```
Ghrehh, line 35: spectrum css isn't loaded for the skeleton rendered in the initial client doc yet, so we need to insert the necessary styles here.

packages/server/src/api/controllers/static/index.ts
```diff
@@ -167,10 +222,23 @@ export const serveApp = async function (ctx: Ctx) {
...
     let appId = context.getAppId()
 
+    const hideDevTools = !!ctx.params.appUrl
+    const sideNav = appInfo.navigation.navigation === "Left"
+    const hideFooter = ctx.user?.license?.features?.includes("branding")
+    const themeVariables = getThemeVariables(appInfo?.theme)
+
     if (!env.isJest()) {
       const plugins = objectStore.enrichPluginURLs(appInfo.usedPlugins)
+      const Skeleton =
```
Ghrehh, line 232: Have to manually load the skeleton component here instead of putting it in the base template. 

Our build step seems to copy svelte files to the server dist folder, where they are processed at run time, so import components in components doesn't really work.

They need to be rendered individually like this or manually copied to where they're expected in the build step, and even then the paths will be different than you would expect, so this method is best imo.

packages/server/src/api/controllers/static/templates/BudibaseApp.svelte
```diff
@@ -95,7 +95,7 @@
   </style>
 </svelte:head>
 
-<body id="app">
```
Ghrehh, line 98: This change was necessary to render multiple components in the app.hbs template body, the location of the client app main svelte entry point remains the same, so it hopefully won't break any tests.

packages/server/src/api/routes/static.ts
```diff
@@ -51,8 +51,8 @@ router
     controller.deleteObjects
   )
   .get("/app/preview", authorized(BUILDER), controller.serveBuilderPreview)
-  .get("/:appId/:path*", controller.serveApp)
   .get("/app/:appUrl/:path*", controller.serveApp)
+  .get("/:appId/:path*", controller.serveApp)
```
Ghrehh, line 55: This was kind of weird, I don't know if this is a typescript thing, with sending two routes with different potential query params to the endpoint, but changing the order of these routes resulted in different values being available in `ctx` in the controller.

Putting them in this order let me check if appUrl was provided, and hence if a published app was being rendered.

Ghrehh:
@deanhannigan @aptkingston I've addressed both of your feedback if you could give this another look when you get a chance please. I updated the original PR message with more context and added some comments with context also which maybe helpful.

packages/server/src/api/controllers/static/templates/BudibaseApp.svelte
adrinr: The issue was that it was using body, or that we had duplicated ids?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -167,10 +222,23 @@ export const serveApp = async function (ctx: Ctx) {
...
+        require("@budibase/frontend-core/src/components/ClientAppSkeleton.svelte").default
```
adrinr, line 233: Do we need the full path? Can't we require `ClientAppSkeleton` from the index?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -167,10 +222,23 @@ export const serveApp = async function (ctx: Ctx) {
     const appInfo = await db.get<any>(DocumentType.APP_METADATA)
     let appId = context.getAppId()
 
+    const hideDevTools = !!ctx.params.appUrl
+    const sideNav = appInfo.navigation.navigation === "Left"
```
adrinr, line 226: Where is this `Left` magic string coming from? Don't we have any type we can use?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -167,10 +222,23 @@ export const serveApp = async function (ctx: Ctx) {
...
+    const hideFooter = ctx.user?.license?.features?.includes("branding")
```
adrinr, line 227: We should try to use `isFeatureEnabled` from `packages/server/src/features.ts`

packages/server/src/api/controllers/static/index.ts
```diff
@@ -146,7 +201,7 @@ const requiresMigration = async (ctx: Ctx) => {
   return requiresMigrations
 }
 
-export const serveApp = async function (ctx: Ctx) {
+export const serveApp = async function (ctx: UserCtx) {
```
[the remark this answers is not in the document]
adrinr: That's just for typing, so there is no runtime change

adrinr (changes requested):

packages/server/src/api/routes/static.ts
Ghrehh: If you're referring to the ordering of the matchers, yes I'm aware, I don't see how the explains the issue. 

The paths are qualified such that I don't see how how both `/app/some_app_url` and `/app_some app_url` would both just match the first one and ignore the second, unless for some reason the `/` characters are matching anything.

Unless you mean two matchers can't point to the same controller function? But I don't really see how that explains the issue either.

packages/server/src/api/controllers/static/index.ts
Ghrehh: It's the positioning of the navigation. I don't believe there's currently any type or constant for it, both the place it's set and the place it's consumed currently use the raw string:

https://github.com/Budibase/budibase/blob/app-skeleton-2/packages/builder/src/pages/builder/app/%5Bapplication%5D/design/%5BscreenId%5D/%5BcomponentId%5D/_components/Navigation/index.svelte#L90
https://github.com/Budibase/budibase/blob/app-skeleton-2/packages/client/src/components/app/Layout.svelte#L41

packages/server/src/api/controllers/static/index.ts
Ghrehh: I'm a little confused by this,  this was throwing a ts build error that I just fixed, and I'm pretty sure the types it's expecting are defined [here])https://github.com/Budibase/budibase/blob/app-skeleton-2/packages/types/src/sdk/licensing/feature.ts) which don't seem to be referenced in the file the function you're referring to is defined in. The features defined there don't seem to match onto the ones defined in the linked types file either.

packages/frontend-core/src/components/ClientAppSkeleton.svelte
aptkingston: For dev tools, could we just have the black background and nothing else? I think that perfectly fits in line with a skeleton. It means the implementation here can be extremely simple since its just a single empty div, but it still means that the page layout is correct. It would just look like the content loads in when it's ready, which I think would be perfectly fine.

I think that's definitely a worthy tradeoff to keep this implementation simpler and less fragile.

packages/server/src/api/controllers/static/index.ts
aptkingston: Yep it's because we just have a single export of all components rather than named exports, so it's not tree-shakable and will pull in all components. Importing it from the path directly here to save on bloat is a good idea 👍

packages/client/src/components/ClientApp.svelte
aptkingston: Hmm I'm a little apprehensive about this change... But if you're not getting any errors then maybe we're ok. With changes like this I always just worry that there's something written to expect data to be available on the initial load, which now there might not be. If everything is perfectly reactive then this won't be a problem, but it's just a risky change.

If it adds a decent amount of value then I'm happy to risk it.

We couldn't try a (less bulletproof) solution of waiting a few milliseconds before removing the skeleton, and keeping this loading blocker, could we? Tree render speed will vary of course but if we could try some small timeout values and seeing if that also fixes the choppy visuals without any risk.

aptkingston (approved):
Amazing work @Ghrehh! You've really gone into detail here and it shows - it's super smooth and massively improves user experience.

Left a few suggestions, but no real issues.

Only minor UI thing I found (and this is extremely minor, but you've minmaxed everything else so I figured you'd want to fix this too) is that when you load a preview URL full screen that includes dev tools, there's a rounded corner that then disappears when the real app loads.

Skeleton:
image

Loaded:
image

adrinr (approved):
Backend-changes look fine to me!

deanhannigan (approved):
LGTM! Looks fantastic!

PR #12900, opened 2024-01-29
Add a check to stringSplit that gives a nicer error message if a non-string is passed.

samwho:
## Description

We're seeing an error in production where an automation is passing a non-string into `stringSplit` and it fails with the error:

```
TypeError: value.split is not a function
```

This PR checks for non-strings and produces a nicer error when it happens so we can figure out what's happening.


mike12345567 (approved):
Good to catch errors properly here as well!

PR #12904, opened 2024-01-30
Test helpers running as javascript

adrinr:
## Description
Test and fix handlebars helpers running as javascript.
Many helpers rely on having handlebars blocks, and it does not make much sense to have it converted to js, given that there are native ways to achieve most of them. Because of this, we are just removing them from the js section (and we should not allow its conversion)

Requires https://github.com/Budibase/handlebars-helpers/pull/20 to be merged and published

### DIFF
--- a/packages/string-templates/src/helpers/list.js
+++ b/packages/string-templates/src/helpers/list.js
@@ -27,6 +27,7 @@ module.exports.getHelperList = () => {
   return helpers
 }
 
+// Some helpers depend on handlebars injecting some parameters. This function adjust the helpers when required
 function adjustJsHelpers(helpers) {
   const result = { ...helpers }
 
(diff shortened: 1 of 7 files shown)


packages/string-templates/src/helpers/list.js
```diff
@@ -3,6 +3,8 @@ const helperList = require("@budibase/handlebars-helpers")
 
 let helpers = undefined
 
+const helpersToRemove = ["sortBy"]
```
samwho, line 6: How come we're removing `sortBy`?

samwho (approved):

packages/string-templates/src/helpers/list.js
adrinr: Our handlebars don't work fine with returned objects, and sort by only returns objects. This is the only one we have, and obviously nobody is using it. This is the current result:
image


This kind of helpers are easily achievable via pure js, so easier to drop them

PR #12906, opened 2024-01-30
Typing improvements around automation loop tests.

samwho:
## Description

No logical changes in here, purely type changes to remove some `any`s from `loop.spec.ts`.


mike12345567 (approved):
LGTM!

PR #12908, opened 2024-01-30
Fix automation looping failing when binding returned an integer.

samwho:
## Description

This PR fixes [this bug](https://app.datadoghq.eu/apm/error-tracking/issue/c62c2616-b558-11ee-940c-da7ad0900005?refresh_mode=sliding&tb=%40http.url&view=spans&from_ts=1705943979351&to_ts=1706548779351&live=true).

What's happening is that the loop `binding` field is returning an integer, which before this PR would make it fall through into the branch in `getLoopIterations` that called `splitString`, except we pass an number and not a string so it fails.

This change makes it so that if the binding returns a standalone integer, we use that an a 1-length array.

Additionally I renamed `executeSychronously` because I was confused when I read it and learned it executed asynchronously. I figured out that it means synchronous as in "not using Bull", so I'm hoping the name `executeInThread` is more clear.

### DIFF
--- a/packages/server/src/automations/tests/loop.spec.ts
+++ b/packages/server/src/automations/tests/loop.spec.ts
@@ -44,4 +44,12 @@ describe("Attempt to run a basic loop automation", () => {
     })
     expect(resp.steps[2].outputs.iterations).toBe(3)
   })
+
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM - nice catch!

PR #12913, opened 2024-01-30
Assorted fixes

PClmnt:
## Description
- Fixes an issue wherechain  automation setting was not being replicated correctly and updated some of the App ID checking related to this on the frontend
- Fixed an issue where you couldn't trigger an automation from your app without a field defined.
- Better handling around trigger fields being undefined when using the bindings drawer in automations

## Addresses
- [BUDI-7305](https://linear.app/budibase/issue/BUDI-7305/app-action-automation-trigger-doesnt-happen-unless-a-field-is-added)


samwho (approved):
I _think_ I follow what's going on here but I have to admit that I don't have great familiarity with the frontend portions of the automation code.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -157,6 +157,7 @@
     let bindings = []
     let loopBlockCount = 0
     const addBinding = (name, value, icon, idx, isLoopBlock, bindingName) => {
+      if (!name) return
```
adrinr, line 160: When will this happen? This is running for each key on an object

adrinr (approved):

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
PClmnt: image

The key can actually be null for this pair so it's just a safeguard against that.

mike12345567 (approved):

packages/builder/src/builderStore/store/frontend.js
```diff
@@ -158,6 +158,7 @@ export const getFrontendStore = () => {
           ...INITIAL_FRONTEND_STATE.features,
           ...application.features,
         },
+        automations: application.automations || {},
```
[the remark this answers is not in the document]
deanhannigan: Ran through the changes there with @PClmnt and it shouldn't be a problem getting it merged in 👍

PR #12914, opened 2024-01-30
Plumbing for showing a maintenance page when SQS is required but missing.

samwho:
## Description

This PR adds a check for the new `maintanence` property on the call to `/api/system/environment` (was previously `infrastructure.sqs` but I generalised it a little) and if it finds it to be not empty, redirects to a page that blocks the user and forces them to perform the required maintenance. The idea is we will show this to self-hosted users who update the Budibase application code to a version that depends on SQS, but we discover that the CouchDB installation hasn't been updated to have SQS yet.

There are 2 new screens: one for showing to builders, and one for showing to app users. The one we show to builders will link to documentation on how to upgrade your Budibase installation, while the one for app users will be more generic and advise contacting the owner of the app.

I haven't yet written the documentation for the SQS migration, but the page will link to it when I have. For the time being, it's impossible to trigger this redirect unless `sqsRequired` has been set in the code.

## Addresses
- https://linear.app/budibase/issue/BUDI-7664/sqs-self-host-ui-for-detecting-lack-of-sqs-support

## Screenshots

Builder desktop:

CleanShot 2024-03-14 at 10 22 18@2x

Builder mobile:

CleanShot 2024-03-14 at 10 22 56@2x

App user desktop:

CleanShot 2024-03-14 at 10 20 01@2x

App user mobile:

CleanShot 2024-03-14 at 10 21 57@2x


aptkingston (commented):
LGTM other than one potential issue when not running self-hosted. I'll give Andy a shout to do a quick mockup of how we could style the error page.

As for end-user apps, I think it depends on whether their apps will be functional or not. If their apps will be broken when in this state, then I think we do need to redirect them to some sort of error page, with different copy than the one for builders.

aptkingston:
In terms of page design, Andy thought to just follow what we have for the other error pages - like the 502 one here (error.html):
image

May need to make it slightly wider for the additional copy though! I'd keep the copy as terse as possible too 👌

samwho:
I've restyled the page and updated the screenshots in the PR description @aptkingston. Let me know what you think!

@mike12345567 I can't remember where we landed re: a separate maintenance page for users. I imagine that installations that get into the state where this maintenance page shows will be failing most screens, should we show a user-specific maintenance page asking them to contact the Budibase admin or some such?

We also need to write a CouchDB migration guide. Right now I'm just linking to the docs homepage as a placeholder. I'll work on some docs for this, though I imagine it's not possible to complete them just because e.g. for Helm the guidance is going to be to update to a specific Helm chart version, which we haven't published yet so we don't know what version.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+  }
+  .title {
+    font-weight: 400;
+    color: var(--spectrum-global-color-gray-900);
+  }
+  .buttons {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-start;
+    margin-top: 15px;
```
mike12345567, line 76: generally when it comes to stuff like padding/margin, we try top use the `var(--spacing-{size})` variables just to try and keep spacing consistent.

mike12345567 (approved):
LGTM - one NAB comment!

mike12345567:
> I've restyled the page and updated the screenshots in the PR description @aptkingston. Let me know what you think!
> 
> @mike12345567 I can't remember where we landed re: a separate maintenance page for users. I imagine that installations that get into the state where this maintenance page shows will be failing most screens, should we show a user-specific maintenance page asking them to contact the Budibase admin or some such?
> 
> We also need to write a CouchDB migration guide. Right now I'm just linking to the docs homepage as a placeholder. I'll work on some docs for this, though I imagine it's not possible to complete them just because e.g. for Helm the guidance is going to be to update to a specific Helm chart version, which we haven't published yet so we don't know what version.

On the question of a separate page for users - not quite sure what you mean - do you mean when they load a client app?

samwho:
> On the question of a separate page for users - not quite sure what you mean - do you mean when they load a client app?

When a builder tries to go somewhere and we determine that they need to upgrade to SQS, we show a page that links to the docs to do that. This page wouldn't be appropriate to show a normal user of the app.

What they will see today is, very likely, a completely broken app. Do we want to instead show them a nicer "this Budibase installation is not configured correctly, contact your administrator to let them know" screen or some such?

mike12345567:
> > On the question of a separate page for users - not quite sure what you mean - do you mean when they load a client app?
> 
> When a builder tries to go somewhere and we determine that they need to upgrade to SQS, we show a page that links to the docs to do that. This page wouldn't be appropriate to show a normal user of the app.
> 
> What they will see today is, very likely, a completely broken app. Do we want to instead show them a nicer "this Budibase installation is not configured correctly, contact your administrator to let them know" screen or some such?

Ah yes I understand what you mean - yeah I think we should build a similar check into the client library as well, but just say to contact the administrator to get the installation updated correctly.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+<div class="main">
+  <div class="content">
+    <div class="hero">
+      <BudibaseLogo />
+    </div>
+    <div>
+      {#each $admin.maintenance as maintenance}
+        {#if maintenance.type === MaintenanceType.SQS_MISSING}
+          <h1 class="title">Please upgrade your Budibase installation</h1>
```
aptkingston, line 15: We should replace this with the BBUI `Heading` component.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+          <p>
```
aptkingston, line 16: We should replace these with BBUI `Body` components.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+            We've detected that the version of Budibase you're using depends on
+            a more recent version of the CouchDB database than what you have
+            installed.
+          </p>
+          <p>
+            To resolve this, you can either rollback to a previous version of
+            Budibase, or follow the migration guide to update to a later version
+            of CouchDB.
+          </p>
+          <div class="buttons">
```
aptkingston, line 26: We should replace this with the BBUI `Button` component.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+</div>
+
+<style>
+  :root {
+    --spectrum-global-color-gray-600: rgb(144, 144, 144);
+    --spectrum-global-color-gray-900: rgb(255, 255, 255);
+    --spectrum-global-color-gray-800: rgb(227, 227, 227);
+    --bb-indigo: #6e56ff;
+    --bb-indigo-light: #9f8fff;
+  }
```
aptkingston, line 44: We don't need to worry about adding these styles - we'll already have them loaded into scope.

packages/builder/src/pages/builder/maintenance/index.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+  a {
+    color: #e7e7e7;
+    font-family: "Roboto", sans-serif;
+    border: none;
+    font-size: 15px;
+    border-radius: 15px;
+    padding: 8px 22px;
+  }
```
aptkingston, line 53: We can remove a lot of these styles when we swap to BBUI components, as they're all pre-styled.

aptkingston (changes requested):
I think I misunderstood our conversation about whether we should use raw HTML or components - I thought this page was going to be a HTML file, like the "Houston we have a problem" page, rather than a svelte file. That's my bad.

Since this page is svelte and within the main frontend bundle, we should definitely use BBUI components here where possible. I've added in the replacements as comments. Swapping out the raw HTML for BBUI components will also improve the sizing of some of these components, which would be my only other comment. The HTML for layout is totally fine - I just specifically mean using BBUI components for the text elements and button.

Code and logic all looks good though

I'll also ping Andy once these changes are made, to sign off on the final design.

aptkingston (approved):
LGTM!

One thing to keep in mind is that older client libraries don't contain this code, so if there was maintenance running then users of old apps wouldn't know about it.

There's no easy solution to that I don't think. You'd have to either:
- Prevent maintenance if you have outdated apps
- Forcefully update all apps to the latest

I don't think either of those are feasible, so we'll just have to accept that outdated apps will just break without an explanation.

PR #12920, opened 2024-01-31
adding lint rule for console.log

shogunpurple:
## Description
We committed a `console.log` debugging statement to prod and it showed up for users. This lint rule prevents this from happening in future as CI will fail the lint check if there's any `console.log` statements lying around in the frontend codebase.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -45,6 +45,12 @@
         "no-prototype-builtins": "off",
         "local-rules/no-budibase-imports": "error"
       }
+    },
+    {
+      "files": ["packages/builder/**/*"],
+      "rules": {
+        "no-console": ["error", { "allow": ["warn", "error", "debug"] } ]
+      }
     }
   ],
   "rules": {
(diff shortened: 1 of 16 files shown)


melohagan (approved):

.eslintrc.json
```diff
@@ -45,6 +45,12 @@
         "no-prototype-builtins": "off",
         "local-rules/no-budibase-imports": "error"
       }
+    },
+    {
+      "files": ["packages/builder/**/*"],
```
mike12345567, line 50: We should probably apply this to the client library as well!

mike12345567 (approved):
LGTM - would just add linting to client as well!

.eslintrc.json
shogunpurple: I have actually already pushed an update with that, but github hasn't processed it yet. Super weird.

aptkingston (approved):
LGTM!

deanhannigan (approved):
LGTM

Issue #12928, opened 2024-02-01
Nested data providers in repeater components don't have the correct context


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.16.0
  - App Version: 2.16.0

**Describe the bug**
Nesting data providers inside repeaters which filters based on the repeater causes the results to show up as duplicated. Instead of showing the matching record it shows whatever component is rendered last.

**Expected behavior**
Allow nested data providers filtered on repeater values to pull out the correct values instead of defaulting to whichever one is rendered last.
  
**App Export**
Engineering team have a reproduced export.

ConorWebb96:
This has been fixed closing.

PR #12931, opened 2024-02-01
Remove proxying of context changes up the chain

aptkingston:
## Description
This PR makes some changes to global bindings to fix some current issues with how context works when using more complicated structures.

### Problem example
An example of a structure which is currently problematic is nesting a global context inside a local context. Inside Budibase, this would look like:
```
- Data provider 1
  - Repeater
    - Data provider 2
```
With this setup, the repeater is providing local context - meaning only components inside the repeater can see its context. The second data provider that is inside the repeater is a global data provider, meaning it will attempt to make its value available everywhere. If the repeater is repeating more than 1 time, this will cause an issue as for every iteration of the repeater, another instance of data provider 2 will update the global value.

The end result is that every usage of data provider 2 will have the same value, even though they should be unique.

### Solution
The solution is to consider global bindings only global until they reach the next local context boundary. In the example above, that would mean that the global bindings for data provider 2 are only global within the context of the repeater. This ensures that every usage of data provider 2 is unique.

It also means that bindings from data provider 2 cannot be used outside of the repeater. Basically once any sort of local context is encountered (which only happens when things are repeated) *any* bindings provided by *any* component inside that repeated tree will only be available from inside this tree. 

### Technical discussion
Previously, in the bugged version of global bindings, there's the concept of proxying context changes up the chain. Context is modeled using a store, and whenever we need local scope (for example a repeater) we add a new layer of context stores containing only the data provided at this level. The total context available to a component is the derivation of all stores.

Previously any context changes in child stores which were deemed "global" would be passed up the chain, so that the very root store would contain all global context. This causes the issue described above, as when using a repeater you would end up with multiple components updating the exact same context key in the root store, meaning all repeated items were using the same value.

I realised that this proxying is entirely unnecessary. All layers of context stores are intrinsically local, and any type of context, including global context, only makes sense to exist within it's own store level. Therefore the solution here was to remove the proxying of changes up the context store tree, and instead store all context within the immediate parent context store of any given component.

## Addresses
- https://github.com/Budibase/budibase/issues/12928

## Feature branch env
[Feature Branch Link](http://fb-nested-context-fix.fb.qa.budibase.net)


ConorWebb96:
LGTM, from testing on the feature branch env a few customer issues that have been raised seem to have been fixed as a result of this PR.

ConorWebb96 (approved):
LGTM, from testing on the feature branch env a few customer issues that have been raised seem to have been fixed as a result of this PR.

packages/builder/src/builderStore/componentUtils.js
```diff
@@ -263,11 +266,59 @@ export const getComponentName = component => {
...
+      buildContextTree(child, tree, currentBranch)
+    })
+  }
 
-  return name
+  return tree
+}
+
+/**
+ * Generates a lookup map of which content branch all components in a component
```
Ghrehh, line 312: NAB: `content branch` instead of `context branch` here I think

Ghrehh (approved):
LGTM

Tested it locally with two tables related one-to-many, the nested data provider and repeater using the relationship of the parent repeater as its datasource.

Can confirm it doesn't work on master, every item in the nested repeater is the same.

This branch fixes it, it behaves as expected, the nested repeater displays different content for each item.

So assuming I understand the bug and the setup this looks good to me 👍

shogunpurple (approved):
LGTM - although it would be really good to have some tests for this

aptkingston:
> LGTM - although it would be really good to have some tests for this

Spoke to Mihail this morning about setting up some QA wolf coverage for bindings 👍 we'll start with basic ones to ensure global/local bindings work, then flesh them out to more complex use cases like nested structures.

packages/builder/src/builderStore/componentUtils.js
aptkingston: TY 👌

PR #12932, opened 2024-02-01
Fix automation JS scripts

mike12345567:
## Description
Quick fix, the context in the `scriptRunner` wasn't generic, this passes through all props from the context rather than just `data` and `params` as transformers would use.


samwho (changes requested):
Could you add a test? 🙏

samwho (approved):
Awesome, thank you 🙏

adrinr (approved):

PR #12933, opened 2024-02-02
Clean up isolates when a request is finished.

samwho:
## Description

The `isolated-vm` migration appears to have caused a memory leak in production.

CleanShot 2024-02-02 at 09 34 17@2x

This PR is an effort to reverse that.


packages/server/src/api/index.ts
```diff
@@ -62,6 +63,8 @@ if (apiEnabled()) {
     .use(auth.auditLog)
     // @ts-ignore
     .use(migrations)
+    // @ts-ignore
+    .use(cleanup)
```
mike12345567, line 67: When does this middleware get executed? Right now based on this flow, it looks like it'll happen right before any controller is actually called, so will it not just end up building another isolate-vm?

packages/server/src/middleware/cleanup.ts
```diff
@@ -0,0 +1,16 @@
+import { Ctx } from "@budibase/types"
+import { context } from "@budibase/backend-core"
+
+export default async (ctx: Ctx, next: any) => {
+  const resp = next()
+
+  const current = context.getCurrentContext()
+  if (current?.cleanup) {
+    for (let fn of current.cleanup || []) {
+      await fn()
```
mike12345567, line 10: should probably wrap this in a `try catch` - just incase any error is thrown during the cleanup, we probably want to finish cleaning up other stuff, and alert in someway that a cleanup failed.

packages/server/src/middleware/cleanup.ts
```diff
@@ -0,0 +1,16 @@
+import { Ctx } from "@budibase/types"
+import { context } from "@budibase/backend-core"
+
+export default async (ctx: Ctx, next: any) => {
+  const resp = next()
```
adrinr, line 5: Next is async, this should be awaited in order to work as expected

adrinr, line 5:
```suggestion
  const resp = await next()
```

adrinr (changes requested):

packages/server/src/middleware/cleanup.ts
```diff
@@ -0,0 +1,30 @@
...
+      // We catch errors here to ensure we at least attempt to run all cleanup
+      // functions. We'll throw the first error we encounter after all cleanup
+      // functions have been run.
+      errors.push(e)
+    }
+  }
+  delete current.cleanup
+
+  if (errors.length > 0) {
+    throw errors[0]
```
mike12345567, line 26: Is this safe - do we run the risk of hiding some issues if there is multiple errors?

mike12345567 (approved):
LGTM - one small question, would probably wait for final confirmation from @adrinr as well as he knows this area much better!

packages/server/src/middleware/cleanup.ts
samwho: If multiple "finalisers" failed, yes, we'd hide some errors. I considered making some sort of `CompositeError` class but then you'd be hiding the stack traces of all errors. Seemed like the lesser of two evils at the time.

adrinr: Should we at least log the errors that we can then see that something else happened?

packages/server/src/jsRunner/index.ts
```diff
@@ -8,25 +8,34 @@ import {
...
         const bbCtx = context.getCurrentContext()
 
-        const vm = bbCtx?.vm
-          ? bbCtx.vm
-          : new IsolatedVM({
-              memoryLimit: env.JS_RUNNER_MEMORY_LIMIT,
-              invocationTimeout: env.JS_PER_INVOCATION_TIMEOUT_MS,
-              isolateAccumulatedTimeout: env.JS_PER_REQUEST_TIMEOUT_MS,
-            }).withHelpers()
+        let vm: VM
```
adrinr, line 19: This 3 if/else could be simplified, to make it more readable, to something like this:
```
let vm = bbCtx?.vm
if(!vm){
  // initialise vm
  // setup cleanup
}
```

adrinr (approved):
Looks good, just a couple of minor comments!

packages/server/src/middleware/cleanup.ts
samwho: I've wrapped each cleanup function in an APM span. This will make them show up in APM tracing, and any error thrown from them will get attached to the span and then show up in DataDog's error tracking dashboard.

PR #12938, opened 2024-02-02
Migrate from `vm` to `isolated-vm`

adrinr:
Reverts Budibase/budibase#12934. [These changes](https://github.com/Budibase/budibase/pull/12769) were already released, but reverted because of some automation issues.

## Description
Replacing vm2 usages in favour of isolated VM. There are multiple reasons for this, one of them being a suggestion from the vm2 developers: `TL;DR The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to [isolated-vm](https://www.npmjs.com/package/isolated-vm).`

Moving to `isolated-vm` will add extra benefits, such as:
- Isolation, removing the vulnerability of allowing users to run malicious code
- Memory limits, preventing some DDoS attacks (both voluntary or involuntary ones)

The main incompatibility between our implementation with vm2 vs the isolated-vm one is the way we inject our handlebars helpers. We could not inject the full library as as objects, so we require following the process below:
1. Use esbuild to generate a file with the required helper functions, without any external dependency
2. Create an isolated-vm module with the generated code
3. Register this module in an isolated-vm script
4. Run the script with the user's code

These changes also improve the performance quite a lot when running heavy calculations. Example, running search requests for a table with 2500 rows with 250 formula columns each. The benchmarks are fetching only 1 row. Running in feature-branches.
### Before
image

### After
image

## Addresses
- [BUDI-7884 - Isolated-vm integration
](https://linear.app/budibase/issue/BUDI-7884/isolated-vm-integration)


packages/string-templates/src/index.cjs
```diff
@@ -36,3 +36,8 @@ if (!process.env.NO_JS) {
     return vm.run(js)
   })
 }
+
```
shogunpurple, line 39: Seems like these files till have a dependency on `vm2` - is that correct?

adrinr: This is not actually used, I am working on removing it completely!

samwho (approved):

packages/server/src/jsRunner/bundles/bson.ivm.bundle.js
```diff
@@ -0,0 +1,7 @@
...(new e(t).id,this.id)}catch{return!1}}toBinary(){return new _(this.id,_.SUBTYPE_UUID)}static generate(){let t=h.randomBytes(gt);return t[6]=t[6]&15|64,t[8]=t[8]&63|128,t}static isValid(t){return t?typeof t=="string"?e.isValidUUIDString(t):Lt(t)?t.byteLength===gt:t._bsontype==="Binary"&&t.sub_type===this.SUBTYPE_UUID&&t.buffer.byteLength===16:!1}static createFromHexString(t){let n=e.bytesFromString(t);return new e(n)}static createFromBase64(t){return new e(h.fromBase64(t))}static bytesFromString(t){if(!e.isValidUUIDString(t))throw new o("UUID string representation must be 32 hex digits or canonical hyphenated representation");return h.fromHex(t.replace(/-/g,""))}static isValidUUIDString(t){return Le.test(t)||Re.test(t)}inspect(t,n,s){return s??=j,`new UUID(${s(this.toHexString(),n)})`}},Y=class e extends D{get _bsontype(){return"Code"}constructor(t,n){super(),this.code=t.toString(),this.scope=n??null}toJSON(){return this.scope!=null?{code:this.code,scope:this.scope}:{code:this.code}}toExtendedJSON(){return this.scope?{$code:this.code,$scope:this.scope}:{$code:this.code}}static fromExtendedJSON(t){return new e(t.$code,t.$scope)}inspect(t,n,s){s??=j;let r=s(this.code,n),i=r.includes(`
```
mike12345567, line 1: Should we be including this bundle in the codebase?

packages/server/src/jsRunner/bundles/index-helpers.ivm.bundle.js
```diff
@@ -0,0 +1,59 @@
...e=Buffer.from(e,"utf8")),mi.createHash("sha1").update(e).digest()}var Lr,zr=Z(()=>{Lr=gi});var yi,Pr,Rr=Z(()=>{ht();zr();yi=we("v5",80,Lr),Pr=yi});var Jr,Gr=Z(()=>{Jr="00000000-0000-0000-0000-000000000000"});function vi(e){if(!ne(e))throw TypeError("Invalid UUID");return parseInt(e.slice(14,15),16)}var Zr,Vr=Z(()=>{$e();Zr=vi});var Qr={};on(Qr,{NIL:()=>Jr,parse:()=>ke,stringify:()=>Tr,v1:()=>Yr,v3:()=>_r,v4:()=>Hr,v5:()=>Pr,validate:()=>ne,version:()=>Zr});var Xr=Z(()=>{Cr();Wr();Br();Rr();Gr();Vr();$e();be();dt()});var en=U((Wu,Kr)=>{var $i=(Xr(),cn(Qr)),bi=Kr.exports;bi.uuid=function(){return $i.v4()}});var nn=U((Fu,mt)=>{var{date:wi,duration:Oi}=Mt(),xi={math:Bt(),array:Me(),number:Kt(),url:tr(),string:ir(),comparison:gr(),object:Mr(),regex:kr(),uuid:en()},rn=["sortBy"];mt.exports.helpersToRemoveForJs=rn;var tn={date:wi,duration:Oi},ie;mt.exports.getJsHelperList=()=>{if(ie)return ie;ie={};for(let e of Object.values(xi))for(let[t,r]of Object.entries(e))ie[t]=(...n)=>r(...n,{});for(let e of Object.keys(tn))ie[e]=tn[e];for(let e of rn)delete ie[e];return Object.freeze(ie),ie}});var{getJsHelperList:Si}=nn(),Ni=Si(),Hu={...Ni,stripProtocol:helpersStripProtocol};export{Hu as default};
```
mike12345567, line 1: Same question should this be included or just built when needed?

packages/builder/src/components/integration/QueryViewer.svelte
```diff
@@ -143,8 +144,20 @@
   const handleScroll = e => {
     scrolling = e.target.scrollTop !== 0
   }
+
+  async function handleKeyDown(evt) {
```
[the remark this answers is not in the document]
adrinr: This is something that comes from @shogunpurple. Is this because of ironbank?

packages/string-templates/package.json
```diff
@@ -20,7 +21,7 @@
   ],
   "scripts": {
     "build": "tsc && rollup -c",
-    "dev": "tsc && rollup -cw",
+    "dev": "concurrently \"tsc --watch\" \"rollup -cw\"",
```
mike12345567, line 24: Nice fix - this was probably causing some problems I'd bet!

packages/server/src/jsRunner/bundles/bson.ivm.bundle.js
adrinr: As I mentioned before, we can either do this or do we need to have all the commands (build, test, etc) to build it first: https://github.com/Budibase/budibase/pull/12963

packages/builder/src/components/integration/QueryViewer.svelte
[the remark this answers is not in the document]
shogunpurple: It is yeah, I added it in my ironbank PR as it was a pet peeve and something that most other query based tools have. Happy to remove this if we want to build an abstraction, but tested it pretty extensively and it works fine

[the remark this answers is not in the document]
mike12345567: @shogunpurple ah fair enough - wouldn't remove it, its a pet peeve of mine as well, was just thinking could be worth looking at moving to a lower level to make it easier to utilise on other modals - wouldn't do it as part of this PR though!

Will tag @aptkingston just for visibility, could be worthwhile expanding on this!

packages/server/src/jsRunner/bundles/bson.ivm.bundle.js
mike12345567: ah sorry - I do remember seeing this before - it probably will never be re-built so probably better to get rid of build steps that won't actually do anything!

mike12345567 (approved):
LGTM!

PR #12939, opened 2024-02-02
A small UX improvement to the automation scripting system

mike12345567:
## Description
As shown during the lab day demos - moving from the old method of showing bindings that are available in the JS scripting block:
image

To something similar to that of our binding drawer experience, using the `BindingPicker` component.
image


aptkingston (commented):
LGTM!

Small suggested design change - could we bump up the binding list slightly? So that the back button is inline with the top of the code editor and the spacing around the search input is the same. Think it would look a little better!

This is what I mean:
image
image

mike12345567:
> LGTM!
> 
> Small suggested design change - could we bump up the binding list slightly? So that the back button is inline with the top of the code editor and the spacing around the search input is the same. Think it would look a little better!
> 
> This is what I mean: image image

Sorted!

image

aptkingston (approved):
LGTM! 

NAB for now, but for future - in general if there are styles needed for one specific use case (like specifically when using the binding picker here in the JS scripting modal) it's a bit cleaner to add those specific styles where the unique use case is rather than a new prop in the generic component, so in this case either adding a global style rule to target the binding picker and set padding top to 0 in this case, or add styles to the other container divs you have access to.

Basically just separation of concerns! This is totally fine to merge now though, but just something for the future.

PR #12943, opened 2024-02-02
Chore/fix worker request error message

melohagan:
## Description
We were doing a `await response.text()` within a try/catch which is not valid: https://github.com/node-fetch/node-fetch/issues/533

Fixed to check for response header first.

### DIFF
--- a/packages/server/src/utilities/workerRequests.ts
+++ b/packages/server/src/utilities/workerRequests.ts
@@ -1,5 +1,4 @@
-import fetch from "node-fetch"
-import { Response } from "node-fetch"
+import { Response, default as fetch } from "node-fetch"
 import env from "../environment"
 import { checkSlashesInUrl } from "./index"
 import {
--- a/packages/server/src/utilities/workerRequests.ts
+++ b/packages/server/src/utilities/workerRequests.ts
@@ -40,22 +40,21 @@ export function request(ctx?: Ctx, request?: any) {
 }
 
 async function checkResponse(
-  response: any,
+  response: Response,
   errorMsg: string,
   { ctx }: { ctx?: Ctx } = {}
 ) {
-  if (response.status !== 200) {
-    let error
-    try {
-      error = await response.json()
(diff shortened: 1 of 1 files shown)


packages/server/src/utilities/workerRequests.ts
```diff
@@ -40,25 +40,21 @@ export function request(ctx?: Ctx, request?: any) {
...
+    } else {
+      responseErrorMessage = await response.text()
     }
-    const msg = `Unable to ${errorMsg} - ${
-      error.message ? error.message : error
-    }`
+    const msg = `Unable to ${errorMsg} - ${responseErrorMessage}`
     if (ctx) {
-      ctx.throw(400, msg)
+      ctx.throw(msg, response.status)
```
samwho, line 57:
```suggestion
      ctx.throw(response.status, msg)
```

melohagan: It actually has to be the other way around to make use of the option status and default to 500:

```ts
/**
         * Throw an error with `msg` and optional `status`
         * defaulting to 500. Note that these are user-level
         * errors, and the message may be exposed to the client.
         *
         *    this.throw(403)
         *    this.throw('name required', 400)
         *    this.throw(400, 'name required')
         *    this.throw('something exploded')
         *    this.throw(new Error('invalid'), 400);
         *    this.throw(400, new Error('invalid'));
         *
         * See: https://github.com/jshttp/http-errors
         */
        throw(message: string, code?: number, properties?: {}): never;
```

samwho: That... is absolutely terrifying.

melohagan: I've never seen anything like it to be fair 😂 

Funny code comment as well: https://github.com/jshttp/http-errors/blob/206aa2c15635dc1212c06c279540972aa90e23ea/index.js#L50

samwho (approved):

PR #12953, opened 2024-02-05
Basic Postgres and Mongo query testcases.

samwho:
## Description

This PR creates a suite of tests for the queries functionality against MongoDB and Postgres. The MongoDB tests are a bit more fleshed out, but the PR is getting big so I'd like to get those in and then flesh out the Postgres tests more in a separate PR.

### DIFF
--- a/packages/server/src/api/routes/tests/queries/mongodb.spec.ts
+++ b/packages/server/src/api/routes/tests/queries/mongodb.spec.ts
@@ -4,7 +4,6 @@ import { databaseTestProviders } from "../../../../integrations/tests/utils"
 import { MongoClient } from "mongodb"
 
 jest.unmock("mongodb")
-jest.setTimeout(3000)
 
 describe("/queries", () => {
   let request = setup.getRequest()
--- a/packages/server/src/api/routes/tests/queries/mongodb.spec.ts
+++ b/packages/server/src/api/routes/tests/queries/mongodb.spec.ts
@@ -5,6 +5,8 @@ import { MongoClient, type Collection } from "mongodb"
 
 jest.unmock("mongodb")
 
+const collection = "test_collection"
+
 describe("/queries", () => {
   let config = setup.getConfig()
   let datasource: Datasource
@@ -20,7 +22,15 @@ describe("/queries", () => {
       transformer: "return data",
       readable: true,
     }
-    return await config.api.query.create({ ...defaultQuery, ...query })
+    const combinedQuery = { ...defaultQuery, ...query }
+    if (
+      combinedQuery.fields &&
+      combinedQuery.fields.extra &&
+      !combinedQuery.fields.extra.collection
+    ) {
+      combinedQuery.fields.extra.collection = collection
+    }
+    return await config.api.query.create(combinedQuery)
   }
 
   async function withClient(
@@ -37,7 +47,6 @@ describe("/queries", () => {
   }
 
   async function withCollection(
-    collection: string,
     callback: (collection: Collection) => Promise<void>
   ): Promise<void> {
     await withClient(async client => {
@@ -61,7 +70,7 @@ describe("/queries", () => {
   })
 
   beforeEach(async () => {
-    await withCollection("test_table", async collection => {
+    await withCollection(async collection => {
       await collection.insertMany([
         { name: "one" },
         { name: "two" },
@@ -73,18 +82,17 @@ describe("/queries", () => {
   })
 
   afterEach(async () => {
-    await withCollection("test_table", async collection => {
+    await withCollection(async collection => {
       await collection.drop()
     })
   })
 
-  it("should execute a query", async () => {
+  it("should execute a count query", async () => {
     const query = await createQuery({
       fields: {
-        json: "{}",
+        json: {},
         extra: {
           actionType: "count",
-          collection: "test_table",
         },
       },
     })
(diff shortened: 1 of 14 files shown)


packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -3,45 +3,45 @@ import { GenericContainer, Wait, StartedTestContainer } from "testcontainers"
...
-    const host = container.getHost()
-    const port = container.getMappedPort(5432)
+export async function start(): Promise<StartedTestContainer> {
+  if (!container) {
+    container = await new GenericContainer("postgres:16.1-bullseye")
+      .withExposedPorts(5432)
+      .withEnvironment({ POSTGRES_PASSWORD: "password" })
+      .withWaitStrategy(
+        Wait.forSuccessfulCommand(
+          "pg_isready -h localhost -p 5432"
```
adrinr, line 13: Good improvement :)

packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -3,45 +3,45 @@ import { GenericContainer, Wait, StartedTestContainer } from "testcontainers"
...
+        ).withStartupTimeout(10000)
```
adrinr, line 14: Do we need this timeout? Should we not rely on the jest timeout?

packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -3,45 +3,45 @@ import { GenericContainer, Wait, StartedTestContainer } from "testcontainers"
...
+      )
+      .start()
+  }
+  return container
+}
+
+export async function datasource(): Promise<Datasource> {
+  const container = await start()
```
adrinr, line 22: Should `if (!container)` not be in here, instead of inside the `start` func?

packages/server/src/tests/utilities/api/query.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  create = async (body: Query): Promise<Query> => {
+    const res = await this.request
+      .post(`/api/queries`)
+      .set(this.config.defaultHeaders())
+      .send(body)
+      .expect("Content-Type", /json/)
+
+    if (res.status !== 200) {
```
adrinr, line 17: Why this instead of using the expected status field as we do in the other utils?

packages/server/src/api/routes/tests/queries/mongodb.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      name: "New Query",
+      parameters: [],
+      fields: {},
+      schema: {},
+      queryVerb: "read",
+      transformer: "return data",
+      readable: true,
+    }
+
+    const res = await request
```
adrinr, line 25: This should be using the `config.api.queries` you just created

packages/server/src/integrations/tests/utils/mongodb.ts
```diff
@@ -0,0 +1,44 @@
...
+    container = await new GenericContainer("mongo:7.0-jammy")
+      .withExposedPorts(27017)
+      .withEnvironment({
+        MONGO_INITDB_ROOT_USERNAME: "mongo",
+        MONGO_INITDB_ROOT_PASSWORD: "password",
+      })
+      .withWaitStrategy(
+        Wait.forSuccessfulCommand(
+          `mongosh --eval "db.version()"`
+        ).withStartupTimeout(10000)
```
adrinr, line 17: Same comment about timeouts

adrinr (commented):
Nice approach, just some comments

packages/server/src/integrations/tests/utils/postgres.ts
samwho: I'm not sure, I copied what existed already. Happy to remove and see what happens.

packages/server/src/integrations/tests/utils/postgres.ts
samwho: Thank you ❤️

packages/server/src/integrations/tests/utils/postgres.ts
samwho: The changes in here are a bit of a relic of a previous attempt to get something working. `datasource()` and `start()` don't really need to be split out anymore, I can recombine them.

packages/server/src/tests/utilities/api/query.ts
samwho: The error message is nicer. If you use `.expect(200)` the error message doesn't show you the body of the failed request. It's been in the back of my mind for a while to try and improve this generally but I've not carved out the time yet. In a completely ideal world I'd like to even surface the error from inside the request, not catching it in any middleware, so it can be displayed in the test failure message.

packages/server/src/integrations/tests/utils/mongodb.ts
```diff
@@ -0,0 +1,41 @@
+import { Datasource, SourceName } from "@budibase/types"
+import { GenericContainer, Wait, StartedTestContainer } from "testcontainers"
+
+let container: StartedTestContainer | undefined
+
+export async function start(): Promise<StartedTestContainer> {
```
mike12345567, line 6: NAB: we should probably create a class/interface for this, as the hope is we expand this, would be nice to make sure this has a consistent interface/if we do build any generic tests all of these can be used under one interface easily.

mike12345567 (approved):
LGTM - one minor comment, not super important right now!

packages/server/src/integrations/tests/utils/mongodb.ts
samwho: There is one but it's a little disconnected in how it's implemented: https://github.com/Budibase/budibase/blob/90a6fef9577130fb4536ca7fb4c1470d3ddc7cf7/packages/server/src/integrations/tests/utils/index.ts#L10

PR #12963, opened 2024-02-06
Move bundle responsibility from string-templates to server

adrinr:
## Description
Moving responsibility from string-templates to server. Also, making the commands less interdependent.
All details described on the new README.md


samwho (approved):
This makes a lot more sense I agree. Really great work!

mike12345567 (approved):

Issue #5591, opened 2022-04-26
Make it clearer that the JSON field is the ROW Schema


melohagan:
**Describe the feature request**
Lots of people in discussions (and myself the first time I used queries) think that the JSON tab is showing the entire transformed payload. 
It needs to be made clearer that it is the JSON Schema for a row. 

Some possible suggestions:
- Rename the field to *Row JSON*
- Add a description label just above or below the tab to explain what is being shown. *E.g. Example row payload*
- Change the JSON payload to show the transformed Raw payload, and only the Schema shows the per row schema.
- Make the schema tab the first/default tab.  

**Screenshots**
![Screenshot 2022-04-26 at 14 01 09](https://user-images.githubusercontent.com/101575380/165305617-3ae2c549-e8bb-46fe-8fb9-765cca7406a3.png)

PR #12964, opened 2024-02-06
Fix BSON usage for isolated-vm

adrinr:
## Description
`Isolated-vm` is causing issues with Mongo queries, as the data comes as bson objects instead of js objects. Because of this, the responses don't know how to process the fields, returning things such as `{ "_id": {} }`.

Following the helpers approach, we inject the bson library and execute the following steps:
1. Serialise the data from bson to buffer outside the isolate.
2. Pass the buffer into the isolate.
3. Load the bson library into the isolate.
4. In the isolate, deserialise the buffer to bson. At this point, we will have the bson object, allowing the user script to use it's properties.
5. At the return of the script, we serialise and deserialise the result to json. This will cast the bson to json, returning primitive types that can be extracted from the isolate.

The bson library is only 50k, so the isolate creation increase should be negligible.

The main drawback is a drop to the performance, as the serialisation from and to bson is slow. Below there is some analytics on the impact:
| Rows returned | BSON serialisation | Without serialisation |
|--------|--------|--------|
| 1k | avg=473.3ms | avg=455.01ms |
| 10k | avg=900.74ms | avg=821.53ms |
| 50k | avg=2.97s | avg=2.82s | 

## Feature branch env
[Feature Branch Link](http://fb-fix-bson.fb.qa.budibase.net)


packages/server/src/integrations/mongodb.ts
```diff
@@ -356,8 +356,8 @@ class MongoIntegration implements IntegrationBase {
   constructor(config: MongoDBConfig) {
     this.config = config
     const options: MongoClientOptions = {
-      tlsCertificateFile: config.tlsCertificateFile || undefined,
-      tlsCertificateKeyFile: config.tlsCertificateKeyFile || undefined,
+      tlsCertificateKeyFile:
```
adrinr, line 359: This had been deprecated. TODO:

- [ ] Test properly

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
```diff
@@ -41,7 +41,7 @@
   </div>
   <div class="content">
     {#if activeTab === "JSON"}
-      <JSONPanel data={rows[0] || {}} />
```
adrinr, line 44: This was displaying only the first results on the display, instead of the actual result
image

packages/server/src/integrations/mongodb.ts
adrinr: Done here: https://github.com/Budibase/budibase/pull/12632

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
mike12345567: This was by design originally - it was supposed to show what a "row" would look like returned from the query. Printing the whole response could be really slow if you have 50K+ rows as theres no virtual rendering.

packages/server/src/jsRunner/bundles/bson.ivm.bundle.js
```diff
@@ -0,0 +1,7 @@
...(new e(t).id,this.id)}catch{return!1}}toBinary(){return new _(this.id,_.SUBTYPE_UUID)}static generate(){let t=h.randomBytes(gt);return t[6]=t[6]&15|64,t[8]=t[8]&63|128,t}static isValid(t){return t?typeof t=="string"?e.isValidUUIDString(t):Lt(t)?t.byteLength===gt:t._bsontype==="Binary"&&t.sub_type===this.SUBTYPE_UUID&&t.buffer.byteLength===16:!1}static createFromHexString(t){let n=e.bytesFromString(t);return new e(n)}static createFromBase64(t){return new e(h.fromBase64(t))}static bytesFromString(t){if(!e.isValidUUIDString(t))throw new o("UUID string representation must be 32 hex digits or canonical hyphenated representation");return h.fromHex(t.replace(/-/g,""))}static isValidUUIDString(t){return Le.test(t)||Re.test(t)}inspect(t,n,s){return s??=j,`new UUID(${s(this.toHexString(),n)})`}},Y=class e extends D{get _bsontype(){return"Code"}constructor(t,n){super(),this.code=t.toString(),this.scope=n??null}toJSON(){return this.scope!=null?{code:this.code,scope:this.scope}:{code:this.code}}toExtendedJSON(){return this.scope?{$code:this.code,$scope:this.scope}:{$code:this.code}}static fromExtendedJSON(t){return new e(t.$code,t.$scope)}inspect(t,n,s){s??=j;let r=s(this.code,n),i=r.includes(`
```
mike12345567, line 1: Should this be included? Would it not be better to have this built as needed and excluded from git?

packages/server/src/utilities/scriptRunner.ts
```diff
@@ -1,62 +1,111 @@
...
+      context: {
+        ...context,
+        data: parseBson
+          ? bson.BSON.serialize({ data: context.data })
+          : context.data,
+      },
+    })
+
+    if (parseBson) {
+      script = `return JSON.parse(JSON.stringify((function(){data=deserialize(data).data;${script}})()));`
```
mike12345567, line 25: Might be worth a comment here to explain whats going on - cause theres a lot at work here and explaining it could be valuable!

mike12345567 (changes requested):
Only major thing that needs done - probably need to revert the `QueryViewerSidePanel` change as it could be risky to display the whole response and it hasn't worked that way before.

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
melohagan: I did raise an issue nearly two years ago relating to this: https://github.com/Budibase/budibase/issues/5591

Closed it because it was stale, but still a thing I suppose.

packages/server/src/utilities/scriptRunner.ts
```diff
@@ -1,62 +1,111 @@
...
-    this.vm.code = code
+
+    this.vm.code = script
   }
 
   execute() {
     this.vm.runScript()
-    const results = this.vm.getValue("results")
-    return results.out
+    const result = this.vm.getResult()
```
melohagan, line 33: NAB: could make this inline: `return this.vm.getResult()`

packages/server/src/utilities/scriptRunner.ts
```diff
@@ -1,62 +1,111 @@
...
+    return result
   }
 }
 
 class IsolatedVM {
-  isolate: ivm.Isolate
-  vm: ivm.Context
+  #isolate: ivm.Isolate
```
melohagan, line 39: NAB: What are the hashtags for?

melohagan (approved):

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
adrinr: We should then trim the response, instead of showing the first row. Otherwise, it will display that the response is an object instead of an array

packages/server/src/jsRunner/bundles/bson.ivm.bundle.js
adrinr: It is the same approach as here: https://github.com/Budibase/budibase/pull/12963
More details on the readme.md

packages/server/src/utilities/scriptRunner.ts
adrinr: That is the way js marks private fields, ensuring it's not accessed from outside

packages/server/src/utilities/scriptRunner.ts
adrinr: On build time this will inlined, and having it as separate lines makes debugging much easier

packages/server/src/utilities/scriptRunner.ts
melohagan: Ah runtime private field. Nice 👍

melohagan (changes requested):
Found an issue on the feature branch with the `aggregate` pipeline:

Screenshot 2024-02-07 at 13 48 52

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
adrinr: I left all the reviews as they were before, except when the result returns multiple rows. In this case, we will show that is an array, but showing a max of 5 results followed by `...`
image

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
```diff
@@ -41,7 +41,7 @@
...
+      <JSONPanel data={rows?.length === 1 ? rows[0] : rows || {}} />
```
melohagan, line 44: Would it not be better to always return `rows` in this case instead of `rows[0]` if our intention is to show the full response, and not the first row?

adrinr: The issue is the way the responses are resolved. Most use cases are not actually arrays, but our code is parsing it to it. Doing this check we display the response as we did before in most of the cases, fe.
image

melohagan: Gotcha - makes sense!

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
adrinr: Done @melohagan 
image

melohagan: Screenshot 2024-02-08 at 13 46 10

Also noticed that the `... x further items` does not work recursively, so nested arrays that have more than five objects show all.

melohagan (approved):
Left a couple of minor additional comments, but otherwise looks good!

packages/server/src/jsRunner/vm/index.ts
```diff
@@ -127,6 +131,68 @@ export class IsolatedVM implements VM {
...
+  withParsingBson(data: any) {
+    this.addToContext({
+      bsonData: bson.BSON.serialize({ data }),
+    })
+
+    // If we need to parse bson, we follow the next steps:
+    // 1. Serialise the data from potential BSON to buffer before passing it to the isolate
+    // 2. Deserialise the data within the isolate, to get the original data
+    // 3. Process script
+    // 4. Stringify the result in order to convert the result from BSON to json
```
samwho, line 147: How come we don't repeat the process of serializing BSON inside the isolate and deserializing it ouside, to make the process truly transparent?

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
mike12345567: > Screenshot 2024-02-08 at 13 46 10
> 
> Also noticed that the `... x further items` does not work recursively, so nested arrays that have more than five objects show all.

I think recursively working this would be too complex - realistically if a user has an array within a row with hundreds of elements, we are going to have to display it no matter what, whether it was the old way or the new way of displaying this.

packages/server/src/jsRunner/vm/index.ts
adrinr: I did it this way because I don't see any specific benefit, and we don't need a post-process conversion, as the output will be already a valid JSON. If needed, we can extend the logic to handle it

packages/builder/src/components/integration/QueryViewerSidePanel/index.svelte
adrinr: Fixed!

melohagan: Looks great! 🥇

PR #12980, opened 2024-02-07
Fix for js binding auto complete

deanhannigan:
## Description

This issue affected the js autocomplete behaviour. It was attempting to destructure an undefined option. The value has now been defaulted to an empty object.

## Addresses
- https://linear.app/budibase/issue/BUDI-7991/auto-completion-in-bindings-is-broken


shogunpurple (approved):
LGTM

Issue #12853, opened 2024-01-24
Display charset does not match export charset -  Issue not fixed


raventechnologiesinc:
**Checklist**

I have searched budibase discussions and github issues to check if my issue already exists

[https://github.com/Budibase/budibase/projects/22#card-76344128](https://github.com/Budibase/budibase/projects/22#card-76344128)

**Hosting**
reventech.budibase.app -->

**Describe the bug**
I've tried every conversion using vscode and iconv...no luck to get the exported data to the charset that BB is using to display on web. Please help with any ideas

BB displays matches mariaDB charset
meta charset = utf-8 on header inspection

BB exports an unidentifiable charset (maybe WIN1252) which can't be converted to native charset

[datasource_plus_a20094ed6b4b4b8c9b0ad54bf1fb81eb__CustomerBill.csv](https://github.com/Budibase/budibase/files/14034210/datasource_plus_a20094ed6b4b4b8c9b0ad54bf1fb81eb__CustomerBill.csv)

MySQL export is correct (Native charset)
ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

[CustomerBill.csv](https://github.com/Budibase/budibase/files/14034243/CustomerBill.csv)

**Desktop (please complete the following information):**

* OS WIn11
* Browser Chrome
* Version 2.15.5

**Additional context**
charset correctly displayed

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/35a79ad5-dc1b-4686-aa35-ceb864d27dbc/85f0b895-2143-4310-a1cc-6991f14f6767?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8zNWE3OWFkNS1kYzFiLTQ2ODYtYWEzNS1jZWI4NjRkMjdkYmMvODVmMGI4OTUtMjE0My00MzEwLWExY2MtNjk5MWYxNGY2NzY3IiwiaWF0IjoxNzA2MDgyMzE3LCJleHAiOjMzMjc2NjQyMzE3fQ.ERjKxvsPBcFY0_ZUBGb3NLfua98ZRBXWoFE2jRMmAAw)

exported to CSV sharset is corrupted

[...]

PR #12990, opened 2024-02-08
Download CSV with UTF-8 support

melohagan:
## Description
The **Export Data** action was not downloading CSVs with correct UTF-8 characters.
I noticed that this did work in the **Data** section, so I simply copied the *Blob* approach that I found there and it works.

## Addresses
- https://github.com/Budibase/budibase/issues/12853


deanhannigan (approved):
LGTM

Issue #1212, opened 2021-02-26
N8N Integration


shogunpurple:
I think, adding n8n.io integration would probably facilitate most of the use-cases mentioned above!

_Originally posted by @arladmin in https://github.com/Budibase/budibase/discussions/891#discussioncomment-381459_

melohagan:
Docs: https://docs.budibase.com/docs/n8n

PR #12992, opened 2024-02-08
n8n automation action integration 

melohagan:
## Description
n8n has been added as an automation action. Supports different http methods, and the 'Authorization' header.

## Addresses
- https://linear.app/budibase/issue/BUDI-4702/n8nio-integration
- https://github.com/Budibase/budibase/issues/1212

## Screenshots
Screenshot 2024-02-08 at 15 45 10

Screenshot 2024-02-08 at 15 44 30

Screenshot 2024-02-08 at 15 44 09

Screenshot 2024-02-08 at 15 45 41


packages/server/src/automations/steps/make.ts
```diff
@@ -34,28 +34,8 @@ export const definition: AutomationStepSchema = {
...
-        value4: {
-          type: AutomationIOType.STRING,
-          title: "Input Value 4",
-        },
-        value5: {
-          type: AutomationIOType.STRING,
-          title: "Input Value 5",
-        },
       },
-      required: ["url", "value1", "value2", "value3", "value4", "value5"],
```
samwho, line 58: How come these're being removed in this PR?

packages/server/src/automations/tests/n8n.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  beforeAll(async () => {
+    await config.init()
+  })
+
+  afterAll()
+
+  it("should be able to run the action and default to 'get'", async () => {
+    const res = await runStep(actions.n8n.stepId, {
+      url: "http://www.test.com",
```
samwho, line 14: test.com is a privately owned domain, I tend to avoid it in tests just in case. It could theoretically point anywhere and if we do accidentally hit it there's no telling what it'll do.

I prefer to use example.com, which is owned by IANA and is a specifically reserved domain: https://www.rfc-editor.org/rfc/rfc2606.

samwho (changes requested):
This automation appears to make a pretty naked `fetch` call, and doesn't appear to have anything n8n-specific in it. Could we not rebrand this as a generic HTTP call step?

packages/server/src/automations/steps/make.ts
melohagan: They were supposed to be fully deprecated in a previous PR and I missed these props at the time. Only noticed today!

https://github.com/Budibase/budibase/commit/b4e0d98973c6395e36040f7789fb2f7995781f10

shogunpurple (commented):
Would agree with @samwho here. It's just duplicated code to make an HTTP call to N8N.  I'd say we:
- either maintain this as a plugin and add a tutorial in the docs
- Add something to the docs to show you how to use our REST connector to do it

EDIT: we actually already do this for make and zapier. It's basically the same thing so no argument if we merge this too. We don't have a direct REST step, so you still have to set up a REST datasource in `data`, point it at N8N and then use it in automations. It saves you a few steps if all you want to do is trigger the webhook I guess.

packages/server/src/automations/steps/n8n.ts
```diff
@@ -0,0 +1,143 @@
...
+import {
+  AutomationActionStepId,
+  AutomationStepSchema,
+  AutomationStepInput,
+  AutomationStepType,
+  AutomationIOType,
+  AutomationFeature,
+} from "@budibase/types"
+
+enum Method {
```
mike12345567, line 12: Could move these into `@budibase/types` and re-use them in the REST integration, just incase these pop up in other places would want to keep from re-writing enums like this all over the place.

melohagan: Added a HttpMethod enum to types, and updated the REST connector.

joebudi:
@melohagan nice work. Is Authorization an optional setting? If so, we should make that clear.

melohagan:
> @melohagan nice work. Is Authorization an optional setting? If so, we should make that clear.

It is optional, and n8n will return a 401 status and a message requesting the auth header if it's needed.

How would you expect required/optional fields to be presented in automations? This would apply to all automation action steps. E.g. bold labels, or an asterisk beside the label

joebudi:
> > @melohagan nice work. Is Authorization an optional setting? If so, we should make that clear.
> 
> It is optional, and n8n will return a 401 status and a message requesting the auth header if it's needed.
> 
> How would you expect required/optional fields to be presented in automations? This would apply to all automation action steps. E.g. bold labels, or an asterisk beside the label

@AndyWhann is best to answer this.

samwho (approved):
I'm broadly happy with this change, though I would wait for an OK from Marty or Joe before going ahead and merging.

PR #12993, opened 2024-02-08
Replace uses of test.com with example.com

samwho:
## Description

The domain `test.com` is privately owned and could, in theory, point anywhere. If we accidentally were to issue requests to it from our tests it's unclear what might happen. It's much safer to use `example.com`, which is owned by IANA and is part of the RFC2606 reserved domains.

This PR replaces all uses of `test.com` with `example.com` and introduces a lint rule to prevent future uses of `test.com`.

Related PR: https://github.com/Budibase/budibase-pro/pull/262

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -43,7 +43,8 @@
         "no-useless-escape": "off",
         "no-undef": "off",
         "no-prototype-builtins": "off",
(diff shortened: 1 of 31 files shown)


mike12345567 (approved):
LGTM

PR #12994, opened 2024-02-08
Remove vm2 from stringtemplates

adrinr:
## Description
Removing vm2 from string-templates. This is not used anymore, only for tests, that will use vm for now

## Feature branch env
[Feature Branch Link](http://fb-remove-vm2-from-stringtemplates.fb.qa.budibase.net)


packages/string-templates/test/javascript.spec.js
```diff
@@ -122,13 +122,6 @@ describe("Test the JavaScript helper", () => {
     const output = processJS(`return process`)
     expect(output).toBe("Error while executing JS")
   })
-
-  it("should prevent sandbox escape", () => {
```
adrinr, line 126: This test depends on the implementation. vm2 would crash, but VM does not. Probably it's better to remove it for now

mike12345567: I would probably keep this test but `x.it` it out, we definitely want to confirm stuff like this on the backend, perhaps we should just move this test to the server?

mike12345567 (approved):
LGTM

packages/string-templates/src/index.js
```diff
@@ -1,398 +1,40 @@
...
-  if (opts.noEscaping) {
-    string = exports.disableEscaping(string)
-  }
-
-  // This does not throw an error when template can't be fulfilled,
-  // have to try correct beforehand
-  const instance = opts.noHelpers ? hbsInstanceNoHelpers : hbsInstance
-  const template = instance.compile(string, {
-    strict: false,
+import vm from "vm"
```
mike12345567, line 1: This will likely break things in Node - might need to require these, hopefully the rollup will convert them back to imports as needed.

packages/string-templates/src/index.js
```diff
@@ -1,398 +1,40 @@
...
+import templates from "./functions.js"
+
+/**
+ * ES6 entrypoint for rollup
+ */
+export const isValid = templates.isValid
```
mike12345567, line 7: Not sure we can do this in a node entry point either - might still need to call this file `.mjs` for node to load it correctly?

adrinr: Good point, moved to the server tests for now!

shogunpurple (approved):
LGTM

PR #12996, opened 2024-02-09
Make import/export fs calls as async as possible.

samwho:
## Description

Lots of `*Sync()` calls in the backups import/export code, so I'm replacing it with promisified `fs/promise` calls instead. Trying to make the impact of the CPU spikes we see when backups are running as little as possible.


mike12345567 (approved):
Gave this a test locally with Geis' big app - seems to be working fine, LGTM!

PR #12997, opened 2024-02-09
Add instrumentation into processStringSync.

samwho:
## Description

Still seeing many examples of `processStringSync` calls that look like this:

CleanShot 2024-02-09 at 11 43 15@2x

The teeny tiny block at the end of this large `processStringSync` is a `runJS` block, which doesn't track for me. How can we spent so much time processing the string, and so little time running JS? I'd like to know, so I'm adding more tracing.

### DIFF
--- a/packages/string-templates/package.json
+++ b/packages/string-templates/package.json
@@ -30,7 +30,8 @@
     "dayjs": "^1.10.8",
     "handlebars": "^4.7.6",
     "lodash.clonedeep": "^4.5.0",
-    "vm2": "^3.9.19"
+    "vm2": "^3.9.19",
+    "dd-trace": "4.20.0"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^17.1.0",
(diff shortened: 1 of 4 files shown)


mike12345567 (approved):

aptkingston:
It's probably long gone by now, but I did a deep dive here too a while back after realising that so much time is consumed processing string templates. I tried caching the templates too, like I see you're doing here, and I'm fairly certain there was a large speed improvement. Considering the memory cost is fairly trivial (and typical use cases, i.e. client apps, tend to reuse them a lot), I'd be in favour of just doing that all the time.

samwho:
@aptkingston I was wondering about the caching. This PR was left as a draft because it's getting weird build errors that I couldn't decipher. If you're feeling generous, I'd love someone who knows the JS build ecosystem better than me to take a look 🙏

samwho:
Though caching being a boundless, non-time-limited object worries me. I'd want to have some sort of LRU cache in place so we know it can't grow indefinitely.

Issue #12973, opened 2024-02-06
Text field values do not persist unless changed


ecmonsen:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 2.17.3
  - App Version: 

**Describe the bug**
In a simple form with one text field that has a default value, and a button, button actions find a blank value for the text field, instead of the default value.

**To Reproduce**
1. Create new screen
2. Add a Form
3. Add Text Field inside the form, with Field = `my_value` and Default value=`1`
4. Add a Button
5. Edit the button's actions, and add a "Show Notification" action
6. Set the notification Message text to `Field value = '{{ Form 1.Fields.my_value }}'. Form value = '{{ Form 1.Value }}'`
7. Save & click Preview
8. Click `New Button`
9. A notification is shown with text `Field value = ''. Form value = '{}'`

**Expected behavior**
9. A notification is shown with text `Field value = '1'. Form value = '{"my_value":"1"}'`
10. Change the value of the text field to `2` and click `New Button`. Now you see the expected result, `Field value = '2'. Form value = '{"my_value":"2"}'`

**Screenshots**
![Screenshot 2024-02-06 at 9 08 02 AM](https://github.com/Budibase/budibase/assets/851923/f62d5153-d2ed-4f0d-a13a-ba284b5da8ba)
![Screenshot 2024-02-06 at 9 08 07 AM](https://github.com/Budibase/budibase/assets/851923/0f3f6dc0-d709-4711-8482-de2c56a29569)
![Screenshot 2024-02-06 at 9 08 13 AM](https://github.com/Budibase/budibase/assets/851923/f37054af-b704-46fd-8cb4-d584958b8370)
![Screenshot 2024-02-06 at 9 08 21 AM](https://github.com/Budibase/budibase/assets/851923/5e7b5c6f-b868-4632-af61-0362b64e3055)
![Screenshot 2024-02-06 at 9 08 30 AM](https://github.com/Budibase/budibase/assets/851923/24f7298c-8b02-4abd-91f4-2ad2578b5ff0)

  
**App Export**

[Debug-export-1707239887630.tar.gz](https://github.com/Budibase/budibase/files/14184042/Debug-export-1707239887630.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS Sonoma 14.2
 - Browser: Firefox
Budibase diagnostics:
```
{
  "browser": {
[...]

Issue #12857, opened 2024-01-24
Can't access apps with spaces in their names with a creator user account


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase 2.15.5 
  - App Version: 2.15.5 
- Cloud
  - Issue occurs on cloud as well.

**Describe the bug**
When assigning a creator to an app that has a space in it. You won't be able to access the app. Instead it will open a new tab and direct you to a page. Thats says;
```
{"message":"AppId could not be found","status":500}
```
**To Reproduce**
1. Go to your Budibase instance
2. Add a new user with the creator permissions
3. Assign them to an app that has a space in its name
4. Login to that user with access to the app
5. Click on the app
6. See new window opens with the error instead of being directed to the app screen.

**Expected behavior**
The app should be available when clicking on the specific app you wish to view. This must be related to the space between words in the title as other apps that don't have a space can be opened. 

**Screenshots**
<img width="1279" alt="Screenshot 2024-01-24 at 13 01 05" src="https://github.com/Budibase/budibase/assets/126772285/78c0543b-4e68-4b54-8596-9a1b9f22ffd7">
<img width="1284" alt="Screenshot 2024-01-24 at 13 01 11" src="https://github.com/Budibase/budibase/assets/126772285/23fd3365-20c6-4f26-b41e-c1ce815ea14f">

**Additional Notes**
If you click to access the app via the side panel on the left it will open the app correctly.

PR #13000, opened 2024-02-09
Various fixes

aptkingston:
WIP: adding more fixes as I go

## Description
- Fix creator users being unable to view apps that they have non-creator access to. Previously this threw an error due to generating an invalid URL. Now published apps can be viewed as intended by clicking the view button, or the whole row:
image
Unpublished apps cannot be clicked at all, as if you only have user-level access and the app is unpublished, then there is nothing for you to use. This is differentiated by the lack of a "View" button, and removal of hover styles on the list item:
image
- Fix certain form bindings not working as expected when using default values. The core issue here was actually deeper and more generic - there was a bit of a race condition in that context changes could happen (and be missed) in between components mounting and them starting to observe context changes. I've moved things around so that there is no longer any potential for context changes to be missed.

## Addresses
- https://github.com/Budibase/budibase/issues/12857
- https://github.com/Budibase/budibase/issues/12973


deanhannigan (approved):
LGTM

Issue #12114, opened 2023-10-19
[BUDI-7649] "Clear Row Selection" from Table Block after Button Click


Doctor-N0:
Similar to the "Clear Form" function, a "Clear Row Selection" function in the Table Block could also be helpful. The function could be added e.g. in the form of a check box "Clear Table Block Selection" under the field "Button". (see screenshot)

As discussed in the articles on Bulk selection (see [here ](https://github.com/Budibase/budibase/discussions/11055#discussioncomment-6370952) and [here](https://github.com/Budibase/budibase/discussions/12111)), this function would be quite helpful for these kind of applications.

![image](https://github.com/Budibase/budibase/assets/105864309/877315e8-6331-411c-9bf5-e660f6f6d7b1)

<sub>[BUDI-7649](https://linear.app/budibase/issue/BUDI-7649/clear-row-selection-from-table-block-after-button-click)</sub>

melohagan:
It would be quite nice to have a **Clear Row Selection** action.

PR #13002, opened 2024-02-09
Add Clear Row Selection action

melohagan:
## Description
As a user, I wanted the option to clear the row selection on button click. Useful when opening side panels as well.

## Addresses
- https://github.com/Budibase/budibase/issues/12114

## Screenshots
Screenshot 2024-02-09 at 14 52 00

Screenshot 2024-02-09 at 14 52 15
*Button click*

Screenshot 2024-02-09 at 14 57 50
*On row click*

**Result**
selected_rows


packages/client/src/components/app/table/Table.svelte
```diff
@@ -46,6 +46,13 @@
       selectedRows.map(row => row._id)
     )
   }
+  // Clear row selection if action was trigger
+  $: {
+    if ($rowSelectionStore?.[$component.id]?.clearSelectedRows) {
```
melohagan, line 51: I did this because the `$rowSelectionStore` was not being used directly to keep track of the selected rows, and I assumed there was a good reason for that.

melohagan: Not needed, refactoring.

melohagan:
> I'm guessing this PR could be rewritten now that we're moving to providing selected rows via the new global binding? This feature could then easily be accomplished by the `Table` and `TableBlock` providing the `Clear Row Selection` action, which simply just sets its local value of `selectedRows` to an empty array.
> 
> We wouldn't need a custom action written - it could just be accomplished using the existing `contextActionHandler`. There are a bunch of actions that follow this pattern today like `Scroll To Field`, `Validate Form` and `Clear Form` to name a few.

Refactored to use the `contextActionHandler` approach - much nicer!

aptkingston (approved):
Perfect! Good job 👌

PR #13003, opened 2024-02-09
Run scripts via vm/vm2

adrinr:
## Description
We found some breaking changes around https://github.com/Budibase/budibase/pull/12938.
Isolated VM requires strict js, but the previously used vm2 did not required it. Instead of rolling back the full PR, we are using the old code for running the scripts until we find a way to fix it

### DIFF
--- a/packages/server/src/environment.ts
+++ b/packages/server/src/environment.ts
@@ -72,9 +72,9 @@ const environment = {
   HTTP_MB_LIMIT: process.env.HTTP_MB_LIMIT,
   FORKED_PROCESS_NAME: process.env.FORKED_PROCESS_NAME || "main",
   JS_PER_INVOCATION_TIMEOUT_MS:
-    parseIntSafe(process.env.JS_PER_INVOCATION_TIMEOUT_MS) || 1000,
+    parseIntSafe(process.env.JS_PER_EXECUTION_TIME_LIMIT_MS) || 1000,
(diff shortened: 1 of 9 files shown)


packages/server/src/environment.ts
```diff
@@ -72,9 +72,9 @@ const environment = {
   HTTP_MB_LIMIT: process.env.HTTP_MB_LIMIT,
   FORKED_PROCESS_NAME: process.env.FORKED_PROCESS_NAME || "main",
   JS_PER_INVOCATION_TIMEOUT_MS:
-    parseIntSafe(process.env.JS_PER_INVOCATION_TIMEOUT_MS) || 1000,
+    parseIntSafe(process.env.JS_PER_EXECUTION_TIME_LIMIT_MS) || 1000,
   JS_PER_REQUEST_TIMEOUT_MS: parseIntSafe(
-    process.env.JS_PER_REQUEST_TIMEOUT_MS
+    process.env.JS_PER_REQUEST_TIME_LIMIT_MS
```
samwho, line 77: Not sure I understand why these are changing. Did we rename them with the release of the isolated-vm changes?

adrinr: Yes, that was caused by a wrong renaming. Changing it back to what it was

samwho (approved):

PR #13004, opened 2024-02-09
Fix for custom branding not appearing on login

PClmnt:
## Description
Fixes an issue where custom branding wasn't being showing for users. Most of the logic is the account-portal (https://github.com/Budibase/account-portal/pull/527). This PR is mostly just the storing of the cookie when the users goes to their tenant page i.e `peter.budibase.app`.

The one drawback to this is that if the user goes straight to `account.budibase.app` they won't have the cookie and therefore won't see the branding. 
## Addresses
[BUDI-7962
](https://linear.app/budibase/issue/BUDI-7962/allow-custom-login-to-be-show-as-default-for-cloud-when-a-premium-user)

## Feature branch env
[Feature Branch Link](http://fb-fix-custom-branding-issue.fb.qa.budibase.net)


mike12345567 (approved):

packages/builder/src/pages/builder/auth/_layout.svelte
```diff
@@ -14,7 +14,8 @@
     $admin.cloud &&
     !$admin.disableAccountPortal &&
     $admin.accountPortalUrl &&
-    !$admin?.checklist?.sso?.checked
+    !$admin?.checklist?.sso?.checked &&
+    !$admin.checklist.branding
```
melohagan, line 18: This makes me nervous. Should it be `$admin?.checklist?.branding` like the line above, or is the line above incorrect?

melohagan (changes requested):

PClmnt:
The more I thought about this, the more I think it's missing something. I talked with @deanhannigan  on Friday and we agreed this is the best approach. But in this case, if a user doesn't have their own custom SSO setup they won't be able to use it at all? i.e I think what [BUDI-7962
](https://linear.app/budibase/issue/BUDI-7962/allow-custom-login-to-be-show-as-default-for-cloud-when-a-premium-user)
 should really be is a ticket so support the use of branding on the account portal login page. Because this way users that were previously able to login via Microsoft or Google will not be able to when branding is set, without configuring custom SSO

packages/builder/src/pages/builder/auth/_layout.svelte
```diff
@@ -9,6 +10,16 @@
     $redirect("../")
   }
 
+  if ($admin?.checklist?.branding) {
+    let url = new URL(window.location.href)
+    let hostname = url.hostname
+    let parts = hostname.split(".")
+    let tenantId = parts[0]
+    let domain = parts.slice(-2).join(".")
+    console.log(domain)
```
mike12345567, line 19: Extra console log! Not sure why this wasn't caught by linter

mike12345567 (approved):
LGTM - one console log needing removed.

melohagan (approved):

PR #13005, opened 2024-02-09
Frontend + Backend CRON validation

mike12345567:
## Description
Adding some basic CRON validation to publishing, currently the error is not explained if it is hit - still need some frontend for this, but this now means that an error is provided to users when attempting to publish, and we can re-use this validation in the automation UI. Need to have both backend and frontend validation as invalid CRONs will already exist, backend makes sure these are error'd on.

Partially addresses: https://linear.app/budibase/issue/BUDI-7928/cron-expressions-should-be-validated

Error message:
image

Frontend: 
image

- Added the error message inline (Peter)
- Also fixed an issue with preset being null for custom crons

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
@@ -8,6 +8,8 @@
   export let value
   let error
 
+  const reboot = "@reboot"
+
   $: {
     const exists = CRON_EXPRESSIONS.some(cron => cron.value === value)
     const customIndex = CRON_EXPRESSIONS.findIndex(
@@ -22,7 +24,9 @@
   }
 
   const onChange = e => {
-    error = helpers.cron.validate(e.detail).err
+    if (value !== reboot) {
+      error = helpers.cron.validate(e.detail).err
+    }
     if (e.detail === value || error) {
       return
     }
@@ -57,7 +61,7 @@
     if (!$flags.cloud) {
       CRON_EXPRESSIONS.push({
         label: "Every Budibase Reboot",
-        value: "@reboot",
+        value: reboot,
       })
     }
   })
--- a/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
@@ -3,6 +3,7 @@
   import { onMount, createEventDispatcher } from "svelte"
   import { flags } from "stores/backend"
   import { helpers } from "@budibase/shared-core"
+
   const dispatch = createEventDispatcher()
 
   export let value
--- a/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
@@ -2,15 +2,31 @@
   import { Button, Select, Input, Label } from "@budibase/bbui"
   import { onMount, createEventDispatcher } from "svelte"
   import { flags } from "stores/backend"
-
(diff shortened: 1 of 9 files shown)


packages/shared-core/src/helpers/cron.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function improveErrors(errors: string[]): string[] {
+  const finalErrors: string[] = []
+
+  for (let error of errors) {
+    if (error.includes(INPUT_CRON_START)) {
+      error = error.split(INPUT_CRON_START)[0].trim()
+    }
+    for (let [key, value] of Object.entries(ERROR_SWAPS)) {
```
melohagan, line 23: NAB: could rename `key, value` to `old, new` to make the code a little easier to follow.

melohagan (approved):
LGTM - nice enhancement 😄

packages/shared-core/src/helpers/cron.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    for (let [oldErr, newErr] of Object.entries(ERROR_SWAPS)) {
+      if (error.includes(oldErr)) {
+        error = error.replace(new RegExp(oldErr, "g"), newErr)
+      }
+    }
+    finalErrors.push(error)
+  }
+  return finalErrors
+}
```
samwho, line 31: Eep. This concerns me. We're very sensitive to the `cron-validate` library changing their wording slightly and breaking our error messages. Makes me wonder if cron validation is actually complex enough to warrant using a library and munging its error messages.

mike12345567: If we lock the library version then this shouldn't be too much of a problem - I doubt we will need to update this often - the error messages themselves are quite hard to read so just wanted to improve them ever so slightly.

packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
```diff
@@ -2,15 +2,36 @@
...
   import { onMount, createEventDispatcher } from "svelte"
   import { flags } from "stores/backend"
+  import { helpers } from "@budibase/shared-core"
 
   const dispatch = createEventDispatcher()
 
   export let value
+  let error
+
+  const reboot = "@reboot"
```
mike12345567, line 12: We should probably have this as a constant in `@budibase/types` so that it can be shared through shared-core with the frontend and backend - there is already a string in the backend for this as well in `packages/server/src/automations/utils.ts`

mike12345567:
✅ the frontend side of it!

packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
```diff
@@ -2,15 +2,36 @@
...
+      CRON_EXPRESSIONS.splice(customIndex, 1)
+    }
+  }
 
   const onChange = e => {
-    if (e.detail === value) {
+    if (value !== reboot) {
+      error = helpers.cron.validate(e.detail).err
+    }
+    if (e.detail === value || error) {
```
melohagan, line 31: NAB: Just wondering what this line does? Is it not possible to have both **value** and **error** set?

PClmnt: If the cron value has not changed (e.detail === value) or there is an error, then we simply return early from the function

melohagan: Ah yes, thank you!

I was reading it like: `e.detail === (value || error)` for some reason.

PClmnt (approved):
Backend LGTM!

PR #13006, opened 2024-02-09
FIX: Error when deleting selected rows that have attachment

melohagan:
## Description
Table (and Table block) rows that have attachment columns can now be deleted via the new **Selected rows** binding which can be found under those settings.

The old binding has been updated to read *Selected Row IDs (DEPRECATED)* to indicate that it should no longer be used. It can be deleted in five minor versions time as per support policy.

## Addresses
- https://linear.app/budibase/issue/BUDI-7996/error-when-deleting-selected-rows-that-have-attachment-columns

## Screenshots

**BEFORE**
old binding

old delete

**AFTER**
deprecated label
*The old binding will work as before, but the label has been updated to warn the user*

new binding
*New selected rows binding is now found under the Table block/Table bindings*

full bindings
*Full row data is now provided, not just the _ids*

delete rows
*Delete rows bug now fixed with the new binding*

### DIFF
--- a/packages/builder/src/builderStore/dataBinding.js
+++ b/packages/builder/src/builderStore/dataBinding.js
@@ -617,6 +617,7 @@ const getDeviceBindings = () => {
 /**
  * Gets all selected rows bindings for tables in the current asset.
  * TODO: remove in future because we don't need a separate store for this
+ * DEPRECATED
  */
 const getSelectedRowsBindings = asset => {
   let bindings = []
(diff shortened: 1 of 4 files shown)


deanhannigan (approved):
LGTM

aptkingston (approved):
I'd just change the all uppercase `DEPRECATED` to just `deprecated` to look less intimidating, but otherwise LGTM!

PR #13011, opened 2024-02-12
[BUG] Don't drop views when updating its table configuration

adrinr:
## Description
We recently introduced a bug that, on table update, its views are dropped. This PR fixes it and adds tests around it

## Addresses
- [BUDI-7784 - [BUG] When adding a new column to a table with views, those views are deleted](https://linear.app/budibase/issue/BUDI-7784/[bug]-when-adding-a-new-column-to-a-table-with-views-those-views-are)

### DIFF
--- a/packages/server/src/api/controllers/table/internal.ts
+++ b/packages/server/src/api/controllers/table/internal.ts
@@ -18,12 +18,16 @@ export async function save(ctx: UserCtx<SaveTableRequest, SaveTableResponse>) {
     _rename?: RenameColumn
   } = {
     _id: generateTableID(),
-    views: {},
     ...rest,
     // Ensure these fields are populated, even if not sent in the request
     type: rest.type || "table",
     sourceType: rest.sourceType || TableSourceType.INTERNAL,
   }
+
+  if (!tableToSave.views) {
+    tableToSave.views = {}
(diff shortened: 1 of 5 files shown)


packages/server/src/api/controllers/table/internal.ts
```diff
@@ -18,10 +18,11 @@ export async function save(ctx: UserCtx<SaveTableRequest, SaveTableResponse>) {
     _rename?: RenameColumn
   } = {
     _id: generateTableID(),
-    ...rest,
-    type: "table",
-    sourceType: TableSourceType.INTERNAL,
     views: {},
```
mike12345567, line 21: it would probably be better to move this outside of the block, as if it is moved after `...rest` then it'll get overwritten, we should probably do something like 
```
if (!tableToSave.views) {
  tableToSave.views = {}
}
```

mike12345567 (approved):
LGTM - just one minor NAB comment.

packages/server/src/api/controllers/table/internal.ts
adrinr: If this was moved again, the tests should catch it. But I can move it out, to make it more explicit

PR #13012, opened 2024-02-12
Add environment variable to log JS errors in bindings

mike12345567:
## Description
As title - a utility for services to log errors which occur in the backend as part of Javascript bindings.


adrinr (approved):
LGTM! ❤️

PR #13026, opened 2024-02-13
Update CouchDB image to bookworm and couchdb 3.3.3.

samwho:
## Description

When running the single image on ARM CPUs, we found we'd get the following error when using `isolated-vm`:

```
Error: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /node_modules/isolated-vm/out/isolated_vm.node)
```

This is because `couchdb:3.2.1`, the image that the single image builds from, uses Debian Bullseye, which ships with glibc 2.31. Debian Bookworm ships with 2.36, which is the Debian that `node:20-slim` is based off of.

However, in order to get this to work I've had to update CouchDB from 3.2.1 to 3.3.3. We will need to test carefully to make sure that this is a safe upgrade for all of our users.

## Addresses
- https://linear.app/budibase/issue/BUDI-7997/fix-dlopen-error-in-arm-single-image-build-when-using-isolated-vm


hosting/couchdb/Dockerfile
```diff
@@ -1,14 +1,105 @@
-FROM couchdb:3.2.1
+# Modified from https://github.com/apache/couchdb-docker/blob/main/3.2.1/Dockerfile
+FROM node:20-slim
+
+# Add CouchDB user account to make sure the IDs are assigned consistently
+RUN groupadd -g 5984 -r couchdb && useradd -u 5984 -d /opt/couchdb -g couchdb couchdb
```
mike12345567, line 5: I'm wondering if there is someway we can pull the CouchDB image Dockerfile and switch its base dynamically, then use this as a local base image for our Couch image, would mean that we don't need to worry about keeping this base up to date - would just mean a funky `sed` to replace the from in the actual base image - could be worth a try to save some future pain?

hosting/single/runner.sh
```diff
@@ -97,10 +97,12 @@ fi
 sleep 10
 
 pushd app
-pm2 start -l /dev/stdout --name app "yarn run:docker"
+pm2 start --name app "yarn run:docker"
```
mike12345567, line 100: What does this change - does this mean that the pm2 logs will come out of the main docker logging?

samwho: As far as I can tell this flag was doing nothing.

hosting/couchdb/Dockerfile
samwho: The amount of magic and hacks we're layering on top of each other to get things to work is worrying me. The `runner.sh` script already changes a lot about the configuration of this image depending on the environment it's running in. Dynamically pulling a Dockerfile and modifying it is a step too far for me, I think.

mike12345567: > The amount of magic and hacks we're layering on top of each other to get things to work is worrying me. The `runner.sh` script already changes a lot about the configuration of this image depending on the environment it's running in. Dynamically pulling a Dockerfile and modifying it is a step too far for me, I think.

Hmm - my main concern is that we're depending an awful lot on there being no issues with the snapshot of the image we've used to create this - I agree there is a fair amount of magic, but it would be good to at least separate what is the Couch image from what is our image - giving us a little bit more control/knowledge of what is ours and what we would need to update in future.

Even if we were to split out the contents into its own Dockerfile, and add in some comments to explain where it is from that would help.

mike12345567: Yeah - fair enough, this will have to do - we don't update this image too often anyway!

mike12345567 (approved):

samwho:
@mike12345567 it occurs to me here we'll need to publish `budibase/couchdb:3.3.3` and then have a follow-up PR that updates the self-host `docker-compose.yaml` and Helm chart.

hosting/couchdb/Dockerfile
```diff
@@ -1,14 +1,111 @@
-FROM couchdb:3.2.1
+# Modified from https://github.com/apache/couchdb-docker/blob/main/3.3.3/Dockerfile
```
adrinr, line 1: I thought we said 3.3.3 was compatible. Why do we need the modifications?

samwho: The `couchdb:3.3.3` base image still uses bullseye. We need bookworm, so we've had to copy the contents of the 3.3.3 Dockerfile but base it on `node:20-slim` (chosen because it's bookworm and also the base image we use elsewhere, so introduces a little bit of consistency).

PR #13027, opened 2024-02-13
Enable non-strict js

adrinr:
## Description
The current implementation of isolated-vm for js runners is making use of modules. These modules run as ESM modules, which are processed as a `strict` code. We realised that plenty of customers usages are `non-strict`.
In order to allow the existing code to run, we updated the usage of isolated-vm to run using scripts instead. These are `non-strict` by default.

### Performance
The initial usage of modules was chosen to get a better performance and reusability of isolates. Luckily, we managed to cache these runs using the script.

#### Test runs
These tests all ran with a table with 10k rows, and with 20 columns with formulas. The tested endpoint is `/search`, as it processes all the formulas for all the rows.
| Requested rows | Current impl. avg response time | New version avg response time |
|--------|--------|--------|
| 1 | 139.66ms | 140.42ms |
| 100 | 2.06s | 1.97s |
| 200 (current max limit) | 3.87s | 3.79s | 

## Feature branch env
[Feature Branch Link](http://fb-enable-sloppy-js.fb.qa.budibase.net)


packages/server/src/jsRunner/bundles/index.ts
```diff
@@ -8,20 +8,21 @@ export const enum BundleType {
...
 const bundleSourceCode = {
-  [BundleType.HELPERS]: "../bundles/index-helpers.ivm.bundle.js",
-  [BundleType.BSON]: "../bundles/bson.ivm.bundle.js",
+  [BundleType.HELPERS]: "./index-helpers.ivm.bundle.js",
+  [BundleType.BSON]: "./bson.ivm.bundle.js",
 }
 
 export function loadBundle(type: BundleType) {
-  if (environment.isJest()) {
+  if (!environment.isBundled) {
```
mike12345567, line 16: I assume this is only set when running it through esbuild?

adrinr: I was not sure if this is the best name...

adrinr: Ok, refactor and not needed :)

samwho (approved):

packages/server/src/jsRunner/vm/isolated-vm.ts
```diff
@@ -268,7 +227,10 @@ export class IsolatedVM implements VM {
   private getFromContext(key: string) {
     const ref = this.vm.global.getSync(key, { reference: true })
     const result = ref.copySync()
-    ref.release()
+
+    new Promise(() => {
+      ref.release()
+    })
```
samwho, line 233: Not sure I understand the purpose of this. What does it fix?

adrinr: Having it async added some unnecessary delay on the process. This does it async, not really caring if it fails. Anyways, the garbage collector will pick it up, this just speeds up the process disposing it

Issue #8799, opened 2022-11-24
Import/Export Data: Make CSV delimiter configurable


oldwired:
**Describe the feature request**
Please make the CSV delimiter configurable. At the moment it is always `,`, but the delimiter excel expects depends on the localization settings. For languages like german Excel expects `;` when just using the open dialog/doubleclicking it.

melohagan:
As an additional improvement to the **Export data** action, I would like to see the column selector match the column selector from the **Data** section.

![Screenshot 2024-01-30 at 09 48 13](https://github.com/Budibase/budibase/assets/101575380/88a036ab-1000-4b39-8161-0d7f4571488f)

*This should become this:*

![Screenshot 2024-01-30 at 09 48 47](https://github.com/Budibase/budibase/assets/101575380/5e10c01b-e480-4ddf-a5c3-4875fb6d670c)

melohagan:
https://docs.budibase.com/docs/data-actions#configure-delimiter

PR #13028, opened 2024-02-13
Export data make CSV delimiter configurable

melohagan:
## Description
As a user I can now select the delimiter to be used in my CSV export from a preset list. Comma is the default.

You can also select columns via the column editor, which supports custom labels as well.  

Tested for both internal and external tables, and tested for backwards compatibility.

## Addresses
- https://github.com/Budibase/budibase/issues/8799

## Screenshots
Screenshot 2024-02-13 at 14 02 57

Screenshot 2024-02-13 at 14 03 12

Screenshot 2024-02-13 at 14 03 45


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ExportData.svelte
```diff
@@ -24,36 +18,60 @@
...
+      value: "|",
+    },
+    {
+      label: "~",
+      value: "~",
+    },
+    {
+      label: "[tab]",
+      value: "  ",
+    },
```
samwho, line 45: `[space]` might be another relatively common one.

adrinr (dismissed):
Not sure if this should be done like this. The CSV is a [standard format](https://www.ietf.org/rfc/rfc4180.txt), comma-separated values.
On the ticket they mention excel, but excel uses their own format. Even so, they allow importing CSV that should work fine with the current csv.
We should probably analyse if this is actually required

packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnEditor.svelte
```diff
@@ -32,6 +32,12 @@
     allowLinks: true,
   })
 
+  $: {
+    value = (value || []).filter(column =>
+      Object.keys(schema || {}).includes(column.name || column)
```
samwho, line 37: Could you not do `(schema || {})[column.name || column] !== undefined` here? Thinking this because `.keys` and `.includes` are both O(n) operations.

packages/server/src/api/controllers/view/exporters.ts
```diff
@@ -1,7 +1,19 @@
 import { Row, TableSchema } from "@budibase/types"
 
-export function csv(headers: string[], rows: Row[]) {
-  let csv = headers.map(key => `"${key}"`).join(",")
+function getHeaders(
+  headers: string[],
+  customHeaders: { [key: string]: string }
+) {
+  return headers.map(header => `"${customHeaders[header] || header}"`)
```
samwho, line 7: Is it strictly correct to wrap this in quotes manually? What if the header already contains quotes?

packages/server/src/api/controllers/view/exporters.ts
```diff
@@ -1,7 +1,19 @@
...
+}
+
+export function csv(
+  headers: string[],
+  rows: Row[],
+  delimiter: string = ",",
+  customHeaders: { [key: string]: string } = {}
```
samwho, line 14: This made me a little bit nervous coming from a Python background, but MDN says it's all good: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#evaluated_at_call_time.

samwho (commented):
Just saw Adri's comment before submitting. I was under the impression it's quite common to use a variety of different delimiters with CSV. For example, the CSV library in Python allows you to customise the delimiter: https://docs.python.org/3/library/csv.html#csv.reader. As do Ruby (https://ruby-doc.org/stdlib-3.0.0/libdoc/csv/rdoc/CSV.html#class-CSV-label-Options) and Go (https://pkg.go.dev/encoding/csv#example-Reader-Options).

melohagan:
> Not sure if this should be done like this. The CSV is a [standard format](https://www.ietf.org/rfc/rfc4180.txt), comma-separated values. On the ticket they mention excel, but excel uses their own format. Even so, they allow importing CSV that should work fine with the current csv. We should probably analyse if this is actually required

That makes sense, although it is something that pgAdmin allows you to do. It also allows you to change the output format to text.

CSVs should be comma separated but from what I could find it's not uncommon to use different delimiters: https://en.wikipedia.org/wiki/Comma-separated_values

packages/server/src/api/controllers/view/exporters.ts
melohagan: That's *interesting* that python works that way; would never expect that behaviour! Something like C# only allows constant default values which is probably the safest approach tbf.

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ExportData.svelte
melohagan: Added. Also updated [tab] to use the `\t` character.

packages/builder/src/components/design/settings/controls/ColumnEditor/ColumnEditor.svelte
melohagan: Nice! Done.

packages/server/src/api/controllers/view/exporters.ts
melohagan: The export data action has always been wrapping the values, including the headers, in double quotes. I imagine now that we have added space as a delimiter, the quotes will be necessary.

samwho: Fair enough. It will break if users have a `"` in their headers, but if it's been there forever that's a separate bug we can address if/when it comes up.

samwho (approved):

PR #13029, opened 2024-02-13
Create isolated-vm toggle

adrinr:
## Description
Add a toggle to enable isolated-vm. By default this will be off, allowing us to run a controlled isolated-vm release

## Feature branch env
[Feature Branch Link](http://fb-reenable-isolatedvm.fb.qa.budibase.net)


packages/server/src/environment.ts
```diff
@@ -126,20 +127,26 @@ const environment = {
   getDefaults: () => {
     return DEFAULTS
   },
+  ISOLATEDVM_QUERY_TRANSFORMERS: !!process.env.ISOLATEDVM_QUERY_TRANSFORMERS,
```
mike12345567, line 130: can we keep the environment variables together above the final exports/functions in the `environment.ts` files? Makes it easier to not miss some environment variables in the list!

mike12345567 (approved):
LGTM

packages/server/src/jsRunner/tests/jsRunner.spec.ts
```diff
@@ -24,18 +9,28 @@ const { runJsHelpersTests } = require("@budibase/string-templates/test/utils")
...
 
-describe("jsRunner", () => {
+describe.each([
+  ["vm", false],
+  ["isolated-vm", true],
+])("jsRunner (using %s)", (_, useIsolatedVM) => {
   const config = new TestConfiguration()
 
   beforeAll(async () => {
+    environment._set("ISOLATEDVM_JS_RUNNER", useIsolatedVM)
```
samwho, line 23: This is going to leak out of these tests and into other tests. I recommend using `config.withEnv`, which makes sure to set values back to what they were prior to being overridden.

mike12345567: Removed this.

Issue #13020, opened 2024-02-12
QUERY_THREAD_TIMEOUT env variable does not affect automations


enpro-github:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

- Self
  - Method: <method> 
  - Budibase Version: 2.19.0

QUERY_THREAD_TIMEOUT works fine for external connections when they're not used inside of an external connector. Otherwise queries will timeout inside of an automation after 12 seconds. 

**To Reproduce**
1. Create an external data connector in automations and run a query that takes longer than 12 seconds

**Expected behavior**
I want the query to complete the same way it does when I'm using the external connector without automations

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome

mike12345567:
Hi @enpro-github - I attempted to recreate this, queries running within automations today have no timeouts, they will run for as long as is required for them to complete.

I've tested this with REST queries that take up to 30 seconds, which continue to work fine.

enpro-github:
@mike12345567 Thanks for the reply, you tried running an automation with an external data connector that connects to a DB? In my app it's connecting to a MYSQL DB and performing a query on it. We don't have an issue with queries that don't take long, this is what we see in the inspect menu. This same query outside of the automation works fine though.

![image](https://github.com/Budibase/budibase/assets/68210951/8e6474ac-ee3f-4854-b42c-bd6284c193ac)

mike12345567:
Hi @enpro-github - thank you for the screenshot, led me to the right place, there will be a fix soon which adds another environment variable for controlling timeouts on automation actions.

PR #13032, opened 2024-02-14
Add AUTOMATION_THREAD_TIMEOUT environment variable

mike12345567:
## Description
This PR adds a new environment variable which allows controlling the timeouts on automation threads. This is particularly useful if you have very slow queries that can lock up the automation for sometime.

environment variable is `AUTOMATION_THREAD_TIMEOUT` and works in the same way as `QUERY_THREAD_TIMEOUT`.

This address: https://github.com/Budibase/budibase/issues/13020


packages/server/src/constants/automations.ts
```diff
@@ -0,0 +1,6 @@
+import { Duration } from "@budibase/backend-core"
+import env from "../environment"
+
+const defaults = env.getDefaults()
+export const AUTOMATION_SYNC_TIMEOUT = Duration.fromMinutes(2).toMs()
+export const AUTOMATION_ASYNC_TIMEOUT = defaults.QUERY_THREAD_TIMEOUT
```
samwho, line 6: This seems a bit weird. So if someone overrides `QUERY_THREAD_TIMEOUT`, this file wouldn't pick it up? It would be whatever the default is?

mike12345567: Got rid of this file and just used the defaults directly everywhere - no point in really duplicating them like this.

samwho (approved):

Issue #12965, opened 2024-02-06
Action chaining not providing result


loureirodnl:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

- Cloud
  - Tenant ID: tdh.budibase.app

**Describe the bug**
Requesting result from a previous action is not working on newly created form blocks

**To Reproduce**
1. Create a form block
2. Add the following action at the end of the save button:
![image](https://github.com/Budibase/budibase/assets/132261713/a07c0fb9-31da-43a2-a60c-4b0146500050)

3. Go to preview and create a new record
4. The result notification is empty:

![image](https://github.com/Budibase/budibase/assets/132261713/c067fdc0-d54e-4a17-8fb0-8c8f5b07e2de)

**Expected behavior**
We should see the result

  
**App Export**
[My app-export-1707217629712.tar.gz](https://github.com/Budibase/budibase/files/14178627/My.app-export-1707217629712.tar.gz)

**Desktop (please complete the following information):**
 - OS: windows 11
 - Browser Microsoft Edge

melohagan:
Thanks for raising @loureirodnl 

This is currently being worked on 👍

melohagan:
Hey @loureirodnl 

In the meantime I think a workaround could be, in your saved row action, to use the default helper per column:

![Screenshot 2024-02-08 at 09 14 24](https://github.com/Budibase/budibase/assets/101575380/1d79b08f-8c97-4319-a05d-d5a503795018)

**default** *form_field_binding* *default_value*

melohagan:
https://linear.app/budibase/issue/BUDI-7964/save-row-action-payload-not-useable-within-a-button-group

genfild:
This bug is observed in version 2.17.8. Everything works in a separate button.

genfild:
This bug is observed in version 2.17.9. Everything works in a separate button.

ConorWebb96:
This seems to be fixed on the latest version of Budibase. Going to close as completed.

PR #13034, opened 2024-02-14
Fix nested button actions

aptkingston:
## Description
This PR adds the nested flag to the button group component settings to prevent early enrichment of actions.

Alternative solution to https://github.com/Budibase/budibase/pull/12989.

## Addresses
- https://linear.app/budibase/issue/BUDI-7964/save-row-action-payload-not-useable-within-a-button-group
- https://github.com/Budibase/budibase/issues/12965


deanhannigan (approved):
LGTM

Issue #13013, opened 2024-02-12
Clicking on a formula settings cog in a grid component causes the app to freeze, hard refresh required to fix.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.19.0
  - App Version: 2.19.0

**Describe the bug**
When trying to view the settings of a grid column which is a formula the application crashes/can't interact with anything else until you hard refresh the page.

**To Reproduce**
1. Add a grid component to the design screen
2. Click the grid component
3. Make sure one of the columns is a static column type
4. On the right hand side of the screen 
5. Click the settings cog related to the formula column
6. See error

**Expected behavior**
I'm not sure formula columns should allow you to view the settings, possibly remove the settings cog to prevent this from happening. Other things such as moving its place etc/hiding it should still work.

**Screenshots**
<img width="539" alt="Screenshot 2024-02-12 at 15 04 11" src="https://github.com/Budibase/budibase/assets/126772285/e30d7de1-a592-4f04-b9b5-d743685f89cc">

**App export**
[grid issue-export-1707750455525.tar.gz](https://github.com/Budibase/budibase/files/14239878/grid.issue-export-1707750455525.tar.gz)

PR #13035, opened 2024-02-14
Use constants for icon info rather than component definitions

aptkingston:
## Description
Fixes a crash when trying to rename a formula column in a grid block. This was happening because we couldn't find the correct icon for a formula column, so this PR updates how we find icons to use the backend-core constants, which is a more accurate source than relying on field component lookups in the manifest (as there are certain data types we don't have components for, like formula columns as seen here).

## Addresses
- https://linear.app/budibase/issue/BUDI-8001/clicking-on-a-formula-settings-cog-in-a-grid-component-causes-the-app
- https://github.com/Budibase/budibase/issues/13013


melohagan (approved):
LGTM!

PR #13036, opened 2024-02-14
Update Svelte Major Version to 4

Ghrehh:
**NOTE**: The first two commits of this PR are the main ones, one has a bunch of repetitive changes temporarily commenting out Svelte warnings, and the other contains the actual real changes.

### Main change
- Bumps the Svelte version in all relevant packages from `3.49.0` to `4.2.10` which at time of writing is the latest stable version, this was mostly painless but has a few run on effects. We also stopped specifying the svelte version in the submodules individually, instead we add it once at the top level.

### Changes to `packages/account-portal`
There is a corresponding account portal change the should be merged first https://github.com/Budibase/account-portal/pull/530

### Changes to `packages/server`
- The [svelte/register](https://svelte.dev/docs/svelte-register) function we previously used to render our client app templates was removed. The Svelte docs in general are a bit lacking on SSR stuff, and the recommended approach for Svelte 4 seems to be "use a bundler or Sveltekit". I opted to add a custom plugin for esbuild, which is the build tool we use for our backend projects, which manually calls the `compile` function Svelte exports. These changes are in `scripts/build.js`.
- This means that Svelte components can be directly imported into our backend code bases and are treated like JS.
- This means we no longer need to import the component at runtime like we did previously.
- These changes also mean that a Svelte file imported and rendered on the backend can itself import and use other Svelte components, which wasn't possible before.
- This does mean that that components and the Svelte runtime are bundled into the main js output, but at a glance this extra code makes no noticeable difference to size, it's at least as small a change as 100kb, but possibly much smaller, that's just how granular `ls` was when I checked. 
- It'd also be pretty easy to handle our `hbs` files in a similar way to the Svelte files, and import them directly and include them in the bundle, currently we're also opening those files at run time.
- I made some general changes to the main `scripts/build.js` to make the js/hbs/svelte build process a little more async.
- I opted to remove the use of `glob` in the above files, because as far as I can tell it's only available for use as it's a dependency of a dependency, and isn't directly pinned anywhere. To avoid adding another dependency I opted to use [`fsPromises.opendir`](https://nodejs.org/api/fs.html#fspromisesopendirpath-options) with the `recursive` flag which seems to fulfil the same task. 
- We had to make some changes to the jest config to get svelte files working, including adding a custom transformer for svelte files.

Hope this work all looks good, I appreciate that making changes to the build process is stepping on other team's toes a bit, so would appreciate any input🙏

### Changes to `packages/client`
- The client package uses Rollup instead of Vite, and the client would trigger any `onMount` hooks unless we updated to the latest version of Vite which is from `2.44.0` -> `4.9.6`. Were still using some deprecated Rollup plugins, but the client seems to function fine.

### Changes to all frontend packages
- Svelte now loudly complains about a lack of accessibility considerations, like onclick events with no button events etc. I've a PR in the works that addresses some of this stuff, so I opted to just add comments to disable them all for now.

## Ticket
https://linear.app/budibase/issue/BUDI-7995/update-svelte-to-latest-4x


aptkingston:
> Sorry to bump in ! Will this require updating the custom components to Svelte 4 too ?

Hey! Custom components will automatically use the version of svelte that Budibase is using, so any custom components already built and being used inside apps will automatically start using version 4. This shouldn't cause any issues - there were no major changes. We will update the skeleton repository for plugins to ship with version 4 too.

You can absolutely update your own plugins to version 4 if you want, but it should only affect the build step as we don't bundle Svelte itself inside custom components.

TLDR: No need to make any changes, unless you want to! Everything should continue to work just fine.

packages/bbui/package.json
```diff
@@ -25,7 +25,7 @@
     "rollup-plugin-postcss": "^4.0.0",
     "rollup-plugin-svelte": "^7.1.0",
     "rollup-plugin-terser": "^7.0.2",
-    "svelte": "3.49.0"
+    "svelte": "^4.2.10"
```
aptkingston, line 28: If we don't pin the version, can we guarantee we'll never accidentally update to a new patch version in one package and not another? I'd hate to accidentally start bundling 2 versions of svelte because of that (and it would increase bloat significantly). That's the main reason we explicitly pinned this before.

aptkingston (approved):
LGTM! Fingers crossed we get some free performance improvements from this too.

poirazis:
Thanks @aptkingston !
I will definetly rebuild all with v4 ! It offers 12% smaller build size they claim and faster performace !
Great job guys !

packages/bbui/package.json
Ghrehh: I'm not 100% on how this works in a monorepo/workspace tbh. The only `yarn.lock` is in the project root , so I don't think this would be possible as long as every package has the same svelte version specified, even with the caret.

As I say though, not 100% so if you're more comfortable pinning it to an exact version I'm happy to do so.

Ghrehh: The team talked offline and we realised we can just specify the svelte version in the root of the monorepo and remove it from the child repos' `package.json`. Probably a bunch of other packages we can do this for too, but will punt on that for now.

PR #13040, opened 2024-02-15
Fix Arrow Key Behaviour for Collapsed Components

Ghrehh:
https://github.com/Budibase/budibase/pull/12764 was merged with a bug where the arrow keys weren't performing actions to navigate between collapsed components correctly, this should fix it.


deanhannigan (approved):
LGTM

PR #13049, opened 2024-02-15
Type workerRequests.ts

samwho:
## Description

While reviewing https://github.com/Budibase/budibase/pull/13048 I noticed that the file has some `any`s floating around, so I decided to obliterate them.

Given that the type for `HeadersInit` is actually quite wide, this wasn't trivial and does include some logic changes. A comment on Mike's PR makes me think this might not be tremendously well-tested code, so advice on how to test this manually would be appreciated 🙏


packages/server/src/utilities/workerRequests.ts
```diff
@@ -167,15 +207,15 @@ export async function getChecklist(): Promise<{
...
+    request({ method: "GET" })
   )
   return checkResponse(response, "get checklist")
 }
 
 export async function generateApiKey(userId: string) {
   const response = await fetch(
     checkSlashesInUrl(env.WORKER_URL + "/api/global/self/api_key"),
-    request(undefined, { method: "POST", body: { userId } })
+    request({ method: "POST", body: JSON.stringify({ userId }) })
```
mike12345567, line 218: This shouldn't be necessary - the request checks if its a JSON type and sets the JSON content type header.

packages/server/src/utilities/workerRequests.ts
```diff
@@ -93,9 +132,9 @@ export async function sendSmtpEmail({
   // tenant ID will be set in header
   const response = await fetch(
     checkSlashesInUrl(env.WORKER_URL + `/api/global/email/send`),
-    request(undefined, {
+    request({
       method: "POST",
-      body: {
+      body: JSON.stringify({
```
mike12345567, line 137: Same comment as later one (sorry wrong way around!)

packages/server/src/utilities/workerRequests.ts
```diff
@@ -10,29 +16,61 @@ import {
...
+}
+
+export function request(request: RequestInit & { ctx?: Ctx }): RequestInit {
+  const ctx = request.ctx
+  request.headers = ensureHeadersIsObject(request.headers)
+
   if (!ctx) {
-    request.headers[constants.Header.API_KEY] = coreEnv.INTERNAL_API_KEY
+    if (coreEnv.INTERNAL_API_KEY) {
+      request.headers.set(constants.Header.API_KEY, coreEnv.INTERNAL_API_KEY)
```
mike12345567, line 47: I'm not sure this is strictly correct for the type `request` actually is - its a request to node fetch in which the headers are simply an object.

packages/server/src/utilities/workerRequests.ts
samwho: If I try and pass it as an object it's a type error. I found the type in the `@types/node-fetch` library: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b2c721481ee38f65eb6630e958cc0c3383c81831/types/node-fetch/index.d.ts#L195.

An example of types not matching the reality of the library?

packages/server/src/utilities/workerRequests.ts
samwho: The type for `request.headers` is `HeadersInit` which is defined here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b2c721481ee38f65eb6630e958cc0c3383c81831/types/node-fetch/index.d.ts#L194. It's a bit wider than just an object, so I created the `ensureHeadersObject` function to make sure it's always a `Headers` object, and used that throughout. Wasn't sure about this as I was doing it but it gets the type checking the pass, and handles the wider definition of what headers can be.

packages/server/src/utilities/workerRequests.ts
mike12345567: ah - yes the type for node-fetch should be a string, however we use the `request` function to build a node-fetch request, and in this instance our function handles checking if the passed in body is actually JSON, sets the header and then stringifies it.

samwho: Where does our function stringify the body?

packages/server/src/utilities/workerRequests.ts
mike12345567: `node-fetch` has a very weird Typescript definition here, so what is happening:
1. When you are building a request to be made by node-fetch, generally you need to just use a map, as your input is simply JSON e.g.
```fetch({
  headers: {...}
})```
This is what we are building towards, so I'm not sure the `.set` functions will be available.
2. When a node-fetch response is received it will come back in the format described by `Headers` - its inconsistent and very frustrating, but it does happen that way.

packages/server/src/utilities/workerRequests.ts
mike12345567: https://github.com/Budibase/budibase/pull/13049/files/dde8f778774ab756edb51e1c4e2a208e80b70629#r1491164920

mike12345567 (approved):
Excellent - looks good! ✅

PR #13052, opened 2024-02-15
Form screen template

deanhannigan:
## Description

- New `Form` mode has been added to the `Create new screen` menu. Pick 1 or more existing datasources and generate a form screen
- Select the initial form mode - `Create`, `Update` or `View`. The screen will contain a `Multistep form block` with your default configuration.
- When creating a new `Grid`, `Table` or `Form` screen, the user will now be redirected to a component path. This way, the newly created component is immediately selected in the designer.
- Tour refactor. Ensuring that internal flags in the TourPopover are appropriately cleared when the tour is changed. Ensure that when destroying tour steps any related tour is stopped.
- Fix for Analytics clients to stop them continuously initialising.

## Addresses
- https://linear.app/budibase/issue/BUDI-7951/form-screen-template

## Screenshots
New `Form` tile when creating a new screen

Select which type of form you wish to create: `Create`, `Update` or`View`

New tour prompts when creating forms. `Create` mode has a unique, 1-step tour and the other modes use a different tour including the `Row Id`

## Feature branch env
[Feature Branch Link](http://fb-feature-form-screen-template.fb.qa.budibase.net)


packages/builder/src/components/portal/onboarding/TourWrap.svelte
```diff
@@ -1,39 +1,62 @@
...
+      tourStepKeyWatch
+    )
+    let currentStep = TOURS[tourKeyWatch]?.steps?.[tourStepIdx]
+    if (currentStep?.scrollIntoView) {
+      let currentNode = $builderStore.tourNodes?.[currentStep.id]
+      if (currentNode) {
+        currentNode.scrollIntoView({ behavior: "smooth", block: "center" })
+      }
+    }
+  }
```
aptkingston, line 40: Might be better to pull this out into a function - I don't think big multi-line reactive statements are a great idea. This will re-evaluate for any change to the `builderStore` for example, which would mean that we scroll the node into view even if a completely unrelated change to the builder store is made while the other conditions are satisfied.

I think it would make sense to pull it into a separate function and call it in a reactive statement, passing in only those observed params that we care about.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte
```diff
@@ -177,9 +177,10 @@
               defaultValue={setting.defaultValue}
               nested={setting.nested}
               onChange={val => updateSetting(setting, val)}
-              highlighted={$componentStore.highlightedSettingKey ===
-                setting.key}
-              propertyFocus={$componentStore.propertyFocus === setting.key}
+              highlighted={$builderStore.highlightedSetting?.key === setting.key
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ComponentSettingsSection.svelte:180]: Can this not just be 
```js
highlighted={$builderStore.highlightedSetting?.key === setting.key}
```
`highlighted` is a boolean prop, so we just need to check if the keys match.

aptkingston (commented):
Couple of small suggestions but overall LGTM!

Could we add a checkbox to the form type selection modal to keep in line with the other templates? That's the only visual change I can see that we need I think.
image

aptkingston (approved):
Could we add a line break into this paragraph of text?
image

Just to look like:
image

Then go ahead and merge - LGTM!

PR #13058, opened 2024-02-16
Change budibase/couchdb default storage location

mike12345567:
## Description
Updating the default to align with what K8S and docker-compose uses, for AAS and single image we control the `TARGETBUILD` environment variable fully.

This is required because we don't always get `TARGET_BUILD=docker-compose` set on docker-compose setups - in which case it will not store data to the volume.


hosting/couchdb/runner.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DATA_DIR=${DATA_DIR:-/data}
+DATA_DIR=${DATA_DIR:-/opt/couchdb/data}
```
samwho, line 3: I'm really not sure if we can make this change safely. It will screw over anyone running this image without having set `TARGETBUILD`. I think we may have to just live with it as it is now?

_some time passes_

It actually occurs to me that the big if-statement below knocks reality out of sync with what the `DATA_DIR` is set to. Maybe `DATA_DIR` isn't actually used in practice anymore?

PR #13063, opened 2024-02-19
Error when deleting selected rows that have attachment

melohagan:
## Description
When trying to delete a row with an empty attachment column, the `map` function was throwing on a null-pointer. 
Fixed to make sure null attachment field values are set to empty array.

## Addresses
- https://linear.app/budibase/issue/BUDI-7996/error-when-deleting-selected-rows-that-have-attachment-columns


samwho (changes requested):
Could you add a test for this? 🙏

melohagan:
@samwho The original fix actually did get caught by an existing unit test, so I've updated the approach and added a unit test for it 👍

samwho (approved):

PR #13064, opened 2024-02-19
Handle branding cookies with different tenant ids

PClmnt:
## Description
Handle the setting of the tenantID for branding purposes whenever a user has been to a different tenant. This will now overwrite the previous cookie value instead of creating a new one. Also adds a check to only do this cookie setting whenever the environment is cloud.


mike12345567 (approved):

packages/builder/src/pages/builder/auth/_layout.svelte
```diff
@@ -10,13 +10,18 @@
     $redirect("../")
   }
 
-  if ($admin?.checklist?.branding) {
+  if ($admin.cloud && $admin?.checklist?.branding) {
```
melohagan, line 13: NAB: Can `$admin` be null?

melohagan (approved):

packages/builder/src/pages/builder/auth/_layout.svelte
PClmnt: Pretty sure no, but better to be safe than sorry.

PR #13075, opened 2024-02-20
Fix re-used context in JS runner.

samwho:
## Description
We have an issue with context not correctly being passed into every instance of the JS runner, this should resolve the problem by making sure to add context every time and cleanup after completion.

### DIFF
--- a/packages/server/src/jsRunner/vm/isolated-vm.ts
+++ b/packages/server/src/jsRunner/vm/isolated-vm.ts
@@ -97,11 +97,11 @@ export class IsolatedVM implements VM {
     return this
   }
 
-  withContext<T>(context: Record<string, any>, f: () => T) {
+  withContext<T>(context: Record<string, any>, executeWithContext: () => T) {
     this.addToContext(context)
 
     try {
-      return f()
+      return executeWithContext()
     } finally {
       this.removeFromContext(Object.keys(context))
     }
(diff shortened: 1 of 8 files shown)


mike12345567:
Might need to implement the `withContext` on the `builtin-vm` as well

packages/server/src/jsRunner/vm/isolated-vm.ts
```diff
@@ -97,10 +97,14 @@ export class IsolatedVM implements VM {
     return this
   }
 
-  withContext(context: Record<string, any>) {
+  withContext<T>(context: Record<string, any>, f: () => T) {
```
adrinr, line 100: Can we name this executeWithContext?

packages/server/src/jsRunner/vm/isolated-vm.ts
```diff
@@ -97,10 +97,14 @@ export class IsolatedVM implements VM {
...
     this.addToContext(context)
 
-    return this
+    try {
+      return f()
+    } finally {
+      this.removeFromContext(Object.keys(context))
```
adrinr, line 106: Might this be deleting staff that was defined somewhere else?

samwho: It's not deleting anything that wouldn't have already been overridden when we set the context, so I figured it's not a concern. Open to situations where it might be, though.

mike12345567 (approved):

PR #13079, opened 2024-02-20
Fix flaky table test.

samwho:
## Description

This test was expecting the result of `/api/tables` to contain a specific table as the first result. This isn't guaranteed due to the parallel nature of our tests. I've modified it to search the results for the table in question.

## Addresses
- https://linear.app/budibase/issue/BUDI-8027/solve-flaky-test-issues


adrinr (approved):
❤️

mike12345567 (approved):
No more corn flakes, yay!

PR #13082, opened 2024-02-20
Dynamic schema generation for query arrays: bug fix and refactor

melohagan:
## Description
- Fix for `Cannot convert undefined or null to object` error when array is empty.
- Refactored to build nested schema from all the rows, as MongoDB schemas are not strictly enforced (properties may be missing on some but not all documents)
- Unit tests added

## Addresses
- https://linear.app/budibase/issue/BUDI-8026/cannot-convert-undefined-or-null-to-object-when-running-query-against


packages/server/src/api/controllers/query/index.ts
```diff
@@ -142,6 +142,78 @@ export async function preview(ctx: UserCtx) {
...
+    if (field)
+      switch (type) {
+        case "boolean":
+          fieldMetadata = makeQuerySchema(FieldType.BOOLEAN, key)
+          break
+        case "object":
+          if (field instanceof Date) {
+            fieldMetadata = makeQuerySchema(FieldType.DATETIME, key)
+          } else if (Array.isArray(field)) {
+            const firstFiveItems = field.slice(0, 5)
```
mike12345567, line 172: wouldn't use just the first five - would bump this to full data set as its easier to understand what is happening rather than explaining its just the first five.

packages/server/src/api/controllers/query/index.ts
```diff
@@ -142,6 +142,78 @@ export async function preview(ctx: UserCtx) {
...
+          break
+        case "number":
+          fieldMetadata = makeQuerySchema(FieldType.NUMBER, key)
+          break
+      }
+    return fieldMetadata
+  }
+
+  function buildSchemaFromArray(
+    firstFiveItems: any[],
```
samwho, line 198: nit: I wouldn't call this `firstFiveItems`, as there's no such limitation inside this function. `items` is probably sufficiently general.

packages/server/src/api/controllers/query/index.ts
```diff
@@ -142,6 +142,78 @@ export async function preview(ctx: UserCtx) {
...
+    }
+  ) {
+    let schema = {}
+    for (const item of firstFiveItems) {
+      if (JsonUtils.hasSchema(item)) {
+        for (const [key, value] of Object.entries(item)) {
+          schema = {
+            ...schema,
+            [key]: getFieldMetadata(value, key, nestedSchemaFields),
+          }
```
samwho, line 210: Is there a reason to use the spread operator here over `schema[key] = ...`? The spread operator will make a copy of the `schema` every time this is done, which is wasteful if not necessary.

packages/server/src/api/controllers/query/index.ts
```diff
@@ -142,6 +142,78 @@ export async function preview(ctx: UserCtx) {
...
+                key,
+                JsonFieldSubType.ARRAY
+              )
+            } else {
+              fieldMetadata = makeQuerySchema(FieldType.ARRAY, key)
+            }
+            nestedSchemaFields[key] = buildSchemaFromArray(
+              firstFiveItems,
+              nestedSchemaFields
+            )
```
samwho, line 185: This modification as a side effect in a function called `getFieldMetadata` doesn't feel great, but it would likely require a substantial refactor to unwind. No action required, just felt compelled to point this out. If you can think of some way to make this nicer that doesn't completely derail the PR, feel free, but don't worry if you can't.

packages/server/src/api/routes/tests/queries/query.seq.spec.ts
```diff
@@ -249,6 +251,126 @@ describe("/queries", () => {
...
+      pg.queryMock.mockImplementation(() => ({
+        rows,
+      }))
+
+      const res = await request
+        .post(`/api/queries/preview`)
+        .send(query)
+        .set(config.defaultHeaders())
+        .expect("Content-Type", /json/)
+        .expect(200)
```
samwho, line 336: Would be good to extract this out into the shared API client we use in tests. The one inside `config`.

packages/server/src/api/controllers/query/index.ts
melohagan: As discussed, removed the limit and performing the schema build across all rows. Didn't appear to have any noticeable performance impact.

packages/server/src/api/controllers/query/index.ts
melohagan: Nice! Done.

packages/server/src/api/routes/tests/queries/query.seq.spec.ts
melohagan: done!

packages/server/src/api/controllers/query/index.ts
melohagan: Good call - I've put this modification into its own function

packages/server/src/api/controllers/query/validation.ts
```diff
@@ -5,9 +5,9 @@ const OPTIONAL_STRING = Joi.string().optional().allow(null).allow("")
 
 export function queryValidation() {
   return Joi.object({
-    _id: Joi.string(),
-    _rev: Joi.string(),
-    name: Joi.string().required(),
+    _id: OPTIONAL_STRING,
+    _rev: OPTIONAL_STRING,
+    name: OPTIONAL_STRING,
```
samwho, line 10: Is this changing from required to optional on purpose?

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -867,23 +868,10 @@ export default class TestConfiguration {
...
 
-  async previewQuery(
-    request: any,
-    config: any,
-    datasource: any,
-    fields: any,
-    params: any,
-    verb?: string
-  ) {
+  async previewQuery(request: any, config: any, queryPreview: QueryPreview) {
```
samwho, line 871: Huh. Sorry, I had no idea this was here. When I made my comment I was thinking of something like `config.api.query`, the `QueryAPI` class. I wasn't clear on that at all, my bad. Would you mind moving this to there, for consistency?

packages/server/src/api/controllers/query/validation.ts
melohagan: Good catch - we recently have been unifying the `QueryPreview` and `Query` objects as there is very little difference between the two, but this is a valid difference.

A query preview can be sent without a name, but a query cannot be saved without a name.

I've update the code to reflect this.

packages/server/src/tests/utilities/TestConfiguration.ts
melohagan: No worries - moved!

packages/server/src/api/controllers/query/index.ts
```diff
@@ -142,6 +158,66 @@ export async function preview(ctx: UserCtx) {
...
+      name: string,
+      subtype?: string
+    ): QuerySchema => ({
+      type,
+      name,
+      subtype,
+    })
+    let type = typeof field,
+      fieldMetadata = makeQuerySchema(FieldType.STRING, key)
+    if (field != null)
```
samwho, line 173: So if I'm following the logic correctly, we're allowing `field` to be `null` or `undefined` and in these cases, we assume it's a string. Is that correct?

If so, this feels like a bug waiting to happen. Would it not be better to raise an error when `field` is `null` or `undefined`, and do a strict check for the `case "string"`, in the spirit of blowing up as early as possible when something we don't expect to happen happens?

packages/server/src/api/controllers/query/index.ts
```diff
@@ -155,51 +231,16 @@ export async function preview(ctx: UserCtx) {
...
-      type: FieldType,
-      name: string,
-      subtype?: string
-    ): QuerySchema => ({
-      type,
-      name,
-      subtype,
-    })
     if (rows?.length > 0) {
       for (let key of [...new Set(keys)] as string[]) {
```
samwho, line 235: I know you didn't write it but I think this is another case of wasteful spread operator usage. I think the following would work fine:

```javascript
for (let key of new Set(keys)) {
  // ...
}
```

And it would save converting from a `Set` to an array for no good reason.

packages/server/src/api/controllers/query/index.ts
```diff
@@ -155,51 +231,16 @@ export async function preview(ctx: UserCtx) {
...
-                    key,
-                    JsonFieldSubType.ARRAY
-                  )
-                } else {
-                  fieldMetadata = makeQuerySchema(FieldType.ARRAY, key)
-                }
-                nestedSchemaFields[key] = getSchemaFields(
-                  field,
-                  Object.keys(field[0])
-                ).previewSchema
```
samwho, line 193: I don't seem to be able to find where you've moved this to. I was looking for usages of `getSchemaFields` and I only find it here, not in any of the new code. What am I missing?

melohagan: `getSchemaFields` is called on line 270: 

```ts
const { previewSchema, nestedSchemaFields } = getSchemaFields(rows, keys)
```

Originally this code was recursive:
```ts
nestedSchemaFields[key] = getSchemaFields(
                  field,
                  Object.keys(field[0])
                ).previewSchema
```

However because of the change to use `buildNestedSchema` which loops over all the rows to generate the schema, I removed the recursive aspect.

packages/server/src/api/controllers/query/index.ts
melohagan: That's right, if a field has `null` or `undefined` in the query response we assume it is string field type.

I believe we have to do this because we're trying to build a schema from dynamic input. If the value is null/undefined, assuming that the type is a string is the safest choice. There is no Unknown field type. 

I don't think we'd ever want to throw an error because we generate the schema from the first row:

```ts
const fieldMetadata = getFieldMetadata(rows[0][key], key)
```

samwho: That makes sense. Could you leave a comment explaining that this is what we're doing next to the string assumption? 🙏

samwho (approved):

mike12345567 (approved):

PR #13094, opened 2024-02-21
Fix flaky user test

adrinr:
## Description
We are having plenty of flakiness on the pipelines, usually on random tests: https://github.com/Budibase/budibase/actions/runs/7985835708/job/21805037961

Doing some research we found out that our testconfig is setting up some default values, that are then used in multiple places. This caused different users, at multiple points, to be created with the same email.

### DIFF
--- a/packages/server/src/tests/utilities/TestConfiguration.ts
+++ b/packages/server/src/tests/utilities/TestConfiguration.ts
@@ -301,7 +301,7 @@ export default class TestConfiguration {
       lastName = generator.last(),
       builder = true,
       admin = false,
-      email,
+      email = generator.email(),
       roles,
     } = config
 
--- a/packages/server/src/tests/utilities/TestConfiguration.ts
+++ b/packages/server/src/tests/utilities/TestConfiguration.ts
@@ -91,7 +91,6 @@ export default class TestConfiguration {
   prodApp: any
   prodAppId: any
   user: any
-  globalUserId: any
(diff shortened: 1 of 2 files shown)


packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -358,9 +354,9 @@ export default class TestConfiguration {
     } = {}
   ): Promise<User> {
     let { id, firstName, lastName, email, builder, admin, roles } = user
-    firstName = firstName || this.defaultUserValues.firstName
-    lastName = lastName || this.defaultUserValues.lastName
-    email = email || this.defaultUserValues.email
+    ;(firstName = firstName || generator.first()),
+      (lastName = lastName || generator.last()),
+      (email = email || generator.email())
```
samwho, line 359: This looks super weird. It would probably make sense to just have multiple `let`s but I'm not all that bothered if you want to leave it as is.

samwho (approved):

packages/server/src/tests/utilities/TestConfiguration.ts
adrinr: Not sure how this even happened 😓 probably some autoformatter after the replaces

adrinr: https://github.com/Budibase/budibase/pull/13096

PR #13096, opened 2024-02-21
Format tests code

adrinr:
## Description
Clean strangely formatted code after https://github.com/Budibase/budibase/pull/13094

### DIFF
--- a/packages/server/src/tests/utilities/TestConfiguration.ts
+++ b/packages/server/src/tests/utilities/TestConfiguration.ts
@@ -353,14 +353,16 @@ export default class TestConfiguration {
       roles?: UserRoles
     } = {}
   ): Promise<User> {
-    let { id, firstName, lastName, email, builder, admin, roles } = user
-    ;(firstName = firstName || generator.first()),
-      (lastName = lastName || generator.last()),
-      (email = email || generator.email())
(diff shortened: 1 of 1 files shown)


packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -353,14 +353,16 @@ export default class TestConfiguration {
...
-    roles = roles || {}
-    if (builder == null) {
-      builder = true
-    }
+    const {
+      id,
+      firstName = generator.first(),
+      lastName = generator.last(),
+      email = generator.email(),
+      builder = true,
```
samwho, line 361: TIL!

```javascript
> obj = {"foo": 1, "bar": 2}
{ foo: 1, bar: 2 }
> const { foo, bar = 3 } = obj
undefined
> foo
1
> bar
2
```

samwho (approved):

packages/server/src/tests/utilities/TestConfiguration.ts
adrinr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value
You can define defaults to all destructures

PR #13097, opened 2024-02-21
Enforce using example.com as a domain for emails.

samwho:
## Description

Same reasoning as in https://github.com/Budibase/budibase/pull/12993. Just gives us an extra layer of safety if we do something wrong.

Fix for account portal lint failure is in https://github.com/Budibase/account-portal/pull/538


shogunpurple (approved):
LGTM

mike12345567 (approved):

PR #13112, opened 2024-02-22
Remove Airtable from data source ui

ConorWebb96:
## Description
@mjashanks asked for this to be removed.

Currently the Airtable connection is broken due to a recent change for how they handle their authentication. What this PR does is directly removes Airtable object from our API meaning that older connections that may still be working can continue to be used. This only hides the connection from the UI.

This is a temporary measure until a decision is made whether we will fix the connection or completely remove it.

## Screenshots
**Before**
Screenshot 2024-02-22 at 11 02 52

**After**
Screenshot 2024-02-22 at 11 03 39


packages/server/src/api/controllers/integration.ts
```diff
@@ -1,13 +1,37 @@
...
+    console.error("Error fetching definitions:", error)
+    ctx.status = 500
+    ctx.body = { error: "Internal server error" }
+
+    return {}
+  }
+}
+
 export async function fetch(ctx: BBContext) {
-  ctx.status = 200
```
shogunpurple, line 5: Is all the other code necessary? Can we not just do:

```
const definitions = await getDefinitions()
delete definitions.AIRTABLE
ctx.status = 200
ctx.body = definitions
```

mike12345567:
Made a few changes - give me a shout for explanation of what I did!

mike12345567 (approved):

PR #13114, opened 2024-02-22
Fix SCIM group creation

adrinr:
## Description
SCIM was created having in mind that we even have ALL users and groups created via SCIM, or none.
We found a bug where mixing groups (that is technically allowed) crashes on group fetch.
This is fixed in this pro PR:
- https://github.com/Budibase/budibase-pro/pull/265

This PR adds tests for this use case


mike12345567 (approved):
LGTM - nice find!

PR #13116, opened 2024-02-22
App list actions

deanhannigan:
## Description

Adds in a new context menu, allowing you to perform some of the handiest app actions straight from the app list screen.

- `Duplicate App` is a new app action that makes a new application from the selected app. This will copy the entire contents of the current dev application including any internal datasource rows and/or attachments.
- `Export lasted edited app` - will export the app with any unpublished changes
- `Export lasted published app` - will only export the currently published version of the application *if* it's available
- `Delete` will prompt users to re-enter their application name and delete their applications. A general update was made to the modal so that you can now click on the app name to automatically populate the confirmation field.
- `FIX` - updated app delete permissions from `GLOBAL_BUILDER` to `BUILDER` to allow app `Creators` to delete their own applications and any application that they have been given `edit` access to. 

## Addresses
- https://linear.app/budibase/issue/BUDI-7677/apps-list-add-more-options-to-the-apps-list

## Screenshots
Access the available `Duplicate`, `Delete` and `Export` options in the app list context menu.

Screenshot 2024-02-21 at 15 35 06

The new duplicate app menu. Enforces the same rules used when creating new applications: no duplicate names or app urls.

The delete menu has been updated to allow users to click on the app name to populate the text field. Should shave of those valuable seconds.

Context menu also accessible in the app preview navigation.

## Feature branch env
[Feature Branch Link](http://fb-feature-app-list-actions.fb.qa.budibase.net)


aptkingston (changes requested):
NAB but there's a space after the clickable app name here, before the question mark:
image

Important change needed:
If you're a creator (so have access to the portal) but only have user level access to someone elses app, you will currently still see the context menu for that app in the side nav, when you shouldn't:
image
The main app list handles this correctly, and hides the context menu and edit button, at least when the app is unpublished. The feature branch got nuked before I could check with a published app.

Everything works perfectly though, so great job!

packages/server/src/api/controllers/application.ts
```diff
@@ -626,6 +641,61 @@ export async function importToApp(ctx: UserCtx) {
...
+        file: {
+          path: tmpPath,
+        },
+      },
+      // Mark the create as a duplicate to kick off the correct event.
+      duplicate: true,
+    },
+  }
+
+  await create(createReq)
```
samwho, line 684: `create` could have its signature changed to `create(ctx: UserCtx<CreateAppRequest, App>)` and then that could flow into here to get rid of the `any` above. In general we're trying not to add any new `any`s into the backend code unless they're unavoidable.

packages/types/src/api/web/application.ts
```diff
@@ -11,6 +11,7 @@ export interface CreateAppRequest {
   includeSampleData?: boolean
   encryptionPassword?: string
   templateString?: string
+  file?: any
```
samwho, line 14: Looking at the place this is used, could this be `file?: { path: string }`?

packages/server/src/api/controllers/application.ts
```diff
@@ -626,6 +641,61 @@ export async function importToApp(ctx: UserCtx) {
   ctx.body = { message: "app updated" }
 }
 
+/**
+ * Create a copy of the latest dev application.
+ * Performs an export of the app, then imports from the export dir path
+ */
+export async function duplicateApp(ctx: UserCtx) {
```
samwho, line 648: Could you add request and response types to `UserCtx` here?

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -589,6 +590,16 @@ export default class TestConfiguration {
...
 
+  async duplicateApp(appId: string, fields: object) {
+    return context.doInTenant(
+      this.tenantId!,
+      async () =>
+        await this._req(appController.duplicateApp, fields, {
+          appId,
+        })
+    )
+  }
```
samwho, line 601: We're in the process of migrating out API calls from the `TestConfiguration` class and into the relevant API classes under `packages/server/src/tests/utilities/api`. Could you move this into the `ApplicationAPI` class, please? 🙏

deanhannigan:
Holding off on merging until this is merged in: https://github.com/Budibase/budibase/pull/13205

packages/server/src/api/controllers/application.ts
```diff
@@ -626,6 +643,67 @@ export async function importToApp(ctx: UserCtx) {
...
+  // Build the new application
+  await create(createRequest)
+  const { body: newApplication } = createRequest
+
+  if (!newApplication) {
+    ctx.throw(500, "There was a problem duplicating the application")
+  }
+
+  ctx.body = {
+    message: "app duplicated",
```
samwho, line 700: Is this field being added to the response to keep the type checker happy and make it such that you can assert agains this message in tests?

If it is, you can remove it and instead use the `Expectations.body` field in your test API calls (you've used `Expectations.status` to assert non-200 statuses in some tests) to assert the contents of an error response without having to pollute the successful type.

MihailHadzhiev2022 (approved):
LGTM

samwho (approved):
Looking good from my perspective! 🚀

aptkingston (approved):

PR #13119, opened 2024-02-22
Plan B: Loosen validation

melohagan:
## Description
Only merge and deploy if the page refresh solution doesn't work.

## Addresses
- https://github.com/Budibase/budibase/pull/13118


packages/server/src/api/controllers/query/validation.ts
```diff
@@ -10,11 +10,14 @@ function baseQueryValidation() {
     fields: Joi.object().required(),
     datasourceId: Joi.string().required(),
     readable: Joi.boolean(),
-    parameters: Joi.array().items(
-      Joi.object({
-        name: Joi.string(),
-        default: Joi.string().allow(""),
-      })
+    parameters: Joi.alternatives(
```
mike12345567, line 13: I think this might not be quite the right format, I think it is supposed to be:
```js
Joi.alternatives().try(
  Joi.array().items(
    Joi.object({
      name: Joi.string(),
      default: Joi.string().allow(""),
    })
  ),
  Joi.object({}).required().unknown(true)
)
```

Ghrehh (approved):

melohagan:
Plan A appears to have worked! Closing.

PR #13128, opened 2024-02-23
SQL Query aliasing

mike12345567:
## Description
This is part of the work for SQS to make the queries we generate for SQS a bit more manageable.

The main focus of this PR is to try and make SQL a bit more manageable/easy to follow. There are two problems that our SQL generation runs into today:
1. As we use the table name to prefix columns when referencing them (to avoid problems with column name overlaps) we run into the 63 character limit for column keys in SQL queries. If a table has a long name, then some columns will be completely inaccessible today. This PR increases the max column length we can support significantly, to 62 characters (table alias is only a single character).
2. The SQL that is usually generated is completely un-readable. For example, before a query with 3 different relationships would look like:
```sql
SELECT "tasks"."executorid" as "tasks.executorid", "tasks"."taskname" as "tasks.taskname", "tasks"."taskid" as "tasks.taskid", "tasks"."completed" as "tasks.completed", "tasks"."qaid" as "tasks.qaid", "products"."productname" as "products.productname", "products"."productid" as "products.productid", "persons"."year" as "persons.year", "persons"."firstname" as "persons.firstname", "persons"."personid" as "persons.personid", "persons"."address" as "persons.address", "persons"."age" as "persons.age", "persons"."type" as "persons.type", "persons"."city" as "persons.city", "persons"."lastname" as "persons.lastname", "persons"."year" as "persons.year", "persons"."firstname" as "persons.firstname", "persons"."personid" as "persons.personid", "persons"."address" as "persons.address", "persons"."age" as "persons.age", "persons"."type" as "persons.type", "persons"."city" as "persons.city", "persons"."lastname" as "persons.lastname" 
FROM (select * from "tasks" order by "tasks"."taskname" asc limit $1) as "tasks" 
LEFT JOIN "products_tasks" as "products_tasks" on "tasks"."taskid" = "products_tasks"."taskid" 
LEFT JOIN "products" as "products" on "products"."productid" = "products_tasks"."productid" 
LEFT JOIN "persons" as "persons" on "tasks"."executorid" = "persons"."personid" or "tasks"."qaid" = "persons"."personid" order by "tasks"."taskname" asc limit $2
```
vs running it with aliasing:
```sql
SELECT "a"."executorid" as "a.executorid", "a"."taskname" as "a.taskname", "a"."taskid" as "a.taskid", "a"."completed" as "a.completed", "a"."qaid" as "a.qaid", "b"."productname" as "b.productname", "b"."productid" as "b.productid", "c"."year" as "c.year", "c"."firstname" as "c.firstname", "c"."personid" as "c.personid", "c"."address" as "c.address", "c"."age" as "c.age", "c"."type" as "c.type", "c"."city" as "c.city", "c"."lastname" as "c.lastname", "c"."year" as "c.year", "c"."firstname" as "c.firstname", "c"."personid" as "c.personid", "c"."address" as "c.address", "c"."age" as "c.age", "c"."type" as "c.type", "c"."city" as "c.city", "c"."lastname" as "c.lastname" 
FROM (select * from "tasks" as "a" order by "a"."taskname" asc limit $1) as "a" 
LEFT JOIN "products_tasks" as "d" on "a"."taskid" = "d"."taskid" 
LEFT JOIN "products" as "b" on "b"."productid" = "d"."productid" 
LEFT JOIN "persons" as "c" on "a"."executorid" = "c"."personid" or "a"."qaid" = "c"."personid" order by "a"."taskname" asc limit $2"
```

They are both very long, but the second one is at least a more to the point.

I have also added some error handling for when a table name or column name contains a dot, while this is valid implementing it to fully work in Budibase would be difficult, it breaks bindings, breaks Knex, and requires a lot of raw implementation to put in place, it is broken today but simply not explained.

image

### Testing
The testing of this feature was a bit of a slog, but it comes in two forms:
1. All existing SQL tests that we have use this system fully. I didn't see much point in adding more end to end tests as the tests we already have cover quite a few different areas of the aliasing.
3. Usually I am not a fan of unit tests, but in this area it can be quite useful to dig into specific edge cases/cover exactly what is happening under the hood of a complex section. I've added a few tests based on query JSON generated by the frontend, and how that converts to SQL that should be ran against Postgres.


packages/server/scripts/integrations/postgres/reset.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 docker-compose down
 docker volume prune -f
+docker volume rm postgres_pg_data
```
samwho, line 4: You could instead run `docker-compose down -v` to delete all volumes from the compose YAML.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -101,6 +104,26 @@ function buildFilters(
...
+function removeRelationships(
+  rowId: string,
+  table: Table,
+  isManyToMany: boolean,
+  colName?: string
+) {
+  const tableId = table._id!
+  const filters = buildFilters(rowId, {}, table)
+  // safety check, if there are no filters on deletion bad things happen
+  if (Object.keys(filters).length !== 0) {
```
samwho, line 116: I've come across this in my review without any context on how it's used, but it strikes me as strange that this function just doesn't do anything under some circumstances. Should it throw an error or some such to let the caller know it didn't run?

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -101,6 +104,26 @@ function buildFilters(
...
+    const op = isManyToMany ? Operation.DELETE : Operation.UPDATE
+    const body = colName && !isManyToMany ? { [colName]: null } : undefined
```
samwho, line 118: It may make for cleaner code to split this out into 2 functions, one for many-to-many and one for, presumably, one-to-many / many-to-one?

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -697,24 +734,46 @@ export class ExternalRequest<T extends Operation> {
...
       }
     }
     await Promise.all(promises)
   }
 
+  async removeRelationshipsToRow(table: Table, rowId: string) {
+    const row = await this.getRow(table, rowId)
+    const related = await this.lookupRelations(table._id!, row)
+    for (let column of Object.values(table.schema)) {
+      if (
```
samwho, line 755: I created a little suite of type-guard functions for this purpose: https://github.com/Budibase/budibase/blob/c60e2bf41cd09483d86000cc7db8d4ccf46da92f/packages/types/src/documents/app/table/schema.ts#L183-L187.

You could, e.g., write:

```javascript
if (!isManyToOne(column)) {
    continue
}
```

And the type-guard on `isManyToOne` would obviate the need for the cast on line 761.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -826,9 +885,21 @@ export class ExternalRequest<T extends Operation> {
...
-    const response = await getDatasourceAndQuery(json)
-    // handle many to many relationships now if we know the ID (could be auto increment)
+    // remove any relationships that could block deletion
+    if (operation === Operation.DELETE && id) {
+      await this.removeRelationshipsToRow(table, generateRowIdField(id))
+    }
+
+    // aliasing can be disabled fully if desired
+    let response
+    if (!env.SQL_ALIASING_DISABLE) {
```
samwho, line 895: nit: double negative conditions read weird, at least to me. I'd flip this:

```javascript
if (env.SQL_ALIASING_DISABLE) {
    // ...
} else {

}
```

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  getAlias(tableName: string) {
+    if (this.aliases[tableName]) {
+      return this.aliases[tableName]
+    }
+    const char = this.character
+    this.aliases[tableName] = char
+    this.tableAliases[char] = tableName
+    this.character = String.fromCharCode(char.charCodeAt(0) + 1)
```
samwho, line 25: You probably want to do 1 of 2 things here:

1. Put a guard in place so that when you get to `"z"`, an error is thrown. The next character is `"{"` and I'm not sure that'd be valid as an alias.
2. Fall back to 2 character aliases, then 3, etc.

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+    return char
+  }
+
+  aliasField(field: string) {
+    const tableNames = this.tableNames
+    if (field.includes(".")) {
```
samwho, line 31: I'm wondering if this is strictly valid. https://github.com/go-gorm/gorm/issues/1280 seems to suggest that `.` is allowed in column names. We may have to be smarter about how we split.

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+      const [tableName, column] = field.split(".")
+      if (tableNames.includes(tableName)) {
+        return `${this.getAlias(tableName)}.${column}`
```
samwho, line 34: I'm also not sure this is strictly valid. Table names and column names can contain all sorts, and will likely need to be quoted to be compatiable with all of our users' tables.

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+      if (tableName) {
+        map[tableName] = this.getAlias(tableName)
+      }
+    }
+    return map
+  }
+
+  async queryWithAliasing(json: QueryJson) {
+    json = cloneDeep(json)
+    const aliasField = (field: string) => this.aliasField(field)
```
samwho, line 73: How come this is required? Why can't you just call `this.aliasField` where it's needed?

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+    for (let tableName of tableNames) {
+      if (tableName) {
+        map[tableName] = this.getAlias(tableName)
+      }
+    }
+    return map
+  }
+
+  async queryWithAliasing(json: QueryJson) {
+    json = cloneDeep(json)
```
samwho, line 72: We do a deep clone here and then proceed to do quite a lot more internal cloning later with spread operators. I don't imagine it's going to be a massive performance hit, but does feel quite wasteful. No real action here, other than if it's easy to do so, it'd be nice for this to be the only copy made of `json`.

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,125 @@
...
+    const map: Record<string, string> = {}
+    for (let tableName of tableNames) {
+      if (tableName) {
+        map[tableName] = this.getAlias(tableName)
+      }
+    }
+    return map
+  }
+
+  async queryWithAliasing(json: QueryJson) {
```
samwho, line 71: Given that this is meant to be the API of this class, it'd be good to fix its return type in place.

packages/server/src/sdk/app/rows/utils.ts
```diff
@@ -1,12 +1,12 @@
...
 import { makeExternalQuery } from "../../../integrations/base/query"
 import { Format } from "../../../api/controllers/view/exporters"
 import sdk from "../.."
 import { isRelationshipColumn } from "../../../db/utils"
 
-export async function getDatasourceAndQuery(json: any) {
+export async function getDatasourceAndQuery(json: QueryJson) {
   const datasourceId = json.endpoint.datasourceId
   const datasource = await sdk.datasources.get(datasourceId)
   return makeExternalQuery(datasource, json)
```
samwho, line 12: Would be nice to give this a type other than `any` if possible.

packages/server/scripts/integrations/postgres/reset.sh
mike12345567: Wasn't aware of this - nice!

packages/server/src/api/controllers/row/alias.ts
mike12345567: Quoting is handled deeper in the SQL layer, we split by `.` and handle quotation.

packages/server/src/sdk/app/rows/utils.ts
mike12345567: That is what this change does.

packages/server/src/integrations/tests/sqlAlias.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    })
+  })
+
+  describe("read", () => {
+    it("should handle basic retrieval with relationships", () => {
+      const queryJson = getJson("basicFetchWithRelationships.json")
+      let query = new Sql(SqlClient.POSTGRES, limit)._query(queryJson)
+      expect(query).toEqual({
+        bindings: [relationshipLimit, limit],
+        sql: `select "a"."year" as "a.year", "a"."firstname" as "a.firstname", "a"."personid" as "a.personid", "a"."address" as "a.address", "a"."age" as "a.age", "a"."type" as "a.type", "a"."city" as "a.city", "a"."lastname" as "a.lastname", "b"."executorid" as "b.executorid", "b"."taskname" as "b.taskname", "b"."taskid" as "b.taskid", "b"."completed" as "b.completed", "b"."qaid" as "b.qaid", "b"."executorid" as "b.executorid", "b"."taskname" as "b.taskname", "b"."taskid" as "b.taskid", "b"."completed" as "b.completed", "b"."qaid" as "b.qaid" from (select * from "persons" as "a" order by "a"."firstname" asc limit $1) as "a" left join "tasks" as "b" on "a"."personid" = "b"."qaid" or "a"."personid" = "b"."executorid" order by "a"."firstname" asc limit $2`,
```
samwho, line 31: Would be nice to format these as multiline strings.

packages/server/src/integrations/base/sql.ts
```diff
@@ -648,6 +686,18 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
     }
     return results.length ? results : [{ [operation.toLowerCase()]: true }]
   }
+
+  log(query: string, values?: any[]) {
```
samwho, line 690: How concerned should we be here about logging sensitive customer information?

packages/server/src/sdk/app/rows/utils.ts
samwho: Sorry, I mean the return type. My editor still says it's `Promise<any>` unless I'm missing something.

packages/server/src/integrations/base/sql.ts
mike12345567: This won't be turned on by default - it is functionality that will mainly be used in self host environments to help users debug issues (has been requested in the past, being able to see queries BB is actually making).

packages/server/src/api/controllers/row/alias.ts
mike12345567: Old habit - I do this sometimes when writing classes in JS, avoids the state of `this` being lost (a horrible thing that can happen) but isn't an issue in TS - will remove.

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: I agree this is a bit weird - but don't really want to make a change here as part of this PR - throwing an error could be risky, I believe its valid right now for this function to do nothing if there is no relationships to remove.

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -22,7 +28,13 @@ export default class AliasTables {
...
     this.aliases[tableName] = char
     this.tableAliases[char] = tableName
-    this.character = String.fromCharCode(char.charCodeAt(0) + 1)
+    this.character =
+      char.substring(0, char.length - 1) +
+      String.fromCharCode(char.charCodeAt(char.length - 1) + 1)
+    // reached end of characters, extend number of characters used
+    if (this.character.charAt(this.character.length - 1) === "z") {
+      this.character = new Array(this.character.length + 1).fill("a").join("")
+    }
```
samwho, line 37: This isn't quite correct. It'll never increase the earlier letters past `"a"`. I had a play with this in a script and came up with this:

```javascript
class CharSequence {
  static alphabet = "abcdefghijklmnopqrstuvwxyz";

  constructor() {
    this.counters = [0];
  }

  get character() {
    return this.counters.map((i) => CharSequence.alphabet[i]).join("");
  }

  next() {
    for (let i = this.counters.length - 1; i >= 0; i--) {
      if (this.counters[i] < CharSequence.alphabet.length - 1) {
        this.counters[i]++;
        return;
      }
      this.counters[i] = 0;
    }
    this.counters.unshift(0);
  }
}

const seq = new CharSequence();

for (let i = 0; i < 100; i++) {
  seq.next();
  console.log(seq.character);
}
```

Can extend the alphabet if we want to, to ensure maximum smallness on the aliases. For example, we might want to include capital letters as well. I believe SQL treats them as distinct to lowercase.

samwho (changes requested):

packages/server/src/api/controllers/row/alias.ts
mike12345567: Ah good catch - was still playing around with this when I logged off last night! Preference is to keep it to lowercase, just to avoid ambiguity, I think this is a pretty big edge case, as if a query includes minimum 26 different tables, its already going to be problematic for Budibase to deal with (performance wise).

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -590,7 +641,7 @@ export class ExternalRequest<T extends Operation> {
         continue
       }
       const isMany = field.relationshipType === RelationshipType.MANY_TO_MANY
-      const tableId = isMany ? field.through : field.tableId
+      const tableId = isMany ? field.through! : field.tableId!
```
adrinr, line 644: This should not be needed after we did our great type refactor. We should check why it's not picking it up

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -609,7 +660,7 @@ export class ExternalRequest<T extends Operation> {
         },
       })
       // this is the response from knex if no rows found
-      const rows = !response[0].read ? response : []
+      const rows: Row[] = response?.[0].read ? [] : (response as Row[])
```
adrinr, line 663: Do you need this `as` cast?

adrinr (commented):
Nice job!

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,158 @@
...
+  constructor() {
+    this.counters = [0]
+  }
+
+  get character() {
+    return this.counters.map(i => CharSequence.alphabet[i]).join("")
+  }
+
+  next() {
+    for (let i = this.counters.length - 1; i >= 0; i--) {
```
adrinr, line 24: Should this next not return the character, so we are sure we don't retrieve the same one twice?

packages/server/src/api/controllers/row/alias.ts
```diff
@@ -0,0 +1,158 @@
...
+  QueryJson,
+  SearchFilters,
+  Table,
+  Row,
+  DatasourcePlusQueryResponse,
+} from "@budibase/types"
+import { getDatasourceAndQuery } from "../../../sdk/app/rows/utils"
+import { cloneDeep } from "lodash"
+
+class CharSequence {
```
adrinr, line 11: This seems pretty critical. Do we have tests for it, specially around what happen when it goes over all the alphabet characters?

mike12345567: Yeap - there is a test in the sqlAlias.spec.ts for this!

packages/server/src/api/controllers/row/alias.ts
mike12345567: Good idea - safer that way!

packages/server/src/integrations/tests/sqlAlias.spec.ts
```diff
@@ -164,14 +164,23 @@ describe("Captures of real examples", () => {
...
-      expect(alias).toEqual("aaay")
+      expect(alias).toEqual("cv")
+    })
+  })
+
+  describe("check some edge cases", () => {
+    it("should handle table names/columns with dots in them", () => {
+      const tableNames = ["hello.world", "foo.bar.baz"]
+      const aliasing = new AliasTables(tableNames)
+      const aliased = aliasing.aliasField("hello.world.field")
```
samwho, line 182: In the SQL, the field would never appear like this. It would be `"hello.world"."field"` or some such. I'd modify this test case to be like that, and also have one that uses grave accents instead of double quotes.

packages/server/src/sdk/app/datasources/plus.ts
```diff
@@ -3,12 +3,33 @@ import {
...
   IntegrationBase,
   Schema,
+  Table,
 } from "@budibase/types"
 import * as datasources from "./datasources"
 import tableSdk from "../tables"
 import { getIntegration } from "../../../integrations"
 import { context } from "@budibase/backend-core"
 
+function checkForSchemaErrors(schema: Record<string, Table>) {
```
samwho, line 13: Do we want to check and make sure first that we haven't got any datasources in prod today that would fail this validation?

mike12345567: I've tested this - even if they do exist, they are 100% not supported. Lots of parts of the system simply do not function if a column or table name contains full stops - if they exist they are in a completely broken state already.

samwho (approved):

samwho (approved):

adrinr (approved):
🚀

PR #13129, opened 2024-02-23
Set correct devDependency for sdk

adrinr:
## Description
We are missing a devDependency that is blocking the [npm releases](https://github.com/Budibase/budibase-deploys/actions/runs/8009591926/job/21878654085).


Ghrehh:
Already open here I think: https://github.com/Budibase/budibase/pull/13120

adrinr:
Already solved in here: https://github.com/Budibase/budibase/pull/13120

PR #13133, opened 2024-02-26
Help popout email support license validation check

ConorWebb96:
## Description
This pr fixes the help menu email support option. Currently its set to business, this changes it so it only shows locked for plans like team, pro and free account types. Additionally it also now displays the correct text if the user doesn't have email support available to them. Premium instead of Business which was shown previously.

## Screenshots
**Before**
Screenshot 2024-02-26 at 09 15 45

**After**
Screenshot 2024-02-26 at 09 21 19

## Launchcontrol
Small logic amend to help popup, this allows email support access for the correct plan types.

### DIFF
--- a/packages/builder/src/components/common/HelpMenu.svelte
+++ b/packages/builder/src/components/common/HelpMenu.svelte
@@ -52,7 +52,7 @@
         <div class="icon">
           <FontAwesomeIcon name="fa-solid fa-play" />
         </div>
-        <Body size="S">Budibase University {isPremiumAndAbove}</Body>
+        <Body size="S">Budibase University</Body>
       </a>
       <div class="divider" />
       {#if isEnabled(TENANT_FEATURE_FLAGS.LICENSING)}
--- a/packages/builder/src/components/common/HelpMenu.svelte
+++ b/packages/builder/src/components/common/HelpMenu.svelte
@@ -2,11 +2,10 @@
   import FontAwesomeIcon from "./FontAwesomeIcon.svelte"
   import { Popover, Heading, Body } from "@budibase/bbui"
   import { isEnabled, TENANT_FEATURE_FLAGS } from "helpers/featureFlags"
-  import { auth } from "stores/portal"
-  import { emailSupportCheck } from "helpers/planTitle"
+  import { licensing } from "stores/portal"
+  import { isPremiumAndAbove } from "helpers/planTitle"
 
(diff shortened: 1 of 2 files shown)


packages/builder/src/helpers/planTitle.js
```diff
@@ -25,3 +25,9 @@ export function getFormattedPlanName(userPlanType) {
   }
   return `${planName} Plan`
 }
+
+export function emailSupportCheck(userPlanType) {
```
melohagan, line 29: This check isn't necessarily restricted to email support, so you could rename the function to `isPremiumAndAbove`, and then have a one line return statement in the function body.

packages/builder/src/components/common/HelpMenu.svelte
```diff
@@ -1,11 +1,12 @@
 <script>
   import FontAwesomeIcon from "./FontAwesomeIcon.svelte"
   import { Popover, Heading, Body } from "@budibase/bbui"
-  import { licensing } from "stores/portal"
   import { isEnabled, TENANT_FEATURE_FLAGS } from "helpers/featureFlags"
+  import { auth } from "stores/portal"
```
melohagan, line 5: What's the difference between using the `licensing` store and the `auth` store?

melohagan (approved):
LGTM! Just a couple of small comments.

packages/builder/src/helpers/planTitle.js
melohagan: Also just wondering if Pro and Team plan types are given premium support or not?

ConorWebb96: PRO and Teams I believe are older plans and didn't have it then. I've asked sales and they agree that they shouldn't have access to it until they move to the new premium plan.

packages/builder/src/components/common/HelpMenu.svelte
ConorWebb96: Not really sure on this, you can output the same thing using either. Same character count which I found freaky. 

The reason I used auth over licensing is due to seeing it being done and used elsewhere in the codebase. `usage.svelte` uses it for the plan type.

```
$licensing?.license.plan.type
$auth.user?.license.plan.type
```

I've asked one of the engineers and they have said to use licensing over auth as its newer and going forward thats the store that will be used to get the plan types.

I'll update it to use this.

PR #13136, opened 2024-02-26
Modernise application.spec.ts

samwho:
## Description

I starting working on a bug to do with custom roles and user groups, which led me to writing some initial tests, which led me to `application.spec.ts` (because the problem is in the `fetch` for applications, I think).

I noticed these tests were written using the old helpers, and the code they were testing was missing types in a lot of places. This PR is an effort to improve the situation.

No business logic changes are in this PR, it's entirely cleanup and typing.


packages/server/scripts/test.sh
```diff
@@ -3,12 +3,12 @@ set -e
 
 if [[ -n $CI ]]
 then
-  export NODE_OPTIONS="--max-old-space-size=4096 --no-node-snapshot"
+  export NODE_OPTIONS="--max-old-space-size=4096 --no-node-snapshot $NODE_OPTIONS"
```
samwho, line 6: This allows VSCode's "auto attach" feature to work. It passes values into `NODE_OPTIONS` that allows it to attach a debugger to any node process started from a VSCode terminal.

packages/server/src/api/controllers/application.ts
```diff
@@ -196,17 +201,19 @@ export async function fetchAppDefinition(ctx: UserCtx) {
...
-  let application = await db.get<any>(DocumentType.APP_METADATA)
+  let application = await db.get<App>(DocumentType.APP_METADATA)
   const layouts = await getLayouts()
   let screens = await getScreens()
   const license = await licensing.cache.getCachedLicense()
 
   // Enrich plugin URLs
   application.usedPlugins = objectStore.enrichPluginURLs(
-    application.usedPlugins
+    application.usedPlugins || []
```
samwho, line 216: `usedPlugins` was not a member of the `App` type, so you'll notice I've added it. Slightly uncertain about that change, but without it this code doesn't type-check correclty.

packages/server/src/api/routes/application.ts
```diff
@@ -4,7 +4,6 @@ import * as deploymentController from "../controllers/deploy"
 import authorized from "../../middleware/authorized"
 import { permissions } from "@budibase/backend-core"
 import { applicationValidator } from "./utils/validators"
-import { importToApp } from "../controllers/application"
```
samwho, line 7: Unused.

packages/worker/scripts/test.sh
```diff
@@ -4,10 +4,10 @@ set -e
 if [[ -n $CI ]]
 then
   # Running in ci, where resources are limited
-  echo "jest --coverage --maxWorkers=2  --forceExit --bail"
-  jest --coverage --maxWorkers=2  --forceExit --bail
+  echo "jest --coverage --maxWorkers=2  --forceExit --bail $@"
```
samwho, line 7: Allows you to run `yarn test <filename>`

qa-core/src/internal-api/tests/tables/tables.spec.ts
```diff
@@ -13,17 +13,6 @@ describe("Internal API - Table Operations", () => {
...
-  async function createAppFromTemplate() {
-    return config.api.apps.create({
-      name: generator.word(),
-      url: `/${generator.word()}`,
-      useTemplate: "true",
-      templateName: "Near Miss Register",
-      templateKey: "app/near-miss-register",
-      templateFile: undefined,
-    })
-  }
```
samwho, line 25: Unused.

packages/server/src/api/controllers/application.ts
```diff
@@ -174,14 +177,16 @@ export const addSampleData = async (ctx: UserCtx) => {
   ctx.status = 200
 }
 
-export async function fetch(ctx: UserCtx) {
+export async function fetch(ctx: UserCtx<null, App[]>) {
```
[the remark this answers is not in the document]
mike12345567: `void` is generally preferred if its the lack there-of we expect - its more specific in terms of meaning absolutely nothing/not there - very similar to `undefined` but stricter.

packages/server/src/api/controllers/application.ts
```diff
@@ -47,6 +47,9 @@ import {
   PlanType,
   Screen,
   UserCtx,
+  CreateAppRequest,
+  FetchAppDefinitionResponse,
+  type FetchAppPackageResponse,
```
mike12345567, line 52: What is the need for the `type` keyword here?

packages/server/src/tests/utilities/api/application.ts
```diff
@@ -1,17 +1,96 @@
...
+    const result = await this.request
+      .post(`/api/applications/${appId}/sync`)
+      .set(this.config.defaultHeaders())
+      .expect("Content-Type", /json/)
+      .expect(statusCode)
+    return result.body
+  }
+
   getRaw = async (appId: string): Promise<Response> => {
+    // While the appPackage endpoint does take an :appId parameter, it doesn't
```
mike12345567, line 85: We probably should update this (not in this PR though!) to use the `appId` in the URL if no header is provided.

packages/types/src/documents/app/app.ts
```diff
@@ -73,3 +75,30 @@ export interface AppFeatures {
 export interface AutomationSettings {
   chainAutomations?: boolean
 }
+
+export interface CreateAppRequest {
```
mike12345567, line 79: Generally we separate the `document` definitions and the API request definitions, these should probably be moved to the `types/src/api/web` layer.

mike12345567 (approved):
LGTM - would update the `UserCtx<void` instances for consistency, and move the API typing into the API layer, but massive improvement to this test, much more readable, and nice to have this stuff better typed!

packages/server/src/api/controllers/application.ts
samwho: This is something my editor seems to be when it auto-imports a type. Will look into configuring it not to do it.

packages/backend-core/src/objectStore/buckets/plugins.ts
```diff
@@ -6,7 +6,7 @@ import { Plugin } from "@budibase/types"
 
 // URLS
 
-export function enrichPluginURLs(plugins: Plugin[]) {
+export function enrichPluginURLs(plugins: Plugin[] | undefined): Plugin[] {
```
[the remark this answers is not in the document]
samwho: 🙈

packages/server/src/api/controllers/application.ts
```diff
@@ -58,23 +61,23 @@ import * as appMigrations from "../../appMigrations"
 async function getLayouts() {
   const db = context.getAppDB()
   return (
-    await db.allDocs(
+    await db.allDocs<Layout>(
```
adrinr, line 64: 😍

adrinr (approved):
Thx for the types!

PR #13146, opened 2024-02-28
Properly type TestConfiguration.ts

samwho:
## Description

I've done my best to properly type `TestConfiguration.ts` here, removing as many `any`s as I could. This has fanned out into quite a lot of files to try and correctly plumb in all of the right types, so this will be a little tricky to review. Sorry about that.

On this journey I did have to make some additions and changes to things in `@budibase/types`. Please give those extra scrutiny, especially the changes to `Layout`, which was very barebones.


packages/server/src/api/controllers/automation.ts
```diff
@@ -207,7 +210,7 @@ export async function find(ctx: UserCtx) {
   ctx.body = await db.get(ctx.params.id)
 }
 
-export async function destroy(ctx: UserCtx) {
+export async function destroy(ctx: UserCtx<void, DocumentDestroyResponse>) {
```
mike12345567, line 213: Even if its identical, we would normally alias these very direct types to something specific to the controller, incase it is ever expanded/changed - e.g. `DeleteAutomationResponse`.

packages/server/src/api/controllers/layout.ts
```diff
@@ -1,9 +1,9 @@
 import { EMPTY_LAYOUT } from "../../constants/layouts"
 import { generateLayoutID, getScreenParams } from "../../db/utils"
 import { events, context } from "@budibase/backend-core"
-import { BBContext, Layout } from "@budibase/types"
+import { BBContext, Layout, UserCtx } from "@budibase/types"
 
-export async function save(ctx: BBContext) {
+export async function save(ctx: UserCtx<Layout, Layout>) {
```
mike12345567, line 6: Similar to above comment - just for clarity we would normally have a type alias for `LayoutRequest` and `LayoutResponse` - NAB at all - just the convention we've normally stuck with.

packages/server/src/api/routes/tests/automation.spec.ts
```diff
@@ -394,7 +394,7 @@ describe("/automations", () => {
     it("deletes a automation by its ID", async () => {
       const automation = await config.createAutomation()
       const res = await request
-        .delete(`/api/automations/${automation.id}/${automation.rev}`)
+        .delete(`/api/automations/${automation._id}/${automation._rev}`)
```
mike12345567, line 397: Odd that this was working at all!

mike12345567 (approved):
Major improvement - looks like there was some painful bits to solve here!

packages/server/src/api/routes/tests/automation.spec.ts
samwho: Right?!

Issue #5769, opened 2022-05-08
Cannot delete a renamed column with a mysql datasource


FlaminWrap:
**Hosting**

- Self
  - Method: Budi-CLI
  - Budibase Version: 1.0.145
  - App Version: 1.0.145

**Describe the bug**
If you try and delete a column that has been renamed in a MYSQL datasource 

**To Reproduce**
1. Create a MYSQL datasource
2. Create a table
3. Create a column called `Test1`
4. Rename the column to `Test2`
5. Delete the column
6. A notification will display `Column Test2 has been deleted`
7. The column has not been deleted

**Expected behavior**
The column is deleted

**Screenshots**
GIF
![SQL](https://user-images.githubusercontent.com/97764630/167306845-17263416-20e4-4410-90d3-2c8db923db26.gif)
Browser console
![image](https://user-images.githubusercontent.com/97764630/167307090-f4b60b4d-c158-4844-a4f3-55d4edbfa3dc.png)
phpMyAdmin screenshot
![image](https://user-images.githubusercontent.com/97764630/167306832-478a0396-28f4-44f1-9aff-5a557145bc0f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 100.0.4896.127

**Additional context**
It looks like BudiBase does not rename the column. The screenshot of phpMyAdmin is after the gif.
Also renaming the column to what it was previously allows you to delete it.

melohagan:
This seems to be related: https://github.com/knex/knex/issues/3358

The **renameColumn** method doesn't appear to work for MySQL.

melohagan:
As mentioned in the past, you cannot rename columns in MySQL through knex (Postgres was fine). It may be worth disabling renaming columns in Budibase for MySQL datasource in the short-term.

PR #13147, opened 2024-02-28
FIX: cannot delete a renamed column with a mysql datasource

melohagan:
## Description
Knex wasn't handling the renaming of a MySQL table column. Added a raw query for this scenario.

Also fixed another bug I noticed where you couldn't delete a SQL column. This was simply an incorrect boolean check on `isIgnoredType`.

Unit tests added.

## Addresses
- https://github.com/Budibase/budibase/issues/5769


mike12345567 (approved):
LGTM - nice typing improvements as well!

Issue #10339, opened 2023-04-18
[BUDI-6874] Clicking on Design tab while Data tab loads does not redirect to the Design page


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.5.2
  * App Version: 2.5.2

**Describe the bug**
Clicking into an app and then quickly clicking Design does not take you to the Design page.

**To Reproduce**

1. Click on an app in the Admin portal
2. Quickly click on Design while the Data section is loading
3. See that Design has been selected, but you remain on the Data page

**Expected behavior**
Clicking on *Design* while *Data* loads should either take you to the Design page, or it should not select Design

**Screenshots**

![side_panel.gif](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/a93fb700-adc0-4f9e-930c-c2e0c7b01d6d/de282d13-cde5-47be-a5e9-f6887e163e00)

<sub>From [SyncLinear.com](https://synclinear.com) | [BUDI-6874](https://linear.app/budibase/issue/BUDI-6874/clicking-on-design-tab-while-data-tab-loads-does-not-redirect-to-the)</sub>

melohagan:
Going to re-open this one just because Joe did happen to mention it recently.

PR #13152, opened 2024-02-28
FIX: clicking on design tab while data tab loads does not navigate correctly

melohagan:
## Description
If the packages was still loading when clicking one of the top level tabs, the incorrect active tab would be shown and the navigation wouldn't occur.
Code has been refactored to ensure that the `goto` function is always fired and doesn't return early.

This was testing by adding the following test code into the `getPackage()` function:

```js
 const delay = ms => new Promise(res => setTimeout(res, ms))
 await delay(3000)
```

## Addresses
- https://github.com/Budibase/budibase/issues/10339

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/42ef1f41-2330-4131-b843-fa9b6b17b674


deanhannigan (approved):
LGTM!

PR #13157, opened 2024-02-29
Unifying how we send requests to the server in packages/server tests.

samwho:
## Description

This isn't finished just yet, I've only converted a few of the descendants of `TestAPI`, but I wanted to get a PR up early to see if there's appetite for the idea before converting everything.

I have 2 goals with this:

1. Make it such that calling backend APIs happens in a consistent way throughout our tests.
2. Output more helpful error messages when a request fails.

For the first point, we have a number of different ways that backend APIs are called and asserted against today. It's on the writer of the particular API client to decide how the user of the client does assertions. Most allow you to pass in something like `{ expectedStatus: 400 }` for when 200 isn't expected, but lots don't, and others have a different name for the key. This PR solves that.

For the 2nd point, the error message from a failed API call today gives essentially no information. All you're told is that there was a status mismatch. You don't get the body of the response, and you don't see any exceptions thrown inside the server that might have caused the problem. This PR solves that.

Before:

CleanShot 2024-02-29 at 09 24 52@2x

After:

CleanShot 2024-02-29 at 09 23 54@2x


packages/server/src/tests/utilities/api/base.ts
```diff
@@ -1,17 +1,164 @@
...
+    return false
+  }
+  const attachedFile = file as AttachedFile
+  return (
+    Object.hasOwnProperty.call(attachedFile, "file") &&
+    Object.hasOwnProperty.call(attachedFile, "name")
+  )
+}
+
+export interface Expectations {
```
mike12345567, line 23: When this is done - i'd probably move these into `@budibase/types` into some testing types - we should probably reuse these ideas in say the worker tests as well.

packages/server/src/api/controllers/permission.ts
```diff
@@ -16,9 +20,9 @@ import {
 import { removeFromArray } from "../../utilities"
 import sdk from "../../sdk"
 
-const PermissionUpdateType = {
-  REMOVE: "remove",
-  ADD: "add",
+enum PermissionUpdateType {
```
adrinr, line 23: nit: can this be a `const enum`?

packages/server/src/api/controllers/query/index.ts
```diff
@@ -17,10 +17,12 @@ import {
   QueryPreview,
   QuerySchema,
   FieldType,
-  type ExecuteQueryRequest,
```
adrinr, line 20: Why removing the type keyword?

packages/server/src/api/controllers/query/index.ts
```diff
@@ -295,7 +297,7 @@ export async function preview(ctx: UserCtx) {
 }
 
 async function execute(
-  ctx: UserCtx<ExecuteQueryRequest, ExecuteQueryResponse | Row[]>,
+  ctx: UserCtx<ExecuteQueryRequest, ExecuteQueryResponse | any[]>,
```
adrinr, line 300: Can we execute this with something that is not a row?

packages/server/src/api/routes/tests/backup.spec.ts
```diff
@@ -19,11 +19,8 @@ describe("/backups", () => {
 
   describe("/api/backups/export", () => {
     it("should be able to export app", async () => {
-      const { body, headers } = await config.api.backup.exportBasicBackup(
-        config.getAppId()!
-      )
+      const body = await config.api.backup.exportBasicBackup(config.getAppId()!)
       expect(body instanceof Buffer).toBe(true)
-      expect(headers["content-type"]).toEqual("application/gzip")
```
adrinr, line 26: This check might be important. We should probably keep it

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -712,11 +712,6 @@ export default class TestConfiguration {
     return this.api.row.save(tableId, config)
   }
 
-  async getRow(tableId: string, rowId: string): Promise<Row> {
```
adrinr, line 715: 🚀

packages/types/src/api/web/app/permission.ts
```diff
@@ -14,3 +14,21 @@ export interface GetResourcePermsResponse {
 export interface GetDependantResourcesResponse {
   resourceByType?: Record<string, number>
 }
+
+export interface AddedPermission {
+  _id: string
+  rev?: string
```
adrinr, line 20: This seems strange. They should be both _id & _rev or id & rev

adrinr (commented):
Good refactor. Added some comments

packages/server/src/api/controllers/query/index.ts
samwho: @mike12345567 asked about them in another PR and I got rid of them there, so I did the same here. Not really any good reason other than that. It wasn't causing problems.

packages/server/src/api/controllers/query/index.ts
samwho: `Row[]` here was actually incorrect, as far as I can tell. Queries don't return `Row`s, they can return absolutely anything. It's raw SQL, we have little to no control over it. `Row` extends `Document`, and both types specify some optional fields that have types that may be wrong. It's entirely possible for a user to have a table with an `_id` column that's a date, for example.

packages/server/src/api/routes/tests/backup.spec.ts
samwho: It's still happening, just lower down the stack. It's in `BackupAPI`, because all callers of this endpoint should assert this type on success.

packages/types/src/api/web/app/permission.ts
samwho: Hmm, seems to have been an oversight on my part. I have made it optional and everything passes type checking. Good spot.

packages/server/src/api/controllers/query/index.ts
mike12345567: I think its purely for dropping the import at runtime, but types is always included anyway, especially with using esbuild fairly sure the `type` keyword is meaningless.

packages/server/src/api/controllers/query/index.ts
mike12345567: This is true, technically we could be a little more specific with a row response of `Record<string, any>` - it will always be an array of objects returned, even if the query returns say a string or number, we will wrap it in an object like `[{ value: 1 }]`.

mike12345567 (approved):
LGTM! Major improvement to testing, much easier to add APIs to the test section now!

packages/types/src/api/web/app/permission.ts
adrinr: I meant the `_`. But as far as I see, this was already like this :/

adrinr (approved):

PR #13163, opened 2024-02-29
Doc writethrough cache

adrinr:
## Description
Implementing a doc writethrough to allow persisting parallel keys into the single document with multiple keys. This will be used for heavy writes such as SCIM logging, where we want to keep adding data into a document without worrying about conflicts.

How does this work?
1. Define a document reference passing its database and doc id.
2. Patch some object, for example `{"a":123, "b":345}`
3. This will queue a new message with the document and data in it
4. A worker will be pulling all the messages and persisting it. If a 409 error is thrown, it means that 2 processes were updating the same document in parallel, so we will retry until it eventually gets persisted

This required the messages to be retried infinitely, and we cannot guarantee the order of this persistence. If different patches are updating the same document, so it should only be used for now for adding keys to a document, instead of editing existing ones


samwho (dismissed):

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+  }
+
+  private async shouldUpdateDb(cache: BaseCache) {
+    const cacheItem = await cache.withCache(this.docInfoCacheKey, null, () =>
+      this.makeNextWriteInfoItem()
+    )
+    return Date.now() >= cacheItem.nextWrite
+  }
```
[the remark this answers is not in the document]
mike12345567: This is probably a point that applies to all our writethrough caches, each cache instance should probably have an interval, say 60 seconds, where it checks if any of the keys it has need to be written, despite no write having occurred in the last 60 seconds.

Our other writethrough caches work on the assumption that the writes will be coming at a steady rate consistently, so you can use the write as a trigger, but that doesn't work if you have bursty write rates, which SCIM logs (what this will be used for) is likely to have.

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  constructor(
+    db: Database,
+    docId: string,
+    writeRateMs: number = DEFAULT_WRITE_RATE_MS
+  ) {
+    this.db = db
+    this._docId = docId
+    this.writeRateMs = writeRateMs
+    this.docInfoCacheKey = `${this.docId}:info`
```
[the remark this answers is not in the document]
adrinr: This explains why I had the db name in it before I removed it thinking it was not needed 😝

packages/backend-core/src/db/instrumentation.ts
```diff
@@ -38,6 +38,13 @@ export class DDInstrumentedDatabase implements Database {
     })
   }
 
+  docExists(id: string): Promise<boolean> {
```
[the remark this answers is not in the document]
adrinr: This required some extra work with the default `exists`, but I agree it is cleaner

packages/backend-core/src/cache/docWritethrough.ts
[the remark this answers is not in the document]
adrinr: Yup, I am checking this. This would require some background work to do these checks...

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,113 @@
...
+              this.docInfoCacheKey,
+              this.makeNextWriteInfoItem()
+            )
+          }
+        }
+      )
+
+      if (!lockResponse.executed) {
+        console.log(`Ignoring redlock conflict in write-through cache`)
+      }
```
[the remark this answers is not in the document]
adrinr: This is quite attached to the internal usage of our redlock. I changed it to make it more verbose and resilient

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,140 @@
...
+    doc = await db.get(docId)
+  } catch {
+    doc = { _id: docId }
+  }
+
+  const keysToPersist = await cache.keys(`${cacheKeyPrefix}:data:*`)
+  for (const key of keysToPersist) {
+    const data = await cache.get(key, { useTenancy: false })
+    doc[data.key] = data.value
+  }
```
samwho, line 84: Redis actually has primitives for dealing with hash maps, e.g. https://redis.io/commands/hgetall/. We should probably create wrappers for using this instead of shoe-horning into lots of different keys.

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,113 @@
...
+    const cacheItem = await cache.withCache(this.docInfoCacheKey, null, () =>
+      this.makeNextWriteInfoItem()
+    )
+    return Date.now() >= cacheItem.nextWrite
+  }
+
+  private async storeToCache(cache: BaseCache, data: Record<string, any>) {
+    for (const [key, value] of Object.entries(data)) {
+      const cacheKey = this.docId + ":data:" + key
+      await cache.store(cacheKey, { key, value }, undefined)
```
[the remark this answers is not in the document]
adrinr: I am checking this in a different PR

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,140 @@
...
+      if (
+        lockResponse.reason !==
+        locks.UnsuccessfulRedlockExecutionReason.LockTakenWithTryOnce
+      ) {
+        console.error("Error persisting docWritethrough", {
+          data: message.data,
+        })
+        throw "Error persisting docWritethrough"
+      }
+      console.log(`Ignoring redlock conflict in write-through cache`)
```
samwho, line 56: Could you explain what's going on here? I don't know what `locks.UnsuccessfulRedlockExecutionReason.LockTakenWithTryOnce` means in practice, and it seems scary to ignore any errors in this code.

packages/backend-core/src/cache/docWritethrough.ts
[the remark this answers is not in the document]
samwho: I left another comment elsewhere that talks about using Redis's hash map primitives for this. I think doing separate requests per key introduces race conditions.

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,140 @@
...
+        cacheKeyPrefix: this.cacheKeyPrefix,
+      },
+      {
+        delay: this.writeRateMs,
+        jobId: this.cacheKeyPrefix,
+        removeOnFail: true,
+        removeOnComplete: true,
+      }
+    )
+  }
```
samwho, line 132: Does reusing the same `jobId` overwrite the job in Bull? If so, won't this continually push the job further forward in time? That would make the name `writeRateMs` misleading, because that's not the rate we're writing to the DB at, it is the amount of time that must pass _without_ a write for the cache to flush to the BD. It also creates a situation where particularly hot documents might not get written to the DB for a long time.

If reusing the `jobId` doesn't overwrite the job, won't we be writing to the DB for every `patch`? Is that not the situation we're trying to avoid in the first place?

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,140 @@
...
+
+docWritethroughProcessorQueue.process(async message => {
+  const { tenantId, cacheKeyPrefix } = message.data
+  await context.doInTenant(tenantId, async () => {
+    const lockResponse = await locks.doWithLock(
+      {
+        type: LockType.TRY_ONCE,
+        name: LockName.PERSIST_WRITETHROUGH,
+        resource: cacheKeyPrefix,
+        ttl: 15000,
```
samwho, line 38: There's no guarantee that the Bull job queue is being processed. If it has an outage and jobs aren't processed for 15 seconds, won't we lose data?

packages/backend-core/src/cache/docWritethrough.ts
adrinr: Good thinking, I added this utility in here! https://github.com/Budibase/budibase/pull/13197

packages/backend-core/src/cache/docWritethrough.ts
adrinr: Removed, after our last discussion

packages/backend-core/src/cache/docWritethrough.ts
adrinr: No, job id is used as idempotency key. So no 2 jobs with the same id might exist. Without the flag this would be "forever", so the same message cannot be reprocessed, but with the `removeOnX` we clean them, allowing newer messages to be queued.
Even so, this was removed after our last discussion

packages/backend-core/src/cache/docWritethrough.ts
adrinr: The message would retry on that case, but the latest discussed solution does not require it

packages/backend-core/src/cache/base/index.ts
```diff
@@ -23,6 +23,18 @@ export default class BaseCache {
     return client.keys(pattern)
```
adrinr, line 23: This file is exposing some of the redis utils into our cache. I ended up not using it, but it is worth keeping

packages/backend-core/src/cache/tests/docWritethrough.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+
+        await waitForQueueCompletion()
+        expect(await db.get(documentId)).toEqual(
+          expect.objectContaining(patches)
+        )
+      })
+    })
+
+    // This is not yet supported
+    it.skip("patches will execute in order", async () => {
```
adrinr, line 263: This test is added for a future change when this will be supported

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,85 @@
...
+)
+
+class DocWritethroughProcessor {
+  init() {
+    docWritethroughProcessorQueue.process(async message => {
+      try {
+        await this.persistToDb(message.data)
+      } catch (err: any) {
+        if (err.status === 409) {
+          logWarn(`409 conflict in doc-writethrough cache`)
```
mike12345567, line 30: We've had issues with stuff like this in the past spamming - this is a valid situation, I think we would only want to log warn if it happens too many times in a row, e.g. if its on its max attempt or something.

adrinr:
We know that this might be causing issues with data replication in clusters, but given that its usage will be only internal for now, I would suggest merging as it is, monitoring, and discussing a solution that will work overall (not only for this use case)

packages/backend-core/src/cache/docWritethrough.ts
```diff
@@ -0,0 +1,85 @@
...
+  docId: string
+  data: Record<string, any>
+}
+
+export const docWritethroughProcessorQueue = createQueue<ProcessDocMessage>(
+  JobQueue.DOC_WRITETHROUGH_QUEUE,
+  {
+    jobOptions: {
+      // We might have plenty of 409, we want to allow running almost infinitely
+      attempts: Number.MAX_SAFE_INTEGER,
```
mike12345567, line 18: As with the error - I might tune this to be a little lower, then if we reach the max attempts me log warn, that way we have some mechanism to track if things are getting clogged up.

packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -70,7 +70,15 @@ export class DatabaseImpl implements Database {
     DatabaseImpl.nano = buildNano(couchInfo)
   }
 
-  async exists() {
+  exists(docId?: string) {
+    if (docId === undefined) {
```
mike12345567, line 74: should this be a `== undefined` to cover null as well just incase?

packages/backend-core/src/cache/docWritethrough.ts
adrinr: We don't know what is a valid max attempts for this. I log the attempt count, we can probably monitor it for a while and then tweak it?

packages/backend-core/src/db/couch/DatabaseImpl.ts
adrinr: This is the typescript way to support overriding. We will have:
1. db.exists()
2. db.exists(docId)

We should never have null, in this case it should be treated as docId in my opinion

packages/backend-core/src/cache/docWritethrough.ts
adrinr: Removed and logging the attempt count on the error

packages/backend-core/src/db/couch/DatabaseImpl.ts
mike12345567: Yeah thats fair - I guess it shouldn't be a valid input!

packages/backend-core/src/cache/docWritethrough.ts
mike12345567: My fear with such a high number is just the queue getting overloaded, theres basically no way to clear it down with this, if the queue gets full we can't wait for it to settle, we'll need to go into Redis and empty it out manually. At least if the number of attempts is something say like `100` - eventually it will clear out, even if that involves dumping some events.

adrinr: Ok, I set it to 100 and we can then see if errored messages are piling up

mike12345567 (approved):

PR #13165, opened 2024-02-29
Rebuild table schema when adding new column to get externalType

melohagan:
## Description
I noticed an edge-case when renaming MySQL columns. Because older version of MySQL require the data type when renaming a column, `externalType` has to be present. 
This is problematic when you create a MySQL column in Budibase, and then try to rename the column afterwards without doing a table fetch/reloading the app. 

To address this, after a new column has been added, the schema is rebuilt for the current table so that the `externalType` of the newly created column is pulled down. 

## Addresses
- https://github.com/Budibase/budibase/issues/5769

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/4f0e6f2d-af29-4ddb-91db-c06dd04e5d58


samwho (changes requested):
This is an excellent find, and a really good PR. Would it be possible to test this new behaviour?

mike12345567 (commented):
Looks good - @samwho is right though - a test for something like this would be worthwhile!

melohagan:
@samwho I have added an integration test suite for MySQL, using the existing Postgres one as a template. An additional test has also been added to cover this fix!

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+  })
+
+  afterAll(async () => {
+    await databaseTestProviders.mysql.stop()
+  })
+
+  beforeEach(async () => {
+    async function createAuxTable(prefix: string) {
+      return await config.createTable({
+        name: `${prefix}_${generator.word({ length: 5 })}`,
```
samwho, line 71: This may not be random enough, and could cause flakes. Any reason to not use a UUID here?

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+      relationshipType: RelationshipType.MANY_TO_ONE,
+    }
+    manyToManyRelationshipInfo = {
+      table: await createAuxTable("m2m"),
+      fieldName: "manyToMany",
+      relationshipType: RelationshipType.MANY_TO_MANY,
+    }
+
+    primaryMySqlTable = await config.createTable({
+      name: `p_${generator.word({ length: 5 })}`,
```
samwho, line 108: Similar here, I'd prefer to see UUIDs used if it's possible to do so. We've seen tests get quite flaky because they were generating random words of quite a limited length.

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+
+  async function createPrimaryRow(opts: {
+    rowData: PrimaryRowData
+    createForeignRows?: {
+      createOneToMany?: boolean
+      createManyToOne?: number
+      createManyToMany?: number
+    }
+  }) {
+    let { rowData } = opts as any
```
samwho, line 203: How come this `as any` is needed?

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+      tableId: primaryMySqlTable._id,
+      ...rowData,
+    })
+
+    return { row, foreignRows }
+  }
+
+  async function createDefaultMySqlTable() {
+    return await config.createTable({
+      name: generator.word({ length: 10 }),
```
samwho, line 270: Same here as other instances of this.

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+      updatedAt: expect.any(String),
+      entities: expect.any(Object),
+    })
+  })
+
+  describe("POST /api/:tableId/rows", () => {
+    const createRow = (tableId: string | undefined, body: object) =>
+      makeRequest("post", `/api/${tableId}/rows`, body)
+
+    describe("given than no row exists", () => {
```
samwho, line 347:
```suggestion
    describe("given that no row exists", () => {
```

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
...
+      _rev: expect.any(String),
+      createdAt: expect.any(String),
+      updatedAt: expect.any(String),
+      entities: expect.any(Object),
+    })
+  })
+
+  describe("POST /api/:tableId/rows", () => {
+    const createRow = (tableId: string | undefined, body: object) =>
+      makeRequest("post", `/api/${tableId}/rows`, body)
```
samwho, line 345: You have access to `config` in these tests, so you should also have access to `config.api.row.create` and such. I know you didn't write this initially, and you're copying what's in `postgres.spec.ts`, but the stuff in there is quite old and crufty compared to how we write tests today. It might not be the best idea to copy it. Going to tag @mike12345567 to get his opinion.

packages/server/src/integration-test/mysql.spec.ts
melohagan: I actually had to reduce this from 6 -> 5 because MySQL was complaining about the identifier lengths. MySQL is limited to 64 characters, so longer join table identifiers were problematic.

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -0,0 +1,1289 @@
+import fetch from "node-fetch"
```
mike12345567, line 1: Would it be possible to avoid this much duplication? I would argue, its probably worthwhile just writing your test case for mysql into this for now, if at some point in the future we want to have a full test suite like the postgres test for all SQL datasources I think we'll need to write something generic which handles it, rather than a separate test case for each.

My preference for this would be to keep the postgres as the all singing all dancing test for SQL, then having the `mysql` test case here purely to test edge case stuff that differs specifically due to MySQL implementation (anywhere, we've needed to do an if mySQL do X).

melohagan: I can do - there may be a couple other tests worth keeping, for instance MySQL uses `DATABASE` instead of `SCHEMA` so there was a small implementation difference in a couple of the tests, but most of them were basically duplicates

packages/server/src/integration-test/mysql.spec.ts
samwho: Looking at the usages of `createAuxTable`, I would personally get rid of the `prefix` argument and make the name always a UUID. A UUID is 36 chars long and shouldn't ever clash.

Been wondering about this for a while so decided to write a little script to see.

```javascript
const uuid = require("uuid")
const chance = require("chance")
const c = new chance.Chance()

const words = new Set()
const uuids = new Set()

const N = 10000000
for (let i = 0; i < N; i++) {
  words.add(c.word({ length: 5 }))
  uuids.add(uuid.v4())
}

function uniqueValues(set) {
  return `${set.size}/${N} (${((set.size / N) * 100).toFixed(2)}%)`
}

console.log(`words uniqueness: ${uniqueValues(words)}`)
console.log(`uuids uniqueness: ${uniqueValues(uuids)}`)
```

Output:

```
words uniqueness: 834493/10000000 (8.34%)
uuids uniqueness: 10000000/10000000 (100.00%)
```

Sort of a flawed test, it looks like the number of possible unique values of the 5 letter words `chance` generates tops out around ~800,000, so comparing against such a large N is silly. But the point is the same: UUIDs will collide way less often.

samwho (approved):
Other than wanting to avoid potential uniqueness problems in test, this change looks good to me. Good work, Mel.

PR #13166, opened 2024-02-29
Updating datasource controller typing

mike12345567:
## Description
Updating datasource controller types, this branched out a little bit to removing as many anys as possible, and allowing `RowValue` to correctly be returned from the `allDocs` function of the DB.

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -15,10 +15,14 @@ import {
   FieldType,
   RelationshipFieldMetadata,
   SourceName,
+  UpdateDatasourceRequest,
   UpdateDatasourceResponse,
   UserCtx,
   VerifyDatasourceRequest,
   VerifyDatasourceResponse,
+  Table,
+  RowValue,
+  DynamicVariable,
 } from "@budibase/types"
 import sdk from "../../sdk"
 import { builderSocket } from "../../websockets"
@@ -90,8 +94,10 @@ async function invalidateVariables(
   existingDatasource: Datasource,
   updatedDatasource: Datasource
 ) {
-  const existingVariables: any = existingDatasource.config?.dynamicVariables
-  const updatedVariables: any = updatedDatasource.config?.dynamicVariables
+  const existingVariables: DynamicVariable[] =
+    existingDatasource.config?.dynamicVariables
+  const updatedVariables: DynamicVariable[] =
+    updatedDatasource.config?.dynamicVariables
   const toInvalidate = []
 
   if (!existingVariables) {
@@ -103,9 +109,9 @@ async function invalidateVariables(
     toInvalidate.push(...existingVariables)
   } else {
     // invaldate changed / removed
-    existingVariables.forEach((existing: any) => {
+    existingVariables.forEach(existing => {
       const unchanged = updatedVariables.find(
-        (updated: any) =>
+        updated =>
           existing.name === updated.name &&
           existing.queryId === updated.queryId &&
           existing.value === updated.value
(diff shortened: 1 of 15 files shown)


packages/server/src/api/controllers/datasource.ts
```diff
@@ -118,24 +124,32 @@ async function invalidateVariables(
...
   await invalidateVariables(baseDatasource, ctx.request.body)
 
   const isBudibaseSource =
     baseDatasource.type === dbCore.BUDIBASE_DATASOURCE_TYPE
 
-  const dataSourceBody = isBudibaseSource
-    ? { name: ctx.request.body?.name }
+  const dataSourceBody: Datasource = isBudibaseSource
+    ? {
+        name: ctx.request.body?.name,
```
mike12345567, line 140: This is the only major logical change, everything else is typing. These types should normally accompany a Budibase datasource type, unsure why they weren't here originally.

packages/types/src/documents/pouch.ts
```diff
@@ -5,15 +5,15 @@ export interface RowValue {
   deleted: boolean
 }
 
-export interface RowResponse<T extends Document> {
+export interface RowResponse<T extends Document | RowValue> {
   id: string
   key: string
   error: string
   value: T | RowValue
```
samwho, line 12: Would this now change to just `value: T`?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -90,8 +94,10 @@ async function invalidateVariables(
   existingDatasource: Datasource,
   updatedDatasource: Datasource
 ) {
-  const existingVariables: any = existingDatasource.config?.dynamicVariables
-  const updatedVariables: any = updatedDatasource.config?.dynamicVariables
+  const existingVariables: DynamicVariable[] =
+    existingDatasource.config?.dynamicVariables
+  const updatedVariables: DynamicVariable[] =
+    updatedDatasource.config?.dynamicVariables
```
samwho, line 100: Given that these use `?.`, don't they need a `|| []` at the end?

packages/server/src/api/controllers/datasource.ts
```diff
@@ -254,9 +268,9 @@ export async function destroy(ctx: UserCtx) {
   if (datasource.type === dbCore.BUDIBASE_DATASOURCE_TYPE) {
     await destroyInternalTablesBySourceId(datasourceId)
   } else {
-    const queries = await db.allDocs(getQueryParams(datasourceId))
+    const queries = await db.allDocs<RowValue>(getQueryParams(datasourceId))
```
samwho, line 271: Given the use of `getQueryParams`, would this not be `db.allDocs<Query>`?

I was thinking recently about whether it would be possible to use conditional types to map `DocumentType` to the returned type here, but haven't gotten any further than dreaming about it.

packages/server/src/api/controllers/integration.ts
```diff
@@ -1,7 +1,10 @@
 import { getDefinition, getDefinitions } from "../../integrations"
 import { SourceName, UserCtx } from "@budibase/types"
 
-const DISABLED_EXTERNAL_INTEGRATIONS = [SourceName.AIRTABLE]
+const DISABLED_EXTERNAL_INTEGRATIONS = [
+  SourceName.AIRTABLE,
+  SourceName.BUDIBASE,
```
samwho, line 6: How come this has been added?

mike12345567: The Budibase source wasn't in the source name list, technically its a valid source name, but don't want that key ever being returned from the integrations API.

packages/server/src/api/controllers/datasource.ts
mike12345567: No - as there isn't an `include_docs` in the request, there won't be a document returned. Instead its just fetching metadata about the documents.

samwho: Ahh I totally missed that. My bad.

samwho (approved):

Issue #12785, opened 2024-01-15
Custom CSS drawer not the correct width


joebudi:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Cloud
  * Tenant ID: budimaster

**Describe the bug**
Custom CSS drawer not spanning the correct width.

![CleanShot 2024-01-15 at 15 22 53@2x](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/58d18e9d-dda8-4a23-a322-15d0109e955b/2a4ff0d1-c77d-412c-8aa1-24fc896ab743?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC81OGQxOGU5ZC1kZGE4LTRhMjMtYTMyMi0xNWQwMTA5ZTk1NWIvMmE0ZmYwZDEtYzc3ZC00MTJjLThhYTEtMjRmYzg5NmFiNzQzIiwiaWF0IjoxNzA1MzQxNTMxLCJleHAiOjMzMjc1OTAxNTMxfQ.tOFsFZvSptsoP23XVNSkZwWybHZ95N45_gNNTI2kNZ0)

joebudi:
This may be the result of reducing the width of the screens panel.

joebudi:
Hey @melohagan . Here's a video that may help:
https://vimeo.com/903603845/c4514562b0?share=copy

melohagan:
Thanks @joebudi Andy was able to reproduce as well 👍

melohagan:
To add to this one, the screen titles don't reveal more even when the panel is widened. 

![Screenshot 2024-01-31 at 13 17 59](https://github.com/Budibase/budibase/assets/101575380/fe805461-7eba-4079-b9bd-688ca9f41043)
*Plenty of space to show the full screen title*

PR #13169, opened 2024-02-29
Live binding evaluation and drawer rework

aptkingston:
## Description
This PR adds multiple improvements and new features when working with bindings and drawers.

## Values for bindings
For most cases, you'll now be able to see the real values of any bindings. This works anywhere inside the design section and with formula columns in the data section. Automations do not support this yet.

The values shown are pretty printed and syntax-highlighted, to make them easily readable. The colour palette is the same palette as we use for code mirror, so it's familiar to users.

Hovering over a binding in the binding side panel:
image

Using the keyboard to select a binding in the code editor:
image

## Live evaluation of expressions
There's a new side panel option within drawers which will evaluate your expression, using the real values for the binding, and show you the result. This gives you a preview of what the overall result is, without having to save and close the drawer and inspect the value in the real application.

These values are also pretty printed and syntax highlighted. Both Handlebars and JS are supported.

This new tab is selected using the "play" icon.

Valid expressions being evaluated:
image
image

Invalid expressions are clearly shown. Currently we don't have much in the way of useful error messages, but we can hopefully improve that in future.

An invalid expression:
image

## Drawer UI rework
The drawer component has been given a lick of paint. The new drawer is rounded, floats on top of the screen, has a cleaner title and fewer borders. There's less internal padding around the various elements, resulting a cleaner overall look.

Drawer titles have also been updated across the board to be more useful, and this is the main reason so many files are edited by this PR.

There have been multiple improvements to the code mirror editor, improving consistency of colours, spacing and fonts across the board.

There are a couple of new features that these drawers also support.

### Stacking drawers
There are a few instances where we need to open more than one drawer, when "going deeper" into a certain component settings. Currently, opening more drawers completely replaces the existing drawer, meaning you lose context of where you are.

Going forwards, drawers now stack by giving the illusion of pushing back the previous drawer - keeping the previous title visible - with exponential falloff to support infinite depths of drawers. Stacking support is fully animated.

Here's a quick example of stacking and closing multiple drawers:

https://github.com/Budibase/budibase/assets/9075550/fe3050ba-04a3-494b-bba9-846427d80894

### Expanding drawers into modal
Certain drawers, particularly ones for editing bindings, can now be expanding into a modal. This gives the user a lot more space to work with. Drawers can be toggled into modals using the expand icon.

Drawers that have been expanded into modals still fully support stacking. All transitions are smoothly animated.

Here's an example of expanding drawers:

https://github.com/Budibase/budibase/assets/9075550/f176f6a1-a360-4228-a9f5-926c2a7be17a

### Intelligent drawer width
Drawers can now also observe another DOM node's dimensions and copy it's width and position. This is particularly useful in the design section where we have a resizable left panel. Drawers will now respect the size of this panel and resize themselves appropriately.

Resizing drawers:

https://github.com/Budibase/budibase/assets/9075550/cd8332a7-d354-4845-8db1-b380c0d17eea

### Theme support
The new code mirror changes, and drawer changes of course, are fully supported across any theme.

Here's a few screenshots of the different themese:
image
image
image
image

### Addresses
- https://linear.app/budibase/issue/BUDI-8004/live-evaluation-testing-bindings-research-spike
- https://linear.app/budibase/issue/BUDI-8006/binding-drawer-tidy-up
- https://linear.app/budibase/issue/BUDI-8005/binding-drawer-stack-drawers-and-responsive
- https://linear.app/budibase/issue/BUDI-8013/live-evaluation-ui
- https://linear.app/budibase/issue/BUDI-7901/custom-css-drawer-not-the-correct-width
- https://linear.app/budibase/issue/BUDI-8008/live-evaluation-return-values-for-bindings
- https://github.com/Budibase/budibase/issues/12785

## Feature branch env
[Feature Branch Link](http://fb-cheeks-lab-day-binding-eval.fb.qa.budibase.net)


packages/bbui/src/Actions/position_dropdown.js
```diff
@@ -38,8 +39,13 @@ export default function positionDropdown(element, opts) {
       styles = customUpdate(anchorBounds, elementBounds, styles)
     } else {
       // Determine vertical styles
-      if (align === "right-outside") {
-        styles.top = anchorBounds.top
+      if (align === "right-outside" || align === "left-outside") {
```
shogunpurple, line 42: NAB: constants instead of strings

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -377,12 +376,13 @@
 <div class="fields">
   {#each schemaProperties as [key, value]}
     {#if canShowField(key, value)}
+      {@const label = getFieldLabel(key, value)}
```
shogunpurple, line 379: TIL `@const` - that's a really clean way to compute the value once 👌

packages/client/src/components/app/blocks/RowExplorer.svelte
```diff
@@ -3,25 +3,35 @@
...
+  const context = getContext("context")
+  const { generateGoldenSample } = getContext("sdk")
 
   let listDataProviderId
   let listRepeaterId
+
+  // Provide additional data context for live binding eval
+  export const getAdditionalDataContext = () => {
+    const rows = get(context)[listDataProviderId]?.rows || []
+    const goldenRow = generateGoldenSample(rows)
```
shogunpurple, line 30: What is the "Golden Sample" in this context?

packages/client/src/index.js
```diff
@@ -84,6 +84,12 @@ const loadBudibase = async () => {
       } else {
         dndStore.actions.reset()
       }
+    } else if (type === "request-context") {
+      const { selectedComponentInstance } = get(componentStore)
+      const context = selectedComponentInstance?.getDataContext()
+      eventStore.actions.dispatchEvent("provide-context", {
+        context: JSON.stringify(context),
```
shogunpurple, line 91: Any scenario where this may fail?

shogunpurple (approved):
giphy

Awesome work cheeks. This takes the general binding experience to a new level altogether, and surprisingly in a lot less code than I would have anticipated. Been playing around with it on the feature branch and it's working really well so far - loving the new drawer too. 

I noticed in automations you don't have access to `trigger.*` in the live eval. I imagine this is just due to it not being available in context until you actually run the automation?

Code looks good to me - just a few minor comments/questions but nothing blocking.

packages/client/src/index.js
aptkingston: I don't think so - `null`, `undefined` etc stringify just fine, in case it was nullish. The only thing that would fail would be objects with cyclic references I think, but there aren't any instances of that in our codebase.

I wonder if you could force it to fail by using a custom JS datasource that deliberately includes cyclic references. Probably worth try/catching this just in case. Good call 👍

packages/client/src/components/app/blocks/RowExplorer.svelte
aptkingston: This was a utility I added to try and provide the "best" example of a row we can find. It's commented a bit at the implementation of `generateGoldenSample` but it will basically loop through all rows and take values for every column that isn't nullish.

It basically just accounts for the fact that if we just used the first row for live eval, there's a chance that the row you pick is mostly empty, so can't provide live eval for some fields.

This alleviates that by searching all rows to find a value for every column, where possible. It also prefers longer values too for things like arrays - basically just tries to give you a nice example for every column.

aptkingston:
Cheers for the feedback!

> I noticed in automations you don't have access to `trigger.*` in the live eval. I imagine this is just due to it not being available in context until you actually run the automation?

Yea I mentioned that in the description, but basically live eval isn't supported for automations right now. We'd need to revisit the whole UX of how and when you run tests, because we'd need to force them to run a test to get values to use, and then continually run more tests as you add steps to get values for the new steps.

To be honest it might even be better to start saving the test results inside the automations themselves, to try and cache as much of that as possible.

In any case, I definitely felt that was beyond the scope of this PR so chose not to feature creep this any more since it already touches quite a lot.

aptkingston:
@shogunpurple addressed the issue of the code mirror plugin showing "undefined" when no context exists as well. I had already updated our own code to ignore live eval when context is missing, but forgot about the code mirror plugin so good catch 👍

deanhannigan (changes requested):
LGTM! One minor note on the CodeMirror updates but the refinements to the Drawer are a huge improvement!

- General Feedback
  - Speed up the resize animation a fraction on the drawer?
  - Binding popover copy and paste
    - It can be a little fiddly to highlight and copy values.
    - A copy behaviour on the hover/suggestion itself would be useful
  - Copy difference in the bindings [search bar](https://www.figma.com/file/pG5v2XS18QeiGemZq103V3/Builder-designs?type=design&node-id=4305-24010&mode=design&t=yhKzn2Ku6FCcMpb9-0)


- `BUG` - Drawer validation messaging missing for invalid HBS bindings
  - The Save button is properly disabled but there is no feedback
  - There used to be red text to indicate an issue.
  - Could possibly switch to preview and display an error?

- `BUG` - Device binding values not updating in the Autocomplete suggestions
  - Updates fine in the binding side panel AND live evaluation window
  - Steps
    - Add an Headline, change the app theme and refresh the page
    - Change app theme again to something different
    - Open the 'Text' binding drawer for the headline
    - Add `{{ App.theme }}` and it will resolve to the old value in the autocomplete suggestion
    
- `BUG` - Autocomplete popover default UX missing from the right-hand bindings panel
  - No hover behaviour or default copy when empty
  - Currently displays the unmapped binding in the autocomplete suggestions.

- `BUG` - Command window
  - z-index places the drawer above the command menu in all modes.
  - Especially awkward if in fullscreen mode.

- `BUG` - Line wrapping in CodeEditor
  - When expanding the binding panel, the line wrapping appears to break. 
text_wrap

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -298,27 +280,194 @@
 </div>
 
 <style>
-  .code-editor.handlebars :global(.cm-content) {
-    font-family: var(--font-sans);
+  /* Editor */
```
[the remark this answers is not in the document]
aptkingston: I basically just found it very difficult to reason about what styles were applied where, and what they affected, because some were defined in CSS and some were defined as part of the theme config. There were inconsistencies around which style overode I found the theme config much more awkward to use, both because you need to pass in a bunch of props to ensure styles are generated, and also just the camel case naming and string value syntax of the styles built up.

I found it much easier to have a single source of truth which we can style with proper CSS (and it's also technically more performant, not that that's a consideration here but it's free).

It was also very difficult to actually reason which styles affected what, which is why I've now broken it down into sections and commented appropriately, to try and help people in future be able to make style changes without trying to work out the consequences.

aptkingston:
Thanks for the in-depth review @deanhannigan!

Addressed those points as follows:

> Speed up the resize animation a fraction on the drawer?

Actually not easy to do since it's inside the `module` context script, so I've had to leave this as is. Not a huge deal as I don't think people will often resize the side panel while a drawer is open (and you could argue the animation is a feature haha).

> Drawer validation messaging missing for invalid HBS bindings

Good catch. After checking with Andy, we want to allow people to save now no matter what, so I've removed the disabled state of the save button. This just means that we allow people to save invalid HBS expressions now, which is consistent with JS (since you've always been able to save garbage if you want to with JS). The preview tab should make it clearer whether things are invalid.

> Device binding values not updating in the Autocomplete suggestions

This was an existing issue - it's just that the completions passed into the `CodeEditor` are not reactive so do not update once the editor is mounted. We just never saw this before since they never changed. I've fixed this by remounting the code editor when completions change, since it's all we can do. Remounting isn't an issue since it only will ever happen once, when the drawer opens and context is refreshed.

> Autocomplete popover default UX missing from the right-hand bindings panel

I didn't think the difference was an issue, but after spending a while working on different ways to make things consistent, I've opted for removing the readable bindings from popovers in the code mirror completion menu (when context exists) rather than adding it into the side panel. The main reason for this was an existing bug where the code mirror autocomplete readable bindings are currently shown as HBS bindings always, even when using JS, and attempting to properly colour the readable binding is so much work that it's not worth it. Basically when there is context now, only values will be shown for both (or no popover if no value). If there is no context (which is only automations now) we will still show readable bindings and descriptions inside the code mirror autocomplete popover (the existing functionality which is bugged because its HBS only) and nothing at all in the side panel. That means there is a small difference in automations only, but we'll be working on this in future to add context so I'm not worried about that for now.

> Command window

Another existing issue (in that it's currently behind modals), but I've gone ahead and fixed it as part of this PR.

> Line wrapping in CodeEditor

Already was fixed in my upstream snippets PR but forgot to say, but well found 👍

deanhannigan (approved):
LGTM!

PR #13173, opened 2024-03-01
Snippets

aptkingston:
## Description
This PR adds the concepts of snippets. Snippets are reusable pieces of JavaScript which you can use anywhere you write JS in Budibase - bindings, formula columns and automations.

## What is a snippet
Snippets are written like a normal JS binding -  they must `return` a value. They can be functions (if you want to make a useful utility to resuse) or just a piece of text, or blob of JSON.

## Creating and editing snippets
Snippets are created, updated and deleted by using the new snippets tab when editing bindings. You'll have access to snippets anywhere you currently write JS.

image

Snippets are searchable to help manage a large list:
image

Clicking on a snippet will insert it into your expression. Hovering over a snippet will show you the snippet code, so you can get a quick reminder of what it does or what parameters it takes.

image

You can create a new snippet by pressing the + icon in the snippets side panel. This will open a new drawer and let you enter a snippet name, and write the snippet.

**Once created, you can't change the name of your snippet. You can edit the code, but the name cannot be changed.**

Snippet names need to be valid Javascript function names, and this validation is shown when entering your name. Casing doesn't matter though - that's up to you - but you can't have any spaces or special characters.

image

Naming validation:
image

You can edit an existing snippet using the edit icon in the snippet list. You can also delete the snippet from here.

image
image

## Using snippets
You can insert snippets by clicking them in the snippet list, but there are faster ways to use them. The syntax for a snippet is simplty
```js
snippet.name
```
Where "name" is the name of your snippet.
If your snippet returns a function, then you'll need to call it like a function:
```js
snippet.square(16)
```
If your snippet is just a value, then simply use it like a normal variable:
```js
return "Sorry, the limit is " + snippets.itemLimit + " items!"
```

There is new autocompletion for snippets, that makes it extremely fast to use them. Just start typing "snippets" to see a list of all your snippets:
image

Or even better, just start typing the name of the snippet itself:
image

### A utility to square a number
Since snippets are just JS, there are many ways you could write a simple function to square a number. The following examples all do the exact same thing, but just serve to highlight the flexiblity of snippets.
```js
return function(num) {
  return num * num
}
```
```js
return num => num * num
```
```js
function square(num) {
  return num * num
}
return square
```
### Constants (text, JSON, etc)
Just like a normal JS binding, snippets can just return static values.
```js
return "Lorem ipsum dolor sit amet"
```
```js
return [
  {
    name: "Bob",
    age: 36
  }
]
```

## Empty state
If you have access to snippets but haven't created any yet, you'll see this placeholder prompting you to create a snippet:
image

## Licensing
Snippets are a paid feature. If you do not have access to snippets, you'll be prompted to upgrade:
image

If you were on a paid plan, used snippets, then drop down to the free plan, existing snippets will continue to work. You won't be able to see, create, edit or delete your snippets anymore, but they will continue to function so that your app does not break.

Snippet autocompletion will also be disabled if you no longer have access to snippets.

Addresses:
- https://linear.app/budibase/issue/BUDI-8066/js-snippets

## Feature branch env
[Feature Branch Link](http://fb-cheeks-snippets-poc.fb.qa.budibase.net)


packages/backend-core/src/context/mainContext.ts
```diff
@@ -122,10 +122,10 @@ export async function doInAutomationContext<T>(params: {
   automationId: string
   task: () => T
 }): Promise<T> {
-  const tenantId = getTenantIDFromAppID(params.appId)
+  await ensureSnippetContext()
```
aptkingston, line 125: Since snippets always need to be part of automation context, it feels correct to me to add the snippets into context here. Since snippet context works by mutating the current context rather than creating a new layer, we aren't creating 2 new contexts or anything by doing this.

packages/backend-core/src/context/mainContext.ts
```diff
@@ -281,6 +281,27 @@ export function doInScimContext(task: any) {
...
+  const db = getAppDB()
+  if (db) {
+    const app = await db.get<App>(DocumentType.APP_METADATA)
+    snippets = app.snippets
+  }
+
+  // Always set snippets to a non-null value so that we can tell we've attempted
+  // to load snippets
+  ctx.snippets = snippets || []
+}
```
aptkingston, line 303: Snippet context works similarly to how we cache the VM isolate - it mutates the current context rather than creating a new layer. The main motivation for this is that there are a large number of places where things like formula columns need to be processed, and attempting to wrap all of those in a new context layer to add snippets while avoiding fetching the snippet definitions multiple times would be extremely difficult.

Doing it this way makes it trivial to ensure that we're only every fetch snippet definitions once per action / API call / automation, as we can simply call `ensureSnippetContext` where needed and not worry about doubling up on app DB reads.

I appreciate this is different than the normal approach though, so I'm open to feedback.

packages/server/src/api/controllers/static/templates/preview.hbs
```diff
@@ -89,6 +90,7 @@
       window["##BUDIBASE_HIDDEN_COMPONENT_IDS##"] = hiddenComponentIds
       window["##BUDIBASE_USED_PLUGINS##"] = usedPlugins
       window["##BUDIBASE_LOCATION##"] = location
+      window["##BUDIBASE_SNIPPETS##"] = snippets
```
aptkingston, line 93: While working inside the builder, snippets are passed across the iframe boundary to ensure that new snippets are immediately available after being updated. Otherwise, we would have to fetch the app definition inside the preview after every change.

packages/server/src/jsRunner/vm/isolated-vm.ts
```diff
@@ -98,6 +98,20 @@ export class IsolatedVM implements VM {
...
+      const snippetDefinitions = ${JSON.stringify(snippets || [])};
+      ${snippetsSource};
+      snippets = snippets.default;
+    `)
+    script.runSync(this.vm, { timeout: this.invocationTimeout, release: false })
+    new Promise(() => {
+      script.release()
+    })
+    return this
+  }
```
aptkingston, line 113: Snippet definitions are stringified and embedded into the isolate. We later read these definitions using the new snippets IVM bundle, to execute at runtime. By compiling the definitions into the isolate rather than adding them via context, like the client does, we can reduce the overall size of context and benefit from the fact we cache the isolate.

packages/server/src/threads/automation.ts
```diff
@@ -625,6 +625,7 @@ export async function executeInThread(job: Job<AutomationData>) {
   })
 
   return await context.doInAppContext(appId, async () => {
+    await context.ensureSnippetContext()
```
aptkingston, line 628: External triggers are the only automation type that I can see where we don't wrap them in automation context, so I'm manually adding them here (just like the other contexts are manually added here).

aptkingston:
Would appreciate some feedback, particularly on the server changes! Sorry for the requests @samwho and @adrinr, but with Mike off it would be great if you'd both be able to quickly review those server / backend-core / context changes specifically. Thanks in advance 🙏

packages/string-templates/package.json
```diff
@@ -12,7 +12,9 @@
       "import": "./dist/bundle.mjs"
     },
     "./package.json": "./package.json",
-    "./test/utils": "./test/utils.js"
+    "./test/utils": "./test/utils.js",
+    "./iife": "./src/iife.js",
+    "./snippet": "./src/helpers/snippet.js"
```
samwho, line 17: I'm guessing this is for testing purposes because it appears to be the `marked` library again. I would name this more descriptively so future people have a better idea what it's for.

packages/server/src/jsRunner/vm/isolated-vm.ts
```diff
@@ -98,6 +98,20 @@ export class IsolatedVM implements VM {
     return this
   }
 
+  withSnippets(snippets?: Snippet[]) {
```
samwho, line 101: Are we ever loading anything other than the full gamut of snippets? If not, I'd personally have the code to fetch them from the app metadata in here and remove adding the snippets to the request context. Or am I missing something?

packages/string-templates/src/helpers/javascript.js
```diff
@@ -48,14 +49,26 @@ module.exports.processJS = (handlebars, context) => {
...
     // Wrap JS in a function and immediately invoke it.
     // This is required to allow the final `return` statement to be valid.
-    const js = `(function(){${atob(handlebars)}})();`
+    const js = iifeWrapper(atob(handlebars))
 
     // Our $ context function gets a value from context.
     // We clone the context to avoid mutation in the binding affecting real
     // app context.
     const sandboxContext = {
       $: path => getContextValue(path, cloneDeep(context)),
```
samwho, line 58: Random thought while I'm here: do we need this `cloneDeep` here? Accessing anything with `$(...)` in the browser does a full clone of the context, which feels a little wasteful. Could we not take a single deep clone and freeze it instead?

packages/string-templates/src/helpers/javascript.js
```diff
@@ -48,14 +49,26 @@ module.exports.processJS = (handlebars, context) => {
...
       helpers: getJsHelperList(),
+
+      // Proxy to evaluate snippets when running in the browser
+      snippets: new Proxy(
+        {},
+        {
+          get: function (_, name) {
+            // This will error if the snippet doesn't exist, but that's intended
+            const snippet = (context.snippets || []).find(x => x.name === name)
```
samwho, line 67: Is it worth gathering these snippets into an object so that accessing them doesn't require a scan across all snippets? I don't imagine it's going to make the world of difference, but potentially unbounded code makes me nervous.

packages/server/src/jsRunner/bundles/snippets.ts
```diff
@@ -0,0 +1,19 @@
...
+  {
+    get: function (_, name) {
+      // Snippet definitions are injected to the isolate global scope before
+      // this bundle is loaded, so we can access it from there.
+      // https://esbuild.github.io/content-types/#direct-eval for info on why
+      // eval is being called this way.
+      // @ts-ignore
+      // eslint-disable-next-line no-undef
+      const snippet = (snippetDefinitions || []).find(x => x.name === name)
+      return [eval][0](iifeWrapper(snippet.code))
```
samwho, line 16: As I understand it, this means we're going to `eval` the snippet source every time we call it? Are there negative performance implications of that? What would be wrong with properly compiling the JavaScript when we `compileScriptSync` in `withSnippets`?

packages/string-templates/src/helpers/javascript.js
aptkingston: That's actually a really good catch. I think I've always skimmed over this assuming that it will be only a single deep clone - but you're right that since that's inside the function body, it's surely doing a deep clone for every single binding access. If that's correct, this could potentially be a massive performance increase! I'll check and fix this.

packages/server/src/jsRunner/bundles/snippets.ts
aptkingston: Absolutely correct, I don't see a reason that we can't optimise this in advance by both your previous suggestion of gathering into an object, and also updating the compiled script source to do a single eval of each snippet. The benefit for a single evaluation is negligible since you're unlikely to use the same snippet more than once, but since we cache the isolate there are definitely performance gains here for things like formulas.

packages/string-templates/src/helpers/javascript.js
aptkingston: Every little helps! ™️

packages/server/src/jsRunner/bundles/snippets.ts
aptkingston: Ah actually there is a reason - doing it this way will only eval the snippets that are actually used. If we eval them all up front, you're evaluating every single snippet even if you use none of them. I'm not sure if that is worth updating or not then, since you're potentially degrading performance if you have many (or very large) snippets. I could try a best of both worlds approach, where we cache snippets which have already been eval'd, but only eval them when required like this?

packages/server/src/jsRunner/vm/isolated-vm.ts
aptkingston: I would love to have them in here, and did initially, but there is a reason we can't - we can't have an async handlebars helper - so we can't do anything async here no matter what. Fetching the snippets from app metadata requires an async call, unless you can think of a clever way to get around that.

packages/string-templates/package.json
aptkingston: The snippet export is a testing leftover yes, so I'll remove that one. The `iife` export is deliberate and required - do you mean rename that to be more descriptive?

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -94,12 +95,35 @@
...
+    }),
+    {
+      decorations: v => v.decorations,
+    }
+  )
 
-  let plugin = ViewPlugin.define(
+  // Match decoration for snippets
+  const snippetMatchDeco = new MatchDecorator({
+    regexp: /snippets.[^\s(]+/g,
```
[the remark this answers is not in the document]
aptkingston: Certainly do. Good catch.

packages/server/src/jsRunner/bundles/snippets.ts
```diff
@@ -0,0 +1,19 @@
+// @ts-ignore
+// eslint-disable-next-line local-rules/no-budibase-imports
+import { iifeWrapper } from "@budibase/string-templates/iife"
+
+export default new Proxy(
+  {},
+  {
+    get: function (_, name) {
```
adrinr, line 8: Should this not be running in isolated vm? Not sure why another bundle is needed

packages/string-templates/src/helpers/javascript.js
aptkingston: Yep this was invoking it every time. For a JS binding with multiple `$` usages, this will now really help performance. I've pushed this up.

packages/backend-core/src/context/mainContext.ts
```diff
@@ -281,6 +281,27 @@ export function doInScimContext(task: any) {
...
+  // If we've already added snippets to context, continue
+  if (!ctx || ctx.snippets) {
+    return
+  }
+
+  // Otherwise get snippets for this app and update context
+  let snippets: Snippet[] | undefined
+  const db = getAppDB()
+  if (db) {
+    const app = await db.get<App>(DocumentType.APP_METADATA)
```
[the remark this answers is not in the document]
aptkingston: They are per app, yea. Eventually maybe we will add the concept of global snippets, but for now we decided to make them just per app.

packages/server/src/jsRunner/bundles/snippets.ts
aptkingston: I've updated both isolated-vm and the browser to still do evaluation at runtime, but then cache the evaluated snippets for future use. This is the best of both worlds approach that I mentioned above - we only eval snippets that are used, and only once.

packages/server/src/jsRunner/vm/isolated-vm.ts
samwho: Blergh, that's a shame.

packages/shared-core/src/constants/index.ts
```diff
@@ -98,6 +98,7 @@ export enum BuilderSocketEvent {
 export const SocketSessionTTL = 60
 export const ValidQueryNameRegex = /^[^()]*$/
 export const ValidColumnNameRegex = /^[_a-zA-Z0-9\s]*$/g
+export const ValidSnippetNameRegex = /^[a-z-_][a-z0-9-_]*$/i
```
samwho, line 101: I've just realised allowing `-` is incorrect, so this should be:

```suggestion
export const ValidSnippetNameRegex = /^[a-z_][a-z0-9_]*$/i
```

samwho (approved):
Awesome feature addition, great work!

packages/shared-core/src/constants/index.ts
aptkingston: Appreciated 🙏

deanhannigan (changes requested):
- General feedback
  - Should snippets be accessible on the free plan?
    - Ran this past @aptkingston earlier and it should be a paid feature. Being updated.
  - Empty snippets menu, currently blank
    - Add "Add your first snippet.." type message and a button?
    - Similar to [here](https://www.figma.com/file/pG5v2XS18QeiGemZq103V3/Builder-designs?type=design&node-id=4337-26072&mode=design&t=ovMVOtJMtI23gxCE-0)
  - You currently can't edit a snippets name after it has been created. Is this intentional?

`Bug - Fixed`
  - Updates to snippets aren't reflected in the preview. The snippet code does make it to the preview but it will only update the displayed value when the binding itself is modified.
    - Add a snippet and use it in the `text` property binding of a headline e.g `return "some text" + snippets.mySnippet($("someval"))`
    - Verify it appears in the preview, then open the binding drawer and edit/delete only the snippet used and save.
    - The preview still reflects the old value with the snippet.
    - Open the binding drawer and alter the "some text" prefix value to something else.
    - Now the value will update completely

`Bug - Fixed`
  - Different snippet insertion UX between sidebar and autocomplete
  - Autocomplete produces `snippets.MySnippet2()`
  - The binding menu produces `snippets.MySnippet2`
    - This is fine for a snippet like `return "hello"` but the editor placeholder suggests that it should be a function.
    - If you return a fuction and use `snippets.MySnippet2` you get no output.
  - It seems snippets can support static returns or fns. I'm not entirely sure if that is expected
  - I ran this past @aptkingston and it is supported but some kind of detection will be needed to apply the correct output.

`Bug - Fixed`
  - Cannot toggle visibility of drawer bindings list anymore

`Bug - Fixed`
  - Cannot save snippet after attempting to edit it
  - Save button is greyed out

`Bug - Fixed`
  - When creating a formula column in the db, clicking in the text area of the code editor closes the add column popover UX.

packages/server/src/jsRunner/bundles/snippets.ts
adrinr: The idea behind these bundles is injecting 3rd party libraries. As far as I understand, the snippets are already js objects that can be injected directly in the isolate. I think it is fine to inject it as you are, but I think this wrapper could be handled on runtime

aptkingston:
Cheers for the in-depth review @deanhannigan!

> Should snippets be accessible on the free plan?

Restricted to premium.

> Empty snippets menu, currently blank

Were no designs for this, but chatted to Andy and decided to keep the same copy as when you don't have access, but with a new button to create a snippet.

> You currently can't edit a snippets name after it has been created. Is this intentional?

Correct, this is by design and mentioned in the PR description for context.

> Updates to snippets aren't reflected in the preview.

Nice catch, I hadn't considered this. I've updated how snippets are consumed in client apps, so that they pass through the normal context flow and are processed by components. With one slight tweak, components will now intelligently detect if they care about snippets (just like any other context key) and enrich bindings again if so. The result is that editing snippets results in immediate updates in client apps now 🥳 

> Different snippet insertion UX between sidebar and autocomplete

Nice work catching the difference (which I've now standardised to no parentheses) but it's not actually possible to detect the type without evaling the snippet, which we can't do in the browser for safety. Therefore we just need to insert no parentheses, and if it's a function then the user will just need to call it like so. If this tuns out to be a big issue, we can look at potentially evaling the snippets in a sandbox at creation, determining the output type, then saving that as metadata to reference later.

packages/server/src/jsRunner/bundles/snippets.ts
aptkingston: The snippets definitions are stringified and injected in, without a bundle, since they are just primitives. I assumed using a proxy object would require pollyfilling to run in the isolate, which is what happens when this is compiled. I can try injecting this code raw (like I do with the definitions) in case it runs fine. I'd rather leave this as is for now in order to get this released, but it's something I can try after.

deanhannigan (approved):
LGTM! Fantastic, can't wait to see it live!! 🚀

joebudi:
For power users, this is a massive accelerator. Also, the implementation is brilliant. It's a unique UX that I've not experienced but feels perfectly fitting for this piece of work. Great job.

PR #13175, opened 2024-03-01
Relationship picker incorrectly renders selections

melohagan:
## Description
Some outstanding issues that have been fixed. I would appreciate a play around on the feature branch (particularly for SQL database relationships) to make sure this is working well.

- Fixed **Update** type form not presenting the correct value if not within the first 100 rows
- Fixed *Default value* not working if more than one _id was provided.

## Addresses
- https://github.com/Budibase/budibase/issues/11478

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/29a71e67-f41c-4ca3-b3dd-6341b5cd2f42


packages/server/src/sdk/app/rows/search/tests/external.spec.ts
melohagan, line 1: I think this entire test file is not running. I'm not sure why.

samwho: It's because on line 24 it's declared with `define.skip("...`

samwho (approved):
Sadly there's no feature branch because there's a bug that happens when your branch name is longer than 63 characters. We'll get that fixed.

Broadly this looks good to me. The test isn't running because that suite is being skipped, left a comment. Would be good to see about unskipping it.

packages/server/src/sdk/app/rows/search/tests/external.spec.ts
melohagan: Wow missed that! That makes sense, will see if I can unskip it / find out why it was skipped.

PR #13179, opened 2024-03-04
Component Node Collapse/Expand Behaviour Changes

Ghrehh:
Some tweaks to the behaviour of the collapsing and expanding behaviour of the component tree:

- Previously, selecting a component in the preview would temporarily make said component visible in the tree, and would collapse again once you selected a different component. We decided this behaviour was a little strange, so now a component expanded this way will stay expanded once the user clicks outside of the tree.
- Slightly changes the behaviour of the "command - up" "command - down" shortcuts that allow you to move a component up and down the tree, previously if the next/previous sibling accepted children, the moved component would become a child of said sibling, now if the target silbing is collapsed the component will be moved above/below the sibling instead. IMO this is much nicer behaviour and is fairly intuitive, but folks can let me know what they think.
- Moved the store that manages collapsed/expanded components to the builder store directory, I think it makes more sense for it to live there.


packages/builder/src/stores/builder/components.js
```diff
@@ -653,6 +653,18 @@ export class ComponentStore extends BudiStore {
...
+      const targetScreen = get(screenStore).screens.find(
+        screen => screen.id === targetScreenId
+      )
+
+      const componentPathIds = findComponentPath(
+        targetScreen?.props,
+        newComponentId
+      ).map(component => component._id)
+
+      componentTreeNodesStore.expandNodes(componentPathIds)
```
aptkingston, line 666: Could we move all this logic into the component tree nodes store to avoid polluting the component store with it? We could have a nice function like `revealNode` or something that takes in a component ID and does this. We don't need to worry finding the right screen, because it will always be the selected screen.
I also don't think this needs to actually be inside the store update function.

packages/builder/src/stores/builder/components.js
```diff
@@ -803,7 +816,10 @@ export class ComponentStore extends BudiStore {
         // sibling
         const previousSibling = parent._children[index - 1]
         const definition = this.getDefinition(previousSibling._component)
-        if (definition.hasChildren) {
+        if (
+          definition.hasChildren &&
+          componentTreeNodes[`nodeOpen-${previousSibling._id}`]
```
aptkingston, line 821: Could we add a function to the component tree node store to check if a component is visible or not? 
i.e. `componentTreeNodesStore.isNodeExpanded(previousSibling._id)`

That would be a lot nicer than having to use this syntax with the magic string `nodeOpen` all over the place. Think it would be a good idea to hide the implementation details away from the usages.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte
```diff
@@ -130,6 +131,13 @@
       error = event.error || "An unknown error occurred"
     } else if (type === "select-component" && data.id) {
       componentStore.select(data.id)
+
+      const componentPath = findComponentPath(
+        $selectedScreen?.props,
+        data.id
+      ).map(component => component._id)
+
+      componentTreeNodesStore.expandNodes(componentPath)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte:140]: Same as mentioned previously - it would be nice to hide this implementation inside the store and just do `componentTreeNodesStore.revealNode(data.id)`.

aptkingston (commented):
Few suggestions to just clean things up a little 👌

packages/builder/src/stores/builder/components.js
Ghrehh: Implemented this, dunno why I didn't do this in the first place. 👍

packages/builder/src/stores/builder/components.js
Ghrehh: Good idea, done 👍

aptkingston (approved):
Nice work on the changes, much tidier 👌

PR #13159, opened 2024-02-29
Revert "App Preview Skeleton"

Ghrehh:
Reverts Budibase/budibase#12898

Seems to be causing issues in production

PR #13180, opened 2024-03-04
Un-revert Skeleton Loader PR

Ghrehh:
## Description
Reimplementation of https://github.com/Budibase/budibase/pull/12898, which was reverted in https://github.com/Budibase/budibase/pull/13159 💀 

More context can be found in the original PR.

This was reverted as it didn't take into account user client versions, and some changes exposed an existing issue in how the free footer was displayed. This PR contains the original skeleton loader implementation and new fixes, new code is linked below under relevant sections.

### Change to account for client versions
[code](https://github.com/Budibase/budibase/pull/13180/files#diff-00a580cc371ed84f938fee90ae8b92184bce3c83c04f6a9205a59449e025451dR230)

We manually insert the skeleton loader into the dom before serving it the user, this is so that is appears asap while assets continue to load. The client app then has code to remove the skeleton once it loads, the problem arises because users on older version of the budibase client will not have this removal code, so the skeleton will stay visible permanently. 

The only real options to fix this I came across were:

Use the `version` property exposed on `app` in the server, which would require some semver parsing logic and and an arbitrary cut-off version for skeleton support that would be hard to orchestrate pinning properly.

The other options was to leverage the `features` property of a client version's `manifest.json`, the problem with the manifest  needs to be retrieved per request before returning the document. I'm not 100% sure on the performance implications of this, the manifest is a pretty large file and I'm unsure if it's retrieved via disk/db/object storage, maybe someone on the backend team could provide some context if they know.

~~We went with the later options for now anyway.~~

__Edit__: Following feedback from Mike, we are still doing the above, but are now caching this value upon app creation, or upon app client version changes, just so we don't need to retrieve this data on every page load.

### Free Footer Issue
[code](https://github.com/Budibase/budibase/pull/13180/files#diff-6373c5f60741d43c8a9e3be3b9104174b08f7e50042bd8ca03539f3ddad4da84R1)

We were using a helper function to retrieve if a app should display the free footer or not, this helper function consumed stores, but did not return a store itself, so if the input values changed the output would not update. This worked fine previously, but a change in the skeleton loader PR made this a problem.

We changed this helper function to return a derived store instead, which fixes the issue.

## Changing client versions locally

Screenshot 2024-03-14 at 10 13 29

- This PR also adds the ability to, while in dev mode, to arbitrarily set an app's `revertableVersion`, and `revertableVersion` now defaults to to `1.3.2` for new apps. This is to allow us to test that changes to the API and builder are compatible with older versions of the client. This UI is unstyled, but is only visible in development, so hopefully that's acceptable. 

- There are changes to the code around manifest and client js app retrieval to allow serving older client versions in dev mode. If the user's app is pointed at `0.0.0` it will still point at the local client.

- You can also pull down arbitrary client versions with `yarn get-past-client-version X.X.X`, and they should then be visible in the above pictured dev only toggle for selection upon restarting your dev server. Note that the version you're pulling down needs to both be available as a GitHub tag and a `@budibase/client` npm package for this to work though, but for all versions I tried locally this seemed to be fine.

## Feature branch env
[Feature Branch Link](http://fb-fix-skeleton-loader.fb.qa.budibase.net)


packages/server/src/api/controllers/static/index.ts
```diff
@@ -165,12 +225,27 @@ export const serveApp = async function (ctx: Ctx) {
   try {
     db = context.getAppDB({ skip_setup: true })
     const appInfo = await db.get<any>(DocumentType.APP_METADATA)
+
     let appId = context.getAppId()
+    const manifest = await getComponentLibraryManifest(
```
mike12345567, line 230: I'm not sure about needing to do this everytime, it will likely slow down app loading time a fair bit, as we need to pull these files from S3 (with varied performance).

Is it possible that we could use the app version, rather than a flag in the manifest for this? Or could we update something into the app metadata to state that the skeleton loader can be used?

If not, perhaps there is a way we can cache this, but it would be good if we could note this somewhere in the app metadata rather than needing to pull the manifest as it will be a lot faster.

packages/server/src/api/controllers/static/index.ts
```diff
@@ -24,14 +23,75 @@ import AWS from "aws-sdk"
...
+  Feature,
+} from "@budibase/types"
 import {
   getAppMigrationVersion,
   getLatestMigrationId,
 } from "../../../appMigrations"
 
 import send from "koa-send"
 
+const getThemeVariables = (theme: string) => {
```
mike12345567, line 40: Can we move this into a theme file in `src/constants`, just to avoid polluting the controllers with spectrum info.

aptkingston:
Could you either update the description or write a comment to explain what's happening now to determine where we can use the skeleton loader? Looking at the code I can see you're caching the feature from the manifest in the app doc, which makes sense, but I'm not sure why there are large changes to how reverting app versions work, and what purpose the script to parse old client library versions does.

Would just make reviewing much easier! Thanks.

Ghrehh:
@aptkingston Apologies, had done this yesterday and forgot to actually submit the update, the PR copy is updated with context for the changes now. I'll also check to see if there's any comments I can add to the code itself to offer more context now.

Ghrehh:
@mike12345567 I addressed your feedback here if you could re-review this when you get a moment please 🙏

packages/builder/src/components/deploy/RevertModalVersionSelect.svelte
```diff
@@ -0,0 +1,28 @@
+<script>
```
Ghrehh, line 1: This is the toggle for selecting arbitrary revert versions, should only be visible in development.

packages/builder/src/components/deploy/VersionModal.svelte
```diff
@@ -28,7 +30,9 @@
     $appStore.upgradableVersion &&
     $appStore.version &&
     $appStore.upgradableVersion !== $appStore.version
-  $: revertAvailable = $appStore.revertableVersion != null
+  $: revertAvailable =
```
Ghrehh, line 33: This is to allow setting the `revertableVersion` of older local apps that don't already have a `revertableVersion` set.

packages/builder/src/components/deploy/clientVersions.json
```diff
@@ -0,0 +1,3 @@
+[
```
Ghrehh, line 1: List of available older client versions that a user can set in dev mode, this is updated automatically by the `getOlderClientVerions` script

packages/server/src/api/controllers/static/index.ts
```diff
@@ -247,18 +263,21 @@ export const serveBuilderPreview = async function (ctx: Ctx) {
...
 export const serveClientLibrary = async function (ctx: Ctx) {
+  const version = ctx.request.query.version
+
   const appId = context.getAppId() || (ctx.request.query.appId as string)
   let rootPath = join(NODE_MODULES_PATH, "@budibase", "client", "dist")
   if (!appId) {
     ctx.throw(400, "No app ID provided - cannot fetch client library.")
   }
-  if (env.isProd()) {
+  if (env.isProd() || (env.isDev() && version !== "0.0.0")) {
```
Ghrehh, line 273: This is what allows serving older client versions locally

packages/server/src/utilities/fileSystem/clientLibrary.ts
```diff
@@ -120,7 +121,12 @@ export async function updateClientLibrary(appId: string) {
       ContentType: "application/javascript",
     }
   )
-  await Promise.all([manifestUpload, clientUpload])
+
+  const manifestSrc = fs.promises.readFile(manifest, "utf8")
+
+  await Promise.all([manifestUpload, clientUpload, manifestSrc])
+
+  return JSON.parse(await manifestSrc)
```
Ghrehh, line 129: Returning the manifest from both `revertClientLibrary` and `updateClientLibrary` so that the controllers that consume these can check the manifest for if the user's client version has skeleton loaders enabled.

packages/server/build/oldClientVersions/1.3.2/app.js
mike12345567, line 1: Think this needs removed!

packages/server/build/oldClientVersions/1.3.2/manifest.json
mike12345567, line 1: remove

packages/server/src/api/controllers/static/index.ts
```diff
@@ -247,18 +263,20 @@ export const serveBuilderPreview = async function (ctx: Ctx) {
...
     ctx.body = await objectStore.getReadStream(
       ObjectStoreBuckets.APPS,
       objectStore.clientLibraryPath(appId!)
     )
     ctx.set("Content-Type", "application/javascript")
-  } else if (env.isDev()) {
+  } else if (env.isDev() && version === "0.0.0") {
```
mike12345567, line 279: Can we turn this into a constant - just to describe this is the development version, might look confusing in future!

packages/server/src/utilities/fileSystem/app.ts
```diff
@@ -35,20 +36,25 @@ export const getComponentLibraryManifest = async (library: string) => {
...
-    ]
-    for (let path of paths) {
-      if (fs.existsSync(path)) {
-        // always load from new so that updates are refreshed
-        delete require.cache[require.resolve(path)]
-        return require(path)
+    const db = context.getAppDB()
+    const app = await db.get<App>(DocumentType.APP_METADATA)
+
+    if (app.version === "0.0.0") {
```
mike12345567, line 42: Would use constant for this, as with other comment, can go in `src/constants`

scripts/getPastClientVersion.js
```diff
@@ -0,0 +1,46 @@
+const fs = require('node:fs/promises');
```
mike12345567, line 1: Just for understanding - what is this script needed for?

mike12345567 (commented):
Backend changes look good - should perform quite a bit better on serving apps, thanks for the change!

scripts/getPastClientVersion.js
Ghrehh: It's for pulling down old client versions to test reverting locally. This originally needed to be unmerged because we didn't test the changes with an older client version, so it seemed appropriate to add the functionality here.

mike12345567: sounds good - nice to be aware of - could be worth adding something to Confluence on how to do this/use this, a very useful little script!

deanhannigan (changes requested):
Some minor suggestions and a couple of bugs, but this is looking great! The dynamic loading of the client library is an amazing addition 🤯 

General Feedback
  - Document `get-past-client-version` usage and associated file locations.
    - Usage `yarn get-past-client-version 2.1.0`
    - Downloads to `/packages/server/build/oldClientVersions/2.1.0`
    - `2.1.0` appears in the revert modal in builder app setting

`Bug - fixed` -  app theme differences
  - Portal app skeleton uses the app theme, but the Builder app skeleton uses the portal theme. 
  - Is this intentional?

`Bug - fixed` - Version number missing when reverting (First time)
  - Version number missing from the success prompt
    - `App reverted successfully to version undefined`
    - Discussed this with @Ghrehh and it can just be omitted on the first revert

`Bug - fixed`
  - Navigation customisation not reflected in the loader.
  - Only shows the top bar variant.

`Bug - fixed` - Legacy feature flag support for older apps.
 - Older applications without a feature flag object don't display in the preview.

packages/builder/src/components/deploy/RevertModalVersionSelect.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+    API.setRevertableVersion(appId, value)
+  }
+
+  $: unselectedClientVersions = clientVersions.filter(
+    version => version !== revertableVersion
+  )
+</script>
+
+<select on:change={handleChange}>
```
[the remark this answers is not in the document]
Ghrehh: Sure, still janky developer ui, but how do you think this looks?
Screenshot 2024-03-24 at 20 39 57
Screenshot 2024-03-24 at 20 40 00

packages/builder/src/components/deploy/RevertModalVersionSelect.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import { API } from "api"
+  import clientVersions from "./clientVersions.json"
+  import { appStore } from "stores/builder"
+  import { Select } from "@budibase/bbui"
+
+  export let revertableVersion
+  $: appId = $appStore.appId
+
+  const handleChange = e => {
+    const value = e.value
```
deanhannigan, line 11: This should be `e.detail`. Reverting without this throws an error.

deanhannigan (changes requested):
One issue with the revert modal behaviour but the other changes are looking grand!

packages/builder/src/components/deploy/RevertModalVersionSelect.svelte
Ghrehh: My bad, I had changed this, no idea how it went back to this.

Ghrehh:
@deanhannigan Feedback addressed, if you could give this another look when you get a chance please 🙏

deanhannigan (approved):
LGTM!

packages/frontend-core/src/themes/midnight.css
```diff
@@ -17,5 +17,8 @@
     --modal-background: var(--spectrum-global-color-gray-50);
     --drop-shadow: rgba(0, 0, 0, 0.25) !important;
     --spectrum-global-color-blue-100: rgba(35, 40, 50) !important;
+
+    --spectrum-alias-background-color-secondary: var(--spectrum-global-color-gray-75);
+    --spectrum-alias-background-color-primary: var(--spectrum-global-color-gray-100);
```
aptkingston, line 22: These are already the default values for these variables - we don't need to redeclare them here.

aptkingston (approved):
LGTM! Just one small suggestion for code we can remove 👍

PR #13183, opened 2024-03-04
Fix issue with filtering client side docs using lucene

PClmnt:
## Description
Fixes an issue in the filtering of client side data like REST, where you can only apply one filter and get results. i.e the logical AND wasn't working correctly, so using two filters would not return results when either one of them did match. You would need to have matched both of them for results to show. 

Also update the test builder for lucene queries to take multiple params. 
## Addresses
BUDI-7993

## Launchcontrol
Fixes an issue in client side filtering where results weren't being returned for multiple filters

## Feature branch env
[Feature Branch Link](http://fb-fix-filtering-lucene-bug.fb.qa.budibase.net)


mike12345567:
Looks good - nice to have a test case for it - but think should get @aptkingston to give final confirmation on this!

packages/shared-core/src/tests/filters.test.ts
```diff
@@ -150,40 +161,87 @@ describe("runLuceneQuery", () => {
...
+      oneOf: { staff_id: [10] },
+      contains: { description: ["box"] },
+    })
+
+    expect(runLuceneQuery(docs, query).map(row => row.order_id)).toEqual([
+      1, 2, 3,
+    ])
+  })
+
+  // what should the name of this test be if it's the same test as above but with different operands
```
melohagan, line 224: You don't need a second test in this case. Simply add parameters to your test that include both the input + expected outputs.

Use `test.each` - there is an example in this file:
https://vitest.dev/api/#test-each

melohagan (dismissed):

packages/shared-core/src/filters.ts
```diff
@@ -390,24 +390,42 @@ export const runLuceneQuery = (docs: any[], query?: SearchQuery) => {
...
+      .filter(
+        ([key, value]: [string, any]) =>
+          !["allOr", "onEmptyFilter"].includes(key) &&
+          Object.keys(value as Record<string, any>).length > 0
+      )
+      .map(([key]) => key as any)
+
+    const results: boolean[] = activeFilterKeys.map(filterKey => {
+      const filterFunction = filterFunctions[filterKey]
+      return filterFunction ? filterFunction(doc) : true
```
aptkingston, line 419: I don't think we should return `true` for a missing filter function here - that would mean that if I sent up an invalid payload of `{ foo: { a : 1 } }` it will fail to find a filter function for `foo`, return `true` and therefore always match an `allOr` query because there's a true result, even if all my other valid filters fail.

I think we should return `false` so that it's completely ignored for a `allOr` query, and so that an "and" query will fail (which it should because this invalid filter function cannot match any rows).

You can also rewrite this as `return filterFunctions[filterKey]?.(doc) ?? false` to be a bit terser too.

packages/shared-core/src/filters.ts
```diff
@@ -390,24 +390,42 @@ export const runLuceneQuery = (docs: any[], query?: SearchQuery) => {
...
+        contains: contains,
+        containsAny: containsAny,
+        notContains: notContains,
+      }
+
+    const activeFilterKeys: SearchQueryOperators[] = Object.entries(query || {})
+      .filter(
+        ([key, value]: [string, any]) =>
+          !["allOr", "onEmptyFilter"].includes(key) &&
+          Object.keys(value as Record<string, any>).length > 0
```
aptkingston, line 413: Do you need to ensure `value` isn't nullish here? Unless typescript does some magic during the casting here, I imagine that sending up `{ string: null }` would throw an error, especially since `value` is typed as `any` here.

PClmnt: Sorted - and have a test case in for this.

aptkingston (approved):
LGTM!

Issue #13178, opened 2024-03-03
Problem with the MySQL table if the table name starts with "Customer"


n05331:
**Hosting**

- Self
- Budibase Version: 2.21.0

The problem started today when I upgraded to the latest version.
If there is a MySQL table whose table name starts with "Customer" it will not work.
I have an existing table named: "Customers" and after the upgrade it stopped working.
I see the following error:
There was a problem loading your grid
Unknown column 'as.id' in 'field list'
![image](https://github.com/Budibase/budibase/assets/102161797/dec1ffa0-4386-49cc-919f-33e08c9dde0a)

I tried to create a new table with the name: "Customer" and it works.
I tried to create a new table with the name: "CustomerVV" and the same problem.
There was a problem loading your grid
Unknown column 'aVV.id' in 'field list'
![image](https://github.com/Budibase/budibase/assets/102161797/541216b0-27c8-4454-800d-365f239fdd5d)

It seems that if the table name starts with "Customer" it is automatically changed to "a" and therefore the table does not work.

mvhaen:
I'm experiencing a similar problem, however our app was not yet updated to the latest version when it started occurring. This is our schema, we don't have a `Customers` table though.

[STR_Translations-dump.sql.zip](https://github.com/Budibase/budibase/files/14482511/STR_Translations-dump.sql.zip)

mike12345567:
Hi @mvhaen - just to confirm, are you self hosted and this issue still occurs for you before 2.21.0?

@n05331 is there anyway you could provide a schema demonstrating the issue you're experiencing? You can retrieve it using this: https://docs.budibase.com/docs/troubleshooting-sql

If you can retrieve a schema, you can send it privately to `support@budibase.com`

mvhaen:
@mike12345567 

I'm still running into the issue. We use the cloud offering by the way, not a local install. 

![image](https://github.com/Budibase/budibase/assets/157020/0336ecef-051e-4947-b1f0-4f103720506d)

Is it possible the fix is not yet in production?

mvhaen:
@mike12345567 

After updating the app, I'm still running into this issue. We are using the cloud offering, not a local install bu the way.

> The app is currently using version 2.21.1.

![image](https://github.com/Budibase/budibase/assets/157020/7980b51b-4294-407f-8624-1cd36d18152c)

Is it possible that the fix was not deployed in production yet?

n05331:
I have now upgraded to the updated version: 2.21.2
[...]

Issue #13177, opened 2024-03-02
Getting error on Azure SQL


dfourmanfsquared:
**Checklist**
- [X ] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: dashingdanstrains

**Describe the bug**
This was working until yesterday, getting the following error whenever creating a new row using a simple row creation form: The multi-part identifier "a.id" could not be bound.

**To Reproduce**
1. Go to ['...'](https://dashingdanstrains.budibase.app/app/traininventory#/train)  (can only be done with authenticated user)
2. Click on create
3. Enter required info

**Expected behavior**
Row should be added with a success message

**Screenshots**
<img width="429" alt="image" src="https://github.com/Budibase/budibase/assets/107331834/bde91822-d97a-4504-bba4-6bb7f24f0f60">
  
**App Export**
See attached
[traininventory-export-1709386995986.tar.gz](https://github.com/Budibase/budibase/files/14469527/traininventory-export-1709386995986.tar.gz)

**Desktop (please complete the following information):**
 - OS: Any
 - Browser Chrome
 - Version Latest

dfourmanfsquared:
@mike12345567 i upgraded to the latest version but still seeing errors

<img width="604" alt="image" src="https://github.com/Budibase/budibase/assets/107331834/79d43de0-be1e-4452-8e93-945846a6d6ba">

PR #13187, opened 2024-03-04
Fix SQL aliasing issue when creating rows

mike12345567:
## Description
Fixing an issue with returning statements in MS-SQL and MySQL, when creating, updating or deleting rows the response is not retrieved correctly when aliased.

Also includes a fix for table names that contain each other, e.g. "Customers" and "Customer". These would sometimes pick each other due to the way we handle wrapping statements, like `""` or `[]` - this is now resolved by checking if something is wrapping, only allowing an increased limit if it is wrapped.

## Addresses
https://github.com/Budibase/budibase/issues/13177
https://github.com/Budibase/budibase/issues/13178


packages/server/src/integrations/tests/sqlAlias.spec.ts
```diff
@@ -160,6 +161,28 @@ describe("Captures of real examples", () => {
...
   })
 
+  describe("returning (everything bar Postgres)", () => {
+    it("should be able to handle row returning", () => {
+      const queryJson = getJson("createSimple.json")
+      const SQL = new Sql(SqlClient.MS_SQL, limit)
+      let query = SQL._query(queryJson, { disableReturning: true })
+      expect(query).toEqual({
+        sql: "insert into [people] ([age], [name]) values (@p0, @p1)",
+        bindings: [222, "awfawf"],
```
samwho, line 171: AWF AWF 🐶

samwho (approved):

samwho (approved):

PR #13188, opened 2024-03-04
Type role controller.

samwho:
## Description

This PR fully types the handlers in `packages/server/src/api/controllers/role.ts`.


packages/server/src/api/routes/tests/application.spec.ts
```diff
@@ -251,10 +251,15 @@ describe("/applications", () => {
...
 
-      const apps = await config.api.application.fetch()
-      expect(apps.length).toBeGreaterThan(0)
+      await config.withUser(user, async () => {
+        const apps = await config.api.application.fetch()
+        expect(apps).toHaveLength(0)
+      })
     })
   })
 })
```
samwho, line 265: I'm planning to expand on this in subsequent PRs, sorry it slipped in here.

packages/backend-core/src/security/roles.ts
```diff
@@ -84,16 +84,18 @@ export function getBuiltinRoles(): { [key: string]: RoleDoc } {
...
-export function isBuiltin(role?: string) {
-  return BUILTIN_ROLE_ID_ARRAY.some(builtin => role?.includes(builtin))
+export function getBuiltinRole(roleId: string): Role | undefined {
+  const role = Object.values(BUILTIN_ROLES).find(role =>
+    roleId.includes(role._id)
+  )
+  if (!role) {
+    return undefined
+  }
+  return cloneDeep(role)
```
samwho, line 98: @mike12345567 I'm actually not sure about this. When you create a custom role, it gets an ID of the form `role_<name>`. If you named your role `role_ADMIN` wouldn't it inherit the permissions of the admin role by default?

I feel like anything that's checking to see if a role ID contains the name of a built in role is flawed, no?

packages/backend-core/src/security/roles.ts
```diff
@@ -161,35 +163,28 @@ export function lowerBuiltinRoleID(roleId1?: string, roleId2?: string): string {
...
     // finalise the ID
-    role._id = getExternalRoleID(role._id, role.version)
+    role._id = getExternalRoleID(role._id!, role.version)
   } catch (err) {
     if (!isBuiltin(roleId) && opts?.defaultPublic) {
       return cloneDeep(BUILTIN_ROLES.PUBLIC)
     }
     // only throw an error if there is no role at all
-    if (Object.keys(role).length === 0) {
+    if (!role || Object.keys(role).length === 0) {
```
melohagan, line 187: NAB: Just curious, do you have a preference of:

`!role || Object.keys(role).length === 0)`

instead of: 

`Object.keys(role || {}).length === 0)`

packages/server/src/api/controllers/role.ts
```diff
@@ -25,43 +31,35 @@ async function updateRolesOnUserTable(
...
-          : roleId
-      const indexOfRoleId = constraints.inclusion!.indexOf(updatedRoleId)
-      if (remove && indexOfRoleId !== -1) {
-        constraints.inclusion!.splice(indexOfRoleId, 1)
-      } else if (!remove && indexOfRoleId === -1) {
-        constraints.inclusion!.push(updatedRoleId)
-      }
-      break
+  const constraints = table.schema.roleId?.constraints
+  if (constraints) {
```
melohagan, line 38: NAB: Could reduce nesting with a guard statement/early return?

melohagan (approved):

packages/backend-core/src/security/roles.ts
samwho: I tend to prefer the latter, come to think of it. Thanks for pointing it out! Fixed. :)

packages/server/src/api/controllers/role.ts
samwho: Really good call. I'm a big fan of early returns / minimised nesting. Agree with basically every point in this video: https://www.youtube.com/watch?v=CFRhGnuXG-4.

melohagan: Haha I think I've actually seen that video before.

packages/backend-core/src/security/roles.ts
samwho: Turns out this was actually important for type checking. Having the `!role` checks makes `role` be `RoleDoc` instead of `RoleDoc | undefined`. Have put this back how it was.

PR #13191, opened 2024-03-05
[fix] BUDI-7710: User groups do not fully support custom roles

samwho:
## Description

This PR fixes the permissions problem that exists if you create a role, attach that role to a group, attach that group to a user, and then try to access an app with that user through the group.

The problem was that when you create a role, it gets created initially into the development database for the app. But when you try to access the app, you're trying to access the production database for the app, where the role does not exist until you publish the app.

This is a tricky UX problem, because we can't just poke data manually in to the prod database because it'll cause conflicts on next publish (publishing is a CouchDB sync between the dev and prod DB). We also can't trigger a publish of the app because we have no idea what state the dev database is in. A user could be working on the app and it may be in a broken state.

The solution @mike12345567 and I came up with was to do a filtered replication when a role is created/updated. We replicate just that role document, nothing else. This means CouchDB won't get into a conflicted state on next publish.

## Addresses
- https://linear.app/budibase/issue/BUDI-7710/user-groups-do-not-fully-support-custom-roles


packages/backend-core/src/db/Replication.ts
```diff
@@ -1,66 +1,57 @@
...
   }
 
-  /**
-   * Two way replication operation, intended to be promise based.
-   * @param opts - PouchDB replication options
-   */
-  sync(opts = {}) {
-    this.replication = this.promisify(this.source.sync, opts)
-    return this.replication
-  }
```
samwho, line 46: This wasn't being used anywhere so I figured I may as well delete it, which would allow me to also delete `promisify` because it will only be used in one place.

packages/backend-core/src/db/Replication.ts
```diff
@@ -1,66 +1,57 @@
+import PouchDB from "pouchdb"
 import { getPouchDB, closePouchDB } from "./couch"
 import { DocumentType } from "../constants"
 
 class Replication {
-  source: any
-  target: any
-  replication: any
```
samwho, line 7: Nobody was using this so I got rid of it.

packages/backend-core/src/db/Replication.ts
```diff
@@ -1,66 +1,57 @@
...
-  sync(opts = {}) {
-    this.replication = this.promisify(this.source.sync, opts)
-    return this.replication
-  }
+  appReplicateOpts(
+    opts: PouchDB.Replication.ReplicateOptions = {}
+  ): PouchDB.Replication.ReplicateOptions {
+    if (typeof opts.filter === "string") {
+      return opts
+    }
```
samwho, line 40: `ReplicateOptions.filter` can be set to the value `"_view"` to invoke some special behaviour, so I'm checking the type here and passing `opts` through without modifying it if the user is trying to do something specific.

packages/server/src/api/controllers/user.ts
```diff
@@ -42,6 +42,7 @@ export async function updateMetadata(
   // this isn't applicable to the user
   delete metadata.roles
   ctx.body = await db.put(metadata)
+  await cache.user.invalidateUser(user._id!)
```
samwho, line 45: This was added during debugging what the problem was, as I thought it could be caching related. It wasn't, but I feel like having this here is correct so I left it in.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -372,21 +373,20 @@ export default class TestConfiguration {
...
       _rev: resp.rev,
       ...user,
     }
   }
 
   async createUser(user: Partial<User> = {}): Promise<User> {
-    const resp = await this.globalUser(user)
-    await cache.user.invalidateUser(resp._id!)
-    return resp
+    return await this.globalUser(user)
```
samwho, line 389: Given that this is now down to just a call to `globalUser`, I think I will do a follow up PR that unifies the uses of this function and gives it a better name. The cache invalidation was the last holdout, and that's now moved into `globalUser` because I feel like it's more correct to have it there.

packages/server/src/api/controllers/role.ts
```diff
@@ -106,6 +106,21 @@ export async function save(ctx: UserCtx<SaveRoleRequest, SaveRoleResponse>) {
...
+  const devDb = context.getDevAppDB()
+  const prodDb = context.getProdAppDB()
+
+  if (await prodDb.exists()) {
+    const replication = new dbCore.Replication({
+      source: devDb.name,
+      target: prodDb.name,
+    })
+    await replication.replicate({
+      filter: (doc: any, params: any) => {
```
mike12345567, line 119: Wondering if we should just always replicate all roles here - same effort but would make sure roles are just always consistently in sync, any role operation makes sure all roles get synced (like for example, if roles depended on each other and some were updated at the same time).

samwho: It's easy to do. I'm struggling to come up with a reason _not_ to do it. I'll defer to your judgement.

packages/server/src/api/controllers/user.ts
mike12345567: Ah - the user here is the global metadata user, technically this comes with no negative - but the user metadata should never have any impact on the actual global user itself. The two concepts are quite separate, despite looking very related, but we try to make sure its not possible to update the global user via the app metadata about the user.

packages/server/src/tests/utilities/TestConfiguration.ts
mike12345567: Yeap, this definitely could be updated, right now its a little confusing.

packages/server/src/api/controllers/user.ts
samwho: Ah, so should I remove this?

packages/backend-core/src/db/Replication.ts
mike12345567: Yeap this is very old - makes sense to remove it, there isn't really any good use cases for it anymore.

packages/server/src/api/controllers/user.ts
mike12345567: Sorry - didn't see this comment earlier - yeap I would remove this!

mike12345567 (approved):
LGTM - thanks for the changes, good to get this one resolved!

packages/server/src/api/controllers/user.ts
```diff
@@ -1,6 +1,6 @@
 import { generateUserFlagID, InternalTables } from "../../db/utils"
 import { getFullUser } from "../../utilities/users"
-import { context } from "@budibase/backend-core"
+import { cache, context } from "@budibase/backend-core"
```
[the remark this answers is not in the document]
samwho: Hm, didn't seem to 🤔

Issue #10762, opened 2023-06-01
[BUDI-7078] Side Panel UX issue: releasing a click off-panel closes the panel


agmcnabb:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: Docker
  - Budibase Version: 2.5.2
  - App Version: 2.5.2

**Describe the bug**
A couple of users have noticed an irritating behaviour with the side panel: when clicking and dragging to highlight text in a text field, if the mouse is hovering off the side panel when the click is released, the app closes the side panel - recording attached.

**To Reproduce**
1. Have a side panel with text field.
2. Click and drag to highlight the text in the text field.
3. Release the click while the mouse is not hovering over the side panel.
4. The click release is interpreted as a click, and the side panel closes.

**Expected behavior**
Releasing a click that started on the side panel should not close the panel.

**Screenshots**

https://github.com/Budibase/budibase/assets/125557661/a4e72b4b-035c-4e74-b45b-00fce7f9c0f6

  
**App Export**

[test app-export-1685626054269.tar.gz](https://github.com/Budibase/budibase/files/11625803/test.app-export-1685626054269.tar.gz)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome, Edge

<sub>[BUDI-7078](https://linear.app/budibase/issue/BUDI-7078/side-panel-ux-issue-releasing-a-click-off-panel-closes-the-panel)</sub>

melohagan:
To add to this one, if you close an error toast it also closes the side panel which is a little frustrating:


https://github.com/Budibase/budibase/assets/101575380/ed858d17-eb74-4185-9df3-835978b2916e

bjornclauw:
would it be possible to add an option under the sidebar so the sidebar can only be closed with the dedicated button? (=clicking outside of the sidebar does nothing)

PR #13193, opened 2024-03-05
Use clickoutside library to stop mouseup closing side panel

melohagan:
## Description
A slightly annoying bug was happening were clicking down within the side panel, and releasing outside the side panel would close it. This could happen when using the mouse to select the contents of a field for example.

This bug is easily fixed by requiring that only **mousedown** outside of the side panel will close it.

## Addresses
- https://github.com/Budibase/budibase/issues/10762

## Screenshots
fields_bug


melohagan:
@aptkingston Well that was easier than I thought! 😄 

Changing the event handler to `mousedown` seems to do the trick. You were right about date pickers not being handled correctly, but that is fixed with this change.
Any other thoughts or areas to test?

aptkingston:
Nearly impossible to know the full extent of any side effects here because clickOutside is used by a large amount of components, but we'll discover any issues over time of there are any. In principle I can't think of any glaring reason that we can't just use mouse down here. The sanity check is just that common things work - modals, popovers, datepickers, grid cell popovers etc all close properly still.

melohagan:
> Nearly impossible to know the full extent of any side effects here because clickOutside is used by a large amount of components, but we'll discover any issues over time of there are any. In principle I can't think of any glaring reason that we can't just use mouse down here. The sanity check is just that common things work - modals, popovers, datepickers, grid cell popovers etc all close properly still.

Hmm, for the Grid view it does alter the behaviour in a slightly bad way:
https://github.com/Budibase/budibase/assets/101575380/c0193023-1283-4565-9dc2-a87d7cc3770a

I would much prefer to be able to pass in the event handler on a per usage basis (like that library). I'll see if I can figure that out.

aptkingston:
> > Nearly impossible to know the full extent of any side effects here because clickOutside is used by a large amount of components, but we'll discover any issues over time of there are any. In principle I can't think of any glaring reason that we can't just use mouse down here. The sanity check is just that common things work - modals, popovers, datepickers, grid cell popovers etc all close properly still.
> 
> Hmm, for the Grid view it does alter the behaviour in a slightly bad way: https://github.com/Budibase/budibase/assets/101575380/c0193023-1283-4565-9dc2-a87d7cc3770a
> 
> I would much prefer to be able to pass in the event handler on a per usage basis (like that library). I'll see if I can figure that out.

I'm not sure we can do it on a per usage basis - we deliberately use a single listener, as if we constructed one every time we would end up with potentially 100s of global click event listeners. You could potentially use a single click and a single mousedown, and support passing in a type, which would then use the appropriate listener.

melohagan:
I think this is a safer approach now if you want to take a look @aptkingston

aptkingston (approved):
Yea this is definitely safer. It does have the side effect of making the `contextmenu` listener completely redundant since nothing is ever configured with an allowedType of `contextmenu`, so you can probably just remove that listener. I don't think it's particularly important to close things on a context menu click.

PR #13197, opened 2024-03-05
Create Redis bulk store utils

adrinr:
## Description
Adding a new util into our redis implementation to allow bulk stores. We use redis pipelines to perform this action atomically on the redis server

## Addresses

## Launchcontrol
Adding `redis.bulkStore` function


packages/backend-core/src/redis/redis.ts
```diff
@@ -274,13 +279,44 @@ class RedisWrapper {
...
+        end
+        `
+    const keys = Object.keys(dataToStore)
+    const values = Object.values(dataToStore)
+    if (expirySeconds !== null) {
+      values.push(expirySeconds)
+    }
+
+    await client.eval(luaScript, keys.length, ...keys, ...values)
+  }
```
samwho, line 311: How come we can't use `mset` directly? https://redis.github.io/ioredis/classes/Redis.html#mset

Slightly concerned that Redis mark the use of `eval` as `@slow` in their docs: https://redis.io/commands/eval/

adrinr: We can, but there is no way to add the expiration in bulk without eval.
We could add the keys first and then modify the expiration (as we do in the actual `store`), but probably it's best to have it as an atomic operation

adrinr: I will play with the pipeline option. It's not transactional, but at least it will ensure atomicity

samwho: `multi` might be more appropriate for this use-case: https://redis.io/commands/multi/

I'm not 100% sure, though, and reading a bit more around scripts they do guarantee atomicity provided you only access the keys you say you will. I suspect the code using `multi` might look a bit nicer, but ultimately leave it up to you to decide.

packages/backend-core/src/redis/redis.ts
```diff
@@ -291,23 +291,16 @@ class RedisWrapper {
...
-            ${
-              expirySeconds !== null
-                ? `redis.call('EXPIRE', key, ARGV[#ARGV])`
-                : ""
-            }
-        end
-        `
-    const keys = Object.keys(dataToStore)
-    const values = Object.values(dataToStore)
+    const pipeline = client.pipeline()
```
samwho, line 294: Pipelines aren't guaranteed to be atomic. I don't believe that is important for this code, the worse that can happen is the expiry is overwritten with a slightly different time, but just wanted to point it out in case you can think of a reason it would matter.

samwho (approved):

packages/backend-core/src/redis/redis.ts
adrinr: Pipelines are meant to be leaner, I think they should be fine for this use case

PR #13201, opened 2024-03-05
Create Redis increment utils

adrinr:
## Description
Adding a new util into our redis implementation to allow `increment`.
I realised that the error responses between `ioredis` and `ioredis-mock` are not the same. Because of this this PR is testing the utils against a real redis instance using testcontainers

## Addresses

## Launchcontrol
Adding `redis.increment` function


packages/backend-core/src/redis/redis.ts
```diff
@@ -331,6 +331,14 @@ class RedisWrapper {
     let items = await this.scan()
     await Promise.all(items.map((obj: any) => this.delete(obj.key)))
   }
+
+  async increment(key: string) {
+    const result = await this.getClient().incr(addDbPrefix(this._db, key))
+    if (isNaN(result)) {
+      throw new Error(`Redis ${key} does not contains a number`)
```
samwho, line 338:
```suggestion
      throw new Error(`Redis ${key} does not contain a number`)
```

packages/backend-core/src/redis/tests/redis.spec.ts
```diff
@@ -1,8 +1,28 @@
...
+    const container = await new GenericContainer("redis")
+      .withExposedPorts(6379)
+      .start()
+
+    env._set(
+      "REDIS_URL",
+      `${container.getHost()}:${container.getMappedPort(6379)}`
+    )
+    env._set("MOCK_REDIS", 0)
+    env._set("REDIS_PASSWORD", 0)
```
samwho, line 22:
```suggestion
    const cleanup = config.withEnv({
      REDIS_URL: `${container.getHost()}:${container.getMappedPort(6379)}`,
      MOCK_REDIS: 0,
      REDIS_PASSWORD: 0,
    })
```

Then in `afterAll` you can call `cleanup()` and it'll reset the environment variables back to their original values.

samwho (commented):
Love this. Would quite like to get rid of `ioredis-mock` entirely.

adrinr:
> `ioredis-mock`

Agreed! This was done in server and worker already, not sure if there are many uses of `ioredis-mock` left

packages/backend-core/src/redis/tests/redis.spec.ts
adrinr: This is set to package and server. Jest will be restoring these settings between files, so no real need to set this here

samwho (approved):

PR #13204, opened 2024-03-06
Create Redis deleteIfValue utils

adrinr:
## Description
Adding a new util into our redis implementation to allow deleting a key only if this is matching a value. This can be used to remove a value ensuring that nobody overrode it. We use lua scripts to perform this action atomically on the redis server

## Addresses

## Launchcontrol
Adding `redis.deleteIfValue` function


packages/backend-core/src/redis/redis.ts
```diff
@@ -339,6 +339,18 @@ class RedisWrapper {
     }
     return result
   }
+
+  async deleteIfValue(key: string, value: any) {
+    const client = this.getClient()
```
mike12345567, line 344: This is very useful - I like it!

mike12345567 (approved):

samwho (approved):

PR #13205, opened 2024-03-06
App favourites

deanhannigan:
## Description
Mark your favourite apps in the portal app list! Favourited apps will shift to the top of the list while also maintaining the selected sort order.
- The apps store has been refactored into a BudiStore
  - App list sort order is now kept in here with a new enriched app store. 
  - When moving between screens you don't have to reapply your sort!
- Updated the Icon component to use the `AbsTooltip` and exposed new customisation options.
- Updated the app preview screen UI and added the app favourite action.
- `FIX` - Refresh auth when a `Creator` creates a new application. After creating an app they could not edit or view their own app without reloading the page.

## Addresses
- https://linear.app/budibase/issue/BUDI-7678/apps-list-add-favouriting

## Screenshots
Screenshot 2024-03-06 at 09 30 45

The sorted app list and favouriting is also reflected in the preview side panel
Screenshot 2024-03-06 at 09 31 09

Preview screen reworked
Screenshot 2024-03-06 at 09 32 02

## Launchcontrol

You can now favourite your apps and bump them to the top of the app list! App list sorting is now also remembered between screens

## Feature branch env
[Feature Branch Link](http://fb-feature-app-favourites.fb.qa.budibase.net)


packages/worker/src/api/controllers/global/self.ts
```diff
@@ -105,16 +110,79 @@ export async function getSelf(ctx: any) {
...
   ctx: UserCtx<UpdateSelfRequest, UpdateSelfResponse>
 ) {
   const update = ctx.request.body
 
   let user = await userSdk.db.getUser(ctx.user._id!)
-  user = {
-    ...user,
-    ...update,
+
+  if ("appFavourites" in update) {
```
adrinr, line 168: Not sure if this should be here. Is there a reason why is this not in a different endpoint that handles only this data?

packages/worker/src/api/controllers/global/self.ts
```diff
@@ -131,3 +199,12 @@ export async function updateSelf(
     _rev: user._rev!,
   }
 }
+
+// export default {
```
adrinr, line 203: Clean leftovers

adrinr (dismissed):

packages/worker/src/api/controllers/global/self.ts
deanhannigan: I didn't separate this into it's own endpoint because it follows the same pattern we have elsewhere for updating distinct user fields. The block that you linked should only ever be hit when explicitly modifying the `appFavourites` as it makes a call to pull apps and sync with the server. This should only ever be run when an update to self includes `appFavourites`.

I did make a change however to ensure that the behaviour on the endpoint is exactly the same when updating favourites as it current excludes anything else.

MihailHadzhiev2022 (approved):
LGTM

aptkingston (commented):
LGTM! Just one suggestion and a couple of questions.

packages/worker/src/api/controllers/global/self.ts
samwho: Where else do we use this pattern? It does feel quite strange.

packages/server/src/api/controllers/application.ts
```diff
@@ -682,7 +683,10 @@ export async function duplicateApp(
   // Build a new request
   const createRequest = {
     roleId: ctx.roleId,
-    user: ctx.user,
+    user: {
+      ...cloneDeep(ctx.user),
+      _id: dbCore.getGlobalIDFromUserMetadataID(ctx.user._id || ""),
```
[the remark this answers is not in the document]
deanhannigan: Yep that's fair enough, this was definitely a reflexive use of cloneDeep when I began to reprocess the context user. It can be safely removed.

packages/worker/src/api/controllers/global/self.ts
deanhannigan: I was just referring to the use of `updateSelf` to update individual fields on the user. The only deviation `appFavourites` has here is an optional sync to clear out invalid apps. I don't think this warrants a separate endpoint to handle. The flow here has been updated further after feedback from @aptkingston.

aptkingston (approved):

PR #13208, opened 2024-03-07
Added custom title field to confirm actions modals

ConorWebb96:
## Description
This allows users to change the title of the confirmation modals for various actions such as; Save Row, Delete Row, Duplicate Row and Execute Query actions

## Addresses
- POC for customer requested this

## Screenshots
**Before**
Screenshot 2024-03-07 at 08 04 09

**After**
Screenshot 2024-03-07 at 08 03 36

## Launchcontrol
Adds some more customisation to action confirmation modals. Allows users to change the titles to better describe what the specific actions are actually doing.


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DeleteRow.svelte
```diff
@@ -45,6 +45,9 @@
     <Checkbox text="Require confirmation" bind:value={parameters.confirm} />
 
     {#if parameters.confirm}
+      <Label small>Confirm title</Label>
```
melohagan, line 48: NAB: Because this setting appears when `Require confirmation` is checked, you could probably just call the label *Title* and the user will know what that is, imo.

melohagan (approved):
LGTM - nice feature to have!

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DeleteRow.svelte
aptkingston: Agreed - these could just be `Title` and `Text`.

aptkingston (approved):

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DeleteRow.svelte
ConorWebb96: Sweet, I've made that change, thanks guys.

PR #13215, opened 2024-03-08
Added text colour option to tag component via manifest

ConorWebb96:
## Description
When using certain colours it was nearly impossible to see what the text inside a tag was. An impossible one would have been white as the custom color select. When this was selected it was a white background with white text which just appears as invisible text.

## Addresses
- Raised by Joe internally

## Screenshots
**Before**
Screenshot 2024-03-08 at 08 39 57

**After**
Screenshot 2024-03-08 at 08 36 26

## Launchcontrol
Small addition to the tag component, allowing for custom colour text selection to avoid clashes with background colours.

### DIFF
--- a/packages/client/src/components/app/Tag.svelte
+++ b/packages/client/src/components/app/Tag.svelte
@@ -35,7 +35,7 @@
         ...styles?.normal,
         "background-color": color,
         "border-color": color,
-        color: textColor ? textColor : "white",
+        color: textColor || "white",
         "--spectrum-clearbutton-medium-icon-color": "white",
       },
     }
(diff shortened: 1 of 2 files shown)


packages/client/src/components/app/Tag.svelte
```diff
@@ -34,7 +35,7 @@
         ...styles?.normal,
         "background-color": color,
         "border-color": color,
-        color: "white",
+        color: textColor ? textColor : "white",
```
shogunpurple, line 38:
```suggestion
        color: textColor ||"white",
```

shogunpurple (approved):
LGTM!

packages/client/src/components/app/Tag.svelte
ConorWebb96: Updated thanks @shogunpurple 🚀

Issue #12502, opened 2023-12-05
[BUDI-7814] Custom SQL query schema does not update when the query body changes


melohagan:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.13.28

**Describe the bug**
When I change the body of an existing SQL query, the schema does not update to reflect this.

**To Reproduce**

1. Add a read query that selects from one table
2. Run and save -> see that the schema is correctly set
3. Change the read query to select from a different table
4. Run and save -> see that the schema has not updated, even though none of the fields match

**Expected behavior**
If you change the body of a SQL query and the fields returned are different, then the schema should be updated.

**Screenshots**

![Screenshot 2023-12-05 at 12 56 23](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/b8cbe431-7154-4d0f-8400-341a1b97a147/21593eac-fefe-45ba-ae20-2fa50ebd0ea2?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9iOGNiZTQzMS03MTU0LTRkMGYtODQwMC0zNDFhMWI5N2ExNDcvMjE1OTNlYWMtZmVmZS00NWJhLWFlMjAtMmZhNTBlYmQwZWEyIiwiaWF0IjoxNzAxNzgxNzQ4LCJleHAiOjE3MDE4NjgxNDh9.lY2n0tM3XIWJQoEmqNsRrU8ci_0fc_BAWmuqvTlL2L8)

*Response after changing the query body*

[...]

PR #13222, opened 2024-03-08
Fix query schema updating

mike12345567:
## Description
Quick fix for #12502 - the schema was never updating in the UI no matter what was input despite the response - cleaned up the key parsing a little on the backend and made sure there is simply one schema that the frontend always uses, respecting whatever the backend responds with.

Before if you ran a query like `select * from people` and saved it, then ran a query to a different table, like `select * from tasks` - the backend would correctly work this out, but the frontend simply would not display the new schema.

Changes on the backend are simply because the old method of returning keys left a lot of duplicate keys, wasn't a problem but was quite confusing when debugging as if a lot of rows were returned it would return a list of a few hundred/thousand keys, without de-duping.

## Addresses
- https://linear.app/budibase/issue/BUDI-7814/custom-sql-query-schema-does-not-update-when-the-query-body-changes
- #12502


melohagan (approved):
LGTM! The code is easier to understand now as well 👍

packages/server/src/threads/query.ts
```diff
@@ -170,7 +170,12 @@ class QueryRunner {
     }
 
     // get all the potential fields in the schema
-    let keys = rows.flatMap(Object.keys)
```
adrinr, line 173: You could just wrap this into a new Set, avoiding the foreach:
`let keys = new Set(rows.flatMap(Object.keys))`

adrinr (approved):
LGTM, small comment left

PR #13229, opened 2024-03-11
Fix type errors that would be caused by `noUncheckedIndexedAccess` in `packages/backend-core`

samwho:
## Description

This PR fixes type errors in `packages/backend-core` that would arise were we to set `noUncheckedIndexAccess` in the tsconfig.json.

This PR doesn't actually set `noUncheckedIndexedAccess` in `packages/backend-core/tsconfig.build.json` because when I do that, while `yarn check:types` works correctly, VSCode doesn't appear to flag type errors caused by the setting in `packages/server`. It flags them inside of `packages/backend-core`, but not `packages/server`. I'm not sure why, and given it's a pretty bad development experience I've left it out for now.

The plan is to slowly work through these errors throughout the codebase until I can turn this setting on everywhere.


samwho:
Closing this for now, is very fiddly and not a high priority.

PR #13230, opened 2024-03-11
Fix - Group members edits only for admins

adrinr:
## Description
With the previous builder users, we have some users who can visualise users without being admins. We don't want these to edit roles, encouraging them to migrate to the new roles.
This PR unifies the UI and endpoints, allowing only users to perform these actions.

Related:
- https://github.com/Budibase/budibase-pro/pull/276

## Launchcontrol
Group member add/remove only for admins


aptkingston (approved):
LGTM!

PR #13235, opened 2024-03-12
Added a setting to allow the title's text alignment to be changed

ConorWebb96:
## Description
This allows the user to change the alignment of the apps title text on the page.

## Addresses
- POC enhancement that a customer asked for within their feedback.

## Screenshots
**Before**
Screenshot 2024-03-12 at 11 27 11

**After**
Screenshot 2024-03-12 at 11 25 55

**Gif of this in action**

## Launchcontrol
Small enhancement to allow the title text to be aligned; left, center or right depending on the users choice.


aptkingston (commented):
Looks grand as it is, but would look even better if you used buttons for this! You could use the `BarButtonList` component (which is what we use for other selects that we style as buttons).

It's nearly a drop in replacement - all you need to do is update your options array to be objects with a `value` and an `icon`. You can find nice spectrum icons for left, center and right align - check the paragraph component.

I reckon that's worth doing 👍

ConorWebb96:
Hey @aptkingston,

I've made that change, this is how the UI now looks.

aptkingston (approved):
Lovely 👌 was there any reason the `BarButtonList` component I suggested didn't work? It does exactly this, just in a ready-made component.

aptkingston (approved):
NAB but one little rename would be great 👌 

Then merge away - LGTM!

PR #13238, opened 2024-03-12
Update row.spec.ts to test against more real databases.

samwho:
## Description

Prior to this PR, `row.spec.ts` ran against CouchDB ("internal"), and Postgres ("external"). This PR expands the external databases tested to include:

- MySQL
- MariaDB
- SQL Server

In the process of getting these tests up and running, I discovered a bug in how JSON columns are handled for SQL Server and MariaDB. In both of these databases, there is no native support for JSON columns. Knex, the SQL builder library we use, falls back to storing JSON data as strings in these databases. However, we do not correctly parse these JSON strings when we fetch them back out.

This PR fixes the above by parsing JSON when fetching data as part of the row API. It doesn't make any changes to the query API, however, because the query API doesn't know the schema of the query being run and cannot accurately do this JSON parsing.

I attempted to also support Oracle in these tests, but the container for Oracle DB (`container-registry.oracle.com/database/free:23.3.0.0`) does not support ARM Macs, even using Rosetta (https://stackoverflow.com/questions/77952213/error-encountered-when-attempting-to-create-oracle-database-using-docker).

## Addresses
- https://linear.app/budibase/issue/BUDI-8067/sql-testing-more-datasource-types

## Launchcontrol

This PR fixes a bug in JSON column handling that was preventing user relationships from working correctly when using SQL Server or MariaDB.


packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -70,7 +77,7 @@ describe.each([
 
   const generateTableConfig: () => SaveTableRequest = () => {
     return {
-      name: uuid.v4(),
+      name: uuid.v4().substring(0, 16),
```
samwho, line 80: MySQL tests were failing due to identifiers being too long without this.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -477,7 +483,7 @@ describe.each([
       expect(row.Story).toBeUndefined()
       expect(row).toEqual({
         ...defaultRowFields,
-        OrderID: 1111,
+        OrderID: createRowResponse.OrderID,
```
samwho, line 486: SQL Server does not by default allow you to insert values into a primary key column.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -641,7 +647,7 @@ describe.each([
       const createdRow = await config.createRow()
 
       const res = await config.api.row.bulkDelete(table._id!, {
-        rows: [createdRow, { _id: "2" }],
+        rows: [createdRow, { _id: "9999999" }],
```
samwho, line 650: This was occasionally failing for me with the value `"2"`, unsure why, suspect race condition.

packages/server/src/integrations/base/sql.ts
```diff
@@ -691,6 +695,35 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
...
+        continue
+      }
+      const fullName = `${table.name}.${name}`
+      for (let row of results) {
+        if (typeof row[fullName] === "string") {
+          row[fullName] = JSON.parse(row[fullName])
+        }
+        if (typeof row[name] === "string") {
+          row[name] = JSON.parse(row[name])
+        }
```
samwho, line 713: I observed queries having both `<table name>.<column name>` and just `<column name>` in these objects, so I'm checking both here to cover the bases.

adrinr (approved):
LGTM, little comments left

packages/server/src/integrations/base/sql.ts
```diff
@@ -691,6 +695,35 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
...
+      }
+    }
+    return results
+  }
+
+  _isJsonColumn(field: FieldSchema) {
```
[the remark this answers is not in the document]
samwho: Could you show me an example of how that would work? I'm not sure that the check results in a clean type at the end of it.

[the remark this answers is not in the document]
adrinr: I realised that the field is not used after the check, as you only continue.
image

Anyways, in typescript we can use type guards to improve our typings. You can see that your code by default checks the type, but the `field` is still of type `FieldSchema`. 
image


As your function is checking the type of the field, you can infer that if `_isJsonColumn` returns true, we have either a JSON or a USERS column type. If you guard it as it follows:
image

the types will be properly inferred:
image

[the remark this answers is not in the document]
samwho: Thank you, I've added the type guards 🙏

PR #13239, opened 2024-03-12
Table/Tableblock deprecated and replaced with Gridblock

deanhannigan:
## Description

- `Deprecation` - `Table` and `Tableblock` components have been deprecated and can no longer be added to your apps. Existing apps making use of these components will continue to work after this update and **no action is required**.
- Tables in Budibase - Going forward, the `Gridblock` has been renamed to `Table` and will now provide the functionality of the legacy components. 
- Template Screens Rework - The previous Table/Grid templates have now been replaced to incorporate the new Table UX.
  - `Table with inline editing` - Generates a screen configured with a datasource containing a default title and the updated Table component. Inline CRUD functionality is enabled by default.
  - `Table with details panel` - Utilising the updated Table component, this screen template provides all the functionality of the legacy Tableblock. The row click behaviour is preconfigured to open a form side panel for editing.
- `Fix` - minor change to the app publish popover. The 'last published' time will now refresh.

## Addresses

https://linear.app/budibase/issue/BUDI-8049/consolidate-table-components

## Screenshots
New template types
Screenshot 2024-03-12 at 17 04 56

The new `Table` component is available in the data section
Screenshot 2024-03-13 at 09 14 04

Table with inline editing
Screenshot 2024-03-12 at 17 29 56

Table with details panel - open
Screenshot 2024-03-12 at 17 34 18

Table with details panel - closed
Screenshot 2024-03-12 at 17 34 48

## Launchcontrol
Table and Tableblock components have been deprecated. Apps with these components will continue to function as normal but these components can no longer be added after this update


shogunpurple (approved):
LGTM, nice work @deanhannigan

packages/builder/src/templates/gridDetailsScreen.js
```diff
@@ -0,0 +1,158 @@
...
+        parameters: {
+          id: createRowSidePanel._json._id,
+        },
+      },
+    ],
+    text: "Create row",
+    type: "cta",
+  })
+
+  buttonGroup.customProps({
```
aptkingston, line 50: We need to adda name to the button group, as currently it doesn't have one. I would move the name from the `createButton` to here instead, since you can't see the name of the button inside the button group.

aptkingston (approved):
LGTM! Just a few suggested naming changes, but all works perfectly.

One other thing that would be worth doing (in addition to flagging the components as deprecated in the manifest) is updating `packages/client/src/components/app/index.js` to move the newly deprecated table block and table down to the section where we put deprecated components. It's just another convention to help us know which components in there are to be left alone. We also move them into the dedicated "deprecated" folder.

aptkingston (approved):
LGTM!

PR #13240, opened 2024-03-12
Remove unused field verificationSent

melohagan:
## Description
`verificationSent` is not used, and will not be used in updated flow.

## Addresses
- https://github.com/Budibase/account-portal/pull/546
- https://linear.app/budibase/issue/GROW-355/worker-create-account


shogunpurple (approved):
LGTM

PR #13245, opened 2024-03-13
Add paths to builder TypeScript config for module resolution aliases with LSP

aptkingston:
## Description
This PR adds the [`paths`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths) option to the builder package TypeScript config. TypeScript config is used by the TypeScript language server to assist with module resolution in IDEs that use [LSP](https://microsoft.github.io/language-server-protocol/), such as VS Code, or other LSP editor plugins (such as one I'm currently using for Sublime Text).

The result of this change is that module resolution aliases support the native "go to definition" function, which is currently not supported.

In the builder we use module resolution aliases all over the place (the full list can be found in the Vite config, and now also the TypeScript config).

Before, failing to resolve the import:

https://github.com/Budibase/budibase/assets/9075550/1a79b06e-fa9b-4bb9-b77a-e4cc46d2ad19

After, resolving the import:

https://github.com/Budibase/budibase/assets/9075550/937af7b1-9168-4a43-bc7f-aa4a0deaedaf

(I have "go to definition" bound to ctrl+click in these examples)


aptkingston:
Don't believe there are any side effects of doing this (does not appear to affect the build in any way) but would appreciate your input @adrinr! Thank you 🙏

adrinr (commented):
What a great task to do, not sure how we survived so many years without this intellisense 😅
We are not using the tsconfig.json at all for the build process, so it should not have any impact on the actual build! So safe to go! 🎉 

Just noting that all the paths defined could be summarised (I reckon) as:
```
    "paths": {
      "assets/*": ["./assets/*"],
      "*": ["./src/*"]
    }
```

aptkingston:
> What a great task to do, not sure how we survived so many years without this intellisense 😅 We are not using the tsconfig.json at all for the build process, so it should not have any impact on the actual build! So safe to go! 🎉
> 
> Just noting that all the paths defined could be summarised (I reckon) as:
> 
> ```
>     "paths": {
>       "assets/*": ["./assets/*"],
>       "*": ["./src/*"]
>     }
> ```

Even better, that does work! I could have sworn I tried a few combination of more generic paths and it didn't work, but after testing it again I can't find any cases that don't work.

I've also added a few more generic paths for `@budibase` imports. BBUI and frontend-core imports never resolved properly with intellisense because LSP can't understand the `svelte` keyword in the package.json, so by adding these new paths:
```js
    "paths": {
      "assets/*": ["./assets/*"],
      "@budibase/*": [
        "../*/src/index.ts",
        "../*/src/index.js",
        "../*",
        "../../node_modules/@budibase/*"
      ],
      "*": ["./src/*"]
    }
```
It will now support intellisense for all `@budibase` imports, and actually navigate to the source code rather than the dist (in the case of shared-core, string-templates etc).

adrinr (approved):
Small change, great impact ❤️

PR #13247, opened 2024-03-13
Handling of files within the REST connector

PClmnt:
## Description
- The beginnings of the handling of files within REST connector
- Takes in a file, such as tar or image and stores it in object storage
- Returns an  `Upload` type including a signed url with a TTL
- Adds a new bucket within the object store called `tmp-file-attachments`. Within that the object will be of a path such as `app_tenant_9df1c2ba1be741fd883ce5651ffa9d26/yourfile.txt`
- Adds a TTL of 1 day (the minimum) on the bucket so objects will expire

## Addresses
https://linear.app/budibase/issue/BUDI-8079/rest-query-handles-file-download

## Screenshots
image

Example of what is returned whenever you call an endpoint that returns a file, here I uploaded a basic text file and received the signed url for it.


packages/server/src/integrations/rest.ts
```diff
@@ -129,42 +130,82 @@ class RestIntegration implements IntegrationBase {
...
+            data = data[keys[0]]
+          }
+          raw = rawXml
+        } else {
+          data = await response.text()
+          raw = data
+        }
       }
     } catch (err) {
+      console.log(err)
```
adrinr, line 205: Remove

packages/server/src/integrations/rest.ts
```diff
@@ -129,42 +130,82 @@ class RestIntegration implements IntegrationBase {
...
-          (await xmlParser(rawXml, {
-            explicitArray: false,
-            trim: true,
-            explicitRoot: false,
-          })) || {}
-        // there is only one structure, its an array, return the array so it appears as rows
-        const keys = Object.keys(data)
-        if (keys.length === 1 && Array.isArray(data[keys[0]])) {
-          data = data[keys[0]]
+      if (filename) {
```
adrinr, line 142: This function is quite long and hard to read. Can we create subfunctions to make it more verbose?

PClmnt: I split out both the file and xml handling part of this function into utils.

packages/server/src/integrations/rest.ts
```diff
@@ -129,42 +130,40 @@ class RestIntegration implements IntegrationBase {
...
 
   async parseResponse(response: any, pagination: PaginationConfig | null) {
-    let data, raw, headers
+    let data, raw, headers, filename
+
     const contentType = response.headers.get("content-type") || ""
+    const contentDisposition = response.headers.get("content-disposition") || ""
+    const matches =
+      /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/.exec(contentDisposition) || []
+    filename = matches[1]?.replace(/['"]/g, "") || ""
```
samwho, line 139: Love a good regex.

This one doesn't appear to perfectly follow the standard for the filename parameter for `Content-Disposition`, though, sadly. Specifically this bit (from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename_2):

CleanShot 2024-03-14 at 14 14 11@2x

The regex takes the first thing that matches, which may or may not be the `filename*` variant.

https://stackoverflow.com/a/67994693/205770 seems to offer a potentially more robust solution, alternately https://www.npmjs.com/package/content-disposition.

We should also be wary of path-traversal attacks. In its current state, the code will parse `filename="../../../../passwd"`. I'm not sure how bad that is for MinIO but I can't imagine any legitimate reason to attempt to use `../` in filenames here, so I'd err on the side of blocking them.

packages/backend-core/src/objectStore/utils.ts
```diff
@@ -2,6 +2,7 @@ import { join } from "path"
 import { tmpdir } from "os"
 import fs from "fs"
 import env from "../environment"
+import { PutBucketLifecycleConfigurationRequest } from "aws-sdk/clients/s3"
```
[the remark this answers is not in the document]
PClmnt: Minio is totally S3 compliant and we don't import the actual minio package anywhere as far as I can tell? When creating the object store we just use the AWS.S3 library also.

packages/server/src/integrations/utils.ts
```diff
@@ -440,3 +445,59 @@ export function removeEmptyFilters(filters: SearchFilters) {
...
+      size: responseBuffer.byteLength,
+      name: processedFileName,
+      url: presignedUrl,
+      extension: fileExtension,
+      key: key,
+    },
+    info: {
+      code: response.status,
+      size: formatBytes(responseBuffer.byteLength),
+      time: `${Math.round(performance.now() - startTime)}ms`,
```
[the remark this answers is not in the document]
PClmnt: I was just maintaining consistency with what previously requests had, but I guess we can remove it.

packages/server/src/integrations/tests/rest.spec.ts
```diff
@@ -611,4 +633,56 @@ describe("REST Integration", () => {
...
+    const responseData = Buffer.from("test file content")
+    const filename = "test.tar.gz"
+    const contentType = "application/gzip"
+
+    ;(fetch as unknown as jest.Mock).mockImplementationOnce(() =>
+      Promise.resolve({
+        headers: {
+          raw: () => ({
+            "content-type": [contentType],
+            "content-disposition": [`attachment; filename="${filename}"`],
```
samwho, line 647: Would be good to have a test that also sets a `filename*` that has some non-ascii characters in it to make sure we don't choke on it, and we correctly prioritise the name.

packages/server/src/integrations/tests/rest.spec.ts
```diff
@@ -611,4 +633,56 @@ describe("REST Integration", () => {
     expect(calledConfig.headers).toEqual({})
     expect(calledConfig.agent.options.rejectUnauthorized).toBe(false)
   })
+
+  it("uploads file to object store and returns signed URL", async () => {
+    const responseData = Buffer.from("test file content")
+    const filename = "test.tar.gz"
+    const contentType = "application/gzip"
+
+    ;(fetch as unknown as jest.Mock).mockImplementationOnce(() =>
```
samwho, line 642: This is a bit grim, isn't it? Can we not spin up a `minio` container and actually do a proper upload? We spin containers up quite frequently for tests against various databases 🤔

packages/server/src/integrations/utils.ts
```diff
@@ -440,3 +445,59 @@ export function removeEmptyFilters(filters: SearchFilters) {
   }
   return filters
 }
+
+export async function handleXml(response: any) {
+  let data
+  const rawXml = await response.text()
+  data =
+    (await xmlParser(rawXml, {
```
samwho, line 453: I'm a little concerned about parsing untrusted XML. I'm aware of attacks like https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html but I don't really see any mention of external entities in the `xml2js` library (https://github.com/Leonidas-from-XIV/node-xml2js/issues?q=is%3Aissue+external+).

The underlying library that `xml2js` is using (`sax-js`) appears to have an allowlist of entities and doesn't support external or custom ones: https://github.com/isaacs/sax-js/blob/main/lib/sax.js#L355. So we're probably safe? 🤔

packages/server/src/integrations/utils.ts
```diff
@@ -440,3 +445,59 @@ export function removeEmptyFilters(filters: SearchFilters) {
...
+}
+
+export async function handleFileResponse(
+  response: any,
+  filename: string,
+  startTime: number
+) {
+  let presignedUrl
+  const responseBuffer = await response.arrayBuffer()
+  const fileExtension = filename.includes(".")
```
samwho, line 473: `filename` _could_ include directory information, and directory names _could_ include dots. I think we probably want to either reject directories, strip them, or handle them somehow in this check.

packages/server/src/integrations/utils.ts
PClmnt: So this was already in the codebase. Adria just commented that the top level function was too messy so I moved everything out. I think it's been there for like 3 years :D

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -146,12 +151,20 @@ export async function upload({
...
 }: UploadParams) {
   const extension = filename.split(".").pop()
-  const fileBytes = fs.readFileSync(path)
-
+  const fileBytes = path ? fs.readFileSync(path) : body
   const objectStore = ObjectStore(bucketName)
-  await makeSureBucketExists(objectStore, bucketName)
+
+  // If this is true the bucket existed previously - so we can conditionally add TTL
+  const bucketExisted = await makeSureBucketExists(objectStore, bucketName)
```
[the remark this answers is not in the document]
PClmnt: Fixed!

packages/server/src/integrations/tests/rest.spec.ts
PClmnt: Added a test container in so we're using live data. Although this specific part is mocking out the fetch implementation, do you mean grim as in the `unknown` cast?

packages/server/src/integrations/rest.ts
PClmnt: Used the content-disposition library which internally handles the `filename` / `filename*` check.

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -146,12 +157,20 @@ export async function upload({
   path,
   type,
   metadata,
+  body,
+  addTTL,
 }: UploadParams) {
   const extension = filename.split(".").pop()
-  const fileBytes = fs.readFileSync(path)
+
+  const fileBytes = path ? (await fsp.open(path)).createReadStream() : body
```
PClmnt, line 165: Like this here @samwho ?

samwho: This looks perfect. You'll want to move from using `objectStore.upload` to `objectStore.streamUpload` further down as well 👍

packages/server/src/integrations/rest.ts
```diff
@@ -133,10 +135,15 @@ class RestIntegration implements IntegrationBase {
     let data, raw, headers, filename
 
     const contentType = response.headers.get("content-type") || ""
+    console.log("content-type: " + contentType)
```
samwho: 👀

packages/server/src/integrations/tests/rest.spec.ts
```diff
@@ -634,55 +625,93 @@ describe("REST Integration", () => {
...
+    afterAll(async () => {
+      await databaseTestProviders.s3.stop()
+    })
+
+    it("uploads file to object store and returns signed URL", async () => {
+      const responseData = Buffer.from("teest file contnt")
+      const filename = "test.tar.gz"
+      const contentType = "application/gzip"
+
+      ;(fetch as unknown as jest.Mock).mockImplementationOnce(() =>
```
samwho, line 642: No action required here, but one day I'd love to remove our `node-fetch` mocks and replace it all with https://www.npmjs.com/package/nock

packages/server/src/integrations/tests/utils/index.ts
```diff
@@ -14,4 +15,4 @@ export interface DatabaseProvider {
   datasource(): Promise<Datasource>
 }
 
-export const databaseTestProviders = { postgres, mongodb, mysql }
+export const databaseTestProviders = { postgres, mongodb, mysql, s3 }
```
samwho, line 18: The `s3` provider doesn't _quite_ fit the interface expected by `DatabaseProvider` above, as it doesn't have a `datasource(): Promise<Datasource>` function. It may be better to split the `s3` provider out into its own `objectStoreTestProviders` package.

samwho (approved):
This is looking great. Thanks for your patience with my comments. Just a couple of extra small suggestions, but nothing I'd consider blocking.

packages/server/src/integrations/tests/utils/index.ts
PClmnt: Sorted, with the move to minio image we now call like this`objectStoreTestProviders.minio.start()`

packages/backend-core/src/objectStore/objectStore.ts
PClmnt: nvm sorted it

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -105,37 +110,43 @@ export function ObjectStore(
...
+  bucketName: string
+): Promise<boolean> {
   bucketName = sanitizeBucket(bucketName)
   try {
     await client
       .headBucket({
         Bucket: bucketName,
       })
       .promise()
+    return false
```
adrinr, line 124: This return might be confusing? It means it exists, that it was created...? If needed, can we type it to something like `return { created: false, exists: true}`?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -105,37 +110,43 @@ export function ObjectStore(
...
         promises[bucketName] = client
           .createBucket({
             Bucket: bucketName,
           })
           .promise()
         await promises[bucketName]
         delete promises[bucketName]
+        return true
+      } else {
+        throw new Error("Access denied to object store bucket.")
```
adrinr, line 143: We should throw the original error, just in case we are hiding something

adrinr (changes requested):

packages/server/src/integrations/tests/rest.spec.ts
samwho: I had originally mistaken this to be related to the objectStore but I realise now it's not. Breaking apart our `node-fetch` mock is a task for another time 😄

packages/backend-core/src/objectStore/objectStore.ts
samwho: I'd originally suggested to Peter to rename this function to make the return a bit more clear, it says whether or not the file was created. You're right, though, it could just as easily represent if it already existed. Not sure how to make this function hard to use incorrectly.

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -138,9 +154,9 @@ export async function createBucketIfNotExists(
           .promise()
         await promises[bucketName]
         delete promises[bucketName]
-        return true
+        return { created: true, exists: false }
       } else {
-        throw new Error("Access denied to object store bucket.")
+        throw new Error("Access denied to object store bucket." + err)
```
samwho, line 159: I'd recommend using the `cause` param here instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#rethrowing_an_error_with_a_cause, it'll preserve the stack trace of the cause that way.

```js
> const e1 = new Error("foo")
undefined
> const e2 = new Error("bar", { cause: e1 })
undefined
> e1
Error: foo
    at REPL1:1:12
    at ContextifyScript.runInThisContext (node:vm:121:12)
    at REPLServer.defaultEval (node:repl:599:22)
    at bound (node:domain:432:15)
    at REPLServer.runBound [as eval] (node:domain:443:12)
    at REPLServer.onLine (node:repl:929:10)
    at REPLServer.emit (node:events:526:35)
    at REPLServer.emit (node:domain:488:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
    at [_line] [as _line] (node:internal/readline/interface:887:18)
> e2
Error: bar
    at REPL2:1:12
    at ContextifyScript.runInThisContext (node:vm:121:12)
    ... 7 lines matching cause stack trace ...
    at [_line] [as _line] (node:internal/readline/interface:887:18) {
  [cause]: Error: foo
      at REPL1:1:12
      at ContextifyScript.runInThisContext (node:vm:121:12)
      at REPLServer.defaultEval (node:repl:599:22)
      at bound (node:domain:432:15)
      at REPLServer.runBound [as eval] (node:domain:443:12)
      at REPLServer.onLine (node:repl:929:10)
      at REPLServer.emit (node:events:526:35)
      at REPLServer.emit (node:domain:488:12)
      at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
      at [_line] [as _line] (node:internal/readline/interface:887:18)
}
```

samwho (approved):

packages/backend-core/src/objectStore/objectStore.ts
PClmnt: This is only available with TS 2022 declarations - we only are at 2020 having a look at `tsconfig.build.json`. And I don't feel comfortable just changing that for this one thing, might have larger ramifications and break stuff?

adrinr (approved):
🎉

PR #13248, opened 2024-03-13
Reduce `row.spec.ts`'s reliance on the `TestConfiguration` helpers.

samwho:
## Description

We're moving away from the helpers in `TestConfiguration` and instead using our various `*API` classes instead. This PR pulls out a lot of uses of `TestConfiguration` helpers from `row.spec.ts`, making them much more independent.


adrinr (approved):
🎉  nice cleanup!

PR #13250, opened 2024-03-13
Convert view.spec.js to TypeScript, remove reliace on TestConfiguration.

samwho:
## Description

Noticed this was still JavaScript, decided to convert it while I was in the area.


adrinr (approved):
🚀

PR #13252, opened 2024-03-14
Remove flakiness from unit tests

adrinr:
## Description
We are facing some flakiness on some docthrough tests. We are generating a lot of data, and some keys repeat between objects, causing flakiness. Changing the keys to guids will fix the errors


samwho (approved):
I almost want to ban `generator.word()`. It has been one of the most common causes of flakiness in our tests.

PR #13266, opened 2024-03-15
Move viewV2 tests out of row.spec.ts and into viewV2.spec.ts.

samwho:
## Description

Pretty much what it says on the tin. `row.spec.ts` is _massive_ and it's very reasonable for these tests to live in `viewV2.spec.ts`, given that they're exercising view-specific functionality (even though they're operating on rows).

I've removed some duplication and simplified some of the tests in the move, hence the negative delta.


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -525,4 +543,475 @@ describe.each([
       expect(row.Country).toEqual("Aussy")
     })
   })
+
+  describe.only("row operations", () => {
```
adrinr, line 547: Remove:
```suggestion
  describe("row operations", () => {
```

samwho: The shame. I might write a lint for this.

adrinr (approved):

PR #13272, opened 2024-03-18
Enable eslint jest plugin.

samwho:
## Description

I left a `.only` in a test in https://github.com/Budibase/budibase/pull/13266 and felt embarrassed about it, so thought I'd write a lint to help me not do that again.

Then @adrinr pointed out that the official eslint jest plugin has a check for this already, and we should use that. So I did, and I also turned on their recommended lint set.

Pro PR:
- https://github.com/Budibase/budibase-pro/pull/280

Account portal PR:
- https://github.com/Budibase/account-portal/pull/552


.eslintrc.json
```diff
@@ -43,9 +43,24 @@
         "no-useless-escape": "off",
         "no-undef": "off",
         "no-prototype-builtins": "off",
-        "local-rules/no-budibase-imports": "error",
+        "local-rules/no-budibase-imports": "error"
+      }
+    },
+    {
+      "files": ["**/*.spec.ts"],
```
adrinr, line 50: Should we use an existing plugin such as `eslint-plugin-jest` instead?

https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-focused-tests.md

adrinr: We could probably also benefit form the other recommended rules

samwho: I'm in the process of enabling the TypeScript recommended rules in another PR. I'll add the Jest ones in there, too. Good call!

.eslintrc.json
```diff
@@ -43,9 +44,34 @@
...
```
adrinr, line 51: Can we add the same plugins (except TS) to `spec.js` as well?

packages/backend-core/scripts/test.sh
```diff
@@ -4,10 +4,10 @@ set -e
...
   # --runInBand performs better in ci where resources are limited
-  echo "jest --coverage --runInBand --forceExit"
-  jest --coverage --runInBand --forceExit
+  echo "jest --coverage --runInBand --forceExit $@"
+  jest --coverage --runInBand --forceExit $@
 else
   # --maxWorkers performs better in development
-  echo "jest --coverage --detectOpenHandles"
-  jest --coverage --detectOpenHandles
+  echo "jest --coverage --forceExit --detectOpenHandles $@"
```
adrinr, line 11: Why is this needed? Don't we actually want to detect the handles?

packages/server/src/api/routes/tests/appImport.spec.ts
```diff
@@ -25,8 +25,8 @@ describe("/applications/:appId/import", () => {
       .expect(200)
     const appPackage = await config.api.application.get(appId!)
     expect(appPackage.navigation?.links?.length).toBe(2)
-    expect(expect(appPackage.navigation?.links?.[0].url).toBe("/blank"))
```
adrinr, line 28: 🤯

packages/server/src/integrations/tests/sql.spec.ts
```diff
@@ -342,25 +342,6 @@ describe("SQL query builder", () => {
     })
   })
 
-  it("should use greater than when only low range specified", () => {
```
adrinr, line 345: Was this duplicated?

samwho: Yes, there is a lint that fires on identical strings in `it` and `describe` blocks and there is an identical test elsewhere in this file.

.eslintrc.json
samwho: Good idea. Do you mind if I do that in a follow-up PR, to keep the size down on this one?

.eslintrc.json
```diff
@@ -34,6 +34,7 @@
     },
     {
       "files": ["**/*.ts"],
+      "excludedFiles": ["qa-core/**"],
```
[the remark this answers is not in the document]
samwho: Soon. Soon.

packages/backend-core/scripts/test.sh
adrinr: Sorry, I meant the `--forceExit`

.eslintrc.json
adrinr: Yes please, otherwise you will start a new rabbit hole ahah

packages/backend-core/scripts/test.sh
samwho: Ohh sorry, I ran these tests once and they got stuck exiting because of an open Redis client. I did this out of laziness and because it's done elsewhere.

adrinr (approved):
image

PR #13283, opened 2024-03-19
Fix hard crash when an automation error exists

aptkingston:
## Description
This PR fixes a hard page crash due to a recently introduced bug. Currently the page will crash if you have a recent automation error, which shows this warning: 
image

This small changes fixes the crash.

Pretty important to roll this out ASAP as it completely blocks usage of Budibase if you have an automation error.


deanhannigan (approved):
LGTM

PR #13286, opened 2024-03-19
Align gh pages with master

adrinr:
## Description
Updating gh-pages to not be so much behind and updating the .gitignore in order to keep only the wanted docs

image


samwho (approved):
It's annoying that `gh-pages` is a branch off `master` at all. We should, at some point, create a new orphan branch (a la https://stackoverflow.com/questions/32596078/how-to-detach-branch-from-the-parent-in-git) and move our Helm repo over to using that instead.

PR #13287, opened 2024-03-19
Reenable no-useless-escape.

samwho:
## Description

Reenables the `no-useless-escape` lint rule, not sure why it was disabled to begin with and not many violations throughout the codebase.

Aiming to get to a point where we have all eslint recommended lints enabled.


adrinr (approved):
Is this even a rule? Who would want to have useless escapes? 😅

samwho:
Us, apparently 😂

PR #13288, opened 2024-03-19
Reenable no-inner-declarations.

samwho:
## Description

Reenables the `no-inner-declarations` lint rule, not sure why it was disabled to begin with and not many violations throughout the codebase.

Aiming to get to a point where we have all eslint recommended lints enabled.


packages/backend-core/src/logging/pino/logger.ts
```diff
@@ -10,6 +10,18 @@ import { formats } from "dd-trace/ext"
 
 import { localFileDestination } from "../system"
 
+function isPlainObject(obj: any) {
```
adrinr, line 13: Should we keep them inner as `const`?

samwho: Because these were pure functions that didn't depend on the surrounding context, I opted to put them at the top level. Don't particularly mind either way, but if there's no need for them to be inner I lean slightly toward them not being inner.

adrinr (approved):

Issue #13259, opened 2024-03-15
Handlebars in query parameters/bindings gets processed


CSLTech:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Describe the bug**
We have an API that takes in handlebars patterns in some JSON properties. However, it seems like the backend of budibase is processing those even if they shouldn't be. This happens **IN ADDITION** the configured body format.

For example, if I have a payload that should look like:
`{"test": "{{ thisIsSomeValue }}"}`
Then what budibase sends as a REST request with a payload that looks like:
`{"test": ""}`

I also validated that it isn't simply sanitized as some static values do get processed correctly, for example:
`{"test": "{{ 'abcdef' }}"}`
Becomes
`{"test": "abcdef"}`

**To Reproduce**
1. Create a REST API datasource
2. Create a body format that includes some bindings
3. Send in Handlebars formatted data as one of those bindings
4. Bad request

**Expected behavior**
Strings should be passed as-is without further processing from the backend.

**Screenshots**

Bindings: 
![image](https://github.com/Budibase/budibase/assets/2613776/aea16caa-280e-49be-9557-0cc33d63af77)
Body:
![image](https://github.com/Budibase/budibase/assets/2613776/27c5d6c0-6a0b-4e99-ab10-e2ea3f292e64)

Sample Call:
![image](https://github.com/Budibase/budibase/assets/2613776/a1cf56ad-5f6f-4a60-9ff9-49c1445186a1)
Response from postman echo:
![image](https://github.com/Budibase/budibase/assets/2613776/3a64492f-b2a2-4ed3-bd41-acca80e08a0f)

**Additional context**
I haven't poked at the context that budibase binds there, but this could lead to security issue if anything private is in this context. An end user could easily generate REST requests with any value in this context and leak them out.

CSLTech:
Please note that I tried escaping it, but couldn't find a good way to get the properly formatted versions output. \{{ value }} ends up as {{ _all_ value }} (For some reason) And quadruple brackets don't get processed at all and don't output a double bracket.

CSLTech:
After a bit more poking around and dumping the context. I can confirm that I can dump sessionIds, CSRF Tokens, etc.

mike12345567:
Hi @CSLTech - this is all intended functionality - I don't think there will be an easy way in Budibase to send Handlebars like statements - we will always process them.

[...]

PR #13289, opened 2024-03-19
Fix for query inputs allowing handlebars

mike12345567:
## Description
This issue details the problem: https://github.com/Budibase/budibase/issues/13259

Users can input handlebars as part of a query input parameter, which is allowed and will be processed. Adding in validation to the API to disallow these inputs as they can cause problems, but we need to process the input context for secrets.


shogunpurple (approved):
LGTM!

adrinr (approved):

PR #13291, opened 2024-03-19
Allow component deselect in preview. Remove screen highlight.

deanhannigan:
## Description

Minor change to remove the selection border from around the screen component.
- Clicking in the empty space around the preview components will now "deselect" the currently selected component
- The border for the `Screen` component is now excluded in the preview, but it still marked as selected in the left navigation.

## Addresses
-  https://linear.app/budibase/issue/BUDI-7676/design-section-remove-selected-state-for-screen-in-the-preview

## Screenshots
Update UX

Previous UX would also highlight the screen.

## Launchcontrol

Clicking in the empty space around builder preview will deselect the currently selected component without adding a border to the entire screen.


packages/client/src/components/app/Layout.svelte
```diff
@@ -281,7 +281,14 @@
         </div>
       </div>
     {/if}
-    <div class="main-wrapper">
+    <div
+      class="main-wrapper"
+      on:click={() => {
+        if ($builderStore.inBuilder) {
+          builderStore.actions.selectComponent(null)
```
aptkingston, line 288: Could we send up `$builderStore.screen._id` here instead? Sending up `null` causes the builder to do an immediate redirect to the screen ID anyway, but we have the screen ID here so we can save the builder the extra work (and slight delay).

packages/client/src/components/preview/SelectionIndicator.svelte
```diff
@@ -5,10 +5,14 @@
...
     : "var(--spectrum-global-color-static-blue-600)"
+
+  $: isScreen = $builderStore.selectedComponentId?.startsWith(
+    $builderStore.screen._id
+  )
 </script>
 
 <IndicatorSet
-  componentId={$builderStore.selectedComponentId}
+  componentId={!isScreen ? $builderStore.selectedComponentId : null}
```
aptkingston, line 15: We can remove any changes to this file if we opt to take my suggested approach.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte
```diff
@@ -130,7 +130,7 @@
       // Catch any app errors
       loading = false
       error = event.error || "An unknown error occurred"
-    } else if (type === "select-component" && data.id) {
+    } else if (type === "select-component" && "id" in data) {
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte:133]: We can remove any changes to this file if we opt for sending up the screen ID as opposed to `null`.

aptkingston (commented):
One quirk with this approach is that hovering over the screen in the component list still shows an indicator around the whole screen, but when you select it you get nothing.

An easier way to implement this is simply removing the `{screenId}` and `{screenId}-dom` class names from the top level nodes in the layout. That will mean that there isn't a matching DOM node for the screen ID, so neither the hover indicator nor selection indicator will show a border around the whole screen, and it doesn't require any changes to those files either 👍 

We'll still need the new on:click handler in the layout to send up the screen ID, but that's it.

packages/client/src/components/preview/SelectionIndicator.svelte
```diff
@@ -5,10 +5,14 @@
   $: color = $builderStore.editMode
     ? "var(--spectrum-global-color-static-green-500)"
     : "var(--spectrum-global-color-static-blue-600)"
+
+  $: isScreen = $builderStore.selectedComponentId?.startsWith(
+    $builderStore.screen._id
+  )
```
aptkingston, line 11: We don't need these changes, since there is no longer any matching DOM nodes for the screen ID, so we can always just let the indicator render with whatever the selected component ID is. Think I said that in my initial review but must have just forgotten to remove it 👍

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte
```diff
@@ -125,7 +125,7 @@
...
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPreview.svelte:128]: Since we aren't sending up nulls any more, we should probably revert this too to ensure we don't ever store a null unintentionally.

packages/client/src/components/preview/SelectionIndicator.svelte
deanhannigan: Apologies, I completely forgot to add this to the commit.

packages/client/src/components/app/Layout.svelte
```diff
@@ -169,15 +169,15 @@
 <!-- svelte-ignore a11y-no-static-element-interactions -->
 <!-- svelte-ignore a11y-click-events-have-key-events -->
 <div
-  class="component {screenId} layout layout--{typeClass}"
+  class="component layout layout--{typeClass}"
   use:styleable={$component.styles}
```
aptkingston, line 173: Just actually tested this, and realised we need to also remove the `use:styleable` helper here (it adds back in the class name, and it's not needed since we don't allow screen styles). Right now you still get the selected border around the screen because of this.

Sorry about so many reviews on this one.

packages/client/src/components/app/Layout.svelte
```diff
@@ -282,7 +282,14 @@
         </div>
       </div>
     {/if}
-    <div class="main-wrapper">
+    <div
+      class="main-wrapper"
+      on:click={() => {
+        if ($builderStore.inBuilder) {
+          builderStore.actions.selectComponent($builderStore.screen._id)
```
aptkingston, line 289: This should be the `screenId` variable that we already have above. The correct ID for the screen being selected is the screen ID suffixed by `-screen`. This still works because the builder thinks this is invalid so will actually redirect to the correct ID, but it caused a double URL update so would be better to update this to send `screenId` instead

aptkingston (approved):
LGTM!

aptkingston:
Happy to merge this as soon as the conflict is resolved.

Issue #13282, opened 2024-03-19
Date field type not displaying a value, however values are being stored in db for Postgres databases [BUILDER]


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.22.1 
  - App Version: 2.22.1

**Describe the bug**
Date fields which are created within external databases won't display their dates in the builder or within the design screen. These can be seen at the database level. Looking at the network requests these fields aren't even being pulled in, they seem to be getting stripped/removed from the network rows object.

**To Reproduce**
1. Create a table using the below create script
2. Add a record to it via the builder or form block
3. See that record is added to database
4. See that no date only fields are being displayed in the builder or the design view.

**Expected behavior**
Date fields should be pulled in and should be able to be seen/displayed.

**Screenshots**
No **accrual_worksheet_received_date** field shown in the network request
<img width="1512" alt="Screenshot 2024-03-19 at 08 41 51" src="https://github.com/Budibase/budibase/assets/126772285/0df4de17-5c92-44e1-982e-b474201487fa">

Database level entry showing the value inputted.
<img width="1024" alt="Screenshot 2024-03-19 at 08 42 47" src="https://github.com/Budibase/budibase/assets/126772285/9c3838cf-434a-4def-bcd8-39ea239d5415">

**App Export**
```sql
CREATE TABLE property_financial_statement_tracker (
    id serial4 NOT NULL,
    property_name text NOT NULL,
    financial_month timestamp with time zone NOT NULL DEFAULT current_timestamp,
    accounting_preparer text NULL,
    accrual_worksheet_received_date date NULL,
    financials_prepared_review_pending date NULL,
    CONSTRAINT property_financial_statement_tracker_pkey PRIMARY KEY (id),
    CONSTRAINT property_financial_statement_tracker_unique UNIQUE (property_name, financial_month)
);
```

**Additional context**
If an engineer wants a pre-setup database, let me know and I'll export an app and send it across for them to see.

mike12345567:
Fixed in release 2.22.4.

PR #13292, opened 2024-03-19
Fix issue with old datasources not having 'isSQL' flag

mike12345567:
## Description
Fix for issue with aliasing not quite working as expected when interacting with very old datasources, there is a flag 'isSQL' which was not set in old versions, this is now set when retrieving datasources to avoid issues with it being unset.

This had an impact on aliasing, but this may have affected a variety of areas.

Addresses: https://github.com/Budibase/budibase/issues/13282


samwho (approved):
Possible to add a test for this?

mike12345567:
> Possible to add a test for this?

I've updated a test now to make sure the `isSQL` is responded with as expected!

adrinr (approved):

Issue #13276, opened 2024-03-18
Budi CLI Broken with Redis Errors and Hanging


pbarmak:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: budi cli
  - Budibase Version: latest

**Describe the bug**
Per this discussion thread:
https://github.com/Budibase/budibase/discussions/13269

And this other bug I found while researching bugs (which confirms this issue in the first paragraph):
https://github.com/Budibase/budibase/issues/13227

It seems the cli has some major issue on install. I have tried 3 new server instances (one arm64, two amd64) with multiple versions of node. Any `budi hosting` command gives a ton of errors and hangs the session (have to log in as another session and kill the thread). I even went as far as installing a redis-server on the machine, just to see, and the cli still hangs (doesn't give the redis error, but hangs). So I doubt the error is the real reason, but here it is, just in case it helps debug:
```
[11:39:04.738] ERROR (5840): [BULL] doc-writethrough=error {"tenantId":"default","bull":{"eventType":"doc-writethrough","event":"error"}}
    err: {
      "type": "Error",
      "message": "connect ECONNREFUSED 127.0.0.1:6379",
      "stack":
          Error: connect ECONNREFUSED 127.0.0.1:6379
              at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1555:16)
      "errno": -111,
      "code": "ECONNREFUSED",
      "syscall": "connect",
      "address": "127.0.0.1",
      "port": 6379
    }
```

In addition, the documentation seems to be incorrect on the install steps. First, I don't think installing only docker.io works - you have to also install docker-compose. Second, the version for the node setup (the curl statement) is way off, node gives a warning. That should be updated. Third, after the setup line, I think you also need to include installing node.js. Finally, running `npm install -g @budibase/cli` without sudo fails. Again, it could be due to whatever bug this is, as I don't recall having that issue before (but I don't remember). I'm just noting all this to help make the docs more complete, since the CLI seems to be the best way for us who don't know docker well to install as it handles backups and restores (I don't see documentation on backing up and restoring everything including users using just docker-compose).

**To Reproduce**
1. Create a new server instance, do not install redis-server on the host
2. Just install following the documentation: https://docs.budibase.com/docs/budibase-cli-setup
3. Try running budi hosting --init
4. Watch it give lots of errors

**Expected behavior**
It creates the docker-compose.yaml file and exits gracefully

[...]

PR #13294, opened 2024-03-19
[Fix] CLI broken with redis errors and hanging

adrinr:
## Description
The new doc-writethrough class is declared in backend-core, and the queue gets initialised by default. We are having issues with CLI, as it tries to initialise this queue (even if it's not needed) but the redis connection might not exist.

image

Updating this class in order to create the queue when actually required.

## Addresses
- Linear: [BUDI-8100 - Budi CLI Broken with Redis Errors and Hanging
](https://linear.app/budibase/issue/BUDI-8100/budi-cli-broken-with-redis-errors-and-hanging)
- https://github.com/Budibase/budibase/issues/13276

## Launchcontrol
Fix CLI issue with the doc-writethrough initialisation

### DIFF
--- a/packages/cli/tsconfig.build.json
+++ b/packages/cli/tsconfig.build.json
@@ -11,6 +11,7 @@
     "types": ["node", "jest"],
     "outDir": "dist",
     "skipLibCheck": true,
+    "baseUrl": ".",
     "paths": {
       "@budibase/types": ["../types/src"],
       "@budibase/backend-core": ["../backend-core/src"],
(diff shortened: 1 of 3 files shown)


packages/cli/tsconfig.build.json
```diff
@@ -11,6 +11,7 @@
     "types": ["node", "jest"],
     "outDir": "dist",
     "skipLibCheck": true,
+    "baseUrl": ".",
```
mike12345567, line 14: Why is this needed out of interest?

mike12345567 (approved):

packages/cli/tsconfig.build.json
adrinr: This is needed in order to properly use the tsconfig-paths: 
image

PR #13305, opened 2024-03-20
Fix unit test flakiness

adrinr:
## Description
Fix flakiness that caused some pipelines to fail randomly: https://github.com/Budibase/budibase/actions/runs/8355591637/job/22870976528

The current test will fail when the generated app id ends with `a`


samwho (approved):
🙈

deanhannigan (approved):
LGTM

PR #13310, opened 2024-03-20
Remove worker aws-sdk, oauth, node-fetch mocks.

samwho:
## Description

I've been thinking about it seeing seeing `nock` in our CouchDB client code, so decided to finally give it a try. I really like it, vastly prefer it to the `__mocks__/node-fetch.ts` we have in the server and worker packages.

This PR introduces `nock` into `packages/worker`, and also removes some other mocks that aren't strictly necessary.


packages/worker/src/api/routes/global/tests/auth.spec.ts
```diff
@@ -352,10 +313,43 @@ describe("/api/global/auth", () => {
...
+        nock("http://localhost").get("/userinfo?schema=openid").reply(200, {
+          sub: "sub",
+          email: "oauth@example.com",
+        })
+
         const configId = await generateOidcConfig()
         const preAuthRes = await config.api.configs.getOIDCConfig(configId)
-
         const res = await config.api.configs.OIDCCallback(configId, preAuthRes)
+        if (res.status > 399) {
```
mike12345567, line 348: I usually prefer `res.status >= 400` - bit clearer that its truly only looking for error status!

mike12345567 (approved):
LGTM - nice improvements here!

PR #13311, opened 2024-03-20
Fix requiring double keypress to create new rows

aptkingston:
## Description
Fix requiring double keypress to create new rows. Now pressing `ctrl/cmd+enter` will always attempt to create a new row regardless of whether you're editing a field or not.

## Addresses
- https://linear.app/budibase/issue/BUDI-8056/double-tap-of-keyboard-required-when-saving-rows


shogunpurple (approved):
Super, this is a handy one. Always run into it!

PR #13316, opened 2024-03-20
Remove mysql mock

samwho:
## Description

Following on the work described in https://github.com/Budibase/budibase/pull/13314. We had a suite of MySQL tests that relied on mocks and overlapped significantly with the things being tested by the MySQL Query tests, so I've folded the former into the latter and deleted the mocks.


shogunpurple (approved):
LGTM

PR #13317, opened 2024-03-20
BUDI-8084 - Single attachment binding as subtype

adrinr:
## Description
Changing single attachments to have a subtype in order to be cleaner on its checks, not relying only on the max length only. This will not be used yet, but it might be used on bindings and other places. Having this subtype single can be extended in the future, as we can have single/multiple in multiple types, such as files, users, relationships, etc.

This change made me fix some types around the utils, getting a cleaner and safer usage of it

## Feature branch env
[Feature Branch Link](http://fb-budi-8084-single-attachment-binding.fb.qa.budibase.net)


packages/server/src/api/controllers/table/utils.ts
```diff
@@ -88,6 +89,24 @@ export async function checkForColumnUpdates(
     // Update views
     await checkForViewUpdates(updatedTable, deletedColumns, columnRename)
   }
+
+  for (const attachmentColumn of Object.values(updatedTable.schema).filter(
+    column =>
+      column.type === FieldType.ATTACHMENT &&
+      column.subtype !== oldTable?.schema[column.name]?.subtype
+  )) {
```
samwho, line 97: This is a little bit gnarly, might be a bit easier to read if it's split out into its own variable:

```typescript
const changedAttachmentColumns = Object.values(updatedTable.schema).filter(
  column =>
    column.type === FieldType.ATTACHMENT &&
    column.subtype !== oldTable?.schema[column.name]?.subtype
)
for (const attachmentColumn if changedAttachmentColumns) {
  // ...
```

samwho (approved):
Really nice type improvements 🚀

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -191,8 +196,10 @@
     // don't make field IDs for auto types
     if (type === AUTO_TYPE || autocolumn) {
       return type.toUpperCase()
-    } else {
+    } else if (type === FieldType.BB_REFERENCE) {
```
mike12345567, line 199: should we just check here if it has a `subtype` and do this? Just to cover all sub-type supporting columns?

adrinr: This was a 'hack' that was added for bb reference. This assumes that these columns' subtypes, and the old attachments have not

mike12345567: Ah I see - makes sense!

mike12345567 (commented):
Forgot to click send my approval, but looks good :)

PR #13318, opened 2024-03-20
Remove mssql mock, unify SQL-based query tests.

samwho:
## Description

Along the same lines as https://github.com/Budibase/budibase/pull/13316, this PR additionally created a unified set of SQL `Query` tests, giving new coverable for queries to mssql and mariadb.


packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -1,26 +1,43 @@
-import { Datasource, Query } from "@budibase/types"
```
mike12345567, line 1: very nice - cool to get these all into one test!

mike12345567 (approved):
LGTM

packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
samwho: Was a bit of a pain picking SQL that worked in all places 😄

Issue #10211, opened 2023-04-05
[BUDI-6846] Can't delete tables created by erroneous relationship creation


kpturner:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: monito

**Describe the bug**
Experimenting with relationships on a setup with an existing onboarded postgresql schema.  The new many-to-many relationships created two new tables in the schema.  The relationships were just experimental so they got deleted. However the two additional tables still exist and I cannot find a way to delete them.  

**To Reproduce**
1. From the data section, create a many to many relationship
2. A new table will be generated
3. Delete the relationship
4. How do I delete the new table?

**Expected behavior**
An intuitive way to remove the unwanted additional table

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome
 - Version 111

<sub>[BUDI-6846](https://linear.app/budibase/issue/BUDI-6846/cant-delete-tables-created-by-erroneous-relationship-creation)</sub>

PR #13319, opened 2024-03-20
Allow all external table deletion

mike12345567:
## Description
Allowing deletion of external tables, whether they were created in Budibase or not.

Previously if tables were fetching from a DB they could not be deleted, this restriction has been lifted as it has caused confusion and we already have enough warnings about what will be deleted.

image

Addresses:
https://linear.app/budibase/issue/BUDI-6846/allow-manual-db-deletion-from-inside-budibase
https://github.com/Budibase/budibase/issues/10211


shogunpurple (approved):
LGTM!

Issue #13320, opened 2024-03-21
Global Bindings broken :DataProvider Selection for Dropdown is not saved-updated when Form element is not inside DP


AlejandroFNadal:
**Checklist**
- [ X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Docker compose
  - Budibase Version: 2.22.7

**Describe the bug**
A form with a Dropdown that uses a Dataprovider as source will show the Dataprovider as a possible source but the selection of it will not work if the Form is not inside the dataprovider. The behavior from the previous version, Global Bindings, is kinda broken.

**To Reproduce**
1. Create an empty app with test data.
2. Create a data provider.
3. Create a form outside the data provider
4. Create a dropdown for the form.
5. Try to use the data provider as source for the form.

**Expected behavior**
I expect the dropdown to be able to use data from a dataprovider when the dropdown is not inside said data provider.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/32724827/63b162b8-548e-4417-aa8f-045a03e6e2a7)
The dataprovider appears but it is removed as soon as it is clicked.

**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 115.7.0esr

**Additional context**
I think this is very serious. We are lucky we make copies of the apps before updating them because otherwise we would have a completely broken system. We have used the behavior of Global Bindings a lot, because it makes apps so much faster (no need for DataProviders inside DataProviders)

AlejandroFNadal:
It is even worse than I imagined, this update is not applied to a app level but to the whole budibase. This means that I see these bugs even if I have not updated an specific app inside budibase. Is there a way to reinstall the previous version?

AlejandroFNadal:
For anyone having the same problem, if you are using Docker Compose:
Go to your docker compose file and add the tag to each budibase image that makes the docker. I went back to 2.17.6. For example:

```
image: budibase.docker.scarf.sh/budibase/apps:2.17.6
```
Don't do this for the minio image, the couchdb image and the watchtower.

Then I run budi hosting --stop and budi hosting --start
This reinstalled the images and now it works again, but all links to all data providers that were not in a hierarchy are now broken and another one was randomly selected. It is not an easy bug to fix from the perspective of budibase users (checking page by page manually, every DP).

Issue #13307, opened 2024-03-20
Can't scroll in conditions drawer even when its expanded


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.22.5
  - App Version: 2.22.5

**Describe the bug**
The conditions drawer doesn't have an overflow scroll on it. Therefore when it extends too far off screen you can't edit the conditions/add new ones to hide and show certain components when conditions are met.

**To Reproduce**
1. Create a new app
2. Add any component
3. Add an update setting condition to the component
4. Try and scroll the entire way to the right
5. See error

**Expected behavior**
Should be able to scroll and see the full condition regardless of screen size.

**Screenshots**
<img width="1063" alt="Screenshot 2024-03-20 at 05 28 54" src="https://github.com/Budibase/budibase/assets/126772285/3a8ba8a8-470b-46ed-9330-9c212bb27d83">

Issue #13212, opened 2024-03-07
Picker component that uses Dataprovider as its source seems to mutate the data. No way to stop this


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.21.3
  - App Version: 2.21.3

**Describe the bug**
When using a data provider as the source for a picker. The data won't respect the pre-sorted structure, it seems to get sorted again once it goes into the picker. I attempted to attach a view as the source of the data provider to then be used by the picker. Even though the sorting is already down in the view, the output of options isn't the same.

**To Reproduce**
1. Import the app export provided below
2. Go to the home screen. 
3. Click preview 
4. See that the options are sorted low to high even though there isn't any sorting checked. E.g. the alphabetical checkbox.

**Expected behavior**
You should be able to add data to a picker component without the component re-sorting it. 

**Screenshots**
View that has already sorted the data
<img width="532" alt="Screenshot 2024-03-07 at 12 39 41" src="https://github.com/Budibase/budibase/assets/126772285/514d68ae-542d-4ab6-88f5-63f80f00f3d2">

Repeater displaying the data correctly  
<img width="1046" alt="Screenshot 2024-03-07 at 12 40 24" src="https://github.com/Budibase/budibase/assets/126772285/caa68c77-d411-4eb7-98a9-9ea3e3608890">

Picker component displaying the data incorrectly
<img width="1374" alt="Screenshot 2024-03-07 at 12 41 17" src="https://github.com/Budibase/budibase/assets/126772285/34d7afd6-96c3-47e0-b141-a1f96e0ac33e">

**App Export**
[issue-export-1709815417747.tar.gz](https://github.com/Budibase/budibase/files/14524155/issue-export-1709815417747.tar.gz)

Issue #13200, opened 2024-03-05
Bug ? [ Builder ] - Custom  Components Refresh when a component is selected


poirazis:
While in the builder, hovering a component works nicely without causing a refresh, but when a component ( any component ) is selected a refresh happens. This only seems to happen to custom components, maybe a store update ? 

Can someone please double check to verify this is happening or is this my own code's fault ?

poirazis:
Hey @melohagan , can you confirm its BB causing the refreshes ?

ConorWebb96:
Hey @poirazis,

I'm not seeing this behaviour for other custom plugins. I am seeing it on your super table component each its clicked or another component is clicked it refreshes.

poirazis:
Hey @ConorWebb96 
it happens only with non-pimitive property types.

To replicate,
create an empty custom component
set a propery of type "datasource"
do a $: console.log( datasource )
you will see the component will refresh 

it wont happen with primitive props
we have double checked with Andrew Kingston

aptkingston:
Just an update on this - the issue is that the error boundary wrapper we wrap all custom components in spreads all props (`{...$$props}`) which breaks svelte's reactivity for any non-primitive types.

Worth nothing that this isn't really a bug as such, and an additional render/reactivity trigger is not typically a problem, but if you configure custom components to do things like API requests reactively based on prop changes then additional API calls will be made when selecting new components.

This won't have any effect in real apps - it's exclusively while working in the builder.

I'll see if there's a quick fix (will probably involved similar hacks to what we do in our core `Component.svelte` to alleviate this), but in the meantime I would encourage just ignoring this as a harmless side effect of working inside the preview.

poirazis:
Hey @aptkingston 
I have used some local checks in the component to not refetch using lodash.isEqual to deep compare data related props circumvent unceseccasy API calls. Maybe its a way other component devs can take.

aptkingston:
Just adding this here for reference, but as @poirazis and I discussed offline - I've added the `memo` and `derivedMemo` utils into the client SDK for plugin authors to use. These utils help ensure reactivity works as expected.

PR #13327, opened 2024-03-21
Various fixes and improvements

aptkingston:
## Description
This is a collection of various fixes, improvements and QoL upgrades.

## Changelog
- Remove "Grid" layout component, which should not have been visible
- Fix recently introduced bug to the `click_outside` helper which allowed non-function handlers to be registered, resulting in runtime errors
- Improve client lib performance by removing a large amount of wasted renders and reactivity triggers
- Improve performance of multiple blocks by improving reactive statement specificity in places where we were poorly observing an entire store
- Massively improve performance of hovering and selecting components. Using the preview to hover and select components feels nearly instant now, fixing a long standing "slowness" (opinion, but I felt it was always laggy for some reason)
- Fix stale state causing flashes of incorrect text or colours of selection indicators
- Add `memo` and `derivedMemo` utils from frontend-core into the SDK for plugin authors https://github.com/Budibase/budibase/issues/13200 (not a fix for the issue, but these are tools to help work around it)
- Allow horizontal overflow in drawers #13307 https://linear.app/budibase/issue/BUDI-8110/cant-scroll-in-conditions-drawer-even-when-its-expanded
- Fix data provider validation not accounting for global bindings #13320 #13065 https://linear.app/budibase/issue/BUDI-8112/global-bindings-broken-dataprovider-selection-for-dropdown-is-not https://linear.app/budibase/issue/BUDI-8021/picker-components-with-options-source-allow-you-to-select-non-nested
- Fix options ordering not being respected when using a data provider options source #13212 https://linear.app/budibase/issue/BUDI-8071/picker-component-that-uses-dataprovider-as-its-source-seems-to-mutate
- Fix snippets not working inside column value transformers
- Add live eval context to table block and table component for the "value" column binding
- Fix code mirror autocomplete popover showing "NO NAME" for certain bindings


poirazis:
Thanks for considering plugin authors !

deanhannigan (approved):
LGTM! 👏

Issue #13329, opened 2024-03-21
"Clear relationships if empty" flag doesn't work in Update Row automation step


andz-bb:
**Checklist**
- [x] I have searched Budibase discussions and GitHub issues to check if my issue already exists

**Hosting**
- Self
  - Method: dev stack
  - Budibase Version: 2.22.5
  - App Version: 2.22.5

**Describe the bug**
![image](https://github.com/Budibase/budibase/assets/110921612/5ea03e13-0ec7-4945-b0f5-f7c49edc82fb)
The Clear relationships if empty checkbox is broken. You would expect it to remove any relationships for a given field if no value is supplied, but the relationships persist. As a workaround, if you provide a string value it does successfully remove relationships.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new app with the sample datasource
2. Create an automation that gets employee rows and then updates them
3. Configure the Update Row action to _Clear relationships if empty_ for the Jobs field, leaving the value blank
4. Test the automation and see that the relationships are not cleared

**Expected behaviour**

- When _Clear relationships if empty_ is selected and the binding value doesn't contain an object or array of objects, any existing bindings should be cleared.
- When _Clear relationships if empty_ is not selected and no value is provided, existing relationships should be preserved.

  
**App Export**
This one just has the automation described as above pre-made for the sample DB:
[Clear Relationships Export.tar.gz](https://github.com/Budibase/budibase/files/14695833/Clear.Relationships.Export.tar.gz)

**Desktop (please complete the following information):**
 - OS: macOS 14
 - Browser: Arc
 - Version: 1.34.0 (Chromium 122)

**Additional context**
Raised by a customer

PR #13332, opened 2024-03-21
fix for boolean values getting stripped when running processObject

andz-bb:
## Description
I found this when investigating the clearRelationships checkbox not working in the Update Row automation step. 

## Addresses

#13329
[BUDI-8118](https://linear.app/budibase/issue/BUDI-8118/clear-relationships-if-empty-flag-doesnt-work-in-update-row-automation)

## Launchcontrol

Fixes certain values getting stripped when processing object in string templates. The update fixes a problem clearing relationships on the update row automation step, and probably other stuff elsewhere.

### DIFF
--- a/packages/string-templates/src/index.ts
+++ b/packages/string-templates/src/index.ts
@@ -94,7 +94,7 @@ export async function processObject<T extends Record<string, any>>(
   for (const key of Object.keys(object || {})) {
     if (object[key] != null) {
       const val = object[key]
-      let parsedValue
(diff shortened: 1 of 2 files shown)


adrinr (approved):

aptkingston (approved):
LGTM! Nice catch, I'm surprised this hasn't caused more issues.

adrinr:
Good finding, I added a test for that! 🙏🏽

PR #13336, opened 2024-03-22
Allow SQL queries to pass null for empty bindings rather than empty string for new queries

mike12345567:
## Description
Adding a new property to queries `nullDefaultSupport` - which allows queries to use null instead of empty string when a binding evaluates to nothing. For most queries this has little impact, as the bindings are going into a query block, which is a string, but for SQL our prepared statements expect null/undefined for empty values.

This means that nullable columns should work better - but we didn't want to change the behaviour for existing queries as it would be a breaking behaviour - we can set this parameter for all queries if desired, but this will mean that new queries will not error when no value is provided for an SQL binding.

I've added two test cases, confirming a scenario that should error on old queries, and doesn't error for new queries.

Also a variety of type updates, as passing these parameters around was quite complicated.

Addresses: https://github.com/Budibase/budibase/issues/11629


packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -12,31 +12,52 @@ const createTableSQL: Record<string, string> = {
...
+
+const MSSQL_SPECIFICS = {
+  // MS-SQL doesn't throw an error here, it casts empty string to nothing
+  nullError: undefined,
+}
+
+const MARIADB_SPECIFICS = {
+  nullError:
+    "Incorrect integer value: '' for column `mysql`.`test_table`.`number` at row 1",
+}
```
samwho, line 53: I'm not sure this is worth it. It's only used in one test and we could reasonably assert that there is an error, and not worry about the specifics of the message.

The purpose behind these tests was to keep them to an absolute minimum variability between databases. If you need to test behaviour for a specific database, it can be a new test suite specific to that database.

packages/server/src/api/controllers/query/index.ts
```diff
@@ -100,10 +99,18 @@ export async function save(ctx: UserCtx<Query, Query>) {
   const datasource = await sdk.datasources.get(query.datasourceId)
 
   let eventFn
-  if (!query._id) {
+  if (!query._id && !query._rev) {
```
adrinr, line 102: Will we ever have rev but not I'd?

mike12345567: if we have an ID it is likely we will have a rev, but just for full confirmation making sure we have both.

packages/server/src/sdk/app/queries/queries.ts
```diff
@@ -65,11 +65,27 @@ export async function fetch(opts: { enrich: boolean } = { enrich: true }) {
   return updateSchemas(queries)
 }
 
+export async function enrichArrayContext(
+  fields: any[],
```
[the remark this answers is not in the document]
mike12345567: Technically what this does is allow the enrichment to run over an array, so `any[]` is correct - I will switch out the object call as it is a little confusing.

samwho (approved):

PR #13341, opened 2024-03-22
Remove `pg` mock.

samwho:
## Description

Saved the worst til last. This gets rid of all mocking of `pg` in `packages/server`, and has involved a significant amount of work to do so.

In the process of this I have:

- Gotten rid of `query.seq.spec.ts` by distributing its tests into more relevant files (creating `rest.spec.ts` and `permissions.spec.ts` in the process, though most of the tests ended up in `generic-sql.spec.ts`).
- Made `executeQuery.spec.ts` run against real databases.
- Introduced `nock` to `packages/server` for `rest.spec.ts`, instead of relying on `__mocks__/node-fetch.ts`, which I intend to remove in future.

The delta on this PR is scary but more than half of it is moving things around.


packages/server/src/sdk/tests/attachments.spec.ts
```diff
@@ -1,4 +1,12 @@
-import newid from "../../db/newid"
+import TestConfig from "../../tests/utilities/TestConfiguration"
```
samwho, line 1: This test is a bit unrelated but was failing for me so I fixed it. It was hardcoding table names in a way that didn't play nice if the test was re-run with `REUSE_CONTAINERS=true` set. I've modernised it and made it play nice.

packages/server/src/api/routes/tests/queries/permissions.spec.ts
```diff
@@ -0,0 +1,47 @@
+import * as setup from "../utilities"
+import { checkBuilderEndpoint } from "../utilities/TestFunctions"
+import TestConfiguration from "../../../../tests/utilities/TestConfiguration"
+import { Datasource, Query, SourceName } from "@budibase/types"
+
+describe("query permissions", () => {
```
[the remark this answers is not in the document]
mike12345567: The tests for this are a bit all over the place - generally we test it as part of the individual test - but its not a great strategy. 

We could probably take a more generic approach to this - could be worth creating an issue for this.

packages/server/src/api/routes/tests/queries/mongodb.spec.ts
```diff
@@ -88,345 +88,491 @@ describe("/queries", () => {
...
-        },
-        extra: {
-          actionType: "findOneAndUpdate",
-        },
-      },
+      expect(result.data).toEqual([{ value: 6 }])
     })
 
-    const result = await config.api.query.execute(query._id!)
+    it("a find query", async () => {
```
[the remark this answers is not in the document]
mike12345567: This is an interesting one - there isn't really any rules today about what we put in the `it` syntax - i've just always followed a rule of making it read like `it should ....` or something along those lines - but we've never really discussed any guidelines to follow as a team, what was the issue with the `should` in this case?

packages/server/jest.config.ts
```diff
@@ -60,6 +54,12 @@ const config: Config.InitialOptions = {
     "!src/db/views/staticViews.*",
     "!src/**/*.spec.{js,ts}",
     "!src/tests/**/*.{js,ts}",
+    // The use of coverage in the JS runner bundles breaks tests
+    "!src/jsRunner/bundles/**/*.{js,ts}",
```
mike12345567, line 58: I do think we should just set the whole `jsRunner` directory to be excluded, within this directory there is a lot of danger when it comes to the code coverage, and excluding it I feel is pretty understandable.

packages/server/package.json
```diff
@@ -143,7 +143,7 @@
     "jest": "29.7.0",
     "jest-openapi": "0.14.2",
     "jest-runner": "29.7.0",
-    "jest-serial-runner": "1.2.1",
+    "nock": "^13.5.4",
```
mike12345567, line 146: Generally we have always pinned package versions - the `yarn.lock` does sort this for us anyway, but for consistency probably worth removing the `^`.

mike12345567 (approved):
Very nice improvement to the test suite - this has come with some significant improvements in some areas as well, like real test cases for the automation query action - some really far ranging improvements to the test suite!

packages/server/package.json
samwho: Ahh I ran `yarn add --dev nock` and this was what it put in by default. Will change.

mike12345567: > Ahh I ran `yarn add --dev nock` and this was what it put in by default. Will change.

Yeap by default this is what happens, sadly theres no real way to change the default `add` functionality, we just tend to pin!

packages/server/src/sdk/tests/attachments.spec.ts
samwho: It was, yeah. Worked fine on a fresh DB, but running a 2nd time (either through `REUSE_CONTAINERS` or `--watch`) it would fail.

packages/server/src/api/routes/tests/queries/mongodb.spec.ts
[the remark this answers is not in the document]
samwho: It was silly, the `describe` block is `describe("execute", ...` and the `should` didn't read nicely in the output. I'm honestly not tied to it at all, and don't mind putting the shoulds back. It's an incredibly unimportant change.

PR #13345, opened 2024-03-25
[Changed] Identity override for event publish + BPM literals

jvcalderon:
## Description

- **Allow identity override for event publish:**
During event publish process, it tries to backfilling (if needed) getting information from the user cache, but sometimes event sending is handled in background. Overrides will be normally performed in automatic actions or operations in async flows (BPM) where the user session is not available.

- **Provides correlation keys, instances keys, values, etc required in onboarding flow BPM process.**

- **Password is optional for tenant**
SSO accounts don't have password.

Note: Some failing `qa-core` tests have been removed. Suite is in process to be ripped out and functionality is covered with tests in server. See: https://linear.app/budibase/issue/BUDI-8106/remove-qa-core


shogunpurple (approved):
LGTM

PR #13346, opened 2024-03-25
Fix to ensure runtimeTime binding is used is automation js modal

deanhannigan:
## Description
Ensures that when adding step bindings into the JS automation modal, the `runtimeBinding` is the value that is used.

Previously the `runtime` and `readable` bindings were identical in automation steps. With customisable step names this is now no longer the case. Adding a binding from a step with a custom name now inserts the custom step name which will cause the step to fail.

## Addresses
-  https://linear.app/budibase/issue/BUDI-8124/custom-automation-step-names-break-js-step-binding-values

## Launchcontrol
Fix to ensure the `runtimeBinding` is added to the JS automation modal and not the `readableBinding`


aptkingston (approved):
LGTM!

PR #13348, opened 2024-03-25
"Download file" app action

adrinr:
## Description
Adding a new app action to allow downloading files. These can be either attachments or URLs.

### Adding new action
image

#### Selecting type 'attachment'
image

#### Binding
*Attachment/single attachment fields are selectable as a binding
image

#### Binding
* All the binding settings are available
image
image

### Client side
1. If the column is either single attachment column or a multiple attachment with a single item, it downloads it as it is
2. If the column is a multiattachment with multiple files, it zips the files and returns downloads the file as `rowDisplayName_columnName.zip`
3. If there it no attachment (empty data), it returns 404 and displays error
4. Wrong URL will display errors

https://github.com/Budibase/budibase/assets/15987277/6c95924a-0c58-4a22-a13e-740301e1b38a

Configurations:
| Header | Header |
|--------|--------|
| Single attachment | image |
| Multiattachment column with a single file | image |
| Download attachment by URL binding | image |
| Multiple attachments | image |
| Empty attachment | image | 
| Chaining actions | image |
| Download via external URL | image |
| Wrong URL | image |

## App Export
[download file-export-1712153141835.tar.gz](https://github.com/Budibase/budibase/files/14853397/download.file-export-1712153141835.tar.gz)

## Launchcontrol
Add "download file" app action

## Feature branch env
[Feature Branch Link](http://fb-budi-8082-download-files-using-a-button-action.fb.qa.budibase.net)


.vscode/launch.json
```diff
@@ -20,6 +19,12 @@
       "runtimeArgs": ["--nolazy", "-r", "ts-node/register/transpile-only"],
       "args": ["${workspaceFolder}/packages/worker/src/index.ts"],
       "cwd": "${workspaceFolder}/packages/worker"
+    },
+    {
+      "type": "chrome",
```
adrinr, line 24: This can be used for VScode to attach the svelte client

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
```diff
@@ -0,0 +1,76 @@
...
+      title="URL"
+      {bindings}
+      value={parameters.url}
+      on:change={value => (parameters.url = value.detail)}
+    />
+    <Label small>File name</Label>
+    <DrawerBindableInput
+      title="File name"
+      {bindings}
+      value={parameters.file_name}
```
aptkingston, line 60: NAB: you don't need to change this, and we never enforce it, but by convention we always just use camel case for variable naming in the frontend. Just an FYI for the future!

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
```diff
@@ -0,0 +1,76 @@
...
+    b => b.fieldSchema?.type === FieldType.ATTACHMENT
+  )
+
+  const fileOptions = [
+    {
+      label: "Attachment",
+      value: "attachment",
+    },
+    {
+      label: "Url",
```
aptkingston, line 20: NAB: `URL` (uppercase) for the label

aptkingston (commented):
Nice work! Just one suggested change about how we could allow users to select attachment fields 👍

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
adrinr: I will check if we can add a lint rule without breaking the existing usages :)

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
```diff
@@ -0,0 +1,76 @@
...
+  />
+  {#if parameters.type === "attachment"}
+    <Label small>Attachment</Label>
+    <DrawerBindableInput
+      title="Attachment"
+      bindings={fileBindings}
+      allowHelpers={false}
+      value={parameters.attachment}
+      on:change={value => (parameters.attachment = value.detail)}
+    />
```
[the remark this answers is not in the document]
adrinr: How will this work, for multiple attachment columns?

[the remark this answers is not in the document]
mike12345567: I guess how we could make it work is if you pick an attachment column with multiple then we just download them all. This would work well for old attachment columns with just a single attachment.

Might be a bit weird though if you have like 10 different files attached, we're going to trigger 10 downloads - but I think this makes some sense and if you just want to download one of the files then you use the URL + name to pick a specific one.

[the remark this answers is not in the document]
aptkingston: @mike12345567 yep I agree - this is what @adrinr and I discussed offline and decided on yesterday as well, so great minds think alike! 👌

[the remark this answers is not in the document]
adrinr: Done! If there is more than 1 file in the column, it gets zipped!

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
adrinr: Done!

adrinr:
I will write some tests once merged, as it will require some refactor around the controller usage

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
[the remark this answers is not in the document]
mike12345567: Just for clarity - this has ended up looking like:
image

The zip name is based on the display column of the row.

packages/frontend-core/src/api/attachments.js
```diff
@@ -88,5 +88,19 @@ export const buildAttachmentEndpoints = API => {
         },
       })
     },
+
+    /**
+     * Download an attachment from a row given its column name.
+     * @param tableId
```
mike12345567, line 94: Probably should fill out descriptions for these to be inline with the other functions here

packages/frontend-core/src/utils/download.js
```diff
@@ -34,3 +34,35 @@ export async function downloadStream(streamResponse) {
...
+    },
+    body: JSON.stringify(body),
+  })
+
+  if (!response.ok) {
+    return false
+  } else {
+    const contentDisposition = response.headers.get("Content-Disposition")
+
+    const matches = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/.exec(
```
mike12345567, line 52: Can we break out this regex into a named constant to signify what it is supposed to do?

I'm guessing this is trying to find the file name from the content disposition, but its not 100% obvious.

packages/server/package.json
```diff
@@ -63,6 +63,7 @@
     "@socket.io/redis-adapter": "^8.2.1",
     "airtable": "0.10.1",
     "arangojs": "7.2.0",
+    "archiver": "^7.0.1",
```
mike12345567, line 66: Should we pin this library version? I know there are some that aren't like the `redis-adapter` above - but we probably should keep pinning things for clarity/safety.

packages/server/src/api/controllers/row/index.ts
```diff
@@ -251,3 +257,59 @@ export const exportRows = async (
...
+
+  if (attachments.length === 1) {
+    const attachment = attachments[0]
+    ctx.attachment(attachment.name)
+    ctx.body = await objectStore.getReadStream(
+      objectStore.ObjectStoreBuckets.APPS,
+      attachment.key
+    )
+  } else {
+    const passThrough = new stream.PassThrough()
```
mike12345567, line 297: Very nice implementation!

packages/string-templates/src/index.ts
```diff
@@ -12,11 +12,12 @@ import {
 } from "./utilities"
 import { convertHBSBlock } from "./conversion"
 import { setJSRunner, removeJSRunner } from "./helpers/javascript"
-import { helpersToRemoveForJs } from "./helpers/list"
```
mike12345567, line 15: What are the changes here needed for out of interest?

mike12345567 (approved):
LGTM! Few comments, none of them critical, had a play with this myself locally, tried downloading a few different things, works really well!

One note-worthy thing I experienced was if you try to download anything that would load HTML - the error is simple "File cannot be downloaded" - for example if you try to put a google drive shared link into the button (I didn't expect this to work, but just wanted to see what would happen).

We might want to add a clearer message in these scenarios, it seems generally what is happening is a CORS issue which gets dumped to the console, we could likely capture this instances and instead state that the URL supplied does not link to a file.

adrinr:
This needs to wait until https://github.com/Budibase/budibase/pull/13399 is merged

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
```diff
@@ -0,0 +1,110 @@
...
+  />
+  {#if parameters.type === "attachment"}
+    <Label>Table</Label>
+    <Select
+      placeholder={null}
+      bind:value={parameters.tableId}
+      {options}
+      getOptionLabel={table => table.label}
+      getOptionValue={table => table.resourceId}
+    />
```
aptkingston, line 64: Rather than have 3 components here that all need configured, we could simply filter the binding list down to only inlude attachment columns. This is what I was originally thinking.

If you do
```js
bindings.filter(x => x.fieldSchema?.type === "attachment")
```
Then you can just have a single picker, with the values being the `runtimeBinding`s of those bindings. Then it's much easier for the user to configure 👍

mike12345567: We can't do this sadly - due to the way zipping has to work. If the user passes in a bunch of different attachments we don't know what table/row the attachments came from. We played around with different ways to try and convert the attachment list back to the information we actually need, but it was super problematic, often requiring extra metadata on the rows.

The only other way would be to download all the files somehow into memory in the browser and then zip it there, but this proved way more complex than allowing the backend to build a zip out of all the attachments.

The other option was to just download all of the files individually, which we felt was a worse experience than it being zipped when there are multiple.

aptkingston (approved):
Works well - nice work @adrinr 👌 Just had my one comment about how the settings are configured, but if it's not possible to work around that then I'm happy to merge as is.

packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DownloadFile.svelte
mike12345567: We tried the egress enrichment - it can sort of work but the use of it is very odd - as we want to download a single row's attachment column, so we have to do something like `attachments[0].tableId` etc - it assumes that if a binding array is passed in we just download whatever is in the first location.

Issue with this is an edge case, but one that cannot be fixed, if a user manages to get attachments from two columns together here then they'll only get one. This in combination with how odd the code looked and how awful the enrichment was going to be, it felt better to just get the information we need from the user directly.

aptkingston: Cool, no problem.

mike12345567: Just to finalise this as well, we did consider another method of building an endpoint that takes a list of attachment URLs and just downloads them, however its very difficult to secure an endpoint like this, it would be very difficult for us to know if the user actually has access to the row which the attachment came from, without having to search by the attachment URL and then working out if the user can access all of the attachments - again rather than going down this route we preferred just getting the info from the user.

packages/string-templates/src/index.ts
adrinr: We were importing just to export, renaming it to `_whatever`. This is not actually needed, but it slipped while refactoring `string-templates` to esm

PR #13350, opened 2024-03-25
Remove jest testcontainers

samwho:
## Description

While working on the Postgres unmocking PR, I noticed that when we run tests, there are two copies of the `testcontainers/ryuk` container running. One as version `0.3.0` and another as `0.5.1`.

On closer inspection, it appears that the `jest-testcontainers` package is depending on `testcontainers@4.7.0` while the Budibase project depends on `testcontainers@10.7.2`.

This discrepancy doesn't seem ideal, so I've removed `jest-testcontainers` in the hope it may help with some of the flakiness we've been seeing. I have no proof that it will, but this at least removes this discrepancy as a culprit.

Pro PR: https://github.com/Budibase/budibase-pro/pull/285


packages/backend-core/src/db/Replication.ts
```diff
@@ -27,7 +27,7 @@ class Replication {
           return resolve(info)
         })
         .on("error", function (err) {
-          throw new Error(`Replication Error: ${err}`)
+          throw err
```
samwho, line 30: This caused an error in a test while I was working on this change, and I moved to throwing the raw error in order to see the full stack trace. Decided to leave it as is because I feel this is more useful.

packages/backend-core/tests/core/utilities/testContainerUtils.ts
```diff
@@ -1,85 +1,66 @@
+import { DatabaseImpl } from "../../../src/db"
```
samwho, line 1: I've changed how this file works because when I moved away from `jest-testcontainers`, this code started crafting invalid CouchDB URLs.

The new code uses `docker ps --format json` to make parsing the output easier, and it also makes sure to always return `testcontainers` containers by looking for the `org.testcontainers=true` label.

packages/server/src/tests/test-couchdb.ini
```diff
@@ -0,0 +1,2 @@
+[log]
```
samwho, line 1: I've taken the opportunity here to calm the CouchDB logs down. By default they log every single HTTP request, which is an unbelievable amount of logging. This made `DEBUG=testcontainers*` unusable, whereas with `level = warn` the output is a little less huge and more useful.

packages/backend-core/src/tests/globalTeardown.ts
```diff
@@ -0,0 +1,12 @@
+import { getContainerRuntimeClient } from "testcontainers"
+
+export default async function teardown() {
+  const client = await getContainerRuntimeClient()
+
+  // @ts-expect-error
+  // eslint-disable-next-line no-undef
+  const containerId = globalThis.__COUCHDB_CONTAINER_ID__
+  const container = client.container.getById(containerId)
+  await client.container.stop(container)
```
adrinr, line 10: Not sure if there is any real benefit of doing this. The containers will stop automatically, and this might add some await time to the tests for no reason

packages/server/src/tests/globalSetup.ts
```diff
@@ -0,0 +1,26 @@
+import { GenericContainer, Wait } from "testcontainers"
+
+export default async function setup() {
+  const container = await new GenericContainer("budibase/couchdb")
+    .withExposedPorts(5984)
+    .withEnvironment({
+      COUCHDB_PASSWORD: "budibase",
```
adrinr, line 7: Should we use the `process.env` to always align with the `jestEnv.ts`?

packages/server/src/tests/globalSetup.ts
```diff
@@ -0,0 +1,26 @@
+import { GenericContainer, Wait } from "testcontainers"
+
+export default async function setup() {
```
adrinr, line 3: Is there any difference with the setup/teardown in backend-core? Should we reuse it?

packages/backend-core/src/tests/globalTeardown.ts
samwho: Ah, do they? This is just for the CouchDB container, I didn't realise there was some existing teardown process.

packages/server/src/tests/globalSetup.ts
samwho: I believe this runs _before_ `jestEnv.ts` runs, so I'm not sure `process.env` will contain values for this yet. I can try tomorrow, though. It would be nice to have a single source of truth for these.

packages/server/src/tests/globalSetup.ts
samwho: Nah, they're identical. I'll look in to a way to unify them.

packages/backend-core/tests/core/utilities/testContainerUtils.ts
mike12345567: This is a nice improvement, I didn't even know you could get JSON out here, very handy!

mike12345567 (approved):
This is a nice improvement, using testcontainers more directly gives us a lot more control and removes a potential candidate for the flakiness we've been seeing, great work!

packages/server/src/tests/globalSetup.ts
samwho: Yeah, sadly `jestEnv.ts` runs after `globalSetup.ts` so there's no `COUCHDB_*` variables in there yet. Going to leave this as is, unless you can think of a way to do this.

adrinr (approved):
Nice to see some much love given to the tests :D

packages/backend-core/src/tests/globalTeardown.ts
adrinr: AFAIK test containers check if they are attached to a process, otherwise, they kill themselves

packages/server/src/tests/globalSetup.ts
mike12345567: I'm trying to check this, but the parameter names in the jest config would suggest that the env comes before the setup files:
image

PR #13351, opened 2024-03-25
Handle files in SMTP automation block

PClmnt:
## Description
This adds the ability for the SMTP block in automations to handle and send files.

- A new field is added to the SMTP block called `Attachments`
- You can add either an array of URLs or a comma separated list.
- URLs can be fully qualified or a partial url, i.e an attachment column returning `files/signed/...`
- Adds tests to handle this

- also did some work as part of this PR to fix the minio test container and moved it into backend-core for use in both server / worker  
## Addresses
https://linear.app/budibase/issue/BUDI-8080/email-automation-action-can-attach-files
## App Export

### Example:
image

Attachment available in Ethereal:
image


packages/worker/src/api/routes/global/tests/realEmail.spec.ts
```diff
@@ -81,4 +93,19 @@ describe("/api/global/email", () => {
...
+    let filename = "test.txt"
+    await objectStore.upload({
+      bucket,
+      filename,
+      body: Buffer.from("test data"),
+    })
+    tk.freeze(mocks.date.MOCK_DATE)
+    let presignedUrl = await objectStore.getPresignedUrl(bucket, filename, 600)
+
+    await sendRealEmail(EmailTemplatePurpose.WELCOME, [presignedUrl])
```
samwho, line 109: As far as I can tell, nothing is asserting that the email that's sent contains the attachment we set. Could we create an assertion for that?

packages/server/src/automations/steps/sendSmtpEmail.ts
```diff
@@ -48,6 +48,10 @@ export const definition: AutomationStepSchema = {
           type: AutomationIOType.STRING,
           title: "HTML Contents",
         },
+        attachments: {
+          type: AutomationIOType.STRING,
```
mike12345567, line 52: Wondering if we should make this a new type, `URLS` - we need this to be a list of URLs to files that should be attached, might be useful from a validation point of view in future.

Also - how are the attachments named? Wondering if we actually need an `AutomationIOType.ATTACHMENT` - which is a specific type containing an array of objects `{ url: ..., name: ... }` as right now, I don't see how the attachment will be named against the email, this should probably be user configurable.

packages/worker/src/api/routes/global/tests/realEmail.spec.ts
PClmnt: Hm, had a look at this and I'm not sure the best way to go about it. Nodemailer doesn't return any info about attachments after sending, and we use ethereal to send a real email so I don't want to spy on and mock the function either.

packages/worker/src/utilities/email.ts
```diff
@@ -162,6 +164,47 @@ export async function sendEmail(
...
+        const isFullyFormedUrl =
+          attachment.url.startsWith("http://") ||
+          attachment.url.startsWith("https://")
+
+        const url = isFullyFormedUrl ? attachment.url : baseUrl + attachment.url
+        let headers
+
+        if (env.isTest()) {
+          headers = {
+            Host: "minio-service",
```
PClmnt, line 180: Not very nice, and kind of hacky, but seems to be the only way to get around the minio host requirement.

samwho: Mm, I'm not the world's biggest fan of test specific code mixed in with production code. I won't block on it, but is this our only option?

packages/worker/src/utilities/email.ts
```diff
@@ -162,6 +164,47 @@ export async function sendEmail(
       contents: opts?.contents,
     }),
   }
+  if (opts?.attachments) {
+    const baseUrl = appContext.getPlatformURL()
+    const attachments = await Promise.all(
+      opts.attachments?.map(async attachment => {
+        const isFullyFormedUrl =
```
samwho, line 171: nit: typically we'd refer to URLs with a scheme and a domain as "absolute" and paths as "relative" but it's not really a problem, don't feel the need to change anything.

packages/worker/src/tests/structures/configs.ts
```diff
@@ -62,8 +62,8 @@ export function smtpEthereal(): SMTPConfig {
       from: "testfrom@example.com",
       secure: false,
       auth: {
-        user: "wyatt.zulauf29@ethereal.email",
-        pass: "tEwDtHBWWxusVWAPfa",
+        user: "mortimer.leuschke@ethereal.email",
```
samwho, line 65: What's the story with this change?

packages/server/src/integrations/utils.ts
```diff
@@ -509,7 +509,7 @@ export async function handleFileResponse(
       type: response.headers["content-type"],
     })
   }
-  presignedUrl = await objectStore.getPresignedUrl(bucket, key, 600)
+  presignedUrl = await objectStore.getPresignedUrl(bucket, key, 60000)
```
samwho, line 512: Was this always incorrect, or are we actually increasing the length of the presigned URL lifetime by 1000x?

samwho (approved):
Broadly happy for this to go in as is, just left a scattering of questions :)

packages/server/src/integrations/utils.ts
mike12345567: I believe this is an addition as part of the `feat/automations-and-rest-file-handling` branch rather than something on master - however I think it would probably be good to update the `getPresignedUrl` function to take the duration as an object specifier like `{ validForSeconds: ... }` to make it more obvious what is being specified. Right now its going from 10 minutes to 0.694 days, which does seem a little odd as well - the `Duration` utility from `backend-core` could be helpful here.

packages/worker/src/utilities/email.ts
mike12345567: Me and Peter took a look into this last week - we tried to find a way to remove the host setting when running in tests, but sadly its unavoidable. In production we'll get away with this as the request will go through the proxy, rather than directly to the service.

There were two other ways this could be solved:
1. Run the budibase/proxy container as well in our test suites to connect through, this would be the most correct to our production environments, but also a little un-necessarily complex.
2. Mocking the request in someway as part of the test, to make sure that the header is attached as expected as part of the test.

mike12345567 (approved):
Looks good to me overall - I would probably take a look at the `getPresignedUrl` function to make it a bit easier to understand/follow and confirm why `60000` seconds is required as part of this update.

packages/server/src/integrations/utils.ts
PClmnt: Yea so I was running into issues with the signed url and just added this to test something, forgot to move it back - I'll switch it to use the Duration utility also

packages/worker/src/tests/structures/configs.ts
PClmnt: The other ethereal email / pass we were using for testing just wasn't working anymore - the account didn't exist so I created a new one for the test config

packages/backend-core/tests/core/utilities/minio.ts
```diff
@@ -21,7 +22,8 @@ export async function start(): Promise<StartedTestContainer> {
     .withWaitStrategy(new ObjectStoreWaitStrategy().withStartupTimeout(30000))
     .start()
 
-  return container
+  const port = container.getMappedPort(9000)
+  env._set("MINIO_URL", `http://0.0.0.0:${port}`)
```
adrinr, line 26: This ip should come from the container. `container.host()` or something like this

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -436,6 +447,33 @@
...
+                    onChange(
+                      {
+                        detail: e.detail.map(({ name, value }) => ({
+                          url: name,
+                          filename: value,
+                        })),
+                      },
+                      key
+                    )}
+                  object={handleAttachmentParams(inputData[key] || {})}
```
adrinr, line 467: It's usually cleaner if this `|| {}` check is part of the function itself, not responsibility of the caller

packages/server/src/automations/steps/sendSmtpEmail.ts
```diff
@@ -72,10 +72,9 @@ export const definition: AutomationStepSchema = {
           title: "Location",
           dependsOn: "addInvite",
         },
-        url: {
-          type: AutomationIOType.STRING,
-          title: "URL",
-          dependsOn: "addInvite",
+        attachments: {
+          type: AutomationIOType.ATTACHMENT,
```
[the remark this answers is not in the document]
adrinr: Single attachment is not yet there, probably I should prioritise it in order to all work to consider this type

packages/types/src/documents/app/automation.ts
```diff
@@ -10,6 +10,7 @@ export enum AutomationIOType {
   ARRAY = "array",
   JSON = "json",
   DATE = "date",
+  ATTACHMENT = "attachment",
```
adrinr, line 13: Shouldn't these strings be coming from the actual types?

packages/backend-core/tests/core/utilities/minio.ts
samwho: I'm actually in the process is explicitly _not_ getting this information from the container because I've found it can be incorrect. Some Docker implementations will bind a port on both IPv4 (127.0.0.1) and IPv6 (::1) localhost, but `container.getHost()` always gives you `localhost`. This can then resolve to either a v4 or v6 address. Then `container.getExposedPort(1234)` seems to usually give back the v6 address.

I have a PR I'm working on that always uses v4 for both things, and bypasses the `testcontainers` functions entirely.

packages/types/src/documents/app/automation.ts
mike12345567: These types are automation input types - they've always been independent from the field types that our tables/columns use.

The types are similar, but not identical, for example automations have an `object` type - which can have custom types like row or table.

They are relatively independent from each other, as all these types are used for is informing the UI what form components to display for the various inputs/outputs of automation steps - they don't really tie up to Budibase data at all.

packages/backend-core/tests/core/utilities/minio.ts
PClmnt: Yep, as Sam mentioned i did try to use the getHost() function but found it just didn't work.

packages/server/src/integrations/utils.ts
PClmnt: So for this i just removed the param altogether and am using the default param on the getPresignedUrl function, which is an hour.

packages/server/src/automations/steps/sendSmtpEmail.ts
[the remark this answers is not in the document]
mike12345567: Yeah - I'm wondering if the type is still needed to tell the automation UI that it needs to support multiple attachments rather than just a single one (key value builder vs a single key value)

PR #13355, opened 2024-03-26
Fix issues with colours in app skeletons

aptkingston:
## Description
This PR fixes an issue where certain colours are not correct in the app skeleton. This is because we weren't properly setting the required spectrum classes. This only affect our custom themes - Nord and Midnight.

Before:
image

After:
image

## Launchcontrol
Fix app skeleton colours for Nord and Midnight themes.


Ghrehh (approved):
LGTM, my b for missing this

aptkingston:
> LGTM, my b for missing this

You had it working, it was me who old told you to remove the extra CSS vars which revealed the real issue!

aptkingston:
Seems to be failing due to some account portal issue. I don't even want to bump the account portal of course, but I'm forced to do so, so now this is forced to be broken until that issue is resolved.

PR #13356, opened 2024-03-26
Single attachment column type

adrinr:
## Description
Adding a new column type for single attachments. This will be stored as the same shape as the current attachments, but as a single object (instead of an array). Having this new type will allow having columns with a single attachment, simplifying validations and bindings.

#### Adding a new single attachment column
image

##### Editing an empty cell
image

##### Editing a none empty cell (will not allow adding more than one file)
image

##### Required single attachment validation
image

#### Inline editing grid
This will work like on the builder end
image

##### New row
image

It will not allow adding more than one image for single attachments
image

##### Editing existing row
image

#### Form block
It will act as the side panel
##### New row
image

##### Editing row
image

#### Multistep block
image

#### Repeater with images
image

## Launchcontrol
Adding a new column type single attachment

## Feature branch env
[Feature Branch Link](http://fb-budi-8122-single-attachment-column-type.fb.qa.budibase.net)


packages/bbui/src/Form/Core/Dropzone.svelte
```diff
@@ -67,7 +67,7 @@
   }
 
   $: showDropzone =
-    (!maximum || (maximum && value?.length < maximum)) && !disabled
+    (!maximum || (maximum && (value?.length || 0) < maximum)) && !disabled
```
adrinr, line 70: This caused a bug on removing all the files when `maximum` was set up value

packages/builder/tsconfig.json
```diff
@@ -1,9 +1,6 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
-    "composite": true,
```
adrinr, line 4: This is not required, and it was causing vscode to complain. This tsconfig was added just because of the paths, as it's not a ts project

packages/server/src/api/controllers/table/utils.ts
```diff
@@ -91,26 +89,6 @@ export async function checkForColumnUpdates(
     await checkForViewUpdates(updatedTable, deletedColumns, columnRename)
   }
 
-  const changedAttachmentSubtypeColumns = Object.values(
```
samwho, line 94: Unused?

samwho (approved):

packages/server/src/api/controllers/table/utils.ts
adrinr: Yup, we got rid of this single subtype in favour of the single attachment column. This has not been used in prod, so safe to remove

packages/builder/src/constants/backend/index.js
```diff
@@ -107,10 +107,18 @@ export const FIELDS = {
...
+    type: FieldType.ATTACHMENT_SINGLE,
+    icon: "Document",
+    constraints: {
+      presence: false,
+    },
+  },
+  ATTACHMENTS: {
+    name: "Attachment List",
+    type: FieldType.ATTACHMENTS,
+    icon: "AppleFiles",
```
mike12345567, line 121: Is there any reason to use `AppleFiles` for attachment list rather than `Folder`? Probably worth keeping this consistent, incase the `Folder` icon has been used elsewhere for attachments today.

adrinr: This is a current inconsistency. We are showing a different icon when adding the column and in the grid header. I just unified it, using AppleFiles as it's displayed in the grid (hence in the client).
image

I am quite sure that this should be unified. Could you confirm @aptkingston ?

mike12345567 (approved):
LGTM! Nice to get this in, much easier to reason about I think than the subtype method!

packages/builder/src/constants/backend/index.js
mike12345567: Ah I see didn't realise it was already inconsistent, nice find! In that case if its now consistent all good!

PR #13358, opened 2024-03-26
Fix additional scrollbar appearing when hiding binding panels

aptkingston:
## Description
Tiny fix for a recent leftover from fixing drawer overflow. This just removes an additional scrollbar that currently appears if you hide any side bars when editing bindings.


deanhannigan (approved):
LGTM

PR #13359, opened 2024-03-26
Change the way we manage `testcontainers`

samwho:
## Description

Prior to this, we spun up a new container for each test that needed them. Lately we've added more tests that test against external databases (e.g. MySQL, MariaDB, Postgres, etc.) using the `testcontainers` package. As a result, we're spinning up and destroying more containers than we used to.

This PR introduces a new environment variables for controlling container reuse. `REUSE_CONTAINERS` will make it so that we use the `testcontainers` `.withReuse()` functionality to make it such that containers aren't killed at the end of a test, and if multiple identical containers are spun up during tests, only 1 gets spun up and the rest reuse that.

This new variable is set by default in CI, and you have the option to use it locally as well. I've personally noticed substantial improvements in test time, especially if you're running a small subset of tests that depend on multiple containers.

To support this reuse, the helpers that spin up containers will now also create a new unique database / collection for users of the `Datasource`. This ensures that any test making use of an external datasource is doing so in a way that cannot be interfered with by other tests.

Additionally, I've added steps to CI for pre-pulling any Docker images the tests need to use. This was because if they get pulled mid-test, it counts against the test timeout. I suspect this is why we've got so many `jest.setTimeout` calls lying around, so I've cleaned those up as well.

## Miscellanea 

This PR also sets the `test-server` CI step to use our new larger runners. This was because pre-pulling all of the images was making it such that sometimes the tests failed due to running out of disk space.


packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -34,12 +37,14 @@ const createTableSQL: Record<string, string> = {
...
-  ["mariadb", databaseTestProviders.mariadb],
-])("queries (%s)", (dbName, dsProvider) => {
+describe.each(
+  [
+    DatabaseName.POSTGRES,
+    DatabaseName.MYSQL,
+    DatabaseName.SQL_SERVER,
+    DatabaseName.MARIADB,
+  ].map(name => [name, getDatasource(name)])
+)("queries (%s)", (dbName, dsProvider) => {
```
samwho, line 47: I changed the way this works for 2 reasons:

1. It's a slightly simpler API, don't have to worry about stopping containers anymore.
2. When you run this test locally, it actually starts up all containers straight away now instead of starting them just before the test runs. This results in faster runtime for tests spinning up multiple containers.

packages/server/src/api/routes/tests/queries/mongodb.spec.ts
```diff
@@ -52,25 +55,25 @@ describe("/queries", () => {
...
   beforeAll(async () => {
     await config.init()
     datasource = await config.api.datasource.create(
-      await databaseTestProviders.mongodb.datasource()
+      await getDatasource(DatabaseName.MONGODB)
     )
   })
 
   beforeEach(async () => {
+    collection = generator.guid()
```
samwho, line 76: Because tests are reused between runs now, we can't use hard-coded table/collection names, so we generate unique ones for each test run.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -107,26 +109,29 @@ export abstract class TestAPI {
     const headersFn = publicUser
       ? this.config.publicHeaders.bind(this.config)
       : this.config.defaultHeaders.bind(this.config)
-    let request = this.request[method](url).set(
+
+    const app = getServer()
+    let req = request(app)[method](url)
```
samwho, line 114: This change serves two purposes:

1. It removes another dependency on `TestConfiguration`.
2. It's an attempt to remove flakiness, though admittedly a hail-mary. `supertest`, in all of their examples, create a new `request(app)` instance for every test. I'm wondering if this is deliberate, and `supertest` isn't actually safe to reuse, and that's causing some of the very odd failures we see. Haven't been able to prove it, but the runtime of doing this didn't seem different to "caching" the supertest instance, so I'd like to leave it in.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -136,11 +141,25 @@ export abstract class TestAPI {
...
+      // We've found that occasionally the connection between supertest and the
+      // server supertest starts gets reset. Not sure why, but retrying it
+      // appears to work. I don't particularly like this, but it's better than
+      // flakiness.
+      if (e.code === "ECONNRESET") {
+        if (attempt > 2) {
+          throw e
+        }
+        return await this._requestRaw(method, url, opts, attempt + 1)
+      }
```
samwho, line 160: This has driven me to the edge today. I was seeing several different types of spurious error when running locally:

1. `invalid localapi request` I tracked this down to be _something_ to do with Tailscale, given that this error string is only found [here](https://github.com/tailscale/tailscale/blob/4cbef205698b532791ccf881bc100f7c57f1d9cf/ipn/localapi/localapi.go#L196) when [looking it up on Google](https://www.google.com/search?q=%22invalid+localapi+request%22&sca_esv=ab02dd2696e7e4b9&source=hp&ei=jgIDZq2oMei-hbIPxKyQ2A0&iflsig=ANes7DEAAAAAZgMQnoLCaHfedW8iA1uk1-NSK1UKtjXr&ved=0ahUKEwitp57utpKFAxVoX0EAHUQWBNsQ4dUDCBc&uact=5&oq=%22invalid+localapi+request%22&gs_lp=Egdnd3Mtd2l6IhoiaW52YWxpZCBsb2NhbGFwaSByZXF1ZXN0IjIHECEYChigATIHECEYChigAUj9FlAAWNAUcAB4AJABAJgBgQGgAfgRqgEEMjAuNrgBA8gBAPgBAZgCGaAC7BHCAgsQABiABBixAxiDAcICDhAuGIAEGLEDGMcBGNEDwgIREC4YgAQYsQMYgwEYxwEY0QPCAg4QLhiABBiKBRixAxiDAcICDhAAGIAEGIoFGLEDGIMBwgILEC4YgAQYsQMYgwHCAg4QLhiDARixAxiABBiKBcICCBAuGIAEGLEDwgIOEC4YgAQYsQMYgwEY1ALCAggQABiABBixA8ICCxAuGIAEGMcBGK8BwgIFEAAYgATCAggQLhixAxiABMICCxAuGIAEGLEDGNQCwgINEAAYgAQYsQMYRhj5AcICCBAAGIkFGKIEwgIIEAAYgAQYogTCAgUQIRigAcICCRAhGAoYoAEYCpgDAJIHBDE3LjigB_mSAQ&sclient=gws-wiz). Uninstalling Tailscale appears to have solved that.
2. `WebSockets request was expected` this appears to be _something_ to do with having a debugger attached while running tests, and indeed I haven't been able to get it to happen if the debugger isn't connected. It's quite rare when the debugger is connected, though.
3. `socket hang up` for whatever reason, we appear to get an `ECONNRESET` from time to time between `supertest` and the HTTP server that `supertest` creates in order to server test requests. I wasn't able to figure out exactly what's up, but I did observe that retrying when this happens usually succeeds the 2nd time. So that's what's going on here.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -170,7 +189,18 @@ export abstract class TestAPI {
...
+        // that happens response.error is sometimes populated with `text` that
+        // gives more detail about the error. The `message` is almost always
+        // useless from what I've seen.
+        if (response.error.text) {
+          response.error.message = response.error.text
+        }
+        throw new Error(message, { cause: response.error })
+      } else {
+        throw new Error(message)
+      }
```
samwho, line 203: Extracting this extra error information is what gave me the clues I needed to find the problems detailed in https://github.com/Budibase/budibase/pull/13359/files#r1539823938.

packages/server/src/tests/utilities/api/base.ts
mike12345567: I feel like this is worth a try, its very possibly causing problems re-using the same request and I doubt this will change performance of the tests much at all - if its even recognisable.

packages/server/src/integrations/tests/utils/postgres.ts
```diff
@@ -1,10 +1,10 @@
...
+import { GenericContainer, Wait } from "testcontainers"
 
-let container: StartedTestContainer | undefined
-
-export async function start(): Promise<StartedTestContainer> {
-  return await new GenericContainer("postgres:16.1-bullseye")
+export async function postgres(): Promise<Datasource> {
+  const container = await new GenericContainer("postgres:16.1-bullseye")
+    .withName("budibase-test-postgres")
+    .withReuse()
```
mike12345567, line 7: This all looks good - one concern I have, just asking how it works, the `withReuse` - is it still tearing down the test containers correctly in dev environments after a test has ran? Just wondering if its leaving these containers running forever, while thats technically faster locally, it might cause some headaches if the containers are never reset.

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -42,10 +41,10 @@ describe.each([
   let datasource: Datasource
 
   function saveTableRequest(
-    ...overrides: Partial<SaveTableRequest>[]
+    ...overrides: Partial<Omit<SaveTableRequest, "name">>[]
   ): SaveTableRequest {
     const req: SaveTableRequest = {
-      name: uuid.v4().substring(0, 16),
+      name: generator.guid().replaceAll("-", "").substring(0, 16),
```
mike12345567, line 47: Why the change here out of interest?

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -41,18 +41,17 @@ describe("postgres integrations", () => {
...
 
-  afterAll(async () => {
-    await databaseTestProviders.postgres.stop()
-  })
-
   beforeEach(async () => {
     async function createAuxTable(prefix: string) {
       return await config.createTable({
-        name: `${prefix}_${generator.word({ length: 6 })}`,
+        name: `${prefix}_${generator
```
mike12345567, line 51: I'm a little concerned about the need to remember that table names can no longer be static, they must be unique across our entire test environment, while this does seem like a major speed improvement, I'm just wondering if theres a better way we could do this.

For most databases we should be able to use a different schema, could this be an option instead of making sure the table names are always unique?

If we must go with unique table names, it would be good to try and centralise this logic somewhere, like a `generateUniqueTableName()` with an explanation of why this function is used, simply to capture this reasoning a bit.

adrinr:
If we reuse containers between tests, can we not have tests polluting each other?

adrinr (commented):
If we reuse containers between tests, can we not have tests polluting each other?

globalSetup.ts
```diff
@@ -2,6 +2,8 @@ import { GenericContainer, Wait } from "testcontainers"
 
 export default async function setup() {
   await new GenericContainer("budibase/couchdb")
+    .withName("budibase-test-couchdb")
```
adrinr, line 5: Do we need the name? What will happen if you are running 2 test consoles locally in parallel?

samwho:
Yes, I've made several changes to tests in this PR to prevent that. This is a problem we already work around when we share the CouchDB container.

packages/server/src/integrations/tests/utils/postgres.ts
samwho: In my experience they do get left up. It's possible to tear them down, but I removed that code on request from @adrinr in a prior PR. I think there was a belief that `testcontainers` would stop them, but I have observed that's not the case. I'm happy to bring that code back, but doing so will lose the performance gain of having containers always running, which is substantial for tests that use them.

packages/server/src/integration-test/postgres.spec.ts
samwho: I did think a little about this, it _is_ possible to create unique schemas / collections. I can give it a whirl if you'd like. I didn't do it initially because it seemed like a more surprising approach than this one, but an extra layer of safety would be nice.

packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -61,7 +66,7 @@ describe.each([
     // We re-fetch the datasource here because the one returned by
     // config.api.datasource.create has the password field blanked out, and we
     // need the password to connect to the database.
-    const ds = await dsProvider.datasource()
+    const ds = await dsProvider
```
adrinr, line 69: This await on a variable seems not natural to me. But not sure what would be the best naming

packages/server/src/integration-test/mysql.spec.ts
```diff
@@ -47,17 +47,13 @@ describe("mysql integrations", () => {
...
   })
 
-  afterAll(async () => {
-    await databaseTestProviders.mysql.stop()
-  })
-
   beforeEach(async () => {
     primaryMySqlTable = await config.createTable({
-      name: uuidv4(),
+      name: generator.guid().replaceAll("-", "_").substring(0, 10),
```
adrinr, line 56: In cases like this, should we not use `generator.word({length})`?

packages/server/src/integration-test/postgres.spec.ts
```diff
@@ -89,7 +88,7 @@ describe("postgres integrations", () => {
     }
 
     primaryPostgresTable = await config.createTable({
-      name: `p_${generator.word({ length: 6 })}`,
+      name: `p_${generator.guid().replaceAll("-", "").substring(0, 6)}`,
```
adrinr, line 91: We should either use `word`, or create a custom helper to DRY these lines

globalSetup.ts
samwho: We don't need it, no, we could have as many containers running in parallel as we want. I did test this and it works fine. But reusing the containers gives a non-trivial performance benefit.

packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
samwho: Yeah, I thought it looked weird as well. I can instead pass in the function, rather than the promise, but I believe this comes with the downside that the containers are not started in parallel, and will lead to each test waiting on its respective container at the start of the tests (if the container is not already running).

packages/server/src/integration-test/mysql.spec.ts
samwho: No, table names need to be unique, and `generator.word` is low-entropy.

packages/server/src/integration-test/postgres.spec.ts
samwho: `word` doesn't provide enough uniqueness. I erred away from creating a helper because the line of code is simple enough to be understood in-place, while a helper would require you to go to its definition to understand what it's doing.

I don't mind creating a helper if that's what you'd prefer, but I don't personally consider the repetition here that much of a problem.

packages/server/src/integration-test/mysql.spec.ts
samwho: I did a comparison of this on one of Mel's PRs a little while ago: https://github.com/Budibase/budibase/pull/13165#discussion_r1510918899

packages/server/src/integrations/tests/utils/postgres.ts
mike12345567: I wonder if there is someway we could detect the teardown is needed outside of CI - my preference would be to teardown locally to avoid containers just running forever, which could lead to weirdness eventually as number of tables builds up etc. Could be worth investigating if we can just throw a teardown in if not in CI.

packages/server/src/integration-test/postgres.spec.ts
mike12345567: I would probably give it a go - I know its slightly surprising, but at least means if you're copying existing tests to create new ones you don't miss the semi-hidden rule of table names needing to be totally unique across all tests.

packages/backend-core/tests/core/utilities/testContainerUtils.ts
```diff
@@ -19,48 +20,71 @@ interface ContainerInfo {
...
-  const couchPort = getExposedPort(couch, 5984)
+  const couchPort = getExposedV4Port(couch, 5984)
   if (!couchPort) {
     throw new Error("CouchDB port not found")
   }
 
   const configs = [
     { key: "COUCH_DB_PORT", value: `${couchPort}` },
-    { key: "COUCH_DB_URL", value: `http://localhost:${couchPort}` },
+    { key: "COUCH_DB_URL", value: `http://127.0.0.1:${couchPort}` },
```
samwho, line 82: This, and the various other ipv4 related changes in this file, were because I noticed that Docker on CI and locally, when you expose a port, was actually exposing 2 ports: one ipv4, the other ipv6. Sometimes these ports overlap with each other, sometimes they overlap with other containers. Port space isn't shared between v4 and v6, so when you connected to `localhost:<port>` it was a gamble as to whether you'd connect to v4 or v6, which could mean you think you're connecting to CouchDB, but end up connecting to MySQL instead.

To solve this, the code in this file ensures we always connect over ipv4 and always connect to the ipv4 port.

packages/server/scripts/test.sh
```diff
@@ -4,8 +4,8 @@ set -e
 if [[ -n $CI ]]
 then
   export NODE_OPTIONS="--max-old-space-size=4096 --no-node-snapshot $NODE_OPTIONS"
-  echo "jest --coverage --maxWorkers=2 --forceExit --workerIdleMemoryLimit=2000MB --bail $@"
-  jest --coverage --maxWorkers=2 --forceExit --workerIdleMemoryLimit=2000MB --bail $@
+  echo "jest --coverage --maxWorkers=4 --forceExit --workerIdleMemoryLimit=2000MB --bail $@"
+  jest --coverage --maxWorkers=4 --forceExit --workerIdleMemoryLimit=2000MB --bail $@
```
samwho, line 8: We've got more vCPUs now, so I've bumped this.

packages/server/src/tests/jestSetup.ts
```diff
@@ -2,17 +2,11 @@ import env from "../environment"
 import { env as coreEnv, timers } from "@budibase/backend-core"
 import { testContainerUtils } from "@budibase/backend-core/tests"
 
-if (!process.env.DEBUG) {
-  global.console.log = jest.fn() // console.log are ignored in tests
-  global.console.warn = jest.fn() // console.warn are ignored in tests
-}
```
samwho, line 8: Removing this didn't really seem to make a difference, and I've been confused in the past about where my `console.log` statements went. Because we lint against this now, I don't think we need it.

packages/server/src/tests/jestSetup.ts
```diff
@@ -2,17 +2,11 @@ import env from "../environment"
...
-
 if (!process.env.CI) {
-  // set a longer timeout in dev for debugging
-  // 100 seconds
+  // set a longer timeout in dev for debugging 100 seconds
   jest.setTimeout(100 * 1000)
 } else {
-  jest.setTimeout(10 * 1000)
+  jest.setTimeout(30 * 1000)
```
samwho, line 9: Increasing the default timeout to 30 seconds because, in practice, no tests should timeout. We much more often so timeouts causing flakes than catching real problems. I've gotten rid of the various calls to `jest.setTimeout` in `packages/server`.

packages/server/src/integrations/tests/utils/postgres.ts
samwho: What I've done instead is control it with an environment variable (`REUSE_CONTAINERS`). It's off by default and we turn it on in CI.

packages/server/src/integration-test/postgres.spec.ts
mike12345567: Not too worried - the table names now don't particularly matter, was just trying to protect against the issue in future!

mike12345567 (approved):
LGTM! Looking forward to seeing the performance improvements this should bring - we'll probably be at the point where the build is slower than the tests often!

PR #13362, opened 2024-03-27
Runs submodule checks only if you changed the submodule ref

adrinr:
## Description
Given that the monorepo is not dependant on the account-portal submodule (if not the other way around), we don't want the submodule getting out of sync to pollute all our pipelines. For this, we will only do proper submodule checks if you are actually updating it. We want to keep the checks in this case to get sure that the submodule is pointing to the latest version instead of an old commit.

Tests:
1. Run with submodule pointing to latest master. The submodule is being checked and passes. https://github.com/Budibase/budibase/actions/runs/8450482156/job/23146754919?pr=13362
image

2. Run with submodule NOT pointing to latest master. The submodule is being checked and fails. https://github.com/Budibase/budibase/actions/runs/8450509151/job/23146834324?pr=13362
image

3. Run without submodule changes. No checks run. https://github.com/Budibase/budibase/actions/runs/8450532668/job/23146907889?pr=13362
image


mike12345567 (approved):

aptkingston (approved):
Awesome! Thank you @adrinr 🙏

jvcalderon (approved):
LGTM

PR #13364, opened 2024-03-27
Ignore merge conflicts on pro submodule checks

adrinr:
## Description
The current flow is quite confusing, as we are always checking if the submodule ref is pointing to master head. This requires the following flow:
1. You create a PR on your submodule
2. You create a PR on the monorepo, with the submodule pointing to your changes
3. Wait until both are green and ready to merge
4. Merge the submodule one
5. Update the monorepo to point to the latest master
6. Merge

This means that the PR checks will be broken after the step 4 until the step 6 is finished.
This happens because the merge to the submodule creates a merge commit, getting the checks out of sync.

This PR's changes will ignore this merge commit.

Test runs:
1. Pro submodule not in sync. It will fail. https://github.com/Budibase/budibase/actions/runs/8451136427/job/23148820110?pr=13364
image

2. Pro submodule in sync with master. Passing. https://github.com/Budibase/budibase/actions/runs/8451114982/job/23148739496?pr=13364
image

3. Pro submodule in sync with master excluding merge commits. Passing. https://github.com/Budibase/budibase/actions/runs/8451154636/job/23148881045?pr=13364
image


packages/pro
adrinr, line 1: This was updated to point to the latest commit ignoring the merge commit:
image

mike12345567 (approved):
I like it - worth making it clear to the team that when updating pro/account-portal in future they can just merge both PRs at the same time now rather than needing to update to the merge commit.

PR #13371, opened 2024-03-28
[Added] POST /api/global/users/sso endpoint

jvcalderon:
## Description

As part of the new internal process handling for onboarding flow, verifiable SSO has an additional level of validation. 
It requires an endpoint to send the already verified ssoId in order to be persisted in `global-info`. This change allows users with 'password' accounts to use SSO if needed to login.


shogunpurple (approved):
LGTM

Issue #7201, opened 2022-08-10
Navigation Sub Menu


lachlanhickey:
I would love to see the integration of sub menu items and dropdowns. Currently the only function from holding me back from implementing Budibase is the minimal ability provide a large amount of navigation options.

mmays77:
I would like to see this feature made available in Budibase.

ConorWebb96:
Closing as this feature has been merged and released.

Issue #6659, opened 2022-07-12
Nested Links in Navigation Menu


poirazis:
While building complex applications, having all the Menu / Navigation Links as Level 1 is very cumbersome. We are often faced with the scenario where we need to nest navigation links as Menu / Submenu

Elliot-Construct:
This also may be offset by role specific menus
See #6790 for info

Elliot-Construct:
not stale still very relevent

darkzone:
still relevant

seandlg:
Check out [this](https://github.com/poirazis/bb-plugin-TabContainer) :-)

scmsystm:
> Check out [this Tabs layout](https://github.com/poirazis/bb-plugin-TabContainer), which may pose a viable alternative :-)

Tabs layout are nice for < 5 contents, tab navigation does not scroll on phone view.
We still need nested links inside native budibase navigation 👍

jkirkpatrick:
not stale issue, please fix bot

alican92:
Defiantly not a stale issue, this would be VERY helpful in instances where a single Budibase APP has multiple pages that need to be grouped.

rafalpawlik1981:
Hi, is there any progress about this topic?

techthrasher:
This functionality is key to making complex applications.

ConorWebb96:
This is now complete, closing this issue. 🤝

PR #13375, opened 2024-03-28
Nested nav links

aptkingston:
## Description
This PR adds new features for app navigation. You can now create groups of nested links, allowing you to create many more navigation links in a denser format.

image

## Configuration
The UI for configuring navigation items is familiar, sharing the same design as the components we use to configure forms in form blocks or groups of buttons.

<sub>Edit: border radius on last nav item in all screenshots has since been fixed, so please ignore it if you're eagle-eyed and spotted it</sub>

image

The settings for each navigation item are displayed in a popover:
image

The new setting available here is the ability to select "Open sub links" for a nav item. This will update the settings to allow you to configure your sub links:
image

The sub link configuration is done in a drawer, and is similar to how we used to configure all links:
image

## Usage
Inside apps, sub link navigation items display an arrow, indicating they can be interacted with to show sub links:
image

When using a top nav, hovering your mouse over the nav item will show the sub links:
image

When using a side nav, all sub links are collapsed by default:
image

Sub links are revealed by clicking on the nav item:
image

## Other stuff
- All navigation related properties are now bindable, meaning you can create more dynamic, reactive text or links for your navigation
- Updated styles for navigation items slightly
- Rewrote entire navigation settings panel to be more consistent with the layout of component settings
- Updated comboboxes to uses popovers (like the picker) to ensure overflow is never an issue
- Add CSS transitions to popovers when they hide or show due to the presence of drawers
- Don't show live eval side panel when context is an empty object

## Addresses
- https://linear.app/budibase/issue/BUDI-8105/nested-links-in-navigation-menu
- https://github.com/Budibase/budibase/issues/6659
- https://github.com/Budibase/budibase/issues/7201

## Launchcontrol
Added the ability to configure nested navigation links.

## Feature branch env
[Feature Branch Link](http://fb-nested-nav-links.fb.qa.budibase.net)


deanhannigan (commented):
Code LGTM! Some minor issues:

- `Bug` - Nav link `Access`/`Nav item` not set by default
    - This affects existing apps and newly created ones
    - A new, manually created `Nav item` *does* default to `Inline link` and `Access` to `Basic`
    - The lack of default `Access` config seems to affect visibility 
      - Added 2 `Public` pages with link and text. Despite being public screens the links didn't appear.
      - `Access` should default to the screen permission and `Nav item` to `Inline link`

~~`Bug` - Nested sublinks do not appear to honour permissions of an internal linked screen.~~

- `Bug` z-index issue for popover nav menu
  - With navigation set to the top of the screen and the screen width reduced
  - The navigation popover appears to display under the subsequent links
Screenshot 2024-04-09 at 11 16 37
- `Bug` scrollbar appearing in navigation for long nav items.
  - Configure the position of navigation to be on the side
  - Add a very long navigation link name
  - Instead of truncating, the navigation begins to scroll
Screenshot 2024-04-09 at 11 28 47

aptkingston:
> `Bug` - Nested sublinks do not appear to honour permissions of an internal linked screen.

Just on this one:
> Bug - Nested sublinks do not appear to honour permissions of an internal linked screen.

This is by design - there is no access setting for nested links, only on the main nav item. Access has to be controlled on the dropdown as a whole.

deanhannigan:
> Just on this one:
> 
> > Bug - Nested sublinks do not appear to honour permissions of an internal linked screen.
> 
> This is by design - there is no access setting for nested links, only on the main nav item. Access has to be controlled on the dropdown as a whole.

Could we check the `isInternal` result on the configured URL to determine if the link is screen? Then apply the screen permission automatically?

aptkingston:
> > Just on this one:
> > > Bug - Nested sublinks do not appear to honour permissions of an internal linked screen.
> > 
> > 
> > This is by design - there is no access setting for nested links, only on the main nav item. Access has to be controlled on the dropdown as a whole.
> 
> Could we check the `isInternal` result on the configured URL to determine if the link is a screen? Then apply the screen permission automatically?

We could but I'd prefer not to - it goes against the idea of explicitly setting the access level for links that we normally do. I think it's better to keep it simple and follow the normal convention here that each top level nav item has an access level setting and that's it.

aptkingston:
Support added for mobile top navs - it'll now properly expand rather than use popovers (there should never be popovers in the side nav):
image

Text overflow updated to handle long names for both links and dropdowns:
image

Working on the automatic link creation now (great catch - totally forgot we auto-create and remove links too).

deanhannigan (approved):
LGTM!

Issue #13390, opened 2024-04-03
On screen load doesn't seem to be triggering for apps in preview mode


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 2.22.14
  - App Version: 2.22.14

**Describe the bug**
On screen load isn't triggering on previewed apps. This does seem to be working for published apps, this makes it a bit of a pain to check your on screen load actions. As you need to publish every time you make a change to see it being affected.

**To Reproduce**
1. Go to your design screen
2. Add an update state action with the value of test to your on screen load 
3. Reference this in a paragraph component
4. Click the preview button
5. See error - no test is outputted.

**Expected behavior**
On screen loads should trigger in preview mode as well as published apps.

**Screenshots**
<img width="1512" alt="Screenshot 2024-04-03 at 09 47 32" src="https://github.com/Budibase/budibase/assets/126772285/085bd5d9-8697-4470-8b6f-40bf95f7fa48">
<img width="311" alt="Screenshot 2024-04-03 at 09 47 47" src="https://github.com/Budibase/budibase/assets/126772285/779ca67d-6dfa-4a2d-a8ef-e99a08628199">
<img width="1425" alt="Screenshot 2024-04-03 at 09 48 00" src="https://github.com/Budibase/budibase/assets/126772285/c45bc911-bd79-4a2b-b1f4-bbf12e4e6cee">
  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

esellier:
Hi,
As a workaround, we use the Interval plugin with an initial timer at 1 sec (which is not perfect, but we cannot have less).
The timer actions:
- Continue If {{ State.Init }} is not 1
- Notify user (just to check it's not running endlessly)
- Set {{ State.Init }} to 1
- Set other required State values for initializing the page

The Interval plugin is of course not displayed (Display component in frontend set to false in builder).
Conditions : Hide component if {{ State.Init }} is not empty
Hiding the Interval component just stops it.

That makes the job even though it's not perfect.
We could improve it by checking the document.readyState==="complete" in step 1 of the Interval's actions.

PR #13394, opened 2024-04-03
Fix race condition when running screen load actions

aptkingston:
## Description
Fixes an issue where screen load actions would sometimes not run due to a race condition.

## Addresses
- https://linear.app/budibase/issue/BUDI-8141/on-screen-load-doesnt-seem-to-be-triggering-for-apps-in-preview-mode
- https://github.com/Budibase/budibase/issues/13390


deanhannigan (approved):
LGTM!

Issue #5587, opened 2022-04-26
S3 File Upload component should work with other S3-compatible services, not just AWS


melohagan:
### Discussed in https://github.com/Budibase/budibase/discussions/5459

<div type='discussions-op-text'>

<sup>Originally posted by **daawaan** April 17, 2022</sup>
Hello, I've been trying to use the S3 File Upload component since the 20MB file upload limit for attachments is too small for our use case which is explained further in #5457. Unfortunately, the S3 component isn't showing any of the S3 datasources in the dropdown for the component which I've already issued in #5454. We're using digitalocean spaces in our case and I was wondering if our datasource wasn't showing up because it isn't from AWS? Its also stated here in #4030 the first pull request for the S3 File Upload component that

> S3 datasources with custom endpoints are not supported. This is mainly due to the non-compliance with presigned URL domain names. Real S3 uses bucket.domain as the FQDN, and using a custom endpoint when generating a presigned URL spits out a value of something like bucket.10.0.0.1 for example, if using a local IP address. Due to this behaviour we just have to put a limitation on this functionality that it will only work with real S3 and not MinIO.
</div>

anborg:
Please Note:  Minio - has changed license terms.  No longer opensource or free.

"Designed for developers who are building open source applications in compliance with the GNU AGPL v3 license and are able to support themselves. It is fully featured. If you distribute, host or create derivative works of the MinIO software over the network, the GNU AGPL v3 license requires that you also distribute the complete, corresponding source code of the combined work under the same GNU AGPL v3 license. This requirement applies whether or not you modified MinIO." 

If you customize budibase and add a form; because  bb uses minio, **you have to release your org bb-custom-code public! Whether or not you modified MinIO**.

Please move away from MinIO.

remcoboerma:
Backblaze support would be really great. It's S3 compatible.

liongkj:
[cloudflare image](https://developers.cloudflare.com/images/cloudflare-images/api-request/) seems to be a nice choice also

matteoveglia:
+1 for this
I should have checked this functionality before building most of my app which is now on hold as this functionality is vital.

I found a bit of a workaround here, it seems to use the same AWS sdk that budibase uses and it works!
https://pilvio.com/en/2021/01/21/uploading-files-from-html-page-to-an-s3-bucket/

[...]

PR #13395, opened 2024-04-03
Add support for file uploads with custom S3 endpoints

CSLTech:
Addresses #5459 and #5587

## Description
Add support for file uploads with custom S3 endpoints. This still may not work with all providers, but most S3 compatible providers do have the same kind of support and similar public URLs (If they are going to be used). I think having a mostly working configuration is much better than the current state of silently not working.

This was tested with google cloud storage and regression tested with S3.

## Addresses
- https://github.com/Budibase/budibase/discussions/5459
- https://github.com/Budibase/budibase/discussions/5587

## Launchcontrol

Add support for file uploads with custom S3 endpoints. (May not work with all providers)


shogunpurple (approved):
LGTM, thanks for the contribution @CSLTech!

PR #13397, opened 2024-04-03
Clean and populate url for single attachment

adrinr:
## Description
Do not persist url for single attachment columns. These are populated for display only, and persisting it to db causes issues with features such as import/export


mike12345567 (approved):
LGTM

mike12345567:
Can't merge this for now - will need to merge this to a separate branch - reverting single attachment column as there are a few issues found in app experience, don't want it to be accidentally released.

PR #13399, opened 2024-04-03
Single attachment column type

mike12345567:
## Description
Adding a new column type for single attachments. This will be stored as the same shape as the current attachments, but as a single object (instead of an array). Having this new type will allow having columns with a single attachment, simplifying validations and bindings.

After discussing with @AndyWhann and @aptkingston, we are also changing the current attachment column to use paperclips icons, to match usages in the design section.

#### Adding a new single attachment column
image

##### Editing an empty cell
image

##### Editing a none empty cell (will not allow adding more than one file)
image

##### Required single attachment validation
image

#### Adding a new field component
image
image

#### Inline editing grid
This will work like on the builder end
image

##### New row
image

It will not allow adding more than one image for single attachments
image

##### Editing existing row
image

#### Form block
It will act as the side panel
##### New row
image

##### Editing row
image

#### Multistep block
image

#### Repeater with images
image

## Launchcontrol
Adding a new column type single attachment

## Feature branch env
[Feature Branch Link](http://fb-budi-8122-single-attachment-column-type.fb.qa.budibase.net)


packages/frontend-core/src/components/grid/cells/AttachmentSingleCell.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+  $: onFileChange = value => {
+    value = value[0] || null
+    onChange(value)
+  }
+</script>
+
+<AttachmentCell
+  {...$$restProps}
+  maximum={1}
```
aptkingston, line 17: We'll need to `bind:api` here to ensure the bind is properly passed through (which makes things like keyboard event or programatic cell focus/blur work), so we'll need to add an explicit `export let api` to facilitate that.

packages/client/manifest.json
```diff
@@ -4322,6 +4322,103 @@
...
+      },
+      {
+        "type": "text",
+        "label": "Extensions",
+        "key": "extensions"
+      },
+      {
+        "type": "number",
+        "label": "Max attachments",
+        "key": "maximum",
```
aptkingston, line 4360: We can remove the max attachments settings for the single attachment field since it doesn't make sense to include it - the max is always 1.

packages/client/manifest.json
```diff
@@ -4322,6 +4322,103 @@
...
+    "styles": ["size"],
+    "requiredAncestors": ["form"],
+    "editable": true,
+    "size": {
+      "width": 400,
+      "height": 200
+    },
+    "settings": [
+      {
+        "type": "field/attachment_single",
```
aptkingston, line 4337: We need to define what settings component should be rendered for this new type `field/attaachment_single`. The existing `FormFieldSelect` will handle it, so we just need to add 
```js
  "field/attachment_single": FormFieldSelect,
```
to `componentSettings.js`.
Without this, the `field` setting is never rendered.

aptkingston (changes requested):
We need to add `attachmentsinglefield` to `componentStructure.json` as otherwise it won't appear as a possible component to add.

Few more bits to review but have a screaming baby to attend to! So here's some initial feedbacl.

packages/client/src/components/app/forms/AttachmentField.svelte
```diff
@@ -78,14 +85,14 @@
   {validation}
   {span}
   {helpText}
-  type="attachment"
+  {type}
   bind:fieldState
   bind:fieldApi
   defaultValue={[]}
```
aptkingston, line 91: We'll also need to override the defaultValue here as the default value of an array is incorrect for a single attachment field, and causes live eval to show that the value of a new single attachment field is an array, which is a bit confusing for the user.

image

So I think we should override this in the single attachment field to set the default value to null.

adrinr:
> `attachmentsinglefield`

The existing attachment list uses a clip instead of the folder we use in the data section. Would you use the clip as well for this new column as well?

image

packages/frontend-core/src/utils/rows.js
```diff
@@ -21,6 +21,8 @@ const isBetterSample = (newValue, oldValue) => {
   }
   if (
     newType === "object" &&
+    Array.isArray(newValue) &&
+    Array.isArray(oldValue) &&
```
aptkingston, line 25: Do we need to add this? This was designed to work with both arrays and objects (like JSON columns). I guess if one if an array and one is an object then we don't want to consider this at all, but in that case I'd rather add a specific check up above this that if one is an array and one is not an array then we return false.

aptkingston (approved):
LGTM other than that one question 👍

packages/frontend-core/src/utils/rows.js
adrinr: I will split this up and I will change this `isArray` check to check nulls. This was the error, as we can have empty objects, and `Object.keys(null)` is not defined

adrinr (approved):

PR #13407, opened 2024-04-04
Handle singleattachment on AttachmentCleanup

adrinr:
## Description
Handle singleattachment on AttachmentCleanup

### DIFF
--- a/packages/server/src/utilities/rowProcessor/attachments.ts
+++ b/packages/server/src/utilities/rowProcessor/attachments.ts
@@ -1,12 +1,6 @@
 import { ObjectStoreBuckets } from "../../constants"
 import { context, db as dbCore, objectStore } from "@budibase/backend-core"
-import {
-  FieldType,
-  RenameColumn,
-  Row,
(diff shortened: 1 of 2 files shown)


packages/server/src/utilities/rowProcessor/attachments.ts
```diff
@@ -25,6 +19,27 @@ export class AttachmentCleanup {
     }
   }
 
+  private static extractAttachmentKeys(
+    type: FieldType,
+    rowData: any
```
samwho, line 24: Would it make sense to type this `rowData: { key: string }`?

samwho (approved):

packages/server/src/utilities/rowProcessor/attachments.ts
adrinr: It might also be an array, and the rows are not typed at all. Also, the shape would need to depend on the FieldType, making the types more messy for a private function. So I think it's not really worth it in this case

PR #13408, opened 2024-04-04
Tests and handle imports

adrinr:
## Description
Testing single attachment columns and handle imports


mike12345567 (approved):
LGTM! nice to cover the import as well.

PR #13414, opened 2024-04-05
NPS survey

melohagan:
## Description
Refactor to use the Posthog client from the analytics file, and not instantiate it within svelte.


shogunpurple (approved):
LGTM

Issue #13416, opened 2024-04-05
Snippets don't get transfered whenever exporting and importing an app


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Docker compose to cloud

**Describe the bug**
Exported an app locally and imported it into a cloud account, noticed that the snippet didn't exist after seeing `error whilst executing JS`.

**To Reproduce**
1. Create a simple snippet
2. Export the app within a snippet
3. Import this app elsewhere
4. See that the snippet isn't imported along with the app.

**Expected behavior**
Snippets created with apps should get imported along with the app.

PR #13417, opened 2024-04-05
Add snippets to app imports

aptkingston:
## Description
Add snippets to app imports.

## Addresses
- https://github.com/Budibase/budibase/issues/13416
- https://linear.app/budibase/issue/BUDI-8149/snippets-dont-get-transfered-whenever-exporting-and-importing-an-app


samwho:
@aptkingston would you mind adding a test for this fix? 🙏 Happy to help if needed.

deanhannigan (approved):
LGTM!

adrinr (approved):

aptkingston:
> @aptkingston would you mind adding a test for this fix? 🙏 Happy to help if needed.

There doesn't appear to be any tests for checking metadata is copied over at the moment, so I'll have to come up with a new suite of tests to ensure all the other keys are copied over too as I'm only adding another key to copy over with this change. I'll maybe store a new app export and use that as reference because I don't think there's a better way to test this.

samwho:
Would it not be possible to create an app, populate its metadata, do an export, and then do an import, then assert agains the newly imported app that things have been populated correctly?

aptkingston:
That would work too but I figured it would require filesystem access to store the temp export, but maybe I can keep it all in memory. Maybe filesystem access is OK too, but just seems a little flaky to me. I'll see if I can do it all in memory.

samwho:
I'm _fairly sure_ we access the filesystem in some other tests and it doesn't cause flakes. I wouldn't worry about it. Filesystem access certainly _shouldn't_ be flaky, if it is that's something we'd want to investigate.

PR #13418, opened 2024-04-05
Signature field and component support

deanhannigan:
## Description
Enhance your app with direct signature capture in Budibase forms!

- Internal data sources now directly support a `Signature` data type. 
- Forms can now capture and store signatures with the `SignatureField` component type.
- Signatures are stored as `.png` files and can be embedded in your applications.
- Supports `mouse` capture and also `touch` entry for mobile users.

## Addresses
- https://linear.app/budibase/issue/BUDI-8104/signature-component
- `Bug` - Single attachment validation behaviour was missing. It has now been added and supports `Required` and `Max upload size` options
- `Bug` - added support for single attachment type in the edit row modal.

## Screenshots
View and edit signature in the data section
signing

Capture signatures in your app forms!
Screenshot 2024-04-08 at 17 06 45

## Launchcontrol
Signature field support added to internal datasources and forms.

## Feature branch env
[Feature Branch Link](http://fb-feature-signature-field-and-component.fb.qa.budibase.net)


packages/backend-core/src/environment.ts
```diff
@@ -104,7 +104,7 @@ const environment = {
   JS_BCRYPT: process.env.JS_BCRYPT,
   JWT_SECRET: process.env.JWT_SECRET,
   JWT_SECRET_FALLBACK: process.env.JWT_SECRET_FALLBACK,
-  ENCRYPTION_KEY: process.env.ENCRYPTION_KEY,
+  ENCRYPTION_KEY: "wgeBLcM6sXQTmaH0UtjpO+hCQhCBDn/c7GQHlnqgPCc=",
```
aptkingston, line 107: Wee leftover, just needs reverted.

packages/bbui/src/Modal/ModalContent.svelte
```diff
@@ -27,6 +27,7 @@
   export let secondaryButtonText = undefined
   export let secondaryAction = undefined
   export let secondaryButtonWarning = false
+  export let enableGrid = true
```
aptkingston, line 30: After reading this, I think we'd be better to call it something more generic, yea. Maybe just call it `custom` - indicating it's a custom modal content rather than spectrum.

packages/builder/src/components/backend/DataTable/TableDataTable.svelte
```diff
@@ -55,6 +59,7 @@
 <div class="wrapper">
   <Grid
     {API}
+    {isDark}
```
aptkingston, line 62: Could we refactor this to be called `darkMode`? I think that's just a bit more standard than `isDark`.

packages/client/src/components/Component.svelte
```diff
@@ -192,6 +193,7 @@
   let pad = false
   $: pad = pad || (interactive && hasChildren && inDndPath)
   $: $dndIsDragging, (pad = false)
+  $: isDark = !$themeStore.theme?.includes("light")
```
aptkingston, line 196: Since we can use the normal `context` to get the theme (like we moved over to doing in the signature field, we can remove this from the core Component.

packages/client/src/components/app/GridBlock.svelte
```diff
@@ -35,6 +35,7 @@
 
   let grid
 
+  $: isDark = $component.isDark
```
aptkingston, line 38: We can use the new context technique to get the theme like the signature field does.

packages/client/src/components/app/forms/SignatureField.svelte
```diff
@@ -0,0 +1,140 @@
...
+
+  $: currentTheme = $context?.device?.theme
+  $: isDark = !currentTheme?.includes("light")
+</script>
+
+<SignatureModal
+  onConfirm={saveSignature}
+  title={fieldSchema?.name}
+  value={fieldState?.value}
+  {isDark}
```
aptkingston, line 85: NAB: would also prefer calling this `darkMode` but not a big deal.

packages/frontend-core/src/components/grid/cells/SignatureCell.svelte
```diff
@@ -0,0 +1,185 @@
...
+
+  $: editable = focused && !readonly
+  $: {
+    if (!focused) {
+      close()
+    }
+  }
+
+  const onKeyDown = () => {
+    return isOpen
```
aptkingston, line 34: If the signature component doesn't actually support any keyboard events (which it looks like it doesn't), then it would be better to return `false` always here - it means that using keys like tab or escape will always work to close the cell.

aptkingston (commented):
Awesome work Dean! Nothing major but just a few suggestions 👌

aptkingston (approved):

mikesealey:
Great work Dean, this will be a well-used feature, I'm sure.

One thing I have noticed is that it still works when the form is set to disabled, both in the builder and on the published version.

[Link to example](https://jam.dev/c/380ffdca-b5b7-4e68-831c-44504c4e23ca)

deanhannigan:
> Great work Dean, this will be a well-used feature, I'm sure.
> 
> One thing I have noticed is that it still works when the form is set to disabled, both in the builder and on the published version.
> 
> [Link to example](https://jam.dev/c/380ffdca-b5b7-4e68-831c-44504c4e23ca)

Niiice catch!

mikesealey:
* Creating a form manually, when I use the signature component, the text/title of the signature input pop-over is `undefined` even when the New Signature component is set up with a field name. Is the title of the popover here fetched from the schema? 

_Steps to replicate:_

- Create a New Form with a Schema set to Custom
- Include a signature field component with a value set for Field
- Publish, and enter a signature

image


* Checking `Disabled` on a form block should disable all components regardless of their individual `Disabled` setting.  The individual `Disabled` setting works well though.

* We also don't get this little menu to allow the component to be narrowed into columns - might be a nice touch to be able to have _sign_ and _print_ side by side in a form. 
image

packages/server/src/utilities/rowProcessor/attachments.ts
```diff
@@ -34,7 +35,7 @@ export class AttachmentCleanup {
       return []
     }
 
-    if (type === FieldType.ATTACHMENTS) {
+    if (type === FieldType.ATTACHMENTS || type === FieldType.SIGNATURE) {
```
adrinr, line 38: I see this too many times in the code, when we should probably have some helpers to avoid repetitions

adrinr: We can leave this for a further clean up

adrinr (commented):
Should this new type behave as Attachments, or should we use the single attachment one? Does it make sense to have a multiple signature column? We used the single attachment to simplify bindings and similar, not sure if we should use the same approach for it

adrinr:
> Should this new type behave as Attachments, or should we use the single attachment one? Does it make sense to have a multiple signature column? We used the single attachment to simplify bindings and similar, not sure if we should use the same approach for it

@deanhannigan I can help refactoring it to be saved as single item if needed

adrinr (dismissed):
Not sure if it happens everywhere, but using firefox I am getting the following:
image

deanhannigan:
> Not sure if it happens everywhere, but using firefox I am getting the following: image

That's really odd. It should be confined to the canvas bounds. I tried to replicate on the latest Firefox(MacOS) but couldn't get the same behaviour. Were there any console errors or specific steps you followed?

adrinr (commented):
We just changed both attachment and user column to be stored as single items instead of arrays to simplify most use cases. The most common use case will be that we have a single signature per column, so we want to keep it simple in order to avoid the array notation on apis and bindings:
image


I think we should keep this data as single and create a `signature list` later on if required. Otherwise, this binding logic gets inconsistent

packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -61,7 +61,8 @@ export async function updateAttachmentColumns(prodAppId: string, db: Database) {
         for (let column of columns) {
           const columnType = table.schema[column].type
           if (
-            columnType === FieldType.ATTACHMENTS &&
+            (columnType === FieldType.ATTACHMENTS ||
+              columnType === FieldType.SIGNATURE) &&
             Array.isArray(row[column])
```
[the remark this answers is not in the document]
deanhannigan: I've refactored the signature field to bring it in line with the attachment and user columns

packages/frontend-core/src/components/grid/cells/SignatureCell.svelte
```diff
@@ -0,0 +1,172 @@
...
+
+<SignatureModal
+  onConfirm={saveSignature}
+  title={schema?.name}
+  {value}
+  darkMode={$props.darkMode}
+  bind:this={modal}
+/>
+
+{#if isOpen}
```
aptkingston, line 91: I think this has probably changed since you originally wrote this, but we now use a `GridPopover` component for any popovers like this. It was an important change to allow overflow going outside the grid.

Updating this is pretty straightforward thankfully - take a look at the attachment cell for an example. Basically just copy that exact usage of the GridPopover and put this stuff inside it (adjusting styles as needed).

aptkingston (approved):
One small suggestion to update the grid cell to use the latest components, but otherwise LGTM! This is so smooth to write on, I love it 😂

mike12345567:
Gave this a go myself, Ubuntu + Firefox, all seems to be working well with my graphics tablet
image

Takes a little bit of getting used to due to it not starting immediately where you click, but works nicely!

aptkingston:
Bit of a random one, but if just click (don't drag) you get lines shooting in from the top left... Not a massive issue lol unless there's an easy fix.
image

deanhannigan:
@aptkingston there appears to be a recent issue around this on the framework itself https://github.com/jakubfiala/atrament/issues/99

Switching to the version from 3 weeks ago might fix the issue.

adrinr (commented):
Still being broken for me using firefox on a mac. Happy to debug today to see if we can reproduce it

deanhannigan:
Had a call with @adrinr this morning to run through this and the error he's seeing is definitely happening. It seems that it only occurs when the component is loaded on a native/laptop monitor 🙃. The canvas get's resized to 800x400 irrespective of the configuration. This was reproducible in Firefox **and** Chrome. Reviewing the issue now

deanhannigan:
@aptkingston @mike12345567 added a fix for the issue around clicking on the canvas. The issue manifests itself if you click in exactly same area without moving. I added a check for where the last draw ended and where the pointer is on click. If they match exactly, the draw behaviour will be canned and the odd lines wont appear.

aptkingston:
> @aptkingston @mike12345567 added a fix for the issue around clicking on the canvas. The issue manifests itself if you click in exactly same area without moving. I added a check for where the last draw ended and where the pointer is on click. If they match exactly, the draw behaviour will be canned and the odd lines wont appear.

Ah I see - I think the common manifestation of this is then if you have *never* clicked yet, it treats it as if you've clicked in the same area since your reference point is null.

PR #13424, opened 2024-04-08
Correct types accepted / returned through the view API.

samwho:
## Description

While working on https://github.com/Budibase/budibase/pull/13419 I ran into a problem caused by the `ViewV2` type not quite being correct. It's being used to represent both the request type and response type from various endpoints, but it's only correct for requests. Responses are sometimes "enriched" and have a wider variety of stuff in them.

This PR fixes that.


mike12345567 (approved):
LGTM!

Issue #13421, opened 2024-04-08
Allow the ability to reference the selected rows for the new table


ConorWebb96:
**Describe the feature request**
Expose the rows that are selected so they can be used for other processes e.g. sending off to an automation or exported as a CSV/JSON.

PR #13426, opened 2024-04-08
Selected row + export data support for grid blocks (new table component)

aptkingston:
## Description
This PR adds support for hooking into the selected rows of the new table component (grid block).

- There's a new binding provided by tables for "Selected rows" which are the full selected row objects. This works for any sort of datasource - a big improvement over the old table component. You can then use this as you see fit.
image

- Support has been added for the "Export data" button action. No config changes needed - you can just select new table components as well as old ones. Behind the scenes I've moved away from the deprecated "row selection store" but there's no difference to the user. Exporting data still only works for table datasources, as always.
image

## Other stuff
When blurring a grid (clicking outside it) we no longer deselect any selected rows. This applies across the board. Hovered and focused cell will still be unset.

## Addresses
- https://linear.app/budibase/issue/BUDI-8151/allow-the-ability-to-reference-the-selected-rows-for-the-new-table
- https://github.com/Budibase/budibase/issues/13421

## Feature branch env
[Feature Branch Link](http://fb-new-table-selected-rows.fb.qa.budibase.net)


ConorWebb96 (approved):
LGTM, there are a few issues with this feature but they don't seem to be related to this PR. I'll log theses issues separately after we get the export functionality for the grid in.

PR #13427, opened 2024-04-09
Quiet option for grid blocks (new table component)

aptkingston:
## Description
This was requested and was very easy to add, so decided to quickly throw it up. This adds a new "Quiet" setting for tables which simply removes the horizontal borders between cells, other than the sticky column which retains its border.

New setting:
image

You'll need to enlarge these images to see the difference properly.

Normal table:
image

Quiet table:
image

## Launchcontrol
Quiet option for tables.


deanhannigan (approved):
LGTM!

PR #13431, opened 2024-04-09
Builder bug fixes

deanhannigan:
## Description
Miscellaneous bug fixes.

## Addresses
- `Bug` - Columns not appearing in settings when specifying a `Data Provider` source
  - For `provider` types,  columns do not appear.
- `Bug` - Height not working for `Table` comp
  - Reduced the min height to `120px` when specifying a custom value. Was locked at `410px` even with a custom value
- `Bug` - Row explorer block into repeater block, crashes builder. 
  - Added `rowexplorer` to the `illegalChildren` config in the manifest. This combination would kill the builder and make recovery very difficult.

## Launchcontrol
Minor bug fixes and UX updates.


Ghrehh (approved):
LGTM

packages/client/src/components/app/GridBlock.svelte
```diff
@@ -147,6 +149,10 @@
     border: 1px solid var(--spectrum-global-color-gray-300);
     border-radius: 4px;
     overflow: hidden;
+    min-height: 120px;
```
aptkingston, line 152: The large min height was deliberate to ensure all types of popovers can comfortable fit inside. I'm happy to shrink it a little if we need to (although I haven't seen anyone complain about this yet), but I wouldn't want to go lower than 230px (which is enough for 5 rows).

The easier way to specify this is just
```css
    min-height: 230px;
    height: 410px;
``
by the way - no need for extra classes etc. The styleable helper will override the `height` setting if they use it.

PR #13432, opened 2024-04-09
[BUDI-7659] First end-to-end test for SQS functionality.

samwho:
## Description

This PR introduces `search.spec.ts`, which currently houses a single test: a search that returns all rows. The thing that's special about it is that it has all of the plumbing in place to execute searches using both Lucene and SQS, as well as external data sources.

The plan is to build out the test suite in future PRs, moving the few search tests from `row.spec.ts` into this file as well as greatly expanding the parts of search that are tested.


mike12345567 (approved):
Massive - great to start testing actual SQS searching!

PR #13435, opened 2024-04-09
Add more search tests.

samwho:
## Description

This adds tests for more variants of search queries against string fields.

The commented add tests fail variously across all datasources, will need to investigate that tomorrow.

### DIFF
--- a/packages/server/src/api/routes/tests/search.spec.ts
+++ b/packages/server/src/api/routes/tests/search.spec.ts
@@ -2,13 +2,7 @@ import { tableForDatasource } from "../../../tests/utilities/structures"
 import { DatabaseName, getDatasource } from "../../../integrations/tests/utils"
 
 import * as setup from "./utilities"
-import {
-  Datasource,
-  FieldType,
-  SearchFilter,
-  SearchFilters,
-  Table,
-} from "@budibase/types"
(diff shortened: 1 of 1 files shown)


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -53,22 +53,41 @@ describe.each([
...
-    ])
+  describe("strings", () => {
+    interface StringSearchTest {
+      query: SearchFilters
+      expected: number[]
+    }
+
+    const stringSearchTests: StringSearchTest[] = [
+      { query: {}, expected: [0, 1] },
+      { query: { string: { name: "foo" } }, expected: [0] },
```
adrinr, line 64: This was not really clear on first read. Could we rename it to expectedRowIdx or similar?

mike12345567 (approved):
Pre-approving this, but agree with @adrinr comment, might be helpful to explain the array of numbers as row indexes.

samwho:
I've gone a slightly different direction, because I didn't really like the indexes at all to be honest. How is it now?

mike12345567 (approved):
Very nice - was wondering if this would be possible, feels easier to just specify what you expect back!

PR #13444, opened 2024-04-10
Unify icons usage in builder

adrinr:
## Description
We have some duplicated icon logic between column configuration icons and header icons. These changes unify this logic, making sure they are always in sync.

Fixing inconsistencies between grid header and the column config
1. Date time: unified to using the full calendar, used in the column confi
image

2. Attachment: unified to use the grid header one
image

3. Multiselect: unified to use the grid header one
image

4. Relationship: unified to use the grid header one
image


aptkingston (approved):
LGTM!

PR #13445, opened 2024-04-10
Add number and date search tests.

samwho:
## Description

Continuing the work to create a comprehensive test suite for our internal search API, this PR adds tests for searching number and date columns.

### DIFF
--- a/packages/server/src/api/routes/tests/search.spec.ts
+++ b/packages/server/src/api/routes/tests/search.spec.ts
@@ -2,7 +2,13 @@ import { tableForDatasource } from "../../../tests/utilities/structures"
 import { DatabaseName, getDatasource } from "../../../integrations/tests/utils"
 
 import * as setup from "./utilities"
-import { Datasource, FieldType, SearchFilters, Table } from "@budibase/types"
+import {
+  Datasource,
+  EmptyFilterOption,
+  FieldType,
+  SearchFilters,
+  Table,
+} from "@budibase/types"
 
 jest.unmock("mssql")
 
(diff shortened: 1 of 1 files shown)


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -76,13 +91,190 @@ describe.each([
...
+        })
+        expect(foundRows).toEqual(
+          expect.arrayContaining(
+            expected.map(r =>
+              expect.objectContaining(savedRows.find(sr => sr.name === r.name)!)
+            )
+          )
+        )
+      }
+    )
```
samwho, line 109: Changed the body of each of these functions to assert that the result contains the body of the _saved_ row, not the bare `{ name: "whatever" }`. Slightly more robust, and in the case of dates lower down they're saved as `Date` objects but return as ISO `string`s. So this gives some resilience against conversions, which I'm sure will come up with some other data types.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -76,13 +91,190 @@ describe.each([
...
+      },
+    ]
+
+    it.each(dateSearchTests)(
+      `should be able to run query: $query`,
+      async ({ query, expected }) => {
+        // TODO(samwho): most of these work for SQS, but not all. Fix 'em.
+        if (isSqs) {
+          return
+        }
```
samwho, line 262: I initially tried to `!isSqs && it.each(...` a few lines up but then Jest complains about having a `beforeEach` in a `describe` with no tests, so this felt slightly less icky than using `!isSqs` in 2 places.

adrinr (approved):
LGTM!

PR #13447, opened 2024-04-10
Improving public API documentation for row search

mike12345567:
## Description
When looking through the search parameters to build up information about validating search inputs better on the API (https://linear.app/budibase/issue/BUDI-8165/validate-search-api-parameters-based-on-table-schema) found the public docs weren't fully up to date - fixing this.

Also generating to include the new attachment single type.

### DIFF
--- a/packages/server/specs/openapi.json
+++ b/packages/server/specs/openapi.json
@@ -1792,7 +1792,7 @@
               },
               "contains": {
                 "type": "object",
-                "description": "Searches for a value, or set of values in an array column types (such as a multi-select), if an array of search options is provided then it must match all.",
+                "description": "Searches for a value, or set of values in array column types (such as a multi-select). If an array of search options is provided then it must match all.",
                 "example": {
                   "arrayColumn": [
                     "a",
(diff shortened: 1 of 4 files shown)


packages/server/specs/openapi.json
```diff
@@ -1786,6 +1789,36 @@
               "oneOf": {
                 "type": "object",
                 "description": "Searches for rows which have a column value that is any of the specified values. The format of this must be columnName -> [value1, value2]."
+              },
+              "contains": {
+                "type": "object",
+                "description": "Searches for a value, or set of values in an array column types (such as a multi-select), if an array of search options is provided then it must match all.",
```
samwho, line 1795:
```suggestion
                "description": "Searches for a value, or set of values in array column types (such as a multi-select). If an array of search options is provided then it must match all.",
```

packages/server/specs/openapi.json
```diff
@@ -1786,6 +1789,36 @@
...
+                "example": {
+                  "arrayColumn": [
+                    "a",
+                    "b"
+                  ]
+                }
+              },
+              "notContains": {
+                "type": "object",
+                "description": "As with the contains search, only functions for array column types, but searches for columns missing the supplied values.",
```
samwho, line 1805:
```suggestion
                "description": "The logical inverse of `contains`. Only works on array column types. If an array of values is passed, the row must not match any of them to be returned in the response.",
```

samwho (approved):
Left a couple of suggestions. I'm guessing there's a source of truth for these docs and the rest are generated from it, but if not, comments apply to all locations 😄

mike12345567:
@samwho OpenAPI spec is based on the `packages/server/specs/resources` typescript files, which define the structure, which is then generated using the `yarn specs` command in the server package.

Once the OpenAPI spec is merged in Readme will re-sync with the spec and update.

PR #13448, opened 2024-04-10
Console.error lint rule

shogunpurple:
## Description
OIDC logging was not occurring because we were only doing `console.log(err)`, which caused pino to skip it. Originally put a fix in to support `console.error(err)`, but decided against it for a few reasons:
- we essentially have this nowhere else in the codebase.
- There are times we may want to log an error in the pod but not to DD
- It's more explicit to add a message before you pass the error object. It's easier to grep, and clearer by reading the code instead of having to debug at runtime.


samwho (approved):

adrinr (approved):
Nice catch! There are still some linting issues left

PR #13450, opened 2024-04-10
Get date search tests passing for SQS.

samwho:
## Description

This PR uncomments the date search tests for SQS and gets them passing.

Most of the heavy lifting to get these to work was done in https://github.com/Budibase/budibase/pull/13449.


mike12345567 (approved):
Looking great!

PR #13456, opened 2024-04-11
Assert length on search tests, fixes bug in SQS around `onEmptyReturn`

samwho:
## Description

The `search.spec.ts` tests weren't correctly asserting when a result set should be empty, so that PR fixes that. It pointed out a problem with SQS where if you set `onEmptyReturn: "none"`, it would actually return all rows.

On looking at the code, I figured this logic could live at the top level of the SDK. So I removed the logic from the `ExternalRequest` class and put it in the SDK. All of the tests pass now.

This PR also reworks how the requests create test data, bringing the runtime on my local machine down from 25 seconds to 14 seconds.


mike12345567 (approved):
Very nice find! This makes more sense in the SDK than part of the external system as well.

PR #13458, opened 2024-04-11
Fix issue where automation meta fields were being overridden

PClmnt:
## Description
The meta fields within an automation object were being cleared thanks to the loop substitution 

## Addresses
https://linear.app/budibase/issue/BUDI-8118/clear-relationships-if-empty-flag-doesnt-work-in-update-row-automation

## Launchcontrol
- Fixes an issue where the Clear Relationship option within automations wasn't being respected when using a loop block.


samwho (approved):
I'm not going to block on this, but is there a way to test this so it doesn't regress?

PClmnt:
> I'm not going to block on this, but is there a way to test this so it doesn't regress?

I'm not sure, the fix is in the the automation thread itself, and for the tests we do have around automations we run the "step" directly, not through the thread. Maybe @mike12345567 might have some ideas?

mike12345567:
> > I'm not going to block on this, but is there a way to test this so it doesn't regress?
> 
> I'm not sure, the fix is in the the automation thread itself, and for the tests we do have around automations we run the "step" directly, not through the thread. Maybe @mike12345567 might have some ideas?

We have some tests under https://github.com/Budibase/budibase/blob/master/packages/server/src/api/routes/tests/automation.spec.ts which we could extend, although its definitely an area where we could do with expansion.

PClmnt:
> I'm not sure, the fix is in the the automation thread itself, and for the tests we do have around automations we run the "step" directly, not through the thread. Maybe @mike12345567 might have some ideas?

Issue being that within those tests the thread is actually mocked, I think that actually testing the entire thread code is a reasonably big task, am I okay to merge this without it?

mike12345567 (approved):
Yeap fine to merge - we should probably create some tickets for creating some real test cases in this area.

PR #13459, opened 2024-04-11
Reduce duplication in search.spec.ts

samwho:
## Description

The search tests were starting to get quite repetitive, and I was noticing that the name of each test wasn't always very helpful:

CleanShot 2024-04-11 at 11 59 37@2x

So this PR attempts to reduce duplication as well as making the test names more descriptive (though still messier than I'm happy with):

CleanShot 2024-04-11 at 12 00 11@2x


mike12345567 (approved):
Very nice! Big improvement to readability!

PR #13463, opened 2024-04-11
adds sidepanel open and close actions, and gives the user the option to disable click-outside closure of sidepanel

### LINKED ISSUE #13429: Configurable actions on Side Panel open/close
Currently, clicking anywhere outside of the side-panel will close the side-panel. Side-panels are often used to contain forms that edit or update data, and clicking out of the side-panel results in the immediate loss of any unsaved changes.

In the Side Panel settings there should be an option to create actions on-open and on-close, and a toggle that allows the developer to decide if clicking out of the side-panel will close the side-panel.

Having the option to run actions on-close would give developers a chance to prompt the user to complete their form.

This feature request stems from an exploration of [Discussion #13401](https://github.com/Budibase/budibase/discussions/13401)

mikesealey:
## Description
This change allows users to set actions on side panel open and side panel close, and toggles wether or not the side panel will close when a user clicks away, outside of the side panel.

## Addresses
Issue: Configurable actions on Side Panel open/close https://github.com/Budibase/budibase/issues/13429
Discussion: Prompt to close side panel is clicking outside the panel https://github.com/Budibase/budibase/discussions/13401

## Screenshots
[Link to screen recording](https://jam.dev/c/057e774e-5c5d-4f32-9d04-663ef9e7db20)

## Launchcontrol
This change allows Budibase developers to set actions on side panel open and on side panel close.

Developers can now toggle wether or not the side panel will close when a user clicks outside of the side panel. An example use-case for this would be preventing the loss of partially-completed forms in the side panel due to mis-click.


mikesealey:

melohagan (changes requested):
Looks good! It's been a requested feature for a while, so a nice addition 😄 

My only main comment is about whether or not we need a *Side Panel Opened* event, because a side panel can never be opened without a Budibase action in the first place. I'm sure others would have some thoughts about this.

The *Side Panel Closed* event definitely makes sense to me though!

packages/client/manifest.json
```diff
@@ -6723,7 +6723,22 @@
...
+        "type": "boolean",
+        "key": "clickOutsideToClose",
+        "label": "Click outside to close",
+        "defaultValue": true
+      },
+      {
+        "type": "event",
+        "key": "sidePanelClose",
+        "label": "Side Panel Close",
+        "info": "Side panel actions configured here will run after the 'Open side panel' action runs, and are not capable of preventing or stopping it. Any form validation should therefore be done before that action is invoked, and not here."
```
melohagan, line 6739: NAB: This text is a bit wordy/technical. I'm not sure if users would broadly understand the term *invoke* for example.

> Side panel actions configured here will run after the 'Open side panel' action runs, and are not capable of preventing or stopping it.

Is this such a common scenario that it needs to be explained. Would a paragraph in the docs be a more appropriate place for this?

packages/client/manifest.json
```diff
@@ -6723,7 +6723,22 @@
...
+      {
+        "type": "boolean",
+        "key": "clickOutsideToClose",
+        "label": "Click outside to close",
+        "defaultValue": true
+      },
+      {
+        "type": "event",
+        "key": "sidePanelClose",
+        "label": "Side Panel Close",
```
melohagan, line 6738: As this is an event, I think it should read: `On side panel close` to be consistent with the other events such as *On click* and *On screen load*

melohagan (approved):
LGTM! Just left a couple of nitpicks around some wording. Great job! 🚀

PR #13469, opened 2024-04-11
Add tests for sorting to search.spec.ts

samwho:
## Description

Alongside new tests for sorted queries, this PR also changes the names of some of the custom matchers to better reflect what they do:

- `toFind` -> `toContains`, because it does not require strict ordering.
- new: `toMatch`, requires ordering but does not do a full object match against the saved row (same as the old behaviour)
- new: `toHaveLength` for the limit check


mike12345567 (approved):
👏

Issue #12483, opened 2023-11-30
[BUDI-7789] Increase services.objectStore.storage default value to 2GB


likke812:
**Checklist**

* [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: k8s
  * Budibase Version: 2.13.20
  * App Version: 2.10.1

**Describe the bug**
After we upgraded Budibase platform using K8S successfully, we can not update Budibase APP from 2.10.1 to 2.13.10 with popup error as below:
`Error updating app: [object Object]`

And there has HTTP 507 error when checking browser inspector:
`POST https://***/api/applications/app_dev_***/client/update 507 (Insufficient Storage)`
`[Builder] HTTP 507 on POST:https://***/api/applications/app_dev_***/client/update `
`Storage backend has reached its minimum free drive threshold. Please delete a few objects to proceed.`

**To Reproduce**

1. Go to 'APP -> Update'
2. Click on 'Update' button and confirm the application update button again
3. See error

**Expected behavior**
Update Budibase application to 2.13.20 successfully

**Screenshots**
Error snapshot:

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/4f9934ac-448e-4baf-a66b-8604166658de/abb94806-6ef4-4c8f-8166-154a65709ae8?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC80Zjk5MzRhYy00NDhlLTRiYWYtYTY2Yi04NjA0MTY2NjU4ZGUvYWJiOTQ4MDYtNmVmNC00YzhmLTgxNjYtMTU0YTY1NzA5YWU4IiwiaWF0IjoxNzAxOTM4MTYwLCJleHAiOjE3MDIwMjQ1NjB9.1mgWktJMAy8NWVvzI3zz532I6b9JLBImMBZlT1ZV1Vo)

**App Export**

**Desktop (please complete the following information):**

* OS: Windows 10
* Browser: Chrome and Firefox
* Version: 119 and 120

**Additional context**
The df output of all of the three DBs:

[...]

PR #13472, opened 2024-04-12
Increase the base Minio storage to 2Gi in Helm chart.

samwho:
## Description

Increases the default Minio storage from 100Mi to 2Gi for Helm chart users. @likke812 pointed out to us that the default simply doesn't work as it's too low. Given that it doesn't work at all, I'm not worried about changing this default as anyone using this chart for real has almost certainly set their own value or isn't using Minio.

## Addresses
- https://github.com/Budibase/budibase/issues/12483

## Launchcontrol

Increases the base storage for Minio to 2Gi in our Helm chart. @likke812 made us aware that the default of 100Mi is too low, and Minio fails to start as a result. Thank you! 🙏

### DIFF
--- a/charts/budibase/values.yaml
+++ b/charts/budibase/values.yaml
@@ -488,7 +488,7 @@ services:
     # do this.
     url: "http://minio-service:9000"
     # -- How much storage to give Minio in its PersistentVolumeClaim.
-    storage: 100Mi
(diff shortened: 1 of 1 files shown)


melohagan (approved):
LGTM!

PR #13479, opened 2024-04-12
Don't delete attachments directly from the frontend

adrinr:
## Description
Nowadays, when deleting an attachment via the Dropzone component, the attachment gets deleted via an API endpoint. This might cause the following issues:
1. The form has not been saved, but the attachment has been deleted
2. The attachment was deleted, but it's still used in the production app
3. There is no easy way to check that the user has access to delete that attachment, as there is no row security metadata or similar in the object.

When updating a row, we have a certain logic that checks if an attachment needs to/can be deleted. Because of this, we can rely on the update endpoint.

Note that the upload will work as it is now. We might have some cases where data has been saved to the storage but not persisted as a row, leaving some orphan blobs. This would require a harder refactor and there is not real benefit other than saving storage space. If required, we could even have a background job that detects these orphan files.

## Addresses
- [BUDI-8143 - Don't delete attachments directly from the frontend](https://linear.app/budibase/issue/BUDI-8143/dont-delete-attachments-directly-from-the-frontend)

## App Export
[BUDI 8143-export-1712921680648.tar.gz](https://github.com/Budibase/budibase/files/14958159/BUDI.8143-export-1712921680648.tar.gz)

## Launchcontrol
Fixing bug when deleting attachments from a row when it's not meant to


mike12345567 (approved):
image
A perfect PR, the best ratio possible 😆 
LGTM!

PR #13483, opened 2024-04-12
SQS relationship support

mike12345567:
## Description
This PR adds in support for SQS relationships - a migration is still required to the link documents, but given a new app this will allow the API to index relationships the same way we do for external databases. This involves creation of the junction tables for the relationships which we can then use for indexing rows.

There is still work needed in this area - this is just the first pass and still need to test/work around multiple relationships between the two same tables, this is not currently managed (although should be achievable just through search).

Also adds in support for aliasing of queries which improves the readability of the SQS queries significantly - had to move aliasing around a bit to make it usable by SQS but otherwise very few changes were required.

Lastly, also updates the search SDK a little to move the search input mapping up to a generic level as it wasn't being performed for SQS search, rather than having it implemented in each search independently added table retrieval to the SDK, passing the search table down into the APIs, which gives us the context to map the input search properly. Also this change will be needed to carry out the search validation work: https://linear.app/budibase/issue/BUDI-8165/validate-search-api-parameters-based-on-table-schema


packages/server/src/sdk/app/rows/sqlAlias.ts
```diff
@@ -156,12 +157,19 @@ export default class AliasTables {
...
-    json: QueryJson
+    json: QueryJson,
+    queryFn?: (json: QueryJson) => Promise<DatasourcePlusQueryResponse>
   ): Promise<DatasourcePlusQueryResponse> {
     const datasourceId = json.endpoint.datasourceId
-    const datasource = await sdk.datasources.get(datasourceId)
+    const isSqs = datasourceId === SQS_DATASOURCE_INTERNAL
+    let aliasingEnabled: boolean, datasource: Datasource | undefined
+    if (isSqs) {
+      aliasingEnabled = true
```
samwho, line 167: Should this respect the value of the environment variable that disables aliasing?

samwho (approved):
Just the one question. I'm leaning heavily here on the tests having good coverage over existing functionality to know that this is unlikely to break anything today, and that we will be writing tests for this functionality in the very near future which will shake out any bugs.

packages/server/src/sdk/app/rows/sqlAlias.ts
mike12345567: I'm not sure about this one - the env variable is mostly for allowing users to disable it if they encounter a problem, with SQS I don't know if we want there to be two internal flows.

But also it could be quite useful for ourselves to have the ability to turn it on and off in the Cloud if we think it was causing a problem. I'll add it in just incase.

samwho: Don't mind which direction you go with it but I follow your reasoning. I'd probably err on the side of leaving it until we have a use-case, but I'm not strongly either way.

PR #13484, opened 2024-04-12
Rename `search.spec.ts` helpers to better reflect their purpose.

samwho:
## Description

This was bothering me as I worked through some new tests so I split it out into its own PR. `toContains` implies extra rows would be fine, but that's not what the assertion does. I've changed it to `toContainsExactly` to reflect that additional rows will cause a test failure. I've also introduced a `toContains` that works as expected.


mike12345567 (approved):
LGTM!

PR #13488, opened 2024-04-15
Adds sidepanel open and close actions, and gives the user the option to disable click-outside closure of sidepanel

mike12345567:
Re-introduces https://github.com/Budibase/budibase/pull/13463

PR opened for @aptkingston to comment on, original creator @mikesealey


packages/client/src/components/app/SidePanel.svelte
```diff
@@ -26,6 +29,10 @@
     }
   }
 
+  $: {
+    sidePanelStore.actions.setSidepanelState(clickOutsideToClose)
```
aptkingston, line 33: We can't use this store to store a single boolean here, because each side panel has its own setting for whether it can be clicked outside to close or not. By using a single store, multiple side panels will override each other and the last to render will "win" in case where we have multiple side panels (which we do have - like the table autoscreen).

One solution would be only setting this value when a certain side panel actually opens - because all we care about is handling the correct behaviour for the side panel that is currently open, and only one can be open at any given time.

The way you could do this is move this line to be inside the
```js
if (open) {
...
}
```
statement down below.

This way the store is updated to reflect the desired behaviour whenever any side panel is opened.

packages/client/manifest.json
```diff
@@ -6723,7 +6723,21 @@
...
     "draggable": false,
-    "info": "Side panels are hidden by default. They will only be revealed when triggered by the 'Open Side Panel' action."
+    "info": "Side panels are hidden by default. They will only be revealed when triggered by the 'Open Side Panel' action.",
+    "sendEvents": true,
+    "settings": [
+      {
+        "type": "boolean",
+        "key": "clickOutsideToClose",
+        "label": "Click outside to close",
+        "defaultValue": true
```
aptkingston, line 6733: We can't default this to true, because all existing apps won't have this value set. For backwards compatibility we need to assume that the default value of any new settings are falsey, and invert the logic.

In this case, this new setting would have to something like "Ignore clicks outside" with a default value of false. This means all old apps will continue to work as normal and the new behaviour is opt in.

packages/client/manifest.json
```diff
@@ -6723,7 +6723,21 @@
...
+      },
+      {
+        "type": "event",
+        "key": "onSidePanelClose",
+        "label": "On side panel close"
```
aptkingston, line 6738: NAB: We should rename this to just "On close" to keep it in one line.

packages/client/src/stores/sidePanel.js
```diff
@@ -32,11 +33,19 @@ export const createSidePanelStore = () => {
     }, 50)
   }
 
+  const setSidepanelState = bool => {
```
aptkingston, line 36: We should use a more descriptive name here - `setSidepanelState` is very generic and doesn't describe what it's actually doing. I'd recommend something like `setIgnoreClicksOutside` (if you follow my suggestion for how to rename the setting, otherwise adapt appropriately).

packages/client/src/stores/sidePanel.js
```diff
@@ -32,11 +33,19 @@ export const createSidePanelStore = () => {
...
+    clearTimeout(timeout)
```
aptkingston, line 37: We don't need to clear the timeout here - that's specifically for improving the opening/closing behaviour.

packages/client/manifest.json
```diff
@@ -6723,7 +6723,21 @@
     "illegalChildren": ["section", "sidepanel"],
     "showEmptyState": false,
     "draggable": false,
-    "info": "Side panels are hidden by default. They will only be revealed when triggered by the 'Open Side Panel' action."
+    "info": "Side panels are hidden by default. They will only be revealed when triggered by the 'Open Side Panel' action.",
+    "sendEvents": true,
```
aptkingston, line 6727: Just confirming this is intentional. `sendEvents` will mean that we will send posthost events whenever this setting is changed by any user. We typically use it to track usage of certain things to improve our default behaviour. Just letting you know that in case this wasn't intentional.

packages/client/manifest.json
```diff
@@ -6723,7 +6723,21 @@
...
+    "settings": [
+      {
+        "type": "boolean",
+        "key": "clickOutsideToClose",
+        "label": "Click outside to close",
+        "defaultValue": true
+      },
+      {
+        "type": "event",
+        "key": "onSidePanelClose",
```
aptkingston, line 6737: NAB: This can just be called `onClose`. This is already in the context of the side panel, so we don't need to repeat the name "side panel" inside the setting.

packages/client/src/components/app/SidePanel.svelte
```diff
@@ -51,6 +64,7 @@
       } else {
         if (target.contains(node)) {
           target.removeChild(node)
+          handleSidePanelClose()
```
aptkingston, line 67: I'm not sure what the default svelte behaviour is here, but I would just double check that this won't invoke the close callback immediately on first render, when the side panel is closed by default.

aptkingston (changes requested):
Well done on your first contribution @mikesealey! I've tried to be reasonably verbose here to help explain the changes I've suggested.

Feel free to ask away with any questions if anything isn't clear :)

mikesealey:
[Link to screen recording demonstration](https://jam.dev/c/05d6de96-3da0-4628-93a6-8afd30a4369b)

aptkingston (approved):
LGTM (looks good to me)! Nice work @mikesealey. 🚀

PR #13489, opened 2024-04-15
Commenting `FieldType` enumeration

mike12345567:
## Description
Commenting the field type enumeration to better explain what all of the types do and how they are represented within Budibase.

### DIFF
--- a/packages/types/src/documents/app/row.ts
+++ b/packages/types/src/documents/app/row.ts
@@ -1,76 +1,110 @@
 import { Document } from "../document"
 
 export enum FieldType {
-  // a primitive type, stores a string, called Text within Budibase. This is one of the default
-  // types of Budibase, if an external type is not fully understood, we will treat it as text.
+  /**
+   * a primitive type, stores a string, called Text within Budibase. This is one of the default
+   * types of Budibase, if an external type is not fully understood, we will treat it as text.
+   */
   STRING = "string",
-  // similar to string type, called Long Form Text within Budibase. This is mainly a frontend
-  // orientated type which disables a larger text input area. This can also be used
-  // in conjunction with the 'useRichText' option to support a markdown editor/viewer.
+  /**
+   * similar to string type, called Long Form Text within Budibase. This is mainly a frontend
+   * orientated type which enables a larger text input area. This can also be used
+   * in conjunction with the 'useRichText' option to support a markdown editor/viewer.
+   */
   LONGFORM = "longform",
-  // similar to string type, called Options within Budibase. This works very similarly to
-  // the string type within the backend, but is validated to a list of options. This will
-  // be displayed a select input within the builder/client.
+  /**
+   * similar to string type, called Options within Budibase. This works very similarly to
+   * the string type within the backend, but is validated to a list of options. This will
+   * display a <select> input within the builder/client.
+   */
   OPTIONS = "options",
-  // a primitive type, stores a number, as a floating point, called Number within Budibase.
-  // this type will always represent numbers as reals/floating point - there is no integer only
-  // type within Budibase.
+  /**
+   * a primitive type, stores a number, as a floating point, called Number within Budibase.
+   * this type will always represent numbers as reals/floating point - there is no integer only
+   * type within Budibase.
+   */
   NUMBER = "number",
-  // a primitive type, stores a boolean, called Boolean within Budibase. This is often represented
-  // as a toggle or checkbox within forms/grids.
+  /**
+   * a primitive type, stores a boolean, called Boolean within Budibase. This is often represented
+   * as a toggle or checkbox within forms/grids.
+   */
   BOOLEAN = "boolean",
-  // a JSON type, this type is always an array of strings, called Multi-select within Budibase.
-  // This type can be compared to the options type, as it functions similarly, but allows picking
-  // multiple options rather than a single option.
+  /**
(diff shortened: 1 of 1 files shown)


packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
 import { Document } from "../document"
 
 export enum FieldType {
+  // a primitive type, stores a string, called Text within Budibase. This is one of the default
+  // types of Budibase, if an external type is not fully understood, we will treat it as text.
   STRING = "string",
+  // similar to string type, called Long Form Text within Budibase. This is mainly a frontend
+  // orientated type which disables a larger text input area. This can also be used
```
samwho, line 8:
```suggestion
  // orientated type which enables a larger text input area. This can also be used
```

packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
...
+  // in conjunction with the 'useRichText' option to support a markdown editor/viewer.
   LONGFORM = "longform",
+  // similar to string type, called Options within Budibase. This works very similarly to
+  // the string type within the backend, but is validated to a list of options. This will
+  // be displayed a select input within the builder/client.
```
samwho, line 13:
```suggestion
  // display a <select> input within the builder/client.
```

packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
...
+  // this type will always represent numbers as reals/floating point - there is no integer only
+  // type within Budibase.
   NUMBER = "number",
+  // a primitive type, stores a boolean, called Boolean within Budibase. This is often represented
+  // as a toggle or checkbox within forms/grids.
   BOOLEAN = "boolean",
+  // a JSON type, this type is always an array of strings, called Multi-select within Budibase.
+  // This type can be compared to the options type, as it functions similarly, but allows picking
+  // multiple options rather than a single option.
   ARRAY = "array",
```
samwho, line 25: If this is a JSON type does that mean it is represented as a string column in some data sources?

packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
...
+  // a complex type, called Relationships within Budibase. This is the most complex type of Budibase,
+  // nothing should be stored against rows under link columns; this type simply represents the
+  // relationship between tables as part of the table schema. When rows are input to the Budibase API
+  // relationships to be made are represented as a list of row IDs to link. When rows are returned
+  // from the Budibase API it will contain a list of row IDs and display column values of the related rows.
   LINK = "link",
+  // a complex type, called Formulas within Budibase. This type has two variants, static and dynamic, with
+  // static only being supported against internal tables. Dynamic formulas calculate a provided HBS/JS binding
+  // based on the row context and enrich it when rows are being returned from the API. Static bindings calculate
+  // this when rows are being stored, so that the formula output can be searched upon within the DB.
```
samwho, line 47: Is there not also a variant where this is either calculated on the frontend or in the backend?

packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
...
   FORMULA = "formula",
+  // a complex type, called Auto Column within Budibase. This type has a few variants, with options such as a
+  // date for created at/updated at, an auto ID column with auto-increments as rows are saved and a user
+  // relationship type which stores the created by/updated by user details. This sub-types all depend on the
+  // date, number of link types respectively.
   AUTO = "auto",
+  // a JSON type, called JSON within Budibase. This type allows any arbitrary JSON to be input to this column
+  // type, which will be represented a string in the row. This type depends on a schema being provided to make the
```
samwho, line 55: It's not always represented as a string in the row. For external datasources that support it (e.g. Postgres) we use the native JSON column type.

samwho, line 55: It's not always represented as a string in the row. For external datasources that support it (e.g. Postgres) we use the native JSON column type.

packages/types/src/documents/app/row.ts
```diff
@@ -1,22 +1,76 @@
...
+  // JSON searchable/bindable, the JSON cannot be fully dynamic.
   JSON = "json",
+  // an internal type, this is an old deprecated type which is no longer used - still represented to note it
+  // could appear in very old tables.
```
samwho, line 59: Should we use the `@deprecated` tag here?

packages/types/src/documents/app/row.ts
mike12345567: This will always be represented as a JSON type if it is available however if there is no JSON type it will default to a string. There are two cases where this happens:
1. Couch - we supply stringified JSON to Couch, but it is represented as JSON for Lucene to search.
2. SQS - SQLite doesn't have a JSON type, so we will represent it as a string or blob here.

packages/types/src/documents/app/row.ts
mike12345567: No, formulas are *always* calculated on the backend.

samwho (approved):
Thank you for writing these 🙏

packages/types/src/documents/app/row.ts
mike12345567: ah - yeah I'm guessing its purely for the live evaluation, that is a good point - I'll add that to the type info just for reference.

Issue #13337, opened 2024-03-22
Color Picker Causes Application Freeze When Applying Conditional Formatting to Paragraph Element


Stropdasman:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: <method> docker compose
  - Budibase Version: <version> v2.22.3
  - App Version: <version> v2.22.3

**Describe the bug**
When attempting to apply conditional formatting to a paragraph element within Budibase, specifically trying to set a text color through the color picker, the application becomes unresponsive and fails to display the color selection interface. This issue occurs after replicating a conditional formatting rule for bold text and attempting to modify it to change the text color. No error messages are visible on the UI, but the JavaScript console displays a TypeError related to the substring method, indicating a potential issue with data handling or type expectation in the color picker component. This bug effectively prevents users from applying customized color-based conditional formatting to text elements.

**To Reproduce**
1. Add paragraph element
2. Add conditional formatting (e.g. bold)
3. Duplicate this formating and change it to color

**Expected behavior**
The application should not freeze and should show color picker

**Desktop (please complete the following information):**
 - OS: WINDOWS 11
 - Browser: Chrome
 - Version Versie 122.0.6261.129

ConorWebb96:
Thanks for the replication steps I can see this issue happening on 2.22.12. Heres the errors that are being thrown in the console.

<img width="660" alt="Screenshot 2024-03-26 at 14 28 18" src="https://github.com/Budibase/budibase/assets/126772285/7cdaf178-eddf-4091-951f-c9db22c6469c">

PR #13490, opened 2024-04-15
Fix crashing when swapping certain conditional UI settings

aptkingston:
## Description
This PR fixes an issue that can crash the builder when swapping between certain settings in conditional UI. The issue happened if you previously had a conditional UI rule applied, with a value, (e.g. update a boolean setting to `true`) then swap the setting to one which has an incompatible value type, e.g. a color setting which expects a string.

This would then pass the previously used value into the new setting, causing a crash.

This PR resolves the issue by wiping the value whenever you change the setting.

## Addresses
- https://linear.app/budibase/issue/BUDI-8121/color-picker-causes-application-freeze-when-applying-conditional
- https://github.com/Budibase/budibase/issues/13337

## Launchcontrol
Fix crashing when swapping certain conditional UI settings


deanhannigan (approved):
LGTM!

PR #13491, opened 2024-04-15
Add tests for array column types, fixing some bugs along the way.

samwho:
## Description

This PR adds tests in `search.spec.ts` for `FieldType.ARRAY` fields (multi-select in practice). While doing this I found several bugs and inconsistencies in external datasources that prevented some search operations against this field type.

## Launchcontrol

Fixes several bugs in searching against multi-select columns in external datasources.

### DIFF
--- a/packages/server/src/constants/index.ts
+++ b/packages/server/src/constants/index.ts
@@ -20,6 +20,7 @@ export enum FilterTypes {
   NOT_EMPTY = "notEmpty",
   CONTAINS = "contains",
   NOT_CONTAINS = "notContains",
+  CONTAINS_ANY = "containsAny",
   ONE_OF = "oneOf",
 }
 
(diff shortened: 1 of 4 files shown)


packages/server/src/constants/index.ts
```diff
@@ -20,6 +20,7 @@ export enum FilterTypes {
   NOT_EMPTY = "notEmpty",
```
mike12345567, line 20: I think there may be two of these `FilterTypes` enums - I know this expands *way* beyond what this PR is about, but I'm wondering if we should get rid of the `FilterType` and `FilterTypes` enums and add a single up to date one to types?

samwho: Yeah, we absolutely should. I'll do that in a follow-up PR if that's alright?

mike12345567 (approved):
LGTM! Just one comment about the `FilterTypes` - I didn't realise that one still existed, one of the old JS remnant days!

PR #13492, opened 2024-04-15
When sending a file attachment in email fetch it directly from our object store

PClmnt:
## Description
Previously we were fetching the signed url directly and attaching the body of the response to the nodemailer object, this was proving problematic in environments where we could not be sure of the platform URL. So updated this section of the code to fetch the object directly from the object store and attaching the readStream from it to the nodemailer object. 

This strategy also allows for a combination of fully formed links and the signed url string that gets returned for our external query connector.

### DIFF
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -535,8 +535,7 @@ export function extractBucketAndPath(
 ): { bucket: string; path: string } | null {
   const baseUrl = url.split("?")[0]
 
-  // eslint-disable-next-line no-useless-escape
-  const regex = /^\/files\/signed\/(?<bucket>[^/]+)\/(?<path>.+)$/
+  const regex = new RegExp(`^${signedFilePrefix}/(?<bucket>[^/]+)/(?<path>.+)$`)
   const match = baseUrl.match(regex)
 
   if (match && match.groups) {
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -524,3 +524,20 @@ export async function getReadStream(
   }
   return client.getObject(params).createReadStream()
 }
+
+/*
+Given a signed url like 'files/signed/tmp-files-attachments/app_123456/myfile.txt' extract
+the bucket and the path from it
+*/
+export function extractBucketAndPath(url: string): {
+  bucket: string
+  path: string
+} {
+  const baseUrl = url.split("?")[0]
+  const parts = baseUrl.split("/")
+  parts.shift()
+  const bucket = parts[2]
+  const path = parts.slice(3).join("/")
+
+  return { bucket, path }
+}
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -524,3 +524,20 @@ export async function getReadStream(
...
+
+/*
+Given a signed url like 'files/signed/tmp-files-attachments/app_123456/myfile.txt' extract
+the bucket and the path from it
+*/
+export function extractBucketAndPath(url: string): {
+  bucket: string
+  path: string
+} {
+  const baseUrl = url.split("?")[0]
```
mike12345567, line 536: this should probably check that it starts with the constant `files/signed` - if it doesn't then there is nothing it can extract - just to avoid any weird scenarios.

samwho: Likewise, I'd check for the total number of `/` characters as well. Really wanna make sure you're not operating on something random, because bugs like that end up being very hard to track down.

PClmnt: sorted this hopefully. This change also had the nice side effect of being able to get rid of that nasty header code in the email function. @mike12345567

mike12345567 (approved):
LGTM!

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -524,3 +525,26 @@ export async function getReadStream(
   }
   return client.getObject(params).createReadStream()
 }
+
+/*
+Given a signed url like 'files/signed/tmp-files-attachments/app_123456/myfile.txt' extract
```
adrinr, line 530: This does not really does what it says. The URL will start with `/`, but the documentation excludes it. Should we trim the initial `/` and accept both, or change the comment?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -524,3 +525,26 @@ export async function getReadStream(
...
+export function extractBucketAndPath(
+  url: string
+): { bucket: string; path: string } | null {
+  const baseUrl = url.split("?")[0]
+  if (!baseUrl.startsWith(signedFilePrefix)) {
+    return null
+  }
+
+  const parts = baseUrl.split("/")
+  if (parts.length < 5) {
```
adrinr, line 542: Could we add a comment or type this magic number?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -524,3 +525,26 @@ export async function getReadStream(
...
+    return null
+  }
+
+  const bucket = parts[3]
```
adrinr, line 546: Same here. This deconstruction seems perfect for a regex. Could we use something like this? `^\/files\/signed\/(?'bucket'[^\/]+)\/(?'path'.+)$`

packages/backend-core/src/objectStore/objectStore.ts
PClmnt: Changed comment

packages/backend-core/src/objectStore/objectStore.ts
PClmnt: Updated to regex!

adrinr (approved):
Regex power! ❤️

PR #13499, opened 2024-04-16
Refetching existing tables with override column schemas

adrinr:
## Description
Skip overriding schemas that already exist. Doing this, we are losing any column settings, such as `date_only` for datetimes and similar.

## Addresses
- [BUDI-8126 - Re fetching existing tables with ignore timezone toggle set, resets this back to off.
](https://linear.app/budibase/issue/BUDI-8126/re-fetching-existing-tables-with-ignore-timezone-toggle-set-resets)

## Launchcontrol
Fix losing column configurations


mike12345567 (approved):
LGTM!

PR #13503, opened 2024-04-16
Add `search.spec.ts` tests for `FieldType.BIGINT` and some additional range query tests.

samwho:
## Description

This PR introduces tests for searches against `FieldType.BIGINT`, and includes some fixes relating the consistency between internal and external datasources for this field type.

To be consistent we have introduced a small breaking change to the way that range searches work through the API (the Budibase UI is not affected by this change). Prior to this, the behaviour was as follows:

1. Searching for a range with both a high and low value performed an inclusive search (so bot h the high and low values would appear in the results if any rows had them as values).
2. Omitting the low or high value (so supplying only a low or high value) would perform a search using either the greater than (`>`) or less than (`<`) operators, so the search would be _exclusive_ of the values being searched against.

So if you had a table with a numeric column called `num` and you had the following rows:

```js
{ num: 1 }
{ num: 2 }
{ num: 3 }
```

You could issue the following query:

```js
{ range: { num: { low: 1, high: 3 } } }
```

To get back all of the rows. However, if you omit `low`:

```js
{ range: { num: { high: 3 } } }
```

You would only get back `{ num: 1 }` and `{ num: 2 }`. 

After this PR is merged, you will receive all 3 rows as expected.

## Launchcontrol

This PR introduces a change to make range queries performed through the [row search API](https://docs.budibase.com/reference/rowsearch) more consistent. Prior to this PR, a search with both a `low` and `high` value would perform and inclusive query where just one of `low` or `high` would be exclusive. After this PR, both variants perform inclusive searches.

This change does not affect how Budibase works through the web UI, it's only the public API that's affected.


mike12345567 (approved):
LGTM - discussed in a call the work I've carried out to fix some of the bugs that these tests unveiled, strictly around the requirement of the `meta.table` and its usage in the underlying SQL layers.

PR #13512, opened 2024-04-17
Fix changing column type

adrinr:
## Description
A couple of fixes caught during QA:
1. Changing between types not working properly for number to bool
2. Keep always the same field order, both adding a new column or editing it's type
3. Keep the schema type on refetch if compatible


mike12345567 (approved):
LGTM!

Issue #13441, opened 2024-04-10
Modal data loading - user data?


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.22.19
  - App Version: 2.22.19
  - Although this bug seems to have been there for around a week or two now

**Describe the bug**
When a modal window is opened which requires access to user data it is not loading that data.
For example if there is a create form which has a user object with a default value of `{{ Current User._id }}`, in a modal it does not set the default value, but if you open that to a full page it does set the default value.
![image](https://github.com/Budibase/budibase/assets/110813989/addba641-3855-47fc-b7b0-8fae97145afe)
![image](https://github.com/Budibase/budibase/assets/110813989/4e34e738-2dfc-4226-8454-52c1b1f864c8)

Another example I have is using the comments plugin. When it is included in a modal it shows an error getting the user first name, but full screen it works as expected.
![image](https://github.com/Budibase/budibase/assets/110813989/a9b88869-81bf-4a39-b462-0812218bfdd0)
![image](https://github.com/Budibase/budibase/assets/110813989/7d705760-1aab-4051-b5f3-cdfdab704704)

**To Reproduce**
1. Create a page and include a create form with a user field
2. Set the user field to have a default value of `{{ Current User._id }}`
3. Create a page to show that page in a modal view
4. Publish the application and open the modal view

**Expected behavior**
The user field should show the form with the default value set.

aptkingston:
Actually noticed this myself when updating the comment box plugin a few days ago. I'll take a look 👍

mike12345567:
Should be fixed in latest.

PR #13513, opened 2024-04-17
Restore render blocker around client apps until data is ready

aptkingston:
## Description
This PR adds back in the render blocker around client apps until data is ready. We always used to have this, but it was removed recently as part of the new app skeleton work to smooth out a few frames of jank from the initial load.

Unfortunately this has caused a few issues so needs to be reverted. This also has the bonus of reducing multiple wasted API calls.

## Addresses
- https://github.com/Budibase/budibase/issues/13441


deanhannigan (approved):
LGTM!

PR #13516, opened 2024-04-18
Allow a displayName to be passed when creating admin user

melohagan:
## Description
When you create an admin user, the first and last name properties are empty. Ideally when signing up with SSO, these fields should be populated.

Preview of first and last name being set for a newly created admin account via Google SSO:

Screenshot 2024-04-18 at 10 21 06

### DIFF
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -116,7 +116,7 @@ const parseBooleanParam = (param: any) => {
 export const adminUser = async (
   ctx: Ctx<CreateAdminUserRequest, CreateAdminUserResponse>
 ) => {
-  const { email, password, tenantId, ssoId } = ctx.request.body
+  const { email, password, tenantId, ssoId, displayName } = ctx.request.body
 
   if (await platform.tenants.exists(tenantId)) {
     ctx.throw(403, "Organisation already exists.")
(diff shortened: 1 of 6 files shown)


jvcalderon (approved):
LGTM

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -145,12 +145,19 @@ export const adminUser = async (
       )
     }
 
+    // Get the first and last names from the SSO display name if available
+    const nameParts = (displayName ?? "").split(" ").filter((p: string) => !!p)
+    const firstName = nameParts.length > 0 ? nameParts[0] : ""
```
adrinr, line 150: This might be conflicting. This might be the most common case, but not the only one. We will have names in multiple languages, and names such as `Mary Jane Doe`. Not sure if we should be inferring the name from the display name.
It could be that the display name is also a nickname or whatever, giving us even weirder results

adrinr: SSO can provide optional meta with the first and second name. Maybe we should check if these are populated depending on the used SSO

melohagan: This definitely seems like a better approach, I was copying some older code. There are some other places that use this code as well which I can update too.

melohagan: Updated to pass in the givenName and familyName which is provided by both Google and Microsoft SSO.

PR #13519, opened 2024-04-18
MSSQL schema conflicts

adrinr:
## Description
Knex has an issue when using `.withSchema` and table aliasing. For this, we are creating the table statement manually, reusing our existing tooling.

## Addresses
- [BUDI-8179 - [MSSQL] changing the default dbo schema to anything else makes the table unavailable when using the other schema to fetch the table](https://linear.app/budibase/issue/BUDI-8179/[mssql]-changing-the-default-dbo-schema-to-anything-else-makes-the)

## Launchcontrol
Fixing MSSQL datasources with non default schemas

### DIFF
--- a/packages/server/scripts/integrations/mssql/docker-compose.yaml
+++ b/packages/server/scripts/integrations/mssql/docker-compose.yaml
@@ -4,6 +4,7 @@ services:
   # user: sa
   # database: master
   mssql:
+    # platform: linux/amd64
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
LGTM - probably a good idea to upgrade Knex, the test cases Sam has been building will be able to confirm that this is a safe upgrade.

PR #13526, opened 2024-04-19
Fix relationships disappearing

adrinr:
## Description
We recently introduced a change where we were persisting existing schema changes to avoid dropping column settings. This caused some issues with columns non-persisted externally when fetching schema from source, such as links or formulas.

This PR is fixing this issue, adding a test that will check that we are not dropping unexpected data when synching from schema

## Addresses
- [BUDI-8181 - Relationships keep disappearing when re-fetching MySQL tables](https://linear.app/budibase/issue/BUDI-8181/relationships-keep-disappearing-when-re-fetching-mysql-tables)
- https://github.com/Budibase/budibase/issues/13523

## Launchcontrol
Fixing some links and formulas dropping when re-fetching from schema

### DIFF
--- a/packages/server/src/api/routes/tests/datasource.spec.ts
+++ b/packages/server/src/api/routes/tests/datasource.spec.ts
@@ -5,10 +5,11 @@ import { context, events } from "@budibase/backend-core"
 import sdk from "../../../sdk"
 
 import tk from "timekeeper"
-import { generator, mocks } from "@budibase/backend-core/tests"
+import { mocks } from "@budibase/backend-core/tests"
 import {
   Datasource,
   FieldSchema,
+  FieldSubtype,
   FieldType,
   QueryPreview,
   RelationshipType,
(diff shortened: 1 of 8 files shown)


packages/server/src/api/routes/tests/datasource.spec.ts
```diff
@@ -223,4 +235,152 @@ describe("/datasources", () => {
...
+          schema: {
+            name: {
+              name: "name",
+              type: FieldType.STRING,
+            },
+          },
+        })
+      )
+
+      type SupportedTypes =
```
mike12345567, line 264: Should probably name this `SupportedSqlTypes` just for clarity

mike12345567 (approved):
LGTM - amazing to finally have a good test in this area!

mike12345567 (approved):
LGTM - quite a bit cleaner and more room for custom stuff like the links logic!

PR #13527, opened 2024-04-19
Prefill card blocks

aptkingston:
## Description
This PR improves card blocks by pre-filling them with relevant bindings. Card titles will default to either the primary display column, if one exists, or else another "readable" type. Readable types are defined as strings, long form text, options, multi-options and numbers.

Card subtitles and descriptions are also filled in, if possible, with more readable fields. For description, we try to use a long form field if one exists.

Card image URLs are filled in with an attachment URL if possible, supporting both single and multi attachment lists.

These properties are all reset if you change the card datasource, so you'll always have working bindings.

Here's a newly created cards block, showing everything filled in. This is obviously test data so please excuse the stock images that are inconsistent with the text.
image

Obviously there's a bit of luck involved here since we are just guessing which columns to use (especially for subtitle and description) but it's very easy to change them.

## Addresses
- https://linear.app/budibase/issue/BUDI-8177/prefill-card-block-with-bindings

## Launchcontrol
Improve card blocks by pre-filling settings.


packages/builder/src/stores/builder/components.js
```diff
@@ -477,7 +552,8 @@ export class ComponentStore extends BudiStore {
       // Mutates the fetched component with updates
       const patchResult = patchFn(component, screen)
 
-      // Mutates the component with any required settings updates
+      // Post processing
+      this.enrichEmptySettings(component, { screen, useDefaultValues: false })
```
aptkingston, line 556: I've added `enrichEmptySettings` inside the core component patch function. This means that after any change to a component, we'll attempt to enrich empty settings again (rather than only on creation, which we do now).
This has a few benefits - the main one being we can now add more logic (like this cards block stuff) in here and ensure that it will be ran again if required. This works well in combination with the "resetOn" manifest key. Use that key to wipe settings, then this enrichment will pick them up and default them again.

shogunpurple (approved):
LGTM, although since it's store logic would be good to have a unit test to cover

packages/builder/src/stores/builder/components.js
deanhannigan: This behaviour seems to have unpredictable results when clearing the card block values. If I delete `Title`, `Subtitle` and `Description`, sometimes all 3 values will remain empty but in some instances maybe 2 values will regenerate. When I deleted all 3 values, some internal defaults appeared and some old config remained. As below:

Screenshot 2024-05-20 at 16 19 41

deanhannigan (commented):
This a fantastic UX update! I threw every type of supported config at the CardBlock and it was always able to figure out the appropriate defaults on create. I did run into an issue with the new update behaviour which I mentioned in my comment.

packages/builder/src/stores/builder/components.js
aptkingston: This seems to just be a race condition thing depending on what field you clear last. The last field you clear will trigger the enrichment and all the settings do get filled back in, but it looks like the blur of the final field sometimes clears whatever you have selected again. 
I'll see if I can put in something to prevent that happening.

aptkingston: Following up on this again - it's an existing issue with any component where if a setting value is `foo`, and you clear it, then empty setting enrichment sets the value back to `foo`, svelte sees that the value is identical and so does not re-render the setting with the same value, yet the input itself has actually been cleared. So you end up in this weird position where the saved value does exist but the input is empty.
Not entirely sure how to solve this without doing something nuclear like key the entire component setting section by screen rev. It's very much an edge case though as it depends on clearing the exact value that will be repopulated by enrichment.

deanhannigan: I see what you mean now. Something to keep in mind but in no way a new thing 👍

deanhannigan (approved):
LGTM!

aptkingston:
> LGTM, although since it's store logic would be good to have a unit test to cover

Done!

PR #13538, opened 2024-04-22
Chart Component Fixes/Improvements. Adds Tooltip to Provide Context on Component Input Types

Ghrehh:
## General Changes

- Fixed an issue where alerts for required fields weren't showing up on freshly added chart components, fixed another related issue hidden by this bug where component controls nested in sections didn't check their parent's `dependsOn` field to check if they're own `dependsOn` validation should be ran.

## Chart Changes

- Upgraded to latest Apex charts
- Removed use of Apex charts svelte shim; we now use the library directly.
- Fixed an issue where selecting nullish or non-scalar values to display in multiselect charts would break the entire chart.
- Histograms now work with negative values.
- `datetime` labels now display a human readable date instead of a UNIX timestamp or raw ISO 8601 string.
- `Horizontal` mode now works for histograms and bar charts with date labels.
- Fixes an issue where switching to certain label types would break a chart component until page refresh.
- `datetime` fields passed as values to charts are now converted to unix time integers, which seems potentially useful. They were previously handled by directly running "parseFloat" on them, only parsing the year of the iso string as a number, which didn't seem very useful.
- In dev mode, if a chart is provided inputs that have no rows, an alert will tell users this is the case.
- Removed the previous chart builder class, instead we explicitly build the Apex chart config object in each chart component, in my opinion this is much easier to understand, even if it's a bit more explicit, there's also less method calling boilerplate.
- Candlestick charts can now accept numbers and string-like values as the date input if the values are correctly formatted.
- Donut and Pie charts now longer break until page reload when a non-numeric value is selected as their value input.

## Explanation Tooltip

- Added a tooltip that provides information on column types and their compatibility with a particular component input when hovering over FieldSelects or MultiFieldSelects.
- These tooltips also have further information that can be accessed on hover explaining certain topics, like how strings can be converted to numbers for numerical inputs.
- There are some other QoL features, like being able to see an at a glance summary of the configuration of a particular column, and the ability to jump straight to things like documentation and the column's table, and example of these are below, though there are more.

|info|image|
|---|---|
|column details menu|Screenshot 2024-04-22 at 09 29 33|
|compatibility details menu|Screenshot 2024-04-22 at 09 29 19|
|explanation tooltip|Screenshot 2024-04-22 at 09 28 57|

## Feature branch env
[Feature Branch Link](http://fb-fix-chart-data-selection-2.fb.qa.budibase.net)


packages/server/src/api/controllers/component.ts
```diff
@@ -20,7 +20,7 @@ export async function fetchAppComponentDefinitions(ctx: UserCtx) {
     const definitions: { [key: string]: any } = {}
     for (let { manifest, library } of componentManifests) {
       for (let key of Object.keys(manifest)) {
-        if (key === "features") {
+        if (key === "features" || key === "typeSupportPresets") {
```
mike12345567, line 23: What do these two manifest keys do exactly? would be good to have a comment, or say an array to check against like `MANIFEST_KEYS_THAT_DO_SOMETHING_SPECIAL.includes(key)`

packages/client/manifest.json
```diff
@@ -1825,19 +1883,19 @@
...
-          "Palette 2",
-          "Palette 3",
-          "Palette 4",
-          "Palette 5",
-          "Palette 6",
-          "Palette 7",
-          "Palette 8",
-          "Palette 9",
-          "Palette 10"
+          { "label": "Custom", "value": "Custom" },
```
deanhannigan, line 1888: A breaking change for existing charts with the palette value already specified. The values are currently stored as "Palette 1".

This is the same for all chart types

packages/client/package.json
```diff
@@ -32,11 +32,12 @@
     "@spectrum-css/tag": "^3.1.4",
     "@spectrum-css/typography": "^3.0.2",
     "@spectrum-css/vars": "^3.0.1",
-    "apexcharts": "^3.22.1",
+    "apexcharts": "^3.48.0",
     "dayjs": "^1.10.8",
     "downloadjs": "1.4.7",
     "html5-qrcode": "^2.2.1",
     "leaflet": "^1.7.1",
+    "lodash": "^4.17.21",
```
deanhannigan, line 40: I don't think it's necessary to add in lodash to the client.

packages/client/src/components/app/charts/ApexChart.svelte
```diff
@@ -1,25 +1,82 @@
 <script>
   import { getContext } from "svelte"
-  import { chart } from "svelte-apexcharts"
-  import Placeholder from "../Placeholder.svelte"
+  import ApexCharts from "apexcharts"
+  import { Icon } from "@budibase/bbui"
+  import { cloneDeep } from "lodash"
```
deanhannigan, line 5: It isn't necessary to import lodash here. There is a simple `cloneDeep` in the `@budibase/bbui` helpers that should do the job.

packages/client/src/components/app/charts/ApexChart.svelte
```diff
@@ -1,25 +1,82 @@
...
+
+  let chartElement
+  let chart
+  let currentType = null
+
+  const updateChart = async newOptions => {
+    // Line charts have issues transitioning between "datetime" and "category" types, and will ignore the provided formatters
+    // in certain scenarios. Rerendering the chart when the user changes label type fixes this, but unfortunately it does
+    // cause a little bit of jankiness with animations.
+    if (newOptions?.xaxis?.type && newOptions.xaxis.type !== currentType) {
```
deanhannigan, line 23: Optional chaining on the right hand side also? 

```javascript
newOptions?.xaxis?.type !== currentType
```

packages/client/src/components/app/charts/ApexChart.svelte
```diff
@@ -1,25 +1,82 @@
...
-  {#key options.customColor}
-    <div use:chart={options} use:styleable={$component.styles} />
-  {/key}
-{:else if $builderStore.inBuilder}
-  <div use:styleable={$component.styles}>
-    <Placeholder />
-  </div>
-{/if}
+{#key optionsCopy?.customColor}
+  <div class:hide use:styleable={$component.styles} bind:this={chartElement} />
```
deanhannigan, line 59: Go for `class:hide={noData}` rather that creating another `hide` variable.

deanhannigan (changes requested):
This is a fantastic piece of work Gerard! The UX improvements to charts make such a huge difference and I genuinely want to see your context menus everywhere in Budibase 🚀 They are a brilliant bit of design and incredibly intuitive!

Some minor comments in the PR and some bugs/notes outlined here:

`Bug` ~~custom color theme palette not loading~~
  - ~~This seems to be an issue on all chart types~~

`Bug` - QR not compatible for Label/Data column
  - The value of a QR code will always resolve to a string
  - In many cases, the text could be appropriate for both string or number like fields.

`Note`
  - ~~Different layout for Pie chart? Is this intentional?~~
  - ~~Prod has a larger diagram with a scrollable legend on the right~~
  - ~~New UX has a small diagram with a static legend at the top.~~

`Note` popover color
  - ~~Nested Props in formblock~~
    - ~~--spectrum-global-color-gray-100~~
  - ~~Context popover background is currently set to~~
    - ~~--spectrum-global-color-gray-75~~
    
`Note`
- I'm not sure the bullet points are required in the primary hover. Purely an opinion, in no way a blocker.

packages/server/src/api/controllers/component.ts
Ghrehh: They're keys used by the frontend that provide access to features based on the client version. They're specified here to avoid having the `@budibase/whatever` prefix appended to them like the components in the manifest. 

Good idea to note down via code or a comment what it does, will do.

packages/bbui/src/Form/Core/Multiselect.svelte
```diff
@@ -17,6 +17,8 @@
   export let customPopoverHeight
   export let open = false
   export let loading
+  export let onOptionMouseenter = () => {}
```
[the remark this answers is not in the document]
Ghrehh: I did this intentionally to match the js event name which is `mouseenter`

packages/client/package.json
Ghrehh: It's used in the client but not explicitly specified in the package.json, and it's not specified at the root level either, made sense to me to add it 🤷

packages/client/src/components/app/charts/ApexChart.svelte
Ghrehh: If the first condition passes then the optional chaining won't be necessary I think.

packages/builder/src/components/design/settings/controls/Explanation/DetailsModal/subjects/Support.svelte
```diff
@@ -0,0 +1,35 @@
...
+      text="Partially compatible"
+    />
+    <span class="body"
+      >Partially compatible with the input, but beware of other caveats
+      mentioned.</span
+    >
+  </Section>
+  <Section>
+    <InfoWord icon="Alert" color="var(--red)" text="Not compatible" />
+    <span class="body">Imcompatible with the component.</span>
```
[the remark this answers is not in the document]
Ghrehh: Doh

packages/builder/src/stores/builder/components.js
```diff
@@ -104,6 +104,7 @@ export class ComponentStore extends BudiStore {
       ...state,
       components,
       customComponents,
+      typeSupportPresets: components?.typeSupportPresets ?? {},
```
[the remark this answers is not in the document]
Ghrehh: Done. Ended up moving this to the app store also.

packages/client/package.json
Ghrehh: Will remove

packages/client/src/components/app/charts/ApexChart.svelte
Ghrehh: Replaced it with [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead

packages/bbui/src/Tooltip/Context.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+  const handleMouseleave = () => {
+    hovering = false
+  }
+</script>
+
+<Portal {target}>
+  <div
+    role="tooltip"
+    on:mouseenter={handleMouseenter}
```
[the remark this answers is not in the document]
Ghrehh: Same as above

packages/client/manifest.json
```diff
@@ -1784,19 +1832,29 @@
         "label": "Label column",
         "key": "labelColumn",
         "dependsOn": "dataProvider",
+        "explanation": {
+          "typeSupport": {
+            "preset": "stringLike"
```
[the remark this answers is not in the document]
Ghrehh: At least `typesupport.preset` is necessary, so components can specify one off typeSupport objects like:

```
 "explanation": {
          "typeSupport": {
                  "supported": ["number", "boolean"],
                  "partialSupport": [
                    { "type": "longform", "message": "stringAsNumber" },
                  ],
                  "unsupported": [
                    { "type": "json", "message": "jsonPrimitivesOnly" }
                  ]
          }
}
```
Could have also implemented it as "if `typeSupport` value is a string then it's a preset",but this was more explicit so I preferred it.


the `explanation` object is just namespacing/futureproofing tbh, if we wanted to add another object/field relating to the explanation modal it's nicer to have them in one object and not have them both in the field configuration object individually. It'd also be hard to add this namespacing in future as it'd be a breaking manifest change and would require some code to work around old/new versions.

packages/client/manifest.json
Ghrehh: Ah yeah that sucks, changed it back.

packages/client/manifest.json
```diff
@@ -1626,19 +1663,29 @@
...
+            "preset": "numberLike"
+          }
+        },
         "required": true
       },
       {
         "type": "select",
         "label": "Format",
-        "key": "yAxisUnits",
+        "key": "valueUnits",
```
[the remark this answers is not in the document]
Ghrehh: Reverted this.

Ghrehh:
@deanhannigan Thanks a bunch for the thorough review, I've addressed or responded to everything, if you could give this another looks when you get a chance I'd appreciate it 🙏

packages/client/src/components/app/blocks/ChartBlock.svelte
```diff
@@ -85,7 +85,7 @@
           legend,
           animate,
           ...colors,
```
deanhannigan, line 87: In order for the custom colors to work, the array needs to be changed to an object.

```javascript
$: colors = c1 && c2 && c3 && c4 && c5 ? { c1, c2, c3, c4, c5 } : null
```

deanhannigan (commented):
Just one minor note for the custom palette issue and one change to the `Explanation` background colour and this is good to go 👍

packages/client/src/components/app/blocks/ChartBlock.svelte
Ghrehh: I'm actually not sure what how the original implementation of this worked, but I changed it just pass `c1`, `c2` to the underlying chart component directly and it works now.

deanhannigan (approved):
LGTM! Great work Gerard!

packages/bbui/src/Form/Core/Picker.svelte
```diff
@@ -199,6 +201,8 @@
             aria-selected="true"
             tabindex="0"
             on:click={() => onSelectOption(getOptionValue(option, idx))}
+            on:mouseenter={e => onOptionMouseenter(e, option)}
+            on:mouseleave={e => onOptionMouseleave(e, option)}
```
aptkingston, line 205: NAB: would be nice to dispatch custom `optionenter` and `optionleave` events here, which can be just proxied on up the component chain rather than having to have a bunch of prop drilling with NOPs but it's not a big deal - works the same.

packages/client/src/components/Component.svelte
```diff
@@ -283,10 +283,23 @@
         const dependsOnKey = setting.dependsOn.setting || setting.dependsOn
         const dependsOnValue = setting.dependsOn.value
         const realDependentValue = instance[dependsOnKey]
+
+        const sectionDependsOnKey =
+          setting.sectionDependsOn?.setting || setting.sectionDependsOn
```
aptkingston, line 288: I see what you're doing with this (I remember the awkwardness before of trying to define the chart settings in a way that the "depends on" stuff worked with only a single key) but to be honest this would motivate me to rewrite dependsOn to just support an array of keys. We could still do what you're doing below and dynamically enrich the setting dependsOn keys with any keys from the section itself.

We can handle that separately though - this PR is big enough already!

aptkingston (approved):
Huge amount of work here! The amount of components added for all the explanation stuff is insane.

No major issues at all, just NABs. I agree with Dean in principle about casing (just because the DOM syntax is a certain way doesn't we can't make it more readable in our own code) but again that's very much NAB.

So I'm happy to merge away as is 👍 

Great to make the charts more stable.

PR #13543, opened 2024-04-22
Single user column type

adrinr:
## Description
Adding a new column type for single users. These new columns will return the data as an object, instead of the existing `bb-reference` that returns an array, allowing better validation and cleaner bindings.
No existing data will be migrated, the old columns will still return objects and it will keep the constraints validations (allowing multiple or single when required)

With this change, we will have the following strategy:
1. The `bb_reference` will be a type that will store an array of internal IDs
2. The `bb_reference_single` will be a type that will store a single internal ID
3. Both column types will have a subtype. For the time being only `user subtype` is allowed, but technically it can support any internal reference such as `apps`, `groups`, `tables`... The subtype will define:
i. Input validation: are the IDs valid for that object?
ii. Output enhancement: what other data needs to be populated when enhancing the id

## Addresses
- [BUDI-8123 - Single user column type](https://linear.app/budibase/issue/BUDI-8123/single-user-column-type)

### Adding a single-user column
| Before | After |
|--------|--------|
| image | image | 

### Adding a multi-user column
| Before | After |
|--------|--------|
| image | image | 

### Display cells
| Before | After |
|--------|--------|
| image | image|

## Launchcontrol
Adding a new column type for single-user values

## Feature branch env
[Feature Branch Link](http://fb-budi-8123-single-user-column-type.fb.qa.budibase.net)


packages/server/src/utilities/rowProcessor/bbReferenceProcessor.ts
```diff
@@ -1,93 +1,190 @@
...
+            firstName: u.firstName,
+            lastName: u.lastName,
+          }))
+        }
+        default:
+          throw utils.unreachable(subtype)
+      }
+    }
+    case FieldType.BB_REFERENCE_SINGLE: {
+      if (!value) {
```
samwho, line 162: Should this uncertainty be reflected in the function signature? Right now there's no indication that `value` could be undefined, or the function can return undefined.

packages/server/src/utilities/schema.ts
```diff
@@ -163,21 +168,31 @@ export function parse(rows: Rows, schema: TableSchema): Rows {
...
 ): boolean {
-  switch (columnSubtype) {
+  if (typeof data !== "string") {
+    return false
+  }
+
+  if (type === FieldType.BB_REFERENCE_SINGLE) {
+    if (!data) {
+      return true
+    }
```
samwho, line 182: ... is this correct?

packages/types/src/documents/app/row.ts
```diff
@@ -100,13 +100,17 @@ export enum FieldType {
    */
   BIGINT = "bigint",
   /**
-   * a JSON type, called User within Budibase. This type is used to represent a link to an internal Budibase
+   * an JSON type, called Users within Budibase. It will hold an array of strings. This type is used to represent a link to an internal Budibase
```
samwho, line 103:
```suggestion
   * a JSON type, called Users within Budibase. It will hold an array of strings. This type is used to represent a link to an internal Budibase
```

packages/client/src/components/app/forms/RelationshipField.svelte
aptkingston, line 1: Thanks for looking at this - it was a real mess and was spamming API calls.

aptkingston (approved):
Frontend integration looks good to me! Nice one 👌

.eslintrc.json
```diff
@@ -53,7 +53,8 @@
             "ignoreRestSiblings": true
           }
         ],
-        "local-rules/no-budibase-imports": "error"
+        "no-redeclare": "off",
+        "@typescript-eslint/no-redeclare": "error"
```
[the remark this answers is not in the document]
adrinr: There is a conflict between this rule and the typescript function overload. These settings were added as recommended by https://typescript-eslint.io/rules/no-redeclare/

packages/builder/src/constants/backend/index.js
```diff
@@ -158,16 +159,29 @@ export const FIELDS = {
...
   USER: {
+    name: "User",
+    type: FieldType.BB_REFERENCE_SINGLE,
+    subtype: FieldSubtype.USER,
+    icon: TypeIconMap[FieldType.BB_REFERENCE_SINGLE][
+      FieldTypeSubtypes.BB_REFERENCE.USER
+    ],
+  },
+  // Used for display of editing existing columns
+  OLD_USER: {
```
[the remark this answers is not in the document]
adrinr: I can see that ahah Renaming it!

packages/server/src/utilities/schema.ts
adrinr: Yes, this was not considering required settings, as we don't do anywhere else. I just changed this specific use case to check it

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -52,18 +51,19 @@
...
-    if (!initialValuesProcessed && primaryDisplay) {
+    if (primaryDisplay && fieldState && !optionsObj) {
       // Persist the initial values as options, allowing them to be present in the dropdown,
       // even if they are not in the inital fetch results
-      initialValuesProcessed = true
-      optionsObj = (fieldState?.value || []).reduce((accumulator, value) => {
+      const valueAsSafeArray =
+        (!fieldState.value || Array.isArray(fieldState.value)
+          ? fieldState.value
+          : [fieldState.value]) || []
```
samwho, line 65: I found this quite hard logic to follow. It looks like you're trying to ensure `fieldState.value` is an array without accidentally creating an array like `[undefined]`? Might be worth creating a function specifically for this, though I wouldn't block you on it.

samwho (approved):

PR #13544, opened 2024-04-23
Update helm charts to account for temp bucket

PClmnt:
## Description
Adds the properties to pull in the temp bucket to our helm chart

### DIFF
--- a/charts/budibase/templates/app-service-deployment.yaml
+++ b/charts/budibase/templates/app-service-deployment.yaml
@@ -153,7 +153,7 @@ spec:
         - name: ENCRYPTION_KEY
           value: {{ .Values.globals.bbEncryptionKey | quote }}
         - name: TEMP_BUCKET_NAME
-          value: {{ .Values.global.tempBucketName | quote }}
+          value: {{ .Values.globals.tempBucketName | quote }}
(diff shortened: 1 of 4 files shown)


charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -152,6 +152,8 @@ spec:
           value: {{ .Values.globals.tenantFeatureFlags | quote }}
         - name: ENCRYPTION_KEY
           value: {{ .Values.globals.bbEncryptionKey | quote }}
+        - name: TEMP_BUCKET_NAME
+          value: {{ .Values.globals.tempBucketName | quote }}
```
calexiou, line 156: This is more about me being picky but could we move this close to the rest of the buckets?

https://github.com/Budibase/budibase/blob/e901ea8a46b163d233abd1d082d93d146eb24269/charts/budibase/templates/app-service-deployment.yaml#L101

It makes more sense to have them all grouped, just because we can search faster and use them as in-code docs.

calexiou (approved):

charts/budibase/templates/app-service-deployment.yaml
PClmnt: Done!

mike12345567 (approved):

PR #13549, opened 2024-04-23
Filter bindings

deanhannigan:
## Description
Added the ability to use bindings when filtering tables or creating views. Supports `handlebars`, `js` and `js snippets`!

- When browsing a table in the Data section you can now apply a number of bindings directly to a filter. 
- You can incorporate a mixture of useful bindings and save them as a reusable `View`.
- Internal datasources utilising `User` field types can filter rows by app user id. The `Current User` bindings can be used to filter rows by the currently logged in user.
- Components in the builder can also apply filter bindings to their respective datasource settings.
- Updated UX utilising the new drawers 

Binding Categories
- `Current User` - metadata for the currently authenticated user. If a view or filter is executed outside of a session e.g. in a public screen, the `Current User` bindings will not resolve.
- `Date` - contains a `Server date` binding. This will reflect the timestamp at the moment the view/filter is executed.
- `Helpers` - the same handlebars helpers used throughout Budibase e.g. `{{ default Current User._firstName 'no user'}}`

## Addresses
- https://linear.app/budibase/issue/BUDI-8152/secure-access-to-my-records

## Screenshots
Previously the filter type only allowed the use of static values. Now there is a second binding option.
Screenshot 2024-04-23 at 15 28 55

A view created with a simple js binding to check for rows created in the last 2 weeks.
Screenshot 2024-04-23 at 14 45 02

A view created with a js snippet.
Screenshot 2024-04-23 at 14 44 23

A view filtering by a user column matching by the sessions current user id. This example utilises the handlebars syntax
Screenshot 2024-04-23 at 14 58 38

## Launchcontrol
Bindings now available when creating views, table filters or filtering builder component datasources.


packages/server/src/api/controllers/row/index.ts
```diff
@@ -198,8 +198,21 @@ export async function destroy(ctx: UserCtx<DeleteRowRequest>) {
 export async function search(ctx: Ctx<SearchRowRequest, SearchRowResponse>) {
   const tableId = utils.getTableId(ctx)
 
+  // Current user context for bindable search
+  const { _id, _rev, firstName, lastName, email, status, roleId } = ctx.user
```
mike12345567, line 202: Can this be a something that lives in the user SDK? Under `server/src/sdk/users/utils.ts` there are some utilities, a function like `getUserContextBindings(ctx.user)` would help describe what is happening here, and would mean you can just use it like:
```ts
const enrichedQuery = await utils.enrichSearchContext(
    { ...ctx.request.body.query },
    {
      user: sdk.users.getUserContextBindings(ctx.user),
    }
  )
```

deanhannigan: Yep that makes perfect sense. I'll get it shifted out 👍

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -156,6 +171,225 @@ describe.each([
     return expectSearch({ query })
   }
 
+  // Ensure all bindings resolve and perform as expected
+  !isSqs &&
```
samwho, line 175: How come you're excluding SQS here? Is it because it _doesn't_ work, or is it because it _shouldn't_ work?

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -35,11 +43,18 @@ describe.each([
...
+      code: "return function (weeks) {\n  const currentTime = new Date();\n  currentTime.setDate(currentTime.getDate()-(7 * (weeks || 1)));\n  return currentTime.toISOString();\n}",
+    },
+  ]
+
   beforeAll(async () => {
     if (isSqs) {
       envCleanup = config.setEnv({ SQS_SEARCH_ENABLE: "true" })
     }
-    await config.init()
+    await config.init({ snippets })
```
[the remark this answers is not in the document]
deanhannigan: Yep, sounds good. When the changes are in I'll revert snippet modifications and go with a dedicated update call 👍

packages/server/src/api/routes/tests/search.spec.ts
deanhannigan: The tests started failing at the SQS stage so I added the exclusion to verify that the rest of the matrix was running without issue. I'll refactor out my snippet changes and try to figure out what the issue is with SQS.

samwho: Don't worry about it, leave this in and put a `TODO(samwho): fix with SQS` or some such. I'll handle it, don't want to block this on SQS. 🙂

packages/server/src/api/routes/tests/search.spec.ts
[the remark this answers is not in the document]
samwho: They're in.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -156,6 +178,230 @@ describe.each([
...
+          contains: { multi_user: ["{{ [user]._id }}"] },
+        }).toContainExactly([
+          {
+            name: "multi user with session user",
+            multi_user: allUsers,
+          },
+        ])
+      })
+
+    !isSqs &&
```
samwho, line 353: Would you mind putting `// TODO(samwho): fix for SQS` comments on these so I remember them later? 🙏

samwho (approved):
From a backend perspective, barring a couple of small questions, I'm happy with this.

Thanks so much for extending `search.spec.ts`, the tests for this are really good!

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -172,173 +179,178 @@ describe.each([
...
+          type: FieldType.BB_REFERENCE,
+          subtype: BBReferenceFieldSubType.USER,
+        },
+        multi_user: {
+          name: "multi_user",
+          type: FieldType.BB_REFERENCE,
+          subtype: BBReferenceFieldSubType.USERS,
+        },
       })
+      await createRows([...rows(config.getUser())])
```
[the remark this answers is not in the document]
deanhannigan: Possibly a leftover from another test. Removed now 👍

samwho (approved):

PR #13557, opened 2024-04-24
Changing client library to be retrieved via pre-signed URL for Cloud

mike12345567:
## Description
Bringing back the old mechanism of returning the client library through a pre-signed URL, rather than always serving through the service.

An attempt to address some of the periods of very slow S3 performance we have been experiencing recently.


samwho (approved):
Other than the lint error this _seems_ reasonable to me. We'll want to test it in QA for sure.

Issue #13537, opened 2024-04-22
Yarn setup run fail in my mac


kiner-tang:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: <method> 
  - Budibase Version: <version> 
  - App Version: <version> 
- Cloud
  - Tenant ID: <tenantId> 

**Describe the bug**
<img width="891" alt="image" src="https://github.com/Budibase/budibase/assets/10286961/4d2747d2-ee01-4d97-be7d-c9104f97925f">

And the `@budibase/pro` version will be change to `latest` automatically when i run `yarn setup` or `yarn`

![image](https://github.com/Budibase/budibase/assets/10286961/d790087c-2872-48a6-a79b-1c3030a2b7ba)

**To Reproduce**

**Expected behavior**

**Screenshots**
  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**

**Additional context**

kiner-tang:
> Is the `Couldn't find package "@budibase/pro"` happening consistently? The change of the package reference to `latest` is expected, but the package `latest` should be available publicly to you: https://www.npmjs.com/package/@budibase/pro/v/latest
> 
> What happens if you run a second time `yarn`?

I get the same issue when I run `yarn` again.
<img width="845" alt="image" src="https://github.com/Budibase/budibase/assets/10286961/0c713445-5cc1-45c6-ad3f-b7efd9c52f11">

adrinr:
Could you share what version of yarn are you using? `yarn -v`

kiner-tang:
> Could you share what version of yarn are you using? `yarn -v`

1.22.19

adrinr:
Could you please pull the latest master from github and check if this PR fixed it for you? https://github.com/Budibase/budibase/pull/13559

kiner-tang:
> Could you please pull the latest master from github and check if this PR fixed it for you? #13559

Thanks, It run!!

PR #13559, opened 2024-04-25
Resolve oss pro via yarn resolutions

adrinr:
## Description
Currently we have a script that changes the pro package to use latest when the code is running without the submodules. This is often confusing and it makes solving OSS issues harder. We removed this logic using yarn resolutions instead. If the submodule exist, yarn will automatically pick the local one. Otherwise it will pull from the npm registry.

| `yarn why` with the submodule loaded | `yarn why` without the submodule |
|--------|--------|
| image | image |

OSS pipeline run:
- https://github.com/Budibase/budibase/actions/runs/8830359032

## Addresses
- [BUDI-8186 - Yarn setup run fail in my mac](https://linear.app/budibase/issue/BUDI-8186/yarn-setup-run-fail-in-my-mac)
- https://github.com/Budibase/budibase/issues/13537


mike12345567 (approved):
LGTM - this does seem to work - its an interesting behaviour! Nice to get rid of the scripts around this.

PR #13561, opened 2024-04-25
Draft: Testing LRU caching of datasource clients

mike12345567:
## Description
Implementing a basic cache for Postgres, using LRU and `merkle-json` (https://www.npmjs.com/package/merkle-json) to hash the configuration.

The goal here is to reduce wasteful client creation/destruction, we've seen in the past over-connecting when using search heavy pages in Budibase, this should help with that as well - it should also speed up self host implementation in theory, as they'll have a pretty much permanent connection when the app is in use.

By default this is set to a TTL of 10 minutes and max of 50 clients, for self host these numbers will be fine, but in the cloud they may need some tweaking. In theory the worst case is that we start to connect too often, but this would need some QA to confirm.

This still requires implementation for all other datasource+ - datasources won't really benefit from this as they are executed within a thread that has a limited lifetime, its not really possible to cache those clients.

### DIFF
--- a/packages/server/src/integrations/base/clientCache.ts
+++ b/packages/server/src/integrations/base/clientCache.ts
@@ -35,14 +35,7 @@ export class ClientCache<T> {
       .digest("base64")
   }
 
-  has(configHash: string): boolean {
-    return clientCache.has(configHash)
-  }
-
-  get(configHash: string): T {
-    if (!this.has(configHash)) {
-      throw new Error("Cannot get client - missing a has check")
-    }
+  get(configHash: string): T | undefined {
     return clientCache.get(configHash) as T
   }
 
--- a/packages/server/src/integrations/base/clientCache.ts
+++ b/packages/server/src/integrations/base/clientCache.ts
@@ -36,7 +36,10 @@ export class ClientCache<T> {
   }
 
   get(configHash: string): T | undefined {
-    return clientCache.get(configHash) as T
+    const value = clientCache.get(configHash) as ClientCacheValue<T>
+    if (value && value.client) {
+      return value.client
+    }
   }
 
   set(configHash: string, client: T) {
--- a/packages/server/src/integrations/base/clientCache.ts
+++ b/packages/server/src/integrations/base/clientCache.ts
@@ -1,21 +1,6 @@
-import { Duration } from "@budibase/backend-core"
+import { Duration, context } from "@budibase/backend-core"
 import LRUCache from "lru-cache"
-
-interface Merkle {
-  hash: (json: Record<string, any>) => string
-  stringify: (json: Record<string, any>) => string
-}
-
-interface MerkleConstructor {
-  new (): Merkle
-}
-
-// no typescript definition - fill it in
-const { MerkleJson } = require("merkle-json") as {
-  MerkleJson: MerkleConstructor
-}
-
-const jsonHash = new MerkleJson()
+import crypto from "crypto"
 
 type ClientCacheValue<T> = {
   destroy: (client: T) => Promise<void>
@@ -42,7 +27,12 @@ export class ClientCache<T> {
   }
 
   hash(config: Record<string, any>) {
-    return jsonHash.hash(config)
+    const tenantId = context.getTenantId()
+    const json = JSON.stringify(config)
+    return crypto
+      .createHash("sha256")
+      .update(tenantId + json)
+      .digest("base64")
   }
 
   has(configHash: string): boolean {
--- a/packages/server/src/integrations/base/clientCache.ts
+++ b/packages/server/src/integrations/base/clientCache.ts
@@ -1,6 +1,19 @@
 import { Duration } from "@budibase/backend-core"
 import LRUCache from "lru-cache"
-import merkle from "merkle-json"
+
+interface Merkle {
+  hash: (json: Record<string, any>) => string
+  stringify: (json: Record<string, any>) => string
+}
+
+interface MerkleConstructor {
+  new (): Merkle
+}
+
+// no typescript definition - fill it in
+const MerkleJson = require("merkle-json") as MerkleConstructor
+
+const jsonHash = new MerkleJson()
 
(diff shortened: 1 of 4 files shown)


packages/server/src/integrations/postgres.ts
```diff
@@ -192,8 +194,10 @@ class PostgresIntegration extends Sql implements DatasourcePlus {
           }
         : undefined,
     }
-    this.client = new Client(newConfig)
-    this.open = false
+    this.clientCache = new ClientCache<Client>(async client => {
+      await this.closeConnection(client)
```
adrinr, line 198: When is this called?

packages/server/src/integrations/base/clientCache.ts
```diff
@@ -0,0 +1,65 @@
...
+  hash: (json: Record<string, any>) => string
+  stringify: (json: Record<string, any>) => string
+}
+
+interface MerkleConstructor {
+  new (): Merkle
+}
+
+// no typescript definition - fill it in
+const { MerkleJson } = require("merkle-json") as {
```
samwho, line 14: `merkle-json` appears to cache the hash values of objects on the object itself under the key `merkleHash`. Does this open up the possibility of users putting `merkleHash` on their DS config objects to try and "steal" a connection to another user's datasource? Should we add something that isolates tenants from each other explicitly instead of relying on the hash being unique?

packages/server/src/integrations/base/clientCache.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const clientCache = new LRUCache<string, ClientCacheValue<any>>({
+  // max 50 active clients per service
+  max: 50,
+  ttl: Duration.fromMinutes(10).toMs(),
+  updateAgeOnGet: true,
+  dispose: async (value: ClientCacheValue<any>) => {
+    await value.destroy(value.client)
+  },
```
samwho, line 33: When the max is reached, will an active connection get killed mid-query? My suspicion here is yes, which is possibly catastrophic. We could end up in situations where users are making requests, they're failing because they get killed by another request, they retry and kill another request, and it thrashes like this for a while.

I think there's probably a good reason that databases typically use pools and not caches for this sort of thing: you open a new identical connection if an existing one is already in use. Which makes me think, could we give out the same connection twice using this code? Is that bad?

packages/server/src/integrations/base/clientCache.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.destroy = destroy
+  }
+
+  hash(config: Record<string, any>) {
+    return jsonHash.hash(config)
+  }
+
+  has(configHash: string): boolean {
+    return clientCache.has(configHash)
+  }
```
samwho, line 50: We could remove `has` entirely and rely on `get` returning undefined when the item is not found. This removes a check from every time we try to get a connection.

I think this also solves a potential problem where `has` could return true, then the TTL could expire, then `get` returns `undefined`.

packages/server/src/integrations/base/clientCache.ts
```diff
@@ -0,0 +1,65 @@
...
+type ClientCacheValue<T> = {
+  destroy: (client: T) => Promise<void>
+  // can't really know the type of the client, they are all so different
+  client: T
+}
+
+const clientCache = new LRUCache<string, ClientCacheValue<any>>({
+  // max 50 active clients per service
+  max: 50,
+  ttl: Duration.fromMinutes(10).toMs(),
```
samwho, line 29: This is almost certainly higher than the time an idle connection will be kept open by the database server, so you may find that after 9 minutes a client goes to use it and you get a connection reset error. I'd probably be looking to keep this to 60 seconds or less.

packages/server/src/integrations/base/clientCache.ts
mike12345567: The merkle-json library only does this if we call like `merkle.hash(json, true)` - the boolean parameter telling it to re-use hashes - although I'm not a huge fan of this library due to its lack of in-built TS support and its relatively low usage, there is probably a better way to do this.

I did consider using the `datasourceId` instead for the key, but the downside to this approach is connections are cached even if the config has changed which would lead to weird behaviour. We could retrieve the tenantId and add it to the start of the hash though, like `tenantId-hash` for the cache key.

packages/server/src/integrations/base/clientCache.ts
mike12345567: The first part is probably my biggest concern with this, I want to give this a test, but realistically it would be safer to mark that the connection is in-use and not close it until the job is finished - I think this is feasible.

The second part - we probably need to use a connection pool in the client - you are right that if a user makes multiple requests to the same DB these will all likely clash - I can adjust this to configure the client as a connection pool so that the cached version is just a pool of clients.

packages/server/src/integrations/base/clientCache.ts
mike12345567: If we use a connection pool then this issue should go away!

packages/server/src/integrations/base/clientCache.ts
samwho: I don't think it'd be terribly challenging to do the hashing ourselves. Concat the tenantId, host, port, user, password, whatever else, do a sha256 on that.

packages/server/src/integrations/base/clientCache.ts
samwho: I think caching a pool instead of a client solves all of the concerns here.

mike12345567: Pooling implemented.

packages/server/src/integrations/base/clientCache.ts
mike12345567: Pooling implemented.

packages/server/src/integrations/base/clientCache.ts
mike12345567: Removed has check.

PR #13564, opened 2024-04-25
Search tests for FieldType.AUTO

samwho:
## Description

Continuing my crusade to write a comprehensive test suite for our search endpoint, this PR adds tests for `FieldType.AUTO`. This is a field type we only support on internal tables, so these tests do not run against external databases. Also, we know that Lucene does not correctly sort numbers (it uses lexicographical sorting instead of numerical), so I have made the sorting tests only run against SQS.


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -59,7 +61,10 @@ describe.each([
   }
 
   async function createRows(rows: Record<string, any>[]) {
-    await Promise.all(rows.map(r => config.api.row.save(table._id!, r)))
+    // await config.api.row.bulkImport(table._id!, { rows })
```
adrinr, line 64: Delete if not needed

adrinr: Was the bulk import an issue? Is it a performance issue or what was the problem?

mike12345567 (approved):

PR #13567, opened 2024-04-26
Support attachment columns in Automations

PClmnt:
## Description
This PR adds support for attachment columns within automations

- Adding attachments using the Create / Update row actions
- Support for both single and multiple attachment columns
- Refactors some of the code out of the Send Email block for common use between actions

## Addresses
https://linear.app/budibase/issue/BUDI-6867/automations-dont-support-attachment-fields

## Screenshots
image

## Launchcontrol
- Adds the ability for the user to add attachments from within the Create Row / Update Row automation steps.


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -222,11 +222,9 @@ export async function streamUpload({
   const objectStore = ObjectStore(bucketName)
   const bucketCreated = await createBucketIfNotExists(objectStore, bucketName)
 
-  if (ttl && (bucketCreated.created || bucketCreated.exists)) {
+  if (ttl && bucketCreated.created) {
     let ttlConfig = bucketTTLConfig(bucketName, ttl)
-    if (objectStore.putBucketLifecycleConfiguration) {
-      await objectStore.putBucketLifecycleConfiguration(ttlConfig).promise()
-    }
+    await objectStore.putBucketLifecycleConfiguration(ttlConfig).promise()
```
PClmnt, line 227: I added this in as part of the last PR, but I got a little concerned that people could potentially update Lifecycles on an existing bucket, i'm not sure we want to do that.

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,52 @@ export function getError(err: any) {
...
+
+        if (content) {
+          await objectStore.streamUpload({
+            bucket: objectStore.ObjectStoreBuckets.APPS,
+            stream: content,
+            filename: s3Key,
+          })
+
+          return {
+            size: 10,
```
PClmnt, line 134: TODO

PClmnt:
Will get some tests in for this as well

packages/backend-core/src/objectStore/utils.ts
```diff
@@ -55,3 +60,50 @@ export const bucketTTLConfig = (
...
+): Promise<BucketedContent> {
+  const result = objectStore.extractBucketAndPath(attachment.url)
+
+  if (result === null) {
+    throw new Error("Invalid signed URL")
+  }
+
+  const { bucket, path } = result
+  const readStream = await objectStore.getReadStream(bucket, path)
+  const fallbackFilename = path.split("/").pop() || ""
```
samwho, line 89: Could we use Node's `path.basename` here instead? 🙏

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
   return typeof err !== "string" ? err.toString() : err
 }
 
+export async function sendAutomationAttachmentsToStorage(
+  tableId: string,
+  row: Row
+) {
+  let table = await sdk.tables.getTable(tableId)
```
samwho, line 105:
```suggestion
  const table = await sdk.tables.getTable(tableId)
```

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
...
+    }
+  }
+
+  for (const [prop, attachments] of Object.entries(attachmentRows)) {
+    if (attachments.length) {
+      row[prop] = await Promise.all(
+        attachments.map(attachment => generateAttachmentRow(attachment))
+      )
+    }
+  }
```
samwho, line 124: Doesn't this omit generating single attachment rows? If the schema above is `FieldType.ATTACHMNENT_SINGLE` won't `value` be a non-array, so this `attachements.length` check will fail?

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
...
+    const attachmentResult = await objectStore.processAutomationAttachment(
+      attachment
+    )
+    const extension = attachment.filename.split(".").pop() || ""
+
+    if ("bucket" in attachmentResult && "path" in attachmentResult) {
+      const { path, content, bucket } = attachmentResult
+      let client = objectStore.ObjectStore(bucket)
+
+      if (path.includes(`${prodAppId}/attachments/`)) {
```
samwho, line 145: Should this be `path.startsWith`?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -19,7 +19,7 @@ const streamPipeline = promisify(stream.pipeline)
 const STATE = {
   bucketCreationPromises: {},
 }
-const signedFilePrefix = "/files/signed"
+export const signedFilePrefix = "/files/signed"
```
mike12345567, line 22: If exporting a const like this, preference is for all upper casing name, like `SIGNED_FILE_PREFIX` - to denote a constant value rather than say a function.

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -41,7 +41,7 @@ type UploadParams = BaseUploadParams & {
 }
 
 type StreamUploadParams = BaseUploadParams & {
-  stream: ReadStream
+  stream: ReadStream | NodeJS.ReadableStream | ReadableStream<Uint8Array> | null
```
mike12345567, line 44: Is this truly `null` or is it optional like `stream?: ReadStream | NodeJS.ReadableStream | ReadableStream<Uint8Array>`

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
...
+  const prodAppId = context.getProdAppId()
+  try {
+    let size: number | null | undefined
+    let s3Key: string
+    const client = objectStore.ObjectStore(objectStore.ObjectStoreBuckets.APPS)
+
+    const attachmentResult = await objectStore.processAutomationAttachment(
+      attachment
+    )
+    const extension = attachment.filename.split(".").pop() || ""
```
samwho, line 139: Could we use `path.extname` here? 🙏

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
...
+      row[prop] = await Promise.all(
+        attachments.map(attachment => generateAttachmentRow(attachment))
+      )
+    }
+  }
+
+  return row
+}
+
+async function generateAttachmentRow(attachment: AutomationAttachment) {
```
samwho, line 129: This is a _very_ complex function to follow. If possible, it would be great to tackle some of the complexity of it. I appreciate you're going on holiday on Friday, so I don't mind it being a `// TODO` in this PR and then something we tackle in a follow-up PR.

packages/server/src/automations/tests/createRow.spec.ts
```diff
@@ -15,6 +18,14 @@ describe("test the create row action", () => {
     }
   })
 
+  beforeAll(async () => {
+    tk.reset()
```
samwho, line 22: Do we actually use `tk` in this file at all?

packages/server/src/automations/automationUtils.ts
```diff
@@ -96,6 +98,93 @@ export function getError(err: any) {
...
+      } else {
+        const processedFileName = `${uuid.v4()}.${extension}`
+        s3Key = `${prodAppId}/attachments/${processedFileName}`
+        await objectStore.streamUpload({
+          bucket: objectStore.ObjectStoreBuckets.APPS,
+          stream: content,
+          filename: s3Key,
+        })
+      }
+      const metadata = await client
```
mike12345567, line 156: Can these calls be encapsulated in the `objectStore` rather than exporting the Client itself and running this logic directly here? 

Like an `objectStore.exists(bucket, path)` would be quite useful and would avoid the need for setting up the client etc here.

packages/worker/src/utilities/email.ts
```diff
@@ -165,39 +163,16 @@ export async function sendEmail(
...
+    let attachments = await Promise.all(
+      opts.attachments?.map(objectStore.processAutomationAttachment)
     )
+    attachments = attachments.map(attachment => {
+      if ("path" in attachment) {
+        const { path, ...rest } = attachment
+        return rest
+      }
+      return attachment
+    })
```
samwho, line 175:
```suggestion
    attachments = attachments.map(attachment => {
      return _.omit(attachment, ["path"])
    })
```

https://lodash.com/docs/4.17.15#omit

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -77,7 +89,31 @@
...
+      {bindings}
+      keyBindings
+      customButtonText={"Add attachment"}
+      keyPlaceholder={"URL"}
+      valuePlaceholder={"Filename"}
+      noAddButton={schema.type === FieldType.ATTACHMENT_SINGLE &&
+        value[field].length >= 1}
+    />
+  </div>
+{:else if ["string", "number", "bigint", "barcodeqr", "array", "attachment"].includes(schema.type)}
```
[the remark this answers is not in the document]
PClmnt: Sorted the first issue, but yep we definitely should. I'll get it in another PR I just don't want to mix up any weird UX side effects that could come of doing that within this PR

packages/server/src/automations/automationUtils.ts
PClmnt: How's that look now?

packages/server/src/automations/automationUtils.ts
PClmnt: You were right, I had changed this recently and forgot to add in the other case. Sorted.

packages/server/src/automations/automationUtils.ts
PClmnt: I tried to refactor it, how does that look now?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -333,7 +329,11 @@ export function getPresignedUrl(
     const signedUrl = new URL(url)
     const path = signedUrl.pathname
     const query = signedUrl.search
-    return `${signedFilePrefix}${path}${query}`
+    if (path.startsWith(signedFilePrefix)) {
+      return `${path}${query}`
+    } else {
+      return `${signedFilePrefix}${path}${query}`
```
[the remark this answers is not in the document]
PClmnt: I removed this if statement as it wasn't actually needed anymore. So it's just the original code for that function.

packages/backend-core/src/objectStore/utils.ts
```diff
@@ -71,7 +71,7 @@ async function processUrlAttachment(
   const fallbackFilename = path.basename(new URL(attachment.url).pathname)
   return {
     filename: attachment.filename || fallbackFilename,
-    content: response.body,
+    content: response.body!,
```
samwho, line 74: How are we certain this `!` is correct? If we're not, it's best to fail early by adding a check above for `!response.body` and throwing an error.

packages/server/src/automations/automationUtils.ts
```diff
@@ -114,12 +118,15 @@ export async function sendAutomationAttachmentsToStorage(
...
-        attachments.map(attachment => generateAttachmentRow(attachment))
-      )
+    if (Array.isArray(attachments)) {
+      if (attachments.length) {
+        row[prop] = await Promise.all(
+          attachments.map(attachment => generateAttachmentRow(attachment))
+        )
+      }
+    } else if (Object.keys(row[prop]).length > 0) {
+      row[prop] = await generateAttachmentRow(attachments)
```
samwho, line 129: We probably need a test in place to prevent the near-miss we had here happening again 🙏

packages/server/src/automations/automationUtils.ts
samwho: Much nicer 🚀

packages/server/src/automations/automationUtils.ts
```diff
@@ -128,55 +135,56 @@ export async function sendAutomationAttachmentsToStorage(
...
+  }
+
+  async function getSize(s3Key: string) {
+    return (
+      (
+        await objectStore.getObjectMetadata(
+          objectStore.ObjectStoreBuckets.APPS,
+          s3Key
+        )
+      ).ContentLength || 0
```
samwho, line 162: Is 0 a correct default here? In what scenarios would `ContentLength` by falsey?

PClmnt: Honestly, I'm not sure. I guess the `S3.HeadObjectOutput` output can be undefined but I don't think we want to fail it if it can't get the size for whatever reason? What do you think?

packages/server/src/automations/automationUtils.ts
PClmnt: done!

samwho (approved):
Thanks for your patience with my feedback, this is looking good!

mike12345567 (approved):
LGTM to me as well - thanks for going through the PR comments!

PR #13569, opened 2024-04-26
Make REUSE_CONTAINERS the default for running tests.

samwho:
## Description

@PClmnt raised that he keeps running in to the following error when using the `--watch` flag to run tests locally:

```
Multiple containers found starting with image: "budibase/couchdb"
```

This was an error I put in place to prevent a situation I found where tests can spin up multiple CouchDB containers but only end up connecting to 1 of them. When you run with `--watch`, `testcontainers` never tears any containers down and so when the tests try to bring up another, it fails.

I looked to see if there's a way to detect when Jest is running in `--watch` mode but it doesn't seem trivial to do so. I had a play with making a `globalTeardown.ts` file as well but that doesn't run until you fully cancel the `--watch` tests.

So I've gone for the simple route of just making container reuse the default and getting rid of the `REUSE_CONTAINERS` environment variable.

Potential problems with this approach:

1. It will leave containers lying around when the tests have finished running, which developers may not expect.
2. These containers will use up resources, which may annoy developers.
3. The containers will slowly accumulate more data in them the more you run tests, which could consume disk space and perhaps fill up the volumes attached to each container.

I think we can cross those bridges if/when they arise, for now this solves an immediate annoyance.


PClmnt (approved):
LGTM - thanks for sorting this!

mike12345567 (approved):

PR #13573, opened 2024-04-26
Remove all traces of flatpickr

aptkingston:
## Description
This PR removes a few leftover references to flatpickr which aren't needed any more, that were missed in the previous PR implementing the new date picker.


aptkingston:
Merging so we can test on QA today.

PR #13579, opened 2024-04-30
[Fix] Schema panel null references

adrinr:
## Description
Fixing 2 small issues around queries:
1. Selecting "empty" in query schema panel caused js to crash
image

2. Selecting "empty" in rest schema panel caused js to crash
image

3. Displaying JSON (instead of nulls) for JSON fields
image

## Launchcontrol
Fixing some minor builder bugs around query schemas


aptkingston (approved):
LGTM!

Issue #13582, opened 2024-04-30
Form blocks don't expose state values whenever creating one related to a field inside the block.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.24.0
  - App Version: 2.24.0

**Describe the bug**
Form blocks don't expose state values whenever creating one related to a field inside the block.

**To Reproduce**
1.  Add a form block
2. Create a state value related to one of your fields.
3. Add a paragraph component
4. Try and reference the value of the state variable created in a paragraph component
5. See there is no state bindings available in the bindings drawer

**Expected behavior**
You should be able to access created state variables from block components specific form blocks.
  
**App Export**
[form state test](https://github.com/Budibase/budibase/files/15162109/form-state-test-1714473793022.tar.gz)

Issue #13581, opened 2024-04-30
Adding a prompt in front of any additional actions that use bindable elements appear as null when referenced


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.24.0
  - App Version: 2.24.0
**Describe the bug**

**To Reproduce**
1. Import the exported app
2. Type in a value to the first name field
3. Go through the confirmation page
4. Click confirm on the prompt
5. See no state is passed through

**Expected behavior**
You should be able to see the binding from the form field. It shouldn't be getting nulled after changing it and going through a prompt modal.=
  
**App Export**
[async test-export-1714470812456.tar.gz](https://github.com/Budibase/budibase/files/15161549/async.test-export-1714470812456.tar.gz)

Issue #13560, opened 2024-04-25
BB Table Rows Filter on Static Formula - all operators disappeared in version 2.23.12


esellier:
It seems there was a (unwanted?) change in table rows filter...
I have a column set as a static formula, it was ok in previous versions as I could use any operator for filtering (in my case a Like operator).
But in version 2.23.12 there are no more operators presented in the Builder...

Tried to revert :

* Reverted to version 2.22.19 in App Builder (Self Hosted BD configuration).
  No Effect...
* Did change the composer yaml to force BB using 2.22.19 image
  Status: Downloaded newer image for [budibase.docker.scarf.sh/budibase/apps:2.22.19](http://budibase.docker.scarf.sh/budibase/apps:2.22.19)
  No Effect...

Solved (with still the 2.22.19 version) when I changed the table, by adding (duplication) another column (duplication of a static formula column).
Maybe an issue with the table schema?

Steps that SHOULD reproduce the issue:
A/ Setup a 2.22.19 version
B/ Create table with two numbers column, and a static formula such as
\-{{ Col1 }}-{{ Col2 }}-
C/ Create screen with a DataProvider on this table
D/ Add a filter on the static formula using Like operator
E/ Upgrade to version 2.23
F/ Check if filter is ok in the DataProvider

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/2d180b66-b458-4491-bd06-55cdd13b5338/e044ec46-256e-4f01-a5eb-c429cae7b360?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8yZDE4MGI2Ni1iNDU4LTQ0OTEtYmQwNi01NWNkZDEzYjUzMzgvZTA0NGVjNDYtMjU2ZS00ZjAxLWE1ZWItYzQyOWNhZTdiMzYwIiwiaWF0IjoxNzE0MDQ3MjAwLCJleHAiOjMzMjg0NjA3MjAwfQ.3jf8VQ0C5Kcz8W8jqhgXcuAVs-w7HfaY2DR-1zmkE6w)

esellier:
I did try to update (again) to 2.23, and I have the issue again.
But this time, even adding or duplicating columns have no effect and the operators are still blanked.
[...]

Issue #13548, opened 2024-04-23
It cannot pass a row id from a table contained in the data provider.


neojigi:
my budibase version is 2.23.10 on self hosted

I have a problem with a side panel with a form block.

I set the row id setting of the formblock correctly to the key value set in the update state of the table.

Table in root --> on row click --> update state {{ Clicked row.\_id }} --> open side panel = row id pass successfully

However, if I do the same manipulation on the table under the data provider, the row id pass doesn't work and always shows the 1st id.

I would be grateful if you could let me know if there is something I am missing.

```[tasklist]
### Tasks
```

neojigi:
Clicked row._id pass do not work within sume situation.

I've been trying for many days and in many ways, but I can't get it to work. T.T

Sometimes when I click on a row in a table, the row id pass does not work.

table is in root, it works great.

But it doesn't work when the table is in the data provider.
It is clear that row id pass does not work if the table is included in the data provider, which passes the click information.



<< My config >>
os : ubuntu 22.04
type : self hosted 2.23.12
network : behind Nginx reverse proxy with SSL

<< nginx conf's location block >>
   location / {
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header X-Forwarded-Proto $scheme;
       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header Host $http_host;
       proxy_http_version 1.1;
       proxy_pass http://192.168.0.247:10000;
   }



<< successful case A - row id pass to another screen >>

* table in root
  - on row click --> (edit screen url)/{{ Clicked.row._id }}

* (edit screen)/:id
  - data provider --> ( _id filter binding = {{ URL.id }}
    - repeater
      - form
        - ...fields --> row id pass success!!!!!



<< successful case B - row id pass to side panel's form block >>

* table in root
  - on row click
    - update state : key name with {{ Clicked row._id }}
    - Open Side Panel

* Side Panel
  - Form Block
    - Row ID = {{ State.(key name) }} --> row id pass success!!!!!



<<< but... >>>

In the above 2 success cases, ROW ID PASS do not work if the source table you are clicking on is under DATA PROVIDER.

{{ Clicked row._id }} value seems to work only when the table is at root.

BennaceurHichem:
Same issue faced here.  ROW ID PASS do not work if the source table you are clicking on is under DATA PROVIDER.

Changing the data provider to Table will make the filters doesn't work at all as well.

[...]

PR #13593, opened 2024-05-02
Teal bug fixes

aptkingston:
## Description
This PR is a collection of fixes for issues discovered during building our app.

## Fixes
- Fix prompts in button action chains breaking event context for following actions
- Rewrite how state keys are detected so they will detect usage inside form steps, form blocks, and any other "nested" component style settings
- Add support for using _id and _rev of clicked rows when using a relationship as a grid datasource
- Fix static formulas having no operator options
image
- Add support for using _id and _rev of clicked rows when using a data provider as a table data source, and that data provider is using a datasource plus source

## Improvements
- Made both settings of the "Prompt user" button action bindable (they were previously just static text)
- Fixed tooltips for form step configuration controls
- Improved grid/table ability to handle more types of error messages
- Remove prettier warnings inside account portal coverage files
- Remove eslint warning about personal config files (if you have one) by add the `root` key to our config
- Refactor click_outside utils to both keep the current fix of not closing side panels when dragging outside, but also fix those annoying drag-outside closing events for good (selecting text inside a modal for example), and fix the current issue of side panels flashing when clicking between rows in grids

## Addresses
- https://github.com/Budibase/budibase/issues/13581
- https://github.com/Budibase/budibase/issues/13582
- https://github.com/Budibase/budibase/issues/13595
- https://github.com/Budibase/budibase/issues/13560
- https://github.com/Budibase/budibase/issues/13548

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FormStepConfiguration.svelte
+++ b/packages/builder/src/components/design/settings/controls/FormStepConfiguration.svelte
@@ -21,26 +21,24 @@
   const currentStep = derived(multiStepStore, state => state.currentStep)
   const componentType = "@budibase/standard-components/multistepformblockstep"
 
+  setContext("multi-step-form-block", multiStepStore)
+
   let cachedValue
   let cachedInstance = {}
 
   $: if (!isEqual(cachedValue, value)) {
     cachedValue = value
   }
-
   $: if (!isEqual(componentInstance, cachedInstance)) {
     cachedInstance = componentInstance
   }
-
-  setContext("multi-step-form-block", multiStepStore)
-
   $: stepCount = cachedValue?.length || 0
   $: updateStore(stepCount)
   $: dataSource = getDatasourceForProvider($selectedScreen, cachedInstance)
   $: emitCurrentStep($currentStep)
   $: stepLabel = getStepLabel($multiStepStore)
   $: stepDef = getDefinition(stepLabel)
-  $: stepSettings = cachedValue?.[$currentStep] || {}
+  $: savedInstance = cachedValue?.[$currentStep] || {}
   $: defaults = Utils.buildMultiStepFormBlockDefaultProps({
     _id: cachedInstance._id,
     stepCount: $multiStepStore.stepCount,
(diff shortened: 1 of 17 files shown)


packages/builder/src/components/design/settings/controls/FormStepConfiguration.svelte
```diff
@@ -92,7 +92,8 @@
   }
 
   const addStep = () => {
-    value = value.toSpliced($currentStep + 1, 0, {})
+    const newInstance = componentStore.createInstance(componentType)
+    value = value.toSpliced($currentStep + 1, 0, newInstance)
```
aptkingston, line 96: Instead of using an empty object, we now create a fake component instance for each form step. This is how we usually handle these settings, and makes it easier to use later with generic utilities because we have a real component type saved for which we can lookup the definition and settings.

packages/builder/src/dataBinding.js
aptkingston, line 1: All nested types are now handled in a single generic way. The old code handled all of these cases apart from `fieldConfiguration`, and was also very duplicated. This is a single generic way to do this now for all these types of settings.

packages/client/src/utils/buttonActions.js
```diff
@@ -542,16 +542,22 @@ export const enrichButtonActions = (actions, context) => {
...
+                  // Enrich and call the next button action if there is more
+                  // than one action remaining
                   const next = enrichButtonActions(
                     actions.slice(i + 1),
                     newContext
                   )
-                  resolve(typeof next === "function" ? await next() : true)
+                  if (typeof next === "function") {
+                    // Pass the event context back into the new action chain
+                    resolve(await next(eventContext))
```
aptkingston, line 557: We were just missing the event context param here to the following button actions. The rest is just moving around some code to make it more readable.

packages/frontend-core/src/components/FilterBuilder.svelte
```diff
@@ -124,6 +124,7 @@
     const fieldSchema = schemaFields.find(x => x.name === filter.field)
     filter.type = fieldSchema?.type
     filter.subtype = fieldSchema?.subtype
+    filter.formulaType = fieldSchema?.formulaType
```
aptkingston, line 127: The formula type was missing from the structure we pass into the lucene utils in shared core, which resulted in us being unable to detect static formulae.

deanhannigan (approved):
LGTM! So many great fixes in here 🚀

PR #13594, opened 2024-05-02
Fix user relationship to user column migration

adrinr:
## Description
Use new user column type when migrating existing user relationships.
Also fixing some issues around the migrated types

## Screenshots
Migrating a table with the following relationships to user...
image

will result as
image

### DIFF
--- a/packages/builder/src/constants/backend/index.js
+++ b/packages/builder/src/constants/backend/index.js
@@ -165,20 +165,11 @@ export const FIELDS = {
       BBReferenceFieldSubType.USER
     ],
   },
-  // Used for display of editing existing columns
-  DEPRECATED_USER: {
-    name: "User",
-    type: FieldType.BB_REFERENCE,
-    subtype: BBReferenceFieldSubType.USER,
-    icon: TypeIconMap[FieldType.BB_REFERENCE_SINGLE][
-      BBReferenceFieldSubType.USER
-    ],
-  },
   USERS: {
     name: "User List",
     type: FieldType.BB_REFERENCE,
-    subtype: BBReferenceFieldSubType.USERS,
-    icon: TypeIconMap[FieldType.BB_REFERENCE][BBReferenceFieldSubType.USERS],
+    subtype: BBReferenceFieldSubType.USER,
+    icon: TypeIconMap[FieldType.BB_REFERENCE][BBReferenceFieldSubType.USER],
     constraints: {
       type: "array",
     },
--- a/packages/server/src/api/controllers/row/utils/basic.ts
+++ b/packages/server/src/api/controllers/row/utils/basic.ts
@@ -1,5 +1,5 @@
 // need to handle table name + field or just field, depending on if relationships used
-import { FieldType, Row, Table } from "@budibase/types"
+import { BBReferenceFieldSubType, FieldType, Row, Table } from "@budibase/types"
 import { generateRowIdField } from "../../../../integrations/utils"
 import { CONSTANT_INTERNAL_ROW_COLS } from "../../../../db/utils"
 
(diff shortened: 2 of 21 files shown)


packages/builder/src/constants/backend/index.js
```diff
@@ -165,20 +165,11 @@ export const FIELDS = {
...
-    subtype: BBReferenceFieldSubType.USER,
-    icon: TypeIconMap[FieldType.BB_REFERENCE_SINGLE][
-      BBReferenceFieldSubType.USER
-    ],
-  },
   USERS: {
     name: "User List",
     type: FieldType.BB_REFERENCE,
-    subtype: BBReferenceFieldSubType.USERS,
-    icon: TypeIconMap[FieldType.BB_REFERENCE][BBReferenceFieldSubType.USERS],
```
samwho, line 181: I'm not sure I understand these changes. Don't we need to keep this old field?

packages/server/src/api/controllers/row/utils/basic.ts
```diff
@@ -111,8 +111,15 @@ export function fixArrayTypes(row: Row, table: Table) {
       try {
         row[fieldName] = JSON.parse(row[fieldName])
       } catch (err) {
-        // couldn't convert back to array, ignore
-        delete row[fieldName]
+        // Handling deprecated  single user type
+        const isDeprecatedSingleUser =
+          schema.type === FieldType.BB_REFERENCE &&
+          schema.subtype === BBReferenceFieldSubType.USER &&
+          schema.constraints?.type !== "array"
```
samwho, line 118: I've seen this check in a couple of places, do we want to extract it out into `shared-core`?

packages/server/src/sdk/app/tables/migration.ts
```diff
@@ -24,25 +25,58 @@ export interface MigrationResult {
...
+
+  const type =
+    oldColumn.relationshipType === RelationshipType.ONE_TO_MANY
+      ? FieldType.BB_REFERENCE_SINGLE
+      : FieldType.BB_REFERENCE
+  const newColumn: FieldSchema = {
+    name: newColumnName,
+    type,
+    subtype: BBReferenceFieldSubType.USER,
+  }
```
samwho, line 68: When I made this API originally the idea was to get the caller to specify the full FieldSchema for the new column, because it kept the API open to doing other types of migration in the future.

I'm not sure I follow why exactly it's being changed away from that. Won't this change make it difficult to extend this in future?

packages/builder/src/constants/backend/index.js
adrinr: This was added to be able to display the old columns properly. This might be confusing as it's not clearly used anywhere (because it's consumed dynamically). Instead, I handled these cases in the actually usage (in `packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte`

packages/server/src/sdk/app/tables/migration.ts
adrinr: If we follow this way, we need the frontend to know about the backend inside-outs. I made these changes in order to keep it as a backend responsibility (depending on the new column type). This also makes it easier to test, as otherwise we would be accepting anything from the frontend

packages/server/src/api/controllers/row/utils/basic.ts
adrinr: I actually thought about this. Refactoring it!

packages/server/src/sdk/app/tables/migration.ts
samwho: That's fair, I did kinda wonder if I was making it more complicated than it needed to be in order to fit use-cases that didn't exist yet.

samwho (approved):
I'm good with this, but would recommend also getting insight from @mike12345567 given that it's quite a fiddly change.

mike12345567 (approved):
LGTM to me as well!

Issue #13598, opened 2024-05-02
New Date Picker component - dates/days incorrect


andyburgessmd:
**Checklist**
I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.24.0
  - App Version: 2.24.0

**Describe the bug**
The updated date picker component is showing the dates matching with the wrong days of the week.

**To Reproduce**
1. Add a date picker component to a page
2. View the date picker in the builder, preview or published app
3. Compare todays date to the day of the week shown

**Expected behavior**
It's expected that the date matched the day of the week. For example 2nd May 2024 is a Thursday. 

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/110813989/5562b94c-5cec-4c9d-9b54-7706683f3209)
Screenshot shows 2nd May 2024 is a Friday. Today (Thursday) is the 2nd May.

aptkingston:
Thanks @andyburgessmd. Deploying a fix very shortly.

aptkingston:
That's released and fixed now 👍

PR #13599, opened 2024-05-02
Fix calendar using locale aware start of week

aptkingston:
## Description
Uses the ISO start of week (Monday) rather than a locale aware start of week (which can be Sunday).

Before (2nd is a Friday, incorrectly):
image

After (2nd is a Thursday, correctly): 
image

## Addresses
- https://github.com/Budibase/budibase/issues/13598


deanhannigan (approved):
LGTM!

PR #13606, opened 2024-05-03
Make auto ID row creation in parallel more robust.

samwho:
## Description

I've noticed a few times in tests that if you create rows with auto ID columns in parallel, you quite often get `409 Conflict` errors.

The reason this happens is because we store the next ID for a new row on the table document itself. When you create a new row, it takes that ID, increments it, saves it back, then uses it in the row. When creating rows in parallel, this read-save-use process is quite likely to conflict.

To make this more robust, I've introduced a function called `getNextAutoId` that will attempt to increment the next ID on the table and save it. If it gets a conflict, it will try again up to 5 times. When it succeeds, the next ID is returned. If it doesn't succeed in 5 attempts an error is thrown.

Because this is attempted 5 times and is done per-column, it makes it much less likely for a 409 to be returned to the user.

### DIFF
--- a/packages/server/src/api/controllers/row/staticFormula.ts
+++ b/packages/server/src/api/controllers/row/staticFormula.ts
@@ -1,20 +1,11 @@
 import { getRowParams } from "../../../db/utils"
 import {
   outputProcessing,
-  processAutoColumn,
   processFormulas,
 } from "../../../utilities/rowProcessor"
-import { context, locks } from "@budibase/backend-core"
-import {
-  Table,
-  Row,
-  LockType,
-  LockName,
-  FormulaType,
-  FieldType,
-} from "@budibase/types"
+import { context } from "@budibase/backend-core"
+import { Table, Row, FormulaType, FieldType } from "@budibase/types"
 import * as linkRows from "../../../db/linkedRows"
-import sdk from "../../../sdk"
 import isEqual from "lodash/isEqual"
 import { cloneDeep } from "lodash/fp"
 
@@ -151,30 +142,7 @@ export async function finaliseRow(
   // if another row has been written since processing this will
   // handle the auto ID clash
   if (oldTable && !isEqual(oldTable, table)) {
-    try {
-      await db.put(table)
-    } catch (err: any) {
-      if (err.status === 409) {
-        // Some conflicts with the autocolumns occurred, we need to refetch the table and recalculate
-        await locks.doWithLock(
-          {
-            type: LockType.AUTO_EXTEND,
-            name: LockName.PROCESS_AUTO_COLUMNS,
-            resource: table._id,
-          },
-          async () => {
-            const latestTable = await sdk.tables.getTable(table._id!)
-            let response = processAutoColumn(null, latestTable, row, {
-              reprocessing: true,
-            })
-            await db.put(response.table)
-            row = response.row
-          }
-        )
-      } else {
-        throw err
-      }
-    }
+    await db.put(table)
   }
   const response = await db.put(row)
   // for response, calculate the formulas for the enriched row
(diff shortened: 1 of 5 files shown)


packages/server/src/api/controllers/row/staticFormula.ts
```diff
@@ -151,30 +142,7 @@ export async function finaliseRow(
...
-          },
-          async () => {
-            const latestTable = await sdk.tables.getTable(table._id!)
-            let response = processAutoColumn(null, latestTable, row, {
-              reprocessing: true,
-            })
-            await db.put(response.table)
-            row = response.row
-          }
-        )
```
samwho, line 173: This locking is no longer required, and didn't work anyway. You'd need to wrap the entire row creation in this lock for it to consistently prevent 409s, and that could be prohibitively expensive.

packages/server/src/sdk/app/rows/tests/internal.spec.ts
```diff
@@ -145,6 +145,7 @@ describe("sdk >> rows >> internal", () => {
                 lastID: 1,
               },
             },
+            _rev: expect.stringMatching("2-.*"),
```
samwho, line 148: Because we're saving the table and updating the `_rev` on it in our new `getNextAutoId` function, this needed adding to this test.

packages/server/src/sdk/app/rows/tests/internal.spec.ts
```diff
@@ -199,7 +199,7 @@ describe("sdk >> rows >> internal", () => {
           await internalSdk.save(table._id!, row, config.getUser()._id)
         }
         await Promise.all(
-          makeRows(10).map(row =>
+          makeRows(20).map(row =>
```
samwho, line 202: Decided to up the number of rows to be a bit more sure we're avoiding conflicts.

packages/server/src/sdk/app/rows/tests/internal.spec.ts
```diff
@@ -189,7 +190,6 @@ describe("sdk >> rows >> internal", () => {
             type: FieldType.AUTO,
             subtype: AutoFieldSubType.AUTO_ID,
             autocolumn: true,
-            lastID: 0,
```
samwho, line 192: Decided to remove this because the code handles it being `undefined`. May as well exercise that.

packages/types/src/documents/app/table/schema.ts
```diff
@@ -219,3 +219,15 @@ export function isAttachmentField(
 ): field is AttachmentFieldMetadata {
   return field.type === FieldType.ATTACHMENTS
 }
+
+export function isAutoColumnField(
+  field: FieldSchema
+): field is AutoColumnFieldMetadata {
```
adrinr, line 225: We should not actually need these wards. If you are doing `field.type === FieldType.AUTO`, the types `AutoColumnFieldMetadata` should be inferred on usage

packages/server/src/api/controllers/row/staticFormula.ts
adrinr: This is a good cleanup :)

packages/server/src/api/controllers/row/staticFormula.ts
```diff
@@ -151,30 +142,7 @@ export async function finaliseRow(
...
-      } else {
-        throw err
-      }
-    }
+    await db.put(table)
```
adrinr, line 145: Do we actually need this? Are we not persisting the table already for the autoid?

samwho: Not necessarily. If the table doesn't have an auto ID column it won't have been persisted.

packages/types/src/documents/app/table/schema.ts
samwho: 🙈 I feel silly.

adrinr (approved):

PR #13609, opened 2024-05-03
Add free trial modal

melohagan:
## Description
When a user is onboarded, they will be presented with a free trial modal. After clicking `Get started` the modal will no longer appear.

**Should be merged in after the related account portal changes have been merged**.

## Addresses
- https://linear.app/budibase/issue/GROW-417/changes-in-the-ui-to-let-the-users-know-we-have-added-a-trial

## Screenshots
Screenshot 2024-05-03 at 14 55 51

account

https://github.com/Budibase/budibase/assets/101575380/18475fbc-f03a-4c2c-a4bd-40074f05ae08

## Launchcontrol
Free trial modal - automatically enrolled on to an Enterprise free trial when creating a new account


packages/builder/src/pages/builder/app/[application]/_layout.svelte
```diff
@@ -102,6 +107,15 @@
     }
   }
 
+  const showFreeTrialModal = (planType, freeTrialModal) => {
+    if (
+      planType === "enterprise_basic_trial" &&
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/_layout.svelte:112]: We can move this const to types and to use it here

packages/builder/src/pages/builder/app/[application]/_layout.svelte
```diff
@@ -192,6 +206,25 @@
...
+  <FreeTrialModal
+    onConfirm={async () => {
+      if (get(auth).user) {
+        try {
+          await API.updateSelf({
+            freeTrialConfirmedAt: new Date().toISOString(),
+          })
+          // Update the cached user
+          await auth.getSelf()
+        } catch {
```
Comment (samwho) [inline packages/builder/src/pages/builder/app/[application]/_layout.svelte]: nit: you could remove this `catch`

jvcalderon (approved):

packages/builder/src/pages/builder/app/[application]/_layout.svelte
```diff
@@ -102,6 +107,15 @@
     }
   }
 
+  const showFreeTrialModal = (planType, freeTrialModal) => {
+    if (
+      planType === "enterprise_basic_trial" &&
+      !$auth.user?.freeTrialConfirmedAt
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/_layout.svelte:113]: Will this not spam the modal until confirmed?

packages/builder/src/pages/builder/app/[application]/_layout.svelte
```diff
@@ -102,6 +107,15 @@
     }
   }
 
+  const showFreeTrialModal = (planType, freeTrialModal) => {
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/_layout.svelte:110]: To encapsulate logic, this should be part of the modal itself, or to have a wrapper around it. We should have the actual `_layout` as lean as possible

adrinr (changes requested):

samwho (approved):

packages/builder/src/pages/builder/app/[application]/_layout.svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/_layout.svelte:113]: The modal has this property: `disableCancel={true}` which means the user must confirm the modal before they can interact with the app.

With that said, if the API call fails for any reason, you would really want the modal to close and stay closed. Might just put in an additional local variable to account for that scenario

Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/_layout.svelte:113]: Although `$: showFreeTrialModal(planType, freeTrialModal)` will only trigger initially, so I think it's alright to stay as is.

packages/builder/src/pages/builder/app/[application]/_layout.svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/_layout.svelte]: using plan type

packages/builder/src/pages/builder/app/[application]/_layout.svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/app/[application]/_layout.svelte]: Refactored

packages/frontend-core/src/constants.js
```diff
@@ -57,6 +57,7 @@ export const PlanType = {
   PRO: "pro",
   BUSINESS: "business",
   ENTERPRISE: "enterprise",
+  ENTERPRISE_BASIC_TRIAL: "enterprise_basic_trial",
```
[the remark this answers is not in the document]
adrinr: If this is the case, we should rename the `showFreeTrialModal` references. The current code is not unified

adrinr (approved):

adrinr (commented):
Looks good, I added few comments

PR #13610, opened 2024-05-03
Table column initial width setting

aptkingston:
## Description
This PR adds an initial width setting to columns in tables. We had a width setting with the old table, but we didn't add it for the new table because
- You can drag to resize columns as you like
- The widths default to those in the data section, so should already be roughly correct

There are still valid use cases for having unique widths per table instance, so this PR restores that width setting. This setting is an "initial width" setting rather than an explicit width - that is to say that users can still resize columns as they like, but this setting will change the initial widths to those specified by the app builder.

image

## Launchcontrol
Add initial width setting to table columns.


packages/builder/src/components/design/settings/controls/DraggableList/DraggableList.svelte
```diff
@@ -100,9 +100,6 @@
       on:click={() => {
         get(store).actions.select(draggableItem.id)
       }}
-      on:mousedown={() => {
-        get(store).actions.select()
-      }}
```
aptkingston, line 105: Not needed - this was just closing the previous popover on mouse down which isn't how we usually handle popovers so I'm removing this to keep things consistent with just clicks.

packages/builder/src/components/design/settings/controls/EditComponentPopover/index.js
aptkingston, line 1: This custom position handler is no longer needed, as my updated logic for the usual position_dropdown util already handles this. I haven't specifically checked usage with the tours yet but I will.

deanhannigan: I had a quick look at the tour stuff there and they are starting but just not visible

deanhannigan (changes requested):
Some minor issues but the changes are solid!

`Bug` - Context menu for the Table now floating to the top in the builder preview

aptkingston:
> Some minor issues but the changes are solid!
> 
> `Bug` - Context menu for the Table now floating to the top in the builder preview 

Fixed 👌 Was actually caused by a previous PR but I've addressed it here since it was easy enough.

deanhannigan (approved):
LGTM! The updates to the grid are solid! After a run through, I can see the context menu is back in place and no other issues remain 👍

PR #13614, opened 2024-05-03
Tests for searching boolean fields, including fixes for various bugs.

samwho:
## Description

This PR introduces tests for searching against boolean fields, which appear to have been quite broken until now. The PR includes fixes for searching against `false` values in Lucene, and a variety of fixes required to get searching working for the SQL-based datasources.


packages/server/src/api/controllers/row/utils/utils.ts
```diff
@@ -117,6 +117,19 @@ export async function validate(
...
+    if (col.type === FieldType.BOOLEAN) {
+      if (row[col.name] === 1) {
+        row[col.name] = true
+      } else if (row[col.name] === 0) {
+        row[col.name] = false
+      }
+    }
+  }
+  return row
+}
```
samwho, line 131: SQS, MariaDB, and MySQL appear to store boolean values as numeric 1s or 0s, this function parses those back into proper booleans.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -67,6 +67,22 @@ describe.each([
...
+        // that are present in the expected row.
+        const searchedObjects = foundRows.map(row => _.pick(row, fields))
+        throw new Error(
+          `Failed to find row: ${JSON.stringify(
+            expectedRow
+          )} in ${JSON.stringify(searchedObjects)}`
+        )
+      }
+      return row
+    }
```
samwho, line 84: Added this so that when tests fail because the row searched for does not match any of the rows returned, the message is _much_ nicer. It used to complain about passing `undefined` to `objectContaining`, now it looks like this:

CleanShot 2024-05-03 at 17 35 21@2x

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -164,8 +164,13 @@ export async function search(
...
 
-      let sql = query.sql,
-        bindings = query.bindings
+      let sql = query.sql
+      let bindings = query.bindings?.map(b => {
+        if (typeof b === "boolean") {
+          return b ? 1 : 0
+        }
+        return b
+      })
```
samwho, line 173: SQLite has no boolean type, and so the SQS API does not accept boolean bindings. So we convert them to 1s and 0s.

packages/backend-core/src/db/lucene.ts
```diff
@@ -282,8 +282,7 @@ export class QueryBuilder<T> {
     }
 
     const equal = (key: string, value: any) => {
-      // 0 evaluates to false, which means we would return all rows if we don't check it
-      if (!value && value !== 0) {
+      if (value === null || value === undefined) {
```
mike12345567, line 285: The `!value` here likely also includes `""` which we really don't want to allow through - its an empty state which isn't super well defined, but the API has always treated it as ignored so we should hold this; it might be helpful to have a function which checks if its an empty state, but not zero or false.

packages/server/src/sdk/app/rows/search/sqs.ts
mike12345567: I'm wondering if we should move this up a layer in some way - if we bake this into the SQL generation rather than the search API, it means if we use SQS anywhere else we don't have to worry about this.

mike12345567 (approved):
LGTM! Thanks for the `isEmpty` improvement!

PR #13636, opened 2024-05-08
Fix for positioning of selected builder navigation tab on page load

deanhannigan:
## Description

The underline on the builder tab navigation would periodically get stuck in the wrong position when the page loaded. This was caused by the header rendering before the fonts had finished loading.

## Addresses
- Adds a font load event to the builder layout. As fonts load, a record is pushed to the builder store. 
- As the `fonts` prop is updated, the tabs can use it to trigger a re-render and position the underline correctly.

## Screenshots
The issue
fonts

## Launchcontrol

Ensure the builder tabs redraw as the fonts load, ensuring the current tab is correctly underlined.


Ghrehh (approved):
LGTM

packages/builder/src/pages/builder/app/[application]/_layout.svelte
```diff
@@ -144,17 +148,21 @@
...
-                quiet
-                selected={$isActive(path)}
-                on:click={topItemNavigate(path)}
-                title={capitalise(title)}
-                id={`builder-${title}-tab`}
-              />
-            </TourWrap>
-          {/each}
+          {#key $builderStore?.fonts}
+            <span style="display:contents">
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/_layout.svelte:152]: NAB: Does a <svelte:fragment> work here instead? Or why do we need to insert another DOM node - do we not have other instances of #each blocks inside #key? Or I guess we could move the #key outside the entire Tabs component too.

aptkingston (approved):

PR #13637, opened 2024-05-08
Remove aws sdk global mock and update tests

PClmnt:
## Description
Removes the global mock from the server and updates associated tests to work correctly.


packages/server/src/integrations/tests/aws-sdk.mock.ts
PClmnt, line 1: Had to add this in here for the s3.spec.ts and dynamodb.spec.ts as we can't use the minio provider for those. We'd need to pull test containers specifically for them and probably out of scope for this

samwho (approved):
Love this.

PR #13644, opened 2024-05-09
Move parallel auto ID row creation test to row.spec.ts.

samwho:
## Description

The original purpose of this PR was to move the tests in `src/sdk/app/rows/tests/internal.spec.ts` into `packages/server/src/api/routes/tests/row.spec.ts`. The latter is a comprehensive suite of functional tests, and the former is just a smattering of unit tests.

However, in doing this it was revealed that creating rows with auto ID columns actually ends up updating datasource entries for external datasources. This creates a huge number of document conflict errors in the rest of the tests, also signalling that this would be quite problematic in production as well. Fortunately, as of right now, you can't create auto ID columns on external datasources. We're going to prevent that and not support it going forward.

The code I wrote in https://github.com/Budibase/budibase/pull/13606 actually turns out to not be more robust, either. If anything, it appeared to make it _more_ likely to give you document conflict errors. For that reason, I've removed it in this PR and gone back to making retrying document conflict errors the responsibility of the API caller.


mike12345567 (approved):
Very nice - LGTM!

Issue #13586, opened 2024-05-02
Altering a external data source's primary key causes the _id and _rev number to become disabled - allow budibase to notify the user when their DB schema has deviated from the one currently stored


ConorWebb96:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.24.0
  * App Version: 2.24.0

**Describe the bug**
When changing the primary key of an external data table e.g. Postgres via an alter statement the `_id` and `_rev` fields are grayed out. This happens in the data area, when trying to copy either of these values from a row.

**To Reproduce**

1. Go to any external table
2. Alter the primary key with an SQL statement
3. Go back to the table
4. See error, see that the fields are grayed out and can't be copied

**Expected behavior**
You should be able to copy the `_id` and` _rev` even after a table SQL alter on the primary key has happened.

**Screenshots**

![Screenshot 2024-05-02 at 08 54 18](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/f83fbeaf-0e0b-4d4c-ae94-1bbcf922eb34/d1c4017e-4050-4f76-8989-a388ed61ca5a?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC9mODNmYmVhZi0wZTBiLTRkNGMtYWU5NC0xYmJjZjkyMmViMzQvZDFjNDAxN2UtNDA1MC00Zjc2LTg5ODktYTM4OGVkNjFjYTVhIiwiaWF0IjoxNzE1MDczODUyLCJleHAiOjMzMjg1NjMzODUyfQ.lshEADuA-tl14XtL3RFAnkzyReO6i5dXh9BeyLJ_syk "=1155x")

**SQL schema**

```sql
CREATE TABLE "public"."timezone" (
    "id" int4 NOT NULL DEFAULT nextval('timezone_id_seq'::regclass),
    "name" varchar(255) NOT NULL,
    "created_at" timestamptz NOT NULL DEFAULT now(),
    "updated_at" timestamptz NOT NULL DEFAULT now(),
    "User" text,
    "birthday" timestamptz,
    PRIMARY KEY ("id")
);
 
-- change primary key after initial create.
[...]

PR #13650, opened 2024-05-09
Fixing issue with grid not allowing copying external ID

mike12345567:
## Description
Fixing an issue with ID/Rev not being copyable from context menu due to the way IDs were split/handled in the grid.

External rows can have row IDs which contain `-` characters, this is used in the grid store to attach the field which is currently focused, this can be avoided by assuming the last split off `-` will leave the ID as the first element and the field as the second element.

## Addresses
- (https://github.com/Budibase/budibase/issues/13586)


mike12345567:
@aptkingston would appreciate your input on this - after some discussion with @samwho realised that we just can't use `-` for joining the ID and field name, as it could be present in either of them, so splitting is very difficult.

Instead using something very unusual to avoid this problem - `‽‽`

aptkingston (approved):
LGTM! 

Technically these are cell IDs rather than row IDs so a rename would be appreciated not definitely not required.

So purely for terminology consistency (since this is only for the grid codebase and we may as well use the same terms) I would opt for
```js
const { rowId, columnName } = parseCellID(cellId)
const cellId = getCellID(rowId, columnName)
```

But that's just me being overly anal about naming. Functionally this looks perfect and is much more robust and maintainable. Meant to do this myself at some stage rather than have the duplicated logic all over the place. Nice one!

PR #13652, opened 2024-05-09
Deleting a column on google spreadsheet messes with the data

adrinr:
## Description
Currently, deleting a column deletes the header cell but not the full column. Because of this, deleting data mixes the order of the columns. We don't want to delete the data as these data might not be created by budibase. Because of this, we will only drop the BB reference but keeping the references in the actual sheet. Adding the deleted column back will backpopulate it in BB.

Also fixing a bug that allowed spamming the `create/edit column` submit, causing strange errors.
https://github.com/Budibase/budibase/assets/15987277/f6a6f5b5-08ec-4c99-b27e-5dc888fb49db

## Addresses
- [[BUDI-8222] Deleting a column on google spreadsheet messes with the data](https://linear.app/budibase/issue/BUDI-8222/deleting-a-column-on-google-spreadsheet-messes-with-the-data)

## App Export

## Launchcontrol
Fixing broken data on google sheet column deletion


packages/server/src/integrations/tests/googlesheets.spec.ts
```diff
@@ -129,10 +129,11 @@ describe("Google Sheets Integration", () => {
       })
       expect(sheet.loadHeaderRow).toHaveBeenCalledTimes(1)
       expect(sheet.setHeaderRow).toHaveBeenCalledTimes(1)
-      expect(sheet.setHeaderRow).toHaveBeenCalledWith(["name"])
-
-      // No undefined are sent
-      expect((sheet.setHeaderRow as any).mock.calls[0][0]).toHaveLength(1)
```
adrinr, line 135: Not sure what this was meant to do, as it was already asserted on the previous lines

samwho (approved):
Took me a little while to get my head around why this worked but I see it now. Great PR, love the typing improvements.

mike12345567 (approved):

PR #13658, opened 2024-05-10
Allow Fancy Input validation to be triggered onBlur

melohagan:
## Description
Part of the work to improve onboarding email address validation: https://github.com/Budibase/account-portal/pull/612

If `validateOn` is set to **blur**, then the validation error message will only trigger on blur, however an existing error message can still be resolved on change.

I.e.

1. I enter an invalid email address (no error message)
2. Blur
3. Error message is shown under the field
4. I change the input to be valid -> the error message goes away (resolved)
5. I then enter an invalid address (no error message, back to step 1)


jvcalderon (approved):
LGTM!

Ghrehh (approved):

PR #13672, opened 2024-05-13
Fix null dereference error when selected cell ID is undefined

aptkingston:
## Description
Fixes a small bug due to the recent cell ID parsing changes which threw a null dereferencing error when there was no selected cell ID. This PR just updates the fallback behaviour of the parsing to return an object of the proper shape rather than `undefined`.


shogunpurple (approved):
LGTM

PR #13676, opened 2024-05-13
Fixes for update row action in automations and attachments

mike12345567:
## Description
Fixing some issues with updating rows with attachments, there were some UI bugs that had been raised (undefined being displayed) and the uploading from a URL was not working as expected, due to the stream not matching the expected format of the S3 SDK - this has been fixed by converting the stream to a true readable stream with `stream.Readable.fromWeb()`.

These issues were raised by @aptkingston and Avalanche.

### DIFF
--- a/packages/backend-core/src/objectStore/utils.ts
+++ b/packages/backend-core/src/objectStore/utils.ts
@@ -9,6 +9,9 @@ import {
   AutomationAttachmentContent,
   BucketedContent,
 } from "@budibase/types"
+import stream from "stream"
+import streamWeb from "node:stream/web"
+
 /****************************************************
  *      NOTE: When adding a new bucket - name       *
  *     sure that S3 usages (like budibase-infra)    *
@@ -53,12 +56,10 @@ export const bucketTTLConfig = (
     Rules: [lifecycleRule],
   }
 
-  const params = {
+  return {
     Bucket: bucketName,
     LifecycleConfiguration: lifecycleConfiguration,
   }
-
-  return params
 }
 
 async function processUrlAttachment(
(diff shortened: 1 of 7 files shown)


packages/backend-core/src/objectStore/utils.ts
```diff
@@ -71,7 +72,7 @@ async function processUrlAttachment(
   const fallbackFilename = path.basename(new URL(attachment.url).pathname)
   return {
     filename: attachment.filename || fallbackFilename,
-    content: response.body,
+    content: stream.Readable.fromWeb(response.body as streamWeb.ReadableStream),
```
adrinr, line 75: Will this always be available?

packages/server/src/automations/automationUtils.ts
```diff
@@ -163,7 +163,7 @@ async function generateAttachmentRow(attachment: AutomationAttachment) {
 
   try {
     const { filename } = attachment
-    const extension = path.extname(filename)
+    const extension = path.extname(filename).replaceAll(".", "")
```
adrinr, line 166: Is this expected? Do we want to trim? What will happen with complex extensions? `fe, .tar.gz`

packages/server/src/automations/automationUtils.ts
```diff
@@ -183,7 +183,7 @@ async function generateAttachmentRow(attachment: AutomationAttachment) {
     return {
       size,
       name: filename,
-      extension,
+      extension: extension,
```
adrinr, line 186: NIT: not needed

adrinr (commented):
Is is worth adding some tests?

packages/backend-core/src/objectStore/utils.ts
mike12345567: Good point - in theory there always should be a response body, but it might not always be a stream.

PR #13682, opened 2024-05-14
Fix schema generation for external tables with view_ in their name

aptkingston:
## Description
If external tables had the string `view_` in their name then we wrongly assumed they were views, as we use external table names are part of their IDs. This PR ensures we only check if IDs start with `view_` rather than contain it.

## Addresses
- https://github.com/Budibase/budibase/issues/13680


andz-bb (approved):
LGTM

Issue #13677, opened 2024-05-13
Percentage Styling not working on Table Component


jmulaaaa:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: Docker
  - Budibase Version: 2.26.1
  - App Version: 2.26.1

**Describe the bug**
When styling the table component, using a percentage does not work, and the table goes to the same size no matter the percentage.
I had a working app that used this styling in version 2.22.18. Now it no longer works, but was able to replicate the issue on a brand new app as well.
**To Reproduce**
1. Create App
2. Add Table Component
3. Change Height to a percentage

**Expected behavior**
For the percentage option to work.
I used to have my tables set to 100% to fill the container they were in, now all of my tables are the same height everywhere.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome

PR #13683, opened 2024-05-14
Rework table height to support percentages again

aptkingston:
## Description
Recent changes to the table component meant that using percentages as a height no longer worked. This was because we were not apply height to the outer DOM node, but instead applying it to the inner grid component itself, breaking relativity to the parent of the table component as the user sees it. Those changes were motivated by a desire to ensure that the min height of the table is respected, ensuring users can't create a table that's too small to be used.

This PR reworks how height is calculated and applied so that percentages work again, but we also still use the correct min-height.

Example of a table with a height of 50%, showing the parent container to highlight that it's working:
image

## Addresses
- https://github.com/Budibase/budibase/issues/13677


packages/client/src/components/app/GridBlock.svelte
```diff
@@ -50,8 +51,7 @@
       metadata: { dataSource: table },
     },
   ]
-  $: height = $component.styles?.normal?.height || "408px"
-  $: styles = getSanitisedStyles($component.styles)
+  $: gridContext?.minHeight?.subscribe($height => (minHeight = $height))
```
aptkingston, line 55: The core grid is the source of truth and makes min height available in context, which we can consume from here and honour.

packages/frontend-core/src/components/grid/layout/Grid.svelte
```diff
@@ -109,8 +109,13 @@
     notifyError,
     buttons,
   })
-  $: minHeight =
-    Padding + SmallRowHeight + $rowHeight + (showControls ? ControlsHeight : 0)
+
+  // Derive min height and make available in context
+  const minHeight = derived(rowHeight, $height => {
```
aptkingston, line 116: Small change to make this a derived store rather than a primitive, so it's observable.

deanhannigan (approved):
LGTM!

packages/client/src/components/app/GridBlock.svelte
Ghrehh: Does unsubscribe ever need to be called on this?

Would storing minHeight as reactive variable and then assigning it to minHeight work?

```
$: contextMinHeight = gridContext.minHeight;
$: minHeight = $contextMinHeight
```

Obviously more verbose, but if the unsubscribe is necessary might be worth it.

Very much NAB anyway.

Ghrehh (approved):
LGTM

packages/client/src/components/app/GridBlock.svelte
aptkingston: Going to move this into the onMount handler to be more explicit about the fact it'll only run once. Makes sense to be explicit about it.

PR #13687, opened 2024-05-14
Update naming of some data types and update order

aptkingston:
## Description
This PR updates the names and orders of data types in the data section. Both are easily readable from the code diff.

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -398,44 +398,50 @@
     if (!externalTable) {
       return [
         FIELDS.STRING,
-        FIELDS.BARCODEQR,
-        FIELDS.LONGFORM,
+        FIELDS.NUMBER,
         FIELDS.OPTIONS,
         FIELDS.ARRAY,
-        FIELDS.NUMBER,
-        FIELDS.BIGINT,
         FIELDS.BOOLEAN,
         FIELDS.DATETIME,
-        FIELDS.ATTACHMENT_SINGLE,
-        FIELDS.ATTACHMENTS,
         FIELDS.LINK,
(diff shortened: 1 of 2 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -398,44 +398,50 @@
...
+        FIELDS.BARCODEQR,
         FIELDS.BIGINT,
-        FIELDS.USER,
       ]
 
-      if (datasource && datasource.source !== SourceName.GOOGLE_SHEETS) {
-        fields.push(FIELDS.USERS)
+      // Filter out multiple users for google sheets
+      if (datasource?.source === SourceName.GOOGLE_SHEETS) {
+        fields = fields.filter(x => x !== FIELDS.USERS)
```
aptkingston, line 437: I've inverted the logic here in order to be able to keep a custom order. By default all types are added then we filter out types that we don't want, rather than appending types that we *do* want to the end of the list.

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -398,44 +398,50 @@
...
       }
-      // no-sql or a spreadsheet
-      if (!externalTable || table.sql) {
-        fields = [...fields, FIELDS.LINK, FIELDS.ARRAY]
+
+      // Filter out SQL-specific types for non-SQL datasources
+      if (!table.sql) {
+        fields = fields.filter(x => x !== FIELDS.LINK && x !== FIELDS.ARRAY)
```
aptkingston, line 442: I've inverted the logic here in order to be able to keep a custom order. By default all types are added then we filter out types that we don't want, rather than appending types that we do want to the end of the list.

mike12345567 (approved):

Issue #13684, opened 2024-05-14
user must be of type string


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.26.1
  - App Version: 2.26.1

**Describe the bug**
This is something to do with the new user changes added recently.
I have a form with a user field which on submit saves a new row with the user and a date. When submitted it now shows the error 'user must be of type string' and does not save the new row.
![image](https://github.com/Budibase/budibase/assets/110813989/4fd297f3-3fc6-4647-8e04-aeb94d37bfd7)
I have tied to revert to 2.24.2 and it is still present.

**To Reproduce**
Steps to reproduce the behavior: This may not be reproducible when creating new - it could be linked to the change being incompatible.
1. Create a table with a user field
![image](https://github.com/Budibase/budibase/assets/110813989/54cda1f4-5c02-456f-af80-50e676a8b911)
2. Create a new form with a user component
![image](https://github.com/Budibase/budibase/assets/110813989/32585db0-1c63-4975-a291-c368b186fef7)
3. Action on the save button 
![image](https://github.com/Budibase/budibase/assets/110813989/262a74fa-be62-4be6-ad67-8e0529926186)
4. Select the user and save the new row

mike12345567:
Hi @andyburgessmd - we have not been able to reproduce this - is it possible you could provide an app export which exhibits this behaviour?

andyburgessmd:
Yes, but can I email it to you instead of sharing on public?

mike12345567:
Hi @andyburgessmd - yes if you send it to support@budibase.com referencing this issue we will take a look.

andyburgessmd:
@mike12345567 on its way now

mike12345567:
Hi @andyburgessmd - this should be fixed in the latest release.

PR #13688, opened 2024-05-14
Fix - user must be of type string

adrinr:
## Description
Some old deprecated user columns had some constrains, forcing the data to be strings. With the recent changes, we were storing them as simple arrays, and this is conflicting with its constrains.

## Addresses
- https://github.com/Budibase/budibase/issues/13684
- [BUDI-8251 - user must be of type string](https://linear.app/budibase/issue/BUDI-8251/user-must-be-of-type-string)

## Launchcontrol
Fixing a bug causing save issues with old deprecated single columns


packages/server/src/sdk/app/rows/search.ts
```diff
@@ -80,7 +80,7 @@ export async function search(
 
   const table = await sdk.tables.getTable(options.tableId)
   options = searchInputMapping(table, options, {
-    isSql: !!table.sql || !!env.SQS_SEARCH_ENABLE,
```
mike12345567, line 83: Why the removal of this?

mike12345567 (approved):
LGTM

mike12345567:
Addresses: https://github.com/Budibase/budibase/issues/13684

mike12345567 (approved):
LGTM - nice improvements, making the bulk import API better!

adrinr:
During these changes, we realised that bulk import (used for tests) was not treating the input row with functions like bbprocessor. This was causing some tests to not work as expected

PR #13689, opened 2024-05-14
SQS cleanup (tables and apps)

mike12345567:
## Description
SQS automatically syncs data from CouchDB that comes via the changes stream, however when tables are deleted or when an app database is deleted it is not fully aware of this.

Keeping this information will not cause any search problems, but would use extra disk space that could be freed up. This solves the problem by making sure when a DB is removed that the on disk status is cleaned up, as well as making sure when a table is deleted it is removed from the SQLite design schema.

Sadly this is quite hard to test as SQS was already aware that rows/DB doesn't exist anymore, so won't respond to the search with them, the only way to confirm that this has worked as expected is by confirming the SQLite databases have been removed from the disk. I've checked/confirmed this and it does appear to be working as expected.

## Addresses
- https://linear.app/budibase/issue/BUDI-7658/sqs-creatingdeleting-tables-when-addedremoved-from-budibase


packages/backend-core/src/db/Replication.ts
```diff
@@ -1,14 +1,34 @@
...
+      target.startsWith(DocumentType.APP)
+    ) {
+      this.direction = ReplicationDirection.TO_PRODUCTION
+    } else if (
+      source.startsWith(DocumentType.APP) &&
+      target.startsWith(DocumentType.APP_DEV)
+    ) {
+      this.direction = ReplicationDirection.TO_DEV
+    } else {
+      this.direction = ReplicationDirection.UNKNOWN
```
adrinr, line 30: Is there a use case for this? Should we allow it?

packages/server/src/api/controllers/table/utils.ts
```diff
@@ -496,5 +497,31 @@ export function setStaticSchemas(datasource: Datasource, table: Table) {
   return table
 }
 
+export async function internalTableCleanup(table: Table, rows?: Row[]) {
+  const db = context.getAppDB()
+  const tableId = table._id!
+  // remove table search index
+  if (!env.isTest() || env.COUCH_DB_URL) {
```
adrinr, line 504: Shouldn't this be checking if there sqs is enabled, instead of `env.COUCH_DB_URL`?

adrinr: What is the reason why this is excluded from tests?

packages/backend-core/src/db/instrumentation.ts
```diff
@@ -62,7 +62,13 @@ export class DDInstrumentedDatabase implements Database {
   ): Promise<DocumentDestroyResponse> {
     return tracer.trace("db.remove", span => {
       span?.addTags({ db_name: this.name, doc_id: id })
-      return this.db.remove(id, rev)
+      if (typeof id === "object") {
```
adrinr, line 65: We could extract the `id` and `rev` from the object, and to check if they are correct. Otherwise, passing an empty object will never complain

packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -281,25 +284,59 @@ export class DatabaseImpl implements Database {
...
+    return await this._sqlQuery<T[]>(url, "POST", {
+      query: sql,
+      args: parameters,
     })
-    if (response.status > 300) {
-      throw new Error(await response.text())
+  }
+
+  // checks design document is accurate (cleans up tables)
+  async sqlCleanup(): Promise<void> {
```
adrinr, line 326: Could we rename this to something more verbose? Not sure what this is doing

packages/backend-core/src/db/Replication.ts
mike12345567: This is valid - because you can be syncing two unknown databases. I'll change this however to make this more obvious by allowing the direction to be undefined.

adrinr (approved):
🚀 SQS!

PR #13692, opened 2024-05-15
Fix: Only show free trial banner/modal to account owner

melohagan:
## Description
Fixes a bug were a non-account owner admin would see the trial banner, which caused issues. This has been replaced to only show to the root admin (account owner).

## Launchcontrol
Bug fix for free trial banner/modal


jvcalderon (approved):
LGTM

PR #13694, opened 2024-05-15
Handle invalid content-dispositions, still allow REST download

mike12345567:
## Description
Adding a hack incase any API we are speaking to does not abide by the expected structure of the content-disposition header, fill in the missing type to allow parsing.

Raised by @andz-bb due to the Carbone API returning a `Content-Disposition` header of `filename="..."` rather than the expected format `disposition-type; filename="...."`.

The `yarn.lock` appears to have massively changed due to a `yarn.lock` being merged without pro/account-portal locks in it (submodule not checked out).


adrinr (approved):
This will not handle use cases where there is a `;` on the file name, but I guess it's pretty unlikely

PR #13696, opened 2024-05-15
remove file size limit on self host

shogunpurple:
## Description
Remove the file size upload limit on self host for builder and client apps. Self hosted users control their own hardware so it doesn't make sense for us to impose an artificial limit.

## Launchcontrol
- Remove the file size upload limit for attachments for self hosted users.


packages/frontend-core/src/components/grid/cells/AttachmentCell.svelte
```diff
@@ -12,7 +12,7 @@
   export let schema
   export let maximum
 
-  const { API, notifications } = getContext("grid")
+  const { API, notifications, isCloud } = getContext("grid")
```
aptkingston, line 15: This will need to be `props` - we store all grid props in a top level `props` store in grid context, but memoize certain props (see the `config` store) and provide those top level for performance so that we can react to them in isolation, but we don't need to do that here.

packages/frontend-core/src/components/grid/cells/AttachmentCell.svelte
```diff
@@ -106,7 +106,7 @@
         on:change={e => onChange(e.detail)}
         maximum={maximum || schema.constraints?.length?.maximum}
         {processFiles}
-        {handleFileTooLarge}
+        handleFileTooLarge={isCloud ? handleFileTooLarge : null}
```
aptkingston, line 109: This will just need to be `$props.isCloud` after updating the import from context above.

aptkingston (commented):
One tiny change needed for the grid but otherwise looks good!

packages/builder/src/components/common/Dropzone.spec.js
aptkingston, line 1: Nice work adding test coverage 🙏

packages/frontend-core/src/components/grid/cells/AttachmentCell.svelte
shogunpurple: good catch, updated!

aptkingston (approved):
LGTM!

PR #13712, opened 2024-05-17
Added flag to ensure onboarding tour isn't cancelled when builder tabs are destroyed.

deanhannigan:
## Description
Ensures the onboarding tour associated with the builder tabs is not killed when the tabs are re-rendered. A [previous fix](https://github.com/Budibase/budibase/pull/13636) for the builder tabs had the unintended effect of ending the onboarding tour. 

Typically when a tour popover target is destroyed, we want the tour to end. We do not require this behaviour for the builder tabs.

- `endTourOnDestroy` flag added to `TourWrap` to allow an exception for the builder tabs. Now when they refresh to accomodate fonts loading, the tour will remain active.

## Launchcontrol
Fix to keep the onboarding tour active when the builder tabs are re-rendered.

### DIFF
--- a/packages/builder/src/components/portal/onboarding/TourWrap.svelte
+++ b/packages/builder/src/components/portal/onboarding/TourWrap.svelte
@@ -5,6 +5,7 @@
   import { builderStore } from "stores/builder"
 
   export let stepKeys = []
+  export let endTourOnDestroy = true
 
   let ready = false
   let registered = {}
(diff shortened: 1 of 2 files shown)


aptkingston (commented):
Unrelated to this change, but I'm just realising that the much easier way to solve the initial problem of fonts not being available is to add `rel="preload"` to the link tags in our core index.html. Then we wouldn't need any custom logic to handle font loading, and we wouldn't need to remount the tabs, meaning we wouldn't need changes here either. Let me try that and see what happens before going ahead with this 👍

PR #13714, opened 2024-05-17
Unable to load grid when using external postgres

adrinr:
## Description
This PR tackles the following points:
1. Support `time with time zone`/`time without time zone` PG columns
2. Fix time-only formatting (storing time format instead of timestamp ISO formats)
3. Using native `input type="time"` in the frontend, to support native features and simplifying the code
4. Fixing timezone issues that was incrementing time wrongly

## Addresses
- [BUDI-8195 - Budibase Unable to Load Grid When Using External Postgres Database with Time Column](https://linear.app/budibase/issue/BUDI-8195/budibase-unable-to-load-grid-when-using-external-postgres-database)

### Empty datetime picker
image

### Filled datetime picker
image

### Empty time picker
image

### Filled time picker
image

## Launchcontrol
Fix time-only column issues


packages/bbui/src/Form/Core/DatePicker/TimePicker.svelte
```diff
@@ -1,46 +1,27 @@
...
   const dispatch = createEventDispatcher()
 
-  $: displayValue = value || dayjs()
+  $: displayValue = value?.format("HH:mm")
 
-  const handleHourChange = e => {
-    dispatch("change", displayValue.hour(parseInt(e.target.value)))
+  const handleChange = e => {
+    const [hour, minute] = e.target.value.split(":").map(x => parseInt(x))
+    dispatch("change", value.hour(hour).minute(minute))
```
aptkingston, line 13: Value can be `null` here (throws an error if you try to change the time when no date is selected), so we'll need to just do something similar to `(value || dayjs())` like we used to have.

aptkingston (commented):
Nice changes! Makes sense to use the native time input - that was an oversight on my part.

I get this ugly black time icon in Chromium though, and only 1 dash after the colon:

image

I wonder if we can improve the look at all?

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -966,6 +966,149 @@ describe.each([
     })
   })
 
+  !isInternal &&
+    describe("datetime - time only", () => {
+      const T_1000 = "10:00"
```
samwho, line 971: https://youtu.be/rXziCock3cY?si=a21znV1jfg3mKI-w&t=120

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -966,6 +966,149 @@ describe.each([
...
+
+      const UNEXISTING_TIME = "10:01"
+
+      beforeAll(async () => {
+        await createTable({
+          time: { name: "time", type: FieldType.DATETIME, timeOnly: true },
+        })
+
+        await createRows(
+          _.shuffle([T_1000, T_1045, T_1200, T_1530, T_0000]).map(time => ({
```
samwho, line 985: I like the shuffle here. I should do this in the rest of the tests.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -966,6 +966,149 @@ describe.each([
...
+      const T_1045 = "10:45"
+      const T_1200 = "12:00"
+      const T_1530 = "15:30"
+      const T_0000 = "00:00"
+
+      const UNEXISTING_TIME = "10:01"
+
+      beforeAll(async () => {
+        await createTable({
+          time: { name: "time", type: FieldType.DATETIME, timeOnly: true },
```
samwho, line 981: Given the trouble we've had with nulls, I think it would be worth adding a row with a null time to make sure it works as expected.

packages/server/src/integrations/base/sql.ts
```diff
@@ -122,18 +122,22 @@ function generateSelectStatement(
...
           `"${tableName}"."${columnName}"::money::numeric as "${field}"`
         )
       }
     }
+    if (
+      knex.client.config.client === SqlClient.MS_SQL &&
+      columnSchema.type === FieldType.DATETIME &&
+      columnSchema.timeOnly
+    ) {
+      return knex.raw(`CONVERT(varchar, ${field}, 108) as "${field}"`)
```
samwho, line 139: It would be worth commenting here that "108" is a "style code" that tells SQL Server to format as `hh:mm:ss`, to save the next person looking it up.

samwho (approved):
Approving backend half of this, with a few small comments.

packages/bbui/src/helpers.js
```diff
@@ -166,7 +166,7 @@ export const stringifyDate = (
   const offsetForTimezone = (enableTime && ignoreTimezones) || timeOnly
   if (offsetForTimezone) {
     // Ensure we use the correct offset for the date
-    const referenceDate = timeOnly ? new Date() : value.toDate()
+    const referenceDate = value.toDate()
```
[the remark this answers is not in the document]
adrinr: I am at gmt+2 and without this, every time I called this function, one extra hour was added to the result. I will try to write unit tests to ensure its behaviour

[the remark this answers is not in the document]
aptkingston: Was this when using a real time type in a postgres DB? This would probably happen if the backing datasource behaved differently than we currently expect compared to using a non-timezone aware timestamp type. I'm not sure a unit test will help here because I'm almost certain it's to do with what the value looks like when it gets saved into the external DB and then returned to the frontend.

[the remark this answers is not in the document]
aptkingston: This is probably just more evidence that a "time only" data type really needs to be its own thing, rather than trying to store it inside an ISO date string.

packages/server/src/api/routes/tests/search.spec.ts
samwho: That said, https://github.com/Budibase/budibase/pull/13728 fixes a bunch of null problems so you may wish to wait until that goes in first.

packages/bbui/src/Form/Core/DatePicker/TimePicker.svelte
adrinr: This was not hit for some reason, but I added the check just in case. The existing behaviour treats default dates as empty and time as `now`. This seems slightly confusing, also, deleting the time value with the `backspace` key sets the value to 0, and the `clear` button clears the date but sets the time to `now` again.
The updated behaviour sets both the date and the time to undefined, so time will be `--:--` (HTML input standard). `Clear` button and `backspace` will remove any time value, getting `--:--` back.

Do you thing it's better to keep the existing behaviour? If so, I can just tweak it to behave like that

aptkingston: Clearing it is more correct so that's good. The old behaviour didn't actually set the time to now when cleared, but just always displayed something (mostly to avoid this issue of trying to mutate a null time). I'm happy to keep the behaviour of actually clearing the field, as long as it's possible to edit the time from empty state without throwing an error. I'll give this another test now after your changes 👍

aptkingston: This still throws an error when the value is null - we need to just update the change handler to use a new dayjs instance whenever we type into a time field that's empty.
image
image
Your original code here was totally fine, but just needed updated to `(value || dayjs()).hour...` like I mentioned.

packages/server/src/api/routes/tests/search.spec.ts
adrinr: I actually changed it in order to be used everywhere

packages/server/src/api/routes/tests/search.spec.ts
adrinr: Done. I actually needed to fix some stuff related to pg 🤷🏽

packages/bbui/src/helpers.js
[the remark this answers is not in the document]
adrinr: I agree that a time only is required. Also in the backend, having the same type with different data shape in the db is also pretty hard to manage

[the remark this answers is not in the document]
adrinr: The issue happened fetching time columns from dbs or "creating" internal db timeOnly fields. Running a deeper testing

packages/bbui/src/Form/Core/DatePicker/TimePicker.svelte
adrinr: Good catch, thx!

samwho (approved):

aptkingston (approved):
LGTM! I'm happy to leave the offset timezone changes in if that fixed an issue for you. There is probably some scenario that works with the old code, but hopefully usage of time-only is low and people won't run into issues.

Eventually we can hopefully move to a proper type for times and avoid that altogether anyway 👍

PR #13724, opened 2024-05-20
Observe size changes in tabs

aptkingston:
## Description
This PR improves tabs by allowing them observe size changes, rather than the mixture of hacks and workarounds we were attempting to use to get around this issue. Observing size changes with a real ResizeObserver is a "pure" and proper solution to this problem.

Quick video of this working:

https://github.com/Budibase/budibase/assets/9075550/1877b1ec-eea5-420a-b3d6-3645c359b691

Fixes the initial issue addressed in https://github.com/Budibase/budibase/pull/13636, and also undoes those changes as they are no longer needed. Also replaces the need for https://github.com/Budibase/budibase/pull/13712.

### DIFF
--- a/packages/bbui/src/Tabs/Tab.svelte
+++ b/packages/bbui/src/Tabs/Tab.svelte
@@ -1,40 +1,28 @@
 <script>
-  import { getContext, onMount, createEventDispatcher } from "svelte"
+  import { getContext, onDestroy } from "svelte"
   import Portal from "svelte-portal"
 
   export let title
   export let icon = ""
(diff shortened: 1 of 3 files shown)


aptkingston:
@deanhannigan could you test and see if this fixes the original issue? I still can't repro the initial thing reliably, even with throttling the browser, but I know you could. Thank you 🙏

deanhannigan (approved):
LGTM! The tabs are resizing on load and the tour popovers are still able to load

Issue #13722, opened 2024-05-19
Options-picker and mutli-select picker not closing when clickling on opened picker


pmeier123:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: aachenemperors

**Describe the bug**
After updating my App from Version 2.23.12 to 2.26.4 one new bug in options-picker handling appears. I cannot close the options-picker or multi-select picker by clicking on the control itself. Was working in 2.23.12. not working anymore in 2.26.4. makes the handling of multi-select pickers very nasty because i have to click anywhere else in the form to close the picker.

**To Reproduce**
1. Go to an options picker
2. Click on options picker to enter a value
3. Try closing the picker by clicking the picker
4. Error: Picker does not get closed

**Expected behavior**
When clicking the opened picker control again, the picker should close. As working in 2.23.12

**Screenshots**

https://github.com/Budibase/budibase/assets/55078905/77ba2849-2096-46c4-9e68-50b2c39af8ba

https://github.com/Budibase/budibase/assets/55078905/bb9e3ed6-8bc4-4e55-aed5-ce928cc9556f

  
**App Export**
I can give you when neccessary.

**Desktop (please complete the following information):**
 - OS: Windows, iOS
 - Browser Chrome
 - Version 125.0.6422.61 (Offizieller Build) (64-Bit)

pmeier123:
Just got fixed

pmeier123:
@deanhannigan @aptkingston thanks for the quick fix!
is there a planned date for a bugfix release including this fix?

pmeier123:
fixed in 2.27.2

aptkingston:
Thanks for updating this yourself @pmeier123. We released last night which fixed this as you noticed :)

PR #13725, opened 2024-05-20
Allow closing options pickers by clicking again

aptkingston:
## Description
Small update to the click_outside utility to fix an issue where you can no longer click on a picker again to close it. The reason this broke was that my previous update to this utility changed a `click` handler to a `mouseup` handler as the even that actually triggers the click_outside behaviour.

By no longer using a click, the existing overrides we had in place (for things like picker) no longer worked since they weren't stopping propagation anymore. By using a click handler instead, we can use these overrides again.

The click_outside utility now uses an initial mousedown event, then uses a click event after that, comparing the DOM targets of each to ensure that we haven't changed node (preventing issues like things closing when selecting text and dragging outside).

## Addresses
- https://github.com/Budibase/budibase/issues/13722


deanhannigan (approved):
LGTM!

PR #13728, opened 2024-05-20
Search tests for user relationship columns.

samwho:
## Description

This PR adds tests and fixes around searching user relationship columns (single and multi).

In doing this work I discovered there were problems around querying against null values, and this appears to be a problem outside of SQS.

Prior to this PR, imagine you had the following rows:

```js
[{ user: null }, { user: "us_1" }, { user: "us_2" }]
```

And you did the following search:

```js
{ notEqual: { user: "us_1 } }
```

You would expect to get back `{ user: null }` and `{ user: "us_2" }`, however you would only get back `{ user: "us_2" }`.

The reason for this is that in SQL, if an expression evaluates to `null` then that `null` often propagates, and the negation of `null` is still `null`. So lots of the conditions we were building would be of the form `WHERE <expr>` and `WHERE NOT <expr>` but if `<expr>` is `null`, you get `null` in both situations and never see the rows with `null` values.

To get around this, I have liberally applied the `COALESCE` function to our clauses (except for SQL Server, where I had to use `CASE` because its `COALESCE` cannot accept boolean expressions as arguments for some reason.)

## Launchcontrol

This fixes columns with a `null` value not being returned as they should be in negation queries like `not equal` or `not contains`.


packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -172,7 +172,8 @@ export async function search(
       sql = sql.replace(/`doc2`.`rowId`/g, "`doc2.rowId`")
 
       const db = context.getAppDB()
-      return await db.sql<Row>(sql, bindings)
+      const rows = await db.sql<Row>(sql, bindings)
```
adrinr, line 175: ❤️  finally somebody adding this instead of removing mines haha

adrinr (approved):

packages/server/src/sdk/app/rows/search/sqs.ts
samwho: It made placing a breakpoint easier 😅

PR #13734, opened 2024-05-21
Making progress towards deleting sql.spec.ts.

samwho:
## Description

The main aim of this PR is to delete `sql.spec.ts`, replacing the things it's testing with integration tests in either `search.spec.ts`, `row.spec.ts`, or `table.spec.ts`.

I haven't achieved that, but I have managed to shorten it so it has much less overlap with existing integration tests.

On my way through this, I renamed `external.spec.ts` to `search.spec.ts` and upgraded it to a generic test against all datasources for our search SDK. Its job is to test whatever can't be tested through the API.

I felt I'd hit a good checkpoint, but would like to follow this up with another PR further trimming down what's in `sql.spec.ts`.


packages/server/src/integrations/tests/sql.spec.ts
```diff
@@ -271,260 +164,6 @@ describe("SQL query builder", () => {
...
-      })
-    )
-    expect(query).toEqual({
-      bindings: [10],
-      sql: `select * from (select * from \`${TABLE_NAME}\` where JSON_OVERLAPS(${TABLE_NAME}.age, '[20,25]') and JSON_OVERLAPS(${TABLE_NAME}.name, '["John","Mary"]') limit ?) as \`${TABLE_NAME}\``,
-    })
-  })
-
-  it("should use ?| operator expression for PostgreSQL when filter is containsAny", () => {
-    const query = new Sql(SqlClient.POSTGRES, 10)._query(
```
mike12345567, line 512: does the search tests have a way of confirming thi?

packages/server/src/integrations/tests/sql.spec.ts
```diff
@@ -551,42 +190,6 @@ describe("SQL query builder", () => {
     })
   })
 
-  it("should handle table names with dashes when performing a LIKE in MySQL", () => {
-    const tableName = "Table-Name-With-Dashes"
```
mike12345567, line 555: Is there a test case for this as well already?

packages/server/src/integrations/tests/sql.spec.ts
```diff
@@ -625,51 +228,6 @@ describe("SQL query builder", () => {
     })
   })
 
-  it("should lowercase the values for Oracle LIKE statements", () => {
-    let query = new Sql(SqlClient.ORACLE, limit)._query(
```
mike12345567, line 629: We don't test oracle today anywhere else as its very difficult - this test case likely isn't covered.

packages/server/src/integrations/tests/sql.spec.ts
```diff
@@ -709,99 +267,4 @@ describe("SQL query builder", () => {
       sql: `insert into "test" ("name") values ($1) returning *`,
     })
   })
-
-  it("should be able to rename column for MySQL", () => {
-    const table: Table = {
```
mike12345567, line 714: Is this covered as part of the table tests now?

packages/server/src/integrations/tests/sql.spec.ts
samwho: Yes, we use UUIDs a lot for table names elsewhere.

packages/server/src/integrations/tests/sql.spec.ts
samwho: Not specifically for the exact SQL used but there is a test for contains in Postgres.

packages/server/src/integrations/tests/sql.spec.ts
samwho: Good catch! Will reinstate this.

packages/server/src/integrations/tests/sql.spec.ts
mike12345567: Would just make sure we have `containsAny` test cases somewhere before removing this - as its a weird case that does come with edge cases from time to time.

packages/server/src/integrations/tests/sql.spec.ts
samwho: Okay looked closer and we do have a test for this, `"updates all the row fields for a table when a schema key is renamed"` in `table.spec.ts`

mike12345567 (approved):

PR #13735, opened 2024-05-21
Audit log SQS

mike12345567:
## Description
Updating the audit log search to use an SQS index instead of the lucene index it used to use. This looks like quite a big PR, however most of the work is moving the SQL generation into `backend-core` so it can be shared across services. I've only moved the minimum which was required for this to work now - which is really just our query JSON conversion to SQL, the `sql.ts`. 

This requires making `knex` is dependency of backend-core, although it'll be included as an external dependency, which meant needing to update the build script.

This PR also introduces pagination for audit logs and the search endpoint. There might be a slightly smarter way to do this, but I wanted to merge working version first so that it can be tested before attempting to do something a little smarter (thinking something in backend-core which gets pagination info based on a passed in `QueryJson`).

Tests have been updated for audit logs to test against both Lucene and SQL based searching - the support for this is enabled by setting the environment variable `SQS_SEARCH_ENABLE` on the worker service.

This also introduces the change for the development team to move to our development SQS image - with both Lucene and SQS. This does not mean the development team will be using SQS, but will install the image and we'll see if there is any compatibility issues.


packages/types/src/documents/global/auditLogs.ts
```diff
@@ -15,5 +16,6 @@ export interface AuditLogDoc extends Document {
   timestamp: string
   metadata: any
   name: string
+  type?: "auditLog"
```
adrinr, line 19: Should this be optional?

adrinr (commented):
Looks good, just a couple comments before I do a final deeper review

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -185,22 +195,45 @@ export async function search(
...
+      const response: SearchResponse<Row> = {
+        rows: finalRows,
+      }
+      const prevLimit = request.paginate!.limit
+      request.paginate = {
+        limit: 1,
+        page: bookmark * prevLimit + 1,
+      }
+      // check if there is another row
+      const nextRow = await runSqlQuery(request, allTables)
```
samwho, line 221: I get you don't want to fuck with this too much, but it'd be better to initially query for `limit + 1` then discard the last row, and set `hasNextPage = true` when that happens.

packages/types/src/sdk/search.ts
```diff
@@ -19,6 +19,9 @@ export enum SearchFilterOperator {
 
 export interface SearchFilters {
   allOr?: boolean
+  // TODO: this is just around for now - we need a better way to do or/and
+  // allows just fuzzy to be or - all the fuzzy/like parameters
+  fuzzyOr?: boolean
```
samwho, line 24: Why is this necessary?

mike12345567: Trying to create a where statement that allows searching over a variety of fields - discussed this the other day, wish there was a method to specify which fields are or'ed and which fields are and'ed together - but right now you can only signal that you want everything to be OR or you want everything to be AND.

The case for audit logs I was trying to achieve was something like `name LIKE %something% or metadata LIKE %something%` but I still need other props (like some other filter) to be AND on the end. Adding this property allows you to create a really fuzzy search across a bunch of different properties, similar to what we were doing with our Lucene search, essentially achieving feature parity with how the audit log search worked.

Again ideally in the future we would build into our search filters the ability to specific which fields are logically combined together, but its outside the scope of this PR and adding a flag to allow it provided a simple solution until this problem is properly solved.

Arguably Lucene was actually better for this purpose than SQL, but this achieves very similar results (enough that any lack of parity isn't noticeable, its just a free text search box used).

packages/types/src/documents/global/auditLogs.ts
mike12345567: Its optional because existing audit logs don't have it - they need to be migrated to have it added. In future it *shouldn't* be optional - but currently all audit logs will be missing this.

I was unsure how to represent this in the types, whether it was better to display what it *should* be, or display what the reality is.

packages/backend-core/src/sql/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export function convertRowId(field: any) {
+  if (Array.isArray(field)) {
+    return field[0]
+  }
+  if (typeof field === "string" && field.match(ROW_ID_REGEX) != null) {
+    return field.substring(1, field.length - 1)
+  }
+  return field
```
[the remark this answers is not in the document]
mike12345567: all of these functions are *very* old - I've simply lifted and shifted them from the server into backend core. I believe this functions goal is to try to convert it to the correct format, but I'd be worried to change what it does now as its used in a lot of core places and has worked the way it is for quite a while.

packages/types/src/documents/global/auditLogs.ts
samwho: I feel like it should always reflect reality, not the ideal. If we go with the ideal we'll end up introducing bugs.

packages/backend-core/src/sql/utils.ts
```diff
@@ -8,7 +8,7 @@ const ROW_ID_REGEX = /^\[.*]$/g
 const ENCODED_SPACE = encodeURIComponent(" ")
 
 export function isExternalTableID(tableId: string) {
-  return tableId.includes(DocumentType.DATASOURCE)
+  return tableId.startsWith(DocumentType.DATASOURCE + SEPARATOR)
```
samwho, line 11: Is this _definitely_ correct? Seems it would be a pretty big problem if not.

samwho (approved):

packages/backend-core/src/sql/utils.ts
mike12345567: Did this in response to: https://github.com/Budibase/budibase/pull/13735#discussion_r1611382875 - table IDs come from here: https://github.com/Budibase/budibase/blob/11f3d9b946df4bc924205e7d767f7f34f3e63b0a/packages/server/src/integrations/utils/utils.ts#L173

The datasource ID will always start with `datasource_` - if you look at the document type that is true for DS+ and DS, all external tables will always have that at the start - internal tables will always start with `ta_`.

PR #13736, opened 2024-05-21
Added support for single user column type in automation block

deanhannigan:
## Description
Single user column support was missing from the `LinkedRowSelector` picker. This meant that the field would be listed in the potential columns for an automation update but the UI picker would not be rendered. 

Both Single and multi user columns should work as expected when creating or updating rows with automations

## Screenshots

The issue
Screenshot 2024-05-21 at 15 09 53

## Launchcontrol

Add single user column support to automations.


aptkingston (approved):
You can't give a before screenshot and not an after screenshot! Nevertheless this LGTM haha.

Issue #13653, opened 2024-05-09
Navbar links never open in a new tab


mohagan9:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: rspb.budibase.app

**Describe the bug**
Links in the navbar do not open in a new tab even when I hold down the Cmd key.
This wasn't ideal because I was filling out a form, and wanted to open a different screen in a new tab that would provide me additional information I needed to complete the form; instead I just lost my progress. 

**To Reproduce**
1. Add navigation links to the navbar
2. Hold down the Cmd key and click on a navigation link
3. The current page changed instead

**Expected behavior**
When using navigation links within a Budibase app, they should behave like hyperlinks on a webpage, i.e. when holding the Cmd key, the link should open in a new tab.≈

**Screenshots**
<img width="616" alt="Screenshot 2024-05-09 at 20 56 00" src="https://github.com/Budibase/budibase/assets/30199063/ac04bab0-3b61-4775-a72b-530c545d819f">
*Also note that initially there is no default radio option selected - a minor UI issue*

<img width="237" alt="Screenshot 2024-05-09 at 20 57 49" src="https://github.com/Budibase/budibase/assets/30199063/65b1af51-e180-485a-949a-965ca14ee5d8">

*When I am on the outings screen, I cannot open the members screen in a new tab, which is inconvenient.*

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Brave

**Additional context**
I could add a Link component to the Members screen to workaround this, but I don't want to! 🙈

andz-bb:
great first issue @mohagan9! many thanks for your contribution 😊

Issue #13237, opened 2024-03-12
Javascript function placeholder does not disappear after Javascript function is deleted


joebudi:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Feature branch for Iive eval

**Describe the bug**
I removed a javascript binding, clicked save, and noticed the Javascript Function placeholder was still present within the data input (settings panel for Heading). This led me to believe the function was not deleted so I checked it, and it was actually deleted but the placeholder persisted. I tried to remove it (using the settings panel), but it wasn't possible. 

I later found out, that to remove it, you have to open the bindings drawer, click text, and then discard Javascript.

It would be good if the placeholder was removed when there was no Javascript function.

PR #13746, opened 2024-05-22
Misc fixes

aptkingston:
## Description
Various fixes.

- Duplicating a screen now adds it to the navigation
- The "Navigate to" button action will still work now even if you forget to prefix the URL with a /
- App navigation links function more like normal links and can be opened in new tabs
- Clearing a JS expression will set the field back to empty, rather than still showing "JavaScript function" placeholder with an empty expression

## Addresses
- https://linear.app/budibase/issue/BUDI-8130/duplicating-a-screen-does-not-update-navigation
- https://linear.app/budibase/issue/BUDI-6730/navigate-to-button-action-internal-navigation-requires-a-slash
- https://github.com/Budibase/budibase/issues/13653
- https://github.com/Budibase/budibase/issues/13237

### DIFF
--- a/packages/builder/src/components/common/bindings/BindingPanel.svelte
+++ b/packages/builder/src/components/common/bindings/BindingPanel.svelte
@@ -186,8 +186,6 @@
   }
 
   const updateValue = val => {
-    console.log("val", val)
-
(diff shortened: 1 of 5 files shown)


deanhannigan (approved):
LGTM! Some really nice UX updates! 🚀

Issue #13729, opened 2024-05-20
Postgres enum columns unexpectedly breaking.


Rdonnohue1:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.26.4
  - App Version:  2.26.4.

**Describe the bug**
Postgres enums are non-responsive after fetching tables. 

**To Reproduce**
Not sure how to reproduce. Combination of defining a relationship, removing a relationship and maybe importing an application.

**Expected behavior**
The columns to behave how they originally did, where they provided a dropdown list of selectable items.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/87651769/227162dc-9de3-4ee5-aafa-736a429c182b)
![image](https://github.com/Budibase/budibase/assets/87651769/6fe51379-8cd4-4007-a6ec-ab16d0ed8419)

{
  "browser": {
    "language": "en-US",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36",
    "platform": "Win32",
    "vendor": "Google Inc."
  },
  "server": {
    "budibaseVersion": "2.26.4",
    "hosting": "docker-compose",
    "nodeVersion": "v20.13.1",
    "platform": "linux",
    "cpuArch": "x64",
    "cpuCores": 2,
    "cpuInfo": "QEMU Virtual CPU version 2.5+",
    "totalMemory": "7.371601104736328GB",
    "uptime": "255 day(s), 23 hour(s), 37 minute(s)"
  }
}

PR #13756, opened 2024-05-23
Enum columns unexpectedly breaking fix

ConorWebb96:
## Description
- Fixed issue with enums not using the correct type for MySQL and Postgres datasources.
- Added tests to check that these fields are typed correctly in the future for these datasources.

## Addresses
- https://github.com/Budibase/budibase/issues/13729

## Launchcontrol
Fixed issue with MySQL and Postgres datasources that cause a specific field type to not show up correctly within the Budibase UI.


samwho (approved):
Great tests, superb PR.

PR #13757, opened 2024-05-23
Grid UX improvements

aptkingston:
## Description
- Increase picker max height before scrolling to 360 across the board (up from 240). More space for data types etc.
- Open grid column edit popover when double clicking on header cells
- New UI for editing options for select types
- Consistent colours between option list and colours in grids
- Actually assign colours to new options
- Validation to prevent duplicate option names
- Improved automatic naming of options
- Fix being able to use grid keyboard shortcuts while editing a column
- Fix tab/arrow keys not working to navigate cells in the new row component
- Use normal dropzones rather than compact ones when editing rows in modals
- Validates existing option sets and prevents saving dupes or empty options

image

image

image

## Addresses
- https://linear.app/budibase/issue/BUDI-8275/lost-focus-on-grid-column-edition
- https://linear.app/budibase/issue/BUDI-8277/grid-tab-key-while-adding-a-new-row
- https://linear.app/budibase/issue/BUDI-8176/grid-ux-improvements (Partly)


deanhannigan (approved):
LGTM!

Issue #13711, opened 2024-05-17
Icon list not showing for table buttons


NicolasBio:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: <method>docker compose 
  - Budibase Version: <version>2.26.4 
  - App Version: <version>2.26.4 

**Describe the bug**

When i want to select a icon for a button on a table the popout were you can select an icon is to small to display an icons. 
This possibly is only apparent on certain screen resolutions.
The icon is also not showing on the button i think?

**To Reproduce**
1. On table component add a button and click on cogwheel.
2. Click on 'pick icon'
3. Observe small popout to select icon were icons do not show.

**Expected behavior**
The icon popout should have enough room to show icons.

**Screenshots**

![image](https://github.com/Budibase/budibase/assets/153834226/64cecd48-24ae-4333-b472-ac062b4718a6)

**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox(126.0) or Egde(124.0.2478.109) 
 - Version: see above

NicolasBio:
Hi, 
I upgraded to the latest version (2.27.4) to see if i could now select an icon. However i'm still seeing the same behaviour as before on firefox. On Edge the issue has been resolved.
Firefox: 
![afbeelding](https://github.com/Budibase/budibase/assets/153834226/f841a38d-4e65-4b61-a7fb-61f91119c927)
Edge:
![afbeelding](https://github.com/Budibase/budibase/assets/153834226/7de86a60-8f74-4945-817b-1ecf62a8bd7e)

Please don't forget us non-chromium based users 😊🙏

aptkingston:
@NicolasBio in your Firefox screenshot it's showing the old version - you can tell because the arrows are blue either side of the "Page 1 of 3". Please make sure you're using the latest version and refresh your page and it'll be fixed.

Don't worry - it'll definitely work on Firefox!

NicolasBio:
> @NicolasBio in your Firefox screenshot it's showing the old version - you can tell because the arrows are blue either side of the "Page 1 of 3". Please make sure you're using the latest version and refresh your page and it'll be fixed.
> 
> Don't worry - it'll definitely work on Firefox!

Ah yes now i see that it works! There was still some cashing probably. Thanks!

PR #13761, opened 2024-05-23
Fix icon select popover and make somewhat presentable

aptkingston:
## Description
Fixes an issue with the icon select popover where it could be tiny if near the edge of the screen. Also gave it a quick lick of paint so that it doesn't look *absolutely terrible*.

Old (awful colours, internal scroll due to 60 massive icons per page):
image

New (slightly improved, no internal scroll with 20 icons per page, more standard colours, fixed spelling and casing):
image

## Addresses
- https://github.com/Budibase/budibase/issues/13711


Ghrehh (approved):
LGTM, looks much better

PR #13763, opened 2024-05-23
Fix for component tree tooltips

deanhannigan:
## Description

- Fix for tooltips in the Component tree in the builder getting stuck in the top left corner. There is an AbsTooltip in the Icon component now and switching to that did the trick. Default config set to type INFO and opening to the right.

## Screenshots
Tooltip issue
Screenshot 2024-05-23 at 16 13 13

## Launchcontrol
Builder component tree icon fixes


aptkingston (approved):
LGTM!

Issue #13768, opened 2024-05-24
Table buttons don't display an icon whenever one is selected


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method:  docker compose
  - Budibase Version: 2.27.3
  - App Version: 2.27.3

**Describe the bug**
Button icons don't display whenever adding them to the table component.

**To Reproduce**
1. Add a table component
2. Scroll down to the bottom of the settings area
3. Add a button
4. Click the settings cog
5. Click any icon to add
6. See no button icon shows up when one is selected.

**Expected behavior**
An icon should be displayed whenever one is selected for the button component.

PR #13769, opened 2024-05-24
Add icons to grid buttons and simplify button column

aptkingston:
## Description
Updates grid buttons to use the icon setting present in the button settings. Also simplifies some of the button column code as we don't actually need a hidden copy to measure.

Even more useful when https://github.com/Budibase/budibase/pull/13761 is merged as that PR makes sure you can actually see the icon popover, and makes it look a lot better.

image

## Addresses
- https://github.com/Budibase/budibase/issues/13768


packages/frontend-core/src/components/grid/layout/ButtonColumn.svelte
```diff
@@ -91,6 +79,9 @@
                   overBackground={button.type === "overBackground"}
                   on:click={() => handleClick(button, row)}
                 >
+                  {#if button.icon}
+                    <i class="{button.icon} S" />
```
Ghrehh, line 83: Is this meant to a template literal?

Ghrehh: Had no idea, 🪄 magic ✨

Ghrehh (approved):
LGTM

PR #13794, opened 2024-05-28
If a user starts the onboarding process, make sure they can still accept an invite

melohagan:
## Description
If a user tried to accept an invite after starting the onboarding process, it would cause an error loop. This has been fixed by updating the NO_TENANCY list to include the check invite endpoint, and also setting the onboarding cookie to complete. 

Also added validation to ensure passwords are at least 8 characters.

Edit: Just noticed that this also fixes the same problem if the user is in an incognito window.

## Addresses
- https://linear.app/budibase/issue/GROW-457/if-a-user-starts-the-onboarding-process-but-doesnt-finish-accepting-an

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/35e93d3d-2bac-4b0b-8b5e-7d164632172f

## Launchcontrol
Niche onboarding bug fix


jvcalderon (approved):
Good one!

PR #13795, opened 2024-05-28
Fix for template apps (old rows/views) in SQS

mike12345567:
## Description
Fix for issue with templated apps - some template apps are *very* old and therefore don't necessarily abide by the rules that are generally followed for new apps, this is something we should resolve, but for now adding an option that allows searching for only a specific document type.

This shouldn't strictly be necessary for new apps as old views don't exist anymore but it is probably a good idea to have this option, allowing restricting the retrieved information to 100% just rows (using the `_id`) when searching internal SQLite tables. Funky things can happen when other documents appear as rows in tables/grids, this avoids the whole area of problems.

The issue stems from the fact that we depend on documents have a `tableId` property to sort them into their tables, in all recent apps (post 2022) rows are the only document that have this property, but in some very old app templates they contain an old style of view document, which also has a `tableId` property. The rows also don't have any other properties other than table ID in these old templates, so there is no other way to split them out other than checking their `_id`.

Issue raised by @samwho in a roundabout way.


samwho (changes requested):
I'm not sure exactly _how_ you'd test this, but I think we do need tests for it.

samwho (approved):
Discussed this with Mike and we decided to merge this PR and do the tests separately

Issue #13797, opened 2024-05-28
The UI freezes when a table is empty and there's a button added to said table


bjornclauw:
**Hosting**
Selfhosted docker  v2.27.4

**Describe the bug**
When I have a table that has 0 rows (either by filtering or empty) and there is a button added to this table, the whole UI freezes (in the builder and also the released app)

**Reproduce**
Create a new app with example data
Add a table, add a button to this table
Add a filter to the table that is guaranteed to not return any results

**note**  `Return all table rows when filter is empty` does not seem to function correctly in a table (returns no rows)

Reftek85:
oh my god! I was thinking I was going crazy!

I updated my version this PM, 
on an empty table, when the table finish loading, it freeze the UI, in the builder and the App, I cannot push create customer button in my main view in the app once it is loaded empty. cannot open a side panel. it is frozen....
version 2.27.4
budibase  hosted on budibase.app cloud instance.

aptkingston:
Thanks @bjornclauw and @Reftek85. I have a PR up to fix this which I'll get deployed ASAP.

raventechsaigon:
I's getting usual UI FREEZES and crashes as well.  With multiple tables and dataprovider on on screen.  And yes I've encounter this behavior for an empty table.

aptkingston:
@raventechsaigon if you have a different, reproducible scenario that causes the UI to freeze then please open another issue with more details so that we can look at it. Thanks.

PR #13800, opened 2024-05-29
Grid improvements and fixes

aptkingston:
## Description
- Offset button column to account for scrollbar if required. This just means we add a bit more padding to the right of the buttons if the scrollbar is in the way.
image
image
- The grid currently ensures that the selected cell is always visible by automatically scrolling to keep it in view. This does not account for buttons, so it can actually be hidden without manual scrolling if you have a large button column. This has been updated to automatically account for the size of the button column.
- Fixes an issue where an error is thrown if you add buttons but have no rows.

## Addresses
- https://github.com/Budibase/budibase/issues/13797


deanhannigan (approved):
LGTM! Tested on a grid with no rows. No further issues with the UI!

Issue #11044, opened 2023-06-27
Ability to revoke invitations


diogenesbrussels:
**Describe the feature request**
I would like to be able to manage invitations to new users:
- Be able to revoke them
- See the remaining validity period
- Extend the validity period?

Discord discussion for ref.;
https://discord.com/channels/733030666647765003/1121369030343786639

aptkingston:
Thanks for raising this @diogenesbrussels 👍

diogenesbrussels:
@aptkingston Is it still planned or considered? Thanks!

diogenesbrussels:
Duplicate #12342  ?

OriginalLimited:
Hello,

When you invite a user and they are sat in "Pending Invites" there are no options to make changes. 
For example:
- Cancel Invite
- Re-send Invite

I found this out when the email settings were not setup correctly - the email did not send and I cannot add the same email addresses again.

Thanks!

PR #13805, opened 2024-05-29
Allow a user invite to be revoked

melohagan:
## Description
User invites can now be deleted. 

## Addresses
- https://github.com/Budibase/budibase/issues/11044
- https://linear.app/budibase/issue/BUDI-7938/allow-a-user-invite-to-be-revoked

## Screenshots
https://github.com/Budibase/budibase/assets/101575380/e775a6b3-4844-44f0-ab96-8bdfdff517d7

## Launchcontrol
Allow user invites to be revoked.


adrinr:
From the UI itself, should the "delete users" and "delete invites" be separate? They are 2 different tables and it does not seem too intuitive. And it will be probably even worst with apps with plenty of users

packages/worker/src/api/routes/global/users.ts
```diff
@@ -108,6 +108,11 @@ router
     buildInviteMultipleValidation(),
     controller.inviteMultiple
   )
+  .post(
```
adrinr, line 111: This should be `DELETE /api/global/users/multi/invite`

packages/frontend-core/src/api/user.js
```diff
@@ -234,6 +234,16 @@ export const buildUserEndpoints = API => ({
     })
   },
 
+  /**
+   * Removes multiple user invites from Redis cache
+   */
+  removeUserInvites: async users => {
```
adrinr, line 240: Shouldn't the param should be invite ids instead of users?

melohagan:
> From the UI itself, should the "delete users" and "delete invites" be separate? They are 2 different tables and it does not seem too intuitive. And it will be probably even worst with apps with plenty of users

I was going off the design in this linear ticket: https://linear.app/budibase/issue/BUDI-7938/allow-a-user-invite-to-be-revoked

I should have linked it in the OP. I think the reason it's one button is because it allows you to delete both users and invites at the same time, and doesn't add additional clutter/buttons to the page.

packages/worker/src/api/routes/global/users.ts
melohagan: That's what I thought initially, but there seems to be a lot of debate whether you should do it or not: https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

There is one place in which we pass a body into a DELETE call, but it's actually a copy+paste mistake, as we only use the ID. 

```ts
const response = await api.del(
    `/api/system/accounts/${account.accountId}/metadata`,
    {
      body: account,
      headers: {
        [Header.API_KEY]: env.BUDIBASE_API_KEY,
      },
    }
  )
```

```ts
export const destroy = async (ctx: any) => {
  const accountId = accounts.metadata.formatAccountMetadataId(
    ctx.params.accountId
  )
  await accounts.metadata.destroyMetadata(accountId)
  ctx.status = 204
}
```

I'll fix the above at least, and happy to use delete on this endpoint, just not sure what the best practice is.

melohagan: From what I can find the body of a delete should not identify the resources being deleted: https://www.mscharhag.com/api-design/rest-deleting-resources

packages/frontend-core/src/api/user.js
melohagan: Refactored to pass the invite codes directly instead of user emails.

packages/worker/src/api/routes/global/users.ts
adrinr: It looks like an edge case on the standard, but plenty of API uses them. Personally, I think that keeping the "delete" verbs makes it safer and more verbose

melohagan: Sure! I'll make change 🙂

adrinr:
> > From the UI itself, should the "delete users" and "delete invites" be separate? They are 2 different tables and it does not seem too intuitive. And it will be probably even worst with apps with plenty of users
> 
> I was going off the design in this linear ticket: https://linear.app/budibase/issue/BUDI-7938/allow-a-user-invite-to-be-revoked
> 
> I should have linked it in the OP. I think the reason it's one button is because it allows you to delete both users and invites at the same time, and doesn't add additional clutter/buttons to the page.

Then I might be challenging the designs. They are technically 2 different tables. You are even using 2 different api actions from it. So I think it might be confusing (it actually confused me a little bit). Thoughts, @aptkingston ?

packages/worker/src/api/routes/global/users.ts
jvcalderon: Well, if we want to follow the REST standard:

DELETE:
- Idempotent: Yes
- Safe: No

POST:
- Idempotent: No
- Safe: No

So here we are creating an idempotent POST method because in the end it is covering a delete under the hood; no matter how many times you make the POST.

Furthermore, as you say, normally DELETE shouldn't contain that kind of information.

Anyway, here we are just discussing about the standard HTTP spec and (as far as I understand you can implement REST following a different RFC); REST principles are not restricted to an specific protocol.

Said that I think currently there isn't an standard solution. Well we can define our own RCF like the teapot protocol 🤣 -> https://datatracker.ietf.org/doc/html/rfc2324 IMO we can go ahead with POST.

jvcalderon (approved):
LGTM

yarn.lock
```diff
@@ -11480,6 +11480,11 @@ formidable@^2.1.2:
     once "^1.4.0"
     qs "^6.11.0"
 
+fp-ts@^2.5.1:
```
adrinr, line 11483: This should not have changed

packages/worker/src/api/routes/global/users.ts
adrinr: So... Nothing to discuss 💩

aptkingston:
Personally I actually think the single action is alright. It's more confusing for us because we understand that they are 2 different tables and very different entities, but for an end user they are basically the same thing. They are all just users and some are active, and some are pending. The data in both tables is identical, and each row does represent a user (just at different activation stages).

The only change I would make to try and make this even more understandable is rename "Pending invites" to "Pending users". That makes the single "Delete users" button more appropriate since you're always working with "users", rather than it actually deleting a combination of "users" and "invites".

So in my opinion, I would say:
- Keep the single delete button like the designs
- Update "Pending invites" to "Pending users"

aptkingston (approved):
LGTM!

packages/account-portal
adrinr, line 1: Is this expected?

adrinr (approved):
All good, just a little question about the account-portal submodule

packages/account-portal
melohagan: Hmm, no.

Issue #13799, opened 2024-05-29
Iframe spam without display


mad4me:
**Checklist**
-  [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.27.3
 
**Describe the bug**
When entering any binding configuration menu, before it loads, the cursor and the menu itself starts to lag and if we check the dev tools, its possible to see the crazy spam of `display: none;` iframes. 

As far as I know, the bug appears only in the bindings menu, and Its the most I use so it can be pretty anyoing, it loads slowly, the cursor bugs out or the menu may close by itself in the middle of the opening.

**To Reproduce**
1. Open the dev tools
2. Open any binding configuration menu
3. Iframe spam ._.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/54853465/6942fb5a-925a-437a-b9dd-6047078e12a3)

**Desktop**
 - OS: Windows 11
 - Browser: Google Chrome
 - Version: 125.0.6422.78

aptkingston:
Thanks for spotting this. I'd actually experienced this lag myself and had never noticed that this weird iframe spam was the reason. I'll get this fixed!

PR #13807, opened 2024-05-29
Drawer eval improvements

aptkingston:
## Description
This PR changes how we process the real evaluations of all data bindings in the binding drawer. This was previously very slow and created numerous zombie iframes on the page.

The improvements in this PR are:
- Use HBS rather than JS to evaluate bindings, as it's many times faster and does not create ephemeral iframes
- Process all bindings in a single `processObject` call as opposed to hundreds of separate `processString` calls (this still happens under the hood, but is fast this way)
- Fix role evaluations (they use a horrible legacy syntax with built in HBS of `trim {roleId}`, but we now account for this)

## Addresses
- https://github.com/Budibase/budibase/issues/13799


Ghrehh (approved):
LGTM, only thing I noticed testing was that previously empty strings wouldn't show a popover at all.

`User.FirstName` is empty for my local user so there was no popover. Now there is, which IMO is now much better, but it wasn't mentioned in the PR so just highlighting it in case it's some weird requirement we accidentally broke.

aptkingston:
> LGTM, only thing I noticed testing was that previously empty strings wouldn't show a popover at all.
> 
> `User.FirstName` is empty for my local user so there was no popover. Now there is, which IMO is now much better, but it wasn't mentioned in the PR so just highlighting it in case it's some weird requirement we accidentally broke.

Nice catch. To be honest I didn't even realise that behaviour had changed. Happy to leave it as this though if you think it's better - I know what you mean that it probably is better to be explicit about empty values rather than showing no tooltip and making users wonder why.

deanhannigan (approved):
LGTM!

PR #13808, opened 2024-05-29
Grid conflict improvements

aptkingston:
## Description
This PR updates the grid to be a bit smarter when dealing with very slow servers or datasources.

- After a request is completed, only reset the local change cache for columns which were saved and have not been further altered. This should prevent any unwanted resetting of local changes made while the request is in flight
- Support grid awareness for multiple in-flight requests for the same row ID
- Ensure single character changes are properly saved when changing row immediately after. There was an issue where these changes would not be saved if you immediately changed row after making it

## Addresses
- https://linear.app/budibase/issue/BUDI-8276/grid-editing-cell-race-conditions


Ghrehh (approved):
Tested the issue highlighted in the ticket with google sheets locally; this seems to have fixed it perfectly, lgtm

packages/frontend-core/src/components/grid/stores/rows.js
```diff
@@ -404,7 +404,7 @@ export const createActions = context => {
 
     // Save change
     try {
-      // Incremenet change count for this row
+      // Increment change count for this row
```
deanhannigan, line 407: 🥚

deanhannigan (approved):
LGTM!

Issue #13806, opened 2024-05-29
Form Block label field does not wrap - long questions spill over beyond the confines of the form


mikesealey:
**Checklist**

* \[`Y`\] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Cloud and Self-host

**Describe the bug**
When creating a Form Block, if the user adds a long enough string to the Label text it will spill outside the confines of the form, instead of line-breaking and continuing underneath.

**To Reproduce**

1. Create Form Block
2. Set Label text to be \~200 characters or more

**Expected behavior**
Text should wrap - Help Text wraps, Label Text does not.

**Screenshots**

![image](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/1f8e10c5-451e-4e25-b58c-1fb7e9201096/6342543a-a8f5-414f-a1ff-55f737ad1303?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8xZjhlMTBjNS00NTFlLTRlMjUtYjU4Yy0xZmI3ZTkyMDEwOTYvNjM0MjU0M2EtYThmNS00MTRmLWExZmYtNTVmNzM3YWQxMzAzIiwiaWF0IjoxNzE2OTg5NjgzLCJleHAiOjMzMjg3NTQ5NjgzfQ.8plm_fL1gp_lCKvVVgEul0Bcvu3ZDxQ1qiIL5QF52GU)

PR #13809, opened 2024-05-29
Allow field labels to wrap

aptkingston:
## Description
This PR updates field labels to allow them to wrap onto new lines.

image

## Addresses
- #13806


Ghrehh (approved):
LGTM

NAB: Still overflows on an incredibly long word with no whitespace, seems like a case of stupid games = stupid prizes, but maybe worth adding `word-break: break-word;` also.

aptkingston:
> LGTM
> 
> NAB: Still overflows on an incredibly long word with no whitespace, seems like a case of stupid games = stupid prizes, but maybe worth adding `word-break: break-word;` also.

Haha yea I realised that too, and deliberately chose to not add it in protest of someone who would actually do that. But there's no downside of adding it so I'll throw it in now that you've caught me 😂

deanhannigan (approved):
LGTM!

PR #13818, opened 2024-05-30
Add readonly view config

adrinr:
## Description
Allow readonly columns configuration on views. This is a paid feature, it will be only enabled for users with the right license but still visible to everybody else.
No actual frontend implementation is developed at this point, just the capability of setting this config value

## Addresses
- [BUDI-8281 - [Readonly views] Update view column UI](https://linear.app/budibase/issue/BUDI-8281/[readonly-views]-update-view-column-ui)

### Free user view
image
image

### Feature available
image
image
image

### Downgraded user using the feature (for security reasons it will still behave as readonly, but it will not be able to set any other value as readonly)
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
+++ b/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
@@ -28,6 +28,7 @@
     showAvatars={false}
     on:updatedatasource={handleGridViewUpdate}
     isCloud={$admin.cloud}
+    allowReadonlyColumns
   >
     <svelte:fragment slot="filter">
       <GridFilterButton />
--- a/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
+++ b/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
@@ -1,6 +1,6 @@
 <script>
   import { viewsV2 } from "stores/builder"
-  import { admin } from "stores/portal"
+  import { admin, licensing } from "stores/portal"
   import { Grid } from "@budibase/frontend-core"
   import { API } from "api"
   import GridCreateEditRowModal from "components/backend/DataTable/modals/grid/GridCreateEditRowModal.svelte"
@@ -28,7 +28,8 @@
     showAvatars={false}
     on:updatedatasource={handleGridViewUpdate}
     isCloud={$admin.cloud}
-    allowReadonlyColumns
+    showReadonlyColumnsOptions
+    canSetReadonlyColumns={$licensing.isViewReadonlyColumnsEnabled}
   >
     <svelte:fragment slot="filter">
       <GridFilterButton />
--- a/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
+++ b/packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
@@ -28,8 +28,7 @@
     showAvatars={false}
     on:updatedatasource={handleGridViewUpdate}
     isCloud={$admin.cloud}
-    showReadonlyColumnsOptions
-    canSetReadonlyColumns={$licensing.isViewReadonlyColumnsEnabled}
+    allowViewReadonlyColumns={$licensing.isViewReadonlyColumnsEnabled}
   >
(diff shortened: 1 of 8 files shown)


packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
```diff
@@ -14,36 +17,64 @@
 
   const toggleColumn = async (column, permission) => {
```
samwho, line 18: Do we need to modify `anyHidden` above?

samwho (approved):

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
adrinr: I did not realise about this logic... I think this should go. It should be active ONLY if open, otherwise it looks like it's focused or similar. Thoughts, @aptkingston ?

https://github.com/Budibase/budibase/assets/15987277/9ccbdc97-3220-4997-806a-9710a2a6792e

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
```diff
@@ -1,9 +1,12 @@
 <script>
   import { getContext } from "svelte"
-  import { ActionButton, Popover, Icon } from "@budibase/bbui"
+  import { ActionButton, Popover, Icon, notifications } from "@budibase/bbui"
   import { getColumnIcon } from "../lib/utils"
   import ToggleActionButtonGroup from "./ToggleActionButtonGroup.svelte"
 
+  export let showReadonlyColumnsOptions = false
```
adrinr, line 7: I don't like this double booleans, but I think it's cleaner than using enums in JS or complex objects as parameters..

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
aptkingston: It was designed to look this way, but I’m not strongly opinionated about it. The same behaviour happens with filters too - it’s in an active state of any filters are applied. Mainly just designed to make it very obvious that you have settings configured that are hiding rows or columns, since it could be confusing if you don’t realise that.

packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
```diff
@@ -28,6 +28,8 @@
     showAvatars={false}
     on:updatedatasource={handleGridViewUpdate}
     isCloud={$admin.cloud}
+    showReadonlyColumnsOptions
+    canSetReadonlyColumns={$licensing.isViewReadonlyColumnsEnabled}
```
aptkingston, line 32: What's the purpose of the `showReadonlyColumnOptions` prop - just to control whether we show that readonly columns are a things only on views and not on other datasources? I would rather derive this from the datasource type, in the grid datasource store.

We do something similar for `hasBudibaseIdentifiers` in there.

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
aptkingston: Yea I agree it's probably better to just use an explicit boolean for whether or not we can use views, rather than passing around a license prop or something and duplicating the logic in here.

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
aptkingston: I think my preference is still show this as highlighted to make it clear that there has been some deviation from the table schema.

aptkingston (approved):
Nice work @adrinr! Just one suggested change to reduce the new external grid props from 2 to 1.

packages/builder/src/components/backend/DataTable/ViewV2DataTable.svelte
adrinr: Done, I was not sure if it was better to have 2 props or having a more coupled grid

PR #13819, opened 2024-05-30
Allow users to only specify a binding when adding attachments

PClmnt:
## Description
Previously with the addition of attachments to automations, both in email and create / row steps the user had to specify both the URL and the Filename of the attachment they wanted to upload or send. However, this left a gap, in that what if the user has an indeterminate amount of attachments they need to send? As it stands now they would need to do:
```
steps.1.response.1.url | filename1.gz
steps.1.response.2.url | filename2.gz
steps.1.response.3.url | filename3.gz
```
and this still doesn't account for if they don't know how many attachments will be returned exactly.

This PR simply allows them to toggle to a normal binding input box where they can specify a binding which satisfies the` {url: string, filename: string}` requirement. This will allow users to say for example use the JS scripting block to build up a more specific list of attachments, dynamically, before passing them to the attachments. 

Also added error handling to fix issues with blank attachments being sent back causing the automation to fail, and also to handle the new scenario where someone might build up an object without the required URL / Filename keys. 
## Addresses
(https://linear.app/budibase/issue/BUDI-8278/update-attachments-in-automations-to-support-bindings-of-url)

## Screenshots

image
image
image

## Feature branch env
[Feature Branch Link](http://fb-fix-allow-users-to-specify-binding-attachment.fb.qa.budibase.net)


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -462,27 +462,75 @@
...
+                  text={"Use bindings"}
+                  size={"XS"}
+                  on:change={e => {
                     onChange(
                       {
-                        detail: e.detail.map(({ name, value }) => ({
-                          url: name,
-                          filename: value,
-                        })),
+                        detail: "",
```
PClmnt, line 473: When a user is toggling from binding to attachment fields, we just clear it out as there's no real way to parse say, a javascript binding to the correct format.

packages/server/src/automations/automationUtils.ts
```diff
@@ -99,6 +99,12 @@ export function getError(err: any) {
   return typeof err !== "string" ? err.toString() : err
 }
 
+export function validateAttachment(attachmentObject: any) {
+  if (!("url" in attachmentObject) || !("filename" in attachmentObject)) {
+    throw new Error("Invalid key. Object must have both URL and Filename keys")
```
samwho, line 104:
```suggestion
    throw new Error(`Attachments must have both "url" and "filename" keys`)
```

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
adrinr: Instead of having inline functions, we could have a `clearAttachment` or similar function and to call it here. It will make it much more verbose

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -462,27 +462,75 @@
...
+                    title={value.title || label}
+                    value={inputData[key]}
+                    panel={AutomationBindingPanel}
+                    type={value.customType}
+                    on:change={e => onChange(e, key)}
+                    {bindings}
+                    updateOnChange={false}
+                  />
+                {:else}
+                  <div class="test">
```
adrinr, line 517: Is this "test" class expected?

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -94,17 +94,22 @@
     dispatch("change", newValue)
   }
 
-  const onChangeSetting = (e, field) => {
-    let fields = {}
-    fields[field] = {
-      clearRelationships: e.detail,
+  const onChangeSetting = (e, field, setting) => {
```
adrinr, line 97: NIT: the parameters of this function look quite strange. I would expect: field, key, value. Now is the other way round, and it gets an event instead of a value.
I needed to check the usage of this function to understand what it was the usage

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -14,6 +20,8 @@
   export let value
   export let bindings
   export let isTestModal
+  export let meta
```
adrinr, line 23: It looks like you are only using `meta` for `meta?.fields?.[field]?.`. This long null checks get things a little bit hard to follow. We could either pass the actual field metadata as input (instead of the full meta) or extract the value as `$: useAttachmentBinding = meta?.fields?.[field]?.useAttachmentBinding` and use that instead

packages/server/src/automations/automationUtils.ts
```diff
@@ -99,6 +99,12 @@ export function getError(err: any) {
   return typeof err !== "string" ? err.toString() : err
 }
 
+export function validateAttachment(attachmentObject: any) {
```
adrinr, line 102: Naming code conventions generally expect validate to return a response. This seems more like `guardAttachment`

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
PClmnt: Yep in this case that's the class for the "test" drawer. It's also used elsewhere in that file

samwho (approved):

Issue #13801, opened 2024-05-29
Add an additional setting that allows notifications to be dismissed after x amount of time


ConorWebb96:
**Describe the feature request**
As a user sometimes I won't to display information to a user for a long enough time were they can read and absorb the information. Currently there isn't a way to do this, either notifications stay forever or they get auto dismissed after a 3 second delay.

PR #13829, opened 2024-05-31
Enhancement: add ability to set custom auto-dismissal duration for notifications

ConorWebb96:
## Description
Implemented support for custom auto-dismissal durations in notifications by passing the duration parameter through the showNotificationHandler. This allows notifications to respect user-specified durations instead of defaulting to 3000ms.

## Addresses
- https://github.com/Budibase/budibase/issues/13801

## Screenshots
**Previous ui**
Screenshot 2024-05-31 at 11 56 26

**Ui with this change**
Screenshot 2024-05-31 at 11 58 14

## Launchcontrol
Added the ability to set custom durations to the notification action.


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/ShowNotification.svelte
```diff
@@ -47,6 +47,15 @@
   />
   <Label />
   <Checkbox text="Auto dismiss" bind:value={parameters.autoDismiss} />
+  {#if parameters.autoDismiss}
+    <Label>Duration (ms)</Label>
+    <DrawerBindableInput
+      title="Duration"
+      {bindings}
+      value={parameters.duration}
+      on:change={e => (parameters.duration = e.detail)}
```
aptkingston, line 56: I think it would be useful to add a placeholder of 3000 here to let users know what the default is, if they don't customise this.

packages/client/src/stores/notification.js
```diff
@@ -1,8 +1,6 @@
 import { writable, get } from "svelte/store"
 import { routeStore } from "./routes"
 
-const NOTIFICATION_TIMEOUT = 3000
```
aptkingston, line 4: I would keep this here and maybe just rename it to DEFAULT_NOTIFICATION_TIMEOUT. It's nice to have constants like this defined in an easy to find location, rather than having hardcoded values buried in files.

packages/client/src/stores/notification.js
```diff
@@ -52,6 +52,7 @@ const createNotificationStore = () => {
           message,
           icon,
           dismissable: !autoDismiss,
+          duration,
```
aptkingston, line 55: `duration` isn't used, so we don't need to store it here. We only need it down below in the `setTimeout` block.

packages/client/src/stores/notification.js
```diff
@@ -18,25 +16,27 @@ const createNotificationStore = () => {
     type = "info",
     icon,
     autoDismiss = true,
+    duration = 3000,
```
aptkingston, line 19: Defaulting this here accounts for `undefined` only, but not for `null` or the more common `""` (empty string) which will happen if users touch the new setting but don't enter anything. So I'd prefer to remove the default here and instead apply it below.

packages/client/src/stores/notification.js
```diff
@@ -60,7 +61,7 @@ const createNotificationStore = () => {
     if (autoDismiss) {
       setTimeout(() => {
         dismiss(_id)
-      }, NOTIFICATION_TIMEOUT)
+      }, duration)
```
aptkingston, line 64: Since this is the only place we actually use `duration`, I think the best option is to ensure that we have a valid value right here. I would update this to `duration || DEFAULT_NOTIFICATION_TIMEOUT` as that account for `undefined`, `null`, empty string and even 0. It's a more robust approach and just defaulting it in the function signature 👍

packages/client/src/stores/notification.js
```diff
@@ -18,25 +16,27 @@ const createNotificationStore = () => {
...
 
-    // If peeking, pass notifications back to parent window
     if (get(routeStore).queryParams?.peek) {
       window.parent.postMessage({
         type: "notification",
         detail: {
           message,
           type,
           icon,
+          duration,
```
aptkingston, line 33: We don't need to worry about checking for a valid value here, as the event is ultimately passed back to the parent window which runs through this same code, and will eventually only be used down in the `setTimeout` block below, where we do default it. Just explaining why I don't think we need to worry about making any changes here - this can stay as you have it.

aptkingston (commented):
Left a few comments about how to make this a little more robust against bad user inputs, but looks good overall!

melohagan (approved):
LGTM! Nice one 👍 

I would put a limit on the duration though, can't imagine you would want a notification to remain permanently on the screen.

PR #13832, opened 2024-05-31
Fix mariadb healthcheck.

samwho:
## Description

The image `mariadb:lts` was updated ~10 hours ago and, with it, the `mysqladmin` binary was removed. We were relying on this to healthcheck the image.

Fortunately, we weren't the only ones and the fix is here: https://github.com/MariaDB/mariadb-docker/issues/512#issuecomment-1608976749.


adrinr (approved):
🚀

PR #13833, opened 2024-05-31
Validate readonly configuration

adrinr:
## Description
Validate view and table schema configurations

### Attempting to make a required column read only
Table with the following columns: name (display column), surname (required) and dob (optional)

https://github.com/Budibase/budibase/assets/15987277/7f12b165-2b48-4423-ab64-20a1d09a27bb

### Updating column to required and readonly in some views

https://github.com/Budibase/budibase/assets/15987277/1dda0bf0-3fd4-4ead-92d8-04cf8bd23834

## Addresses
- [BUDI-8282 - [Readonly views] Validation for view column configuration](https://linear.app/budibase/issue/BUDI-8282/[readonly-views]-validation-for-view-column-configuration)

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -349,8 +349,7 @@ describe.each([
         await config.api.viewV2.create(newView, {
           status: 400,
           body: {
-            message:
-              'Field "name" cannot be readonly as it is a required field',
+            message: 'You can\'t make the required field "name" read only',
             status: 400,
           },
         })
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -1424,7 +1424,8 @@ describe.each([
             status: 400,
             body: {
               status: 400,
-              message: 'Required field "name" is missing in view "view a"',
+              message:
+                'To make field "name" required, this field must be present and writable in views: view a.',
             },
           }
         )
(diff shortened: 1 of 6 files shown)


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -349,8 +345,7 @@ describe.each([
         await config.api.viewV2.create(newView, {
           status: 400,
           body: {
-            message:
-              'Field "name" cannot be readonly as it is a required field',
+            message: 'You can\'t make read only the required field "name"',
```
samwho, line 348: This reads a little strange, I think: "You can't make field "name" readonly because it is a required field." would be better.

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -1348,4 +1343,123 @@ describe.each([
...
+                constraints: { presence: true },
+              },
+            },
+          },
+          {
+            status: 400,
+            body: {
+              status: 400,
+              message:
+                'Invalid body - Required field "name" is missing in view "view a"',
```
samwho, line 1427: I think it's going to be unclear to users what this means. What are they supposed to do to make this action possible? Add the field "name" to the view "view a"? If so, maybe: "To field "name" required, the field must be present and writable in views: <list of views>".

packages/server/src/sdk/app/views/index.ts
```diff
@@ -39,9 +39,7 @@ async function guardViewSchema(
   tableId: string,
   viewSchema?: Record<string, ViewUIFieldMetadata>
 ) {
-  if (!viewSchema || !Object.keys(viewSchema).length) {
-    return
-  }
+  viewSchema ??= {}
```
samwho, line 42: TIL.

samwho (approved):

PR #13835, opened 2024-05-31
Fixing some REST file download issues

mike12345567:
## Description
Some issues raised by @andz-bb - the following:

1. Some APIs (Carbone) will return errors, but also still return content dispositions suggesting the body should be downloaded. Making sure that we only attempt to download the response if it is a successful response code
2. Streamed responses aren't always complete when we try to send them to S3/MinIO, so need to wait for the stream to finish before passing it along to avoid uploading corrupt files.
3. Some APIs don't tell us how big the file will be that we are downloading, for our metadata we need to ask MinIO/S3 how big the file is once we have completed the streamed download.

Also an issue I found while investigating, dynamic variables weren't working exactly as expected due to bindings not taking their default parameters properly - it depended on the main query having the same bindings and default value as the one being called to get dynamic variables.

### DIFF
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -41,10 +41,7 @@ type UploadParams = BaseUploadParams & {
   path?: string | PathLike
 }
 
-export type StreamTypes =
-  | ReadStream
-  | NodeJS.ReadableStream
-  | ReadableStream<Uint8Array>
+export type StreamTypes = ReadStream | NodeJS.ReadableStream
 
 export type StreamUploadParams = BaseUploadParams & {
   stream?: StreamTypes
@@ -222,6 +219,9 @@ export async function streamUpload({
   extra,
   ttl,
 }: StreamUploadParams) {
+  if (!stream) {
+    throw new Error("Stream to upload is invalid/undefined")
+  }
   const extension = filename.split(".").pop()
   const objectStore = ObjectStore(bucketName)
   const bucketCreated = await createBucketIfNotExists(objectStore, bucketName)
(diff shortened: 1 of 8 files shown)


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -251,14 +251,35 @@ export async function streamUpload({
...
   }
-  return objectStore.upload(params).promise()
+
+  // make sure we have the stream before we try to push it to object store
+  if (stream.on) {
+    await new Promise((resolve, reject) => {
+      stream.on("finish", resolve)
+      stream.on("error", reject)
+    })
+  }
```
samwho, line 270: Does this not defeat the purpose of streaming? The idea is to never hold more than a small amount of the object in memory. If we're waiting for the stream to end, we're going to be buffering the whole object before uploading it.

yarn.lock
```diff
@@ -11904,6 +11904,17 @@ glob@^10.0.0, glob@^10.2.2:
     minipass "^7.0.4"
     path-scurry "^1.10.2"
 
+glob@^10.3.7:
```
samwho, line 11907: How come this file has been updated?

samwho (changes requested):
If possible, would be nice to get some tests on these fixes to make sure they don't regress.

packages/backend-core/src/objectStore/objectStore.ts
mike12345567: So I thought this as well - however it seems like when we pass the stream onto S3 you only get a partial of the file, as it simply streams up to where-ever the stream was at - I tried a few ways to get around this, but we need the stream to be complete before we can pass it to S3 to avoid partial uploads of files - I couldn't find a lot of information about this though, it seems to be a very niche problem and I don't know how we've never ran into this before.

Considered making the wait for finish a flagged option so that we only use it when working with web streams (pulled through node-fetch).

yarn.lock
mike12345567: this happens occasionally - generally its due to a yarn.lock being committed to master that doesn't include the submodule dependencies (due to the way NX/Lerna works these are all piled into one file)

packages/backend-core/src/objectStore/objectStore.ts
samwho: I don't think we can allow this into production as-is. We have a lot of memory problems already, and they appear to be part of what causes container restarts.

I'd instead opt to stream these files first to disk, and then from disk to S3. This would keep memory use low at the cost of ephemeral disk. The pods in cloud today don't have any ephemeral disk in their resources config, but I can change that for you if we go down this road.

mike12345567: Updated to use `tmp` to write the file through disk rather than straight as a stream.

samwho (approved):

adrinr (approved):

PR #13849, opened 2024-06-04
Add PR size and team labeler

adrinr:
## Description
Automatically add the following tags:
1. Size, using the action https://github.com/CodelyTV/pr-size-labeler
2. `firestorm` for PRs created by firestorm devs, using https://github.com/rodrigoarias/auto-label-per-user


mike12345567 (approved):
Very nice - will be nice to actually label all firestorm PRs - I do use this regularly to check what all is complete/filter down the list etc!

PR #13854, opened 2024-06-04
SQS migrations/app migrations

mike12345567:
## Description
This PR has two parts to it:

1. The app migration system which @adrinr implemented back in December 2023 - it finishes this off so that its ready to use and adds the first migration
2. The migration which is needed for SQS to perform correctly - this comes in three parts:

- Updating all link documents so that they can be correctly indexed by SQS (have the links be in a deterministic order (which one is doc1/2) and add a `tableId` for SQS to recognise it as a row.
- Making sure that the SQS design document exists.
- iterating through each table and running a simple `select * from table limit 1` query to kick off the migration.


packages/server/src/api/controllers/migrations.ts
```diff
@@ -27,7 +27,7 @@ export async function getMigrationStatus(ctx: Ctx) {
 
   const latestAppliedMigration = await getAppMigrationVersion(appId)
 
-  const migrated = latestAppliedMigration === getLatestMigrationId()
+  const migrated = latestAppliedMigration === getLatestEnabledMigrationId()
```
samwho, line 30: Does this need to instead be a `>` check? Thinking about the situation where a migration is dynamically disabled after having been applied (never intended to happen but we _are_ creating the environment for it to happen if mistakes are made). If that does happen, we still want to consider the app migrated if the latest migration is _greater than_ the latest enabled migration.

packages/server/src/appMigrations/index.ts
```diff
@@ -10,22 +10,36 @@ export * from "./appMigrationMetadata"
 export type AppMigration = {
   id: string
   func: () => Promise<void>
+  // disabled so that by default all migrations listed are enabled
+  disabled?: boolean
 }
 
-export const getLatestMigrationId = () =>
-  MIGRATIONS.map(m => m.id)
+export function getLatestEnabledMigrationId() {
```
samwho, line 17: It's not going to a problem at the moment but long-term we'll _probably_ want to avoid creating a copy of the migration list for every request.

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
```diff
@@ -0,0 +1,59 @@
+import { context } from "@budibase/backend-core"
+import { allLinkDocs } from "../../db/utils"
+import LinkDocumentImpl from "../../db/linkedRows/LinkDocument"
+import sdk from "../../sdk"
+import env from "../../environment"
+import { DBError } from "@budibase/types"
+
+const migration = async () => {
+  const linkDocs = await allLinkDocs()
```
samwho, line 9: Does this have the potential to be very large? Do we instead want to paginate?

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
```diff
@@ -0,0 +1,59 @@
...
+    // do these one by one - running in parallel could cause problems
+    for (let table of tables) {
+      try {
+        await db.sql(`select * from ${table._id} limit 1`)
+      } catch (err) {
+        if (typeof err === "object") {
+          const dbErr = err as DBError
+          throw new Error(`Failed to run initial SQS search - ${dbErr.message}`)
+        }
+        throw err
```
samwho, line 53: Do we really need this catch at all?

packages/server/src/appMigrations/tests/migrations.spec.ts
```diff
@@ -52,4 +53,26 @@ describe("migrations", () => {
       },
     })
   })
+
+  it("should disable migrations if any migration is disabled", () => {
```
samwho, line 57: Name of this test isn't quite accurate anymore.

packages/server/src/appMigrations/tests/migrations.spec.ts
```diff
@@ -52,4 +53,26 @@ describe("migrations", () => {
...
+    // remove all migrations
+    migrations.MIGRATIONS.splice(0, migrations.MIGRATIONS.length)
```
samwho, line 59: Modifying global state in a test is risky. I'd recommend one of two things:

1. Put this back how you found it when you're done.
2. Make `getLatestEnabledMigrationId()` accept a parameter that is the list of migrations to operate over, removing the need to modify global state.

I like 2 more than 1.

packages/server/src/appMigrations/tests/migrations.spec.ts
```diff
@@ -52,4 +53,26 @@ describe("migrations", () => {
...
+    migrations.MIGRATIONS.push({
+      id: "20231211105810_new-test",
```
samwho, line 61: To aid test readability, could you extract these strings out into variables?

```typescript
const MIGRATION_ID_1 = "...";
const MIGRATION_ID_2 = "...";
```

packages/server/src/middleware/appMigrations.ts
```diff
@@ -1,9 +1,16 @@
 import { UserCtx } from "@budibase/types"
 import { checkMissingMigrations } from "../appMigrations"
+import { env } from "@budibase/backend-core"
 
 export default async (ctx: UserCtx, next: any) => {
   const { appId } = ctx
 
+  // migrations can be disabled via environment variable, or can be disabled
+  // due to some of the migrations not being ready to run - disables all migrations
```
samwho, line 9: I think this comment is no longer quite correct.

packages/server/src/sdk/app/tables/internal/sqs.ts
```diff
@@ -107,8 +108,18 @@ async function buildBaseDefinition(): Promise<PreSaveSQLiteDefinition> {
 
 export async function syncDefinition(): Promise<void> {
   const db = context.getAppDB()
+  let existing: SQLiteDefinition | undefined
+  if (await db.exists(SQLITE_DESIGN_DOC_ID)) {
+    existing = await db.get<SQLiteDefinition>(SQLITE_DESIGN_DOC_ID)
+  }
```
samwho, line 114: Could remove a DB call by wrapping just a `db.get` in a try-catch. Not sure how often this function is going to run in practice to know whether that's worth doing.

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
mike12345567: I think the scale of this is likely to be small enough (its very hard to create a lot of internal relationships today) that pagination adds enough complexity that its not worth it.

packages/server/src/appMigrations/index.ts
mike12345567: In this case it isn't copying the whole migration anymore - it creates an array of references to the enabled migrations (not cloned) and then maps out the IDs of them, theres never a full copy of the migrations.

samwho (approved):

packages/server/src/appMigrations/index.ts
```diff
@@ -10,22 +10,36 @@ export * from "./appMigrationMetadata"
...
-export const getLatestMigrationId = () =>
-  MIGRATIONS.map(m => m.id)
+export function getLatestEnabledMigrationId(migrations?: AppMigration[]) {
+  const enabledMigrations: AppMigration[] = []
+  for (let migration of migrations || MIGRATIONS) {
+    // if a migration is disabled, all migrations after it are disabled
+    if (migration.disabled) {
+      break
+    }
+    enabledMigrations.push(migration)
```
adrinr, line 24: Do we need to keep pushing it? Could we not just order on the foreach and keep getting the latest it?

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (docsToUpdate.length) {
+    await db.bulkDocs(docsToUpdate)
+  }
+
+  // at the end make sure design doc is ready
+  await sdk.tables.sqs.syncDefinition()
+  // only do initial search if environment is using SQS already
+  // initial search makes sure that all the indexes have been created
+  // and are ready to use, avoiding any initial waits for large tables
+  if (env.SQS_SEARCH_ENABLE) {
```
adrinr, line 42: I don't feel too comfortable with it... Should the migration run only if the SQS is enabled? What would happen if it runs when it's not enabled, and then it gets enabled?

mike12345567: that's fine - disabling/enabling this stuff is ok, i've tested disabling/enabling over and over again and it doesn't cause any problems (as the migration is backwards compatible and the integrity test confirms you can re-run).

adrinr (approved):

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
mike12345567: The worst case scenario is that the indexes aren't generated by the migration (SQS isn't available, so it just doesn't call it) but these indexes will be generated on the fly as well - essentially the user will have a slow first query experience, but after that problem is solved.

PR #13860, opened 2024-06-05
Prepare Helm chart for SQS.

samwho:
## Description

This PR adds the necessary gubbins to our Helm chart to make it ready for the transition to SQS. You will be able to enable SQS in your `values.yaml` file by adding:

```yaml
globals:
  sqs:
    enabled: true
```

If you're using the chart-supplied CouchDB, there will be nothing more you need to do. Currently, we don't default to an SQS-enabled CouchDB image, though, so if you do want to switch early (please don't, if you're reading this and don't work at Budibase!) you'll need to switch to using `budibase/couchdb:v3.3.3-sqs`. We're planning to do this in our QA environment imminently.


mike12345567 (approved):
LGTM! Nice work - great to get this together!

PR #13863, opened 2024-06-06
Adding current user global ID binding

mike12345567:
## Description
Adding a global binding for `globalId` which is retrieved as part of the self call - making sure this is available as a binding (this is not in the user metadata table, just for bindings).

Useful feature suggested by Ronan.


samwho (approved):
Test? 😇

Issue #13839, opened 2024-06-02
Cannot save time value to internal DB


mohagan9:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: rspb.budibase.app

**Describe the bug**
I cannot save time values to the internal DB.

**To Reproduce**
1. Add date/time columns to an internal Budibase DB
2. Create a form with time only fields for these date/time columns
3. Try to save data
4. See in network tab validation constraint being hit

**Expected behavior**
I should be able to save time only date/time values to the internal Budibase DB.

**Screenshots**
<img width="505" alt="Screenshot 2024-06-02 at 23 08 56" src="https://github.com/Budibase/budibase/assets/30199063/f92fc780-b4cc-45bc-b59d-a168e3200d52">

<img width="618" alt="Screenshot 2024-06-02 at 23 09 24" src="https://github.com/Budibase/budibase/assets/30199063/34f7747a-bbb3-478c-9574-40deec9d7506">

<img width="1027" alt="Screenshot 2024-06-02 at 23 11 41" src="https://github.com/Budibase/budibase/assets/30199063/1020b710-2091-47b9-8543-fb999a8b5864">

<img width="348" alt="Screenshot 2024-06-02 at 23 12 45" src="https://github.com/Budibase/budibase/assets/30199063/5032def9-4574-4b75-b045-e329ab8f703e">

<img width="321" alt="Screenshot 2024-06-02 at 23 14 31" src="https://github.com/Budibase/budibase/assets/30199063/22d71b61-8e26-49a6-90d0-627c10893616">

**Additional context**
This definitely worked fine a month ago.

ConorWebb96:
Re-create app export here.

[...]

PR #13865, opened 2024-06-06
Fix time-only form fields for internal db

adrinr:
## Description
We recently fixed some issues related to time-only fields, storing these as time strings instead of date times. This is now conflicting with the datetime values with time-only frontend, as the API is still expecting a iso datetime.
For this, the field will know if the actual value to store is a datetime or a time-only, parsing it accordingly.

As this is a frontend implementation concern, the API will still expect the right formats for each field depending of its configuration.

Also, date-only fields will hide the time by default (it can still be added back manually)

https://github.com/Budibase/budibase/assets/15987277/f0414eea-c1f5-4222-a078-c0b2499c9e5a

## Addresses
- https://github.com/Budibase/budibase/issues/13839

## Launchcontrol
Fixing time-only form fields for internal db

### DIFF
--- a/packages/client/src/components/app/blocks/FormBlockComponent.svelte
+++ b/packages/client/src/components/app/blocks/FormBlockComponent.svelte
@@ -69,9 +69,13 @@
         }
       },
       [FieldType.DATETIME]: (_field, schema) => {
-        return {
+        const props = {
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
LGTM!

PR #13866, opened 2024-06-06
Update Lerna to 8.x

adrinr:
## Description
Recently we are having some flakiness on the pipelines because lerna started picking up the account-portal root as a package, causing some scripts to be called multiple times:
image

The issue was that some scripts were forcing this to be treated as a package, even if it was meant to be ignored.
On the process to fix that, we updated lerna to 8.x, and running lerna repair (that caused some extra .json files to change)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "kill-port": "^1.6.1",
     "lerna": "8.1.3",
     "madge": "^6.0.0",
-    "nx": "16.4.3",
     "nx-cloud": "16.0.5",
     "prettier": "2.8.8",
     "prettier-plugin-svelte": "^2.3.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -37,8 +35,8 @@
     "build": "NODE_OPTIONS=--max-old-space-size=1500 lerna run build --stream",
     "build:apps": "yarn build --scope @budibase/server --scope @budibase/worker",
     "build:cli": "yarn build --scope @budibase/cli",
-    "build:oss": "NODE_OPTIONS=--max-old-space-size=1500 lerna run build --stream --ignore @budibase/account-portal --ignore @budibase/account-portal-server --ignore @budibase/account-portal-ui",
-    "build:account-portal": "NODE_OPTIONS=--max-old-space-size=1500 lerna run build --stream --scope @budibase/account-portal --scope @budibase/account-portal-server --scope @budibase/account-portal-ui",
```
adrinr, line 41: `--scope @budibase/account-portal` was actually causing the conflicts

adrinr:
Closing because it needs some investigation. It will be reopened in July

PR #13869, opened 2024-06-06
Enable SQS in dev 🎉

mike12345567:
## Description
As title - this updates environment in dev to make sure that everyone is using SQS rather than Lucene.


samwho (approved):

adrinr (approved):
🚀

PR #13870, opened 2024-06-06
Some automation fixes and refactoring

PClmnt:
## Description
Fixes a couple of issues as well as ~~refactors a lot of the looping code out of automations in to its own file. It was becoming  very difficult and hard to understand what was happening the looping section so I refactored it out into multiple functions. None of the functionality had fundamentally changed but the `execute()` function within the `Orchestrator` is a lot easier to understand, especially considering that most of the time a loop isn't even invoked.~~

Just now refactors the binding replacement out and leaves pretty much everything else intact, also renames a couplemof variables because they were confusingly named before. 

## Addresses
- https://linear.app/budibase/issue/BUDI-8260/automation-conditions-are-not-working-with-booleans-when-using
- https://linear.app/budibase/issue/BUDI-8167/issue-in-automation-collect-data-stepsnsuccess-fails


packages/server/src/automations/steps/filter.ts
```diff
@@ -73,7 +73,12 @@ export async function run({ inputs }: AutomationStepInput) {
   try {
     let { field, condition, value } = inputs
     // coerce types so that we can use them
-    if (!isNaN(value) && !isNaN(field)) {
+    if (
+      !isNaN(value) &&
+      !isNaN(field) &&
+      typeof field !== "boolean" &&
```
adrinr, line 79: Shouldn't this be checking if the `typeof field === "string"`?

packages/server/src/automations/loopUtils.ts
```diff
@@ -0,0 +1,136 @@
...
+  updateContextAndOutput: Function
+) {
+  updateContextAndOutput(loopStepNumber, step, tempOutput, {
+    status: AutomationErrors.MAX_ITERATIONS,
+    success: true,
+  })
+}
+
+export function checkFailureCondition(
+  context: any,
```
adrinr, line 112: Could we type these any? And could we be passing the actual step `currentItem` as a parameter, to make the functions and its parameters more verbose?

packages/server/src/automations/loopUtils.ts
```diff
@@ -0,0 +1,136 @@
+import { AutomationErrors } from "../constants"
+import * as automationUtils from "./automationUtils"
+import env from "../environment"
+import { LoopInput } from "../definitions/automations"
+
+export function initializeTempOutput(loopSteps: any, iterationCount: number) {
+  return { items: loopSteps, iterations: iterationCount }
+}
+
+export function typecastLoopInput(
```
mike12345567, line 10: I don't know if the name of this is quite right - or it should probably return the typecast loop input rather than operating on an in/out parameter mechanism. I assume the boolean returned is to state whether or not it was able to update typecast, but I think a lot of logic could be extracted from this into the caller if all this did was try to typecast and if it fails throw an error. Then the caller could manage setting the incorrect type and whether to proceed.

packages/server/src/threads/automation.ts
```diff
@@ -192,6 +203,57 @@ class Orchestrator {
...
+      )
+    ) {
+      loopStep = undefined
+      loopSteps = undefined
+    }
+
+    originalStepInput = replaceFakeBindings(originalStepInput, loopStepNumber)
+
+    if (checkMaxIterations(index, loopStep?.inputs.iterations)) {
+      handleMaxIterations(
```
samwho, line 235: Similar to `updateContext`, this is only used in one place. I'd just inline this back how it was personally, especially considering this function takes 3 `any`s and an unspecified `Function`.

packages/server/src/automations/loopUtils.ts
```diff
@@ -0,0 +1,136 @@
...
+    updateContextAndOutput(loopStepNumber, step, tempOutput, {
+      status: AutomationErrors.INCORRECT_TYPE,
+      success: false,
+    })
+    return false
+  }
+  return true
+}
+
+export function getLoopItems(binding: any, option: string | undefined): any[] {
```
mike12345567, line 31: Would it be possible to provide better typing for some of this stuff than `any`? It looks like this function handles if the binding is either a string or an array, this could likely be typed with `string | any[]` even if we don't know what the array contains.

packages/server/src/threads/automation.ts
```diff
@@ -192,6 +203,57 @@ class Orchestrator {
...
   }
 
+  handleLooping(
+    loopSteps: LoopStep[] | undefined,
+    iterationCount: number,
+    loopStep: LoopStep | undefined,
+    loopStepNumber: number,
+    index: number,
+    originalStepInput: any,
+    step: any
```
mike12345567, line 213: I feel like in general lot of the `any` is coming from the lack of step/input typing - I think if we fixed these then it would be a lot easier to fill out the typing on the `loopUtils` - would be worthwhile improving the typing as part of this refactor as it just provides a good bit more safety.

samwho (changes requested):
I have to honest I'm afraid and say that I don't think this refactor increases clarity. It reduces the size of the `execute` function, but all of the utilities that have been split out take a large number of parameters, many of them `any` or `Function` (which seems to be to call into instance methods). A large number of disparate parameters suggests tight coupling, and I think we do see that in most of them.

I think a larger-scale refactor of the automation code is what we need. Something that takes all of the step types and creates a handler for each, then the main body of the `execute` function looks like this:

```typescript
function execute(automation: Automation) {
  for (const step of automation.definition.steps) {
    executeStep(step);
  }
}

function executeStep(step: Step) {
  switch(step.stepId) {
    case LOOP_STEP_ID:
      executeLoopStep(step as LoopStep);
    case ...
  }
}
```

Likely with some way to keep state between steps so you can reference the output of each step.

With a larger refactor we can also pay close attention to making sure the types flow throughout the automation code. I think one of the problems at the moment is that the code is so co-dependent and infected with `any` that it's difficult to get types flowing anywhere.

packages/server/src/automations/loopUtils.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+  } else if (typeof value === "string") {
+    value = automationUtils.substituteLoopStep(value, `steps.${loopStepNumber}`)
+  }
+  return value
+}
+
+export function checkMaxIterations(
+  index: number,
+  iterations: string | undefined
```
mike12345567, line 91: `iterations?: string`

packages/server/src/automations/steps/filter.ts
PClmnt: I think we want to specifically check for not boolean here

samwho (approved):

Issue #13792, opened 2024-05-28
Add settings in Nav Links to prompt user before navigating away


andz-bb:
### Discussed in https://github.com/Budibase/budibase/discussions/13783

<div type='discussions-op-text'>

<sup>Originally posted by **mikesealey** May 28, 2024</sup>
When using Actions in the `On Click` of buttons we have the option to prompt a user "Are you sure you want to continue" or similar. 

Occasionally, users can mistakenly click navigation links whilst completing a form, which would result in the loss of any unsaved data. We should add the option to include an "Are you sure. you wish to leave this page?" prompt. 

This could take the form of a single configurable setting called "Confirm Navigation" (or similar), or it could follow the same format as  `Button` `On Click` Actions, with careful consideration given to appropriate permitted/forbidden actions.

![image](https://github.com/Budibase/budibase/assets/39269380/55e1ad85-681e-4406-ab45-c13c4e5ea61a)

Currently, my workaround is to hide the navigation entirely, and rebuild it using _many_ containers, but this leaves a lot to be desired from a maintainability perspective. Example: [Prompt User Before Navigating Away-export-1716890940854.tar.gz](https://github.com/Budibase/budibase/files/15467569/Prompt.User.Before.Navigating.Away-export-1716890940854.tar.gz)

</div>

aptkingston:
I think this would need to be configured on a per-screen basis rather than on the navigation or button action chains, since it's dependent on the content of the screen you're coming from rather than going to. We could then potentially hook into the routing in the client and prompt before any sort of navigation away from a certain page, rather than needing users to configure prompts in multiple different components or links.
It would need something similar to conditional UI though to configure whether or not you want to prompt when leaving, since otherwise it would be pretty annoying to always prompt when leaving a page, even if you touched nothing.

PR #13879, opened 2024-06-07
13792 add settings in nav links to prompt user before navigating away

ConorWebb96:
## Description
Added a confirmation utility function that shows a prompt that must be confirmed before a user can navigate away.

In some cases users may not want to accidentally navigate away from there page as it looses any data that they have added to the form fields. With this feature the intent is to avoid that from happening if you configure your screens correctly. 

## Addresses
- https://github.com/Budibase/budibase/issues/13792

## Screenshots
**Screen setting showcase**
Screenshot 2024-06-07 at 13 34 52

**Confirmation Modal**
Screenshot 2024-06-07 at 13 35 27

## Launchcontrol
Added the option of a modal that must be confirmed if the user wishes to change screens. (Prevents data loss if configured for users.)


aptkingston:
I've pushed up code to https://github.com/Budibase/budibase/tree/navigation-confirmation which allows for intercepting navigation requests at a very low level and preventing them (if desired) without causing any re-renders, meaning you could do thinks like prevent losing work in an unsaved form.

I've designed that in a way that components can specify when and if they want to block navigation - my reasoning being that you don't want to always warn when leaving a screen. It's always conditional on a component being in a certain state, such as only after you've made a change to a form, but not if you haven't made a change.

I'll flesh that out into a real PR at some stage, but we need to decide how to integrate it with components.

My preference is adding it as settings to certain components (just form for now?) as a checkbox ("Confirm discarding unsaved changes" or something), then a checkbox if that is ticked for the prompt (to avoid hardcoded english).

We would then only conditionally prompt users if they've actually modified the form, but not annoy them if they haven't changed it.

That can then be rolled out to whatever other components which make sense to have it.

shogunpurple:
Can we close this @ConorWebb96 based on @aptkingston comment?

PR #13882, opened 2024-06-07
Modernise datasource.spec.ts

samwho:
## Description

I opened `datasource.spec.ts` and quickly realised it could use a little love, so love I have given it.

1. Moved it to using `config.api.datasource` wherever applicable.
2. Removed unused `tk` mock.
3. Moved as many tests as I could into the `describe` that loops over 4 external datasources.
4. Moved away from snapshot testing where it wasn't applicable (surprised that didn't actually cause flakes)
5. Removed all `any`s


mike12345567 (approved):
LGTM - nice improvement!

PR #13883, opened 2024-06-07
Move some mysql.spec.ts and postgres.spec.ts tests into datasource.sp…

samwho:
## Description

I'm on the hunt for tests that live in `mysql.spec.ts` and `postgres.spec.ts` that can reasonably live in more general spec files, where as many external datasource types are being tested.

Going to try and keep these PRs small and focused.


mike12345567 (approved):
LGTM!

PR #13895, opened 2024-06-10
Slim down postgres.spec.ts, most of its content is tested elsewhere now across many datasources.

samwho:
## Description

For everything I deleted in `postgres.spec.ts`, a test existed in another file. I made some minor modification to those tests to bring the quality of the assertions up to what was in `postgres.spec.ts`.


mike12345567 (approved):
Wow - we had a fair amount of duplication, good to clean this up! LGTM

PR #13899, opened 2024-06-10
Various QoL and table fixes

### LINKED ISSUE #13838: When creating a new row in a budibase table via the modal window, no images are uploaded
**Hosting**
Selfhosted 2.27.5 (docker)

**Describe the bug**
When creating a new row in a budibase table via the modal window, my uploaded image is not saved when i save the row

**To Reproduce**
1. create a new app
2. add a budibase table
3. add a "single attachment" column
4. add a new row via the modal form
5. save the row, the image is not visible in the attachment column

**Screenshots**
image
image
image

**Additional context**
Granted, this is a very minor bug, but i am experiencing some weird behavior regarding the "single attachment", most notable randomly images not being uploaded via my app, and previews in the table showing the "no image" icon (not immediately, but after a while), when i open the url i am getting "404 key expired", when i edit a field in the same row the correct image is visible again. Maybe related?

### LINKED ISSUE #13878: Glitch while editing resized number cell in Grid
Latest

to recreate
create an internal db table
create a number column
resize it ( make it smaller than the default )
click to add / edit record 

https://github.com/Budibase/budibase/assets/80512170/d6de0e0f-8c27-47c2-954c-c3c927c47cb3

### LINKED ISSUE #13894: New table validation errors are not displayed to users
**Describe the bug**
New validation errors relating to required fields are not displayed to the user, making it very difficult to work out why things aren't working. I think these were introduced as part of the readonly column work, but I'm not sure. I've just pulled the latest master and couldn't work out why no schema mutations were working. I then realised that the API calls were failing with errors similar to
`To make field "Cost" required, this field must be present and writable in views: gdfg.`

This affected a few of my existing tables so I'm sure other users will be affected too. We just need to surface these new errors to users so they understand what the problems are and can fix them.

Since you were allowed to get into this state previously, but now we block all changes to schema in this state, I would consider this a breaking change since it requires manual intervention to restore functionality to your tables. I think it would have been better to either migrate this automatically for users or grandfather in existing tables with errors so that they continued to work.

I'm already working on a bunch of fixes so will add in a change to surface these errors.

aptkingston:
## Description
A few fixes for bugs identified while using the platform.

Fixes:
- Prevent inline text editing in the preview making any changes unless actually edited. The issue here was that if you used a JS or HBS expression to render your text, double clicking it in the preview would immediately update the saved value to be the resultant plain text string, nuking your expressions. This change means double clicking to edit mode won't do anything by default until you actually type to edit the value.
- Fix an issue where defining any custom column options in table blocks would override the width of all columns, instead of inheriting all other column values.
- Highlight new schema validation errors to users, which prevent any schema modifications being made (fixes https://github.com/Budibase/budibase/issues/13894). Also now handle these errors and revert attempted schema modifications if saving fails.
- Fix issue where the highlighted row in tables underneath the mouse is incorrect when the page is scrolled
- Fix long column names in the new column settings button in grids in the data section having unlimited length. They now truncate properly and have sensible min and max widths. Also tweaked padding a little.
- Fix attachments not working when editing rows in modals (fixes https://linear.app/budibase/issue/BUDI-8315/when-creating-a-new-row-in-a-budibase-table-via-the-modal-window-no and fixes https://github.com/Budibase/budibase/issues/13838) 
- Fix options not clearing properly when adding columns
- Fix issue where inputs can exceed cell bounds in firefox (fixes https://github.com/Budibase/budibase/issues/13878)


deanhannigan (approved):
LGTM! So many awesome enhancements in here! 🚀 The only thing I came across during testing was that the `required` toggle in `CreateEditColumn` modal won't clear when creating consecutive columns. This isn't a new issue but I came across it when trying to verify the options clearing changes.

aptkingston:
> LGTM! So many awesome enhancements in here! 🚀 The only thing I came across during testing was that the `required` toggle in `CreateEditColumn` modal won't clear when creating consecutive columns. This isn't a new issue but I came across it when trying to verify the options clearing changes.

Nice, thanks Dean. Will investigate and fix in a follow up 👍

PR #13902, opened 2024-06-10
Create Tenant endpoint + tenant_info doc

melohagan:
## Description
Part of a larger piece of work to decouple Account from Tenant. 

A **TenantOwner** is the root account of a Tenant. In future this will be tied to a Budibase platform user via the `budibaseUserId`.
We save this data in the `tenant_info` doc of the tenant global DB.

An **Account** owner/holder on the other-hand will relate to billing. This person may not be a platform user, therefore the `budibaseUserId` would ideally not be stored here. 

## Addresses
- https://linear.app/budibase/issue/GROW-469/save-the-tenant-object-to-couchdb

## Screenshots
onboarding
*Onboarding flow*

doc
*CouchDB doc created from the onboarding*

## Launchcontrol
Backend chore


packages/backend-core/src/tenancy/db.ts
```diff
@@ -1,6 +1,24 @@
...
+  const db = getTenantDB(tenantInfo.id)
+  try {
+    // save the tenant info to db
+    await db.put({
+      _id: "tenant_info",
+      ...tenantInfo,
+    })
+  } catch (err: any) {
+    if (err.status === 409) {
+      throw "Tenant info exists already"
```
samwho, line 19: The wording of this is leaving me with questions. Is tenant info write-once? Should it never update?

samwho (approved):

packages/backend-core/src/tenancy/db.ts
melohagan: No longer catching the error.

packages/types/src/documents/global/tenantInfo.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface TenantInfo extends Document {
+  owner: {
+    email: string
+    password?: string
+    ssoId?: string
+    givenName?: string
+    familyName?: string
+    budibaseUserId?: string
+  }
+  id: string
```
mike12345567, line 12: What is this `id` property? How does this differ from the `_id` prop of a document?

packages/worker/src/api/controllers/global/tenant.ts
```diff
@@ -0,0 +1,6 @@
+import { tenancy } from "@budibase/backend-core"
+import { TenantInfo, Ctx } from "@budibase/types"
+
+export const save = async (ctx: Ctx<TenantInfo>) => {
+  await tenancy.saveTenantInfo(ctx.request.body)
```
mike12345567, line 5: I would usually recommend returning something in the body - e.g. the `_id` of the document, and a message to say saved successfully.

mike12345567 (commented):
Looks good - would it be possible to throw a quick `worker/src/api/routes/global/tests/tenant.spec.ts` together just to make sure this does what is expected?

packages/types/src/documents/global/tenantInfo.ts
melohagan: id is the `tenantId` - happy to rename the prop for clarification

melohagan: renamed to `tenantId`

packages/worker/src/api/controllers/global/tenant.ts
melohagan: returning body with _id

PR #13914, opened 2024-06-11
Lock starting containers.

samwho:
## Description

We noticed while debugging a CI problem that it's possible for multiple of the same DB container to be started. This eats up resources for no gain, because everything should connect to just one of the containers, so I'm using `proper-lockfile` to create a filesystem lock when bringing up containers. This should prevent more than 1 starting.


mike12345567 (approved):
LGTM!

mike12345567 (approved):

PR #13921, opened 2024-06-12
UX Changes to Component Name Editing Flow

Ghrehh:
## Description
We discussed in the avalanche standup this week that it wasn't immediately obvious to a user how to edit the name of a component, this PR makes some changes that will hopefully make it more intuitive.

## Changes

- Added an option to the component tree context menu to automatically highlight the name field in the component settings panel.
- The cursor when hovering over the name input is now a `text` cursor, which is the one normally shown for inputs.
- The style of the input now changes to something closer to our typical spectrum input when hovered.

## Notes
- The hotkey for editing a components name is `mod+g` which is a little strange, but every letter in `rename` is a fairly major OS/browser level hotkey a user might be upset if we override (at least on Mac).
- The style of the input on hover is not identical to our normal component settings panel input, I was limited on options due to how the Panel component is designed, at least without a large series of changes/a refactor.

|Info|Screenshot|
|---|---|
|Context menu|Screenshot 2024-06-12 at 09 08 35|
|Editing|Screenshot 2024-06-12 at 09 08 41|
|Clicking directly|Jun-12-2024 09-09-53|


aptkingston (commented):
I like the hover changes (makes it much more obvious that it's editable) but I'm not sure about the context menu addition. I reckon if you already know how to change a name then you'll continue to do it the normal way, and if you don't know how to change a name then I'm not actually sure you'd notice that it focuses the field up there...

I think I would expect that to open a modal or something to change name. But if we did use a modal, then people would probably just assume that's how name changes are done and never discover the inline editing.

Just imagining that if we ship this we'll get a few issues that clicking rename "does nothing" as far as they're concerned, other than select the component.

Happy to get other opinions on that though.

deanhannigan (commented):
LGTM! The design updates on the component name make for a much clearer UX 👍. I agree with @aptkingston regarding the context menu though: I like the how the new addition looks, but I would expect it to open a modal of sorts.

Ghrehh:
@deanhannigan @aptkingston I get what you guys mean, but for my money, the blinking cursor and change to the field upon highlight _should_ be indication enough that the field is selected, it'd be more obvious if we used any sort of `outline` upon focusing an input, but I think we have them all manually disabled.

The main reason I thought it'd be nice to have the context menu option is that despite there being a hover state for the input now, it's still hidden by default. Currently on master, I can easily imagine a user clicking around trying to figure out how to rename a component, and basically going "there's no field for this, there's a field for everything else but not this. You mustn't be able to rename a component". This is kind of exacerbated, imo, by the `Screen` settings panel looking stylistically the same as the component one, but actually having a field to rename it.

But I'm happy to remove the context menu change, I realise this isn't even a requested change, I don't think anyone's even complained about it. So maybe I'm making a mountain out of a molehill 😆 I'll push up a commit now.

deanhannigan (approved):
LGTM!

aptkingston (approved):

PR #13922, opened 2024-06-12
Chart Component and Explanation Modal Fixes and Improvements

Ghrehh:
## Description
Fixes some issues with the chart components and the explanation modal

#### Charts

- Fixed an issue where the format (thousands/millions) option was not being passed from the chart block to the underlying chart component.
- Removed the automatic sorting of datetime labels from Bar/Line/Area charts as it was breaking in certain cases.

#### Explanation Modal
- Fixed an issue where certain columns weren't pointing to their doc pages.
- Fixed an issue where certain columns weren't displaying the correct user facing names.
- Added icons for JSON array and object column types, and both now point to the JSON docs.
- Implemented proper handling of the indefinite article (a/an) in the explanation modal line that mentions column type.
- The explanation modals for the inputs of non-block chart components now link to their individual budibase docs. The chart block links to the chart overview doc page.

### Under the hood

- Some refactoring around how data is passed into the explanation modal from the inputs that use it.


deanhannigan (approved):
LGTM! Really nice updates to the explanation code! 🚀 

There appears to be an issue in master related to charts displaying at the minute, but it precedes your changes.

aptkingston (approved):
LGTM!

PR #13923, opened 2024-06-12
Add endpoint to GET tenant info from tenant global DB.

melohagan:
## Description
GET endpoint for TenantInfo

Removed `check:types` for account-portal, as this is already run in the account-portal workflows. 
The coupling between account-portal and Budibase core is a bit of a risk. The sooner we can separate these the better.

## Launchcontrol
Backend chore


packages/backend-core/src/tenancy/db.ts
```diff
@@ -14,3 +14,8 @@ export async function saveTenantInfo(tenantInfo: TenantInfo) {
     ...tenantInfo,
   })
 }
+
+export async function getTenantInfo(tenantId: string): Promise<TenantInfo> {
+  const db = getTenantDB(tenantId)
+  return await db.get("tenant_info")
```
shogunpurple, line 20: `await` not needed here

packages/worker/src/api/controllers/global/tenant.ts
```diff
@@ -8,3 +8,12 @@ export const save = async (ctx: Ctx<TenantInfo>) => {
     _rev: response.rev,
   }
 }
+
+export const get = async (ctx: Ctx) => {
+  const response = await tenancy.getTenantInfo(ctx.params.id)
+  if (response) {
+    ctx.body = response
+  } else {
+    ctx.throw(404, "Tenant info not found.")
```
shogunpurple, line 17: Does this do what we think it does? If tenant info call throws (404, 500 etc) I believe it will skip this block altogether and throw the error up the chain. Worth testing it

packages/backend-core/src/tenancy/db.ts
melohagan: Removed

.github/workflows/budibase_ci.yml
```diff
@@ -73,9 +73,9 @@ jobs:
       - name: Check types
         run: |
           if ${{ env.USE_NX_AFFECTED }}; then
-            yarn check:types --since=${{ env.NX_BASE_BRANCH }}
+            yarn check:types --since=${{ env.NX_BASE_BRANCH }} --ignore @budibase/account-portal-server --ignore @budibase/account-portal-ui
```
shogunpurple, line 76: Don't think the `ui` is necessary as it doesn't use TS

shogunpurple (approved):

PR #13931, opened 2024-06-12
Expose old row through the emitter

PClmnt:
## Description
As part of some of the work going on to allow users to trigger automations based on specific parts of a row that are updating, we want to expose the old row both through the event emitter (for use in further work) as well exposing the binding directly within the bindings drawer. 

## Addresses
https://linear.app/budibase/issue/BUDI-8335/automation-row-updated-oldrow-data-binding

## Launchcontrol
Exposes a binding that allows the user to see the previous row whenever using an Update Row trigger. Allowing comparison of the previous and new rows in subsequent steps.


packages/server/src/api/controllers/row/external.ts
```diff
@@ -39,8 +39,8 @@ export async function handleRequest<T extends Operation>(
 
 export async function patch(ctx: UserCtx<PatchRowRequest, PatchRowResponse>) {
   const tableId = utils.getTableId(ctx)
+  const oldRow = ctx.request.body
```
mike12345567, line 42: This will contain the new row - not the old row, the old row will only be retrievable by performing a `db.get(rowId)` at the start here, then putting it through `outputProcessing`.

packages/server/src/automations/steps/updateRow.ts
```diff
@@ -35,6 +35,11 @@ export const definition: AutomationStepSchema = {
           customType: AutomationCustomIOType.ROW,
           title: "Table",
         },
+        oldRow: {
+          type: AutomationIOType.OBJECT,
+          customType: AutomationCustomIOType.ROW,
+          title: "Row",
```
mike12345567: should this have a title of "Row" - should it not be "Old Row"?

packages/server/src/api/controllers/row/external.ts
PClmnt: Sorry yes, I need to update that to follow the pattern used in internal.

packages/server/src/automations/triggerInfo/rowUpdated.ts
```diff
@@ -27,6 +27,11 @@ export const definition: AutomationTriggerSchema = {
     },
     outputs: {
       properties: {
+        oldRow: {
+          type: AutomationIOType.OBJECT,
+          customType: AutomationCustomIOType.ROW,
+          description: "The row that was updated",
```
mike12345567, line 33: Likely a better description would be "The row before it was updated"

packages/server/src/automations/steps/updateRow.ts
PClmnt: I honestly wasn't sure about this lol. Where is the title even used - I had assumed the title was to do with the type of the binding (i.e a row) for some reason. I can change it for sure though

packages/server/src/api/routes/tests/automation.spec.ts
```diff
@@ -436,4 +443,36 @@ describe("/automations", () => {
       expect(res).toEqual(true)
     })
   })
+
+  describe("Update Row Old / New Row comparison", () => {
+    it.each([
+      { oldCity: "asdsadsadsad", newCity: "new", expectedResult: false },
+      { oldCity: "Belfast", newCity: "Belfast", expectedResult: true },
```
samwho, line 450: nit: you could very easily derive the `expectedResult` by doing `oldCity === newCity`. If we ever expanded these tests it would make them a fair bit easier to write, removing the scope of potential human error.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -608,6 +618,32 @@ describe.each([
...
+        _id: beforeRow._id!,
+        _rev: beforeRow._rev!,
+        tableId: table._id!,
+        name: "Updated Name",
+      })
+
+      expect(row.name).toEqual("Updated Name")
+      expect(row.description).toEqual(beforeRow.description)
+
+      const event = await rowUpdateEventPromise
```
samwho, line 639: Usually with stuff like this where you have some beginning and then some end and something happens in between I like to wrap it in a callback function like this:

```typescript
const event = await waitForEvent("row:update", async () => {
  const row = await config.api.row.patch(table._id!, {
    _id: beforeRow._id!,
    _rev: beforeRow._rev!,
    tableId: table._id!,
    name: "Updated Name",
  })

  expect(row.name).toEqual("Updated Name")
  expect(row.description).toEqual(beforeRow.description)
})
```

Leaves a little bit less room for mistakes.

```typescript
function waitForEvent<T extends Event>(name: string, callback: () => Promise<void>): Promise<T> {
  const p = new Promise((resolve: any) => emitter.once(name, resolve))
  await callback()
  return p
}
```

Thinking a bit more about this, do we have to be careful that the event emitter is a global thing? So it could be that if tests are running in parallel the event that we get back in here is not the event from the test that we're running, it's from some other test.

samwho (commented):
Just a couple of small comments. While I was reading through the PR I did notice that the event emitting code, the types there have quite a lot of optional fields and the functions have quite a lot of optional arguments which I think is a sign that we should probably take a look at that code at some point and see if we can make it a bit more concrete.

packages/server/src/api/routes/tests/row.spec.ts
PClmnt: > do we have to be careful that the event emitter is a global thing

I think it is in this instance yea - not sure how to handle this exactly. Do you think something like instantiating a new emitter for this specific test would work better? Or what way should we handle this?

samwho: Okay, I think I have a solution.

```typescript
interface WaitOptions {
  name: string,
  matchFn?: (event: any) => boolean,
}
async function waitForEvent(opts: WaitOptions, callback: () => Promise<void>): Promise<any> {
  const p = new Promise((resolve: any) => {
    const listener = (event) => {
      if (opts.matchFn && !opts.matchFn(event)) {
        return
      }
      resolve(event)
      emitter.off(name, listener)
    };
    emitter.on(name, listener)
  })
}
```

This gives you the ability to listen to an event with a function that allows you to filter which event you get back So what we can then do is use this function to match a specific row ID, like this:

```typescript
const opts = { name: "row:updated", matchFn: (event) => event.row._id === row._id }
const event = waitForEvent(opts, () => {
  // do stuff
})
```

I'm not fully sold on the API I've created there, but you see the idea. This gets around the problem of it being global, because you're looking for an event that you know that only you're going to match.

samwho (approved):

mike12345567 (approved):

PR #13940, opened 2024-06-13
Move Role constants from frontend core to shared core

melohagan:
## Description
Refactor

## Addresses
- https://linear.app/budibase/issue/GROW-301/move-role-enum-to-shared-core


mike12345567:
One thing to mention, the backend also has some cases where this is duplicated, to be honest we could move this to an enumeration in `@budibase/types` to make it accessible everywhere, in the backend we have `BUILTIN_ROLE_IDS` and `BUILTIN_IDS` which list these in `backend-core/src/security/roles.ts`.

samwho (commented):
I see Mike has given some advice on this but I want to quickly note that it's probably best to remove the launchcontrol section of the PR description rather than put "chore" in it. We now strip out empty launchcontrol content in Discord so getting rid of it just removes clutter.

packages/backend-core/src/security/roles.ts
```diff
@@ -7,18 +7,11 @@ import {
   doWithDB,
 } from "../db"
 import { getAppDB } from "../context"
-import { Screen, Role as RoleDoc } from "@budibase/types"
+import { Screen, Role as RoleDoc, Roles } from "@budibase/types"
```
mike12345567, line 10: For enums we follow the rule of the enumeration names being singular, this can sometimes read a little bit better, e.g. `Options.One` is a bit weirder than `Option.One`

melohagan: Renamed to `BuiltInRole` to avoid conflict with the `Role` document type.

Issue #13505, opened 2024-04-17
Nested data providers return no results when multiple filters are applied to them.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker-compose
  - Budibase Version: 2.23.6
  - App Version: 2.23.6

**Describe the bug**
Nested filtered data providers return no results even though their filtering logic seem to be correct. This only happens when more than 1 filter is added to the nested data provider. At the top level data provider no issues seem to occur.

**To Reproduce**
1. Import the attached app below
2. Go to the home screen
3. Look at the filters and see that it should return 2 records.

**Expected behavior**
6 records should be shown whenever referencing the nested data provider.

**Screenshots**
Main filter nothing is set returns all rows within a table.
<img width="1512" alt="Screenshot 2024-04-17 at 09 51 59" src="https://github.com/Budibase/budibase/assets/126772285/fae587fc-96eb-4cb1-b213-294529b9a73d">
**Nested filter 2 filters are set** nothing is returned even though every row within main provider is either value a or b.
<img width="1512" alt="Screenshot 2024-04-17 at 09 52 11" src="https://github.com/Budibase/budibase/assets/126772285/a526aba8-df50-432c-a2ce-11962ac2155a">
  
**App Export**
[test nested-export-1713343985090.tar.gz](https://github.com/Budibase/budibase/files/15008505/test.nested-export-1713343985090.tar.gz)

aptkingston:
The issue with this one is that applying filters to a data provider which is using another data provider as its source, will only do fake in-memory filtering. Looking at the code, right now that code does not fully support the "match any" constraint. It will properly check if any operator passes, but for each operator is does a "match all" on all the values.

i.e. this fails because it is 2 filters using the same "equal" operator.

This is a bug in the code so can be fixed. However I would still discourage this setup because it's doing in-memory filtering rather than server-side filtering.

PR #13941, opened 2024-06-13
Test in-memory search alongside all of our other search types, to ensure parity.

samwho:
## Description

This PR makes sure that our in-memory searching that we use in the frontend behaves the same as the backend searches by plugging them into the same test suite the backend ones use.


mike12345567 (approved):
LGTM - nice to get this actually tested!

aptkingston:
Any idea if this PR would address https://github.com/Budibase/budibase/issues/13505#issuecomment-2060860157? I had noticed before we didn't properly support the "match any" function in the in-memory filtering.

samwho:
@aptkingston Yes, I think it probably will. There was a bug in the way that the in-memory filter did matching where it wasn't correctly obeying the allOr param and I fixed that in this PR and I think what you describe in that comment would be fixed by what I did.

aptkingston:
> @aptkingston Yes, I think it probably will. There was a bug in the way that the in-memory filter did matching where it wasn't correctly obeying the allOr param and I fixed that in this PR and I think what you describe in that comment would be fixed by what I did.

Super! Thanks.

Issue #13943, opened 2024-06-14
Close a side panel component if you are no longer on the page were it exists.


ConorWebb96:
**Describe the feature request**
As a user sometimes I change screens using the browsers history. E.g. the forward and back arrows that come with most browsers. Currently it doesn't seem like side panels close whenever changing pages in this way.

I would like a way for these to be closed whenever navigating between pages in this way.

PR #13944, opened 2024-06-14
Add function to close side panel based on the browser's history.

ConorWebb96:
## Description
Adds a function that will close the side panel based on the browser's history changes. E.g. using the forward and back arrows to changes between pages.

## Addresses
- https://github.com/Budibase/budibase/issues/13943

## Launchcontrol
Closes side panel components whenever switching between pages that don't have the one thats currently open.


aptkingston (approved):
I would use the `hashchange` event rather than `popstate`, just because we technically do only use hash bashed routing in client apps (and it has slightly better support for old browsers). I would expect this to function exactly the same though!

PR #13947, opened 2024-06-14
Add the rest of our tests to DataDog.

samwho:
## Description

In https://github.com/Budibase/budibase/pull/13901 we enabled test instrumentation in Datadog and this PR just enables it in the rest of our backend tests.


mike12345567:
Going to close this one - think this may be a little pricey per month, we can turn it on when we need it!

PR #13958, opened 2024-06-17
Remove deprecated properties

melohagan:
## Description
Need to remove these properties so that I can avoid using this type in account-portal

## Addresses
- https://linear.app/budibase/issue/GROW-472/remove-use-of-types-package-in-account-portal


samwho:
https://github.com/Budibase/budibase/pull/13958/commits/943a6199b12cda4aca2f6cc94a723610a116b0c8

This seems like a fairly large change, is this something that we've discussed wider and agreed on?

melohagan:
> [943a619](https://github.com/Budibase/budibase/commit/943a6199b12cda4aca2f6cc94a723610a116b0c8)
> 
> This seems like a fairly large change, is this something that we've discussed wider and agreed on?

No it hasn't been. I've been trying to detangle the types between Budibase core and account-portal.

What I can do is merge this in to remove the **tier** property, update account-portal to no longer use this property, and then re-add the account-portal tests run back into the Budibase core CI again?

samwho (approved):

PR #13960, opened 2024-06-17
Use `knex` instead of `rawQuery` everywhere, remove `rawQuery` from codebase.

samwho:
## Description

This PR moves our tests to using `knex` in favour of our custom `rawQuery`. `knex` has a `knex.raw()` method we can use if we need to issue raw SQL, and those instances should be few and far between. `knex`'s DSL for running queries should be preferred everywhere.

### DIFF
--- a/packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
+++ b/packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
@@ -10,37 +10,11 @@ import * as setup from "../utilities"
 import {
   DatabaseName,
   getDatasource,
-  rawQuery,
+  knexClient,
 } from "../../../../integrations/tests/utils"
 import { Expectations } from "src/tests/utilities/api/base"
 import { events } from "@budibase/backend-core"
-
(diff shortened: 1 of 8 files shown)


packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -82,21 +57,34 @@ describe.each(
...
+    // The Datasource API doesn ot return the password, but we need it later to
+    // connect to the underlying database, so we fill it back in here.
     datasource.config!.password = rawDatasource.config!.password
 
-    await rawQuery(datasource, createTableSQL[datasource.source])
-    await rawQuery(datasource, insertSQL)
+    client = await knexClient(rawDatasource)
+
+    await client.schema.dropTableIfExists("test_table")
+    await client.schema.createTable("test_table", table => {
```
mike12345567, line 67: this is class - so good!

mike12345567 (approved):
LGTM - this is such an improvement, so much easier than the raw queries!

PR #13965, opened 2024-06-18
Table UI changes

aptkingston:
## Description
This PR adds a couple of UI adjustments to the table.

- The space around the inside of the table has been reduced further, to as little as possible while still allowing space for the necessary controls
- The "quiet" table setting now hides more borders, making it more quiet
- An empty cell has been added to the new row component in the button column (if one exists) to make layout more consistent
- A few small pixel adjustments for consistency
- The render position of the new row component now updates in real time, in case you scroll vertically using the scrollbar while adding a new row
- Vertical padding takes into account whether a horizontal scrollbar exists or not, meaning it reduces it as much as possible depending on the current needs of the table


deanhannigan (approved):
LGTM! Really nice refinements to the table!

Issue #13963, opened 2024-06-18
Add the option to change the confirm and cancel button texts within the confirmation modals.


ConorWebb96:
**Describe the feature request**
As a user sometimes I would like to change the text of the confirm and cancel buttons within modals. This is more so for users whom want the button text to be displayed in their native language.

PR #13966, opened 2024-06-18
Add the option to change the confirm and cancel button texts within the confirmation modals

ConorWebb96:
## Description
This adds the ability to change the confirm button texts of both cancel and confirm within the confirmation modals for actions like `Save Row`, `Duplicate Row` etc.

## Addresses
- https://github.com/Budibase/budibase/issues/13963

## Launchcontrol
Minor change to add in another 2 fields to allow the changing of button's text within the action confirmation modals.


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/DeleteRow.svelte
```diff
@@ -53,6 +53,12 @@
         placeholder="Are you sure you want to delete?"
         bind:value={parameters.confirmText}
       />
+
+      <Label small>Confirm Text</Label>
+      <Input placeholder="Confirm" bind:value={parameters.confirmButtonText} />
+
+      <Label small>Cancel Text</Label>
+      <Input placeholder="Cancel" bind:value={parameters.confirmCancelText} />
```
melohagan, line 61: this parameter should be called `cancelButtonText` instead of `confirmCancelText`

melohagan (approved):
LGTM! Just a variable name suggestion

PR #13968, opened 2024-06-18
Support external datasources in the bulkImport endpoint.

samwho:
## Description

This is the backend half of https://linear.app/budibase/issue/BUDI-8220/support-updating-existing-rows-for-external-databases-using-csv-import. It adds backend support for doing bulk imports of rows, supporting upserts against keys that already exist.

The only wrinkle is that we can't support upserts for SQL Server yet because `knex` does not have support for it. There's a PR open adding support for it that has some recent activity, so perhaps this will be something we can enable in the near future: https://github.com/knex/knex/pull/6050.


packages/server/src/api/controllers/table/external.ts
```diff
@@ -98,7 +98,7 @@ export async function bulkImport(
     table = processed.table
   }
 
-  await handleRequest(Operation.BULK_CREATE, table._id!, {
+  await handleRequest(Operation.BULK_UPSERT, table._id!, {
```
mike12345567, line 101: is there any use in the `BULK_CREATE` operation anymore for SQL? Should we just have both of those operations go through the bulk upsert process?

mike12345567 (approved):

packages/server/src/api/controllers/table/external.ts
samwho: It appears to still be used in the Google Sheets integration.

samwho: Oh, in SQL. I'm not quite sure what you mean... What do you mean by "both of those"?

PR #13969, opened 2024-06-18
Counting total rows in query (Datasource+)

mike12345567:
## Description
This has been a feature that has been desirable for quite some time, but up until now it was difficult to implement due to our Lucene search implementation limitation on internal tables.

With SQS this is no longer an issue, so adding support for counting the rows that will come back in query, as well as fixing some other issues with pagination.

I've also updated the external table implementation to use a single query to work out if another page is available, previously it would make the main query and then make a secondary query for a single row. This should help quite a bit for reading rows from SQL DBs with relatively high latency.

I've made this an option for the search API, you can set a `countRows: boolean` property which will include the total rows that would be returned after all pages have been searched.

This was requested by @aptkingston for improved grid integrations.


packages/backend-core/src/sql/sql.ts
```diff
@@ -602,24 +605,33 @@ class InternalBuilder {
...
       foundLimit = paginate.limit
     }
-    // start building the query
-    let query = this.knexWithAlias(knex, endpoint, tableAliases)
-    query = query.limit(foundLimit)
-    if (foundOffset) {
+    // always add the found limit, unless counting
+    if (!counting) {
+      query = query.limit(foundLimit)
+    }
```
samwho, line 611: The large number of `if (!counting)` conditions in this function makes me feel like this should actually be 2 separate functions. What do you think?

packages/backend-core/src/sql/sql.ts
```diff
@@ -602,24 +605,33 @@ class InternalBuilder {
...
     // add sorting to pre-query
-    query = this.addSorting(query, json)
+    if (!counting) {
+      query = this.addSorting(query, json)
+    }
     const alias = tableAliases?.[tableName] || tableName
     let preQuery = knex({
       [alias]: query,
-    } as any).select(selectStatement) as any
+    } as any)
```
samwho, line 627: `as any` 👀

packages/backend-core/src/sql/sql.ts
```diff
@@ -602,24 +605,33 @@ class InternalBuilder {
...
+    if (counting) {
+      preQuery = preQuery.count("* as total")
+    } else {
+      preQuery = preQuery.select(selectStatement)
+    }
     // have to add after as well (this breaks MS-SQL)
-    if (this.client !== SqlClient.MS_SQL) {
+    if (this.client !== SqlClient.MS_SQL && !counting) {
```
samwho, line 634: It just occurred to me, do we not allow sorting for SQL Server? Because before you added the condition at the end here, this was never adding sorting for SQL Server. Is that right?

packages/backend-core/src/sql/sql.ts
```diff
@@ -630,6 +642,12 @@ class InternalBuilder {
       endpoint.schema,
       tableAliases
     )
+
+    // add a base query over all
```
samwho, line 646: This comment doesn't make sense.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -663,11 +671,17 @@ export class ExternalRequest<T extends Operation> {
 
     // aliasing can be disabled fully if desired
     let response
-    if (env.SQL_ALIASING_DISABLE) {
-      response = await getDatasourceAndQuery(json)
+    const aliasing = new sdk.rows.AliasTables(Object.keys(this.tables))
+    // if it's a counting operation there will be no more processing, just return the number
+    if (this.operation === Operation.COUNT) {
+      return (await aliasing.countWithAliasing(
```
samwho, line 677: How come we're not respecting the aliasing disable environment variable here?

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -105,6 +106,17 @@ describe.each([
...
+    private async performSearchFullResponse(): Promise<SearchResponse<Row>> {
+      if (isInMemory) {
+        return { rows: dataFilters.search(_.cloneDeep(rows), this.query) }
+      } else {
+        return config.api.row.search(table._id!, {
+          ...this.query,
+          tableId: table._id!,
+        })
+      }
+    }
```
samwho, line 118: I think it would be better to delete this and make the other performSearch function return the full response and then pull the rows out in each of the places where it's called. I don't think it makes sense to have two of these that are almost entirely identical.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -1797,4 +1837,40 @@ describe.each([
           { two: "foo", other: [{ _id: otherRows[0]._id }] },
         ]))
     })
+
+  // lucene can't count, and in memory there is no point
```
samwho, line 1841: I would still test this for in-memory.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -1797,4 +1837,40 @@ describe.each([
...
+
+      it("should be able to count rows when option set", () =>
+        expectSearch({
+          countRows: true,
+          query: {
+            notEmpty: {
+              name: true,
+            },
+          },
+        }).toHaveProperty([{ key: "totalRows", value: 2 }, { key: "rows" }]))
```
samwho, line 1863: Is this not just the jest `toMatch` assertion? Could we implement `expectSearch({ ... }).toMatch({ totalRows: 2, rows: expect.any(Array) })` so it fits more naturally into jest?

packages/server/src/api/routes/utils/validators.ts
```diff
@@ -109,6 +109,7 @@ export function internalSearchValidator() {
       sortOrder: OPTIONAL_STRING,
       sortType: OPTIONAL_STRING,
       paginate: Joi.boolean(),
+      countRows: Joi.boolean(),
```
samwho, line 112: Is this not optional?

packages/server/src/integrations/base/query.ts
```diff
@@ -8,8 +8,8 @@ import { getIntegration } from "../index"
 import sdk from "../../sdk"
 
 export async function makeExternalQuery(
-  datasource: Datasource,
-  json: QueryJson
+  json: QueryJson,
+  datasource?: Datasource
```
samwho, line 12: I don't think I understand why these have been flipped around and datasource has been made optional. You're always going to need a datasource here, right?

packages/server/src/sdk/app/rows/search/external.ts
```diff
@@ -69,24 +75,22 @@ export async function search(
...
       sort,
       paginate: paginateObj as PaginationJson,
       includeSqlRelationships: IncludeRelationship.INCLUDE,
-    })
+    }
+    let rows = await handleRequest(Operation.READ, tableId, parameters)
+    let totalRows: number | undefined
+    if (countRows) {
+      totalRows = await handleRequest(Operation.COUNT, tableId, parameters)
+    }
```
samwho, line 88: Is it worth structuring these in such a way that you can issue both of the `handleRequest` calls and wrap them in a `Promise.all` so that they're executing at the same time?

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -95,14 +95,29 @@ function buildTableMap(tables: Table[]) {
     // update the table name, should never query by name for SQLite
     table.originalName = table.name
     table.name = table._id!
+    // need a primary for sorting, lookups etc
+    table.primary = ["_id"]
```
samwho, line 99: Do we not need a check here to see if the table already has a primary key set?

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -95,14 +95,29 @@ function buildTableMap(tables: Table[]) {
...
   return tableMap
 }
 
-async function runSqlQuery(json: QueryJson, tables: Table[]) {
+function runSqlQuery(json: QueryJson, tables: Table[]): Promise<Row[]>
+function runSqlQuery(
+  json: QueryJson,
+  tables: Table[],
+  opts: { countTotalRows: boolean }
+): Promise<number>
```
samwho, line 110: Is this correct? What if they specify countTotalRows as false? It wouldn't return a number in that case, right?

samwho (commented):
Honestly, I'm not convinced this is the right approach. There are so many conditionals scattered throughout the code now to handle counting, it really does feel to me like it should be a separate endpoint.

packages/server/src/api/routes/tests/search.spec.ts
mike12345567: It doesn't have any implementation for counting, do you think we should add it?

packages/server/src/api/routes/utils/validators.ts
mike12345567: Joi properties are optional by default.

packages/server/src/sdk/app/rows/search/sqs.ts
mike12345567: This will never happen in for internal tables (SQS) but I can add a throw just incase.

packages/backend-core/src/sql/sql.ts
mike12345567: We do allow sorting on SQL Server, essentially if a sort isn't applied it always applies a base sort (I believe).

samwho (approved):

packages/backend-core/src/sql/sql.ts
```diff
@@ -410,13 +410,32 @@ class InternalBuilder {
...
   addSorting(query: Knex.QueryBuilder, json: QueryJson): Knex.QueryBuilder {
-    let { sort, paginate } = json
+    let { sort } = json
     const table = json.meta.table
+    const primaryKey = table.primary
     const tableName = getTableName(table)
     const aliases = json.tableAliases
     const aliased =
       tableName && aliases?.[tableName] ? aliases[tableName] : table?.name
+    if (!Array.isArray(primaryKey)) {
```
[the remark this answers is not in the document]
mike12345567: In this case we want it to be empty, its not an array e.g. there is no primary key, an empty array (zero length) is invalid as well.

PR #13977, opened 2024-06-19
Adds _id and _rev back to internal datasource filter options

PClmnt:
## Description
The ability to select _id and _rev columns when filtering a table were missing due to the fact we now use the table schema which doesn't include these. Just adding them manually on the frontend so they can be usable again.


packages/frontend-core/src/components/FilterBuilder.svelte
```diff
@@ -28,14 +28,30 @@
   export let allowBindings = false
   export let filtersLabel = "Filters"
 
+  $: {
+    if (
+      tables.find(
+        table =>
+          table._id === datasource.tableId &&
+          table.sourceId === "datasource_internal_bb_default"
```
PClmnt, line 36: Just need to make this a const somewhere

packages/shared-core/src/constants/index.ts
```diff
@@ -179,3 +179,5 @@ export enum BpmStatusValue {
   VERIFYING_EMAIL = "verifying_email",
   COMPLETED = "completed",
 }
+
+export const DEFAULT_BB_DATASOURCE_ID = "datasource_internal_bb_default"
```
mike12345567, line 183: Is this duplicated, I believe there is somewhere in backend-core or server that defines this as well?

PClmnt: Yea I wanted to add it in to frontend-core as it didn't exist there, so thought it made sense to add to shared-core for that purpose. I can't import from backend-core or server in shared-core can I?

mike12345567: You can remove it from everywhere else though and then import it from shared-core to be exported from backend-core - in `backend-core/src/constants/db.ts` you can do `export { DEFAULT_BB_DATASOURCE_ID } from "@budibase/shared-core`

PClmnt: Sorted this now @mike12345567

mike12345567 (approved):
LGTM!

PR #13979, opened 2024-06-19
[BUDI-8220] Support updating existing rows for external databases using CSV import

samwho:
## Description

So before this PR, you could only import a CSV file into internal tables. What this PR implements is the ability to import a CSV file into external tables.

One of the differences that had to be put in for this for external tables is the only way I could get this to work is if it uses the `ON CONFLICT` functionality of SQL which requires the primary keys to be the keys used for deciding whether or not to create a new row. So with the internal version of this, you can choose any columns you want. With the external version, you're forced into using the primary key. So the front end has been updated to reflect that.

Additionally, I couldn't get this to work with MSSQL so it's disabled there for now.

## Addresses
- https://linear.app/budibase/issue/BUDI-8220/support-updating-existing-rows-for-external-databases-using-csv-import

## Screenshots

https://github.com/Budibase/budibase/assets/338027/93eb1d4b-0158-43fa-8aba-8cbab572dc97

## Launchcontrol

You can now import CSV data and update existing rows by their primary key with external data sources.


packages/builder/src/components/backend/TableNavigator/ExistingTableDataImport.svelte
```diff
@@ -185,20 +185,22 @@
...
-    <br />
-    <Toggle
-      bind:value={updateExistingRows}
-      on:change={() => (identifierFields = [])}
-      thin
-      text="Update existing rows"
-    />
-    {#if updateExistingRows}
+  <br />
+  <Toggle
```
samwho, line 189: @aptkingston I'd like to hide this toggle when the datasource is MSSQL, but I'm not sure how to best do that. I don't want to call out to the datasources API because I can see in the Chrome network tab that we already fetch the datasources, so I'm guessing there's some way to get at that data but I just don't know how. Halp? 😇

mike12345567: The datasources will be available at all times in components like this in the stores, you can get the datasource by its ID using:
```
import { datasources } from "stores/builder"
```

from here you can just get the datasource using the svelte store fetch function `$` - like:
```
const datasourceList = $datasources.list
```

You can then search this list by `_id` to find the datasource this is occurring for, the store also contains the current datasourceId under `$datasources.selectedDatasourceId` - if this is set then it is an external datasource and you can pick it out of the list, then check its sourceType!

aptkingston: You're correct - we do indeed already have the required info. You'd achieve this by doing:
```js
import { tables, datasources } from "stores/builder"

export let tableId // already have this prop

$: table = $tables.list.find(table => table._id === tableId)
$: datasource = $datasources.list.find(ds => ds._id === table?.sourceId)
```

packages/server/src/api/controllers/table/external.ts
```diff
@@ -16,6 +16,7 @@ import {
 import sdk from "../../../sdk"
 import { builderSocket } from "../../../websockets"
 import { inputProcessing } from "../../../utilities/rowProcessor"
+import _ from "lodash"
```
mike12345567, line 19: usually preferred to use `import { isEqual } from "lodash"` as it allows just that function to be pulled in here as part of the esbuild.

packages/server/src/api/controllers/table/external.ts
```diff
@@ -82,15 +83,30 @@ export async function bulkImport(
...
+    identifierFields.length > 0 &&
+    !_.isEqual(identifierFields, table.primary)
+  ) {
+    // This is becuse we make use of the ON CONFLICT functionality in SQL
+    // databases, which only triggers when there's a conflict against a unique
+    // index. The only unique index we can count on atm in Budibase is the
+    // primary key, so this functionality always uses the primary key.
+    ctx.throw(
+      400,
+      "Identifier fields are not supported for bulk import into an external datasource."
```
mike12345567, line 100: Would it not be better here to give an exact error "Only primary keys can be used for bulk importing of rows"- something that should be obvious to the user.

samwho: This error should, in theory, never be shown to the user because the UI doesn't let them specify the identifier fields. Happy to change it either way, but the idea is they'd never see it 😅

packages/builder/src/components/backend/TableNavigator/ExistingTableDataImport.svelte
samwho: Thank you very much, gentlemen 🙏

mike12345567 (approved):
LGTM!

PR #13993, opened 2024-06-20
Disallow prohibited columns

mike12345567:
## Description
As the title - this was already done, but it was case sensitive. Budibase columns are case in-sensitive, meaning that the prohibited columns could still be added as long as they were a different casing - this has been rectified, backend validation added and test cases for it added.

### DIFF
--- a/packages/server/src/api/routes/tests/table.spec.ts
+++ b/packages/server/src/api/routes/tests/table.spec.ts
@@ -276,6 +276,31 @@ describe.each([
         })
     })
 
+    it("shouldn't allow duplicate column names", async () => {
+      const saveTableRequest: SaveTableRequest = {
+        ...basicTable(),
+      }
+      saveTableRequest.schema["Type"] = { type: FieldType.STRING, name: "Type" }
+      await config.api.table.save(saveTableRequest, {
+        status: 400,
+        body: {
+          message:
+            'Column "type" is duplicated - make sure there are no duplicate columns names, this is case insensitive.',
+        },
+      })
+      saveTableRequest.schema = {
+        foo: { type: FieldType.STRING, name: "foo" },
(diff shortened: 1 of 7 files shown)


packages/server/src/api/routes/tests/table.spec.ts
```diff
@@ -276,6 +276,35 @@ describe.each([
...
+          type: FieldType.STRING,
+          name: "Type",
+        }
+        await config.api.table.save(saveTableRequest, {
+          status: 400,
+          body: {
+            message:
+              'Column "type" is duplicated - make sure there are no duplicate columns names, this is case insensitive.',
+          },
+        })
```
samwho, line 294: I don't feel too strongly about this but I'd probably get rid of this section of the test. It doesn't test anything that the foo test below doesn't test and it depends on a detail of the basicTable method that may potentially change.

packages/shared-core/src/table.ts
```diff
@@ -51,3 +52,22 @@ export function canBeDisplayColumn(type: FieldType): boolean {
 export function canBeSortColumn(type: FieldType): boolean {
   return !!allowSortColumnByType[type]
 }
+
+export function findDuplicateInternalColumns(table: Table): string | undefined {
```
samwho, line 56: For the sake of a slightly better error message, is it worth here returning a list of all duplicated columns? So in the error we can give the user all the columns they need to pay attention to?

samwho (approved):

packages/server/src/api/routes/tests/table.spec.ts
mike12345567: Added some more usefulness to it - this checks for prohibited columns and then just straight duped columns, as well as confirming multiple can be errored on.

PR #13998, opened 2024-06-21
Allow filtering on relationship column, or related table name

mike12345567:
## Description
This is a two part fix as well as a minor test update to cover these problems:

1. A suggestion from Poirazis to list the column names of the relationship rather than the table name, allowing either option to be used from the API, will need to update frontend to use the relationship column names as this might add some extra context.

2. Column names or table names within table names, for example if I have a `ProductTasks` and `Product`, if I try to filter something on a relationship to `ProductTasks` it might swap the product component with the product table ID and not update the tasks. I've solve this by sorting the lists by longest first, so that we replace using the longest best fit, which avoids sub-strings.


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -248,6 +163,117 @@ export class ExternalRequest<T extends Operation> {
...
+    // replace any relationship columns initially, table names and relationship column names are acceptable
+    const relationshipColumns = sdk.rows.filters.getRelationshipColumns(table)
+    filters = sdk.rows.filters.updateFilterKeys(
+      filters,
+      relationshipColumns
+        .map(({ name, definition }) => {
+          const { tableName } = breakExternalTableId(definition.tableId)
+          return {
+            original: name,
+            updated: tableName!,
```
samwho, line 184: It's a bit annoying that this is necessary. It seems like `breakExternalTableId` is one of those functions that could take nothing and if it does it returns nothing. And I can see the value in that leniency, but it does create situations like this where you have to override the type checker. I wonder if it would be difficult to make `breakExternalTableId` only take known strings rather than undefined or to type it in such a way that if it doesn't receive undefined it doesn't return an undefined result.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -248,6 +163,117 @@ export class ExternalRequest<T extends Operation> {
...
+        // don't update table names - include this for context incase a column would be replaced
+        .concat(
+          tables.map(table => {
+            const tableName = table.originalName || table.name
+            return {
+              original: tableName,
+              updated: tableName,
+            }
+          })
+        )
```
samwho, line 196: I'm not sure that I understand the purpose of this. Was this in place for when we were doing the sorting? And now that we've fixed that, it's not necessary?

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -248,6 +163,117 @@ export class ExternalRequest<T extends Operation> {
...
+    )
+    const primary = table.primary
+    // if passed in array need to copy for shifting etc
+    let idCopy: undefined | string | any[] = cloneDeep(id)
+    if (filters) {
+      // need to map over the filters and make sure the _id field isn't present
```
samwho, line 202: I know that you didn't write this, but how come? I can't think of a good reason you'd need to do this.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -248,6 +163,117 @@ export class ExternalRequest<T extends Operation> {
...
+    }
+    // there is no id, just use the user provided filters
+    if (!idCopy || !table) {
+      return filters
+    }
+    // if used as URL parameter it will have been joined
+    if (!Array.isArray(idCopy)) {
+      idCopy = breakRowIdField(idCopy)
+    }
+    const equal: any = {}
```
samwho, line 228: Again, I know this is just moved, but is there a way we can be more specific than any here?

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -260,10 +286,22 @@ export class ExternalRequest<T extends Operation> {
...
+  async retrieveMetadata(datasourceId: string) {
+    if (!this.datasource) {
+      this.datasource = await sdk.datasources.get(datasourceId)
+      if (!this.datasource || !this.datasource.entities) {
+        throw "No tables found, fetch tables before query."
+      }
+      this.tables = this.datasource.entities
+      this.tableList = Object.values(this.tables)
+    }
+  }
```
samwho, line 299: It's a shame that this removes the type narrowing on this.datasource. Is there any way to retain that? For example, could we return the datasource from this function and assign it to this.datasource at the calling location? Because right now you've had to add an exclamation mark to the end of this.datasource at the call site.

packages/server/src/api/controllers/row/utils/basic.ts
```diff
@@ -99,7 +99,7 @@ export function basicProcessing({
         row,
         tableName: table._id!,
         fieldName: internalColumn,
-        isLinked: false,
```
samwho, line 102: Was this just an oversight?

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -1950,59 +1950,77 @@ describe.each([
...
+            name: "baz",
+            productCat: [],
           }),
         ])
+      })
 
-        rows = await config.api.row.fetch(table._id!)
+      it("should be able to filter by relationship using column name", async () => {
+        await expectQuery({
+          equal: { ["productCat.name"]: "foo" },
```
samwho, line 2007: I think this change makes it possible to run this specific variant of the test against in-memory search. I'm not sure exactly how best to structure the test to do that, but I think it's worth doing.

packages/server/src/sdk/app/rows/search/filters.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function updateFilterKeys(
+  filters: SearchFilters,
+  updates: { original: string; updated: string }[]
+): SearchFilters {
+  const makeFilterKeyRegex = (str: string) =>
+    new RegExp(`^${str}\\.|:${str}\\.`)
+  for (let filter of Object.values(filters)) {
+    if (typeof filter !== "object") {
```
samwho, line 35: I don't think this is quite doing what you're hoping it to do. I fell over this myself recently, for example:

```javascript
> typeof null
'object'
> typeof []
'object'
```

I _think_ you want https://lodash.com/docs/4.17.15#isPlainObject because even `_.isObject` returns true for arrays.

packages/server/src/sdk/app/rows/search/filters.ts
```diff
@@ -0,0 +1,54 @@
...
+      name: entry[0],
+      definition: entry[1] as RelationshipFieldMetadata,
+    }))
+}
+
+export function getTableIDList(
+  tables: Table[]
+): { name: string; id: string }[] {
+  return tables
+    .filter(table => table.originalName)
```
samwho, line 24: For completeness, do you also want to filter on the _id field here? Because you use it in the map below.

packages/server/src/sdk/app/rows/search/filters.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function getRelationshipColumns(table: Table): {
+  name: string
+  definition: RelationshipFieldMetadata
+}[] {
+  return Object.entries(table.schema)
+    .filter(entry => entry[1].type === FieldType.LINK)
+    .map(entry => ({
+      name: entry[0],
+      definition: entry[1] as RelationshipFieldMetadata,
```
samwho, line 16: Does this cast create a future danger that if we have subtypes of FieldType.LINK they may not be a RelationshipFieldMetadata?

mike12345567: TLDR - I typed a long message about this, relationship fields already have some subtypes, this is the base type of all relationships already - however it would be nice if type narrowing could actually achieve this (with raw array filtering it can't without a cast of some format).

I've changed how this is done so that type narrowing achieves it, but it is a bit of a painful one with union types.

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: Just went through and updated this - it wasn't actually too bad!

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: the `_id` field doesn't exist in the external DB, so it needs to be extracted so that it can be handled as a separate case, rather than passed down to the SQL layer to generate, we convert it to a filter based on the primary keys (the later part of this function).

samwho (approved):

PR #13999, opened 2024-06-21
Release SQS for self-host users (single image, docker compose, Kubernetes)

samwho:
## Description

This PR will release SQS for people using the Docker Compose and Budi CLI self-hosting methods. So the next time that they update, they will have an SQS enabled deployment. This will mean that the first time they load an app after running this update, there will be a one-time migration to make their data compatible with SQS.

Because we're not quite ready to release this into the wild yet, I'm going to leave this PR in a draft state so it doesn't get merged by accident.


packages/cli/src/hosting/utils.ts
```diff
@@ -46,7 +46,8 @@ export function setServiceImage(service: string, image: string) {
 export async function downloadDockerCompose() {
   const filename = composeFilename()
   try {
-    await downloadFile(COMPOSE_URL, `./${filename}`)
+    fs.copyFileSync("../../hosting/docker-compose.yaml", `./${filename}`)
+    //await downloadFile(COMPOSE_URL, `./${filename}`)
```
mike12345567, line 50: remove comment

mike12345567 (approved):

mike12345567 (approved):

mike12345567 (approved):
LGTM - thanks for test case!

mike12345567 (approved):

PR #14004, opened 2024-06-24
fix issue where schema wasn't updating types when a query was run

PClmnt:
## Description
Fixes an issue where the schema wasn't being updated correctly if the types had changed. 

We were assigning the existing schema back to the preview schema if the key existed at all, which meant that if the schema had a new type, (i.e it had changed say from string -> number) then it was just using the old type (string). This simply changes it so that if the key doesn't exist in the new schema then assign it from the existing schema (i.e a manually added key)
## Addresses
https://linear.app/budibase/issue/BUDI-8296/queries-dont-re-fetch-the-correct-schema-if-you-re-send-to-build-it

## Launchcontrol
Fixes an issue where schema types weren't being updated correctly if your query response changed


packages/server/src/api/controllers/query/index.ts
```diff
@@ -311,8 +311,8 @@ export async function preview(
 
   // if existing schema, update to include any previous schema keys
   if (existingSchema) {
-    for (let key of Object.keys(previewSchema)) {
-      if (existingSchema[key]) {
+    for (let key of Object.keys(existingSchema)) {
```
mike12345567, line 314: This does seem like quite a silly mistake, logically I can see how this was completely wrong - is there anyway we could put a quick test together for this to confirm that it works? This feels like a logical mistake which could easily regress.

mike12345567 (approved):

PR #14006, opened 2024-06-24
Fix missing operator options when filtering on relationships

aptkingston:
This PR fixes the issue where there were no operator options when attempting to filter on relationships in both the design section and automations. I've also removed a weird double invocation of `getFields` which seemed to do nothing.

Fixes https://linear.app/budibase/issue/BUDI-8389/builder-data-provider-filters-empty-operator-for-relationships


deanhannigan (approved):
LGTM! Had a test run through an internal db and an external PostgreSQL db; everything looked grand 👍

PR #14023, opened 2024-06-26
Multi-cell operations for tables

aptkingston:
**Sorry about the size of this PR. Apart from a few tiny changes, this is entirely self-contained to the table codebase and is therefore much lower risk than the size of the PR would imply. Manual testing to verify stuff works as expected is more useful than a code review, probably.**

This PR adds multiple features to tables, in both the data section and apps. Most of these features aim to provide a more "spreadsheet-like" experience when manipulating your data.

Main new features:
- Multi-cell selection
- Copy and paste multiple cells at once (max 50 rows, unlimited columns)
- Delete multiple cells at once (max 50 rows, unlimited columns)
- Duplicate multiple rows at once (max 50)

All of these new multi-cell/multi-row operations work using sequential row save calls, throttled to ensure we never hit rate limits. This is also why they're capped to 50 rows. I'd love to use proper bulk update endpoints, but I had a look and adding relationship support to the current import endpoint (the closest to a bulk update we have) is not trivial (confirmed with @mike12345567). If we had an endpoint for this in future, I will absolutely update the table to use that endpoint and remove the limits.

## Multi-cell selection

https://github.com/Budibase/budibase/assets/9075550/f398e5a5-1334-4113-b639-6c7b5af21912

Multiple cells can now be selected at once. This is how most of the new features operate - select a bunch of cells, then do something with them.

Cells can be selected in a few different ways:
- Click and drag
- Select a cell, then select another cell while holding shift. All cells in between will be selected.
- Hold shift, then use the arrow keys. This will expand the current selection by 1 row or 1 column, depending on the arrow pressed.

### Copying
Select a range of cells, then either press ctrl+c/cmd+c or right click and choose "Copy" in the context menu, just like copying the value of a single cell.

image

### Pasting
How paste works will depend on what data you've copied.

If you've copied multiple cell values, then paste will work in one of the following ways:
- If you've selected multiple cells to paste into, then it will attempt to fill those cells as well as possible. If you have copied more columns than you are pasting, it will simply use the first N columns. If you have not copied enough columns, then it will only paste as many columns as you have copied. For rows, it will repeat the copied rows as required to fill the desired cells to paste into
- If you only have one cell selected, the entire extent of the pasted data will be pasted (space allowing). The updated cells will be selected to show you the extent of the updated data.

If you've copied a single cell value and have multiple cells selected while pasting, then that value will be duplicated into all selected cells.

### Deleting
Pressing delete or backspace, or right clicking and choosing the menu option, will attempt to delete all selected cells.

### Progress indicators
All multi-cell operations have progress indicators, to keep you updated and progress and provide insight into the expected total duration. The progress is accurate, and is not just an estimate.

https://github.com/Budibase/budibase/assets/9075550/7c18a4b1-502e-4729-b287-cdca6c35f062

## Multi-row duplication
If you have multiple rows selected (which you can now do by holding shift and selecting another row), then you'll get a new option in the context menu to duplicate all those rows. This will highlight the new rows after duplication, and also has a progress indicator.

https://github.com/Budibase/budibase/assets/9075550/3d8e5981-33c9-48d4-9c86-7380ea0657e5

## Error handling
All of these new operations function on a per-row basis, and will handle errors gracefully. If any rows fail to save then the user will be notified, but the other rows will still be saved. Basically it will try and save as much as possible. At the end of the operation, the notifications will include information on how many rows saved or failed to save.

## Other stuff
There's also a bunch of refactoring in here - cleaning up and simplifying large parts of the code base. The table should feel faster over all as many wasted store updates have been removed, and the refactored column handling logic should also reduce reactive statement triggers.

There are also a few visual improvements, fixing some pixel inconsistencies and gaps.

- Fixes an issue with long field names in settings popovers
- Fixes an issue with text overflow in settings popovers

## Feature branch env
[Feature Branch Link](http://fb-table-improvements-2.fb.qa.budibase.net)


deanhannigan (approved):
LGTM! An amazing update 🤯. The UX is really intuitive and robust, I really tried to break it but it simply refused.

PR #14028, opened 2024-06-26
Row Step Automation UX updates and Automation refactoring

deanhannigan:
## Description

Updates to Automation row steps to allow request composition! Previously, a row step would show every single field in the table schema. This took up a tonne of screen real estate and implied that empty fields would be cleared by the request.

- Empty fields in the previous UX were *completely ignored*. That is not the case now. If you actively add a field to your row step and leave it empty, this means **the field will be cleared** in an `Update Row` step
- Existing update automations **will behave in exactly the same way they did before**. No action will be required on your part and empty fields will still be ignored. Previously empty fields will now no longer be listed in the new request UI for `Update Row` and `Created Row` step types when loaded.
- Relationship/Link field config `clearRelationships` has been deprecated. Any fields marked with it as `true` will still be added to your automation request, whether its empty or not.
- UX updates have also been extended to the Test Modal for all Row trigger types.

## Addresses
- https://linear.app/budibase/issue/BUDI-8243/automations-update-row-ux

## Screenshots
Getting started. The new UX lets you pick a table but won't load every field. Clicking on `Add fields` will show you all available fields in the table schema. Add and remove as necessary.  
Screenshot 2024-06-26 at 14 12 40

Select fields relevant to your update.

The new update step layout.

### Updates

This work also includes some general refactoring and bug fixing. The updates should improve stability in Automations in general.
- `General changes`
  - Replaced the JSON row picker with the `CodeEditor`
  - Trigger table not locked in `Test Modal`
    - For all row trigger types, the user could select any table, even though it should fixed to match the trigger step.
    - This is now locked and will update and clear the automation `testData` when the trigger is changed.
  - Refactored `FieldSelector` as a lot of the coercion logic present was never used.
  - Removed `Password`, `Email` automation schema fields as they were no longer used.
  - Schema type `fields` now has labels in test modal.
  - 
  - Added a `PropField` component to handle Automation form layouts.
    - Defaults to `320` component width and the label takes the remaining space
    - Supports full width mode; `JSON`, `Long form`, `Attachment` and `Signature` field types.
  - Composable layouts - all row types now use the same composable structure to display the new row layout. This coupled with the `onChange` logic in the `AutomationBlockSetup` should pave the way for more complex automation layouts.
  - Removed several instances of mutation in form controls. We try not to use `bind:value` as it can have unintended consequences.
- `Bug Fixes` - found along the way
  - Removed `undefined` from the `url` field when adding a new attachment
  - Too many updates in FieldSelector.
    - On blur triggers a full update which is unnecessary
  - Mouseover for string type is truncated - present on field name tooltips, in this case.
    - General fix. Altered TooltipWrapper to use AbsTooltip 
  - FieldSelector Mutating the value. Required refactor.
  - "Use binding" Attachment UX doesn't work in test modal.
  - Long field names push the fields out of the automation step
  - Double header border in automations header ui
  - `Set Up Webhook` button visible in the test modal - Removed.
  - Required output fields not marked correctly in test modal. Now marking in step and test modal
  - z Position bug in TestModal Binding drawer.
  - Signature type says "Add attachment" and allows adding more than one.

## Launchcontrol

New Automation row step UX updates and some bug fixes

## Feature branch env
[Feature Branch Link](http://fb-feature-automation-row-ux-update.fb.qa.budibase.net)


aptkingston:
Going to review this properly of course, but just had a though while looking at the screenshots. Instead of the double X icons (which I know are needed here because the outermost one controls removing the field from the update), could we not just remove fields from the update but unselecting them in the multiselect? That would clean up the look of this even more, and feels natural since we allow deselection from multiselects in the other places we use them.

What do you reckon?

deanhannigan:
> Going to review this properly of course, but just had a thought while looking at the screenshots. Instead of the double X icons (which I know are needed here because the outermost one controls removing the field from the update), could we not just remove fields from the update by unselecting them in the multiselect? That would clean up the look of this even more, and feels natural since we allow deselection from multiselects in the other places we use them.

@aptkingston, yep agreed. I'm not a huge fan of them myself and it's trivial to remove the `X` icons 👍

aptkingston (commented):
Awesome updates @deanhannigan! Going through the code now, but from usage here are the things that I found:

Functional issues:
- If you add a field and enter a value for it, you can't remove the field again. You need to first delete the value before it will let you remove the field.
- If you have an existing automation (from before this PR) with a row created or row updated trigger, the test data modal will not have a table selected. That also means you can't add any fields to the test data, and since it's disabled it's not possible to get around this. (The fix is change to another table and change back, but obviously we don't want users to have to do this)
image
- In the test data modal, if you swap to JSON and edit the table ID to a non existing table, then swap back to values, then it will crash the builder with a `RowSelector.svelte:94 Uncaught (in promise) TypeError: schemaFields is not iterable` error. Putting my real QA hat on here to find this one 😂 

Small UI issues:
- We should remove the disabled look of fields that get selected in the field picker, since we can now deselect them. This made sense before deselection was possible.
- My preference would be to remove the additional borders around things like attachments, users and signatures, but I'm open to other opinions on that.
- I think it might be nice to sort the fields alphabetically to make it easier for people to find the field they want. This could also be rolled out into other parts of the platform, but I think it would definitely be a nice enhancement here.

packages/bbui/src/Tooltip/TooltipWrapper.svelte
```diff
@@ -1,33 +1,25 @@
...
+      <AbsTooltip text={tooltip}>
+        <div
+          class="icon"
+          class:icon-small={size === "M" || size === "S"}
+          on:focus
+        >
+          <Icon name="InfoOutline" size="S" {disabled} hoverable />
         </div>
-      {/if}
+      </AbsTooltip>
```
aptkingston, line 22: Nice update - fixes a bunch of issues across the board.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -8,31 +8,32 @@
...
   import { cloneDeep } from "lodash/fp"
+  import { memo } from "@budibase/frontend-core"
+  import { AutomationEventType } from "@budibase/types"
 
   let failedParse = null
   let trigger = {}
   let schemaProperties = {}
 
+  const memoTestData = memo($selectedAutomation.testData)
+  $: memoTestData.set($selectedAutomation.testData)
```
aptkingston, line 19: This is the way 🪄

packages/builder/src/components/automation/SetupPanel/FieldSelector.svelte
```diff
@@ -38,77 +44,58 @@
...
-      <RowSelectorTypes
-        {isTestModal}
-        {field}
-        {schema}
-        {bindings}
-        {value}
-        {onChange}
-      />
+      <PropField label={field}>
+        {#if ["string", "number", "array"].includes(schema.type)}
```
aptkingston, line 62: NAB but might be nice to use the `FieldType` enums here rather than magic strings for the data types.

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -35,153 +36,197 @@
...
                       ? {
                           url: e.detail[0].name,
                           filename: e.detail[0].value,
                         }
-                      : {}
+                      : null
                     : e.detail.map(({ name, value }) => ({
                         url: name,
                         filename: value,
                       })),
```
aptkingston, line 200: NAB this `onChange` callback is a little bit gnarly with the multiple ternary statements etc... Would be a lot more readable to break it out into a function and not use all the conditionals but not a huge deal.

packages/server/src/automations/steps/updateRow.ts
```diff
@@ -82,39 +82,73 @@ export async function run({ inputs, appId, emitter }: AutomationStepInput) {
...
-    ) {
-      delete inputs.row[propKey]
-    }
+  // Base update
+  let rowUpdate: Record<string, any>
+
+  // Legacy
+  // Find previously set values and add them to the update. Ensure empty relationships
+  // are added to the update if clearRelationships is true
+  const legacyUpdated = Object.keys(inputs.row || {}).reduce(
```
aptkingston, line 91: Just a question on this - when the new frontend next makes a change to this automation, will inputs.row be removed from the saved definition? Just noticing that `legacyUpdated` overrides `parsedUpdate`, so if it didn't get cleared then the saved values in `inputs.row` will forever override the new meta info.

I assume this does happen, but thought it was faster to ask than try to find the relevant code in the frontend haha.

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -1,61 +1,163 @@
...
-    dispatch("change", value)
+
+  $: parsedBindings = bindings.map(binding => {
+    let clone = Object.assign({}, binding)
+    clone.icon = "ShareAndroid"
+    return clone
+  })
+
+  $: tableId = $memoStore?.row?.tableId
+  $: if (tableId) {
```
aptkingston, line 65: This is a pretty huge reactive statment! Looks like it will react to changes in `tableId`, `memoStore`, `editableRow` and `editableFields` because they are used as values too.

I would definitely refactor this to be a function, with obvious props passed in that we want to react to. It would be really hard in future to work with this otherwise, and there are so many potential triggers. This also has the side effect of mutating `editableFields` many times, whereas we could do it in a single update if we used a function.

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -1,61 +1,163 @@
...
+    // Purge any row keys that are not present in the schema.
+    for (const rowKey of Object.keys(editableRow)) {
+      if (!schemaKeys.includes(rowKey)) {
+        delete editableRow[rowKey]
+        delete editableFields[rowKey]
+      }
+    }
   }
 
+  $: typeToField = Object.values(FIELDS).reduce((acc, field) => {
```
aptkingston, line 150: NAB but this can be a const - it's immutable since it's only a derivation of FIELDS.

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -86,130 +193,175 @@
...
-    let newField = {}
-    newField[field] = {
-      [key]: value,
-    }
-
-    let updatedFields = {
-      ...meta?.fields,
-      ...newField,
+  const onChange = update => {
+    const customizer = (objValue, srcValue) => {
```
aptkingston, line 205: Is this basically doing a recursive deep merge? Hard to tell from the code but I think that's what is going on...

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -86,130 +193,175 @@
...
-<style>
-  .field-width {
-    width: 320px;
-  }
+<Popover
+  align="center"
+  bind:this={customPopover}
+  anchor={popoverAnchor}
+  minWidth={popoverAnchor?.getBoundingClientRect()?.width}
+  maxWidth={popoverAnchor?.getBoundingClientRect()?.width}
```
aptkingston, line 309: The popover has a prop called `useAnchorWidth` which should do this for you 👍

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -43,31 +43,56 @@
...
   import { FIELDS } from "constants/backend"
+  import PropField from "./PropField.svelte"
 
   export let block
   export let testData
   export let schemaProperties
   export let isTestModal = false
 
+  // Stop unnecessary rendering
+  const memoBlock = memo(block)
```
aptkingston, line 68: 💪

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -117,15 +145,248 @@
       }
     }
   }
-  const onChange = Utils.sequential(async (e, key) => {
+
+  $: customStepLayouts($memoBlock, schemaProperties)
```
aptkingston, line 149: This is basically what I'm suggesting to use in `RowSelector` as well - a nice function with explicit params to react to rather than a monster reactive statement.

aptkingston (commented):
Absolute monster of a code refactor! I can tell that was definitely not easy or fun to work through all that logic, but the result is way better for the user 👌 

Only a couple of minor suggestions for refactors etc but can't see anything major 👍

packages/server/src/automations/steps/updateRow.ts
deanhannigan: Yep, the second the user alters the automation in the frontend, everything that shouldn't be present is purged. The legacy stuff in the server is just there to accomodate any unchanged automations. That's why it iterates over the `inputs.row` and not `inputs.meta.fields`.

deanhannigan:
> If you have an existing automation (from before this PR) with a row created or row updated trigger, the test data modal will not have a table selected.
I haven't been able to reproduce this one with older trigger types, but I did run into something else. I disabled this picker because it now reflects the correct trigger table at all times. However, you used to have to set it manually and you could pick any table at all. I think I need to add a check if the trigger/testData tableIds are matched and reset it if it's not.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -117,15 +145,248 @@
...
+              },
+              bindings,
+              updateOnChange: false,
+              forceModal: true,
+            },
+          },
+        ]
+      }
+
+      stepLayouts[block.stepId] = {
```
PClmnt, line 218: @deanhannigan Hi Dean, see for this, there was some work done recently to expose the `oldRow` within the frontend. This is an input for the Update Row trigger as well so the user needs to be able to add it within their Test Data. I think it was just generated out of the trigger schema before but doesn't seem to be there now, what do you think the best way to handle this is?

deanhannigan: Hey Peter,  good spot! The layout for the screen has its own configuration and its the only step type that does this at the minute. The UX just needs to be configured and mapped to the correct location in the testData. I'll have a look at this now 👍

packages/builder/src/components/automation/SetupPanel/FieldSelector.svelte
```diff
@@ -38,77 +47,58 @@
...
-
-  const onChange = (e, field, type) => {
-    value[field] = coerce(e.detail, type)
-    dispatch("change", value)
   }
 </script>
 
-{#if schemaFields.length && isTestModal}
-  <div class="schema-fields">
+{#if schemaFields?.length && isTestModal}
```
Ghrehh, line 61: Is this component ever actually used? Was having trouble triggering it to render, only place it's consumed is 
[here](https://github.com/Budibase/budibase/blob/feature/automation-row-ux-update/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte#L950) seemingly, but I couldn't get it to render with any of the automation step types I tried.

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -1,61 +1,168 @@
...
+        if (!Object.hasOwn(table?.schema, column)) {
+          delete editableFields[column]
+        }
       }
-    })
-  }
-  const onChangeTable = e => {
-    value["tableId"] = e.detail
-    dispatch("change", value)
+      editableFields = editableFields
```
Ghrehh, line 100: Very NAB, feel free to ignore: Is the assignment here necessary to trigger an update? 

If so, since we're cloning the `editableFields` anyway, would it be a little nicer to return a `newEditableFields` from this function and change the reactive statement that invokes it to `$: editableFields = initData(tableId, $memoStore?.meta?.fields, $memoStore?.row)`?

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -1,61 +1,168 @@
...
+          // remove this and place the field in the editable row.
+          delete editableFields[key]?.clearRelationships
+
+          // Default the field
+          editableRow = {
+            ...editableRow,
+            [key]: emptyField,
+          }
+        } else {
+          // Purge from the update as it's presence is not necessary.
```
Ghrehh, line 136: NAB: Wrong `it's` here I think.

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -35,153 +36,210 @@
...
+    }
 
     if (!Array.isArray(keyValueObj) && keyValueObj) {
       keyValueObj = [keyValueObj]
     }
 
     if (keyValueObj.length) {
       for (let param of keyValueObj) {
-        params[param.url] = param.filename
+        params[param.url || ""] = param.filename || ""
```
Ghrehh, line 55: Might be missing something, but will this not potentially just add a random empty string key to the params object? Does a guard or something make more sense? 

```js
for (let param of keyValueObj) {
     if (!params.url) {
       continue;
     }
     params[param.url] = param.filename || ""
}
```

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
```diff
@@ -195,20 +253,41 @@
...
-  .json-input-spacing {
-    margin-top: var(--spacing-s);
-    margin-bottom: var(--spacing-l);
+  .attachment-field-spacing {
+    border: 1px solid var(--spectrum-global-color-gray-400);
+    border-radius: 4px;
+    padding: var(--spacing-s);
+  }
+
+  .field-wrap.json-field {
```
Ghrehh, line 279: NAB: Is there a way to make the code-editor resizeable? At a glance I don't see any obvious props that make it possible, but perhaps you know more about it than I do.

The reason I asked, is that 80px is a little titchy IMO, it'd be hard to write or view any complex JSON.

Screenshot 2024-07-03 at 12 43 00

If resizing isn't possible, would bumping the size up make sense?

packages/builder/src/components/automation/SetupPanel/TableSelector.svelte
```diff
@@ -25,4 +26,5 @@
   options={filteredTables}
   getOptionLabel={table => table.name}
   getOptionValue={table => table._id}
+  {disabled}
```
Ghrehh, line 29: Smart change 🥚

Ghrehh (approved):
LGTM, major improvement. Some small comments left on code and some more general ones below that weren't really appropriate to leave anywhere in particular:

- The `Add fields` menu UX is a little odd when adding multiple fields, the popover stays in place as the fields appear underneath it, gradually pushing the button that opens it away, it will also jump once the user attempts to scroll. TBH I don't really know a great fix for this, it kind of seems like an inherent flaw of the design. 
Some possible options would be moving the `Add Fields` button above the fields, or not adding fields until the user closes the menu. Not sure!

- NAB: Is it possible for the JSON code editor to have syntax highlighting?

- Changing the name of a field in the data section then hides that field in the automations field view, not sure if this is a concern or not.

- These were patch notes you left in the original PR body I have questions about it:
`Schema type fields now has labels in test modal.`
What is a schema type? I tried everything I could think of but I couldn't figure it out.
`Required output fields not marked correctly in test modal. Now marking in step and test modal`
I checked this, and at least for string fields, this doesn't seem to be the case, they weren't marked as required.

packages/builder/src/components/automation/SetupPanel/FieldSelector.svelte
deanhannigan: This component is found in `App action` trigger types. It's only visible in the test modal. It renders the fields in the trigger and makes them editable.

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
deanhannigan: The main goal here was to just not get `undefined` in the url text box when adding an attachment for the first time. The empty string is just a starting point for the url. The main reason why I don't like the above approach is that if you delete the url, it completely removes the entry from the attachments list, even if you've entered your file name.

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
deanhannigan: At the minute no, it behaves in the same way as the old JSON editor. It allows you to quickly glance over your content and if they want more screen real-estate, they can click on the lightning bolt.

deanhannigan:
Hey @Ghrehh, thanks again for taking a look! My responses to your main comment are below:

> The Add fields menu UX is a little odd when adding multiple fields

Agreed. This seems to be a limitation of the `Popover` component where is doesn't follow the anchor element. We can definitely improve the general behaviour here.

> NAB: Is it possible for the JSON code editor to have syntax highlighting?

It is possible, but we would have to add the Codemirror JSON language to the editor. Definitely on the cards for future work as the CodeEditor component expands into other areas of Budibase

> Changing the name of a field in the data section then hides that field

Yep, this makes sense. It's no longer a valid input so it will be removed from the automation.

> What is a schema type? I tried everything I could think of but I couldn't figure it out.

I'm referring to automation schema types. `AutomationCustomIOType` in the automation schema is used to govern how the automation is rendered on screen. `fields` is not present as a real `AutomationCustomIOType` type as it is only used in the front end. This is how the `App action` fields are rendered in the test modal.

> Required output fields not marked correctly in test modal. Now marking in step and test modal`

In `rowUpdated.ts`, the required outputs for the update row trigger are `["row", "id"]`. In the test modal, `id` is the only one we render as a field, so only it gets marked with required. Also, they're not strictly required here as this is just the test data. I think i'll remove this for now as it seems to be confusing. 


In `updateRow.ts`, the required *inputs* are `["row", "rowId"]`. Like the test modal we only render the `rowId` field as a text field

packages/builder/src/components/automation/SetupPanel/RowSelectorTypes.svelte
deanhannigan: I bumped the default height by another 40 pixels and be we should definitely look into making it resizable.

packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
deanhannigan: I think this line might be a hangover from an earlier iteration. I don't believe this is necessary anymore. 

Also, as `editableFields` is directly altered elsewhere in the same component, I usually prefer to use a standard variable declaration in those scenarios. I find directly altering a reactive variable quite odd.

PR #14038, opened 2024-06-27
Initial implementation of solving SQS issue with case insensitivity

mike12345567:
## Description
Fix for issues with columns named after other internal columns, like `type` or `createdAt` - these are not supposed to be usable but the check was case sensitive, with SQS this is a problem.

Quite a few instances of columns named things like `Type` exist, which we have to support, we can't migrate these easily without breaking bindings. This PR introduces a change where all user columns are prefixed as `data_...` in the SQLite table, meaning they can't overlap with anything.

We already did case sensitivity checks vs other columns in the table, but just didn't handle the columns which are internal to the table.


packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -90,6 +92,33 @@ function cleanupFilters(
...
+      const found = userColumnList.find(possibleColumn =>
+        key.endsWith(possibleColumn)
+      )
+      if (found) {
+        const newKey = key.replace(found, userColumnMap[found])
+        filter[newKey] = filter[key]
+        delete filter[key]
+      }
+    }
+  }
```
samwho, line 120: Could we not simplify this section quite a lot?

```typescript
function keyInAnyTable(key: string): boolean {
  return allTables.some(table => table.schema[key])
}

for (const filter of Object.values(filters)) {
  for (const key of Object.keys(filter)) {
    const [prefix, rawKey] = splitKeyAndNumericPrefix(key) // don't think this exists yet we'd have to write it
    if (keyInAnyTable(rawKey)) {
      filter[`${prefix}${mapToUserColumn(rawKey)}`] = filter[key]
      delete filter[key]
    }
  }
}
```

We also remove a bunch of array scans and any possibility of doing an incorrect replacement.

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -106,6 +135,26 @@ function buildTableMap(tables: Table[]) {
...
+        const newKey =
+          key.slice(0, lastIndex) + key.slice(lastIndex + prefixLength)
+        finalRow[newKey] = row[key]
+      } else {
+        finalRow[key] = row[key]
+      }
+    }
+    return finalRow
+  })
+}
```
samwho, line 156: Are we not running the risk here of the user having columns that contain the prefix? e.g. does this not fail for the column name `data_foo`? It gets prefixed to `data_data_foo` and then we find the last occurrence of `data_`, which is not the correct one.

Would be worth writing some tests regardless that create column names that contain the prefix in various ways to make sure we do this wrapping/unwrapping correctly.

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -147,6 +196,8 @@ async function runSqlQuery(
   const response = await alias.queryWithAliasing(json, processSQLQuery)
   if (opts?.countTotalRows) {
     return processRowCountResponse(response)
+  } else if (Array.isArray(response)) {
+    return reverseUserColumnMapping(response)
```
samwho, line 200: Do we need to wrap the `return response` in the else block below in this function?

packages/server/src/sdk/app/rows/search/sqs.ts
mike12345567: I've written a new function which requires metadata from the tables to work this out instead.

samwho (approved):

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2108,4 +2108,26 @@ describe.each([
...
+    it("should be able to query a column with data_ in it", async () => {
+      await expectSearch({
+        query: {
+          equal: {
+            ["1:name_data_test"]: "a",
+          },
+        },
+      }).toContainExactly([{ name_data_test: "a" }])
+    })
+  })
```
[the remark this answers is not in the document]
mike12345567: Done - parameterised tests for this!

PR #14039, opened 2024-06-27
Remove all Redis mocking in `packages/server`

samwho:
## Description

Part of my crusade to run as much real code in tests as possible, removing `ioredis-mock` has been on my list for a while. This ended up being a wide change but there are only a few key changes:

1. `ioredis-mock` is entirely gone, as are a bunch of custom Redis mocks we had.
2. I've simplified how we connect to Redis. We had a custom retry mechanism that didn't seem necessary given that `ioredis` has its own, so I got rid of it. It only applied during initial connection anyway, it doesn't handle reconnecting after the initial attempts.
3. I've removed all global instances of `TestConfiguration` because data sharing between them complicated tracking down bugs in the process of making this change. The only truly shared global state in tests should be the Koa server.
4. I've converted a whole bunch of `.js` tests to `.ts` to make tracking down the `TestConfiguration` sharing easier.


packages/backend-core/src/redis/redis.ts
samwho, line 1: This is where the meat of removing `ioredis-mock` and simplifying Redis connections is. Because tests needed to be able to `await` Redis startup, this file has been considerably reworked to allow that.

packages/backend-core/src/redis/utils.ts
```diff
@@ -89,6 +89,7 @@ export function getRedisConnectionDetails() {
 export function getRedisOptions() {
   const { host, password, port } = getRedisConnectionDetails()
   let redisOpts: Redis.RedisOptions = {
+    enableReadyCheck: true,
```
samwho, line 92: This enables the `ready` state in `ioredis` and makes sure we aren't using a Redis connection before it's ready.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -225,21 +224,28 @@ export default class TestConfiguration {
 
   // use a new id as the name to avoid name collisions
   async init(appName = newid()) {
-    if (!this.started) {
+    if (this.initialised) {
```
samwho, line 227: We had to start doing a lot more in here because we needed to be able to `await` server startup. This is because we need to wait until Redis is ready, which is done during server startup. When we were using `ioredis-mock` this wasn't necessary because there was nothing really to wait for.

packages/server/src/threads/automation.ts
```diff
@@ -39,9 +43,6 @@ import env from "../environment"
 import tracer from "dd-trace"
 
 threadUtils.threadSetup()
-const FILTER_STEP_ID = actions.BUILTIN_ACTION_DEFINITIONS.FILTER.stepId
```
samwho, line 42: These references to `BUILTIN_ACTION_DEFINITIONS` ended up causing an import cycle at one point so I removed them from the top level of the file.

packages/server/src/utilities/redis.ts
```diff
@@ -25,9 +24,7 @@ export async function init() {
 
   // Duplicate the socket client for pub/sub
   socketClient = await redis.clients.getSocketClient()
-  if (!env.isTest()) {
-    socketSubClient = socketClient.getClient().duplicate()
-  }
```
samwho, line 30: This was only gated because we used `ioredis-mock`. Now that's gone we can use Redis proper.

packages/server/src/api/routes/public/tests/users.spec.ts
```diff
@@ -1,26 +1,66 @@
...
+    return user()
+  }),
+  saveGlobalUser: jest.fn(() => {
+    return { _id: "user", _rev: "rev" }
+  }),
+  deleteGlobalUser: jest.fn(() => {
+    return { message: "deleted user" }
+  }),
+  removeAppFromUserRoles: jest.fn(),
+}))
```
samwho, line 44: These used to live in a file almost everything imported, but they're only used here. Moved here to reduce confusion.

packages/server/src/api/routes/tests/analytics.spec.ts
samwho, line 1: In all cases where I'm doing a `.js` to `.ts` conversion I've changed only the minimum to pass type checks. I think there are a couple of exceptions where minor logical changes were needed, but I didn't end up having to rework any of these in great detail.

packages/server/src/app.ts
```diff
@@ -14,21 +14,27 @@ import Koa from "koa"
...
+  await started
 }
 
 start().catch(err => {
   console.error(`Failed server startup - ${err.message}`)
   throw err
 })
 
-export function getServer(): Server {
+export async function getServer(): Promise<Server> {
```
samwho, line 37: This is a key change. In order for tests to wait for everything to properly initialise, this had to be made async so it can be `await`ed.

samwho:
Putting this to the side. It has consumed too much of my time and the end isn't clearly in sight.

PR #14059, opened 2024-07-01
Fix contains search on multi-user column.

samwho:
## Description

We had a bug report this morning that search was broken for contains on multi-user columns and it turns out the frontend sends a request to the backend that doesn't match the type for search requests on the backend. So this PR makes accommodations for that but I'm not very happy with it because it seems that this is now an outlier and I wonder if the frontend is sending searches that don't match the type in more cases.

### DIFF
--- a/packages/types/src/sdk/search.ts
+++ b/packages/types/src/sdk/search.ts
@@ -54,7 +54,7 @@ export interface SearchFilters {
     [key: string]: any[]
   }
   [SearchFilterOperator.CONTAINS]?: {
-    [key: string]: any[] | any
+    [key: string]: any[]
   }
   [SearchFilterOperator.NOT_CONTAINS]?: {
     [key: string]: any[]
--- a/packages/types/src/sdk/search.ts
+++ b/packages/types/src/sdk/search.ts
@@ -54,7 +54,7 @@ export interface SearchFilters {
     [key: string]: any[]
   }
   [SearchFilterOperator.CONTAINS]?: {
-    [key: string]: any[]
+    [key: string]: any[] | any
   }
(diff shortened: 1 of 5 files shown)


packages/types/src/sdk/search.ts
```diff
@@ -54,7 +54,7 @@ export interface SearchFilters {
     [key: string]: any[]
   }
   [SearchFilterOperator.CONTAINS]?: {
-    [key: string]: any[]
+    [key: string]: any[] | any
```
samwho, line 57: Everything else in this type takes an array of any, and now this is the only one that takes an array of any OR any, and that makes me feel a bit uneasy.

adrinr: Is this because what the frontend sends? Could we not tweak the frontend instead?

adrinr: Is this because what the frontend sends? Could we not tweak the frontend instead?

mike12345567: > Is this because what the frontend sends? Could we not tweak the frontend instead?

Sadly if the frontend has been sending something which isn't an array, we can't really change that, as these searches will be baked into apps, it isn't super easy to update them all (since client libraries aren't updated with the server version right now).

I think rather than updating our typing around this, could we instead update the API controller to look over the search operators and re-jig any which won't fit into our expected structure? That way we could keep the typing clean and just have the fact that the frontend doesn't consistently send good search parameters hidden in the API layer?

samwho: Done! Much nicer. Thanks folks 🙂

mike12345567 (approved):
LGTM!

PR #14063, opened 2024-07-01
Updating filter fields to use relationship name rather than related table name

mike12345567:
## Description
Using the relationship name rather than the table name - the relationship may have a different name to the related table, this makes it a bit more sensible.

### DIFF
--- a/packages/server/src/sdk/app/rows/search/sqs.ts
+++ b/packages/server/src/sdk/app/rows/search/sqs.ts
@@ -340,8 +340,7 @@ export async function search(
       await sdk.tables.sqs.syncDefinition()
       return search(options, table)
     }
-    // previously the internal table didn't error when
-    console.log(JSON.stringify(err))
+    // previously the internal table didn't error when a column didn't exist in search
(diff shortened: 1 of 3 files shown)


packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -331,12 +334,17 @@ export async function search(
...
     const syncAndRepeat =
-      (err.status === 400 && msg?.match(NO_SUCH_COLUMN_REGEX)) ||
+      (err.status === 400 && msg?.match(USER_COLUMN_PREFIX_REGEX)) ||
       (err.status === 404 && msg?.includes(SQLITE_DESIGN_DOC_ID))
     if (syncAndRepeat) {
       await sdk.tables.sqs.syncDefinition()
       return search(options, table)
     }
+    // previously the internal table didn't error when
+    console.log(JSON.stringify(err))
```
samwho, line 344: 👀

mike12345567: woops - thank you!

samwho (approved):

PR #14069, opened 2024-07-02
Automation grouping

deanhannigan:
## Description

Automations are now grouped by trigger type in the side navigation. The icons for each trigger type are shown along side each automation.

## Addresses
- https://linear.app/budibase/issue/BUDI-8364/group-automations-by-triggers-add-icons-for-side-nav

## Launchcontrol

Automations now grouped by type in the side navigation.


packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
```diff
@@ -30,6 +30,16 @@
       return lowerA > lowerB ? 1 : -1
     })
 
+  $: groupedAutomations = filteredAutomations.reduce((acc, auto) => {
+    acc[auto.definition.trigger.event] = acc[auto.definition.trigger.event] || {
```
Ghrehh, line 34: NAB: [Nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) maybe a bit better here?

```
 acc[auto.definition.trigger.event] ??= {
    icon: auto.definition.trigger.icon,
    name: (auto.definition.trigger?.name || "").toUpperCase(),
    entries: [],
  }
```

packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
```diff
@@ -82,6 +101,17 @@
...
+  .nav-group {
+    padding-top: var(--spacing-l);
+  }
+  .nav-group-header {
+    color: var(--spectrum-global-color-gray-600);
+    padding: 0px calc(var(--spacing-l) + 4px);
+    padding-bottom: var(--spacing-l);
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
Ghrehh, line 113: Great future-proofing here, considering I don't think any of the names even can overflow at the mo 🫡

Ghrehh (approved):
LGTM

PR #14072, opened 2024-07-02
Handle view visibility schema

adrinr:
## Description
A few weeks back @Ghrehh found an inconsistency with the view schemas (https://github.com/Budibase/budibase/pull/13625). Analysing the impact on this we realised we had some issues on the way we stored some view information, as some endpoints were trimming the non-visible columns completely while others were not.

After a discussion, we decided to never trim this information for different reasons. The main one is that the view schema might have other metadata rather than the visible/non-visible config, such as width, order, etc.
When retrieving a row from a view, we will trim out the non-visible fields, but we will never trim any info from the view itself.

Related:
- https://github.com/Budibase/budibase/pull/13625

## Addresses
- [BUDI-8413 - Views - trim invisible properties from rows](https://linear.app/budibase/issue/BUDI-8413/views-trim-invisible-properties-from-rows)

## Launchcontrol
Fixing some incoherences around view schemas


mike12345567:
Fixes: https://linear.app/budibase/issue/BUDI-8413/views-trim-invisible-properties-from-rows

mike12345567 (approved):

PR #14074, opened 2024-07-02
Enable SQS by default in the Helm chart.

samwho:
## Description

Enables SQS by default for our Helm chart. I'll release a version 3.1.0 by hand after this goes in, which will be after we've released everywhere else.

The large change to the README.md is sadly due to `helm-docs` changing how they output the markdown table.


mike12345567 (approved):

samwho:
Superseded by https://github.com/Budibase/budibase/pull/13999

PR #14083, opened 2024-07-03
Fix an issue where null values were causing an error in automation loops

PClmnt:
## Description
Null or undefined values were causing an issue in the loop binding replacement code and throwing an error. This fix simply does nothing for those values, (since they're empty and wont contain and handlebars statements that need updated). 

Also added some tests to cover this issue and potential other problem areas such as empty arrays / objects. 
## Addresses
https://linear.app/budibase/issue/BUDI-8410/looping-in-automations-cannot-convert-undefined-or-null-to-object

## Launchcontrol
- Fixes an issue where looping an object or array with null values was causing undefined errors


packages/server/src/automations/loopUtils.ts
```diff
@@ -18,14 +18,18 @@ function replaceBindingsRecursive(
...
+
   if (typeof value === "object") {
     for (const [innerKey, innerValue] of Object.entries(value)) {
       if (typeof innerValue === "string") {
         value[innerKey] = automationUtils.substituteLoopStep(
           innerValue,
           `steps.${loopStepNumber}`
         )
-      } else if (typeof innerValue === "object") {
+      } else if (typeof innerValue === "object" && innerValue !== null) {
```
samwho, line 32: I know this isn't your code but `typeof innerValue === "object"` almost certainly doesn't do what the author is expecting it to. 

```
❯ node
Welcome to Node.js v20.10.0.
Type ".help" for more information.
> typeof {} === "object"
true
> typeof [] === "object"
true
> typeof null === "object"
true
```

samwho (approved):
Beautiful tests.

packages/server/src/automations/loopUtils.ts
PClmnt: I think it is mine from a good while ago - but yes you're right. We would be better checking for the existence of keys in the object here I think

PClmnt: UPD2: I added some more checks in that if, in order to be a bit more clear on what we're doing.

packages/server/src/automations/loopUtils.ts
```diff
@@ -29,7 +29,12 @@ function replaceBindingsRecursive(
           innerValue,
           `steps.${loopStepNumber}`
         )
-      } else if (typeof innerValue === "object" && innerValue !== null) {
+      } else if (
+        innerValue &&
+        typeof innerValue === "object" &&
+        !Array.isArray(innerValue) &&
+        Object.keys(innerValue).length > 0
```
samwho, line 36: If you're wanting to detect that something is an object, `_.isPlainObject` is a bit clearer.

mike12345567: this can be re-written to
```
import { isPlainObject } from "lodash"
....
innerValue && isPlainObject(innerValue) && Object.keys(innerValue).length > 0
```

PClmnt: Done, thanks!

Issue #14079, opened 2024-07-02
Unable to create documents in CouchDB datasource


g-serge:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.29.6
  - Couchdb images versions tested : budibase/couchdb & ibmcom/couchdb3 (latest version as of 2024-07-02)

**Describe the bug**
It's not possible to create a document using CouchDB datasource with a Create query.

**To Reproduce**
1. Go to the data section of an app and add a CouchDB datasource
2. Create a new query to your CouchDB datasource
3. Select "Create" in Function dropdown
4. Add a valid JSON document inside Query text field
5. Click on "Run query" button
6. Observe error "Query Error: this.client[command] is not a function"

**Expected behavior**
The query is run and a document is created.

**Screenshots**
![image](https://github.com/Budibase/budibase/assets/145403199/d27506d5-eaf5-41fb-bdad-001f9ee64f3f)

**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox 127.0.2 (64 bits) / Chrome

**Additional context**
Clicking on "Run query" button triggers a request to "/api/queries/preview" wich fails with the following response : 
```
{"message":"this.client[command] is not a function","status":400}
``` 
The datasource is the same CouchDB server used by Budibase.

mike12345567:
This issue will be fixed in the next release (2.29.13)

PR #14086, opened 2024-07-03
Typescript improvements for CouchDB integration

mike12345567:
## Description
Fix for CouchDB integration, the typing in it was very poor and wise hiding issues, I've updated how this is implemented so that static typing can catch issues with it.

This also addresses: https://github.com/Budibase/budibase/issues/14079


packages/server/src/integrations/couchdb.ts
```diff
@@ -79,21 +82,19 @@ class CouchDBIntegration implements IntegrationBase {
...
     }
     return response
   }
 
-  async query(
-    command: string,
-    errorMsg: string,
-    query: { json?: object; id?: string }
-  ) {
+  async query<T>(operation: () => Promise<T>, errorMsg: string) {
```
samwho, line 95: This function isn't really doing what it advertises. It runs a given function and logs the error before rethrowing. It could be better called `logError(...` or gotten rid of entirely, I don't personally look at the error logs when those same errors are more usefully visible in traces.

samwho (approved):

packages/server/src/integrations/couchdb.ts
mike12345567: Fair point - simplified!

PR #14088, opened 2024-07-03
Allow updating email via SCIM

adrinr:
## Description
Allow updating emails for users synced via SCIM. Some clients' AD integrations are having issues with some synchronisation caused by email changes.

Related PR:
- https://github.com/Budibase/budibase-pro/pull/304

## Addresses
- [BUDI-8416 - Allow updating email via SCIM
](https://linear.app/budibase/issue/BUDI-8416/allow-updating-email-via-scim)

## Launchcontrol
Allow updating emails for users synced via SCIM.


packages/backend-core/src/users/test/db.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+              _rev: expect.stringMatching(/^2-\w+/),
+            })
+          )
+        })
+      })
+
+      it("updating emails frees previous emails", async () => {
+        await config.doInTenant(async () => {
+          const previousEmail = user.email
+          const newEmail = generator.email({})
```
mike12345567, line 160: why the empty object to the generator?

mike12345567 (approved):
LGTM

packages/backend-core/src/users/test/db.spec.ts
adrinr: We added a custom lint rule that "requires" a parameter. An empty object works, for this

samwho: 😭

PR #14089, opened 2024-07-03
Separate SQS migration into another environment variable

mike12345567:
## Description
As title - giving the option to run migrations while not actually enabling the search functionality.


adrinr (approved):
We will need to remember to cleanup all these flags

samwho (approved):

PR #14090, opened 2024-07-03
Detect secrets in error messages.

samwho:
## Description

This pull request seeks to prevent us ever leaking secrets out through uncaught thrown exceptions and it does it by maintaining a list of environment variables that we consider to be secret values and then for each error message we check that the error does not contain any of those secret values.

There are two downsides to this approach that I can think of:

1. It relies on us adding new secrets to the list of secret environment variables which would be easy to forget if you don't know about this mechanism.
2. If you're using a sufficiently generic secret like "password" or something like that you could end up filtering out error messages that aren't actually problematic


packages/backend-core/src/environment.ts
```diff
@@ -202,6 +202,21 @@ const environment = {
   DISABLE_SCIM_CALLS: process.env.DISABLE_SCIM_CALLS,
 }
 
+type EnvironmentKey = keyof typeof environment
+export const SECRETS: EnvironmentKey[] = [
+  "API_ENCRYPTION_KEY",
```
mike12345567, line 207: There are some environment variables which are specific to services, like `OPENAI_API_KEY`, `BB_ADMIN_USER_PASSWORD`, `SENDGRID_API_KEY`, `AWS_SESSION_TOKEN` which are from the `packages/server/src/environment.ts` and `packages/worker/src/environment.ts` - these may not be in the backend-core environment list, although all of the ones I've listed are common.

One thing I've thought about doing is getting rid of the worker/server environment implementations and keeping one full list in backend-core, which helps a bit with managing the secret list as well.

packages/backend-core/src/security/secrets.ts
```diff
@@ -0,0 +1,20 @@
+import environment, { SECRETS } from "../environment"
+
+export function stringContainsSecret(str: string) {
+  if (str.includes("-----BEGIN PRIVATE KEY-----")) {
```
mike12345567, line 4: Smart thinking!

mike12345567 (approved):

PR #14097, opened 2024-07-04
Squash the oldRow prop in the external row controller

deanhannigan:
## Description
I ran into an issue when testing update triggers for an external database with relationships. When the update event was triggered and the server attempted to queue the automation, it got stuck into a cyclical loop trying to serialise the `oldRow` prop.

## Addresses
- Squashes the `oldRow` event property so that it can be serialised when queuing an automation.
- Adds a try/catch for development to ensure issues encountered when queuing automations won't kill the local server.

## Launchcontrol

Minor fix to ensure oldRow is squashed for external sources.


mike12345567 (approved):
LGTM!

PR #14098, opened 2024-07-04
Ellipsis on long emails

adrinr:
## Description
Fixing some UI breaks with really long emails, specially visible for SCIM users

| Before | After |
|--------|--------|
| image | image | 

## Launchcontrol
Fix really long email UI breaks

### DIFF
--- a/packages/builder/src/pages/builder/portal/users/users/_components/EmailTableRenderer.svelte
+++ b/packages/builder/src/pages/builder/portal/users/users/_components/EmailTableRenderer.svelte
@@ -5,7 +5,17 @@
   export let row
 </script>
 
-{value}
+<span class="email">
+  {value}
+</span>
 {#if row.scimInfo?.isSync}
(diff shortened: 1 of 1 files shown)


packages/builder/src/pages/builder/portal/users/users/_components/EmailTableRenderer.svelte
```diff
@@ -5,7 +5,17 @@
   export let row
 </script>
 
-{value}
+<span class="email">
```
deanhannigan, line 8: Could you add a `title={value}` attribute so that the user can see the full email on hover?

deanhannigan (approved):
LGTM! One minor suggestion but not a blocker.

packages/builder/src/pages/builder/portal/users/users/_components/EmailTableRenderer.svelte
adrinr: Done

PR #14109, opened 2024-07-05
Fix for builder relationship picker 

deanhannigan:
## Description

Minor change to account for `datasource` not being fully initialised as the FilterBuilder loads

## Addresses
- https://linear.app/budibase/issue/BUDI-8411/relationship-picker-filtering-is-broken


Ghrehh (approved):
LGTM 👍

PR #14116, opened 2024-07-08
Per-tenant enablement of SQS

samwho:
## Description

This PR makes it possible to enable SQS on a per-tenant basis. Actually, it _requires_ enabling SQS on a per-tenant basis for the time being, I've made it such that if no tenants are specified then SQS isn't enabled for anyone (felt safer to fail-closed than fail-open).

To enable SQS you need to specify `SQS_SEARCH_ENABLE=true` (which kinda felt redundant but it seems to be used in some non-tenant-specific places so I kept it in), and set `SQS_SEARCH_ENABLE_TENANTS=tenant_id_1,tenant_id_2,etc`.


packages/server/src/api/controllers/table/utils.ts
```diff
@@ -15,7 +15,7 @@ import { getViews, saveView } from "../view/utils"
 import viewTemplate from "../view/viewBuilder"
 import { cloneDeep } from "lodash/fp"
 import { quotas } from "@budibase/pro"
-import { events, context } from "@budibase/backend-core"
+import { events, context, db } from "@budibase/backend-core"
```
mike12345567, line 18: Need to rename this - but for safety can we import this as `db as dbCore` - try to reserve the `db` keyword to avoid overlap!

mike12345567 (approved):
LGTM!

packages/server/src/api/controllers/table/utils.ts
samwho: Totally understand this. Is it worth, in a different PR, changing this export to be called dbCore so you never have to rename it like this?

mike12345567: We probably could do that - we'd just need to go through everywhere its imported and update it to remove the as!

PR #14120, opened 2024-07-08
CSV export fails for table with composite primary key

adrinr:
## Description
Support composite primary keys on filters, allowing actions such as CSV exporting specific rows.

## Addresses
- [BUDI-7065 - CSV export fails for table with composite primary key](https://linear.app/budibase/issue/BUDI-7065/csv-export-fails-for-table-with-composite-primary-key)

## Screenshoots
_Table with name+age as primary key_

https://github.com/Budibase/budibase/assets/15987277/e8879d5f-4912-4c5b-950d-eba64f323d71

image

## Launchcontrol
Allow exporting CSV exports for tables with composite primary key


packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -1452,6 +1436,46 @@ describe.each([
         { status: 404 }
       )
     })
+
+    // MSSQL needs a setting called IDENTITY_INSERT to be set to ON to allow writing
+    // to identity columns. This is not something Budibase does currently.
+    providerType !== DatabaseName.SQL_SERVER &&
+      it("should handle filtering by composite primary keys", async () => {
+        const tableRequest = saveTableRequest({
+          primary: ["number", "string"],
```
samwho, line 1445: The last time I was looking into this, for external data sources, only the first element in this array gets used. Are you sure that this test is actually creating a composite primary key?

adrinr: Not 100% sure if in the database it gets created as identity, but for the logic of these export tests we shoudn't really care. We want to treat both columns as the primary key, and we want to get sure that the select queries work as expected (ignoring if they are or are not primary keys on the external dbs).

Even so, we could double-check it, but I think it gets actually set on the DB. For this reason, I needed to disable SQL Server, as we were getting actual DB exceptions

adrinr: Just to clarify, without this PR code changes, this test would fail

samwho (approved):

packages/backend-core/src/sql/sql.ts
```diff
@@ -170,6 +170,8 @@ function convertBooleans(query: SqlQuery | SqlQuery[]): SqlQuery | SqlQuery[] {
   return query
 }
 
+const COMPLEX_ID_OPERATOR = "_complexIdOperator"
```
mike12345567, line 173: Could we move this perhaps into type (I know its not strictly a type, but we put enums and other stuff in there as well, we probably would want to use it for typing in the future as well

adrinr: You mean in @budibase/types? We could also add strict types if they live in backend core, as this is where the only "usage" lives

adrinr: Amended!

mike12345567 (approved):
Very nice typing!

PR #14124, opened 2024-07-09
Fix notification message from display the wrong phrasing

ConorWebb96:
## Description
Enable and disabled were displaying the wrong way round. Amended so it outputs the correct message at the right time.

## Launchcontrol

Minor fix to automation notification message.


PClmnt (approved):
LGTM!

PR #14125, opened 2024-07-09
Fix Client Modal Component `Ignore clicks outside` Setting

Ghrehh:
## Description
https://linear.app/budibase/issue/BUDI-8425/modal-component-closes-on-click-outside-ignoring-the-settings

The `ignoreClickOutside` prop for modal components in the client was not wired up correctly, so the `Ignore clicks outside` checkbox wasn't doing anything, this should be fixed and working as intended now.


ConorWebb96 (approved):
LGTM!

PR #14128, opened 2024-07-09
Fix bug in oneOf search.

samwho:
## Description

This PR fixes a problem we were seeing with is-in queries where the values that arrive from the front end as comma separated strings were not being correctly split into individual strings.

In fixing this bug I've realized a wider problem in the way that search works, which is that the type we have in the back end for the layout of a search query doesn't match what the front end sends us. And there's quite a lot of fudging going on in different places to make it match, but this is done in a kind of piecemeal way. What we should probably do is have two types, one type for what the front end is sending us, and then one type, the existing type, for what the back end can operate on, and have an explicit conversion in one place between these two types.

I spoke to Michael Drury about this and he's going to create a sustaining ticket because I suspect it's going to be a couple of days of work to clean all this up.

### DIFF
--- a/packages/shared-core/src/filters.ts
+++ b/packages/shared-core/src/filters.ts
@@ -327,6 +327,35 @@ export const buildQuery = (filter: SearchFilter[]) => {
   return query
 }
 
+// The frontend can send single values for array fields sometimes, so to handle
+// this we convert them to arrays at the controller level so that nothing below
+// this has to worry about the non-array values.
+export function fixupFilterArrays(filters: SearchFilters) {
+  const arrayFields = [
+    SearchFilterOperator.ONE_OF,
+    SearchFilterOperator.CONTAINS,
+    SearchFilterOperator.NOT_CONTAINS,
+    SearchFilterOperator.CONTAINS_ANY,
+  ]
+  for (const searchField of arrayFields) {
+    const field = filters[searchField]
+    if (field == null) {
+      continue
+    }
+
+    for (const key of Object.keys(field)) {
+      if (!Array.isArray(field[key])) {
+        if (typeof field[key] !== "string") {
+          field[key] = [field[key]]
+        } else {
+          field[key] = field[key].split(",").map(x => x.trim())
+        }
+      }
+    }
+  }
+  return filters
+}
+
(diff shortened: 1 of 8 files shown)


packages/shared-core/src/filters.ts
```diff
@@ -323,6 +323,35 @@ export const buildQuery = (filter: SearchFilter[]) => {
   return query
 }
 
+// The frontend can send single values for array fields sometimes, so to handle
+// this we convert them to arrays at the controller level so that nothing below
+// this has to worry about the non-array values.
+export function fixupFilterArrays(filters: SearchFilters) {
+  const arrayFields = [
```
adrinr, line 330: This could come from a type/array next to the SearchFilterOperator. So when we add a new operator, we don't need to look for all the possible usages, being centralised in a single place

packages/shared-core/src/filters.ts
```diff
@@ -323,6 +323,35 @@ export const buildQuery = (filter: SearchFilter[]) => {
...
+    SearchFilterOperator.ONE_OF,
+    SearchFilterOperator.CONTAINS,
+    SearchFilterOperator.NOT_CONTAINS,
+    SearchFilterOperator.CONTAINS_ANY,
+  ]
+  for (const searchField of arrayFields) {
+    const field = filters[searchField]
+    if (field == null) {
```
adrinr, line 338: We should handle other invalid cases. Couldn't we even receive a non-object that would blow the forof below?

adrinr (commented):
Nice catch, and nice to have better testing around all these use cases!

packages/shared-core/src/filters.ts
samwho: Yeah, I was thinking about this. I really dislike having this list here, because as you mentioned, it's so detached from the type that it would be easy to miss upon adding a new field. Do you have some preferred way of signaling it in the type? Or would you recommend just putting this definition near the type?

samwho: Tried a lil' something I think you might like. Let me know!

mike12345567 (approved):
Very nice - good idea to split up the operators!

PR #14131, opened 2024-07-09
Fixing issue with multiple relations to same table

mike12345567:
## Description
There was an issue with multiple relationships to the same table and SQS - as SQS uses a single junction table (no matter how many relationships) between the two tables. This is a bit more efficient than creating junction tables for every relationship and we have the information in the junction to decipher the relationships between their different columns (although we weren't doing this).

This created a problem when you had multiple relationships to the same table where all rows that were related would end up in every column, if you removed from one column it wouldn't understand what to do, as it might not be present in that column.

I've added a test case for internal tables for this, it doesn't impact external tables and is a relatively pointless test as relationships depend on foreign keys and specific configuration for external.


packages/server/src/api/controllers/row/utils/sqlUtils.ts
```diff
@@ -22,6 +22,18 @@ export function isManyToMany(
   return !!(field as ManyToManyRelationshipFieldMetadata).through
 }
 
+function isCorrectRelationship(
+  relationship: RelationshipsJson,
+  row: Row
+): boolean {
+  const junctionTableId = relationship.through!
```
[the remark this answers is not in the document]
mike12345567: Updated this to pass in the two individual tables - this is currently true for SQS (as all relationships are many to many in the background) but might not be true forever.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2085,6 +2085,58 @@ describe.each([
...
+        await expectSearch({
+          query: {},
+        }).toContainExactly([
+          {
+            name: "test",
+            related1: [{ _id: relatedRows[0]._id }],
+            related2: [{ _id: relatedRows[1]._id }],
+          },
+        ])
+      })
```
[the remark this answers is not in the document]
mike12345567: Added!

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2133,8 +2142,47 @@ describe.each([
...
           },
+          {
+            name: "test2",
+            related1: [{ _id: relatedRows[2]._id }],
+            related2: [{ _id: relatedRows[3]._id }],
+          },
         ])
       })
+
+      isSqs &&
```
samwho, line 2153: Are these tests not applicable to all datasources?

samwho (approved):

packages/server/src/api/routes/tests/search.spec.ts
mike12345567: Lucene can't do this and I've limited these test cases to internal because this doesn't have a lot of impact on external DBs, this is a problem which is specific to the internal implementation of relationships (and it's odd use of junction tables for efficiency)

samwho: I'd argue that while this problem affects SQS today, there's no reason it won't affect external datasources later. I'd strongly lean toward enabling this test in as many places as we can, to ensure parity is being maintained and future regressions don't affect places we weren't expecting.

Issue #14136, opened 2024-07-11
When `return all table rows` is selected the Is in filter returns nothing instead of all rows


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.29.16
  - App Version: 2.29.16

**Describe the bug**
Data isn't returned for the `is in` filter whenever the value is empty, when the `return all table rows` is selected.

**To Reproduce**
1. Use the below app export
2. Go to the home screen
3. See currently no results are returning even though the filter value has been left blank
4. Switch to equals and save it will return results

**Expected behavior**
All records should be returned if no value is present.
  
**App Export**
[is in test-export-1720682926923.tar.gz](https://github.com/user-attachments/files/16173014/is.in.test-export-1720682926923.tar.gz)

adrinr:
Fixed in release [2.29.18](https://github.com/Budibase/budibase/releases/tag/2.29.18)

PR #14140, opened 2024-07-11
When `return all table rows` is selected the is in filter returns nothing instead of all rows

adrinr:
## Description
Remove `is in` filter when there is no values in the filtered info

## Addresses
- [BUDI-8449 - When return all table rows is selected the Is in filter returns nothing instead of all rows
](https://linear.app/budibase/issue/BUDI-8449/when-return-all-table-rows-is-selected-the-is-in-filter-returns)
- https://github.com/Budibase/budibase/issues/14136

### Filtering with a valid filter
image

### Filtering with an empty value filter (and return all fallback)
image

### Filtering with an empty value filter (and return none fallback)
image

### Filtering with an non-matching value filter
image

## Launchcontrol
Honour `onEmptyFilter` selection when `is in` filter is set without proper values

### DIFF
--- a/packages/shared-core/src/filters.ts
+++ b/packages/shared-core/src/filters.ts
@@ -106,7 +106,7 @@ export const NoEmptyFilterStrings = [
   OperatorOptions.NotEquals.value,
   OperatorOptions.Contains.value,
   OperatorOptions.NotContains.value,
-  OperatorOptions.NotContains.value,
+  OperatorOptions.ContainsAny.value,
   OperatorOptions.In.value,
 ] as (keyof SearchQueryFields)[]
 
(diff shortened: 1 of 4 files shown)


packages/shared-core/src/filters.ts
```diff
@@ -106,31 +106,49 @@ export const NoEmptyFilterStrings = [
   OperatorOptions.NotEquals.value,
   OperatorOptions.Contains.value,
   OperatorOptions.NotContains.value,
+  OperatorOptions.NotContains.value,
```
adrinr, line 109: Not sure if this object should still be called `NoEmptyFilterStrings`

samwho (approved):
Love this, much cleaner.

PR #14144, opened 2024-07-11
Re-sync SQS definitions consistently

mike12345567:
## Description
Resync if it is found that a table or column is missing, this was previously done for specific cases but have expanded to cover all, but only retry once (not get into an infinite loop).

### DIFF
--- a/packages/server/src/sdk/app/rows/search/sqs.ts
+++ b/packages/server/src/sdk/app/rows/search/sqs.ts
@@ -249,7 +249,7 @@ function resyncDefinitionsRequired(status: number, message: string) {
 export async function search(
   options: RowSearchParams,
   table: Table,
-  opts?: { retry?: boolean }
+  opts?: { retries?: number }
 ): Promise<SearchResponse<Row>> {
   let { paginate, query, ...params } = options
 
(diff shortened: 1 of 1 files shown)


packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -376,9 +377,9 @@ export async function search(
     return response
   } catch (err: any) {
     const msg = typeof err === "string" ? err : err.message
-    if (resyncDefinitionsRequired(err.status, msg)) {
+    if (!opts?.retry && resyncDefinitionsRequired(err.status, msg)) {
       await sdk.tables.sqs.syncDefinition()
-      return search(options, table)
+      return search(options, table, { retry: true })
```
samwho, line 382: It's that thing again where from the perspective of the caller Passing in retry true makes me think that I'm asking the function to retry But I think here you're using it as a marker to let it know that it is a retry and it shouldn't retry again I'm sure we ran into this somewhere else and we decided to use a different name, but I don't know what it was

samwho (approved):

samwho (approved):

packages/server/src/sdk/app/rows/search/sqs.ts
```diff
@@ -240,18 +237,19 @@ async function runSqlQuery(
...
     // no design document found, needs a full sync
     (status === 404 && message?.includes(SQLITE_DESIGN_DOC_ID))
   )
 }
 
 export async function search(
   options: RowSearchParams,
-  table: Table
+  table: Table,
+  opts?: { retries?: number }
```
adrinr, line 252: This seems confusing to me, as I would expect this to be the remaining attempts, not the current ones. For this case, either having a private function or just a boolean "retrying" would work

PR #14145, opened 2024-07-11
DynamicFilter fixes

deanhannigan:
## Description
Some general Builder bug fixes.

## Addresses
- https://linear.app/budibase/issue/BUDI-8367/fix-design-section-component-settings-icon-tooltip
  - Fix component panel header in Design section. Updated tooltip for the component icon hover.
- Fix for dynamic filter button size. The button size config was not passed to the component.
- Fix to remove invalid FieldTypes in `FilterBuilder`. Attachment single and Signature removed.
- Fix to remove invalid field types from the `DynamicFilter` component.
- Added in `_id` as a filterable column to `DynamicFilter`
- Removed `_rev` as a filter option in general.

## Launchcontrol

General bug fixes and DynamicFilter can now filter by _id


packages/client/src/components/app/dynamic-filter/DynamicFilter.svelte
```diff
@@ -69,7 +69,10 @@
         }
       })
     }
+
     return Object.values(clonedSchema || {})
+      .filter(field => !Constants.BannedSearchTypes.includes(field.type))
+      .concat([{ name: "_id", type: "string" }])
```
aptkingston, line 75: `_id` only makes sense for DS+, so we might need to try and be smart here and check if it's a DS+ or not, since currently this is adding it for any data source.

aptkingston (commented):
Just a question about how we handle _id but otherwise looks good!

packages/client/src/components/app/dynamic-filter/DynamicFilter.svelte
deanhannigan: That's totally correct. Nice spot 🎉

aptkingston (approved):

PR #14150, opened 2024-07-12
Fix a very rare flake in row.spec.ts.

samwho:
## Description

This one is fun because some attempt was obviously made to make these objects unique but the problem is, with external data sources, we only take the first column in a primary key so it was possible to create an object that had the same number column but different string columns making this an exceptionally rare flake.


adrinr (approved):
Not sure what `unique` does 100%, but the issue might be that we were returning objects instead of primitive types, so technically `{number:1}` and `{number:1}` are different

PR #14152, opened 2024-07-12
Row action CRUD endpoints

adrinr:
## Description
Scaffolding for the new row action CRUD endpoints. At the moment no logic is added to these actions, only the name field is stored.
These controllers are slightly different from our existing ones, and I would like to use this PR to discuss various pros and cons.

The main difference between this and other existing endpoints are:
1. Endpoint requests and responses are not aware of any repository implementation. We don't expose couch db fields.
2. If we need some specific data (such `_rev`), we can map them in/out. This will allow us exposing it in the shape we actually need, not being attached to the `_rev` format (for example, in this case).
3. The create/update endpoints will never return data outside the requested element. For example, creating views will return all the full table with the updated views. This makes all our endpoints dependent on the db shape.
4. On the endpoints, we don't use the full `ctx.request.body`. We extract what we need, making its usage much cleaner.
5. On the endpoints, controllers have only validation and mapping logic, the business and repository logic is in the SDKs.

Having the approach of dettaching API from db objects will allow us mutating data on the backend much easier, as we can do the mappings that we need. Our tests will also ignore couchdb fields and other values that are not actually useful for the specific feature.

## Addresses
- [BUDI-8428 - Row action API (storage and retrieval)
](https://linear.app/budibase/issue/BUDI-8428/row-action-api-storage-and-retrieval)


packages/server/src/api/controllers/rowAction/crud.ts
```diff
@@ -0,0 +1,80 @@
...
+async function getTable(ctx: Ctx) {
+  const { tableId } = ctx.params
+  const table = await sdk.tables.getTable(tableId)
+  if (!table) {
+    ctx.throw(404)
+  }
+  return table
+}
+
+export async function find(ctx: Ctx<void, RowActionsResponse>) {
```
samwho, line 19: Should these not use `UserCtx` instead of `Ctx`?

packages/server/src/api/routes/tests/rowAction.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+    rowAction: CreateRowActionRequest,
+    expectations?: Expectations,
+    opts?: { publicUser?: boolean }
+  ) {
+    return await config.api.rowAction.save(
+      tableId,
+      rowAction,
+      {
+        ...expectations,
+        status: expectations?.status || 201,
```
samwho, line 37: I'd probably put this default in `config.api.rowAction.save` instead of having this helper for it. I've done similar in other endpoints that have a non-200 success status.

packages/server/src/api/routes/tests/rowAction.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+      },
+      opts
+    )
+  }
+
+  function createRowActionRequest(): CreateRowActionRequest {
+    return {
+      name: generator.word(),
```
samwho, line 45: Do names have to be unique? If yes, let's not use `generator.word()`.

samwho (commented):
In principal I like the approach, though the problem we will inevitably have is that this will stand out as a separate approach to the rest of our code until the rest of our code does the same as this (which is likely to be never).

The age-old problem of wanting to improve but being held back by a legacy you will likely never have the time to change.

packages/server/src/api/controllers/rowAction/crud.ts
adrinr: I'm not sure if it makes any difference in this case. `UserCtx` will have the user typed, but nothing else, and we don't use it here at all

packages/server/src/api/routes/tests/rowAction.spec.ts
adrinr: Good point. No more flakiness

adrinr:
> In principal I like the approach, though the problem we will inevitably have is that this will stand out as a separate approach to the rest of our code until the rest of our code does the same as this (which is likely to be never).
> 
> The age-old problem of wanting to improve but being held back by a legacy you will likely never have the time to change.

As discussed we will go for it for this feature, as it's pretty independent from the existing flows. If we decide that we want to keep consistency in the future, we can always change it and return the full objects without the mappings

packages/backend-core/src/middleware/joi-validator.ts
```diff
@@ -6,7 +6,7 @@ function validate(
   property: string,
   opts?: { errorPrefix?: string; allowUnknown?: boolean }
 ) {
-  const errorPrefix = opts?.errorPrefix || `Invalid ${property}`
+  const errorPrefix = opts?.errorPrefix ?? `Invalid ${property}`
```
samwho, line 9: What's the difference between `||` and `??` here?

samwho (approved):

packages/backend-core/src/middleware/joi-validator.ts
adrinr: `||` is a [logical or](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR), so it "converts" the values to boolean. `??` is a [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing), checks only for nulls or undefineds.

Main difference:
'' || 'default' -> 'default'
'' ?? 'default' -> ''

PR #14157, opened 2024-07-12
Default table - in-memory/on-disk representation duplicates

mike12345567:
## Description
There is a risk with default tables that the schema may exist in the DB as well as existing in memory - we should prefer the CouchDB representation as this may have been adjusted by the user.

This problem is specific to sample data tables - it doesn't exist outside of this as all tables should only have a single table document/representation of them.


adrinr (approved):
Do we have some tests that could catch problems with this?

mike12345567:
> Do we have some tests that could catch problems with this?

Sadly we don't really have any test suites around default tables - they are very much their own thing and theres nothing testing those flows today - I'll create a sustaining ticket to expand this, until SQS I basically forgot this feature existed which is pretty dangerous!

PR #14169, opened 2024-07-15
Allow Opening Certain Context Menus With Right Click

Ghrehh:
## Description
Allows the user to open certain context menus by right-clicking anywhere on the component the context menu is for.

## Addresses
- https://linear.app/budibase/issue/BUDI-5426/right-click-should-activate-meatball-dropdown-menu

## Changes

- Adds a top level store and component to handle the displaying on context menus, this guarantees there'll only be one displayed at a time, even across different resources.
- Adds a unified method to display the menu items, allowing for icons, names, hotkey icons, callbacks and checks for disabling and fully hiding buttons. 
- Removed a file relating to automations that I don't think was being used.


deanhannigan (changes requested):
Love it! Enriches the UX all over the builder and portal! Also really tidies up the code of a lot of boilerplate context menu stuff. Really nicely done 🚀 

Just some bugs I ran into:
- `BUG` component copy/paste not working in the builder
- `BUG` - export published on the portal app list not working
- `BUG` - edit option shouldn't be present in the DataNavItem for external data source tables
  - Editing a table name for an external table will just create a new entry that doesn't work.
- `BUG` - Error when duplicating a newly created query.

packages/builder/src/components/ContextMenu.svelte
```diff
@@ -0,0 +1,68 @@
...
+  import { Popover, Menu, MenuItem } from "@budibase/bbui"
+
+  let dropdown
+  let anchor
+
+  const showDropdown = dropdown => {
+    dropdown?.show()
+  }
+
+  $: showDropdown(dropdown)
```
aptkingston, line 12: You can simplify showing the popover a little - there's a prop called `open` on the Popover component, so you can actually remove this `showDropdown` function, as well as the second `{#if ...}` statement around the popover, and instead just pass in `open={$contextMenuStore.visible}`.

packages/builder/src/components/ContextMenu.svelte
```diff
@@ -0,0 +1,68 @@
...
+/>
+
+{#key $contextMenuStore.position}
+  {#if $contextMenuStore.visible}
+    <Popover
+      bind:this={dropdown}
+      {anchor}
+      resizable={false}
+      align="left"
+      on:close={contextMenuStore.close}
```
aptkingston, line 41: Maybe just personal preference, but I think this is much nicer if we remove the animation with `animate={false}`. It's just much snappier, and feels more typical for a context menu.

aptkingston (changes requested):
Awesome work @Ghrehh - this feels really nice. Hit the same issues as @deanhannigan but he's already caught them, so I've just left a few comments for suggested changes which I think make this even better.

packages/builder/src/components/backend/DatasourceNavigator/QueryNavItem.svelte
```diff
@@ -0,0 +1,101 @@
...
+      },
+      {
+        icon: "Duplicate",
+        name: "Duplicate",
+        keyBind: null,
+        visible: true,
+        disabled: false,
+        callback: async () => {
+          try {
+            const newQuery = await queries.duplicate(query)
```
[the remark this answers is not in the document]
Ghrehh: Good spot, some `goto` weirdness, sorted now.

packages/builder/src/components/ContextMenu.svelte
Ghrehh: This was originally done because the popover wouldn't "notice" a change to it's provided anchor element, so the menu would show once then never again. I found a cleaner way to do it however, by doing as you suggested and putting the key statement around the anchor element itself.

packages/builder/src/components/ContextMenu.svelte
Ghrehh: Don't disagree, though I think no animation at all looks a little weird too. I changed it be a bit more like the MacOS right-click menu animation, where it's a subtle fade in, bit longer on the out.

You can let me know if you think that looks better, if not I'll just remove the animation entirely.

Ghrehh:
@aptkingston @deanhannigan Thanks for the reviews, responded to your feedback there. Some additional notes below

> `BUG` component copy/paste not working in the builder
I can't replicate this @deanhannigan, copy and pasting components seems to work fine for me, both into screens and the component tree. Could you provide a more detailed repro please? Maybe I'm missing something obvious.

## Other changes

- Added the ability to right-click an empty area of the component tree and paste any copied components (I found myself trying to do this and thought that's a good indication its intuitive). Let me know if you think this is a good idea or not.
- It was possible to access the context menu for the internal Budibase datasource which shouldn't be allowed, this is now fixed.

aptkingston:
I think a subtle fade would be fine if you want some animation, but I really don't like the Y animation - could you remove that? You can try a fade like you said, if you want to. The grid uses no animations at all for context menus and I still think thats the nicest, but happy to try a short fade too.

Ghrehh:
@aptkingston Sorry, wasn't clear, I've already done that. There's no y-movement, just a fade.

aptkingston:
Ah sorry must just be an issue on my end. Telling me I'm up to date but still seeing the Y transition. No probs though, sounds good.

aptkingston:
Seeing it now 👍 personally I still prefer no animation at all (which also keeps things consistent with the grid) but will wait for Dean / others to weigh in with their opinions too.

deanhannigan:
> I can't replicate this @deanhannigan, copy and pasting components seems to work fine for me, both into screens and the > component tree. Could you provide a more detailed repro please? Maybe I'm missing something obvious.

Yep, I had pinged you on discord about the issue while I was reviewing the PR. The bug list above contains everything I found  in the review

deanhannigan:
> Seeing it now 👍 personally I still prefer no animation at all (which also keeps things consistent with the grid) but will wait for Dean / others to weigh in with their opinions too.

@aptkingston Yep agree,  I prefer zero animation and it reflects the grid UX

Ghrehh:
@aptkingston @deanhannigan No problemo, will remove it. If you guys could give this another looks I'd appreciate it 🙏 

## Changes

- Removed the animation on the context menu.
- Fixed the issue with pasting into a component Dean was mentioning.
- Also fixed an issue where a context menu could stay open on page transitions. Since the context menu sat at the same level as the Routify router, it wouldn't get rerendered if the url changed. Moving it inside the router fixes this.

deanhannigan:
> Also fixed an issue where a context menu could stay open on page transitions. Since the context menu sat at the same level as the Routify router, it wouldn't get rerendered if the url changed. Moving it inside the router fixes this.

Nice one. 🤯

aptkingston (approved):
LGTM!

deanhannigan (approved):
LGTM!

Issue #14159, opened 2024-07-12
On Screen Load - Open Side Panel issue - Side Panel will open and immediately close


mikesealey:
**Checklist**
- `[Y]` I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self & Cloud

**Describe the bug**
Navigating to a screen that has a sidepanel, and an `On Screen Load` action to open the side panel will open and immediately close.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create two screens
2. Give one a side panel
3. Set `On Screen Load` to open the side panel
4. navigate from elsewhere in the app to the screen with the side panel
5. side panel will open and immediately close

**Expected behaviour**
Side Panel should open `On Screen Load` and stay open.

**Screenshots**
Jam - https://jam.dev/c/9e0836cd-6e70-4a73-92ab-2770adc1782a
  
**App Export**
[ Side Panel 12 Jul-export-1720791585599.tar.gz](https://github.com/user-attachments/files/16194836/NT.Aus.Ben.Martinson.Side.Panel.12.Jul-export-1720791585599.tar.gz)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser chrome
 - Version 126.0.6478.127

**Additional context**
Issue raised by customer, arrived in support inbox

benaunt:
Just confirming the latest version fixed this for us.

PR #14182, opened 2024-07-16
Add logic to prevent panel and modals from closing when screen load actions are available

ConorWebb96:
## Description
Prevent modal /sidepanel from closing when on-screen load actions are setup to open them

## Addresses
- https://github.com/Budibase/budibase/issues/14159 

## Launchcontrol
Prevents modals/side-panels from immediately closing whenever on screen load actions are opening them.


shogunpurple (approved):
LGTM!

PR #14185, opened 2024-07-16
Conditional table cell and row styles

aptkingston:
## Description
This PR adds the ability to conditionally style rows or cells in tables, based on their value.

Conditions are configured on a per-column basis, just like their name or width:
image

Conditions are edited inside a drawer, which feels familiar to the one we use for our existing component conditions feature:
image

When defining conditions you're able to update either the background colour or the text colour of the row or the cell. These are configured for each condition:
image
image

## New colours
To support this feature, a new range of subtle colours have been added to the colour picker. These colours work very well as background colours, when you usually don't want a strong colour:
image

## Examples
Let's say we have a table with employee performance numbers, and we want to highlight any employees with a poor performance rating (less than 50 in this case). We can add a condition that will update the performance rating cell background color to red:
image

This updates are table to look like this:
image

Let's say we'd also like to highlight employees that are performing well (>90 rating). We can add another condition to update their text colour to green:
image
image

Let's also update the text colour, for even more contrast. We can duplicate our conditions, and change them to update the text colour as well:
image
image

Maybe we'd rather update the entire row, to really stand out. To do this we just need to update our conditions to update the row background color rather than the cell:
image
image

## Complex example
Sometimes we might want to highlight a row based on the value of more than one field. Using our example above, maybe we only want to highlight underperforming employees who are also at a senior level. We can achieve this by creating a new formula column which will look at both the rating and their level, and return whether or not we should highlight them.

Firstly we create our formula column, and get it to just return true or false:
image
image

Now we can just add a condition on this column, to update the row background colour if the value is true:
image
image

We also don't actually want to see the values of this column in our table, so we can turn off the column in the table settings. The conditions will continue to work even with the column disabled:
image

## Feature branch env
[Feature Branch Link](http://fb-conditional-table-colours.fb.qa.budibase.net)


deanhannigan (commented):
This is amazing work @aptkingston. I gave the feature a thorough run through only found a small number of bugs, all of which we covered. The main issues were:

- `Fixed` - show all behaviour when there is no rule match
- `Fixed` - Signature, Attachment, Attachments. Not supported but present in the UI
- `Fixed` - No lookups for User Fields or Single/Multi selects

deanhannigan (approved):
LGTM! All issues resolved and some nice new updates. This is a phenomenal update @aptkingston! 🎉

PR #14186, opened 2024-07-16
[BUDI-8436] Make sure a field cannot both have a default value and be required.

samwho:
## Description

This PR handles the check for making sure that columns cannot be both required and have a default value. Also tests for the transitions between having a default value and not having one, when the field is not required.

### DIFF
--- a/packages/server/src/api/controllers/table/index.ts
+++ b/packages/server/src/api/controllers/table/index.ts
@@ -29,6 +29,7 @@ import sdk from "../../../sdk"
 import { jsonFromCsvString } from "../../../utilities/csv"
 import { builderSocket } from "../../../websockets"
 import { cloneDeep, isEqual } from "lodash"
+import { helpers } from "@budibase/shared-core"
 
 function pickApi({ tableId, table }: { tableId?: string; table?: Table }) {
   if (table && isExternalTable(table)) {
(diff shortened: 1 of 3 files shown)


packages/server/src/api/controllers/table/index.ts
```diff
@@ -40,6 +40,20 @@ function pickApi({ tableId, table }: { tableId?: string; table?: Table }) {
   return internal
 }
 
+function checkDefaultFields(table: Table) {
+  for (const [key, field] of Object.entries(table.schema)) {
+    if (!("default" in field) || field.default == null) {
+      continue
+    }
+    if (field.constraints?.presence) {
```
adrinr, line 48: We should use `packages/shared-core/src/helpers/schema.ts, isRequired`. This might be a complex object, and this function encapsulates the logic

adrinr (changes requested):

packages/server/src/api/controllers/table/index.ts
samwho: Just got a little bitten by this because the type that `isRequired` accepts is `FieldConstraints | undefined`. Because all of the fields in `FieldConstraints` are optional, any object can be successfully passed in. So I initially tried `isRequired(field)` instead of `isRequired(field.constraints)` and nothing shouted at me, but all the tests failed.

I'm not sure what the answer is, but it's a bit of a sharp edge as it currently stands.

adrinr (approved):

PR #14193, opened 2024-07-18
Support space as last char in column name in SQS

samwho:
## Description

This PR is in response to [this error](https://app.datadoghq.eu/apm/trace/669584d4000000003a7d9a55792a5878?colorBy=service&env=prod&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=1539363709037386526&spanViewType=metadata&timeHint=1721074900479.001&view=spans) we've seen in production. There are tables that have column names that end in a space, and this is not handled correctly when generating the SQLite query in the search endpoint. This PR fixes that. I've left comments explaining the fix throughout the PR, as it's quite subtle and has probably shaved a few months off my life expectancy.

[This comment](https://github.com/Budibase/budibase/pull/14193/files#r1682629994) is the key to the fix.


packages/backend-core/src/sql/sql.ts
```diff
@@ -113,34 +114,45 @@ function generateSelectStatement(
...
+    //     in the case of relationships, where it's table.doc1.column. In that
+    //     case, we want to split it into `table`.`doc1.column` for reasons that
+    //     aren't actually clear to me, but `table`.`doc1` breaks things with the
+    //     sample data tests.
+    return knex.raw(
+      `${quote(client, table)}.${quote(
+        client,
+        [column, ...rest].join(".")
+      )} as ${quote(client, field)}`
+    )
```
samwho, line 155: This is the actual fix. The previous `${field} as ${field}` gets passed to `knex.select(...)` which, when `as` is present, parses the string.

So you get the following:

```typescript
knex.select("foo as bar")  // `foo` as `bar`
knex.select("foo  as bar") // `foo` as `bar`
knex.select(" foo as bar") // `foo` as `bar`
```

This change forces the spaces to preserved.

packages/backend-core/src/sql/sql.ts
```diff
@@ -42,27 +42,28 @@ const envLimit = environment.SQL_MAX_ROWS
...
-      throw new Error("Unknown client generating like key")
+      return `\`${str}\``
   }
-  const parts = key.split(".")
-  key = parts.map(part => `${start}${part}${end}`).join(".")
+}
+
+// Takes a string like a.b.c and returns a quoted identifier like [a].[b].[c]
+// for SQL Server and `a`.`b`.`c` for MySQL.
+function quotedIdentifier(client: SqlClient, key: string): string {
```
samwho, line 62: This is just `likeKey` from before renamed to be more general (because I'm using it in a context that has nothing to do with `like` queries). It is logically the same as `likeKey` was, except split into another `quote` function so I can use that logic elsewhere.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2182,8 +2188,7 @@ describe.each([
         }).toContainExactly([{ name: "baz", productCat: undefined }])
       })
     })
-
-  isInternal &&
+  ;(isSqs || isLucene) &&
```
samwho, line 2182: Semi-colon is annoyingly required or Node parses this as an attempted function call.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2371,6 +2376,7 @@ describe.each([
       beforeAll(async () => {
         await config.api.application.addSampleData(config.appId!)
         table = DEFAULT_EMPLOYEE_TABLE_SCHEMA
+        rows = await config.api.row.fetch(table._id!)
```
samwho, line 2370: This is the one change that was required to make this test work with the in-memory search tests.

packages/backend-core/src/sql/sql.ts
```diff
@@ -42,27 +42,28 @@ const envLimit = environment.SQL_MAX_ROWS
...
-      break
+      return `"${str}"`
     case SqlClient.MS_SQL:
-      start = "["
-      end = "]"
-      break
+      return `[${str}]`
     default:
-      throw new Error("Unknown client generating like key")
+      return `\`${str}\``
```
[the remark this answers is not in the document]
adrinr: Should this be an exhaustive switch, if this is never meant to be?

packages/server/src/api/routes/tests/search.spec.ts
adrinr: 💩 yes, I really dislike this

packages/backend-core/src/sql/sql.ts
adrinr: This function name is not really verbose. Should this not be `parseIdentifier` or something similar? And is this quoted, if it returns square brackets or others?

samwho: I'm not sure what a better word would be than "quoted". I don't really have strong feelings about the name other than `likeKey` was incorrect after the way I used it in this PR. Happy to go with `parseIdentifier` if you'd like, though I'm not sure that's more clear.

packages/backend-core/src/sql/sql.ts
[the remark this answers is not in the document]
samwho: Good call, done!

packages/backend-core/src/sql/sql.ts
```diff
@@ -113,34 +114,81 @@ function generateSelectStatement(
...
+    if (table) {
+      return knex.raw(
+        `${quote(client, table)}.${quote(client, column)} as ${quote(
+          client,
+          field
+        )}`
+      )
+    } else {
+      return knex.raw(`${quote(client, field)} as ${quote(client, field)}`)
+    }
```
samwho, line 191: This is the actual fix. The previous `"${field} as ${field}"` gets passed to `knex.select(...)` which, when `as` is present, parses the string.

So you get the following:

```typescript
knex.select("foo as bar")  // `foo` as `bar`
knex.select("foo  as bar") // `foo` as `bar`
knex.select(" foo as bar") // `foo` as `bar`
```

This change forces the spaces to preserved.

packages/backend-core/src/sql/sql.ts
```diff
@@ -113,34 +114,81 @@ function generateSelectStatement(
...
+    // A table name and a column name, e.g.: "table.column"
+    if (parts.length === 2) {
+      table = parts[0]
+      column = parts[1]
+    }
+
+    // A link doc, e.g.: "table.doc1.fieldName"
+    if (parts.length > 2) {
+      table = parts[0]
+      column = parts.slice(1).join(".")
```
samwho, line 143: This is honestly something I don't fully understand, but when querying against link tables you get gives fields like `a.doc1.fieldName`, which splits to `["a", "doc1", "fieldName"]`. Previously, with `"${field} as ${field}"` as the return value, the `"fieldName"` segment is preserved despite not being used below, as `columnName = fieldNames[1]` would result in `columnName = "doc1"`, which I don't actually think appears in any table's schema.

However, _not_ preserving this `"fieldName"` segment results in weird errors in other tests. Preserving it fixes them. I did not dig deeply into this.

packages/backend-core/src/sql/sql.ts
adrinr: I was just a little bit confused, as technically it does not always quotes the id

adrinr (approved):

PR #14194, opened 2024-07-18
Improved handling around invalid relationship column

deanhannigan:
## Description
If you attempt to paste invalid data into a relationship field in the grid, it will kill the builder. 

- This change will handle the UI issue more gracefully.
- The backend will now identify invalid request data, rejecting the row update, then returning an error.

## Screenshots
Screenshot 2024-07-18 at 09 55 52

## Launchcontrol
Improve error handling when pasting data into relationship columns


packages/server/src/db/linkedRows/LinkController.ts
```diff
@@ -172,6 +172,11 @@ class LinkController {
       const rowField = row[fieldName]
       const field = table.schema[fieldName]
       if (field.type === FieldType.LINK && rowField != null) {
+        // Expects an array of docs with at least their _id
+        if (!Array.isArray(rowField)) {
+          throw new Error("Relationship Error: Invalid request")
```
samwho, line 177: Is this an error we expect users to be able to hit, or is this more something you hit during development if you send a request to the API that isn't well-formed?

If it's something the user could reasonably encounter, I think we should try to provide a better error message. Something that includes guidance on how to fix the situation they're in.

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -29,17 +29,35 @@
...
+    // Is it unset or a valid array? 1+
+    const isValid = val =>
+      !val ||
+      (Array.isArray(val) &&
+        (val.length === 0 ||
+          val.some(relEntry => Object.hasOwn(relEntry, "_id"))))
+
+    const stf = JSON.stringify(value)
+    if (!cacheStr || cacheStr !== stf) {
+      cacheStr = stf
```
adrinr, line 55: What is the logic behind this cache? Just not to call the isValid?

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -29,17 +29,35 @@
...
+      return isValid(value) ? [...(value || [])] : []
+    }
+    return [...(value || [])]
+  }
+
```
aptkingston, line 60: Following up on what Adri said, could this be a little simpler?

```js
const parseValue = value => {
  if (Array.isArray(value) && value.every(x => x?._id)) {
    return value
  }
  return []
}
```

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
deanhannigan: I ran into some issues with reactivity at one point and was caching to help debug. As @aptkingston, this can be made way simpler.

packages/server/src/db/linkedRows/LinkController.ts
deanhannigan: Users will get this error when pasting invalid content into a relationship cell, like an attachment or a number. This is just a catch-all error for scenarios like that. I used the error on line 230 as a rough template. 

I updated the message to `Relationship Error: Invalid value` to better imply that the pasted data was the issue and not some general issue with the request.

aptkingston (approved):

PR #14195, opened 2024-07-18
Fix Automations Breaking When Their Datasource is Deleted

Ghrehh:
## Bug Description

Deleting a table an automation uses bricks the builder.

## Repro

- Create a table `data`
- Create an automation that triggers whenever a row is created for `data`
- Add a step to the automation to update a row, select `data` as the table for this also, Row ID doesn't need to be set.
- Delete the table `data`.
- View your automation.
- The app should cease function and print some errors in the console.

## Fix

Seems to be caused by trying to iterate over schema fields while it's undefined. Moving it outside of a guard clause fixes the issue.


deanhannigan (approved):
LGTM! Nice catch!

PR #14198, opened 2024-07-18
Remove unnecessary save button for automation steps

adrinr:
## Description
Remove the unnecessary save button while adding a new step into an automation. The current modal has no other input than selecting the step type, and if added by mistake it can be easily removed. Because of this, removing this extra click gives a much better UX

## Addresses
- [BUDI-8452 - Row action frontend (data UI)](https://linear.app/budibase/issue/BUDI-8452/row-action-frontend-data-ui)

### Before

https://github.com/user-attachments/assets/148fa7df-4eb2-4b58-9dd2-d922fcbcd782

### After

https://github.com/user-attachments/assets/b8e3cd88-81f5-4618-8d94-f6d83ea9596d

## Launchcontrol
Remove the save button while saving new automation steps.

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
@@ -81,7 +81,7 @@
   // Check the schema to see if required fields have been entered
   $: isError =
     !isTriggerValid(trigger) ||
-    !trigger.schema.outputs.required.every(
+    !trigger.schema.outputs.required?.every(
       required => $memoTestData?.[required] || required !== "row"
     )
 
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -81,7 +81,7 @@
   // Check the schema to see if required fields have been entered
   $: isError =
     !isTriggerValid(trigger) ||
-    !trigger.schema.outputs.required.every(
+    !trigger.schema.outputs.required?.every(
```
adrinr, line 84: This is not mandatory on the document, so it might cause undefined exceptions

shogunpurple (approved):
LGTM! I was so used to this I never thought about it that much, but it makes total sense

PClmnt (approved):
Nice one, LGTM

PR #14200, opened 2024-07-19
Create automation on adding new row action

adrinr:
## Description
Adding a new automation with `row action` type when adding a new row action. This automation will be deleted when the original row action is deleted.
No UI changes at all exist on this PR. There are some TODO left, that for the time being are not being used at all. I will be analysing its usage, as I think they some of them are obsolete fields that can be removed.

## Addresses
- [BUDI-8430 - Row action API - auto-create automation](https://linear.app/budibase/issue/BUDI-8430/row-action-api-auto-create-automation)

### DIFF
--- a/packages/server/src/api/controllers/rowAction/crud.ts
+++ b/packages/server/src/api/controllers/rowAction/crud.ts
@@ -31,7 +31,12 @@ export async function find(ctx: Ctx<void, RowActionsResponse>) {
     actions: Object.entries(actions).reduce<Record<string, RowActionResponse>>(
       (acc, [key, action]) => ({
         ...acc,
-        [key]: { id: key, tableId: table._id!, ...action },
+        [key]: {
+          id: key,
+          tableId: table._id!,
+          name: action.name,
+          automationId: action.automationId,
+        },
       }),
       {}
     ),
@@ -50,7 +55,9 @@ export async function create(
 
   ctx.body = {
     tableId: table._id!,
-    ...createdAction,
+    id: createdAction.id,
+    name: createdAction.name,
+    automationId: createdAction.automationId,
   }
   ctx.status = 201
 }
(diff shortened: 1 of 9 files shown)


packages/server/src/api/controllers/rowAction/crud.ts
```diff
@@ -31,7 +31,12 @@ export async function find(ctx: Ctx<void, RowActionsResponse>) {
     actions: Object.entries(actions).reduce<Record<string, RowActionResponse>>(
       (acc, [key, action]) => ({
         ...acc,
-        [key]: { id: key, tableId: table._id!, ...action },
```
adrinr, line 34: Removing magic mappings that can be exposing unexpected fields

packages/server/src/api/routes/tests/rowAction.spec.ts
```diff
@@ -129,9 +132,24 @@ describe("/rowsActions", () => {
...
       expect(await config.api.rowAction.find(tableId)).toEqual({
         actions: {
-          [responses[0].id]: { ...rowActions[0], id: responses[0].id, tableId },
-          [responses[1].id]: { ...rowActions[1], id: responses[1].id, tableId },
-          [responses[2].id]: { ...rowActions[2], id: responses[2].id, tableId },
+          [responses[0].id]: {
+            name: rowActions[0].name,
+            id: responses[0].id,
+            tableId,
+            automationId: expect.any(String),
```
samwho, line 139:
```suggestion
            automationId: expectAutomationId(),
```

For this and the 2 below.

packages/server/src/sdk/app/rowActions.ts
```diff
@@ -39,15 +45,53 @@ export async function create(tableId: string, rowAction: { name: string }) {
     doc = { _id: rowActionsId, actions: {} }
   }
 
+  const { name: tableName } = await tables.getTable(tableId)
+
   ensureUniqueAndThrow(doc, action.name)
 
+  const automation = await automations.create({
+    name: `${tableName}: ${action.name}`,
+    appId: context.getAppId()!,
```
samwho, line 54: Would it be better to instead check that `context.getAppId()` is defined and throw an error if not? The `!` is making me nervous, don't want to store bad data.

samwho (approved):

PR #14203, opened 2024-07-19
[BUDI-8460] Add New Step to Table Screen Template

Ghrehh:
## Ticket
https://linear.app/budibase/issue/BUDI-8460/templates-add-new-step-to-creating-table-screen

## Changes
- Added a modal variant of the table screen template.
- Added a variant of the table screen template that opens the edit and create forms in separate screens.
- The above creates 3 different screens for a single template, which required refactoring all the existing template generating functions. All these functions have a uniform interface which should be very extensible in future.
- All the table modal variants are now grouped together in a new second stage of the table template flow.
- Style changes to the second stage of the form template flow so that it and the second table template stage look uniform.
- Some style changes to the datasource selection stage of screen templates so that they look more uniform with the above, based on some designs Andy left lying around.
- The datasource selection modal now no longer overflows if the datasource (not the table or view, this was already fixed) has a very long name.
- All images used in this flow are now svgs, which should load much quicker than their PNG equivalents.
- The containers of these images now have a set height, so there shouldn't be any noticeable "jumping around" as the images load in.
- Removed the role indicators added in #14192 for reasons discussed there.
- Renames a bunch of existing code that refers to `table` as a `grid` as `table` seems to be the new preferred nomenclature.

|description|image|
|---|---|
|New Screen Modal|Screenshot 2024-08-06 at 14 36 25|
|Datasource selection|Screenshot 2024-08-06 at 14 36 32|
|Table subtype|Screenshot 2024-08-06 at 14 36 38|
|Form subtype|Screenshot 2024-08-06 at 14 36 45|
|Table New Screen Index|Screenshot 2024-08-06 at 15 09 01|
|Table New Screen Edit|Screenshot 2024-08-06 at 15 09 11|
|Table New Screen Create|Screenshot 2024-08-06 at 15 09 17|


packages/builder/src/helpers/data/datasourceSelectFormat.js
```diff
@@ -0,0 +1,19 @@
+const table = (table, datasources) => ({
```
Ghrehh, line 1: These formatters are to ensure that tables/views selected for generated templates _exactly_ match the expected format of the settings control/client prop they get fed into. 

Feeding in the raw table or view actually works, but the table selector will for instance trigger an update to massage the table into the format it expects, and when it does this it can trigger unexpected updates to other settings that depend on the table.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ScreenList/ScreenNavItem.svelte
```diff
@@ -20,17 +20,16 @@
...
     let duplicateScreen = Helpers.cloneDeep(screen)
     delete duplicateScreen._id
     delete duplicateScreen._rev
     duplicateScreen.props = makeComponentUnique(duplicateScreen.props)
 
     // Attach the new name and URL
-    duplicateScreen.routing.route = sanitizeUrl(screenUrl)
+    duplicateScreen.routing.route = sanitizeUrl(route)
     duplicateScreen.routing.homeScreen = false
-    duplicateScreen.props._instanceName = screenName
```
Comment (Ghrehh) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/ScreenList/ScreenNavItem.svelte:33]: screenName was always undefined

deanhannigan (changes requested):
Nice work @Ghrehh! Such a great refinement of the UX with solid updates to screen generation too. Minor comments and notes below.

`Bugs` - 🤷 

`Notes`

- Reduce default form size of the modal from `large` to `medium`. 
- Set `ignore clicks outside` to `true` for the modal table screen. Protect us from ourselves
-  Datasource selection modal title and description.  I don't think these changes are necessary as we also display views in the list. In the ticket the modal is marked as 'same as today'
    - Updated 
      - Title - `Choose a table to create your screen`
      - Description - removed
    - Original 
      - Title - `Autogenerated Screens`
      - Description text - `Select which datasources you would like to use to create your screens`

- Some of the urls are either inconsistent or deviated from the designs.
  - Form Screen
    - Create - `{datasource name}/new`. This pattern differs from the table create url.
  - Table Screen
    - Create - `{datasource name}/create` deviates from the design.
      - Original Design - `/{datasource name}/new/row`
    - View/Update `{datasource name}/edit/:id` deviates from the design.
      - Original Design - `{datasource name}/:id`

packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte
```diff
@@ -44,136 +43,83 @@
...
+    loadNewScreen(newScreens[0])
   }
 
-  const createFormScreen = async formType => {
-    let firstScreen = null
-
-    for (let tableOrView of selectedTablesAndViews) {
-      const screenTemplate = formScreen(
+  const createFormScreen = async type => {
+    const screenTemplates = selectedTablesAndViews.flatMap(tableOrView =>
```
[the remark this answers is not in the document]
Comment (Ghrehh) [reply packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte:109]: Sorry, I'd taken note to mention this in the main body of the PR, but looks like I forgot.

It's just so that all these generator functions have the same interface, so because one of them needs to return an array of screens, it makes sense (to me, at least) just to have them all follow that pattern. I think having them all use/return the same interface will save us some pain once we introduced more templates and we're forced to DRY up the new screen modal flow.

Ghrehh:
@deanhannigan thanks a bunch for the review, if you could give this another look when you get a chance please 🙏 

> Reduce default form size of the modal from `large` to `medium`.

Could you explain what you mean by this one please? Only form type setting I can think of is the one where you set the blank screen name, and that's already size M.


> Set `ignore clicks outside` to `true` for the modal table screen. Protect us from ourselves

Personally it drives me up the wall if you can't click outside a modal to close it (though that's maybe just a me thing and not something every user would feel). I also think this modal alone preventing outside clicks to close might be a little strange. I do get what you mean about it potentially blowing away a lot of stuff a user selected if they accidentally outside clicked. Maybe a confirmation dialog? It functions this way in master currently, so maybe we could punt on this for now, let me know if that sounds okay.

> Datasource selection modal title and description.  I don't think these changes are necessary as we also display views in the list. In the ticket the modal is marked as 'same as today'

Yeah that's fair, I thought it was a little strange that the title was "autogenerated screens", which seems more like a general title for the whole flow instead of a title for that individual step, but was definitely a little presumptuous of me to just go ahead and change all this. I changed it back 👍 


>  Urls

Ah, good catch on the "create" name, that was an oversight by me, should be `new`. 

When me and Cheeks talked through the [previous form screen ticket](https://linear.app/budibase/issue/BUDI-8363/form-screen-templates-hook-up-id-parameter) we decided on `name/edit/:id` and `name/new/:id`, so I was just trying to make the "new screen" grid template use the same pattern, I assumed the urls specified in the ticket were just a rough stand-in and were okay to change for consistency with the other templates. Let me know if that seems okay.

deanhannigan:
> Could you explain what you mean by this one please? Only form type setting I can think of is the one where you set the blank screen name, and that's already size M.

Yep, I was referring to the modal that the form is inside, not the form itself. It seemed a little on the big side.

> Set ignore clicks outside to true for the modal table screen. Protect us from ourselves

Now that you mention it, I remember there being a change in the works around that exact behaviour.

> Personally it drives me up the wall if you can't click outside a modal to close it (though that's maybe just a me thing and not something every user would feel)

Fair enough, totally personal preference here and they have the toggle there anyways. 

Regarding the urls and other changes, I only mentioned them here because I couldn't find any info on them in the ticket. My assumption is always that, at least regarding design, that the ticket is the final spec and any updates to that would be in the PR 👍

I'll give this a quick run through and get it approved 🎉

deanhannigan (approved):
LGTM!

PR #14204, opened 2024-07-19
[BUDI-8433] Default value column UI

samwho:
## Description

This PR creates the UI for default values. It's currently behind a feature flag so it will be hidden in production but shown in QA.

https://github.com/user-attachments/assets/91e29968-2938-4054-9fd5-a1082389ec26

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -133,7 +133,9 @@
   }
   $: initialiseField(field, savingColumn)
   $: checkConstraints(editableColumn)
-  $: required = !!editableColumn?.constraints?.presence || primaryDisplay
+  $: required = hasDefault
+    ? false
+    : !!editableColumn?.constraints?.presence || primaryDisplay
   $: uneditable =
     $tables.selected?._id === TableNames.USERS &&
     UNEDITABLE_USER_FIELDS.includes(editableColumn.name)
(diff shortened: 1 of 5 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -165,11 +168,20 @@
     editableColumn?.type !== AUTO_TYPE &&
     editableColumn?.type !== JSON_TYPE &&
     !editableColumn.autocolumn
+  $: canHaveDefault =
+    isEnabled(TENANT_FEATURE_FLAGS.DEFAULT_VALUES) &&
+    (editableColumn?.type === FieldType.NUMBER ||
```
adrinr, line 173: Is this controlled only via frontend? Don't we have a backend validation?

melohagan: NAB: could also refactor into `[FieldType.NUMBER, FieldType...].includes(editableColumn?.type)`

melohagan (approved):
LGTM - awesome feature! 😄 

I noticed that the default value supports bindings. Have you tried a few scenarios?

I wonder if this would allow you to do auto-increments, or other sequencing formulas?

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
samwho: Currently only controlled via the frontend. Should be easy to add backend validation if you thought it was worth it, but at the time I felt like it was easier to reason about this only being toggled in the frontend.

samwho: @melohagan In an ideal would I'd like to do a broader refactor of this file to rely far less on checking hardcoded field types, and instead have the FieldSchema dictate what can and cannot be done with a field, then genericise this component using the FieldSchema data.

samwho:
@melohagan there are some simple examples of using bindings as default values in `row.spec.ts` that I added in https://github.com/Budibase/budibase/pull/14177/files#diff-428e0cad26d239dd58cf676e4dfef825a1caca2ee998ef8183dfc44e9103572eR681. If you've got more complex ideas in mind it might be a neat idea to make testcases for them to make sure they work. 🙂

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
adrinr: Just worried of having inconsistencies if these values are set somehow. It it's not too hard, having a helper in shared-core that is used both in front and back would be great

adrinr (approved):
LGTM, just one final comment

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
samwho: Have had a go at what I think you're looking for, PTAL 🙂

PR #14209, opened 2024-07-22
Row action automation display in frontend

adrinr:
## Description
Displaying row action information, such as table name next to action name and trigger information banners. This required changing the fetch automation endpoint in order to make it more extensible without polluting the database objects or having copies of reference names

## Addresses
- [BUDI-8430 - Row action API - auto-create automation
](https://linear.app/budibase/issue/BUDI-8430/row-action-api-auto-create-automation)

## Screenshots
image

### DIFF
--- a/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
+++ b/packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
@@ -81,7 +81,7 @@
   // Check the schema to see if required fields have been entered
   $: isError =
     !isTriggerValid(trigger) ||
-    !trigger.schema.outputs.required?.every(
+    !(trigger.schema.outputs.required || []).every(
       required => $memoTestData?.[required] || required !== "row"
     )
 
--- a/packages/frontend-core/src/api/automations.js
+++ b/packages/frontend-core/src/api/automations.js
@@ -26,9 +26,13 @@ export const buildAutomationEndpoints = API => ({
   /**
    * Gets a list of all automations.
    */
-  getAutomations: async () => {
+  getAutomations: async ({ enrich }) => {
+    let url = "/api/automations?"
+    if (enrich) {
+      url += "enrich=true"
+    }
     return await API.get({
-      url: "/api/automations",
+      url,
     })
   },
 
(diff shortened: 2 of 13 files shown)


packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -81,7 +81,7 @@
   // Check the schema to see if required fields have been entered
   $: isError =
     !isTriggerValid(trigger) ||
-    !trigger.schema.outputs.required?.every(
+    !(trigger.schema.outputs.required || []).every(
```
mike12345567, line 84: Why is this change needed?

packages/frontend-core/src/api/automations.js
```diff
@@ -26,9 +26,13 @@ export const buildAutomationEndpoints = API => ({
   /**
    * Gets a list of all automations.
    */
-  getAutomations: async () => {
+  getAutomations: async ({ enrich }) => {
+    let url = "/api/automations?"
+    if (enrich) {
+      url += "enrich=true"
```
mike12345567, line 32: Can we use `URLSearchParams` here to add this to the URL? Helpful if this ever needs to be extended with other parameters.
e.g.
```
const params = new URLSearchParams()
if (enrich) {
  params.set("enrich", true)
}
const url = `/api/automations?${params.toString()}`
```

packages/server/src/api/controllers/automation.ts
```diff
@@ -73,8 +74,17 @@ export async function update(ctx: UserCtx) {
   builderSocket?.emitAutomationUpdate(ctx, automation)
 }
 
-export async function fetch(ctx: UserCtx) {
-  ctx.body = await sdk.automations.fetch()
+export async function fetch(ctx: UserCtx<void, FetchAutomationResponse>) {
+  const query = ctx.request.query || {}
```
mike12345567, line 78: Might be nice to type this, even inline

mike12345567 (approved):
LGTM - some NAB comments just!

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
adrinr: If we don't have any required field this will always be true, as `!undefined`. So, not having required fields, will always result on `isError = true`

mike12345567: Ah makes sense!

mike12345567 (approved):

PR #14210, opened 2024-07-22
Fix to show Formula column type picker again

deanhannigan:
## Description
Recent SQS changes have altered the way in which internal tables set the value `table.sql`. There were a couple of areas in the UI that used this to determine if a data source was internal.

## Addresses
  - Fix for the hidden `Formula` column type picker. This should be visible for Internal tables only as they support both `Static` formulas and the default, `Dynamic`.
    - This also ensures that the picker won't mistakenly appear for Google Sheets, which doesn't support `Static` formulas.
  - Fix to ensure internal sources get the `_rev` binding in the builder. The `!table.sql` check was excluding it from list.
    - `_rev` binding will be included for `table`, `link` and `viewV2` sources.
    - This also excludes the `_rev` binding for Google Sheets sources.

## Screenshots
This picker was inadvertently hidden when configuring a Formula column in an internal table.
Screenshot 2024-07-22 at 12 42 59

## Launchcontrol
Fix to show the Formula column type again for internal tables.

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -682,7 +682,7 @@
       {errors}
     />
   {:else if editableColumn.type === FORMULA_TYPE}
-    {#if !table.sql}
+    {#if !externalTable}
       <div class="split-label">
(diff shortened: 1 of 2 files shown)


aptkingston (approved):
LGTM!

PR #14212, opened 2024-07-22
Fix for duplicate columns in SQS

mike12345567:
## Description
There is an issue which was found last week around duplicate columns, columns that have the same name (user defined) but just different casing. This isn't a valid scenario and isn't really supported, but we don't want it to error when this occurs.

I've written in a case to handle this, updating the SQLite definitions to not include the duplicate as SQS simply cannot handle this today.


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2524,4 +2526,38 @@ describe.each([
...
+  isSqs &&
+    describe("duplicate columns", () => {
+      beforeAll(async () => {
+        table = await createTable({
+          name: {
+            name: "name",
+            type: FieldType.STRING,
+          },
+        })
+        await context.doInAppContext(config.appId!, async () => {
```
samwho, line 2539:
```suggestion
        await context.doInAppContext(config.getAppId(), async () => {
```

samwho (approved):

PR #14215, opened 2024-07-22
Trim data on automation sdk

adrinr:
## Description
Updating automation crud to not allow storing or retrieving any data that is not expected based its interface. This will automatically trim any unexpected data, otherwise any data sent via the requests can be persisted, adding noise to the database

## Launchcontrol
Ensure persisted automation data is the expected one

### DIFF
--- a/packages/server/src/sdk/app/automations/crud.ts
+++ b/packages/server/src/sdk/app/automations/crud.ts
@@ -1,4 +1,9 @@
-import { Automation, Webhook, WebhookActionType } from "@budibase/types"
+import {
+  Automation,
+  RequiredKeys,
+  Webhook,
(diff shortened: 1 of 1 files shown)


packages/server/src/sdk/app/automations/crud.ts
```diff
@@ -116,8 +124,7 @@ export async function create(automation: Automation) {
 }
 
 export async function update(automation: Automation) {
-  automation = { ...automation }
-
+  automation = trimUnexpectedObjectFields({ ...automation })
```
samwho, line 127: Do we need these spreads? Does this not result in two copies of the object when we only need one?

samwho (approved):

PR #14216, opened 2024-07-22
Support primitives in feature flags, make flag types flow, remove some obsolete feature flag systems.

samwho:
## Description

This is mostly a cleanup PR that's laying the foundation for adding PostHog to the list of feature flag sources in Budicloud.

I've taken the existing flag system that lives in `packages/backend-core/src/features/index.ts` and rejigged it to allow flags with values other than booleans, made types flow through it nicely, and made it very easy to add new flags with arbitrary default values.

Corresponding pro change: https://github.com/Budibase/budibase-pro/pull/306


packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
-      if (!tenantFeatureFlags[tenantId]) {
-        tenantFeatureFlags[tenantId] = []
-      }
-      tenantFeatureFlags[tenantId].push(feature)
-    })
-  })
+const DEFAULTS = Object.keys(FLAGS).reduce((acc, key) => {
+  const typedKey = key as keyof typeof FLAGS
+  // @ts-ignore
+  acc[typedKey] = FLAGS[typedKey].defaultValue
```
samwho, line 27: I couldn't find a nice way to make this type check because `FLAGS[typedKey].defaultValue` will have a type that's the union of all flags. So it's currently a boolean but when you add, say, a number flag it'll fail. I suspect it's possible to make this check correctly but I couldn't figure it out.

packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
+  return flags[name]
+}
+
+type BooleanFlags = {
+  [K in keyof typeof FLAGS]: (typeof FLAGS)[K] extends Flag<boolean> ? K : never
+}[keyof typeof FLAGS]
+
+// Convenience function for boolean flag values. This makes callsites more
+// readable for boolean flags.
+export async function isEnabled<K extends BooleanFlags>(
```
samwho, line 103: It required some type shenanigans but the outcome is that this function is _only_ usable with the boolean flag names. If you try to specify a non-boolean or non-existent flag it will not type check. This is an improvement in safety from just using `string`.

It will also mean that when it comes time to remove a flag, removing it from `FLAGS` will produce type check errors everywhere the flag is used.

packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
 
-export enum TenantFeatureFlag {
-  LICENSING = "LICENSING",
-  GOOGLE_SHEETS = "GOOGLE_SHEETS",
-  USER_GROUPS = "USER_GROUPS",
-  ONBOARDING_TOUR = "ONBOARDING_TOUR",
+// Gets a single feature flag value. This is a convenience function for
+// `fetch().then(flags => flags[name])`.
+export async function get<K extends keyof Flags>(name: K): Promise<Flags[K]> {
+  const flags = await fetch()
```
samwho, line 93: In a follow-up PR when I introduce PostHog, I suspect we will need to start caching feature flags per-tenant to avoid excessively calling out to the PostHod `decide` API.

packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
+        value = false
+      }
+
+      if (!isFlagName(feature)) {
+        throw new Error(`Feature: ${feature} is not an allowed option`)
+      }
+
+      if (typeof flags[feature] !== "boolean") {
+        throw new Error(`Feature: ${feature} is not a boolean`)
+      }
```
samwho, line 80: I was torn between making these errors or making them log a warning and get ignored. Because they're an environment variable, I erred toward erroring. It's reasonable to expect the environment variable to be strictly correct, imo.

When I come to add PostHog to this, I _think_ it makes more sense to ignore unknown flags. This will let us set them up before use and we wouldn't want adding a new flag in the PostHog UI to suddenly crash all of Budicloud.

packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
+
+      // @ts-ignore
+      flags[feature] = value
```
samwho, line 83: For reasons that make no sense to me, this throws the error `boolean cannot be assigned to never`. It thinks `flags[feature]` is of type `never`. It absolutely isn't, as demonstrated by the tests.

packages/backend-core/src/features/index.ts
mike12345567: Yeah I think I would generally tend towards ignoring anything we don't understand rather than error - while its good to enforce strictness here, if any of this is linked up to posthog there will be a risk of a catastrophic error, when it probably would be better to ignore and log the issue.

packages/backend-core/src/features/index.ts
mike12345567: That's very odd - perchance is this just the language server in VSCode having a hard time? I've checkout this branch and removed the `// ts-ignore` - with a build and a check:types I wasn't seeing any errors here.

packages/backend-core/src/features/index.ts
mike12345567: That's very odd - perchance is this just the language server in VSCode having a hard time? I've checkout this branch and removed the `// ts-ignore` - with a build and a check:types I wasn't seeing any errors here.

packages/backend-core/src/features/index.ts
```diff
@@ -1,79 +1,108 @@
...
+  private constructor(public defaultValue: T) {}
+}
 
-  env.TENANT_FEATURE_FLAGS.split(",").forEach(tenantToFeatures => {
-    const [tenantId, ...features] = tenantToFeatures.split(":")
+// This is the primary source of truth for feature flags. If you want to add a
+// new flag, add it here and use the `fetch` and `get` functions to access it.
+// All of the machinery in this file is to make sure that flags have their
+// default values set correctly and their types flow through the system.
+const FLAGS = {
```
mike12345567, line 17: One thing we should probably do - all these current flags are *very* old - I think we should probably remove them - we could instead use one of the features that is actually feature flagged right now instead, would it be worth considering moving SQS to use this mechanism instead?

All the feature flags that are here can 100% be removed as they should never be disabled now - the features are ancient, we've just never removed them!

packages/backend-core/src/features/index.ts
samwho: Huh. Yeah, it works. Must have been! 🙈

packages/backend-core/src/features/index.ts
samwho: Happy for me to do this in a follow-up PR?

packages/backend-core/src/features/index.ts
samwho: The idea would be to ignore missing flags for PostHog and error here for the environment variables. If you'd prefer I ignore for the environment variables it's easy to change, though I think it's better to error.

packages/backend-core/src/features/index.ts
mike12345567: Yeap sounds like a plan!

mike12345567 (approved):
LGTM!

PR #14217, opened 2024-07-22
Fix oracle parsing for CRUD operations, remove need to install oracle client

shogunpurple:
## Description
Fixes an issue with CRUD operations in Oracle, because of the `COALESCE` statement not working the way oracle expects. Oracle has `COALESCE` but I've just opted for a straight equals as I believe it has the same effect here.

Upgraded our oracle client to 6.x as well, which means we can take advantage of the "thin" client written in JS, and get rid of all the supporting code and scripts we had to install the oracle client in our docker containers. This should also allow the oracle connector to work locally on your macbook.

## Addresses
- https://linear.app/budibase/issue/BUDI-8469/oracle-throws-ora-00907-missing-right-parenthesis-when-performing-crud

## Feature branch env
[Feature Branch Link](http://fb-fix-oracle-parse.fb.qa.budibase.net)


adrinr:
Really nice changes @shogunpurple! Nice to finally get rid of this dependency ❤️
But there are some issues with the docker: https://github.com/Budibase/budibase-deploys/actions/runs/10055241880/job/27791512415

I can have a look at it for you if you want!

adrinr (commented):
Really nice changes @shogunpurple! Nice to finally get rid of this dependency ❤️
But there are some issues with the docker: https://github.com/Budibase/budibase-deploys/actions/runs/10055241880/job/27791512415

I can have a look at it for you if you want!

shogunpurple:
@adrinr that's an odd one, seems network related - I'll try a re-run here first just in case then dive a bit deeper. I'm not too sure how it could have been affected by removing the oracle script

adrinr (approved):
Rerunning 3 times worked 🤷🏽 
I tried it locally and it works fine! Just some unit test failing, but good change!

packages/backend-core/src/sql/sql.ts
```diff
@@ -435,6 +435,10 @@ class InternalBuilder {
             )} = ? THEN 1 ELSE 0 END = 1`,
             [value]
           )
+        } else if (this.client === SqlClient.ORACLE) {
+          query = query[fnc](`${quotedIdentifier(this.client, key)} = ?`, [
+            value,
+          ])
```
samwho, line 441: I was the one who originally put the coalesce calls in and they do serve a purpose when handling nulls. Because nulls in SQL tend to propagate, `anything = null` results in `null` rather than a boolean. This then continues to spread throughout the operation, resulting in strange behaviour.

If we were able to test Oracle in our search.spec.ts, I think what we'd find with this change is that it wouldn't be at parity with the other databases. This could be fine in the short term, because it's better than it being flat broken as it currently is, but long term we will want to make this work with coalescing.

samwho (approved):
Tentative approval, though I would say doing this "correctly" today would be worthwhile. I won't delay getting a fix out, though, and am happy for that to be a follow-up PR with comments alongside the Oracle if-statements linking to the bug.

packages/backend-core/src/sql/sql.ts
shogunpurple: Makes total sense - I've updated the PR and tests to use an oracle compatible version of `COALESCE`. The problem is that Oracle doesn't like it being used with booleans and it needs a condition at the end. Seems to be working well with the oracle connector after testing too 👍

PR #14218, opened 2024-07-23
Prevent edits and deletions for row action automations

adrinr:
## Description
Prevent deleting and editing row actions via the frontend (and via the API)

## Addresses
- [BUDI-8430 - Row action API - auto-create automation](https://linear.app/budibase/issue/BUDI-8430/row-action-api-auto-create-automation)

## Screenshots
image

image

image

### DIFF
--- a/packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
+++ b/packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
@@ -89,7 +89,7 @@
   on:contextmenu={openContextMenu}
   {icon}
   iconColor={"var(--spectrum-global-color-gray-900)"}
-  text={automation.name}
+  text={automation.displayName}
   selected={automation._id === $selectedAutomation?._id}
   hovering={automation._id === $contextMenuStore.id}
   on:click={() => automationStore.actions.select(automation._id)}
--- a/packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
+++ b/packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
@@ -35,47 +35,53 @@
   }
 
   const getContextMenuItems = () => {
-    return [
-      {
-        icon: "Delete",
-        name: "Delete",
-        keyBind: null,
-        visible: true,
-        disabled: false,
-        callback: confirmDeleteDialog.show,
+    const isRowAction = automation.definition.trigger.name === "Row Action"
+    const result = []
+    if (!isRowAction) {
+      result.push(
+        ...[
+          {
+            icon: "Delete",
+            name: "Delete",
+            keyBind: null,
+            visible: true,
+            disabled: false,
+            callback: confirmDeleteDialog.show,
+          },
+          {
+            icon: "Edit",
+            name: "Edit",
+            keyBind: null,
+            visible: true,
+            disabled: false,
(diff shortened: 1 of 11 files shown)


packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
```diff
@@ -35,45 +35,53 @@
...
-          automationStore.actions.toggleDisabled(
-            automation._id,
-            automation.disabled
-          )
-        },
+    const isRowAction = automation.definition.trigger.name === "Row Action"
+    const result = []
+    if (!isRowAction) {
+      result.push(
+        ...[
```
samwho, line 42: How come this spread is needed? Could you not ditch the array entirely and pass the objects in bare?

packages/server/src/api/controllers/automation.ts
```diff
@@ -94,6 +95,13 @@ export async function find(ctx: UserCtx) {
 export async function destroy(ctx: UserCtx<void, DeleteAutomationResponse>) {
   const automationId = ctx.params.id
 
+  const automation = await sdk.automations.get(ctx.params.id)
+  if (
+    automation.definition.trigger.stepId === AutomationTriggerStepId.ROW_ACTION
+  ) {
+    ctx.throw("Row actions cannot be deleted", 403)
```
samwho, line 102: nit: https://www.rfc-editor.org/rfc/rfc9110#status.403 suggests the 403 status is to do with authorisation, I suspect 400 is slightly more accurate here because you're attempting to do something that's invalid regardless of who you are.

packages/server/src/sdk/app/automations/crud.ts
```diff
@@ -281,4 +285,14 @@ function guardInvalidUpdatesAndThrow(
...
+
+  if (isRowAction(automation) && automation.name !== oldAutomation.name) {
+    throw new Error("Row actions cannot be renamed")
+  }
+}
+
+function isRowAction(automation: Automation) {
+  const result =
+    automation.definition.trigger.stepId === AutomationTriggerStepId.ROW_ACTION
+  return result
```
samwho, line 297: Should we throw this into shared-core so we can use it on the frontend too?

samwho (approved):

packages/builder/src/components/automation/AutomationPanel/AutomationNavItem.svelte
adrinr: Yes, that would be the same than 3 separate `.push(object)`.

packages/server/src/api/controllers/automation.ts
adrinr: I changed this 100 times and I did quite a lot of read. I agree that is not a 403, but, at the same time, it's not a 400. The request is not "bad" in terms of [format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400).

I will change it again to be a 400, as 403 might have more implications and anyways this is something that should never be called

packages/server/src/sdk/app/automations/crud.ts
adrinr: Good call!

packages/server/src/api/controllers/automation.ts
adrinr: Changed to 422, as it seems a good use case for it

PR #14224, opened 2024-07-23
Move "find row" to sdk

adrinr:
## Description
Moving the `row.getById` function to SDK. Currently, the code is dependent on controller context, and it makes it really hard to reuse in other contexts. We will need it for row actions, hence this refactor.
We also added a new find user metadata test, as this was causing the context dependency

## Feature branch env
[Feature Branch Link](http://fb-budi-8441-find-row-to-sdk.fb.qa.budibase.net)


mike12345567 (approved):
Very nice!

PR #14228, opened 2024-07-24
Don't show datasource options for array bindings that lack table IDs

aptkingston:
## Description
Fixes an issue where invalid datasources were being listed for data providers. This PR removes any array-like datasources generated from bindings that lack a table ID.

## Addresses
- https://linear.app/budibase/issue/BUDI-8388/%5Bqa-wolf%5D-repeater-does-not-get-data-from-data-providers-remove


Ghrehh (approved):
LGTM

PR #14229, opened 2024-07-24
Fix failing test due testcontainer conflicts

adrinr:
## Description
Fixing mssql testcontainer wait strategy. The image has been updated and the script was broken, not allowing starting the containers properly. Changing the tag to not the latest fixes it temporarily.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -10,7 +10,18 @@
   },
   "targetDefaults": {
     "build": {
-      "inputs": ["{workspaceRoot}/scripts/*", "{workspaceRoot}/lerna.json"]
+      "inputs": [
+        "{workspaceRoot}/scripts/*",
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -10,7 +10,18 @@
   },
   "targetDefaults": {
     "build": {
-      "inputs": ["{workspaceRoot}/scripts/*", "{workspaceRoot}/lerna.json"]
```
adrinr, line 13: This will enforce running build and test when the workflows are updated. Otherwise, something might be skipped if there are no changes on the actual packages

adrinr:
Related to https://github.com/microsoft/mssql-docker/issues/892

mike12345567 (approved):

PR #14237, opened 2024-07-24
Lock CI images to specific SHAs

mike12345567:
## Description
Fixing images to use a locked SHA, this means that we shouldn't have issues with CI breaking due to using images which are receiving updates (we've experienced two CI breakages due to MariaDB and MS-SQL updates).

Hopefully should avoid issues like: https://github.com/microsoft/mssql-docker/issues/892 breaking CI again in future.


packages/server/src/api/routes/tests/table.spec.ts
```diff
@@ -28,11 +28,11 @@ const { basicTable } = setup.structures
...
 describe.each([
-  ["internal", undefined],
+  // ["internal", undefined],
   [DatabaseName.POSTGRES, getDatasource(DatabaseName.POSTGRES)],
-  [DatabaseName.MYSQL, getDatasource(DatabaseName.MYSQL)],
-  [DatabaseName.SQL_SERVER, getDatasource(DatabaseName.SQL_SERVER)],
-  [DatabaseName.MARIADB, getDatasource(DatabaseName.MARIADB)],
+  // [DatabaseName.MYSQL, getDatasource(DatabaseName.MYSQL)],
+  // [DatabaseName.SQL_SERVER, getDatasource(DatabaseName.SQL_SERVER)],
+  // [DatabaseName.MARIADB, getDatasource(DatabaseName.MARIADB)],
```
samwho, line 35: 👀

mike12345567: woops - thank you!

packages/server/src/integrations/tests/utils/images.ts
```diff
@@ -0,0 +1,17 @@
+import dotenv from "dotenv"
+import { join } from "path"
+
+const path = join(__dirname, "..", "..", "..", "..", "datasource-sha.env")
+dotenv.config({
+  path,
+})
+
+export function getImageSHAs() {
```
adrinr, line 9: Out of curiosity, why is this a function instead just exporting the strings?

mike12345567: good point!

samwho (approved):

PR #14239, opened 2024-07-24
Adding bulkRemove utility for CouchDB

mike12345567:
## Description
Adding a small utility to DB layer for bulk removing documents, this is a problem that Mel ran into, the fact it doesn't default to throwing errors, I've updated a few cases and added functionality for it to maintain compatibility with the old way of doing things (errors silenced).

@melohagan hopefully this helps!


packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -246,6 +246,35 @@ export class DatabaseImpl implements Database {
...
+    for (let res of response) {
+      if (res.error) {
+        errorFound = true
+        errorMessage += res.error
+      }
+    }
+    if (errorFound) {
+      throw new CouchDBError(errorMessage, {
+        name: this.name,
+        status: 400,
```
melohagan, line 273: Might be nice to surface the error status code from CouchDB ?

melohagan (approved):
LGTM!

packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -56,24 +56,24 @@ class CouchDBError extends Error implements DBError {
   constructor(
     message: string,
     info: {
-      status: number | undefined
-      statusCode: number | undefined
+      status?: number | undefined
```
adrinr, line 59: The `| undefined`s are now redundant and they can be removed

packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -246,6 +246,35 @@ export class DatabaseImpl implements Database {
     })
   }
 
+  async bulkRemove(documents: Document[], opts?: { silenceErrors?: boolean }) {
+    const response: Nano.DocumentBulkResponse[] = await this.performCall(db => {
+      return () =>
+        db.bulk({
+          docs: documents.map(doc => ({
```
adrinr, line 253: Is this atomic, or can we get an inconsistent state?

packages/backend-core/src/db/couch/DatabaseImpl.ts
adrinr, line 273: Why 400?

packages/backend-core/src/platform/users.ts
```diff
@@ -113,15 +113,12 @@ export async function addUser(
...
-  const toDelete = userDocs.rows.map((row: any) => {
-    return {
-      ...row.doc,
-      _deleted: true,
-    }
-  })
-  await db.bulkDocs(toDelete)
+  await db.bulkRemove(
+    userDocs.rows.map(row => row.doc!),
+    { silenceErrors: true }
```
adrinr, line 122: Why do we want to silence errors? Can't we have issues with concurrent update & delete, or similar?

packages/backend-core/src/db/couch/DatabaseImpl.ts
mike12345567: It isn't atomic no, but there isn't really anyway to get around that.

packages/backend-core/src/db/couch/DatabaseImpl.ts
mike12345567: Unsure what status code to use - it isn't a 500 as we got a 200 back from the bulk API which lead to this, generally CouchDB uses 400 when its not a 409 and its just due to some sort of user input - this felt like the best code based on the scenario (something in the list couldn't be deleted, but it didn't go totally wrong).

packages/backend-core/src/platform/users.ts
mike12345567: This is just for consistency with how we used to use this, previously when we used this without a utility errors would be ignored, I wanted to use this in a few places as it made the typing better, but I didn't want to introduce a new failure point, some of these APIs are years old and have been ignoring errors in deletion for all that time, I didn't want to change the behaviour.

PR #14247, opened 2024-07-25
Row action run api

adrinr:
## Description
Create endpoint to allow running row action automations. This will require some further refactor and deeper testing when building the frontend for it, but it will work as initial implementation. This is a new endpoint, so it's safe to merge as it's not currently use by anything


packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -7,7 +9,6 @@ import {
 import tk from "timekeeper"
 import emitter from "../../../../src/events"
 import { outputProcessing } from "../../../utilities/rowProcessor"
-import * as setup from "./utilities"
```
adrinr, line 10: This was causing issues with the js order loading, with the barrel files of the SDK and supertest

samwho (approved):

PR #14252, opened 2024-07-26
Fix importing internal tables

adrinr:
## Description
With the recent fixes preventing duplicated table ids, internal tables imports are broken. The checks are preventing using protected column names such as `_id`, but the export/import should support it.

## Launchcontrol
Fix importing internal tables with protected column names such as `_id`

### DIFF
--- a/packages/server/src/sdk/app/tables/internal/index.ts
+++ b/packages/server/src/sdk/app/tables/internal/index.ts
@@ -31,7 +31,7 @@ export async function save(
     tableId?: string
     rowsToImport?: Row[]
     renaming?: RenameColumn
-    isImport: boolean
+    isImport?: boolean
   }
 ) {
   const db = context.getAppDB()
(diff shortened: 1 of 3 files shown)


packages/server/src/sdk/app/tables/internal/index.ts
```diff
@@ -47,7 +48,9 @@ export async function save(
   }
 
   // check for case sensitivity - we don't want to allow duplicated columns
-  const duplicateColumn = findDuplicateInternalColumns(table)
+  const duplicateColumn = findDuplicateInternalColumns(table, {
+    internalRowsAllowed: !!opts?.isImport,
```
mike12345567, line 52: Can we name this to something which explains it a bit better? `internalRowsAllowed` doesn't really explain that its not checking for duplicates in the internal column names.

adrinr: Done as agreed!

mike12345567 (approved):

PR #14253, opened 2024-07-26
Fix/verify sso bug

melohagan:
## Description
Niche bug with Microsoft SSO and password accounts. 

When verifying an MS SSO email account, if this same email address had been used to create a password account by an admin, you would be unable to verify the email again if the user was deleted and re-added.

This is because the SSO user profile CouchDB doc was not being removed on deletion of the user. 

## Screenshots

two user docs
*When a Microsoft user is added with a password account, but signs in with SSO, you get multiple documents in Couch.*

Screenshot 2024-07-26 at 13 32 13
*Deletion of the user in this screen was not removing the SSO user profile, which prevents subsequent email verification if that user is re-added.*

**Also updated the password reset screen:**
https://linear.app/budibase/issue/GRO-560/make-the-reset-your-password-support-the-enter-key

https://github.com/user-attachments/assets/80e336f3-7dd3-4df6-ab88-608c085128a5

## Launchcontrol
Minor bug fix relating to Microsoft SSO


packages/backend-core/src/users/lookup.ts
```diff
@@ -36,7 +36,7 @@ export async function searchExistingEmails(emails: string[]) {
 // lookup, could be email or userId, either will return a doc
 export async function getPlatformUser(
   identifier: string
-): Promise<PlatformUser | null> {
+): Promise<PlatformUser[] | PlatformUser | null> {
```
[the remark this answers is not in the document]
melohagan: It returns an array if the following Couch view finds more than one document:

```ts
const viewJs = `function(doc) {
    if (doc.tenantId) {
      emit(doc._id.toLowerCase(), doc._id)
    }

    if (doc.ssoId) {
      emit(doc.ssoId, doc._id)
    }
  }`
```

In this case it is these two documents:

Screenshot 2024-07-26 at 15 41 23

*The document has the property ssoId, so emit that property as the key* 

Screenshot 2024-07-26 at 15 41 46

*The document has the property tenantId, so emit the _id property as the key*

This will result in two documents matching the SSO ID.

packages/backend-core/src/platform/users.ts
```diff
@@ -25,6 +25,11 @@ export async function getUserDoc(emailOrId: string): Promise<PlatformUser> {
   return db.get(emailOrId)
 }
 
+export async function updateUserDoc(platformUser: Partial<PlatformUserById>) {
```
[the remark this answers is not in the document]
melohagan: Partial removed!

packages/backend-core/src/users/lookup.ts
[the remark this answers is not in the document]
adrinr: If this is the case, we might have all the usages wrong. You are casting the results as `PlatformUser`, so if it returns an array we will have issues. Should we handle the possible arrays on the usages? (fe. `packages/backend-core/src/users/db.ts, line 454`)

packages/backend-core/src/users/db.ts
```diff
@@ -446,9 +453,32 @@ export class UserDB {
       creator => !!creator
     ).length
 
+    const ssoUsersToDelete: AnyDocument[] = []
     for (let user of usersToDelete) {
+      const platformUser = (await getFirstPlatformUser(
+        user._id!
+      )) as PlatformUserById
```
adrinr, line 460: If I am right, this cast should not be needed anymore

packages/backend-core/src/users/lookup.ts
```diff
@@ -34,15 +34,27 @@ export async function searchExistingEmails(emails: string[]) {
...
+  })
+}
+
+export async function getFirstPlatformUser(
+  identifier: string
+): Promise<PlatformUser | null> {
+  // use the view here and allow to find anyone regardless of casing
+  // Use lowercase to ensure email login is case insensitive
+  const platformUserDocs = await getPlatformUsers(identifier)
+  if (platformUserDocs.length > 0) {
```
adrinr, line 54: This could be simplified to `platformUserDocs[0] ?? null`

packages/builder/src/components/common/users/PasswordRepeatInput.svelte
```diff
@@ -1,40 +1,52 @@
...
-  import { Layout, Input } from "@budibase/bbui"
+  import { FancyForm, FancyInput } from "@budibase/bbui"
   import { createValidationStore, requiredValidator } from "helpers/validation"
 
   export let password
+  export let passwordForm
   export let error
 
+  const validatePassword = value => {
+    if (!value || value.length < 12) {
```
adrinr, line 10: Is this 12 right? We use 8 by default on the backend, and it can be specified case by case. Should we use the same `PASSWORD_MIN_LENGTH` value?

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -71,6 +72,15 @@ export const addSsoSupport = async (ctx: Ctx<AddSSoUserRequest>) => {
       userByEmail.userId,
       userByEmail.tenantId
     )
+    // Need to get the _rev of the user doc to update
+    const userById = (await platform.users.getUserDoc(
+      userByEmail.userId
+    )) as PlatformUserById
```
adrinr, line 78: Same, probably not needed if using the new `getFirstPlatformUser` function

adrinr (approved):

packages/backend-core/src/users/db.ts
melohagan: It is unfortunately because of the line below which assumes the presence of the ssoId. The platform user types are a little strange because of the multiple Couch docs:

```ts
/**
 * doc id is user email
 */
export interface PlatformUserByEmail extends Document {
  tenantId: string
  userId: string
}

/**
 * doc id is userId
 */
export interface PlatformUserById extends Document {
  tenantId: string
  email?: string
  ssoId?: string
}

/**
 * doc id is a unique SSO provider ID for the user
 */
export interface PlatformUserBySsoId extends Document {
  tenantId: string
  userId: string
  email: string
  ssoId?: string
}

export type PlatformUser =
  | PlatformUserByEmail
  | PlatformUserById
  | PlatformUserBySsoId
```

PR #14255, opened 2024-07-26
"Create screen" component context option

deanhannigan:
## Description
A `Create screen` option has been added to the component context menu in the builder. Selecting the option will:
- Generate a brand new screen with the `Basic` role
- Clone the selected component and place it in the the screen.
- The default screen path will be `/new-screen`. If you create multiple screens in a row a suffix will be appended to avoid collision e.g. `/new-screen-2`

## Addresses
- `Fix` - component context options were being cached between screens. When pasting or cloning into new screens, it wasn't immediately clear what elements had become broken due to broken context references until full page refresh. The component options will now clear when the active screen is changed.
- Minor refactoring of the screen store to centralise some screen autogeneration functionality.

## Screenshots
Screenshot 2024-07-26 at 14 58 07

## Launchcontrol
Create a new screen from a selected component

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte
@@ -12,7 +12,6 @@
     builderStore,
   } from "stores/builder"
   import { auth } from "stores/portal"
-  import { get } from "svelte/store"
   import getTemplates from "templates"
   import { Roles } from "constants/backend"
   import { capitalise } from "helpers"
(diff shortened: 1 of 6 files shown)


packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte
```diff
@@ -53,11 +52,13 @@
...
+        if (
+          screenStore.hasExistingUrl(screen.routing.route, screenAccessRole)
+        ) {
           let suffix = 2
-          let candidateUrl = makeCandidateUrl(screen, suffix)
-          while (hasExistingUrl(candidateUrl)) {
-            candidateUrl = makeCandidateUrl(screen, ++suffix)
+          let candidateUrl = screenStore.makeCandidateUrl(screen, suffix)
+          while (screenStore.hasExistingUrl(candidateUrl, screenAccessRole)) {
+            candidateUrl = screenStore.makeCandidateUrl(screen, ++suffix)
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/_components/NewScreen/CreateScreenModal.svelte:61]: Rather than all this logic (some of which I know was pre-existing), we should use our `getSequentialName` utility which is designed to do exactly this!

packages/client/src/components/ClientApp.svelte
```diff
@@ -46,11 +46,24 @@
...
+
+  // Clear component context entries from the global context
+  // when the active page changes.
+  const purgeGlobalComponentContext = screenId => {
+    if (!screenId || cachedActiveScreenId == screenId) return
+    cachedActiveScreenId = screenId
+    globalContext.actions.clearComponentContext()
+  }
+
+  $: purgeGlobalComponentContext($screenStore?.activeScreen?._id)
```
aptkingston, line 66: I understand the purpose of this, but I think we should come up with a cleaner way of handling it than trying to whitelist all properties that we want to persist between screen changes. I have a few ideas in mind... For now I think it would be better to take this out of this PR as it's a separate issue, and we can tackle it elsewhere.

aptkingston (commented):
Nice one Dean! Happy to get this in if we remove the context stuff for now. Also one wee change suggested which should remove a bunch of code as we already have a utility for generating sequential names 👌

PR #14257, opened 2024-07-26
Fix for searching by row ID (with a limit) when the row has many related rows

mike12345567:
## Description
When searching by row ID with external DBs/SQS we can get into a situation where the limit of 1 which is applied by the frontend can cause problems, with many to many relationships we need to retrieve multiple rows (all of the joined related rows). 

This bug exhibits itself in one part of the platform, when attempting to a row by ID in a form block that has multiple many to many relationships. The frontend needs to be able to send a limit of 1 incase it is using a form block but hasn't gotten a row ID (this can happen in preview/the builder) and it just wants to populate with a row for display.

Thanks to @poirazis for raising this issue!


packages/server/src/sdk/app/rows/search/utils.ts
```diff
@@ -108,3 +108,18 @@ export function searchInputMapping(table: Table, options: RowSearchParams) {
...
   return options
 }
+
+export function isSearchingByRowID(query: SearchFilters): boolean {
+  for (let searchField of Object.values(query)) {
+    if (typeof searchField !== "object") {
+      continue
+    }
+    const hasId = Object.keys(searchField).find(
+      key => key.endsWith("_id") && searchField[key]
```
adrinr, line 118: Why is this endsWith? Could this be conflicting with things such as `department_id` or any other column name ending with `_id`?

adrinr (changes requested):

packages/server/src/sdk/app/rows/search/utils.ts
mike12345567: True - I was trying to handle for prefixes - but I'll do it a different way!

adrinr: We have a util that trims the prefixes. We should use that

adrinr (approved):
🚀

PR #14263, opened 2024-07-29
Fix Oracle support as a datasource and add tests for it.

samwho:
## Description

This PR adds Oracle to our list of databases that we spin up during tests and hit for real. It ensures parity between Oracle and other external databases within Budibase.

Adding Oracle to various test suites highlighted that it was broken in a lot of fundamental ways, and this PR also includes scores of fixes to the various ways it was broken. As such, it is a large and complicated PR. The result, however, is that Oracle should work as well as Postgres and others as an external datasource in Budibase.

## Launchcontrol

Fixes dozens of problems with using OracleDB as an external datasource for Budibase, and adds Oracle as one of the databases we hit in our integration tests to help prevent it breaking in future.


packages/server/src/integrations/oracle.ts
```diff
@@ -386,7 +500,10 @@ class OracleIntegration extends Sql implements DatasourcePlus {
       password: this.config.password,
       connectString,
     }
-    return oracledb.getConnection(attributes)
+    const tz = Intl.DateTimeFormat().resolvedOptions().timeZone
```
samwho, line 503: TODO: add comment explaining why this

mike12345567 (approved):
LGTM! We discussed this on a call and addressed some comments there.

I've created a sustaining ticket: https://linear.app/budibase/issue/BUDI-8540/backend-sql-generation-break-down-sqlts

Would like to revisit this in the future to try and solve the issue of `this.client === ORACLE || MSSQL` that get littered throughout the SQL generation.

PR #14265, opened 2024-07-29
Fix issue with stepId being undefined when importing app

PClmnt:
## Description
Fixes an issue where importing an app caused an error due to an undefined stepId in automations. 

## Addresses
https://linear.app/budibase/issue/BUDI-8516/importing-an-app-with-an-older-automation-seems-to-breaking-the-import


adrinr (approved):

packages/shared-core/src/sdk/documents/automations.ts
```diff
@@ -2,12 +2,12 @@ import { Automation, AutomationTriggerStepId } from "@budibase/types"
 
 export function isRowAction(automation: Automation) {
   const result =
-    automation.definition.trigger.stepId === AutomationTriggerStepId.ROW_ACTION
+    automation.definition.trigger?.stepId === AutomationTriggerStepId.ROW_ACTION
```
adrinr, line 5: This is not nullable on the type definition, probably we should check how come this comes null.
Should we check also if `definition` is null?

PClmnt: Yea i'm not sure how I got like that, but it will be in the definition for it now regardless. Added a check on the definition property as well just to be sure.

PR #14273, opened 2024-07-30
update account-portal ref

melohagan:
## Description
Point to correct account-portal master branch


shogunpurple (changes requested):
I would get rid of the `yarn.lock` changes here, they aren't relevant

shogunpurple (approved):

PR #14279, opened 2024-07-30
Disallow adding row action automations

mike12345567:
## Description
Fixes some issues with row actions which were allowing them to be added from the automation section (which is not allowed) as well as breaking app access once they were added - this hides them properly so they can't be added as well as fixing the issue when they exist.

### DIFF
--- a/packages/server/src/automations/utils.ts
+++ b/packages/server/src/automations/utils.ts
@@ -9,10 +9,8 @@ import { cloneDeep } from "lodash/fp"
 import { quotas } from "@budibase/pro"
 import {
   Automation,
-  AutomationActionStepId,
   AutomationJob,
   AutomationStepSchema,
-  AutomationTriggerStepId,
 } from "@budibase/types"
 import { automationsEnabled } from "../features"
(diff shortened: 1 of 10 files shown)


adrinr (approved):
Good catch!

packages/server/src/automations/utils.ts
```diff
@@ -111,10 +117,16 @@ export async function updateTestHistory(
   )
 }
 
-export function removeDeprecated(definitions: any) {
+export function removeInvalidDefinitions(
+  definitions: Record<string, AutomationStepSchema>
+) {
+  const disallowedStepIds: (
```
adrinr, line 123: Should this not be handled on the frontend? We just want not to be able to add these types, but they should be part of the definitions. These are used for automation logs and similar, so they should be there

mike12345567: For now I think they should be fully removed - since row actions isn't released yet I think its best if there's no mention of it at all.

mike12345567: Discussed - definitions are used for automation history as well, going to leave automation definitions full for now instead, logic for disabling creation is on frontend.

PR #14281, opened 2024-07-30
Improve typings

adrinr:
## Description
Type untyped APIs and improve schema typing

### DIFF
--- a/packages/server/src/api/routes/tests/row.spec.ts
+++ b/packages/server/src/api/routes/tests/row.spec.ts
@@ -32,6 +32,7 @@ import {
   TableSourceType,
   UpdatedRowEventEmitter,
   TableSchema,
+  JsonFieldSubType,
 } from "@budibase/types"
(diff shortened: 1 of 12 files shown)


packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -102,7 +103,7 @@ describe.each([
   ): SaveTableRequest {
     const defaultSchema: TableSchema = {
       id: {
-        type: FieldType.AUTO,
```
adrinr, line 105: The `id` field is mandatory for external ds, and having it as auto causes conflicts with parallel creations. Having them as `number` solves that issue and its closer to the real BB use cases

mike12345567: This is correct anyway - no columns actually have a `FieldType.AUTO` type.

mike12345567 (approved):

PR #14292, opened 2024-07-31
SQL pagination improvements

mike12345567:
## Description
This PR includes a change to pagination which makes sure if the 5000 max row limit is hit that pagination still kicks in. This means that you can eventually return all rows, although for very large tables you may hit rate limits (if you have thousands of rows related to each row in your table).

Previously our external and SQS SQL implementations assumed that each page contained the limit of rows, but due to the 5000 row internal limit when tables have a lot of relationships pages could be required even if the page limit isn't reached.

For example if you have 50 rows in our table and search with a page limit of 100, but each of the 50 rows has 1000 rows related to it, after 5 rows you would hit the internal row limit. This method makes sure that the bookmark returned denotes where it stopped returning rows rather than assuming its returned everything.

This also puts a fix in place for SQS when queries generate a request that would return over 2000 columns - this limits the response and forces it to always return under 2000 fields.


packages/server/src/sdk/app/rows/search/external.ts
```diff
@@ -105,37 +106,42 @@ export async function search(
...
-      queries.push(handleRequest(Operation.COUNT, tableId, parameters))
+      queries[1] = handleRequest(Operation.COUNT, tableId, parameters)
     }
     const responses = await Promise.all(queries)
-    let rows = responses[0] as Row[]
-    const totalRows =
-      responses.length > 1 ? (responses[1] as number) : undefined
+    let rows = responses[0].rows
+    const rawResponseSize = responses[0].rawResponseSize
+    const totalRows = responses.length > 1 ? responses[1] : undefined
```
samwho, line 119: Can make this a bit more readable with some sneaky destructuring:

```typescript
const [{ rows, rawResponseSize }, totalRows] = await Promise.all([
  handleRequest(Operation.READ, tableId, parameters),
  countRows ? handleRequest(Operation.COUNT, tableId, parameters) : Promise.resolve(undefined)
])

**Review by samwho — COMMENTED** [review]
This very much needs tests, sorry 😅

File: packages/server/src/sdk/app/rows/search/external.ts [hunk 2]
```diff
@@ -105,37 +106,42 @@ export async function search(
...
     }
-    const queries: Promise<Row[] | number>[] = []
-    queries.push(handleRequest(Operation.READ, tableId, parameters))
+    const queries: [
+      Promise<ExternalReadRequestReturnType>,
+      Promise<number> | undefined
+    ] = [handleRequest(Operation.READ, tableId, parameters), undefined]
     if (countRows) {
-      queries.push(handleRequest(Operation.COUNT, tableId, parameters))
+      queries[1] = handleRequest(Operation.COUNT, tableId, parameters)
```
adrinr, line 114: What is this 1? Why not a push?

mike12345567: This is just so that I could type the first and second array elements, the second array element is typed as undefined or a count result, so needs replaced - but doesn't matter with Sam's above comment.

samwho (approved):

PR #14294, opened 2024-07-31
Dont allow protected column names on import

adrinr:
## Description
We have some protected column names, such as _id and tableId, that can not be used in the schema. When exporting some csv we are exporting some of these columns, and when creating a new table via CSV upload allowed creating these columns as part of the schema.
This PR ensures that this is not the case, both on the API level and on the client side

https://github.com/user-attachments/assets/bee97c6c-b575-48d5-a387-91ca53f3418a

## Launchcontrol
Handle protected column names on table imports

### DIFF
--- a/packages/server/src/api/routes/tests/table.spec.ts
+++ b/packages/server/src/api/routes/tests/table.spec.ts
@@ -123,6 +123,29 @@ describe.each([
         body: basicTable(),
       })
     })
+
+    it("does not persist the row fields that are not on the table schema", async () => {
+      const table: SaveTableRequest = basicTable()
+      table.rows = [
+        {
+          name: "test-name",
+          description: "test-desc",
+          nonValid: "test-non-valid",
+        },
+      ]
+
+      const res = await config.api.table.save(table)
+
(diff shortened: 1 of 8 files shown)


packages/server/src/api/routes/tests/table.spec.ts
```diff
@@ -1053,6 +1115,70 @@ describe.each([
...
+            id: true,
+            name: true,
+            [columnName]: false,
+          },
+        })
+      })
+
+      isInternal &&
+        it.each(
+          isInternal ? PROTECTED_INTERNAL_COLUMNS : PROTECTED_EXTERNAL_COLUMNS
```
samwho, line 1151: Is this needed? Doesn't the outer condition make this redundant?

samwho (approved):

packages/server/src/api/routes/tests/table.spec.ts
adrinr: Amended here: https://github.com/Budibase/budibase/pull/14302

PR #14297, opened 2024-08-01
Remove global node-fetch mock.

samwho:
## Description

This PR removes the global `__mocks__/node-fetch.ts` file in `packages/server` in favour of using `nock` where needed.

Additionally I've set up `nock` to disallow any requests to the internet outside of localhost.


packages/server/src/api/controllers/plugin/index.ts
```diff
@@ -75,14 +84,9 @@ export async function create(ctx: any) {
     const doc = await pro.plugins.storePlugin(metadata, directory, source)
 
     clientAppSocket?.emit("plugins-update", { name, hash: doc.hash })
-    ctx.body = {
```
mike12345567, line 78: Nice find!

samwho: It fell out after I typed the response object.

mike12345567 (approved):
LGTM! Very nice!

PR #14301, opened 2024-08-01
Remove all mocking from rest.spec.ts.

samwho:
## Description

Following on from https://github.com/Budibase/budibase/pull/14297, this PR dramatically reduces the reliance on mocks for `rest.spec.ts`.

In the process of doing this, I believe I've discovered some functionality that never worked. The `legacyHttpParser` setting sets an option in `node-fetch` that only appears to be available in version 3 and above, where we're stuck on version 2.


mike12345567 (approved):
Pretty massive refactor, nicely done!

PR #14302, opened 2024-08-02
User column csv import error

adrinr:
## Description
Fix importing user/users columns when exported as CSV. The current logic is handling CSV strings as JSON, but this is not true for our CSV exported objects. Adding some tests to check that the code works as expected.

The current export/import flow does:
1. Export as CSV or JSON
2. Import the file
3. If it's a CSV, call the endpoint `/csvtojson`
4. Validate the outputting json via `/validateNewTableImport` or `/validateExistingTableImport`
5. Import the data via table or bulk row import

Because of this flow is highly dependent on the previous steps, we added a test that covers the full flow

## Launchcontrol
Fix importing user/users data

## Feature branch env
[Feature Branch Link](http://fb-budi-8396-user-column-csv-import-error.fb.qa.budibase.net)


mike12345567 (approved):
Very nice work - great to have massively improved tests around exporting!

PR #14305, opened 2024-08-02
Fixing export rows to use correct search

mike12345567:
## Description
Refactoring search SDK to make it obvious the different search methods, exports was using lucene always when doing internal export rows which shouldn't be the case, should go through the complete search SDK.


adrinr (approved):
Nice cleanup!

PR #14307, opened 2024-08-02
Make sure to delete SSO ID doc as well

melohagan:
## Description
When an account is deleted, it correctly removes associated CouchDB docs based on email address, but it was leaving behind the SSO doc. 
This has been fixed to make sure that the SSO doc is also deleted.

## Launchcontrol
Minor bug fix relating to Microsoft SSO


samwho (approved):
It would be good to get a test for this, to prevent it from regressing in future.

PR #14315, opened 2024-08-05
Assorted builder fixes and a fix for Google sheets filtering

deanhannigan:
## Description

Mixed bag of Builder bug fixes and Google Sheets filtering

## Addresses
  - Single User Field not usable in custom schema - `field/bb_reference_single` wasn't configured in the builder. 
    - https://linear.app/budibase/issue/BUDI-8521/-unable-to-use-the-user-field-component
  - Fix signature display. The signature would stretch in tables with larger row heights
    - https://linear.app/budibase/issue/BUDI-8504/improve-display-of-signature-cells-in-table-component
  - Unable to deselect a `Select` field IF a placeholder was set, then cleared. The placeholder is then always empty. 
    - An empty placeholder will remove the "Choose an option" option and the field is no longer clearable 
    - Affected the following form fields: `Single Select`, `Relationship(ManyToOne)`, `Single User`
 - Fix for GoogleSheets filtering 
   - Added optional `findInDoc` function parameter to `runQuery` as the current default (`deepGet`) assumes it receives a simple object. In the case of GoogleSheets it actually receives a `GoogleSpreadsheetRow` object
  https://linear.app/budibase/issue/BUDI-8462/search-feature-in-data-table-not-working-for-google-sheets
 - Fixes to ensure that `Single select` and `Single user` fields relay their binding status in automations. Previously, both fields just said `Choose an option` and it was not clear that a binding was present in the field at all.
   - https://linear.app/budibase/issue/BUDI-8505/no-ui-indication-for-a-user-field-binding-in-row-automation-steps 
   - Handlebars string
   Screenshot 2024-08-06 at 11 19 07
   - Javascript binding
   Screenshot 2024-08-06 at 11 46 19

## Launchcontrol
Bug fixes and fix for Google sheets filtering

## Feature branch env
[Feature Branch Link](http://fb-dean-fixes.fb.qa.budibase.net)


packages/bbui/src/Form/Core/Select.svelte
```diff
@@ -102,7 +102,9 @@
   {onOptionMouseenter}
   {onOptionMouseleave}
   isPlaceholder={value == null || value === ""}
-  placeholderOption={placeholder === false ? null : placeholder}
+  placeholderOption={placeholder === false
```
melohagan, line 105: NAB: Would be nice if placeholder was of type `string` instead of `string | false`

packages/builder/src/components/common/LinkedRowSelector.svelte
```diff
@@ -15,14 +15,25 @@
...
   $: linkedTableId = linkedTableId || schema.tableId
   $: linkedTable = $tables.list.find(table => table._id === linkedTableId)
   $: fetchRows(linkedTableId)
 
+  const getFieldValue = val => {
+    const linkedIds = (Array.isArray(val) ? val : [])?.map(
+      row => row?._id || row
+    )
+    if (schema.relationshipType === "one-to-many") {
+      return linkedIds?.[0]
```
melohagan, line 29: NAB: `linkedIds` will always be non-null here.

packages/builder/src/components/common/LinkedRowSelector.svelte
```diff
@@ -15,14 +15,25 @@
...
+    } else if (schema.type === "bb_reference_single") {
```
melohagan, line 30: NAB: This deviates from the logic below - just checking this is intended.

I.e. `{:else if schema.relationshipType === "one-to-many" || schema.type === "bb_reference_single"}`

Implies that `linkedIds[0]` should also be used in this `bb_reference_single` case.

melohagan (approved):
LGTM!

packages/bbui/src/Form/Core/Select.svelte
deanhannigan: The inclusion of `false` is just for scenarios where we explicitly want to switch off the placeholder behaviour entirely e.g. an enum. If the placeholder value is nullish, svelte will default to "Choose an option" and the user will be able to unintentionally clear the picker. 

This fix still maintains that behaviour, but in the event that it isn't explicitly turned off (`false`), it still needs a final check to ensure that even when its cleared to an empty string `""`, it must switch to the default string or the placeholder option will not be added.

packages/builder/src/components/common/LinkedRowSelector.svelte
deanhannigan: Niiiiice catch! I cannot remember now why I made this change. It still returns a list of ids and not a single value. I'll get this updated and give it a test.

packages/builder/src/components/common/LinkedRowSelector.svelte
```diff
@@ -15,14 +15,26 @@
...
-  $: linkedIds = (Array.isArray(linkedRows) ? linkedRows : [])?.map(
-    row => row?._id || row
-  )
+  $: fieldValue = getFieldValue(linkedData)
   $: label = label || capitalise(schema.name)
   $: linkedTableId = linkedTableId || schema.tableId
   $: linkedTable = $tables.list.find(table => table._id === linkedTableId)
   $: fetchRows(linkedTableId)
 
+  const getFieldValue = val => {
```
aptkingston, line 24: This consumes `schema` but it's not a parameter - so we might want to update this so the reactive statement also passes it in.

aptkingston (approved):
LGTM! Just one suggested change to ensure reactivity works as expected 👍  Great work on these bugs.

packages/builder/src/components/common/LinkedRowSelector.svelte
deanhannigan: Great catch!

PR #14318, opened 2024-08-05
Added unhandledRejection error handling to koa app

deanhannigan:
## Description

Uncaught promise rejections in the Google Sheets integration would eventually be caught in the `uncaughtException` handler in the koa app. This would then terminate the server.

This fix adds general handling for uncaught rejections in the koa app and logs any events to `bb-alert`

## Addresses
- https://github.com/Budibase/node-google-spreadsheet/pull/2

## Launchcontrol
Add handling for `uncaughtRejection` in the koa app.

### DIFF
--- a/packages/server/src/koa.ts
+++ b/packages/server/src/koa.ts
@@ -69,6 +69,10 @@ export default function createKoaApp() {
     server.destroy()
   }
 
+  process.on("unhandledRejection", err => {
+    logging.logAlert("Unhandled rejection.", err)
+  })
+
   process.on("uncaughtException", err => {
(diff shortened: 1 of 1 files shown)


mike12345567:
Hey @deanhannigan - we actually intentionally don't handle stuff like this, as an unhandled promise like is coming from the googlesheets library is a severe issue we want the process to crash and make us aware something very bad has happened.

We had build protections in place for our backend services to avoid unhandled promises via static analysis but for libraries we don't have this luxury - however when it does happen its best we're made aware as directly as possible.

PR #14325, opened 2024-08-06
Add Oracle to datasource.spec.ts.

samwho:
## Description

Missed this one in the big Oracle PR. Only one change needed to get this to pass, a more strict not-null check. It was sometimes finding nullable columns to be not-null.


packages/server/src/api/routes/tests/datasource.spec.ts
```diff
@@ -285,9 +286,6 @@ describe("/datasources", () => {
           [FieldType.STRING]: {
             name: stringName,
             type: FieldType.STRING,
-            constraints: {
-              presence: true,
-            },
```
samwho, line 290: Ended up removing this because when we create the table, we don't set DB-level not-null checks. I'm uncertain why, but adding this here and then having an exclusion for it later on in the `objectContaining` check seemed weird.

mike12345567 (approved):

PR #14328, opened 2024-08-06
Fix tables showing up multiple times for Oracle in the datasource info endpoint.

samwho:
## Description

Fairly simple fix, the function in the Oracle integration that was getting table names was doing it from a query designed to list out columns. So each table was being listed N times, where N was the number of columns it had. De-duped these using a `Set<string>` and it works as expected.


mike12345567 (approved):
Very nice - good test cases as well !

PR #14331, opened 2024-08-06
In-memory support for logical operators

adrinr:
## Description
Adding support for conditional types for in-memory rows.

## Addresses
- [BUDI-8506 - In memory support for logical operators](https://linear.app/budibase/issue/BUDI-8506/in-memory-support-for-logical-operators)

## Launchcontrol
Allowing conditional filtering for in-memory at the API level

### DIFF
--- a/packages/shared-core/src/filters.ts
+++ b/packages/shared-core/src/filters.ts
@@ -464,14 +464,18 @@ export const runQuery = (docs: Record<string, any>[], query: SearchFilters) => {
     ) =>
     (doc: Record<string, any>) => {
       for (const [key, testValue] of Object.entries(query[type] || {})) {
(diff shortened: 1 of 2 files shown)


samwho (approved):

packages/shared-core/src/filters.ts
```diff
@@ -674,15 +678,34 @@ export const runQuery = (docs: Record<string, any>[], query: SearchFilters) => {
 
   const and = match(
     LogicalOperator.AND,
-    (_docValue: Record<string, any>, _testValue: any) => {
-      // TODO
-      return false
+    (docValue: Record<string, any>, conditions: any) => {
```
mike12345567, line 681: would it be possible to type conditions any better that `any`?

PR #14337, opened 2024-08-07
Allow feature flags to be set via PostHog as well as by environment variables.

samwho:
## Description

This PR introduces PostHog into the path for determining the value of feature flags. This is going to allow us to do gradual and targeted rollouts of new features into Budicloud.

Also in this PR, I've removed the `node-fetch` mock from the `backend-core` and `worker` packages in favour of using `nock`. This followed on from removing the global PostHog mock we were using in `backend-core` to make it easier to mock specific responses from PostHog.

There's no caching of flags in this PR because it appears that the PostHog library does its own caching. I tested it out and flag updates do get picked up very quickly, so I'm not entirely sure how it works. Will need to test in QA using DataDog traces to see how much impact it has on request performance.


packages/backend-core/package.json
```diff
@@ -45,7 +45,7 @@
     "passport-oauth2-refresh": "^2.1.0",
     "pino": "8.11.0",
     "pino-http": "8.3.3",
-    "posthog-node": "1.3.0",
+    "posthog-node": "4.0.1",
```
samwho, line 48: This was required in order to fetch all feature flags at once. `posthog.getAllFeatureFlagsAndPayloads` didn't exist in 1.3.0.

packages/backend-core/src/environment.ts
```diff
@@ -208,6 +209,32 @@ const environment = {
...
+export function withEnv<T>(envVars: Partial<typeof environment>, f: () => T) {
+  const cleanup = setEnv(envVars)
+  const result = f()
+  if (result instanceof Promise) {
+    return result.finally(cleanup)
+  } else {
+    cleanup()
+    return result
+  }
+}
```
samwho, line 236: This is only used in tests, and exists to the mirror the functionality we have in `TestConfiguration` in `packages/server`. I'm not a big fan of it lying around like this but wasn't sure where else to put it.

packages/backend-core/src/events/processors/posthog/tests/PosthogProcessor.spec.ts
samwho, line 1: All of the changes in this file were made as a result of removing the global PostHog mock.

packages/backend-core/src/plugin/tests/validation.spec.ts
```diff
@@ -1,12 +1,129 @@
...
+import nock from "nock"
 
-const repoUrl =
-  "https://raw.githubusercontent.com/Budibase/budibase-skeleton/master"
-const automationLink = `${repoUrl}/automation/schema.json.hbs`
-const componentLink = `${repoUrl}/component/schema.json.hbs`
-const datasourceLink = `${repoUrl}/datasource/schema.json.hbs`
+const automationLink = `http://example.com/automation/schema.json`
+const componentLink = `http://example.com/component/schema.json`
+const datasourceLink = `http://example.com/datasource/schema.json`
```
samwho, line 8: This was previously calling out to GitHub every time we ran these tests.

packages/backend-core/src/redis/redis.ts
```diff
@@ -111,6 +111,10 @@ function init(selectDb = DEFAULT_SELECT_DB) {
   CLIENTS[selectDb] = client
 }
 
+export function closeAll() {
+  Object.values(CLIENTS).forEach(client => client.disconnect())
+}
```
samwho, line 116: I've added this because `redis.spec.ts` was giving me an error when the tests finished running saying there were open connections that could prevent Jest from finishing.

packages/worker/src/api/routes/global/tests/auth.spec.ts
```diff
@@ -292,9 +292,9 @@ describe("/api/global/auth", () => {
       it("redirects to auth provider", async () => {
         nock("http://someconfigurl").get("/").times(1).reply(200, {
           issuer: "test",
-          authorization_endpoint: "http://localhost/auth",
-          token_endpoint: "http://localhost/token",
-          userinfo_endpoint: "http://localhost/userinfo",
+          authorization_endpoint: "http://example.com/auth",
```
samwho, line 295: Using `localhost` here now that `nock` is in use made the tests fail because we allow legit connections to `localhost` in order to make requests to containers, so I changed it to `example.com`.

packages/worker/src/api/routes/global/tests/realEmail.spec.ts
```diff
@@ -30,19 +57,8 @@ describe("/api/global/email", () => {
...
-                status: 301,
-                errno: "ETIME",
-              }),
-            20000
-          )
-        )
-      await Promise.race([config.saveEtherealSmtpConfig(), timeout()])
-      await Promise.race([config.saveSettingsConfig(), timeout()])
+      await withTimeout(20000, config.saveEtherealSmtpConfig())
+      await withTimeout(20000, config.saveSettingsConfig())
```
samwho, line 61: These were changed because in the previous implementation we were leaking `setTimeout` objects and Jest was getting upset about it.

packages/backend-core/src/environment.ts
adrinr: Can we then remove the ones from `server` and to use only these ones?

packages/backend-core/src/events/processors/posthog/tests/PosthogProcessor.spec.ts
```diff
@@ -73,6 +74,8 @@ describe("PosthogProcessor", () => {
     describe("rate limiting", () => {
       it("sends daily event once in same day", async () => {
         const processor = new PosthogProcessor("test")
+        const spy = jest.spyOn(processor.posthog, "capture")
```
adrinr, line 77: If we are using this spy in every test, should we move outside and use a single spy?

packages/backend-core/src/features/index.ts
```diff
@@ -1,24 +1,91 @@
...
 import { cloneDeep } from "lodash"
+import { PostHog, PostHogOptions } from "posthog-node"
+import { IdentityType } from "@budibase/types"
+import tracer from "dd-trace"
+
+let posthog: PostHog | undefined
+export function init(opts?: PostHogOptions) {
+  if (env.POSTHOG_TOKEN) {
+    posthog = new PostHog(env.POSTHOG_TOKEN, {
+      host: "https://us.i.posthog.com",
```
adrinr, line 12: We should have this string as env (maybe with a default value), just in case something changes from their end

packages/backend-core/src/features/index.ts
```diff
@@ -1,24 +1,91 @@
...
+      ...opts,
+    })
+  }
+}
+
+abstract class Flag<T> {
```
adrinr, line 18: Clean and nice strategy!

packages/backend-core/src/features/index.ts
```diff
@@ -1,24 +1,91 @@
...
-  USER_GROUPS: Flag.withDefault(false),
-  ONBOARDING_TOUR: Flag.withDefault(false),
+  LICENSING: Flag.boolean(false),
+  GOOGLE_SHEETS: Flag.boolean(false),
+  USER_GROUPS: Flag.boolean(false),
+  ONBOARDING_TOUR: Flag.boolean(false),
+
+  _TEST_BOOLEAN: Flag.boolean(false),
+  _TEST_STRING: Flag.string("default value"),
+  _TEST_NUMBER: Flag.number(0),
```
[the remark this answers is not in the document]
adrinr: Could these test values not be injected via jest.mocks?

packages/backend-core/src/features/index.ts
```diff
@@ -53,38 +120,84 @@ function isFlagName(name: string): name is keyof Flags {
...
+
+        if (!isFlagName(feature)) {
+          throw new Error(`Feature: ${feature} is not an allowed option`)
+        }
+
+        if (typeof flags[feature] !== "boolean") {
+          throw new Error(`Feature: ${feature} is not a boolean`)
+        }
+
+        // @ts-ignore
```
adrinr, line 155: Why this ignore?

packages/backend-core/src/features/index.ts
```diff
@@ -53,38 +120,84 @@ function isFlagName(name: string): name is keyof Flags {
...
+    for (const [tenantId, ...features] of split) {
+      if (!tenantId || (tenantId !== "*" && tenantId !== currentTenantId)) {
+        continue
       }
 
-      if (!isFlagName(feature)) {
-        throw new Error(`Feature: ${feature} is not an allowed option`)
+      for (let feature of features) {
+        let value = true
+        if (feature.startsWith("!")) {
```
adrinr, line 141: Do we support negative flags? What is the use case of this?

packages/backend-core/src/plugin/tests/validation.spec.ts
adrinr: We will drop our traffic, now :(

packages/backend-core/src/redis/redis.ts
adrinr: I think we have been hunting this one for a while... 🤞🏽

packages/backend-core/src/features/index.ts
samwho: We do. I'm not sure, but it's a thing that existed before this PR and I'm maintaining the behaviour.

packages/backend-core/src/environment.ts
samwho: Good idea, pushed a commit doing this!

packages/backend-core/src/features/index.ts
samwho: CleanShot 2024-08-09 at 10 34 40@2x

If you can help me figure out why this happens I would be very grateful.

packages/backend-core/src/features/index.ts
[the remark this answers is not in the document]
samwho: Went a slightly different route. Created a `FlagSet` class where you can have different sets of flags, exported the production set from index.ts and created a testing set for the tests.

packages/backend-core/src/features/index.ts
adrinr: Mm.. ts generic typing bullshit. There are ways to solve this, but not much more elegant.
One thing it can be easily done is `flags = { ...flags, [feature]: value }`

packages/backend-core/src/features/index.ts
adrinr: Interesting 😅

adrinr (approved):

packages/backend-core/src/features/index.ts
samwho: I'm reticent to add a copy just to please the type system. I feel confident enough in the tests to keep the ignore in.

Issue #14340, opened 2024-08-08
Can no longer copy values outside of an app into the table component.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.29.29
  - App Version: 2.29.29

**Describe the bug**
You can no longer copy values outside of Budibase into the rows of the table components. This is likely related to the changes done to bulk duplicating, copying and deleting rows.

**To Reproduce**
1. Create a new app with sample data
2. Copy a value within your note pad and try and paste this into one of the cells in the data area
4. See issue

**Expected behavior**
Should be able to copy across values from areas outside of Budibase into the table component.

Issue #14288, opened 2024-07-31
Unable to use Conditions to Update Settings >> Details on Multi Step Form Block


mikesealey:
**Checklist**
- `[Y]` I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
Observed on Self (Docker single image) and Cloud (2.29.26) miketest1.budibase.app

**Describe the bug**
_Bug raised on behalf of customer_
Unable to use conditions to update settings on Multi Step Form Block - "Error updating component prop".

**To Reproduce**
1. Add a Multi Step Form Block
2. Add a Condition to update setting `Details`
3. Make a change to fields shown or title or description - UI changes, error appears message appears.

**Expected behavior**
Should be able to configure settings with "Update Setting > Details" settings

**Screenshots**
JAM https://jam.dev/c/ea1d4e87-7487-47f7-b017-290415a16741
  
**App Export**
[Geis MultiStep Form BLock-export-1722423403250.tar.gz](https://github.com/user-attachments/files/16440280/Geis.MultiStep.Form.BLock-export-1722423403250.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 126.0.6478.127 (Official Build) (arm64)

**Additional context**

aptkingston:
Just chipping in to say that I think it's a lot more likely we remove conditional UI for monster settings like "Details" in the multi-step form block rather than make this to work... Rendering that entire section inside the drawer looks terrible. If this customer does need conditional multi-step form block field settings then I think having multiple form blocks which are shown/hidden using conditional UI is probably the way to go.
But I'll take a look in case there's an easy fix anyway.

shogunpurple:
@aptkingston did you get a chance to look at this? Might be better us removing it as you say for new applications and keeping legacy support around

aptkingston:
Just a heads up that when this fix is shipped, this customer will have to delete their condition and recreate it in order to benefit from the fix.

PR #14342, opened 2024-08-08
Fixes + table native paste

aptkingston:
## Description
- Fixes an issue where certain settings would break when used inside conditional UI (https://linear.app/budibase/issue/BUDI-8525/unable-to-use-conditions-to-update-settings-details-on-multi-step-form, https://github.com/Budibase/budibase/issues/14288)
- Allows native paste from clipboard into table cells. This only works if you actually focus the cell, as in you've double clicked into it and are editing the text. If you simply select the cell once but don't start editing it then this won't work, as the browser native paste event can't do anything here (https://github.com/Budibase/budibase/issues/14340)
- Changes how actions are discovered so that they understand global bindings and respect local scope (https://linear.app/budibase/issue/BUDI-8402/clear-form-action-can-select-a-form-that-is-inside-a-repeater-but, https://github.com/Budibase/budibase/issues/14032)


packages/builder/src/dataBinding.js
```diff
@@ -286,44 +282,39 @@ export const getComponentContexts = (
...
-    const action = actions.find(x => x.type === actionType)
-    if (action) {
-      let runtimeBinding = component._id
-      if (action.suffix) {
-        runtimeBinding += `-${action.suffix}`
-      }
-      providers.push({
-        readableBinding: component._instanceName,
-        runtimeBinding,
+        action: context.contexts[0].actions.find(x => x.type === actionType),
```
melohagan, line 302: NAB: How do you know that you can always use the first context?

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ConditionalUIDrawer.svelte
```diff
@@ -147,6 +147,15 @@
       onOperatorChange(condition, condition.operator)
     }
   }
+
+  const onSettingChange = (e, condition) => {
+    const setting = settings.find(x => x.key === e.detail)
+    if (setting?.defaultValue != null) {
+      condition.settingValue = setting?.defaultValue
```
Comment (melohagan) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ConditionalUIDrawer.svelte:154]: NAB: can be `setting.defaultValue`

packages/frontend-core/src/components/grid/overlays/KeyboardManager.svelte
```diff
@@ -58,7 +58,7 @@
         case "c":
           return handle(() => dispatch("copy"))
         case "v":
-          return handle(() => dispatch("paste"))
+          return dispatch("paste")
```
melohagan, line 61: NAB: Why is this different to the "copy" case?

melohagan (approved):

packages/frontend-core/src/components/grid/overlays/KeyboardManager.svelte
aptkingston: It's because we want to enable the native browser paste. If we wrap this in `handle` then that will `preventDefault` on the event, meaning it stops the browser native paste from ever working. This was what used to happen before this PR.

Copy is different because we already put the copied value into the clipboard ourselves, so we don't rely on the native copy functionality. Pasting is different as you cannot forcefully read from the clipboard in JS (for security reasons, and at least not without annoying popups), otherwise I would continue to ignore the native paste functionality and just read from the clipboard in JS and handle it ourselves.

packages/builder/src/dataBinding.js
melohagan: Hmm, I'll take your word for it 😅 
If they're was a reasonable way to pass a single object instead of a list it might be worth doing that, but no worries if it's a hassle!

deanhannigan (approved):
LGTM! Ran through the issues and all are resolved 🚀

packages/builder/src/dataBinding.js
aptkingston: I've added in some optional chaining to it wouldn't error even if this structure didn't exist. It's 100% safe without this at the moment, but it's sensible to be more cautious since it could potentially change in future and introduce a bug, so this will prevent that happening 👍

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ConditionalUIDrawer.svelte
Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Component/ConditionalUIDrawer.svelte]: Done 👍

PR #14344, opened 2024-08-08
Auto login the user after password is set

melohagan:
## Description
When setting your password in various screens, most of the time you are taken to the login screen. 
Updated to directly login, and allow enter key to be used.

## Addresses
- https://linear.app/budibase/issue/GRO-458/all-password-related-endpoints-should-auto-login-the-user

## Screenshots
https://github.com/user-attachments/assets/8c07d330-2f80-4321-a2b4-72839fd31d2b

## Launchcontrol
Login after setting password. Support enter key in password forms.

### DIFF
--- a/packages/builder/src/pages/builder/admin/index.svelte
+++ b/packages/builder/src/pages/builder/admin/index.svelte
@@ -21,6 +21,7 @@
   let submitted = false
 
   $: tenantId = $auth.tenantId
+  $: passwordMinLength = $admin.passwordMinLength ?? 12
 
   async function save() {
     form.validate()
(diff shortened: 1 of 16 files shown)


packages/builder/src/pages/builder/admin/index.svelte
```diff
@@ -83,9 +94,13 @@
           validate={() => {
             let fieldError = {}
 
-            fieldError["password"] = !formData.password
-              ? "Please enter a password"
-              : undefined
+            if (!formData.password) {
+              fieldError["password"] = "Please enter a password"
+            } else if (formData.password.length < 12) {
```
mike12345567, line 99: can we remove this and just depend on the backend validation of this - so that dev environment respects it?

melohagan: Updated admin store to expose `passwordMinLength`. Example:

example

Defaults to 12 if the variable is missing.

mike12345567 (approved):
Nice one - thanks for sorting out that dev environment stuff!

PR #14345, opened 2024-08-08
feat: only use requirepass if password's set

calexiou:
## Description
Avoid using the `requirepass` flag if the user is not setting a `REDIS_PASSWORD` if the `REDIS_PASSWORD`'s value is an empty string. 

## Launchcontrol

Disable `--requirepass` on empty or non-existing `REDIS_PASSWORD`

### DIFF
--- a/hosting/single/runner.sh
+++ b/hosting/single/runner.sh
@@ -80,7 +80,7 @@ ln -s ${DATA_DIR}/.env /worker/.env
 # make these directories in runner, incase of mount
 mkdir -p ${DATA_DIR}/minio
 chown -R couchdb:couchdb ${DATA_DIR}/couch
-if [[ -z "${REDIS_PASSWORD}" ]]; then
+if [[ -z "${REDIS_PASSWORD}" || "${REDIS_PASSWORD}" = "" ]]; then
   redis-server --requirepass $REDIS_PASSWORD > /dev/stdout 2>&1 &
 else
   redis-server > /dev/stdout 2>&1 &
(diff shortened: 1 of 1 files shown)


hosting/single/runner.sh
```diff
@@ -80,7 +80,11 @@ ln -s ${DATA_DIR}/.env /worker/.env
 # make these directories in runner, incase of mount
 mkdir -p ${DATA_DIR}/minio
 chown -R couchdb:couchdb ${DATA_DIR}/couch
-redis-server --requirepass $REDIS_PASSWORD > /dev/stdout 2>&1 &
+if [[ -z "${REDIS_PASSWORD}" || "${REDIS_PASSWORD}" = "" ]]; then
```
melohagan, line 83: I think this should be `if [[ -n "${REDIS_PASSWORD}" ]]; then`

If what you want to do is only pass `--requirepass` if REDIS_PASSWORD is non-null or non-empty.

Not a bash wizard though so might be wrong here!

melohagan (changes requested):

hosting/single/runner.sh
calexiou: I believe we need a hybrid. Both to check if the `REDIS_PASSWORD` is unset and if it's non-emtpy. My original version was wrong indeed - I wanted to check for `!=` forgot the negation.

See: https://github.com/Budibase/budibase/pull/14345/commits/6325ff3593dfef6ec6da5f3da8835edfb0b5dff1

calexiou: Yup, you're right. I've forgotten the difference between `-z` and `-n`. The latter is the one we want!

shogunpurple (approved):
LGTM

melohagan (approved):
LGTM!

PR #14347, opened 2024-08-08
Fix: Handle setting of env vars in .env gracefully

calexiou:
## Description
Use indirect expansion of the `ENV_VAR` instead of assigning value to temporary variable `temp`. If `REDIS_PASSWORD` is set to any value (including an empty string), it won't be overwritten. Only truly unset variables will be assigned a random UUID.

## Screenshots

https://github.com/user-attachments/assets/94dcfa8d-d96b-41ba-af7d-b30a01c90fe8

## Launchcontrol

 Make sure that environment variables from `ENV_VARS` array are respected, whether set to a specific value or left empty.


hosting/single/runner.sh
```diff
@@ -45,8 +45,7 @@ fi
 # randomise any unset environment variables
 for ENV_VAR in "${ENV_VARS[@]}"
 do
-    temp=$(eval "echo \$$ENV_VAR")
-    if [[ -z "${temp}" ]]; then
+    if [[ -z "${!ENV_VAR}" ]]; then
```
shogunpurple, line 48: Are we sure this doesn't break backwards compat with existing single image users?

calexiou: Yes, if a user was setting any variable that the image supported, the existing solution didn't respect the value, thus randomly generating one for them and storing it in a file on disk so that it would remain persistent during container restarts.

What this should do is start respecting whatever they were adding as variables. If they weren't using them and defaulting to randomly generated ones, their values would be on disk, and if they're not, they'll still be randomly generated.

TLDR: the change ensures that [what we claim](https://docs.budibase.com/docs/docker#method-1---docker-run) in the documentation holds.

shogunpurple (approved):
LGTM

PR #14353, opened 2024-08-09
Make it possible to get flag values from the license.

samwho:
## Description

This PR further extends that feature flags capabilities in `backend-core` to be able to source flags from the user's license.

This PR also includes a refactoring of the tests to make it easier to specify new cases.

### DIFF
--- a/packages/backend-core/src/features/index.ts
+++ b/packages/backend-core/src/features/index.ts
@@ -83,22 +83,18 @@ class NumberFlag extends Flag<number> {
 }
 
 export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
-  private readonly flags: T
-
-  constructor(flags: T) {
-    this.flags = flags
-  }
+  constructor(private readonly flagSchema: T) {}
 
   defaults(): FlagValues<T> {
-    return Object.keys(this.flags).reduce((acc, key) => {
+    return Object.keys(this.flagSchema).reduce((acc, key) => {
       const typedKey = key as keyof T
-      acc[typedKey] = this.flags[key].defaultValue
+      acc[typedKey] = this.flagSchema[key].defaultValue
       return acc
     }, {} as FlagValues<T>)
   }
 
   isFlagName(name: string | number | symbol): name is keyof T {
-    return this.flags[name as keyof T] !== undefined
+    return this.flagSchema[name as keyof T] !== undefined
   }
 
   async get<K extends keyof T>(
@@ -120,9 +116,7 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
   async fetch(ctx?: UserCtx): Promise<FlagValues<T>> {
     return await tracer.trace("features.fetch", async span => {
       const tags: Record<string, any> = {}
-
-      const flags = this.defaults()
-
+      const flagValues = this.defaults()
       const currentTenantId = context.getTenantId()
       const specificallySetFalse = new Set<string>()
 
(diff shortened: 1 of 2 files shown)


packages/backend-core/src/features/index.ts
```diff
@@ -150,6 +156,26 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
...
+            continue
+          }
+
+          if (flags[feature] === true || specificallySetFalse.has(feature)) {
+            // If the flag is already set to through environment variables, we
+            // don't want to override it back to false here.
+            continue
+          }
+
+          // @ts-ignore
```
melohagan, line 173: NAB: linters prefer @ts-expect-error, ideally with a comment or ticket link for addressing in future.

packages/backend-core/src/features/index.ts
```diff
@@ -150,6 +156,26 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
...
 
+      const license = ctx?.user?.license
+      if (license) {
+        for (const feature of license.features) {
+          const flag = this.flags[feature]
+          if (!flag) {
+            continue
+          }
+
+          if (flags[feature] === true || specificallySetFalse.has(feature)) {
```
melohagan, line 167: NAB: I don't understand the variable `specificallySetFalse` . What is that?

packages/backend-core/src/features/index.ts
```diff
@@ -150,6 +156,26 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
         }
       }
 
+      const license = ctx?.user?.license
+      if (license) {
+        for (const feature of license.features) {
+          const flag = this.flags[feature]
```
melohagan, line 162: NAB: As far as I can see, this `flag` var is not used anywhere else, therefore it could be simplified to indicate that it's not used anywhere else:
`if (!this.flags[feature]) continue`

Another point is that we have a readonly property called `flags` as well as the locally scoped `flags` variable. 
Maybe one could be named differently to make the code a little easier to read?

packages/backend-core/src/features/tests/features.spec.ts
```diff
@@ -12,207 +12,214 @@ const schema = {
...
       expected: flags.defaults(),
     },
   ])(
-    'should find flags $expected for $tenant with string "$flags"',
-    ({ tenant, flags: envFlags, expected }) =>
-      context.doInTenant(tenant, async () =>
-        withEnv({ TENANT_FEATURE_FLAGS: envFlags }, async () => {
-          const values = await flags.fetch()
-          expect(values).toMatchObject(expected)
+    "$it",
```
melohagan, line 139: Awesome tests! 😄

melohagan (approved):
LGTM!

packages/backend-core/src/features/index.ts
samwho: Good call, have updated all instances of this. It's all the same error, which I'm not sure there's a good workaround for sadly.

packages/backend-core/src/features/index.ts
samwho: When an environment variable specifies a _negative_ feature flag, e.g. `TENANT_FEATURE_FLAGS=tenant1:!DEFAULT_VALUES` they're wanting to disable `DEFAULT_VALUES`. This `specificallySetFalse` set tracks those flags and makes sure that later steps in the flag fetching process don't set them back to true.

It's up for debate whether this is the correct behaviour or not, but I went with it because this negative flag behaviour already exists and might be in use by self-hosters. I think a flag suddenly becoming enabled when they're set it false on purpose could be a bad experience.

packages/backend-core/src/features/index.ts
samwho: I've updated the names of a few variables to better reflect their purpose. I definitely went overboard with `flag`/`flags` 😅

packages/backend-core/src/features/index.ts
melohagan: Makes sense!

melohagan (approved):

PR #14362, opened 2024-08-12
Make generated passwords longer

melohagan:
## Description
Bug fix were backend validation was being hit because the generated passwords weren't long enough.

## Screenshots
longer password

## Launchcontrol
Minor bug fix relating to auto-generated passwords.

### DIFF
--- a/packages/builder/src/pages/builder/portal/users/users/_components/AddUserModal.svelte
+++ b/packages/builder/src/pages/builder/portal/users/users/_components/AddUserModal.svelte
@@ -16,7 +16,7 @@
 
   export let showOnboardingTypeModal
 
-  const password = generatePassword()
+  const password = generatePassword(12)
   let disabled
   let userGroups = []
 
@@ -44,7 +44,7 @@
       {
         email: "",
         role: "appUser",
-        password: generatePassword(),
+        password: generatePassword(12),
         forceResetPassword: true,
         error: null,
       },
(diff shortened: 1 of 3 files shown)


packages/builder/src/pages/builder/portal/users/users/_components/AddUserModal.svelte
```diff
@@ -69,6 +69,14 @@
     return userData[index].error == null
   }
 
+  // Minimum password length is 12 characters, but Math random will generate at most 11 characters, so add another 5.
+  function generatePassword() {
+    return (
+      Math.random().toString(36).substring(2, 22) +
+      Math.random().toString(36).substring(2, 7)
+    )
```
samwho, line 77: This is a lil' janky, and I don't think `Math.random` is guaranteed to be a strong source of randomness.

Could we instead do something like:

```javascript
crypto.randomBytes(15).toString("base64")
```

samwho (changes requested):

packages/builder/src/pages/builder/portal/users/users/_components/AddUserModal.svelte
samwho:
```javascript
function generatePassword(length) {
    const array = new Uint8Array(length);
    window.crypto.getRandomValues(array);
    return Array.from(array, byte => byte.toString(36).padStart(2, '0')).join('').slice(0, length);
}

const password = generatePassword(12);
```

Seems to be the equivalent way to get strong random values in the browser, but it makes me tremendously uneasy we don't do this password generation on the server.

melohagan: Yeah there's really no reason to do that is there. We can just pass up a flag to request an auto-generated password when we create the user

samwho (approved):

PR #14364, opened 2024-08-12
Disable PostHog feature flags in prod.

samwho:
## Description

Until we get the local-evaluation working (which I've set the groundwork for in this PR), I'm disabling PostHog feature flags in production (nothing uses them yet anyway). This is because hitting PostHog takes ~300ms and we'd like to reduce that as much as possible.


shogunpurple (approved):
LGTM

PR #14369, opened 2024-08-13
Add more automation tests

PClmnt:
## Description
Adds more automation tests specifically around multiple loops in an automation (which we have had issues with in the past) and external querying and creating rows. 

Also updated the output type for looping to be correct.

### DIFF
--- a/packages/server/src/automations/tests/scenarios/scenarios.spec.ts
+++ b/packages/server/src/automations/tests/scenarios/scenarios.spec.ts
@@ -8,13 +8,7 @@ import {
   FieldType,
 } from "@budibase/types"
 import { createAutomationBuilder } from "../utilities/AutomationBuilder"
-import {
-  DatabaseName,
-  getDatasource,
-  knexClient,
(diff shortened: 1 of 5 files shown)


packages/server/src/automations/tests/scenarios/scenarios.spec.ts
```diff
@@ -157,4 +236,101 @@ describe("Automation Scenarios", () => {
...
+            type: FieldType.NUMBER,
+            presence: true,
+          },
+        },
+      },
+    })
+
+    let client: Knex
+
+    let db = await getDatasource(DatabaseName.MYSQL)
```
mike12345567, line 266: Might be worth breaking this out into a function that can be re-used through the automation tests, the ability to create an external database with table, with rows, for querying purposes could be very useful if reusable.

mike12345567 (approved):
LGTM - one NAB comment, otherwise these tests are great!

adrinr (approved):

PR #14378, opened 2024-08-14
More automation typing in preparation for branching

PClmnt:
## Description
This PR contains numerous improvements to typing around the automations area. 

- All of the steps now have their own types which are extended from a base schema (Thanks to @adrinr for sorting this by doing the bulk of work here, and basically fixing my poorly done implementation 😄 ) 
- Moved step definitions to a new type as they aren't technically an automation step
- Added a branch type
- Fixed some casting and typing in the automation thread around looping now that we can access the types automatically

- ## Addresses
https://linear.app/budibase/issue/BUDI-8528/update-automation-typing-to-support-branching


packages/server/src/automations/tests/utilities/AutomationBuilder.ts
```diff
@@ -143,14 +194,16 @@ class AutomationBuilder {
...
-    inputs: T
+  private step<TStep extends AutomationActionStepId>(
+    stepId: TStep,
+    stepSchema: Omit<AutomationStep, "id" | "stepId" | "inputs">,
+    inputs: AutomationStepInputs<TStep>
   ): this {
     this.automationConfig.definition.steps.push({
       ...stepSchema,
-      inputs,
+      inputs: inputs as any,
```
samwho, line 204: 👀 what's stopping us getting rid of this `as any`?

packages/server/src/tests/utilities/structures.ts
```diff
@@ -160,7 +162,7 @@ export function automationTrigger(
   return {
     id: utils.newid(),
     ...triggerDefinition,
-  }
+  } as AutomationTrigger
```
samwho, line 165: How come this is required?

packages/server/src/threads/automation.ts
```diff
@@ -331,7 +325,7 @@ class Orchestrator {
                 }
                 try {
                   loopStep.inputs.binding = automationUtils.typecastForLooping(
-                    loopStep.inputs as LoopInput
+                    loopStep.inputs
```
samwho, line 328: ❤️

packages/server/src/automations/tests/utilities/AutomationBuilder.ts
PClmnt: @adrinr Might be able to explain this better

adrinr: `stepSchema` might be anything, so ts will complain if the types are not typed correctly... I fought with it but the only solution I found was having guard rails and similar, adding too much of a boilerplate

packages/server/src/tests/utilities/structures.ts
PClmnt: Because here we're having to create a Trigger from the definition which isn't actually a fully formed AutomationTrigger. Once we get away from having to define the definitions like this and use the builder it will be much better.

adrinr: `triggerDefinition` might be anything, and sadly typing go nuts... I tried my best, but not sure if there is a cleaner way to solve it

adrinr (approved):

PR #14379, opened 2024-08-14
Cache feature flags per-request, set default values flag to false by default.

samwho:
## Description

Local-evaluation of flags works in QA and brings the feature flag retrieval down from 300ms to 5ms, but 5ms is still quite a lot of you're doing a few flag checks per request. Because of this, I'm caching the flags in the request context so that subsequent checks should be near instant.

I also erroneously set the DEFAULT_VALUES flag to true by default, it should be false.

### DIFF
--- a/packages/backend-core/src/features/index.ts
+++ b/packages/backend-core/src/features/index.ts
@@ -130,10 +130,7 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
 
   async fetch(ctx?: UserCtx): Promise<FlagValues<T>> {
     return await tracer.trace("features.fetch", async span => {
-      const requestContext = context.getCurrentContext()
-      const cachedFlags = requestContext?.featureFlagCache?.[this.setId] as
-        | FlagValues<T>
-        | undefined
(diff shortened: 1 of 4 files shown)


packages/backend-core/src/features/index.ts
```diff
@@ -236,6 +252,11 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
         }
       }
 
+      if (requestContext) {
+        requestContext.featureFlagCache ??= {}
+        requestContext.featureFlagCache[this.setId] = flagValues
```
mike12345567, line 257: Normally we would have getters/setters in the context rather than directly utilising the context - makes it a bit easier to use for people unfamiliar with it.

While we do usually do the setting via `doIn...` type functions but something to just set in the context works fine as well, like `context.setFeatureFlags()` and `context.getFeatureFlags()`

mike12345567 (approved):

packages/backend-core/src/features/index.ts
samwho: Addressed 🙂

mike12345567 (approved):

PR #14381, opened 2024-08-14
Use HBS context over helpers when names overlap (formulas)

mike12345567:
## Description
This implements a fix for when helpers names and fields in the context overlap - for example if you have a table with a column named `test` - then it will overlap with the test helper for regex.

This fix means that if a helper name overlaps that in context then the helper will become inaccessible - it has been decided this is better as being able to access the columns in your table easily is far more important than accessing the helpers - this is an edge case scenario but when it occurs can be quite confusing.

To do this we prefix any keys in HBS statements which overlap in context/helpers with `./` - this is explained in the handlebars docs here: https://handlebarsjs.com/guide/expressions.html#disambiguating-helpers-calls-and-property-lookup

## Addresses
- https://linear.app/budibase/issue/BUDI-8488/formula-columns-cannot-reference-other-columns-whose-names-clash-with

### DIFF
--- a/packages/string-templates/src/index.ts
+++ b/packages/string-templates/src/index.ts
@@ -47,7 +47,10 @@ function testObject(object: any) {
   }
 }
 
-function findOverlappingHelpers(context: object) {
+function findOverlappingHelpers(context?: object) {
+  if (!context) {
+    return []
+  }
   const contextKeys = Object.keys(context)
   return contextKeys.filter(key => helperNames.includes(key))
 }
--- a/packages/string-templates/src/index.ts
+++ b/packages/string-templates/src/index.ts
@@ -1,17 +1,18 @@
 import { Context, createContext, runInNewContext } from "vm"
 import { create, TemplateDelegate } from "handlebars"
 import { registerAll, registerMinimum } from "./helpers/index"
-import { preprocess, postprocess } from "./processors"
+import { postprocess, preprocess } from "./processors"
 import {
   atob,
   btoa,
-  isBackendService,
-  FIND_HBS_REGEX,
   FIND_ANY_HBS_REGEX,
+  FIND_HBS_REGEX,
   findDoubleHbsInstances,
+  isBackendService,
+  prefixStrings,
 } from "./utilities"
 import { convertHBSBlock } from "./conversion"
-import { setJSRunner, removeJSRunner } from "./helpers/javascript"
+import { removeJSRunner, setJSRunner } from "./helpers/javascript"
 
 import manifest from "./manifest.json"
 import { ProcessOptions } from "./types"
@@ -23,6 +24,7 @@ export { iifeWrapper } from "./iife"
 
 const hbsInstance = create()
 registerAll(hbsInstance)
+const helperNames = Object.keys(hbsInstance.helpers)
 const hbsInstanceNoHelpers = create()
 registerMinimum(hbsInstanceNoHelpers)
 const defaultOpts: ProcessOptions = {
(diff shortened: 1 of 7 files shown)


packages/string-templates/src/index.ts
```diff
@@ -60,7 +74,25 @@ function createTemplate(string: string, opts?: ProcessOptions) {
...
+    : []
+
+  string = preprocess(string, {
+    ...opts,
+    disabledHelpers: overlappingHelpers,
+  })
+
+  if (context && !opts?.noHelpers) {
+    if (overlappingHelpers.length > 0) {
+      for (let block of findHBSBlocks(string)) {
```
melohagan, line 88: NAB: Should be const instead of let ?

packages/string-templates/src/processors/preprocessor.ts
```diff
@@ -56,7 +56,10 @@ export const processors = [
   }),
   new Preprocessor(
     PreprocessorNames.FINALISE,
-    (statement: string, opts: { noHelpers: any }) => {
+    (
+      statement: string,
+      opts: { noHelpers: any; disabledHelpers?: string[] }
```
melohagan, line 61: NAB: prefer unkown to any type, if genuinely unknown.

packages/string-templates/src/utilities.ts
```diff
@@ -66,3 +66,16 @@ export const btoa = (plainText: string) => {
 export const atob = (base64: string) => {
   return Buffer.from(base64, "base64").toString("utf-8")
 }
+
+export const prefixStrings = (
```
melohagan, line 70: Please add unit tests for this method!

packages/string-templates/test/helpers.spec.ts
```diff
@@ -483,3 +483,26 @@ describe("uuid", () => {
...
   })
 })
+
+describe("helper overlap", () => {
+  it("should use context over helpers (regex test helper)", async () => {
+    const output = await processString("{{ test }}", { test: "a" })
+    expect(output).toEqual("a")
+  })
+
+  it("should use helper if no sum in context, return the context value otherwise", async () => {
```
melohagan, line 493: Do we need a test for when you explicitly want to use the sum helper instead of the sum context var?

melohagan (changes requested):
Looks good, just a couple comments relating to testing.

packages/string-templates/test/helpers.spec.ts
```diff
@@ -483,3 +483,26 @@ describe("uuid", () => {
     expect(output).toMatch(UUID_REGEX)
   })
 })
+
+describe("helper overlap", () => {
```
samwho, line 487: Would be good to have some tests here that pass in `noHelpers` to make sure all those conditionals work correctly.

Also do any of these tests exercise `prefixStrings`? I can't quite tell.

mike12345567: Yeap - these tests specifically have to exercise it - `sum` and `test` are helpers, so for their values to be returned they have to be prefixed.

packages/string-templates/test/helpers.spec.ts
mike12345567: There are test cases for this already - if `sum` is in context then the helper is completely inaccessible, this is expected behaviour and a limitation we have accepted.

packages/string-templates/src/utilities.ts
mike12345567: Generally we don't unit test individual functions anymore as a black box test is preferred - it is better to confirm this works through the usage of it than confirm an individual function works (if we started to go down the path of testing all functions we'd never finish!)

packages/string-templates/src/index.ts
```diff
@@ -60,7 +74,25 @@ function createTemplate(string: string, opts?: ProcessOptions) {
...
+        string = string.replace(
+          block,
+          prefixStrings(block, overlappingHelpers, "./")
```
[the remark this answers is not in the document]
mike12345567: Defined in handlebar documentation here https://handlebarsjs.com/guide/expressions.html#disambiguating-helpers-calls-and-property-lookup (mentioned in PR).

packages/string-templates/src/utilities.ts
melohagan: Fair enough, it was mostly the regex that made me nervous, but if you're happy it's covered in other tests then all good!

melohagan (approved):

samwho (approved):

packages/string-templates/src/utilities.ts
mike12345567: Regex always makes me nervous too - but all its doing is taking a user input, escaping anything which isn't regex friendly, then making a list of strings to find - the JS for this would be quite simple but regex is *a lot* faster and performance in this area is critical!

PR #14386, opened 2024-08-15
Use new feature flag API for SQS.

samwho:
## Description

This PR gets rid of the `SQS_SEARCH_ENABLE` and `SQS_MIGRATION_ENABLE` environment variables in favour of creating an `SQS` feature flag.

`SQS_MIGRATION_ENABLE` has been entirely removed and the function it used to serve (enabling the SQS migration) is no longer needed because the functionality always happens. This is fine because SQS is running in all environments now.

`SQS_SEARCH_ENABLE` checks have been replaced with `features.flags.isEnabled("SQS")`.

Pro PR: https://github.com/Budibase/budibase-pro/pull/313

## Feature branch env
[Feature Branch Link](http://fb-feature-flag-sqs.fb.qa.budibase.net)


shogunpurple (approved):
LGTM! Nice tests too

packages/server/src/tests/jestEnv.ts
```diff
@@ -3,6 +3,7 @@ import { tmpdir } from "os"
 process.env.SELF_HOSTED = "1"
 process.env.NODE_ENV = "jest"
 process.env.MULTI_TENANCY = "1"
+process.env.APP_PORT = "0"
```
mike12345567, line 6: should we do the same for the worker - just to avoid the issue that you ran into?

charts/budibase/templates/app-service-deployment.yaml
```diff
@@ -212,10 +212,6 @@ spec:
         - name: APP_FEATURES
           value: "api"
         {{- end }}
-        {{- if .Values.globals.sqs.enabled }}
```
adrinr, line 215: Don't we need to add the new feature flag env in here?

packages/worker/src/tests/api/environment.ts
```diff
@@ -9,6 +9,7 @@ export class EnvironmentAPI extends TestAPI {
   getEnvironment = () => {
     return this.request
       .get(`/api/system/environment`)
+      .set(this.config.defaultHeaders())
```
mike12345567, line 12: Out of interest - why is this needed now? The environment endpoint usually can be called publicly I believe.

mike12345567 (approved):
LGTM - just a question and NAB :)

packages/worker/src/tests/api/environment.ts
samwho: Ah, without a tenantId it will now fail because the tenantId is needed to check the feature flag. Will this be a problem in practice?

packages/server/src/api/controllers/table/index.ts
```diff
@@ -118,6 +119,8 @@ export async function save(ctx: UserCtx<SaveTableRequest, SaveTableResponse>) {
   ctx.eventEmitter &&
     ctx.eventEmitter.emitTable(`table:save`, appId, { ...savedTable })
   ctx.body = savedTable
+
+  savedTable = await processTable(savedTable)
```
adrinr, line 123: Why is this needed?

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
```diff
@@ -36,16 +36,6 @@ const migration = async () => {
 
   // at the end make sure design doc is ready
   await sdk.tables.sqs.syncDefinition()
-  // only do initial search if environment is using SQS already
```
adrinr, line 39: Why is this not needed? What about updates from a non-sqs version to this one?

packages/server/src/websockets/builder.ts
```diff
@@ -102,10 +101,9 @@ export default class BuilderSocket extends BaseSocket {
   }
 
   emitTableUpdate(ctx: any, table: Table, options?: EmitOptions) {
-    // This was added to make sure that sourceId is always present when
-    // sending this message to clients. Without this, tables without a
-    // sourceId (e.g. ta_users) won't get correctly updated client-side.
-    table = processTable(table)
+    if (table.sourceId == null || table.sourceId === "") {
```
adrinr, line 104: Are we excluding undefineds for a reason?

adrinr (approved):
Looks good, just a couple of comments left!

charts/budibase/templates/app-service-deployment.yaml
samwho: The only users of this so far is us, and we're going to move to using PostHog to flag this on and off, so I decided to just get rid of it. When it comes time for this to be turned on for users, it'll be done by default.

packages/server/src/api/controllers/table/index.ts
samwho: Let me see if I can remember, I know there was a reason I discussed it with @mike12345567. Here's the start of a long discussion about it: https://discord.com/channels/733030666647765003/1022445411098710016/1273586235893612636

tl;dr `processTable` wasn't async, I've had to make it async in this PR. But `processTable` was used in a non-async function in `packages/server/src/websockets/builder.ts`. To avoid having to make all of that code async, I instead lifted `processTable` out of that file and put some checks in there instead.

packages/server/src/appMigrations/migrations/20240604153647_initial_sqs.ts
samwho: All this was doing in practice was kicking off a CouchDB replication into SQLite. It was an optimisation, it isn't a necessary step for SQS to work. I discussed this with @mike12345567 (https://discord.com/channels/733030666647765003/1022445411098710016/1273597160428146689) and we decided to get rid of it. The reason it came up is that these migrations aren't a user-triggered action, so we wouldn't be able to consult PostHog for the status of a flag here. That could lead to a mismatch in flags being disabled here but enabled elsewhere. Because of this it felt like the cleanest option was to remove this and rely on the backfill being triggered by a real search.

It's a bit academic in practice because for most users this code has already been run in Budicloud.

packages/worker/src/tests/api/environment.ts
mike12345567: It would be worth checking if its a problem - if you publish an app with a public screen, then try to go to that screen URL in the app I would suspect you'll get some error messages if the environment endpoint now requires it.

We will probably have to assume feature flags are disabled for public apps/pages until the feature is fully released, since there is no real way to confirm the tenant ID.

packages/server/src/websockets/builder.ts
samwho: This is a little bit on faith. I'm trusting that the comment that was here before is correct, and any missing `sourceId` is a problem. I'm not sure _why_ it's a problem, so I went wide on this conditional.

mike12345567 (approved):
Looks good - thanks for checking the public stuff was working!

PR #14387, opened 2024-08-15
Update automation test builder to support branching

PClmnt:
## Description
This PR updates the automation test builder to support (nested) branching of automations. These don't actually run successfully yet as the implementation is not started within the thread but I was able to verify an automation with the correct inputs (conditions and name) as well as children is being built up as we would expect. 

## Addresses
https://linear.app/budibase/issue/BUDI-8531/automation-branching-test-automation-builder-improvements


packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -43,69 +46,55 @@ type TriggerOutputs =
...
-  }
-  private config: TestConfiguration = setup.getConfig()
-  private triggerOutputs: TriggerOutputs
-  private triggerSet: boolean = false
-
-  constructor(options: { name?: string } = {}) {
-    this.automationConfig.name = options.name || `Test Automation ${uuidv4()}`
+type BranchConfig = {
+  [key: string]: {
+    steps: (stepBuilder: StepBuilder) => StepBuilder
```
samwho, line 51:
```suggestion
    steps: (stepBuilder: StepBuilder) => void
```

Because you don't use the return value of this function, I'd make it void.

samwho (approved):

packages/server/src/automations/steps/branch.ts
```diff
@@ -0,0 +1,47 @@
...
+        },
+        children: {
+          type: AutomationIOType.ARRAY,
+        },
+      },
+      required: ["conditions"],
+    },
+    outputs: {
+      properties: {
+        output: {
```
mike12345567, line 39: I think the branch step should output which branch it decided to go down, so the branch name and the result of the condition, similar to the Filter step today - I don't think there is must else it can output.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -43,69 +46,55 @@ type TriggerOutputs =
...
-  ) {
-    this.triggerOutputs = outputs
-    return this.trigger(
-      TRIGGER_DEFINITIONS.ROW_DELETED,
-      AutomationTriggerStepId.ROW_DELETED,
-      inputs,
-      outputs
-    )
-  }
+  branch(branchConfig: BranchConfig): this {
```
mike12345567, line 73: I think the branch step should return a more limited set, rather than `this` - since you can't add steps after the branch we can just return the `run` function.

mike12345567 (approved):
LGTM - some minor comments!

PR #14389, opened 2024-08-15
Fix for feature flag issue

mike12345567:
## Description
Fixing an issue where old feature flags would cause issues, all feature flags are expected to be known, this allows unknown feature flags to just be ignored.

### DIFF
--- a/packages/backend-core/src/features/index.ts
+++ b/packages/backend-core/src/features/index.ts
@@ -159,11 +159,11 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
             specificallySetFalse.add(feature)
           }
 
-          if (!this.isFlagName(feature)) {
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/features/index.ts
```diff
@@ -159,11 +159,11 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
             specificallySetFalse.add(feature)
           }
 
-          if (!this.isFlagName(feature)) {
-            throw new Error(`Feature: ${feature} is not an allowed option`)
```
samwho, line 163: I'd just change this one line to be `continue` instead of combining with the next conditional, think it makes the intent a little bit clearer.

Also the way you've done it here I think will push junk flags into flagValues and it's not been caught because that line has a `@ts-expect-error` above it.

samwho (approved):

PR #14394, opened 2024-08-16
Search logical operator fixes

mike12345567:
## Description
This includes some small fixes for improvements to filters - namely around making sure that cleanup/prefixing of filters is working as expected. Previously filters were a flat structure but now recursing over the structure is required (when `$and` or `$or` keys are found).

Also includes a small temporary fix for filtering - makes sure that filters get applied as expected when working with relationships - previously and filters wouldn't work. This will eventually be updated with a new mechanism for pagination which removes the need for the sub-query.


packages/server/src/api/controllers/row/views.ts
```diff
@@ -58,8 +58,9 @@ export async function searchView(
         })
       })
     } else {
+      const operator = query.allOr ? "$or" : "$and"
       query = {
-        $and: {
+        [operator]: {
```
adrinr, line 63: This should always be an $and. We need to get sure we AND the view filters with the user filters. Otherwise a user can bypass view filters

packages/backend-core/src/sql/sql.ts
```diff
@@ -371,10 +373,11 @@ class InternalBuilder {
             ),
             castedTypeValue.values
           )
-        } else if (!opts?.relationship && !isRelationshipField) {
+        } else if (!isRelationshipField) {
           const alias = getTableAlias(tableName)
           fn(alias ? `${alias}.${updatedKey}` : updatedKey, value)
-        } else if (opts?.relationship && isRelationshipField) {
+        }
+        if (opts?.relationship && isRelationshipField) {
```
adrinr, line 380: Why removing this else?

adrinr (changes requested):

packages/backend-core/src/sql/sql.ts
mike12345567: This is somewhat of a hack for now - basically if there are `$or` relationship filters we need to have both filters below here - it depends on the logic, but previously we only had the filters for relationships here, then we did the filters for the main table in the sub-query.

This is a short term fix as the changes in https://github.com/Budibase/budibase/pull/14410 completely remove the need for this - technically it could be removed since once the new pagination mechanism is merged is completely fixes this problem.

packages/server/src/api/controllers/row/views.ts
mike12345567: Discussed on call - removing this change.

adrinr (approved):
Thx for looking at it!

PR #14405, opened 2024-08-19
auto generate encryption key, and use secrets

shogunpurple:
## Description
Update for enterprise prospect - the `API_ENCRYPTION_KEY` (`Values.apiEncryptionKey`) and `ENCRYPTION_KEY` (`Values.bbEncryptionKey`) need to be specified in `values.yaml`.

The K8S docs erroneously state `You don't need to set this if `createSecrets` is true.` I've updated this to be the case.

## Launchcontrol
Update helm chart to automatically generate secrets for encryption keys automatically as well as use encryption keys from K8S secrets


calexiou (approved):
LGTM

PR #14406, opened 2024-08-19
Split columns setting button/content

adrinr:
## Description
Refactoring the ColumnsSettingButton between the button and content. The content will be reused for view joins, so this PR is just splitting the code to make the next PRs easier to review. This PR breaks a component into two.
Along with these changes, we are also fixing a bug where new columns were shown as non-visible on the column picker, but they are visible by default from the API.

## Launchcontrol
Fix default visibility for columns in the ColumnsSettings component


mike12345567 (approved):
LGTM!

PR #14411, opened 2024-08-19
Poc authenticated iframe

shogunpurple:
## Description
PoC for an enterprise prospect.

Adds the ability to use data from the enclosing iframe as a binding in Budibase. This will enable "Authenticated iframe" type use cases where an enclosing application needs to pass down information like tokens to the embedded Budibase app, and Budibase can use them to make calls to APIs for example. It's done through the introduction of a new Binding that sits under an `Embed` category. 

### Prereqs 
- Your app needs to be public (required anyway for embeds)
- You must be running Budibase and your enclosing web app on the same domain. (Different subdomains are fine, just the TLD needs to be the same) This is not a strict requirement because it uses the `postMessage` API underneath, but I wanted to keep things secure by enforcing the same origin policy.

### Feedback
Raised this as a draft for now as there's a few things I'd like feedback on:
- The name of the binding (`ParentWindow`) - should this be a less technical term, or will users who use it need to be aware of the technical aspect anyway
- The icon for the embed binding (see video) - just grabbed one from spectrum. Open to others

## Screenshots
This demo shows how the binding works - basically it's just `{{ ParentWindow.whateverKey }}` and you can then post messages from the enclosing application and your budibase app will reactively update when a message is received.
https://jam.dev/c/1d267676-9b44-405e-bf4e-bec76a70bc8b

## Launchcontrol
- Allow information to be passed down to an embedded Budibase application from the app that renders it. Useful for tokens/auth


packages/builder/src/dataBinding.js
```diff
@@ -813,6 +815,26 @@ export const getActionBindings = (actions, actionId) => {
...
+  const safeEmbed = makePropSafe("embed")
+
+  bindings = [
+    {
+      type: "context",
+      runtimeBinding: `${safeEmbed}`,
+      readableBinding: `ParentWindow`,
+      category: "Embed",
+      icon: "DistributeVertically",
+      display: { type: "object", name: "Parent Window" },
```
aptkingston, line 832: NAB: The display name is optional here and overrides the readable binding, so we may as well remove it and just let it use the readable binding (which I'd also put a space in). Icon looks grand to me.

packages/client/src/components/context/EmbedProvider.svelte
```diff
@@ -0,0 +1,41 @@
...
+  let data = {}
+
+  export function extractDomainFromUrl(url) {
+    const { hostname } = new URL(url)
+    const parts = hostname.split('.');
+    const tld = parts.slice(-2).join(".")
+    return tld
+  }
+
+  export function handleMessage(event) {
```
aptkingston, line 14: NAB: we don't need to export these 2 functions since they're never consumed outside the component.

packages/client/src/components/context/EmbedProvider.svelte
```diff
@@ -0,0 +1,41 @@
...
+    // Validate the event origin to ensure it's coming from a trusted source
+    // Allow different subdomains but must match TLD
+    const appOrigin = extractDomainFromUrl(window.location.origin)
+    const eventOrigin = extractDomainFromUrl(event.origin)
+
+    if (appOrigin !== eventOrigin) {
```
aptkingston, line 20: I think this logic is inverted - it should be `===` here.

aptkingston (commented):
Looks good, just one small logic change needed.

My other thought on this is that postMessage is a common API and we also use it internally for passing data between the builder and app preview. That means that this handler picks up the events every time you make a change in the builder, or even hover your mouse.

I think there are 2 options to get around this:
- Explicitly filter out our own BB events by checking if the message payload has a `isBudibaseEvent` property set to `true` (which we set on our events because we had issues with browser plugins and other things firing `message` events and causing proplems)
- Add a specific structure to the message payload for this use case, and only process those events which match. I think this is my preferred approach, which also avoids the issue of accidentally intercepting message events fired for any other reason. For this, we could go for something like
```js
{
  type: "bb-parent-window-context",
  data: { // something }
}
```

packages/client/src/components/context/EmbedProvider.svelte
shogunpurple: Ah, good catch. Was messing around for testing

packages/builder/src/dataBinding.js
shogunpurple: Removed display and just kept `ParentWindow` as one because the binding looks cleaner

aptkingston (approved):

PR #14413, opened 2024-08-20
SQS - broken relationship fix

mike12345567:
## Description
Fixing an issue where searching on a relationship column which is in a broken state would break in SQS (previously this was allowed and would return empty relationships).


shogunpurple (approved):
LGTM - would be good to have a test that proves it out but can be a follow up

PR #14418, opened 2024-08-20
Invalid primary display setting (no relationship columns)

mike12345567:
## Description
Adding a test case for primary display columns, ignore when it has been set to a relationship, instead use another column which is valid in the table.

This can cause a circular structure to occur as the primary display can be a relationship pointing back at the original table, meaning the full row gets pulled in, causing a loop which breaks things.

### DIFF
--- a/packages/server/src/db/linkedRows/index.ts
+++ b/packages/server/src/db/linkedRows/index.ts
@@ -229,12 +229,12 @@ function getPrimaryDisplayValue(row: Row, table?: Table) {
     invalid = INVALID_DISPLAY_COLUMN_TYPE.includes(primaryDisplaySchema.type)
   }
   if (invalid || !primaryDisplay) {
-    const validKey = Object.keys(table?.schema || {}).filter(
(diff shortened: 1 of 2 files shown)


packages/server/src/db/linkedRows/index.ts
```diff
@@ -206,6 +215,34 @@ export async function attachFullLinkedDocs(
...
+  const primaryDisplay = table?.primaryDisplay
+  let invalid = true
+  if (primaryDisplay) {
+    const primaryDisplaySchema = table?.schema[primaryDisplay]
+    invalid =
+      INVALID_DISPLAY_COLUMN_TYPE.includes(primaryDisplaySchema.type) &&
+      row[primaryDisplay]
+  }
+  if (invalid || !primaryDisplay) {
+    const validKey = Object.keys(table?.schema || {}).filter(
```
adrinr, line 232: This could be a find instead of a filter. You are using the first, anyways

adrinr (approved):
One minor comment, but looks good!

PR #14422, opened 2024-08-21
Fix saving huge amounts of unnecessary data in settings

aptkingston:
## Description
This PR fixes an issue which was introduced a few months ago and resulted in us saving the entire datasource definition of a given datasource when you chose a table from that datasource in a component setting.

The feature that introduced this was intended to provide additional information in datasource settings by showing the datasource name, but it was implemented in a way that introduced this issue.

It's important we fix this as not only does it save massive amounts of bloated data which isn't needed, but it's also blocking the upcoming "connected screens" feature.


deanhannigan (approved):
LGTM!

PR #14423, opened 2024-08-21
Fix query conditions on views

adrinr:
## Description
Adding tests around view user queries and fixing a couple of issues found around queries

## Launchcontrol
Add tests and fix view queries


packages/backend-core/src/sql/sql.ts
```diff
@@ -339,7 +339,7 @@ class InternalBuilder {
     if (!filters) {
       return query
     }
-    filters = this.parseFilters(filters)
+    filters = this.parseFilters({ ...filters })
```
mike12345567, line 342: Why is this necessary?

mike12345567 (approved):
LGTM - just a question!

PR #14425, opened 2024-08-21
View UI ability to pick relationships on columns

adrinr:
## Description
Frontend part of new relationship picker. It will be only one level down (no nested relationships)

### In tables
https://github.com/user-attachments/assets/4d1654f4-da35-463e-b35c-fefae04d69ba

### In views
https://github.com/user-attachments/assets/e0fd6886-7d36-45d6-be00-32a53b041c88

### Link and formula fields are not display and field types icons are used
image

#### Relation display field
image
image

#### Required fields
image

image

### DIFF
--- a/packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
+++ b/packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
@@ -2,16 +2,29 @@
   import { getContext } from "svelte"
   import { ActionButton, Popover } from "@budibase/bbui"
   import ColumnsSettingContent from "./ColumnsSettingContent.svelte"
+  import { FieldPermissions } from "../../../constants"
 
   export let allowViewReadonlyColumns = false
 
-  const { columns } = getContext("grid")
+  const { columns, datasource } = getContext("grid")
 
   let open = false
   let anchor
 
   $: anyRestricted = $columns.filter(col => !col.visible || col.readonly).length
   $: text = anyRestricted ? `Columns (${anyRestricted} restricted)` : "Columns"
+
+  $: permissions =
+    $datasource.type === "viewV2"
+      ? [
+          FieldPermissions.WRITABLE,
+          FieldPermissions.READONLY,
+          FieldPermissions.HIDDEN,
+        ]
+      : [FieldPermissions.WRITABLE, FieldPermissions.HIDDEN]
+  $: disabledPermissions = allowViewReadonlyColumns
+    ? []
+    : [FieldPermissions.READONLY]
 </script>
 
 <div bind:this={anchor}>
@@ -28,5 +41,9 @@
 </div>
 
 <Popover bind:open {anchor} align="left">
-  <ColumnsSettingContent columns={$columns} {allowViewReadonlyColumns} />
+  <ColumnsSettingContent
+    columns={$columns}
+    {permissions}
+    {disabledPermissions}
+  />
 </Popover>
--- a/packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
+++ b/packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
@@ -48,46 +48,53 @@
   }
 
   $: displayColumns = columns.map(c => {
-    const isRequired = helpers.schema.isRequired(c.schema.constraints)
-    const requiredTooltip = isRequired && "Required columns must be writable"
-    const editEnabled =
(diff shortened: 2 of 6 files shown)


packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
```diff
@@ -94,16 +186,51 @@
...
 </div>
 
+{#if allowRelationshipSchemas}
+  <Popover
+    on:close={() => (relationshipFieldName = null)}
+    open={!!relationshipField}
+    anchor={relationshipPanelAnchor}
+    align="right-outside"
+  >
+    <svelte:self
```
aptkingston, line 220: Love to see usage of `svelte:self` 👌

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
```diff
@@ -2,16 +2,29 @@
   import { getContext } from "svelte"
   import { ActionButton, Popover } from "@budibase/bbui"
   import ColumnsSettingContent from "./ColumnsSettingContent.svelte"
+  import { FieldPermissions } from "../../../constants"
```
mike12345567, line 5: Should we move these to `@budibase/types`?

packages/frontend-core/src/components/grid/stores/datasource.js
```diff
@@ -147,19 +168,34 @@ export const createActions = context => {
   }
 
   // Adds a schema mutation for a single field
-  const addSchemaMutation = (field, mutation) => {
+  const addSchemaMutation = (field, mutation, fromNestedField) => {
```
aptkingston, line 171: I like the nice logical separation of `schemaMutations` and `subSchemaMutations`. Could we add a dedicated function for `addSubSchemaMutation` instead of keeping both types of mutation inside `addSchemaMutation`?

packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
```diff
@@ -1,87 +1,179 @@
...
+    (acc, c) => ({
+      ...acc,
+      [c]: {
+        disabled: disabledPermissions.includes(c),
+      },
+    }),
+    {}
+  )
+
+  $: allowRelationshipSchemas = true // TODO
```
mike12345567, line 33: Extra `// TODO`

packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
```diff
@@ -1,87 +1,179 @@
...
+
+  $: displayColumns = columns.map(c => {
```
mike12345567, line 35: could we break this out into a function? Right now I think this will react to a lot of different possible changes - maybe that is desired, but I'm guessing it should only react if the columns change

packages/frontend-core/src/components/grid/stores/cache.js
aptkingston, line 1: Very glad to see you found this file and updated it to cache the full table definitions! My comment prediction turned out to be true after all 😂

aptkingston (approved):
Looks great @adrinr! You've integrated the concept of `subSchemaMuations` really nicely with the existing codebase 👌 Made one tiny suggestion but looks super.

packages/frontend-core/src/components/grid/controls/ColumnsSettingButton.svelte
adrinr: No, they are only used in the frontend. The API expects the `hidden` and `readonly` fields, mapped accordingly on the frontend

packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
adrinr: This is left on propose, it will require the feature flag checks. It will be added separately in a further PR

packages/frontend-core/src/components/grid/stores/datasource.js
adrinr: Done

packages/frontend-core/src/components/grid/controls/ColumnsSettingContent.svelte
adrinr: Most of these things will be quite static, but we depend on reacting to all of them if loaded async or something changes via its parameters. For this it was created like this, otherwise it would need a function that gets almost all the same parameters

PR #14427, opened 2024-08-21
Persist view related schemas

adrinr:
## Description
We are going to enable picking certain fields from related fields. These fields (and only these ones) will be used when enriching relations. This is a wide change, this PR only updates the types in order to support the proper configuration at the API level.
1. Add a new field `schema` on relationship metadata. This will persist the field visibility for that given relationship (using the same writable/readonly/hidden strategy used in views)
2. Update the view schema fields metadata to also support this configuration, allowing the views to also store the permission information
3. Updating the view dto mapping to persist these new fields (this is not required for tables as we don't use API dtos and mappers in their save endpoints)

## Launchcontrol
Updating types around relationship fields

### DIFF
--- a/packages/server/src/api/controllers/view/viewsV2.ts
+++ b/packages/server/src/api/controllers/view/viewsV2.ts
@@ -3,11 +3,12 @@ import {
   CreateViewRequest,
   Ctx,
   RequiredKeys,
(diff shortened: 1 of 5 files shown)


mike12345567 (approved):
LGTM!

PR #14435, opened 2024-08-22
Do not force include primary display

adrinr:
## Description
Do not force include primary display. The primary display column will be used as enrichment as `primaryDisplay` field (as it is now, to avoid breaking changes on the api), but the actual field (being `name`, `title` or whatever it might be) it can be included or not, being up to the user configuration


adrinr:
https://linear.app/budibase/issue/BUDI-8563/view-join-column-test-cases

samwho (approved):

PR #14439, opened 2024-08-23
Enrich on squashing

adrinr:
## Description
Enabling squashing using the new relationships schemas. The new changes allow setting the schemas on relationships, so this PR ensures that we are mapping the expected fields on the row responses.
This PR includes the testing of the different retrieval methods, and sets SQS testing for the `row.spec` API tests.
This is enrichment it will be feature flagged for V3.

Includes:
- [BUDI-8563 View join column test cases](https://linear.app/budibase/issue/BUDI-8563/view-join-column-test-cases)

## Launchcontrol
Honor relationship schema on row retrieval

### DIFF
--- a/packages/server/src/db/linkedRows/index.ts
+++ b/packages/server/src/db/linkedRows/index.ts
@@ -272,7 +272,8 @@ export async function squashLinksToPrimaryDisplay(
         const obj: any = { _id: link._id }
         obj.primaryDisplay = getPrimaryDisplayValue(link, linkedTable)
 
-        if (schema.schema) {
+        const allowRelationshipSchemas = true // TODO
+        if (schema.schema && allowRelationshipSchemas) {
           for (const relField of Object.entries(schema.schema)
             .filter(([_, field]) => field.visible !== false)
             .map(([fieldKey]) => fieldKey)) {
--- a/packages/server/src/db/linkedRows/index.ts
+++ b/packages/server/src/db/linkedRows/index.ts
@@ -253,20 +253,33 @@ export async function squashLinksToPrimaryDisplay(
   // will populate this as we find them
   const linkedTables = [table]
   const isArray = Array.isArray(enriched)
-  let enrichedArray = !isArray ? [enriched] : enriched
-  for (let row of enrichedArray) {
+  const enrichedArray = !isArray ? [enriched] : enriched
+  for (const row of enrichedArray) {
     // this only fetches the table if its not already in array
     const rowTable = await getLinkedTable(row.tableId!, linkedTables)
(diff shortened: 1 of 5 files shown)


packages/server/src/utilities/rowProcessor/tests/outputProcessing.spec.ts
```diff
@@ -10,6 +10,14 @@ import { outputProcessing } from ".."
 import { generator, structures } from "@budibase/backend-core/tests"
 import * as bbReferenceProcessor from "../bbReferenceProcessor"
 
+jest.mock("@budibase/backend-core", () => ({
+  ...jest.requireActual("@budibase/backend-core"),
+  db: {
+    ...jest.requireActual("@budibase/backend-core").db,
+    isSqsEnabledForTenant: () => true,
+  },
+}))
```
samwho, line 19: Can this not be done using environment variables? Mindful of avoiding jest mocks for that beautiful day we migrate away from jest.

samwho (approved):

packages/server/src/db/linkedRows/index.ts
```diff
@@ -253,20 +253,34 @@ export async function squashLinksToPrimaryDisplay(
...
       }
       const newLinks = []
-      for (let link of row[column]) {
+      for (const link of row[column]) {
         const linkTblId = link.tableId || getRelatedTableForField(table, column)
         const linkedTable = await getLinkedTable(linkTblId!, linkedTables)
         const obj: any = { _id: link._id }
         obj.primaryDisplay = getPrimaryDisplayValue(link, linkedTable)
+
+        const allowRelationshipSchemas = true // TODO
```
mike12345567, line 275: What was the `TODO` here about - is it left over or was there something intended?

mike12345567 (approved):

packages/server/src/db/linkedRows/index.ts
adrinr: This is waiting for the feature branch

packages/server/src/utilities/rowProcessor/tests/outputProcessing.spec.ts
adrinr: I did it this way as `outputProcessor` now checks if SQS is enabled to trim some response fields. Because of this, we are adding a dependency to the context. Because these tests are simple unit tests, they fail because the context is not set.
I could use the TestConfiguration, but it adds extra unneeded boilerplate, making it slower. And anyway, this mock will be removed once the SQS feature flag is gone. If we add the full TestConfig we will never remember to remove it even if it will be not needed anymore

PR #14441, opened 2024-08-23
Change Automations CodeEditorModal to update on close

deanhannigan:
## Description

The `on change` behaviour was too chatty for the code editor causing issues with the UX. The API update behaviour has now been changed to only fire when the modal itself closes. This should greatly improve the editor experience.

## Addresses
- https://linear.app/budibase/issue/BUDI-8587/unpredictable-behaviour-when-typing-quickly-in-js-scripting-automation

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -1078,13 +1078,8 @@
                   >
                     <CodeEditor
                       value={inputData[key]}
-                      on:blur={e => {
+                      on:change={e => {
                         // need to pass without the value inside
-                        inputData[key] = e.detail
-                      }}
-                      on:insert={e => {
-                        // Blur events only happen once after inserting a binding
-                        // State will now update on insert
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -1065,7 +1065,12 @@
                 value={inputData[key]}
               />
             {:else if value.customType === "code"}
-              <CodeEditorModal>
+              <CodeEditorModal
+                on:hide={() => {
+                  // Push any pending changes when the window closes
+                  onChange({ [key]: inputData[key] })
```
aptkingston, line 1071: Does this account for saving vs cancelling? Like if you close the modal by clicking the X icon or clicking outside it, would we not want to ignore the change rather than save it?

aptkingston: Although I appreciate that this is how the old modal used to work, since it updated on every key stroke. So this is probably fine since that's how it's always worked.

aptkingston (approved):

PR #14447, opened 2024-08-26
Row action view security crud

adrinr:
## Description
Allow CRUD for row action view security.

### DIFF
--- a/packages/server/src/api/routes/rowAction.ts
+++ b/packages/server/src/api/routes/rowAction.ts
@@ -50,6 +50,16 @@ router
     authorizedResource(PermissionType.TABLE, PermissionLevel.READ, "tableId"),
     rowActionController.remove
   )
+  .post(
+    "/api/tables/:tableId/actions/:actionId/permissions/:viewId",
+    authorizedResource(PermissionType.TABLE, PermissionLevel.READ, "tableId"),
+    rowActionController.setViewPermission
+  )
+  .delete(
+    "/api/tables/:tableId/actions/:actionId/permissions/:viewId",
+    authorizedResource(PermissionType.TABLE, PermissionLevel.READ, "tableId"),
+    rowActionController.unsetViewPermission
+  )
 
   // Other endpoints
   .post(
--- a/packages/server/src/api/routes/tests/rowAction.spec.ts
+++ b/packages/server/src/api/routes/tests/rowAction.spec.ts
@@ -491,36 +491,54 @@ describe("/rowsActions", () => {
     })
   })
 
-  describe("setViewPermission", () => {
-    unauthorisedTests((expectations, testConfig) =>
-      config.api.rowAction.setViewPermission(
-        tableId,
-        generator.guid(),
-        generator.guid(),
-        expectations,
-        testConfig
+  describe("set/unsetViewPermission", () => {
+    describe.each([
+      ["setViewPermission", config.api.rowAction.setViewPermission],
+      ["unsetViewPermission", config.api.rowAction.unsetViewPermission],
+    ])("unauthorisedTests for %s", (__, delegateTest) => {
+      unauthorisedTests((expectations, testConfig) =>
+        delegateTest(
+          tableId,
+          generator.guid(),
+          generator.guid(),
+          expectations,
+          testConfig
+        )
       )
-    )
+    })
 
-    it("can set permission views", async () => {
+    let tableIdForDescribe: string
+    let actionId1: string, actionId2: string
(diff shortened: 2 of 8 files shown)


packages/server/src/sdk/app/rowActions.ts
```diff
@@ -97,55 +100,97 @@ export async function docExists(tableId: string) {
...
+    return actionsDoc
+  })
+}
+
+async function guardView(tableId: string, viewId: string) {
+  let view
+  try {
+    view = await sdk.views.get(viewId)
+  } catch (e: any) {
+    if (e?.message !== "Unable to extract table ID, is not a view ID") {
```
samwho, line 149: Not a fan on relying on error messages like this. Is there a function we could use to check if the ID is a valid view ID instead?

packages/types/src/api/web/app/rowAction.ts
```diff
@@ -8,6 +8,7 @@ export interface RowActionResponse extends RowActionData {
   id: string
   tableId: string
   automationId: string
+  allowedViews: string[] | undefined
```
samwho, line 11:
```suggestion
  allowedViews?: string[]
```

samwho (commented):
Is this the permissions stuff we were talking about this morning in standup? If so, is this still relevant or are we going to tie view action permissions to read permissions?

packages/server/src/api/routes/rowAction.ts
```diff
@@ -50,6 +50,16 @@ router
     authorizedResource(PermissionType.TABLE, PermissionLevel.READ, "tableId"),
     rowActionController.remove
   )
+  .post(
+    "/api/tables/:tableId/actions/:actionId/permissions/:viewId",
+    authorizedResource(PermissionType.TABLE, PermissionLevel.READ, "tableId"),
```
mike12345567, line 55: These endpoints are only used for setting the view permissions - I think these should probably only be authorized as builder actions - these should never be called by real users in the end app.

packages/server/src/sdk/app/rowActions.ts
mike12345567: I would agree - it likely makes sense to expose the `isViewID` function from the SDK so this can check the error and handle the scenario in its own unique way rather than relying on the exception.

mike12345567 (changes requested):
Looks good - just need to update the API authorization middlewares for the builder only endpoints.

packages/server/src/api/routes/tests/rowAction.spec.ts
```diff
@@ -462,4 +490,158 @@ describe("/rowsActions", () => {
...
+    beforeAll(async () => {
+      tableIdForDescribe = tableId
+      for (const rowAction of createRowActionRequests(3)) {
+        await createRowAction(tableId, rowAction)
+      }
+      const persisted = await config.api.rowAction.find(tableId)
+
+      const actions = _.sampleSize(Object.keys(persisted.actions), 2)
+      actionId1 = actions[0]
+      actionId2 = actions[1]
```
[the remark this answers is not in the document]
adrinr: This cannot be done here, as the `actionId1` and `actionId2` are declared outside the `beforeAll`

packages/server/src/sdk/app/rowActions.ts
adrinr: I agree that that's cleaner. Done!

packages/types/src/api/web/app/rowAction.ts
adrinr: This type is use as type response. Not having the `?` will enforce us to declare this field, even if it is set to undefined. Otherwise, we might be missing some mappings and the type will not complain

adrinr:
> Is this the permissions stuff we were talking about this morning in standup? If so, is this still relevant or are we going to tie view action permissions to read permissions?

This is still required. Each row action will be created per table, but to trigger it from a view it will need to be explicitly toggled. My question during the standup was who can actually trigger it?
In short, the flow will be as follows:
1. A row action A is called via the view B
2. First check: is row action A enabled for the view B (this is what this PR crud is about)
3. Second check: does the user calling have enough permissions to trigger this action (how this is done is yet to be discussed and implement in a further PR)

samwho (approved):

adrinr:
> Looks good - just need to update the API authorization middlewares for the builder only endpoints.

This is done in here: https://github.com/Budibase/budibase/pull/14448

mike12345567 (approved):

PR #14459, opened 2024-08-27
Feature flagging pick relationship fields

adrinr:
## Description
Feature flagging pick relationship fields


packages/backend-core/src/features/index.ts
```diff
@@ -267,4 +267,5 @@ export class FlagSet<V extends Flag<any>, T extends { [key: string]: V }> {
 // default values set correctly and their types flow through the system.
 export const flags = new FlagSet({
   DEFAULT_VALUES: Flag.boolean(false),
+  [FeatureFlag.ENRICHED_RELATIONSHIPS]: Flag.boolean(false),
```
samwho, line 270: It'll be nice when the frontend is all TypeScript as well because we won't need enums. We'd get full type checking on the literal strings.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -2648,6 +2667,67 @@ describe.each([
...
+                relWithEmptySchema: [otherRows[1]],
+                relWithFullSchema: [otherRows[2]],
+                relWithHalfSchema: [otherRows[3]],
+                relWithIllegalSchema: [otherRows[4]],
+              })
+
+              const retrieved = await retrieveDelegate(row)
+
+              expect(retrieved).toEqual(
+                expect.objectContaining({
```
samwho, line 2693: Is this a correct way to assert what we want here? Won't this assertion also pass if the row _is_ enriched?

packages/backend-core/src/features/index.ts
adrinr: Yes, it will be. But I think I still prefer enums for these use cases, as it is easier to find references and refactor. Otherwise we have just typed magic strings.

packages/server/src/api/routes/tests/row.spec.ts
adrinr: Yes. `expect.objectContaining` does not act in nested properties. In this case we want to make sure that `relWithNoSchema` and any other field only have `_id` and `primaryDisplay`

packages/backend-core/src/features/index.ts
samwho: What specifically is made easier by the enums? With the typed strings, your editor will suggest to you the correct options after typing `isEnabled(`, and if you delete a flag, TypeScript will error all of the places that flag is checked.

samwho (approved):

PR #14461, opened 2024-08-27
Refactor automation orchestrator to support branching

PClmnt:
## Description
This refactors the automation orchestrator to facilitate branching and in the process improves how we handle looping. We now have an entry point for loop / branch / normal steps

## Addresses
https://linear.app/budibase/issue/BUDI-8530/update-automation-orchestrator-support-branching

## Feature branch env
[Feature Branch Link](http://fb-feat-support-branching-orchestrator.fb.qa.budibase.net)


packages/server/src/automations/tests/scenarios/scenarios.spec.ts
```diff
@@ -23,42 +23,42 @@ describe("Automation Scenarios", () => {
...
+        .appAction({ fields: {} })
+        .serverLog({ text: "Starting automation" })
+        .branch({
+          topLevelBranch1: {
+            steps: stepBuilder =>
+              stepBuilder.serverLog({ text: "Branch 1" }).branch({
+                branch1: {
+                  steps: stepBuilder =>
+                    stepBuilder.serverLog({ text: "Branch 1.1" }),
+                  condition: { notEmpty: { column: 10 } },
```
PClmnt, line 42: All the conditions here are just placeholders right now

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +294,252 @@ class Orchestrator {
...
+    })
+
+    try {
+      if (this._stopped) {
+        this.updateExecutionOutput(step.id, step.stepId, {}, STOPPED_STATUS)
+        return
+      }
+
+      let originalStepInput = cloneDeep(step.inputs)
+      if (this._loopStep) {
```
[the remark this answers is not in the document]
PClmnt: We could check iteration, we would also need to pass it down through to the `handleStepOutput` function as well, do you think that's okay?

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +294,252 @@ class Orchestrator {
...
+            status: AutomationErrors.MAX_ITERATIONS,
+            success: true,
+          }
+        )
+        break
+      }
+
+      let isFailure = false
+      const currentItem = this.getCurrentLoopItem(loopStepIndex)
+      if (currentItem && typeof currentItem === "object") {
```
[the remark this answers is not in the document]
PClmnt: I remember trying to change this before and it breaking, I think it actually might be okay that it is true for arrays.

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +294,252 @@ class Orchestrator {
...
+        ) {
           await this.updateMetadata(metadata)
         }
         return this.executionOutput
       }
     )
   }
+
+  private async executeSteps(steps: AutomationStep[]): Promise<void> {
+    let stepIndex = 0
```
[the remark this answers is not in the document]
PClmnt: stepIndex isn't used outside of `executeSteps` though, at least not that I noticed :D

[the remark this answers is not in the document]
samwho: It's passed around a fair bit, which you could get rid of if you wanted to have it as state on the class. Mostly through `executeLoopStep`.

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +294,252 @@ class Orchestrator {
...
+      } catch (err) {
+        this.updateContextAndOutput(
+          currentIndex,
+          this._loopStep,
+          {},
+          {
+            status: AutomationErrors.INCORRECT_TYPE,
+            success: false,
+          }
+        )
```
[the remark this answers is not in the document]
PClmnt: yep, missing a break.

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +294,252 @@ class Orchestrator {
...
+    while (stepIndex < steps.length) {
+      if (timeoutFlag) {
+        this._parentSpan?.addTags({ timedOut: true })
+        break
+      }
+
+      if (timeout) {
+        setTimeout(() => {
+          timeoutFlag = true
+        }, timeout || env.AUTOMATION_THREAD_TIMEOUT)
```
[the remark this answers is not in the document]
PClmnt: How does that look now?

samwho (approved):
Really nice work!

packages/types/src/documents/app/automation/StepInputsOutputs.ts
```diff
@@ -100,7 +100,7 @@ export type FilterStepOutputs = BaseAutomationOutputs & {
 export type LoopStepInputs = {
   option: LoopStepType
   binding: any
-  iterations?: number
+  iterations?: string
```
adrinr, line 103: Why this? Should this not always be a number and cast any possible usage that comes as a string?

packages/server/src/threads/automation.ts
```diff
@@ -70,6 +72,9 @@ class Orchestrator {
   _emitter: any
   _context: AutomationContext
   _job: Job
+  _loopStepOutputs: LoopStep[]
+  _stopped: boolean
+  _loopStep?: LoopStep
```
adrinr, line 77: Could all these `_whatever` be marked as `private`? I guess this is what we are trying to make with the `_`

packages/types/src/documents/app/automation/StepInputsOutputs.ts
PClmnt: Using parseInt on a number gives type errors. And the frontend always sends this back as a string.

packages/server/src/threads/automation.ts
PClmnt: Sorted this!

packages/types/src/documents/app/automation/StepInputsOutputs.ts
adrinr: Maybe there is something else we should handle then. Not sure if the solution is to make our types more dirty

PClmnt: Changed the type back to number and made sure we run it through a parser function before use

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +290,288 @@ class Orchestrator {
...
+
+  private async evaluateBranchCondition(
+    conditions: SearchFilters
+  ): Promise<boolean> {
+    const toFilter: Record<string, any> = {}
+
+    const processFilter = (filter: SearchFilters) => {
+      Object.entries(filter).forEach(([_, value]) => {
+        Object.entries(value).forEach(([field, _]) => {
+          const fromContext = processStringSync(
```
mike12345567, line 479: This turned out nice!

packages/server/src/threads/automation.ts
```diff
@@ -544,13 +290,288 @@ class Orchestrator {
...
+            `{{ literal ${field} }}`,
+            this.context
+          )
+          toFilter[field] = fromContext
+        })
+      })
+      return filter
+    }
+    processFilter(conditions)
```
mike12345567, line 488: Why are both calls required here? Should the recursing of the logical operators not automatically process filters at the top of the condition?

mike12345567 (approved):

packages/server/src/threads/automation.ts
PClmnt: In the `recurseLogicalOperators` function it only runs the passed function if it sees a logical operator, so I run the function on my top level first then we can do any recursive stuff after that.

mike12345567: ah - I see, we should probably have a utility function which is just `recurseSearchFilters` which does both jobs in one!

PClmnt: Done!

PR #14467, opened 2024-08-28
Refactor view usage

adrinr:
## Description
Refactoring the way we handle the view endpoints. Until now, row endpoints had the same behaviour for both views and tables. The only difference was that some data was trimmed on some middleware.
Now we will need some extra logic specific to the views, so we want to have this information in the controller. For this, we don't mutate the source data and we handle it properly on the controller level.

Also, adding SQS tests for the row controller.

## Launchcontrol
Code refactor

### DIFF
--- a/packages/server/src/sdk/app/rows/search.ts
+++ b/packages/server/src/sdk/app/rows/search.ts
@@ -122,9 +122,8 @@ export async function fetchRaw(tableId: string): Promise<Row[]> {
 }
 
 export async function fetchView(
-  tableId: string,
   viewName: string,
   params: ViewParams
 ): Promise<Row[]> {
-  return pickApi(tableId).fetchView(viewName, params)
+  return internal.fetchView(viewName, params)
 }
--- a/packages/server/src/sdk/app/rows/search.ts
+++ b/packages/server/src/sdk/app/rows/search.ts
@@ -121,9 +121,9 @@ export async function fetchRaw(tableId: string): Promise<Row[]> {
   return pickApi(tableId).fetchRaw(tableId)
 }
 
-export async function fetchView(
+export async function fetchLegacyView(
   viewName: string,
   params: ViewParams
 ): Promise<Row[]> {
-  return internal.fetchView(viewName, params)
+  return internal.fetchLegacyView(viewName, params)
 }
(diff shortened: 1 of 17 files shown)


packages/server/src/sdk/app/rows/search.ts
```diff
@@ -122,9 +122,8 @@ export async function fetchRaw(tableId: string): Promise<Row[]> {
 }
 
 export async function fetchView(
-  tableId: string,
   viewName: string,
   params: ViewParams
 ): Promise<Row[]> {
-  return pickApi(tableId).fetchView(viewName, params)
+  return internal.fetchView(viewName, params)
```
mike12345567, line 128: Should this always go to the internal API?

packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -349,11 +350,19 @@ export async function outputProcessing<T extends Row[] | Row>(
   }
   // remove null properties to match internal API
   const isExternal = isExternalTableID(table._id!)
-  if (isExternal) {
+  if (isExternal || db.isSqsEnabledForTenant()) {
```
mike12345567, line 353: This check might cause a problem with @samwho feature flagging work - worth checking in on that.

packages/server/src/sdk/app/rows/search.ts
adrinr: I just renamed it.
It is for the legacy views. We don't have external legacy views. It was working fine as the current code checks if the provided table id is external to pick the external sdk. Otherwise, it uses the internal. Because the table id was undefined on this controller, it did always fall-back to internal.

packages/server/src/utilities/rowProcessor/index.ts
adrinr: This has been cherry-picked from the work in `feat/pick-relationship-fields`

mike12345567 (approved):
LGTM - thanks for addressing comment!

packages/server/src/utilities/rowProcessor/index.ts
mike12345567: Ah makes sense!

PR #14470, opened 2024-08-28
The UI should indicate who the account holder is

melohagan:
## Description
In the users screen, the account holder is now sorted to the top of the list and indicated with an additional access level called *Account holder*.

Row selection is disabled for this user.

This will only be available for *Cloud* accounts that have been created since the **20th June 2024**, as this was the date that the TenantOwner info was saved to CouchDB when onboarding.

All older and self-hosted accounts will display the tenant owner as **Account admin**.
This will exhibit the behaviour that is seen today where you can "delete" the account holder, but nothing happens.

### Additional changes

- Removed the Joi validation when saving TenantInfo, because this is metadata about the Tenant that we want to save regardless of the shape of the payload. Don't want a change in Tenant structure to cause a quiet bug like it did in this case.
- Added backend validation to prevent account holder role being changed
- Made front-end changes to remove options to delete or change the role of the *Account holder*

## Addresses
- https://linear.app/budibase/issue/GRO-456/the-ui-should-indicate-who-the-account-holder-is

## Screenshots
Screenshot 2024-08-28 at 16 32 06

**Account holder** appears at the top of the list and cannot be selected. 

Screenshot 2024-08-28 at 16 33 36

Select all does not select the **Account holder**.

image

**Account holder** cannot have their role changed, or be deleted.

## Launchcontrol
Display account holder in users screen


jvcalderon (approved):
Good one! 😃 LGTM

packages/frontend-core/src/fetch/UserFetch.js
```diff
@@ -47,10 +47,17 @@ export default class UserFetch extends DataFetch {
         limit,
       }
       const res = await this.API.searchUsers(opts)
+      let tenantInfo
+      try {
+        tenantInfo = await this.API.getTenantInfo({ tenantId })
```
shogunpurple, line 52: NAB: would it make more sense to return an empty response from the API if there's no tenantInfo for the tenant, rather than just relying on it failing?

shogunpurple (commented):
LGTM, but I'm wondering if the `tenantInfo` is necessary at all - I thought that we were able to determine the account holder without this because we have `ctx.user.account` in the backend that gets added from their session

packages/frontend-core/src/fetch/DataFetch.js
```diff
@@ -62,6 +62,7 @@ export default class DataFetch {
       cursors: [],
       resetKey: Math.random(),
       error: null,
+      tenantOwner: null,
```
aptkingston, line 65: I'd prefer not to add specific metadata like tenant owner into the generic data fetch - we should try and keep this fully generic. I also don't think it belong at this level. I think that if you need to know the tenant owner then you should do the required API call to get that data, rather than always fetching the tenant owner every time you do an API call to get users.

packages/bbui/src/Table/SelectEditRenderer.svelte
```diff
@@ -6,10 +6,11 @@
   export let onEdit
   export let allowSelectRows = false
   export let allowEditRows = false
+  export let allowSelectedOverride
```
aptkingston, line 9: I think a slightly better way to handle this would just be reading it from the row itself, rather than pulling it into a separate prop. The flag already exists in the row, and we have `data` as an available prop (which is the row in this case), but it just isn't currently used here.

So we can just add `export let data` and consume it as `data.allowSelectedOverride`.

packages/builder/src/pages/builder/portal/users/users/index.svelte
```diff
@@ -110,11 +112,18 @@
           })
         }
       })
+      user.tenantOwnerEmail = res.tenantOwner?.email
```
aptkingston, line 115: We should just load the tenant info separately here, and combine that with the user list to generate our data to pass into the table. This may also have outgrown being a reactive statement, so it might be a good idea to pull this into a dedicated function now and pass in `$fetch.rows`, `$groups` and the tenant owner (and any other dependencies I've missed).

aptkingston (changes requested):
UI changes look good!

Left a comment about how you're bundling the tenant owner info into the core data fetch and user fetch, as I'd prefer to not bundle this logic in there and instead just load is as required using its own endpoint that we're already using elsewhere.

Other minor thing is just in terms of naming - I find `allowSelectedOverride` a little verbose and difficult to understand. What we do elsewhere is prefix these internal flags with a double underscore to avoid clashing with other property names and also make it obvious that it's just an internal flag, so in this case i think` __selectable: false` would be a great candidate for this flag to indicate that it can't be selected inside the table. 

As for consuming this flag, as I've mentioned in a comment we can then just always refer to `row.__selectable` within the table codebase (which most of the time you're doing anyway, other than pulling it out into a separate prop for the `SelectEditRenderer`.

packages/worker/src/api/routes/global/tests/users.spec.ts
```diff
@@ -412,6 +412,28 @@ describe("/api/global/users", () => {
       expect(events.user.permissionBuilderRemoved).toHaveBeenCalledTimes(1)
     })
 
+    it("should not be able to update an account holder user to a basic user", async () => {
```
deanhannigan, line 415: 🥚

deanhannigan (approved):
LGTM! Nice work on the updates! I ran through all the issues we discussed yesterday and all have been resolved 👍

aptkingston (approved):
LGTM! One suggestion just to save on API calls, but looks good 👍

PR #14478, opened 2024-08-29
Fix for force password reset

melohagan:
## Description
Set password to 12 characters when forcing password reset.

## Addresses
- https://linear.app/budibase/issue/BUDI-8597/force-password-reset-hitting-validation-error


jvcalderon (approved):
LGTM

PR #14482, opened 2024-08-30
Display relationship fields

adrinr:
## Description
Given that we introduced the ability to pick field relations, we want to display them on the grid. This will be used to give feedback to the user about the updates they have made.

### Relationships with all data visible, with some long texts in both column and data

https://github.com/user-attachments/assets/8d202b8e-e3f3-42ef-acfe-1d92003b1c69

### All the relationships will show in the same position for each cell

https://github.com/user-attachments/assets/62dc78b3-60b4-4ddc-861c-023fcafd30cf

### Having some relationships with no-related data populated
See `a4`, the pill just does not display 

https://github.com/user-attachments/assets/525b302d-cacd-4ee5-abf7-d7d1873f6351

### Relationship fields will only display when focused (but not editing), to avoid too much noise

https://github.com/user-attachments/assets/7b4f5bc7-77d5-478c-a62f-4a8458ca2463

## Launchcontrol
Display relationship fields

### DIFF
--- a/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
@@ -469,6 +469,10 @@
   .values.wrap .badge:hover::before {
     opacity: 0.7;
   }
+  .values.wrap .badge.extra-info {
+    cursor: pointer;
+  }
+
   .badge span {
     overflow: hidden;
     white-space: nowrap;
@@ -544,12 +548,14 @@
   .relationship-fields {
     margin: var(--spacing-m) var(--spacing-l);
     display: grid;
-    grid-template-columns: auto auto;
+    grid-template-columns: minmax(auto, 75%) auto;
     grid-row-gap: var(--spacing-m);
     grid-column-gap: var(--cell-spacing);
   }
 
-  .values.wrap .badge.extra-info {
-    cursor: pointer;
+  .relationship-field-name {
+    text-transform: uppercase;
+    color: var(--spectrum-global-color-gray-600);
+    font-size: var(--font-size-xs);
   }
 </style>
--- a/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
@@ -42,6 +42,20 @@
     }
   }
 
+  $: relationFields = fieldValue?.reduce((acc, f) => {
+    const fields = {}
+    for (const [column] of Object.entries(schema?.columns || {}).filter(
+      ([key, column]) =>
+        column.visible !== false && f[key] !== null && f[key] !== undefined
+    )) {
+      fields[column] = f[column]
+    }
+    if (Object.keys(fields).length) {
+      acc[f._id] = fields
+    }
+    return acc
+  }, {})
+
   const parseValue = value => {
     if (Array.isArray(value) && value.every(x => x?._id)) {
       return value
@@ -223,20 +237,7 @@
   }
 
   const displayRelationshipFields = relationship => {
-    const fields = {}
-    for (const column of Object.entries(schema.columns)
-      .filter(
-        ([key, column]) =>
-          column.visible !== false &&
-          relationship[key] !== null &&
-          relationship[key] !== undefined
-      )
-      .map(([key]) => key)) {
-      fields[column] = relationship[column]
-    }
-    if (Object.keys(fields).length) {
-      relationshipFields = fields
-    }
+    relationshipFields = relationFields[relationship._id]
   }
 
   const hideRelationshipFields = () => {
--- a/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
+++ b/packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
@@ -56,6 +56,12 @@
     return acc
   }, {})
 
+  $: showRelationshipFields =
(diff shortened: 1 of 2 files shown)


packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -478,4 +531,19 @@
...
+    grid-template-columns: auto auto;
+    grid-row-gap: var(--spacing-m);
+    grid-column-gap: var(--cell-spacing);
+  }
+
+  .values:not(.wrap) .badge.extra-info {
+    cursor: pointer;
+  }
+  .values:not(.wrap) .badge.extra-info:hover::before {
+    opacity: 0.7;
```
aptkingston, line 547: NAB: instead of using a pseudo ::before element and changing the opacity like this, it would be simpler (and more performant) to add something like `filter: brightness(1.2)` to the badge on hover.

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -322,6 +351,21 @@
   </GridPopover>
 {/if}
 
+{#if relationshipFields && !focused}
+  <GridPopover {open} {anchor}>
+    <div class="relationship-fields">
+      {#each Object.entries(relationshipFields) as [fieldName, fieldValue]}
+        <div class="relationship-field-name">
```
aptkingston, line 358: NAB: I think it might look nicer with a different font style for column names. Something like all caps, grey-600 or gery-700, slightly smaller font size (to balance the caps).

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -322,6 +351,21 @@
...
+          {fieldName}
+        </div>
+        <div class="relationship-field-value">
+          {fieldValue}
```
aptkingston, line 362: NAB: I think we should add a min width here to ensure we can at least render a few words for the value before line wrapping, then also use a line clamp of 3 lines to prevent values getting too long. You'll also need a few new styles to ensure long values get properly truncated with ellipses properly.

aptkingston (commented):
Nice quality of life improvement! Left a view comments about styles. 

As discussed in the group chat, I'd much prefer to only show this information when the cell is selected to prevent flashing these popovers unintentionally.

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
adrinr: Good shout, it looks much nicer!

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
adrinr: I did not know about this css API. Much easier :)

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -322,6 +357,21 @@
   </GridPopover>
 {/if}
 
+{#if relationshipFields}
```
aptkingston, line 360: NAB/performance optimisation: Does this work the same if we instead do `{#if showRelationshipFields}...`? Just think that otherwise we actually end up and unmounting this grid popover rapidly when moving the mouse over the grid, even if the cells aren't focused (since we still have the mouseover handler). Updating the if statement to use `showRelationshipFields` would ensure it only gets mounted when we intend to show it. I thiynk we can just leave the `open` prop out then, since it defaults to true.

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -248,7 +277,13 @@
     >
       {#each fieldValue || [] as relationship}
         {#if relationship[primaryDisplay] || relationship.primaryDisplay}
-          <div class="badge">
+          <div
+            class="badge"
+            class:extra-info={!!relationFields[relationship._id]}
```
aptkingston, line 282: NAB/performance optimsation: Since svelte has no nice way to conditionally attach handlers, we could disable pointer events using CSS when the cell is not focused, which would improve performance since these mouseover/leave handlers wouldn't fire unless the cell is focused.

We could achieve that with something like
`class:disabled={!focused}`
and CSS of
```css
.values.disabled {
  pointer-events: none;
}
```

I would attach this to the `.values` div rather than each badge, since that encompasses all badges.

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
```diff
@@ -322,6 +357,21 @@
   </GridPopover>
 {/if}
 
+{#if relationshipFields}
+  <GridPopover open={showRelationshipFields} {anchor}>
```
aptkingston, line 361: NAB: I think we should increase with min width here so that average length text values have more room 👍  Something like `minWidth={300}`, or more, depending what you think looks best.

aptkingston (commented):
Looks much better with those changes, nice one! Left a couple more performance improvement ideas, just because the grid can be heavy on the browser at times 👍

packages/frontend-core/src/components/grid/cells/RelationshipCell.svelte
adrinr: Done: 
image

aptkingston (approved):
LGTM!

PR #14486, opened 2024-08-30
Disable creation of createdAt and createdBy auto columns when default values is enabled.

samwho:
## Description

Because it's possible to replicate the functionality of CREATED_AT and CREATED_BY columns, we disable their creation when the DEFAULT_VALUES feature flag is enabled.


mike12345567 (approved):
LGTM!

PR #14491, opened 2024-08-30
Make sure setEnrichedUsers is called when tenantOwnerLoaded

melohagan:
## Description
Issue found in QA that wasn't found locally because of differences in race-conditions.

Need to make sure that when the tenant owner query is complete that the users are enriched and set.

### DIFF
--- a/packages/builder/src/pages/builder/portal/users/users/index.svelte
+++ b/packages/builder/src/pages/builder/portal/users/users/index.svelte
@@ -100,7 +100,7 @@
   $: pendingSchema = getPendingSchema(schema)
   $: userData = []
   $: inviteUsersResponse = { successful: [], unsuccessful: [] }
-  $: setEnrichedUsers($fetch.rows)
+  $: setEnrichedUsers($fetch.rows, tenantOwnerLoaded)
 
   const setEnrichedUsers = async rows => {
     if (!tenantOwnerLoaded) {


shogunpurple (approved):
LGTM!

aptkingston:
The usual way to do this is just pass in `tenantOwner` itself as a param to `setEnrichedUsers`. Then you don't need a separate loading flag, and the reactive statement will run again whenever the value of `tenantOwner` changes.

No need to change this since it's still safe to do it this way as tenant owner is only set on mount, but just a FYI for the future 👍

melohagan:
> The usual way to do this is just pass in `tenantOwner` itself as a param to `setEnrichedUsers`. Then you don't need a separate loading flag, and the reactive statement will run again whenever the value of `tenantOwner` changes.
> 
> No need to change this since it's still safe to do it this way as tenant owner is only set on mount, but just a FYI for the future 👍

In this particular case the loaded flag is necessary because I can't guarantee that the tenant_info doc, and thus tenantOwner will be found. If it's not found then `undefined` will be returned and the standard behaviour of treating *Account holders* as regular *Account admins* will be adopted. 

We may be able to back-fill older accounts (before June 20th) or this strategy of using the tenant-info doc might change in future. We'll see!

aptkingston:
Cool, didn't know that the tenant owner was a new concept. You're completely right then!

PR #14499, opened 2024-09-02
Add LICENSE_TYPE

calexiou:
## Description
Passes a `LICENSE_TYPE` to the workflow dispatch for the feature branches action.  

## Addresses
- fixes failed pipelines as a result of a license generation error


shogunpurple (approved):
LGTM

PR #14503, opened 2024-09-03
Support updating names

adrinr:
## Description
The new link fields relationships stores the name of the columns to define which fields are returned with link data. Because we don't have any immutable identifier, this data can get stale for some views when a column is renamed.
This PRs ensures that, on column rename, any view referring to that column will also be renamed. The applied method is not the most efficient one, as it needs to analyse many possible views. But given that this action can only happen on the builder, it should be fine.

https://github.com/user-attachments/assets/beb400a9-9673-4ee3-9353-df2226a9ce2a

## Launchcontrol
Handle view relationship fields while renaming columns

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -1204,18 +1204,18 @@ describe.each([
           )
           await config.api.table.save({
             ...table,
-              schema: {
+            schema: {
               ...table.schema,
-                aux: {
-                  name: "aux",
-                  relationshipType: RelationshipType.ONE_TO_MANY,
-                  type: FieldType.LINK,
-                  tableId: auxTable._id!,
-                  fieldName: "fk_aux",
+              aux: {
+                name: "aux",
+                relationshipType: RelationshipType.ONE_TO_MANY,
+                type: FieldType.LINK,
+                tableId: auxTable._id!,
+                fieldName: "fk_aux",
                 constraints: { type: "array" },
-                },
               },
-            })
+            },
+          })
           // Refetch auxTable
           auxTable = await config.api.table.get(auxTable._id!)
 
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -1238,11 +1238,11 @@ describe.each([
 
         const renameColumn = async (table: Table, renaming: RenameColumn) => {
           const newSchema = { ...table.schema }
-          ;(newSchema[renaming.updated] = {
+          newSchema[renaming.updated] = {
             ...table.schema[renaming.old],
             name: renaming.updated,
-          }),
-            delete newSchema[renaming.old]
+          }
+          delete newSchema[renaming.old]
 
           await config.api.table.save({
             ...table,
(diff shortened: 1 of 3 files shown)


packages/server/src/sdk/app/views/index.ts
```diff
@@ -252,3 +252,49 @@ export function syncSchema(
...
+export async function renameLinkedViews(table: Table, renaming: RenameColumn) {
+  const relatedTables: Record<string, Table> = {}
+
+  for (const field of Object.values(table.schema)) {
+    if (field.type !== FieldType.LINK) {
+      continue
+    }
+
+    relatedTables[field.tableId] ??= await sdk.tables.getTable(field.tableId)
+    break
```
samwho, line 265: I don't fully understand the purpose of this `break`. Wouldn't you want to fetch all related tables?

packages/server/src/sdk/app/views/index.ts
```diff
@@ -252,3 +252,49 @@ export function syncSchema(
...
+      sdk.views.isV2
+    )
+    if (!viewsV2) {
+      continue
+    }
+
+    for (const view of viewsV2) {
+      for (const relField of Object.keys(view.schema || {}).filter(f => {
+        const tableField = relatedTable.schema[f]
+        if (tableField.type !== FieldType.LINK) {
```
samwho, line 280: Is `tableField` guaranteed to be non-undefined here? Think this might be a case where not having strict index checks is giving us a false sense of security.

packages/server/src/sdk/app/views/index.ts
```diff
@@ -252,3 +252,49 @@ export function syncSchema(
...
+          return false
+        }
+
+        return tableField.tableId === table._id
+      })) {
+        const columns = view.schema && view.schema[relField]?.columns
```
samwho, line 286: Can you express this as:

```suggestion
        const columns = view.schema?.[relField]?.columns
```

packages/server/src/sdk/app/views/index.ts
adrinr: Good catch! This is left from a initial implementation that was behaving differently. I will check why the test I wrote did not catch that :/

adrinr: Tests added

packages/server/src/sdk/app/views/index.ts
adrinr: It should not be undefined, as there are other processes that cleanse old ones. But just in case, as technically it might be not the case, I will add a guard!

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -1200,48 +1210,68 @@ describe.each([
...
+            schema,
           })
 
+        const renameColumn = async (table: Table, renaming: RenameColumn) => {
+          const newSchema = { ...table.schema }
+          ;(newSchema[renaming.updated] = {
+            ...table.schema[renaming.old],
+            name: renaming.updated,
+          }),
+            delete newSchema[renaming.old]
```
samwho, line 1245: This looks kinda weird, how come this leading `;` is needed?

samwho (approved):

packages/server/src/api/routes/tests/viewV2.spec.ts
adrinr: It is not, just some weird autoformatting due some wrong parenthesis 🤷🏽‍♀️

PR #14508, opened 2024-09-03
Add API validation and some tests for automation branching

PClmnt:
## Description
This PR adds validation for automation branching to our API layer and also some tests to verify it is working correctly. 

Also fixed an issue where the branch step wasn't appearing the execution output that I noticed as part of this work, and updated some tests to cover this. 

## Addresses
https://linear.app/budibase/issue/BUDI-8533/validate-automation-structure-at-api


packages/server/src/tests/utilities/structures.ts
PClmnt, line 1: I know I know, I hate this ugly structures file as well, but had to add some as the Automation Test Builder doesn't really work for this case as it doesn't allow you (because of typing) to generate an incorrect structure, say a step after a branch etc. So this is really simulating some bad request coming from the frontend.

packages/server/src/api/routes/tests/automation.spec.ts
```diff
@@ -121,6 +123,78 @@ describe("/automations", () => {
...
+        { name: "test", condition: { equal: { "steps.1.success": "true" } } },
+      ]
+      automation.definition.steps[0].inputs.children = {}
+
+      const res = await request
+        .post(`/api/automations`)
+        .set(config.defaultHeaders())
+        .send(automation)
+        .expect("Content-Type", /json/)
+        .expect(400)
```
samwho, line 150: I'd add this post call to `packages/server/src/tests/utilities/api/automation.ts` and use the `Expectations` type you'll find reference to in there to assert the shape of the body (e.g. `{ message: "Branch steps are only allowed as the last step" }`. This should make these tests shorter and easier to parse, also more in line with the bulk of our other integration tests.

packages/server/src/api/routes/utils/validators.ts
```diff
@@ -88,7 +91,7 @@ export function datasourceValidator() {
   )
 }
 
-function filterObject() {
+function filterObject(unknown = true) {
```
samwho, line 94:
```suggestion
function filterObject(opts?: { unknown: boolean }) {
  const { unknown = true } = opts || {}
```

samwho: This just makes it a little bit easier to understand whehn this function is called, what this boolean is actually doing or referring to.

packages/server/src/api/routes/utils/validators.ts
```diff
@@ -267,13 +275,35 @@ function generateStepSchema(allowStepTypes: string[]) {
...
+    inputs: Joi.when("stepId", {
+      is: AutomationActionStepId.BRANCH,
+      then: Joi.object({
+        branches: Joi.array().items(branchSchema).min(1).required(),
+        children: Joi.object()
+          .pattern(Joi.string(), Joi.array().items(Joi.link("#step")))
+          .required(),
+      }).required(),
+      otherwise: Joi.object(),
+    }),
```
samwho, line 287: This is really cool. I had no idea you could do this.

packages/server/src/api/routes/utils/validators.ts
```diff
@@ -267,13 +275,35 @@ function generateStepSchema(allowStepTypes: string[]) {
...
 
+const validateStepsArray = (
+  steps: AutomationStep[],
+  helpers: Joi.CustomHelpers
+) => {
+  for (let i = 0; i < steps.length - 1; i++) {
+    if (steps[i].stepId === AutomationActionStepId.BRANCH) {
+      return helpers.error("branchStepPosition")
+    }
+  }
```
samwho, line 304:
```suggestion
  for (const step of steps.slice(0, -1)) {
    if (step.stepId === AutomationActionStepId.BRANCH) {
      return helpers.error("branchStepPosition")
    }
  }
```

samwho: I'm assuming here that the ignoring of the last element of the array is intentional. I almost missed it on my first readthrough. I think this slice way of doing it makes the intention much more explicit.

packages/server/src/tests/utilities/structures.ts
samwho: In a lot of other parts of the test code that we write, when we're creating invalid structures that TypeScript does not like, we typically just annotate them with a `@ts-ignore` comment. If that allows you to not add this structure here, then I think that's probably the way

packages/server/src/api/routes/utils/validators.ts
PClmnt: I know right, I noticed that Adria had done it elsewhere in the file so I copied him, can't take the credit :)

packages/server/src/tests/utilities/structures.ts
PClmnt: I actually misunderstood my own code, it's because on a branch step in the test builder we literally only allow a run function:

```
  branch(branchConfig: BranchConfig): {
    run: () => Promise<AutomationResults>
  } {
    this.addBranchStep(branchConfig)
    return {
      run: () => this.run(),
    }
  }

```

I can try and  this if you want to try and make it more flexible, what do you think?

PClmnt: I actually just updated the builder and the tests to support this @samwho

samwho (approved):

mike12345567 (approved):

PR #14511, opened 2024-09-04
Row action view security

adrinr:
## Description
Adding row action trigger security to only allow running a row action if the user triggering the action has read access to the given resource (either view or table).

Given that this is a critical function, the tests are testing all the possible security use case (any combination of roles, permissions on table, permission on views, and permissions on views via table inheritance)

## Launchcontrol
Adding extra row action trigger security


packages/server/src/middleware/triggerRowActionAuthorised.ts
```diff
@@ -1,40 +1,52 @@
...
+    )(ctx, () => {})
 
-    if (isTableId && !rowAction.permissions.table.runAllowed) {
+    // Check is the row action can run for the given view/table
+    const rowAction = await sdk.rowActions.get(tableId, rowActionId)
+    if (!viewId && !rowAction.permissions.table.runAllowed) {
       ctx.throw(
         403,
-        `Row action '${rowActionId}' is not enabled for table '${sourceId}'`
+        `Row action '${rowActionId}' is not enabled for table '${tableId}'`
```
aptkingston, line 44: Row actions are always allowed to be run against the table (assuming you have read access to the table, which I assume the `authorizedResource` takes care of), so I'm just curious what determines if this `table.runAllowed` flag is set or not?

aptkingston (approved):
LGTM! Just one question out of curiousity.

packages/server/src/middleware/triggerRowActionAuthorised.ts
adrinr: I created it in a way that could allow running for tables, but given that the UI does not support it, I will change it to be off by default

PR #14523, opened 2024-09-05
Drag and drop improvements

aptkingston:
## Description
This PR adds a few improvements to drag and drop.

- Restore the new screen placeholder which had vanished
image

- Dragging new components onto screens now lets you target the screen, and highlights it. Previously this didn't work at the screen level.

https://github.com/user-attachments/assets/14d1caf4-8562-4547-9746-4f4dcf2cf160

- Dragging new components now works with the new grid layout

https://github.com/user-attachments/assets/9eee7689-909e-4b52-8b62-abf473b8681a

- Enable empty states to work with grid layout. Previously empty containers were invisible. Now they have the normal empty state styles and placeholders - the same as flex.

https://github.com/user-attachments/assets/a7b37711-f09c-4643-a92b-fbc94e9e1268

- DND component placeholders are properly sized for both flex and grid layouts. They should always roughly reflect the size the real component will be when you add them. This works for mobile as well.


shogunpurple (approved):
LGTM - clean impl too - you actually removed code???

Issue #13820, opened 2024-05-30
Rows in a table go missing when multiple relationships are attached to a table.


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.27.5
  - App Version: 2.27.5

**Describe the bug**
Whenever multiple relationships are attached to a table, not all rows will be displayed. Additionally, you can't scroll down as you normally would to be able to get more rows. Also rows for relationship plus rows in the table don't display anywhere near the 5000 row limit that is set by default.

**To Reproduce**
1. Import app attached below as a new app
2. Create new postgres database
3. Import the tables
4. Run the insert queries to generate data for the tables
5. Reconfigure database to work with your new locally setup one one
6. Go to the builder and select the regions table 
7. See issue 10 region rows should be displayed, only one available.

**Expected behavior**
All rows within the table should be visible. A decent amount of relationship related rows should also be visible.

**Screenshots**
Builder row display for regions table
<img width="1211" alt="Screenshot 2024-05-30 at 12 40 01" src="https://github.com/Budibase/budibase/assets/126772285/51226ff4-e630-42b2-9f3a-3942a9bc5614">

Database row display for regions table
<img width="566" alt="Screenshot 2024-05-30 at 12 40 34" src="https://github.com/Budibase/budibase/assets/126772285/18a4fb72-468f-4ce5-b5f3-cba367d1134f">

Network request screenshot
<img width="734" alt="Screenshot 2024-05-30 at 13 26 22" src="https://github.com/Budibase/budibase/assets/126772285/98284118-2766-4ea3-a645-67c87ae4bc19">
  
**App Export**
[replicated-issue.tar.gz](https://github.com/Budibase/budibase/files/15499224/replicated-issue.tar.gz)

```mysql
CREATE TABLE regions (
    region_id SERIAL PRIMARY KEY,
    region_name VARCHAR(255) NOT NULL
);

CREATE TABLE countries (
    country_id SERIAL PRIMARY KEY,
    country_name VARCHAR(255) NOT NULL,
    region_id INT NOT NULL,
    FOREIGN KEY (region_id) REFERENCES regions(region_id)
);

CREATE TABLE cities (
    city_id SERIAL PRIMARY KEY,
    city_name VARCHAR(255) NOT NULL,
    region_id INT NOT NULL,
[...]

PR #14532, opened 2024-09-06
SQL - utilise JSON aggregations for relationships

mike12345567:
## Description
This PR addresses a number of issues with loading a large number of relationships in Budibase, primarily documented in: https://github.com/Budibase/budibase/issues/13820.

We also have a linear ticket: https://linear.app/budibase/issue/BUDI-8311/rows-in-a-table-go-missing-when-multiple-relationships-are-attached-to which discusses the change this PR implements, moving to use JSON aggregations to retrieve relationships rather than joining them onto the main table.

The issue with our old method is that as the number of many relationships grows, the size of the response data grows exponentially. For example if a table has 10 many relationships attached to it, each with 1000 rows related to a row in the main table, this is not a response size of 10001 rows (1000 rows for each of the 10 relationships, and 1 row for the main table row) - it is somewhere in the region of a response size 1000 to the power of 10. This is an un-avoidable issue with joining, usually this would be avoided by simply not joining everything, but that is not how Budibase works, we need all relationship data available for the page so that we can calculate formulas (amongst other requirements, like relational filtering etc).

Even if we do not return the full result set we still end up in this scenario, as the sheer amount of data getting joined can cause the query to never respond. We see this when working with SQS, we can cause threads to stop responding with heavily related data.

This fix means that the main table never has any joins on it - all joins are singular and part of sub-queries. This means that the response time will always scale O(n) - with the number of rows that need joined nothing more. The sub-queries can retrieve a large number of rows from the related table (up to our hard limit) and then merge these using JSON aggregation functions into a single column per row, this works well as this is what Budibase would normally do with the data set anyway.

This also fixes some issues we have with pagination - when we limit the main query to 100, we will always get 100 rows from the main table if there is that many rows - this improvement helps implement some other features.


packages/backend-core/src/sql/sql.ts
```diff
@@ -745,16 +855,222 @@ class InternalBuilder {
...
+        unaliased = parts.join(".")
+        tableField = this.quote(unaliased)
+      }
+      let separator = ","
+      switch (sqlClient) {
+        case SqlClient.ORACLE:
+          separator = " VALUE "
+          break
+        case SqlClient.MS_SQL:
+          separator = ":"
```
mike12345567, line 884: remove this.

packages/backend-core/src/sql/sql.ts
```diff
@@ -745,16 +855,222 @@ class InternalBuilder {
...
+            const to = relationship.to
+            // @ts-ignore
+            this.orOn(`${toAlias}.${toPrimary}`, `${throughAlias}.${to}`)
+          }
+        })
+    }
+    return query
+  }
+
+  addRelationships(
```
mike12345567, line 1069: Check if still used

packages/server/src/api/controllers/row/utils/basic.ts
```diff
@@ -101,6 +109,54 @@ export function basicProcessing({
...
+    })
+    const array: Row[] = Array.isArray(value)
+      ? value
+      : typeof value === "string"
+      ? JSON.parse(value)
+      : undefined
+    if (array) {
+      thisRow[col] = array
+      // make sure all of them have an _id
+      if (Array.isArray(thisRow[col])) {
```
mike12345567, line 135: Move this higher up

packages/server/src/api/routes/tests/queries/generic-sql.spec.ts
```diff
@@ -832,10 +832,12 @@ describe.each(
         },
       })
       expect(res).toHaveLength(1)
-      expect(res[0]).toEqual({
-        id: 2,
-        name: "two",
-      })
+      expect(res[0]).toEqual(
```
mike12345567, line 835: Check why this is needed

samwho (approved):
Went through this on a call with Mike just now, approval in principal. Just a few little things to look into that we've left in comments throughout.

packages/server/datasource-sha.env
```diff
@@ -1,4 +1,4 @@
-MSSQL_SHA=sha256:c4369c38385eba011c10906dc8892425831275bb035d5ce69656da8e29de50d8
+MSSQL_SHA=sha256:3b913841850a4d57fcfcb798be06acc88ea0f2acc5418bc0c140a43e91c4a545
```
adrinr, line 1: Why this update? Could we add a comment on what is the actual version of the sha?

adrinr (approved):

packages/server/datasource-sha.env
mike12345567: This was reducing our MS-SQL version as the latest versions have JSON aggregation functions, while older versions use a different syntax - we want to test against the minimum version we would want to support.

PR #14535, opened 2024-09-06
[BUDI-8579] Create more realistic tests for Google Sheets

samwho:
## Description

In this PR I've completely overhauled how we test our Google Sheets integration. I've created a new utility class called `GoogleSheetsMock` whose job it is to intercept Google Sheets API calls via `nock` and maintain the state of a mocked `Spreadsheet` within itself. This makes it a crude but useful simple reimplementation of the Google Sheets API that we can write tests against.

I'm about 90% sure in the process of doing this, and in https://github.com/Budibase/budibase/pull/14542, we've fixed the linked bug.

I want to do at least one follow-up PR adding a few more tests, mainly around deletes.

## Addresses
- https://linear.app/budibase/issue/BUDI-8579/issue-with-google-sheets

## Launchcontrol

Fixes several bugs around creating and updating rows in Google Sheets.


packages/server/src/integrations/googlesheets.ts
```diff
@@ -395,7 +397,12 @@ class GoogleSheetsIntegration implements DatasourcePlus {
           sheet,
         })
       case Operation.CREATE_TABLE:
-        return this.createTable(json?.table?.name)
+        if (json.table === undefined) {
```
adrinr, line 400: Any reason on why excluding null checks?

packages/server/src/integrations/tests/utils/googlesheets.ts
```diff
@@ -0,0 +1,774 @@
+// In this file is a mock implementation of the Google Sheets API.  It is used
```
adrinr, line 1: Could we get all these types from the google SDK, to detect if something changes in some npm update?

adrinr (approved):

packages/server/src/integrations/googlesheets.ts
samwho: No, I'll fix this. 🙏

mike12345567 (approved):
LGTM - massive improvement!

packages/server/src/integrations/googlesheets.ts
```diff
@@ -397,14 +399,19 @@ export class GoogleSheetsIntegration implements DatasourcePlus {
           sheet,
         })
       case Operation.CREATE_TABLE:
-        if (json.table === undefined) {
+        if (json.table == null) {
```
adrinr, line 402: Why not `if(!json.table)` to support both more natively?

adrinr (approved):
Nice to have this cleaner and cleaner every day :D

PR #14539, opened 2024-09-09
Fix relationship popover click behaviour

adrinr:
## Description
Fixing a small UX bug with the recently introduced popover showing the relationship fields. This new popup is absorbing outside clicks, and it causes issues when trying to remove a relationship.

#### Before
https://github.com/user-attachments/assets/f2b12627-1ab5-42ad-8aed-e62ec3693485

#### After

https://github.com/user-attachments/assets/80d9776a-2d6f-471d-bda7-ba90b399d3f7

### DIFF
--- a/packages/frontend-core/src/components/grid/overlays/GridPopover.svelte
+++ b/packages/frontend-core/src/components/grid/overlays/GridPopover.svelte
@@ -44,6 +44,7 @@
   {wrap}
   portalTarget="#{gridID} .grid-popover-container"
   offset={0}
+  clickOutsideOverride
 >
   <div
     class="grid-popover-contents"


aptkingston (approved):
Interesting... I had also made this change. But I'm now thinking it was when reviewing the original PR and I probably never said about it. But yep LGTM!

adrinr:
> Interesting... I had also made this change. But I'm now thinking it was when reviewing the original PR and I probably never said about it. But yep LGTM!

It might be in your redesign branch... I also remember talking about that 🤷🏽

adrinr:
> Interesting... I had also made this change. But I'm now thinking it was when reviewing the original PR and I probably never said about it. But yep LGTM!

Might it be in your redesign branch?

aptkingston:
> > Interesting... I had also made this change. But I'm now thinking it was when reviewing the original PR and I probably never said about it. But yep LGTM!
> 
> Might it be in your redesign branch?

Probably!

PR #14540, opened 2024-09-09
Fix long popover for relationship column pickers

adrinr:
## Description
Fixing a UX bug when having too many columns on a relationship field picker

### Before

https://github.com/user-attachments/assets/99c39d53-b71b-4ad1-8f5a-360676fa1e34

### After

https://github.com/user-attachments/assets/3751476c-cdb1-4a41-bdb2-c59d7d9b6a3c


aptkingston (approved):
I've added a new option specifically for nested popovers / context menus but it's on my V3 UI branch so we can't use it yet. Once we are doing the V3 merge we should update this to use that, but for now this is the best option 👍

adrinr:
> I've added a new option specifically for nested popovers / context menus but it's on my V3 UI branch so we can't use it yet. Once we are doing the V3 merge we should update this to use that, but for now this is the best option 👍

I will merge this to master and we can then use the new version on your branch :)

PR #14543, opened 2024-09-09
Part 1 of 2: Platform level AI Configuration

shogunpurple:
## Description
This PR adds the ability to create AI/LLM configurations that will be used to drive AI functionality across the Budibase platform. 

These configs are stored at the global DB (tenant) level, like other settings such as SMTP or OIDC. You can select your provider (OpenAI, Anthropic, Azure etc) and name your config, then set your API key and save. API keys are never exposed to the client after saving - the backend will return a placeholder that is used to tell the API that we should not overwrite API keys unless they are specifically overridden.

Matching pro PR for the quotas/licensing stuff: https://github.com/Budibase/budibase-pro/pull/318

Designs:
https://www.figma.com/proto/qEwKmsoVUrxQhvCKLuFQNp/Portal-Settings?page-id=0%3A1&node-id=120-2016&viewport=544%2C210%2C0.09&t=bg9PA7F9cDMI24UZ-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=11%3A571

## Addresses
https://linear.app/budibase/issue/BUDI-8608/ai-platform-level-config 

There are 2 parts to the above:
1) Allowing the user to create custom configs, and to meter them for license/quota purposes (This PR)
2) enabling anywhere in the platform that uses AI to actually make use of these configs in the DB (TBD - this will be another PR)

Thought it best to get config CRUD reviewed before plugging it into the wider automations piece.

## Test Plan
Quite a few different behaviours in this piece based on licensing so went through all the scenarios manually to make sure everything was working. Still working through these cases.
https://docs.google.com/spreadsheets/d/19KtVi6oMqOe6KkhXhdvZjKbjIGuOP1EB37SFS0H2H18/edit?gid=0#gid=0

## Screenshots
**Free users where AI is behind paywall**
Screenshot 2024-09-09 at 18 16 39

**Configs**
Screenshot 2024-09-09 at 18 11 26

**Creating a new config**
Screenshot 2024-09-09 at 18 12 16

## Launchcontrol
Allows the user to configure custom AI/LLM configurations that will drive their AI functionality in Budibase. Support for default Budibase AI config


packages/builder/src/pages/builder/portal/settings/ai/AIConfigTile.svelte
```diff
@@ -0,0 +1,130 @@
...
+  <div class="icon">
+    {#if config.name === "Budibase AI"}
+      <BudibaseLogo height="30" width="30" />
+    {:else if config.provider === Providers.OpenAI.name}
+      <OpenAILogo height="30" width="30" />
+    {:else if config.provider === Providers.Anthropic.name}
+      <AnthropicLogo height="30" width="30" />
+    {:else if config.provider === Providers.TogetherAI.name}
+      <TogetherAILogo height="30" width="30" />
+    {/if}
```
samwho, line 28: It feels like there ought to be a structure under `config.provider` that includes a `config.provider.logo` to make this a bit more robust to change.

packages/builder/src/pages/builder/portal/settings/ai/constants.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const ConfigMap = {
+  OpenAI: {
+    name: "OpenAI",
+    baseUrl: "https://api.openai.com",
+  },
+  Anthropic: {
+    name: "Anthropic",
+    baseUrl: "",
```
samwho, line 47: Are these going to be filled in later?

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -0,0 +1,171 @@
...
+    Tags,
+    Tag,
+  } from "@budibase/bbui"
+  import { admin, licensing } from "stores/portal"
+  import { API } from "api"
+  import AIConfigModal from "./ConfigModal.svelte"
+  import AIConfigTile from "./AIConfigTile.svelte"
+
+  const ConfigTypes = {
+    AI: "ai",
```
samwho, line 21: What other options do we anticipate in here?

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -0,0 +1,171 @@
...
+      await API.saveConfig(fullAIConfig)
+      notifications.success(`Successfully saved and activated AI Configuration`)
+      await fetchAIConfig()
+    } catch (error) {
+      notifications.error(
+        `Failed to save AI Configuration, reason: ${
+          error?.message || "Unknown"
+        }`
+      )
+    }
```
samwho, line 79: Given that `fetchAIConfig` does its own toast on failure and can never throw, I'd move it out of this try to improve clarity.

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -0,0 +1,171 @@
...
+  async function deleteConfig(key) {
+    // We don't store the default BB AI config in the DB
+    delete fullAIConfig.config.budibase_ai
+    // Delete the configuration
+    delete fullAIConfig.config[key]
+
+    try {
+      await API.saveConfig(fullAIConfig)
+      notifications.success(`Deleted config`)
+      await fetchAIConfig()
```
samwho, line 91: Same here.

packages/builder/src/pages/builder/portal/settings/ai/logos/Anthropic.svelte
```diff
@@ -0,0 +1,27 @@
+<script>
+  export let height
+  export let width
+</script>
+
+<svg
+  id="katman_1"
```
samwho, line 7: Do you need this `id` and `data-name`? They don't appear to be used within the SVG.

packages/builder/src/pages/builder/portal/settings/ai/logos/OpenAI.svelte
```diff
@@ -0,0 +1,18 @@
...
+  export let height
+  export let width
+</script>
+
+<svg
+  fill="#000000"
+  {width}
+  {height}
+  viewBox="0 0 24 24"
+  role="img"
```
samwho, line 11: Should this `role="img"` be on all of the SVGs in this PR?

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -197,6 +209,20 @@ async function verifyOIDCConfig(config: OIDCConfigs) {
...
 }
 
+export async function verifyAIConfig(
+  config: AIConfig,
+  existingConfig?: AIConfig
+) {
+  if (!existingConfig) return
+
+  // ensure that the redacted API keys are not overwritten in the DB
+  for (let uuid in existingConfig.config) {
```
samwho, line 219:
```suggestion
  for (const uuid in existingConfig.config) {
```

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -197,6 +209,20 @@ async function verifyOIDCConfig(config: OIDCConfigs) {
   await verifySSOConfig(ConfigType.OIDC, config.configs[0])
 }
 
+export async function verifyAIConfig(
+  config: AIConfig,
+  existingConfig?: AIConfig
+) {
+  if (!existingConfig) return
```
samwho, line 216: This is a bit weird. A caller could do `verifyAIConfig(config)` without realising it would never do anything at all. If `existingConfig` is required, it make it required as a type and throw an error if it's not provided. Think it makes more sense for this responsibility to be on the caller.

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -304,6 +333,30 @@ function enrichOIDCLogos(oidcLogos: OIDCLogosConfig) {
   )
 }
 
+async function enrichAIConfig(aiConfig: AIConfig) {
+  // Strip out the API Keys from the response so they don't show in the UI
+  for (let key in aiConfig.config) {
```
samwho, line 338:
```suggestion
  for (const key in aiConfig.config) {
```

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -304,6 +333,30 @@ function enrichOIDCLogos(oidcLogos: OIDCLogosConfig) {
...
+
+  // Return the Budibase AI data source as part of the response if licensing allows
+  const budibaseAIEnabled = await pro.features.isBudibaseAIEnabled()
+  const defaultConfigExists = Object.keys(aiConfig.config).some(key => aiConfig.config[key].isDefault)
+  if (budibaseAIEnabled) {
+    aiConfig.config["budibase_ai"] = {
+      provider: "OpenAI",
+      active: true,
+      isDefault: !defaultConfigExists,
+      defaultModel: env.BUDIBASE_AI_DEFAULT_MODEL,
```
samwho, line 352: Do we want some sort of validation in this env var in case someone changes it to something invalid?

packages/worker/src/api/controllers/global/tests/configs.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+import * as pro from "@budibase/pro"
+import { find, verifyAIConfig } from "../configs"
+
+jest.mock("@budibase/backend-core", () => ({
+  ...jest.requireActual("@budibase/backend-core"),
+  configs: {
+    getConfig: jest.fn(),
+    save: jest.fn(),
+  },
+}))
```
samwho, line 12: Are these configs saved into CouchDB? Would it be difficult to _not_ mock them?

packages/worker/src/api/controllers/global/tests/configs.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+        ai: {
+          apiKey: "--secret-value--",
+          baseUrl: "https://api.example.com",
+        },
+      },
+    })
+  })
+
+  it("Should return the default BB AI config when the feature is turned on", async () => {
+    pro.features.isBudibaseAIEnabled = jest.fn(() => true)
```
samwho, line 48: Does this assignment need to be reset when this test is finished?

packages/builder/src/pages/builder/portal/settings/ai/constants.ts
shogunpurple: Anthropic doesn't have a baseURL - it has it's own SDK so we will just use that if someone chooses Anthropic, the baseURL is mostly for OpenAI compatible providers

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
shogunpurple: None, was just to avoid putting the string directly

packages/worker/src/api/controllers/global/configs.ts
shogunpurple: I think it's ok, as it will be us who sets it and it can only be from the list of GPT models, so fairly low risk

packages/worker/src/api/controllers/global/tests/configs.spec.ts
shogunpurple: Updated to use spies, which will be cleaned up with `resetAllMocks`

packages/builder/src/pages/builder/portal/settings/ai/constants.ts
samwho: In that case do we want to use tagged unions here instead of having some fields be optional but only for some providers?

e.g. something like

```ts
interface OpenAIProvider {
  name: "OpenAI"
  baseUrl: string
}

interface AnthropicProvider {
  name: "Anthropic"
}

type Provider = OpenAIProvider | AnthropicProvider
```

packages/worker/src/api/controllers/global/tests/configs.spec.ts
shogunpurple: They are saved in couch, but the main purpose of this test was to just test the mapping code for secrets and the default Budibase AI config. I didn't think it was necessary to have a DB save step for testing this controller

samwho: We've generally been trying to test as much real code as possible, and avoiding mocks unless absolutely necessary. One of the motives is because it'll make any eventual move away from Jest easier, as other testing frameworks seem to do mocking differently.

It's not a huge deal, but if it's easy enough to actually do the DB save that would be preferable in the long term.

packages/builder/src/pages/builder/portal/settings/ai/AIConfigTile.svelte
shogunpurple: Updated to use `svelte:component` - unfortunately you can't just pull in through a string, so just defined the logos above as constants and at least we don't have to keep extending this `if` statement

shogunpurple:
> Just a small aesthetic point - the list along the left-hand side is getting a little long.
> 
> Is it worth adding groups, and alphabetically sorting within those groups?
> 
> Somewhat similar to what we do with components: Screenshot 2024-09-09 at 19 26 49

@melohagan I think we are probably ok - there's only 8 items. I've updated them to be in alphabetical order though to make it a bit easier

> Member

There's only 8 items so don't think we are at groups level yet, but I've sorted them alphabetically!

melohagan:
> @melohagan I think we are probably ok - there's only 8 items. I've updated them to be in alphabetical order though to make it a bit easier

Putting them alphabetically was the main thing - perfect!

samwho (approved):

PR #14548, opened 2024-09-10
More google sheets tests

samwho:
## Description

Made the `GoogleSheetsMock` a bit more true to the actual API behaviour by "trimming" `ValueRange`s that come back from the API. I observed that this is how Google Sheets API works in practice, so I've mimicked it here and built more tests on top of it as a result.


mike12345567 (approved):
Very nice!

PR #14552, opened 2024-09-11
Pass accountId as param instead of Account type

melohagan:
## Description
Having some type issues in account-portal, and this helps out. 

Ideally these types should be deleted from here and moved into account-portal itself as it's the only place that uses this code.


jvcalderon (approved):
LGTM

PR #14554, opened 2024-09-11
Update automations to default to using names for accessing context

PClmnt:
## Description
This PR updates both the backend and the frontend of the automations system to ensure that the primary method of accessing automation context is using names instead of numerical indexing. 

So the method of accessing a step result moves from `steps.1.success` -> `steps.MyLogStep.success`. 

Previously we just had a frontend concept of this where the name would be shown in the inputs but the actual runtime binding would still contain the index. This updates it so both the runtime and readable binding will be named, this is in support of branching where multiple branches could result in multiple `steps.1.success` for example. On the backend this is done by creating a new context called `stepsByName` which we pass to handlebars as well as the standard context, that way if something is being accessed by name or by number index it will compile both correctly. 

This also has the nice side effect of being able to provide complete binding updates across the step definitions wherever they are. i.e If you change the name of your first step this will update the binding name all the way down in step 7. 

**_This only can apply to new blocks that are added_**. Existing blocks when bound will simply use the name of the automation. i.e binding an existing block to another step will now be `steps.Create Row.success` and will use `steps.3.row` as the runtime. This is okay for automations that exist before branching as there can only be one path. So if a branch gets added the context will still be correct

Also, if a user adds a new block, it will simply be given a name which is +1 of any existing blocks of that same type. i.e 2 `Backend Log` blocks exist, user adds a new block. It is named `Backend Log 3`
`
## Addresses
https://linear.app/budibase/issue/BUDI-8601/use-step-names-for-accessing-automation-context

## Video

https://www.loom.com/share/f9586f05ee40492b9c037439b977d112?sid=04f5a02d-c157-474c-b35e-68a73ec21c76
## Launchcontrol
- Updates the automation system to be primarily name based. Automation step outputs are now accessed using the name of an automation step, rather than the step index. 

## Feature branch env
[Feature Branch Link](http://fb-feat-automation-step-naming-updates.fb.qa.budibase.net)


packages/builder/src/helpers/tests/nameHelpers.spec.js
```diff
@@ -0,0 +1,177 @@
+import { cloneDeep } from "lodash"
+import {
+  updateBindingsInInputs,
+  updateBindingsInSteps,
+} from "../automations/nameHelpers"
+describe("Automation Binding Update Functions", () => {
```
mike12345567, line 6: Nice to see this - useful case for a unit test!

packages/server/src/automations/tests/scenarios/scenarios.spec.ts
```diff
@@ -49,151 +49,188 @@ describe("Automation Scenarios", () => {
...
+        name: "Test Row Save and Create",
       })
-      .run()
 
-    expect(results.steps).toHaveLength(1)
-    expect(results.steps[0].outputs.rows).toHaveLength(2)
-  })
+      const results = await builder
+        .appAction({ fields: {} })
+        .queryRows({
```
mike12345567, line 68: we should build a test scenario that loops over the results of a query rows - not for this PR, but just got me thinking as I see it! Feels like one of the most common use cases for it.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -99,66 +105,74 @@ class BaseStepBuilder {
   }
 
   // STEPS
-  createRow(inputs: CreateRowStepInputs): this {
+  createRow(inputs: CreateRowStepInputs, stepName?: string): this {
```
mike12345567, line 108: Should the step name be within an object specifier? E.g. `opts?: { name?: string }` - that way if there are further optional parameters in future they are easy to add without needing to update everywhere these are used.

packages/types/src/documents/app/automation/automation.ts
```diff
@@ -124,6 +124,7 @@ export interface Automation extends Document {
   definition: {
     steps: AutomationStep[]
     trigger: AutomationTrigger
+    stepNames?: Record<string, string>
```
mike12345567, line 127: What is this a map of? Would be good to comment to explain what it does.

PClmnt: {
    "0DepA21qN": "My Step Name"
}

Key is the ID of the step. For easy lookup. I'll get a comment in for it.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
PClmnt: Sorted 👍

packages/server/src/automations/tests/scenarios/scenarios.spec.ts
PClmnt: I've got another branch with a few tests on it and this is one of them, definitely an important one to get covered.

mike12345567 (approved):

PR #14555, opened 2024-09-11
onEmptyFilter fixes for logical operators

mike12345567:
## Description
Test cases for return none/all with the logical operators, as well as updates to the `hasFilter` to support this correctly.

This changes the `hasFilters` to be properly aware of logical operators, making sure it can recurse as is needed to get through the `$and` and `$or` groups.

Also adds test cases for this - empty filters with `all` and `none`.


deanhannigan (approved):
LGTM! 👍

PR #14557, opened 2024-09-11
SQL - use CTEs to avoid pagination issues with large tables

mike12345567:
## Description
As the title - this updates our SQL generation to use a CTE, avoiding issues when attempting to jump to large page offsets.

This is required as the `OFFSET` parameter doesn't stop the rows before the offset from being selected, this includes our JSON aggregation and therefore this can perform poorly.

The CTE correctly picks out the rows that we are going to page on (including using our `WHERE EXISTS` step to filter the relationships) then selects the aggregations just for those rows.

This was not initially implemented as SQS could not support it yet.


PClmnt (approved):
Code LGTM, although I have not so much context on the overall goal

PR #14562, opened 2024-09-11
Update logic when no branch condition is met

PClmnt:
## Description
As I was writing a test for this I realised the behaviour wasn't exactly correct. So adding in a specific case that when no branch conditions are met we return with a status. Added a test to cover this eventuality as well. 

Big delta for this because i split out the branching and looping tests from the main automation scenarios file.


packages/server/src/automations/tests/scenarios/branching.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+            },
+          },
+        },
+      })
+      .run()
+
+    expect(results.steps[1].outputs.message).toContain("Special user")
+  })
+
+  it("should stop the branch automation when no conditions are met", async () => {
```
PClmnt, line 159: This is the only new test

mike12345567 (approved):

PR #14564, opened 2024-09-11
updating budibase apps image to use alpine

shogunpurple:
## Description
Updating our `budibase/apps` image to use alpine instead of `slim`. This is part of the ongoing work to reduce vulnerabilities in our image for an enterprise prospect. Following this merge, I can use the new alpine based hardened image as the base image for both apps and worker.

**Steps:**
- converted server Dockerfile and built it locally
- Used the new alpine based dockerfile in a docker compose configuration. Works as expected
- `docker image ls` shows that this image is now `880MB` instead of the standard `budibase/apps` 1.16GB, so a nice saving just for updating to alpine

## Addresses

### DIFF
--- a/packages/server/Dockerfile
+++ b/packages/server/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-slim
+FROM node:20-alpine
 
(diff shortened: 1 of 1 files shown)


calexiou (approved):
lgtm!

PR #14570, opened 2024-09-12
Role metadata and view controls

mike12345567:
## Description
Taking working from `new-rbac-ui` branch and separating it into its own PR, so that other work can be based on this from master.

This work includes two parts:
1. The ability to add UI metadata to roles, which can be used in the frontend representation of them
2. Views role control is now free/accessible by all

This work has largely been completed on @aptkingston branch, but bringing it into master so further role work (around multi-inheritance) can be based on it.

Tests to follow - as part of further PRs once this has all been merged up the chain of new UI branches.


aptkingston (approved):
LGTM! Thanks for pulling this out.

PR #14572, opened 2024-09-12
updating JS dependencies to get rid of vulns

shogunpurple:
## Description
- Further fixes for worker and app service vulnerabilities. Mainly these just consist of going through the list on the scanner, looking at the CVEs online (eg. https://nvd.nist.gov/vuln/detail/cve-2022-0235) and updating to the patched versions.)

## Addresses

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -63,7 +63,7 @@
     "@koa/router": "8.0.8",
     "@socket.io/redis-adapter": "^8.2.1",
     "@types/xml2js": "^0.4.14",
-    "airtable": "0.10.1",
+    "airtable": "^0.12.2",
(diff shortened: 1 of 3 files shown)


packages/server/package.json
```diff
@@ -63,7 +63,7 @@
     "@koa/router": "8.0.8",
     "@socket.io/redis-adapter": "^8.2.1",
     "@types/xml2js": "^0.4.14",
-    "airtable": "0.10.1",
+    "airtable": "^0.12.2",
```
mike12345567, line 66: we should likely pin this, as we did with the old dependency.

mike12345567 (approved):
LGTM!

PR #14579, opened 2024-09-13
Role name and display name updates

aptkingston:
## Description
Small update to the recent role work.

- Don't use display name as name, since this can contain spaces and results in an illegal name being stored in the doc
- The old UI will therefore function as it always has, and the new UI will send up role names as UUIDs (which will also create proper UUID _ids as well) and display names as the readable name
- Restore usual names for built in roles (ADMIN, BASIC etc) rather than the new display names which again contain spaces and are only part of the new UI


mike12345567 (approved):
LGTM!

PR #14586, opened 2024-09-16
More google sheets tests

samwho:
## Description

Last batch of Google Sheets tests: updating and deleting stuff. This will be the last set of tests for the time being, we have the basic CRUD operations covered in a much more realistic way than they were before. Should stave off some of the most simple regressions for a while.


mike12345567 (approved):
LGTM!

PR #14588, opened 2024-09-16
Fix Postgres "cannot pass more than 100 arguments" issue

mike12345567:
## Description
Adding function parameter limit control for different SQL DBs, every DB has different limits with Postgres being the lowest at 100. We need to fix for wide tables which are related.


samwho (approved):
Can we add tests for this?

PR #14589, opened 2024-09-16
Fix Google Sheets pagination.

samwho:
## Description

Pagination was broken for Google Sheets because at some point we must have stopped using the `query.pagination.page` param and instead started passing `limit` and `offset` directly (I vaguely remember us doing this as part of the SQS work.)

I've updated the code to support the passing of either limit/page or limit/offset, and added a test to prevent regression.

## Launchcontrol

- Fixes an issue with pagination in the Google Sheets integration.


mike12345567 (approved):
LGTM!

PR #14597, opened 2024-09-18
Default view permissions to parent table roles

aptkingston:
## Description
This PR updates view creation so that the read and write roles get set to the explicit read and write roles from the parent table. This is required as V3 is doing away with the concept of inheriting access from tables.

This PR basically consists of:
- Moving some permission code from the controller into the SDK so we can use it from multiple places
- Updating views upon creation to set explicit permissions matching their parent table


packages/server/src/api/routes/tests/permissions.spec.ts
```diff
@@ -125,6 +125,12 @@ describe("/permission", () => {
     })
 
     it("should be able to access the view data when the table is set to public and with no view permissions overrides", async () => {
+      // Make view inherit table permissions
+      await config.api.permission.revoke({
+        roleId: STD_ROLE_ID,
+        resourceId: view.id,
+        level: PermissionLevel.READ,
+      })
```
samwho, line 133: I'm a teensy bit confused, let me see if I understand.

Before this PR, views inherit the permissions set on their underlying table. I assume there exists code that does those checks somewhere outside of this PR.

In this PR, on view creation, you are setting permissions on the view explicitly, based on permissions on the table.

What is confusing me is why you need to do these revokes in the tests. I would have thought the permission on the views is identical to how it would have been before your changes? 🤔

aptkingston: You're correct. These additional revokes are just needed because the old behaviour was to inherit access, while now it defaults to having explicit permissions. These tests were deliberately testing inheritance, so I need to revoke the explicit permissions in order to continue to test inheritance.

samwho (approved):

aptkingston:
Just before you merge this -  I have a table with 10K rows and 2 identical views, one with related columns shown and one without, but otherwise identical. The one without related columns can do a search call in about 200ms, but the one with them takes 15-20 seconds. 

This is reliable reproducible by creating a new view (fast) then showing even a single related column (extremely slow).

Issue #14595, opened 2024-09-18
Automation - delete row not triggering


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 2.32.5
  - App Version: 2.32.5

**Describe the bug**
When there is an automation which has Row Deleted as the trigger the automation does not trigger when a row is deleted.

**To Reproduce**
1. Create an app with a internal DB table
2. Create an interface where you can add and delete from the table
3. Add an automation with row deleted on that table
4. Publish the app
5. Create and delete a row
6. View automation logs

**Expected behavior**
The automation should trigger and show in the automation logs. It does neither.

andyburgessmd:
@mikesealey I have sent the export to your support email address.

mikesealey:
@andyburgessmd Thanks for providing your app-export. I took this on a call with some of the engineering team, who are currently working on a fix.

andyburgessmd:
@mikesealey thanks, fixed now :)

PR #14601, opened 2024-09-18
Add table ID to row deletion requests that only contain string IDs

aptkingston:
## Description
Fixes an issue where the row deletion trigger does not work when the bulk request rows endpoint is used with only string IDs. The endpoint accepts an array of either rows or row IDs, and when row IDs are passed they are not enriched enough to trigger row deletion automations, which depend on a table ID property being present.

This PR adds the table ID property to the enrich row objects created when only IDs are passed in.

Also added a test to cover this case.

## Addresses
- https://github.com/Budibase/budibase/issues/14595#issuecomment-2358530978


samwho (approved):
Great work, thanks for jumping on this <3

PR #14602, opened 2024-09-18
Part 2 of 2 - Budi 8608 ai platform level config

shogunpurple:
## Description
This PR allows budibase to inject the AI configs that can be created with https://github.com/Budibase/budibase/pull/14543 into LLM providers. Right now, the only place we have AI in the platform is in automations, so support is provided there to allow users who already use the OpenAI integration to continue as is, but users who have access to Budibase AI on paid plans can use custom configs, such as Anthropic instead of OpenAI.

All the LLM code (a majority of this work) is wrapped up in this budibase pro repo. There's further context there on how the LLM integration works, but I tried to keep it as abstracted as possible (including reading the config etc) in pro, so it can just be pulled in and used in a few lines anywhere in the platform that we need AI, and has a dependency on `@budibase/pro`.
https://github.com/Budibase/budibase-pro/pull/320

## Addresses
https://linear.app/budibase/issue/BUDI-8608/ai-platform-level-config

There are 2 parts to the above:

- Allowing the user to create custom configs, and to meter them for license/quota purposes (Done - https://github.com/Budibase/budibase/pull/14543)
- enabling anywhere in the platform that uses AI to actually make use of these configs in the DB (This PR)

## Launchcontrol
Add the ability to use custom AI configs and the built in budibase AI in cloud for automations


packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -56,10 +56,13 @@
     } else {
       // We don't store the default BB AI config in the DB
       delete fullAIConfig.config.budibase_ai
+
       // unset the default value from other configs if default is set
       if (editingAIConfig.isDefault) {
         for (let key in fullAIConfig.config) {
-          fullAIConfig.config[key].isDefault = false
+          if (key !== id) {
```
samwho, line 63: I don't quite understand this change. The old code looks correct, what does this fix?

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -372,7 +347,13 @@ export async function find(ctx: UserCtx) {
...
 
       if (type === ConfigType.AI) {
-        await enrichAIConfig(scopedConfig)
+        await pro.sdk.ai.enrichAIConfig(scopedConfig)
+        // Strip out the API Keys from the response so they don't show in the UI
+        for (const key in scopedConfig.config) {
+          if (scopedConfig.config[key].apiKey) {
+            scopedConfig.config[key].apiKey = PASSWORD_REPLACEMENT
+          }
+        }
```
samwho, line 356: We need a test to prevent us ever breaking this behaviour by accident.

samwho (changes requested):

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
shogunpurple: Since it iterates through all the configs, It unsets the new one you actually want to set to default back to `false`, so it needs to know not to do that

samwho: Ahhh damn, I missed that. Good catch.

samwho (approved):

Issue #14605, opened 2024-09-19
Update Setting "Is In"


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
SELF & CLOUD
- Self
  - Method: docker compose
  - Budibase Version: 2.32.2
- Cloud
  - Tenant ID:  miketest1 - 2.32.5

**Describe the bug**
Following a discussion https://github.com/Budibase/budibase/discussions/14604 - the "Update Setting" condition 'Is In" erroneously returns true when compared to an empty string, or a `null` value

**To Reproduce**
1. Add a headline with text "False"
2. Add a form with two text fields
3. Add a condition `Update Setting Text to TRUE IF {{ New Form.Fields.this_value }} is in binding {{ New Form.Fields.that_value }}`
4. In preview or published app, add "1" to the first field, and "1,2,3" to the second to see "true" then make the second field empty - it also returns true. "2,3,4" will return false as expected.

The following values should return false, but instead return true - null, "", []

**Expected behavior**
Empty strings, empty text field, empty array, and `null` should all return "false"

**Screenshots**
See app-export below
![image](https://github.com/user-attachments/assets/96886c3f-be22-48e2-b3ed-7846a842421e)

**Screencapture**
https://jam.dev/c/ab9f539d-3b78-40f9-a566-d0c43b5e416d

**App Export**
[14604-export-1726728104323.tar.gz](https://github.com/user-attachments/files/17054702/14604-export-1726728104323.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 128.0.6613.114 (Official Build) (arm64)

**Additional context**
Related: https://github.com/Budibase/budibase/discussions/14604

aptkingston:
After reading this I had an inclination that this probably affects all operators rather than just "is in" and it looks like that's correct. The reason for this is due to the "on empty filter" operator which is designed to return all rows when all filters are empty. We use this same code behind the scenes to evaluate conditions, but this behaviour means that when your condition value is "empty" it effectively skips it and returns true.

So this same thing happens if you just do "equals" with an empty value too.

The fix is to ensure that we use "none" for the "on empty filter" parameter of our query when evaluating conditions in the client library. This is an easy fix 👌

aptkingston:
Threw up a quick fix!

PR #14606, opened 2024-09-19
Fix conditions with empty values falsely evaluating to true

aptkingston:
## Description
Fixes an issue when evaluating conditions which meant that empty values were falsely treated as the condition passing.

Before - empty value matching the condition and the new text setting being incorrectly applied:
image

After - not matching on an empty value:
image

After - matching when the condition is actually true:
image

## Addresses
- https://github.com/Budibase/budibase/issues/14605


deanhannigan (approved):
LGTM!

PR #14607, opened 2024-09-19
Require a flag to be set to query PostHog for feature flags.

samwho:


packages/backend-core/src/environment.ts
```diff
@@ -143,6 +143,7 @@ const environment = {
   POSTHOG_TOKEN: process.env.POSTHOG_TOKEN,
   POSTHOG_PERSONAL_TOKEN: process.env.POSTHOG_PERSONAL_TOKEN,
   POSTHOG_API_HOST: process.env.POSTHOG_API_HOST || "https://us.i.posthog.com",
+  POSTHOG_FEATURE_FLAGS: process.env.POSTHOG_FEATURE_FLAGS,
```
shogunpurple, line 146: NAB: Might be better to call this `POSTHOG_FEATURE_FLAGS_ENABLED` or something since it's a boolean flag

shogunpurple (approved):
LGTM

PR #14610, opened 2024-09-19
`fetchSchema` tests for Google Sheets integration

samwho:
## Description

This PR adds tests and a small fix for a problem with `fetchSchema` in the Google Sheets integration. It looks like we weren't correctly handling the situation when no header row was found, partly because there was an unawaited promise (fixed in https://github.com/Budibase/node-google-spreadsheet/pull/3) and partly because we weren't matching on the right error from the library we use.


shogunpurple (approved):
Nice work, this has been an issue for a while!

PR #14613, opened 2024-09-20
Shrink the size of the attachment fields set to "compact"

mikesealey:
## Description
The existing setting "compact" for the attachment fields shrinks the "drag and drop" zone, but has very little effect on the file preview window, which remains quite large, which is particularly noticeable for Multi-Attachment fields where the "drag and drop" zone always displays, as well as the file preview.

This PR shrinks the size of the file preview window, offering a more appreciable difference when using the "compact" setting. 

## Addresses
BUDI-8588 Multi-attachment field "compact" setting not compact enough

## Screenshots
**Before Changes**
**Data Section Before Changes**
Multi Attachment All Images, All Files, and Mixed Images-and-Files
Screen Recording 2024-09-24 at 13 53 10
**Data Section After Changes**
Screen Recording 2024-09-24 at 14 22 06

**Published App Single Attachment Field (with images) Before Changes**
image

**After changes scrolling through multi-attachment field files**
Screen Recording 2024-09-24 at 14 17 28

**After changes scrolling through multi-attachment field files and images**
Screen Recording 2024-09-24 at 14 14 36

**Considerations**

## Launchcontrol

_Shrink the size of attachment fields set to "compact"._


aptkingston:
How does this look when using an attachment that does support previewing (images basically). The preview was kept roughly the same size so you're still able to actually see the image previews, which I don't think you'd be able to if shrunk to this size.

The other thing to check is if this change affects tables or not (including in the data section) as I think they use a compact attachment component.

mikesealey:
> How does this look when using an attachment that does support previewing (images basically)? The preview was kept roughly the same size so you're still able to actually see the image previews, which I don't think you'd be able to if shrunk to this size.

Images
image

Files
image

> 
> The other thing to check is if this change affects tables or not (including in the data section) as I think they use a compact attachment component.
Data tab images
image


Data tab files
image

aptkingston (approved):
Thanks for the images! LGTM!

PR #14614, opened 2024-09-20
Update view handling of filters

aptkingston:
## Description
This PR updates how views handle filters and when they're attached to requests. This fixes some pre-existing issues and some newly introduced ones, that resulted in things like double queries and attaching view filters in the search queries (which is unnecessary as they are saved into the view definition).

### DIFF
--- a/packages/frontend-core/src/components/grid/stores/filter.js
+++ b/packages/frontend-core/src/components/grid/stores/filter.js
@@ -1,4 +1,4 @@
-import { writable, get, derived } from "svelte/store"
+import { get, derived } from "svelte/store"
 import { FieldType, FilterGroupLogicalOperator } from "@budibase/types"
 import { memo } from "../../../utils/memo"
 
--- a/packages/frontend-core/src/components/grid/stores/filter.js
+++ b/packages/frontend-core/src/components/grid/stores/filter.js
@@ -1,12 +1,13 @@
 import { writable, get, derived } from "svelte/store"
 import { FieldType, FilterGroupLogicalOperator } from "@budibase/types"
+import { memo } from "../../../utils/memo"
 
 export const createStores = context => {
   const { props } = context
 
   // Initialise to default props
-  const filter = writable(get(props).initialFilter)
-  const inlineFilters = writable([])
+  const filter = memo(get(props).initialFilter)
+  const inlineFilters = memo([])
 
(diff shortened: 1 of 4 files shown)


packages/frontend-core/src/components/grid/stores/filter.js
```diff
@@ -19,19 +20,26 @@ export const deriveStores = context => {
...
+          },
+        ],
+      }
+      // Just use inline if no filter
+      if (!$filter?.groups?.length) {
+        return allFilters
+      }
+      // Join them together if both
+      allFilters.groups = [...allFilters.groups, ...$filter.groups]
+      return allFilters
```
aptkingston, line 42: Just rewrote this to be more explicit and construct the minimal complexity of query possible.

packages/frontend-core/src/fetch/DataFetch.js
```diff
@@ -365,7 +364,9 @@ export default class DataFetch {
     let refresh = false
     const entries = Object.entries(newOptions || {})
     for (let [key, value] of entries) {
-      if (JSON.stringify(value) !== JSON.stringify(this.options[key])) {
+      const oldVal = this.options[key] == null ? null : this.options[key]
+      const newVal = value == null ? null : value
```
aptkingston, line 368: Added this in as `undefined` and `null` were treating as being different and therefore causing a fetch to refresh all data.

packages/frontend-core/src/fetch/ViewV2Fetch.js
```diff
@@ -55,12 +48,6 @@ export default class ViewV2Fetch extends DataFetch {
       this.options.sortOrder = definition.sort.order
     }
 
-    const parsed = utils.processSearchFilters(filter)
-
-    if (!parsed?.groups?.length && definition.query?.groups?.length) {
-      this.options.filter = definition.query
-    }
```
aptkingston, line 62: We never want to set the filter to the query definition, as that is redundant and causes the filters to be sent up in the search request.

deanhannigan (approved):
LGTM!

packages/frontend-core/src/fetch/DataFetch.js
deanhannigan: Nice catch!

PR #14627, opened 2024-09-23
Fixing AI test cases.

mike12345567:
## Description
As title - fixing a test case which was failing in `server/src/automations/tests/openai.spec.ts`

### DIFF
--- a/packages/server/src/automations/tests/openai.spec.ts
+++ b/packages/server/src/automations/tests/openai.spec.ts
@@ -23,14 +23,15 @@ jest.mock("openai", () => ({
     },
   })),
 }))
-
 jest.mock("@budibase/pro", () => ({
   ...jest.requireActual("@budibase/pro"),
   ai: {
-    LargeLanguageModel: jest.fn().mockImplementation(() => ({
-      init: jest.fn(),
-      run: jest.fn(),
-    })),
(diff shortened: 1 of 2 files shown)


packages/server/src/automations/tests/openai.spec.ts
```diff
@@ -121,11 +123,14 @@ describe("test the openai action", () => {
...
+      "gpt-4o-mini"
+    )
 
-    // @ts-ignore
-    const llmInstance = pro.ai.LargeLanguageModel.mock.results[0].value
-    expect(llmInstance.init).toHaveBeenCalled()
+    const llmInstance =
+      mockedPro.ai.LargeLanguageModel.forCurrentTenant.mock.results[0].value
+    // init does not appear to be called currently
+    // expect(llmInstance.init).toHaveBeenCalled()
```
samwho, line 133: Because forCurrentTenant is mocked, which usually calls it.

samwho (approved):
Have I ever mentioned that I hate mocks?

mike12345567:
> Have I ever mentioned that I hate mocks?

Agreed - I'd like to see this test through the new automation framework, if we must mock I think we can just mock the `anthropic/sdk` and thats it.

PR #14628, opened 2024-09-23
Initial passing test for view calculations.

samwho:
## Description

This PR focuses around a single new test: https://github.com/Budibase/budibase/blob/e760dc4dc6cd1a0857a7ed09fcee32b93ea7a8f8/packages/server/src/api/routes/tests/viewV2.spec.ts#L2395-L2426

For this to work, I have had to break through 2 fundamental assumptions about how Budibase works:

1. All rows returned from any database have a unique identifier.
2. All fields on a row can be found in its table's schema.

There are _many_ places in the code base that assume these two things, and more that _enforce_ them. They all needed to be changed because view calculation columns will never be on the table schema, and rows from a view calculation query represent an aggregate operation over many other rows.


packages/backend-core/src/sql/sql.ts
```diff
@@ -1236,10 +1270,18 @@ class InternalBuilder {
...
-    query = !counting
-      ? query.select(this.generateSelectStatement())
-      : this.addDistinctCount(query)
+    const aggregations = this.query.resource?.aggregations || []
+    if (counting) {
+      query = this.addDistinctCount(query)
+    } else if (aggregations.length > 0) {
+      query = query.select(
+        this.knex.raw("ROW_NUMBER() OVER (ORDER BY (SELECT 0)) as _id")
+      )
```
samwho, line 1279: This is to give the aggregation rows a unique ID. It's a hack for the rest of the codebase because we make some assumptions that rows have a unique ID. I think the frontend uses it to know what rows to update when updates come through the websockets, though that will never happen for these rows.

This plus the large number of other changes in this PR make me think it's all a bit _too_ hacky and we should make more explicit affordances for "aggregation rows" (rows that do not have an ID).

packages/server/src/sdk/app/rows/search/internal/sqs.ts
```diff
@@ -294,7 +295,7 @@ export async function search(
   table: Table,
   opts?: { retrying?: boolean }
 ): Promise<SearchResponse<Row>> {
-  let { paginate, query, ...params } = options
+  let { paginate, query, ...params } = cloneDeep(options)
```
samwho, line 298: This `cloneDeep` is to avoid mutations to `options` percolating up and out of this function call.

packages/server/src/sdk/app/rows/search/internal/sqs.ts
```diff
@@ -425,6 +445,6 @@ export async function search(
     if (err.status === 400 && msg?.match(MISSING_COLUMN_REGEX)) {
       return { rows: [] }
     }
-    throw new Error(`Unable to search by SQL - ${msg}`, { cause: err })
+    throw err
```
samwho, line 448: Think I changed this to make the stack trace nicer in the tests, it'll get put back before merging.

packages/types/src/documents/app/view.ts
```diff
@@ -33,15 +33,24 @@ export interface View {
...
+export interface RelationSchemaField extends UIFieldMetadata {
   readonly?: boolean
 }
 
+export interface ViewCalculationFieldMetadata extends BasicViewUIFieldMetadata {
+  calculationType: CalculationType
+  field: string
+}
+
+export type ViewUIFieldMetadata =
```
samwho, line 50: I have a feeling when I started this PR, this type was called `ViewUIFieldMetadata` and somewhere it got changed to `ViewFieldMetadata` but that got lost. I didn't add `UI` to this name in this PR.

packages/server/src/api/controllers/row/utils/basic.ts
```diff
@@ -108,6 +117,11 @@ export function basicProcessing({
       thisRow[fieldName] = value
     }
   }
+
+  for (let aggregation of aggregations || []) {
+    thisRow[aggregation.name] = row[aggregation.name]
+  }
```
samwho, line 123: Almost everything in this PR is in service of a few bits of code that look like this. We plumb `aggregations` all the way to the bottom of the SQL layer in order to slightly change the SQL query, and then make sure those fields get pulled out and don't get left behind by the several places that make sure _only_ schema fields get pushed back out through the API. It feels very messy to me.

packages/types/src/sdk/row.ts
```diff
@@ -19,6 +25,7 @@ export interface SearchParams {
   indexer?: () => Promise<any>
   rows?: Row[]
   countRows?: boolean
+  aggregations?: Aggregation[]
```
samwho, line 28: This is the API that takes us from the controller level into the SDK level. Everything below the SDK level has to interrogate this field and change behaviour based on it.

packages/types/src/sdk/search.ts
```diff
@@ -154,6 +155,7 @@ export interface QueryJson {
   }
   resource?: {
     fields: string[]
+    aggregations?: Aggregation[]
```
samwho, line 158: Then this is the API that takes us from the SDK layer into the SQL layer. It being a nested optional field feels really weird but it made sense to me to have it live next to `fields`.

packages/backend-core/src/sql/sql.ts
mike12345567: I think while this is hacky its a pretty good idea to get the DB to allocate these numbers, I'm not against continuing down this path.

There is a beneficial piece of work we could do that would make this easier however, right now all tables are assumed to have full CRUD capabilities. If we had the ability to make tables as "read-only" then they wouldn't need to have primary keys - a more significant project but this would also unlock the ability to import views to Budibase which are read-only and don't have primary keys.

packages/backend-core/src/context/mainContext.ts
```diff
@@ -394,3 +394,20 @@ export function setFeatureFlags(key: string, value: Record<string, any>) {
...
+}
+
+export function setTableForView(viewId: string, table: Table) {
+  const context = getCurrentContext()
+  if (!context) {
+    return
+  }
+  context.viewToTableCache ??= {}
+  context.viewToTableCache[viewId] = table
+}
```
samwho, line 413: Because we plan to use the `ViewV2` to get to the `Table` lot of times per request, I've decided to cache that lookup in the context so it's free every time after the first. This is nicer than passing both a `Table` and a `ViewV2` around everywhere.

packages/backend-core/src/docIds/params.ts
```diff
@@ -77,13 +76,16 @@ export const isTableId = (id: string) => {
...
+
 /**
  * Check if a given ID is that of a datasource or datasource plus.
- * @returns {boolean}
  */
-export const isDatasourceId = (id: string) => {
+export const isDatasourceId = (id: string): boolean => {
   // this covers both datasources and datasource plus
-  return id && id.startsWith(`${DocumentType.DATASOURCE}${SEPARATOR}`)
+  return !!id && id.startsWith(`${DocumentType.DATASOURCE}${SEPARATOR}`)
```
samwho, line 88: Noticed this while I was in the file, this was returning `string | boolean` probably unintentionally.

packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -159,17 +161,41 @@ function isEditableColumn(column: FieldSchema) {
 
 export class ExternalRequest<T extends Operation> {
   private readonly operation: T
-  private readonly tableId: string
-  private datasource?: Datasource
-  private tables: { [key: string]: Table } = {}
+  private readonly source: Table | ViewV2
+  private datasource: Datasource
```
samwho, line 165: Changing how instances of this class are constructor by using a static initialiser meant we could remove the ? from here and tidy up some code throughout the class. I generally like to make sure classes, once instantiated, are in a known-good state.

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -76,7 +76,7 @@ async function waitForEvent(
 }
 
 describe.each([
-  ["internal", undefined],
+  ["lucene", undefined],
```
samwho, line 79: Funny bug that must have existed for a while, but when this was set to "internal" it was skipping both SQS and Lucene tests. Whoops.

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -2196,28 +2196,6 @@ describe.each([
...
-        })
-
-        await config.api.viewV2.search(view.id, { query: {} })
-        expect(searchSpy).toHaveBeenCalledTimes(1)
-
-        expect(searchSpy).toHaveBeenCalledWith(
-          expect.objectContaining({
-            fields: ["id"],
-          })
-        )
```
samwho, line 2218: This was changed so that the ViewV2 is passed down and `fields` is undefined. `sdk.rows.search` has been coded to know that if `fields` is undefined, it derives all of the visible fields from the view. So this test isn't needed anymore.

packages/server/src/db/utils.ts
```diff
@@ -318,12 +318,8 @@ export function generateViewID(tableId: string) {
   }${SEPARATOR}${tableId}${SEPARATOR}${newid()}`
 }
 
-export function isViewID(viewId: string) {
-  return viewId?.split(SEPARATOR)[0] === VirtualDocumentType.VIEW
-}
```
samwho, line 323: Moved this to live with is `isFooId` functions in backend-core.

packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -99,10 +105,17 @@ export async function processAutoColumn(
...
+          tableMutated = true
         }
         break
     }
   }
+
+  if (tableMutated) {
+    const db = context.getAppDB()
+    const resp = await db.put(table)
+    table._rev = resp.rev
```
samwho, line 117: This is a substantial change that sort-of doesn't belong in this PR but made a few things in this PR much easier.

Up until now, when calling `inputProcessing`, there was a change that the `Table` you pass in gets modified. This happens if you're using an internal table, it has an auto ID column, and you're creating a row. Scattered throughout the codebase were a lot of places where we cloned the `Table` prior to calling `inputProcessing` and diff old and new afterwards, saving if they are different.

These checks became a pain when it was not `Table` anymore but `Table | ViewV2` so I decided enough of this, I'm just going to save the table at the point it changes. This takes the complexity away from higher up the stack at the code of possibly doing more saves than before.

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: This is nice - I like it

packages/server/src/api/controllers/row/utils/basic.ts
```diff
@@ -108,21 +126,30 @@ export function basicProcessing({
...
-  } else {
-    columns = columns.concat(PROTECTED_EXTERNAL_COLUMNS)
-    for (let internalColumn of [...PROTECTED_INTERNAL_COLUMNS, ...columns]) {
-      thisRow[internalColumn] = extractFieldValue({
-        row,
-        tableName: table._id!,
-        fieldName: internalColumn,
-        isLinked,
-      })
+  if (!isCalculationView) {
```
mike12345567, line 137: Would it be possible to avoid the double `if` statement? I sometimes find these hard to reason about, its easier to read `if (!isCalculationView && !sqs) .... else if (!isCalculationView) ...` as each block is quite clear in its intentions.

packages/server/src/api/controllers/table/utils.ts
```diff
@@ -139,8 +139,7 @@ export async function importToRows(
     const processed = await inputProcessing(user?._id, table, row, {
       noAutoRelationships: true,
     })
-    row = processed.row
-    table = processed.table
```
mike12345567, line 143: What happened to the table here - I assume this is the part where the table is getting saved inside `inputProcessing` rather than being returned?

packages/server/src/sdk/app/rows/internal.ts
```diff
@@ -20,21 +20,25 @@ export async function save(
...
+
   if (!inputs._rev && !inputs._id) {
     inputs._id = db.generateRowID(inputs.tableId)
   }
 
-  // this returns the table and row incase they have been updated
-  const dbTable = await sdk.tables.getTable(inputs.tableId)
-
-  // need to copy the table so it can be differenced on way out
-  const tableClone = cloneDeep(dbTable)
```
mike12345567, line 31: Unsure what this cloning was about - I get the feeling the `inputProcessing` can make changes to the schema etc in someway of the passed in table/source, has this changed and we don't need to worry about it anymore?

Realistically we shouldn't need to worry about this, but just raising it as this likely was added to fix a bug at some point.

packages/server/src/sdk/app/rows/internal.ts
```diff
@@ -49,24 +53,18 @@ export async function save(
...
-export async function find(tableOrViewId: string, rowId: string): Promise<Row> {
-  const { tableId, viewId } = tryExtractingTableAndViewId(tableOrViewId)
-
-  const table = await sdk.tables.getTable(tableId)
-  let row = await findRow(tableId, rowId)
-
-  row = await outputProcessing(table, row, { squash: true, fromViewId: viewId })
-  return row
+export async function find(sourceId: string, rowId: string): Promise<Row> {
+  const source = await getSource(sourceId)
```
mike12345567, line 59: Its nice how much cleaner the API level is because of this change!

packages/server/src/sdk/app/rows/queryUtils.ts
```diff
@@ -110,6 +110,9 @@ export const getQueryableFields = async (
     "_id", // Querying by _id is always allowed, even if it's never part of the schema
   ]
 
+  if (fields === undefined) {
```
mike12345567, line 113: since this is an exported function, should it check for both `null` and `undefined`? A `fields == null` might be safer incase a caller were to do something like: `const fields: string[] | null = null` and then pass this

packages/server/src/sdk/app/rows/search.ts
```diff
@@ -71,34 +72,43 @@ export async function search(
...
+    if (options.viewId) {
+      source = await sdk.views.get(options.viewId)
+      table = await sdk.views.getTable(source)
+      options = searchInputMapping(table, options)
+    } else if (options.tableId) {
+      source = await sdk.tables.getTable(options.tableId)
+      table = source
+      options = searchInputMapping(table, options)
+    } else {
+      throw new Error(`Invalid source ID: ${options.viewId || options.tableId}`)
```
mike12345567, line 86: Will this error message ever be correct? if both `options.viewId` and `options.tableId` are undefined then it'll just print `Invalid source ID: `

packages/server/src/sdk/app/rows/search/internal/sqs.ts
```diff
@@ -70,13 +97,18 @@ function buildInternalFieldList(
     })
     return junctionFields
   }
-  fieldList = fieldList.concat(
-    PROTECTED_INTERNAL_COLUMNS.map(col => `${table._id}.${col}`)
-  )
-  for (let key of Object.keys(table.schema)) {
+  if (sdk.tables.isTable(source)) {
+    for (const key of PROTECTED_INTERNAL_COLUMNS) {
+      if (allowedFields && !allowedFields.includes(key)) {
```
mike12345567, line 102: Could this be done with a filter instead - `PROTECTED_INTERNAL_COLUMNS.filter(key => !allowedFields?.includes(key)).map(key => ${table._id}.${key})`

packages/server/src/api/controllers/table/utils.ts
samwho: Correct, we no longer need to "carry" the table through the loop.

packages/server/src/sdk/app/tables/utils.ts
```diff
@@ -9,3 +9,7 @@ export function isExternal(opts: { table?: Table; tableId?: string }): boolean {
   }
   return false
 }
+
+export function isTable(table: any): table is Table {
+  return table.type === "table"
```
mike12345567, line 14: Its probably also worth checking the ID to see if it is in the form of a table ID - `isTableId` (I now realise these functions break the rule of capitalised `ID`

packages/server/src/middleware/triggerRowActionAuthorised.ts
```diff
@@ -15,7 +15,7 @@ export function triggerRowActionAuthorised(
       const rowActionId: string = ctx.params[actionPath]
 
       const isTableId = docIds.isTableId(sourceId)
-      const isViewId = utils.isViewID(sourceId)
+      const isViewId = docIds.isViewId(sourceId)
```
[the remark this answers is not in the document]
samwho: Hmm, this doesn't seem to be the case:

https://github.com/Budibase/budibase/blob/d550aad63ae4aafe80e0de9b42569eadfb24825d/packages/backend-core/src/docIds/params.ts#L67-L87

Happy to make them consistent in a follow-up PR though.

packages/server/src/sdk/app/rows/internal.ts
samwho: You're correct, yeah. I address this in another comment, but `inputProcessing` can modify the table in very rare cases and we leak that out through to the caller and leave them to handle it. This PR changes that so that `inputProcessing` saves the table itself now.

packages/server/src/sdk/app/views/index.ts
```diff
@@ -40,16 +40,87 @@ export async function getEnriched(viewId: string): Promise<ViewV2Enriched> {
...
+      )
+    }
+  }
+
+  const groupByFields = helpers.views.basicFields(view)
+  for (const groupByFieldName of Object.keys(groupByFields)) {
+    const targetSchema = table.schema[groupByFieldName]
+    if (!targetSchema) {
+      throw new HTTPError(
+        `Group by field "${groupByFieldName}" does not exist in the table schema`,
```
mike12345567, line 89: should this error message state that it doesn't exist or is an invalid type, since the `basicFields` are the only ones returned, it could be a relationship column etc

packages/server/src/sdk/app/rows/search/internal/sqs.ts
samwho: It certainly could, but this felt a little cleaner. Happy to change it if you want but don't think the benefit is that large.

packages/server/src/utilities/rowProcessor/index.ts
mike12345567: Think this is fine - the `auto ID` is used a lot more rarely now that it used to be - this also explains why the worry about cloning has disappeared!

packages/server/src/sdk/app/tables/utils.ts
samwho: I feel like if we detect `type == "table"` and _not_ a table ID, that's an error no? We should probably throw?

packages/server/src/sdk/app/views/index.ts
samwho: I'm not sure I understand what you mean. The current wording is strictly correct when compared against what the code is doing. I'm uncertain how relationship fields come into play.

packages/server/src/middleware/triggerRowActionAuthorised.ts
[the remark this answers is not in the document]
mike12345567: Yeap realised later in the chain when investigating the `isFoo` functions don't keep this consistent, we can refactor that in follow-up!

packages/server/src/sdk/app/rows/search/internal/sqs.ts
mike12345567: I generally find the nature of `.filter` and `.map` a bit more readable than a for loop since its clear what its doing, making a new sub-array rather than performing any explicit actions, but very much preference!

packages/server/src/sdk/app/tables/utils.ts
mike12345567: Likely yes - there is cases where that could happen - we probably could add that here, the core to my point though is that you're better to look at the ID of the table than the `type` field, generally checking the ID is the preferred way to confirm what type a document/object is.

packages/server/src/sdk/app/views/index.ts
mike12345567: Ah I think I mis-understood this functions purpose - is it for validating the view when its being saved? I was basing this on what I thought the `basicFields` function does, but I now realise this simply retrieves all the fields which aren't calculation type from the schema - I understand now - sorry!

packages/server/src/sdk/app/tables/utils.ts
samwho: Ahh I see. Done! Also changed isView to use isViewId.

packages/server/src/sdk/app/rows/search.ts
samwho: Good catch, have made this error message better.

packages/server/src/sdk/app/views/index.ts
mike12345567: Actually - raises a separate point, the group-by and calculation columns can come from a related table, so we need to validate the columns exist in the main table, or a table they are related to (making sure they correctly reference that other table).

samwho: Yeah this is for when views get saved. `basicFields` maybe not the best name. 😅

mike12345567 (approved):
LGTM!

PR #14630, opened 2024-09-23
MariaDB - avoid using correlated sub-queries

mike12345567:
## Description
This is a problem that was raised by our QA MariaDB database which was hitting a limit around how much data can be returned from an aggregate response - the JSON that was being returned was malformed and not performing as expected.

The issue was that we did not apply a limit to MySQL/MariaDB queries as it wasn't possible to limit these without the use of a correlated sub-query, something which is not allowed - information about this [here.](https://mariadb.com/kb/en/subquery-limitations/#:~:text=Subqueries%20in%20the%20FROM%20clause,when%20the%20query%20is%20executed)

To get around this we have used a limit which can be applied within the aggregation itself, avoiding the malformed response. This is not as performant as using a correlated sub-query, but we were very limited in terms of options and this gave us the best set of fixed problems while performing within acceptable limits.

### DIFF
--- a/packages/backend-core/src/sql/sqlTable.ts
+++ b/packages/backend-core/src/sql/sqlTable.ts
@@ -210,14 +210,25 @@ function buildDeleteTable(knex: SchemaBuilder, table: Table): SchemaBuilder {
 
 class SqlTableQueryBuilder {
   private readonly sqlClient: SqlClient
+  private extendedSqlClient: SqlClient | undefined
 
   // pass through client to get flavour of SQL
   constructor(client: SqlClient) {
     this.sqlClient = client
   }
 
-  getSqlClient(): SqlClient {
+  getBaseSqlClient(): SqlClient {
     return this.sqlClient
+  }
+
+  getSqlClient(): SqlClient {
+    return this.extendedSqlClient || this.sqlClient
+  }
+
+  // if working in a database like MySQL with many variants (MariaDB)
+  // we can set another client which overrides the base one
+  setExtendedSqlClient(client: SqlClient) {
+    this.extendedSqlClient = client
   }
 
   /**
--- a/packages/server/src/api/routes/tests/search.spec.ts
+++ b/packages/server/src/api/routes/tests/search.spec.ts
@@ -45,14 +45,14 @@ import { generateRowIdField } from "../../../integrations/utils"
 import { cloneDeep } from "lodash/fp"
 
 describe.each([
-  // ["in-memory", undefined],
-  // ["lucene", undefined],
-  // ["sqs", undefined],
-  // [DatabaseName.POSTGRES, getDatasource(DatabaseName.POSTGRES)],
-  // [DatabaseName.MYSQL, getDatasource(DatabaseName.MYSQL)],
-  // [DatabaseName.SQL_SERVER, getDatasource(DatabaseName.SQL_SERVER)],
+  ["in-memory", undefined],
+  ["lucene", undefined],
+  ["sqs", undefined],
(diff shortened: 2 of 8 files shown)


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2262,6 +2279,31 @@ describe.each([
...
+        })
+        for (let i = 0; i < 11; i++) {
+          await config.api.row.save(relatedTable._id!, {
+            name: i,
+            product: [mainRow._id!],
+          })
+        }
+      })
+
+      it("can only pull 500 related rows", async () => {
```
samwho, line 2300: nit: test name doesn't make sense

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2262,6 +2279,31 @@ describe.each([
...
+        await withCoreEnv({ SQL_MAX_RELATED_ROWS: "10" }, async () => {
+          const response = await expectQuery({}).toContain([{ name: "foo" }])
+          expect(response.rows[0].productCat).toBeArrayOfSize(10)
```
samwho, line 2303: I'd also do a version of this query without the `SQL_MAX_RELATED_ROWS` env var to make sure it does return 11 as expected.

packages/server/src/sdk/app/tables/external/index.ts
```diff
@@ -231,7 +232,10 @@ export async function save(
   // remove the rename prop
   delete tableToSave._rename
 
-  datasource.entities[tableToSave.name] = tableToSave
+  datasource.entities = {
+    ...datasource.entities,
+    ...tables,
+  }
```
samwho, line 238: How come this is required?

packages/server/src/sdk/app/tables/external/utils.ts
```diff
@@ -22,12 +22,16 @@ export function cleanupRelationships(
   tables: Record<string, Table>,
   oldTable?: Table
 ) {
+  if (!oldTable) {
+    return
+  }
```
samwho, line 27: Would prefer to see `oldTable` become a non-optional parameter and push the handling of that scenario up to the callers. As it is now it's easy to call the function assuming `oldTable` is fine to omit, only to realise the function then does nothing.

packages/server/src/sdk/app/tables/external/index.ts
mike12345567: throughout this function it builds up a list of updated tables in the `tables` map, based on other tables getting updated when relationships are added. These weren't always getting stored correctly.

samwho (approved):

packages/backend-core/src/sql/sqlTable.ts
```diff
@@ -210,16 +210,27 @@ function buildDeleteTable(knex: SchemaBuilder, table: Table): SchemaBuilder {
 
 class SqlTableQueryBuilder {
   private readonly sqlClient: SqlClient
+  private extendedSqlClient: SqlClient | undefined
```
samwho, line 213: Ooh how come this is necessary? Why can't we just use `sqlClient`?

mike12345567: We need to know that the implementation/dialect is `MY_SQL` - but also its MariaDB - for Knex it needs to know its MySQL, technically we could get around this by setting everywhere that needs to know its MySQL to MariaDB when needed, but I like the idea of separating the difference in client here, since its still a MySQL datasource, it just reads a different version once it connects.

PR #14633, opened 2024-09-24
Update the version of PostHog used in the cli package.

samwho:
## Description

This needs to always line up with what `backend-core` uses because `cli` depends on `backend-core`. If you don't do this, you end up getting errors like this:

```
➜  ~ budi hosting --init
/Users/jvcalderon/.nvm/versions/node/v20.10.0/lib/node_modules/@budibase/cli/dist/index.js:892534
    this.posthog = new import_posthog_node3.PostHog(token);
                   ^

TypeError: import_posthog_node3.PostHog is not a constructor
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -28,7 +28,7 @@
     "inquirer": "8.0.0",
     "lookpath": "1.1.0",
     "node-fetch": "2.6.7",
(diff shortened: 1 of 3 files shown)


jvcalderon (approved):
LGTM

PR #14634, opened 2024-09-24
More automation tests

PClmnt:
## Description
Added tests around:

- Looping query rows and updating / deleting / verifying correctness after
- Same thing but non looping
- Filter test cases


packages/server/src/automations/tests/scenarios/looping.spec.ts
```diff
@@ -269,4 +270,149 @@ describe("Loop automations", () => {
...
+    })
+
+    const rows = [
+      { name: "Row 1", value: 1, tableId: table._id },
+      { name: "Row 2", value: 2, tableId: table._id },
+      { name: "Row 3", value: 3, tableId: table._id },
+    ]
+
+    for (const row of rows) {
+      await config.createRow(row)
```
samwho, line 303: nit: prefer `config.api.row.bulkImport(table._id!, { rows })`

packages/server/src/automations/tests/scenarios/looping.spec.ts
```diff
@@ -269,4 +270,149 @@ describe("Loop automations", () => {
 
     expect(results.steps[1].outputs.message).toContain("- 3")
   })
+
+  it("should run an automation with a loop and update row step", async () => {
+    const table = await config.createTable({
```
samwho, line 275: nit: prefer `config.api.table.save`

packages/server/src/automations/tests/scenarios/scenarios.spec.ts
```diff
@@ -233,4 +319,168 @@ describe("Automation Scenarios", () => {
...
+        value: 10,
+        rowValue: 15,
+        expectPass: false,
+      },
+    ]
+
+    testCases.forEach(({ condition, value, rowValue, expectPass }) => {
+      it(`should ${
+        expectPass ? "pass" : "fail"
+      } the filter when condition is "${condition}" and value is ${value}`, async () => {
```
samwho, line 450: nit: prefer `it.each(testCases)("...")`, if the reason you didn't do this is the ternary inside of the string I wouldn't worry too much about grammatically correct test names.

samwho (commented):
Just a few small comments. These are really good tests.

packages/server/src/automations/tests/scenarios/scenarios.spec.ts
```diff
@@ -444,43 +444,48 @@ describe("Automation Scenarios", () => {
...
-          .filter({
-            field: "{{ steps.2.rows.0.value }}",
-            condition,
-            value,
+    it.each(testCases)(
+      "should pass the filter when condition is %s",
+      async ({ condition, value, rowValue, expectPass }) => {
+        it(`should ${
+          expectPass ? "pass" : "fail"
+        } the filter when condition is "${condition}" and value is ${value}`, async () => {
```
samwho, line 452: You shouldn't need this anymore, I suspect this doesn't actually work as-is.

PClmnt: I was... not concentrating at all when I made the last commit. Just left the existing code under it lol, should be sorted now!

mike12345567 (approved):
Looks good to me, nice use of `it.each` to get quite a nice set of tests here, lovely!

PR #14635, opened 2024-09-24
Default values for options and array types (frontend)

aptkingston:
## Description
Few changes to default values, and some fixes around schema mutations when editing columns and primary display.

- Adds default value setting for options type (renders a picker instead of a bindable text field, since it needs to be a valid option as the default value)
- Adds default value setting for multi-select type (renders a multi-select picker instead of a bindable text field, since it needs to be a valid option as the default value)
- Ensures that schema properties are consistent upon saving (e.g. required deletes any default value and vice versa)
- Fixes an issue where you could set default value on primary display columns (as the logic for determining whether a field is required or not used an incorrect order of precedence)
- Show default value setting as long as the feature is enabled, but disable it if it cannot be used
- Ensures primary display columns are required and do not have default values when saving
- When changing primary display via the grid, it now also mutates the schema to make the field required and have no default value. Before this, you were able to make any field the primary display column - meaning you could have non-required primary display columns which is not meant to be possible.

Single select:
image

Multi select:
image

Addresses:
- https://linear.app/budibase/issue/BUDI-8678/add-the-ability-to-apply-default-values-to-single-and-multi-select


packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -134,8 +134,10 @@ async function processDefaultValues(table: Table, row: Row) {
 
   for (const [key, schema] of Object.entries(table.schema)) {
     if ("default" in schema && schema.default != null && row[key] == null) {
-      const processed = await processString(schema.default, ctx)
-
+      const processed =
+        typeof schema.default === "string"
+          ? await processString(schema.default, ctx)
+          : schema.default
```
aptkingston, line 140: Default value is not always a string any more, as multi-select types have a default value that is an array. Therefore we only run binding enrichment now if it's a string.

packages/shared-core/src/table.ts
```diff
@@ -53,8 +53,9 @@ const allowDefaultColumnByType: Record<FieldType, boolean> = {
   [FieldType.DATETIME]: true,
   [FieldType.LONGFORM]: true,
   [FieldType.STRING]: true,
+  [FieldType.OPTIONS]: true,
+  [FieldType.ARRAY]: true,
```
samwho, line 57: Could you add some tests for default values in fields of this type? 🙏

packages/server/src/utilities/rowProcessor/index.ts
samwho: I'm not sure where we currently land on backend changes going to master vs going to this branch. It's messy to do any kind of split, we should commit to one or the other.

samwho (changes requested):

packages/server/src/utilities/rowProcessor/index.ts
aptkingston: Ah yes, good point. It's a nightmare developing this way, but I'll split this apart and do a backend PR to master first, then leave this as frontend only.

aptkingston: Done, should be a clean split now.

samwho (approved):

PR #14637, opened 2024-09-24
Enrich squashed links values

adrinr:
## Description
Add tests around link fields squashing and enriching fields values. Without this, some fields such as user columns don't use the same format between row values and linked row values.

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -2291,8 +2291,17 @@ describe.each([
             },
           })
 
+          const flags = [`*:${FeatureFlag.ENRICHED_RELATIONSHIPS}`]
+          if (isLucene) {
+            flags.push("*:!SQS")
+          } else if (isSqs) {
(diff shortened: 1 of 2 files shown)


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -2213,6 +2215,163 @@ describe.each([
...
+                  age: { visible: true, readonly: true },
+                },
+              },
+            },
+          })
+
+          const response = await withCoreEnv(
+            { TENANT_FEATURE_FLAGS: `*:${FeatureFlag.ENRICHED_RELATIONSHIPS}` },
+            () => config.api.viewV2.search(view.id)
+          )
```
samwho, line 2297: I've noticed and fixed this in my current branch, but you need to be careful to also set the SQS flag correctly here, otherwise it gets overwritten and the lucene tests use sqs by accident.

Makes me think a feature flag helper for tests is in order.

samwho (changes requested):

samwho (approved):

PR #14639, opened 2024-09-25
Display related columns in table

adrinr:
## Description
Display related fields on the table grid. These new cells are readonly.

There is different logic depending on the relationship type:
1. If we are for single-value relationships, we just display the field as it is.
2. For multi-value relationship (many-to-many or the many side ot one-to-many), we aggregate the values using the following logic:
   1. String and number types is displayed as CSV
   2. Date times are also displayed CSV, honoring the format of the column (date only/timezones/etc.)
   3. Booleans are stringified and CSVed (we don't have a multi-boolean cell, stringifying it for now should be enough)
   4. User and multi-user get displayed as multi-user, removing duplicates
   5. Option get displayed as multi-options
   6. Multi-option aggregates the values removing duplicates
   7. Attachment and multi-attachment get displayed as multi-attachment
   8. Signature get displayed as readonly attachments (that already support the desired display)
   9. Json values get display as an array of the json objects

By default, these new columns will be named as `relationshipColumn.relationshipField`. This can be renamed in the design section if required by the final user.

### One-sided column with all possible types
image
image

### Multi-sided column with all possible types
image
image

### In data section, the columns might be hidden, resized and reordered

https://github.com/user-attachments/assets/1ea243eb-fc66-46df-8825-65a7bde41ea6

### In design, the columns might be hidden, renamed, resized and reordered

https://github.com/user-attachments/assets/2efff250-12af-4f41-8274-6ed192f02c96

### In design, the relation column might be hidden leaving the other columns visible. On data section, this means that will the subvalues are to be hidden

https://github.com/user-attachments/assets/98fa2682-be3a-40c3-bde3-39cf767b3207

### Signature (and the multi-signature) are visible for all themes

https://github.com/user-attachments/assets/2775ed4b-7b65-41f0-9c81-61ae7d689a94

### Sort and make primary display is not available on related columns
image

### Hide column button works as it does from the picker

https://github.com/user-attachments/assets/597de104-d511-404c-93b4-7fb87149aa0b

### Sort picker does not contain the related columns (also fixed a race-condition that caused flashes)

https://github.com/user-attachments/assets/2f7cef84-a820-4d7f-bf42-cd982e72698c

## Launchcontrol
Display related fields on the table grid.

## Feature branch env
[Feature Branch Link](http://fb-budi-8565-related-columns.fb.qa.budibase.net)


packages/backend-core/src/features/index.ts
```diff
@@ -277,5 +277,7 @@ export const flags = new FlagSet({
   AUTOMATION_BRANCHING: Flag.boolean(env.isDev()),
   SQS: Flag.boolean(env.isDev()),
   [FeatureFlag.AI_CUSTOM_CONFIGS]: Flag.boolean(env.isDev()),
-  [FeatureFlag.ENRICHED_RELATIONSHIPS]: Flag.boolean(false),
+  [FeatureFlag.ENRICHED_RELATIONSHIPS]: Flag.boolean(
+    env.isDev() && !env.isTest()
```
adrinr, line 281: The tests are need to be rewritten to support the `ENRICHED_RELATIONSHIPS` by default

mike12345567: I think we should probably update the test cases to do this rather than putting the `!env.isTest()` in here, that could be quite dangerous/missed or forgotten.

mike12345567 (approved):
LGTM - thanks for all the videos/screenshots, really demonstrates how this works well! I'll let @aptkingston comment on the the store updates as I know that was an area of concern.

packages/frontend-core/src/utils/relatedColumns.js
```diff
@@ -0,0 +1,111 @@
...
+  let result = ""
+  try {
+    if (fromSingle) {
+      result = row[field.related.field]?.[0]?.[field.related.subField]
+    } else {
+      const parser = columnTypeManyParser[field.type] || (value => value)
+
+      result = parser(
+        row[field.related.field]
+          .flatMap(r => r[field.related.subField])
```
aptkingston, line 86: This will throw if there isn't a relationship for a certain row - when I span this up locally on a view with an optional relationship column that had mostly empty values and enabled a related column, this was throwing a lot.

packages/frontend-core/src/utils/relatedColumns.js
```diff
@@ -0,0 +1,111 @@
...
+          FieldType.DATETIME,
+          FieldType.LONGFORM,
+          FieldType.BARCODEQR,
+        ].includes(field.type)
+      ) {
+        result = result.join(", ")
+      }
+    }
+  } catch (e) {
+    result = "Not rendable"
```
aptkingston, line 106: I don't think we should return anything here - I think an empty value would be better.

packages/frontend-core/src/components/grid/stores/columns.js
```diff
@@ -73,16 +79,24 @@ export const deriveStores = context => {
...
+      const { related } = column
+      const mutation = { width }
+      if (!related) {
+        datasource.actions.addSchemaMutation(column.name, mutation)
+      } else {
+        datasource.actions.addSubSchemaMutation(
+          related.subField,
+          related.field,
+          mutation
+        )
```
aptkingston, line 97: I've seen this in a few places, where we now need to check the type of column then call a different function to make the schema mutation. This got me thinking that it might be nice to abstract those column differences into the datasource store, and just have a single `addSchemaMutation` which internally checks the type of column and then updates either `schemaMutations` or `subSchemaMutations` as required.

It's not a change we need to make for now - just something I might look to refactor in future!

aptkingston (changes requested):
Great job integrating this @adrinr. Was very pleased to see this works natively with cell conditions too.

Just found a few issues while testing it which I've noted below.

You're currently able to set these related columns as the primary display column, which errors of course if you actually click it. We should disable this for related columns:
image

You can currently try to sort by these columns as well, which returns in the server throwing a 500 with `{"message":"Cannot read properties of undefined (reading 'type')","status":500}`. The fact it throws a 500 is another bug separate to this I assume, but we should disable the ability to sort by these columns.

You can also attempt to hide these columns via the header context menu, which again doesn't actually work. So we should either disable this, or even better if we're able to make this actually update the related column to be hidden again.

(Stylistic comment, feel free to ignore:)
Completely NAB, but I personally find the combination of array.reduce with single letter params and no comments quite difficult to read and understand. In future I would appreciate a few comments or more verbose params or alternatives to array.reduce in some instances to help make the code more readable and maintainable 👍

packages/frontend-core/src/utils/relatedColumns.js
```diff
@@ -0,0 +1,111 @@
+import { FieldType, RelationshipType } from "@budibase/types"
+import { Helpers } from "@budibase/bbui"
+
+const columnTypeManyTypeOverrides = {
```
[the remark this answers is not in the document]
adrinr: We don't have unit test in frontend-core. It might be a good opportunity to add them, but I will add them separately and it needs few changes just for the vitest setup

packages/frontend-core/src/components/grid/stores/columns.js
adrinr: I was thinking something similar. I can try to refactor it to make this logic abstract to the consumer! I will do it in a different PR to keep this one from growing much more

aptkingston (approved):
LGTM!

aptkingston:
Just before you merge this - I have a table with 10K rows and 2 identical views, one with related columns shown and one without, but otherwise identical. The one without related columns can do a search call in about 200ms, but the one with them takes 15-20 seconds.

This is reliable reproducible by creating a new view (fast) then showing even a single related column (extremely slow).

adrinr:
> Just before you merge this - I have a table with 10K rows and 2 identical views, one with related columns shown and one without, but otherwise identical. The one without related columns can do a search call in about 200ms, but the one with them takes 15-20 seconds.
> 
> This is reliable reproducible by creating a new view (fast) then showing even a single related column (extremely slow).
> 
> Edit: in a table with 36 rows, the search call goes from 100ms (no rel columns) to 600ms (with rel columns). Definitely seems to be a huge performance impact.
> 
> Hiding the rel column again does not fix the issue - it gets stuck in the "slow" state.
> 
> I realise this PR is just the frontend changes though, which don't affect this issue at all.
> 
> Could someone else investigate and check?

Merging this PR and investigating the backend issue separately

PR #14649, opened 2024-09-27
Fix small issue with conditions in branches

PClmnt:
## Description
For some reason only known to previous me I decided to write the branching tests without handlebars syntax in the conditions. This updates the tests and fixes the evaluation code.

### DIFF
--- a/packages/server/src/threads/automation.ts
+++ b/packages/server/src/threads/automation.ts
@@ -516,10 +516,8 @@ class Orchestrator {
       filter => {
         Object.entries(filter).forEach(([_, value]) => {
           Object.entries(value).forEach(([field, _]) => {
-            const fromContext = processStringSync(
-              `{{ literal ${field} }}`,
-              this.context
-            )
+            const updatedField = field.replace("{{", "{{ literal ")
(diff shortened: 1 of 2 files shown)


packages/server/src/threads/automation.ts
```diff
@@ -516,10 +516,8 @@ class Orchestrator {
       filter => {
         Object.entries(filter).forEach(([_, value]) => {
           Object.entries(value).forEach(([field, _]) => {
-            const fromContext = processStringSync(
-              `{{ literal ${field} }}`,
-              this.context
-            )
+            const updatedField = field.replace("{{", "{{ literal ")
```
samwho, line 519:
```suggestion
            const updatedField = field.replace("{{", "{{ literal")
```

samwho (approved):

packages/server/src/threads/automation.ts
PClmnt: I changed this, but changed it back as I realised we need the space. There needs to be a space after literal for the actual binding i.e `{{ literal steps.1.success }}` instead of `{{ literalsteps.1.success }}`

PR #14650, opened 2024-09-27
Enriched relationship view performance improvement

mike12345567:
## Description
This improves the performance of views when the enriched relationships feature has been enabled. There was an issue that caused the `squashLinks` and `outputProcessing` to loop onto each other until things broke down - this fixes the issue.

I've introduced a new function `coreOutputProcessing` which just processes rows as needed for output, leaving squashing and such to the main `outputProcessing`.


adrinr (approved):
Nice catch!

PR #14651, opened 2024-09-27
Fix OSS build issues

adrinr:
## Description
Updating budibase/pro to latest on the lock file to ensure the ai references are pulled. Otherwise it fails due some missing references


adrinr:
This is a nice catch from this pipeline :)

mike12345567:
Very nice catch - great find @adrinr ! Glad we have the pipeline for this sort of thing!

mike12345567 (approved):

PR #14660, opened 2024-09-30
View security allow filters specified on view to use any

adrinr:
## Description
Getting sure that view filters are always respected, even if the fields used are not visible to the table or to the user. These filters are set by the builder, so they need to be honored. Otherwise, there are risks of displaying some unexpected data.

Also, fixing some data

### View respects the filters even if the columns are hidden
https://github.com/user-attachments/assets/1ffb4fc8-9499-4ebf-aec4-843eda8443cb

### User bindings can be used in views
https://github.com/user-attachments/assets/2054e487-8ef7-49f1-b667-8716012a2cac

### View filters cannot be bypassed in the design section

https://github.com/user-attachments/assets/554f6ba9-de57-4118-9378-e879857faf08

### View filters cannot be bypassed in the client
https://github.com/user-attachments/assets/005940a1-e316-49a3-bcc6-6a1050ae3e56

## Launchcontrol
Improving view security

### DIFF
--- a/packages/server/src/api/controllers/row/views.ts
+++ b/packages/server/src/api/controllers/row/views.ts
@@ -3,14 +3,9 @@ import {
   ViewV2,
   SearchRowResponse,
   SearchViewRowRequest,
-  SearchFilterKey,
-  LogicalOperator,
 } from "@budibase/types"
-import { dataFilters } from "@budibase/shared-core"
 import sdk from "../../../sdk"
-import { db, context, features } from "@budibase/backend-core"
-import { enrichSearchContext } from "./utils"
-import { isExternalTableID } from "../../../integrations/utils"
+import { context } from "@budibase/backend-core"
 
 export async function searchView(
   ctx: UserCtx<SearchViewRowRequest, SearchRowResponse>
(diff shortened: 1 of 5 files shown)


packages/server/src/api/controllers/row/views.ts
```diff
@@ -27,58 +22,23 @@ export async function searchView(
 
   const { body } = ctx.request
 
-  // Enrich saved query with ephemeral query params.
-  // We prevent searching on any fields that are saved as part of the query, as
-  // that could let users find rows they should not be allowed to access.
-  let query = dataFilters.buildQuery(view.query || [])
```
mike12345567, line 33: I believe this is needed to handle if the `view.query` is of type `SearchFilter[]` not `SearchFilters`, which will be the case in the DB, so we need to keep a call to `buildQuery.`

adrinr: Yes, I moved it in the sdk, otherwise the SDK and the controller would need to be highly coupled

mike12345567 (approved):
LGTM - nice job!

PR #14663, opened 2024-09-30
Add the ability to look up context by ID in automation thread

PClmnt:
## Description
With the change to the frontend to use IDs as the runtime binding the backend needed to be updated to allow for ID support also.


packages/server/src/definitions/automations.ts
```diff
@@ -15,6 +15,7 @@ export interface TriggerOutput {
 
 export interface AutomationContext extends AutomationResults {
   steps: any[]
+  stepsById?: Record<string, any>
```
samwho, line 18: How come this is optional? Would it not be safer to default it to `{}`? All of the code I can see referencing this does so with an `!` so it feels a bit pointless to have it listed here as optional.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -106,80 +105,98 @@ class BaseStepBuilder {
...
       inputs,
-      opts?.stepName
+      opts
     )
   }
 
-  serverLog(input: ServerLogStepInputs, opts?: { stepName?: string }): this {
+  serverLog(
+    input: ServerLogStepInputs,
+    opts?: { stepName?: string; id?: string }
```
samwho, line 193: nit: I'd either call `stepName` -> `name`, or `id` -> `stepId` to be consistent.

samwho (approved):

PR #14673, opened 2024-10-01
Backport V3 backend view change

mike12345567:
## Description
backport of V3 backend changes for search filters on view, giving this the correct type to support logical operators (and the new grouped filter UI).

This is work completed by @aptkingston and @deanhannigan.


packages/server/src/api/controllers/row/views.ts
```diff
@@ -17,73 +20,90 @@ export async function searchView(
 ) {
   const { viewId } = ctx.params
 
-  const view = await sdk.views.get(viewId)
+  const view: ViewV2 = await sdk.views.get(viewId)
```
samwho, line 23: How come this annotation is needed? `sdk.views.get` returns a `Promise<ViewV2>`

packages/server/src/api/controllers/row/views.ts
```diff
@@ -17,73 +20,90 @@ export async function searchView(
...
-    if (
-      !isExternalTableID(view.tableId) &&
-      !(await features.flags.isEnabled("SQS"))
-    ) {
+    if (!supportsLogicalOperators) {
+      // In the unlikely event that a Grouped Filter is in a non-SQS environment
+      // It needs to be ignored entirely
+      let queryFilters: LegacyFilter[] = Array.isArray(view.query)
+        ? view.query
+        : []
```
samwho, line 55: I feel an error would be more appropriate here than continuing silently, unless I'm missing something? Discarding `view.query` feels wrong.

packages/server/src/api/controllers/row/views.ts
```diff
@@ -17,73 +20,90 @@ export async function searchView(
...
+  const sqsEnabled = await features.flags.isEnabled("SQS")
+  const supportsLogicalOperators = isExternalTableID(view.tableId) || sqsEnabled
+
   // Enrich saved query with ephemeral query params.
   // We prevent searching on any fields that are saved as part of the query, as
   // that could let users find rows they should not be allowed to access.
-  let query = dataFilters.buildQuery(view.query || [])
+  let query = dataFilters.buildQueryLegacy(view.query)
+
+  delete query?.onEmptyFilter
```
samwho, line 44: How come we're deleting this? Isn't this something the user has requested?

packages/shared-core/src/filters.ts
```diff
@@ -304,10 +307,143 @@ export class ColumnSplitter {
...
+      query[queryOperator]![field] = value
+    }
+  }
+
+  return query
+}
+
+export const buildQueryLegacy = (
+  filter?: LegacyFilter[] | SearchFilters
+): SearchFilters | undefined => {
```
samwho, line 441: Can this actually return `undefined`? If so, can it be written in such a way it doesn't?

It's opinion but I think it is _generally_ incorrect to allow `f(undefined) === undefined`. Functions should take real arguments and checking for undefined should generally be left to the caller.

packages/shared-core/src/filters.ts
```diff
@@ -368,13 +504,15 @@ export const buildQuery = (filter: SearchFilter[]) => {
...
+      ["contains", "notContains", "containsAny"].includes(
+        operator.toLocaleString()
+      ) &&
       type === "array" &&
       typeof value === "string"
     ) {
       value = value.split(",")
     }
-    if (operator.startsWith("range") && query.range) {
+    if (operator.toLocaleString().startsWith("range") && query.range) {
```
samwho, line 515: How come `toLocaleString` is needed here?

packages/shared-core/src/filters.ts
```diff
@@ -401,7 +539,7 @@ export const buildQuery = (filter: SearchFilter[]) => {
         }
       }
     } else if (isLogicalSearchOperator(queryOperator)) {
-      // TODO
+      // ignore
```
samwho, line 542: I know this isn't really part of the PR, but what's the plan here long term?

packages/shared-core/src/filters.ts
```diff
@@ -423,14 +561,68 @@ export const buildQuery = (filter: SearchFilter[]) => {
...
+  const globalOnEmpty = parsedFilter.onEmptyFilter
+    ? parsedFilter.onEmptyFilter
+    : null
+
+  const globalOperator: LogicalOperator =
+    operatorMap[parsedFilter.logicalOperator as FilterGroupLogicalOperator]
+
+  const coreRequest: SearchFilters = {
+    ...(globalOnEmpty ? { onEmptyFilter: globalOnEmpty } : {}),
+    [globalOperator]: {
```
samwho, line 604: This can be simplified a fair bit:

```suggestion
  const globalOperator = operatorMap[parsedFilter.logicalOperator]
  const coreRequest: SearchFilters = {
    ...(parsedFilter.onEmptyFilter || {}),
    [globalOperator]: {
```

packages/shared-core/src/filters.ts
```diff
@@ -423,14 +561,68 @@ export const buildQuery = (filter: SearchFilter[]) => {
...
+  return coreRequest
+}
+
 // The frontend can send single values for array fields sometimes, so to handle
 // this we convert them to arrays at the controller level so that nothing below
 // this has to worry about the non-array values.
 export function fixupFilterArrays(filters: SearchFilters) {
+  if (!filters) {
+    return filters
+  }
```
samwho, line 625: The types don't suggest that `undefined` is possible, do they need to be updated or can this be removed?

packages/shared-core/src/utils.ts
```diff
@@ -1,5 +1,20 @@
...
+  SearchFilters,
+  BasicOperator,
+  ArrayOperator,
+} from "@budibase/types"
 import * as Constants from "./constants"
+import { removeKeyNumbering } from "./filters"
+
+// an array of keys from filter type to properties that are in the type
+// this can then be converted using .fromEntries to an object
+type WhitelistedFilters = [
```
samwho, line 14: Would prefer to see `AllowedFilters` here.

packages/shared-core/src/utils.ts
```diff
@@ -104,3 +119,96 @@ export function isSupportedUserSearch(query: SearchFilters) {
...
+ * If filters is not an array, the migration is skipped
+ *
+ * @param {LegacyFilter[] | SearchFilterGroup} filters
+ */
+export const processSearchFilters = (
+  filters: LegacyFilter[] | SearchFilterGroup | undefined
+): SearchFilterGroup | undefined => {
+  if (!filters) {
+    return
+  }
```
samwho, line 136: Ideally would prefer to see this check pushed out to the caller and the type signature of this function to have the `undefined`s removed.

packages/shared-core/src/utils.ts
```diff
@@ -104,3 +119,96 @@ export function isSupportedUserSearch(query: SearchFilters) {
...
+
+  // Base search config.
+  const defaultCfg: SearchFilterGroup = {
+    logicalOperator: FilterGroupLogicalOperator.ALL,
+    groups: [],
+  }
+
+  const filterWhitelistKeys = [
```
samwho, line 144: Would prefer `filterAllowedKeys`.

packages/shared-core/src/utils.ts
```diff
@@ -104,3 +119,96 @@ export function isSupportedUserSearch(query: SearchFilters) {
...
+          acc.onEmptyFilter = value
+        } else if (key === "operator" && value === "allOr") {
+          // Group 1 logical operator
+          baseGroup.logicalOperator = FilterGroupLogicalOperator.ANY
+        }
+
+        return acc
+      }
+
+      const whiteListedFilterSettings: WhitelistedFilters =
```
samwho, line 184: `allowedFilterSettings`

packages/types/src/documents/app/view.ts
```diff
@@ -65,7 +65,9 @@ export interface ViewV2 {
   name: string
   primaryDisplay?: string
   tableId: string
-  query?: SearchFilter[]
+  query?: LegacyFilter[] | SearchFilters
+  // duplicate to store UI information about filters
+  queryUI?: SearchFilterGroup
```
samwho, line 70: I was under the impression we would no longer need to go down this road. What happened?

mike12345567: Discussed on Discord, the `SearchFilterGroup` is still required for the UI.

packages/shared-core/src/utils.ts
mike12345567: As this is in the JS realm it needs to protect against the possibility of undefined, as we can't really protect against it in the JS code.

packages/shared-core/src/filters.ts
mike12345567: Not a clue haha

packages/shared-core/src/filters.ts
mike12345567: Again unsure, but would prefer not to remove something that may have a very explicit purpose - it needs a string and perhaps there are things in the JS that configure the `operator` as a number etc.

packages/shared-core/src/filters.ts
mike12345567: Again this is in the JS realm so having the `undefined` check is just part of it, we can't protect from this on the caller side.

samwho (approved):

mike12345567:
Converting to draft PR - @aptkingston let us know when there is time to get this ready to be pulled in!

adrinr (approved):

aptkingston:
As long as we don't pull it into V3 UI (which will break the feature branch) it can go into master any time! After the RABC stuff is done I'll circle back to this and do the required updates on V3 UI to support the new `queryUI` thing.

PR #14676, opened 2024-10-01
Fix the binding drawer for default values.

samwho:
## Description

Prior to this PR the default values binding draw included the row context, which is incorrect as we don't allow users to use the row context in default value bindings. This PR fixes that.

It also fixes a case where when you're adding a column to a table without any columns, no binding drawer was shown at all.


aptkingston (approved):
LGTM!

PR #14678, opened 2024-10-01
Implement `GROUP BY` for calculation views.

samwho:
## Description

This PR implements the ability for calculation views to GROUP BY fields existing on the underlying table. They do this by specifying them as normal view fields as you would for any other view. Under the hood, we infer that these fields should always be grouped by because there's no other logical thing to do with them.


packages/backend-core/src/sql/sql.ts
```diff
@@ -1508,18 +1519,35 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
...
+      table.sourceId === INTERNAL_TABLE_SOURCE_ID
+    ) {
+      if (!table._id) {
+        return
+      }
+      // SQS uses the table ID rather than the table name
+      name = table._id
+    }
+    return aliases?.[name] ? aliases[name] : name
+  }
```
samwho, line 1538: There was logic scattered around a few places on how to determine a table name in the light of SQS being slightly different to other SQL sources, and the potential for table names to be aliased to shorter strings.

I consolidated all of that logic in to this one place and made everywhere* use it.

*not quite, there's a class later in this file that has it's own distinct implementation because it's slightly different

packages/backend-core/src/sql/sql.ts
```diff
@@ -813,30 +799,66 @@ class InternalBuilder {
     return query
   }
 
+  isSqs(t?: Table): boolean {
```
adrinr, line 802: NIT: I am not too fan of these optional parameters in here, as it might be easy to miss in some refactors or just while using it. This allows the caller to use `isSqs()`, skipping the parameter, so it's not obvious if it's meant to be passed or not.
Also, 1-letter parameter in functions makes the code harder to follow

packages/backend-core/src/sql/sql.ts
```diff
@@ -813,30 +799,66 @@ class InternalBuilder {
...
-      this.table.name && aliases?.[this.table.name]
-        ? aliases[this.table.name]
-        : this.table.name
-    if (!primary) {
+    if (!this.table.primary) {
       throw new Error("SQL counting requires primary key to be supplied")
     }
-    return query.countDistinct(`${aliased}.${primary[0]} as total`)
+    return query.countDistinct(
+      `${this.getTableName()}.${this.table.primary[0]} as total`
```
adrinr, line 845: Can we have issues here with tables containing a `total` column? Should we prefix them to avoid conflicts?

packages/backend-core/src/sql/sql.ts
```diff
@@ -1508,23 +1527,40 @@ class SqlQueryBuilder extends SqlTableQueryBuilder {
...
+      table.sourceType === TableSourceType.INTERNAL ||
+      table.sourceId === INTERNAL_TABLE_SOURCE_ID
+    ) {
+      if (!table._id) {
+        return
+      }
+      // SQS uses the table ID rather than the table name
+      name = table._id
+    }
+    return aliases?.[name] ? aliases[name] : name
```
adrinr, line 1545:
```suggestion
    return aliases?.[name] || name
```

packages/backend-core/src/sql/sql.ts
samwho: This did start life having no parameter at all but there's an annoying, singular case where the table that's being checked is _not_ `this.table`, which made me make this take a param and then because most of the uses would pass `this.table` it felt a bit nicer to make this optional.

The 1 letter param name is just because if I call it `table` I then have to call the const a little below it something else and that felt tedious.

Happy to go with another way of doing it if you have any suggestions.

packages/backend-core/src/sql/sql.ts
samwho: Good question. I'll change it to be safe.

packages/backend-core/src/sql/sql.ts
adrinr: I see the reasoning. Maybe this might be cleaned with a "private method" that requires the table, and a public one that will use the `privateIsSqs(this.table)`?

packages/backend-core/src/sql/sql.ts
```diff
@@ -813,30 +806,63 @@ class InternalBuilder {
...
-      this.table.name && aliases?.[this.table.name]
-        ? aliases[this.table.name]
-        : this.table.name
-    if (!primary) {
+    if (!this.table.primary) {
       throw new Error("SQL counting requires primary key to be supplied")
     }
-    return query.countDistinct(`${aliased}.${primary[0]} as total`)
+    return query.countDistinct(
+      `${this.getTableName()}.${this.table.primary[0]} as __bb_total`
```
adrinr, line 849: Should we have this magic string (`__bb_total`) stored in an enum or a similar constant?

adrinr (approved):
Nice and easy :D

packages/backend-core/src/sql/sql.ts
samwho: I'll sneak that into the next PR.

PR #14679, opened 2024-10-01
Test all aggregation types.

samwho:
## Description

This PR adds tests for all view calculation aggregation types (count, sum, avg, min, max). Through these tests I've discovered that `COUNT(*)` returns a string in MySQL, Postgres, and MariaDB. I've made sure to coerce them to number before returning through the API.


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -2490,6 +2490,61 @@ describe.each([
...
+                  return numbers.reduce((a, b) => a + b, 0) / numbers.length
+                case CalculationType.MIN:
+                  return Math.min(...numbers)
+                case CalculationType.MAX:
+                  return Math.max(...numbers)
+              }
+            }
+
+            const prices = rows.map(row => row.price)
+            const expected = calculate(type, prices)
```
melohagan, line 2536: NAB: Why not put the expected aggregate into the test input instead?

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -2490,6 +2490,61 @@ describe.each([
...
+            const actual = response.rows[0].aggregate
+
+            if (type === CalculationType.AVG) {
+              // The average calculation can introduce floating point rounding
+              // errors, so we need to compare to within a small margin of
+              // error.
+              expect(actual).toBeCloseTo(expected)
```
melohagan, line 2543: NAB: Nice to know! Curious what precision this works to?

packages/server/src/integrations/postgres.ts
```diff
@@ -272,7 +272,8 @@ class PostgresIntegration extends Sql implements DatasourcePlus {
     try {
       const bindings = query.bindings || []
       this.log(query.sql, bindings)
-      return await client.query(query.sql, bindings)
+      const result = await client.query(query.sql, bindings)
+      return result
```
melohagan, line 276: NAB: Why?

melohagan (approved):

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho: Because the test data is generated randomly.

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho: 2 decimal places by default. You can change it but I didn't feel the need to.

packages/server/src/integrations/postgres.ts
samwho: Ah, I did this so I could set a breakpoint with a way to check what was in here. Forgot to put it back when I was done.

packages/server/src/api/routes/tests/viewV2.spec.ts
melohagan: Makes sense 🙂

PR #14680, opened 2024-10-01
Refresh data when adding columns

aptkingston:
## Description
This PR refreshes data in grids when adding columns. We actually already do this when editing columns, we just don't when adding which is an oversight.

Before:

https://github.com/user-attachments/assets/d3d4f89f-80d9-43ed-a29a-bdf4fcf1087c

After:

https://github.com/user-attachments/assets/9ac4ab6c-8ce7-4928-b722-163996655ea7


shogunpurple (commented):
LGTM! Nice one

shogunpurple (approved):

PR #14681, opened 2024-10-01
Make sure calculation views are created and returned correctly from the API.

samwho:
## Description

Adds code and tests to make sure calculation views are treated correctly by the create and get view APIs. Prior to this, calculation fields were being stripped due to not being part of the table schema.


packages/server/src/sdk/app/views/index.ts
```diff
@@ -299,15 +292,24 @@ export async function enrichSchema(
...
+  const visibleTableFields = Object.keys(tableSchema).filter(
+    key => tableSchema[key].visible !== false
+  )
+  const visibleFields = new Set([...visibleSchemaFields, ...visibleTableFields])
+  for (const key of visibleFields) {
     // if nothing specified in view, then it is not visible
     const ui = viewSchema[key] || { visible: false }
     schema[key] = {
-      ...tableSchema[key],
+      ...(tableSchema[key] || {}),
```
adrinr, line 310: Is this required? Spread should work even if it is undefined

packages/server/src/sdk/app/views/index.ts
```diff
@@ -299,15 +292,24 @@ export async function enrichSchema(
...
   const anyViewOrder = Object.values(viewSchema).some(ui => ui.order != null)
-  for (const key of Object.keys(tableSchema).filter(
-    k => tableSchema[k].visible !== false
-  )) {
+
+  const visibleSchemaFields = Object.keys(viewSchema).filter(key => {
+    if (helpers.views.isCalculationField(viewSchema[key])) {
+      return viewSchema[key].visible !== false
+    }
+    return key in tableSchema && tableSchema[key].visible !== false
```
adrinr, line 300: Is this not bypassing the view visibility?

adrinr (changes requested):

packages/server/src/sdk/app/views/index.ts
samwho: Huh, I didn't know this. I assumed it wouldn't work. Will put this back how it was!

packages/server/src/sdk/app/views/index.ts
samwho: I broke some tests while working on this and the current state of the code passes the various visibility tests I broke. I haven't changed any visibility tests. In what way would this break view visibility?

adrinr: Interesting... Good we have tests for this. I were just worried that we do `const visibleFields = new Set([...visibleSchemaFields, ...visibleTableFields])`, including the table fields. It might be trimmed in some post-process then

adrinr (approved):

PR #14689, opened 2024-10-02
Fix browser crash under certain circumstances with forms

aptkingston:
## Description
This is an extremely nasty bug that was very hard to find, and caused your browser to lock up. It's existed for a very long time and may explain some of the random browser lock up issues we sometimes see.

The core issue was that we were mutating an array in-place while reversing it, and under certain circumstances this lead to an endless loop.

## Addresses
- https://linear.app/budibase/issue/BUDI-8688/builder-crashes-in-preview-with-modal-data-provider-repeater-form


deanhannigan (approved):
LGTM!

PR #14691, opened 2024-10-02
Default values for options and array types (backend)

aptkingston:
## Description
This PR allows default values to be set for options and array types.
Tests are included.
Frontend will be in a separate PR to V3-UI to avoid conflicts.

## Addresses
- https://linear.app/budibase/issue/BUDI-8678/add-the-ability-to-apply-default-values-to-single-and-multi-select


samwho (approved):
Thanks for splitting these up, and adding tests 🙏

PR #14692, opened 2024-10-02
More view calculation tests, and COUNT(DISTINCT ...) implementation.

samwho:
## Description

This PR adds a new suite of tests for updating calculation views.

It also adds implementation for `COUNT(DISTINCT ...)` through calculation views, which wasn't part of the requirements but was easy enough to include that I thought why not.

The API for adding a `COUNT(DISTINCT ...)` is to add a view schema field that looks like:

```js
{
  visible: true,
  calculationType: "count",
  distinct: true,
  field: "foo"
}
```

The `field` argument, when `distinct` is false or unspecified, is not required.

This PR also includes some fixes for CLOBs in OracleDB in relation to calculation views.


aptkingston (commented):
Thanks Sam! Looking forward to getting this in 👍

mike12345567 (approved):
LGTM - good set of tests as well!

PR #14693, opened 2024-10-03
Pull the error object out of isolated-vm when a user script throws an error.

samwho:
## Description

This is very much a stepping stone PR. It is an improvement on the status-quo but by no means the last work that should be done on improving these error messages.

Concerns:

- I've changed the API of `processStringSync` a bit. Where it used to just return the binding pre-evaluation if an error happened, now for the `js` helper it will throw a `UserScriptError` that I have caught in the relevant places. I may have missed some places, so please scrutinise deeply.
- This doesn't tell you _where_ the error happened, what line. I would like to try doing that in a follow-up PR.

## Screenshots

JS:

CleanShot 2024-10-03 at 11 32 47@2x

Before:

CleanShot 2024-10-03 at 12 10 41@2x

After:

CleanShot 2024-10-03 at 12 11 59@2x

Bindings panel:

image

## Launchcontrol

- Returns the error type and message from JS execution errors, instead of the generic "Error while executing JS" message.


packages/string-templates/src/index.ts
```diff
@@ -463,7 +467,27 @@ export function defaultJSSetup() {
         setTimeout: undefined,
       }
       createContext(context)
-      return runInNewContext(js, context, { timeout: 1000 })
+
+      js = `
```
melohagan, line 471: If you are always overwriting the **js** argument, should you remove the parameters from the signature?

```ts
setJSRunner((js: string, context: Context) ...
```

becomes

```ts
setJSRunner((context: Context) ...
```

packages/string-templates/src/index.ts
```diff
@@ -229,7 +230,10 @@ export function processStringSync(
     } else {
       return process(string)
     }
-  } catch (err) {
+  } catch (err: any) {
```
melohagan, line 233: NAB: Is any needed?

melohagan (approved):
LGTM! Awesome feature 😁

packages/string-templates/src/helpers/javascript.ts
```diff
@@ -100,6 +100,9 @@ export function processJS(handlebars: string, context: any) {
     if (error.name === "ExecutionTimeoutError") {
       return "Request JS execution limit hit"
     }
+    if (error.code === "USER_SCRIPT_ERROR") {
```
adrinr, line 103: This could be `UserScriptError.code` if made static, to avoid magic strings

packages/string-templates/src/index.ts
samwho: Sadly yes, without it there's a type error that fails the library tests.

mike12345567 (approved):
LGTM - massive improvement!

PR #14695, opened 2024-10-03
Handle internal relationships referencing rows that don't exist

mike12345567:
## Description
Fixing an issue with corrupt relationship records referencing rows which don't exist, this is a temporary measure as these relationships should be cleaned up correctly but for now ignore any which reference rows which no longer exist.


packages/server/src/db/linkedRows/LinkController.ts
```diff
@@ -221,9 +221,15 @@ class LinkController {
                 link.id !== row._id && link.fieldName === linkedSchema.name
             )
 
+            // check all the related rows exist
+            const foundRecords = await this._db.getMultiple(
+              links.map(l => l.id),
+              { allowMissing: true, excludeDocs: true }
+            )
```
samwho, line 228: I'm confused, this doesn't seem to line up with the PR description. It looks like you're checking for existing links to a document on save, not looking for links to documents that don't exist?

mike12345567: we're deciding whether or not the validation is correct by getting a list of rows, the way the system works it retrieves a list of link documents, this then makes sure those link documents reference something valid.

samwho (approved):

adrinr (approved):

Issue #14533, opened 2024-09-06
Postgres Integration: json query parsing leads to broken queries


sdsys-ch:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.28.4

**Describe the bug**
Adding a new SQL query for a Postgres DB (haven't checked internal or others) results in `Query Error: syntax error at or near "\"` if the query contains a json constructor (pot. only related to COALESCE statements, but don't think s - untested).

Looking at the logfiles, the query contains `\n` or '\\t' chars that haven't been stripped by the sql preprocessing code apparently:

```
bbpostgres  | 	SELECT
bbpostgres  | 	    i.*,
bbpostgres  | 	    COALESCE(json_build_object('contact', ic.contact_info),'{"contact": {}}'::json) AS contact_details,\n    COALESCE(json_build_object('name', in.name_info),'{"name": {}}'::json) AS name_details
```

Removing the line breaks in the create query window makes the query succeed.

**To Reproduce**

Create table and table data in your DB:

```
CREATE TABLE individuals (
    id SERIAL PRIMARY KEY,
    first_name VARCHAR(50) NOT NULL,
    last_name VARCHAR(50) NOT NULL,
    email VARCHAR(100),
    phone VARCHAR(20),
    birth_date DATE,
    nationality VARCHAR(50)
);

-- sample data
INSERT INTO individuals (first_name, last_name, email, phone, birth_date, nationality) VALUES
('Jane', 'Austen', 'jane.austen@lite....com', '1775-1817', '1775-12-16', 'English'),
('Ernest', 'Hemingway', 'ernest@lost....org', NULL, '1899-07-21', 'American'),
('Virginia', 'Woolf', NULL, '1882-1941', '1882-01-25', 'British'),
('Franz', 'Kafka', NULL, NULL, '1883-07-03', 'Czech'),
('Gabriel', 'García Márquez', 'gabo@mag....com', '1927-2014', '1927-03-06', 'Colombian'),
('Agatha', 'Christie', 'agatha@myst....com', '1890-1976', '1890-09-15', 'British'),
('Leo', 'Tolstoy', NULL, '1828-1910', '1828-09-09', 'Russian'),
('Chimamanda', 'Ngozi Adichie', 'chimamanda@con....lit', NULL, '1977-09-15', 'Nigerian');
```

[...]

PR #14699, opened 2024-10-03
Postgres - newlines between JSON aggregations

mike12345567:
## Description
Fixing an issue with newlines between coalesce statements in Postgres - we were escaping newlines even if they were valid when given a list of JSON operations to perform.

## Addresses
- https://github.com/Budibase/budibase/issues/14533
- https://linear.app/budibase/issue/BUDI-8629/postgres-integration-json-query-parsing-leads-to-broken-queries


samwho (approved):

packages/server/src/integrations/postgres.ts
```diff
@@ -41,7 +41,7 @@ if (types) {
   types.setTypeParser(1184, (val: any) => val) // timestampz
 }
 
-const JSON_REGEX = /'{.*}'::json/s
+const JSON_REGEX = /'{\s*.*?\s*}'::json/gs
```
adrinr, line 44: Is this just to exclude empty `{}`? And if the `g` flag is needed, can we add a test that processes multiple matches, to ensure this is not deleted in the future?

mike12345567: Its to make sure multiple matches don't get combined into one - the test case I've added with this is already checking for that (does the JSON statement twice).

PR #14700, opened 2024-10-04
Add uiMetadata to viewV2

aptkingston:
## Description
Tiny PR to add a UI metadata property to views. This is needed for the new view calculations feature. It's totally opt-in and is untouched by the server so is completely safe to merge into master without feature flags.

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -177,6 +177,9 @@ describe.each([
               visible: true,
             },
           },
+          uiMetadata: {
+            foo: "bar",
+          },
         }
         const res = await config.api.viewV2.create(newView)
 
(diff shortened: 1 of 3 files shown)


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -177,6 +177,9 @@ describe.each([
               visible: true,
             },
           },
+          uiMetadata: {
+            foo: "bar",
+          },
```
samwho, line 182: You probably want at least one assertion somewhere to make sure we do actually return this when fetching views from the API.

aptkingston: That's what the update test does already isn't it? It's asserting that the response from the update call matches the payload sent up, which includes the `uiMetadata` field that should have been saved.

samwho: Oh I completely missed that, sorry!

samwho (approved):

adrinr (approved):
Nice to see this ui metadata not directly added in the schema :)

PR #14701, opened 2024-10-04
Fixes an issue where id and revision weren't passed to row action automations

PClmnt:
## Description
`id` and `revision` weren't being populated for a row action automation trigger. 

## Addresses
https://linear.app/budibase/issue/BUDI-8690/automation-update-row-step-creates-a-row


aptkingston (approved):
Nice one, thanks Peter!

PR #14702, opened 2024-10-04
Validate that you cannot create a calculation view with more than 5 calculation fields.

samwho:
## Description

We're keeping a lid on how many aggregates can be put onto calculation views to make sure they stay within reasonable bounds, both for performance and complexity reasons.


adrinr (approved):
Is 5 not a really low number? Do we know the performance impact of having many of these?

samwho:
I've not done performance testing myself but I think the last time we spoke about this we decided we would err on the side of caution. It is much easier to raise a limit than it is to lower it.

PR #14703, opened 2024-10-04
persist app sort method selection by saving it against the user

andz-bb:
## Description
Every time you load the apps list in the portal it defaults to sorting the apps alphabetically. If you select a different method of sorting it works but then the selection is lost on refresh. This PR adds a new property to the user `appSort` which is then updated when a new sorting method is selected by the user. This is then used to sort the apps on subsequent page loads so the user preference is persisted.

## Launchcontrol

save the user's app sorting method selection in the portal apps screen


packages/builder/src/stores/portal/apps.js
```diff
@@ -140,43 +148,50 @@ export class AppsStore extends BudiStore {
...
+      ? $store.apps.map(app => ({
+          ...app,
+          deployed: app.status === AppStatus.DEPLOYED,
+          lockedYou: app.lockedBy && app.lockedBy.email === $auth.user?.email,
+          lockedOther: app.lockedBy && app.lockedBy.email !== $auth.user?.email,
+          favourite: $auth.user?.appFavourites?.includes(app.appId),
+        }))
+      : []
+
+    if ($sortBy === "status") {
```
andz-bb, line 169: $store.sortBy -> $sortBy on lines 169 and 179

sorry for the messy diff

aptkingston (approved):
Really clean implementation @andz-bb!

- Excellent derived store usage
- Instant UI updates with async persistence
- Full backwards compatibility

Perfect. Huge QoL improvement as well!

samwho (approved):

PR #14704, opened 2024-10-04
Validate that there are no duplicate calculations in calculation views.

samwho:
## Description

We don't want users to have multiple columns doing the same calculation in a calculation view, so this PR adds validation that throws an error on creation and save if there are multiple columns doing the same work.


adrinr (commented):
What is the reasoning for this? This will require disabling the duplicate column and other. I don't see many use cases for users to do this, but if they want to do it (to keep a copy of a column while playing with some editions, for example), is there any downside that we want to avoid?

samwho:
I have no idea what the reasoning is, I'm going off what's written in https://linear.app/budibase/issue/BUDI-8611/calculation-view-crud-api-updates. @mike12345567 would be the person to ask but he's out until Wednesday.

One potential upside for doing this that I can think of is for if users accidentally try to create identical columns without realising. Maybe they typed the wrong field and meant a different one.

adrinr:
Happy to go with it if already working. Could you check what happen when we try to "duplicate" a column via UI?

samwho:
I think to do that we'll need to merge this to master and then merge master into the v3-ui branch.

adrinr (approved):

PR #14705, opened 2024-10-04
Make the new throwing behaviour of `processStringSync` opt-in.

samwho:
## Description

Because there's only really one place that needs an actual exception to be thrown from `processStringSync` when JS execution fails (the binding panel when you're editing them), I've changed this behaviour to be opt-in. Everywhere else will get a string back containing the error message.


aptkingston (approved):
Definitely prefer this approach - excellent idea to make it optional!

PR #14709, opened 2024-10-04
Make new tables require ADMIN permissions to read and write.

samwho:
## Description

As we're planning for users to make more use of views for access-control going forward, having table permissions default to `BASIC` defeats the purpose of that. So tables created after this PR goes on will default to `ADMIN` level read/write permissions.

## Addresses
- https://linear.app/budibase/issue/BUDI-8686/new-tables-should-be-admin-by-default

## Launchcontrol

- Newly created tables will default to needing admin-level permissions to read and write.

## Feature branch env
[Feature Branch Link](http://fb.fb.qa.budibase.net)

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/ManageAccessModal.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/ManageAccessModal.svelte
@@ -94,7 +94,7 @@
   loadDependantInfo()
 </script>
 
-<ModalContent showCancelButton={false} confirmText="Done">
+<ModalContent showCancelButton={false} showConfirmButton={false}>
   <span slot="header">
     Manage Access
     {#if requiresPlanToModify}
--- a/packages/server/src/utilities/security.ts
+++ b/packages/server/src/utilities/security.ts
@@ -1,5 +1,6 @@
 import { permissions, roles } from "@budibase/backend-core"
 import { DocumentType, VirtualDocumentType } from "../db/utils"
+import { getDocumentType, getVirtualDocumentType } from "@budibase/types"
 
 export const CURRENTLY_SUPPORTED_LEVELS: string[] = [
   permissions.PermissionLevel.WRITE,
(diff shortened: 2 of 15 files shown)


packages/server/src/utilities/security.ts
```diff
@@ -8,14 +9,16 @@ export const CURRENTLY_SUPPORTED_LEVELS: string[] = [
 ]
 
 export function getPermissionType(resourceId: string) {
-  const docType = Object.values(DocumentType).filter(docType =>
-    resourceId.startsWith(docType)
-  )[0]
```
samwho, line 13: There's some light refactoring in this file because my implementation originally went in here but I later switched tactics. I did, however, discover a bug in this code here. Because `DocumentType`s can overlap (e.g. `datasource_` is a subset of `datasource_plus_`), you need to search through this list in order of longest to shortest. So I extracted a helper to do just that.

adrinr (dismissed):

packages/backend-core/src/security/tests/permissions.spec.ts
```diff
@@ -17,10 +17,6 @@ describe("levelToNumber", () => {
     expect(permissions.levelToNumber(permissions.PermissionLevel.WRITE)).toBe(2)
   })
 
-  it("should return 3 for ADMIN", () => {
-    expect(permissions.levelToNumber(permissions.PermissionLevel.ADMIN)).toBe(3)
-  })
```
samwho, line 22: How come this and the test below are being removed?

packages/types/src/sdk/permissions.ts
```diff
@@ -2,7 +2,6 @@ export enum PermissionLevel {
   READ = "read",
   WRITE = "write",
   EXECUTE = "execute",
-  ADMIN = "admin",
```
samwho, line 5: Was this unused?

packages/backend-core/src/security/tests/permissions.spec.ts
adrinr: This was causing a lot of confusion while trying to use the new "base" permission, and it should not be used. It used to be used before the app creator changes, that was a while ago.
But because cleaning it is not needed anymore, and at the risk of breaking some really old installations, I just undid these removals

packages/types/src/sdk/permissions.ts
adrinr: Same as above

packages/builder/src/components/backend/DataTable/modals/ManageAccessModal.svelte
```diff
@@ -94,7 +94,7 @@
   loadDependantInfo()
 </script>
 
-<ModalContent showCancelButton={false} confirmText="Done">
+<ModalContent showCancelButton={false} showConfirmButton={false}>
```
adrinr, line 97: This button does nothing, and it can give a false impression of an action not being persisted. The update on the dropdowns trigger the "save"

https://github.com/user-attachments/assets/570b65e6-b624-4491-b07f-5012f439b0ab

samwho:
Approved ✅

adrinr (approved):
Nice job 🚀

PR #14710, opened 2024-10-04
Freeze the JS context instead of cloning it.

samwho:
## Description

We're seeing a lot of time spent in the `deepClone` inside of `processJS` in production. I'm attempting in this PR to freeze it instead. The goal is to prevent mutation, and this should do that without copying anything and hopefully a lot faster.

You'll see in the benchmark below that freezing is actually _slower_ than copying, but the benefit would be that we only freeze once. So this will benefit requests with lots of formulas, while hurting requests with only 1-2.

## Benchmark

Script used:

<details>

```typescript
import Benchmark from "benchmark";
import { cloneDeep } from "lodash";

function deepFreeze<T>(o: T): Readonly<T> {
  if (o == null) {
    return o;
  }

  // We assume that if the top-level object is frozen, it has likely already
  // been frozen by this function and we don't recurse.
  if (Object.isFrozen(o)) {
    return o;
  }

  Object.freeze(o);

  Object.getOwnPropertyNames(o).forEach(function (prop) {
    const key = prop as keyof typeof o;
    if (
      o[key] !== null &&
      (typeof o[key] === "object" || typeof o[key] === "function") &&
      !Object.isFrozen(o[key])
    ) {
      deepFreeze(o[key]);
    }
  });

  return o;
}

function buildObject(depth: number, keys: number) {
  const o: any = {};
  for (let i = 0; i < keys; i++) {
    if (depth === 1) {
      o[i] = i;
      continue;
    } else {
      o[i] = buildObject(depth - 1, keys);
    }
  }
  return o;
}

const suite = new Benchmark.Suite();

suite.add("buildObject small", () => {
  buildObject(1, 100);
});

suite.add("buildObject medium", () => {
  buildObject(2, 100);
});

suite.add("buildObject large", () => {
  buildObject(3, 100);
});

suite.add("deepFreeze small", () => {
  const obj = buildObject(1, 100);
  deepFreeze(obj);
});

suite.add("deepFreeze medium", () => {
  const obj = buildObject(2, 100);
  deepFreeze(obj);
});

suite.add("deepFreeze large", () => {
  const obj = buildObject(3, 100);
  deepFreeze(obj);
});

suite.add("cloneDeep small", () => {
  const obj = buildObject(1, 100);
  cloneDeep(obj);
});

suite.add("cloneDeep medium", () => {
  const obj = buildObject(2, 100);
  cloneDeep(obj);
});

suite.add("cloneDeep large", () => {
  const obj = buildObject(3, 100);
  cloneDeep(obj);
});

suite.on("cycle", function (event: any) {
  console.log(String(event.target));
});

suite.run();
```

</details>

Results:

```
buildObject small  x 3,561,547 ops/sec ±0.40%  (98 runs sampled)
buildObject medium x 32,567    ops/sec ±0.76%  (90 runs sampled)
buildObject large  x 280       ops/sec ±1.18%  (81 runs sampled)
deepFreeze  small  x 50,162    ops/sec ±4.26%  (97 runs sampled)
deepFreeze  medium x 518       ops/sec ±0.29%  (95 runs sampled)
deepFreeze  large  x 5.01      ops/sec ±2.45%  (17 runs sampled)
cloneDeep   small  x 66,381    ops/sec ±10.32% (75 runs sampled)
cloneDeep   medium x 813       ops/sec ±2.31%  (93 runs sampled)
cloneDeep   large  x 7.98      ops/sec ±5.35%  (24 runs sampled)
```

### DIFF
--- a/packages/string-templates/package.json
+++ b/packages/string-templates/package.json
@@ -26,7 +26,7 @@
     "manifest": "ts-node ./scripts/gen-collection-info.ts"
   },
   "dependencies": {
-    "@budibase/handlebars-helpers": "^0.13.2",
+    "@budibase/handlebars-helpers": "^0.14.0",
     "dayjs": "^1.10.8",
     "handlebars": "^4.7.8",
     "lodash.clonedeep": "^4.5.0"
--- a/packages/string-templates/src/helpers/javascript.ts
+++ b/packages/string-templates/src/helpers/javascript.ts
@@ -1,5 +1,4 @@
 import { atob, isJSAllowed } from "../utilities"
-import cloneDeep from "lodash/fp/cloneDeep"
 import { LITERAL_MARKER } from "../helpers/constants"
 import { getJsHelperList } from "./list"
 import { iifeWrapper } from "../iife"
--- a/packages/string-templates/src/helpers/javascript.ts
+++ b/packages/string-templates/src/helpers/javascript.ts
@@ -49,7 +49,7 @@ const getContextValue = (path: string, context: any) => {
 }
 
 function deepFreeze<T>(o: T): Readonly<T> {
-  if (o === undefined) {
+  if (o == null) {
     return o
   }
 
--- a/packages/string-templates/src/helpers/javascript.ts
+++ b/packages/string-templates/src/helpers/javascript.ts
@@ -49,6 +49,16 @@ const getContextValue = (path: string, context: any) => {
   return data
 }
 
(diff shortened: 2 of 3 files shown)


samwho:
Closing this because @adrinr has reminded me that `isolated-vm` already copies the context over the isolation boundary, so on the backend this deep clone is totally redundant. I'm just going to remove it.

PR #14717, opened 2024-10-07
Fix SQS error handling.

samwho:
## Description

We were attempting to parse the body as JSON before we knew it was JSON, resulting in parse failures before the error handling could kick in.


packages/backend-core/src/db/couch/DatabaseImpl.ts
```diff
@@ -371,11 +371,15 @@ export class DatabaseImpl implements Database {
     return this.performCall(() => {
       return async () => {
         const response = await directCouchUrlCall(args)
-        const json = await response.json()
-        if (response.status > 300) {
-          throw json
+        if (response.status >= 300) {
+          const text = await response.text()
```
adrinr, line 375: Will this always be text?

adrinr (approved):

packages/backend-core/src/db/couch/DatabaseImpl.ts
samwho: Correct me if I'm wrong but I thought `.text()` just got the plain old body contents without trying to parse it, so I thought it was always safe to use.

adrinr: Yes, you are right!

PR #14718, opened 2024-10-07
JS backend deep clone optimisation.

samwho:
## Description

On the backend, the implementation of `string-templates`' `getContextValue` is used by the return value of it gets copied when it crosses the `isolated-vm` memory boundary. This makes the `deepClone` in `processJS` redundant, so I'm removing it.

I did notice that the frontend works slightly differently to the backend and left a comment about it on the code.

### DIFF
--- a/packages/string-templates/src/helpers/javascript.ts
+++ b/packages/string-templates/src/helpers/javascript.ts
@@ -58,18 +58,7 @@ const getContextValue = (path: string, context: any) => {
     data = data[removeSquareBrackets(key)]
   })
 
-  // When data is returned from this function in the backend, isolated-vm copies
-  // it across the vm boundary. We need to replicate that behaviour on the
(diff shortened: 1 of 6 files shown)


aptkingston (approved):
LGTM! Well found.

packages/string-templates/src/helpers/javascript.ts
```diff
@@ -70,9 +89,8 @@ export function processJS(handlebars: string, context: any) {
     // Our $ context function gets a value from context.
     // We clone the context to avoid mutation in the binding affecting real
```
adrinr, line 94: This comment need updating

adrinr (commented):
Will this not add breaking changes for js snippets that run only on frontend?

adrinr (approved):

PR #14720, opened 2024-10-07
Allow calculation views to hide required fields.

samwho:
## Description

This was raised by @aptkingston while implementing the frontend for this feature. When we return a view from the views API, all of the table fields that are not present on the view are "enriched" onto it but set to `visible: false`. So when making modifications to the view, the update request sent will include these invisible fields.

We need to support calculation views being allowed to hide required fields, which I believe is okay because rows cannot be written to calculation views.


aptkingston (approved):

packages/server/src/sdk/app/views/index.ts
```diff
@@ -178,14 +178,15 @@ function checkRequiredFields(
       continue
     }
 
-    if (!viewSchemaField?.visible) {
+    if (!helpers.views.isCalculationView(view) && !viewSchemaField?.visible) {
```
adrinr, line 181: Should this not be done first thing it this function? Or even in the caller, as it's a private method?

adrinr (changes requested):

packages/server/src/sdk/app/views/index.ts
samwho: How come? Is it not relevant to only this conditional?

adrinr: This will be true for any field in a calculation view, will it not? If the view is a calculation one, no required fields are actually needed. Or did I get this wrong?

samwho: Ohh you're suggesting I skip `checkRequiredFields` entirely for calculation views? That may actually be the correct play, let me check.

samwho: Done!

packages/server/src/sdk/app/views/index.ts
```diff
@@ -114,7 +114,11 @@ async function guardViewSchema(
   }
 
   await checkReadonlyFields(table, view)
-  checkRequiredFields(table, view)
+
+  if (!helpers.views.isCalculationView(view)) {
```
adrinr, line 118: Yup, this is what I meant. Thanks!

adrinr (approved):

PR #14723, opened 2024-10-07
Mark calculation views explicitly instead of figuring it out implicitly.

samwho:
## Description

This PR makes it such that calculation views require `type: "calculation"` to be passed on their creation, and once set you cannot change the `type` of a view.

We decided to do this because validation rules on what you can do with field visibility are different for the two types of view, and the ability for a view to transition between being a calculation view or not just by the presence of calculation fields makes validation problematic. It is possible to "brick" a table or view by transitioning what type it is.


aptkingston (approved):
LGTM! Thanks Sam.

packages/types/src/documents/app/view.ts
```diff
@@ -79,10 +79,15 @@ export enum CalculationType {
   MAX = "max",
 }
 
+export enum ViewV2Type {
+  CALCULATION = "calculation",
```
adrinr, line 83: Should we have a name for the 'non-calculation' views, and to start populating them as well?

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -127,13 +127,13 @@ export async function create(ctx: Ctx<CreateViewRequest, ViewResponse>) {
...
   const parsedView: Omit<RequiredKeys<ViewV2>, "id" | "version"> = {
     name: view.name,
+    type: view.type,
     tableId: view.tableId,
     query: view.query,
     queryUI: view.queryUI,
     sort: view.sort,
     schema,
     primaryDisplay: view.primaryDisplay,
-    uiMetadata: view.uiMetadata,
```
adrinr, line 136: Is this meant to be removed?

packages/types/src/documents/app/view.ts
aptkingston: I thought this too, but unless we run a migration to backfill it then the inconsistency would annoy me. I don't mind either way though.

packages/server/src/api/controllers/view/viewsV2.ts
samwho: Yes. Its only purpose prior to this was for the frontend to identify calculation views. This is now possible with `type`.

packages/types/src/documents/app/view.ts
samwho: I would argue no, it's a bit less messy to assume if type is not there, it's a "vanilla" ViewV2. If we start populating it, we will infinitely have to check for either not present, or equal to some value. I'd prefer one state to two.

adrinr (approved):
We should discuss things like adding the types. Probably we don't need to backfill it in the dbs, but it would be nice if the apis were populating these default values if not populated, so we get consistent APIs. This is a good middle-ground between half-baked types and backfilling

samwho:
If we go down that road what I think we really need to do is split our types into internal and external. The API layer should be a distinct beast to what we handle internally. I don't think there is a way for us to have one set of types we work with given the way our code has evolved.

PR #14724, opened 2024-10-07
Fix enterprise vulns

shogunpurple:
## Description
Fixing a vuln for an enterprise client picked up by their scanner around XSS - we aren't susceptible to XSS, but the fact you can save it in the DB is still enough to trip the validation.

I've added a new env var - `XSS_SAFE_MODE`, that will actually prevent a row saving if any of it's string inputs return `true` in a regex match against potentially dangerous input.


packages/server/src/sdk/app/rows/tests/utils.spec.ts
```diff
@@ -332,4 +333,46 @@ describe("validate", () => {
...
+      "</script><script>alert('test')</script>",
+      "<div onmouseover=\"alert('XSS')\">Hover over me!</div>",
+      "'; EXEC sp_msforeachtable 'DROP TABLE ?'; --",
+      "{alert('Injected')}",
+      "UNION SELECT * FROM users",
+      "INSERT INTO users (username, password) VALUES ('admin', 'password')",
+      "/* This is a comment */ SELECT * FROM users",
+      '<iframe src="http://malicious-site.com"></iframe>',
+    ])("test potentially unsafe input: %s", async input => {
+      environment.XSS_SAFE_MODE = true
```
samwho, line 367: We have a helper called `withEnv` with nice types and makes sure everything is cleaned up correctly that would be handy here. If `validate` throws we leave the env dirty.

samwho (approved):

packages/server/src/sdk/app/rows/tests/utils.spec.ts
shogunpurple: Lovely - updated!

PR #14725, opened 2024-10-07
Test helper for setting feature flags.

samwho:
## Description

This PR introduces a helper for setting feature flags that:

1. Correctly cleans up after itself, the same way our env helpers do.
2. Can be safely nested, so you don't accidentally overwrite flags.


adrinr (approved):
Nice way to cleanup magic strings! ❤️

PR #14735, opened 2024-10-08
Make sure calculation views cannot be used to write or modify rows.

samwho:
## Description

Pretty much what it says in the title. Code and tests to make sure we can't modify data through calculation views.

### DIFF
--- a/packages/server/src/api/controllers/row/external.ts
+++ b/packages/server/src/api/controllers/row/external.ts
@@ -27,6 +27,7 @@ import {
 } from "../../../utilities/rowProcessor"
 import { cloneDeep } from "lodash"
 import { generateIdForRow } from "./utils"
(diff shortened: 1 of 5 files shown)


packages/server/src/api/controllers/row/external.ts
```diff
@@ -42,6 +43,11 @@ export async function handleRequest<T extends Operation>(
 
 export async function patch(ctx: UserCtx<PatchRowRequest, PatchRowResponse>) {
   const source = await utils.getSource(ctx)
+
+  if (sdk.views.isView(source) && helpers.views.isCalculationView(source)) {
```
mike12345567, line 47: We can't delete rows through a calculation view as well right? We should likely add that to this PR too.

mike12345567 (approved):

PR #14742, opened 2024-10-09
Validate link field names cannot overlap with table field names.

samwho:
## Description

Noticed this while creating a view test case: if the `fieldName` of a link field overlaps with a field name on the table itself all sorts of horrendous errors occur. So I'm checking for it on table creation to make sure users get a nice error message when it happens.


samwho:
Hit some strange errors with this that makes me think it works differently to how I think, going to leave it for now.

PR #14744, opened 2024-10-09
Still fetch flags when the user is not logged in.

samwho:
## Description

We weren't fetching flags from PostHog when the user had no login, resulting in public endpoints behaving quite differently (in this case, no SQS, which is required for some newer features).

This PR uses the hashed IP address of the logged our user as their user ID, and it also falls back to getting the current tenant ID if the one on the identity cannot be fetched.

## Feature branch env
[Feature Branch Link](http://fb.fb.qa.budibase.net)


mike12345567 (approved):
LGTM!

aptkingston (approved):
Nice one 👍 Not an easy scenario to etst!

samwho:
I've rejigged the way we fetch IP address and removed the code that reads flags from the license (this was unused anyway, should have no impact on anything). PTAL!

mike12345567 (approved):

PR #14748, opened 2024-10-09
Prevent deleting rows through a calculation view.

samwho:
## Description

Following up on https://github.com/Budibase/budibase/pull/14735#discussion_r1793339447.


mike12345567 (approved):
LGTM!

packages/server/src/api/controllers/row/external.ts
```diff
@@ -102,6 +103,11 @@ export async function patch(ctx: UserCtx<PatchRowRequest, PatchRowResponse>) {
 
 export async function destroy(ctx: UserCtx) {
   const source = await utils.getSource(ctx)
+
+  if (sdk.views.isView(source) && helpers.views.isCalculationView(source)) {
```
adrinr, line 107: Can this not be done at the higher level, so it is not duplicated between external and internal?

samwho: It can, though it would cause a refetch of the table/view. I put them where I did because it didn't cause any extra CouchDB fetches. It might be possible to have the best of both worlds but it would be a larger change.

adrinr: After discussing it, let's leave it as it is. This is the common pattern now, we should check if we can improve it in the future clean up, rethinking how we do differentiate between internal and external logic

Issue #14747, opened 2024-10-09
There seems to be issues with relationships in Budibase at the moment `Cannot convert undefined or null to object`


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 2.32.13
  - App Version: 2.32.13

**Describe the bug**
When trying to change the relationships in the owner table of the attached schema. It seems to throw an error. `Cannot convert undefined or null to object"`

**To Reproduce**
1. Import the attached app and schema
2. Configure your Postgres connection to match your local instance
3. Mess around with adding and removing the relationships from the owner's table

**Expected behavior**
Should be able to append and edit relationships in all relationship column types. Rather than getting an error `Cannot convert undefined or null to object"`

**Screenshots**
<img width="934" alt="Screenshot 2024-10-09 at 13 13 05" src="https://github.com/user-attachments/assets/83fe1fec-1307-4e48-9661-7596982e8561">

**App Export**
[bb relationship setup-export-1728477508419.tar.gz](https://github.com/user-attachments/files/17308243/bb.relationship.setup-export-1728477508419.tar.gz)

**Postgres Schema**
**Budibase generated relationship columns**
[BB relationship build.txt](https://github.com/user-attachments/files/17308146/BB.relationship.build.txt)

**SQL build foreign keys**
[precreated relationships.txt](https://github.com/user-attachments/files/17308174/precreated.relationships.txt)

PR #14749, opened 2024-10-09
Fix SQL removing related row from many side 

mike12345567:
## Description
This fixes an issue raised by @ConorWebb96 when attempting to remove related rows from a relationship specifically from the many side. Adding a test case to cover this in future.

## Addresses
- https://linear.app/budibase/issue/BUDI-8715/there-seems-to-be-issues-with-relationships-in-budibase-at-the-moment
- https://github.com/Budibase/budibase/issues/14747


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -743,9 +745,19 @@ export class ExternalRequest<T extends Operation> {
...
-      : await aliasing.queryWithAliasing(json, makeExternalQuery)
+    let response: DatasourcePlusQueryResponse
+    // there's a chance after input processing nothing needs updated, so pass over the call
+    // we might still need to perform other operations like updating the foreign keys on other rows
+    if (
+      this.operation === Operation.UPDATE &&
+      Object.keys(row || {}).length === 0
+    ) {
+      response = [config.row!]
+    } else {
```
samwho, line 756: I don't think I fully understand this 🤔 So if the row is empty we return `[{}]`?

packages/backend-core/src/sql/sql.ts
```diff
@@ -325,14 +325,18 @@ class InternalBuilder {
...
+        const { column } = this.splitter.run(key)
+        const schema = this.table.schema[column]
+        if (!schema) {
+          continue
+        }
+        body[key] = this.parse(value, schema)
       }
-      body[key] = this.parse(value, schema)
+    } catch (err) {
+      console.log(err)
```
samwho, line 339: Is it correct to just return the body as-is, un-parsed, when anything goes wrong in parsing? Feels like something that's destined to cause future bugs.

packages/server/src/api/controllers/row/ExternalRequest.ts
mike12345567: Ah - in this case the `row` is the processed row, which may have been stripped down if nothing has changed, while `config.row` is unprocessed - I'll rename to make it more obvious.

packages/backend-core/src/sql/sql.ts
mike12345567: Oh woops - I should have removed this try-catch, was just during my debugging, nice catch!

samwho (approved):

PR #14756, opened 2024-10-10
Fix table updates wiping out view calculation fields.

samwho:
## Description

Fixes another problem @aptkingston ran into: when updating a table, all of the view schemas are synced with the table. This would remove any field in the view not found in the table schema. I've added a check for calculation fields in there.


packages/server/src/api/routes/utils/validators.ts
```diff
@@ -20,29 +20,28 @@ const OPTIONAL_NUMBER = Joi.number().optional().allow(null)
...
+  const required = Object.keys(table.schema).filter(key =>
+    isRequired(table.schema[key].constraints)
+  )
+  if (required.length === 0) {
+    return table
+  }
+  for (const view of Object.values(table.views)) {
+    if (!sdk.views.isV2(view) || helpers.views.isCalculationView(view)) {
+      continue
+    }
```
samwho, line 36: This is the only meaningful change in here, but I did a little refactoring to reduce the nesting and make this more readable while I was here.

packages/server/src/sdk/app/views/index.ts
```diff
@@ -96,6 +96,13 @@ async function guardCalculationViewSchema(
       continue
     }
 
+    if (!schema.field) {
+      throw new HTTPError(
+        `Calculation field "${name}" is missing a "field" property`,
+        400
+      )
+    }
```
samwho, line 104: This a separate problem I ran into accidentally. It would error previously, but the message was unhelpful. This just improves the error message.

packages/server/src/sdk/app/views/index.ts
```diff
@@ -371,7 +378,8 @@ export function syncSchema(
 
   if (view.schema) {
     for (const fieldName of Object.keys(view.schema)) {
-      if (!schema[fieldName]) {
+      const viewSchema = view.schema[fieldName]
+      if (!helpers.views.isCalculationField(viewSchema) && !schema[fieldName]) {
```
samwho, line 382: This is the primary fix for the problem this PR addresses.

mike12345567 (approved):
LGTM - thanks for the comments explaining the changes!

adrinr (approved):

PR #14757, opened 2024-10-10
Validation for search api

adrinr:
## Description
Adding zod for validating the search request. This is used only for the view and table search input api. It's currently hide behind a feature flag in order to test it and release it properly


samwho (approved):
Love it.

PR #14758, opened 2024-10-10
Properly clean up row actions on table deletion.

samwho:
## Description

@ConorWebb96 found a bug where deleting a table that has row actions associated with it will cause several other endpoints to start returning errors, rendering the app unusable.

This happens because when a table is deleted, its row actions become orphaned and reference a table that no longer exists. Other things fetch these actions and then attempt to fetch the table, resulting in errors.

This PRs deletes all row actions associated with a table when the table is deleted.

### DIFF
--- a/packages/types/src/sdk/db.ts
+++ b/packages/types/src/sdk/db.ts
@@ -129,7 +129,12 @@ export interface Database {
   name: string
 
   exists(): Promise<boolean>
+  /**
+   * @deprecated the plan is to get everything using `tryGet` instead, then rename
+   * `tryGet` to `get`.
+   */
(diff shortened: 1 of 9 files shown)


packages/types/src/sdk/db.ts
```diff
@@ -129,7 +129,12 @@ export interface Database {
   name: string
 
   exists(): Promise<boolean>
+  /**
+   * @deprecated the plan is to get everything using `tryGet` instead, then rename
```
adrinr, line 133: ❤️

adrinr (commented):
LGTM!

adrinr (approved):
Now it LGTM!

PR #14762, opened 2024-10-10
Fix for an issue found with custom role naming

mike12345567:
## Description
Silly bug for if a role name contained `role_` - this would break the ID, correcting this and updating an existing test to cover this check.


samwho (approved):
lol

PR #14763, opened 2024-10-10
Allow sorting by calculation fields.

samwho:
## Description

@aptkingston pointed out to me that sorting by calculation fields throws an error currently, and would be really nice to have. This PR fixes it and adds tests.


mike12345567 (approved):
LGTM!

PR #14770, opened 2024-10-11
View joins with any and all not working correctly

adrinr:
## Description
Fix filtering issues with relationship filtering.

### DIFF
--- a/packages/server/src/api/routes/tests/search.spec.ts
+++ b/packages/server/src/api/routes/tests/search.spec.ts
@@ -2331,82 +2331,132 @@ describe.each([
         })
 
         describe("logical filters", () => {
-          it("should allow nested ands with single conditions", async () => {
-            await expectQuery({
-              $and: {
-                conditions: [
-                  {
-                    $and: {
-                      conditions: [
-                        {
-                          equal: { ["productCat.name"]: "foo" },
-                        },
-                      ],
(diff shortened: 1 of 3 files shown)


aptkingston:
I assume this would fix https://linear.app/budibase/issue/BUDI-8738/filter-by-relationship-not-working-on-views?

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -2329,6 +2330,211 @@ describe.each([
...
+                            },
+                          ],
+                        },
+                      },
+                    ],
+                  },
+                }).toContainExactly([
+                  { name: "foo", productCat: [{ _id: productCatRows[0]._id }] },
+                  { name: "bar", productCat: [{ _id: productCatRows[1]._id }] },
+                  // { name: "baz", productCat: undefined }, // TODO
```
adrinr, line 2509: This is the current behaviour and fixing this required quite a lot of changes. Doing this in a different branch while investigating the best approach

adrinr:
> I assume this would fix https://linear.app/budibase/issue/BUDI-8738/filter-by-relationship-not-working-on-views?

Yes, this should fix the same issue

samwho (approved):
The tests 👨‍🍳 💋

PR #14772, opened 2024-10-11
Allow BB_REFERNCE_SINGLE to have default values.

samwho:
## Description

Pretty much what it says on the tin.


aptkingston:
Could we update this to include multi user? I've added both in my frontend branch.

samwho:
I'll do multi in a follow-up PR.

aptkingston (approved):

PR #14774, opened 2024-10-11
Fix all or property

adrinr:
## Description
Fixing filter "behaviour". This has been broken and it always behaves as "all", even if "any" is selected

https://github.com/user-attachments/assets/8894f2fe-5559-45a0-9bcf-aa2ba61bcac3

## Launchcontrol
Fixing filter "behaviour"

### DIFF
--- a/packages/shared-core/src/utils.ts
+++ b/packages/shared-core/src/utils.ts
@@ -172,7 +172,7 @@ export const processSearchFilters = (
         .sort((a, b) => {
           return a.localeCompare(b)
         })
-        .filter(key => key in filter)
+        .filter(key => filter[key])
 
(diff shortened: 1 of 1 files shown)


packages/shared-core/src/utils.ts
```diff
@@ -172,7 +172,7 @@ export const processSearchFilters = (
         .sort((a, b) => {
           return a.localeCompare(b)
         })
-        .filter(key => key in filter)
```
adrinr, line 175: `key in filter` will be true if the key exists, even if the value is empty

mike12345567 (approved):

aptkingston (approved):

PR #14775, opened 2024-10-11
Fix searching by bindings when those bindings have been stored in the `view.query` field.

samwho:
## Description

This was a tricky one to track down and fix. What was happening is that the code that handles and-ing together a `view.query` with the query passed in from the search endpoint wasn't evaluating bindings correctly.

The story for binding evaluation in search queries is actually quite complicated, especially when they're user bindings. We have special code in place to ensure any user IDs passed in our the global variant (`us_...`) instead of the user metadata variant (`ro_ta_users_us...`). This is fundamentally what wasn't being correctly applied.

### DIFF
--- a/packages/server/src/sdk/app/rows/search.ts
+++ b/packages/server/src/sdk/app/rows/search.ts
@@ -81,6 +81,10 @@ export async function search(
       options.query = {}
     }
 
+    if (context) {
+      options.query = await enrichSearchContext(options.query, context)
+    }
+
     // need to make sure filters in correct shape before checking for view
     options = searchInputMapping(table, options)
 
@@ -100,6 +104,7 @@ export async function search(
       const sqsEnabled = await features.flags.isEnabled("SQS")
       const supportsLogicalOperators =
         isExternalTableID(view.tableId) || sqsEnabled
+
       if (!supportsLogicalOperators) {
         // In the unlikely event that a Grouped Filter is in a non-SQS environment
         // It needs to be ignored entirely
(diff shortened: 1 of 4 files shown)


packages/server/src/sdk/app/rows/search.ts
```diff
@@ -81,6 +81,10 @@ export async function search(
       options.query = {}
     }
 
+    if (context) {
+      options.query = await enrichSearchContext(options.query, context)
+    }
```
samwho, line 86: This has been moved from line 140 to here. This has to happen before `searchInputMapping` because `searchInputMapping` is what corrects metadata user IDs to global ones.

packages/server/src/sdk/app/rows/search.ts
```diff
@@ -92,12 +96,15 @@ export async function search(
       // Enrich saved query with ephemeral query params.
       // We prevent searching on any fields that are saved as part of the query, as
       // that could let users find rows they should not be allowed to access.
-      let viewQuery = dataFilters.buildQueryLegacy(view.query) || {}
-      delete viewQuery?.onEmptyFilter
+      let query = await enrichSearchContext(view.query || {}, context)
```
samwho, line 99: This `enrichSearchContext` was missing such that bindings were never being evaluated on `view.query` as far as I could tell.

packages/server/src/sdk/app/rows/search.ts
```diff
@@ -92,12 +96,15 @@ export async function search(
...
+      query = dataFilters.buildQueryLegacy(query) || {}
+      query = checkFilters(table, query)
```
samwho, line 101: This `checkFilters` was extracted from within `searchInputMapping` for the sole purpose of being used here to correct user IDs.

packages/server/src/sdk/app/rows/search.ts
adrinr, line 99: Can we rename this back to view query? Otherwise it's really confusing between the requested query and the view query

mike12345567 (approved):
LGTM!

PR #14777, opened 2024-10-11
Add default value toggle for both single and multi user columns

aptkingston:
## Description
This PR adds a default value setting for both single and multi user columns. The setting is just a toggle that will set the default value to the current user ID.

Single:
image

Multi:
image

Result:
image
image

### DIFF
--- a/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
+++ b/packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
@@ -840,8 +840,8 @@
     {:else if editableColumn.subtype === BBReferenceFieldSubType.USER}
       {@const defaultValue =
         editableColumn.type === FieldType.BB_REFERENCE_SINGLE
-          ? SINGLE_USER_DEFAULT
-          : MULTI_USER_DEFAULT}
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -49,19 +49,15 @@
...
   import { getUserBindings } from "dataBinding"
+  import { makePropSafe as safe } from "@budibase/string-templates"
 
-  const AUTO_TYPE = FieldType.AUTO
-  const FORMULA_TYPE = FieldType.FORMULA
-  const LINK_TYPE = FieldType.LINK
-  const STRING_TYPE = FieldType.STRING
-  const NUMBER_TYPE = FieldType.NUMBER
-  const JSON_TYPE = FieldType.JSON
-  const DATE_TYPE = FieldType.DATETIME
```
aptkingston, line 59: Just removing these aliases as they serve no purpose.

aptkingston:
The broken lint will be fixed by another open PR to master.

deanhannigan (approved):
LGTM!

PR #14778, opened 2024-10-11
Calculation Views UI

aptkingston:
## Description
This PR adds the UI for calculation views.

View type is specified upon creation. "Normal" views are called "Table" views.
image

Calculation views are configured using the "Configure calculations" button in the grid toolbar:
image

This opens the calculation editor when you can define which calculations you'd like to add to your view.
image

Example of calculation results:
image

Clarification of functionality:
- Calculation views are completely empty by default. You need to configure some calculations to see data.
- You cannot filter on calculation columns
- You *can* sort on calculation columns
- Calculation views do not support adding, editing or deleting rows
- Calculations can only be defined on numeric columns (number, bigint)
- There is a max of 5 calculations per calculation view
- There is no limit to the number of group bys


packages/builder/src/components/backend/DataTable/buttons/grid/GridViewCalculationButton.svelte
```diff
@@ -0,0 +1,259 @@
...
+        on:click={addCalc}
+        disabled={calculations.length >= 5}
+      >
+        Add calculation
+      </ActionButton>
+    </div>
+    <InfoDisplay
+      icon="Help"
+      quiet
+      body="Calculations only work with numeric columns and a maximum of 5 calculations can be added at once."
```
shogunpurple, line 237: Is this a hard technical limitation or soft limit?

shogunpurple (approved):
LGTM! Nice work

packages/builder/src/components/backend/DataTable/buttons/grid/GridViewCalculationButton.svelte
aptkingston: Soft limit as far as I know. This was just specified in the ticket, but I'm not sure if there's any real reason for it other than to limit the potential performance hit of running these.

aptkingston:
Test failures are random flakes. Going to merge this in and let us use it in QA.

PR #14783, opened 2024-10-13
Automation Branching UX

deanhannigan:
## Description
Adds the ability to create new branched automations in the builder.

- `FIX` - loop block binding added multiple times. Consecutive loops will continue to add `currentItem` bindings to the available bindings pool.
- `FIX` - error reporting in loop blocks would inadvertently overwrite the wrong results. If a loop failed it would update at the loop failure index and not the loop step index.
- `Update` - existing automations using step index runtime bindings (`steps.2.myBindings`), will be migrated automatically to use `step id` bindings. This change was made to better accomodate branching as a feature but it also means that we don't have to continually parse steps when you alter the step order. **No change is required by the user. Existing automations will continue to run without modification**

## Addresses
-  https://linear.app/budibase/project/automation-branching-0ddfe414a2aa/overview

## Screenshots
Incoming

## Launchcontrol
Adds support for creating branched automations in the builder!

## Feature branch env
[Feature Branch Link](http://fb-feature-automation-branching-ux.fb.qa.budibase.net)


aptkingston (commented):
Amazing work Dean! Such a massive feature and adds so much value. For sure one of the best features we've ever shipped 🚀 

Given the size of this, I'm just going to use it and note down anything I find.

## Bugs (blocking)
- I'm still seeing the bug where drag placeholders are not reactive, so they're wrong after dragging once
- The automation history results in settings is cut off, I assume due to the layout changes
image
- I can't actually get branching to work, but I'm not sure if that's more of a binding issue than a branching issue. I did a simple test with a create row trigger, and tried to match on a field being "a" or "b". I created a row with "a" and it fails saying no branch conditions matched. Here's the input and condition which both look correct, but it fails as shown:
image
image
image
- Not sure if related to your changes or not, but when using a create or update row trigger we don't have bindings for `_id` or `_rev`, and `_id` at least is very important to use in your automation later
- We should remove "branch" from the automation step modal, since it errors and we have a dedicated button anyway
image

## UX improvements (non blocking)
Most of these we've already talked about and most don't need to block merging as we can continue to improve the UX over time.
- My preference would be zoom without holding control to be consistent with RBAC, but we can get other opinions here
- We should have a min zoom level of 20% or something as you can currently make it invisible
- Pan while dragging near the edges of the screen would be nice. Acceleration not required, as zooming out when dragging via the scroll wheel is even easier
- The pan handler grab zone seems to only be the insides of the automation flow, so if you zoom out you can't pan by dragging outside the flow, which feels like its broken. Not sure how hard this is to add pan anywhere on the screen.
- Would be nice to center zooming around the mouse position
- Is it possible to update state immediately upon drag and drop? Right now when dropping, the drag state resets and then it waits for the API call to complete before updating the automation. Not the end of the world, but will be more of an issue in cloud where it's slower. Feels a bit jumpy. The same applies for moving branches left and right.
- I think the placeholder text for creating branching needs to be something different, as it's too similar to the "Add branch" button when you already have branching, which makes you think it'll do the same thing. Maybe something like "Create branching" or "Insert branching" to make it clear that it will start a new branching tree, rather than append a branch to an existing branching tree.
- Considering you can nest branching as many times as you want by adding it above the current tree, is there any reason we can't allow branching as the first child of a branch? We hide the icon right now, but I don't see any reason we can't allow it. This is where I mean:
image
And the use case would just be to dedupe conditions. So like doing 2 nested branching structures with 2 branches each containing 1 condition each, rather than one branching structure with 4 branches and needing 2 conditions in each.

deanhannigan:
Hey @aptkingston, thanks for taking a look! I ran through the bug list and all the issues you raised have been resolved. The only one I didn't address was the app action binding issue where `{{ trigger.row.First Name }}` should be `{{ trigger.row.[First Name] }}`. As we discussed, this is an existing issue with the trigger step.

aptkingston:
Excellent thanks Dean! I'm going to approve this to mark it as ready to test by the wider team, and we can keep working on the UX improvements.

aptkingston:
In terms of release plan, we'll not merge this into `v3-ui` but leave it on its own branch right up until release. We're going to update QA to deploy from this branch instead though, so we can use this in QA.

aptkingston (approved):

PR #14787, opened 2024-10-14
Changes to support processing bindings on both sides of branch condition evaluation

PClmnt:
## Description
Previously the processing of bindings was only supported on one side of the expression. Updating it so bindings get processed on both sides. Also added tests to support this.


deanhannigan (approved):
LGTM! 🙏

adrinr (approved):

samwho (approved):

PR #14788, opened 2024-10-14
Move conditionals to env section

calexiou:
## Description
Simplifies the conditionals that decide which license type text to use to dispatch to the remote workflow.

### DIFF
--- a/.github/workflows/deploy-featurebranch.yml
+++ b/.github/workflows/deploy-featurebranch.yml
@@ -2,13 +2,11 @@ name: deploy-featurebranch
 
 on:
   pull_request:
-    types: [
-      labeled,
-      # default types below (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)
-      opened,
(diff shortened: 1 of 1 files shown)


aptkingston (approved):
Happy to merge this in too, but I'm not sure if this will fix feature branch deployment. I think the issue is that in the BB deploys workflow the licensing step only runs when the license is set in the input context, but it doesn't check the`github.event.client_payload` context: `if: ${{ github.event.inputs.LICENSE_TYPE != 'free' }}`.

It was also that workflow where I meant that we could use the `env` context to simplify thing a lot, but happy to see it here as well of course!

calexiou:
You're right. We had these uncommited changes [`a5d41ba`](https://github.com/Budibase/budibase-deploys/commit/a5d41bac4b4b1606fdeff630e946e02767198ba8) and [`4677a71`](https://github.com/Budibase/budibase-deploys/commit/4677a710650951d52d3436aaece39362233c0d98) that should be addressing that

PR #14790, opened 2024-10-14
Fix view join parser for non-array values any "many" types

aptkingston:
## Description
Fixes a crash when viewing certain types of view join columns with non-array values.

## Addresses
- https://linear.app/budibase/issue/BUDI-8737/one-of-my-views-is-borked


deanhannigan (approved):
LGTM! 🥖

PR #14795, opened 2024-10-15
Fix negated join filters on relationships

adrinr:
## Description
Treating relationships on filter negations. These should behave as "blacklisting" rows, but they are not treating the rows with empty relationships accordingly. Currently these are always getting excluded

## Addresses
- https://linear.app/budibase/issue/BUDI-8752/fix-bugs-relationship-bugs

### Having the following data...
image

### Filter will behave as expected, returning empty relations when it should (blacklisting filters)
image
image

### But not returning them when it should not (whitelisting filters)
image
image

## Launchcontrol
Fixing certain filters on relationships


adrinr:
Related: https://linear.app/budibase/issue/BUDI-8752/fix-bugs-relationship-bugs

mike12345567 (approved):
Thanks for this, great work again, I didn't consider this scenario, thank you for rectifying it!

PR #14798, opened 2024-10-15
Fix multiple relationships same table

adrinr:
## Description
Fixing some recently introduced issue with multiple relationship between tables. Adding and fixing some tests to ensure this is supported and the search works as it should

## Addresses
- https://linear.app/budibase/issue/BUDI-8752/fix-bugs-relationship-bugs

### Given the following data, where both relationships point to the same table...
image

### Filtering by the first relation
image
image

### Filtering by the second relation
image
image

## Launchcontrol
Fixing some issues related multiple relationships on the same tables via different fields


adrinr:
Related: https://linear.app/budibase/issue/BUDI-8752/fix-bugs-relationship-bugs

mike12345567 (approved):
Thanks for this @adrinr - really great work here!

PR #14801, opened 2024-10-15
Add settings to automation context.

samwho:
## Description

Adds the following items to automation context:

- `settings.url`
- `settings.logo`
- `settings.company`

All taken from the global configuration.

## Launchcontrol

- Add `settings` to the context for automation bindings, allowing you to fetch the installation URL, company logo URL, and company name.


packages/server/src/threads/automation.ts
```diff
@@ -263,6 +263,13 @@ class Orchestrator {
         this.context.env = await sdkUtils.getEnvironmentVariables()
         this.context.user = this.currentUser
 
+        const { config } = await configs.getSettingsConfigDoc()
```
adrinr, line 266: Will this always exist? If so, should we make the fields not optional on the type?

samwho: I'm going by the types entirely, I'm not sure.

mike12345567: If the document exists the config should exist on it - I believe the type is correct as well, it is not optional on it. The get will fail and 404 the endpoint if the document doesn't exist, therefore the `{ config }` will always be safe.

samwho: I _think_ I've implemented what you're suggesting, PTAL 🙏

mike12345567 (approved):
LGTM - thanks for the fix!

PR #14813, opened 2024-10-17
Hide power role for v3 apps

adrinr:
## Description
Don't show POWER role for apps >= 3.x. This required adding a `createdVersion` entry in the appmetadata, and to be able to run/test fine locally, we are updating the way we infer the current running version on dev, getting the value from `lerna.json` (until now everything was set as 0.0.0)

## Launchcontrol
Do not show POWER role for apps >= 3.x


aptkingston (approved):
LGTM!

Just curious, what was the motivation for the new `creationVersion` concept rather than adding a new flag? Was it so that we don't have to keep adding new flags going forward, since we can just reference version for everything? It's a nice change 👌

packages/server/src/sdk/app/appMetadata/metadata.ts
```diff
@@ -0,0 +1,18 @@
+import { context, DocumentType } from "@budibase/backend-core"
+import { App } from "@budibase/types"
+
+/**
+ * @deprecated the plan is to get everything using `tryGet` instead, then rename
+ * `tryGet` to `get`.
+ */
+export async function get() {
```
mike12345567, line 8: There is an `application` layer to the SDK - would it not be better to add this there?

packages/backend-core/src/security/roles.ts
```diff
@@ -366,6 +383,18 @@ export async function getAllRoles(appId?: string): Promise<RoleDoc[]> {
   }
 }
 
+async function shouldIncludePowerRole(db: Database) {
+  const app = await db.tryGet<App>(DocumentType.APP_METADATA)
+  const creationVersion = app?.creationVersion
```
samwho, line 388: I'd be tempted to actually put this on the root `Document` type and call it `_creationVersion`, anticipating that we will likely introduce this idea onto other documents going forward.

adrinr:
> LGTM!
> 
> Just curious, what was the motivation for the new `creationVersion` concept rather than adding a new flag? Was it so that we don't have to keep adding new flags going forward, since we can just reference version for everything? It's a nice change 👌

Exactly. Having this new "creationVersion" will give us more flexibility. For example, it can be used in the future for certain migrations or similar, as we know when the app was created and what we don't need to support/migrate for that specific app

packages/worker/src/api/routes/global/tests/roles.spec.ts
```diff
@@ -69,6 +77,53 @@ describe("/api/global/roles", () => {
...
+        expect(res.body[appId].roles.map((r: any) => r._id)).toEqual([
+          ROLE_NAME,
+          roles.BUILTIN_ROLE_IDS.ADMIN,
+          roles.BUILTIN_ROLE_IDS.BASIC,
+          roles.BUILTIN_ROLE_IDS.PUBLIC,
+        ])
+      }
+    )
+
+    it.each(["2.9.0", "1.0.0"])(
```
samwho, line 96: I'd recommending including some of the other version strings we use, e.g. `0.0.0` (feature branches) and `2.32.17_2146.b125a7c` (qa)

packages/server/src/sdk/app/appMetadata/metadata.ts
adrinr: Not sure if that is exactly the same. I was actually doubting where to put that. I understood that the current application is for the crud of the apps itself, but this is for the app metadata specific. Do you think it's best to merge both?

packages/backend-core/src/security/roles.ts
samwho: Yeah, then in `DatabaseImpl.post` we know when a doc is new because it has no ID, we could slap `_creationVersion` onto it there. Though it _is_ a wide-reaching change that likely has risk.

packages/server/src/sdk/app/appMetadata/metadata.ts
mike12345567: I think it probably makes sense to have metadata functions under the application, the CRUD of the application does fall under that part of the SDK, but the metadata is pretty much the only application document there is and I would default to looking for it under that part of the SDK.

packages/backend-core/src/security/roles.ts
mike12345567: I'm unsure about this - I don't think we need this on every document, since every DB will have an `app_metadata` (it won't load without this/is corrupt) - in theory noting it here is enough since its a static document that is always easy to retrieve.

mike12345567 (approved):
LGTM - just giving pre-approval, I think it would be good to move the metadata SDK into the existing application, but that's just a preference, thing, I've went to that a couple times looking for it in the past!

PR #14816, opened 2024-10-17
Multi-inheritance RBAC (backend)

mike12345567:
## Description
As required for the new RBAC experience roles need a way to reference multiple other roles, meaning that more complex flows of roles can be created.

The main difference within this is the `inherits` type of a role can be `string | string[]` - the string format will apply to old roles and built-ins, while the array will be used for new roles to reference a set of roles below them.

With this, there is a high chance of role loops (this was possible previously as you could set two roles to inherit each other, however this would be un-unusable since it doesn't reference any built in (so essentially two public roles). The API has been updated to protect against loops, as well as providing a function in `shared-core` that can check for loops given a list of roles.

A general pass around prefixing has been performed, making sure that roles always take their external format when returned from the API as well as storing the roles with their complete prefixed format in the DB.

Finally, a few new test cases have been added:
1. Checks for the `accessible` endpoint which lists the roles a user has access to, making sure this works with multi-inheritance.
2. Checks for screen access, making sure that only the screens a user can access based on their `routing.roleId` are returned.
3. Checks for role loops in the role save API, i've also updated this test case to Typescript.
4. Checks for resource permissions, confirming that resources can be accessed based on role multi-inheritance.


packages/types/src/api/web/role.ts
```diff
@@ -1,12 +1,14 @@
 import { Role, RoleUIMetadata } from "../../documents"
+import { PermissionLevel } from "../../sdk"
 
 export interface SaveRoleRequest {
   _id?: string
   _rev?: string
   name: string
-  inherits: string
+  inherits?: string | string[]
```
adrinr, line 8: Is this meant to be optional?

packages/server/src/api/controllers/role.ts
```diff
@@ -53,11 +65,16 @@ async function updateRolesOnUserTable(
...
-  ctx.body = await roles.getAllRoles()
+  ctx.body = (await roles.getAllRoles()).map(role => externalRole(role))
 }
 
 export async function find(ctx: UserCtx<void, FindRoleResponse>) {
-  ctx.body = await roles.getRole(ctx.params.roleId)
+  const role = await roles.getRole(ctx.params.roleId)
+  if (!role) {
+    ctx.throw(404, { message: "Role not found" })
+  } else {
```
adrinr, line 75: NIT: redundant else

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -428,6 +428,34 @@ export default class TestConfiguration {
 
   // HEADERS
 
+  // sets the role for the headers, for the period of a callback
+  async setRole(roleId: string, cb: () => Promise<unknown>) {
```
adrinr, line 432: Is this the right naming? Is this "login in as role"?

packages/server/src/tests/utilities/api/role.ts
```diff
@@ -21,20 +22,28 @@ export class RoleAPI extends TestAPI {
   }
 
   save = async (body: SaveRoleRequest, expectations?: Expectations) => {
+    // the tests should always be creating the "new" version of roles
+    if (body.version === undefined) {
```
adrinr, line 26: Should this not be part of the interface, then?

packages/types/src/api/web/role.ts
mike12345567: This is correct, the `Role` has always allowed inherits to be empty - it could be worth refactoring this as I think inherits can always be set now, but its quite big and would probably do outside this RP.

packages/server/src/tests/utilities/api/role.ts
mike12345567: true - I can detect if the role doesn't have a `_rev` set it

packages/server/src/api/routes/tests/role.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+        level: PermissionLevel.READ,
+      })
+      const res = await config.api.roles.fetch()
+      expect(res.length).toBeGreaterThan(0)
+      const power = res.find(role => role._id === BUILTIN_ROLE_IDS.POWER)
+      expect(power?.permissions[table._id!]).toEqual(["read"])
+    })
+  })
+
+  describe("destroy", () => {
```
samwho, line 201: Could we have some tests of deleting roles from the middle of inheritance chains, to make sure the chains are correctly severed?

packages/server/src/middleware/currentapp.ts
```diff
@@ -59,11 +59,15 @@ export default async (ctx: UserCtx, next: any) => {
       // Ensure the role is valid by ensuring a definition exists
       try {
         if (roleHeader) {
-          await roles.getRole(roleHeader)
```
samwho, line 62: wat

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -428,6 +428,34 @@ export default class TestConfiguration {
 
   // HEADERS
 
+  // sets the role for the headers, for the period of a callback
+  async loginAsRole(roleId: string, cb: () => Promise<unknown>) {
+    const roleUser = await this.createUser({
+      roles: {
+        [this.prodAppId!]: roleId,
```
samwho, line 435: `prodAppId` is often undefined because I believe it's only set when `config.publish` is called. I'd call `getProdAppId()` so there's proper error checking, or `getAppId()` if it's not super important to be the prod one.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -428,6 +428,34 @@ export default class TestConfiguration {
...
+      admin: { global: false },
+    })
+    await this.login({
+      roleId,
+      userId: roleUser._id!,
+      builder: false,
+      prodApp: true,
+    })
+    const temp = this.user
+    this.user = roleUser
```
samwho, line 447: There's a `withUser` function you can use here to do this cleanup for you.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -459,6 +487,7 @@ export default class TestConfiguration {
 
     const headers: any = {
       Accept: "application/json",
+      Cookie: "",
```
samwho, line 490: How come this is necessary?

packages/server/src/tests/utilities/api/role.ts
```diff
@@ -27,14 +28,18 @@ export class RoleAPI extends TestAPI {
...
-    return await this._delete(`/api/roles/${roleId}`, {
+  destroy = async (role: Role, expectations?: Expectations) => {
+    return await this._delete(`/api/roles/${role._id}/${role._rev}`, {
       expectations,
     })
   }
 
-  accesssible = async (expectations?: Expectations) => {
+  accessible = async (
+    headers: Record<string, string | string[]>,
```
samwho, line 38: I'm not fond of this. Would prefer to see some sort of `config.withHeaders` helper, so it fits with other stuff that we do and doesn't pollute the API here for people that don't want to set headers.

packages/shared-core/src/helpers/roles.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // mark this role has been fully checked
+    checking.delete(roleId)
+    checked.add(roleId)
+
+    return false
+  }
+
+  return !!roles.find(role => hasLoop(role._id!))
+}
```
samwho, line 55: This could probably also be greatly simplified with a `walkRoleHierarchy` helper:

```typescript
export function checkForRoleInteritanceLoops(roles: Role[]): boolean {
  for (const role of roles) {
    let hasLoop = false
    walkRoleHierarchy(role, (child) => {
      if (role._id === child._id) {
        hasLoop = true
      }
    })
    if (hasLoop) {
      return true
    }
  }
  return false
}
```

Alternately if `walkRoleHierarchy` is written as a generator:

```typescript
export function checkForRoleInteritanceLoops(roles: Role[]): boolean {
  for (const role of roles) {
    for (const child of walkRoleHierarchy(role)) {
      if (role._id === child._id) {
        return true
      }
    }
  }
  return false
}
```

packages/shared-core/src/helpers/tests/roles.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  afterEach(() => {
+    allRoles = []
+  })
+
+  function check(hasLoop: boolean) {
+    const result = checkForRoleInheritanceLoops(allRoles)
+    expect(result).toBe(hasLoop)
+  }
+
+  describe("checkForRoleInheritanceLoops", () => {
```
samwho, line 46: Would be good to see if this also returns true for self-referential loops, e.g. role1 inherits role1.

packages/backend-core/src/security/roles.ts
```diff
@@ -170,12 +190,36 @@ export async function roleToNumber(id: string) {
...
+      const highestBuiltin: number | undefined = role.inherits
+        .map(roleId => {
+          const foundRole = hierarchy.find(role => role._id === roleId)
+          if (foundRole) {
+            return findNumber(foundRole) + 1
+          }
+        })
+        .filter(number => !!number)
+        .sort()
+        .pop()
```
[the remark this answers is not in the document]
mike12345567: Will investigate a general switch to `.toSorted()` in another PR, where can work out linter issues.

packages/backend-core/src/security/roles.ts
```diff
@@ -236,26 +301,67 @@ async function getAllUserRoles(
   userRoleId: string,
   opts?: { defaultPublic?: boolean }
 ): Promise<RoleDoc[]> {
+  const allRoles = await getAllRoles()
+  if (helpers.roles.checkForRoleInheritanceLoops(allRoles)) {
+    throw new Error("Loop detected in roles - cannot list roles")
```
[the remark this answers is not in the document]
mike12345567: Removed in latest

packages/backend-core/src/security/roles.ts
```diff
@@ -236,26 +301,67 @@ async function getAllUserRoles(
...
+  }
+  const foundRoleIds: string[] = []
+  return roles.filter(role => {
+    if (role._id && !foundRoleIds.includes(role._id)) {
+      foundRoleIds.push(role._id)
+      return true
+    } else {
+      return false
+    }
+  })
```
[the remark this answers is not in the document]
mike12345567: using lodash now.

packages/backend-core/src/security/roles.ts
```diff
@@ -236,26 +301,67 @@ async function getAllUserRoles(
...
-    currentRole = await getRole(currentRole.inherits)
-    if (currentRole) {
-      roles.push(currentRole)
+  const rolesFound = (ids: string | string[]) => {
+    if (Array.isArray(ids)) {
+      return ids.filter(id => roleIds.includes(id)).length === ids.length
+    } else {
+      return roleIds.includes(ids)
+    }
+  }
```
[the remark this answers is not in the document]
mike12345567: Has been completely removed.

packages/backend-core/src/security/roles.ts
```diff
@@ -236,26 +301,67 @@ async function getAllUserRoles(
...
+          iterateInherited(role)
+          break
+        } else {
+          roleIds.push(role.inherits)
+          role = findRole(role.inherits, allRoles, opts)
+          if (role) {
+            roles.push(role)
+          }
+        }
+      }
```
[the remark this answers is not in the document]
mike12345567: Has been added.

packages/server/src/tests/utilities/TestConfiguration.ts
mike12345567: This just makes sure the cookie header is definitely overridden to be nothing, if its being mixed with some other headers.

packages/shared-core/src/helpers/roles.ts
mike12345567: Discussed offline, the traversal uses the loop detection currently - may attempt to update this, but once the traversal is protected against loops it can't check for loops, chicken and egg problem.

samwho (approved):

PR #14818, opened 2024-10-17
Make `queryUI` on `ViewV2` something you can write on its own, and `query` gets filled in for you

samwho:
## Description

This PR is on request of @aptkingston, it makes it such that the various view APIs allow the frontend to only work with `queryUI` and can ignore `query`.

Here's broadly what's happening:

- On view create/update, if you set `queryUI` it will generate `query` for you (regardless what you specify for `query`, it will get overwritten because `queryUI` is the source of truth).
- On view fetch (either directly or through the table), if there is a `query` but no `queryUI`, `queryUI` will get generated from `query`*.

*This only happens if `query` is of type `LegacyFilter[]`, which is true of all `ViewV2` in prod on Budicloud today (we checked). When we convert `queryUI` to `query`, we don't convert to a `LegacyFilter[]` but instead to a `SearchFilters`. There's currently no function to convert between `SearchFilters` and `UISearchFilter` (renamed from `SearchFilterGroup`) because it's technically lossy, and in theory we don't have to because from now on the frontend will _always_ set `queryUI`.

I'm sorry for how confusing this all is. There's a lot of history around this part of the codebase and I'm doing the best I can to slot in new requirements to an already complex part of the code.

## Feature branch env
[Feature Branch Link](http://fb-queryui-default.fb.qa.budibase.net)


packages/server/src/sdk/app/tables/getters.ts
```diff
@@ -12,17 +12,37 @@ import {
...
+}
 
 export async function processTable(table: Table): Promise<Table> {
   if (!table) {
     return table
   }
 
   table = { ...table }
+  if (table.views) {
+    processViews(Object.values(table.views))
```
adrinr, line 44: Could this be a pure function and to return the processed views instead?

packages/shared-core/src/utils.ts
```diff
@@ -1,18 +1,27 @@
...
 import * as Constants from "./constants"
-import { removeKeyNumbering } from "./filters"
-
-// an array of keys from filter type to properties that are in the type
-// this can then be converted using .fromEntries to an object
-type AllowedFilters = [keyof LegacyFilter, LegacyFilter[keyof LegacyFilter]][]
+import { removeKeyNumbering, splitFiltersArray } from "./filters"
+import _ from "lodash"
+
+const FILTER_ALLOWED_KEYS = [
```
adrinr, line 15: I guess this is the subset of LegacyFilter, right? Could we type the `const _FILTER_ALLOWED_KEYS: (keyof LegacyFilter)[]`, to help removing magic strings?

packages/shared-core/src/filters.ts
```diff
@@ -308,308 +312,193 @@ export class ColumnSplitter {
...
+export function splitFiltersArray(filters: LegacyFilter[]) {
+  const split: LegacyFilterSplit = {
+    filters: [],
   }
 
-  if (!Array.isArray(filter)) {
-    return query
+  for (const filter of filters) {
+    if ("operator" in filter && filter.operator === "allOr") {
+      split.allOr = true
```
adrinr, line 438: Could this technically not be "false"?

packages/server/src/sdk/app/views/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { ViewV2 } from "@budibase/types"
+import { utils, dataFilters } from "@budibase/shared-core"
+import { isPlainObject } from "lodash"
+import { HTTPError } from "@budibase/backend-core"
+
+function isEmptyObject(obj: any) {
+  return obj && isPlainObject(obj) && Object.keys(obj).length === 0
```
adrinr, line 7: Are these checks ( `isPlainObject(obj)` && `Object.keys(obj).length === 0`) not redundant?

packages/types/src/api/web/searchFilter.ts
```diff
@@ -1,23 +1,57 @@
...
+    | "rangeLow"
+    | "rangeHigh"
+  // Field name will often have a numerical prefix when coming from the frontend,
+  // use the ColumnSplitter class to remove it.
   field: string
-  type?: FieldType
   value: any
+  type?: FieldType
   externalType?: string
+  noValue?: boolean
```
adrinr, line 32: What is this? I cannot see any usage of it

packages/server/src/sdk/app/tables/getters.ts
samwho: I didn't return anything because it modifies what's passed in. I tend to avoid copying unless it's necessary because lots of it puts pressure on the GC.

packages/types/src/api/web/searchFilter.ts
samwho: The frontend sends it so I thought it would be worth representing in the type. Happy to remove it if you think it just adds confusion, though.

packages/shared-core/src/filters.ts
samwho: No. The object the frontend sends is `{"operator": "allOr"}` and as far as I can tell it is always true. @aptkingston am I right on that?

packages/shared-core/src/utils.ts
samwho: Great call, will do!

samwho: This is actually quite interesting. It calls out `valueType` and `formulaType` as not being keys of `SearchFilter` (can't do `LegacyFilter` because that includes the allOr and onEmptyFilter objects, and makes the type functionally a `never[]`).

If I add them to the type, they won't be used anywhere in TS code but they will be in JS code.

packages/types/src/api/web/searchFilter.ts
samwho: Corollary: https://github.com/Budibase/budibase/pull/14818#discussion_r1810295475

packages/server/src/sdk/app/views/utils.ts
samwho: Not quite, the problem is arrays.

```typescript
> _.isPlainObject({})
true
> _.isPlainObject([])
false
> Object.keys({}).length === 0
true
> Object.keys([]).length === 0
true
```

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -1602,7 +1602,7 @@ describe.each([
         })
       })
 
-    describe.each([FieldType.ARRAY, FieldType.OPTIONS])("%s", () => {
+    describe("arrays", () => {
```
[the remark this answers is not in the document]
samwho: So I just tried this and it has actually always been incorrect. `FieldType.OPTIONS` only takes single values, so these tests don't work for it anyway.

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -864,6 +958,21 @@ describe.each([
...
               { operator: "equal", field: "newField", value: "thatValue" },
             ],
+            // Should also update queryUI because query was not previously set.
+            queryUI: {
+              groups: [
+                {
+                  logicalOperator: "all",
+                  filters: [
+                    {
+                      operator: "equal",
```
mike12345567, line 968: Sometimes the enum `BasicOperator` is used and sometimes its a string here, is that intended?

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -170,17 +247,33 @@ describe.each([
...
           schema: {
             id: { visible: true },
             Price: {
               visible: true,
             },
           },
+          queryUI: {
+            groups: [
+              {
+                logicalOperator: UILogicalOperator.ALL,
```
mike12345567, line 261: There's a part of me wonder if this might be a bit more visually obvious if we had something in the structure utilities to create a default query UI, I assume that is what all these additions are for, to create the kind of basic view that the frontend will send up.

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -2851,398 +3013,534 @@ describe.each([
...
+                      value: "foo2",
+                    },
+                  ],
+                },
+              ],
+            },
             schema: {
               id: { visible: true },
-              one: { visible: false },
+              one: { visible: true },
```
mike12345567, line 3042: What is the change that lead to this? From what I can see the old test case specified some fields visible and some not, while this no longer does that, is there another test case that replicate the behaviour this was testing now?

packages/server/src/sdk/app/views/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function ensureQueryUISet(view: ViewV2) {
+  if (!view.queryUI && view.query && !isEmptyObject(view.query)) {
```
mike12345567, line 11: I understand this affects the input therefore it isn't required to return and maybe this is a personal preference, but even if the input is changed I still like when a function returns the altered state - therefore you know the function has a goal of changing/updating something, the returned response will be it.

It would also be possible to do a:
```
return {
  ...view,
  queryUI: utils.processSearchFilters(view.query),
}
```
To get a response which does not affect the input; I understand this is less performant and more GC time however. Having it just `return view` even if the input has changed is helpful to the caller, since it makes it a bit more obvious what it did.

packages/server/src/sdk/app/views/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+    view.queryUI = utils.processSearchFilters(view.query)
+  }
+}
+
+export function ensureQuerySet(view: ViewV2) {
+  // We consider queryUI to be the source of truth, so we don't check for the
+  // presence of query here. We will overwrite it regardless of whether it is
+  // present or not.
+  if (view.queryUI && !isEmptyObject(view.queryUI)) {
+    view.query = dataFilters.buildQuery(view.queryUI)
```
mike12345567, line 37: Same as comment with previous function.

packages/server/src/sdk/app/views/utils.ts
samwho: I'd argue the opposite, it is confusing to the caller when a function both modifies its input and returns it, and we've had this bite us recently with `array.sort()`.

I think functions should do one or the other, and if there's no clear need to copy I think mutating makes the most sense.

mike12345567: In this case I think I would probably prefer the copy, as the function is called quite a bit and its not very obvious if its just error checking. I do also like having the response since it makes it properly chainable e.g. `checkOtherViewParameters(ensureQueryUISet(...))` - although I guess this is still achievable with in/out parameters.

Middle-ground could be a name which explains it mutates the input, just so its obvious what it will do.

samwho: I feel like this:

```js
ensureQueryUISet(view)
checkOtherViewParameters(view)
```

Is clearer than:

```js
checkOtherViewParameters(ensureQueryUISet(view))
```

But if your preference is copying I'll change it.

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho: I do have a helper further down in the file called `simpleQuery`. For now I've removed the optional fields (they weren't optional when I started) to reduce the verbosity.

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho: Weird, it hasn't changed as far as I can tell:
https://github.com/Budibase/budibase/blob/44fbe7074050a3420025638f0542fcd267bd5d78/packages/server/src/api/routes/tests/viewV2.spec.ts#L2687-L2691

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho: No 🙈

mike12345567 (approved):
LGTM - very nice work, great to get all the tests updated as well!

adrinr (approved):

packages/shared-core/src/filters.ts
aptkingston: Correct. If used, it will be exactly `{ operator: "allOr" }`, indicating we want to "or" all the filters. If we don't want the `allOr` functionality then this won't be present.

PR #14825, opened 2024-10-18
Support user in automation context

PClmnt:
## Description
This passes through the user from row based triggers as well as the app trigger into the automation context. This will allow the binding of user values within automations. Frontend will need to go in on the v3 branch. 

## Addresses
https://linear.app/budibase/issue/BUDI-8721/automations-current-user-binding-not-available


packages/server/src/automations/triggers.ts
```diff
@@ -140,9 +141,15 @@ function rowPassesFilters(row: Row, filters: SearchFilters) {
...
-  params: { fields: Record<string, any>; timeout?: number; appId?: string },
+  params: {
+    fields: Record<string, any>
+    timeout?: number
+    appId?: string
+    user?: User
+  },
   { getResponses }: { getResponses?: boolean } = {}
 ): Promise<any> {
+  console.log("user: " + params.user)
```
samwho, line 152: 👀

packages/server/src/automations/triggers.ts
```diff
@@ -189,6 +196,7 @@ export async function externalTrigger(
       appId: context.getAppId(),
       automation,
     }
+    console.log(data)
```
samwho, line 199: 👀

packages/server/src/api/controllers/automation.ts
```diff
@@ -159,6 +160,7 @@ export async function trigger(ctx: UserCtx) {
         automation,
         {
           fields: ctx.request.body.fields,
+          user: ctx.user as User,
```
samwho, line 163: This is quite tragic isn't it. I had a look at why you needed to do this and `UserCtx` has a `ContextUser` (lol) as the `user` prop, and it's just an `Omit<User, "roles">`. Should `externalTrigger()` be updated to accept a `ContextUser` instead of `User`?

PClmnt: Then we have to update the `User` all the way day into `AutomationDataEvent` to `ContextUser`. Which I guess makes sense, because it is a **_user that comes from context_**. But slightly disgusting all the same lol.

packages/server/src/api/controllers/row/index.ts
```diff
@@ -168,7 +183,15 @@ async function deleteRows(ctx: UserCtx<DeleteRowRequest>) {
   }
 
   for (let row of rows) {
-    ctx.eventEmitter && ctx.eventEmitter.emitRow(`row:delete`, appId, row)
+    ctx.eventEmitter &&
+      ctx.eventEmitter.emitRow(
+        `row:delete`,
```
mike12345567, line 188: would probably be wroth changing this into an object specifier like:
```
{ appId: string, row: Row, table?: Table, oldRow?: Row, user: User }
```

This way user can be required and some elements can be optional, without needing to specify `null`.

packages/server/src/automations/triggers.ts
```diff
@@ -140,7 +141,12 @@ function rowPassesFilters(row: Row, filters: SearchFilters) {
 
 export async function externalTrigger(
   automation: Automation,
-  params: { fields: Record<string, any>; timeout?: number; appId?: string },
+  params: {
+    fields: Record<string, any>
+    timeout?: number
+    appId?: string
+    user?: UserBindings | undefined
```
mike12345567, line 148: Why does `undefined` need specified here as well as the optional `user?:` - only one should be needed.

packages/types/src/documents/global/user.ts
```diff
@@ -68,6 +68,16 @@ export interface User extends Document {
   appSort?: string
 }
 
+export interface UserBindings extends Document {
+  firstName?: string
+  lastName?: string
+  email?: string
+  status?: string
+  roleId?: string | undefined | null
```
mike12345567, line 76: Ideally this should be `roleId?: string | null` - should cover the same typing, although I expect `null` isn't required.

mike12345567 (approved):
LGTM! Good to get this in, left a few comments about typing improvements/clarifications, but giving a pre-approval as none of these changes are blocking.

packages/server/src/automations/triggers.ts
PClmnt: Yea, I can just remove the undefined. It was just a leftover from a previous type I had. No longer needed

adrinr (approved):

packages/server/src/api/controllers/row/index.ts
PClmnt: Done

packages/server/src/api/routes/tests/rowAction.spec.ts
```diff
@@ -783,6 +783,7 @@ describe("/rowsActions", () => {
                 ...(await config.api.table.get(tableId)),
                 views: expect.anything(),
               },
+              user: expect.anything(),
```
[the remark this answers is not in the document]
PClmnt: Done :)

PR #14837, opened 2024-10-21
deprecates Section component

mikesealey:
## Description
Deprecates `Section` component. 

Since Grid layout, there isn't a lot of need for a section component. There were a couple of known-bugs that came up with it, but there's little use in fixing them when the grid layout achieves the same end-result (and more).

Deletes "Section" from componentStructure.json
moves component into deprecated folder
Updates filepath to deprecated folder


aptkingston (approved):
LGTM!

PR #14839, opened 2024-10-21
Default user columns to current user for templates

adrinr:
## Description
With the new templates containing user columns, we need a way to set valid values for it. Otherwise, the data will be trimmed on the way out, causing some unexpected behaviours on the template.

This PR ensures that when a new app comes from a template, the user column data will be defaulted to the creator of the app.

Also, updating multiple interface and typings that were not correct and misleading 

### Using the following data as a template
image

### Before, using that template, the user column was empty
image

### After, the data gets populated with the user creator
image

### DIFF
--- a/packages/server/src/api/controllers/application.ts
+++ b/packages/server/src/api/controllers/application.ts
@@ -250,9 +250,9 @@ async function performAppCreate(ctx: UserCtx<CreateAppRequest, App>) {
     useTemplate,
     key: templateKey,
   }
-  if (ctx.request.files && ctx.request.files.templateFile) {
+  if (ctx.request.files && ctx.request.files.fileToImport) {
     instanceConfig.file = {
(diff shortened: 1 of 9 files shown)


samwho (approved):

packages/server/src/api/controllers/application.ts
```diff
@@ -279,6 +276,10 @@ async function performAppCreate(ctx: UserCtx<CreateAppRequest, App>) {
     const instance = await createInstance(appId, instanceConfig)
     const db = context.getAppDB()
 
+    if (instanceConfig.useTemplate && !instanceConfig.file) {
+      await updateUserColumns(appId, db, ctx.user._id!)
```
adrinr, line 280: This is the actual work that does the required job, the rest is just typing and cleaning

samwho (approved):

PR #14843, opened 2024-10-22
get user doc by email for account holder login

melohagan:
## Description
Testing out login without `budibaseUserId`

Although  believe the type change to include `budibaseAccess` boolean is valid regardless

## Addresses
- https://linear.app/budibase/issue/GRO-721/spike-can-account-holder-login-via-email-no-budibase-user-id


melohagan:
spike complete

PR #14844, opened 2024-10-22
Preview role - multi-inheritance fixes

mike12345567:
## Description
There were a few issues detected by @aptkingston and @deanhannigan with how the backend handled multi-inheritance changes  and the preview role. This is selected when previewing the design of an app, it can set `x-budibase-role` to force the app to appear as a certain role, rather than just as a builder.

There were a couple of areas that needed fixed:
1. Custom roles were being created without any inheritance specified, this assumes they are public roles. Now the API makes sure roles have at least the `BASIC` built in role to require custom roles to be logged in (public roles don't make any sense, they cannot be used).
2. The `currentapp.ts` had an issue as it was checking the existence of a role specified in the role header, however this was a double check as it does this later in the function when it has a complete app context, cleaned out the old un-used bit and added a test case to make sure this works.
3. `accessible` API didn't handle when roles had no inheritance, in this case it just listed all roles. Fixed this in two ways, the API doesn't let creating roles like this anymore and a test case which makes sure in these cases the API just returns that role and nothing else.

### DIFF
--- a/packages/backend-core/src/security/roles.ts
+++ b/packages/backend-core/src/security/roles.ts
@@ -220,7 +220,7 @@ export function validInherits(
   if (!inherits) {
     return false
   }
-  const find = (id: string) => allRoles.find(r => compareRoleIds(r._id!, id))
+  const find = (id: string) => allRoles.find(r => roleIDsAreEqual(r._id!, id))
   if (Array.isArray(inherits)) {
     const filtered = inherits.filter(roleId => find(roleId))
     return inherits.length !== 0 && filtered.length === inherits.length
@@ -306,7 +306,7 @@ export function lowerBuiltinRoleID(roleId1?: string, roleId2?: string): string {
     : roleId1
 }
 
-export function compareRoleIds(roleId1: string, roleId2: string) {
+export function roleIDsAreEqual(roleId1: string, roleId2: string) {
   // make sure both role IDs are prefixed correctly
   return prefixRoleID(roleId1) === prefixRoleID(roleId2)
 }
@@ -339,7 +339,7 @@ export function findRole(
     roleId = prefixRoleID(roleId)
   }
   const dbRole = roles.find(
-    role => role._id && compareRoleIds(role._id, roleId)
+    role => role._id && roleIDsAreEqual(role._id, roleId)
   )
   if (!dbRole && !isBuiltin(roleId) && opts?.defaultPublic) {
     return cloneDeep(BUILTIN_ROLES.PUBLIC)
@@ -573,7 +573,7 @@ export class AccessController {
     }
 
     return (
-      roleIds?.find(roleId => compareRoleIds(roleId, tryingRoleId)) !==
+      roleIds?.find(roleId => roleIDsAreEqual(roleId, tryingRoleId)) !==
       undefined
     )
   }
--- a/packages/backend-core/src/security/roles.ts
+++ b/packages/backend-core/src/security/roles.ts
@@ -213,6 +213,22 @@ export function getBuiltinRole(roleId: string): Role | undefined {
   return cloneDeep(role)
 }
 
+export function validInherits(
+  allRoles: RoleDoc[],
+  inherits?: string | string[]
+): boolean {
+  if (!inherits) {
+    return false
+  }
+  const find = (id: string) => allRoles.find(r => compareRoleIds(r._id!, id))
+  if (Array.isArray(inherits)) {
+    const filtered = inherits.filter(roleId => find(roleId))
+    return inherits.length !== 0 && filtered.length === inherits.length
+  } else {
+    return !!find(inherits)
+  }
+}
+
 /**
  * Works through the inheritance ranks to see how far up the builtin stack this ID is.
  */
--- a/packages/server/src/api/controllers/role.ts
+++ b/packages/server/src/api/controllers/role.ts
@@ -23,6 +23,7 @@ import {
 import { RoleColor, sdk as sharedSdk, helpers } from "@budibase/shared-core"
 import sdk from "../../sdk"
 import { builderSocket } from "../../websockets"
+import { roleIDsAreEqual } from "@budibase/backend-core/src/security/roles"
 
 const UpdateRolesOptions = {
   CREATED: "created",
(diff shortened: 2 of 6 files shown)


packages/backend-core/src/security/roles.ts
```diff
@@ -213,6 +213,22 @@ export function getBuiltinRole(roleId: string): Role | undefined {
...
 }
 
+export function validInherits(
+  allRoles: RoleDoc[],
+  inherits?: string | string[]
+): boolean {
+  if (!inherits) {
+    return false
+  }
+  const find = (id: string) => allRoles.find(r => compareRoleIds(r._id!, id))
```
samwho, line 223: `compareRoleIds` naming makes this a bit tricky to understand what's going on at first glance. Would suggest a rename to `roleIdsAreEqual(r._id!, id)`

packages/server/src/api/controllers/role.ts
```diff
@@ -125,6 +126,17 @@ export async function save(ctx: UserCtx<SaveRoleRequest, SaveRoleResponse>) {
     ctx.throw(400, "Cannot change custom role name")
   }
 
+  // custom roles should always inherit basic - if they don't inherit anything else
+  if (!inherits && roles.validInherits(allRoles, dbRole?.inherits)) {
+    inherits = dbRole?.inherits
+  } else if (!roles.validInherits(allRoles, inherits)) {
+    inherits = [roles.BUILTIN_ROLE_IDS.BASIC]
+  }
+  // assume write permission level for newly created roles
```
samwho, line 135: Why?

packages/server/src/api/controllers/role.ts
```diff
@@ -125,6 +126,17 @@ export async function save(ctx: UserCtx<SaveRoleRequest, SaveRoleResponse>) {
     ctx.throw(400, "Cannot change custom role name")
   }
 
+  // custom roles should always inherit basic - if they don't inherit anything else
+  if (!inherits && roles.validInherits(allRoles, dbRole?.inherits)) {
+    inherits = dbRole?.inherits
+  } else if (!roles.validInherits(allRoles, inherits)) {
+    inherits = [roles.BUILTIN_ROLE_IDS.BASIC]
```
samwho, line 133: This feels like it would result in a confusing API experience. Specifying invalid inherits should _probably_ be an error, no?

packages/server/src/api/controllers/role.ts
```diff
@@ -226,35 +238,20 @@ export async function accessible(ctx: UserCtx<void, AccessibleRolesResponse>) {
   if (!roleId) {
     roleId = roles.BUILTIN_ROLE_IDS.PUBLIC
   }
+  // If a custom role is provided in the header, filter out higher level roles
+  const roleHeader = ctx.header?.[Header.PREVIEW_ROLE] as string
+  const isBuilder = ctx.user && sharedSdk.users.isAdminOrBuilder(ctx.user)
```
samwho, line 243: When would `ctx.user` ever not be present? I thought `UserCtx` guaranteed it would be, throwing an auth error otherwise?

packages/server/src/api/controllers/role.ts
```diff
@@ -226,35 +238,20 @@ export async function accessible(ctx: UserCtx<void, AccessibleRolesResponse>) {
   if (!roleId) {
     roleId = roles.BUILTIN_ROLE_IDS.PUBLIC
   }
+  // If a custom role is provided in the header, filter out higher level roles
+  const roleHeader = ctx.header?.[Header.PREVIEW_ROLE] as string
```
adrinr, line 242: Either this type should allow undefined or we should handle the `header?`

packages/server/src/api/routes/tests/role.spec.ts
```diff
@@ -149,6 +159,19 @@ describe("/roles", () => {
...
+      const res = await config.api.roles.save({
+        ...basicRole(),
+        inherits: [BUILTIN_ROLE_IDS.ADMIN],
+      })
+      // save again
+      const updatedRes = await config.api.roles.save({
+        ...res,
+        inherits: undefined,
+      })
+      expect(updatedRes.inherits).toEqual([BUILTIN_ROLE_IDS.ADMIN])
```
samwho, line 173: I'm not sure I understand why we expect this behaviour. I don't think many other things in our API work this way. How come you can't erase/reset the `inherits` field to `undefined` or some default value?

packages/server/src/api/routes/tests/role.spec.ts
samwho, line 1: Would be nice to have a testcase for an `inherits` full of totally bogus values.

packages/backend-core/src/security/roles.ts
```diff
@@ -213,6 +213,22 @@ export function getBuiltinRole(roleId: string): Role | undefined {
   return cloneDeep(role)
 }
 
+export function validInherits(
+  allRoles: RoleDoc[],
+  inherits?: string | string[]
+): boolean {
+  if (!inherits) {
```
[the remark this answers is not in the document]
mike12345567: Yeah I originally had it that it didn't accept `undefined`, however that means you need to do `!role.inherits 
 || validInherits(role.inherits)` to check if its valid everywhere, I like being able to just pass the prop from the document and check if its valid in one go.

packages/backend-core/src/security/roles.ts
```diff
@@ -213,6 +213,22 @@ export function getBuiltinRole(roleId: string): Role | undefined {
...
+    return false
+  }
+  const find = (id: string) => allRoles.find(r => compareRoleIds(r._id!, id))
+  if (Array.isArray(inherits)) {
+    const filtered = inherits.filter(roleId => find(roleId))
+    return inherits.length !== 0 && filtered.length === inherits.length
```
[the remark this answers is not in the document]
mike12345567: In this case it is fine, if a role is duplicated its still valid - it will still produce an array of same length.

packages/server/src/api/controllers/role.ts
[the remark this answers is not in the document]
mike12345567: The frontend will respect this, however for API users it was decided we can make sure roles are always valid, we could instead throw an error here and tell the user to set `BASIC` if they have set nothing - however in a lot of APIs we tend to fill in missing components when not found.

packages/server/src/api/routes/tests/role.spec.ts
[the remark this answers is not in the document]
mike12345567: it takes the role from the DB user if it isn't specified - this was the previous behaviour but it was not tested for.

packages/server/src/api/controllers/role.ts
mike12345567: Discussed on Adri's comment.

packages/server/src/api/controllers/role.ts
mike12345567: discussed on Adri's comment.

packages/server/src/api/routes/tests/role.spec.ts
mike12345567: Discussed in Adri's comment, this is old behaviour being tested for now.

packages/server/src/api/controllers/role.ts
mike12345567: Addressed.

samwho (approved):

PR #14845, opened 2024-10-22
Allow serving old apps on local dev

adrinr:
## Description
For a while, when running locally we always used 0.0.0. This meant that we could not run the upgrade/downgrade system locally, and this was changed recently to use the latest version coming from the local `lerna.json` value.
This is now causing serving the client locally, as the server expects 0.0.0 to return the local client.

These changes are adding a `+local` flag to the prerelease locally, and we are serving the build one if this flag is found. We also return the local build if the version is 0.0.0 for backwards compatibility.
These checks also allow us to import production apps (or manually removing the `+local` tag) and to run with older versions locally.

### DIFF
--- a/packages/server/src/api/controllers/static/index.ts
+++ b/packages/server/src/api/controllers/static/index.ts
@@ -268,8 +268,8 @@ export const serveClientLibrary = async function (ctx: Ctx) {
     ctx.throw(400, "No app ID provided - cannot fetch client library.")
   }
 
-  const serverLocally = shouldServeLocally(version || "")
-  if (!serverLocally) {
+  const serveLocally = shouldServeLocally(version || "")
+  if (!serveLocally) {
     ctx.body = await objectStore.getReadStream(
       ObjectStoreBuckets.APPS,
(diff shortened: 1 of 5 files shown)


packages/server/src/api/controllers/static/index.ts
```diff
@@ -257,25 +258,29 @@ export const serveBuilderPreview = async function (ctx: Ctx) {
...
+  }
+
   const appId = context.getAppId() || (ctx.request.query.appId as string)
   let rootPath = join(NODE_MODULES_PATH, "@budibase", "client", "dist")
   if (!appId) {
     ctx.throw(400, "No app ID provided - cannot fetch client library.")
   }
-  if (env.isProd() || (env.isDev() && version !== devClientVersion)) {
+
+  const serverLocally = shouldServeLocally(version || "")
```
mike12345567, line 271: `serveLocally` rather than `serverLocally`

mike12345567 (approved):
LGTM!

aptkingston (approved):
Thanks for the quick turnaround @adrinr!

PR #14849, opened 2024-10-23
Row action name cleanup

adrinr:
## Description
On the initial design of row actions, the crud of these (specially editions and deletions) were done in the data section. With the latest designs, these are managed in the automation section, so there is quite a lot of code that can be simplified to achieve this.
Main changes:
- Remove the persisted row action name, as it should use the automation one. This is only used in the builder, so we don't need to worry for the extra db check.
- Remove the row action update endpoint, as this is just renaming the actual action
- Simplify some API responses
- Remove the update row action frontend modal, as it's doing the same than the automation modal

## Addresses
- [BUDI-8770 - Row Action automation names are not updated in Automation Logs when names are changed
](https://linear.app/budibase/issue/BUDI-8770/row-action-automation-names-are-not-updated-in-automation-logs-when)


packages/server/src/api/controllers/rowAction/crud.ts
```diff
@@ -68,26 +69,6 @@ export async function create(
   ctx.status = 201
 }
 
-export async function update(
```
adrinr, line 71: This logic has not been moved to the automation screen, so this is not needed anymore

packages/server/src/api/controllers/rowAction/crud.ts
```diff
@@ -96,38 +77,36 @@ export async function remove(ctx: Ctx<void, void>) {
...
+export async function setTablePermission(
+  ctx: Ctx<void, RowActionPermissionsResponse>
+) {
   const table = await getTable(ctx)
   const tableId = table._id!
   const { actionId } = ctx.params
 
   const action = await sdk.rowActions.setTablePermission(tableId, actionId)
   ctx.body = {
-    tableId,
```
adrinr, line 106: This data is not actually required, and it would require fetching the automation names. Instead of adding the extra call, we can just simplify the responses

aptkingston (approved):
LGTM! Always happy to see simplification 👍

PR #14853, opened 2024-10-23
Role validation - allow permissionId to be optional

mike12345567:
## Description
Frontend now expects to be able to supply role without a `permissionId`, this update makes it so that the validators allow this.

Added test cases for the behaviour - also realised the types in this area were pretty incorrect based on the validators, `permissionId` was just a `string` when it was actually typed, fixed this.

### DIFF
--- a/packages/backend-core/src/security/roles.ts
+++ b/packages/backend-core/src/security/roles.ts
@@ -1,5 +1,4 @@
 import semver from "semver"
-import { BuiltinPermissionID, PermissionLevel } from "./permissions"
 import {
   prefixRoleID,
   getRoleParams,
@@ -14,6 +13,8 @@ import {
   RoleUIMetadata,
   Database,
   App,
+  BuiltinPermissionID,
+  PermissionLevel,
 } from "@budibase/types"
 import cloneDeep from "lodash/fp/cloneDeep"
 import { RoleColor, helpers } from "@budibase/shared-core"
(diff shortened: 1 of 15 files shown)


packages/server/src/api/routes/utils/validators.ts
```diff
@@ -214,8 +215,8 @@ export function roleValidator() {
       }).optional(),
       // this is the base permission ID (for now a built in)
       permissionId: Joi.string()
-        .valid(...Object.values(permissions.BuiltinPermissionID))
-        .required(),
+        .valid(...Object.values(BuiltinPermissionID))
```
mike12345567, line 218: One of two code changes in this PR.

samwho (approved):

packages/backend-core/src/security/roles.ts
```diff
@@ -50,7 +51,7 @@ export class Role implements RoleDoc {
   _id: string
   _rev?: string
   name: string
-  permissionId: string
+  permissionId: BuiltinPermissionID
```
adrinr, line 54: Will this not be used for custom roles, as well?

mike12345567: Yes, this is used in all roles, but its never a `string` - its just a very old type, we've always validated the Role to make sure it is one of the values in `BuiltinPermissionID`.

PR #14858, opened 2024-10-24
Remove lightest and dark themes and add core types and utils for themes

aptkingston:
## Description
This PR removes the lightest and dark theme options.

I've also cleaned up and improved our management of themes in the codebase, adding a proper enum and type for metadata, moving utils into shared-core and using them across the board.

We previously stored themes different in the builder and client app (sometimes prefixed with `spectrum--` and sometimes not) and the code was littered with magic strings. We now also have proper constants for the available and default themes for apps and the builder.

In terms of backwards compatibility:
- Anyone currently using a removed theme for the builder will have it migrated to the closest available option (lightest > light, dark > darkest)
- For apps, your theme will also be migrated once opening the builder, but you won't need to update the client library. I've had to add a new client manifest feature flag to achieve this, but the result should be that whether or not you update your client library you'll be able to change your theme and see the correct (migrated) theme both in the SSR'd loading skeleton and the builder preview.


samwho (approved):
Just sign-off on the few changes in `packages/server`, haven't reviewed the majority of stuff but don't want you getting stuck on CODEOWNER approval. 🙂

aptkingston:
> Just sign-off on the few changes in `packages/server`, haven't reviewed the majority of stuff but don't want you getting stuck on CODEOWNER approval. 🙂

Cheers Sam. Added a few basic sanity tests for the shared-core utils there too.

shogunpurple (approved):
LGTM

Just wondering if there's actually any need to have logic to perform the fallbacks etc. Can we just remove the option to select it as a theme, and keep it around behind the scenes for apps/builders that currently use it?

aptkingston:
> LGTM
> 
> Just wondering if there's actually any need to have logic to perform the fallbacks etc. Can we just remove the option to select it as a theme, and keep it around behind the scenes for apps/builders that currently use it?

I would have, but it presents a few problems:
- Need some UI to reference that an non existing option is selected
- Would have to still handle the different value types stored for builder theme and app theme (`lightest` and `spectrum--lightest` for builder and app respectively), which have now been unified

I've tried to alleviate the pain of that by moving to the closest match though, so lightest will become light, and dark will become darkest. So it's not as if there should be huge visual differences at least.

If you think it's too much to force migrate them at all though then we can shelve this for now. Doesn't need to go in.

PR #14859, opened 2024-10-24
Delete row actions when removing a datasource

adrinr:
## Description
When deleting a table, we are currently deleting their related row actions and their automations. This PR ensures that the same happens when deleting a datasource, not leaving orphan row actions

### DIFF
--- a/packages/server/src/tests/utilities/api/automation.ts
+++ b/packages/server/src/tests/utilities/api/automation.ts
@@ -23,17 +23,6 @@ export class AutomationAPI extends TestAPI {
     })
   }
 
-  fetchEnriched = async (
-    expectations?: Expectations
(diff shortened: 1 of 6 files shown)


packages/server/src/tests/utilities/api/automation.ts
```diff
@@ -23,17 +23,6 @@ export class AutomationAPI extends TestAPI {
     })
   }
 
-  fetchEnriched = async (
```
adrinr, line 26: This is not used, after the recent cleanups. `enrich` is not needed anymore

samwho (approved):
Good catch!

shogunpurple (approved):
LGTM - great tests 🧪

PR #14860, opened 2024-10-24
Remove row actions when unfetching tables

adrinr:
## Description
Similar to https://github.com/Budibase/budibase/pull/14859, we might have orphan row actions when some tables are "unfetched". This PR ensures that row actions are removed when a table has been indirectly removed.

image

### DIFF
--- a/packages/server/src/sdk/app/datasources/plus.ts
+++ b/packages/server/src/sdk/app/datasources/plus.ts
@@ -9,6 +9,7 @@ import * as datasources from "./datasources"
 import tableSdk from "../tables"
 import { getIntegration } from "../../../integrations"
 import { context } from "@budibase/backend-core"
+import sdk from "../.."
 
 function checkForSchemaErrors(schema: Record<string, Table>) {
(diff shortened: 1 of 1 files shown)


packages/server/src/sdk/app/datasources/plus.ts
```diff
@@ -96,6 +97,15 @@ export async function buildSchemaFromSource(
   const datasource = await datasources.get(datasourceId)
 
   const { tables, errors } = await buildFilteredSchema(datasource, tablesFilter)
+
+  const oldTables = datasource.entities || {}
+  const tablesToRemove = Object.keys(oldTables).filter(
+    t => !Object.keys(tables).includes(t)
+  )
+  for (const table of tablesToRemove) {
+    await sdk.rowActions.deleteAll(oldTables[table]._id!)
```
adrinr, line 106: I don't like adding the row action stuff mixed with the datasource, but this is the only way to do it without redesigning how this currently works..

samwho (approved):

PR #14861, opened 2024-10-24
Add tests for SQL injection attacks on table/view creation and search.

samwho:
## Description

This PR includes a substantial refactor of `sql.ts` to remove as many calls to Knex raw functionality as we can, as well as shoring up the raw calls we can't remove with checks to make sure the input to them is sanitised.


packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -3470,5 +3472,70 @@ describe.each([
           ])
         })
       })
+
+    isSql &&
+      describe("SQL injection", () => {
+        const badStrings = [
+          "1; DROP TABLE test;",
```
adrinr, line 3479: Should this table names not match the created ones? If the table "test" was actually deleted, would the test catch it?

samwho: I considered that and decided it wasn't really necessary, because the query will fail as the table is missing, causing the test to fail all the same.

packages/backend-core/src/sql/sql.ts
```diff
@@ -795,103 +865,109 @@ class InternalBuilder {
...
+            value,
+          ])
         } else if (this.client === SqlClient.ORACLE) {
           const identifier = this.convertClobs(key)
-          return q[fnc](
-            `(${identifier} IS NOT NULL AND ${identifier} != ?) OR ${identifier} IS NULL`,
-            [value]
+          return (
+            q
+              .where(subq =>
```
mike12345567, line 940: This is a nice improvement, I do find the knex structure a lot easier to follow than interpolated strings!

mike12345567 (approved):
Very nice job - this is a massive improvement, really like the addition of the `rawQuotedIdentifier` - very useful function!

packages/server/src/api/routes/tests/search.spec.ts
samwho: I have addressed this, PTAL 🙏

adrinr (approved):
🚀

mike12345567 (approved):

PR #14862, opened 2024-10-24
Missing old row on automations

adrinr:
## Description
We had an issue where some chained automations were not passing the proper data, missing some data such as "oldRow". The problem was that it was not properly sent, and while debugging the issue I realised that there is not much typing around automation emitters.
This PR is mainly typing everything around emitters to stop using magic strings, parameters and `any`s. It also fixes the actual issue, passing the `oldRow` accordingly.

image

### DIFF
--- a/packages/server/src/events/AutomationEmitter.ts
+++ b/packages/server/src/events/AutomationEmitter.ts
@@ -1,7 +1,15 @@
 import { rowEmission, tableEmission } from "./utils"
 import mainEmitter from "./index"
 import env from "../environment"
-import { Table, Row, DocumentType, App, ContextEmitter } from "@budibase/types"
+import {
+  Table,
+  Row,
+  DocumentType,
+  App,
+  ContextEmitter,
+  EventType,
+  UserBindings,
+} from "@budibase/types"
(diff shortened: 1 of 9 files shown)


packages/server/src/events/AutomationEmitter.ts
```diff
@@ -54,7 +66,9 @@ class AutomationEmitter {
       appId,
       row,
       table,
+      oldRow,
```
adrinr, line 69: This is the actual fix

packages/types/src/sdk/koa.ts
```diff
@@ -49,3 +60,32 @@ export interface UserCtx<RequestBody = any, ResponseBody = any>
 export interface BBContext extends Ctx {
   user?: ContextUser
 }
+
+export interface ContextEmitter {
```
PClmnt, line 64: This is really nice, I was actually confused about this the other day when trying to track down some emitter issues - the fact we had the AutomationEmitter and the standard emitter etc.  Really nice stuff!

PClmnt (approved):
LGTM!

PR #14865, opened 2024-10-24
Misc V3 improvements

aptkingston:
## Description
- Don't send a wasted API request when changing datasource access if the value is unchanged
- Show warning in user side panel if you haven't published your app, letting users know they'll need to publish before other people can access it

image


shogunpurple (approved):
LGTM 👍

PR #14868, opened 2024-10-24
Open API - security update

mike12345567:
## Description
@shogunpurple noticed there was an issue with using the public API to access development apps from non-builder roles, when this occurred the response was a re-direct as we expect in the browser, however this makes little sense in an API call.

I've updated the test cases a bit around the public API to make this a testable scenario, as well as updating `currentapp.ts` to 403 in these cases rather than a re-direct, by detecting it is using an API key.

I've also removed the `isTest` which was hiding this from our test cases and instead check if the call is coming from a browser.

Final small update is the OpenAPI spec was a little hard to use as you had to work out all the variables you needed, I've defaulted these so that when imported to tools that support OpenAPI specifications it shows all the variables listed with descriptions, making it a lot quicker to get up and running.

### DIFF
--- a/packages/server/specs/openapi.json
+++ b/packages/server/specs/openapi.json
@@ -47,6 +47,7 @@
         "required": true,
         "description": "The ID of the app which this request is targeting.",
         "schema": {
+          "default": "{{ appId }}",
           "type": "string"
         }
       },
@@ -56,6 +57,7 @@
         "required": true,
         "description": "The ID of the app which this request is targeting.",
         "schema": {
+          "default": "{{ appId }}",
           "type": "string"
         }
       },
--- a/packages/server/specs/openapi.json
+++ b/packages/server/specs/openapi.json
@@ -8,19 +8,15 @@
   "servers": [
     {
       "url": "https://budibase.app/api/public/v1",
-      "description": "Budibase Cloud API"
-    },
-    {
-      "url": "{protocol}://{hostname}/api/public/v1",
-      "description": "Budibase self hosted API",
+      "description": "Budibase Cloud API",
       "variables": {
-        "protocol": {
-          "default": "http",
-          "description": "Whether HTTP or HTTPS should be used to communicate with your Budibase instance."
+        "apiKey": {
+          "default": "<user API key>",
+          "description": "The API key of the user to assume for API call."
         },
-        "hostname": {
-          "default": "localhost:10000",
-          "description": "The URL of your Budibase instance."
+        "appId": {
+          "default": "<App ID>",
+          "description": "The ID of the app the calls will be executed within the context of, this should start with app_ (production) or app_dev (development)."
         }
       }
     }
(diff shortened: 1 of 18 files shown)


shogunpurple (approved):
LGTM - great tests

packages/server/specs/openapi.json
```diff
@@ -833,7 +831,8 @@
                       "type": "string",
                       "enum": [
                         "static",
-                        "dynamic"
+                        "dynamic",
+                        "ai"
```
samwho, line 835: Need to update the `description` field here as well.

packages/server/src/api/routes/public/tests/Request.ts
```diff
@@ -0,0 +1,110 @@
+import { User, Table, SearchFilters, Row } from "@budibase/types"
+import { HttpMethod, MakeRequestResponse, generateMakeRequest } from "./utils"
+import TestConfiguration from "../../../../tests/utilities/TestConfiguration"
+import { Expectations } from "../../../../tests/utilities/api/base"
+
+type RequestOpts = { internal?: boolean; appId?: string }
+
+type PublicAPIExpectations = Omit<Expectations, "headers" | "headersNotPresent">
```
samwho, line 8: Given you only use status and body it's probably worth making a new type here.

packages/server/src/api/routes/public/tests/Request.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export class PublicAPIRequest {
+  private makeRequest: MakeRequestResponse | undefined
+  private appId: string | undefined
+  private _tables: PublicTableAPI | undefined
+  private _rows: PublicRowAPI | undefined
+  private _apiKey: string | undefined
```
samwho, line 15: This is a bit of a mess, and makes a lot of the rest of the code of this class quite awkward. I know it somewhat mirrors `TestConfiguration` but I eventually want us to fix that, too.

I'd recommend doing something like this:

```typescript
// The name `PublicAPIRequest` feels incorrect, it's not a request, it's
// a bridge to other APIs such as `tables` and `rows`.
export class PublicAPI {
  private makeRequest: MakeRequestResponse
  private apiKey: string
  private appId?: string

  tables: PublicTableAPI
  rows: PublicRowAPI

  constructor(config: TestConfiguration, user: User, opts?: RequestOpts) {
    this.apiKey = await config.generateApiKey(user._id)
    this.makeRequest = generateMakeRequest(this.apiKey, opts)
    this.appId = opts?.appId
    this.tables = new PublicTableAPI(this)
    this.rows = new PublicRowAPI(this)
  }
}
```

Then I think it would be ideal if this lived alongside `config.api` as `config.publicApi`. That would make it much nicer when used, rather than having to create these object by hand when needed.

samwho (changes requested):

PR #14871, opened 2024-10-25
Fix new filtering for with queries

adrinr:
## Description
Fixing an issue that was causing frontend filtering not to work fine, breaking query filtering and possibly other client filtering datasources.

#### All dataset
image

#### Filtering
image


aptkingston:
I think the proper fix for this is changing
```js
    if (!query && this.features.supportsSearch) {
      query = buildQuery(filter)
    }
```
to
```js
    if (!query) {
      query = buildQuery(filter)
    }
```
on line 187. I think that used to always be that way but must have changed recently. I don't see why we shouldn't build the query if we don't support server side searching... We always want the built query, as we use that for client side filtering too.

aptkingston:
> > I think the proper fix for this is changing
> > ```js
> >     if (!query && this.features.supportsSearch) {
> >       query = buildQuery(filter)
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```js
> >     if (!query) {
> >       query = buildQuery(filter)
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > on line 187. I think that used to always be that way but must have changed recently. I don't see why we shouldn't build the query if we don't support server side searching... We always want the built query, as we use that for client side filtering too.
> 
> I thought about that, but this means we will be sending it to the backend, and I assumed that the current change tried to avoid that...

I don't think we will - if you check `QueryFetch` for example, you'll see it never uses `query` when fetching data from the backend, since it knows that it's not supported. So I think it's fine to always have a `query`, as the individual fetch classes know to only use it when they support it.

adrinr:
> > > I think the proper fix for this is changing
> > > ```js
> > >     if (!query && this.features.supportsSearch) {
> > >       query = buildQuery(filter)
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > to
> > > ```js
> > >     if (!query) {
> > >       query = buildQuery(filter)
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > on line 187. I think that used to always be that way but must have changed recently. I don't see why we shouldn't build the query if we don't support server side searching... We always want the built query, as we use that for client side filtering too.
> > 
> > 
> > I thought about that, but this means we will be sending it to the backend, and I assumed that the current change tried to avoid that...
> 
> I don't think we will - if you check `QueryFetch` for example, you'll see it never uses `query` when fetching data from the backend, since it knows that it's not supported. So I think it's fine to always have a `query`, as the individual fetch classes know to only use it when they support it.

Cool! Just changed it, I thought that was added for another reason and something would break

aptkingston (approved):

PR #14872, opened 2024-10-25
Gro 738 refactor code that returns the account holder from user

melohagan:
## Description
- Now getting the account holder by email from the acc_metadata (platform info db)
- Removed tenant_info doc and usage 

Should only be merged after the account-portal side of this has been deployed.

## Addresses
- https://linear.app/budibase/issue/GRO-738/refactor-code-that-returns-the-account-holder-from-user-id
- https://linear.app/budibase/issue/GRO-740/remove-tenant-info-code

## Screenshots

**Account holder:**
account_holder

**Admin user:**
admin


packages/backend-core/src/users/utils.ts
```diff
@@ -67,21 +65,26 @@ export async function validateUniqueUser(email: string, tenantId: string) {
...
+      email: string
     }
+  | undefined
+> {
+  if (!env.SELF_HOSTED && !env.DISABLE_ACCOUNT_PORTAL) {
+    const accountMetadata = await getExistingAccounts(
+      users.map(user => user.email)
+    )
+    return users.find(user =>
+      accountMetadata.map(metadata => metadata.email).includes(user.email)
```
shogunpurple, line 87: will `accountMetadata` be updated if an account holder changes their email address like in the v2 flows?

packages/builder/src/pages/builder/portal/users/users/[userId].svelte
```diff
@@ -206,7 +206,9 @@
     if (!user?._id) {
       $goto("./")
     }
-    tenantOwner = await users.tenantOwner($auth.tenantId)
+    // Just take the first app because the account holder will belong to all apps.
+    const appId = availableApps[0].devId
```
Comment (shogunpurple) [inline packages/builder/src/pages/builder/portal/users/users/[userId].svelte:210]: what happens if there's no apps?

packages/frontend-core/src/api/user.js
```diff
@@ -296,9 +296,9 @@ export const buildUserEndpoints = API => ({
     })
   },
 
-  getTenantInfo: async ({ tenantId }) => {
+  getAccountHolder: async ({ appId }) => {
     return await API.get({
-      url: `/api/global/tenant/${tenantId}`,
+      url: `/api/global/users/holder/${appId}`,
```
shogunpurple, line 301: would make this `accountholder` or `accountowner` for consistency

packages/backend-core/src/users/utils.ts
melohagan: It will have to be. Currently account metadata is created and deleted as part of the onboarding/offboarding flows. With this change, account metadata will be a core document that will need to be handled within transactions in our new flows.

packages/builder/src/pages/builder/portal/users/users/[userId].svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/portal/users/users/[userId].svelte:210]: Screenshot 2024-10-28 at 10 44 51

Currently you can't get into a valid state without at least one app. When you delete your last app, you are presented with the create first app flow.

If you do manage to get into a state with no apps, it throws errors in other parts of the platform.

packages/frontend-core/src/api/user.js
melohagan: Renamed.

packages/backend-core/src/users/utils.ts
```diff
@@ -67,21 +65,26 @@ export async function validateUniqueUser(email: string, tenantId: string) {
...
-    }
-
-    const budibaseUserId = account.budibaseUserId
-    if (userIds.includes(budibaseUserId)) {
-      return account
+export async function getAccountHolderFromUsers(
+  users: Array<{
+    userId: string
+    email: string
+  }>
```
samwho, line 74: Given that this `{ userId: string, email: string }` type is used in various places, it likely makes sense to extract it out into its own named type.

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -302,6 +309,20 @@ export const tenantUserLookup = async (ctx: any) => {
   }
 }
 
+/**
+ * This will be paginated to a default of the first 50 users,
+ * So the account holder may not be found until further pagination has occurred
+ */
+export const accountHolderLookup = async (ctx: any) => {
```
samwho, line 316: Can this `ctx: any` be a more specific type?

packages/worker/src/api/controllers/global/users.ts
```diff
@@ -302,6 +309,20 @@ export const tenantUserLookup = async (ctx: any) => {
...
+  const appId = ctx.params.appId
+  ctx.body = {
+    appId,
+  }
+  await getAppUsers(ctx)
```
samwho, line 321: Could we instead call `userSdk.db.getUsersByAppAccess` here?

packages/builder/src/pages/builder/portal/users/users/[userId].svelte
Comment (melohagan) [reply packages/builder/src/pages/builder/portal/users/users/[userId].svelte]: Refactored to not rely on appId.

shogunpurple (approved):
LGTM!

PR #14876, opened 2024-10-25
Use account email instead of budibaseUserId

melohagan:
## Description
Refactoring `identifyUser` code to use the email address of the account instead of the budibaseUserId

## Addresses
- https://linear.app/budibase/issue/GRO-738/refactor-code-that-returns-the-account-holder-from-user-id


jvcalderon (approved):
LGTM

PR #14877, opened 2024-10-25
Support multiple relationships same tables

adrinr:
## Description
Fixing a couple of issues related to relationships:
1. Fixing deleting entries with many-to-many linked data
2. Fixing updates with multiple many-to-many between the same tables

### DIFF
--- a/packages/server/src/api/controllers/row/ExternalRequest.ts
+++ b/packages/server/src/api/controllers/row/ExternalRequest.ts
@@ -443,7 +443,6 @@ export class ExternalRequest<T extends Operation> {
         rows: Row[]
         isMany: boolean
         tableId: string
-        field: string
       }
     > = {}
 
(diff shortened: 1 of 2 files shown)


packages/server/src/api/controllers/row/ExternalRequest.ts
```diff
@@ -425,8 +427,12 @@ export class ExternalRequest<T extends Operation> {
    */
   async lookupRelations(tableId: string, row: Row) {
```
mike12345567, line 428: Could be worth defining the type for this return - so it has a named type and its bit easier to follow, an array is a good choice.

mike12345567 (commented):
Looks good - one question around this - should we merge this to `master` - or should it go to `v3-ui` as this is a feature/enhancement?

Also - we will need to remove the checks from the frontend which disable this.

samwho (approved):

packages/server/src/api/controllers/row/ExternalRequest.ts
adrinr: I refactored the returned type to be back to object, with a helper to get the key. This way we extract the logic from the consumer of this function

mike12345567 (approved):

PR #14878, opened 2024-10-25
Allow bigints in calculations

samwho:
## Description

This PR fixes that bigints can't be used in calculations. This was disallowed in the frontend, which was easy to fix, but I noticed while testing that we weren't accurate with bigints close to the 64bit mark. This was due to them being converted to JS numbers before leaving the API, subjecting them to the limitations of JS's int53 numbers.

I fixed this and added tests to make sure we support summing up to the 2 ** 63 - 1 boundary of a signed 64 bit integer.


packages/backend-core/src/sql/sql.ts
```diff
@@ -179,12 +179,6 @@ class InternalBuilder {
     return this.table.schema[column]
   }
 
-  private supportsILike(): boolean {
-    return !(
-      this.client === SqlClient.ORACLE || this.client === SqlClient.SQL_LITE
-    )
-  }
```
samwho, line 186: Was unused, a holdover from my SQL security PR from yesterday.

packages/backend-core/src/sql/sql.ts
```diff
@@ -216,8 +210,30 @@ class InternalBuilder {
     return formatter.wrap(value, false)
   }
 
-  private rawQuotedValue(value: string): Knex.Raw {
```
samwho, line 219: Was unused, a holdover from my SQL security PR from yesterday.

mike12345567 (approved):
LGTM - nice work!

PR #14880, opened 2024-10-25
Revert "Write Redis data to the persistent data dir in single image."

CSLTech:
Reverts Budibase/budibase#14797

This causes the redis database to become corrupted when running under hosting that shares files for multiple instances (Eg: Google Cloud Run)

### DIFF
--- a/hosting/single/Dockerfile
+++ b/hosting/single/Dockerfile
@@ -69,9 +69,6 @@ WORKDIR /minio
 COPY scripts/install-minio.sh ./install.sh
 RUN chmod +x install.sh && ./install.sh
 
-# setup redis
-COPY hosting/single/redis.conf /etc/redis/redis.conf
-
 # setup runner file
 WORKDIR /
 COPY hosting/single/runner.sh .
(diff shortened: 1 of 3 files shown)


samwho:
Hey @CSLTech! Could you explain the problem you're seeing a bit more? You're running multiple instances of the Budibase image that share persistent file storage?

CSLTech:
@samwho We're running budibase on google cloud run (https://docs.budibase.com/docs/google-cloud-run). However, the redis database keeps getting corrupted since this update.

The google cloud run uses the FILESHARE_IP environment variable to mount a shared NFS disk for persistance.

samwho:
And you're running multiple instances of Budibase all using the same fileshare?

CSLTech:
The issue was still happening when setting the number of instances to 1, but I think cloud run may still start multiple instances in the background. In all cases, running multiple instances would be the default for cloud run.

samwho:
Didn't realise closing this would also lock it. If you can't reply anymore, and you have a problem with the proposed solution, please file an issue and ping me directly in it. 🙏

CSLTech:
Thank you for your work on this @samwho. This is certainly a viable option (Could also have piggybacked on the existing FILESHARE_IP option).

samwho:
It crossed my mind, but it felt a little more descriptive/future-proof to introduce a new variable for this purpose. Thanks for your understanding on all this!

PR #14888, opened 2024-10-28
Data section toolbar standardisation

aptkingston:
## Description
This PR updates all the controls in the data section toolbar to use the same component - detail popovers. Currently there are 4 different components used for the various controls - popovers, detail popovers, modals and drawers. Having a single experience for all controls is a much nicer experience for the user.

Functionality wise, the only change is to the export button. Since it's only available on tables (where filters cannot be applied) I could massively reduce the complexity of that component. Now you can only export all data, or the selected rows.

### Import button
image
image

## Export button
image
image

## Filter button
image
image

## Sort button
image
image

## Size button
image
image

## Columns button
image
image

## View calculations button
image
image


deanhannigan (approved):
LGTM! Awesome update @aptkingston! Really refines the UX in the data section. Only one issue found during testing around the import popover not locking down the types, but I've already gone through that with you. 👍

Issue #14870, opened 2024-10-24
Automation history pagination changes to disabled state for premium user if filter is selected


hawkinsjon:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: Digital Ocean
  - Budibase Version: 2.32.10
  - App Version: 2.32.10

**Describe the bug**
As a premium user, I should be able to view up to 30 days of my automation history. 
I'm able to do this as long as I don't touch the filtering options in the automation log area
If any of the filters are selected, the pagination arrow no longer lets me get past page 1

**Screenshots**
Notice when all filter options are showing (ALL) the pagination is accessible. 

<img width="897" alt="Screenshot 2024-10-24 at 2 42 04 PM" src="https://github.com/user-attachments/assets/e161a8c3-0923-426d-9a25-ec2f51090139">

And when even one filter options is selected, pagination is no longer accessible.

<img width="912" alt="Screenshot 2024-10-24 at 2 42 24 PM" src="https://github.com/user-attachments/assets/286c8a53-f8a6-48d7-a049-014f0e9fd5c3">

mike12345567:
This should be fixed in the latest release (2.33.12).

hawkinsjon:
Thanks!

PR #14895, opened 2024-10-29
Fix automation log pagination

adrinr:
## Description
We detected an issue with automation logs: pagination breaks when adding filters. This PR adds some typing, tests and fixes that issue

Related PR:
- https://github.com/Budibase/budibase-pro/pull/329

## Addresses
- https://github.com/Budibase/budibase/issues/14870

## Screenshots

https://github.com/user-attachments/assets/427cca82-1174-4570-82ee-b29d45db5f90

## Launchcontrol
Fixing automation logs filtering

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
@@ -38,8 +38,11 @@
   let loaded = false
   $: app = $appsStore.apps.find(app => $appStore.appId?.includes(app.appId))
   $: licensePlan = $auth.user?.license?.plan
-  $: page = $pageInfo.page
-  $: fetchLogs(automationId, status, page, timeRange)
+
+  // Reset the page every time that a filter gets updated
+  $: pageInfo.reset(), automationId, status, timeRange
+
+  $: fetchLogs(automationId, status, $pageInfo.page, timeRange)
   $: isCloud = $admin.cloud
   $: chainAutomations = app?.automations?.chainAutomations ?? !isCloud
   const timeOptions = [
--- a/packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
@@ -42,7 +42,8 @@
   // Reset the page every time that a filter gets updated
   $: pageInfo.reset(), automationId, status, timeRange
 
-  $: fetchLogs(automationId, status, $pageInfo.page, timeRange)
+  $: page = $pageInfo.page
+  $: fetchLogs(automationId, status, page, timeRange)
   $: isCloud = $admin.cloud
   $: chainAutomations = app?.automations?.chainAutomations ?? !isCloud
   const timeOptions = [


packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
```diff
@@ -38,8 +38,11 @@
   let loaded = false
   $: app = $appsStore.apps.find(app => $appStore.appId?.includes(app.appId))
   $: licensePlan = $auth.user?.license?.plan
-  $: page = $pageInfo.page
-  $: fetchLogs(automationId, status, page, timeRange)
+
+  // Reset the page every time that a filter gets updated
+  $: pageInfo.reset(), automationId, status, timeRange
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:43]: @aptkingston not sure if this the most "svelte-way" to achieve this. Could I get your thoughts?

mike12345567 (approved):
Approval on backend/pro side - will await @aptkingston opinion on the log fetching!

samwho (approved):

packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:43]: This is fine yea. We could really do with making a proper `AutomationLogFetch` or something at some stage because I really don't like this custom pagination implementation, but this reactive statement is totally fine in the meantime to achieve what we need.

packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte
```diff
@@ -38,8 +38,11 @@
   let loaded = false
   $: app = $appsStore.apps.find(app => $appStore.appId?.includes(app.appId))
   $: licensePlan = $auth.user?.license?.plan
-  $: page = $pageInfo.page
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:41]: Is it safe to remove this? We probably pulled this out so that it's stored in a primitive and can be properly compared. I'm worried that `fetchLogs(automationId, status, $pageInfo.page, timeRange)` could cause an endless loop as `$pageInfo.page` will re-trigger every time the `pageInfo` changes at all, which happens when we load data, which would trigger another refresh of data... etc.

I'm assuming this doesn't happen since you would have noticed it, but I would prefer to keep this as is and pass in the primitive `page` as the param to `fetchLogs` to keep reactivity to a minimum.

Comment (adrinr) [reply packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:41]: I can leave it as it is, but if there is a loop it would happen like this as well, will it not? It would be a reactivity chain anyways

Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:41]: Svelte uses shallow comparison (`==`) to determine if a value has changed or not, so if you have
```js
$: page = $pageInfo.page
$: console.log(page)
```
And the `pageInfo` store updates but the `page` (number) value stays the same, then the the log will not fire, because the primitive value stored in the first reactive statement will be equal when using shallow comparison. That's a benefit of using primitives for reactive statement values. If it was an object, then it would always retrigger no matter what.
```js
$: console.log($pageInfo.page)
```
Would trigger every time `pageInfo` updates even if no change is made to `page` directly, since it is observing the entire `pageInfo` store rather than just the primitive `page`.

It's a useful svelte trick to know.

Comment (adrinr) [reply packages/builder/src/pages/builder/app/[application]/settings/automations/index.svelte:41]: Cool, thx for the info! I did undo these changes :)

aptkingston (approved):

PR #14896, opened 2024-10-29
Fix default values when using multi-option column and supplying empty array.

samwho:
The expected behaviour here is to revert back to the default value, rather than continuing with an empty array. Same behaviour should apply to empty strings.


mike12345567 (approved):
LGTM!

PR #14899, opened 2024-10-29
Support view RBAC with forms and cache view definitions

aptkingston:
## Description
This PR fixes an issue where forms always attempted to load the definition of their parent table when being used with a view, which is unnecessary and threw permission errors if the user did not also have access to this table.

Another improvement on this PR will massively help reduce API calls to the server - caching view definitions in client memory. We already cache table definitions, but we did not do the same for view definitions. These cannot change at runtime so it is safe to cache. On a table autoscreen with side panels, this reduces **8** duplicate API calls for the view definition into 1.

Before:
image

After:
image


adrinr (approved):
LGTM!

PR #14904, opened 2024-10-29
Account for logicalOperator inside UISearchFilter groups

aptkingston:
## Description
Fixes an issue where logical operators inside groups in frontend filter structures were ignored when converting to a query.

Given a sample frontend filter group (`SearchFilterGroup`) of:
```js
{
    "logicalOperator": "any",
    "filters": [
        {
            "valueType": "Binding",
            "field": "Country",
            "type": "string",
            "constraints": {
                "type": "string",
                "length": {},
                "presence": false
            },
            "operator": "fuzzy",
            "noValue": false,
            "value": ""
        },
        {
            "valueType": "Binding",
            "field": "Region",
            "type": "string",
            "constraints": {
                "type": "string",
                "length": {},
                "presence": false
            },
            "operator": "fuzzy",
            "noValue": false,
            "value": ""
        }
    ]
}
```
Notice the `any` value for `logicalOperator`, meaning perform an "or".

This was previously converted into:
```js
{
    "$and": {
        "conditions": [
            {
                "fuzzy": {
                    "Country": ""
                }
            },
            {
                "fuzzy": {
                    "Region": ""
                }
            }
        ]
    }
}
```
Notice the `$and` operator which is incorrect.

With this PR it is now correctly converted into:
```js
{
    "$or": {
        "conditions": [
            {
                "fuzzy": {
                    "Country": ""
                }
            },
            {
                "fuzzy": {
                    "Region": ""
                }
            }
        ]
    }
}
```

This is safe to go to `v3-ui` rather than master as these structures do not exist in master yet.


mikesealey (approved):

samwho (changes requested):
Could you add a test for this? 🙏 There's a big group of them in `viewV2.spec.ts` around line 4059.

aptkingston:
> Could you add a test for this? 🙏 There's a big group of them in `viewV2.spec.ts` around line 4059.

I looked for tests inside shared-core where this resides and there were none, but looks like that's because the tests for it live elsewhere! Will do - I was going to if I had found them.

samwho:
Yeah, sorry about that. I didn't want to make assertions on the output of `buildQuery` because it would be quite brittle. Fundamentally what matters is whether it results in the right rows coming back to you, so I decided to put them in `viewV2.spec.ts`.

samwho (approved):

PR #14916, opened 2024-10-30
Reenable cache

adrinr:
## Description
Reenabling the cache after figuring it out what the issue was. We follow the pattern defined in [the docker docs](https://docs.docker.com/build/cache/optimize/) around caching, specially to solve the slowness while building multiplatform architectures.
We figure out that the issue was that the parallel runs were causing problems with the yarn cache, one overriding the other. Because of this we were seeing some build working while others would fail.

This PR reenables the cache we removed, and fixes the issue using different directories per each architecture.

### DIFF
--- a/.github/workflows/budibase_ci.yml
+++ b/.github/workflows/budibase_ci.yml
@@ -324,5 +324,12 @@ jobs:
         with:
           node-version: 20.x
           cache: yarn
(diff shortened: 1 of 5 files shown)


mike12345567 (approved):
LGTM!

PR #14918, opened 2024-10-30
Ai fixes

shogunpurple:
## Description
Some fixes for the AI functionality for v3. Importing an AI field from a table import caused empty values to be written, and the AI column functionality was not correctly wired up to our posthog feature flagging system.

## Addresses
- https://linear.app/budibase/issue/BUDI-8786/bulk-importing-an-ai-field-does-not-work-also-overwrites-existing-data
- https://linear.app/budibase/issue/BUDI-8787/ai-columns-not-feature-flagged

### DIFF
--- a/packages/server/src/utilities/rowProcessor/index.ts
+++ b/packages/server/src/utilities/rowProcessor/index.ts
@@ -216,10 +216,6 @@ export async function inputProcessing(
     if (field.type === FieldType.FORMULA) {
       delete clonedRow[key]
     }
-    // remove any AI values, they are to be generated
-    if (field.type === FieldType.AI) {
-      delete clonedRow[key]
-    }
     // otherwise coerce what is there to correct types
     else {
       clonedRow[key] = coerce(value, field.type)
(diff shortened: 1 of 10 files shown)


samwho (approved):

packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -216,10 +216,6 @@ export async function inputProcessing(
     if (field.type === FieldType.FORMULA) {
       delete clonedRow[key]
     }
-    // remove any AI values, they are to be generated
```
adrinr, line 219: Why this? Do we want to persist/analyse these imports?

samwho: As I understood it, this was the thing stripping AI column results when doing a bulk import.

shogunpurple: that's it - when you re-import a CSV that has AI columns they were being stripped. We don't want to make the user generate them again, unless they specifically click to do so

PR #14921, opened 2024-10-30
Conditions reactivity fix + row actions improvement

aptkingston:
## Description
- Fixes an issue where component settings added then removed by conditions would fail to be reflected in components https://linear.app/budibase/issue/BUDI-8771/conditions-on-child-components-arent-updating-when-a-data-provider-is
- Automatically refreshes datasources after triggering row actions via button actions


deanhannigan (approved):
LGTM!

PR #14922, opened 2024-10-30
Search filter typing

adrinr:
## Description
While debugging I realised we were handling conditions incorrectly, as `Object.keys` was treating equally basic operation and logical operations. I refactored the code in order to simplify it and to make a better usage of the types

### DIFF
--- a/packages/shared-core/src/filters.ts
+++ b/packages/shared-core/src/filters.ts
@@ -142,25 +142,6 @@ export function recurseLogicalOperators(
   return filters
 }
 
-export function recurseSearchFilters(
-  filters: SearchFilters,
-  processFn: (filter: SearchFilters) => SearchFilters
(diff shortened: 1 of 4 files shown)


packages/shared-core/src/filters.ts
```diff
@@ -142,25 +142,6 @@ export function recurseLogicalOperators(
   return filters
 }
 
-export function recurseSearchFilters(
```
mike12345567, line 145: Where did this move to?

mike12345567 (approved):

packages/shared-core/src/filters.ts
adrinr: This usage was slightly wrong and only used in one place. This merged it and refactored it in `packages/server/src/threads/automation.ts` directly

PR #14923, opened 2024-10-30
Fix builder serve endpoint error rate

mike12345567:
## Description
Fixing an issue with serving the builder, increased error rate due to requests that hit the / endpoint, these were 500ing due to a lack of app ID.

### DIFF
--- a/packages/server/src/api/controllers/static/index.ts
+++ b/packages/server/src/api/controllers/static/index.ts
@@ -6,7 +6,6 @@ import { ObjectStoreBuckets } from "../../../constants"
 import { processString } from "@budibase/string-templates"
 import {
   loadHandlebarsFile,
-  streamFile,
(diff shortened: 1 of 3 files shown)


packages/server/src/api/controllers/static/index.ts
```diff
@@ -145,6 +146,16 @@ const requiresMigration = async (ctx: Ctx) => {
 }
 
 export const serveApp = async function (ctx: UserCtx) {
+  if (ctx.url.includes("apple-touch-icon")) {
```
samwho, line 149: Should we be more specific here? e.g. `ctx.url.includes("/apple-touch-icon.png")`?

samwho (approved):

shogunpurple (approved):
LGTM

PR #14924, opened 2024-10-30
Fix view calculations with relationships

samwho:
## Description

Prior to this PR, creating a calculation view on a table with any relationship columns would make that calculation view always return no data. This PR fixes that, and adds a test.

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -3659,6 +3659,109 @@ describe.each([
               )
             })
 
+            it("should be able to filter on relationships", async () => {
+              const companies = await config.api.table.save(
+                saveTableRequest({
+                  schema: {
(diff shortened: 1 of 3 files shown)


packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -3659,6 +3659,109 @@ describe.each([
...
+                {
+                  name: "Aperture Science Laboratories",
+                }
+              )
+
+              const blackMesa = await config.api.row.save(companies._id!, {
+                name: "Black Mesa",
+              })
+
+              await config.api.row.save(employees._id!, {
```
mike12345567, line 3733: Might be nice to group these together for speed with a `Promise.all([config.api.row.save...`

mike12345567 (approved):
LGTM! Pain about the lack of bulk import but good tests!

PR #14925, opened 2024-10-30
Throw an error when attempting to bulkImport a relationship field into an internal table.

samwho:
## Description

This cost myself and @mike12345567 an hour or so today. If you bulk import a relationship field into an internal table it does entirely the wrong thing. It should just be an error for now, to avoid confusion in future.

This shouldn't be reachable behaviour in the UI, because the CSV that gets important is first sent to a validation endpoint that strips out relationship fields. This is mostly to help idiots like me write tests without wasting time.


mike12345567 (approved):
LGTM - thanks for getting this together!

PR #14928, opened 2024-10-31
Allow fields on count calculations.

samwho:
## Description

Prior to this PR, calculation view "count" columns didn't take a `field`. They just assumed the field was `*`. This PR makes `field` required for counts, which allows you to count specific non-null values of a given column.


aptkingston (approved):
Thanks Sam!

PR #14929, opened 2024-10-31
Allow tables to handle _id values in non DS+ rows

aptkingston:
## Description
This PR allows the usage of existing values for an `_id` column on non-datasource-plus sources. An example is Mongo, which has it's own concept of `_id` separate to Budibases. The current table component assumes that non DS+ sources will not contain this field, and therefore overwrites it with a UUID in order to be able to uniquely identify each row.

This PR will now use existing `_id` values if they exist, **trusting that they are unique**, and continuing to generate them if not.

I've also updated the additional context we pull from tables to better reflect the available data. This means that fields like `_id` and `__idx` which the table generates for internal use will not appear in live binding evaluation.

## Assumptions
The grid has 2 remaining assumptions.
- `_id` values in non DS+ rows are unique
- `_id` values in non DS+ rows do not start with `‽‽` (that's not a question mark, it's an obscure character chosen since it's extremely rare and will never be used)

## Addresses
A customer issue.


deanhannigan (approved):
LGTM!

PR #14930, opened 2024-10-31
Allow counting non-numeric fields in calculation views

aptkingston:
## Description
This PR allows counting non-numeric fields in calculation views. Previously this was only possible on numeric fields, and counts didn't correctly rehydrate in the UI as the `field` property was stripped.

Example - counting the amount of rows with each country (an options field) in a table:
image
image


deanhannigan (approved):
LGTM!

PR #14932, opened 2024-10-31
Fix flaky export test.

samwho:
## Description

We've seen this failure in CI a few times lately:

image

Looked into it and found our CSV generation code on row export does not properly escape strings, so quotes can mess with the structure of the CSV generated.


mike12345567 (approved):
LGTM!

PR #14937, opened 2024-11-01
Sample data fix

mike12345567:
## Description
Fixing an issue with recent [PR](https://github.com/Budibase/budibase/pull/14925) for throwing an error on bulk import - this makes sure that sample data can be imported (as these schemas have relationship columns, but no values).


PClmnt (approved):
LGTM!

PR #14939, opened 2024-11-01
Minor OpenAPI updates

mike12345567:
## Description
Some OpenAPI changes, fixing defaults for app ID, role endpoint permissions and addition to router (they weren't accessible) and export role examples.

This is from my testing with the new view public APIs, backporting as they won't be in V3 launch.


shogunpurple (approved):
LGTM

PR #14945, opened 2024-11-04
Update color picker to account for new theme values

aptkingston:
## Description
This PR fixes an issue with the color picker in component settings, where it would not correctly account for the app theme in certain circumstances.


deanhannigan (approved):
LGTM!

PR #14946, opened 2024-11-04
Fix issue with branch step bindings

PClmnt:
## Description
The branch object wasn't being added to context correctly so adding a branch binding such as `branch.status` was not working.


deanhannigan (approved):
LGTM! 🚀

PR #14949, opened 2024-11-04
Patch row improvements

adrinr:
## Description
Updating the way that patch row works to fix a couple of issues around row patching

## Launchcontrol
Row patch api improvements


mike12345567 (approved):

mike12345567 (approved):

packages/server/src/integrations/tests/utils/index.ts
```diff
@@ -102,6 +102,9 @@ function createDummyTest() {
 }
 
 export function datasourceDescribe(opts: DatasourceDescribeOpts) {
+  // tests that call this need a lot longer timeouts
+  jest.setTimeout(120000)
```
adrinr, line 106: Why is this introduced in here? Nothing we did should really affect that

mike12345567: It shouldn't - but it seems that under certain circumstances (guessing github free runners under load) the tests can fail due to timeouts in oracle/DS startup. This branch was failing consistently the other day without this.

PR #14954, opened 2024-11-04
Fix issue counting number of filters when legacy structure is used

### LINKED ISSUE #14947: Update Setting Filtering to -> "No flinters set" even when filters have been set
**Checklist**

**Hosting**

- Cloud - miketest1
- AND
- Self - v3.0.0 Docker Single Image

**Describe the bug**
In an existing app that has a cards-block, "Update Setting" conditions are added to change filtering. In the app, there are 6 cases where filtering is updated, all of which have filters established, but all rows show "No filters set".

Showing no fitlers set
image

When clicking into "no filters set", filters exist
image

**To Reproduce**
Unable to create a fresh app where this occurs - I believe it may be to do with the way v3 Filter Groups interact with established filters.

**Expected behavior**
Button should reflect how many filters are already set.

**Screenshots**
https://jam.dev/c/b3d99d34-1bdb-463c-a182-ea761efb61ff
  
**App Export**
App-export available on request

 - OS: MacOS
 - Browser Chrome

**Additional context**
Issue discovered in a customer's app

aptkingston:
## Description
This PR fixes a visual issue where we displayed "No filters set" even when filters were set, if the legacy filter structure of a flat array was used (which would be the case for all existing apps prior to V3).

I've also removed a few extra `cloneDeep` usages which weren't needed as we clone inside the `CoreFilterBuilder` anyway, so these were just dupes hurting performance.

Before:
image

After:
image

## Addresses
- Fixes https://github.com/Budibase/budibase/issues/14947

### DIFF
--- a/packages/builder/src/components/design/settings/controls/FilterEditor/FilterEditor.svelte
+++ b/packages/builder/src/components/design/settings/controls/FilterEditor/FilterEditor.svelte
@@ -12,6 +12,7 @@
   import FilterBuilder from "./FilterBuilder.svelte"
   import { tables, selectedScreen } from "stores/builder"
   import { search } from "@budibase/frontend-core"
+  import { utils } from "@budibase/shared-core"
 
   const dispatch = createEventDispatcher()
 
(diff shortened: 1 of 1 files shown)


deanhannigan (approved):
LGTM!

PR #14956, opened 2024-11-04
Disable major update from actions

adrinr:
## Description
Remove the "major" option when releasing new versions. If we run a major version by mistake, we will not be able to roll back easily, so we are removing this option from the picker. If we need a v4 upgrade, we will add this option, but removing it is the safest.


samwho (approved):

mike12345567 (approved):
Good change - smart thinking!

Issue #14958, opened 2024-11-04
v3 AI Config issue


andyburgessmd:
**Checklist**
- I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.0.2
  - App Version: 3.0.2

**Describe the bug**
Using a update form shows AI settings realted errors but nothing is using any AI. The errors was showing an error something like 'Error: No AI Config' so I went to settings where the /builder/portal/settings/ai was shown as default and I added a provider to see if that would fix it. Then I got the message 'Error: No config selected as default AI config',
![image](https://github.com/user-attachments/assets/f6ac9fc7-1172-4d0f-a841-60e98fb85bed)

I have tried to go back to settings to set the AI provider as default but it won't allow that and shows this message.
![image](https://github.com/user-attachments/assets/d4b4da1f-e988-4f2a-a406-0f3b02754ce6)

I've only found this on the first form I have tried after updating an app to 3.0.2 so nothing else has changed and I'm not going to continue with any others yet in case they break

shogunpurple:
Hi @andyburgessmd - looking into this, this feature should be feature flagged and not required if there's no AI config. Must be something to do with paid tiers. Will keep you updated. Best to roll back for now if you are on the premium plan

andyburgessmd:
@shogunpurple rolled the app back but still shows this error

andyburgessmd:
@shogunpurple 3.0.3 has fixed it. Thanks

PR #14959, opened 2024-11-04
Flag ai correctly for premium

shogunpurple:
## Description
Fixing an issue described in https://github.com/Budibase/budibase/issues/14958#issuecomment-2455778091 which causes AI columns to be rolled out for premium self host users - updated the areas of the code so it respects the overall feature flag ahead of the AI launch

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/CronBuilder.svelte
@@ -9,7 +9,7 @@
   } from "@budibase/bbui"
   import { onMount, createEventDispatcher } from "svelte"
   import { flags } from "stores/builder"
-  import { featureFlags } from "stores/portal"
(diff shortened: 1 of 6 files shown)


adrinr (approved):
A lot of duplicated checks. Maybe worth either having some helpers or adding the license check within the feature flag check

shogunpurple:
@adrinr I considered this but the overall feature flag will be removed after the release, so I'll do another PR to remove it. This is the first paid feature that we have flagged, so it requires both checks unfortunately - the overall one and in the license

adrinr:
> @adrinr I considered this but the overall feature flag will be removed after the release, so I'll do another PR to remove it. This is the first paid feature that we have feature flagged, so it requires both checks unfortunately - the overall one and in the license

Ok, it makes sense if it's a hard one to tackle

PR #14965, opened 2024-11-05
further AI config updates for case where there's no configuration

shogunpurple:
## Description
Fixing a few issues noticed with AI config, feature flags and licensing.

- Reset quota to zero when you remove the entire AI config
- Only show the AI settings panel when the feature flag is on
- Updates in `pro` here (https://github.com/Budibase/budibase-pro/pull/331) make sure that you can run as normal on self host when you have no custom configs enabled


packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -518,7 +525,11 @@ export async function destroy(ctx: UserCtx) {
     await db.remove(id, rev)
     await cache.destroy(cache.CacheKey.CHECKLIST)
     if (id === configs.generateConfigID(ConfigType.AI)) {
-      await pro.quotas.removeCustomAIConfig()
+      await pro.quotas.set(
```
adrinr, line 528: Just to double check, this is not meant to use the new helper added here? https://github.com/Budibase/budibase-pro/pull/331/files#diff-a0aad30e4928dd505d83d6c01819c048f2b553140dea91a84c97ef945602d8edR32

adrinr (approved):

packages/worker/src/api/controllers/global/configs.ts
shogunpurple: Nope, this particular one is to just reset all the quotas back to 0 - I could actually get rid of that helper tbh as it's not really necessary

PR #14967, opened 2024-11-05
Fix race conditions between automation test data and runs

adrinr:
## Description
When running automation tests, we don't use the updated automation data properly. This might cause test runs with "stale" data.
The issue is that `AutomationBlockSetup` updates the data via API, and refetches it. In parallel, we run the test run, passing the locally fetched test data. If this happens before the refetch finishes (that will happen almost every time with a minimum of latency) , the data will be stale.
This PR ensures that the data is properly passed to the test.

### Before

https://github.com/user-attachments/assets/8491de38-1ef0-4051-ad2e-7ba803b4a280

### After

https://github.com/user-attachments/assets/04cac9a1-e211-4552-b95c-fb5ca32f9faf

## Launchcontrol
Fix race condition when running automation tests


adrinr:
Low priority. Do not merge this until the V3 release is completed.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
```diff
@@ -152,7 +150,7 @@
   {#if selectedValues}
     <div class="tab-content-padding">
       <AutomationBlockSetup
-        testData={$memoTestData}
+        bind:testData
```
deanhannigan, line 153: We're trying to steer away from using `bind` when working with props in the builder. It can lead to issues with mutation.

deanhannigan (approved):
LGTM! Had a run through this and it seems to resolve the issue. I did make one minor note but its not a blocker.

packages/builder/src/components/automation/AutomationBuilder/FlowChart/TestDataModal.svelte
adrinr: I assume that, but it would require quite a lot of reactors around the stores... Given that we are going to type/cleanup the stores soon, happy if we check it then?

PR #14968, opened 2024-11-05
fix for conditional automation step options always showing

andz-bb:
## Description
Automation step options that should conditionally render based on other settings are currently always showing. I think the only place this is used is the send email automation step where calendar invite options should be shown only when **Add calendar invite** is checked.

## Addresses
[BUDI-8820](https://linear.app/budibase/issue/BUDI-8820/conditional-automation-step-options-always-rendering-send-email)

## Screenshots
**before**
SCR-20241105-kzny

**after**
SCR-20241105-lbqe

## Launchcontrol
fix for send email automation step settings

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -865,7 +865,7 @@
   <!-- Custom Layouts -->
   {#if stepLayouts[block.stepId]}
     {#each Object.keys(stepLayouts[block.stepId] || {}) as key}
-      {#if canShowField(key, stepLayouts[block.stepId].schema)}
+      {#if canShowField(stepLayouts[block.stepId].schema)}
         {#each stepLayouts[block.stepId][key].content as config}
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -865,7 +865,7 @@
   <!-- Custom Layouts -->
   {#if stepLayouts[block.stepId]}
     {#each Object.keys(stepLayouts[block.stepId] || {}) as key}
-      {#if canShowField(key, stepLayouts[block.stepId].schema)}
+      {#if canShowField(stepLayouts[block.stepId].schema)}
```
andz-bb, line 868:
```ts
  function canShowField(value) {
    const dependsOn = value?.dependsOn
    return !dependsOn || !!inputData[dependsOn]
  }
```
this function used to have key and value parameters but key was removed, meaning all cases where this was being called with two arguments were broken

deanhannigan (approved):
LGTM! Nice catch!

Issue #14974, opened 2024-11-06
Show a visual indicator whenever a file is uploaded via an attachment field


ConorWebb96:
**Describe the feature request**
As a user, sometimes I upload files to a Budibase app using slower Wi-Fi. I would like to see that the upload is still in progress. This feature would greatly improve my user experience, as it would provide a simple spinner or any visual cue to let me know that something is still indeed happening, reducing my uncertainty.

Currently, nothing is shown. It just updates once the file has finished uploading.

Issue #13827, opened 2024-05-31
Table using an external database with a formula column as the primaryDisplay show it on the far right instead of respecting the order of the dataview.


ConorWebb96:
**Checklist**

- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

* Self
  * Method: docker compose
  * Budibase Version: 2.27.5
  * App Version: 2.27.5

**Describe the bug**
This issue only appears if the table component is linked to a `dataprovider` as its source. If you directly reference the table the order will be correctly reflected.

**To Reproduce**

1. Import app attached below.
2. Use the table create script attached below to create the database.
3. Fetch the tables in Budibase
4. Add a formula column, doesn't matter what its called.
5. Set the formula column to be the primaryDisplay column.
6. Create a new screen that has a dataprovider linked to the newly created table.
7. Add a table component nested inside the data provider, make sure the table is using the data provider as the source of its info.
8. See the error, if the formula is the primary display it will go to the far right of the screen instead of being the first column.

**Expected behavior**
primaryDisplay should always be the first column shown in the table.

**Screenshots**

![Screenshot 2024-05-31 at 10 37 59](https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/8c398297-06b8-4dbd-85df-1bc3bf1b9423/734a213c-bac0-4dc4-85cf-46f3cd828260?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC84YzM5ODI5Ny0wNmI4LTRkYmQtODVkZi0xYmMzYmYxYjk0MjMvNzM0YTIxM2MtYmFjMC00ZGM0LTg1Y2YtNDZmM2NkODI4MjYwIiwiaWF0IjoxNzE3MTQ4NjA0LCJleHAiOjMzMjg3NzA4NjA0fQ.IJ3pCFVHABtPD2UuSt8AEevVztje2nOJoZuhwgJ-Gu0 "=1512x")

[...]

PR #14982, opened 2024-11-06
Misc fixes

aptkingston:
## Description
- Fix display column position being incorrect when using a DP as a table source https://github.com/Budibase/budibase/issues/13827 https://github.com/Budibase/budibase/issues/15011
- Increase width of create/edit column popover in data section to better display configuration for types like relationships
Old (very cramped and cut off):
image
New:
image
- Add loading spinner when uploading files. This also prevents attempting to upload additional files while an upload is in progress as I've disabled the dropzone while a file is uploading. https://github.com/Budibase/budibase/issues/14974
image
- Prevent adding multiple fields to single attachment cells and fields. You can now only select one file as intended. https://linear.app/budibase/issue/BUDI-8503/can-select-multiple-attachments-for-an-attachment-single-column


deanhannigan (commented):
Ran through the all the issues listed here and have some feedback

- The display column is now showing on the left as expected when a table uses a data provider. Also, the upload field changes are working perfectly 👍 
- The CreateEditColumn popover change is visible when editing an existing column but not when creating a new one. Could this be extended to the create flow as well? 

- When adding multiple attachments to the same single attachment field, I still see the same behaviour
Screenshot 2024-11-15 at 12 39 53

aptkingston:
> * The CreateEditColumn popover change is visible when editing an existing column but not when creating a new one. Could this be extended to the create flow as well?

It's the same component, and it definitely is there for creating a new column... What did you click to see the small one?

Creating: 360 width
image

Editing: 360 width 
image

> * When adding multiple attachments to the same single attachment field, I still see the same behaviour
>   Screenshot 2024-11-15 at 12 39 53

How did you add multiple? You can't select more than one file in the file picker now (unless it's a multi attachment column of course), and you can't upload a new one until you delete the current file.

deanhannigan (approved):
LGTM!

PR #14983, opened 2024-11-06
fix helm chart secret to correctly reapply the existing secret

### LINKED ISSUE #14981: helm chart secrets are not correctly reapplied if one of the secret keys contains an empty string
**Checklist**

**Hosting**

- Self
  - Method: k8s(helm) 
  - Budibase Version: 2.33.2 
  - Chart Version: 3.0.216 

**Describe the bug**
When the helm chart is upgraded or synced and there is an existing secret that has an empty string for one or more of the keys, the keys with the empty strings are removed from that secret

but since the empty string is a valid base64 encoded value for the empty string this will cause budibase to crash since the environment variable is now missing 

**To Reproduce**
1. create a secret with one of the data keys as the empty string "" e.g `apiEncryptionKey`
2. upgrade the chart to a newer version or sync the state from the chart back to the cluster
3. wait
4. the budibase app will crash because of a missing env variable like `apiEncryptionKey`

**Expected behavior**
keys will not be removed from the existing secret

**Additional context**
this happens because inside the `charts/budibase/templates/secrets.yaml` the secret is read and reapplied from a lookup variable and not quoted before readding

cstruck:
## Description
This PR fixes #14981 

## Addresses
- `#14981`


cstruck:

cstruck:
recheck

shogunpurple (approved):
LGTM, thanks for the contribution @cstruck !

PR #14984, opened 2024-11-06
Use real Bull queues for testing

adrinr:
## Description
We are using `inMemoryQueue` when running unit tests instead of Bull, as we do when running Budibase. This is causing some inconsistencies in the way we validate things when depends on some queue features, as they don't behave the same way (for example, cron jobs). As we already have all the setup available to get rid of `inMemoryQueue`, this PR replaces it for the proper bull usage. This is only done when needed, as if all the tests share the same Redis instance we might have tests interfering with each other.

Pro PR:
- https://github.com/Budibase/budibase-pro/pull/332


packages/backend-core/src/cache/tests/docWritethrough.spec.ts
```diff
@@ -242,50 +248,48 @@ describe("docWritethrough", () => {
...
+        expect(await db.tryGet(documentId)).toEqual(
           expect.objectContaining(patches)
         )
       })
     })
 
-    // This is not yet supported
-    // eslint-disable-next-line jest/no-disabled-tests
-    it.skip("patches will execute in order", async () => {
+    it("patches will execute in order", async () => {
```
adrinr, line 273: We can finally enable this test! Queues behave the same way in tests and in bb instances! 🕺🏽

packages/backend-core/tests/core/utilities/utils/queue.ts
```diff
@@ -0,0 +1,31 @@
...
+        `until redis-cli  ping | grep -q PONG; do
+          echo "Waiting for Redis to be ready..."
+          sleep 1
+        done
+        echo "Redis is ready!"`
+      ).withStartupTimeout(10000)
+    )
+    .start()
+
+  const port = redis.getMappedPort(6379)
```
samwho, line 21: `getMappedPort` causes problems with dual-stack (v4/v6) systems, see comment [here](https://github.com/Budibase/budibase/blob/ee20f2eba8d8c4a2dcd21661871183787b1d8601/packages/server/src/integrations/tests/utils/index.ts#L220-L228).

I _strongly_ recommend using the `startContainer` function in the linked file. It works around several problems that used to plague us when using `testcontainers`.

packages/backend-core/tests/core/utilities/utils/queue.ts
```diff
@@ -0,0 +1,31 @@
...
+    .withExposedPorts(6379)
+    .withEnvironment({})
+    .withLabels({ "com.budibase": "true" })
+    .withReuse()
+    .withWaitStrategy(
+      Wait.forSuccessfulCommand(
+        `until redis-cli  ping | grep -q PONG; do
+          echo "Waiting for Redis to be ready..."
+          sleep 1
+        done
```
samwho, line 15: I don't think we should loop here. `Wait.forSuccessfulCommand` already implements its own loop to retry this command.

packages/backend-core/src/cache/tests/docWritethrough.spec.ts
```diff
@@ -10,15 +15,14 @@ import {
...
-
 const initialTime = Date.now()
 
 async function waitForQueueCompletion() {
-  const queue: InMemoryQueue = DocWritethroughProcessor.queue as never
-  await queue.waitForCompletion()
+  await utils.queue.processMessages(DocWritethroughProcessor.queue)
 }
 
+beforeAll(() => utils.queue.useRealQueues())
```
samwho, line 24: Would it be worth treating `redis` like we treat `minio` and `couchdb`, adding them to `globalSetup.ts` and making sure they're always running during tests?

adrinr: I started this way, but we might have conflicts sharing redis instances as all the tests might affect each other...

packages/backend-core/tests/core/utilities/utils/queue.ts
adrinr: It this not happening because we are creating the test container and retrieving it separately? Will this happen with using the test container directly?

packages/backend-core/src/cache/tests/docWritethrough.spec.ts
samwho: Ahhh yes, I am remembering now that I also did try this. Redis doesn't have great ways to isolate tests away from each other like the DBs do (where we create a new database every time).

packages/backend-core/tests/core/utilities/utils/queue.ts
adrinr: Done!

samwho (approved):

PR #14989, opened 2024-11-06
Remove watchtower

mike12345567:
## Description
As title - this has been part of our self-host stack for a while now, however it only supports updating our docker-compose services. We have decided to remove this as it has caused enough trouble before now - instead replacing it with a link on how to update your self-host installation safely.


hosting/proxy/nginx.prod.conf
```diff
@@ -107,10 +106,6 @@ http {
       proxy_pass      $apps;
     }
 
-    location = /v1/update {
```
shogunpurple, line 110: There's a potential breaking scenario here I think - if someone upgrades using BB CLI and they have a custom NGINX config it won't boot I believe

shogunpurple (approved):
LGTM

hosting/proxy/nginx.prod.conf
mike12345567: Discussed on Discord, should be safe, this is just a route update.

PR #14991, opened 2024-11-06
Issue where looping wasn't working due to missing context

PClmnt:
## Description
Due to the update to bindings to use step Ids, getting the loop iterations was broken due to the context we passed being incorrect. This resulted in loops not working at all


mike12345567 (approved):
LGTM - thanks for the solid test case!

PR #14994, opened 2024-11-07
Create JWT secret keys on boot.

samwho:
## Description

I noticed when profiling our tests that we pass `JWT_SECRET` as a string to `jwt.sign`. This causes the `jwt` library to create a secret key every time we encode or decode a `jwt`, which is quite often in tests and at least once per request in production.

This is accounting for a little under 1% of CPU time in production:

CleanShot 2024-11-07 at 17 07 00@2x

`createPublicKey` appears to be what takes the most time, and I assume it fails. The code to create a KeyObject in the `jwt` library looks like this:

```js
if (secretOrPublicKey != null && !(secretOrPublicKey instanceof KeyObject)) {
  try {
    secretOrPublicKey = createPublicKey(secretOrPublicKey);
  } catch (_) {
    try {
      secretOrPublicKey = createSecretKey(typeof secretOrPublicKey === 'string' ? Buffer.from(secretOrPublicKey) : secretOrPublicKey);
    } catch (_) {
      return done(new JsonWebTokenError('secretOrPublicKey is not valid key material'))
    }
  }
}
```

Where `secretOrPublicKey` is the `JWT_SECRET` string we pass in.

If I try this locally I get an error that looks like it could be environmental, but given that we don't _actually_ pass in a properly formatted public key in production, and both `createPublicKey` and `createSecretKey` can be seen in the profiles, I assume this fails in production too:

```js
> const { createPublicKey } = require("crypto")
undefined
> createPublicKey("foo")
Uncaught Error: error:1E08010C:DECODER routines::unsupported
    at createPublicKey (node:internal/crypto/keys:619:12) {
  library: 'DECODER routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_UNSUPPORTED'
}
> 
```

To test that this isn't going to cause problems on deployment, I ran `yarn dev` locally with the current JWT_SECRET values, logged in, then I changed the code so it called `createSecretKey` in `environment.ts` and refreshed the page. I was still logged in, which I believe means the JWT in the cookies correctly verifies after this change.


mike12345567 (approved):
Very nice find!

adrinr (approved):

PR #14999, opened 2024-11-08
Allow standalone field components without forms

aptkingston:
## Description
This PR allows using field components without forms. That means they do not need to be wrapped in a `Form` component (but still can be, of course) and provide a global binding of their value to use as you like.

Almost all field types support this usage. The only exceptions are attachments (both single and multiple), signatures and relationship pickers, as these require being bound to a datasource to be functional.

Clarification:
- The `field` setting is now optional when using fields standalone. The component name rather than the field setting will be used to generate the global binding
- All settings work as expected - validation, default values, on change etc.

### Quick example of filtering using a standalone field

Here's a screen with a table, and I've dropped an options picker above it:
image

Here's my component tree - just the table and some fields without a form:
image

I can filter my table using the new global bindings:
image

The field configuration can be left completely blank (I've added a placeholder but that's optional of course):
image

### Quick example of saving rows
You also don't need a form to create rows when using fields like this. You can simply add a button, leave datasource to "None" and bind up the values:
image

## Launchcontrol
Allow using fields without forms.


deanhannigan (approved):
LGTM 🚀 ! This is game changer! It removes another barrier to rapidly building richer, leaner apps. It's so much more intuitive to simply drag in a custom field component and hook it into your app without a form in sight. For the S3 and QR field, especially, it really makes things more accessible.

I ran through all the field types and didn't come across any issues. I just made the following observations which are not blockers:

- For the `Multiselect` field, should the `Schema` option be available in the `Options source` setting when not inside a form?
  - At the minute it does nothing as there is no parent form.
  - Could the `Schema` settings adapt when the field is isolated to allow picking a datasource and column to populate the picker?
- Should there now be a `Clear field/s` action that will list one or more isolated fields to clear?

aptkingston:
> LGTM 🚀 ! This is game changer! It removes another barrier to rapidly building richer, leaner apps. It's so much more intuitive to simply drag in a custom field component and hook it into your app without a form in sight. For the S3 and QR field, especially, it really makes things more accessible.
> 
> I ran through all the field types and didn't come across any issues. I just made the following observations which are not blockers:
> 
> * For the `Multiselect` field, should the `Schema` option be available in the `Options source` setting when not inside a form?
>   
>   * At the minute it does nothing as there is no parent form.
>   * Could the `Schema` settings adapt when the field is isolated to allow picking a datasource and column to populate the picker?
> * Should there now be a `Clear field/s` action that will list one or more isolated fields to clear?

Thanks for the review!

For the options sources, unfortunately I can't do anything here although it would be nice if we could. Everything is generic, so we can't conditionally change these settings depending on how the field component is used. We'd have to create a new settings type specifically for "options source" and I don't think that's worth it in this case.

Regarding actions - that's a good point and something I'll maybe look at in a follow up PR. There are a few enhancements we could maybe make with actions to make the integration even smoother, but I don't want to hold up this PR as it may be a while before I get round to that.

PR #15000, opened 2024-11-08
prevent attachment column updates when importing budibase templates

andz-bb:
## Description
We want to be able to import apps from templates while preserving the attachment data from those template dbs. `updateAttachmentColumns` is great when importing an app with attachments from a file, but wrecks the attachment data in our templates which is always pointing to a public s3 bucket rather than the instance's own object storage

for example, an attachment in my template app db:
```json
"Image":{"size":604516,"name":"amsterdam.jpg","extension":"jpg","key":"amsterdam.jpg","url":"https://prod-budi-templates.s3.eu-west-1.amazonaws.com/attachments/expense-approval/amsterdam.jpg"}
```

gets completely stripped out by the above function. I've added in a flag to prevent it from running when we're importing from a template rather than a file (i.e. when `template.key` has a value)

## Launchcontrol
fix for attachments in budibase templates

### DIFF
--- a/packages/server/src/sdk/app/backups/imports.ts
+++ b/packages/server/src/sdk/app/backups/imports.ts
@@ -170,7 +170,10 @@ export async function importApp(
   appId: string,
   db: Database,
   template: TemplateType,
-  opts: { importObjStoreContents: boolean } = { importObjStoreContents: true }
+  opts: {
+    importObjStoreContents: boolean
+    updateAttachmentColumns?: boolean
+  } = { importObjStoreContents: true, updateAttachmentColumns: true }
 ) {
   let prodAppId = dbCore.getProdAppID(appId)
   let dbStream: any
@@ -219,7 +222,9 @@ export async function importApp(
   if (!ok) {
     throw "Error loading database dump from template."
   }
-  await updateAttachmentColumns(prodAppId, db)
+  if (opts.updateAttachmentColumns) {
+    await updateAttachmentColumns(prodAppId, db)
+  }
   await updateAutomations(prodAppId, db)
   // clear up afterward
   if (tmpPath) {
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM!

packages/server/src/sdk/app/backups/imports.ts
```diff
@@ -170,7 +170,10 @@ export async function importApp(
   appId: string,
   db: Database,
   template: TemplateType,
-  opts: { importObjStoreContents: boolean } = { importObjStoreContents: true }
+  opts: {
+    importObjStoreContents: boolean
+    updateAttachmentColumns?: boolean
+  } = { importObjStoreContents: true, updateAttachmentColumns: true }
```
aptkingston, line 176: It would be better to make `updateAttachmentColumns` not optional here so we don't accidentally miss usages of it. Defaulting the params like this works well if the entire `opts` param is undefined, but if you explicitly did something like `importApp(id, db, { importObjStoreContents: true })` then `updateAttachmentColumns` would be `undefined`, whereas the default behaviour before this was was to always we do.

We actually do this in `updateWithExport` today, and therefore the behaviour there would change as we would *not* update attachments any more.

Basically all we need to do is make `updateAttachmentColumns` non-optional, then your IDE will correctly flag any missing usages of it (like `updateWithExport`) and prompt you to explicitly specify it and set it to `true`.

aptkingston (changes requested):
Just one sneaky change needed due to how default params work, but otherwise LGTM!

packages/server/src/sdk/app/backups/imports.ts
andz-bb: good spot, thanks cheeks

aptkingston (approved):

PR #15001, opened 2024-11-08
Automation Branching UX updates

deanhannigan:
## Description

Some additional UX updates to the new Automation branching work.

## Addresses
-  Added autoscrolling when dragging and dropping a step. Now when you drag a step to the extreme edge of the automation canvas, the automation will now scroll in that direction.
- UI updates - added the same colouring and background pattern as the new role management UI. Both screens have a very similar UX this makes sure they share the same visual style.
- Updated the canvas dragging behaviour to allow infinite dragging. CSS classes can be passed as a prop to customise the canvas dragging behaviour for whatever content is being displayed

## Screenshots

Auto scrolling when dragging

drag

Updated UI - added in the same styling as the role management UI. Also reflowed some of the buttons at the top
Screenshot 2024-11-08 at 14 13 13

## Launchcontrol

Automation branching UX updates!


aptkingston (approved):
Awesome work @deanhannigan! This feels so much better and fixes a bunch of the weird edge cases I had encountered before.

PR #15002, opened 2024-11-08
Formula response types

mike12345567:
## Description
As demonstrated in watercooler - lab day work to support formula response types.

Currently formulas will generally return strings, the only exception is dynamic Javascript formulas can return native types, however it is quite undefined. Having a method to actually define what the response means that we can coerce the results to check they fit this result type.

For now I've restricted to a couple of core primitive types that can be returned from the rows API - strings, numbers, booleans and datetimes.

This is done at the API layer, so all rows returned from internal/external tables with a formula will now have the correct primitive type based on what is set. Formulas without anything set will continue to have the exact functionality they have today (HBS string, JS less defined).

This also means handlebars can actually return a primitive now from a formula, which can be very helpful.

## Screenshots
UI:
image

Grid correctly representing formulas of these response types:
image


aptkingston (approved):
Frontend changes LGTM!

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -386,6 +387,7 @@
       editableColumn.relationshipType = RelationshipType.MANY_TO_MANY
     } else if (editableColumn.type === FieldType.FORMULA) {
       editableColumn.formulaType = "dynamic"
+      editableColumn.responseType = FIELDS.STRING.type
```
adrinr, line 390: Will this persist responseTypes? Will this not override on new editions?

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -3227,35 +3233,106 @@ datasourceDescribe(
...
+        expect(rows[0].formula).toBeUndefined()
+      })
+
+      it("should coerce handlebars", async () => {
+        await updateFormulaColumn("{{ add 1 1 }}", {
+          responseType: FieldType.NUMBER,
+        })
+        const { rows } = await config.api.row.search(table._id!)
+        expect(rows[0].formula).toBe(2)
+      })
```
samwho, line 3309: It's a bit nitty but could you add a test case that does `{{ add 1 1 }}` but returns a string? Just to make sure the string get passed through correctly, didn't see that in other tests.

packages/server/src/utilities/rowProcessor/utils.ts
```diff
@@ -80,12 +93,17 @@ export async function processFormulas<T extends Row | Row[]>(
...
           rows[i] = {
             ...row,
             [column]: tracer.trace("processStringSync", {}, span => {
               span?.addTags({ table_id: table._id, column, static: isStatic })
-              return processStringSync(formula, context)
+              const result = processStringSync(formula, context)
+              try {
+                return responseType ? coerce(result, responseType) : result
+              } catch (err) {
+                // if the coercion fails, we return empty row contents
```
samwho, line 104: In case it matters down the line, wanna slap the error message into the trace?

```typescript
span.addTags({ coercionError: err.message })
```

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -784,7 +784,7 @@
           ]}
           getOptionLabel={option => option.name}
           getOptionValue={option => option.type}
-          tooltip="Formulas by default will return a string - however if you need a native type the response can be coerced."
+          tooltip="Formulas by default will return a string - however if you need a another type the response can be coerced."
```
samwho, line 787:
```suggestion
          tooltip="Formulas by default will return a string - however if you need another type the response can be coerced."
```

samwho (approved):

PR #15003, opened 2024-11-10
Csp updates

shogunpurple:
## Description
This PR contains code that moves the management of our `Content-Security-Policy` to our node backend, rather than managing it through NGINX. This was a security requirement for an enterprise customer, and in order to remove the `unsafe-inline` rule from our CSP, which allows any kind of inline script, we need to dynamically generate a hash (or nonce) and have that returned dynamically in the CSP header, as well as injected into any script tags that Budibase needs to run. Doing this has the added benefit of allowing people to control and switch off CSP altogether with an environment variable - `DISABLE_CONTENT_SECURITY_POLICY`. Today, users do this by bind mounting the NGINX config and disabling it directly there, which means that they have effectively "ejected" their NGINX config, and if we update it they won't get the update in later versions of our `proxy-service` - causing reverse proxy issues in some cases.

It also turns the CSP on in dev. I think this is valuable, as it means we can catch CSP issues before pushing to QA or even prod. If there's any dev specific CSP exceptions that need to be made, we can do it in the CSP middleware rather than turning it off altogether.

**Testing**
- [x] Test backwards compatibility with old NGINX configuration
- [x] Test that embeds and plugins still work as expected
- [x] Potentially add another env var (on by default) that allows people to keep the `unsafe-inline` rule, to ease migration, as this is technically a change for one customer therefore the exception rather than the rule

## Addresses
- https://linear.app/budibase/issue/BUDI-8834/move-csp-into-serverworker-code-instead-of-nginx

## Launchcontrol
Move Content-Security-Policy management into the node app code, rather than hardcoding it in NGINX. New `DISABLE_CONTENT_SECURITY_POLICY` env var to allow people to switch it off without editing NGINX config


packages/server/src/environment.ts
```diff
@@ -114,6 +114,7 @@ const environment = {
     DEFAULTS.JS_RUNNER_MEMORY_LIMIT,
   LOG_JS_ERRORS: process.env.LOG_JS_ERRORS,
   DISABLE_USER_SYNC: process.env.DISABLE_USER_SYNC,
+  DISABLE_CONTENT_SECURITY_POLICY: process.env.DISABLE_CONTENT_SECURITY_POLICY,
```
mike12345567, line 117: Don't need to duplicate this - if its something that is common across services its fine for it to just be in `backend-core/src/environment.ts`

packages/worker/src/environment.ts
```diff
@@ -46,6 +46,7 @@ const environment = {
   SMTP_FALLBACK_ENABLED: process.env.SMTP_FALLBACK_ENABLED,
   DISABLE_DEVELOPER_LICENSE: process.env.DISABLE_DEVELOPER_LICENSE,
   BUDIBASE_ENVIRONMENT: process.env.BUDIBASE_ENVIRONMENT,
+  DISABLE_CONTENT_SECURITY_POLICY: process.env.DISABLE_CONTENT_SECURITY_POLICY,
```
mike12345567, line 49: Same as previous comment.

mike12345567 (approved):
LGTM! One minor comment about environment files - thats it!

PR #15004, opened 2024-11-10
dependency upgrades for security scanners

shogunpurple:
## Description
Updating some of our dependencies for passing vulnerability checks. This is still WIP - there's a few more I need to test and make sure are ready before merging.

Pro PR: https://github.com/Budibase/budibase-pro/pull/333

## Addresses


packages/server/src/api/controllers/query/import/sources/curl.ts
```diff
@@ -2,9 +2,8 @@ import { ImportSource, ImportInfo } from "./base"
 import { Query } from "../../../../../definitions/common"
 import { URL } from "url"
 
-const curlconverter = require("curlconverter")
-
-const parseCurl = (data: string): any => {
+const parseCurl = async (data: string): Promise<any> => {
+  const curlconverter = await import("curlconverter")
```
mike12345567, line 6: out of interest, why is it necessary to import this way? Is there a problem with a top level import like `import curlconverter from "curlconverter"`

packages/server/src/integrations/microsoftSqlServer.ts
```diff
@@ -281,8 +281,14 @@ class SqlServerIntegration extends Sql implements DatasourcePlus {
...
           const { domain, trustServerCertificate } =
             this.config.ntlmConfig || {}
+
+          if (!domain) {
+            throw Error("Domain must be provided for NTLM config")
+          }
+
           clientCfg.authentication = {
             type: "ntlm",
+            // @ts-ignore
```
mike12345567, line 291: Could you add a comment to explain why the ts-ignore is necessary? Going forward going to try and follow the `ts-expect-error` practice to note there is a problem that can't be solved here.

packages/server/src/api/controllers/query/import/sources/curl.ts
shogunpurple: Yep, the later versions of the `curlconverter` package use top level `await`, and you can't just import it normally without changing `"type": "module"` in `package.json`. This is actually causing the rest tests to fail, which I'm looking at atm

mike12345567 (approved):
LGTM!

packages/server/src/api/controllers/query/import/sources/curl.ts
mike12345567: Very interesting - didn't realise this would happen with top level awaits, good to know that they make the library a lot less usable!

packages/server/src/api/controllers/query/import/sources/curl.ts
```diff
@@ -1,10 +1,9 @@
 import { ImportSource, ImportInfo } from "./base"
 import { Query } from "../../../../../definitions/common"
 import { URL } from "url"
-
 const curlconverter = require("curlconverter")
```
mike12345567, line 4: Its probably worth throwing a comment in here to explain the pain of trying to convert curlconverter to `import`!

PR #15015, opened 2024-11-16
enable proxy buffering for calls to OIDC endpoints

shogunpurple:
## Description
Sometimes OIDC responses are very heavy - lots of headers and big response bodies. This was causing an issue in our NGINX layer for MSFT AD.


calexiou (approved):
lgtm

PR #15017, opened 2024-11-18
Clean enriched relationships flag

adrinr:
## Description
Clean enrich relationships flag, as it's shipped 100%


packages/builder/src/components/backend/DataTable/buttons/grid/GridColumnsSettingButton.svelte
```diff
@@ -48,7 +48,7 @@
   </svelte:fragment>
   <ColumnsSettingContent
     columns={$tableColumns}
-    canSetRelationshipSchemas={isEnabled(FeatureFlag.ENRICHED_RELATIONSHIPS)}
+    canSetRelationshipSchemas
```
samwho, line 51: Do we not want to remove this property altogether?

samwho (approved):

packages/builder/src/components/backend/DataTable/buttons/grid/GridColumnsSettingButton.svelte
adrinr: This is used internally within a self. I will check if it can be cleanup and rely only on the backend data

adrinr: Worked fine :)

PR #15021, opened 2024-11-18
fix dnd handlers

shogunpurple:
## Description
Fixing a problem with inline handlers and drag and drop in the builder. We were using a direct inline handler to prevent the default drag behaviour. I've updated this to use sveltes system so we don't trigger the CSP

We also no longer need the `unsafe-inline` script fallback, because it doesn't work with the nonce based system we have implemented.


aptkingston (approved):
LGTM!

PR #15025, opened 2024-11-18
Detach account portal submodule

adrinr:
## Description
Cleaning up all the account-portal references from the monorepo

Related:
- https://github.com/Budibase/account-portal/pull/842


nx.json
```diff
@@ -2,7 +2,6 @@
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
   "tasksRunnerOptions": {
     "default": {
-      "runner": "nx-cloud",
```
adrinr, line 5: This is not actually used, removing to avoid confusions

lerna.json
```diff
@@ -2,11 +2,6 @@
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "version": "3.2.4",
   "npmClient": "yarn",
-  "packages": [
-    "packages/*",
```
shogunpurple, line 6: is this the default? Do we not want to keep `packages/*` around for the standard bb packages?

samwho (approved):

PR #15026, opened 2024-11-18
Fix block native search

aptkingston:
## Description
Block native search (only available on the cards block now) had not been updated for the new search structure.

## Addresses
- Issue raised on Discord


PClmnt (approved):
LGTM!

PR #15027, opened 2024-11-18
Direct describe rather than datasourceDescribe

mike12345567:
## Description
Draft discussion - converting one test case back to use direct `describe`.

This is necessary since some IDEs (like webstorm) need specific `describe` statements for their Jest integrations to work, otherwise all useful tooling is lost.


adrinr (approved):

adrinr:
I think it's a good idea, otherwise there is too much abstraction that makes it hard to use if you are not familiar with the set up, as well

samwho (approved):

PR #15032, opened 2024-11-19
Bypass transformers that are trivial, allow users to configure bson buffer size.

samwho:
## Description

This PR checks if transformers are simply just returning the data they are given and if that's the case bypasses them entirely. This avoids the serialization/deserialization overhead of passing data into and out of the JS isolate, especially for MongoDB which does BSON serialization and deserialization.

I'm also making the BSON internal buffer size configurable with an environment variable so that we can change it to whatever we want as can self-host users. Currently you cannot return more than 17 megabytes of data from a MongoDB query and this limitation is imposed by the BSON library because 16 megabytes is the MongoDB document size limit. Queries could be returning multiple documents so this limit could be quite constricting for no good reason.


mike12345567 (approved):
LGTM!

packages/server/src/threads/query.ts
```diff
@@ -136,21 +136,23 @@ class QueryRunner {
...
+      transformer != null &&
+      transformer.length > 0 &&
+      transformer.trim() !== "return data" &&
+      transformer.trim() !== "return data;"
+
+    if (transformer && hasTransformer) {
       transformer = iifeWrapper(transformer)
       let vm = new IsolatedVM()
       if (datasource.source === SourceName.MONGODB) {
         vm = vm.withParsingBson(rows)
```
adrinr, line 152: Should `hasTransformer` only skip this? Can we run the `return data` without the isolated-vm? 🤔

samwho: I'm pretty sure if the transformer just returns the data as-is, it can be skipped. If we only skipped `vm.withParsingBson` we'd have error about trying to pass objects across the isolate boundary.

adrinr (approved):

PR #15033, opened 2024-11-19
Remove all traces of legacy custom layouts

aptkingston:
## Description
This PR removes all legacy code dealing with custom layouts. We removed support for custom layouts in July 2022, so I think the time has definitely come to stop supporting them altogether. Current support (before this PR) was only for migrating away from them, but it's been impossible to edit custom layouts for over 2.5 years. I highly doubt there are apps running that are over 2.5 years old but using the latest version of Budibase.

The reason this cannot curently be merged is that all our templates (aside from the 3 new ones) were created using custom layouts. That means that using one of those old templates will create an app that is missing whatever navigation settings existed in the app, as any migration logic that understood the legacy structure has been removed. So until we do something with the multitude of old app templates, this cannot be merged.

I still feel like this is an important piece of work to do - simplify things and remove tonnes of deprecated code - so is worth keeping the PR around as motivation to think about how we want to handle legacy templates.


packages/client/src/components/Router.svelte
```diff
@@ -20,6 +23,55 @@
...
+  $: logoUrl = $orgStore?.logoUrl
+  $: layoutSettings = getLayoutSettings(
+    screen,
+    embedded,
+    application,
+    navigation,
+    logoUrl
+  )
+
+  const getLayoutSettings = (screen, embedded, app, navigation, logoUrl) => {
```
aptkingston, line 40: This is the only real addition in this PR. Logic for how the layout component is rendered has moved from the screen store into the Router here.

PR #15036, opened 2024-11-19
Dependency checking as part of package build

mike12345567:
## Description
For now this is limited to the core backend components, `backend-core`, `server` and `worker`. the `yarn build` command will now fail if a dependency is detected as missing from the package.json. This can easily be rectified by making sure the package is listed correctly (in dev or an actual production dependency).

This is required as the monorepo can hide missing dependencies, since packages can find each other dependencies and therefore have access to things in the monorepo they wouldn't have in production.

I had to rectify quite a few issues that weren't necessarily causing problems in production, a lot of them were dev dependencies and therefore only mattered in the monorepo where they were accessible. This is the main downside of this approach, we need to list all the dependencies of each package here to get CI to pass, when a lot of the dev dependencies would be accessible through the monorepo - however I feel like this isn't that big of a loss since each package listing all of its dependencies very specifically is not a bad thing.


samwho (approved):
Probably want to get sign off from @adrinr here as well, he knows JS/TS build tooling a lot better than I do.

adrinr (commented):
Do we want to add this on the build step? Should this not be another script and part of the pipeline? I am just concerned about adding unnecessary steps on the build, as it's used in nodemon

mike12345567:
> Do we want to add this on the build step? Should this not be another script and part of the pipeline? I am just concerned about adding unnecessary steps on the build, as it's used in nodemon

Ah I didn't realise nodemon used this - ok I'll add it as a separate step!

adrinr (approved):
I love these little wins :) Good idea!

PR #15044, opened 2024-11-20
Fix test-libraries yarn invocation.

samwho:
## Description

I broke our `yarn` invocation for `packages/builder` in `test-libraries` in CI in this PR: https://github.com/Budibase/budibase/pull/15040.

We use `vitest` in `packages/builder` which takes different flags to `jest`.


shogunpurple (approved):
LGTM

PR #15050, opened 2024-11-21
Tenant lookup endpoint - validate email input

mike12345567:
## Description
Adding some validation for tenant lookup endpoint, fail with a better error if something goes wrong with email processing.


packages/worker/src/api/controllers/global/users.ts
```diff
@@ -300,6 +301,10 @@ export const find = async (ctx: any) => {
 
 export const tenantUserLookup = async (ctx: any) => {
   const id = ctx.params.id
+  // is email, check its valid
+  if (id.includes("@") && !emailValidator.validate(id)) {
```
melohagan, line 305: I don't think this will work because the ID could be a user id `us_...` I believe,

melohagan (changes requested):

melohagan (approved):

PR #15052, opened 2024-11-21
Additional instrumentation for DD around CouchDB.

samwho:
## Description

This PR is doing a pass on our `DDInstrumentedDatabase` class, adding some information we noticed to be missing while debugging a recent problem.

It also gets rid of an `any` and fixes the resulting typing around it as a little bonus.

### DIFF
--- a/packages/backend-core/src/db/instrumentation.ts
+++ b/packages/backend-core/src/db/instrumentation.ts
@@ -27,7 +27,7 @@ export class DDInstrumentedDatabase implements Database {
 
   exists(docId?: string): Promise<boolean> {
     return tracer.trace("db.exists", span => {
-      span?.addTags({ db_name: this.name, doc_id: docId })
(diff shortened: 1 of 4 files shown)


packages/backend-core/src/db/instrumentation.ts
```diff
@@ -92,83 +102,115 @@ export class DDInstrumentedDatabase implements Database {
...
+      span.addTags({
+        db_name: this.name,
+        batch_limit: opts?.batch_limit,
+        batch_size: opts?.batch_size,
+        style: opts?.style,
+        timeout: opts?.timeout,
+        num_doc_ids: opts?.doc_ids?.length,
+        query_params: opts?.query_params,
+        view: opts?.view,
+        selector: opts?.selector,
```
mike12345567, line 184: including the `selector` and `query_params` here could be dangerous, this could include user inputs.

mike12345567 (approved):
LGTM

PR #15053, opened 2024-11-21
Make bigint fields return as string values when fetched through relationships.

samwho:
## Description

In https://github.com/Budibase/budibase/pull/14878 I did some work to make sure bigints were never converted to JavaScript numbers while being processed inside of Budibase, because doing so subjects them to the inaccuracies of JavaScript's [int53 limitation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). In doing this I created a bug where when queried directly, bigints were returned as strings but queried indirectly, through relationships, they were still numbers in external databases.

The cause of this is how we fetch relationship data. What we do is a "join and squish" approach. To avoid JOINs creating many "duplicate" rows, we squash related records into a JSON array column such that queries return 1 row for each row in the original table, and then up to 500 rows squished into a JSON array for each relationship. 

When constructing JSON objects in most databases we support, bigints are inserted into the JSON as numbers. This is technically correct because JSON numbers have arbitrary size. Most of our database drivers parse this JSON using the built-in `JSON.parse` from JavaScript, which will convert numbers to JavaScript's native number type, losing precision.

To fix this, this PR essentially casts bigint columns to strings in relationship queries. To achieve that, however, took significant effort. Column schema data for relationships needed to be plumbed all the way down into `sql.ts` mandatorily, and it was not mandatory to do so prior to this PR. The vast majority of this PR is that plumbing work.

Pro PR: https://github.com/Budibase/budibase-pro/pull/339

## Addresses
- https://linear.app/budibase/issue/BUDI-8874/improve-bigint-support-for-external-dbs

## Launchcontrol

- Bigint fields will not work correctly when fetched through relationships.


packages/server/src/api/controllers/table/external.ts
```diff
@@ -54,7 +54,7 @@ export async function updateTable(
     return table
   } catch (err: any) {
     if (err instanceof Error) {
-      ctx.throw(400, err.message)
+      throw err
```
samwho, line 57: Put this back as it was.

packages/server/src/sdk/app/rows/search/internal/sqs.ts
```diff
@@ -495,6 +485,7 @@ export async function search(
     if (err.status === 400 && msg?.match(MISSING_COLUMN_REGEX)) {
       return { rows: [] }
     }
-    throw new Error(`Unable to search by SQL - ${msg}`, { cause: err })
+    // throw new Error(`Unable to search by SQL - ${msg}`, { cause: err })
+    throw err
```
samwho, line 489: Put this back how it was.

mike12345567 (approved):
LGTM - discussed in call, follow-up PR to come with change to remove `entityId` and `datasourceId` from the `EnrichedQueryJson` as they are no longer used at the SQL generation layer and could add confusion.

packages/types/src/sdk/search.ts
```diff
@@ -171,11 +171,9 @@ export interface QueryJson {
   sort?: SortJson
   paginate?: PaginationJson
   body?: Row | Row[]
-  table?: Table
-  meta: {
-    table: Table
-    tables?: Record<string, Table>
+  meta?: {
     renamed?: RenameColumn
+    oldTable?: Table
```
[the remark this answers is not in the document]
samwho: The answer is "probably not." I think there's a lot more work we can do to simplify this type, but in future PRs.

PR #15054, opened 2024-11-22
Some automation bugs

PClmnt:
## Description
Fixes a couple of issues with automations

- Issue where the selected JS bindings were invalid, updated the backend to support the use of `steps` context when being passed into a step run
- Stringify the value when an object is passed to the backend log step instead of a string
- Disallow filtering on relationship fields in automation filters as this is not supported. 
- 
## Addresses
- https://linear.app/budibase/issue/BUDI-8860/bindings-in-automations-are-not-working-as-expected
- https://linear.app/budibase/issue/BUDI-8782/filtering-on-relationship-fields-in-automations-doesnt-work

## Feature branch env
[Feature Branch Link](http://fb-fix-automation-adjacent-bugs.fb.qa.budibase.net)


packages/server/src/threads/automation.ts
```diff
@@ -385,7 +385,7 @@ class Orchestrator {
     stepIdx: number,
     pathIdx?: number
   ): Promise<number> {
-    await processObject(loopStep.inputs, this.processContext(this.context))
+    await processObject(loopStep.inputs, this.mergeContexts(this.context))
```
mike12345567, line 388: What is this change doing? Is there anyway we can test for why this change was required?

packages/server/src/automations/steps/serverLog.ts
```diff
@@ -61,8 +61,10 @@ export async function run({
   inputs: ServerLogStepInputs
   appId: string
 }): Promise<ServerLogStepOutputs> {
+  if (typeof inputs.text === "object") {
+    inputs.text = JSON.stringify(inputs.text)
+  }
   const message = `App ${appId} - ${inputs.text}`
-  console.log(message)
```
mike12345567, line 65: I think this `console.log` is still required, this is the core function of the `serverLog.ts`

packages/server/src/threads/automation.ts
PClmnt: So two changes 

- First I renamed processContext -> mergeContext as it makes much sense.
- The actual fix is on line 648, stepsById / steps / stepsByName contexts were not being merged together before being sent to the step function. This meant that specifically for JS Scripting block, that when it tried to use it's own `processObject` the context which it was looking for was not there. I can get a test in for it

packages/server/src/automations/steps/serverLog.ts
PClmnt: So stupid on my behalf, my lint threw a warning and I wasn't even paying attention and just removed it. Sorted

samwho: I can 100% see myself doing the same thing. It would be worth adding a comment above the line disabling the lint and explaining that this is a load bearing console.log.

packages/builder/src/stores/builder/automations.js
```diff
@@ -519,6 +524,10 @@ const automationActions = store => ({
         runtimeName = `loop.${name}`
       } else if (idx === 0) {
         runtimeName = `trigger.${name}`
+      } else if (
+        currentBlock?.stepId === AutomationActionStepId.EXECUTE_SCRIPT
+      ) {
+        runtimeName = `steps["${pathSteps[idx]?.id}"].${name}`
```
samwho, line 530: The presence of the `?` indicates that this might end up being `steps["undefeind"].name`. Would that be a bug? If yes, we likely need checks around this so we can remove the question mark.

Same with the pre-existing line below.

packages/server/src/automations/steps/serverLog.ts
```diff
@@ -61,6 +61,9 @@ export async function run({
   inputs: ServerLogStepInputs
   appId: string
 }): Promise<ServerLogStepOutputs> {
+  if (typeof inputs.text === "object") {
```
samwho, line 64: nit: probably a tiny bit more correct to say `typeof inputs.text !== string`.

packages/builder/src/stores/builder/automations.js
PClmnt: think it's more just being defensive here, we just don't want it to error out if somehow `pathSteps[idx] `is undefined. Having a check beforehand is going to result in the same unusable binding regardless.

samwho: How come we don't want it to error out? Isn't an error explaining the problem preferable to silently generating incorrect code?

PClmnt: Because binding generation can continue on for other steps etc. This binding gets inserted onto the user's input. They'll see immediately that it's undefined. I added in some code to surface if this specific binding index was undefined, what do you think?

mike12345567 (approved):
LGTM!

PR #15055, opened 2024-11-22
OpenAPI for Views

### LINKED ISSUE #15041: The CORS error when making requests to the Budibase API from the client
**Checklist**

**Hosting**
- Self
  - Method: docker
  - Budibase Version:3.2.5
- Cloud
  - Tenant ID: hytest

**Describe the bug**
We want to use Budibase for backend management in our React project, but we are getting a CORS error when making requests from the client. As an alternative, we installed Budibase locally on our machine and ran it on localhost:10000, and even though we added the CORS origin settings as environment variables in the Docker Compose files, we are still encountering the CORS error.

 - OS: MacOS
 - Browser Brave
 - Version 1.73.89 Chromium: 131.0.6778.69 (Resmi Derleme) (arm64) Sürümü

mike12345567:
## Description
Expanding the public API to have a set of endpoints for CRUD on views as well as searching views.

The following endpoints have been added:
- POST `/api/public/v1/views` - create a view
- PUT `/api/public/v1/views/:viewId` - update a view
- DELETE `/api/public/v1/views/:viewId` - delete a view
- GET `/api/public/v1/views/:viewId` - retrieve a single view by ID
- POST `/api/public/v1/views/search` - search for a view across all tables by name
- POST `/api/public/v1/views/:viewId/rows/search` - search for rows within a view

This also fixes https://github.com/Budibase/budibase/issues/15041 by setting the CORS headers for the public API correctly to allow using the public API directly from the browser.

Also important to note most of the new additions are the updates to the `openapi.json` and `openapi.yaml` that need committed to the repo so that readme has access to them.


packages/server/src/api/controllers/public/mapping/views.ts
```diff
@@ -0,0 +1,32 @@
+import { View } from "./types"
+import { ViewV2, Ctx } from "@budibase/types"
+import { dataFilters } from "@budibase/shared-core"
+
+function view(body: ViewV2): View {
+  return {
```
adrinr, line 6: Could we use `RequiredKeys` to ensure any further field is mapped?

packages/server/src/api/controllers/public/rows.ts
```diff
@@ -22,13 +22,13 @@ export function fixRow(row: Row, params: any) {
   return row
 }
 
-export async function search(ctx: UserCtx, next: Next) {
+function getSearchParameters(ctx: UserCtx) {
```
adrinr, line 25: Is the name of this function right? Could we make it more verbose?

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -125,6 +126,12 @@ export async function get(ctx: Ctx<void, ViewResponseEnriched>) {
   }
 }
 
+export async function fetch(ctx: Ctx<void, ViewFetchResponseEnriched>) {
```
adrinr, line 129: Why is this needed?

packages/server/src/api/routes/public/tests/Request.ts
```diff
@@ -1,20 +1,32 @@
...
+  ViewV2Type,
+  PublicAPIView,
+} from "@budibase/types"
 import { HttpMethod, MakeRequestResponse, generateMakeRequest } from "./utils"
 import TestConfiguration from "../../../../tests/utilities/TestConfiguration"
-import { Expectations } from "../../../../tests/utilities/api/base"
 
 type RequestOpts = { internal?: boolean; appId?: string }
 
+type Response<T> = { data: T }
```
adrinr, line 16: ❤️

packages/server/src/api/controllers/view/viewsV2.ts
mike12345567: As part of the public APIs we always offer an endpoint to list all resources of that type - this can then be searched by name. I've added this to the internal API so the public API can utilise it.

adrinr (approved):
LGTM!

PR #15056, opened 2024-11-25
further security updates

shogunpurple:
## Description

More security updates for `svelte`, `dd-trace` and a `resolution` override for `cross-spawn`

Pro PR: https://github.com/Budibase/budibase-pro/pull/336


samwho (approved):

adrinr (approved):

package.json
```diff
@@ -25,11 +25,12 @@
     "prettier": "2.8.8",
     "prettier-plugin-svelte": "^2.3.0",
     "proper-lockfile": "^4.1.2",
-    "svelte": "^4.2.10",
+    "svelte": "4.2.19",
```
adrinr, line 28: We should define our best practices in here, as we always have a mix. Should we pin or use minor versions?

adrinr (approved):
LGTM

package.json
shogunpurple: I think we should pin - as it keeps the `yarn.lock` more consistent

PR #15063, opened 2024-11-25
Typescript in svelte in builder

adrinr:
## Description
Enabling typescript on svelte in the builder. This PR only sets the right configuration for enabling using typescript in svelte, making use of it in a file in the builder and one in bbui.
This PR will be followed by another enabling using typescript instead of js, and further refactors of more files.
Some of the complex types had been just mapped to `any` to keep this one simple. Better typing across the stack will follow.

## Launchcontrol
Enable ts in svelte in the builder

## Feature branch env
[Feature Branch Link](http://fb-budi-8866-builder-vite-tooling-svelte.fb.qa.budibase.net)

### DIFF
--- a/packages/bbui/src/Form/Core/DateRangePicker.svelte
+++ b/packages/bbui/src/Form/Core/DateRangePicker.svelte
@@ -2,13 +2,6 @@
   import CoreDatePicker from "./DatePicker/DatePicker.svelte"
   import Icon from "../../Icon/Icon.svelte"
 
-  export let value = null
-  export let disabled = false
-  export let readonly = false
-  export let error = null
-  export let appendTo = undefined
-  export let ignoreTimezones = false
-
   let fromDate
   let toDate
 </script>
--- a/packages/client/src/components/app/charts/AreaChart.svelte
+++ b/packages/client/src/components/app/charts/AreaChart.svelte
@@ -18,8 +18,6 @@
   export let palette
   export let c1, c2, c3, c4, c5
 
-  // Area specific props
-  export let area
   export let stacked
   export let gradient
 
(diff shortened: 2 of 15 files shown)


packages/bbui/src/Form/Core/DateRangePicker.svelte
```diff
@@ -2,13 +2,6 @@
   import CoreDatePicker from "./DatePicker/DatePicker.svelte"
   import Icon from "../../Icon/Icon.svelte"
 
-  export let value = null
-  export let disabled = false
-  export let readonly = false
-  export let error = null
-  export let appendTo = undefined
-  export let ignoreTimezones = false
```
samwho, line 10: Unused?

packages/client/src/components/app/charts/AreaChart.svelte
```diff
@@ -18,8 +18,6 @@
   export let palette
   export let c1, c2, c3, c4, c5
 
-  // Area specific props
-  export let area
```
samwho, line 22: Unused?

samwho (approved):

packages/bbui/src/Form/Core/DateRangePicker.svelte
adrinr: Yes. And typescript complained about that 😄

aptkingston: I've actually removed this file altogether in my branch. BBUI is consumed by frontend-core so I saw the same issues, and this file is actually entirely unused. You can leave it in though - I can deal with the small conflict.

adrinr: I had the feeling that it was not used at all. I just did not want to remove it just in case some old app was using it somehow 🪄

mike12345567 (approved):
LGTM - very nice, didn't require much configuration change!

PR #15065, opened 2024-11-26
Typescript in builder

adrinr:
## Description
Enable ts files (instead of js) in the builder. This PR sets the config for it, and transforms both navigation and plugin store to typescript. Also converted the API file to ts, as it's consumed by these stores. All the other usages (that are consuming it from js) keep working, so this proves that the hybrid model is supported, consuming ts from js. This will allow a phased transformation to ts.

## Launchcontrol
Enabling typescript files in the builder.

## Feature branch env
[Feature Branch Link](http://fb-budi-8866-builder-enable-typescript.fb.qa.budibase.net)


packages/builder/.gitignore
```diff
@@ -5,4 +5,5 @@ package-lock.json
 release/
 dist/
 routify
-.routify/
\ No newline at end of file
+.routify/
+.rollup.cache
```
samwho, line 9: How come this is needed?

samwho (approved):

packages/builder/.gitignore
adrinr: Rollup is now building the ts files, and it creates this cache folder with a lot of bundled files. We don't want this to be in the repo

adrinr: image

mike12345567 (approved):
LGTM - very nice work!

packages/builder/vite.config.mjs
```diff
@@ -3,6 +3,7 @@ import replace from "@rollup/plugin-replace"
 import { defineConfig, loadEnv } from "vite"
 import { viteStaticCopy } from "vite-plugin-static-copy"
 import path from "path"
+import typescript from "@rollup/plugin-typescript"
```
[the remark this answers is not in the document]
adrinr: I think it's required due to some issues with svelte... Without this we get the following error:
image

PR #15072, opened 2024-11-26
Convert client to use vite instead of rollup

aptkingston:
## Description
This PR converts the client library to use vite instead of rollup for bundling.

## Addresses
- https://linear.app/budibase/issue/BUDI-8868/client-library-convert-to-vite


packages/client/tsconfig.json
```diff
@@ -12,6 +16,5 @@
       ],
       "*": ["./src/*"]
     }
-  },
-  "include": ["src/**/*"]
```
adrinr, line 16: Why removing this?

adrinr (approved):
Nice one!

PR #15074, opened 2024-11-27
Remove rollup from bbui

adrinr:
## Description
Remove rollup from bbui

## Feature branch env
[Feature Branch Link](http://fb-budi-8872-remove-rollup-bbui.fb.qa.budibase.net)


packages/bbui/vite.config.js
```diff
@@ -0,0 +1,29 @@
+import { defineConfig } from "vite"
+import { svelte } from "@sveltejs/vite-plugin-svelte"
+import postcss from "postcss"
+import path from "path"
+
+export default defineConfig({
+  build: {
+    sourcemap: true,
```
aptkingston, line 8: Should we only generate sourcemaps for dev environments to reduce bundle size?

packages/bbui/vite.config.js
```diff
@@ -0,0 +1,29 @@
...
+    },
+  },
+  plugins: [
+    svelte({
+      emitCss: true,
+    }),
+  ],
+  css: {
+    postcss,
+  },
```
aptkingston, line 22: I don't think we need to include the default postcss plugin - I think vite should handle it natively.

packages/bbui/package.json
```diff
@@ -14,17 +14,12 @@
...
-    "@rollup/plugin-commonjs": "^16.0.0",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-node-resolve": "^11.2.1",
+    "@sveltejs/vite-plugin-svelte": "1.4.0",
     "postcss": "^8.2.9",
-    "rollup": "^2.45.2",
-    "rollup-plugin-postcss": "^4.0.0",
-    "rollup-plugin-svelte": "^7.1.0",
-    "rollup-plugin-terser": "^7.0.2"
+    "vite-plugin-commonjs": "^0.10.4"
```
aptkingston, line 22: We aren't using this in the vite config so we can remove it.

packages/bbui/package.json
```diff
@@ -14,17 +14,12 @@
...
   "scripts": {
-    "build": "rollup -c"
+    "build": "vite build"
   },
   "devDependencies": {
-    "@rollup/plugin-commonjs": "^16.0.0",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-node-resolve": "^11.2.1",
+    "@sveltejs/vite-plugin-svelte": "1.4.0",
     "postcss": "^8.2.9",
```
aptkingston, line 21: We can remove the postcss plugin if we don't need it in the vite config as mentioned above.

packages/bbui/vite.config.js
adrinr: I saw that not having this created a different output, but it all works fine without it!

aptkingston (commented):
When I build this locally, it generates `bbui.es.mjs` as the bundle, but `package.json` still expects `bbui.es.js`. We can fix this by explicitly telling vite how to name the file:
`fileName: () => "bbui.es.js"`. Or we can update `package.json` to point to the new name.

This now also spits out css as a separate `style.css` file, which we would need to include in all dependent packages. Previously we bundled the CSS into the same JS file, which is what I'm also doing in the client now (via the `vite-plugin-css-injected-by-js`) plugin. I think it might be a better idea to use that plugin here too, rather than update the other packages. The main reason is that plugins or other standalone usages of BBUI would also be required to now pull in this CSS file.

However, I haven't worked out why everything still looks correct in the product with this external CSS that is not imported anywhere... Edit: it's because in dev we don't use the built bundle -  we use the source instead.

aptkingston (approved):

PR #15081, opened 2024-11-28
Automation fixes

### LINKED ISSUE #15059: Values are not passed correct in the automation test screen
**Checklist**

**Hosting**

- Self
  - Method: docker compose 
  - Budibase Version: 3.2.11.
  - App Version: 3.2.11.

**Describe the bug**
When testing the automation, the last value is displayed. However, this is not passed in the search.
With the same values but freshly entered, the search returns a row.

image

image

image

deanhannigan:
## Description
This addresses some general automation issues:
- The recent [race condition fix](https://github.com/Budibase/budibase/pull/14967) inadvertently caused another issue, where the `Test Modal` would appear to clear on blur. This has now been resolved
- During testing, I saw that for row automation triggers like, `Row Updated`, the test config would be reset if you didn't add any fields at all. This was because the default config values were initially ignored even though they were correctly initialised.

## Addresses
- This issue was technically addressed in `3.2.13` but, as I mentioned above, another issue was introduced with that fix https://github.com/Budibase/budibase/issues/15059


aptkingston (approved):
LGTM!

PR #15089, opened 2024-11-28
Remove exited containers when running tests.

samwho:
## Description

Sometimes when you're running tests, containers will die and hang around in your Docker daemon. This causes future attempts to run tests that use a specific database to fail because the container name is already taken. We use a specific name in order to prevent multiple databases of the same type running, but these exited containers are useless and just need `docker rm`ing. This PR automates that.


mike12345567 (approved):
LGTM - this will help me quite a bit!

PR #15090, opened 2024-11-28
Always use the current docker context.

samwho:
## Description

We discovered recently that if you have multiple Docker runtimes open at the same time, it's possible for the packages/server tests to pick one runtime during globalSetup.ts and the other one during the test runs. This results in the tests failing because CouchDB and minio have not been started in the Docker runtime the tests are using.

I was able to reproduce this locally, and setting `process.env.DOCKER_HOST` fixes it. `testcontainers` reads this environment variable when creating its Docker client.

I've duplicated the code between globalSetup.ts and testContainerUtils.ts because I don't want globalSetup.ts important from anything within Budibase in case it triggers `testcontainers` to generate its runtime config. It only does this once then caches it.

### DIFF
--- a/globalSetup.ts
+++ b/globalSetup.ts
@@ -6,6 +6,26 @@ import {
 import { ContainerInfo } from "dockerode"
 import path from "path"
 import lockfile from "proper-lockfile"
+import { execSync } from "child_process"
+
+interface DockerContext {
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM!

Issue #15049, opened 2024-11-21
Bindings in Automations are not working as expected.


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.2.9
**and**
- Cloud
  - Tenant ID: miketest1

**Describe the bug**
When using the bindings menu to populate bindings in an automation flow
* For the JS Scripting Block
It populates with return steps.SupLYakvO.rows
![Screen Recording 2024-11-21 at 10 21 04](https://github.com/user-attachments/assets/d2b572a8-df64-4e7c-bf30-c17fddc4eaa3)

Which then errors
{
"success": false,
"response": "TypeError: Cannot read properties of undefined (reading 'rows')"
}
![image](https://github.com/user-attachments/assets/49c35a3e-1f40-4fdf-ae9c-2554faba2e8d)

* Backend Log populates and runs the handelbars binding fine, though there is a difference between the displayed input in the bindings drawer and the returned input in the automation test

![image](https://github.com/user-attachments/assets/804cba1a-9bc8-4258-9fa1-304e7af86ea5)

However, switching to JavaScript in the bindings drawer
Input
![image](https://github.com/user-attachments/assets/0d0f893f-f9fd-4324-813a-129f206e6779)

Output
![image](https://github.com/user-attachments/assets/dfd1ade4-235a-4e20-9759-b308f031fdd7)

* The difference could be to do with wether or not the returned value is a string. When fleshing out the bug, I have used an if-block to check if the result of Query Rows (called with the handlebars binding) is the same as the result of Query Rows (called with the JS binding). They do not match, and appear to be a string vs a JSON object.

(For reference, comparison value _(JavaScript function)_ is `return $("steps.Query rows.rows")`
![image](https://github.com/user-attachments/assets/d5d7f920-8bc1-460e-8450-44a152a681bd)

**To Reproduce**
Steps to reproduce the behaviour:
This has been an incredibly difficult one to pin down. It's unlikely that you'll be able to reproduce it in a new app.  Please use this app-export.
[Automations Bug-export-1732186126825.tar.gz](https://github.com/user-attachments/files/17844372/Automations.Bug-export-1732186126825.tar.gz)

1. Go to the automations tab
2. trigger an automation
3. see the results

**Expected behavior**
Using the bindings menu should populate _consistently_ with human-readable bindings, both in handlebars and in JS
[...]

PR #15091, opened 2024-11-28
fix/loop context lost when max iterations are reached. Add/test cases…

imnaveenk:
## Description
Max iterations in loop step is being considered as error, but this should be a success case and outputs should be available to subsequent steps

## Addresses
-  [https://github.com/Budibase/budibase/issues/15049](https://github.com/Budibase/budibase/issues/15049)

## App Export
- [automation-export-1732820062011.tar.gz](https://github.com/user-attachments/files/17951526/automation-export-1732820062011.tar.gz)
-  Run automation with/without this PR's changes 

## Launchcontrol

Fix : Consider max iterations as success and run subsequent steps normally


imnaveenk:

packages/server/src/threads/automation.ts
```diff
@@ -480,7 +470,9 @@ class Orchestrator {
               success: true,
             }
           : {
-              success: true,
+              success: reachedMaxIterations
+                ? AutomationStepStatus.MAX_ITERATIONS
+                : true,
               items: this.loopStepOutputs,
```
PClmnt, line 476: @imnaveenk What I think we should here do is keep `success` to be a boolean. And if `reachedMaxIterations` is true then conditionally add `status: AutomationStepStatus.MAX_ITERATIONS `

PClmnt:
@imnaveenk This is great, thanks so much for getting this fix in and and taking the time to write some tests to cover it! We really appreciate it. I just left one comment

packages/server/src/threads/automation.ts
imnaveenk: @PClmnt resolved. I have added no iteration in condition as well this is debatable as for max_iterations=0 this case would arise. I feel its better to return no_iterations in that case.

packages/server/src/automations/tests/scenarios/looping.spec.ts
```diff
@@ -152,6 +152,44 @@ describe("Loop automations", () => {
...
+      .loop({
+        option: LoopStepType.ARRAY,
+        binding: ["test", "test2", "test3"],
+        iterations: 2,
+      })
+      .serverLog({ text: "{{loop.currentItem}}" })
+      .serverLog({ text: "{{steps.1.iterations}}" })
+      .run()
+
+    expect(results.steps[1].outputs.iterations).toBe(2)
```
PClmnt, line 171: So, just a little weirdness with our automation system here that you've fallen victim to. This should actually be

`expect(results.steps[0].outputs.iterations).toBe(2)`

As we don't count a loop as a _real_ step. But within the handlebars that's correct as the trigger is counted. I will document this somewhere in the test builder to avoid this confusion, my fault!

packages/server/src/automations/tests/scenarios/looping.spec.ts
```diff
@@ -152,6 +152,44 @@ describe("Loop automations", () => {
...
+      .loop({
+        option: LoopStepType.ARRAY,
+        binding: ["test", "test2", "test3"],
+        iterations: 2,
+      })
+      .serverLog({ text: "{{loop.currentItem}}" })
+      .serverLog({ text: "{{steps.1.iterations}}" })
+      .run()
+
+    expect(results.steps[2].outputs.message).toContain("- 2")
```
PClmnt, line 190: Same here:

`expect(results.steps[1].outputs.message).toContain("- 2")
`

PClmnt (commented):
Just a couple of comments based on test failures also.

packages/server/src/automations/tests/scenarios/looping.spec.ts
imnaveenk: Understood. Fixed both.

PClmnt (approved):

PR #15092, opened 2024-11-29
Don't fetch account for identifyTenantGroup posthog event

melohagan:
## Description
Fixes a potential race condition during onboarding
 
## Addresses
- https://linear.app/budibase/issue/GRO-839/solve-the-problem-with-tenant-creation-account-not-found


packages/backend-core/src/events/identification.ts
```diff
@@ -121,34 +121,20 @@ const identifyInstallationGroup = async (
...
-  let companySize: string | undefined
-
-  if (account) {
-    profession = account.profession
-    companySize = account.size
-    hosting = account.hosting
-  } else {
-    hosting = getHostingFromEnv()
-  }
+  const hosting = Hosting.CLOUD
```
jvcalderon, line 130: There's no possibility to be self?

melohagan: Updated to support self-host

jvcalderon (approved):
LGTM

PR #15096, opened 2024-11-29
Convert frontend-core API client to Typescript

aptkingston:
## Description
This PR contains the full conversion of the frontend-core API client to typescript.

Along with the TS conversion, certain endpoints have had their arguments flattened and the calling code updated accordingly.

My usage of `type` vs `interface` is pretty much arbitrary so I welcome any recommendations or changes to that, as well as the location of types.

All new typescript code is contained in `frontend-core/src/api`. Every other change is just to accommodate function signature changes, which are a combination of flattening params and rearranging things slightly to suit types better.


mike12345567:
> Ah this is now failing due to the new server endpoint typing changes. I'll mark this as a draft again until that's completed and I've gone through and updated all these endpoints again.

Sorry - I didn't mean to slow this work - the last PR on the app service side will be merged today!

aptkingston:
> > Ah this is now failing due to the new server endpoint typing changes. I'll mark this as a draft again until that's completed and I've gone through and updated all these endpoints again.
> 
> Sorry - I didn't mean to slow this work - the last PR on the app service side will be merged today!

All good don't worry! I need to go back over all the endpoints again and see what has changed but at least it's faster than first time round.

packages/server/nodemon.json
```diff
@@ -1,13 +1,19 @@
...
-    "../backend-core",
-    "../pro",
-    "../types",
-    "../shared-core",
-    "../string-templates"
+    "../backend-core/src",
+    "../pro/src",
+    "../types/src",
+    "../shared-core/src",
+    "../string-templates/src"
```
aptkingston, line 8: Specifying only the `src` directories appears to fix the intermittent issues of endless restarts via nodemon (due to a loop of nodemon > build > something changes > nodemon restart > build... etc).

packages/builder/src/components/backend/DataTable/buttons/ExportButton.svelte
```diff
@@ -56,28 +56,19 @@
...
-      },
-    }
+    let payload = {}
     if (selectedRows?.length) {
       payload.rows = selectedRows.map(row => row._id)
     }
     if (sorting) {
-      payload.search.sort = sorting.sortColumn
-      payload.search.sortOrder = sorting.sortOrder
+      payload.sort = sorting.sortColumn
```
adrinr, line 68: ❤️
 I guess this was broken? Or were we supporting both formats?

packages/builder/src/stores/portal/plugins.ts
```diff
@@ -1,17 +1,13 @@
...
-interface Plugin {
-  _id: string
-}
-
 export function createPluginsStore() {
   const { subscribe, set, update } = writable<Plugin[]>([])
 
   async function load() {
-    const plugins = await API.getPlugins()
+    const plugins = (await API.getPlugins()) as Plugin[]
```
adrinr, line 10: Could the API not be typed, instead?

packages/server/nodemon.json
adrinr: Good catch!

adrinr (approved):
LGTM! Good job :)
Just left some minor comments

packages/frontend-core/src/api/index.ts
```diff
@@ -174,29 +151,31 @@ export const createAPIClient = config => {
...
+        } else if (parseResponse) {
           return await parseResponse(response)
         } else {
-          return await response.json()
+          return (await response.json()) as ResponseT
         }
       } catch (error) {
         delete cache[url]
-        return null
+        throw `Failed to parse response: ${error}`
```
adrinr, line 170: Do you know the reason why we were not throwing, before?

packages/frontend-core/src/api/types.ts
```diff
@@ -0,0 +1,136 @@
...
+  message?: string
+  url?: string
+  method?: HTTPMethod
+  json: any
+  status: number
+  handled: boolean
+  suppressErrors: boolean
+}
+
+export type APIClient = BaseAPIClient &
```
adrinr, line 104: Hectic,  but good typing! :)

packages/frontend-core/src/api/index.ts
aptkingston: I think it was probably just a mistake from before, but I don't honestly know... We need to throw now with TS, since we can't return the proper type when we hit this - but yea I think we probably should have been always throwing, for correctness.

packages/builder/src/stores/portal/plugins.ts
aptkingston: The API actually is typed and returns a `FetchPluginResponse`, which is typed as
```js
export type FetchPluginResponse = Plugin[]
```
so I'm only casting to `Plugin[]` because I'd prefer to store that type, even though there is no difference. It just feels wrong to store `FetchPluginResonse` in the store when it's actually just an array of plugins. I'm happy remove this if you'd prefer - I'd actually appreciate knowing what you think is better because I knew this casting was redundant but it just seemed "nicer".

packages/builder/src/components/backend/DataTable/buttons/ExportButton.svelte
aptkingston: It wasn't broken, but I've just flattened the structure a little to make it simpler. We used to do this logic inside the API client code, but there was no point in wrapping things in a `search` object just to pull them out again later. So the result sent to the API is identical but it's a bit simpler to use it now :)

aptkingston:
Thanks for the review @adrinr! This was a monster 😅

packages/builder/src/pages/builder/app/[application]/settings/backups/index.svelte
```diff
@@ -149,24 +144,14 @@
...
+      await backups.deleteBackup($appStore.appId, detail.backupId)
       await fetchBackups(filterOpt, page)
     } else if (detail.type === "backupRestore") {
-      await backups.restoreBackup({
-        appId: $appStore.appId,
-        backupId: detail.backupId,
-        name: detail.restoreBackupName,
-      })
-      await fetchBackups(filterOpt, page)
-    } else if (detail.type === "backupUpdate") {
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/settings/backups/index.svelte:164]: Is the backup update no longer required?

packages/builder/src/stores/portal/plugins.ts
mike12345567: I think my preference is to do something like `const plugins: Plugin[] = await API.getPlugins()` in these type of scenarios, this isn't casting in this case as Typescript will infer whether the types overlap enough for it to work.

Using `as Plugin[]` is more dangerous, because you are forcing the conversion - there is some protections still if the two types don't overlap at all/it makes no sense, but you'll lose protections around things like undefined checks.

packages/builder/src/stores/portal/users.js
```diff
@@ -35,7 +35,24 @@ export function createUsersStore() {
   }
 
   async function invite(payload) {
-    return API.inviteUsers(payload)
+    const users = payload.map(user => {
```
mike12345567, line 38: I assume all this extra work is to make sure that the user fits the API typing?

packages/builder/src/stores/portal/plugins.ts
aptkingston: Sounds good! I'll go for that approach instead of the `as xxx`. Thank you 🙏

packages/builder/src/pages/builder/app/[application]/settings/backups/index.svelte
Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/settings/backups/index.svelte:164]: Nope I believe we actually stopped using and allowing custom naming of backups a while ago, so this code existed but was never used. So just pruning dead code with this.

packages/frontend-core/src/api/rowActions.ts
```diff
@@ -1,50 +1,68 @@
...
     })
   },
 
-  /**
-   * Updates a row action.
-   * @param name the new name of the row action
-   * @param tableId the ID of the table
-   * @param rowActionId the ID of the row action to update
-   */
-  update: async ({ tableId, rowActionId, name }) => {
```
mike12345567, line 33: Was the update no longer needed?

packages/server/nodemon.json
mike12345567: Amazing!

mike12345567 (approved):
LGTM - massive improvement here, I can see how this will populate well to the stores!

packages/frontend-core/src/api/rowActions.ts
aptkingston: Yep this was also dead code! Row actions are now renamed by renaming the automation, using the normal automation endpoint, rather than via this one.

PR #15100, opened 2024-12-02
API typing - app service A -> D

mike12345567:
## Description
This PR contains the typing for A -> D APIs in the app service - this includes automations and applications.


packages/types/src/api/web/app/user.ts
```diff
@@ -7,3 +7,5 @@ export interface SetFlagRequest {
   flag: string
   value: any
 }
+
+export type AppSelfResponse = ContextUserMetadata | {}
```
adrinr, line 11: Why this empty body?

packages/types/src/api/web/deployment.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface PublishAppResponse extends DeploymentDoc {}
+
+export type DeploymentProgressResponse =
+  | {
+      _id: string
+      appId: string
+      status?: DeploymentStatus
+      updatedAt: number
+    }
+  | undefined
```
adrinr, line 12: Should undefined be part of the usage, instead of the type?

packages/server/src/api/controllers/auth.ts
```diff
@@ -45,9 +45,9 @@ export async function fetchSelf(ctx: UserCtx) {
     try {
       const userTable = await sdk.tables.getTable(InternalTables.USER_METADATA)
       // specifically needs to make sure is enriched
-      ctx.body = await outputProcessing(userTable, user as Row)
+      ctx.body = (await outputProcessing(userTable, user as Row)) as ContextUser
```
adrinr, line 48: If we remove the `as Row`, is the `as ContextUser` required?

adrinr (commented):
It's so nice seeing this!

packages/types/src/api/web/app/user.ts
mike12345567: This is a possible response - I didn't want to change any functionality, just document what it actually does!

packages/types/src/api/web/deployment.ts
mike12345567: In this case - this API isn't used today, so I'll make a small change so the API can't return undefined, but I would say if the API can return nothing/undefined then its worth capturing in the types so the frontend knows.

packages/server/src/api/controllers/auth.ts
mike12345567: Silly - should have checked, didn't realise `outputProcessing` can handle context user directly!

adrinr (approved):
🚀

PR #15106, opened 2024-12-03
Fix MySQL options column imports that have commas or other type names in them.

samwho:
## Description

We had a user reporting that their enum columns weren't being imported correctly. On investigation, I found 2 separate bugs with enum imports in MySQL:

- We weren't correctly parsing commas in enum fields, so enum variants containing commas could break it.
- We were incorrectly checking enum variants to see if they were type names, confusing the heuristics that determine what Budibase type to assign to a field.

This PR fixes both and adds tests to prevent regression.

## Launchcontrol

- Fixes some bugs around importing tables that have enum column where the enum variants contain commas or type names.


mike12345567 (approved):
LGTM! Thanks for quick fix

Issue #15113, opened 2024-12-03
Saving an empty snippet leads to the builder crashing


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1
  
**Describe the bug**

Snippets are a premium feature. Steps to reproduce assume that you're using a premium instance
**To Reproduce**
1. In a screen, add a headline binding
2. Open the bindings drawer and switch to JS
3. Click the Snippets icon and add a new snippet
4. Save  an empty snippet
![image](https://github.com/user-attachments/assets/9ab4d78e-208e-4632-97ea-c716103848d0)
5. Try to add it to the input box in the bindings drawer by clicking on it
6. _builder becomes unresponsive_.

**Expected behavior**
empty snippets should not break the builder. 

**Screenshots**
Jam https://jam.dev/c/81a6d8b8-0e21-4b06-b185-30f94e403005

For no obvious reason, Jam didn't capture this error at 21 seconds (when the snippet is clicked from the list in the bindings menu)
![Screenshot 2024-12-03 at 17 58 20](https://github.com/user-attachments/assets/d49f60eb-bd6b-4e31-9b7a-cd1b12906a7b)

  
**App Export**
[Snippets-export-1733248847467.tar.gz](https://github.com/user-attachments/files/17997134/Snippets-export-1733248847467.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - Version 131.0.6778.86 (Official Build) (arm64)

**Additional context**
Issue raised by CSE on behalf of customer.

As part of the initial inquiry the customer also stated that snippets that they had written were _gone_, which is how they discovered that empty snippets cause the builder to crash.

mike12345567:
Fixed in: https://github.com/Budibase/budibase/pull/15114

PR #15114, opened 2024-12-03
prevent empty snippets being saved, don't crash for old empty snippets

shogunpurple:
## Description
Fixes some issues with snippets:
- Prevents saving of empty snippets
- Prevents the builder crashing from empty snippets saved before this
- Fixes name validation - it was causing an error when you tried to edit an existing snippet because the name existed

## Addresses
- https://github.com/Budibase/budibase/issues/15113

## Launchcontrol

Fix issues with snippets crashing builder UI


PClmnt (approved):
LGTM!

PR #15118, opened 2024-12-04
3.0 metrics

shogunpurple:
## Description
This PR adds telemetry for the 3.0 launch features for features that didn't have them

- [x] - row actions
- [x] - views, calculation views and view joins
- [x] - AI column
- [x] - grouped filtering
- [x] - Default values

Remove view v1 events - these weren't being tracked (for updates), weren't seeing much use in general, and are now legacy.

## Addresses
- https://linear.app/budibase/issue/BUDI-8880/30-launch-metrics

## Launchcontrol
Adds telemetry for 3.0 features


packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -160,6 +164,57 @@ export async function create(ctx: Ctx<CreateViewRequest, CreateViewResponse>) {
...
+      await events.view.filterUpdated(properties)
+    }
+  }
+}
+
+async function handleViewEvents(existingView: ViewV2, view: ViewV2) {
+  // Grouped filters
+  if (view.queryUI?.groups) {
+    await handleViewFilterEvents(existingView, view)
+  }
```
samwho, line 186: This isn't a reliable indicator that grouped filters have been used. For backwards compatibility, we populate this `view.queryUI` field from `view.query` when `view.queryUI` hasn't been set: https://github.com/Budibase/budibase/blob/6865052ef0a64e8d5f43299d5d5f98ff8d829ce6/packages/server/src/sdk/app/views/utils.ts#L10-L32

In practice, this field will ~always be set.

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -160,6 +164,57 @@ export async function create(ctx: Ctx<CreateViewRequest, CreateViewResponse>) {
...
+
+  // if new columns in the view
+  for (const key in view.schema) {
+    if (!existingView?.schema?.[key]) {
+      // view calculations
+      // @ts-expect-error non calculation types just won't have the calculationType field
```
samwho, line 192: Prefer a check here over disabling TS, e.g.:

```typescript
if (!("calculationType" in view.schema[key])) {
```

packages/server/src/api/controllers/view/viewsV2.ts
```diff
@@ -160,6 +164,57 @@ export async function create(ctx: Ctx<CreateViewRequest, CreateViewResponse>) {
...
+        await events.view.calculationCreated({
+          calculationType,
+          tableId: view.tableId,
+        })
+      }
+    }
+
+    // view joins
+    if (view.schema[key].columns) {
+      for (const column in view.schema[key]?.columns) {
```
samwho, line 204: Can get rid of a level of nesting here if you want:

```typescript
for (const column in view.schema[key]?.columns ?? [])
```

packages/server/src/api/routes/tests/view.spec.ts
```diff
@@ -73,25 +73,12 @@ describe("/views", () => {
...
-      expect(events.view.created).toHaveBeenCalledTimes(1)
-    })
-
     it("creates a view with a calculation", async () => {
       jest.clearAllMocks()
 
       const view = await saveView({ calculation: ViewCalculation.COUNT })
 
       expect(view.tableId).toBe(table._id)
-      expect(events.view.created).toHaveBeenCalledTimes(1)
```
samwho, line 87: How come all the deletions here? (not that I'm not pleased to see mocks getting deleted)

packages/backend-core/src/events/publishers/table.ts
```diff
@@ -20,14 +21,34 @@ async function created(table: Table, timestamp?: string | number) {
...
-async function updated(table: Table) {
+async function updated(oldTable: Table, newTable: Table) {
+  // only publish the event if it has fields we are interested in
+  let defaultValues, aiColumn
+
+  // check that new fields have been added
+  for (const key in newTable.schema) {
+    if (!oldTable.schema[key]) {
+      const newColumn = newTable.schema[key]
+      if ("default" in newColumn) {
```
samwho, line 32: Possibly worth making sure `default` isn't `null` or `undefined` here. I think that's how it gets unset.

packages/server/src/api/routes/tests/view.spec.ts
shogunpurple: These events are no longer tracked for the legacy views - we just track them in v2 now

packages/server/src/api/controllers/view/viewsV2.ts
samwho: As far as I remember, these get run even on fetch. We wanted to make absolutely sure that these are always set, because the frontend chokes without them. I really wouldn't recommend relying on them.

mike12345567 (approved):
LGTM - thanks for the view filter event changes!

PR #15119, opened 2024-12-04
Add automation survey popup

aptkingston:
## Description
This PR adds a popup in the automations section prompting users to complete a survey.

The [primary link](https://t.maze.co/310149185) points to the maze survey.
The [terms link](https://drive.google.com/file/d/12-qk_2F9g5PdbM6wuKoz2KkIyLI-feMX/view?usp=sharing) points to the PDF which I've hosted in Google Drive and made publicly readable.

Visibility is controlled by a simple local storage flag, so it won't reappear after being dismissed unless local storage is cleared, which typically happens infrequently. For a short term popup like this I don't think it's correct to start saving permanent metadata against the user doc.

The popup will be hidden by either clicking the close icon or by clicking the link.

The popup has a slight delay before sliding in from the right, to catch the eye of the user.

https://github.com/user-attachments/assets/364fbb12-39ad-4a1f-bb50-e7ef7f2fc104

## Addresses
- https://linear.app/budibase/issue/PRO-267/feedback-button-for-automations

## Screenshots
image
image

## Launchcontrol
Add automation survey popup.

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/automation/_layout.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/automation/_layout.svelte
@@ -89,6 +89,7 @@
           href="https://t.maze.co/310149185"
           target="_blank"
           rel="noopener noreferrer"
+          on:click={() => surveyDismissed.set(true)}
         >
           Complete our survey on Automations</a
         >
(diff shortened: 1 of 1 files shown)


ben-budibase:
Approved sire

PClmnt (approved):

PR #15120, opened 2024-12-04
Final app service API typing

mike12345567:
## Description
As title, the finishes off the API typing for the app service, in line with the two PRs before it:

1. https://github.com/Budibase/budibase/pull/15100
2. https://github.com/Budibase/budibase/pull/15112

This contains APIs like views, webhooks and tables.


packages/backend-core/src/middleware/auditLog.ts
```diff
@@ -1,6 +1,6 @@
-import { BBContext } from "@budibase/types"
+import { Ctx } from "@budibase/types"
 
-export default async (ctx: BBContext | any, next: any) => {
+export default async (ctx: Ctx, next: any) => {
```
samwho, line 3: Can we not just delete this? It does nothing.

packages/server/src/api/routes/script.ts
```diff
@@ -1,10 +0,0 @@
-import Router from "@koa/router"
```
samwho, line 1: I'm assuming this was unused?

samwho (approved):

packages/backend-core/src/middleware/auditLog.ts
mike12345567: It is a place holder that is overriden by pro - its one of the oddities that we have to deal with due to the closed source submodule.

packages/server/src/api/routes/script.ts
mike12345567: Yeap - theres no API that uses this and it didn't even really function properly/was even tested - it didn't serve a purpose and probably never would!

PR #15123, opened 2024-12-05
Fix schema exporting.

samwho:
## Description

Prior to this PR, schema exporting for our databases that support it (Postgres, MSSQL, MySQL, and MariaDB) resulted in broken schemas most of the time. 

This PR introduces a test that does the following:

1. Create a table through Budibase.
2. Export datasource schema.
3. Delete the created table.
4. Run the exported schema.
5. Import the table back into Budibase.
6. Assert the old table matches the new one.

This was failing variously for all of our databases prior to this PR, but is passing after.

## Launchcontrol

- A variety of fixes to schema exporting for Postgres, MSSQL, MySQL, and MariaDB.


mike12345567 (approved):
Very nice work on this - great idea with the test as well, good way to prove it!

PR #15126, opened 2024-12-05
Update to TypeScript 5.7.2.

samwho:
## Description

For reasons that aren't clear to me, this also required some fiddling with `testcontainers`.

1. I disabled the `ryuk` testcontainers feature. This cleaned up images after tests had finished, but we don't need it because we tend to reuse images or delete them manually. I did this because for whatever reason it was reading from the wrong Docker socket on my local machine all of a sudden, and failing the backend-core tests.
2. I upgraded to `testcontainers` 10.16.0 in an attempt to fix the above. It didn't work, but I figured I'd stick on this version anyway. It allows us to get rid of the code that removes stopped images, because the latest version of `testcontainers` can restart stopped images.

### DIFF
--- a/examples/nextjs-api-sales/package.json
+++ b/examples/nextjs-api-sales/package.json
@@ -22,6 +22,6 @@
     "@types/react": "17.0.39",
     "eslint": "8.10.0",
     "eslint-config-next": "12.1.0",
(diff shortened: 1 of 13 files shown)


adrinr (approved):
LGTM!

PR #15128, opened 2024-12-05
Tidy monorepo setup

adrinr:
## Description
Making a better use of the yarn workspace, cleaning up and removing some unnecessary processes:
1. Removing unnecessary build steps, as we can use the src directly
2. Unifying the tsconfigs and paths, removing plenty of duplicated code
3. Unifying @types/node 

Related:
- https://github.com/Budibase/budibase-pro/pull/345

## Feature branch env
[Feature Branch Link](http://account-fb-labday-monorepo-setup.fb.qa.budibase.net)


samwho (approved):

packages/backend-core/src/environment.ts
```diff
@@ -134,10 +134,10 @@ const environment = {
   BUDIBASE_ENVIRONMENT: process.env.BUDIBASE_ENVIRONMENT,
   JS_BCRYPT: process.env.JS_BCRYPT,
   JWT_SECRET: process.env.JWT_SECRET
-    ? createSecretKey(Buffer.from(process.env.JWT_SECRET))
+    ? createSecretKey(process.env.JWT_SECRET, "utf8")
```
adrinr, line 137: This is a strange case... We are using node 20, and based on their docs, it should accept "buffer" as a parameter. But the types don't allow it. Until now it was working as we were wrongly using node 22 types

PR #15130, opened 2024-12-05
Update eslint

samwho:
## Description

Upgrades all of our eslint-related dependencies, as well as updating us to the new eslint config format.

Unfortunately I wasn't able to get `eslint-plugin-import` working in this new world. I would appreciate someone else taking a stab at it, it would be a shame to use it. For now, I've just gotten things working without it.

Pro PR: https://github.com/Budibase/budibase-pro/pull/344


mike12345567 (approved):
LGTM - good to get closer to recommended!

Issue #15057, opened 2024-11-25
Bug in default values in multi-select picker


aconjaer:
I have a checkbox multi-select picker with 3 options which should be filtering a repeater. I would like all options to be selected by default, but when I do this, it's not filtering the repeater anymore when you uncheck one of the boxes. When I don't use default values, it all works fine. Looks like a bug to me, but maybe I'm missing something. Any help would be appreciated!
[Multi select bug-export-1732529415502.tar.gz](https://github.com/user-attachments/files/17901089/Multi.select.bug-export-1732529415502.tar.gz)

https://github.com/user-attachments/assets/27944869-21ec-49bc-b6df-6db193c67976

https://github.com/user-attachments/assets/2e8a9201-8b4c-43a7-b1b7-e3b1420b9e7b

mikesealey:
I think there are a few more things at play here, so I've filled out a bug report template to add to this issue. In future, please use the template.

**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.2.12
  - App Version: 1

**Describe the bug**
A multi-select picker (checkboxes) is being used to apply filters to a data provider. When setting the default value for the for the multi-select picker filtering breaks.

**To Reproduce**
Steps to reproduce the behavior:
1. In the data section, create a table - "SampleData" with two columns, ID and Label
2. create rows usings pairs of values 
ID 1 Value - A, 
ID 2 - Value B,
ID 3 Label - C, etc
3. In the design section, add a data provider set to fetch SampleData
4. Nest a form inside that contains a Multi Select Picker
5. Set the Multi Select Picker to Checkboxes, sourcing options and values from the Data Provider, Label column = Label, value column = ID 
![image](https://github.com/user-attachments/assets/3085efe3-dbfb-4a14-96ff-a776940d9380)
6. Add anohter data provider to the screen (at the highest level), set to fetch from the SampleData table, containing a repeater that displays the values in the rows of the SampleData table
7. Add a filter to the data provider to only return rows that are checked
![image](https://github.com/user-attachments/assets/c0862520-1067-4212-8f4d-b77b0e673eb6)
8. Set the default value of the Multi Select Picker to be an array [1,2,3] (should show all boxes checked, may need to reload the builder to confirm
9. (Bonus extra step) Add a headline that shows the {{ New Form.Value }} 
10. preview, and check/uncheck

[...]

PR #15136, opened 2024-12-06
Stop coercing all values to strings when using CoreCheckboxGroup

aptkingston:
## Description
Stop coercing all values to strings when using CoreCheckboxGroup.

## Addresses
- https://linear.app/budibase/issue/BUDI-8862/bug-in-default-values-in-multi-select-picker
- https://github.com/Budibase/budibase/issues/15057


deanhannigan (approved):
LGTM!

PR #15140, opened 2024-12-09
handle cases where datasourceName is undefined when sorting datasources

andz-bb:
## Description
Users table doesn't have a datasource name so currently crashes the builder when you create a table.

## Screenshots
image

## Launchcontrol
Fix bug with sorting datasource names


deanhannigan (approved):
LGTM!

melohagan (approved):

PR #15143, opened 2024-12-09
Enable a variety of useful looking lints.

samwho:
## Description

I went through the rules for eslint and typescript-eslint and enabled everything that looked useful.


mike12345567 (approved):
LGTM - some good rules added!

PR #15147, opened 2024-12-09
Disable filter value binding input when no column is selected in filter screen

andz-bb:
## Description
Binding button was enabled for filter value in the filter drawer:

image

This meant that a filter condition could be created without the column selected or the operator. Without an operator being properly set, the app screen would crash.

The PR disables the binding button along with the field itself meaning conditions without an operator can no longer be created.

image

## Addresses
[BUDI-8909](https://linear.app/budibase/issue/BUDI-8909/builder-crashes-when-incomplete-filters-are-saved)

## App Export
[mike tag team-export-1733759728902.tar.gz](https://github.com/user-attachments/files/18064662/mike.tag.team-export-1733759728902.tar.gz)

## Launchcontrol
Prevent invalid filter conditions from being created


deanhannigan (commented):
Hey @andz-bb, this needs to go a little further and allow that state to be recoverable in the builder on load and it should clear invalid filter config on save. I'll take a look at this, this morning and push some additional updates to your branch 👍

aptkingston (approved):
LGTM!

andz-bb:
tested and works well @deanhannigan, thanks so much

Issue #15138, opened 2024-12-08
Postgresql connection error


Zighy:
**Checklist**
- [ x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.2.25

**Describe the bug**
When I create a connection to a PostgreSQL (version 9.5.25 on x86_64-pc-linux-gnu) i get this error : "Error creating datasource: error: function to_regclass(text) does not exist".

**To Reproduce**
1. Create new app
2. Go to 'Data'
3. Click on 'PostgreSQL'
4. Enter connection parameters (host, port, database, user, password, schema)
5. Click "Connect"
6. See error

**Expected behavior**
Te connection to the database should be correctly establisched.

**Screenshots**
[error_budibase](https://github.com/user-attachments/assets/55b77d7b-1772-4e3b-85aa-1f019bfd8206)

  
**App Export**
[budibase-export.tar.gz](https://github.com/user-attachments/files/18050342/budibase-export.tar.gz)

**Desktop (please complete the following information):**
 - OS: Linux Mint
 - Browser: Brave
 - Version: 1.73.91 Chromium: 131.0.6778.85 (official build) (64 bits) (https://brave.com/latest/)

**Additional context**
I used DBeaver to connect on the same PostgreSQL database and executed the following query : SELECT to_regclass('public.users');
The return of the query was "users", the function to_regclass exists.

PR #15155, opened 2024-12-10
Fix for fetching schemas in Postgres 9.5

mike12345567:
## Description
Updating how we list columns from a table in Postgres - to avoid using `to_regclass` which is not available in Postgres 9.5.

This is a supported version of Postgres - updated test case to check it.

## Addresses
- https://github.com/Budibase/budibase/issues/15138
- https://linear.app/budibase/issue/BUDI-8905/postgresql-connection-error


samwho (changes requested):
You're going to need to update the CI GitHub Actions YAML to include `postgres_legacy` as a datasource that gets tested.

packages/server/src/api/routes/tests/datasource.spec.ts
```diff
@@ -590,13 +590,13 @@ if (descriptions.length) {
 }
 
 const datasources = datasourceDescribe({
-  exclude: [DatabaseName.MONGODB, DatabaseName.SQS, DatabaseName.ORACLE],
+  only: [DatabaseName.POSTGRES_LEGACY],
```
adrinr, line 593: Is this expected?

packages/server/src/api/routes/tests/datasource.spec.ts
```diff
@@ -647,6 +647,13 @@ if (datasources.length) {
             // can load it. We're using postgres 16 in tests at the time of writing.
             schema = schema.replace("SET transaction_timeout = 0;", "")
           }
+          if (isPostgres && isLegacy) {
```
adrinr, line 650: Should this not be `!isLegacy`, if it's not supported in older versions?

mike12345567: In this case its removing removing something from the export - rather than adding something.

adrinr (approved):

PR #15161, opened 2024-12-11
Fix client error SVG

aptkingston:
## Description
After moving from rollup to vite we need to import SVG files slightly differently. Currently the error image when you don't have access to an app is broken.

Before:
image

After:
image


deanhannigan (approved):
LGTM!

adrinr (approved):

Issue #15158, opened 2024-12-11
Unexpected rows appear in user-specific views when multiple users create records at the same time


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 3.2.27
  - App Version: 3.2.27

**Describe the bug**
Filters for multi-dev collaboration aren't applied when users are on the same screen at the same time, creating rows. Refreshing resolves this issue.

**To Reproduce**
1. Import the attached app
2. Publish the app
3. Login to 2 separate users
4. Navigate to the app on the 2 user sessions
5. Create a row for one user
6. Switch to the other user and see that the row that is only meant for that user is displayed.

**Expected behavior**
Users shouldn't be able to see data that isn't meant for them.

**Screenshots**
<img width="480" alt="Screenshot 2024-12-11 at 10 47 38" src="https://github.com/user-attachments/assets/e242d4f0-3f0c-4b01-92b6-9387cdbc86d0">
  
**App Export**
[current user-export-1733914081086.tar.gz](https://github.com/user-attachments/files/18094052/current.user-export-1733914081086.tar.gz)

PR #15163, opened 2024-12-11
Prevent syncing row changes between users for views filtered by current user

aptkingston:
## Description
This PR prevents syncing row changes between users when using views that are filtered by any current user property. This is required as it can otherwise expose data which should not be seen by other users. I'm just stringifying and searching the `query` property for the presence of a current user binding as my litmus test. 

Currently the view doc is fetched and checked whenever a user connects to a certain datasource, which is infrequent - typically just once per page load, so I don't think there's much to be gained my computing this on update and storing a flag into the doc, but opinions are welcome.

Also contains a tiny fix for scrollbar styles as they have looked wrong for a few months and it's about time they were fixed!

## Addresses
- https://linear.app/budibase/issue/BUDI-8917/unexpected-rows-appear-in-user-specific-views-when-multiple-users
- https://github.com/Budibase/budibase/issues/15158

### DIFF
--- a/packages/server/src/websockets/grid.ts
+++ b/packages/server/src/websockets/grid.ts
@@ -1,7 +1,7 @@
 import authorized from "../middleware/authorized"
 import currentApp from "../middleware/currentapp"
 import { BaseSocket } from "./websocket"
-import { auth, permissions } from "@budibase/backend-core"
+import { auth, permissions, context } from "@budibase/backend-core"
 import http from "http"
 import Koa from "koa"
 import { getSourceId } from "../api/controllers/row/utils"
@@ -10,6 +10,7 @@ import { Socket } from "socket.io"
 import { GridSocketEvent } from "@budibase/shared-core"
 import { userAgent } from "koa-useragent"
 import { createContext, runMiddlewares } from "./middleware"
+import sdk from "../sdk"
 
 const { PermissionType, PermissionLevel } = permissions
 
@@ -24,9 +25,26 @@ export default class GridSocket extends BaseSocket {
       const ds = payload.datasource
       const appId = payload.appId
       const resourceId = ds?.type === "table" ? ds?.tableId : ds?.id
+      let valid = true
 
-      // Ignore if no table or app specified
+      // Validate datasource
       if (!resourceId || !appId) {
+        // Ignore if no table or app specified
+        valid = false
+      } else if (ds.type === "viewV2") {
+        // If this is a view filtered by current user, don't sync changes
+        try {
+          await context.doInAppContext(appId, async () => {
+            const view = await sdk.views.get(ds.id)
+            if (JSON.stringify(view.query).includes("[user]")) {
+              valid = false
+            }
+          })
+        } catch (err) {
+          valid = false
+        }
+      }
+      if (!valid) {
         socket.disconnect(true)
         return
       }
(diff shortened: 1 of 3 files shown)


packages/server/src/websockets/grid.ts
```diff
@@ -18,15 +24,46 @@ export default class GridSocket extends BaseSocket {
...
+    return findHBSBlocks(JSON.stringify(view.query))
+      .map(binding => {
+        const sanitizedBinding = binding.replace(/\\"/g, '"')
+        if (isJSBinding(sanitizedBinding)) {
+          return decodeJSBinding(sanitizedBinding)
+        } else {
+          return sanitizedBinding
+        }
+      })
+      .some(binding => binding?.includes("[user]"))
```
adrinr, line 38: Will this always be the only true case? Could we not get other issues such as formulas or similar?

packages/server/src/websockets/grid.ts
```diff
@@ -18,15 +24,46 @@ export default class GridSocket extends BaseSocket {
...
+            const view = await sdk.views.get(ds.id)
+            if (this.containsCurrentUserBinding(view)) {
+              valid = false
+            }
+          })
+        } catch (err) {
+          valid = false
+        }
+      }
+      if (!valid) {
```
adrinr, line 66: More of a product discussion, but are we exposing this somehow to the user? Otherwise they might think it's actually broken

adrinr (changes requested):
It looks like the issue is not related to the user filter, but on any filter. Create a different view that, for example, filters by "title starting with a". The same issue will happen, as the grid socket will push the row ignoring any filter.

https://github.com/user-attachments/assets/f157c3f6-0013-49d2-9ecc-8b39daac7c5b

aptkingston:
> It looks like the issue is not related to the user filter, but on any filter. Create a different view that, for example, filters by "title starting with a". The same issue will happen, as the grid socket will push the row ignoring any filter.
> 
>  Screen.Recording.2024-12-12.at.10.40.37.mov

The same thing does happen with all filters yes, but it's less significant for other filters. We are prioritising a quick solution specifically for user-segregated data as that has larger security implications compared to just seeing rows which don't match the currently applied filters.

I agree we need a wider solution for all filters, but it's not an easy fix. It will probably require evaluating all filters server side against rows before broadcasting them down the websocket.

packages/server/src/websockets/grid.ts
aptkingston: For this fix we only care about data segregated per user. There's a wider problem of data not matching filters in general, but by identifying user bindings we can fix cases where users are seeing row they should not be able to access. Only the current user bindings create this security risk. This PR is deliberately only targetting user bindings and is not meant to be a wider fix.

packages/server/src/websockets/grid.ts
aptkingston: We aren't, no. But I think that's ok, as you wouldn't expect to be able to see rows created by other users if using a view that's filtered to show only your own rows.

aptkingston:
I think we'll need to release this in the meantime, because the other solution is a much larger change. Right now you just subscribe to a datasource and that's the identifier for changes. We would have to refactor things so that users resubscribe when their query changes (their local frontend query which can be attached to any DS+).

I think the flow would look something like:
- Clients resubscribe every time they are searching using a new query, in the case of using a table or appending additional filters to a view
- Serverside we store every connection's query
- Serverside we fetch and cache the current user doc for every connection
- Whenever a document change is detected, for each connection we use the current user doc, view built-in query (if a view), and any additional frontend filters to create a single enriched query. If this is a view with a built in query and the user also has an additional frontend query, these must be joined together, then the whole thing is enriched with the current user
- Evaluate that query against the doc
- Broadcast the change to users which the query matches against

It's pretty significant and there is some time pressure to release this one, so I think we should proceed with this for now.

adrinr (approved):
Good to go with this solution temporally

PR #15166, opened 2024-12-12
Custom role in groups issue

adrinr:
## Description
Fixing an issue with custom roles. If we have a single custom role, it has role order 0, but the check it not contemplating this properly. This PR just adds a test catching the error and fixes the issue.
Also, fixing the client dev build to not be minimised, as I had some struggles while trying to find the origin of the issue

Related:
- https://github.com/Budibase/budibase-pro/pull/346

## Launchcontrol
Fixing an issue with custom roles in groups


mike12345567 (approved):
LGTM!

aptkingston (approved):

PR #15169, opened 2024-12-12
Only select required columns from sql databases

adrinr:
## Description
Edit the SQL generation (for both external and sqs) in order to not include `select *`. Instead, pull only the required fields. This applies both on the requested table rows plus the linked relationships. This change will increase performance, not pulling non-needed data.

There is an exception for this when there is a formula field on the requested fields. Given that this formula might use other columns, we need to pull it all in order to compute them.

## Launchcontrol
Smarter SQL select, requesting only the required data.


packages/backend-core/src/sql/sql.ts
```diff
@@ -292,11 +281,9 @@ class InternalBuilder {
 
     const alias = this.getTableName(table)
     const schema = this.table.schema
-    if (!this.isFullSelectStatementRequired()) {
```
adrinr, line 295: The logic for it has been moved to `sqlUtils`

packages/server/src/api/controllers/row/utils/sqlUtils.ts
```diff
@@ -118,45 +119,131 @@ export async function buildSqlFieldList(
...
+    const requiredFields: string[] = []
+    if (table.primary) {
+      requiredFields.push(...table.primary)
+    }
+    if (table.primaryDisplay) {
+      requiredFields.push(table.primaryDisplay)
+    }
+
+    if (!sql.utils.isExternalTable(table)) {
+      requiredFields.push(...PROTECTED_INTERNAL_COLUMNS)
```
mike12345567, line 145: This may lead to some duplication as `.primary` is likely to be set to `_id` for internal tables - but I don't think thats a problem here.

mike12345567 (approved):
LGTM - the `sqlQueryJson` test cases are a bit of a nightmare to maintain - if they are being problematic we could consider getting rid of them - they circumvent the API which makes them a lot harder to keep up to date.

packages/server/src/api/controllers/row/utils/sqlUtils.ts
adrinr: For this we use a set for the return, to avoid having duplicates

packages/backend-core/src/sql/sql.ts
```diff
@@ -311,34 +298,35 @@ class InternalBuilder {
...
+        // HH:mm format
 
-        if (table) {
-          return this.rawQuotedIdentifier(`${table}.${column}`)
-        } else {
-          return this.rawQuotedIdentifier(field)
-        }
-      })
+        // TODO: figure out how to express this safely without string
+        // interpolation.
```
samwho, line 317: Oh sorry this TODO is outdated. I believe what exists there now is safe.

samwho (approved):

Issue #15139, opened 2024-12-08
Cannot read properties of undefined (reading 'schema')


JBSDS:
Hello everyone,

I'm facing an issue since I updated Budibase: I upgraded from version 2.32.11 to the latest available version (above 3.0 ). After this update, I started encountering the following error whenever I use filters that reference "relationships" in the database :

Cannot read properties of undefined (reading 'schema').

For context, I am using MySQL as my database.

I tested a downgrade to version 2.32.11 , and everything works fine in this version. However, my goal is to move to the latest version to take advantage of the new features and improvements. 🙄

Has anyone else encountered this issue ? Or does anyone have ideas on how to diagnose/fix this problem ? Any help or guidance would be greatly appreciated ! 🙏

Thank you in advance for your feedback

```
{
  "browser": {
    "language": "fr-FR",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36",
    "platform": "Win32",
    "vendor": "Google Inc."
  },
  "server": {
    "budibaseVersion": "3.2.25",
    "hosting": "docker-compose",
    "nodeVersion": "v20.18.1",
    "platform": "linux",
    "cpuArch": "x64",
    "cpuCores": 32,
    "cpuInfo": "13th Gen Intel(R) Core(TM) i9-13900K",
    "totalMemory": "31.22834014892578GB",
    "uptime": "0 day(s), 7 hour(s), 36 minute(s)"
  }
}

```
![Capture d’écran 2024-12-08 125700](https://github.com/user-attachments/assets/9c290c49-589e-4d9a-877f-e279bace26d4)

WoisWoi:
I'm also having this problem.

mike12345567:
Hi @WoisWoi and @JBSDS - would it be possible to provide a schema for a database which exhibits this behaviour? You can export this using the following functionality in Budibase: https://docs.budibase.com/docs/troubleshooting-sql#sql-schema-export

JBSDS:
Hi @mike12345567, here is the database schema.
[MySQL-dump (1).zip](https://github.com/user-attachments/files/18083675/MySQL-dump.1.zip)

mike12345567:
[...]

PR #15172, opened 2024-12-12
Relationship searching on tables with spaces

mike12345567:
## Description
It is possible to search for rows using the table name - this is something that data providers will often do - but if the external table name contained a space there could be an issue with the table name not lining up after it comes in through the API (in the body of the request).

Added a test case for this as well, only applies to external tables as the internal table name is not used for searching.

Addresses: https://github.com/Budibase/budibase/issues/15139

### DIFF
--- a/packages/backend-core/src/sql/utils.ts
+++ b/packages/backend-core/src/sql/utils.ts
@@ -66,6 +66,14 @@ export function buildExternalTableId(datasourceId: string, tableName: string) {
   return `${datasourceId}${DOUBLE_SEPARATOR}${tableName}`
 }
 
+export function checkTableId(tableId: string) {
+  if (isExternalTableID(tableId) && tableId.includes(" ")) {
+    return encodeURIComponent(tableId)
+  } else {
+    return tableId
+  }
+}
+
 export function breakExternalTableId(tableId: string) {
   const parts = tableId.split(DOUBLE_SEPARATOR)
   let datasourceId = parts.shift()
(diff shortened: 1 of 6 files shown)


packages/backend-core/src/sql/utils.ts
```diff
@@ -66,6 +66,14 @@ export function buildExternalTableId(datasourceId: string, tableName: string) {
   return `${datasourceId}${DOUBLE_SEPARATOR}${tableName}`
 }
 
+export function checkTableId(tableId: string) {
```
adrinr, line 69: Should this not be encoding always? It will encode only if it is actually needed

samwho (approved):

packages/backend-core/src/sql/utils.ts
adrinr, line 69: Also, this function name is confusing. It does not check the table id, it encodes it

packages/server/src/api/controllers/row/utils/utils.ts
```diff
@@ -65,19 +65,21 @@ export function getSourceId(ctx: Ctx): { tableId: string; viewId?: string } {
     const { sourceId } = ctx.params
     if (docIds.isViewId(sourceId)) {
       return {
-        tableId: utils.extractViewInfoFromID(sourceId).tableId,
+        tableId: sql.utils.checkTableId(
+          utils.extractViewInfoFromID(sourceId).tableId
```
adrinr, line 69: Should the `checkTableId` not be used internally on `extractViewInfoFromID`?

packages/backend-core/src/sql/utils.ts
mike12345567: true - I can adjust them both - I was just replicating how it builds the table IDs today!

adrinr (approved):

PR #15174, opened 2024-12-12
Convert portal apps store to typescript

aptkingston:
## Description
Convert portal apps store to typescript.

### DIFF
--- a/packages/builder/src/stores/portal/apps.ts
+++ b/packages/builder/src/stores/portal/apps.ts
@@ -1,19 +1,39 @@
 import { derived } from "svelte/store"
+// @ts-ignore
 import { AppStatus } from "constants"
 import { API } from "api"
 import { auth } from "./auth"
-import BudiStore from "../BudiStore" // move this
+import BudiStore from "../BudiStore"
+import { App, UpdateAppRequest, User } from "@budibase/types"
 
-// properties that should always come from the dev app, not the deployed
-const DEV_PROPS = ["updatedBy", "updatedAt"]
-
-export const INITIAL_APPS_STATE = {
-  apps: [],
+interface AppIdentifierMetadata {
+  devId?: string
+  devRev?: string
+  prodId?: string
+  prodRev?: string
 }
 
-export class AppsStore extends BudiStore {
+interface AppUIMetadata {
+  deployed: boolean
+  lockedYou: boolean
+  lockedOther: boolean
+  favourite: boolean
+}
+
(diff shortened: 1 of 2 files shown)


packages/builder/src/stores/portal/apps.ts
```diff
@@ -1,19 +1,39 @@
 import { derived } from "svelte/store"
+// @ts-ignore
```
aptkingston, line 2: Needed because the tsconfig refuses to properly map "constants" to "src/constants/index.js" even if specifying an explicit path. It instead prefers to resolve this to some constants module from node_modules and throws an error saying AppStatus does not exist.

mike12345567: On this one - I had a look into this, it can be aliased, but it appears we just need to update the constants alias to `@constants` - I'm not sure where the overlap it coming from exactly, but splitting it feels like a better option to disabling Typescript completely.

mike12345567: After a discussion - going to approve this PR with the ignore, but we're going to update all aliases to follow the `@` pattern often seen with aliases - this avoids overlap with builtin Node modules which cannot be overridden.

mike12345567 (approved):

packages/builder/src/stores/portal/apps.ts
```diff
@@ -146,22 +174,23 @@ export class AppsStore extends BudiStore {
...
-          lockedOther: app.lockedBy && app.lockedBy.email !== $auth.user?.email,
-          favourite: $auth.user?.appFavourites?.includes(app.appId),
-        }))
-      : []
+    const enrichedApps: EnrichedApp[] = $store.apps.map(app => {
+      const user = $auth.user as User | null
+      return {
+        ...app,
+        deployed: app.status === AppStatus.DEPLOYED,
+        lockedYou: app.lockedBy != null && app.lockedBy.email === user?.email,
```
adrinr, line 189: Why is this a simple `!= null` check instead of using !!app.lockedBy

packages/builder/src/stores/portal/apps.ts
adrinr: I found this issues to... 🤷🏽

packages/builder/src/stores/portal/apps.ts
aptkingston: No reason! It's an object so both are functionally equivalent - I just tend to use `!= null` more often for some reason.

PR #15179, opened 2024-12-13
S3Upload UI fixes

deanhannigan:
## Description
A couple of minor fixes for the `S3Upload` component for issues that I came across while testing.

## Addresses
- `Fix` - When making a `put` request to S3, the response body is empty. When the api client encountered this, the call to `reponse.json()` was triggering an error. Instead a call to `.text()` is made and verified. If the response text is empty, an empty object is returned instead.
- If you moved the `S3Upload` component in the builder the progress circle would appear and remain visible
- Minor change to allow binding of the `S3Upload` picker value within its own settings. This is specifically useful if you want to map the name of a file you've added in file picker directly to the `File name` setting.
  - Using the existing notation will now output the file name `{{ My File Upload.Value.0.name  }}`
  - This will output the browser mimetype `{{ My File Upload.Value.0.type  }}`

## Screenshots
Original S3 issue. This is now prevented
Screenshot 2024-12-13 at 09 02 02

## Launchcontrol
Minor fix for S3Upload component and added nested prop binding.

## Feature branch env
[Feature Branch Link](http://account-fb-s3-upload-fixes.fb.qa.budibase.net)


packages/client/src/components/app/forms/S3Upload.svelte
```diff
@@ -78,12 +83,15 @@
   }
 
   const upload = async () => {
+    const processedFileKey = processStringSync(key, {
+      [$component.id]: fieldState,
```
aptkingston, line 87: This means the only context available is itself, but we need to add the normal context in here too don't we? Since you have access to all normal bindings when configuring this setting in the builder.

So we just need to pull in normal context as well:
```js
const context = getContext("context")
...
const processedFileKey = processStringSync(key, {
  ...$context,
  [$component.id]: fieldState,
})
```

aptkingston (changes requested):
LGTM but one small change needed to allow all normal bindings to work.

packages/client/src/components/app/forms/S3Upload.svelte
deanhannigan: Yep, that's correct, the other context elements should be present. `$context` also contains the correct parameter so there's no need for the `[$component.id]` entry now. Cheers!

packages/frontend-core/src/api/index.ts
```diff
@@ -163,7 +163,13 @@ export const createAPIClient = (config: APIClientConfig = {}): APIClient => {
...
           return await parseResponse(response)
         } else {
-          return (await response.json()) as ResponseT
+          const text = await response.text()
+          // If the response has no body at all e.g s3.put
+          if (!text) {
+            // Empty response
+            return {} as ResponseT
+          }
+          return JSON.parse(text) as ResponseT
```
aptkingston, line 172: Rather than handle this here and kind of lie to TS, we could add a `parseResponse` property to the API PUT call for the external upload. That's the proper way of handling a unique type of response. In this case, if we don't care about the response we can add `parseResponse: () => {}` or equivalent. That would go around line 103 in `attachments.ts`. Just think that's a little cleaner than assuming any empty response is a valid `ResponseT`.

deanhannigan: Wayyyyy neater approach. I've reverted the api change and added in a `parseResponse` exclusively for the externalUpload. Thanks!

aptkingston (approved):
LGTM!

Issue #14935, opened 2024-11-01
When creating and testing query in the S3 connector the builder crashes/reloads


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1 AND miketest1qa

**Describe the bug**

I'm trying to set up a Create query - When clicking "Run query" with the following settings specified the query fails _and then the page reloads_ 
![image](https://github.com/user-attachments/assets/3683f6dc-1075-49f7-8880-7c0642cd270e)

**To Reproduce**
1. Create s3 connection
2. start a new query
3. Name - Create // Function - Create // Access - Admin // ACL - private // New Bucket - Bucket1

**Expected behavior**
If query test-run fails, builder should not refresh

**Screenshots**
https://jam.dev/c/2ba4e0bc-43fb-4da2-9581-2435700acd07
  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 130.0.6723.70 (Official Build) (arm64)

**Additional context**
Discovered in QA, also reproduced in Prod

PR #15180, opened 2024-12-13
Fix s3 integration error handling

deanhannigan:
## Description

The default behaviour for handling integration errors is to throw the error and have koa relay it as a `400`.
If an error object thrown by an integration contains a `statusCode`, koa will ignore the `400` and use that instead.

In development, if you encountered a permission restriction with your S3 integration, the AWS client would return a `403` and log you out. In prod, the page you were on would reload and you would lose the error message.

All errors in the S3 integration throw a new instance of `Error` with any relevant messaging. This will then be relayed with the default `400` status.

## Addresses
- https://github.com/Budibase/budibase/issues/14935

## Launchcontrol

S3 integration error handling updated

### DIFF
--- a/packages/server/src/integrations/s3.ts
+++ b/packages/server/src/integrations/s3.ts
@@ -177,8 +177,12 @@ class S3Integration implements IntegrationBase {
     try {
       await this.client.listBuckets().promise()
       response.connected = true
-    } catch (e: any) {
-      response.error = e.message as string
+    } catch (e: unknown) {
+      if (e instanceof Error) {
+        response.error = e.message as string
+      } else {
+        response.error = "AWS S3: Unable to process the request"
+      }
     }
     return response
   }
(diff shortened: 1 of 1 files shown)


aptkingston (approved):
LGTM but will let someone from firestorm review too.

packages/server/src/integrations/s3.ts
```diff
@@ -209,7 +213,20 @@ class S3Integration implements IntegrationBase {
...
+      return response
+    } catch (e: unknown) {
+      let message = "Unable to process the request"
+      if (e instanceof Error) {
+        message = e.message
+      }
+
+      // Do not rethrow AWS errors as the statusCode prop will be
+      // interpreted by koa
+      throw new Error("AWS S3: " + message)
```
samwho, line 228: Could you add `{ cause: e }` to this to make sure we don't lose the stack trace? I think you'd need to do something like this:

```typescript
try {
  // ...
} catch (e: unknown) {
  if (e instanceof Error) {
    throw new Error(e.message, { cause: e })
  }
  throw new Error("AWS S3: Unable to process request")
}
```

Alternately, is it enough to target the problem Koa behaviour directly like this:

```typescript
try {
  // ...
} catch (e: unknown) {
  if ("statusCode" in e && e.statusCode === 403) {
    e.statusCode = 400
  }
  throw e
}
```

Alternately alternately should we even trust 403s from integrations at all? Is there ever a case where we want to log the user out if an integration returns a 403? Should we go right to the source and map an integration 403 to 400 in all cases in whatever relevant bit of Koa is handling that?

PR #15200, opened 2024-12-17
Update DS and integration store to BudiStore

mike12345567:
## Description
Updating datasource and sorted integration store to be classes, that extend the BudiStore base implementation.

Also had to rename the datasource store `update` function to `save` for this to work due to the overlap with the stores update function.

### DIFF
--- a/packages/builder/src/stores/builder/datasources.ts
+++ b/packages/builder/src/stores/builder/datasources.ts
@@ -1,4 +1,4 @@
-import { writable, derived, get } from "svelte/store"
+import { derived, get } from "svelte/store"
 import {
   IntegrationTypes,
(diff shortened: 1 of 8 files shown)


packages/builder/src/stores/builder/datasources.ts
```diff
@@ -41,102 +42,131 @@ class TableImportError extends Error {
...
+
+  select(id: string) {
+    this.store.update(state => ({
       ...state,
       selectedDatasourceId: id,
     }))
   }
 
-  const updateDatasource = (
+  private updateDatasourceInStore(
```
aptkingston, line 139: Oooh I forgot that we can use `private` now... Awesome.

aptkingston (approved):
LGTM!

PR #15201, opened 2024-12-17
Fix patching rows from views with hidden relationships

adrinr:
## Description
Fixing the following scenario:
1. Create an external table with related data
2. Create a view excluding the relations
3. Update a row from that view
4. Go back to the main table, see the relationship being dropped

This PR adds tests catching the issue and fixes to the proper behaviour, where the relationship is not dropped. Also, it makes sure that any field that is not part of the views can be modified.

https://github.com/user-attachments/assets/43b30030-df50-4cd1-8384-6ea409f2f447

## Launchcontrol
Fixing wrongly dropped relationships for trimmed tables


mike12345567 (approved):
LGTM! Nice catch

PR #15205, opened 2024-12-18
Update builder aliases

aptkingston:
## Description
This PR updates all root level alieased imports in the builder packages. Previously we manually aliased all top level imports such as `components`, `stores` and `constants` but we've recently ran into an issue where typescript is not happy about a root level `constants` import.

This PR updates the aliases so that anything imported as `@/foo` is resolved from the top level `src` directory, and also keeps TS happy. This has the added benefit of being able to import anything from its top level path without needing to manually add another alias for it like we used to.

Examples of before and after:
`import { Something } from "constants"` > `import { Something } from "@/constants"`
`import { Component } from "components/common/data"` > `import { Component } from "@/components/common/data"`


PClmnt (approved):

adrinr (approved):
LGTM, I also got similar issues because some base packages. I like the idea of prefixing all of them

PR #15215, opened 2024-12-18
Update AWS SDK to V3

mike12345567:
## Description
For a while now we have needed to migrate from AWS SDK V2 to V3 - this is quite a big change with a lot of changing APIs, but luckily our usage of the AWS SDK is quite small.

I initially followed the guide here: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/migrating.html and the tool it provides to start the migration off, however a lot of it did need to be done by hand as we make heavy use of presigned URLs and this is something that has changed quite a lot in the newer version.

The biggest difference is the fact that pre-signed URL generation is now an async function, previously this was a sync call and a lot of our call stack depended upon that - which is where the majority of the updates were required. I don't believe this is impactful from a performance point of view as the function is still doing the same work, but it is async *incase* the command you are pre-signing requires async work, which none of ours require.

Ignore the very large file difference, the majority of this comes from the `yarn.lock` changes required as AWS SDK V3 is broken down into lots of smaller packages rather than one gigantic one.


adrinr (approved):
Good job! This was not an easy migration :)

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -194,11 +199,12 @@ export async function upload({
...
   }
-  const config: any = {
+  const config: PutObjectCommandInput = {
     // windows file paths need to be converted to forward slashes for s3
+    Bucket: sanitizeBucket(bucketName),
     Key: sanitizeKey(filename),
-    Body: fileBytes,
-    ContentType: contentType,
+    Body: fileBytes as stream.Readable | Buffer,
+    ContentType: contentType!,
```
samwho, line 207: nit: is this ! required? how come TypeScript isn't convinced this exists?

packages/backend-core/src/objectStore/buckets/tests/app.spec.ts
samwho, line 1: How come a bunch of tests are being removed here?

packages/backend-core/src/objectStore/buckets/tests/global.spec.ts
samwho, line 1: Same here, how come we're getting rid of minio/S3 tests?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -284,35 +297,42 @@ export async function streamUpload({
...
+  if (!response.Body) {
+    throw new Error("Unable to retrieve object")
+  }
   // currently these are all strings
   if (STRING_CONTENT_TYPES.includes(response.ContentType)) {
-    return response.Body.toString("utf8")
+    return response.Body.toString()
   } else {
-    return response.Body
+    return response.Body as stream.Readable
```
samwho, line 317: How come this cast is required?

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -284,35 +297,42 @@ export async function streamUpload({
...
   }
-  const response: any = await objectStore.getObject(params).promise()
+  const response = await objectStore.getObject(params)
+  if (!response.Body) {
+    throw new Error("Unable to retrieve object")
+  }
   // currently these are all strings
   if (STRING_CONTENT_TYPES.includes(response.ContentType)) {
-    return response.Body.toString("utf8")
+    return response.Body.toString()
```
samwho, line 315: Do we no longer need to pass an encoding here?

samwho (approved):

packages/backend-core/src/objectStore/objectStore.ts
mike12345567: yeap - this `toString` has been updated, it always casts to utf8 now, there is no parameter.

packages/backend-core/src/objectStore/objectStore.ts
mike12345567: Sadly this is due to the typing in SDK V3 - it thinks it could be a node or browser response, which confuses it quite a bit, I'll see if I can cast it in a slightly better way!

packages/backend-core/src/objectStore/objectStore.ts
mike12345567: didn't need this!

packages/backend-core/src/objectStore/buckets/tests/global.spec.ts
mike12345567: Thanks for noticing these, I removed them when testing, forgot to put them back!

PR #15220, opened 2024-12-19
Typing the table store

adrinr:
## Description
Type table store

### DIFF
--- a/packages/builder/src/stores/builder/tables.ts
+++ b/packages/builder/src/stores/builder/tables.ts
@@ -141,9 +141,6 @@ export class TableStore extends DerivedBudiStore<
   }
 
   async delete(table: { _id: string; _rev: string }) {
-    if (!table?._id) {
-      return
-    }
     await API.deleteTable(table._id, table._rev)
     this.replaceTable(table._id, null)
   }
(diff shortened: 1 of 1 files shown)


packages/builder/src/stores/builder/tables.ts
```diff
@@ -99,28 +133,33 @@ export function createTablesStore() {
...
+      await this.fetch()
     } else {
-      await Promise.all(tableIdsToFetch.map(id => singleFetch(id)))
+      await Promise.all(tableIdsToFetch.map(id => this.singleFetch(id)))
     }
     return savedTable
   }
 
-  const deleteTable = async table => {
+  async delete(table: { _id: string; _rev: any }) {
```
mike12345567, line 143: Why does `_rev` need to be any here - is `_rev?: string` not sufficient?

packages/builder/src/stores/builder/tables.ts
```diff
@@ -198,26 +237,12 @@ export function createTablesStore() {
     }
   }
 
-  const removeDatasourceTables = datasourceId => {
-    store.update(state => ({
+  removeDatasourceTables(datasourceId: any) {
```
mike12345567, line 240: can datasource ID not be specified as a string?

packages/builder/src/stores/builder/tables.ts
adrinr: Yup, mistake. Done. It should always be required, and it looks like there is a single usage were we are passing the full table

mike12345567 (approved):
Nice job!

PR #15221, opened 2024-12-19
allow users to select firestore databases other than the default

andz-bb:
## Description
Currently the firestore connector can only use the `(default)` firestore DB. In cases where a user has more than one db, this allows them to specifically connect to one of them by id. The (default) is added by default. For visibility in cases where a connection setup already exists, the field now shows with `(default)` as a placeholder in the datasource config fields. To allow this, I have had to add support for configurable placeholders on each relevant config field type.

## Screenshots
_before_
image

_after_
image

## Launchcontrol
add support for database selection in firestore connector

### DIFF
--- a/packages/server/src/integrations/firebase.ts
+++ b/packages/server/src/integrations/firebase.ts
@@ -12,6 +12,7 @@ interface FirebaseConfig {
   email: string
   privateKey: string
   projectId: string
+  databaseId?: string
 }
 
 const SCHEMA: Integration = {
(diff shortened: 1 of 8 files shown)


shogunpurple (approved):
LGTM, nice work!

packages/server/src/integrations/firebase.ts
```diff
@@ -30,12 +31,21 @@ const SCHEMA: Integration = {
...
     projectId: {
       type: DatasourceFieldType.STRING,
+      display: "Project ID",
       required: true,
     },
+    databaseId: {
+      type: DatasourceFieldType.STRING,
+      display: "Database ID",
+      required: false,
+      default: "(default)",
```
melohagan, line 46: NAB: Should this be `"default"` instead of `"(default)"`

packages/server/src/integrations/firebase.ts
```diff
@@ -97,6 +107,7 @@ class FirebaseIntegration implements IntegrationBase {
     this.config = config
     this.client = new Firestore({
       projectId: config.projectId,
+      databaseId: config.databaseId || "(default)",
```
melohagan, line 110: Same here, should this have brackets?

melohagan (approved):

packages/server/src/integrations/firebase.ts
andz-bb: it seems to be specifically `(default)` in firestore

`default` won't connect but `(default)` and my second db named `second-db` both work

image

PR #15224, opened 2024-12-20
Typing viewV2 store

adrinr:
## Description
Typing viewV2 store


mike12345567 (approved):
LGTM - nice use of type predicates!

PR #15227, opened 2024-12-20
Typing automation store

PClmnt:
## Description
Types the automation store. There are still quite a few `any` in here but I thought I'd get the PR up


packages/builder/package.json
```diff
@@ -94,6 +94,7 @@
     "@sveltejs/vite-plugin-svelte": "1.4.0",
     "@testing-library/jest-dom": "6.4.2",
     "@testing-library/svelte": "^4.1.0",
+    "@types/shortid": "^2.2.0",
```
PClmnt, line 97: Had to add types for shortid as it was complaining, so the lockfile changed massively. In case anyone was wondering about the PR size

packages/builder/src/stores/builder/automations.ts
```diff
@@ -686,114 +636,113 @@ const automationActions = store => ({
...
   },
 
-  processBlockInputs: async (block, data) => {
+  processBlockInputs: async (
+    block: AutomationStep,
+    data: Record<string, any>
+  ) => {
     // Create new modified block
-    let newBlock = {
+    let newBlock: any = {
```
adrinr, line 745: Could this be `AutomationStep`?

adrinr (commented):
Looks good. I added a few comments

packages/builder/src/dataBinding.js
```diff
@@ -1528,6 +1533,7 @@ export const migrateReferencesInObject = ({ obj, label = "steps", steps }) => {
       migrateReferencesInObject({
         obj: obj[key],
         steps,
+        originalIndex,
```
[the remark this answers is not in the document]
PClmnt: We were actually missing a param in here - this is needed when migrating from the old `steps.1.whatever` binding - the number was missing so the migration wasn;t working correctly. A nice bug teased out by typescript.

adrinr (approved):
Nice cleanup!

PR #15228, opened 2024-12-20
Typing grid columns store

adrinr:
## Description
Typing grid columns store

### DIFF
--- a/packages/builder/tsconfig.json
+++ b/packages/builder/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
+    "outDir": "./dist",
     "baseUrl": ".",
     "paths": {
       "assets/*": ["./assets/*"],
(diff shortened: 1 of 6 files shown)


packages/builder/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
+    "outDir": "./dist",
```
adrinr, line 4: IDE were complaining for this field missing

packages/frontend-core/src/components/grid/stores/columns.ts
```diff
@@ -1,7 +1,40 @@
...
+export interface DerivedStore {
+  tableColumns: Readable<Column[]>
+  displayColumn: Readable<Column>
+  columnLookupMap: Readable<Record<string, Column>>
+  visibleColumns: Readable<Column[]>
+  scrollableColumns: Readable<Column[]>
+  hasNonAutoColumn: Readable<boolean>
+}
+
+type Column = FieldSchema & {
```
adrinr, line 19: We should probably refactor this when everything is typed

adrinr: Nice! Good idea having that centralised :)

PR #15229, opened 2024-12-20
Typing grid datasources store

adrinr:
## Description
Typing grid datasources store

### DIFF
--- a/packages/frontend-core/src/components/grid/stores/datasources/nonPlus.ts
+++ b/packages/frontend-core/src/components/grid/stores/datasources/nonPlus.ts
@@ -1,7 +1,28 @@
 import { SortOrder } from "@budibase/types"
 import { get } from "svelte/store"
+import { Store as StoreContext } from ".."
 
-export const createActions = context => {
+interface NonPlusActions {
+  nonPlus: {
+    actions: {
+      saveDefinition: () => Promise<void>
+      addRow: () => Promise<void>
+      updateRow: () => Promise<void>
+      deleteRows: () => Promise<void>
+      getRow: () => Promise<void>
+      isDatasourceValid: (datasource: {
+        type: string
+        id: string
+        tableId: string
+      }) => boolean
+      canUseColumn: (name: string) => boolean
+    }
+  }
+}
+
+export type Store = NonPlusActions
+
+export const createActions = (context: StoreContext): NonPlusActions => {
   const { columns, table, viewV2 } = context
 
   const saveDefinition = async () => {
@@ -20,7 +41,11 @@ export const createActions = context => {
     throw "This datasource does not support fetching individual rows"
   }
 
-  const isDatasourceValid = datasource => {
+  const isDatasourceValid = (datasource: {
+    type: string
+    id: string
+    tableId: string
+  }) => {
     // There are many different types and shapes of datasource, so we only
     // check that we aren't null
     return (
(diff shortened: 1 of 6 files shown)


packages/frontend-core/src/components/grid/stores/index.ts
```diff
@@ -44,20 +45,36 @@ const DependencyOrderedStores = [
   Users,
   Menu,
   Pagination,
-  Config,
+  Config as any,
```
adrinr, line 48: This needs to be typed as any for now, as we have some conflicts with the outputting types. It will be refactored when all the grid stores are typed

packages/frontend-core/src/components/grid/stores/datasources/table.ts
```diff
@@ -1,32 +1,55 @@
...
 import { get } from "svelte/store"
+import { Store as StoreContext } from ".."
 
 const SuppressErrors = true
 
-export const createActions = context => {
+interface TableActions {
+  table: {
+    actions: {
+      saveDefinition: (newDefinition: SaveTableRequest) => Promise<void>
```
mike12345567, line 16: Brilliant to see this kind of typing, all of this is very hidden otherwise!

packages/frontend-core/src/components/grid/stores/datasources/viewV2.ts
```diff
@@ -1,16 +1,42 @@
...
-export const createActions = context => {
+interface ViewActions {
+  viewV2: {
+    actions: {
+      saveDefinition: (newDefinition: UpdateViewRequest) => Promise<void>
+      addRow: (row: SaveRowRequest) => Promise<Row>
+      updateRow: (row: SaveRowRequest) => Promise<Row>
+      deleteRows: (rows: (string | Row)[]) => Promise<void>
+      getRow: (id: string) => Promise<Row>
+      isDatasourceValid: (datasource: {
```
mike12345567, line 20: Can we make a type for this - it looks quite different to a backend datasource given the `id` rather than `_id` and the `tableId` - would be good to document this in `@budibase/types` under `types/src/ui/stores` as a UIDatasource or something like that

packages/frontend-core/src/components/grid/stores/datasources/table.ts
```diff
@@ -1,32 +1,55 @@
...
+      addRow: (row: SaveRowRequest) => Promise<SaveRowResponse>
+      updateRow: (row: SaveRowRequest) => Promise<SaveRowResponse>
+      deleteRows: (rows: (string | Row)[]) => Promise<void>
+      getRow: (id: string) => Promise<Row>
+      isDatasourceValid: (datasource: { type: string; tableId: any }) => boolean
```
mike12345567, line 21: Similar to previous point, would be good to have a type for this that they all use (even for us to replace/update in future)

packages/frontend-core/src/components/grid/stores/datasources/nonPlus.ts
```diff
@@ -1,7 +1,28 @@
...
-export const createActions = context => {
+interface NonPlusActions {
+  nonPlus: {
+    actions: {
+      saveDefinition: () => Promise<void>
+      addRow: () => Promise<void>
+      updateRow: () => Promise<void>
+      deleteRows: () => Promise<void>
+      getRow: () => Promise<void>
+      isDatasourceValid: (datasource: {
```
mike12345567, line 13: Same comment - would be good for this to have UI type.

mike12345567 (approved):
LGTM - some nice stuff captured here!

PR #15235, opened 2024-12-24
Configure frontend-core

adrinr:
## Description
Configuring frontend-core to validate ts properly on the IDEs. Without this, some checks such as `svelte-checks` would complain but not the IDE, giving us a bad dev experience

### DIFF
--- a/packages/frontend-core/tsconfig.json
+++ b/packages/frontend-core/tsconfig.json
@@ -6,12 +6,7 @@
     "moduleResolution": "bundler",
     "outDir": "./dist",
     "skipLibCheck": true,
-    "allowJs": true,
-    "paths": {
-      "@budibase/types": ["../types/src"],
-      "@budibase/shared-core": ["../shared-core/src"],
-      "@budibase/bbui": ["../bbui/src"]
-    }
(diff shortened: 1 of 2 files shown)


packages/frontend-core/tsconfig.json
```diff
@@ -1,13 +1,12 @@
...
+  "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "target": "ESNext",
+    "module": "preserve",
     "moduleResolution": "bundler",
+    "outDir": "./dist",
     "skipLibCheck": true,
-    "paths": {
-      "@budibase/types": ["../types/src"],
-      "@budibase/shared-core": ["../shared-core/src"],
```
shogunpurple, line 8: Will this mess up the import aliases?

shogunpurple (approved):
Just one comment but assuming it’s not an issue or it wouldn’t build. LGTM

packages/frontend-core/tsconfig.json
adrinr: No, this is now defined in the root `tsconfig.json`

PR #15236, opened 2024-12-24
Typing grid datasources stores

adrinr:
## Description
Typing grid datasource. Given it's complexity, there are quite a lot of anys that will need to be refactored in the future. This will hopefully be done while updating the usages and the other stores, but we have an initial typed version we can use without changing the actual code.
Some types are conflicting between APIs, `type` field, etc. making it quite hard to work as it is without some refactors.


PClmnt (approved):
LGTM! So nice to get this typed!

PR #15239, opened 2024-12-24
Typing grid users stores

adrinr:
## Description
Typing grid users stores


packages/types/src/ui/stores/grid/user.ts
```diff
@@ -0,0 +1,6 @@
+import { User } from "@budibase/types"
+
+export interface UIUser extends User {
+  sessionId: string
+  gridMetadata?: { focusedCellId?: string }
+}
```
aptkingston, line 6: The store also adds `color` and `label` properties to users, which are both strings. Those are extracted and enriched via a derived store which is mixed back into the `users` export from `createStores`, but I can see that this sort of pattern (using an additional derived store to change the store "output" compared to the store "input") might be a bit of a problem with TS.

aptkingston (approved):

PR #15250, opened 2024-12-27
Typing grid clipboard store

adrinr:
## Description
Typing grid clipboard store


packages/frontend-core/src/components/grid/stores/clipboard.ts
```diff
@@ -1,10 +1,41 @@
...
-  const clipboard = writable({
+type ClipboardStoreData =
+  | {
+      value: string[][]
+      multiCellCopy: true
+    }
+  | {
+      value: string | null | undefined
+      multiCellCopy: false
+    }
```
aptkingston, line 15: This is really cool - I didn't know you can basically conditionally type it based on the value of another field like this.

packages/frontend-core/src/components/grid/stores/clipboard.ts
```diff
@@ -1,10 +1,41 @@
...
 import { Helpers } from "@budibase/bbui"
 import { parseCellID, getCellID } from "../lib/utils"
 import { NewRowID } from "../lib/constants"
+import { Store as StoreContext } from "."
 
-export const createStores = () => {
-  const clipboard = writable({
+type ClipboardStoreData =
+  | {
+      value: string[][]
```
aptkingston, line 9: `value` here can be any cell value, so it doesn't need to just be a string. We could be specific about the range possible types but it can be any data type that we support, so `any` is probably appropriate.

adrinr: I thought that the value would always be stringified. Updated to any

aptkingston (approved):

PR #15255, opened 2024-12-27
Typing grid rows stores

adrinr:
## Description
Typing grid rows stores


packages/frontend-core/src/components/grid/stores/datasources/nonPlus.ts
```diff
@@ -97,34 +90,5 @@ export const initialise = (context: StoreContext) => {
...
-        const $fetch = get(fetch)
-        if (!isSameDatasource($fetch?.options?.datasource, $datasource)) {
-          return
-        }
-        $fetch.update({
-          sortOrder: $sort.order || SortOrder.ASCENDING,
-          sortColumn: $sort.column,
-        })
-      })
-    )
```
aptkingston, line 128: These are needed for inline filtering and sorting to work - what was the reason for removing them?

Normal usage of these features:
- Right click column header and change sorting
- Seaching via inline filtering (hovering over the column header and clicking the search icon, then entering a value)

packages/frontend-core/src/components/grid/stores/datasources/table.ts
```diff
@@ -108,34 +112,5 @@ export const initialise = (context: StoreContext) => {
...
-        const $fetch = get(fetch)
-        if ($fetch?.options?.datasource?.tableId !== $datasource.tableId) {
-          return
-        }
-        $fetch.update({
-          sortOrder: $sort.order || SortOrder.ASCENDING,
-          sortColumn: $sort.column,
-        })
-      })
-    )
```
aptkingston, line 139: Same as above.

packages/frontend-core/src/components/grid/stores/datasources/viewV2.ts
```diff
@@ -207,38 +201,6 @@ export const initialise = (context: StoreContext) => {
...
-        }
-        const $fetch = get(fetch)
-        if ($fetch?.options?.datasource?.id !== $datasource.id) {
-          return
-        }
-        $fetch.update({
-          filter: $allFilters,
-        })
-      })
-    )
```
aptkingston, line 240: Same as above.

aptkingston (changes requested):
LGTM other than the removal of code required for sorting and inline searching.

packages/frontend-core/src/components/grid/stores/datasources/nonPlus.ts
adrinr: I totally missed that. I actually wanted to ask you, as I did not see the actual usage of that, and I thought it was some legacy code. Fixed!

aptkingston (approved):

PR #15262, opened 2024-12-30
Builder store typescript conversions

PClmnt:
## Description
Converts app.js and builder.js to TS


adrinr (approved):

packages/builder/src/stores/builder/builder.ts
```diff
@@ -125,27 +144,27 @@ export class BuilderStore extends BudiStore {
...
+  startBuilderOnboarding(): void {
     this.update(state => ({
       ...state,
       onboarding: true,
       tourKey: TOUR_KEYS.TOUR_BUILDER_ONBOARDING,
     }))
   }
 
-  endBuilderOnboarding() {
+  endBuilderOnboarding(): void {
```
aptkingston, line 155: Do we always need to specify these `void` return types? I've always left them out because it can trivially work out that it's a void return.

aptkingston (approved):

packages/builder/src/stores/builder/builder.ts
PClmnt: Sorry I merged before I noticed your review, ill get them removed in my next PR here

PR #15264, opened 2024-12-30
Typing grid config store

adrinr:
## Description
Typing grid config store

### DIFF
--- a/packages/frontend-core/src/components/grid/layout/Grid.svelte
+++ b/packages/frontend-core/src/components/grid/layout/Grid.svelte
@@ -47,7 +47,6 @@
   export let buttonsCollapsedText = null
   export let darkMode = false
   export let isCloud = null
-  export let rowConditions = null
   export let aiEnabled = false
 
   // Unique identifier for DOM nodes inside this instance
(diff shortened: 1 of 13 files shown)


packages/frontend-core/src/components/grid/layout/Grid.svelte
```diff
@@ -47,7 +47,6 @@
   export let buttonsCollapsedText = null
   export let darkMode = false
   export let isCloud = null
-  export let rowConditions = null
```
adrinr, line 50: This is not used anywhere

packages/frontend-core/src/components/grid/lib/utils.ts
```diff
@@ -14,7 +14,9 @@ export const getCellID = (rowId: string, fieldName: string) => {
   return `${rowId}${CellIDSeparator}${fieldName}`
 }
 
-export const parseEventLocation = (e: MouseEvent & TouchEvent) => {
+export const parseEventLocation = (event: MouseEvent | TouchEvent) => {
+  const e = event as MouseEvent & TouchEvent
```
adrinr, line 18: Mixing params with actual usage. The current code uses `clientX` to detect what type is it. We could add some guard typing, but this does the job for now without modifying the executable code

aptkingston (approved):

PR #15267, opened 2024-12-30
Component, Integrations and componentTreeNode stores converted to typescript

PClmnt:
## Description
Converts Components.js / Integrations.js / componentTreeNodes.js to typescript

Also removed unneeded voids


packages/builder/src/stores/builder/componentTreeNodes.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface OpenNodesState {
+  [key: string]: boolean
+}
+
+export class ComponentTreeNodesStore extends BudiStore<OpenNodesState> {
+  private baseStore = createSessionStorageStore(
+    "openNodes",
+    {} as OpenNodesState
+  )
```
aptkingston, line 16: NAB: would it work to add a flag to `BudiStoreOpts` to control what sort of store is created - as in if we should persist it in session storage, local storage, or no persistence (i.e. a normal store)? The stores return from all of those should implement `Writable<T>` so the only difference would be how the main store is created in the BudiStore constructor. Maybe there's a reason that wouldn't work in practice, but just thought that might be a cleaner way to handle this instead of adding a new `baseStore` and overriding a bunch of things.

PClmnt: @aptkingston What do you think about that? was that what you meant?

PClmnt: (there's currently an error from this though so investigating)

aptkingston (approved):
LGTM! Yep that was exactly what I meant 👌

PR #15273, opened 2024-12-31
Respect grid config in builder

adrinr:
## Description
We found an issue with the builder, not updating order and width settings of the columns if these are not directly specified. Because of this, views without specific order or width, instead of using the default values, will use the last provided settings. This will cause inconsistency, as the same view will be displayed differently depending on the user journey.

Having the following views:
1. Narrow columns with columns being reordered
image

2. Default view, with no settings modified
image

3. Wide view, with only column width being modified
image

Previously, the column order and width were not respected in view 2, and neither was the order in view 3:
https://github.com/user-attachments/assets/7cfe408c-d48e-42c8-9287-f2ab23dda5d7

After the fix:

https://github.com/user-attachments/assets/8a7c3b0a-7394-4cf4-a97c-59c3d1f028ed

## Launchcontrol
Fix grid width and order behaviour when using default values


aptkingston (approved):
Yea I can see that in the specific case of swapping between views in the data section then the "old column" logic would use order and size from a column in a different view. I can't fully remember why I added that in the first place, but if it's needed for something then I'm sure it'll appear again and we can add a solution for both.

PR #15278, opened 2025-01-02
Ignore default values for view sort

adrinr:
## Description
While QAing an issue related to views was found. Due the frontend setting default values for grid sort, unnecessary updates are submitted when creating new views. This is happening in both views and calculation views, and due some recent changes on the backend calculation views are displaying some errors. Even so, the views are created correctly.

https://github.com/user-attachments/assets/8ad9fc11-b121-421a-a9f5-3f8fd9c1579b

Instead of fixing the error on the backend to not throw, the frontend has been fixed in order to not create these calls.

The problematic flow does as it follows:
1. Create a new view (by default, the sort configuration is not set)
2. Update the datasource in the frontend accordingly
3. The frontend sees that the sort config is not defined, so it sets the default order as ascending
4. As there is a change in the config, it tries to update the definition via API (this happens as it cannot differentiate if it was a user change or an "automatic" one)

This is currently happening in production

## Launchcontrol
Remove wrong and confusing message when creating new views

### DIFF
--- a/packages/frontend-core/src/components/grid/stores/datasources/viewV2.ts
+++ b/packages/frontend-core/src/components/grid/stores/datasources/viewV2.ts
@@ -151,6 +151,35 @@ export const initialise = (context: StoreContext) => {
       })
     )
 
+    function sortHasChanged(
+      newSort: {
+        column: string | null | undefined
+        order: SortOrder
+      },
+      existingSort?: {
+        field: string
+        order?: SortOrder
+      }
+    ) {
+      const newColumn = newSort.column ?? null
+      const existingColumn = existingSort?.field ?? null
+      if (newColumn !== existingColumn) {
+        return true
+      }
+
+      if (!newColumn) {
+        return false
+      }
+
+      const newOrder = newSort.order ?? null
+      const existingOrder = existingSort?.order ?? null
+      if (newOrder !== existingOrder) {
+        return true
+      }
+
+      return false
+    }
+
     // When sorting changes, ensure view definition is kept up to date
     unsubscribers.push(
       sort.subscribe(async $sort => {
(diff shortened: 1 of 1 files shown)


aptkingston:
Could we fix this in a more simple way by just checking if a sort column is set before saving the view definition?
```js
        // If we can mutate schema then update the view definition
        if (get(config).canSaveSchema && $sort?.column) {
```
Either that or just terminate this whole block immediately if sort column is undefined? Since I don't think it's actually possible for a view sort column to become undefined while one is mounted in the frontend. It's only possible if you delete the column from the table (you can't delete columns while using a view).

adrinr:
> Could we fix this in a more simple way by just checking if a sort column is set before saving the view definition?
> 
> ```js
>         // If we can mutate schema then update the view definition
>         if (get(config).canSaveSchema && $sort?.column) {
> ```
> 
> Either that or just terminate this whole block immediately if sort column is undefined? Since I don't think it's actually possible for a view sort column to become undefined while one is mounted in the frontend. It's only possible if you delete the column from the table (you can't delete columns while using a view).

We can't do this as we have the option of removing this config to go back to the default behaviour. This would not allow it.

aptkingston:
Ah yes I forgot it's possible to unset sorting again using the popover.

I think I was just confused because it seems like a lot of code for a small logic change - it's equivalent to this isn't it?

Changing 
```js
        // Skip if nothing actually changed
        if (
          $sort?.column === $view.sort?.field &&
          $sort?.order === $view.sort?.order
        ) {
          return
        }
```

to
```js
        // Skip if nothing actually changed
        if (
          $sort?.column === $view.sort?.field &&
          (!$sort?.column || $sort?.order === $view.sort?.order)
        ) {
          return
        }
```

But if you prefer the new function to be more verbose and explicit about the logic then I don't mind merging that instead 👍 Since it's a little hard to follow inline like my example.

aptkingston (approved):

PR #15281, opened 2025-01-02
Flags / Layout / Preview stores to TS

PClmnt:
## Description
Converts Flags / Layout / Preview stores to Typescript


packages/builder/src/stores/builder/layouts.ts
```diff
@@ -69,6 +79,24 @@ export class LayoutStore extends BudiStore {
...
+              layout => layout._id === $store.selectedLayoutId
+            ) || null,
+        }
+      })
+    }
+    super(INITIAL_LAYOUT_STATE, makeDerivedStore)
+  }
+}
+
+export const selectedLayout = new SelectedLayoutStore(layoutStore)
```
aptkingston, line 102: NAB: we can leave this as is since you've already converted it, but I think in this case we could leave the selected layout as a normal derived store rather than a derived budi store. It's just a simple piece of derived state rather than a full store with functions etc.

aptkingston (approved):

packages/builder/src/stores/builder/layouts.ts
PClmnt: I'm gonna move it back, it was much cleaner before!

PR #15285, opened 2025-01-02
Convert portal environment store to TS

aptkingston:
## Description
Convert environment store to TS


packages/builder/src/stores/portal/environment.ts
```diff
@@ -0,0 +1,79 @@
...
+    const status = await API.checkEnvironmentVariableStatus()
+    this.update(store => {
+      store.status = status
+      return store
+    })
+  }
+
+  async loadVariables() {
+    if (get(licensing).environmentVariablesEnabled) {
+      const envVars: string[] = (await API.fetchEnvironmentVariables())
```
PClmnt, line 41: NAB: Don't think we need to type this do we? `string[]` will be inferred from the API? I could be wrong

PClmnt (approved):

packages/builder/src/stores/portal/environment.ts
aptkingston: We do actually, only because the API response type is a custom thing (something along the lines of FetchEnvironvmentVariablesResponse) which is itself just aliased to a `string[]`, but I'd rather store a normal string array in the store instead of the response type, hence the casting here.

PR #15291, opened 2025-01-03
User and websocket stores converted to typescript

PClmnt:
## Description
Converted the users store to typescript and updated the websocket and websocket utils files also.


packages/builder/src/stores/builder/users.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.actions = {
+      init: this.init.bind(this),
+      updateUser: this.updateUser.bind(this),
+      removeUser: this.removeUser.bind(this),
+      reset: this.reset.bind(this),
+    }
+  }
+
+  init(users: UIUser[]) {
+    this.store.set(users)
```
aptkingston, line 24: NAB: BudiStore already binds `set`, `subscribe` and `update` to itself, so you can just do `this.update`, `this.set` etc. Again, no functional difference though!

aptkingston (approved):

adrinr (approved):

PR #15292, opened 2025-01-03
Convert portal feature store to TS

aptkingston:
## Description
Convert portal feature store to TS.

Also updated `DerivedBudiStore` to remove the restriction of `DerivedT` having to extend `T`. The result being that the derived state can be a completely different shape than the one stored, which is helpful for stores like this.

Also added another small util to the types packages to help determine whether the value returned from the API when trying to fetch a config (typed as `Config | {}`) is indeed a valid `Config`.


samwho (approved):
Neat! I hadn't seen our DerivedBudiStore type yet and I see it's being used as a way to essentially "fold" 2 different state sources into a single boolean that's easier to consume.

aptkingston:
> Neat! I hadn't seen our DerivedBudiStore type yet and I see it's being used as a way to essentially "fold" 2 different state sources into a single boolean that's easier to consume.

Exactly. We quite often use a pattern where we have a normal, writable svelte store for some state, but then use an additional derived svelte store to either enrich or transform that state for consumption by components - then combine those 2 stores together (by way of overriding the `subscribe` property of the writable store to be that of the derived store), creating a pseudo store that stores data of one type, but when read provides data of another type.

Such stores aren't directly mutated from components (as it would be weird to store something, then read it back as something different), but internally will fetch some raw data from the API and store it, then provide some actions for components to consume.

The `DerivedBudiStore` is a typescript-friendly way of handling that pattern - store type `T` and read type `DerivedT`.

Not sure if I would arrived at this same pattern if we had used typescript from day 1, but it worked very well pre-typescript and I think this is a reasonably clean way of continuing to support it.

PR #15294, opened 2025-01-03
Convert portal groups store to TS

aptkingston:
## Description
Convert portal groups store to TS.


packages/builder/src/stores/portal/groups.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private async refreshGroup(groupId: string) {
+    const group = await API.getGroup(groupId)
+    this.updateStore(group)
+  }
+
+  async save(group: UserGroup) {
+    const { ...dataToSave } = group
+    delete dataToSave.scimInfo
```
samwho, line 40: It's a real shame there's not a more elegant way to do this. I've seen `const { scimInfo, ...rest } = group` used in the past but the linter doesn't like it.

packages/builder/src/stores/portal/groups.ts
```diff
@@ -0,0 +1,93 @@
...
+    group._id = response._id
+    group._rev = response._rev
+    this.updateStore(group)
+    return group
+  }
+
+  async delete(group: UserGroup) {
+    await API.deleteGroup(group._id!, group._rev!)
+    this.update(state => {
+      state = state.filter(state => state._id !== group._id)
```
samwho, line 51: More of an academic question than something to act on: does this copy `state`? Is there a way to do it without copying? e.g. `state.find` followed by `state.splice`?

Something I was extremely surprised by in the course of writing my queueing blog post is just how fast `array.splice` is in JS. It appears to be O(1), iirc.

packages/builder/src/stores/portal/groups.ts
```diff
@@ -0,0 +1,93 @@
...
+      return state
+    })
+  }
+
+  async addUser(groupId: string, userId: string) {
```
samwho, line 56: It might make for a less error-prone API if this were an object `{ groupId: string, userId: string }`. Passing 2 strings in the correct order is a weirdly difficult task for human beings.

packages/builder/src/stores/portal/groups.ts
samwho: https://jsbm.dev/9sJxjgmMBsVe5

samwho (approved):
Just a few small comments that I don't mind if you action or ignore. 🙂

packages/builder/src/stores/portal/groups.ts
aptkingston: That's a very nice benchmarking site - TIL! Interesting that find/splice is about 2x faster. It actually got slightly  faster again (15 op/s to 16 on this M1 MBP) when I replaced `array.find` with `array.findIndex`, which I think is what you meant to use.

aptkingston:
Worth noting that I've only updated this file to TS and haven't written or changed any of those lines, but I do agree with your points! 
I'll update the filter to find/splice though because I can't say no to free performance gains.

PR #15295, opened 2025-01-03
Typing frontend-core fetch apis

adrinr:
## Description
Typing all the fetch classes from frontend-core. These conversions still contain some anys and some code could be simplified and cleaned (specially around nulls).
As this is a quite critical change, the approach was to modify as little as possible. Once merged in master, we can tackle case by case, improving the quality of these classes if required.


packages/builder/src/components/design/settings/controls/DataSourceSelect/DataSourceSelect.svelte
```diff
@@ -116,8 +115,11 @@
       }
     })
   $: fields = bindings
-    .filter(x => arrayTypes.includes(x.fieldSchema?.type))
```
adrinr, line 119: This was causing the attachments and array fields to not be displayed properly on the dataprovider selector

packages/frontend-core/src/api/views.ts
```diff
@@ -20,7 +28,7 @@ export const buildViewEndpoints = (API: BaseAPIClient): ViewEndpoints => ({
   fetchViewData: async (name, { field, groupBy, calculation }) => {
     const params = new URLSearchParams()
     if (calculation) {
-      params.set("field", field)
+      params.set("field", field!)
```
mike12345567, line 31: is this safe? Should we not check if field is set when this function is called and error if not (or not set the param)

packages/frontend-core/src/components/grid/stores/rows.ts
```diff
@@ -10,17 +10,18 @@ import {
...
+import DataFetch from "../../../fetch/DataFetch"
 
 interface IndexedUIRow extends UIRow {
   __idx: number
 }
 
 interface RowStore {
   rows: Writable<UIRow[]>
-  fetch: Writable<UIFetchAPI | null>
+  fetch: Writable<DataFetch<any, any, any> | null>
```
mike12345567, line 24: These anys, can we document what they should be at some point in the future?

packages/frontend-core/src/components/grid/stores/datasource.ts
```diff
@@ -184,9 +184,9 @@ export const createActions = (context: StoreContext): ActionDatasourceStore => {
   const refreshDefinition = async () => {
     const def = await getDatasourceDefinition({
       API,
-      datasource: get(datasource),
+      datasource: get(datasource) as any,
```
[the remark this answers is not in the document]
samwho: I feel like we can reasonably treat all occurrences of `any` as todos.

packages/frontend-core/src/api/views.ts
adrinr: It should, as the frontend requires field to be mandatory if the calculation is set. I did try to create a type that reflects this but it got a little bit too messy.
To simplify it, I just removed the `!` in favour of a proper check

packages/frontend-core/src/fetch/DataFetch.ts
```diff
@@ -130,26 +211,26 @@ export default class DataFetch {
...
-    const features = this.determineFeatureFlags(definition)
+    const features = await this.determineFeatureFlags()
     this.features = {
       supportsSearch: !!features?.supportsSearch,
       supportsSort: !!features?.supportsSort,
       supportsPagination: paginate && !!features?.supportsPagination,
     }
 
-    // Fetch and enrich schema
-    let schema = this.getSchema(datasource, definition)
```
mike12345567, line 147: Is it safe to remove the enrichment and just check the base definition?

adrinr: I think this is actually not needed, but added it back just in case. It could be required for views v1

mike12345567 (approved):
LGTM - a lot of conversion!

PR #15299, opened 2025-01-03
Theme / Row actions / Snippers stores converted to typescript. 

PClmnt:
## Description
Converts Theme / Row actions / Snippers stores to typescript and removes an unused method in the Row Actions store.


packages/builder/src/stores/builder/rowActions.ts
```diff
@@ -34,26 +45,30 @@ export class RowActionStore extends BudiStore {
...
     // Get a unique name for this action
     if (!name) {
-      const existingRowActions = get(this.store)[tableId] || []
+      const existingRowActions = get(this)[tableId] || []
       name = getSequentialName(existingRowActions, "New row action ", {
         getName: x => x.name,
       })
     }
 
+    if (!name) {
```
mike12345567, line 68: Is this potentially an error scenario?

mike12345567 (approved):
LGTM - just a question about a possible error scenario.

packages/builder/src/stores/builder/rowActions.ts
PClmnt: You're right, let's be more explicit and actually return an error here. I can't see a case in which it gets here, something would need to break in the `getSequentialName` function which is itself a fallback. But will throw here just to be safe

PR #15286, opened 2025-01-02
Convert portal feature flags store to TS

aptkingston:
## Description
Convert portal feature flags store to TS.

PR #15308, opened 2025-01-06
Put `FeatureFlags` type in `packages/types`, simplify flag types

samwho:
## Description

This PR simplifies the types around our feature flagging system. The value they were meant to be providing hasn't been realised, so I'm paring it back to something simpler and easier to reason about.

This PR:

- Makes it such that flags can only be booleans for now. We haven't used strings or numbers yet so the code is more complex than it needs to be. We can add them back in when we need them.
- Exposes the core `FeatureFlags` type in `packages/types` so that it can be used in `GetGlobalSelfResponse` in thus the frontend code (which should make https://github.com/Budibase/budibase/pull/15286 nicer).
- Gets rid of a pet peeve I've had with the API of feature flags. It used to be that you had to do `features.flags.isEnabled(FeatureFlag.WHATEVER)` which always felt wordy and repetitive. I've made it such that the `flags` bit in the middle is no longer required.

Companion pro PR: https://github.com/Budibase/budibase-pro/pull/348


mike12345567 (approved):
LGTM - nicely simplified, the downside of not being able to set per environment could be a slight issue, the main thing being dev vs prod flags, but we can deal with that if it comes up again!

PR #15309, opened 2025-01-06
Add save button to user screen

melohagan:
## Description
Solves minor race conditions with the first name and last name fields of the user editing screen.
Now using a `Save` button that enabled on input.

## Addresses
- https://linear.app/budibase/issue/GRO-789/improve-users-section-of-budibase-portal

## Screenshots

Screenshot 2025-01-06 at 13 13 01
*Initial state*

Screenshot 2025-01-06 at 13 13 29
*Changes made*

Screenshot 2025-01-06 at 13 13 51
*Save clicked*


samwho (approved):
The backend changes appear to be semantically identical, just avoiding a lookup right? If so, this LGTM from me as far as backend changes go.

packages/builder/src/pages/builder/portal/users/users/[userId].svelte
```diff
@@ -164,40 +165,45 @@
...
+      await users.save({ ...user, ...userFieldsToUpdate })
+      userFieldsToUpdate = {}
       await fetchUser()
     } catch (error) {
       notifications.error("Error updating user")
     }
   }
 
+  async function updateUserFirstName(evt) {
+    userFieldsToUpdate = { ...userFieldsToUpdate, firstName: evt.target.value }
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/portal/users/users/[userId].svelte:179]: NAB: we don't need to clone the object here - we can just do `userFieldsToUpdate.firstName = evt.target.value`.
Same applies to the other updates as well.

aptkingston (approved):

PR #15310, opened 2025-01-06
Convert role and published store to TS

PClmnt:
## Description
Convert role and published store to TS


packages/builder/src/stores/builder/roles.ts
```diff
@@ -0,0 +1,96 @@
...
+      })
+    }
+  }
+}
+
+const store = new RoleStore()
+export const roles = {
+  ...store,
+  subscribe: store.enriched.subscribe,
+}
```
aptkingston, line 96: This is another prime candidate for a derived budi store, but this works too! 👍

aptkingston (approved):

PR #15313, opened 2025-01-06
Screen store TS conversion

deanhannigan:
## Description
Store and JSDoc updates required to convert the `Screen` builder store to TypeScript.

## Launchcontrol
Convert the `Screen` builder store into TypeScript.


packages/builder/src/stores/builder/components.ts
```diff
@@ -33,17 +33,28 @@ import { Utils } from "@budibase/frontend-core"
 import { Component, FieldType, Screen, Table } from "@budibase/types"
 import { utils } from "@budibase/shared-core"
 
-interface ComponentDefinition {
+export interface ComponentState {
+  components: Record<string, ComponentDefinition>
+  customComponents: string[]
+  selectedComponentId?: string | null
```
aptkingston, line 39: We should probably use either an optional property or type it as `string | null` shouldn't we? Rather than both. I've been opting to go for optional properties - the only change being we'll need to ensure we always clear these by setting them to `undefined` rather than `null`.

packages/builder/src/stores/builder/components.ts
```diff
@@ -440,6 +442,11 @@ export class ComponentStore extends BudiStore<ComponentState> {
    * @returns
    */
   createInstance(componentName: string, presetProps: any, parent: any) {
+    const screen = get(selectedScreen)
+    if (!screen || !selectedScreen) {
```
aptkingston, line 446: NAB: the `!selectedScreen` part is redundant here.

packages/builder/src/stores/builder/screens.ts
```diff
@@ -13,15 +13,33 @@ import {
...
+  Screen,
+  Component,
+  SaveScreenResponse,
+} from "@budibase/types"
+import { ComponentDefinition } from "./components"
+
+interface ScreenState {
+  screens: Screen[]
+  selectedScreenId?: string
+  selected?: Screen
```
aptkingston, line 28: I can't see any external usages of this `selected` property other than in the `afterAction` of the history store constructor, so it'll always be null... I'm assuming this is maybe a leftover and we can refactor it out? I think in after action we should be reading the `selectedScreen` derived store and using that to select a component instead of this `selected` property, which we can remove.

aptkingston (commented):
Few initial comments but looks good overall!

aptkingston (approved):

adrinr (approved):

PR #15326, opened 2025-01-08
Type portal user store

aptkingston:
## Description

## Addresses
- ...more if required

## App Export

## Screenshots

## Launchcontrol

### DIFF
--- a/packages/backend-core/src/users/db.ts
+++ b/packages/backend-core/src/users/db.ts
@@ -22,6 +22,7 @@ import {
   PlatformUserBySsoId,
   PlatformUserById,
   AnyDocument,
+  BaseUser,
 } from "@budibase/types"
 import {
   getAccountHolderFromUsers,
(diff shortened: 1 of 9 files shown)


packages/backend-core/src/users/db.ts
```diff
@@ -362,13 +363,13 @@ export class UserDB {
...
         // create the promises array that will be called by bulkDocs
-        newUsers.forEach((user: any) => {
+        newUsers.forEach(user => {
           usersToSave.push(
             UserDB.buildUser(
               user,
               {
                 hashPassword: true,
-                requirePassword: user.requirePassword,
+                requirePassword: false,
```
aptkingston, line 372: `requirePassword` never exists on any user type, so I don't know why this was here. It's also never referenced at all in the frontend, so we definitely don't mix this in to the user type before sending it up.

PR #15328, opened 2025-01-08
Related to account-portal change

melohagan:
## Description
Account portal already hashes the password.

**Do not merge** - Deployment strategy needed

## Addresses
- https://linear.app/budibase/issue/GRO-886/issue-with-forgot-password-reset-email-not-being-received-for-tenant


melohagan:
Going to move this code into account-portal instead.

PR #15330, opened 2025-01-09
Convert views and published stores to TS

PClmnt:
## Description
Convert views and published stores to TS


packages/builder/src/stores/builder/views.ts
```diff
@@ -0,0 +1,94 @@
...
+  select = (name: string) => {
+    this.store.update(state => ({
+      ...state,
+      selectedViewName: name,
+    }))
+  }
+
+  delete = async (view: View) => {
+    if (!view.name) {
+      return
```
adrinr, line 58: Is this an expected use case? Should we throw an error?

packages/builder/src/stores/builder/views.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (table?.views && view.name) {
+        delete table.views[view.name]
+      }
+      return { ...state }
+    })
+  }
+
+  save = async (view: View & { originalName?: string }) => {
+    if (!view.name) {
+      return
```
adrinr, line 74: Same here

adrinr (approved):

PR #15331, opened 2025-01-09
Typing notification store

adrinr:
## Description
While QAing, I found an existing issue for apps trying to access pg instances that are not available. On that case the API returns the following response:
image

The current code, when the errors don't contain a message, it falls back to showing the "error" field. In this case it can be objects, displaying a confusing message to the end user:
image

While diagnosing the issue, I typed the notification store and the related client stores.

The end result is not ideal, as the server should probably throw a better formatted error, but this makes sure that errors throwned to the client are at least readable:
image

## Launchcontrol


packages/client/src/api/api.ts
```diff
@@ -1,6 +1,10 @@
 import { createAPIClient } from "@budibase/frontend-core"
-import { authStore } from "../stores/auth.js"
-import { notificationStore, devToolsEnabled, devToolsStore } from "../stores/"
+import { authStore } from "../stores/auth"
+import {
+  notificationStore,
+  devToolsEnabled,
+  devToolsStore,
+} from "../stores/index.js"
```
mike12345567, line 7: Is the `index.js` needed?

mike12345567 (approved):
LGTM!

packages/client/src/api/api.ts
adrinr: No, removed

PR #15336, opened 2025-01-09
Explicitly fetch the data from fetch

adrinr:
## Description
Updating data fetchers to do not fetch data by default on the constructor. We have some use cases where we don't want to do this, and the way we use to handle this was passing an empty datasource on the constructor. This was changed in here: https://github.com/Budibase/budibase/pull/15332
This PR change the behaviour to work as expected.

Also, reading the reverted changes from here: https://github.com/Budibase/budibase/pull/15332, fixing an issue while loading the forms.

### QA process
1. Grid calls a single search, new forms are not performing any search and edit forms run a single search with limit 1

https://github.com/user-attachments/assets/a71599c7-3be3-4694-a36c-179aad109bbd

2. Forms with relationships run one extra search for relationship field

https://github.com/user-attachments/assets/02620ab1-5338-4c9e-bcdb-b471c07943d8

3. Grid on data section loads as expected

https://github.com/user-attachments/assets/b56c2392-9d9d-4e27-ade7-4fb9a52830af

4. Multiple datasources are properly displayed on the clients: as tables, forms, repeaters...

https://github.com/user-attachments/assets/98340e10-5889-4260-a076-43a290021bb1

### DIFF
--- a/packages/client/src/utils/schema.js
+++ b/packages/client/src/utils/schema.js
@@ -31,7 +31,7 @@ const getDatasourceFetchInstance = datasource => {
   if (!handler) {
     return null
   }
-  return new handler({ API })
+  return new handler({ API, datasource })
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM - I don't know this area massively well, but if it doesn't spam requests and the form is fixed that's what matters!

PR #15337, opened 2025-01-09
Fix date-only searching.

samwho:
## Description

Bit of a thorny one, this. Mostly affected SQS.

The problem boils down to how we store dates as strings in SQS. The problem is that we haven't been consistent in how we do it: it's possible to save a date as either `2020-01-01` or `2020-01-01T00:00:00Z`. It's also possible to search for a date with either format. Because they're strings, despite being semantically identical (if you assume UTC both sides), they do not match.

What I've done is make sure that regardless of whether you search or save the date in either format, you can always find them.

## Addresses
- https://linear.app/budibase/issue/BUDI-8957/date-filter-equals-does-not-work

## Launchcontrol

- Fixes a problem that caused no rows to be returned when applying filters to date-only fields.


packages/backend-core/src/middleware/errorHandling.ts
```diff
@@ -32,8 +32,12 @@ export async function errorHandling(ctx: any, next: any) {
     }
 
     if (environment.isTest() && ctx.headers["x-budibase-include-stacktrace"]) {
+      let rootErr = err
+      while (rootErr.cause) {
+        rootErr = rootErr.cause
+      }
       // @ts-ignore
-      error.stack = err.stack
+      error.stack = rootErr.stack
```
samwho, line 40: This was bothering me when tests failed. I only got shown the outermost error, which was never where the real problem was. This makes the error stack traces substantially more useful.

packages/backend-core/src/sql/sql.ts
```diff
@@ -816,14 +816,29 @@ class InternalBuilder {
...
+            schema?.type === FieldType.DATETIME &&
+            schema.dateOnly
+          ) {
+            for (const value of values) {
+              if (value != null) {
+                q = q.or.whereLike(key, `${value.toISOString().slice(0, 10)}%`)
+              } else {
+                q = q.or.whereNull(key)
+              }
+            }
```
samwho, line 838: It's a shame we can't easily use `whereIn` here. We _could_, I suppose, if we expanded out, say, `2020-01-01` to `["2020-01-01", "2020-01-01T00:00:00Z"]` but that felt more fraught than a bunch of LIKE queries.

packages/backend-core/src/sql/utils.ts
```diff
@@ -149,15 +149,7 @@ export function isInvalidISODateString(str: string) {
...
-  const trimmedValue = str.trim()
-  if (!ISO_DATE_REGEX.test(trimmedValue)) {
-    return false
-  }
-  let d = new Date(trimmedValue)
-  if (isNaN(d.getTime())) {
-    return false
-  }
-  return d.toISOString() === trimmedValue
+  return ISO_DATE_REGEX.test(str.trim())
```
samwho, line 152: This function is used in a single place, and it's where we parse incoming data to be searched on. I want both dates and date time to get parsed into Date objects, so I've modified this function to include anything that looks like `2020-01-01` or `2020-01-01T00:00:00Z`.

It stood out to me that the regex we use does not consider any other timezones, not sure how big a deal that is in practice.

packages/server/src/api/routes/tests/search.spec.ts
```diff
@@ -1683,6 +1683,151 @@ if (descriptions.length) {
                 })
               })
 
+            describe("datetime - date only", () => {
+              describe.each([true, false])(
+                "saved with timestamp: %s",
+                saveWithTimestamp => {
+                  describe.each([true, false])(
+                    "search with timestamp: %s",
+                    searchWithTimestamp => {
```
samwho, line 1692: Yep I hate it too, but we're in a position where dates could have been saved into SQS with or without time components, and you can also search with or without a time component. For dateOnly fields, I think the best trade-off is to handle the cross product of all eventualities in the same way: ignore time.

packages/server/src/utilities/rowProcessor/index.ts
```diff
@@ -411,6 +411,15 @@ export async function coreOutputProcessing(
...
       }
+    } else if (column.type === FieldType.DATETIME && column.dateOnly) {
+      for (const row of rows) {
+        if (typeof row[property] === "string") {
+          row[property] = new Date(row[property])
+        }
+        if (row[property] instanceof Date) {
+          row[property] = row[property].toISOString().slice(0, 10)
+        }
+      }
```
samwho, line 422: We used to return date only fields with a timestamp sometimes, in a format that wasn't consistent (depended how each database returned the data, e.g. postgres would do `2020-01-01 +00:00:00`). This unifies them all to return `2020-01-01`.

packages/shared-core/src/filters.ts
```diff
@@ -699,7 +699,27 @@ export function runQuery<T extends Record<string, any>>(
...
+      return false
+    }
+
+    const leftDate = dayjs(docValue)
+    if (leftDate.isValid()) {
+      const rightDate = dayjs(testValue)
+      if (rightDate.isValid()) {
+        return leftDate.isSame(rightDate)
+      }
+    }
```
samwho, line 720: This is to help in-memory search match date-only fields correctly when either side is a date/datetime.

mike12345567 (approved):
LGTM!

PR #15338, opened 2025-01-10
Typification of util schema

adrinr:
## Description
Converting the util schema to ts. While doing this I realised that some of the conversions I used for the datafetchers were much more complex than they needed, and they were not 100% accurate. I updated to use their own definitions, and after all these are the definitions used from the svelte files.


packages/client/src/sdk.js
```diff
@@ -29,7 +29,7 @@ import { ActionTypes } from "./constants"
 import {
   fetchDatasourceSchema,
   fetchDatasourceDefinition,
-} from "./utils/schema.js"
+} from "./utils/schema.ts"
```
mike12345567, line 32: is the `.ts` needed?

packages/frontend-core/src/fetch/ViewFetch.ts
```diff
@@ -1,7 +1,14 @@
-import { Table, View } from "@budibase/types"
+import { Table } from "@budibase/types"
 import DataFetch from "./DataFetch"
 
-type ViewV1 = View & { name: string }
+type ViewV1 = {
```
mike12345567, line 4: Should this live in the BB types package, is it a UI only thing?

packages/client/src/sdk.js
adrinr: No. Removed!

packages/frontend-core/src/fetch/ViewFetch.ts
adrinr: It's a UI only thing. It is the interface that the svelte files will use to use it. I will rename it to match the equivalent usages in the other fetchers

mike12345567 (approved):
LGTM!

Issue #15193, opened 2024-12-17
Error: Unable to find table named


JBSDS:
Hi everyone,

I’m encountering an error on Budibase when creating an automation with a "Query row" and the table name contains multiple words. The error I get is: "Error: Unable to find table named 'Historique%2520anciennete%2520abo'".
The problem appeared when upgrading from version 2.32.11 to 3.2.29.

Does anyone have any ideas on how to resolve this issue?

Thank you in advance for your help!

`{
  "browser": {
    "language": "fr-FR",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36",
    "platform": "Win32",
    "vendor": "Google Inc."
  },
  "server": {
    "budibaseVersion": "3.2.29",
    "hosting": "docker-compose",
    "nodeVersion": "v20.18.1",
    "platform": "linux",
    "cpuArch": "x64",
    "cpuCores": 32,
    "cpuInfo": "13th Gen Intel(R) Core(TM) i9-13900K",
    "totalMemory": "31.22834014892578GB",
    "uptime": "0 day(s), 6 hour(s), 50 minute(s)"
  }
}`

![image](https://github.com/user-attachments/assets/b5070bad-596e-4fdb-8eb4-98059252cadb)

JBSDS:
Hi,
I just got the migration pop-up related to the CouchDB service. I went ahead and did it, and it resolved my issue.
https://docs.budibase.com/docs/migrations

JBSDS:
Hi, I obviously spoke a bit too soon. I just noticed that I have the same issue with other automations 😢.
![image](https://github.com/user-attachments/assets/f64eb32d-a6a0-440e-818e-564c74646765)

JBSDS:
Hi,

Here's the workaround I used to solve this issue: I ended up recreating the entire database from scratch. However, I ran into another problem when trying to re-import the CSV files into my database using Budibase. 
Specifically, I noticed that the relationship data from my MySQL database wasn't importing.

To address this, I used this solution: I converted all "fk_" columns related to relationships into "arrays", as that is the format required to import relationships via CSV. I'm not sure why this isn't the default export format, and I couldn’t find an option to handle this directly.

Here's what I did:

1 - I renamed the "fk_" columns in my CSV to match the names of the corresponding relationship columns.
[...]

PR #15346, opened 2025-01-10
Fix query rows step when querying a table containing spaces in the name

mike12345567:
## Description
Updating query row to correctly decode the input table Id as it will be encoded. The CTX expects this to be passed in its decoded state as Koa would do to encoded URLs.

## Addresses
- https://linear.app/budibase/issue/BUDI-8930/error-unable-to-find-table-named
- https://github.com/Budibase/budibase/issues/15193


packages/server/src/automations/steps/queryRows.ts
```diff
@@ -122,9 +122,10 @@ export async function run({
     sortType =
       fieldType === FieldType.NUMBER ? FieldType.NUMBER : FieldType.STRING
   }
+  // when passing the tableId in the Ctx it needs to be decoded
   const ctx = buildCtx(appId, null, {
     params: {
-      tableId,
+      tableId: decodeURIComponent(tableId),
```
samwho, line 128: I'm confused as to why spaces in the _name_ of the table would affect the- ohhhh the ID for external tables encodes the name within it doesn't it?

samwho (approved):

packages/server/src/automations/steps/queryRows.ts
mike12345567: Yeap exactly this - external tables the ID is built from the table name today.

PR #15353, opened 2025-01-13
Type client utils

adrinr:
## Description
Type client utils


packages/client/src/sdk.js
```diff
@@ -74,7 +73,6 @@ export default {
   fetchData,
   QueryUtils,
   ContextScopes: Constants.ContextScopes,
-  getAPIKey,
```
aptkingston, line 77: This is unused in our code, but the SDK is provided to plugin authors. This was added to allow them to authenticate with the public API more easily, so it needs to stay.

packages/client/vite.config.mjs
```diff
@@ -67,6 +67,10 @@ export default defineConfig(({ mode }) => {
           find: "constants",
           replacement: path.resolve("./src/constants"),
         },
+        {
+          find: "@/constants",
+          replacement: path.resolve("./src/constants"),
+        },
```
aptkingston, line 73: You can replace this with the catch-all `"@": path.resolve(__dirname, "src")`, as long as you put it after the `@budibase/xxx` aliases below.

aptkingston (changes requested):
LGTM but we just need to keep `getAPIKey` in the SDK for plugin authors 👍

packages/client/src/sdk.js
adrinr: Great, I added a comment to avoid future removals

packages/client/vite.config.mjs
adrinr: I did try that initially, but getting the following error:
image

aptkingston: Very interesting. I wonder what the difference is between this and the builder package, since we use that there.

aptkingston (approved):

PR #15354, opened 2025-01-13
Typing client api patches overrides

adrinr:
## Description
Typing client api patches overrides. We had a few bugs coming from it, so this should make refactors safer!


packages/client/src/api/patches.ts
```diff
@@ -90,13 +91,14 @@ export const patchAPI = API => {
     return await enrichRows(rows, tableId)
   }
 
-  // Wipe any HBS formulae from table definitions, as these interfere with
+  // Wipe any HBS formulas from table definitions, as these interfere with
```
aptkingston, line 94: NAB: this is also correct English - both are correct plurals of formula. Just in case you didn't know!

packages/client/src/api/patches.ts
```diff
@@ -90,13 +91,14 @@ export const patchAPI = API => {
...
   // handlebars enrichment
   const fetchTableDefinition = API.fetchTableDefinition
   API.fetchTableDefinition = async tableId => {
     const definition = await fetchTableDefinition(tableId)
     Object.keys(definition?.schema || {}).forEach(field => {
       if (definition.schema[field]?.type === "formula") {
+        // @ts-expect-error TODO check what use case removing that would break
         delete definition.schema[field].formula
```
aptkingston, line 102: Judging by the comment, I'd say this was added because formula column schemas can contain bindings in plaintext - e.g. `{{ user.name }}` etc, and therefore if the table schema was attached to an object which was later fully recursively processed by string-templates, as we do with our component settings, it would end up attempting to processing these bindings inside formula column definitions and fail.

That would end up mutating the definition of the table, which I assume could then be erroneously sent up later and saved.

This was 3 years ago so who knows if that is even still a problem, but that's my guess.

aptkingston (approved):
LGTM!

packages/client/src/api/patches.ts
adrinr: TIL that. I had no idea, thx :)

PR #15358, opened 2025-01-13
Add tests to ensure date sorting works, and you can't bulk import dates in an invalid format.

samwho:
## Description

This PR makes sure that date-only fields sort correctly by adding tests, and it makes sure that bulk importing rows cannot introduce invalid data into datetime fields.

## Launchcontrol

- Bulk importing rows (e.g. from a CSV) now requires dates to be formatted in a strict way, e.g. date-times should be YYYY-MM-DDTHH:MM:SSZ, date-only should be YYYY-MM-DD, and time-only should be HH:MM:SS.


adrinr (approved):
Nice to see improvement and tests around that area! It's always pretty confusing...

PR #15363, opened 2025-01-14
Type fetch data usages

adrinr:
## Description
Type fetch data usages


packages/frontend-core/src/fetch/DataFetch.ts
```diff
@@ -44,22 +44,31 @@ interface DataFetchDerivedStore<TDefinition, TQuery>
...
-> {
+export interface DataFetchParams<TDatasource, TQuery = SearchFilters> {
   API: APIClient
   datasource: TDatasource
   query: TQuery
-  options?: {}
+  options?: Partial<DataFetchOptions<TQuery>>
+}
+
+export interface DataFetchOptions<TQuery = SearchFilters> {
```
mike12345567, line 54: since this is exported, would it be beneficial to put this in the types package?

packages/frontend-core/src/fetch/index.ts
```diff
@@ -1,19 +1,30 @@
...
+import RelationshipFetch, { RelationshipDatasource } from "./RelationshipFetch"
+import NestedProviderFetch, {
+  NestedProviderDatasource,
+} from "./NestedProviderFetch"
+import FieldFetch, { FieldDatasource } from "./FieldFetch"
 import JSONArrayFetch from "./JSONArrayFetch"
-import UserFetch from "./UserFetch"
-import GroupUserFetch from "./GroupUserFetch"
-import CustomFetch from "./CustomFetch"
+import UserFetch, { UserDatasource } from "./UserFetch"
```
mike12345567, line 11: This was my previous point - its nice to just have these in types instead of needing to export like this.

mike12345567 (approved):
LGTM - just a comment about how we handle typing like this - I've been trying to expand the UI typing in types, I think it saves some future effort when trying to import these.

PR #15364, opened 2025-01-14
Re-add tests files in backend-core/dist

adrinr:
## Description
After a monorepo refactor (https://github.com/Budibase/budibase/pull/15128) we stopped adding the tests files in the backend-core npm package. These are used in places such as account-portal, and they might be used somewhere else.
This PR adds this files explicitly, to fix it avoiding to be oversighted again.

Updated dist folder:
image

Matches the working version:
image

## Launchcontrol
Re-add test files in backend-core/dist


melohagan (approved):
Thank you!

PR #15366, opened 2025-01-14
Create a new automation step for executing JS

samwho:
## Description

In the process of upgrading the development experience of writing JavaScript in automations, we want to port improvements we've made in the JS experience in the designer: live evaluation, access to helpers, access to snippets, showing test values in the bindings drawer.

Unfortunately one problem with all this is that the way JavaScript gets executed on the backend for the existing JavaScript step in automations differs quite a lot to JavaScript elsewhere in Budibase. For example, there is no `$("foo")` syntax in automations to access the execution context, you would just say `foo`. 

To bring automation JS in line with the rest of the app, we're introducing a new JavaScript execution step and deprecating the old one. Existing old ones will continue to work, but you will not be able to create new instances of the old JavaScript step.

Code written for the old JS step, e.g.:

```javascript
return trigger.row.name;
```

Will no longer work in the new one, it will need to be translated to:

```javascript
return $("trigger.row.name");
```

## Launchcontrol


deanhannigan (approved):
LGTM! Nicely done! I've been running this locally and it's handling both `ExecuteScript` and `ExecuteScriptV2` steps as expected. 🎉

deanhannigan:
Hold off on merging for the moment. There seems to be an issue now on the child branch around live evaluation.

mike12345567 (approved):

PR #15372, opened 2025-01-15
`BindingPanel` typing improvements

mike12345567:
## Description
As part of the debugging work I've started by typing some of the components that will need updated to support the JS logging functionality - this includes `BindingPanel.svelte`, `EvaluationSidePanel.svelte` and `CodeEditor.svelte` - there is a few other bits that needed updated but these are the main components focused on here.


packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -71,7 +73,7 @@
 
   $: {
     if (autofocus && isEditorInitialised) {
-      editor.focus()
+      editor?.focus()
```
adrinr, line 76: This should not be required if you remove the ` | undefined` from `let textarea: HTMLDivElement`

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -100,35 +102,47 @@
...
 
-    if (editor.state.doc.toString() !== value || queuedRefresh) {
+    if (
+      editor &&
+      value &&
+      (editor.state.doc.toString() !== value || queuedRefresh)
+    ) {
       editor.dispatch({
-        changes: { from: 0, to: editor.state.doc.length, insert: value },
+        changes: { from: 0, to: editor.state.doc.length, insert: value! },
```
adrinr, line 122:
```suggestion
        changes: { from: 0, to: editor.state.doc.length, insert: value },
```

packages/builder/src/components/common/bindings/EvaluationSidePanel.svelte
```diff
@@ -45,11 +48,11 @@
   }
 
   const copy = () => {
-    let clipboardVal = expressionResult.result
+    let clipboardVal = expressionResult
```
adrinr, line 51: What that broken, having the `.result`?

packages/frontend-core/src/utils/jsonFormatter.ts
```diff
@@ -0,0 +1,71 @@
...
+const defaultColors: ColorsOptions = {
+  keyColor: "dimgray",
+  numberColor: "lightskyblue",
+  stringColor: "lightcoral",
+  trueColor: "lightseagreen",
+  falseColor: "#f66578",
+  nullColor: "cornflowerblue",
+}
+
+const entityMap: Record<string, string> = {
```
adrinr, line 21: It might be better not having the `Record<string, string>`, so the type gets properly typed automatically

packages/builder/src/components/common/bindings/EvaluationSidePanel.svelte
mike12345567: It doesn't seem to ever actually be passed in like this - this is actually the reason why I started the process of converting - I couldn't see where the `.result` was coming from.

adrinr (approved):

PR #15373, opened 2025-01-15
Remove old migration system.

samwho:
## Description

This PR removes the old global migration system.

As far as I was able to tell, this is not used anymore. We have a newer migration system that works per-app, and confusingly the endpoint to check the status of these new migrations was intermingled with endpoints that operated on the old migration system.

## Launchcontrol

- Removes an old migration system. The following internal endpoints will no longer be present: 
  - `POST /api/migrations/run`
  - `GET /api/migrations/status`
  - `POST /api/system/migrations/run`
  - `GET /api/system/migrations/definitions`

## Feature branch env
[Feature Branch Link](http://account-fb-remove-dead-code.fb.qa.budibase.net)


adrinr (commented):
Not 100% sure how I feel about this. What would happen if somebody tries to update a really old installation? Will we simply not support it, as it's probably an unlikely use case? @mike12345567

adrinr (approved):
Thanks for all the checks!

PR #15383, opened 2025-01-16
Extract automation step definitions out into shared-core.

samwho:
## Description

This is so that we can make use of them in the frontend to generate data in tests.

## Addresses
- https://linear.app/budibase/issue/PRO-279/pull-automation-definitions-into-shared-core


packages/shared-core/src/automations/steps/bash.ts
```diff
@@ -0,0 +1,47 @@
...
+  description: "Run a bash script",
+  type: AutomationStepType.ACTION,
+  internal: true,
+  features: {
+    [AutomationFeature.LOOPING]: true,
+  },
+  stepId: AutomationActionStepId.EXECUTE_BASH,
+  inputs: {},
+  schema: {
+    inputs: {
```
adrinr, line 23: Would not be enough moving inputs and outputs? Will we be using anything else in the frontend?

samwho: cc @deanhannigan, I believe the big win from this would be testing.

adrinr (approved):

PR #15384, opened 2025-01-17
Typing some components and returning 404 for views not found

adrinr:
## Description
As a ground work for the FormBlock data issues I typed some of the used components to typescript. At the same time, I realised that requesting for non-existing views (for a deleted one, for example) would return 500 instead of a 404. Updated it to be more REST compliant. (ideally should be returning a 410, but we don't have this information at the m

## Launchcontrol
/get/view/:id returns 404 if not found


packages/client/src/components/app/blocks/form/FormBlock.svelte
```diff
@@ -1,37 +1,44 @@
...
+  export let showSaveButton: boolean
+  export let saveButtonLabel: boolean
+  export let deleteButtonLabel: boolean
 
-  const { fetchDatasourceSchema, generateGoldenSample } = getContext("sdk")
-  const component = getContext("component")
-  const context = getContext("context")
+  const { fetchDatasourceSchema, generateGoldenSample } = getContext<SDK>("sdk")
+  const component = getContext<Component>("component")
+  const context = getContext<Context>("context")
```
samwho, line 39: I wonder if it's worth us wrapping this with something we can overload to not require these type parameters, e.g.

```typescript
function getContext(name: "sdk"): SDK;
function getContext(name: "component"): Component;
function getContext(name: "context"): Context;
function getContext(name: string) {
 // ...
}
```

samwho (approved):

packages/client/src/components/app/blocks/form/FormBlock.svelte
adrinr: That's a magnificent idea! Checking if it's easy to do with svelte

PR #15385, opened 2025-01-17
BBUI TS scaffolding + helpers conversion

aptkingston:
## Description
This PR adds the required config so that BBUI can properly build typescript code in both svelte components and .ts files.

I've converted the helpers and the Heading component to TS as an example of both.

I've also added a `dev` script so that `bbui` is continually built in dev, highlighting any build issues which may otherwise go unnoticed.


packages/bbui/src/Typography/Heading.svelte
```diff
@@ -2,10 +2,10 @@
   import "@spectrum-css/typography/dist/index-vars.css"
 
   // Sizes
-  export let size = "M"
-  export let textAlign = undefined
-  export let noPadding = false
-  export let weight = "default" // light, heavy, default
+  export let size: "XS" | "S" | "M" | "L" = "M"
+  export let textAlign: string | undefined = undefined
```
adrinr, line 6: I think that the `| undefined` is not required, as it gets inferred by the `= undefined`

adrinr (approved):
Thanks for that!

packages/bbui/src/Typography/Heading.svelte
aptkingston: You got me hopeful as I have done this in loads of places and would love to remove it!

But it does seem required. If I remove it:
image

adrinr: Mm.. I just used in here: https://github.com/Budibase/budibase/pull/15384/files#diff-615c4b1d1bda62d5c443c2b06e9226cb16b173146419c01e458d72d1dafe7383

adrinr: Oh, I see. It's caused to adding the default tsconfig. The `strictNullChecks` is the culprit

aptkingston: Ah I think I understand the difference. In this case `textAlign` actually is optional, so I want to type is as possibly undefined. If i change `strictNullChecks` to false then it lets me set it as undefined, but it is typed as `string` only which is incorrect for this use case and potentially dangerous (unless all our code was TS, then it would be easy to spot any errors). 

I'll keep this as-is, because I think it's more correct for this use case, in that it actually is optional 👍

PR #15402, opened 2025-01-20
JS console log support

mike12345567:
## Description

This PR introduces support for `console.log` type functions when writing JS bindings. These will be presented as part of the preview - displayed as such:

image
error state: 
image

This will correctly handle `warn` and `error` - creating a visual distinction from other log types. This will also handle comma separated log statements and will correctly display different types (e.g. correctly highlighting JSON).

All logging is handled purely in the frontend, so while you can preview them for formulas or automation bindings (when preview is supported) they will not be logged on the backend currently.

Logs in the design section will carry through to the app currently, so that when previewing a real page you can check how bindings are evaluating (say in the preview of the app).

Note: At the moment `.table()` is supported, but does not get displayed any differently to a log statement (it is not fully supported currently).


packages/string-templates/src/helpers/javascript.ts
```diff
@@ -114,8 +120,49 @@ export function processJS(handlebars: string, context: any) {
...
     }
 
+    const logs: Log[] = []
+    // logging only supported on frontend
+    if (!isBackendService()) {
+      // this counts the lines in the wrapped JS *before* the user's code, so that we can minus it
+      const jsLineCount = frontendWrapJS(js).split(js)[0].split("\n").length
+      const buildLogResponse = (type: LogType) => {
+        return (...props: any[]) => {
+          if (!isTest()) {
```
adrinr, line 130: Why removing it for tests?

packages/string-templates/src/helpers/javascript.ts
```diff
@@ -114,8 +120,49 @@ export function processJS(handlebars: string, context: any) {
...
+          props.forEach((prop, index) => {
+            if (typeof prop === "object") {
+              props[index] = JSON.stringify(prop)
+            }
+          })
+          // quick way to find out what line this is being called from
+          // its an anonymous function and we look for the overall length to find the
+          // line number we care about (from the users function)
+          // JS stack traces are in the format function:line:column
+          const lineNumber = new Error().stack?.match(
```
adrinr, line 142: This is a really smart but strange way to get the line number 😅

packages/bbui/src/bbui.css
```diff
@@ -45,6 +45,11 @@
   --purple: #806fde;
   --purple-dark: #130080;
 
+  --error-bg: rgba(226, 109, 105, 0.3);
```
adrinr, line 48: This is slightly confusing, as it is not the background of the error. This is the error icon, right?

adrinr (commented):
Nice job!

joebudi:
Great work @mike12345567 . There are two arrows at the right of the screenshot that seem to be out of place.

packages/string-templates/src/helpers/javascript.ts
mike12345567: Jest seems to detect the error/warn and end the test suite based on this - doesn't happen locally but seems to happen in CI

packages/bbui/src/bbui.css
mike12345567: No this is actually a background in this case, there is an error background for when there's a literal JS error - I just gave it a specific colour that fits inline with the icon colours

adrinr (approved):

PR #15403, opened 2025-01-20
Bindings explorer panel

aptkingston:
## Description
This PR adds a new panel to explore available bindings for the selected component.

Also contains typescriptification of a few BBUI components, BBUI utils and builder components.

image
image

Example of the colours of all possible data types:
image

You can dig in to all the available bindings, including going deep into complex structures - letting you see the full extent of the data available to you.

Interactions:
- You can expand or collapse an object or array by clicking on either the arrow or the name of the value
- Values are initially only 1 line. You can expand long values to reveal their full extent by clicking:
image
image
- When hovering over any binding, a copy icon will appear on the right side. Clicking this icon will copy the binding to the clipboard (e.g. `{{ Data provider.Rows.0.Price }}`. This can then be pasted into a binding drawer for example.

Worth noting that this change also removes the title and add component button from the components tab. The only way to add components is now using the add component FAB in the preview (or ctrl/cmd + enter).
image

## Addresses
- https://linear.app/budibase/issue/BUDI-8979/create-bindings-data-context-panel

## Feature branch env
[Feature Branch Link](http://fb-bindings-panel.fb.qa.budibase.net)


packages/bbui/src/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+  Bottom = "bottom",
+  Left = "left",
+}
+
+export enum TooltipType {
+  Default = "default",
+  Info = "info",
+  Positive = "positive",
+  Negative = "negative",
+}
```
aptkingston, line 23: We need to export enums from `.ts` file to avoid linting errors. There's a configuration issue with eslint, typescript and svelte right now that will cause linting errors if exporting from a svelte component directly. This is a clean solution.

PClmnt (approved):
Pulled this down and had a play with it, and it's such an improvement to the app building experience, great stuff! LGTM

PR #15406, opened 2025-01-20
login platform user after forced password reset

melohagan:
## Description
Small bug fix relating to platform users with generated passwords. Originally the user would be logged out after changing their password, not they are automatically logged in. 

## Addresses
- https://linear.app/budibase/issue/GRO-822/when-a-user-is-invited-and-they-reset-password-auto-login

## Screenshots
https://github.com/user-attachments/assets/52577806-90c4-4fd4-accf-1ecf4c882840


jvcalderon (approved):
LGTM

PR #15407, opened 2025-01-20
Improve typing around automations and automation testing

samwho:
## Description

This started life as a PR to implement a new endpoint for testing individual steps in an automation but we've decided not to go with that approach. While I was doing that work, I had done some typing improvements in the area that I didn't want to lose so I've created this PR to get those merged in.

I've left a bunch of comments inline to explain each of the changes that I've done.


packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -398,21 +398,19 @@ class AutomationBuilder extends BaseStepBuilder {
 
   async run() {
     const automation = await this.save()
-    const results = await testAutomation(
-      this.config,
-      automation,
-      this.triggerOutputs
+    const response = await this.config.api.automation.test(
+      automation._id!,
+      this.triggerOutputs as TestAutomationRequest
```
samwho, line 403: I don't like having to do this cast but `TriggerOutputs` isn't compatible with `TestAutomationRequest`. This is a problem that we should rectify, but I didn't get to it in this PR.

packages/server/src/api/controllers/automation.ts
```diff
@@ -231,24 +231,25 @@ export async function test(
...
-    occurredAt: new Date().getTime(),
+  const automation = await db.tryGet<Automation>(ctx.params.id)
+  if (!automation) {
+    ctx.throw(404, `Automation ${ctx.params.id} not found`)
+  }
+
+  const { request, appId } = ctx
+  const { body } = request
+
+  ctx.body = await withTestFlag(automation._id!, async () => {
```
samwho, line 242: Rolled up `setTestFlag` and `clearTestFlag` into this `withTestFlag` helper that correctly handles what happens when an exception is thrown. While I was here, I did a little refactoring to make this handler easier to read.

packages/server/src/automations/triggers.ts
```diff
@@ -156,14 +149,26 @@ export function isAutomationResults(
...
+export async function externalTrigger(
+  automation: Automation,
+  params: AutomationTriggerParams,
+  options: { getResponses: true }
+): Promise<AutomationResults | DidNotTriggerResponse>
+export async function externalTrigger(
+  automation: Automation,
+  params: AutomationTriggerParams,
+  options?: { getResponses: false }
+): Promise<AutomationJob | DidNotTriggerResponse>
```
samwho, line 168: This function either queues an `AutomationJob`, or runs an automation synchronously and returns an `AutomationResults` based on what the value of `options.getResponses` is. These overloads narrow down these types to help the testing endpoint know it has hold of an `AutomationResults` object, never actually has to deal with the `AutomationJob` type.

packages/server/src/automations/steps/createRow.ts
```diff
@@ -15,7 +18,7 @@ export async function run({
 }: {
   inputs: CreateRowStepInputs
   appId: string
-  emitter: EventEmitter
+  emitter: ContextEmitter
```
samwho, line 21: All of the automation action handlers wanted an `EventEmitter` as an argument. Digging down into why, I found that they actually _don't_ want an `EventEmitter` at all. This `emitter` is passed down to eventually go into a function called `buildCtx`, which creates a fake `UserCtx` object for use by action handlers to directly call controller endpoint functions. `UserCtx.eventEmitter` is of type `ContextEmitter`, so I've updated these types to match that.

packages/server/src/automations/triggers.ts
```diff
@@ -33,14 +34,6 @@ const JOB_OPTS = {
 import * as automationUtils from "../automations/automationUtils"
 import { doesTableExist } from "../sdk/app/tables/getters"
 
-type DidNotTriggerResponse = {
-  outputs: {
-    success: false
-    status: AutomationStatus.STOPPED
-  }
-  message: AutomationStoppedReason.TRIGGER_FILTER_NOT_MET
-}
```
samwho, line 42: This got moved to `@budibase/types` because I wanted to use it in the `TestAutomationResponse` type.

packages/server/src/utilities/index.ts
```diff
@@ -58,57 +58,32 @@ export function checkSlashesInUrl(url: string) {
 export async function updateEntityMetadata(
   type: string,
   entityId: string,
-  updateFn: any
+  updateFn: (metadata: Document) => Document
```
samwho, line 61: This function is called by the test automation endpoint and I noticed this any and wanted to get rid of it. While here, I did some light refactoring to make these functions easier to follow.

mike12345567 (approved):
LGTM - good to capture some of these types a bit better and also good to know the test/trigger outputs aren't compatible - something we should probably sort out.

PR #15408, opened 2025-01-20
Convert history store

adrinr:
## Description
Converting history store to ts. This required some utils to be typed as well.


packages/builder/src/stores/builder/screens.ts
```diff
@@ -138,13 +148,6 @@ export class ScreenStore extends BudiStore {
       }
 
       const definition = componentStore.getDefinition(component._component)
-      // Reset whitelist for direct children
-      legalDirectChildren = []
```
adrinr, line 142: I cannot see this being referenced anywhere. I assume is safe to remove. Could you confirm, @aptkingston ?

aptkingston:
Dean already has a PR up for the screen store, so I'll let you 2 work out what's the best way to go haha. https://github.com/Budibase/budibase/pull/15313

adrinr:
@deanhannigan @aptkingston my PR also typed the history store. I resolved the conflicts after @deanhannigan's PR, so there is only the history store left

packages/frontend-core/src/utils/utils.ts
```diff
@@ -45,9 +54,9 @@ export const sequential = fn => {
  * @param minDelay the minimum delay between invocations
  * @returns a debounced version of the callback
  */
-export const debounce = (callback, minDelay = 1000) => {
-  let timeout
-  return async (...params) => {
+export const debounce = (callback: Function, minDelay = 1000) => {
+  let timeout: NodeJS.Timeout
```
aptkingston, line 58: NAB: Since this normally runs in the browser, this will actually just be a number. Sam and I ran into this same problem, where TS thinks this is running in a node env rather than a browser one.

The solution I've been using (until we find a proper fix to let TS know the correct environment) is just type this as `ReturnType<typeof setTimeout>`.

(I don't think this really affects anything here, but it's just more correct).

aptkingston (approved):
LGTM!

PR #15413, opened 2025-01-21
Aligning the way that tableselector and datasourceselector display

adrinr:
## Description
We have two setting components that allow users to select datasources from a dropdown. One allows only tables and views, while the other allows other types such as queries, custom sources, etc.

These designs currently don't match:
| Datasource select | Table select |
|--------|--------|
| image|image|

This PR aligns both components to be lookalike, as well as adding some enhances and fixes:
1. Prefixing views with the datasource name
2. Fixing a bug that caused views with the same name to be wrongly marked as "selected"
image
3. Do not prefix at all if there is only a datasource (to keep it simple)
image

## Launchcontrol
Enhancing the datasource picker


aptkingston (approved):
LGTM!

PR #15414, opened 2025-01-21
Validate datasource setting on components

adrinr:
## Description
Validate "table" settings on components. This is used in the components `form block`, `multistep form block` and `row explorer`. If the set table/view does not exist anymore, we will display a friendly error instead of a confusing error.

### Previous 
image

### New behaviour
image

## Launchcontrol
Improve verbosity for wrong datasource settings values


mike12345567 (approved):
This LGTM - but I'll leave @aptkingston to comment as well - given it crosses the builder/client boundary!

packages/builder/src/stores/builder/screenComponent.ts
```diff
@@ -0,0 +1,80 @@
...
+) {
+  const friendlyNameByType = {
+    table: "table",
+    view: "view",
+    viewV2: "view",
+  }
+
+  const result: Record<string, string[]> = {}
+  for (const component of findComponentsBySettingsType(screen, "table")) {
+    const { resourceId, type, label } = component.dataSource
```
aptkingston, line 69: NAB: This makes the assumption that all settings of type `table` are stored under the key `dataSource`. I know that is true of the manifest today, but it's not something we can guarantee as keys are otherwise arbitrary. This also may not account for custom components which may use other keys too.

The complete solution is more complicated because you'd have to return both the component and key for set setting you find. It's not super important to do this as the current manifest satisfies this constraint, but worth keeping in mind I guess.

packages/builder/src/helpers/screen.ts
```diff
@@ -0,0 +1,30 @@
...
+      result.push(component)
+    }
+    component._children?.forEach(child => {
+      recurseFieldComponentsInChildren(child, type)
+    })
+  }
+
+  recurseFieldComponentsInChildren(screen?.props, type)
+  return result
+}
```
aptkingston, line 30: This doesn't account for custom components, which are stored separately, but that's probably fine. Just as long as it's a known limitation

packages/builder/src/stores/builder/screenComponent.ts
```diff
@@ -0,0 +1,80 @@
...
+          return {
+            errors: getInvalidDatasources($selectedScreen, datasources),
+          }
+        }
+      )
+    }
+
+    super({}, makeDerivedStore)
+  }
+}
```
aptkingston, line 34: NAB: Fair play for making this a full derived BudiStore, but since this is only deriving data (not storing any) and doesn't have any actions, I would probably just make this a native derived store to keep things easier. It might also be better to just name this `componentErrors` or something and make it a `Record<string, string[]>` directly rather than object - again just keeping things simple. `ScreenComponentStore` is a little generic and not obvious what it does, but I know you're probably just thinking ahead of other use cases for it.

packages/client/src/components/Component.svelte
```diff
@@ -368,6 +374,9 @@
     }
   }
 
+  // Check for invalid settings
+  $: invalidSettings = $componentErrors[id]
```
aptkingston, line 378: Could we hoist this up a bit to be beside the other reactive statements? This component is a monster and I find it much easier to read with all the reactive logic in one place 🙏

aptkingston (commented):
Nice work @adrinr! Glad this approach worked out 👌 Mostly just a few opinions here - nothing major other than just being aware that this currently wouldn't support component plugins, which is probably fine.

adrinr:
Found an issue that causes the component to render to a frame with the wrong data, calling the API with the invalid data. This displays an unexpected error notification. Investigating why

packages/builder/src/helpers/screen.ts
adrinr: I will create a ticket for a "nice to have", and to tackle this if we have some extra time

adrinr:
> Found an issue that causes the component to render to a frame with the wrong data, calling the API with the invalid data. This displays an unexpected error notification. Investigating why

Solved

aptkingston (approved):
LGTM!

PR #15425, opened 2025-01-22
Budi 9006 allow automation test data modal to select real data to

samwho:
## Description

## Addresses
- ...more if required

## App Export

## Screenshots

## Launchcontrol

### DIFF
--- a/packages/bbui/src/Form/Core/Checkbox.svelte
+++ b/packages/bbui/src/Form/Core/Checkbox.svelte
@@ -1,22 +1,23 @@
-<script>
+<script lang="ts">
   import "@spectrum-css/checkbox/dist/index-vars.css"
   import "@spectrum-css/fieldgroup/dist/index-vars.css"
   import { createEventDispatcher } from "svelte"
+  import { ChangeEventHandler } from "svelte/elements"
 
   export let value = false
-  export let id = null
-  export let text = null
+  export let id: string | undefined = undefined
+  export let text: string | undefined = undefined
   export let disabled = false
   export let readonly = false
-  export let size
+  export let size: string = "M"
   export let indeterminate = false
 
   const dispatch = createEventDispatcher()
-  const onChange = event => {
-    dispatch("change", event.target.checked)
+  const onChange: ChangeEventHandler<HTMLInputElement> = event => {
+    dispatch("change", event.currentTarget.checked)
   }
 
-  $: sizeClass = `spectrum-Checkbox--size${size || "M"}`
+  $: sizeClass = `spectrum-Checkbox--size${size}`
 </script>
 
 <label
--- a/packages/bbui/src/Form/Core/Checkbox.svelte
+++ b/packages/bbui/src/Form/Core/Checkbox.svelte
@@ -2,7 +2,7 @@
   import "@spectrum-css/checkbox/dist/index-vars.css"
   import "@spectrum-css/fieldgroup/dist/index-vars.css"
   import { createEventDispatcher } from "svelte"
-  import { ChangeEventHandler } from "svelte/elements"
+  import type { ChangeEventHandler } from "svelte/elements"
 
   export let value = false
   export let id: string | undefined = undefined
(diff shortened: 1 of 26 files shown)


packages/bbui/src/Form/Core/Checkbox.svelte
```diff
@@ -1,22 +1,23 @@
...
   export let readonly = false
-  export let size
+  export let size: string = "M"
   export let indeterminate = false
 
   const dispatch = createEventDispatcher()
-  const onChange = event => {
-    dispatch("change", event.target.checked)
+  const onChange: ChangeEventHandler<HTMLInputElement> = event => {
+    dispatch("change", event.currentTarget.checked)
```
aptkingston, line 17: `currentTarget` can differ from `target` depending on how listeners are set up and if the event is bubbling. In this case they should be the same as the element that triggers the event (`target`) is the same as the element which the event listener is attached to (`currentTarget`), as they should both be the input element. But just something to keep in mind in future as this could be a breaking change if we were manually adding the listener at a higher level (like on the full document).

packages/bbui/src/Form/Core/Checkbox.svelte
```diff
@@ -1,22 +1,23 @@
...
 
   export let value = false
-  export let id = null
-  export let text = null
+  export let id: string | undefined = undefined
+  export let text: string | undefined = undefined
   export let disabled = false
   export let readonly = false
-  export let size
+  export let size: string = "M"
```
aptkingston, line 12: You can probably take this further and type this as `"S" | "M" | "L"`. Not sure if XS and XL are valid options here but they might be as well - would need to check the CSS file imported at the top here.

packages/bbui/src/Form/Core/TextArea.svelte
```diff
@@ -1,33 +1,38 @@
...
 
   let focus = false
-  let textarea
+  let textarea: HTMLTextAreaElement
   const dispatch = createEventDispatcher()
-  const onChange = event => {
-    dispatch("change", event.target.value)
+  const onChange = (event: FocusEvent) => {
+    const target: HTMLTextAreaElement | null =
+      event.target as HTMLTextAreaElement
```
aptkingston, line 23: I've found the typing around events to be pretty awful to work with. Target is not actually nullable - it will always exist, so I usually just forcefully cast this to be an element type.

packages/frontend-core/src/utils/utils.ts
```diff
@@ -1,36 +1,34 @@
...
-export const sequential = fn => {
-  let queue = []
-  return (...params) => {
-    return new Promise((resolve, reject) => {
+export const sequential = <R, T extends (...args: any[]) => Promise<R>>(
+  fn: T
+): ((...params: Parameters<T>) => Promise<R>) => {
+  let queue: (() => Promise<void>)[] = []
+  return (...params: Parameters<T>): Promise<R> => {
+    return new Promise<R>((resolve, reject) => {
```
aptkingston, line 18: I'm sure this was fun to type 😂 Nice one.

aptkingston (approved):
I'll let Dean comment on the files that contain automation-specific logic (`TestDataModal` and `AutomationBlockSetup`) as my knowledge of those isn't great, but all the BBUI and other builder changes look good to me.

packages/bbui/src/Form/Core/Checkbox.svelte
samwho: `.target` doesn't exist on the event type, which is the only reason I changed it.

packages/frontend-core/src/utils/utils.ts
samwho: I was readily admit that I got help from ChatGPT 😅

packages/bbui/src/Form/Core/Checkbox.svelte
aptkingston: This is sort of what I was alluding to in my comment below about the event types being difficult to work with - `target` 100% does exist in the real event, so I'm not sure why it doesn't in the type.

PR #15427, opened 2025-01-23
Fix view v1 on tables

adrinr:
## Description
Fixing broken tables for viewV1, probably broken during the typification of the data fetchers: https://github.com/Budibase/budibase/pull/15295/files

This PR fixes it, while ensuring that not extra calls are made during some initialisation of forms or similar (we had that case in the past with some changes)

## Launchcontrol
Fixing tables for viewV1


packages/frontend-core/src/components/grid/stores/datasource.ts
```diff
@@ -1,5 +1,3 @@
-// TODO: datasource and defitions are unions of the different implementations. At this point, the datasource does not know what type is being used, and the assignations will cause TS exceptions. Casting it "as any" for now. This should be fixed improving the type usages.
```
adrinr, line 1: All these TODOS are not needed anymore

packages/frontend-core/src/fetch/ViewFetch.ts
```diff
@@ -21,7 +21,7 @@ export default class ViewFetch extends BaseDataFetch<ViewV1Datasource, Table> {
 
   getSchema(definition: Table) {
     const { datasource } = this.options
-    return definition?.views?.[datasource.name]?.schema
+    return definition?.views?.[datasource?.name]?.schema
```
adrinr, line 24: This was causing errors when trying to fetch the data, as at some reactivity updates the datasource was undefined

aptkingston (approved):

PR #15431, opened 2025-01-23
Improving design section JS binding performance

mike12345567:
## Description
This PR introduces [vm-browserify](https://github.com/budibase/vm-browserify) as a direct dependency, rather than receiving this as part of the string-templates build process and a browser polyfill for the `vm` Node library.

I have forked the library to add an optimisation to re-use the iframe when processing JS - previously each JS binding on the page would add an iframe, run the JS and then destroy said iframe - from our testing this has some particularly large performance costs.

The difference can be seen here, when rendering a page with a repeater block, containing a headline with a `return 1` JS binding - this is for 1000 components.

Before the library update:
image

After the library update:
image

Here we can see the total time to prepare the data for rendering has gone from 32.8 seconds to 1.4 seconds - a massive improvement in performance.

We have also introduced support for the iframe sandbox capabilities that are now generally supported by browsers (not an option when the library was originally built) - information can be found here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

This restricts the JS binding from performing certain actions like opening a popover - we have limited it to `allow-scripts` and `allow-same-origin` sandboxing behaviours.

The re-use of the iframe did raise a question of global state management, however after much testing it was found that there is little difference, since the parent body is always accessible in the JS via `window.parent` - this means there has always been the potential of global state.


aptkingston (approved):
🚀 I love that reusing the iframe had such a big impact. Definitely opens the doors for introducing bindings into grids for things like custom columns, and also hopefully lets me simplify some client library logic, which was extremely heavily optimised (at the cost of complexity) to minimize binding evaluation at all costs.

PR #15432, opened 2025-01-23
Fix selecting 'old row' in row updated trigger.

samwho:
## Description

If you have an automation with the trigger "row updated" and you open the test data modal and select "old row", it currently does not work.

CleanShot 2025-01-23 at 17 34 19@2x

Was a simple missing null check, this PR fixes it.

## Launchcontrol

- Fixes a missing null check when selecting "old row" to populate test data for an automation run.


deanhannigan (approved):
LGTM 🔥

PR #15433, opened 2025-01-23
Validate datasource setting on tables

adrinr:
## Description
Following https://github.com/Budibase/budibase/pull/15414, validate the datasource picker value. The previous validation (TableSelect.svelte) only allowed selecting tables and views. Now, validating the `DataSourceSelect.svelte` that allow selecting the following.
- Tables
- Views v1
- Views v2
- Queries
- Relationships
- Fields
- JSON array fields
- Data providers
- Custom/CSV datasets

For this PR, we only validate the "backend" datasources: tables, views and queries. The other ones depend on some screen components, so it requires some extra work.

### Having the following valid configuration...
image
image

### when deleting, we get the following
image
image

## Before, it would look broken
image
image

## Launchcontrol
Validate settings for Datasource selector


mike12345567 (approved):
LGTM!

PR #15435, opened 2025-01-24
JS Console log UX improvements

mike12345567:
## Description
@aptkingston recommended some improvements to console logging experience:
1. Fix word wrapping to split in the middle of long words
2. Fix border/separator to look more correct
3. Get console logged JSON to be displayed correctly

Below you can see these improvements in action:
image


aptkingston (approved):
Awesome, nice work! This solves all the mentioned issues, but could you also either
- Remove the horizontal padding around each element
- Or make the items full width
Either of those changes would make the logged text line up with the title text "Preview" and I think look even better 👌

mike12345567:
@aptkingston - sorted!
image

PR #15439, opened 2025-01-27
Validate screen relationship datasource settings

adrinr:
## Description
Adding validation for relationship fields in datasource selectors. Follow up of https://github.com/Budibase/budibase/pull/15433

### Before
If a component was referencing an element it did not exist on the component chain, it would not complain. It would just return no rows.

https://github.com/user-attachments/assets/a3697f35-e2d8-43e7-8d45-4a2cbbf2f995

The same would happen if the relationship were dropped
image

### After
The component using the invalid datasource will complain (either because of a wrong component tree or a dropped relationship)

https://github.com/user-attachments/assets/931e3d78-65d9-4680-b92f-563bb1842a2c

image

## Launchcontrol
Validation relationship fields in datasource settings in components.

### DIFF
--- a/packages/builder/src/helpers/bindings.ts
+++ b/packages/builder/src/helpers/bindings.ts
@@ -1,0 +1,50 @@
+import { makePropSafe } from "@budibase/string-templates"
+import { UIBinding } from "@budibase/types"
+
+export function extractRelationships(bindings: UIBinding[]) {
+  return (
+    bindings
+      // Get only link bindings
+      .filter(x => x.fieldSchema?.type === "link")
+      // Filter out bindings provided by forms
+      .filter(x => !x.component?.endsWith("/form"))
+      .map(binding => {
+        const { providerId, readableBinding, fieldSchema } = binding || {}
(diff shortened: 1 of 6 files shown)


packages/builder/src/helpers/bindings.ts
```diff
@@ -0,0 +1,74 @@
+import { makePropSafe } from "@budibase/string-templates"
```
adrinr, line 1: This has been moved from `packages/builder/src/components/design/settings/controls/DataSourceSelect/DataSourceSelect.svelte`

packages/types/src/ui/bindings/binding.ts
```diff
@@ -24,3 +24,18 @@ export type InsertAtPositionFn = (_: {
   value: string
   cursor?: { anchor: number }
 }) => void
+
+export interface UIBinding {
```
deanhannigan, line 28: 🙏

deanhannigan (approved):
LGTM!

PR #15442, opened 2025-01-27
[BUDI-9006] Allow automation test data modal to select real data

samwho:
## Description

This PR implements the ability to select real row data in the automation testing modal.

## Addresses
- https://linear.app/budibase/issue/BUDI-9006/allow-automation-test-data-modal-to-select-real-data-to-populate-test

## Screenshots

https://github.com/user-attachments/assets/e746689d-7ff1-4b8c-83e4-e3036ba2a1b4

## Launchcontrol

- Allow the selection of real row data when testing automations with row-related triggers.

## Feature branch env
[Feature Branch Link](http://fb-budi-9006-allow-automation-test-data.fb.qa.budibase.net)


packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -140,6 +149,31 @@
       ? [hbAutocomplete([...bindingsToCompletions(bindings, codeMode)])]
       : []
 
+  $: fetch = tableId
+    ? fetchData({
```
aptkingston, line 153: It's probably fine to construct the fetch instance this way for this use case, but one thing to keep in mind is that if you construct a brand new fetch every time any of these properties changes it'll cause a wasted API call to fetch the datasource definition every time. Constructing a fetch instance is a little heavier than just updating the search or sorting parameters for example, which can be done by `fetch.update({ query, sortColumn })` etc.

A pattern I typically use would be to only construct a new fetch instance when the datasource (`tableId` in this case) changes, and keep it up to date using a new reactive statement. That usually looks something like:
```js
  $: fetch = createFetch(dataSource)
  $: fetch.update({
    query,
    sortColumn,
    sortOrder,
    limit,
  })

  const createFetch = datasource => {
    return fetchData({
      API,
      datasource,
      options: {
        query,
        sortColumn,
        sortOrder,
        limit,
      },
    })
  }
```
It's basically just taking more control over the reactive statement triggers in order to avoid some wasted API calls.

Having said all that... For this use case, which is relatively infrequent, I think it's probably totally fine to leave this as is. Just wanted to mention this as it may be relevant in future.

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
```diff
@@ -140,6 +149,31 @@
...
+          },
+          limit: 20,
+        },
+      })
+    : null
+  $: fetchedRows = $fetch?.rows
+  $: fetchLoading = $fetch?.loading
+  $: primaryDisplay = table?.primaryDisplay
+  $: rowSearchTerm = ""
+  $: selectedRow = null
```
aptkingston, line 175: If `rowSearchTerm` and `selectedRow` are simply normal variables that you're declaring here, it would be better to just declare them with `let` instead of a reactive statement. Functionally this will be the same, but a reactive statement like this is ambiguous as to whether it's declaring a new variable or assigning a value to an existing one (if you had previously declared it above). So for readability, I always prefer to only use reactive statements for logic that we actually want to be reactive. Declaring a variable via either a reactive statement or via `let` etc has no impact on whether other code can "react" to changes of those variables.

aptkingston (commented):
No issues, just a couple of small suggestions!

packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
samwho: Ah, I copied this from `RelationshipField.svelte`. It looks like the codebase has a few instances of this pattern dotted around.

aptkingston (approved):

PR #15447, opened 2025-01-27
Type Checkbox.svelte, CheckboxGroup.svelte, and Combobox.svelte.

samwho:
## Description

This PR converts a few bbui types to TypeScript. I'm trying to target components that are leaf nodes, or almost leaf nodes, as I think going bottom-up on this will be easier than top-down.

### DIFF
--- a/packages/bbui/src/Form/Core/CheckboxGroup.svelte
+++ b/packages/bbui/src/Form/Core/CheckboxGroup.svelte
@@ -1,3 +1,8 @@
+<script lang="ts" context="module">
+  type O = any
+  type V = any
+</script>
+
 <script lang="ts" generics="O, V">
   import "@spectrum-css/fieldgroup/dist/index-vars.css"
   import "@spectrum-css/radio/dist/index-vars.css"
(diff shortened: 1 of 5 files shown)


packages/bbui/src/Form/Core/CheckboxGroup.svelte
```diff
@@ -1,19 +1,24 @@
-<script>
+<script lang="ts" context="module">
+  type O = any
+  type V = any
+</script>
+
+<script lang="ts" generics="O, V">
```
aptkingston, line 6: TIL about the `generics` attribute here.

packages/bbui/src/Form/Core/CheckboxGroup.svelte
```diff
@@ -25,6 +30,8 @@
   }
 </script>
 
+import context from "src/context"
```
aptkingston, line 33: Not sure how this got here, but I don't think it should be!

aptkingston (approved):
LGTM!

packages/bbui/src/Form/Core/CheckboxGroup.svelte
samwho: Huh, neither do I. Almost certainly auto-added by VSCode accident. 🙈

packages/bbui/src/Form/Core/CheckboxGroup.svelte
samwho: Yeah it's cool!

Slightly wonky because to make eslint happy about it you have to add that extra script block (https://stackoverflow.com/a/78981916), but I'm optimistic that's the sort of thing that will be improved on over time.

PR #15453, opened 2025-01-28
Do a typing pass on automation.spec.ts

samwho:
## Description

This updates `automation.spec.ts` and a few surrounding files to make use of `config.api.automation` instead of the older `config.createAutomation` and friends.


packages/server/src/tests/utilities/api/index.ts
samwho, line 1: I just re-ordered the lines in this file to be alphabetical so it's a bit easier to scan through them.

mike12345567 (approved):
LGTM!

PR #15454, opened 2025-01-29
Fix adding component after undo

adrinr:
## Description
Fixing a small bug that broke adding a component after an undo action. In that state, the current selected element is empty, and this breaks adding a new component until a new component is selected. Based on the code, in this case we were meant to select the top level screen. This PR fixes it.

### Before

https://github.com/user-attachments/assets/871363c5-6bdf-403e-b4e3-4b90861215ca

### After

https://github.com/user-attachments/assets/c3950be9-f7c1-4011-8631-cb5108b7f511

## Launchcontrol
Fix adding a component immediately an undo action


deanhannigan (approved):
LGTM!

PR #15457, opened 2025-01-29
Table column "format" setting

aptkingston:
## Description
This PR adds a new setting to table columns to allow controlling the format of values displayed in each column. A simple example of this might be prefixing a numerical value with a currency symbol, or transforming a date into the relative time from now.

JavaScript is fully supported.

The format is configured as a normal bindable expression ("Format" in this screenshot - the last setting):
image

You have the entire row available to you to bind to, so you can not only reference values from the current cell, but for all row values.

Here's an example of adding a dollar sign to a numerical value, and the result in the table:
image
image

Another quick example to show JS works - just formatting a value as random numbers:
image
image

Adding the format setting to any column will render the value as plain text. This means if you have a relationship, multi-select, user or other complex data type and use a custom format setting, it will still expect a string to be returned. This is mostly because we can't guarantee that the user won't mess up the expected data shape. Some help text is included below the format setting to explain this. This is an edge case anyway - I don't expect this to be heavily used for types other than primitives.

## Launchcontrol
Add format setting to table columns.

## Feature branch env
[Feature Branch Link](http://fb-grid-column-formatting.fb.qa.budibase.net)


packages/frontend-core/src/components/grid/stores/rows.ts
```diff
@@ -114,26 +115,44 @@ export const createStores = (): RowStore => {
...
+          (map, column) => {
+            map[column.name] = column.format!(row)
+            return map
+          },
+          {}
+        )
+        return {
+          ...row,
+          ...relatedValues,
+          __formatted: formattedValues,
```
aptkingston, line 152: The "formatted" value of each cell is cached inside the enriched row. This is because there's no need to evaluate the formatted binding on every render on the cell - it's wasteful.

adrinr (commented):
If we have this controlled in the frontend, it might cause some confusion/issue with backend related code such as formulas or automations. Should this not be return already formatted from the backend?

shogunpurple (approved):
LGTM!

Issue #15449, opened 2025-01-28
MongoDB request fails in v3.3.1 - worked fine before


esellier:
**Checklist**
- [X ] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker
  - Budibase Version: 3.3.1

**Describe the bug**
A mongo DB request fails with a response : 
{"message":"Unexpected token '{' [<isolated-vm>:1:6]","status":400}

The query works fine if I remove the Transformer. Anything else than the default "return Data;" transformer will fail.

**To Reproduce**
Create query, add a custom transformer => Fails
Remove transformer => OK

ConorWebb96:
Hey @esellier,

I've tried to recreate this. 

I'm able to do this `return data.name` (just getting a specific field from an object). Is this also broken for you? Is there a specific piece of JS that you are using that is broken?

Are you using Docker Compose or Docker Single Image? If you let me know which one you are using, I'll try to recreate it on those as well.

esellier:
I tried on my own Macbook, with a new clone of the git repo and running in Docker (npm run dev:docker), it works fine.

esellier:
The problem is in query.ts:155, when executing the transformer in VM.
There is an exception :
`{
    "userScriptError": {},
    "code": "USER_SCRIPT_ERROR"
  }`
The script passed to vm.execute is OK.

mike12345567:
Hi @esellier - I've been trying to replicate this - but the only way I've been able to get the error is with invalid JS. In your example `data?.name` is not valid JS, the conditional access operator is not available in Node.

I've tried quite a few different examples in 3.3.3 and have not been able to recreate with any valid transformer - below is your example from Discord.

![Image](https://github.com/user-attachments/assets/f7cfeeb1-95fb-4c37-a607-0bd234f95076)

esellier:
Found the issue, and fixed.
in isolated-vm.ts:158, it should be : 
```        const result = ${iifeWrapper(code)}```
and 
```        const result = ${code}```

I no longer have the issue when I fix that in query.js (directly in docker).

esellier:
Well... not exactly.
The transformer is not executed, I may have missed something

esellier:
Extracted from /tmp/.budibase/systemlogs/budibase.log
[...]

PR #15371, opened 2025-01-15
Minimise npm projects

adrinr:
## Description
Minimise files on npm projects and don't publish pro sourcemap

PR #15460, opened 2025-01-30
Fix isolatedvm query issues

adrinr:
## Description
After introducing minifications for the esbuild bundles (https://github.com/Budibase/budibase/pull/15371) we broke some usages on the implementation for isolated-vm. Isolated-vm and bson required polyfilling some functions and classes, but this stopped working fine because the modification transformations.
This PR does as it follows:
1. Extract the polyfills to a static js that will be bundled in the isolated-vm instance
2. Remove the bridge functions (in favour of full polyfills). This adds extra security, and isolated-vm is not 100% isolated
3. Add some tests to cover the bson document transformations

## Launchcontrol
Fix issues with MongoDB and queries

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,3 +10,4 @@ packages/builder/.routify
 packages/sdk/sdk
 packages/pro/coverage
 **/*.ivm.bundle.js
+!**/bson-polyfills.ivm.bundle.js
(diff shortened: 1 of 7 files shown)


mike12345567 (approved):
LGTM - very nicely done, great to have a test case for this edge case.

adrinr:
This one should resolve the issue https://github.com/Budibase/budibase/issues/15449

PR #15464, opened 2025-01-30
Fixes to address state usage misses

PClmnt:
## Description
Just adds some new code to better traverse the component tree to find properties that might contain state. Examples that now work

e.g this button group object

```
{
    "_id": "ce1b3eec8e51e4082ba0dea325eca599a",
    "_component": "@budibase/standard-components/buttongroup",
    "buttons": [
        {
            "_id": "cfcfc9e21bc184659b342dae92c0da0df",
            "_component": "@budibase/standard-components/button",
            "_styles": {
                "normal": {},
                "hover": {},
                "active": {}
            },
            "_instanceName": "ce9d0a50340e247c5a9c38b2259c3f79b",
            "text": "Button 1",
            "type": "cta",
            "size": "M",
            "gap": "M",
            "_rootId": "ce1b3eec8e51e4082ba0dea325eca599a",
            "onClick": [
                {
                    "parameters": {
                        "field": "bvbb",
                        "type": "set",
                        "value": "{{ [state].[test] }}"
                    },
                    "##eventHandlerType": "Update Field Value",
                    "id": "AHEtAQB6h"
                }
            ]
        },
        {
            "_id": "c0bc31ba96cfa4499af11a7ef7fcd3033",
            "_component": "@budibase/standard-components/button",
            "_styles": {
                "normal": {},
                "hover": {},
                "active": {}
            },
            "_instanceName": "c70a3357d081c4d6b8bdbe220e9f0fae2",
            "text": "Button 2",
            "type": "primary",
            "size": "M",
            "gap": "M",
            "_rootId": "ce1b3eec8e51e4082ba0dea325eca599a",
            "onClick": [
                {
                    "parameters": {
                        "key": "in group",
                        "type": "set",
                        "value": "v"
                    },
                    "##eventHandlerType": "Update State",
                    "id": "Fx1Q6oDca"
                }
            ]
        }
    ],
    "direction": "row",
    "hAlign": "left",
    "vAlign": "top",
    "size": "shrink",
    "gap": "M"
}
```

We previously weren't finding the state updates within each of the buttons, but now we traverse through these and find them, recursively throughout the screen object.


aptkingston (commented):
This PR prompted me to have a little look at the logic for finding state updates and references, and I think we can substantially simplify it! All we need to do is stringify the value of each setting and check that whole thing (at least when it comes to finding usages). I see today you're recursing through objects and arrays, and even splitting out conditions and styles separately. I don't think we need any of that - pretty sure something like this does the exact same thing!

```js
  const getSettingsWithState = (component: any, stateKey: string): string[] => {
    return Object.entries(component)
      .filter(([_, value]) => hasStateBinding(JSON.stringify(value), stateKey))
      .map(([key]) => key)
  }
```

You're also already filtering out `_children`, `_conditions` and `_styles` before it touches `getSettingsWithState` so you don't need to check the keys again. In my example above, I'm deliberately including conditions and styles since we can check those the same way.

In terms of the text returned that the user can see for the setting names, I can see you're using the component key. That's not ideal because the key is not the label, and isn't necessarily the same as the label that the user will see in the component settings panel. It would be much better to use the component definition to get the label for the associated setting - that ensures we use the correct user readable string.

Another thing which would substantially simplify things and make your life easier when looking for event settings is to use the definition again, and search by the `type` of a setting. I can see you're using a hardcoded list of setting keys, which is flaky and would need maintained over time, making it very easy to miss things. If we instead use the component definition to find settings of the relevant type, we can guarantee we'll not miss anything and it requires no maintenance 👌 

If you need a hand with any of these changes or any more clarification just let me know.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/StatePanel.svelte
```diff
@@ -64,32 +65,71 @@
...
-          }
-        })
-      }
+    let eventHandlers: string[] = []
+    if ($selectedScreen) {
+      let componentSettings = findComponentsBySettingsType(
+        $selectedScreen,
+        "event",
+        $componentStore.components
+      )
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/StatePanel.svelte:74]: Can I just check the logic here - so you're wanting to find components updating state, and this `findComponentsUpdatingState` is only concerned with a single component at a time. The first thing that happens is that you check all components on the entire screen for any `event` settings, then make an array of all unique keys that those event settings have, then check all those keys against the current component.

All you need to worry about is the component being passed in, so you can just check the settings for that one component. A better solution would be using `componentStore.getComponentSettings(component._component)`, which gives you an array of all the settings for this component, then filter those down to `event` types. My suggestion for using this `findComponentsBySettingsType` util was if we were just checking the whole screen at once, rather than one component at a time recursively. It still would probably be more performant again to just do the whole screen at once though.

You could also use the same util to find all `buttonConfiguration` settings and check those, since they can contain state update actions like you mentioned.

Happy to jump on this branch and make those changes if that helps 👍

aptkingston (commented):
Nice improvements @PClmnt! Still a couple more changes that I think we need though before we're good to go.

aptkingston (approved):

PR #15465, opened 2025-01-30
Avoid duplicated names on creating components

adrinr:
## Description
Add a counter to the `New whatever` component name to avoid creating multiple items with the same name. We don't validate that they are not called the same, so any existing duplicates or renaming to have the same duplicates will be allowed

## Screenshots

https://github.com/user-attachments/assets/c3313193-7c1d-45e4-9b11-10e2ad6fad0a

## Launchcontrol
Avoid duplicated names on creating components

### DIFF
--- a/packages/builder/src/helpers/duplicate.ts
+++ b/packages/builder/src/helpers/duplicate.ts
@@ -76,13 +76,15 @@ export const getSequentialName = <T extends any>(
   {
     getName,
     numberFirstItem,
+    separator = "",
   }: {
     getName?: (item: T) => string
     numberFirstItem?: boolean
+    separator?: string
   } = {}
 ) => {
   if (!prefix?.length) {
-    return null
+    return ""
   }
   const trimmedPrefix = prefix.trim()
   const firstName = numberFirstItem ? `${prefix}1` : trimmedPrefix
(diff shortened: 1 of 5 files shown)


packages/builder/src/helpers/duplicate.ts
```diff
@@ -76,13 +76,15 @@ export const getSequentialName = <T extends any>(
   {
     getName,
     numberFirstItem,
+    separator = "",
```
aptkingston, line 79: The way to handle a space is just to include a space in the prefix - so just add a space to the end of `New ${definition.friendlyName || definition.name}` down below and it has the same effect without needing a separator param 👌 This util is written to trim the prefix and handle it properly when looking for existing names.

aptkingston (approved):
LGTM! We can keep things simpler by not needing the separator param 👍

packages/builder/src/helpers/duplicate.ts
adrinr: This was my initial approach, but I felt it was confusing. I set that the prefix (aka the name I actually want) with the whitespace, but the result is without the whitespace if there is a count. As a user, I would expect to respect the whitespace by default

PR #15469, opened 2025-01-31
Validate custom components

adrinr:
## Description
Validate custom components. While doing this I realised we should be using the existing definitions instead us using the manifest.json, as it contains plugins and the snapshot of the definitions based on the app version.
I also renamed the feature flag, to avoid wrong usage of manifest for versions before this one.

## Screenshots

https://github.com/user-attachments/assets/0671b6d8-f1e2-41aa-81c1-4a9cefceaa1b

## Launchcontrol
Validate custom component settings


aptkingston (approved):
LGTM!

PR #15472, opened 2025-01-31
clear results context on every run

shogunpurple:
## Description
Fix for formula columns persisting errors across runs.

## Addresses
- https://linear.app/budibase/issue/BUDI-9023/when-a-formula-column-errors-it-affects-other-formula-columns-within-a


mike12345567 (approved):
LGTM - thanks for adding test case!

PR #15477, opened 2025-02-03
Design errors popover

adrinr:
## Description
Displaying a popover aggregating all the component errors in a single place. This required changing the store to use an array instead of an object, in order to be able to display the errors in order.

### No errors
image

### Errors

https://github.com/user-attachments/assets/ec14573d-ca51-4288-86b4-79873044a755

## Launchcontrol
Displaying a popover aggregating all the component errors in a single place.

## Feature branch env
[Feature Branch Link](http://fb-budi-9016-design-errors-popup.fb.qa.budibase.net)


adrinr:
Dependant on https://github.com/Budibase/budibase/pull/15483

ben-budibase:
Looks good to me @adrinr nice one

mike12345567 (approved):
Everything looks good to me, checked it out and had a play with it, very nice! I would get one final confirmation from @aptkingston as the update the the screen components store is obviously quite large - worth checking over that!

adrinr:
> Everything looks good to me, checked it out and had a play with it, very nice! I would get one final confirmation from @aptkingston as the update the the screen components store is obviously quite large - worth checking over that!

Sure! The screen components was recently created for this epic, so it should be quite safe anyways :)

aptkingston (commented):
Code LGTM!

Had a small play around with the design and wondered if something like this was slightly better? I don't think we need to include the setting name in the title, since it's already mentioned in the message. And I thought using a new line for text and slightly more contrast in the text color looked better. Also shrank the icon and tweaked spacing a little. Maybe a question for Ben?
image

Only other visual thing would be the tooltip - could we change it to be a top tooltip, remove the capital "E" from "Errors" and remove the exclamation mark?

Fantastic feature though - great work!

joebudi:
I'm not a fan of the new line because I like the condensed feel of the popover. Also, the smaller the popover the better. I agree with everything else.

aptkingston:
> I'm not a fan of the new line because I like the condensed feel of the popover. Also, the smaller the popover the better. I agree with everything else.

They still take up 2 lines in the current version - it already wraps onto the second line. So this is the same size, but just more readable. I also removed some text by removing the settings name from the title. But if you prefer the denser text that's totally valid - just wanted to point out it doesn't increase the size.

aptkingston:
> This required changing the store to use an array instead of an object, in order to be able to display the errors in order.

Would this be a breaking change for the previous client library version which expects an object? That's already been deployed hasn't it?

joebudi:
> > I'm not a fan of the new line because I like the condensed feel of the popover. Also, the smaller the popover the better. I agree with everything else.
> 
> They still take up 2 lines in the current version - it already wraps onto the second line. So this is the same size, but just more readable. I also removed some text by removing the settings name from the title. But if you prefer the denser text that's totally valid - just wanted to point out it doesn't increase the size.

If you remove the setting name, that's more space saved.

ben-budibase:
> Code LGTM!
> 
> Had a small play around with the design and wondered if something like this was slightly better? I don't think we need to include the setting name in the title, since it's already mentioned in the message. And I thought using a new line for text and slightly more contrast in the text color looked better. Also shrank the icon and tweaked spacing a little. Maybe a question for Ben? image
> 
> Only other visual thing would be the tooltip - could we change it to be a top tooltip, remove the capital "E" from "Errors" and remove the exclamation mark?
> 
> Fantastic feature though - great work!

I like this Cheeks, this would be my preferred option!

adrinr:
> > This required changing the store to use an array instead of an object, in order to be able to display the errors in order.
> 
> Would this be a breaking change for the previous client library version which expects an object? That's already been deployed hasn't it?

No, as it's passed as an object to the client

PR #15478, opened 2025-02-03
Modernise automation step tests.

samwho:
## Description

This PR is in preparation of adding a suite of automation trigger tests. I've done the following:

1. Moved the step tests into their own subdirectory called `steps`, and added the `trigger` subdirectory. Found there are already tests for webhook triggers, so they're the first resident of the trigger directory.
2. Reduced existing test dependency on `setup.getConfig()`, something we've been doing in lots of other files to reduce the likelihood of inter-test interference.
3. Removed a whole bunch of `any`s I found lying around.
4. Removed a lot of uses of the old `runStep` way of running automation tests, these are now using `createAutomationBuilder` instead.
5. Removed various uses of `config.createTable` and `config.createRow` and replaced them with the corresponding `config.api` calls.
6. Removed mocking in some OpenAI tests, and added AI quota checks to make sure we're counting quota correctly.


packages/server/src/automations/tests/scenarios.spec.ts
```diff
@@ -484,12 +486,18 @@ if (descriptions.length) {
...
+        datasourceId: datasource._id!,
+        parameters: [],
+        fields: {
+          sql: client(tableName).select("*").toSQL().toNative().sql,
+        },
+        transformer: "",
+        schema: {},
+        readable: true,
+        queryVerb: "read",
+      })
```
samwho, line 500: This was the only use of `saveTestQuery`, so I moved it inline and deleted the function.

packages/server/src/automations/tests/steps/bash.spec.ts
```diff
@@ -1,30 +1,30 @@
...
+    automation.shutdown()
+    config.end()
+  })
 
   it("should use trigger data in bash command and pass output to subsequent steps", async () => {
-    const result = await createAutomationBuilder({
-      name: "Bash with Trigger Data",
-      config,
-    })
+    const result = await createAutomationBuilder({ config })
```
samwho, line 27: You'll notice this pattern a lot. It's sufficient in almost all cases to just pass in `config` and nothing else. The name does nothing and is just visual noise.

packages/server/src/automations/tests/steps/delay.spec.ts
samwho, line 1: This shows up as a new file because enough of it was rewritten in the move, it's actually moved from the previous directory.

packages/server/src/automations/tests/steps/discord.spec.ts
samwho, line 1: Similarly here, this is moved but has changed enough that it's tracked as a new file.

packages/server/src/automations/tests/steps/filter.spec.ts
samwho, line 1: And here, moved but largely rewritten in the process.

packages/server/src/automations/tests/steps/loop.spec.ts
samwho, line 1: This file is now a combination of the `loop.spec.ts` and `looping.spec.ts` file from the `scenarios/` directory that no longer exists. The two files were both testing looping, so I decided to merge them to keep all that testing in one place.

packages/server/src/automations/tests/steps/make.spec.ts
samwho, line 1: Again, moved and largely rewritten in the process.

packages/server/src/automations/tests/steps/openai.spec.ts
samwho, line 1: This has had its jest mocks removed and replaced with `nock`.

packages/server/src/automations/tests/steps/outgoingWebhook.spec.ts
samwho, line 1: Again, moved but largely rewritten.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -294,9 +387,8 @@ class AutomationBuilder extends BaseStepBuilder {
         stepNames: {},
       },
       type: "automation",
-      appId: options.appId ?? setup.getConfig().getAppId(),
+      appId: options.appId ?? this.config.getAppId(),
```
samwho, line 390: This was incorrect, if you passed a `config` but not an `appId` it would try to pull an `appId` from the uninitialised `setup.getConfig()`, resulting in an error.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -356,6 +448,16 @@ class AutomationBuilder extends BaseStepBuilder {
     )
   }
 
+  cron(inputs: CronTriggerInputs, outputs?: CronTriggerOutputs) {
```
samwho, line 451: I'm iffy about the ordering here, this is only used in one place at the moment but it feels more natural to have the arguments this way around (the rest of them are reversed).

adrinr (approved):
Nice to see cleanups :)

PR #15482, opened 2025-02-04
Handle outside clicks when clicking the iframe

adrinr:
## Description
The builder uses a `click_outside` hander in order to perform some operations such as closing modals and side panels. If we are in the design section, if we click in the iframe, this absorbs the click and the builder is not aware of it. This causes some inconsistencies in the UX.

https://github.com/user-attachments/assets/96bdec76-e0a0-4364-bb04-f8f3a9d668ca

This is even more visible in the new error popover:

https://github.com/user-attachments/assets/ff0477a0-5679-4802-8a1a-d4fddfbbc7d0

Happy to discuss the taken approach and other possible solutions.

### DIFF
--- a/packages/bbui/src/Actions/click_outside.ts
+++ b/packages/bbui/src/Actions/click_outside.ts
@@ -34,14 +34,14 @@ let candidateTarget: HTMLElement | undefined
 // Processes a "click outside" event and invokes callbacks if our source element
 // is valid
 const handleClick = (e: MouseEvent) => {
-  const target = e.target as HTMLElement
+  const target = e.target as HTMLElement | null
 
   // Ignore click if this is an ignored class
-  if (target.closest('[data-ignore-click-outside="true"]')) {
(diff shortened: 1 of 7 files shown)


aptkingston:
Had a look at this myself and I think there's a much simpler way to achieve this - https://github.com/Budibase/budibase/compare/detect-iframe-clicks. What do you think?

adrinr:
> Had a look at this myself and I think there's a much simpler way to achieve this - #15483. What do you think?

That's a much nicer approach, as we remove any logic from the client! Good thinking 🚀

aptkingston:
> > Had a look at this myself and I think there's a much simpler way to achieve this - #15483. What do you think?
> 
> That's a much nicer approach, as we remove any logic from the client! Good thinking 🚀

Nice. If you approve that one then we can get it merged in 👍

adrinr:
Closed in favour of https://github.com/Budibase/budibase/pull/15483

PR #15483, opened 2025-02-04
Detect iframe clicks via loss of focus

aptkingston:
## Description
This PR updates the `click_outside` utility to handle clicks of the app preview iframe, via detecting loss of focus on the main window.

https://github.com/user-attachments/assets/d73d569a-702b-497b-a08d-0412b9a64c2c


adrinr (approved):
Nice solution! ❤️

packages/bbui/src/Actions/click_outside.ts
```diff
@@ -91,9 +91,19 @@ const handleMouseDown = (e: MouseEvent) => {
...
+    handleClick(
+      new MouseEvent("click", { relatedTarget: document.activeElement })
+    )
+  }
+}
+
 // Global singleton listeners for our events
 document.addEventListener("mousedown", handleMouseDown)
 document.addEventListener("contextmenu", handleClick)
+window.addEventListener("blur", handleBlur)
```
adrinr, line 106: Should this be in bbui, or should it be part of the builder only?

aptkingston: I don't think it matters if it's in BBUI - all it'll do is trigger an additional click (with an iframe being the target) if focus shifts to an iframe element. I think it's valid to include this everywhere, as otherwise iframe clicks are always lost (e.g. if inside a client app and clicking on an iframe - today we would not catch this, but we would now).

PR #15487, opened 2025-02-05
Show screens that are linked to the selected query

PClmnt:
## Description
Adds a button in both the datasource / rest query screens that shows the user how many screens are linked to this particular query. 

Uses the same component as the grid button that so look / usage is the same. 

## Screenshots

image

image

## Launchcontrol
- Added a button/ popover to both the Datasource and REST Query screens displaying the number of screens linked to the respective query


packages/types/src/ui/components/popover.ts
```diff
@@ -0,0 +1,6 @@
+export enum PopoverAlign {
+  Left = "left",
+  Right = "right",
+  LeftOutside = "left-outside",
+  RightOutside = "right-outside",
+}
```
aptkingston, line 6: `PopoverAlignment` already is an enum exported from BBUI - so you can resue that instead of adding this! 👌

aptkingston (approved):
LGTM! Just one small thing we can remove.

PR #15488, opened 2025-02-05
Create an automation cron trigger test.

samwho:
## Description

Creates a test to assert cron triggers are set up correctly.

We rely on the bull queue to do the right thing with regards to how frequently the cron trigger fires so there's no test for that here because it would require using bull properly rather than our in-memory queue.


mike12345567 (approved):
LGTM - nice to be able to test this now, some good improvements to the in memory queue mock as well!

I think realistically we need to be able to mock this simply due to the nature of what automations do - triggers like this simply are easily testable without some form of mock.

PR #15489, opened 2025-02-05
Fix searching for a date on a string field.

samwho:
## Description

We were erroneously converting date formatted strings into date objects for searches against fields that were not `FieldType.DATETIME`, which caused these searches to return an error.

## Addresses
- https://linear.app/budibase/issue/BUDI-9018/searching-budidb-column-headings-in-the-data-tab-breaks-grid

## Launchcontrol

- Fix searching for date-formatted strings on non-date fields.


mike12345567 (approved):
Very nice! LGTM

PR #15494, opened 2025-02-05
Refactor automation builder.

samwho:
## Description

Most calls to `createAutomationBuilder` followed the same pattern: they passed in `config`, set a name, maybe set an `appId` that always matched `config.getAppId()`. I've simplified this pattern so that you only have to do `createAutomationBuilder(config)`. The name was never used for anything anyway, but I've made it so that you can set it if you want with a new `name(n: string)` method on the builder.

I've also set `appAction({ fields: {} })` to be the default trigger, as most uses of `createAutomationBuilder` were doing this manually.

Lastly, I removed a lot of duplication in the `AutomationBuilder.ts` file by making the creation of the step and filter functions the result of another function call, shortening them to one line each and making sure they're consistent. This will also make it a bit easier to add new ones in future.


PClmnt (approved):
Really nice refactor! LGTM

PR #15497, opened 2025-02-06
New component drag-and-drop improvements

aptkingston:
## Description
This PR updates and refactors a lot of the drag and drop code, particularly when adding components. The primary new feature is the ability to drag components directly on to grid screens, in the position of your choice.

Here's an example of adding a few new components to a grid screen:

https://github.com/user-attachments/assets/d4de2efc-4a75-4929-a1ee-24e9b7d67a2d

Dragging components on to flex screens is unaffected.

## Other stuff
- Updates all client import aliases to use new `@/xxx` approach (same as builder)
- Converts all DND related components and stores to typescript
- Converts all code relating to "indicators" (selected component, hovered component, new dragged component) to typescript
- Improves performance of DND


deanhannigan (approved):
LGTM! Love it.  A really intuitive UX enhancement and some really nice updates to the core DnD behaviour! 🎉

Gave this a rinse locally and didn't run into any issues!

`NOTE` - Small default size for the following. Have to resize to see the help UX. Not a blocker at all.
  - Dynamic filter
  - Date range
  - Icon

aptkingston:
> LGTM! Love it. A really intuitive UX enhancement and some really nice updates to the core DnD behaviour! 🎉
> 
> Gave this a rinse locally and didn't run into any issues!
> 
> `NOTE` - Small default size for the following. Have to resize to see the help UX. Not a blocker at all.
> 
> * Dynamic filter
> * Date range
> * Icon

Cheers Dean. Nice one finding those components that look weird by default too, due to error states. I think I'll update the icon to just pick an icon by default to solve that one.

For the others it's not so easy, since they need a data provider. And when there is a data provider, they are much smaller than the error state. I'll add some logic to attempt to auto fill in data provider type settings, which will help reduce the frequency of initial error states.

PR #15499, opened 2025-02-06
Data section resource deletion warning

mike12345567:
## Description
Previously we had some warnings for deletion of data resources - but they were not always accurate and did not cover all of the data resources that are available within the system.

We have implemented a new endpoint `POST /api/screens/usage/:sourceId` that will determine what type of data resource the source is and then attempt to find its usages throughout the screens within the app - this can then be used to warn users before deleting a data resource that it may break the screen to remove it.

## Screenshots
A query, not in use:
image

A query that is in use:
image

A datasource which is in use (through queries) and has two queries attached to it:
image

An internal table, with views, that is in use:
image

An external table, with views, but no screens:
image

An external table, with no views and no screens:
image

A view used in screens:
image


packages/server/src/sdk/app/common/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (docIds.isTableId(sourceId)) {
+    return SourceType.TABLE
+  } else if (docIds.isViewId(sourceId)) {
+    return SourceType.VIEW
+  } else if (docIds.isDatasourceId(sourceId)) {
+    return SourceType.DATASOURCE
+  } else if (docIds.isQueryId(sourceId)) {
+    return SourceType.QUERY
+  }
+  throw new Error("Unknown source type - cannot find document type")
```
samwho, line 14: _Probably_ worth including the `sourceId` in the output here, make it easier to debug when it goes wrong.

packages/server/src/tests/utilities/structures.ts
```diff
@@ -517,6 +523,18 @@ export function basicScreen(route = "/") {
...
+  return createTableScreen(datasourceName, table)
+}
+
+export function viewScreen(datasourceName: string, view: ViewV2) {
+  return createViewScreen(datasourceName, view)
+}
+
+export function queryScreen(datasourceId: string, query: Query) {
+  return createQueryScreen(datasourceId, query)
+}
```
samwho, line 536: These feel redundant? Why not call the inner methods directly?

packages/shared-core/src/sdk/documents/screens.ts
```diff
@@ -0,0 +1,23 @@
+import { Screen } from "@budibase/types"
+
+export function findInSettings(screen: Screen, toFind: string) {
+  const foundIn: { setting: string; value: string }[] = []
+  function recurse(props: Record<string, any>, parentKey = "") {
+    for (let key of Object.keys(props)) {
```
samwho, line 6:
```suggestion
    for (const key of Object.keys(props)) {
```

Also, given you're not reassigning `props[key]`, it might be a bit nicer to use `for (const [key, value] of Object.entries(props))`.

packages/types/src/ui/common.ts
```diff
@@ -0,0 +1,3 @@
+export type UIEvent = Event & {
+  currentTarget: EventTarget & HTMLInputElement
+} & { key?: string } & { target?: any }
```
samwho, line 3: Haha, finally snapped huh?

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,224 @@
...
+    const screenCount = affectedScreens.length
+    let message = `Removing ${source?.name} `
+    let initialLength = message.length
+    if (sourceType === SourceType.TABLE) {
+      const views = "views" in source ? Object.values(source?.views ?? []) : []
+      message += `will delete its data${
+        views.length
+          ? `${screenCount ? "," : " and"} views (${views.length})`
+          : ""
+      }`
```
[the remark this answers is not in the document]
mike12345567: Discussed on Discord

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,224 @@
...
+      <p class="warning">
+        {buildMessage(sourceType)}
+        {#if affectedScreens.length > 0}
+          <span class="screens">
+            {#each affectedScreens as item, idx}
+              <Link overBackground target="_blank" href={item.url}
+                >{item.text}{idx !== affectedScreens.length - 1
+                  ? ","
+                  : ""}</Link
+              >
```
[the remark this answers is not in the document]
mike12345567: Yeah, just seems to be the way the linter wants to do it!

packages/types/src/ui/common.ts
samwho: I think you might be better served here by something like:

```typescript
type BudibaseEventHandler<E extends Event = Event, T extends EventTarget = Element, V extends EventTarget = Element> = (
  event: E & { currentTarget: EventTarget & T, target: EventTarget & V }
)
```

To follow Svelte's own types:

CleanShot 2025-02-06 at 17 10 01@2x

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,224 @@
...
+
+  async function deleteDatasource(datasource: Datasource) {
+    try {
+      const isSelected =
+        get(datasources).selectedDatasourceId === datasource._id
+      await datasources.delete(datasource)
+      notifications.success("Datasource deleted")
+      if (isSelected) {
+        $goto("./datasource")
+      }
```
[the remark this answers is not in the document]
mike12345567: As far as I can tell this function is synchronous, it cannot fail here, as this essentially loads the router with where you'd like to go, so that the router can move the app to that page when it sees fit - if there is an error in the routing it does not appear to be possible for it to be thrown from the goto, it'll instead render an error state for the whole app (as routing has failed, theres no where to go).

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -85,7 +86,7 @@
 
   async function deleteView(view: ViewV2 | View) {
     try {
-      if ("version" in view && view.version === 2) {
+      if ("version" in view && helpers.views.isCalculationView(view)) {
```
samwho, line 89: Is this a mistake? I'm not sure calculation view is the right check here.

packages/server/src/api/routes/tests/screen.spec.ts
```diff
@@ -67,39 +67,30 @@ describe("/screens", () => {
...
     })
 
     async function checkScreens(roleId: string, screenIds: string[]) {
       await config.loginAsRole(roleId, async () => {
         const res = await config.api.application.getDefinition(
-          config.prodAppId!,
-          {
-            status: 200,
-          }
+          config.prodAppId!
```
samwho, line 93: nit: `config.getProdAppId()`, will properly check if it's set and fail with a nice error message if it's not.

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
mike12345567: Ah woops - auto-complete has led me astray!

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,225 @@
...
+    switch (sourceType) {
+      case SourceType.TABLE:
+        return await deleteTable(source as Table)
+      case SourceType.VIEW:
+        return await deleteView(source as ViewV2)
+      case SourceType.QUERY:
+        return await deleteQuery(source as Query)
+      case SourceType.DATASOURCE:
+        return await deleteDatasource(source as Datasource)
+    }
```
adrinr, line 143: We could use `utils.unreachable` now that we have ts

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,225 @@
...
+  async function deleteView(view: ViewV2 | View) {
+    try {
+      if (helpers.views.isV2(view)) {
+        await viewsV2.delete(view as ViewV2)
+      } else {
+        await views.delete(view as View)
+      }
+      notifications.success("View deleted")
+    } catch (error) {
+      console.error(error)
```
adrinr, line 96: To remove

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
```diff
@@ -0,0 +1,225 @@
...
+    screens: { url: string; _id: string }[]
+  ): { text: string; url: string }[] {
+    return screens.map(({ url, _id }) => ({
+      text: url,
+      url: `/builder/app/${$appStore.appId}/design/${_id}`,
+    }))
+  }
+
+  function hideDeleteDialog() {
+    sourceType = undefined
```
adrinr, line 64: Should this not close the dialog, instead of removing it's content?

packages/frontend-core/src/api/screens.ts
```diff
@@ -2,12 +2,14 @@ import {
...
   SaveScreenRequest,
   SaveScreenResponse,
+  UsageOfScreensResponse,
 } from "@budibase/types"
 import { BaseAPIClient } from "./types"
 
 export interface ScreenEndpoints {
   saveScreen: (screen: SaveScreenRequest) => Promise<SaveScreenResponse>
   deleteScreen: (id: string, rev: string) => Promise<DeleteScreenResponse>
+  usageOfScreens: (sourceId: string) => Promise<UsageOfScreensResponse>
```
adrinr, line 12: Should this not be `usageInScreens`?

packages/server/src/sdk/app/views/index.ts
```diff
@@ -81,6 +81,14 @@ export function isView(view: any): view is ViewV2 {
   return view.id && docIds.isViewId(view.id) && view.version === 2
 }
 
+export function isInternal(viewId: string) {
+  if (!docIds.isViewId(viewId)) {
+    return false
```
adrinr, line 86: Should this not be an exception?

packages/shared-core/src/sdk/documents/screens.ts
```diff
@@ -0,0 +1,23 @@
+import { Screen } from "@budibase/types"
+
+export function findInSettings(screen: Screen, toFind: string) {
+  const foundIn: { setting: string; value: string }[] = []
+  function recurse(props: Record<string, any>, parentKey = "") {
```
adrinr, line 5: Should `props` not be of type Component?

packages/types/src/ui/common.ts
```diff
@@ -0,0 +1,3 @@
+export type UIEvent = Event & {
```
adrinr, line 1: Why all these unions instead of having a single object?

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
mike12345567: It does close the dialog, this is just a function that runs before it closes (the dialog component will always close itself, but can notify its about to close).

packages/server/src/sdk/app/views/index.ts
mike12345567: Don't actually need this anymore - so just removed it!

packages/builder/src/components/backend/modals/DeleteDataConfirmationModal.svelte
adrinr: Oh, I see. I think the standard should be calling it `onHideDeleteDialog`

adrinr (approved):
LGTM! Great UX improvement :)

PR #15505, opened 2025-02-07
Disabling grid cell formatting

mike12345567:
## Description
Disabling cell formatting, it is currently incompatible with cell/row conditions in grids and to avoid breaking those for now we are removing the capability until both can be a aligned to use the same mechanism of binding evaluation.

## Addresses
- discussion: https://github.com/Budibase/budibase/discussions/15502


aptkingston (approved):
Cheeky log left in there but otherwise good to go.

PR #15512, opened 2025-02-10
Create tests for the row saved trigger.

samwho:
## Description

This PR creates tests for the row saved automation trigger. One of the key things that changes in this PR is that I have introduced a new helper method, `captureAutomationRuns`, that will capture all automation runs that happen while a given function is executing. I expect to use this in quite a few of the trigger tests going forward.


mike12345567 (approved):
LGTM!

PR #15514, opened 2025-02-10
Row deleted trigger tests.

samwho:
## Description

Adds tests for the row deleted trigger. Also introduces a more specific variant of the helper that captures automation results, where you can pass in an automation to only get results that relate to that automation.


mike12345567 (approved):
LGTM!

PR #15515, opened 2025-02-10
Add tests for the row updated trigger.

samwho:
## Description

Almost identical to https://github.com/Budibase/budibase/pull/15514 but for the row updated trigger.


mike12345567 (approved):
LGTM - nice to see how easy these tests are to create now!

PR #15521, opened 2025-02-11
Fix js completions

adrinr:
## Description
Fixing the way we handle autocomplete for javascript. The current behaviour works fine for `fields`, but it does not for `helpers` and `snippers`. Also, updating the autocomplete package to make use of `displayLabel`, that will simplify the filter selection.

### Helpers
Before, they were displaying as autocomplete options with `$("")`. This generates outputs such as `$("random")`, that are not valid. The result should be `helpers.random()`.
image

https://github.com/user-attachments/assets/41647fd2-7c59-4d2b-a0f6-fd126d6f90ed

There is no autocomplete for the right `helpers.whatever()`

After:
1. there is an autocomplete for `helpers.whatever()`
2. helpers will not appear in `$("")`

### Snippets
The same issues we got for autocomplete will occur here. After the changes, we the snippets will be formatted and it will behave as the helpers

https://github.com/user-attachments/assets/329ca553-0671-4cc3-a2fd-69111597655a

## Final result

https://github.com/user-attachments/assets/8af50242-6396-4396-b495-1e0651b791c7

### Helpers
image

It will filter any part of the helper name
image

### Snippets
They will be formatted to be aligned with the helpers formats
image

## Launchcontrol
Improve javascript bindings autocomplete


packages/builder/package.json
```diff
@@ -53,7 +53,7 @@
     "@budibase/shared-core": "*",
     "@budibase/string-templates": "*",
     "@budibase/types": "*",
-    "@codemirror/autocomplete": "^6.7.1",
+    "@codemirror/autocomplete": "6.9.0",
```
adrinr, line 56: Fixing this, as while trying to get newer ones I got conflicts with the other packages

packages/builder/src/types/bindings.ts
```diff
@@ -0,0 +1,8 @@
+import { CompletionContext, Completion } from "@codemirror/autocomplete"
+
+export type BindingCompletion = (context: CompletionContext) => {
```
adrinr, line 3: These types are only used in the builder, and they have dependencies that are not used anywhere else. It makes more sense to live in here that in the `types` package

aptkingston (commented):
Huge value for a reasonably small change! Feels much better to use now. 🚀 

I couldn't find any issues other than 1 small one - when you're writing snippets you now get suggestions for helpers, but you can't actually use helpers inside snippets. So we just need to remove those suggestions.

aptkingston (approved):
LGTM

PR #15527, opened 2025-02-11
Table settings updates

aptkingston:
## Description
This PR restores the table "Format" setting, which had been reverted due to it breaking the existing table "Conditions" setting. This PR allows both conditions and formatting to work as intended.

There is an intended difference to the context (bindings) available to each setting.
- Conditions can access all normal bindings, but cannot reference values in the row itself
- Formatting cannot access normal bindings, and can only reference values in the row itself

This difference is intended and is mainly due to performance constraints.

This fix was not possible without adding (another) new concept to the manifest - `contextAccess`, which acts as an escape hatch for this and any future binding problems caused by how we render certain settings using a pseudo component definition. This has been added to the appropriate type, and should not normally ever need used.

## Other stuff
Fixed some inconsistencies between which bindings are available in real settings versus the same setting when configuring conditional UI.


deanhannigan (approved):
LGTM! Ran into one small issue when running through the condition behaviour. Other than that, the binding restrictions are in place and the features are working great 🚀 

`BUG` - Preview will crash if a newly added condition is saved on a GridBlock without setting an initial value.

Issue #15530, opened 2025-02-11
Plugins no longer showing in preview / published app


melohagan:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 3.4.6
  - App Version: 3.4.6

**Describe the bug**
Plugins are no longer showing in my previewed and published apps. This had been working fine until recently. Last tested to be working on **3.4.3**

**To Reproduce**
1. Add a plugin to a screen - shows in Design
2. Hit preview - no plugin
3. Publish - no plugin

**Screenshots**
<img width="752" alt="Image" src="https://github.com/user-attachments/assets/ddc2c3a5-1f9f-40ab-8c6e-047d53863cd1" />

<img width="752" alt="Image" src="https://github.com/user-attachments/assets/81c6d323-6765-4b59-bdcf-620bf9e77f0d" />

*Screen on preview and published is empty*

**Additional context**
Last working recently on version **3.4.3**

mike12345567:
Should be fixed in master/next release.

PR #15533, opened 2025-02-12
Fix plugin loading in production apps

mike12345567:
## Description
This fixes an issue found last night with plugins loading in production apps - the issue was that generation of pre-signed URLs now returns a promise - something that Typescript handled everywhere else in the code, but the rendering of the SSR Svelte component (`BudibaseApp.svelte`) was not type checked.

Sadly Svelte rendering is not type aware, even when the component you're trying to render has typing on the exports. I've gotten around this by passing on a single property `props` which is typed to the component, therefore both the rendering and the component can respect the same type contract.

This isn't 100% full proof, because you can still pass whatever you want to the untyped `Component.render` function of svelte, but its a lot better than the previous completely untyped nature.


samwho (approved):

mike12345567:
Thanks @adrinr for your help - good to find a way to run SSR on TS svelte components 🥇

mike12345567:
Fixes: https://github.com/Budibase/budibase/issues/15530

PR #15535, opened 2025-02-12
App action tests

samwho:
Create tests for the app action trigger type for automations.


packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -207,6 +209,15 @@ class AutomationRunner<TStep extends AutomationTriggerStepId> {
 
     return response
   }
+
+  async trigger(
```
mike12345567, line 213: should it be made more obvious this triggers the automation rather than adding a trigger to the automation you're building?

mike12345567 (approved):
Pre-emptive LGTM - but think it would be nice to differentiate between the addition of a trigger and triggering the automation.

PR #15536, opened 2025-02-12
Cleaning up loop.spec.ts, moved tests from automation.spec.ts to more appropriate places.

samwho:
## Description

- `loop.spec.ts` had some tests that weren't using `createAutomationBuilder` so I converted them.
- `automation.spec.ts` was doing some mocking that we don't need anymore now that we have `createAutomationBuilder`, so I wrote a new test in `appAction.spec.ts` that covered what it was doing, moved another test from it into `automationUtils.spec.ts`, and deleted the file.


packages/server/src/automations/tests/steps/loop.spec.ts
```diff
@@ -34,41 +30,46 @@ describe("Attempt to run a basic loop automation", () => {
...
+    const result = await createAutomationBuilder(config)
+      .onAppAction()
+      .loop({
+        option: LoopStepType.ARRAY,
+        binding: "{{ 1 }}",
+      })
+      .serverLog({ text: "log statement" })
+      .test({ fields: {} })
+
+    expect(result.steps[0].outputs.iterations).toBe(1)
```
samwho, line 72: These should be identical to their `runLoop` counterparts, but using the new builder instead of `runLoop`.

packages/server/src/automations/tests/triggers/appAction.spec.ts
```diff
@@ -42,4 +42,33 @@ describe("app action trigger", () => {
...
+      })
+    })
+
+    expect(jobs).toHaveLength(1)
+    expect(jobs[0].data.event.fields).toEqual({
+      text: "1",
+      number: 2,
+      boolean: true,
+    })
+  })
```
samwho, line 73: This is new to cover a test that was happening in `automation.spec.ts`

mike12345567 (approved):

PR #15540, opened 2025-02-12
Remove typescript from SSR app - fix app preview

mike12345567:
## Description
Reverting svelte config update for SSR typescript - this doesn't seem to work well at all, rather than updating this to Typescript for one type definition I have added it via a type comment instead.


deanhannigan (approved):
LGTM! Ran through it locally and all app previews are loading as expected 👍

PR #15543, opened 2025-02-13
Convert client builder store

adrinr:
## Description
Convert client builder store


packages/client/src/stores/builder.ts
```diff
@@ -2,9 +2,35 @@ import { writable, get } from "svelte/store"
...
+  selectedComponentId: string | null
+  editMode: boolean
+  previewId: string | null
+  theme: string | null
+  customTheme: string | null
+  previewDevice: PreviewDevice
+  navigation: string | null
+  hiddenComponentIds: []
+  usedPlugins: { name: string; hash: string }[] | null
+  eventResolvers: {}
```
aptkingston, line 25: I don't think this is actually ever used - I think it belongs in the event store instead. So we can just remove this property as part of this conversion.

packages/client/src/stores/builder.ts
```diff
@@ -2,9 +2,35 @@ import { writable, get } from "svelte/store"
...
+  metadata: { componentId: string; step: number } | null
```
aptkingston, line 26: I can't find any usages of this metadata... maybe we can remove it too?

aptkingston (commented):
A few type suggestions to make this stricter 👌

deanhannigan (approved):
LGTM! Ran through it locally and didn't run into any issues

packages/client/src/stores/builder.ts
adrinr: This is used in `MultiStepFormblock.svelte`

PR #15545, opened 2025-02-13
Pull out new code editor field and drawer bindable slot into single component

aptkingston:
## Description
Ensures we keep the drawer and new code editor field in sync.


deanhannigan (approved):
LGTM! Besides the lint issues, this looks great. I ran through it locally here and it successfully alleviates the timing issue on blur. I'll be sure to mirror this approach in the automation sidebar updates too 🚀 Cheers!

PR #15546, opened 2025-02-13
Automation Orchestrator refactor.

samwho:
## Description

The motivation for this is a bug in the automation test output when combining a filter step with loop steps: https://linear.app/budibase/issue/BUDI-9015/looped-automation-steps-showing-up-in-test-results-after-a-failed.

In reading the code to try and fix this, I decided to take the plunge and change the way the Orchestrator works in order to make this bug fix easier.

My recommendation for reviewing this is to pull the branch into your local environment and start reading from `packages/server/src/threads/automation.ts`. That's where the `Orchestrator` class, the thing that runs automations, lives and where most of the changes are concentrated.

## Addresses
- https://linear.app/budibase/issue/BUDI-9015/looped-automation-steps-showing-up-in-test-results-after-a-failed

## Feature branch env
[Feature Branch Link](http://fb-fix-automation-loop-test-output-2.fb.qa.budibase.net)


packages/server/src/threads/automation.ts
```diff
@@ -330,286 +310,224 @@ class Orchestrator {
...
+      env.AUTOMATION_MAX_ITERATIONS
+    )
 
-      this.context.steps[pathStepIdx] = {
-        currentItem: this.getCurrentLoopItem(loopStep, loopStepIndex),
+    const items: Record<string, any>[] = []
+    let iterations = 0
+    let iterable: any[] = []
+    try {
+      iterable = getLoopIterable(loopStep)
```
mike12345567, line 432: What is this? The response type appears to be `any` and the loop below makes it look like a number, but the name `iterable` suggests its some sort of array/iterable object

packages/server/src/threads/automation.ts
```diff
@@ -330,286 +310,224 @@ class Orchestrator {
...
+                const stepToLoop = steps[stepIndex + 1]
+                const result = await this.executeLoopStep(ctx, step, stepToLoop)
+
+                ctx.steps.push(result.outputs)
+                ctx.stepsById[stepToLoop.id] = result.outputs
+                ctx.stepsByName[stepToLoop.name || stepToLoop.id] =
+                  result.outputs
+
+                stepOutputs.push(result)
+                stepIndex += 2
```
mike12345567, line 381: this is mildly scary looking - a comment/explanation for the odd increment here would be helpful!

packages/server/src/threads/automation.ts
```diff
@@ -100,53 +156,47 @@ class Orchestrator {
...
 
   constructor(job: AutomationJob) {
-    let automation = job.data.automation
-    let triggerOutput = job.data.event
-    const metadata = triggerOutput.metadata
-    this.chainCount = metadata ? metadata.automationChainCount! : 0
+    this.automation = job.data.automation
+
+    const triggerOutput = job.data.event
+    if (
```
mike12345567, line 168: It would be good to explain this, looks a bit random just appearing in the constructor, a comment or a utility function to perform the function of this (I'm guessing this is filling in missing information in a cron trigger if its not supplied)

packages/server/src/threads/automation.ts
```diff
@@ -330,286 +310,224 @@ class Orchestrator {
...
-      const stepName = steps[stepToLoopIndex].name || steps[stepToLoopIndex].id
-      this.context.stepsByName[stepName] = tempOutput
-      this.context.steps[this.context.steps.length] = tempOutput
-      this.context.steps = this.context.steps.filter(
-        item => !item.hasOwnProperty.call(item, "currentItem")
-      )
+      if (matchesLoopFailureCondition(loopStep, currentItem)) {
+        return {
+          ...result,
+          outputs: {
```
mike12345567, line 461: it would be nice to have a utility for setting a loop failing, just to reduce the visual load of this repeating, all that really matters here is the change in status, something like `return loopFailure(result, AutomationStepStatus.FAILURE_CONDITION)` would make this all look a lot clearer (to me at least)

packages/server/src/threads/automation.ts
samwho: The response type is `any[]`.

packages/builder/src/stores/builder/automations.ts
```diff
@@ -1134,7 +1138,7 @@ const automationActions = (store: AutomationStore) => ({
    * @returns
    */
   shiftBranch: (pathTo: Array<any>, block: AutomationStep, direction = -1) => {
```
adrinr, line 1140: Should the block input not be `BranchStep` instead?

packages/server/src/definitions/automations.ts
```diff
@@ -13,19 +13,17 @@ export interface TriggerOutput {
   timestamp?: number
 }
 
-export interface AutomationContext extends AutomationResults {
+export interface AutomationContext {
+  trigger: any
```
adrinr, line 17: Should this not be an AutomationTrigger?

samwho: It's the output of an `AutomationTriggerResult`. I've created a new type for this that aliases `Record<string, any>`, because I don't think we have better typing that than yet.

packages/builder/src/stores/builder/automations.ts
samwho: Good call! Much nicer.

packages/server/src/threads/automation.ts
```diff
@@ -83,650 +196,408 @@ export async function enrichBaseContext(context: Record<string, any>) {
...
+    await processObject(step.inputs, prepareContext(ctx))
+
+    const maxIterations = getLoopMaxIterations(step)
+    const items: Record<string, any>[] = []
+    let iterations = 0
+    let iterable: any[] = []
+    try {
+      iterable = getLoopIterable(step)
+    } catch (err) {
+      return stepFailure(stepToLoop, {
```
mike12345567, line 474: This is nice, much more explicit!

packages/server/src/threads/automation.ts
```diff
@@ -83,650 +196,408 @@ export async function enrichBaseContext(context: Record<string, any>) {
...
-/**
- * The automation orchestrator is a class responsible for executing automations.
- * It handles the context of the automation and makes sure each step gets the correct
- * inputs and handles any outputs.
- */
 class Orchestrator {
-  private chainCount: number
-  private appId: string
-  private automation: Automation
+  private readonly job: AutomationJob
```
mike12345567, line 215: This is amazing, gotten rid of so much state!

mike12345567 (approved):
This is fantastic, the main automation thread is a lot easier to follow now! Excellent to see the majority of state has been removed, it was definitely disrupting the flow of code, much easier to follow how the context gets updated/shifted around now.

packages/server/src/threads/automation.ts
```diff
@@ -83,650 +196,408 @@ export async function enrichBaseContext(context: Record<string, any>) {
...
+    ctx: AutomationContext,
+    step: LoopStep,
+    stepToLoop: AutomationStep
+  ): Promise<AutomationStepResult> {
+    await processObject(step.inputs, prepareContext(ctx))
+
+    const maxIterations = getLoopMaxIterations(step)
+    const items: Record<string, any>[] = []
+    let iterations = 0
+    let iterable: any[] = []
```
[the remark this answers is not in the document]
samwho: `iterable` is the thing we're looping over, which I believe is a legitimate use of `any` seeing as it could really be anything if the user decides to give us some random JavaScript. Correct me if I'm wrong.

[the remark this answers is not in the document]
mike12345567: I believe `any` is correct yes, I was just thinking an alias for `any` - so that anywhere we're using step responses we can can tell that it is a step response, not just an unknown `any` - more for knowledge in the code rather than any type checking!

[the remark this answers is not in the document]
samwho: I'm going to leave this for the time being, because it feels fairly minor and we can follow-up if we really want to.

Generally I've been a bit annoyed at our use of this pattern, but it comes down entirely to tooling. I _wish_ VSCode would tell me the shape of the underlying type when I hover of it. Currently I see this:

CleanShot 2025-02-19 at 09 32 55@2x

I would love love love it to show me a recursive resolution of the type, and not make me cmd+click my way down to see what's actually going on.

adrinr (approved):

PR #15548, opened 2025-02-13
Revert setting a default icon

aptkingston:
## Description
This can cause undesired effects for plugin authors, so I'm reverting this. We'll continue to show an error when adding icon components.

### DIFF
--- a/packages/builder/src/stores/builder/components.ts
+++ b/packages/builder/src/stores/builder/components.ts
@@ -311,8 +311,6 @@ export class ComponentStore extends BudiStore<ComponentState> {
             component[setting.key] = fieldOptions[0]
             component.label = fieldOptions[0]
           }
-        } else if (setting.type === "icon") {
-          component[setting.key] = "ri-star-fill"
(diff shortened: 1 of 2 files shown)


poirazis:
@aptkingston maybe set the default value of the property via the manifest schema of the icon component instead of the icon property so the error state can be avoided too

aptkingston:
> @aptkingston maybe set the default value of the property via the manifest schema of the icon component instead of the icon property so the error state can be avoided too

Do you do that yourself with your icon settings - setting a default value via the manifest? Just curious if that works for plugins as well!

poirazis:
> > @aptkingston maybe set the default value of the property via the manifest schema of the icon component instead of the icon property so the error state can be avoided too
> 
> Do you do that yourself with your icon settings - setting a default value via the manifest? Just curious if that works for plugins as well!

I use the schema to set the default value and also set a default value on the component property , and yeah it was working !

deanhannigan (approved):
LGTM!

Issue #15428, opened 2025-01-23
a Helper function to decode _id


poirazis:
When working with forms / relationship pickers with SQL db , the numeric ids get encoded.
Often we need to grab an id from a form or a picker to pass to a custom query or filter, and we need the unencoded id.
this is not hard to do, with JS, but it so redundant that would be easier to have a helper funtion.

it would look like `{{ decodeId Form.Relationship.0_id }} `

you can curretnly decode it using 

```
if ( val && val.length )
  return val[0]._id.substring(3, val[0]._id.length - 3)
```

mikesealey:
https://github.com/Budibase/budibase/pull/15549

PR #15549, opened 2025-02-13
Budi 9011 - DecodeId helper

mikesealey:
## Description
_Carbon copy of_ `packages/backend/backend-core/src/sql.utils.ts line 125` - this PR surfaces a Handlebars Helper to allow the user to un-encode a numeric ID from an SQL table. This is useful when fetching IDs from Form Blocks.

## Addresses
https://github.com/Budibase/budibase/issues/15428

## Screenshots
When using the ID from the Form Block in the bindings, it is returned as an encoded array `%5B /* ID HERE */%5D`, but with this new helper decodeId we can extract the ID itself without any URI encoded square-brackets.
* Black is the unprocessed ID from the form block `{{ New Form Block.actor._id }}`
* Red is the new decodeId `{{ decodeId New Repeater Block.actor._id  }}`
* Blue is the existing decodeURI helper `{{ decodeURI New Repeater Block.actor._id  }}`

image

Helper also returns a comma-separated string, when passed an array of numeric IDs.
* Black is original, unprocessed array of numeric IDs `%5B1%2C2%2C3%5D`
* Red is the new decodeId`{{ decodeId "%5B1%2C2%2C3%5D"  }}`
* Blue is the existing decodeURI `{{ decodeURI "%5B1%2C2%2C3%5D"  }}`

image

## Launchcontrol
Adds a decodeId helper for cleanly retrieving IDs from form-blocks


mike12345567 (commented):
Looks good! Only thing that would be helpful is to add a test case or two for this into the string-templates test cases!

I would test both invalid cases (passing strings that simply are not an encoded ID, as well as properly encoded IDs).

mike12345567 (approved):
Tests look good to me, nice job!

PR #15552, opened 2025-02-14
Validate handlebars

adrinr:
## Description
Validating handlebars on the binding panel. For now, we will be only validating the following:
1. ~~Fields parameters~~ (TODO in a further PR)
2. Helpers parameters
3. If helper required a body, validate it is set
4. Within a body of a helper, we only validate other helpers (as we don't know what keywords can the helper use, such as `index`, `this`...)

### No errors
image

### Helper with missing parameters
image

### Helper with too many parameters
image

### Helper with missing required body 
image

### Wrong helper within a helper
image

### Invalid HBS code
image

## Launchcontrol
Validate helpers usages in hbs


packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -245,10 +250,113 @@
     ]
   }
 
+  function isMustacheStatement(
+    node: hbs.AST.Statement
```
mike12345567, line 254: This is cool - didn't realise HBS provides an AST for this!

mike12345567:
Checked this out and had a look at it - ran into a few issues - it seems handlebars statements with spaces in them are not validated correctly, for example the `Query Params` based on the URL:
image

Also components which have spaces in their name suffer from this same problem, e.g. `table - table 1` (auto-generated screens can appear like this).

adrinr:
> Checked this out and had a look at it - ran into a few issues - it seems handlebars statements with spaces in them are not validated correctly, for example the `Query Params` based on the URL: image
> 
> Also components which have spaces in their name suffer from this same problem, e.g. `table - table 1` (auto-generated screens can appear like this).

This is caused by the difference between `runtime` vs `readable` values. This required a much bigger job, as we will need to handle whitespaces properly. I will be doing this separately, as it might grow quite a lot. We will need to handle also other type of expressions, such as arrays.
Meanwhile, I updated this branch to only validate hbs helpers, as this will already add value.

aptkingston (approved):
Works well in my testing 👍  Nice job @adrinr.

Issue #15400, opened 2025-01-20
Saved value does not show in Select Picker if the value is not in the first 100 possible rows


mikesealey:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
When using a form block set to edit, if the saved value for a single-select would not appear in the first 100 rows, the value does not populate in the field. 

**To Reproduce**
1. Create a table (A) with a few hundred rows, and another table (B) with a handful of rows
2. create a relationship between the two tables, and set a relationship from TableB to a row in tableA outside of the first 100
3. create a relationship between the two tables, and set a relationship from TableB to a row in tableA from the first few rows
4. In the design section, add a table and a form block to the screen
5. Set the on-row-click action of the table to store the _id in state
6. set the form to edit, and set the edit-row-id to be the state._id
7. In Preview - switch between rows to see a row that was assigned with a relationship < 100 rows in, and a relationship > 100 rows in

Less than 100 Rows in will display the relationship as intended
![Image](https://github.com/user-attachments/assets/0eb7756a-befc-4444-aeac-aa2cb56e2fac)

No value saved
![Image](https://github.com/user-attachments/assets/94a96fe1-653b-4cc3-949d-f850d719d8a6)

Relationship exists beyond the first 100 rows
![Image](https://github.com/user-attachments/assets/319fe9dd-f1f0-4642-a335-62c9f14d8a42)

But more than 100 rows in, the user will need to open the dropdown and scroll to the 100 that contains the value before it will populate - See Jam for more context

**Expected behavior**
This value should populate when the form loads the existing values, and should not have to scroll to find the answer.

**Screenshots**
JAM https://jam.dev/c/7a83b1a5-f365-44ab-8aeb-c58c73323d0f

**App Export**
[NTG Aus Single Select-export-1737373895445.tar.gz](https://github.com/user-attachments/files/18476912/NTG.Aus.Single.Select-export-1737373895445.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 131.0.6778.205 (Official Build) (arm64)

**Additional context**
Issue raised by CSE on behalf of enterprise customer enquiry

PR #15553, opened 2025-02-14
Relationship field - make sure selected value correctly displays

mike12345567:
## Description
If the selected value of a relationship field is not within the first 100 elements then the contents will not be fetched correctly. This PR updates the component to Typescript (as well as performing a very basic conversion of some of the ancillary components in the area) and fetches if the value is not found in the options list.

Fixes: https://linear.app/budibase/issue/BUDI-8998/saved-value-does-not-show-in-select-picker-if-the-value-is-not-in-the
https://github.com/Budibase/budibase/issues/15400


packages/client/src/components/app/forms/Field.svelte
```diff
@@ -86,8 +86,9 @@
     )
   }
 
-  const updateLabel = e => {
+  const updateLabel = (e: any) => {
     if (touched) {
+      // @ts-expect-error updateProp doesn't appear to exist, needs investigation
```
aptkingston, line 91: `updateProp` is in the builder store - it's just not typescript at the moment (although Adri has a PR up), I assume that's why it's not happy? But yea it's there at line 41 in the builder store.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -1,39 +1,55 @@
...
+  export let disabled: boolean = false
+  export let readonly: boolean = false
+  export let validation: any
+  export let autocomplete: boolean = true
+  export let defaultValue: string | undefined = undefined
+  export let onChange: any
+  export let filter: SearchFilter[]
+  // not really obvious how to type this - some components pass other things here
+  // but it looks like the component data fetch should only work with tables
+  export let datasourceType: "table" = "table"
```
aptkingston, line 28: I think this was added to support the other BB reference types. For example in `BBReferenceField` we pass in `datasourceType="user"`, and looking at the possible types I think this could either be `table`, `user` or `groupUser`.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -220,13 +276,12 @@
       this={component}
       options={enrichedOptions}
       {autocomplete}
-      value={selectedValue}
+      value={componentValue()}
```
aptkingston, line 279: This won't be properly reactive, so changes in `selectedValue` won't actually result in this necessarily be re-rendered with the new value being passed in. Can we do `value={selectedValue as any}` instead?

aptkingston (commented):
This is a disgusting file hacked by too many people, so well done for successfully making a change - it's not easy. Maybe we'll be able to fully rewrite it at some point, and having TS will make that much easier so thank you!

packages/client/src/components/app/forms/RelationshipField.svelte
mike12345567: So this causes type errors on line 37 with:
```
  $: fetch = fetchData({
    API,
    datasource: {
      type: datasourceType,
      tableId: linkedTableId,
    },
    options: {
      filter,
      limit: 100,
    },
  })
```

This seems to think the only valid `datasourceType` is `"table"` - I didn't want to fiddle with the typing of data fetching, but something doesn't seem quite right there.

packages/client/src/components/app/forms/RelationshipField.svelte
mike12345567: Sadly this isn't valid Typescript, I'll see what I can do to get around it - a reactive statement casting it to something like `castedSelectedValue` will probably be the method

packages/client/src/components/app/forms/Field.svelte
mike12345567: ah this is likely the issue - I'll keep note for once this is merged/sorted out!

aptkingston (approved):
LGTM! I don't have a datasource capable of testing this in the product, but code changes look good 👍

PR #15564, opened 2025-02-17
Convert builder components helpers to ts

adrinr:
## Description
Convert builder components helpers to ts


packages/builder/src/helpers/components.ts
```diff
@@ -8,27 +8,32 @@ import {
...
+export const findComponentType = (rootComponent: Component, type: string) => {
   return searchComponentTree(rootComponent, comp => comp._component === type)
 }
 
 /**
  * Recursively searches for the parent component of a specific component ID
  */
-export const findComponentParent = (rootComponent, id, parentComponent) => {
+export const findComponentParent = (
+  rootComponent: Component | undefined,
```
deanhannigan, line 33: Could these all be set as optional?

packages/builder/src/helpers/components.ts
```diff
@@ -51,7 +56,11 @@ export const findComponentParent = (rootComponent, id, parentComponent) => {
  * Recursively searches for a specific component ID and records the component
  * path to this component
  */
-export const findComponentPath = (rootComponent, id, path = []) => {
+export const findComponentPath = (
+  rootComponent: Component,
+  id: string | undefined,
```
deanhannigan, line 61: Same here, an optional string for the id param instead of `string | undefined`

packages/builder/src/helpers/components.ts
```diff
@@ -97,17 +109,17 @@ export const findAllMatchingComponents = (rootComponent, selector) => {
...
 
 /**
  * Finds the closest parent component which matches certain criteria
  */
 export const findClosestMatchingComponent = (
-  rootComponent,
-  componentId,
-  selector
+  rootComponent: Component,
+  componentId: string | undefined,
```
deanhannigan, line 121: Possible optional value

packages/builder/src/stores/builder/components.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: analise and fix all the undefined ! and ?
```
deanhannigan, line 1: 🥚 Just checking if this was intentionally left in

deanhannigan (commented):
Everything looks grand, just some minor comments some of the optional function parameters.

`BUG` - when dragging a new component to the screen I get
  - `TypeError: Cannot read properties of null (reading '_children')`

packages/builder/src/helpers/components.ts
aptkingston: rootComponent and id shouldn't be optional (or even allowed to be undefined) but it's just because the code that calls this probably isn't all TS - so we can't guarantee that we're not passing in something undefined by mistake. The 3rd param is optional, since it's intended to be blank when called externally, but it's used when it invokes itself via recursion.

packages/types/src/ui/components/index.ts
```diff
@@ -56,3 +58,19 @@ export interface ComponentSetting {
...
 }
+
+export interface ComponentContext {
+  type: ComponentContextType
+  scope: ComponentContextScopes
+  actions: any[]
+  url?: Record<string, any>
+  query?: string
+  state?: Record<string, any>
+}
```
aptkingston, line 69: I think ComponentContext here is trying to be too many things at once.

One is a structure representing the context(s) provided by components, defined in the manifest, and that will not include `url`, `query` or `state`. For acions, it will be of the shape `{ type = "action", scope, actions }` as you have defined here, but for data context it will be of the shape `{ type = "static" | "schema" | "form", scope? = "local" | "global", suffix?: string, values?: { label, key, type }[]`. Since the function `getComponentContexts` handles both actions and data, I think the full type should be

```ts
type ComponentContextType = "action" | "static" | "schema" | "form"

interface ComponentAction {
  type: string,
  suffix?: string
}

interface ComponentStaticContextValue {
  label: string
  key: string
  type: string // technically this is a long list of options but there are too many to enumerate
}

interface ComponentContext {
  type: ComponentContextType
  scope?: ComponentContextScopes
  actions?: ComponentAction[]
  suffix?: string
  values?: ComponentStaticContextValue[]
}
```

 The other is the actual full app context, which will contain `url`, `query` and `state` as well as `embed`, `device`, `user`, `rowSelection` (deprecated) and`snippets`, and also extends `Record<string, any>`.

packages/types/src/ui/stores/preview.ts
```diff
@@ -1,2 +1 @@
 export type PreviewDevice = "desktop" | "tablet" | "mobile"
-export type ComponentContext = Record<string, any>
```
aptkingston, line 2: This is the other context I was talking about, which is the full app context. This should just be `Record<string, any>`.

aptkingston (commented):
Conversion looks good but just a query about the types 👍

packages/builder/src/stores/builder/components.ts
adrinr: Yes, this was left intentionally as it will require a much bigger refactor

packages/builder/src/helpers/components.ts
adrinr: As @aptkingston mentioned this is reflecting the current usages... Most of them should not be nullable, but until everything is fully typed will not be safe to refactor

deanhannigan (approved):
LGTM!

PR #15566, opened 2025-02-17
Add tests for cron stopping.

samwho:
## Description

This PR rounds out the testing of cron triggers by adding end-to-end tests for filling in timestamps, and making sure crons stop when they error more than 5 times.

Annoyingly it's hard to test the stopping perfectly, so I'm asserting that the job has a stopped status. I couldn't figure out a nice way to make a test against us sending the message deletion request to Bull.


packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -83,6 +87,12 @@ class InMemoryQueue implements Partial<Queue> {
   async process(concurrencyOrFunc: number | any, func?: any) {
     func = typeof concurrencyOrFunc === "number" ? func : concurrencyOrFunc
     this._emitter.on("message", async message => {
+      // For the purpose of testing, don't trigger cron jobs immediately.
+      // Require the test to trigger them manually with timestamps.
+      if (message.opts?.repeat != null) {
+        return
+      }
```
samwho, line 94: It used to be that when you called `config.api.application.publish()`, all crons were triggered straight away. This was odd behaviour to write tests against, so I decided to defer all cron triggers to be done manually in tests because it's less spooky.

packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -164,13 +174,14 @@ class InMemoryQueue implements Partial<Queue> {
...
-    // TODO: implement for testing
-    console.log(cronJobId)
+  async removeRepeatableByKey(id: string) {
+    for (const [idx, message] of this._messages.entries()) {
+      if (message.opts?.jobId?.toString() === id) {
+        this._messages.splice(idx, 1)
+        this._emitter.emit("removed", message)
+        return
+      }
+    }
```
samwho, line 184: This was put in to try and test us removing repeatable jobs from Bull but in practice it never gets called, due to my change in how crons get triggered. Because we trigger them manually, the job message doesn't have a `job.opts.repeat` anymore. I couldn't figure a way around this, we can't easily mimic Bull's repeat job functionality in our `InMemoryQueue` class.

packages/server/src/automations/tests/triggers/cron.spec.ts
```diff
@@ -49,4 +61,82 @@ describe("cron trigger", () => {
...
+    expect(results).toHaveLength(5)
+
+    await config.withProdApp(async () => {
+      const {
+        data: [latest, ..._],
+      } = await automations.logs.logSearch({
+        automationId: runner.automation._id,
+      })
+      expect(latest.status).toEqual(AutomationStatus.STOPPED_ERROR)
+    })
```
samwho, line 96: This was the best I could do for now to make sure cron jobs stop when they hit a certain number of errors (5 at present). It doesn't assert that we remove the job from the Bull queue, because in tests I'm not sure how we can do that.

packages/server/src/automations/tests/utilities/AutomationTestBuilder.ts
```diff
@@ -220,10 +220,34 @@ class AutomationRunner<TStep extends AutomationTriggerStepId> {
...
+            `Automation with ID ${
+              this.automation._id
+            } not found in app ${this.config.getAppId()}. You may have forgotten to call config.api.application.publish().`,
+            { cause: e }
+          )
+        } else {
+          throw e
+        }
+      }
+    })
```
samwho, line 250: Decided to flesh this out to make it easier to use in tests, and easier to know what to do in some common failure scenarios.

packages/server/src/automations/tests/utilities/index.ts
```diff
@@ -61,9 +101,18 @@ export async function captureAllAutomationResults(
...
     // eslint-disable-next-line no-unmodified-loop-condition
-    while (messagesReceived > 0) {
+    while (messagesOutstanding > 0) {
       await helpers.wait(50)
+      waited += 50
+      if (waited > waitMax) {
+        // eslint-disable-next-line no-unsafe-finally
+        throw new Error(
+          `Timed out waiting for automation runs to complete. ${messagesOutstanding} messages waiting for completion.`
+        )
```
samwho, line 114: Previously tests would hang forever if a message was added to the queue and not removed (as cron messages now are). Put in this failsafe to make sure the tests do eventually fail.

packages/server/src/threads/automation.ts
```diff
@@ -192,44 +181,42 @@ class Orchestrator {
...
-        "Failed to write automation metadata",
-        db.name,
-        automation._id!,
-        err
-      )
+    const metadata = await db.tryGet<AutomationMetadata>(metadataId)
+    return metadata || { _id: metadataId, errorCount: 0 }
+  }
+
+  async incrementErrorCount() {
```
samwho, line 202: I was finding that the previous method of incrementing the error count would very frequently run into document conflicts. Rewrote it to be more robust to that failure scenario.

packages/server/src/automations/tests/utilities/index.ts
```diff
@@ -34,6 +34,42 @@ export async function runInProd(fn: any) {
...
+    messages.push(message)
+  }
+
+  queue.on("message", messageListener)
+  try {
+    await f()
+    // Queue messages tend to be send asynchronously in API handlers, so there's
+    // no guarantee that awaiting this function will have queued anything yet.
+    // We wait here to make sure we're queued _after_ any existing async work.
+    await helpers.wait(100)
```
mike12345567, line 53: Is there anyway to check if something has been queued? Just wondering if in tests this 100ms could add up quite a bit

mike12345567 (approved):
LGTM - nice to have test cases for a lot of this now!

packages/server/src/automations/tests/utilities/index.ts
samwho: Unfortunately it can (and often does) happen async, so by the time we get to here the async task that does the queueing hasn't happened yet. I agree with you that this is not ideal

PR #15569, opened 2025-02-18
Test cron removal.

samwho:
## Description

In https://github.com/Budibase/budibase/pull/15566 I was annoyed that I wasn't able to find a nice way to test that cron automations get removed correctly from Bull. In this PR, I think I've figured out a nice enough way to do it.

It boils down to capturing the cron message that goes into the queue and exposing a method to re-trigger that message in tests. With this, we ensure that when the job triggers, it's using the correct cron message, and this then correctly triggers the removal from the queue when it fails 5 times.

### DIFF
--- a/packages/backend-core/src/queue/inMemoryQueue.ts
+++ b/packages/backend-core/src/queue/inMemoryQueue.ts
@@ -3,6 +3,7 @@ import { newid } from "../utils"
 import { Queue, QueueOptions, JobOptions } from "./queue"
 import { helpers } from "@budibase/shared-core"
 import { Job, JobId, JobInformation } from "bull"
+import { cloneDeep } from "lodash"
 
 function jobToJobInformation(job: Job): JobInformation {
   let cron = ""
@@ -33,7 +34,7 @@ function jobToJobInformation(job: Job): JobInformation {
   }
 }
 
-interface JobMessage<T = any> extends Partial<Job<T>> {
+export interface TestQueueMessage<T = any> extends Partial<Job<T>> {
   id: string
   timestamp: number
   queue: Queue<T>
@@ -47,15 +48,15 @@ interface JobMessage<T = any> extends Partial<Job<T>> {
  * internally to register when messages are available to the consumers - in can
  * support many inputs and many consumers.
  */
-class InMemoryQueue implements Partial<Queue> {
+export class InMemoryQueue<T = any> implements Partial<Queue<T>> {
   _name: string
   _opts?: QueueOptions
-  _messages: JobMessage[]
+  _messages: TestQueueMessage<T>[]
   _queuedJobIds: Set<string>
   _emitter: NodeJS.EventEmitter<{
-    message: [JobMessage]
-    completed: [Job]
-    removed: [JobMessage]
+    message: [TestQueueMessage<T>]
+    completed: [Job<T>]
+    removed: [TestQueueMessage<T>]
   }>
   _runCount: number
   _addCount: number
@@ -86,10 +87,13 @@ class InMemoryQueue implements Partial<Queue> {
    */
   async process(concurrencyOrFunc: number | any, func?: any) {
     func = typeof concurrencyOrFunc === "number" ? func : concurrencyOrFunc
-    this._emitter.on("message", async message => {
+    this._emitter.on("message", async msg => {
+      const message = cloneDeep(msg)
+
+      const isManualTrigger = (message as any).manualTrigger === true
       // For the purpose of testing, don't trigger cron jobs immediately.
       // Require the test to trigger them manually with timestamps.
-      if (message.opts?.repeat != null) {
+      if (!isManualTrigger && message.opts?.repeat != null) {
         return
       }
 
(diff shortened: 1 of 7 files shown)


packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -86,10 +87,13 @@ class InMemoryQueue implements Partial<Queue> {
    */
   async process(concurrencyOrFunc: number | any, func?: any) {
     func = typeof concurrencyOrFunc === "number" ? func : concurrencyOrFunc
-    this._emitter.on("message", async message => {
+    this._emitter.on("message", async msg => {
+      const message = cloneDeep(msg)
```
samwho, line 91: Hilariously this is what was causing those `loop.spec.ts` tests to pass without timing out. The orchestrator removes `appId` from the job message, which mutates the original message, causing all future invocations of that message to fail. Preserving the `appId` fixes these, but causes `loop.spec.ts` to then start timing out because it's doing a lot more work than before.

packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -133,7 +136,14 @@ class InMemoryQueue implements Partial<Queue> {
...
    * @param repeat serves no purpose for the import queue.
    */
-  async add(data: any, opts?: JobOptions) {
+  // add(name: string, data: T, opts?: JobOptions): Promise<Job<T>>;
+  async add(data: T | string, optsOrT?: JobOptions | T) {
+    if (typeof data === "string") {
+      throw new Error("doesn't support named jobs")
+    }
+
+    const opts = optsOrT as JobOptions
```
samwho, line 145: These changes were required to do the correct generic typing on the class (`InMemoryQueue<T>` as opposed to just `InMemoryQueue`).

packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -176,7 +186,7 @@ class InMemoryQueue implements Partial<Queue> {
 
   async removeRepeatableByKey(id: string) {
     for (const [idx, message] of this._messages.entries()) {
-      if (message.opts?.jobId?.toString() === id) {
+      if (message.id === id) {
```
samwho, line 189: In `getRepeatableJobs` we transform each message to a `JobInformation` object. This sets `.key` to be the same as `.id`. We're relying on that fact here, as messages are not stored with a `.key` property.

packages/server/src/automations/logging/index.ts
```diff
@@ -32,7 +32,16 @@ export async function storeLog(
...
+  try {
+    await automations.logs.storeLog(automation, results)
+  } catch (e: any) {
+    if (e.status === 413 && e.request?.data) {
+      // if content is too large we shouldn't log it
+      delete e.request.data
+      e.request.data = { message: "removed due to large size" }
+    }
+    logging.logAlert("Error writing automation log", e)
+  }
```
samwho, line 44: This handling was moved from the automation orchestrator to here because it seems to me that we should probably always log this when a log fails because it's too big.

packages/server/src/automations/tests/steps/loop.spec.ts
```diff
@@ -21,6 +21,11 @@ describe("Attempt to run a basic loop automation", () => {
   })
 
   beforeEach(async () => {
+    const { automations } = await config.api.automation.fetch()
+    for (const automation of automations) {
+      await config.api.automation.delete(automation)
+    }
```
samwho, line 27: Fix for the timeouts mentioned earlier in this review. This makes it such that there are fewer other automations to trigger when working with each test's specific automation.

packages/server/src/automations/tests/utilities/index.ts
```diff
@@ -20,6 +21,17 @@ export function afterAll() {
   config.end()
 }
 
+export function getTestQueue(): queue.InMemoryQueue<AutomationData> {
+  return getQueue() as unknown as queue.InMemoryQueue<AutomationData>
+}
```
samwho, line 26: This is what I was trying to avoid in the previous PR, but I bit the bullet and did it. I wanted to just use methods on the `Bull.Queue` interface but that was holding me back from being able to manually retrigger cron messages.

packages/server/src/automations/tests/utilities/index.ts
```diff
@@ -87,7 +131,8 @@ export async function captureAllAutomationResults(
   }
   const messageListener = async (message: Job<AutomationData>) => {
     // Don't count cron messages, as they don't get triggered automatically.
-    if (message.opts?.repeat != null) {
+    const isManualTrigger = (message as any).manualTrigger === true
```
adrinr, line 134: Could we capture this `manualTrigger` to a type, of is `Job` too attach to bull?

packages/server/src/threads/automation.ts
```diff
@@ -200,24 +189,29 @@ class Orchestrator {
...
-        logging.logAlertWithInfo(
-          "Failed to update error count in automation metadata",
-          db.name,
-          this.automation._id!,
-          err
-        )
+        return metadata.errorCount
+      } catch (error: any) {
+        err = error
+        await helpers.wait(1000 + Math.random() * 1000)
```
adrinr, line 204: Why the randomness?

adrinr (approved):

packages/server/src/threads/automation.ts
samwho: When multiple things are fighting to write to a document, a cheap way to make it less likely they'll conflict is to wait a random amount of time. It's not strictly necessary but makes conflicts less likely.

PR #15574, opened 2025-02-18
Remove mocking from sqlUtils.spec.ts.

samwho:
## Description

One day it would be nice to migrate away from Jest and use vitest or, if it's stable enough by then, Bun test. 

To do this, I've been slowly working through all of our uses of `jest.mock` and getting rid of them. This PR removes a single `jest.mock` call from `sqlUtils.spec.ts`.

To do this it required significant reworking of the test. It wasn't using any real database, so some of the constructions of tables and views didn't conform with reality.

I recommend reviewing this with the "hide whitespace" option, as a lot of lines have been indented one extra level.

CleanShot 2025-02-18 at 17 07 55@2x


packages/server/src/api/controllers/row/utils/tests/sqlUtils.spec.ts
```diff
@@ -1,511 +1,532 @@
...
-    withRelationshipColumns(
-      field: string,
-      columns: Record<string, { visible: boolean }>
-    ) {
-      this._view.schema ??= {}
-      this._view.schema[field] ??= {}
-      this._view.schema[field].columns = columns
-      return this
-    }
+        it("excludes non-sql fields fields", async () => {
```
mike12345567, line 224: fields repeated here

mike12345567 (approved):
LGTM!

PR #15578, opened 2025-02-19
Validate js helpers

adrinr:
## Description
Validate js helpers and return values.

### No errors on empty code
image

### Validates valid code
image

### Allows returning empty values
image

### Requires return values
image

### Validates valid helpers
image

### Validates wrong parameters on helpers
image
image

### Validates wrong helpers
image

## Launchcontrol
Add some js validations

## Feature branch env
[Feature Branch Link](http://fb-budi-9038-validate-js-helpers.fb.qa.budibase.net)


aptkingston (approved):
Awesome improvement @adrinr!

mike12345567 (approved):
This is very nice - great to be able to directly see where the problem is rather than needing to run it/work out what the error means!

PR #15580, opened 2025-02-19
Fix getting proper contexts when validating datasources

adrinr:
## Description
Fixing an issue raised by @poirazis. After implementing https://github.com/Budibase/budibase/pull/15433, some further merge conflicts caused this feature to work properly for links and other contexts. This PR fixes it to behave as it should.
I will be checking if we can add some unit tests around this.


mike12345567 (approved):
LGTM!

PR #15584, opened 2025-02-19
Fix CouchDB datasource.

samwho:
## Description

This started out as some mock removal in `couchdb.spec.ts` but after removing the mock I found that a couple of things about this integration were broken.

1. `testConnection` would fail if the DB specified did not exist. I suspect this used to rely on us creating a DB for pretty much every CouchDB call we made, but we changed that a little while back. I changed how it worked to query documents and if that call succeeds, we're connected.
2. Deleting records never passed in a `_rev`, and so can never have worked. This has been fixed.

## Launchcontrol

- Several fixes to the CouchDB datasource: fixed testing connection, fixed delete queries.


mike12345567 (approved):
LGTM!

PR #15586, opened 2025-02-19
Improve APM traces around automations.

samwho:
## Description

This PR adds more detail to our automation APM traces in DataDog.

### DIFF
--- a/packages/backend-core/src/context/mainContext.ts
+++ b/packages/backend-core/src/context/mainContext.ts
@@ -123,7 +123,7 @@ export async function doInAutomationContext<T>(params: {
   task: () => T
 }): Promise<T> {
   await ensureSnippetContext()
-  return newContext(
+  return await newContext(
     {
       tenantId: getTenantIDFromAppID(params.appId),
       appId: params.appId,
--- a/packages/server/src/automations/utils.ts
+++ b/packages/server/src/automations/utils.ts
@@ -40,39 +40,35 @@ function loggingArgs(job: AutomationJob) {
 }
 
 export async function processEvent(job: AutomationJob) {
-  return tracer.trace(
-    "processEvent",
-    { resource: "automation" },
-    async span => {
-      const appId = job.data.event.appId!
-      const automationId = job.data.automation._id!
+  return tracer.trace("processEvent", async span => {
+    const appId = job.data.event.appId!
+    const automationId = job.data.automation._id!
 
-      span?.addTags({
-        appId,
-        automationId,
-        job: {
-          id: job.id,
-          name: job.name,
-          attemptsMade: job.attemptsMade,
-          opts: {
-            attempts: job.opts.attempts,
-            priority: job.opts.priority,
-            delay: job.opts.delay,
-            repeat: job.opts.repeat,
-            backoff: job.opts.backoff,
-            lifo: job.opts.lifo,
-            timeout: job.opts.timeout,
-            jobId: job.opts.jobId,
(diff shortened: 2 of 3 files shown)


packages/backend-core/src/context/mainContext.ts
```diff
@@ -123,7 +123,7 @@ export async function doInAutomationContext<T>(params: {
   task: () => T
 }): Promise<T> {
   await ensureSnippetContext()
-  return newContext(
+  return await newContext(
```
samwho, line 126: I'm noticing in DataDog that our automation traces look incomplete, and part of me is wondering if it's because we're not awaiting some promises. This seemed like a potential culprit, and should be a semantically safe change.

packages/server/src/automations/utils.ts
```diff
@@ -40,39 +40,35 @@ function loggingArgs(job: AutomationJob) {
...
+        delay: job.opts.delay,
+        repeat: job.opts.repeat,
+        backoff: job.opts.backoff,
+        lifo: job.opts.lifo,
+        timeout: job.opts.timeout,
+        jobId: job.opts.jobId,
+        removeOnComplete: job.opts.removeOnComplete,
+        removeOnFail: job.opts.removeOnFail,
+        stackTraceLimit: job.opts.stackTraceLimit,
+        preventParsingData: job.opts.preventParsingData,
```
samwho, line 65: These have been unnested because they appear to get ignored when they're nested past a certain level.

packages/server/src/threads/automation.ts
```diff
@@ -310,87 +310,83 @@ class Orchestrator {
...
-        const ctx: AutomationContext = {
-          trigger: trigger.outputs,
-          steps: [trigger.outputs],
-          stepsById: {},
-          stepsByName: {},
-          user: trigger.outputs.user,
-        }
-        await enrichBaseContext(ctx)
+    return await tracer.trace("execute", async span => {
+      span.addTags({ appId: this.appId, automationId: this.automation._id })
```
samwho, line 314: I don't know why GitHub is being so weird about this diff, it should just be an indentation change.

packages/server/src/threads/automation.ts
```diff
@@ -446,74 +442,92 @@ class Orchestrator {
...
+      return [stepFailure(step, { status: AutomationStatus.NO_CONDITION_MET })]
+    })
   }
 
   private async executeStep(
     ctx: AutomationContext,
     step: Readonly<AutomationStep>
   ): Promise<AutomationStepResult> {
-    return tracer.trace("Orchestrator.executeStep", async span => {
+    return await tracer.trace(step.stepId, async span => {
```
samwho, line 530: Gave this a more specific name so that it shows up nicer in the trace.

packages/server/src/threads/automation.ts
```diff
@@ -557,6 +572,7 @@ class Orchestrator {
         ;(outputs as any).status = AutomationStatus.STOPPED
       }
 
+      span.addTags({ outputsKeys: Object.keys(outputs) })
```
samwho, line 575: Didn't want to fully add the inputs and outputs as they could be PII, so just adding the keys so we have an idea of the shape of things.

mike12345567 (approved):
LGTM - I'm not sure the `return await` will change anything, because returning the promise directly should be identical, the only difference is that if the awaited function throws an error it is now going to be thrown within that function rather than being thrown back to the receiver awaiting it.

If the promise isn't being awaited where the `ensureSnippetContext` is being called then that might explain it, but we should be handling that already.

PR #15587, opened 2025-02-20
Ops/aas with minio gateway

calexiou:
## Description
Configures minio to run under gateway mode to allow persistent storage when Budibase is running in Azure App Service. Requires `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` to be set when executing the container.

### DIFF
--- a/hosting/single/runner.sh
+++ b/hosting/single/runner.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-echo "Starting runner.sh"
+echo "Starting runner.sh..."
 
 # set defaults for Docker-related variables
 export APP_PORT="${APP_PORT:-4001}"
@@ -96,7 +96,9 @@ if [[ -n "${REDIS_PASSWORD}" ]]; then
 else
     redis-server "${REDIS_CONFIG}" >/dev/stdout 2>&1 &
 fi
-/bbcouch-runner.sh &
+
+echo "Starting callback CouchDB runner..."
+./bbcouch-runner.sh &
 
 # only start minio if use s3 isn't passed
 if [[ -z "${USE_S3}" ]]; then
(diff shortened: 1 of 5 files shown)


hosting/single/runner.sh
```diff
@@ -79,31 +69,42 @@ ln -s ${DATA_DIR}/.env /worker/.env
 # make these directories in runner, incase of mount
 mkdir -p ${DATA_DIR}/minio
 mkdir -p ${DATA_DIR}/redis
-chown -R couchdb:couchdb ${DATA_DIR}/couch
+#mkdir -p ${DATA_DIR}/couch
+#chown -R couchdb:couchdb ${DATA_DIR}/couch
```
samwho, line 73: How come these are no longer required?

hosting/single/runner.sh
```diff
@@ -19,14 +19,7 @@ declare -a DOCKER_VARS=("APP_PORT" "APPS_URL" "ARCHITECTURE" "BUDIBASE_ENVIRONME
...
 #  export CUSTOM_DOMAIN=budi001.custom.com
 
-# Azure App Service customisations
-if [[ "${TARGETBUILD}" = "aas" ]]; then
-    export DATA_DIR="${DATA_DIR:-/home}"
-    WEBSITES_ENABLE_APP_SERVICE_STORAGE=true
-    /etc/init.d/ssh start
-else
-    export DATA_DIR=${DATA_DIR:-/data}
-fi
```
samwho, line 29: How come this is no longer required?

calexiou: https://github.com/Budibase/budibase/blob/master/hosting/couchdb/runner.sh#L14

Same reason as above - it's handled in the `bbcouch-runner.sh` script.

hosting/single/runner.sh
calexiou: https://github.com/Budibase/budibase/blob/master/hosting/couchdb/runner.sh#L7

The `bbcouch-runner.sh` script that's being called a few lines below is creating the same directories and changing the permissions as well.

adrinr (commented):
Did we run a test of this single image and on feature-branches?

calexiou:
> Did we run a test of this single image and on feature-branches?

We did, yes. Here's the run -> https://github.com/Budibase/budibase-deploys/actions/runs/13437085341 and [here's the branch](https://github.com/Budibase/budibase-deploys/commit/598d4f81d3d5011111895e561cfd099b38030a5f) that hardcodes `budibase/budibase:experimental`. The tag contains the Dockerfile and runner script changes from this PR's branch.

The feature branch [behaves the same way ](https://fb-master.fb.qa.budibase.net/) as the regular ones.

adrinr (approved):

PR #15590, opened 2025-02-20
Type sidepanel

adrinr:
## Description
Type sidepanel.
Also, adding an eslint rule that ensures that types are properly imported in svelte files on builder. Without this, we might have some runtime errors as it tries to find types that are not on the final bundler, possibly breaking some components on runtime


PClmnt (approved):

mike12345567 (approved):
LGTM!

PR #15595, opened 2025-02-20
Move snippets to sidepanel

adrinr:
## Description
Moving snippets section in bindings in order to match the UX of helpers and other bindings. We discussed keeping the existing section for crud, but this made it harder to understand and to reuse some logic. Happy to discuss improvements and different approaches.

### Snippets will be moved to the bindings side panel
image

### Snippets will not appear on the text section
https://github.com/user-attachments/assets/aaa5545d-97f4-4e94-85b9-83d6a95c42f0

### Snippets can be added from the side panel

https://github.com/user-attachments/assets/71933e44-3e1f-4e59-a8ed-7bf3604ff270

### Snippets can be edited from the side panel
https://github.com/user-attachments/assets/3153976f-487d-445d-b185-f9e5af5096a2

### Snippets can be deleted from the side panel
https://github.com/user-attachments/assets/ac7f30e0-e1b9-462e-a249-1b9e1ed7933c

### Upgrade flow will still be in there for non-licenced users

https://github.com/user-attachments/assets/22ee70ca-e340-4713-b2e6-eb06f3c3a30d

## Launchcontrol
Moving snippets section to the binding side panel

## Feature branch env
[Feature Branch Link](http://fb-budi-9068-move-snippets-to-sidepanel.fb.qa.budibase.net)


packages/builder/src/components/common/bindings/BindingSidePanel.svelte
```diff
@@ -295,10 +409,53 @@
...
+              {/if}
+            </div>
+          {/if}
+        </div>
+      {/if}
     {/if}
   </Layout>
 </div>
 
+{#if showSnippetDrawer}
```
adrinr, line 452: This is done as it was the easier way to clean the state. Otherwise, when creating multiple snippets, the state would be persisted instead of starting from a clean slate everytime

aptkingston: Was there any way to work around this? It's not a big problem, but unmounting the component like this when hiding means we lose the drawer close animation - it just instantly disappears with a small flash rather than fading out with an animation.

Not super important but would be a nice to have I guess. Happy to keep it this way since I assume you already tried resetting state instead, but just wanted to note the downside of doing it like this.

aptkingston (approved):
LGTM! Just the one comment about losing the animation due to unmounting the snippet drawer, but not a big problem. All works perfectly 👍

adrinr:
> LGTM! Just the one comment about losing the animation due to unmounting the snippet drawer, but not a big problem. All works perfectly 👍

Good shout. I updated it to handle it internally

PR #15608, opened 2025-02-24
Fix bug with branches when no conditions are set.

samwho:
## Description

When you first create a branch in the automations UI, a branch step with 2 options is created and the default conditions set are:

```json
{ "onEmptyFilter": "none", "$and": { "conditions": [] } }
```

It's the `onEmptyFilter` that was causing problems, a check in `isLogicalFilter` wasn't expecting it and throws an exception. However, that exception was being swallowed and ignored. I've fixed both of those things in this PR.

## Launchcontrol

- Fix lack of automation test result output if your automation contained a branch step with no conditions set. This will now fail as expected.


mike12345567 (approved):
LGTM!

PR #15609, opened 2025-02-24
Support filtering relationships by _id

samwho:
## Description

Due to an oversight on my part, it wasn't possible to filter on the `_id` field of relationships. This PR fixes that.

## Addresses
- https://linear.app/budibase/issue/BUDI-9055/provide-a-filtering-option-for-a-relationships-id-when-using-the

## Launchcontrol

- Support filtering relationships by their `_id` field.


mike12345567 (approved):
LGTM!

PR #15611, opened 2025-02-24
Fix for automation traversal

deanhannigan:
## Description
The `terminated` flag was not being set on the step references when traversing. This would manifest in the UI as automation flows that would seemingly come to an unexpected end and `Collect Data` would no longer be a selectable step type.

## Addresses
- https://linear.app/budibase/issue/BUDI-9079/cannot-add-collect-data-step-as-ui-no-longer-recognises-the-end-of-an

## Launchcontrol
Minor fix to ensure terminated flag is set when traversing automations.


samwho (approved):

shogunpurple (approved):
LGTM

PR #15623, opened 2025-02-25
Typescript conversion for some BBUI components

PClmnt:
## Description
Converts a number of BBUI components to Typescript


packages/bbui/src/Stores/notifications.ts
```diff
@@ -2,9 +2,21 @@ import { writable } from "svelte/store"
...
+  action: (() => void) | null
+  actionMessage: string | null
+  wide: boolean
+  dismissTimeout: number
+}
+
 export const createNotificationStore = () => {
-  const timeoutIds = new Set()
-  const _notifications = writable([], () => {
+  const timeoutIds = new Set<NodeJS.Timeout>()
```
adrinr, line 18: We had some issues with this when running on the frontend. 
```suggestion
  const timeoutIds = new Set<ReturnType<typeof setTimeout>>()
```

packages/bbui/src/Table/CellRenderer.svelte
```diff
@@ -8,14 +8,14 @@
   import InternalRenderer from "./InternalRenderer.svelte"
   import { processStringSync } from "@budibase/string-templates"
 
-  export let row
-  export let schema
-  export let value
-  export let customRenderers = []
-  export let snippets
+  export let row: any
```
adrinr, line 11: Should this be `Record<string, any>`?

packages/bbui/src/Table/CellRenderer.svelte
```diff
@@ -8,14 +8,14 @@
...
+  export let schema: any
```
adrinr, line 12: Same with the schema. Should this be Record<string, any>?

packages/bbui/src/Table/CellRenderer.svelte
```diff
@@ -8,14 +8,14 @@
...
+  export let value: any
+  export let customRenderers: { column: string; component: any }[] = []
+  export let snippets: any
```
adrinr, line 15: Same

packages/bbui/src/Table/DateTimeRenderer.svelte
```diff
@@ -1,13 +1,13 @@
...
-  export let value
-  export let schema
+  export let value: string | Date
+  export let schema: { timeOnly?: boolean; dateOnly?: boolean }
 
   // adding the 0- will turn a string like 00:00:00 into a valid ISO
   // date, but will make actual ISO dates invalid
   $: time = new Date(`0-${value}`)
-  $: isTimeOnly = !isNaN(time) || schema?.timeOnly
+  $: isTimeOnly = !isNaN(time.getTime()) || schema?.timeOnly
```
adrinr, line 10: Is this expected? Would this not change its behaviour?

PClmnt: So isNan only takes a number, but we were passing a date to it here, but it was getting coerced internally by isNan. I've casted it to any instead, just to be safe in this instance and to not change any behaviour it relies on.

adrinr (approved):
LGTM!

PR #15624, opened 2025-02-25
Test ElasticSearch integration against a real ElasticSearch docker container

samwho:
## Description

Part of my ongoing quest to remove mocking from the codebase, this PR removes our global `@elastic/elasticsearch` mock from the `packages/server` tests.

@mike12345567 could you add the elasticsearch variant of `test-server` to the required steps for CI, please? 🙏


packages/server/src/api/routes/tests/datasource.spec.ts
```diff
@@ -165,7 +165,8 @@ describe("/datasources", () => {
 })
 
 const descriptions = datasourceDescribe({
-  exclude: [DatabaseName.MONGODB, DatabaseName.SQS],
+  plus: true,
```
samwho, line 168: You'll notice a lot of changes like this one dotted around. I didn't want adding a new datasource to the list of datasources to mean it gets added to all of the places that specify just an `exclude` list on `datasourceDescribe`. I noticed all of the places that do this are doing tests that should pass for all DS+, so I've added this parameter to constrain those.

mike12345567 (approved):
LGTM - very nice!

PR #15626, opened 2025-02-25
Remove mocks from static.spec.ts

samwho:
## Description

Fairly easy one, it mostly passed on its own. Just needed some asserts around signed URLs updating.


mike12345567 (approved):
LGTM!

PR #15629, opened 2025-02-25
Rewrite relationship field

aptkingston:
## Description
This PR is a full rewrite of our `RelationshipField` which fixes countless bugs and improves performance.

I've done my best to test as many configurations as possible but there could well still be a bug hidden somewhere.

Known bug fixes:
- Searching doesn't work at all
- Default values of strings with commas would not work if the IDs were outside the first page
- If the default value is an invalid ID it is impossible to deselect this value as there is no option for it
- Default values flashed raw IDs while waiting for API requests

Known improvements:
- You can now endlessly scroll through row options to keep loading more
- Huge API request reduction (up to 4X fewer search requests in some configurations)
- All options are cached once encountered so they never go missing if updating search term
- No flashing or jank. If we're loading options for default values which aren't in our option set, we show a blank value until we have them.
- Gracefully handle IDs which are invalid and ensure it's possible to deselect them
- Avoid API requests entirely if we don't need them - this is the case when using something like an update form which includes a relationship. The row will contain a basic relationship structure of just `_id` and `primaryDisplay` for each relationship, and this is enough information for us to render options for these rows without having to search.


deanhannigan (approved):
LGTM!

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -1,17 +1,23 @@
...
   import Field from "./Field.svelte"
   import type {
     SearchFilter,
     RelationshipFieldMetadata,
-    Table,
     Row,
   } from "@budibase/types"
-
-  const { API } = getContext("sdk")
+  import TableFetch from "@budibase/frontend-core/src/fetch/TableFetch"
```
adrinr, line 18: This should not be needed if this one gets applied: https://github.com/Budibase/budibase/pull/15650/commits/fa2357c1d34ead3a5ef7b628775d94bdec6bc2d5

adrinr (changes requested):

adrinr (approved):
🚀

PR #15634, opened 2025-02-26
Group UI improvements

mike12345567:
## Description
Issues have been raised around how groups function - such as: https://linear.app/budibase/issue/BUDI-9066/adding-users-to-groups-does-not-grant-access-to-apps-associated-with

Groups work as expected, however looking at the UI its not clear at all that the users are receiving access to the app via the groups they are part of - we have fixed this in two places:

1. In the user portal, we now correctly list the number of apps users have access to, based on their direct roles and all of their group assignments.
2. In the user side panel (in app builder) it now displays that the user is within groups (that pertain to this app) and that their app access is controlled by the group.

## Screenshots
image


melohagan (approved):
LGTM - the only thing would be to see what happens for very long group names, and make sure everything is reasonably formatted

packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
```diff
@@ -741,16 +744,33 @@
...
                 <div class="auth-entity">
                   <div class="details">
-                    <div class="user-email" title={user.email}>
-                      {user.email}
+                    <div class="user-groups">
+                      <div class="user-email" title={user.email}>
+                        {user.email}
+                      </div>
+                      {#if userGroups.length}
+                        <div class="group-info">
```
Comment (deanhannigan) [inline packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:759]: This takes up a lot of space alongside the email With longer emails it pushes the UI completely out of site.  It may work better to move the group count under the user email or just remove the '2 groups' and leave the icon?

Screenshot 2025-02-27 at 09 07 54

packages/shared-core/src/sdk/documents/users.ts
```diff
@@ -129,3 +130,30 @@ export function containsUserID(value: string | undefined): boolean {
...
+  groups?: UserGroup[]
+) {
+  const prodAppId = getProdAppID(appId)
+  const userGroups = getUserGroups(user, groups)
+  return userGroups.filter(group =>
+    Object.keys(group.roles || {}).find(app => app === prodAppId)
+  )
+}
+
+export function userAppAccessList(user: User, groups?: UserGroup[]) {
```
deanhannigan, line 152: Couldn't this be used to populate the user app list in the portal?

deanhannigan (commented):
Hey @mike12345567, had a run through this, this morning and didn't run into any issues! 

Just one question. In the linked ticket it shows in the user portal that, even with group access to an app, their app list still says `This user doesn't have access to any apps`.  While testing, my users still show this behaviour in the portal. Is this supposed to be populated now?

packages/shared-core/src/sdk/documents/users.ts
mike12345567: Yeap - sorry I missed the app list in the portal - thanks for catching that!

packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte
Comment (mike12345567) [reply packages/builder/src/pages/builder/app/[application]/_components/BuilderSidePanel.svelte:759]: Thanks for pointing this out, I've gone with truncating long email addresses as that is what would happen if the email was very long already.
image

packages/shared-core/src/sdk/documents/users.ts
mike12345567: Updated this to a similar experience in the user side panel.

image

deanhannigan (approved):
LGTM! Gave this a rinse locally and it's working perfectly. A really great update! It definitely makes group access permissions way easier to understand in both the portal and builder! 🎉

PR #15638, opened 2025-02-26
Improve types around `nodemailer` (take 2)

samwho:
Reverts the initial revert at Budibase/budibase#15637

I've pointed out in a comment where the bug from the first attempt was. I've substantially extended the testing around email sending in `packages/worker`. Instead of doing it half with Jest mocks and half by firing off real emails to `seamus99@ethereal.email`, it now spins up an SMTP server in the tests using https://github.com/maildev/maildev and sends off real emails, then retrieves them through `maildev`'s API. This results in much better test realism, and better assertions on the results of the tests.

## Feature branch env
[Feature Branch Link](http://fb-revert-15637-revert-15612-smtp-types.fb.qa.budibase.net)


packages/types/src/api/web/global/email.ts
```diff
@@ -10,19 +11,18 @@ export enum EmailTemplatePurpose {
 }
 
 export interface SendEmailRequest {
-  workspaceId?: string
   email: string
-  userId: string
+  userId?: string
```
samwho, line 15: Automations don't send this, it has always been optional.

packages/types/src/api/web/global/email.ts
```diff
@@ -10,19 +11,18 @@ export enum EmailTemplatePurpose {
...
   purpose: EmailTemplatePurpose
   contents?: string
   from?: string
   subject: string
-  cc?: boolean
-  bcc?: boolean
+  cc?: string
+  bcc?: string
```
samwho, line 21: These were just flatly wrong, assuming an oversight in the original TS conversion.

packages/worker/src/api/controllers/global/email.ts
```diff
@@ -24,13 +23,15 @@ export async function sendEmail(
...
   } = ctx.request.body
-  let user: any
+  let user: User | undefined = undefined
   if (userId) {
     const db = tenancy.getGlobalDB()
-    user = await db.get<User>(userId)
+    user = await db.tryGet<User>(userId)
+    if (!user) {
+      ctx.throw(404, "User not found.")
+    }
```
samwho, line 32: This is where the original bug was. I had this `if (!user)` check outside of the `if (userId)` block, meaning whenever a `userId` wasn't specified (e.g. in automations), it would always return a 404.

This was an oversight on my part that we had no tests to catch. I've added tests that catch this.

packages/worker/src/utilities/email.ts
```diff
@@ -158,10 +143,21 @@ export async function sendEmail(
...
+    case EmailTemplatePurpose.PASSWORD_RECOVERY:
+      if (!opts.user || !opts.user._id) {
+        throw new HTTPError("User must be provided for password recovery.", 400)
+      }
+      code = await cache.passwordReset.createCode(opts.user._id, opts.info)
+      break
+    case EmailTemplatePurpose.INVITATION:
+      code = await cache.invite.createCode(email, opts.info)
+      break
+  }
```
samwho, line 157: This got moved from `getLinkCode` and I added the checks to `opts.user` because the newly corrected typing was showing the user of `opts.user` to be incorrect.

packages/worker/src/utilities/email.ts
```diff
@@ -106,8 +91,8 @@ async function buildEmail(
     throw "Unable to build email, missing base components"
   }
 
-  let name = user ? user.name : undefined
-  if (user && !name && user.firstName) {
+  let name: string | undefined
```
mike12345567, line 94: this change, just to check, `user.name` - is that a property that was ever populated/even exists? I'm guessing that is why it was removed from here as I don't think thats a prop our users have

packages/worker/src/utilities/email.ts
mike12345567: Out of interest, what was the reason for lifting it out of `getLinkCode`?

packages/worker/src/tests/mocks/email.ts
```diff
@@ -1,10 +1,170 @@
-export function mock() {
-  // mock the email system
-  const sendMailMock = jest.fn()
-  const nodemailer = require("nodemailer")
-  nodemailer.createTransport.mockReturnValue({
-    sendMail: sendMailMock,
-    verify: jest.fn(),
-  })
-  return sendMailMock
+import MailDev from "maildev"
```
mike12345567, line 1: This is fantastic - one thing that comes to mind is that we should probably make this mail server available through `backend-core` as I suspect @melohagan will be very interested in using this in the account portal as well.

mike12345567 (approved):
LGTM - great to get an iCal invite test case together as well, this is a much better method of testing this, avoids the multitude of issues we've ran into with Ethereal being unavailable etc as well as giving us a lot more options in terms of how we test email.

packages/worker/src/tests/mocks/email.ts
samwho: I'm led to believe Mel is trying to reduce the account-portal's reliance on backend-core, so it's more likely a copy-paste job. Either way, I wrote it in a way that should be easily portable.

packages/worker/src/utilities/email.ts
samwho: As far as I can tell it doesn't exist.

packages/worker/src/utilities/email.ts
samwho: `getLinkCode` was only used here and had as many arguments as it did significant lines of code. It's probably a matter of taste more than anything, but it felt like unnecessary indirection to me.

mike12345567: Was the `getLinkCode` function removed in that case? If it is no longer used we should get rid of it (although it maybe used in the account portal potentially)

samwho: It was, yes 🙂 It was only in this file and not exported.

PR #15639, opened 2025-02-26
Fix MSSQL tests.

samwho:
## Description

A while ago we hit a problem where our `mssql` Docker container would fail to run on Linux kernel version 6.8.0, which was the kernel used on `ubuntu-latest` at the time. So we pinned to `ubuntu-22.04`, which used kernel 6.5.0.

`ubuntu-22.04` recently updated from 6.5.0 to 6.8.0 and as a result our `mssql` tests have been failing. This PR does a couple of things:

1. Updates to `ubuntu-24.04`, which is also kernel 6.8.0, seeing as we don't have any other reason to pin to an older runner.
2. Updates our `mssql` image from 2019 to 2022, as this newer version runs successfully on 6.8.0.
3. Updates our `mssql` container healthchecks to support that TLS is now on by default and the location of the tool we use to do the healthchecking has moved.


mike12345567 (approved):
LGTM - thanks for investigating this - good to get past the kernel issue!

PR #15641, opened 2025-02-27
Skip ancestors validation on plugins components

adrinr:
## Description
After adding the new component ancestor validations, we realised it is not fully compatible with plugin components (raised by @poirazis). These components are served built, so we don't know their inner component structure. Because of this, we cannot validate ancestors properly, and it will always complain.
This PR skips this validation if there is a custom component in the ancestor chain. Also, we update the order of this validation to validate first the ancestors requirement before any setting validation.

Also, removing an obsolete index.js that was left over after the ts refactor

## Screenshots
image

## Launchcontrol
Skip ancestors validation on plugins components


mike12345567 (approved):
LGTM!

PR #15642, opened 2025-02-27
Handle spaces on hbs binding validations

adrinr:
## Description
Recently we introduced Handlebars binding validation. This uncovered some issues around the way we handle readable vs runtime values. Specifically, the runtime values are wrapped around square brackets, and this was dropped on the readable version. Handlebars does not read this as a single parameter, instead, it reads as many parameters as words separated by whitespace we have.
image

This PR fixes this, keeping the brackets when the value contains whitespace.
### In bindings
image

### In formulas

https://github.com/user-attachments/assets/6b1c622d-b2f7-4f42-acef-83ba769f7603

It will not be wrapped between [] if there are no whitespace
image

### Bindings with whitespace and complex keys
image

image

During this PR we also found out a couple of other issues, fixed as well:
1. Validate optional parameters on the hbs expressions
image
image

2. Validate parameters on functions with body
image
image

3. Validate that bodies are only allowed for functions that accept them
image

4. Make date helpers parameters optional
image

## Launchcontrol
Fix whitespaces handling and improve handlebars binding validations

## Feature branch env
[Feature Branch Link](http://fb-budi-9082-handle-spaces-on-binding-va.fb.qa.budibase.net)


packages/string-templates/scripts/gen-collection-info.ts
```diff
@@ -163,7 +159,6 @@ function run() {
         .map(tag => tag.description!.replace(/`/g, "").split(" ")[0].trim())
       collectionInfo[name] = fixSpecialCases(name, {
         args,
-        numArgs: args.length,
```
adrinr, line 166: This is not used anywhere, and it just makes the manifest bigger for no reason

mikesealey:
image

Passing simply `{{ date }}` still seems to return a valid output, despite the validation requesting 2-3 arguments. Is this a valid use-case of `{{ date }}`?

adrinr:
> {{ date }

That's a good shout. In theory, just passing `{{ date }}` is not supported, but I can see quite a lot of usage for it. I will check if we can support the validation for it

adrinr:
> > {{ date }
> 
> That's a good shout. In theory, just passing `{{ date }}` is not supported, but I can see quite a lot of usage for it. I will check if we can support the validation for it

Done!

packages/builder/src/components/backend/DataTable/formula.js
```diff
@@ -62,6 +62,12 @@ export function getBindings({
 
     const label = path == null ? column : `${path}.0.${column}`
     const binding = path == null ? `[${column}]` : `[${path}].0.[${column}]`
+
+    let readableBinding = label
+    if (readableBinding.includes(" ")) {
+      readableBinding = `[${readableBinding}]`
```
mike12345567, line 68: Is this safe? If the binding is something like `Named Repeater.rows.0.a column` will it not lead to `[Named Repeater.rows.0.a column]` suggesting that the entire binding is a constant string, ignoring the dot notation?

packages/string-templates/scripts/gen-collection-info.ts
```diff
@@ -34,15 +33,13 @@ const outputJSON: Manifest = {}
 const ADDED_HELPERS = {
   date: {
     date: {
-      args: ["datetime", "format"],
-      numArgs: 2,
+      args: ["[datetime]", "[format]", "[options]"],
```
mike12345567, line 36: Why are the square brackets necessary here?

packages/string-templates/src/manifest.json
```diff
@@ -91,7 +82,6 @@
         "a",
         "b"
       ],
-      "numArgs": 2,
```
mike12345567, line 94: Removing the `numArgs` - is this used anywhere/displayed in the builder?

packages/builder/src/components/backend/DataTable/formula.js
adrinr: You are right. Updated it to support dot notation

packages/string-templates/scripts/gen-collection-info.ts
adrinr: This is the [JSDoc](https://jsdoc.app/tags-param) code for optional parameters

mike12345567 (approved):
LGTM - very nice work!

PR #15643, opened 2025-02-27
Remove mocks from public variant of `users.spec.ts`

samwho:
## Description

This PR removes the mocking in the public variant of `users.spec.ts`. To enable this I've had to introduce a way to hit the public API that's consistent with how we hit the internal API in tests. I also decided to add the same error stack trace propagation to our public API during testing that we have on our internal API, to make debugging easier.


packages/server/src/api/routes/public/tests/users.spec.ts
```diff
@@ -1,132 +1,121 @@
...
+    .post(`/api/global/users`)
+    .reply(200, (uri, body) => {
+      const newUser = body as User
+      if (!newUser._id) {
+        newUser._id = `us_${generator.guid()}`
+      }
+      users[newUser._id!] = newUser
+      return newUser
+    })
+    .persist()
```
samwho, line 41: These two HTTP mocks create a very crude implementation of the worker API for creating and retrieving users. It just takes what it's given and stores it into a map by the user ID.

In an ideal world, I'd like to _actually_ hit the worker API here. I don't see why we wouldn't be able to. That or we start the work to merge the server and worker packages.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -46,33 +46,34 @@ export interface RequestOpts {
 export abstract class TestAPI {
   config: TestConfiguration
   request: SuperTest<Test>
+  prefix = ""
```
samwho, line 49: Overridden later in the `PublicAPI` subclass.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -236,3 +239,24 @@ export abstract class TestAPI {
     ).body
   }
 }
+
+export abstract class PublicAPI extends TestAPI {
+  prefix = "/api/public/v1"
```
samwho, line 244: Chosen specifically so that the URL we pass in children of this class will match up with the URL specified in the routes files.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -236,3 +239,24 @@ export abstract class TestAPI {
...
+    const headers: Record<string, string | string[]> = {
+      Accept: "application/json",
+      Host: this.config.tenantHost(),
+      "x-budibase-api-key": apiKey,
+      "x-budibase-app-id": this.config.getAppId(),
+      ...extras,
+    }
+
+    return headers
+  }
```
samwho, line 261: This override ensures that public API methods are always authed using a valid API key, which is created if it doesn't already exist.

packages/server/src/api/routes/public/middleware/testErrorHandling.ts
```diff
@@ -0,0 +1,24 @@
+import { Ctx } from "@budibase/types"
+import environment from "../../../../environment"
+
+export default async (ctx: Ctx, next: any) => {
+  try {
+    await next()
+  } catch (err: any) {
+    if (
+      !(environment.isTest() && ctx.headers["x-budibase-include-stacktrace"])
```
adrinr, line 9: Could we throw an error if this is used outside `environment.isTest()` at register time instead of runtime?

samwho: Good idea! Done.

mike12345567 (approved):
LGTM!

PR #15657, opened 2025-03-03
Fix app deletion issue in development

mike12345567:
## Description
Fixes an issue found by @deanhannigan and @aptkingston - if your object store is empty then deletion would fail under the new AWS SDK.

Also - appears like apps can zombie in this situation, however I believe this is just app cache not getting invalidated correctly - I've made sure app invalidation works as expected.

### DIFF
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -478,7 +478,7 @@ export async function deleteFolder(
   if (existingObjectsResponse.Contents?.length === 0) {
     return
   }
-  const deleteParams: any = {
+  const deleteParams: { Bucket: string; Delete: { Objects: any[] } } = {
     Bucket: bucketName,
     Delete: {
       Objects: [],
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -489,10 +489,12 @@ export async function deleteFolder(
     deleteParams.Delete.Objects.push({ Key: content.Key })
   })
 
-  const deleteResponse = await client.deleteObjects(deleteParams)
-  // can only empty 1000 items at once
-  if (deleteResponse.Deleted?.length === 1000) {
-    return deleteFolder(bucketName, folder)
+  if (deleteParams.Delete.Objects.length) {
```
adrinr, line 492: Does this mean we will leave empty buckets? Is this expected?

mike12345567: Its a single bucket used for all apps - so if we didn't add anything theres simply nothing to do - we use static buckets for everything, so deleting them would make it harder to create new things.

packages/server/src/api/routes/tests/application.spec.ts
```diff
@@ -1,4 +1,7 @@
 import { DEFAULT_TABLES } from "../../../db/defaultData/datasource_bb_default"
+import { setEnv } from "../../../environment"
+
+setEnv({ USE_LOCAL_COMPONENT_LIBS: "0" })
```
samwho, line 4: nit: to avoid changing global state in this test, I'd recommend putting this in `beforeAll`, assigning the result to `envCleanup: () => void`, and then calling it in `afterAll`.

samwho: Also if this is what triggers MinIO to get used, I'd put a comment above it explaining that.

adrinr: Jest don't allow changing global state, so it should not be neede

mike12345567: Done!

samwho (approved):

PR #15659, opened 2025-03-03
De-mock ArangoDB tests.

samwho:
## Description

Makes the ArangoDB test hit an actual ArangoDB container, removing the mocks in the process.


mike12345567:
Going to close this - I'll create a separate PR to remove the test cases and disable the integration in the UI.

PR #15665, opened 2025-03-03
Group performance improvements

mike12345567:
## Description
The primary purpose of this PR is to introduce two functions, `isCreatorSync` and `isCreatorAsync` - the first takes a list of groups and determines if the user is a creator, including checking the groups and the second will retrieve the groups and check the user (only taking the user as a parameter).

This is useful as there are many places where lots of users need to be assessed to check if they are a creator, in this case its better to retrieve a list of all the groups that pertained to those users and then synchronously checking the users against the groups - previously when this was necessarily the list of groups would be retrieved over and over again.

The existing group test cases confirm this doesn't affect the functionality of the system - the reason the async function exists is really as a utility incase you are doing this for a single user, you still need the full list of groups that pertain to them, this function encapsulates that logic for situations like middlewares where it needs to check a single user.

Pro PR: https://github.com/Budibase/budibase-pro/pull/356


packages/backend-core/src/users/utils.ts
```diff
@@ -16,30 +16,55 @@ export const hasAdminPermissions = sdk.users.hasAdminPermissions
 export const hasBuilderPermissions = sdk.users.hasBuilderPermissions
 export const hasAppBuilderPermissions = sdk.users.hasAppBuilderPermissions
 
-export async function isCreator(user?: User | ContextUser) {
+async function getGroups(groupIds: string[]) {
+  if (groupIds.length) {
```
samwho, line 20: This feels like an optimisation we should do at the `getMultiple` level. It's broadly applicable.

packages/backend-core/src/users/utils.ts
```diff
@@ -16,30 +16,55 @@ export const hasAdminPermissions = sdk.users.hasAdminPermissions
...
+  }
+  return []
+}
+
+export async function creatorsInList(
+  users: (User | ContextUser)[],
+  groups?: UserGroup[]
+) {
+  if (!groups) {
+    const groupIds = users.flatMap(user => user.userGroups || [])
```
samwho, line 32: Probably want to dedup this list as well, right?

packages/backend-core/src/users/utils.ts
```diff
@@ -16,30 +16,55 @@ export const hasAdminPermissions = sdk.users.hasAdminPermissions
...
+    if (groupIds.length) {
```
samwho, line 33: You don't need this check thanks to the optimisation you've done in `getGroups`. This then allows you to get rid of the `|| []` a few lines below.

samwho (approved):

PR #15666, opened 2025-03-03
Type BBUI multiselect

adrinr:
## Description
Type BBUI multiselect


packages/bbui/src/Form/Core/Multiselect.svelte
```diff
@@ -1,22 +1,26 @@
-<script>
+<script lang="ts" context="module">
+  type Option = any
```
PClmnt, line 2: Is Option an `any` though? At least for the label it can probably only be string, right?

adrinr: This component supports any kind of `options` object, and we have the props `getOptionLabel/Value` to extract what is needed

PClmnt (approved):

PR #15669, opened 2025-03-04
More BBUI types and tidy up

PClmnt:
## Description
More BBUI types and deleted a few unneeded files


packages/bbui/src/Layout/Layout.svelte
```diff
@@ -1,12 +1,12 @@
...
-  export let alignContent = "normal"
-  export let justifyItems = "stretch"
+<script lang="ts">
+  export let horizontal: boolean = false
+  export let paddingX: "S" | "M" | "L" | "XL" | "XXL" = "M"
+  export let paddingY: "S" | "M" | "L" | "XL" | "XXL" = "M"
+  export let noPadding: boolean = false
+  export let gap: "XXS" | "XS" | "S" | "M" | "L" | "XL" = "M"
+  export let noGap: boolean = false
+  export let alignContent: string = "normal"
```
adrinr, line 8: Could this be typed? It should be a close set of values, right?

packages/bbui/src/Modal/CustomContent.svelte
```diff
@@ -1,11 +1,11 @@
-<script>
+<script lang="ts">
   import { getContext } from "svelte"
   import Context from "../context"
 
-  const { hide } = getContext(Context.Modal)
+  const { hide } = getContext(Context.Modal) as { hide: () => void }
```
adrinr, line 5: Could we "autotype" it as we do using in `packages/client/src/context.d.ts`?

adrinr (approved):

PR #15674, opened 2025-03-04
Speed up `loop.spec.ts`

samwho:
## Description

This should solve the timeouts we've been seeing with `loop.spec.ts` recently. I've left comments inline explaining, it was quite a journey to debug.

CleanShot 2025-03-04 at 18 04 13@2x

### DIFF
--- a/globalSetup.ts
+++ b/globalSetup.ts
@@ -88,6 +88,16 @@ export default async function setup() {
           content: `
           [log]
           level = warn
+
+          [httpd]
+          socket_options = [{nodelay, true}]
+
+          [couchdb]
+          single_node = true
+
+          [cluster]
+          n = 1
+          q = 1
         `,
           target: "/opt/couchdb/etc/local.d/test-couchdb.ini",
         },
--- a/packages/backend-core/src/queue/inMemoryQueue.ts
+++ b/packages/backend-core/src/queue/inMemoryQueue.ts
@@ -3,7 +3,6 @@ import { newid } from "../utils"
 import { Queue, QueueOptions, JobOptions } from "./queue"
 import { helpers } from "@budibase/shared-core"
 import { Job, JobId, JobInformation } from "bull"
-import { cloneDeep } from "lodash"
 
 function jobToJobInformation(job: Job): JobInformation {
   let cron = ""
@@ -89,7 +88,7 @@ export class InMemoryQueue<T = any> implements Partial<Queue<T>> {
   async process(concurrencyOrFunc: number | any, func?: any) {
     func = typeof concurrencyOrFunc === "number" ? func : concurrencyOrFunc
     this._emitter.on("message", async msg => {
-      const message = cloneDeep(msg)
+      const message = msg
 
       // For the purpose of testing, don't trigger cron jobs immediately.
       // Require the test to trigger them manually with timestamps.
@@ -165,6 +164,9 @@ export class InMemoryQueue<T = any> implements Partial<Queue<T>> {
         opts,
       }
       this._messages.push(message)
+      if (this._messages.length > 1000) {
+        this._messages.shift()
+      }
       this._addCount++
       this._emitter.emit("message", message)
     }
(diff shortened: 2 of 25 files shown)


globalSetup.ts
```diff
@@ -88,6 +88,16 @@ export default async function setup() {
...
+
+          [httpd]
+          socket_options = [{nodelay, true}]
+
+          [couchdb]
+          single_node = true
+
+          [cluster]
+          n = 1
+          q = 1
```
samwho, line 100: I went through https://docs.couchdb.org/en/stable/maintenance/performance.html and enabled a few things. I don't think it had much of a performance impact, but it did get rid of some errors in the logs so I'm going to leave it in.

packages/server/src/api/routes/tests/viewV2.spec.ts
samwho, line 1: The quota stuff in this file, and a few others, is leftover from my attempt to speed up tests by disabling quota checking everywhere except tests that relied on it. I've gotten rid of it now but liked these changes, so kept them.

packages/server/src/definitions/automations.ts
```diff
@@ -20,9 +20,12 @@ export interface TriggerOutput {
 
 export interface AutomationContext {
   trigger: AutomationTriggerResultOutputs
-  steps: [AutomationTriggerResultOutputs, ...AutomationStepResultOutputs[]]
-  stepsById: Record<string, AutomationStepResultOutputs>
+  steps: Record<
```
samwho, line 23: Previously we would track context in this shape and before each time we used it, we'd call a function called `prepareContext` that slightly transformed it. This ended up being quite expensive, so I made it such that the transformation is no longer needed.

packages/server/src/definitions/automations.ts
```diff
@@ -31,4 +34,6 @@ export interface AutomationContext {
     company?: string
   }
   loop?: { currentItem: any }
+  _stepIndex: number
+  _error: boolean
```
samwho, line 38: `_stepIndex` tracks what numerical index to use next for step output (e.g. for `{{ steps.0.output }}`), and `_error` tracks whether a step has failed. This removes the need for us to loop over all of the output to check.

packages/server/src/events/AutomationEmitter.ts
```diff
@@ -32,6 +32,8 @@ class AutomationEmitter implements ContextEmitter {
 
     if (chainAutomations === true) {
       return MAX_AUTOMATIONS_ALLOWED
+    } else if (env.isTest()) {
+      return 0
```
samwho, line 36: This was the key to solving the reason `loop.spec.ts` has been timing out. We had lots of automations that triggered themselves, which was causing havoc in the background while other tests were running. This prevents that, unless chaining is explicitly enabled in the app metadata.

packages/server/src/tests/utilities/api/base.ts
```diff
@@ -146,8 +146,9 @@ export abstract class TestAPI {
       }
     }
 
+    let resp: Response | undefined = undefined
     try {
-      return await req
+      resp = await req
```
samwho, line 151: Did this to make debugging easier.

packages/server/src/threads/automation.ts
```diff
@@ -304,41 +292,37 @@ class Orchestrator {
...
   }
 
   async execute(): Promise<AutomationResults> {
     return await tracer.trace("execute", async span => {
       span.addTags({ appId: this.appId, automationId: this.automation._id })
 
-      const job = cloneDeep(this.job)
-      delete job.data.event.appId
-      delete job.data.event.metadata
+      const data = cloneDeep(this.job.data)
```
samwho, line 302: This was another big part of why automation tests became slow. `this.job` contains a reference to `InMemoryQueue`, which in turn has a reference to `_messages` which contains all automation queue messages fired during tests. Cloning it in full was expensive.

packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -165,6 +162,9 @@ export class InMemoryQueue<T = any> implements Partial<Queue<T>> {
         opts,
       }
       this._messages.push(message)
+      if (this._messages.length > 1000) {
+        this._messages.shift()
```
adrinr, line 166: Can this have a side effect? Could we not remove it when processed?

adrinr (approved):
LGTM! Thanks for the detailed comments!

Issue #15670, opened 2025-03-04
[Internal DB] views that have public access can't upload attachments if the base table's permissions are higher than it


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: self-host - docker-compose
  - Budibase Version: 3.4.22
  - App Version: 3.4.22

**Describe the bug**
I have a table called attachments. The base table permissions are for app admin. I created a view for this table that public users will use, including the attachment column. Permissions are set to public on this view. 

If I publish the app and access the page as a non logged-in user, fill out the details and then try to upload an attachment. I get hit with a notification saying `Not Authorized`.

```json
err: {
    "type": "ForbiddenError",
    "message": "Not Authorized",
    "stack":
    ForbiddenError: Not Authorized
    at Object.throw (/Users/name/budibase/node_modules/koa/lib/context.js:97:11)
    at checkAuthorized (/Users/name/budibase/packages/server/dist/index.js:187172:25)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/name/budibase/packages/server/dist/index.js:187236:9
    at async middleware3 (/Users/name/budibase/packages/server/dist/index.js:174406:20)
    at async /Users/name/budibase/node_modules/koa-compress/lib/index.js:38:5
    at async errorHandling (/Users/name/budibase/packages/server/dist/index.js:141930:5)
    at async /Users/name/budibase/node_modules/koa-mount/index.js:52:26
    at async userAgent (/Users/name/budibase/node_modules/koa-useragent/dist/index.js:12:5)
    at async contentSecurityPolicy (/Users/name/budibase/packages/server/dist/index.js:142013:5)
    at async ip_default (/Users/name/budibase/packages/server/dist/index.js:142174:16)
    "status": 403,
    "statusCode": 403,
    "expose": true
}
```

**To Reproduce**
1. Import the attached app
2. Publish the app
3. View as a non logged in user
4. Try and upload attachments to the attachment field
5. See error

**Expected behavior**
Views with public permissions should allow the ability to upload attachments.

You shouldn't have to set the base table to public in order for this to work.
  
**App Export**
[public attachments-export-1741079442694.tar.gz](https://github.com/user-attachments/files/19067344/public.attachments-export-1741079442694.tar.gz)

PR #15679, opened 2025-03-05
Allow uploading attachments through views

aptkingston:
## Description
This PR updates the attachment field to allow uploading via the view ID (therefore using the view permissions) rather than using the table ID.

## Addresses
- https://github.com/Budibase/budibase/issues/15670
- https://linear.app/budibase/issue/BUDI-9100/%5Binternal-db%5D-views-that-have-public-access-cant-upload-attachments-if


deanhannigan (approved):
LGTM! Ran through the attached app and verified that I could successfully upload attachments to permitted views.

Issue #15676, opened 2025-03-04
Clear Selected Rows action does not find any tables to interact with


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
Using the "Clear Selected Rows" action should provide an option of each of the tables on the screen.
![Image](https://github.com/user-attachments/assets/d4947fc4-646c-459a-97dc-e78e8b8201c3)

**To Reproduce**
1. Add a Table component to the screen, with a data source.
2. Add a button with an "On Click" action to "Clear Selected Rows"

**Expected behavior**
Dropdown should display all available tables

**Screenshots**

  
**App Export**
[Clear Row Selection-export-1741110326093.tar.gz](https://github.com/user-attachments/files/19075296/Clear.Row.Selection-export-1741110326093.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 133.0.6943.142 (Official Build) (arm64)

**Additional context**
Issue raised by CSE on behalf of Premium customer

PR #15684, opened 2025-03-06
Documentation link in App Action

mikesealey:
## Description
This PR adds options for showing a link to relevant documentation in the Inline Alert, as well as implementing this in the "Clear Selected Rows" action.

## Addresses
- https://github.com/Budibase/budibase/issues/15676 Raised by CSE on behalf of a customer - it wasn't clear that this action was for the deprecated Table Block.

## Screenshots
image

Other implementations of `InlineAlert.svelte` remain unchanged

image


packages/bbui/src/InlineAlert/InlineAlert.svelte
```diff
@@ -49,6 +51,20 @@
...
+    <div id="docs-link">
+      <a href={link} target="_blank" rel="noopener noreferrer" class="doc-link">
+        {linkText}
+        <svg
+          class="spectrum-Icon spectrum-Icon--sizeS spectrum-Toast-typeIcon"
+          focusable="false"
+          aria-hidden="true"
+        >
+          <use xlink:href="#spectrum-icon-18-LinkOut" />
+        </svg>
```
aptkingston, line 64: It would be nicer to use our Icon component (just import it from "../Icon/Icon.svelte"), but it'll function the same! Style wise I would just make it a touch smaller and with a little space between the link and the text.

aptkingston (commented):
Just a small style tweak suggested!

packages/bbui/src/InlineAlert/InlineAlert.svelte
aptkingston: Using our icon, you can make it smaller by just doing
```svelte
<Icon name="LinkOut" size="XS" /> 
```
for example.

aptkingston (approved):

PR #15686, opened 2025-03-06
Convert middleware tests to TS, remove mocks.

samwho:
## Description

It appeared that `selfhost.ts` middleware was no longer used, so I deleted it and its tests.


mike12345567 (approved):
LGTM!

PR #15696, opened 2025-03-07
New automation JS editor fixes.

mike12345567:
## Description
A couple of small fixes for automation JS editor experience, making sure that the first use of `$` correctly jumps to within the binding and handling template string usage.

While it is a bit hacky, it seems that the first use of a completion just simply will not re-anchor the cursor, so breaking the cursor setting out of it to be after the text has been altered seems to rectify this issue.

EDIT: appears these problems are global - affecting all JS bindings.


deanhannigan (approved):
LGTM! Had a run through this with @mike12345567 and confirmed that the behaviour was across all JS binding inputs in the builder. Nice one!

Issue #15658, opened 2025-03-03
REST API import is broken


casidiablo:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
I tried this in the cloud version, and then ran a local version (Docker) just to figure out where the error comes from.

**Describe the bug**

I wanted to import an OpenAPI spec, but get this error always:

<img width="744" alt="Image" src="https://github.com/user-attachments/assets/f6a52d42-07af-4815-840b-2e83587d443c" />

**To Reproduce**
1. Go to the Data tab
2. Add a new REST API data source
3. Try to import any file into it

**Expected behavior**

It should create the datasource out of the OpenAPI spec

**Screenshots**

In both cases (cloud and locally on Docker), I see that there's an error on the JS, even before the file is sent to budibase.

<img width="1115" alt="Image" src="https://github.com/user-attachments/assets/1b30598b-0532-48a1-bdaa-519d0edd5676" />

And yes, I have tried this in different browsers

  
**App Export**

No relevant

**Desktop (please complete the following information):**

Not relevant

**Additional context**

ConorWebb96:
Hey @casidiablo,

Thanks for bringing this up. I replicated the issue using the collections provided in our documentation. The following documentation files work when imported into Postman and other REST GUIs:

[Collection docs link](https://docs.budibase.com/docs/public-api#getting-the-openapi-specification)

mike12345567:
Fixed in: https://github.com/Budibase/budibase/pull/15698 - will be included in the next release.

PR #15698, opened 2025-03-07
Import OpenAPI fixes

mike12345567:
## Description
There was an issue with the recent Typescript refactors of stores, the `queries` store used to have an alias for `import` which is no longer available, converted everything in the area to Typescript and updated to use the `importQueries` function instead.

## Addresses
https://linear.app/budibase/issue/BUDI-9099/rest-api-import-is-broken
https://github.com/Budibase/budibase/issues/15658


packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/panels/Queries/RestImportQueriesModal.svelte
```diff
@@ -55,9 +58,9 @@
       const body = {
         data: dataString,
         datasourceId,
+        datasource,
       }
-
-      const importResult = await queries.import(body)
+      const importResult = await queries.importQueries(body)
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/data/datasource/[datasourceId]/_components/panels/Queries/RestImportQueriesModal.svelte:63]: This is the actual fix, from `queries.import` -> `queries.importQueries`

PClmnt (approved):

packages/builder/src/components/backend/DatasourceNavigator/modals/ImportRestQueriesModal.svelte
mike12345567, line 1: Was duplicated and not used, removed.

Issue #15700, opened 2025-03-08
Custom S3 Endpoints broken


CSLTech:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Describe the bug**
The custom S3 endpoints are broken. This seems to be a regression introduced in 52d8d7d1dcfd52ca922c28ad11ce1755a0d86d96

**To Reproduce**
1. Create an S3 endpoint with a custom endpoint
2. The code still tries to contact the default endpoint

**Expected behavior**
It should try to contact the custom endpoint

mike12345567:
Release going out now - fix will contain it.

PR #15701, opened 2025-03-08
Fix S3 integration custom endpoints usage

CSLTech:
## Description
Custom S3 endpoints are broken. This fixes the S3 integration usage of custom S3 endpoints

## Addresses
-  #15700 

## Launchcontrol

Fix S3 integration when using custom endpoints

### DIFF
--- a/packages/server/src/integrations/s3.ts
+++ b/packages/server/src/integrations/s3.ts
@@ -168,6 +168,7 @@ class S3Integration implements IntegrationBase {
         secretAccessKey: config.secretAccessKey,
       },
       region: config.region,
+      endpoint: config.endpoint,
     }
     if (config.endpoint) {
       this.config.forcePathStyle = true


samwho:
Hey @CSLTech, thanks for the contribution!

I'll get this merged when the tests pass.

samwho:
Actually, @mike12345567 pointed out to me that folks may have set this to invalid values and forgotten they've done it since it does nothing right now. Enabling this could break them.

What we'd like to do is introduce a new config parameter called `customEndpoint`, and deprecate this `endpoint` parameter. Would you be able to make this change?

What you need to do is add `deprecated: true` to the config entry in the `SCHEMA` on line 45. Then add a new `customEndpoint` to the schema, as well as the to `S3Config` type above the schema. Then plumb it in like you did with `config.endpoint` 🙏

CSLTech:
Ermm, I can make the change, but this should be somewhat documented as this PR in it's current state fixes an issue that is a regression from a working behavior. Moreover if they had it misconfigured it would have made the signed links behave in a (Most likely broken) weird way.

samwho (approved):

PR #15706, opened 2025-03-10
Type fields

adrinr:
## Description
Type fields components


packages/client/src/components/app/forms/Field.svelte
```diff
@@ -1,43 +1,56 @@
...
+  import type { Readable } from "svelte/store"
   import { writable } from "svelte/store"
   import { Icon } from "@budibase/bbui"
   import { memo } from "@budibase/frontend-core"
   import Placeholder from "../Placeholder.svelte"
   import InnerForm from "./InnerForm.svelte"
-  import type { FieldApi } from "."
+  import type { FieldApi, FieldState } from "."
+  import type { FieldSchema, FieldType } from "@budibase/types"
+  import type { FieldValidation, FormField } from "@/index"
```
aptkingston, line 11: The types for these field values seem to be stored in quite a few different locations - would it be better to add a dedicated `fields` or `forms` file to the types package and contain them all there instead? Or even a more generic `client` one to contain the types that we currently just store in the root `index.ts`?

aptkingston (commented):
Changes all look good, just one question about where we store all these new types 👍

aptkingston (approved):
LGTM!

PR #15709, opened 2025-03-10
Default app design improvements

aptkingston:
## Description
This PR makes multiple improvements to how new apps look and feel, with the goal of giving a better impression to new users.

 Full list of changes:
- All new apps (from scratch only) come with sample data automatically. This does not include imports or templates.
- All new apps (from scratch only) come with a screen already created, which is a table screen template with sideabar for the inventory table. This table includes conditions (highlighting a row in red) and formatting (adding a currency symbol to a price). This helps showcase some of these features.
- The inventory table data and schema has been improved to make the data more interesting in the sample screen.
- A new single text component has been added, replacing headline and paragraph. This component supports markdown formatting and has a new component to configure it - a text input which dynamically grows in size up to 5 rows and is full width in the right panel. To make life easier, searching for "headline" or "paragraph" in the new component panel will show this new "text" component in the results panel. This component is also used in the new sample screen.
- The nav is now blue by default, with white text
- Nav links styling has been improved - they have a background color selected state and padding and margin has been adjusted
- The broken "portal" link in the nav has been replaced with a new user avatar with a dropdown menu. This menu has multiple actions: update details (same as the portal, for changing name), update password (same as the portal), go to portal (to go back to the app list for normal users or go to the builder for builders), and log out.
- Table styles have been improved. The whitespace at the bottom has been shrank, header font is bold, row number colors contrast lowered, and the "quiet" setting has been updated to add multiple opinionated style changes - no header icons, no contrast for header background).
- There are new "style" options for table header background color and stripe row color (if choosing to stripe rows)
- Checkboxes in the table are hidden if they serve no purpose
- New color options added to color picker, adding static colours which don't change with the app theme. This is used for the new nav background.

## Addresses
- (most of) https://linear.app/budibase/issue/BUDI-8493/app-navigation-updates
- https://docs.google.com/document/d/1ApIU1IM5qVH6vTlYTCeym_N_8tR2GXNbL6xNyJ9mK98/edit?tab=t.qx1jlguei4hr

### New sample screen which comes by default, also showing new nav and new quiet table styles
image

### New user menu in apps (hidden for public users)
image

### New side nav styles, showing new user menu for side navs
image

### New text component and multiline text setting
image

### New color options in color picker
image


aptkingston:
@mike12345567 if you could just glance over the server changes please that would be great 🙏 They are significantly smaller than the rest of this PR, so should hopefully be quick to review!

packages/server/src/api/controllers/application.ts
```diff
@@ -293,11 +304,25 @@ async function performAppCreate(
...
     const db = context.getAppDB()
+    const isImport = !!instanceConfig.file
+    const addSampleData = !isImport && !useTemplate
 
     if (instanceConfig.useTemplate && !instanceConfig.file) {
       await updateUserColumns(appId, db, ctx.user._id!)
     }
 
+    // Add link to sample data page if required
+    const newLinks = addSampleData
```
mike12345567, line 315: Could we perhaps break out the navigation updates into an `addSampleNavigation` function like we have for the others parts of the sample data? I was looking for a minute trying to work out where the screen got added to the nav.

mike12345567 (approved):
LGTM on server-side, a couple of minor comments, I do wonder about whether sample screens/tables should be added to the backup stats, but thats not something this PR should worry about and its a problem today, its just going to be highlighted more by this work.

packages/server/src/api/controllers/application.ts
aptkingston: The links are part of the app metadata, so if I break them out into a function then I'd need to do another update of the metadata after the app is created or something which is wasteful, since I can just add it in the initial doc that is created.

Either that or just have a function that mutates the metadata passed in and adds that one link, but that seems a little convoluted... but would still move it out at least. I'll add that if you'd prefer it. This at least won't do another couch read/write.

mike12345567: I don't necessarily think that doing a separate write is a bad idea, my reasoning being that the navigation has been updated here before the data/screens has been added yet, in the same vein as my previous comment, the app is going to exist even if the sample data/screens fail to be created, so theres a potential scenario right now where the navigation has a link to a screen that does not exist yet.

Doing this as a separate update after the data/screens has been added in the `if (addSampleData) ...` block means that if something fails there, the nav will be the last thing to be updated, so worst case the nav will never have links to non-existent screens added to it.

deanhannigan (changes requested):
So many fantastic quality of life improvements! Nicely done! 🎉 

Gave this a run through locally and its really solid. Only ran into a couple of minor issues:

- `BUG` - log out enabled and usable in the preview (portal and builder preview modal)
- `BUG` - Portal button is now active and usable in the preview (portal and builder preview modal).
  - This was previously disabled to keep users from leaving the preview or infinitely navigating through the portal.

packages/bbui/src/Form/Core/TextArea.svelte
```diff
@@ -55,8 +76,10 @@
     {disabled}
     {readonly}
     {id}
+    on:input={onChange}
     on:focus={() => (focus = true)}
-    on:blur={onChange}
+    on:blur={onBlur}
+    on:blur
```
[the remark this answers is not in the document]
aptkingston: This is deliberate - one is a custom handler and the other means that the blur event is proxied up the chain to parent components. Svelte supports adding multiple handlers like this 👍

aptkingston:
> So many fantastic quality of life improvements! Nicely done! 🎉
> 
> Gave this a run through locally and its really solid. Only ran into a couple of minor issues:
> 
> * `BUG` - log out enabled and usable in the preview (portal and builder preview modal)
> * `BUG` - Portal button is now active and usable in the preview (portal and builder preview modal).
>   
>   * This was previously disabled to keep users from leaving the preview or infinitely navigating through the portal.

Cheers for giving this a run through.

Yea I'm aware everything is functional in the preview, and to be honest I thought it was better to let it be functional. If someone really wants to recurse into a million iframes then good for them 😂 

I suppose log out is slightly different, in that if you log out while using the preview iframe it'll kick you back to login in the iframe, but the parent window will also have been logged out but not visually updated - so if they close the iframe they'll get a 403 on the next action they take, and also be kicked back to login. So log out specifically should probably be disabled - but I think I'll leave the portal link functional.

deanhannigan:
> I suppose log out is slightly different, in that if you log out while using the preview iframe it'll kick you back to login in the iframe, but the parent window will also have been logged out but not visually updated - so if they close the iframe they'll get a 403 on the next action they take, and also be kicked back to login. So log out specifically should probably be disabled - but I think I'll leave the portal link functional.

Agreed 👍

aptkingston:
Log out and portal links are both disabled when loaded inside an iframe, or embedded. That accounts for the in builder preview modal, when viewing via the portal, as well as any custom embed usages.

image

I just did the portal too since for some of those use cases it probably doesn't make sense to let them leave the app (specifically authenticated embeds).

deanhannigan (approved):
LGTM! 🚀

mike12345567 (approved):
LGTM - thanks for making that change!

aptkingston:
Spectrum 1200 colour variants have been manually added. They've also been added to the colour picker, and are now used by default for the app accent theme and nav background.

image

PR #15711, opened 2025-03-11
Initial scaffolding for using LLMs to generate JS in Budibase.

samwho:
## Description

This feature is behind a feature flag while in development, so this PR won't be releasing anything to users.

What's in here:

- A new endpoint, `POST /api/ai/js` that the frontend talks to as a proxy to whatever LLM we use.
- Some new UI elements for showing a "generate with AI" button on JS codeblocks.
- A little bit of TypeScript conversion in bbui for good measure.

Pro PR: https://github.com/Budibase/budibase-pro/pull/359


packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -139,6 +149,46 @@
     }
   }
 
+  // Please write me a function to get the current number of seconds past the hour.
```
PClmnt, line 152: Don't think we need this comment?

PClmnt (approved):
LGTM!

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
samwho: Whoops, I left it there so I could copy it when I needed it instead of typing it out each time 😅

packages/bbui/src/Form/TextArea.svelte
```diff
@@ -5,30 +5,39 @@
 
   export let value: string | undefined = undefined
   export let label: string | undefined = undefined
-  export let labelPosition: string = "above"
+  export let labelPosition = "above"
```
adrinr, line 8: Should this be using the position enum, instead?

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -139,6 +149,45 @@
...
+    promptLoading = true
+    popoverWidth = 30
+    const { code } = await API.generateJs({ prompt })
+    value = code
+    editor.dispatch({
+      changes: { from: 0, to: editor.state.doc.length, insert: code },
+    })
+    suggestedCode = code
+    popoverWidth = 100
+    promptLoading = false
```
adrinr, line 167: Should this be wrapped in a try/finally? Otherwise it will be stuck to loading forever

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -428,6 +477,50 @@
   <div tabindex="-1" bind:this={textarea} />
 </div>
 
+{#if aiGenEnabled}
+  <button
```
adrinr, line 481: Is there a reason on why not using bbui buttons? Are the designs not matching?

packages/server/src/api/controllers/ai.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export async function generateJs(
+  ctx: UserCtx<GenerateJsRequest, GenerateJsResponse>
+) {
+  if (!(await features.isEnabled(FeatureFlag.AI_JS_GENERATION))) {
+    ctx.throw(400, "AI JS generation is disabled")
+  }
+
+  if (!env.OPENAI_API_KEY) {
+    ctx.throw(400, "OpenAI API key is not set")
```
adrinr, line 20: These 400 look strange to me. None of them is "bad request", both of them are due server conditions, no because of the request

packages/server/src/api/routes/ai.ts
```diff
@@ -0,0 +1,9 @@
+import Router from "@koa/router"
+import * as controller from "../controllers/ai"
+import { auth } from "@budibase/backend-core"
+
+const router: Router = new Router()
+
+router.post("/api/ai/js", auth.builderOrAdmin, controller.generateJs)
```
adrinr, line 7: Should this not be builder only?

packages/bbui/src/Form/TextArea.svelte
samwho: Which enum are you referring to? I dug into the components this gets passed to and they are all either using `string` or not typed yet.

packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
samwho: This started life as a <div>, but that caused a lint error about divs not being keyboard accessible. I didn't know how to attach a class to a <Button> so I went with <button>.

It is highly likely this will change before we ship this, my plan was to ask Cheeks or Peter to have a once over and make it look nice.

packages/server/src/api/routes/ai.ts
samwho: I'm not sure to be honest. Why not admin as well?

packages/bbui/src/Form/TextArea.svelte
adrinr: Oh... We don't have a type for this. Probably we should add one, but we can leave it outside this PR scope

packages/server/src/api/routes/ai.ts
adrinr: You are right, this is on the app scope!

adrinr (approved):
LGTM!

Issue #15714, opened 2025-03-11
Validation text on Lorem helper incorrect


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
Validation text for `{{ lorem }}` suggests that it requires 1 parameter, however it is actually set up to work without a parameter, as well as with a single numeric value as a parameter.
![Image](https://github.com/user-attachments/assets/1726a814-966d-4166-9893-60b963e13162)

`{{ lorem }} -> "Lorem ipsum"`
`{{ lorem 25 }} -> // Returns 25 characters of Lorem Ipsum`

**To Reproduce**
1. Open the bindings drawer and enter `{{ lorem }}`
2. See the validation text

**Expected behavior**
Validation should account for no parameter being passed

PR #15716, opened 2025-03-11
Updates Lorem Helper validation

mikesealey:
## Description
`{{ Lorem }}` helper will work without a number-perameter and return `Lorem ipsum` (equivilent to` {{ lorem 11 }}`)

This PR makes a change to the parameter so that the validaiton understand that the number is optional

Please note: This PR contains a 1-line change, and _many_ lint changes.
My work can be found here:
image

## Addresses
https://github.com/Budibase/budibase/issues/15714

## Screenshots
Before:
image

After:
image
(No validation warnings)

## Launchcontrol
_Updates validation on Lorem helper_


adrinr (changes requested):
This file is autogenerated, it should come from https://github.com/Budibase/handlebars-helpers
I know that is some tedious extra work, but it would be nice to change it in there and to get it being autogenerated. Otherwise, on the next "update" your changes will be lost.
This would require:
1. Update the parameter change in https://github.com/Budibase/handlebars-helpers
2. Merge it to master
3. Publish a new version (@mike12345567 can help with it)
4. On this branch, update to use the latest handlebars-helpers package and run `yarn manifest` from within `package/string-templates`

PR #15719, opened 2025-03-11
Don't show row test data selector outside of test modal.

samwho:
## Description

Oversight on my part, I didn't realise the code for the test modal was reused for steps within the automation. Thanks @deanhannigan for noticing this!

## Launchcontrol

- Fixes automation row update step showing UI elements it shouldn't have.

### DIFF
--- a/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
+++ b/packages/builder/src/components/automation/SetupPanel/AutomationBlockSetup.svelte
@@ -387,6 +387,55 @@
         ]
       }
 
+      const getTestDataSelector = () => {
+        if (!isTestModal) {
+          return []
+        }
+        return [{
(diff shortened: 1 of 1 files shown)


deanhannigan (changes requested):
LGTM! One minor thing, could you add something like this to around line `222` to ensure the selected row is populated on load? Right now it only populated when a change is triggered in the test modal, rather than on load.

```javascript
  const updateSelectedRow = testData => {
    selectedRow = testData?.row
  }

  $: updateSelectedRow(testData)

  $: customStepLayouts(
    $memoBlock,
    schemaProperties,
    stepState,
    fetchedRows,
    selectedRow
  )
  ```

deanhannigan (approved):
LGTM! Had another quick click through and everything's resolved 🚀

PR #15732, opened 2025-03-13
Initial oAuth2 config crud

adrinr:
## Description
Basic scaffolding and basic implementation of the OAuth2 config endpoints. Don't pay much attention to the data persisted, I kept it simple and I extended it when needed in the usage of this [configs](https://github.com/Budibase/budibase/pull/15733).
The following deletion/updates/etc. will be implemented as needed


packages/types/src/api/web/app/index.ts
samwho, line 1: ❤️

samwho (approved):

PR #15733, opened 2025-03-13
Use OAuth2 on rest

adrinr:
## Description
Initial implementation of OAuth2 usages. This PR supports setting an OAuth2Config in a rest endpoint (backend only), obtaining and using the generated token.
The PR tests 2 different scenarios:
1. The generation of the tokens, using a real oauth2 service via test containers
2. The usage of an oauth2 configuration in rest, using nock


packages/server/images-sha.env
```diff
@@ -4,4 +4,5 @@ POSTGRES_SHA=sha256:bd0d8e485d1aca439d39e5ea99b931160bd28d862e74c786f7508e9d0053
 MONGODB_SHA=sha256:afa36bca12295b5f9dae68a493c706113922bdab520e901bd5d6c9d7247a1d8d
 MARIADB_SHA=sha256:e59ba8783bf7bc02a4779f103bb0d8751ac0e10f9471089709608377eded7aa8
 ELASTICSEARCH_SHA=sha256:9a6443f55243f6acbfeb4a112d15eb3b9aac74bf25e0e39fa19b3ddd3a6879d0
-DYNAMODB_SHA=sha256:cf8cebd061f988628c02daff10fdb950a54478feff9c52f6ddf84710fe3c3906
\ No newline at end of file
+DYNAMODB_SHA=sha256:cf8cebd061f988628c02daff10fdb950a54478feff9c52f6ddf84710fe3c3906
+KEYCLOCK_IMAGE=quay.io/keycloak/keycloak@sha256:2ce6c7c70994c70dbbd70b372a5422c3b4eebb32583175eac03751320609e52c
```
samwho, line 8:
```suggestion
KEYCLOAK_IMAGE=quay.io/keycloak/keycloak@sha256:2ce6c7c70994c70dbbd70b372a5422c3b4eebb32583175eac03751320609e52c
```

samwho, line 8: You'll also want to pre-pull this in tests, somewhere around here: https://github.com/Budibase/budibase/blob/0e676f018365af79d602b1acef43131656ec3344/.github/workflows/budibase_ci.yml#L215

packages/server/src/sdk/app/oauth2/tests/utils.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        { source: volumePath, target: "/opt/keycloak/data/import/" },
+      ])
+      .withCommand(["start-dev", "--import-realm"])
+      .withWaitStrategy(Wait.forLogMessage("Listening on: http://0.0.0.0:8080"))
+      .withStartupTimeout(60000)
+      .start()
+
+    keycloakUrl = `http://${container.getHost()}:${container.getMappedPort(
+      8080
+    )}`
```
samwho, line 37: You _probably_ don't want to rely on `testcontainers` to get you the right port. We've seen trouble with that in the past around ipv6/ipv4. See the comment here: https://github.com/Budibase/budibase/blob/0e676f018365af79d602b1acef43131656ec3344/packages/backend-core/tests/core/utilities/testContainerUtils.ts#L186-L193

I recommend using that `startContainer` function here so you don't have to go through the same pain I did with running containers across local and CI. 🫠  It'll handle getting the ipv4 port so you can reliably connect to `127.0.0.1` (don't use localhost, you don't know if it'll resolve to a v4 or v6 address). It also handles setting the name and labels.

packages/server/src/sdk/app/oauth2/tests/utils.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+import TestConfiguration from "../../../../tests/utilities/TestConfiguration"
+import { generateToken } from "../utils"
+import path from "path"
+import { KEYCLOCK_IMAGE } from "../../../../integrations/tests/utils/images"
+
+const config = new TestConfiguration()
+
+const volumePath = path.resolve(__dirname, "docker-volume")
+
+jest.setTimeout(60000)
```
samwho, line 13: You might not need such a high timeout if you pre-pull the image.

packages/server/src/sdk/app/oauth2/tests/utils.spec.ts
adrinr: I implemented the changes, but this does not support killing containers and other... Not sure where the original issue came from, but we have some usages (like in packages/backend-core/src/redis/tests/redis.spec.ts) that has been running for ages without issues. Not sure if the issue was a one-off one

samwho (approved):

packages/server/src/integrations/rest.ts
```diff
@@ -377,28 +377,35 @@ export class RestIntegration implements IntegrationBase {
...
+              headers.Authorization = `Basic ${Buffer.from(
+                `${config.username}:${config.password}`
+              ).toString("base64")}`
+              break
+            case RestAuthType.BEARER:
+              headers.Authorization = `Bearer ${config.token}`
+              break
+            default:
+              throw utils.unreachable(type)
+          }
```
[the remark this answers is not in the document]
adrinr: Done!

PR #15736, opened 2025-03-13
Support attachment and signature columns for SQL databases

mike12345567:
## Description
Adds support for attachment columns in SQL databases, this turned out to be quite simple as a JSON column covers everything we need to support this and the row enrichment layers already do all of the work in input/output processing. The common row API absorbs 100% of the effort of implementing attachment columns, as long as it can store some JSON against the row.

Tests have been updated to include the SQL datasources when running the attachment and signature test cases.

Attachments in a Postgres table:
image


adrinr (approved):
LGTM!

PR #15742, opened 2025-03-13
New rest auth picker

adrinr:
## Description
Updating the rest authentication picker interface. This will allow integrating with the new oauth2 configs.

### Picker without configurations
image

### Picker with configurations but none selected
image

### Picker with selected configuration
image

### Select/unselect flow

https://github.com/user-attachments/assets/5f03c23a-5624-4d28-a39b-ae73eae63e39

### On "add new" click

https://github.com/user-attachments/assets/7bf47330-c453-4e39-8891-81ef00231061

## Launchcontrol
Updating rest auth picker


packages/bbui/src/List/ListItem.svelte
```diff
@@ -89,14 +89,13 @@
   /* Selection is only meant for standalone list items (non stacked) so we just set a fixed border radius */
   .list-item.selected {
     background-color: var(--spectrum-global-color-blue-100);
-    border-color: var(--spectrum-global-color-blue-100);
+    border: 1px solid var(--spectrum-global-color-blue-400);
```
aptkingston, line 92: I'd prefer to keep 100 for the border - 400 is quite strong.

aptkingston (commented):
Nice work!

The title in the modal seems to say "Basic" regardless of the type of auth - like in your screenshot you have a bearer token config which also is under the "Basic" title. I think we should just remove that title, or have something generic like "Pick an authentication strategy".

The button also is called "Add Basic" but you can add other types of auth configs can't you? So I'd maybe just update that to "Add auth config" or "Add auth strategy" (whatever terminology we use elsewhere).

aptkingston (approved):

PR #15744, opened 2025-03-14
Fix save button state on rest

adrinr:
## Description
The rest button has a save button, and this has a "disable" state. This button is currently not working as expected, as it only checks if the name has been set. This causes some issues as it's not clear what data has been saved and what is not.

This PR changes the way we handle state on the resteditor, enabling the save button when there is some change to be saved.

## Screenshots

https://github.com/user-attachments/assets/5428f5f8-56b2-4898-b1ee-b435b3e82290

## Launchcontrol
Improvements on the rest configuration editor

### DIFF
--- a/packages/builder/src/components/integration/RestQueryViewer.svelte
+++ b/packages/builder/src/components/integration/RestQueryViewer.svelte
@@ -504,7 +504,9 @@
             on:save={saveQuery}
           />
           <div class="controls">
-            <ConnectedQueryScreens sourceId={query._id} />
+            {#if query._id}
+              <ConnectedQueryScreens sourceId={query._id} />
+            {/if}
             <div class="access">
               <Label>Access</Label>
               <AccessLevelSelect {query} {saveId} />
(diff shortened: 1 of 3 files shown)


mike12345567 (approved):
LGTM!

packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -50,18 +50,20 @@
     toBindingsArray,
   } from "@/dataBinding"
   import ConnectedQueryScreens from "./ConnectedQueryScreens.svelte"
+  import _ from "lodash"
```
aptkingston, line 53: Not sure if the bundler would treeshake this or not, but we typically just import the function we need rather than the entirely of lodash. I doubt it actually makes a difference (and there's probably somewhere else in the codebase we import the whole thing like this), but we may as well just import what we need in case it does.

aptkingston (approved):

packages/builder/src/components/integration/RestQueryViewer.svelte
adrinr: It probably work fine, but it might be better to play safe. And I realised we were already importing it :P

PR #15745, opened 2025-03-14
Autosave rest authentication settings

adrinr:
## Description
Autosave rest authentication settings. The current flow requires an extra save button that is pretty confusing from the UX point of view.

## Screenshots

https://github.com/user-attachments/assets/10b24df7-1b09-4d50-8b43-7b703b7e3621

## Launchcontrol
Autosave rest authentication settings


mike12345567 (approved):

deanhannigan (approved):
LGTM! Confirmed the issue on master and verified the fix 👍

PR #15747, opened 2025-03-14
Typescript conversion of `CreateEditColumn.svelte`

mike12345567:
## Description
As title, this is a critical file in table column creation, I wanted to convert this as best as possible to TS. There were a few issues which I came across that this PR has not resolved yet - but was hoping to have a stab at dealing with those separately.

I've needed to create a new type `FieldSchemaConfig` - which is essentially a partial of all of our column options. This is an annoyingly complex type, but using union types really doesn't work with the current method of creating columns, the structure of a column is entirely fluid until it is saved at which point it makes sure it fits the `FieldSchema` type.

For now I've had to create something which covers all possible column configurations that can then be narrowed, but it is not ideal.

## Feature branch env
[Feature Branch Link](http://fb-fix-ts-convert-create-edit-column.fb.qa.budibase.net)


packages/bbui/src/Form/Multiselect.svelte
```diff
@@ -1,40 +1,41 @@
...
   export let readonly = false
   export let labelPosition = "above"
-  export let error = null
-  export let placeholder = null
-  export let options = []
-  export let getOptionLabel = option => option
-  export let getOptionValue = option => option
+  export let error: string | undefined = undefined
+  export let placeholder: string | undefined = undefined
+  export let options: any[] = []
```
adrinr, line 13: We should type these any from generics, so the options in function will be typed. 
```
<script lang="ts" generics="O">
export let options: O[] = []
export let getOptionLabel = (option: O) => option
```

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -1,17 +1,27 @@
+<script context="module" lang="ts">
+  type ValueType = string | string[]
```
adrinr, line 2: This should be done using generics

packages/shared-core/src/tests/filters.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import {
+  BasicOperator,
+  EmptyFilterOption,
+  FieldType,
+  UILogicalOperator,
+  UISearchFilter,
+} from "@budibase/types"
+
+describe("filter to query conversion", () => {
+  it("handles a filter with 1 group", () => {
```
adrinr, line 11: Thank you for doing this, I started it a couple of times in different PRs but never ended up merging it!

packages/types/src/api/web/searchFilter.ts
```diff
@@ -38,11 +38,19 @@ export type SearchFilter = {
 // involved. We convert this to a SearchFilters before use with the search SDK.
 export type LegacyFilter = AllOr | OnEmptyFilter | SearchFilter
 
+// A search filter group should either contain groups or filters, but not both
```
adrinr, line 41: Typing power! ♥️

adrinr (commented):
Code looks good except the generic approach as commented. Going to run some QA around it before giving the final ✅

adrinr (commented):
LGTM! Thanks for the generic linting fight :)

adrinr (approved):

PR #15748, opened 2025-03-14
Bind the password form

melohagan:
## Description
Bug found via PlayWright

## Addresses
- https://linear.app/budibase/issue/GRO-1031/fix-reset-password-bug


deanhannigan (approved):
LGTM!

Issue #15741, opened 2025-03-13
Cascading Dropdown Filters do not clear out old values


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.5.2

**Describe the bug**
Using an options picker and a filtered relationship picker, the filtered relationship picker holds on to old data.

[https://docs.budibase.com/docs/cascading-dropdown-filters](In this guide) we allow the user to select a country, and then show them cities based on the country they selected.

If the user selects a country, and then views the City picker, they will see cities in that country. When they select a different country, the City picker will show them cities in Country 1 and country 2. This behaviour can be repeated until the whole, unfiltered cities list is available in the City picker.

**To Reproduce**
_It's fairly detailed, but if you want to skip to the Jam or straight to the export that's fine too_

1. In the data section, create two tables with a one-to-many relationship (My example Countries and Cities works nicely for this) - you should populate these tables with familiar data
2. In the Design section, add a data provider fetching Countries (or the 1-side of your 1-to-Many) and use an Options picker to display the rows from the Data Provider. Set the Label and Value column to display `country_name`
3. Inside the data provider, also add a Form component containing a relationship picker.
4. On the Relationship picker, add the following filter - `country.country_name EQUALS {{ [New Options Picker].Value }}`
5. Select a country, and view the Cities available in the relationship picker - the cities should be filtered accordingly
6. Select a different country - the cities picker will contain cities from both the previous and current country.

**Expected behavior**
Old cities should be cleared out when the country value is changed.

**Screenshots**
https://jam.dev/c/594222a9-9f17-4e82-9ffd-324bfe0841f5
  
**App Export**
[filter forms-export-1741885360617.tar.gz](https://github.com/user-attachments/files/19233551/filter.forms-export-1741885360617.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS 14.6.1 (23G93)
 - Browser Chrome 
 - Version 134.0.6998.46 (Official Build) (arm64)

**Additional context**
CSE raised this bug on behalf of enterprise customer. This issue is not present in the example hosted in the documentation, but downloading the app-export provided there will demonstrate the issue.

Using On-Change actions on the Country picker to Clear Form or Update Field Value - reset to default does not help

Network traffic shows that the correct items are being returned, but the component itself seems to hold onto the previous data as well.

[...]

PR #15749, opened 2025-03-14
Filtering updates

aptkingston:
## Description
- Supports converting nested UI filters with groups into a query. The types allow this, but our code did not correctly convert such filter structures properly
- Added tests to verify filter > query conversion works for a few cases
- Tightened up types around filters to enforce that a single SearchFilterGroup may only contain filters or groups, but not both
- Updated relationship picker to clear options when a filter changes. This fixes an issue where the caching of options meant that options which no longer match the currently applied filter were still being shown
- Fixed an issue where the type of the filter prop in the relationship picker was wrong. A runtime error would be thrown if you attempted to search while also having a filter applied

## Addresses
- https://github.com/Budibase/budibase/issues/15741


mike12345567 (approved):
LGTM -thanks for the expanded tests as well!

PR #15750, opened 2025-03-14
Style tweaks

aptkingston:
## Description
- Fix height of text component when used inside flex container
- Use 600 weight for table headers and markdown generated headline elements
- Fix table header cell background color in data section and non-quiet tables in apps
- Use a theme-aware reactive color for app accent color again


joebudi (approved):
LGTM. Another improvement for core styles.

PR #15752, opened 2025-03-17
Modal to prevent losing REST datasource changes

adrinr:
## Description
REST datasources currently behave quite differently from the other datasources and the rest of the platform, as it is one of the few points where we don't autosave on actions: the user is required to save via the save click. This might be confusing as the UI and some of its elements might give the impression that the data is persisted on change. This will be even more problematic as we are changing some of its configuration elements from dropdowns to context buttons and menus.
This PR introduces a mechanism to detect if you are leaving a REST config with unsaved changes. If this is the case, the user will have 2 options:
1. Leave discarding the changes
2. Leave saving the changes
3. Close the modal and stay on the current page (this required some extra props on the confirm modal)

### No modal when no changes
no modal

### Modal appears when there are changes
modal on changes

### Saving on exiting the screen
save on quit

### Discard changes on exit
discarding

### Closing the modal will prevent from exiting
cancelling

### On new endpoint, we will validate if it's saved
new

### We will not validate new endpoint "save" status if there is no changes at all
empty new

## Launchcontrol
Confirm modal when leaving a rest configuration page with unsaved changes

### DIFF
--- a/packages/builder/src/components/integration/RestQueryViewer.svelte
+++ b/packages/builder/src/components/integration/RestQueryViewer.svelte
@@ -63,6 +63,7 @@
   let nestedSchemaFields = {}
   let saving
   let queryNameLabel
+  let mounted = false
 
   $: staticVariables = datasource?.config?.staticVariables || {}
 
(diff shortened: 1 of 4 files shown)


packages/builder/src/components/integration/RestQueryViewer.svelte
```diff
@@ -474,6 +480,38 @@
...
+
+    mounted = true
+  })
+
+  $beforeUrlChange(async () => {
+    if (!isModified) {
+      return true
+    }
+
+    return await confirm({
```
aptkingston, line 492: This is a nice util to have 👌

aptkingston (approved):
LGTM! The confirm util is very nice.

NAB - could we update the "Save and continue" button to be a CTA variant instead? The current red button makes it look like saving is a bad/dangerous thing, but it's probably what they want to press.

adrinr:
> LGTM! The confirm util is very nice.
> 
> NAB - could we update the "Save and continue" button to be a CTA variant instead? The current red button makes it look like saving is a bad/dangerous thing, but it's probably what they want to press.

Good catch! Updated:
image

PR #15755, opened 2025-03-17
Implement edit and delete config endpoints

adrinr:
## Description
Implement edit and delete config endpoints

### DIFF
--- a/packages/server/src/api/routes/tests/oauth2.spec.ts
+++ b/packages/server/src/api/routes/tests/oauth2.spec.ts
@@ -151,7 +151,9 @@ describe("/oauth2", () => {
         },
         {
           status: 400,
-          body: { message: "Name is not available." },
+          body: {
+            message: `OAuth2 config with name '${config2.name}' is already taken.`,
+          },
         }
       )
     })
--- a/packages/server/src/api/routes/tests/oauth2.spec.ts
+++ b/packages/server/src/api/routes/tests/oauth2.spec.ts
@@ -1,11 +1,11 @@
-import { CreateOAuth2ConfigRequest, VirtualDocumentType } from "@budibase/types"
+import { UpsertOAuth2ConfigRequest, VirtualDocumentType } from "@budibase/types"
 import * as setup from "./utilities"
 import { generator } from "@budibase/backend-core/tests"
 
 describe("/oauth2", () => {
   let config = setup.getConfig()
 
-  function makeOAuth2Config(): CreateOAuth2ConfigRequest {
+  function makeOAuth2Config(): UpsertOAuth2ConfigRequest {
     return {
       name: generator.guid(),
       url: generator.url(),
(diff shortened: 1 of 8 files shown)


packages/server/src/api/routes/tests/oauth2.spec.ts
```diff
@@ -92,4 +97,94 @@ describe("/oauth2", () => {
...
+        existingConfigs.push({ ...oauth2Config, id: result.config.id })
+      }
+    })
+
+    it("can delete an existing configuration", async () => {
+      const { id: configId } = _.sample(existingConfigs)!
+
+      await config.api.oauth2.delete(configId, { status: 204 })
+
+      const response = await config.api.oauth2.fetch()
```
samwho, line 178: Isn't this potentially flaky?

If the test below runs before this one, 5 configs get created and 0 deleted. Then this test runs, `existingConfigs` is reset to `[]` and 5 more configs get created. There are now 10 configs for the tenant. Then you delete 1 and compare the total to 4, but there are 9.

What am I missing?

packages/server/src/sdk/app/oauth2/crud.ts
```diff
@@ -48,3 +48,46 @@ export async function get(id: string): Promise<OAuth2Config | undefined> {
...
+  if (!doc.configs[config.id]) {
+    throw new HTTPError(`OAuth2 config with id '${config.id}' not found.`, 404)
+  }
+
+  if (
+    Object.values(doc.configs).find(
+      c => c.name === config.name && c.id !== config.id
+    )
+  ) {
+    throw new HTTPError("Name is not available.", 400)
```
samwho, line 68:
```suggestion
    throw new HTTPError(`OAuth config with name "${config.name}" not found`, 400)
```

packages/server/src/sdk/app/oauth2/crud.ts
```diff
@@ -48,3 +48,46 @@ export async function get(id: string): Promise<OAuth2Config | undefined> {
...
+  }
+
+  doc.configs[config.id] = {
+    ...config,
+  }
```
samwho, line 73: Do we need to copy `config` here?

packages/server/src/api/routes/tests/oauth2.spec.ts
adrinr: It should not. First, jest test run always in order per each file. And we are creating new tenants per each run, so each test is independent

packages/server/src/sdk/app/oauth2/crud.ts
adrinr: No, this is meant to be like this. It checks that the name is not already taken for another config, avoiding duplicates

packages/server/src/sdk/app/oauth2/crud.ts
adrinr: Not really in here, but I will need it in further PRs for things like this without mutating the passed object:
image

packages/server/src/sdk/app/oauth2/crud.ts
samwho: Sorry, not "not found" but "already taken." Essentially if "name is not available" were shown to the user I'm not sure they'd know exactly what went wrong. Being more explicit that the chosen name is already taken would be good.

samwho (approved):

PR #15759, opened 2025-03-18
PWA Support

PClmnt:
## Description
Adds the ability to enable PWA apps for your Budibase application on Enterprise tier. 

## Addresses
https://linear.app/budibase/issue/BUDI-9178/pwa-support

### Install button on desktop 
image

### Installing via button on phone

https://github.com/user-attachments/assets/0cb19c11-17b8-4183-b750-f095cbe2d4b3

## Launchcontrol
- Adds the ability to enable PWA support for Enterprise users

## Feature branch env
[Feature Branch Link](http://fb-feat-pwa-apps.budibase.dev)


packages/backend-core/src/objectStore/buckets/app.ts
```diff
@@ -51,3 +52,26 @@ export async function getAppFileUrl(s3Key: string) {
     return await objectStore.getPresignedUrl(env.APPS_BUCKET_NAME, s3Key)
   }
 }
+
+export async function enrichPWAImages(
+  images: PWAManifestImage[]
+): Promise<PWAManifestImage[]> {
+  if (!images || images.length === 0) {
```
adrinr, line 59: The parameter is not optional, so either this `!images` is not needed or we should make the param optional

packages/bbui/src/Form/File.svelte
```diff
@@ -12,7 +12,8 @@
   export let extensions: string[] | undefined = undefined
   export let error: string | undefined = undefined
   export let title: string | undefined = undefined
-  export let value: File | undefined = undefined
+  export let value: File | { name: string; type: string } | undefined =
```
adrinr, line 15: This does not match the core one, as it does not have `size`. Is that needed?

packages/builder/src/pages/builder/app/[application]/settings/pwa.svelte
```diff
@@ -0,0 +1,304 @@
...
+          .trim()
+      )
+    } catch (error) {
+      console.error("Error converting CSS variable:", error)
+      return "#FFFFFF"
+    }
+  }
+
+  async function handlePWAZip(file: File) {
+    if (!file) return
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/settings/pwa.svelte:73]: Should this display an error?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -243,8 +327,45 @@ export const serveApp = async function (ctx: UserCtx<void, ServeAppResponse>) {
 
       const { head, html, css } = AppComponent.render({ props })
       const appHbs = loadHandlebarsFile(appHbsPath)
+
+      let extraHead = ""
+      const pwaEnabled = true // await pro.features.isPWAEnabled()
```
adrinr, line 332: This needs to be properly set

packages/server/src/api/controllers/static/index.ts
```diff
@@ -397,3 +518,75 @@ export const getSignedUploadURL = async function (
...
+
+    if (appInfo.pwa.icons && appInfo.pwa.icons.length > 0) {
+      try {
+        manifest.icons = await objectStore.enrichPWAImages(appInfo.pwa.icons)
+
+        const desktopScreenshot = manifest.icons.find(
+          icon => icon.sizes === "1240x600" || icon.sizes === "2480x1200"
+        )
+        if (desktopScreenshot) {
+          manifest.screenshots?.push({
```
adrinr, line 560: This will not push the object if `screenshots` is undefined. Is this what you are attempting?
Otherwise, you would need to ensure that `manifest.screenshots` exists: `manifest.screenshots ??= []`

adrinr (commented):
Looks good, just a few comments here and there. I know this might be hard, but is there value adding some tests around it?

packages/server/src/api/controllers/static/index.ts
PClmnt: I can't add the license on the feature branch right now as account portal has stopped deploys, so this is just temporary.

packages/bbui/src/Form/Core/File.svelte
```diff
@@ -57,7 +58,16 @@
 <!-- svelte-ignore a11y-no-static-element-interactions -->
 <!-- svelte-ignore a11y-click-events-have-key-events -->
 <div class="field">
-  {#if value}
+  {#if statusText}
```
PClmnt, line 61: Needed a state for the File component where it didn't take a `File` type, just some text. And I thought this was cleaner than mixing types on the `value` prop.

aptkingston (commented):
Client changes LGTM!

packages/server/src/api/controllers/static/index.ts
PClmnt: Screenshots is always initialised as an empty array in the manifest definition on line 559.

packages/client/src/components/ClientApp.svelte
```diff
@@ -93,6 +94,16 @@
 
   // Load app config
   onMount(async () => {
+    if ("serviceWorker" in navigator) {
```
adrinr, line 97: Should this not be part of the `InstallPrompt` in order to encapsulate this logic?

packages/server/src/api/controllers/static/index.ts
```diff
@@ -397,3 +528,74 @@ export const getSignedUploadURL = async function (
 
   ctx.body = { signedUrl, publicUrl }
 }
+
+export async function servePwaManifest(ctx: UserCtx<void, any>) {
+  const appId = context.getAppId()
+  if (!appId) {
+    ctx.status = 404
```
adrinr, line 535: These 3 lines could be simplified as `ctx.throw(404, message)`. Same on the other 404

packages/server/src/api/controllers/static/index.ts
adrinr: I see, so probably the `?` is not required

adrinr (approved):
Some minor comments, but LGTM!

PR #15760, opened 2025-03-18
AI class refactor

samwho:
## Description

This change is mostly the set of companion changes required for the corresponding pro PR. The details of the changes are on that PR.

Pro PR: https://github.com/Budibase/budibase-pro/pull/364


packages/server/src/automations/steps/openai.ts
```diff
@@ -41,18 +41,9 @@ export async function run({
...
-    let llmWrapper
-    if (budibaseAIEnabled || customConfigsEnabled) {
-      llmWrapper = await pro.ai.LargeLanguageModel.forCurrentTenant(
-        inputs.model
-      )
-    }
-
-    response = llmWrapper?.llm
-      ? await llmWrapper.run(inputs.prompt)
+    const llm = await pro.ai.getLLM(inputs.model)
```
samwho, line 44: `getLLM` performs the pro features checks that were previously being performed here. `getLLM` is designed to return an LLM if the user has access to one, and nothing if they don't.

packages/server/src/utilities/rowProcessor/tests/utils.spec.ts
```diff
@@ -79,59 +60,4 @@ describe("rowProcessor utility", () => {
       expect(fixAutoColumnSubType(schema)).toEqual(schema)
     })
   })
-
-  describe("processAIColumns", () => {
```
samwho, line 83: What was being tested here is covered by more realistic tests of AI columns in `row.spec.ts.

packages/server/src/utilities/rowProcessor/utils.ts
```diff
@@ -126,10 +126,8 @@ export async function processAIColumns<T extends Row | Row[]>(
     const numRows = Array.isArray(inputRows) ? inputRows.length : 1
     span?.addTags({ table_id: table._id, numRows })
     const rows = Array.isArray(inputRows) ? inputRows : [inputRows]
-    const llmWrapper = await pro.ai.LargeLanguageModel.forCurrentTenant(
-      "gpt-4o-mini"
-    )
-    if (rows && llmWrapper.llm) {
+    const llm = await pro.ai.getLLM()
+    if (rows && llm) {
```
samwho, line 130: It's nicer here now that we don't have to reach into the returned class and check that a property of it is not null (`llmWrapper.llm` from old diff) before we can use it.

packages/types/src/documents/global/config.ts
```diff
@@ -111,18 +111,26 @@ export interface SCIMInnerConfig {
 
 export interface SCIMConfig extends Config<SCIMInnerConfig> {}
 
-export type AIProvider = "OpenAI" | "Anthropic" | "TogetherAI" | "Custom"
+export type AIProvider =
+  | "OpenAI"
+  | "Anthropic"
+  | "AzureOpenAI"
```
samwho, line 117: `"AzureOpenAI"` was missing from this list, it is referenced in several places outside this file.

packages/types/src/documents/global/config.ts
```diff
@@ -111,18 +111,26 @@ export interface SCIMInnerConfig {
...
+  | "TogetherAI"
+  | "Custom"
+
+export interface ProviderConfig {
+  provider: AIProvider
+  isDefault: boolean
+  isBudibaseAI?: boolean
```
samwho, line 124: This field is new, and much more explicitly labels a config as Budibase AI. We were previously inferring this from other fields. Budibase AI config is always enriched at runtime, so this won't be stored anywhere.

packages/types/src/sdk/licensing/license.ts
```diff
@@ -17,4 +17,5 @@ export interface License {
   plan: PurchasedPlan
   billing?: Billing
   testClockId?: string
+  tenantId?: string
```
samwho, line 20: https://github.com/Budibase/account-portal/pull/1165

packages/types/src/documents/global/config.ts
adrinr: Did the ts not complain about this missing?

adrinr (approved):

adrinr:
Thanks for the detailed comments!

packages/types/src/documents/global/config.ts
samwho: No, the typing in this area is a bit loose. This PR tightens it up a little, but there's likely still more we could do.

PR #15761, opened 2025-03-18
Enabling svelte generics support in eslint

mike12345567:
## Description
As title - enabling feature: https://github.com/sveltejs/svelte-eslint-parser/pull/477


adrinr (approved):
Good catch!

PR #15763, opened 2025-03-18
Edit/delete configs frontend implementation

adrinr:
## Description
Edit/delete configs frontend implementation

### Editing an existing config
https://github.com/user-attachments/assets/225f020a-ecc5-451e-89e1-5a3fc7e4686f

### Passwords are protected, they will be left as they are by default

https://github.com/user-attachments/assets/d5646337-559b-4e4d-8fb0-9cde39e72373

### Deletion confirmation
https://github.com/user-attachments/assets/d224ce75-8ee0-498a-8452-d78c70238bf5


packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte
```diff
@@ -1,11 +1,38 @@
...
+      title: "Confirm Deletion",
+      body: `Deleting "${row.name}" cannot be undone. Are you sure?`,
+      okText: "Delete Configuration",
+      warning: true,
+      onConfirm: async () => {
+        try {
+          await oauth2.delete(row.id)
+          notifications.success(`Config '${row.name}' deleted successfully`)
+        } catch (e: any) {
+          notifications.error("Error deleting config")
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte:32]: I think its always worth exposing the error in a scenario like this - its unlikely, but if it does happen it'll help diagnose what went wrong know what the message was.

packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte
```diff
@@ -0,0 +1,144 @@
...
+    try {
+      if (isCreation) {
+        await oauth2.create(validationResult.data)
+        notifications.success("Settings created.")
+      } else {
+        await oauth2.edit(config!.id, validationResult.data)
+        notifications.success("Settings saved.")
+      }
+    } catch (e: any) {
+      notifications.error(e.message)
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte:82]: Worth adding a user readable message to this as well, like `Failed to save config - ${e.message}`

packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte
```diff
@@ -0,0 +1,144 @@
...
+      full URL.
+    </Body>
+  </div>
+  <Input
+    label="Client ID*"
+    placeholder="Type here..."
+    bind:value={data.clientId}
+    error={errors.clientId}
+  />
+  <Input
```
Comment (mike12345567) [inline packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte:124]: One thing I believe we discussed was support for environment variable usage on the client id/secret, for production/development flagging and to make it easier to secure this information; has this been implemented yet?

mike12345567 (approved):
LGTM - just a couple of questions for my understanding!

packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte
Comment (adrinr) [reply packages/builder/src/pages/builder/app/[application]/settings/oauth2/OAuth2ConfigModalContent.svelte:124]: Not yet, it's in the todo list

PR #15769, opened 2025-03-19
Validate config

adrinr:
## Description
Validate configuration before saving

### Validate new config
https://github.com/user-attachments/assets/8cb52814-eb5b-46dc-b35c-ae2f0586cff2

### Validate existing config without modifying password

https://github.com/user-attachments/assets/5226c217-7e89-4eb8-9494-2f222b32028e

### Validate password changes

https://github.com/user-attachments/assets/0506b599-ac2a-4fa9-9528-372a66df93a5

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte
@@ -45,5 +45,5 @@
 </ActionMenu>
 
 <Modal bind:this={modal}>
-  <OAuth2ConfigModalContent config={row} />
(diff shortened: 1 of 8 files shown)


packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte
```diff
@@ -45,5 +45,5 @@
 </ActionMenu>
 
 <Modal bind:this={modal}>
-  <OAuth2ConfigModalContent config={row} />
+  <OAuth2ConfigModalContent config={{ ...row }} />
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/settings/oauth2/MoreMenuRenderer.svelte:48]: This will make a copy of the row, otherwise the in-memory one will be modified before being saved

mike12345567 (approved):
LGTM!

PR #15772, opened 2025-03-19
Track usage

adrinr:
## Description
Track oauth usages. We will update that usage every time that we get an oauth2 token.
The last usage will be picked from the most recent between dev and prod, as we want to track its usage, regarding if used in prod or dev

## Screenshots
image

### DIFF
--- a/packages/server/src/api/routes/oauth2.ts
+++ b/packages/server/src/api/routes/oauth2.ts
@@ -28,7 +28,7 @@ function oAuth2ConfigValidator() {
 function oAuth2ConfigValidationValidator() {
   return middleware.joiValidator.body(
     Joi.object({
-      id: Joi.string().required(),
+      id: Joi.string(),
       ...baseValidation,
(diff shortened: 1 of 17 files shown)


packages/server/src/api/routes/oauth2.ts
```diff
@@ -28,7 +28,7 @@ function oAuth2ConfigValidator() {
 function oAuth2ConfigValidationValidator() {
   return middleware.joiValidator.body(
     Joi.object({
-      id: Joi.string().required(),
+      id: Joi.string(),
```
mike12345567, line 31: Why is a separate `id` needed when the config has a `_id`? Is there multiple OAuth configs within the single oauth document?

adrinr: This might validate existing configs or new configs

mike12345567 (approved):
LGTM!

PR #15777, opened 2025-03-20
TS conversion for ColorPicker component

PClmnt:
## Description
Converts ColorPicker component to TS

### DIFF
--- a/packages/bbui/src/ColorPicker/ColorPicker.svelte
+++ b/packages/bbui/src/ColorPicker/ColorPicker.svelte
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
   import Popover from "../Popover/Popover.svelte"
   import Layout from "../Layout/Layout.svelte"
   import { createEventDispatcher } from "svelte"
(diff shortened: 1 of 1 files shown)


packages/bbui/src/ColorPicker/ColorPicker.svelte
```diff
@@ -124,20 +126,20 @@
...
+  const getThemeClasses = (theme: Theme | undefined) => {
     if (!theme) {
       return ""
     }
     theme = ensureValidTheme(theme, DefaultAppTheme)
     return getThemeClassNames(theme)
   }
 
-  const onChange = value => {
+  const onChange = (value: string | null) => {
```
mike12345567, line 137: I think we should be trying to get rid of `null` usage, as under typescript its very specific, `undefined` should work in all scenarios like this.

mike12345567 (approved):
LGTM!

packages/bbui/src/ColorPicker/ColorPicker.svelte
PClmnt: I had this set to auto merge, woops. I will fix this in my next PR, sorry!

PR #15779, opened 2025-03-20
Add accordion component

andz-bb:
## Description
Adds a new collapsible accordion component to Budibase. Child components show and hide when clicking the header.

## Screenshots
image

image

## Launchcontrol
Add accordion component to Budibase

## Feature branch env
[Feature Branch Link](http://fb-budi-9129-convert-accordion-plugin-to.budibase.dev)


packages/client/src/components/app/Accordion.svelte
```diff
@@ -0,0 +1,11 @@
+<script>
+  import { Accordion } from "@budibase/bbui"
+
+  export let bold
+  export let initialOpen
+  export let label
+</script>
+
+<Accordion header={label} {bold} {initialOpen}>
```
aptkingston, line 9: We should change this to `header={label || ""}` just to ensure it doesn't say `undefined` when there's no text.

aptkingston (commented):
Nice work Andy! Works perfectly 👌 

Just a few small finishing touches I think we could put on it. Looking at the default state when adding a new instance:
image

I think we could:
- Give it a default title (e.g. just "Accordion")
- Make sure it doesn't say `undefined` when there's no text (I've commented in the code how to do this)
- Increase the size a bit when added to grid layouts (I would go for the same as a container, so width 500 and height 200 in the `size` key in the manifest)
- Since it's basically a container, I think you always want to grid sizing to be "stretch", which also cleans up the settings bar. So I think we should set both `hAlign` and `vAlign` to `stretch` in the manifest

aptkingston (approved):

PR #15781, opened 2025-03-20
Automation side bar

deanhannigan:
## Description
A general overhaul of the automations workflow UI in the builder. 

- Step configuration has been moved out of the steps and into a new side panel. This should make configuration much easier and provide more room for the UI to grow.
- Automation test results are now overlayed directly onto the automation itself. Previously this would be listed in a sidebar.
  - The new approach should give a much more explicit picture of the execution path especially for complex, branched automations.
  - New labels are displayed above any steps that were executed in the test run; `Completed`, `Failed` or `Stopped` for conditional workflow states. Clicking `Completed` will open the `Data out` tab and clicking `Failed` or `Stopped` will take you to the `Errors` tab.
- New tabbed `Data` section under the step config.
  - `Data in` - show the data context available to each step on the automaton tree. Padded out initially with static content and stub data and then populated with real data upon triggering a test run.
  - `Data out` - displays the step results after a successful test run.
  - `Errors` - list any errors with the automation. Displayed in the following order; `Error`, `Warning`, `Default`
- A new `Logs` link has been added to the top of the UI that opens a new tab to the production logs for the current automation. 
  - If the published version of the automation is experiencing errors, an error count will be displayed to draw the eye.
- Automations Typescript conversion has been expanded in the frontend and backend.
- General refactoring of the Automations code to break it apart.

## Addresses
- https://linear.app/budibase/issue/BUDI-8992/step-configuration-sidebar

## Screenshots
Updated layout. New step config panel and slimmed down automation.
Screenshot 2025-03-20 at 15 31 54

Test results - the path taken during an automation test run is now overlayed on the automation itself.
Screenshot 2025-03-20 at 22 41 07

`Data in` tab - before running a test. Static data and step metadata detailing the `type` and `description` for the expected input. This is an example of a `Javascript` step available somewhere in the automation flow.
Screenshot 2025-03-20 at 22 35 47

`Data in` tab - *after* running the test. Now populated with real test data for the `Javascript` step.
Screenshot 2025-03-21 at 09 11 21

`Data out` tab - displaying any data output by the selected step during the test. In this example, the `rows` output from a `Query rows` step.
Screenshot 2025-03-21 at 09 09 14

`Errors` tab - displays the current error count and lists any errors
Screenshot 2025-03-20 at 22 44 04

`Logs` link - opens a new tab to the production logs for the current automation for debugging.
Screenshot 2025-03-20 at 22 44 19

Clicking the log link will take you to the Automation logs in your app settings. If there are errors, it will filter by the target automation and preselect any error logs.
Screenshot 2025-03-21 at 11 55 30

## Launchcontrol
Automations redesign. Step configuration moved to a side bar with evaluation context explorer.


aptkingston:
image
image

aptkingston (commented):
Absolutely massive piece of work Dean, well done. This adds so much insight into whats happening in automations and will be a life saver for anyone trying to build or debug any non-trivial automation.

This works flawlessly - the only things I could find are small UX things:
- The divider between step stetings and data in/out seems heavily in favour of data in/out by default. I would give a lot more space to the step settings and less to data in/out by default. It also seems to have a very large minimum size, which prevents me from shrinking data in/out as much as I want to.
- The divider also seems to sometimes move when clicking on different steps (particularly the trigger in my case) which is just a bit annoying
- Clicking on the copy icon in the data in/out sections copies a full JS object rather than just the value. We just need to update the code to copy the `value` property of the callback payload
- Not a big deal at all, but some errors aren't properly displayed. For example, add an update row step and don't add any settings, then run a test. It displays `object Object` for the message, but in data out you can see the real error. Not sure if we should try and exrtact an inner `message` property from error objects?
image
- NAB: we don't show an icon when there's no error, so I think I would remove this floating icon too - I don't think it adds anything and looks a little out of place
image

Tried my best to break it but I couldn't!

deanhannigan:
Thanks again for looking through all this @aptkingston. It's very much appreciated 🙏 

I've run through pretty much all the UX updates you requested except for one:

> * The divider also seems to sometimes move when clicking on different steps (particularly the trigger in my case) which is just a bit annoying

The jump only occurs when switching from a trigger to step and vice-versa. As triggers don't have any action buttons in the header bar, the config panel and data panel shift up.

aptkingston:
> Thanks again for looking through all this @aptkingston. It's very much appreciated 🙏
> 
> I've run through pretty much all the UX updates you requested except for one:
> 
> > * The divider also seems to sometimes move when clicking on different steps (particularly the trigger in my case) which is just a bit annoying
> 
> The jump only occurs when switching from a trigger to step and vice-versa. As triggers don't have any action buttons in the header bar, the config panel and data panel shift up.

Gotcha, didn't even notice it was the buttons causing the jump!

aptkingston (approved):

PR #15782, opened 2025-03-20
Split oauthconfig per document

adrinr:
## Description
As discussed on the 1:1, splitting the oauthconfig to each own document instead of having an aggregate


mike12345567 (approved):
LGTM!

Issue #15768, opened 2025-03-19
Cannot add bindings to Queries on external DBs


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: CLI/Dev stack
  - Budibase Version: <version> 3.5.3
  - App Version: 3.5.0

**Describe the bug**
Cannot add binding to Query for external DB (MS SQL, Postgres)

**To Reproduce**
1. connect an external data source
2. add query
3. add binding

**Expected behavior**
Should be able to add a binding and define it's default value

**Screenshots**
https://jam.dev/c/f918f414-d1e5-4aba-a2d4-e788dd08583b

**Desktop (please complete the following information):**
 - OS: MacOS 14.6.1
 - Browser Chrome
 - Version 134.0.6998.46 (Official Build) (arm64)

**Additional context**
**Issue observed on self-host 3.5.3 but not replicated on Cloud**

mike12345567:
Fixed in next release.

PR #15783, opened 2025-03-20
Fixing an issue where query bindings could not be added

mike12345567:
## Description
As title - can add bindings again.

Addresses: https://github.com/Budibase/budibase/issues/15768

image


adrinr (approved):
LGTM

PR #15784, opened 2025-03-20
SMTP - Cloud send email issue

mike12345567:
Fixing an issue with SMTP config, in the Cloud it is possible to configure send email without having added SMTP configuration, this will always fail.

This is supposed to appear as: 
image

But in the cloud it's always enabled.

## Addresses
- https://linear.app/budibase/issue/BUDI-9094/cloud-allows-adding-send-email-step-to-automation-without-smtp


adrinr (commented):
LGTM

adrinr (approved):

PR #15788, opened 2025-03-21
PDF screen template

aptkingston:
## Description
This PR adds a new PDF screen template. This is an opinionated screen template which is always sized as a portrait A4 page, and includes a title and a button to download the PDF.

New screen template:
image

Default state of new PDF screen:
image

Some notes about PDF screens:
- You can't preview them on different devices - it's desktop only
- The pages themselves are always light themed
- PDF screens are always grid containers

## Page breaks
As you add content to your page, it will continue to expand downards. Once you exceed the size of a single page, a new page will be added automatically and a line indicating a page break will be added. You do not need to leave space either side of the page break - it will be inserted automatically when you download your PDF - so you can have content either side of the page break.

image

Includes the approved PRs:
- https://github.com/Budibase/budibase/pull/15838
- https://github.com/Budibase/budibase/pull/15844

## Feature branch env
[Feature Branch Link](http://fb-pdf-screen-template.budibase.dev)


deanhannigan (commented):
Really solid work! Paging and display of table data working perfectly across all datasource types!

One bug: 
If you add any component to the PDF page, it will initially not display at full width when saved. If you resize the component and save the PDF again, the dimension will be correct.

One minor observation:
Buttons aren't currently respecting the screen accent config. Most components seem to use the accents without issue.

aptkingston:
> Really solid work! Paging and display of table data working perfectly across all datasource types!
> 
> One bug: If you add any component to the PDF page, it will initially not display at full width when saved. If you resize the component and save the PDF again, the dimension will be correct.
> 
> One minor observation: Buttons aren't currently respecting the screen accent config. Most components seem to use the accents without issue.

Cheers Dean. Both should be fixed 👌

deanhannigan (approved):
Awesome! Ran through everything and all issues are resolved 🎉

PR #15792, opened 2025-03-24
Add new pill to auth picker

adrinr:
## Description
Following designs, displaying the new pill on the auth picker. This will only be visible if there is no auth config selected.

### None selected, closed picker
image
### None selected, open picker
image
### Selected auth
image


deanhannigan (approved):
LGTM!

Issue #15778, opened 2025-03-20
Different result between Budibase and MySQL CLI


Philippe-M:
**Checklist**
- [ *] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 3.5.3
  - App Version: 3.5.3

**Describe the bug**
The result of a query is different when executed from Budibase compared to when it is executed directly in the MySQL console.

**To Reproduce**
1. Go to query in datasource (mysql) 
2. Add this dump : [MySQL-dump.zip](https://github.com/user-attachments/files/19362612/MySQL-dump.zip)
3. Create a new query : checkSscc
5. `SELECT count(ssccNum) AS nbsscc FROM sdelivery WHERE ssccNum = "{{ ssccnum }}"`
6. add binding parameter : ssccnum
7. save
8. in new screen add form with et text field name : ssccnum. And add text field name result
9. add button and add 3 action on click : 
10. reset to default value result field
11. execute query checkSscc with parameter : {{ ssccnum.Value }}
12. add update value to result field

![Image](https://github.com/user-attachments/assets/f8c0965c-d1cf-4a81-913b-cecca235dc05)

![Image](https://github.com/user-attachments/assets/c3563758-c6e7-47ca-8cf5-866168ba71e4)

![Image](https://github.com/user-attachments/assets/c0f62ce9-6878-46eb-aab0-16c6b8492104)

14. In sdelivery table i've this data

![Image](https://github.com/user-attachments/assets/1733f34f-6679-423b-975a-69579939f36b)

15. Publish your app and test

In mysql console sdelivery table : 

![Image](https://github.com/user-attachments/assets/861a78cb-46b6-4383-8e74-a945d7092642)

**My test**

![Image](https://github.com/user-attachments/assets/4727bba0-700f-4a60-98a5-5ffd0e1a1d58)

But with mysql cli

![Image](https://github.com/user-attachments/assets/29808c46-52a3-4355-9fd5-306615e3fd66)

If i'm testing with "small" value 

![Image](https://github.com/user-attachments/assets/00225652-7ea4-4634-9b54-d911b5c265ad)

![Image](https://github.com/user-attachments/assets/baf0e9ce-d8b3-4a76-ab5f-6d09d82e658d)

**Desktop (please complete the following information):**
 - OS: Linux Fedora 41
 - Browser Firefox
 - Version 136.01

mike12345567:
[...]

PR #15795, opened 2025-03-24
MySQL ANSI quotes support for queries

mike12345567:
## Description
In MySQL by default constant character strings can be input as `"String"` or `'String'` - this was causing issues in our binding system.

This fix addresses the issue defined here: https://github.com/Budibase/budibase/issues/15778

If `ANSI_QUOTES` are disabled then in theory we will try to replace bindings within identifier strings but bindings in these situations do not work anyway so either case will be an error.

### DIFF
--- a/packages/server/src/integrations/queries/sql.ts
+++ b/packages/server/src/integrations/queries/sql.ts
@@ -1,10 +1,30 @@
 import { findHBSBlocks } from "@budibase/string-templates"
-import { DatasourcePlus } from "@budibase/types"
+import { DatasourcePlus, SourceName } from "@budibase/types"
 import sdk from "../../sdk"
 
-const CONST_CHAR_REGEX = new RegExp("'[^']*'", "g")
+function getConstCharRegex(sourceName: SourceName) {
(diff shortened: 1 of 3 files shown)


packages/server/src/integrations/queries/sql.ts
```diff
@@ -1,10 +1,30 @@
...
-import { DatasourcePlus } from "@budibase/types"
+import { DatasourcePlus, SourceName } from "@budibase/types"
 import sdk from "../../sdk"
 
-const CONST_CHAR_REGEX = new RegExp("'[^']*'", "g")
+function getConstCharRegex(sourceName: SourceName) {
+  // MySQL clients support ANSI_QUOTES mode off, this is by default
+  // but " and ' count as string literals
+  if (sourceName === SourceName.MYSQL) {
+    return new RegExp(`"[^"]*"|'[^']*'`, "g")
```
samwho, line 9: Could you define these regexes as consts at the global level? Constructing them has a non-zero cost.

samwho (approved):

PR #15798, opened 2025-03-24
[Fixed] Wrong user/creator calculation on app init

jvcalderon:
## Description

Creators was being set to 0 on application init. Account holder was being considered as user.

Now if `isAccountHolder` flag is given, it is considered as a creator addition. This change is needed because it was causing problems when the first user (the holder) was being created.

## Addresses

https://linear.app/budibase/issue/GRO-1050/wrong-usercreator-calculation-on-app-init

### DIFF
--- a/packages/backend-core/src/users/db.ts
+++ b/packages/backend-core/src/users/db.ts
@@ -261,9 +261,13 @@ export class UserDB {
       }
     }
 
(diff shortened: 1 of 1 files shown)


shogunpurple (approved):
LGTM

samwho:
Could we add a test to prevent this regressing in future?

PR #15805, opened 2025-03-25
Cache token on oauth2 usage

adrinr:
## Description
Cache token on oauth2 usage to speed up requests and to avoid issues with 429 (too many requests). This is recommended in the [RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1).
This PR also handles retries: if using a token in a rest request we get a 401, we will clean the cached token and retry ONE extra time. This will avoid issues with revoked tokens or other conflicting issues. Otherwise, all the requests will fail until the TTL expires

### DIFF
--- a/packages/server/src/sdk/app/oauth2/utils.ts
+++ b/packages/server/src/sdk/app/oauth2/utils.ts
@@ -6,7 +6,7 @@ import {
   OAuth2CredentialsMethod,
   OAuth2GrantType,
 } from "@budibase/types"
-import { cache, context, docIds, Duration } from "@budibase/backend-core"
+import { cache, context, docIds } from "@budibase/backend-core"
 
 interface OAuth2LogDocument extends Document {
   lastUsage: number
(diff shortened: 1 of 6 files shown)


packages/server/src/sdk/app/oauth2/utils.ts
```diff
@@ -62,24 +62,32 @@ const trackUsage = async (id: string) => {
...
+    async () => {
+      const config = await get(id)
+      if (!config) {
+        throw new HttpError(`oAuth config ${id} count not be found`)
+      }
+
+      const resp = await fetchToken(config)
+
+      const jsonResponse = await resp.json()
+      if (!resp.ok) {
```
samwho, line 77: I think you probably want to check `resp.ok` _before_ grabbing `resp.json()`, in case it fails in such a way that it doesn't return JSON. Then when formatting the error, call `resp.text()` which is less likely to throw.

samwho (approved):

packages/server/src/sdk/app/oauth2/utils.ts
adrinr: This should cover controlled errors. `401` or `403` and `error_description` is part of the oauth2 spec, so we want to extract it if provided. Otherwise, the `resp.statusText` should get other issues, such as connection ones

Issue #15791, opened 2025-03-24
Row selection unavailable since version 3.6.0


Colinka38:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: modra.budibase.app -->

**Describe the bug**
I have just upgraded to version 3.6.0 and it seems that it is no more possible to select the rows in a table (internal table).
Is it expected behaviour?

**Expected behavior**
Before we had the option to activate: Allow row selection on a table component.

**Screenshots**

![Image](https://github.com/user-attachments/assets/1bb88e8c-4255-441d-ae66-886da7bb7556)

**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [chrome]
 - Version []

**Additional context**

filippobenozzi:
Easy and temporary fix.
Add an embed component in the page where the table is and add this code inside:
```
<style>
.grid .checkbox{
  display: block !important;
}
</style>
```

Colinka38:
Oh great @filippobenozzi thank you for the workaround. It's working well. I also found that when I enable "delete rows", the checkboxes are back.

aptkingston:
Hey. We deliberately removed the checkboxes if you have the table configured to prevent adding or removing rows, to keep it cleaner. We assume that if you can't add or edit rows then the checkboxes serve no purpose, but I'm now realising that you could potentially use the "selected rows" binding even with a readonly table.

Is it the selected rows binding that you're using here? Or what else do you use the checkboxes for if you've disable adding and removing rows?

I can easily add this back in if it's been an oversight on our part.

LukiWasTaken:
Hi @aptkingston,

not the one who opened this issue but wanted to add that the "Export Data" action also requires the option to select rows. I added the option to add rows for now so the checkbox becomes visible again, but that is not a function I want to have for that table.

![Image](https://github.com/user-attachments/assets/75feb5b1-11db-427b-a486-9077f02cc2bc)

![Image](https://github.com/user-attachments/assets/e32ef80c-6e0d-41c0-8a77-0311d0b2fcd8)

Colinka38:
@aptkingston 
Hello and thank you for looking at the problem. Actually yes it was for selected rows usage, as well as the 'Export data'. It could be good to have the checkboxes back again :)
Have a good day
Katarina

[...]

Issue #15764, opened 2025-03-18
Lightning bolt incorrectly positioned New Column popover


mikesealey:
**Checklist**
- [X] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: docker compose
  - Budibase Version: 3.5.3

**Describe the bug**
Lightning bolt for bindings drawer isn't contained inside the field.

**To Reproduce**
1. Add a new column

**Expected behaviour**

![Image](https://github.com/user-attachments/assets/b44141a4-7f17-426a-9bf8-0836e1cda14c)

**Screenshots**

![Image](https://github.com/user-attachments/assets/c4c6496a-348d-4681-a9cf-6bd842db2bcc)

**Desktop (please complete the following information):**
 - OS: MacOS Sonoma 14.6.1 
 - Browser Chrome
 - Version 134.0.6998.46 (Official Build) (arm64)

aptkingston:
Thanks for noticing. Will have a fix up soon.

Issue #15756, opened 2025-03-17
Data import will not accept properly formatted dates


heyjoeway:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: nui34t9grn89vy

**Describe the bug**
When importing data to a new table and setting a field format to "Date", I receive an error that the formatting isn't correct even though it is:

`Invalid format for field "Test": "2024-10-31T00:00:00Z". Datetime fields must be in ISO format, e.g. "YYYY-MM-DDTHH:MM:SSZ".`

**To Reproduce**
Import a CSV to a new table with the following content:
```csv
Test,Test2
2024-10-31T00:00:00Z,1234
```
OR a JSON with the following content:
```json
[
	{
		"Test": "2024-10-31T00:00:00Z",
		"Test2": "1234"
	}
]
```
Make sure that the "Test" field type is set to "Date".

**Expected behavior**
The data should import without error.

**Screenshots**
![Image](https://github.com/user-attachments/assets/02a8dd00-01b6-4911-aa94-8b584e1b1049)
![Image](https://github.com/user-attachments/assets/be072efb-02de-426a-bf54-3d784daff4e7)
  
**App Export**
This is on a brand new app.

**Desktop (please complete the following information):**
 - OS: Windows 11 24H2 (26100.3476)
 - Browser: Chrome
 - Version: 134.0.6998.89

**Additional context**
N/A

heyjoeway:
It seems like the message is misleading. Exporting an existing date from Budibase and reimporting does work, but the format is slightly different from what is indicated:
- For dates, you can use ISO 2014 and simply leave out the time portion, eg. `2024-10-31`
- For times, you need to explicitly specify the decimal portion, eg. `2024-10.31T00:00:00.000Z`

PR #15814, opened 2025-03-26
Misc fixes

aptkingston:
## Description
Fixes:
- Ensures binding icon is in the correct position for both inputs and text areas, regardless of other settings
- Restores row selection on tables regardless of settings
- Allow text component to accept any data types for the text setting, and try its best to stringify it
- Fix bug with date validation when importing data into tables (millis are now optional)
- Add proper validation of dates to the pre-import checks, to ensure that invalid dates will show before you try to import the data (previously we should show this as success, but then fail when it comes to the real import, bow it uses the same logic to ensure the result is the same)
- Coerce string values into arrays to ensure they are valid values when importing strings into array fields (e.g. you can import "foo" into an array field, and it will save it as ["foo"]). This used to work, but broke at some point in time.
- Add tests for the above

## Addresses
- https://github.com/Budibase/budibase/issues/15764
- https://github.com/Budibase/budibase/issues/15791
- https://github.com/Budibase/budibase/issues/15823
- https://github.com/Budibase/budibase/issues/15756

### DIFF
--- a/packages/backend-core/src/sql/utils.ts
+++ b/packages/backend-core/src/sql/utils.ts
@@ -139,17 +139,17 @@ export function breakRowIdField(_id: string | { _id: string }): any[] {
   }
 }
 
-export function isInvalidISODateString(str: string) {
+export function isValidISODateString(str: string) {
   const trimmedValue = str.trim()
   if (!ISO_DATE_REGEX.test(trimmedValue)) {
-    return true
+    return false
   }
-  let d = new Date(trimmedValue)
-  return isNaN(d.getTime())
+  const d = new Date(trimmedValue)
+  return !isNaN(d.getTime())
 }
 
-export function isValidISODateString(str: string) {
-  return ISO_DATE_REGEX.test(str.trim())
+export function isInvalidISODateString(str: string) {
+  return !isValidISODateString(str)
 }
 
 export function isValidFilter(value: any) {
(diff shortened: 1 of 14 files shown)


deanhannigan (approved):
LGTM! All fixes verified against prod! 🚀

packages/backend-core/src/sql/utils.ts
```diff
@@ -142,7 +142,7 @@ export function breakRowIdField(_id: string | { _id: string }): any[] {
 export function isInvalidISODateString(str: string) {
   const trimmedValue = str.trim()
   if (!ISO_DATE_REGEX.test(trimmedValue)) {
-    return false
+    return true
```
aptkingston, line 145: This was a logic bug that could have been causing us date problems, as it is incorrectly saying invalid dates are actually valid.

adrinr: I think this happened because usually this should be `isValidISODateString`, instead of `isInvalid` 😅  I can see why this was confusing

samwho: I wonder if it makes more sense to have this be:

```typescript
export function isInvalidISODateString(str: string) {
  return !isValidISODateString(str)
}
```

adrinr (approved):

packages/backend-core/src/sql/utils.ts
aptkingston: Yea I was thinking the same. The invalid check is more thorough and actually tries constructing a date after the regex check. I'll pull the more thorough check into the `isValid`, then update isInvalid` to just invert `isValid`.

PR #15829, opened 2025-03-27
Bull queue observability.

samwho:
## Description

This PR adds APM and metrics for our bull queues. It does this by wrapping the `BullQueue` type in a new `BudibaseQueue`, which instruments the adding of items to the queue and the processing callback.

The result of this is that we now have:

- `queue.add` and `queue.process` APM spans, where `queue.process` links back to the `queue.add` that originally added it.
- A variety of metrics tracking the add rate and job size for each queue.

I've tested this by running `datadog-agent` locally and seeing things happening in our dashboards.

Pro PR: https://github.com/Budibase/budibase-pro/pull/371

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -56,6 +56,7 @@
     "koa-pino-logger": "4.0.0",
     "lodash": "4.17.21",
     "node-fetch": "2.6.7",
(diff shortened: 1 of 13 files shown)


mike12345567 (approved):
LGTM - very nice work!

PR #15833, opened 2025-03-28
Load CSS/image assets without going through middlewares

mike12345567:
## Description
Small improvement to remove middleware and licensing from loading static assets, these are public anyway.


adrinr (approved):
LGTM!

PR #15838, opened 2025-03-28
PDF table component

aptkingston:
## Description
This PR adds a new "Plain Table" component (naming TBD), which is primarily designed to display tabular data in a printable way for the new PDF screen.

image

Some features of this component:
- It never scrolls. Content will be squished in horizontally as required, and cutoff at the side if you keep adding too many columns.
- All content is rendered as plain text. Every data type is represented in text form, ensuring it's fully readable, printable and searchable.

This table has most of the usual settings for fetching data - datasource, sorting, limit and filtering. It also contains a setting to control which columns are included in the table, and what to label them as.

Settings:
image

Column settings:
image

As this table is typically space constrained, it defaults to only showing the first 3 columns of the selected datasource. This ensures it doesn't look horribly cramped out of the box.

## Addresses
- https://linear.app/budibase/issue/BUDI-9157/create-pdf-compatible-table-component

## Other stuff
Also added a new size setting for the text component.


deanhannigan (changes requested):
Nice work! I ran through it locally and the table component is really solid! Also, the size config updates to the `Text` component are working great 🚀 

Some minor observations and one theming bug.

- Columns UX for JSON/CSV
  - Theres currently no way to alter or customize the columns used. Limited to 3 columns as expected, but then column UI is unusable.
  - I realise that column config is not generally supported in these scenarios, so it should probably not be displayed.
- Filters not clearing when the datasource is changed. Just needs configured in the manifest
- Light default font colour on all themes except light.
Screenshot 2025-04-01 at 10 09 10

deanhannigan (approved):
LGTM! 🚀

PR #15844, opened 2025-03-31
Single row provider component

aptkingston:
## Description
This PR adds a "Single Row Provider" component. It's just a faster way of fetching data for a single row without having to use a data provider and repeater.

This shows up as the first option now under the "Data" category:
image

It has 2 simple settings - a datasource and a row ID:
image

It provides context of a single row, with no need to access the 0th element of an array or anything:
image

Also added a user readable error message when trying to fetch a single row, as it was previously not handled well at all.
- Sending up a datasource ID and an invalid row ID threw a 400 with `Couch error: missing`
- Sending up a datasource ID and a valid row ID for a different source was completely unhandled a threw a 500 internal server error

Also fixed a bug that was causing the component order to be wrong; the first and second component in each category were getting swapped.

## Addresses
- https://linear.app/budibase/issue/BUDI-9158/create-single-row-provider-block


packages/server/src/api/controllers/row/index.ts
```diff
@@ -144,9 +144,11 @@ export async function find(ctx: UserCtx<void, FindRowResponse>) {
   const { tableId, viewId } = utils.getSourceId(ctx)
   const sourceId = viewId || tableId
   const rowId = ctx.params.rowId
-
-  const response = await sdk.rows.find(sourceId, rowId)
-  ctx.body = response
+  try {
+    ctx.body = await sdk.rows.find(sourceId, rowId)
+  } catch (e) {
+    ctx.throw(400, "That row couldn't be found")
```
samwho, line 150: 404?

aptkingston: Sure thing, that does make sense for a dedicated "fetch one row" endpoint.

samwho (approved):
Approving for the backend portion of the PR 🙂

deanhannigan (approved):
LGTM! Everything worked great in the builder with just two observation:

-  Not configured for the`Refresh Datasource` behaviour in button actions.
-  The row _id setting should have a `resetsOn` value set to the datasource prop.
    - It currently doesn't clear when the source is changed, throwing an error until its removed

aptkingston:
> LGTM! Everything worked great in the builder with just two observation:
> 
> * Not configured for the`Refresh Datasource` behaviour in button actions.
> * The row _id setting should have a `resetsOn` value set to the datasource prop.
>   
>   * It currently doesn't clear when the source is changed, throwing an error until its removed

Yea I deliberately didn't hook it up to refresh datasource, since I didn't think it was needed... In general manual refreshing shouldn't be (since it should automatically refresh), and specifically for only fetching a single row I thought that reduced the importance again. I suppose it's pretty trivial to support it, so I'll add it in to avoid someone asking for it 6 months down the line.

And yea good call on resetting the row ID when changing datasource. That was on my todo list and I obviously forgot to add it!

adrinr (approved):
So simple and so needed ❤️

PR #15852, opened 2025-03-31
Allow numeric static formulas to be used in view calculations

mike12345567:
## Description
Adding support for static formulas with a numeric response type when creating view calculations, this only applies to the internal DB but offers a way to transform say string columns to numbers and then calculate on these.

A static formula on a string column with a numeric response type (this string column only contains numbers)
image

Configuring view:
image

Displaying the max: 
image

### DIFF
--- a/packages/server/src/api/routes/tests/viewV2.spec.ts
+++ b/packages/server/src/api/routes/tests/viewV2.spec.ts
@@ -38,10 +38,7 @@ import {
   FormulaType,
 } from "@budibase/types"
 import { generator, mocks } from "@budibase/backend-core/tests"
-import {
-  datasourceDescribe,
-  DatabaseName,
-} from "../../../integrations/tests/utils"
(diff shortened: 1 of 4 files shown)


adrinr (approved):
Nice one :)

packages/server/src/api/routes/tests/viewV2.spec.ts
```diff
@@ -3865,6 +3866,45 @@ if (descriptions.length) {
...
+                        formulaType: FormulaType.STATIC,
+                        responseType: FieldType.NUMBER,
+                        formula: "{{ string }}",
+                      },
+                    },
+                  })
+                )
+                await config.api.row.save(table._id!, {
+                  string: "1",
+                })
```
samwho, line 3890: Could you create more than 1 row for this test, just to make sure it is actually maxing something and not passing by some fluke?

samwho (changes requested):

samwho (approved):

PR #15855, opened 2025-04-01
Format the plan type label

melohagan:
## Description
Noticed the "Premium" plan was displaying as "premium_plus". Using the label helper to present this nicely. Also added bold to make it more obvious.

## Addresses
- https://linear.app/budibase/issue/GRO-1075/replace-premium-plus-with-premium-in-the-budibase-platform-ui

## Screenshots
plan


aptkingston (approved):
The layout of the tags is pretty nasty here (prettier looks like it really messed it up) but obviously that doesn't affect functionality at all, so LGTM.

PR #15856, opened 2025-04-01
Fix size of boilerplate screen components on mobile

aptkingston:
## Description
Fixes a small issue where the title and button of the boilerplate screen did not have mobile sizing set, meaning they were a bit cramped by default.

Before:
image

After:
image


adrinr (approved):
🎉

PR #15857, opened 2025-04-01
Create new global DB for self-host quota.

samwho:
## Description

This PR tracks quota differently for self-host users making use of Budicloud services. Instead of creating a global DB per self-host tenant, we create a single global DB for all self-host tenants and then create tenant-specific `quota_usage_*` documents.

The idea behind this is that we want to maintain the idea that a single global DB represents a single cloud tenant. We don't want to muddy that water with self-host users.

Pro PR: https://github.com/Budibase/budibase-pro/pull/374

### DIFF
--- a/packages/backend-core/src/context/mainContext.ts
+++ b/packages/backend-core/src/context/mainContext.ts
@@ -157,6 +157,12 @@ export async function doInTenant<T>(
   return newContext(updates, task)
 }
 
+// We allow self-host licensed users to make use of some Budicloud services
+// (e.g. Budibase AI). When they do this, they use their license key as an API
+// key. We use that license key to identify the tenant ID, and we set the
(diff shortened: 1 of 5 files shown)


packages/backend-core/src/context/mainContext.ts
```diff
@@ -157,6 +157,27 @@ export async function doInTenant<T>(
   return newContext(updates, task)
 }
 
+export async function doInSelfHostTenantUsingCloud<T>(
```
mike12345567, line 160: I'd probably add a comment on this context function, the name does explain it, but just covers this whole area of functions explaining that self host licenses can connect to the cloud to use certain cloud services.

mike12345567 (approved):
LGTM!

Issue #15860, opened 2025-04-01
Inserting rows using automations don't work on single select columns


hassan-insurely:
As a result of this PR https://github.com/Budibase/budibase/pull/15781, you can no longer set the value of a single select when creating a row based on input data, but can only be hard-selected from the dropdown list

PR #15863, opened 2025-04-01
Fix for Automation bindings in Row actions

deanhannigan:
## Description
The binding drawer was mistakenly removed as part of the recent automation UI rework here: https://github.com/Budibase/budibase/pull/15781

This meant that it was no longer possible to configure bindings for select fields when working with Row actions.

## Addresses
- https://github.com/Budibase/budibase/issues/15860

## Launchcontrol
Fix for row step binding support when creating or updating row.


aptkingston (approved):
Nice work Dean!

PR #15865, opened 2025-04-02
Add license feature for PDF and update UI to account for it

aptkingston:
## Description
Makes PDF a premium feature. Need to merge this before I can update pro and account portal to include the new feature in licenses.

Also just swaps the order of the form and PDF templates so that PDF is last.

image

Pro PR: https://github.com/Budibase/budibase-pro/pull/375


deanhannigan (approved):
LGTM!

PR #15881, opened 2025-04-03
Fix popover root selector

aptkingston:
## Description
Popover root selected was misconfigured, resulting in an invalid selector.


deanhannigan (approved):
LGTM!

PR #15883, opened 2025-04-03
Fix update queries against Google Sheets datasources.

samwho:
## Description

As part of this PR I've added tests for all 4 query verbs against Google Sheets datasources to make sure this doesn't regress in future. Had to update the way our Google Sheets mock works when deleting data to make this work correctly, the way that the Google Sheets API does this is to bring all rows below the deleted row up 1. We weren't doing this, which caused a mismatch between tests and reality.

## Addresses
- https://linear.app/budibase/issue/BUDI-9153/gsheets-cannot-read-properties-of-undefined-reading-schema

## Launchcontrol

- Fixes an undefined property error when attempting to run an update query against a Google Sheets datasource.

### DIFF
--- a/packages/server/src/api/controllers/datasource.ts
+++ b/packages/server/src/api/controllers/datasource.ts
@@ -1,5 +1,4 @@
 import { getQueryParams, getTableParams } from "../../db/utils"
-import { getIntegration } from "../../integrations"
 import { invalidateCachedVariable } from "../../threads/utils"
 import { context, db as dbCore, events } from "@budibase/backend-core"
 import {
@@ -173,14 +172,6 @@ export async function update(
   )
   await events.datasource.updated(datasource)
   datasource._rev = response.rev
-
-  // Drain connection pools when configuration is changed
-  if (datasource.source && !isBudibaseSource) {
-    const source = await getIntegration(datasource.source)
-    if (source && source.pool) {
-      await source.pool.end()
-    }
-  }
 
   ctx.message = "Datasource saved successfully."
   ctx.body = {
--- a/packages/server/src/integrations/base/query.ts
+++ b/packages/server/src/integrations/base/query.ts
@@ -3,7 +3,7 @@ import {
   EnrichedQueryJson,
   QueryJson,
 } from "@budibase/types"
-import { getIntegration } from "../index"
+import { getIntegration, isDatasourcePlusConstructor } from "../index"
 import sdk from "../../sdk"
 import { enrichQueryJson } from "../../sdk/app/rows/utils"
 
@@ -29,8 +29,7 @@ export async function makeExternalQuery(
 
   const Integration = await getIntegration(json.datasource.source)
 
-  // query is the opinionated function
-  if (!Integration.prototype.query) {
+  if (!isDatasourcePlusConstructor(Integration)) {
     throw "Datasource does not support query."
   }
 
--- a/packages/server/src/integrations/googlesheets.ts
+++ b/packages/server/src/integrations/googlesheets.ts
@@ -18,7 +18,6 @@ import {
   DatasourcePlusQueryResponse,
   BBReferenceFieldSubType,
   EnrichedQueryJson,
-  TableSchema,
 } from "@budibase/types"
 import { OAuth2Client } from "google-auth-library"
 import {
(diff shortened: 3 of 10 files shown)


packages/server/src/integrations/tests/utils/googlesheets.ts
```diff
@@ -540,9 +540,28 @@ export class GoogleSheetsMock {
...
 
-    this.iterateRange(range, cell => {
-      cell.userEnteredValue = this.createValue(null)
-    })
+    const sheet = this.getSheetById(range.sheetId)
+    if (!sheet) {
+      throw new Error(`Sheet ${range.sheetId} not found`)
+    }
+
+    if (!range.startRowIndex || !range.endRowIndex) {
```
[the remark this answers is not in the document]
samwho: Good bot.

packages/server/src/api/controllers/datasource.ts
```diff
@@ -174,14 +173,6 @@ export async function update(
...
   datasource._rev = response.rev
 
-  // Drain connection pools when configuration is changed
-  if (datasource.source && !isBudibaseSource) {
-    const source = await getIntegration(datasource.source)
-    if (source && source.pool) {
-      await source.pool.end()
-    }
-  }
-
```
samwho, line 184: As far as I can tell, `source.pool` is never a thing. The only integration that uses a `pool` is MSSQL, and it's an instance property not a static one. Please correct me if I'm wrong @mike12345567 🙏

packages/server/src/integrations/googlesheets.ts
```diff
@@ -626,7 +626,7 @@ export class GoogleSheetsIntegration implements DatasourcePlus {
     sheet: string
     rowIndex: number
     row: any
-    table: Table
+    table?: Table
```
samwho, line 629: This was the root of the problem: `table` doesn't appear to always be passed (ever be passed?). So I've handled it not being there. Not sure if it's ever passed to be honest, and whether it should be removed entirely.

The typing on `IntegrationBase` is quite bad, lots of anys, allowing this sort of thing through.

packages/server/src/integrations/tests/utils/googlesheets.ts
```diff
@@ -540,9 +540,28 @@ export class GoogleSheetsMock {
...
+      throw new Error(`Sheet ${range.sheetId} not found`)
+    }
+
+    if (range.startRowIndex === undefined || range.endRowIndex === undefined) {
+      throw new Error("Range must have start and end row indexes")
+    }
+
+    const totalRows = sheet.data[0].rowData.length
+    if (totalRows < range.endRowIndex) {
+      throw new Error(
```
adrinr, line 554: ❤️

packages/types/src/documents/app/query.ts
```diff
@@ -8,6 +8,8 @@ export interface QuerySchema {
   subtype?: string
 }
 
+export type QueryVerb = "read" | "create" | "update" | "delete"
```
adrinr, line 11: Are these query verbs not used for mongo queries, as well?

packages/server/src/integrations/base/query.ts
```diff
@@ -29,8 +29,7 @@ export async function makeExternalQuery(
 
   const Integration = await getIntegration(json.datasource.source)
 
-  // query is the opinionated function
-  if (!Integration.prototype.query) {
+  if (!isDatasourcePlusConstructor(Integration)) {
     throw "Datasource does not support query."
```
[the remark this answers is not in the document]
samwho: The `query` method is present on `DatasourcePlus` instances and not `IntegrationBase` instances. I changed the conditional purely for the type narrowing.

packages/types/src/documents/app/query.ts
samwho: That's a good point, MongoDB appears to also support `"aggregate"`. If I add that here, though, I get some type check failures because `aggregate` is not a function on `IntegrationBase`.

I think the reality is that typing around this area isn't great today, and it's probably a bigger piece of work to fix it in full. I've improved it a tiny bit in this PR but I think I'd like to just leave it at that for now.

adrinr (approved):
LGTM

PR #15898, opened 2025-04-07
Add CSP whitelist feature to custom app scripts

aptkingston:
## Description
Opening for testing in feature branches.

This PR adds the ability to specify domains to be added to the content security policy, allowing resources and AJAX requests to domains which would otherwise be blocked. These domains are added to the `default-src` CSP directive, whitelisting them for most common use cases.

Notes on usage:
- If you add a third party script from an external domain in a custom app script, you *do not* need to also add that domain to the CSP whitelist, because the custom app script feature uses a nonce to automatically authenticate that resource
- The CSP is active in cloud and active by default in self host, but it can be disabled in self host. If disabled, you don't need this feature because all domains will be allowed.
- You only need this feature if the scripts you add either fetch additional resources at runtime from other domains or make requests to other domains

TODO: get docs written about this feature and add link to frontend.

image
image

### DIFF
--- a/packages/backend-core/src/middleware/contentSecurityPolicy.ts
+++ b/packages/backend-core/src/middleware/contentSecurityPolicy.ts
@@ -1,6 +1,7 @@
 import crypto from "crypto"
 import { app } from "../cache"
-import { Feature } from "@budibase/types"
+import { Feature, Ctx } from "@budibase/types"
+import { Next } from "koa"
 
 const CSP_DIRECTIVES = {
   "default-src": ["'self'"],
@@ -88,7 +89,7 @@ const CSP_DIRECTIVES = {
   "worker-src": ["blob:"],
 }
 
-export async function contentSecurityPolicy(ctx: any, next: any) {
+export async function contentSecurityPolicy(ctx: Ctx, next: Next) {
   const nonce = crypto.randomBytes(16).toString("base64")
   ctx.state.nonce = nonce
   let directives = { ...CSP_DIRECTIVES }
--- a/packages/backend-core/src/middleware/contentSecurityPolicy.ts
+++ b/packages/backend-core/src/middleware/contentSecurityPolicy.ts
@@ -1,7 +1,7 @@
 import crypto from "crypto"
 import { app } from "../cache"
 import { Feature } from "@budibase/types"
-import { env as coreEnv } from "@budibase/backend-core"
+import coreEnv from "../environment"
 
 const CSP_DIRECTIVES = {
   "default-src": ["'self'"],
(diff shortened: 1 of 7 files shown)


packages/server/src/api/index.ts
```diff
@@ -66,9 +71,13 @@ if (apiEnabled()) {
...
     .use(pro.licensing())
     .use(currentApp)
-    .use(auth.auditLog)
-    .use(migrations)
-    .use(cleanup)
+
+  // Add CSP as soon as possible - depends on licensing and currentApp
+  if (!coreEnv.DISABLE_CONTENT_SECURITY_POLICY) {
+    router.use(middleware.csp)
+  }
```
aptkingston, line 78: The CSP middleware has been removed from the base koa file and instead spliced in here, after the licensing and current app middle which this feature depends on.

packages/backend-core/src/middleware/contentSecurityPolicy.ts
```diff
@@ -87,27 +89,45 @@ const CSP_DIRECTIVES = {
 }
 
 export async function contentSecurityPolicy(ctx: any, next: any) {
```
samwho, line 91: While you're here, can you fix these anys?

```suggestion
import { Ctx } from "@budibase/types"
import { Next } from "koa"
export async function contentSecurityPolicy(ctx: Ctx, next: Next) {
```

samwho (approved):

packages/builder/src/pages/builder/app/[application]/settings/scripts.svelte
```diff
@@ -144,6 +144,17 @@
         minHeight={200}
         placeholder="&lt;script&gt;...&lt;/script&gt;"
       />
+      <Label size="L">
+        CSP whitelist<br />
+        <Link href="https://docs.budibase.com/docs/app-scripts" target="_blank">
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/settings/scripts.svelte:149]: Is this link correct?

Comment (aptkingston) [reply packages/builder/src/pages/builder/app/[application]/settings/scripts.svelte:149]: Nope it's just a placeholder - I have a TODO in the description above just saying I need to get a proper docs link 👍

packages/backend-core/src/middleware/contentSecurityPolicy.ts
aptkingston: I can indeed, no problemo.

aptkingston: Actually I can't type `next` as backend-core doesn't have `koa` as a dependency. So it's either leave it as `any` or add koa as a dependency. Happy to do either. Or move this middleware over to the server I suppose. Any preference?

aptkingston: It already depends on some other koa packages, so I've just added in koa itself too which lets me type everything properly.

Issue #15904, opened 2025-04-08
When importing .CSV to create a table, column types do not align


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
When creating a new table using the internal DB, the user(builder) is offered a handful of column-types to choose from for each column, though the names do not match up.

A column from which a single option can be chosen is referred to as a `Single Select` column in the database, with its counterpart being a `Multi-select` column. When importing CSV, these are referred to as Options and Multi-select, respectively. This could be especially confusing for users who do not speak English as a first language.

![Image](https://github.com/user-attachments/assets/56a7ab94-f8d6-4fe3-9be6-94ff6aefb80c)

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the data section
2. import a .csv file
3. change the column types

**Expected behaviour**
Should refer to the single-select column by name, instead of "Options"

**Screenshots**
Importing CSV
![Image](https://github.com/user-attachments/assets/ebbb7a24-5fd4-4961-93b9-58d45fc1409b)

Once the column has been created
![Image](https://github.com/user-attachments/assets/a4968847-9b48-4af6-b7cf-e863f80658f5)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 135.0.7049.42 (Official Build) (arm64)

PR #15905, opened 2025-04-08
updates column name in import modal

mikesealey:
## Description
Updates the name of the single-option olumn type in the Import CSV modal from Options to Single-select in order to conform with column types in the internal database.

## Addresses
https://github.com/Budibase/budibase/issues/15904
## Screenshots
Before
image

After
image

## Launchcontrol
_Improves column-type conformity in TableDataImport modal._


aptkingston (approved):
LGTM, but I would remove the hyphen since we don't use that elsewhere for single select.

mikesealey:
Removed hyphen from both Single select and Multi select columns

image

Which aligns with the Column type names in the database
image

PR #15906, opened 2025-04-08
Rating Field Component

andz-bb:
## Description
A new rating field component loosely based on my star rating plugin. It's a Budibase form field component that sets a numeric value. Icon style, colour, quantity etc. can all be configured. Actions are available on change with the value available on context. It respects form settings like disabled and readonly. Validation is supported

## Addresses
[BUDI-9128](https://linear.app/budibase/issue/BUDI-9128/convert-star-rating-plugin-to-native-component)

## Screenshots
image

## Launchcontrol
Add rating field component

## Feature branch env
[Feature Branch Link](http://fb-budi-9128-convert-star-rating-plugin.budibase.dev)


packages/client/src/components/app/forms/RatingField.svelte
```diff
@@ -0,0 +1,152 @@
...
+</script>
+
+<Field
+  {label}
+  {field}
+  {disabled}
+  {readonly}
+  {validation}
+  type={FieldType.NUMBER}
+  defaultValue="0"
```
aptkingston, line 87: We should probably use numeric 0 here rather than string 0, as otherwise we can send up sending up strings for number columns. The API does seem to handle this gracefully, but for correctness I think option for the number would be better.

aptkingston (approved):
Great work @andz-bb! You've really thought of everything here. There's a really comprehensive list of settings, all of which have sensible defaults and limits. Everything worked perfectly and I couldn't break it.

I've left one very minor suggestion, and the only other thing I spotted is that there's no default value setting. Was that deliberately left out for any reason? Probably fine without it anyway, but just curious.

Super job!

packages/client/src/components/app/forms/RatingField.svelte
andz-bb: agreed, it's how I had it before I wrote the component in TypeScript. I'm not sure how to do it other than by expanding the type definition in a load of areas. Is this something we would want to do?

```diff
Field.svelte (and lots of other places where it's only typed as a string/undefined)

- defaultValue: string | undefined
+ defaultValue: string | number | undefined
```

NumberField.svelte handles it like this - should I do something similar? 

```ts
<script>
  import StringField from "./StringField.svelte"

  export let defaultValue

  const parseNumber = val => {
    if (val == null) {
      return null
    }
    return isNaN(val) ? null : parseFloat(val)
  }
</script>

<StringField
  {...$$props}
  type="number"
  defaultValue={parseNumber(defaultValue)}
/>
```

aptkingston: Ah I didn't realise that defaultValue is typed like that in the core `Field`. That's incorrect - it should be `any` or (even better) a generic matching the type of the real value.

In that case I'm happy to ignore this for now and stick with how you have it 👍

PR #15927, opened 2025-04-10
Provide user with context if license key has already been activated

melohagan:
## Description
If a license key has already been activated, inform the user. 

## Addresses
- https://linear.app/budibase/issue/GRO-879/strange-ux-for-used-license-keys

## Screenshot
Screenshot 2025-04-10 at 09 49 13


jvcalderon (approved):
LGTM

PR #15930, opened 2025-04-10
CSP whitelist validation

aptkingston:
Validate CSP whitelist domains before applying them to prevent illegal characters.

## Addresses
- https://linear.app/budibase/issue/BUDI-9231/app-scripts-no-validation-on-csp-whitelist-invalid-character


packages/backend-core/src/middleware/contentSecurityPolicy.ts
```diff
@@ -89,6 +89,8 @@ const CSP_DIRECTIVES = {
   "worker-src": ["blob:", "'self'"],
 }
 
+const CSPDomainRegex = /^[A-Za-z0-9-*:/.]+$/
```
samwho, line 92: Is this definitely good enough? If someone enters "........" which this regex will consider valid, is that going to cause problems? Does Node not have some robust way to validate domains?

packages/backend-core/src/middleware/tests/contentSecurityPolicy.spec.ts
```diff
@@ -132,4 +132,49 @@ describe("contentSecurityPolicy middleware", () => {
     expect(app.getAppMetadata).toHaveBeenCalledWith(appId)
     expect(next).toHaveBeenCalled()
   })
+
+  it("should filter out invalid domains", async () => {
```
samwho, line 136: Should we instead have some integration tests? Make an actual request and verify the header?

packages/backend-core/src/middleware/contentSecurityPolicy.ts
aptkingston: Yea if you enter something like that it'll log a different error saying it's an invalid source and it will be ignored, which I think is harmless. As long as it doesn't brick entirely, which happens if illegal characters are entered. Validating the domains would certainly be more robust but I don't think it's necessarily required here, since if someone wants to enter harmless garbage then they can.

packages/backend-core/src/middleware/tests/contentSecurityPolicy.spec.ts
aptkingston: Do you not think this is good enough? It's hitting the exact same logic and verifying the header set in the koa context. Either way though, I'd like to get this important fix out in the meantime and I can investigate integration tests separately if you think they're needed as well.

samwho (approved):

packages/backend-core/src/middleware/tests/contentSecurityPolicy.spec.ts
samwho: It'll do for the time being but generally we've been moving away from tests like this on Firestorm. Mocking almost everything is very fragile and often ends up masking real problems. I don't know to what extent we'd have caught this CSP problem with integration tests, probably not at all, but generally I like to hit as much real code in tests as possible.

packages/backend-core/src/middleware/contentSecurityPolicy.ts
samwho: I'm iffy about it, I think we could try and be more helpful if the user makes a mistake, but in the spirit of getting this fix out it's fine.

PR #15934, opened 2025-04-10
Add tests for budibase AI self host quota.

samwho:
## Description

To make sure that we're storing quota in the right place for self-host requests to Budibase AI, I've created a new test that asserts on the document ID.


packages/server/src/api/routes/tests/ai.spec.ts
```diff
@@ -301,13 +309,29 @@ describe("BudibaseAI", () => {
...
+    async function getQuotaUsage() {
+      return await context.doInSelfHostTenantUsingCloud(
+        config.getTenantId(),
+        async () => {
+          return await quotas.getQuotaUsage()
+        }
+      )
+    }
+
+    it.only("handles correct chat response", async () => {
```
adrinr, line 321:
```suggestion
    it("handles correct chat response", async () => {
```

adrinr (approved):
LGTM, just remove the `only`

PR #15935, opened 2025-04-10
Check for DB prefix in a slightly more robust way.

samwho:


melohagan (approved):

samwho:
These test failures are very strange. For some reason, changing this function makes a logout test fail because it can't find a tenant ID. Going to close this for now.

PR #15953, opened 2025-04-14
Expose LLM provider config via /api/global/self.

samwho:
## Description

This is so that the frontend can know if the user has an AI provider configured and, if not, show the right calls to action in the UI.

## Addresses
- https://linear.app/budibase/issue/BUDI-9244/expose-current-llm-provider-in-apiself


PClmnt (approved):
LGTM, thanks for getting that in so fast!

PR #15954, opened 2025-04-14
Capture codegen accept/reject events in PostHog.

samwho:
## Description

Creates analytics events for whether users accept or reject code generated with LLMs, so we can track whether changes are improving quality or not.

## Addresses
- https://linear.app/budibase/issue/BUDI-9145/send-events-to-posthog-on-acceptreject-of-ai-suggestions


shogunpurple (approved):
LGTM!

PR #15955, opened 2025-04-14
Remove the check for Feature.BUDIBASE_AI

samwho:
## Description

Pro PR: https://github.com/Budibase/budibase-pro/pull/390


packages/server/src/api/controllers/row/staticFormula.ts
```diff
@@ -156,14 +155,9 @@ export async function finaliseRow(
...
 
-  const aiEnabled =
-    (await pro.features.isBudibaseAIEnabled()) ||
-    (await pro.features.isAICustomConfigsEnabled())
-  if (aiEnabled) {
-    row = await processAIColumns(table, row, {
-      contextRows: [enrichedRow],
-    })
-  }
+  row = await processAIColumns(table, row, {
```
adrinr, line 158: One think is that everybody has access to it, but does it not need to be enabled?

samwho: Currently it is enabled by default if you have access to it. This PR can't be merged until we put a control in place that has it disabled by default, and requires toggling a setting to activate. I believe Ben and Joe are working on designs for that today.

shogunpurple (approved):
just need to fix that wee test then LGTM 👍

PR #15966, opened 2025-04-16
Show AI JS prompt in expanded state when in drawer

PClmnt:
## Description
Updates the JS generator component to open fully when in binding editor mode.

### DIFF
--- a/packages/builder/src/components/common/CodeEditor/AIGen.svelte
+++ b/packages/builder/src/components/common/CodeEditor/AIGen.svelte
@@ -40,12 +40,14 @@
   let switchOnAIModal: Modal
   let addCreditsModal: Modal
 
+  const thresholdExpansionWidth = 350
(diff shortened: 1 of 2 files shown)


packages/builder/src/components/common/CodeEditor/AIGen.svelte
```diff
@@ -43,11 +43,22 @@
...
+      ? true
+      : expanded
+
   $: creditsExceeded = $licensing.aiCreditsExceeded
   $: disabled = suggestedCode !== null || !aiEnabled || creditsExceeded
-  $: if (expandedOnly) {
+
+  $: if (
+    expandedOnly ||
+    (expanded && parentWidth !== null && parentWidth > 350)
```
adrinr, line 57: Could we type this magic number?

adrinr (approved):

PR #15975, opened 2025-04-17
AI table generation UI

adrinr:
## Description
Frontend on the AI table generation. This does not have any backend logic, just the UI for it. The piping and the "post" creation will come on further PRs. This is hidden behind a FF, so it is not visible to anybody.

### New UI
image

### Submitting a prompt

https://github.com/user-attachments/assets/097070d6-5faf-4fbf-899b-dd74f69c1001

### Using an "example" prompt

https://github.com/user-attachments/assets/2c95f91f-6304-48bc-a575-31b193cbc0a0

### Responsive designs
image


packages/frontend-core/src/api/ai.ts
```diff
@@ -23,4 +25,8 @@ export const buildAIEndpoints = (API: BaseAPIClient): AIEndpoints => ({
       body: req,
     })
   },
+  generateTables: async prompt => {
+    console.warn({ prompt })
```
adrinr, line 29: This will be implemented in further PRs

PClmnt (approved):

aptkingston:
Looks great @adrinr 👌 

I know it'll probably be addressed in the PR when we actually handle the prompt, but it would make sense to disable submitting the input if there is no prompt specified.

The only other thing I can see is that the the "expand" logic doesn't seem to work - it doesn't expand and the border disappears instead of animating. Maybe you know that already but just checking!

Super minor but I would also add a few pixels left margin to the logo as well. I think 4px looks good to given even spacing all around:
image

PR #15977, opened 2025-04-17
Inline filter component

deanhannigan:
## Description
A new filter component has been added to the builder. You can now directly add a filter component to `Table` and `Data Provider` components with customisable layout and configurable display names.
- The filter also supports optional persistence in local storage. So any complex filters you applied will be maintained when refreshing the page.
- You can also optionally provide a 'Clear all' button to allow user to reset their filter entirely instead of removing each filter button manually.

## Addresses
- https://linear.app/budibase/project/filter-component-544306c4894c
- **Fix** - Stop `NumberInput` throwing CSP errors when focusing on the field. The inline action to trigger a select on fix has been replaced.

## Screenshots
Builder configuration
Screenshot 2025-04-22 at 09 21 18

Supported components will appear in `Target Component` drop down
Screenshot 2025-04-22 at 09 23 15

In app
Screenshot 2025-04-22 at 09 21 50

## Launchcontrol
New customisable filter added to the builder

## Feature branch env
[Feature Branch Link](http://fb-feature-filter-component.budibase.dev)


packages/client/src/components/app/GridBlock.svelte
```diff
@@ -51,15 +54,84 @@
...
+    let groups = Object.entries(extensions).map(([_, entry]) => {
+      // Assuming there should only ever be 1
+      return entry
+    })
+
+    // Combine all groups into the base
+    extended.groups = [...extended.groups, ...groups]
+
+    return extended
+  }
```
aptkingston, line 133: Just a bit of a terser version which is easier to read 👌 
```js
  const extendFilter = (initialFilter, extensions) => {
    if (!Object.keys(extensions || {}).length) {
      return initialFilter
    }
    return {
      groups: (initialFilter ? [initialFilter] : []).concat(
        Object.values(extensions)
      ),
      logicalOperator: UILogicalOperator.ALL,
      onEmptyFilter: EmptyFilterOption.RETURN_NONE,
    }
  }
```

packages/builder/src/stores/builder/components.ts
```diff
@@ -213,8 +212,10 @@ export class ComponentStore extends BudiStore<ComponentState> {
     }
 
     const def = this.getDefinition(enrichedComponent?._component)
-    const filterableTypes = def?.settings?.filter(setting =>
-      setting?.type?.startsWith("filter")
+    const filterableTypes = def?.settings?.filter(
+      setting =>
+        setting?.type?.startsWith("filter") &&
+        setting?.type !== "filterConfiguration"
```
aptkingston, line 218: I think it would make sense to move to a whitelist here to keep things more maintainable. So we just need to include `filter` and `filter/relationship`.

packages/client/src/components/app/filter/FilterButton.svelte
```diff
@@ -0,0 +1,250 @@
...
+
+  .spectrum-Button-label.truncate {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: 200px;
+  }
+
+  .spectrum-Button-label .display {
+    color: rgba(179, 222, 254, 1);
```
aptkingston, line 232: I find it quite hard to read the filters when applied:
image
We should probably use a theme aware variable with more contrast 👍

aptkingston (changes requested):
Nice work Dean! Huge improvement to the end-user filtering experience.

I'm still reviewing a few things, one of which is the changes to the BBReferenceField and RelationshipField, but I'm unable to do so because I can't seem to add a relationship field into the filter - is that expected or a bug? In my table here I have the column `Relationship`:
image
But it's not available as an option to include:
image

Figured there was enough feedback here to get started in the meantime though!

One other UX point is that this is usable in the builder iframe preview, but the filters don't actually get applied since all button actions are disabled in the preview. So I think we should probably disable interacting with this at all in the builder preview.

aptkingston (approved):
LGTM!

PR #15979, opened 2025-04-17
Backend changes to support new AI settings page

samwho:
## Description

Pro PR: https://github.com/Budibase/budibase-pro/pull/394

See description on pro PR for details.


packages/backend-core/src/platform/tenants.ts
```diff
@@ -99,3 +101,19 @@ export async function removeTenant(tenantId: string) {
...
+    return undefined
+  }
+
+  return await doInTenant(tenantId, async () => {
+    const settings = await getSettingsConfigDoc()
+    if (settings.createdAt === undefined) {
+      return undefined
+    }
+    return new Date(settings.createdAt)
+  })
```
samwho, line 118: This was the best way I could think of to determine the age of the tenant. The settings config doc appears to be created on install in self-host and at tenant creation in cloud.

packages/backend-core/src/platform/tenants.ts
```diff
@@ -99,3 +101,19 @@ export async function removeTenant(tenantId: string) {
     throw err
   }
 }
+
+export async function getTenantCreatedAt(tenantId: string) {
+  if (isSelfHostUsingCloud()) {
+    // we have no way of knowing when self-host tenants were created, so we
+    // can't return a createdAt date
+    return undefined
```
adrinr, line 109: As mentioned in here, if we store the version where they were created, we could solve this

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -219,14 +220,30 @@ async function verifyOIDCConfig(config: OIDCConfigs) {
...
+  }
+
+  let numBudibaseAI = 0
+  for (const config of Object.values(newConfig)) {
+    if (config.provider === "BudibaseAI") {
+      numBudibaseAI++
+      if (numBudibaseAI > 1) {
+        throw new BadRequestError("Only one Budibase AI provider is allowed")
+      }
+    } else {
```
adrinr, line 241: NIT: unnecessary if nesting

packages/worker/src/tests/api/configs.ts
```diff
@@ -17,21 +22,44 @@ export class ConfigAPI extends TestAPI {
...
+    const { status = 200, body } = expectations || {}
+
+    const resp = await this.request
       .post(`/api/global/configs`)
       .send(data)
       .set(this.config.defaultHeaders())
-      .expect(200)
       .expect("Content-Type", /json/)
+
+    if (resp.status !== status) {
```
adrinr, line 46: Why this instead of `.expect(status)`?

samwho: `.expect(status)` gives you no information about the content of the response, and makes debugging test failure really hard. It's why we moved away from it entirely in `packages/server` in favour of our own `Expectations`-based method.

packages/backend-core/src/platform/tenants.ts
samwho: What do you mean?

adrinr: Sorry, here is [here](https://github.com/Budibase/budibase-pro/pull/394#discussion_r2053620813)

.github/workflows/budibase_ci.yml
```diff
@@ -239,6 +239,7 @@ jobs:
       - name: Test server
         env:
           DATASOURCE: ${{ matrix.datasource }}
+          DEBUG: testcontainers*
```
adrinr, line 242: Probably not needed

adrinr (approved):
Nice work!

.github/workflows/budibase_ci.yml
samwho: I was getting timeouts in the oauth2 tests and they went away when I added this to debug them 🫠

packages/worker/src/api/controllers/global/configs.ts
samwho: It's not unnecessary as far as I can tell. `config.apiKey` is required for all providers except `BudibaseAI`. If I unnested this bit, `BudibaseAI` would get checked and fail in all cases (because `apiKey` is never set for it).

PR #15980, opened 2025-04-17
Portal app creation updates

andz-bb:
### Onboarding improvements
* When onboarding, the first app created will have the sample screen and data. 
* The representation of the first app on the onboarding side panel has been updated to reflect the new sample screen.
SCR-20250423-nejr

### App creation
* Subsequent new apps will start from scratch with empty design and data.
* `builder/portal/apps/create` and `builder/portal/apps/templates` screens have been removed. Instead, apps can be created or imported directly from the main apps screen.
image
image

* Templates screen has been replaced with a modal with the 3.0 templates.

image

## Launchcontrol
improved onboarding, streamlined app creation, new templates flow

## Feature branch env
[Feature Branch Link](http://fb-portal-app-creation-updates.budibase.dev)


aptkingston (commented):
Awesome work Andy! This is much neater and speeds things up.

Still reviewing, but a few UX things I think we could improve so far:
- When using a template we could generate nicer names. Right now it's all lowercase and. I think something like `Expense Approval App 1` would be better.
image
- For subsequent apps that don't get the sample data or screen, it still throws you into the design section after creation. 2 of the 3 screen options (table and form) can't actually be used because there's no data at all. I understand not wanting the sample screen for subsequent apps, but personally I'd probably keep the sample data for all apps since it's completely harmless and can be deleted if not wanted. Either that or we need to either disable those screen templates until data exists.
image

aptkingston (dismissed):
Not sure the app names have changed - it's still lowercase like `expense approval 3`.

~~When starting from scratch, I also get an error now when creating the workspace and being automatically pushed to the design section. This happens regardless of whether the workspace apps flag is enabled or disabled. So we'd better address that 👍~~ I think this is on master too, probably safe to ignore.
image

deanhannigan:
I had a quick run through these this morning and ran into a few minor issues. We should probably go through these at some point and make sure they're still working as intended. Other than that, the updates look great!

- Change request app

  - `/review-change-requests` - Status filter column is unset on both templates here for some reason. Loading the page throws `Invalid filter field: Change Request.Status`
  - `/administration` - needs some reflow on the pickers.

- Expense approval app

  - Missing attachments - there are a lot of image resources embedded in many of the screens and they are not present
  - Approval flow not kicking off.

- Risk management app

  - Didn't run into any immediate problems with this one 👍

packages/server/src/api/routes/tests/application.spec.ts
```diff
@@ -178,35 +194,43 @@ describe("/applications", () => {
           path.resolve(__dirname, "data", "agency-client-portal.tar.gz")
         )
 
-      const app = await config.api.application.create({
+      const newApp = await config.api.application.create({
```
deanhannigan, line 197: Would it be possible to switch this to one of the new templates like `expense-approval`? Rather that depending on one we're getting rid of. Not a blocker at all!

deanhannigan (approved):
LGTM! Nice work! I ran through this locally and didn't run into any issues. The new template cards look awesome 👍

adrinr (changes requested):
We need to handle loading status on template usage. In the video below I triple click it (no loading indicators, multiple submition allowed)

https://github.com/user-attachments/assets/a168c76d-bc80-4ed4-9584-8995a54d0352

andz-bb:
> We need to handle loading status on template usage. In the video below I triple click it (no loading indicators, multiple submition allowed)
> 
>  Screen.Recording.2025-08-04.at.12.25.15.mov

good spot, thanks @adrinr - addressed in latest commits

https://github.com/user-attachments/assets/5618672c-08fa-4e99-a952-062277a30cba

deanhannigan (commented):
Ran through this again and it seems to be working nicely. Good job @andz-bb!

adrinr (approved):
Good work! I did some extensive QA and everything works great 🚀

PR #15987, opened 2025-04-22
Markdown performance improvements

aptkingston:
## Description
This PR greatly improves performance of markdown rendering for view-only markdown components (i.e. Markdown Viewer and the new Text component).

Even 10K instances now has no noticeable performance impact on my machine. 50K lags initially, but doesn't crash the page. Both of these numbers should be well beyond what we can reasonably expect anyone to put on a single screen.

Other stuff:
- Added new aliases for text component ("heading", "title" and "markdown"):
- Added "New" badge on text component to highlight that it's new
image
- Shrunk the size of the medium size h1 (i.e. `# Title`) so that it has no overflow in grid layouts. Previously the default medium size h1 had a vertical scrollbar in its default state.
- Deprecated the markdown viewer component

## Addresses
- Fixes https://github.com/Budibase/budibase/issues/15967


deanhannigan (approved):
LGTM! Ran through some tests with 10k rows locally and, although an unlikely scenario, it was able to handle the load. Performance has improved massively. Nicely done 🎉

PR #15989, opened 2025-04-22
Safari fixes

aptkingston:
## Description
Couple of fixes for Safari.

- Fix the primary builder nav not taking up full height
Before:
image
After:
image
- Fix data section tables having a horizontal scroll bar
Before:
image
After:
image
- Fix text selection in grids, which was erroneously selecting text in all cells when dragging to bulk select cells


deanhannigan (approved):
LGTM! Reviewed behaviour against Safari and Firefox and all issues are resolved

PR #15999, opened 2025-04-23
Remove tours

aptkingston:
## Description
This PR removes all code relating to the tours we previously used for some basic onboarding and a couple of features. We haven't updated these in a long time so we're removing them.

## Feature branch env
[Feature Branch Link](http://fb-remove-tour.budibase.dev)


deanhannigan (commented):
Do we also need to remove everything around `onboardedAt` on the user/auth object if tours are removed? 

```js
// ../backend-core/src/users/events.ts
await events.user.onboardingComplete(user)
```

aptkingston:
> Do we also need to remove everything around `onboardedAt` on the user/auth object if tours are removed?
> 
> ```js
> // ../backend-core/src/users/events.ts
> await events.user.onboardingComplete(user)
> ```

Nice catch, should be removed now!

deanhannigan (approved):
LGTM!

samwho (approved):
Approving for the few back-end changes that there are.

PR #16010, opened 2025-04-24
Feature flag debug UI

samwho:
## Description

This PR introduces command palette entries for flipping feature flags. You can only access it if you have the DEBUG_UI feature flag enabled, this is intended for internal use only.

https://github.com/user-attachments/assets/608358a2-090b-4589-bd32-66707cf6f697


aptkingston (approved):
Awesome work 🚀  will make our lives a lot easier.

PR #16014, opened 2025-04-25
Don't swallow errors when serving apps

aptkingston:
## Description
Currently we swallow errors when serving apps, and try to render the app again for some reason - without required props - which of course still results in a broken app. This PR stops that behaviour, and lets the server throw a real 500 with a message if we encounter an unhandled error.


adrinr (approved):
LGTM!

PR #16015, opened 2025-04-25
Barcode QRcode

mikesealey:
## Description
This PR adds a new component to display QR Codes and Barcodes based on the value passed. This could be a static value, or one passed in using bindings/JS.

image

This component also has options such as "Display Value"
image

And also "Show logo"
image

QR Codes can also select their primary color
image

Barcode/QR Code Generators can also be used in repeaters
image

Barcode/QR Code Generator in PDF screens
image

## Addresses
- ...more if required

## App Export
[Barcode Gen-export-1745587460313.tar.gz](https://github.com/user-attachments/files/19909231/Barcode.Gen-export-1745587460313.tar.gz)

## Screenshots

## Launchcontrol

_Adds Barcode/QR Code generator._


packages/client/src/components/app/CodeGenerator.svelte
```diff
@@ -0,0 +1,146 @@
...
+    {#if codeType === "QR Code"}
+      <div class="qr-container">
+        <div bind:this={qrContainer} />
+        <div class="qr-value" style="color: {primColor}; max-width: {size}px;">
+          {showValue ? value : ""}
+        </div>
+      </div>
+    {:else}
+      <div id="barcode-container">
+        <div id="logo-and-barcode" style="height: {size}px, width: 100%">
```
aptkingston, line 80: I can see that you're using quite a lot of ids here. You should never use the same ID more than once on a page, but adding multiple of this component will create multiple elements with the same ID.

The 2 options here are:
- Generate a unique ID for each instance of the component. This is harder to style because you'll need to get this ID into your styles section, so I don't recomment this approach.
- Don't use ID attributes at all. You can style elements using classes, and you can get the reference to your DOM element for jsbarcode by using binding, like you do with the qrContainer. I checked the source and it looks like it supports passing a DOM element in as the parameter instead of a string selector.

packages/client/src/components/app/CodeGenerator.svelte
```diff
@@ -0,0 +1,146 @@
...
+      generateQr()
+    }
+  })
+
+  afterUpdate(() => {
+    if (codeType === "Barcode") {
+      generateBarcode()
+    } else {
+      generateQr()
+    }
```
aptkingston, line 61: Could you just double check that this can't cause an endless loop? `afterUpdate` fires after any component DOM update, and you're mutating the DOM inside the callback, which smells a little risky. I'd also just double check that this does properly react to all settings changes as expected - e.g. try changing all of these props at runtime and check the component redraws the barcode as expected without needing a page refresh.

Basically if you could just add a console log or something, and ensure it's not constantly firing! And just check both barcode and QR code to be sure.

The usual way to handle this would be doing a (rather large) reactive statement and passing in all the required props that we'd want to observe and re-render the barcode when they change, but this is fine too as long as it doesn't cause too many wasted invocation.

aptkingston (approved):
Nice work addressing that feedback 👌 LGTM!

PR #16017, opened 2025-04-25
Fix serving apps without themes due to defaulting a string to an object

aptkingston:
## Description
Fix serving apps without themes due to defaulting a string to an object.
This should be left as is without defaulting to an object as the `ensureValidTheme` call properly handles this and defaults it to a real theme if required.


deanhannigan (approved):
LGTM!

PR #16029, opened 2025-04-29
Table generation speed improments

adrinr:
## Description
Table generation speed improvements. Main improvements:
1. Prompt trimming
2. Parallelisation
3. Avoid processing AI columns multiple times

Pro PR:
- https://github.com/Budibase/budibase-pro/pull/398


packages/builder/src/stores/builder/screenComponent.ts
```diff
@@ -17,16 +17,9 @@ import { bindings } from "@/helpers"
...
-function reduceBy<TItem extends {}, TKey extends keyof TItem>(
-  key: TKey,
-  list: TItem[]
-): Record<string, TItem> {
-  return list.reduce<Record<string, TItem>>((result, item) => {
-    result[item[key] as string] = item
-    return result
-  }, {})
-}
+const reduceBy = utils.toMap
```
samwho, line 22: Feels like the sort of thing lodash can do.

samwho (approved):

packages/builder/src/stores/builder/screenComponent.ts
adrinr: Yup, I am never sure if I like lodash for basic use case, as it adds some external lib dependencies for stuff that is pretty simple in js/ts

PR #16044, opened 2025-04-30
Show evaluation side panel on AI suggestion

PClmnt:
## Description
Triggers the evaluation side panel whenever a user has an AI JS suggestion


packages/builder/src/components/common/CodeEditor/CodeEditor.svelte
```diff
@@ -423,6 +423,7 @@
     })
     isAIGeneratedContent = true
     dispatch("change", code)
+    dispatch("ai_suggestion")
```
PClmnt, line 426: It's much cleaner to just dispatch a new event than try and add modify the change object

samwho (approved):

PR #16045, opened 2025-04-30
Typing builder screen components

adrinr:
## Description
Typing builder screen components, ahead of some work around the area


packages/builder/src/components/common/NavItem.svelte
```diff
@@ -53,11 +53,8 @@
     scrollApi.scrollTo(bounds)
   }
 
-  const getStyle = (indentLevel, selectedBy) => {
+  const getStyle = (indentLevel: number) => {
     let style = `padding-left:calc(${indentLevel * 14}px);`
-    if (selectedBy) {
-      style += `--selected-by-color:${helpers.getUserColor(selectedBy)};`
-    }
```
aptkingston, line 60: Was this not required to be able to show if other users have resources selected?

aptkingston (approved):
LGTM! Just one question about the removal of the colouring for other users in NavItems.

packages/builder/src/components/common/NavItem.svelte
adrinr: This did not work. `selectedBy` is an array of users, this will not work fine. The different coloring is already  handled in the `UserAvatar` component

PR #16046, opened 2025-04-30
Remove VerificationPromptBanner

melohagan:
## Description
An unverified Account can no longer be created. Removing unused code.

## Addresses
- https://linear.app/budibase/issue/GRO-1169/remove-old-endpoint-call-in-budibase


jvcalderon (approved):
LGTM

PR #16057, opened 2025-05-01
Add option to refresh all data providers

### LINKED ISSUE #15920: Add action to refresh all data sources on a page
As a user, I would like to have more control over data refresh behavior In Budibase. Especially when working with screen models in this case.

### To address this limitation, I propose one of the following solutions:
1. Automatically Refresh All Data Providers on Modal Close:
Implement an action to automatically refresh all data providers on the source screen when a modal is closed.
2. Provide a Configurable Option:
Allow users to choose whether or not to trigger a data refresh on modal close, either globally, per screen, or data provider.

This becomes more of an issue when custom queries are used as data sources on screens. While data providers do refresh automatically when the form and the provider share the same source, this is not the case when they differ. No refresh occurs in those situations, especially when a form is submitted within a screen modal. Currently, there is no built-in mechanism to force a refresh of the parent screen's data providers, as the system lacks the needed context.

aptkingston:
## Description
Adds an option to refresh all data providers when using the "Refresh data provider" action.

image

Fixes https://github.com/Budibase/budibase/issues/15920.

## Launchcontrol
Add option to refresh all data providers when using the "Refresh data provider" action.


deanhannigan (approved):
LGTM! 🎉  Ran through all the refreshable sources and everything worked perfectly!

PR #16062, opened 2025-05-02
Ensure primary display column is always required

### LINKED ISSUE #15899: Display Column isn't required
**Checklist**

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
Display Column should be required - shows as "required - checked" in the "edit column" menu, but it is still possible to write a row with no value in the display column.

**To Reproduce**
1. Using the employees sample data, save a row with no information populated.

**Expected behavior**
Should prevent the user/builder from saving a row without a value in the display column.

**Screenshots**
https://jam.dev/c/21a3f02c-9ee5-44d3-bf2c-8628320a5a13

 - OS: MacOS
 - Browser Chrome
 - Version 134.0.6998.166 (Official Build) (arm64)

aptkingston:
## Description
Fixes an issue where the server did not enforce primary display columns to be required, but instead relied on the correct constraint also being present.

Now, the server validates that primary display column are present regardless of other validation rules.

Test added.

## Addresses
- Fixes https://linear.app/budibase/issue/BUDI-9217/display-column-isnt-required
- Fixes https://github.com/Budibase/budibase/issues/15899


samwho (commented):
Is this not going to cause backward compatibility problems? I feel like it's quite likely folks are going to be out there in the wild relying on the fact that the primary display column isn't required and this will break their workflows. 

Also, with the way this is implemented, wouldn't it be the case that the primary display column doesn't look like it's required in the UI because that constraint is never saved on the table, it's added in at validation time?

aptkingston:
> Is this not going to cause backward compatibility problems? I feel like it's quite likely folks are going to be out there in the wild relying on the fact that the primary display column isn't required and this will break their workflows.
> 
> Also, with the way this is implemented, wouldn't it be the case that the primary display column doesn't look like it's required in the UI because that constraint is never saved on the table, it's added in at validation time?

In terms of backwards compatibility, it quite probably will cause validation errors for the small subset of people who have managed to create this scenario. I'm not sure that's necessarily a bad thing though, because we make plenty of assumptions elsewhere that primary display columns should always exist, and I don't think we should continue to allow this. This will force people to address the problem. We've seen other issues (such as relationship pickers not being able to display rows properly) caused by the lack of a primary display value. I'd rather force people to address this instead of pushing a more passive change which allows them continue indefinitely.

As for the UI, it always shows the required toggle as active (and is disabled) when editing the primary display column. I'd already made an assumption on the frontend tha the display column is also always required. The constraint is added by the UI when setting primary display - this issue only occurs when the primary display is not created via the frontend - so things like table import or some other operation which forces a new primary display to be set by the backend. I can add additional checks though to ensure the frontend sets the required flag when editing the display column, in case it isn't there for some reason. That won't make a functional difference after this PR, but at least makes the constraints consistent with behaviour.

samwho (approved):

PR #16063, opened 2025-05-02
Fix AI Config screen UI issue and add some tests

PClmnt:
## Description
When loading the AI Config page you would occasionally see a brief flicker of the banner before it loaded fully, just added a mounted check to ensure this doesn't happen. 

Also added some tests to check certain screen states and a vite config change to support lifecycle methods in Vitest.


packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -153,80 +154,83 @@
       aiConfig = (await API.getConfig(ConfigType.AI)) as AIConfig
       const licenseKeyResponse = await API.getLicenseKey()
       hasLicenseKey = licenseKeyResponse?.licenseKey
+      mounted = true
```
adrinr, line 157: What would happen if there is an error getting this? The full section will not be rendered. also, this might cause confusion as the mounted will be false for mounted components that failed. It should either be moved outside the try/catch or changed to `aiConfig` loaded or similar. On the other hand, if it's the latest, it could be as easy to have the check to be `if(aiConfig && licenseKey)`

adrinr (approved):

packages/builder/src/pages/builder/portal/settings/ai/index.svelte
```diff
@@ -159,74 +159,76 @@
...
-
-  {#if !activeProvider && !$bannerStore}
-    <div class="banner">
-      <div class="banner-content">
-        <div class="banner-icon">
-          <img src={BBAI} alt="BB AI" width="24" height="24" />
-        </div>
-        <div>Try BB AI for free. 50,000 tokens included. No CC required.</div>
+{#if aiConfig}
+  <Layout noPadding>
```
adrinr, line 163: The flicker will go away, but will this not the full section to flash in?

PClmnt: The UI won't load until the onMount lifecycle finishes?

PR #16076, opened 2025-05-06
Fix issue where clicking icon was clearing ai prompt

PClmnt:
## Description
Clicking on the BBAI icon in an expanded state would cause the prompt to be cleared. This checks to see if it is an `expandedOnly` state and doesn't do anything if it is.


packages/builder/src/components/common/ai/AIInput.svelte
```diff
@@ -93,8 +93,10 @@
       class:loading={promptLoading}
       class:disabled={expanded && disabled}
       on:click={e => {
```
adrinr, line 95: Could we remove the `on:click` completely, so we don't have the hover effect?

packages/builder/src/components/common/CodeEditor/AIGen.svelte
```diff
@@ -26,11 +26,11 @@
 
   const thresholdExpansionWidth = 350
 
-  $: expanded =
+  $: expandedOnly =
```
adrinr, line 29: Is this right? `expandedOnly` should depend on the parent width?

samwho (approved):

PR #16083, opened 2025-05-06
Graceful termination of the worker/app servers

calexiou:
## Description
Worker/app servers might terminate without waiting all pending operations to be completed. This introduces a graceful termination mechanism using `http-graceful-shutdown`.


calexiou:
@adrinr thanks for the review! If you have some time, give it another go. Addressed your points.

adrinr (approved):
LGTM! Thanks for addressing that!

PR #16085, opened 2025-05-07
Don't clear user prompt if the user rejects a suggestion

PClmnt:
## Description
Fixes an issue where if the user rejected an AI JS suggestion the prompt would be cleared regardless.


aptkingston (approved):
LGTM! Title confused me since I think that should say "only clear if the user *accepts*", but I can see in the code it only clears when rejecting, so that's perfect!

PClmnt:
> LGTM! Title confused me since I think that should say "only clear if the user _accepts_", but I can see in the code it only clears when rejecting, so that's perfect!

Yep, whoops. Updated the title

PR #16090, opened 2025-05-07
Move from Source Sans Pro to variable weight Source Sans 3

aptkingston:
## Description
Updates everything (builder, apps, BBUI) to use the variable weight Source Sans 3 font instead of Source Sans Pro. This gives us access to all weights between 200 and 900.

We've always bundled the fonts locally - previously we installed SSP from NPM and then copied it from node modules into the dist folder. Instead, I've directly sourced and included the required WOFF2s and now copy those from source to dist. So it's all locally bundled, and now with no dependencies.


adrinr (approved):
LGTM!

PR #16092, opened 2025-05-07
Icon picker to ts

adrinr:
## Description
Converting icon picker to ts. Also removing unused props that will allow reusing this component without extra logic


deanhannigan (approved):
LGTM!

PR #16095, opened 2025-05-08
Proxy the 'refresh all data providers' action across screen peek modals

aptkingston:
## Description
This PR proxies the "refresh all data providers" action across "peek" screen modals. This means that if you use the "refresh data provider" action inside a screen that you're rendering inside a modal, then choose the "all data providers" option, this will now proxy to the parent app and all data providers on the parent screen will refresh. 

This already happens with the "refresh data provider" action if you choose a specific provider - any providers on the parent screen using that datasource would refresh. This just makes it work with the new "all data providers" option.

Also a little cleanup of messages sent between instances of the client app.

## Feature branch env
[Feature Branch Link](http://account.fb-refresh-all-datasources-peek.budibase.dev)


deanhannigan (approved):
LGTM! This is awesome! Didn't run into any issues 🎉

PR #16106, opened 2025-05-09
Filter pill buttons are now resizable

deanhannigan:
## Description
Added a `size` prop to the filter component to allow setting the size of the filter pill buttons
- The default size is set to `Medium`
- `Small`, `Medium` and `Large` options are available.
- The `Persist filters` and `Clear filters` options have also been moved above `Fields` to stop them getting pushed out of the way by the `Fields` list.

## Screenshots
Screenshot 2025-05-09 at 16 53 29

## Launchcontrol
Filter pill buttons are now resizable


aptkingston (approved):
LGTM!

PR #16120, opened 2025-05-12
Simplify Redis initialisation logic, instrument with DataDog

samwho:
## Description

Pro PR: https://github.com/Budibase/budibase-pro/pull/410

To debug some ongoing problems in production with Redis reconnection, I've refactored our Redis client logic to be much simpler and I've also instrumented it with DataDog spans so we can better see what's going on.


shogunpurple (approved):
LGTM

PR #16129, opened 2025-05-13
New project apps backend crud

adrinr:
## Description
First PR for the project app. Creating the basic crud and the migration for it.
The new structure will change from:
`app -> screens`
to:
`app -> project apps -> screens`

This will change the appPackage response from containing an array of screens to contain an array of project apps. The migration will run at runtime when fetching it (as it's a lean one), so we will not need to run a global migration.

Contains: 
- https://github.com/Budibase/budibase/pull/16145
- https://github.com/Budibase/budibase/pull/16164


packages/server/src/api/controllers/projectApp.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export async function edit(
+  ctx: Ctx<UpdateProjectAppRequest, UpdateProjectAppResponse>
+) {
+  const { body } = ctx.request
+
+  if (ctx.params.id !== body._id) {
+    ctx.throw("Path and body ids do not match", 400)
```
samwho, line 47: Why have both?

packages/server/src/api/routes/projectApp.ts
```diff
@@ -0,0 +1,52 @@
...
+  ...baseSchema,
+})
+
+const updateSchema = Joi.object({
+  _id: Joi.string().required(),
+  _rev: Joi.string().required(),
+  ...baseSchema,
+})
+
+function projectAppValidator(
```
samwho, line 26: ❤️

packages/server/src/sdk/app/projectApps/index.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export async function get(id: string): Promise<ProjectApp | undefined> {
+  const db = context.getAppDB()
+  const projectApp = await db.tryGet<ProjectApp>(id)
+  return projectApp
+}
+
+export async function create(
+  projectApp: Omit<ProjectApp, "_id" | "_rev" | "createdAt" | "updatedAt">
```
samwho, line 32: Kinda feels like this pattern would be a nice utility type. `WithoutDocMetadata<ProjectApp>` or something.

packages/server/src/sdk/app/screens/screens.ts
```diff
@@ -1,13 +1,53 @@
...
         include_docs: true,
       })
     )
   ).rows.map(el => el.doc!)
+
+  if (screens.every(s => s.projectAppId)) {
+    return screens
+  }
+
+  await migrateToProjectApp(screens)
```
samwho, line 21: I'm a touch uneasy about this. We'll never, ever be able to remove it. I feel like having migrations all in one place (e.g. the migration system we built for SQS) is better than having them scattered throughout the codebase.

I don't feel very strongly about this, just a mild sense that this isn't The Way.

packages/server/src/sdk/app/screens/screens.ts
```diff
@@ -1,13 +1,53 @@
...
+async function migrateToProjectApp(screens: Screen[]) {
+  const application = await sdk.applications.metadata.get()
+  if (screens.every(s => s.projectAppId)) {
+    return screens
+  }
+
+  const allProjectApps = await sdk.projectApps.fetch()
+  let projectAppId = allProjectApps.find(p => p.name === application.name)?._id
+  if (!projectAppId) {
+    // Forcing the _id to avoid concurrency issues
```
samwho, line 34: What concurrency issues?

packages/types/src/api/web/app/application.ts
```diff
@@ -38,7 +38,7 @@ export interface FetchAppDefinitionResponse {
 export interface FetchAppPackageResponse {
   application: App
   licenseType: PlanType
-  screens: Screen[]
+  projectApps: (ProjectApp & { screens: Screen[] })[]
```
samwho, line 41: Are screens not always a part of the `ProjectApp` type?

packages/server/src/api/controllers/projectApp.ts
adrinr: We don't really need it, but we used that pattern across all the endpoints and this just follows the current standard

packages/server/src/sdk/app/screens/screens.ts
adrinr: We could have parallel gets. This is solved using the built-in migrations

packages/server/src/sdk/app/screens/screens.ts
adrinr: You are right, I wanted to avoid having a migration that "blocks" the UI, but it's probably better than having a migration living next to the code

packages/server/src/api/controllers/projectApp.ts
samwho: That's fair.

samwho (approved):

PR #16139, opened 2025-05-14
New screen flow as modal

adrinr:
## Description
New screen flow as modal.

On the wizard flow navigation:
- Either clicking the back button or pressing the escape key, go to the previous state
- Either clicking outside the modals or on the X on the top, closing the full flow

https://github.com/user-attachments/assets/ed9e22e3-cce0-430e-9dad-6085d1093986

## Feature branch env
[Feature Branch Link](http://fb-budi-9296-new-screen-as-modal.budibase.dev)


aptkingston (changes requested):
Great work Adri 👌 

The main issue I found is that when you have no screens the builder still redirects you to `/design/new` which has no content on it, and you get stuck in a blank page without the ability to create a screen:
image


On other super minor thing is that when PDF is disabled the Form template is 2.5 pixels shorter than PDF, but the selected border is applied to the outer div which results in the border being too tall:
image

We can either change from grid layout to flex layout to fix this, or change where the selected border is applied, or apply a negative margin to the "Premium" badge to fix the height offset.

aptkingston (commented):
I like the idea of the CSS hack - but you'll need to update it to get it to work.

You'll need to change the CSS to:
```
  display: flex;
  justify-content: center;
  align-items: center;
  flex: 1 1 auto;
```
in the `new-screen-picker` class.

Before:
image

After:
image

aptkingston (approved):
LGTM!

PR #16147, opened 2025-05-14
Fix query parameter binding when passing a 0, it shouldn't fall back to the default value.

samwho:
## Description

Query params were being considered falsey if they were the number 0, and the default value being substituted in their place. This PR corrects the offending if statement and adds a test for this case.

## Addresses
- https://linear.app/budibase/issue/BUDI-9189/passing-0-into-query-using-js-returns-default-value

## Launchcontrol

Fixes a case where query bindings were falling back to their default value if the value given was a 0.

### DIFF
--- a/packages/server/src/api/controllers/query/index.ts
+++ b/packages/server/src/api/controllers/query/index.ts
@@ -169,8 +169,10 @@ function enrichParameters(
   validateQueryInputs(requestParameters)
   // make sure parameters are fully enriched with defaults
   for (const parameter of query.parameters) {
-    let value: string | null =
-      requestParameters[parameter.name] || parameter.default
+    let value = requestParameters[parameter.name]
+    if (value == null) {
+      value = parameter.default
+    }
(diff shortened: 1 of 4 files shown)


packages/server/src/api/controllers/query/index.ts
```diff
@@ -169,8 +169,10 @@ function enrichParameters(
   validateQueryInputs(requestParameters)
   // make sure parameters are fully enriched with defaults
   for (const parameter of query.parameters) {
-    let value: string | null =
-      requestParameters[parameter.name] || parameter.default
```
melohagan, line 173: Can you use `??` here instead?

melohagan (approved):
LGTM!

packages/server/src/api/controllers/query/index.ts
samwho: You probably could now I think of it. Not going to go in and change, but a good spot.

PR #16156, opened 2025-05-15
stringify object values

shogunpurple:
## Description
Fixes an issue where object values in the temporary/debug state variable input were causing crashes

## Addresses
https://linear.app/budibase/issue/BUDI-9299/state-variable-temporary-value-resets-to-object-object-and

### DIFF
--- a/packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/StatePanel.svelte
+++ b/packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/StatePanel.svelte
@@ -56,7 +56,13 @@
   ) => {
     if (screen && key) {
       searchComponents(screen, key)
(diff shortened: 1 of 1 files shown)


aptkingston (commented):
Since we're checking if the type is `object` first, that would mean that if the real state value was `null` or `undefined` it would stringify that and put the literal string `"null"` or `"undefined"` into the editor, which would in turn update state to actually set it to those string values wouldn't it?

Should the order maybe be
- Check if nullish (use empty string as editor value)
- Check if type is `string` (use it directly)
- Otherwise stringify

Unrelated to your changes but this would also fall over if the real state values were bools or any sort of more complicated JSON objects. But I guess that was never intended to be supported. I reckon if we at least account for `undefined` since that will definitely appear in real apps.

Issue #16082, opened 2025-05-06
Possible bug in Data Provider ?


poirazis:
I think the Data Provider https://github.com/Budibase/budibase/blob/master/packages/client/src/components/app/DataProvider.svelte needs an onDestroy hander to clear any pending timers.

I think the current implementation leaves pending timers behind.

mikesealey:
Hey @poirazis, could you provide some more details on this? Perhaps some steps to reproduce the scenario you're describing and/or an export

poirazis:
Sure
Make a new screen
Add a data Provider
Set to Autorefresh every 10s

Create a second screen

Preview the app

Navigate to the screen with the data provider 
Then navigate away to the other screen

Open Dev Tools / Network , you will see the Data Provider still pulling data

I think what is missing is 

`onDestroy(() => {
    if (interval) clearInterval(interval);
  });`

PR #16159, opened 2025-05-15
adds onDestroy handler for auto refresh Data Prov

mikesealey:
## Description
If a DataProvider is set to Auto-refresh, the timers don't get removed/cancelled when the user navigates away to a different page. This can be demonstrated with a 10-second interval, navigating elsewhere in the app. Looking at the browser's network tab you'll see the request fires even after leaving the page.

## Addresses
- https://github.com/Budibase/budibase/issues/16082

## App Export
[BUDI 9297-export-1747321455930.tar.gz](https://github.com/user-attachments/files/20230539/BUDI.9297-export-1747321455930.tar.gz)

## Screenshots
Check the Network tab in Jam:
https://jam.dev/c/6c61d570-ceea-4e4b-ae2f-33a345f77f52

## Launchcontrol

_Removes AutoRefresh timers on Data Providers after navigating away_

### DIFF
--- a/packages/client/src/components/app/DataProvider.svelte
+++ b/packages/client/src/components/app/DataProvider.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { getContext } from "svelte"
+  import { getContext, onDestroy } from "svelte"
   import { Pagination, ProgressCircle } from "@budibase/bbui"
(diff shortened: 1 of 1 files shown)


aptkingston (approved):
Nice one! Really good catch.

PR #16165, opened 2025-05-16
Always fetch a new presigned url for attachments

PClmnt:
## Description
Fixes an issue where `url` was populated in for an attachment. Therefore a new signed URL was never generated.


packages/server/src/utilities/rowProcessor/tests/outputProcessing.spec.ts
PClmnt, line 1: For some reason, these tests were assuming that a URL existed in the object before. Thoughts @adrinr Safe to remove?

packages/server/src/utilities/rowProcessor/tests/outputProcessing.spec.ts
```diff
@@ -177,10 +177,6 @@ describe("rowProcessor - outputProcessing", () => {
       expect(output2.attach[0].url?.split("?")[0]).toBe(
         "/files/signed/prod-budi-app-assets/test.jpg"
       )
-
-      row.attach[0].url = "aaaa"
```
adrinr, line 181: I think the right test here would be having the `url` but no `key`. If this ever happen (I don't really see how) we want that to be the behaviour

adrinr (approved):
LGTM!

PR #16176, opened 2025-05-19
Fix for resetOn behaviour

deanhannigan:
## Description
The `resetOn` behaviour was too aggressive when it came to altering a `dataSource` prop. In the attached ticket, when the `queryParams` were altered in the builder, the `resetOn` behaviour considered this reason enough to clear out the `filters` parameter. In reality, this should only trigger if the query is replaced with a completely one.

- Updated the component tests to include more coverage for `updateComponentSettings` in general.
- Added the entire `@budibase/client` manifest to dev dependencies to replace stale fixtures. This means that we have a much more accurate representation of the real client definitions. 
- Updated the mocks to allow setting stores that are consistent throughout the test run.
- This fix takes into account all datasource types but focuses mainly on queries.

## Addresses
- https://github.com/Budibase/budibase/issues/16124

## Launchcontrol
Update 'resetOn' behaviour to only trigger when the datasource has been completely replaced.


packages/builder/src/stores/builder/components.ts
```diff
@@ -1085,6 +1085,40 @@ export class ComponentStore extends BudiStore<ComponentState> {
...
+      viewNameChange = currentSetting.name !== update.name
+    }
+
+    return (
+      !currentSetting ||
+      currentSetting.type !== update.type ||
+      currentSetting?.[resourceId] !== update[resourceId] ||
+      viewNameChange
+    )
+  }
```
aptkingston, line 1120: Could we simplify this right down since it's only queries that have these extra params? I'm basically thinking that if the newly selected datasource type is anything other than `query` we can immediately return true that it changed. If the new type is `query`, compare the type and `_id` property with the previous value.

aptkingston (commented):
Great work on fleshing out the tests 👌 Just left a question around if we can simplify this - but happy to merge either way.

packages/builder/src/stores/builder/components.ts
deanhannigan: Yep, agreed 👍  My first approach was exactly that; only catering to the `query._id` case as no other datasources allowed additional config. The only reason I went with this approach was future proofing. I'm happy to remove the other cases.

packages/builder/src/stores/builder/components.ts
```diff
@@ -1100,12 +1126,25 @@ export class ComponentStore extends BudiStore<ComponentState> {
...
       settings.forEach((setting: ComponentSetting) => {
         const needsReset =
           name === setting.resetOn ||
           (Array.isArray(setting.resetOn) && setting.resetOn.includes(name))
+
         if (needsReset) {
+          // Ignore the reset if the updated datasource was not replaced
+          if (isDatasource && !sourceModified) {
+            return
+          }
```
aptkingston, line 1147: NAB: we can hoist this right up to just above where we compute `isDatasource` and `sourceModified is computed, since if this is a datasource setting and there's no change we want to just abort entirely.

In fact, the snippet up above (line 1130) could effectively just be:
```js
if (updatedSetting?.type === "dataSource") {
  if (!this.isDatasourceUpdated(component, updatedSetting, value)) {
    return false
  }
}
```

Just no need to have this logic down in the loop, since if the DS didn't change this should just be a NOP.

aptkingston (approved):
LGTM! Not a blocker, but one small suggested refactor to simplify again.

packages/builder/src/stores/builder/components.ts
deanhannigan: Yo, thanks for taking a look! I ran through that scenario and if you bail out at that point, the update on line 1169 will be skipped entirely. In this instance we aren't trying to abort the update, just any side-affects of the resetOn if they haven't been met.

PR #16181, opened 2025-05-21
Retrieve only screens for workspace apps

adrinr:
## Description
Updating the way we retrieve screens for a given app. The changes are only on the backend.
These changes will update the endpoint to allow passing the loaded routeurl in order to retrieve what workspace app is being loaded


packages/server/src/api/controllers/application.ts
```diff
@@ -271,13 +272,34 @@ export async function fetchAppPackage(
...
+  // Only filter screens if the user is not a builder call
+  const isBuilder = users.isBuilder(ctx.user, appId) && !utils.isClient(ctx)
+  if (!isBuilder) {
     const userRoleId = getUserRoleId(ctx)
     const accessController = new roles.AccessController()
     screens = await accessController.checkScreensAccess(screens, userRoleId)
   }
 
+  if (await features.flags.isEnabled(FeatureFlag.WORKSPACE_APPS)) {
+    const fromRoute = ctx.request.query.fromRoute
```
aptkingston, line 284: Is there a reason we can't use the real HTTP referrer? Since that would be a bit more robust and not (as easily) forged.

adrinr: Rethinking this, this makes more sense. We will have different apps with URL prefixes, the hash part will be to navigate within that app screens. This is more natural, and it works out of the box without passing this values!

aptkingston (approved):
LGTM!

PR #16186, opened 2025-05-21
Font tweaks

aptkingston:
## Description
This PR adds a few CSS tweaks to improve compatibility with Source Sans 3.

Changes:
- Adjusted button padding
- Adjusted text input padding
- Inject latest font CSS into server-rendered HTML to bypass client versioning and ensure Source Sans 3 works in old apps

The main font fix here **does not** require an app update, which is the whole point of it. It immediately fixes the font in old client apps as soon as the platform is updated, but does not need an app update.

Before (notice the font is wrong, as it is incorrectly attempting to use Source Sans Pro as it is using versioned CSS):
image

After:
image

### DIFF
--- a/packages/bbui/src/Button/Button.svelte
+++ b/packages/bbui/src/Button/Button.svelte
@@ -44,6 +44,7 @@
       }
     }}
   >
+    sdfdsf
     {#if $$slots && reverse}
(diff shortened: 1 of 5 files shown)


deanhannigan (approved):
LGTM!

PR #16191, opened 2025-05-22
Improve handling of app versions in dev mode

aptkingston:
## Description
This PR fixes a few issues with versioning client apps in dev. A while ago we added a feature to be able to use old client versions in dev, but this was buggy and currently the handling of app versioning in dev mode is broken. It currently says updates are available, but trying to update errors. This doesn't make sense, as in dev you should always be using the local version from your file system.

This PR prevents updating or reverting client apps in dev. The associated UI components are hidden, and the settings will now explicitly state that you're using the latest version from the file system since you're in dev.

Settings when in dev mode:
image

I've also removed all code relating to that feature of fetching and using old client app versions, so the update modal is more minimal again. When only update is available:
image

When update and revert are available:
image

## Feature branch env
[Feature Branch Link](http://fb-dev-client-fix.budibase.dev)


deanhannigan (approved):
LGTM! Span it up locally and didn't run into any issues 🎉

adrinr (approved):
LGTM

PR #16194, opened 2025-05-23
Charts on click

mikesealey:
## Description
This PR gives all Charts on-click actions, which can be used to surface data or events. 

#### Area Chart, Bar Chart, Candlestick Chart, Line Chart
* Clicked row - The row responsible for the datapoint clicked. 
image

#### Pie Chart, Donut Chart
* Clicked Row - the row responsible for the datapoint clicked. 
* Segment Index
* Percentage of the pie/donut that this datapoint occupies
image

#### Histogram Chart
* Clicked bucket rows - that fall into the clicked bucket
* Lower Limit
* Upper Limit
image

## App Export
[Charts-export-1749561271411.tar.gz](https://github.com/user-attachments/files/20673407/Charts-export-1749561271411.tar.gz)

## Screenshots
Demonstration of published results: https://jam.dev/c/20fd23ad-e645-4902-8709-13c222c97fb3

Bindings drawer:
image

## Launchcontrol

_Adds on-click actions for charts_


packages/client/src/components/app/charts/AreaChart.svelte
```diff
@@ -153,4 +168,4 @@
   }
 </script>
 
-<ApexChart {options} />
+<ApexChart {options} on:click{handleLineClick} />
```
aptkingston: Does this handler ever do anything? I don't think the apex chart component broadcasts a click event. If it did, this would also then break because the params passed to the callback are not what `handleLineClick` expects. We should remove this `on:click` from the ApexChart component in every file here that it's been added to.

aptkingston (commented):
LGTM other than removing the on:click props as mentioned. Nice work!

aptkingston (approved):

PR #16200, opened 2025-05-26
Builder nav updates

aptkingston:
## Description
This PR converts the primary builder top nav into a collapsible side nav. All previous functionality is still present, as well as some new functionality. This sits behind the `WORKSPACE_APPS` feature flag.

image

The only new addition is the user avatar, which brings in the settings available when currently only browsing the portal (or a subset in the end user app):
image

The side nav is open by default, but can be collapsed at any time using the icon in the top right:
image

Collapsed side nav:
image

When collapsed, the nav can be expanded again by hovering over it. It can be permanently opened again by clicking the same icon used to collapse it.

https://github.com/user-attachments/assets/58f17a35-22f7-4454-94aa-470b8ef5a1e5

## Feature branch env
[Feature Branch Link](http://fb-nav-updates.budibase.dev)


deanhannigan (approved):
LGTM

PR #16202, opened 2025-05-27
[BUDI-9179] Fix "duplicate key violation" errors when updating many-to-many fields.

samwho:
## Description

This was a fun (read: not fun) one to debug.

There were 2 problems:

1. When building up `ManyRelationship` objects, we were not using the correct fields for many-to-many relationships.
2. When searching for existing relationship rows, the predicate was not correctly matching join rows for many-to-many relationships.

However, things worked just fine for many-to-many relationships created inside of Budibase because of the way that we name fields. The wrong field we were using in 1 happened to line up correctly for Budibase-created relationships, and wasn't causing a problem. For 2, the reason that the predicate wasn't matching was because of the `id` field in the join table being present on existing relationships but not necessarily on new ones coming in from the API (even though "new" ones could be existing relationships.)

## Addresses
- https://linear.app/budibase/issue/BUDI-9179/duplicate-key-value-violates-unique-constraint-errors-on-many-to-many

## Launchcontrol

Fixes an error that crops up when you attempt to update relationships in many-to-many tables that have been imported into Budibase from an external source.


mike12345567 (approved):
LGTM - nice work, there's quite a few niggly changes in here that were required!

PR #16203, opened 2025-05-27
Navigation per workspace app

adrinr:
## Description
Implementing navigation per workspace app, both in the builder and in the client. Also, updating the app panel to include a "view live" link: 
image

Includes the already approved PR, merged to be able to run a proper final QA:
- https://github.com/Budibase/budibase/pull/16221

 

## Feature branch env
[Feature Branch Link](http://fb-budi-9294-navigation-per-project-app.budibase.dev)


samwho (commented):
Have only looked at the backend changes for the time being.

packages/server/src/api/controllers/application.ts
```diff
@@ -188,34 +189,47 @@ async function addSampleDataScreen() {
...
 
-  await db.put(app)
+    await db.put(app)
 
-  // remove any cached metadata, so that it will be updated
-  await cache.app.invalidateAppMetadata(app.appId)
+    // remove any cached metadata, so that it will be updated
+    await cache.app.invalidateAppMetadata(app.appId)
+  } else {
+    const workspaceApp = (await sdk.workspaceApps.get(screen.workspaceAppId!))!
```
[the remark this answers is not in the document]
adrinr: This will be always true given the creation flow. I refactored the code to do both app and link creation at once, making it clearer and cleaner

samwho (approved):

aptkingston (commented):
Nice work Adri! Everything performs as expected, including the migration of old apps 👌 

Found one small design issue which is that the home screen icons should be right aligned, but they sit quite far out (this is probably from a previous PR but i've just noticed it):
image

The only other UX thing I found (which we might not want to fix) is that navigating to the root URL of a workspace doesn't redirect you to the first workspace app, assuming they all have prefix URLs. Not sure if we even want to change that though.

One style change which might also be nice with workspace apps is increasing the default size of the screen list, as it's currently very small by default compared to the component list. To do that we just need to adjust the min and max height css in `ScreenList/index.svelte`. I like something around 320px:

```css
.screens {
    min-height: 320px;
    max-height: calc(100% - 320px);
    ...
}
```

aptkingston:
Again probably caused by a previous workspace app PR, but it seems like the routing needs updated so that if you land on `/design` it'll properly redirect you to the first workspace app. Right now it pushes you to the `/new` URL, and means there's no way of escaping that without either adding a new screen or manually knowing and typing the workspace app URL prefix.

packages/builder/src/stores/builder/navigation.ts
```diff
@@ -29,10 +58,18 @@ export class NavigationStore extends BudiStore<AppNavigationStore> {
   }
 
   async save(navigation: AppNavigation) {
-    const appId = get(appStore).appId
-    const app = await API.saveAppMetadata(appId, { navigation })
-    if (app.navigation) {
-      this.syncAppNavigation(app.navigation)
+    if (!featureFlag.isEnabled(FeatureFlag.WORKSPACE_APPS)) {
```
samwho, line 61: Maybe a silly question but how come this doesn't use `$featureFlags` like above in the constructor?

packages/builder/src/stores/builder/tests/screens.test.js
```diff
@@ -281,88 +285,16 @@ describe("Screens store", () => {
...
 
-    expect(routeSpy).toHaveBeenCalled()
+    expect(appStore.refresh).toHaveBeenCalledOnce()
     expect(saveSpy).toHaveBeenCalled()
 
     // On save, the screen is spliced back into the store with the saved content
     expect(bb.store.screens[2]).toStrictEqual(existingScreens[2].json())
   })
 
-  it("Sync API data to relevant stores on save. Updated plugins", async ({
```
samwho, line 298: How come this is being removed?

samwho (approved):

packages/builder/src/stores/builder/navigation.ts
adrinr: `$featureFlags` is a store, so it's not accessible. We could use `get(featureFlags)` instead, but it would be the same as reactivity would not apply either

packages/builder/src/stores/builder/tests/screens.test.js
adrinr: This function is not called anymore, as we to a hard refresh on the appStore (already tested on another test)

aptkingston (approved):
LGTM! Awesome work Adri. Gave this a good test locally and it works well 👍 

Edit - it looks like new apps don't have any nav items. I think they currently ship with one nav item for the home screen?

Prod:
image

This PR:
image

Not a big deal though!

Issue #16080, opened 2025-05-06
Enhance the navigation settings area and include some more features


ConorWebb96:
**Describe the feature request**
Having more control over the navigation area would be nice as a user.

Some settings that would be great are:

- Logo position: top or bottom of the nav bar.
- Title formatting: size, colour
- Background colour, border, shadow

Add more settings for Navbar links formatting: size, colour, and styling.

zedascouves:
I'd like to add more:

- Ability to add a datasource. This would enable, for example, showing counts for pages containing lists.

PR #16204, opened 2025-05-27
16080 enhance the navigation settings area and include some more features

ConorWebb96:
## Description
Adds a range of different improvements to navigation settings.

- Adds the ability to supply custom styles to each navigation item
- Adds the ability to supply conditional statements to each navigation item. 
- Add a few different settings to the navigation area
   - Title colour
   - Title size
   - Logo position (Top or bottom) only when its set to the side

## Addresses
- https://github.com/Budibase/budibase/issues/16080

## Screenshots
**Navigation settings before**
Screenshot 2025-05-27 at 10 30 42

**Navigation settings after**
Screenshot 2025-05-27 at 10 32 57

**Navitem popover before**
Screenshot 2025-05-27 at 10 39 27

**Navitem popover after**
Screenshot 2025-05-27 at 10 51 13

## Launchcontrol
Adds more functionality to the navigation settings area for screens, allows users to have more control over: title size, colour, navigation custom styles, conditions and logo positioning.


packages/client/src/components/app/Layout.svelte
```diff
@@ -284,22 +304,38 @@
...
+                  <Logo
+                    {logoUrl}
+                    {logoLinkUrl}
+                    {openLogoLinkInNewTab}
+                    {hideLogo}
+                    {title}
+                    {linkable}
+                    {isInternal}
+                    {getSanitizedUrl}
+                    {logoHeight}
```
aptkingston: We don't need to pass in logo height here, as we already set a CSS variable which controls height and defaults to 24px. 
- Remove passing in logo height here
- Remove the logo height prop from Logo
- Remove all `style` attributes in Logo
- Add CSS to Logo to use the predefined variable:
```html
<style>
  img {
    height: var(--logoHeight);
  }
</style>
```

Without this, existing apps don't have any size set:
image

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Navigation/index.svelte
```diff
@@ -131,25 +175,26 @@
...
+          />
+          <PropertyControl
+            label="Title color"
+            control={ColorPicker}
+            onChange={color => update("titleColor", color)}
+            value={$nav.titleColor}
+            props={{
+              spectrumTheme: $themeStore.theme,
+            }}
+          />
```
Comment (aptkingston) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Navigation/index.svelte:196]: I would strongly prefer to not have a separate prop for title color. We already have a "text color" setting for the navigation that sets the text color - I don't think we need a dedicated one so that you can have different colour links from the title... Unless this was specifically part of the request I guess.

packages/bbui/src/Typography/Heading.svelte
```diff
@@ -6,11 +6,17 @@
...
 </script>
 
 <h1
   data-testid="typography-heading"
-  style={textAlign ? `text-align:${textAlign}` : ``}
+  style={[
+    textAlign ? `text-align: ${textAlign}` : "",
+    titleColor ? `color: ${titleColor}` : "",
+  ]
+    .filter(Boolean)
```
aptkingston, line 18: Sneaky little trick. Very terse, I like it. Although if we can remove the title colour setting then we can also undo changes to this file, but I'll let that be informed by whether or not a unique title colour was part of the customer request. If we're keeping this, then we at least need to rename this setting to `color` rather than `titleColor`.

aptkingston (commented):
Sorry I forgot about this one for so long! Left a couple of comments. In general good job on the implementation - works well.

Only other point:
I'm not sure I understand the purpose of the logo position setting - it only moves the logo but the title remains in place. So you end up with something like this, with an unusal logo at the bottom and title at the top:
image

Maybe this was also part of the customer request but it looks very unnatural to me.

packages/client/src/components/app/Layout.svelte
ConorWebb96: Thanks for the feedback.

packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Navigation/index.svelte
Comment (ConorWebb96) [reply packages/builder/src/pages/builder/app/[application]/design/[screenId]/[componentId]/_components/Navigation/index.svelte:196]: I completely agree with you; however, this was part of the initial request. It mainly gives users more flexibility. That said, it does open up the possibility for colour clashes. I think many users won't change the settings and will stick with the default text colour.

packages/bbui/src/Typography/Heading.svelte
ConorWebb96: As it was part of the customer request, I've renamed it to just color.

packages/client/src/components/app/Layout.svelte
```diff
@@ -59,7 +59,7 @@
 
   export let logoPosition = "top" // "top" or "bottom"
   export let titleSize = "S"
-  export let titleColor // CSS color string, only affects title
+  export let color // CSS color string, only affects title
```
aptkingston: I think this was an accidental change - this should remain as `titleColor`. Only the BBUI prop should be called `color`. If you wanted to rename this one you'd also need to update the navigation settings component which saves it, since it currently is still using `titleColor`, so right now this wouldn't actually work.

Best course of action - just renaming this back to `titleColor`, but keeping the BBUI Heading component using `color`.

aptkingston (approved):
LGTM!

PR #16207, opened 2025-05-27
Use form context instead of component type to find nearest form component

### LINKED ISSUE #16198: [ Custom Components ] - Form Context and Properties enrichement
I have created a custom Form component
Its the exact replica of the build in Form / InnerForm

The component exposes context type "form", as per stock Form

When you drop ( custom or stock ) field components inside my custom form, the field properties of the child components dont get enriched, as if the form field enrichement pipeline doent take into account custom components.

The Form is still properly exposed into the various Form actions.

https://github.com/user-attachments/assets/8b451521-f091-4737-9c28-2b5a49024ced

aptkingston:
## Description
This enhancement improves support for custom forms. Currently we populate "form field" settings by looking for the nearest form component and extracting the schema it uses. Today that uses an exact match on component type, only searching for our own form component.

This PR updates that logic to instead locate the nearest component that provides form context, which means custom forms are supported.

Quick example with a form plugin, using a text field inside https://github.com/poirazis/bb-component-SuperFormSimple:

Form component using my "Data types" schema:
image

Text field component correctly listing text types:
image

## Addresses
- Closes https://github.com/Budibase/budibase/issues/16198


deanhannigan (approved):
LGTM! Nice one! Ran through this locally and compared it against the behaviour on master. Didn't run into any issues 🎉

PR #16208, opened 2025-05-27
New API Request Step

deanhannigan:
## Description
Added a new automation step type, specifically for integration API queries into your automation! 

We currently support this behaviour as part of the `External Data Connector` but API requests have now been broken out into their own dedicated step.

No action is required for any existing `External Data Connector` steps as these will continue to work as configured. From now on though, any API requests should be configured in the new `API Request` step.

## Addresses
- https://linear.app/budibase/issue/BUDI-9319/update-external-data-connector-automation-step
- `Bug` - evaluation context fix for the `External Data Connector` step. The evaluate button now appears in the binding drawer as expected
- `Update` - General Looping config has been shifted into its own collapsable section at the top of the step sidebar

## Screenshots
New step layout
Screenshot 2025-05-27 at 15 02 55

The new `Query Details` section gives a quick summary of your request at a glance. With `Auth`, `Header`, `Request`, `Body` and `Transformer` sections.
Screenshot 2025-05-27 at 15 06 45

## Launchcontrol
New APIRequest Step added to automations.


packages/builder/src/components/automation/AutomationBuilder/StepPanel.svelte
```diff
@@ -55,6 +63,11 @@
   $: isAppAction &&
     automationStore.actions.setPermissions(role, $memoAutomation)
 
+  // Reset the panel scroll when the target node is changed
+  $: if (configPanel && configPanel?.scrollTop > 0 && selectedNodeId) {
```
adrinr, line 67: Will this not evaluate every time scrollTop changes?

packages/builder/src/components/automation/SetupPanel/layouts/APIRequest.svelte
```diff
@@ -0,0 +1,311 @@
...
+      return !query?.fields?.disabledHeaders?.[key]
+    })
+    .map(([key, val]: [string, string]) => {
+      return [key, runtimeToReadableBinding(restBindings, val)]
+    })
+  $: parsedHeaders = Object.fromEntries(parsedHeaderEntries)
+
+  const getRESTPreview = (query: Query | undefined) => {
+    if (!query) return
+    return query?.fields?.queryString
```
adrinr, line 105: NIT: extracting the `{queryString, path} = query.fields` might make this more readable

packages/builder/src/stores/builder/queries.ts
```diff
@@ -26,6 +26,7 @@ interface BuilderQueryStore {
 
 interface DerivedQueryStore extends BuilderQueryStore {
   selected?: Query
+  listByType?: Query[]
```
adrinr, line 29: Is this used? I cannot see it being used anywhere

deanhannigan: Nope, nice catch! 👍

aptkingston (commented):
This looks so much better Dean, nice one. Proper first class experience for API requests in automations now! Haven't looked at the code yet but was giving this a rinse locally and these were the first few things I spotted (only one being an actual bug):

- I created an example PATCH request with everything filled in - headers, bindings, params, body etc, and it causes the builder to lock up when selecting it (just hard crashes, unrecoverable). I think it's maybe something to do with CodeMirror? Will send you the app privately.
- NAB: could we use the common `NewPill` component so that we get the usual purple "new" badge instead of the blue one, beside the new API request step title?
- NAB: when you have some queries but haven't selected one, the ability to add a new one is disabled. You have to pick an existing query in order to be able to create a new one, which feels like a restriction we don't need
image
- NAB: We call the request "Request" in the picker, but the section below it is called "Query details". Might be nice to standardise that to one term.

aptkingston (approved):
All feedback addressed, LGTM!

PR #16213, opened 2025-05-28
Remove client dep from builder

deanhannigan:
## Description
Added the @budibase/client as a dev dependency on the builder and this seems to cause a significant memory leak.

Original PR: https://github.com/Budibase/budibase/pull/16176

## Launchcontrol
Fix for dev memory leak


deanhannigan:
This issue was resolved here: https://github.com/Budibase/budibase/pull/16214

PR #16216, opened 2025-05-28
Strip user information if the user does not need it

mike12345567:


packages/types/src/documents/global/user.ts
```diff
@@ -75,6 +75,12 @@ export interface User extends Document {
   tours?: Record<string, Date> // deprecated and no longer saved
 }
 
+export interface StrippedUser extends Document {
```
samwho, line 78: Should this instead be:

```typescript
export type StrippedUser = Pick<User, "_id" | "tenantId" | "email" | "userId">
```

samwho (approved):

PR #16218, opened 2025-05-29
Spectrum to phosphor icon migration

andz-bb:
## Description
This PR migrates all icons from spectrum and font awesome to phosphor. It also unifies all icon usage to properly consume the BBUI icon component, rather than the multitude of uses of raw spectrum SVG components, which will make maintenance far easier.

Changes:
- Removed spectrum icons entirely, including all CSS targeting them
- Replace all usage of custom SVG icons with BBUI icon component
- Removed font awesome
- Added phosphor icon equivalents globally
- Left in a map of spectrum > phosphor icon names as a fallback, so any missed icons or icons saved outside our codebase (e.g. old automation step definitions saved in the DB, or old client libs) will still render properly

Notes:
- All icons are slightly differently sized now. We will likely need many updates of icon sizes across the board until this is ready to be shipped, but everything should be functional and look correct other than sizing being slightly off.

## Launchcontrol
Migrated from Spectrum icons to Phosphor icons.

## Feature branch env
[Feature Branch Link](http://fb-phosphor-3.budibase.dev)


deanhannigan (approved):
LGTM!

PR #16221, opened 2025-05-29
Handle screen conflicts between workspace apps

adrinr:
## Description
Allowing "duplicated" settings between workspace apps. Given this new setup, we can allow things such as:
- duplicated screen names between apps
- multiple home pages (while being in different apps)

Some of this logic has been moved from the frontend, now living on the backend

### DIFF
--- a/packages/server/src/appMigrations/migrations/20250514133719_workspace_apps.ts
+++ b/packages/server/src/appMigrations/migrations/20250514133719_workspace_apps.ts
@@ -29,9 +29,14 @@ const migration = async () => {
     }))
 
   const designDoc = await db.get<DesignDocument>("_design/database")
-  // Remove the routing view to be recreated (and updated to support workspace apps)
-  delete designDoc.views?.[ViewName.ROUTING]
-  docsToUpdate.push(designDoc)
+  if (
+    designDoc.views?.[ViewName.ROUTING] &&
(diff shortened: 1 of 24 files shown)


samwho (approved):

packages/server/src/appMigrations/migrations/20250514133719_workspace_apps.ts
```diff
@@ -21,14 +21,24 @@ const migration = async () => {
...
+    .filter(s => !s.workspaceAppId)
+    .map<Screen>(s => ({
+      ...s,
+      workspaceAppId,
+    }))
+
+  const designDoc = await db.get<DesignDocument>("_design/database")
+  if (
+    designDoc.views?.[ViewName.ROUTING] &&
+    designDoc.views[ViewName.ROUTING].version !== 2
```
samwho: We probably need to be a lil' careful here, what happens if version === 3? Will it be correct to trigger this block?

samwho (approved):

packages/server/src/appMigrations/migrations/20250514133719_workspace_apps.ts
adrinr: Good thinking. I'll be more explicit. Even so, it should be safe in this case, as the view will be just recreated

PR #16226, opened 2025-05-29
Respect what's in values.yaml when updating secrets through the Helm chart.

samwho:
## Description

Up until this PR, if you tried to change the value of e.g. `internalApiKey` it wouldn't work. The Helm chart is written in such a way that secret values prefer the current value to anything found in values.yaml, so weren't updateable after creation.

This PR changes that such that what's in values.yaml is the preference, allowing you to change the value of secrets through values.yaml.

## Launchcontrol

Makes it possible to update secret values through the Helm chart.

### DIFF
--- a/charts/budibase/templates/secrets.yaml
+++ b/charts/budibase/templates/secrets.yaml
@@ -12,12 +12,12 @@ metadata:
 type: Opaque
 data:
   {{- if $existingSecret }}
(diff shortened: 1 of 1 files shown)


calexiou (approved):

samwho:
We didn't need this in the end and it's quite a scary change to make, so I'm going to close it for the time being.

PR #16230, opened 2025-05-30
AI Data extraction automation step

PClmnt:
## Description
This features allows a user to pass an image or a file alongside a Schema to our LLM, and in return receive some structured data. 

This supports both
- Passing a file from our object storage, say for example an attachment in a row in your db, e.g coming from a `Query Rows` step. 
- Passing a generic URL, for this you need to specify your content type. 

I also added new logic to the automation schema to allow the showing / hiding of a field based on the value of another field

## Screenshots
image

You can see here how an array of objects based on the schema we have provided is returned. This is using an attachment from a row
image

And the same again but using a URL. 
image

Pro PR: https://github.com/Budibase/budibase-pro/pull/419

## Feature branch env
[Feature Branch Link](http://fb-feat-extraction-automation-step.budibase.dev)


packages/server/src/automations/steps/ai/extract.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  try {
+    const llm = await ai.getLLMOrThrow()
+
+    let filename
+    let contentType
+    let fileIdOrDataUrl
+    let request
+
+    if (
```
adrinr, line 34: For readability and verbosity, on long functions like this would be good to break some logic into subfunctions

packages/shared-core/src/automations/steps/ai/extract.ts
```diff
@@ -0,0 +1,66 @@
...
+          enum: Object.values(DocumentSourceType),
+          title: "Source",
+          description: "The source of the document to extract data from.",
+        },
+        fileType: {
+          dependsOn: { field: "source", value: "URL" },
+          type: AutomationIOType.STRING,
+          enum: Object.values(SupportedFileType),
+          title: "File Type",
+          description: "The type of file at the URL",
```
adrinr, line 40: NIT: either all or none descriptions should finish with a dot

packages/types/src/documents/app/automation/StepInputsOutputs.ts
```diff
@@ -219,7 +219,18 @@ export type GenerateTextStepOutputs = {
...
+  file: RowAttachment | string
+  source: "URL" | "Attachment"
+  fileType?: string
+  schema: Record<string, any>
+}
 
+export type ExtractFileDataStepOutputs = {
+  response?: string
+  success: boolean
+  data?: Record<string, any>
```
adrinr, line 232: Should we use smarter types such as `{ success: false } || { success: true, data: Record<string, any>}`, making data required for successful calls?

adrinr (approved):
Minor comments, but LGTM!

Issue #16219, opened 2025-05-29
Add option to customise upload text in the attachment component


ConorWebb96:
**Describe the feature request**
As a user, I would like to customise all the text in the attachment component, including the upload prompt. Currently, all other text elements seem customisable except this one. This would be especially helpful when building apps for users in different languages.

**Why is this important?**
- Enables localisation/multilingual support
- Provides more control over UI consistency
- Improves accessibility and user experience for diverse audiences

**Screenshots**
<img width="607" alt="Image" src="https://github.com/user-attachments/assets/f61d5001-d7dd-4bb8-80bf-9db3c57ac13c" />

PR #16234, opened 2025-05-30
16219 add option to customise upload text in the attachment component

ConorWebb96:
## Description
As requested by sales, a few fields were added to allow more customisation over text in a couple of components.

- Single Attachment
- Attachment list
- New filter component

## Addresses
- https://github.com/Budibase/budibase/issues/16219

## Screenshots
**Attachment setting changes**
Screenshot 2025-05-30 at 12 30 56

**Filter setting changes**
Screenshot 2025-05-30 at 12 35 19

## Launchcontrol
Sales requested some changes to a few components. This PR adds new settings to attachments and the filter component, allowing more text to be customised.


packages/client/manifest.json
```diff
@@ -4919,6 +4919,51 @@
...
+      },
+      {
+        "section": true,
+        "name": "Uploader Text Settings",
+        "settings": [
+          {
+            "type": "text",
+            "label": "Title Text",
+            "key": "titleText",
+            "dependsOn": [
```
deanhannigan: Nice detail 👨‍🍳 . In this instance you could add a single `dependsOn` section to your `Uploader Text Settings` for the upload fields. Then you don't need one for each field. You can then keep the `compact` check for `Title Text`

```json        
"dependsOn": [
  {
    "setting": "disabled",
    "value": false
  }
]
```

packages/client/src/components/app/filter/FilterPopover.svelte
```diff
@@ -360,7 +361,7 @@
             hide()
           }}
         >
-          Apply
+          {buttonText}
```
deanhannigan: You would need to ensure that this can't be empty. So add in a fallback to `Apply` to match the placeholder

packages/client/manifest.json
```diff
@@ -5668,6 +5758,12 @@
         "defaultValue": "M",
         "wide": true
       },
+      {
+        "type": "text",
+        "key": "buttonText",
+        "label": "Button text",
```
deanhannigan, line 5744: Add the following so the user can see what this will be if left blank.
```json
"placeholder": "Apply",
```

deanhannigan (changes requested):
Some really useful updates to the components! A few minor suggestions but great work!

deanhannigan (approved):
LGTM!

Issue #16241, opened 2025-06-02
User cannot hide table columns


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: miketest1

**Describe the bug**
No table columns can ever be hidden by the user.

**To Reproduce**
1. In the template app, preview
2. Click the three dots of a table column header
3. Hide column is always disabled

**Expected behavior**
Should be able to hide a column

**Screenshots**

![Image](https://github.com/user-attachments/assets/ef564b17-2ee4-45f2-b1d9-ca27a36811e6)

**Desktop (please complete the following information):**
 - OS: MacOS 14.6.1
 - Browser Chrome
 - Version 137.0.7151.56 (Official Build) (arm64)

PR #16242, opened 2025-06-02
Fixes condition checking if a column can be hidden

mikesealey:
## Description
"Hide column" was checking for a truthy value on a key that no longer exists. This PR removes that check.

## Addresses
https://github.com/Budibase/budibase/issues/16241

## App Export
N/A

## Screenshots
Previously: 
image

With this PR:
image

Display column remains un-hidable
image

## Launchcontrol

Fixes 'hide column' option in Table


aptkingston (approved):

aptkingston:
Might be worth removing the `showControls` prop from the GridBlock component in the client, as it still passes that in even though its not used.

PR #16243, opened 2025-06-02
Centralise `.env` for worker/server

mike12345567:
## Description
Today we have a separate `.env` for the worker and the server - this isn't really ideal as most variables need to be set for both. Useful when setting flags like `TENANT_FEATURE_FLAGS`.

This is specifically for development environments.


samwho (approved):

adrinr (commented):
Could you add this new path to the `nodemon` watch? 🙏🏽

PR #16247, opened 2025-06-02
Fix a bug around dates without timezones, unify output of date fields without a timezone.

samwho:
## Description

A while back we introduced stricter formatting requirements for the data we get for datetime fields. We required them to be in an ISO8601-type format, and only accepted UTC dates.

When you use a datetime field and set the "ignore timezone" option, the UI passes the backend datetimes with no timezone. These were always being rejected. This PR adds logic to accept strings without a timezone in this case.

I also noticed that the output format of datetime strings when using "ignore timezone" is not standardised, it changes based on what database you're reading from. I've unified them to always be `YYYY-MM-DDTHH:MM:SS.MMM`.

## Addresses

- https://linear.app/budibase/issue/BUDI-9188/timestamp-fields-no-longer-post-after-most-recent-update-insertupdate

## Launchcontrol

Fix a bug where data was being lost when writing to datetime fields if the value did not have a timezone, e.g. when using the "ignore timezone" setting.


aptkingston (approved):
LGTM!

PR #16252, opened 2025-06-03
Fixing memory issue and "redis" issue

adrinr:
## Description
Fixing an issue introduced in this [PR](https://github.com/Budibase/budibase/pull/16083) that causes the processes to not stop properly.
The issue is that the process was not proper existing, so both signals (SIGINT & SIGTERM) were being listen. The second to run would have issues closing the redis connections (as they would be already closed), causing an infinite loop because the `uncaughtException`/`unhandledRejection` handlers.
This scenario caused Redis to seem the culprit, while it was not their fault at all. Sorry, Redis.

Along the way I also updated the shutdowns to be proper async, hopefully helping us to catch unhandled errors in the future

### DIFF
--- a/packages/backend-core/src/events/index.ts
+++ b/packages/backend-core/src/events/index.ts
@@ -9,7 +9,7 @@ import { processors } from "./processors"
 
 export function initAsyncEvents() {}
 
-export const shutdown = () => {
-  processors.shutdown()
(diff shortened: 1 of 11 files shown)


mike12345567 (approved):
Very nice find - this could have gone un-noticed for a very long time!

PR #16255, opened 2025-06-03
Enable sourcemaps

adrinr:
## Description
While debugging some live issues I realised that the stack trace it's pretty useless, as we are getting the minified version:
image

Enabling sourcemaps when running with the node scripts should push the sourcemap on the stacktraces, allowing better debugging


adrinr:
Just worried that this might bring some performance issue. I will keep an eye to it when released

samwho (approved):
CleanShot 2025-06-03 at 12 54 51@2x

Good luck.

PR #16257, opened 2025-06-03
Google sheet improvements

adrinr:
## Description
Fixing the error handling while fetching google sheets. The errors were not handled properly, causing the frontend to break.
Also:
1. Fetching only the filtered sheets. This will improve some calls, and it will fix others. For example, I was getting an error because a non selected sheet had some invalid format. With this change, we don't even check their values
2. Closing the modal after a successful fetch. Otherwise it seems like the fetch didn't work

## Launchcontrol
Improve google sheet configuration integration

### DIFF
--- a/packages/server/nodemon.json
+++ b/packages/server/nodemon.json
@@ -16,5 +16,5 @@
     "client/**/*",
     "builder/**/*"
   ],
-  "exec": "yarn build && node --no-node-snapshot ./dist/index.js"
+  "exec": "yarn build && node --enable-source-maps --no-node-snapshot ./dist/index.js"
 }
--- a/packages/worker/nodemon.json
+++ b/packages/worker/nodemon.json
@@ -10,5 +10,5 @@
   ],
   "ext": "js,ts,json",
   "ignore": ["**/*.spec.ts", "**/*.spec.js", "../*/dist/**/*"],
-  "exec": "yarn build && node --no-node-snapshot dist/index.js"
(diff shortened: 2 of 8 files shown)


packages/server/nodemon.json
```diff
@@ -16,5 +16,5 @@
     "client/**/*",
     "builder/**/*"
   ],
-  "exec": "yarn build && node --no-node-snapshot ./dist/index.js"
+  "exec": "yarn build && node --enable-source-maps --no-node-snapshot ./dist/index.js"
```
samwho, line 19: Is this intentional?

packages/worker/nodemon.json
```diff
@@ -10,5 +10,5 @@
   ],
   "ext": "js,ts,json",
   "ignore": ["**/*.spec.ts", "**/*.spec.js", "../*/dist/**/*"],
-  "exec": "yarn build && node --no-node-snapshot dist/index.js"
+  "exec": "yarn build && node --enable-source-maps --no-node-snapshot dist/index.js"
```
samwho, line 13: And this.

samwho (approved):

packages/server/nodemon.json
adrinr: Yes, it is. I realised we are not using sourcemap neither in dev mode, and if we revert [this](https://github.com/Budibase/budibase/pull/16255) we will still get that in

PR #16259, opened 2025-06-03
Fix timeout errors causing step history to be lost.

samwho:
## Description

If an automation times out, we never appended the step results to the automation results causing it to look like the automation didn't do anything. This PR fixes that by keeping track of intermediate results and using them if the automation times out.

## Launchcontrol

No longer loses automation step history if the automation times out.


adrinr (approved):
Thx for the tests!

PR #16266, opened 2025-06-04
Allow selection of relationship fields in junction tables when using the Create Row automation step.

samwho:
## Description

The frontend code that filtered out fields for selection in the create row step made the assumption that all autocolumns are unassignable, but this isn't true. Foreign keys can (and often need to be) assigned. This PR changes that frontend logic to allow selecting foreign keys.

## Launchcontrol

Fixed a bug where you couldn't select foreign keys to write to in the "create row" automation step.


packages/builder/src/components/automation/SetupPanel/RowSelector.svelte
```diff
@@ -71,6 +71,10 @@
 
   $: initData(tableId, $memoStore?.meta?.fields, $memoStore?.row)
 
+  const isAutoincrement = field => {
+    return field.autocolumn && field.autoReason !== AutoReason.FOREIGN_KEY
```
adrinr, line 75: Will this always be true? Would be checking if it is from type `AutoFieldSubType.AUTO_ID` work?

samwho: It doesn't look like we set that on primary key fields imported from external tables, e.g.:

CleanShot 2025-06-09 at 15 58 14@2x

vs:

CleanShot 2025-06-09 at 15 58 43@2x

I'm not sure if it will always be true but so far it looks like the only solid signifier available to me.

adrinr (approved):

PR #16270, opened 2025-06-04
Chore/upgrade lerna

melohagan:
## Description
Use Python 3.13


samwho:
This appears to use python 3.13 in the worker Dockerfile and nowhere else. Should we also update `.tool-versions` `.python-version` and the other Dockerfile, e.g. `packages/server/Dockerfile` and `hosting/single/Dockerfile`?

PR #16274, opened 2025-06-05
[BUDI-9382] Fix SSO enforcement preventing bulk user imports

samwho:
When a bulk create is happening and SSO is enabled, remove any passwords attached to the user so that we don't trigger the validation around password reset being disabled for SSO.


packages/backend-core/src/users/db.ts
```diff
@@ -372,21 +372,26 @@ export class UserDB {
...
+            delete user.password
+          }
+
           usersToSave.push(
             UserDB.buildUser(
               user,
               {
                 hashPassword: true,
-                requirePassword: user.requirePassword,
+                requirePassword: user.requirePassword !== false,
```
adrinr: Confusing condition 😅

packages/backend-core/src/users/db.ts
```diff
@@ -372,21 +372,26 @@ export class UserDB {
...
     const account = await accountSdk.getAccountByTenantId(tenantId)
+    const isSSOEnforced = await UserDB.features.isSSOEnforced()
+
     return UserDB.quotas.addUsers(
       newUsers.length,
       newCreators.length,
       async () => {
-        // create the promises array that will be called by bulkDocs
         newUsers.forEach((user: any) => {
+          if (isSSOEnforced && user.password) {
```
adrinr: Should this not be just checking `if (isSSOEnforced)`, to get sure we don't have empty strings or other kind of value on `password`?

adrinr (approved):

packages/backend-core/src/users/db.ts
samwho: You're right, I've done a bunch of cleanup and in the process I realised that I'm _fairly sure_ nothing actually passes `requirePassword`. Correcting the types in this function revealed that it's not actually a field on the input argument to the function, and I didn't find any references in the frontend to setting it. I simplified it down to `requirePassword: !isSSOEnabled`

packages/backend-core/src/users/db.ts
samwho: Good call, done.

PR #16286, opened 2025-06-06
[Added] Models needed for actions tracking

jvcalderon:
## Description

An action is:
- An action step within automations - example: the send email automation is run
- A write operation within an app - example: user submits form 
- A request completed by an agent - example: user asks agent “what are the rules around ticket escalation?”

This PR adds `action` property to quota and license structures in order to track its value.

## Addresses

https://linear.app/budibase/issue/GRO-1209/actions-tracking

### DIFF
--- a/packages/builder/src/stores/portal/licensing.ts
+++ b/packages/builder/src/stores/portal/licensing.ts
@@ -57,8 +57,10 @@ interface LicensingState {
   userCount?: number
   userLimit?: number
   aiCreditsLimit?: number
+  actionsLimit?: number
   userLimitReached: boolean
   aiCreditsExceeded: boolean
+  actionsExceeded: boolean
   errUserLimit: boolean
 }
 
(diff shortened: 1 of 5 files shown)


packages/builder/src/stores/portal/licensing.ts
```diff
@@ -133,6 +136,16 @@ class LicensingStore extends BudiStore<LicensingState> {
     return aiCredits > aiCreditsLimit
   }
 
+  actionsExceeded(
+    actions: number,
+    actionsLimit = get(this.store).actionsLimit
+  ) {
+    if (actionsLimit === UNLIMITED || actionsLimit === undefined) {
```
melohagan, line 143: NAB: I don't think `actionsLimit === undefined` is needed because `undefined > actionsLimit` will always be false.
```suggestion
    if (actionsLimit === UNLIMITED) {
```

melohagan (approved):

PR #16289, opened 2025-06-06
fix(RowExplorer): add repeater context to prevent multiple ejections

Stropdasman:
When ejecting a RowExplorer, the repeater incorrectly generates 10 spectrumcard components, even though only 1 should be generated.
This happens because the spectrumcard was missing the correct context, causing Budibase to misinterpret the data scope.

This PR explicitly sets context="repeater" and correctly binds the props to the current row in the repeater.
This ensures that only one spectrumcard is generated per row during eject.

Before: 
image

After:
image


Stropdasman:

melohagan (approved):
Thank you! Ran into this one recently myself 👍

PR #16294, opened 2025-06-09
Apps to workspaces exports

adrinr:
## Description
Renaming "apps" to "workspaces" around the workspaces exports. This renaming is behind a feature flag


packages/builder/src/types/app.ts
```diff
@@ -10,3 +12,21 @@ export type AppTemplate =
       background: string
       icon: string
     }
+
+export interface AppIdentifierMetadata {
```
mike12345567, line 16: Can we keep types in `@budibase/types`? I think having a single type import of `import type {...} from @budibase/types` is a lot easier than having to import types from multiple places.

mike12345567 (approved):
LGTM - just one comment about trying to keep types in budibase types package - having a type/interface within a file which is only used by that file is fine, but I'd rather keep all shared types in one place across the whole project.

adrinr:
> LGTM - just one comment about trying to keep types in budibase types package - having a type/interface within a file which is only used by that file is fine, but I'd rather keep all shared types in one place across the whole project.

I think we should define that. In my opinion, it makes sense to share types across packages in `budibase/types`, but these are types used locally on the builder, so adding them outside only adds noise and makes it unclear what the actual usage is. I see it as if any type declared in a single file were declared in a shared location

PR #16267, opened 2025-06-04
Add AI Agents with Tool Source Management

shogunpurple:
## Description

This PR introduces a comprehensive AI Agents system to Budibase, providing users with intelligent chat capabilities powered by configurable tool sources. The implementation includes a full chat interface, tool source management system, and integration with external services like GitHub and Confluence.

## TODOs
- remove code to show tool debugging information. Going to build a more comprehensive set of debugging/log tools so you can see what each tool call does
- Add Budibase to the tools config panel by default as you don't need to connect it
- PH events
- Support for injecting the guidelines (additional prompting) on the toolsource into the system prompt
- UI polish based on the designs
- Testing on the workflows defined in the AI agents pitch

### 🤖 AI Chat Interface
- Interactive chat interface with message history
- Real-time agent responses with tool execution debugging
- Chat session persistence and management
- Auto-scrolling and focus management for optimal UX

### 🔧 Tool Source Management
- Support for multiple tool sources: Budibase, GitHub, Confluence
- Dynamic tool discovery and configuration
- Per-tool enable/disable toggles for fine-grained control
- Secure API key management with guidelines support

### 🛠️ Backend Infrastructure
- RESTful API endpoints for agent chat and tool source management
- Database schemas for chat history and tool source configuration
- Integration with OpenAI for LLM capabilities
- Extensible tool system supporting both built-in and external tools

### 📊 Technical Improvements
- Comprehensive TypeScript type definitions
- Full type safety across frontend and backend
- Proper error handling and validation
- Performance optimizations for real-time chat

### New Components
- `AgentChat` interface with full message history
- `AgentToolSource` and `AgentToolSourceWithTools` types for tool management
- Chat UI with history panel and tool configuration panel
- Logo components for various tool sources

### API Endpoints
- `POST /api/agent/chat` - Send messages to AI agent
- `GET /api/agent/history` - Retrieve chat history
- `GET /api/agent/toolsource` - Fetch available tool sources
- `POST /api/agent/toolsource` - Create new tool source
- `PUT /api/agent/toolsource` - Update tool source configuration

### Database Schema
- Agent chat document storage with CouchDB
- Tool source configuration persistence
- Proper document ID generation and management

## Technical Challenges Solved

1. **Type Safety**: Implemented comprehensive TypeScript types across the entire stack
2. **Real-time UX**: Optimized chat interface with proper scrolling and focus management
3. **Tool Integration**: Created extensible system for adding new tool sources
4. **State Management**: Proper Svelte store management for complex agent state

[...]

PR #16300, opened 2025-06-09
streaming support for agents

shogunpurple:
## Description
Streaming support for the AI agent created in https://github.com/Budibase/budibase/pull/16267

Working for OpenAI directly, but have added some "stubs" to the anthropic and BudibaseAI setup to still call the streaming method, but just return the full response in one go until it's implemented properly

Pro PR: 
- https://github.com/Budibase/budibase-pro/pull/429
- https://github.com/Budibase/budibase-pro/pull/434


packages/frontend-core/src/api/agents.ts
```diff
@@ -9,8 +9,36 @@ import {
 } from "@budibase/types"
 import { BaseAPIClient } from "./types"
 
+export interface StreamChunk {
```
samwho: Is there not already a type for this on the backend? It would be good if the frontend and backend shared the same type so they don't fall out of sync.

packages/frontend-core/src/api/agents.ts
```diff
@@ -32,6 +60,67 @@ export const buildAgentEndpoints = (API: BaseAPIClient): AgentEndpoints => ({
     })
   },
 
+  agentChatStream: async (chat, onChunk, onError) => {
```
samwho, line 42: Why not also make this an async generator?

shogunpurple: It would require an update to how the UI handles it, and with no real benefit - the callback based approach works the same way

samwho: Fair.

samwho (approved):

PR #16306, opened 2025-06-09
Filtered publishing (feature flagged behind workspace apps)

mike12345567:
## Description

This PR implements filtered publishing functionality that allows selective deployment of specific automations and workspace apps along with their dependencies, rather than publishing entire applications.

### 🎯 Selective Resource Publishing

  - Automation Filtering: Publish only specified automations by providing `automationIds`
  - Workspace App Filtering: Publish only specified workspace apps by providing `workspaceAppIds`
  - Dependency Resolution: Automatically includes all related resources (tables, datasources, queries, row actions) used by the filtered automations/workspace apps
  
### 🔧 Core Implementation

1. Enhanced Publish Endpoint (`packages/server/src/api/controllers/deploy/index.ts`)
  - Modified `publishApp` to accept `PublishAppRequest` body with filtering parameters
  - Integrated `buildPublishFilter()` to create replication filters
  - Maintains backward compatibility - works with or without filtering parameters

2. Resource Analysis Engine (`packages/server/src/sdk/app/resources/index.ts`)
  - `analyseMinimal()`: Identifies basic resources used by automations/workspace apps
  - `analyseAll()`: Performs comprehensive dependency analysis including:
    - Tables and datasources referenced in automations/screens
    - Queries associated with datasources
    - Row actions attached to tables
    - Row action automations

3. Publishing Filter (`packages/server/src/api/controllers/deploy/filters.ts`)
  - Creates filter functions for selective document replication
  - Filters by document type: automations, screens, workspace apps, tables, datasources, queries, row actions
  - Ensures only specified resources and their dependencies are published
  
 Description generated by Claude Code.
 
 Screenshot of feature in action:
 
image


packages/shared-core/src/automations/triggers/rowAction.ts
```diff
@@ -10,11 +10,11 @@ import {
 export const definition: AutomationTriggerDefinition = {
   type: AutomationStepType.TRIGGER,
   name: "Row Action",
-  event: AutomationEventType.ROW_ACTION, // TODO
-  icon: "Workflow", // TODO
+  event: AutomationEventType.ROW_ACTION,
```
mike12345567, line 13: Noticed this when I was building out row actions for tests - these `TODO` statements really aren't valid - this is fully deployed now and we are dependent on these components of the row action - added a description rather than a TODO (description is not user viewable, would just be seen in the database).

packages/bbui/src/Accordion/Accordion.svelte
```diff
@@ -32,7 +33,13 @@
         <use xlink:href="#spectrum-css-icon-Chevron100" />
       </svg>
     </h3>
-    <div class="spectrum-Accordion-itemContent" role={itemName}>
+    <div
+      class="spectrum-Accordion-itemContent"
+      role={itemName}
+      style={noPadding
+        ? "padding-left: 20px; padding-bottom: 0;"
+        : "padding-left: 30px;"}
```
samwho, line 41: I'm not an expert but don't we tend to use the spacing variables for this, instead of hardcoded pixel values?

packages/builder/src/components/deploy/AppActions.svelte
```diff
@@ -13,23 +13,39 @@
...
   let showNpsSurvey = false
   let publishButton
   let publishSuccessPopover
+  const workspaceEnabled = featureFlag.isEnabled(FeatureFlag.WORKSPACE_APPS)
 
   $: updateAvailable =
     $appStore.upgradableVersion &&
     $appStore.version &&
     $appStore.upgradableVersion !== $appStore.version
+  $: inAutomations = $page.path.includes("automation")
```
samwho: This and the next line feel risky, what if these strings are part of the tenant ID which will feature in app IDs and the like?

packages/builder/src/components/deploy/AppActions.svelte
```diff
@@ -13,23 +13,39 @@
...
   import { admin } from "@/stores/portal"
   import VersionModal from "@/components/deploy/VersionModal.svelte"
+  import PublishModal from "@/components/deploy/PublishModal.svelte"
 
   let versionModal
+  let publishModal
   let showNpsSurvey = false
   let publishButton
   let publishSuccessPopover
+  const workspaceEnabled = featureFlag.isEnabled(FeatureFlag.WORKSPACE_APPS)
```
samwho: There's a feature flag store that's properly reactive an stuff, search `$featureFlag` to find usage examples.

packages/builder/src/components/deploy/PublishModal.svelte
```diff
@@ -0,0 +1,223 @@
...
+
+  export function hide() {
+    publishModal.hide()
+  }
+
+  async function getUsedResources(appIds: string[], automationIds: string[]) {
+    if (!appIds.length && !automationIds.length) {
+      return
+    }
+    let { resources } = await API.resource.analyse({
```
samwho: I'm not sure if you've added this API in this PR, but the word analyze is sufficiently general I don't really know what it's doing.

packages/builder/src/components/deploy/PublishModal.svelte
```diff
@@ -0,0 +1,223 @@
...
+
+  function removeTarget<T extends { _id?: string }>(
+    target: string,
+    list: T[]
+  ): T[] {
+    return list?.filter(item => item._id !== target) || []
+  }
+
+  function removeRowActionAutomations(automations: Automation[]) {
+    return automations?.filter(
```
samwho:
```suggestion
    return automations.filter(
```

packages/builder/src/stores/builder/deployment.ts
```diff
@@ -83,16 +83,13 @@ class DeploymentStore extends DerivedBudiStore<
...
     try {
       this.update(state => ({ ...state, isPublishing: true }))
-      await API.publishAppChanges(get(appStore).appId)
-      if (showNotification) {
-        notifications.send("App published successfully", {
-          type: "success",
-          icon: "GlobeCheck",
-        })
-      }
+      await API.publishAppChanges(get(appStore).appId, opts)
```
samwho, line 92: How come you're getting rid of showing the notification here?

packages/server/src/api/controllers/deploy/filters.ts
```diff
@@ -0,0 +1,61 @@
...
+  DocumentType,
+  ResourceType,
+  Screen,
+  SEPARATOR,
+} from "@budibase/types"
+import sdk from "../../../sdk"
+
+type Filter = PouchDB.Replication.ReplicateOptions["filter"]
+
+const AUTOMATION_ID_FORMAT = DocumentType.AUTOMATION + SEPARATOR,
```
samwho, line 23: Don't we already have a wide variety of functions lying around to check if a document is of a specific type?

packages/server/src/api/controllers/resource.ts
```diff
@@ -0,0 +1,23 @@
...
+} from "@budibase/types"
+import sdk from "../../sdk"
+
+export async function analyse(
+  ctx: UserCtx<ResourceAnalysisRequest, ResourceAnalysisResponse>
+) {
+  const { workspaceAppIds, automationIds } = ctx.request.body
+
+  if (!workspaceAppIds?.length && !automationIds?.length) {
+    ctx.throw(401, "No workspace apps or automations specified.")
```
samwho: This should likely be a 400. I think a 401 is to do with authentication (https://http.cat/401)

packages/server/src/sdk/app/resources/index.ts
```diff
@@ -0,0 +1,158 @@
...
+        type: ResourceType.DATASOURCE,
+      }))
+    )
+
+  let internalTableFound = false
+  const resources: UsedResource[] = []
+  const searchForResource = (json: string) => {
+    for (let search of toSearchFor) {
+      if (
+        json.includes(search.id) &&
```
samwho, line 57: This doesn't feel very solid, users could put resource IDs in places without using those resource IDs. I know it's a lot of extra work to parse the structure of the screen but I think it's correct to do so.

packages/server/src/sdk/app/resources/index.ts
```diff
@@ -0,0 +1,158 @@
...
+        })
+      }
+    }
+  }
+
+  if (workspaceAppIds?.length) {
+    const screens = await sdk.screens.fetch()
+    const workspaceAppScreens: Record<string, Screen[]> = {}
+
+    for (let screen of screens) {
```
samwho: nit: lots of `let`s here that should be `const`s

packages/server/src/sdk/app/resources/index.ts
```diff
@@ -0,0 +1,158 @@
...
+    const automations = await sdk.automations.find(automationIds, {
+      allowMissing: true,
+    })
+    for (let automation of automations) {
+      const json = JSON.stringify(automation)
+      searchForResource(json)
+    }
+  }
+
+  // internal table found, need to make sure all internal tables have been added
```
samwho: Why?

packages/server/src/sdk/app/resources/index.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+  return resources
+}
+
+export async function analyseAll(toCheck: {
+  automationIds?: string[]
+  workspaceAppIds?: string[]
+}) {
+  const minimalResources = await analyseMinimal(toCheck)
+  // copy list, will expand it with queries and row actions
```
samwho: It feels like this should instead be one function where you pass in a list of the resource types you want to find.

packages/server/src/sdk/app/resources/index.ts
```diff
@@ -0,0 +1,158 @@
...
+  )
+  const automations = await sdk.automations.find(missingAutomationIds)
+  // add missing automations to resources
+  automations.map(automation =>
+    resources.push({
+      id: automation._id!,
+      name: automation.name,
+      type: ResourceType.AUTOMATION,
+    })
+  )
```
samwho:
```suggestion
  resources.push(...automations.map(automation => {
    return {
      id: automation._id!,
      name: automation.name,
      type: ResourceType.AUTOMATION,
    }
  })
```

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -471,6 +474,15 @@ export default class TestConfiguration {
     }
   }
 
+  async defaultToProduction(cb: () => Promise<unknown>) {
```
samwho: Why not use `withProdApp`?

packages/builder/src/stores/builder/deployment.ts
mike12345567: the notification was only needed for one operation, it shouldn't be in the store - it should be done in the UI where it is needed so I moved it there.

packages/bbui/src/Accordion/Accordion.svelte
mike12345567: This utilises some hard-coded pixel values already see master: https://github.com/Budibase/budibase/blob/1b19e5e185a10e91a4d1e0e6e2ac64186fbf15d4/packages/bbui/src/Accordion/Accordion.svelte#L50 which I don't really want to change - I'm not attempting to adjust existing accordions, I just need an accordion that is compliant with the designs here.

packages/server/src/api/controllers/deploy/filters.ts
mike12345567: This is a filter that is used as part of replication - this filter executes inside PouchDB/CouchDB and doesn't have access to our entire codebase.

We could use these functions here, but it ties those functions to being usable here, I've found in the past this is dangerous as changing a function elsewhere in the system can lead to a weird Pouch/Couch breakage, keeping this function as simple as possible and keeping it isolated avoids that issue.

I'll add a comment to explain this to the top of the file.

samwho: Ahhhh of course, I didn't think about that. A comment would be very handy.

packages/server/src/tests/utilities/TestConfiguration.ts
mike12345567: I did not realise this was available - good to know!

packages/server/src/sdk/app/resources/index.ts
mike12345567: We treat the internal tables as a single internal datasource today - the single "Budibase DB" - in future these maybe broken down into different datasources, but for now we need to make sure that if any internal resources are published we publish the entire internal datasource (tables with relationships etc - this is the lowest risk way to manage this).

packages/server/src/sdk/app/resources/index.ts
mike12345567: I think it is quite a bit more dangerous to attempt to find where the ID is used, there is really two approaches here:
1. look for the IDs as a raw string search - its possible for the users to specify resource IDs in their screen/automation manually which the app isn't necessarily tied to, but all that happens is these are published alongside the app (not breaking).
2. Attempt to work out anywhere that a resource ID could be used in the structure - this requires an intimate knowledge of these structures which will need updated regularly - if we miss anywhere that a resource ID could go we will *not* publish needed resources during the publish leading to a likely broken app.

In my opinion the first option is a lot safer since over-publishing is very unlikely to lead to a problem (closer to what we do today) and the second option is close to impossible - everywhere that a resource ID could be specified could also contain user input which might be a resource ID un-intentionally leading to the same problem.

mike12345567: Discussed on Discord, sticking with current approach.

packages/server/src/sdk/app/resources/index.ts
mike12345567: I've done this - got Claude to help with it, after much back and forth it is done - quite the monster function but gets it altogether.

samwho (approved):

PR #16313, opened 2025-06-10
[BUDI-9039] Fix conditional formatting refresh for external tables.

samwho:
## Description

For external datasources we always use the `_rev` `"rev"`. Conditional formatting was relying on this value changing to know when to recalculate its conditions, and so wasn't working for external datasources.

This PR explicitly checks for a `_rev` of `"rev"` and makes sure we always recalculate for those rows.

## Addresses
- https://linear.app/budibase/issue/BUDI-9039/refresh-data-providers-dont-refresh-conditional-appearences-on-tables

## Launchcontrol

Fix conditional cell formatting not applying when data changed in external datasources.


adrinr (approved):
Worth adding a test for it?

PR #16314, opened 2025-06-10
Create a default admin user in local dev that Claude knows about.

samwho:
## Description

This will make local browser use much easier.


adrinr (approved):
LGTM, happy to update this if it improve testing flows

PR #16316, opened 2025-06-10
Support importing workspaces

adrinr:
## Description
Support importing workspace apps. These changes will cause the migrations doc to be imported as well, so if there is a need to run any migration, it will ran. Otherwise, it will not as the migration document will tell us that is already implemented


packages/server/src/utilities/fileSystem/clientLibrary.ts
```diff
@@ -175,7 +175,7 @@ export function shouldServeLocally(version: string) {
     return false
   }
 
-  if (version === "0.0.0") {
+  if (env.isDev()) {
```
adrinr, line 178: This was removed recently, so we should always serve locally on dev:
- #16191

packages/backend-core/src/constants/db.ts
```diff
@@ -78,3 +78,8 @@ export const DEFAULT_INVENTORY_TABLE_ID = "ta_bb_inventory"
 export const DEFAULT_EXPENSES_TABLE_ID = "ta_bb_expenses"
 export const DEFAULT_EMPLOYEE_TABLE_ID = "ta_bb_employee"
 export { DEFAULT_BB_DATASOURCE_ID } from "@budibase/shared-core"
+
+export const enum DesignDocuments {
+  SQLLITE = SQLITE_DESIGN_DOC_ID,
```
mike12345567: Spelling here - `SQLITE`

mike12345567 (approved):
LGTM - a test case using the template import is likely a good solution to test this and much more but I think it should be done separately as it is probably a bigger job.

One small comment about the spelling in an enum, might cause a problem otherwise!

packages/backend-core/src/constants/db.ts
adrinr: Good catch!

adrinr:
> LGTM - a test case using the template import is likely a good solution to test this and much more but I think it should be done separately as it is probably a bigger job.
> 
> One small comment about the spelling in an enum, might cause a problem otherwise!

I am tackling the template usage separately. I'll add the test in there

PR #16321, opened 2025-06-10
Support workspaces when using templates

adrinr:
## Description
Support workspaces when using templates


mike12345567 (approved):
Very nice - template test was pretty simple which is great!

adrinr:
> Very nice - template test was pretty simple which is great!

Not simple to mock, but good and easy :)

PR #16323, opened 2025-06-10
Set maximum number of active sessions per user to 3 (from 1)

samwho:
## Description

This PR allows users to be logged into 3 active sessions at a time instead of 1. This gives people more flexibility in how they use Budibase, e.g. on a desktop and a mobile device at the same time.

The way it works is by attaching a header to the response of the login request that indicates if you logged out any other sessions with this login. If you did, that information gets stored in `localStorage` (because several redirects happen on login) and pulled out of `localStorage` to display a `notification.info` when the page has settled, explaining that another session has been logged out.

https://github.com/user-attachments/assets/6f1ea505-f954-476c-a88e-8cd1b1d70ed8

## Launchcontrol

Allow users to be logged into 3 active sessions at a time, up from 1. You will no longer get logged out of Budibase if you log in elsewhere, provided you're not logged in to more than 3 sessions at a time.


packages/builder/src/pages/builder/portal/apps/_layout.svelte
```diff
@@ -14,7 +14,7 @@
   import PortalSideBar from "./_components/PortalSideBar.svelte"
 
   // Don't block loading if we've already hydrated state
-  let loaded = !!$appsStore.apps?.length
+  let loaded = !!$appsStore.apps?.lengthcheckInvalidSessionsAndNotify
```
aptkingston: Looks like an accidental paste.

aptkingston (commented):
LGTM other than the comment mentioned 👌 I think local storage is a great choice here to ensure we handle any number of hard or soft redirects.

I don't remember off the top of my head, but do we handle redirects after login if you originally requested another URL?

I'm thinking of a flow like:
- I try to go to a builder URL for a specific app
- I get redirected to login as I'm not logged in
- I log in and get redirected back to the original URL, the builder

In this scenario, not only would I not see the notification (since I bypassed the portal apps page), but I'd also erroneously see the popup the next time I go to the portal apps page, as it still exists in local storage.

If that's a valid consideration, I think we should move the logic handling this popup into the core `builder/_layout.svelte` to ensure that we show it anywhere. There's still the possibility that the user is redirected to a client app though, so do we also need this logic added to the client library? It's also the case that the session limit applies there too.

adrinr:
> LGTM other than the comment mentioned 👌 I think local storage is a great choice here to ensure we handle any number of hard or soft redirects.
> 
> I don't remember off the top of my head, but do we handle redirects after login if you originally requested another URL?
> 
> I'm thinking of a flow like:
> 
>     * I try to go to a builder URL for a specific app
> 
>     * I get redirected to login as I'm not logged in
> 
>     * I log in and get redirected back to the original URL, the builder
> 
> 
> In this scenario, not only would I not see the notification (since I bypassed the portal apps page), but I'd also erroneously see the popup the next time I go to the portal apps page, as it still exists in local storage.
> 
> If that's a valid consideration, I think we should move the logic handling this popup into the core `builder/_layout.svelte` to ensure that we show it anywhere. There's still the possibility that the user is redirected to a client app though, so do we also need this logic added to the client library? It's also the case that the session limit applies there too.

Valid point. Yes, we handle redirects after login, even if it goes via the portal homepage anyways (it looks flaky but it might "solve" the notification issue you mentioned)

packages/frontend-core/src/utils/login.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function checkIfSessionsInvalidatedAndNotify() {
+  const invalidatedCount = parseInt(
+    localStorage.getItem(SESSIONS_INVALIDATED_KEY) || "0",
+    10
+  )
+
+  if (invalidatedCount > 0) {
+    const sessionText = invalidatedCount === 1 ? "session" : "sessions"
+    notifications.info(
```
aptkingston: I'm not sure this will work in the client library (but maybe it did if it's pushed up?) as it uses its own notification system that doesn't use BBUI notifications. (No, I'm not sure why and we should really just use BBUI notifications across the board). You might need to update this to accept some sort of "notify" param and pass in `notificationStore.actions.info` from the client library instantiation.

aptkingston (commented):
LGTM but I think we'll just need a small change to get the client library implementation working.

packages/frontend-core/src/utils/login.ts
samwho: I'll be honest, I assumed this would be fine when I saw other bbui things referenced in packages/client 😓 Will fix tomorrow.

aptkingston (approved):

PR #16325, opened 2025-06-11
Typing user section

adrinr:
## Description
Typing user section on the portal done during the workspace app migrations. No logic changes


packages/builder/src/types/user.ts
```diff
@@ -0,0 +1,8 @@
+import { Constants } from "@budibase/frontend-core"
+
+export interface UserInfo {
```
mike12345567, line 3: Could we move this type to Budibase types, under the UI section?

mike12345567 (approved):
LGTM!

PR #16288, opened 2025-06-06
Rename app labels to workspaces

adrinr:
## Description
Initial renaming from `app` to `workspace`. This came with quite a lot of ts conversion and cleanups.
The converted areas are mainly the dashboard apps and their modals.
Note that not all the references across the builder have been converted, just creating a PR to avoid this from growing massively. Everything is behind the feature flag

PR #16331, opened 2025-06-11
Fix importing app

adrinr:
## Description
Fix issues after some refactor in https://github.com/Budibase/budibase/pull/16288
We updated to use json instead of FormData, but I missed that this is required in order to accept file upload.... "Reverting" these changes to get it working

Tested scenarios:
1. Create an app from scratch
2. Create an app from a template
3. Create an app from a template via url import
4. Import backup
5. Duplicate app

### DIFF
--- a/packages/frontend-core/src/api/app.ts
+++ b/packages/frontend-core/src/api/app.ts
@@ -38,7 +38,7 @@ export interface AppEndpoints {
   revertAppClientVersion: (appId: string) => Promise<RevertAppClientResponse>
   releaseAppLock: (appId: string) => Promise<ClearDevLockResponse>
   getAppDeployments: () => Promise<FetchDeploymentResponse>
-  createApp: (app: CreateAppRequest) => Promise<CreateAppResponse>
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):

packages/frontend-core/src/api/app.ts
```diff
@@ -136,6 +136,14 @@ export const buildAppEndpoints = (API: BaseAPIClient): AppEndpoints => ({
    * @param app the app to create
    */
   createApp: async app => {
+    if (app instanceof FormData) {
+      return await API.post({
+        url: "/api/applications",
+        body: app,
+        json: false,
```
melohagan, line 143: Instead of doing this `instanceof` check, you could simply pass the **json** boolean as an optional flag that defaults to true

melohagan (dismissed):

packages/frontend-core/src/api/app.ts
adrinr: This means that the consumer needs to pass it, instead of being infered. If that was the case, we could have wrong usages

PR #16332, opened 2025-06-11
Fix nav item widths and scrolling

aptkingston:
## Description
Fixes a few issues with nav item widths, scrolling and text overflow, which are currently live.

Currently:
image
image

This branch:
image
image


ConorWebb96 (approved):
LGTM!

PR #16333, opened 2025-06-11
Allowing admins and builders to access agent chat/history

mike12345567:
## Description
As title - permissions were set to admin only.


shogunpurple (approved):
LGTM

PR #16334, opened 2025-06-11
[Added] Actions tracking

jvcalderon:
## Description

Emit an event every time an action is done.

An action is:
- An action step within automations - example: the send email automation is run
- A write operation within an app - example: user submits form 
- A request completed by an agent - example: user asks agent “what are the rules around ticket escalation?”

`aiAgentExecuted` method has been created for the last point (a request completed by an agent) but it is not used yet.

## Addresses
https://linear.app/budibase/issue/GRO-1209/actions-tracking

## Screenshots

Action events are sent to Posthog for tracking

Screenshot 2025-06-11 at 16 26 15


packages/backend-core/src/events/publishers/action.ts
```diff
@@ -0,0 +1,34 @@
...
+} from "@budibase/types"
+
+async function automationStepExecuted(
+  action: ActionAutomationStepExecuted,
+  timestamp?: string | number
+) {
+  await publishEvent(Event.ACTION_AUTOMATION_STEP_EXECUTED, action, timestamp)
+}
+
+async function crudExecuted(
```
melohagan, line 16: It's not CRUD it's CUD 😁 Maybe there's some other name that would work like "writeQuery", idk

*Read* does not trigger an action increment.

packages/backend-core/src/events/publishers/action.ts
```diff
@@ -0,0 +1,34 @@
...
+  timestamp?: string | number
+) {
+  await publishEvent(Event.ACTION_CRUD_EXECUTED, action, timestamp)
+}
+
+async function aiAgentExecuted(
+  action: ActionAiAgentExecuted,
+  timestamp?: string | number
+) {
+  await publishEvent(Event.ACTION_CRUD_EXECUTED, action, timestamp)
```
melohagan: Using the wrong event enum

packages/types/src/sdk/events/event.ts
```diff
@@ -196,6 +196,11 @@ export enum Event {
 
   // ROW ACTION
   ROW_ACTION_CREATED = "row_action:created",
+
+  //ACTION
+  ACTION_AUTOMATION_STEP_EXECUTED = "action:automation_step:executed",
+  ACTION_CRUD_EXECUTED = "action:crud:executed",
```
melohagan, line 202: As mentioned before, CRUD doesn't quite represent what this is

melohagan (changes requested):

packages/backend-core/src/events/publishers/action.ts
jvcalderon: Fixed. Thanks.

packages/backend-core/src/events/publishers/action.ts
jvcalderon: - crudExecuted -> Everyone will associate it with persistence operations, but it suggests that a read operation is also included (but currently we don't trigger the event for read operations).

- cudExecuted -> Theoretically, it expresses exactly what we want, but it's less well-known, sounds strange, and may seem like a typo.

In my opinion, we should leave crudExecuted because the use of this event depends on the context in which it's used, not on the method by itself. I mean, If for any reason in the future we don't fire the event with delete operations either, we shouldn't change the event to cuExecuted. It should always be crudExecuted. For the same reason, if we want to trigger it for read in the future, shouldn't be changed to crud again.

crudExecuted expects to receive an object of type QueryVerb, which contains all operations, including reading, in order to use it with flexibility depending on the needs.

packages/types/src/sdk/events/event.ts
jvcalderon: `action:crud:executed` event contains a body with the following information (example):

```json
{ "type": "create" }
```

If for some reason, in the future we want to trigger this event for reading operation as well, it will be:

```json
{ "type": "read" }
```

(Without having to change the event name)

melohagan (approved):
LGTM

PR #16338, opened 2025-06-12
Add new Automation Logs viewer 

PClmnt:
This PR introduces moves the Automation Logs into the automation section, as well as totally updating how a user views and interacts with those logs

**New Feature Overview:**
- **Dual View Mode System**: Toggle between Editor mode (for building automations) and Logs mode (for viewing execution history)
- **Automation Logs Panel**: Browse historical automation runs with filtering by status and date range
- **Log Details Panel**: Detailed step-by-step view of automation execution with input/output data and error information
- **Visual Flow Integration**: Interactive automation flow that shows only executed steps with success/error indicators
- **Smart Data Mapping**: Intelligently maps log data to automation steps using stepId matching

**Key Components:**
1. `AutomationLogsPanel.svelte` - Lists automation execution history with pagination and filtering
2. `LogDetailsPanel.svelte` - Detailed view with tabbed interface for step data (Data In, Data Out, Issues)
3. Enhanced `FlowChart.svelte` - Supports dual view modes and step selection for log viewing
4. Enhanced flow components - Visual indicators and click handling for log data viewing
5. `ViewMode` enum - Type-safe view mode management replacing string literals

**User Workflow:**
1. User switches to "Logs" tab in automation builder
2. Browses list of past automation executions with filtering options
3. Selects a specific log entry to examine
4. Views the automation flow as it actually executed (filtered to show only executed steps)
5. Clicks individual steps to view detailed input/output data and any errors

## Screenshots
New Editor / Logs switcher
image

Logs Panel
image

Logs Detail Panel default state
image

Selecting a block to view it's log output
image

## Launchcontrol
Added comprehensive automation logs viewer with execution history browsing, step-by-step debugging, and visual flow integration for improved automation monitoring and troubleshooting.

## Feature branch env
[Feature Branch Link](http://fb-feat-automation-runs-ui.budibase.dev)


mike12345567:
Had a play with this - great new experience, love it! One issue I had was that triggers don't always have "data in" (i'm not sure they ever do?) and this is kind of confusing, for example an App Action

image

For triggers would it be a good idea to remove the "Data in" tab since it'll never be populated?

deanhannigan (changes requested):
Hey Pete, I ran through everything there and only came across one minor bug!

- `BUG` - Issue selecting steps in `Logs` mode. There is no selection border around the step as you make a selection. 
- Also, if you select a step before switching to `Logs` mode, it will retain its selection border regardless of what other steps you click.

Below is just some feedback on the feature itself.

-  Zoom in, zoom out buttons should use the same design as the undo/redo. The extra border and margin distorts the other buttons in the group.
- "Editor/Logs" buttons at the top floats along the header as the automation name changes. The button could just go back to the previous location, on the top right, which would stop it shifting around.
- In `Logs` mode, the `Trigger` step is not showing any `Data in` in the side bar.
  - It should show the available context elements like in `Editor` mode or, if that doesn't make sense, some kind of info copy to explain why it's blank. 
- When switching to `Logs` mode it should auto-select a step, like the trigger.
  - At the moment, it shows an empty side panel with "Select a step in the automation flow to view its data"
- Do we need entirely new side panels to display or navigate the log data flow?
  - The list of available logs makes sense but the drill-down side panel is largely empty and, in places, largely identical to the normal block data section.
  - Can the existing side panel and UI accomodate this flow instead? I think it could be a less complicated approach than mode switching and hopping through side panels. It was a little confusing.
-  Can the log UX show when logs are available? Maybe change colour or show a number?

deanhannigan (approved):
LGTM! Ran through everything there and the issues are resolved! 👍

PR #16339, opened 2025-06-12
Workspaces on my app list

adrinr:
## Description
Handling workspaces in "my apps" portal. A new endpoint was created, as the current `/apps` endpoint return "workspaces" instead of apps, and it's used in multiple places along the builder. This will need renaming in the future, but for now we are creating a new `client apps` endpoint, ensuring everything works as it should

image

As discussed, we will be displaying it a flat list, ignoring name duplicates for now


packages/server/src/api/routes/application.ts
```diff
@@ -21,6 +21,7 @@ router
   )
   .get("/api/applications/:appId/definition", controller.fetchAppDefinition)
   .get("/api/applications", controller.fetch)
+  .get("/api/applications/client", controller.fetchClientApps)
```
mike12345567: I'm not sure about this endpoint name - its not clear what this does, would it be worth creating a new resource `/api/workspace/apps` to start expanding upon? I think at some stage in the future we're going to need to update the APIs to use the workspace word as we need the "apps" word back.

I don't think this is move is for right now, but I do think new APIs should be created with regards to the workspace, especially if they are workspace focused/aware endpoints.

adrinr: After discussing it, renamed to `/api/client/applications`

mike12345567 (approved):

PR #16341, opened 2025-06-12
Render apps on builder index

adrinr:
## Description
Handling workspace app urls on portal homepage

### Not having a default route
| Before | After |
|--------|--------|
|image| image |


packages/server/src/api/controllers/application.ts
```diff
@@ -247,10 +247,28 @@ export const addSampleData = async (
...
+      result.push({
+        ...app,
+        defaultWorkspaceAppUrl: workspaceApps[0]?.urlPrefix || "",
+      })
+    }
+  } else {
+    result.push(...apps.map(a => ({ ...a, defaultWorkspaceAppUrl: "" })))
+  }
+
+  ctx.body = result
```
samwho: Does it not make more sense to handle this at the sdk level? 🤔

samwho (approved):

packages/server/src/api/controllers/application.ts
adrinr: I did not want to add it the document, and it's only use in here. But I will move it to the SDK for extraction and future reusability

mike12345567 (approved):

aptkingston (approved):

PR #16346, opened 2025-06-12
[BUDI-9202] Fix table ID parsing in <SaveRow> so that it works with sample tables

samwho:
## Description

The actual fix for this PR is a one-liner, we weren't correctly parsing the table ID out of a view ID in dataBinding.js. But the one liner required moving a bunch of ID-related functions out of packages/server and into packages/shared-core.

Pro PR: https://github.com/Budibase/budibase-pro/pull/431

## Addresses
- https://linear.app/budibase/issue/BUDI-9202/no-columns-available-when-using-a-view-as-table-source-on-save-row

## Launchcontrol

Fixed a bug that would sometimes cause no columns to be selectable in a "Save Row" on click action.


packages/builder/src/dataBinding.js
```diff
@@ -896,7 +897,7 @@ export const getSchemaForDatasourcePlus = (resourceId, options) => {
     ? {
         type: "viewV2",
         id: resourceId,
-        tableId: resourceId.split("_").slice(1, 3).join("_"),
+        tableId: getTableIdFromViewId(resourceId),
```
samwho, line 900: This is the actual bug fix.

packages/shared-core/src/ids.ts
samwho, line 1: This is code moved from the `docIds.ts` file, cleaned up a little bit.

packages/shared-core/src/ids.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ *  This function checks if the given ID is a table ID, including datasource
+ *  plus tables.
+ *
+ *  Example of IDs that will return true:
+ *  - "ta_12345"
+ *  - "datasource_plus_12345__ta_67890"
+ */
+export const isTableIdOrExternalTableId = (id: string): boolean => {
```
samwho, line 51: This used to be called `isTableId` but that felt potentially dangerous to me so I've given it a more explicit name.

Same for `isDatasourceOrDatasourcePlusId`.

packages/shared-core/src/tests/ids.test.ts
samwho, line 1: New tests to make sure ID parsing/checking functions work as expected.

mike12345567 (approved):
LGTM!

PR #16347, opened 2025-06-12
[BUDI-9220] Allow uploading attachments to anyone with table write permissions, not just creators.

samwho:
## Description

Attachments were being uploaded with the `API.uploadBuilderAttachment` API call, which requires you to be a builder to use. This PR introduces a code path where if a table ID is available, `API.uploadAttachment` is used, which allows anyone with table write access to upload a file.

## Addresses
- https://linear.app/budibase/issue/BUDI-9220/updating-an-sql-attachment-field-doesnt-work-for-anyone-other-than

## Launchcontrol

Fixed a bug where only app creators could upload SQL attachments.


packages/frontend-core/src/components/grid/cells/AttachmentCell.svelte
```diff
@@ -55,7 +55,13 @@
       data.append("file", fileList[i])
     }
     try {
-      return await API.uploadBuilderAttachment(data)
+      const tableId = $datasource?.tableId || $datasource?.id
+      if (tableId) {
+        return await API.uploadAttachment(tableId, data)
+      } else {
+        // Fallback to builder endpoint if no table context
+        return await API.uploadBuilderAttachment(data)
```
samwho, line 63: It's not obvious to me why this endpoint was chosen to use in the first place. @mike12345567 do you know?

packages/server/src/api/routes/tests/row.spec.ts
```diff
@@ -2398,6 +2398,95 @@ if (descriptions.length) {
             `${uuid.v4()}.png`
           )
         })
+
+        !isInternal &&
+          describe("external database attachment authorization", () => {
```
samwho, line 2403: These tests were written by Claude and just make sure that these upload endpoints can be access by non-creators. They always could, but this prevents any regression.

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -416,6 +416,7 @@ export default class TestConfiguration {
           sessionId: this.sessionIdForUser(userId),
           tenantId: this.getTenantId(),
           email: user.email,
+          csrfToken: this.csrfToken,
```
samwho, line 419: This bit me while iterating on the test Claude wrote. Any attempt to log in as a different user during a test would fall foul of the CSRF check if you were using a content-type that gets CSRF checked.

packages/frontend-core/src/components/grid/cells/AttachmentCell.svelte
mike12345567: I'm not sure why it was this way - I'd bet this was just the "upload attachment" endpoint originally, it got renamed and its permissions updated, then the new upload attachment endpoint was added and it was just never fixed here.

mike12345567 (approved):

PR #16360, opened 2025-06-16
Individual publish status API

mike12345567:
## Description
There needs to be a way to check if the workspace apps/automations have been published and when they were last published - for now this impacts the URL that is displayed at the top of the builder preview but it will be used on the index pages that @aptkingston is working on.

This works by pulling information from both development and production and then comparing them - ideally this will be expanded to cover change detection as well.


packages/backend-core/src/db/Replication.ts
```diff
@@ -63,11 +63,15 @@ class Replication {
 
     return {
       ...opts,
-      filter: (doc: any, params: any) => {
+      filter: (doc: Document, params: any) => {
         // don't sync design documents
         if (toDev && doc._id?.startsWith("_design")) {
           return false
         }
+        // always replicate deleted documents
```
adrinr, line 71: Why this? Will this create documents that will be "immediately" deleted?

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -148,6 +163,88 @@ export async function deploymentProgress(
...
+        }
+      }
+    }
+
+    ctx.body = {
+      automations,
+      workspaceApps,
+    }
+  } catch (err) {
+    ctx.throw(500, "Error fetching data for deployment status")
```
samwho: Better to let this bubble up and be handled by the error handling middleware, that way we get stack traces in DataDog with the appropriate context. Catching and rethrowing like this throws away the stack trace and creates a new one, making it almost impossible to track down the root of the problem.

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -148,6 +163,88 @@ export async function deploymentProgress(
...
+export async function publishStatus(ctx: UserCtx<void, PublishStatusResponse>) {
+  if (!(await features.isEnabled(FeatureFlag.WORKSPACE_APPS))) {
+    return (ctx.body = { automations: {}, workspaceApps: {} })
+  }
+  try {
+    const prodDb = context.getProdAppDB()
+    const productionExists = await prodDb.exists()
+    type State = { automations: Automation[]; workspaceApps: WorkspaceApp[] }
+    let developmentState: State = { automations: [], workspaceApps: [] }
+    let productionState: State = { automations: [], workspaceApps: [] }
```
samwho:
```suggestion
    const developmentState: State = { automations: [], workspaceApps: [] }
    const productionState: State = { automations: [], workspaceApps: [] }
```

packages/frontend-core/src/api/types.ts
```diff
@@ -140,7 +141,8 @@ export type APIClient = BaseAPIClient &
   TemplateEndpoints &
   UserEndpoints &
   FeatureFlagEndpoints &
-  ViewEndpoints & {
+  ViewEndpoints &
+  DeploymentEndpoints & {
```
adrinr: Could we prefix it, the same way we use the ones below?

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -26,6 +36,11 @@ import { buildPublishFilter } from "./filters"
 // the max time we can wait for an invalidation to complete before considering it failed
 const MAX_PENDING_TIME_MS = 30 * 60000
 
+async function getAppMetadata(opts: { production: boolean }) {
```
adrinr: Could this live in the sdk, away from the controller layer?

packages/server/src/api/routes/tests/deploy.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+  let config = setup.getConfig()
+
+  afterAll(setup.afterAll)
+
+  beforeAll(async () => {
+    features.testutils.setFeatureFlags("*", {
+      WORKSPACE_APPS: true,
+    })
+    await config.init()
+  })
```
samwho: You'll want to store the function returned by `setFeatureFlags` and call it in `afterAll` to make sure we're cleaning up after ourselves.

packages/server/src/api/controllers/deploy/index.ts
```diff
@@ -148,6 +163,88 @@ export async function deploymentProgress(
   }
 }
 
+export async function publishStatus(ctx: UserCtx<void, PublishStatusResponse>) {
+  if (!(await features.isEnabled(FeatureFlag.WORKSPACE_APPS))) {
+    return (ctx.body = { automations: {}, workspaceApps: {} })
+  }
+  try {
+    const prodDb = context.getProdAppDB()
```
adrinr: It would be better to have this living in the sdk, extracting the koa objects and being able to easily write tests for it

packages/types/src/documents/app/app.ts
```diff
@@ -32,6 +32,8 @@ export interface App extends Document {
   updatedBy?: string
   pwa?: PWAManifest
   scripts?: AppScript[]
+  // stores a list of IDs and when they were last published (timestamp)
```
samwho: IDs for what?

packages/types/src/documents/app/app.ts
```diff
@@ -32,6 +32,8 @@ export interface App extends Document {
...
+  lastPublishedAt?: Record<string, string>
```
adrinr: This might be really confusing. Will this hold automations and screen changes?

samwho (commented):
Just a few minor comments.

adrinr (commented):
I like the idea, but I think this is missing the actual screen checks. The main issue will not be if a "workspace app" is modified, but if their screens are

packages/backend-core/src/db/Replication.ts
mike12345567: No - it is necessary to actually delete documents in production - I ran into this issue with deleting an automation in development, then publishing and the automation wouldn't be removed.

packages/types/src/documents/app/app.ts
mike12345567: Updated comment.

packages/types/src/documents/app/app.ts
mike12345567: It is simply a list of resources which have been published - any ID to a timestamp of when it was published - it should be generic, simply a map to look against.

adrinr: Should then be `resourcesPublishedAt`? Otherwise it looks like it is related to the app itself

adrinr (approved):
LGTM, just a last comment about namings

PR #16383, opened 2025-06-18
Enable workspace app on backend only

adrinr:
## Description
We have had the workspace app on development for a while, and it's growing quite a lot. We also, found out some issues around migrations:
1. Some of them ran accidentaly, so we have production apps in different states
2. When shipping, we don't currently have a way to enable the migration using the feature toggle, so we will have an intermidiate stage where the migration ran and the flag went on.

Because of this last point, we decided to run the migration for all and to get this working properly on the backend. When we decide to enable the toggle, that will only affect the frontend experience, as most of the backend will already be supporting the changes.

This means that:
1. We will run the migration for all, handling all test case scenarios (apps that already contained the migration, apps that did not, half-migrated...). This will create a snapshot of the appMetadata at the time.
2. We will be storing the screens linked to the existing workspace app (the user will not yet be aware of it)
3. On any change on the navigation (that will use appMetadata or workspaceApp depending on the flag status) we will save it to both appMetadata and workspaceApp (only if it's in the default workspaceApp). This will allow us to toggle on and off the flag without losing any data. All this logic will be removed when cleaning the flag.


package.json
```diff
@@ -42,7 +42,7 @@
     "build:apps": "DISABLE_V8_COMPILE_CACHE=1 yarn build --scope @budibase/server --scope @budibase/worker",
     "build:cli": "yarn build --scope @budibase/cli",
     "build:dev": "lerna run --stream prebuild && yarn nx run-many --target=build --output-style=dynamic --watch --preserveWatchOutput",
-    "check:types": "yarn check:dependencies && lerna run --concurrency 2 check:types",
+    "check:types": "yarn build && yarn check:dependencies && lerna run --concurrency 2 check:types",
```
mike12345567, line 45: :confetti_ball: thanks - I've ran into this a few times!

mike12345567 (approved):
LGTM - nice to see all the `workspaceAppId` being made required!

PR #16385, opened 2025-06-19
undo manifest mistakes from icon migration to fix divider component

andz-bb:
## Description
I ran a script to help with the icon migration. I missed a mistake it made that renamed the divider component, breaking it for existing users.

## Launchcontrol
Fix divider component

### DIFF
--- a/packages/client/manifest.json
+++ b/packages/client/manifest.json
@@ -753,9 +753,9 @@
       }
     ]
   },
-  "separator": {
-    "name": "Separator",
-    "description": "This component is a separator",
+  "divider": {
+    "name": "Divider",
+    "description": "A basic divider",
     "icon": "line-segments",
     "illegalChildren": ["section"],
     "size": {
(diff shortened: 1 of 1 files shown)


packages/client/manifest.json
```diff
@@ -1559,7 +1561,7 @@
   },
   "link": {
     "name": "Link",
-    "description": "This component is a link",
+    "description": "A basic link component for internal and external links",
```
melohagan, line 1564: NAB: Why basic?

melohagan (approved):
LGTM!

mikesealey (approved):

packages/client/manifest.json
andz-bb: I'm just reverting to what was there before. I think my cursor went a bit rogue and changed some of these descriptions in my original commit for #16218 without me realising

melohagan: All good then!

PR #16386, opened 2025-06-19
Icon fixes

aptkingston:
## Description
Fixes a few icon related issues:
- Fixes the BBUI button component not showing icons
- Fixes group icons in the user side panel
- Fixes icons in the BBUI table component header cells
- Fixes a few instances of edit icons incorrectly appearing in portal level user and group settings

### DIFF
--- a/packages/client/src/components/app/Button.svelte
+++ b/packages/client/src/components/app/Button.svelte
@@ -1,6 +1,5 @@
 <script>
   import { getContext } from "svelte"
-  import { Icon } from "@budibase/bbui"
   import "@spectrum-css/button/dist/index-vars.css"
 
   const { styleable, builderStore } = getContext("sdk")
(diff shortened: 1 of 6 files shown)


melohagan (approved):

packages/client/src/components/app/Button.svelte
```diff
@@ -66,7 +65,7 @@
     contenteditable={$component.editing && !icon}
   >
     {#if icon}
-      <Icon name={icon} {size} />
+      <i class="{icon} {size}" />
```
adrinr, line 68: Why not using the bbui one?

aptkingston: We use remix icons here in the client library - so it has to be unique. Hopefully at some stage we can replace this with phosphor.

PR #16391, opened 2025-06-20
Filter and Gridblock fixes

deanhannigan:
## Description
Some fixes around the `Filter` and `GridBlock` components. The `Filter` component was holding on to action context from the `GridBlock` provider. As this context isn't cleared when leaving the page, upon return there was a window in which the old context could be grabbed and effectively disconnect the Filter and Grid.

## Addresses
- Fix to ensure the `Filter` will update its copy of the `GridBlock` action callbacks when navigating to another page and back
- Adds a fix to ensure that when processing row conditions for a `GridBlock`, in then event there is no cached record of metadata for a row, a request to parse the metadata is made. When refreshing a page with a `GridBlock` and a `Filter` configured to persist it would retrigger the conditions checks but if there were new rows that hadn't been processed it would crash the builder.
- Switches the filter icon to the phosphor `plus-circle` icon

## Launchcontrol
Fixed to ensure filters use the correct action context on page load and stop Grid for throwing an error when processing metadata for conditions.

### DIFF
--- a/packages/client/src/components/app/filter/Filter.svelte
+++ b/packages/client/src/components/app/filter/Filter.svelte
@@ -112,8 +112,10 @@
     : ActionTypes.RemoveDataProviderQueryExtension
 
   // Register extension actions
-  $: addExtension = componentId ? getAction(componentId, addAction) : null
-  $: removeExtension = componentId ? getAction(componentId, removeAction) : null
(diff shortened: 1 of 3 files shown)


aptkingston (approved):
LGTM!

PR #16395, opened 2025-06-20
Add user count to usages

melohagan:
## Description
Include users in the usage page. Also updated the invite user modal to use an envelope icon instead of the browser.

## Addresses
- https://linear.app/budibase/issue/GRO-1113/add-user-count

## Screenshots
show users count

send_email_invites

## Launchcontrol
Add missing users count to the usage page


jvcalderon (approved):
LGTM

PR #16397, opened 2025-06-20
New Component: Radar chart

mikesealey:
## Description
This PR adds a new chart from the ApexCharts library - the Radar Chart.

The Radar Chart is particularly good at plotting things like View Counts - in the examples below we can see Workload of tickets assigned to developers, plotted by priority, and Units Sold By Region. 

## App Export
[Radar Chart-export-1750691348930.tar.gz](https://github.com/user-attachments/files/20867799/Radar.Chart-export-1750691348930.tar.gz)

## Screenshots
image

image

## Launchcontrol

_Adds a new component - Radar Chart_


packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,132 @@
...
+          const dataPointIndex = opts.dataPointIndex
+          const row = dataProvider.rows[dataPointIndex]
+
+          // Prevents clicking on the chart but not a specific point from triggering actions
+          if (dataPointIndex !== -1) {
+            handleSpokeClick(row)
+          }
+        },
+      },
+      foreColor: "#808080", // Numbers/Scale up the spoke
```
samwho, line 75: Does this look good in both dark and light mode?

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,132 @@
+<script>
```
samwho, line 1: We should convert this to TypeScript, and maybe introduce a lint rule that says all new components should be TypeScript.

samwho: I'm happy to help with this if you're not familiar with TS :)

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,132 @@
...
+      foreColor: "#808080", // Numbers/Scale up the spoke
+    },
+    yaxis: {
+      show: spokeLabels,
+    },
+  }
+
+  function handleSpokeClick(spoke) {
+    onClick?.({ spoke })
+  }
```
samwho, line 84: This is a bit trivial, could remove it in favour of calling `onClick` inline where this function is used.

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,132 @@
...
+
+  const getSeries = (dataProvider, valueColumns) => {
+    const rows = dataProvider.rows ?? []
+
+    return valueColumns.map(column => {
+      const data = rows.map(row => {
+        const value = row?.[column]
+
+        if (dataProvider?.schema?.[column]?.type === "datetime" && value) {
```
samwho, line 93: This expression (`dataProvider?.schema?.[column]?.type === "datetime"`) is used twice in this file and it a bit of a beast, probably worth making a helper for it:

```javascript
function columnIsDatetime(column) {
  return dataProvider?.schema?.[column]?.type === FieldType.DATETIME
}
```

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,132 @@
...
+          return Date.parse(value)
```
samwho, line 94: This can fail.

```
> Date.parse("fdakjlfdjsakl")
NaN
```

Are we correctly handling this possibility?

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,147 @@
...
+
+  const getLabels = (dataProvider, labelColumn, labelType) => {
+    const rows = dataProvider.rows ?? []
+
+    return rows.map(row => {
+      const value = row?.[labelColumn]
+
+      if (!["string", "number", "boolean"].includes(typeof value)) {
+        return ""
+      } else if (labelType === "datetime") {
```
samwho, line 116: Generally prefer to use enums in favour of naked strings in places like this. `FieldType.DATETIME` > `"datetime"`.

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,147 @@
...
+    })
+  }
+
+  const getLabels = (dataProvider, labelColumn, labelType) => {
+    const rows = dataProvider.rows ?? []
+
+    return rows.map(row => {
+      const value = row?.[labelColumn]
+
+      if (!["string", "number", "boolean"].includes(typeof value)) {
```
samwho, line 114: Are datetimes always strings? Can they ever be `Date` objects? If they can, this conditional branch will capture them and return the incorrect label.

packages/client/src/components/app/charts/RadarChart.svelte
```diff
@@ -0,0 +1,147 @@
...
+    })
+  }
+</script>
+
+<ApexChart {options} />
+<script>
+  options = {
+    chart: {
+      events: {
+        mounted: function(chartContext) {
```
samwho, line 130: Does this colour look good in both dark and light mode?

I believe we generally avoid hard coding colours, preferring instead to use colour vars so that they change appropriately based on the theme in use.

samwho (changes requested):

Issue #16379, opened 2025-06-18
App Action - Delay


mikesealey:
**Describe the feature request**
A configurable delay in the app-action flow that temporarily pauses the action-flow for an amount of time.

Related: https://github.com/Budibase/budibase/issues/16378

Maximebaker:
Reopen Please!!!

Issue #16378, opened 2025-06-18
Scroll to Field action does not work in On-Screen Load actions


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Cloud
  - Tenant ID: miketest1

**Describe the bug**
Using a "Scroll to field" app action in the "On screen load" action flow does not work.

**To Reproduce**
1. Using the sample data employees table, create a form block
2. Set an On Screen Load action to scroll to "End Date"

**Expected behavior**
Should scroll to the last field in the form

**Screenshots**
https://jam.dev/c/18cfb60f-8d70-4b27-9b11-a3bff0f56d64
  
**App Export**

[Scroll to field Kiirakii-export-1750238859109.tar.gz](https://github.com/user-attachments/files/20792985/Scroll.to.field.Kiirakii-export-1750238859109.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS Sequoia 15.5
 - Browser Chrome
 - Version 137.0.7151.104 (Official Build) (arm64)

**Additional context**
Issue raised on behalf of Discord user `kiirakii`

aptkingston:
I feel like I've seen this exact issue before... but anyway - it's because the screen has no way of knowing when all components are "ready", so it can't wait for them before executing the actions. The actions are basically just executed before the field is mounted and rendered, so it doesn't work.

I don't see a reliable fix for this. There are hacks like "wait 3 seconds before executing screen on load actions" but nothing that is guaranteed to work while also not impacting other use cases.

I think we should just remove "scroll to field" from the available "on load" actions. Either that or we add a "delay" action, so people can delay actions for a second and then scroll to the field.

mikesealey:
> I think we should just remove "scroll to field" from the available "on load" actions. Either that or we add a "delay" action, so people can delay actions for a second and then scroll to the field.

I suspected this would be the case - in the recreation, I think the "Show notification" step fires before the form properly loads in. Would you object to a new action with a configurable delay?

aptkingston:
Nope that was my suggestion above - an action to delay. I think that's the only solution to these sort of issues.

Maximebaker:
Reopen please!!

PR #16398, opened 2025-06-20
New app-action: delay

mikesealey:
## Description
Adds new app-action - Delay. This will help with an issue for the "Scroll to field" action when used in on-screen-load actions, which was likely caused by the form not being loaded by the time the on-screen-load actions have started running.

## Addresses
- https://github.com/Budibase/budibase/issues/16378
- https://github.com/Budibase/budibase/issues/16379

## App Export
[Delay-export-1751646728552.tar.gz](https://github.com/user-attachments/files/21061997/Delay-export-1751646728552.tar.gz)

## Screenshots
image

https://jam.dev/c/8b90af5d-979f-4d46-b3f2-e34e41babeea

## Launchcontrol

Adds new app-action: Delay


packages/client/src/utils/buttonActions.js
```diff
@@ -514,6 +514,10 @@ const rowActionHandler = async action => {
   })
 }
 
+const delayHandler = async action => {
+  return new Promise(resolve => setTimeout(resolve, action.parameters.duration))
```
aptkingston, line 531: This will work regardless of what value `duration` is, but we should probably sanity check it anyway. There's also no guarantee that it's a number - so I would use a normal number input for the setting instead of a drawer bindable input, and also do `action.parameters.duration || 0` here to account for null/undefined/empty string.

packages/client/src/utils/buttonActions.js
```diff
@@ -550,6 +581,8 @@ const handlerMap = {
   ["Download File"]: downloadFileHandler,
   ["Row Action"]: rowActionHandler,
   ["Copy To Clipboard"]: copyToClipboardHandler,
+  ["Delay"]: delayHandler,
+  ["Scroll To Field"]: scrollToFieldHandler,
```
aptkingston, line 605: If we're replacing the existing "Scroll To Field" implentation then we should remove the existing implementation. Currently I think it's only forms that do this, and have their own function called `handleScrollToField` which will be ignored now if we add this. So if this new implementation works fine for forms then let's remove the current one. We wouldn't need to expose a `ScrollTo` action at all from the real form component in this case.

We should also delete the `Scroll to Field` case from `getActionContextKey` in this file then too, as it is also redundant.

aptkingston: To be honest, it would actually be better to put this updated logic inside the `InnerForm` component, replacing the existing logic, instead of having it in this file - as it directly pertains to forms.

packages/client/src/utils/buttonActions.js
```diff
@@ -527,6 +527,57 @@ const rowActionHandler = async action => {
   })
 }
 
+const delayHandler = async action => {
+  return new Promise(resolve => setTimeout(resolve, action.parameters.duration))
```
deanhannigan, line 531: On a screen where a `setTimeout` has been executed, unless you clear it manually, it will stay in the background for however long it was configured for. If you navigate away from the page before it executes, it will still execute. In this scenario I noticed that _all pending steps_ from the previous screen would also execute.

All `delayHandler` promises/timeouts would have to be registered to the page, possibly during enrichment, then terminated manually before moving away from the page.

Additionally, the screen would need be aware of what screen the onload actions were meant for. If the screen doesn't match, the execution flow should be aborted. Let me know if that makes sense! Happy to hop on a call.

deanhannigan (changes requested):
Just gave this a rinse locally and it's pretty solid! 👍. Only ran into one issue, outlined below.

packages/client/src/utils/buttonActions.js
mikesealey: That's a great catch. I'll see what I can do.

Maximebaker:
Reopen please!!

mikesealey:
> Reopen please!!

I've reopened this for now, but it's likely to be easier to start with a fresh PR that achieves the same (and addresses the feedback). 

No timeline available on when this work might take place.

PR #16409, opened 2025-06-23
Fix automation name grouping

PClmnt:
## Description
Fixes an issue where with the renaming of the "App Action" trigger to "User Action" you could have multiple groups of the same trigger type. This just ensures that they all appear as one group in the automations side panel.

### DIFF
--- a/packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
+++ b/packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
@@ -49,7 +49,14 @@
       }
       // Otherwise group by trigger
       if (!category) {
-        category = auto.definition?.trigger?.name || "No Trigger"
+        let triggerName = auto.definition?.trigger?.name || "No Trigger"
+        // The definition was changed from "App Action" to "User action" in the definition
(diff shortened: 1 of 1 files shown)


packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
```diff
@@ -49,7 +49,13 @@
       }
       // Otherwise group by trigger
       if (!category) {
-        category = auto.definition?.trigger?.name || "No Trigger"
+        let triggerName = auto.definition?.trigger?.name || "No Trigger"
+        // The definition was changed from "App Action" to "User action" in the definition
```
samwho, line 53: Ah is this because automation definitions get saved in the DB alongside the automation, so still have the old name?

samwho (approved):

packages/builder/src/components/automation/AutomationPanel/AutomationPanel.svelte
PClmnt: Well, not the definition, but the saved automation is built from the definition, and since the update users will have been creating triggers with the "User Action" while still having pre-existing "App Action" ones.

PR #16412, opened 2025-06-24
Add ability to duplicate automation step

PClmnt:
## Description
Adds a button that allows the user to duplicate their automation step - it will simply add the step below the currently selected one. 

## Screenshots
image

https://linear.app/budibase/issue/BUDI-9331/add-the-ability-to-add-copy-and-paste-automation-steps


packages/builder/src/components/automation/AutomationBuilder/StepPanel.svelte
```diff
@@ -143,6 +144,29 @@
...
+          on:click={async () => {
+            if (!blockRef || !$memoBlock || isTrigger($memoBlock)) {
+              return
+            }
+            const duplicatedBlock = {
+              ...$memoBlock,
+              id: generate(),
+            }
+
+            await automationStore.actions.addBlockToAutomation(
```
deanhannigan, line 164: Would it be possible to add in something like the following when duplicating? 

```javascript
// This would duplicate an entire name, so `log 2` would become `log 2 2`. 
// Then merge it into `duplicatedBlock` before saving
const newName = getNewStepName($memoAutomation, duplicatedBlock)
```
We don't allow users to have duplicate names because it causes collisions when trying to build readable bindings down the line. If we don't change the name, we bypass the check

packages/builder/src/components/automation/AutomationBuilder/StepPanel.svelte
```diff
@@ -143,6 +144,29 @@
       >
         Delete
       </ActionButton>
+      {#if $memoBlock && !isBranchStep($memoBlock)}
```
deanhannigan, line 148: 👨‍🍳

deanhannigan (changes requested):
Gave this a rinse locally and only ran into one minor issue. Really nice update!

packages/builder/src/components/automation/AutomationBuilder/StepPanel.svelte
PClmnt: @deanhannigan Done :)

deanhannigan (approved):
LGTM!  👍

PR #16413, opened 2025-06-24
Treat citext as string

melohagan:
## Description
Importing **citext** was being treated as *Options* internal type, which was causing problems. Updated to treat **citext** as *Text* instead.

## Addresses
- https://linear.app/budibase/issue/GRO-1312/email-field-is-imported-as-dropdown-and-throws-error-when-trying-to

## Screensh
Screenshot 2025-06-24 at 11 16 02

*Correctly importing as Text instead of Options*

## Launchcontrol
Support **citext** data type in Postgres

### DIFF
--- a/packages/server/src/integration-test/postgres.spec.ts
+++ b/packages/server/src/integration-test/postgres.spec.ts
@@ -210,6 +210,31 @@ if (mainDescriptions.length) {
           expect(row.price).toBe("400.00")
         })
       })
+
+      describe("citext field", () => {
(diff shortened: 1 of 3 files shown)


mike12345567:
Would it be possible to add a test case for this? `packages/server/src/integration-test/postgres.spec.ts` you can use Knex to create a table with any column type and then try fetching the schema and seeing how Budibase treats it.

mike12345567 (approved):

Issue #16410, opened 2025-06-24
Unhide Column Toggle in Data View for Budibase DB Tables


digitizedcurrencynow:
**Checklist**
- [ ] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
**Title:** Unhide Column Toggle in Data View for Budibase DB Tables

**Description:**
Currently, when a user hides a column from the Data View in the internal Budibase DB, there is no way to unhide it via the UI. This makes it difficult to recover visibility of columns without workarounds like re-importing data or editing schema manually.

**Request:**
Please add a column visibility toggle (eye icon or checkbox list) in the Data View (not just Views or app screens), so users can manage hidden fields easily, similar to spreadsheet-style column selectors.

**Context:**
Tested in Budibase cloud version as of mid-2025. No "unhide" control appears once a column is hidden.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**
  
**App Export**
If possible - please attach an export of your budibase application for debugging/reproduction purposes.

**Desktop (please complete the following information):**

**Additional context**

aptkingston:
You shouldn't be able to hide columns in the main table view - so this is a bug. Hiding columns should only be available for views.

For dev reference - I remember someone made a change to this recently because it was always disabled. Looks like it went too far and now has enabled hiding columns even when you shouldn't be able to.

ConorWebb96:
Closing as it looks to be fixed.

PR #16414, opened 2025-06-24
Prevents columns in Data section from being hidden on base-tables

mikesealey:
## Description
This PR gives some context to HeaderCell to help us understand where it is being used.

HeaderCell being used as part of GridBlock as a Table Component will have a `componentId`, which means the end-user should have the option to hide a column (except the display column).

HeaderCell being used in the Data Section in a table should not be able to hide a column, as this is a difficult state to recover from.

## Addresses
- https://github.com/Budibase/budibase/issues/16410

## App Export

## Screenshots
Table as a Datasource in use in the Table Component
image

The same Table (datasource) in the Data Section
image

## Launchcontrol

_Prevents columns from being hidden in the datasource table_

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "version": "3.13.6",
+  "version": "3.13.8",
   "npmClient": "yarn",
   "concurrency": 20,
   "command": {
--- a/packages/backend-core/src/db/instrumentation.ts
+++ b/packages/backend-core/src/db/instrumentation.ts
@@ -183,7 +183,10 @@ export class DDInstrumentedDatabase implements Database {
     })
   }
 
-  dump(stream: Writable, opts?: DatabaseDumpOpts | undefined): Promise<any> {
+  dump(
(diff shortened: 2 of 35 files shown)


aptkingston (changes requested):
Rather than add domain logic concerning things like components, it would be better to add a discrete, reusable flag to control whether or not you can hide columns. We use this pattern for all other grid features - `canAddRows`, `canDeleteRows`, `canSaveSchema` etc.

I think we should add a new flag called `canHideColumns`, which defaults to true, and we explicitly set it to false from the data section.

You also don't need to pass this prop down all the way to the header cells - you can read it from the `config` store directly inside the HeaderCell component.

Rough steps to do this:
- Add a `canHideColumns` prop to the `Grid`, defaulting to `true`
- Add the property to the `BaseStoreProps` type too
- Add `canHideColumns` into the `props` store alongside all other props (Grid.svelte line 82)
- Update `HeaderCell` to check `$config.canHideColumns`
- Pass in `canHideColumns={false}` to the table grid in the data section

aptkingston (approved):
LGTM!

PR #16416, opened 2025-06-24
Fix an issue where automation mode wasn't being reset to Editor

PClmnt:
## Description
If you had the logs view selected and you selected a new automation, the editor / logs toggles would both be active.

### DIFF
--- a/packages/builder/src/stores/builder/automations.ts
+++ b/packages/builder/src/stores/builder/automations.ts
@@ -1610,6 +1610,10 @@ const automationActions = (store: AutomationStore) => ({
       delete state.testResults
       state.showTestModal = false
       delete state.selectedNodeId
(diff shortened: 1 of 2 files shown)


packages/builder/src/stores/builder/automations.ts
```diff
@@ -1610,6 +1610,10 @@ const automationActions = (store: AutomationStore) => ({
       delete state.testResults
       state.showTestModal = false
       delete state.selectedNodeId
+      state.showLogsPanel = false
+      state.showLogDetailsPanel = false
+      delete state.selectedLog
```
melohagan, line 1615: NAB: could group the delete statements and the boolean statements together.

melohagan (approved):
LGTM

mike12345567 (approved):

PR #16422, opened 2025-06-24
[Fixed] Use getExistingPlatformUsers to get email

jvcalderon:
## Description

Replaced `getUserDoc` with `getExistingPlatformUsers` in order to find by lower case email

`getUserDoc` is causing a problem with SSO when it tries to retrieve the emails with capital letters.

### DIFF
--- a/packages/worker/src/api/controllers/global/users.ts
+++ b/packages/worker/src/api/controllers/global/users.ts
@@ -113,7 +113,7 @@ export const addSsoSupport = async (
   try {
     const [userByEmail] = await users.getExistingPlatformUsers([email])
     if (!userByEmail) {
-      ctx.throw(400, "User not found")
+      ctx.throw(404, "Not Found")
     }
 
     await platform.users.addSsoUser(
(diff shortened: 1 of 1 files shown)


packages/worker/src/api/controllers/global/users.ts
```diff
@@ -112,10 +112,11 @@ export const addSsoSupport = async (
 ) => {
   const { email, ssoId } = ctx.request.body
   try {
-    // Status is changed to 404 from getUserDoc if user is not found
-    const userByEmail = (await platform.users.getUserDoc(
-      email
-    )) as PlatformUserByEmail
+    const [userByEmail] = await users.getExistingPlatformUsers([email])
```
melohagan, line 114: NAB: Could also just do a `email.toLowerCase()` ? But if `getExistingPlatformUsers` is preferable then fair enough

melohagan (approved):

packages/worker/src/api/controllers/global/users.ts
jvcalderon: I think `getExistingPlatformUsers` is preferable because we are storing documents with capital letters as Id. `getExistingPlatformUsers` will search into lowercase index.

melohagan: There is a performance trade-off here I think, in that the `getUserDoc` just straight up gets the document by **_id** and does not need to query

Issue #15964, opened 2025-04-16
Add customisable colours for Data Labels on all charts/chart blocks


mikesealey:
**Describe the feature request**
Issue/Feature request raised following a discussion raised by @dado-ca

Data labels are hard-coded to be white, but this isn't helpful when using light-themes and the labels spill over onto the white of the chart itself.

Add a colour-picker to select the colour of the text for this.

**Screenshots**

![Image](https://github.com/user-attachments/assets/74f1237d-b6e7-48d6-b7e7-da9370e81ab1)

PR #16425, opened 2025-06-24
Improves visibility of data labels in charts

mikesealey:
## Description
Users were having difficulty seeing Data Labels in charts that were wider than the bar on which they were displayed. After playing around with the ApexCharts library for a little while, Drop Shadow seems to be the way forward. It's already in use in Pie and Donut charts, and Area and Line charts use a different label setup.

## Addresses
Issue: https://github.com/Budibase/budibase/issues/15964
Discussion: https://github.com/Budibase/budibase/discussions/15957 

## App Export

## Screenshots
Before/Without Drop Shadow
image

With Drop Shadow
image

## Launchcontrol

_Improves visibility of Data Labels on Charts_

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "version": "3.12.20",
+  "version": "3.12.21",
   "npmClient": "yarn",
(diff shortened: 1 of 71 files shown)


ConorWebb96 (approved):
LGTM!

Issue #16418, opened 2025-06-24
Undo/Redo buttons don't work


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: miketest1

**Describe the bug**
Undo/Redo buttons don't work (though they flash when the creator makes a change

**To Reproduce**
1. Go to the builder, make a change, click undo

**Expected behaviour**
Should undo/redo

**Desktop (please complete the following information):**
 - OS: MacOS Sequioa 15.5
 - Browser Chrome
 - Version 137.0.7151.120 (Official Build) (arm64)

**Additional context**

andz-bb:
shouldn't they be on the left as well? 

<img src="https://uploads.linear.app/7ecfa297-cc77-4fdf-9117-fa5afdb8f908/11ac8948-7c5e-4d2c-8a43-fd823a80de21/04b090ce-4111-446c-9d8b-64000e721d62?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzdlY2ZhMjk3LWNjNzctNGZkZi05MTE3LWZhNWFmZGI4ZjkwOC8xMWFjODk0OC03YzVlLTRkMmMtOGE0My1mZDgyM2E4MGRlMjEvMDRiMDkwY2UtNDExMS00NDZjLTlkOGItNjQwMDBlNzIxZDYyIiwiaWF0IjoxNzUwNzczMDQ1LCJleHAiOjMzMzIxMzMzMDQ1fQ.jHIj_zR7f6QPZEptajM1bmlxMG-UKh-hkQGstVSnTBw " alt="image.png" width="902" data-linear-height="266" />

PR #16433, opened 2025-06-25
Fixing undo/redo

mike12345567:
## Description
Removing complete refresh on screen save/delete and sending the drawer-container down z-index to stop it blocking buttons.

Addresses: https://linear.app/budibase/issue/BUDI-9427/undoredo-buttons-dont-work
https://github.com/Budibase/budibase/issues/16418


adrinr (approved):
LGTM! As mentioned, this will need to be checked for the workspace apps flag on. I will cover it before the flag gets turned on

PR #16434, opened 2025-06-25
Allow skipping migration redirects

adrinr:
## Description
Fixing a bug around migrations. When deleting an app, if this needs to be migrated, the response on the deletion would contain the "should migrate" flag, that would cause a redirect to check if the migration ran succesfully. Because the app deletion, this will never succeed.

### Current issue

https://github.com/user-attachments/assets/a187fdb9-85aa-47bc-9937-d3ee0e2c6423


samwho (changes requested):
Could use a test

samwho (approved):

PR #16440, opened 2025-06-26
Fix: migration related issues

adrinr:
## Description
Fixing a couple of issues seen releated to the latest migration:
1. @PClmnt got a migrated app, but the routing view did not refresh as it should. Forcing the recreation of this view by changing the key of it
2. Some migrations are stuck while getting the error below. This should never happen, as the code is already checking if an app with this name already exists.... For this, as the flag has never been on for anybody, we can assume that if there is any workspace app, there is no need to create another one (as nobody could create any workspace app manually)
image

### DIFF
--- a/packages/backend-core/src/constants/db.ts
+++ b/packages/backend-core/src/constants/db.ts
@@ -21,7 +21,7 @@ export enum ViewName {
   USER_BY_EMAIL = "by_email2",
   BY_API_KEY = "by_api_key",
   LINK = "by_link",
-  ROUTING = "screen_routes",
+  ROUTING = "screen_routes_2",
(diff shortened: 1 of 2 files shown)


packages/backend-core/src/constants/db.ts
```diff
@@ -21,7 +21,7 @@ export enum ViewName {
   USER_BY_EMAIL = "by_email2",
   BY_API_KEY = "by_api_key",
   LINK = "by_link",
-  ROUTING = "screen_routes",
+  ROUTING = "screen_routes_2",
```
adrinr, line 24: This will force all the screen routes views to be "recreated", as they will be considered as a new one

PClmnt (approved):

PR #16441, opened 2025-06-26
Improve migration ux

adrinr:
## Description
Most migrations run in milliseconds, almost instantly. Even this, as with any migration we are having a redirect on the frontend, it looks much more than it is, as it needs to redirect, check, and go back.
With this PR, if a migration needs to run, the flow will be as follows:
1. Apply migrations
2. Wait for a couple of seconds to see if the migration ran successfully
3. If so, continue as if nothing happened. Otherwise, keep the current flow of redirect to the "updating" page


packages/server/src/appMigrations/index.ts
```diff
@@ -58,17 +57,43 @@ export async function checkMissingMigrations(
...
 export const isAppFullyMigrated = async (appId: string) => {
   const latestMigration = getLatestEnabledMigrationId()
   if (!latestMigration) {
     return true
   }
   const latestMigrationApplied = await getAppMigrationVersion(appId)
-  return latestMigrationApplied >= latestMigration
+  return (
+    getMigrationIndex(latestMigrationApplied) >=
+    getMigrationIndex(latestMigration)
```
aptkingston, line 97: Relying on `getMigrationIndex` will mean that if there's an unknown future migration, we'll never consider the app migrated and can potentially get into the broken migration loop we experienced recently. By relying on the chronologically sorted migration names we don't need to worry about that. So that's just a consideration.

adrinr: That was the initial approach, but we got a migration that did not follow the timestamp pattern and we got into similar issues... We should probably improve the resilience around this and add unit tests to ensure everything has the format it should.
And what do you mean, by `unknown future migration`?

samwho (approved):

packages/server/src/appMigrations/index.ts
adrinr: Also, the migrations are idempotent and they should be safe to rerun. Anyway, I want to revisit all this once everything is stable, to prevent running old migrations when not needed

PR #16446, opened 2025-06-26
Fix TS not liking a missing prop injected by routify

aptkingston:
## Description
Fix TS not liking a missing prop injected by routify.


deanhannigan (approved):
LGTM

PR #16449, opened 2025-06-26
Check for /embed url when matching workspace apps for routing

deanhannigan:
## Description
When fetching app routes, the `fromUrl` can either be `/app/my-app` or in the case of an embedded app `/embed/my-app`. This fix adds in the check for `/embed/` prefixed app urls when resolving workspace apps. Now embedded apps should resolved their routes correctly and display screens again.


adrinr (approved):
Nice finding! Confirmed that works as expected :)

PR #16455, opened 2025-06-26
Prune majority of phosphor icons and serve required files as static assets

aptkingston:
## Description
We noticed the client library bundle had ballooned to 38 MB of JS in a recent release. Phosphor icons was the culprit, as we were importing the entirety of the phosphor NPM package.

This PR extacts only the required parts of phosphor - the woff2s for the weights we use, and the required CSS (which I've also slightly compressed) - and serves them as static assets rather than bundling them inside JS.

This completely fixes the bundle bloating and also means these assets can be cached (and that cache can be shared between builder and client apps now too).


deanhannigan (approved):
LGTM!

adrinr (commented):
Could we not keep it as it was (npm) but marking it as external? It would be easier to upgrade to newer versions

aptkingston:
> Could we not keep it as it was (npm) but marking it as external? It would be easier to upgrade to newer versions

The primary issue with the bloating was that we can't bundle the woff2s inside the client (this caused 30+ MB of JS), so it needs to be served as static assets somehow and excluded from the bundle.

It would potentially be possible to instead copy the files from `node_modules` into the output assets directly as part of the builder vite config, if we were just very explicit about the paths. The main issue with that is that the CSS included in the phosphor bundle references multiple different font assets (woff, woff2, SVG etc), so I'm not sure how we can handle that if copying from `node_modules`, since we need to modify the CSS as we only want to extract and use the woff2s. We would have to write the style ourselves for the font families and override the default styles exported from phosphor, but that's not exactly very clean either.

So it's possible to consume it from npm but I don't think it's much cleaner than this approach - and this approach shrinks the bundle as much as possible. Upgrading wouldn't actually be too difficult here either - just grab the styles and woffs of the new version and override them.

adrinr (approved):

mike12345567 (approved):

PR #16458, opened 2025-06-27
Dev & pro migration conflicts

adrinr:
## Description
We designed the migration system assuming that:
1. Migrations will never run in parallel
2. A migration needs to handle idempotency, allowing it to run multiple times if something fails
This should allow us to ensure that, for example, we never create more than 1 workspace app.

Even so, we found an issue where this was not the case, as we were getting plenty of users with 2 workspace apps.
After some research, we found the issue. This only happened in published apps, due the following flow:
1. The prod app is accessed. Migration needs to run, the workspace app does not exists. Creates the workspace app with id `foo`
2. The dev app is accessed. Migration is not present, so it runs, creating workspace app with id `bar`.
3. The sync run from prod to dev, leaving dev with both `foo` and `bar` workspace apps.

This made us rethink the migration system, having the following flow:
1. If an app is not published, the migration will run only on the dev app (no issues for to date)
2. If an app is published, we follow the same flow, either accessing the dev or the prod app
a. For each migration, run the migration in prod
b. Sync prod to dev
c. Run migration to dev (as it is idempotent, only docs not handled by the application might be updated)


packages/server/src/appMigrations/migrations/tests/20250618162639_workspace_apps.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      await config.doInContext(appId, async () => {
+        await updateAppMigrationMetadata({
+          appId,
+          version: "",
+        })
+
+        // remove workspace apps to simulate it comes from an older installation
+        const workspaceApps = await sdk.workspaceApps.fetch()
+        const db = context.getAppDB()
+        await db.bulkRemove(workspaceApps)
```
samwho: Is it enough to delete them like this? In the future we may have other cleanup to do before delete them. Would it be worth introducing an SDK method for this?

packages/server/src/appMigrations/migrationsProcessor.ts
```diff
@@ -7,68 +7,104 @@ import {
...
-
-          const currentIndexMigration = migrations.findIndex(
-            m => m.id === currentVersion
-          )
+    console.log(`Acquiring app migration lock for "${appId}"`)
 
-          const pendingMigrations = migrations.slice(currentIndexMigration + 1)
+    const devAppId = db.getDevAppID(appId)
+    const prodAppId = db.getProdAppID(appId)
+    const isPublished = await db.dbExists(prodAppId)
```
samwho: What happens if migrations try to run when this returns false but then a publish happens during the migration? Do we need some other lock to prevent publishes happening during migrations?

packages/server/src/appMigrations/tests/migrationsProcessor.spec.ts
```diff
@@ -4,161 +4,282 @@ import {
...
+          executionOrder.push(`${db.name}-migration-3`)
+        },
+      },
+    ]
+
+    await runMigrations(testMigrations)
+
+    await expectMigrationVersion(testMigrations[2].id)
+
+    expect(executionOrder).toEqual([
```
samwho, line 97: This is a really nice test.

packages/server/src/appMigrations/migrations/tests/20250618162639_workspace_apps.spec.ts
adrinr: We have an sdk method for it. I did not use it exactly because of this, as we might have future changes that can break this (like not allowing to delete the last workspace app, for example). But better I'll use the SDK method and we will handle possible issues in the future, if they arise

packages/server/src/appMigrations/migrationsProcessor.ts
adrinr: I thought about this, but I thought it might be an edge case. But actually, implementing it has not been too hard, so I did it. I spent too much time trying to write a test for it, but it's pretty hard to do something nice around locks and everything, so for now I will leave this part untested... :/

samwho (approved):

PR #16464, opened 2025-06-29
Settings modal

deanhannigan:
## Description
All settings screens in Budibase have been consolidated into one common settings modal. Portal settings like user management, licensing, audit logs and AI are now available globally. So no more back and forth between the builder and the portal!

App settings have also been migrated and will appear in the modal when in the app builder.

Also:
- Licensing messaging and UX have been unified across all settings screens. 
- All routes are created using access aware config, so it should be a simple process to move pages around and manage the UX.
- Models now have unique internal ids to ensure nested content is parented correctly. Content relying on the nested modal context was attaching to the first modal and not the nested ones.

## Screenshots
Accessible in the top-right of the Portal or Builder.
Screenshot 2025-06-29 at 17 56 35

Portal user settings
Screenshot 2025-06-29 at 17 40 27

App settings
Screenshot 2025-06-29 at 17 45 49

## Launchcontrol
Move all settings pages for the portal and application into one global modal.


aptkingston (commented):
This is awesome @deanhannigan! Will attempt to do some code review, but I've just been using it for now and it works flawlessly - couldn't find any issues with it whatsoever.

The only tiny thing that I found that needs changed is that the command palette includes an option for navigate to settings, which no longer does anything. So we could just update that to also open settings.

The only minor things I think we could do with improving are just design related. First things first we should pull in master to get the Inter changes, then work from there. Looking at Figma, I think a few valuable tweaks would include:
- Tweaking some font sizes and weights, especially titles
- Using 1px borders
- Substantially increasing horizontal padding for the settings content (right side) panel
- Using gray-100 and 200 for the left and right panels respectively, instead of 50 and 100

mike12345567 (dismissed):
This looks great @deanhannigan - amazing job! The typescript improvements throughout are fantastic and it generally feels like a more cohesive settings experience.

Also great job pulling all of this together into a single place, no easy feat!

I'm going to approve this, but mention one thing, I know usually we don't worry too much about how features perform on mobile, but I do think it is useful to be able to edit settings from a mobile phone, like add a user, or reset a password - the default mobile experience isn't the best right now:
image

Perhaps if on mobile the side panel menu could be hidden then it might make it a bit easier to use?

packages/builder/src/components/integration/rest/AuthPicker.svelte
```diff
@@ -40,7 +41,7 @@
   }
 
   function addOAuth2Configuration() {
-    $goto(`/builder/workspace/${$appStore.appId}/settings/oauth2`)
+    bb.settings("/general/oauth2")
```
adrinr, line 44: Can this be called "goToSetting" or something like this? Ideally it would be nice to have it typed an an enum, to get sure we are always pointing to the right settings

packages/builder/src/components/settings/ModalSideBar.svelte
```diff
@@ -0,0 +1,231 @@
...
+  // Only dispatch when collapsed actually changes
+  $: {
+    if (initialized && previousCollapsed !== collapsed) {
+      dispatch("toggle", collapsed)
+    }
+    previousCollapsed = collapsed
+  }
+
+  export const unPin = () => {
+    // We need to ignore pointer events for a while since otherwise we would
```
adrinr, line 32: Is this needed? This is not triggered on collapsed change? Why is this reacting to a mouseenter?

packages/builder/src/components/settings/SettingsModal.svelte
```diff
@@ -0,0 +1,426 @@
...
+    isRouteHREF,
+    isSettingIcon,
+    type MatchedRoute,
+    type Route,
+  } from "@/types/routing"
+  import { beforeUrlChange, goto } from "@roxi/routify"
+  import ModalSideBar from "./ModalSideBar.svelte"
+  import SideNavLink from "@/pages/builder/workspace/[application]/_components/SideNav/SideNavLink.svelte"
+
+  export const show = () => {
```
adrinr: Usually is best to get this a one liner, so if show returns something, it will be passed over (the same that doing `return modal.show()`)

packages/builder/src/components/settings/SettingsModal.svelte
```diff
@@ -0,0 +1,426 @@
...
+  // Show/Hide the settings modal as required
+  $: modal, toggleSettings(open)
+
+  $: groupEntries = Object.entries(routesByGroup || {})
+
+  // Reset scroll when path changes
+  $: resetScroll(matchedRoute?.entry?.path)
+
+  // Pull out the default route
+  $: defaultRoute = $flattenedRoutes.find(r => r.path === "/general/info")
```
adrinr, line 60: Can we have this magic string in a constant or something to make sure this path is always valid?

packages/builder/src/pages/builder/workspace/[application]/_layout.workspace.svelte
```diff
@@ -0,0 +1,93 @@
+<script>
```
Comment (adrinr) [inline packages/builder/src/pages/builder/workspace/[application]/_layout.workspace.svelte:1]: I still don't see why this can't be left as _layout.svelte but wrapping it with the `key`

packages/builder/src/settings/pages/ai/AISettings.spec.js
```diff
@@ -1,10 +1,24 @@
 import { it, expect, describe, vi } from "vitest"
+import { writable } from "svelte/store"
 import AISettings from "./index.svelte"
 import { render, waitFor } from "@testing-library/svelte"
 import { admin, licensing, featureFlags } from "@/stores/portal"
 import { notifications } from "@budibase/bbui"
 import { API } from "@/api"
 
+const Hosting = {
```
adrinr: This already exists in types

packages/builder/src/types/routing.ts
```diff
@@ -0,0 +1,174 @@
...
+export const isRouteHREF = (href: string | RouteHREF): href is RouteHREF => {
+  return typeof href === "object" && href !== null && "url" in href
+}
+
+export const isSettingIcon = (
+  icon: string | RouteIcon | undefined
+): icon is RouteIcon => {
+  return (
+    typeof icon === "object" &&
+    icon !== null &&
```
adrinr: To handle undefineds
```suggestion
    icon &&
```

adrinr:
@claude

packages/builder/src/components/settings/ModalSideBar.svelte
deanhannigan: Yep, this gets triggered if the local storage `$pinned` value is set to false. The next time the user opens the settings menu `$pinned` is loaded from local storage as false which triggers `unpin`. This ensures the side bar component correctly reflects the closed state.

adrinr:
@claude

packages/builder/src/components/settings/SettingsModal.svelte
deanhannigan: This path is derived directly from the config itself and I thought that it would be more intuitive to format the requests as composable urls. For example, `/people/user/123` is a feasible path and its parameter of 123 can be processed and passed to the settings modal users page.

adrinr (approved):
LGTM! All detected issues sorted

PR #16472, opened 2025-06-30
Fix for Make connector where body wasn't being parsed correctly

PClmnt:
## Description
Fixes an issue where when having a single quote in your Make payload would result in a malformed payload being sent.
## Addresses
https://linear.app/budibase/issue/BUDI-9407/json-containing-apostrophes-gets-many-added-in

## Screenshots

Make working correctly previously broken payload:
image
## Launchcontrol
Fix an issue where the Make connector was sending malformed data.

### DIFF
--- a/packages/server/src/automations/steps/make.ts
+++ b/packages/server/src/automations/steps/make.ts
@@ -1,6 +1,10 @@
 import fetch from "node-fetch"
 import { getFetchResponse } from "./utils"
-import { ExternalAppStepOutputs, MakeIntegrationInputs } from "@budibase/types"
+import {
+  ExternalAppStepOutputs,
+  JSONValue,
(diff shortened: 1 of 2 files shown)


mike12345567 (approved):
LGTM - nice test cases.

packages/server/src/automations/steps/make.ts
```diff
@@ -9,9 +9,27 @@ export async function run({
 }): Promise<ExternalAppStepOutputs> {
   const { url, body } = inputs
 
-  let payload = {}
+  let payload: any = {}
```
mike12345567: if we could type this better than `any` that would be great, we actually have a type which we use when the type is unknown but it is JSON, `JSONValue` which can be imported from types.

PR #16473, opened 2025-07-01
"On input" actions

aptkingston:
## Description
This PR adds a setting to allow users to trigger "on change" actions by "input" events as well. This only makes sense for things like text inputs that fire input events separately to change events, so it's only added to these fields for now:
- Text field
- Number field
- Big int field
- Password field

image

Enabling this setting run the actions on "input" events **as well as** change events, not instead of.

Also fixed an unrelated issue which caused placeholders of `false` everywhere.

## Addresses
- https://linear.app/budibase/issue/BUDI-9455/on-input-actions

### DIFF
--- a/packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
+++ b/packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
@@ -13,7 +13,7 @@
   export let value: any = ""
   export let bindings: any[] = []
   export let title: string | undefined = undefined
-  export let placeholder: string | undefined = undefined
+  export let placeholder: string | false | undefined = undefined
   export let label: string | undefined = undefined
   export let disabled: boolean = false
   export let allowHBS: boolean = true
@@ -75,7 +75,7 @@
     on:change={event => onChange(event.detail)}
     on:blur={onBlur}
     on:scrollable={e => (scrollable = e.detail)}
-    {placeholder}
+    placeholder={placeholder || undefined}
     {updateOnChange}
     {autocomplete}
   >
@@ -96,7 +96,7 @@
   on:drawerHide={onDrawerHide}
   on:drawerShow
   bind:this={bindingDrawer}
-  title={title ?? placeholder ?? "Bindings"}
+  title={title || placeholder || "Bindings"}
   {forceModal}
 >
   <Button cta slot="buttons" on:click={saveBinding}>Save</Button>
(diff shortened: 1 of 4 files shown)


packages/client/src/components/app/forms/StringField.svelte
```diff
@@ -12,6 +12,7 @@
   export let defaultValue = ""
   export let align
   export let onChange
+  export let runOnInput = false
```
mike12345567, line 15: Why false, should it not be undefined?

packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
```diff
@@ -13,7 +13,7 @@
   export let value: any = ""
   export let bindings: any[] = []
   export let title: string | undefined = undefined
-  export let placeholder: string | undefined = undefined
+  export let placeholder: string | false | undefined = undefined
```
mike12345567, line 16: Why is false an option, can it not just be left undefined?

packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
```diff
@@ -75,7 +75,7 @@
     on:change={event => onChange(event.detail)}
     on:blur={onBlur}
     on:scrollable={e => (scrollable = e.detail)}
-    {placeholder}
+    placeholder={placeholder || undefined}
```
deanhannigan, line 78: 👨‍🍳

deanhannigan (approved):
LGTM! Ran through the supported field types and didn't run into any issues 👍

packages/client/src/components/app/forms/StringField.svelte
aptkingston: It's a boolean setting (you can check the manifest to confirm)  with a default value of false, so should be initialised to `false` here too.

packages/builder/src/components/common/bindings/DrawerBindableInput.svelte
aptkingston: I wish that was the case yea - we used to use an explicit `null` in some places to mean "don't show a placeholder option at all for selects", and someone changed that recently so that `false` for a placeholder now means that for selects. The problem is that there is a bunch of core logic that passes the placeholder into a variety of components, so we are passing `false` in to here at times too. So that's why it needs to support being a bool. Further down I just coerce this into `undefined` if it is false, so at least other components down the chain don't need to know about this weird typing.

PR #16474, opened 2025-07-01
Fix for embed issues in safari

deanhannigan:
## Description
Embedded apps weren't loading in Safari due to issues with CSP violations. The main culprit was the injection of iframes by svelte. Using the svelte bindings `bind:clientWidth`, `bind:clientHeight`, `bind:offsetWidth` or `bind:offsetHeight` will inject an iframe into the app, which it uses to monitor these values. This was more of an issue for newer apps because the bindings are used in our new default layout `GridContainer`. It also affects two deprecated components: `Table (Original)` and `Section`

The iframe svelte injects:
- `data:text/html,<script>onresize=function(){parent.postMessage(0,'*')}</script>`
- This both requires `data:` in the `frame-src` and the `<script>` element doesn't have a nonce. Both conflict with the CSP and Safari has issues loading the app.

This replaces all instances of the svelte bindings removes any need for changes to our CSP at all.

## Addresses
- https://linear.app/budibase/issue/BUDI-9439/ios-safari-browser-dont-show-an-iframes-content

## App Export
- Sample app from @ConorWebb96 containing `Table` and `Section` components.
- [everything-export-1751362759916.tar.gz](https://github.com/user-attachments/files/20997149/everything-export-1751362759916.tar.gz)

## Launchcontrol
Fix for embed CSP issues cause problems for Safari


samwho:
I had a look and don't see a whole lot that I can really comment on, so I'm going to rely on Cheeks here for review.

packages/bbui/src/Table/Table.svelte
```diff
@@ -330,16 +332,37 @@
...
   }
+
+  // Instead of svelte bind:offsetWidth
+  // Svelte injects an iframe causing issues with CSP, this avoids it
+  const setupResizeObserver = (element: HTMLElement) => {
+    const resizeObserver = new ResizeObserver(entries => {
+      if (!entries?.[0]) {
+        return
+      }
+      const bounds = entries[0].target.getBoundingClientRect()
```
aptkingston, line 343: Are you not able to use the bounds provided by the resize observer? Which would just be `entries[0].contentRect`.

aptkingston (commented):
Nice work! LGTM other than the one commented question.

packages/bbui/src/Table/Table.svelte
deanhannigan: The original svelte binding was for `offsetHeight` so it needed to include the border size too.`.contentRect` doesn't include the border.

adrinr (approved):
LGTM

packages/bbui/src/Table/Table.svelte
aptkingston: Tangent, but this is why `box-sizing: border-box` should always be used! But cool. I'd still rather just add 1 or 2px here then manually rather than `getBoundingClientRect()` for performance reasons, but it's totally an optimsation that's unnecessary.

aptkingston (approved):

PR #16478, opened 2025-07-01
[BUDI-9221] Add a new button to duplicate internal tables without their data.

samwho:
## Description

Introduces the ability to duplicate internal tables. This creates a copy of the table schema without the table data.

https://github.com/user-attachments/assets/2c9588ff-1afe-4d7b-ac8c-d8dc359d9854

## Addresses
- https://linear.app/budibase/issue/BUDI-9221/duplicate-internal-db-tables

## Launchcontrol

Adds the ability to easily duplicate internal tables.


packages/server/src/api/controllers/table/index.ts
```diff
@@ -263,3 +263,21 @@ export async function migrate(
 
   ctx.body = { message: `Column ${oldColumn} migrated.` }
 }
+
+export async function duplicate(ctx: UserCtx<void, SaveTableResponse>) {
+  const tableId = ctx.params.tableId as string
+  const table = await sdk.tables.getTable(tableId)
+
+  if (isExternalTable(table)) {
+    throw new HTTPError("Cannot duplicate external tables", 400)
```
adrinr: NIT: this should probably throw a 422 instead of a 400. The request is not "bad", it's just not valid

packages/server/src/api/controllers/table/index.ts
```diff
@@ -263,3 +263,21 @@ export async function migrate(
...
+  }
+
+  const duplicatedTable = await sdk.tables.duplicate(table, ctx.user._id)
+  await events.table.created(duplicatedTable)
```
adrinr: Should these events be dispatched within the SDK?

packages/server/src/sdk/app/tables/duplicate.ts
```diff
@@ -0,0 +1,64 @@
...
+import { getAllInternalTables } from "./getters"
+
+/**
+ * Generates a unique name for a duplicated table by checking existing table names
+ * and adding an incrementing number if needed
+ * @param originalName - The name of the original table
+ * @param existingTables - Array of existing tables to check against
+ * @returns A unique table name
+ */
+function generateUniqueName(
```
adrinr: We have this util already (or similar enough) in `packages/shared-core/src/helpers/duplicate.ts`. Can it be reused?

adrinr (changes requested):
Looks good, just a comment about reusing the "duplicate name" function

packages/server/src/api/controllers/table/index.ts
samwho: It looks like we more commonly dispatch these events in controllers than the sdk, but the sdk does feel more logical.

adrinr (approved):
LGTM!

PR #16479, opened 2025-07-01
Typing multiselect field component in client

mike12345567:
## Description
Typing the multiselect field in the client ahead of fixing a bug in this area.


packages/types/src/ui/fields/dataProvider.ts
```diff
@@ -0,0 +1,25 @@
+import { FieldSchema, UIDatasourceType, Row, JSONValue } from "../../"
+
+export type UIFieldDataProviderSchema = FieldSchema & {
+  optionColors?: Record<string, string>
+}
+
+export interface UIFieldDataProvider {
```
aptkingston: Might be worth renaming this to make it more clear what it actually is and let it be reused in future - this is the context provided by a data provider. So I'd probably go for something explicit like `DataProviderContext`.

packages/types/src/ui/fields/dataProvider.ts
```diff
@@ -0,0 +1,25 @@
...
+  rows: Row[]
+  datasource: {
+    label: string
+    tableId?: string
+    resourceId?: string
+    type: UIDatasourceType
+  }
+  schema: Record<string, UIFieldDataProviderSchema>
```
aptkingston: Schema here should just be `TableSchema | null` - that's what it's typed as in the real DataProvider component that exposes this. I'm sure that'll cause some other TS issue though lol but that's more correct.

packages/client/src/types/forms.ts
```diff
@@ -1,22 +1,20 @@
...
 export interface FormContext {
   formApi?: {
     registerField: (
       field: string,
       type: FieldType,
-      defaultValue: string | undefined,
+      defaultValue: string | string[] | undefined,
       disabled: boolean,
       readonly: boolean,
-      validation: FieldValidation | undefined,
```
adrinr, line 12: Not sure why I set this type. If this does not break any type, I guess it's ok. Checking my typing PR, I don't see the reason why I did that, but just worried that changing from a delegate to an object, nothing complains 😅

adrinr (approved):

packages/types/src/ui/fields/actions.ts
```diff
@@ -0,0 +1,7 @@
+import { Row } from "../../"
+
+export type UIFieldEventContext = { row: Row } | { value: any }
```
aptkingston: The param is always just `value` I think, unless I've missed something.

packages/client/src/types/forms.ts
mike12345567: Yeah, I don't think this is very well hooked up to be honest - I'd like to look at this more, but I have confirmed what is being passed in at the top of the stack and its not a delegate - the structure is based on runtime information.

aptkingston (commented):
You're brave, diving in to typing this stuff! Would have made a world of difference if we'd been TS from the start... But excellent job, looks good other than a few questions I've left 👌

Issue #16428, opened 2025-06-25
Default values of options picker broken


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: CLI/Devstack
  -  Budibase Version: 3.12.21
  - App Version: 3.12.21

**Describe the bug**
When setting default bindings of the Multi-Select picker with text, values can be added, but the original values cannot be un-selected.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a multi-select picker and populate it with some values
2. specify default values using text ie "1,3"
3. Default values will populate but cannot be unselected
(many examples in Jam and Export)

**Expected behavior**
Default values should be un-selectable

**Screenshots**
https://jam.dev/c/4a949199-faf7-47c6-88da-cb6abd6ef9ef
  
**App Export**
[Multi Select Picker Export-export-1750845348130.tar.gz](https://github.com/user-attachments/files/20901243/Multi.Select.Picker.Export-export-1750845348130.tar.gz)

**Desktop (please complete the following information):**
 - OS: MacOS Sequoia 15.5
 - Browser Chrome
 - Version 137.0.7151.120 (Official Build) (arm64)

**Additional context**
Issue raised by CSE on behalf of enterprise customer

mikesealey:
Additional context surrounding this bug: Switching from Dropdown to Checkboxes exhibits buggy behaviour. When using text to set the default values of the Multi-Select Picker (checkbox), the default values are not indicated in the component, though it's value reflects that they are. Those values can be checked, which adds another element to the array, but the original/default cannot ever be removed. (See JAM)

https://jam.dev/c/38dc26c5-d51d-46c6-ae85-53c22ae68c4a 

Expanded App-Export: 

[Multi Select Picker Export-export-1750852298256.tar.gz](https://github.com/user-attachments/files/20903016/Multi.Select.Picker.Export-export-1750852298256.tar.gz)

PR #16481, opened 2025-07-01
Fix de-selecting multi-select defaulted values

mike12345567:
## Description
Fixes an issue where de-selecting multi-select default values if the multi-select has values which are not of type string would not work. (Quite the mouthful!)

I considered a few options here but I think this can only really be fixed at the comparison stage as otherwise it would mean the multi-select field could only ever return strings - as far as I'm aware it is possible for it to return different types, although the default values will always be of type string.

## Addresses
- https://linear.app/budibase/issue/BUDI-9430/default-values-of-options-picker-broken
- https://github.com/Budibase/budibase/issues/16428


packages/bbui/src/Form/Core/Multiselect.svelte
```diff
@@ -81,7 +81,10 @@
   const makeToggleOption = (map: Record<string, boolean>, value: string[]) => {
     return (optionValue: string) => {
       if (map[optionValue]) {
-        const filtered = value.filter(option => option !== optionValue)
+        // comparison needs to take into account different types, always compare them as strings
+        const filtered = value.filter(
+          option => option.toString() !== optionValue.toString()
```
melohagan, line 96: Going by the types, these values should already be a string and not need the `toString` ?

```ts
value: string[]
```

```ts
return (optionValue: string)
```

Is it possible to prevent the wrong types being passed into the toggle function in the first place? Or update the type signatures to reflect the real possible types?

melohagan (changes requested):

packages/bbui/src/Form/Core/Multiselect.svelte
mike12345567: This would be possible if all of the code calling this was Typescript, but without another big push to convert the builder + client to Typescript fully it would be very difficult to protect against callers passing incorrect types.

I've updated the types to more accurately reflect how the multiselect is used today.

melohagan: No worries, thanks Mike!

melohagan (approved):

PR #16482, opened 2025-07-01
[BUDI-9240] Propagate OAuth2 information to external data connector and API Request automation steps

samwho:
## Description

We weren't currently propagating SSO user OAuth2 tokens through to External Data Connector or API Request steps, so those steps weren't able to make use of a user's SSO auth. This PR plumbs that data through and tests to make sure it works correctly, and refreshes correctly if required.

## Addresses
- https://linear.app/budibase/issue/BUDI-9240/make-currentuseroauth-available-through-app-trigger-automation-to-the

## Launchcontrol

SSO users' OAuth2 tokens now propagate through to External Data Connector and API Request steps if they are available (e.g. during App Action triggers).


packages/server/src/automations/steps/utils.ts
```diff
@@ -19,15 +19,15 @@ export async function getFetchResponse(fetched: any) {
...
 // throw added to them, so that controllers don't
 // throw a ctx.throw undefined when error occurs
-// opts can contain, body, params and version
+// opts can contain, body, params, version, and user
 export function buildCtx(
   appId: string,
   emitter?: ContextEmitter | null,
   opts: any = {}
 ) {
   const ctx: any = {
```
adrinr, line 28: Should we type these `any` to ensure that stuff like user is properly populated?

adrinr (approved):

packages/server/src/automations/steps/utils.ts
samwho: It's _really_ hard to type this correct. `UserCtx` is a really complex object.

PR #16484, opened 2025-07-02
[Fix] Autogenerate screens from data

adrinr:
## Description
Fix an issue around "generate screens" related to workspace apps. This PR types the failing component and fixes the issue itself (passing the default workspace app for now, we will see how it will work if we ship the full feature)

## Screenshots
https://github.com/user-attachments/assets/c5c7164f-7e1d-46d5-b68f-723b524755db

## Launchcontrol
Fixing an issue while autogenerating screens from the data section

### DIFF
--- a/packages/builder/src/components/backend/DataTable/buttons/grid/GridGenerateButton.svelte
+++ b/packages/builder/src/components/backend/DataTable/buttons/grid/GridGenerateButton.svelte
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
   import { ActionButton, ListItem, notifications } from "@budibase/bbui"
   import { getContext } from "svelte"
   import {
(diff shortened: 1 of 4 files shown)


packages/builder/src/components/backend/DataTable/buttons/grid/GridGenerateButton.svelte
```diff
@@ -165,7 +169,7 @@
   {/if}
 </DetailPopover>
 
-<CreateScreenModal bind:this={createScreenModal} />
+<CreateScreenModal bind:this={createScreenModal} {workspaceAppId} />
```
adrinr, line 174: This is the actual fix

PClmnt (approved):

PR #16485, opened 2025-07-02
Change email endpoint for account holder tenant users

melohagan:
## Description
Part of the change email process for account holders. 
Once the account email has been changed, the user doc needs to also be updated to reflect this, in CouchDB, for each cloud tenant the account holder has. 

## Addresses
- https://linear.app/budibase/issue/GRO-1332/add-a-restricted-endpoint-in-budibase-for-changing-user-email

## Screenshots
put

*After:*

platform user

first

second

### DIFF
--- a/packages/server/.gitignore
+++ b/packages/server/.gitignore
@@ -8,3 +8,4 @@ db/dev.db/
 dist
 coverage/
 watchtower-hook.json
+.claude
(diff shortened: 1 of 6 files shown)


packages/server/.gitignore
```diff
@@ -8,3 +8,4 @@ db/dev.db/
 dist
 coverage/
 watchtower-hook.json
+.claude
```
jvcalderon, line 11: NAB You can use a local `.gitignore`.

jvcalderon (approved):
LGTM

Issue #16011, opened 2025-04-25
Expand "Export to CSV" app-action functionality for users


mikesealey:
**Describe the feature request**
When the Table component fetches a large data source, it does so 100 rows at a time, fetching another 100 rows when the user scrolls down. 

This prevents a user from selecting all rows for a button to "Export Data" to export everything there as .CSV. Clicking the 'select-all' checkbox will only select rows that have already been returned (ie the first hundred), meaning the "Export Data" action will download a .CSV of only 100 rows

![Image](https://github.com/user-attachments/assets/779ae5ae-90d7-4de9-8a86-1ef9290607fa)

The Export Data action is also _not available_ for use with queries, which limits it's functionality where a business might want to use Budibase to generate reports based on parameters passed into queries.

Building the query for the table
![Image](https://github.com/user-attachments/assets/c9afaa55-6659-4d9c-aa5c-67700007cf6d)

Trying to export data from the table

![Image](https://github.com/user-attachments/assets/ec74a3d4-4e40-4e45-a693-914a8b7242ad)

It might be worth changing the "Export Data" action step to be configurable to fetch the whole query-result or the whole table, or specifically selected rows.

In summary - I'd like to be able to use queries in tables as part of an "Export data" action step, and configure how many rows are returned to the table to facilitate larger "Export Data" action steps.

![Image](https://github.com/user-attachments/assets/8370a4f3-a76b-41d2-9b81-7f4324375d5a)

melohagan:
Difficulty: Medium

To make Export Data useful for large tables we need to let the action request the full set (not just currently paged rows) and accept query-based data sources. That means adding options on the action step to choose between the entire table, a specific query result, or manually selected rows, and wiring the export runner to fetch rows directly from the API rather than relying on the client selection/per-page cache. Additional UI hints and tests should ensure paginated tables and queries can trigger exports without missing rows.

ConorWebb96:
Closing as done

PR #16486, opened 2025-07-02
16011 expand export to csv app action functionality for users

ConorWebb96:
## Description
Extended the exportDataHandler to support exporting data from the table component even when the data source isn’t an existing Budibase table.

This includes other data sources tables can use, such as custom queries, REST queries, etc. A new flag (checkbox) has been added, allowing users to enable this functionality, instead of being limited to fetched Budibase tables.

A utility function was added to sanitize and format the provided JSON data, converting it into CSV or JSON format. The resulting data is then stringified and downloaded as a CSV or JSON file, depending on the selected export type.

## Addresses
- `https://github.com/Budibase/budibase/issues/16011

## Screenshots
**New setting**
Screenshot 2025-07-02 at 11 19 56

## Launchcontrol
You can now export any data shown in your tables, not just Budibase tables, including data from DB custom or REST queries. A new checkbox lets you enable this feature, making it easier to download your table data as a CSV or JSON file.


packages/client/src/utils/buttonActions.js
```diff
@@ -347,47 +348,54 @@ const exportDataHandler = async action => {
...
 
-  // Handle legacy configs using the row selection store
+  // Handle no rows selected
   if (!rows?.length) {
-    const selection = rowSelectionStore.actions.getSelection(tableComponentId)
-    if (selection?.selectedRows?.length) {
-      rows = selection.selectedRows
-      tableId = selection.tableId
-    }
+    notificationStore.actions.error("Please select at least one row")
```
mike12345567, line 353: I think its safer for the caller to handle the notification/what to do if something goes wrong - its better for utilities to throw an error and then the component calling to handle that error with a notification.

I know this was previously handled here - might just be worth an investigation if it can be handled externally.

packages/shared-core/src/helpers/rowsHelper.ts
```diff
@@ -0,0 +1,67 @@
+import { Row } from "@budibase/types"
+
+export function cleanExportRows(
```
mike12345567: This function name isn't very clear, it appears this returns a bunch of rows in a format - its not just cleaning.

packages/client/src/utils/buttonActions.js
mike12345567: Ignore this previous comment - did an investigation - the button actions aren't exactly "utilities" - this is the handler for very dynamic behaviour - its not possible to separate the error handling.

mike12345567 (approved):

PR #16487, opened 2025-07-02
Fix disappearing dev app

adrinr:
## Description
Fixing the way we handle backup imports to make it more resilient. CouchDB does not provide us with a way to override a full database atomically, so we've implemented a more robust backup import strategy to prevent data loss and corruption during restore operations. This might be the issue with the existing dev app disappearing (as most users were paid ones, with access to backups).

Updated steps:
1. Run the import into a temporary database, rather than directly importing into the live dev application database
2. Only after a successful import into the temporary database, remove the existing app database
3. Replicate from the temporary database to the target database

This is still prone to errors, as there is no atomicity we can use for it.
Possible pitfalls:
1. If something fails on step 1 (it might be the heavier one), the app db does not get overridden.
2. Step 2 should not fail, and if it fails it might mean that the db was never deleted.
3. We are relying on native replication, but this is now the critical point.

Pro PR:
- https://github.com/Budibase/budibase-pro/pull/440/files

## Launchcontrol
Make restoring backups more resilient

### DIFF
--- a/packages/backend-core/src/db/instrumentation.ts
+++ b/packages/backend-core/src/db/instrumentation.ts
@@ -183,7 +183,10 @@ export class DDInstrumentedDatabase implements Database {
     })
   }
 
-  dump(stream: Writable, opts?: DatabaseDumpOpts | undefined): Promise<any> {
+  dump(
+    stream: Writable,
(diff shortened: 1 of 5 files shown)


mike12345567 (approved):
LGTM

PR #16491, opened 2025-07-02
Testing groups correctly handle apps that don't exist

mike12345567:
## Description
Some fixes for user groups, make sure that groups correctly adjust when apps they list do not exist.

Pro PR: https://github.com/Budibase/budibase-pro/pull/441


packages/worker/src/api/routes/global/tests/groups.spec.ts
```diff
@@ -294,6 +294,23 @@ describe("/api/global/groups", () => {
...
+
+      const { body: savedGroup } = await config.api.groups.saveGroup(group)
+      await config.api.groups.updateGroupApps(savedGroup._id, {
+        add: [{ appId: fakeAppId, roleId: "BASIC" }],
+      })
+      const { body: retrievedGroup } = await config.api.groups.find(
+        savedGroup._id
+      )
+
+      expect(Object.keys(retrievedGroup.roles)).not.toContain(fakeAppId)
```
adrinr, line 310: Could we add a "real" group, to make sure we only strip what is meant to be stripped?

adrinr (approved):

deanhannigan (approved):
LGTM! Verified that, upon deleting an app from the builder, the role entry no longer counts towards the users total apps in the app list 🎉

PR #16492, opened 2025-07-02
Allow Postgres enums to be searched

melohagan:
## Description
`ilike` only works for Postgres enum fields when you cast to text. 

## Addresses
- https://linear.app/budibase/issue/BUDI-8889/database-enum-breaks-built-in-filter-all-columns-all-views-including

## Screenshots
https://github.com/user-attachments/assets/d7b31470-988f-46c2-bd8a-9b4f4f6e0b34

## Launchcontrol
Allow Postgres enum fields to be searched


samwho (changes requested):
Could you add a test please? 🙏

melohagan:
Test added

samwho (approved):

PR #16493, opened 2025-07-02
[BUDI-9478] Fix exporting and importing a CSV to the same table.

samwho:
## Description

It should always succeed when you export data from a table and then import that data back into the same table (except for situations where constraints are violated).

This PR fixes 2 instances where this was not true:

- Datetime fields with `ignoreTimezone: true` set.
- Datetime fields with `ignoreTimezone: true` set _and_ where the database server is not UTC.

The first point was a fairly simple validation oversight, but the second point was quite tricky.

Some of our database integrations, when creating a datetime field without a timezone (e.g. the `schema.datetime({ useTz: false })` option in knex), will take the `Date` object they're given in their bindings and strip the timezone naively. No conversion from the source timezone. So 10:30AM in BST suddenly becomes 09:30AM. 

You can see this surface in production today if you create a MySQL server not using UTC, create a datetime field without a timezone, and write to it. What you choose in the UI is not what you will see after the row is saved.

How this PR fixes that is to parse all timezoneless dates it finds as if they were UTC (by appending a "Z" to them) and then making sure when we connect to each database, we set the time zone of that connection to UTC. This ensures consistency between Budibase and the database server.

In the case of Postgres, it didn't seem to be enough to just set the timezone. The `pg` driver mishandles dates, so it was necessary to convert dates to strings before sending them.

## Addresses
- https://linear.app/budibase/issue/BUDI-9478/csv-reimport-fails-on-date-field-with-ignore-timezones-enabled

## Launchcontrol

Fixes several bugs around datetime fields without timezones being written and read incorrectly throughout Budibase.


samwho:
Converted back to draft because some tests still failing.

packages/backend-core/src/sql/sql.ts
```diff
@@ -431,9 +431,9 @@ class InternalBuilder {
         return new Date(date)
       } else if (schema.ignoreTimezones) {
         if (isValidISODateString(input)) {
-          return new Date(input.trim())
```
adrinr, line 441: Why this? Should this not always be trimmed?

packages/backend-core/src/sql/utils.ts
```diff
@@ -14,9 +14,10 @@ import environment from "../environment"
 const DOUBLE_SEPARATOR = `${SEPARATOR}${SEPARATOR}`
 const ROW_ID_REGEX = /^\[.*]$/g
 const ENCODED_SPACE = encodeURIComponent(" ")
-const ISO_DATE_REGEX = /^\d{4}-\d{2}-\d{2}(?:T\d{2}:\d{2}:\d{2}(?:.\d{3})?Z)?$/
+const ISO_DATE_REGEX =
+  /^\d{4}-\d{2}-\d{2}(?:[T ]\d{2}:\d{2}:\d{2}(?:.\d{3})?Z)?$/
 const ISO_DATE_REGEX_NO_TIMEZONE =
-  /^\d{4}-\d{2}-\d{2}(?:T\d{2}:\d{2}:\d{2}(?:.\d{3})?)?$/
+  /^\d{4}-\d{2}-\d{2}(?:[T ]\d{2}:\d{2}:\d{2}(?:.\d{3})?)?$/
```
adrinr, line 20: Where does this come from? Is having a whitespace instead of `T` ISO compliant?

samwho: It's not but we've had several complaints from people about the strictness of the date format we expect. At one point dates were coming back from the DB in this format with a space not a T and I decided fuck it, this little bit of leniency does probably make sense.

packages/backend-core/src/sql/sql.ts
samwho: I removed these because if the string matches our valid date rexeges, it by definition cannot need trimming.

adrinr (approved):

Issue #16073, opened 2025-05-06
Limit user picker component to current app users


ConorWebb96:
**Describe the feature request**
As a user, I would like the ability to limit the users that can be selected/searched in the user picker form component to only those who have access to the specific app I am on. Currently, it pulls all users within the Budibase platform.

melohagan:
Difficulty: Medium

We need to tighten the user picker’s query so it only returns users linked to the current app (likely via the workspace/app membership tables) and surface a setting so builders can choose between app-specific or global users. That requires backend filtering, updating the API that backs the picker, and verifying the picker still works for inviting new users when the app-specific list is empty.

ConorWebb96:
Closing as this was done a while ago.

PR #16497, opened 2025-07-03
Added "App users" option to user pickers

deanhannigan:
## Description
User lookup fields currently return the global user list in the portal when loading options. A new `App users` prop has been added to the `BBReferenceField` and `BBReferenceFieldSingle` to allow users to restrict the options to the current apps users.

- `App users` defaults to `false`, reflecting the previous search behaviour, so no action is required.
- If you do wish to switch to an app only user search, just activate the toggle in the field settings.

## Addresses
- https://github.com/Budibase/budibase/issues/16073

## Screenshots
Screenshot 2025-07-03 at 11 39 42

## Launchcontrol
Added "App users" to User pickers. This changes user search from global to app only.


packages/client/src/components/app/forms/BBReferenceField.svelte
```diff
@@ -27,15 +28,21 @@
...
-  {type}
-  datasourceType={"user"}
-  primaryDisplay={"email"}
-  defaultValue={updatedDefaultValue}
-  {defaultRows}
-  {multi}
-  on:rows
-/>
+<!-- Trigger a re-render when changing the appUsersOnly but only in the builder -->
+{#key $builderStore.inBuilder ? `active-${appUsersOnly}` : "inactive"}
```
aptkingston: Could we not just make the relationship field be reactive to the `appUsersOnly` prop? That way we don't need any changes to this file at all, and I think that's a more robust solution since right now it's only when the relationship field is wrapped in this BB reference field that it's reactive.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -139,14 +147,14 @@
...
-    const datasource =
+    const datasource: DataFetchDatasource =
       dsType === "table"
         ? {
             type: dsType,
             tableId: linkedTableId!,
           }
         : {
-            type: dsType,
+            type: appUsersOnly ? "table" : dsType,
```
aptkingston: As long as pass `appUsersOnly` in to `createFetch` here as a param, it'll be reactive.

packages/client/src/components/app/forms/RelationshipField.svelte
```diff
@@ -200,6 +208,14 @@
     optionsMap = optionsMap
   }
 
+  const parseId = (id: string) => {
+    // Normalise app table users to the global format.
+    if (datasourceType === "user" && appUsersOnly) {
```
aptkingston: This function consumes both `datasourceType` and `appUsersOnly` but is not reactive to either, so changes in this variables wouldn't invalidate this function. This may be alright in practice, since other upstream functions may get invalidated which will in turn call this. But just something to be aware of.

aptkingston (commented):
I think we can simplify how we consume `appUsersOnly` to avoid changing the BB reference field, but otherwise looks good to me!

packages/client/src/components/app/forms/BBReferenceField.svelte
deanhannigan: Yep, agreed. I've shifted down to the RelationshipField

packages/client/src/components/app/forms/RelationshipField.svelte
deanhannigan: `processOptions` and `loadMissingOptions` will both now get triggered when either `appUsersOnly` or `datasourceType` are changed.

melohagan (approved):
LGTM

The only thought I have is that the term `App users` is confusing. In the docs, an App user is defined at the workspace level and just means any user that can use any app: https://docs.budibase.com/docs/user-roles#access-permission

In this case we're talking about App users of the current app.

So I would change the label to `Current app users` or `Local app users`. Or even just being very explicit albeit wordy: 

`Only include users that have access to this app [x] `

ConorWebb96 (approved):
I just re-ran through this and it seems to still be working. The only thing that might be an issue is the user columns on the tables. These still allow you to select any user in your instance.

deanhannigan:
> I just re-ran through this and it seems to still be working. The only thing that might be an issue is the user columns on the tables. These still allow you to select any user in your instance.

Yep, that's good shout. We should extend this behaviour into other components that surface the `User` type fields too!

PR #16498, opened 2025-07-03
Fix for budi backups tool on WSL

andz-bb:
## Description
Users running the `budi backups` CLI tool on WSL have run into an issue where log files from the process are locked and can't be cleaned up at the end of the process. This led to an EBUSY error shown below rather than confirmation that the backup or restore has completed.

image

This PR fixes this by [closing the connections to PouchDB](https://pouchdb.com/api.html#close_database) after the CouchDB part of the import/export is complete.

## Screenshots
Backup completing as expected running the updated version of the CLI
image

## Context
Raised by an enterprise customer and a community member [on this discord thread](https://discord.com/channels/733030666647765003/1377670799770517585/1377670799770517585)

## Launchcontrol
fix Budibase instance backups for Windows users


mike12345567 (approved):
nice fine - surprising this is an issue but nice one!

packages/cli/src/backups/index.ts
```diff
@@ -104,9 +108,9 @@ async function importBackup(opts: BackupOpts) {
 
 async function pickOne(opts: BackupOpts) {
   if (opts["import"]) {
-    return importBackup(opts)
+    return await importBackup(opts)
```
mike12345567: NAB: Just letting you know, this isn't technically necessary - doing this means that error handling happens within this function so:
```js
async function something() {
  try {
    return someOtherAsyncFunction()
  } catch (err)...
}
```
Versus:
```js
async function something() {
  try {
    return await someOtherAsyncFunction()
  } catch (err)...
}
```
The first function the try catch will never be triggered because the calling function handles any errors that erupt from the `someOtherAsyncFunction` but in the second example the error will be caught within the function. If the calls are not wrapped in any kind of error handling then it is going to bubble up anyway and the await does nothing.

andz-bb: thanks, will just take the awaits out to clean it up a bit

Issue #16461, opened 2025-06-27
My app disappeared after trying to revert


cnrodrig:
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: demouc

**Describe the bug**
I have an app "Cierre Procesos", published and working ok. I tried the last update and because a feature stop working i tried to revert the update... a strange error appeared (like error trying to revert) and then the app is nowhere to be seen...

The strange thing is that in Account -> Usage it shows 2 apps 

**To Reproduce**
1. Go to Account -> Usage
2. Check the number of Apps
3. Go to the builder home
4. Only one app is displayed

**Expected behavior**
It should show two apps on the builder

**Screenshots**

![Image](https://github.com/user-attachments/assets/9943ed75-9750-4993-aa4d-4a31236fe50a)

![Image](https://github.com/user-attachments/assets/dc728580-7989-4633-843f-216b00ed3e2c)

  

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 137.0.7151.122

adrinr:
@cnrodrig we are investigating the issue that caused your app to disappear. Meanwhile, we restored your app so it should be back in there. We will keep you update it about the final fix 🙏🏽

PR #16499, opened 2025-07-03
Resilient revert

adrinr:
## Description
Using a system of queues to process reverts. If something fails (as it happened in the past) we might get into an inconsistent state where the dev app disappears. Using these queues, the system should eventually fix itself.
For this PR I added a generic queue system that might be reused in the future, ensuring retries and properly tested

## Addresses
- #16461


packages/backend-core/src/queue/inMemoryQueue.ts
```diff
@@ -186,7 +208,25 @@ export class InMemoryQueue<T = any> implements Partial<Queue<T>> {
     } else {
       pushMessage()
     }
-    return { id: jobId } as any
+    return {
+      id: jobId,
+      finished: () =>
+        new Promise((resolve, reject) => {
+          this._emitter.on("error", (job, error) => {
```
samwho: Are these going to build up over time as the tests run? Could we remove the event handler when we find the thing we're looking for?

packages/server/src/api/controllers/dev.ts
```diff
@@ -89,51 +82,24 @@ export async function clearLock(ctx: Ctx<void, ClearDevLockResponse>) {
...
-      message: "Reverted changes successfully.",
+      status: "applied",
     }
-    await events.app.reverted(appDoc)
-  } catch (err) {
-    ctx.throw(400, `Unable to revert. ${err}`)
-  } finally {
-    await replication.close()
+  } catch (err: any) {
+    ctx.throw(500, err.message)
```
samwho: Could we let this bubble instead? The error handling middleware will handle it and, in tests, attach a stack trace to make it easier to debug the problem.

packages/server/src/sdk/dev/devRevertProcessor.ts
```diff
@@ -0,0 +1,97 @@
...
+  ): Promise<{ message: string }> {
+    const { appId } = data
+    const productionAppId = dbCore.getProdAppID(appId)
+
+    // App must have been deployed first
+    try {
+      const db = context.getProdAppDB({ skip_setup: true })
+      const exists = await db.exists()
+      if (!exists) {
+        throw new Error("App must be deployed to be reverted.")
```
samwho: Does this count as an error that ends up getting retried? If it does that feels a bit wasteful, I think this is a situation where we don't want to retry because we know the revert can't succeed.

adrinr:
Really good points, @samwho. Everything updated

samwho (approved):

PR #16502, opened 2025-07-03
CI check for client library size

mike12345567:
## Description
Adding a check to make sure the client library build is less than 8MB.


aptkingston (approved):
LGTM!

Issue #16504, opened 2025-07-04
Disabled menu options look no different to enabled options


mikesealey:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**

- Self
  - Method: Dev Stack/CLI
  - Budibase Version: 3.13.8
  - App Version: 3.13.8

**Describe the bug**
In the data section, in the column kebab-menu, options that are disabled are not greyed out. Same for the Table Component.

**To Reproduce**
1. Go to the Employees table in the sample data, and open the kebab-menu for any of the columns

1. Go to the table component and open the kebab menu in the column of a table component.

**Expected behavior**
Disabled menu options should be greyed out.

**Screenshots**
Table Component - Cannot hide a display column. The option in the menu isn't greyed out (though it is still disabled)

![Image](https://github.com/user-attachments/assets/8b3b35a3-35db-40af-8f5e-dc5b6a1f3abe)

Data section table

![Image](https://github.com/user-attachments/assets/b3508fc3-5e38-47bc-a219-2a36523ac5f1)

**Desktop (please complete the following information):**
 - OS: MacOS Sequoia 15.5
 - Browser Chrome
 - Version 138.0.7204.50 (Official Build) (arm64)

PR #16505, opened 2025-07-04
fixes styling issue on disabled menu items

mikesealey:
## Description
A slight UI issue has been introduced somewhere along the line in the kebab-menu of column header cells - disabled options were not greyed out.

## Addresses
- https://github.com/Budibase/budibase/issues/16504

## App Export

## Screenshots
Disabled menu items were not greyed out - for the display column "Use as display column" as well as "move left", "move right", "duplicate column" and "edit column" are disabled, and so should be greyed out
Before (Data Section):
image

Before (Table Component):
image

After (Data Section):
image

After: (Table Component):
image

## Launchcontrol

_fixes UI issue in column's kebab menu_


packages/bbui/src/Menu/Item.svelte
```diff
@@ -94,8 +94,8 @@
     display: grid;
     place-items: center;
   }
-  .is-disabled .key {
-    color: var(--spectrum-global-color-gray-600);
+  .is-disabled span {
```
deanhannigan: Try to avoid `!important` as much as possible. It has a habit of throwing a spanner in the works down the line. In this case, the more specific rule takes precedence now.

  ```css
  .is-disabled .spectrum-Menu-itemLabel {
    color: var(--spectrum-global-color-gray-600);
  }
  .spectrum-Menu-itemLabel {
    color: var(--spectrum-global-color-gray-900);
  }
  ```

deanhannigan (approved):
LGTM!

PR #16515, opened 2025-07-07
Yaml config email templates

deanhannigan:
## Description
Email templates can now also be loaded via a yaml template document.

- A new environment variable added to the worker for loading yaml templates: `EMAIL_TEMPLATE_PATH`
- On load, the worker will check if there is content in this location, parse it, then update any templates that differ to those in the portal.
- All current email types are supported: `base`, `password_recovery`, `invitation`, `welcome`, `custom`
- Also, two new buttons have been added to the Email settings page: `Export default` downloads a fresh set of yaml config and `Export` will download your current templates, including any edits. `Export` will not appear if you have never edited your templates.

Sample default doc: 
[bb_default_email_templates.yaml.zip](https://github.com/user-attachments/files/21099256/bb_default_email_templates.yaml.zip)

## Addresses
- https://github.com/Budibase/budibase/issues/15950

## Screenshots
A download section has been added to email settings
Screenshot 2025-07-07 at 09 27 02

## Launchcontrol
Email templates now configurable via yaml config.


packages/worker/src/api/controllers/global/templates.ts
```diff
@@ -93,3 +103,55 @@ export async function destroy(
...
   ctx.body = { message: `Template ${ctx.params.id} deleted.` }
 }
+
+function templatesToYaml(
+  templates: Array<{ contents: string; purpose: string }>
+): string {
+  const doc = new yaml.Document()
+
+  // Build the template object
+  const templatesObj: any = {}
```
samwho:
```suggestion
  const templatesObj: Record<string, yaml.Scalar> = {}
```

Does this work?

packages/worker/src/api/controllers/global/templates.ts
```diff
@@ -93,3 +103,55 @@ export async function destroy(
...
+      // Combine header and content
+      const finalYaml = header + yamlContent
+      ctx.attachment(
+        `${type === "custom" ? "bb_email_templates" : "bb_default_email_templates"}.yaml`
+      )
+      const path = join(objectStore.budibaseTempDir(), v4())
+      fs.writeFileSync(path, finalYaml)
+      ctx.body = fs.createReadStream(path)
+    } catch (err: any) {
+      console.log("Could not download email templates ", err.message)
```
samwho: I'd let this error bubble up. Swallowing it will lead to confusion.

packages/worker/src/constants/templates/index.ts
```diff
@@ -19,6 +20,76 @@ export const EmailTemplates = {
   [EmailTemplatePurpose.CORE]: readStaticFile(join(__dirname, "core.hbs")),
 }
 
+const normaliseTemplateContent = (content: string) => {
+  return content.replace(/>\s+</g, "><").replace(/\s+/g, " ").trim()
```
samwho, line 24: It's not obvious to me what the purpose of this is.

packages/worker/src/constants/templates/index.ts
```diff
@@ -19,6 +20,76 @@ export const EmailTemplates = {
...
+}
+
+export async function loadTemplateConfig(pathTo: string) {
+  let templateConfig: any
```
samwho: Is there a more concrete type we could use for this?

packages/worker/src/constants/templates/index.ts
```diff
@@ -19,6 +20,76 @@ export const EmailTemplates = {
...
+          }
+          return acc
+        },
+        []
+      )
+
+      if (updates.length) {
+        const updateList = updates.map(u => u.purpose)
+        const db = tenancy.getGlobalDB()
+        await db.bulkDocs([...updates])
```
samwho: Do we need to copy `updates` here?

packages/worker/src/constants/templates/index.ts
```diff
@@ -19,6 +20,76 @@ export const EmailTemplates = {
...
+        console.log(`Email templates updated: ${updateList.join(",")}`)
+      } else {
+        console.log(`Email templates unchanged`)
+      }
+    } catch (e: any) {
+      console.log("Error persisting email templates", e.message)
```
samwho, line 90: It's less obvious to me what we should do with these errors, as the code is run on server start. Logging is likely to go unnoticed by the user, but crashing the server on startup isn't ideal either. 🤔

samwho:
We should also have some tests in `packages/worker` that validate this template parsing on startup works as expected.

packages/worker/src/constants/templates/index.ts
deanhannigan: All this does is remove all white space between html tags in a template. I was getting false positives comparing the yaml values to the template content returned by the server, even though the content was the same.

samwho: In that case I'd call it something like `removeWhitespaceBetweenTags` instead of `normaliseTemplateContent`

packages/worker/src/constants/templates/index.ts
deanhannigan: Yep. My thoughts are that if they're loading email yaml templates from an environment path location they'll be checking the logs when updates don't appear in the UI.

packages/worker/src/constants/templates/tests/templates.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    await config.beforeAll()
+  })
+
+  afterAll(async () => {
+    await config.afterAll()
+  })
+
+  it("should ignore an invalid path", async () => {
+    const consoleSpy = jest.spyOn(console, "log").mockImplementation(() => {})
```
samwho: Could extract this out into a beforeEach and the `consoleSpy.mockRestore()` into an afterEach as they're used in all tests as far as I can tell. Is minor, though, ignore if you don't feel like it.

samwho (approved):
Cheers for those tests, Dean! 🙏

packages/worker/src/constants/templates/tests/templates.spec.ts
deanhannigan: Yep, done! Thanks for taking a look Sam!

PR #16521, opened 2025-07-07
[BUDI-9364] Give sorting in views better labels based on field type.

samwho:
## Description

It used to be that sorting in views always showed A-Z or Z-A regardless of the column type. This PR gives more contextually relevant labels to these options.

https://github.com/user-attachments/assets/309f887e-fc84-43ae-9fc9-0a0aa492961e

## Addresses
- https://linear.app/budibase/issue/BUDI-9364/sorting-in-views-does-not-have-context-of-a-columns-data-type

## Launchcontrol

Sorting options for columns in views now have contextually relevant labels, e.g. "low to high" for numeric columns and "oldest to newest" for dates.


deanhannigan (approved):
LGTM!

PR #16528, opened 2025-07-08
[BUDI-9355] Bulk add existing users to groups via csv upload

samwho:
## Description

Adds the ability to bulk add users to a group using a CSV with an `email` column.

https://github.com/user-attachments/assets/b2ca5d8e-7177-452e-b498-6f58c85d59c1

Pro PR: https://github.com/Budibase/budibase-pro/pull/443

## Addresses
- https://linear.app/budibase/issue/BUDI-9355/bulk-add-existing-users-to-groups-via-csv-upload

## Launchcontrol

Adds the ability to bulk add users to a group using a CSV with an `email` column.


packages/builder/src/pages/builder/portal/users/groups/_components/BulkAddUsersModal.svelte
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  const reset = () => {
+    selectedFile = undefined
+    results = undefined
+  }
+</script>
+
+<ModalContent
+  title="Bulk Add Users via CSV"
```
adrinr: This UI does not look too much of BB UX. Should we just say "import" as we do in other places?

adrinr (commented):
If I get this right, this will not add new users, just add existing ones. Is there a reason why? Can we not use the same logic we use everywhere else?

packages/builder/src/pages/builder/portal/users/groups/_components/BulkAddUsersModal.svelte
samwho: "Import" doesn't feel _quite_ right. I don't particularly mind, though. This felt somewhat descriptive. Happy to change it.

samwho:
I was going off what the ticket requested. @ConorWebb96 @mike12345567 what do you think to Adri's question above?

packages/builder/src/pages/builder/portal/users/groups/_components/BulkAddUsersModal.svelte
adrinr: This is my point. Why is this not doing an import, as we do in the actual user section. Either this or we make it more explicit, as it might be confusing otherwise

ConorWebb96: Are we saying it should do both?

- Create new users if they don't exist in the CSV and assign them to the group
- Assign existing users inside the CSV to the group as well

If yes, I don't see why not.

The user here specifically asked for a way to bulk add existing users, but I can also see it being used to add new users.

samwho: Do we not already have a way to bulk add users somewhere?

It doesn't make sense to me for a bulk add users feature to exist inside of a specific group's UI.

samwho: CleanShot 2025-07-09 at 14 33 06@2x

ConorWebb96: I think the issue is the terminology. "Bulk add" implies, to me, that you can create new users.

Would `Bulk assign` work better here as the text for the button?

I'm not a designer, but the icon on the button seems strange. We don't have icons for any other buttons in this area.

- Could shorten the text of the button to  just "Bulk assign"
- Drop the button icon
- Copy the UI for the modal for the original import, big upload button etc. keep it consistent.

I would probably get @joebudi input for this design, though.

mike12345567:
> I was going off what the ticket requested. @ConorWebb96 @mike12345567 what do you think to Adri's question above?

The feature should only add existing users to the group - actually adding new users to the system while editing the group would be a new feature and beyond the scope of this ticket.

Conor's comment is correct to me, `Bulk assign` would be the correct terminology to use here when targeting a group.

samwho:
CleanShot 2025-07-16 at 16 57 01@2x

CleanShot 2025-07-16 at 17 05 49@2x

@mike12345567 @ConorWebb96 how dis?

mike12345567 (approved):
LGTM!

PR #16531, opened 2025-07-08
Fix issue with collect blocks in button actions

PClmnt:
## Description
Previously we weren't correctly discovering Collect blocks that existed in branch children. That meant that the Button Action bindings weren't being updated properly. 

(also a tiny fix to remove a gate on the branch separator that shouldn't be there, was causing a slight visual discrepancy). 

## Addresses
https://linear.app/budibase/issue/BUDI-9116/the-collect-data-step-doesnt-work-if-its-at-the-end-of-a-branch-step.

## Launchcontrol
- Fix an issue with automation collect block bindings in automations

### DIFF
--- a/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/TriggerAutomation.svelte
+++ b/packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/TriggerAutomation.svelte
@@ -2,8 +2,8 @@
   import { Select, Label, Input, Checkbox, Icon, Body } from "@budibase/bbui"
   import { automationStore } from "@/stores/builder"
   import DrawerBindableInput from "@/components/common/bindings/DrawerBindableInput.svelte"
-  import { TriggerStepID, ActionStepID } from "@/constants/backend/automations"
-  import { AutomationActionStepId } from "@budibase/types"
+  import { TriggerStepID } from "@/constants/backend/automations"
+  import { sdk as coreSdk } from "@budibase/shared-core"
 
   export let parameters = {}
   export let bindings = []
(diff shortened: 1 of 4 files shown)


packages/builder/src/components/design/settings/controls/ButtonActionEditor/actions/TriggerAutomation.svelte
```diff
@@ -32,8 +33,35 @@
...
+                return true
+              }
+            }
+          }
+        }
+
+        return false
+      }
+
+      let hasCollectBlock = hasCollectBlockRecursive(
```
melohagan:
```suggestion
      const hasCollectBlock = hasCollectBlockRecursive(
```

melohagan (approved):
LGTM

packages/shared-core/src/sdk/documents/automations.ts
```diff
@@ -15,3 +19,36 @@ export function isWebhookAction(automation: Automation) {
 export function isAppAction(automation: Automation) {
   return automation.definition.trigger?.stepId === AutomationTriggerStepId.APP
 }
+
+function hasCollectBlockRecursive(steps: any[]): boolean {
```
samwho: Go straight to `any` jail.

Don't we have an `AutomationStep` type?

packages/shared-core/src/sdk/documents/automations.ts
```diff
@@ -15,3 +19,36 @@ export function isWebhookAction(automation: Automation) {
...
+      step.stepId === AutomationActionStepId.BRANCH &&
+      step.inputs?.children
+    ) {
+      // Branch children is an object where keys are branch IDs and values are arrays of steps
+      for (const branchId in step.inputs.children) {
+        const branchSteps = step.inputs.children[branchId]
+        if (hasCollectBlockRecursive(branchSteps)) {
+          return true
+        }
+      }
```
samwho:
```suggestion
      for (const { children } of Object.values(step.inputs.children)) {
        if (hasCollectBlockRecursive(children)) {
          return true
        }
      }
```

samwho (approved):
Minor suggestions.

Issue #16192, opened 2025-05-23
External data source columns not properly hidden after re-fetch


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Self
  - Method: docker compose
  - Budibase Version: 
  - App Version: 

**Describe the bug**
When a new column is added to an external table that already exists in Budibase, re-fetching the data and viewing the table in the design screen shows the column. The new column appears in the UI, even though it's toggled off in the table's settings. To hide it properly, you have to toggle it on and then off again.

This becomes especially frustrating when the same data source is used across multiple screens, as the issue has to be manually fixed in each instance. Ideally, newly added columns should be hidden by default if their toggle is off.

**To Reproduce**
1. Add any external source: MySQL, MS SQL, Postgres
2. Fetch a table
3. Add a table component to a screen with your selected table as the source
4. Create a new column in the database directly
5. Re-fetch the table
6. Notice the setting for the new column is toggled off
7. Notice the new column is visible in the UI on the design screen even though this setting shows off

**Expected behavior**
Column shouldn't appear in the design area if the setting for the column is toggled off.

**Screenshots**
<img width="1208" alt="Image" src="https://github.com/user-attachments/assets/06546929-cf9f-436c-9914-c83b9b30316f" />

melohagan:
Difficulty: Medium

The row is that the table component isn’t re-applying column visibility metadata when a schema refresh happens. To resolve it we should make the data source sync process persist the hidden flag for each column (even if it’s newly discovered) and ensure the builder/rendering layer reads those flags rather than assuming new columns are visible. The effort spans the fetch/metadata merge logic plus the client-side rendering state so it needs regression checks on schema re-fetch.

PR #16532, opened 2025-07-08
Fixing addition of new columns when some are hidden

mike12345567:
## Description
Previously attempted to fix this, but there was a mis-communication in terms of what was desired - there was an issue causing new columns to be visible despite the settings suggesting they weren't. I've updated the rules in this area to follow the correct rule set:
1. Default everything visible if nothing has been set hidden
2. Hide all other columns if visibility has been set on any column.

## Addresses
https://linear.app/budibase/issue/BUDI-9348/external-data-source-columns-not-properly-hidden-after-re-fetch
https://github.com/Budibase/budibase/issues/16192


aptkingston (approved):
LGTM!

PR #16546, opened 2025-07-10
Nav update conflicts

adrinr:
## Description
Cleaning the code around navigation for the frontend. This was meant to be a temporally status, so we had a patch that was updating both workspace app and app metadata navigation if required. As this might stay longer, and we found a bug around this (where the store was not properly updated and we had some silent errors), we are moving this logic into the backend.

The logic is as it was:
1. If we are updating a navigation part of the default workspace app, we update both workspace app navigation and the app metadata
2. If it's not the default workspace app, we only update the workspace app

## Launchcontrol
Ensure default workspace app and app metadata navigation are always on sync

### DIFF
--- a/packages/server/src/api/routes/tests/navigation.spec.ts
+++ b/packages/server/src/api/routes/tests/navigation.spec.ts
@@ -62,7 +62,7 @@ describe("/navigation", () => {
       )
     })
 
-    it("should not app navigation app navigation when updating not default workspaces", async () => {
+    it("should not update app navigation app navigation when updating not default workspaces", async () => {
       const { workspaceApp: otherWorkspaceApp } =
         await config.api.workspaceApp.create({
           ...structures.workspaceApps.createRequest(),
--- a/packages/server/src/api/routes/tests/navigation.spec.ts
+++ b/packages/server/src/api/routes/tests/navigation.spec.ts
@@ -1,0 +1,109 @@
+import { structures } from "@budibase/backend-core/tests"
+import * as setup from "./utilities"
+import { AppNavigation, WithRequired, WorkspaceApp } from "@budibase/types"
+
+describe("/navigation", () => {
+  let request = setup.getRequest()
+  let config = setup.getConfig()
+  let workspaceApp: WithRequired<WorkspaceApp, "_id">
+
+  beforeAll(async () => {
+    await config.init()
+  })
+
+  beforeEach(async () => {
+    await config.newTenant()
+    const { workspaceApps } = await config.api.workspaceApp.fetch()
+    workspaceApp = workspaceApps[0]
+  })
+
+  afterAll(setup.afterAll)
+
+  const sampleNavigation: AppNavigation = {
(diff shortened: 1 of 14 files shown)


packages/server/src/api/routes/tests/navigation.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    it("should return 400 for invalid workspace app id", async () => {
+      await config.api.navigation.update(
+        "invalid_id",
+        { navigation: sampleNavigation },
+        { status: 400 }
+      )
+    })
+
+    it("should not app navigation app navigation when updating not default workspaces", async () => {
```
samwho: Typo?

samwho: Also it's not immediately clear to me why we wouldn't update navigation for non-default workspaces. I'm not terribly familiar with this part of Budibase, sorry 🙈

packages/server/src/api/routes/tests/navigation.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      })
+
+      const updatedApp = await config.api.workspaceApp.find(workspaceApp._id)
+      expect(updatedApp).toBeDefined()
+      expect(updatedApp!.navigation).toEqual(emptyNavigation)
+
+      const appMetadata = await config.api.application.get(config.getAppId())
+      expect(appMetadata.navigation).toEqual(emptyNavigation)
+    })
+  })
```
samwho, line 99: I wonder if it makes sense for us to have a test that asserts what should happen if either of the two writes to CouchDB were to fail. We don't want to leave data in an inconsistent state.

adrinr: I was thinking about this... As we don't have transactions, the only way would be queueing or having a manual revert. This would be overkill for this purpose, as it is only intended to maintain backwards compatibility.
Any ideas on how to achieve this without too much overengineering? I don't think we have this done anywhere. In fact, it is in my labday list

samwho: Can you bulk write them? Does that have any atomicity guarantees? 🤔

adrinr: No. Bulk write does not garantee transaction. The result of it is an array of documents, indicating if they were updated or the reason why not

packages/server/src/api/routes/tests/navigation.spec.ts
adrinr: This is done only for backwards compatibility for the feature, tbh I will clean it up soon as part of it has already been merge behind the feature flag. The idea is that the default workspace app (created by the migration) is a copy of the app metadata (where it lived this until now). We want to keep the default workspace app and metadata navigation changes sync, just in case we need to consume the app metadata for any reason (reverts, app downgrades, etc)

samwho (approved):

Issue #16540, opened 2025-07-09
Creating a few different columns as the first column for a table softlocks the table


ConorWebb96:
**Checklist**
- [x] I have searched budibase discussions and github issues to check if my issue already exists

**Hosting**
- Cloud
  - Tenant ID: cprem

**Describe the bug**
If you create any of the following fields first, it prevents you from creating other columns. This may be related to the primary display setting.

- Attachment
- Signature
- Multi-attachment
- Relationship
- single user
- multi user

Anytime a new column is created, an error will be thrown: `Error saving column: Column "table 1" cannot be used as a display type.`

**To Reproduce**
1. Import the below app export
2. Try to add a column to any of the tables

**Expected behavior**
Prevent these columns from being added as the first column.

**Screenshots**
<img width="1512" height="864" alt="Image" src="https://github.com/user-attachments/assets/8b676471-fd63-4b66-9f7d-921ef55b3624" />
  
**App Export**
[softlocked tables.tar.gz](https://github.com/user-attachments/files/21144238/softlocked.tables.tar.gz)

ConorWebb96:
fixed and merged, will be included `2.13.24`

PR #16549, opened 2025-07-10
Fix: added check to ensure certain columns couldn't be added as the first column of a table

ConorWebb96:
## Description
Specific columns are not allowed to be used as the primary display. Previously, it was possible to set them as the primary display by adding them as the first column in a new table.

This fix introduces a list of disallowed fields that cannot be selected when a table has no existing columns.

## Addresses
- https://github.com/Budibase/budibase/issues/16540

## Screenshots
**Before**
Screenshot 2025-07-10 at 14 09 25

**After**
Screenshot 2025-07-10 at 14 09 58

## Launchcontrol
Some columns aren't supposed to be used as the main one shown in a table. But there was a bug where it would still get set as the main one if you added one of these special columns first.

This fix ensures those columns aren't picked first when creating a new table.


samwho (changes requested):
We should also enforce this on the backend and have tests to make sure this behaviour gets tested for regression.

packages/server/src/api/controllers/table/index.ts
```diff
@@ -71,6 +71,21 @@ function checkDefaultFields(table: Table) {
 async function guardTable(table: Table, isCreate: boolean) {
   checkDefaultFields(table)
 
+  // Filter out protected/internal columns
+  const userSchemaKeys = Object.keys(table.schema || {}).filter(
```
adrinr: Why extracting this instead of checking the `table.primaryField` value? If this is done to get the "default" primary value automatically, can this not cause weird errors to the user, as they will not now what's going on? Would it not be better than, instead of picking the first one, picking the first "valid" one?

packages/builder/src/components/backend/DataTable/modals/CreateEditColumn.svelte
```diff
@@ -716,6 +724,10 @@
     getOptionValue={field => field.fieldId}
     getOptionIcon={field => field.icon}
     isOptionEnabled={option => {
+      // Disable certain fields if they are the first column
+      if (isFirstColumn && disallowedFirstFields.includes(option.type)) {
+        return false
+      }
```
samwho: Is it obvious in the UI why they're disabled? Can we have hover tooltips or such explaining it?

packages/builder/src/stores/builder/tables.ts
```diff
@@ -4,12 +4,29 @@ import {
...
+  excludeA?: string,
+  excludeB?: string
+): string | undefined {
+  const candidates = Object.keys(draft.schema).filter(n => {
+    if (n === excludeA || n === excludeB) return false
+    const f = draft.schema[n]
+    return f && isAllowedDisplayField(n, f.type)
+  })
+  return candidates[0]
+}
```
samwho:
```suggestion
function pickFallbackDisplayField(
  draft: SaveTableRequest,
  ...exclude: string[],
): string | undefined {
  for (const [name, field] of Object.entries(draft.schema)) {
    if (exclude.includes(name)) {
        continue
    }
    if (isAllowedDisplayField(name, field.type)) {
      return name
    }
  }
  return undefined
}
```

Couple of things here:
- It's a bit icky to have multiple optional params that serve ~the same purpose, better to have a vararg array at the end. More generalisable.
- No need to loop over the whole schema just to return the first thing we find, we can end the function when we find the thing.

packages/builder/src/stores/builder/tables.ts
```diff
@@ -174,14 +191,19 @@ export class TableStore extends DerivedBudiStore<
...
+
     if (hasPrimaryDisplay) {
       draft.primaryDisplay = field.name
-    } else if (draft.primaryDisplay === originalName) {
-      const fields = Object.keys(draft.schema)
-      // pick another display column randomly if unselecting
-      draft.primaryDisplay = fields.filter(
-        name => name !== originalName || name !== field.name
-      )[0]
+    } else if (!draft.primaryDisplay && isEligible) {
```
samwho: This `hasPrimaryDisplay` argument is weirdly named, and sort of not required? If I'm reading it right, it indicates whether this field being displayed is currently the primaryDisplay field, right?

Can't you derive this information from `draft.primaryDisplay`? Won't `wasPrimary` always be the same value as `hasPrimaryDisplay`?

packages/server/src/api/routes/tests/table.spec.ts
```diff
@@ -37,6 +39,10 @@ const { basicTable } = setup.structures
 const ISO_REGEX_PATTERN = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/
 
 const descriptions = datasourceDescribe({ plus: true })
+const disallowedDisplayColumns = Object.values(FieldType).filter(type => {
+  if (typeof type !== "string") return false
```
samwho: When is this ever the case?

samwho (approved):

PR #16550, opened 2025-07-10
Links not working

adrinr:
## Description
For some reason, probably due the issues we found around migrations, we are getting some users with screens without workspaceAppId. This causes problems, as this is meant to be a required fields. This might also be due some client upgrading and downgrading their instances, and we will need to explore it.
There are 2 changes done in this PR, being a patch until we release the migrations fix:
- When we fetch screens, if any of them has no workspace app id, we map it back to the default workspace app (this will fix builder issues)
- On the client route endpint, when checking for the default workspace app, we will check for both workspace app id matches and `empty` workspace apps (fixing client routing issues)

## Launchcontrol
Fix issue with broken links and some screens not being able to save

### DIFF
--- a/packages/server/src/utilities/routing/index.spec.ts
+++ b/packages/server/src/utilities/routing/index.spec.ts
@@ -1,0 +1,160 @@
+import { getRoutingInfo as sut } from "./index"
+
+import TestConfiguration from "../../tests/utilities/TestConfiguration"
+import { structures } from "@budibase/backend-core/tests"
+import { basicScreen } from "../../tests/utilities/structures"
+import { DesignDocument, Screen, WorkspaceApp } from "@budibase/types"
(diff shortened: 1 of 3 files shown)


packages/server/src/utilities/routing/index.spec.ts
```diff
@@ -0,0 +1,160 @@
+import { getRoutingInfo as sut } from "./index"
```
samwho, line 1: Feels like an odd name. What does it mean? "system under test"?

samwho (approved):

packages/server/src/utilities/routing/index.spec.ts
adrinr: Subject under test. It is (or it used to be 🧓🏽 ) a common unit test keyword

PR #16569, opened 2025-07-14
Add Endpoint and EndpointGroup mechanism for specifying API routes

mike12345567:
## Description
Previously we directly specified routes by create koa routers, adding routes to them and then eventually retrieving all these routes from the sub-routers and applying them to a main router. This worked, but if you specified something like `subRouter.use(middleware)` this would apply to all routes applied to the main router after that sub router was applied.

To get around this in the public API we create a list of routes to eventually apply, then apply them all to a single router, listing the middlewares for each of them individually - with the addition of the endpoint groups we have specify the middleware once and apply it to all the endpoints when they are added to the router.

This makes it a lot easier to apply middlewares to all routes in a resource path - which was the original reason for desiring this - if you want to add a protection like say recaptcha to a whole resource today, the easiest way to do it is specify it one by one on every route.

The endpoint group also has a `lockMiddleware` function which is used by the standard groups, making sure that no unexpected middlewares are added to these groups later (which would apply to every route within).


packages/backend-core/src/Endpoint/EndpointGroup.ts
```diff
@@ -0,0 +1,104 @@
+import Endpoint, { CtxFn, Method } from "./Endpoint"
+import Router from "@koa/router"
+
+type ArrayOneOrMore<T> = {
```
adrinr: That's a funny way to define this. I've usually seen it like `type ArrayOneOrMore<T> = [T,...T[]] `

packages/backend-core/src/Endpoint/Endpoint.ts
```diff
@@ -0,0 +1,74 @@
...
+  constructor(method: Method, url: string, controller: CtxFn) {
+    this.method = method
+    this.url = url
+    this.controller = controller
+    this.middlewares = []
+    this.outputMiddlewares = []
+  }
+
+  addMiddleware(middleware: CtxFn, opts?: { first?: boolean }) {
+    if (opts?.first) {
```
adrinr, line 36: Why using this `first` flag instead of actually relying on the addition order?

packages/server/src/api/routes/apikeys.ts
```diff
@@ -1,12 +1,6 @@
...
+import { builderGroup } from "./endpointGroups"
 
-const router: Router = new Router()
-
-router
-  .get("/api/keys", authorized(permissions.BUILDER), controller.fetch)
-  .put("/api/keys/:key", authorized(permissions.BUILDER), controller.update)
-
-export default router
+builderGroup
```
adrinr: Is `group` the right name? Would this not be clearer as `builderRoutes` or similar?

packages/backend-core/src/Endpoint/Endpoint.ts
mike12345567: the group middlewares are independent from the middlewares that have been added to the endpoint directly, generally the pre-middlewares will always come first and the post-middlewares will always come last. I wanted to provide a way to flip this behaviour if needs be, this is needed for the authorization middlewares, they come after the resource IDs have been picked out. I wanted a way to still apply these to the group, but to specify they should be added after the endpoint middlewares.

packages/server/src/api/routes/apikeys.ts
mike12345567: Yeah you're right, its basically just like the sub-routers, i'll update that now!

samwho (approved):
Scary but good change. Great work.

packages/backend-core/src/Endpoint/EndpointGroup.ts
mike12345567: That is actually a lot nicer lol

PR #16517, opened 2025-07-07
Fix MSSQL config validation

adrinr:
## Description
MSSQL config validation has been broken for a while for advanced configuration. This might have affected all nestedGroup configurations. This PR fixes it and ensures that we only validate what it actually needs to be validated.

#### Basic validation
https://github.com/user-attachments/assets/fbf361a6-34ec-444c-b404-7c183cd2b89e

#### Nested validation

https://github.com/user-attachments/assets/eb55cf2b-5a9d-446a-a3cb-f76a4858ca84

## Launchcontrol
Fix MSSQL config validation

PR #16574, opened 2025-07-16
Fix Google Sheet datasource creation.

samwho:
## Description

I believe this got broken in https://github.com/Budibase/budibase/pull/16517 but what's happening here is that we pull a parameter Google gives us after the OAuth flow from the query string (`?continue_google_setup`) and throw it into the config store, to be pulled out later and sent to the backend.

Adri's changes make this store more strict in its validation, removing anything that shouldn't belong on the config object, but sadly this meant the `?continue_google_setup` parameter got thrown out. But we need it in order to fetch a `refreshToken` from Google when we create a spreadsheet.

This PR very simply ensures that parameter gets preserved through validation.

## Launchcontrol

Fix the creation of new Google Sheet datasources.


packages/builder/src/components/backend/Datasources/ConfigEditor/stores/validatedConfig.ts
```diff
@@ -143,6 +143,14 @@ export const createValidatedConfigStore = (
 
       const config: Record<string, any> = {}
 
+      // During the auth flow, we get a continueSetupId from Google's OAuth flow
+      // that gets saved in the config store but is not something we expect the
+      // user to configure. It needs preserving in the validated config or no
+      // one is able to create a Google Sheets datasource.
+      if (integration.name === "GOOGLE_SHEETS") {
```
PClmnt: Could I be super annoying here and ask for a `IntegrationTypes.GOOGLE_SHEETS`

PClmnt (approved):

packages/builder/src/constants/backend/index.ts
```diff
@@ -226,23 +226,23 @@ export const DB_TYPE_INTERNAL = "internal"
 export const DB_TYPE_EXTERNAL = "external"
 
 export const IntegrationTypes = {
-  POSTGRES: "POSTGRES",
-  MONGODB: "MONGODB",
+  AIRTABLE: "AIRTABLE",
```
samwho, line 229: Sorted these alphabetically.

packages/types/src/sdk/datasources.ts
```diff
@@ -47,23 +47,23 @@ export enum DatasourceFieldType {
 }
 
 export enum SourceName {
-  POSTGRES = "POSTGRES",
+  AIRTABLE = "AIRTABLE",
```
samwho, line 50: Sorted these alphabetically.

PClmnt (approved):

PR #16575, opened 2025-07-17
Use timestamps again in migrations

adrinr:
## Description
Initially, we had the migrations running timestamp based. This was updated to be array-based as we got some corrupted version, where the timestamp could not be extracted.
This PR changes it back to be timestamp based, as this gives us more flexibility. Plenty of testing was added to ensure if works as expected


packages/server/src/appMigrations/index.ts
```diff
@@ -31,8 +31,16 @@ export function getLatestEnabledMigrationId(migrations?: AppMigration[]) {
...
 
-function getMigrationIndex(versionId: string) {
-  return MIGRATIONS.findIndex(m => m.id === versionId)
+export function getTimestamp(versionId: string): number {
+  const timestampStr = new RegExp(/^(?<timestamp>\d{14})_.+$/).exec(versionId)
+    ?.groups?.["timestamp"]
+
+  if (!timestampStr) {
+    return -1
+  }
```
samwho, line 40: Would it not be better to throw an exception in this case? Doesn't returning minus one put us in the same position that led us to remove timestamps in the first place?

packages/server/src/appMigrations/tests/migrations.integrity.spec.ts
```diff
@@ -22,4 +23,21 @@ describe("migration integrity", () => {
       }
     })
   })
+
+  it("all migrations should have a valid id with timestamps", () => {
+    for (const migration of migrations.MIGRATIONS) {
+      const timestamp = getTimestamp(migration.id)
+      expect(timestamp).toBeGreaterThan(1)
```
samwho, line 30: Okay, I see. I think even with this test I'd still throw an exception. I think it's better to blow up in this case than to apply a migration out of order if something goes drastically wrong.

PR #16576, opened 2025-07-17
Introduce a suite of tests that test upgrading between Budibase versions.

samwho:
## Description

This PR creates a new package: `packages/upgrade-tests`. This package is totally dedicated to running a set of tests that make sure upgrading between Budibase versions goes smoothly.

https://github.com/user-attachments/assets/45bc940c-6481-406a-a3a4-3a574e608917

You run it like so:

```
cd packages/upgrade-tests
yarn test:upgrade --from 3.9.0
```

It broadly works like so:

1. An older Budibase single image is spun up, in this case `budibase/budibase:3.9.0`.
2. A set of Jest tests are run against this Budibase instance.
3. The instance is spun down and a new instance, built from the tip of the current branch, replaces it.
4. Another set of Jest tests runs against that.

Here's an example of what the tests look like:

```typescript
import { BudibaseClient } from "../index"
import { it } from "../utils/upgradeTest"

describe("Automation Upgrade Tests", () => {
  let client: BudibaseClient

  beforeAll(async () => {
    client = await BudibaseClient.authenticated()
  })

  it("should preserve automation counts", {
    pre: async () => {
      const automations = await client.automation.fetch()
      return { automationCount: automations.length }
    },
    post: async ({ automationCount }) => {
      const currentAutomations = await client.automation.fetch()
      expect(currentAutomations.length).toBe(automationCount)
    },
  })
})
```

I've created a custom `it` function that takes a `pre` and `post` function. Whatever is returned from `pre` is saved to disk and then passed into `post`. This allows tests to verify before/after state, e.g. number of records, checksum of record content, and so on.

## CI

In CI, we run the upgrade tests against the last 5 minor versions of Budibase, and we run against all of the app exports we find in `packages/upgrade-tests/src/fixtures`.

I suspect this may get too large too quickly, so we may need to tweak the way this works in future.


.github/workflows/upgrade-tests.yml
```diff
@@ -0,0 +1,223 @@
...
+    paths:
+      # Server-side packages that affect upgrade compatibility
+      - "packages/worker/**"
+      - "packages/server/**"
+      - "packages/backend-core/**"
+      - "packages/shared-core/**"
+      # The upgrade tests themselves
+      - "packages/upgrade-tests/**"
+      # Docker and hosting changes that affect the build
+      - "hosting/**"
```
adrinr, line 14: We could also add the root `package.json` or the `yarn.lock` (even if this might cause too many unnecessary runs), as it might have some hoisting implications

packages/server/package.json
```diff
@@ -196,8 +196,12 @@
       },
       "build": {
         "inputs": [
+          "{projectRoot}/src/**/*",
```
adrinr: NIT:  I think that the "/**/*" part is redundant

packages/upgrade-tests/tsconfig.build.json
```diff
@@ -0,0 +1,21 @@
+{
```
adrinr, line 1: Do we need this config? Will this ever be "built for production"?

packages/upgrade-tests/src/api/application.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  async get(appId: string): Promise<App> {
+    const { data } = await this.client.get<App>(`/api/applications/${appId}`)
+    return data
+  }
+
+  async import(filePath: string, name?: string): Promise<string> {
+    // Read the file as a buffer
+    const fileBuffer = fs.readFileSync(filePath)
+    const filename = filePath.split("/").pop() || "imported-app.tar.gz"
```
adrinr: Is this default "imported-app.tar.gz" expected?

packages/upgrade-tests/src/api/row.ts
```diff
@@ -0,0 +1,11 @@
+import { Row } from "@budibase/types"
+import type { BudibaseClient } from "./BudibaseClient"
+
+export class RowAPI {
+  constructor(private client: BudibaseClient) {}
+
+  async fetch(tableId: string): Promise<Row[]> {
+    const { data } = await this.client.get<Row[]>(`/api/${tableId}/rows`)
```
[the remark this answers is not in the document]
adrinr: Our restfulness is being attacked...

packages/upgrade-tests/src/tests/automations.upgrade.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  beforeAll(async () => {
+    client = await BudibaseClient.authenticated()
+  })
+
+  it("should preserve automation counts", {
+    pre: async () => {
+      const automations = await client.automation.fetch()
+      return { automationCount: automations.length }
+    },
+    post: async ({ automationCount }) => {
```
adrinr, line 16: I like this idea of the pre output being the post input :)

packages/upgrade-tests/src/utils/upgradeTest.ts
```diff
@@ -0,0 +1,133 @@
...
+export function it<T extends JSONValue>(
+  testName: string,
+  options: UpgradeTestOptions<T>
+): void {
+  const { app, pre, post } = options
+
+  // Check if we should skip based on app filter
+  const currentApp = process.env.TEST_APP
+  if (app && currentApp && currentApp !== app) {
+    jestIt.skip(testName, () => {})
```
adrinr, line 52: What would be this scenario?

packages/upgrade-tests/src/utils/upgradeTest.ts
```diff
@@ -0,0 +1,133 @@
...
+  } else if (isPostUpgradePhase() && post) {
+    jestIt(testName, async () => {
+      const context = loadContext() as T
+      const result = post(context)
+      if (isPromise(result)) {
+        await result
+      }
+    })
+  } else {
+    jestIt.skip(testName, () => {})
```
adrinr: Should this ever happen?

adrinr (commented):
Good job, Claude 😄  Just some questions, more to understand what's going on than anything else

packages/upgrade-tests/src/api/application.ts
samwho: Bleh, no. Claude has a habit of making sure everything has a fallback value, where I'd much rather it threw an exception. It also kept adding fallbacks for the BUDIBASE_URL and I had to keep telling it off.

packages/upgrade-tests/tsconfig.build.json
samwho: Good question, I was copying the layout of our other packages. `tsconfig.json` depends on this file currently so I think strictly yes? Though if you think this should be structured differently let me know how and I'll do it.

packages/upgrade-tests/src/tests/automations.upgrade.test.ts
samwho: It's clean, isn't it? Allows you a lot of freedom in how you want to verify the upgrade was successful. Also wonderfully typesafe 😁

packages/upgrade-tests/src/utils/upgradeTest.ts
samwho: So, nothing uses this currently, but it's possible to limit your tests to specific apps. I'm envisioning a future where we have several app exports we're testing against, and people wanting to specifically target their tests to just one of them.

```typescript
it("should something", {
  app: "car-rental",
  pre: async () => { ... },
  post: async () => { ... },
})
```

packages/upgrade-tests/src/utils/upgradeTest.ts
samwho: No, good spot!

packages/upgrade-tests/tsconfig.build.json
adrinr: It's not really important, but in this case, `tsconfig.json` (used for dev and jest) and `tsconfig.build.config` (used for production builds could be merged, as otherwise, it implies we might have some production build somehow (this is where the initial thoughts came from)

packages/upgrade-tests/src/api/application.ts
adrinr: So should we remove this one?

adrinr (approved):
LGTM, great initial approach!

PR #16577, opened 2025-07-17
Improve migration integrity tests

adrinr:
## Description
Updating the way we run the migration integrity tests to use the `migrationProcessor`. As the logic in there changed, with migrations and app syncs, we want to replicate the same flow.

Along the way, I realised that we are updating the createdAt/updatedAt wrongly on user sync. Fixed it and added some tests for it


packages/server/src/appMigrations/tests/migrations.integrity.spec.ts
```diff
@@ -9,16 +12,55 @@ describe("migration integrity", () => {
...
+        expect(await getAppMigrationVersion(appId)).toBe(latestMigration)
 
-        await migration.func()
-        const latestDocs = await db.allDocs({ include_docs: true })
+        const latestDevDocs = await getDocs(devDb)
+        const latestProdDocs = await getDocs(prodDb)
 
-        expect(docs).toEqual(latestDocs)
+        expect(latestDevDocs).toEqual(devDocs)
+        expect(latestProdDocs).toEqual(prodDocs)
```
samwho: Are these valid assertions? Aren't migrations by definition going to change some documents?

packages/server/src/sdk/users/tests/utils.spec.ts
```diff
@@ -62,6 +64,96 @@ describe("syncGlobalUsers", () => {
     })
   })
 
+  it("app users are synced", async () => {
```
samwho, line 67: This test seems unrelated to the PR?

samwho: Ah you mention this in the description, my mistake.

packages/server/src/appMigrations/tests/migrations.integrity.spec.ts
adrinr: This is what this test is testing. Each migration should be idempotent, and you should be able to run it multiple times. If it succeeded once, subsequent runs should not change anything (as everything is migrated already). I renamed the variables to make it more verbose

samwho (approved):

packages/server/src/appMigrations/tests/migrations.integrity.spec.ts
samwho: Ahhh I totally misunderstood, the rename is much better. Thank you 🙏

PR #16581, opened 2025-07-17
Convert worker APIs to groups

mike12345567:
## Description
As with app service conversion, doing the same for the worker APIs! Previous work can be found here: https://github.com/Budibase/budibase/pull/16569

Moved some bits into backend-core, did a little bit of renaming to make it a bit more consistent with how its used and then updated all routes to follow in the same way.

Last thing to do is to convert pro to do something similar, allowing pro to expose groups of routes for each of the APIs it offers, but i'll do this at a later point - just wanted the worker to be consistent for now.


packages/worker/src/api/routes/endpointGroups/standard.ts
```diff
@@ -0,0 +1,22 @@
+import cloudRestricted from "../../../middleware/cloudRestricted"
```
adrinr, line 1: This is nice :)

packages/worker/src/api/routes/system/status.ts
```diff
@@ -1,8 +1,4 @@
-import Router from "@koa/router"
 import * as controller from "../../controllers/system/status"
+import { loggedInRoutes } from "../endpointGroups"
 
-const router: Router = new Router()
-
-router.get("/api/system/status", controller.fetch)
-
-export default router
+loggedInRoutes.get("/api/system/status", controller.fetch)
```
adrinr, line 4: I saw a couple of these. Was this requiring being logged in previously? Can it cause some weird 403?

mike12345567: every endpoint in the worker has a logged in middleware on it, there are a few endpoints that don't, but they are managed in a slightly different way. I'd like to fix that now that we have a proper way of managing routes, but I didn't want to do that in this PR.

Right now if you look in `worker/src/api/index.ts` you'll see it just lists a bunch of endpoints that are public, ones that don't require tenancy etc. Realistically we can break down the `loggedInRoutes` to three, `publicRoutes`, `noTenancyRoutes` and `loggedInRoutes` - removing the middlewares covering every route simplifying.

packages/worker/src/api/routes/global/users.ts
```diff
@@ -69,92 +72,69 @@ function buildChangeTenantOwnerEmailValidation() {
...
   )
   .post(
     "/api/global/users/multi/invite/delete",
-    auth.builderOrAdmin,
     controller.removeMultipleInvites
   )
+  .post("/api/global/users/invite/update/:code", controller.updateInvite)
+  .get("/api/global/users/invites", controller.getUserInvites)
+  // global endpoint but needs to come at end (blocks other endpoints otherwise)
+  .get("/api/global/users/:id", controller.find)
```
samwho, line 120: Isn't this gonna get reordered by the sort that takes colons into account?

samwho (changes requested):

packages/worker/src/api/routes/global/users.ts
mike12345567: It is, thats just an old comment - I'll remove it.

adrinr (approved):
LGTM!

PR #16583, opened 2025-07-18
App level Recaptcha V2 support

mike12345567:
## Description
Implements support for for Recaptcha V2 on apps - if an app has this enabled then it cannot be accessed unless the user has been through the recaptcha flow. This protects the complete app rather than protecting individual forms similar to how Stack overflow works.

There is two new settings inputs, one for configuring your recaptcha keys and one to enable recaptcha on the app that you'd like to protect (its not enabled for every app globally).

There is a new screen within the client that will be brought up if this is enabled, the user has an enterprise license and they do not have recaptcha verification. The verification takes the form of a cookie which will expire after a period of time.

Pro PR to add the license features: https://github.com/Budibase/budibase-pro/pull/446

## Screenshots
Button to disable/enable Recaptcha support in the app level settings:
image

This has a lock icon and no options if the user is not on the enterprise license.

If recaptcha token is not found:
image

Once accepted user is sent onto the app itself. This will expire after 30 minutes and they will be returned to this screen to re-validate.

Recaptcha settings menu globally
image


packages/builder/src/pages/builder/portal/settings/recaptcha.svelte
```diff
@@ -0,0 +1,103 @@
...
+
+  async function saveConfig() {
+    try {
+      loading = true
+      const config = get(values)
+      let recaptchaConfig: RecaptchaConfig = {
+        type: ConfigType.RECAPTCHA,
+        config: {
+          siteKey: config.siteKey!,
+          secretKey: config.secretKey!,
```
samwho: Could we validate this instead of banging it?

packages/client/src/components/RecaptchaV2.svelte
```diff
@@ -0,0 +1,47 @@
+<script>
```
samwho: ts pls

packages/client/src/components/RecaptchaV2.svelte
```diff
@@ -0,0 +1,47 @@
...
+    })
+  })
+
+  async function handleSubmit(event) {
+    event.preventDefault()
+    if (!token) {
+      throw new Error("No token provided")
+    }
+    const { verified } = await API.recaptcha.verify(token)
+    if (verified) {
```
samwho, line 29: Should we do something if this is false?

packages/server/src/api/controllers/recaptcha.ts
```diff
@@ -0,0 +1,75 @@
...
+    "https://www.google.com/recaptcha/api/siteverify",
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+      },
+      body: new URLSearchParams({
+        secret: config.config.secretKey,
+        response: token,
+      }),
```
samwho: Kinda weird seeing URLSearchParams as the body, is this correct/normal?

packages/server/src/api/controllers/static/templates/BudibaseApp.svelte
```diff
@@ -132,6 +132,7 @@
       <p />
     {/if}
   </div>
+  <script src="https://www.google.com/recaptcha/api.js"></script>
```
samwho: Should we only load this if they have the feature enabled? I can imagine some people being a little upset about this being unconditionally loaded.

packages/server/src/api/routes/tests/recaptcha.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+jest.mock("@budibase/backend-core", () => {
+  const actual = jest.requireActual("@budibase/backend-core")
+  return {
+    ...actual,
+    configs: {
+      ...actual.configs,
+      getRecaptchaConfig: () => mockConfig,
+    },
+  }
+})
```
samwho, line 26: Could we actually create the config instead of mocking?

packages/server/src/tests/utilities/TestConfiguration.ts
```diff
@@ -494,9 +494,24 @@ export default class TestConfiguration {
...
 
+    let cookie: (string | string[])[] = [
+      `${constants.Cookie.Auth}=${authToken}`,
+    ]
+    const tempHeaderCookie = this.temporaryHeaders?.["Cookie"]
+    const hasAuth = Array.isArray(tempHeaderCookie)
+      ? tempHeaderCookie.find(cookie => cookie.includes(constants.Cookie.Auth))
+      : typeof tempHeaderCookie === "string"
+        ? tempHeaderCookie.includes(constants.Cookie.Auth)
+        : false
```
samwho: Nested ternaries are really hard to read, could we split this out?

packages/server/src/tests/utilities/api/recaptcha.ts
```diff
@@ -0,0 +1,30 @@
+import { VerifyRecaptchaRequest } from "@budibase/types"
+import { Expectations, RequestOpts, TestAPI } from "./base"
+import { Response } from "supertest"
+
+export class RecaptchaAPI extends TestAPI {
+  verify = async (
+    request: Partial<VerifyRecaptchaRequest>,
+    expectations?: Expectations
+  ): Promise<Response> => {
```
samwho: Could we type this response more specifically?

packages/server/src/tests/utilities/api/recaptcha.ts
```diff
@@ -0,0 +1,30 @@
...
+    return await this._requestRaw("post", "/api/recaptcha/verify", {
+      body: request,
+      expectations,
+    })
+  }
+
+  check = async (
+    cookie?: string,
+    expectations?: Expectations
+  ): Promise<Response> => {
```
samwho: And this.

packages/server/src/utilities/redis.ts
```diff
@@ -114,3 +117,14 @@ export function getSocketPubSubClients() {
...
+export async function setRecaptchaVerified() {
+  const id = utils.newid()
+  await recaptchaClient.store(id, { verified: true }, RECAPTCHA_SESSION_SECONDS)
+  return id
+}
+
+export async function isRecaptchaVerified(sessionId: string) {
+  const session = await recaptchaClient.get(sessionId)
+  return !!(session && session.verified)
+}
```
samwho, line 133: How come we need this? Could we not set a TTL on the cookie?

packages/worker/src/api/controllers/global/configs.ts
```diff
@@ -273,6 +276,21 @@ export async function processAIConfig(
   }
 }
 
+export async function processRecaptchaConfig(
+  config: RecaptchaInnerConfig,
+  existingConfig?: RecaptchaInnerConfig
+) {
+  if (!(await pro.features.isRecaptchaEnabled())) {
+    throw new ForbiddenError("License does not allow use of recaptcha")
+  }
```
samwho, line 285: Is this going to cause problems if people enable it then downgrade and lose the flag on their license?

packages/server/src/api/routes/tests/recaptcha.spec.ts
mike12345567: I explained this in the comment above it - configs are all managed in the worker service today, we don't really have an easy way to create them without either moving all the logic to manage configs into backend-core so that it could be shared for testing, or simply trying to fake it by force creating the document by hand - I'm not really sure its much better to actually create the document vs just mocking the document being returned.

packages/server/src/utilities/redis.ts
mike12345567: This is potentially incase we need to support extending the session, I'd prefer to manage it on the backend and then we can update they are still verified while they use the app - there hasn't been a decision about whether it should just always expire after a period of time, or whether it should stay valid until the user stops using the app.

packages/worker/src/api/controllers/global/configs.ts
mike12345567: No, the middleware becomes inactive if they don't have recaptcha enabled, so once they downgrade, recaptcha is essentially fully disabled, they can't set/remove the config anymore, but recaptcha is non-functional anyway.

packages/server/src/api/controllers/recaptcha.ts
mike12345567: This appears to be the recommended way to fill out a form, you could encode the URI elements manually yourself, but for the `x-www-form-urlencoded` content type this is the best approach!

packages/server/src/api/routes/tests/recaptcha.spec.ts
samwho: Ahhh that's fair, I completely forgot about that.

packages/client/src/components/RecaptchaV2.svelte
```diff
@@ -0,0 +1,52 @@
...
+      notifications.error("Recaptcha verification failed, please try again")
+    }
+  }
+</script>
+
+<div class="wrapper">
+  <Body>Human verification step required to continue.</Body>
+  <div id="recaptcha-container" />
+  <div>
+    <Button cta on:click={handleSubmit}>Send</Button>
```
mikesealey: NAB: I feel like "send" isn't the right terminology here. Perhaps "Submit".

samwho (approved):

PR #16602, opened 2025-07-23
Fix out of sync migrations

adrinr:
## Description
Recent changes modified the migrationProcessor to keep in sync prod and dev. As this was not the case before, we have already out-of-sync apps (where only dev or prod has been migrated). As this is not expected, the current code did not support it.

## Launchcontrol
Fix migration systems for half-migrated apps

### DIFF
--- a/packages/server/src/appMigrations/migrationsProcessor.ts
+++ b/packages/server/src/appMigrations/migrationsProcessor.ts
@@ -8,11 +8,73 @@ import { AppMigration, doInMigrationLock } from "."
 import { MIGRATIONS } from "./migrations"
 import sdk from "../sdk"
 
+async function getPendingMigrationsForApp(
+  appId: string,
+  allMigrations: AppMigration[]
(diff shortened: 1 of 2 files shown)


packages/server/src/appMigrations/migrationsProcessor.ts
```diff
@@ -8,11 +8,73 @@ import { AppMigration, doInMigrationLock } from "."
...
+}: {
+  migrationId: string
+  migrationFunc: () => Promise<void>
+  appId: string
+}): Promise<void> {
+  await context.doInAppMigrationContext(appId, async () => {
+    console.log(`Running migration "${migrationId}" for app "${appId}"`)
+    await migrationFunc()
+    console.log(`Migration "${migrationId}" ran for app "${appId}"`)
+  })
```
samwho: Worth wrapping this in a `tracer.trace`?

samwho (approved):

PR #16604, opened 2025-07-23
Flesh out the suite of upgrade tests

samwho:
## Description

This PR primarily adds a bunch more upgrade tests for our core domain objects:

- Configs
- Datasources
- Queries
- Plugins
- Roles
- Rows
- Screens
- Users
- Views

In the process of doing this I've added some new helpers to ease the writing of these tests:

- `shouldNotChange` is an alternative to `it` that simplifies asserting some data should not change between upgrades.
- `upgradeSpansCommit` and `upgradeSpansVersion` are helpers that allow you to check if the upgrade include a given commit or version, so that logic can be modified if changes are expected.


packages/upgrade-tests/src/api/BudibaseClient.ts
```diff
@@ -24,13 +32,21 @@ export interface RequestOptions {
 }
 
 export class BudibaseClient {
-  private config: ClientConfig
+  private _config: ClientConfig
```
adrinr, line 35: Out of curiosity, why prefixing it?

packages/upgrade-tests/src/tests/automations.upgrade.test.ts
```diff
@@ -1,21 +1,30 @@
...
-    pre: async () => {
-      const automations = await client.automation.fetch()
-      return { automationCount: automations.length }
-    },
-    post: async ({ automationCount }) => {
-      const currentAutomations = await client.automation.fetch()
-      expect(currentAutomations.length).toBe(automationCount)
-    },
+  shouldNotChange("IDs", async () => {
+    const automations = await client.automation.fetch()
```
adrinr, line 13: I don't get these tests. Are they checking that the ids and other properties are not modified? Is this a real case scenario?

packages/upgrade-tests/src/api/BudibaseClient.ts
samwho: So that I can expose `config` as the `ConfigAPI`.

packages/upgrade-tests/src/tests/automations.upgrade.test.ts
samwho: The thinking is to make sure we don't accidentally delete something or make it otherwise not return from the API. It felt to me like if we asserted the entire structure of each automation these tests would just fail whenever we made a migration against things, so I wanted to make sure nothing catastrophic happened without introducing a source of friction.

mike12345567 (approved):
LGTM - i like the approach of breaking down the various resources into specific upgrade tests. Hopefully this means we could run a lot of scenarios like this in parallel.

PR #16605, opened 2025-07-23
Automation State block

deanhannigan:
## Description
Added a new `State` block to automations. Named variables can be now set in state and updated throughout the automations runtime. 
- A new `State` category has been added to the bindings surfacing all available state variables.
- State can now be seen in the automation context on the bottom right of the screen.
- `FIX` - Added the looping feature flag to the AI steps: Classify, Extract doc data, Generate text. In order to utilise the looping behaviour the steps will need to be recreated.
- `FIX` - Fix for looping UI so that it only shows when steps are configured with the feature flag.

## Screenshots
Below I have declared a variable called `myVar1`. In my step, `Sample Step 1`, I initialise the value and then update it in `Sample Step 4`. Before running the automation, all the `State` steps that modify a state variable will be listed. 
Screenshot 2025-07-23 at 16 36 14

## Launchcontrol
Added new state block to automations


packages/server/src/automations/tests/steps/extractState.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    const failureStep = results.steps[0]
+    expect(failureStep.outputs.success).toBe(false)
+    expect(results.status).toEqual(AutomationStatus.ERROR)
+
+    const successStep = results.steps[1]
+    expect(successStep.outputs.success).toBe(true)
+    expect(successStep.outputs.value).toBe(5)
+
+    // The failure should not cause a total automation failure
```
samwho, line 91: Why not?

samwho (approved):

packages/server/src/automations/tests/steps/extractState.spec.ts
deanhannigan: Hey Sam, thanks for taking a look! What I meant was that, a failure in the state block shouldn't cause any unhandled exceptions and that any state data that was correctly processed should remain intact.

mike12345567 (changes requested):
Sorry for time taken to get to this - the code looks good! I had a play with it and it works the way I would expect, one thing I noticed though, the value is called "code"

image

I think Key and Value are better terms to mimic the experience we have as part of the button:
image

mike12345567:
I also tried using a loop with the state block and it is just saying it throws an error which is a little odd - I'll investigate this further tomorrow!

image

mike12345567 (approved):
Gave this a test - its great! It'll be fantastic once we have the looping block in and the capability to exit branches, such a useful little utility!

Tested it with adding up array of numbers, worked really well - love it!

PR #16615, opened 2025-07-25
Fix scroll on automation side-bar-nav class.

ConorWebb96:
## Description
After x amount of automations were added, the automation list couldn't be scrolled through. This adds a max height so the overflow: auto can scroll through the list. `- 118px` accounts for the nav, side-bar-controls height, and padding applied to the top of the automation list.

## Addresses
- Customer report

## Screenshots
**Before**

**After**

## Launchcontrol
Fixed an issue where you couldn't scroll through an extensive list of automations.


deanhannigan (approved):
LGTM! Nice one!

PR #16621, opened 2025-07-25
Add workspace app disabling and updating status API

mike12345567:
## Description
This returns the workspace apps feature to using the old system for publishing, but now allows for workspace apps to be disabled.

This is pretty much the same experience as we have for automations today and I'll use the new updates to the status API as part of the index pages.

The disabled toggle can be seen in the preview top nav when inside a workspace app.
image

When publishing the popover now provides information about the automations/apps it has published as well:
image

I've also updated the language around automations from "Paused" and "Activated" to "Disabled" and "Enabled" as the word paused makes very little sense for apps and I didn't want them to use different wording.

When an app is disabled it is not returned as part of the app package and no routing information is provided for it - this stops it from being renderable.

The status API doesn't have a purpose in this PR but I'll update the index pages to use the state that it now returns.

I have disabled most of the filtered publishing functionality but haven't removed it entirely from the backend as the functionality could be useful, but I have removed the frontend components for now. The resource usage analysis on the backend is still quite useful for detecting if there are unpublished changes or not.


packages/bbui/src/Form/Core/Switch.svelte
```diff
@@ -14,7 +15,7 @@
   }
 </script>
 
-<div class="spectrum-Switch spectrum-Switch--emphasized">
+<div class="spectrum-Switch spectrum-Switch--emphasized" class:noPadding>
```
adrinr, line 18:
```suggestion
<div class="spectrum-Switch spectrum-Switch--emphasized" class:noPadding={noPadding}>
```

packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPanel.svelte
```diff
@@ -39,8 +40,33 @@
...
+                noPadding
+                on:change={() =>
+                  workspaceAppStore.toggleDisabled(
+                    selectedWorkspaceAppId,
+                    !selectedWorkspaceApp?.disabled
+                  )}
+                value={!selectedWorkspaceApp?.disabled}
+              />
+              <Label
+                >{selectedWorkspaceApp?.disabled
```
Comment (adrinr) [inline packages/builder/src/pages/builder/app/[application]/design/[screenId]/_components/AppPanel.svelte]: OCD, but can this be idented as 
```
<Label>
...
</Label>
```

adrinr (commented):
LGTM, just a bbui comment that needs to be addressed

packages/bbui/src/Form/Core/Switch.svelte
mike12345567: This is short-hand for the same thing.

adrinr: Nvm, I was still sleeping

adrinr (approved):
🚀

PR #16622, opened 2025-07-28
Move away from anonymous functions for middleware

samwho:
## Description

Anonymous function middleware shows up in DataDog with the span name "anonymous", whereas named functions get given names. As it stands today, most of our middleware is anonymous and it's can be tricky to know what's going on in the middleware portion of DataDog spans.

This PR gives all middleware functions names to remedy this.


packages/server/src/middleware/ensureTenantAppOwnership.ts
```diff
@@ -1,7 +1,7 @@
 import { tenancy, utils, context } from "@budibase/backend-core"
 import { UserCtx } from "@budibase/types"
 
-async function ensureTenantAppOwnership(ctx: UserCtx, next: any) {
+export async function ensureTenantAppOwnership(ctx: UserCtx, next: any) {
```
adrinr: Should this not be renamed to `ensureTenantAppOwnershipMiddleware` to follow the same pattern?

packages/server/src/middleware/trimViewRowInfo.ts
```diff
@@ -4,7 +4,7 @@ import sdk from "../sdk"
 import { Next } from "koa"
 import { getSourceId } from "../api/controllers/row/utils"
 
-export default async (ctx: Ctx<Row, Row>, next: Next) => {
+export async function trimViewRowInfo(ctx: Ctx<Row, Row>, next: Next) {
```
adrinr: Same, renaming to `trimViewRowInfoMiddleware`?

adrinr (approved):
Just a question on why some functions were renamed to be "middleware" prefixed and others not

PR #16624, opened 2025-07-28
Serialize the upload and download of directories to S3.

samwho:
## Description

The goal here is to control peak memory use of the Budibase app service container. For large apps, with thousands of attachments, kicking off thousands of download streams uses a lot of memory at peak. Serializing it will make the download slower, but gives us more confidence that peak memory use shouldn't pass the stream high watermark in NodeJS, which defaults to 64kb.

### DIFF
--- a/packages/backend-core/src/objectStore/objectStore.ts
+++ b/packages/backend-core/src/objectStore/objectStore.ts
@@ -318,25 +318,34 @@ export async function retrieve(
   bucketName: string,
   filepath: string
 ): Promise<string | stream.Readable> {
-  const objectStore = ObjectStore()
-  const params = {
-    Bucket: sanitizeBucket(bucketName),
-    Key: sanitizeKey(filepath),
-  }
-  const response = await objectStore.getObject(params)
-  if (!response.Body) {
-    throw new Error("Unable to retrieve object")
-  }
-  if (STRING_CONTENT_TYPES.includes(response.ContentType)) {
-    return response.Body.transformToString()
-  } else {
-    // this typecast is required - for some reason the AWS SDK V3 defines its own "ReadableStream"
-    // found in the @aws-sdk/types package which is meant to be the Node type, but due to the SDK
-    // supporting both the browser and Nodejs it is a polyfill which causes a type clash with Node.
-    const readableStream =
-      response.Body.transformToWebStream() as ReadableStream
-    return stream.Readable.fromWeb(readableStream)
-  }
+  return await tracer.trace("retrieve", async span => {
+    span.addTags({ bucketName, filepath })
+    const objectStore = ObjectStore()
+    const params = {
+      Bucket: sanitizeBucket(bucketName),
+      Key: sanitizeKey(filepath),
+    }
+    const response = await objectStore.getObject(params)
+    if (!response.Body) {
+      throw new Error("Unable to retrieve object")
+    }
+    span.addTags({
+      contentLength: response.ContentLength,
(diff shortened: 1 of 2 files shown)


packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -388,50 +415,61 @@ export async function getPresignedUrl(
...
       })
-    )
-  }
-  await Promise.all(writePromises)
-  return writePath
+    }
+
+    span.addTags({ numObjects })
+    return writePath
+  })
```
samwho: This function has been reworked. `listAllObjects` is now an async generator, to prevent loading thousands of file names into memory at once, and instead of gathering all promises together and calling Promise.all on them, we await them individually.

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -503,26 +541,26 @@ export async function uploadDirectory(
...
           stream: fs.createReadStream(local),
         })
-      )
+      }
     }
-  }
-  await Promise.all(uploads)
-  return files
+    return files
+  })
```
samwho, line 553: This function got the same treatment as `retrieveDirectory`. On top of serializing it, I have also replaced uses of `fs` with `fsp` where possible, to prevent blocking the event loop on disk IO.

samwho (commented):
Other than these two comments, the rest of the PR is adding DataDog traces.

mike12345567 (approved):
Good addition of all the traces, at the very least we'll learn quite a bit more about how all of this operates!

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -388,50 +415,61 @@ export async function getPresignedUrl(
...
+  return await tracer.trace("retrieveToTmp", async span => {
+    span.addTags({ bucketName, filepath })
+    bucketName = sanitizeBucket(bucketName)
+    filepath = sanitizeKey(filepath)
+    const data = await retrieve(bucketName, filepath)
+    const outputPath = join(budibaseTempDir(), v4())
+    span.addTags({ outputPath })
+    if (data instanceof stream.Readable) {
+      span.addTags({ stream: true })
+      data.pipe(fs.createWriteStream(outputPath))
```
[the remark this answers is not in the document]
samwho: Ooh this is a good point, cc @mike12345567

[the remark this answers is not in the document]
mike12345567: Very good point and very good catch!

packages/backend-core/src/objectStore/objectStore.ts
```diff
@@ -438,7 +437,7 @@ export async function retrieveDirectory(bucketName: string, path: string) {
     span.addTags({ bucketName, path })
 
     let writePath = join(budibaseTempDir(), v4())
-    await fsp.mkdir(writePath)
+    await fsp.mkdir(writePath, { recursive: true })
```
mike12345567, line 440: Why does `mkdir` need to be recursive after the pipeline change?

samwho: It doesn't, this was another CoPilot suggested change from https://github.com/Budibase/budibase/pull/16624#discussion_r2236800772.

mike12345567: Ah I see, I suppose it probably doesn't create any problems!

mike12345567 (approved):

PR #16629, opened 2025-07-29
Close banners

adrinr:
## Description
Closing the new banners on the index pages. The state of "closeness" will be persisted on localStorage for now. In the future, if these were to stay for longer, we can check if it's worth to store it in the DB, to prevent cache cleanups and different browsers from popping out again

https://github.com/user-attachments/assets/8b564c0f-8c53-4fb2-bfc7-775b7dd15eed

### DIFF
--- a/packages/builder/src/stores/portal/banners.test.ts
+++ b/packages/builder/src/stores/portal/banners.test.ts
@@ -1,0 +1,121 @@
+import { it, expect, describe, beforeEach, afterEach, vi } from "vitest"
+import { get } from "svelte/store"
+import { BannerType } from "@/constants/banners"
+import { BannerStore } from "./banners"
+
+describe("BannerStore", () => {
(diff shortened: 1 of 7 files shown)


packages/builder/src/stores/portal/banners.test.ts
```diff
@@ -0,0 +1,121 @@
+import { it, expect, describe, beforeEach, afterEach, vi } from "vitest"
+import { get } from "svelte/store"
+import { BannerType } from "@/constants/banners"
+import { BannerStore } from "./banners"
+
+describe("BannerStore", () => {
+  const STORAGE_KEY = BannerStore.STORAGE_KEY_PREFIX
```
deanhannigan, line 7: 👨‍🍳

deanhannigan (approved):
LGTM! Gave it a quick run through there and everything works perfectly 👍

PR #16632, opened 2025-07-29
Support status correctly on index pages

mike12345567:
## Description
As title, adds "Published", "Unpublished" and "Disabled" as possible states for workspace apps and automations.


adrinr (approved):
LGTM

PR #16637, opened 2025-07-29
Deduplicate workspace apps

adrinr:
## Description
Adding a new migration to deduplicate workspace apps caused by a bug around migrations. This will ensure we can move forward with a unique default workspace app


packages/server/src/appMigrations/migrations/20250729134531_workspace_cleanups.ts
```diff
@@ -0,0 +1,32 @@
+import { context } from "@budibase/backend-core"
+import sdk from "../../sdk"
+import { Document } from "@budibase/types"
+
+const migration = async () => {
+  const allWorkspaceApps = await sdk.workspaceApps.fetch()
+  const defaultWorkspaceApps = allWorkspaceApps.filter(a => a.isDefault)
+  if (defaultWorkspaceApps.length > 1) {
+    const [workspaceAppToUse, ...toDelete] = defaultWorkspaceApps.sort((a, b) =>
```
samwho, line 9: Forgive me, I haven't followed this work super closely. Is there a possibility that the user has modified one of the duplicate workspace apps and this selection might not choose the correct one? Should we perhaps select by the updatedAt field?

packages/server/src/sdk/app/workspaceApps/crud.ts
```diff
@@ -37,8 +37,8 @@ export async function create(
 
   const response = await db.put(
     {
-      _id: docIds.generateWorkspaceAppID(),
       ...workspaceApp,
+      _id: docIds.generateWorkspaceAppID(),
```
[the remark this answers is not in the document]
adrinr: Nah, in this case we want the opposite. A creation should always generate a new id

packages/server/src/appMigrations/migrations/20250729134531_workspace_cleanups.ts
adrinr: This should not be needed, as we did not add anything that allows modifying the workspace app but not the app metadata. But I implemented it as you said just in case :)

mike12345567 (approved):
LGTM - hopefully this solves the dupe workspace issue once and for all!

PR #16641, opened 2025-07-29
Make sure tooltip position is at the bottom for banners

melohagan:
## Description
Make sure tooltip position is at the bottom for banners

## Addresses
- https://linear.app/budibase/issue/BUDI-9532/minor-tooltip-issue

## Launchcontrol
Minor tooltip fix for banners

### DIFF
--- a/packages/builder/src/components/portal/licensing/licensingBanners.js
+++ b/packages/builder/src/components/portal/licensing/licensingBanners.js
@@ -73,6 +73,7 @@ const buildUsageInfoBanner = (
     criteria: () => {
       return appLicensing?.usageMetrics[metricKey] >= percentageThreshold
     },
+    showCloseButton: false,
     tooltip: appLicensing?.quotaResetDate,
   }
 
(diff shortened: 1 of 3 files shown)


packages/builder/src/components/portal/licensing/licensingBanners.js
```diff
@@ -73,6 +73,7 @@ const buildUsageInfoBanner = (
     criteria: () => {
       return appLicensing?.usageMetrics[metricKey] >= percentageThreshold
     },
+    showCloseButton: false,
```
adrinr, line 76: Is this expected?

adrinr (approved):
Small comment, but LGTM

packages/builder/src/components/portal/licensing/licensingBanners.js
melohagan: Yeah, we want the usage banner to be permanent if you're reaching a limit.

PR #16642, opened 2025-07-30
Index pages - automations

adrinr:
## Description
Implement index page automations:
- Displaying right last update
- Display trigger type
- Implement filters
- Avoid duplicate names while duplicating automations
- Implement pause/activate


packages/builder/src/pages/builder/app/[application]/automation/_flagged/index.new.svelte
```diff
@@ -24,21 +27,38 @@
...
+    {
+      label: "Published",
+      filterValue: PublishResourceState.PUBLISHED,
+    },
+    {
+      label: "Disabled",
+      filterValue: PublishResourceState.DISABLED,
+    },
+    {
+      label: "Drafts",
```
Comment (deanhannigan) [inline packages/builder/src/pages/builder/app/[application]/automation/_flagged/index.new.svelte]: Should this be "Unpublished" too? Or should the labels on the automation list say "Drafts"?

deanhannigan (approved):
LGTM! One minor question but I didn't run into any issues during testing 🎉

packages/builder/src/pages/builder/app/[application]/automation/_flagged/index.new.svelte
Comment (adrinr) [reply packages/builder/src/pages/builder/app/[application]/automation/_flagged/index.new.svelte]: Good catch. We will probably need to discuss all these terms, but for now I will align them

PR #16643, opened 2025-07-30
Improve put retrievedoc typing

adrinr:
## Description
Improve typing and fix small issue (that would have been detected by the typing) related to the following PR:
- https://github.com/Budibase/budibase/pull/16625


samwho (approved):
Oof, I didn't notice this. Great catch.

